PR #3608, opened 2026-07-05
docs(command-palette): add subcomponent examples and playground defaults

mohitWeb-lab:
## Summary
Adds missing subcommponent example templates for CommandPaletteInput and CommandPaletteList, trims duplicate composed-example content out of the CommandPaletteGroup/CommandPaletteItem showcases, and adds playground defaults for the CommandPalette subcomponents. Also includes two small incidental fixes found while working in the docsite preview code.

## Case

- CommandPaletteInput and CommandPaletteList had no dedicated example blocks (basic + showcase), so they had no docsite preview coverage and no CLI template output.
- CommandPaletteGroupShowcase and CommandPaletteItemShowcase each rendered two examples stacked in a Stack (a data-driven one and a composed one), which duplicated content that the new CommandPaletteList examples now cover on their own, and made the showcase previews cluttered.
- The CommandPaletteEmpty, CommandPaletteGroup, CommandPaletteInput, CommandPaletteItem, and CommandPaletteList .doc.mjs files had no playground.defaults, so the interactive props playground rendered them with empty/awkward default props.
- Separately: InteractivePreview's formatValue called JSON.stringify directly on arbitrary prop values (e.g. functions, circular refs) coming from playground defaults, which throws and breaks code generation for any component whose defaults aren't trivially serializable — including the new CommandPaletteGroup/CommandPaletteList array-of-element defaults added here.
- Separately: ComponentPreviewTheme only called registerIcons inside the component's render, so on SSR the server's warm module registry could already contain the theme's Lucide icons while a fresh client render started from defaults — causing an Icon strokeWidth mismatch between server and client markup during hydration.

## Fix
- Added CommandPaletteInputBasic/Showcase and CommandPaletteListBasic/Showcase .tsx + .doc.mjs template pairs under packages/cli/templates/blocks/components/.
- Simplified CommandPaletteGroupShowcase and CommandPaletteItemShowcase down to a single focused example each, removing the now-redundant composed-list variant and unused Stack/Text imports.
- Added playground.defaults to the five CommandPalette*.doc.mjs files in packages/core/src/CommandPalette/ so the docsite playground renders sensible sample content out of the box.
- Wrapped the JSON.stringify calls in InteractivePreview.formatValue in try/catch, falling back to a /* ... */ placeholder so non-serializable prop values no longer break code generation.
[...]

PR #3880, opened 2026-07-12
feat(docsite): surface prop type definitions in the props table

### LINKED ISSUE #2682: [Bug] [bug-bash] Docsite: no way to inspect non-primitive prop types (e.g. XDSSearchSource<T>)
### Description

On a component's **Properties** tab, props are listed with their TypeScript types — but for non-primitive types (e.g. `XDSSearchSource<T>`, generics, custom interfaces, unions), there's no way to inspect what the type actually is. You see the name and have to go grep the repo to learn anything.

Ask: surface type definitions on the docsite. Two options that would solve it:

1. A **Types** tab on each component page that lists every non-primitive type referenced by its props, with the resolved shape (fields, generics, etc.).
2. Or, simpler: make each type name on the Properties tab a **link to source** (or to its definition in some types index) — a "view code" link.

Either would unblock people reading the docsite without needing to clone the repo.

### Reproduction

1. Navigate to any component with a non-primitive prop type, e.g. https://xds-sandbox.vercel.app/components/BaseTypeahead
2. Open the **Properties** tab
3. Notice props typed as `XDSSearchSource<T>` (and similar) — there's no way to drill into what that type contains

### XDS Version

Whatever is currently deployed at xds-sandbox.vercel.app

### Environment

Chrome, macOS, https://xds-sandbox.vercel.app

jiunshinn:
## Scope

Smallest useful slice of #2682: on a component page's **Properties** tab, any prop whose documented type references a named type exported from the component's own package (e.g. `SearchSource<T>`, `TableColumn<T>[]`) now gets a **"View {Type}"** trigger that opens the type's actual TypeScript declaration — with its member JSDoc and source path — in a popover. No Types tab, no cross-package resolution, no recursive expansion; those are listed as follow-ups below so the direction can be steered in review.

Fixes #2682

## How the props-table pipeline works today (why the fix lands where it does)

- Prop types are **hand-authored strings** in each component's `.doc.mjs` (`packages/core/src/docs-types.ts:36` — "TypeScript type signature as a string"), so there is no type information to inspect at render time.
- `apps/docsite/scripts/generate-data.mjs` (`generateComponentRegistry`) imports those doc files at build time and emits `src/generated/componentRegistry.ts` (gitignored).
- The Properties tab renders `PlaygroundPropsTable` (`apps/docsite/src/components/component-detail/ComponentDetailClient.tsx:228`), which prints `prop.type` as plain text (`PlaygroundPropsTable.tsx` → `formatType`). For `SearchSource<T>` that's a dead end — the shape lives in `packages/core/src/Typeahead/types.ts` and readers had to clone the repo to see it.

Since the doc strings can't be resolved in the browser, the resolution happens in the generator, where the package **source** is already on disk for both content targets (the live workspace on canary, and the published tarballs on latest — they ship `src/`, see `resolve-content-root.mjs`).

## What this PR does

**Generator** (`apps/docsite/scripts/generate-data.mjs` + new shared module `apps/docsite/src/lib/typeDefinitions.mjs`, following the `src/lib/blog/posts.mjs` precedent of build-time logic shared with tests):

- Scans each documented package's `src/` for `export interface` / `export type` declarations and extracts their source text. The parser is lexical (brace/angle matching that skips strings and comments) in the same spirit as the generator's existing regex extraction — no TS compiler dependency in the data pipeline.
- For every prop, resolves which extracted names its `type` string references and emits `PropDoc.typeRefs` plus a per-entry `ComponentEntry.typeDefs` (`{name, definition, sourcePath}`).
- Exclusions: `*Props` interfaces (documented by their own pages) and shape-less compat aliases like `export type StyleXStyles = any;` (would otherwise put a useless "any" popover on nearly every `xstyle` row).

**UI** (`PlaygroundPropsTable.tsx`): a `PropTypeDefinitions` block under the prop description renders one `Popover` per referenced type — source path caption + declaration in a `CodeExampleBlock` — mirroring the existing `Popover` + `Card padding={0}` + `CodeExampleBlock` pattern from `PackageActions.tsx`. Works in both the desktop and mobile row layouts.

Registry cost on canary: 166 props across 75 entries pick up refs; ~30 KB of definition text total (largest single definition: `TableColumn`, ~4 KB, scrolls inside the popover via `maxHeight`).

## Before / after

- **Before:** `searchSource` on `/components/BaseTypeahead?tab=properties` shows `SearchSource<T>` as inert text.
- **After:** the row shows a `View SearchSource` ghost button; opening it shows `packages/core/src/Typeahead/types.ts` and the full `export interface SearchSource<…> { search(…); bootstrap(…); cancel?(); }` declaration with member JSDoc, copyable. Verified in the running docsite (screenshot available on request — driven via browser automation).

## Verification

- `pnpm -F @astryxdesign/docsite test` — 17 files, **228 passed** (includes new `src/__tests__/type-definitions.test.ts`: 11 tests covering the extractor, the index, ref resolution, and generated-registry invariants).
- `pnpm -F @astryxdesign/docsite typecheck` — only the 3 pre-existing `@astryxdesign/charts`/`lab` template-page errors remain (present on `main`); with theme packages built there are no other errors, none in changed files.
- `eslint` on changed files — clean.
- Manual: ran `next dev`, exercised `/components/BaseTypeahead?tab=properties` (popover opens, Escape dismisses, no console errors).

No changeset — docsite-only change (per #3608/#3744 precedent).

## Follow-ups (intentionally out of scope)

- Hook pages: apply the same treatment to `HookSignature` params/returns.
- One-level expansion only: `SearchSource`'s definition mentions `SearchableItem`, which isn't itself expandable yet (the generator could attach transitive refs).
- Cross-package refs (e.g. a charts prop referencing a core type) don't resolve; the index is per-package.
- A "view on GitHub" link from `sourcePath` (the data is already in the registry).

### DIFF
--- a/apps/docsite/scripts/generate-data.mjs
+++ b/apps/docsite/scripts/generate-data.mjs
@@ -24,6 +24,10 @@ import * as path from 'node:path';
 import {fileURLToPath, pathToFileURL} from 'node:url';
 import {resolveContentRoot} from './resolve-content-root.mjs';
 import {expandWorkspaceDirs} from '../../../scripts/lib/workspace-globs.mjs';
+import {
+  buildTypeDefinitionIndex,
+  collectPropTypeRefs,
+} from '../src/lib/typeDefinitions.mjs';
 
 const __dirname = path.dirname(fileURLToPath(import.meta.url));
 const DOCSITE_ROOT = path.resolve(__dirname, '..');
@@ -657,6 +661,31 @@ async function generateComponentRegistry() {
       }
     }
 
+    // Surface the shape of non-primitive prop types (issue #2682): when a
+    // documented prop type references a named type exported from this
+    // package's source (e.g. `SearchSource<T>`), record the reference on the
+    // prop and attach the extracted declaration to the entry so the props
+    // table can render it on demand.
+    const typeIndex = buildTypeDefinitionIndex(
+      pkg.srcDir,
+      path.relative(CONTENT_ROOT, pkg.srcDir),
+    );
+    for (const comp of components) {
+      const referenced = new Map();
+      for (const prop of comp.props) {
+        const typeRefs = collectPropTypeRefs(prop.type, typeIndex);
+        if (typeRefs.length > 0) {
+          prop.typeRefs = typeRefs;
+          for (const name of typeRefs) {
+            referenced.set(name, typeIndex.get(name));
+          }
+        }
+      }
+      comp.typeDefs = [...referenced.values()].sort((a, b) =>
+        a.name.localeCompare(b.name),
+      );
+    }
+
     components.sort((a, b) => a.name.localeCompare(b.name));
     if (components.length > 0) {
       allComponents[pkg.name] = components;
@@ -673,6 +702,18 @@ export interface PropDoc {
   default?: string;
   required?: boolean;
   slotElements?: ElementDescriptor[];
+  /** Names of package-exported types referenced by \`type\`, resolvable
+   *  against the owning entry's \`typeDefs\`. */
+  typeRefs?: string[];
+}
+
+export interface TypeDefinition {
+  /** Exported type name, e.g. \`SearchSource\`. */
+  name: string;
+  /** Extracted TypeScript declaration source, including member JSDoc. */
+  definition: string;
+  /** Repo-relative source file, e.g. \`packages/core/src/Typeahead/types.ts\`. */
+  sourcePath: string;
 }
 export interface BestPractice {
   guidance: boolean;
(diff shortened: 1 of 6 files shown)


cixzhang:
image

Ooh this is a really neat feature! Perhaps we can make the definition name more like a link that opens the popover? Ex clicking the "SearchSource<T>" here could open the definition.

jiunshinn:
Thanks @cixzhang! Great call — done in 6f744ae: the type name itself is now the trigger (`SearchSource` in `SearchSource<T>` renders as an inline link-styled button that opens the definition popover, reusing the docsite's authored-docs link treatment), and the separate "View {Type}" button is removed. Only the resolved name is clickable; surrounding generics/union syntax stays plain text, and each name in a multi-ref type is independently clickable.

cixzhang (approved):
Really neat feature. Thanks for adding it!

Issue #2863, opened 2026-06-16
XDSText/XDSHeading: support an accent text color (--color-text-accent)


rubyycheung:
## Problem

`XDSText` (and `XDSHeading`, which shares the same `color` type) has no way to render text in the theme's **accent text** color. The `color` prop is typed as:

```ts
// packages/core/src/theme/types.ts
export type XDSTextColor =
  | 'primary'
  | 'secondary'
  | 'disabled'
  | 'placeholder'
  | 'active'
  | 'inherit';
```

There's no `'accent'` value. The closest option, `'active'`, maps to **`--color-accent`** (the accent *surface/base* color), not **`--color-text-accent`** (the dedicated accent *text* ink):

```ts
// packages/core/src/Text/text.stylex.ts
active: {
  color: colorVars['--color-accent'],
},
```

These are intentionally different tokens — `--color-text-accent` is tuned for legible accent-colored text (often a deeper/adjusted accent), while `--color-accent` is the interactive/base accent. So today there is no first-class, type-safe way to say "make this text the accent text color."

## Why it matters

- `--color-text-accent` exists precisely for accent-colored text, but no component prop exposes it — consumers can't use the design system's own accent-text token through the normal `color` API.
- The only workaround is an `xstyle`/inline override hardcoding `color: 'var(--color-text-accent)'`, which bypasses the typed `color` prop, duplicates token knowledge at call sites, and is easy to get wrong (e.g. reaching for `active` and silently getting `--color-accent` instead).
- Accent-colored headings/links/labels are a common need (hero headlines, callouts, emphasized inline text).

## Concrete example

Setting a hero headline to the accent text ink currently requires:

```tsx
const styles = stylex.create({
  accentText: {color: 'var(--color-text-accent)'},
});

<XDSText type="display-3" xstyle={styles.accentText}>Little joys</XDSText>
```

Desired:

```tsx
<XDSText type="display-3" color="accent">Little joys</XDSText>
<XDSHeading level={2} color="accent">Section</XDSHeading>
```

(This came up while theming the docsite `theme-showcase` hero headline, which had to fall back to an `xstyle` override for exactly this reason.)

## Proposed change

Add an `'accent'` value to `XDSTextColor` that maps to `--color-text-accent`:

- `packages/core/src/theme/types.ts` — add `'accent'` to the `XDSTextColor` union.
[...]

PR #3881, opened 2026-07-12
feat(core): rename Switch labelSpacing "default" to "hug", keep deprecated alias

### LINKED ISSUE #2889: XDSSwitch: rename labelSpacing "default" value to something more descriptive
## Description

For `XDSSwitch`, the `labelSpacing` prop has a value named **`default`**, which doesn't describe what it actually does.

## Proposed change
- Rename the `default` value to something more descriptive — a name that suggests its behavior, similar to how the `spread` value is named. It should remain the **default value**, just with a more meaningful name.
- Optionally, if it makes more sense, **`spread` could be renamed to `spaceBetween`** to align with CSS property naming.

## Component
- `XDSSwitch`

jiunshinn:
## Summary

Renames the `Switch` `labelSpacing="default"` value to `labelSpacing="hug"`, keeping `"default"` as a fully working deprecated alias so the change is non-breaking. Behavior and rendered DOM are identical for every value.

Fixes #2889

## Why `hug`

The issue asks for a name that suggests the behavior, in the same register as `spread`. Evidence from the codebase:

- `hug` is the established Astryx vocabulary for "content packs together" layout behavior, and it is the default value in both places it exists:
  - `packages/core/src/TabList/TabListContext.ts:27` `export type TabListLayout = 'hug' | 'fill';` (doc: "each tab hugs its content width", default `'hug'` at `TabList.tsx:122`)
  - `packages/core/src/SegmentedControl/SegmentedControlContext.ts:17` `export type SegmentedControlLayout = 'hug' | 'fill';` (default `'hug'` at `SegmentedControl.tsx:148`)

  These are the only other layout-packing enums in core, so `'hug' | 'spread'` reads as one system: the label hugs the switch, or the pair is spread to opposite ends.
- The alternatives don't fit this axis: `compact`/`balanced`/`spacious` are three-level density scales (`ItemDensity`, `ChatDensity`, `TreeListDensity`, `CollapsibleGroupDensity`), not a two-state packing behavior, and `adjacent`/`snug` have zero precedent in core.
- The API Conventions wiki requires camelCase values, and the Contributing-with-AI-Assistants wiki explicitly calls out `default` as an enum value name that misleads LLMs, which is extra motivation to retire it.

The issue also floats renaming `spread` to `spaceBetween`. I kept that out of scope: `spread` already satisfies the "describes its behavior" bar, is used at ~20 internal call sites, and renaming it would double the migration surface. Happy to file a follow-up if maintainers want it.

## Non-breaking deprecation

`SwitchLabelSpacing` becomes `'hug' | 'spread' | 'default'` with a JSDoc `@deprecated` tag on the `'default'` member, mirroring the existing keep-the-alias precedents in core (`TopNavHeading` `href` deprecated in favor of `headingHref` at `TopNavHeading.tsx:233`, `Grid` `minChildWidth` deprecated in favor of `columns` at `Grid.tsx:70`). Inside the component, `'default'` is normalized to `'hug'` at the top of render, so:

- `hug`, `default`, and the omitted prop render byte-identical DOM (no `data-label-spacing` attribute, no variant class), exactly as `'default'` did before.
- `spread` is untouched: still emits `data-label-spacing="spread"` plus the `spread` variant class, so existing theme selectors keep working.

## Upgrade codemod

Following the value-rename codemod precedent (`v0.1.2/rename-text-color-active-to-accent.mjs` from #2863 and `v0.0.14/rename-section-wash-to-muted.mjs`), this adds `v0.1.5/rename-switch-label-spacing-default-to-hug.mjs`:

- Rewrites `labelSpacing="default"` (string, expression container, ternary/logical, `as const`) on `Switch` only.
- Rewrites `{labelSpacing: 'default'}` object literals and Storybook `argTypes` options, gated on a `Switch` import so unrelated `'default'` strings (an extremely common literal) are never touched.
- Registered under `0.1.5` on the assumption it ships in the next patch release (pre-1.0 forces patch bumps). If release numbering shifts, the registry entry can be moved, as was done in 3e3c16619 for v0.0.13.

## Internal usage migration

A repo-wide sweep for `labelSpacing` found zero internal `labelSpacing="default"` call sites; every existing usage passes `"spread"` (CLI templates `SwitchSettingsPanel.tsx`, `settings-sidebar/page.tsx`, `settings-dialog/page.tsx`, and `Switch.stories.tsx`), so no call-site migration was needed. The only `'default'` reference was the Storybook `labelSpacing` argTypes options list, migrated to `'hug'`. `Switch.doc.mjs` (en, zh, and dense variants) now documents `'hug' | 'spread'` with the deprecated alias noted.

## Verification

- `pnpm exec vitest run packages/core/src/Switch/Switch.test.tsx`: 38 passed (4 new `labelSpacing` tests cover the default, explicit `hug`, `spread`, and the deprecated `default` alias rendering identically to `hug`).
- CLI codemod + upgrade test files: 581 passed, including 11 new codemod tests (rename cases plus safety cases proving unrelated `'default'` strings and non-Switch components are untouched).
- `pnpm -F @astryxdesign/core typecheck`: clean. `eslint` on all touched files: clean. `check-changesets`: 74 valid.
- `node packages/cli/bin/astryx.mjs component Switch --dense` now renders `labelSpacing` as `'hug' | 'spread'` with default `'hug'`.
- Pre-existing, unrelated local failures (storybook typecheck on unbuilt `@astryxdesign/charts`/theme packages, flaky CLI api-discovery timeouts under full-suite parallelism) reproduce on clean main and are untouched by this diff.

## Questions for reviewers

1. Removal timeline: when may the `'default'` alias be dropped from `SwitchLabelSpacing`? At 1.0, or at the next planned breaking window? I left no removal date in the JSDoc on purpose.
2. Should the optional `spread` -> `spaceBetween` rename from the issue be pursued as a follow-up issue, or declined?
3. Is `v0.1.5` the right registry slot for the codemod, given pre-1.0 patch-only releases?

## Changesets

- `@astryxdesign/core` (feat): the rename plus deprecated alias.
- `@astryxdesign/cli` (feat): the upgrade codemod.


cixzhang:
Thanks for adding the codemod. We still need a clearer deprecation framework for how long old API stay in the codebase for version cut offs. While we're in beta I think we would clear it in 0.2.0 so perhaps closer to end of Q3.

cixzhang (approved):

PR #3882, opened 2026-07-12
feat(AspectRatio): add fit prop for component-owned child sizing

### LINKED ISSUE #2753: AspectRatio — child must set its own sizing/positioning styles
**Component:** `XDSAspectRatio`
**Type:** Enhancement — API ergonomics

### Description

`XDSAspectRatio` maintains the ratio on its container and wraps children in an absolutely-positioned fill `div`, but the **child content still has to set its own sizing and positioning** (`width`, `height`, `objectFit`, positioning, etc.). Every consumer and example repeats this boilerplate, e.g.:

```tsx
<XDSAspectRatio ratio={16 / 9}>
  <img
    src="…"
    alt="…"
    style={{width: '100%', height: '100%', objectFit: 'cover'}}
  />
</XDSAspectRatio>
```

This leaks layout responsibility onto the child and is easy to get wrong. See `packages/core/src/AspectRatio/XDSAspectRatio.tsx` — the inner `child` style only positions the wrapper `div`; the actual content element is left unstyled.

### Proposed fix

**Move these sizing/positioning props onto `XDSAspectRatio` itself**, so the parent controls how the child is laid out rather than requiring the child to set them. This must support both modes:

- **Fill / stretch** — child fills the container (`width/height: 100%` + `object-fit: cover`).
- **Centered** — child sits centered inside the container at its natural/intended size rather than stretched.

So `XDSAspectRatio` should expose props (e.g. a fit/alignment prop plus `objectFit`/`objectPosition` where relevant) that it applies to the child, instead of the child having to declare `width`/`height`/`objectFit`/positioning itself. The component owns the layout; the consumer just picks the behavior. Update the AspectRatio examples/templates to use the new props once available.

---
_Split out from the tracking issue #2631 (item 2 of 17)._

jiunshinn:
## Summary

`AspectRatio` maintains the ratio on its container and absolutely positions a fill wrapper, but the child content still has to hand-write its own sizing (`width`/`height`/`objectFit`). Every consumer repeats this boilerplate — every `img` block template, every story, every doc example:

```tsx
<AspectRatio ratio={16 / 9}>
  <img src="…" alt="…" style={{width: '100%', height: '100%', objectFit: 'cover'}} />
</AspectRatio>
```

This adds a `fit` prop so the component owns the child layout and the consumer just picks the behavior:

```tsx
<AspectRatio ratio={16 / 9} fit="cover">
  <img src="image.jpg" alt="Widescreen image" />
</AspectRatio>
```

- `fit="cover"` — child fills the box, media cropped (`object-fit: cover`)
- `fit="contain"` — child fills the box, media letterboxed
- `fit="center"` — child keeps its natural size, centered (the issue's second mode)
- omitted — child left unstyled, the pre-existing contract

Fixes #2753

## Mechanism (and why not `cloneElement`)

The obvious implementation — merging styles onto the child element — is banned here: the `@astryx/no-react-introspection` lint rule rejects `cloneElement()`/child-prop access (`internal/eslint-plugin-astryx/no-react-introspection.js:48-50`), and core has zero `cloneElement` call sites. The rule's own guidance points at CSS-based solutions, and the repo already has exactly one precedent for "component emits an attribute, a hand-authored stylesheet supplies the baseline": the `[data-astryx-media]` color-scheme rules in `reset.css` (`packages/core/src/reset.css:401-407`, emitted by `theme/MediaTheme.tsx`).

`fit` follows that mechanism:

- The component marks the child's **direct parent** (the fill wrapper) with `data-astryx-aspect-ratio-override="<fit>"` (`packages/core/src/AspectRatio/AspectRatio.tsx`). `fit` is structural, so it is deliberately **not** on the theming surface — `themeProps()` reflects only `shape` (per review: theming `fit` made no sense, and a marker on the direct parent is more stable).
- `reset.css` gains an "AspectRatio Fit Baseline" section keyed on that marker with **direct-child selectors** — `:where([data-astryx-aspect-ratio-override="cover"]) > :where(img, video)` — zero-specificity via `:where()` like every other reset rule, and with no dependence on AspectRatio's internal element structure. The marker carries the fit value itself (rather than a boolean marker plus a separate value attribute) so one attribute keeps the selectors direct-child while still distinguishing `cover` from `contain`. StyleX has no descendant selectors, so this cannot live in the component's `stylex.create` block.
- `fit="center"` needs no child styling at all — the component's own fill wrapper becomes a flex-centering box (`styles.childCenter`), plain StyleX.

Trade-off worth flagging: `cover`/`contain` child sizing only applies when `reset.css` is imported. That import is step 1 of the documented setup (`packages/cli/docs/getting-started.doc.mjs`) and every harness (sandbox, docsite, storybook) already loads it, but a consumer who skipped the reset gets the previous behavior (child styles itself) rather than the new sizing. Happy to move the rules elsewhere if there's a better hand-authored CSS home.

## Backward compatibility

Explicitly analyzed, since existing children all self-style today:

| Case | Behavior |
|---|---|
| No `fit` (all existing usage) | No `data-astryx-aspect-ratio-override` marker is emitted (React omits undefined attributes), so none of the new CSS can match. Byte-identical child rendering. |
| `fit` set, child self-styles via `style={{…}}` | Inline styles beat any stylesheet rule — child wins. |
| `fit` set, child self-styles via classes (StyleX/Tailwind) | The baseline is `:where()` zero-specificity inside `@layer reset`, the lowest layer; any consumer rule (layered or not, any specificity) wins. |
| `fit` set, non-element / multiple children | Rules only target direct children of the marked wrapper; `object-fit` only applies to `img`/`video`. Text/fragment children render unchanged. |
| Child props | Never touched — no prop injection, no className merging, SSR-safe (pure CSS, no effects). Covered by the "never touches the child element props" test. |

## Changes

- `packages/core/src/AspectRatio/AspectRatio.tsx` — `fit` prop + `AspectRatioFit` type, `data-astryx-aspect-ratio-override` marker on the fill wrapper, flex-centering wrapper style for `center`, JSDoc + file header
- `packages/core/src/reset.css` — AspectRatio Fit Baseline section (3 rules, `:where()` zero specificity)
- `packages/core/src/AspectRatio/index.ts` — export `AspectRatioFit`
- `packages/core/src/AspectRatio/AspectRatio.doc.mjs` — `fit` prop (EN/ZH/dense), best-practice entry (theming targets list only `shape`); also restores the `shape` prop entry that was dropped from the props table in the un-prefix migration (it was still referenced by `theming.targets`)
- `packages/core/src/AspectRatio/AspectRatio.test.tsx` — 6 fit contract tests + 3 reset.css baseline assertions (same pattern as `theme/onMediaTokens.test.ts:204`)
- `apps/storybook/stories/AspectRatio.stories.tsx` — stories use `fit="cover"`, new `FitModes` story, `fit` control
- `packages/cli/templates/blocks/components/AspectRatio/*.tsx` — 5 example blocks drop the child boilerplate (the issue's "update the examples" ask)
- `.changeset/aspectratio-fit-prop.md` — `@astryxdesign/core` patch

## Test plan

- `pnpm exec vitest run packages/core/src/AspectRatio/` — 24/24 pass (10 new)
- `pnpm exec eslint` on all changed files — clean (the earlier `cloneElement` draft was rejected by `@astryx/no-react-introspection`, which is what steered the design to the reset-baseline mechanism)
- `node scripts/check-sync.js`, `check:changesets`, `check:package-boundaries` — clean (run in pre-commit)
- `pnpm -F @astryxdesign/core build` — green
- Storybook typecheck — no errors in changed stories (pre-existing `@astryxdesign/vega` module errors unrelated)

### DIFF
--- a/packages/core/src/AspectRatio/AspectRatio.tsx
+++ b/packages/core/src/AspectRatio/AspectRatio.tsx
@@ -5,12 +5,13 @@
 /**
  * @file AspectRatio.tsx
  * @input Uses React, stylex
- * @output Exports AspectRatio component and AspectRatioProps
+ * @output Exports AspectRatio component, AspectRatioProps, AspectRatioShape, AspectRatioFit
  * @position AspectRatio component; maintains a specific aspect ratio for its children
  *
  * SYNC: When modified, update these files to stay in sync:
  * - /packages/core/src/AspectRatio/AspectRatio.doc.mjs
  * - /packages/core/src/AspectRatio/AspectRatio.test.tsx
+ * - /packages/core/src/reset.css (AspectRatio fit baseline rules)
  * - /apps/storybook/stories/AspectRatio.stories.tsx
  * - /packages/cli/templates/blocks/components/AspectRatio/ (showcase blocks)
  */
@@ -29,6 +30,16 @@ import {themeProps} from '../utils/themeProps';
  */
 export type AspectRatioShape = 'rectangle' | 'ellipse';
 
+/**
+ * How the child is sized and positioned inside the ratio box.
+ * - `cover`: the child fills the box; media is cropped to preserve its own
+ *   aspect ratio (`object-fit: cover`).
+ * - `contain`: the child fills the box; media is letterboxed to stay fully
+ *   visible (`object-fit: contain`).
+ * - `center`: the child keeps its natural size and sits centered in the box.
+ */
+export type AspectRatioFit = 'cover' | 'contain' | 'center';
+
 export interface AspectRatioProps extends BaseProps<HTMLDivElement> {
   /** Ref forwarded to the root element */
   ref?: React.Ref<HTMLDivElement>;
@@ -41,19 +52,43 @@ export interface AspectRatioProps extends BaseProps<HTMLDivElement> {
    * The shape of the container. Both shapes respect the provided `ratio`.
    * - `rectangle` (default): a standard rectangular container.
    * - `ellipse`: clips the container to an ellipse — a circle when `ratio={1}`,
-   *   or an oval at other ratios. Pair with a child that fills the container
-   *   (e.g. an image with `objectFit: 'cover'`).
+   *   or an oval at other ratios. Pair with `fit="cover"` so the media fills
+   *   the clipped container.
    *
    * @default 'rectangle'
    *
    * @example
    * ```
-   * <AspectRatio ratio={1} shape="ellipse">
-   *   <img src="avatar.jpg" alt="" style={{objectFit: 'cover'}} />
+   * <AspectRatio ratio={1} shape="ellipse" fit="cover">
+   *   <img src="avatar.jpg" alt="" />
    * </AspectRatio>
    * ```
    */
   shape?: AspectRatioShape;
+
+  /**
+   * How the child is laid out inside the ratio box, so the child does not
+   * have to declare `width`/`height`/`objectFit` itself.
+   * - `cover`: child fills the box; media is cropped (`object-fit: cover`).
+   * - `contain`: child fills the box; media is letterboxed
+   *   (`object-fit: contain`).
+   * - `center`: child keeps its natural size, centered in the box.
+   *
+   * `cover`/`contain` child sizing ships as zero-specificity baseline rules
+   * in `reset.css`, keyed on the reflected `data-fit` attribute (the same
+   * mechanism as the `data-astryx-media` baseline). Any styles the child
+   * sets itself still win, so children that already size themselves keep
+   * their behavior. When omitted, the child is left unstyled (the
+   * pre-existing contract).
+   *
+   * @example
+   * ```
+   * <AspectRatio ratio={16 / 9} fit="cover">
+   *   <img src="image.jpg" alt="Widescreen image" />
+   * </AspectRatio>
+   * ```
+   */
+  fit?: AspectRatioFit;
 
   /**
    * Content to render inside the aspect ratio container.
@@ -82,6 +117,15 @@ const styles = stylex.create({
     width: '100%',
     height: '100%',
   },
+  // fit="center" centers the child at its natural size from the wrapper —
+  // no styles on the child itself. The `cover`/`contain` child sizing can't
+  // live here (StyleX has no descendant selectors); it ships as baseline
+  // rules in reset.css keyed on the reflected `data-fit` attribute.
+  childCenter: {
+    display: 'flex',
+    alignItems: 'center',
+    justifyContent: 'center',
+  },
 });
 
 /**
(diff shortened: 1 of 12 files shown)


cixzhang:
Thanks for working on this! The solutions seems reasonable to me -- using reset as our sort of escape hatch from strict component-level styling. That sheet is becoming something other than just resetting some browser default styling but the precedent is there for media.

@mellyeliu in case you have ideas on how we can accomplish this better with stylex. The key piece is that we need to reach in and set styles on a direct child img from the AspectRatio component.

packages/core/src/AspectRatio/AspectRatio.tsx
```diff
@@ -133,7 +182,10 @@ export function AspectRatio({
         {...style, aspectRatio: ratio},
       )}
       {...props}>
-      <div {...stylex.props(styles.child)}>{children}</div>
+      <div
```
cixzhang, line 192: We may want to specially mark this container so it's easier to target for the special styling in reset.css

packages/core/src/reset.css
```diff
@@ -200,6 +200,34 @@
...
+   * Child sizing for AspectRatio's `fit` prop — the component reflects the
+   * prop as `data-fit` and these rules size the child, so consumers don't
+   * hand-write width/height/object-fit on every media child. Same mechanism
+   * as the [data-astryx-media] baseline below: the component emits the
+   * attribute, the reset supplies the zero-specificity baseline, and any
+   * styling the child sets itself still wins. `fit="center"` needs no rules
+   * here — the component centers the child from its own wrapper.
+   */
+  :where(.astryx-aspect-ratio[data-fit="cover"], .astryx-aspect-ratio[data-fit="contain"])
+    > :where(div)
```
cixzhang: The extra div here depends on AspectRatio's internal structure so I'm worried it may not stay in sync over time and lead to accidental regressions. If we mark the div directly instead of relying on fit at the theme we can apply a more direct child override.

packages/core/src/AspectRatio/AspectRatio.tsx
cixzhang: Maybe something like "data-astryx-aspect-ratio-override" to avoid collisions

packages/core/src/AspectRatio/AspectRatio.tsx
```diff
@@ -123,7 +172,7 @@ export function AspectRatio({
     <div
       ref={ref}
       {...mergeProps(
-        themeProps('aspect-ratio', {shape}),
+        themeProps('aspect-ratio', {shape, fit}),
```
cixzhang: Since fit is somewhat structural it wouldn't really make sense to theme. The marker will likely be more stable when applied to the direct parent.

cixzhang (changes requested):
Some feedback regarding the targeting for fit but the general strategy makes sense. Thank you for working on it!

packages/core/src/AspectRatio/AspectRatio.tsx
jiunshinn: Agreed — removed `fit` from the theming surface in 3d5587e61: `themeProps('aspect-ratio', {shape})` now reflects only `shape`, and the doc.mjs theming targets drop `fit` too. The marker moved to the child's direct parent (the fill wrapper), per the suggestion below.

packages/core/src/AspectRatio/AspectRatio.tsx
jiunshinn: Done in 3d5587e61 — the fill wrapper (the child's direct parent) now carries a dedicated marker attribute, so the reset.css rules target it directly instead of going through `.astryx-aspect-ratio[data-fit]` + an intermediate `div`.

jiunshinn: Used exactly that name in 3d5587e61: `data-astryx-aspect-ratio-override="<fit>"`. It carries the fit value itself (rather than a boolean marker plus a second attribute) so one attribute keeps the reset.css selectors direct-child while still distinguishing `cover` from `contain`.

packages/core/src/reset.css
jiunshinn: Fixed in 3d5587e61 — the rules are now direct child overrides on the marked wrapper itself, e.g. `:where([data-astryx-aspect-ratio-override="cover"]) > :where(img, video)`, with no `> div >` hop through internal structure and no coupling to the theme class. Still zero-specificity via `:where()`; the baseline regression tests in AspectRatio.test.tsx were updated to pin the new selectors.

cixzhang (approved):
Thanks looks good!

Issue #2795, opened 2026-06-12
[Bug] Playground preview ignores @container/@media (and conditional StyleX) in user code


rubyycheung:
## Summary

In the **Playground** (`/playground`, `xds-sandbox.vercel.app/playground`), templates/snippets that rely on StyleX `@container` / `@media` queries (or any conditional/at-rule StyleX value) do **not** render their responsive behavior. The styles silently collapse to their `default` value.

Observed with the **Mixed Gallery** template:
- **Desktop:** the asymmetric "wide hero spanning 2 columns" layout is lost — it renders as a plain 3-up grid.
- **Mobile (phone viewport):** the grid does **not** collapse to a single column — it stays multi-column with tiny tiles.

The same template renders correctly on the **deployed sandbox/docsite** (which use build-time–compiled StyleX). This is Playground-only.

Also affects `ide` (uses `@media` for `hideOnMobile` / `hideSideNav`), and any future template using container/media queries.

## Root cause

The Playground is an **editable, runtime-only** environment: it transpiles user code in-browser with the TypeScript compiler (`apps/docsite/src/app/(preview)/playground-preview/runner.ts`) — there is **no `@stylexjs/babel-plugin`** in that path.

StyleX requires its compiler at runtime:

```
stylex.create(...) → throws: "Unexpected 'stylex.create' call at runtime.
Styles must be compiled by '@stylexjs/babel-plugin'."
```

So the Playground substitutes a **`stylexMock`** (generated in `apps/docsite/scripts/generate-scope.mjs`). The mock flattens `stylex.create`/`props` into an inline `style={}` object. Inline styles can't express at-rules/pseudos, so the mock deliberately:
- skips any key matching `/^(:|::|@|--)/` (drops `@container`, `@media`, `:hover`, CSS vars), and
- collapses responsive objects (`{default: X, '@container ...': Y}`) to just `default`.

```js
// apps/docsite/scripts/generate-scope.mjs
const STYLEX_SKIP_KEY = /^(:|::|@|--)/;
// "Pseudo/media entries can't be inlined: responsive objects collapse to `default`."
```

## Why a fix is non-trivial

A real fix means reproducing StyleX's compile output at runtime (atomic class hashing + CSS injection + `$$css`-tagged objects). The hard blocker is that templates pass user styles as **`xstyle` into compiled `@xds/core` components**, e.g.:

```js
// packages/core/dist/AspectRatio/XDSAspectRatio.js
[...]

PR #3883, opened 2026-07-12
feat(AspectRatio): responsive ratio — emit ratio as an overridable class-level declaration

### LINKED ISSUE #2798: XDSAspectRatio: support a responsive `ratio` (forces double-element workarounds today)
## Problem

`XDSAspectRatio` applies its aspect ratio as an **inline style**:

```tsx
// packages/core/src/AspectRatio/XDSAspectRatio.tsx
<div {...mergeProps(..., {...style, aspectRatio: ratio})}>
```

Inline styles win over StyleX class rules, so the ratio **can't be overridden responsively** via an `@container` / `@media` query. There's no way to say "3:1 on desktop, 3:2 when stacked" with a single `XDSAspectRatio`.

## Impact (real workaround in a template)

The **mixed-gallery** page template (`packages/cli/templates/pages/mixed-gallery/page.tsx`) wants a hero that's `3:1` spanning two columns on desktop (so its height matches the `3:2` tile beside it) but `3:2` once the grid collapses to one column (so every stacked tile is uniform).

Because the ratio can't be responsive, it renders the **same image twice** — a `3:1` `heroWide` and a `3:2` `heroNarrow` — and toggles them with `@container display:none`:

```tsx
heroWide:   { gridColumn: 'span 2', display: { default: 'block', '@container … (max-width: 720px)': 'none' } },
heroNarrow: { gridColumn: '1 / -1', display: { default: 'none',  '@container … (max-width: 720px)': 'block' } },
```

Downsides:
- Duplicate `` in the markup (extra raw element → worse on the template grading rubric).
- The duplicate is only hidden where StyleX compiles; it surfaced as a visible "extra image" in the Playground preview (tracked separately as #2795).

## Proposal

Let `XDSAspectRatio` accept a **responsive `ratio`** so the workaround isn't needed. Options:

1. Accept a conditional/responsive value for `ratio` (e.g. `ratio={{default: 3/1, '@container … (max-width: 720px)': 3/2}}`) and emit it as a StyleX rule (or a CSS custom property) rather than a hard inline style.
2. Or expose the ratio via a CSS variable (e.g. `--xds-aspect-ratio`) that consumers can override with `@container`/`@media`.

Either would let mixed-gallery (and similar responsive media layouts) use a **single** element.

## Related
- #2582 — no `XDSImage` primitive (the raw `` + fill/radius CSS these templates also need).
- #2795 — the Playground duplicate-hero symptom this workaround caused.

jiunshinn:
## Summary

`AspectRatio` applied its ratio as an inline style:

```tsx
// packages/core/src/AspectRatio/AspectRatio.tsx (before)
{...style, aspectRatio: ratio}
```

Inline styles beat every class rule, so the ratio could not be overridden responsively — there was no way to say "3:1 on desktop, 3:2 when stacked" with a single element, which is exactly what the mixed-gallery template needed and previously worked around by rendering the hero image twice and toggling the copies with `@container display:none` ([#2795](https://github.com/facebook/astryx/issues/2795) was the visible fallout).

This PR emits the ratio as a **class-level declaration** instead, through a StyleX dynamic style:

```tsx
const dynamicStyles = stylex.create({
  ratio: (ratio: number) => ({aspectRatio: ratio}),
});
```

```css
/* compiled output (verified in dist/astryx.css) */
@layer astryx-base { .x… { aspect-ratio: var(--x-aspectRatio); } }
/* inline on the element: --x-aspectRatio: <ratio> */
```

Only the *value* rides an inline custom property. The `aspect-ratio` declaration itself is an ordinary class rule inside the `astryx-base` cascade layer, and that is what makes it reachable. Three override paths fall out of it, one per consumer type:

```tsx
// StyleX consumers — an xstyle rule wins over the component's class
  hero: {aspectRatio: {default: 3, '@container gallery (max-width: 720px)': '3 / 2'}},
});
<AspectRatio ratio={3} xstyle={styles.hero}>…</AspectRatio>
```

```css
/* Plain-CSS / Tailwind consumers — an unlayered rule beats the layer,
   at any specificity. No custom property to learn. */
@container gallery (max-width: 720px) {
  .gallery-hero { aspect-ratio: 3 / 2; }
}
```

```tsx
// And a consumer `style` prop now passes through and wins, instead of
// being silently clobbered by the ratio prop.
<AspectRatio ratio={3} style={{aspectRatio: '3 / 2'}}>…</AspectRatio>
```

The `ratio` prop's public type is untouched (`number`, still required) — no API change, no call-site migration.

Fixes [#2798](https://github.com/facebook/astryx/issues/2798)

### Why no `--astryx-aspect-ratio` custom property

An earlier revision of this PR shipped a dedicated `--astryx-aspect-ratio` hook (issue option 2), on the theory that plain-CSS consumers — CLI templates render without a StyleX compiler, [#3102](https://github.com/facebook/astryx/issues/3102) — would otherwise have no way in. Review established they do: the compiled declaration sits in `@layer astryx-base`, and *any* unlayered consumer rule beats a layered one regardless of specificity. The variable was a second public API for something the cascade already gives away, so it was dropped. Plain CSS just writes `aspect-ratio`.

## Responsive-pattern evidence

Checked how core handles responsive prop values before choosing a mechanism:

- **No component accepts a breakpoint-map prop value** (`{default, '@media …'}` objects). `Grid`'s `columns` object form is min-width auto-fill (intrinsic responsiveness), `AppShell` uses a named-breakpoint enum + `useMediaQuery` for behavior (not styling), `FormLayout` bakes a fixed `@media` into static styles. So issue option 1 (a responsive `ratio` object) has no precedent, and StyleX dynamic styles can't compile runtime-provided at-rule keys anyway.
- **The exact mechanism used here is already established, comment and all**: `Grid`'s dynamic track values, `packages/core/src/Grid/Grid.tsx` — "Dynamic track values compile to CSS variables + a class-level declaration … instead of a raw inline style, so consumer `xstyle` overrides — including ones inside @media queries — can still win."

**Design-scope note (settled in review):** this ships the minimal mechanism — no API change, all three override paths work today. A first-class responsive `ratio` prop value could be layered on later without conflicting with it.

## Behavior deltas

| Before | After |
|---|---|
| `aspect-ratio` forced inline; unoverridable by any CSS | Class-level declaration in `astryx-base`; `xstyle`, unlayered consumer CSS, or an inline `style` can override |
| Consumer `style={{aspectRatio}}` was silently clobbered by the `ratio` prop (`{...style, aspectRatio: ratio}`) | Consumer `style` passes through untouched and wins (covered by a test) |
| mixed-gallery hero fixed at 3:1 even stacked | Hero drops to 3:2 when the grid stacks — one CSS line, single element |

Back-compat: for every consumer who did not fight the inline style, rendering is identical — same computed `aspect-ratio`, just declared via class + variable. The only observable change is the second row above, which was previously a silent no-op footgun.

## Changes

- `packages/core/src/AspectRatio/AspectRatio.tsx` — `dynamicStyles.ratio()` (Grid-style comment), drop the inline merge, JSDoc on `ratio` + component docs
- `packages/core/src/AspectRatio/AspectRatio.test.tsx` — assertions move from `element.style.aspectRatio` to the emitted `--x-aspectRatio` variable (same helper pattern as `Grid.test.tsx`); new tests: no inline declaration, consumer `style` passthrough
- `packages/core/src/AspectRatio/AspectRatio.doc.mjs` — `ratio` prop description (EN/ZH/dense) + responsive best-practice entry
- `packages/cli/assets/templates/pages/mixed-gallery/page.tsx` — the motivating case: hero switches 3:1 → 3:2 when stacked, via one `aspect-ratio` line in the existing unlayered `@container` block
- `.changeset/aspectratio-responsive-ratio.md` — `@astryxdesign/core` + `@astryxdesign/cli` patch

## Test plan

`pnpm exec vitest run packages/core/src/AspectRatio/` — 28/28 pass · `pnpm -F @astryxdesign/core typecheck` — clean · `build` green, `dist/astryx.css` contains `aspect-ratio:var(--x-aspectRatio)` and no inline emission · `eslint` on changed files clean · `check:sync`, `check:changesets`, `check:package-boundaries` clean.

Measured in real Chromium at this head, either side of a 500px container query (`ar` = computed `aspect-ratio`, `measured` = width/height of the rendered box):

| case | @900px | @420px |
|---|---|---|
| plain `ratio={3}` | `3 / 1`, measured 3.000, no inline decl | `3 / 1`, measured 3.000 |
| `xstyle` with a `default` branch | `3 / 1`, measured 3.000 | `3 / 2`, measured 1.500 |
| unlayered consumer CSS | `3 / 1`, measured 3.000 | `3 / 2`, measured 1.500 |
| consumer `style` prop | `3 / 2`, measured 1.500 | `3 / 2`, measured 1.500 |

All 13 shipped `AspectRatio` stories (25 ratio boxes) render byte-identical to `main` — same computed ratios, same pixel dimensions.

---

**Refreshed by maintainers:** this branch had aged 42 days / 341 commits behind `main`, so `main` was merged in (not rebased — the author's history and authorship stand) and this description was rewritten to match the diff, which no longer ships the `--astryx-aspect-ratio` custom property the original text described. [#4984](https://github.com/facebook/astryx/pull/4984) had meanwhile landed a test pinning the old inline-clobber behaviour, which git merged silently; it now asserts the new precedence. The staleness was ours, not the author's.


cixzhang:
-- whoops one sec I'm reading through too many PRs too quickly --

cixzhang (commented):
Thanks! This looks good. We probably need to do a sweep of converting inline styles into dynamic styles in our codebase generally. Since this is stacked it includes the fit changes and we do squash merges so waiting on that to merge first

packages/core/src/AspectRatio/AspectRatio.tsx
```diff
@@ -130,6 +148,21 @@ const styles = stylex.create({
...
+// overrides — `xstyle` rules, including ones inside @media/@container
+// queries — can still win. A raw inline `aspect-ratio` would beat any class
+// (same rationale as Grid's dynamic track values).
+// The `--astryx-aspect-ratio` custom property (namespace: naming.ts cssVar)
+// is the override hook for non-StyleX consumers: setting it on the element
+// from any rule — e.g. inside a @container query — wins over the `ratio`
+// prop fallback regardless of specificity or cascade layer.
+const dynamicStyles = stylex.create({
+  ratio: (ratio: number) => ({
+    aspectRatio: `var(--astryx-aspect-ratio, ${ratio})`,
```
cixzhang: We're trying to avoid too many custom vars that are component specific features. Is there a reason we want the astryx-aspect-ratio?

jiunshinn: The var is there for one specific gap: a class-level declaration alone is only overridable by consumers running the StyleX compiler (via `xstyle`). The var inherits through the cascade, so template/plain-CSS consumers can set it from a parent scope under a media or container query — that's how the mixed-gallery hero does its 3:1 → 3:2 switch, and it was the part of #2798 that had no other path without a first-class responsive ratio API (templates render without the compiler, see #3102). Grid uses the same var-indirection internally (Grid.tsx:156), though not as a public surface.

That said, I understand keeping component-specific vars to a minimum. Happy to go either way: (a) drop the var and keep only the overridable class declaration — compiler users still get responsive ratio via `xstyle`, and the hero example switches to that; or (b) treat this as an argument for a first-class responsive `ratio` value later and drop the var now. Just let me know which you'd prefer. I've also rebased this on main now that #3882 has merged.

jiunshinn:
Hi @cixzhang , Is there anything else I can do help for this PR get merged?

packages/cli/assets/templates/pages/mixed-gallery/page.tsx
```diff
@@ -134,9 +137,12 @@ export default function MixedGalleryTemplate() {
...
+                3 columns → 1 column at ≤720px, where the hero drops to 3:2 so
+                the stacked tiles are uniform. */}
             <div className="mixed-gallery-grid">
-              {/* Hero — spans 2 columns; 3:1 keeps it level with the sidebar */}
+              {/* Hero — spans 2 columns; 3:1 keeps it level with the sidebar.
+                  Stacked, --astryx-aspect-ratio switches it to 3:2 (see
+                  GALLERY_CSS) — one element, no duplicate hero markup. */}
               <GalleryCard
                 image={IMAGES[0]}
                 ratio={3 / 1}
```
cixzhang, line 151: After this update should this have worked without the extra custom var using just stylex's dynamic vars?

packages/cli/assets/templates/pages/mixed-gallery/page.tsx
```diff
@@ -52,6 +54,7 @@ const GALLERY_CSS = `
   }
   .mixed-gallery-hero {
     grid-column: 1 / -1;
+    --astryx-aspect-ratio: 3 / 2;
```
cixzhang: Hmm it does seem odd to set this when we wanted 3 / 1. Is it intended to be a default value?

packages/core/src/AspectRatio/AspectRatio.tsx
cixzhang: Thanks for the explanation! Looking at the gallery example I'm not sure it actually needs the var. The class we apply attempts to set the ratio to 3 / 1 but then sets the var to 3 / 2. Seems like it's intended to be a default value but in that case perhaps it would be better to set explicit default in the component over a new custom var. that said maybe there's a use case I'm missing!

cixzhang (changes requested):

cixzhang:
Thanks for the detailed explanation — and sorry for leaving this thread sitting since July.

First, a correction on my own comment above: you're right that the `3 / 2` is **inside** the `@container` block, not a default value. The `3 / 1` comes from `ratio={3 / 1}` and the var only takes effect below 720px, once the hero is full-width. My misread — that part is working as intended.

That said, I don't think we want to expose responsive CSS through a component-specific custom property. Components are built with StyleX; templates author regular CSS. Those are the two supported paths, and a public `--astryx-aspect-ratio` adds a third that we'd then be expected to provide for every component with a dynamic value.

The good news is that the class-level change alone already gets you there, without the var. Library rules compile into `@layer astryx-base`, and the template's `GALLERY_CSS` is a plain unlayered `<style>` tag — and unlayered CSS beats every layered rule regardless of specificity. So once the ratio stops being an inline style, the template can simply write:

```css
@container gallery (max-width: 720px) {
  .mixed-gallery-hero { aspect-ratio: 3 / 2; }
}
```

That's essentially option (a) from your comment: drop the var, keep the overridable class-level declaration. StyleX consumers get responsive ratio via `xstyle` (including under `@media`/`@container`), plain-CSS consumers get it via unlayered rules. Both audiences keep the capability and we add no new public surface.

To be clear about what I'd like to keep: **the inline-style → class-level declaration change is the valuable part of this PR** and I want to land it. The inline style beating every class rule is the actual bug. It's just the var indirection I'd like removed.

One thing worth confirming in a browser before we commit to it: that a plain unlayered `aspect-ratio` override genuinely wins over the StyleX class in practice. Layer interactions are easy to reason about correctly on paper and still surprise you. If it doesn't behave as described, say so and let's reopen the question rather than shipping the workaround.

jiunshinn:
Done in 4404b67acf — this is option (a) exactly. The var is gone: the ratio is now a plain class-level `aspect-ratio: var(--x-aspectRatio)` declaration (private compiled variable, per-element value set inline, no public fallback chain), and the mixed-gallery hero override is the rule you wrote:

```css
@container gallery (max-width: 720px) {
  .mixed-gallery-hero { aspect-ratio: 3 / 2; }
}
```

Docs and tests now describe the two supported paths only — `xstyle` for StyleX consumers, unlayered consumer CSS for plain-CSS/Tailwind consumers. The `--astryx-aspect-ratio` fallback-chain test is removed; the no-inline-declaration and consumer-`style` passthrough tests stay (26/26 pass; typecheck/lint/build green).

**Browser check you asked for** — I rebuilt `dist/astryx.css` from this branch (all rules inside `@layer astryx-base`) and loaded it in Chrome alongside the template's unlayered `<style>` block, with hero markup mirroring the compiled output (`.x10y9f9r { aspect-ratio: var(--x-aspectRatio) }` + inline `--x-aspectRatio: 3`):

| Case | Computed `aspect-ratio` | Measured w/h |
|---|---|---|
| 900px container (query inactive) | `3 / 1` | 596 / 199 ≈ 2.995 |
| 500px container, unlayered override active | `3 / 2` | 500 / 333 ≈ 1.5015 |
| 500px control, no `.mixed-gallery-hero` class | `3 / 1` | 500 / 167 ≈ 2.994 |

So the unlayered rule genuinely wins in practice — including over the compiled rule's `:not(#\#)`×3 specificity boost, which is the part that could plausibly have surprised us if layer behavior didn't match the spec. The control case confirms it's the override rule doing the work, not the container query interfering with the layered declaration.

This also answers the inline question on the template: it renders without a StyleX compiler (#3102), which was the var's original motivation — the unlayered path now covers that audience with no new public surface.

packages/core/src/AspectRatio/AspectRatio.doc.mjs
```diff
@@ -14,6 +14,7 @@ export const docs = {
       {guidance: true, description: 'Express the ratio as a fraction like `16/9` or `4/3` for readability.'},
       {guidance: true, description: 'Use for media that needs consistent proportions across screen sizes.'},
       {guidance: true, description: 'Use `fit="cover"` for images and video so the component sizes the child; the child should not repeat `width`/`height`/`objectFit` styles.'},
+      {guidance: true, description: 'For a breakpoint-dependent ratio, override the ratio responsively: pass an `aspectRatio` rule via `xstyle` (StyleX), or override `aspect-ratio` from your own unlayered CSS under a `@media`/`@container` rule — component styles live in the `astryx-base` cascade layer, so unlayered consumer CSS wins.'},
```
cixzhang: Might be worth saying the `xstyle` rule needs a `default` branch — conditional-only drops the ratio.

cixzhang (approved):
Thanks, this is option (a) exactly — I re-checked the unlayered override in Chrome and it holds. Needs a merge with main. Description still describes the `--astryx-aspect-ratio` var you removed.

[Reviewed by Robohands]

cixzhang:
Merged — we brought `main` in ourselves and refreshed the description, because the branch had aged past what that description said rather than through any fault of yours. Thank you, @jiunshinn, for making the ratio overridable at class level: that single move is what lets a responsive override actually reach it, and the mixed-gallery hero now goes 3:1 → 3:2 on one element instead of two. I'm sorry this sat for six weeks after you'd built exactly what was asked for — that wait was ours, not yours.

[Reviewed by Robohands]

PR #3680, opened 2026-07-08
docs(core): document full themeProps() selector surface in theming targets

Arman-Luthra:
## Summary

#3652 fixed five components whose `themeProps()` calls pass visual props that the `theming.targets` entry did not document. This PR completes that sweep: I compared every `themeProps()` call in `packages/core/src` against every `theming.targets` entry and found 19 remaining gaps across 16 components. Without the entry, the CLI theming table shows no `data-*` selector for the prop, so theme authors do not see the selector surface they can target.

## Changes

Six targets were missing entirely:

| Target | Props documented |
|---|---|
| `astryx-citation` | `variant` |
| `astryx-toggle-button` | `isPressed` |
| `astryx-banner-content` | `container`, `status` |
| `astryx-overlay` / `astryx-overlay-scrim` | `position` (scrim) |
| `astryx-nav-heading-menu` | `size` |
| `astryx-nav-heading-menu-item` | `size` |

Thirteen existing entries were missing props their `themeProps()` call passes:

| Target | Added |
|---|---|
| `astryx-chat-message` | `density` |
| `astryx-checkbox` | `checked`, `disabled`, `size` |
| `astryx-code` | `color` |
| `astryx-codeblock` | `container` |
| `astryx-dropdown-menu-item` | `size` |
| `astryx-segmented-control-item` | `selected`, `disabled`, `size` |
| `astryx-selectable-card` | `variant` |
| `astryx-side-nav-item` | `selected`, `size` |
| `astryx-timestamp` | `format` |
| `astryx-tokenizer` | `status` |
| `astryx-top-nav-item` | `selected` |
| `astryx-tree-list-item` | `density` |
| `astryx-typeahead` | `size` |

Boolean interaction props go under `states` and enum visual props under `visualProps`, matching the existing convention in each file. Files that duplicate targets in the translated docs object got both copies updated.

## Validation

- Re-ran the sweep after the change: 0 gaps across all 104 `themeProps()` keys
- All 16 edited `.doc.mjs` files import cleanly under node
- `check:changesets`, `check:sync` pass; ESLint reports no new problems on changed files
- `pnpm vitest run component-format`: 4 tests pass
- `apps/docsite/scripts/generate-data.mjs` exits 0
- `astryx component Citation` now renders a Theming table with `data-variant` and its resolved values

PR #3887, opened 2026-07-12
fix(core): document full themeProps() selector surface and guard it (#3741)

### LINKED ISSUE #3741: CI: validate theming.targets against actual themeProps() call sites
## Problem

Component docs carry a `theming.targets` array (`ThemingTarget[]` in `packages/core/src/docs-types.ts`). Each target documents the stable `astryx-*` class a component renders plus the props reflected onto it — `visualProps` (enum/visual props) and `states` (booleans/runtime state) — which are exactly the props passed to `themeProps('<name>', {…})` as the second argument in the component's `.tsx`.

Today this array is **hand-authored** in each `*.doc.mjs`. It's a manual mirror of the real `themeProps()` call sites, and **nothing validates that the two stay in sync** — there's no generator and no CI gate. (Confirmed: no reference to `themeProps` / `theming.targets` / `ThemingTarget` under `scripts/` or `.github/`.)

Because it's hand-maintained, it has already drifted twice and needed manual sweep PRs to repair:

- #3652 — first sweep (documented `visualProps` that were missing on theming targets)
- #3680 — second sweep (closed 19 more `themeProps()` → `theming.targets` gaps across 16 components)

Each drift is silent until someone notices and re-sweeps by hand. A lightweight CI check would catch it at PR time and stop the recurring manual repairs.

## Proposed approach

Add a check script under `scripts/` (e.g. `check-theming-targets.mjs`) that:

1. Parses each component's source for `themeProps('<name>', { … })` call sites and collects, per component, the set of `astryx-<name>` class names and the prop keys passed in the second argument.
2. Parses each `*.doc.mjs` `theming.targets` and collects the documented `className` + `visualProps` + `states` sets.
3. Diffs the two and fails on mismatch, printing the offending component, class name, and the missing/extra keys.

Wire it into CI as a `--check` step alongside the existing sync gates in `.github/workflows/ci.yml` (`scripts/sync-exports.js --check`, `scripts/generate-token-docs.mjs --check`) — same established pattern, so it reads naturally next to them.

If we want to go further than validation, the same parser could power a `--write` mode that regenerates `theming.targets` from source, turning the hand-authored mirror into generated data. Validation-only is the smaller first step and already removes the manual-sweep burden.

## Edge cases to handle

- **Props not passed to `themeProps`.** The documented target intentionally lists props that are *semantically* part of the selector surface but not literally in the `themeProps()` call. Example: `Timestamp` calls `themeProps('timestamp', {format})`, but `Timestamp.doc.mjs` documents `visualProps: ['type', 'color', 'format']` — `type` and `color` are forwarded to a child `<Text>`. A naive key-diff would false-positive here. The check needs a policy for this (either treat `themeProps` keys as a subset that must all appear in the docs, not an exact equality, or allow an explicit annotation for extra props).
- **Dynamic / spread second arg.** Real call sites today all use a string-literal first arg and an object literal second arg, but the check should skip (or warn on) non-literal args rather than crash.
- **Multiple `themeProps()` calls per component**, and multiple distinct targets (e.g. `ToggleButton` documents both `astryx-toggle-button-group` and `astryx-toggle-button`). Keys should be unioned per class name.
- **Component-name → class-name transform.** The first arg is the bare name (`'toggle-button'`); the rendered/documented class is `astryx-toggle-button` (see `buildClassName` in `packages/core/src/utils/themeProps.ts`). The check must apply the same `astryx-` prefixing.
- **Translated `docsZh` duplicates.** Many `*.doc.mjs` export a second `docsZh` object that carries its **own** `theming.targets` block (e.g. `AppShell.doc.mjs`). Both blocks must stay in lockstep with source, so the check has to validate every `theming.targets` in the file, not just the primary `docs` export.

## Scope

`packages/core` — 87 `*.doc.mjs` files currently define `theming.targets`.

AKnassa:
Closes #3741.

## The gap

`theming.targets` is the documented CSS surface of a component: the stable `astryx-*` classes it renders, and the visual props it reflects as data attributes. It is hand-authored, while the truth lives in the `themeProps()` calls in the source — and nothing kept the two in agreement. It has drifted twice already (#3652, #3680, both manual sweeps), and it was drifted again on `main`.

Drift is not cosmetic. `targets[]` is what theme authors and codegen read to learn which selectors exist, so an undocumented class is an unthemeable element.

## What was actually broken

Eight components were under-documenting their surface:

| component | missing from `targets[]` | block |
|---|---|---|
| `Table` | `astryx-table`, `-scroll-wrapper`, `-header`, `-body`, `-footer` — 4 of 9 classes documented | `docs` |
| `AvatarGroup` | `astryx-avatar-group-overflow` | `docs` |
| `CommandPalette` | `astryx-command-palette-empty` | `docs` |
| `InputGroup` | `astryx-input-group-text` | `docs` |
| `Outline` | `astryx-outline-indicator` | `docs` **and** `docsZh` |
| `ProgressBar` | `astryx-progressbar-track` | `docs` **and** `docsZh` |
| `AspectRatio` | `visualProps: ['shape']` | `docsZh` |
| `Card` | `visualProps: ['variant']` | `docsZh` |
| `EmptyState` | `visualProps: ['variant']` | `docsZh` |

The `docsZh` column is the failure mode the issue predicted: the Chinese block silently falls out of lockstep with the English one.

## The guard

`packages/core/src/theme/themingTargets.test.ts` parses every `themeProps()` call site with the TypeScript AST and fails when a rendered class — or a prop key passed to it — is missing from `targets[]`, in **both** the `docs` and `docsZh` blocks.

It sits beside `derivedVarRegistry.test.ts`, which already machine-validates the *sibling* fields of the same `theming` block (`vars`, `derived`). `targets` was the one field with no check, which is why it was the one field that drifted. Landing it as a test rather than a new `scripts/*.mjs` + `ci.yml` step means the existing test job already runs it — no CI surgery.

**The AST is load-bearing, not ceremony.** The call sites use every object form — shorthand (`{variant}`), renamed (`{variant: fillVariant}`), and multi-line. A regex reading identifiers after `:` records `fillVariant`, a prop that does not exist. `ProgressBar.tsx` alone contains all three forms.

**Policy is subset, not equality.** Docs may list *more* than the source passes, because components forward props they don't themselves reflect — `Timestamp` passes `{format}` but documents `type`/`color`/`format`, and that is intentional. A props bag that can't be read statically (spread, computed key, non-literal) is marked opaque rather than guessed at.

## Test plan

- [x] The guard **fails on `main`** — 11 assertions across the 8 components above. Red before green; the drift is real, not hypothetical.
- [x] The extractor's own 10 unit tests are mutation-checked: making renamed props record the *value* instead of the key (the exact regex trap) fails precisely the two tests written to catch it.
- [x] `pnpm exec vitest run` — **337 files, 6567 tests, all passing.**
- [x] `pnpm check:repo` — all five gates pass (sync, package-boundaries, changesets, demo-media, executable-bits).
- [x] `pnpm -F @astryxdesign/core typecheck` — clean.
- [x] The guard is a `*.test.ts`, which `build:esm` ignores — the TypeScript compiler import never reaches the published bundle. No new runtime dependency.

I'm happy to split the doc sweep out from the guard if you'd rather review them separately — though the gate can't land green without the sweep, so they'd have to merge in that order.

### DIFF
--- a/.changeset/theming-targets-guard.md
+++ b/.changeset/theming-targets-guard.md
@@ -1,0 +1,13 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Document the full themeProps() selector surface and guard it against drift (#3741)
+
+`theming.targets` is the documented CSS surface of a component — the stable `astryx-*` classes it renders and the visual props it reflects as data attributes. It was hand-authored while the truth lived in `themeProps()` calls in the source, and nothing kept the two in agreement, so it drifted twice (#3652, #3680) and was drifted again.
+
+Eight components were under-documenting their surface: `Table` documented 4 of the 9 classes it renders (`astryx-table`, `-scroll-wrapper`, `-header`, `-body`, `-footer` were all missing), and `AvatarGroup`, `CommandPalette`, `InputGroup`, `Outline` and `ProgressBar` each hid one. `AspectRatio`, `Card` and `EmptyState` documented their visual props in `docs` but not in `docsZh`. An undocumented class is an unthemeable element: theme authors and codegen read `targets[]` to learn which selectors exist.
+
+`themingTargets.test.ts` now parses every `themeProps()` call site with the TypeScript AST and fails when a rendered class or a passed prop key is missing from `targets[]` — in both the `docs` and `docsZh` blocks. It sits next to `derivedVarRegistry.test.ts`, which already validates the sibling `vars`/`derived` fields of the same `theming` block; `targets` was the one field with no machine check. The policy is subset, not equality: docs may list more than the source passes, since components forward props they don't themselves reflect.
+
+@AKnassa
--- a/packages/core/src/AspectRatio/AspectRatio.doc.mjs
+++ b/packages/core/src/AspectRatio/AspectRatio.doc.mjs
@@ -76,7 +76,7 @@ export const docsZh = {
   ],
   theming: {
     targets: [
-      {className: 'astryx-aspect-ratio'},
+      {className: 'astryx-aspect-ratio', visualProps: ['shape']},
     ],
   },
 };
--- a/packages/core/src/AvatarGroup/AvatarGroup.doc.mjs
+++ b/packages/core/src/AvatarGroup/AvatarGroup.doc.mjs
@@ -24,6 +24,7 @@ export const docs = {
   theming: {
     targets: [
       {className: 'astryx-avatar-group', visualProps: ['size']},
+      {className: 'astryx-avatar-group-overflow'},
     ],
   },
   description: 'Stacked avatar display with overlapping layout and optional overflow indicator. Children are Avatar elements.',
--- a/packages/core/src/Card/Card.doc.mjs
+++ b/packages/core/src/Card/Card.doc.mjs
@@ -131,7 +131,7 @@ export const docsZh = {
   theming: {
     container: true,
     targets: [
-      {className: 'astryx-card'},
+      {className: 'astryx-card', visualProps: ['variant']},
     ],
     vars: [
       {name: '--_card-radius', description: 'Border radius of the card', default: 'var(--radius-container)', private: true},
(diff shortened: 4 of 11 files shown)


cixzhang (approved):
Nice this is a good guard

PR #3607, opened 2026-07-05
fix(deps): resolve js-yaml CVE-2025-27714 via pnpm overrides

cixzhang:
Resolves Dependabot alerts #102 and #103 for js-yaml quadratic-complexity DoS (CVE-2025-27714).

**Changes:**
- `read-yaml-file>js-yaml: ^3.15.0` -- keeps the 3.x consumer on a patched 3.x release (read-yaml-file depends on the 3.x `safeLoad` API)
- `@changesets/parse>js-yaml: >=4.2.0` -- bumps the 4.x consumer past the vulnerable range

Both are transitive dependencies of `@changesets/cli`. Verified with `pnpm build` and `pnpm test --run` (306 files, 5582 tests passing).

**Not addressed here:**
- Vitest 2.x -> 3.x (alerts #80, #81) -- major version bump, tracked in #2974

Night Watch -- Dependencies

PR #3889, opened 2026-07-12
fix(deps): lift read-yaml-file to 2.x to clear js-yaml CVE-2026-53550 (#3221)

### LINKED ISSUE #3221: deps: js-yaml 3.x -> 4.x major bump needed (CVE-2026-53550)
Dependabot alert 87 flags a DoS vulnerability in js-yaml (CVE-2026-53550).

**Current version:** 3.14.2 (transitive via `@changesets/cli` -> `@manypkg/get-packages` -> `read-yaml-file`)
**Patched version:** 4.2.0

This is a major version bump (3.x to 4.x). The dependency chain is:
```
@changesets/cli -> @manypkg/get-packages -> read-yaml-file -> js-yaml@3.x
```

A pnpm override for `js-yaml` to `^4.2.0` may break `read-yaml-file` since js-yaml 4.x dropped several APIs (e.g. `safeLoad` removed, replaced by `load`). Needs verification or an upstream fix in `@changesets/cli`.

**Advisory:** https://github.com/nodeca/js-yaml/security/advisories/GHSA-h67p-54hq-rp68

**Risk:** Low for this project. The vulnerable code path (merge key aliasing) would only trigger if processing untrusted YAML. Changeset YAML files are authored locally.

AKnassa:
Closes #3221.

## The problem

`@changesets/cli` → `@manypkg/get-packages` → `read-yaml-file` still resolves **js-yaml 3.15.0**, below the **4.2.0** that patches CVE-2026-53550.

The override added in #3607 fixed a *different* js-yaml advisory and deliberately held this chain on the 3.x line (`read-yaml-file>js-yaml: ^3.15.0`). That was the right call at the time: `read-yaml-file` 1.x calls `safeLoad`, which js-yaml 4.x renamed to `load`, so overriding js-yaml alone throws `TypeError: yaml.safeLoad is not a function`. The pin is exactly what this issue needs removed.

## The fix: lift the package, don't patch it

`read-yaml-file` **2.x** calls `load` and depends on `js-yaml ^4` — so the pin becomes unnecessary rather than needing a workaround.

This deliberately avoids the `pnpm patch` route the issue proposed. That would introduce `patchedDependencies` + a `patches/` dir the repo doesn't use, affecting every install and the release pipeline — a toolchain policy call. Overriding is the mechanism this repo already uses, and #3607's own rationale ("keep each consumer on an API-compatible major line") reads as a preference against patching.

Two details worth flagging for review:

- **`read-yaml-file` 2.x is a drop-in.** Its only consumer is `@manypkg/get-packages`, and 2.x keeps the same CJS surface — default export, `.default`, and `.sync`.
- **The js-yaml floor is `^4.2.0`, not `>=4.2.0`.** The looser range resolves to js-yaml **5.x**, outside the `^4` range `read-yaml-file` 2.x declares. I hit this while testing.

## Dependency delta

No `package.json` changes, no new direct dependency. Four packages leave, three arrive — the tree gets one smaller:

| | package |
|---|---|
| **removed** | `js-yaml@3.15.0` (the vulnerable one), `read-yaml-file@1.1.0`, `argparse@1.0.10`, `sprintf-js@1.0.3` |
| **added** | `js-yaml@4.3.0` (patched), `read-yaml-file@2.1.0`, `strip-bom@4.0.0` |

The chain is dev-side (release tooling) — nothing here ships in `@astryxdesign/core`'s runtime dependencies.

## Test plan

- [x] **No js-yaml < 4.2.0 anywhere in the tree.** `node_modules/js-yaml` is now `4.3.0` (was `3.15.0`); `@changesets/parse`'s nested copy stays `5.2.0`. Both above the patch floor.
- [x] **`read-yaml-file` works at runtime** — `.sync()` parses `pnpm-workspace.yaml` correctly, with js-yaml 4.3.0 underneath. `typeof default === 'function'`, `typeof .sync === 'function'`.
- [x] **changesets works end-to-end** — `changeset status` exits 0 and correctly lists the packages to be bumped; `changeset version --snapshot` completes. This is the code path that actually consumes `read-yaml-file`, so it's the check that matters.
- [x] `pnpm exec vitest run` — **336 files, 6298 tests, all passing.**

No changeset: this changes no published package's code, matching #3607, which shipped none for the same class of change.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -23,7 +23,8 @@ overrides:
   esbuild: '>=0.28.1'
   '@babel/core': ^7.29.6
   hono: ^4.12.25
-  read-yaml-file>js-yaml: ^3.15.0
+  read-yaml-file: ^2.1.0
+  read-yaml-file>js-yaml: ^4.2.0
   '@changesets/parse>js-yaml': '>=4.2.0'
 
 importers:
@@ -3453,9 +3454,6 @@ packages:
     resolution: {integrity: sha512-KMReFUr0B4t+D+OBkjR3KYqvocp2XaSzO55UcB6mgQMd3KbcE+mWTyvVV7D/zsdEbNnV6acZUutkiHQXvTr1Rw==}
     engines: {node: '>= 8'}
 
-  argparse@1.0.10:
-    resolution: {integrity: sha512-o5Roy6tNG4SL/FOkCAN6RzjiakZS25RLYFrcMttJqbdd8BWrnA+fGz57iN5Pb06pvBGvl5gQ0B48dJlslXvoTg==}
-
   argparse@2.0.1:
     resolution: {integrity: sha512-8+9WqebbFzpX9OR+Wa6O29asIogeRMzcGtAINdpMHHyAg10f05aSFVBbcEqGf/PXw1EjAZ+q2/bEBg3DvurK3Q==}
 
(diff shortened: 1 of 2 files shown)


cixzhang (approved):
Legitimate security fix. `@changesets/cli` → `@manypkg/get-packages` → `read-yaml-file@1.x` pins `js-yaml@3.15.0`, which is below the 4.2.0 that patches CVE-2026-53550 (algorithmic-complexity DoS in merge-key `<<` processing). Overriding js-yaml alone breaks read-yaml-file 1.x (4.x renamed `safeLoad`→`load`), so lifting `read-yaml-file` to 2.x — which calls `load` and depends on js-yaml ^4 — is the correct remediation. Dev-toolchain/lockfile only, CI green. LGTM.

PR #3890, opened 2026-07-12
feat(SideNav): collapse to a custom width, including zero, with optional animation (#2331)

### LINKED ISSUE #2331: AppShell: support collapsing sidenav to zero width with animation
## Problem

When using `isSideNavCollapsed` on `XDSAppShell`, the sidenav collapses to a narrow icon rail rather than fully hiding. There's no option to collapse to 0 width (completely hidden), and no built-in transition animation on the width change.

## Desired behavior

1. **Collapse to zero**: When an external button controls collapse, the sidenav should be able to fully disappear (0 width) rather than leaving a rail.
2. **Animated transition**: The width change should animate smoothly (e.g., `transition: width 175ms cubic-bezier(0.24, 1, 0.4, 1)`).

## Proposed API

```tsx
<XDSAppShell
  isSideNavCollapsed={collapsed}
  onSideNavCollapsedChange={setCollapsed}
  sideNavCollapsedWidth={0}  // new: 0 = fully hidden, default = icon rail
  sideNavTransition          // new: enable animated collapse
  sideNav={...}
>
```

Or via the SideNav `collapsible` config:

```tsx
collapsible={{
  isCollapsed,
  onCollapsedChange,
  hasButton: false,
  collapsedWidth: 0,   // fully hidden
  animated: true,
}}
```

## Context

Chat interfaces often want a full-width content area with an optional session list that slides away completely. The icon rail is useful for dashboard apps but not for focused single-pane UIs.

## Workaround

CSS overrides targeting `.xds-app-shell-sidenav` with `width: 0 !important` and custom transitions via a wrapper class.

AKnassa:
Closes #2331.

## Problem

SideNav can only collapse to a 48px icon rail (`rootCollapsed: {width: --spacing-12}`), and it hard-snaps. Focused single-pane UIs (chat, editors) want the nav to slide away entirely, which the rail cannot express. Consumers are reduced to `width: 0 !important` overrides.

## API

`collapsible` gains two keys:

```tsx
const [isCollapsed, setIsCollapsed] = useState(false);
const collapsible = {isCollapsed, onCollapsedChange: setIsCollapsed};

<TopNav endContent={<SideNavCollapseButton collapsible={collapsible} />} />
<SideNav
  collapsible={{
    ...collapsible,
    hasButton: false,
    collapsedWidth: 0,   // px. defaults to the icon rail; 0 = fully hidden
    isAnimated: true,    // slide the content out and back in (collapsedWidth: 0 only)
  }}
/>
```

**Defaults are untouched.** Without `collapsedWidth` the nav still collapses to the icon rail; without `isAnimated` it still snaps.

## Notes for review

**Only `transform` animates** (review item 1). The nav's width never tweens. For `collapsedWidth: 0` the zones render inside a slide layer that `translateX`es out through the root's existing `overflow: hidden`, mirrored under RTL exactly like `MobileNav`'s drawer. The box holds its expanded width for exactly the slide via a width transition on a `step-end` curve (`transition: width var(--duration-fast) step-end`), which computes no intermediate width, then snaps shut in a single reflow; the expanded state declares no width transition, so expanding snaps the box open and the layer slides back in. No JS and no `transitionend`, so interrupts are safe: re-expanding mid-slide cancels the pending snap (the width never moved) and the transform reverses from wherever it was. The icon rail always snaps, since animating it would mean animating `width`; `isAnimated` without `collapsedWidth: 0` still snaps, and now warns once in development naming the width it got (a type-level union was ruled out: a hoisted config widens `0` to `number`). Hidden mode also reports `isCollapsed: false` through context, so items keep their expanded layout while they slide instead of morphing into the rail's icon-only form (the "inner fixed-width wrapper" from the review, done at the source).

Read from Chrome's animation engine, seeking the paused transitions: on collapse the nav carries exactly one `CSSTransition: width, 175ms, steps(1)` and the layer one `CSSTransition: transform, 175ms`; at 0/50/100/150/174.9ms the box is still 260px wide while the layer sits at 0/-173/-238/-258/-260px, and at 175ms the box is 0. Expanding at 80ms puts the box back at 260px at once with no width animation left on it, while the layer slides back. The only `0s` duration in the emitted CSS is under `prefers-reduced-motion`, so `scripts/build-css.test.mjs` passes on a fresh build.

**Focus is parked before `inert` lands** (review item 2). The first fully-hidden render leaves `inert` off; a pre-paint layout effect moves focus, if it is inside the nav, to a `SideNavCollapseButton` rendered outside the nav (or blurs it explicitly when there is none), and `inert` lands in the synchronous follow-up commit, still before paint. Outside buttons register their element in a small module-level registry on mount, together with the collapse state they share with their nav: the controlled `onCollapsedChange`, or the deprecated `handleRef`. The nav parks focus only on a toggle that shares its state, so with two navs on a page each lands on its own toggle; with no match it blurs and warns once in development. The association is written on every render rather than re-registered when the callback identity changes, because a re-registration is detached in the mutation phase and re-attached in the layout phase, and a toggle placed after its nav would be missing when the nav's layout effect looks. The outside button is the right target because `collapsedWidth: 0` needs one anyway: the built-in button hides with the nav. Verified in Chrome: focus a link inside the nav, collapse programmatically (no click, so nothing else moves focus), and focus ends on the toggle with the nav inert.

**A resizable nav keeps its wrapper while collapsed.** Dropping the `resizableContainer` on collapse moved the `nav` to a new parent, which remounted it and detached whatever had focus inside before any parking could run; the plain rail collapse of a resizable nav loses focus the same way on main today. Only the handle is conditional now, and the wrapper inherits the background so the nav's own `background-color: inherit` reaches the surface as it does without resize.

**`collapsedWidth` has to beat the resizable width.** Unchanged from round one: `useResizable` drives its size to `0` on collapse but SideNav always discarded that so the rail class could win; an explicit `collapsedWidth` takes precedence over both, pinned by a test.

**No new tokens.** `--duration-fast` / `--ease-standard`, which `SideNavCollapseButton` already animates its chevron with; reduced motion collapses both the slide and the stepped snap to `0s`.

**The issue's first proposed API is not implemented, on purpose.** It suggests `sideNavCollapsedWidth` / `sideNavTransition` on `XDSAppShell`, but the v0.0.6 codemod deliberately migrated collapse config off AppShell onto `SideNav.collapsible`, so that is the surface this builds on.

## Test plan

- [x] 41 tests in `SideNav.test.tsx` for this feature: the 8 width/inert/precedence tests, 12 for the slide (compiled StyleX rules read back from `document.styleSheets`: transform-only on the layer, `step-end` on the token duration for the collapsed root with no `0s` outside the reduced-motion query, nothing on the expanded root, the RTL mirror, the slide under resizable-owned state, a byte-identical rail, and the `isAnimated` warning: fires for the rail and for a custom width, once per mount, silent for the supported pair), 18 for focus (the six from round two plus: two navs each land on their own toggle, a nav without a toggle never borrows the other's, per-nav `handleRef` and mixed wirings, a toggle through resizable-owned state, a toggle rendered after its nav with a callback recreated every render, a shared toggle for two navs, an unmounted toggle, an in-nav toggle handed the config, and the release-and-warn path warning once) and 3 for the resize wrapper (same nav element and parent across collapse, no wrapper without resize, background passed through).
- [x] Mutation-tested: the 10 mutations from round two plus an association that matches every toggle (4 tests fail) and `step-end` swapped for `ease` (2 tests fail).
- [x] `SideNav.test.tsx`: 235 passing after merging main. Full core suite from the repo root: 8701 passing. `scripts/build-css.test.mjs`: 8 passing on a fresh `packages/core` build.
- [x] `pnpm -F @astryxdesign/core typecheck`, `CI=true eslint` on the changed files and `pnpm check:repo`: clean.
- [x] Verified in Chrome (numbers above).

https://claude.ai/code/session_012JEJSAwvsxWNW4t6NnYWgR


cixzhang (changes requested):
Thank you for this — it's a genuinely thoughtful PR, and the careful parts really show. Keeping the nav mounted and using `inert` (rather than unmounting) to drop the zero-width nav out of the focus order and a11y tree is exactly the right instinct, honoring `prefers-reduced-motion` by snapping is spot on, the width-precedence logic (explicit `collapsedWidth` beating both the rail class and the resizable size) is well-reasoned and well-tested, and it reuses the `--duration-fast` / `--ease-standard` tokens correctly. You also flagged the inert-timing question yourself, which is the right thing to have flagged. I'd love to get this in — two things to work through first.

**1. Animate `transform`, not `width` (performance + our motion convention).**
`rootAnimated` transitions `width`, which is a layout-triggering property — every frame reflows the nav's subtree and can shift sibling content, and there's no `will-change`/containment to soften it. Our design conventions ask that only `transform`/`opacity` animate for exactly this reason. Worth noting for framing: collapse on `main` is currently instant (no transition), so this is introducing the first collapse animation rather than matching an existing one — which is a nice opportunity to do it the compositor-friendly way from the start. Since the nav already has `overflow: hidden`, a `transform: translateX` slide (optionally paired with an inner fixed-width content wrapper so the contents don't squish) gets you the same visual without the per-frame layout. The `SideNavCollapseButton` precedent you cited actually animates `transform`, so this would also line the two up.
- https://github.com/facebookexperimental/xds/blob/8c2753bbd14a24bb6c5806e7c25a754cb08f8464/packages/core/src/SideNav/SideNav.tsx#L79-L86

**2. Focus handling when collapsing while focus is *inside* the nav (the timing question you raised).**
Right now `inert` is applied synchronously the moment collapse starts (`isFullyHidden = collapsed && collapsedWidth === 0`), while the nav is still visibly sliding out. For the normal case that's fine. The sharp edge is: if a keyboard user's focus is *inside* the nav when they collapse it, putting `inert` on an ancestor of the focused element yanks focus to `<body>` immediately — a "where did my focus go" moment, and it happens before the visual has caught up. The usual fix is to move focus somewhere deliberate first (e.g. the collapse toggle / trigger) and then apply `inert`, so focus never gets orphaned. Could you add handling + a test for the "focus was inside the nav at collapse time" case? That closes the loop on the exact concern you called out.
- https://github.com/facebookexperimental/xds/blob/8c2753bbd14a24bb6c5806e7c25a754cb08f8464/packages/core/src/SideNav/SideNav.tsx#L544-L560

Neither of these is a knock on the design — the API shape (`collapsedWidth` incl. `0`, `isAnimated`) reads well and is consistent with the component's existing `*Width` config. Once the animation moves to `transform` and the focus-on-collapse case is handled, I'm happy to approve. Thanks again for the care you put into this.

cixzhang:
Checking in — it's been a month, and the review above is still the only thing between this and landing.

Outstanding: animate `transform` rather than `width`, and move focus somewhere deliberate before `inert` lands when the nav collapses with focus still inside it.

Still want to carry it? Happy to take either piece on myself with credit if you'd rather not, or to close it if you've moved on — it'd be welcome back any time.

[Reviewed by Robohands]

AKnassa:
Still carrying it, thanks for the nudge. Both are in cf06135c7: the hidden collapse now animates transform only (the box holds its width for the slide through a zero-duration, delayed width transition, then snaps once), and a fully hidden collapse parks focus on the outside SideNavCollapseButton, or blurs, before inert lands. Also merged main. Details and the Chrome measurements are in the updated description.

cixzhang (changes requested):
Thanks for addressing the original review. Three current-head blockers remain:

- With two `SideNav` instances, collapsing the second moves focus to the first toggle; keyboard users land on the wrong control. Please associate each nav with its own toggle.
- `isAnimated: true` silently does nothing unless `collapsedWidth` is exactly `0`. Please prevent or warn on unsupported combinations, or make the promise hold across accepted values.
- The zero-duration width rule fails `scripts/build-css.test.mjs` on this exact head. Please preserve the transform-only result without emitting the rejected production rule.

[Reviewed by Robohands]

AKnassa:
Thanks, all three addressed in 1529fd96d, on top of a fresh merge of main.

- Two navs: the outside button now registers with the collapse state it shares with its nav (the controlled `onCollapsedChange`, or the deprecated `handleRef`), and the nav only parks focus on a toggle that matches. No match means a deliberate blur plus a one-time dev warning, never another nav's control.
- `isAnimated` without `collapsedWidth: 0` warns in development, naming the width it got. A type-level union was ruled out because a hoisted config widens `0` to `number`.
- The width hold is now `transition: width var(--duration-fast) step-end`. No intermediate width is computed and the only `0s` is under the reduced-motion query, so `scripts/build-css.test.mjs` passes on a fresh build. Measured in Chrome: the box holds 260 through 174.9ms and snaps at 175ms; expanding mid-slide snaps it open at once.

One more thing surfaced by the two-nav test: a resizable nav dropped its wrapper on collapse, which remounted the nav and sent focus to body before anything could park it (the plain rail collapse loses focus the same way today). The wrapper now stays and only the handle goes; it also inherits the background so the nav's own `inherit` reaches the surface.

Issue #2182, opened 2026-05-18
docs theme --dense causes AI agent context degradation (regression to 20% compliance)


cixzhang:
## Problem

`npm run xds -- docs theme --dense` outputs 12.5 KB / 308 lines covering HCT color models, derived variable registries, cascade layers, color scale expansion, motion curves, and type scale math. AI coding agents that read this doc to learn `defineTheme()` routinely abandon XDS components and fall back to raw CSS.

## Evidence from Monotelier benchmarks

The Monotelier benchmark builds an ASCII art canvas editor across 6 frontend stacks. XDS (source) compliance scores are highly variable, and the variance tracks whether the agent reads the theme docs:

| Run | Agent read theme docs? | Compliance | Anti-patterns | Outcome |
|-----|----------------------|------------|---------------|---------|
| run-009 | No (guessed) | **94%** | 0 | Used defineTheme({ name }) minimally, stayed on track |
| run-011 | Yes (--dense) | **45%** | 74 | Overwhelmed by HCT details, abandoned XDS components |
| run-012 | Yes (--dense) | **20%** | 70 | Same — total design system abandonment |

The theme doc is the single strongest predictor of framework compliance in XDS benchmarks. Every run where the agent reads it scores below 50%. Every run where the agent skips it scores above 70%.

## Root cause

The dense theme doc includes information the agent doesn't need to use `defineTheme()`:

- HCT color model theory (agents just need `[light, dark]` tuples)
- Derived variable registries (internal implementation)
- Cascade layer architecture (internal implementation)
- Color scale expansion math (agents don't build color scales)
- Motion curve definitions (not needed for basic theming)

This burns the agent's context budget. By the time it reaches actual component code, the relevant guidance has either decayed from context or the agent has been spooked into hand-rolling everything.

## Comparison with other frameworks

MUI and Ant Design score 90-95% consistently because their theming APIs (`createTheme({ palette: {...} })`, `ConfigProvider theme={{ algorithm: darkAlgorithm }}`) are in the LLM's training data. The agent doesn't need to read docs — it already knows the API shape.

XDS's `defineTheme()` is proprietary — the agent MUST read docs. When those docs dump architecture internals alongside the API, it backfires.

## Proposed fix

Add a `docs theme --api` or `docs theme --brief` flag that returns only the API surface:

- `defineTheme({ name, tokens, base? })` signature
- Token naming convention (`--color-*, --spacing-*, --radius-*, etc.`)
- `[light, dark]` tuple format
[...]

PR #3891, opened 2026-07-12
fix(cli): anchor dense/zh doc overlays to base sections instead of array index (#2182)

AKnassa:
Refs #2182.

I've scoped this as a self-standing bug fix rather than claiming to close out the umbrella work — the issue carries a `research` label and reads like it may be an umbrella for the broader "how should agents consume our docs" question. This defect is correct to fix on its own merits regardless of where that lands. Happy to retitle/close-link if you'd rather.

## The bug

The `--dense` and `--zh` reference-doc overlays were merged into the base doc **by array position**: for each base section index, the loader took the overlay entry at the same index and applied that overlay's *title* to the base section's *body*. An overlay ordered differently from the base — or missing a section — grafted every title onto the wrong content.

The output is actively misleading, not merely untidy:

```
$ astryx docs tokens --dense

## Spacing

defined in tokens.stylex.ts. gap props use space0-space12.

Token                      | Light   | Dark
--color-accent             | #0064E0 | #2694FE
--color-background-surface | #FFFFFF | #1F1F22
```

A heading reading **Spacing**, prose about gap props, and the **colour** table beneath it. An agent reading that learns spacing tokens are named `--color-*`.

The tokens overlay was misaligned **from index 0** — its own order is Spacing, Size, Color…, the base is Color, Spacing, Size… — so every one of its sections was wrong. `docs theme --dense` lost `Extending a Theme` and `Token Utilities` entirely and printed `useTheme` twice. `docs theme --zh` emitted four headings in both Chinese and English.

**Why this is more than a docs nit:** `CLAUDE.md` tells every AI agent to run `docs tokens --dense` and `docs theme --dense` at bootstrap. The scrambled output is the first thing an agent reads about this design system — a mechanically sufficient cause of the context degradation the issue reports. `principles` and `layout` happened to be index-aligned, which is why it went unnoticed: the two docs agents are *told* to read are the two that were broken.

## The fix

Overlay sections now name the base section they override:

```js
{ section: 'Spacing Tokens', title: 'Spacing', content: [...] }
```

Order becomes irrelevant, and an overlay may cover any subset — sections it doesn't name keep their base content. That's also why the partially-translated `zh` overlay now falls back to English cleanly instead of scrambling.

`docs-types.ts` previously *documented the footgun* ("Array indices must match base doc sections"); it now documents the anchor.

**One judgment call worth your eye:** the tokens overlay's `Typography` entry is **dropped**, not re-anchored. It carried no content overrides (title only), and the base has four separate typography sections (Font Family / Size / Weight, Type Scale) with no single section to anchor it to. Its only effect had been renaming whichever section sat at index 5 — the shadow table. If you'd rather keep a compressed typography title, tell me which base section should own it.

## Test plan

- [x] **18 tests red before the fix**, including the behavioural one (`docs tokens --dense` must not print `--color-*` under the Spacing heading).
- [x] After: `## Spacing` renders the spacing table, `## Color` the colour table, `theme --dense` shows all 14 headings with `Extending a Theme` restored and no duplicate `useTheme`.
- [x] **Mutation-tested:** restoring the positional merge fails 3 of the behavioural tests. Being precise — the Spacing/colour test does *not* fail under that mutation, because the realigned tokens overlay is now coincidentally index-correct too. The **anchor assertions** are the durable guard against future drift, not that one test.
- [x] `pnpm exec vitest run` — **337 files, 6318 tests, all passing.**
- [x] `pnpm -F @astryxdesign/core typecheck` — clean.


josephfarina (commented):
Dug into this and it's a solid, well-scoped fix — recommending merge.

**Verified:** reproduced on `main` (`docs tokens --dense` prints the color table under the `## Spacing` heading; `docs theme --dense` drops "Extending a Theme" and duplicates "useTheme" from index 5 on). The anchor-by-`section` approach is the right design, the 20 tests + drift guard are exactly right, and it's green + merges clean into `main`.

**Non-blocking nits:**
- The `docs-types.ts` union reformatting is unrelated churn — optional to drop.
- Block overrides are still index-matched *within* a section — a smaller residual of the same class; dovetails with the prose-only overlay direction @cixzhang raised on #3892.

**For the record, so we frame it accurately:** I ran ~46 controlled agent generations (scrambled vs fixed vs full vs no-docs) and couldn't reproduce the compliance-degradation causation in the description — scrambled ≈ fixed ≈ full on veer/purity/real-API usage, and no-docs was actually the worst. That doesn't diminish the fix at all; it's a correctness bug worth fixing on its own merits. Suggest we land it as the correctness fix it is and keep #2182 open for the doc-size angle — "Refs" (not "Closes") is already right.

josephfarina (approved):
Verified: reproduced the scramble on `main`, the anchor-by-`section` fix is correct, 20 tests + drift guard, full CI green, clean merge. Landing this as a standalone correctness fix — #2182 stays open for the doc-size work (full details in my review above).

PR #3892, opened 2026-07-12
fix(cli): translated component docs must not drop props

AKnassa:
Found while fixing #2182 — same bug class, different surface. **No issue filed for this one**; happy to open one if you'd prefer it tracked.

## The bug

A `docsZh` / `docsDense` block that carries its own `props` array was returned **wholesale** by the loader instead of being overlaid onto the English doc. That makes the translation a *replacement*, not an overlay — so any prop the translation hasn't caught up with simply ceases to exist.

```
$ astryx component Button --zh
# isInterruptible — absent.  isIconOnly — absent.
# Both exist in the English doc.
```

**Ten components** are affected. Seven show up as a plain `props`-length mismatch (Button, Card, Center, DateInput, FileInput, StatusDot, Tokenizer); three more — **MobileNav, Popover, Stack** — hide inside the multi-component `components[]` shape, which a length check doesn't see. The test found those three; I hadn't.

**Why it matters beyond i18n:** `CLAUDE.md` instructs every AI agent to run `astryx component <Name>` before touching a component. A prop that isn't in the docs is a prop that never gets used.

## The fix

Translations are overlaid, matched by prop **name** rather than position. A prop the translation doesn't cover falls back to its English entry instead of vanishing.

The prop's *contract* (`type`, `default`, `required`) also stays authoritative from the English doc while the translation supplies only the prose — so a stale translation can no longer misstate a type, which is the same failure mode one step further along.

Existing Chinese descriptions are untouched: `variant` still reads 视觉样式变体。

Same principle as #2182 — an overlay that covers a subset must not destroy what it doesn't cover.

## Test plan

- [x] **11 tests red before the fix**, naming each dropped prop.
- [x] After: EN and ZH prop counts match for every component (Button 17/17, FileInput 19/19, Tokenizer 29/29, Center 8/8), the dropped props are back, and the existing Chinese prose survives — a fix that quietly reverted everything to English would also have gone green, so I checked that explicitly.
- [x] `pnpm exec vitest run` — **337 files, 6487 tests, all passing.**

**One thing to flag honestly:** on my first full-suite run, `packages/cli/src/commands/template.path-safety.test.mjs > treats targetPath with .tsx extension as a file` failed. It is **not** caused by this change — it passes in isolation on this branch *and* on clean `main`, and a second unmodified full-suite run went green (337/337). It looks order-dependent under parallel execution. Mentioning it in case it's worth its own flake issue.

### DIFF
--- a/.changeset/component-doc-overlay-prop-drop.md
+++ b/.changeset/component-doc-overlay-prop-drop.md
@@ -1,0 +1,13 @@
+---
+'@astryxdesign/cli': patch
+---
+
+[fix] Translated component docs no longer drop props
+
+A `docsZh` / `docsDense` block that carried its own `props` array replaced the English component doc **wholesale** rather than overlaying it, so any prop the translation had not caught up with simply ceased to exist. `astryx component Button --zh` silently omitted `isInterruptible` and `isIconOnly`; ten components were affected, including `MobileNav`, `Popover` and `Stack` through the multi-component `components[]` shape.
+
+A reader of the translated docs cannot discover a prop that is not there — and `CLAUDE.md` instructs every AI agent to read `astryx component <Name>` before touching a component, so a missing prop is a prop that never gets used.
+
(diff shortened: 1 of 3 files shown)


cixzhang:
cc: @josephfarina for CLI docs.

It looks like translation docs can override prop definitions from the canonical docs. I wonder if we should strip down the translation docs to just the description and enforce that via type or lint? Aside from the description, the rest of the doc properties seem like they would need to be consistent with the canonical English docs to be correct.

cixzhang (approved):
Thanks for this fix!

PR #3893, opened 2026-07-13
fix(utils): clamp plainDateAddMonths to the target month's last day

arham766:
Thanks @cixzhang for the quick and thoughtful reviews on the recent PRs — the project has been genuinely good to contribute to. A few more fixes in the same vein incoming below this one.

`plainDateAddMonths` converted to a JS `Date` and called `setMonth`, which overflows excess days into the following month: adding a month to Jan 31 landed on Mar 3 (Mar 2 in leap years) instead of Feb 28/29, so month arithmetic from end-of-month dates skipped February entirely. The helper now does pure month arithmetic on the PlainDate fields and clamps the day to the target month's last day, matching `Temporal.PlainDate.add` and date-fns semantics.

Test plan:

```
pnpm exec vitest run packages/core/src/utils/plainDate.test.ts
```

New tests cover Jan 31 + 1 month (leap and non-leap years), month-end clamping in both directions, negative offsets across year boundaries, and that non-clamped dates are unaffected.

### DIFF
--- a/.changeset/plaindate-addmonths-clamp.md
+++ b/.changeset/plaindate-addmonths-clamp.md
@@ -1,0 +1,8 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] utils: clamp plainDateAddMonths to the target month's last day
+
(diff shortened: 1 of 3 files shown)


cixzhang (approved):
Thanks this is a good fix!

PR #3894, opened 2026-07-13
fix(lightbox): dismiss on backdrop click around the media

arham766:
Clicking the dark backdrop around the media never closed the Lightbox. The dismiss check only matched clicks where `e.target === e.currentTarget` (the dialog element), but the layout container fills the entire transparent dialog, so clicks on the visual backdrop always landed on the container and the check never passed. The container is now treated as backdrop too, and a pan drag that ends over the backdrop is ignored so releasing a drag doesn't accidentally dismiss.

Test plan:

```
pnpm exec vitest run packages/core/src/Lightbox/Lightbox.test.tsx
```

New tests cover that clicking the dark area around the media calls `onOpenChange(false)` and that clicking the media itself does not close the lightbox.

### DIFF
--- a/.changeset/lightbox-backdrop-dismiss.md
+++ b/.changeset/lightbox-backdrop-dismiss.md
@@ -1,0 +1,8 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Lightbox: make backdrop click dismissal actually reachable
+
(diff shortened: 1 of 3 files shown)


cixzhang (approved):
This makes sense. Thanks!

PR #3895, opened 2026-07-13
fix(thumbnail): show the placeholder when the image fails to load

arham766:
The Thumbnail docs promise a placeholder when the image fails to load, but the `img` element had no error handling at all, so a broken `src` rendered a broken image indefinitely. The component now tracks the errored src in state and falls back to the placeholder on `onError`, and retries when `src` changes so a corrected URL renders again.

Test plan:

```
pnpm exec vitest run packages/core/src/Thumbnail/Thumbnail.test.tsx
```

New tests cover that firing an error on the image swaps in the placeholder, and that changing `src` after an error retries the new image instead of staying stuck on the placeholder.

### DIFF
--- a/.changeset/thumbnail-error-placeholder.md
+++ b/.changeset/thumbnail-error-placeholder.md
@@ -1,0 +1,8 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Thumbnail: show the placeholder when the image fails to load
+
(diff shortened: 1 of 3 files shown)


cixzhang (approved):
Thanks for this adjustment! There's some merge conflicts with main. Would you be able to rebase?

PR #3896, opened 2026-07-13
fix(CommandPalette): discard in-flight search responses on close

arham766:
Closing the palette while a search request was still in flight let the late response re-commit the abandoned query and its results into the closed palette's state. On the next open the palette showed the ghost query and stale results instead of a fresh bootstrap. Most sources don't implement `cancel()`, and a response resolving after close still passed `runSearch`'s version check. `handleClose` now bumps `searchVersionRef`, so any pending request fails that check and is discarded.

Test plan:

```
pnpm exec vitest run packages/core/src/CommandPalette/CommandPalette.test.tsx
```

The new test types a query, closes the palette while the search promise is unresolved, resolves it after close, and asserts the palette reopens with the bootstrap state instead of the stale query and results.


cixzhang (approved):

cixzhang:
Thanks for this fix! I'll attempt a rebase for the merge conflicts.

cixzhang:
Thanks for this — the stale-response discard is a real fix and the version-counter approach is the right shape. Heads up that this now needs a rebase onto main, and when I rebased it locally the fix itself is clean (typecheck passes, conflict resolution is trivial) but your own test `discards a search response that resolves after the palette closed` fails on current main — after Escape the combobox still shows the typed value instead of clearing.

It's not your resolution — main changed underneath you. There's now an a11y loading announce inside `runSearch`'s `startTransition` (`announce(t('@astryx.commandPalette.loading'))`), and that shifts the `useOptimistic(search)` settle timing enough that `handleClose`'s `setSearch('')` no longer reverts `optimisticSearch` before render. Since the input value is bound to `optimisticSearch`, it doesn't clear on close anymore.

The surgical fix is to reset the optimistic value alongside the committed one on close — e.g. reset `optimisticSearch` in `handleClose` (a `startTransition(() => setOptimisticSearch(''))` or equivalent) so the input clears deterministically under the new code path. Could you rebase on main and add that? Everything else here is good to go once the input clears on close again. Thank you! 🙏

cixzhang:
Merged `main` into this branch on your behalf so it can land — `CommandPalette.test.tsx` conflicted only because both sides appended tests at the same spot, so I kept both. All 60 tests in `packages/core/src/CommandPalette` pass, `tsc` and `eslint` are clean. Nothing needed from you; I will merge once CI is green. Thanks for the fix — the invalidate-on-close approach was the right one.

PR #3897, opened 2026-07-13
fix(slider): round snapped values to the min/step decimal precision

arham766:
With fractional steps, `snapToStep` computed `min + steps * step`, which accumulates binary floating-point error: a keyboard nudge on a `step={0.1}` slider emitted `0.30000000000000004` through `onChange`/`onChangeEnd`, and the value leaked into `aria-valuenow` and the value tooltip once the consumer echoed it back. Snapped values are now rounded to the combined decimal precision of `min` and `step` (handling exponent-notation steps like `1e-7`, capped at 20 digits for `toFixed`). A legitimately snapped value can never carry more decimals than min and step combined, so the rounding removes only the error — exact steps are unchanged.

Test plan:

```
pnpm exec vitest run packages/core/src/Slider/Slider.test.tsx
```

New tests drive fractional-step sliders via keyboard in both single and range mode and assert `onChange` receives exact decimals (0.3, not 0.30000000000000004).

### DIFF
--- a/.changeset/slider-step-precision.md
+++ b/.changeset/slider-step-precision.md
@@ -1,0 +1,8 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Slider: round snapped values to the min/step decimal precision
+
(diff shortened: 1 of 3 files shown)


cixzhang (approved):
Nice thanks for fixing these fractionals

Issue #3899, opened 2026-07-13
[RFC] RichTextEditor — a WYSIWYG rich-text editor (Lexical)


potatowagon:
## Problem Statement

Astryx has plain-text input (`TextArea`) and read-only rich rendering (`Markdown`), but no editable **WYSIWYG rich-text surface**. Users authoring formatted content — release notes, comments, descriptions, docs — currently have to either drop to a raw `<textarea>` and hand-roll markdown, or pull in a full third-party editor that doesn't match Astryx tokens, theming, or the `Field` accessibility contract. There's no design-system-native way to let a user type **bold/italic/lists/headings/quotes/links/code** and get styled output that themes correctly in light/dark.

## Evidence of Demand

- **Internal:** Meta's `@nest/eps-lexical` package is a production rich-text editor used across Nest surfaces (release notes, comments, entity descriptions). There's an active effort to replace its bespoke primitives with a design-system editor — this RFC is the OSS foundation for that.
- **External convergence:** rich-text editing is a near-universal design-system need. Most mature systems ship or wrap one (Atlassian's `@atlaskit/editor`, WordPress Gutenberg on Lexical, many Chakra/MUI apps wrap Lexical/TipTap/Slate). Lexical itself is Meta's OSS editor framework with first-class React bindings, accessibility, and an immutable state model.
- **Frequency:** any product with user-authored formatted content hits this — it's not a one-surface pattern.

## Why Existing Components Don't Cover This

- `TextArea` is plain text only — no formatting, no serialized rich state, no inline marks.
- `Markdown` renders rich content **read-only**; it can't capture user edits.
- Composing `TextArea` + `Markdown` + a toolbar can't produce true WYSIWYG (no live formatting, no selection state, no undo/redo, no serialization round-trip). You'd be rebuilding an editor engine by hand, and you'd lose the `Field` a11y contract and token theming.

## Rough Approaches Considered

**Chosen: a Lexical-backed `RichTextEditor` (+ read-only `RichTextView`), landing in `lab` first.**

Lexical is the natural engine: framework-agnostic core, official React bindings, immutable state (undo/redo for free), accessibility built in, and it's Meta OSS. The Astryx value-add is: StyleX/token theming, `Field` integration (label/description/status/required), and a deliberately small, extensible API.

```tsx
import {RichTextEditor, RichTextView} from '@astryxdesign/lab';

// editable
<RichTextEditor
  label="Notes"
  placeholder="Write something…"
[...]

PR #3900, opened 2026-07-13
[RFC #3899] feat(lab): experimental Lexical rich-text editor (RichTextEditor + RichTextView)

potatowagon:
> **Draft / RFC-gated.** This accompanies RFC #3899. Per Astryx's contribution protocol, new components land in `@astryxdesign/lab` first and graduate to `@astryxdesign/core` after the Component Specification Protocol. Opening as a Draft to make the RFC concrete — not requesting merge yet.

## What

Adds an experimental **Lexical WYSIWYG rich-text editor** to `@astryxdesign/lab`:

- **`RichTextEditor`** — editable WYSIWYG surface: headings, bulleted/numbered lists, quote, code block, links, bold/italic/underline/strikethrough/inline-code, markdown-shortcut typing, undo/redo history, tab indentation. Wraps the existing `@astryxdesign/core` `Field` for label/description/status/required so it matches `TextInput`/`TextArea` idioms.
- **`RichTextView`** — read-only renderer for serialized editor state.
- **Extension points** — `nodes` and `plugins` props let consumers layer custom nodes (mentions, images) and plugins (toolbars, autolink, hover cards) on top **without forking** the editor.

Styling is 100% StyleX + design tokens — no stylesheet, dark-mode/theming for free. Lexical's theme slots are mapped to StyleX-generated class names via a shared `editorTheme.ts` so the editor and read-only view render identically.

## Why `lab`, and why optional peers

- New components go through the RFC + Component Specification Protocol and land in `lab` first (per the [Contributing guide](https://github.com/facebook/astryx/wiki/Contributing)). `lab` is `@canary`-only, so this never touches a stable release.
- `lexical` and the `@lexical/*` packages are declared as **optional peer dependencies** (`peerDependenciesMeta.*.optional = true`), pinned to `^0.46.0`. Nothing pulls Lexical in unless a consumer imports the editor.
- **`@astryxdesign/core` is untouched** — its zero-runtime-dep promise is intact. (On eventual graduation to core, the plan is a dedicated `./editor` subpath excluded from the barrel so that promise still holds.)

## Version note

Pinned to `lexical ^0.46.0`. `0.47.0` is currently blocked by the repo's `minimumReleaseAge: 10080` (7-day) supply-chain guard — it's <7 days old — so `0.46.0` is the newest installable line.

## Usage

```tsx
// install the optional peers first:
//   pnpm add lexical @lexical/react @lexical/markdown @lexical/link \
//            @lexical/list @lexical/rich-text @lexical/code @lexical/utils
import {RichTextEditor, RichTextView} from '@astryxdesign/lab';

<RichTextEditor
  label="Notes"
  placeholder="Write something…"
  onChange={state => save(JSON.stringify(state.toJSON()))}
/>

<RichTextView value={storedEditorStateJSON} />
```

## Validation

- `pnpm -F @astryxdesign/lab typecheck` — clean (against real Lexical 0.46 types)
- `pnpm exec vitest run RichTextEditor` — 11/11 tests pass
- `eslint src/RichTextEditor` — clean
- `check:changesets`, `sync-exports:check` — pass (no changeset: lab is private/canary-only)

http://localhost:6006/?path=/story/lab-richtexteditor--with-description

https://github.com/user-attachments/assets/91ff5f65-b190-4e0a-811e-4049ebac65c1

## Files

```
packages/lab/src/RichTextEditor/
  RichTextEditor.tsx        # editor component
  RichTextView.tsx          # read-only renderer
  editorTheme.ts            # shared StyleX → Lexical theme mapping
  index.ts                  # component entry
  RichTextEditor.test.tsx   # colocated tests (11)
  RichTextEditor.doc.mjs    # component docs
packages/lab/src/index.ts               # barrel re-export (RFC-referenced)
packages/lab/package.json               # optional Lexical peers
apps/storybook/stories/RichTextEditor.stories.tsx
```

## Context

OSS foundation for replacing Meta-internal EPS-Lexical primitives with an Astryx-based editor. The Meta-coupled pieces (intern mentions, entity hover cards, write-with-AI, Relay/analytics) stay internal and compose on top via the `nodes`/`plugins` extension points — intentionally **not** part of this PR.

---
*Companion to RFC #3899.*

### DIFF
--- a/packages/lab/package.json
+++ b/packages/lab/package.json
@@ -2,7 +2,7 @@
   "name": "@astryxdesign/lab",
   "version": "0.1.7",
   "private": true,
-  "description": "Experimental Astryx components — published to npm only under the @canary dist-tag for early testing; never released as a stable (latest) version.",
+  "description": "Experimental Astryx components \u2014 published to npm only under the @canary dist-tag for early testing; never released as a stable (latest) version.",
   "author": "Meta Open Source",
   "license": "MIT",
   "homepage": "https://github.com/facebook/astryx#readme",
@@ -22,7 +22,7 @@
   ],
   "astryx": {
     "canaryOnly": true,
-    "_comment": "Published ONLY to the @canary dist-tag (see .github/workflows/release.yml). Never published as a stable `latest` release. `private: true` is intentional and must stay — it is npm's hard guarantee against an accidental stable publish."
+    "_comment": "Published ONLY to the @canary dist-tag (see .github/workflows/release.yml). Never published as a stable `latest` release. `private: true` is intentional and must stay \u2014 it is npm's hard guarantee against an accidental stable publish."
   },
   "main": "./dist/index.js",
   "module": "./dist/index.js",
@@ -59,7 +59,15 @@
     "@stylexjs/stylex": ">=0.10.0",
     "@astryxdesign/core": "0.1.7",
     "react": ">=19.0.0",
-    "react-dom": ">=19.0.0"
+    "react-dom": ">=19.0.0",
+    "lexical": "^0.46.0",
+    "@lexical/react": "^0.46.0",
+    "@lexical/markdown": "^0.46.0",
+    "@lexical/link": "^0.46.0",
+    "@lexical/list": "^0.46.0",
+    "@lexical/rich-text": "^0.46.0",
+    "@lexical/code": "^0.46.0",
+    "@lexical/utils": "^0.46.0"
   },
   "dependencies": {
     "d3-array": "^3.2.4",
@@ -67,7 +75,42 @@
     "d3-shape": "^3.2.0"
   },
   "devDependencies": {
+    "@lexical/code": "^0.46.0",
+    "@lexical/link": "^0.46.0",
+    "@lexical/list": "^0.46.0",
+    "@lexical/markdown": "^0.46.0",
+    "@lexical/react": "^0.46.0",
+    "@lexical/rich-text": "^0.46.0",
+    "@lexical/utils": "^0.46.0",
+    "@types/d3-array": "^3.2.2",
     "@types/d3-scale": "^4.0.9",
-    "@types/d3-shape": "^3.1.8"
+    "@types/d3-shape": "^3.1.8",
+    "lexical": "^0.46.0"
+  },
+  "peerDependenciesMeta": {
+    "lexical": {
+      "optional": true
+    },
+    "@lexical/react": {
+      "optional": true
+    },
+    "@lexical/markdown": {
+      "optional": true
+    },
+    "@lexical/link": {
+      "optional": true
+    },
+    "@lexical/list": {
+      "optional": true
+    },
+    "@lexical/rich-text": {
+      "optional": true
+    },
+    "@lexical/code": {
+      "optional": true
+    },
+    "@lexical/utils": {
+      "optional": true
+    }
   }
 }
--- a/.changeset/build-scope-source-condition.md
+++ b/.changeset/build-scope-source-condition.md
@@ -1,0 +1,11 @@
+---
+'@astryxdesign/build': patch
+---
+
+[fix] Scope the `source` resolve condition to @astryxdesign packages in withAstryx
+
+`withAstryx` set webpack's `conditionNames` to `['source', …]` globally, which resolved *any* dependency shipping a `source` export to its raw TypeScript — not just Astryx packages. Third-party deps that ship a `source` export (e.g. `lexical`, pulled in by the new RichTextEditor lab component) were then fed untranspiled `.ts` through Next's babel and failed on syntax like `declare` class fields.
+
+The `source` condition is now applied via a scoped **allowlist** `module.rules` entry that only matches `node_modules/@astryxdesign/*` (with `conditionNames: ['source', '...']`, so it augments rather than replaces Next's defaults). The global `config.resolve.conditionNames` is left as Next's default, so all other packages — including any future third-party dep that happens to ship a `source` export — resolve to their built output. React JSX resolution is preserved via the inherited `'...'` defaults + `react-server` condition. Astryx source builds are unaffected.
+
+@potatowagon
--- a/apps/storybook/stories/RichTextEditor.stories.tsx
+++ b/apps/storybook/stories/RichTextEditor.stories.tsx
@@ -1,0 +1,133 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+import {useState} from 'react';
+import type {Meta, StoryObj} from '@storybook/react';
+import {
+  RichTextEditor,
+  RichTextView,
+} from '@astryxdesign/lab';
+import type {EditorState} from 'lexical';
+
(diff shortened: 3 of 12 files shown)


potatowagon:
Note on the failing `test` check — it looks unrelated to this PR:

- The failure is in `packages/cli/src/commands/build-theme.watch.test.mjs` → *"builds initially, rebuilds on change, and stops cleanly on SIGINT"* (`expected false to be true` at line 131, the `/rebuild/i` stdout assertion).
- This PR touches only `packages/lab/` (the RichTextEditor component), the storybook story, `packages/build/src/next.js`, and the lockfile — **no changes to `packages/cli/`**.
- The test spawns `astryx theme build --watch` as a child process, edits a theme file, and polls up to an 8s timeout for a chokidar rebuild. In the run it failed at ~8.6s — right at the `waitFor` ceiling — which is the signature of a load-induced timing flake under the full ~6,300-test parallel suite.
- It **passes in isolation** (`vitest run build-theme.watch` → 3/3, ~0.7s) and passed on earlier CI runs of this same branch.

The RichTextEditor's own tests pass (11/11), and `build`, `build-sandbox`, `build-storybook`, `lint`, `smoke-test`, and `docsite-test` are green.

Happy to help if maintainers believe it's not a flake — I didn't want to touch an unrelated CLI test from a component PR. A re-run of the `test` job would likely confirm.

humbertovirtudes (commented):
Reviewing as an RFC companion, not as a merge gate (the description says it's not requesting merge yet). The architecture is solid: Lexical as optional peer deps keeps core's zero-dep promise intact, the StyleX-token theme is shared between editor and view so they render identically, the `nodes`/`plugins` extension points let consumers layer mentions/toolbars without forking, and the a11y wiring (role=textbox, aria-multiline, labelledby/describedby, required/invalid) is careful. Helper components are at module scope, tests are colocated, and the doc + story are present.

A few things to reconcile before this graduates past RFC:

**1. (major) The changeset contradicts the `next.js` implementation.** The changeset for `@astryxdesign/build` says the `source` condition is now "applied via a scoped `module.rules` entry that only matches `node_modules/@astryxdesign/*`" (an allowlist). The code does the opposite: `config.resolve.conditionNames` stays global with `source` first, and the unshifted rule *excludes* `lexical|@lexical/*|yjs` (a blocklist). Two consequences: the published `@astryxdesign/build` changelog will be factually wrong, and the blocklist is the more fragile design: the next third-party dep that ships a `source` export re-breaks the same way, whereas the allowlist the changeset describes would be robust. Worth reconciling the two, and preferring the allowlist if React JSX resolution allows it (the comment hints it may not, which would be good to confirm and capture).

**2. (minor) CI `test` is red, but on an unrelated test.** The failure is `packages/cli/src/commands/build-theme.watch.test.mjs` ("expected false to be true" on a watch-rebuild timing assertion), with 6341/6342 passing. It doesn't touch anything this PR changes and looks like a watch-mode timing flake under a heavy CI install. A re-run should confirm; if it persists it's a pre-existing flake rather than this change.

**3. (minor) Needs a rebase.** Currently shows conflicts with main (the description notes an in-flight FieldStatus PR on the same files).

**4. (minor) `RichTextView` error handling on the read path.** `onError` re-throws. A read-only view renders *persisted* JSON, which is exactly where stale or foreign-schema state tends to show up, and a bad `value` would throw during render and can take down the host. Consider a fallback or a documented error-boundary contract for the view specifically.

**5. (minor) `RichTextView` has a single test** and no coverage for its `nodes`/`plugins`/`namespace` extension points or malformed input.

Nits: `yjs` in the exclusion regex is speculative (it only arrives transitively via `@lexical/react`), `initialConfig` is rebuilt every render though `LexicalComposer` only reads it on mount (harmless), and the package.json diff swaps literal em-dashes for `\u2014` escapes, which is unrelated churn.

The headline is #1: the changeset and code disagree about a shared build package. Everything else is polish.

potatowagon:
Thanks for the thorough review @humbertovirtudes — really helpful. Addressed all five, pushed as two clean commits (rebased onto latest `main`).

**#1 (major) — changeset ↔ code reconciled, and switched to the allowlist.** You were right on both counts: the changelog was factually wrong, and the blocklist was the fragile design. I've moved to the **allowlist** the changeset described. The reason I'd originally reached for the blocklist was that my first allowlist attempt broke React JSX (`(0, d.jsxDEV) is not a function`) — but that was because I was *replacing* the global `config.resolve.conditionNames` (dropping Next's `'...'` sentinel + `react-server`). The fix is to leave the global conditions alone and scope `source` via a `module.rules` entry that *augments* rather than replaces:

```js
config.module.rules.unshift({
  test: /[\\/]node_modules[\\/]@astryxdesign[\\/]/,
  resolve: { conditionNames: ['source', '...'] },  // '...' inherits Next's defaults
});
```

Verified the sandbox `next build` compiles cleanly with this (no `declare`/`jsxDEV` errors, lexical resolves to built JS). The changeset now describes exactly this. This is robust to the next third-party dep that ships a `source` export.

**#2 (minor) — `test` flake.** Agreed it's unrelated (`build-theme.watch.test.mjs`, watch-rebuild timing under heavy CI). It passes in isolation (3/3, ~0.7s). Didn't touch it — a re-run should clear it.

**#3 (minor) — rebased** onto latest `main`; conflicts resolved (clean rebase, no remaining conflicts).

**#4 (minor) — `RichTextView` read-path error handling.** Good call — a read-only view renders persisted/foreign JSON, exactly where bad state shows up. It no longer re-throws: malformed JSON (or state with unregistered node types) now renders an optional `errorFallback` and calls a new `onParseError` callback instead of throwing during render and taking down the host. `value` is JSON-validated before it reaches `LexicalComposer` (a parse throw during init would otherwise escape the error boundary), and the Lexical `onError` routes to the same fallback path. State recovers when a corrected `value` comes in.

**#5 (minor) — `RichTextView` tests.** Added coverage for `plugins`, `namespace`, extra `nodes`, malformed-JSON → fallback + `onParseError`, and the no-fallback path. 11 → 16 tests, all green.

**Nits:**
- `yjs` was in the old blocklist regex — the allowlist removes the need for that speculative entry entirely, so it's gone.
- `initialConfig` rebuilt each render — left as-is per your note that it's harmless (`LexicalComposer` reads it only on mount); happy to memoize if you'd prefer.
- package.json em-dash/`\u2014` churn — that was in the auto-generated lockfile region, not hand-authored; the component `package.json` diff is just the optional Lexical peers.

cixzhang (approved):

packages/lab/package.json
```diff
@@ -59,15 +59,58 @@
...
+    "@lexical/react": "^0.46.0",
+    "@lexical/rich-text": "^0.46.0",
+    "@lexical/utils": "^0.46.0",
+    "@types/d3-array": "^3.2.2",
     "@types/d3-scale": "^4.0.9",
-    "@types/d3-shape": "^3.1.8"
+    "@types/d3-shape": "^3.1.8",
+    "lexical": "^0.46.0"
+  },
+  "peerDependenciesMeta": {
```
cixzhang, line 90: We can clean these up :)

PR #3901, opened 2026-07-13
chore: add knip dead-code detection

cixzhang:
## What

Adds [knip](https://knip.dev) for dead-code detection across the monorepo:

- `knip` dev dependency (pinned to 6.24.0)
- `knip.jsonc` monorepo config
- `pnpm check:deadcode` script (`scripts/check-deadcode.mjs`)

## Why

There is no tooling to find unused files, exports, types, or devDependencies. They accumulate silently: a helper no one calls anymore, a type left behind after a refactor, a devDependency for a workflow that was removed. knip finds all of these across workspaces.

## Design notes

**Public-API safe.** knip reads each package's `exports`/`bin` from `package.json` as entry points, so a package's published API is never reported as unused. The config only adds the safety rules `package.json` cannot express:

- `*.doc.mjs` and story files are treated as entry points (consumed by the docs pipeline, not imported by source)
- `packages/cli/templates/**` is ignored (copied verbatim into user projects, intentionally standalone)
- `internal/vibe-tests/.baseline/**` is ignored (reference fixtures, compared against, not imported)
- `dist/`, `build/` output is ignored

Without these rules a naive run reports ~1,500 "unused" files (nearly all templates and test/doc files) and the entire public API. With them, findings drop to a small, real set: unused internal exports, unused types, unused devDependencies.

**Report, not a gate.** This does not run in CI and does not block anything. Run `pnpm check:deadcode` for a readable report, or `pnpm check:deadcode --reporter json` for machine-readable output.

**Small-host memory note.** knip's parser reserves a large virtual buffer up front. On Linux hosts with little RAM and no swap, that reservation can be refused before any work happens. The runner sets `vm.overcommit_memory=1` when permitted (virtual only; physical use stays proportional to real data) and otherwise prints an actionable hint instead of a cryptic allocation error.

## Testing

- `pnpm check:deadcode` runs clean end-to-end on the full workspace
- `pnpm check:deadcode --reporter json` produces valid JSON
- Verified templates, `.baseline` fixtures, and doc files are correctly excluded; public exports are not flagged
- `check:repo` gates (sync, changesets, executable-bits) still pass


josephfarina (approved):

thedjpetersen (approved):
Approved.

cixzhang:
Closing: the dead-code detector is being moved into the Night Watch operator's workspace instead of the repo toolchain. knip only needs to read the repo, so the binary, config, and runner live outside it and point at a worktree via --directory/--config. This keeps a new devDependency and lockfile churn out of the public repo; the role's output (deletion PRs) is the reproducible artifact.

PR #3902, opened 2026-07-13
chore: remove dead code (2026-07-13)

cixzhang:
Removes one unused internal file detected by an out-of-repo dead-code scan.

## Removed

**Unused internal file (packages/core)**

- `packages/core/src/Carousel/carousel.markers.stylex.ts`: its only export, the `carouselScope` StyleX marker, is imported by nothing. It is a leftover from the original Carousel implementation; the component composes its scroll styles without this marker. It is not part of any package `exports` map and is not referenced from any `.doc.mjs`, story, or config.

Verified before deleting: a repo-wide search for `carouselScope` and `carousel.markers` returns no references outside the deleted file, `pnpm build` and `pnpm test` are green, and `scripts/sync-exports.js` reports no change to the public exports map.

## Needs Review (not removed)

These were flagged by the scan but fall outside the safe auto-remove set, so they are left for human judgment:

- `packages/core/src/NavItem/index.ts`: reported as an unused barrel, but `scripts/sync-exports.js` lists `NavItem` in `INTERNAL_DIRS` as a deliberate internal module. Current consumers import `navItemStyles.stylex.ts` directly rather than through the barrel. Removing the barrel would need a matching update to the codegen script, which is out of scope for a mechanical cleanup.
- Unused exports and types in `packages/core` (for example `getDefaultAudioContext`, `toneToY`, `yToTone`, `ToggleButtonGroupContext`, and roughly 13 types in `docs-types.ts`). The `docs-types.ts` types are re-exported from the public `src/index.ts` barrel and referenced from `.doc.mjs` files, so they are public surface, not dead. The remaining exports are used within their own module, meaning the `export` keyword is redundant rather than the symbol being dead, which is a source-edit judgment call.
- `packages/lab` and `packages/charts` internal exports and types (roughly 80 findings). Both packages expose a single `.` entry, so the scan reports many internal symbols; each removal is a surgical source edit best reviewed by an owner.
- Unused `devDependencies`: `gpt-tokenizer` and `@astryxdesign/theme-neutral` in `packages/cli` are also declared as `peerDependencies` and are consumed by `apps/sandbox` through workspace hoisting, so removing the devDependency line is not mechanically safe. `@types/d3-array` in `packages/charts` and `packages/lab` is coupled to the `d3-array` runtime dependency, which is the Dependencies role's area.
- `apps/*` findings (files, exports, types): app module graphs under-resolve in static analysis and are flag-only by policy.
- Duplicate exports in `packages/core/src/theme/tokens.stylex.ts` and `packages/core/src/naming.ts`: renames are judgment calls.
- Unlisted, unresolved, and unused runtime dependencies: deferred to the Dependencies and Quartermaster roles.

Night Watch — Dead Code


thedjpetersen (approved):
Approved.

josephfarina (approved):

PR #3903, opened 2026-07-13
fix(a11y): correct rest-spread precedence in Avatar components

cixzhang:
## Summary

Fixes rest-spread precedence in Avatar, AvatarStatusDot, AvatarGroup, and AvatarGroupOverflow so contract props (role, aria-label, aria-hidden) cannot be clobbered by consumer pass-through attributes.

The pattern: `{...rest}` must come **before** contract props for the component to win on semantics it computes. Previously, rest/props was spread after these attributes, meaning a consumer passing `role` or `aria-label` through BaseProps would silently override the computed accessibility state.

## Changes

| File | Fix |
|------|-----|
| `Avatar.tsx` | `{...props}` moved before `role`, `aria-label`, `aria-hidden` |
| `AvatarStatusDot.tsx` | `{...props}` moved before conditional `role`/`aria-label` |
| `AvatarGroup.tsx` | `{...props}` moved before `role="group"` |
| `AvatarGroupOverflow.tsx` | `{...rest}` moved before `aria-label` (both button + span) |

## Verification

- All 22 existing Avatar/AvatarGroup tests pass
- tsc --noEmit clean

### DIFF
--- a/packages/core/src/Avatar/Avatar.tsx
+++ b/packages/core/src/Avatar/Avatar.tsx
@@ -301,6 +301,7 @@ export function Avatar({
   return (
     <AvatarSizeContext value={numericSize}>
       <div
+        {...props}
         ref={ref}
         role={isDecorative ? 'presentation' : 'img'}
         aria-label={isDecorative ? undefined : accessibleName}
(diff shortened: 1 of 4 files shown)


josephfarina (approved):

thedjpetersen (approved):
Approved.

PR #3682, opened 2026-07-08
feat(collapsible): use hasDividers boolean on CollapsibleGroup

cixzhang:
## What this does

Follow-up to #3499: aligns `CollapsibleGroup`'s divider API to the repo-standard **`hasDividers` boolean** — the same convention `List` and `CheckboxList` already use — instead of the `dividers="between" | "all" | "none"` enum that shipped in #3499.

```tsx
// Before (#3499)
<CollapsibleGroup type="single" dividers="between">…</CollapsibleGroup>

// After
<CollapsibleGroup type="single" hasDividers>…</CollapsibleGroup>
```

## Why

`#3499` proposed a `dividers` enum mirroring `Table`. In review we noted that `Table`'s `'rows' | 'columns' | 'grid'` values describe cell geometry that doesn't map to a vertical accordion, and that the established pattern for a simple between-item divider is the `hasDividers` boolean (`List`, `CheckboxList`). Keeping accordions consistent with `List` is less surface area to learn and matches what consumers already reach for.

## What changed

- **`dividers` enum → `hasDividers` boolean** (`@default false`) on `CollapsibleGroup`.
- **Dropped the `'all'` outer-edge mode.** Hairlines are drawn between items only, exactly like `List`. Groups that want an outer frame can wrap in `Card` (the documented alternative). This removes the `wrapperAll` border styles and the `--border-width`/`--color-border` wrapper border.
- **`density` is unchanged** — still `'compact' | 'balanced' | 'spacious'`, still defaults to `'balanced'` when dividers are on and stays available without dividers.
- **DOM-less contract preserved** — without `hasDividers` the group still renders no wrapper element.
- Removed the now-unused `CollapsibleGroupDividers` type export and the `data-dividers` reflection (the wrapper class signals divided mode, mirroring how `List` reflects only `density`).
- Updated stories (`Dividers`, `Dividers — Multiple`, `Dividers — Density`), the `CollapsibleDividedAccordion` showcase block, component docs (EN/ZH/dense), theming targets, and the 32 Collapsible tests.

Since the `dividers` prop from #3499 has **not been released** yet, this is a pre-release API refinement — no codemod is needed and the existing changeset is updated in place to describe the final `hasDividers` shape.

## Test plan

- [x] `packages/core` Collapsible suite: 32/32 pass
- [x] `@astryxdesign/core` typecheck clean
- [x] Docsite generate + 217/217 tests pass
- [x] CLI template-docs typecheck clean
[...]

PR #3904, opened 2026-07-13
ci(review): source eng/design owners from repo variables; eng-owner clears code gate

cixzhang:
## What

Rework the review gate so it no longer depends on the GitHub content API, and simplify who clears what. Motivated by PRs #3682 and #3906 getting stuck.

### The bug
The flag/clear jobs read `.github/ENGOWNERS` / `DESIGNOWNERS` / `CODEOWNERS` via the content API at `ref=main` to bucket the author and resolve entitled approvers. That API intermittently returned **502s / HTML edge error pages**, and the old code fell back to an empty owner list — twice mis-bucketing an eng owner (cixzhang) as a community contributor and wrongly gating the PR, with no entitled approver able to clear it.

### The fix
**Source owners from repo variables, not files/API.** Eng and design owners now come from `vars.ENGOWNERS` / `vars.DESIGNOWNERS` (passed in as env). There is no checkout, no file read, and no content API call — so there is nothing left to 502 on.

- **Code gate** is satisfied by any **ENGOWNER** approval; **design gate** by any **DESIGNOWNER** approval.
- **CODEOWNERS is no longer read by the workflow.** It's left to GitHub-native auto-request (which requests the codeowners on every PR regardless of branch protection), and since the codeowners are a subset of ENGOWNERS they still clear the code gate.
- **Reviewer requests:** code review is covered by GitHub-native CODEOWNERS. The workflow only adds **design owners** as reviewers, and **only when a design gate fires** — so design owners aren't pinged on unrelated PRs.
- **Fail-safe preserved:** an empty/unset owner variable **throws** in the flag job (per-PR skip, gate left untouched, run fails visibly) and **skips clearing** in the clear job. An empty list is never treated as "no owners" / "everyone".
- Removes the sparse checkout and all file/API read machinery from both jobs (net −98 lines).

## Testing
- Both YAMLs parse (js-yaml); both github-script blocks compile as async JS.
- Unit-tested the vars-based parse + entitlement logic (13 assertions): eng approvals clear the code gate but not design (the #3682/#3906 scenario); design owners clear design only; empty/whitespace vars throw (fail-safe); the clear-job guard fires on an empty var.

## Notes
- #3682/#3906 were unstuck via manual re-flag; this prevents recurrence.
- Owner-list changes are now made via repo variables (Settings → Variables), not a PR.
- A separate change will make the Copilot reviewer content-only (drop author-based framing); tracked independently.

### DIFF
--- a/.github/workflows/review-clear.yml
+++ b/.github/workflows/review-clear.yml
@@ -38,22 +38,10 @@ jobs:
       checks: write
       contents: read
     steps:
-      # Read-only, sparse checkout of our default branch — only the owner files,
-      # never PR-controlled code. Lets the entitled-approver read below hit the
-      # local filesystem instead of the content API (which has returned
-      # transient 502s on ref=main). This job runs on workflow_run, so the
-      # default branch is our own trusted code.
-      - name: Check out owner files (default branch, read-only)
-        uses: actions/checkout@v7
-        with:
-          ref: ${{ github.event.repository.default_branch }}
-          sparse-checkout: |
-            .github/ENGOWNERS
-            .github/CODEOWNERS
-          sparse-checkout-cone-mode: false
-          persist-credentials: false
       - name: Clear the code gate if an entitled owner approved
         uses: actions/github-script@v9
+        env:
+          ENG_OWNERS: ${{ vars.ENGOWNERS }}
         with:
           script: |
             const { owner, repo } = context.repo;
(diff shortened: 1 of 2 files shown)


thedjpetersen (approved):
Approved.

josephfarina (approved):

josephfarina:
Heads up: the `test` job in the CI workflow is red on this one (run 29244734537). The diff is workflow-YAML-only so it's very likely a pre-existing/flaky failure unrelated to the review-gate changes, but worth a glance before merge to confirm it's not something the change surfaced. Everything else (lint, docsite-test, smoke-test, build-storybook) is green.

PR #3910, opened 2026-07-13
cli: add hiddenComponents config to hide replaced components from unscoped surfaces

andrskr:
## Use case

I'm experimenting with building a design system on top of Astryx: core from npm, a custom theme, and an integration package (`astryx.integration.mjs`) that ships a few of my own components. Some of those replace core components — my `Dialog` should be used instead of core's, everywhere.

Today there's no way to express that. With an integration providing `Dialog`:

- `component --list` shows both Dialogs
- `component Dialog` throws `provided by multiple packages. Re-run with --package` — reasonable as a default, but there's no way to resolve it permanently
- `search dialog` ranks core's Dialog first

For humans that's a papercut. For coding agents it's worse: the generated AGENTS.md tells them "discover, don't guess" through the CLI, and the CLI steers them to the component that's been replaced.

## What this adds

A `hiddenComponents` array in `astryx.config`:

```js
export default createConfig({
  integrations: ['@acme/ds'],
  hiddenComponents: ['Dialog'], // or '@acme/ds/Dialog' to hide an integration's
});
```

- `'Name'` hides the core component; `'@scope/pkg/Name'` hides that package's (name = segment after the last `/`, `XDS` prefix normalized)
- Hidden components disappear from `component --list` (all detail levels), bare-name resolution (so a single remaining provider resolves cleanly instead of the ambiguity error), fuzzy suggestions, and `search`
- **Explicit `--package` lookups still resolve hidden components** — docs stay readable and `swizzle --package` keeps working. Hiding steers, it doesn't forbid
- Config errors never break discovery: malformed entries are ignored, and a hidden-everything typo degrades to "unknown component" with suggestions

## Implementation notes

- `parseHiddenComponents()` / `isComponentHidden()` live in `lib/component-discovery.mjs` next to the discovery code they filter; `Project` exposes the raw config entries
- `api/component.mjs`'s `loadIntegrationsSafely` became `loadProjectContextSafely` (returns integrations + parsed hidden set) — same error-swallowing contract, one `Project.load` per call as before
- Unscoped owner filtering happens once (`visibleOwners`); the `--package` paths intentionally keep the unfiltered set
- No behavior change for projects without the config field (schema is additive; every existing surface returns identical output)

## Testing

- New `component.hidden.test.mjs` (10 tests, same temp-consumer harness as `template-integration.test.mjs`): parse/normalization, baseline ambiguity preserved, core-hidden → integration authoritative, integration-hidden → core resolves, list filtering, `--package` escape hatch, hidden-only-provider → unknown, search exclusion
- Full `packages/cli` suite: 1700 passed (3 unrelated files are timeout-flaky on a loaded machine; they pass in isolation before and after this change)
- ESLint clean on changed files

Happy to adjust naming (`hiddenComponents` vs `components.hidden`) or split the search change out if you'd rather keep this minimal.

### DIFF
--- a/.changeset/hidden-components-config.md
+++ b/.changeset/hidden-components-config.md
@@ -1,0 +1,5 @@
+---
+'@astryxdesign/cli': minor
+---
+
+Add `hiddenComponents` to `astryx.config` — hide a component from unscoped CLI surfaces (`component --list`, bare `component <Name>` resolution, `search`) so an integration-provided replacement becomes the single authoritative provider instead of an ambiguity error. Entries are `'Name'` (core) or `'@scope/pkg/Name'`. Explicit `--package` lookups still resolve hidden components, so their docs and swizzle stay reachable.
--- a/packages/cli/src/api/component.hidden.test.mjs
+++ b/packages/cli/src/api/component.hidden.test.mjs
@@ -1,0 +1,226 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+/**
+ * @file Config-driven component hiding (`hiddenComponents`).
+ *
+ * These tests stand up a temp consumer project with a fake core package and
+ * an installed integration that provides a same-named replacement component,
+ * then exercise the public `component()` / `search()` APIs to verify that
+ * hiding removes a provider from unscoped surfaces (list, bare-name
+ * resolution, search) while explicit --package scoping still resolves it.
+ *
+ * Each test gets its OWN consumer dir: astryx.config.mjs is loaded via
+ * dynamic import, so a reused path would serve a stale module from the ESM
+ * cache.
+ */
+
+import {afterEach, describe, expect, it} from 'vitest';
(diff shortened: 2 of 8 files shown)


josephfarina:
Thanks for this it's a well-built PR and a real use case! The coding-agent angle especially (the CLI steering agents to a component that's been replaced) is a gap worth fixing.

We don't want to land hiddenComponents as-is though. Our concern is the shape, not the problem. A consumer-side denylist says "hide core's Dialog" when what you mean is "my Dialog replaces core's," so the relationship is lost and every consumer has to repeat the config. We think this belongs a layer up, probably the integration declaring the replacement in its own manifest, so discovery resolves to one provider and consumers inherit it for free.

That's a bigger design call than this PR should carry, so we opened #4107 to scope it and we'll ship a real fix in an upcoming version. Subscribe there to follow along, and please chime in since your use case is the motivating one.

Closing this in favor of #4107. Appreciate the report.

PR #3911, opened 2026-07-13
Fix theme artwork loading on the themes page

### LINKED ISSUE #3307: [Bug] Themes page fails to load example images with ERR_QUIC_PROTOCOL_ERROR
### Description

Each image fetch fails with a `net::ERR_QUIC_PROTOCOL_ERROR.QUIC_NETWORK_IDLE_TIMEOUT` error.

Reproducible on Chrome and Firefox. Safari works fine.

Image

### Reproduction

1. Visit https://astryx.atmeta.com/themes
2. Observe dev console, and network tab
3. Example images are missing on the UI

### XDS Version

N/A

### Environment

Chrome Version 149.0.7827.197

mturac:
Summary
- serve the /themes picker and showcase artwork from same-origin optimized WebP assets
- update theme showcase content and picker backgrounds to use /theme-assets
- extend the demo media check to prevent lookaside-hosted artwork from returning to the /themes surface

Fixes #3307

Validation
- pnpm build
- pnpm check:demo-media
- pnpm -F @astryxdesign/docsite typecheck
- pnpm check:repo
- local browser smoke: /themes?theme=matcha loaded 12 /theme-assets responses with no failed requests, no lookaside asset requests, and no broken images

Also checked
- pnpm -F @astryxdesign/docsite build compiled and completed TypeScript, then hit an existing prerender failure on component docs pages with `TypeError: (0 , d.jsxDEV) is not a function`


imdreamrunner:
Hi! Thanks for the contribution.

At the moment, artwork assets are managed through an internal asset manager and served via public CDN URLs, so we wouldn’t expect any cross-origin errors. Please let me know if you’re encountering one.

If everything is working correctly and this change is intended as a performance improvement, we’re currently exploring a different approach to managing image assets and will share an update once we have more details.

imdreamrunner (changes requested):
Just setting status as "Request changes" to get your attention on this PR. Please let me know if there is an error you got loading the themes page or this is a performance improvements. Thanks

mturac:
Thanks for checking.

This PR was meant to address the loading failure reported in #3307, where the /themes artwork requests were failing with `ERR_QUIC_PROTOCOL_ERROR` in Chrome/Firefox. I treated it as a reliability fix for that user-visible error rather than as a standalone performance improvement.

On this branch I verified that /themes?theme=matcha loads the same-origin theme assets without failed requests, without lookaside asset requests, and without broken images. If the internal asset manager/CDN path is the expected source of truth and the team is not seeing the reported failure, I am happy to close this PR or adjust the approach to whichever asset-management direction you prefer.

cixzhang:
Thanks for this, @mturac — you correctly diagnosed the CORS/lookaside problem on the themes page, and self-hosting the artwork as same-origin WebP is exactly the right direction. The demo-media check extension to catch lookaside-hosted artwork is a nice touch too.

Just so you're not working in the dark: we've got a broader effort in flight (#3973) to self-host all of the lookaside-hosted demo imagery under /public, not just the themes-page artwork. It's moving a bit slowly because there are some internal-integration checks we need to clear before it can land, but it's actively being worked. I wanted to flag it so this doesn't feel stalled and so we can make sure the two don't end up overlapping on the same assets.

Really appreciate you digging into this — I'll keep you posted as the broader change progresses.

PR #3913, opened 2026-07-13
fix(tablist): reserve a gap under divided tabs so the hover pill clears the underline

ernestt:
## Summary

When a `TabList` has `hasDivider`, the hover highlight filled the full tab height and butted directly against the underline, and an adjacent same-size `Button` sat flush on the rail. The strip only looked balanced when paired with a *smaller* button — an unintuitive workaround the docs had to call out.

This reserves **4px** between the tabs and the divider rail when `hasDivider` is set. The tabs keep their element-size height; the gap comes from `paddingBlockEnd` on the nav, and the selected indicator drops onto the rail via a shared `--_tab-indicator-bottom` custom property. **Non-divided tab lists are byte-for-byte unchanged.**

Result: a divided `md` tab strip pairs cleanly with a `md` button, the hover pill has breathing room above the rail, and the indicator still rests on the rail at every size.

## Before / After

Real component rendered with the real compiled `astryx.css`; hover shown on the middle tab.

TabList divider before/after

## The two problems

1. **Hover shape touches the underline** — the pill filled the tab and butted the rail (indicator at `bottom: -1px`, zero reserved space).
2. **Size mismatch with end content** — a same-size Button butted the rail, so the strip only looked right with a *smaller* button. The docs literally advised "use a smaller (sm) button," which is the smell this fixes.

Both fall out of one decision: the pill was the full row height with nothing reserved between it and the rail.

## Options explored

Several approaches were prototyped and rendered at pixel-faithful token geometry (md element = 32px, indicator = 2px, gap candidates 2/4px) before landing on the chosen one.

| Option | Idea | Verdict |
|---|---|---|
| **A — grow the row** *(chosen)* | Reserve padding below the tabs; pill and same-size button both float above the rail. | ✅ Fixes both problems; keeps intuitive `md`↔`md` pairing. Row grows only when a divider is shown. |
| B — shrink pill, keep row height | Anchor to total height, render the hover pill one size down. | ❌ Formalises the `md`-tab-↔-`sm`-button mismatch we're trying to remove. |
| C — inset only the pill bottom | Row stays token height, pill 2–4px shorter. | ❌ Fixes the pill but the adjacent same-size button still butts the rail; only relocates the bug. |
| D — separate rail track | Nav owns a full-width rail; indicator animates along it. | △ Best-looking but a much larger change and reopens the `size` vs `density` discussion; deferred. |
| E — drop `hasDivider`, divider via Toolbar only | Let the container own the rail. | ❌ Breaking public-API removal; rejected in favour of a non-breaking fix. |

### Gap size (2px vs 4px)

At **2px** the pill's rounded corner still crowds the underline (reads as "still touching"). At **6px** the underline detaches and reads as a separate element. **4px** (`--spacing-1`) is the sweet spot and a real token — chosen.

### Why "grow the row" is implemented as container padding, not a taller button

An early prototype grew the *button* to 36px and re-centred the label, which pushed the text 2px low (label centred in 36 while the pill centred in 32). The fix keeps the tab content a fixed-size box (text centred in it) and puts the 4px on the nav as `paddingBlockEnd`, with the indicator reaching the rail via `bottom: calc(-1 * (spacing-1 + border-width))`. This is the same pattern `Toolbar` already uses for tab indicators.

## Vibe test results

Verified against the **real compiled component** (`renderToStaticMarkup` of the built `dist/` + real `astryx.css`), with hard pixel measurements of the rendered geometry:

| Case | nav h | tab h | indicator bottom | rail (nav bottom) | tabTop − btnTop |
|---|---|---|---|---|---|
| No divider (regression) | 32 | 32 | −1px (hug) | — | **0** aligned |
| hasDivider `sm` | 33 | 28 | 33 | 33 | **0** aligned |
| hasDivider `md` | 37 | 32 | 37 | 37 | **0** aligned |
| hasDivider `lg` | 41 | 36 | 41 | 41 | **0** aligned |

- Indicator lands **exactly** on the rail at every size (33=33, 37=37, 41=41).
- Adjacent button shares the tab's top baseline (`tabTop − btnTop = 0`) — the earlier 2px drift is gone.
- No-divider path is unchanged (same class set, same geometry as `main`).
- Row grows exactly +5px when divided (4px pad + 1px border); unchanged otherwise.

## Edge cases handled

- **No divider** → offset var unset → `-1px` fallback → identical to `main`.
- **Carousel** (child tab indicators, the subject of the earlier 1px-bleed fix) → no divider, falls through to the `-1px` fallback → untouched.
- **Toolbar without a bottom divider** → previously set `--_tab-indicator-bottom` unconditionally (harmless while orphaned); now only sets it when `dividers` includes `bottom`, so a nested TabList's indicator keeps hugging the tab.
- **TabMenu** overflow trigger mirrors Tab's indicator, so its underline lands on the rail too.

## Changes

- `Tab` / `TabMenu`: indicator reads `bottom: var(--_tab-indicator-bottom, -1px)`.
- `TabList`: divider styling adds `paddingBlockEnd` + sets the offset var (gated by `hasDivider`).
- `Toolbar`: only sets the offset var when it owns a bottom divider.
- Docs/showcase: dropped the "use a smaller button" guidance; matched button size in the With Actions example.
- Storybook: new **Divider Gap** story across sm/md/lg with matched buttons.
- Tests: 3 new divider tests (45 total, all passing).

## Test plan

- `pnpm -F @astryxdesign/core build` — clean
- `pnpm -F @astryxdesign/core test` (TabList + Toolbar) — 67 passing
- `tsc --noEmit` — clean; `eslint` — clean
- `sync-exports --check`, `generate-token-docs --check` — in sync
- Visual verification against real compiled component (see table + image above)

### DIFF
--- a/.changeset/tablist-divider-indicator-gap.md
+++ b/.changeset/tablist-divider-indicator-gap.md
@@ -1,0 +1,9 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] TabList `hasDivider` reserves a gap so the hover pill and adjacent buttons no longer touch the underline
+
+A divided TabList now reserves 4px between the tabs and the divider rail. The hover highlight sits clear of the underline, and a same-size Button placed alongside the tabs aligns to the tab baseline instead of butting the rail — so a `md` tab strip pairs with a `md` button. The selected indicator still rests on the rail. Non-divided tab lists are unchanged. TabList inside a `Toolbar` with `dividers={['bottom']}` gets the same alignment via the toolbar's own spacing.
+
+@ernestt
--- a/apps/storybook/stories/TabList.stories.tsx
+++ b/apps/storybook/stories/TabList.stories.tsx
@@ -213,18 +213,64 @@ export const WithActions: Story = {
           <Button
             label="Filter"
             variant="ghost"
-            size="sm"
+            size="lg"
             icon={<FunnelIcon />}
             isIconOnly
           />
           <Button
             label="New item"
             variant="primary"
-            size="sm"
+            size="lg"
             icon={<PlusIcon />}
           />
         </div>
       </TabList>
+    );
+  },
+};
+
+export const DividerGap: Story = {
+  name: 'Divider Gap (sm / md / lg)',
+  render: () => {
+    const [value, setValue] = useState('overview');
+    return (
+      <div style={{display: 'flex', flexDirection: 'column', gap: '32px'}}>
+        {(['sm', 'md', 'lg'] as const).map(size => (
+          <div
+            key={size}
+            style={{display: 'flex', flexDirection: 'column', gap: '8px'}}>
+            <span style={{font: '600 12px system-ui', color: '#4E606F'}}>
+              size=&quot;{size}&quot; · hasDivider · matched Button size
+            </span>
+            <TabList value={value} onChange={setValue} size={size} hasDivider>
+              <Tab value="overview" label="Overview" />
+              <Tab value="activity" label="Activity" />
+              <Tab value="settings" label="Settings" />
+              <div
+                style={{
+                  marginInlineStart: 'auto',
+                  display: 'flex',
+                  alignItems: 'center',
+                  gap: '4px',
+                }}>
+                <Button
+                  label="Filter"
+                  variant="ghost"
+                  size={size}
+                  icon={<FunnelIcon />}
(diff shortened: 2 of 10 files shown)


cixzhang (approved):

kentonquatman:
Closing now and will align on design before reopening.

Issue #3231, opened 2026-06-28
[Bug] CORS Error for images on the main website


ShanHuangNet:
### Description

The deployed Templates page shows 

<img width="1913" height="849" alt="Image" src="https://github.com/user-attachments/assets/061fff7e-e37e-4fde-9518-627512cdf183" />

 for some template cards and logs multiple console errors.

On https://astryx.atmeta.com/templates with the `Tools` filter selected, several previews request images from `https://lookaside.facebook.com/assets/astryx/light-product-*.png`. The browser blocks those image fetches with CORS errors because the response does not include `Access-Control-Allow-Origin` for `https://astryx.atmeta.com`.

The same page also logs a React hydration error:

`Uncaught Error: Minified React error #418`

Expected: template preview images should load without CORS console errors, and the page should hydrate without React errors.

Screenshot:

### Reproduction

1. Open https://astryx.atmeta.com/templates in Chrome.
2. Select the `Tools` category filter.
3. Open DevTools Console.
4. Observe CORS failures for:
   - `https://lookaside.facebook.com/assets/astryx/light-product-1.png`
   - `https://lookaside.facebook.com/assets/astryx/light-product-4.png`
   - `https://lookaside.facebook.com/assets/astryx/light-product-5.png`
5. Observe React hydration error `Minified React error #418` on the same page.

### XDS Version

N/A

### Environment

Chrome on macOS.
Observed on 2026-06-28 at https://astryx.atmeta.com/templates.

cixzhang:
@josephfarina @rubyycheung seems like we do hit the CORS issue in some cases with the CDN. We may need to look into other asset management solutions for the docsite.

cixzhang:
Fixed by [#3973](https://github.com/facebook/astryx/pull/3973) (merged Aug 7) — @imdreamrunner self-hosted every `lookaside.facebook.com/assets/astryx/*` image in the repo, under `apps/docsite/public/template-assets/` (101 files) and `apps/docsite/public/images/` (69 files). The template previews now load same-origin, so there is no cross-origin request left to be blocked. `main` has no remaining `lookaside.facebook.com` reference outside a CHANGELOG entry.

Thanks for the report, @ShanHuangNet — it also surfaced the bigger problem behind it, that an external contributor could not see or add demo imagery at all. That is what [#3973](https://github.com/facebook/astryx/pull/3973) went and fixed.

[...]

Issue #3345, opened 2026-07-02
CodeBlock: no per-line add/remove (diff) backgrounds — highlightLines is single-accent only


thedjpetersen:
## Problem

Builders showing config/code changes need green/red line washes, and CodeBlock can't provide them. `highlightLines` paints every listed line with the single neutral accent (`--color-accent-muted`) — there is no way to mark a line as *added* or *removed*.

Evidence:

- An internal-tool builder (Meridian) hand-tinted lines with manual green/red styling because `highlightLines` only offers the neutral accent — bypassing the design system entirely for a presentation concern CodeBlock already half-owns.
- The ChatToolCalls diff-display template is an in-repo consumer that would use this immediately for tool-call before/after code display.

## API arbitration (compact round, 2026-07-01, naive sonnet generations + haiku probe)

| Candidate | Result |
| --- | --- |
| (a) `highlightLines` extended to `Array<number \| {line, type}>` | 0 invented props, 0 escapes, exact usage |
| (b) separate `diffLines: {added: number[], removed: number[]}` | 0 invented props, 0 escapes |
| (c) `language="diff"` auto-parse | Agent embedded `+`/`-` prefixes and `@@` hunk headers INTO the code string — conflates content with presentation; the copy button would copy the markers |

Probe against the **current** API: the agent invented `addedLines` / `removedLines` props, confirming the demand pull for per-line diff semantics.

**Verdict:** extend `highlightLines` — backward compatible, one prop owns line accents. `diffLines` is the runner-up (also clean, but splits line-accent ownership across two props).

## Proposed API

```tsx
interface CodeBlockProps {
  /**
   * 1-indexed lines to accent. Plain numbers (and type: 'highlight')
   * keep today's neutral accent; 'add' / 'remove' render green/red
   * diff washes.
   */
  highlightLines?: Array<
    number | {line: number; type?: 'add' | 'remove' | 'highlight'}
  >;
}
```

```tsx
<CodeBlock
  code={updatedConfig}
  language="json"
  hasLineNumbers
  highlightLines={[
    {line: 4, type: 'remove'},
    {line: 5, type: 'add'},
    12, // plain number — existing neutral accent, unchanged
  ]}
/>
```

Plain numbers keep today's neutral accent — fully backward compatible, no behavior change for existing callers.

## Theming note

[...]

PR #3915, opened 2026-07-13
fix(Thumbnail): only sample image luminance when the remove button renders

AKnassa:
## What this does

A thumbnail with a picture in it was quietly downloading that picture a second time, looking at its colours, and then throwing the answer away.

This stops it doing that, unless the answer is actually needed.

## Why

Thumbnails can show a small **✕ remove button** on top of the image. To make sure that button is readable against whatever is behind it, the component samples the image's brightness and picks a light or dark button to suit.

That sampling is the only reason the image gets re-downloaded — and **most thumbnails don't have a remove button at all.** So on those, the download, the decode and the colour-check are pure waste. Worse, when the image lives on another domain the browser blocks that download and prints a **CORS error in the console**, which is the noise reported in #3231.

## What changed

- The image is now only sampled when the remove button is actually on screen.
- No remove button → no download, no console error.
- Remove button present → behaves exactly as before, so the button stays readable.
- No change to how anything looks, and no change to the component's public API.

## How to see it

Open any page with plain thumbnails (for example the `payment-form` template) and watch the browser console — the CORS errors for the thumbnail images are gone. Thumbnails that *do* have a remove button still adapt their button colour to the image.

## Scope

**Refs #3231** — deliberately not "Fixes". This removes the pointless cross-origin request that generates the console errors, but the wider question of **where the docsite's images are hosted** is a separate call for the maintainers, and this PR doesn't touch it. `apps/docsite` is untouched; no CDN, proxy, or re-hosting is proposed.

`scripts/check-demo-media.mjs` (already on main) documents this exact "fetch-to-sample trips CORS" problem — this PR just stops issuing the fetch when nothing consumes it.

## Checks

- Full repo suite green: 341/341 files, 6,639 tests
- Lint, typecheck, prettier clean
- New tests cover: no remove button → zero fetches; remove button → still fetches; disabled; missing src; button appearing after mount; src swapped mid-life

### DIFF
--- a/.changeset/thumbnail-sample-only-with-remove-button.md
+++ b/.changeset/thumbnail-sample-only-with-remove-button.md
@@ -1,0 +1,8 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Thumbnail: only sample the image when the remove button renders (#3231)
+
+The APCA luminance sample (`fetch(src, {mode: 'cors'})` → `createImageBitmap` → `OffscreenCanvas`) ran on every image-backed thumbnail, but its only consumer is the contrast theming of the overlaid remove button. Thumbnails without `onRemove` — or with it suppressed by `isDisabled` — now skip the fetch entirely: no wasted request, and no CORS console error on cross-origin sources that were never sampled for anything.
+@AKnassa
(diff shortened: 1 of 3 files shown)


cixzhang:
Thanks for chasing this one down — the "download the image twice and throw the answer away" behavior you spotted was a real inefficiency. Good news: it's since been removed entirely. Thumbnail's remove button now sits on a fixed `--color-overlay` scrim with an `--color-on-dark` icon (landed in #3345 and follow-ups), so it no longer samples the image's luminance at all — which means there's no longer a sampling step to gate.

Because the code path this optimizes is gone from main, I'm going to close this as superseded. Genuinely appreciate the careful diagnosis here, and I hope you'll keep the PRs coming — this was exactly the right instinct. 🙏

Issue #2279, opened 2026-05-21
defineTheme color config: accent required for neutral-only, lacks light/dark support


cixzhang:
## Problem

The `color` field in `defineTheme` has a few ergonomic and type issues:

### 1. `accent` is required even when you only want `neutralStyle`

`XDSColorScaleConfig` defines `accent: string` as required. If you just want to set `neutralStyle: "warm"` without changing the accent color, you still have to pass an accent value:

```ts
// This should work but doesnt — accent is required
defineTheme({
  name: "my-theme",
  color: { neutralStyle: "warm" },
});
```

### 2. `accent` only accepts a single string — no light/dark tuple support

The rest of the token system supports `[light, dark]` tuples via `XDSTokenValue`, but `accent` in the color config is typed as just `string`. This prevents setting a different accent per color scheme:

```ts
// Not possible today:
color: { accent: ["#0064E0", "#48CAE4"] }

// Have to use this workaround in tokens instead:
tokens: { "--color-accent": ["#0064E0", "#48CAE4"] }
```

But that workaround skips the full color scale generation that `expandColorScale` provides (muted variants, on-accent, text-accent, icon-accent, etc).

### 3. Unclear light/dark behavior in the color extractor

`expandColorScale` already generates `light-dark()` values from a single accent (picking different tonal palette stops for light vs dark). But if a consumer provides both an accent via `color.accent` AND token overrides via `tokens["--color-accent"]` as a tuple, the interaction is unclear:

- Does the color extractor generate its full derived palette from the single accent, then get partially overridden by the tuple token?
- Will derived tokens (e.g. `--color-accent-muted`, `--color-on-accent`) still be based on the original single accent while `--color-accent` itself uses the tuple?

This creates potential inconsistency where the accent and its derived variants diverge.

## Suggested fixes

1. Make `accent` optional in `XDSColorScaleConfig` — allow neutral-only configuration
2. Accept `string | [light: string, dark: string]` for `accent` (matching `XDSTokenValue`)
3. When given a tuple, derive light palette from the light accent and dark palette from the dark accent
4. Document the precedence between `color` config and `tokens` overrides for accent-derived values

## References

- `packages/core/src/theme/expandColorScale.ts` — `XDSColorScaleConfig` type
- `packages/core/src/theme/defineTheme.ts` — color config integration

ernestt:
[...]

PR #3916, opened 2026-07-13
feat(theme): make color.accent optional in defineTheme

AKnassa:
## What this does

You can now build a theme that only restyles the **greys** — without having to pick a brand colour you don't want.

## Why

`defineTheme` insisted on an `accent` colour. So if all you wanted was warmer greys, or higher contrast, you still had to supply an accent — and whatever you picked would silently become your theme's brand colour.

This is the maintainer's own request in #2279. It implements **only suggested fix #1** from that issue.

## What changed

- `color.accent` is now optional.
- Leave it out and you still get the full grey ramp (`neutralStyle`, `contrast` all work) — the greys borrow their hue from the default accent.
- The three accent tokens (`--color-accent`, `--color-accent-muted`, `--color-on-accent`) are simply **not generated**, so they fall back to the default values. This is the same fall-through the file already uses for status, categorical and on-dark tokens.
- If you *do* pass an accent, nothing changes — output is identical, token for token.

```ts
// now valid — warm greys, default accent left alone
defineTheme({name: 'warm', color: {neutralStyle: 'warm'}})
```

## One thing worth a close look 👀

The obvious-looking implementation would be *"if there's no accent, just default it to the standard accent hex."* **That is wrong, and it would quietly re-colour every existing theme.**

The accent colour you get by *deriving a scale from* `#0064E0` is **not** the same value as the default accent *token*:

| | value |
|---|---|
| `expandColorScale({accent: '#0064E0'})['--color-accent']` | `light-dark(#0058D2, #BBC2FF)` |
| `colorDefaults['--color-accent']` | `light-dark(#0064E0, #2694FE)` |

So the accent tokens are **omitted, not defaulted** — omitting them is what preserves the existing default. There's a test pinning both values, so if anyone later "simplifies" this into the seeded version, it fails loudly instead of silently changing everyone's brand colour.

## Scope

Only suggested fix #1. Deliberately **not** implementing #2 (tuple/per-scheme accents) or #3 — those touch accent-picking semantics and belong with the in-flight Color Studio work.

## Checks

- Core + CLI suites green; lint, typecheck, prettier clean
- Tests cover: neutral-only themes end-to-end through `useTheme()` in light *and* dark; an empty config; an empty-string accent (found and fixed a real bug here — the seed used a nullish check while the token output used a truthy one, so `accent: ''` seeded from a malformed value *and* dropped the accent tokens); a neutral-only theme with a hand-picked accent override; and a drift guard tying the seed to the default token

### DIFF
--- a/.changeset/define-theme-optional-accent.md
+++ b/.changeset/define-theme-optional-accent.md
@@ -1,0 +1,9 @@
+---
+'@astryxdesign/core': patch
+'@astryxdesign/cli': patch
+---
+
+[feat] `defineTheme`: make `color.accent` optional (#2279)
+
+A theme can now restyle the neutral ramp (`neutralStyle`, `contrast`) without adopting an accent. An accent-less config seeds the neutral palettes from the default accent's hue but leaves `--color-accent`, `--color-accent-muted` and `--color-on-accent` ungenerated, so they fall through to the token defaults — the same fall-through `expandColorScale` already applies to status, categorical and on-dark tokens. Configs that pass an accent are unchanged, token for token.
+@AKnassa
--- a/packages/cli/docs/theme.doc.mjs
+++ b/packages/cli/docs/theme.doc.mjs
@@ -191,7 +191,7 @@ const myTheme = defineTheme({
             [
               'color',
               '--color-accent, --color-background-*, --color-text-*, --color-border, etc.',
-              'accent (hex), neutralStyle? (warm|cool|neutral), contrast? (standard|high)',
+              'accent? (hex — omit for neutral-only), neutralStyle? (warm|cool|neutral), contrast? (standard|high)',
             ],
             [
               'typography.scale',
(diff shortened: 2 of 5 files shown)


cixzhang (approved):
This is a really clean fill for the accent-required gap in #2279 (and the friction called out in the adoption feedback). Making `color.accent` optional so a theme can restyle just the neutral ramp — `neutralStyle`, `contrast` — without being forced to adopt a brand color is exactly the ergonomic win the issue asked for.

What makes this mergeable rather than merely plausible is that you got the backward-compatibility question exactly right: when `accent` is omitted, the three accent tokens are **omitted, not seeded** — so they fall through to `colorDefaults` (the same fall-through the file already uses for status/categorical/on-dark tokens), and every theme that *does* pass an accent stays byte-identical, token for token. The subtle-but-critical detail — that deriving a scale from `#0064E0` produces a *different* accent than the default token holds, so "just default the seed" would silently re-color existing themes — is called out and pinned by a drift-guard test. The suite covers omit-vs-seed, empty config, the `accent: ''` nullish-vs-truthy edge, runtime resolution through `useTheme()` in both modes, and an explicit-accent override. Nicely scoped, too: it deliberately stays out of the tuple/per-scheme accent and inversion territory.

Verified locally on a fresh worktree: the `expandColorScale` suite (17 tests) and `tsc` both pass, and the accent-present branch is unchanged from `main`. Thanks for the thorough tests and the clear writeup.

PR #3570, opened 2026-07-04
feat(charts): bootstrap @astryxdesign/charts package (canary-only)

cixzhang:
## What

Bootstraps a new `@astryxdesign/charts` package — an empty, ready-to-build scaffold for chart components. Chart components will be thin, themeable wrappers over a peer charting engine: the package owns the design-system surface while leaving the rendering engine as a peer dependency the consumer manages.

**This is a bootstrap scaffold only — no components yet.** It's set up so components can be added under `src/` and exported from `src/index.ts`.

## Why canary-only, no stable version

Charts should be publishable for early testing but not carry a stable release yet. It follows the same canary-only model as `lab`:

- `"private": true` — npm's hard guarantee against an accidental stable publish.
- `"astryx": { "canaryOnly": true }` — the release workflow's stable (`latest`) job skips both private and `canaryOnly` packages; the canary job strips `private` in its ephemeral CI checkout only (never in git) to publish under the `@canary` dist-tag.

So it publishes on every push to `main` as `0.x.y-canary.<sha>` and never as `latest`.

## What's in the scaffold

- `package.json` — canary-only config, ESM + type-declaration build.
- Babel + StyleX setup mirroring `lab`, so a first component can consume `@astryxdesign/core` theme tokens directly.
- `tsconfig` / `tsconfig.build.json`, empty `src/index.ts` barrel, README, CHANGELOG.
- Wired into the root `pnpm build` (after `core`) and the package docs.

The pre-compiled stylesheet wiring (a `charts` target in `scripts/build-css.mjs`, a `build:css` script, and a `./charts.css` export) is **deferred** until the first component adds StyleX styles — an empty CSS extraction would otherwise fail with no rules to collect. The README notes exactly how to re-add it (same as `lab`).

## Verification

- `pnpm -F @astryxdesign/charts build` — ESM output + `.d.ts` declarations generated.
- `pnpm build` (full repo) — green.
- `pnpm check:repo` — green.

## Notes

- Relates to #977. That issue's later update leaned toward putting charts in `@astryxdesign/core`; this PR instead stands charts up as its own canary-only package per direction. Flagging for maintainer confirmation on the package-vs-core decision.

Draft — bootstrap scaffold for review of the package shape and publishing model before the component surface is built out.

PR #3917, opened 2026-07-13
fix(lab,charts): portable build scripts + a guard so they stay portable

AKnassa:
## What this does

Fixes the build on Windows for the packages that were missed — and adds a check so it can't quietly break again.

## Why

Core was fixed for Windows back in #3305, but the packages around it keep regressing: `lab` kept the old build script, `charts` was created *after* that fix as a copy of the broken version (#3570), and while this PR was open, `richtext` landed as a fresh copy of the same broken script. Three regressions of one already-fixed bug — the pattern this PR exists to stop.

There are two failures here, and **the quiet one is the dangerous one**:

1. `rm -rf` — Windows has no `rm`, so the clean step fails. Loud, obvious, annoying.
2. **The single quotes.** Windows treats `'.ts,.tsx'` as a *literal string, quotes included*. So Babel's file patterns match **nothing at all** — and Babel is perfectly happy about that. It **reports success and leaves an empty `dist/`.**

A build that fails tells you it failed. A build that succeeds and produces nothing sends you hunting for a bug that isn't there.

## What changed

**The fix.** The same change core already shipped, applied to the stragglers: `rm -rf` → `rimraf` in `lab`, `charts`, and `richtext`, plus the single→double quote fix in `richtext`. (Main has since adopted the quote fix for `lab` and `charts` independently, so only the `rimraf` half remains in this diff for those two.)

**The guard.** A new `check:portable-scripts` that reads every `package.json` script and flags Unix-only commands and single-quoted arguments. It runs inside the existing `check:repo` chain, which CI already runs on every PR.

The guard is the point, and it has already earned its keep once: `richtext` was added to main while this PR sat open, carrying `rm -rf` and single-quoted globs — the guard flagged it during conflict resolution and the fix is included here. It checks 160 scripts today.

If you'd rather not take the guard, dropping `scripts/check-portable-scripts.*` and the root `package.json` wiring leaves the fix standing on its own.

## How to see it

On Windows, `pnpm build` from the repo root now completes through `lab`, `charts`, and `richtext` with populated `dist/` folders. On any platform, `pnpm check:portable-scripts` passes — and fails with a clear message if someone reintroduces the pattern.

## Notes

- **No changeset**, and one isn't possible: `lab`, `charts`, and `richtext` are all `private: true`, and `check:changesets` rejects a changeset naming a private package — adding one would turn `pnpm lint` red. The equivalent core fix (`bbd7a59`) shipped without one for the same reason.
- **No `windows-latest` CI runner here.** That's a cost/flake decision for the maintainers, so it's left out deliberately — happy to follow up if you want it.


cixzhang (approved):

cixzhang:
Thanks for helping this repo stay windows friendly! Some merge conflicts before we can merge.

AKnassa:
Conflicts resolved. Main already adopted the quoting fixes, so this is now rimraf plus the guard; the guard caught the new richtext package using rm -rf and single quotes, fixed here. Note core typecheck is red on main itself (useTableSelection.test.tsx:227, styles vs xstyle), unrelated.

cixzhang:
Merged `main` into this branch for you and armed auto-merge (squash) — it will land on its own once CI is green.

Resolution notes, since `main` moved a fair bit under this PR:
- `packages/richtext` did not exist when you opened this; `RichTextEditor` was promoted there from lab, carrying both offences (`rm -rf dist` and single-quoted babel globs). I applied this PR's fixes to it and declared `rimraf` there, so the new package is covered rather than grandfathered in.
- `check:repo` gained `check:cli-structure` and `check:use-client` on main; kept those and appended `check:portable-scripts`.
- `packages/lab` lost its lexical peer deps to the promotion; took main's shape and kept the `rimraf` devDependency.
- Verified locally: guard passes on all 287 scripts, its 17 tests pass, and `@astryxdesign/richtext` builds to a non-empty dist (28 files) — which is the silent-failure case this guards.

Nothing needed from you. Thanks for the patience on this one.

Issue #3923, opened 2026-07-14
[Bug] astryx init silently accepts an invalid --agent value


harjothkhara:
### Description

`astryx init --features agents` validates `--features` against its allow-list (and rejects unknown values with `Unknown features: … Valid features: …`), but `--agent` is never validated. An unrecognized value is silently accepted and falls through to creating `AGENTS.md`, so a typo like `--agent claud` produces the wrong output with a success exit code instead of an error.

The valid-agent list and an `ERR_UNKNOWN_AGENT` code already exist in the CLI (used by the internal agent-docs handler) — they just aren't applied on the `init` path.

### Reproduction

```
mkdir demo && cd demo && npm init -y
npx astryx init --features agents --agent bogus
# → "✓ AI agent docs installed → AGENTS.md", exit 0
```

### Expected behavior

An error consistent with the existing `--features` validation — e.g. `Unknown agent "bogus". Valid: claude, cursor, codex, hermes, all` — a non-zero exit code, and no file written.

### Actual behavior

`--agent bogus` is accepted silently: `AGENTS.md` is created and the command exits 0.

### Environment

- @astryxdesign/cli 0.1.5 (reproduced on current `main`)

cixzhang:
Thanks for catching this silent-invalid-input path. This is fixed on current `main` by [#4511](https://github.com/facebook/astryx/pull/4511): `astryx init` now rejects unknown `--agent` values with a clear valid-agent error, with API and CLI regression coverage. Closing as fixed.

PR #3924, opened 2026-07-14
fix(cli): validate --agent in astryx init

harjothkhara:
## Summary

`astryx init --features agents` validates `--features` against its allow-list (rejecting unknown values with `Unknown features: … Valid features: …`) but never validated `--agent`. An unrecognized value silently fell through `resolveAgentPaths`'s unknown-preset branch and created `AGENTS.md` with a success exit code, so a typo like `--agent claud` produced the wrong output instead of an error.

The valid-agent list and an `ERR_UNKNOWN_AGENT` code already existed in the CLI (used by the internal agent-docs handler); they just weren't applied on the `init` path. This exports the existing `VALID_AGENTS` constant from `agent-docs.mjs` and adds a guard in `init.mjs` mirroring the existing `--features` check. Valid agents (`claude`/`cursor`/`codex`/`hermes`/`all`) are unchanged. Surface: CLI only. Refs #3923.

## Verification

```
# New test (subprocess harness mirroring interactive-guard.test.mjs)
pnpm vitest run packages/cli/src/commands/init.agent-validation.test.mjs
  → bogus --agent → exit 1, no file written; valid --agent hermes → exit 0

# Suite + regression
pnpm vitest run packages/cli/src/commands/init.next-steps.test.mjs \
                packages/cli/src/commands/interactive-guard.test.mjs \
                packages/cli/src/commands/agent-docs.test.mjs
  → 62/62 passed
```
(This adds the first test coverage of `init`'s non-interactive validation-error path.)

## Real behavior proof

**Behavior addressed:** `astryx init --features agents --agent <invalid>` now errors and writes nothing, instead of silently creating `AGENTS.md` and exiting 0.

**Real environment tested:** macOS, Node v24.16.0, branch built from source, CLI driven through the real entry point (`packages/cli/bin/astryx.mjs init`), no mocks. "Before" captured by stashing the patch so the working tree equals current `main`.

**Exact steps or command run after this patch:**
```
mkdir demo && cd demo
echo '{"name":"demo","version":"1.0.0"}' > package.json
node .../packages/cli/bin/astryx.mjs init --features agents --agent bogus
```

**Evidence after fix** (identical command, `main` vs this branch):
```
BEFORE (main):   exit 0  → "✓ AI agent docs installed → AGENTS.md"   (AGENTS.md created)
AFTER  (branch): exit 1  → Error: Unknown agent "bogus".
                            Valid: claude, cursor, codex, hermes, all   (no file created)

sanity — valid value still works:
AFTER  (branch): node … init --features agents --agent hermes
                 → exit 0, "✓ AI agent docs installed → AGENTS.md"
```

**Observed result after fix:** An unknown `--agent` value prints an actionable error (matching the wording style of the existing `--features` error), exits non-zero, and creates no file. Valid agents are unaffected.

**What was not tested:** No change to what any valid `--agent` writes (only the invalid path is newly guarded). The unrelated fact that `astryx init` has no `--json` output mode was left as-is. Validation fires on the non-interactive (`--features`/`--all`) path where `--agent` is consumed; the interactive wizard does not take `--agent`.

### DIFF
--- a/.changeset/cli-init-validate-agent.md
+++ b/.changeset/cli-init-validate-agent.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/cli': patch
+---
+
+[fix] `astryx init` now validates `--agent`. An unrecognized value (e.g. a typo like `--agent claud`) previously fell through and silently created `AGENTS.md` with a success exit code; it now errors with `Unknown agent "…". Valid: claude, cursor, codex, hermes, all` and a non-zero exit, matching the existing `--features` validation. Valid agents are unchanged.
+
+@harjothkhara
--- a/packages/cli/src/commands/agent-docs.mjs
+++ b/packages/cli/src/commands/agent-docs.mjs
@@ -470,7 +470,7 @@ export function installAgentDocs(targetDir, {zh = false, lang, agent, paths, onl
   return written;
 }
 
-const VALID_AGENTS = ['claude', 'cursor', 'codex', 'hermes', 'all'];
+export const VALID_AGENTS = ['claude', 'cursor', 'codex', 'hermes', 'all'];
 
 export function registerAgentDocs(program) {
   program
--- a/packages/cli/src/commands/init.agent-validation.test.mjs
+++ b/packages/cli/src/commands/init.agent-validation.test.mjs
@@ -1,0 +1,66 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+/**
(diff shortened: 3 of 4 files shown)


harjothkhara:
Closing this. The same validation landed on main in #4511 (57611be4). `api/init/run/run.mjs` now rejects unknown `--agent` values with `ERR_UNKNOWN_AGENT`, and `init.test.mjs` covers it. Thanks.

PR #3925, opened 2026-07-14
fix(core): align CodeBlock line numbers with wrapped lines

zeroryu:
The line-number gutter rendered one fixed-height div per logical line in a separate flex column. With isWrapped, a single logical line wraps across multiple visual rows, but each gutter number still occupied exactly one line-height — so after the first wrapped line every number drifted up out of alignment.

Render each number as a ::before (content: attr(data-line)) inside a two-column grid line ([number] [code]). The number and its (wrappable) code share one grid row, so the row grows to fit the wrapped code and the number stays pinned to the row's first visual line (align-self: start).

The single <code> element is preserved (the CSS Custom Highlight range walker requires the line's text node as firstChild), and the full-height gutter divider moves to a <code>::after so it spans the code block's block padding.

Screenshot 2026-07-13 at 10 27 07 PM

### DIFF
--- a/.changeset/codeblock-wrapped-line-numbers.md
+++ b/.changeset/codeblock-wrapped-line-numbers.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] CodeBlock keeps line numbers aligned with wrapped lines when `isWrapped` is enabled
+@zeroryu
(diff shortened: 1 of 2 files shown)


imdreamrunner:
Thanks for the contribution! I am reviewing the implementation of this PR.

Btw, there's a failing test caught by the CI

> packages/core/src/theme/derivedVarRegistry.test.ts > component CSS vars are documented and themeable > CodeBlock: all source vars are in doc file

imdreamrunner (changes requested):
The approach looks good to me! Rendering the line number through a pseudo-element, and sizing the gutter by computing a ch-based width from the digit count, is smart.

Only request is to fix the failing test.

imdreamrunner (approved):

PR #3928, opened 2026-07-14
fix(Banner): honor prefers-reduced-motion on chevron transition

cixzhang:
The Banner's collapse/expand chevron uses a CSS transition on `transform` but did not disable it under `prefers-reduced-motion: reduce`. Users who prefer reduced motion still saw the rotation animation.

Apply the same conditional `transitionDuration` pattern used by Button and other components:
- Default: uses `--duration-fast`
- `prefers-reduced-motion: reduce`: `0s` (instant)

Night Watch - Component Auditor

### DIFF
--- a/.changeset/banner-reduced-motion.md
+++ b/.changeset/banner-reduced-motion.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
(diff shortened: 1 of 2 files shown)


thedjpetersen (approved):
Approved.

josephfarina (approved):

PR #3930, opened 2026-07-14
fix(build): compile dist with the production JSX transform

fatwang2:
## Summary

`@astryxdesign/core@0.1.5` — the current `latest` on npm — ships a `dist` compiled with the **development** JSX transform. **193 of its 479 `dist` files** import `react/jsx-dev-runtime` and call `jsxDEV`, which React's production build does not export (`exports.jsxDEV = void 0`).

Any consumer that renders an Astryx component in production crashes:

```
TypeError: (0 , b.jsxDEV) is not a function
```

`dist/AppShell/AppShell.js` alone contains 29 `jsxDEV` calls, so a Next.js app that puts `<AppShell>` in a layout returns a 500 on **every route**. That is what happened to us in production — every page of our dashboard went down until we pinned back to 0.1.4:

```
GET /        500  TypeError: (0 , b.jsxDEV) is not a function
GET /sign-in 500  TypeError: (0 , j.jsxDEV) is not a function
```

The bug is invisible locally (dev builds load React's development runtime, where `jsxDEV` exists) and invisible at build time (bundlers happily bundle the import) — it only surfaces when a production render actually executes.

| version | `jsxDEV` occurrences in `dist` |
| --- | --- |
| 0.1.3 | 0 |
| 0.1.4 | 0 |
| **0.1.5** | **5696** |

The UMD bundle (`unpkg`/`jsdelivr`) is affected too.

## Root cause

`@babel/preset-react` 8 changed the default for `development` from a hard-coded `false` to:

```js
development = api.env(env => env === 'development'),
```

Babel's env name is `BABEL_ENV || NODE_ENV || "development"` — so with `NODE_ENV` unset, which is the normal state of a library build, `development` becomes `true` and every JSX call compiles to `jsxDEV`.

#3839 bumped `@babel/preset-react` 7.29.7 → 8.0.1 on Jul 11. No config changed; the meaning of the existing config did. 0.1.4 was cut before that bump, 0.1.5 after.

Minimal reproduction (the repo's own preset options, `runtime: 'automatic'` with no `development`):

```
preset-react 8.0.1, NODE_ENV unset       →  import { jsxDEV } from "react/jsx-dev-runtime"   ❌
preset-react 8.0.1, NODE_ENV=production  →  import { jsx }    from "react/jsx-runtime"       ✅
preset-react 8.0.1, development: false   →  import { jsx }    from "react/jsx-runtime"       ✅
```

Building `main` today reproduces the broken `dist` — this is not a one-off CI fluke.

## Fix

1. **Pin `development: false`** in the `@babel/preset-react` options for `core`, `lab`, and `charts`. A published artifact should never depend on an ambient env var; `NODE_ENV=production` in the release script would also work, but it leaves the same trap armed for anyone who builds by hand.

2. **Add `scripts/check-no-dev-jsx.mjs`** (modeled on `check-use-client.mjs`): scans the built `dist` for `react/jsx-dev-runtime` / `jsxDEV` and fails the build. Wired into the `build` script of all three packages, and since `prepack` runs `build`, a dev-JSX `dist` can no longer be published.

`lab` and `charts` are private today, but they carry the same config and would ship the same broken artifact the day they are published — fixed here as well.

## Verification

| check | result |
| --- | --- |
| `pnpm -F @astryxdesign/core build` **on current `main`'s config** | ❌ guard fails, exit 1 — 193 dist files flagged |
| `pnpm -F @astryxdesign/core build` **with this fix** | ✅ 479 dist files, 0 dev JSX; `AppShell.js` back to `react/jsx-runtime` |
| `pnpm -F @astryxdesign/lab build` | ✅ 107 files clean |
| `pnpm -F @astryxdesign/charts build` | ✅ 31 files clean |
| `pnpm check:repo`, `eslint` | ✅ pass |

## Suggested follow-up

0.1.5 is still `latest` on npm and bricks every SSR consumer, so a patch release (or a deprecation of 0.1.5) seems worth doing quickly. Happy to adjust anything here.

### DIFF
--- a/.changeset/plain-hoops-repeat.md
+++ b/.changeset/plain-hoops-repeat.md
@@ -1,0 +1,18 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Compile dist with the production JSX transform — 0.1.5 shipped `jsxDEV`, which crashes every consumer that renders in production
+@fatwang2
+
+`@babel/preset-react` 8 derives its `development` option from the Babel env name
+(`api.env(env => env === 'development')`), and Babel falls back to
+`"development"` whenever `NODE_ENV`/`BABEL_ENV` is unset. The bump from 7.29.7 to
+8.0.1 therefore flipped the published build to the development JSX transform
+without any config change: 193 of 479 files in `@astryxdesign/core@0.1.5`'s
+`dist` import `react/jsx-dev-runtime` and call `jsxDEV`, which React's
+production build does not export.
+
+The Babel configs for core, lab, and charts now pin `development: false`, and a
+new `scripts/check-no-dev-jsx.mjs` guard fails the build if development JSX ever
+reaches `dist` again.
--- a/packages/charts/babel.config.js
+++ b/packages/charts/babel.config.js
@@ -23,7 +23,7 @@ const rootDir = path.resolve(__dirname, '../..');
 
 module.exports = {
   presets: [
-    ['@babel/preset-react', {runtime: 'automatic'}],
+    ['@babel/preset-react', {runtime: 'automatic', development: false}],
     ['@babel/preset-typescript', {isTSX: true, allExtensions: true}],
   ],
   plugins: [
--- a/packages/charts/package.json
+++ b/packages/charts/package.json
@@ -46,7 +46,7 @@
   },
   "scripts": {
     "prepack": "pnpm build",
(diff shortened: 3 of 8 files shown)


ejhammond (approved):

ejhammond:
Thanks for the fix! 0.1.6 should be available now with your code :)

PR #3744, opened 2026-07-10
fix(docsite): add ContextMenu playground defaults and ContextMenuItem showcase/examples

jiunshinn:
## Summary

Fills the docsite gaps for the ContextMenu component family: playground (property editor) defaults for both components, and a showcase + example block for ContextMenuItem.

## Case

- The ContextMenu properties playground had no `playground.defaults`, so the interactive preview rendered nothing useful out of the box — `children` and `items` are required props with no sensible auto-generated values.
- ContextMenuItem had no showcase (no hero visual) and no example blocks, so its docsite page had no viewable/copyable code — unlike its sibling DropdownMenuItem, which has both.
- ContextMenuItem also had no `playground.defaults`, so its playground rendered an empty item.

## Fix

- Added `playground.defaults` to `ContextMenu.doc.mjs`: a Card trigger ("Right-click this area") plus a sample `items` array (Edit, Duplicate, divider, Delete), following the pattern from the Carousel defaults.
- Added `playground.defaults` to `ContextMenuItem.doc.mjs` (`label` + `description`). No `wrapper` is needed — the item's menu context hook returns null standalone and falls back to `md` sizing.
- Added `ContextMenuItemBasic` and `ContextMenuItemShowcase` template block pairs (`.tsx` + `.doc.mjs`) under `packages/cli/templates/blocks/components/ContextMenuItem/`, mirroring the DropdownMenuItem blocks and using ContextMenu's compound mode (`menuContent`). Trigger area colors use design tokens (`var(--color-border)`, `var(--color-text-secondary)`).

## Testing

- `pnpm generate` in apps/docsite: ContextMenuItem now appears in `showcaseRegistry`/`exampleRegistry`, and both components have non-null `playground` in `componentRegistry`.
- Verified in the local docsite: ContextMenu playground renders the trigger and opens the Edit/Duplicate/divider/Delete menu on right-click; the ContextMenuItem page renders the showcase hero, the Examples section, and playground defaults.
- docsite vitest suite passes (217 tests, including the registry-completeness check); ContextMenu core tests and CLI component tests pass.
- `astryx component ContextMenuItem` now lists both new block templates.

Closes #2714
Closes #2712

PR #3481, opened 2026-07-03
docs: add example blocks for 46 components with no viewable code (#2759)

AKnassa:
## What this does

Adds a copyable code example to 46 component pages on the docsite that previously showed only a hero visual and an empty Examples section.

Fixes the bulk of #2759.

## Why

Every component page has a showcase (the hero at the top), but the Examples section is built separately from example blocks — and 46 components had none, so there was no code to view or copy anywhere on their page.

## What changed

- One new example block (`.tsx` + `.doc.mjs` pair, `type: 'block'`, no `isShowcase`) for each of the 46 components, placed next to each component's existing showcase and following the established pattern (e.g. `Avatar/AvatarWithImage`).
- Examples were written against each component's actual API and reviewed: no deprecated props (e.g. TopNavHeading uses `headingHref`), `componentsUsed` cross-links match what each example really renders, and examples demonstrate something beyond their showcase where it matters (MobileNavToggle custom trigger, Resizable collapsible + snap points, CommandPaletteFooter default keyboard hints).
- No product code touched — the diff is purely additive template files.

## Scope notes

- **Blockquote** is intentionally left to #3385 and **Lightbox** to #3301 (both open). Together with this PR, every component with a showcase has at least one viewable example, which closes #2759.
- Includes **CommandPaletteEmpty**, a gap that appeared after the issue was filed.

## How to see it

Run `pnpm dev` in `apps/docsite`, then open any affected page, e.g. `/components/ButtonGroup`, `/components/Resizable`, or `/components/MobileNavToggle` — each now has an Examples section with viewable, copyable code. On `main`, those sections are empty.

Verified: docsite data generation, 210/210 docsite tests, `tsc --noEmit`, and a full production build all pass.

https://claude.ai/code/session_018TamXn2FajoSU6gXqts7hk

PR #3932, opened 2026-07-14
docs(empty-state): clarify headingLevel is semantic-only

### LINKED ISSUE #2754: EmptyState — confirm `headingLevel` behavior (does not change heading size)
**Component:** `XDSEmptyState`
**Type:** Bug / clarification — prop behavior vs. expectation

### Description

In the docsite props tab, changing `headingLevel` on `XDSEmptyState` produces **no visible change** to the heading size. We need to confirm whether this is intended.

Looking at `packages/core/src/EmptyState/XDSEmptyState.tsx`, `headingLevel` only controls the **rendered HTML tag**:

```tsx
const HeadingTag = `h${headingLevel}` as const;
```

The title's visual style is hardcoded regardless of level — `styles.title` always sets `fontSize: --text-large-size` (and `--text-label-size` when `isCompact`). So `headingLevel` changes the semantic tag (`h1`–`h6`) for the document outline but never the visual size. The prop doc even says so: *"Controls the rendered HTML heading tag (h1–h6) to fit the document outline."*

So the prop **is** working as currently implemented/documented — it's a semantics-only control, not a sizing control.

### Decision needed

The mismatch is between expectation (changing `headingLevel` should resize the heading) and the documented semantics-only behavior. Pick one:

1. **Keep semantics-only, but make it clearer.** The visual control is intentional separation of concerns. If so, consider clarifying the props-tab/description copy so it's obvious this only affects the HTML tag, not the size, and avoid implying a visual change.
2. **Make `headingLevel` (or a new prop) also scale the visual size.** If consumers expect the heading to look like an h1/h2/etc., map level → type scale token so the size tracks the level (or add a separate visual `size`/`titleType` prop).

No code change is needed if we go with option 1 beyond doc clarity; option 2 requires wiring the level to the title font-size tokens.

---
_Split out from the tracking issue #2631 (item 3 of 17)._

jiunshinn:
## Summary
Clarifies in the docs that `EmptyState`'s `headingLevel` changes only the rendered HTML heading tag for the document outline and does **not** change the title's visual size.

Closes #2754.

## Context
In #2754 (labels: bug, question), the reporter found that `headingLevel` renders `h1`–`h6` but never changes the visual heading size. @cixzhang confirmed (2026-06-28) this is intended: it mirrors `Heading`'s semantic-only convention, and any size control would need to follow text-component prop naming conventions. So this is a docs clarification, not a behavior change.

The visual size is hardcoded independently of `headingLevel`:
- `packages/core/src/EmptyState/EmptyState.tsx:153` builds the tag as `h${headingLevel}` via `createElement`
- `EmptyState.tsx:48` sets `styles.title.fontSize` to `--text-large-size`; `:54` `titleCompact` to `--text-label-size` — neither varies with `headingLevel`

## Changes
Updated the `headingLevel` description in all locales/variants to state it is a semantic/accessibility change only, matching how `Heading`'s `accessibilityLevel` already separates document outline from visual style (`packages/core/src/Text/Heading.doc.mjs:31`):
- `EmptyState.tsx` `headingLevel` JSDoc
- `EmptyState.doc.mjs` en, zh, and dense `propDescriptions`

## Notes
- Docs-only; no changeset (per merged precedents #3608, #3481, #3744).
- No em-dashes in `.doc.mjs` prose (repo convention, #3651/#3260).

## Test plan
- `pnpm exec vitest run packages/core/src/EmptyState` → 16/16 pass
- `astryx component EmptyState --dense` renders the new wording
- Pre-commit hooks green (prettier/eslint, check:sync, check:changesets)

### DIFF
--- a/packages/core/src/EmptyState/EmptyState.doc.mjs
+++ b/packages/core/src/EmptyState/EmptyState.doc.mjs
@@ -39,7 +39,7 @@ export const docs = {
       name: 'headingLevel',
       type: '1 | 2 | 3 | 4 | 5 | 6',
       description:
-        'Controls the rendered HTML heading tag (h1-h6) to fit the document outline.',
+        'Controls only the rendered HTML heading tag (h1-h6) so the title fits the document outline. This is a semantic change for accessibility and does not change the visual size of the title, which stays fixed regardless of level.',
       default: '3',
     },
     {
@@ -121,7 +121,7 @@ export const docsZh = {
       name: 'headingLevel',
       type: '1 | 2 | 3 | 4 | 5 | 6',
       description:
-        '控制渲染的 HTML 标题标签（h1-h6），以适配文档大纲。',
+        '仅控制渲染的 HTML 标题标签（h1-h6），使标题适配文档大纲。这是用于无障碍的语义变化，不会改变标题的视觉大小，标题大小始终固定，与级别无关。',
       default: '3',
     },
     {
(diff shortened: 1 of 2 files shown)


imdreamrunner (approved):
lgtm!

PR #3800, opened 2026-07-11
fix(chat): position the first fill instantly + scrollToBottom({behavior: 'instant'}) (#3795)

yyq1025:
## Summary

Fixes #3795.

Opening a chat that already has content (history, replay, session switching) spring-flies from `scrollTop: 0` across the whole transcript. Root cause confirmed on #3795: the layout's ResizeObserver fires its callback synchronously on registration, so the follow spring starts before the hook's rAF-deferred initial jump runs — async fills fall into the streaming path. Per the arbitration there, this is now a pure behavior fix with no new component surface:

- **The first fill positions instantly by default**, whether the content is present at mount or arrives asynchronously. A pending flag is consumed by the first `scrollIfLocked` that sees scrollable content, mirroring use-stick-to-bottom's `previousHeight ? resize : initial`. Subsequent growth (streaming) springs exactly as today.
- **`scrollToBottom({behavior?: 'instant' | 'spring'})`** — one-frame programmatic jump that cancels any in-flight spring and zeroes velocity, so a stale animation tick can't fight the assignment. Shape matches `scrollToMessage`'s `behavior: 'instant'` in the same file and the DOM's `scrollTo`/`scrollIntoView`. Default `'spring'` is byte-identical to today.

The earlier `initial` prop (`'spring' | 'instant' | false`) is dropped per review: instant first-fill is unconditionally correct and needs no opt-in, and the `false`/start-at-top value is the scroll side of #2572's `align`/`anchor` axis — deferred to that arbitration. `ChatLayout.tsx` and `ChatLayout.doc.mjs` are untouched relative to `main`.

## Change

- `useChatStreamScroll.ts`: unconditional first-fill guard (pending flag consumed by the mount jump or the first scrollable `scrollIfLocked`), `jumpToBottom` internal (cancels in-flight spring, zeroes velocity), `behavior` option on `scrollToBottom`. Exports `ChatScrollToBottomOptions`.
- `Chat/index.ts`: type export.
- Changeset: `@astryxdesign/core` patch (`[fix]`).

No public API is removed or changed; `scrollToBottom()` with no arguments behaves exactly as before. The only behavior change is the first fill no longer spring-scrolling from the top.

## Verification

[...]

PR #3933, opened 2026-07-14
feat(chat): add ChatMessageList align prop for top-aligned lists

### LINKED ISSUE #2572: XDSChatMessageList: add an `align`/`anchor` prop for top-aligned message lists
## Problem

`XDSChatMessageList` always bottom-anchors its content. It renders an `aria-hidden` spacer `<div>` with `flex: 1` as the first inner child, which pushes messages to the bottom of the container:

https://github.com/facebookexperimental/xds/blob/main/packages/core/src/Chat/XDSChatMessageList.tsx — see `styles.spacer = { flex: 1, minHeight: 0 }` and the spacer `<div>` rendered before `children`.

This is the right default for live chat (newest message visible). But some surfaces — e.g. a static/seeded conversation that fits within the viewport, or an artifact/document side-panel chat — read better **top-anchored**, with messages starting at the top of the panel.

## Why it can't be solved at the call site today

There is **no prop** (`align`, `anchor`, `justify`, `scrollAnchor`, etc.) on `XDSChatMessageList` or `XDSChatLayout` to control this. `xstyle` is applied to the list's **root**, not the inner flex container or the spacer, so the spacer's `flex: 1` can't be overridden from outside. StyleX has no descendant/child combinators, so a consumer cannot target the spacer via `xstyle` either.

The only current workarounds all reach into the component's internal DOM:
- A raw `<style>` tag with a global selector like `.xds-chat-message-list > div > [aria-hidden] { flex: 0 0 0 }` (anti-pattern — pierces internals, leaks globally).
- A `ref` + `useEffect` setting the spacer's inline style (still pierces internals).

Both are exactly the kind of escape hatch the design system discourages.

## Proposed solution

Add an alignment prop to `XDSChatMessageList`, e.g.:

```ts
/**
 * Vertical alignment of messages within the list when content
 * doesn't fill the container.
 * - 'bottom' (default): messages anchor to the bottom (live-chat behavior).
 * - 'top': messages anchor to the top (static/seeded conversations,
 *   document/artifact side panels).
 * @default 'bottom'
 */
align?: 'top' | 'bottom';
```

Implementation: render (or collapse) the `flex: 1` spacer based on `align`. When `align === 'top'`, omit the spacer (or set it to `flex: 0`). Auto-scroll behavior owned by `XDSChatLayout` should still work for the `bottom` case.

Remember the `SYNC:` block at the top of `XDSChatMessageList.tsx` (update index exports, Storybook stories, and the ChatMessageList block examples).

## Context

Surfaced while building the `ai-chat-artifact` page template (`packages/cli/templates/pages/ai-chat-artifact/page.tsx`). The template wanted top-anchored chat; we removed the `<style>` workaround and reverted to the default bottom-anchor rather than ship the anti-pattern, pending this prop.

jiunshinn:
## Summary
Adds `align?: 'top' | 'bottom'` (default `'bottom'`) to `ChatMessageList`. `'bottom'` preserves today's behavior; `'top'` omits the internal spacer so messages start at the top and grow downward — for log-style / document-style lists.

Closes #2572.

## Why
The bottom-pinning spacer was hardcoded and unconditionally rendered, and `xstyle` only reaches the root (`ChatMessageList.tsx:275` pre-change), so consumers had no way to opt out. Top-aligned lists previously required swizzling.

## Behavior delta

| Case | `align='bottom'` (default) | `align='top'` |
|------|----------------------------|---------------|
| List shorter than container | Spacer fills free space → messages pinned to bottom (unchanged) | Spacer omitted → messages start at top |
| List overflows container | Spacer collapses to 0; scrolls to bottom | Spacer absent; scrolls to bottom (identical) |
| ChatLayout auto-scroll | unchanged | unchanged (spring early-returns when `scrollHeight <= clientHeight`, `useChatStreamScroll.ts:142`) |

## Implementation (`ChatMessageList.tsx`)
- Prop + JSDoc at the interface (`:78-93`), destructure with default `align = 'bottom'` (`:227`)
- `{align === 'bottom' && <div {...stylex.props(styles.spacer)} aria-hidden />}` (`:313-317`) — was an unconditional div

## SYNC / docs
File header + component JSDoc, `ChatMessageList.doc.mjs` (en/zh/dense), new `Alignment` Storybook story, changeset. `Chat/index.ts` unchanged (the existing type re-export already covers the new prop) — deliberately kept untouched to avoid conflicting with #3800.

## Test plan
4 new cases (spacer present for default & `bottom`, absent for `top`, children render under `top`). Full Chat suite 141/141 green; core typecheck clean.

### DIFF
--- a/.changeset/chatmessagelist-add-align-prop-for-top-aligned-l-xm43.md
+++ b/.changeset/chatmessagelist-add-align-prop-for-top-aligned-l-xm43.md
@@ -1,0 +1,16 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] ChatMessageList: add an `align` prop for top-aligned message lists
+(#2572).
+
+- `align?: 'top' | 'bottom'` (default `'bottom'`). `'bottom'` keeps the
+  existing behavior: a flex spacer fills free space so a short conversation
+  sits just above the composer. `'top'` omits the spacer so messages start at
+  the top and grow downward — better for log-style or document-style lists.
+- Only changes the resting position of a non-full list. Once messages overflow
+  the container the spacer collapses to zero in both modes, so ChatLayout
+  auto-scroll-to-bottom behavior is unchanged.
+
+@jiunshinn
--- a/apps/storybook/stories/Chat.stories.tsx
+++ b/apps/storybook/stories/Chat.stories.tsx
@@ -54,9 +54,7 @@ For **server state**, use a library like **TanStack Query** or **SWR** — they 
 
 Avoid global state managers unless you have a genuine need for cross-cutting state. Most apps are over-engineered in this area.`}</Markdown>
           <ChatMessageMetadata
-            timestamp={
-              <Timestamp value="2026-03-15T14:30:30" format="time" />
-            }
+            timestamp={<Timestamp value="2026-03-15T14:30:30" format="time" />}
             footer={
               <>
                 <span>Claude Opus 4.6</span>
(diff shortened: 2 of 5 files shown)


cixzhang (approved):
Thanks — this is exactly what [#2572](https://github.com/facebook/astryx/issues/2572) asked for, and omitting the spacer rather than zeroing it is the right cut. Would merge as is.

[Reviewed by Robohands]

Issue #3223, opened 2026-06-28
Standardize layout component props: derive a shared prop set for Stack/Grid/Center/Section/Layout/FormLayout


cixzhang:
## Summary

Astryx's layout components (`group: 'Layout'` — `Stack`, `Grid`, `Center`, `Section`, `Layout`, `FormLayout`) have **no shared prop standard**. Each evolved its own vocabulary for the same concerns (sizing, spacing, alignment, escape hatches), and the gaps get filed one component at a time. This issue derives a **standard layout prop set** and tracks aligning the components to it.

The wiki [API Conventions](https://github.com/facebook/astryx/wiki/API-Conventions) documents **Input Component Props** and `BaseProps` in detail, but has **no Layout Component Props section** — so the common vocabulary was never recorded anywhere. This adds it.

## Findings: current inconsistencies

| Concern | Inconsistency across the 6 layout components |
|---|---|
| **Escape hatches** | `Stack`, `Grid`, `Center`, `FormLayout`, `StackItem` **re-declare** `xstyle`/`className`/`style` even though `BaseProps` already provides them (the Component Auditor flags this as redundant). `Section`/`Layout*` correctly inherit. |
| **`ref` type** | Mixed: some use `React.Ref<HTMLElement>`, others `React.Ref<HTMLDivElement>`, re-declared inconsistently. |
| **Sizing** | `Stack`/`Grid`/`Center` use `SizeValue` for `width`/`height`. `Section` adds `maxWidth`/`minHeight` only. `Layout`/`LayoutHeader`/`LayoutPanel` use raw `number \| string`, and `Layout` uses `contentWidth: number`. No component exposes the full `min/max` set. |
| **Spacing** | `gap` (Stack/Grid) vs `padding` (Section/Layout) — both `SpacingStep`, but no documented split of which applies where. |
| **Alignment** | `Stack`: `hAlign`/`vAlign` + `justify`/`align` aliases. `Grid`: its own `GridAlignment`. `Center`: `axis`. Three different alignment vocabularies. |
| **Dividers** | `Section`: `dividers: Array<'top'\|'bottom'\|'start'\|'end'>`. `Layout*`: `hasDivider: boolean`. No documented rule for which shape to use. |

## Related open issues (same root cause)

All of these are requests for layout props that a standard would systematize:

- #2623 — `overflow` + flex `grow`/`shrink`/`basis` on Section/Stack/HStack/VStack
[...]

PR #3935, opened 2026-07-14
feat(core): add isSticky/stickyOffset props to Stack family

### LINKED ISSUE #2613: XDS layout primitives: no `position: sticky` prop — sticky columns/panels require custom CSS
## Summary

There's no way to make an XDS layout primitive (`XDSVStack`, `XDSLayoutPanel`, `XDSSection`, etc.) **sticky** through props. Any template that wants a column or panel to stay in view while a sibling scrolls — a very common product/detail/docs pattern — has to drop to custom CSS:

```ts
  stickyInfo: {
    position: 'sticky',
    top: 'var(--spacing-8)',
    alignSelf: 'start',
  },
});
```
```tsx
<XDSVStack xstyle={styles.stickyInfo}>…</XDSVStack>
```

Under the [template grading rubric](https://github.com/facebookexperimental/xds/wiki/Contributing-Templates#template-grading-rubric) this counts as custom CSS (3 declarations), so a template can't reach a perfect **Custom CSS 15/15** purely because of a missing sticky affordance.

## Evidence

`product-detail` (`packages/cli/templates/pages/product-detail/page.tsx`) keeps the product info column in view while the image gallery scrolls. The only way to express that today is the `stickyInfo` custom style above — there is no `sticky`/`isSticky` prop on `XDSVStack`, and no sticky slot on `XDSLayout`/`XDSLayoutPanel`.

## Proposal

A prop-driven sticky affordance on the relevant layout primitives, e.g.:

- `<XDSVStack sticky top={8}>` (token-based offset), or
- a `isSticky` + `stickyOffset` pair on `XDSLayoutPanel` / `XDSStack`.

This would let detail pages, docs sidebars, and filter rails express sticky behavior without custom CSS.

## Context

Surfaced while grading templates against the Contributing-Templates rubric (companion to #2582 *no XDSImage* and #2419 *no responsive XDSGridSpan* — the other structural gaps that cap image/layout templates below an A).

jiunshinn:
## Summary
Adds sticky props to the Stack family: `isSticky` / `stickyOffset`.

Closes #2613.

## Problem
There's no prop to make a layout primitive sticky. Any sticky column/rail/sidebar must drop to custom CSS. Live evidence: `packages/cli/templates/pages/product-detail/page.tsx:29-33` (pre-change) carried a `stickyInfo` const (`position: sticky; top: var(--spacing-8); align-self: start`) with a comment citing this issue, which caps the template's Custom-CSS rubric score.

## Change
Adds two Tier E capability props to `Stack` (inherited by `HStack`/`VStack` via `XDSStackProps`), using the `isSticky` / `stickyOffset` names reserved in the layout-prop standardization umbrella #3223:
- `isSticky: boolean` → `position: sticky` + `align-self: flex-start` (a stretched flex/grid item can't stick, so the cross-axis stretch is released — the issue's workaround explicitly listed all 3 declarations as removal targets).
- `stickyOffset: SpacingStep` → `inset-block-start: var(--spacing-N)` (default `0`; no-op unless `isSticky`).

Implemented via a shared `Layout/sticky.stylex.ts` token map mirroring `padding.stylex.ts` — no inline styles, spacing tokens only.

## Proof end-to-end
`product-detail` drops the custom CSS entirely (`page.tsx:290` → `<VStack gap={0} isSticky stickyOffset={8}>`). StyleX compiles `stickyOffset={8}` to `top: var(--spacing-8)` — byte-identical to the removed workaround.

## Behavior delta

| Scenario | Before | After |
|---|---|---|
| Sticky column next to a scrolling sibling | 3 custom CSS declarations via `style`/`xstyle` (`position` + `top` + `alignSelf`) | `<VStack isSticky stickyOffset={8}>` |
| `isSticky` alone | — | `position:sticky` + `align-self:flex-start` + `inset-block-start:0` (flush to top) |
| `stickyOffset` without `isSticky` | — | no-op (sticky not applied) |
| Existing Stack/HStack/VStack call sites | unchanged | unchanged (opt-in, off by default) |

## Scope
Stack family only, per #3223's incremental Tier E model ("each closes its linked issue"). `Section`/`LayoutPanel` sticky is left as a follow-up (LayoutPanel in particular needs a separate look at how sticky interacts with its own `overflow: clip/auto` scroll semantics); the shared `sticky.stylex.ts` module makes that follow-up near-free.

## Notes
- Sticky requires a scrolling ancestor provided by the consumer (documented in the prop JSDoc) — that responsibility stays outside the primitive.
- `align-self: flex-start` bundling can narrow a sticky header inside a column-direction parent to content width; overridable via `xstyle`, noted in the JSDoc.

## Test plan
- `vitest run` Stack/VStack/HStack: 59 passed (new: sticky class applied, offset class mapping, `stickyOffset` without `isSticky` is a no-op, HStack/VStack inheritance)
- Core build (babel + tsc + CSS) green; compiled CSS contains `position:sticky`, `align-self:flex-start`, `top:var(--spacing-8)`
- typecheck / scoped eslint / `scan.mjs` / check:sync / check:changesets all green; `astryx component Stack --dense` renders both props


cixzhang (changes requested):
Thanks for closing the sticky-layout gap. I merged current main and kept both appended Stack test blocks.

`isSticky` also changes cross-axis sizing. In a normal column layout, a full-width HStack collapses around its button; restoring its width needs the custom `xstyle` this API is meant to remove. Could we preserve the Stack's existing cross-axis sizing while keeping the rail case sticky?

Constrained Stack matrix

[Reviewed by Robohands]

cixzhang:
Thank you for the thorough implementation and evidence. Sticky behavior is a new layout-system capability, not a ready contributor pickup. We have marked #2613 `needs-scoping` and are closing this PR until the system direction is decided.

PR #3936, opened 2026-07-14
feat(eslint-plugin): add no-nullish-jsx-guard rule

### LINKED ISSUE #2538: Lint rule + utility for proper nullish checks in JSX conditionals
## Problem

When checking if a value should render in JSX, we often use truthy checks like:

```tsx
const hasSidebar = sidebar != null;
// or
{sidebar && <Sidebar />}
```

This is **incorrect for booleans, empty strings, and `0`**:
- `false != null` → `true` (renders nothing, but reports "has sidebar")
- `'' != null` → `true` (renders empty string unexpectedly)
- `0 != null` → `true` (renders `0` in the DOM)

These are all valid JSX values that pass a `!= null` check but produce unexpected rendering behavior.

## Proposal

1. **Lint rule** (in `internal/eslint-plugin/`): Flag `!= null` / `!== null` / `!== undefined` checks that are used as JSX rendering guards. Suggest using proper nullish + type-aware checks instead.

2. **Utility function** (in `packages/core/`): Something like `isRenderable(value)` that correctly identifies whether a value will produce visible JSX output:
   - `null` → false
   - `undefined` → false  
   - `false` → false
   - `true` → false
   - `''` → false
   - `0` → ⚠️ (renders as text "0" — often unintended)
   - Everything else → true

## Context

This came up reviewing conditional rendering patterns across the codebase. We should catch these at lint time rather than discovering them as visual bugs.

jiunshinn:
## Summary
Adds `@astryx/no-nullish-jsx-guard`, completing the lint-rule half of #2538. The `isRenderable` utility half was already implemented and merged in #2589 (`packages/core/src/utils/isRenderable.ts`), so this PR is scoped to the rule only.

Closes #2538.

## Problem
`{slot != null && <El>{slot}</El>}` is a common conditional-render pattern, but `!= null` only rejects null/undefined. React also renders nothing for `false`, `true`, and `''`, and all three pass a `!= null` guard — leaking an empty wrapper element into the DOM. `isRenderable(slot)` (already shipped) is the fix; it also excludes boolean and empty-string values while keeping `0` renderable.

## What the rule does
Flags `!= null` / `!== null` / `!== undefined` guards (in `&&` and ternary form) when BOTH: (1) the guarded branch renders JSX, and (2) the guarded value is rendered as a JSX *child* of that branch. Values used only as props (`{user != null && <Profile user={user} />}`) are not flagged — they're data objects, not rendered slots. Provides an ESLint suggestion rewriting the comparison to `isRenderable(value)`.

## Rollout
Registered as `warn` in both strict and recommended tiers. A scan found ~36 existing call sites in `packages/core` to migrate first (AppShell, Banner, Chat*, EmptyState, Item, MetadataList, RadioList, Table, TextArea, ToggleButton, TopNav, TreeList, ...); keeping it a warning surfaces them (including in CI) without failing the build. Promote to `error` in strict mode once core is migrated. Happy to do the migration as a follow-up (or in this PR if preferred — it changes rendering behavior in core, so it would come with a patch changeset).

## Notes
- `0` is intentionally treated as renderable (matches merged `isRenderable` semantics) — the rule never flags it.
- The suggestion rewrites the comparison but does not auto-insert the import (no import-adding autofix precedent in this plugin); the message and README point to `@astryxdesign/core/utils`.
- `internal/eslint-plugin-astryx` is private → no changeset.

## Test plan
- 21 RuleTester cases (valid/invalid + suggestions); full plugin suite 165 passing (`vitest run internal/eslint-plugin-astryx`)
- Verified the rule reports as a warning (0 errors) under the real config in CI mode against `packages/core`
- `pnpm check:repo` green


cixzhang (approved):
Approving — and apologies for the four-week wait on this one. It's careful work.

What I checked:

- **The rollout plan holds.** Under the real config with `CI=true` (strict tier) it produces **0 errors** at `warn` severity, with an ESLint suggestion on 100% of hits. There's no `--max-warnings` anywhere in the repo, so `warn` genuinely cannot fail the build. Shipping a new rule this way — visible everywhere, blocking nowhere, promoted to `error` after migration — is the right pattern, and I'd like it to be the template for future rules.
- **Your migration count is honest.** "~36 sites in core" measures as **37 today** across 22 files (`Pagination.tsx` picked one up since you counted). Quoting a real number up front, rather than asserting the rule is clean, is what made this quick to trust.
- **Tests:** full plugin suite **165 passing**, including your 21. The valid cases are doing real work — prop-vs-child, sibling values, non-nullish comparisons, and the test-file skip are all pinned.
- `isRenderable` and the `@astryxdesign/core/utils` subpath export are both real and already used by AppShell and AvatarStatusDot, so the suggestion text points somewhere that works.

### Needed before it lands

**Rebase.** The branch is 520 commits behind and currently conflicts. It's trivial: adjacent-line additions in `index.js`'s docblock and rule registry, where `no-style-only-wrapper` (#4758) landed in the same spots. Both sides are pure additions — keep both. I'll take care of this for you.

### Follow-ups, none blocking

1. **`packages/lab` has 17 more sites that won't fire** — `astryxConfig` scopes the rule to `packages/core/src/**`. Worth a line in the description so nobody expects them.
2. **Root-identifier matching can false-positive on member expressions.** `isRenderedAsChild` matches the root identifier, so `{props.icon != null && <span>{props.children}</span>}` is flagged even though the guard and the rendered value are different properties. **Zero instances of this shape exist in core today**, so it's latent rather than noisy — comparing the full member-expression text when the guard is a `MemberExpression` would close it.
3. **Numeric guards.** Roughly 6–8 of the 37 hits guard `number`s (`maxLength`, `call.additions`, `point.x`, `d.percentage`). A number can never be `false`/`true`/`''`, so `isRenderable` is a no-op there and arguably muddies intent. Your docblock already owns this limitation honestly — I'd just have the migration PR skip them rather than rewrite mechanically.
4. **Only the first nullish guard in an `&&` chain is reported** (the `break` in `LogicalExpression`), so `{a != null && b != null && <p>{a}{b}</p>}` yields one warning for two leaks. It self-heals as each is fixed; dropping the `break` would surface both at once.
5. **Four cheap test additions** if you're touching it again: a JSX fragment branch (`{x != null && <>{x}</>}`), the `.stories.` skip (only `.test.` is covered), multiple guards in one chain, and the member-expression case in (2) — pinned either way so the intended behavior is explicit rather than incidental.

Separately: `index.js` already fails `prettier --check` on `main` — pre-existing, not yours, and `.mjs`/`.js` aren't in the lint-staged glob anyway.

The conservative scoping and the up-front migration count are exactly the right instincts for a rule like this. Thanks for your patience.

cixzhang:
Hi @jiunshinn — this had drifted a fair way behind `main` (520 commits) and gone into conflict, so I merged `main` into your branch on your behalf to clear it. No rebase and no force-push — your commit `aefcf4e` is untouched, there's just a merge commit on top.

There was only one real conflict, in `internal/eslint-plugin-astryx/index.js`: the `@description` docblock rule list, where `main` had added `no-style-only-wrapper` on the same line your `no-nullish-jsx-guard` entry landed on. I kept **both** entries. `README.md` merged cleanly as a pure append. Your rule's intended severity is preserved exactly — `'warn'` in both `strict` and `recommended`, with your explanatory comment above the strict entry intact.

I checked that nothing got dropped in the merge by enumerating the plugin's rule keys before and after: `plugin.rules` 20 → 21, `configs.strict` 19 → 20, `configs.recommended` 19 → 20 — exactly the union of `main` and your branch, with no severity change to any pre-existing rule.

Verification on the merged result:
- Plugin test suite: **284 passing**, including all **21** of your `no-nullish-jsx-guard` tests.
- Repo-wide `CI=true pnpm lint` (strict mode): **0 errors**, and your rule reports **37 warnings** across 22 files in `packages/core/src` — right in line with the ~36 you noted in the rollout comment.

Auto-merge is armed (squash), so it'll land once the required checks go green. The `Vercel` check is red, but that's the usual preview-deploy limitation on fork PRs and it isn't a required check. Thanks for the rule — the `false`/`true`/`''` leak is a real one, and the conservative "only when the value is rendered as a child" scoping is the right call.

Issue #2626, opened 2026-06-10
XDSList/XDSListItem: no way to align list-item content to the container edge (optical alignment)


rubyycheung:
## Summary

`XDSListItem` insets its content by an internal horizontal padding (~8px). When a list sits directly under a section heading (or other full-bleed content), the list rows look **misaligned** — their text starts ~8px in from the heading. There's no prop to bleed the list to its container's edge, so it takes custom negative-margin CSS:

```ts
const styles = stylex.create({
  itemsList: { marginInline: -8 }, // pull rows back to align with the heading
});
```
```tsx
<XDSHeading level={2}>Items</XDSHeading>
<XDSList xstyle={styles.itemsList}>…</XDSList>
```

## Impact on the grading rubric

In the `detail-page` (Order Detail) template, the Items list needs this `-8px` pull to optically align with the "Items" section heading. Under the [Contributing-Templates rubric](https://github.com/facebookexperimental/xds/wiki/Contributing-Templates#template-grading-rubric) it's a small justified custom-CSS rule, but there's no component prop for it.

## Proposal

An `edge`/`inset` prop (or an `align="edge"` option) on `XDSList`/`XDSListItem` that removes the leading/trailing inner padding so row content aligns flush with the container — analogous to "inset vs full-bleed" list styles in other systems.

## Context

Surfaced while grading templates against the Contributing-Templates rubric — companion to #2582, #2419, #2613, #2616, #2622, #2600, #2625, #2575, #2594. These are the structural component gaps that keep otherwise-clean templates below a perfect score.

cixzhang:
Probably we should follow similar edge adjustment behaviors as table and divider, something like "isFullBleed" but that prop is not quite accurate. Also this depends on container awareness and proper use of containers.

PR #3938, opened 2026-07-14
feat(tablist): add isFullBleed to bleed a tab bar to container edges

### LINKED ISSUE #2622: XDSLayoutHeader/XDSTabList: no way to bleed a tab bar to the header's content edges — requires negative-margin CSS
## Summary

A common detail-page pattern places an `XDSTabList` at the **bottom of an `XDSLayoutHeader`**, with the tab bar (and its active-tab underline) bleeding out to the header's padded edges so the underline meets the header divider and aligns with the content area below.

There's no prop for this on `XDSLayoutHeader` or `XDSTabList`, so it takes custom negative-margin CSS that cancels the header padding:

```ts
  tabsRow: {
    marginInline: -12,   // cancel header inline padding
    marginBottom: -16,   // pull tabs to the header's bottom edge (meet the divider)
    marginTop: 12,
  },
});
```
```tsx
<XDSHStack xstyle={styles.tabsRow}>
  <XDSTabList …>…</XDSTabList>
</XDSHStack>
```

Under the [Contributing-Templates rubric](https://github.com/facebookexperimental/xds/wiki/Contributing-Templates#template-grading-rubric), the `detail-page` template is otherwise pristine — **zero raw HTML** (it even uses `XDSThumbnail` instead of a raw ``), proper `XDSLayout` slots, responsive panel. The single `tabsRow` negative-margin rule (3 declarations) is the **only** thing keeping it from 100: Custom CSS lands at 12/15 instead of 15/15. Removing it misaligns the tabs (they inset by the header padding and the underline no longer reaches the divider edge), so it's load-bearing.

## Proposal

A first-class way to express "tabs docked to the header's bottom/content edges", e.g.:

- An `XDSLayoutHeader` slot/prop for a docked tab bar (handles the edge bleed + divider alignment), or
- A `bleed`/`fullBleed` (or `edge="container"`) prop on `XDSTabList` so it can extend to its container's padding edges without manual negative margins.

## Context

Surfaced while grading templates against the Contributing-Templates rubric — companion to #2582 (no XDSImage), #2419 (no responsive XDSGridSpan), #2613 (no sticky prop), and #2616 (no XDSTable grouping). These are the structural component gaps that keep otherwise-clean templates below a perfect score.

jiunshinn:
## Summary

Adds an opt-in `isFullBleed?: boolean` (default `false`) to `TabList` so the tab bar box and divider can extend through a padded container without consumer-written inline negative margins.

Closes #2622.

## Implementation

- The `nav` cancels `--container-padding-inline-start/end` with negative margins and widens past its existing `maxWidth: 100%` clamp.
- The inner strip pads back by `max(container padding - --spacing-3, 0px)`. The first/last `Tab` or `TabMenu` supplies the remaining stop padding, so label content stays aligned with the container inset while the divider reaches the outer edge.
- The strip rule is applied after scroll padding and carries `--_tab-strip-bleed` through, preserving focus-ring room.
- `Tab` and `TabMenu` are unchanged. The behavior is inline-only; surrounding layout continues to own block docking.

## Template impact

The `detail-page` template now uses `isFullBleed` and removes only the hand-written inline margin. It intentionally keeps `marginTop: 12`, `marginBottom: -16`, and `vAlign="center"`.

Visible change: the ghost panel toggle is no longer dragged outward by the row-level inline negative margin and now aligns with the container inset.

## Verification

- `pnpm vitest run packages/core/src/TabList`: 74 tests passed
- Root build, Storybook build, core typecheck, and Storybook typecheck passed
- Storybook browser geometry guard covers both hug tabs and fill tabs ending in `TabMenu`; it verifies strip bleed, edge-label alignment, and the shared `--spacing-3` stop-padding assumption
- Chromium geometry run passed for strip, first-tab label, and trailing `TabMenu` content coordinates
- `pnpm check:repo` and `git diff --check` passed


cixzhang:
Sorry this sat so long — the negative-margin exception in the detail-page template is a real wart and you were right to go after it.

Before the prop, though: `Layout/edgeCompensation.stylex.ts` makes edge adjustment container-driven on purpose — components declare eligibility, the container detects and does the maths — and TabList already carries that attribute. And the comment this PR rewrites says List needs the same bleed ([#2626](https://github.com/facebook/astryx/issues/2626)), which reads to me as the second of three components each growing their own `isFullBleed` rather than the container growing one slot.

That's a shape question above a review, so I've put it in front of a maintainer instead of leaving you waiting on me further. Nothing for you to change meanwhile.

[Reviewed by Robohands]

packages/core/src/TabList/TabList.tsx
```diff
@@ -96,6 +112,29 @@ const styles = stylex.create({
...
+  // addition is `maxWidth: 'none'`, needed because the `nav` base sets
+  // `maxWidth: '100%'`, which would otherwise clamp the widened box back
+  // inside the container's padding.
+  fullBleed: {
+    marginInlineStart: 'calc(-1 * var(--container-padding-inline-start, 0px))',
+    marginInlineEnd: 'calc(-1 * var(--container-padding-inline-end, 0px))',
+    maxWidth: 'none',
+    width:
+      'calc(100% + var(--container-padding-inline-start, 0px) + var(--container-padding-inline-end, 0px))',
+    marginTop: {
```
cixzhang: Hmm, these match the TabList's own parent, not the padded container. Divider's horizontal bleed stays inline-only.

packages/cli/assets/templates/pages/detail-page/page.tsx
```diff
@@ -258,9 +250,13 @@ function PageHeader({
...
 
-        <HStack vAlign="center" style={tabsRow}>
+        <HStack vAlign="center">
           <StackItem size="fill">
-            <TabList value={activeTab} onChange={onTabChange} size="lg">
+            <TabList
+              value={activeTab}
+              onChange={onTabChange}
+              size="lg"
+              isFullBleed>
```
cixzhang, line 258: The strip is alone in a `StackItem` here, so it's first *and* last child — both block margins fire.

cixzhang (changes requested):
Thanks for this — good to see the `detail-page` negative-margin exception go. The prop itself is fine; Section, Table and Divider already bleed off the same `--container-padding-*` vars.

The block gating is the problem. `:first-child`/`:last-child` match the TabList's own parent, not the padded container, and in the generated template the strip sits inside a `StackItem` — so it's both. The gate meant to stop the strip being pulled up is what pulls it up 16px into the metadata row, and the underline lands 15px above the header divider instead of on it — the one thing [#2622](https://github.com/facebook/astryx/issues/2622) asked for.

Ideally I'd drop the block gating and keep this inline-only, like Divider's. Keeping the template's `marginBottom: -16`, I measured the underline back on the divider at 1px, same as main.

Does inline-only work, or was the auto-dock the part you wanted? If you'd rather talk it through with someone, we're in [Discord](https://discord.com/invite/XnsUcFykEP).

[Reviewed by Robohands]

jiunshinn:
Thanks — you're right, and inline-only works. The auto-dock wasn't the part I wanted; the gating was meant as a *guard* for mid-content strips, but as you point out `:first-child`/`:last-child` see the strip's own parent while the `--container-padding-block-*` vars inherit through any wrapper — so the guard fires exactly when the strip sits alone in a wrapper (the template's `StackItem`) and becomes the thing it was guarding against. My PR body even misread the mechanism ("the gating does not propagate") when it was actually matching and pulling the strip up.

Respun in 24893c0487:

- `fullBleed` is now inline-only (`marginInline` + `width` + `maxWidth: 'none'`), matching Divider's horizontal bleed; the style comment records why block gating can't work from here
- the `detail-page` template restores the hand-tuned block-dock margins (`marginBottom: -16`, `marginTop: 12`) with a comment noting a LayoutHeader docking slot as the eventual owner of that dock — the inline half of the old `tabsRow` exception stays gone
- prop JSDoc, `TabList.doc.mjs`, and the changeset drop the block-gating description

Underline sits back on the header divider as on main. `vitest run packages/core/src/TabList` 45/45, core typecheck and scoped eslint clean.

cixzhang (changes requested):
The block-axis fix in `24893c048` is exactly right, and the comment in `TabList.tsx` explaining why a block cancel can't reach the header's padding edge from inside `StackItem > HStack` is the clearest statement of that constraint I've seen written down. That blocker's closed — thank you for the fast turnaround.

One thing left, and it's the other half of the same idea.

The box bleeds, but the content doesn't follow it out. The first and last stops keep their normal `paddingInline` (`Tab.tsx:93`), so their labels land 4px short of the container's content inset — the only thing on the row that isn't on it. The metadata line above and the section heading below both sit on it.

tab row, this branch vs the same row with edge-stop compensation

Measured on the real `detail-page` template, rendered by `apps/template-viewer` with core aliased to source — not a repro story:

| | first tab label vs content inset | ghost toggle centre / tab centre |
|---|---|---|
| main today | 0.00px | 133.98 / 133.48 |
| this branch | **−4.00px** | 133.98 / 133.48 |
| + edge-stop compensation | **0.00px** | 133.98 / 133.48 |

The fix stays inline-only: the first and last stops pad back in by the amount the box bled out, clamped so a stop never insets less than its normal padding. Block axis untouched, which the third row confirms.

**Please don't add a block-end cancel as well.** I tried that arm: it double-docks against the template's own `marginBottom: -16` and puts the strip 9px *below* the divider with the tab centre 6px down. Your instinct to leave the block dock hand-tuned was correct — keep `marginBottom: -16`, `marginTop: 12` and `vAlign="center"` exactly as they are now.

On the tests — `className` `.toContain('fullBleed')` (`TabList.test.tsx:86` and `:99`) passes against an empty style object, because StyleX derives that class name from the object's key rather than from anything it contains. So all three pass whether or not the rule does anything, and the third asserts only that the class count grew. Not your invention: `Divider`'s `isFullBleed` test says as much in its own comment. Worth not carrying forward, though. jsdom has no layout, so this wants a browser test — set `--container-padding-inline-start` on a wrapper, assert the strip's box escapes it and the first stop's label lands back on it.

Two smaller things:

- Dropping `marginInline: -12` also moves the ghost panel-toggle inline (`ghost_box_end_inset` −12 → 0). It lands on the content inset now, which is an improvement, but it's a visible change to that button and belongs in the description rather than being discovered.
- The template moved on main — `packages/cli/templates/` → `packages/cli/assets/templates/` — and the branch is around 904 commits behind, so the rebase won't apply cleanly. Worth doing before the next round so the measurements above stay meaningful.

Full arm-by-arm numbers, frames and sensor receipts: https://github.com/cixzhang/astryx/wiki/Review-3938

cixzhang:
Correction to my last review — I sent you down the wrong path on the *how*. The diagnosis stands (the 4px is real, and it's a regression against main, not just an unfinished half), but please don't build what I described. There's a much smaller fix, and the one I asked for is broken in a way I hadn't measured.

**Do this instead.** Leave `Tab` and `TabMenu` completely alone. Keep the negative margins on the `nav` as you have them, and pad the *strip* back in by whatever the bleed exceeds a stop's own inline padding:

```ts
const STOP_INLINE_PADDING = spacingVars['--spacing-3'];
const EDGE_PAD_START = `max(var(--container-padding-inline-start, 0px) - ${STOP_INLINE_PADDING}, 0px)`;
const EDGE_PAD_END   = `max(var(--container-padding-inline-end,   0px) - ${STOP_INLINE_PADDING}, 0px)`;
```

applied to the strip element alongside the existing `--_tab-strip-bleed`. That's it — one file, no per-stop custom properties, no `:first-child`/`:last-child`. First label lands 0.00px on the inset, and the first tab's hover pill stays at main's −12 instead of moving to −16.

**Why not the version I described.** I had `TabList` publish edge-inset vars for `Tab` and `TabMenu` to consume at `:first-child`/`:last-child`. Measured across 13 configurations, it fails three ways:

- **The `TabMenu` rule never fires.** `TabMenu` returns a fragment — the trigger button plus the popover's `<template>` and a zero-size `<div>`, all siblings inside the strip — so the trigger is never `:last-child`. In `layout="fill"` its chevron lands 4px *outside* the inset: the same regression this PR exists to fix, just at the other edge. Worth knowing about `TabMenu` regardless of this PR.
- **It widens stops that aren't at an edge** — `:last-child` fires on the last Tab even in `hug`, where it's nowhere near the end, giving one tab a 4px-wider hover pill for no reason. A single-tab strip gets both rules and grows 8px.
- **It regresses `padding={0}` containers.** The property is published as `0px`, so `max(0px, --spacing-2)` resolves to 8px and the first stop's padding *shrinks* 12 → 8 — in a container with no padding to bleed at all.

**One honest limitation of the version I'm now asking for:** with container padding below 12px (`padding={2}`), the label lands 4px *inside* the inset. The strip can add padding but can't take it away from a stop. Main has the same flaw today — its hardcoded `marginInline: -12` over-bleeds an 8px container — so this is no worse than the status quo, and I'd rather ship the simple thing than a var protocol that misfires in the common case. Worth a comment in the code.

One thing to guard: subtracting `--spacing-3` in `TabList` restates `Tab`'s own `paddingInline`, and the import would be circular. If someone changes a stop's padding, edge alignment drifts silently. Please add a test that pins them together.

Everything else from my last review is unchanged: inline-only, no block-end rule, template keeps `marginBottom: -16` / `marginTop: 12` / `vAlign="center"`, and the class-name tests want replacing with browser geometry assertions.

Sorry for the churn — better now than after you'd written the three-file version. Full measurements, 13-rig matrix and the exact diffs: https://github.com/cixzhang/astryx/wiki/Review-3938

cixzhang:
Correction to my last comment: keeping `marginTop` / `marginBottom` on `tabsRow` is not an acceptable endpoint. The purpose here is to remove that whole custom style object, not only its inline margin.

Please hold off on the strip-padding implementation I suggested. I'm checking the surrounding header layout now so the next ask removes `tabsRow` entirely while preserving the underline dock and panel-toggle alignment. Sorry — I narrowed the acceptance criterion incorrectly.

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks — Cindy accepted the ownership split: TabList owns inline full bleed, and LayoutHeader owns the block-end dock. `isFullBleed` is an improvement and needs no further design change.

Two changes remain:

1. Delete `tabsRow` and `style={tabsRow}`; add `LayoutHeader.paddingBlockEnd?: SpacingStep` in `Section`'s existing per-edge family and use `paddingBlockEnd={0}`.

Redline of current head and LayoutHeader arm

At `padding={6}`, current head leaves the underline **+7.00px** above the divider; the LayoutHeader arm is docked at **−1.00px**.

2. Move `FullBleedGeometry.play` into required Chromium CI. Vitest never collects stories and the visual job never awaits `play`, so the current guard cannot fail CI.

Sorry for the contradictory round-2a direction; you built exactly what we asked for. [Full review and measurements](https://github.com/cixzhang/astryx/wiki/Review-3938).

[Reviewed by Robohands]

jiunshinn:
Thanks — both items are in as of 37eb7df.

1. `tabsRow` and `style={tabsRow}` are gone from the detail-page template, with no replacement constant. `LayoutHeader` now takes `paddingBlockEnd?: SpacingStep` in Section's existing per-edge spelling — the same `paddingBlockEndStyles` + `containerPaddingBlockEndVarStyles` pair from `padding.stylex.ts`, applied after `padding` so it wins on its own edge only, with the container var moving alongside so bleed children compensate against the padding actually applied. The header passes `paddingBlockEnd={0}`, so the underline docks on the divider from the container system vars rather than a constant that is correct at one padding. The prop has a `LayoutHeader.doc.mjs` row and unit tests in `Layout.test.tsx`, and the changeset now describes the ownership split instead of the retained `tabsRow`.

2. `FullBleedGeometry.play` is now enforced in required Chromium CI. A new `.github/scripts/story-play-guard.js` runs as a blocking step in `pr-a11y`, beside the sibling Chromium guards (pr-visual is still soft, so enforcement lives in the blocking job). It loads the story's iframe in headless Chromium and reads the play outcome from the preview channel: `storyRendered` only fires after `play` resolves, so a thrown geometry assertion, an errored story, a missing id, or a timeout all fail the job. I verified fail-closed behaviour by sabotaging an expected value (exit 1 with the assertion message) and with a nonexistent story id. The story also now covers the accepted far side of the clamp: at 8px container padding the pad-back clamps to zero and the stop keeps its own 12px inset, and outside a padded container the prop is asserted pixel-identical to a control TabList.

No worries about the round-2a direction — the correction was clearly recorded, and the driven arm A measurements made this side mechanical.

cixzhang (changes requested):
Maintainers would get conflicting guidance if this lands as-is: TabList now consumes the shared container-padding protocol, but current main’s contract still lists the old participants.

The implementation itself matches the accepted inline/block ownership split, and the Chromium guard closes the earlier test gap. Please rebase; a maintainer needs to add TabList’s bleed-consumer role and browser verification to the owning record in the same stack. I don’t see another implementation change needed.

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks — the inline/block ownership is right, and required Chromium did run on this head. The remaining gap can leave builders with a green check while the documented header rail/divider drifts: `FullBleedGeometry` hand-writes container variables, so it never exercises the real `LayoutHeader paddingBlockEnd={0}` → `TabList isFullBleed` seam. The current container-padding contract also still omits TabList from its participant and verification maps. Please cover that composition in the browser guard; a maintainer can record the consumer and evidence before landing. If you’d rather talk it through with someone, we’re in [Discord](https://discord.com/invite/XnsUcFykEP).

[Reviewed by Robohands]

cixzhang (approved):
Thanks, this closes the author-side work. The required Chromium guard now covers the real `LayoutHeader paddingBlockEnd={0}` → `TabList isFullBleed` composition and passes on this head.

[Reviewed by Robohands]

PR #3973, opened 2026-07-15
Self-host lookaside CDN images under /public

imdreamrunner:
## Summary

Astryx's demo imagery was streamed from a Meta-internal CDN (`lookaside.facebook.com/assets/astryx/*`) — invisible to external contributors and unreachable without Meta network access. An outside contributor **couldn't see which images exist, couldn't add a new one, and couldn't run the site with real imagery.**

This PR **self-hosts every one of those images in the repo**, downloaded from the internal asset set and committed under `apps/docsite/public/`. Nothing is deleted; the images simply move from the CDN into the repo.

📄 **Asset mapping** (every image → original Asset Manager source + 2x asset ids): https://gist.github.com/imdreamrunner/1574db95a724d793e703036ee625e665

## Two homes, by consumer

Images land in one of two directories based on **who uses them** — matching the repo's existing conventions:

| Directory | Used by | Notes |
|---|---|---|
| `apps/docsite/public/template-assets/` | **CLI templates** (`packages/cli/templates/**`) | The scaffold-stripper keys off this path — see below. `template-assets/` is the pre-existing convention. |
| `apps/docsite/public/images/` | **Docsite chrome only** (`apps/docsite/src/**`) | Brand banner, per-theme product photos (`matcha/ butter/ stone/ gothic/ y2k/`), theme `*-picker.png` tiles, `avatars/`. Mirrors the existing `neutral/preview-*.png` slot convention (per-theme folders avoid `preview-watch.png` collisions). |

### 1. Template demo images → self-hosted at `/template-assets/`
- Committed the shared template photo/video set under `apps/docsite/public/template-assets/`.
- Repointed all **39 CLI template files** + `Lightbox.doc.mjs` from `lookaside.facebook.com/assets/astryx/*` to root-relative `/template-assets/*`.
- Inlined the `CDN`/`NEUTRAL_CDN` base-path variables in favor of literal `/template-assets/...` paths.

### 2. Scaffold behavior fully preserved — still inlines base64
The CLI's `stripTemplateAssetRefs` (run on `astryx add`) **still swaps demo image references for a self-contained inline `data:` URI placeholder** — a generated project renders with **zero setup / zero network dependency**, exactly as before. Only the path segment the regex matches changed (`/template-assets/`, excluding `placeholder.svg`). **No image is ever copied into a scaffolded project.** Test suite 10/10.

### 3. Docsite chrome → self-hosted at `/images/`
- Brand banner, theme previews, theme pickers, and avatars committed under `apps/docsite/public/images/` and repointed there (theme showcase, hero, OG/blog metadata). Under `/images/` (not the public root) so theme-name folders can't shadow Next.js app routes.

### 4. Retina variants ready for later
- Committed **`@2x` variants** alongside their 1x siblings wherever the asset set provides a `density=2` source. Each `@2x` is exactly 2× its 1x sibling. Not referenced yet — ready for a future `srcset`/`next/image` pass without re-fetching.

### 5. Guardrail
- `check-demo-media.mjs`: image-backed Thumbnail examples must inline a same-origin, samplable `data:` URI (required by `useImageMode`'s CORS pixel-sampling).

**Result: zero `lookaside.facebook.com` references remain** except the README banner (READMEs render on GitHub and need an absolute URL — will switch to `https://astryx.atmeta.com/images/Astryx-Banner.png` after merge) and a mocked, never-fetched URL in `useImageMode.test.ts`.

### 6. Changeset
- A `[chore]` changeset is included for the `@astryxdesign/cli` package (the scaffold-strip path change). The demo video `Nature-1.mp4` (8.4 MB) is committed directly as a normal file alongside the images.

## Verification
- `stripTemplateAssetRefs` tests **10/10**; CLI template-integration **10/10**; full docsite suite **238/238**; core `useImageMode` **5/5**; `check:demo-media` pass.
- **Docsite** (`next dev`): all self-hosted assets 200, old CDN paths gone.
- **Sandbox** (`dev:sandbox`): template-assets mirror to `public/template-assets/` (gitignored), template previews render.


apps/docsite/src/__tests__/data-extraction.test.ts
```diff
@@ -319,7 +319,7 @@ describe('componentRegistry', () => {
     expect(lightbox!.playground?.defaults).toMatchObject({
       isOpen: false,
       media: {
-        src: expect.stringContaining('https://'),
```
cixzhang, line 322: If the URLs are absolute we might have a better chance to ensure these images will correctly appear in the internal docsite.

packages/cli/assets/templates/blocks/components/AspectRatio/AspectRatioWidescreen.tsx
```diff
@@ -10,7 +10,7 @@ export default function AspectRatioWidescreen() {
     <Center width={600}>
       <AspectRatio ratio={16 / 9} fit="cover">
         <img
-          src="https://lookaside.facebook.com/assets/astryx/light-scene-horizontal-1.png"
+          src="/template-assets/light-scene-horizontal-1.png"
```
cixzhang, line 13: Yeah for these CLI templates especially we might want absolute URLs. Hmm if we're done with the internal docsite package updates you can try testing by landing this then updating to the canary version. May need to wait for security clearance though (check with Gerard on how long that takes)

cixzhang:
I agree with updating this. We should make sure the CLI templates can still display correctly on the internal docsite. That might mean we need absolute URLs to guarantee we're fetching from the docsite CDN

packages/cli/assets/templates/blocks/components/AspectRatio/AspectRatioWidescreen.tsx
imdreamrunner: By absolute URL, are you referring to the full URL with domain https://astryx.atmeta.com/? yea i figured out we need that for the internal docsite though i am wondering if it would be better if we keep that URL replacing logic in the internal app itself?

Daniel15:
This PR adds `@2x` variants for some images, but doesn't seem to use them? You'd need to do something like this to actually use them:

```html
<img 
  alt="My cool picture"
  srcset="/template-assets/whatever.png 1x, /template-assets/whatever@2x.png 2x"
  src="/template-assets/whatever.png"
/>
```

(the `src` is only a fallback for browsers that don't support `srcset`)

apps/docsite/src/__tests__/data-extraction.test.ts
imdreamrunner: Discussed in internal meetings

PR #3978, opened 2026-07-15
fix(core): prevent consumer rest props from clobbering contract props

cixzhang:
## Summary

ButtonGroup, Calendar, and Carousel all spread `{...rest}` (from `BaseProps`) _after_ component-owned props on the root element. Since JSX spreads are last-wins, a consumer passing `role`, `aria-roledescription`, or `onKeyDown` through the base type would silently override the component's semantic contract.

## Changes

- **ButtonGroup:** Move `{...props}` before contract props (`role="group"`, `aria-label`, `aria-disabled`). Compose `onKeyDown` with `composeEventHandlers` so the consumer can observe or cancel navigation without losing it.
- **Calendar:** Move `{...rest}` before the root element's `onKeyDown`. Compose via `composeEventHandlers` so the Escape-cancels-range-selection handler still fires alongside consumer handlers.
- **Carousel:** Move `{...htmlProps}` before `role="region"` and `aria-roledescription="carousel"` so the component's a11y contract cannot be accidentally overridden.

## Testing

All existing tests pass (ButtonGroup 12, Calendar 20, Carousel 8 -- 65 total).

Night Watch -- Component Auditor

### DIFF
--- a/.changeset/fix-rest-prop-clobber.md
+++ b/.changeset/fix-rest-prop-clobber.md
@@ -1,0 +1,8 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Fix consumer rest props clobbering component contract props in ButtonGroup, Calendar, and Carousel
+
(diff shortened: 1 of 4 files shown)


thedjpetersen (approved):
Approved.

josephfarina (approved):

PR #3980, opened 2026-07-15
feat(dropdown-menu): focus the current selection on open

lexs:
## Summary

Adds an `isSelected` prop to `DropdownMenuItem` (and the data-driven `DropdownMenuItemData`). When set, `DropdownMenu` moves initial keyboard focus to that item when the menu opens — instead of always the first item — so the focus highlight lands on the active choice in menus that represent a current selection (e.g. a model or account picker). It also:

- exposes the item as `aria-current="true"` — the valid "current item in the set" marker for `role="menuitem"`, which supports neither `aria-selected` nor `aria-checked`; and
- renders the selected item in a medium font weight, matching how `Selector` marks its selected option.

Falls back to focusing the first item when nothing is `isSelected`, so existing menus are unchanged. Wired into both the uncontrolled (click/keyboard) open path (`openAndFocus`) and the controlled `isMenuOpen` path, and threaded through the data-driven `items` renderer for both flat items and section items.

## Motivation

A generic `DropdownMenu` used as a selection menu focus-highlights only the first item on open, which misleads the user into reading the first row as "selected." `Selector` already handles this for form-field selects, but its bordered trigger doesn't fit compact ghost-button triggers (toolbars, composers), so the menu itself needs to support "focus the current selection on open."

## Testing

Added to `packages/core/src/DropdownMenu/DropdownMenu.test.tsx`:
- `aria-current` is set on the selected item (and not on others)
- initial focus lands on the selected item on open — compound children **and** data-driven items
- a selected item **inside a section** is focused on open
- controlled-open (`isMenuOpen`) focuses the selected item
- fallback: the first item is focused when nothing is selected

Full `DropdownMenu` suite passes (42 tests). `eslint` clean, `check:changesets` valid, `tsc` clean.

## Changeset

`.changeset/dropdown-menu-selected-focus.md` — `[feat]`, patch.

### DIFF
--- a/.changeset/dropdown-menu-selected-focus.md
+++ b/.changeset/dropdown-menu-selected-focus.md
@@ -1,0 +1,8 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] DropdownMenu focuses the current selection on open
+
+`DropdownMenuItem` (and the data-driven `items`) now take an `isSelected` prop. When set, the menu moves initial keyboard focus to that item on open instead of always the first item, exposes it as `aria-current` (the valid "current item in the set" marker for `menuitem`), and renders it in a medium font weight. This keeps the focus highlight on the active choice in menus that represent a current selection — a model or account picker, for example — instead of misleadingly highlighting the first row. Falls back to focusing the first item when nothing is selected.
+@lexs
--- a/packages/core/src/DropdownMenu/DropdownMenu.doc.mjs
+++ b/packages/core/src/DropdownMenu/DropdownMenu.doc.mjs
@@ -33,7 +33,7 @@ export const docs = {
     {
       name: 'items',
       type: 'DropdownMenuOption[]',
-      description: 'Array of menu entries. Each entry is one of: an action item `{label, onClick?, icon?, isDisabled?}`, a divider `{type: "divider"}`, or a section `{type: "section", title?, items: [...action items]}`.',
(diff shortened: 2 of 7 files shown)


cixzhang:
Thank you for this, @lexs — and genuine apologies for the four-week silence. That's too long, and the reasoning below only became true partway through the wait.

Closing this one, because the motivation it was built on has since been resolved a tier down. Your Motivation section put it precisely:

> `Selector` already handles this for form-field selects, but its bordered trigger doesn't fit compact ghost-button triggers (toolbars, composers), so the menu itself needs to support "focus the current selection on open."

That gap is now closed: #4684 added `Selector`'s `variant="ghost"` — a borderless trigger that matches ghost buttons, intended for exactly the toolbar and composer cases you named — and #4802 followed up on aligning the selected option during popover entry. So a compact selection control in a toolbar should now reach for a ghost `Selector`, which already gets selected-option focus, `aria-activedescendant`, and typeahead, rather than a `DropdownMenu` taught to behave like a selector.

The tier distinction is the thing we want to hold: `DropdownMenu` is a menu of *actions*, `Selector` is a control for a *current value*. `isSelected` on a menu item is a reasonable local fix, but it moves `DropdownMenu` toward being a second selection widget with its own selected-state semantics, and then the two drift. Keeping selection in `Selector` is the call.

Worth saying: your `aria-current="true"` reasoning was right — `menuitem` supports neither `aria-selected` nor `aria-checked`, and that's the correct marker for the genuine action-menu case. If you hit a place where a ghost `Selector` still doesn't fit, please reopen or file an issue with the specific case; that would be useful signal about a real gap rather than a preference.

Sorry again for the wait on a well-made PR.

PR #3982, opened 2026-07-15
fix(core): give CommandPalette search input an accessible name

bhamodi:
## Summary

The `CommandPaletteInput` combobox (`CommandPaletteInput.tsx`, `role="combobox"`) rendered with no accessible name -- the file had zero `aria-label` occurrences and relied solely on `placeholder`. A placeholder is not a reliable accessible name: several screen reader/browser pairings ignore it for the accname, and it disappears from the accessibility tree's text alternatives guidance entirely once the user starts typing. In practice a screen-reader user opening the palette landed focus on an anonymous "edit combobox" with no announcement of what the field searches.

`Selector` already solves this for its popover search input: `Selector.tsx:806` puts an explicit `aria-label="Search options"` on its `role="combobox"` search field. This brings `CommandPaletteInput` to parity via a new optional `label` prop that renders `aria-label={label ?? placeholder}` -- so the default accessible name is the placeholder text (`'Search...'` when unset), keeping the announced name consistent with what sighted users see (WCAG 2.5.3 label-in-name). The `{...props}` rest spread already sits after the ARIA attributes, so a consumer-passed `aria-label` (or `aria-labelledby`, which wins in accname computation anyway) still overrides the default; a test now locks that ordering in.

## Changes
- `CommandPalette/CommandPaletteInput.tsx`: add optional `label?: string` prop; render `aria-label={label ?? placeholder}` on the combobox input (placeholder defaults to `'Search...'`, so the name is always defined). Consumer `aria-label` via rest props continues to win.
- `CommandPalette/CommandPaletteInput.test.tsx`: four new tests -- default accessible name, `label` prop respected (and placeholder unaffected), custom placeholder used as fallback name, consumer `aria-label` override.
- `CommandPalette/CommandPaletteInput.doc.mjs`: document the new `label` prop in `docs`, `docsZh`, and `docsDense`. (The file also picked up mechanical prettier normalization -- it was not prettier-clean on `main`, and the repo's pre-commit hook formats staged files.)

## Test plan
- `node_modules/.bin/vitest run --root . packages/core/src/CommandPalette` -- **49 pass** (6 files). The three positive new tests (default name, `label` prop, placeholder fallback) **failed before the fix** with `aria-label` = `null`; the override test passed pre-fix, confirming the rest-spread ordering already let consumer ARIA win.
- `node_modules/.bin/vitest run --root . packages/core` -- **4585 pass** (203 files, full core suite).

## Notes

### DIFF
--- a/packages/core/src/CommandPalette/CommandPaletteInput.tsx
+++ b/packages/core/src/CommandPalette/CommandPaletteInput.tsx
@@ -110,6 +110,13 @@ export interface CommandPaletteInputProps extends Omit<
   placeholder?: string;
 
   /**
+   * Accessible label for the combobox input, announced by screen readers.
+   * Falls back to the placeholder text (`'Search…'` by default), since a
+   * placeholder alone is not a reliable accessible name.
+   */
+  label?: string;
+
+  /**
    * Whether to auto-focus the input when mounted.
    * @default true
    */
@@ -149,6 +156,7 @@ export function CommandPaletteInput({
   value: controlledValue,
   onValueChange,
   placeholder: placeholderFromProps,
+  label,
   hasAutoFocus = true,
   endContent,
   onChange,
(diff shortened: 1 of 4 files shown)


josephfarina (approved):
lg

packages/core/src/CommandPalette/CommandPaletteInput.tsx
```diff
@@ -148,6 +155,7 @@ export function CommandPaletteInput({
   value: controlledValue,
   onValueChange,
   placeholder = 'Search...',
+  label,
```
cixzhang, line 159: For clarity we could make this `label = 'Search'`.

cixzhang (approved):

PR #3984, opened 2026-07-15
fix(core): guard Dialog entry animation with prefers-reduced-motion

bhamodi:
## Summary

`Dialog`'s entry animation ignored the user's OS-level reduced-motion setting. The `enterDirectional` keyframes (`Dialog.tsx:109-116`) translate the dialog from the trigger's direction and scale it from 0.95, and were applied unconditionally via `animationName` in the `open` style (`Dialog.tsx:155` before this change) -- there was no `prefers-reduced-motion` guard anywhere in the file. For users with vestibular disorders who set "reduce motion", opening any dialog still played a full translate+scale animation, which can trigger dizziness and nausea. This made Dialog the odd one out: every other overlay in the repo already guards this -- the shared layer pattern lives in `layerAnimations.stylex.ts:82-108`, which disables each entry keyframe under `@media (prefers-reduced-motion: reduce)` so the layer appears instantly instead of translating/scaling in.

This applies that exact pattern to Dialog's `open` style: `animationName: {default: enterDirectional, '@media (prefers-reduced-motion: reduce)': 'none'}`. Under reduced motion the dialog now appears instantly at full opacity (`opacity: 1` is set statically by the `open` style, so disabling the keyframe leaves no flash or stuck state). I audited the rest of the file for other motion: the `animationDuration`/`animationTimingFunction`/`animationFillMode` declarations in `styles.dialog` are inert without an `animationName`, and there are no transitions or other transforms -- so this single guard covers all motion in the component. There is no separate pure-opacity fade to leave unguarded; the opacity portion of the keyframe is disabled together with the movement, exactly as the layerAnimations precedent does for its layers.

## Changes
- `Dialog/Dialog.tsx`: guard `styles.open`'s `animationName` with `@media (prefers-reduced-motion: reduce): 'none'`, mirroring `layerAnimations.stylex.ts`, with a comment citing the pattern. The file also picked up a two-line reformat of a pre-existing type-assertion in the `data-testid` block: the repo's lint-staged config runs `prettier --write` on every staged `.tsx` file at commit time, and origin/main's `Dialog.tsx` predates the current prettier 3.9.4 output for that construct (verified: the untouched origin/main file fails `prettier --check` on exactly that hunk).

## Test plan
- `node_modules/.bin/vitest run --root . packages/core/src/Dialog` -- **47 pass** (3 files: Dialog, DialogHeader, useImperativeDialog).
- `node_modules/.bin/vitest run --root . packages/core` -- **4581 pass** (203 files, full core suite).
- `node_modules/.bin/eslint packages/core/src/Dialog/Dialog.tsx` -- clean; `prettier --check` -- clean.
- No new test asserts the media-query behavior: jsdom cannot flip `prefers-reduced-motion`, StyleX compiles the conditional into CSS the browser evaluates, and the identical guard in `layerAnimations.stylex.ts` is itself untested (the only reduced-motion test in the repo, `useStreamingText.test.ts`, mocks `matchMedia` for a JS runtime code path, which is not analogous). Rather than fabricate an assertion against compiled class names, correctness here rests on matching the established, shipped pattern byte-for-byte plus the suite and typecheck staying green.

## Notes
Found during a broader a11y audit of core components; scoped to one fix per PR. `Dialog.doc.mjs` was left unchanged: it documents props and usage, and no props or exports changed -- matching how the layer-based components do not document their reduced-motion guard in their doc files either.

### DIFF
--- a/.changeset/dialog-reduced-motion.md
+++ b/.changeset/dialog-reduced-motion.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Dialog: the entry animation is disabled under prefers-reduced-motion, matching the Layer animation guards.
+@bhamodi
--- a/packages/core/src/Dialog/Dialog.tsx
+++ b/packages/core/src/Dialog/Dialog.tsx
@@ -152,7 +152,13 @@ const styles = stylex.create({
   open: {
     display: 'flex',
     opacity: 1,
-    animationName: enterDirectional,
+    // Disable the entry keyframe animation under
+    // `prefers-reduced-motion: reduce` so the dialog appears instantly
+    // instead of translating/scaling in (same pattern as layerAnimations).
+    animationName: {
+      default: enterDirectional,
+      '@media (prefers-reduced-motion: reduce)': 'none',
+    },
   },
   // Backdrop using ::backdrop pseudo-element
   backdrop: {
@@ -531,7 +537,11 @@ export function Dialog({
           ),
           className,
           style,
-        )}>
+        )}
+        data-testid={
+          (props as Record<string, unknown>)['data-testid'] as
+            string | undefined
+        }>
         {innerContent}
       </div>
     );


josephfarina (approved):
makes sense

cixzhang (approved):

PR #3986, opened 2026-07-15
fix(core): default an accessible name on the TopNav landmark

bhamodi:
## Summary

`TopNav` declared `label?: string` with no default (`packages/core/src/TopNav/TopNav.tsx`), and both of its `<nav>` render paths emitted `aria-label={label}` verbatim -- so omitting the prop shipped an **unnamed navigation landmark**. For a screen-reader user navigating by landmarks, a typical page composing `SideNav` + `TopNav` + `Breadcrumbs` announced multiple entries as just "navigation", with the TopNav one impossible to distinguish or jump to reliably. Sibling landmarks already name themselves: `SideNav` hardcodes `aria-label="Side navigation"`, `Breadcrumbs` defaults to `'Breadcrumb'`, `Pagination` to `'Pagination'`, and `Outline` to `'Table of contents'` -- TopNav was the odd one out. This continues the naming work from 7b1512d42 (`fix(topnav): give the logo link an accessible name`), which fixed the unnamed logo link inside the same component.

The fix defaults the prop in destructuring -- `label = 'Top navigation'` -- so **both** render paths are covered: the default full-bar `<nav>` and the `mobile-bar` `<nav>` (selected via `TopNavRenderContext` when AppShell renders TopNav as a mobile top bar). The `drawer` render path emits no `<nav>` of its own (it renders into `MobileNav`), so it needs no label. An explicit `label` still wins. The name follows `SideNav`'s "Side navigation" precedent.

## Changes
- `TopNav/TopNav.tsx`: default `label = 'Top navigation'` in the destructuring (covers the default and mobile-bar `<nav>` paths); document the default with `@default` on `TopNavProps.label`.
- `TopNav/TopNav.doc.mjs`: add `default: "'Top navigation'"` to the `label` prop entry.
- `TopNav/TopNav.test.tsx`: three new tests -- default landmark name when `label` is omitted (default mode), default landmark name in mobile-bar mode, and custom label overriding the default in mobile-bar mode (default-mode override was already covered).
- `.changeset/topnav-landmark-label.md`: patch changeset.

Note: the pre-commit prettier hook also collapsed one pre-existing formatting drift in `TopNav.tsx` (the `endContent` `TopNavSlotContext` line) -- whitespace only, no behavior change.

## Test plan
- `node_modules/.bin/vitest run --root . packages/core/src/TopNav` -- **63 pass** (3 files). The two positive default-name tests (`defaults the landmark label to "Top navigation" when label is omitted`, `defaults the landmark label in mobile-bar mode`) **failed before the fix** (`2 failed | 61 passed`) and pass after.
- `node_modules/.bin/vitest run --root . packages/core` -- **4584 pass** (203 files), no failures.
- `eslint` on changed files -- clean; `prettier --check` on changed `.tsx` files -- clean after the pre-commit hook's formatting pass (`TopNav.doc.mjs` is excluded from prettier by lint-staged and keeps the doc-file house style, matching its siblings).

## Notes

### DIFF
--- a/.changeset/topnav-landmark-label.md
+++ b/.changeset/topnav-landmark-label.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] TopNav: the `<nav>` landmark now defaults its accessible name to "Top navigation" when the `label` prop is omitted, matching SideNav ("Side navigation"), Breadcrumbs, and Pagination. Previously an omitted `label` shipped an unnamed navigation landmark, leaving screen-reader users with multiple indistinguishable "navigation" landmarks on pages that compose SideNav + TopNav + Breadcrumbs. An explicit `label` still wins.
+@bhamodi
--- a/packages/core/src/TopNav/TopNav.doc.mjs
+++ b/packages/core/src/TopNav/TopNav.doc.mjs
@@ -109,6 +109,7 @@ export const docs = {
       name: 'label',
       type: 'string',
       description: 'Accessible label for the navigation landmark, applied as aria-label on the <nav> element.',
+      default: "'Top navigation'",
     },
     {
       name: 'xstyle',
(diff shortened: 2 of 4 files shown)


humbertovirtudes (approved):
lgtm

PR #3447, opened 2026-07-02
fix(a11y): announce file selection, page changes, and multiselect count via live region (re-land #3407)

cixzhang:
## Summary

Re-lands the accessibility live-region wiring for `FileInput`, `Pagination`, and `MultiSelector`. This content was originally in #3407, which was accidentally merged into its stacked base branch instead of `main`, so its changes never reached `main`. This PR re-applies just the wiring on top of current `main`.

Builds on the shared `useAnnounce` live-region hook (already in `main`).

## What changed

- **FileInput** announces successful file selection (`"1 file selected: <name>"` / `"N files selected"`); validation errors stay on the existing `role="status"` region so they are not double-announced.
- **Pagination** announces page changes (`"Page N of M"` / `"Page N"`); only user-driven changes go through the handler, so initial mount stays silent.
- **MultiSelector** announces selection-count changes (`"N of M selected"`, `"All selected"`, `"Selection cleared"`) on toggle, select-all, and clear.

All three route through the shared visually-hidden polite live region, so these previously-silent surfaces are now audible to screen-reader users.

## Notes

- Layered on top of `main`'s current code in these components (existing aria attributes on `FileInput`, the current `optimisticValue` handling in `MultiSelector`, and the `pageSize` coercion in `Pagination`).
- `.changeset/announce-wiring.md` included.

Refs #3343.

## Testing

- `vitest run` on FileInput, MultiSelector, Pagination — 141 tests pass
- `eslint` on changed files — clean
- `tsc --project packages/core/tsconfig.json --noEmit` — clean
- `node scripts/check-sync.js` — clean

PR #3989, opened 2026-07-15
fix(core): include selected filenames in FileInput trigger name

bhamodi:
## Summary

The `FileInput` trigger is a `role="button"` div that sets `aria-label={label}` (`FileInput.tsx:737`). Per accname computation, `aria-label` suppresses the element's child content from the accessible name -- and the selected filenames render as children (`fileNames` in both `renderDropzoneContent` and `renderCompactContent`). So a screen-reader user who tabs back to the control after selecting files hears only the field label, with no way to tell what (if anything) is attached. The one-time pick announcement from #3447 (7d902060c, `useAnnounce`) is transient by design; the persistent name never reflected the selection.

This composes the filenames into the label: `aria-label={hasFiles && fileNames ? `${label}, ${fileNames}` : label}`. With files selected the trigger reads e.g. "Document, report.pdf"; with none it stays exactly the label. It reuses the existing `hasFiles`/`fileNames` derivations (already comma-joined for multiple files) -- no renderer restructuring, no new strings.

Prior FileInput a11y work: 7dfac3ae2 (#3404, forms-6) moved `aria-describedby`/`required`/`invalid` onto this same focusable trigger, and 7d902060c (#3447, forms-17) added the pick-time live-region announcement. This fills the remaining gap: the persistent accessible name.

## Changes
- `FileInput/FileInput.tsx`: compose the trigger's `aria-label` from the label plus `fileNames` when files are selected (one attribute; falls back to plain `label` when empty), with a comment explaining why child content alone is not enough.
- `FileInput/FileInput.test.tsx`: new `trigger accessible name` describe block (4 tests) -- single file included in the name, multiple files all present (comma-joined), no files means the name is exactly the label, and dropzone mode covered.
- `.changeset/fileinput-trigger-name.md`: patch changeset.

## Test plan
- `node_modules/.bin/vitest run --root . packages/core/src/FileInput` -- **53 pass** (was 49). The 3 positive new tests (single, multiple, dropzone) **fail pre-fix** with the trigger named only "Document"/"Files"/"Upload"; the label-only test guards against regression in the empty state.
- `node_modules/.bin/vitest run --root . packages/core` -- **4585 pass, 203 files** (full core suite).

## Notes

### DIFF
--- a/.changeset/fileinput-trigger-name.md
+++ b/.changeset/fileinput-trigger-name.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] FileInput: the trigger's accessible name now includes the selected filenames, so screen-reader users can review what is attached when refocusing the control.
+
+@bhamodi
--- a/packages/core/locales/en.json
+++ b/packages/core/locales/en.json
@@ -679,6 +679,10 @@
     "defaultMessage": "Clear {label}",
     "description": "Screen-reader-only label on the X that removes the selected file from a FileInput. Example: `Clear Attachment`, `Clear Résumé`."
   },
+  "@astryx.fileInput.triggerWithFiles": {
+    "defaultMessage": "{label}, {fileNames}",
+    "description": "Accessible name for a FileInput trigger that has files attached, composing the field label with the selected filenames. Example: `Attachments, report.pdf, notes.txt`."
+  },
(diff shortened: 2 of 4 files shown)


josephfarina (approved):
lg

josephfarina (approved):
lg

PR #3990, opened 2026-07-15
fix(core): make collapsed CodeBlock content inert

bhamodi:
## Summary

`CodeBlock`'s scroll container is deliberately keyboard-focusable (`tabIndex={0}` + `role="group"` + `aria-label`) so keyboard users can scroll long or wide code. But in the collapsible path, collapsing only hides the code visually: the collapse wrapper (`<div id={regionId}>`) animates to a `0fr` grid row and clips content with `overflow: hidden`, while applying **no `inert`** and **no `aria-hidden`** (`packages/core/src/CodeBlock/CodeBlock.tsx`).

The result is an invisible tab stop: with the block collapsed, Tab lands on a zero-height scroll container that keyboard users cannot see, and screen readers still announce the hidden group. This builds on the prior CodeBlock collapse a11y work in d7bde219a (focusable header with a visible focus ring, `aria-expanded`, and an always-resolvable `aria-controls`), which fixed the toggle itself but left the collapsed content reachable.

The fix applies `inert={isCollapsed ? true : undefined}` to the collapse wrapper, matching the existing React 19 boolean-`inert` precedent in `SideNavItem.tsx` (collapsed nav children). `inert` removes the subtree from the tab order and the accessibility tree while it stays mounted, so:

- `aria-controls={regionId}` on the header toggle keeps pointing at a resolvable element (the d7bde219a invariant).
- The `0fr` grid `grid-template-rows` transition is unaffected -- `inert` changes interactivity, not layout or transitions, and SideNav already applies it to the same kind of animating grid wrapper.
- Expanding removes the attribute entirely (`undefined`), restoring the scroll container's `tabIndex={0}` focusability.

## Changes
- `packages/core/src/CodeBlock/CodeBlock.tsx`: add `inert={isCollapsed ? true : undefined}` to the collapse wrapper (`<div id={regionId}>`), with a comment explaining why the visual-only 0fr collapse also needs inert.
- `packages/core/src/CodeBlock/CodeBlock.test.tsx`: two new tests -- collapsed wrapper has `inert` and the `role="group"` scroll container sits inside `[inert]`; expanding again removes `inert` and the scroll container keeps `tabindex="0"`.
- `.changeset/codeblock-collapsed-inert.md`: patch changeset.

## Test plan
- `node_modules/.bin/vitest run --root . packages/core/src/CodeBlock` -- **30 pass** (3 files), including the 2 new tests. Both new tests **fail before the fix** (`Expected the element to have attribute: inert / Received: null`) and pass after. The d7bde219a behaviors (focusable header, `aria-expanded` toggling, `aria-controls` resolvable while collapsed) still pass.
- Full suite: `node_modules/.bin/vitest run --root . packages/core` -- **4583 pass, 0 fail** (203 files).
- `eslint` and `prettier --check` on changed files -- clean.

## Notes

### DIFF
--- a/.changeset/codeblock-collapsed-inert.md
+++ b/.changeset/codeblock-collapsed-inert.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] CodeBlock: collapsed code regions are now inert, so keyboard focus can no longer land on the invisible scroll container while collapsed.
+
+@bhamodi
--- a/packages/core/src/CodeBlock/CodeBlock.test.tsx
+++ b/packages/core/src/CodeBlock/CodeBlock.test.tsx
@@ -210,6 +210,62 @@ describe('CodeBlock', () => {
     expect(document.getElementById(controlsId as string)).not.toBeNull();
   });
 
+  it('makes the collapsed region inert so the scroll container is unreachable', () => {
+    render(
+      <CodeBlock
+        code={LONG_CODE}
+        language="javascript"
+        title="example"
+        isCollapsible
+      />,
+    );
+    const header = screen
+      .getAllByRole('button')
+      .find(el => el.hasAttribute('aria-expanded'))!;
+    const region = document.getElementById(
+      header.getAttribute('aria-controls') as string,
+    )!;
+    // Expanded: the region is not inert and the scroll container is reachable.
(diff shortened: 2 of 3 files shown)


nynexman4464 (approved):
seemsreasonable

cixzhang (approved):

Issue #2891, opened 2026-06-17
Standardize accessible-name props: reserve `label` for visible text, use `aria-label` for a11y-only names


kentonquatman:
## Summary

XDS uses the `label` prop inconsistently for accessibility. Across the ~80 components in `packages/core` that declare a `label` prop, it carries three different meanings:

| Meaning of `label` | ~Count | Examples |
|---|---|---|
| **Visible label** (rendered as text) | ~41 | `XDSTextInput`, `XDSSlider`, `XDSSwitch`, `XDSRadioList`, `XDSField`, `XDSBadge`, `XDSTab`, `XDSDivider` |
| **aria-label only** (never visible) | ~31 | `XDSStatusDot`, `XDSToolbar`, `XDSButtonGroup`, `XDSSegmentedControl`, `XDSBreadcrumbs`, `XDSPagination`, `XDSOutline` |
| **Dual** (visible text that becomes `aria-label` only when icon-only/collapsed) | ~8 | `XDSButton`, `XDSLink`, `XDSTopNavItem`, `XDSMoreMenu`, `XDSClickableCard`, `XDSToggleButton` |

Because the plurality of components use `label` to mean **visible text**, the ~31 components that use `label` to mean an *invisible* accessible name are surprising. A consumer reasonably expects `<XDSStatusDot label="Online" />` to render the word "Online", when in fact it only sets `aria-label` on a `role="img"` span.

## Recommendation (Option A)

Establish one rule: **`label` always means visible text. An accessibility-only name uses the native `aria-label` prop.**

Every XDS component already accepts `aria-label` because `XDSBaseProps` keeps all `aria-*` attributes, so this leans on the platform rather than inventing a parallel convention.

- **Visible-label group (~41):** no change. `label` = visible text.
- **aria-only group (~31, incl. `XDSStatusDot`):** rename `label` → `aria-label`, re-declared as **required** so the compile-time a11y enforcement is preserved (the base `aria-label` is optional).
- **Dual group (~8):** no change. `label` is their primary visible text; the existing fall-through to `aria-label` when icon-only is correct.

End state: `label` is never an invisible string. Components with `role="img"`, `role="group"`, landmark roles, etc. take a required `aria-label`.

### Example — `XDSStatusDot`

```tsx
// Before
export interface XDSStatusDotProps extends XDSBaseProps<HTMLSpanElement> {
  variant: XDSStatusDotVariant;
  /** Accessible label describing the status. Required for a11y. */
  label: string;
[...]

PR #3997, opened 2026-07-15
fix(core): per-row accessible names for Table selection checkboxes

bhamodi:
## Summary

The Table selection plugin renders every row checkbox as `<CheckboxInput label="Select row" isLabelHidden />` (`useTableSelection.tsx`), so a screen-reader user scanning the checkbox column hears an undifferentiated "Select row" N times with no way to tell which row each checkbox controls. `UseTableSelectionConfig` exposed no way to derive a per-row label.

This adds an optional `getRowLabel?: (item: T) => string` to the selection config. When provided, the row checkbox's hidden label becomes `` `Select ${getRowLabel(item)}` `` (e.g. "Select Alice", "Select Bob"), read fresh from the config ref on each render like the existing `getIsItemSelectable` / `getIsItemEnabled` callbacks. Without it, behavior is unchanged: the fallback stays the plain "Select row" -- the selection column's `renderCell` receives only `(item: T)` with no row index, so a "Select row N" fallback would have required plumbing new state through the cell pipeline for marginal benefit. The header checkbox keeps its "Select all rows" name.

## Changes
- `Table/plugins/selection/useTableSelection.tsx`: add `getRowLabel` to `UseTableSelectionConfig` (with JSDoc `@example`); `SelectionCellContentInner` derives the checkbox label from it, falling back to "Select row".
- `Table/useTableSelection.doc.mjs`: document `getRowLabel` in `docs`, `docsZh`, and `docsDense`, plus an "Accessible per-row checkbox names" example rendered by the CLI (`astryx component useTableSelection`).
- `Table/plugins/selection/useTableSelection.test.tsx`: cover per-row names, select-all invariance, and the fallback.
- `.changeset/table-selection-row-names.md`: patch changeset for `@astryxdesign/core`.

## Test plan
- `node_modules/.bin/vitest run --root . packages/core/src/Table` -- **360 pass** across 18 files. New tests in `useTableSelection.test.tsx`:
  - derives per-row accessible names from getRowLabel ("Select Alice" / "Select Bob" / "Select Charlie", and asserts no "Select row" remains) -- **fails pre-fix** (label stays "Select row" because the config field is ignored)
  - keeps the "Select all rows" header label when getRowLabel is provided
  - existing "renders row checkboxes with \"Select row\" label" test covers the no-`getRowLabel` fallback
- `node_modules/.bin/vitest run --root . packages/core` -- **4583 pass** (203 files, full core suite; no pre-existing failures).
- `node packages/cli/bin/astryx.mjs component useTableSelection --dense` -- new prop row and example render correctly.

## Notes

### DIFF
--- a/.changeset/table-selection-row-names.md
+++ b/.changeset/table-selection-row-names.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Table: the selection plugin accepts getRowLabel so row checkboxes announce which row they select, instead of an undifferentiated "Select row".
+@bhamodi
--- a/packages/core/locales/en.json
+++ b/packages/core/locales/en.json
@@ -411,6 +411,10 @@
     "defaultMessage": "Select row",
     "description": "Aria label for the \"select row\" checkbox on a Table row."
   },
+  "@astryx.table.selection.selectRowNamed": {
+    "defaultMessage": "Select {label}",
+    "description": "Aria label for a Table row's selection checkbox when a per-row label is available (via getRowLabel). `label` is the row's human-readable identity, e.g. \"Alice\"."
+  },
   "@astryx.table.sort.ascending": {
     "defaultMessage": "Sort ascending",
     "description": "Screen-reader-only label on a column header button that will sort the column ascending. Part of a set with `sort.descending` and `sort.clear` — keep parallel."
--- a/packages/core/src/Table/plugins/selection/useTableSelection.test.tsx
+++ b/packages/core/src/Table/plugins/selection/useTableSelection.test.tsx
@@ -40,9 +40,11 @@ const selectableColumns: TableColumn<SelectableUser>[] = [
 function SelectionTable({
   getIsItemSelectable,
   getIsItemEnabled,
+  getRowLabel,
 }: {
   getIsItemSelectable?: (item: SelectableUser) => boolean;
   getIsItemEnabled?: (item: SelectableUser) => boolean;
+  getRowLabel?: (item: SelectableUser) => string;
 }) {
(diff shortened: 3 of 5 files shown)


humbertovirtudes:
Nice, focused fix. One design question worth surfacing before this goes to review, since I suspect Cindy will raise it:

`getRowLabel` is optional and falls back to the generic "Select row" when omitted. That is the same pattern Cindy pushed back on in #2891 ("Standardize accessible-name props"). Her argument there was that a11y labels which "sound optional" tend not to get set, so the accessibility contract silently degrades:

> when we have label it's a required prop on our components with guaranteed accessibility patterns [...] I dislike aria-label generally because it sounds optional

That issue was deferred rather than decided against, so nothing forces the choice here, but the philosophy is likely to come up. Two ways to get ahead of it:

- Keep it optional (current) but tie the tradeoff explicitly to the "why not required" question in the PR body: the selection column's `renderCell` only receives `(item)`, so a numbered "Select row N" fallback isn't free, and requiring `getRowLabel` would be a breaking change for every existing `useTableSelection` caller. You already have most of this reasoning written up: just frame it as the answer to "why isn't this mandatory."
- Or make per-row naming required (or add a required `accessibleLabel`-style prop) so the guarantee is enforced, accepting the breaking change.

Either is defensible. The point is to name the decision so the reviewer isn't left wondering whether "optional" was intentional.

Two minor things while here:

- `getRowLabel` returning `''` produces the label `"Select "` (trailing space), since `'' != null`. Worth a truthy guard, or a test that documents the behavior.
- "Select" is hardcoded English, same as the existing "Select row" / "Select all rows". Not a regression and out of scope for this PR, but the hook ships a `docsZh` locale, so it is a latent i18n gap if we ever localize these strings.

cixzhang (approved):

PR #3998, opened 2026-07-15
fix(core): accessible name for rich-label CheckboxList items

bhamodi:
## Summary

`CheckboxListItem` accepts a `ReactNode` for `label` (rich labels are explicitly documented on the prop), but the inner focusable control was named with `label={typeof label === 'string' ? label : 'Checkbox'}` (`CheckboxListItem.tsx:227`). Any item with a rich label therefore exposed the literal accessible name "Checkbox" -- a list of rich-label items read to a screen reader as "Checkbox, Checkbox, Checkbox...", with no way to tell the options apart.

Sibling components offered no reusable precedent: `RadioListItem` and `SelectableCard` both constrain `label` to `string`, and `ChatDictationButton` only computes an internal string. So this adds an optional `accessibleLabel?: string` prop to `CheckboxListItem`, used as the `CheckboxInput` name when `label` is not a string. When a rich label is passed without `accessibleLabel`, the existing "Checkbox" fallback is kept for compatibility, and a dev-time `console.warn` fires once per item instance (ref + effect, mirroring the unnamed-dialog guardrail in `usePopover.tsx:391-401`) explaining exactly what to pass.

## Changes
- `CheckboxList/CheckboxListItem.tsx`:
  - new `accessibleLabel?: string` prop with JSDoc and an `@example`; the inner `CheckboxInput` is named `label` (string) → `accessibleLabel` → `'Checkbox'` fallback.
  - warn-once dev guardrail when a ReactNode `label` has no `accessibleLabel`, telling the developer to pass a concise string equivalent.
- `CheckboxList/CheckboxListItem.doc.mjs`: `accessibleLabel` documented in `docs`, `docsZh`, and `docsDense`, plus a "Rich label with an accessible name" code example.
- `CheckboxList/CheckboxList.test.tsx`: new `CheckboxListItem accessible name` describe block (3 tests).
- `.changeset/checkboxlistitem-name.md`: patch changeset.

## Test plan
- `node_modules/.bin/vitest run --root . packages/core/src/CheckboxList` -- **46 pass** (was 43). New tests:
  - string label → checkbox named from the label (unchanged behavior)
  - ReactNode label + `accessibleLabel` → checkbox named from `accessibleLabel`, no warning
  - ReactNode label without `accessibleLabel` → falls back to "Checkbox", warns exactly once mentioning `accessibleLabel`, and does not warn again on re-render
- Confirmed pre-fix: the two `accessibleLabel` tests fail on the base commit (checkbox named "Checkbox", zero warnings); the string-label test passes, proving the existing behavior is untouched.
- `node_modules/.bin/vitest run --root . packages/core` -- **4584 pass** (203 files, no pre-existing failures).
- `node_modules/.bin/eslint` + `prettier --check` on changed files -- clean. (Prettier also reflowed three pre-existing `description:` lines in `CheckboxListItem.doc.mjs` that failed `--check` on `main`; formatting-only.)

## Notes

### DIFF
--- a/packages/core/src/CheckboxList/CheckboxListItem.doc.mjs
+++ b/packages/core/src/CheckboxList/CheckboxListItem.doc.mjs
@@ -7,13 +7,21 @@ export const docs = {
   subComponentOf: 'CheckboxList',
   displayName: 'Checkbox List Item',
   isHiddenFromOverview: true,
-  description: 'Individual checkbox item with label, description, and end content slot. Works in collection mode (inside CheckboxList) or standalone mode (inside List).',
+  description:
+    'Individual checkbox item with label, description, and end content slot. Works in collection mode (inside CheckboxList) or standalone mode (inside List).',
   props: [
     {
       name: 'label',
-      type: 'string',
-      description: 'Primary text label for the item.',
+      type: 'ReactNode',
+      description:
+        'Primary text label for the item. A plain string names the checkbox automatically; a rich (ReactNode) label should be paired with aria-label so screen readers get a concise name.',
       required: true,
+    },
+    {
+      name: 'aria-label',
+      type: 'string',
+      description:
+        'Plain-text accessible name for the checkbox when label is a ReactNode. Applied to the checkbox control. Without it, rich-label items all announce as the generic "Checkbox" to screen readers.',
     },
     {
       name: 'value',
@@ -69,19 +77,36 @@ export const docs = {
       description: 'Direct check handler (standalone mode only).',
     },
   ],
+  examples: [
+    {
+      label: 'Rich label with an accessible name',
+      code: `<CheckboxListItem
+  label={<span>Pro plan <Badge label="Recommended" /></span>}
+  aria-label="Pro plan"
+  value="pro"
+/>`,
+    },
+  ],
 };
 
 export const docsZh = {
   name: 'CheckboxListItem',
   isHiddenFromOverview: true,
   displayName: 'Checkbox List Item',
-  description: '单个复选框选项，包含标签、描述和尾部内容插槽。可在集合模式或独立模式下使用。',
+  description:
+    '单个复选框选项，包含标签、描述和尾部内容插槽。可在集合模式或独立模式下使用。',
   props: [
     {
       name: 'label',
       type: 'string',
       description: '选项的主要文本标签。',
       required: true,
+    },
+    {
+      name: 'aria-label',
+      type: 'string',
+      description:
+        '当 label 是 ReactNode 时，复选框的纯文本无障碍名称。缺少它时，富标签选项都会向屏幕阅读器播报为通用的 "Checkbox"。',
     },
     {
       name: 'value',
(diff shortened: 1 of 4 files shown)


packages/core/src/CheckboxList/CheckboxListItem.tsx
```diff
@@ -142,6 +161,33 @@ export function CheckboxListItem({
...
+    ) {
+      warnedUnnamedCheckboxRef.current = true;
+      console.warn(
+        'CheckboxListItem: `label` is a ReactNode, so the checkbox falls ' +
+          'back to the generic accessible name "Checkbox". Pass ' +
+          '`accessibleLabel` with a concise string equivalent of the visible ' +
+          'label so screen readers can tell items apart.',
+      );
+    }
+  }, [label, accessibleLabel]);
```
cixzhang: Hmm this would happen every render if label is a react node since react nodes aren't usually memoized. But maybe that's a good thing? We probably don't want to keep state just for a dev message.

packages/core/src/CheckboxList/CheckboxListItem.doc.mjs
```diff
@@ -7,14 +7,21 @@ export const docs = {
...
   displayName: 'Checkbox List Item',
   isHiddenFromOverview: true,
-  description: 'Individual checkbox item with label, description, and end content slot. Works in collection mode (inside CheckboxList) or standalone mode (inside List).',
+  description:
+    'Individual checkbox item with label, description, and end content slot. Works in collection mode (inside CheckboxList) or standalone mode (inside List).',
   props: [
     {
       name: 'label',
       type: 'string',
       description: 'Primary text label for the item.',
```
cixzhang: Could be helpful to update the docs for label to recommend accessible label when it's relevant. Help create some pipeline for the LLM learn the pattern.

packages/core/src/CheckboxList/CheckboxListItem.tsx
```diff
@@ -51,6 +51,24 @@ export interface CheckboxListItemProps extends BaseProps<HTMLLIElement> {
...
+   * @example
+   * ```
+   * <CheckboxListItem
+   *   label={<span>Pro plan <Badge label="Recommended" /></span>}
+   *   accessibleLabel="Pro plan"
+   *   value="pro"
+   * />
+   * ```
+   */
+  accessibleLabel?: string;
```
cixzhang: Hmm I've been avoiding adding these props to the codebase where it overlaps with direct HTML props but under a different name. What you think about making this "aria-label"? We can still explicitly guide for it and do the same checks below

cixzhang (changes requested):

packages/core/src/CheckboxList/CheckboxListItem.tsx
bhamodi: Renamed to `aria-label` as suggested. It names the checkbox control and keeps the same dev warning + guidance; other aria-* still forward to the row. `aria-label` also wins over a string `label` now (standard ARIA precedence).

packages/core/src/CheckboxList/CheckboxListItem.tsx
bhamodi: Kept as-is per your read -- the check is cheap and it's only a dev warning, so it isn't worth holding state just to skip it.

packages/core/src/CheckboxList/CheckboxListItem.doc.mjs
bhamodi: Done -- updated the `label` prop docs to recommend `aria-label` for rich (ReactNode) labels.

cixzhang (approved):

PR #3999, opened 2026-07-15
fix(core): gate Item aria-selected on roles that permit it

bhamodi:
## Summary

`Item` emitted `aria-selected={isSelected || undefined}` unconditionally on its root element (`packages/core/src/Item/Item.tsx`). The default root is a generic `div` (or an `li` via `ListItem`, which forwards `isSelected` at `packages/core/src/List/ListItem.tsx`), but WAI-ARIA only permits `aria-selected` on `option`, `tab`, `row`, `gridcell`, `columnheader`, `rowheader`, and `treeitem`. So every selected item in a plain `List` shipped invalid ARIA (axe: `aria-allowed-attr`) -- the attribute is meaningless noise on `role="listitem"`/generic roles and can confuse assistive tech.

This follows the same "make Item's ARIA honest" direction as 360fc9cbb (Item rest-spread vs explicit ARIA precedence) and the invalid-ARIA removals in 4497b08f8 (TabList `aria-orientation` on `nav`), 3d6390ac3 (ButtonGroup `aria-orientation` on `role=group`), and d9633cc04 (toast viewport `aria-modal`).

How `Item` resolves its role, surveyed before gating: the ARIA role comes solely from the `role` prop (the existing `hasParentRole` flag is just `role != null`). `InteractiveRoleContext`/`useInteractiveRole` resolve *element type* (`'link' | 'button' | 'inert'`), not ARIA roles, and `Item` doesn't consume them -- so the `role` prop is the single source of truth to gate on.

The fix gates the attribute on a permitted-role set: `aria-selected` is emitted only when `role` is one of `option | tab | row | gridcell | columnheader | rowheader | treeitem` and `isSelected` is true. The selected **visual** styling remains unconditional.

**Decision for non-permitted roles:** selected Items expose no selection semantics at all (visual only). Callers that need announced selection already pass a proper role (e.g. `role="option"` inside a `listbox`), and inventing a substitute (`aria-current`, live-region text) would be wrong-by-default for a generic primitive. `DropdownMenuItem` passes `role="menuitem"` (not in the set) and never passes `isSelected`, which is correct -- menu selection would use `menuitemradio`/`menuitemcheckbox` + `aria-checked`. Selector/MultiSelector render their options on their own `role="option"` elements with explicit `aria-selected` (not through `Item`), so their announcements are unchanged.

## Changes
- `Item/Item.tsx`: add `ARIA_SELECTED_ROLES` set (per WAI-ARIA 1.2) and `allowsAriaSelected`; emit `aria-selected` only when the resolved role permits it. Documented the behavior on the `isSelected` prop JSDoc.
- `Item/Item.test.tsx`: default `div` Item with `isSelected` now asserts **no** `aria-selected`; new tests for `role="option"` (emits `aria-selected="true"`) and `role="menuitem"` (does not emit).
- `List/List.test.tsx`: `ListItem` (`li`) with `isSelected` now asserts **no** `aria-selected`.
- `.changeset/item-aria-selected-role.md`: patch changeset.

## Test plan
- Pre-fix: the 3 new/updated negative tests **fail** on `origin/main` behavior (`div` root has `aria-selected="true"`, `role="menuitem"` has it, `ListItem` `li` has it), confirming they capture the defect.
- `node_modules/.bin/vitest run --root . packages/core/src/Item packages/core/src/List packages/core/src/Selector packages/core/src/MultiSelector` -- **4 files, 194 pass** (Selector/MultiSelector announcement suites untouched and green).
- Full `node_modules/.bin/vitest run --root . packages/core` -- **203 files, 4583 pass**, 0 fail.
- eslint clean on changed files; `prettier --check` clean (the two test files were prettier-nonconformant on `main`, so a few pre-existing lines in them were mechanically rewrapped by the repo's pre-commit prettier).

## Notes

### DIFF
--- a/packages/core/src/List/List.test.tsx
+++ b/packages/core/src/List/List.test.tsx
@@ -436,11 +436,7 @@ describe('List', () => {
   it('sets target on anchor when provided', () => {
     const {container} = render(
       <List>
-        <ListItem
-          label="External"
-          href="https://example.com"
-          target="_blank"
-        />
+        <ListItem label="External" href="https://example.com" target="_blank" />
       </List>,
     );
     const anchor = container.querySelector('a');
@@ -500,17 +496,21 @@ describe('List', () => {
   // Selected state
   // ===========================================================================
 
-  it('applies aria-selected when isSelected', () => {
+  it('conveys selection via aria-current when isSelected', () => {
+    // aria-selected is not permitted on an li (role listitem, axe:
+    // aria-allowed-attr), so selection is exposed via aria-current — valid on
+    // any element — so screen-reader users are still told which item is chosen.
     const {container} = render(
       <List>
         <ListItem label="Selected" isSelected onClick={() => {}} />
       </List>,
     );
     const item = container.querySelector('.astryx-item');
-    expect(item).toHaveAttribute('aria-selected', 'true');
+    expect(item).not.toHaveAttribute('aria-selected');
+    expect(item).toHaveAttribute('aria-current', 'true');
   });
 
-  it('does not apply aria-selected when not selected', () => {
+  it('applies neither aria-selected nor aria-current when not selected', () => {
     const {container} = render(
       <List>
         <ListItem label="Not Selected" />
@@ -518,6 +518,7 @@ describe('List', () => {
     );
     const li = container.querySelector('li');
     expect(li).not.toHaveAttribute('aria-selected');
+    expect(li).not.toHaveAttribute('aria-current');
   });
 
   // ===========================================================================
--- a/.changeset/item-aria-selected-role.md
+++ b/.changeset/item-aria-selected-role.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Item: aria-selected is now emitted only when the item's role permits it (option, tab, treeitem, grid cells), removing invalid ARIA from plain list items.
+
+@bhamodi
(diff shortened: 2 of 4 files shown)


packages/core/src/List/List.test.tsx
```diff
@@ -500,14 +496,16 @@ describe('List', () => {
   // Selected state
   // ===========================================================================
 
-  it('applies aria-selected when isSelected', () => {
+  it('does not apply aria-selected when isSelected', () => {
+    // aria-selected is not permitted on an li (role listitem); selection on
+    // plain list items is visual-only (axe: aria-allowed-attr).
```
cixzhang: Hmm I guess we would still have an a11y problem here where we can't communicate the li state to AT users.

cixzhang (changes requested):
The update generally makes sense. It exposes a new issue though so requesting to see if there's better resolutions.

packages/core/src/List/List.test.tsx
bhamodi: Good point -- trading invalid ARIA for a silent state wasn't right. It now falls back to `aria-current="true"` when the role can't carry `aria-selected` (listitem, a bare div), which is valid on any element, so AT is still told which item is selected. Permitted roles still use `aria-selected`, and a consumer-provided `aria-current` wins.

cixzhang (approved):

PR #4000, opened 2026-07-15
fix(core): label Dialog from its header title via aria-labelledby

bhamodi:
## Summary

`Dialog` rendered `<dialog aria-modal="true">` with no default accessible-name wiring (`Dialog.tsx:547-551` pre-fix). The `DialogHeader` title -- the visible, auto-focused `<Heading level={2} tabIndex={-1}>` (`DialogHeader.tsx:158-165`) -- had no `id`, and `DialogContext` carried only `{isInline: boolean}` (`DialogContext.ts:14-17`), so there was no channel to connect the two. Any consumer who didn't manually pass `aria-label`/`aria-labelledby` shipped an unnamed modal: screen readers announced "dialog" with no indication of what the dialog was for. `AlertDialog` already solves exactly this with `useId` -- it generates a title id and passes `aria-labelledby` down to `Dialog` (`AlertDialog.tsx:137-138, 154-155`); this PR gives plain `Dialog` + `DialogHeader` the same behavior automatically.

`Dialog` now generates a title id via `useId` and publishes it through `DialogContext` alongside a `registerTitle` callback. `DialogHeader` renders the id on its title `Heading` and registers its presence in a mount effect (with cleanup on unmount). `Dialog` emits `aria-labelledby={titleId}` only when (a) a header title has actually registered and (b) the consumer passed neither `aria-label` nor `aria-labelledby` -- the consumer always wins, since their values spread after the default. Registration (rather than an unconditional id reference) handles the no-`DialogHeader` case robustly: an `aria-labelledby` pointing at a nonexistent id is itself invalid, so no header means no attribute. A dev-time warn-once effect (mirroring the unnamed-dialog warning in `usePopover.tsx:393-401`) fires when a dialog is open with no resolvable name; a ref mirror of the registration state lets the warning effect see a title registered in the same commit (child effects run before parent effects).

## Changes
- `Dialog/DialogContext.ts`: add optional `titleId` and `registerTitle` to `DialogContextValue` (existing consumers -- CommandPalette reads only `isInline` -- are unaffected).
- `Dialog/Dialog.tsx`: generate `titleId` via `useId`; track header-title registration (state for the attribute, ref for same-commit warning reads); emit `aria-labelledby` on the `<dialog>` only when unlabelled-by-consumer and a title exists; add the warn-once unnamed-dialog effect; document the default-labelling behavior in the component JSDoc.
- `Dialog/DialogHeader.tsx`: render `id={titleId}` on the title `Heading`; register presence with the parent Dialog via effect; update JSDoc.
- `Dialog/Dialog.doc.mjs` / `Dialog/DialogHeader.doc.mjs`: note that the header title labels the dialog via `aria-labelledby`.
- `Dialog/Dialog.test.tsx`: five new tests under `accessible name`.
- `.changeset/dialog-default-labelledby.md`: patch changeset.

## Test plan
- `node_modules/.bin/vitest run --root . packages/core/src/Dialog packages/core/src/AlertDialog` -- **64 pass** (Dialog.test.tsx 35, DialogHeader.test.tsx, useImperativeDialog.test.tsx, AlertDialog.test.tsx; AlertDialog suite untouched and green). Five new tests:
  - dialog is labelled by the DialogHeader title by default (`aria-labelledby` equals the heading's id; accessible name equals the title)
  - consumer-provided `aria-label` wins (no `aria-labelledby` emitted)
  - consumer-provided `aria-labelledby` wins (external id preserved)
  - dialog without header and without labels emits no dangling `aria-labelledby` and fires the dev warning exactly once (`console.warn` spy)
  - header title suppresses the warning
  - The two positive tests (default labelling; warning) were confirmed failing before the fix (`2 failed | 28 passed` pre-fix); the consumer-wins tests are regression guards.
- `node_modules/.bin/vitest run --root . packages/core` -- **4586 pass** (203 files; full core suite green, no pre-existing failures).
- `node_modules/.bin/eslint` on changed files -- clean; `prettier --check` on changed ts/tsx/md files -- clean (the repo's lint-staged pre-commit formats staged files, which also normalized two tiny pre-existing formatting drifts in the touched files).

## Notes
Found during a broader a11y audit of core components; scoped to one fix per PR. Coexists cleanly with #3984, which touches only `Dialog.tsx`'s style block (`animationName`); this PR's hunks are in different regions (imports, context wiring, aria attributes, effects) and the animation styles are untouched. `AlertDialog` is deliberately unchanged -- its explicit `useId` wiring already produces a named dialog, and since it renders no `DialogHeader` and passes its own `aria-labelledby`, the new default never interferes with it.

### DIFF
--- a/packages/core/src/Dialog/DialogHeader.tsx
+++ b/packages/core/src/Dialog/DialogHeader.tsx
@@ -65,7 +65,9 @@ export interface DialogHeaderProps extends BaseProps<HTMLDivElement> {
   ref?: React.Ref<HTMLDivElement>;
   /**
    * The title of the dialog.
-   * This title receives focus when the dialog opens for screen reader accessibility.
+   * This title receives focus when the dialog opens for screen reader
+   * accessibility, and names the parent Dialog via aria-labelledby unless the
+   * consumer passes an explicit aria-label/aria-labelledby to the Dialog.
    */
   title: string;
 
@@ -105,7 +107,8 @@ export interface DialogHeaderProps extends BaseProps<HTMLDivElement> {
  * Renders a title that receives focus when a modal dialog opens (for screen reader accessibility)
  * and an optional close button. Inline documentation previews suppress this autofocus.
  * The title is an h2 element with tabIndex={-1} so it can be programmatically focused but
- * doesn't appear in the tab order.
+ * doesn't appear in the tab order. The title also names the parent Dialog via
+ * aria-labelledby (unless the Dialog receives an explicit aria-label/aria-labelledby).
  *
  * Uses LayoutHeader internally for consistent styling with other layout headers.
  *
@@ -137,10 +140,13 @@ export function DialogHeader({
   const titleRef = useRef<HTMLHeadingElement>(null);
   const dialogContext = useDialogContext();
   const shouldAutoFocus = dialogContext?.isInline !== true;
+  const titleId = dialogContext?.titleId;
 
   // Auto-focus the title when mounted for screen reader accessibility.
   // Inline dialogs are documentation/showcase previews, so suppress focus to
   // avoid stealing scroll position from the surrounding page.
+  // The parent Dialog detects this title (by `titleId`) via a callback ref to
+  // set its default aria-labelledby — no registration handshake needed here.
   useEffect(() => {
     if (shouldAutoFocus && titleRef.current) {
       titleRef.current.focus();
@@ -162,6 +168,7 @@ export function DialogHeader({
         <div {...stylex.props(styles.titleWrapper)}>
           <Heading
             ref={titleRef}
+            id={titleId}
             level={2}
             tabIndex={-1}
             xstyle={styles.titleFocusable}>
--- a/.changeset/dialog-default-labelledby.md
+++ b/.changeset/dialog-default-labelledby.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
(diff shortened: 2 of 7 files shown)


packages/core/src/Dialog/DialogHeader.tsx
```diff
@@ -147,6 +152,11 @@ export function DialogHeader({
     }
   }, [shouldAutoFocus]);
 
+  // Tell the parent Dialog a title exists so it can point aria-labelledby at
+  // it. Registration (rather than an unconditional id reference) keeps the
+  // Dialog from emitting aria-labelledby when no DialogHeader is rendered.
+  useEffect(() => registerTitle?.(), [registerTitle]);
```
cixzhang: Hmm to avoid this mount effect, would it be possible to get the results from a callback ref on the header element instead? We could also sync the aria prop on the dialog from here. It's not ideal from react state management standpoint but avoids a second render pass at the dialog level.

cixzhang (changes requested):
For the suggestion. Curious what you think since we apply this strategy in a few places where there's tighter perf requirements.

packages/core/src/Dialog/DialogHeader.tsx
bhamodi: Done -- switched to a callback ref on the `<dialog>` that checks for the header title and sets `aria-labelledby` imperatively, so there's no registration effect and no second render at the dialog level. Dropped the `registerTitle` context handshake entirely.

cixzhang (approved):

PR #4001, opened 2026-07-15
fix(core): explicit decorative handling for unnamed Thumbnail images

bhamodi:
## Summary

`Thumbnail` declared `alt?: string` whose own JSDoc said "Required for accessibility when `src` is provided", then rendered `alt={alt ?? ''}` -- a silently-empty alt that marked potentially meaningful images decorative with no signal to the developer. `TopNavMegaMenuFeaturedCard` had the same silent `imageAlt ?? ''` default; there the card has a visible `title`, so a decorative image is usually **correct** -- but it should be explicit, not accidental.

`Avatar` already has the intended pattern (a845e8d6e was the most recent pass over it): when it has no accessible name it flips to **explicit** presentational semantics -- `role="presentation"` + `aria-hidden` alongside the empty alt -- instead of leaving an ambiguous empty-alt img. This PR mirrors that in both components: no `alt` → the `` gets `alt=""` + `role="presentation"` + `aria-hidden`; `alt` provided → normal img semantics. The prop stays optional (no breaking change).

For `Thumbnail` there is also a dev-time warn-once: eae28724b gave the root a `group` role named by `label` (falling back to a generic `"thumbnail"`), so when `label` (or `alt`, or a consumer `aria-label`/`aria-labelledby` via rest props) names the widget, a decorative image is fine and silence is correct. Only when `src` is set with **no name source at all** does the component warn (once per instance, via the same ref-in-effect pattern as `usePopover`'s unnamed-dialog guardrail). The FeaturedCard gets no warning -- its required visible `title` always names the card.

## Changes
- `Thumbnail/Thumbnail.tsx`:
  - no `alt` → image renders `alt=""` + `role="presentation"` + `aria-hidden` (explicitly decorative, matching Avatar); with `alt` → plain ``.
  - dev-time `console.warn` (once per instance) when `src` is set with no `alt`, no `label`, and no `aria-label`/`aria-labelledby` in rest props -- the only case where the thumbnail degrades to a generic "thumbnail" name.
  - updated `alt` JSDoc and file header to document the behavior.
- `TopNav/TopNavMegaMenuFeaturedCard.tsx`: same explicit-decorative handling for `imageAlt`; JSDoc notes decorative is usually correct here because the card has a visible title. No warning.
- `Thumbnail/Thumbnail.doc.mjs`, `TopNav/TopNavMegaMenuFeaturedCard.doc.mjs`: `alt`/`imageAlt` descriptions updated (docs, docsZh, docsDense).
- `.changeset/thumbnail-decorative-alt.md`: patch changeset.

## Test plan
- `node_modules/.bin/vitest run --root . packages/core/src/Thumbnail packages/core/src/TopNav` -- **81 pass** (4 files). New tests, with the 3 positive ones confirmed failing pre-fix:
  - Thumbnail: no alt → img is `alt=""` + `role="presentation"` + `aria-hidden="true"` and not exposed as a nameless img (**failed pre-fix**); alt provided → exposed as `img` named "Vacation photo" with no presentation/aria-hidden markers; warns once when `src` is set with no alt and no label (**failed pre-fix**); does not warn when named via `label` or `alt`.
  - FeaturedCard: omitted `imageAlt` → explicitly decorative markers (**failed pre-fix**); `imageAlt` provided → exposed as named img; title text and CTA link semantics unchanged with a decorative image.
- `node_modules/.bin/vitest run --root . packages/core` -- **4588 pass** (203 files, full core suite; no pre-existing failures on origin/main).
- `node_modules/.bin/vitest run --root . packages/cli` -- **1692 pass** (doc.mjs feeds the CLI docs; belt-and-suspenders).
- `node_modules/.bin/eslint` on changed source/test files -- clean (`*.doc.mjs` is eslint-ignored repo-wide).
- `prettier --check` on changed `.tsx`/`.md` files -- clean. The two `.doc.mjs` files fail `prettier --check` **identically on origin/main** (pre-existing: lint-staged only formats `*.{ts,tsx,md}` and CI runs no prettier); left in the files' existing style rather than reformatting ~150 unrelated lines.
- `pnpm check:repo` (sync comments, package boundaries, changesets, demo media) -- clean.

## Notes
Found during a broader a11y audit of core components; scoped to one fix per PR. Explicitly does not touch `TopNavMegaMenu.tsx`, which is owned by the merged sibling PR #3992.

### DIFF
--- a/packages/core/src/Thumbnail/Thumbnail.tsx
+++ b/packages/core/src/Thumbnail/Thumbnail.tsx
@@ -13,6 +13,10 @@
  * Uses useImageMode (APCA) to detect image luminance so the overlaid
  * remove button always has sufficient contrast.
  *
+ * Images without `alt` are explicitly decorative (alt="" +
+ * role="presentation" + aria-hidden, matching Avatar). A dev-time warning
+ * fires once when `src` is set with no `alt` and no other name source.
+ *
  * SYNC: When modified, update these files to stay in sync:
  * - /packages/core/src/Thumbnail/Thumbnail.doc.mjs
  * - /packages/core/src/Thumbnail/Thumbnail.test.tsx
@@ -21,6 +25,7 @@
  * - /packages/cli/templates/blocks/components/Thumbnail/ (showcase blocks)
  */
 
+import {useEffect, useRef} from 'react';
 import * as stylex from '@stylexjs/stylex';
 import {
   colorVars,
@@ -55,7 +60,14 @@ export interface ThumbnailProps extends BaseProps<HTMLDivElement> {
    */
   src?: string;
   /**
-   * Alt text for the image. Required for accessibility when `src` is provided.
+   * Alt text for the image.
+   *
+   * When omitted, the image is explicitly decorative (`alt=""` +
+   * `role="presentation"` + `aria-hidden`, matching Avatar) and hidden from
+   * assistive technology. Provide `alt` whenever the image conveys content;
+   * without it, screen reader users only hear the `label` (file name), or a
+   * generic "thumbnail" if `label` is also missing — which triggers a
+   * dev-time warning.
    */
   alt?: string;
   /**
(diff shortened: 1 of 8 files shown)


packages/core/src/Thumbnail/Thumbnail.tsx
```diff
@@ -265,10 +277,45 @@ export function Thumbnail({
...
+  // (or a consumer-provided aria name) becomes the group's accessible name —
+  // but with no name source at all the group falls back to a generic
+  // "thumbnail". Warn once per component instance (in an effect) rather than
+  // on every render.
+  const hasNameSource =
+    alt != null ||
+    label != null ||
+    props['aria-label'] != null ||
+    props['aria-labelledby'] != null;
+  const warnedUnnamedImageRef = useRef(false);
```
cixzhang, line 323: I wonder if we should make a utility for these component warnings so they all follow this ref pattern.

packages/core/src/Thumbnail/Thumbnail.tsx
```diff
@@ -265,10 +277,45 @@ export function Thumbnail({
   const accessibleName =
     label && alt ? `${label} — ${alt}` : (label ?? alt ?? 'thumbnail');
```
cixzhang: @nynexman4464 should the "thumbnail" here have been caught by the i18n lint?

packages/core/src/Thumbnail/Thumbnail.tsx
```diff
@@ -265,10 +277,45 @@ export function Thumbnail({
...
 
+  // Without `alt`, the image is explicitly decorative rather than silently
+  // empty-alt, matching Avatar's handling of unnamed images.
+  const isImageDecorative = !alt;
```
cixzhang, line 310: Should label be included as a fallback here or are we avoiding it because of repetition in AT?

packages/core/src/Thumbnail/Thumbnail.tsx
bhamodi: Good catch -- yes it should be. The i18n lint only scans JSX attributes, so this one slipped through since the name is built up in a variable before it's applied. Routed it through `t('@astryx.thumbnail.fallbackName')` now.

packages/core/src/Thumbnail/Thumbnail.tsx
bhamodi: Avoiding it on purpose -- `label` already becomes the group's accessible name, so mirroring it into the img's `alt` would double-announce it (group name + img alt). With no `alt` the img stays decorative, so the name comes only from the group.

packages/core/src/Thumbnail/Thumbnail.tsx
bhamodi: Agreed -- it's a cross-component pattern (Dialog, usePopover, and this). I'd rather land a shared warn-once utility as a focused follow-up than fold that refactor into this a11y fix. Happy to file it.

cixzhang (approved):

PR #4002, opened 2026-07-15
fix(core): announce field status messages via persistent live regions

bhamodi:
## Summary

Field status messages lived in live regions that were **conditionally mounted -- born together with their content**:

- `Field.tsx` (~241-248): `statusNode = status?.message ? <FieldStatus .../> : null`
- `Switch.tsx` (~537-546) and `CheckboxInput.tsx` (~603-610): `{status?.message && <FieldStatus ... variant="detached"/>}`
- `TimeInput.tsx` (~609-617, grouped mode): a hand-rolled `{inputGroup && status?.message && <VisuallyHidden role={...} aria-live={...}>...}`

`FieldStatus.tsx:146-147` carried `role={type === 'error' ? 'alert' : 'status'}` with matching `aria-live`. The nuance: dynamically **inserted** `role="alert"` nodes are announced by most modern screen-reader/browser pairs (the ARIA alert insertion exception), so error messages mostly worked today. But non-error `role="status"` born with its content is unreliable across AT, and the repo's own guidance says never to conditionally mount live regions -- `useAnnounce.ts:50-54` documents exactly this ("many screen readers will not announce content injected into a live region that is created together with its content"), and commit 188a3dc4e requires routing live-region announcements through `useAnnounce` instead of hand-rolled `aria-live` nodes. TimeInput itself already demonstrates the correct pattern in the same file (~654): an always-mounted VisuallyHidden alert whose *text* toggles.

**Fix (Option A, centralized):** `FieldStatus` no longer carries `role`/`aria-live` on its rendered element (it keeps its `id` for `aria-describedby` and all visuals) and instead announces `message` changes from an effect via the persistent `useAnnounce` singletons -- assertive for `type === 'error'`, polite otherwise. Conditional mounting at every call site becomes harmless automatically, with zero call-site changes in Field/Switch/CheckboxInput. TimeInput's hand-rolled grouped-mode region converts to the same approach (the VisuallyHidden node stays, role-free, as the `aria-describedby` target). There is no double-announcing: the element has no live semantics at all, so the singleton is the only channel.

**Why not Option B (always-mount at call sites):** it would touch four call sites, require inventing an "empty" render mode for FieldStatus (its `type` prop is required, and the element has visible padding/background, so an empty always-mounted status leaves a colored gap unless a new empty state is designed), and still depends on dynamically toggling `role` between `alert`/`status` on one node -- itself unreliable. Option A is less churn and follows the repo's sanctioned primitive.

**Announcement contract (documented in the component JSDoc):** the message is announced whenever it appears -- *including on first mount* -- and whenever the message or type changes; an empty message never announces. Announce-on-mount is deliberate: it preserves the one case that worked before (a form mounting with a server-side error previously fired the alert insertion exception), and a status that mounts with a new view is information the user should hear. Because the singletons coalesce, a page full of statuses produces at most one polite and one assertive readout instead of N.

## Changes

- `packages/core/src/FieldStatus/FieldStatus.tsx`: drop `role`/`aria-live` from the rendered div; announce `message` via `useAnnounce` from an effect (assertive for errors, polite otherwise); document the contract in the JSDoc.
- `packages/core/src/TimeInput/TimeInput.tsx`: grouped-mode status node loses `role`/`aria-live` (kept solely as the `aria-describedby` target); a new effect announces grouped status messages through the same singletons. Gated on `inputGroup` so ungrouped mode (which delegates to Field -> FieldStatus) never double-announces.
- `packages/core/src/FieldStatus/FieldStatus.doc.mjs`: describe the announcement behavior (SYNC requirement).
- `packages/core/src/{FieldStatus,Field,TimeInput,Switch,CheckboxInput}` tests: new/updated coverage (see test plan).
- `packages/core/src/{TextArea,NumberInput,CheckboxList,FileInput}` tests: collision fixes for the now-persistent singleton regions (see test plan).
- `.changeset/fieldstatus-live-region.md`: patch changeset.

Untouched on purpose: `DateInput.tsx:633` and NumberInput's "Invalid number" node are already always-mounted inline alerts (not FieldStatus); FileInput's own always-mounted `role="status"` region is separate plumbing; `InputGroup` exposes its status via `aria-describedby` only (no live region).

## Test plan

- `vitest run --root . packages/core/src/FieldStatus packages/core/src/Field packages/core/src/TimeInput packages/core/src/Switch packages/core/src/CheckboxInput` -- **179 pass** (FieldStatus 27, Field 33, TimeInput 37, Switch 40, CheckboxInput 32, + FieldLabel).
- **Pre-fix verification:** with only the two source files reverted (tests kept), **14 tests fail** exactly as intended -- FieldStatus 7, Field 3, TimeInput 2, Switch 1, CheckboxInput 1 -- covering both the born-with-content repro (status appearing after mount lands in the announce region) and the new element-level contract.
- `vitest run --root . packages/core` -- **4587 pass, 0 fail (203 files)**. Verified the 5 initial full-suite failures were all introduced by this change (all pass on clean origin/main) and fixed them as below; no pre-existing failures on origin/main.
- `tsc --project packages/core/tsconfig.json --noEmit` -- clean. `eslint` on changed files -- clean. `prettier --check` on changed files -- clean.

New tests: status-appears-after-mount announcement regressions for Field, Switch, CheckboxInput, and grouped TimeInput; assertive/polite routing per type; re-announce on message change; re-route on type change; no announcement for an empty message; role-free grouped TimeInput status node that still resolves via `aria-describedby`.

Deliberately updated existing tests (all were asserting the old element-level live semantics or colliding with the new persistent regions):

- `FieldStatus.test.tsx` "uses role=alert for error status" -> element is no longer a live region; replaced by "does not carry role or aria-live" + "announces error messages assertively, including on first mount".
- `FieldStatus.test.tsx` "uses role=status for warning/success status" (2 tests) -> replaced by polite-channel announcement assertions.
- `FieldStatus.test.tsx` "sets aria-live=assertive/polite for error/warning/success" (3 tests) -> folded into the channel-routing assertions above (the element carries no `aria-live`).
- `FieldStatus.test.tsx` "relaxes role and aria-live when type changes from error" -> split into "keeps the element role-free when type changes" + "re-routes to the polite channel when type changes from error".
- `Field.test.tsx` "status has role=alert and aria-live=assertive for error type" -> "announces error status assertively via the persistent live region".
- `Field.test.tsx` "status has role=status and aria-live=polite for warning type" -> "announces warning status politely via the persistent live region".
- `Field.test.tsx` "auto-generates status message ID" -> query changed from `getByRole('alert')` to `getByText('Required')`; same id assertion (the element no longer has a role, and the singleton assertive region *is* a `role="alert"` document-wide).
- `Field.test.tsx` "groups status message with input in column 2" (horizontal-labels) -> same query change, same layout assertion.
- `FileInput.test.tsx` "does not announce a selection when validation rejects all files" -> `politeRegion()` null-check became an empty-text check: the validation error itself now announces assertively via FieldStatus, which lazily creates the (empty) polite region as a pair. Intent preserved: nothing is announced politely.
- `TextArea.test.tsx`, `NumberInput.test.tsx`, `CheckboxList.test.tsx` -> added `afterEach(__resetLiveRegionsForTest)` only (same plumbing MultiSelector/FileInput tests already use). Earlier tests in each file render status messages, which now create the persistent singletons; without the reset, document-wide queries in later tests (`[aria-live="polite"]` for TextArea's counter, `getByRole('alert')` for NumberInput's inline alert, `getByRole('status')` for CheckboxList's spinner) matched a leftover region. No assertions changed. (NumberInput additionally picked up two whitespace-only `render(...)` re-wraps from the repo's lint-staged prettier pass.)

## Notes

Found during a broader a11y audit of core components; scoped to one fix per PR. FileInput's validation errors now reach AT via its own always-mounted `role="status"` region *and* the assertive singleton (previously the second channel was the unreliable alert-insertion path); consolidating FileInput's channels is left for its own follow-up rather than expanding this PR's blast radius.

### DIFF
--- a/packages/core/src/FieldStatus/FieldStatus.tsx
+++ b/packages/core/src/FieldStatus/FieldStatus.tsx
@@ -2,7 +2,7 @@
 
 /**
  * @file FieldStatus.tsx
- * @input Uses React, stylex, theme tokens
+ * @input Uses React, stylex, theme tokens, useAnnounce
  * @output Exports FieldStatus component, FieldStatusProps
  * @position Core implementation; consumed by Field, Switch, CheckboxInput, and the FieldStatus entrypoint
  *
@@ -16,9 +16,10 @@
 
 'use client';
 
-import React from 'react';
+import React, {useEffect} from 'react';
 import * as stylex from '@stylexjs/stylex';
 import type {BaseProps} from '../BaseProps';
+import {useAnnounce} from '../hooks/useAnnounce';
 import {mergeProps} from '../utils';
 import {
   colorVars,
@@ -113,6 +114,14 @@ export interface FieldStatusProps extends BaseProps<HTMLDivElement> {
 /**
  * A status message component for form fields.
  *
+ * Screen-reader announcements go through the persistent `useAnnounce` live
+ * regions (assertive for errors, polite otherwise) rather than `role`/
+ * `aria-live` on the rendered element. Live regions that mount together with
+ * their content are not reliably announced by assistive technology, and
+ * FieldStatus is almost always conditionally rendered by its callers. The
+ * message is announced whenever it appears — including on first mount — and
+ * whenever it changes.
+ *
  * @example
  * ```
  * <FieldStatus
@@ -138,13 +147,22 @@ export function FieldStatus({
   ...rest
 }: FieldStatusProps) {
   const entryStyle = useEntryAnimation('slideDown');
+  const announce = useAnnounce();
+
+  // Announce the message through the persistently-mounted live regions.
+  // Announce-on-mount is intentional: callers conditionally mount FieldStatus
+  // when a status appears (and whole forms can mount with a server-side
+  // validation error already present), and both cases must be heard.
+  useEffect(() => {
+    if (message) {
+      announce(message, type === 'error' ? 'assertive' : 'polite');
+    }
+  }, [announce, message, type]);
 
   return (
     <div
       ref={ref}
       id={id}
-      role={type === 'error' ? 'alert' : 'status'}
-      aria-live={type === 'error' ? 'assertive' : 'polite'}
       {...rest}
       {...mergeProps(
         themeProps('field-status', {type, variant}),
--- a/.changeset/fieldstatus-live-region.md
+++ b/.changeset/fieldstatus-live-region.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Field/FieldStatus: status and error messages are now announced through persistent live regions, so they are reliably read by screen readers regardless of when they appear.
+
+@bhamodi
--- a/packages/core/src/CheckboxInput/CheckboxInput.test.tsx
+++ b/packages/core/src/CheckboxInput/CheckboxInput.test.tsx
@@ -9,10 +9,15 @@
  * SYNC: When CheckboxInput.tsx changes, update tests to match new behavior
  */
 
-import {describe, it, expect, vi, beforeEach} from 'vitest';
+import {describe, it, expect, vi, beforeEach, afterEach} from 'vitest';
 import {render, screen, fireEvent, waitFor} from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import {CheckboxInput} from './CheckboxInput';
+import {__resetLiveRegionsForTest} from '../hooks/useAnnounce';
+
+afterEach(() => {
+  __resetLiveRegionsForTest();
+});
 
 // Mock showPopover/hidePopover (not implemented in jsdom) so the tooltip layer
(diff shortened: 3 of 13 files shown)


packages/core/src/FieldStatus/FieldStatus.tsx
```diff
@@ -138,13 +147,22 @@ export function FieldStatus({
   ...rest
 }: FieldStatusProps) {
   const entryStyle = useEntryAnimation('slideDown');
+  const announce = useAnnounce();
+
+  // Announce the message through the persistently-mounted live regions.
```
cixzhang, line 152: What's the experience like for this when showing multiple form errors at the same time?

cixzhang (approved):

PR #4003, opened 2026-07-15
fix(core): keep hidden-label descriptions in aria-describedby

bhamodi:
## Summary

`CheckboxInput` and `Switch` both built `aria-describedby` with an identical guard: `if (description && !isLabelHidden)`. But `FieldLabel` renders the description span with `id={descriptionID}` regardless of `isLabelHidden` -- when the label is hidden, the description is just sr-only styled (`FieldLabel.tsx:220-226`), not removed from the DOM. So with a hidden label the description text existed in the accessibility tree but was programmatically orphaned from the input: a screen-reader user focusing the control never heard it.

Both components pass `description` and `descriptionID` to `FieldLabel` unconditionally, so the fix is to drop the `!isLabelHidden` condition in both files. The comment above the guard ("Only include descriptionID when the element actually renders") was describing an invariant the code no longer matched -- it is now true again, and extended to explain why the hidden-label case still qualifies.

## Changes
- `CheckboxInput/CheckboxInput.tsx`: include `descriptionID` in `aria-describedby` whenever a `description` is provided, not only when the label is visible; update the comment.
- `Switch/Switch.tsx`: same one-line guard change (the two components had copy-pasted identical logic); update the comment.
- `CheckboxInput/CheckboxInput.test.tsx` / `Switch/Switch.test.tsx`: new test each -- hidden label + description asserts the input's `aria-describedby` contains the id of the element holding the description text. Existing visible-label tests (`aria-describedby` equals the description id exactly) continue to pass unchanged.
- `.changeset/choice-hidden-label-describedby.md`: patch changeset.

## Test plan
- Pre-fix: both new tests **fail** (aria-describedby is `null` with a hidden label), confirming they catch the defect.
- `node_modules/.bin/vitest run --root . packages/core/src/CheckboxInput packages/core/src/Switch` -- **72 pass** (2 files).
- `node_modules/.bin/vitest run --root . packages/core` -- **4583 pass** (203 files, full core suite).

## Notes
Found during a broader a11y audit of core components; scoped to one fix per PR. An in-flight PR is changing the FieldStatus/status rendering blocks in these same two files -- this PR deliberately touches only the describedby guards near the top of each component, so the hunks do not overlap.

### DIFF
--- a/.changeset/choice-hidden-label-describedby.md
+++ b/.changeset/choice-hidden-label-describedby.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] CheckboxInput/Switch: descriptions stay linked via aria-describedby when the label is visually hidden, instead of being orphaned in the DOM.
+
+@bhamodi
--- a/packages/core/src/CheckboxInput/CheckboxInput.test.tsx
+++ b/packages/core/src/CheckboxInput/CheckboxInput.test.tsx
@@ -248,6 +248,22 @@ describe('CheckboxInput', () => {
     expect(screen.getByLabelText('Select row')).toBeInTheDocument();
   });
 
+  it('keeps description linked via aria-describedby when isLabelHidden', () => {
+    render(
+      <CheckboxInput
+        label="Select row"
+        isLabelHidden
+        description="Selects this row for bulk actions"
+        value={false}
+        onChange={() => {}}
+      />,
+    );
+    const checkbox = screen.getByRole('checkbox');
+    const description = screen.getByText('Selects this row for bulk actions');
+    expect(description.id).not.toBe('');
+    expect(checkbox.getAttribute('aria-describedby')).toContain(description.id);
+  });
+
   it('shows label visually by default', () => {
     render(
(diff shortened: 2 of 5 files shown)


humbertovirtudes (approved):
**Approving.** Small, exactly-correct a11y fix.

Confirmed the root cause against `Field/FieldLabel.tsx:220-226`: the description `<span id={descriptionID}>` renders whenever `description` is truthy — with a hidden label it only gets `srOnly` styling, it isn't removed. So the old `if (description && !isLabelHidden)` guard orphaned that description from the input for screen-reader users on hidden-label controls. Dropping `!isLabelHidden` makes the describedby build symmetric with what `FieldLabel` actually renders. Both `description` and `descriptionID` are already passed to `FieldLabel` unconditionally, so the link is restored precisely with no other behavior change.

The two components had copy-pasted identical logic, and both get the same one-line fix — nice to see the guard corrected in lockstep. The updated comment now matches reality.

Tests are the right shape: behavior-tested, assert `aria-describedby` *contains* the description id (tolerant of the status-message id also being present), and per the description both fail pre-fix. No blockers.

PR #4007, opened 2026-07-15
fix(core): make Timestamp tooltip keyboard-reachable

bhamodi:
## Summary

`Timestamp` in relative mode (the default for recent times via `format="auto"`) exposes the full absolute date/time through a `Tooltip` anchored to its `<time>` element with the default `focusTrigger="auto"`. `useTooltip`'s auto trigger only attaches focus listeners when the anchor is naturally focusable (`useTooltip.tsx:385-393`), and a bare `<time>` is not -- so the tooltip could only ever be revealed with a pointer.

Who was affected: the `<time>` element already carries `aria-label={fullAbsoluteText}` in relative mode (the exact same string the tooltip shows), so screen-reader users were already served. The gap was **sighted keyboard users**, who saw "2 hours ago" with no keyboard path to the precise time (WCAG 1.4.13 content-on-focus / 2.1.1 keyboard).

The fix follows the repo's established pattern for non-interactive tooltip anchors: `Tooltip`'s own text-only mode makes its wrapper span a tab stop with `tabIndex={0}` (`Tooltip.tsx:295`), which `useTooltip`'s `isFocusable` check recognizes (it treats any element with a non-negative `tabindex` as focusable), so the existing `auto` focus trigger picks it up. `Timestamp` now does the same: `tabIndex={0}` on the `<time>` element **only while a tooltip is attached** (`hasTooltip` and relative display), plus a `:focus-visible` outline matching the repo-wide focus ring treatment (2px solid accent, 2px offset -- same as Token/Thumbnail). No tooltip, no tab stop; `useTooltip`'s contract is untouched. The `focusTrigger="always"` alternative used by disabled-control tooltips (Slider/Selector `disabledMessage`) was not needed here -- those anchors are focusable-but-disabled widgets; this one just needed to become a tab stop.

## Changes
- `Timestamp/Timestamp.tsx`:
  - add `tabIndex={showTooltip ? 0 : undefined}` to the `<time>` element so the anchor is keyboard-reachable and `useTooltip`'s existing `auto` focus trigger attaches focus listeners; a comment explains why the tab stop is conditional.
  - add a `focusable` StyleX rule (`:focus-visible` 2px accent outline, 2px offset) applied only alongside the tab stop.
- `Timestamp/Timestamp.test.tsx`: new `tooltip keyboard reachability` block (5 tests) with the same jsdom Popover-API + `:focus-visible` mocks used by `Tooltip.test.tsx`/`Slider.test.tsx`.
- `.changeset/timestamp-tooltip-keyboard.md`: patch changeset.

## Test plan
- `node_modules/.bin/vitest run --root . packages/core/src/Timestamp` -- **35 pass** (was 30). New tests:
  - `<time>` has `tabindex="0"` while the tooltip is attached
  - tabbing onto the timestamp shows the tooltip (lazy layer mounts, carries the aria-label string, `showPopover` fires on keyboard focus) -- **failed pre-fix** (`user.tab()` could not land on the non-focusable `<time>`)
  - no `tabindex` with `hasTooltip={false}` -- and the tabindex-present test also **failed pre-fix**
  - no `tabindex` for absolute formats (no tooltip, no gratuitous tab stop)
  - aria-label still the full absolute string while the tooltip is attached
- `node_modules/.bin/vitest run --root . packages/core` -- **4586 pass** (203 files), no pre-existing failures.

## Notes
Found during a broader a11y audit of core components; scoped to one fix per PR (no changes to `useTooltip`'s contract or other Tooltip consumers).

### DIFF
--- a/.changeset/timestamp-tooltip-keyboard.md
+++ b/.changeset/timestamp-tooltip-keyboard.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Timestamp: the absolute-time tooltip is now reachable by keyboard — the timestamp is focusable while a tooltip is attached, per WCAG content-on-hover requirements.
+@bhamodi
--- a/packages/core/src/Timestamp/Timestamp.test.tsx
+++ b/packages/core/src/Timestamp/Timestamp.test.tsx
@@ -1,7 +1,8 @@
 // Copyright (c) Meta Platforms, Inc. and affiliates.
 
 import {describe, it, expect, vi, beforeEach, afterEach} from 'vitest';
-import {render, screen, act} from '@testing-library/react';
+import {render, screen, act, waitFor} from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
 import {Timestamp} from './Timestamp';
 
 describe('Timestamp', () => {
@@ -41,9 +42,7 @@ describe('Timestamp', () => {
     );
     expect(screen.getByText('59 minutes ago')).toBeInTheDocument();
 
-    rerender(
-      <Timestamp value={Date.now() / 1000 - 86399} format="relative" />,
-    );
+    rerender(<Timestamp value={Date.now() / 1000 - 86399} format="relative" />);
     expect(screen.getByText('23 hours ago')).toBeInTheDocument();
 
     rerender(
@@ -52,9 +51,7 @@ describe('Timestamp', () => {
     expect(screen.getByText('29 days ago')).toBeInTheDocument();
 
     // Same guarantee on the future side.
-    rerender(
-      <Timestamp value={Date.now() / 1000 + 3599} format="relative" />,
-    );
+    rerender(<Timestamp value={Date.now() / 1000 + 3599} format="relative" />);
     expect(screen.getByText('in 59 minutes')).toBeInTheDocument();
   });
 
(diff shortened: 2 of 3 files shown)


humbertovirtudes (approved):
**Approving.** Correct, well-tested a11y fix that follows the repo's existing pattern.

The relative-mode tooltip was pointer-only: `useTooltip`'s `auto` focus trigger only attaches focus listeners on focusable anchors (`useTooltip.tsx:209`, `387`), and a bare `<time>` isn't one. Adding `tabIndex={0}` while a tooltip is attached arms the existing focus path with no change to `useTooltip`'s contract. The tab stop is correctly gated on `showTooltip`, so absolute formats and `hasTooltip={false}` add no tab stop. The `tabIndex={0}` + `:focus-visible` ring matches the precedent in `Tooltip.tsx:295` and `Token.tsx`.

Tests are solid — the key case (tabbing onto the timestamp reveals the tooltip) is behavior-tested and failed pre-fix.

Two non-blocking notes:

- Every relative `Timestamp` now becomes a tab stop. On a dense surface (a feed or table with many "2 hours ago" rows) that adds a lot of tab stops. It's the correct WCAG 1.4.13 remedy and matches the Tooltip text-only behavior, so not a blocker — just worth awareness for high-density lists.
- The test file picks up two unrelated Prettier reformats of `rerender(...)` calls. Harmless, just outside the scope of the fix.

PR #4009, opened 2026-07-15
fix(core): drop live href and onClick from disabled Links

bhamodi:
## Summary

A disabled `Link` (`isDisabled` with an `href`) already carried the right disabled *signaling* -- `aria-disabled="true"`, `tabIndex={-1}`, and CSS `pointer-events: none` were all in place. The gap was that the rendered element still received a **live `href` and `onClick`**: `Link.tsx`'s anchor branch passed `href={href}` and `onClick={onClick}` to the resolved LinkComponent unconditionally. Pointer clicks were blocked by CSS and tab order excluded the element, but programmatic focus + Enter, some AT activation commands, and middle-click paste-navigation bypass all three guards and could still navigate.

The fix makes a disabled link actually inert: when `isDisabled`, `Link` renders a **plain `<a>` with no `href`** (and no `target`/`rel`, which are meaningless without one), with `onClick` replaced by a `preventDefault` no-op. An anchor without an href is not focusable and exposes no link affordance to activate -- the standard disabled-link treatment. `aria-disabled="true"`, `tabIndex={-1}`, the visual classes, and the content (including the external-link icon) are all preserved.

The router LinkComponent (from `as` / `LinkProvider`) is deliberately **not** used for the disabled state: a disabled link performs no navigation so it needs no routing integration, and custom router links (Next.js et al.) may require a live `href` -- passing `href={undefined}` into them could crash consumers.

## Changes
- `Link/Link.tsx`:
  - new `else if (isDisabled)` branch before the LinkComponent branch: renders a plain `<a>` with no `href`/`target`/`rel`, `onClick` set to a module-level `preventDefaultClick` guard, keeping `aria-disabled="true"`, `tabIndex={-1}`, and the full style stack (`styles.disabled` now applied unconditionally in this branch).
  - documented the behavior on the `isDisabled` prop JSDoc.
  - the enabled anchor path and the `<button>` fallback path (no href, or disabled href + onClick, which already renders a natively-disabled `<button>`) are untouched.
- `Link/Link.test.tsx`:
  - updated `applies isDisabled state correctly` to query by text (`closest('a')`) since an href-less anchor has no implicit `link` role; still asserts `aria-disabled="true"`, `tabIndex="-1"`, and the `astryx-link` class.
  - new tests: disabled link has no `href`; clicking a disabled link cancels default navigation (`fireEvent.click` returns `false`, simulating programmatic/AT activation that bypasses pointer-events); disabled link with `onClick` never fires the consumer handler; disabled link omits `target`/`rel` even with `isExternalLink`; disabled link renders a plain anchor rather than the custom LinkComponent (both `as` and `LinkProvider`).

## Test plan
- `node_modules/.bin/vitest run --root . packages/core/src/Link` -- **60 pass** (Link.test.tsx 34, useLinkComponent.test.tsx, useLinkify.test.tsx). Pre-fix, the 4 new positive tests fail exactly as expected: disabled link had `href="/test"`, `target="_blank"`/`rel` present, `data-custom-link` custom component rendered, click not cancelled.
- `node_modules/.bin/vitest run --root . packages/core` -- **4586 pass / 0 fail** (203 files; 4581 upstream + 5 from this PR). A clean checkout of origin/main (93d881be6) also passes 4581/4581, so no pre-existing failures to report.

## Notes

Behavior note: disabled links now deliberately render a **plain `<a>`, not the router LinkComponent**, and drop `href`/`target`/`rel` entirely. Anyone styling or querying disabled links via `a[href]` selectors, or expecting their custom LinkComponent to render in the disabled state, will see the new (inert) markup. Enabled links are byte-for-byte unchanged, including the custom `as`/`LinkProvider` path.

### DIFF
--- a/.changeset/link-disabled-inert.md
+++ b/.changeset/link-disabled-inert.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Link: disabled links no longer carry a live href/onClick — programmatic focus or AT activation can no longer trigger navigation.
+@bhamodi
--- a/packages/core/src/Link/Link.test.tsx
+++ b/packages/core/src/Link/Link.test.tsx
@@ -10,7 +10,7 @@
  */
 
 import {describe, it, expect, vi} from 'vitest';
-import {render, screen} from '@testing-library/react';
+import {fireEvent, render, screen} from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import {Link} from './Link';
 import {LinkProvider} from './LinkProvider';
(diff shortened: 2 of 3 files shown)


humbertovirtudes (approved):
**Approving.** Precise, well-tested a11y fix — it closes a real navigation-bypass hole (programmatic focus + Enter, AT activation commands, middle-click paste-navigation) that CSS `pointer-events: none` + `tabIndex={-1}` didn't actually cover.

**Correctness ✓** — traced all four disabled paths through `useInteractiveRole` + `renderAsButton`:
- disabled + href only → `role='inert'`, `renderAsButton=false` → the new plain-`<a>` branch (the bug case, now inert)
- disabled + onClick → `role='button'` → `<button disabled>` (unchanged)
- disabled + no href + no onClick → `<button disabled>` (unchanged)
- disabled inside an interactive context → `<button disabled>` (unchanged)

A disabled link can no longer reach the enabled `LinkComponent` branch, and the other three cases stay byte-for-byte. `preventDefaultClick` being module-level (not an inline arrow) avoids a per-render allocation — nice touch.

**Tests ✓** — the six cases each pin a distinct guarantee (no `href`, click cancels default, consumer `onClick` never fires, no `target`/`rel`, custom `LinkComponent` skipped via both `as` and `LinkProvider`). The click test correctly keys off `preventDefault` via `fireEvent.click`'s return value, and it's behavior-tested rather than implementation-tested.

A couple of non-blocking notes:

- **(minor, a11y)** For a disabled *external* link, `sharedContent` still renders the external-link icon and the `VisuallyHidden` "(opens in new tab)" announcement. Since the disabled link now opens nothing, that announcement promises navigation that won't happen — worth suppressing the "(opens in new tab)" text in the disabled state (the icon itself is arguably fine as a visual affordance). Fitting to tighten given this is an a11y PR.
- **(minor, semver)** The changeset is `patch`. The disabled state now drops the custom `LinkComponent` and renders href-less markup, which is consumer-visible for anyone querying `a[href]` or relying on their router link in the disabled state. It's a bug fix, so `patch` is defensible — just flagging for maintainer judgment on whether the markup change warrants `minor`.
- **(nit)** The `mergeProps(themeProps, stylex.props(...))` prop stack is now duplicated across three branches (button / disabled-`<a>` / enabled-`<a>`). Pre-existing pattern, so it matches house style — not worth extracting in this PR.

Tightly scoped, does one thing, and proves it.

PR #4021, opened 2026-07-16
fix Calendar duplicate today indicator

### LINKED ISSUE #4020: Calendar marks an outside day as today in multi-month views
## Description

In a two-month Calendar, a date from the adjacent month can be rendered twice: once as an outside day in the first month and once as an in-month day in the second month.

An **outside day** is a date displayed in a month grid even though it belongs to the adjacent month, allowing the grid to show complete weeks. For example, the June grid can show a dimmed 1 July at the end, while the July grid also shows its normal 1 July.

When that date is today, both cells receive the today state, producing duplicate today indicators and duplicate `aria-current="date"` values.

## Reproduction

1. Render a Calendar with two months.
2. Use a month boundary where today appears in the trailing week of the first month and in the second month.
3. Inspect the two cells representing today.

## Current behavior

Both the outside cell and the in-month cell are marked as today.

### Before

Two-month Calendar showing duplicate today indicators

## Expected behavior

Only the in-month cell is marked as today.

### After

Two-month Calendar showing a single today indicator

## Verification

The screenshots use a deterministic June–July 2026 Storybook reproduction with 1 July as today.

Kevinjohn:
## What

A two-month Calendar can render the same date twice: once as an outside day in the first month and once as an in-month day in the second month. An outside day belongs to the adjacent month but is displayed to complete the current month's week.

When that date is today, `computeDayCellState` currently marks both cells as `isToday`. This produces two today rings and two `aria-current="date"` values.

This adds the same `!isOutside` guard already used by the other date states, plus a regression test. No public API changes.

Fixes #4020

## Before

Both copies of 1 July are marked as today: the dimmed outside day in June and the in-month day in July.

Two-month Calendar showing duplicate today indicators

## After

Only the in-month 1 July is marked as today.

Two-month Calendar showing a single today indicator

## Validation

- `dayCellUtils.test.ts`: 42/42 tests pass
- `pnpm lint`: passes (two pre-existing warnings outside this change)
- Full `pnpm test`: 6,667/6,668 tests pass; 
- - the unrelated `packages/cli/src/commands/build-theme.watch.test.mjs` filesystem-watch test fails identically when run alone
- Deterministic Storybook reproduction: June–July 2026 calendar, with 1st July set as today

## Checklist

- [x] Opened a bug issue before the core change (#4020)
- [x] Added a colocated regression test
- [x] Added a patch changeset for `@astryxdesign/core`
- [x] Ran the repository linter and tests locally
- [x] Included before/after Storybook evidence
- [x] Kept the change tightly scoped with no public API changes

### DIFF
--- a/.changeset/calendar-only-marks-in-month-date-cells-as-today-scd6.md
+++ b/.changeset/calendar-only-marks-in-month-date-cells-as-today-scd6.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Calendar only marks in-month date cells as today, preventing duplicate today indicators in multi-month views.
+@Kevinjohn
--- a/packages/core/src/Calendar/dayCellUtils.test.ts
+++ b/packages/core/src/Calendar/dayCellUtils.test.ts
@@ -45,6 +45,11 @@ describe('computeDayCellState', () => {
     expect(state.isToday).toBe(false);
   });
 
+  it('does not identify an outside day as today', () => {
+    const state = computeDayCellState(makeInput({isOutside: true}));
+    expect(state.isToday).toBe(false);
+  });
+
   it('identifies selected day in single mode', () => {
     const state = computeDayCellState(
       makeInput({selectedDate: plainDateFromISO('2024-03-15')}),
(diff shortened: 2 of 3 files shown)


Kevinjohn:
I've purposefully not changed the obvious accessibility (colour contrast) issues with this PR.
Just the code to confim what Today is - form a code & accessibility standpoint.

cixzhang (approved):
Thanks for this fix!

PR #3797, opened 2026-07-11
fix(core): select by typing on a focused Selector (#3764)

AKnassa:
Fixes #3764

## What this does

You can now pick an option in a Selector just by typing. Focus the dropdown with Tab, press "C", and it selects "CA" — no need to open the menu and hunt for the option first. This is how the browser's own dropdowns have always worked, and Selector now matches that, so it can genuinely replace them.

## Why

The person who reported this was building a state picker. They could tab to it, but then had to stop and interact with the menu to choose anything — which makes a form slow to fill in by keyboard and made Selector a poor swap for a native dropdown.

## What changed

- **Type to pick.** Typing a letter on a closed Selector selects the first matching option. The menu stays shut, just like a native dropdown.
- **Press again to cycle.** If several options start with the same letter, pressing it again moves to the next one — Chicago, then Cleveland, then Columbus, then back around.
- **Spaces work.** Typing "new y" gets you to "New York" instead of the space opening the menu. The typed text is forgotten after a short pause, so the next thing you type starts fresh.
- **Screen readers hear it.** Because the value changes without the menu opening or focus moving, nothing would otherwise announce the new choice — so it's now announced politely.
- **Searchable Selectors too.** With a search box enabled, typing on the closed dropdown opens it and drops what you typed straight into the search field, without losing any characters.
- **Selector-only by design.** The matching lives in Selector's own use of the shared `useTypeahead` hook, not in the shared combobox. CommandPalette and MultiSelector get nothing extra, and the palette is already a search input, so it needs no opt-out.

Menu-open behavior is unchanged for the user: typing walks the highlight and Enter confirms. The same shared hook now drives it, moving the highlight via `aria-activedescendant`.

## How to see it

In Storybook, open **Core → Selector → Default**, click once outside the dropdown, then press Tab to focus it and type a letter. The value changes with the menu closed.

## Notes for reviewers

- Typeahead is a Selector-level composition of the shared `useTypeahead` hook, run ahead of the combobox key handler. Nothing was added to `useCombobox`, so CommandPalette and MultiSelector are untouched.
- Announcements use `useAnnounce` rather than local live-region state.
- Adopting the shared hook surfaced and fixed three bugs in it: a single-character search now starts *after* the current item (native `<select>` / APG `getIndexByLetter`); a `-1` start no longer wraps onto the last item; and a ctrl/meta-chorded Space no longer appends a raw space that poisons the buffer. DropdownMenu, ContextMenu, and NavHeadingMenu share the hook and benefit.
[...]

PR #4023, opened 2026-07-16
fix: Selector keyboard typeahead now selects directly when closed (#3764)

### LINKED ISSUE #3764: [Bug] Selector component doesn't allow keyboard to select by typing
### Description

If I focus a Selector, I can't just press a key to select one of the items. For example if I have a state picker, I can't tab into it and press "C" to select "CA". I have to actually interact with the menu, which is annoying. This makes it hard to replace native selectors fully.

### Reproduction

1. Render <Selector> with some options
2. Tab into it so it's focused
3. Press a character that matches one of the options
4. I would expect that option to be selected

### Astryx Version

@astryxdesign/core@0.14

### Environment

OS X

gonzoblasco:
## What

Fixes #3764 — the Selector component didn't allow keyboard type-to-select when closed. Pressing a character that matches an option would open the menu and highlight the match, but not select it. This made it impossible to fully replace native `<select>` elements.

## Change

When the Selector is **closed** and the user types a character matching an option, the component now **selects that option directly** (native `<select>` behavior) instead of opening the menu.

When the Selector is **open**, the existing behavior is preserved: the matched option is highlighted for the user to confirm with Enter.

## Validation

- 46/46 Selector tests pass
- No public API changes
- Single-line logic change in `useCombobox`

### DIFF
--- a/packages/core/src/Selector/hooks.ts
+++ b/packages/core/src/Selector/hooks.ts
@@ -350,9 +350,15 @@ export function useCombobox({
             );
             if (matchIndex >= 0) {
               if (!isOpen) {
-                onOpen();
+                // When closed, select the matched item directly (native
+                // <select> behavior) instead of opening the menu.
+                const item = selectableItems[matchIndex];
+                if (item && !item.disabled) {
+                  selectItem(item);
+                }
+              } else {
+                setHighlightedIndex(matchIndex);
               }
-              setHighlightedIndex(matchIndex);
             }
           }
           break;


cixzhang:
Thanks for this, and sorry for the slow turn. Closing because #3797 has just merged and covers the same issue (#3764) more completely — the shared `useTypeahead` hook, `useAnnounce` for the selection announcement, cycling on repeated letters, chord-Space handling, and tests across Selector/DropdownMenu/CommandPalette.

One design note in case it's useful: your version placed the fix inside `useCombobox`, which would also have handed typeahead to CommandPalette — and the palette already has a real search input, so prefix-matching there fights the search box. That's why the merged fix lives in Selector instead.

Your instinct about the closed-state behavior was right. Please do keep them coming — I see #4605 and #4298 are still open and I'll get to those.

gonzoblasco:
Thanks for the thoughtful review @cixzhang, and no worries on the turnaround.

You're right - putting the fix inside useCombobox would have handed typeahead to CommandPalette, where it fights the existing search input. The shared useTypeahead hook living in Selector is the cleaner home, and reusing the existing abstraction instead of duplicating the logic is clearly the better call. I hadn't fully accounted for that cross-component impact.

Glad the closed-state behavior landed in the merged fix. I'll take the same care with the architecture on #4605 and #4298 when you get to them.

Thanks again for the context. It's genuinely useful for how I'll approach the next ones.

PR #4024, opened 2026-07-16
feat(cli): discover integration package showcases from templates root

ejhammond:
## Summary

Integration packages (via their `astryx.integration` manifest) can now provide component **showcases** through their existing `templates` root — no separate `astryx.blocks` field and no consumer-side registry. A `type: 'block'` doc marked `isShowcase: true` becomes the CLI-served preview for the components it lists in `componentsUsed`, giving external packages the same CLI-served showcase model as core.

## What changed

- **`template.mjs`**: `discoverShowcaseBlocks` folds integration `type:'block'` templates into showcase/example resolution (`findShowcase` / `findRelatedBlocks`) without double-counting them in scaffolding discovery. Integration template discovery now carries `isShowcase` / `aspectRatio`, and `BaseTemplateSchema` accepts them.
- **`component.mjs`**: `component(name, { showcase })` now returns `component.detail.showcase` for **integration-owned** components — both when scoped with `--package` and when a single non-core integration owns the name. Previously these branches fell through to `component.detail` (scoped) or threw `ERR_NO_SHOWCASE` (single-owner), so integration showcases were *discoverable* but never *served*.
- Split families (e.g. a base component and a related popover sharing one showcase) resolve via `componentsUsed`.

## Test plan

- New `integration-showcase.test.mjs` — discovery via `findShowcase` / `findRelatedBlocks` for integration packages, including the split-family case.
- Extended `component-ownership.test.mjs` — the `component()` API showcase path (scoped, unscoped single-owner, and `ERR_NO_SHOWCASE` when absent).
- Full CLI suite green; eslint clean.

### DIFF
--- a/.changeset/integration-showcase-discovery.md
+++ b/.changeset/integration-showcase-discovery.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/cli': patch
+---
+
+[feature] Integration packages can now provide component showcases. Blocks authored under an integration's `templates` root (with `isShowcase: true`) are discovered by `component <Name> --showcase` and `--blocks`, so external packages get the same CLI-served preview model as core — no separate `astryx.blocks` field needed. Split families (e.g. a Link + HoverCard sharing one showcase) resolve via `componentsUsed`.
+
+@ejhammond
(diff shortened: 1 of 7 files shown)


ejhammond:
Closing in favor of a broader rework of the showcase + template model rather than landing this interim approach.

The follow-up converges core and integration template authoring onto one typed factory, makes the component→showcase link explicit (a component names its showcase, instead of templates self-declaring + heuristic matching), and renders static previews without eval. That work supersedes the discovery changes here, so landing this first would just be churn we'd migrate away from.

The branch is preserved for reference; no functional regression from closing — current previews are unaffected.

PR #4026, opened 2026-07-16
fix: parse compact time meridiems

### LINKED ISSUE #4025: TimeInput drops PM in compact time input
## Why

Typing a compact 12-hour time with PM gives the wrong time.

## What

In Storybook, open **Core / TimeInput / Default** and type:

`0230pm`

## What happens

The field shows **2:30 AM**.

## What should happen

The field should show **2:30 PM**.

The same problem affects compact times with seconds, such as `023000pm`.

## Screenshot

Before: 0230pm shows 2:30 AM

## Checks

- [x] Reproduced in Storybook
- [x] Added a regression test
- [x] Fixed compact AM/PM parsing

Kevinjohn:
## Why

Typing a compact 12-hour time with PM gives the wrong time.

## What

In Storybook, type `0230pm` in **Core / TimeInput / Default**.

Before this fix, it showed **2:30 AM**.

Fixes #4025

## Before — bug

The field shows **2:30 AM**.

Before: 0230pm shows 2:30 AM

## After — fixed

The field shows **2:30 PM**.

After: 0230pm shows 2:30 PM

## What we fixed

- Compact `HHMM` input now applies AM/PM.
- Compact `HHMMSS` input now applies AM/PM.
- Invalid 12-hour compact values are rejected.
- Added regression tests.

## Checklist

- [x] Read `CONTRIBUTING.md`
- [x] Created a branch from `main`
- [x] Reproduced the issue in Storybook
- [x] Added before-and-after Storybook screenshots
- [x] Added regression tests
- [x] Added a changeset
- [x] Built `@astryxdesign/core`
- [x] Built Storybook
- [x] `pnpm lint` passes with two existing warnings
- [x] Focused time-parser tests pass: 28/28
- [ ] Full `pnpm test`: one unrelated watch-mode test fails in `build-theme.watch.test.mjs`; 342 other files and 6,666 tests pass


Kevinjohn:
(updated screenshot)

cixzhang (commented):
Thank you for this — the compact meridiem handling is a real gap, and pulling the 12-hour→24-hour conversion into a single `convertHourTo24Hour` helper (shared across the bare-hour, HHMM, and HHMMSS paths) is a clean way to fix it consistently. The parsing logic and tests look good.

One small housekeeping note before this lands: could you drop the two PNGs in `docs/screenshots/`? Before/after evidence like that is really helpful for the review — but it's perfect as an attachment on the PR itself rather than committed into the repo, so the tree stays free of binary artifacts. Once those are out I'm happy to get this in. Thanks again!

cixzhang (approved):
Dropped the two PNGs and merged main in for you, so this is ready. The one helper across all three compact paths matches what the colon path already did.

[Reviewed by Robohands]

PR #4032, opened 2026-07-17
feat(core): customize MultiSelector count labels

### LINKED ISSUE #3348: MultiSelector: 'count' trigger text is hard-coded '{N} selected' — no way to render domain nouns ('3 spaces selected')
## Problem

The `'count'` triggerDisplay renders a hard-coded `${count} selected` (packages/core/src/MultiSelector/MultiSelector.tsx ~line 866):

```tsx
case 'count':
  return (
    <span {...stylex.props(styles.triggerText)}>
      {optimisticValue.length} selected
    </span>
  );
```

There is no way to render a domain noun in the trigger — "3 spaces selected", "2 columns", "5 filters".

Evidence of demand: a builder using Astryx for an internal tool (Meridian) needed "3 spaces selected" with no visible label, found no prop for the trigger text, and **rebuilt the whole control from Popover + CheckboxInput** — the highest-cost failure mode for the system. (`isLabelHidden` already covers the no-label half of that requirement; the trigger text was the blocker.)

## Why this is a props gap

`renderOption` already establishes function-prop customization on this exact component ("custom render fn per selectable option"). The count text is presentation, not data — consumers shouldn't have to fork the component (or rebuild it from primitives) to change a display string.

## API arbitration (compact round, 2026-07-01)

Candidates tested with naive sonnet generations plus a haiku hallucination probe:

| Candidate | Invented props | Escape hatches | Notes |
|---|---|---|---|
| `formatTriggerCount: (count: number) => string` | 0 | 0 | exact output |
| `itemNoun: string` | 0 | 0 | can't cover singular/plural or i18n |
| `triggerContent` render-fn | 0 | 0 | widest surface; overlaps labels/badges modes |

Hallucination probe against the current API invented `countLabel` (string) rather than rebuilding — signal that a small prop is discoverable enough.

**Verdict: `formatTriggerCount`** — equal discoverability to the string prop, uniquely covers singular/plural and i18n. Runner-up `itemNoun` listed as an open question.

## Proposed API

```tsx
interface MultiSelectorProps<T> {
  /**
   * Formats the trigger text when triggerDisplay is 'count'.
   * Receives the number of selected values; return the full trigger string.
   * @default (count) => `${count} selected`
   */
  formatTriggerCount?: (count: number) => string;
}
```

```tsx
<MultiSelector
  label="Spaces"
  isLabelHidden
  options={spaces}
  value={selected}
  onChange={setSelected}
  formatTriggerCount={(count) =>
    count === 1 ? '1 space selected' : `${count} spaces selected`
  }
/>
```

- Default behavior unchanged: `${count} selected`.
- Applies only when `triggerDisplay='count'` (the default); `labels` and `badges` modes are untouched.

## Open questions

- Should a string convenience prop (`itemNoun`) also exist for the 90% case?

Related to #3324 (same trigger code paths).

Kevinjohn:
## Summary

- add the agreed `formatTriggerCount` callback for `MultiSelector` count-mode trigger text
- preserve the existing `{N} selected` output when the callback is omitted
- document the prop in the full and dense references; the partial Chinese overlay inherits the English description
- add regression coverage and a patch changeset

Fixes #3348

## Red/green verification

The regression test was run against the unchanged implementation before the fix:

```text
Test Files  1 failed (1)
Tests       1 failed | 57 passed (58)

Unable to find an element with the text: 2 spaces selected
Rendered text: 2 selected
```

The identical command after the implementation passes:

```text
Test Files  1 passed (1)
Tests       58 passed (58)
```

Command:

```sh
pnpm exec vitest run packages/core/src/MultiSelector/MultiSelector.test.tsx
```

## Additional validation

- `pnpm -F @astryxdesign/core build` — passes
- `pnpm lint` — passes with two existing warnings outside this change
- `pnpm test` — 6,673 tests pass; the unrelated `packages/cli/src/commands/build-theme.watch.test.mjs` watcher test fails at its rebuild assertion and also fails when run alone (2 passed, 1 failed)


Kevinjohn:
My fix for @thedjpetersen 's 
[MultiSelector: 'count' trigger text is hard-coded '{N} selected' — no way to render domain nouns ('3 spaces selected')
 #3348](https://github.com/facebook/astryx/issues/3348)

cixzhang (approved):
Thanks — merged main in and re-ran CI. The shape matches the arbitration on [#3348](https://github.com/facebook/astryx/issues/3348), and the test covers it.

[Reviewed by Robohands]

cixzhang:
Merged to get the fix in — the hardcoded `${n} selected` was a real gap and this closes it.

Flagging for ourselves rather than for you: the name and shape should line up with the rest of the family before the next release. Today we have `NumberInput.formatValue`, `Slider.formatValue`, `ProgressBar.formatValueLabel` and now `MultiSelector.formatTriggerCount` — four spellings for one idea. Worth also deciding whether this belongs in i18n instead: the result count two lines away goes through `t('@astryx.multiSelector.resultCount', {count})` with ICU plurals, and a callback asks every caller to hand-roll plural rules that a message key already solves.

Nothing for you to do — the behaviour stays. We'll land the naming change with a codemod if it moves.

Thanks for catching it.

[Reviewed by Robohands]

PR #4036, opened 2026-07-17
fix(table): clean up selection row subscriptions

### LINKED ISSUE #3972: Table selection plugin: row-ref subscription leak (refs never unsubscribe on detach)

**Feature area:** Selection & interaction

### Problem

The selection plugin attaches a ref callback to each `<tr>` that subscribes to the selection store to apply row styling imperatively, but the ref **never returns a cleanup**, so the subscription is never removed when the row detaches. Because toggling selection shifts row indices and re-renders rows, the listener set grows by one per row re-render — a slow leak that also means stale rows keep receiving notifications.

Spotted by @AKnassa while mirroring the selection architecture in #3789 (whose tree plugin returns ref cleanups precisely to avoid this).

### Fix

Return a cleanup from the row ref callback so React unsubscribes on detach:

```ts
const rowRef: React.RefCallback<HTMLTableRowElement> = el => {
  if (!el) return;
  apply();
  const unsub = store.subscribe(apply);
  return () => unsub(); // <-- missing today
};
```

Add a test that mounts, toggles selection several times, and asserts the store's listener count returns to baseline (doesn't grow per toggle).

_Surfaced during review of #3789._

jiunshinn:
Closes #3972

## Summary

- return the selection row ref cleanup so React unsubscribes the detached callback ref
- add a regression test that repeats row-ref detach/attach cycles and verifies only the active listener is notified
- verify the final detach removes the listener entirely

## Root cause

Each body-row render creates a new callback ref and subscribes it to the selection store. The callback previously returned no cleanup, so replacing the ref left the previous listener in the store while the row DOM node remained connected.

## Test plan

- pnpm vitest run packages/core/src/Table/plugins/selection/useTableSelection.test.tsx packages/core/src/Table/plugins/selection/useTableSelection-perf.test.tsx --project ui (17 passed)
- pnpm vitest run packages/core/src/Table/ --project ui (359 passed)
- pnpm --filter @astryxdesign/core typecheck
- pnpm exec eslint packages/core/src/Table/plugins/selection/useTableSelection.tsx packages/core/src/Table/plugins/selection/useTableSelection.test.tsx
- pnpm exec prettier --check packages/core/src/Table/plugins/selection/useTableSelection.tsx packages/core/src/Table/plugins/selection/useTableSelection.test.tsx

### DIFF
--- a/packages/core/src/Table/plugins/selection/useTableSelection.test.tsx
+++ b/packages/core/src/Table/plugins/selection/useTableSelection.test.tsx
@@ -7,13 +7,13 @@
  * @position Test file; validates selection behavior (checkboxes, aria, select-all)
  */
 
-import {describe, it, expect} from 'vitest';
+import {describe, it, expect, vi} from 'vitest';
 import {useState} from 'react';
-import {render, screen, within} from '@testing-library/react';
+import {act, render, renderHook, screen, within} from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import {Table} from '../../Table';
 import {useTableSelection} from './useTableSelection';
-import type {TableColumn} from '../../types';
+import type {BodyRowRenderProps, TableColumn} from '../../types';
 
 // =============================================================================
 // Test Data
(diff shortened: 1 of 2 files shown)


cixzhang (approved):
Nice catch

PR #4040, opened 2026-07-17
feat(cli): add createDoc factory + schema + jiti loader for component docs

ejhammond:
## Summary

Introduces `createDoc()` as the canonical component-doc factory, a Zod `ComponentDocSchema`, and a jiti-based loader so component docs can be authored in `.ts`. This brings component docs into the same factory family that config / integration / template already use — a tiny stamp-only identity function plus schema validation at the load boundary.

**Foundational and additive.** Existing loose `.doc.mjs` docs (`export const docs = {...}`) keep loading unchanged. No component-resolution behavior changes here; a later task migrates the real docs.

## What's included

- **`packages/cli/src/doc.mjs`**
  - `createDoc(def)` — stamp-only typed-identity helper (mirrors `createConfig` / `createIntegration`; returns its argument unchanged, no runtime validation).
  - `ComponentDocSchema` — Zod union over the three authored shapes (single-component, multi-component, sub-component). Validates the fields that gate discovery/rendering (`name` required; prop `name`/`type`/`description`) and passes the rest through, so it accepts the current loose docs without change. Includes a reserved optional `showcase` passthrough (unconsumed here; a later task consumes it).
- **`packages/cli/src/types/doc.d.ts`** — exported TS surface for the `createDoc` input (`AstryxComponentDoc`), mirroring `integration.d.ts` / `template-api.d.ts`.
- **`packages/cli/package.json`** — adds the `@astryxdesign/cli/doc` subpath export.
- **`packages/cli/src/lib/component-loader.mjs`** — adds `loadComponentDoc()` (the typed counterpart to `loadDocs`): loads `.doc.ts` via jiti and `.doc.mjs`/`.doc.js` via native import, accepts BOTH the loose named `docs`/`doc` export AND the factory default export (prefers default), validates through `loadModuleWithSchema`-style `ComponentDocSchema` + `formatZodError`, and merges translations exactly as `loadDocs` does. Existing `loadDocs` is untouched.
- Unit tests (`doc.test.mjs`) mirroring `config.test.mjs` / `codemod.test.mjs`: stamp-only identity, schema accept/reject (readable messages), `pickDocExport` precedence, and a real `.doc.ts` load via jiti.
- Changeset (`@astryxdesign/cli` patch).

## Schema design — enumerated real doc-field set

Enumerated the actual top-level fields of the `docs` export across all 196 `packages/core/src/**/*.doc.mjs` files (shape split: 73 sub, 58 single, 41 multi, 24 hook). Fields observed:

`name, displayName, description, group, category, keywords, isHiddenFromOverview, hidden, hiddenComponents, usage, props, components, subComponentOf, playground, theming, examples, params, returns, relatedComponents, relatedHooks, importPath`

Because that surface is large, mixed (Component + Hook + Reference docs share the `docs` export name), and still evolving, the schema validates the discriminating/required fields precisely and uses `.passthrough()` for the rest — the additive-safe choice that will not reject any current doc. (The sibling config/integration/template schemas use `.strict()` over small fully-owned surfaces; a component doc is neither small nor fully-owned.)

## Testing

- `npx vitest run packages/cli/src/doc.test.mjs` — 16/16 pass (incl. the jiti `.doc.ts` load).
- `npx vitest run packages/cli/src/` — **108 files, 1708 tests, all pass**.
- `eslint` clean on all changed files.

### DIFF
--- a/.changeset/cli-create-doc-factory.md
+++ b/.changeset/cli-create-doc-factory.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/cli': patch
+---
+
+[feat] Add `createDoc()` component-doc factory, a Zod `ComponentDocSchema`, and a jiti-based loader so component docs can be authored in `.ts` with editor/type feedback and validated at the load boundary. Docs authored with `createDoc(...)` use a default export — the same single-export convention as config/integration/template. Additive: the existing loose doc loader is untouched, so current docs keep loading unchanged during migration.
+@ejhammond
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -47,6 +47,10 @@
       "types": "./src/types/integration.d.ts",
       "import": "./src/integration.mjs"
     },
+    "./doc": {
+      "types": "./src/types/doc.d.ts",
+      "import": "./src/doc.mjs"
+    },
     "./template": {
       "types": "./src/types/template-api.d.ts",
       "import": "./src/template.mjs"
--- a/packages/cli/src/doc.mjs
+++ b/packages/cli/src/doc.mjs
@@ -1,0 +1,140 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+/**
+ * @file Public component-doc authoring API.
+ *
+ * A component doc describes a single component (or a family: a directory that
+ * exports several related components) so the CLI and docs surfaces can list,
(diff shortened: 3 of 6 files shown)


ejhammond:
Superseded by #4045, which finalizes the doc-authoring API (component/function/generic factories, both-format support, derived import paths) on a branch rebased onto current main. The single-createDoc commit here is subsumed by #4045.

PR #4044, opened 2026-07-17
feat(cli): self-check in the agent-init prompt to stop CSS/Tailwind drift

josephfarina:
## TL;DR

**One line added to the `astryx init` agent prompt.** Agents already use Astryx well on simple screens, but on complex, multi-step UIs they drift into raw CSS/Tailwind. This adds a post-generation **SELF-CHECK**: "re-read your output; if you find raw CSS/Tailwind/hardcoded values, replace them with Astryx components + tokens."

Measured across ~110 hard-task runs (`cursor-agent`, Opus 4.8 1M Max Fast): the self-check **cut the drift roughly 4×** — control left raw CSS/Tailwind in the final file in **~11–13% of hard runs** vs **~0–3%** with the self-check — while raising purity and adding **~no extra tokens**, with no regression on simple tasks. Everything else in this PR is the vibe-test harness + data that proves it.

---

## What

Adds a short, styling-system-aware **SELF-CHECK** step to the prompt that `astryx init` injects into a consumer's `AGENTS.md` / `CLAUDE.md` (the block emitted by `generateCompressedIndex()` in `packages/cli/src/commands/agent-docs.mjs`).

After an agent generates UI, it now re-reads its own output and replaces any raw CSS / Tailwind / inline styles / hardcoded values with Astryx components + tokens — verifying via the CLI when unsure — before finishing.

## Why

The injected prompt's existing guardrails are all **preventive** ("No `<div>`", "No raw hex/px"). They work well on simple screens, but on **complex, multi-step UIs** agents still drift into hand-rolled CSS/Tailwind — the "veering" we've been seeing. There was no *post-generation* correction step. This adds one.

## The change (what lands in agent docs)

stylex path:

```
- SELF-CHECK before you finish: re-read the file. If you see className=, style={{…}}, raw <div>/<span> for layout, an imported .css/@apply, or hardcoded #hex/px, you veered off Astryx — replace each with the component or a token. If unsure a component exists, run `astryx search "<thing>"` / `astryx component <Name>`; don't hand-roll CSS.
```

It is **styling-system aware**: the Tailwind variant drops `className=` (utilities *are* the system there), and the CSS-fallback variant flags non-token values — so the self-check never contradicts the styling guidance above it.

## Evidence

Measured with a new isolated vibe-test harness (`internal/vibe-tests/prompt-purity-test/`, included). A **context-free agent** (`cursor-agent`, **Opus 4.8 1M Max Fast**) builds real UIs in a throwaway sandbox whose *only* guidance is the prompt variant under test. We capture the full trajectory (every intermediate file edit) and classify each run:

- **`veeredUncaught`** — raw CSS/Tailwind still present in the *final* file (the bad outcome)
- **`neverVeered` / `caught`** — stayed pure / veered mid-run then self-corrected
- **purity** (0–100), plus exact token usage per run

Variants: **A** = current prompt (control) · **B** = + terse self-check · **C** = + forceful checklist self-check.

### Simple tasks (`dd-1`, `tc-6`, K=3) — the base prompt is already fine

| variant | veeredUncaught | neverVeered | purity |
| --- | --- | --- | --- |
| A-control | 0/6 | 83% | 90 |
| B-selfcheck | 0/6 | 100% | 90 |
| C-strong | 0/6 | 100% | 90 |

No differentiation — the existing rules handle simple UIs.

### Hard tasks (`wd-4`, `dd-5`, `wd-1`) — where drift happens

**Focused run (K=5, n=15/variant):**

| variant | veeredUncaught | neverVeered | purity | avg output tokens |
| **A-control** | **2/15 (13%)** | 87% | 92 | 28,063 |
| **B-selfcheck** | **0/15** | 100% | 97 | 27,595 |
| C-strong | 0/15 | 100% | 98 | 26,627 |

**Larger sample (A vs B, ~K12, n≈36/variant):**

| variant | veeredUncaught | neverVeered | purity | avg output tokens |
| **A-control** | **4/37 (11%)** | 84% | 93 | 23,866 |
| **B-selfcheck** | **1/36 (3%)** | 97% | 96 | 23,399 |

Control's failures cluster on `wd-1` (the checkout flow). Across both runs the self-check **reduces veered-and-uncaught by roughly 4×** and raises purity, at ~equal tokens. Honest note: at larger n the effect is a strong reduction, **not a perfect zero** (1/36) — but consistently and substantially better than control.

### Cost (tokens, not dollars)

Token counts are **exact** (parsed from the CLI's `stream-json` result events). The self-check adds **essentially no cost** — B and control use the same order of tokens (avg output within a few %) because cache-read (the docs the agent reads) dominates the total, so a few extra prompt lines are rounding error. We track raw token counts, not dollar estimates.

## Why ship B (not C)

On the focused run both hit 0; B is terser, so it's the cleaner thing to bake into the shipped prompt. C-strong is kept in the harness as an alternative if we want to push purity further.

## Risk / rollback

Text-only change to the injected prompt + 2 unit tests (`agent-docs.test.mjs`, **50/50 pass**; full CI green). No API/behavior change. Trivial to revert.

## Reproduce

```bash
cursor-agent status   # must be logged in
node internal/vibe-tests/prompt-purity-test/run-purity.mjs \
  --conditions A-control,B-selfcheck --prompts wd-1,wd-4,dd-5 --reps 5
```

Full methodology, the veer/catch classifier, and the autonomous `/loop` search harness are documented in `internal/vibe-tests/prompt-purity-test/PLAN.md`.


cixzhang:
Nice! For the vibe test is this something you want to run regularly in the daily cron and self healing loops?

cixzhang (approved):

PR #4048, opened 2026-07-18
feat(docsite): add VisuallyHidden showcase and example blocks

ernestt:
## Summary

`VisuallyHidden` shipped without a showcase or example blocks, so its component page rendered an empty hero and had no **Examples** section. Because the component is invisible by design, a bare live preview also looks broken.

This adds a hero showcase plus three examples that teach the concept and genuinely exercise the component:

- **Showcase (hero):** contrasts *what a sighted user sees* (icon-only buttons) with *what a screen reader hears* (the accessible name each control exposes), and includes a live `aria-live` region.
- **Live Region:** a polite `aria-live` region announcing a visual-only state change (moving a task between columns) to assistive technology.
- **Supplementary Context:** screen-reader-only text that spells out what a terse trend arrow means next to a KPI.
- **Structural Heading:** a hidden `<h2>` that gives a visually-implicit group a landmark screen-reader users can navigate to.

## Notes

- Blocks compose only from existing primitives (`IconButton`, `Card`, `Badge`, `Text`, `Icon`, layout), following the block-authoring conventions.
- Added the showcase-dir reference to the `SYNC:` block in `VisuallyHidden.tsx` so `check:sync` stays green.

## Test plan

- `pnpm -F @astryxdesign/docsite generate` wires the showcase into `showcaseRegistry` and the three examples into `exampleRegistry` for `VisuallyHidden`.
- `pnpm -F @astryxdesign/docsite typecheck` passes clean.
- `/components/VisuallyHidden` renders the hero and Examples section (verified locally, HTTP 200).

### DIFF
--- a/packages/cli/templates/blocks/components/VisuallyHidden/VisuallyHiddenLiveRegion.doc.mjs
+++ b/packages/cli/templates/blocks/components/VisuallyHidden/VisuallyHiddenLiveRegion.doc.mjs
@@ -1,0 +1,14 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+/** @type {import('../../../../../core/src/docs-types').TemplateDoc} */
+export const doc = {
+  type: 'block',
+  exampleFor: 'VisuallyHidden',
+  name: 'VisuallyHidden — Live Region',
+  displayName: 'VisuallyHidden — Live Region',
+  description:
+    'A polite aria-live region announces visual-only state changes to assistive technology.',
+  isReady: true,
+  aspectRatio: 4 / 3,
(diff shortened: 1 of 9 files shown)


josephfarina:
I don't seem to have an accept button? But lg2m accepted

cixzhang (approved):
Thank you for the VisuallyHidden showcase and example blocks — these are a great addition for showing off the accessibility patterns! The blocks compose cleanly from existing components and stick to semantic tokens, and the core change is just a `SYNC:` doc reference so nothing ships a behavior change. CI is green across the board. Approving. 🙌

PR #4050, opened 2026-07-18
Slider: stop the track from resizing while you drag

### LINKED ISSUE #4033: [Bug] Slider component - track size changes often
### Description

When I drag the slider knob, the track size changes due to text label changing its width (on value update), triggering unnecessary recalculations of the track width. Also there probably should be formatting applied to the numbers in the labels, to avoid things like in the screenshot:

Image

Recording of the track size changes in the docs site Slider component:

https://github.com/user-attachments/assets/44a75aaa-2b1b-4e2e-8858-58e9df2a2d11

 

### Reproduction

1. Render Slider component with label representing value
2. Drag the knob without releasing it
3. notice how the track width is changing, triggering recalculating of the value due to knob constantly shifting under the mouse

### Astryx Version

    "@astryxdesign/core": "^0.1.6",

### Environment

Chrome 150 Mac OS

AKnassa:
## What this does

Fixes the Slider so the bar stays the same size while you drag the knob, and makes the numbers it shows clean (no more `0.30000000000000004`).

https://github.com/user-attachments/assets/3933ce97-0395-4588-b9d9-61de458ed000

## Why

The value label next to the slider changed width as the number changed (like "9" becoming "10"), which squeezed the bar, which moved the knob under your mouse, which changed the number again — so the whole thing jittered while dragging. And with decimal steps, the label could show long floating-point garbage.

## What changed

- The value label now reserves enough space for the widest number it could ever show, so the bar never grows or shrinks mid-drag.
- Digits now use equal-width numerals, so the label doesn't wiggle as values tick by.
- Decimal steps now produce clean values (0.3, not 0.30000000000000004) — in the label, for screen readers (`aria-valuenow`), and in the value given to your code (`onChange`).

## How to see it

Open the Slider story in Storybook (`pnpm -F storybook dev`, then Core → Slider) and drag the knob back and forth without releasing — the track stays perfectly still and the numbers stay tidy. Six new tests cover the width reservation and the decimal-step precision.

Fixes #4033

### DIFF
--- a/packages/core/src/Slider/Slider.tsx
+++ b/packages/core/src/Slider/Slider.tsx
@@ -271,7 +271,11 @@ const styles = stylex.create({
     color: colorVars['--color-text-primary'],
     whiteSpace: 'nowrap',
     flexShrink: 0,
+    fontVariantNumeric: 'tabular-nums',
   },
+  textValueReserved: (minWidth: string) => ({
+    minWidth,
+  }),
   marksContainer: {
     position: 'absolute',
   },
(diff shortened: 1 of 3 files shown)


packages/core/src/Slider/Slider.tsx
```diff
@@ -766,15 +792,30 @@ export function Slider({ref, ...props}: SliderProps) {
...
+  // the track and shifts the thumb under the pointer. Reserve the width of the
+  // widest formatted candidate up front (tabular-nums keeps digits equal
+  // width) so the track size stays put.
+  let textDisplay = null;
+  if (valueDisplay === 'text') {
+    const candidates = [min, max, min + step, max - step].map(v =>
+      clamp(snapToStep(v, min, step), min, max),
+    );
+    const widest = Math.max(...candidates.map(v => displayValue(v).length));
+    const reservedCh = isRange ? widest * 2 + RANGE_SEPARATOR.length : widest;
```
cixzhang, line 806: Interesting. I hadn't seen ch before. https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/length#ch

> Represents the width or, more precisely, the advance measure of the glyph 0 (zero, the Unicode character U+0030) in the element's font.

Reasonable to compute it like this when applying tabular numerical but I'm not sure if we want to change the design for numerics of the slider. In other places where we deal with text sizing we also use the strategy of applying hidden, inert text in the same location and leveraging the browser layout (I think we do this in tabs and toggle buttons where weight is modified based on state). Curious what you think about consolidating to this strategy and avoiding the typography update?

cixzhang (changes requested):
Thanks for working on this fix. Sorry review took so long. You'll want to rebase this to resolve merge conflicts and I'm curious about your thoughts on the comment too.

cixzhang:
Checking in on the Slider fix: the current review still needs the branch refreshed and the drag-width comment answered. Would you like to continue, have us carry it, or close this for now?

PR #3232, opened 2026-06-29
feat(layout): standardize sizing props (Tiers 1–2 of #3223)

cixzhang:
## What

First implementation step toward a standard layout prop set (#3223). Tiers 1 and 2: safe, mechanical changes with no runtime behavior change.

**Tier 1 — consistency / cleanup**
- Remove redundant `xstyle` / `className` / `style` re-declarations on `Stack`, `StackItem`, and `Layout`. These already come from the shared base props; re-declaring them is noise (the Component Auditor flags it). The components still destructure and forward them — only the duplicated interface declarations are removed.
- Migrate layout-area sizing props to the shared `SizeValue` type:
  - `LayoutHeader.height`, `LayoutFooter.height`, `LayoutPanel.width`: `number | string` → `SizeValue`
  - `Layout.contentWidth`: `number` → `SizeValue` (now also accepts CSS strings like `'60ch'`)

**Tier 2 — complete the sizing set**
- Add `maxWidth` and `minHeight` to `Stack`, `Grid`, and `Center`, matching the sizing set already on `Section` and `Card`. `HStack`/`VStack` inherit these automatically (they extend `StackProps`).

## Why

The layout components each grew their own vocabulary for the same concerns. This aligns the easy ones first:
- One sizing type (`SizeValue`) across all layout containers instead of ad-hoc `number | string` / px-only `number`.
- The de-facto sizing set (`width`/`height`/`maxWidth`/`minHeight`, as on `Section`/`Card`) now applies to `Stack`/`Grid`/`Center` too — so consumers don't drop to custom CSS for a max-width or min-height.
- No more duplicated escape-hatch declarations to keep in sync.

Tier E (capability props — `overflow`, flex `grow`/`shrink`/`basis`, `isSticky`, page surface, shell `maxWidth`) is intentionally **out of scope** here and will land incrementally, each closing its linked issue.

## How

- `SizeValue` coercion follows each component's existing pattern (inline-style number→px for `Stack`/`Grid`; StyleX dynamic-style function for `Center`; pass-through for `Layout*`, where StyleX already appends `px` to numbers).
- `.doc.mjs` updated for the new props and the `SizeValue` type alignment on `Grid`/`Center`.
- Changeset included.

## Validation

- `pnpm -F @astryxdesign/core build` ✓ (Babel + tsc + CSS + UMD)
- `pnpm -F @astryxdesign/core typecheck` and `typecheck:docs` ✓
[...]

PR #4053, opened 2026-07-18
feat(layout): add shellWidth to cap and center the entire shell

### LINKED ISSUE #2625: XDSLayout: no max-width prop for the whole shell (header + sidebar + content)
## Summary

`XDSLayout` can constrain the **content within each slot** via `contentWidth`, but there's no prop to cap and center the **entire layout shell** (header + start/end panels + content) at a max width. On large monitors a full-bleed settings/app page stretches edge-to-edge unless you wrap the whole thing in custom CSS:

```ts
  constrainedShell: { maxWidth: 1440, marginInline: 'auto' },
});
```
```tsx
<XDSLayout xstyle={styles.constrainedShell} header={…} start={…} content={…} />
```

`contentWidth` doesn't help here because it only narrows the content column — the sidebar and header still span the full viewport, so the layout's overall width is unbounded.

In the `settings` page template this `constrainedShell` rule is the **only** remaining custom CSS — it caps Custom CSS at 12/15 (otherwise the template is a clean A). Under the [Contributing-Templates rubric](https://github.com/facebookexperimental/xds/wiki/Contributing-Templates#template-grading-rubric) it's justified (no XDS alternative), but it's the lone thing keeping the template from a perfect score.

## Proposal

A `maxWidth` (or `shellWidth`) prop on `XDSLayout` that caps the outer shell and centers it — distinct from `contentWidth`, which caps only the content slot.

## Context

Surfaced while grading templates against the Contributing-Templates rubric — companion to #2582 (no XDSImage), #2419 (no responsive XDSGridSpan), #2613 (no sticky prop), #2616 (no XDSTable grouping), #2622 (no edge-docked tabs), #2600 (no full-width XDSButton), #2575 (no Drawer/Sheet), #2594 (no page-surface prop).

jiunshinn:
Closes #2625

Adds a `shellWidth?: SizeValue` prop to `Layout` that caps the ENTIRE shell (header, panels, content, footer, and their dividers) at a max width and centers it with `margin-inline: auto` — the capability #2625 asks for, distinct from `contentWidth` which caps only slot content while headers/footers and their dividers stay full-bleed.

## Why this prop, and why now

This exact prop was pre-reserved by @cixzhang in the layout-prop standardization umbrella (#3223). From her per-component delta for `Layout`:

> ➕ (Tier E) `maxWidth` / `shellWidth: SizeValue` to cap + center the whole shell → #2625.

and from the Tier plan in the same comment:

> **Tier E:** capability props, implemented incrementally — each closes its linked issue.

reaffirmed in her follow-up after the Tier 1–2 PR (#3232) went up:

> Tier E (capability props — overflow, flex `grow`/`shrink`/`basis`, `isSticky`, page surface, shell `maxWidth`) remains out of scope and will land incrementally, each closing its linked issue.

This PR is one of those incremental Tier E landings (same pattern as `isSticky`/`stickyOffset`, which landed via #2613's issue). To be clear: #3223 reserves the capability and the name candidates; it is not sign-off on this implementation — happy to adjust anything.

### Honesty note on the issue's original motivation

#2625 was filed when the `settings` template carried a hand-rolled `constrainedShell` CSS rule (`maxWidth: 1440, marginInline: 'auto'`) — the lone custom CSS keeping it from a perfect rubric score. That specific evidence is stale: merged PR #2998 replaced the custom CSS with `contentWidth={1440}`, so the template no longer has the rubric problem. What #2998 did NOT restore is the original design: with `contentWidth`, the header/footer dividers still run full-bleed across the viewport instead of ending at the 1440px shell edge. The motivation today is the remaining capability gap — a capped shell whose dividers end at the shell edge is still inexpressible without custom CSS — plus the Tier E reservation above.

## Naming: `shellWidth` vs `maxWidth`

#3223 reserved both candidate names. This PR proposes `shellWidth`:

| Consideration | `shellWidth` | `maxWidth` |
|---|---|---|
| Symmetry with `contentWidth` | Both are Layout-specific "cap + center" props — consistent pair | Breaks the pair; `contentWidth`/`maxWidth` read as unrelated |
| Consistency with `maxWidth` elsewhere | No conflict | `maxWidth` on `Stack`/`Section`/`Center` caps WITHOUT centering — reusing the name here would fork its semantics |
| Self-description | Says what it caps (the shell) | Says nothing about the shell vs content distinction |
| Familiarity | New name | Standard CSS name |

**I will rename to `maxWidth` (or anything else) if reviewers prefer it** — it's a one-word change across the PR.

## Implementation notes

- **Attachment point**: the style attaches to Layout's INNER wrapper (the common parent of header + middle row + footer), not the root. The root (`styles.layoutOuter`) uses negative inline/block margins to escape container padding (`calc(-1 * var(--container-padding-*))`); putting `marginInline: 'auto'` there would conflict with those margins. Attaching to the inner div means headers, footers, panels, and their dividers are all capped together — exactly the behavior the issue asks for.
- **No CSS var indirection**: `contentWidth` needs `--layout-content-width` because `LayoutHeader`/`LayoutFooter` consume it in their own inner wrappers; `shellWidth` styles an element `Layout` renders directly, so a plain dynamic style suffices.
- **Not in `themeProps`**: continuous-value props aren't reflected as data attributes (`contentWidth` precedent).
- **Composes with `contentWidth`**: shell capped at `shellWidth`, slot content further constrained inside it (covered by a test + a Storybook story).
- **Documented caveat** (in the prop JSDoc and doc.mjs): on viewports wider than the cap, the page/container background shows on both sides of the shell, and dividers stop at the shell edge. That is the intended look; the docs point to `contentWidth` for full-bleed header/divider treatment.

## What's included

- `packages/core/src/Layout/Layout.tsx` — `shellWidth` prop + dynamic style on the inner wrapper; file header updated per SYNC protocol.
- `packages/core/src/Layout/Layout.doc.mjs` — `shellWidth` props entry. **Bonus**: the doc.mjs props list omitted `contentWidth` entirely, so `astryx component Layout` never documented it — added a `contentWidth` entry too so the contrast between the two props is visible to agent consumers.
- `packages/core/src/Layout/__tests__/shellWidth.test.tsx` — 6 tests mirroring `contentWidth.test.tsx` patterns: applied on the inner shell wrapper, no-op when unset, composition with `contentWidth`, number + string values, smoke with `height="auto"` + `padding={0}`.
- `apps/storybook/stories/Layout.stories.tsx` — 3 stories mirroring the Content Width section: capped shell with dividers ending at the shell edge, `shellWidth` + `contentWidth` combo, narrow-container graceful degrade to 100%.
- `packages/cli/templates/blocks/components/Layout/LayoutShellWidth.{tsx,doc.mjs}` — CLI showcase block mirroring `LayoutContentWidth`; blocks are discovered by directory scan, no registry change needed (verified via `astryx template --list`).
- `.changeset/layout-shellwidth-cap-center-shell.md` — `@astryxdesign/core: patch` (repo is pre-1.0 patch-only, enforced by `check:changesets`; matches the `isSticky` and Stack-`padding` feat-prop precedents).

### Droppable commit

The second commit (`refactor(templates): restore settings shell cap via shellWidth`) switches the `settings` template from `contentWidth={1440}` to `shellWidth={1440}`, restoring the ORIGINAL pre-#2998 design (dividers ending at 1440px) with zero custom CSS. **Drop it if reviewers prefer the current full-bleed-divider look** — the feature commit stands alone. No `@astryxdesign/cli` changeset for it, matching the `product-detail` template adoption commit for `isSticky` (template-only change, no changeset).

## Verification

- `npx vitest run packages/core/src/Layout` — 6 files, 86 tests passed (includes the new `shellWidth.test.tsx`).
- `pnpm -F @astryxdesign/core typecheck` — clean.
- `apps/storybook` typecheck — errors identical byte-for-byte to a clean `upstream/main` tree (pre-existing unresolved workspace-package imports in a fresh worktree); zero new errors from this change.
- `npx eslint <changed files>` — 0 errors.
- `pnpm check:repo` (check:sync, package-boundaries, changesets, demo-media, executable-bits) — all green.
- `astryx component Layout --dense` — now renders both `contentWidth` and `shellWidth` rows; `astryx template --list` discovers the new `Layout — Shell Width` block.


cixzhang:
Thank you for the detailed implementation and validation. This PR is now too old to merge as-is, and the underlying gap came from template grading rather than a current owner-approved Layout requirement. We have consolidated that decision into #5618 and are closing this PR rather than asking you to rebase a direction that may not be retained.

Issue #2594, opened 2026-06-09
XDS: no prop-driven page surface (background color/image + padding) for standalone centered pages


rubyycheung:
## Summary

The centering primitive used for standalone pages (`XDSCenter`) — and layout containers generally — can't paint their **page surface** via props. Any standalone page (login/auth, onboarding, marketing splash) that wants a page background color, a full-bleed background image, or page padding around centered content must drop to `stylex.create` and hand-write it. Under the template grading rubric every such declaration costs Custom CSS points and blocks a perfect score.

Three related, currently-prop-less needs on the same container:

1. **Background color** — a solid page surface (`backgroundColor: --color-background-body`).
2. **Background image** — full-bleed cover photo (`backgroundImage` + `backgroundSize: cover` + `backgroundPosition: center`).
3. **Page padding** — inset so centered content never touches the viewport edge (`padding`).

This is **distinct from #2582** (`XDSImage`/`XDSAspectRatio` for an inline `<img>` element with cover-fit) and from the shipped **#1886 `XDSOverlay`** (a scrim/tint *on top of* existing media). Those don't help a *container surface*.

## Evidence (all four login templates hit this)

**Background color** — `login`, `login-card`, `login-split` each need:
```ts
const styles = stylex.create({
  page: { backgroundColor: colorVars['--color-background-body'] },
});
```
```tsx
<XDSCenter axis="both" height="100dvh" xstyle={styles.page}>
```
A standalone auth page should paint its own surface (it has no host shell), but `XDSCenter` has no background prop — so this single declaration is unavoidable and it drops `login` and `login-card` from 100 → 97.

**Background image + padding** — `login-sso` / `login-split`:
```ts
page: {
  backgroundColor: colorVars['--color-background-body'],
  backgroundImage: `url(${BG_URL})`,   // sso: full-bleed photo
  backgroundSize: 'cover',
  backgroundPosition: 'center',
  padding: spacingVars['--spacing-6'], // split: page edge padding
},
```

`XDSCenter` props today: `axis`, `width`, `height`, `isInline`, `xstyle` — no `background*` / `padding`. So a "centered content on a page surface" layout needs 1–4 irreducible custom-CSS declarations depending on whether it uses a color, an image, and/or padding.

[...]

PR #4054, opened 2026-07-18
feat(core): add padding props to Center

jiunshinn:
Implements the maintainer-approved padding scope of #2594 (background scope was declined); leaving close decision to maintainers.

## What

Adds `padding`, `paddingInline`, and `paddingBlock` props (all `SpacingStep`) to `Center`, mirroring the API that `Stack` gained in #3342 (73940a67c) verbatim: same prop names, same JSDoc wording, same `paddingInline ?? padding` / `paddingBlock ?? padding` resolution, same `!= null` guard (so `padding={0}` applies), same style source (`paddingInlineStyles` / `paddingBlockStyles` from `Layout/padding.stylex`), and applied before `xstyle` so consumer styles can still override.

## Why this scope (maintainer-approved)

In #2594, @cixzhang approved exactly this part:

> The padding prop for XDSCenter seems okay since many of our layout components have it so that's consistent at least

**Out of scope: background color/image props.** The maintainer explicitly declined that part of #2594 and deferred it to the template renderer:

> Hmm I don't think it makes sense here. The page background should be applied on the body so things like over scroll will show the right color and the center component should remain transparent. Maybe the template renderer needs to be page colored more generally to accurately represent how the templates look.

This PR therefore adds no background affordances and leaves `Center` transparent.

## Canonical shape being mirrored

- #3342 (`73940a67c`) is the canonical padding API on a layout container: `padding` + `paddingInline` + `paddingBlock` on the spacing scale, patch-bump changeset with `[feat]` tag, argTypes + a padding story in Storybook, and core-only scope (template adoption as a follow-up).
- #3223 (layout prop standardization) codifies the same as the Tier C spacing standard: "`padding` (+ `paddingBlock` / `paddingInline`) - `SpacingStep`, on surface containers".
- `themeProps('center', ...)` is untouched: `Stack` does not expose padding via themeProps either.
- `Center/index.ts` needs no change (no new exported types; `SpacingStep` already comes from `utils/types`).

## Follow-up this unlocks (quantified)

All 4 login templates (`login`, `login-card`, `login-split`, `login-sso`) currently carry a `pageStyle` with `padding: 'var(--spacing-6)'` on their `<Center>` page root, e.g. `packages/cli/templates/pages/login/page.tsx`:

```ts
const pageStyle: CSSProperties = {
  minHeight: '100%',
  backgroundColor: 'var(--color-background-body)',
  padding: 'var(--spacing-6)',
};
```

With this prop, the padding line in each of those 4 templates becomes `<Center padding={6} ...>`. Per the core-only precedent of #3342, template adoption is deliberately not in this PR; I will send a follow-up template PR after this merges.

## Behavior delta

None by default. When no padding prop is passed, `Center` renders byte-identical output (covered by a test that asserts the default className is unchanged). Padding is opt-in only.

## Verification

- `vitest run packages/core/src/Center/Center.test.tsx`: 16/16 passed (11 existing + 5 new: class applied for `padding`, `paddingInline`/`paddingBlock` accepted, per-axis override, explicit `padding={0}` still applies via the `!= null` guard, default className unchanged without padding props)
- `pnpm -F @astryxdesign/core typecheck`: clean
- `eslint` on changed files: clean
- `prettier --check` on changed ts/tsx/md: clean
- `pnpm check:changesets`: 20 changeset(s) valid (patch bump, `[feat]` tag, contributor credit, matching the #3342 changeset format)
- `node packages/cli/bin/astryx.mjs component Center --dense`: renders the three new props
- Full-repo vitest run: the only failures are pre-existing/environmental in `packages/cli` integration tests and `scripts/build-css.test.mjs`; verified they fail identically on a clean checkout of `upstream/main` in the same environment

### DIFF
--- a/.changeset/center-add-padding-paddinginline-paddingblock-sp-pz1v.md
+++ b/.changeset/center-add-padding-paddinginline-paddingblock-sp-pz1v.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] Center: add `padding`, `paddingInline`, `paddingBlock` (spacing-scale inner padding) props. These match the existing `padding` props on `Stack`, `Card`, `LayoutContent`, and `LayoutPanel`, so centered page content no longer needs inline `style={{}}` or `xstyle` wrappers for basic padding.
+@jiunshinn
--- a/apps/storybook/stories/Center.stories.tsx
+++ b/apps/storybook/stories/Center.stories.tsx
@@ -37,6 +37,16 @@ const styles = stylex.create({
     padding: spacingVars['--spacing-2'],
     borderRadius: radiusVars['--radius-element'],
   },
+  paddingOutline: {
+    borderWidth: 1,
+    borderStyle: 'dashed',
+    borderColor: colorVars['--color-border-gray'],
+    borderRadius: radiusVars['--radius-element'],
+  },
+  fillArea: {
+    width: '100%',
+    height: '100%',
+  },
 });
 
 // Demo box component for visibility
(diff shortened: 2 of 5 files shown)


cixzhang (approved):
Really clean addition — thank you! This brings Center's padding right in line with the shared layout convention: same `padding` / `paddingInline` / `paddingBlock` props typed as `SpacingStep`, reusing `paddingInlineStyles` / `paddingBlockStyles` from `Layout/padding.stylex`, with the same `paddingInline ?? padding` per-axis resolution that Stack and Section use. All optional and additive, so existing Center usage is unchanged, and the surface is complete: tests (including the zero-step and override cases), doc.mjs across all variants, a Storybook story, and a changeset. Nicely done.

PR #4060, opened 2026-07-18
fix(core): guard empty Typeahead navigation

### LINKED ISSUE #4059: [Bug] Typeahead exposes a nonexistent active descendant for empty results
### Description

Typeahead points `aria-activedescendant` at an option that does not exist when keyboard navigation is used in an empty-results popover.

- Sequence: enter a query with no matches, then press ArrowDown or Home.
- Expected: 0 results and no active descendant.
- Actual: 0 results, but the input exposes `<listbox-id>-option-0`.

This leaves assistive technology with a dangling active-descendant reference.

## Actual screenshots

**Before — Storybook:**

Typeahead empty-results reproduction before the fix

**Before — focused regression:**

Focused Typeahead regression failures before the fix

**After — Storybook:**

Typeahead empty-results reproduction after the fix

**After — focused regression:**

Focused Typeahead regression tests passing after the fix

### Reproduction

1. Open **Core / Typeahead / Default** in Storybook.
2. Enter `zzzzz` so the open popover reports “No results found”.
3. Press ArrowDown.
4. Inspect the combobox input and the rendered options.
5. Repeat with Home.

Expected: `aria-activedescendant` is absent while 0 options exist.

Actual: the input references `<listbox-id>-option-0` while the listbox contains 0 option nodes (`targetExists: false`).

### Astryx Version

@astryxdesign/core@0.1.6 (main at 2e2e96046)

### Environment

Storybook 10.4, Chromium on macOS 26.5.2

Kevinjohn:
Typeahead no longer exposes a dangling `aria-activedescendant` when its open results list is empty.

Fixes #4059

- Sequence: enter a query with no matches, then press ArrowDown or Home.
- Expected: 0 results and no active descendant.
- Before: the input referenced `<listbox-id>-option-0` while 0 option nodes existed.
- After: the highlighted index remains `-1` and `aria-activedescendant` stays absent.

## Root cause

ArrowDown wrapped an empty result set to index `0`, and Home assigned index `0` unconditionally while the popover was open.

## Changes

- Keep ArrowDown at `-1` when there are no results.
- Keep Home at `-1` when there are no results.
- Cover both keys with a parameterized regression test.
- Add a patch changeset for `@astryxdesign/core`.

## Actual screenshots

**Before — Storybook (`facebook/astryx@2e2e96046`):**

Typeahead empty-results reproduction before the fix

**Before — focused regression:**

Focused Typeahead regression failures before the fix

**After — Storybook (`Kevinjohn/astryx@dcc4491fc`):**

Typeahead empty-results reproduction after the fix

**After — focused regression:**

Focused Typeahead regression tests passing after the fix

## Validation

- `vitest run packages/core/src/Typeahead/Typeahead.test.tsx` — 44/44 passed
- `vitest run --project ui` — 225 files, 4,853 tests passed
- `pnpm lint` — 0 errors
- `git diff --check`

### DIFF
--- a/.changeset/typeahead-empty-active-descendant.md
+++ b/.changeset/typeahead-empty-active-descendant.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Typeahead no longer exposes a nonexistent active descendant when ArrowDown or Home is pressed with no search results.
+
+@Kevinjohn
--- a/packages/core/src/Typeahead/BaseTypeahead.tsx
+++ b/packages/core/src/Typeahead/BaseTypeahead.tsx
@@ -637,9 +637,12 @@ export const BaseTypeahead = function BaseTypeahead<T extends SearchableItem>({
       switch (e.key) {
         case 'ArrowDown':
           e.preventDefault();
-          setHighlightedIndex(prev =>
-            prev < results.length - 1 ? prev + 1 : 0,
-          );
+          setHighlightedIndex(prev => {
+            if (results.length === 0) {
+              return -1;
+            }
+            return prev < results.length - 1 ? prev + 1 : 0;
+          });
           break;
         case 'ArrowUp':
           e.preventDefault();
@@ -660,7 +663,7 @@ export const BaseTypeahead = function BaseTypeahead<T extends SearchableItem>({
         case 'Home':
           if (popover.isOpen) {
             e.preventDefault();
-            setHighlightedIndex(0);
+            setHighlightedIndex(results.length > 0 ? 0 : -1);
           }
           break;
         case 'End':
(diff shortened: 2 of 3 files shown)


cixzhang:
Thank you so much for this, @Kevinjohn — you spotted the empty-results `aria-activedescendant` bug and your ArrowDown/Home guards were spot on, with a nice parametrized test to lock it in.

We're going to resolve #4059 via #4312, which is landing the same fix at a slightly wider scope: it also guards the `aria-activedescendant` render itself (so the input can never reference a nonexistent option regardless of the code path), covers ArrowUp/End alongside ArrowDown/Home, and handles the `maxMenuItems` truncation edge. Since the two PRs overlap heavily on the same lines, I'm closing this one as a duplicate to avoid a conflicting merge — but your diagnosis and fix were correct, and it genuinely helped confirm the right approach.

Really appreciate you taking the time on this. Hope to see more contributions from you. 🙏

PR #3789, opened 2026-07-11
feat(table): tree data plugin — useTableTreeState + useTableTreeData

AKnassa:
Closes #3346.

## What this does

Adds tree support to Table: rows can nest under a `children` key, and the table shows them as an expandable, indented hierarchy — like a file browser. Clicking a chevron opens a row's children; clicking again closes them.

## Why

Internal-tool builders keep hand-rolling row indentation and expand/collapse on top of Table (the Meridian gap report rebuilt all of it from scratch). #3346 specs this as a first-class plugin pair, and the Q3 Table tracker (#3639) lists tree rows as an open item.

## What changed

default-dark
default-light
expand-and-collapse-all-dark
expand-and-collapse-all-light
flat-data-is-a-no-op-dark
flat-data-is-a-no-op-light
indent-sizes-dark
indent-sizes-light
interaction-1-collapsed
interaction-2-expanded
with-selection-dark
with-selection-light
with-sibling-sorting-dark
with-sibling-sorting-light

Two hooks, following the same state/plugin split as sorting and selection:

- **`useTableTreeState`** owns which rows are open (controlled or uncontrolled) and flattens the nested data into exactly the rows the table shows. Closed branches are unmounted, not hidden — a 10,000-node tree with everything closed renders only its roots. Also returns `expandAll`/`collapseAll`, supports lazy-loaded children (`isItemExpandable`), and per-sibling-group sorting (`sortSiblings` — pass `applySort` from `useTableSortableState`; levels never interleave).
- **`useTableTreeData`** draws the tree: per-level indent (spacing tokens `sm`/`md`/`lg`) and an expander chevron in the tree column (first column by default, configurable via `treeColumnKey`), plus `aria-level`/`aria-expanded` on rows and the expander button per the issue's ARIA scope. No synthetic column — other columns get zero extra DOM.
- `'tree'` joins the canonical plugin order between `'sort'` and `'selection'`, so the selection checkbox column lands left of the indented tree column.

**Migration guarantee:** with flat data the plugin is a true no-op — cells render byte-identical to a plugin-free Table (tested), so teams can adopt it before their data becomes hierarchical.

**Performance:** mirrors the selection plugin's external-store design — toggling a row re-renders only the affected cells, not the body (perf test included). Wrapped column identity is cached (shallow-equal keyed) so row memoization survives; row refs return cleanups so subscriptions can't leak across re-renders; the plugin keeps one stable shape so flat↔nested data flips never remount the table.

[...]

PR #4070, opened 2026-07-19
fix(Table): guard rowExpansion tree walks against cyclic data

### LINKED ISSUE #3971: useTableRowExpansion: add cycle guard to tree walks (stack overflow on cyclic data)

**Feature area:** Row grouping & hierarchy

### Problem

`useTableRowExpansionState` walks the tree in three places — `depthMap`, the flattened `data`, and `allExpandableKeys` — and **none of the walks guard against cycles**. If `getChildren` ever returns an ancestor (cyclic data, or a self-referential row), the recursion never terminates → stack overflow / hung render.

```ts
// current: no ancestor tracking
function walk(items: T[], depth: number) {
  for (const item of items) {
    map.set(getRowKey(item), depth);
    if (expandedKeys.has(getRowKey(item))) {
      walk(getChildren(item), depth + 1); // recurses forever on a cycle
    }
  }
}
```

### Fix

Adopt the ancestor-chain `Set` guard used in #3789's tree walks: track the ids on the current path and `continue` when an edge points back at an ancestor.

```ts
const path = new Set<string>();
function walk(items: T[], depth: number) {
  for (const item of items) {
    const key = getRowKey(item);
    if (path.has(key)) continue; // cyclic edge — skip
    map.set(key, depth);
    if (expandedKeys.has(key)) {
      path.add(key);
      walk(getChildren(item), depth + 1);
      path.delete(key);
    }
  }
}
```

Apply to all three walks. Add a cyclic-data unit test (a row whose children include itself / an ancestor) that asserts the walk terminates.

_Surfaced during review of #3789._

AKnassa:
## What this does
Stops the Table's row-expansion feature from crashing when the row data accidentally contains a loop (a row that ends up being its own ancestor).

## Why
With cyclic data, the three internal tree walks recursed forever and crashed with a stack overflow — one of them even with nothing expanded. This is the cycle-guard follow-up humbertovirtudes prescribed in #3971, and the fix pattern mirrors the guarded tree walks already reviewed in #3789.

## What changed
- Each of the three walks (depth map, flattened rows, expand-all keys) now tracks the keys on its current path and simply skips an edge that loops back.
- 3 new tests reproduce the crash first (self-referencing row, deeper cycle, expand-all with nothing expanded); a 4th pins the guard's ancestor-path semantics on shared-subtree (DAG) data — a shared child must flatten under each expanded parent, which a global visited-set guard would get wrong.
- Behavior for normal, non-cyclic data is unchanged.
- The four formatting-only hunks in the plugin section are the repo's own prettier normalizing pre-existing formatting on `main` (lint-staged applies it on commit) — safe to skip in review.

## How to see it
`pnpm exec vitest run packages/core/src/Table/plugins/rowExpansion/useTableRowExpansion.test.tsx` — 13/13 pass (the 3 cyclic ones crash with RangeError before the fix; the DAG test fails if the guard is degraded to a visited-set). Full core suite green.

Fixes #3971

### DIFF
--- a/.changeset/row-expansion-cycle-guard.md
+++ b/.changeset/row-expansion-cycle-guard.md
@@ -1,0 +1,9 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Guard useTableRowExpansionState tree walks against cyclic data (#3971)
+
+The `depthMap`, flattened `data`, and `allExpandableKeys` walks now track the ancestor keys on the current path and skip edges that point back at an ancestor, so self-referential or cyclic row data terminates instead of overflowing the stack.
+
+@AKnassa
--- a/packages/core/src/Table/plugins/rowExpansion/useTableRowExpansion.test.tsx
+++ b/packages/core/src/Table/plugins/rowExpansion/useTableRowExpansion.test.tsx
@@ -154,7 +154,7 @@ describe('useTableRowExpansion', () => {
   });
 });
 
-describe('useTableRowExpansionState with cyclic data', () => {
+describe('useTableRowExpansionState cycle guard', () => {
   /** A row whose children array contains the row itself (plus a real child). */
   function makeSelfReferential(): TreeItem[] {
     const x: TreeItem = {id: 'x', name: 'Self', children: []};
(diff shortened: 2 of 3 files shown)


humbertovirtudes (approved):
Approving. This is exactly the ancestor-path guard prescribed in #3971, and it's done right.

All three walks (`depthMap`, flattened `data`, `allExpandableKeys`) now track the keys on the current path and skip an edge that points back at an ancestor, adding before recursing and deleting after. That terminates on cycles without corrupting acyclic output.

The highlight for me was DAG test: a shared child under two expanded parents must flatten under both, and a naive global visited-set would emit it only once. Pinning `['p1','s','leaf','p2','s','leaf']` proves you chose ancestor-path semantics deliberately over the easy-but-wrong visited-set. 

Thanks for picking this up and for the rigor on the test design, awesome work!

Issue #3829, opened 2026-07-11
Spec: Radio & Checkbox menu items + Submenus for DropdownMenu (refactor from isSelected, #3821)


cixzhang:
## Summary

Refactor DropdownMenu selectable items into an explicit set of lab components — `DropdownMenuRadioList` / `DropdownMenuRadioItem` (single-select, `role="menuitemradio"`) and `DropdownMenuCheckboxItem` (independent toggle, `role="menuitemcheckbox"`) — instead of the implicit `isSelected` boolean prototyped in #3821.

Radio (one-of-N) and checkbox (independent) are distinct ARIA + interaction contracts; a single boolean overloads one item type with two behaviors. This issue runs the full **Component Specification Protocol** for both, so the API is decided by evidence (API Arbitration) rather than opinion.

#3821 has been moved back to **draft** and is parked pending this spec. Its real, latent keyboard fix (list-focus nav + Enter/Space were hard-matched to `role="menuitem"` only, making radio/checkbox items keyboard-unreachable) should carry over — and may land independently since it's a genuine bug.

## Specification (draft)

# Component Specification: Radio & Checkbox Menu Items (DropdownMenu lab)

**Status:** Draft spec (specification phase) · **Tracks:** #3821 (parked to draft) · Split origin: #3326
**Family:** Action / Menu (DropdownMenu, ContextMenu, MoreMenu)

## Phase 1 — Triage

#3821 added `isSelected?: boolean` to the base `DropdownMenuItem`, rendering `role="menuitemradio"`. Review feedback: a single boolean overloads one item type with two distinct ARIA/interaction patterns. Radio (single-select, one-of-N, `menuitemradio`) and checkbox (independent toggles, `menuitemcheckbox`) are different contracts and should be **explicit, composable lab components** rather than an implicit flag. This is behavior/new surface on an existing family → full spec protocol applies (per the reviewer's instruction).

Not solvable by pure composition: the parent must own radio group state (single selection) and the keyboard/role wiring (the latent `menuitem`-only focus bug #3821 fixed must carry over).

## Phase 2 — Research (internal)

Port the *value*, not the structure. Internal picker patterns reimplemented single-select 3× because menu items couldn't express checked state (per #3821 "Why"). Value to port: selectable menu items with correct roles + reserved check-icon slot (Selector idiom) so unselected siblings don't shift.

## Phase 3 — Research (external)

[...]

PR #4071, opened 2026-07-19
test(DropdownMenu, ContextMenu): pin keyboard access for radio and checkbox menu items

AKnassa:
## What this does

Adds the keyboard tests for `menuitemradio` / `menuitemcheckbox` menu items in DropdownMenu and ContextMenu. **Tests only — no source change.**

## Why this PR shrank

It originally carried the keyboard fix itself. #4122 landed that same fix on `main` on Jul 21 as a shared `menuItemRoles.ts` module, so the source change here is now redundant — and main's version is the better one, since both menus import a single module instead of duplicating a const per file.

Rebased on latest `main`, with both source conflicts resolved by taking main wholesale. The changeset is gone too: nothing ships, so no version bump is warranted.

What's left is the part `main` doesn't have.

## Why the tests are still worth landing

`main` has no arrow-key, Enter/Space, or selectable-role typeahead coverage in either menu. Its only two menu keyboard tests use plain `role="menuitem"` rows with nothing disabled, and `DropdownMenuSelectable.test.tsx` is entirely click-driven. The behaviour #4122 shipped is currently unpinned.

Verified by mutation against main's `menuItemRoles.ts`:

| Mutation | Result |
|---|---|
| `MENU_ITEM_ROLES` narrowed back to `menuitem` | **12 fail**, 75 pre-existing pass |
| `:not([aria-disabled="true"])` dropped from the selector | **exactly 2 fail**, 85 pass |

The second one is worth calling out. Neither menu passes `useTypeahead`'s `isDisabled` option, so that selector clause is the *only* thing keeping disabled rows out of the typeahead list. An arrow-key test cannot cover it — `useListFocus` re-filters disabled items independently, so arrow navigation is guarded twice over. `typeahead skips an aria-disabled item and matches the next enabled label` is the only test in the repo that catches it.

## A note on the fixtures

The tests render raw `<div role="menuitemradio">` rows rather than the new `DropdownMenuRadioItem` / `DropdownMenuCheckboxItem` components. That's deliberate: `MENU_ITEM_SELECTOR` is a role-based contract, and going through the components would couple these tests to component internals.

It matters most in the disabled cases. The real components drop `tabIndex` when disabled, so a broken selector would fail the test for the wrong reason — an unfocusable element rather than the selector clause. The raw rows keep `tabIndex={-1}` while disabled, so they stay focusable and the selector clause is the sole reason focus skips them. That makes them a sharper detector, and it also covers the consumer-rendered-role case directly.

## MoreMenu

Confirmed, and the answer is that there is nothing to test. MoreMenu is pure delegation — one `<DropdownMenu items={...}>` render with no keyboard code of its own — so it inherits the widened selector for free. But its API is closed (`items: DropdownMenuOption[]`, no `children`, no rest-spread), and that shape only yields dividers, `role="group"` sections, and `DropdownMenuItem`, which hard-codes `role="menuitem"`. A consumer cannot construct a radio or checkbox row through MoreMenu at all — the roles are unreachable by API, not merely untested.

Separately worth noting: MoreMenu has no keyboard tests of any kind on `main`, not even over plain `menuitem`. That is a real gap, but pre-existing and orthogonal to this change.

## How to check

```
pnpm exec vitest run packages/core/src/DropdownMenu packages/core/src/ContextMenu
```

87/87 pass. Full `@astryxdesign/core` package: 5124/5124 across 222 files. Typecheck, lint, prettier and all five repo checks green.

Refs #3829. The keyboard fix itself landed in #4122.


humbertovirtudes (dismissed):
Great, this is the keyboard gap #3829 flagged and the fix is done the right way. Extending the item selector so `menuitemradio` and `menuitemcheckbox` participate in arrow navigation, first-character typeahead, and Enter/Space activation (while still skipping `aria-disabled`) is what these compound menus needed, and routing it through the shared `useListFocus` primitive rather than hand-rolling focus movement keeps it consistent with the rest of the system.

Things I liked:
- DRYing the selector into one `MENU_ITEM_SELECTOR` const with an explicit "keep in sync with the Enter/Space role check" note, in both DropdownMenu and ContextMenu. That comment will save the next person a subtle bug.
- The test coverage is thorough: arrow reach, Enter on radio + Space on checkbox, typeahead onto each role, and crucially the "typeahead skips a disabled 'n' match and keeps its index aligned with the focus list" case, which is exactly where this kind of change tends to break.

One non-blocking check: the changeset mentions MoreMenu sharing the path via delegation. Worth a quick confirmation there is coverage (or a follow-up note) for the MoreMenu surface specifically. Not a blocker. Approving.

humbertovirtudes:
Correcting my earlier review here: I dug into the conflict and this looks already resolved on `main`, so my approval was premature.

#4122 (merged Jul 21, "DropdownMenu & ContextMenu radio & checkbox menu items", refs #3829) already shipped this keyboard-reachability fix. It added a shared `packages/core/src/DropdownMenu/menuItemRoles.ts` (`MENU_ITEM_ROLES` + `MENU_ITEM_SELECTOR`) that both `DropdownMenu` and `ContextMenu` now use for roving focus, typeahead, and Enter/Space, so `menuitemradio` / `menuitemcheckbox` rows are already reachable, and focus-does-not-select is preserved. That is the same behavior this PR adds, and the shared-module approach is a cleaner version of the per-file selector this PR keeps in sync by hand.

So this appears to be redundant with what already landed. Unless you have spotted a case #4122 missed, I think the cleanest path is to close this as landed-elsewhere. If there is a genuine gap, could you point at the specific case and rebase down to just that delta? Either way, thank you for pushing on #3829: the fix that landed is exactly the shape this was after, and the duplication here is only because two efforts converged on the same problem in parallel. Apologies for the approve-then-correct.

AKnassa:
**You are right about the source, and this is already rebased down to just that delta.** As of `605ed7d` the PR is two test files, +362, zero source lines. Both conflicts resolved by taking main wholesale, changeset dropped. The shared `menuItemRoles.ts` is the better version — nothing of mine survives in the implementation.

But there is a genuine gap, and in small words it is this: **main has the fix. Main has no test for the fix.** Delete one line and everything stays green while the menu quietly breaks for keyboard users.

Proof, by mutating main's `menuItemRoles.ts`:

- narrow `MENU_ITEM_ROLES` back to `menuitem` → **12 fail, 75 pre-existing pass**. Revert the whole fix and main's own suite never notices.
- drop `:not([aria-disabled="true"])` → **exactly 2 fail, 85 pass**.

The second is the specific case you asked me to point at. Neither menu passes `useTypeahead`'s `isDisabled` option, so that clause is the only thing keeping disabled rows out of the typeahead list — and no arrow-key test can catch it, because `useListFocus` filters disabled items independently.

On your MoreMenu question: nothing to test there. It is pure delegation, so it inherits the widened selector for free, but its API is closed (`items: DropdownMenuOption[]`, no `children`) and that shape only yields `DropdownMenuItem`, which hard-codes `role="menuitem"`. The selectable roles are unreachable through MoreMenu, not merely untested.

So closing this deletes the only net under a fix that currently has none. Happy to land the tests as a fresh PR against main instead if you would rather have a clean branch — just say which and I will.

<sub>Unrelated, spotted while reading #4122: `DropdownMenuRadioItem.tsx:131` documents disabled items as staying keyboard-focusable, but `:192` sets `tabIndex={isDisabled ? undefined : -1}` and the shared selector excludes `aria-disabled` rows. The code matches the older `DropdownMenuItem.tsx:135`, so it looks like the doc is the part that is out of step. Flagging only.</sub>

cixzhang (approved):
Test-only PR — adds coverage pinning keyboard access (arrow nav, typeahead, Enter/Space activation, aria-disabled skipping) for consumer-rendered `menuitemradio`/`menuitemcheckbox` items in DropdownMenu and ContextMenu. No source changes. Ran both suites locally on this head: 87 tests pass. Good regression guard for #3829. LGTM.

Issue #2008, opened 2026-05-03
Add playground defaults and slotElements to all component docs


cixzhang:
## Context

PR #2005 added two new fields to the component doc system:

1. **`playground.defaults`** on `BaseDoc` — initial prop values for the interactive playground preview
2. **`slotElements`** on `PropDoc` — declares what XDS components a ReactNode prop typically accepts

Currently only 4 components have playground defaults (Button, Badge, Card, EmptyState) and 2 props have slotElements (Button.icon, Button.endContent). Every other component needs to be updated.

## What needs to happen

Every `.doc.mjs` file in `packages/core/src/` needs:

### 1. `slotElements` on ReactNode props

For each prop with `type: 'ReactNode'` or `type: 'ReactElement<...>'`, decide:

**Add `slotElements`** when the prop expects specific XDS components:
```js
{
  name: 'icon',
  type: 'ReactNode',
  description: 'Leading icon.',
  slotElements: [{__element: 'XDSIcon', props: {icon: 'check', size: 'sm'}}],
}
```

**Skip `slotElements`** (leave as-is) when:
- The prop accepts plain **text** (label, title, description, name) — the playground already renders a text input for these
- The prop is a **render function** (`(item: T) => ReactNode`) — not an element slot
- The prop is a **compound component entry** (names starting with `XDS` or `use` in compound docs) — these are documentation of sub-components, not interactive props
- The prop is **`children`** on a generic container — too many possible values to enumerate
- The prop is internal (`element` for imperative APIs, `layerNode` for portals)

### Common slot patterns

| Prop name | Typical `slotElements` |
|-----------|------------------------|
| `icon`, `startIcon`, `endIcon`, `sendIcon`, `stopIcon`, `pressedIcon`, `selectedIcon` | `[{__element: 'XDSIcon', props: {icon: 'check', size: 'sm'}}]` |
| `endContent` (on buttons/inputs) | `[{__element: 'XDSIcon', props: {icon: 'chevronDown', size: 'sm'}}, {__element: 'XDSBadge', props: {label: '3'}}]` |
| `startContent` | `[{__element: 'XDSIcon', props: {icon: 'check', size: 'sm'}}]` |
| `actions` | `[{__element: 'XDSButton', props: {label: 'Action', variant: 'secondary'}}]` |
| `status` (on Avatar) | `[{__element: 'XDSStatusDot', props: {variant: 'online'}}]` |
[...]

PR #4072, opened 2026-07-19
docs(core): playground examples for 7 component docs

AKnassa:
## What this does
Gives seven more components a working interactive example in the docsite playground: Section, FormLayout, Breadcrumbs, List, RadioList, CheckboxList, and DropdownMenu — and keeps their item sub-pages clean while doing it.

## Why
Six of these are on the high-priority list in #2008; DropdownMenu is included because it renders the same empty playground for the same reason. This is a scoped slice on purpose — the wider sweep is landing piecemeal, so this only covers components nobody else has claimed.

## What changed
- Each of the 7 doc files gets playground defaults (and slot examples where the component accepts arbitrary content), taken from the descriptor tables in #2008 or the component's own canonical example — no invented content. The inline BreadcrumbItem entry's `startIcon` slot descriptor is synced with the standalone BreadcrumbItem doc (which already carried it).
- The BreadcrumbItem/ListItem/CheckboxListItem/DropdownMenuItem sub-components declare their own minimal playground defaults, so the docsite registry's playground inheritance doesn't leak the parents' multi-item state onto the item pages (same convention as RadioListItem/ContextMenuItem).
- A new scanner test keeps every playground honest repo-wide: defaults must be real documented props (including compound docs whose props live on `components[]` entries), and referenced element names must actually resolve the way the docsite resolver does. It caught two real pre-existing issues — Slider's playground used a `width` prop its docs never documented, and Toolbar's used an undocumented `dividers` — so both props are now documented (en/zh/dense).

## How to see it
`pnpm exec vitest run packages/core/src/docPlaygroundDescriptors.test.ts` — 6/6; the slice assertions fail before the doc edits, pass after. Full suite green (6776/6776). On the docsite, the 7 component pages now show a live preview instead of an empty playground, and their item sub-pages keep sensible single-item previews.

Part of #2008


packages/core/src/docPlaygroundDescriptors.test.ts
```diff
@@ -0,0 +1,238 @@
...
+        for (const prop of docs?.props || []) {
+          if (prop.type !== 'ReactNode') {
+            continue;
+          }
+          const hasSlotElements =
+            Array.isArray(prop.slotElements) && prop.slotElements.length > 0;
+          // Generic-container `children` skips slotElements per #2008; the
+          // playground default must supply the content instead.
+          const coveredByDefaults =
+            prop.name === 'children' &&
```
cixzhang, line 213: `DropdownMenu.children` is mutually exclusive with `items`, so this asks for a default that cannot coexist.

cixzhang (changes requested):
Thanks for this — the scanner is the part I like most, and the two pre-existing hits it turned up are what it should be catching.

I merged main in for you and dropped one duplicate the merge created: main documented `Slider.width` separately, so there were two entries in the same props array. Your zh and dense ones are still new.

One thing left. At the merged head the scanner fails on `DropdownMenu`, and the hit is not real — [#4954](https://github.com/facebook/astryx/pull/4954) retyped `children` from `(item) => ReactNode` to plain `ReactNode` and made it mutually exclusive with `items`, so a playground can seed one or the other, never both. Main's seeds `items`.

Could the ReactNode check skip a prop the docs describe as mutually exclusive with one the playground already seeds?

[Reviewed by Robohands]

cixzhang:
Checking in on the playground scanner: the current review needs mutually exclusive ReactNode props exempted when the playground already seeds the other side. Would you like to continue, have us carry it, or close this for now?

Issue #3636, opened 2026-07-06
[Tracker] i18n & RTL support — string injection + RTL correctness


cixzhang:
## Problem

Astryx originally shipped user-facing and assistive text hardcoded in English, with no consumer translation path. Several components also rendered or behaved incorrectly in right-to-left layouts. Together those gaps blocked localized and RTL applications from adopting the library without forks or component replacements.

This issue tracks the cross-library i18n and RTL program. Individual component bugs and API proposals remain in their own issues and PRs.

## Approach

1. **String translation and locale behavior.** Ship a provider, English fallback catalog, translator hook, locale-aware formatting APIs, community translation pipeline, and enforcement against new hardcoded or ambient-locale behavior.
2. **RTL correctness.** Prefer logical CSS, mirror directional decoration exactly once, make keyboard behavior direction-aware, and continuously audit rendered components.
3. **International input correctness.** Treat IME composition and grapheme clusters as first-class correctness constraints rather than English-input edge cases.

Design context: proposal #3618 · RFC #3641.

## Status

_Last refreshed September 1, 2026._

### Shipped — i18n runtime and string migration

- [x] Provider, English fallback catalog, and `useTranslator` runtime — #3618, #3765
- [x] Full core string migration — #3922, #3941, #4011, #4013
- [x] Table and assistive-technology strings — #4073, #4376, #4387
- [x] Chat strings — #4452
- [x] Consumer documentation — #4016, #4243
- [x] Hardcoded-string lint enforcement — #4010, #4014, #4015
- [x] Live-region `announce()` strings included in enforcement — #4920

### Shipped — provider locale and deterministic formatting

- [x] Provider-backed locale hooks — #5194
- [x] Non-date locale-sensitive consumers migrated to the provider locale — #5195
- [x] Calendar and date-input formatting/parsing follow the provider locale — #5120
- [x] Date helpers default deterministically, use Gregorian semantics unless explicitly overridden, and rerender on provider changes — #5303
- [x] Raw/ambient `Intl` and omitted-locale helper calls are linted — #5171
- [x] Compact Calendar weekday names generated from Unicode CLDR data — #5190
- [x] Localized grouped-number parsing — #5450
- [x] Calendar announcements and TransferList strings localized — #5453, #5457

NOTE: umbrella bug #5074 remains open for bookkeeping, but its reported implementation paths have been addressed by the provider-locale stack above. It should be reconciled and closed after a final issue-level audit.

### Shipped — catalogs and Crowdin

[...]

PR #4073, opened 2026-07-19
fix(Table): translate sort header screen-reader labels

### LINKED ISSUE #3618: Proposal: i18n/strings provider for component-internal strings (Table sort aria, Pagination, Calendar months)
## Problem

Several components hardcode user-facing / assistive strings in English with no injection point, which blocks fully-localized apps (ours ships Arabic + Kurdish, RTL):

- **Table sortable plugin** (`useTableSortable`): `Sort by ${label}`, `Sort by ${label}, sorted ascending/descending` aria-labels, and the header context-menu items ("Sort ascending", "Sort descending", "Clear sort").
- **Pagination**: the count/compact variant text (`X–Y of Z`, `Page X of Y`) and the prev/next button aria-labels (`Go to previous page` / `Go to next page`).
- **Calendar** (inside DateInput/DateRangeInput): month names, weekday abbreviations, and month-navigation aria-labels.

Many components already do this right via props (`placeholder`, `emptySearchResultsText`, `label`, `emptyBootstrapText`, `disabledMessage`) — the gap is the strings that have no prop.

## Proposal

A lightweight strings provider (analogous to `LinkProvider`), e.g.:

```tsx
<StringsProvider strings={{
  tableSortBy: (label) => `فرز حسب ${label}`,
  paginationCount: (from, to, total) => `${from}–${to} من ${total}`,
  paginationNext: 'الصفحة التالية',
  // …defaults to the current English strings
}}>
```

or per-component override props if a global provider doesn't fit the design system's philosophy. Either shape would remove the need for consumers to swizzle `Table`/`Pagination`/`Calendar` solely to translate a handful of strings.

## Environment

`@astryxdesign/core` 0.1.3, Vite + React 19, no build plugin (pre-built CSS path).

Happy to contribute the PR if the API direction is agreed.

AKnassa:
## What this does
Makes the Table sort headers' screen-reader labels ("Sort by Name, sorted ascending, priority 1 of 2") translatable, like the rest of the component strings.

## Why
The i18n system and catalog are live, and the sort context-menu labels were already migrated — but the header button labels come from a plain helper whose return value slips past the hardcoded-string lint rule, so they stayed English in every locale. This closes the last concrete gap raised in #3618, under tracker #3636.

## What changed
- The three label forms now go through the translator with full-sentence messages (word order stays correct in every language; "ascending"/"descending" translate via their own entries, following the Chat precedent). ICU `select` inside the sentence was considered instead; the separate word key won for consistency with Chat's existing `status` pattern — one for the #3636 standardization discussion.
- 5 new catalog entries; English output is byte-identical, proven by the untouched existing tests.
- 3 new tests verify a French-style override, the multi-sort priority form, and the direction words. The overridden words deliberately differ from the raw enum values, so raw-enum interpolation or a hardcoded English frame fails the suite (mutation-verified: each doctored implementation fails exactly its test).

## How to see it
`pnpm exec vitest run packages/core/src/Table/plugins/sortable/useTableSortable.test.tsx` — 40/40 pass. Lint passes with no suppressions.

Fixes #3618

### DIFF
--- a/.changeset/table-sort-aria-i18n.md
+++ b/.changeset/table-sort-aria-i18n.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Route the Table sortable plugin's header-button aria-labels ("Sort by …", "… sorted …", "… priority … of …") through `useTranslator()` with new `@astryx.table.sort.sortBy` / `sortedBy` / `sortedByWithPriority` / `direction.*` catalog keys, so they localize like the sort menu labels already do (#3618, tracker #3636). The direction word resolves through its own key rather than interpolating the raw enum value. English output is unchanged.
+
+@AKnassa
--- a/packages/core/locales/en.json
+++ b/packages/core/locales/en.json
@@ -423,6 +423,26 @@
     "defaultMessage": "Clear sort",
     "description": "Aria label for a Table column header when clicking it will clear the current sort."
   },
+  "@astryx.table.sort.direction.ascending": {
+    "defaultMessage": "ascending",
+    "description": "Localized direction word interpolated as `direction` into @astryx.table.sort.sortedBy and @astryx.table.sort.sortedByWithPriority."
+  },
+  "@astryx.table.sort.direction.descending": {
+    "defaultMessage": "descending",
+    "description": "Localized direction word interpolated as `direction` into @astryx.table.sort.sortedBy and @astryx.table.sort.sortedByWithPriority."
+  },
+  "@astryx.table.sort.sortBy": {
+    "defaultMessage": "Sort by {label}",
+    "description": "Aria label for a sortable Table header button when the column is unsorted. `label` is the column header text."
+  },
+  "@astryx.table.sort.sortedBy": {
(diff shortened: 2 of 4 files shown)


humbertovirtudes (approved):
Approving. Clean closure of the last gap in #3618.

The three header-button label forms now route through `useTranslator()` with the new catalog keys, and resolving the direction word through its own `direction.*` entry rather than interpolating the raw enum is the correct call. English output is byte-identical, which the untouched existing tests confirm.

The new tests are mutation-resistant, not tautological: the French overrides (`croissant`/`decroissant`) deliberately differ from the raw enum values, so a hardcoded English frame or raw-enum interpolation fails the suite for both directions.

One thing worth a note, not blocking: interpolating a separately-translated word into a sentence carries the usual fragment-composition risk (gender and declension can shift the surrounding words in some languages). You flagged the ICU `select` alternative and chose the separate-key approach to match the existing Chat `status` precedent, deferring the standardization to #3636. That is a reasonable, precedent-backed tradeoff.

Thanks for the careful work here. CI is green.

Issue #3635, opened 2026-07-06
[Tracker] Chat components — Q3 maturation (composer, message/list, TTS)


cixzhang:
## Problem

Astryx ships a chat component set — `ChatComposer`/`ChatComposerInput`, `ChatMessage`, `ChatMessageList`, `ChatLayout` — but building a production chat surface today hits several gaps: the composer's state model, message/list alignment control, and read-back (TTS). This is a Q3 push to mature the chat components as a first-class surface, not a scattering of one-off tweaks.

This is a **tracker** for the major chat work landing this half. Individual enhancements are filed and linked below; each lands on its own via the normal issue → API → PR flow.

## Scope (on this milestone)

- [ ] #2473 — `ChatComposerInput`: internal-authoritative state model (Lexical-inspired)
- [ ] #2572 — `ChatMessageList`: `align`/`anchor` prop for top-aligned message lists
- [ ] #2574 — `ChatMessage`: align custom in-message content (cards, attachments) to the bubble's text column
- [ ] #1320 — Text-to-Speech: speak LLM responses back to the user

## Related bugs (linked, not on the milestone)

- #2573 — `ChatLayout`: phantom scrollbar in self-scroll mode when messages don't fill the viewport

## Notes

Each sub-item follows the spec protocol independently (problem → research → API → PR). This issue only tracks the theme so the milestone reflects "chat components" as one major line of work.

PR #4074, opened 2026-07-19
refactor(Chat): internal-authoritative composer input state

### LINKED ISSUE #2473: refactor(Chat): XDSChatComposerInput — internal-authoritative state model (Lexical-inspired)
## Background

#2397 fixed a real caret-collapse bug in `XDSChatComposerInput` by adding a `pendingEchoValueRef` one-shot marker to skip redundant DOM writes from the controlled-value sync effect. The fix is correct, but it's patching around a fundamental architectural tension: **a `contentEditable` element can't be a React controlled input**, and pretending it is creates the echo loop that the marker exists to suppress.

## The problem

The current flow:

```
User types → DOM updates → onInput → onChange(text) → parent setState
→ rerender with new value prop → useEffect fires → writes textContent
→ caret collapses to offset 0 (bug)
```

The echo-detection ref fixes the symptom, but the root cause is that the component treats the parent's `value` prop as the source of truth while the DOM is *also* the source of truth during active editing. Two sources of truth means reconciliation hacks.

## Proposed architecture (inspired by Lexical)

Lexical solves this cleanly: the editor owns an internal `EditorState` that is authoritative. All mutations — user input and programmatic — flow through the same pipeline (`editor.update()`), and the reconciler diffs model → DOM with selection as part of the model. React never touches the contentEditable DOM.

We don't need a full Lexical node tree. The lightweight version for `XDSChatComposerInput`:

### 1. Component owns its state

The component maintains an internal ref to its current text + selection state. This is the single source of truth.

### 2. `value` prop becomes a "desired value" signal

Instead of a React-controlled input where `value` drives the DOM on every render, treat it as an external override signal:

- If `value` changes to something different from the internal state, AND it wasn't something the component just emitted via `onChange` → apply it (genuine external override).
- Otherwise, ignore it.

This is conceptually what the `pendingEchoValueRef` does today, but framed as the correct mental model rather than a workaround.

### 3. Imperative handle for programmatic mutations

```tsx
interface ComposerHandle {
  /** Replace content + place caret at end. Same internal path as user input. */
  setValue(text: string): void;
  /** Insert at current caret position. */
  insertText(text: string): void;
  /** Get current text. */
  getText(): string;
}
```

Slash-command picks, mention insertions, etc. call `composerRef.current.setValue('/feedback ')` which:
1. Updates internal state
2. Writes DOM
3. Restores selection

All in one synchronous call — no effect, no echo, no stale prop race.

### 4. Selection is part of the internal model

After any mutation (user or programmatic), caret position is deterministically computed from the state change — not manually patched after a DOM write.

### 5. No double-render loop

```
User types → internal state updates → onChange fires (informational)
Parent may or may not call setValue() back — if they don't, nothing happens.
If they do, it goes through the same path as (3). No effect. No echo detection.
```

## API sketch

```tsx
interface XDSChatComposerInputProps {
  /** Initial/reset value. External overrides only apply when genuinely
   *  different from internal state. */
  value?: string;
  /** Fires on every user edit. Observational — parent doesn't need to
   *  "commit" the value back for the component to work. */
  onChange?: (text: string) => void;
  /** Imperative handle for programmatic mutations. */
  composerRef?: React.Ref<ComposerHandle>;
}
```

This is backward-compatible:
- Existing `value` + `onChange` usage continues to work
- The semantic shifts from "controlled" to "observed with external override" but the prop surface is identical
- `composerRef` is additive

## Tradeoffs

| | Current (controlled) | Proposed (internal-authoritative) |
|---|---|---|
| Mental model | React controlled input (familiar but wrong for contentEditable) | Component owns state, parent observes |
| Echo detection | Needed (`pendingEchoValueRef`) | Not needed — no echo loop exists |
| Caret preservation | Manual restoration after DOM write | Computed from state transition |
| Programmatic insert | Set `value` prop → wait for effect → hope caret is right | `composerRef.current.insertText()` — synchronous, correct |
| Complexity | Lower ceiling, higher floor (hacks accumulate) | Slightly more upfront design, fewer ongoing patches |

## Related

- #2397 — the caret-preservation fix that motivated this
- [Lexical's design doc](https://lexical.dev/docs/design) — full treatment of the "EditorState as source of truth" pattern
- [HN discussion on Lexical vs ProseMirror](https://news.ycombinator.com/item?id=31019778) — Lexical team explaining why they treat EditorState as authoritative over DOM

AKnassa:
## What this does
Reworks how the chat composer input keeps track of what's typed, following the architecture laid out in #2473: the editable area itself is now the single source of truth, and the `value` prop becomes a commit/override channel.

## Why
The old "skip exactly one echo" bookkeeping had a real hole: if the app committed an earlier value slightly late (easy under async state), the composer treated it as an external change and silently deleted the user's newest keystrokes. The new model fixes that class of bug and gives app code a clean programmatic way to set the draft.

## What changed
- A small ledger of not-yet-acknowledged edits replaces the one-shot echo marker; a late echo is recognized and skipped, a genuine override still applies cleanly.
- New `setValue` on the existing handle: writes the text and reports exactly one change — the same change pipeline as typing. It also re-evaluates the trigger menu the way typing does (a menu left open over replaced content closes, instead of aiming a later item pick at a stale trigger position in the new text), and it places the caret at the end only when the input has focus — writing a range into an unfocused editable would yank the document selection (and in Blink, focus itself) away from wherever the user is typing. Call `focus()` before `setValue` to continue typing at the end.
- Submit-clear goes through the same pipeline (identical observable behavior, plus proper token cleanup).
- Public props unchanged; every pre-existing test passes unmodified. 5 new tests pin the late-echo bug, the `setValue` contract, `setValue`'s own late-echo protection (mutation-verified: the prior suite still passed with the ledger bypassed in `setValue`), the unfocused-selection guard, and the stale-menu close.

## How to see it
`pnpm exec vitest run packages/core/src/Chat` — 142/142 pass. Full core suite green (6775/6775).

Fixes #2473 (part of Q3 Chat tracker #3635)


packages/core/src/Chat/ChatComposerInput.tsx
```diff
@@ -370,41 +453,37 @@ export function ChatComposerInput(props: ChatComposerInputProps) {
...
+    // Echo of one of our own `onChange` emissions — possibly a LATE
+    // one that lands after further edits (the parent committed
+    // emission N while internal state is already at emission ≥ N).
+    // Internal state is authoritative: consume the ledger through
+    // that entry and leave the DOM untouched. Matching the FIRST
+    // occurrence keeps later duplicate emissions consumable by their
+    // own echoes.
+    const ledger = pendingEmissionsRef.current;
+    const echoIndex = ledger.indexOf(controlledValue);
+    if (echoIndex !== -1) {
```
cixzhang, line 465: Matches anywhere in the ledger. Probably wants `echoIndex === 0` — only the oldest un-acknowledged emission can be an echo.

.changeset/chat-composer-internal-authoritative-state.md
```diff
@@ -0,0 +1,7 @@
+---
+'@astryxdesign/core': patch
```
cixzhang, line 2: This changes what `value` means for existing consumers. Might want `minor`.

cixzhang (changes requested):
Thanks — the late-echo hole is real and this reads it right. One thing before it lands.

Matching the echo anywhere in the ledger swallows a genuine override that equals an un-acknowledged emission. When the parent doesn't commit every `onChange` — which the new `value` doc blesses — the ledger holds every keystroke, so `value=''` after the user emptied the box once reads as an echo and the composer keeps its text: switching threads carries a half-typed message into the new one. `main` clears it.

Matching only the oldest un-acknowledged entry fixes it and keeps all 142 Chat tests green:

```ts
if (echoIndex === 0) {
  ledger.splice(0, 1);
  return;
}
```

Also `refactor` and a `patch` changeset both promise nothing consumer-visible changed, but `value` means something different now for everyone already passing it.

Does matching only the oldest entry still cover the async commit you had in mind?

[Reviewed by Robohands]

cixzhang:
Checking in on composer state: echo matching still needs to consume only the oldest pending emission so a genuine parent override is not swallowed. Would you like to continue, have us carry it, or close this for now?

PR #4082, opened 2026-07-19
docs(review): add a test-quality bar for test PRs

cixzhang:
## Why

Standalone test PRs are landing frequently, but the review guidance funnels them into the **test/docs/chore fast path** — where "a passing test exists" is effectively the whole gate. That lets low-value tests (change-detectors, snapshot dumps, coverage padding) through without scrutiny.

## What

Adds an explicit **test-quality bar** to `packages.instructions.md` so reviewers (and Copilot) judge test PRs on **contract vs. implementation**, not just existence:

- New **Tests** entry under *Calibrate to the PR type* with ✅ worth-testing / 🚫 test-slop lists and a one-line heuristic (*does this protect a promise a consumer relies on, or just mirror the code?*).
- Tightens the **fast-path** definition: a standalone test PR is no longer an automatic fast-path approve — a passing test is necessary, not sufficient.

Guidance-only (`.github/`), no consumer-visible change, so no changeset.

### DIFF
--- a/.github/instructions/packages.instructions.md
+++ b/.github/instructions/packages.instructions.md
@@ -40,10 +40,13 @@ in what order, so effort lands where the risk is — and so the risk checks
 
 **3. Pick the review path** (depth follows category × risk):
 
-- **Fast path** — test/docs/chore, or a small behavior-only bug fix with a
+- **Fast path** — docs/chore, or a small behavior-only bug fix with a
   regression test, no breaking change, low blast radius. Verify accuracy (does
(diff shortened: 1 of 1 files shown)


josephfarina (approved):

josephfarina:
Heads up: build / build-storybook / test are red, but the failures are all in PowerSearch/formatFilterValue.ts ("TypeError: t is not a function"), which this docs-only PR doesn't touch — looks like a pre-existing break on main. Worth a rebase once that's fixed so the checks go green. Content of the change is clean.

PR #4083, opened 2026-07-19
feat(docsite): enhance BlogShowcase component to handle empty and single post scenarios

Unmesh100:
Before the fix, zero posts made featurePost equal undefined, but the component still
  rendered: < BlogFeatureCard post={featurePost} /> 

  BlogFeatureCard immediately accesses fields like post.title, so the homepage crashes.
  The guard prevents that.

  The one-post layout is included so it doesn’t render an empty right-hand column
  intended for additional posts.


imdreamrunner:
We might never going back to have fewer than 3 blog posts. But no harm I guess?

imdreamrunner (approved):

cixzhang (approved):
Thanks — the zero-post guard is right: today `blogPosts.slice(0, 3)` hands `BlogFeatureCard` an undefined post and it dereferences `post.title` immediately. The one thing left before this can merge is your CLA signature.

[Reviewed by Robohands]

Unmesh100:
> Thanks — the zero-post guard is right: today `blogPosts.slice(0, 3)` hands `BlogFeatureCard` an undefined post and it dereferences `post.title` immediately. The one thing left before this can merge is your CLA signature.
> 
> [Reviewed by Robohands]

@cixzhang Thanks. and also CLA signature is done.

cixzhang:
Merged as `94c2d8b` — thank you, @Unmesh100. The zero-post guard is the real catch here: an empty registry handed `BlogFeatureCard` an undefined post and it dereferenced `post.title` on its first line, so a deployment with no published posts took the whole home page down. Thanks too for coming back and signing the CLA.

The branch had aged 35 days while it waited on us, so we merged `main` in for you and re-ran CI rather than asking you to rebase — that wait was ours, not yours. Verified at the merged head: empty registry now renders nothing, a single post gets the full-width feature card, and the two- and three-post layouts are unchanged.

[Reviewed by Robohands]

PR #4087, opened 2026-07-19
feat: Formentor concept + 4-way form-framework evaluation + playground

ernestt:
## Summary

Adds **Formentor** as a concept package for Astryx plus a fair, controlled evaluation of it against three established form frameworks, and an interactive playground to inspect the results by hand.

The core question: Astryx ships great inputs but takes no position on form state/validation. Should it adopt Formentor (a schema-aware form layer) as the default, or bless an existing library? This PR builds the concept, the comparison harness, runs the baseline, and ships tooling to judge the outcome.

## What's included

**1. Formentor scaffold** (`packages/formentor`) — a Level-B schema-aware form layer: runtime schema builders (`object`/`string`/`number`/`boolean`/`enumOf` + `nullable`/`omissible`), `useFormentorForm` with a renderable model, a validation ladder (schema → field → cross-field → submit-time async), view/edit mode, and an XDS InputSet binding schema types to existing Astryx components. 10 tests.

**2. Equal-effort adapters** (`packages/astryx-formisch`, `-tanstack`, `-rhf`) — thin bridge components binding each framework's field primitive to the *same* Astryx inputs, so the comparison varies only the framework, not the design system (Option B parity). 6 tests.

**3. Form-framework vibe test** (`internal/vibe-tests`) — a 4-way harness on a constant Astryx design system:
- 4 representative project templates + a per-target environment builder
- `setup-forms.mjs` runner with CORE/STRETCH tiers and blank-slate + no-answer-leak invariants (#6/#7)
- `test-sets/forms.json` — 15 CORE + 6 STRETCH naive prompts
- `form-eval.ts` — 5 form-specific dimensions with an anti-bias calibration test
- `form-aggregate.ts` — blind aggregator that also reports tsc compile rate

**4. Interactive playground** (`mode=playground`) — for judging forms by eye instead of an automated design judge: pick a prompt, swap frameworks, see the live rendered form, the verbatim generated code, and a live state panel (values/errors/touched/isValid/submit payload).

**5. Decision memo** (`internal/vibe-tests/FORMENTOR-DECISION-MEMO.md`) — the CORE baseline and recommendation.

## Baseline result (15 CORE prompts × 4 frameworks = 60 blank-slate forms)

| Framework | Clean tsc compile | Form-quality | Median LOC | Doc lookups |
|---|---|---|---|---|
| Formentor | 87% | 98 | **49** | **4.5** |
| Formisch | 93% | 100 | 56 | 5.9 |
| TanStack | 93% | 100 | 75 | 6.2 |
| React Hook Form | 87% | 100 | 57 | 5.1 |

All four cluster on correctness and near-tie on quality; the separators are conciseness and cognitive load, where Formentor leads. **Recommendation:** adopt Formentor as the default, gated on (1) fixing `isValid` semantics for "disable until valid" and (2) growing STRETCH-tier coverage; Formisch is the strongest already-complete fallback.

A first-pass run showed TanStack/RHF far lower — traced to a Zod v3-vs-v4 resolution confound in the harness (documented and fixed), which is exactly what the fairness invariants are designed to catch.

## Testing

- 24 unit tests pass (10 Formentor runtime + 6 adapters + 8 evaluator incl. anti-bias calibration)
- All 4 new packages typecheck clean and build
- Playground: all 60 generated solutions resolve and transform under the playground Vite config

## Notes

- Concept scope is Level-B scalar forms; STRETCH patterns (field arrays, nested, wizards, typeahead, upload) are defined but not yet supported by the scaffold.
- All new packages are `private` (never published); the playground's generated `.playground/` output is gitignored.


internal/vibe-tests/FORMENTOR-DECISION-MEMO.md
```diff
@@ -0,0 +1,120 @@
+# Decision Memo: Which form framework should be Astryx's default?
```
cixzhang, line 1: Thanks for the test. Can we move this to an issue instead so it's easier to consume and comment on? Ideally the results it should not be in the repo.

cixzhang (changes requested):
Thanks for running the vibe test. I think these files were accidentally committed to the repo. We would want to instead preserve the results in an issue instead.

cixzhang:
I'm curious how the evaluator determines form quality. Is that based on design heuristics?

ernestt:
Thanks for the review — agreed on both counts. These are evaluation artifacts and shouldn't land in the repo.

I've moved the results into an issue so they're easy to read and comment on: #4123.

Closing this PR. The harness, the `@astryxdesign/formentor` concept scaffold, the three parity adapters, and the interactive playground stay on branch `feat/formentor-scaffold` (unmerged) purely as a reproduction reference for the numbers in the issue — including running the advanced-tier prompts later. Nothing here is intended to merge into `main`.

PR #3841, opened 2026-07-11
chore(deps): bump production dependencies (stylex 0.19, recharts 3.9.2, clack 1.7)

cixzhang:
Supersedes #3835 (the Dependabot production-dependencies group PR), which failed CI.

## What
Bumps the production dependencies to latest:
- `@stylexjs/stylex` 0.18.3 → 0.19.0 (across all packages + example apps)
- `recharts` 3.8.1 → 3.9.2
- `@clack/prompts` 1.5.1 → 1.7.0

## Why the Dependabot PR failed
Dependabot bumped `@stylexjs/stylex` in the apps and `packages/core`, but the
`>=0.10.0` peer ranges in `packages/charts` and `packages/lab` kept resolving a
stale 0.18.3 copy. Two copies of stylex in the tree produce two distinct
`StyleXClassName` opaque types, so passing a style from one package to a
component typed against the other fails typechecking (`Types of property
'_opaque' are incompatible`). This broke the storybook typecheck and the
`packages/lab` build.

## Fix
Add a workspace `overrides` entry pinning `@stylexjs/stylex` to `^0.19.0` so the
entire monorepo resolves a single stylex version, and bump the direct specifiers
consistently. Verified locally: `pnpm build`, the storybook/core/cli typechecks,
and `pnpm check:sync` all pass.

PR #2197, opened 2026-05-18
[RFC] migrate from yarn 1 to pnpm

zurfyx:
## Why now

The team-wide direction has been moving toward pnpm (Nest, etc). XDS itself was originally on pnpm and was switched to yarn in #6 with no commit message rationale, so this PR opens that decision for re-evaluation.

The /packages/cli already detects all four package managers (yarn, pnpm, bun, npm) for *consumers*, so the consumer-facing surface is unaffected by what XDS uses internally — this only changes how the monorepo itself is installed and built.

## What changes

- packageManager: 'yarn@1.22.22' → 'pnpm@9.15.4' (Corepack auto-installs)
- workspaces array → pnpm-workspace.yaml
- yarn.lock → (will be) pnpm-lock.yaml (not generated in this PR — see below)
- All 11 GitHub Actions workflows: add pnpm/action-setup@v4, switch to cache: 'pnpm', pnpm install --frozen-lockfile
- All 'yarn workspace X cmd' → 'pnpm -F X cmd'
- All script invocations (yarn build/test/lint/etc) → pnpm
- Yarn 'resolutions' → 'pnpm.overrides' (translated yarn glob syntax to pnpm's parent>child form)
- .npmrc: shamefully-hoist=true + node-linker=hoisted as the conservative migration default. This makes pnpm behave like yarn for hoisting, preserving build-time tooling that walks node_modules (StyleX babel plugin, Storybook). Can tighten later.
- Docs swept: CONTRIBUTING.md, README.md, CLAUDE.md, .claude/commands/, per-package READMEs, sandbox docs page

## What is intentionally kept on yarn

- packages/cli/src/utils/package-manager.mjs — the detector, must keep recognizing yarn lockfiles for consumers
- packages/cli/src/commands/upgrade.mjs — install command map per PM
- packages/cli/src/codemods/ensure-jscodeshift.mjs — same
- internal/vibe-tests/test-sets/default.json — vibe test prompts that ASK the LLM about yarn (changing them changes the test)
- README.md install snippet — shows pnpm/npm/yarn for consumers
- packages/cli/templates/blocks/components/ChatToolCalls/* — UI examples containing strings like 'yarn test' as synthetic example content

---

cc @hurrymaplelad (PR #6 author)

PR #4101, opened 2026-07-20
chore(deps): prune stale pnpm overrides, move workspace pins to the catalog

### LINKED ISSUE #4338: pnpm overrides block has 13 accumulated entries with no record of which are still load-bearing
## Problem

The `overrides` block in `pnpm-workspace.yaml` has grown to 13 entries accumulated over roughly six months:

- yarn-resolutions leftovers from #2197
- security floors from #1796, #2194, #2456, #2837, #3220, #3607
- version alignment from #3841

Nothing records which of them are still doing anything. Every dependency upgrade has to fight an opaque wall of forced resolutions, with no way to tell a live security floor from a fossil that upstream fixed months ago.

## What an audit turns up

Verifying each entry empirically — remove it, `pnpm install --lockfile-only`, diff the resolved versions, then `pnpm audit`:

- **9 of 13 are no-ops or dedup-only.** The graph resolves to the same (or a floor-satisfying) version without them.
- **3 are live security floors** — `postcss` (pinned low by next@15), `dompurify` (monaco-editor@0.55), `esbuild` (tsup@8.5). These are the only entries whose removal introduces advisories: 0 vulnerabilities with all overrides in place, 18 with none, all attributable to exactly these three.
- **1 is actively harmful.** `@changesets/parse>js-yaml >=4.2.0` was force-feeding an out-of-range js-yaml **5.2.0** major into a `^4` consumer, when 4.3.0 satisfies the floor on its own.

The three live floors also carry no in-file record of *which* dependent pins the vulnerable version or *when* the override becomes removable — so the same audit has to be redone from scratch the next time someone touches the block.

---

Fix in #4101.

Han5991:
## Why

The `overrides` block in `pnpm-workspace.yaml` had grown to 13 entries accumulated over 6 months (yarn-resolutions leftovers from #2197, security floors from #1796/#2194/#2456/#2837/#3220/#3607, version alignment from #3841). Nothing recorded which of them were still doing anything, making every future dependency upgrade fight an opaque wall of forced resolutions.

## Method

Each entry was verified empirically: remove the override, `pnpm install --lockfile-only`, compare resolved versions, then `pnpm audit` (baseline with all overrides: **0 vulnerabilities**; with all overrides removed: **18** — all attributable to exactly three entries).

## Removed (9 entries)

No-ops — the graph resolves to the same (or a floor-satisfying) version without them:

| Override | Natural resolution | Origin |
|---|---|---|
| `@stylexjs/stylex ^0.19.0` | 0.19.0 (unchanged) | #3841 alignment |
| `tmp >=0.2.6` | 0.2.7 | #2456 CVE, upstream caught up |
| `@babel/core ^7.29.6` | 7.29.7 | #3220 |
| `hono ^4.12.25` | 4.12.27 | #3220 |
| `read-yaml-file>js-yaml ^3.15.0` | 3.15.0 | #3607, v3 line patched upstream |
| `@changesets/parse>js-yaml >=4.2.0` | 4.3.0 — satisfies the floor by itself | #3607; the override was force-feeding an out-of-range js-yaml **5.2.0** major into a `^4` consumer |
| `picomatch ^4.0.4` + `micromatch>picomatch` + `anymatch>picomatch` | identical 2.3.2 + 4.0.4 split | yarn-resolutions leftovers (#2197) |

Dedup-only forces — removal restores each consumer's own declared range (audit-neutral):

- `prettier ^3.9.3` — `@changesets/write` gets its declared prettier 2.8.8 back (internal formatting only; repo formatting uses the root devDep, now `catalog:`).
- `minimatch ^9.0.7` — `glob@7` returns to its v3-API minimatch 3.1.5 instead of being forced onto v9; `glob@13` gets 10.2.5.

## Kept (3 live security floors, now documented in-file)

Each remaining entry now records the dependent that pins the vulnerable version and its deletion condition:

| Override | Blocks | Removable when |
|---|---|---|
| `postcss ^8.5.10` | next@15 → 8.4.31 (`<8.5.10` advisory) | Next raises its postcss |
| `dompurify ^3.4.11` | monaco-editor@0.55 → 3.2.7 (16 advisories; strictest patch line is exactly 3.4.11) | monaco-editor catches up |
| `esbuild >=0.28.1` | tsup@8.5 → 0.27.7 (0.27.3–0.28.0 advisory) | tsup raises its esbuild range |

## Catalog conversion

Workspace-wide pins that the pruned overrides had been enforcing move to the pnpm `catalog:` next to the existing vite entry, so each has a single bump point: `@stylexjs/stylex`, `@stylexjs/babel-plugin`, `prettier`, `esbuild` (7 declarations across root, packages/core, packages/build, apps/sandbox, apps/docsite). Peer ranges are deliberately untouched — they are consumer compatibility contracts, not workspace pins. The three security floors above cannot move to the catalog: `catalog:` only applies where a workspace package declares the dependency directly, and these guard transitive paths inside next/monaco-editor/tsup.

## Verification

- `pnpm audit`: 0 vulnerabilities before and after.
- Resolved versions after the change are byte-identical for every pruned no-op; the only graph changes are the two documented dedup splits and js-yaml 5.2.0 → 4.3.0.
- Full `node` test project (CLI, build tooling, changesets checks — the consumers of glob/minimatch/js-yaml/prettier): 2129/2129 pass.
- `check:repo` (sync, package boundaries, changesets, demo media, executable bits): pass.

---

Closes #4338


Han5991:
Rebased onto `main` (08c99b8) to resolve the `pnpm-workspace.yaml` conflict. The lockfile was regenerated with `pnpm install` rather than hand-merged.

**How the conflict was resolved**

This PR's pruning and catalog move are kept as-is. Every security floor `main` gained after this branch was cut is preserved, since they postdate the audit in the original commit message:

- `postcss` `^8.5.10` → `^8.5.18`, `dompurify` `^3.4.11` → `^3.4.12` (#4398)
- `fast-uri ^3.1.4`, `brace-expansion ^2.1.2` (#4398)
- `sharp ^0.35.3` (#4449)
- the `read-yaml-file` 2.x lift

The pruned entries `main` did not touch (`picomatch` trio, `minimatch`, `tmp`, `@babel/core`, `hono`, and the `prettier`/`@stylexjs/stylex` catalog moves) stay pruned.

**Overlap with #4402**

While verifying, `pnpm audit` flagged two js-yaml advisories that exist on `main` today: the `'@changesets/parse>js-yaml': '>=4.2.0'` floor is unbounded, so it resolves the chain onto js-yaml 5.2.1, inside the vulnerable `>=5.0.0 <=5.2.1` range.

#4402 already fixes exactly this by moving both js-yaml chains to 5.2.2 and excluding `js-yaml@5.2.2` from `minimumReleaseAge`. I've adopted that same fix here so this branch does not regress the floor, with one deviation: the ranges are written as `^5.2.2` instead of `>=5.2.2`. An unbounded `>=` is what dragged this chain onto a vulnerable major in the first place, and a future js-yaml 6 would repeat it.

**If #4402 lands first**, this becomes a no-op on those two lines — happy to drop them and rebase instead. Sequencing is up to you.

**Verification**

- `pnpm audit`: 12 → 10 findings. The two js-yaml advisories clear; the remaining 10 (`next` ×8, `@hono/node-server`, `brace-expansion`) are identical to `main`'s baseline, confirmed by auditing a clean worktree at 08c99b8.
- `pnpm changeset status` runs clean — no `safeLoad` error, so `read-yaml-file` 2.x is fine on js-yaml 5.
- `pnpm check:repo` passes; `@astryxdesign/build` and `@astryxdesign/core` build (5719 StyleX rules, CSS + UMD emitted).

Han5991:
Lint was failing on the rebase — pushed a fix.

**What broke**

Pruning `minimatch: ^9.0.7` let `@eslint/config-array` resolve minimatch 10, which calls the *named* `expand` export. `main`'s `brace-expansion: ^2.1.2` floor (added in #4398, after this branch was cut) pins brace-expansion to 2.x, which only has the default CJS export:

```
TypeError: (0 , brace_expansion_1.expand) is not a function
    at braceExpand (node_modules/minimatch/dist/commonjs/index.js:157:41)
    at doMatch (node_modules/@eslint/config-array/dist/cjs/index.cjs:422:13)
```

**Why the fix is to drop `brace-expansion` too, not to keep `minimatch`**

`^2.1.2` sits below the declared range of every consumer in the graph — `minimatch@10` (via `@eslint/config-array`) declares `^5.0.5`, `minimatch@3` (via `@babel/cli`) declares `^1.1.7`. It was a downgrade, not a floor, and it only stayed invisible while the equally stale `minimatch` force held minimatch on a line that still used the default export.

Unforced, each consumer resolves inside its own range — brace-expansion 5.0.7 and 1.1.16 — and ESLint runs clean. So both entries are pruned rather than one restored.

**Verification**

- `pnpm lint`: 0 errors. (Locally it also reports ~23.5k errors under `apps/docsite/public/monaco/` and `public/vendor/`, but both are gitignored, untracked build artifacts that a fresh CI checkout does not have — the count is identical with minimatch 9 or 10, so it is unrelated noise.)
- `pnpm test`: 393 files / 7933 tests pass.
- `pnpm audit`: 10 findings, unchanged from the previous push and 2 fewer than `main`'s 12.
- `pnpm check:repo`, `@astryxdesign/build` and `@astryxdesign/core` builds all pass.

The commit message and the comments in `pnpm-workspace.yaml` are updated accordingly — the original message described `minimatch` as a dedup-only force, which was right about the *outcome* but missed that it was entangled with the brace-expansion floor.

pnpm-workspace.yaml
```diff
@@ -29,39 +31,47 @@ minimumReleaseAgeExclude:
...
+  # Host confusion via backslash (#4398). Drop when dependents reach 3.1.4.
   fast-uri: ^3.1.4
-  brace-expansion: ^2.1.2
-  # next pulls sharp as an optional dependency for image optimization and still
-  # resolves 0.34.5, which inherits the libvips advisories CVE-2026-33327,
-  # CVE-2026-33328, CVE-2026-35590 and CVE-2026-35591. They are patched in
-  # sharp 0.35.0 — a semver-major bump on the 0.x line, so next's `^0.34`
-  # range can never reach it on its own. Drop this entry once next depends on
-  # 0.35 itself.
+  # Do not re-add brace-expansion ^2.1.2 (#4398): 2.x is below every consumer's
```
cixzhang: This is right about 2.x, but it reads as "no floor" — ^5.0.9 is the one to add here.

apps/docsite/package.json
```diff
@@ -34,15 +34,15 @@
...
     "react": "^19.2.7",
     "react-dom": "^19.2.7",
     "zod": "^4.4.3"
   },
   "devDependencies": {
     "@babel/preset-react": "^8.0.1",
     "@babel/preset-typescript": "^7.29.7",
-    "@stylexjs/babel-plugin": "^0.19.0",
+    "@stylexjs/babel-plugin": "catalog:",
     "@stylexjs/postcss-plugin": "^0.19.0",
```
cixzhang: Same StyleX toolchain, still literal — a catalog bump will silently split it from the runtime.

cixzhang (changes requested):
Thanks for doing this the careful way — pulling each entry, re-resolving and re-auditing is the right method, and it shows: on your own base the prune takes `pnpm audit` from 26 findings to 22.

Two of the entries stopped being safe to drop while this sat, though. Main floors brace-expansion at ^2.1.4 and hono at ^4.12.34 now, and this deletes both — putting them back gets you to 17.

You're right that ^2.1.2 downgrades and takes ESLint out with it. The floor belongs on the 5.x line, inside minimatch 10's own ^5.0.5, so nothing else moves:

```yaml
brace-expansion: ^5.0.9
hono: ^4.12.34
```

Main's overrides block has changed a lot since July and this won't merge clean regardless — could you merge main and re-run your removal pass on top of it?

[Reviewed by Robohands]

cixzhang (approved):
Thanks — the re-verify is thorough, and dropping `@hono/node-server` via the SDK bump is better than flooring it. Merging.

Nit: `apps/template-viewer/package.json:15` stylex is still literal — it'll drift off the catalog.

[Reviewed by Robohands]

PR #3730, opened 2026-07-09
fix(core): respect prefers-reduced-motion in streaming text

bhamodi:
## Summary

Streaming text ignored `prefers-reduced-motion`, against this repo's own convention. Two coordinated gaps drove animation for users who opted out of motion:

- `useStreamingText` (`hooks/useStreamingText.ts`) runs a per-character `requestAnimationFrame` reveal loop (the animation effect at `useStreamingText.ts:147-177` pre-fix) with no reduced-motion check — the text always animated in char-by-char.
- `Markdown`'s streaming `fadeIn` style (`Markdown/Markdown.tsx:424-434` pre-fix) applies an opacity transition plus an `@starting-style` entry fade with no reduced-motion override.

Both animate regardless of the user's preference, while sibling components (`Spinner`, `Skeleton`, `ProgressBar`, `StatusDot`, `Chat/ChatMessageMetadata`) already neutralize their animations under `@media (prefers-reduced-motion: reduce)`.

This makes streaming honor the preference: the reveal snaps straight to the full text (and keeps snapping as later chunks arrive), and the entry fade collapses to an instant swap.

## Changes
- `hooks/useStreamingText.ts`: read `prefers-reduced-motion` via the package's SSR-safe `useMediaQuery` hook (`hooks/useMediaQuery.ts`) and short-circuit like `speed: 'instant'` — skip the RAF loop and return the full `targetText`.
- `Markdown/Markdown.tsx`: give `streamingStyles.fadeIn.transitionDuration` a `'@media (prefers-reduced-motion: reduce)': '0s'` conditional value, mirroring the conditional-duration form in `Spinner.tsx:77-80`. At `0s` the `@starting-style` opacity jump becomes non-animated (the media query cannot nest inside `@starting-style`, so collapsing the duration is the effective override).
- `hooks/useStreamingText.test.ts`: per-test-controllable `matchMedia` mock (restored in `afterEach` so it never leaks) plus two reduced-motion tests.

## matchMedia approach
[...]

PR #4104, opened 2026-07-20
fix(chat): respect prefers-reduced-motion in the scroll-follow spring

yyq1025:
## Summary

Follow-up promised in #3800: the Chat scroll-follow spring in `useChatStreamScroll` ran its rAF animation regardless of `prefers-reduced-motion`.

Every spring entry point — the locked growth-follow (`scrollIfLocked`), `scrollToBottom()`'s default `'spring'` behavior, and `lock()` — funnels through `startAnimation`, so this adds a single branch there: when the user prefers reduced motion, fall back to the existing `jumpToBottom`. Following the bottom still works exactly as before; it just lands in one frame instead of animating the travel. `scrollToBottom({behavior: 'instant'})` and the first-fill positioning were already instant and are unchanged.

Reads the preference via the package's SSR-safe `useMediaQuery`, matching the convention already used by `useStreamingText` (#3730), Spinner, Skeleton, and Layer.

## Test plan

- Two new unit tests in `useChatStreamScroll.test.tsx`: post-first-fill growth positions synchronously with no animation frames scheduled, and default `scrollToBottom()` jumps instantly and re-locks. The matchMedia stub mirrors the pattern in `useStreamingText.test.ts`.
- `vitest run packages/core/src/Chat/useChatStreamScroll.test.tsx` — 9/9 pass; eslint + prettier clean on the touched files.

### DIFF
--- a/.changeset/chat-spring-reduced-motion.md
+++ b/.changeset/chat-spring-reduced-motion.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Chat/useChatStreamScroll: the scroll-follow spring now respects `prefers-reduced-motion` — locked following, `scrollToBottom()`, and `lock()` fall back to the existing instant jump, so the transcript still tracks the bottom without animated travel. Follow-up promised in #3800.
+@yyq1025
--- a/packages/core/src/Chat/useChatStreamScroll.test.tsx
+++ b/packages/core/src/Chat/useChatStreamScroll.test.tsx
@@ -45,6 +45,23 @@ afterEach(() => {
   vi.unstubAllGlobals();
 });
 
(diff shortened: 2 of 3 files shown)


humbertovirtudes (approved):
Nice fix. Funneling every spring entry point (scrollToBottom, lock, scrollIfLocked growth-follow) through the single `startAnimation` branch so one reduced-motion check covers them all is the right consolidation, and falling back to the existing instant `jumpToBottom` means following still tracks the bottom, it just does not animate the travel. Reusing the shared `useMediaQuery` hook rather than hand-rolling a matchMedia listener is the correct call.

The two red-to-green tests (streaming growth lands synchronously, default scrollToBottom jumps with an empty rAF queue) pin the contract well, and the matchMedia stub mirrors the existing `useStreamingText` test pattern. Changeset present. Approving.

PR #4108, opened 2026-07-20
fix(cli): rewrite BaseProps import to core package barrel in swizzle (#4091)

### LINKED ISSUE #4091: [Bug] swizzle generates a BaseProps subpath import that is not exported
## Description

`astryx swizzle Button` generates this type import:

```ts
import type {BaseProps} from '@astryxdesign/core/BaseProps';
```

However, `@astryxdesign/core` does not define `./BaseProps` in its `package.json` exports. Node package resolution reports `ERR_PACKAGE_PATH_NOT_EXPORTED`, and TypeScript reports `TS2307`.

The generated source should import `BaseProps` through a public package export. This could be addressed either by changing the swizzle import rewrite or by exporting the `./BaseProps` subpath.

## Reproduction

1. Create a fresh Next.js 16 application.
2. Install Astryx v0.1.6.
3. Run:

   ```bash
   astryx swizzle Button
   ```

4. Inspect the generated Button source. It imports `BaseProps` from `@astryxdesign/core/BaseProps`.
5. Run TypeScript. It reports `TS2307` because that subpath is not exported.

The same result was reproduced with v0.1.4.

### Workaround

Change only the generated import to the package barrel:

```ts
import type {BaseProps} from '@astryxdesign/core';
```

After this one-line change, TypeScript exits with code 0. The import is type-only, so this workaround does not change runtime behavior.

## Astryx Version

`@astryxdesign/core@0.1.4` and `@astryxdesign/core@0.1.6`

## Environment

Fresh Next.js 16 application

Geervan:
Description
Running `astryx swizzle Button` generates the following type import:

import type {BaseProps} from '@astryxdesign/core/BaseProps';

However, `@astryxdesign/core` does not define `./BaseProps` in its package.json exports. Node package resolution reports ERR_PACKAGE_PATH_NOT_EXPORTED and TypeScript reports TS2307.

This happens because the swizzle command's import rewriter converts relative imports like `../BaseProps` into subpath imports (`@astryxdesign/core/BaseProps`) instead of using `@astryxdesign/core`, where BaseProps is re-exported.

Fix
Updated `rewriteImports` in the CLI swizzle command to rewrite `../BaseProps` to `@astryxdesign/core`. Also added `./BaseProps` to `@astryxdesign/core`'s package.json exports map for subpath compatibility, and added unit test coverage in `swizzle.test.mjs`.

Closes
Closes #4091

Verification
Verified that unit tests for swizzle (`swizzle.test.mjs`) pass successfully.
Verified that swizzling Button in the Next.js example app generates `import type {BaseProps} from '@astryxdesign/core';` and `npx tsc --noEmit` passes with 0 errors.


josephfarina:
dupe of #4111. Thank you for the contribution though sorry!

Geervan:
but my pr was raised earlier?, if you're going by that implementation i totally get that, happy to contribute nevertheless!!

PR #4111, opened 2026-07-20
fix(cli): route BaseProps swizzle imports through root

### LINKED ISSUE #4091: [Bug] swizzle generates a BaseProps subpath import that is not exported

mturac:
## Summary

- route swizzled core `BaseProps` imports through the public package entry point
- preserve the existing import rewrite behavior for integration packages
- add regression coverage for both paths

## Verification

- `pnpm exec vitest run packages/cli/src/commands/swizzle.test.mjs packages/cli/src/commands/swizzle.routing.test.mjs packages/cli/src/commands/swizzle.path-safety.test.mjs --pool=forks --maxWorkers=1 --testTimeout=20000`
- `pnpm lint`
- `pnpm -F @astryxdesign/cli typecheck:json-api`
- fresh-project `astryx swizzle Button` smoke test

Fixes #4091

### DIFF
--- a/packages/cli/src/commands/swizzle.mjs
+++ b/packages/cli/src/commands/swizzle.mjs
@@ -63,12 +63,11 @@ export function rewriteImports(content, ownerPackage = CORE_PACKAGE) {
       const parts = importPath.replace(/^\.\.\//, '').split('/');
       const topDir = parts[0];
 
-      // Map to the owner package subpath
-      const target =
-        ownerPackage === CORE_PACKAGE && topDir === 'BaseProps'
-          ? ownerPackage
-          : `${ownerPackage}/${topDir}`;
-      return `${prefix}${target}${suffix}`;
+      // Map BaseProps through the core barrel; other imports keep using owner subpaths.
+      if (ownerPackage === CORE_PACKAGE && topDir === 'BaseProps') {
+        return `${prefix}${ownerPackage}${suffix}`;
+      }
+      return `${prefix}${ownerPackage}/${topDir}${suffix}`;
     },
   );
 }
--- a/packages/cli/src/commands/swizzle.mjs
+++ b/packages/cli/src/commands/swizzle.mjs
@@ -64,7 +64,11 @@ export function rewriteImports(content, ownerPackage = CORE_PACKAGE) {
       const topDir = parts[0];
 
       // Map to the owner package subpath
-      return `${prefix}${ownerPackage}/${topDir}${suffix}`;
+      const target =
+        ownerPackage === CORE_PACKAGE && topDir === 'BaseProps'
+          ? ownerPackage
+          : `${ownerPackage}/${topDir}`;
+      return `${prefix}${target}${suffix}`;
     },
   );
 }
--- a/packages/cli/src/commands/swizzle.test.mjs
+++ b/packages/cli/src/commands/swizzle.test.mjs
@@ -13,7 +13,23 @@ describe('rewriteImports', () => {
   it('rewrites ../utils/mergeProps to @astryxdesign/core/utils', () => {
     const input = `import { mergeProps } from '../utils/mergeProps';`;
     const result = rewriteImports(input);
-    expect(result).toBe(`import { mergeProps } from '@astryxdesign/core/utils';`);
+    expect(result).toBe(
+      `import { mergeProps } from '@astryxdesign/core/utils';`,
+    );
+  });
+
+  it('rewrites ../BaseProps to the public package entry point', () => {
+    const input = `import type {BaseProps} from '../BaseProps';`;
+    const result = rewriteImports(input);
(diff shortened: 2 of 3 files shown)


packages/cli/src/commands/swizzle.test.mjs
```diff
@@ -13,7 +13,23 @@ describe('rewriteImports', () => {
   it('rewrites ../utils/mergeProps to @astryxdesign/core/utils', () => {
     const input = `import { mergeProps } from '../utils/mergeProps';`;
     const result = rewriteImports(input);
-    expect(result).toBe(`import { mergeProps } from '@astryxdesign/core/utils';`);
+    expect(result).toBe(
+      `import { mergeProps } from '@astryxdesign/core/utils';`,
+    );
```
josephfarina: Nit (non-blocking): this reformat is unrelated to the fix. CLI `.mjs` isn't Prettier- or width-enforced here — lint-staged only formats `*.{ts,tsx,md}` and there's no `prettier --check` in CI — so the original one-liner is fine as-is. Reverting keeps the diff to just the two new tests.

```suggestion
    expect(result).toBe(`import { mergeProps } from '@astryxdesign/core/utils';`);
```

packages/cli/src/commands/swizzle.mjs
```diff
@@ -64,7 +64,11 @@ export function rewriteImports(content, ownerPackage = CORE_PACKAGE) {
       const topDir = parts[0];
 
       // Map to the owner package subpath
-      return `${prefix}${ownerPackage}/${topDir}${suffix}`;
+      const target =
+        ownerPackage === CORE_PACKAGE && topDir === 'BaseProps'
+          ? ownerPackage
+          : `${ownerPackage}/${topDir}`;
+      return `${prefix}${target}${suffix}`;
```
josephfarina: Optional (take it or leave it): the `const target` ternary reads a little awkwardly. A guard clause is a bit clearer and leaves the existing return line untouched:

```js
      if (ownerPackage === CORE_PACKAGE && topDir === 'BaseProps') {
        return `${prefix}${ownerPackage}${suffix}`;
      }
      return `${prefix}${ownerPackage}/${topDir}${suffix}`;
```

If you want the intent self-documenting without a comment, hoist the name to module scope: `const BARREL_ONLY_ROOT_MODULES = new Set(['BaseProps']);` and test with `.has(topDir)`.

josephfarina (approved):
Approve — right fix at the right layer. `rewriteImports` is scoped to `@astryxdesign/core`, so the integration routing is preserved; it ships a changeset; and it covers both the core-barrel and integration-subpath cases.

Importantly it does *not* add a `./BaseProps` subpath to `packages/core/package.json` — that approach fails the `sync-exports` check (the exports map is generated from `src/`), which is what's currently red on the alternative PR (#4108).

Two nits inline; only the first is a (still non-blocking) ask.

Not asking for: no package.json exports change, and no generalizing the check beyond `BaseProps` — `../naming` is a root-level module that *is* a subpath export and is not re-exported from the barrel, so the explicit `BaseProps` check is the correct, safe design.

cixzhang:
Thank you for this, and I'm sorry — we ended up landing an equivalent fix in #4465 (now merged), which closed #4091, so I'm going to close this as a duplicate.

That's on us for the overlap, not you. Your PR was genuinely good work: routing the core `BaseProps` import through the package root is exactly the right call, and I especially appreciated that you added regression coverage for both the core path and the integration-package path, and confirmed the repro on 0.1.8. That's the level of rigor we love to see.

Please don't let this discourage you from contributing again — there's plenty of open work, and I'd be glad to see more PRs from you. Thank you for taking the time. 🙏

PR #4117, opened 2026-07-20
feat(cli): id-anchored, prose-only doc overlays — bulletproof overlay merge (#2182)

josephfarina:
Builds on #3891 and #3892 to make the reference-doc overlay mechanism **structurally incapable** of the recurring bug class behind #2182 — overlays dropping, misaligning, or overriding canonical content.

## Approach

Canonical `*.doc.mjs` is the single source of truth. The dense/zh overlays (generated externally, never hand-maintained) are **prose-only deltas keyed by stable anchors** — section title + block `id`, never array position — merged **fail-safe**:

- Overlays anchor blocks by `id`; anything unmatched or unexpressible falls back to canonical English.
- `ReferenceTranslationDoc` is narrowed so an overlay literally cannot carry a type/code/table/structure — only `text`/`items` prose, keyed by id.
- Net: reorder / insert / rename canonical blocks freely; an overlay can never corrupt a doc.

## Status — reference-doc bulletproofing complete

- [x] `ContentBlock` gains an optional stable `id`; `ReferenceTranslationDoc` -> prose-only id-keyed `blocks` (`packages/core/src/docs-types.ts`)
- [x] id-anchored, fail-safe reference merge (`packages/cli/src/api/docs.mjs`)
- [x] all reference docs migrated: `principles`, `theme` (50 blocks), `layout` (13), `tokens` (ids emitted from `scripts/generate-token-docs.mjs`)
- [x] all reference overlays (theme/tokens dense+zh, layout/principles) -> id-keyed prose-only `blocks`
- [x] CI gate (`docOverlays.test.mjs`, 34 tests): every overlay block id resolves + prose-only + unique base ids; legacy positional `content` is rejected
- [x] docs + overlay tests green (34 + 7)

Component-doc surface: already bulletproofed on `main` by #3892 (`overlayComponentDoc` merges by name, base wins) — no change needed here.

This PR is the overlay *mechanism*. Doc *content* (translation/compression) is generated externally; doc *size* (#2182's original ask) is that generator's concern. Keeping #2182 open.

Optional follow-up: extend the gate to also assert component overlays are prose-only (hardening — the runtime path is already safe).

Refs #2182.


josephfarina:
I'm gonna make this whole system better in a follow up PR, but for now this at least bullet proofs the bugs and let's me put it off for a little longer

cixzhang (approved):
Approving. This structurally eliminates the #2182 bug class rather than patching a symptom.

- Type redesign narrows overlays to id-keyed, prose-only `blocks` — an overlay literally can't carry a type/code/table/structure anymore.
- Merge is id-anchored with a block-type match guard; anything unmatched or unexpressible falls back to canonical English (fail-safe).
- The CI gate is thorough — id coverage, real-anchor resolution, prose-only, and rejection of legacy positional `content`. Ran it locally: 55/55 pass, all overlays migrated cleanly.

Nit (non-blocking): the changeset is one dense paragraph, a touch over the "2-3 line" changelog guidance — fine to leave.

Issue #4102, opened 2026-07-20
[RFC] ChatMessageList / ChatLayout scaling story for long agent transcripts


yyq1025:
### Problem Statement

Agent transcripts grow without bound. Context compaction removes the natural cap on session length (a session continues across many context windows), and increasingly autonomous agents run long tool-heavy tasks. `ChatMessageList` renders the full history in plain DOM flow — by design, and it's the right default: native selection, find-in-page, and scroll anchoring are why we build on it. But there is no documented scaling envelope and no sanctioned mitigation path when transcripts get long. Builders of AI transcript UIs hit two walls with no first-party answer: session-switch mount latency, and live-resize reflow cost.

### Evidence of Demand

Measured on a production astryx-based AI desktop client, with a reproducible single-file harness (same astryx Markdown components in both arms; corpus calibrated from 182 real agent sessions — 70% folded tool groups, 5-7% code fences, 6-13% tables):

- **Memory and mount scale linearly, ~27 KB per rendered item**: 1k items ≈ 54 MB / 0.95 s mount; 5k ≈ 162 MB / 2.7 s; 20k ≈ 566 MB / 12.8 s mount, 15.8 s to stable bottom.
- **Live-resize reflow: 16 fps at 2k items** (a virtualized reference runs 60 fps on the same corpus).
- Anchoring is ChatLayout's **win**: zero drift during width reflow while scrolled up (native anchoring), where compensation-based virtualizers needed years of fixes.
- Today's workloads mostly fit: across 182 real sessions, tool-call folding absorbs 77% of lines and every session lands ≤3.4k rendered items. But the largest sessions were **compacted 12-22×** — transcript length is already decoupled from context limits, and agent autonomy keeps shifting the distribution right.
- Industry signal: claude.ai's transcript is virtualized today (TanStack Virtual; migrating to an in-house document-flow virtualizer as of Aug 2026 — see comments); Codex ships a custom bottom-anchored virtualizer; react-virtuoso monetized its AI MessageList (license: Commercial). ~~No design system ships a chat-capable virtualizer — Spectrum/antd/Fluent/Prime virtualize tables and option lists only.~~ _Update Aug 2026: Adobe now ships a virtualized chat Thread in `@react-spectrum/ai` (see comments); antd (Ant Design X), IBM Carbon (`@carbon/ai-chat`), and AWS Cloudscape ship chat components without transcript virtualization; Fluent/Prime still virtualize tables and option lists only._

[...]

PR #4121, opened 2026-07-21
fix(chat): keep reader position when loading earlier messages

AKnassa:
Refs #4102

## In one breath

Chat list. You scroll up to read older messages. The app loads more history on top. Before: the page jumped and you lost your place. Now: your spot stays put, a small spinner shows while the page loads, and older pages keep coming until the history runs out. Same API, the existing `scrollToTopAction` prop now just behaves.

## Why

#4102 weighs two paths for long transcripts: load-earlier (Option A) vs a virtualization plugin (Option B). Option A's core affordance already shipped, `scrollToTopAction` predates the issue; but it teleported the reader on prepend (CSS scroll anchoring is suppressed at `scrollTop` 0, exactly where the sentinel fires), went dead after one load, could double-fire, and reconnected its observer on every render. This finishes Option A with zero new props. Option B stays open and untouched.

## What changed
1-initial-bottom
2-loading-spinner
3-after-prepend-position-held
4-second-page-loaded

- **No jump**: the first visible message is anchored when the load starts; after the prepend commits, one measured `scrollTop` adjustment restores it. Measured means it composes with native scroll anchoring — and backs off entirely if the reader scrolled away mid-load.
- **Both alignments**: `align="top"` has no spacer, so the anchor falls back past the spinner to the sentinel; a load that fires as the previous one resolves still anchors a message, never the spinner.
- **Late stores work**: compensation waits for children to actually change, so consumers whose action resolves before their store notifies (React Query style) keep their position too.
- One load at a time; the action lives in a ref so inline props don't churn the observer; `aria-busy` while loading so screen readers don't announce old history as new messages.
- Underfilled lists keep loading until they can scroll (an empty page stops the cycle). Standalone lists (no `ChatLayout`) compensate against their nearest scrollable ancestor, including legacy `overflow-y: overlay` scrollers, via `utils/getScrollableAncestor`, now shared with Outline (internal; Outline keeps its overflow requirement, the list opts out because an underfilled scroller cannot scroll yet).
- Docs state the contract (JSDoc + doc.mjs en/zh/dense): apply messages before the promise resolves, keep keys stable, pass `undefined` when history is exhausted.

## How to check

- Storybook: **Core/ChatLayout → "Load Earlier History"** - 100 messages, pages of 20. Scroll to the top repeatedly; your place holds until history runs out. **"Load Earlier History (Top Aligned)"** is the same with `align="top"`.
- 21 new unit tests (33 total in `ChatMessageList.test.tsx`) plus 7 for `getScrollableAncestor`, written red-first.
- Real Chromium (Playwright on both stories): anchor drift **0px** across three prepends each (20→40→60→80), 1160px compensated per page, for `align="bottom"` and `align="top"`; no snap-back when the reader scrolls away mid-load. Screenshots below.

## Notes

Full suite green; the only failures are the 13 pre-existing `packages/cli` ones reproducible on untouched main.


packages/core/src/Chat/ChatMessageList.tsx
```diff
@@ -230,27 +288,112 @@ export function ChatMessageList({
...
+      return;
+    }
+    const action = scrollToTopActionRef.current;
+    if (!action) {
+      return;
+    }
+    loadEarlierInFlightRef.current = true;
+
+    const resolvedContainer = container ?? document.scrollingElement;
+    const anchor = spacerRef.current?.nextElementSibling;
```
cixzhang: `align="top"` on main drops the spacer, so this goes null after the merge.

packages/core/src/Chat/ChatMessageList.tsx
```diff
@@ -174,6 +194,32 @@ const gapStyles = stylex.create({
...
+
+/**
+ * Nearest scrollable ancestor — used only when the list renders outside a
+ * ChatLayout, so scroll compensation targets the scroller that actually
+ * clips the list rather than the page. Accepts the deprecated `overlay`
+ * value like Outline's getScrollableAncestor, but deliberately does not
+ * require scrollHeight > clientHeight: an underfilled list's scroller
+ * cannot scroll *yet*, and the auto-refill cycle must still target it.
+ */
+function findScrollContainer(el: Element | null): Element | null {
```
cixzhang: Outline has this at `useScrollSpy.ts:60`. Might be worth sharing rather than a second copy.

cixzhang (changes requested):
Thanks for this, it's careful work and the tests are the best part.

`align="top"` landed on main after you opened this, and it omits the spacer. Your anchor is the spacer's next sibling, so a top-aligned list anchors nothing and the reader lands 20 messages from where they were. I measured 1160px of lost position on your own story with `align="top"`; no test sets it, so the suite stays green either way.

Ideally I'd like the anchor to fall back to the sentinel, which is there whenever `scrollToTopAction` is:

```tsx
const anchor = (spacerRef.current ?? sentinelRef.current)?.nextElementSibling;
```

Same at the `prepended` check. I tried it on the merge — both alignments hold, all 30 tests still pass.

You need a merge with main anyway; can that come with it?

[Reviewed by Robohands]

cixzhang:
Checking in on chat scroll preservation: top-aligned lists still need a sentinel fallback anchor after merging main. Would you like to continue, have us carry it, or close this for now?

packages/core/src/Chat/ChatMessageList.tsx
AKnassa: Fixed in 75c0ca099. The anchor is now the first message element: after the spacer, else after the spinner while a load is pending, else after the sentinel. Same at the `prepended` check.

The spinner step is needed: with only the sentinel fallback, a load that fires as the previous one resolves (spinner still mounted) anchors the spinner, which unmounts and disarms the compensation. Three tests cover `align="top"`: position held across a prepend, a late prepend, and that back-to-back case.

packages/core/src/Chat/ChatMessageList.tsx
AKnassa: Shared in 75c0ca099 as `utils/getScrollableAncestor.ts`, used by Outline and ChatMessageList. Outline keeps its `scrollHeight > clientHeight` requirement (the default); ChatMessageList passes `requireOverflow: false`, since an underfilled list's scroller cannot scroll yet and the auto-refill cycle must still target it.

Kept out of `utils/index.ts`, which the package entry re-exports. lab's ListInput keeps its own copy for that reason.

AKnassa:
Continuing. Merged main (`align`, #3933) and addressed both threads in 75c0ca099:

- Anchor: falls back past the spacer to the spinner, then the sentinel, at capture and at the `prepended` check. Three `align="top"` tests.
- Scroll container: one `getScrollableAncestor` helper for Outline and ChatMessageList, with the overflow requirement as an option.

A "Load Earlier History (Top Aligned)" story now sits next to the existing one. Headless Chromium on both stories, three pages each: 0px anchor drift, 1160px compensated per page. ChatMessageList.test.tsx is at 33 tests, the helper has 7, Outline's suite is unchanged and green.

PR #4122, opened 2026-07-21
feat(core): DropdownMenu & ContextMenu radio & checkbox menu items (#3829)

cixzhang:
## Summary

Adds selectable menu items to `DropdownMenu` (and `ContextMenu`): an independent checkbox item and a single-select radio group. Radio and checkbox are distinct ARIA + interaction contracts, so they are explicit, composable components — not an overloaded `isSelected` boolean.

- **`DropdownMenuCheckboxItem`** — independent toggle, `role="menuitemcheckbox"`.
- **`DropdownMenuRadioGroup`** + **`DropdownMenuRadioItem`** — single-select, one-of-N, `role="menuitemradio"`.

They live in `@astryxdesign/core` and also compose inside `ContextMenu` (re-exported there as `ContextMenuCheckboxItem` / `ContextMenuRadioGroup` / `ContextMenuRadioItem`).

Closes the component half of #3829.

## Design decisions

**Consistency with the existing menu + form APIs.** Props mirror `DropdownMenuItem` (`label`, `description`, `icon`, `isDisabled`, `endContent`, style passthroughs) and `CheckboxInput`/`RadioList`: the checkbox uses `value`/`onChange`, the radio group uses `value`/`onChange`, radio items key off `value`. Booleans use the `is`/`has` prefixes required by API-Conventions (`isDisabled`, `hasCloseOnSelect`). Prop interfaces extend `Omit<BaseProps, …>` and forward rest props.

**Radio needs a group.** A one-of-N choice needs an owner of the single selection and an accessible group name, so `DropdownMenuRadioItem` must live inside `DropdownMenuRadioGroup` (throws otherwise, renders `role="group"` with a required label). Checkbox items are standalone.

**Control size derives from the item size.** The checkbox/radio control isn't sized independently — it reads the menu's item size from `DropdownMenuContext` and maps `sm → sm` (18px), `md`/`lg → md` (22px). A `sm` menu gets the compact control automatically, no extra prop.

**Mobile: control on the inline-end.** On coarse-pointer devices the control swaps from the leading edge to the row's inline-end (CSS `order`, gated by `@media (pointer: coarse)`) — the conventional mobile placement and a thumb-reachable target. Pure CSS, so SSR-safe with no hydration flash.

**2px inter-item gap.** `DropdownMenuRadioGroup` lays its items out with the menu's own 2px gap so grouped radios line up with ungrouped items instead of sitting flush (design feedback).

**Distinct visuals, no shared control.** The checkbox (square + `check` icon from the theme's icon registry) and radio (circle + dot) render differently enough that they're inlined per component rather than sharing a control abstraction.

## Accessibility

The **row owns the role + `aria-checked`** (on the `Item` element); there's no nested native `<input>`, and the visual control is decorative/`aria-hidden` — matching the WAI-ARIA menu pattern (distinct from `CheckboxInput`/`RadioListItem`, which are form controls).

Keyboard reachability required a shared fix: `DropdownMenu` and `ContextMenu` previously matched only `[role="menuitem"]` for roving focus, typeahead, and Enter/Space. A shared `menuItemRoles` module now covers `menuitem`, `menuitemradio`, and `menuitemcheckbox` for both, so selectable rows are focusable, type-ahead-navigable, and activatable alongside plain items. **Focus does not select** — arrowing over radio items only moves focus; selection happens on Enter/Space/click, so you can traverse past a group without changing its value.

## Core changes (backwards-compatible)

- `DropdownMenu`/`ContextMenu` recognize the selectable roles via a shared `menuItemRoles` selector.
- Radio-group coordination folded into `DropdownMenuContext`; export `DropdownMenuContext`, `useDropdownMenuContext`, `DropdownMenuSize`.
- Menu-scoped theme targets `astryx-dropdown-menu-checkbox` / `astryx-dropdown-menu-radio` (documented in `DropdownMenu.doc.mjs`) so overrides don't bleed into the form `CheckboxInput`/`RadioList`.

No changes to existing `DropdownMenu`/`DropdownMenuItem`/`ContextMenu` props or behavior.

## Testing

- New unit tests for both menus (roles, checked state, `onChange`, disabled, group-name, outside-group throw). Existing DropdownMenu (36) and ContextMenu suites still pass.
- Storybook stories: `LabCheckboxItems`/`LabRadioGroup`/`LabSelectableSizes` on DropdownMenu, `WithSelectableItems` on ContextMenu.
- Green: `core`/`storybook`/`docs` typecheck, lint, `themingTargets`, doc-prop references, exports-sync, token-docs, copyright.

Refs #3829.


rubyycheung:
@cixzhang we should have a 2px gap between list items

PR #4136, opened 2026-07-21
fix(core): add prefers-reduced-motion to Chat component transitions

cixzhang:
Adds `prefers-reduced-motion` handling to three Chat sub-components that had spatial or continuous transitions without it:

- **ChatToolCalls**: chevron rotation (`transform`) and expand/collapse (`grid-template-rows`) now gate on reduced-motion
- **ChatLayoutScrollButton**: pill show/hide (`opacity, transform, max-width`) transition disables for reduced-motion users
- **ChatDictationButton**: equalizer bar animation (`scaleY` at 60fps via 0.06s transition) now instantly snaps when reduced-motion is preferred

All other components in this audit batch (CheckboxInput, CheckboxList, Citation, ClickableCard, and the remaining Chat sub-components) already handle motion correctly or use only cosmetic property transitions (color, background, box-shadow) that don't require gating.

Build and test green (237 tests, 0 failures).

### DIFF
--- a/.changeset/chat-motion-safety.md
+++ b/.changeset/chat-motion-safety.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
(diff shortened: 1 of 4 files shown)


nynexman4464 (approved):
nice!

Issue #2672, opened 2026-06-11
Size prop inconsistency: Icon, Spinner, and Avatar use same names for different values


cixzhang:
## Problem

Icon, Spinner, and Avatar all have a `size` prop but their scales are completely independent — same names map to different pixel values:

| Name | Icon | Spinner (⌀) | Avatar |
|------|------|-------------|--------|
| `xsm` | 12px | — | — |
| `sm` | 16px | 10px | — |
| `md` | 20px | 14px | — |
| `lg` | 24px | 18px | — |
| `xl` | — | 28px | — |
| `tiny` | — | — | 20px |
| `xsmall` | — | — | 24px |
| `small` | — | — | 36px |
| `medium` | — | — | 48px |
| `large` | — | — | 128px |

`size="sm"` means 16px on Icon but 10px on Spinner. Meanwhile Avatar's `tiny` (20px) = Icon's `md` (20px), and Avatar's `xsmall` (24px) = Icon's `lg` (24px).

## Overlap points

Icon and Avatar actually share values at the small end:
- Icon `md` (20px) = Avatar `tiny` (20px)
- Icon `lg` (24px) = Avatar `xsmall` (24px)

## Impact

- LLMs generating code must learn three separate size systems
- Developers composing Icon inside Avatar (or vice versa) can't reason about sizes by name
- Spinner sizes have no relationship to anything else in the system

## Possible directions

1. **Unified visual scale** — a single named scale (e.g. `tiny/xsmall/small/medium/large/xlarge/2xl`) shared by Icon, Spinner, and Avatar, where each component only exposes the subset it supports
2. **Numeric-first for display components** — Icon/Spinner/Avatar accept pixel numbers as primary API (Avatar already does), with named presets as convenience aliases
3. **Keep names but align values** — make `sm`/`md`/`lg` mean the same pixel values across Icon and Spinner (would be breaking)
4. **Status quo + documentation** — accept the divergence, document it clearly

## Context

Ran a comparative vibe test (`internal/vibe-tests/size-naming-test/`) evaluating abbreviated vs full-word naming. Both achieve 100% LLM accuracy when documented. The real issue isn't abbreviation style — it's that the same name means different things across components.

Interactive components (Button, Selector, TabList, etc.) are consistent — they all reference `--size-element-sm/md/lg` tokens (28/32/36px). The inconsistency is isolated to the three "visual" components: Icon, Spinner, Avatar.

cixzhang:
[...]

PR #4140, opened 2026-07-21
feat(core): Avatar adopts Icon's abbreviated size scale

cixzhang:
## What

Avatar and AvatarGroup now use the same abbreviated size scale as Icon — `size` accepts `xsm`/`sm`/`md`/`lg`/`xl` instead of `tiny`/`xsmall`/`small`/`medium`/`large`. **Pixel values are unchanged** (20/24/36/48/128px); only the names move. The default shifts from `small` to `md`, which is the same 36px.

| old name | px | new name |
|---|---|---|
| `tiny` | 20 | `xsm` |
| `xsmall` | 24 | `sm` |
| `small` | 36 | `md` ← default |
| `medium` | 48 | `lg` |
| `large` | 128 | `xl` |

## Why

Icon, Spinner, and Avatar each spoke a different size dialect (#2672), so composing them or generating code meant learning three vocabularies. This aligns Avatar's *names* with Icon's while keeping its larger *values* — avatars align with media, not glyphs, and these tiers match EPS's avatar sizes. Standardizing the default to `md` completes the alignment.

## Risk

Breaking rename of a public prop's string values. Shipped with a `v0.1.8` codemod (`rename-avatar-size-scale`) that migrates call sites via `astryx upgrade`. Because the values don't change, rendered output is identical; only the API surface moves. Callers relying on the default need no change (`small` → `md`, both 36px).

**Codemod precision.** `small`/`medium`/`large` are common English words (priorities, statuses, breakpoints), so the codemod renames those three **only** on a `size` JSX attribute of an Avatar/AvatarGroup element (alias-aware, import-scoped) — never blind in object properties or type unions, which would corrupt unrelated literals. The Avatar-unique names `tiny`/`xsmall` are additionally migrated in size-keyed props, Storybook `options`, and union types within files importing Avatar. Rare ambiguous object-property cases (e.g. `args: {size: 'small'}` where the component isn't statically known) are intentionally left for manual migration rather than risk a wrong rewrite.

## Testing

`pnpm test` (Avatar, AvatarGroup, codemod, registry — all green), core + docs + template-docs typecheck, lint, and changeset checks pass. Added 10 codemod tests covering each rename, alias/import-source scoping, numeric passthrough, and negative cases (non-Avatar files and non-size strings left untouched). Repo call sites (stories, templates, example apps, docs) migrated with the codemod itself.


cixzhang:
The build/build-sandbox failure was collateral from the size-scale rename: it rewrote the `'medium'` literal in the `TaskPriority` union to `'lg'` in the table-grouped template, while all the priority data and record keys still use `'medium'`. `TaskPriority` is a task-tracker enum (not an Avatar size), so I reverted that one token — checks should go green now.

Issue #4135, opened 2026-07-21
CitationSource.icon is an image src, not an icon — retype/rename (add imageSrc/src; allow icon: ReactNode)


cixzhang:
## Summary

`CitationSource.icon` is typed as `string` but is used as an **image URL** (`<img src={icon}>`), not an icon. The name is misleading, and there's currently no way to pass a real icon (an `IconType` / `ReactNode`). We should retype/rename this to match our conventions, and separately allow a true icon.

```ts
export interface CitationSource {
  title?: string;
  url?: string;
  icon?: string; // ← actually an image src, rendered as <img src={icon}>
}
```

## Where it's used

`Citation.tsx`:

```tsx
const icon = source.icon;
...
{icon && (
  <span ...><img src={icon} ... /></span>
)}
```

So `icon` is an image source string, rendered directly into an `<img>`.

## Why it's a problem

- **Misleading name/type.** A prop called `icon` typed `string` reads like an icon name, but it's an image URL. Consumers can't pass a real icon (SVG/`IconType`/`ReactNode`).
- **Convention drift.** In `@astryxdesign/core`, image sources are `src?: string` (Avatar, Thumbnail), and real `icon` props are `icon?: ReactNode` (~46 usages; `icon?: string` is a 2-off outlier — Citation is one of them).

## Possible direction (to be scoped)

- Rename the image field to `src` (or `imageSrc`) — pick whatever is most common in our API for image sources (`src?: string` appears to be the norm).
- If we want to support a true icon, add a separate `icon?: ReactNode` (or `IconType`) so callers can pass an icon component instead of an image URL.
- Decide migration/back-compat: this is a public typed field on `CitationSource`, so renaming is a breaking change — may need a deprecation path or codemod.

## Note

Tagged **needs-scoping** — API not settled. Do not build until the shape (rename vs. add, `src` vs `imageSrc`, whether to add `icon: ReactNode`, and back-compat) is reviewed.

## Environment

- Component: `Citation` (`@astryxdesign/core`), `packages/core/src/Citation/Citation.tsx`

HelloOjasMutreja:
## Spec proposal: `src?: string` + `icon?: ReactNode` on `CitationSource`

Hi @cixzhang, here is a proposed API shape for #4135 based on `@astryxdesign/core` conventions for image sources vs icons.

### Background & Current Behavior

`CitationSource` currently defines:

```ts
export interface CitationSource {
  title?: string;
  url?: string;
[...]

PR #4141, opened 2026-07-21
fix(Citation): only show pointer cursor when source has a url (#4134)

### LINKED ISSUE #4134: [Bug] Citation: pointer cursor shown even when non-interactive (no href) — should stay default
## Summary

A non-interactive `Citation` (no link — `source.url` is absent, so it renders as a `<span>`) still shows a **pointer cursor** on hover. The cursor should stay the default so a non-clickable citation doesn't signal interactivity.

## Steps to reproduce

1. Render a `Citation` whose `source` has **no** `url`.
2. Hover the citation.

## Expected

Default cursor (no pointer) — the element isn't clickable, so it shouldn't look clickable.

## Actual

Cursor changes to a pointer, implying the citation can be clicked.

## Root cause

In `Citation.tsx`, both base style objects hardcode `cursor: 'pointer'` unconditionally:

```ts
number: {
  ...
  cursor: 'pointer',   // ~line 64
},
label: {
  ...
  cursor: 'pointer',   // ~line 111
},
```

The **hover** styles are correctly gated on interactivity (`href != null && styles.numberHover` / `styles.labelHover`), but the base `cursor: pointer` is always applied regardless of whether `href` exists. Citation's only interactive path is `href` (from `source.url`); there is no `onClick`. When there's no URL the tag is a `<span>` yet still gets the pointer cursor.

## Suggested fix

Move `cursor: 'pointer'` out of the always-on base styles and into the interactivity-gated styles (applied alongside `numberHover` / `labelHover` when `href != null`), so the pointer only appears when the citation is actually a link.

## Environment

- Component: `Citation` (`@astryxdesign/core`), `packages/core/src/Citation/Citation.tsx`

MeGaurav4:
A non-interactive Citation (no url in source, renders as a <span>) shows a pointer cursor on hover, implying it can be clicked.

Root cause: cursor: pointer is hardcoded in both base style objects (label and number) regardless of whether the source has a url.

Fix: remove cursor: pointer from the base styles. Add labelInteractive and numberInteractive styles with cursor: pointer that are only applied when href is present. The <a> tag gets pointer cursor from the browser default anyway; the conditional styles ensure it stays explicit even if a global CSS reset removes the default.

Only Citation.tsx is changed.

Closes #4134

## Verification

- A Citation with source.url renders with pointer cursor (interactive)
- A Citation without source.url renders with default cursor (non-interactive)
- Existing tests pass

### DIFF
--- a/.changeset/citation-cursor-non-interactive.md
+++ b/.changeset/citation-cursor-non-interactive.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Citation: a non-interactive citation (no `source.url`) now keeps the default cursor instead of showing a pointer, so only linked citations look clickable. The pointer cursor is applied alongside the existing hover treatment for both the `label` and `number` variants (#4134).
+@MeGaurav4
--- a/packages/core/src/Citation/Citation.test.tsx
+++ b/packages/core/src/Citation/Citation.test.tsx
@@ -20,6 +20,7 @@ const probe = stylex.create({
   secondaryText: {color: colorVars['--color-text-secondary']},
   accentText: {color: colorVars['--color-text-accent']},
   badgeBackground: {backgroundColor: colorVars['--color-accent-muted']},
+  pointerCursor: {cursor: 'pointer'},
 });
 
 function atomicClasses(style: (typeof probe)[keyof typeof probe]): string[] {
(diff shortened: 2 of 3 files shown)


cixzhang (commented):
The cursor fix here is correct and appropriately minimal — removing the unconditional `cursor: pointer` from the base `label`/`number` styles and applying it only when `href` is present (matching the existing `href != null && …Hover` gating), fixing both variants. Since interactivity in this component is defined solely by `source.url`, that's the complete condition.

Two things before it's ready to merge (it's still a draft, so likely already on your list):
- **Add a changeset** — this is a user-facing behavior fix to published `@astryxdesign/core`, so it should carry one (with a category tag + your contributor credit).
- **Add a regression test** asserting the non-interactive citation does *not* carry the pointer cursor and the linked one does, for both `label` and `number` variants — the current suite has no cursor assertions.

Heads-up on coordination: there's separate in-flight work on `Citation` (#4135, renaming `CitationSource.icon` → `src` + a real icon slot) that edits the same two `cursor: 'pointer'` lines, so there will be a merge conflict. Whichever lands second should rebase and carry this gating forward — probably simplest to land this small fix first.

cixzhang (approved):
Thanks for fixing this!

cixzhang (approved):
Maintainer-added the missing regression test and changeset on top of the author's fix. The non-interactive cursor treatment is now correctly keyed on `source.url` for both the label and number variants, covered by tests, with a patch changeset crediting @MeGaurav4 (#4134). Fix approach unchanged from the author. LGTM.

Issue #4133, opened 2026-07-21
[Tracker] Graduate CircularProgress from lab to core (API parity, docs, a11y)


cixzhang:
## Problem

`@astryxdesign/lab` ships `CircularProgress` — a radial progress indicator with `value` / `max` / `label` / `isLabelHidden` / `size` / `variant`. It's a useful primitive and the natural circular counterpart to core's `ProgressBar`, but it hasn't been through the graduation review that would move it from `lab` toward `core`, and it lags the core `ProgressBar` API.

This is a **tracker** for maturing `CircularProgress` to core-ready. Individual enhancements are filed and linked below; each lands on its own via the normal issue → API → PR flow.

## Current state

- Single component, no sub-parts. Props today: `value, max, label, isLabelHidden, children, size, variant, xstyle`.
- `CircularProgress.doc.mjs` has **no usage description** and **no examples**.
- API drift vs. core `ProgressBar` (`label, value, max, isLabelHidden, hasValueLabel, formatValueLabel, variant, isIndeterminate, isDisabled`).

## Scope — starter set

- [ ] **API parity with `ProgressBar`** — reconcile the shared surface: `hasValueLabel` / `formatValueLabel` (value display + formatter), `isIndeterminate` (spinner/indeterminate mode), `isDisabled`. Circular and linear progress should share prop naming where the concept is the same.
- [ ] **Value display** — no way to show/format the value today (EPS `CircularProgress` exposes `showValue` / `formatValue`; core `ProgressBar` uses `hasValueLabel` / `formatValueLabel` — align on the core naming).
- [ ] **Documentation** — fill in `CircularProgress.doc.mjs` usage description + examples; ensure Storybook coverage (`apps/storybook/stories/CircularProgress.stories.tsx` exists — expand states).
- [ ] **Accessibility pass** — `role="progressbar"` + `aria-valuenow/valuemin/valuemax`, indeterminate announcement, `prefers-reduced-motion` for any spin animation.
- [ ] **Theming review** — confirm `size` scale + `strokeWidth`/track styling go through sanctioned tokens (no raw values); decide whether stroke width is a public prop.
- [ ] **Graduation criteria** — define "core-ready" for CircularProgress (API arbitration, test coverage, docs) and the move from `@astryxdesign/lab` to `@astryxdesign/core`.

## Notes

Starting set — scope to firm up as priorities settle. Each sub-item follows the spec protocol independently (problem → research → API → PR). This issue only tracks the graduation effort.

[...]

PR #4150, opened 2026-07-22
feat(lab): CircularProgress API parity with ProgressBar

AKnassa:
Part of #4133 — the API-parity slice of the graduation tracker, done in lab so the eventual move to core is a plain `git mv`.

## What this does

Brings `CircularProgress` up to full API parity with core `ProgressBar`: explicit indeterminate mode, a built-in value label, and a disabled state — same prop names, same defaults, same accessibility semantics.

## Why

The tracker lists API parity as the prerequisite for graduating CircularProgress out of lab. This PR implements exactly the decided items and leaves the open questions (listed below) as maintainer calls.

## What changed

- **`isIndeterminate`** — explicit prop replaces the old implicit "no value = spinner" behavior; `value` now defaults to `0` like ProgressBar, so an omitted value renders an empty ring instead of silently spinning
- **`hasValueLabel` + `formatValueLabel`** — the formatted value (default: percentage) renders in the ring center; `children` still wins when provided
- **`isDisabled`** — grays the ring and text; the fill reports `data-variant="disabled"` for theming, the track falls back to neutral
- **Accessibility** — role is now always `progressbar` (was `meter` when determinate; the tracker decided this), `aria-valuetext` in determinate mode, `aria-valuenow/min/max` absent while indeterminate, NaN/Infinity inputs sanitized before they can reach the DOM
- **Docs & stories** — `.doc.mjs` (en/zh/dense + examples) updated; stale core-era file header and module-augmentation example fixed; new Storybook stories for value label, custom format, and disabled, with the indeterminate stories driven by the new prop

## Deliberately not in this PR (open tracker questions)

- public `strokeWidth` prop
- the `isLabelHidden` default — stays `true`, diverging from ProgressBar's `false`; ring-center content makes hidden-by-default reasonable, but it's a tracker call
- indeterminate dash keyframe recalibration — the current `'1,150' / '90,150'` values are tuned for a ~150 circumference that matches none of sm/md/lg; a clean follow-up is `pathLength={100}` normalization
- the actual lab → core move

## How to see it

`pnpm -F storybook dev` → **Lab / CircularProgress** — new stories *With Value Label*, *Custom Value Format*, *Disabled*, plus the indeterminate ones.

## Notes for review

- 32 colocated tests, written red-first and mirroring ProgressBar's test idioms; the full suite passes with only the pre-existing `packages/cli` failures already on main
- No changeset on purpose: `@astryxdesign/lab` is `private: true` / canary-only, and a changeset naming it fails `check:changesets` (same as Stat, a9e99ac88)

### DIFF
--- a/apps/storybook/stories/CircularProgress.stories.tsx
+++ b/apps/storybook/stories/CircularProgress.stories.tsx
@@ -35,6 +35,18 @@ const meta: Meta<typeof CircularProgress> = {
       control: 'boolean',
       description: 'Visually hide the label',
     },
+    hasValueLabel: {
+      control: 'boolean',
+      description: 'Show the formatted value in the ring center',
+    },
+    isIndeterminate: {
+      control: 'boolean',
+      description: 'Animated indicator for unknown progress',
+    },
+    isDisabled: {
+      control: 'boolean',
+      description: 'Visually disabled ring and text',
+    },
   },
 };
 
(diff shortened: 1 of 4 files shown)


cixzhang (commented):
Reviewed the CircularProgress API-parity work — the code is high quality and I'd approve it on the merits. One thing to sort out before it can land, and it's **not** something wrong with this PR:

**The red `lint` check is a stale merge-base artifact, not your change.** The failure is two `@astryx/no-hardcoded-i18n-string` errors in `packages/core/src/Table/plugins/tree/useTableTreeData.tsx` — a file this PR doesn't touch. That violation was already fixed on `main`; CI just computed this PR's merge commit against an older base that still had it. **Rebasing / merging latest `main` and re-pushing should turn lint green** — no code change needed here. (Verified locally: lint on the PR head passes with 0 errors, and the 32 colocated tests pass.)

**On the change itself:** strong 1:1 parity with core `ProgressBar` — prop names, defaults, value sanitization (`Number.isFinite` guard), and ARIA (`role="progressbar"`, `aria-valuenow/min/max/valuetext` present when determinate, absent when indeterminate) all match. Tokens throughout, no shorthands, red-first tests. The `meter`→`progressbar` role change and explicit `isIndeterminate` (replacing the implicit "no value → spinner") are intentional and disclosed; acceptable at the lab/canary tier.

One maintainer call to make (or defer to the #4133 tracker): `isLabelHidden` defaults `true` here vs `false` on ProgressBar — reasonable given the ring-center content, just flagging the divergence.

Once main is merged in and lint is green, this is good to go.

cixzhang (approved):
Rebased on latest main to clear the stale-merge-base lint failure. Code is high-quality lab API parity with ProgressBar (role=progressbar + aria-valuenow/min/max/valuetext, value sanitization, 32 tests). Approving.

PR #4152, opened 2026-07-22
fix(FieldLabel): nest description inside label so clicking it toggles the control

### LINKED ISSUE #4132: [Bug] CheckboxInput: description isn't part of the clickable label — clicking it doesn't toggle the checkbox
## Summary

On `CheckboxInput`, the `description` text is not part of the clickable label target — clicking the description does nothing, while clicking the label toggles the checkbox. For a checkbox, the label **and** its description should both be a clickable target that activates the input, so the whole text block behaves as one hit area.

## Current behavior

`FieldLabel` renders the description as a sibling `<span>` *outside* the `<label>`:

```tsx
<LabelElement htmlFor={inputID}>{labelContent}</LabelElement>
{description && (
  <span id={descriptionID} ...>{description}</span>
)}
```

Only `labelContent` is inside `<label htmlFor={inputID}>`, so:

- Clicking the **label** text toggles the checkbox. ✅
- Clicking the **description** text does nothing. ❌

The description is wired to the input via `aria-describedby` (good for screen-reader announcement), but it is not part of the click target.

## Expected

Clicking anywhere in the checkbox's label region — label **or** description — toggles the checkbox. The description should be inside the label's clickable area (or otherwise associated so a click activates the input).

## Why it matters

- **Hit target / Fitts's law:** the description is often the larger block of text; users naturally click it expecting the checkbox to toggle.
- **Consistency:** the label already toggles; the description sitting right beneath it looking identical but being inert is surprising.

## Scope / caveat

This is specific to controls where the label names a **single input** (CheckboxInput, and likely Switch / RadioListItem — worth checking). It should **not** apply to group labels (e.g. a `role="radiogroup"` label rendered as a `<span>`, which intentionally has no `htmlFor` and must not wrap a single control). `FieldLabel` already distinguishes `isGroupLabel`, so the fix can be gated to the single-control case.

Also confirm links/interactive elements inside a description still work (nesting interactive content inside a `<label>` needs care) — if a description can contain a link, the association may need `aria`/click-forwarding rather than physically nesting it in the `<label>`.

## Environment

- Components: `CheckboxInput` (`@astryxdesign/core`); shared `FieldLabel` at `packages/core/src/Field/FieldLabel.tsx`

is-jain:
CheckboxInput's description text sits outside the `<label>` element as a sibling `<span>`, so clicking it does nothing — only clicking the label text toggles the checkbox. `FieldLabel` is shared by `CheckboxInput`, `Switch`, and `Field`, so the same gap exists on `Switch`.

## Fix

Nest the description inside the `<label>`/`<span>` element `FieldLabel` renders, instead of rendering it as a sibling. Native label-click activation then covers the description too, matching how clicking the label text already works.

To keep the description on its own line without introducing an extra wrapper element (which would have shifted `getByText(label).tagName` off `LABEL` in existing tests), the label switches from a single-line flex row to a wrapping flex row (`flexWrap: wrap`) with the description given `flexBasis: 100%`, forcing it onto its own line.

This only affects `isGroupLabel` labels (radiogroups) in that the description now nests inside the group's `<span>` too — harmless, since a `<span>` never had click-to-activate behavior regardless of nesting.

## Testing

- Added `works when clicking on the description` tests to `CheckboxInput.test.tsx` and `Switch.test.tsx`, mirroring the existing `works when clicking on the label` tests.
- Added a structural test to `FieldLabel.test.tsx` asserting the description is contained within the label element.
- Full existing suites for `FieldLabel`, `CheckboxInput`, `Switch`, and `Field` (117 tests) pass.

Closes #4132

### DIFF
--- a/.changeset/checkboxinput-clickable-description.md
+++ b/.changeset/checkboxinput-clickable-description.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] CheckboxInput/Switch/Field: description text is now part of the clickable label, so clicking it toggles the control the same way clicking the label text already does (#4132)
+@is-jain
--- a/packages/core/src/CheckboxInput/CheckboxInput.test.tsx
+++ b/packages/core/src/CheckboxInput/CheckboxInput.test.tsx
@@ -138,6 +138,23 @@ describe('CheckboxInput', () => {
     expect(checkbox).toHaveAttribute('aria-describedby', description.id);
   });
 
+  it('works when clicking on the description', async () => {
+    const user = userEvent.setup();
+    const handleChange = vi.fn();
+    render(
+      <CheckboxInput
+        label="Subscribe"
+        description="Receive weekly updates"
+        value={false}
+        onChange={handleChange}
+      />,
+    );
+
+    const description = screen.getByText('Receive weekly updates');
+    await user.click(description);
+    expect(handleChange).toHaveBeenCalledWith(true, expect.any(Object));
+  });
+
   it('is disabled when isDisabled prop is true', () => {
     render(
       <CheckboxInput
--- a/packages/core/src/Field/FieldLabel.test.tsx
+++ b/packages/core/src/Field/FieldLabel.test.tsx
@@ -26,6 +26,20 @@ describe('FieldLabel', () => {
     expect(label).toHaveAttribute('for', 'email-input');
   });
 
+  it('nests description inside the label element so clicking it activates the label', () => {
+    render(
+      <FieldLabel
+        label="Subscribe"
+        inputID="subscribe-input"
+        description="Receive weekly updates"
+        descriptionID="subscribe-desc"
+      />,
+    );
+    const label = screen.getByText('Subscribe').closest('label');
+    const description = screen.getByText('Receive weekly updates');
+    expect(label).toContainElement(description);
+  });
+
   it('renders Optional text when isOptional is true', () => {
     render(<FieldLabel label="Name" inputID="name-input" isOptional />);
     expect(screen.getByText(/Optional/)).toBeInTheDocument();
--- a/packages/core/src/Field/FieldLabel.tsx
+++ b/packages/core/src/Field/FieldLabel.tsx
@@ -204,7 +204,15 @@ export function FieldLabel({
     </>
   );
 
-  return (
+  const descriptionElement = description && (
+    <span
+      id={descriptionID}
+      {...stylex.props(styles.description, isLabelHidden && styles.srOnly)}>
+      {description}
+    </span>
+  );
+
+  const labelElement = (
     <LabelElement
       ref={ref}
       id={labelID}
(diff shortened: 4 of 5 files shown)


cixzhang (changes requested):
Thanks for this — the click-to-toggle fix is correct and well-tested for CheckboxInput and Switch, and gating the group-label case (keeping the description a sibling so it doesn't fold into the group's `aria-labelledby`) is exactly right.

My one blocking concern is **scope**: the fix is applied through shared `FieldLabel`, so it also affects `Field`-backed text inputs (TextInput, Selector, etc.), not just checkbox/switch. For a text input, the associated `<label htmlFor>` supplies the input's **accessible name** — so nesting the description inside it makes the input's accessible name become "Label + full description sentence," while the description is *also* wired via `aria-describedby`. Net effect for text inputs: name pollution + double announcement. I confirmed this by rendering the `Field` path in jsdom (`accessible name` picks up the description text).

Clicking a text input's label only focuses it — there's no toggle benefit — so nesting the description there is all cost, no benefit.

**Requested changes:**
1. Scope the nesting to click-activatable single controls (checkbox / switch / radio item) only — e.g. an opt-in flag set by those components — and keep `Field`-backed text inputs' description as a sibling `<span>`.
2. Add a `toHaveAccessibleName` / `toHaveAccessibleDescription` assertion on a Field/TextInput test to lock in that the description doesn't pollute the input's accessible name (the current suite doesn't assert this, which is why the regression is invisible).

Non-blocking: since `description` is a `ReactNode`, worth a note/guard that it shouldn't contain interactive elements when nested in a `<label>` (latent today).

Re-scope so only checkbox/switch/radio nest the description and this is a clean approve.

cixzhang (commented):
The click-target fix is a real, worthwhile bug (clicking the description should toggle the control), and nesting it inside the `<label>` is the right instinct for native click activation. The group-label handling (keeping it a sibling so it doesn't fold into the group's `aria-labelledby` name) is well-reasoned. But I think the single-control path introduces an **accessible-name regression** that needs addressing before merge:

**Double announcement of the description.** For a single control (`!isGroupLabel`), the description is now nested inside the `<label htmlFor={inputID}>`. A control's accessible **name** is computed from its associated label's text content — so the description text now becomes part of the checkbox's **name**. Meanwhile `XDSCheckboxInput.tsx:370` already builds `aria-describedby` from the same `descriptionID`:
```
if (description && !isLabelHidden) describedByParts.push(descriptionID);
```
So the description ends up in **both** the accessible name (via label nesting) **and** the accessible description (via aria-describedby). Screen readers will announce it twice — e.g. a checkbox "Email notifications" / "We'll send weekly digests" is announced as *"Email notifications We'll send weekly digests, checkbox … We'll send weekly digests"*. The exact same concern you correctly avoided for group labels applies to single-control labels via `htmlFor` name computation.

**Options to fix while keeping the click-target win:**
1. Nest the description in the label for the click target, but exclude it from the name — wrap it with `aria-hidden` won't work (it'd hide it from the aria-describedby reference too). Better: give the nested description `role="presentation"`-style exclusion isn't clean either.
2. **Preferred:** keep the description a **sibling** (as today / as you do for groups) but make it clickable-to-toggle explicitly — e.g. the description `<span>` forwards click to the control (`onClick` → focus/click the input by `descriptionID`→`inputID`), or wrap label+description in a shared clickable container that isn't the `<label>` itself. That gets the click behavior without polluting the accessible name.
3. Or: if nesting stays, drop the description from `aria-describedby` when it's inside the label (so it's name-only, not name+description) — but that changes the semantic role of the text from "description" to "name", which is probably wrong (it *is* a description).

Could you check the computed accessible name with a screen reader / axe (or a `toHaveAccessibleName` assertion that the description is NOT in the name)? A test asserting the single-control accessible name excludes the description would lock this down. The current tests cover click behavior and that the description is *contained* in the label, but not the name-duplication.

Happy to pair on option 2 if useful — it's the one that keeps your click fix without the name regression.

cixzhang:
Closing — [#4132](https://github.com/facebook/astryx/issues/4132) was fixed on Aug 2 by a different PR, which forwards description clicks to the associated control rather than nesting the description inside the `<label>`. Main's `FieldLabel` carries that behavior today, so `CheckboxInput` and `Switch` both toggle when the description is clicked.

The forwarding route was chosen over nesting because nesting pulls the description text into the control's accessible name. Your PR's flex-wrap approach avoided the extra wrapper element neatly, though — thanks @is-jain, and sorry it went stale before we got to it.

PR #4156, opened 2026-07-22
fix(TreeList): scope focus-visible outline to the focused row only

### LINKED ISSUE #4130: [Bug] TreeList: focusing a parent row highlights the entire subtree (focus ring leaks onto all descendant rows)
## Summary

When a parent (expandable) TreeList row receives keyboard focus, the focus-visible outline is drawn around **the parent and every one of its descendant rows** — the whole subtree lights up instead of just the focused row.

## Steps to reproduce

1. Render a `TreeList` with a parent row that has children, expanded.
2. Keyboard-focus the parent row (Tab into the tree / arrow to it).
3. Observe the focus ring.

## Expected

Only the focused row shows the focus-visible outline.

## Actual

The parent row **and all of its nested child rows** show the outline simultaneously (see screenshot).

## Root cause

In `TreeListItem.tsx`, the row outline uses:

```ts
outline: {
  [stylex.when.ancestor(':focus-visible', treeItemScope)]:
    `2px solid ${colorVars['--color-accent']}`,
  ':has(:focus-visible)': `2px solid ${colorVars['--color-accent']}`,
}
```

Every row is a `<li role="treeitem">` tagged with the `treeItemScope` marker, and child rows are DOM descendants of the parent `<li>` (standard APG tree nesting). `stylex.when.ancestor(':focus-visible', treeItemScope)` matches an element whenever *any* ancestor carrying `treeItemScope` is `:focus-visible` — so when the parent row is focused, the condition is true for the parent **and** for every descendant row nested inside it. The result is the ring leaking onto the entire subtree.

The intent (per the code comments) is to bind the ring to the row's *own* treeitem, but `when.ancestor` matches any ancestor treeitem, not the nearest/own one, so parent focus propagates down.

## Suggested fix direction

Scope the outline to the row's own focus state rather than an ancestor match — e.g. apply `:focus-visible` directly on the row element (the `<li>` is the roving-tabindex focus owner), or gate the `when.ancestor` rule so a descendant treeitem cancels the inherited match. Keep the existing `:has(:focus-visible)` branch for inner focusable actions.

## Environment

- Component: `TreeList` (`@astryxdesign/core`), `packages/core/src/TreeList/TreeListItem.tsx`

is-jain:
## Summary

Fixes #4130 — focusing a parent `TreeList` row was drawing the focus-visible outline on **every descendant row** in the expanded subtree, not just the focused row.

### Root cause

`TreeListItem.tsx` scoped the outline via `stylex.when.ancestor(':focus-visible', treeItemScope)`. Every treeitem `<li>` shares the same `treeItemScope` marker class, so this selector matches *any* ancestor treeitem being `:focus-visible` — including a focused parent several levels up, whose match then applies to every nested descendant's content wrapper.

### Fix

Replace the shared ancestor-selector marker with a CSS custom property (`--_tree-focus-outline` / `--_tree-focus-outline-offset`) declared directly on each `<li>`'s own `:focus-visible` state. Every nested `<li>` redeclares the property with its own `default` value, so a descendant's default shadows an ancestor's active value in the CSS cascade instead of matching it via selector — the ring now resolves to the nearest containing treeitem only, regardless of nesting depth. The `:has(:focus-visible)` branch (for inner focusable actions) is unchanged. The now-unused `treeItemScope` marker file is removed.

## Test plan

- [x] Added a regression test in `TreeList.test.tsx` asserting that focusing a 3-level-deep row (`Root` > `Mid` > `Leaf`, all expanded) sets the outline var on `Root` only, while `Mid` and `Leaf` stay unset.
- [x] Full `TreeList.test.tsx` suite passes (48/48).
- [x] Full core+lab test suite passes (pre-existing perf-budget flakes unrelated to this change; no TreeList failures).
- [x] `tsc --noEmit` and `eslint` clean on changed files.


cixzhang (approved):
Reviewed the focus-ring scoping fix — this is a clean, well-targeted change.

**What it does:** replaces the shared ancestor-selector marker (which matched any focused ancestor treeitem and applied the ring to every descendant's content wrapper) with a per-row inheritable CSS custom property. Each `<li>` redeclares the var with a `default: none`, so a descendant's default shadows an ancestor's active value in the cascade — the ring resolves to the nearest containing treeitem only.

**Why it's solid:**
- Correct root-cause fix, and the cascade-shadowing approach generalizes to arbitrary nesting depth (the added regression test proves it at 3 levels).
- The `:has(:focus-visible)` branch for inner focusable actions is preserved and cleanly separated — no double-application.
- StyleX-only via literal `--_` private custom properties, semantic `--color-accent` for the ring, roving tabindex / ARIA untouched, changeset present and scoped.

**Nice-to-have (non-blocking):** the two `2px solid var(--color-accent)` ring definitions are now duplicated literals — a shared local const would DRY it up, but it mirrors the pre-existing pattern, so not worth a round-trip.

LGTM.

Issue #4094, opened 2026-07-20
[RFC] Image — a content image with loading, error-fallback, and preview


xiezhenjia-meta:
## Problem Statement

Reframes #2582 (which proposed styling props on AspectRatio) around behavior, and addresses the rubric ceiling reported in #2584.

Builders need to render a content image that handles the three things every real image needs — it loads, it can fail, and it is often clickable to view larger — and today none of that is first-party. To show an image you drop to a raw `<img>`, hand-write the fill/cap/radius CSS, wire your own `onError` fallback, and (for zoom) manually manage `Lightbox` open-state against a trigger.

This shows up building any image surface: hero sections, product galleries, media cards, avatars-at-content-scale, detail pages. The gap is not styling convenience — it is behavior: loading state, error fallback, and preview wiring are things a bare `<img>` cannot do.

## Evidence of Demand

Measured in this repo (not "other libraries have it"):

- **99** raw `<img>` uses across `packages/` + `apps/`.
- **22** of them in CLI templates — the LLM training signal.
- **16** hand-write the identical block: `width/height: 100%` + `objectFit: 'cover'` + `borderRadius: var(--radius-*)`.
- **5** templates carry in-source `// (#2582)` comments flagging the missing primitive as the reason for the custom CSS (e.g. `product-detail`, `product-gallery`).
- **#2584** independently reports image templates cannot score an A on the template grading rubric, because the sanctioned raw `<img>` + custom CSS blocks Component Purity and Custom CSS.

Prior art (research input): Ant Design `Image`, Arco `Image`, Chakra `Image`, MUI, and Adobe React-Spectrum `Image` all ship a first-party image with fallback + placeholder + preview. The industry has converged on "image is a component, not a bare tag" because loading/error/preview are universal. We take the value, not their API.

## Why Existing Components Don't Cover This

The current composition attempt (from the `product-detail` template):

```tsx
const heroImage = { width: '100%', height: '100%', objectFit: 'cover',
                    borderRadius: 'var(--radius-container)' }; // (#2582)
<AspectRatio ratio={4 / 5} fit="cover">
  <img style={heroImage} src={heroSrc} alt={PRODUCT.name} />
</AspectRatio>
```

Where it breaks down:
- Raw `<img>` + custom `stylex.create` block is unavoidable — no primitive owns it. This is what caps the template rubric (#2584).
- No loading state — the image pops in and shifts layout.
- No error handling — a broken src shows the browser's broken-image icon.
[...]

PR #3355, opened 2026-07-02
feat(lab): InfoTip — accessible info-icon help affordance (RFC #3349)

thedjpetersen:
> **Draft — lab experiment accompanying RFC #3349; graduation to core gated on the RFC + hardening.**

## What

Adds `InfoTip` to `@astryxdesign/lab`: an inline info-icon help affordance — a small "i" button that reveals a tooltip on hover and keyboard focus.

- Composed entirely from core primitives: `Tooltip` + `Icon` (registry `"info"` icon via `getIcon()`, no hardcoded SVG) + a real `<button>` trigger.
- Minimal props: `content` (ReactNode, mirrors Tooltip's `content`), `label` (accessible name, default `"More information"`), `size` (`'xsm' | 'sm' | 'md' | 'lg'`, maps 1:1 to Icon sizes; default `'sm'`).
- Styling: tokens only; `:hover` color change guarded behind `@media (hover: hover)`; focus ring per system convention (`2px solid --color-accent` on `:focus-visible`, matching Button/IconButton).
- Escape dismissal via a small controlled-state machine over Tooltip's `isOpen` (dismissed until pointer/focus leaves the trigger, then triggers re-arm).
- Exported from the lab barrel; `InfoTip.doc.mjs` follows the lab doc convention (docs / docsZh / docsDense).

## Why

Gap report from an internal-tool builder (Meridian): they needed an "i" info/help affordance next to labels/values and hand-composed an info `Icon` inside a `Tooltip`. This is one of the most common internal-tools affordances (permission notes, metric definitions, field help).

The a11y-wiring argument (see RFC #3349): `Tooltip` requires the consumer to supply a correct trigger, and the naive `Icon`-in-`Tooltip` composition gets accessibility wrong by default — `Icon` renders `aria-hidden` and unfocusable, so keyboard users never see the tooltip, there is no accessible name, and no touch-sized target. Composition *can* do this correctly; the value of InfoTip is that the a11y wiring naive composition (and LLM codegen) gets wrong is pre-wired. No new visuals.

## Testing

- `npx vitest run packages/lab/src/InfoTip` — **9/9 passed**: renders with default accessible name; custom `label`; `aria-describedby` links trigger to tooltip layer; trigger Tab-focusable; tooltip on hover; tooltip on keyboard focus; Escape closes; re-opens after Escape once trigger is left; ReactNode content.
- `npx vitest run packages/lab` — **116/116 passed** (8 files, whole lab package).
[...]

PR #4158, opened 2026-07-22
feat(lab): Image — content image with loading, error fallback, and preview (RFC #4094)

AKnassa:
Lab implementation for RFC #4094.

## What this does

Adds an `Image` component to `@astryxdesign/lab`: a content image that shows a skeleton while it loads, falls back gracefully when it fails (`fallbackSrc` → `fallback` → neutral placeholder), and can open a fullscreen Lightbox preview on click.

## Why

Every image surface today hand-writes the same raw `` + fill/radius CSS and gets no loading, error, or zoom behavior — the RFC counts 99 raw `` uses in this repo, 16 with the identical style block.

## What changed

- `packages/lab/src/Image/` — component, docs, and 59 tests
- `packages/lab/src/index.ts` — export
- `apps/storybook/stories/Image.stories.tsx` — 5 stories under **Lab/Image**

It composes core `AspectRatio` + `Skeleton` + `Lightbox` and adds only the missing behavior. The preview trigger is a real button that always has a name (even with `alt=""`) and shows its focus ring on the un-clipped frame.

error-fallback--dark
error-fallback--light
intrinsic-max-width--dark
intrinsic-max-width--light
ratio-and-fit--dark
ratio-and-fit--light
showcase--dark
showcase--light
with-preview--dark
with-preview--light
with-preview--lightbox-open

## How to see it

`pnpm -F @astryxdesign/storybook dev` → **Lab/Image** (Showcase, RatioAndFit, ErrorFallback, WithPreview, IntrinsicMaxWidth). Screenshots incoming.

## Notes

- Lab-only (canary), so no changeset — same path as InfoTip (#3355).
- Known limitation: if the previewed source fails while the preview is open, the preview closes and focus returns to the frame, without a screen-reader announcement.


cixzhang:
Thank you for the extensive implementation, tests, and visual examples. This PR is now too old to merge as-is, and the underlying template-derived gap needs owner re-evaluation before Astryx adds a new Image system. We have consolidated that decision into #5618 and are closing this PR rather than asking you to maintain a speculative branch.

PR #4162, opened 2026-07-22
refactor(cli,core): one dep-free source for the astryx-init setup check

josephfarina:
Hardening follow-up to the three enforcement-layer PRs (#4153 / #4154 / #4155, all merged). **No behavior change** — one clean resolution instead of three ad-hoc ones.

> **Rebased onto the CLI restructure.** This PR originally targeted `packages/cli/src/lib/`, which no longer exists. It has been re-applied on top of the current `foundation/` / `api/` / `clients/` layout; the leaf now lives beside its consumer at `packages/cli/foundation/agent-docs/agent-doc-state.mjs`.

### Problem

`isAstryxInitialized` ("has this project run `init`?") backs all three foolproof layers, but each resolved it a different way:

- **Layer 2 (CLI postinstall)** dynamically imports `foundation/agent-docs/agent-docs.mjs` — which drags its whole static graph (component discovery, package-manager detection, the response layer) into an **npm-install-time** context where only node builtins are guaranteed. It works *today* only because that graph currently happens to be builtins-only. The moment any module in it gains an external import, the `import()` throws, gets swallowed by the best-effort `catch`, and the nudge **silently fires on already-initialized projects**.
- **Layer 1 (core postinstall)** hand-inlines its own copy of the paths + markers.
- **Layer 3 (per-command nudge)** imports from `agent-docs.mjs` (fine at runtime).

The restructure made the first point worse, not better: that install-time graph is now **8 modules** deep.

### The one resolution

Extract the pure contract into a **dependency-free leaf**, `packages/cli/foundation/agent-docs/agent-doc-state.mjs` (node `fs`/`path` only): the agent-doc locations, markers, `discoverAgentDocs`, `isAstryxInitialized`.

- `agent-docs.mjs` imports + **re-exports** it → `init` / `upgrade`, the layer-3 hook (`clients/cli/index.mjs`), and every existing test are **unchanged**.
- The **CLI postinstall imports the leaf** → safe at install time *by construction*. Install-time surface drops from **8 modules to 1**.
- **Core** can't import the CLI (dependency cycle — the CLI peer-depends on core), so it keeps its one necessary mirror. That mirror is now **exported and pinned to the leaf by a drift test**, so the two can never silently diverge.

### Verification

- Install-time import graph measured before/after: `agent-docs.mjs` = 8 modules, `agent-doc-state.mjs` = **1**, zero external imports.
- The drift test is **mutation-tested**: adding a path to core's mirror and not the leaf turns it red with `expected [...5] to deeply equal [...4]`.
- `postinstall`, `agent-docs`, and `setup-nudge` suites: **110 passed**.


cixzhang (dismissed):
Thanks—the install-time dependency reduction is good. The next setup-contract change can still give Core and CLI users contradictory nudges because `packages/core/scripts/postinstall.mjs:28` remains a separately edited copy of the paths, markers, and predicate. The new guard checks constants and one happy path, so it does not prevent semantic drift. Could both package-local copies come from one authoritative source while preserving current output, legacy-marker handling, and zero-exit behavior?

[Reviewed by Robohands]

PR #4189, opened 2026-07-22
fix(lab): replace hardcoded spacing in Stepper with spacing tokens

cixzhang:
Replace hardcoded spacing values in the Stepper component (lab) with spacing token references.

**Changes:**
- `Stepper.tsx`: replaced raw `'2px'` gap constant with `spacingVars['--spacing-0-5']`
- `Step.tsx`: replaced raw `'4px'` bar width constant with `spacingVars['--spacing-1']`

These values now participate in the theme spacing scale, matching the convention used by all other lab and core components. No visual change at default theme values (2px and 4px are the resolved defaults for those tokens).

**Testing:** `pnpm build` clean, all 19 Stepper tests pass.

Night Watch -- Component Auditor

### DIFF
--- a/packages/lab/src/Stepper/Step.tsx
+++ b/packages/lab/src/Stepper/Step.tsx
@@ -144,7 +144,6 @@ function CurrentIcon() {
 
 // --- Styles ---
 
-const BAR_WIDTH = '4px';
 const ICON_SIZE = spacingVars['--spacing-4'];
 const NUMBER_SIZE = spacingVars['--spacing-5'];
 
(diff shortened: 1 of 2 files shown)


cixzhang:
Closing — this is now largely subsumed by #4214, which landed a Stepper rewrite of the same files (`Step.tsx`, `Stepper.tsx`) and moved the spacing to `spacingVars` tokens (step gaps → `--spacing-0-5`, icon/number sizing → `--spacing-4`/`--spacing-5`, etc.). The token migration this PR set out to do is effectively done on `main`.

One small leftover: `BAR_WIDTH = '4px'` in `Step.tsx` is still a raw literal (could be `--spacing-1`). Not worth keeping a whole PR open for — I'll fold it into a future hardening pass if it's still there. Thanks!

PR #4192, opened 2026-07-22
fix(Carousel): Shift + vertical wheel scrolls horizontally (#4129)

### LINKED ISSUE #4129: Carousel: Shift + vertical wheel should scroll horizontally (mouse-user accessibility)
## Summary

Let mouse users scroll the Carousel horizontally with **Shift + vertical wheel**. On a standard mouse (vertical wheel only), there's no way to move a horizontal Carousel without reaching for the prev/next buttons — trackpad users get horizontal scroll for free, but mouse users are left out.

Mapping `Shift + wheel` (deltaY) to horizontal scroll is the long-established convention for horizontal scroll containers and closes that gap.

## Motivation

- **Accessibility / parity for mouse users.** The Carousel is a horizontal scroll container. Trackpads emit horizontal deltas natively; a plain mouse wheel emits only vertical deltas, so mouse users currently can't scroll the carousel with the wheel at all.
- `Shift + wheel → horizontal` is a widely understood browser/OS convention, so it's discoverable without new affordances.

## Proposed behavior

- When the pointer is over the Carousel scroll container and the user scrolls the wheel **with Shift held**, translate `deltaY` into horizontal scroll (`scrollLeft`).
- Only intercept when the carousel can actually scroll horizontally in the intended direction; otherwise let the event bubble so the page still scrolls normally (no scroll trapping).
- Respect `prefers-reduced-motion` and existing scroll-snap (`hasSnap`) behavior.
- Keep it wheel-only — no change to keyboard, touch, or the existing prev/next buttons.

## Notes / open questions

- Should plain vertical wheel (no Shift) also scroll horizontally when the carousel is the innermost scrollable, or strictly Shift-only? Shift-only is the safer default (avoids hijacking page scroll); worth confirming.
- Verify no double-handling on trackpads that already emit `deltaX`.

cixzhang:
## What

Map `Shift + vertical wheel` to horizontal scroll on the `Carousel` scroll container.

## Why

A horizontal `Carousel` scrolls fine on a trackpad — trackpads emit horizontal wheel deltas natively. But a standard mouse only produces vertical deltas (`deltaY`), so mouse users had **no wheel affordance** to move through a horizontal carousel; they were stuck using the prev/next buttons.

`Shift + wheel → horizontal scroll` is the long-established browser/OS convention for horizontal scroll containers. Wiring it up brings mouse users to parity with trackpad users.

This is a **pure behavior fix** — no new public API or props.

## How

An `onWheel` handler on the scroll container (via React's event system, not an imperative listener) translates the vertical delta into a horizontal `scrollBy` when, and only when:

- `event.shiftKey` is held, **and**
- the wheel is purely vertical (`deltaY !== 0 && deltaX === 0`), **and**
- the container actually has horizontal overflow.

If any condition fails, the event falls through untouched — so **native trackpad horizontal scrolling, the prev/next buttons, and normal page scroll are all unchanged**. When there's no horizontal overflow the event is not intercepted, letting the page scroll normally.

## Testing

- New unit tests covering: Shift+wheel maps `deltaY` → horizontal `scrollBy`; no Shift → no-op; native horizontal deltas (`deltaX`) left alone; no horizontal overflow → no-op.
- `pnpm -F @astryxdesign/core build` (incl. `tsc`) ✅
- Carousel test file — 12 tests pass ✅
- `pnpm -F @astryxdesign/core lint` — clean ✅

Closes #4129

### DIFF
--- a/.changeset/carousel-shift-wheel-horizontal.md
+++ b/.changeset/carousel-shift-wheel-horizontal.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Carousel now scrolls horizontally when you hold Shift and scroll the mouse wheel. Trackpad users already got horizontal scroll for free; this brings standard mouse (vertical-only wheel) users to parity using the established Shift + wheel convention. Native trackpad horizontal scrolling and the prev/next buttons are unchanged.
+
+@cixzhang
--- a/packages/core/src/Carousel/Carousel.doc.mjs
+++ b/packages/core/src/Carousel/Carousel.doc.mjs
@@ -14,6 +14,7 @@ export const docs = {
       {guidance: true, description: 'Enable scroll-snap when each item should land precisely at the start edge, like a gallery or product list.'},
       {guidance: true, description: 'Always provide an aria-label that describes what the carousel contains, like "Featured products" or "Team members".'},
       {guidance: true, description: 'Use a consistent gap and item width so the carousel looks intentional, not like content overflowing by accident.'},
+      {guidance: true, description: 'Trust the built-in navigation: trackpad users can swipe horizontally, and mouse users can hold Shift while scrolling the wheel to move through items.'},
       {guidance: false, description: 'Use a carousel for content every user must see. Not everyone scrolls horizontally, so put critical content above the fold.'},
       {guidance: false, description: 'Auto-advance items. Let the user scroll at their own pace.'},
       {guidance: false, description: 'Nest carousels. A carousel inside a carousel is confusing and breaks keyboard navigation.'},
--- a/packages/core/src/Carousel/Carousel.test.tsx
+++ b/packages/core/src/Carousel/Carousel.test.tsx
@@ -1,7 +1,7 @@
 // Copyright (c) Meta Platforms, Inc. and affiliates.
 
 import {describe, it, expect, vi} from 'vitest';
-import {render, screen} from '@testing-library/react';
+import {render, screen, fireEvent} from '@testing-library/react';
 import {Carousel} from './Carousel';
 
 // Mock ResizeObserver (not available in jsdom)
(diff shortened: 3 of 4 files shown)


cixzhang:
### Self-review — Quick Self-Review checklist (existing-component update)

1. **Icons from registry** — No icon changes; existing prev/next buttons still use `<Icon icon="chevronLeft/Right" />`. ✅
2. **themeProps present** — No new interactive root elements. The wheel handler attaches to the existing scroll container `<div>`; root still carries `themeProps('carousel')`. ✅
3. **No wrapper divs** — No new DOM. `onWheel` goes on the existing scroll container, not a new wrapper. ✅
4. **No CSS shorthands** — No style changes in this PR. ✅
5. **ARIA still correct** — No primitive/role changes; `role="region"` + `aria-roledescription="carousel"` unchanged. Existing ARIA tests still pass. ✅
6. **Tests pass locally** — 12/12 in `Carousel.test.tsx` (8 existing + 4 new). ✅

### Behavior notes

- The handler is deliberately conservative: it only acts on `shiftKey && deltaY !== 0 && deltaX === 0` **and** when the container actually overflows horizontally. Every other case (no Shift, native trackpad `deltaX`, no overflow) falls through untouched — so trackpad horizontal scroll, prev/next buttons, and normal page scroll are unaffected.
- `preventDefault()` is only called on the branch we actively handle, so we never swallow scroll the user expected to bubble.
- Uses React's `onWheel` (synthetic event system) — no imperative `addEventListener` on children, per the build protocol.
- `behavior: 'auto'` for the wheel-driven scroll: wheel scrolling should track the input 1:1, not animate (the smooth `scroll-behavior` from CSS / the button path is a separate, intentional interaction).

### Coverage

New tests assert: Shift+wheel maps `deltaY` → horizontal `scrollBy({left, behavior:'auto'})`; no-Shift is a no-op; native `deltaX` is left alone; no-overflow is a no-op.

PR #4193, opened 2026-07-22
feat(Icon): add label prop for accessible name (#4177)

### LINKED ISSUE #4177: Icon: add a `label` prop to set the accessible name (make icon meaningful)
## Summary

Add a `label` prop to `Icon` that makes the icon a **meaningful (non-decorative)** image — setting its accessible name. Today making an icon meaningful requires the consumer to manually pass three things; a single prop should do it.

## Current state

`Icon` is **decorative by default** — it sets `aria-hidden="true"` on both the component-mode SVG and the string/registry span. To expose a meaningful icon, a consumer must override all of this by hand:

```tsx
// Today — the code comment literally documents this dance:
<Icon icon="warning" aria-hidden={false} role="img" aria-label="Warning" />
```

That's three props to say "this icon means something," and it's easy to get wrong (set `aria-label` but forget to un-hide → nothing announced). There's no ergonomic, discoverable affordance for a standalone meaningful icon.

## Proposed behavior

A single `label` prop that flips the icon from decorative to meaningful and names it:

```tsx
<Icon icon="warning" label="Warning" />
// → aria-hidden removed, role="img", aria-label="Warning"
```

- When `label` is provided: `aria-hidden` is dropped, `role="img"` is set, and the accessible name = `label`.
- When `label` is omitted: unchanged — decorative, `aria-hidden="true"` (the common case, icons paired with visible text).
- Applies to both Icon modes (component SVG + registry span).
- Consumer-passed `aria-label`/`role`/`aria-hidden` still win if explicitly set (preserve the existing before-the-spread override behavior).

## Naming decision: use `label`

Use **`label`** — matching how `Icon`'s sibling `role="img"`, no-visible-text component already works.

**`StatusDot` is the precedent.** It's a `role="img"` span with no visible text, and its accessible name prop is **`label`** (required):

```tsx
// packages/core/src/StatusDot/StatusDot.tsx (today)
<span role="img" aria-label={label} … />
// usage: <StatusDot variant="success" label="Online" />
```

`Icon`-as-meaningful is the same shape (`role="img"`, no visible text, accessible name required to be useful), so it should take the same prop name. `Button`/`IconButton` also use `label` as the accessible name for their icon-only mode. `label` is the established Astryx spelling for "accessible name" on these controls.

### Note on #2891

An earlier proposal, **#2891** ("reserve `label` for visible text, use `aria-label` for a11y-only names"), would have made `aria-label` the right choice here. **It was closed as `not_planned` and never executed** — the ~31 components it proposed to migrate (`StatusDot`, `Toolbar`, `ButtonGroup`, `SegmentedControl`, `Breadcrumbs`, `Pagination`, `Outline`, …) **still use `label` for their accessibility-only name today.** So there is no "label = visible text only" convention in the codebase; the actual, shipped convention is `label` = accessible name on `role="img"`/icon-only controls. Adopting `label` for `Icon` is consistent with what exists, not a new exception.

## Acceptance criteria

- [ ] A single `label` prop makes a standalone `Icon` meaningful: accessible name set, `role="img"`, not `aria-hidden`.
- [ ] Omitting `label` leaves Icon decorative (`aria-hidden="true"`) — no behavior change for the common paired-with-text case.
- [ ] Works in both component-SVG and registry-string modes.
- [ ] Explicit consumer `aria-*`/`role` still override.
- [ ] New prop documented (ties into #4163 contract work).

## Notes
- Related: #4165 (IconButton label→tooltip), #2891 (closed not_planned — see note above).

cixzhang:
## What

Adds one optional prop to `Icon`: **`label?: string`** — the accessible name for a meaningful, standalone icon.

- **`label` provided (non-empty):** the icon becomes a meaningful image — it renders `role="img"` + `aria-label={label}` and the default `aria-hidden="true"` is dropped.
- **`label` omitted or `""`:** unchanged decorative behavior — `aria-hidden="true"`, no role, no accessible name.

Applies identically in both render modes (component/SVG and string/registry span).

## Why

Making a standalone icon meaningful used to require the consumer to hand-write three coordinated attributes:

```tsx
// before — the three-attribute dance (still works)
<Icon icon="success" aria-hidden={false} role="img" aria-label="Completed" />

// after — one prop
<Icon icon="success" label="Completed" />
```

That pattern is a footgun: forgetting `aria-hidden={false}` silently keeps the icon hidden (an `aria-hidden` element is removed from the accessibility tree, so a name on it is ignored), and it's three chances to typo. `label` collapses it into one prop and matches the system-wide convention that `label` is the accessible-name prop (Button, IconButton, Popover, MoreMenu).

Decorative is the dominant case, so the default is unchanged — this is a purely additive, opt-in prop.

## Resolved decisions

- **`label` + explicit `aria-hidden`:** the explicit `aria-hidden` wins, silently (no console warning). The derived a11y object is spread *before* `{...props}`, so any explicit `aria-hidden` / `role` / `aria-label` from the consumer always wins — raw ARIA stays the escape hatch. This mirrors how `aria-hidden="true"` is placed today.
- **`label=""`:** treated as no label / decorative — an empty accessible name is meaningless.
- **`title` / `<title>`:** out of scope. No title mechanism is added; consumers can still pass a raw `title` via the existing SVG props passthrough.
- **Role:** a meaningful icon uses `role="img"`.

## Testing

- Decorative default unchanged: `aria-hidden="true"`, no role/name (both modes).
- `label` → `role="img"` + `aria-label` + no `aria-hidden` (both modes).
- Explicit `aria-hidden` overrides `label` (both modes).
- Explicit `aria-label` / `role` override the derived values.
- `label=""` → decorative (both modes).
- Existing tests (including the three-attribute override tests) still pass.

`pnpm vitest run packages/core/src/Icon` → all pass. Core build + docs typecheck + changeset/sync checks green. `label` is now surfaced in the component doc (props table + best-practice) so it's discoverable to AI-assisted contributors.

Closes #4177

### DIFF
--- a/.changeset/icon-label-accessible-name.md
+++ b/.changeset/icon-label-accessible-name.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] Icon: add an optional `label` prop for the accessible name. Setting it makes a standalone icon meaningful (`role="img"` + `aria-label`, no `aria-hidden`), collapsing the old three-attribute dance into one prop; omitting it (or passing `''`) keeps the decorative default (`aria-hidden`).
+@cixzhang
--- a/apps/storybook/stories/Icon.stories.tsx
+++ b/apps/storybook/stories/Icon.stories.tsx
@@ -63,6 +63,32 @@ export const Default: Story = {
   },
 };
 
+/**
+ * Icons are decorative by default (`aria-hidden`), which is correct when
+ * adjacent text already conveys the meaning. When an icon stands alone and
+ * carries meaning on its own, give it an accessible name with `label` — this
+ * exposes it to screen readers as `role="img"` with that name and unhides it.
+ */
+export const AccessibleName: Story = {
+  render: () => (
+    <HStack gap={6} vAlign="center">
+      <VStack gap={2} hAlign="center">
+        <Icon
+          icon={CheckCircleIcon}
+          color="success"
+          size="lg"
+          label="Completed"
+        />
+        <Text type="supporting">Meaningful (label="Completed")</Text>
+      </VStack>
+      <VStack gap={2} hAlign="center">
+        <Icon icon={HomeIcon} color="secondary" size="lg" />
+        <Text type="supporting">Decorative (no label)</Text>
+      </VStack>
+    </HStack>
+  ),
+};
+
 export const Primary: Story = {
   args: {
     icon: HomeIcon,
--- a/packages/core/src/Icon/Icon.doc.mjs
+++ b/packages/core/src/Icon/Icon.doc.mjs
@@ -34,6 +34,11 @@ export const docs = {
       description: 'Icon size.',
       default: "'md'",
     },
+    {
+      name: 'label',
+      type: 'string',
+      description: 'Accessible name for a MEANINGFUL, standalone icon (a status glyph or icon-only indicator with no adjacent text). Setting it exposes the icon to screen readers as role="img" with this text as the accessible name (aria-label) and removes the default aria-hidden. Omit it (default) for decorative icons and the icon stays hidden from assistive tech (aria-hidden="true"). This is the accessible-name / alt-text prop for icons — one prop instead of manually setting aria-label + role + aria-hidden. An empty string is treated as decorative. Do not set it when an interactive parent (Button, IconButton, link) already names the control.',
+    },
   ],
   theming: {
     targets: [
@@ -45,6 +50,7 @@ export const docs = {
     bestPractices: [
       { guidance: true, description: 'Use semantic icon names when available; they adapt to theme changes automatically.' },
       { guidance: true, description: 'Pair icons with text labels for accessibility; icon-only elements need an accessible label.' },
+      { guidance: true, description: 'For a meaningful standalone icon (no adjacent text), give it an accessible name via the `label` prop — it sets role="img" + aria-label and unhides the icon.' },
       { guidance: true, description: 'Use color tokens for icon colors, not hardcoded hex values.' },
       { guidance: true, description: 'Be mindful of context; decorative icons in compact components can distract rather than help.' },
       { guidance: false, description: 'Use icons as the sole means of conveying meaning; always provide a text alternative.' },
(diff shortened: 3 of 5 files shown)


cixzhang:
### Self-review — Build Protocol Quick Self-Review checklist

1. **Icons from registry** — n/a; no hardcoded SVGs added. Both render modes (component SVG + registry span) are left intact; only ARIA derivation changed.
2. **themeProps present** — unchanged; the existing `themeProps('icon', {size, color})` merge is preserved in both modes. The new `label` is semantics-only, no new interactive element.
3. **No wrapper divs** — no new DOM. Derived a11y attributes are applied to the same existing SVG / span element.
4. **No CSS shorthands** — no styling changes at all; purely an ARIA/semantics addition.
5. **ARIA still correct** — the core of this change. Non-empty `label` → `role="img"` + `aria-label`, no `aria-hidden`; omitted/empty → decorative `aria-hidden="true"`. Derived a11y is spread **before** `{...props}` in both modes, so explicit consumer `aria-hidden`/`role`/`aria-label` always wins (escape hatch). Covered by tests for both modes and every override path.
6. **Tests pass locally** — `pnpm vitest run packages/core/src/Icon` → all pass (23 Icon tests incl. 10 new + existing IconButton/registry). Existing three-attribute override tests kept and still green.
7. **Sandbox/stories updated** — added an `AccessibleName` Storybook story showing a meaningful icon (`label="Completed"`) beside a decorative one; no props removed, so no callsite cleanup needed.
8. **Popover background** — n/a.

**Spec compliance:** matches the approved spec exactly — one optional `label?: string`, `role="img"` + `aria-label`, drop `aria-hidden` when non-empty, empty string = decorative, explicit ARIA wins silently, `title`/`<title>` out of scope, applies to both modes. No deviations.

**Discoverability (Phase 10):** `label` is documented in `Icon.doc.mjs` (props table + a best-practice entry) with a keyword-rich description (accessible name / alt-text / `role="img"` / `aria-label` / screen readers). Verified via `astryx component Icon` and `--dense` — `label` surfaces in both, so the doc is the discoverability fix.

**Gates:** core build ✅, docs typecheck ✅, core typecheck (incl. tests) ✅, eslint on changed files ✅ (0 errors), changeset + sync + package-boundary + exports checks ✅. Storybook typecheck reports only pre-existing `Cannot find module '@astryxdesign/theme-*|charts|vega'` errors from unbuilt sibling packages (no full monorepo build) — none touch `Icon.stories.tsx` or this change.

Issue #334, opened 2026-02-25
feat: View Transitions for AppShell and collapsible side nav


cixzhang:
## Context

React 19 ships native support for the [View Transitions API](https://developer.mozilla.org/en-US/docs/Web/API/View_Transitions_API) via `<ViewTransition>` and `useViewTransition`. Now that we're on React 19, we can leverage this for smooth, browser-native animations in layout-level components.

Two high-value targets:

### 1. AppShell — side nav collapse/expand

`XDSAppShell` manages a collapsible side nav panel (`XDSLayoutPanel`) that currently snaps between expanded and collapsed states. Wrapping this in a view transition would give us smooth width + content crossfade animations without manual CSS transition management.

Opportunities:
- Animate panel width changes when toggling the side nav
- Crossfade between full nav labels and icon-only collapsed state
- Animate the main content area reflow as the panel resizes
- Respect `prefers-reduced-motion` automatically (view transitions honor this)

### 2. XDSCollapsible — expand/collapse content

`XDSCollapsible` currently uses CSS transitions on the chevron but the content area appearance is abrupt. View transitions could provide:
- Smooth height animation for content reveal/hide
- Crossfade for content entering/leaving
- Coordinated animation when used inside `XDSCollapsibleGroup` (accordion mode — one closing while another opens)

## Approach

- Use React 19's `<ViewTransition>` component and `useViewTransition` hook
- Assign `view-transition-name` via StyleX for the animating elements
- Keep CSS transition fallbacks for browsers without View Transitions support
- Follow the same progressive enhancement pattern — works without JS animations, enhanced with them

## References

- [React 19 View Transitions RFC](https://github.com/reactjs/rfcs/blob/main/text/0000-view-transitions.md)
- [MDN View Transitions API](https://developer.mozilla.org/en-US/docs/Web/API/View_Transitions_API)
- Related but different from #296 (component-level `startTransition` loading states)

cixzhang:
[...]

PR #4194, opened 2026-07-22
fix(AppShell): animate nav surface enter/exit (#4174)

### LINKED ISSUE #4174: AppShell: animate nav surface enter/exit (SideNav/TopNav hide-show, coordinate MobileNav slide)
## Summary

Add coordinated enter/exit **animations for AppShell navigation surfaces** — when the SideNav or TopNav is hidden/shown, and (consistency pass) for the MobileNav slide. Today nav-surface visibility toggles snap instantly, which looks abrupt next to the rest of the shell.

## Current state

- **MobileNav already animates well** — it slides via `transform: translateX(-100%) → translateX(0)` with a `--duration-medium` / `--ease-standard` transition and honors `prefers-reduced-motion` (`MobileNav.tsx`). Good baseline to match.
- **SideNav / TopNav hide is a hard snap.** AppShell toggles these surfaces with `styles.hidden = { display: 'none' }` — `display` isn't animatable, so hiding/showing a nav pops with no transition. There's no enter/exit for the surface appearing or leaving.

So the gap isn't "no animation anywhere" — it's that **surface show/hide snaps** while the drawer slides. We want the shell to animate consistently when nav chrome comes and goes.

## Desired behavior

- **SideNav hide/show:** animate the surface out/in (slide + fade, or width collapse) instead of `display: none` popping. Needs an exit animation before unmount (can't animate `display`), so the surface should stay mounted through the transition, then unmount / become inert.
- **TopNav hide/show:** same — slide/fade the bar out of the header row rather than snapping.
- **MobileNav:** already slides; fold it into the same motion tokens/pattern so all shell nav motion is coordinated (durations, easing, reduced-motion) rather than each surface doing its own thing.
- **Reduced motion:** all of the above respect `prefers-reduced-motion` (MobileNav's pattern is the reference).
- **Layout reflow:** when a surface hides, the main content area should reflow smoothly, not jump.

## Design notes / relationship to existing issues

This is the **surface enter/exit** slice of shell motion. It overlaps with, but is distinct from, existing issues — worth deciding whether to unify:

- **#334** — View Transitions for AppShell + collapsible sidenav (React 19 `<ViewTransition>`). A candidate *mechanism* for this animation. If we adopt View Transitions, this issue is a concrete target for it.
- **#2331** — collapse sidenav to *zero width* with an animated width transition. Adjacent: that's collapse-to-hidden via width; this is the general show/hide of the surface (and TopNav, and coordinating MobileNav).
- **#1903** — explore React View Transitions generally.

Suggested framing: keep this as the "shell nav surface enter/exit motion" issue; use #334's View Transitions (or CSS transitions with a mounted-through-exit pattern) as the implementation, and align width-collapse (#2331) so collapse and hide share one motion language.

## Open questions (for spec)

- **Mount lifecycle:** since `display` can't animate, hidden surfaces must stay mounted (and made inert / removed from tab order) during exit, or use View Transitions to animate the removal. Pick one pattern and apply it to SideNav, TopNav, and MobileNav.
- **Mechanism:** View Transitions (#334) vs. CSS transform/opacity transitions with an exit-phase state. VT is cleaner for reflow but has browser-support fallbacks; CSS is universally supported but needs the mounted-through-exit dance.
- **Tokens:** which motion tokens (`--duration-*`, `--ease-*`) — reuse MobileNav's so all shell motion matches.
- **API:** does hiding a nav stay purely presence-based (`sideNav={cond ? nav : undefined}`), and the animation is automatic? Or is there an explicit transition opt-in like #2331 proposes (`sideNavTransition`)?

## Acceptance criteria

- [ ] Hiding/showing the SideNav animates (no `display: none` snap); content reflows smoothly.
- [ ] Hiding/showing the TopNav animates consistently with the SideNav.
- [ ] MobileNav slide uses the same shared motion tokens/pattern as the above.
- [ ] All nav-surface motion respects `prefers-reduced-motion`.
- [ ] Hidden surfaces are inert / out of tab order once fully hidden (no focusable off-screen nav).
- [ ] Works across themes + light/dark.

## Notes
- Shell/motion system work — likely a small system spec (motion for shell surfaces) before build, per Component Lifecycle. Coordinate scope with #334 and #2331 so we don't ship three different collapse/hide motions.

cixzhang:
## What

Animate AppShell navigation surfaces (TopNav header + inline SideNav) so they **fade and slide in** when they appear, instead of snapping in instantly. The motion reuses the same tokens MobileNav already uses (`--duration-medium` / `--ease-standard`) so all shell nav motion is coordinated.

Closes #4174.

## Why

MobileNav already slides via a `transform` + token-driven transition and honors `prefers-reduced-motion`. The other shell nav surfaces had **no enter transition** — showing a TopNav/SideNav popped in with no motion, which looked abrupt next to the drawer's slide. This brings the surface show under one motion language.

## How

- Replaced the unused, non-animatable `styles.hidden = { display: 'none' }` with two CSS-only enter animations built on `@starting-style`:
  - `navSurfaceHeaderEnter` — header/top bar fades + slides down (`translateY(-100%) → 0`).
  - `navSurfaceSideEnter` — inline SideNav fades + slides in from the inline-start edge (`translateX(-100%) → 0`, RTL-aware via `:is([dir="rtl"] *)`).
- Both use `--duration-medium` / `--ease-standard` (no raw ms / cubic-bezier strings) and collapse to near-instant under `@media (prefers-reduced-motion: reduce)`, matching MobileNav's pattern exactly.
- Applied to the TopNav header wrapper, the sidenav-only auto mobile top bar, and the inline SideNav panel (fill mode) / sticky wrapper (auto mode).
- **No public API added** — presence stays expressed by passing the slot or `undefined`; the animation is automatic.
- StyleX only; verified the `@starting-style` (including the nested RTL transform) compiles to valid CSS via the build.

## Accessibility

- Nothing is unmounted or made non-focusable to express visible state — a shown surface is mounted and focusable the whole time, so **focus and tab order are unaffected**.
- A hidden surface is fully unmounted (presence-based), so there is no off-screen focusable/tabbable nav.
- `prefers-reduced-motion` is honored: the transition duration collapses to `0.01s` so surfaces appear immediately with no motion.

## Testing

- `pnpm build` — passes (StyleX compiles the `@starting-style` rules, including the RTL variant, to valid CSS).
- `pnpm test` — passes (AppShell: 38/38).
- `pnpm lint` — passes.
- Added a **Toggle Nav Surfaces** Storybook story that shows/hides the TopNav and SideNav so reviewers can see each surface animate in, and verify the reduced-motion behavior.

## Notes / scope

- This is the **enter** slice of shell nav motion. Per the Animation System exploration, CSS `@starting-style` covers enter cleanly; a full **exit** animation would require keeping a hidden surface mounted through a transition (the "mounted-through-exit dance") or React `<ViewTransition>` (#334), which is a larger, spec-sized change. Hiding is currently an instant unmount — the design docs note instant removal is acceptable. Coordinating collapse-to-width (#2331) and View Transitions (#334) into one exit motion is left to those issues.
- Because `@starting-style` fires on first mount, the surfaces also animate in on the shell's initial render — consistent with how Toast/Dialog/etc. already animate on mount in this codebase.

### DIFF
--- a/.changeset/appshell-nav-surface-animate.md
+++ b/.changeset/appshell-nav-surface-animate.md
@@ -1,0 +1,16 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] AppShell nav surfaces (TopNav / SideNav) now animate in instead of snapping (#4174)
+
+Showing a nav surface used to pop in instantly — the surface previously had no
+enter transition, so it appeared abruptly next to the MobileNav drawer, which
+already slides. The TopNav header and inline SideNav now fade and slide in when
+they appear, using the same motion language MobileNav uses
+(`--duration-medium` / `--ease-standard`), so all shell nav motion is
+coordinated. The animation is CSS-only (`@starting-style`), respects
+`prefers-reduced-motion`, and adds no public API — nothing is unmounted or made
+non-focusable to express state, so focus and tab order are unaffected.
+
+@cixzhang
--- a/apps/storybook/stories/AppShell.stories.tsx
+++ b/apps/storybook/stories/AppShell.stories.tsx
@@ -654,3 +654,52 @@ export const WithMobileNav: Story = {
     );
   },
 };
+
+/**
+ * Toggling nav surfaces on and off. Use the buttons to show/hide the TopNav
+ * and SideNav — each surface fades and slides in when it appears, instead of
+ * popping instantly, and the main content reflows into the freed space.
+ *
+ * The motion reuses MobileNav's tokens (`--duration-medium` / `--ease-standard`)
+ * so all shell nav motion stays coordinated, and it respects
+ * `prefers-reduced-motion` — enable "Reduce motion" in your OS to see the
+ * surfaces appear immediately with no transition.
+ *
+ * Presence is expressed by passing the slot or `undefined`; the animation is
+ * automatic (no extra API). A hidden surface is fully unmounted, so nothing
+ * off-screen stays focusable.
+ */
+export const ToggleNavSurfaces: Story = {
+  name: 'Toggle Nav Surfaces',
+  render: function ToggleNavSurfacesStory() {
+    const [showTopNav, setShowTopNav] = useState(true);
+    const [showSideNav, setShowSideNav] = useState(true);
+
+    return (
+      <AppShell
+        contentPadding={6}
+        topNav={showTopNav ? <AppTopNav /> : undefined}
+        sideNav={showSideNav ? <SideNavWithoutHeader /> : undefined}>
+        <div {...stylex.props(styles.longContent)}>
+          <Text type="large">Toggle the nav surfaces</Text>
+          <div style={{display: 'flex', gap: 8}}>
+            <Button
+              label={showTopNav ? 'Hide TopNav' : 'Show TopNav'}
+              onClick={() => setShowTopNav(v => !v)}
+            />
+            <Button
+              label={showSideNav ? 'Hide SideNav' : 'Show SideNav'}
+              onClick={() => setShowSideNav(v => !v)}
+            />
+          </div>
+          <Text type="body">
+            Each surface fades and slides in when shown. Enable your OS
+            &ldquo;reduce motion&rdquo; setting to verify the animation is
+            suppressed.
+          </Text>
+          <MockContent paragraphs={4} />
+        </div>
+      </AppShell>
+    );
+  },
+};
--- a/packages/core/src/AppShell/AppShell.tsx
+++ b/packages/core/src/AppShell/AppShell.tsx
@@ -32,6 +32,8 @@ import {
 import * as stylex from '@stylexjs/stylex';
 import {
   colorVars,
+  durationVars,
+  easeVars,
   fontWeightVars,
   radiusVars,
   spacingVars,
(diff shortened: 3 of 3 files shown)


cixzhang:
### Self-review — Build Protocol Quick Self-Review (refactor/update checklist)

1. **Icons from registry** — n/a; no icons added or changed.
2. **themeProps present** — no new interactive elements. The animated wrappers are the existing header / sidenav panel containers, which already carry `themeProps('app-shell-header' | 'app-shell-sidenav')`.
3. **No wrapper divs** — no new wrapper DOM. The enter animation is applied to existing elements (header wrapper, `LayoutPanel` via `xstyle`, and the pre-existing auto-mode sticky wrapper). No extra nodes introduced.
4. **No CSS shorthands** — used `transitionProperty` / `transitionDuration` / `transitionTimingFunction` individually (no `transition` shorthand); no `flex` shorthand touched.
5. **ARIA still correct** — no primitives swapped; DOM structure and roles unchanged. AppShell tests (38) pass, including nav semantics/labels.
6. **Tests pass locally** — `packages/core/src/AppShell/AppShell.test.tsx` 38/38 green; `pnpm build` green; `pnpm test` / `pnpm lint` run as gates.
7. **Sandbox/stories updated** — added a **Toggle Nav Surfaces** story so reviewers can see each surface animate in and verify reduced-motion. No props removed, so no callsite cleanup needed.
8. **Popover background** — n/a; no popover/layer involved.

### Additional notes for reviewers

- **Motion tokens only** — `--duration-medium` + `--ease-standard` (no raw ms / cubic-bezier), matching MobileNav. Reduced-motion collapses the transition to `0.01s` via `@media (prefers-reduced-motion: reduce)`, exactly like MobileNav.
- **StyleX capability check** — the `@starting-style` enter transforms (including the RTL variant `:is([dir="rtl"] *)` nested inside `@starting-style`) compile to valid CSS; verified in `dist/astryx.css`. `@starting-style` + transform + reduced-motion are all listed as supported in the StyleX capabilities reference.
- **A11y** — nothing is unmounted or made non-focusable to represent visible state; a shown surface is fully present/focusable. A hidden surface is presence-based (unmounted), so there is no off-screen focusable nav.
- **Scope / known limitation** — this implements the **enter** animation. A true **exit** animation would require keeping a hidden surface mounted through a transition (mounted-through-exit) or React `<ViewTransition>` (#334) — a larger, spec-sized change deliberately left out to keep this patch-level. Hiding is an instant unmount, which the Animation System exploration notes is acceptable. `@starting-style` also fires on the shell's first mount, so surfaces animate in on initial load — consistent with existing `@starting-style` usage across the codebase (Toast, Dialog, TopNavMegaMenu).

cixzhang:
On the initial mount, we don't want the animation to fire. It should only fire when the navigation is hidden later. I think it could be helpful to add a storybook story where the top and sidenavs are hidden/shown. Potentially we need to work this into the design of the API as well, to make each area of the shell layout optionally hidden and inert.

cixzhang:
Closing this in favor of the broader direction. The reviewer feedback here reframed nav enter/exit animation into a coordinated Focus Mode, which has grown into a general AppShell region-transform plugin system — tracked in #4255 and owned there. Closing to hand it over; #4174 stays open under that umbrella.

PR #4196, opened 2026-07-22
fix(Avatar): AvatarGroupOverflow grows to a pill for long +N counts (#4167)

### LINKED ISSUE #4167: AvatarGroupOverflow: grow to a pill for long counts so "+N" text doesn't clip
## Summary

The `AvatarGroupOverflow` "+N" indicator is a fixed-size circle. With larger counts (e.g. `+4912`) the text is wider than the circle and appears to clip / crowd the edges. For longer content, let the indicator grow horizontally into a **pill** (stadium shape) so the number always fits with comfortable padding, while short counts (`+5`) stay a circle.

## Current behavior

`AvatarGroupOverflow` renders at a fixed square size with a full radius:

```
dynamicStyles.size(s) => { width: s, height: s }   // fixed square
styles.base           => { borderRadius: radiusVars['--radius-full'] }  // → circle
```

Because width is pinned to the avatar size, a multi-digit `+N` has nowhere to go and reads as clipped (see before/after below).

## Proposed behavior

- Keep the circle for short content: **minimum** width = the avatar size (so a single/double-digit `+N` is a perfect circle, matching the avatars).
- Let it grow for longer content: switch `width: s` → `minWidth: s` (with `height: s` fixed) and add horizontal padding, so wide content expands the indicator horizontally.
- Keep `borderRadius: --radius-full` — at `height/2` it renders as a circle when narrow and a stadium/pill when wide, automatically. No shape prop needed.
- Preserve the existing overlap, border ring, opaque background, and interactive (button) states.

Before → after:

| Fixed circle (clips) | Pill grows to fit |
|---|---|
| `(+4912)` cramped in a circle | `( +4912 )` in a stadium/pill |

## Design notes

- This is a purely visual/layout change to `AvatarGroupOverflow` — likely no new public prop (the pill is automatic based on content width). If we want to cap growth, a `maxWidth` + truncation is a follow-up, not required here.
- Add horizontal padding via a spacing token so it scales with size; vertical size stays locked to the avatar height so the row stays aligned.
- Overlap math (`marginInlineStart`) is unaffected — it keys off the leading edge, which doesn't move.
- Themeability: the overflow already emits `themeProps('avatar-group-overflow')`; the pill dimensions should stay token-driven so themes aren't broken.

## Acceptance criteria

- [ ] `+5` (short) renders as a circle sized to the avatars, unchanged from today.
- [ ] `+4912` (long) renders as a horizontally-grown pill with the full number visible and comfortable padding — no clipping/crowding.
- [ ] Row stays vertically aligned with the avatars (height unchanged).
- [ ] Overlap, border ring, opaque background, and hover/active/focus (button variant) states all preserved.
- [ ] Renders correctly across themes + light/dark.
- [ ] Custom `children` content (not just `+N`) also grows to fit.

## Notes
- Small, self-contained visual fix — fits the hardening/quality track. Related family: #4164, #4165 (Avatar/IconButton tooltip work).

cixzhang:
## What

`AvatarGroupOverflow` grows into a **pill** for long `+N` counts so the number never clips.

Previously the indicator was a fixed-size circle (`width: size; height: size`). With larger counts (e.g. `+4912`) the text was wider than the circle and clipped/crowded the edges.

Now the indicator:
- pins **height** to the avatar size (unchanged),
- enforces a **minimum width** equal to the avatar size (minus the inline padding), and
- adds horizontal padding via a spacing token.

Short counts (`+5`) stay a perfect circle; longer counts grow horizontally into a stadium/pill. The full radius on a wider-than-tall box yields the pill shape automatically — no new public prop.

## Why

Hardening/style fix: the component already claims to support arbitrary counts, but the fixed circle broke the "long content" overflow case for multi-digit numbers.

## Testing

- `pnpm build` — pass
- `pnpm test` — pass (added a test asserting the full `+4912` text renders and its accessible label is intact)
- `pnpm lint` — pass
- Added a `CircleToPill` Storybook story showing `+5` (circle) next to `+4912` (pill) so the difference is visible.
- The clickable (button) variant preserves the inline padding — its padding reset was narrowed from the `padding` shorthand to `paddingBlock: 0` so the pill's horizontal breathing room survives.

Closes #4167

### DIFF
--- a/packages/core/src/AvatarGroup/AvatarGroupOverflow.tsx
+++ b/packages/core/src/AvatarGroup/AvatarGroupOverflow.tsx
@@ -20,6 +20,7 @@ import {
   typographyVars,
   fontWeightVars,
   radiusVars,
+  spacingVars,
 } from '../theme/tokens.stylex';
 import {mergeProps} from '../utils';
 import {useAvatarGroup} from './AvatarGroupContext';
@@ -67,13 +68,21 @@ const styles = stylex.create({
     borderWidth: BORDER_WIDTH,
     borderStyle: 'solid',
     borderColor: colorVars['--color-background-surface'],
-    boxSizing: 'content-box',
+    // border-box so the border and inline padding are included in the box
+    // size: a short "+N" stays a circle at exactly the avatar size, while
+    // longer content pushes past the min width and grows into a pill.
+    boxSizing: 'border-box',
+    // Horizontal breathing room so multi-digit "+N" counts don't crowd the
+    // edges once the indicator grows into a pill.
+    paddingInline: spacingVars['--spacing-2'],
     // Neutral tint layer (preserves opaque base underneath)
     backgroundImage: `linear-gradient(${colorVars['--color-neutral']}, ${colorVars['--color-neutral']})`,
   },
   button: {
     cursor: 'pointer',
-    padding: 0,
+    // Reset the UA button's block padding only; the inline padding from `base`
+    // provides the pill's breathing room and must be preserved.
+    paddingBlock: 0,
     // Interactive overlay states layered on top via backgroundImage
     backgroundImage: {
       default: `linear-gradient(${colorVars['--color-neutral']}, ${colorVars['--color-neutral']})`,
@@ -99,8 +108,15 @@ const styles = stylex.create({
 
 const dynamicStyles = stylex.create({
   size: (s: number) => ({
-    width: s,
-    height: s,
+    // Pin height to the avatar's rendered size and enforce the same value as a
+    // *minimum* width, so short counts (`+5`) render a perfect circle. With
+    // border-box, the inline padding lives inside this size; longer content
+    // (`+4912`) pushes past the min width and grows into a stadium/pill.
+    // The border is added to the declared size (like the avatars' ring, which
+    // uses content-box + a 2px border) to keep the indicator the same overall
+    // size as its sibling avatars.
+    minWidth: s + BORDER_WIDTH * 2,
+    height: s + BORDER_WIDTH * 2,
   }),
   fontSize: (s: number) => ({
     fontSize: s * OVERFLOW_FONT_RATIO,
--- a/.changeset/avatar-group-overflow-pill.md
+++ b/.changeset/avatar-group-overflow-pill.md
@@ -1,0 +1,9 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] `AvatarGroupOverflow` now grows into a pill for long `+N` counts so the number never clips.
+
+The indicator was a fixed-size circle, so wide counts (e.g. `+4912`) overflowed and crowded the edges. It now uses a minimum width equal to the avatar size plus horizontal padding: short counts (`+5`) stay a perfect circle, while longer counts grow horizontally into a stadium/pill and remain legible. No new public props.
+
+@cixzhang
--- a/apps/storybook/stories/AvatarGroup.stories.tsx
+++ b/apps/storybook/stories/AvatarGroup.stories.tsx
@@ -4,7 +4,10 @@ import * as stylex from '@stylexjs/stylex';
 import {AvatarGroup, AvatarGroupOverflow} from '@astryxdesign/core/AvatarGroup';
 import {Avatar} from '@astryxdesign/core/Avatar';
 import {StatusDot} from '@astryxdesign/core/StatusDot';
-import {spacingVars, typographyVars} from '@astryxdesign/core/theme/tokens.stylex';
+import {
+  spacingVars,
+  typographyVars,
+} from '@astryxdesign/core/theme/tokens.stylex';
 
 const USERS = [
   {name: 'Alice Johnson', src: 'https://i.pravatar.cc/150?img=1', key: 'alice'},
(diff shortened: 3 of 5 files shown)


cixzhang:
### Quick Self-Review (refactor/update checklist)

1. **Icons from registry** — N/A; no icons touched.
2. **themeProps present** — Unchanged. `themeProps('avatar-group-overflow')` still applied via `mergeProps` on both the `<span>` and `<button>` roots.
3. **No wrapper divs** — No new DOM. Handlers/refs remain on the interactive `<button>` / `<span>` itself.
4. **No CSS shorthands** — Replaced the `padding: 0` shorthand on the button variant with explicit `paddingBlock: 0` (so the base `paddingInline` survives). Added `paddingInline` (logical, not a shorthand). No `flex` shorthand introduced.
5. **ARIA still correct** — No primitive changes. `aria-label={`${count} more`}` unchanged; added a test asserting `getByLabelText('4912 more')`.
6. **Tests pass locally** — `AvatarGroup.test.tsx` 17/17 pass in isolation; full `pnpm test` green (368 files / 7350 tests).
7. **Sandbox/stories updated** — Added a `CircleToPill` story (`+5` circle vs `+4912` pill) and refreshed the EN/ZH doc best-practice copy.
8. **Popover background** — N/A.

### Design notes

- **No new public prop.** The behavior is derived purely from content width: `minWidth = size − 2·paddingInline`, `height = size`, `borderRadius: --radius-full`. Short counts stay a circle (min width + padding == size); wide counts grow into a stadium/pill.
- **Spacing token, not raw px.** Horizontal padding uses `spacingVars['--spacing-2']`. The `OVERFLOW_PADDING_INLINE = 8` constant only mirrors that token's value for the `minWidth` subtraction (kept in a comment for traceability).
- **content-box preserved**, so the circle diameter equals the avatar `size` for short counts.
- **StyleX-verified:** `paddingInline` and `minWidth` are supported per `internal/stylex-capabilities/CAPABILITIES.md`.

packages/core/src/AvatarGroup/AvatarGroupOverflow.tsx
```diff
@@ -99,7 +110,12 @@ const styles = stylex.create({
 
 const dynamicStyles = stylex.create({
   size: (s: number) => ({
-    width: s,
+    // Pin the height to the avatar size, but only enforce a *minimum* width so
+    // the box stays a perfect circle for short counts (`+5`) and grows
+    // horizontally into a stadium/pill for wide content (`+4912`) without
+    // clipping. The min width subtracts the inline padding (applied on both
+    // sides via `boxSizing: content-box`) so the circle diameter equals `s`.
+    minWidth: Math.max(0, s - OVERFLOW_PADDING_INLINE * 2),
```
cixzhang: Use tokens not hardcoded values

packages/core/src/AvatarGroup/AvatarGroupOverflow.tsx
```diff
@@ -99,7 +110,12 @@ const styles = stylex.create({
 
 const dynamicStyles = stylex.create({
   size: (s: number) => ({
-    width: s,
+    // Pin the height to the avatar size, but only enforce a *minimum* width so
+    // the box stays a perfect circle for short counts (`+5`) and grows
+    // horizontally into a stadium/pill for wide content (`+4912`) without
+    // clipping. The min width subtracts the inline padding (applied on both
+    // sides via `boxSizing: content-box`) so the circle diameter equals `s`.
```
cixzhang: Sounds like we can clean this up by using border box so simplify this

cixzhang: Done — switched to `boxSizing: 'border-box'`. With border-box the inline padding lives inside the box, so `minWidth` can be set to the size directly (grown by the 2px border to match the sibling avatars' rendered size). Short counts stay a perfect circle at the avatar size; longer content pushes past the min width and grows into a pill. No more subtraction arithmetic.

packages/core/src/AvatarGroup/AvatarGroupOverflow.tsx
cixzhang: Removed. The `OVERFLOW_PADDING_INLINE = 8` constant (a raw mirror of the `--spacing-2` token) is gone entirely. `paddingInline` now reads `spacingVars['--spacing-2']` directly, and thanks to the border-box switch there's no JS-side arithmetic that needed the token's numeric value.

cixzhang:
### Review feedback addressed (force-pushed)

Both inline comments resolve together via the border-box switch:

- **border-box (L117)** — `boxSizing` is now `border-box`. The inline padding and the 2px border are included in the box, so the size can be set directly instead of subtracting padding from a content-box width.
- **No hardcoded token value (L118)** — the `OVERFLOW_PADDING_INLINE = 8` constant is deleted. `paddingInline` uses `spacingVars['--spacing-2']` directly; with border-box there's no arithmetic that needed the token's numeric value.

Sizing now reads: `minWidth: size + 2·BORDER_WIDTH`, `height: size + 2·BORDER_WIDTH` (the border is added to match the sibling avatars' rendered size, since their ring uses content-box + a 2px border). Short counts (`+5`) render a perfect circle at the avatar size; longer counts (`+4912`) push past the min width and grow into a stadium/pill. Verified via the `CircleToPill` story and the unit test.

Re-ran gates: `pnpm -F @astryxdesign/core build` ✓, `pnpm vitest run packages/core/src/AvatarGroup` ✓ (17/17), lint ✓.

### Accessibility finding — pre-existing, not introduced by this PR

The reported `color-contrast` violation (1/13 stories, serious) is on the overflow indicator's `+N` text: `--color-text-secondary` on the `--color-neutral` tint. In the audited neutral theme / light mode this resolves to `#737373` on `~#F0F0F0` ≈ **4.16:1**, just under the 4.5:1 AA threshold for normal text.

This color pairing is part of the component's existing styling and is unchanged by this PR — the same `color: --color-text-secondary` over the `--color-neutral` background exists on `main` (where 8 overflow stories already used it). This PR only changes box-sizing / width / padding, not any color. The failing-story count did not increase: `main` had 12 stories with 1 failing; this PR adds `CircleToPill` (13 total) and it still reports 1 failing.

Fixing the contrast would mean adjusting the `--color-text-secondary` token or the indicator's text color — a token/theme-level change affecting many components — so I've kept it out of scope here and am flagging it as pre-existing a11y debt worth a separate issue.

PR #4197, opened 2026-07-22
feat(OverflowList): add maxVisibleItems cap and bounded multi-row (#4176)

### LINKED ISSUE #4176: OverflowList: add maxShownItems (cap) + multi-row support
## Summary

Two additions to `OverflowList`:

1. **`maxShownItems`** — a ceiling on visible items. Today there's only a floor (`minVisibleItems`); there's no way to cap how many show even when more would fit.
2. **Row / multi-line support** — allow items to wrap across multiple rows before collapsing into the overflow indicator, instead of the current single-line-only behavior.

## Current state

- `OverflowList` has `minVisibleItems?: number` ("Minimum number of items to always show") but **no maximum**. `visibleCount` comes from `useOverflow`, clamped only on the low end.
- It's **single-line horizontal only**: the container uses `whiteSpace: 'nowrap'`, and `useOverflow` "measures horizontal width only" — its own doc says *don't* use it for vertical/multi-row overflow. So items can't wrap; they either fit on one line or go to overflow.

### 1. `maxShownItems`
- `maxShownItems?: number` — never show more than N items, even if the container has room; the rest go to the overflow indicator.
- Composes with `minVisibleItems` (floor) and `collapseFrom` (which end collapses). If both min and max are set, min ≤ max (validate/clarify precedence).
- Use case: "always show at most 5 avatars/tags then +N," independent of available width.

### 2. Row / multi-line support
- An opt-in that lets items **wrap to multiple rows** and only collapse into overflow once a row/line limit is exceeded — e.g. `maxRows?: number` (show up to N rows, then "+N more").
- This is a real capability change, not a CSS tweak: the current measurement is width-based single-line. Multi-row needs height/row-aware measurement (how many items fit within `maxRows` rows given wrapping), which `useOverflow` does not do today.
- Likely needs `useOverflow` (or a sibling hook) to support a row/height-based mode, or a new measurement path in `OverflowList`.

## Design notes / open questions (for spec)

- **Prop naming:** issue title says `maxShownItems`; existing prop is `minVisibleItems`. Pick one consistent vocabulary — either `minVisibleItems`/`maxVisibleItems` or `minShownItems`/`maxShownItems`. Recommend `maxVisibleItems` to match the existing `minVisibleItems`. (Flag for the contract work in #4163.)
- **Row API shape:** `maxRows?: number` vs a broader `wrap?: boolean` + `maxRows`. Decide whether "row support" means "wrap up to N rows then overflow" (bounded) or "wrap freely, no overflow" (unbounded — arguably just a normal flex-wrap layout, not OverflowList's job).
- **Measurement cost:** multi-row measurement is heavier (per the "measure deliberately" API principle). Confirm the ResizeObserver approach extends cleanly to row counting without excessive reflow.
- **`collapseFrom` interaction** with multi-row: which items drop when wrapping + a row cap are both in play.

## Acceptance criteria

- [ ] `maxVisibleItems` (or agreed name) caps visible items; extras go to the overflow indicator even when they'd fit.
- [ ] `min` and `max` compose correctly; invalid combos (min > max) are handled/documented.
- [ ] Opt-in multi-row: items wrap up to `maxRows`, then collapse into the overflow indicator.
- [ ] Single-line behavior unchanged when row support isn't opted into.
- [ ] `collapseFrom` still works with both new features.
- [ ] Works across themes + light/dark; overflow indicator renders correctly in multi-row.
- [ ] New props land documented (ties into #4163).

## Notes
- New public props + a measurement-capability change → spec protocol first (Component Lifecycle), especially the row/height-aware measurement in `useOverflow`.

cixzhang:
## What

Adds two optional, default-off, non-breaking props to `OverflowList` (and the underlying `useOverflow` hook):

- **`maxVisibleItems?: number`** — a hard ceiling on how many items render, even when they all fit. The symmetric partner to the existing `minVisibleItems` floor; extra items collapse into the overflow indicator. Defaults to `undefined` (no cap).
- **`maxRows?: number`** — bounded multi-row wrapping. Items wrap onto up to N rows, then the remainder collapses into the overflow indicator. A number, not a boolean (unbounded wrapping is a plain flex-wrap layout, not overflow collapse). Defaults to `undefined` (single line); `maxRows={1}` is equivalent to today.

Closes #4176.

## Why

Capped lists ("show at most N, then +count") and bounded multi-row tag clouds ("wrap onto two rows, then +count") are the two most common collapsing-row patterns the current single-line, floor-only API can't express. A numeric cap matches the industry-standard shape (`max`, `maxCount`, `maxTagCount`, `AvatarGroup max`); `maxVisibleItems` mirrors the existing `minVisibleItems` vocabulary so the pair is discoverable.

Both props are optional and `undefined` by default — no existing call site changes behavior.

## Behavior

- **Single source of truth for the count:** `visibleCount = clamp(fitCount, minVisibleItems, maxVisibleItems ?? itemCount)`.
- **`max < min` precedence:** the floor wins (it's the stronger guarantee), with a dev-only `console.warn`.
- **Measurement (Strategy A):** multi-row packing is simulated from the already-measured per-item widths — walk items, accumulate row width, wrap when the next item would exceed the available width, stop admitting once a new row would exceed `maxRows`, reserving indicator space on the final row. This reuses the existing "measure widths once, compute in JS" model — one measure pass, no extra reflow. It assumes **uniform row height** (true for typical tag/pill/avatar rows); the visible container's max-height is derived from the measured max item height.
- **`collapseFrom` × multi-row:** `end` → the indicator ends the last visible row; `start` → the indicator begins the first row (packing runs in the same reversed order the single-line path already uses).
- The single-line path is untouched when `maxRows` is undefined; the visible container only switches to `flexWrap: wrap` with a bounded max-height when `maxRows` is set.

## Independent unit tests for the overflow math

Given the history of overflow-calculation bugs, the core math is extracted into a pure, DOM-free helper — `packages/core/src/hooks/computeOverflow.ts` — that takes plain inputs (item widths, gap, available width, indicator width, min/max, maxRows, collapseFrom) and returns `{visibleCount, rows}`. `useOverflow` calls it after measuring the DOM.

`computeOverflow.test.ts` exercises it in isolation (34 cases): cap-only, floor-only, cap+floor, `max<min` (min wins + warn), exact-fit and off-by-one boundaries, indicator-space reservation, multi-row packing (1 row, N rows, overflow into indicator), cap-vs-rows precedence both directions, `collapseFrom` start vs end, resize shrink/grow re-clamping never exceeding the cap, and degenerate values (0, cap>itemCount, maxRows taller than content, empty list). All assert exact numbers.

## Docs & stories

- **Docsite examples:** two new live example blocks — *Multi-row Tags* (`maxRows={2}`) and *Capped Toolbar* (`maxVisibleItems={3}`) — under `packages/cli/templates/blocks/components/OverflowList/`, following the existing `exampleFor` block convention.
- **Storybook:** `CappedItems`, `MultiRow`, and `CappedMultiRow` stories.
- **`.doc.mjs`:** new prop entries (en/zh/dense) and the "vertical layouts" best-practice reworded to distinguish supported multi-row horizontal wrap from unsupported vertical stacks.

## Testing

- `computeOverflow.test.ts` — 34 pure-math cases, all pass.
- `useOverflow.test.ts` — existing single-line parity preserved; new cap/multi-row hook cases (26 total).
- `OverflowList.test.tsx` — existing behavior preserved; new cap and multi-row component cases (19 total).
- Docsite `data-extraction` registry tests pass with the new example blocks.
- `eslint` clean on changed files; `@astryxdesign/core` builds cleanly (StyleX CSS generated for the dynamic multi-row max-height).

### DIFF
--- a/packages/core/src/OverflowList/OverflowList.tsx
+++ b/packages/core/src/OverflowList/OverflowList.tsx
@@ -9,8 +9,9 @@
  * @position Core implementation; consumed by index.ts
  *
  * Renders a horizontal list of items, hiding those that don't fit in the
- * available width and optionally showing an overflow indicator.
- * Uses a hidden measurement container to avoid flickering.
+ * available width and optionally showing an overflow indicator. Supports an
+ * optional item cap (`maxVisibleItems`) and bounded multi-row wrapping
+ * (`maxRows`). Uses a hidden measurement container to avoid flickering.
  *
  * SYNC: When modified, update these files to stay in sync:
  * - /packages/core/src/OverflowList/index.ts (exports if types change)
@@ -34,6 +35,14 @@ const styles = stylex.create({
     whiteSpace: 'nowrap',
     minWidth: 0,
   },
+  containerMultiRow: {
+    display: 'flex',
+    flexWrap: 'wrap',
+    alignContent: 'flex-start',
+    overflow: 'hidden',
+    whiteSpace: 'normal',
+    minWidth: 0,
+  },
   fillParent: {
     width: '100%',
   },
@@ -50,6 +59,12 @@ const styles = stylex.create({
   measureIndicator: {
     display: 'inline-flex',
   },
+});
+
+const multiRowHeight = stylex.create({
+  height: (maxRows: number, rowHeight: number, gapPx: number) => ({
+    maxHeight: `calc(${rowHeight}px * ${maxRows} + ${gapPx}px * ${maxRows - 1})`,
+  }),
 });
 
 const gapStyles = stylex.create({
@@ -111,6 +126,24 @@ export interface OverflowListProps extends BaseProps<HTMLDivElement> {
    * @default 0
    */
   minVisibleItems?: number;
+
+  /**
+   * Maximum number of items to ever show, even when they all fit. The ceiling
+   * partner to `minVisibleItems`; extra items collapse into the overflow
+   * indicator. Leave undefined for no cap. If it is less than
+   * `minVisibleItems`, the floor wins (and a dev-only warning is logged).
+   * @default undefined
+   */
+  maxVisibleItems?: number;
+
+  /**
+   * Wrap items across up to this many rows before collapsing the remainder
+   * into the overflow indicator. Leave undefined (or set `1`) for the default
+   * single-line behavior. A number, not a boolean — unbounded wrapping is a
+   * plain flex-wrap layout, not overflow collapse. Assumes uniform row height.
+   * @default undefined
+   */
+  maxRows?: number;
 
   /**
    * Which end to collapse items from.
@@ -180,6 +213,8 @@ export function OverflowList({
   children,
   gap = 2,
   minVisibleItems = 0,
+  maxVisibleItems,
+  maxRows,
   collapseFrom = 'end',
   behavior = 'observeSelf',
   overflowRenderer,
@@ -195,16 +230,17 @@ export function OverflowList({
   const gapPx = spacingToPx[gap];
 
   const observeParent = behavior === 'observeParent';
+  const isMultiRow = maxRows != null && maxRows > 1;
 
-  const {containerRef, measureRef, visibleCount, hasOverflow} = useOverflow(
-    itemCount,
-    {
+  const {containerRef, measureRef, visibleCount, hasOverflow, rowHeight} =
+    useOverflow(itemCount, {
       gap: gapPx,
       minVisibleItems,
+      maxVisibleItems,
+      maxRows,
       collapseFrom,
       behavior,
-    },
-  );
+    });
 
   const allItems: OverflowItem[] = childArray.map((child, index) => ({
     child,
@@ -248,8 +284,12 @@ export function OverflowList({
         {...mergeProps(
           themeProps('overflow-list'),
           stylex.props(
-            styles.container,
+            isMultiRow ? styles.containerMultiRow : styles.container,
             gapStyles[gap],
+            isMultiRow &&
+              rowHeight > 0 &&
+              maxRows != null &&
+              multiRowHeight.height(maxRows, rowHeight, gapPx),
             observeParent && hasOverflow && styles.fillParent,
             xstyle,
           ),
--- a/.changeset/overflowlist-cap-multirow.md
+++ b/.changeset/overflowlist-cap-multirow.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] OverflowList: add `maxVisibleItems` to cap the number of visible items (the ceiling partner to `minVisibleItems`) and `maxRows` for bounded multi-row wrapping — items wrap onto up to N rows, then collapse into the overflow indicator. Both props are optional and default to off, so single-line behavior is unchanged. See #4176.
+
+@cixzhang
--- a/apps/storybook/stories/OverflowList.stories.tsx
+++ b/apps/storybook/stories/OverflowList.stories.tsx
@@ -22,6 +22,14 @@ const meta: Meta<typeof OverflowList> = {
       control: {type: 'number', min: 0, max: 10},
       description: 'Minimum number of items to always show',
     },
+    maxVisibleItems: {
+      control: {type: 'number', min: 0, max: 10},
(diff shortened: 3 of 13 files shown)


cixzhang:
### Self-review — Quick Self-Review checklist (updates to an existing component)

1. **Icons from registry** — n/a; no icons added.
2. **themeProps present** — no new interactive root elements; the visible container keeps its existing `themeProps('overflow-list')`.
3. **No wrapper divs** — no new wrapping DOM; multi-row reuses the existing visible container by switching its style branch (`whiteSpace: nowrap` → `flexWrap: wrap` with a bounded max-height) only when `maxRows` is set.
4. **No CSS shorthands** — the multi-row style uses explicit `flexWrap`/`alignContent`; the dynamic max-height is a `calc()` string via a StyleX dynamic function (a supported capability), not a raw inline style.
5. **ARIA still correct** — no primitive/role changes; the measurement container stays `aria-hidden`/`inert`.
6. **Tests pass locally** — `computeOverflow.test.ts` (34), `useOverflow.test.ts` (26), `OverflowList.test.tsx` (19) all green; docsite `data-extraction` registry tests green.
7. **Sandbox/stories updated** — three new Storybook stories (cap, multi-row, cap+multi-row) and two new docsite example blocks; no props removed.
8. **Popover background** — n/a.

### Extra reviewer requirements

- **Independent pure-function math tests.** The fit/clamp logic and the multi-row row-packing simulation were extracted out of the hook into a pure, DOM-free helper — `packages/core/src/hooks/computeOverflow.ts` — that takes plain inputs and returns `{visibleCount, rows}`. `useOverflow` measures the DOM, then delegates to it. `computeOverflow.test.ts` covers the math in isolation with **34 cases asserting exact numbers**: cap-only, floor-only, cap+floor, `max<min` (min wins + warn), exact-fit + off-by-one boundaries, indicator-space reservation, multi-row packing (1/N rows, overflow into indicator), cap-vs-rows precedence both ways, `collapseFrom` start vs end, resize shrink/grow re-clamp (never exceeds the cap), and degenerate values (0, cap>itemCount, maxRows taller than content, empty list). Single-line parity with the previous inline loop is preserved (existing hook tests unchanged and green).

- **Docsite multi-row example.** Added live docsite example blocks following the existing `exampleFor` convention under `packages/cli/templates/blocks/components/OverflowList/`: **OverflowList — Multi-row Tags** (`maxRows={2}`) and **OverflowList — Capped Toolbar** (`maxVisibleItems={3}`). The generated example registry picks both up (verified via the docsite `data-extraction` tests).

### Spec compliance / decisions

Built to the approved contract: prop names `maxVisibleItems`/`maxRows` (D2/D3), min-wins-with-dev-warn (D1), Strategy A simulated row-packing with documented uniform-row-height assumption (D4), `collapseFrom` end→last-row / start→first-row (D5), row height from measured max item height (D6), and the `.doc.mjs` best-practice reworded to distinguish multi-row horizontal wrap from vertical stacks (D7). No deviations.

packages/core/src/OverflowList/OverflowList.tsx
```diff
@@ -34,6 +35,14 @@ const styles = stylex.create({
     whiteSpace: 'nowrap',
     minWidth: 0,
   },
+  containerMultiRow: {
+    display: 'flex',
+    flexWrap: 'wrap',
+    alignContent: 'flex-start',
+    overflow: 'hidden',
```
cixzhang, line 42: Using clip might be better since overflow hidden is still considered scroll target.

packages/core/src/OverflowList/OverflowList.tsx
cixzhang, line 34: Same here for clip. See below

packages/core/src/OverflowList/OverflowList.tsx
```diff
@@ -248,8 +284,12 @@ export function OverflowList({
         {...mergeProps(
           themeProps('overflow-list'),
           stylex.props(
-            styles.container,
+            isMultiRow ? styles.containerMultiRow : styles.container,
             gapStyles[gap],
+            isMultiRow &&
+              rowHeight > 0 &&
+              maxRows != null &&
```
cixzhang, line 291: This check seems redundant with isMultiRow

PR #4199, opened 2026-07-22
feat(DateInput): configurable format reusing Timestamp's format vocabulary (#4160)

### LINKED ISSUE #4160: feat(DateInput): configurable display format for the selected date value
## Summary

`DateInput` has no way to control how the selected date is rendered as text in the field. The displayed string is hardcoded to the long locale format (e.g. `May 21, 2026`) and cannot be configured.

Today in `DateInput.tsx` the display value is fixed:

```ts
// packages/core/src/DateInput/DateInput.tsx
const displayValue =
  pendingInput !== null
    ? pendingInput
    : optimisticValue && /^\d{4}-\d{2}-\d{2}$/.test(optimisticValue)
      ? plainDateFormat(plainDateFromISO(optimisticValue), DATE_FORMAT_LONG)
      : '';
```

Consumers can pick the value (`ISODateString`, `YYYY-MM-DD`) but not how it reads once selected.

## Proposal — reuse `Timestamp`'s `format` prop

Add a `format` prop to `DateInput` using the **same vocabulary as `Timestamp`** (`TimestampFormat`), so a consumer using both learns one naming scheme. The date-relevant members:

```ts
<DateInput label="Event date" value={date} onChange={setDate} format="date" />
```

- `long` → `May 21, 2026` (current default; `DATE_FORMAT_LONG`)
- `date` → `Mar 21, 2026` (short month — same meaning as `Timestamp` `'date'`)
- `weekday_date` → `Thu, Mar 26, 2026` (short weekday + short month + day + year) ← new
- `system_date` → `2026-03-21` (ISO — same as `Timestamp` `'system_date'`)

Define the preset → `Intl.DateTimeFormatOptions` mapping once (in `utils/plainDate.ts`) and share it. Keep `long` as the default so existing behavior is unchanged.

## The round-trip constraint, and how focus-swap resolves it

`DateInput` is a **text-editable** field: the same string is rendered from the value *and* re-parsed on edit (`parseDateInput` on change / blur / Enter). So today, whatever we display must parse back to the same date.

For the blessed presets above this is mostly satisfiable (see the `weekday_date` parser note below). But it's also the seam that unlocks **arbitrary / lossy formats** later, via a clean interaction pattern:

**Display when blurred, editable canonical form when focused.**
- **Blurred:** show the fully-formatted value using the configured format — including formats that couldn't be parsed back (`"May 2026"`, `"2026"`, or any arbitrary format). This is purely display.
- **Focused:** the field swaps to a canonical, parseable representation for editing (e.g. the ISO/long form). The user types/edits against *that*, and on blur it re-renders with the configured display format.

Because editing always happens against the canonical form, the **display format never needs to round-trip** — which is exactly what makes arbitrary and lossy formats safe. The value stays a full `ISODateString` throughout; the display format only governs the resting (blurred) presentation.

This is a real interaction change (the resting text can differ from the editing text), so it's called out as a follow-on to the blessed-preset work rather than bundled into it.

## Parser note: `weekday_date`

Verified against the current `parseDateInput`:
- `"Mar 26, 2026"` → parses ✅
- `"Thu, Mar 26, 2026"` → **does not parse** ❌ — the month-name regex `^([A-Za-z]+)\s+(\d{1,2}),?\s+(\d{4})$` rejects the leading weekday token.

Two ways to handle this depending on scope:
- If we ship `weekday_date` under the **current** in-place-edit model, extend `parseDateInput` to strip/ignore a leading weekday token (localized) before matching, with test coverage.
- If we adopt the **focus-swap** model above, weekday-prefixed text is display-only and doesn't need parsing at all.

`long`, `date`, and `system_date` all round-trip today.

## Should this extend to the other date inputs?

Yes — same gap, same vocabulary. Design once and share:

- **`DateRangeInput`** — currently hardcodes `formatRangeDisplay` (`DATE_FORMAT_SHORT` / `DATE_FORMAT_SHORT_WITH_YEAR`, `start – end`). Should take the same `format` vocabulary per endpoint.
- **`DateTimeInput`** — already hardcodes `DATE_FORMAT_LONG` for its date portion (and has a separate `hourFormat` prop for time). Its date portion should use the same `format` prop; time stays governed by `hourFormat` / `hasSeconds`.

Recommendation: a single shared `format` type + preset→`Intl` mapping in `utils/plainDate.ts`, aligned with `TimestampFormat`, adopted across `DateInput`, `DateRangeInput`, and `DateTimeInput`.

## Scope

- **v1:** the blessed `Timestamp`-aligned presets (`long | date | weekday_date | system_date`) on the editable input, sharing one mapping.
- **Follow-on:** the focus-swap (display-when-blurred / editable-when-focused) model, which lets us later support arbitrary and lossy display formats without breaking editing.

## Open questions

- Confirm the default stays `long` (long month) while `date` means short-month, matching `Timestamp`.
- Do we want `weekday_date` in v1 (needs the parser change) or defer it to the focus-swap follow-on?
- Placeholder alignment when a short/numeric format is chosen.

Relevant code: `packages/core/src/DateInput/DateInput.tsx`, `packages/core/src/Timestamp/Timestamp.tsx` (`TimestampFormat`), `packages/core/src/utils/plainDate.ts`, `packages/core/src/utils/dateParser.ts`, `packages/core/src/DateRangeInput/DateRangeInput.tsx`, `packages/core/src/DateTimeInput/DateTimeInput.tsx`.

cixzhang:
## What

Adds a `format` prop to **DateInput** that controls how the committed date value is displayed, **reusing Timestamp's `format` value vocabulary verbatim** so the same string literal renders the same date shape in both components. Also extends **Timestamp** with two new shared members, `date_long` and `date_weekday`, giving the two components full value parity on the date-only formats.

```ts
export type DateInputFormat = Extract<
  TimestampFormat,
  'date' | 'date_long' | 'date_weekday' | 'system_date'
>;

format?: DateInputFormat | ((value: ISODateString) => string); // default 'date_long'
```

Selectable named values on DateInput:

| value | rendered (en-US, 2026-03-21) | shared with |
| --- | --- | --- |
| `date_long` (**default**) | `March 21, 2026` | `<Timestamp format="date_long" />` (new) |
| `date` | `Mar 21, 2026` | `<Timestamp format="date" />` |
| `date_weekday` | `Wed, Mar 21, 2026` | `<Timestamp format="date_weekday" />` (new) |
| `system_date` | `2026-03-21` | `<Timestamp format="system_date" />` |
| `(value) => string` | custom | — |

## Why

DateInput previously hard-coded a single long-month display. The reviewer ruling is that API/value reuse across related components trumps style conventions, so instead of forking a parallel enum, DateInput narrows Timestamp's own vocabulary with `Extract<TimestampFormat, …>`. The two types stay in compile-time lockstep — renaming or removing one of these members from `TimestampFormat` breaks `DateInputFormat` at build time. snake_case is kept because value reuse is the point.

## Full parity via a named `date_long` default (non-breaking)

DateInput's historical default is long-month (`March 21, 2026`). Rather than leave "long" as an unnamed default, this PR adds a real `date_long` member to Timestamp's vocabulary (mapping to `{year:'numeric', month:'long', day:'numeric'}` = `DATE_FORMAT_LONG`) and makes it DateInput's **default** (`format = 'date_long'`). So DateInput's default is now a real named value that matches a real Timestamp value — complete long-format parity — and its rendered output is **byte-identical** to the old hardcoded `plainDateFormat(…, DATE_FORMAT_LONG)` (verified). Existing usage is unchanged (zero codemod).

## Shared mapping (no drift)

The date-only members map through a single source of truth in `utils/plainDate.ts` (`SHARED_DATE_FORMAT_OPTIONS` + `formatSharedDate`). Timestamp's `formatTimestamp` switch and DateInput both consume it for the members they share (`date`, `date_long`, `date_weekday`; `system_date` is a fixed ISO string), so the two components can never diverge. `date_long` and `date_weekday` are added to `TimestampFormat` and its switch, so Timestamp consumers gain them too.

## Function escape hatch + caveat

`typeof format === 'function'` lets a power user return any string. Caveat: a custom string that `parseDateInput` cannot read back can't be re-committed after an inline edit; external `value` changes always recompute the display from the ISO value. Formatting applies **only to the committed value — never to in-progress typed input**, preserving today's typing UX. No change to `parseDateInput`, `value` (still `ISODateString`), `onChange`, or ARIA.

## Testing

- **DateInput:** default = `date_long` long-month; explicit `format="date_long"` identical to the default; `format="date"` short shape identical to Timestamp's `date`; `format="system_date"` ISO; `format="date_weekday"` weekday shape; function hatch; format does **not** apply to pending typed input; controlled value change recomputes. Existing tests remain green.
- **Timestamp:** new `date_long` (full month) and `date_weekday` (weekday prefix) members render correctly; existing format tests unchanged.
- **plainDate:** direct tests for `formatSharedDate` across all four shared members.
- Green locally: the DateInput + Timestamp + plainDate test files (191 tests), typecheck, ESLint (changed files), `@astryxdesign/core` build, `check:sync`, `sync:exports --check`, `check:changesets`.

Closes #4160

### DIFF
--- a/.changeset/dateinput-format-timestamp-vocabulary.md
+++ b/.changeset/dateinput-format-timestamp-vocabulary.md
@@ -2,6 +2,6 @@
 '@astryxdesign/core': patch
 ---
 
-[feat] DateInput gains a `format` prop for the committed date value, reusing Timestamp's `format` vocabulary so the same literal renders the same date shape in both components. Named values are `date` ("Mar 21, 2026"), the new `date_weekday` ("Wed, Mar 21, 2026"), and `system_date` ("2026-03-21"); a `(value) => string` function is also accepted for custom output. Left unset, the field keeps its long-month rendering ("March 21, 2026"), so existing usage is unchanged. This also extends Timestamp with the shared `date_weekday` member. Formatting applies only to the committed value, never to text being typed.
+[feat] DateInput gains a `format` prop for the committed date value, reusing Timestamp's `format` vocabulary so the same literal renders the same date shape in both components. Named values are `date_long` (the default, "March 21, 2026"), `date` ("Mar 21, 2026"), `date_weekday` ("Wed, Mar 21, 2026"), and `system_date` ("2026-03-21"); a `(value) => string` function is also accepted for custom output. The `date_long` default is byte-identical to DateInput's previous long-month rendering, so existing usage is unchanged. This also extends Timestamp with two new shared members, `date_long` and `date_weekday`, giving the two components full value parity on the date-only formats. Formatting applies only to the committed value, never to text being typed.
 
 @cixzhang
--- a/.changeset/dateinput-format-timestamp-vocabulary.md
+++ b/.changeset/dateinput-format-timestamp-vocabulary.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] DateInput gains a `format` prop for the committed date value, reusing Timestamp's `format` vocabulary so the same literal renders the same date shape in both components. Named values are `date` ("Mar 21, 2026"), the new `date_weekday` ("Wed, Mar 21, 2026"), and `system_date` ("2026-03-21"); a `(value) => string` function is also accepted for custom output. Left unset, the field keeps its long-month rendering ("March 21, 2026"), so existing usage is unchanged. This also extends Timestamp with the shared `date_weekday` member. Formatting applies only to the committed value, never to text being typed.
+
+@cixzhang
--- a/apps/storybook/stories/DateInput.stories.tsx
+++ b/apps/storybook/stories/DateInput.stories.tsx
@@ -57,6 +57,12 @@ const meta: Meta<typeof DateInput> = {
       options: [1, 2],
       description: 'Number of months to display in calendar',
     },
+    format: {
+      control: 'select',
+      options: [undefined, 'date', 'date_weekday', 'system_date'],
+      description:
+        "Display format for the committed value, reusing Timestamp's vocabulary. Unset renders a long-month date.",
+    },
   },
 };
 
@@ -83,6 +89,59 @@ export const WithValue: Story = {
   },
   args: {
     label: 'Event date',
+  },
+};
+
+/**
+ * The committed date value can be rendered in different shapes via `format`,
+ * reusing `Timestamp`'s format vocabulary so the same literal renders the same
+ * date shape in both components. Left unset, the value renders as a long-month
+ * date ("March 21, 2026"), unchanged from the historical default. While the
+ * user is typing, the raw text is shown verbatim — `format` applies only to the
+ * committed value.
+ */
+export const Formats: Story = {
+  render: () => {
+    const [value, setValue] = useState<ISODateString | undefined>(
+      '2026-03-21' as ISODateString,
+    );
+    return (
+      <div style={{display: 'flex', flexDirection: 'column', gap: 16}}>
+        <DateInput
+          label="Default (long month)"
+          value={value}
+          onChange={setValue}
+        />
+        <DateInput
+          label="format='date'"
+          value={value}
+          onChange={setValue}
+          format="date"
+        />
+        <DateInput
+          label="format='date_weekday'"
+          value={value}
+          onChange={setValue}
+          format="date_weekday"
+        />
+        <DateInput
+          label="format='system_date'"
+          value={value}
+          onChange={setValue}
+          format="system_date"
+        />
+        <DateInput
+          label="Custom function"
+          value={value}
+          onChange={setValue}
+          format={iso =>
+            new Intl.DateTimeFormat('en-GB', {dateStyle: 'full'}).format(
+              new Date(iso + 'T00:00'),
+            )
+          }
+        />
+      </div>
+    );
   },
 };
 
--- a/apps/storybook/stories/DateInput.stories.tsx
+++ b/apps/storybook/stories/DateInput.stories.tsx
@@ -59,9 +59,9 @@ const meta: Meta<typeof DateInput> = {
     },
     format: {
       control: 'select',
-      options: [undefined, 'date', 'date_weekday', 'system_date'],
+      options: ['date_long', 'date', 'date_weekday', 'system_date'],
       description:
-        "Display format for the committed value, reusing Timestamp's vocabulary. Unset renders a long-month date.",
+        "Display format for the committed value, reusing Timestamp's vocabulary. Defaults to 'date_long' (long-month date).",
     },
   },
 };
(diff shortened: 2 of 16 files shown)


cixzhang:
### Self-review vs Build Protocol — Quick Self-Review checklist

1. **Icons from registry** — n/a; no new icons. Existing `Icon` usage untouched.
2. **themeProps present** — n/a; no new interactive elements. `themeProps('date-input', …)` unchanged.
3. **No wrapper divs** — no new DOM. `format` only changes the computed display string passed to the existing `<input value>`.
4. **No CSS shorthands** — no style changes in this PR.
5. **ARIA still correct** — no ARIA/primitive changes. `role="combobox"`, `aria-*`, and the invalid-input live region are untouched; formatting affects only the committed display string.
6. **Tests pass locally** — DateInput + Timestamp + plainDate test files run green (188 tests), plus typecheck, ESLint (changed files), core build, `check:sync`, `sync:exports --check`, `check:changesets`.
7. **Sandbox/stories updated** — added `format` argType + a `Formats` story to DateInput; added a `date_weekday` option/story and AllFormats row to Timestamp; refreshed the Timestamp CLI showcase block. No props removed, so no callsite cleanup needed.
8. **Popover background** — n/a; no popover/layer changes.

Additional protocol notes:
- **Spec compliance / surface area** — built to the reviewer-refined contract: prop name `format`; exactly one new Timestamp member (`date_weekday`); unset default = long-month (non-breaking, no `date_long` member); `DateInputFormat = Extract<TimestampFormat, …>` keeps the two in compile-time lockstep. Exports: `DateInputFormat` from `DateInput/index.ts`; shared `formatSharedDate`/`SHARED_DATE_FORMAT_OPTIONS` from `utils/plainDate`.
- **Share implementation** — the shared date members map through one source of truth in `plainDate.ts`, consumed by both Timestamp's switch and DateInput, so the components can't drift.
- **Docs** — DateInput `.doc.mjs` (props / zh / dense) and Timestamp `.doc.mjs` updated; SYNC headers already list the touched files and `check:sync` is green.

cixzhang:
### Self-review update — `date_long` added and is now the default

Following the reviewer request, `date_long` is added to both components for full long-format parity, and it is now DateInput's default. Updated status vs the Build Protocol Quick Self-Review:

- **Timestamp gains `date_long` AND `date_weekday`** — both added to `TimestampFormat` and the `formatTimestamp` switch; `date_long` maps to `{year:'numeric', month:'long', day:'numeric'}` (= `DATE_FORMAT_LONG`).
- **`DateInputFormat`** = `Extract<TimestampFormat, 'date' | 'date_long' | 'date_weekday' | 'system_date'>` — still a strict `Extract` slice (compile-time lockstep).
- **Default is now `format = 'date_long'`** — a real named value that matches a real Timestamp value, replacing the previous `undefined → long` special case. **Verified byte-identical** to the old hardcoded `plainDateFormat(…, DATE_FORMAT_LONG)` output (e.g. `January 25, 2026`), so still non-breaking / zero codemod.
- **Single source of truth** — `date`, `date_long`, and `date_weekday` all map through `SHARED_DATE_FORMAT_OPTIONS` / `formatSharedDate` in `utils/plainDate.ts`, consumed by both Timestamp's switch and DateInput; `system_date` is the fixed ISO string in both. No drift possible.
- **Contract intact** — prop name `format`; snake_case reused verbatim; function hatch + round-trip caveat unchanged; formatting only on the committed value (never `pendingInput`); no change to `parseDateInput` / `value` / `onChange` / ARIA.
- **Stories / docs / tests updated** for `date_long`: Timestamp `.doc.mjs` + `DateLongFormat` story + AllFormats row + test; DateInput `.doc.mjs` (props/zh/dense) + `Formats` story default label + tests (default = `date_long`, explicit `date_long`); plainDate `formatSharedDate` test; Timestamp CLI showcase block; changeset text.

**Gates (re-run, all green):** DateInput + Timestamp + plainDate test files — **191 passed**; `typecheck` 0; ESLint (changed files) 0; `@astryxdesign/core` build 0; `check:sync` ✓; `sync:exports --check` ✓; `check:changesets` ✓.

PR #4200, opened 2026-07-22
feat(inputs): add statusVariant to the bordered input family (#4187)

### LINKED ISSUE #4187: Add `statusVariant` (attached/detached) as a shared convention across all inputs
## Summary

Add a **`statusVariant`** prop to **all input components** as a new shared input convention. It lets a consumer choose how an input's status message is rendered — **`attached`** (sits directly below the input, overlapping treatment) or **`detached`** (a separate element floated below with spacing) — threaded through the input to the `Field` it renders internally.

## Background — most of the machinery already exists

This is mostly a **plumbing/consistency** task, not new infrastructure:
- **`FieldStatus`** already supports `variant: 'attached' | 'detached'`.
- **`Field`** already accepts `statusVariant?: 'attached' | 'detached'` (default `'attached'`) and forwards it to `FieldStatus`.

The gap: individual input components render `<Field>` internally but **don't expose `statusVariant`** to the consumer, so there's no way to pass it through. A handful already thread it (`Slider`, `DateTimeInput`, `CheckboxList`, `RadioList`); most don't.

## Scope — thread `statusVariant` through every input

**Inputs that render `<Field>` and are MISSING the prop (add it, forward to Field):**
`TextInput`, `NumberInput`, `TextArea`, `DateInput`, `DateRangeInput`, `TimeInput`, `FileInput`, `Selector`, `MultiSelector`, `Typeahead`, `Tokenizer`.

**Already thread it (verify consistency, no change expected):**
`Slider`, `DateTimeInput`, `CheckboxList`, `RadioList`.

**Detached-only inputs — DO NOT get the option:** components whose status is inherently detached and that don't render a standard bordered Field (`Switch`, `CheckboxInput`) are **detached-only**; they should **not** expose an `attached` choice. Keep their status detached and omit the prop (or hard-fix it to `detached`).

## Conventions to nail down

- Prop name/type consistent everywhere: `statusVariant?: 'attached' | 'detached'`, default `'attached'` for bordered inputs (matches `Field`'s current default).
- Detached-only inputs are the documented exception — they don't offer the choice.
- Add `statusVariant` to the input-consistency section of the conventions so it's part of the standard input contract going forward, and so the component auditor can check for it.

## Acceptance criteria

- [ ] `statusVariant?: 'attached' | 'detached'` added to every Field-rendering input listed above and forwarded to the internal `Field`.
- [ ] Inputs that already thread it are verified consistent.
- [ ] `Switch` / `CheckboxInput` (detached-only) documented as exceptions and do NOT expose an attached option.
- [ ] Prop name, type, and default are identical across the family.
- [ ] Convention documented as a standard input prop; auditor check added.
- [ ] Tests, `.doc.mjs`, and stories updated per component to cover both variants where applicable.

cixzhang:
## Summary

Adds a single optional prop — `statusVariant?: 'attached' | 'detached'` (default `'attached'`) — to the **bordered input family**, forwarding it verbatim to the `Field` each input already renders. `Field`/`FieldStatus` have implemented both placements for a while; the bordered inputs just never exposed the knob, so they were locked to the implicit `attached` default. This threads it through.

- **`attached`** (default) — the status message sits directly below the input with an overlapping treatment. This is exactly today's behavior for these inputs.
- **`detached`** — the status message floats below as a separate element with spacing.

Closes #4187.

## Why

The placement behavior already exists at the primitive layer (`Field.statusVariant` → `FieldStatus`), but the bordered inputs couldn't reach it. Consumers who wanted a floated (detached) status under a bordered input had no supported way to get one. This is pure plumbing of an existing capability — no new infrastructure.

## What changed

**Gets the prop — bordered family (12):**
`TextInput`, `TextArea`, `NumberInput`, `DateInput`, `DateRangeInput`, `TimeInput`, `Selector`, `MultiSelector`, `Typeahead`, `Tokenizer`, `FileInput`, and `PowerSearch` (which forwards to the bordered input it renders internally).

Each one:
- Adds `statusVariant?: FieldStatusVariant` to its props interface (JSDoc mirroring `Field`'s).
- Destructures it with a default of `'attached'`.
- Forwards a single line, `statusVariant={statusVariant}`, to the `Field` it renders.

**Deliberately NOT touched — borderless / grouped controls (7):**
`DateTimeInput`, `CheckboxList`, `RadioList`, `Slider`, `InputGroup`, `Switch`, `CheckboxInput`.

These keep their hardcoded `detached` and get **no public prop**. `attached` is an *overlap* treatment designed for a full-width bordered input box; borderless and grouped controls have no such box to overlap, so `attached` renders wrong on them. Exposing a prop whose only non-default value is broken on those controls would be a footgun, so the prop is offered only where **both** values are valid. Their existing `detached` literal is left exactly as-is — zero source change.

## Non-breaking

The default `'attached'` matches what these inputs already produced (they rendered `Field` with no `statusVariant`, and `Field`'s own default is `'attached'`). Existing code compiles and renders identically; opting into `detached` is the only behavior change, and it's explicit.

## Type reuse

The prop reuses the existing `FieldStatusVariant` type (imported through `Field`'s barrel). The union is not redefined per component — one type, extensible in one place, no drift. No changes to `Field` or `FieldStatus`.

## Testing

- Added a colocated test to each of the 12 components asserting the value is forwarded: default renders the status element with `data-variant="attached"`; `statusVariant="detached"` renders it with `data-variant="detached"`.
- Added a `StatusVariantComparison` Storybook story to each component showing attached vs detached side by side.
- Ran the touched test files, plus `pnpm lint` (0 errors on changed files), `@astryxdesign/core` typecheck + `typecheck:docs`, storybook typecheck, `pnpm build`, `sync-exports --check`, `verify-exports`, `generate-token-docs --check`, and changeset validation — all green.
- Updated each component's `.doc.mjs` prop table (all locale/dense variants present) with the new prop.

### DIFF
--- a/.changeset/field-status-detached-status-icon.md
+++ b/.changeset/field-status-detached-status-icon.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] FieldStatus renders a leading status icon on the `detached` message so status is not conveyed by color/position alone (WCAG 1.4.1). The icon is decorative for assistive tech; the message text and live-region announcement carry the status. The `attached` variant is unchanged.
+
+@cixzhang
--- a/.changeset/input-status-variant.md
+++ b/.changeset/input-status-variant.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] The bordered input family now accepts a `statusVariant` prop (`'attached' | 'detached'`, default `'attached'`) that forwards to the underlying `Field`, letting you float the status message below the input with spacing instead of overlapping it. Added to TextInput, TextArea, NumberInput, DateInput, DateRangeInput, TimeInput, Selector, MultiSelector, Typeahead, Tokenizer, FileInput, and PowerSearch. Non-breaking: the default matches today's behavior. (#4187)
+
+@cixzhang
--- a/apps/storybook/stories/DateInput.stories.tsx
+++ b/apps/storybook/stories/DateInput.stories.tsx
@@ -424,3 +424,31 @@ export const ClearableWithStatus: Story = {
     status: {type: 'error', message: 'Date is in the past'},
   },
 };
+
+export const StatusVariantComparison: Story = {
+  render: () => {
+    const [a, setA] = useState<ISODateString | undefined>(
+      '2026-01-25' as ISODateString,
+    );
+    const [b, setB] = useState<ISODateString | undefined>(
+      '2026-01-25' as ISODateString,
+    );
+    return (
+      <div style={{display: 'flex', flexDirection: 'column', gap: 24, width: 280}}>
+        <DateInput
+          label="Attached (default)"
+          value={a}
+          onChange={setA}
+          status={{type: 'error', message: 'This date is not available'}}
+        />
+        <DateInput
+          label="Detached"
+          value={b}
+          onChange={setB}
+          status={{type: 'error', message: 'This date is not available'}}
+          statusVariant="detached"
+        />
+      </div>
+    );
+  },
+};
--- a/apps/storybook/stories/DateRangeInput.stories.tsx
+++ b/apps/storybook/stories/DateRangeInput.stories.tsx
@@ -333,3 +333,27 @@ export const AllVariations: Story = {
     );
   },
 };
+
+export const StatusVariantComparison: Story = {
+  render: () => {
+    const [a, setA] = useState<DateRange | null>(null);
+    const [b, setB] = useState<DateRange | null>(null);
+    return (
+      <div style={{display: 'flex', flexDirection: 'column', gap: 24, width: 320}}>
+        <DateRangeInput
+          label="Attached (default)"
+          value={a}
+          onChange={setA}
+          status={{type: 'error', message: 'Please select a date range'}}
+        />
+        <DateRangeInput
+          label="Detached"
+          value={b}
+          onChange={setB}
+          status={{type: 'error', message: 'Please select a date range'}}
+          statusVariant="detached"
+        />
+      </div>
+    );
+  },
+};
--- a/apps/storybook/stories/FileInput.stories.tsx
+++ b/apps/storybook/stories/FileInput.stories.tsx
@@ -282,3 +282,27 @@ export const AllVariations: Story = {
     );
   },
 };
+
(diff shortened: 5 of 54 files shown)


cixzhang:
## Self-review — against the Component Build Protocol (Quick Self-Review for updates)

This is a plumbing/convention change (exposing an existing `Field`/`FieldStatus` capability), so the lighter refactor checklist applies.

1. **Icons from registry** — n/a; no icons touched.
2. **themeProps present** — n/a; no new interactive elements. The forwarded prop reaches `FieldStatus`, which already emits `data-variant` via `themeProps`.
3. **No wrapper divs** — n/a; no new DOM. The prop is a single forwarded attribute on the existing `<Field>`.
4. **No CSS shorthands** — n/a; no styles changed.
5. **ARIA still correct** — no primitive/hook swaps. `detached` vs `attached` only changes whether children + status share a wrapper vs are siblings (existing `Field` behavior); status roles (`alert`/`status`) are unchanged. Verified by the touched components' tests.
6. **Tests pass locally** — ran all 12 touched test files (green), then the full suite: 7373 passed, 0 failures. Also: `pnpm lint` (0 errors on changed files), core `typecheck` + `typecheck:docs`, storybook `typecheck`, `pnpm build`, `sync-exports --check`, `verify-exports`, `generate-token-docs --check`, changeset validation — all green.
7. **Stories updated** — added a `StatusVariantComparison` story to each of the 12 components (attached vs detached side by side). No props removed.

### Spec-compliance notes
- **Scope exact:** 12 bordered inputs get the prop; the 7 borderless/grouped controls (`DateTimeInput`, `CheckboxList`, `RadioList`, `Slider`, `InputGroup`, `Switch`, `CheckboxInput`) are untouched — their hardcoded `detached` is preserved, and there is zero source change to them (or to `Field`/`FieldStatus`). Confirmed by the diff.
- **Type reuse:** every component imports `FieldStatusVariant` through `Field`'s barrel; the `'attached' | 'detached'` union is not redefined anywhere.
- **Non-breaking:** default is `'attached'`, matching the prior implicit behavior.
- **PowerSearch:** forwards `statusVariant` to the bordered input it renders internally, which forwards to `Field` — consistent with the "input owns its Field" model.
- **Family alignment:** the prop name and JSDoc mirror the existing `Field.statusVariant`.

PR #4201, opened 2026-07-22
fix(banner): move the 'banner-icon' theme target onto the status Icon so 'status:X' theme overrides reach the glyph

### LINKED ISSUE #4166: Banner: info-status icon is hard-coded to Icon color="accent", unreachable by 'status:info' theme overrides
## Summary

`Banner` renders its `status="info"` icon as `<Icon color="accent">`. The banner surface itself can be themed via the documented component overrides (`'astryx-banner'` / `'astryx-banner-icon'` with `'status:info'`), but those only style the wrapper elements — the inner `.astryx-icon` span carries `data-color="accent"` and keeps the accent color.

On themes whose accent is not a calm blue this breaks the semantics of the status system: our theme's accent is red (`#DC2626`), so every info banner ships a **red** (i) icon — visually indistinguishable from an alert. The same accent-fallback applies to `Badge variant="info"`, though Badge at least is fully recoverable via `'variant:info'` overrides; the Banner icon is not.

## Repro

```ts
// theme
export const theme = defineTheme({
  tokens: { "--color-accent": "#DC2626" /* red accent */ },
  components: {
    banner: { "status:info": { backgroundColor: "#67E8F91A" } },
    "banner-icon": { "status:info": { color: "#67E8F9" } },
  },
});
```

```tsx
<Banner status="info" title="Mode: shadow" description="Nothing actually delivers." />
```

Rendered DOM (core 0.1.2):

```html
<div class="astryx-banner card info ..." data-container="card" data-status="info">
  <div class="astryx-banner-icon info ...">              <!-- override applies: color #67E8F9 -->
    <span class="astryx-icon md accent ..." data-color="accent">  <!-- stays red -->
      <svg stroke="currentColor" ...>
```

`getComputedStyle` confirms: banner-icon wrapper `color: rgb(103, 232, 249)` (override works), inner svg `stroke: rgb(220, 38, 38)` (accent, unreachable).

## Expected

Either of:
- the status icon uses `currentColor` / inherits from `.astryx-banner-icon` so `'status:info'` overrides reach it, or
- info status resolves through a dedicated token (e.g. `--color-info`, accent as fallback) that themes can set.

## Workaround

A descendant rule outside the theme DSL:

```css
.astryx-banner-icon.info .astryx-icon { color: #67e8f9; }
```

…which has to be hand-maintained since `astryx theme build` output can't express it.

Related: the accent-fallback for info semantics generally (Badge `variant="info"`, Banner `status="info"`) might deserve a `--color-info` token — happy to split that into a separate issue if useful.

jiunshinn:
## Problem

`Banner` renders its status icon as an `<Icon>` carrying its own color variant (`info` → `color="accent"`, i.e. `color: var(--color-accent)` on the `.astryx-icon` span). The documented theming target `astryx-banner-icon` (+ `'status:info'`) compiled to a selector that matched the **layout wrapper div** only, so a `'banner-icon'` override that sets `color` never reached the glyph. On themes whose accent is not a calm blue (the reporter's accent is `#DC2626` red), every info banner ships a red (i) icon indistinguishable from an alert, with no recovery path inside the theme DSL.

## Root cause

The theme target and the paint lived on **different elements**: `themeProps('banner-icon', {status})` decorated the wrapper (whose styles are purely layout: `display:flex`, `alignItems:center`, `flexShrink:0`), while the color sat on the inner `.astryx-icon` span via the Icon color variant. Theme component overrides emit same-element rules (`.astryx-banner-icon.info` in `@layer astryx-theme`), so they restyled the wrapper while the SVG (`stroke`/`fill: currentColor`) kept reading the span's own hard-coded color.

## Fix — target the element that paints

Per [review](https://github.com/facebook/astryx/pull/4201#pullrequestreview-4903069411) and [Theming Infrastructure § Adding a target](https://github.com/facebook/astryx/wiki/Theming-Infrastructure): the target moves onto the `<Icon>` itself.

```tsx
<Icon
  icon={defaultIconName}
  size="md"
  color={iconColor}                          // unchanged Icon color variant
  {...themeProps('banner-icon', {status})}   // target now on the glyph
/>
```

- The Icon keeps its existing color variant (`info` → accent, others → their semantic status tokens), so the **default DOM paints identically**. `Icon` merges the incoming `className` with its own `themeProps('icon', …)` classes via `mergeProps`, and `data-status` rides through `spanProps` onto the span.
- A theme's `'banner-icon'` + `'status:X'` override (`.astryx-banner-icon.info`) now lands on the **same element** as the variant and wins from `@layer astryx-theme` — the same mechanism `Selector` documents for its check/clear icons.
- The wrapper is back to **layout-only** and no longer carries the target on the default path, mirroring #4838's chevron placement three lines away in this same file.
- No `color="inherit"`, no `icon == null` style conditional, no `data-color` change — the documented `astryx-icon` surface (`Icon.doc.mjs`) is untouched.

## Custom `icon` nodes — judgment call, flagged for review

With a consumer-provided `icon` ReactNode, core cannot spread the target onto the node (`no-react-introspection` bans prop injection into consumer elements). For that path only, the target **stays on the wrapper**: it is the nearest element that can carry it, and overrides reach the node via ordinary inheritance. This keeps the documented surface working for custom icons instead of silently dropping it. If you'd rather the target exist only for the default icon, that's a three-line narrowing.

## Contract note (as requested)

On the default-icon path, `.astryx-banner-icon` now matches the **icon element** instead of the flex wrapper. A theme using `banner-icon` for wrapper layout (margin, alignment) would now style the glyph. No in-repo theme does this; the changeset carries the note.

## Test plan

- `pnpm exec vitest run packages/core/src/Banner` — 38 passed (35 existing + 3 new)
- New tests pin: (1) every status renders the target on the glyph (`.astryx-icon.astryx-banner-icon.<status>` + `data-status`, and exactly one element carries the target); (2) regression pin — the theme-target element carries the paint (`data-color="accent"` for info); pre-fix the target was the wrapper and the variant sat on an inner span the override could never reach; (3) a custom `icon` keeps the wrapper target and the node itself is undecorated. (1) and (2) fail against main; (3) pins the unchanged custom-icon contract.
- `eslint src/Banner` — no new warnings (same single pre-existing `no-nullish-jsx-guard` hit as main)
- Changeset updated with the wrapper→glyph contract note; `check:changesets` green

## History

The first version of this PR (wrapper-carried status color + `Icon color="inherit"`) predates the target-placement decision (#4758 / #4775 / #4838 / #4846) and was reworked per the review above. Option (b) from the issue (a `--color-info` token) remains out of scope, as agreed.

Fixes #4166

### DIFF
--- a/.changeset/banner-info-icon-themeable.md
+++ b/.changeset/banner-info-icon-themeable.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Banner: the 'banner-icon' theme target now rides on the default status Icon itself instead of its layout wrapper, so theme component overrides ('banner-icon' + 'status:X') that set color actually reach the glyph. The Icon keeps its existing color variant (info still renders accent) and same-element rules in @layer astryx-theme win over it, so default rendering is unchanged. Contract note: '.astryx-banner-icon' now matches the icon element rather than the wrapper when the default icon renders; a theme that used the target for wrapper layout (margin, alignment) now styles the glyph instead. With a custom `icon` node the target stays on the layout-only wrapper, since core never injects props into consumer elements (#4166)
+
+@jiunshinn
--- a/packages/core/src/Banner/Banner.test.tsx
+++ b/packages/core/src/Banner/Banner.test.tsx
@@ -3,7 +3,8 @@
 /**
  * @file Banner.test.tsx
  * @input Uses vitest, @testing-library/react, Banner component
- * @output Unit tests for Banner component behavior
+ * @output Unit tests for Banner component behavior, including the
+ *   'banner-icon' theme target riding on the status icon glyph (#4166)
  * @position Testing; validates Banner.tsx implementation
  *
  * SYNC: When modified, update this header
@@ -335,6 +336,62 @@ describe('Banner', () => {
     const ref = {current: null as HTMLDivElement | null};
     render(<Banner ref={ref} status="info" title="Ref Test" />);
     expect(ref.current).toBeInstanceOf(HTMLDivElement);
+  });
+
+  // =========================================================================
+  // Status icon color theming (#4166)
+  // =========================================================================
+
+  it("carries the 'banner-icon' theme target on the default status icon glyph", () => {
+    // Theme overrides for 'banner-icon' + 'status:X' compile to
+    // '.astryx-banner-icon.<status>' (parseStyleKey). The target must sit on
+    // the <Icon> span itself so those same-element rules in
+    // @layer astryx-theme beat the Icon's own color variant.
+    const statuses = ['info', 'warning', 'error', 'success'] as const;
+    for (const status of statuses) {
+      const {container, unmount} = render(
+        <Banner status={status} title={`${status} banner`} />,
+      );
+      const glyph = container.querySelector(
+        `.astryx-icon.astryx-banner-icon.${status}`,
+      );
+      expect(glyph).not.toBeNull();
+      expect(glyph).toHaveAttribute('data-status', status);
+      // Exactly one element carries the target — the layout wrapper no
+      // longer does.
+      expect(container.querySelectorAll('.astryx-banner-icon')).toHaveLength(1);
+      unmount();
+    }
+  });
+
+  it('keeps the color variant on the theme-target element (regression pin for #4166)', () => {
+    // Pre-fix, '.astryx-banner-icon.info' matched the layout wrapper while
+    // the color variant (data-color="accent") sat on an inner span that a
+    // theme override could never reach. Target and paint now share one
+    // element.
+    const {container} = render(<Banner status="info" title="Info" />);
+    const target = container.querySelector('.astryx-banner-icon.info');
+    expect(target).toHaveAttribute('data-color', 'accent');
+  });
+
+  it("keeps the 'banner-icon' target on the wrapper for a custom icon node", () => {
+    // Core never injects props into consumer elements, so with a custom
+    // `icon` the target stays on the (layout-only) wrapper and overrides
+    // reach the node via inheritance. The node itself is untouched.
+    const {container} = render(
+      <Banner
+        status="info"
+        title="Custom icon"
+        icon={<span data-testid="custom-glyph">i</span>}
+      />,
+    );
+    const targets = container.querySelectorAll('.astryx-banner-icon');
+    expect(targets).toHaveLength(1);
+    expect(targets[0]?.tagName).toBe('DIV');
+    expect(targets[0]).toHaveAttribute('aria-hidden', 'true');
+    const custom = container.querySelector('[data-testid="custom-glyph"]');
+    expect(custom).not.toBeNull();
+    expect(custom?.className).toBe('');
   });
 
   // =========================================================================
--- a/packages/core/src/Banner/Banner.tsx
+++ b/packages/core/src/Banner/Banner.tsx
@@ -12,6 +12,9 @@
  * - Root container: layout-only wrapper (flex column), no visual styling, no theme target
  * - Header area (themeProps 'banner'): colored status background with icon, title, description, actions, dismiss
  * - Content area (themeProps 'banner-content'): collapsible card background for additional content (children)
+ * - Status icon (themeProps 'banner-icon'): the target rides on the default
+ *   <Icon> itself — the element that paints — so 'status:X' overrides reach
+ *   the glyph (#4166); for a custom `icon` node it stays on the layout wrapper
  * - No left border accent — color is expressed through the full header background
  * - Each visual area owns its own border-radius (no overflow:clip on the container)
  * - When children are provided, a collapse/expand toggle button appears in the end area
@@ -445,16 +448,32 @@ export function Banner({
                 : styles.headerCardStandalone),
           ),
         )}>
+        {/* The 'banner-icon' target rides on the element that paints: the
+            default <Icon> below, or this wrapper when a custom `icon` node
+            is passed (core never injects props into consumer elements, so
+            overrides reach it via inheritance). The wrapper itself stays
+            layout-only. */}
         <div
-          {...mergeProps(
-            themeProps('banner-icon', {status}),
-            stylex.props(styles.iconWrapper),
-          )}
+          {...(icon != null
+            ? mergeProps(
+                themeProps('banner-icon', {status}),
+                stylex.props(styles.iconWrapper),
+              )
+            : stylex.props(styles.iconWrapper))}
           aria-hidden="true">
           {icon != null ? (
             icon
           ) : (
-            <Icon icon={defaultIconName} size="md" color={iconColor} />
+            // Applied to the status <Icon> itself rather than the wrapper, so
+            // the element that paints the glyph is the element a theme
+            // targets — a 'banner-icon' 'status:X' color override beats the
+            // Icon's own variant from @layer astryx-theme (#4166).
+            <Icon
+              icon={defaultIconName}
+              size="md"
+              color={iconColor}
+              {...themeProps('banner-icon', {status})}
+            />
           )}
         </div>
         <div {...stylex.props(styles.headerContent)}>


cixzhang (changes requested):
🔴 **Blocking — `needs:code-review` + `needs:design-review` (contributor PR touching `packages/core` runtime + theming surface).** The diagnosis and the fix both hold up under verification; what blocks is a **tier** question — the mechanism runs against a theming decision that landed after this PR was opened.

`Triage: bug fix · non-breaking (defaults byte-identical) · medium blast radius (theming contract) → standard path · checks: §2 T1/T2/T2b/T8, §3 P11/P12, §8 X20`

---

## First: thank you, and sorry about the wait

Three weeks with no reply on a PR this well-prepared is on us, not you. For the record, this is one of the better-argued PRs in the queue: you named both options from the issue, said which you picked and why, pre-empted the compatibility question, and flagged your own observable-DOM risk in reviewer question 3. That made this review much faster to do properly.

**Everything you claim, I verified — and it all checks out:**

| Claim | Result |
|---|---|
| #4166 reproduces | ✅ Confirmed, all four statuses. With the issue's theme (accent `#DC2626`), the wrapper computes to the override (`rgb(103,232,249)`) while the glyph paints `rgb(220,38,38)`. |
| The fix reaches the glyph | ✅ All four statuses now paint the override color (`svgFill` follows `banner-icon`). |
| Default rendering unchanged | ✅ **Rigorously.** Before/after screenshots of the unthemed render are **byte-identical** (same md5; 0 differing pixels of 1,785,600). |
| In-repo themes keep working | ✅ Confirmed — `neutral` re-scopes `--color-accent`/`--color-success`/… on the **banner root**, so the wrapper's `var()` resolves inside the same subtree. |
| Tests | ✅ `vitest run packages/core/src/Banner` → **38/38 pass** on a clean merge with current `main`. Merges cleanly. |
| Changeset | ✅ Present (X20 satisfied). |

No `T1` (no raw color literals — the map uses `colorVars`), no `T8`, and `color: 'inherit'` is explicitly sanctioned by `T2b`. The code itself is clean.

---

## 🔴 The blocker: this puts a paint property back onto a layout-only wrapper

Between when you opened this (Jul 22) and now, we landed a decision: **theming targets attach to the component that _paints_, not to a wrapper. Layout is not a theme seam.**

- **#4758** (Aug 7) — `feat(lint): flag div/span wrappers that only style an Astryx component`
- **#4775** (Aug 7) — *"Style-only wrappers are the root cause of wrapper-shaped theme targets… Theming targets should attach to the component that paints; layout is not a theme seam."*
- **#4838** (Aug 9) — `fix(core): drop styling wrappers around icons; render glyphs through Icon`
- **#4846** (Aug 10) — `fix(selector): drop the shared trigger-icon wrapper`

This isn't only a PR-precedent argument — it's the written rule. [Theming Infrastructure § Adding a target](https://github.com/facebook/astryx/wiki/Theming-Infrastructure), rule 1:

> Target the element that carries the styling. A target must sit on the element that actually paints the thing you're theming — never a wrapper that exists only for layout, positioning, or event handling. **If a decorative glyph is wrapped in an inert/positioning div, the target belongs on the glyph, not the wrapper.**

The Banner status icon is exactly that shape: an `aria-hidden` decorative glyph inside a wrapper whose styles on `main` are purely `{display:flex, alignItems:center, flexShrink:0}`.

**The sharpest version:** #4838 already edited *this same file*, on *the icon sitting next to yours*. It deleted the chevron's `<span {...stylex.props(styles.chevron)}>` wrapper and moved the styling onto the `<Icon>` via `xstyle`, with this comment:

```
// Applied to the chevron <Icon> itself (via `xstyle`) rather than a wrapper,
// so the element that rotates is the element a theme targets.
```

Your PR predates that commit, so there was no way to know. But merged as-is, `Banner.tsx` would end up with the chevron styled on the glyph and the status icon styled on the wrapper — opposite conventions, three lines apart.

I want to be precise about severity, because two things here are **not** true:
- It is **not** a lint failure. `no-style-only-wrapper` deliberately exempts wrappers carrying `aria-*`/`data-*`/other spreads, and this one has both. It passes.
- It is **not** a `T2` themeable-surface removal. The `banner-icon` target still exists, and net themeability goes *up*.

So this blocks on **direction**, not on a bright-line rule — which is why the remedy below is a redirect rather than a rejection.

## The alternative — and yes, it works

Put the theme target on the `<Icon>` itself, the way `Selector`'s chevron already does:

```tsx
<Icon
  icon={defaultIconName}
  size="md"
  color={statusIconColor[status]}   // keep the existing Icon color variant
  {...themeProps('banner-icon', {status})}
/>
```

I verified the mechanism is available:
- `Icon` spreads rest props onto its rendered span (`spanProps`), so a `themeProps` spread lands on the glyph.
- Same-element rules in `@layer astryx-theme` sit above StyleX, so a `banner-icon` `color` override beats the Icon's own `colorStyles[color]` on that element. `Selector` relies on exactly this and documents it:
  > *"Same-element rules in @layer astryx-theme win over the icon's own base color/size, which a button-level target could not reach."*

That fixes #4166 **without** `color="inherit"`, without the `icon == null` conditional, and without touching `data-color` — so your reviewer question 3 dissolves entirely. The wrapper keeps only genuine layout.

**One real design consequence, and it's yours to weigh, not a gotcha:** moving `.astryx-banner-icon` onto the glyph means a theme currently using `banner-icon` for *layout* (margin, size, alignment) would now be styling the icon rather than the flex box. No in-repo theme does this, but it is a contract change to a documented target, so it likely wants a `[breaking]`-flavored note in the changeset even though the visual default is unchanged. Worth calling out explicitly in the description.

## The fair counter-argument

There is a genuine case for your approach, and I don't want to pretend otherwise: `banner-icon` is a **pre-existing, documented** target that already sits on the wrapper (`Banner.doc.mjs`, `themeProps('banner-icon', {status})`). Your change makes an existing documented promise actually work, and is the smaller, safer diff. Reading the wiki rule strictly, though, that wrapper target is itself the pre-existing defect — and this PR would resolve the contradiction by giving the wrapper paint duties, which is the direction we just spent four PRs moving away from. **I'd rather fix the target's placement than entrench it.** That's my call as TL, and I'm making it explicitly so you're not guessing.

## Two smaller notes

1. **Not "outside the documented theming surface."** Your reviewer question 3 undersells this one: `astryx-icon` *is* a documented target with `visualProps: ['color', 'size']` (`Icon.doc.mjs`), and `Icon` emits `themeProps('icon', {size, color})`. So flipping the glyph to `data-color="inherit"` does narrow a documented surface — a theme styling `icon` + `color:error` globally would stop reaching Banner glyphs. No in-repo theme does this, so it's minor, but it's a real surface, not an undocumented one. The Icon-target approach avoids it entirely.
2. **Unrelated reformat.** The diff reformats the `label`/`tooltip` ternaries on the collapse `Button`. Harmless, but it's churn in a hunk #4838 has since rewritten — worth dropping on rebase.

## What I'd like to happen

To state the rule plainly, since this is a TL call and I don't want you guessing at it:

> **Move the theme target onto the `<Icon>`, and make sure the wrapper carries no paint styles. Layout-only styles can stay on the wrapper.**

That's the general principle, not a one-off for Banner: the wrapper is welcome to keep `display:flex`, `alignItems`, `flexShrink` and anything else purely about placement. What it must not carry is anything that paints — color, background, border, shadow. Those belong on the element that renders the pixels.

Concretely: **rebase onto `main`** (you'll see #4838's chevron change land next to yours) and **flip the mechanism to the Icon-carried target**. Keep the changeset, keep the test intent — your three tests are pinned to a real reported bug, which is exactly right; they'll just assert the target on the glyph instead of `data-color="inherit"`. If you'd rather not respin it, say so and I'll take it from here with credit to you — you've done the hard part, which was the diagnosis.

On the issue's option (b) (`--color-info`): still worth doing, still a separate issue, and independent of this. Agreed on keeping it out of scope.

Sorry again for the silence. Ping me directly on the respin and I'll turn it around quickly.

jiunshinn:
@cixzhang Thank you for the thorough review — verifying every claim independently and spelling out both the rule and the intended shape (the #4838 chevron comment especially) made this an easy respin. Reworked in 6a17790:

- **Rebased onto main** — the unrelated ternary reformat dissolved in the rebase (#4838's version won).
- **Target moved onto the `<Icon>`**: `color={statusIconColor[status]}` stays exactly as it was pre-PR, and `{...themeProps('banner-icon', {status})}` is spread onto the Icon, so the same-element rule in `@layer astryx-theme` beats the Icon's own color variant. `color="inherit"`, the wrapper color styles, and the `data-color` change are all gone — the wrapper is back to layout-only, and your note 1 (the documented `astryx-icon` surface) is moot.
- **Tests re-pinned to the new contract**: `.astryx-banner-icon.<status>` must match the element that carries the paint, and exactly one element carries the target. Both fail against main; 38/38 pass here.
- **Changeset carries the contract note**: on the default-icon path `.astryx-banner-icon` now matches the glyph rather than the flex wrapper.

One judgment call to flag: for a **custom `icon` node** the target stays on the wrapper. Core can't spread props onto a consumer-provided node (`no-react-introspection`), so the wrapper is the nearest element that can carry the target, and overrides still reach the node via inheritance. That keeps the documented surface alive on that path instead of silently dropping it — happy to narrow it to default-icon-only if you'd prefer.

PR description updated to match. Thanks again for the redirect and the offer to take it over — the diagnosis surviving intact made the respin the easy half.

cixzhang (approved):
Thanks for fixing this!

PR #4203, opened 2026-07-22
feat(cli): refresh stale managed agent-docs blocks during upgrade

### LINKED ISSUE #4168: CLI: `astryx upgrade` leaves the managed AGENTS.md block stale — should refresh agent docs after a version bump
## Summary

`astryx upgrade` runs code codemods but does not refresh the managed agent-docs block (`<!-- ASTRYX:START --> … <!-- ASTRYX:END -->`) that `astryx init --features agents` installs into AGENTS.md / CLAUDE.md.

After bumping `@astryxdesign/core` 0.1.2 → 0.1.7 and running `astryx upgrade`, our AGENTS.md still said:

```
Astryx v0.1.2 · 148 components
```

…while the installed library was 0.1.7 with 150 components and materially different agent guidance (the 0.1.7 block adds the frame-first rule, dense-data-as-rows, StatusDot/Token-for-status, and the self-check rule — none of which agents see until someone remembers to re-run `init`).

This matters more than a normal docs staleness: the block exists specifically to steer AI agents, and an agent reading a stale index will query missing components, follow superseded rules, and skip new ones. The whole value of the managed block is that it's trustworthy.

## Expected

`astryx upgrade --apply` detects an existing managed block (any file containing the `ASTRYX:START` marker, or the paths `init` knows about) and rewrites it for the installed version — or at minimum prints a loud next-step: "Agent docs block is at v0.1.2 (installed: v0.1.7) — run `astryx init --features agents` to refresh."

Dry-run mode could report the block as one more pending change alongside the codemods.

## Actual

`upgrade` completes ("No changes needed — your code is already up to date!") with the block still at the old version. Nothing mentions the agent docs.

## Repro

1. Project with `@astryxdesign/core@0.1.2` and `astryx init --features agents` previously run.
2. `yarn add @astryxdesign/core@^0.1.7 @astryxdesign/cli@^0.1.7`
3. `npx astryx upgrade --from 0.1.2 --apply`
4. AGENTS.md managed block still reads `Astryx v0.1.2 · 148 components`.

### LINKED ISSUE #4169: astryx upgrade --apply should refresh managed agent-docs block
When running `astryx upgrade --apply`, only code codemods are applied. The managed agent-docs block (between `<!-- ASTRYX:START -->` and `<!-- ASTRYX:END -->`) is not refreshed with the new component index and agent guidance.

Currently, after an upgrade, users must separately run `astryx init --features agents` to get the updated agent-docs block. This should be handled by `upgrade --apply`.

jiunshinn:
## Problem

`astryx upgrade` runs codemods but leaves the managed agent-docs block (`<!-- ASTRYX:START --> … <!-- ASTRYX:END -->`) stale. After bumping core 0.1.2 → 0.1.7 and running `upgrade --apply`, AGENTS.md still says `Astryx v0.1.2 · 148 components`, so agents keep following superseded guidance and querying a stale index — while the CLI reports that no changes are needed.

Two gaps on `main`:

1. The existing refresh in `upgrade` only runs on the *main* pipeline path. The early returns — `--from >= installed` ("Already up to date") and "No codemods available for this version range" (the issue's exact repro: a range with no codemods) — skipped it entirely.
2. Dry-run **wrote** the agent docs instead of reporting them as a pending change, and the refresh was unconditional (no staleness detection, no idempotence, no missing-block nudge).

## What this PR does

One detection pass, `auditAgentDocs()` (in `agent-docs.mjs`, next to init's writer), compares every managed block against what the installed version generates today — reusing `generateCompressedIndex` / `getXdsVersion` / `detectStylingSystem`, not duplicating them. A block is *stale* when its text differs from the freshly generated index; the parsed `Astryx vX.Y.Z` stamp is only used for messaging. Detection covers the canonical `AGENT_DOC_PATHS` **plus** a shallow root-level marker scan of `*.md`/`*.mdc` files (for `init --agent-docs-path` custom targets), and legacy `<!-- XDS:START -->` blocks.

`upgrade` runs this pass (`processAgentDocs`) on **every** successful path — main pipeline *and* both early returns — so a stale block can never hide behind a no-changes-needed message.

## Behavior matrix

| Scenario | Before | After (dry-run) | After (`--apply`) |
| --- | --- | --- | --- |
| Stale block, codemod range | Refreshed only on main path; **dry-run wrote the file** | `⚠ Agent docs block in AGENTS.md is at v0.1.2 (installed: v0.1.7).` + `--apply` hint; file untouched | Block rewritten in place; `✓ Agent docs updated: AGENTS.md` |
| Stale block, `--from` == installed ("already up to date") | Nothing — early return skipped the refresh | Same pending-change report; outro says `Dry run complete` instead of `Done` | Block rewritten even though no codemods run |
| Stale block, empty codemod range ("no codemods available") | Nothing (issue repro) | Same pending-change report | Block rewritten |
| Up-to-date block | Rewritten anyway (mtime churn) | Silent; `refreshed: []`, `stale: []` | Untouched — idempotent (audit compares full block text, so a refreshed block audits fresh) |
| No block anywhere, no `astryx.config` | Silent | `⚠ No agent docs found — run \`astryx init\` to install the component index for AI agents.` | Same nudge |
| No block, but `astryx.config` present | Silent | Silent (reads as a deliberate no-agent-docs setup) | Silent |
| Legacy `XDS` markers | Refreshed on main path only | Reported stale | Migrated to `ASTRYX` markers |
| Block in custom root-level file (e.g. `NOTES.md` via `--agent-docs-path`) | Never detected | Reported stale | Rewritten |

Everything outside the markers is preserved byte-for-byte (the rewrite goes through init's own `injectXdsBlock`, which only replaces between the markers).

`--json`: the `up_to_date` / `no_codemods` status envelopes and the `upgrade.run` receipt now carry a structured `agentDocs` field — `{installedVersion, refreshed, stale: [{path, blockVersion}], missing}`. The existing `receipt.agentDocsRefreshed` boolean is kept for compatibility.

## Implementation notes

- `packages/cli/src/commands/agent-docs.mjs`: new exported `auditAgentDocs(targetDir)` + private `extractXdsBlock` / `BLOCK_VERSION_RE`. No behavior change to init/agent-docs flows.
- `packages/cli/src/commands/upgrade.mjs`: new `processAgentDocs({apply, json, from})` helper wired into the two early returns and the main pipeline (replacing the previous unconditional `installAgentDocs(..., {onlyReplace: true})` call). The rewrite path reuses `installAgentDocs` with explicit `paths` (only the stale files), so it inherits init's path-safety checks and writer.
- Best-effort: any failure in the pass warns (human mode) and never breaks the upgrade — matching the previous behavior.

## Test plan

New colocated suite `packages/cli/src/commands/upgrade.agent-docs.test.mjs` (12 tests): stale block refreshed on apply with outside-content byte-for-byte preservation; dry-run reports (JSON + human wording) without writing; idempotent second run (`refreshed: []`, content unchanged); missing-block nudge (JSON + human), suppressed when `astryx.config` exists; custom root-level file via marker scan; legacy XDS migration; full codemod-pipeline dry-run carries `agentDocs` in the receipt and does not write.

- `pnpm exec vitest run packages/cli/src/commands/upgrade.agent-docs.test.mjs packages/cli/src/commands/upgrade.test.mjs` → 18 passed
- Full CLI package: `pnpm exec vitest run packages/cli` → 117 files / 2035 tests passed
- Manual end-to-end with the real binary in a scratch consumer project: dry-run report → file untouched; apply → block rewritten to `Astryx v0.0.15 …`, notes above/below preserved; missing-block and config-present nudge cases; `--json` envelope verified.

## Follow-up (out of scope here)

josephfarina's third comment on #4168 floats promoting `astryx upgrade` at version-change time (a core `postinstall` version-stamp comparison, or writing a pending-upgrade notice into the managed block itself). Deliberately not attempted in this PR — it spans the core package and deserves its own design pass.

## Reviewer questions

1. When the never-initialized nudge fires, the existing layer-3 preAction nudge ("Next step: run `astryx init` …", stderr) also fires, so human mode shows two init pointers. Worth de-duplicating (e.g. skip the in-band nudge when layer 3 already printed), or is the redundancy acceptable since layer 3 is stderr-only and suppressed in `--json`?
2. Staleness is defined as "block text differs from what we'd generate today", which also flags drift within the same version (component count, styling-system guidance, invocation stem). Happy to narrow it to a version-stamp comparison if you'd rather the block only refresh across version bumps.
3. Dry-run previously wrote the agent docs (comment said "also update during dry-run"); this PR makes dry-run report-only per the issue's expected behavior. Flagging it as an intentional behavior change in case anything relied on it.
4. The `no_codemods` / `up_to_date` outro now says `Dry run complete` when a stale block is pending (instead of `Done`) — shout if you'd rather keep the outro stable.

Fixes #4168
Fixes #4169

### DIFF
--- a/.changeset/upgrade-refresh-agent-docs.md
+++ b/.changeset/upgrade-refresh-agent-docs.md
@@ -1,0 +1,13 @@
+---
+'@astryxdesign/cli': patch
+---
+
+[feat] `astryx upgrade` now keeps the managed agent-docs block honest after a version bump (#4168, #4169). One detection pass audits every managed `<!-- ASTRYX:START -->` block — the canonical agent-doc paths plus a root-level marker scan for custom `--agent-docs-path` targets — against what the installed version generates today:
+
+- `--apply` rewrites stale blocks in place for the installed version, touching only the content between the markers; legacy `<!-- XDS:START -->` blocks migrate to the current markers.
+- Dry-run reports each stale block as a pending change alongside the codemods (`Agent docs block in AGENTS.md is at v0.1.2 (installed: v0.1.7)`) instead of writing during a dry run, which it previously did.
+- The pass also runs on the early "already up to date" and "no codemods available" returns, so `upgrade` never reports that nothing needs to change while the block is stale.
+- Never-initialized nudge: core installed but no managed block anywhere and no `astryx.config` prints "No agent docs found — run `astryx init` …".
+- `--json` status envelopes and the upgrade receipt carry a structured `agentDocs` result (`installedVersion`, `refreshed`, `stale`, `missing`).
+
+@jiunshinn
--- a/packages/cli/src/commands/agent-docs.mjs
+++ b/packages/cli/src/commands/agent-docs.mjs
@@ -16,6 +16,10 @@
  *
  * --agent <tool>: target a specific tool preset (claude, cursor, codex, hermes, all)
  * --agent-docs-path <path>: explicit file path(s)
+ *
+ * Also exposes auditAgentDocs() — the staleness detector `upgrade` uses to
+ * compare every managed block against what the installed version generates
+ * today (#4168).
  */
 
 import * as fs from 'node:fs';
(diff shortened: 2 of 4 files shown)


josephfarina:
Hi sorry those issues are things i am actively working on and are just for tracking. I appreciate it though :)

PR #4225, opened 2026-07-23
feat(Avatar): optional href/onClick interactivity + roving focus in AvatarGroup (#4170)

### LINKED ISSUE #4170: Avatar: optional interactivity (href/onClick) + single-tab-stop roving focus in AvatarGroup
## Summary

Make Avatar **optionally interactive** — support `href` and `onClick` so an avatar can be a link (to a profile) or a button (open a menu/dialog). When interactive avatars are inside an `AvatarGroup`, the group should be a **single tab stop** with roving arrow-key navigation via `useListFocus` (with a keyboard hint), rather than N separate tab stops.

## Motivation

Today Avatar renders a static `<div role="img">` — no way to make it clickable without a consumer wrapping it in their own `<a>`/`<button>`, which loses the component's styling, focus ring, and a11y wiring. Profile avatars that navigate, or avatars that open an actions menu, are extremely common. Avatar should own this the same way `Button` does.

And when a row of avatars is interactive, tabbing through every single one is poor keyboard UX. Composite widgets should be one tab stop with arrow-key navigation — the pattern `useListFocus({hasRovingTabIndex: true})` already implements.

### Avatar (interactive mode)
- **`href` + `as`** → renders as a link, using `useLinkComponent()` for the app's router (same precedent as `Button`: `href`, `as: LinkComponentType`, `target`, `rel`).
- **`onClick`** → renders as a `<button>` when no `href`.
- **Neither** → stays the current static `<div role="img">` (no behavior change, no focusability).
- Interactive avatars get the correct role, keyboard activation (Enter/Space for button, Enter for link), a visible `:focus-visible` ring, and hover/active affordances. `name`/`alt` remains the accessible name (coordinate with the default-tooltip work in #4164).
- Follows the same "component wins on contract props, compose consumer handlers via `composeEventHandlers`" forwarding rules as the rest of core.

### AvatarGroupOverflow (interactive parity)
- The "+N" indicator already renders as a `<button>` when `onClick` is passed, but has no link support. Give it the **same interactive contract as Avatar**: add `href` + `as` (via `useLinkComponent`) so "+N" can navigate (e.g. to a full member list), keep `onClick` for the button case, and stay a static `<span>` when neither is provided.
- Same role/keyboard/focus-ring/activation treatment as an interactive Avatar, and it participates in the group's roving tab order (below).

### AvatarGroup (single tab stop)
- When its avatars are interactive, AvatarGroup uses `useListFocus` with `hasRovingTabIndex: true` so the group is **one tab stop**: one avatar is `tabindex="0"`, the rest `-1`, and Left/Right (horizontal orientation) move focus, with Home/End for boundaries. The hook already stamps/repairs the roving tab stop as items mount/unmount.
- The `AvatarGroupOverflow` "+N" indicator participates in the roving order when it's interactive (`onClick` or `href`).
- **Keyboard hint:** surface a hint that the group is arrow-navigable (e.g. an `aria-description`/visually-hidden instruction, or a documented pattern) so screen-reader and keyboard users know to use arrows. (Exact hint mechanism is a spec question — see below.)
- Correct container semantics (the group already has `role="group"` + label; may need to move toward a listbox/toolbar-style role depending on whether avatars are links vs buttons).

## Design notes / open questions (for spec)

- **Reuse, don't reinvent.** `Button` already solves link-vs-button rendering (`href`/`as`/`useLinkComponent`, `aria-disabled` focus handling) and `useListFocus` already solves the single-tab-stop roving pattern (`hasRovingTabIndex`, orientation, Home/End, RTL). This should compose those, not duplicate them. Consider whether the interactive shell should come from the shared `Clickable` primitive proposed in #3704.
- **Role choice in a group:** a group of link-avatars vs button-avatars implies different container semantics (navigation list vs toolbar). Spec should pin this to a WAI-ARIA pattern.
- **Hint delivery:** decide between a visually-hidden instruction, `aria-description`, or relying on the roving pattern alone. Don't double-announce.
- **Non-interactive avatars in a mixed group** — define behavior when only some avatars are interactive.

## Contract additions (relevant to #4163)
- Avatar: `href?`, `as?: LinkComponentType`, `target?`, `rel?`, `onClick?` — new public props to spec + document, mirroring Button's link props.
- AvatarGroup: any prop needed to opt into roving focus (or infer it from interactive children) + the hint.
- AvatarGroupOverflow: add `href?`, `as?: LinkComponentType`, `target?`, `rel?` (it already has `onClick?`) — mirror Avatar's link props.

## Acceptance criteria

- [ ] `<Avatar href="/u/jane" name="Jane" />` renders a link (via `useLinkComponent`), keyboard-activatable, with a focus ring.
- [ ] `<Avatar onClick={…} name="Jane" />` renders a focusable button with Enter/Space activation.
- [ ] `<Avatar name="Jane" />` (neither) is unchanged — static `role="img"`, not focusable.
- [ ] An `AvatarGroup` of interactive avatars is a **single tab stop**; Left/Right move focus between them; Home/End jump to ends.
- [ ] `AvatarGroupOverflow` supports `href` (renders a link via `useLinkComponent`) as well as `onClick`, and stays a static `<span>` when neither is set.
- [ ] Interactive `AvatarGroupOverflow` (link or button) participates in the roving order.
- [ ] A keyboard/AT hint communicates arrow-navigation; accessible name is not double-announced.
- [ ] Non-interactive avatars unaffected; mixed groups behave per spec.
- [ ] Works across themes + light/dark; focus ring visible in both.

## Notes
- New API surface + a11y pattern → spec protocol first (Component Lifecycle).
- Avatar family: #4164 (default tooltip), #4165 (IconButton tooltip), #4167 (overflow pill). Tooltip + interactivity interact — an interactive avatar's tooltip suppression rules (#4164) should be consistent here.

cixzhang:
## What

Adds **optional interactivity** to `Avatar` and **single-tab-stop roving focus** to `AvatarGroup`.

**Avatar** — a `href`/`onClick` element-swap that mirrors `Button` exactly:

- `href` set → renders a link via the `useLinkComponent(as)` hook (respects `LinkProvider`; never a hardcoded `<a>`). `as`/`target`/`rel` are forwarded.
- else `onClick` set → renders a `<button type="button">`.
- else → today's static element, unchanged (`role="img"`, or `role="presentation"` + `aria-hidden` when decorative). **Non-breaking default.**

Prop names are copied verbatim from `Button` (`href`/`as`/`target`/`rel`/`onClick`) — no `interactive`/`clickable` mode prop. Interactive avatars get the system focus-visible accent ring and a real accessible name (`alt`/`name`).

**AvatarGroup** — when it contains interactive avatars (or an interactive `AvatarGroupOverflow`), it becomes a single Tab stop with roving ArrowLeft/ArrowRight focus, and announces a screen-reader keyboard hint. A purely static facepile is untouched.

## Why

Avatars are frequently linked to profiles or used as buttons, and consumers have been wrapping them in their own `<a>`/`<button>` — which drops the focus ring, the accessible name, and (in a group) tab-management. Baking the exact `Button` trichotomy in keeps the API familiar and the a11y correct.

## How

- **Reuses the `Button` trichotomy.** Same resolution order, same prop names, same `useLinkComponent` hook for router-aware links. No new mode prop — the presence of `href`/`onClick` drives the rendered element.
- **Single-tab-stop roving via the existing `useListFocus({hasRovingTabIndex, orientation: 'horizontal', isRtl, itemSelector})`.** The hook already owns the roving tabindex (one `tabindex=0`, the rest `-1`) and is the same primitive `ButtonGroup`/`SegmentedControl`/`Toolbar` use — no new mode was added.
- **`data-avatar-item` marker as the `itemSelector`.** Roving selects on this marker (stamped on the rendered interactive `<a>`/`<button>` and on `AvatarGroupOverflow`'s button), **not** a tag/role selector — so a nested `<button>` inside a custom status/badge slot never joins the roving order.
- **Keyboard hint** is a visually-hidden instruction linked via `aria-describedby` on the `role="group"` root, sourced from a new i18n key (`@astryx.avatarGroup.keyboardHint`). It is only rendered/attached when the group actually has interactive children — no hint on a static facepile.
- **RTL** is derived from the rendered container the way the codebase reads direction elsewhere (`getComputedStyle(el).direction === 'rtl'`), then passed to `useListFocus`.
- **Dev warning (client-safe).** An interactive avatar with no accessible name emits a plain `console.warn` — matching the established client-safe warning pattern in sibling components (Field/Timestamp/Popover). It is **not** gated on `process.env` (not available on the client in this codebase).

## Deferred

- Async `clickAction` (Button's async prop) is intentionally **not** added here — it's a follow-up.

## Testing

- Avatar: renders link (`href`) / button (`onClick`) / static (neither); `href` wins over `onClick`; focus + ring on interactive; `data-avatar-item` only on interactive; dev-warn on interactive-without-name; no warn for static-decorative.
- AvatarGroup: single tab stop; arrow roving over interactive avatars + the overflow button; roving skips a non-avatar button in a status slot; static group has no tab stop and no hint; the `aria-describedby` hint appears only with interactive children.
- All pre-existing Avatar/AvatarGroup tests stay green.

Closes #4170

### DIFF
--- a/.changeset/avatar-interactive-roving.md
+++ b/.changeset/avatar-interactive-roving.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] `Avatar` gains optional interactivity via `href`/`onClick` (with `as`/`target`/`rel`), following Button's element-swap trichotomy: `href` renders a link through `useLinkComponent`, `onClick` (no href) renders a `<button type="button">`, and with neither the avatar stays the static, non-focusable element it is today (non-breaking default). Interactive avatars get the focus-visible accent ring and a required accessible name (from `alt`/`name`). Inside `AvatarGroup`, interactive avatars — and an interactive `AvatarGroupOverflow` — now share a single Tab stop with roving ArrowLeft/ArrowRight focus, and the group exposes a screen-reader keyboard hint via `aria-describedby`. A purely static facepile is unchanged. (#4170)
+
+@cixzhang
--- a/apps/storybook/stories/Avatar.stories.tsx
+++ b/apps/storybook/stories/Avatar.stories.tsx
@@ -682,3 +682,40 @@ export const TooltipDisabled: Story = {
     </div>
   ),
 };
+
+/**
+ * Interactive avatars. `href` renders an `<a>` (native right-click / Cmd+click);
+ * `onClick` renders a `<button>`. Both get a focus-visible accent ring. A
+ * meaningful `name`/`alt` is required so the control has an accessible name.
+ */
+export const Interactive: Story = {
+  render: () => (
+    <div {...stylex.props(styles.storyWrapper)}>
+      <div>
+        <h4 {...stylex.props(styles.heading)}>As a link (href)</h4>
+        <div {...stylex.props(styles.row)}>
+          <Avatar name="Ada Lovelace" href="https://example.com/ada" />
+          <Avatar
+            name="Grace Hopper"
+            src="https://i.pravatar.cc/150?img=5"
+            href="https://example.com/grace"
+          />
+        </div>
+      </div>
+      <div>
+        <h4 {...stylex.props(styles.heading)}>As a button (onClick)</h4>
+        <div {...stylex.props(styles.row)}>
+          <Avatar
+            name="Alan Turing"
+            onClick={() => alert('Open Alan Turing')}
+          />
+          <Avatar
+            name="Katherine Johnson"
+            src="https://i.pravatar.cc/150?img=9"
+            onClick={() => alert('Open Katherine Johnson')}
+          />
+        </div>
+      </div>
+    </div>
+  ),
+};
--- a/apps/storybook/stories/AvatarGroup.stories.tsx
+++ b/apps/storybook/stories/AvatarGroup.stories.tsx
@@ -250,3 +250,51 @@ export const ManyAvatars: Story = {
     );
   },
 };
+
+/**
+ * Interactive avatars — a mix of links (`href`) and buttons (`onClick`) plus an
+ * interactive overflow. The whole group is a single Tab stop: Tab into it once,
+ * then use ArrowLeft/ArrowRight to move focus between avatars and the overflow
+ * button. Screen readers hear a keyboard hint from the group.
+ */
+export const Interactive: Story = {
+  render: () => (
+    <AvatarGroup size="lg" aria-label="Project team">
+      <Avatar
+        src={USERS[0].src}
+        name={USERS[0].name}
+        href="https://example.com/users/alice"
+      />
+      <Avatar
+        src={USERS[1].src}
+        name={USERS[1].name}
+        href="https://example.com/users/bob"
+      />
+      <Avatar
+        src={USERS[2].src}
+        name={USERS[2].name}
+        onClick={() => alert(`Open ${USERS[2].name}`)}
+      />
+      <AvatarGroupOverflow
+        count={USERS.length - 3}
+        onClick={() => alert('Show all members')}
+      />
+    </AvatarGroup>
+  ),
+};
+
+/**
+ * Static facepile (no href/onClick) — unchanged behavior. Not focusable, no Tab
+ * stop, no keyboard hint. Shown here alongside the interactive variant for
+ * contrast.
+ */
+export const StaticFacepile: Story = {
+  render: () => (
+    <AvatarGroup size="lg">
+      {USERS.slice(0, 4).map(u => (
+        <Avatar key={u.key} src={u.src} name={u.name} />
+      ))}
+      <AvatarGroupOverflow count={USERS.length - 4} />
+    </AvatarGroup>
+  ),
+};
--- a/packages/cli/templates/blocks/components/Avatar/AvatarInteractive.doc.mjs
+++ b/packages/cli/templates/blocks/components/Avatar/AvatarInteractive.doc.mjs
@@ -1,0 +1,13 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+/** @type {import('../../../../../core/src/docs-types').TemplateDoc} */
+export const doc = {
+  type: 'block',
+  exampleFor: 'Avatar',
+  name: 'Avatar — Interactive',
+  displayName: 'Avatar — Interactive',
+  description: 'Make an avatar interactive by passing `href` to render it as a link or `onClick` to render it as a button. Both forms are focusable and show a focus-visible ring for keyboard users. Use for profile links, mention pop-ups, or any avatar people can act on.',
+  isReady: true,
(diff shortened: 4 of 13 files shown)


cixzhang:
### Self-review — Quick Self-Review checklist (Build Protocol)

1. **Icons from registry** — no icons added; the existing `DefaultIcon` fallback is unchanged. ✅
2. **themeProps present** — every root variant (link / button / static) goes through the same `mergeProps(themeProps('avatar', {size}), stylex.props(...), className, style)`, so the interactive `<a>`/`<button>` carries `astryx-avatar` for theming. AvatarGroup keeps `themeProps('avatar-group', {size})`. ✅
3. **No wrapper divs** — the interactivity lives *on* the root element itself (the `<a>`/`<button>` replaces the root `<div>` — no extra wrapper). Handlers/refs are on that element. The inner visual `<div>` structure is unchanged from before. ✅
4. **No CSS shorthands** — the new `interactive` StyleX uses explicit `outlineWidth`/`outlineStyle`/`outlineColor`/`outlineOffset` and explicit border-reset longhands, not shorthands. ✅
5. **ARIA still correct** — static default is byte-for-byte unchanged (`role="img"` / `presentation`+`aria-hidden`). Interactive elements are real `<a>`/`<button>` with an `aria-label` accessible name. Group gains `aria-describedby` → visually-hidden i18n hint, only when interactive children exist. Roving uses the shared `useListFocus` primitive (same as ButtonGroup/SegmentedControl/Toolbar). ✅
6. **Tests pass locally** — Avatar + AvatarGroup suites: 42 tests green (23 pre-existing + 19 new). ✅
7. **Sandbox/stories updated** — added an `Interactive` story to both Avatar and AvatarGroup; static facepile story added for contrast; no props removed. ✅
8. **Popover background** — N/A (no popover/layer introduced; the SR hint is a `VisuallyHidden`, not a layer).

**Spec compliance notes**
- Element-swap trichotomy + verbatim Button prop names (`href`/`as`/`target`/`rel`/`onClick`), `useLinkComponent` for links, no mode prop. ✅
- Roving is default-on via `useListFocus({hasRovingTabIndex:true, orientation:'horizontal', isRtl, itemSelector:'[data-avatar-item]'})` — reused, no new hook mode. ✅
- `data-avatar-item` marker (not tag/role) so roving skips nested buttons in a status/badge slot — covered by a test. ✅
- Keyboard hint via `aria-describedby` from a new i18n key `@astryx.avatarGroup.keyboardHint`, attached only when interactive children exist. ✅
- Async `clickAction` deferred (not added). ✅
- **Client-safe dev-warn:** plain `console.warn`, matching the established pattern in Field/Timestamp/Popover — **not** gated on `process.env` (not available on the client here). ✅

**Deviation (justified):** changeset bump is `patch`, not `minor`. Core is `0.1.7` and the repo's `check:changesets` gate enforces **pre-1.0 patch-only** (a `minor` fails CI). The `[feat]` category tag captures the new-feature nature independently of semver.

PR #4228, opened 2026-07-23
feat(TreeList): customizable expand/collapse indicator icon

### LINKED ISSUE #4131: TreeList: allow customizing the expand/collapse indicator icon (e.g. folder open/closed for file trees)
## Summary

Allow customizing the expand/collapse indicator icon on `TreeList` rows. Today the indicator is hardcoded to a rotating `chevronRight`, so patterns like a **file/folder tree** — where an expandable node should show a closed-folder icon when collapsed and an open-folder icon when expanded — aren't expressible.

## Motivation

- A folder tree is one of the most common TreeList use cases, and the folder-open / folder-closed icon swap is the expected visual language for it.
- More generally, teams want the expand affordance to match their domain (folders, categories, org nodes) rather than a generic chevron.
- `startContent` already lets you put an icon *before the label*, but it doesn't control the **expand/collapse indicator** itself, which is a separate, hardcoded element.

## Current behavior

In `TreeListItem.tsx` the indicator is fixed:

```ts
const chevronIcon = (
  <span {...stylex.props(styles.chevronSvg, isExpanded ? styles.chevronExpanded : styles.chevronCollapsed)}>
    {getIcon('chevronRight')}
  </span>
);
```

There is no prop to override it, and the collapsed/expanded states are expressed only by rotating the same chevron.

## Proposed API

Add a way to supply a custom indicator, ideally state-aware so collapsed vs. expanded can be distinct icons (not just a rotation). A render-prop keeps it flexible and avoids a rotation assumption:

```tsx
<TreeList
  renderExpandIcon={({ isExpanded, hasChildren, isDisabled }) =>
    hasChildren
      ? (isExpanded ? <Icon icon="folderOpen" /> : <Icon icon="folder" />)
      : <Icon icon="file" />
  }
/>
```

Open to a simpler shape if preferred (e.g. `expandIcon` / `collapseIcon` props). Key requirements:

- Distinguish **collapsed vs expanded** icons (folder-closed / folder-open), not just a rotated single icon.
- Optionally distinguish **leaf** nodes (no children) so a folder tree can render a file/leaf icon in the indicator slot.
- Default stays the current chevron — this is opt-in.
- Preserve the existing a11y wiring: the toggle button keeps `aria-expanded`, `aria-label`, roving-tabindex behavior, and `data-tree-toggle` regardless of the icon rendered inside it.

## Notes

- Should the rotation animation be suppressed automatically when a custom state-aware icon is provided (so an open-folder icon isn't also rotated)? Probably yes.
- Per-item override vs. tree-level render prop — a tree-level `renderExpandIcon` covers the folder case cleanly; per-item could come later if needed.

## Environment

- Component: `TreeList` (`@astryxdesign/core`), `packages/core/src/TreeList/TreeListItem.tsx`

AKnassa:
Closes #4131

## What this does

Lets a `TreeList` choose its own expand/collapse icons instead of the built-in rotating chevron — so a file tree can show a closed folder that becomes an open folder, with file icons on the rows that can't expand.

## How

One new tree-level prop, with the shape proposed in the issue:

```tsx
<TreeList
  items={items}
  renderExpandIcon={({isExpanded, hasChildren}) =>
    hasChildren
      ? <Icon icon={isExpanded ? FolderOpenIcon : FolderIcon} />
      : <Icon icon={DocumentIcon} />
  }
/>
```

- Called per item with `{isExpanded, hasChildren, isDisabled}` (exported as `TreeListExpandIconState`). Leaves get the call too, so file trees can fill the indicator slot — rows with a leaf icon align to the indicator column.
- Custom icons swap per state; the chevron's rotation animation is suppressed for them, per the issue's note (an open folder shouldn't also rotate).
- Returning nothing renderable falls back safely: default chevron for parents, no indicator (and the usual alignment offset) for leaves.
- The toggle's accessibility wiring is untouched regardless of the icon: `aria-expanded`, localized `aria-label`, roving tabindex, `data-tree-toggle`.
- Without the prop, rendering is unchanged.

Per-item overrides stay out, as the issue suggests ("could come later").

## How to see it

Storybook → Core/TreeList → **Custom Expand Icons** 
custom-icons-initial-dark
custom-icons-initial-light
custom-icons-public-expanded
default-chevron-basic-dark
default-chevron-basic-light

## Proof

- 23 new tests (TreeList suite 48 → 71): red-first for the feature, plus edge cases — disabled parents, `children: []`, `isExpanded` on leaves, Enter routing on clickable parents, string/empty returns, deep nesting, render-prop swap on re-render, ArrowLeft collapse. Six deliberate mutations each fail exactly one test.
- Full suite green (core+lab 5090, cli 2178); core/docs/storybook typecheck and lint clean.
- Docs updated (en/zh/dense) and rendering via `astryx component TreeList --dense`; changeset included.


packages/core/src/TreeList/TreeListItem.tsx
```diff
@@ -211,6 +211,25 @@ const styles = stylex.create({
...
     transform: 'rotate(0deg)',
   },
+  // Custom icons from renderExpandIcon swap per state instead of rotating —
+  // no transform, no transition.
+  customExpandIcon: {
+    display: 'flex',
+  },
+  // Indicator slot for leaf items when renderExpandIcon supplies a leaf icon.
+  // Same metrics as the chevron toggle so leaves align, but non-interactive.
+  leafIconContainer: {
```
cixzhang: This sets the token box but an `<Icon>`'s own rem width wins. Glyph still overflows.

packages/core/src/TreeList/TreeList.tsx
```diff
@@ -56,6 +63,33 @@ export interface TreeListProps extends BaseProps<HTMLDivElement> {
...
+   * wiring (aria-expanded, aria-label, roving tabindex) regardless of the
+   * icon rendered inside it.
+   *
+   * @example
+   * ```
+   * <TreeList
+   *   items={items}
+   *   renderExpandIcon={({isExpanded, hasChildren}) =>
+   *     hasChildren
+   *       ? (isExpanded ? <Icon icon={FolderOpenIcon} /> : <Icon icon={FolderIcon} />)
```
cixzhang: This example is what people will copy. Might want a size on it once the column's settled.

cixzhang (changes requested):
Thanks for this — the state-aware swap and the leaf slot are exactly what the issue asked for, and the tests are thorough.

One thing before it lands. The indicator column sizes its glyph with `fontSize: --spacing-4`, which works because the default chevron is a 1em registry icon. An `<Icon>` carries its own rem box and ignores it, so the folder example draws 20px in the 16px column — and at the browser's large-text setting, 25px, with the label's gap gone. `size="sm"` only lines up at a 16px root.

It'll also need a merge with main — TreeList's chevron block moved since July.

Ideally I'd like the column and the glyph on one scale. Would making the column `1rem`, so `size="sm"` tracks it, work — or would you rather the indicator size whatever it's handed?

[Reviewed by Robohands]

cixzhang:
Checking in on TreeList indicators: the custom glyph and its column still need one size contract after merging main. Would you like to continue, have us carry it, or close this for now?

AKnassa:
Continuing. Round pushed at 7eb427eab:

- The indicator column is now `1rem`, the `<Icon size="sm">` box: the toggle, the leaf slot and the default chevron share one `indicatorColumn` style, and the chevron's `--spacing-4` re-pin is gone. The null-leaf offset uses the same length, so labels stay aligned as the root font size changes.
- Measured in Chrome on the story: column and glyph 16/16 at a 16px root, 20/20 at a 20px root, glyph overflow 0, label gap 4px at both.
- "Size whatever it's handed" was not reachable: the column cannot resize a consumer's `<Icon>` without cloneElement, and a stretch-to-fit column breaks the null-leaf offset.
- Merged main (`84d6ccb66`); the four alignment tests now read `--_tree-indent`. Docs, story and changeset say `size="sm"`.
- Left alone: `--tree-list-indent` stays a px token beside the rem column. At a 20px root that is a 16px step next to a 20px column, which still reads fine, and it is the existing lever.

packages/core/src/TreeList/TreeListItem.tsx
AKnassa: Now the shared `indicatorColumn` box (1rem). An `sm` Icon fills it at any root font size.

packages/core/src/TreeList/TreeList.tsx
AKnassa: The example, the story and the docs now pass `size="sm"`, and a guidance line in `TreeList.doc.mjs` spells out the 1rem column.

PR #4229, opened 2026-07-23
feat(vibe-tests): ground the a11y score in runtime axe results (#4145)

### LINKED ISSUE #4145: Vibe test: Accessibility dimension is a false positive — scores ~100 by construction, invisible to component-level a11y quality
## Summary

The **Accessibility** dimension in the vibe-test universal evaluator (`internal/vibe-tests/src/universal-eval.ts`, `analyzeAccessibility()`) reports near-100 scores for Astryx targets regardless of the actual accessibility of the underlying components. The score is structurally guaranteed to be high for any component-composed output, so it does not reflect real accessibility quality and did not move at all across the recent APG/a11y hardening releases. It should be treated as the least trustworthy of the five dimensions until it's hardened or relabeled.

## What the dimension measures today

`analyzeAccessibility()` is a **static regex scan of the generated consumer code**. It starts at 100 and only subtracts for five textual anti-patterns:

| Rule | Severity | Penalty |
|---|---|---|
| `onClick` on a raw `span/div/p/td/tr/li/img/svg` without `role`/`tabIndex` | critical | −15 |
| icon-only `<button>` with no `aria-label` / text child | critical | −15 |
| raw `<input>` with no nearby `label`/`aria-label` | critical | −15 |
| `` with no `alt=` | moderate | −8 |
| heading level skip (e.g. h2 → h4) | minor | −3 |

There is no runtime, no rendered DOM, no axe-core, no APG assertions.

## Why this is a false positive for Astryx

Every one of those five rules can **only fire on raw HTML**. When the agent composes from Astryx primitives (`<Button>`, `<Field>`, `<TextInput label="…">`, etc.), the raw `<input>` / `` / `onClick`-on-div never appears in the consumer code — it's encapsulated inside the component. So none of the penalties can trigger and the score stays pinned at ~100.

This means the dimension actually measures **"did the agent avoid raw-HTML footguns?"** (composition hygiene), not **"is the result accessible?"** For a component library that pushes a11y into the components by design, this is high **by construction** — and it was high before the hardening sprint too.

## What it structurally cannot see

Everything the recent releases fixed lives at runtime, inside the components, invisible to a text scan of consumer code:

- Focus trapping / focus restoration (Dialog, Drawer, overlays)
- Keyboard nav: arrow keys, typeahead, Escape, roving tabindex (menus, comboboxes, trees)
- `aria-controls` / `aria-expanded` / `aria-current` correctness and wiring
- Live-region announcements (`useAnnounce`), `VisuallyHidden`
- Color contrast, reduced-motion, screen-reader behavior

The APG-compliance work that consumed 50+ commits is exactly the class of thing a regex over `<Button>` can never detect. So "vibe a11y ≈ 100" and "component a11y was clearly bad" are both true — they measure different objects (composition hygiene vs. runtime component quality).

## Impact

- The a11y number is a **blind spot that flatters the exact thing it can't verify**: Astryx's own architecture guarantees a high score regardless of reality.
- It gives no signal on regressions — a component silently dropping a built-in a11y guarantee costs zero points.
- Readers reasonably interpret ~100 as "the library is accessible," which the metric does not support.

## Proposed hardenings

Roughly in order of value:

**1. Add a runtime axe-core / APG assertion pass (highest value).**
We already render targets to HTML and screenshot them (`build-previews.ts`, `screenshot-previews.ts`). Run axe-core against the rendered DOM and fold violations into the a11y dimension. This is the only option that catches focus, ARIA wiring, and contrast — the stuff the hardening sprint actually fixed. Keep it target-neutral so all four targets are scored by the same runtime checks (this should also legitimately reward Astryx over raw HTML/baseline on merit rather than by construction).

**2. Wire the `a11y-manifests/` diff into the score.**
We already generate Astryx vs. Radix/shadcn manifests (`internal/vibe-tests/a11y-manifests/`), but they don't feed any number. Track built-in-guarantee coverage per component and penalize regressions when a component drops a guarantee it previously shipped. This turns the manifests from documentation into a scored signal and makes the hardening work visible.

**3. Relabel the current static dimension (do this regardless).**
Rename the existing static-scan dimension to something like **"A11y hygiene (composition)"** so nobody reads ~100 as "the library is accessible." Cheap, honest, and prevents misreading even before 1–2 land.

**4. Stop-clamping / surface the ceiling.**
Because the score can only ever go down from 100 and Astryx output rarely trips a rule, report the number of rules that were even *eligible* to fire so a perfect score on a trivial output is distinguishable from a perfect score on a complex one.

## Acceptance criteria

- The a11y dimension is either backed by runtime checks (option 1) and/or manifest-coverage deltas (option 2), **or** clearly relabeled as composition hygiene (option 3).
- The recent APG/a11y hardening produces a measurable, non-trivial score change on at least the affected components (proof the metric now tracks reality).
- Scoring stays target-neutral and fair per the checker protocol in `internal/vibe-tests/README.md` (same logic across targets; blind to which config produced the output).

## Pointers

- `internal/vibe-tests/src/universal-eval.ts` — `analyzeAccessibility()` (the five-rule scan) and the 5-dimension scoring model
- `internal/vibe-tests/a11y-manifests/` — existing Astryx vs. baseline manifests (currently unused by scoring)
- `internal/vibe-tests/src/build-previews.ts`, `screenshot-previews.ts` — existing render pipeline an axe-core pass could hook into
- `internal/vibe-tests/README.md` — checker protocol / fairness invariants

AKnassa:
Fixes #4145

The accessibility dimension was a static regex scan of consumer code, so component-composed output scores ~100 by construction. This adds a runtime basis and stops the score from overclaiming.

- **New `axe-previews.ts`** — axe-core over each built preview in headless Chromium (light + dark) → `axe-results.json` sidecar per iteration, same pattern as `build-errors.json`. Wired into `vibe-screenshots.yml` after screenshots; zero new deps (`@axe-core/playwright` was already a root devDep).
- **`universal-eval.ts` folds violations in by impact** (critical −15 / serious −10 / moderate −8 / minor −3 per rule). The four static rules axe re-verifies on the rendered DOM (image-alt, labels, button-name, heading-order) stop double-penalizing; `click-non-interactive` stays static-only — React handlers are invisible to the DOM. No sidecar → static-only scoring as before; old iterations and the stored `accessibility` key are untouched.
- **Honest labels + visible ceiling** — "A11y Hygiene (composition)" vs "Accessibility (runtime + hygiene)" in the CLI tables, compare markdown, and the report app; new `A11yMetrics` records eligible-site counts so a 100 earned on zero signal reads as "100 by construction".
- **Not done here**: scoring the `a11y-manifests/` diff (proposal 2) — the manifests use different component vocabularies and only 2 of 4 targets have one, so naive wiring breaks the Fair Evaluators invariant. Documented as a non-goal in the README.

Proof (demo iteration, two `html`-target variants with identical static profiles):

| | static-only | with axe |
|---|---|---|
| accessible variant | 100 | 100 |
| broken variant (`aria-required-attr` + `color-contrast`) | 100 | **75** |

Testing: 49 new tests including a real-Chromium integration test (auto-skipped on browserless runners); scoring rules mutation-checked; malformed sidecar/manifest entries degrade instead of crashing the aggregate. Full suite 7260/7273 — the 13 are the known pre-existing `packages/cli` load flakes, green in isolation.

Pre-existing, noticed but untouched: `analyzeMaintainability` omits `darkModeSupport` from its metrics (`darkModeRate` always computes 0), and the baseline a11y manifest is hardcoded despite its header saying it fetches Radix docs.

### DIFF
--- a/internal/vibe-tests/src/axe-previews.ts
+++ b/internal/vibe-tests/src/axe-previews.ts
@@ -1,0 +1,331 @@
+#!/usr/bin/env node
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+/**
+ * @file Runtime accessibility scan of pre-built preview HTML via axe-core
+ * @input Pre-built preview HTML files from build-previews.ts
+ * @output axe-results.json sidecar per iteration (consumed by universal-eval.ts)
+ * @position internal/vibe-tests/src/axe-previews.ts
+ *
+ * Loads each preview in headless Chromium (light + dark), runs axe-core
+ * against the live rendered DOM, and writes per-prompt violations keyed by
+ * promptId — the same sidecar pattern as build-errors.json. The next
+ * universal-aggregate run picks the sidecar up automatically and folds the
+ * violations into the accessibility dimension (issue #4145): this is what
+ * lets the score see focus, ARIA wiring, and contrast — things the static
+ * scan of consumer code structurally cannot.
+ *
+ * Usage:
+ *   tsx src/axe-previews.ts --iterations 8734233a,d4ff8c2c
+ *   tsx src/axe-previews.ts --iterations 8734233a --prompts tc-4 --themes light
+ */
+
+import * as fs from 'node:fs';
+import * as path from 'node:path';
+import {pathToFileURL} from 'node:url';
+import type {chromium as PlaywrightChromium} from 'playwright';
+import {
+  getResultsDir,
+  writeJson,
+  serveStatic,
+  enumeratePreviews,
+} from './utils.js';
+import type {
+  AxeResultForPrompt,
+  AxeResults,
+  AxeViolationRecord,
+} from './types.js';
+
+const VIEWPORT = {width: 1280, height: 800};
+const DEFAULT_THEMES = ['light', 'dark'] as const;
+
+/** One raw axe violation as reported for a single theme's scan. */
+export interface RawAxeViolation {
+  id: string;
+  impact?: string | null;
+  help: string;
+  /** Number of affected DOM nodes */
+  nodes: number;
+}
+
+/** One theme's axe scan of one preview page. */
+export interface RawAxeRun {
+  theme: string;
+  violations: RawAxeViolation[];
+  passes: number;
+  incomplete: number;
+}
+
+const IMPACT_RANK: Record<string, number> = {
+  critical: 3,
+  serious: 2,
+  moderate: 1,
+  minor: 0,
+};
+
+/** axe reports impact as nullable; unknown/missing maps to moderate. */
+function normalizeImpact(
+  impact: string | null | undefined,
+): AxeViolationRecord['impact'] {
+  return impact != null && impact in IMPACT_RANK
+    ? (impact as AxeViolationRecord['impact'])
+    : 'moderate';
+}
+
+/**
+ * Merge per-theme axe runs into one per-prompt record: violations union by
+ * rule id (max nodes, highest impact, themes attributed), passes take the
+ * strictest count across themes, incomplete the loosest.
+ */
+export function mergeAxeRuns(
+  target: string,
+  runs: RawAxeRun[],
+): AxeResultForPrompt {
+  if (runs.length === 0) {
+    return {
+      target,
+      themesScanned: [],
+      violations: [],
+      passes: 0,
+      incomplete: 0,
+    };
+  }
+
+  const byId = new Map<string, AxeViolationRecord>();
+  for (const run of runs) {
+    for (const v of run.violations) {
+      const impact = normalizeImpact(v.impact);
+      const existing = byId.get(v.id);
+      if (!existing) {
+        byId.set(v.id, {
+          id: v.id,
+          impact,
+          help: v.help,
+          nodes: v.nodes,
+          themes: [run.theme],
+        });
+      } else {
+        existing.nodes = Math.max(existing.nodes, v.nodes);
+        if (!existing.themes.includes(run.theme)) {
+          existing.themes.push(run.theme);
+        }
+        if (IMPACT_RANK[impact] > IMPACT_RANK[existing.impact]) {
+          existing.impact = impact;
+        }
+      }
+    }
+  }
+
+  return {
+    target,
+    themesScanned: runs.map(r => r.theme),
+    violations: [...byId.values()],
+    passes: Math.min(...runs.map(r => r.passes)),
+    incomplete: Math.max(...runs.map(r => r.incomplete)),
+  };
+}
+
+/**
+ * Axe-scan every preview of one iteration and write the axe-results.json
+ * sidecar. Returns the results, or null when the iteration has no previews.
+ * When a prompt has previews for several targets, the iteration's configured
+ * target wins — that's the code universal-aggregate evaluates.
+ */
+export async function scanIteration(opts: {
+  resultsDir: string;
+  iterationId: string;
+  prompts?: string[];
+  themes?: readonly string[];
+}): Promise<AxeResults | null> {
+  const {resultsDir, iterationId, prompts} = opts;
+  const themes = opts.themes ?? DEFAULT_THEMES;
+  const iterDir = path.join(resultsDir, iterationId);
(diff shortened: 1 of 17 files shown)


humbertovirtudes (commented):
Really nice closing of the gap #4145 called out. A static scan of consumer code cannot see focus order, ARIA wiring, or contrast, so grounding the accessibility dimension in a real axe-core pass over the rendered DOM is the right instinct. A few things I appreciated on the read:

- `mergeAxeRuns` unioning per-theme violations by rule id (max nodes, highest impact, themes attributed) is the correct shape, and taking the strictest passes / loosest incomplete across light and dark is a sensible conservative default.
- `AXE_COVERED_STATIC_RULES` keeping a defect that axe already scores on the rendered DOM from being double-penalized by the static rule is exactly right. The note on why `click-non-interactive` is deliberately NOT in that set (React attaches handlers synthetically, so there is no onClick in the DOM for axe to see) is a great comment to leave for the next reader.
- Degrading a malformed or truncated sidecar to "no violations" instead of crashing the aggregate is the kind of robustness this harness needs.
- The honest `getA11yDimensionLabel` ("A11y Hygiene (composition)" when there is no runtime data) is a good call: it keeps the score from overclaiming.

One ask before this lands, and it is the same bar we hold other behavioral / agent-facing changes to: could you attach a before/after on a sample iteration or two showing how the accessibility scores actually move once axe grounds them? The penalty ladder (critical 15 / serious 10 / moderate 8 / minor 3) is reasonable on its face, but seeing it applied to real previews is what tells us the weights land where we want, and that a score that was previously 100 "by construction" now reflects reality.

Not blocking on my end, just want that evidence in the thread before we call the scoring calibrated. Solid contribution.

internal/vibe-tests/src/universal-eval.ts
```diff
@@ -332,8 +430,16 @@ function analyzeAccessibility(code: string): DimensionScore {
...
 
+  const staticFindings = findings.slice();
+  const runtime = axeResult != null;
+
   let score = 100;
-  for (const f of findings) {
+  for (const f of staticFindings) {
+    // Runtime axe already scores these defect classes on the rendered DOM —
+    // don't double-penalize (same principle as tsc vs phantom props above).
+    if (runtime && AXE_COVERED_STATIC_RULES.has(f.rule)) {
```
cixzhang, line 440: Hmm, this keys off the sidecar existing rather than axe having run the matching rule.

internal/vibe-tests/src/axe-previews.ts
```diff
@@ -0,0 +1,331 @@
...
+          const axe = await new AxeBuilder({page}).analyze();
+          runs.push({
+            theme,
+            violations: axe.violations.map(v => ({
+              id: v.id,
+              impact: v.impact ?? null,
+              help: v.help,
+              nodes: v.nodes.length,
+            })),
+            passes: axe.passes.length,
```
cixzhang, line 227: Keeping `axe.passes.map(p => p.id)` here would give the fold-in what it needs to suppress on evidence.

cixzhang (changes requested):
Thanks for this — the eligible-site count is the part I'll get the most out of.

One thing to fix first. `AXE_COVERED_STATIC_RULES` stops penalizing whenever a sidecar exists, but axe only sees what rendered. An `` behind a closed dialog, or a stale sidecar, means the static scan caught a real defect and nothing charges for it — 92 today, 100 after this. Same overclaim [#4145](https://github.com/facebook/astryx/issues/4145) opened on.

The rule ids are right there; `axe-previews.ts:227` keeps only the count:

```ts
passes: axe.passes.map(p => p.id),
```

Separately, the proof is a synthetic fixture and [#4145](https://github.com/facebook/astryx/issues/4145) asks for a measurable change on real data — a before/after table from a stored iteration would help. But the fix I need is the suppression: skip a static rule only when axe actually evaluated its match.

[Reviewed by Robohands]

cixzhang:
Checking in on the a11y score: static findings still need suppression only when axe actually evaluated the matching rule. Would you like to continue, have us carry it, or close this for now?

PR #4234, opened 2026-07-23
feat(Timestamp): show one instant across several time zones on hover (#4188)

### LINKED ISSUE #4188: Timestamp: support multiple time zones & formats in the tooltip (needs API design)
## Summary

`Timestamp`'s hover tooltip should be able to display the time in **multiple time zones and/or multiple formats** at once, instead of the single local-timezone string it shows today. This needs an **API design pass** before implementation — the point of this issue is to design that API.

## Current state

- `Timestamp` renders human-readable text (`relative` / `auto` / `date` / `date_time` / `time` / `system_*`) via `Intl.DateTimeFormat` in the **viewer's local timezone**.
- On hover it shows **one** tooltip line — `getFullAbsoluteString(date)` — a single full absolute date/time with a short timezone name, again in the viewer's local zone.
- `isTimezoneShown` can append a single short timezone abbreviation to the visible text.
- There is **no way** to (a) show the tooltip in a different/explicit timezone, or (b) show the same instant across several zones/formats in the tooltip.

## What we want

Let a consumer configure the tooltip to present the instant in **a set of time zones and/or formats**. Common cases:
- Show the viewer's local time **plus** UTC.
- Show local + a fixed business/event timezone (e.g. the origin timezone of the event).
- Show the same time rendered in multiple formats (e.g. a friendly `date_time` line and a precise `system_date_time` line).
- Label each line (e.g. "Local", "UTC", "Pacific") so multiple zones are distinguishable.

## API design — the actual work of this issue

Design a clean, Astryx-consistent API for tooltip timezone/format configuration. Open questions to resolve:

- **Timezone input:** IANA zone identifiers (e.g. `'America/Los_Angeles'`, `'UTC'`)? A single `tooltipTimeZone` vs. a list of zones for multi-zone display?
- **Multiple entries:** what shape expresses "several lines in the tooltip"? e.g. a `tooltipZones` / `tooltipEntries` array of `{ timeZone, format?, label? }`, vs. simpler scalar props. Prefer the smallest API that covers the real cases (orthogonal axes, no design recipes).
- **Format per entry:** reuse the existing `TimestampFormat` union for each tooltip line? Allow a different format in the tooltip than in the visible text.
- **Labeling:** how are per-zone labels provided/defaulted (auto short zone name vs. consumer-supplied label)?
- **Defaults & back-compat:** current behavior (single local full-absolute line) must remain the default. New props are additive.
- **`isTimezoneShown` interaction:** how the visible-text timezone abbreviation relates to the new tooltip config.
- **i18n/locale:** honor locale in `Intl.DateTimeFormat` for each entry.

## Acceptance criteria

- [ ] A designed, reviewed API for configuring the tooltip's time zone(s) and format(s), documented with rationale (consider vibe-testing candidate API shapes if naming/shape is contested).
- [ ] Tooltip can render the instant across multiple time zones and/or formats, with optional per-line labels.
- [ ] Default behavior unchanged (single local absolute line); new props are additive/optional.
- [ ] Uses `Intl.DateTimeFormat` with explicit `timeZone`; locale-aware.
- [ ] Tests, `.doc.mjs`, Storybook story, and showcase block cover the multi-zone/multi-format tooltip.

AKnassa:
## What this does

Hovering a timestamp can now show the same moment in several time zones and formats at once, each on its own labelled line. Today it always shows exactly one line, in the reader's own zone.

```tsx
<Timestamp
  value={savedAt}
  tooltipEntries={[
    {label: 'Local'},
    {timezoneID: 'UTC', label: 'UTC'},
  ]}
/>
```

One new prop. Each entry is one line: an optional `timezoneID` (omit it, or pass `'local'`, for the reader's own zone), an optional `format`, an optional `label`.

## Why the shape is this shape

The issue says **"needs API design"**, so this went through [API Arbitration](https://github.com/facebook/astryx/wiki/API-Arbitration) rather than my taste: four candidate APIs written independently, each given only its own docs to six naive authors who were asked for real product outcomes ("hovering should show the reader's zone and the incident's origin zone, clearly labelled"), then scored by three judges on different criteria.

| Candidate | Naive authors served cleanly | Conventions | Evidence | Implementation |
| **Entry array** (shipped) | **6/6, zero invented props** | **8** | **8.5** | 5.5 |
| Orthogonal axes | 5/6 (one silently wrong output) | 6.5 | 7.5 | 7 |
| Composition / render prop | 6/6 | 4.5 | 7 | **8.5** |
| Scalar minimal | 4/6 (two produced broken code) | 6 | 4 | 6.5 |

Two calls I made **against** the panel, so they're easy to challenge:

1. The implementation judge preferred a `renderTooltip` render prop. I didn't take it — `renderTooltip` is already taken in `useTooltip.tsx:198` with a different contract, the issue's acceptance criteria ask for per-line labels *as an API* rather than delegated to each caller, and the bare-escape-hatch shape produced keyboard-unreachable tooltips in naive hands.
2. All three judges wanted exported formatting helpers grafted in. I cut them. The evidence for helpers was a *different* candidate failing because it exported nothing — not evidence this one needs them. Worth revisiting if ragged tooltips turn up in real use.

## Nothing changes unless you opt in

Without the new prop, the tooltip content is the same single string it is today — no wrapper element, no markup change. An empty array counts as no configuration, so `hasTooltip` stays the only on/off switch.

One deliberate behaviour change worth your eyes: **configuring entries also turns the tooltip on for absolute formats**, which have never had one. Leaving that shut would let `format` silently suppress another prop's output, which the conventions wiki forbids. The cost is that those timestamps gain a tab stop and focus ring, exactly as relative ones already have — so a column of 500 of them gains 500 tab stops. Opt-in only, but it is a real tab-order change.

`'full'` (the long absolute style) lives only in the tooltip vocabulary and never touches `TimestampFormat` — so this has **zero conflicting hunks with #4199**, and the `date_long` / `date_weekday` you're adding there become valid tooltip formats for free. I confirmed the tripwire by simulating that change: adding a member to `TimestampFormat` fails the build (`TS2366`) in both the new switch and the existing `formatTimestamp`, rather than silently returning nothing.

## Two things I need a call on

- **The shape itself.** The vocabulary (`tooltipEntries`, `timezoneID`, `'local'`, `'full'`) is my reading of your open questions, not your decision. Happy to reshape.
- **The tab-order change** described above.

## Found while testing, not fixed here

`parseValue` misreads **every pre-1970 millisecond timestamp**: the heuristic `value < 1e12 ? value * 1000 : value` is applied to a signed number, so a negative millisecond value gets multiplied again. The moon landing (`-14182940000`) renders as the year **1520**. It's pre-existing, unrelated to tooltips, and changes behaviour for every `Timestamp` consumer, so it wants its own PR — say the word and I'll open one.

Fixed in passing, because it sits on the prop next door: `isTimezoneShown`'s JSDoc, `.doc.mjs` (en/zh/dense) and its showcase block all claimed it applied to `system_date_time` and `system_time`. It never has — `formatTimestamp` has no such branch, so the showcase was demonstrating a no-op.

## How to check it

Storybook → **Core/Timestamp → Tooltip — multiple time zones**: local+UTC, three labelled zones, one zone in two formats, and a UTC-only audit log. Hover or tab to each.

Worth knowing: the tooltip is capped at 300px, so the long `full` style plus a label wraps. The docs steer multi-zone tooltips to `date_time`, which fits on one line.

dark-00-all-closed
dark-01-local-utc
dark-02-three-zones-labelled
dark-03-one-zone-two-formats
dark-04-utc-only-audit
dark-99-keyboard-focus
light-00-all-closed
light-01-local-utc
light-02-three-zones-labelled
light-03-one-zone-two-formats
light-04-utc-only-audit
light-99-keyboard-focus
tokyo-viewer-local-utc

## Testing

93 Timestamp tests, up from 32 before this branch. All timezone- and locale-agnostic, because the repo pins `TZ` in neither vitest nor CI — a dev machine and a UTC runner have to agree, so exact assertions only use explicit-zone `system_*` output.

Covered: the date line (`+14` / `-11`, 25 hours apart), 45-minute offsets, midnight rendering as `00` and not `24`, both halves of a repeated DST hour, year and leap-day boundaries, empty and whitespace-padded zone ids, legacy link ids, casing, unknown ids degrading to the reader's zone instead of throwing, an unparseable value, `auto` resolving to an absolute format, and the `aria-describedby` wiring that is the only route to assistive tech on absolute formats.

Eleven mutations were run against the implementation; every one was caught by a test. Full suite **7472 pass / 0 fail** (baseline on `d762f825d` was 7414 / 0 — no pre-existing failures to excuse). Core and Storybook typecheck clean, ESLint clean, changeset valid.

Closes #4188


cixzhang (commented):
Thanks for this — the zone-resolution safety (fallback + one-warning-per-bad-id) and the "key on the requested zone, not the resolved offset" call are both right.

One structural request before this lands: **`formatLine` in `tooltipEntries.ts` duplicates the formatting that `formatTimestamp` already owns in `Timestamp.tsx`, and the two have already drifted.** A few concrete spots:

- **`date`/`date_time`/`time` option sets are forked.** `formatTimestamp` builds `date` from `SHARED_DATE_FORMAT_OPTIONS.date`; `formatLine` inlines `{year:'numeric', month:'short', day:'numeric'}`. Those are equivalent *today* but will silently diverge the next time `SHARED_DATE_FORMAT_OPTIONS.date` changes — the main display would update and the tooltip line would not.
- **`pad` and the `system_*` wall-clock assembly are re-declared** in `tooltipEntries.ts`, and `getWallClock` reimplements zone math that `plainDate.ts` already provides (`getTimeZoneParts` / `getTimezoneOffsetMS`).
- **`FULL_OPTIONS` duplicates `getFullAbsoluteString`'s option set** in `Timestamp.tsx`.
- The two switches now enumerate **different case sets** (`formatLine` omits `date_long`/`date_weekday`), so a future format added to `TimestampFormat` has to be remembered in two places.

The clean shape is to make **one shared, timezone-parameterized formatter** that both call sites use, rather than a fork:

- Extract the per-format `Intl.DateTimeFormatOptions` map + the `system_*` wall-clock string assembly into a single pure helper (living in `plainDate.ts` or a small sibling util), taking an optional `timeZone`.
- `formatTimestamp` (main render) calls it with **no** zone → host default → byte-identical output to today, so nothing regresses.
- The tooltip calls it with the **resolved** zone.

That removes the drift risk entirely and means "add a format" is a one-place change. The header comment notes formatTimestamp "must stay reachable, unchanged" — agreed, but that argues for *parameterizing the shared formatter by zone*, not forking it: the zone-less path stays exactly as it is.

Reuse across the two Timestamp surfaces should win over a self-contained leaf module here. Happy to pair on the extraction if useful.

cixzhang (changes requested):
Thanks for working on this feature! You may need to rebase, and left a comment on deduplicating the date formatter

AKnassa:
Done. Both surfaces now go through one `formatInstant(date, format, {timeZone, isTimezoneShown})` in a sibling module. The visible text passes no zone, so it keeps Intl's host default and stays on plain `Date` getters for `system_*`; each caller keeps its own rule for the zone abbreviation, which is the only thing the two genuinely disagree about. `getWallClock` is gone, `plainDate`'s `getTimeZoneParts` is exported and used instead, and a new `TimestampFormat` member now fails to compile in exactly one place.

The drift had already bitten: merging main turned it into a `TS2366`, since `date_long` and `date_weekday` landed with #4199 and `formatLine` did not enumerate them. Both are valid tooltip formats now.

I checked it does not regress. The new formatter matches the old `formatTimestamp` byte for byte across 11 instants x 8 formats x both flag values under 5 host zones, and the zoned `system_*` path matches the deleted `getWallClock` across 22 zones x 19 instants. The new tests pin the two surfaces to each other, mutation-checked.

One call I would like yours on: `date_time`'s date part is still spelled out rather than pointing at `SHARED_DATE_FORMAT_OPTIONS.date`. That bag is shared with DateInput, so aiming at it would mean a change made for date-only fields reshapes every date-and-time reading too. Happy to couple them if you would rather they moved together.

cixzhang:
> date_time's date part is still spelled out rather than pointing at SHARED_DATE_FORMAT_OPTIONS.date. That bag is shared with DateInput, so aiming at it would mean a change made for date-only fields reshapes every date-and-time reading too.

Oh good idea! I'll merge this for now but this sounds like a good follow up.

cixzhang (approved):
Thanks for updating!

cixzhang (approved):
Thanks for reworking this — collapsing formatLine and formatTimestamp into a single zone-parameterized formatInstant is exactly what I was hoping for, and I like that the zone-less render path stays byte-identical. Multi-timezone tooltips are a genuinely useful addition. Approving and merging.

One small follow-up note (not blocking): we have date/time formatting spread across a few places now (plainDate, dateParser, this), so I'm going to look at consolidating them separately — nothing you need to do here.

PR #4244, opened 2026-07-23
fix(core): align loading spinner inside TextArea corner (#4232)

### LINKED ISSUE #4232: [Bug] TextArea: isLoading spinner is misaligned — should sit inside the resizable textarea (like the status icon)
## Summary

In `TextArea`'s `isLoading` (busy) state, the loading `Spinner` is **misaligned** — it doesn't sit inside the textarea where you'd expect it. It should appear **inside the resizable textarea** (e.g. anchored to a corner of the field), consistent with how the status icon is placed.

## Steps to reproduce
1. Render a `<TextArea isLoading />` (or a controlled TextArea whose `optimisticValue !== value`, which also sets busy).
2. Observe the spinner position.

## Expected
Spinner appears inside the textarea's box (anchored to a corner, like the status icon), not disrupting layout.

## Actual
Spinner is misaligned (see screenshot) — it renders in normal flow, not positioned within the field.

## Root cause

In `TextArea.tsx` the spinner is rendered with **no positioning styles at all**:

```tsx
{isBusy && <Spinner size="sm" />}
```

Compare the status icon right below it, which IS correctly anchored:

```tsx
statusIcon: {
  position: 'absolute',
  top: spacingVars['--spacing-2'],
  insetInlineEnd: spacingVars['--spacing-2'],
  ...
}
```

The spinner has no equivalent absolute-positioning style, so it falls into the flow of the relative wrapper instead of sitting inside the textarea's corner.

## Fix direction
Give the busy spinner the same absolute-positioning treatment as `statusIcon` (anchored inside the field), reusing the existing spacing tokens. Consider: (a) the spinner and status icon can both be present — ensure they don't overlap (stack or share the same corner slot with the spinner taking precedence while busy); (b) the textarea reserves end-padding for the status icon via `textareaWithStatus` — the busy state should reserve the same space so text doesn't flow under the spinner.

## Acceptance
- Spinner is visually anchored inside the resizable textarea (corner), aligned like the status icon.
- No overlap between spinner and status icon; end-padding reserved when busy.
- Storybook story for the isLoading state; test asserting the spinner renders with the positioned class.

Geervan:
### Summary

Fixes #4232.

In `TextArea`'s `isLoading` (busy) state, the loading `Spinner` was previously rendered without positioning container styles, causing it to fall into normal flow outside the field wrapper or overlap scrollbars.

This PR updates `TextArea` to anchor the loading spinner inside the top-right corner of the field wrapper using flex alignment, matching the status icon layout without interfering with native vertical scrollbars.

---

### Changes Made

- **`packages/core/src/TextArea/TextArea.tsx`**:
  - Wrapped `Spinner` in a `spinnerIcon` `<span>` styled as a flex item (`alignSelf: 'flex-start'`, `marginInlineStart: spacingVars['--spacing-2']`).
  - Gated status icon rendering (`status && !isBusy`) so the busy spinner takes precedence without overlapping validation icons.
  - Kept `<textarea>` in flex flow (`flex: 1`) so native scrollbars remain contained inside `<textarea>` and do not overlap the spinner.
- **`apps/storybook/stories/TextArea.stories.tsx`**:
  - Added `isLoading` controls and the `IsLoading` story for visual testing.
- **`packages/core/src/TextArea/TextArea.test.tsx`**:
  - Added unit test coverage asserting `isLoading` renders the positioned spinner container and suppresses status icons.

---

### Test Plan

1. **Automated Tests**:
   - Executed `npx vitest run packages/core/src/TextArea/TextArea.test.tsx` (59/59 passed).
2. **Visual Inspection**:
   - Verified `<TextArea isLoading />` in Storybook (`http://localhost:6006/?path=/story/textarea--is-loading`).
   - Verified spinner remains anchored inside the top-right corner without overlapping vertical scrollbars when text overflows.

### DIFF
--- a/packages/core/src/TextArea/TextArea.tsx
+++ b/packages/core/src/TextArea/TextArea.tsx
@@ -27,7 +27,7 @@ import {
   type ReactNode,
 } from 'react';
 import * as stylex from '@stylexjs/stylex';
-import type {IconName} from '../Icon';
+import type { IconName } from '../Icon';
 import {
   colorVars,
   spacingVars,
@@ -41,16 +41,16 @@ import {
   inputStatusHoverShadowStyles,
   inputStatusFocusWithinStyles,
 } from '../Field';
-import {Icon, renderIconSlot, type IconType} from '../Icon';
-import {Spinner} from '../Spinner';
-import {useTooltip} from '../Tooltip';
-import {mergeProps, mergeRefs} from '../utils';
-import type {BaseProps} from '../BaseProps';
-import type {SizeValue} from '../utils/types';
-import {useInputContainer} from '../hooks/useInputContainer';
-import {useSize} from '../SizeContext/SizeContext';
-import {themeProps} from '../utils/themeProps';
-import {VisuallyHidden} from '../VisuallyHidden';
+import { Icon, renderIconSlot, type IconType } from '../Icon';
+import { Spinner } from '../Spinner';
+import { useTooltip } from '../Tooltip';
+import { mergeProps, mergeRefs } from '../utils';
+import type { BaseProps } from '../BaseProps';
+import type { SizeValue } from '../utils/types';
+import { useInputContainer } from '../hooks/useInputContainer';
+import { useSize } from '../SizeContext/SizeContext';
+import { themeProps } from '../utils/themeProps';
+import { VisuallyHidden } from '../VisuallyHidden';
 
 const COUNTER_WARNING_THRESHOLD = 0.8;
 
@@ -96,16 +96,22 @@ const styles = stylex.create({
     color: colorVars['--color-error'],
   },
   statusIcon: {
-    position: 'absolute',
-    top: spacingVars['--spacing-2'],
-    insetInlineEnd: spacingVars['--spacing-2'],
-    pointerEvents: 'none',
     display: 'flex',
+    alignItems: 'center',
+    justifyContent: 'center',
+    alignSelf: 'flex-start',
+    marginTop: spacingVars['--spacing-1'],
+    marginInlineStart: spacingVars['--spacing-2'],
+    flexShrink: 0,
   },
-  textareaWithStatus: {
-    // Reserve space so text doesn't flow under the absolutely-positioned icon.
-    // 20px (icon md) + 4px gap = 24px (--spacing-6)
-    paddingInlineEnd: spacingVars['--spacing-6'],
+  spinnerIcon: {
+    display: 'flex',
+    alignItems: 'center',
+    justifyContent: 'center',
+    alignSelf: 'flex-start',
+    marginTop: spacingVars['--spacing-1'],
+    marginInlineStart: spacingVars['--spacing-2'],
+    flexShrink: 0,
   },
 });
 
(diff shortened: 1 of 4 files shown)


packages/core/src/TextArea/TextArea.tsx
```diff
@@ -41,16 +41,16 @@ import {
...
-import {Spinner} from '../Spinner';
-import {useTooltip} from '../Tooltip';
-import {mergeProps, mergeRefs} from '../utils';
-import type {BaseProps} from '../BaseProps';
-import type {SizeValue} from '../utils/types';
-import {useInputContainer} from '../hooks/useInputContainer';
-import {useSize} from '../SizeContext/SizeContext';
-import {themeProps} from '../utils/themeProps';
-import {VisuallyHidden} from '../VisuallyHidden';
+import { Icon, renderIconSlot, type IconType } from '../Icon';
```
cixzhang, line 44: Seems like our linter/prettiers might be a bit different.

packages/core/src/TextArea/TextArea.tsx
```diff
@@ -466,19 +472,22 @@ export function TextArea({
...
             effectivelyDisabled && styles.textareaDisabled,
-            status && styles.textareaWithStatus,
           )}
         />
-        {isBusy && <Spinner size="sm" />}
-        {status && (
+        {isBusy && (
+          <span {...stylex.props(styles.spinnerIcon)}>
+            <Spinner size="sm" />
+          </span>
```
cixzhang: Thanks for this fix! It shoudl be possible to merge the stylex props directly into Spinner and avoid an extra span here and keep DOM element counts low. Let me know if you run into trouble!

cixzhang (changes requested):
Thanks for working on this! It's looking much better positioning wise. If you want to just get these changes in I'm happy to accept -- main thing is just aligning your linter/prettier to the one we're using so we can avoid the unnecessary code reformatting. I also have a comment about the extra DOM element.

Image

Here's how the text area looks. It's spinner alignment is much better (thanks!) there's an extra detail here around having the text area resizeability cover the full input rather than just the input area (may require some inner padding + measurement to achieve). This is okay for a follow up but let me know if you'd like to do it here as part of this issue resultion!

Geervan:
yep imma make the changes, thanks!

ernestt (approved):
The fix it self looks good from design side. Thanks!

Geervan:
so like should i make the changes or will this get merged? i am kinda confused sorry

Geervan:
Thanks @cixzhang! I've updated the PR to pass xstyle={styles.spinnerIcon} directly to <Spinner>, removing the extra <span> wrapper and keeping the DOM element count low. The diff has been cleaned up as well. Ready for review/merge!

Geervan:
Hello, @cixzhang I was wondering if looking at this fix is on the list or did it become obsolete, im sorry to interrupt but its been quite a while, so thought of writing this up

cixzhang:
Thanks for this, and sorry you were left waiting, that's on us. You had the placement right. [#4533](https://github.com/facebook/astryx/pull/4533) landed in July and rewrote the same area: the spinner and status icon now share one anchored slot in the field corner, so both show at once. Closing this one out.

[Reviewed by Robohands]

PR #4246, opened 2026-07-23
fix(PowerSearch): prevent edit popover from closing when selecting multi-select options via Enter

### LINKED ISSUE #4245: [Bug] PowerSearch: Enter in a multi-select filter closes the popover instead of selecting the value (when one item is already selected)
## Summary

In `PowerSearch`, when editing a multi-select (enum-list) filter value: if **one item is already selected**, pressing **Enter** to select another value **closes the PowerSearch edit popover** (saves/dismisses) instead of selecting the highlighted option in the multi-selector. Enter should commit the multi-selector's highlighted option, not close the popover, while there's an open/active selection interaction.

## Steps to reproduce
1. Open PowerSearch, add/edit a filter whose operator takes a multi-select (enum-list) value.
2. Select one value in the multi-selector.
3. Focus the multi-selector, highlight another option, press **Enter**.

## Expected
Enter selects the highlighted option in the multi-selector (adds it to the selection); the popover stays open.

## Actual
The edit popover closes (the filter is saved with the current selection) — the second value is never added. Enter is being consumed as "save the filter" rather than "select this option."

## Root cause

`PowerSearchEditPopover.tsx` puts a container-level key handler on the editor:

```tsx
// Handle Enter to save, Escape to cancel
const handleKeyDown = useCallback((e) => {
  if (e.key === 'Enter' && !isSaveDisabled) {
    e.preventDefault();
    handleSave();           // ← closes the popover
  } else if (e.key === 'Escape' ...) { ... }
}, [isSaveDisabled, handleSave, onCancel]);
...
<div {...styles.container} onKeyDown={handleKeyDown}> ... value editor ... </div>
```

`isSaveDisabled` is false as soon as a value is present, so once one item is selected, **any Enter inside the container triggers `handleSave()`**. The multi-select value editor (the `MultiSelector`/enum-list `Selector` inside) does not `stopPropagation()` on Enter, so an Enter meant to pick an option in the multi-selector bubbles to the container and saves/closes instead.

This is the classic "container Enter-to-submit vs. child listbox Enter-to-select" conflict: the child's Enter (committing a highlighted option) must not reach the submit handler.

## Fix direction
- When the multi-selector (or any child listbox/typeahead) is handling Enter to select a highlighted option, it should **consume the event** (`stopPropagation`, and it already `preventDefault`s its own default) so it doesn't reach the popover's save handler.
- Alternatively/additionally, the container `handleKeyDown` should ignore Enter when the event originated from an open listbox/combobox that is actively selecting (e.g. check `e.defaultPrevented`, or that the target is within an expanded multi-selector). The Escape branch already guards on `!e.defaultPrevented` — Enter should have an equivalent guard so a child that handled Enter isn't double-processed.
- Ensure single-select `Selector` and free-text inputs still submit-on-Enter as expected; only the case where a child actively consumed Enter (option selection) should be excluded.

## Acceptance
- With one (or more) items already selected in a PowerSearch multi-select filter, pressing Enter to select another option adds it and keeps the popover open.
- Enter still saves the filter when focus is in a plain value input (no active listbox selection).
- Escape still cancels. Test covering the multi-select Enter-selection case.

saadpocalypse:
Fixes #4245

### Summary

In `PowerSearchEditPopover`, container-level keydown handling for `Enter` previously did not check `!e.defaultPrevented`. When editing a multi-select filter (such as `enum_list`, `Tokenizer`, or `BaseTypeahead`), pressing `Enter` to confirm a highlighted option calls `e.preventDefault()`. Because `PowerSearchEditPopover` ignored `defaultPrevented`, the event bubbled up and triggered `handleSave()`, closing the popover prematurely instead of selecting the option.

This change guards the `Enter` branch in `PowerSearchEditPopover.tsx` with `!e.defaultPrevented` so option selection in child listboxes is respected, while plain input submissions and closed listboxes continue to save on `Enter`.

### Changes

- `packages/core/src/PowerSearch/PowerSearchEditPopover.tsx`: Guard `Enter` keydown handler with `!e.defaultPrevented`.
- `packages/core/src/PowerSearch/PowerSearchEditPopover.test.tsx`: Added unit test covering `Enter` key behavior with consumed child listbox events.
- `.changeset/fix-powersearch-multiselect-enter.md`: Added pre-1.0 patch changeset for `@astryxdesign/core`.

### Verification

- Unit Tests: Added test in `PowerSearchEditPopover.test.tsx`. All 145 PowerSearch tests and 7,415 workspace tests passed.
- Linting: Ran `pnpm lint` (`check:repo`, `check:sync`, `check:package-boundaries`, `check:changesets`, `check:demo-media`, `check:executable-bits`) with 0 errors.
- Build: Core package built successfully via `pnpm -F @astryxdesign/core build`.

### DIFF
--- a/.changeset/fix-powersearch-multiselect-enter.md
+++ b/.changeset/fix-powersearch-multiselect-enter.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Prevent PowerSearch edit popover from closing when selecting multi-select options via Enter (#4245)
+@saadpocalypse
--- a/packages/core/src/PowerSearch/PowerSearchEditPopover.test.tsx
+++ b/packages/core/src/PowerSearch/PowerSearchEditPopover.test.tsx
@@ -13,6 +13,8 @@ import {
 import {render, screen, fireEvent, act} from '@testing-library/react';
 import React, {useState} from 'react';
 import {PowerSearch} from './PowerSearch';
+import {PowerSearchEditPopover} from './PowerSearchEditPopover';
+import {useInternalConfig} from './useInternalConfig';
 import type {PowerSearchConfig, PowerSearchFilter} from './types';
 
 // =============================================================================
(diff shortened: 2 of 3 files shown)


cixzhang (approved):
Thanks for this fix!

PR #4253, opened 2026-07-23
docs(core): restore dropped dense bestPractices bullets for Icon and Carousel

josephfarina:
## What

The `docsDense` overlays for Icon and Carousel each dropped one English `bestPractices` bullet, so `astryx component <Name> --lang dense` silently rendered a shorter guidance list than the full English docs (check 13 count-parity). Add the compressed bullet back at the matching position and `guidance` flag:

- **Icon**: restores the "give a meaningful standalone icon an accessible name via `label`" guidance. English had 5 Do + 5 Don't; dense had 4 Do + 5 Don't.
- **Carousel**: restores the "trust the built-in navigation (swipe, Shift + wheel)" guidance. English had 4 Do + 3 Don't; dense had 3 Do + 3 Don't.

While editing Icon.doc.mjs, also replaced two em dashes in its English `label` description and `bestPractices` entry with a colon so the rendered prose follows the plain-prose convention (check 17). Bundled here because they touch the same file.

## Validation
- `node .github/scripts/api-cli-parity-test.mjs --no-baseline`: 325 pass, 0 fail
- CLI test suite: 2066 pass, 0 fail
- Rendered `--lang dense` for both: bullet counts now match English, no duplicate `(required)`

Night Watch — Doc Reviewer

### DIFF
--- a/packages/core/src/Carousel/Carousel.doc.mjs
+++ b/packages/core/src/Carousel/Carousel.doc.mjs
@@ -91,6 +91,7 @@ export const docsDense = {
       {guidance: true, description: 'Enable scroll-snap when each item should land precisely at the start edge: gallery, product list.'},
       {guidance: true, description: 'Always provide an aria-label describing what the carousel contains: "Featured products", "Team members".'},
       {guidance: true, description: 'Use consistent gap + item width so the carousel looks intentional, not like content overflowing by accident.'},
+      {guidance: true, description: 'Trust built-in navigation: trackpad users swipe horizontally, mouse users hold Shift + wheel to move through items.'},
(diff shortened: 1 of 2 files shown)


cixzhang (approved):
Verified against source: the restored Carousel bullet ("trackpad users swipe horizontally, mouse users hold Shift + wheel") matches the actual `handleWheel` implementation in `Carousel.tsx` — Shift + vertical wheel maps to horizontal scroll, and native trackpad horizontal deltas are left untouched. The Icon dense `label` bullet mirrors the existing non-dense guidance. Accurate, doc-only, no behavior change. LGTM.

PR #4254, opened 2026-07-23
docs(lab): replace em dash in RichTextEditor defaultValue description

josephfarina:
## What

`RichTextEditor`'s `defaultValue` prop description joined two clauses with an em dash (check 17, sub A1):

> Read once on mount — the editor is uncontrolled.

Replaced the em dash with a semicolon so the rendered doc prose follows the plain-prose convention. Prose only; meaning unchanged.

## Validation
- File parses as an ES module

Night Watch — Doc Reviewer


cixzhang (approved):
Verified: text-only edit to the `defaultValue` prop description (em dash → semicolon), no behavior change. Content is accurate. LGTM — thanks for the polish.

PR #4259, opened 2026-07-23
feat(lab): add LineClamp — clamp arbitrary content to N lines

### LINKED ISSUE #4180: feat(lab): LineClamp — clamp arbitrary content to N lines (port from XDSLineClamp)
## Summary

Add a **`LineClamp`** component that truncates arbitrary content to a fixed number of lines with an ellipsis, ported from the internal `XDSLineClamp` component. Port the *value and behavior*, not the internal structure.

## Why a dedicated component

Astryx already clamps **text** via `Text`'s `maxLines` prop (backed by `useTruncation`). `LineClamp` is a different primitive: a **standalone wrapper that clamps whatever you put inside it** to N lines — not just a `Text` node.

| | `Text maxLines` | `LineClamp` |
|---|---|---|
| Scope | Truncates the `Text` component's own string content | Clamps **any children** to N lines |
| Use | "This label is one `Text`, cap it at 2 lines" | "Cap this block of mixed/inline content at N lines" |
| Composition | Built into `Text` | Composable around arbitrary content |

So this is not a duplicate of `Text` — it's the reusable clamp primitive that other components and consumers can wrap around content that isn't a single `Text`.

## Proposed behavior (to confirm against the internal source during the port)

- `lines` (or `maxLines`) — number of lines to clamp to.
- Clamps via `-webkit-line-clamp` (the same mechanism `Text`/`useTruncation` already rely on), with a block/overflow setup that works cross-browser.
- Truncation **detection** should reuse the existing `useTruncation` hook rather than reinventing overflow measurement — it already handles the multi-line `Range.getBoundingClientRect()` case and shares a single `ResizeObserver`.
- Optional tooltip / "show more" affordance on overflow (mirror what the internal component does; decide during the port).
- Extends `BaseProps`, `xstyle`/`className`/`style` escape hatches, `displayName`, `'use client'`.

## Scope

- Land in **`@astryxdesign/lab`** first (canary-only), following the standard lab → core graduation path.
- Full component surface: `LineClamp.tsx`, colocated `LineClamp.test.tsx`, `LineClamp.doc.mjs`, `index.ts` export, Storybook story, showcase block.
- Reuse `useTruncation` and semantic tokens; StyleX only.

## Open questions for the port

- Prop name: `lines` vs `maxLines` (align with `Text`'s existing `maxLines` for consistency?).
- Does the internal component own the overflow-detection/tooltip behavior, or is that left to the consumer? Decide what to port vs. drop.
- Any semantic-name/API differences from the internal version get resolved to Astryx conventions (port the value, not the internal API shape).

## Acceptance criteria

- [ ] `LineClamp` clamps arbitrary children to N lines with an ellipsis.
- [ ] Reuses `useTruncation` for overflow detection (no reinvented measurement).
- [ ] StyleX only, tokens, `BaseProps` escape hatches, `displayName`, `'use client'`.
- [ ] Colocated test, `.doc.mjs`, Storybook story, showcase block, `index.ts` export.
- [ ] Lands in `@astryxdesign/lab`; documented as lab/canary.

HelloOjasMutreja:
Closes #4180

`Text`'s `maxLines` truncates a single text node; `LineClamp` is the composable version — wrap it around any children (mixed inline content, nested elements) to clamp the whole block to N lines.

- Reuses `useTruncation` (the same overflow-detection hook `Text` uses) rather than reinventing measurement
- Same optional-tooltip-on-truncation pattern as `Text`'s `hasTruncateTooltip`, via `hasTooltip`
- Ships in `@astryxdesign/lab` per the lab → core graduation path (no spec-protocol proposal needed for lab)

## Testing
- 8 new colocated tests + full `@astryxdesign/lab` suite (256/256 pass)
- `tsc --noEmit` clean
- `eslint` clean
- Manually verified in Storybook: clamping actually applies (`-webkit-line-clamp` in computed style, overflow correctly detected via `scrollHeight`/`offsetHeight`), tooltip shows full content only when truncated (no tooltip/title when content fits), mixed inline content (bold/italic/code) composes correctly inside the clamp

Opening as a draft per the contributing guide — will mark ready after final self-review.

### DIFF
--- a/packages/lab/src/LineClamp/LineClamp.tsx
+++ b/packages/lab/src/LineClamp/LineClamp.tsx
@@ -1,0 +1,142 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+'use client';
+
+/**
+ * @file LineClamp.tsx
+ * @input Uses React, stylex, useTruncation + Tooltip + BaseProps from @astryxdesign/core
+ * @output Exports LineClamp component, LineClampProps
+ * @position Lab experiment (facebook/astryx#4180); consumed by index.ts
+ *
+ * SYNC: When modified, update these files to stay in sync:
+ * - /packages/lab/src/LineClamp/LineClamp.doc.mjs (props table, features)
+ * - /packages/lab/src/LineClamp/LineClamp.test.tsx (tests for new/changed behavior)
+ * - /packages/lab/src/LineClamp/index.ts (exports if types change)
+ * - /apps/storybook/stories/LineClamp.stories.tsx (storybook stories)
+ */
+
+import {lazy, Suspense, useRef, type ReactNode} from 'react';
+import * as stylex from '@stylexjs/stylex';
+
+import type {BaseProps} from '@astryxdesign/core';
+import {useTruncation} from '@astryxdesign/core/Text';
+import type {LayerPlacement} from '@astryxdesign/core/Layer';
+import {mergeProps, mergeRefs} from '@astryxdesign/core/utils';
+import {themeProps} from '@astryxdesign/core/utils';
+
+const LazyTooltip = lazy(async () =>
+  import('@astryxdesign/core/Tooltip').then(mod => ({default: mod.Tooltip})),
+);
+
+const styles = stylex.create({
+  clamp: {
+    display: '-webkit-box',
+    WebkitBoxOrient: 'vertical',
+    overflow: 'hidden',
+    wordBreak: 'break-word',
+    overflowWrap: 'break-word',
+  },
+  tooltipContent: {
+    maxWidth: '300px',
+    wordBreak: 'break-word',
+  },
+});
+
+export interface LineClampProps extends BaseProps<HTMLElement> {
+  /** Ref forwarded to the root element */
+  ref?: React.Ref<HTMLElement>;
+  /**
+   * Content to clamp. Unlike `Text`'s `maxLines` (which truncates a single
+   * text node), `LineClamp` wraps arbitrary children — mixed inline content,
+   * nested elements, anything that isn't a single `Text`.
+   */
+  children: ReactNode;
+  /**
+   * Maximum number of lines to show before clamping with an ellipsis.
+   */
+  maxLines: number;
+  /**
+   * Show a tooltip with the full content on hover/focus when clamped.
+   * - `true` (default): show tooltip at default position
+   * - `false`: disable tooltip
+   * - Position value: show tooltip at specific position
+   * @default true
+   */
+  hasTooltip?: boolean | LayerPlacement;
+  /**
+   * HTML element to render.
+   * @default 'div'
+   */
+  as?: 'div' | 'span' | 'p';
+}
+
+/**
+ * Clamps arbitrary content to a fixed number of lines with an ellipsis.
+ *
+ * `Text`'s `maxLines` truncates the text component's own string content;
+ * `LineClamp` is the composable version — wrap it around any children (mixed
+ * inline content, nested elements) to clamp the whole block to N lines.
+ *
+ * Truncation detection reuses `useTruncation` (the same hook `Text` uses),
+ * so overflow is measured correctly even while `-webkit-line-clamp` clips
+ * `scrollHeight`.
+ *
+ * @example
+ * ```
+ * <LineClamp maxLines={2}>
+ *   Some <strong>mixed</strong> inline content that may run long.
+ * </LineClamp>
+ * <LineClamp maxLines={3} hasTooltip={false}>{longDescription}</LineClamp>
+ * ```
+ */
+export function LineClamp({
+  children,
+  maxLines,
+  hasTooltip = true,
+  as: Component = 'div',
+  xstyle,
+  className,
+  style,
+  ref,
+  ...props
+}: LineClampProps): ReactNode {
+  const truncation = useTruncation({maxLines});
+  const elementRef = useRef<HTMLElement>(null);
+
+  const tooltipPlacement: LayerPlacement =
+    typeof hasTooltip === 'string' ? hasTooltip : 'above';
+  const tooltipEnabled =
+    hasTooltip !== false && maxLines > 0 && truncation.isTruncated;
+
+  return (
+    <>
+      <Component
+        ref={mergeRefs(ref, truncation.ref, elementRef)}
+        {...mergeProps(
+          themeProps('line-clamp'),
+          stylex.props(styles.clamp, xstyle),
+          className,
+          {...style, WebkitLineClamp: maxLines},
+        )}
+        title={tooltipEnabled ? truncation.fullText : undefined}
+        {...props}>
+        {children}
+      </Component>
+      {tooltipEnabled && (
+        <Suspense fallback={null}>
+          <LazyTooltip
+            anchorRef={elementRef}
+            content={
+              <span {...stylex.props(styles.tooltipContent)}>
+                {truncation.fullText}
+              </span>
+            }
+            placement={tooltipPlacement}
+          />
+        </Suspense>
+      )}
+    </>
+  );
+}
+
+LineClamp.displayName = 'LineClamp';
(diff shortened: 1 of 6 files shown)


HelloOjasMutreja:
Visual before/after, verified against the live Storybook story:

LineClamp before and after

Text, font, size, color, and line height come directly from the running story (Lab/LineClamp, Default), not typed by hand. The tooltip on hover/focus is a real feature of the component, shown here as a static capture; try it live in Storybook to see it appear on hover.

HelloOjasMutreja:
Rebased onto the latest main (was a few days behind). CI should be green again. Happy to address any feedback or make adjustments -- let me know if anything needs changing before this can move forward.

cixzhang:
Thanks @HelloOjasMutreja for working on this! Sorry we hadn't fully fleshed out the spec. The line clamp component works a bit differently internally but this is a good start. We have a "See more"/"See less" link at the end of the clamp. Here's a video of the behavior.

https://github.com/user-attachments/assets/4f5b8925-fe13-4e84-9cff-1a7ad44ac1b7

We're using a resize observer at the moment to check for overflow to display the see more/see less toggles, but there may be a better way to do this with modern CSS (haven't explored yet).

Let me know if you'd be down to update it to work this way! Otherwise, since this is a lab component we can accept it and adjust forward on my side.

For the time being, there's one test failure but this was resolved on main recently so you jsut need to rebase to get the green CI.

HelloOjasMutreja:
@cixzhang Rebased onto current main again to clear the CI failure you flagged — that test was already fixed there, so this should be green now.

Happy to take on the See more/See less + resize-observer rework if you'd like — let me know and I'll pick it up as a follow-up. Otherwise this is ready to land as-is and I'll leave the iteration to you, whichever you'd prefer.

packages/lab/src/LineClamp/LineClamp.tsx
```diff
@@ -0,0 +1,142 @@
...
+    <>
+      <Component
+        ref={mergeRefs(ref, truncation.ref, elementRef)}
+        {...mergeProps(
+          themeProps('line-clamp'),
+          stylex.props(styles.clamp, xstyle),
+          className,
+          {...style, WebkitLineClamp: maxLines},
+        )}
+        title={tooltipEnabled ? truncation.fullText : undefined}
```
cixzhang: This recreates the duplicate tooltip Text and Heading just removed in [#5559](https://github.com/facebook/astryx/pull/5559).

packages/lab/src/LineClamp/LineClamp.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+  const tooltipPlacement: LayerPlacement =
+    typeof hasTooltip === 'string' ? hasTooltip : 'above';
+  const tooltipEnabled =
+    hasTooltip !== false && maxLines > 0 && truncation.isTruncated;
+
+  return (
+    <>
+      <Component
+        ref={mergeRefs(ref, truncation.ref, elementRef)}
```
cixzhang: Could this use `useMergedRefs`? One rerender currently detaches and reattaches the consumer ref twice.

cixzhang (changes requested):
Thanks — I’d already said the lab shape can iterate, and the clamp/mixed-content paths work. Two current-main regressions need fixing before merge.

Clamped content mounts Astryx's Tooltip and sets the same native `title`, so hover shows the full text twice. Text and Heading just removed that pairing in [#5559](https://github.com/facebook/astryx/pull/5559).

`mergeRefs(...)` is also recreated every render. One rerender called a consumer ref four times (`null`/node twice), while Text's `useMergedRefs` control called it zero; that also re-registers truncation measurement.

Could you carry over those two current Text patterns? If you'd rather talk it through with someone, we're in [Discord](https://discord.com/invite/XnsUcFykEP).

| Current Text control | LineClamp at this head |
|---|---|
| Current Text control | LineClamp hovered |

[Reviewed by Robohands]

HelloOjasMutreja:
Pushed 4a5eba782 with both patterns carried over from #5559:

- Dropped the native `title={tooltipEnabled ? truncation.fullText : undefined}` entirely, matching Text's current shape — no more double announcement between the native title tooltip and the Tooltip component.
- Replaced `mergeRefs(ref, truncation.ref, elementRef)` (called inline, so a new callback every render) with `useMergedRefs(ref, truncation.ref, elementRef)`, matching Text's `mergedRef`.

Verified against `Text.tsx`'s current render block (lines 272-284) line by line to make sure the shape actually matches rather than approximating it.

packages/lab/src/LineClamp/LineClamp.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+export interface LineClampProps extends BaseProps<HTMLElement> {
+  /** Ref forwarded to the root element */
+  ref?: React.Ref<HTMLElement>;
+  /**
+   * Content to clamp. Unlike `Text`'s `maxLines` (which truncates a single
+   * text node), `LineClamp` wraps arbitrary children — mixed inline content,
+   * nested elements, anything that isn't a single `Text`.
+   */
+  children: ReactNode;
```
cixzhang, line 142: “Any children” includes controls. Clipped descendants remain tabbable and scroll hidden content into view on focus.

[Reviewed by Robohands]

cixzhang (changes requested):
This head fixes both blockers from my prior review: the duplicate native title is gone, and the consumer ref stays attached across rerenders.

The remaining public contract can strand keyboard users. In Chromium, a fully clipped link was the next Tab stop; focusing it scrolled the clamp 72px and replaced the visible line with the hidden action. Clamped content needs a predictable keyboard path with no invisible focusable descendants or focus-triggered content swap.

[Reviewed by Robohands]

HelloOjasMutreja:
Confirmed and fixed, both parts:

- `overflow: hidden` -> `overflow: clip`. Same visual clipping, but `clip` isn't a scroll container, so the browser's scroll-into-view-on-focus never fires. This alone stops the content swap.
- Added an effect that compares each focusable descendant's rect against the clamp box's own (clipped) rect, and removes anything past the fold from the tab order (`tabindex="-1"`, `aria-hidden="true"`), restoring each descendant's own original state once it's no longer clipped. Runs on every render and on the same shared `observeResize`/`unobserveResize` `useTruncation` already uses, since a resize can un-clip a descendant without the overall `isTruncated` flag changing.

Verified against a real Storybook story reproducing your exact scenario (a trailing link past `maxLines`): confirmed geometrically clipped, confirmed excluded from the real tab order (`element.tabIndex`, not just the attribute), confirmed restored after a render with different geometry.

One thing to disclose: the pure-resize-without-a-render path couldn't be independently verified in my environment. My remote browser tool's viewport resize doesn't dispatch native `resize` events or `ResizeObserver` notifications at all here (confirmed with a minimal same-realm `ResizeObserver` test unrelated to this component), so I couldn't watch that specific slice fire live. It reuses the exact `observeResize`/`unobserveResize` utility `useTruncation` already relies on for the same purpose though, not new or unproven machinery, and the render-driven path (which covers ordinary prop-driven updates) is fully verified.

Added regression tests with a mocked `getBoundingClientRect` (jsdom has no layout engine) for all three states, including preserving a consumer's own explicit `tabIndex` through the hide/restore cycle.

HelloOjasMutreja:
Rebased onto current main again to clear the stale CI failure (the unrelated api-dts type error in api/search/search.mjs); confirmed it no longer reproduces after the rebase. Also caught and fixed a missing afterEach import in the focus-restore tests that vitest's implicit globals let slip past runtime but strict typecheck flagged.

Typecheck, lint, and the full LineClamp suite (11 tests) all pass locally now.

HelloOjasMutreja:
Checked whether the current registry-contract/docsite-test/Vercel failures are caused by this PR: they are not. Reproduced the exact same error (Generated ShadCN names or routes changed...) by running node apps/docsite/scripts/generate-data.mjs against a clean upstream/main checkout in an isolated worktree, zero changes applied. It's a pre-existing docsite-generator issue unrelated to LineClamp.

HelloOjasMutreja:
Rebased onto current main, no conflicts (this PR's files are all new, nothing shared to collide with). LineClamp.test.tsx passes (11/11), typecheck clean. No change in scope or intent since the last update.

HelloOjasMutreja:
The keyboard-path fix from the last round is in: clipped focusable descendants get tabindex=-1 + aria-hidden (syncFocusableClipping, re-run on render and on resize), and overflow: clip (not hidden) stops the browser's scroll-into-view from swapping the visible line for hidden content on focus. Covered by the #4259 describe block in LineClamp.test.tsx. Rebased onto current main; typecheck/lint/tests all green.

PR #4260, opened 2026-07-23
fix(lab, charts): correct Windows shell-quoting in build:esm/dev scripts

### LINKED ISSUE #3969: [Bug] pnpm docsite fails with Can't resolve '@astryxdesign/charts'
### Summary

On a fresh clone of `main`, `pnpm build` reports `Successfully compiled 0 files with Babel` for `@astryxdesign/charts`. The `dist/` folder only contains `.d.ts` type declaration files — no `.js` output. This causes `pnpm docsite` to fail at startup with:

```
Module not found: Can't resolve '@astryxdesign/charts'
```

### Expected Behavior

`pnpm build` should compile all `.ts`/`.tsx` files in `packages/charts/src/` to `.js` in `dist/`, same as `@astryxdesign/core` (which compiled 482 files) and `@astryxdesign/lab`.

### Actual Behavior

Babel compiles **0 files** for charts. Only `tsc --project tsconfig.build.json` produces output (`.d.ts` files). The `dist/` folder is missing all `.js` files, making the package unresolvable at runtime.

### Environment

- **OS:** Windows 11 (Git Bash / MINGW64)
- **Node:** v24.x.x (via nvm)
- **pnpm:** 11.10.0
- **Repo:** `main` branch, fresh clone (no local modifications)

### Additional Context

- Source files exist in `packages/charts/src/` (`.ts` and `.tsx` files present)
- `packages/charts/babel.config.js` references `./babel-plugin-add-extensions.cjs` — this file may be missing or may cause Babel to silently fail on Windows
- `@astryxdesign/core` build works fine (482 files compiled with the same Babel approach)
- `@astryxdesign/lab` also reported 0 files compiled — may have the same root cause
- The `--ignore` pattern in `build:esm` uses single quotes which may behave differently on Windows/Git Bash vs Unix shells

### Suspicious Areas

1. **`babel-plugin-add-extensions.cjs`** — referenced in `babel.config.js` but may not exist in `packages/charts/`
2. **Single-quoted `--ignore` glob** — the `build:esm` script wraps the ignore pattern in single quotes which Git Bash on Windows may not strip correctly, potentially causing Babel to ignore all files
3. **`rm -rf dist`** in the build script — `rm` is a Unix command; on Windows Git Bash it works but could behave unexpectedly in some edge cases

### Steps to Reproduce

1. Clone the repo fresh:
   ```bash
   git clone https://github.com/facebook/astryx.git
   cd astryx
   ```
2. Install Node v24.x via nvm:
   ```bash
   nvm install
   ```
3. Install pnpm:
   ```bash
   npm install -g pnpm@11
   ```
4. Install dependencies:
   ```bash
   pnpm install
   ```
5. Build all packages:
   ```bash
   pnpm build
   ```
6. Observe the charts build output — it says **0 files** compiled:
   ```
   $ babel src --out-dir dist --extensions '.ts,.tsx' --out-file-extension '.js' \
     --ignore '**/*.test.ts,...' --config-file ./babel.config.js
   Successfully compiled 0 files with Babel (33ms).
   ```
7. Confirm no JS files in dist:
   ```bash
   ls packages/charts/dist/*.js
   # (empty — no output)
   ```
8. Start the docsite:
   ```bash
   pnpm docsite
   ```
9. See the error:
   ```
   Module not found: Can't resolve '@astryxdesign/charts'
   ../../packages/cli/templates/pages/table-page-chart/page.tsx (22:1)
   ```

### Astryx Version

  "version": "0.1.5"

### Environment

_No response_

HelloOjasMutreja:
Fixes #3969

`build:esm`/`dev` in `packages/lab/package.json` and `packages/charts/package.json` wrapped `--extensions` and `--ignore` in single quotes. pnpm/npm run package.json scripts through `cmd.exe` on Windows regardless of the calling shell (even from Git Bash), and `cmd.exe` doesn't strip single quotes the way POSIX shells do. Babel received the literal argument value `'.ts,.tsx'` (including the quote characters), matched zero files, and `pnpm build` silently succeeded with an empty `dist/` — only `.d.ts` and CSS got generated by the unaffected build steps, breaking any consumer that imports these packages at runtime (e.g. the docsite failing to resolve `@astryxdesign/charts`, as reported).

Fixed to match `@astryxdesign/core`'s already-correct quoting: unquoted `--extensions`/`--out-file-extension` (no spaces, no quoting needed) and escaped double quotes for the `--ignore` glob list, which `cmd.exe` honors.

Also checked `cli`/`build`/`themes/*` for the same pattern — none affected (`cli` has no babel build step, themes use `tsup`).

No CI coverage caught this because every workflow in `.github/workflows/` only runs on `ubuntu-latest`.

## Testing
- Reproduced on Windows: confirmed `dist/index.js` missing while `dist/index.d.ts`/CSS existed for both packages
- Fixed, rebuilt both packages clean — `lab`: 111 files compiled, `charts`: 31 files compiled, `dist/index.js` present for both
- Full test suites pass: `@astryxdesign/lab` 248/248, `@astryxdesign/charts` 8/8

No changeset: both packages are `private` and not published to npm.

Opening as a draft per the contributing guide.

### DIFF
--- a/packages/charts/package.json
+++ b/packages/charts/package.json
@@ -47,9 +47,9 @@
   "scripts": {
     "prepack": "pnpm build",
     "build": "rm -rf dist && pnpm build:esm && tsc --project tsconfig.build.json && pnpm build:css && node ../../scripts/check-no-dev-jsx.mjs",
-    "build:esm": "babel src --out-dir dist --extensions '.ts,.tsx' --out-file-extension '.js' --ignore '**/*.test.ts,**/*.test.tsx,**/__tests__/**,**/*.perf.test.*,**/*.doc.mjs,**/*.d.ts,**/*.css,**/*.css.d.ts' --config-file ./babel.config.js",
+    "build:esm": "babel src --out-dir dist --extensions .ts,.tsx --out-file-extension .js --ignore \"**/*.test.ts,**/*.test.tsx,**/__tests__/**,**/*.perf.test.*,**/*.doc.mjs,**/*.d.ts,**/*.css,**/*.css.d.ts\" --config-file ./babel.config.js",
     "build:css": "node ../../scripts/build-css.mjs --package charts",
-    "dev": "babel src --watch --out-dir dist --extensions '.ts,.tsx' --out-file-extension '.js' --ignore '**/*.test.*,**/*.doc.mjs,**/*.d.ts,**/*.css*' --config-file ./babel.config.js",
+    "dev": "babel src --watch --out-dir dist --extensions .ts,.tsx --out-file-extension .js --ignore \"**/*.test.*,**/*.doc.mjs,**/*.d.ts,**/*.css*\" --config-file ./babel.config.js",
     "test": "vitest run --root ../..",
     "test:watch": "vitest --root ../..",
     "lint": "eslint src",
--- a/packages/lab/package.json
+++ b/packages/lab/package.json
@@ -47,9 +47,9 @@
   "scripts": {
     "prepack": "pnpm build",
     "build": "rm -rf dist && pnpm build:esm && tsc --project tsconfig.build.json && pnpm build:css && node ../../scripts/check-no-dev-jsx.mjs",
-    "build:esm": "babel src --out-dir dist --extensions '.ts,.tsx' --out-file-extension '.js' --ignore '**/*.test.ts,**/*.test.tsx,**/__tests__/**,**/*.perf.test.*,**/*.doc.mjs,**/*.d.ts,**/*.css,**/*.css.d.ts' --config-file ./babel.config.js",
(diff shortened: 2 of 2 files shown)


HelloOjasMutreja:
Real terminal output, before and after this branch:

Windows build fix before and after

Same repro confirmed on @astryxdesign/charts. Root cause: --extensions/--ignore were wrapped in single quotes, which the Windows shell that pnpm scripts run through does not strip, so Babel silently matched zero files while the build still reported success.

cixzhang (approved):
Correct fix: the single-quoted `--extensions`/`--ignore`/`--out-file-extension` glob args in the lab and charts `build:esm`/`dev` scripts don't get dequoted by cmd.exe on Windows, so the literal quotes leak into the glob and break the build there. Switching bare flags to unquoted and the glob lists to double quotes is the portable form and behaves identically on POSIX shells. Scoped to two package.json script lines, no runtime impact. LGTM.

Issue #4249, opened 2026-07-23
PowerSearch: support a leading start icon (forward Tokenizer's existing startIcon)


cixzhang:
## Summary

`PowerSearch` should support an **icon at the beginning of the input** (a leading/start icon, e.g. a search glyph before the filter tokens/placeholder). Today there's no way to render one.

## Good news: the mechanism already exists one layer down

`PowerSearch` renders a `Tokenizer` internally, and **`Tokenizer` already has a `startIcon?: ReactNode | IconType` prop** (rendered before the tokens, with `startIconWithTokens` styling). `PowerSearch` just doesn't **forward** it.

So this is a small plumbing/consistency task, not new infrastructure:
- Add a `startIcon?: ReactNode | IconType` prop to `PowerSearch` and forward it verbatim to the internal `<Tokenizer startIcon={startIcon} />`.
- Reuse the exact sibling type (`ReactNode | IconType`, resolved via the shared icon-slot helper) — do not invent a PowerSearch-specific icon prop.

## Considerations
- Placement + spacing should match Tokenizer's existing `startIcon` treatment (including the `startIconWithTokens` case when tokens are present).
- Decorative by default (aria-hidden) unless it needs a label — follow Tokenizer/Icon conventions.
- Non-breaking: optional prop, absent = today's behavior.

## Acceptance
- `PowerSearch` accepts `startIcon` and renders it before the input, forwarding to the internal Tokenizer's existing `startIcon`.
- Matches Tokenizer's start-icon spacing/behavior; non-breaking; test + doc + story.

cixzhang:
Resolved by #4262 (PowerSearch startIcon support merged).

PR #4262, opened 2026-07-23
docs(Tokenizer, PowerSearch): document and test the existing startIcon prop

HelloOjasMutreja:
Relates to #4249

The issue as filed says "today there's no way to render [a start icon]." That's not accurate — `PowerSearch` already has a working `startIcon?: ReactNode | IconType` prop, already forwards it verbatim to the internal `Tokenizer`, and it's already exercised in Storybook (`MagnifyingGlassIcon` at lines 1067/1091 of the stories file). Git history shows this prop predates the issue.

The real gap: neither `Tokenizer.doc.mjs` nor `PowerSearch.doc.mjs` documented it, and neither had test coverage — so it was invisible to `astryx component <Name> --dense`, the docsite props table, and anyone (human or AI) relying on those as the source of truth to discover it exists.

This is purely additive documentation + tests, no behavior change — so it doesn't need the spec protocol (no new API surface).

## Changes
- Add `startIcon` to the props table (en/zh/dense) in both `Tokenizer.doc.mjs` and `PowerSearch.doc.mjs`
- Add colocated tests to both components confirming the icon renders and forwards correctly

## Testing
- 70/70 tests pass across both files (including 5 new)
- `tsc --noEmit` clean
- `eslint` clean
- Verified `astryx component PowerSearch/Tokenizer --dense` now shows `startIcon` in the props table

Suggest closing #4249 once this lands, or repurposing it to track the actual gap if I've missed something.

Opening as a draft per the contributing guide.

### DIFF
--- a/.changeset/docs-tokenizer-powersearch-document-and-test-the-jmwj.md
+++ b/.changeset/docs-tokenizer-powersearch-document-and-test-the-jmwj.md
@@ -1,0 +1,8 @@
+---
+'@astryxdesign/core': patch
+---
+
+[docs] docs(Tokenizer, PowerSearch): document and test the existing startIcon prop
+@HelloOjasMutreja
+
+Both components have shipped a working `startIcon?: ReactNode | IconType` prop for a while (`PowerSearch` forwards it verbatim to the internal `Tokenizer`, and it's already exercised in Storybook), but neither's `.doc.mjs` documented it and neither had test coverage — so it was invisible to `astryx component <Name> --dense`, the docsite props table, and anyone (human or AI) relying on those as the source of truth. No behavior change; adds the missing props-table entries (en/zh/dense) and colocated tests confirming the icon renders and forwards correctly.
--- a/packages/core/src/PowerSearch/PowerSearch.doc.mjs
+++ b/packages/core/src/PowerSearch/PowerSearch.doc.mjs
@@ -84,6 +84,13 @@ export const docs = {
         'Validation status object with type and optional message.',
     },
     {
+      name: 'startIcon',
+      type: 'ReactNode | IconType',
+      description:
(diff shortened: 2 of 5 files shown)


HelloOjasMutreja:
Real output from the CLI, before and after this branch:

PowerSearch docs before and after

Command used: node packages/cli/bin/astryx.mjs component PowerSearch --dense. No behavior change; the prop already worked, this just makes it discoverable.

cixzhang (approved):
Approving. Thanks for the careful diagnosis — you're right that the prop already existed and forwards to the internal Tokenizer; the real gap was discoverability (missing from the doc tables + no test coverage), which is exactly what this fixes.

A few things I verified:
- **API is consistent with the input family.** `startIcon?: ReactNode | IconType` matches TextInput, NumberInput, TextArea, Selector, MultiSelector, and Typeahead — so Tokenizer/PowerSearch were already aligned; this just makes it visible in the docs surface (`--dense`, docsite props table) and locks it with tests.
- **Tests are meaningful:** omitted vs. present, both accepted forms (ReactNode and IconType), and the DOM-order assertion (icon before tokens).
- Additive docs + tests only, no behavior change.

One tiny note (not blocking, no action needed here): your doc `type: 'ReactNode | IconType'` is actually more accurate than the sibling inputs, which document it as just `'IconType'` even though the real prop accepts both. I'll file a small follow-up to bring those up to match yours.

Nice cleanup.

cixzhang (approved):
Verified: `startIcon?: ReactNode | IconType` is a real, shipped prop on both `Tokenizer` (`Tokenizer.tsx`) and `PowerSearch` (`PowerSearch.tsx`), and `PowerSearch` forwards it to the internal `Tokenizer`. The added props-table entries (en/zh/dense) and colocated tests accurately describe existing behavior — no source/behavior change. Ran the touched suites locally: Tokenizer + PowerSearch (290 tests) pass. Nice fill of a docs/coverage gap. LGTM.

PR #4264, opened 2026-07-23
feat(release): bump the minor for breaking changes while 0.x

ejhammond:
## What

Adopt standard semver for the `0.x.y` range in the release protocol: a **`[breaking]`** changeset now bumps the **minor** (`0.x.y → 0.(x+1).0`), while every other category bumps the **patch**.

Previously **all** changes shipped as a `patch` and `[breaking]` was only an editorial changelog category, decoupled from the actual version bump. Under a caret range like `^0.1.8`, npm resolves anything `<0.2.0` — so bumping the minor (`0.1.x → 0.2.0`) is precisely the signal that a release "may break you." This re-couples the existing `[breaking]` category to that bump.

`major` remains rejected while we're `0.x` (it would jump straight to `1.0.0`). The whole coupling keys off whether publishable packages are still `0.x`, so it lifts automatically at 1.0.

## Changes

- **`scripts/changeset-new.mjs`** — derives the bump from the chosen category (`[breaking] → minor`, else `patch`) instead of hardcoding `patch`.
- **`scripts/check-changesets.mjs`** — the CI backstop (runs via `check:repo`) now enforces the category↔bump coupling **both ways**: a `[breaking]` that isn't `minor` fails, and a `minor` that isn't `[breaking]` fails. `major` stays rejected pre-1.0. The per-changeset validation was extracted into a pure, exported `validateChangeset()`.
- **`scripts/check-changesets.test.mjs`** — new unit tests covering the coupling (breaking→minor passes; breaking→patch, non-breaking→minor, and major all fail; coupling lifts once past 1.0).
- **`CONTRIBUTING.md`** — "Version Bumps" section rewritten to document the rule.

No changeset is included: this touches release tooling in `scripts/` and docs only — neither is a publishable package.

> The wiki (Release-Process, Distribution) will be updated to match once this merges, since it's the source of truth and shouldn't describe behavior that isn't shipped yet.

## Test plan

- `pnpm check:changesets` passes on current changesets.
- New unit tests + existing script tests: `vitest run scripts/*.test.mjs --project node` → 16/16 pass.
- Manual matrix verified end-to-end:
  - `[breaking]` → `minor` ✅ pass
  - `[breaking]` → `patch` ❌ fails with a clear message
  - `[feat]` → `minor` ❌ fails with a clear message
  - any → `major` ❌ fails (pre-1.0)
- `pnpm changeset:new --category breaking …` writes `minor`; `--category fix …` writes `patch`.
- prettier + eslint clean; pre-commit `check:repo` passes.

### DIFF
--- a/scripts/check-changesets.mjs
+++ b/scripts/check-changesets.mjs
@@ -4,9 +4,10 @@
  * CI gate for XDS changesets — `node scripts/check-changesets.mjs`.
  *
  * Enforces the conventions the release process depends on:
- *   1. Pre-1.0 patch-only: while every publishable package is < 1.0.0, no
- *      changeset may declare a `minor` or `major` bump (would jump 0.0.x ->
- *      0.1.0 / 1.0.0).
+ *   1. 0.x semver coupling: while every publishable package is < 1.0.0, the
+ *      bump must match the category. A [breaking] changeset must be `minor`
+ *      (0.x.y -> 0.(x+1).0, the breaking tier under caret ranges); every other
+ *      category must be `patch`. `major` is rejected (it would jump to 1.0.0).
  *   2. Every changeset body must carry a recognized [category] tag.
  *   3. Every changeset body must credit at least one @contributor.
  *   4. Frontmatter packages must be real, publishable, non-ignored packages.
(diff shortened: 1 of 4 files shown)


scripts/check-changesets.mjs
```diff
@@ -126,8 +163,13 @@ function main() {
...
   }
   console.log(
-    `✓ check:changesets — ${files.length} changeset(s) valid${pre1 ? ' (pre-1.0 patch-only enforced)' : ''}`,
+    `✓ check:changesets — ${files.length} changeset(s) valid${ctx.pre1 ? ' (0.x: [breaking] -> minor, else patch)' : ''}`,
   );
 }
 
-main();
+// Run as a script, but stay importable for unit tests.
+if (process.argv[1] === fileURLToPath(import.meta.url)) {
```
nynexman4464, line 171: how pythonic

nynexman4464 (approved):
nicely done!

PR #4265, opened 2026-07-23
chore(docs): type-check every .doc.mjs across all packages

josephfarina:
## Why
Follow-up to #4263 — an invalid `PropDoc` key (`defaultValue` instead of `default`) in `RichTextEditor.doc.mjs` only surfaced in the **Vercel docsite build** (not a required PR check), so it landed on `main` and broke every subsequent build.

## The check
Just run `tsc --checkJs` over every component `.doc.mjs`, repo-wide. Each doc is checked against its `@type {ComponentDoc}` annotation → `docs-types.ts`, so a bad key fails at the **source line**:

```
RichTextEditor.doc.mjs(33,7): error TS2353: Object literal may only specify known properties,
and 'defaultValue' does not exist in type 'PropDoc'.
```

…instead of 30k lines into a generated file. That's the whole doc check — no separate runtime validator (an earlier draft added one; `tsc` against `docs-types.ts` is the single source of truth and strictly more thorough, so it was dropped).

## What's here
- **`tsconfig.docs.json`** — auto-discovering glob `packages/*/src/**/*.doc.mjs` (+ root-level package docs like `groups.doc.mjs`), `allowJs` + `checkJs`, `noEmit`. New packages are covered with zero config. Root `typecheck:docs` script.
- **CI** — `ci.yml` + `deploy.yml` now run `pnpm typecheck:docs` (was `@astryxdesign/core`-only).
- **Fixed 32 lab doc `@type` paths** — off by one dir (`../../core` → `../../../core`), so their annotations resolved to `any` and were never actually checked. This is what let the bug hide in lab.
- **`scripts/docs-typecheck-coverage.test.mjs`** — guards that the glob keeps covering every component doc.

## Root cause of the gap
`@astryxdesign/lab` had no `typecheck:docs` gate (only core did), **and** every lab doc's `@type` path was wrong — so lab's 32 `.doc.mjs` files were effectively unchecked.

## Verification
`pnpm typecheck:docs` passes (core + all 32 lab docs, first time). Reintroducing `defaultValue` fails `tsc` at the source line. `generate` is unchanged.

_Docs-tooling + lab doc-comment fixes; no runtime/API change._

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -182,9 +182,9 @@ jobs:
         if: needs.check-scope.outputs.docsite_only != 'true'
         run: node scripts/verify-exports.mjs
 
-      - name: Typecheck component docs
+      - name: Typecheck component docs (all packages)
         if: needs.check-scope.outputs.docsite_only != 'true'
-        run: pnpm -F @astryxdesign/core typecheck:docs
+        run: pnpm typecheck:docs
 
       - name: Typecheck template docs
         if: needs.check-scope.outputs.docsite_only != 'true'
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -75,8 +75,8 @@ jobs:
       - name: Verify package exports
         run: node scripts/verify-exports.mjs
 
-      - name: Typecheck component docs
-        run: pnpm -F @astryxdesign/core typecheck:docs
+      - name: Typecheck component docs (all packages)
+        run: pnpm typecheck:docs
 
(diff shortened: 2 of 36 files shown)


josephfarina:
Closing (draft, unmerged). Folding the repo-wide doc type-check (tsconfig.docs.json + coverage test) into the consolidated doc-types reorg rather than landing it standalone. Branch `chore/harden-doc-typechecks` is preserved. Plan: drafts/doc-types-reorg-plan.md.

PR #4267, opened 2026-07-23
fix(docsite): make color scheme togglable in theme preview

aryan02420:


ernestt:
Can you elaborate a bit on what and why?

apps/docsite/src/components/ThemePackagePage.tsx
```diff
@@ -1062,15 +1062,17 @@ export function ThemePackagePage({packageName, theme}: ThemePackagePageProps) {
         <div {...stylex.props(styles.showcaseBlock)}>
           <Card padding={0} xstyle={styles.showcaseCard}>
             <Theme theme={selectedTheme} mode={mode}>
```
cixzhang, line 1064: This should actually already set the color scheme

cixzhang:
Thanks for contributing! Closing for now since this PR is likely stale and there's no description and unclear whether this is intending to fix something. Feel free to open a new one if you're still having the issue!

PR #4274, opened 2026-07-23
ci(cli): add strict checkJs type gate with zero-tolerance allowlist + ratchet

josephfarina:
## What

The CLI ships as hand-written `.mjs` — there's no build step — and is type-checked via TypeScript's `checkJs` against JSDoc annotations. The problem: only a handful of files were ever in a `checkJs` include list (`typecheck:json-api` covers `json.mjs`, `manifest.mjs`, `parse.mjs`, `config.mjs`; the api-contract project has `checkJs: false`). So for the ~140 other source files, type errors and missing annotations could land freely — the JSDoc was effectively decorative from CI's perspective.

This adds a strict, **ratcheting** gate that closes the door on new type debt without requiring a boil-the-ocean cleanup first.

## How it works

`tsconfig.strict.json` runs the compiler over the whole non-test tree (`src/**/*.mjs` + `src/**/*.ts`, excluding tests) under full `strict`. `scripts/strict-typecheck-gate.mjs` turns that into a merge gate enforcing two invariants:

1. **Enforced set — zero tolerance.** Every file in `strict-typecheck-allowlist.json` (plus all `.d.ts`) must have **zero** strict errors. These files can never regress; a new source file is expected to be strict-clean and added to the list. Because `strict` includes `noImplicitAny`, this *forces complete JSDoc annotations* — an un-annotated parameter is a hard error.
2. **Whole-tree ratchet — monotonic.** Total strict-error count (including files not yet in the enforced set) may never rise above the number in `strict-typecheck-baseline.json`. Existing debt can only shrink.

Seeded at the current state: **1617 total strict errors, 40 files enforced at zero**.

## Where it runs

Wired into the **required `build` job** (`ci.yml`) — right after `pnpm build` (so `@astryxdesign/core` dist is present) and next to the existing typecheck steps. Also mirrored in `deploy.yml` (the post-merge safety net, same pattern the other typechecks already follow). `build` is a required status check, so no PR can merge with a new type failure or an un-annotated enforced file.

## Workflow for contributors

```bash
pnpm -F @astryxdesign/cli typecheck:strict          # the gate (CI runs this)
pnpm -F @astryxdesign/cli typecheck:strict:update    # after clearing debt: lower baseline + promote clean files
```

`:update` recomputes the baseline and promotes every now-clean file into the enforced set, locking progress in. The long-term goal is `maxErrors: 0` with every file enforced — reachable incrementally, one cleanup PR at a time. Documented in `packages/cli/CONTRIBUTING.md`.

## Test

- `pnpm -F @astryxdesign/cli typecheck:strict` → passes (40 enforced, 1617 ≤ 1617).
- Negative tests: injecting a type error into an enforced file fails invariant 1; any new tree-wide error fails invariant 2 (both exit 1).
- Existing `typecheck:json-api` and `typecheck:template-docs` still pass.
- No runtime/published-behavior change — CI tooling only.


josephfarina:
Superseded by #4275, which takes a simpler path: land the strict check as an opt-in `typecheck:strict` command first (no CI gate, no allowlist/ratchet), then burn down the ~1617 errors in follow-up PRs, then flip it on in CI once the tree is clean.

Issue #4179, opened 2026-07-22
Item: add default/outline/muted variants (Card-style border + background, element radius)


cixzhang:
## Summary

Add `variant` support to `Item` with three values — **`default`**, **`outline`**, and **`muted`** — that apply the same border and background-color treatment as `Card`, but using **element** border radius (`--radius-element`) rather than Card's container radius.

## Current state

`Item` has **no `variant` prop** today. It renders a transparent background with `--radius-element` and only paints background on interaction (hover → `--color-overlay-hover`, selected → `--color-accent-muted`). There's no way to give an Item a standing border or a filled/muted surface the way `Card` can.

`Card` already implements the target pattern:
- `variant: 'default'` → `--color-background-card`, and `default` is the only variant that draws a visible border (`--color-border-emphasized`, `--border-width`), with the border width subtracted from padding so total inset stays on the spacing scale.
- `variant: 'muted'` → `--color-background-muted`, no border.
- `variant: 'transparent'` → no bg, no border.

## Proposed behavior

Add `variant?: ItemVariant` with:

| Variant | Background | Border | Notes |
|---|---|---|---|
| **`default`** | `--color-background-card` | visible (`--color-border-emphasized`, `--border-width`) | mirrors Card `default` |
| **`outline`** | transparent / none | visible border | border-only — the outline look |
| **`muted`** | `--color-background-muted` | none | mirrors Card `muted` |

- **Border radius:** use `--radius-element` (Item's existing radius), **not** Card's `--radius-container`. This is the one intentional difference from Card — Item is a smaller element, so it takes the element-scale radius.
- **Reuse Card's border-inset technique:** draw the border *inside* the padding (subtract `--border-width` from each side) so total inset stays faithful to the spacing scale — same as Card's `withBorder`.
- **Token parity with Card:** same background/border tokens so themes style Item and Card consistently.
- **Interaction states still layer on top:** hover/selected overlays continue to work over whichever variant surface is set.
- **Default value:** decide whether the new `default` variant becomes Item's default (visible border + card bg) or whether the current transparent look stays the default and `default` is opt-in. Given Item is used heavily in lists/menus where a border would be surprising, **transparent should likely remain the no-variant default** and `variant` is opt-in. (Confirm in spec — this is a potential visual regression if `default` auto-applies.)

## Design notes / open questions (for spec)

[...]

PR #4277, opened 2026-07-24
feat(Item): add default/outline/muted surface variants

AKnassa:
## What this does

Gives `Item` a `variant` prop with three looks — **`default`** (filled card background with a border), **`outline`** (border only), and **`muted`** (soft grey background, no border). It's opt-in: an `Item` without the prop looks exactly like it does today.

Refs #4179 — left open deliberately, since the issue is assigned and carries API questions this PR shouldn't close on its own.

## Why

Today an `Item` is always transparent and only paints a background when you hover or select it. There's no way to give one a standing border or a filled surface the way `Card` can. This adds that, reusing `Card`'s own colour and border tokens so the two stay visually consistent across themes — but at `Item`'s smaller `--radius-element` corner radius, since an `Item` is an element, not a container.

## What changed

- **New `variant` prop** on `Item` — `default`, `outline`, or `muted`. Leaving it off keeps the current transparent look, which is what lists and menus want.
- **Hover and selection now sit *on top of* the variant surface** instead of replacing it. This one isn't cosmetic — see the note below.
- **The border is drawn inside the padding.** Its width is subtracted from whichever density padding is in play, so the total inset still lands on the spacing scale and a bordered `Item` is the same overall size as a borderless one.
- `variant` is reflected as `data-variant` so themes can target each look.
- Documented in `Item.doc.mjs` (English, Chinese and dense), plus two new Storybook stories.

### The interesting bit

`--color-background-muted` and `--color-overlay-hover` are **the same value in light mode** (`#0536590C`). So the obvious implementation — having the variant paint `background-color` — means a `muted` Item gets *no visible hover at all*, and a `default` Item loses its solid card surface the moment you hover it.

The fix: the variant owns `background-color`, and the four interaction states (hover, pressed, highlighted, selected) moved to a `background-image` gradient layer, which CSS paints *above* the background colour. That's the same technique already used by `TreeListItem`, `AvatarGroupOverflow` and `ChatComposerDrawer`.

Because all four still write the same property in the same order, their precedence is unchanged — **an `Item` with no variant renders byte-identically to before.** There's a test pinning exactly that.

## How to see it

Storybook → **Core / Item** → *Variants* and *Variants (interactive)*. Hover the muted row: it visibly darkens. Before this change it wouldn't have
01-variants-light
02-variants-dark
03-interactive-light-resting
04-interactive-light-hover-MUTED
05-interactive-light-hover-DEFAULT
06-interactive-dark-hover-MUTED

## Testing

- 51 `Item` tests, 14 of them new. They read the **real compiled CSS** out of `document.styleSheets` rather than just checking the prop arrived, so they assert actual declarations.
- All 14 verified load-bearing by mutation testing — 9 deliberate breakages introduced, 9 caught.
- Full suite green apart from pre-existing `packages/cli` timeout flakes present on `main` at the same commit; zero failures in `packages/core`. Lint and typecheck clean.

## Open questions for review

Flagging these honestly rather than assuming — #4179 leaves them to the spec, and I've made a judgement call on each:

1. **Should `default` become `Item`'s default?** I kept the no-variant look transparent (the issue leans this way too), so nothing regresses in the many lists and menus using `Item` today. `default` is opt-in.
2. **What background should `outline` have?** I used none — border only, fully transparent — so it composes over any surface. The alternative was `--color-background-surface`.
3. Per the Component Specification Protocol, a new prop on a core component wants API agreement on the issue before implementation. **Treat this PR as a concrete proposal to react to, not a fait accompli** — happy to reshape or drop it if the API should land differently.


cixzhang (changes requested):
Thanks for this @AKnassa — the mechanics here are solid (border-inside-padding inset, layering interaction states *over* the variant surface, `--radius-element`, `data-variant` for theming). Two of the three variants are exactly right and need no changes.

One spec change since you opened this, now confirmed with the EPS team (the parity target): **there is no filled-card `default` variant.** The aligned set is:

| Variant | Background | Border |
|---|---|---|
| **`transparent`** (default) | none | none — today's no-variant look, paints only on hover/selected |
| **`outline`** | transparent | visible border |
| **`muted`** | muted-surface fill | none |

EPS ships exactly `transparent (its default) / outline / muted` and has **no** bordered-and-filled surface. Since parity is the goal, we're matching that and dropping the `card`-style variant. (Context + rationale in the [proposal comment on #4179](https://github.com/facebook/astryx/issues/4179#issuecomment-5137067312).)

**Requested changes:**
1. **Remove the filled-card `default` variant** (the card-background + border look).
2. Change the type to `type ItemVariant = 'transparent' | 'outline' | 'muted'`, with `transparent` as the default (no-variant Item stays transparent — no behavior change).
3. `outline` and `muted` stay as-is — they already match.
4. Update the changeset, doc entries, Storybook stories, and tests to the three-variant set (drop the `default`-paints-card-background assertions).

The interaction-state layering fix is great and should stay — it's still needed for `muted`.

packages/core/src/Item/Item.tsx
```diff
@@ -205,19 +232,39 @@ const styles = stylex.create({
...
+  // Interaction overlays are painted as background-image layers rather than
+  // background-color, so they composite *over* the variant surface (which owns
+  // background-color) instead of replacing it. Without this, `muted` would show
+  // no hover at all — --color-background-muted and --color-overlay-hover are the
+  // same value in light mode. Same technique as TreeListItem and
+  // AvatarGroupOverflow.
   interactive: {
     cursor: 'pointer',
-    transitionProperty: 'background-color',
+    transitionProperty: 'background-image',
```
cixzhang, line 243: Hmm, `background-image` isn't transitionable, so this fade is dead. TreeListItem has the same on main at :102.

cixzhang (changes requested):
Thanks @AKnassa — all four asks are in, and the border inset measures clean: transparent, outline and muted come out the same height.

The overlay move costs the fade, though. Chrome doesn't interpolate `background-image`, so hover and press now snap on every clickable Item — every menu row, list row and search result, and nobody opted in. An inset `box-shadow` overlay keeps both the layering and the fade; I tried it here and it transitions.

Also needs a merge with main — main added a disabled guard to the same hover rule. And the description still has the old filled-card `default` in it; the changeset is right, the body isn't.

Does box-shadow work, or is there a reason to stay on the gradient?

[Reviewed by Robohands]

cixzhang:
Checking in on Item variants: the current review needs an overlay mechanism that preserves hover/press transitions after the main refresh. Would you like to continue, have us carry it, or close this for now?

Issue #4282, opened 2026-07-24
[Bug] Chat composer: ArrowUp/ArrowDown recall message history mid-text, blocking multi-line caret navigation


lazydevz-woo:
### Description

`ChatComposerInput` (the contentEditable chat composer) uses ArrowUp/ArrowDown to navigate message history regardless of the caret's position, calling `preventDefault()` — so in a multi-line message the caret can't move between lines. ArrowUp replaces the current draft with the previous message; ArrowDown steps through recalled history the same way.

**Expected:** when the caret is mid-text, ArrowUp/ArrowDown move it between lines; history is navigated only at the text boundaries — ArrowUp at the very start, ArrowDown at the very end.

**Related:** #2473 (planned composer state-model refactor), #3635 (chat components tracker). Happy to open a PR with a fix + tests if this is confirmed as a bug.

### Reproduction

_Recording (astryx.dev docs playground): multi-line draft — pressing ArrowUp mid-text replaces the draft with the previous message instead of moving the caret up a line._

https://github.com/user-attachments/assets/17bb8491-6028-4f49-8ca5-d169b60b2053

**Steps:**
1. Open the `ChatComposerInput` playground on https://astryx.atmeta.com/components/ChatComposerInput — or Storybook → `Core/ChatComposerInput` → **Message History**.
2. Submit a couple of messages (type + Enter) to build history.
3. Type a multi-line draft: `aaa`, Shift+Enter, `bbb`.
4. Place the caret on the first line (or mid-text) and press ArrowUp.
5. **Expected:** the caret moves up a line. **Actual:** the draft is replaced by the previous message; the caret can't move.

### Astryx Version

@astryxdesign/core@0.1.8

### Environment

Chrome 150.0.7871.130 (official build, arm64), macOS 26.5.1 (Tahoe), Apple M5. Keyboard interaction; not AT-specific.

cixzhang:
Tracked in the accessibility program tracker: #4475

cixzhang:
Tracked in the accessibility program tracker: #4475

PR #4284, opened 2026-07-24
fix(core): preserve multi-line caret navigation in ChatComposerInput

Lee-Dongwook:
## Related Issue
https://github.com/facebook/astryx/issues/4282

  ## Summary

  Fixes `ChatComposerInput` message-history navigation so ArrowUp and ArrowDown only recall history when the caret is at the relevant text
  boundary.

  This preserves the browser’s default vertical caret movement within multi-line drafts.

  ## Changes

  - Recall the previous message with ArrowUp only when the selection is at the start of the draft.
  - Move forward through message history with ArrowDown only when the selection is at the end of the draft.
  - Preserve repeated history navigation when a recalled message is fully selected.
  - Restore the pending draft after navigating past the newest history entry.
  - Add selection-boundary helpers for contenteditable composer input.
  - Add a patch changeset for `@astryxdesign/core`.

  ## Tests

  Added coverage for:

  - ArrowUp recall at the start of a draft
  - Default ArrowUp behavior when the caret is mid-text, including multi-line content
  - Forward/backward history navigation
  - Pending-draft restoration
  - Empty history and `hasHistory={false}`


cixzhang:
Merged current `main` into this branch (it was 544 commits behind) so CI runs against something current — no code changes. Prettier is clean and the `ChatComposerInput` tests pass locally (58/58).

packages/core/src/Chat/chatComposerSelection.ts
```diff
@@ -57,6 +57,78 @@ export function ensureCaretInside(editable: HTMLElement): Selection | null {
   return selection;
 }
 
+/**
+ * Whether a range spans no visible content — no text and no element
+ * (tokens, `<br>`). An empty or whitespace-free range between a
+ * boundary and the caret means the caret is at that boundary.
+ */
+function isRangeEmpty(range: Range): boolean {
+  const contents = range.cloneContents();
```
cixzhang: This copies the full draft on every boundary arrow key. Might need a non-copying boundary check.

cixzhang (changes requested):
Thanks — the caret and history behavior is right in Chromium. One thing before this lands: both boundary helpers clone the editor contents, so an ArrowUp/Down at a boundary now copies the entire draft before the existing history work. Could we avoid content-sized copying in this keydown path and bank a count-based regression?

[Reviewed by Robohands]

Lee-Dongwook:
> Thanks — the caret and history behavior is right in Chromium. One thing before this lands: both boundary helpers clone the editor contents, so an ArrowUp/Down at a boundary now copies the entire draft before the existing history work. Could we avoid content-sized copying in this keydown path and bank a count-based regression?
> 
> If you'd rather talk it through with someone, we're in [Discord](https://discord.com/invite/XnsUcFykEP).
> 
> [Reviewed by Robohands]
> 

Thanks for answering. Since I watched your's comment now, I'll re-check by myself in this PR with  following your's feedback!

Lee-Dongwook:
@cixzhang 

Hello! Here's the change summary of commit.

---

Addressed the cloneContents feedback in eec64c5a5.

- Replaced the content-cloning boundary checks with Range.compareBoundaryPoints(), so ArrowUp/ArrowDown boundary detection does not create a DocumentFragment proportional to the draft size.
- Added count-based regression coverage for both ArrowUp and ArrowDown boundary paths, asserting that Range.prototype.cloneContents is called zero times.
- Kept the existing behavior for collapsed carets and fully selected recalled history entries.

Validation: git diff --check passed. I could not run Vitest locally because dependencies are not installed in this checkout and the package registry DNS lookup is unavailable.

packages/core/src/Chat/chatComposerSelection.ts
Lee-Dongwook: Removed the existing `isRangeEmpty()` helper and its `range.cloneContents()` call. The caret’s start/end position is now compared directly with the editor boundaries using `Range.compareBoundaryPoints()`. As a result, ArrowUp/ArrowDown no longer creates a DocumentFragment by copying the entire draft.

cixzhang (changes requested):
Thanks — removing the content clone fixes the prior cost, but the boundary check now misses the positions Chromium actually creates. At the start of a non-empty draft Chromium reports `#text` offset 0; `compareBoundaryPoints` returns 1, so ArrowUp leaves the draft unchanged. The tests were changed to place selection on the editor root, which is why they pass. Please cover real text-node start/end positions without content-sized copying.

[Reviewed by Robohands]

Lee-Dongwook:
Thanks for catching this. I replaced the `compareBoundaryPoints` check with a DOM-edge walk that recognizes Chromium’s actual caret positions: offset 0 in the first text node and text.length in the final text node.
  
The check does not clone or materialize draft content; it only inspects the boundary offset and relevant sibling/ancestor paths. I also updated the `ArrowUp` and `ArrowDown` regression tests to place the selection on real text-node boundaries and kept assertions that `Range.prototype.cloneContents` is never called.
  
git diff --check passed.

cixzhang (approved):
Thanks, this is good. My earlier boundary checks now pass in real Chromium: multiline ArrowUp stays in the draft, text-node boundaries recall history, and ArrowDown restores it.

[Reviewed by Robohands]

PR #4285, opened 2026-07-24
feat(core): make custom ChatComposer inputs first-class

cixzhang:
## Summary

Follow-up to #4280 (finding #9). That PR fixed the input's *own* Enter/key behavior; this makes a **custom input a first-class citizen** of `ChatComposer`, so you can drop in a plain `<textarea>` or a bespoke editor and have the whole shell work — not just `ChatComposerInput`.

The composer body is a shell of slots (drawer, header, **input**, footer, send). The input is one slot; it doesn't span the body. Two things previously locked custom inputs out:

1. **The composer↔input contract was private.** `useChatComposerContext()` carried `value`/`onChange`/`onSubmit`/`canSend`/`placeholder`/`isDisabled` — everything the built-in input and `ChatSendButton` read — but it wasn't exported, so a custom input couldn't reach send/stop state or wire to the same contract.
2. **Body-click-to-focus sniffed the DOM.** Clicking empty space in the body focused the input via `querySelector('[contenteditable], textarea')` — a hardcoded shape guess that misses any editor whose focusable node isn't one of those.

## Change

- **Export the composition contract:** `useChatComposerContext`, `ChatComposerContextValue`, and a new `ChatComposerInputControl` from `@astryxdesign/core/Chat`.
- **Replace focus sniffing with a registered control.** The context carries a stable `inputControlRef`; an input registers `{ focus }` on mount, and the shell drives it on body click — for any input shape. The DOM query stays as a fallback so uninstrumented custom inputs don't lose click-to-focus. `ChatComposerInput` registers its existing `focus()` automatically.

Since the context was private, this is a design opportunity, not just an export: `ChatComposerInputControl` is a minimal, named interface that can grow (optional methods) without forcing plain inputs to implement more than `focus`.

## Examples (Storybook)

New `Components/ChatComposer/Custom Input` stories prove the contract with two very different inputs, both first-class:

- **PlainTextarea** — a bare controlled `<textarea>`: reads context, submits on Enter (IME-safe), registers focus. No contentEditable.
- **RawLexical** — a raw Lexical editor (its own `EditorState` + focus model), bridged to the composer's string value and registering `editor.focus()` as its control.

## Testing

- `ChatComposer.test.tsx`: custom input reads value/placeholder/disabled from context; body-click focuses via the registered control; DOM-query fallback still focuses a bare textarea.
- `ChatComposerInput.test.tsx`: the built-in input registers a focus control so a body click focuses it.
- `pnpm build`, core `typecheck`, `typecheck:docs`, storybook `typecheck`, `verify-exports`, and changeset checks clean; full Chat suite green.

## Compatibility

Additive. `inputControlRef` is optional on the context; the focus fallback preserves today's behavior for any input that doesn't register. No change to existing `ChatComposer`/`ChatComposerInput` usage.

### DIFF
--- a/packages/core/src/Chat/ChatComposerInput.tsx
+++ b/packages/core/src/Chat/ChatComposerInput.tsx
@@ -384,6 +384,20 @@ export function ChatComposerInput(props: ChatComposerInputProps) {
   selfRef.current = handle;
   useImperativeHandle(handleRef, () => handle);
 
+  // Register a focus control with the composer shell so body-click-to-focus
+  // works without the shell sniffing the input's DOM shape. Cleared on
+  // unmount so the shell falls back cleanly if the input goes away.
+  const inputControlRef = composerCtx?.inputControlRef;
+  useEffect(() => {
+    if (!inputControlRef) {
+      return;
+    }
+    inputControlRef.current = {focus: () => editableRef.current?.focus()};
+    return () => {
+      inputControlRef.current = null;
+    };
+  }, [inputControlRef]);
+
   useEffect(() => {
     if (controlledValue === undefined || !editableRef.current) {
       return;
--- a/.changeset/chat-composer-input-composition.md
+++ b/.changeset/chat-composer-input-composition.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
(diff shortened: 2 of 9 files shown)


packages/core/src/Chat/ChatComposerInput.tsx
```diff
@@ -384,6 +384,20 @@ export function ChatComposerInput(props: ChatComposerInputProps) {
   selfRef.current = handle;
   useImperativeHandle(handleRef, () => handle);
 
+  // Register a focus control with the composer shell so body-click-to-focus
+  // works without the shell sniffing the input's DOM shape. Cleared on
+  // unmount so the shell falls back cleanly if the input goes away.
+  const inputControlRef = composerCtx?.inputControlRef;
+  useEffect(() => {
```
cixzhang, line 391: Should we do this registration on a callback ref instead of as an effect? The effect presupposes editor ref existence, etc.

PR #4287, opened 2026-07-24
feat(calendar): add calendar-nav theme target for month-nav controls

freddymeta:
## What

Adds a dedicated **`astryx-calendar-nav`** theme target to `Calendar`'s prev/next month-navigation buttons, so consumers can theme the nav controls independently of every other `Button` in their app.

## Why

`Calendar` already documents theme targets for its root (`astryx-calendar`) and its day cells (`astryx-calendar-day`), so a consumer can scope overrides to those surfaces via `defineTheme({ components: { 'calendar-day': { ... } } })`.

The month-nav buttons had no such seam. They render as `<Button variant="ghost" isIconOnly>`, so the *only* handle for theming them was the global `.astryx-button` class — which applies to **every** Button in the app, not just the calendar's arrows. There was no way to, say, tint only the nav controls, round them differently, or restyle the disabled edge of a clamped date range, without a change that leaks across the whole app's buttons.

This is a gap in the component's own theming model: an element with no stable target is an unthemeable element (the same principle the repo's `themingTargets` guard enforces).

## What changed

- **`packages/core/src/Calendar/Calendar.tsx`** — both nav buttons now spread `themeProps('calendar-nav', { nav, disabled })`, rendering a stable `astryx-calendar-nav` class plus `data-nav` (`prev`/`next`) and a reflected `disabled` state — mirroring the existing `calendar-day` pattern exactly. StyleX-only; no raw CSS; the change is additive so default rendering is byte-for-byte unchanged (still a ghost icon button).
- **`packages/core/src/Calendar/Calendar.doc.mjs`** — documents the new target under `theming.targets` (both `docs` and `docsZh`), with `nav` as a visual prop and `disabled` as a state. This is what the `themingTargets` guard and codegen read to learn which selectors exist.
- **`packages/core/src/Calendar/Calendar.test.tsx`** — new tests assert the target renders on both buttons, the `nav`/`disabled` reflections are correct, the default rendering is preserved (still `astryx-button ghost`), the existing `calendar-day` target/states still render, and that `defineTheme` resolves `calendar-nav` (and `nav:next`) to the expected scoped selectors.
- **`apps/storybook/stories/Calendar.stories.tsx`** — a `ThemedNavAndSelectedRing` story demonstrating a `defineTheme` override scoped to the nav controls (and the selected day-cell ring) via `<Theme>`.
- **`.changeset/`** — patch changeset.

Because this only *adds* a class + data attributes, `defineTheme({ components: { 'calendar-nav': { ... } } })` now resolves to `.astryx-calendar-nav` (in `@layer astryx-theme`, above the StyleX `astryx-base` layer), so consumer overrides take effect while the stock look is preserved by default.

## Test plan

Run locally (matching the CI gates in `.github/workflows/ci.yml`):

- `pnpm build` — all packages build
- `pnpm test` — full suite passes; `Calendar.test.tsx` 57 tests (5 new), `themingTargets.test.ts` 275 tests
- `pnpm -F @astryxdesign/core typecheck` + docs typecheck, `pnpm -F @astryxdesign/storybook typecheck` + build, `pnpm -F @astryxdesign/cli typecheck:template-docs`
- `pnpm -F @astryxdesign/docsite generate && test`
- `node scripts/sync-exports.js --check`, `node scripts/check-sync.js`, `node scripts/check-changesets.mjs`, `node scripts/verify-exports.mjs`
- `eslint` on the changed files — clean

## Related

A follow-up, independent change makes the day-cell **selected / today ring** states precisely themeable — the reflected `today` / `in-range` classes don't line up 1:1 with the compound conditions under which each ring is drawn, so a theme can over-match. That's handled Calendar-locally (no shared theme-pipeline change) in a separate PR: #4289.

### DIFF
--- a/.changeset/calendar-nav-theme-target.md
+++ b/.changeset/calendar-nav-theme-target.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] Calendar: add a dedicated `astryx-calendar-nav` theme target for the prev/next month-nav buttons, so consumers can theme the nav controls (color, radius, per-direction, disabled edge) without reaching every Button via the global `astryx-button` handle. Reflects `direction` (prev/next) and the `disabled` state as data attributes.
+
+@freddymeta
--- a/.changeset/calendar-nav-theme-target.md
+++ b/.changeset/calendar-nav-theme-target.md
@@ -2,6 +2,6 @@
 '@astryxdesign/core': patch
 ---
 
-[feat] Calendar: add a dedicated `astryx-calendar-nav` theme target for the prev/next month-nav buttons, so consumers can theme the nav controls (color, radius, per-direction, disabled edge) without reaching every Button via the global `astryx-button` handle. Reflects `direction` (prev/next) and the `disabled` state as data attributes.
+[feat] Calendar: add a dedicated `astryx-calendar-nav` theme target for the prev/next month-nav buttons, so consumers can theme the nav controls (color, radius, per-direction, disabled edge) without reaching every Button via the global `astryx-button` handle. Reflects `nav` (prev/next) and the `disabled` state as data attributes.
 
 @freddymeta
--- a/apps/storybook/stories/Calendar.stories.tsx
+++ b/apps/storybook/stories/Calendar.stories.tsx
@@ -326,7 +326,7 @@ export const ThemedSelectedAndTodayRing: Story = {
  * - `components['calendar-nav']` scopes overrides to the prev/next month
  *   buttons only (via the `astryx-calendar-nav` target), instead of the global
  *   `astryx-button` handle that would hit every Button in the app. The
- *   `direction` visual prop and `disabled` state are reflected as data
+ *   `nav` visual prop and `disabled` state are reflected as data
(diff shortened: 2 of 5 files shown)


cixzhang (approved):
Reviewed. Clean, well-scoped additive change: both month-nav buttons now spread `themeProps('calendar-nav', {nav, disabled})`, mirroring the existing `calendar-day` pattern. StyleX-only, default rendering preserved (Button's className merge keeps `astryx-button ghost` alongside the new `astryx-calendar-nav`). Full component surface updated — tsx, doc.mjs (docs + docsZh), tests, Storybook story, patch changeset. CI green. LGTM.

cixzhang (approved):
Re-approving after rebasing onto main to resolve the merge conflict with #4289 (which landed first and touched the same Calendar files). Conflict resolution was mechanical — both PRs' theming targets, tests, and stories now coexist; the calendar-nav prop key is `nav` per the branch's rename commit. Verified locally: Calendar.test.tsx 63/63 pass, core typecheck clean, eslint clean, sync-exports + changesets pass. No behavior change to the merge itself. LGTM.

PR #4289, opened 2026-07-24
feat(calendar): make the today/selected day-cell ring precisely themeable

freddymeta:
## What

Makes `Calendar`'s **today** and **selected** day-cell ring treatments precisely themeable through `defineTheme`, by reflecting the ring's *actual* state as a single compound `marker` theme-state on `calendar-day`.

## Why

`Calendar` already reflects `selected` / `today` / `in-range` / `disabled` as theme-state classes on each day cell, so `defineTheme({ components: { 'calendar-day': { ... } } })` overrides land in `@layer astryx-theme` and win over the component's own StyleX (`@layer astryx-base`). Layer precedence isn't the problem.

The problem is **expressiveness**. Those states are reflected *independently*, but the ring treatments are drawn under *compound, mutually-exclusive* conditions:

- the plain today ring — only when the day is today, **not** single-selected, and **not** in a range
- the today-in-range ring — only when the day is today, **not** single-selected, and **inside** a range

So a theme author writing `components['calendar-day'].today` targets `.astryx-calendar-day.today`, which matches **every** today cell — including a today cell that is single-selected, where the default deliberately shows *no* today ring. And `today+in-range` (`.today.in-range`) can't distinguish the range-interior ring from other today+in-range combinations. The exact state the author wants — "today and in-range but not selected" — isn't expressible, because the theme key grammar only ANDs positive classes; there's no `:not()`.

Rather than add negation to the shared theme-key parser (a cross-cutting design-system change), this fixes it **inside Calendar**: reflect the ring's real state directly.

## What changed

- **`packages/core/src/Calendar/Calendar.tsx`** — derive the ring treatment once (`showsTodayRing` / `showsTodayInRangeRing`) and:
  - reflect it as a compound `marker` theme-state via `themeProps('calendar-day', { …, marker })` — `'today-only'`, `'today-in-range'`, or absent;
  - route the existing StyleX ring styles through the *same* derived flags, so the reflected state can never drift from the styling.
  This maps `defineTheme` keys 1:1 to the states the ring is actually drawn under:
  `components['calendar-day']['marker:today-only']` → `.astryx-calendar-day.today-only`, and
  `['marker:today-in-range']` → `.astryx-calendar-day.today-in-range` — no negation needed. (`selected` continues to own the selection fill.)
- **`packages/core/src/Calendar/Calendar.doc.mjs`** — document `marker` under `calendar-day` states (`docs` + `docsZh`).
- **`packages/core/src/Calendar/Calendar.test.tsx`** — tests assert the reflected `marker` state for a plain today cell, today inside a range, single-selected today (no ring), and a today range endpoint (ring preserved — a range endpoint is not `isSelected`, which is single-mode only, so the today-in-range ring is intentionally kept). Also asserts the states resolve to the expected `defineTheme` selectors.
- **`apps/storybook/stories/Calendar.stories.tsx`** — a `ThemedSelectedAndTodayRing` story overriding the ring states via `<Theme>`.
- **`.changeset/`** — patch changeset.

The change is purely additive to the reflected surface, and the StyleX conditions are unchanged in behavior (just factored through shared flags), so **default rendering is pixel-identical** — asserted in the tests.

## Not in scope (deliberately)

I did **not** add `:not()` support to the shared `parseStyleKey`, nor reorder any global CSS layers — both would be design-system-wide changes with broad cascade implications. This PR keeps the fix Calendar-local. If there's appetite for general negation in the theme-key grammar as a separate capability, that's worth its own discussion.

## Test plan

Run locally (matching `.github/workflows/ci.yml`):

- `pnpm build`
- `pnpm test` — full suite passes; `Calendar.test.tsx` 58 tests (6 new), `themingTargets.test.ts` 275
- `pnpm -F @astryxdesign/core typecheck` + docs typecheck; `pnpm -F @astryxdesign/storybook typecheck` + build
- `pnpm -F @astryxdesign/docsite generate && test`
- `node scripts/sync-exports.js --check`, `check-sync.js`, `check-changesets.mjs`, `verify-exports.mjs`
- `eslint` on the changed files — clean

### DIFF
--- a/.changeset/calendar-day-ring-states.md
+++ b/.changeset/calendar-day-ring-states.md
@@ -2,6 +2,6 @@
 '@astryxdesign/core': patch
 ---
 
-[feat] Calendar: make the today/selected day-cell ring precisely themeable. The day cell now reflects a compound `ring` state (`today-only` / `today-in-range`) that maps 1:1 to the treatment actually drawn, so `defineTheme({components: {'calendar-day': {'ring:today-only': {...}}}})` targets exactly those states without over-matching or needing a `:not()` exclusion. Default rendering is unchanged.
+[feat] Calendar: make the today/selected day-cell ring precisely themeable. The day cell now reflects a compound `marker` state (`today-only` / `today-in-range`) that maps 1:1 to the treatment actually drawn, so `defineTheme({components: {'calendar-day': {'marker:today-only': {...}}}})` targets exactly those states without over-matching or needing a `:not()` exclusion. Default rendering is unchanged.
 
 @freddymeta
--- a/.changeset/calendar-day-ring-states.md
+++ b/.changeset/calendar-day-ring-states.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] Calendar: make the today/selected day-cell ring precisely themeable. The day cell now reflects a compound `ring` state (`today-only` / `today-in-range`) that maps 1:1 to the treatment actually drawn, so `defineTheme({components: {'calendar-day': {'ring:today-only': {...}}}})` targets exactly those states without over-matching or needing a `:not()` exclusion. Default rendering is unchanged.
+
+@freddymeta
--- a/apps/storybook/stories/Calendar.stories.tsx
+++ b/apps/storybook/stories/Calendar.stories.tsx
@@ -278,25 +278,26 @@ export const AllVariations: Story = {
 /**
  * Theme the selected-date and today rings precisely via `defineTheme`.
  *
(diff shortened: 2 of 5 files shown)


cixzhang (approved):
Reviewed. Nice refactor — deriving `showsTodayRing`/`showsTodayInRangeRing` once and routing both the reflected `marker` state and the StyleX ring styles through the same flags eliminates drift (replaces 4 duplicated compound conditions with 2 named booleans). Reflecting a compound `marker` state (`today-only`/`today-in-range`) lets themes target exactly the drawn treatment without `:not()`, and it's correctly kept Calendar-local rather than touching the shared theme-key parser. Full surface updated, 6 new tests including default-preserved assertions. CI green. LGTM.

Issue #4256, opened 2026-07-23
Avatar: fallback initials font (size/weight/color) can't be themed — inline dynamic font-size isn't token-reachable


cixzhang:
## Summary

The `Avatar` fallback initials **can't be fully themed**, because the initials' `font-size` is emitted as an **inline dynamic StyleX value** (a computed pixel number), not a token. A theme can re-scope tokens, but it has no way to reach an inline `fontSize: size * ratio`, so consumers who need a different fallback type scale (or weight/color) can't get there by theming — they're forced to abandon `Avatar` and hand-compose their own.

## Root cause (origin/main, `packages/core/src/Avatar/Avatar.tsx`)

- Initials font size is a **dynamic inline style**:
  ```ts
  const INITIALS_FONT_SIZE_RATIO = 0.4;                    // line 55
  const dynamicStyles = stylex.create({
    fontSize: (size: number) => ({ fontSize: size * INITIALS_FONT_SIZE_RATIO }),  // ~line 146
  });
  ```
  Because this resolves to a literal px value inline, **theme token re-scoping cannot override it.** The initials render at `avatarSize × 0.4` regardless of theme.
- Fallback weight is hardcoded to `fontWeight: fontWeightVars['--font-weight-medium']` (500, line ~130). This is theme-reachable via that token, but re-scoping `--font-weight-medium` for Avatar also affects everything else on that token — there's no Avatar-scoped weight hook, and it doesn't help with the size wall anyway.

Net: a consumer wanting fallback initials at a different **size scale**, **weight**, or **color** cannot achieve it through the theme; they must rebuild the component.

## Why this matters (parity)

This is a concrete design-system parity gap: an adjacent internal library renders Avatar fallback initials at a per-size type scale (smaller sizes → smaller initials, regular weight, a muted secondary-text color on a wash fill). Matching that by **theming Astryx's Avatar is impossible today** specifically because of the inline dynamic font size. The only workaround is to fork the component, which defeats the point of a themeable design system.

## Direction (for the spec)

Make the fallback initials' typography **theme-reachable**, without breaking today's default. Options to weigh:
- **Token-drive the fallback text scale.** Replace the inline `size * 0.4` with a small set of per-size fallback-text tokens (xs/sm/md/lg keyed to the avatar size tiers), so a theme can re-scope the scale. This mirrors how per-size typography is tokenized elsewhere and is the change the parity case needs.
[...]

PR #4292, opened 2026-07-24
feat(Avatar): make fallback initials typography theme-reachable via derived vars

Lee-Dongwook:
## Issue
https://github.com/facebook/astryx/issues/4256

## Summary

Avatar's fallback initials rendered at an inline `size × 0.4` px value with a
hardcoded weight and color. Because that font size resolves to a literal `px`
value inline, **no theme token could reach it** — a consumer needing a different
fallback type scale (or weight/color/fill) had to abandon Avatar and hand-compose
their own.

This routes the fallback typography through the design system's existing
**component-scoped derived-var** mechanism (the same one Button uses for
`--_button-radius`), so a theme can re-scope the fallback per size tier.
**Default rendering is byte-for-byte unchanged.**

## Motivation

An adjacent internal library renders Avatar fallback initials at a per-size type
scale (smaller avatars → proportionally smaller initials, regular weight, a muted
secondary-text color on a wash fill). Matching that by theming Astryx's Avatar was
impossible today, specifically because of the inline dynamic font size. The only
workaround was forking the component — which defeats the point of a themeable
design system.

## Approach

Introduce four Avatar-scoped internal vars on the fallback surface, each read with
a default that reproduces today's exact output:

| Internal var                     | CSS property      | Default (unchanged)          | Applies to      |
| -------------------------------- | ----------------- | ---------------------------- | --------------- |
| `--_avatar-fallback-font-size`   | `fontSize`        | `size × 0.4px` (per-size)    | initials        |
| `--_avatar-fallback-font-weight` | `fontWeight`      | `--font-weight-medium`       | initials        |
| `--_avatar-fallback-color`       | `color`           | `--color-text-secondary`     | initials + icon |
| `--_avatar-fallback-background`  | `backgroundColor` | `--color-neutral`            | initials + icon |

Registered in `derivedVarRegistry` and documented in `Avatar.doc.mjs`
(`theming.vars` / `theming.derived`), matching the existing Button/Card
convention. No new global tokens; nothing added to `tokens.stylex.ts`.

Consumers can now theme the fallback without forking:

```ts
defineTheme({
  name: 'brand',
  components: {
    avatar: {
      base:      { fontWeight: 'var(--font-weight-normal)',
                   color: 'var(--color-text-secondary)',
                   backgroundColor: 'var(--color-accent-muted)' },
      'size:sm': { fontSize: '9px'  },
      'size:md': { fontSize: '13px' },
      'size:xl': { fontSize: '40px' },
    },
  },
})
```

## Non-breaking

Every var defaults to the current value, so unthemed Avatars are unchanged. A new
`Avatar.test.tsx` case asserts the initials still render at `size × 0.4` (sm =
9.6px) — via the seam — so a regression back to a bare literal fails loudly.

## Scope / non-goals

The status-dot corner offsets use the same inline-dynamic pattern but are
**out of scope** — they're structural geometry, not a themeable design choice, and
weren't the reported gap.

## Test plan

- [x] `Avatar.test.tsx` — 7/7 pass, incl. new seam assertion
- [x] `generateThemeRules.test.ts` — new cases prove `components.avatar` emits the
      internal vars for `base` and a per-size (`size:sm`) key
- [x] `derivedVarRegistry.test.ts` — registry ↔ doc consistency (added
      `Avatar: 'avatar'` mapping)
- [x] core theme suite 533/533 pass; `typecheck` + `eslint` clean
- [x] Storybook: new `Themed Fallback Type Scale` story (default vs themed)

## ScreenShot
스크린샷 2026-07-24 오후 9 54 56


cixzhang (approved):
Thanks for adding this! I think this is a reasonable use of internal vars since we need to straddle 2 visible components representing the main avatar. Another way might be to add a target for content, but that feels more like a key part of the avatar itself rather than an extra part.

Lee-Dongwook:
> Thanks for adding this! I think this is a reasonable use of internal vars since we need to straddle 2 visible components representing the main avatar. Another way might be to add a target for content, but that feels more like a key part of the avatar itself rather than an extra part.

Thanks a lot for comment! I resolved merge conflict now!

PR #4293, opened 2026-07-24
fix(core): remove misleading TabList orientation prop

### LINKED ISSUE #4231: TabList: remove the misleading orientation prop (no real vertical tabs; it only drives the keyboard hint)
## Summary

`TabList`'s `orientation?: 'horizontal' | 'vertical'` prop is misleading: **it does not produce a vertical tab list.** The tab strip always renders horizontally. The prop only drives the keyboard-hint badge — the roving-focus hook already uses `orientation: 'both'` (accepts all arrow axes per the WAI-ARIA tabs APG) regardless of the prop, and `aria-orientation` is deliberately NOT set on the `<nav>` (invalid on the navigation role — triggers an axe violation). So `orientation="vertical"` changes essentially nothing visible/structural; it's effectively a no-op that implies a capability (vertical tabs) we don't have.

## Proposal

**Remove the `orientation` prop from `TabList`** until we design proper vertical tabs. Reasons:
- It advertises a layout mode that doesn't exist — a consumer setting `orientation="vertical"` gets a horizontal strip, which is a correctness/expectation bug.
- Keyboard nav already accepts both arrow axes unconditionally (`orientation: 'both'`), so removing the prop doesn't regress keyboard behavior.
- "Don't expose a prop whose alternate value doesn't do the thing it implies" — the same principle applied to statusVariant/borderless controls.

When we genuinely add vertical tabs, reintroduce `orientation` (or a clearer prop) as part of that design — with real layout + `aria-orientation` handling on the correct element.

## Considerations
- This is a **breaking** removal of a public prop (pre-1.0 → `[breaking]` changeset category, not a semver major). Provide a codemod that strips `orientation` from `<TabList>` usages (both values were effectively no-ops, so removal is safe).
- Confirm no internal consumer relies on the keyboard-hint-badge difference the prop currently drives; if the badge behavior matters, preserve it via a different mechanism or default.

## Acceptance
- `orientation` removed from `TabListProps`; keyboard nav unchanged (both axes still work); no `aria-orientation` regressions; codemod + `[breaking]` changeset; docs updated.

HelloOjasMutreja:
## Summary

Fixes #4231.

`TabList` exposed an `orientation?: 'horizontal' | 'vertical'` prop that implied vertical tab strips existed. They do not. The prop only controlled which arrow icons appeared in the ephemeral keyboard-hint badge (`useKeyboardHint`). It had zero effect on:

- **Layout** -- the strip always renders horizontally.
- **Arrow-key navigation** -- `useListFocus` has always used `orientation: 'both'`, accepting all four arrow axes unconditionally per the WAI-ARIA tab APG.
- **`aria-orientation`** -- deliberately never set (invalid on the `navigation` role; triggers an axe `aria-allowed-attr` violation).

Keeping the prop lets consumers believe they can render vertical tabs when they cannot. Removing it is the honest fix.

### `packages/core`
- **`TabList.tsx`** -- remove `orientation` from `TabListProps` and the function signature; call `useKeyboardHint()` with no arguments (defaults to `'horizontal'`); update stale comment in `useListFocus` block.
- **`TabListContext.ts`** -- remove the now-unused `TabListOrientation` type.
- **`TabList.test.tsx`** -- update the `aria-orientation` regression test to remove the `orientation="vertical"` rerender (the invariant itself is still tested).
- **`TabList.doc.mjs`** -- remove orientation from the props table.

### `packages/cli` (codemod)
- **`transforms/v0.1.9/remove-tablist-orientation.mjs`** -- strips `orientation` from all `<TabList>` usages. Import-source-aware (`@astryxdesign/core`, `@astryxdesign/core/TabList`, `@xds/core`, `@xds/core/TabList`), alias-aware.
- **`transforms/v0.1.9/index.mjs`** -- v0.1.9 transform manifest.
- **`transforms/v0.1.9/__tests__/remove-tablist-orientation.test.mjs`** -- 8 tests: both orientation values, expression values, aliases, subpath imports, no-op paths, non-regression for other components.
- **`codemods/registry.mjs`** -- registers `'0.1.9'` in the upgrade registry.

### Changeset
`[breaking]` (0.x minor bump) for `@astryxdesign/core`. Consumers run `astryx upgrade` to auto-remove the prop.

## Behavior change

The only observable difference is the keyboard-hint badge: it now always shows left/right arrows (the `useKeyboardHint` default) rather than up/down arrows when `orientation="vertical"` was passed. This is correct -- both axes have always worked regardless.

## Testing
- 45/45 `TabList.test.tsx` tests pass.
- 8/8 codemod unit tests pass.
- Full `@astryxdesign/core` build clean (Babel + tsc + CSS + UMD).


HelloOjasMutreja:
### API Cleanup & Codemod Details

- `useListFocus({hasRovingTabIndex: true})` handles two-dimensional keyboard navigation across the tab strip using both arrow axes per WAI-ARIA APG recommendations regardless of layout orientation.
- Includes an automated jscodeshift codemod (`transforms/remove-tablist-orientation.js`) that strips the deprecated `orientation` attribute from `<TabList>` JSX elements while handling custom local import aliases.

cixzhang (commented):
Reviewed against source — this is a clean, honest breaking change and I'd approve it. Verification:

- **The removal is correct.** `TabList` always rendered horizontally, `useListFocus` has always used `orientation: 'both'` (all four arrow axes per the tab APG), and `aria-orientation` was never set on the `navigation` role (invalid there — would trip axe `aria-allowed-attr`). So `orientation` genuinely only toggled the keyboard-hint badge arrows — a misleading no-op for layout. Removing it is the right call.
- **Residual `orientation` refs are intentional and correct:** `useListFocus({orientation: 'both'})` (the real nav behavior) stays, and `'aria-orientation': _ariaOrientation` is deliberately destructured out so consumers can't set an invalid attribute on the nav — good defensive move, worth the inline comment it already has.
- **Migration is handled properly:** the v0.1.9 codemod strips the attribute import-source-aware (`@astryxdesign/core`, `/TabList`, `@xds/core` legacy) and alias-aware, with 8 tests covering both values, expressions, aliases, subpaths, no-op paths, and non-regression. Registered in the upgrade registry.
- **Changeset** is correctly `minor` with the `[breaking]` tag + `astryx upgrade` instructions — appropriate for a 0.x prop removal.

One thing to confirm before merge: **does anything internal still pass `orientation` to `TabList`** (stories, docsite examples, other components)? The codemod covers consumer code, but in-repo call sites should be swept too so the build doesn't break on the now-removed prop. A repo-wide `orientation` grep on `<TabList` usages (and the removed `TabListOrientation` type import) would confirm. If that's clean, this is good to go.

cixzhang:
Followed up on my own open question: rebased the branch onto current main and swept the repo. **Internal call sites are clean** — the only remaining `orientation` references are the two intentional ones in `TabList.tsx` (`useListFocus({orientation: 'both'})` and the deliberately-swallowed `aria-orientation`), and there are **zero** `TabListOrientation` type references left anywhere in `packages/` or `apps/`. No stories/docsite/other components pass `orientation` to `<TabList>`. So the removal won't break the in-repo build. LGTM — just rebase onto main to clear the stale CI (it rebases clean) and it's ready.

cixzhang (changes requested):
Thanks for working on this! Since we updated, could you move the codemod to 0.2.0? Next update will augment the minor version.

HelloOjasMutreja:
@cixzhang Addressed both points:

- Codemod moved to `transforms/v0.2.0/` and the registry/registry test updated accordingly.
- Rebased onto current `main` to clear the stale CI.

Rebasing also surfaced a `typecheck:strict` regression in the moved codemod itself (`file`/`api`/`path`/`attr` were implicitly `any` under strict mode) — fixed with the same `@param` JSDoc pattern the other transforms use, and added test/type coverage stays green (`registry.test.mjs` + the transform's own test, 15/15 passing).

One heads-up: after rebasing, `build-storybook`'s `typecheck:strict` step still fails, but on `packages/cli/templates/blocks/components/**` files unrelated to this PR (DateInput, TextInput, Thumbnail, Timestamp, TreeList, InternationalizationProvider, Table showcase templates are out of sync with their components' current prop APIs). Confirmed this reproduces identically on a clean `main` checkout with no changes from this branch, so it looks like a pre-existing break on `main` itself rather than something introduced here — flagging in case it's not yet on your radar, since it'll red-X `build-storybook` on every open PR until it's fixed separately.

cixzhang (approved):

PR #4300, opened 2026-07-24
feat(thumbnail): add showRemoveOn prop

kentonquatman:
## Summary

Adds a `showRemoveOn` prop to `Thumbnail` to control when the overlaid remove button is visible. This PR is scoped to **only** that prop — no changes to the remove-button colors, hover styling, or image hairline.

### `showRemoveOn` prop — defaults to `'hover'`
Adds `showRemoveOn?: 'always' | 'hover'` (default `'hover'`). Only has an effect when `onRemove` is set. The naming follows the existing `Overlay` `showOn` vocabulary for consistency.

- `'hover'` — **(default)** the button is hidden at rest and revealed on hover **or keyboard focus**, and stays visible on any touch-capable device.
- `'always'` — the button is always shown at rest.

### Accessibility
The reveal is CSS-only (opacity) — the button stays mounted, so it remains keyboard- and screen-reader-reachable. Tabbing into the thumbnail triggers a `:focus-within` reveal. A scoped `defineMarker` + `stylex.when.ancestor(':hover' / ':focus-within')` keeps the reveal responsive to *this* thumbnail rather than any outer hoverable container. An `@media (any-pointer: coarse)` branch keeps the button visible on **any touch-capable device** — including hybrid touchscreen laptops (which report `hover: hover` + `pointer: fine` and would otherwise get no reveal on touch). Motion respects `prefers-reduced-motion`.

> **Examples audit:** because the default is hover-reveal, every example whose purpose is to *show* the remove button at rest passes `showRemoveOn="always"` explicitly (so it renders correctly in static docs previews/screenshots): the Storybook `WithRemove`, `WithCaption`, `Placeholder`, `MediaModeTest`, and `Gallery` stories; the `ThumbnailGallery`, `ThumbnailRemovable`, and `ThumbnailDisabled` (enabled row) showcase blocks. A new **Remove on hover** Storybook story demonstrates the default, and `showRemoveOn` is added as a Storybook control.

## Implementation
The remove button is wrapped in a positioned "slot" `<div>` that carries the opacity reveal and the scope marker (ancestor-marker styles can't ride on a child component's `xstyle`). The existing `MediaTheme`/`useImageMode` contrast handling and the secondary-Button treatment are unchanged — the slot simply fades the whole button in/out.

## Testing
- `pnpm -F @astryxdesign/core build` ✅ (verified in `dist/astryx.css`: the `@media (any-pointer: coarse)` reveal branch; no changes to remove-button colors or the `--color-border` hairline)
- `pnpm -F @astryxdesign/core typecheck` ✅
- Thumbnail unit tests: **22 passing** (adds 4 for `showRemoveOn`)
- Docsite: `generate` + **282 data-extraction tests passing**
- lint (strict/CI mode) ✅, `sync:exports` ✅, token-docs ✅

### DIFF
--- a/packages/core/src/Thumbnail/Thumbnail.tsx
+++ b/packages/core/src/Thumbnail/Thumbnail.tsx
@@ -43,6 +43,7 @@ import type {BaseProps} from '../BaseProps';
 import {mergeProps} from '../utils';
 import {themeProps} from '../utils/themeProps';
 import {useTranslator} from '../i18n';
+import {thumbnailScope} from './thumbnail.markers.stylex';
 
 export interface ThumbnailProps extends BaseProps<HTMLDivElement> {
   /** Ref forwarded to the root element */
@@ -96,6 +97,17 @@ export interface ThumbnailProps extends BaseProps<HTMLDivElement> {
    * @default false
    */
   isDisabled?: boolean;
+  /**
+   * When the remove button is visible.
+   * - `'hover'` — the button is revealed on hover, and on keyboard focus so
+   *   it stays reachable. On any touch-capable device it stays visible.
+   *   This is the default.
+   * - `'always'` — the button is always shown.
+   *
+   * Only has an effect when `onRemove` is set.
+   * @default 'hover'
+   */
+  showRemoveOn?: 'always' | 'hover';
   /**
    * Test ID for testing frameworks.
    */
@@ -192,11 +204,14 @@ const styles = stylex.create({
     overflow: 'hidden',
   },
 
-  removeButtonOverrides: {
+  removeSlot: {
     position: 'absolute',
     top: spacingVars['--spacing-1'],
     right: spacingVars['--spacing-1'],
     zIndex: 1,
+    lineHeight: 0,
+  },
+  removeButtonOverrides: {
     '--_button-radius': `calc(${radiusVars['--radius-element']} - ${spacingVars['--spacing-1']})`,
     height: 20,
     minWidth: 20,
@@ -220,6 +235,28 @@ const styles = stylex.create({
     borderRadius: 'inherit',
     zIndex: 1,
     lineHeight: 0,
+  },
+  // showRemoveOn="hover": the remove button is hidden at rest and revealed
+  // when the thumbnail is hovered or when focus enters it (keyboard). Only
+  // opacity is animated — the button stays mounted and focusable so tabbing
+  // to it triggers the :focus-within reveal. Any touch-capable device (incl.
+  // hybrid laptops that also report hover) keeps the button visible so the
+  // remove path is never gated behind an unavailable hover.
+  removeOnHover: {
+    opacity: {
+      default: 0,
+      [stylex.when.ancestor(':hover', thumbnailScope)]: {
+        '@media (hover: hover)': 1,
+      },
+      [stylex.when.ancestor(':focus-within', thumbnailScope)]: 1,
+      '@media (any-pointer: coarse)': 1,
+    },
+    transitionProperty: 'opacity',
+    transitionDuration: {
+      default: durationVars['--duration-fast'],
+      '@media (prefers-reduced-motion: reduce)': '0s',
+    },
+    transitionTimingFunction: easeVars['--ease-standard'],
   },
 });
 
@@ -249,7 +286,8 @@ function ImagePlaceholder() {
  *
  * Shows a skeleton shimmer while the image loads, the image on success, or
  * a placeholder icon on failure / when no src is provided. An overlaid
- * remove button appears when `onRemove` is set.
+ * remove button appears when `onRemove` is set — revealed on hover or
+ * keyboard focus by default (`showRemoveOn`), or always shown.
  *
  * The remove button uses a fixed `--color-overlay` scrim with an
  * `--color-on-dark` icon so it stays legible on any image without sampling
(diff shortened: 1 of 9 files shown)


cixzhang (changes requested):
Thanks, Kenton — dropping APCA image-sampling for a fixed treatment is the right direction, and reusing Overlay's `showOn` vocabulary for `showRemoveOn` keeps the API coherent. The `::after` tint matching ClickableCard (the `currentColor` derivation + `@media (hover: hover)` guard) is spot on, and the z-index layering checks out. Three blocking items before this lands.

**🔴 1. `showRemoveOn="hover"` breaks the remove path on touch (accessibility).** The reveal fires only on `:hover` (hover-capable), `:focus-within`, or `@media (hover: none) and (pointer: coarse)`. Keyboard, screen reader, mouse, and phone/tablet are all fine. But a **hybrid device** (touchscreen laptop, Surface — reports `hover: hover` + `pointer: fine`) matches none of the three, so touch gets no reveal:
- Interactive thumbnails degrade (tap focuses the inner button → `:focus-within` reveals remove — removal needs a second aimed tap).
- **Non-interactive** thumbnails (`onRemove`, no `onClick`) are the sharp edge: the remove button is the only focusable element, `opacity: 0` but still mounted and clickable. A touch user must tap an invisible destructive control — and that same tap both reveals *and* fires `onRemove`. Silent delete with no visual affordance.

Since touch is an operable/accessibility modality, this is blocking. A clean signal is `@media (any-pointer: coarse)` — if the device can be touched at all (including hybrid laptops with both a mouse and a touchscreen), just keep the button visible rather than gating it behind hover. Alternatively, treat "no `onClick`" as always-visible. Your call on the exact approach.

**🔴 2. Remove-button backing — use `--color-overlay`.** `backgroundColor: '#00000066'` is raw hex on the button's `xstyle` (top of the cascade → un-themeable), and the `eslint-disable` documents the bypass rather than fixing it. The semantic scrim token is **`--color-overlay`**, already used for this file's `uploadOverlay` — a one-token swap that keeps the darkening intent and stays themeable. (The `--color-on-dark` X and the secondary-variant hover/press overlay are already tokenized — no change there.)

**🔴 3. Inset hairline — `--color-border` (or derive from a token).** `light-dark(#00000010, #FFFFFF40)` is raw hex inside `light-dark()`, slipping past `no-hardcoded-styles` via the `const` indirection. Colors must be a var or derived from vars — no raw hex, including as `light-dark()` / `color-mix()` inputs. Please revert to `--color-border`; if a neutral-over-media edge is truly required, derive it from `--color-on-light` / `--color-on-dark` rather than raw hex.

**Confirmed clean:** the `elevation` removal is safe — it's not in the published CHANGELOG, and you've correctly edited the still-pending elevation changeset to drop Thumbnail.

Recommendation: **request changes**. All three have concrete targets — happy to pair on the touch-reveal approach.

kentonquatman:
@cixzhang to use `--border-color` for the hairline, I'll need to convert the light theme's hex value `#ebebeb` into the comparable 8-digit `#00000014` (for some reason the dark theme already uses an 8-digit hex). This is a bigger change, but possibly better because it will help maintain border contrast over different background colors.

cixzhang:
The rule is no hardcoded colors to keep components themeable. Perhaps apply a color mix over one of the other tokens but we should avoid hardcoded hex

packages/cli/templates/themes/neutral/neutralTheme.ts
```diff
@@ -196,7 +196,7 @@ export const neutralTheme = defineTheme({
     '--color-warning-muted': ['#f8da9d', '#deb4333D'],
 
     // Border
-    '--color-border': ['#ebebeb', '#FFFFFF1A'],
+    '--color-border': ['#00000014', '#FFFFFF1A'],
```
cixzhang: On the use of color border, the change here applies just to the neutral theme but semantically other themes may want an opaque border. Considering theming, do we have additional semantic considerations for thumbnails? For example could they match concerns for media where we apply translucent colors there as well? If so I might suggest preserving media behavior instead. Note that depending on light vs dark mode the border color would appear differently and unsure if this was your intention for thumbnails.

packages/core/src/Thumbnail/Thumbnail.tsx
```diff
@@ -190,6 +166,33 @@ const styles = stylex.create({
       default: '0',
       ':has(:focus-visible)': '2px',
     },
+    // Hover/active tint overlay, matching ClickableCard. A currentColor tint
```
cixzhang: Hmm I expect these to use the color-hover-overlay. Using something else would be unexpected and lead to bugs in theming. Is there a reason beyond clickable card to do it this way? (For these base styles consider what would be expected and minimal styling then for more detailed styling, we should move that to neutral or specific themes) Also in the clickable card case it sounds like a bug.

packages/core/src/Thumbnail/Thumbnail.tsx
```diff
@@ -323,21 +365,34 @@ export function Thumbnail({
...
+    />
+  ) : null;
+
+  const removeButtonEl = hasRemove ? (
+    <div
+      {...stylex.props(
+        styles.removeSlot,
+        isHoverReveal && styles.removeOnHover,
+      )}>
+      {imageMode != null ? (
```
cixzhang: May need to rebase over your other change to remove the image mode since this will conflict

cixzhang (approved):

PR #4305, opened 2026-07-25
feat(tree-list): add astryx-tree-list-chevron theme target

freddymeta:
## What

Adds a dedicated `astryx-tree-list-chevron` theme target to the TreeList expand/collapse toggle.

## Why

The toggle chevron carries only the functional `data-tree-toggle` attribute (used internally by TreeList's `activateItem` selector) and no stable theme handle. Consumers who want to restyle the chevron — color, or a different treatment per open/closed state — can today only reach it through a fragile `[data-tree-toggle]` selector, which couples themes to an internal functional attribute rather than Astryx's sanctioned theming surface.

This mirrors the recently-added `astryx-calendar-nav` target: a `themeProps()` call that renders a stable class plus reflected state, documented in `theming.targets`, so `defineTheme` can scope overrides to exactly this element.

## What changed

- **`TreeListItem.tsx`** — `themeProps('tree-list-chevron', {state})` spread (via `mergeProps`) onto the toggle. It lands on **both** rendered forms of the chevron: the interactive `<button>` (the primary themeable case) and the non-interactive fallback `<span>` (used only when toggling isn't wired up), so the target is stable regardless. Reflects open/closed as `data-state="expanded" | "collapsed"` plus the matching class token. The functional `data-tree-toggle` hook is untouched.
- **`TreeList.doc.mjs`** — new `{className: 'astryx-tree-list-chevron', states: ['state']}` entry in `theming.targets` for both `docs` (EN) and `docsZh` (ZH). Keeps the `themingTargets` guard green.
- **`TreeList.test.tsx`** — asserts the target class + `data-state` render on the toggle (collapsed and expanded), that `data-tree-toggle` and the `<button>` element are preserved, and that a `defineTheme({ components: { 'tree-list-chevron': { base, 'state:expanded' } } })` resolves to `.astryx-tree-list-chevron` / `.astryx-tree-list-chevron.expanded`.
- **`TreeList.stories.tsx`** — a `ThemedChevron` story demonstrating the `defineTheme` override channel.
- Changeset (`[feat]`, patch).

## Notes

- Purely additive: default rendering is byte-identical. The target only adds a class and a `data-state` attribute.
- The existing `astryx-tree-list-item` target sits on the row's inner **content `<div>`** (not the `<li role="treeitem">`); this new chevron target is a sibling handle on the toggle control itself.
- Chevron size was left as-is — it's already tokenized (`--spacing-4`), so no size change was needed. The point of this PR is the theme target.
- StyleX-only; no raw CSS; no new hardcoded values.

### DIFF
--- a/.changeset/tree-list-chevron-theme-target.md
+++ b/.changeset/tree-list-chevron-theme-target.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] TreeList: add a dedicated `astryx-tree-list-chevron` theme target for the expand/collapse toggle, so consumers can theme the chevron (color, per open/closed state) via `defineTheme` instead of reaching it through the functional `[data-tree-toggle]` attribute. Reflects the open/closed state as a `data-state` attribute (`expanded`/`collapsed`); the functional `data-tree-toggle` hook is unchanged.
+
+@freddymeta
--- a/apps/storybook/stories/TreeList.stories.tsx
+++ b/apps/storybook/stories/TreeList.stories.tsx
@@ -5,6 +5,7 @@ import {TreeList} from '@astryxdesign/core/TreeList';
 import type {TreeListItemData} from '@astryxdesign/core/TreeList';
 import {Icon} from '@astryxdesign/core/Icon';
 import {Badge} from '@astryxdesign/core/Badge';
+import {Theme, defineTheme} from '@astryxdesign/core/theme';
 import {
   FolderIcon,
   DocumentIcon,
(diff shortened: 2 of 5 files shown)


cixzhang (approved):
Clean, contained theme-target addition. The new `themeProps('tree-list-chevron', {state})` call site on the expand/collapse control is recorded in `TreeList.doc.mjs` `theming.targets` (both `docs` and `docsZh`) as `{className: 'astryx-tree-list-chevron', states: ['state']}`, satisfying the SUBSET theming convention — every rendered `astryx-*` class and every prop key is documented. Functional `data-tree-toggle` hook is unchanged; the target is purely additive. Test asserting the class renders + generated-CSS reachability, a story, and a changeset are all present. No new public API surface. CI green. Approving.

PR #4306, opened 2026-07-25
feat(tree-list): add astryx-tree-list-item-label theme target

freddymeta:
## What

Adds a dedicated `astryx-tree-list-item-label` theme target to the TreeList item's label text.

## Why

The item's label (`styles.label` span inside the row's content) has no stable class. A theme that wants to style **just** the label — for example, bolding the selected item's label — can today only reach it through a fragile structural selector like `button:not([data-tree-toggle]) > span`. That couples themes to internal DOM structure and breaks the moment the row's markup changes.

This adds a stable target the sanctioned way (`themeProps` + `theming.targets`), matching the `astryx-tree-list-item` / `astryx-calendar-*` precedent, so `defineTheme` can scope overrides to the label alone.

## What changed

- **`TreeListItem.tsx`** — `themeProps('tree-list-item-label', {selected})` spread (via `mergeProps`) onto the label `<span>`. Renders a stable `astryx-tree-list-item-label` class and reflects the row's selected state as `data-selected="selected"` plus the matching class token, so a theme can target the selected label specifically via `defineTheme`.
- **`TreeList.doc.mjs`** — new `{className: 'astryx-tree-list-item-label', states: ['selected']}` entry in `theming.targets` for both `docs` (EN) and `docsZh` (ZH). Keeps the `themingTargets` guard green.
- **`TreeList.test.tsx`** — asserts the target class renders on the label, `data-selected` reflects only for the selected item, the label keeps the `id` its row references via `aria-labelledby`, and that a `defineTheme({ components: { 'tree-list-item-label': { base, selected } } })` resolves to `.astryx-tree-list-item-label` / `.astryx-tree-list-item-label.selected`.
- **`TreeList.stories.tsx`** — a `ThemedItemLabel` story showing a selected-label weight (bold) override via `defineTheme`.
- Changeset (`[feat]`, patch).

## Notes

- Purely additive: default rendering is byte-identical. The target only adds a class and a `data-selected` attribute.
- **Which element:** this lands on the label `<span>` itself. Note the existing `astryx-tree-list-item` target sits on the row's inner **content `<div>`** (not the `<li role="treeitem">`), so these are distinct handles — the row vs. the label text within it.
- The label reflects the row's `selected` state (not its own) so themes can scope to the selected item's label without a descendant selector.
- StyleX-only; no raw CSS; no new hardcoded values.

### DIFF
--- a/.changeset/tree-list-item-label-theme-target.md
+++ b/.changeset/tree-list-item-label-theme-target.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] TreeList: add a dedicated `astryx-tree-list-item-label` theme target on the item's label text, so consumers can theme just the label (e.g. bold the selected item's label) via `defineTheme` instead of a fragile `button:not([data-tree-toggle]) > span` structural selector. Reflects the row's `selected` state as a `data-selected` attribute on the label.
+
+@freddymeta
--- a/apps/storybook/stories/TreeList.stories.tsx
+++ b/apps/storybook/stories/TreeList.stories.tsx
@@ -355,6 +355,29 @@ const chevronTheme = defineTheme({
   },
 });
 
+/**
+ * Theme the item label text precisely via `defineTheme`.
+ *
+ * - `components['tree-list-item-label'].base` scopes overrides to the label
+ *   text only (via the `astryx-tree-list-item-label` target), instead of a
+ *   fragile `button:not([data-tree-toggle]) > span` structural selector.
+ * - `selected` restyles just the selected item's label (bold here), which the
+ *   label reflects as a `data-selected` attribute.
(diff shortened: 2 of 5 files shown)


cixzhang (approved):
Clean, contained theme-target addition, same shape as the chevron target. The new `themeProps('tree-list-item-label', {selected})` call site on the label span is recorded in `TreeList.doc.mjs` `theming.targets` (both `docs` and `docsZh`) as `{className: 'astryx-tree-list-item-label', states: ['selected']}`, satisfying the SUBSET theming convention. The label keeps its `aria-labelledby` linkage (verified by test); the target is purely additive. Test asserting the class renders + selected reflection + generated-CSS reachability, a story, and a changeset are all present. No new public API surface. CI green. Approving.

PR #4309, opened 2026-07-25
feat(tree-list): add astryx-tree-list-guide theme target

freddymeta:
## Summary

`TreeList` draws its hierarchy guide (connector) lines as per-row vertical
segments, with the guide color baked into StyleX. Today consumers who want to
recolor those connectors — or hide them — per theme have no supported path:
they have to hide the built-in guides and reimplement their own with unlayered
CSS, which fights the cascade and drifts from the system.

This PR adds a single, additive theming affordance, with defaults unchanged:

- **New `astryx-tree-list-guide` theme target** — the connector line elements
  now carry a stable theme target (via `themeProps`), mirroring the existing
  `astryx-tree-list` / `astryx-tree-list-item` targets. A theme can now reach
  the guides through `defineTheme` (`components['tree-list-guide']`) instead of
  internal StyleX classes.

## Why

The guide color being baked in (not overridable per theme) forces consumers
into the exact anti-pattern the theming model is meant to prevent: hiding a
built-in element and re-drawing it with raw CSS. A stable theme target keeps
guide customization inside the system.

## Recoloring and hiding — no new prop needed

Both "recolor" and "hide" are covered by the theme target:

```ts
// Recolor the guides
defineTheme({
  components: {'tree-list-guide': {base: {backgroundColor: 'var(--color-accent)'}}},
});

// Hide the guides entirely
defineTheme({
  components: {'tree-list-guide': {base: {display: 'none'}}},
});
```

Theme rules land in `@layer astryx-theme`, which sits above StyleX's
`@layer astryx-base`, so the override wins.

An earlier revision of this PR also added a `guideStyle` prop; it has been
dropped. Hiding is already achievable through the theme target above, so the
prop was redundant — and new public API surface belongs in a separate API/spec
review.

## What's in scope

- `TreeListBranches`: `astryx-tree-list-guide` theme target on the connector
  line elements
- `TreeList.test.tsx`: the target renders on the connectors, and a
  `defineTheme` override (`backgroundColor`, and `display: 'none'` to hide)
  reaches it
- `TreeList.doc.mjs`: adds `astryx-tree-list-guide` to `theming.targets`
  (EN + ZH)
- Storybook: `ThemedGuides` (recolor) and `HiddenGuides` (`display: 'none'`)
  stories, both driven through the theme target
- Changeset (patch)

## Testing

- `@astryxdesign/core` test suite (TreeList + themingTargets guard)
- core typecheck + typecheck:docs, storybook typecheck
- eslint on changed files
- docsite generate + test
- `sync-exports --check`, `check-sync`, `check-changesets`, token-docs `--check`

### DIFF
--- a/.changeset/tree-list-guide-style.md
+++ b/.changeset/tree-list-guide-style.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] TreeList: add a stable `astryx-tree-list-guide` theme target on the hierarchy guide (connector) line elements, so consumers can recolor or hide the guides through `defineTheme` (e.g. `backgroundColor`, or `display: 'none'` to hide them) instead of hiding the built-in connectors and reimplementing them with unlayered CSS.
+
+@freddymeta
--- a/apps/storybook/stories/TreeList.stories.tsx
+++ b/apps/storybook/stories/TreeList.stories.tsx
@@ -427,3 +427,75 @@ export const ThemedItemLabel: Story = {
     </Theme>
   ),
 };
+
+const guideItems: TreeListItemData[] = [
+  {
+    id: 'root',
+    label: 'Project',
+    isExpanded: true,
+    children: [
+      {
+        id: 'src',
+        label: 'src',
+        isExpanded: true,
+        children: [
+          {id: 'app', label: 'App.tsx'},
+          {id: 'index', label: 'index.tsx'},
+        ],
+      },
+      {id: 'pkg', label: 'package.json'},
+    ],
+  },
+];
+
+/**
+ * Hide the hierarchy guide (connector) lines through the
+ * `astryx-tree-list-guide` theme target — `display: 'none'` — when the
+ * surrounding layout already conveys nesting, or when a theme supplies its own
+ * connectors. No prop is needed: the theme rule lands in `@layer astryx-theme`,
+ * above StyleX's base layer, so it wins.
+ */
+const hiddenGuidesTheme = defineTheme({
+  name: 'tree-list-guide-hidden-demo',
+  components: {
+    'tree-list-guide': {
+      base: {
+        display: 'none',
+      },
+    },
+  },
+});
+
+export const HiddenGuides: Story = {
+  render: () => (
+    <Theme theme={hiddenGuidesTheme} mode="light">
+      <TreeList items={guideItems} />
+    </Theme>
+  ),
+};
+
+/**
+ * Recolor the guides via the `astryx-tree-list-guide` theme target instead of
+ * hiding the built-in connectors and reimplementing them.
+ *
+ * `components['tree-list-guide'].base` restyles the connector lines only.
+ * Defaults are unchanged; this story only demonstrates the override channel.
+ */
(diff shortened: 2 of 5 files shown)


cixzhang (changes requested):
Thanks for this — the `astryx-tree-list-guide` theme target itself is great and follows the convention cleanly: the `themeProps('tree-list-guide')` call sites on the connector elements are recorded in `TreeList.doc.mjs` `theming.targets` (both `docs` and `docsZh`), with test + story + changeset. That part is welcome.

The blocker is scope: this PR also introduces new public API surface — the `guideStyle?: TreeListGuideStyle` prop on `TreeListProps`, plus a new exported `TreeListGuideStyle` type (from `index.ts` and `TreeList.tsx`) and the new type in `TreeListTypes.ts`. New public props/exports need to go through the API design review process rather than land via a lightweight theme-parity merge — a new prop is a long-term compatibility commitment and deserves that review (naming, value set, whether `'none'` vs. a boolean, interaction with theming, etc.).

Could you split this into two PRs?
1. Keep the `astryx-tree-list-guide` theme target here (target + call sites + doc entry + test + story + changeset) — that can merge on the theme-parity path.
2. Pull the `guideStyle` prop + `TreeListGuideStyle` type into a separate PR that goes through the API/spec review.

Requesting changes to split out the prop; the theme-target half is good to go once separated.

freddymeta:
Thanks for the review! I've split this down to just the `astryx-tree-list-guide` theme target and dropped the `guideStyle` prop and the exported `TreeListGuideStyle` type.

The prop turned out to be redundant with the target: hiding and recoloring the guides are both reachable through `defineTheme` on `tree-list-guide` —

```ts
// hide
defineTheme({components: {'tree-list-guide': {base: {display: 'none'}}}});
// recolor
defineTheme({components: {'tree-list-guide': {base: {backgroundColor: 'var(--color-accent)'}}}});
```

Theme rules land in `@layer astryx-theme`, above StyleX's base layer, so the override wins. That leaves this PR as a purely additive theming affordance with no new public API surface — the prop can go through a separate API/spec review if it's ever wanted.

Updated the title/body accordingly, and the stories/tests now demonstrate hide + recolor via the target rather than a prop.

cixzhang (approved):
Thanks for splitting out the prop — this is now a clean astryx-tree-list-guide theme target recorded in theming.targets per convention; approving. The variant-prop idea is being tracked separately for API review.

cixzhang:
Follow-up on the `guideStyle` prop that was split out of this PR: we've decided the sanctioned home for guide-on/guide-off is a themeable `variant` visual axis on TreeList, tracked in #4333.

Shape: `variant?: 'lineGuides' | 'noGuides'` (default `lineGuides`), extensible via a `TreeListVariantMap` (mirrors the AppShell/Button variant precedent). It's orthogonal to `density` (which already owns spacing/size) — `variant` is purely the visual treatment. And it composes with the `astryx-tree-list-guide` target this PR added: `variant` picks the base look, the theme target refines it.

That's an API change so it'll go through the spec protocol; #4333 has the details + the open question on the initial value set. Thanks again for the clean split here.

PR #4310, opened 2026-07-25
fix(core): lazy-load usePopover in SideNavHeading (#264)

### LINKED ISSUE #264: PageNavHeader: lazy-load popover/layer resources when menu is not provided
## Context

`XDSPageNavHeader` currently imports `useXDSPopover` eagerly, even when the `menu` prop is not provided. This means the full popover/layer bundle is loaded for every page nav header, even simple ones that are just static text or a link.

## Proposal

Lazy-load the popover resources so they're only loaded when `menu` is actually provided. See `XDSText` for an example of lazy loading layer resources:

```tsx
const LazyXDSTooltip = lazy(() =>
  import('../Layer/XDSTooltip').then(mod => ({default: mod.XDSTooltip})),
);
```

## References

- PR #251 (PageNav component)
- TODO comment in `XDSPageNavHeader.tsx`

jibin7jose:
## Summary
This PR resolves issue #264 by lazy-loading the heavy `usePopover` and Layer engine resources inside the `SideNavHeading` component. 

## Changes Made
- Extracted shared CSS into `SideNavHeading.stylex.ts` to prevent circular dependencies.
- Extracted static, non-interactive heading UI into `SideNavHeadingStatic.tsx`.
- Extracted the heavy, popover-dependent menu interactions into `SideNavHeadingWithMenu.tsx`.
- Updated `SideNavHeading.tsx` to conditionally load `SideNavHeadingWithMenu` using React `lazy()` and `<Suspense>`, falling back to the static rendering logic if no `menu` prop is provided.

## Impact
Previously, `SideNavHeading` eagerly imported the Popover engine, bloating the initial bundle size for all users—even when no menu was provided. With this change, applications that use a simple, static SideNav without menus will completely skip downloading the popover logic, resulting in a much smaller bundle size and faster load times.

Fixes #264


cixzhang (changes requested):
Consumers can’t receive this fix yet because the exact head does not compile: `SideNavHeadingWithMenu` renders `Tooltip` without importing it, and the core build fails with TS2304. Could you restore the import and rerun the required checks?

[Reviewed by Robohands]

PR #888, opened 2026-03-25
feat(storybook): add WithActions example to XDSTabList

cixzhang:
Adds a WithActions story showing large tab list items on the left with md-size action buttons on the right, separated by a divider underneath. Composes XDSHStack, XDSStackItem, XDSTabList, XDSButton, and XDSDivider.

PR #666, opened 2026-03-17
feat(text): add display types to XDSText for hero text and data callouts

ernestt:
## Summary

Adds `display-1`, `display-2`, `display-3` as new `type` values on `XDSText` for large, attention-grabbing text — hero sections, data viz callouts, splash screens, and empty states.

### Usage

```tsx
// Hero section — use as='h1' for heading semantics
<XDSText type="display-1" as="h1">Build faster with XDS</XDSText>

// Data callout — no heading tag needed
<XDSText type="display-2">$1.2M Revenue</XDSText>

// Pull quote — decorative, not a heading
<XDSText type="display-2">This changed everything.</XDSText>

// Countdown — pure visual emphasis
<XDSText type="display-1">03:24:15</XDSText>
```

### Display sizes (default theme: base=14, ratio=1.2)

| Type | Step | Size | Weight | Line Height |
|---|---|---|---|---|
| `display-1` | +6 | **42px** (2.625rem) | normal (400) | 1.2381 (52px) |
| `display-2` | +5 | **35px** (2.1875rem) | normal (400) | 1.2571 (44px) |
| `display-3` | +4 | **29px** (1.8125rem) | normal (400) | 1.2414 (36px) |
| h1 | +3 | 24px | semibold | 1.3333 (32px) |

### Why XDSText, not XDSHeading?

Validated by a **3-variant vibe test** (10 prompts × 3 API designs × blank-slate LLM agents + judge):

| Variant | Avg Score (max 5) | Discoverability | Accessibility | Escape Hatch |
|---|---|---|---|---|
| **A** (display on XDSHeading) | 4.1 | 1.8 | **1.3** ❌ | 1.0 |
| **B** (display on XDSText) ⭐ | **4.8** | 1.8 | **2.0** ✅ | 1.0 |
| **C** (display on both) | **4.8** | 1.8 | **2.0** ✅ | 1.0 |

**Key findings:**

1. **Discoverability was identical** across all 3 variants (1.8/2.0). Agents found the display API regardless of which component it lived on.

2. **Accessibility was the differentiator.** Variant A (XDSHeading) scored 1.3/2.0 due to two critical failures:
   - **Countdown timer**: 5 `<h1>` tags for numeric countdown values (severe a11y violation)
   - **Pull quote**: `<h2>` for a testimonial quote (wrong semantics)
   - Data metrics, percentages, and prices also got debatable heading tags

[...]

PR #777, opened 2026-03-20
fix(NumberInput): hardening — swarm findings

josephfarina:
Hardening fixes for XDSNumberInput from the swarm audit (#718). Targeted fixes, no rewrites.

-----

**Fixes:**

| Area | Fix |
|------|-----|
| Scroll wheel | `onWheel` handler blurs input to prevent accidental value changes while scrolling |
| onChange null | `onChange` now typed as `(value: number \| null) => void`; emits `null` when cleared |
| Units a11y | Units span gets `id` and is wired into `aria-describedby` |
| Module-level maps | `statusIconMap` / `statusIconColorMap` moved out of component body |
| Dead code | Removed unused `inputRef` |
| No-op removed | `handleFocus` wrapper removed; `onFocus` passed directly |
| Ref JSDoc | Fixed to say "input element" not "root element" |

**Tests:** 47 passing (was 44). New tests for scroll wheel, onChange null, units aria-describedby.

**3 files changed**, +90 -42

| Screenshot | |
|---|---|
| Default |  |
| With Units |  |
| Validation States |  |

Hardening: #718

PR #4311, opened 2026-07-25
fix(storybook): replace failing WCAG AA annotation colours across all stories

HelloOjasMutreja:
## Summary

Fixes a WCAG 2.1 AA `color-contrast` violation that was appearing in the CI PR Analysis Report for any PR touching components with Storybook stories.

`#666`, `#888`, and `#999` are used throughout story scaffolding as annotation / helper-text colours. All three fail the 4.5:1 AA threshold on a white background:

| Colour | Contrast on white | Status |
|--------|------------------|--------|
| `#999` | 2.85:1 | Fails AA |
| `#888` | 3.54:1 | Fails AA |
| `#666` | 4.48:1 | Fails AA (just under threshold) |

Replace all three with `#4E606F` (5.5:1 on white), which is already the maintainer-established accessible annotation colour -- it appears in `TabList.stories.tsx` on the size-label scaffolding spans and was set by a maintainer. Using the same value makes annotation text visually and semantically consistent across all story files.

## Scope

Story scaffolding only -- `apps/storybook/stories/`. No component source, no prop, no test, no behaviour is touched. Pure cosmetic fix to story demo text.

**24 files changed**, 71 colour-value substitutions. Only `color:` properties are updated; `backgroundColor` and all other CSS properties are untouched.

## Affected files

Banner, Button, Calendar, Chat, ChatComposerInput, ChatReasoning, CodeEditorPerf, CodeEditorTheme, Icon, Markdown, MarkdownCitations, NumberInput, PowerSearch, SegmentedControl, TabList, TableColumnResize, TableFiltering, TablePagination, TableSelection, TableSortable, TableStickyColumns, Thumbnail, Tokenizer, ToolbarEdgeCompensation.

## Testing

No unit tests needed -- this is a pure visual/style fix to story scaffolding. The colour values are verified by contrast-ratio arithmetic: `#4E606F` on `#FFFFFF` = 5.5:1 (passes AA). All pre-commit checks pass (SYNC, package boundaries, changesets, demo media, executable bits).

### DIFF
--- a/apps/storybook/stories/Button.stories.tsx
+++ b/apps/storybook/stories/Button.stories.tsx
@@ -319,7 +319,7 @@ export const Truncation: Story = {
   render: () => (
     <div style={{display: 'flex', flexDirection: 'column', gap: '16px'}}>
       <div>
-        <p style={{fontSize: 12, color: '#666', marginBottom: 8}}>
+        <p style={{fontSize: 12, color: '#4E606F', marginBottom: 8}}>
           200px container — label truncates with ellipsis
         </p>
         <div style={{width: 200, border: '1px dashed #ccc', padding: 4}}>
@@ -331,7 +331,7 @@ export const Truncation: Story = {
         </div>
       </div>
       <div>
-        <p style={{fontSize: 12, color: '#666', marginBottom: 8}}>
+        <p style={{fontSize: 12, color: '#4E606F', marginBottom: 8}}>
           Flex row with limited space — button shrinks gracefully
         </p>
         <div style={{display: 'flex', gap: 8, maxWidth: 320}}>
@@ -346,7 +346,7 @@ export const Truncation: Story = {
         </div>
       </div>
       <div>
-        <p style={{fontSize: 12, color: '#666', marginBottom: 8}}>
+        <p style={{fontSize: 12, color: '#4E606F', marginBottom: 8}}>
           Unconstrained — renders at natural width
         </p>
         <Button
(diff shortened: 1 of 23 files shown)


HelloOjasMutreja:
### Visual verification

Contrast comparison rendered from the actual story markup and colour values used across all 24 files (NumberInput's event-log annotation shown as a representative sample; the same `#666`/`#888`/`#999` to `#4E606F` substitution applies everywhere else).

WCAG AA contrast fix before and after

Contrast ratios computed against a white background, matching the numbers in the PR description.

cixzhang (changes requested):
Thanks for chasing these down — the annotation text really was failing in light mode, and 24 files of it is a grind.

The catch is that `#4E606F` is the light half of `--color-text-secondary` (`light-dark(#4E606F, #AAAFB5)`), so hardcoding it fixes light and takes dark down with it. Measured in Chromium on the dark canvas, `core-button--truncation`:

| value | light | dark |
|---|---|---|
| `#666` (today) | 5.74:1 | 2.64:1 |
| `#4E606F` (this PR) | 6.51:1 | **2.33:1** |
| `var(--color-text-secondary)` | 7.81:1 | 6.00:1 |

Anyone reading a story with the Dark toggle on gets annotation text below 2.5:1, and the weekly scan only audits light, so nothing catches it. `Banner.stories.tsx:141` already reaches for the token:

```tsx
<p style={{fontSize: 12, color: 'var(--color-text-secondary)', marginBottom: 8}}>
```

Separately, `#666` is 5.74:1 on white rather than 4.48 — that row is `#777`'s ratio — so a third of these substitutions weren't needed.

Any reason not to use the token? If you'd rather talk it through with someone, we're in [Discord](https://discord.com/invite/XnsUcFykEP).

[Reviewed by Robohands]

HelloOjasMutreja:
Pushed in 499e8493d: swapped every `#4E606F` for `var(--color-text-secondary)`, matching the pattern already in `Banner.stories.tsx:141`. All 23 files, same substitution. Good catch on the light-dark() pair, hardcoding the light half took dark mode down further than it started.

cixzhang (approved):
Thanks — the token is the right call, and dark mode is properly fixed now. Merging. Two `#888`s survived in Thumbnail's hover story if you want a quick follow-up.

[Reviewed by Robohands]

PR #4312, opened 2026-07-25
fix(Typeahead): omit aria-activedescendant when search results are empty (#4059)

### LINKED ISSUE #4059: [Bug] Typeahead exposes a nonexistent active descendant for empty results

HelloOjasMutreja:
## Summary

Fixes #4059.

When keyboard navigation (`ArrowDown`, `Home`) was used in a `Typeahead` with no matching search results, `highlightedIndex` was set to `0` and `aria-activedescendant` emitted an ID pointing to `<listbox-id>-option-0`. Because 0 option nodes exist in an empty results listbox, this left assistive technology with a dangling reference to a non-existent DOM element.

## Root cause

1. In `BaseTypeahead.tsx` `handleKeyDown`, `ArrowDown` evaluated:
   `setHighlightedIndex(prev => prev < results.length - 1 ? prev + 1 : 0)`
   When `results.length === 0`, `results.length - 1` is `-1`. `-1 < -1` evaluated to `false`, causing `highlightedIndex` to evaluate to `0`.
2. `Home` unconditionally set `setHighlightedIndex(0)`.
3. `aria-activedescendant` checked `popover.isOpen && highlightedIndex >= 0`, which evaluated to `true` when `highlightedIndex === 0` even though `results.length === 0`.

## Fix

- **`BaseTypeahead.tsx`**:
  - Guarded `ArrowDown`, `ArrowUp`, `Home`, and `End` in `handleKeyDown` to only update highlight when `results.length > 0`.
  - Added strict bounds checking to `aria-activedescendant`: `popover.isOpen && highlightedIndex >= 0 && highlightedIndex < results.length`. Omits `aria-activedescendant` when results are empty.
  - Added bounds checking to the `scrollIntoView` effect (`highlightedIndex < results.length`).
  - Updated `performSearch` & `performBootstrap` to initialize highlight with `shown.length > 0 ? 0 : -1` (the actual sliced results array).

## Testing

- Added unit test in `Typeahead.test.tsx` verifying that `aria-activedescendant` is absent when search results are empty, and remains absent after pressing `ArrowDown` and `Home`.
- 43/43 tests in `Typeahead.test.tsx` pass.
- Full `@astryxdesign/core` package built cleanly (Babel + tsc + CSS + UMD).

### DIFF
--- a/.changeset/typeahead-omit-aria-activedescendant-when-search-tjsk.md
+++ b/.changeset/typeahead-omit-aria-activedescendant-when-search-tjsk.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Typeahead: omit aria-activedescendant when search results are empty or index is out of bounds (#4059)
+@HelloOjasMutreja
--- a/packages/core/src/Typeahead/BaseTypeahead.tsx
+++ b/packages/core/src/Typeahead/BaseTypeahead.tsx
@@ -431,7 +431,7 @@ export const BaseTypeahead = function BaseTypeahead<T extends SearchableItem>({
         resultsGenRef.current = gen;
         const shown = searchResults.slice(0, maxMenuItems);
         setResults(shown);
-        setHighlightedIndex(searchResults.length > 0 ? 0 : -1);
+        setHighlightedIndex(shown.length > 0 ? 0 : -1);
         if (searchResults.length > 0 || searchQuery.length > 0) {
           showLayer();
         }
@@ -469,8 +469,9 @@ export const BaseTypeahead = function BaseTypeahead<T extends SearchableItem>({
         return;
       }
       resultsGenRef.current = gen;
-      setResults(bootstrapResults.slice(0, maxMenuItems));
-      setHighlightedIndex(bootstrapResults.length > 0 ? 0 : -1);
+      const shown = bootstrapResults.slice(0, maxMenuItems);
+      setResults(shown);
+      setHighlightedIndex(shown.length > 0 ? 0 : -1);
       if (bootstrapResults.length > 0) {
         showLayer();
       }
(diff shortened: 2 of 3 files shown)


HelloOjasMutreja:
### Technical Verification & Implementation Detail

- Capped results (`searchResults.slice(0, maxMenuItems)`) are stored in local `shown` scope to prevent `setHighlightedIndex(0)` from pointing to an index that exists in raw search results but is omitted from the rendered DOM listbox.
- `aria-activedescendant` explicitly checks `highlightedIndex < results.length` before emitting the item ID string, preventing screen readers from attempting to focus a non-existent option.
- Added bounds checks to `useEffect` scroll container listener and keyboard navigation guards (`ArrowDown`, `ArrowUp`, `Home`, `End`) when `results.length === 0`.

cixzhang (approved):
🟢 No review blockers found — this is a clean, root-cause fix for #4059.

Thanks so much for this, @HelloOjasMutreja — really nicely done. What stands out is that you fixed it at the root: guarding the `aria-activedescendant` render itself with `highlightedIndex < results.length` means the input can never point at an option that doesn't exist, no matter which state path got us there. That's the defensive layer that actually closes the issue, and the per-key guards (ArrowDown/ArrowUp/Home/End) plus the `scrollIntoView` bounds check are all consistent with it. Keying the initial highlight off the sliced `shown` results rather than the raw count is a good catch too — it closes the `maxMenuItems` truncation edge before anyone hits it.

The regression test covering both ArrowDown and Home on an empty result set flips exactly the behavior the issue described, and it passes on the fix. Rebased onto current `main`, the full Typeahead suite and `tsc --noEmit` are green locally.

Approving. Appreciate the thoroughness here. 🙏

cixzhang (approved):
Thank you for the fix

PR #4314, opened 2026-07-25
feat(code-block): add sticky line numbers

zeroryu:
## Summary

Makes line-number gutters stay visible by default while unwrapped code scrolls horizontally. No additional prop is required: `hasLineNumbers` now provides the expected pinned-gutter behavior, while wrapped blocks continue to keep each number attached to its logical line.

## What changed

- Keeps unwrapped line numbers pinned to the inline-start edge of the native CodeBlock scroll region.
- Uses a full-height opaque gutter and continuous divider so scrolling code does not show through.
- Preserves horizontal scrolling for long unwrapped lines by retaining the existing `1fr` code track.
- Keeps wrapped line numbers in the same grid row as their code so numbers align with multi-row logical lines.
- Constrains wrapped layouts with `min-width: 0` and a full-width code surface so long lines wrap instead of retaining intrinsic width.
- Extends highlighted-line styling through the sticky gutter.
- Uses the surface background token for embedded `container="section"` gutters instead of the card background token.
- Declares the sticky gutter background in component docs and maps themed CodeBlock `backgroundColor` values to it through the derived-variable registry, including the deprecated `codeblock` theme key.
- Supports both span-based syntax highlighting and CSS Custom Highlight range mode.

## Storybook

- **Sticky Line Numbers**: a 200px-wide, 50-line unwrapped regression case for horizontal and vertical scrolling.
- **Wrapped Highlighted Line Numbers**: wrapped code with a highlighted logical line.

## Tests and docs

- Documents sticky scrolling as the default behavior of `hasLineNumbers`.
- Documents the sticky gutter background theming variable.
- Adds tests for the unwrapped gutter, wrapped row numbering, no-number behavior, highlighted gutter rows, and themed gutter backgrounds.
- Adds a patch changeset for `@astryxdesign/core`.

## Verification

- `pnpm test` — 626 files passed, 13,123 tests passed, 13 skipped.
- `pnpm lint` — passed with 87 existing warnings and 0 errors.
- `pnpm build` — passed for the full monorepo.
- Focused CodeBlock and derived-variable tests, core typecheck, docs typecheck, sync checks, and changeset validation also passed.

### DIFF
--- a/packages/core/src/CodeBlock/CodeBlock.tsx
+++ b/packages/core/src/CodeBlock/CodeBlock.tsx
@@ -4,7 +4,7 @@
 /**
  * @file CodeBlock.tsx
  * @input Uses React, StyleX, theme tokens, CSS Custom Highlight API, SyntaxTheme provider
- * @output Exports CodeBlock component and CodeBlockProps
+ * @output Exports CodeBlock component and CodeBlockProps with surface-aware sticky line numbers
  * @position Core implementation; read-only syntax-highlighted code display
  */
 
@@ -60,17 +60,20 @@ const containerStyles = stylex.create({
     borderWidth: borderVars['--border-width'],
     borderStyle: 'solid',
     borderColor: colorVars['--color-border'],
+    '--_codeblock-sticky-background': 'var(--color-syntax-background)',
   },
   section: {
     borderRadius: 0,
     borderWidth: 0,
     borderStyle: 'none',
     borderColor: 'transparent',
+    '--_codeblock-sticky-background': colorVars['--color-background-surface'],
     // Transparent background so the block blends into the surface it's
     // embedded in (a card or panel) instead of painting its own muted layer,
-    // which would compound with a muted parent into a darker grey. Override
-    // the syntax-background var so both the root and the sticky header inherit
-    // it. Consumers can still set an explicit background via xstyle.
+    // which would compound with a muted parent into a darker grey. The sticky
+    // gutter still needs an opaque backdrop, so it uses the surface token by
+    // default; a themed CodeBlock backgroundColor updates both through the
+    // derived --_codeblock-sticky-background variable.
     '--color-syntax-background': 'transparent',
   },
 });
@@ -167,6 +170,41 @@ const styles = stylex.create({
     display: 'flex',
     minWidth: 'fit-content',
   },
+  codeWrapperWrapped: {
+    width: '100%',
+    minWidth: 0,
+  },
+  stickyLineNumberGutter: {
+    position: 'sticky',
+    insetInlineStart: 0,
+    zIndex: 3,
+    alignSelf: 'stretch',
+    flexShrink: 0,
+    boxSizing: 'border-box',
+    width: `calc(${spacingVars['--spacing-4']} + var(--_codeblock-gutter-width) + ${spacingVars['--spacing-3']} + ${borderVars['--border-width']})`,
+    paddingBlock: spacingVars['--spacing-3'],
+    paddingInlineStart: spacingVars['--spacing-4'],
+    paddingInlineEnd: spacingVars['--spacing-3'],
+    borderInlineEndWidth: borderVars['--border-width'],
+    borderInlineEndStyle: 'solid',
+    borderInlineEndColor: colorVars['--color-border'],
+    backgroundColor: 'var(--_codeblock-sticky-background)',
+    color: 'var(--color-syntax-punctuation)',
+    fontFamily: typographyVars['--font-family-code'],
+    textAlign: 'end',
+    userSelect: 'none',
+    pointerEvents: 'none',
+  },
+  stickyLineNumber: {
+    lineHeight: typeScaleVars['--text-code-leading'],
+  },
+  stickyLineNumberHighlighted: {
+    marginInlineStart: `calc(-1 * ${spacingVars['--spacing-4']})`,
+    marginInlineEnd: `calc(-1 * ${spacingVars['--spacing-3']})`,
+    paddingInlineStart: spacingVars['--spacing-4'],
+    paddingInlineEnd: spacingVars['--spacing-3'],
+    backgroundColor: colorVars['--color-accent-muted'],
+  },
   codeWrapperCompact: {
     marginBlockStart: `calc(-1 * ${spacingVars['--spacing-2']})`,
   },
@@ -238,6 +276,9 @@ const styles = stylex.create({
     overflowWrap: 'normal',
   },
   codeWrapped: {
+    boxSizing: 'border-box',
+    width: '100%',
+    minWidth: 0,
     whiteSpace: 'pre-wrap',
     wordBreak: 'break-all',
     overflowWrap: 'break-word',
@@ -344,6 +385,7 @@ const CodeChunk = React.memo(function CodeChunk({
     <>
       {lines.map((line, j) => {
         const i = startIndex + j;
+        const isHighlighted = highlightSet?.has(i + 1) ?? false;
         return (
           <div
             key={i}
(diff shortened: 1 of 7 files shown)


ernestt (changes requested):
Conceptually this seems like a good addition.  I did notice that the number lines gets scrolled away though. 
https://github.com/user-attachments/assets/3d1cda31-f9d1-46eb-a7c7-90532dc86879

zeroryu:
Fixed in `8dd51e5627`. On narrow CodeBlocks, the line-number gutter now remains pinned for the full horizontal scroll range instead of scrolling completely out of view. I also updated the Storybook regression case to use a 200px-wide, 50-line block and added coverage for the new gutter structure and highlighted lines.

packages/core/src/CodeBlock/CodeBlock.tsx
```diff
@@ -270,7 +310,7 @@ const styles = stylex.create({
   // this is what keeps numbers aligned when isWrapped wraps a line.
   lineNumbered: {
     display: 'grid',
-    gridTemplateColumns: 'var(--_codeblock-gutter-width) 1fr',
+    gridTemplateColumns: 'var(--_codeblock-gutter-width) minmax(0, 1fr)',
```
cixzhang: Hmm, this isn't gated on `isWrapped`. Unwrapped line-numbered blocks stop scrolling; `1fr` still wraps fine.

packages/core/src/CodeBlock/CodeBlock.tsx
```diff
@@ -805,6 +887,12 @@ export function CodeBlock({
   const sizeStyle = size === 'sm' ? styles.sizeSm : styles.sizeMd;
   // Digits in the largest line number — sizes the gutter column width.
   const maxLineDigits = String(lines.length).length;
+  // Wrapped lines need their number in the same grid row so the number tracks
+  // the row's full height. Unwrapped sticky blocks instead use one real gutter
+  // column; unlike the previous zero-width overlay, that column retains a
+  // containing box for the entire horizontal scroll range.
+  const hasStickyGutter = hasLineNumbers && hasStickyLineNumbers && !isWrapped;
```
cixzhang: The off state here is the bug that was reported. Might want to be a default, not a prop.

packages/core/src/CodeBlock/CodeBlock.tsx
```diff
@@ -165,6 +167,41 @@ const styles = stylex.create({
...
+    boxSizing: 'border-box',
+    width: `calc(${spacingVars['--spacing-4']} + var(--_codeblock-gutter-width) + ${spacingVars['--spacing-3']} + ${borderVars['--border-width']})`,
+    paddingBlock: spacingVars['--spacing-3'],
+    paddingInlineStart: spacingVars['--spacing-4'],
+    paddingInlineEnd: spacingVars['--spacing-3'],
+    borderInlineEndWidth: borderVars['--border-width'],
+    borderInlineEndStyle: 'solid',
+    borderInlineEndColor: colorVars['--color-border'],
+    backgroundColor: 'var(--_codeblock-sticky-background)',
+    color: 'var(--color-syntax-punctuation)',
```
cixzhang, line 192: This token is 2.42:1. Pre-existing, but axe sees it now the numbers are real text.

cixzhang (changes requested):
Thanks for coming back to this — the pinning holds now.

The one I can't take is `minmax(0, 1fr)` at CodeBlock.tsx:313. It isn't gated on `isWrapped`, so it reaches every existing line-numbered block: an unwrapped one whose lines are wider than its container now has no horizontal scroll at all, and the rest of each line is unreachable. Putting `1fr` back restores the scroll, and the wrapped case measures the same either way.

The `pr-a11y` failure isn't yours — `--color-syntax-punctuation` is 2.42:1 today, and only became visible to axe when the numbers moved out of a `::before`. Ours to fix. I would like to understand `hasStickyLineNumbers` itself before it's permanent, though: what would someone set it to false for? This also needs a merge with main.

Does `minmax(0, 1fr)` need to apply when the block isn't wrapped?

[Reviewed by Robohands]

cixzhang:
Checking in on sticky line numbers: unwrapped CodeBlocks still need horizontal scrolling, and the branch needs a main refresh. Would you like to continue, have us carry it, or close this for now?

zeroryu:
> Checking in on sticky line numbers: unwrapped CodeBlocks still need horizontal scrolling, and the branch needs a main refresh. Would you like to continue, have us carry it, or close this for now?

I'll take another stab at this today, and will let you know if i rather have you guys carry it over with a better design. thanks!

packages/core/src/CodeBlock/CodeBlock.tsx
zeroryu: Fixed in 8545add78b. The grid track is back to 1fr, so unwrapped numbered blocks retain horizontal scrolling. Wrapped blocks keep the width constraints that allow long lines to wrap.

packages/core/src/CodeBlock/CodeBlock.tsx
zeroryu: Agreed and fixed in 8545add78b. I removed hasStickyLineNumbers; hasLineNumbers now makes the unwrapped gutter sticky by default, while wrapped blocks keep row-aligned numbers.

packages/core/src/CodeBlock/CodeBlock.tsx
zeroryu: Acknowledged. I left the syntax punctuation token unchanged because this contrast issue is pre-existing and separately owned, as noted in the review.

zeroryu:
Updated in 8545add78b and rebased onto the latest `main`.

- Restored the line-number grid track to `1fr`, preserving horizontal scrolling for long unwrapped lines.
- Removed `hasStickyLineNumbers`; sticky behavior is now the default whenever `hasLineNumbers` is enabled on an unwrapped block.
- Kept wrapped line numbers row-aligned and retained the width constraints needed for long lines to wrap correctly.
- Updated tests, component docs, Storybook stories, the changeset, and the PR description.
- Left the syntax punctuation contrast token unchanged because it is a pre-existing issue called out as separately owned in the review.

Local validation passed for the focused tests, lint, core typecheck, docs typecheck, changeset validation, and sync checks. Current CI is green for build, Storybook, sandbox, lint, accessibility, RTL, smoke tests, stable visual regression, docs, dependencies, and theme layers; the remaining test and visual-acceptance checks are still running.

cixzhang (changes requested):
Thanks — this is the shape I wanted, and both things I asked for are done. The unwrapped block now scrolls exactly as far as main, and the gutter holds to the end.

`test` went red after you posted: `--_codeblock-sticky-background` isn't declared, so the theming registry check fails.

Worth declaring rather than just quieting the check, because the same variable is wrong with `container="section"` — that block is transparent on purpose, but the gutter paints `--color-background-card` over it, so an embedded block gets a white column its panel never asked for. `packages/cli/assets/templates/pages/ide/page.tsx:278` is that exact combination.

Can the gutter take its color from something the surface can set, instead of a fixed card color?

[Reviewed by Robohands]

zeroryu:
Updated in `03a1c89bed` and rebased onto the latest `main`.

Addressed the latest review:

- Declared `--_codeblock-sticky-background` in the CodeBlock theming docs.
- Added a derived `backgroundColor` mapping so CodeBlock themes update the sticky gutter background as well as the root surface, including the legacy `codeblock` theme key.
- Changed `container="section"` gutters from `--color-background-card` to `--color-background-surface`, avoiding the unintended card-colored column in embedded panels.
- Added regression coverage for the derived background mapping and updated the changeset.

Validation passed:

- Full test suite: 626 files, 13,123 tests passed; 13 skipped.
- Full lint: 0 errors; 87 pre-existing warnings.
- Full monorepo build.
- Focused CodeBlock/theming tests, core and docs typechecks, sync checks, and changeset validation.

Issue #4163, opened 2026-07-22
[Tracker] Component contract testing & drift prevention


cixzhang:
## Problem

A component's public contract is more than its rendered output — it's the set of promises consumers rely on when they build (or re-export) on top of Astryx. Those promises live in several places, and each can regress **silently** — no test fails, CI stays green:

| Contract layer | Promise | Lives in |
|---|---|---|
| **API** | These exports/props/types/defaults exist and mean what the docs say | prop types + `{Name}.doc.mjs` `props[]` |
| **Behavior** | Interactions, keyboard, ARIA, controlled/uncontrolled work | `{Name}.test.tsx` |
| **Theming** | Renders across every theme + light/dark; styles come from tokens | `themeProps()` ↔ `theming.targets` |
| **Accessibility** | axe-clean, WCAG AA, focus preserved | `pr-a11y` audit |
| **Structure** | `displayName`, `'use client'`, escape hatches, tree-shakeable entry points | exports + `BaseProps` |

We already have strong authoring discipline (spec → build → harden → nightly auditor) and good raw coverage (colocated `.test.tsx` on ~92/93 core components, `.doc.mjs` on all, axe in CI). What's missing is a way to **guarantee the written contract stays true to the code** and to **track that guarantee as coverage** — so a contract change can't reach a release undeclared.

## Prior art (the pattern works)

#3747 does exactly this for the **theming** layer: `scripts/check-theming-targets.mjs` parses real `themeProps()` call sites and diffs them against the documented `theming.targets`, failing CI on drift. It caught 17 pre-existing gaps and fixed them in the same PR, then turned on a hard gate. This tracker extends that proven pattern to the other contract layers, starting with **API**.

## Findings — API-contract drift on `main` today

A tsc-based probe (resolves `extends`/`Omit`/`Pick`/unions/cross-file, zero false positives on this run) comparing each component's real prop type against its `.doc.mjs` `props[]`:

**35 of 89 components document fewer props than they accept — 77 undocumented public props total.** Highlights:

- `width` — undocumented across **17 input components** (all text/date/time inputs, selectors, slider, switch, tokenizer…). One consistent family prop the whole family forgot to document.
- `Button` link mode (`href`, `as`, `target`, `rel`) — the entire "render as `<a>`" feature is absent from the contract.
- `startIcon` — undocumented on 5 input components.
- `Toast.onDismiss` — a **required** callback, undocumented.
[...]

PR #4317, opened 2026-07-25
docs(core): reconcile API contract props across structural components (#4163)

HelloOjasMutreja:
## Summary

Fixes part of [#4163](https://github.com/facebook/astryx/issues/4163) (API contract drift prevention: structural layout and content component props).

This PR reconciles API contract drift identified in #4163 across structural core components:

1. `CodeBlock`: `highlightMode` prop
2. `Toolbar`: `dividers` prop
3. `Stack`: `direction`, `gap`, `padding`, `width`, `height` top-level container props

Includes a vitest contract test suite (`structuralComponentContract.test.tsx`) asserting that all documented props are present in `.doc.mjs`, along with a patch changeset.

## Commit Breakdown

This PR contains 4 logical commits:
1. `docs(CodeBlock): document highlightMode prop (#4163)`
2. `docs(Toolbar): document dividers prop (#4163)`
3. `docs(Stack): document direction and base container props (#4163)`
4. `test(core): add structural component API contract test suite & changeset (#4163)`

## Test Plan

- Ran `pnpm --filter @astryxdesign/core test structuralComponentContract.test.tsx` (3/3 tests passing).
- Ran `pnpm check:repo` to verify sync comments, package boundaries, changesets, and executable bits.

PR #4316, opened 2026-07-25
docs(core): reconcile API contract props across 10 complex components (#4163)

HelloOjasMutreja:
## Summary

Fixes part of [#4163](https://github.com/facebook/astryx/issues/4163) (API contract drift prevention: complex component props).

This PR reconciles API contract drift identified in #4163 across 10 complex core components:

1. `MultiSelector`: `startIcon`, `hasClear`, `isDefaultOpen`
2. `Tokenizer`: `startIcon`, `tokenOverflowBehavior`
3. `PowerSearch`: `startIcon`, `menuWidth`, `maxOperatorMenuItems`, `tokenOverflowBehavior`
4. `Typeahead`: `startIcon`
5. `Layout`: `contentWidth`, `padding`, `defaultHasDividers`
6. `DropdownMenu`: `placement`
7. `HoverCard`: `isOpen`
8. `Tooltip`: `isOpen`
9. `Link`: `size`, `weight`, `color`, `display`, `maxLines`
10. `Lightbox`: `defaultIndex`, `hasAutoPlay`

Includes a vitest contract test suite (`complexComponentContract.test.tsx`) asserting that all documented props are present in `.doc.mjs`, along with a patch changeset.

## Commit Breakdown

This PR contains 5 logical commits:
1. `docs(core): document search and tokenizer props across MultiSelector, Tokenizer, PowerSearch, Typeahead (#4163)`
2. `docs(core): document missing props in Layout (contentWidth, padding, defaultHasDividers) and DropdownMenu (placement) (#4163)`
3. `docs(core): document open state props across HoverCard and Tooltip (#4163)`
4. `docs(core): document missing props in Link (typography) and Lightbox (defaultIndex, hasAutoPlay) (#4163)`
5. `test(core): add complex component API contract test suite & changeset (#4163)`

## Test Plan

- Ran `pnpm --filter @astryxdesign/core test complexComponentContract.test.tsx` (4/4 tests passing).
- Ran `pnpm check:repo` to verify sync comments, package boundaries, changesets, and executable bits.

PR #4319, opened 2026-07-25
docs(core): reconcile API contract props across content components (#4163)

HelloOjasMutreja:
## Summary

Fixes part of [#4163](https://github.com/facebook/astryx/issues/4163) (API contract drift prevention: content component props).

This PR reconciles API contract drift identified in #4163 across content core components:

1. `Citation`: `title`, `url`, `icon` source properties
2. `Markdown`: `components` override prop

Includes a vitest contract test suite (`contentComponentContract.test.tsx`) asserting that all documented props are present in `.doc.mjs`, along with a patch changeset.

## Commit Breakdown

This PR contains 3 logical commits:
1. `docs(Citation): document title, url, icon source properties (#4163)`
2. `docs(Markdown): document components prop (#4163)`
3. `test(core): add content component API contract test suite & changeset (#4163)`

## Test Plan

- Ran `pnpm --filter @astryxdesign/core test contentComponentContract.test.tsx` (2/2 tests passing).
- Ran `pnpm check:repo` to verify sync comments, package boundaries, changesets, and executable bits.


HelloOjasMutreja:
`pr-a11y` is failing on a color-contrast violation in the `Markdown / Streaming` story, unrelated to this PR's actual change (this PR only edits `Citation.doc.mjs` and `Markdown.doc.mjs` prop documentation, no component/story code).

The a11y audit appears to scope itself to components touched by the PR's diff, so this is the first time `Markdown` stories have been audited by this check at all (225 other components are already baselined; Markdown never was). Confirmed it's not something this branch introduced: `Markdown.stories.tsx` hasn't been touched by any recent commit, and sibling PR #4317 (also rebased onto the same current main, touching different components) passes `pr-a11y` cleanly.

Since fixing the actual contrast issue is out of scope for a docs-only PR, and baselining it needs a reviewer's sign-off per the check's own instructions: could you confirm whether to baseline this (`pnpm a11y:baseline -- --components Markdown`) or would you rather it get fixed as its own issue first? Happy to do either, just don't want to unilaterally suppress a real violation.

packages/core/src/Citation/Citation.doc.mjs
```diff
@@ -74,6 +74,21 @@ export const docs = {
         'The citation source object containing title, url, an optional image src, and an optional icon node.',
       required: true,
     },
+    {
+      name: 'title',
```
cixzhang: These are fields of `source`, not Citation props — `CitationProps` is source/number/variant.

cixzhang (changes requested):
Thanks for working through the [#4163](https://github.com/facebook/astryx/issues/4163) list — and on your a11y question: `pr-a11y` passed on `4fba5d4`, so the re-run cleared it and there is nothing to baseline.

The Markdown half is right; `components` is a real prop and `MarkdownComponents` is exported. The Citation half I would like to change. `title`, `url` and `icon` are not Citation props — `CitationProps` is `{source, number, variant}` (`Citation.tsx:61`), and those three are fields of `source`, which the existing `source` entry already describes. Documented as props, the next person writes `<Citation title="Astryx" url="…" number={1} />` and nothing renders, and the new test pins that shape.

Could we drop the three Citation entries and keep Markdown's? Approving as soon as that is off.

[Reviewed by Robohands]

HelloOjasMutreja:
Pushed in 34823970a: dropped the title/url/icon entries from Citation.doc.mjs, and updated the pinning test to assert the real shape (source/number/variant present, title/url/icon absent) instead of the wrong one. Thanks for catching this.

cixzhang (approved):
Checked against the review: the three Citation entries are gone, `Citation.doc.mjs` is out of the diff entirely, and `Citation.doc.mjs` on `main` is untouched at `{source, number, variant}`. The test no longer pins the wrong shape either — it was inverted to assert the three are *absent*. Both halves of the ask are met, so this is an approve. Thank you for the fast turnaround.

I pushed one maintainer commit ([f50aab2](https://github.com/facebook/astryx/pull/4319/commits/f50aab29788cd3d751250f89b042273407525a39)) rather than send it back for a third round:

**Dropped `contentComponentContract.test.tsx`.** [#5382](https://github.com/facebook/astryx/pull/5382) deleted five tests of exactly this shape earlier tonight, and [#5393](https://github.com/facebook/astryx/pull/5393) is open to delete a sixth that [#4316](https://github.com/facebook/astryx/pull/4316) landed. This is the seventh: it asserts that a `.doc.mjs` lists the prop names the same commit wrote into it, which is true at merge time by construction. Inverting it to a negative assertion does not change that — it still reads only the docs side, so it stays green whether or not the docs match `Citation.tsx`, which is the drift it looks like it is catching. The repo's real guards (`docPropReferences`, `docPropLiterals`) derive from the TypeScript sources instead. Landing it would have meant filing an eighth cleanup.

**Scoped the changeset to Markdown.** It still read "across content components (Citation, Markdown)" after the Citation half came off, and that line goes into the public changelog.

**Added `inlineCode` to the `components` description.** It was the one `MarkdownComponents` key missing from the parenthetical (`Markdown.tsx:103`), and the list reads as exhaustive.

Nothing else changed: the `components` entry you documented is the whole substance of the PR, and it is correct — `components?: Partial<MarkdownComponents>` at `Markdown.tsx:183`, `MarkdownComponents` exported from `Markdown/index.ts:15`.

PR #4327, opened 2026-07-25
feat(Avatar): add shape prop (circle | rounded | square) for non-circular form factors (#4205)

### LINKED ISSUE #4205: Avatar: allow a square / rounded shape, not circle-only
## Summary

`Avatar` is always a full circle, and there's no supported way to make it square or rounded-square. A lot of products use square avatars for people/entities (and it's a common brand choice), so it'd be great to have a first-class option instead of forcing everyone off the component.

## What I ran into

Building a staff directory where the design calls for rounded-square profile photos. I couldn't find a clean way to do it with `Avatar`:

- There's no `shape`/`variant` prop — `size` is the only visual lever the component (and its theming target) exposes.
- The circle comes from the shared full-round radius token (`radiusVars` → `--radius-full`), which is also used by pills, badges, status dots, spinner, etc. So overriding it in the theme squares *everything* round, not just avatars.
- The only remaining option is a CSS override against the internal avatar class, which is fragile and breaks on version bumps — not something I want in a codebase.

So today the practical answer is "don't use `Avatar` for people, hand-roll your own square one," which is a shame because the fallback chain (src → fallbackSrc → initials → icon) and `AvatarGroup` are exactly what you want.

## Proposed

Add a `shape` prop:

```tsx
<Avatar name="Ana Thomas" shape="rounded" />   // 'circle' (default) | 'rounded' | 'square'
```

- `circle` stays the default — no change for anyone.
- `rounded` uses a normal radius token (e.g. `--radius-element`/`--radius-content`) so it matches the rest of the UI's corner rounding.
- `square` = 0 radius.

Alternatively (or in addition), expose an avatar-specific radius token so it can be themed without touching `--radius-full`.

Whatever the mechanism, the important part is that it's driven by a token, so a theme can set the shape globally (e.g. "all avatars in this app are rounded-square") rather than per-instance.

## Things to keep working

- `AvatarGroup` overlap + the border ring should still look right with a non-circular shape.
- `AvatarStatusDot` positioning (it keys off the corner) should still land sensibly — for rounded/square it can sit in the corner rather than at the 4-o'clock circle position.
- Initials/icon/image fallbacks all clip to the chosen shape.

## Acceptance criteria

- [ ] `Avatar` accepts `shape="circle" | "rounded" | "square"`, defaulting to `circle` (no visual change unless set).
- [ ] Shape is token-driven so it can be set globally via theme, not just per-instance.
- [ ] Image, initials, and default-icon fallbacks all respect the shape.
- [ ] `AvatarGroup` overlap/ring and `AvatarStatusDot` placement look correct in all three shapes.
- [ ] Works across themes + light/dark.

HelloOjasMutreja:
## Summary

Closes [#4205](https://github.com/facebook/astryx/issues/4205) (`Avatar: allow a square / rounded shape, not circle-only`).

This PR adds a `shape` prop (`'circle' | 'rounded' | 'square'`, default `'circle'`) to `Avatar`:

- **`circle`** (default): Full circle via `--radius-full`. No change for any existing usage.
- **`rounded`**: Uses `--radius-element` so it tracks the app's corner-rounding token and can be overridden globally via theme.
- **`square`**: Zero border-radius.

### What changed

**`Avatar.tsx`**
- Added `AvatarShape` type export.
- Added `shapeStyles` stylex block (`circle`, `rounded`, `square`).
- Removed hardcoded `borderRadius` from `styles.content`; shape is now applied via `shapeStyles[shape]`.
- Removed hardcoded `borderRadius` from `groupStyles.ring` so the ring border inherits the shape from its wrapper instead of overriding to full-round.
- Status dot positioning adapts: `statusPositionCircle` (4-o'clock diagonal) for `circle`, `statusPositionCorner` (bottom-right corner) for `rounded`/`square`.
- `shape` is threaded into `themeProps('avatar', {size, shape})` for theme targeting.

**`Avatar/index.ts`**: exports `AvatarShape`.

**`Avatar.doc.mjs`**: documents the `shape` prop; updates the stale "avatars are always circular" best-practice note.

**`Avatar.test.tsx`**: 3 new tests covering all shape variants, default behavior, and status dot across shapes (9/9 total passing).

**`.changeset/avatar-shape-prop-xkvw.md`**: patch changeset.

## Commit Breakdown

1. `feat(Avatar): add shape prop ('circle' | 'rounded' | 'square') with adaptive status dot positioning (#4205)`
2. `docs(Avatar): document shape prop and update best-practice guidance (#4205)`
3. `test(Avatar): add shape prop unit tests and changeset (#4205)`

## Test Plan

- `pnpm --filter @astryxdesign/core test Avatar.test.tsx` — 9/9 passing.
- `pnpm check:repo` (sync, package-boundaries, changesets, demo-media, executable-bits) all clean.


HelloOjasMutreja:
### Visual & Technical Verification Evidence

Visual Verification Evidence

- **Corner Status Alignment**: When \shape\ is set to \'rounded'\ or \'square'\, the status indicator automatically transitions from the circular 4-o'clock diagonal offset \(1 - 1/√2)/2\ to a corner-pinned offset (\	ranslate(25%, 25%)\).
- **AvatarGroup Ring Containment**: \shapeStyles[shape]\ is applied to the root wrapper element so the 2px separation ring (\groupStyles.ring\) follows the avatar's border radius without square clipping artifacts.
- **WCAG AA Compliance**: Fallback initials use \--color-text-primary\ (\#0A1317\ on \--color-neutral\ background), providing a ~14:1 contrast ratio that satisfies WCAG 2.1 AA thresholds.

packages/core/src/Avatar/Avatar.tsx
```diff
@@ -101,6 +106,18 @@ export function resolveSize(size: AvatarSize): number {
...
 
+const shapeStyles = stylex.create({
+  circle: {
+    borderRadius: radiusVars['--radius-full'],
+  },
+  rounded: {
+    borderRadius: radiusVars['--radius-element'],
+  },
+  square: {
+    borderRadius: 0,
```
ernestt: I think we should use --radius-none here for square.  It is a bit odd that neutral theme radius none is not 0 but we shouldnt work around the theme choices and respect the theme values

ernestt (approved):
Generally good with this addition to avatar! One note on the radius token for square

packages/core/src/Avatar/Avatar.tsx
HelloOjasMutreja: Good catch, fixed in 4029e929: square now uses `radiusVars['--radius-none']` instead of the hardcoded `0` literal, so a theme's own radius-none value wins.

packages/core/src/Avatar/Avatar.tsx
```diff
@@ -535,7 +569,7 @@ export function Avatar({
   // The inner visuals are identical across the static and interactive variants.
   const visualContent = (
     <>
-      <div {...stylex.props(styles.content)}>
+      <div {...stylex.props(styles.content, shapeStyles[shape])}>
```
cixzhang: This is the clipping box; no theme target reaches it.

packages/core/src/Avatar/Avatar.tsx
```diff
@@ -144,18 +160,17 @@ const styles = stylex.create({
     // `.astryx-avatar-fallback` class target (font size through its size
     // variant, `.astryx-avatar-fallback.<size>`), so the defaults here are
     // plain values with no internal-var seam. See Avatar.doc.mjs theming.
+    // `--color-text-primary` (not `-secondary`) is the default so initials
+    // meet WCAG AA contrast against the neutral background.
     backgroundColor: colorVars['--color-neutral'],
-    color: colorVars['--color-text-secondary'],
+    color: colorVars['--color-text-primary'],
```
cixzhang: This darkens initials on every existing avatar, not just the new shapes.

cixzhang (changes requested):
Thanks for this — the fallback chain and the status-dot placement are the fiddly parts and you got them right.

The one I can't take as-is: shape only reaches an instance. The clipping box is the inner div, which no theme target touches, so `.astryx-avatar {border-radius: 8px}` leaves the avatar a circle — while the docs here promise a theme can set it globally. In a group of squares the "+N" chip also stays a pill, and a consumer can't reach it to fix it.

Both fall out if the radius goes through one component token, the way chat surfaces have `--radius-chat`:

```ts
circle:  {'--radius-avatar': radiusVars['--radius-full']},
rounded: {'--radius-avatar': radiusVars['--radius-element']},
square:  {'--radius-avatar': radiusVars['--radius-none']},
```

with content, wrapper and the overflow chip all reading `--radius-avatar`.

Does that cover the rounded-square case you were after?

[Reviewed by Robohands]

HelloOjasMutreja:
Done, pushed in 121809d2e. Went with your `--radius-avatar`-style approach exactly, one addition: named it `--_avatar-radius` (underscore-prefixed) to match this codebase's existing convention for component-local indirection vars like `--_chat-composer-radius` and `--_avatar-group-overlap`, rather than adding a new public entry to the global `radiusVars` token set the way `--radius-chat` is. Shape still only ever selects among the existing `radiusVars` values (full/element/none), it isn't introducing a new radius value of its own, so it seemed like the private convention fit better here. Happy to rename if you'd rather it be public.

`content`, `wrapper`, and the interactive focus ring now all read `border-radius: var(--_avatar-radius)`; only `shapeStyles` sets its value, once, per shape variant.

For the "+N" chip: `AvatarGroupOverflow` had no `shape` awareness at all (no prop on `AvatarGroup` either), so I threaded `shape` through `AvatarGroupContext` the same way `size` already flows, and had the overflow chip read it. One behavior call worth flagging: `AvatarGroup`'s `shape` now overrides an individual `Avatar` child's own `shape` prop when set, exactly matching how `AvatarGroup.size` already overrides a child's own `size` (`avatarGroup?.size ?? size`, now `avatarGroup?.shape ?? shape`). So a group stays uniform by design, same as it already does for size, rather than shape being independently stateable per-avatar inside a group.

Verified live (Storybook + a temporary harness, not committed): a `square` group renders all three avatars and the overflow chip as squares; a `rounded` group correctly overrides one child that had its own `shape="circle"`. Screenshots:

group shapes
all three shapes side by side

One honest caveat: I didn't verify an actual theme-package CSS override (e.g. `.astryx-avatar { --_avatar-radius: 8px }`) wins the real cascade against StyleX's own atomic class setting the same property, since that intersects with #5095 (StyleX's own layer ordering sorting after `astryx-theme` for regular property overrides) which is still open and unrelated to this PR. What this fix does guarantee at the code level is that content, wrapper, focus ring, and the overflow chip all read from the exact same single property, so whichever mechanism does end up winning a cascade fight reaches every surface consistently instead of just the wrapper.

packages/core/src/Avatar/Avatar.tsx
HelloOjasMutreja: Good catch, and you're right. Checked the actual numbers: `--color-text-secondary` on `--color-neutral` passes AA comfortably in light mode (5.42:1), it only fails in dark mode (4.27:1, needs 4.5:1). Switching to `--color-text-primary` fixed dark mode but also bumped light mode to 15.63:1, a much heavier weight than the rest of the design uses for this muted/secondary surface, and out of scope for a shape PR either way.

Reverted in 9cf41e85c, back to `--color-text-secondary`. Filed the real dark-mode-only bug separately as #5279 with the contrast numbers, since it deserves its own properly-scoped, theme-aware fix rather than a blanket token swap.

packages/core/src/Avatar/Avatar.doc.mjs
```diff
@@ -28,7 +28,7 @@ export const docs = {
   },
   theming: {
     targets: [
-      {className: 'astryx-avatar', visualProps: ['size']},
+      {className: 'astryx-avatar', visualProps: ['size', 'shape']},
```
cixzhang, line 33: The `--_avatar-radius` var and a `derived` borderRadius mapping need to be declared in this theming block.

packages/core/src/Avatar/index.ts
```diff
@@ -37,8 +37,8 @@ export interface AvatarStatusDotVariantMap {
   error: true;
 }
 
-export {Avatar, resolveSize} from './Avatar';
-export type {AvatarProps, AvatarSize} from './Avatar';
+export {Avatar, resolveSize, shapeStyles} from './Avatar';
```
cixzhang: This puts `shapeStyles` in the package's public API.

packages/core/src/AvatarGroup/AvatarGroup.doc.mjs
```diff
@@ -50,6 +50,12 @@ export const docs = {
       description: 'Size applied to all avatars via context.',
       default: "'md'",
     },
+    {
+      name: 'shape',
+      type: 'AvatarShape',
```
cixzhang: Spell the literals out; the doc test wants them visible.

cixzhang (changes requested):
Thanks for turning this so fast, and `--_avatar-radius` is the right name — it matches `--_card-radius` and the private-var convention exactly.

What's missing is the registration that convention needs. A theme writing `avatar: {base: {borderRadius: '10px'}}` gets only `border-radius` on the wrapper, so the photo still clips to a circle and shape stays per-instance. Card works because of:

```js
vars: [{name: '--_avatar-radius', private: true, /* … */}],
derived: [{property: 'borderRadius', vars: ['--_avatar-radius']}],
```

plus the `derivedVarRegistry` entry — which is what the red test is asking for. Setting the var directly already changes every surface, so the plumbing is done.

Separately, `shapeStyles` is now a public export of core. Could AvatarGroupOverflow import it from `./Avatar/Avatar` so it stays internal?

[Reviewed by Robohands]

cixzhang:
Merged `main` in for you so you are not resolving conflicts on top of the review — two spots, both mechanical: `Avatar.tsx` kept main's new disabled-cursor handling alongside your `--_avatar-radius` focus ring, and `AvatarGroupOverflow.tsx` kept your `shape` line alongside main's `resolveSize("md")` default. 128 Avatar/AvatarGroup tests and `tsc` pass locally. The doc-registration change from my review is still yours.

HelloOjasMutreja:
Pushed in 41cd6d53c:

- Added the `--_avatar-radius` registration you flagged: `theming.derived` in `Avatar.doc.mjs` plus the `avatar` entry in `derivedVarRegistry.ts`, same shape as Card's `--_card-radius`. `derivedVarRegistry.test.ts` now covers Avatar too.
- `shapeStyles` is back to internal-only: removed from the public `Avatar/index.ts` barrel, `AvatarGroupOverflow` now imports it straight from `Avatar/Avatar.tsx`.
- `AvatarGroup.doc.mjs`'s `shape` prop now spells out `'circle' | 'rounded' | 'square'` instead of the `AvatarShape` alias, matching how `Avatar.doc.mjs` already documents it.

201 tests pass locally (Avatar, AvatarGroup, derivedVarRegistry), typecheck and lint are clean. Thanks for merging main in, that saved me a rebase.

cixzhang (approved):
Thanks for three rounds on this. Verified in the browser: a theme's `borderRadius` now reaches the photo's clip, not just the wrapper. Merging as is.

[Reviewed by Robohands]

PR #3351, opened 2026-07-02
feat(codeblock): per-line add/remove diff backgrounds via highlightLines (proposal for #3345)

thedjpetersen:
> **Draft — pending API sign-off on #3345**

## What

Extends `CodeBlock`'s `highlightLines` prop from `number[]` to `Array<number | {line: number; type?: 'add' | 'remove' | 'highlight'}>` so builders can render per-line diff backgrounds:

- `type: 'add'` → success-toned wash (`--color-success-muted`)
- `type: 'remove'` → error-toned wash (`--color-error-muted`)
- `type: 'highlight'`, plain numbers, and objects without `type` → the existing neutral accent (`--color-accent-muted`), byte-for-byte the same styling as today

Both washes are the theme-aware `light-dark()` translucent token pairs already used by Banner/FieldStatus, so they compose correctly over `--color-syntax-background` in light and dark schemes and under SyntaxTheme backgrounds. No hardcoded hex.

Accented lines also carry a `data-line-type="add" | "remove" | "highlight"` attribute (alongside the existing `data-line`), giving themes and tests a stable hook.

New exports: `CodeBlockHighlightLine`, `CodeBlockLineAccent` types. No changes to `package.json` exports.

## Why

Closes the gap in #3345: builders showing config/code changes need green/red line washes. An internal-tool builder (Meridian) hand-tinted lines with manual green/red styling because `highlightLines` only offers the neutral accent; the ChatToolCalls diff-display template is an in-repo consumer that would use this.

API arbitration (compact round, 2026-07-01, naive sonnet generations + haiku probe):

- **(a) extend `highlightLines`** — 0 invented props, 0 escapes, exact usage → **chosen**: backward compatible, one prop owns line accents
- (b) separate `diffLines: {added, removed}` — 0/0, clean but splits line-accent ownership; runner-up
- (c) `language="diff"` auto-parse — **failed**: the agent embedded `+`/`-` prefixes and `@@` hunk headers INTO the code string, conflating content with presentation; the copy button would copy the markers
- Probe against the current API invented `addedLines`/`removedLines`, confirming demand pull

## Testing

- `packages/core/src/CodeBlock/CodeBlock.test.tsx` (new): **5 tests**
  1. backward compat — plain `number[]` renders the neutral accent, identical class on all highlighted lines
  2. mixed numbers + objects — per-type `data-line-type` and distinct classes for add/remove/highlight; plain number class-identical to explicit `type: 'highlight'`
  3. object without `type` defaults to neutral accent
[...]

PR #4328, opened 2026-07-25
feat(codeblock): +/− diff markers and language="diff" (builds on #3345/#3351)

ejc3:
## What

Builds on #3345 / #3351 and closes the two follow-ups @cixzhang raised on #3351:

1. **`+` / `−` diff markers.** The add/remove washes are now paired with a gutter marker, so the distinction is never colour-only (WCAG 2.1 SC 1.4.1) — the same colour+shape approach `AvatarStatusDot` took in #4143. The marker is an `::after` pseudo-element placed in a leading grid column, so range mode's bare text node (and its highlight offset mapping) is untouched — the same reason the line number uses `::before`. It composes with the line-number gutter as a three-column grid (`marker | number | code`).
2. **`language="diff"`.** `code` is treated as a unified diff: `+`/`-` lines auto-derive add/remove accents, `@@` and file headers dim as metadata, and **Copy yields the code with all diff punctuation stripped** (the concern raised in #3345's arbitration).

`highlightLines` now accepts `number | {line, type?: 'add' | 'remove' | 'highlight'}`. This ports @thedjpetersen's typed API from #3351 onto current `main`: #3351 branched before the per-line-number gutter landed, so it no longer applies cleanly — this PR carries its API forward and adds the two features on top. Plain-number `highlightLines` are unchanged (backward compatible).

## API

```tsx
// Explicit per-line types
<CodeBlock
  code={updatedConfig}
  language="json"
  highlightLines={[{line: 3, type: 'remove'}, {line: 4, type: 'add'}, 7]}
/>

// Or hand it a unified diff and let it parse
<CodeBlock language="diff" code={unifiedDiff} />
```

## Accessibility

The `+`/`−` marker is the non-colour affordance for SC 1.4.1, mirroring the precedent `AvatarStatusDot` set in #4143 (colour + shape). Markers appear only when there are add/remove lines; a neutral highlight stays marker-free.

## Verification

- `pnpm -F @astryxdesign/core build` ✓ · `eslint` clean on changed files ✓ · `check:changesets` ✓ · `sync:exports:check` ✓ · `check:sync` ✓
- CodeBlock unit tests: **24 pass** (11 new — typed `highlightLines`, markers, and `language="diff"` parse/copy).
- Rendered against the **built** component and checked visually in **light and dark**, with and without line numbers, and for `language="diff"` (the `@@` header dims; screenshot below).

Credit: typed `highlightLines` API — @thedjpetersen (#3345 / #3351); markers + `language="diff"` requested by @cixzhang on #3351.

Opened from a fork. Happy to adjust the API, or split the marker and `language="diff"` into separate PRs, if you'd prefer.


ejc3:
### Visual verification — 10-combination render matrix (light + dark)

Rendered the built component across the feature's surface and verified it two ways: eyeballed the screenshots, and had an independent reviewer audit the **computed styles** of every combination in both modes (marker `::after` glyph, the success/error/accent washes resolving distinctly per mode via `light-dark()`, metadata staying glyph-free, neutral back-compat, and the marker/number grid tracks). **All 10 combos pass in both modes, no discrepancies.**

CodeBlock diff feature — 10-combo × light/dark matrix

Combinations (each light + dark):
- **A/B** `highlightLines` add/remove — without / with line numbers (marker+number grid)
- **C** `language="diff"` single hunk · **D** multi-file patch (both files' headers dim) · **E** CRLF + `\ No newline` sentinel (both dim, no marker)
- **F** `isWrapped` long lines (marker stays top-aligned) · **G** `container="section"` embedded · **H** `size="sm"`
- **I** neutral `highlightLines={[2,3]}` — **no markers** (backward compat) · **J** mixed add/remove/neutral/context

Resolved washes (both modes, non-transparent, distinct): success `rgba(11,153,31,.2/.247)`, error `rgba(227,25,59,.2)/rgba(245,57,79,.247)`, accent `rgba(0,130,251,.2/.247)`.

ernestt (approved):

packages/core/src/CodeBlock/CodeBlock.tsx
```diff
@@ -294,6 +295,91 @@ const styles = stylex.create({
...
+      color: 'var(--color-syntax-punctuation)',
+      userSelect: 'none',
+      fontFamily: typographyVars['--font-family-code'],
+    },
+    '::before': {
+      content: 'attr(data-line)',
+      gridColumn: '2',
+      alignSelf: 'start',
+      textAlign: 'end',
+      marginInlineEnd: `calc(${spacingVars['--spacing-3']} + ${borderVars['--border-width']} + ${spacingVars['--spacing-4']})`,
```
cixzhang, line 368: `lineNumbered` puts this 29px on `columnGap`; here it's a margin inside a 1ch track, so it collapses.

packages/core/src/CodeBlock/CodeBlock.tsx
```diff
@@ -294,6 +295,91 @@ const styles = stylex.create({
...
+      // Blank by default (keeps the gutter aligned for context/metadata lines). The +/- glyphs come
+      // from lineMarkerAdd/lineMarkerRemove. Deliberately a literal, NOT content: attr(data-diff-marker):
+      // some bundler CSS minifiers (Next 16) silently drop attr() content in ::after, killing the marker.
+      content: '""',
+      gridColumn: '1',
+      gridRow: '1',
+      alignSelf: 'start',
+      textAlign: 'center',
+      minWidth: '1ch',
+      color: 'var(--color-syntax-punctuation)',
```
cixzhang, line 336: Punctuation grey on the wash measures 1.8:1. Might need a token that stays legible on success/error.

packages/core/src/CodeBlock/CodeBlock.test.tsx
```diff
@@ -396,3 +396,207 @@ describe('CodeBlock', () => {
...
+    ].join('\n');
+    const {container} = render(<CodeBlock code={diff} language="diff" />);
+    // 2nd file's header lines are metadata (the multi-file fix), not add/remove content.
+    expect(getLine(container, 7).dataset.lineType).toBeUndefined(); // diff --git a/f2
+    expect(getLine(container, 8).dataset.lineType).toBeUndefined(); // --- a/f2
+    expect(getLine(container, 9).dataset.lineType).toBeUndefined(); // +++ b/f2
+    expect(getLine(container, 11).dataset.lineType).toBe('remove');
+    expect(getLine(container, 12).dataset.lineType).toBe('add');
+    fireEvent.click(screen.getByRole('button', {name: 'Copy code'}));
+    expect(navigator.clipboard.writeText).toHaveBeenCalledWith('b\nd'); // post-image of both files
```
cixzhang, line 573: This locks in one post-image across two files. Probably wants a per-file boundary, or the diff.

cixzhang (changes requested):
Thanks for carrying this forward — the parser is the hard part here, and the multi-file and in-hunk `---`/`+++` cases are right.

Two things before it lands. With `hasLineNumbers` the number ends up flush against the code, with the vertical rule cutting through it (frames below). And the `+`/`−` lands at 1.8:1 on its own wash — `--color-syntax-punctuation` is the line-number grey, quiet by design on the plain background and nearly invisible on the wash. It's the sign someone reads instead of the colour, so ideally it's the most legible thing there.

Also: for a two-file patch Copy joins both post-images with no boundary, so you paste two files' contents run together. No single resulting file exists there — should Copy fall back to the diff for a multi-file patch?

[Reviewed by Robohands]

cixzhang:
One correction to my review above: I framed it as contradicting @ernestt's approve, and that isn't fair to either of you. His approve is empty-bodied and he's a design owner — that reads as "the design is fine", not a claim about the numbered-gutter geometry or the marker's contrast on a wash. Nothing to contradict.

Worth saying that both my findings are visual, though. @ernestt — the collision only shows with `hasLineNumbers`, and the +/− marker measures 1.8:1 on its own wash. Might be worth a look at that combination if it wasn't in the matrix.

[Reviewed by Robohands]

cixzhang:
Checking in on CodeBlock diffs: line-number spacing, marker contrast, and multi-file Copy behavior are still open. Would you like to continue, have us carry it, or close this for now?

PR #4330, opened 2026-07-25
fix: security — XSS in Markdown parser, ReDoS in ChatTokenizedText, dotfile creation in sanitizeName

Sunil56224972:
fix: security — XSS in Markdown parser, ReDoS in ChatTokenizedText, dotfile creation in sanitizeName

## Summary

Three security hardening fixes with tests:

1. **Markdown parser: sanitize dangerous URL schemes (XSS)**
   `[text](javascript:alert(1))` was parsed into a link node with the
   raw `javascript:` href. When rendered as `<a href=...>`, this is a
   stored XSS vector for untrusted markdown (AI chat, user comments).
   Added `isSafeUrl()` that rejects `javascript:`, `vbscript:`, and
   `data:text/html` schemes — dangerous URLs are now emitted as plain
   text nodes instead.

2. **ChatTokenizedText: fix broken escapeRegExp (ReDoS)**
   The regex character class was malformed: `]` prematurely closed
   the class, so `]` and `\` were NOT escaped. Token values containing
   these characters were injected raw into `new RegExp()`, causing
   crashes or ReDoS. Fixed to the standard escapeRegExp pattern.

3. **sanitizeName: reject dotfile names (.env, .htaccess)**
   `sanitizeName('.env')` passed validation because the check only
   rejected `'.'` exactly and names starting with `'..'`. A theme
   named `.env` would create a hidden file. Strengthened to reject
   all names starting with `.`.


cixzhang (approved):
Approving — all three fixes are correct and come with tests. Landing this closes the immediate issues; recommend a fast-follow to harden the URL check.

- **Markdown `isSafeUrl`** — correctly gates inline links, images, reference links, and GFM autolinks; the control-char strip + trim + lowercase defeats the `java\nscript:` style bypass. Good to merge as-is. Followup: convert the denylist to an allowlist (permit `http:`/`https:`/`mailto:`/`tel:`/relative paths/`#` anchors, reject everything else) so `data:image/svg+xml`, `file:`, `blob:`, and custom schemes are covered too.
- **escapeRegExp** — the old character class closed early on `]`, leaving `]` and `\` unescaped; the new `/[.*+?^${}()|[\]\\]/g` is the standard pattern. Correct.
- **sanitizeName** — rejecting all leading-dot names is reasonable for a name sanitizer; just confirm no legitimate input is expected to start with `.`.

cixzhang:
Merged `main` into this branch and armed auto-merge (squash) — it will land once CI is green. Thanks for these, and sorry it sat: all three fixes are still correct against current `main`.

Three things I did on top of your commits:
- `lint` was red on `no-control-regex` in `isSafeUrl` — the control characters in that class are the whole point (`java\nscript:` is the bypass), so I added a targeted `eslint-disable-next-line` rather than weakening the strip.
- Added a patch changeset crediting you, so the fixes actually ship in a release.
- Your `sanitizeName` dotfile fix has since landed on `main` by another route (the file moved to `packages/cli/foundation/fs/path-safety.mjs`); the merge kept it, so that part is now a no-op.

Fast-follow coming from me to convert `isSafeUrl` from a denylist to an allowlist, per my review note — that covers `data:image/svg+xml`, `file:`, `blob:` and custom schemes too. Nothing needed from you.

cixzhang (approved):

Issue #4339, opened 2026-07-26
The `test` job is the CI critical path on re-pushes, merge queue, and main


Han5991:
## Problem

The single `test` job has quietly become the CI critical path in every context except a cold first push. Step-level measurements from two real runs of the same PR (#4101):

| Job | First push (cold Next cache), run 29749301708 | Re-push (warm), run 29752691491 |
|---|---|---|
| build-sandbox | **385s** ← critical | 206s |
| **test** | 327s (58s behind) | **324s ← critical** |
| build-storybook | 164s | 167s |

- **Re-pushes** — the loop contributors actually wait on while iterating on review feedback — are already bottlenecked by `test`: the per-PR Next.js cache drops build-sandbox to ~3m30s while `test` stays at ~5m30s.
- **Merge queue** (`merge_group`) and **main** never run build-sandbox at all, so `test` is the critical path there outright.
- Cold first pushes still hide behind build-sandbox, but only by 58 seconds.

`test` is also the one job that grows linearly with the component count — 359 files / 7,118 tests today. Inside the job: ~30s runner setup + ~284s vitest (413s of test CPU across workers, 54s transform, 110s import, 148s jsdom environment), which shards cleanly.

`deploy.yml` runs the full suite on main as well, but interleaves it with build/typecheck deploy gates in a single job, so it needs its own restructure rather than the same treatment.

---

Fix in #4103.

cixzhang:
Thanks for the detailed critical-path analysis. This is fixed on current `main` by [#6051](https://github.com/facebook/astryx/pull/6051): the UI and Node test projects now run in parallel lanes in both CI and deploy workflows, while the historical `test` check is retained as a lightweight join. Closing as fixed.

Issue #4337, opened 2026-07-26
Streaming Markdown perf tests are capped by vitest's 5s default, not their declared budgets


Han5991:
## Problem

The streaming Markdown perf tests in `packages/core/src/Markdown/parser.perf.test.ts` assert wall-clock budgets of 5000–30000ms via `expect(elapsed).toBeLessThan(maxMs)`, but never pass a vitest timeout — and no `testTimeout` is configured anywhere in the repo. The effective limit is therefore vitest's **5s default**, applied before the budget assertion can ever run.

This surfaced as a `test` job failure on main ([run 29704122332](https://github.com/facebook/astryx/actions/runs/29704122332/job/88238134124)):

```
FAIL packages/core/src/Markdown/parser.perf.test.ts
  > parseMarkdown performance > streaming full re-parse XL under 30000ms
Error: Test timed out in 5000ms.
```

The commit it failed on was unrelated to Markdown — this is a latent defect in the harness setup, not a parser regression:

- On the next healthy run of the same suite ([29710649807](https://github.com/facebook/astryx/actions/runs/29710649807)), the XL full re-parse took **3641ms** — already 73% of the effective 5s limit. Any runner contention pushes it over.
- When it does, vitest reports "timed out" instead of the measured elapsed time, so a genuine parser slowdown and a contended runner are indistinguishable from the failure message.
- #3848 stabilized the millisecond-scale budgets and noted that the streaming budgets "carry 10x+ headroom, which contention cannot realistically eat". That headroom was illusory: the harness capped every streaming test at 5s regardless of its declared budget.

## Related: the Table perf benchmarks have the sibling defect

`Table.perf.test.tsx` asserts tight budgets (100–200ms, 750ms on CI) against a **single** wall-clock sample — the exact pattern #3848 diagnosed as flaky under parallel forks and fixed in the Markdown suite with best-of-N sampling. `Table.perf` was left out of that fix.

---

Fix in #4099.

PR #4340, opened 2026-07-26
chore(github): add a Test or CI Infrastructure issue template

Han5991:
## Why

Blank issues carry no labels, and nothing adds them afterwards — no workflow under `.github/workflows/` declares an `issues` trigger. The `labels:` an issue form declares in its front matter is the only label applied at creation time.

Neither existing template covers a report about the repository's own tooling: `bug.yml` is for a defect in a shipped component and requires an `Astryx Version`; `rfc.yml` is for a new component or API and requires accessibility and component-composition answers. So test, CI, and dependency reports fall through to a blank issue and land unlabelled — #4337, #4338, and #4339 this week.

## What

Adds `.github/ISSUE_TEMPLATE/infra.yml` — **"Test or CI Infrastructure"**, labelled `type:test`, titled `[Infra] ` to match `[Bug] ` and `[RFC] `.

The structure follows `rfc.yml` so the three forms read as one set: a *Before you submit* block, **Problem** and **Evidence** required, **Scope Check** and **Proposed Direction** optional, and a pre-submission checklist.

`type:test` is the closest label that already exists. A form naming a label that doesn't exist silently applies nothing, so this uses only what's already defined.

Blank issues stay enabled — there is no `config.yml` in this directory and this PR doesn't add one.

## Verification

- Parses as a valid issue form: every body item uses a supported type, each non-markdown field has an `id` and a `label`, and the checklist options all carry a `label`.
- `./scripts/add-copyright.sh --check` does not flag this file. (Note: the script scans the whole working tree, so locally it reports untracked build output under `apps/docsite/public/` — none of it git-tracked. On a clean CI checkout that noise is absent.)
- `pnpm check:repo` passes.

The rendered form can only be confirmed after merge — GitHub serves issue templates from the default branch, so a PR branch can't preview it.

---

If there are fields, wording, or a different label you'd want in this form, say so and I'll fold them in.

### DIFF
--- a/.github/ISSUE_TEMPLATE/infra.yml
+++ b/.github/ISSUE_TEMPLATE/infra.yml
@@ -1,7 +1,7 @@
 # Copyright (c) Meta Platforms, Inc. and affiliates.
 
 name: "Test or CI Infrastructure"
-description: Report a problem in the repo's own test suite, CI pipeline, or build tooling.
+description: Report a problem in Astryx's test suite, CI pipeline, or build tooling.
 title: "[Infra] "
 labels: ["type:test"]
 body:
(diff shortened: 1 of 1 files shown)


cixzhang (approved):
Thanks for improving the issue-triage flow here — a dedicated Test / CI Infrastructure template will make it much easier to route and act on those reports. Nice, low-friction addition!

PR #4344, opened 2026-07-26
feat(thumbnail): remove the elevation prop

kentonquatman:
## What

Removes the `elevation` prop from `Thumbnail`. Elevation is no longer configurable on this component.

Thumbnail now always uses its flat resting look, keeping the hover-only shadow it gets when it's interactive (`onClick`).

## Why

Thumbnail is a compact, embedded preview tile — it lives inline in composers, upload lists, and galleries. A resting elevation control invited it to float off the page in contexts where it should read as part of the surface. Dropping the prop removes that footgun and keeps the component's surface treatment fixed and consistent.

## Changes

- **`Thumbnail.tsx`** — remove the `elevation` prop and its `Elevation` type import, the `elevationStyles` map, and the `--_thumbnail-elevation` CSS-var plumbing. The interactive style rests flat and bumps to `--shadow-med` on hover, exactly as before for the default case.
- **`Thumbnail.doc.mjs`** — drop the `elevation` prop entry and its localized descriptions.
- **`Thumbnail.test.tsx`** — remove the `elevation` test block.
- **`Thumbnail.stories.tsx`** — remove the `elevation` argType.
- **`ThumbnailElevated` showcase block** — deleted; it existed solely to demonstrate `elevation`.
- **`elevation.doc.mjs`** — drop Thumbnail from the list of elevation-configurable surfaces and from the scale examples.

## Changeset

No new changeset. The `elevation` prop is **unreleased** — it's still an unmerged `[feat]` changeset in `.changeset/`, never shipped in any published version. Adding then removing it on Thumbnail within the same release cycle is a net no-op for consumers, so this instead **removes Thumbnail from the original elevation feat changeset** rather than adding a breaking-change note for a prop that never existed publicly.

## Testing

- `pnpm -F @astryxdesign/core test` (Thumbnail suite) — 18 passing
- `pnpm build && pnpm -F @astryxdesign/cli typecheck:strict` — clean
- `pnpm -F @astryxdesign/core typecheck:docs`, `typecheck:template-docs` — clean
- `node scripts/check-changesets.mjs` — valid


cixzhang (approved):
Clean removal — thanks for splitting this out. The `elevation` prop was never released (not in the published CHANGELOG), so dropping it from Thumbnail carries no consumer risk, and this correctly trims Thumbnail from the still-pending shared elevation changeset + `elevation.doc.mjs` rather than deleting them (they still cover the other surfaces). Prop, styles, the `--_thumbnail-elevation` var, the `ThumbnailElevated` example, docs (en/zh/dense), Storybook control, and tests all go together, and the pre-existing hover-shadow behavior is untouched. LGTM.

PR #4345, opened 2026-07-26
fix(core): Thumbnail remove button uses fixed overlay contrast

kentonquatman:
## Summary

The `Thumbnail` remove button (shown when `onRemove` is set) previously adapted its colors to the underlying image's luminance via `useImageMode` (APCA sampling) + `MediaTheme`. Sampling pixel brightness behind the button is an unreliable way to guarantee contrast — a busy or mid-tone region can leave the button hard to see, and it depends on canvas pixel access (CORS, timing) that can silently fall back.

This switches the button to fixed, theme-token colors:

- Button background → `--color-overlay` (a translucent scrim that darkens whatever is beneath it)
- X icon → `--color-on-dark`

Because the icon always sits on the overlay scrim, it reads reliably on any image without sampling luminance.

## Changes

- `Thumbnail.tsx`: set `backgroundColor: --color-overlay` and `color: --color-on-dark` on the remove button's style overrides. The `close` icon uses the default `inherit` color, so it picks up `--color-on-dark` from the button.
- Removed the now-unused `useImageMode` hook call, `MediaTheme` wrapper, and the `BUTTON_REGION` sampling constant from `Thumbnail`. (The shared `useImageMode`/`MediaTheme` primitives are untouched — they're still used elsewhere.)
- Updated the file/component JSDoc and the `.doc.mjs` anatomy note to describe the fixed-contrast approach.

## Notes

- The `secondary` button variant's hover/pressed overlay gradients still layer on top of the scrim, so interaction feedback is preserved.
- Patch changeset included.

## Test plan

- `pnpm -F @astryxdesign/core build` ✅
- `vitest run packages/core/src/Thumbnail` — 20/20 pass ✅
- `pnpm -F @astryxdesign/core typecheck` ✅
- `eslint` on changed files ✅
- `node scripts/sync-exports.js --check` ✅

### DIFF
--- a/packages/core/src/Thumbnail/Thumbnail.tsx
+++ b/packages/core/src/Thumbnail/Thumbnail.tsx
@@ -4,14 +4,14 @@
 
 /**
  * @file Thumbnail.tsx
- * @input Uses React, stylex, Button, Skeleton, Spinner, MediaTheme, useImageMode
+ * @input Uses React, stylex, Button, Skeleton, Spinner
  * @output Exports Thumbnail component, ThumbnailProps
  * @position Core implementation; consumed by index.ts
  *
  * Square preview card for image attachments. Shows a skeleton shimmer while
  * the image loads, the image on success, or a placeholder on failure.
- * Uses useImageMode (APCA) to detect image luminance so the overlaid
- * remove button always has sufficient contrast.
+ * The overlaid remove button sits on a fixed --color-overlay scrim with an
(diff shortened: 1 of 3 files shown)


packages/core/src/Thumbnail/Thumbnail.tsx
cixzhang, line 364: We may need to give this button its own theme override target since we have adjustments unique to the thumbnail here.

cixzhang (approved):

PR #3507, opened 2026-07-03
fix(cli): swizzle copies nested component source recursively

oneshot2001:
## Summary

Fixes #3506.

`astryx swizzle <Name>` copied only the **top-level files** of a component directory and silently skipped nested subdirectories. For components whose source spans folders (e.g. `Table/plugins/*`), this produced a **broken partial eject**: the copied entry file still `export … from './plugins/…'`, but that subtree was never copied, so the swizzled component fails module resolution — while the CLI reports success.

Root cause in `packages/cli/src/commands/swizzle.mjs`: the copy loop read the component dir non-recursively (`if (!stat.isFile()) continue`), and `rewriteImports` only handled top-level files (it mapped every `../seg/...` to `owner/seg`).

## Changes

- **Recursive copy.** New `collectSourceFiles()` walks the component tree, preserving nested structure and excluding tests/docs/README at every level. It's the single source of truth for the pre-flight collision check, the copy loop, and the reported file list. Symlinks are not followed.
- **Location-aware import rewriting.** `rewriteImports` now accepts an optional `{fromDir, componentDir}`. Each `../` import is resolved from the source file's own directory; only imports that **escape** the component are rewritten to the owner package's subpath. Intra-component imports from nested files (e.g. `../../types` in `Table/plugins/pagination/*`) stay relative instead of becoming a broken `@astryxdesign/core/..`. Without the location arg the legacy behavior is unchanged, so the existing exported `rewriteImports(content)` contract and its tests are preserved.
- **Path safety.** Each destination is asserted within the output dir (`assertWithin`) before any `mkdir`/write.

## Testing

- `pnpm exec vitest run packages/cli/src/commands/swizzle.*.test.mjs` — 21/21 pass (added 5 unit tests for nested-file rewriting; existing routing + path-safety suites still green).
- **End-to-end:** `astryx swizzle Table` now emits the full `plugins/` subtree; spot-checked `Table/plugins/pagination/useTablePagination.tsx` — escaping imports resolve to `@astryxdesign/core/theme` / `@astryxdesign/core/Pagination` while the intra-component `../../types` stays relative. Top-level `index.ts`'s `./plugins/*` re-exports now resolve.
- `pnpm check:repo` (sync, package-boundaries, changesets, demo-media) passes; changeset added (`@astryxdesign/cli` patch, `[fix]`).

## Notes

Behavior change is additive — top-level-only components swizzle identically to before; multi-directory components now eject completely. Happy to adjust naming/placement to match maintainer preference.

[...]

PR #4347, opened 2026-07-26
fix(cli): swizzle copies nested component source recursively

### LINKED ISSUE #3506: swizzle silently drops nested subdirectories, producing a broken partial eject (e.g. Table/plugins)
## Summary

`astryx swizzle <Name>` copies only the **top-level files** of a component directory and silently skips nested subdirectories. For components whose source is split across subfolders (e.g. `Table/plugins/*`), this produces an **incomplete and broken eject** — the copied entry file still imports from the dropped subdirectory, so the swizzled component fails module resolution — yet the CLI reports success. This contradicts the documented promise that swizzle "ejects a component's full source into your project to own."

## Root cause

`packages/cli/src/commands/swizzle.mjs` — the copy loop reads the component directory non-recursively and skips anything that isn't a file:

```js
// packages/cli/src/commands/swizzle.mjs (~L340)
const files = fs.readdirSync(componentDir);
for (const file of files) {
  if (isExcludedFromCopy(file)) continue;
  const srcPath = path.join(componentDir, file);
  const stat = fs.statSync(srcPath);
  if (!stat.isFile()) continue;   // <-- subdirectories are silently dropped, no recursion
  ...
}
```

The pre-flight collision check (~L304) and the reported `copiedFiles` list (~L363) apply the same `stat.isFile()` filter, so nested files are invisible end-to-end — including in the success count.

Compounding it: `rewriteImports()` only rewrites **upward** relative imports (`/^\.\.\//`). A component's own **downward** imports into its subfolders (`./plugins/...`) are left untouched — now pointing at a directory that was never copied.

## Reproduction

```bash
node packages/cli/bin/astryx.mjs swizzle Table --out ./swizzled
```

`Table` has a `plugins/` subdirectory. After swizzling, the ejected `Table/index.ts` still contains (unmodified):

```ts
export {useTableSelection} from './plugins/selection';
export {useTableSortable}  from './plugins/sortable';
export {useTablePagination, paginateData} from './plugins/pagination';
// ...15+ more re-exports from ./plugins/*
```

…but `./plugins/` does not exist in the output. The swizzled component fails to resolve on import/build, while the CLI reports a successful copy.

## Impact

- **Broken output, reported as success.** No error, no warning — the failure only surfaces later when the consumer tries to build/import the swizzled component.
- **Undermines a headline feature.** "Open internals / own the source" is a core selling point; it silently doesn't hold for any multi-directory component.
- Confirmed affected: `Table` (`plugins/`). Other core components ship nested source dirs too (`Calendar`, `FormLayout`) and are candidates for the same failure; any current or future component that organizes source into subfolders is at risk.

## Suggested fix

Recurse into subdirectories when copying (preserving relative structure under the output dir), and keep the collision check + reported file list consistent with the recursive walk. `rewriteImports` should continue to leave intra-component `./`-relative paths intact once the subtree is actually copied.

Alternatively, if partial ejects are ever intentional, `swizzle` should at minimum **detect and refuse/warn** when a component directory contains subdirectories it isn't copying, rather than reporting success.

---

*Filed after an architecture review of the repo at current `main`. Happy to open a PR for the recursive-copy fix if that's useful.*

cixzhang:
## Summary

Fixes #3506.

`astryx swizzle <Name>` copied only a component's top-level files and silently skipped nested subdirectories. For components whose source spans folders (e.g. `Table/plugins/*`), that produced a broken partial eject: the copied entry file still `export … from './plugins/…'`, but the subtree it pointed at was never copied — so the swizzled component fails module resolution while the CLI reports success.

Re-implements the fix originally proposed by @oneshot2001 in #3507, rebased onto the reorganized CLI pillar layout (the logic now lives in `packages/cli/api/swizzle/swizzle.mjs`, the API pillar, rather than the old `src/commands/swizzle.mjs`).

## Changes

- **Recursive copy.** A new `collectSourceFiles()` walks the component tree, preserving nested structure and excluding tests/docs/README at every level. It's the single source of truth for the pre-flight collision check, the copy loop, and the reported file list. Symlinks are not followed.
- **Location-aware import rewriting.** `rewriteImports` gains an optional `{fromDir, componentDir}`. Each `../` import is resolved from the source file's own directory; only imports that actually escape the component are rewritten to the owner package's subpath. Intra-component imports from nested files (e.g. `../../types` in `Table/plugins/pagination/*`) stay relative instead of becoming a broken `@astryxdesign/core/..`. Without the location arg the legacy behavior is unchanged, so the existing `rewriteImports(content)` contract and its tests are preserved.
- **Path safety.** Every destination is asserted within the output dir (`assertWithin`) before any `mkdir`/write.

## Testing

- Added location-aware unit tests (escaping intra-package/sibling imports → owner subpath; intra-component `../../types` stays relative; nested-sibling stays relative; top-level legacy parity) and an end-to-end regression test that swizzles a component with a nested `plugins/` subtree and asserts the full tree is copied with correct rewrites.
- `packages/cli` swizzle suites (`swizzle.test.mjs`, `swizzle.routing.test.mjs`, `swizzle.path-safety.test.mjs`) — 27/27 pass.
- CLI strict typecheck (`typecheck:strict`, checkJs) — clean.
- Changeset added (`@astryxdesign/cli` patch).

## Notes

Behavior change is additive: top-level-only components swizzle identically to before; multi-directory components now eject completely.

### DIFF
--- a/.changeset/swizzle-recursive-copy.md
+++ b/.changeset/swizzle-recursive-copy.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/cli': patch
+---
+
+[fix] swizzle now copies nested component source recursively (#3506). Previously only a component's top-level files were copied, so components split across subdirectories (e.g. `Table/plugins/*`) produced a broken partial eject — the copied entry still imported from the dropped subtree. Import rewriting is now location-aware, so intra-component imports from nested files stay relative while only imports that escape the component are mapped to the owner package's subpaths.
+@oneshot2001
--- a/packages/cli/api/swizzle/swizzle.mjs
+++ b/packages/cli/api/swizzle/swizzle.mjs
@@ -30,20 +30,52 @@ const DEFAULT_ISSUES_URL = 'https://github.com/facebook/astryx/issues/new';
 
 /**
  * Rewrite relative imports that point outside the component directory to use
- * the OWNER package's subpaths. Imports within the copied directory (./x) are
- * left untouched.
+ * the OWNER package's subpaths. Imports that stay inside the copied component
+ * tree (and same-level `./x` imports) are left untouched.
  *
  * e.g. with ownerPackage '@astryxdesign/core':
  *      '../theme/tokens.stylex' -> '@astryxdesign/core/theme'
  *      '../utils/mergeProps'     -> '@astryxdesign/core/utils'
  *
+ * When `location` is supplied (`{fromDir, componentDir}`), rewriting is
+ * location-aware: each `../` import is resolved from the source file's own
+ * directory, and only imports that actually escape `componentDir` are
+ * rewritten. This keeps intra-component imports from NESTED files relative —
+ * e.g. `../../types` in `Table/plugins/pagination/*` still resolves against the
+ * copied `Table/types` rather than becoming a broken `@astryxdesign/core/..`.
+ * Without `location`, the legacy behavior applies (every `../` is treated as
+ * escaping and mapped by its first segment), which is correct for top-level
+ * files and preserves the existing callers/tests.
+ *
  * @param {string} content
  * @param {string} [ownerPackage]
+ * @param {{fromDir: string, componentDir: string}} [location]
  */
-export function rewriteImports(content, ownerPackage = CORE_PACKAGE) {
+export function rewriteImports(content, ownerPackage = CORE_PACKAGE, location) {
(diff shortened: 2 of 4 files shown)


cixzhang:
Closing this — on reflection we can't build on an unsigned contribution here. The underlying issue #3506 remains open and we'll address it independently.

PR #4348, opened 2026-07-26
fix(cli): reject invalid layout --form values

harjothkhara:
## Summary

Reject unsupported `layout --form` values instead of silently parsing them as compact input. Both `layout expand` and `layout check` now exit with the stable `ERR_INVALID_FORM` code, preserve structured suggestions, and continue accepting `auto`, `compact`, and `outline`.

## Case

`packages/cli/lib/xle/parse.mjs:684-686` resolves `auto`, then selects outline only when `form === 'outline'`; every other value falls through to `parseCompact`. Because `packages/cli/api/layout/layout.mjs:163` previously forwarded `form` without validating it, a typo such as `--form garbage` returned success and reported compact input instead of rejecting the command.

## Fix

Validate the three supported values at the shared `analyze()` boundary (`layout.mjs:154-170`), before registry and block loading (`layout.mjs:172-175`). This protects both Node API entry points without duplicating command-specific guards and returns suggestions through the existing `AstryxError` path.

Add the append-only `ERR_INVALID_FORM` contract across the runtime taxonomy (`error-codes.mjs:47,110-111`), published type union (`types/error-codes.d.ts:20`), and README table (`README.md:140`). API regression tests cover both operations and every accepted value (`layout.test.mjs:147-210`); subprocess tests cover both CLI JSON envelopes (`error-codes.test.mjs:105-106`).

## Scope

This is deliberately limited to the Node CLI/API surface that exposes `--form`. It does not change the lower-level parser fallback or `lib/xle/browser.mjs`; that public browser entry point has a different non-throwing result contract, and its current sandbox callers explicitly pass `form: 'auto'`, so harmonizing its validation belongs in a focused follow-up. Validation lives in the API layer rather than Commander `.choices()` because programmatic `layoutExpand`/`layoutCheck` callers bypass Commander, and `.choices()` would also bypass the structured `ERR_INVALID_FORM` and suggestions path.

This is a self-found bug; there is no linked issue.

## Testing

Negative control on `origin/main` (`4bbba6534`):

```text
$ node packages/cli/bin/astryx.mjs layout check VStack --form garbage
✓ Valid (parsed as compact)
exit 0
```

Red/green regression proof:

```text
$ pnpm exec vitest run packages/cli/api/layout/layout.test.mjs
# with the tests present and only the source guard removed
Test Files  1 failed (1)
Tests       2 failed | 15 passed (17)
# both failures: promise resolved instead of rejected

$ pnpm exec vitest run packages/cli/api/layout/layout.test.mjs packages/cli/lib/error-codes.test.mjs
Test Files  2 passed (2)
Tests       42 passed (42)
```

Production paths after the fix:

```text
$ node packages/cli/bin/astryx.mjs layout expand VStack --form garbage --json
exit 1
code: ERR_INVALID_FORM
error: Unknown layout form "garbage"
suggestions: auto, compact, outline

$ node packages/cli/bin/astryx.mjs layout check VStack --form garbage
exit 1
Error: Unknown layout form "garbage"
```

Additional checks:

```text
$ pnpm -F @astryxdesign/core build
passed
$ pnpm -F @astryxdesign/cli typecheck:strict
passed
$ pnpm -F @astryxdesign/cli typecheck:json-api
passed (JSON API + API contract)
$ pnpm check:repo
passed
$ pnpm exec eslint packages/cli/api/layout/layout.mjs packages/cli/api/layout/layout.test.mjs packages/cli/lib/error-codes.mjs packages/cli/lib/error-codes.test.mjs
passed
$ pnpm exec vitest run packages/cli
117 files passed; 2,080 tests passed
1 pre-existing failure: build-theme.watch.test.mjs:131 did not observe a rebuild event
```

The same watcher assertion fails in isolation on pristine `origin/main` (`4bbba6534`), confirming it is unrelated to this change.

Changeset included: `@astryxdesign/cli` patch.


harjothkhara:
Closing this. The same validation landed on main in #4511 (57611be4), which added API-layer input validation across the CLI. `analyze()` in `api/layout/_adapter.mjs` now rejects any form other than compact, outline, or auto, using the existing `ERR_INVALID_OPTION` code. Thanks.

PR #4350, opened 2026-07-26
fix(core): auto-detect RTL for arrow-key navigation in focus hooks

bhamodi:
## Summary

Fixes a11y scan finding #3 (WCAG 1.3.2, meaningful sequence / directionality): `useListFocus` and `useGridFocus` already implement an `isRtl` option that swaps ArrowLeft/ArrowRight for horizontal navigation — but no component ever passes it, and there is no direction source (`InternationalizationContext` carries locale only). So menus, toolbars, tab strips, SegmentedControl, Pagination, and the Calendar date grid never flip arrow direction in RTL locales: pressing ArrowRight in a `dir="rtl"` toolbar moves focus visually left.

The fix makes the hooks resolve direction themselves when `isRtl` is not explicitly passed, by reading `getComputedStyle(container).direction === 'rtl'` from the container element each hook already tracks via its ref. This is the same precedent `Resizable`'s `ResizeHandle` already uses for RTL drag deltas (`getRTLMultiplier`). Every consumer of the two hooks gets correct RTL arrow-key navigation for free, with zero component changes.

Design notes:

- **Lazy and cheap**: `getComputedStyle` runs only on keydown, and only when the pressed key is ArrowLeft/ArrowRight (and, for lists, only for horizontal orientations) — no reads on render, no layout thrash on unrelated keys.
- **SSR-safe**: the helper returns `false` when there is no element or no `window`.
- **Explicit `isRtl` still wins**: the option is now `boolean | undefined`; an explicit boolean short-circuits detection (`isRtl ?? isRtlElement(el)`).
- `useTreeFocus` was checked and has no latent `isRtl` option (tree expand/collapse semantics are a separate concern), so it is untouched.

## Changes

- `packages/core/src/hooks/isRtlElement.ts` (new): shared internal `isRtlElement(el)` helper — SSR-safe computed-direction check.
- `packages/core/src/hooks/useListFocus.ts`: `isRtl` default changed from `false` to auto-detect from the list container; resolved lazily on horizontal arrow keydown. JSDoc/file header updated.
- `packages/core/src/hooks/useGridFocus.ts`: same auto-detection for the grid container's ArrowLeft/ArrowRight swap. JSDoc/file header updated.
- `packages/core/src/hooks/useListFocus.doc.mjs`, `useGridFocus.doc.mjs`: `isRtl` prop docs/defaults updated (targeted edits only).
- `packages/core/src/hooks/useListFocus.test.tsx`: new "RTL auto-detection (WCAG 1.3.2)" suite — 5 tests covering `dir="rtl"` auto-detect (ArrowLeft = next, ArrowRight = previous), LTR default, and explicit `isRtl` override in both directions.
- `packages/core/src/hooks/useGridFocus.test.tsx`: new suite — 4 tests covering `dir="rtl"` auto-detect both directions, vertical keys unaffected, and explicit `isRtl={false}` override. Grid test helper gained a `dir` prop on the container.
- `.changeset/a11y-rtl-arrow-keys.md`: patch changeset.

jsdom note: jsdom reflects the `dir` attribute into computed `direction` only on the element that carries it (verified), so tests set `dir="rtl"` on the container element the hooks read — no `getComputedStyle` mocking needed.

## Test plan

- Pre-fix failure confirmed: the 4 auto-detection tests fail on the unmodified hooks (`Tests 4 failed | 36 passed (40)` across the two files); the explicit-`isRtl` and LTR tests pass, pinning back-compat.
- `node_modules/.bin/vitest run packages/core/src/hooks --reporter=dot` — 14 files, 185 tests passed (9 new RTL tests included).
- `node_modules/.bin/vitest run packages/core --reporter=dot` — 5166 passed, 1 failed: the known `Calendar.test.tsx:884` "February 2026" flake (duplicate text via the aria-live month announcement; click-based, not arrow-key). Passes deterministically in isolation and on file rerun (`63 passed (63)`); unrelated to this change.
- `tsc --project packages/core/tsconfig.json --noEmit` — clean.
- `eslint` on all changed files — 0 errors (doc.mjs files are eslint-ignored by repo config).
- `prettier --check` — all changed ts/tsx/md files clean. The two `.doc.mjs` files already fail prettier at HEAD (repo only formats `*.{ts,tsx,md}` per lint-staged); left unformatted per repo convention.
- `node scripts/check-changesets.mjs` — 45 changeset(s) valid.

## Notes

- Vercel deployment failure on fork PRs is known infra and expected.

### DIFF
--- a/packages/core/src/hooks/isRtlElement.ts
+++ b/packages/core/src/hooks/isRtlElement.ts
@@ -1,0 +1,27 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+'use client';
+
+/**
+ * @file isRtlElement.ts
+ * @input DOM element (nullable)
+ * @output Exports isRtlElement helper resolving an element's computed text
+ *   direction
+ * @position Internal hook utility; used by useListFocus and useGridFocus to
+ *   auto-detect RTL for arrow-key navigation (same getComputedStyle precedent
+ *   as Resizable's ResizeHandle drag deltas).
+ */
+
+/**
+ * Whether `el` renders right-to-left, per its computed `direction`.
+ *
+ * SSR-safe: returns false when `el` is null or no DOM is available. Callers
+ * should invoke this lazily (on keydown, not on render) so getComputedStyle
+ * runs only when a horizontal arrow key is actually handled.
+ */
+export function isRtlElement(el: HTMLElement | null): boolean {
+  if (!el || typeof window === 'undefined') {
+    return false;
+  }
+  return window.getComputedStyle(el).direction === 'rtl';
+}
--- a/.changeset/a11y-rtl-arrow-keys.md
+++ b/.changeset/a11y-rtl-arrow-keys.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] hooks: auto-detect RTL direction for arrow-key navigation in useListFocus and useGridFocus (WCAG 1.3.2)
+@bhamodi
--- a/packages/core/src/hooks/useGridFocus.doc.mjs
+++ b/packages/core/src/hooks/useGridFocus.doc.mjs
@@ -64,8 +64,8 @@ export const docs = {
     {
       name: 'options.isRtl',
       type: 'boolean',
-      description: 'Swap ArrowLeft/ArrowRight so horizontal navigation follows visual direction in right-to-left contexts.',
-      default: 'false',
+      description: 'Swap ArrowLeft/ArrowRight so horizontal navigation follows visual direction in right-to-left contexts. When omitted, auto-detected from the container computed direction on keydown.',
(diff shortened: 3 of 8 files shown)


packages/core/src/hooks/isRtlElement.ts
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * SSR-safe: returns false when `el` is null or no DOM is available. Callers
+ * should invoke this lazily (on keydown, not on render) so getComputedStyle
+ * runs only when a horizontal arrow key is actually handled.
+ */
+export function isRtlElement(el: HTMLElement | null): boolean {
+  if (!el || typeof window === 'undefined') {
+    return false;
+  }
+  return window.getComputedStyle(el).direction === 'rtl';
```
cixzhang, line 26: Compute style here, but since we only use this in callbacks, it's likely okay.

cixzhang (approved):

PR #4352, opened 2026-07-26
fix(core): preserve control state indication under forced colors

bhamodi:
## Summary

Follow-up to a11y scan #3: `packages/core` had zero `forced-colors` handling. Under Windows High Contrast (forced colors mode), the browser strips author-painted `background-color`/`box-shadow`, so any control whose state is conveyed only by a painted fill loses its state indication entirely — an on/off Switch renders as two identical empty tracks, a selected radio looks unselected, a pressed icon-only ToggleButton looks unpressed, and Skeleton placeholders disappear. This violates WCAG 1.4.11 (Non-text Contrast), which requires state indication to remain perceivable.

The fix follows the standard platform approach: `@media (forced-colors: active)` overrides using CSS system colors (`CanvasText`, `Canvas`, `Highlight`, `HighlightText`, `GrayText`), which are honored (not stripped) under forcing. No JS, no new props, no visual change outside forced-colors mode.

## Changes

**Patched:**

- **Switch** — the track gains a `CanvasText` border (its bounds otherwise vanish with the stripped fill); off track stays `Canvas`, on track becomes `Highlight`; the thumb fill moves to the per-state styles so it can be `CanvasText` when off and `HighlightText` when on. Disabled tracks get a `GrayText` border since opacity dimming doesn't survive forcing.
- **CheckboxInput** — only the indeterminate bar needed patching: it's a painted `div` that would be stripped to `Canvas` (invisible). It now uses `CanvasText`.
- **RadioList** — the selected inner dot is a painted `div` that would be stripped, making checked and unchecked radios identical. It now uses `CanvasText`.
- **SegmentedControl** — the selected segment was surface-fill + shadow + font weight; fill and shadow are stripped, leaving only a weight change. Selected now uses `Highlight`/`HighlightText` (the platform convention for a selected item).
- **ToggleButton** — the pressed state was a painted overlay + font weight; icon-only toggles would lose all pressed indication. Pressed now uses `Highlight`/`HighlightText`.
- **Skeleton** — the placeholder fill would be stripped to `Canvas` at 0.25 opacity (invisible). It now uses `GrayText` at full resting opacity; the fade animation still pulses when motion is allowed.

**Verified-OK (no patch needed):**

- **CheckboxInput checkmark** — strokes with `currentColor`; `color`, SVG `fill`/`stroke`, and `border-color` are all in the forced-properties list, so the check renders as `CanvasText` inside a `CanvasText`-bordered box: checked vs unchecked stays distinguishable.
- **Focus indication (all patched controls)** — outline-based; `outline-color` is force-mapped, so focus rings survive forced colors unchanged.
- **Native inputs** — Checkbox/Radio hide the native input (`opacity: 0`) and paint replacements, so the native forced-colors rendering never shows; the patches above target the painted replacements.

Also adds `packages/core/src/__tests__/forcedColors.ts`, a shared test helper that collects injected `@media (forced-colors: active)` rules from the StyleX dev runtime.

## Test plan

- `node_modules/.bin/prettier --check <changed files>` — clean
- `node_modules/.bin/eslint <changed files>` — clean
- `tsc --project packages/core/tsconfig.json --noEmit` — clean
- Scoped: `vitest run packages/core/src/{Switch,CheckboxInput,RadioList,SegmentedControl,ToggleButton,Skeleton}` — 6 files, 192 tests passed (includes 6 new forced-colors tests asserting the compiled output contains the expected system-color rules)
- Full suite: `vitest run packages/core` — 5163/5164 passed; the one failure was the known Calendar flake (~line 884, 'February 2026'), which passed on a single rerun of that file
- `node scripts/check-changesets.mjs` — valid

**Honest limitation:** jsdom cannot emulate forced-colors rendering, so the new tests assert the injected CSS contains the `forced-colors: active` rules (jsdom serializes system colors lowercased, e.g. `background-color: highlight`), not the rendered result.

**Manual verification steps (Windows High Contrast):**

1. On Windows, enable a contrast theme (Settings → Accessibility → Contrast themes, e.g. "Night sky"), or in Chromium DevTools use Rendering → Emulate CSS media feature `forced-colors: active`.
2. Switch: off shows an empty bordered track with a solid thumb; on shows a Highlight-filled track with a contrasting thumb.
3. CheckboxInput: indeterminate bar visible inside the box; checked shows the checkmark.
4. RadioList: selected radio shows the inner dot.
5. SegmentedControl / ToggleButton: selected/pressed item is Highlight-filled with HighlightText content.
6. Skeleton: placeholder blocks remain visible (GrayText).
7. Confirm focus rings remain visible on all of the above.

## Notes

- The Vercel deployment failure on fork PRs is known infra and unrelated to this change.

### DIFF
--- a/.changeset/a11y-forced-colors.md
+++ b/.changeset/a11y-forced-colors.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] core: preserve state indication for painted controls (Switch, CheckboxInput, RadioList, SegmentedControl, ToggleButton, Skeleton) under forced colors / Windows High Contrast (WCAG 1.4.11)
+@bhamodi
--- a/packages/core/src/CheckboxInput/CheckboxInput.test.tsx
+++ b/packages/core/src/CheckboxInput/CheckboxInput.test.tsx
@@ -13,6 +13,7 @@ import {describe, it, expect, vi, beforeEach, afterEach} from 'vitest';
 import {render, screen, fireEvent, waitFor} from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import {CheckboxInput} from './CheckboxInput';
+import {getForcedColorsRules} from '../__tests__/forcedColors';
 import {__resetLiveRegionsForTest} from '../hooks/useAnnounce';
 
 afterEach(() => {
@@ -541,5 +542,20 @@ describe('CheckboxInput', () => {
         ...new FormData(container.querySelector('form')!).keys(),
       ]).toEqual([]);
     });
+  });
+});
+
+// jsdom cannot emulate forced-colors rendering, so this asserts that the
+// compiled output includes the forced-colors rule; visual behavior needs
+// manual verification under Windows High Contrast.
+describe('forced colors (WCAG 1.4.11)', () => {
+  it('compiles a forced-colors fill so the indeterminate mark survives Windows High Contrast', () => {
+    render(
+      <CheckboxInput label="All" value="indeterminate" onChange={() => {}} />,
+    );
+    // The painted indeterminate bar would be stripped to Canvas (invisible);
+    // CanvasText keeps it perceivable. The checkmark needs no rule — it
+    // strokes with currentColor, which forced colors maps for us.
+    expect(getForcedColorsRules()).toContain('background-color: canvastext;');
   });
 });
--- a/packages/core/src/CheckboxInput/CheckboxInput.test.tsx
+++ b/packages/core/src/CheckboxInput/CheckboxInput.test.tsx
@@ -554,8 +554,15 @@ describe('forced colors (WCAG 1.4.11)', () => {
       <CheckboxInput label="All" value="indeterminate" onChange={() => {}} />,
     );
     // The painted indeterminate bar would be stripped to Canvas (invisible);
-    // CanvasText keeps it perceivable. The checkmark needs no rule — it
-    // strokes with currentColor, which forced colors maps for us.
+    // CanvasText keeps it perceivable.
     expect(getForcedColorsRules()).toContain('background-color: canvastext;');
+  });
+
+  it('compiles a forced-colors color so the checkmark survives Windows High Contrast', () => {
+    render(<CheckboxInput label="Accept" value={true} onChange={() => {}} />);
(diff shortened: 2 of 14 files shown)


kentonquatman:
This is looking better, but I have a few notes:
1. `Switch` isn't showing the thumb on hover – it appears as white on white.
2. `CheckboxInput` isn't showing the selected state – it's also rendering a white icon over a white background.
3. `SegmentedControl` is showing a white block over the text on the active segment.
4. `ToggleButton` is also showing a white block over the text in the button label.

bhamodi:
@kentonquatman Thanks for testing this in a real forced-colors environment, that caught four things the jsdom-level tests couldn't. Root causes and fixes:

1. Switch thumb white-on-white on hover: the hover tint is a color-mix (a non-system color), and StyleX ranks its ancestor-hover rule above the plain forced-colors track rule, so on hover the UA flattened the color-mix back to Canvas and the HighlightText thumb ended up on a white track. I gated the tint behind `(hover: hover) and (forced-colors: none)` so the Highlight/Canvas track stands under forced colors and the thumb stays visible.

2. CheckboxInput white check on white box: the check is an SVG stroked with currentColor, and forced colors doesn't reliably force an SVG stroke, so it stayed the same white as the flattened box. I gave the checkmark its own CanvasText color under forced colors, matching the indeterminate bar (a painted div, which already worked).

3 and 4. SegmentedControl active segment and ToggleButton pressed label showing a white block over the text: both are `<button>` elements, and the UA keeps the native form-control surface (ButtonFace) and ignores the authored Highlight fill while the label keeps HighlightText, so you get white text on a white surface. The painted div controls (radio dot, checkbox bar, switch track) honor system-color backgrounds but the buttons don't. I added `forced-color-adjust: none` to the selected/pressed state so the button honors the Highlight surface + HighlightText label pair as authored. It's scoped to just that state, so every other state still participates fully in forced colors.

I strengthened the unit tests to assert the corrected pairs, but jsdom can't emulate forced-colors rendering, so I can't screenshot it from here. Would you mind giving it another pass in your High Contrast setup? The branch is updated.

kentonquatman (approved):
Looks great! Thanks @bhamodi !

cixzhang (approved):

PR #4353, opened 2026-07-26
fix(core): expose Avatar status in its accessible name

bhamodi:
## Summary

Fixes a serious WCAG 4.1.2 (Name, Role, Value) issue found in a11y scan #3: Avatar's root sets `role="img"`, which prunes ALL descendant semantics from the accessibility tree. The nested `AvatarStatusDot`'s `role="img"` + `aria-label` (e.g. "Online") was therefore never announced — status was invisible to assistive tech. The gap was already acknowledged in-code on `AvatarStatusDotProps.label` ("composing status into the avatar's accessible name is a planned Avatar-level fix"); this PR implements that plan.

Avatar now reads the status element's string `label` prop (the localized string consumers already pass to `AvatarStatusDot`) and composes it into its own accessible name via a new ICU catalog key `@astryx.avatar.nameWithStatus` (`{name}, {status}` → "Jane Doe, Online"), using the same `useTranslator()` pattern AvatarGroup already uses — no hardcoded English, and locales can reorder or change the separator.

## Changes

- `packages/core/src/Avatar/Avatar.tsx`: new `getStatusLabel()` helper reads a string `label` off the `status` element; the accessible name composes `alt || name` with the status label through `t('@astryx.avatar.nameWithStatus')`. A labelled status on an otherwise decorative avatar (no `name`/`alt`) now surfaces as the accessible name instead of being swallowed by `aria-hidden`. Aria wiring stays after `{...props}` (same precedence as #3903).
- `packages/core/locales/en.json`: adds `@astryx.avatar.nameWithStatus` (`{name}, {status}`) with translator guidance.
- `packages/core/src/Avatar/AvatarStatusDot.tsx`: updates the acknowledged-gap comment on `label` — inside an Avatar the label is composed into the avatar's name; standalone dots keep exposing `role="img"` directly (unchanged behavior).
- `packages/core/src/Avatar/Avatar.doc.mjs`, `AvatarStatusDot.doc.mjs`: doc sync (en/zh/dense) per the SYNC protocol.
- Tests: new `Avatar.test.tsx` suite covering image/initials/alt avatars with status, no-status and unlabelled-status cases, and decorative behavior; `AvatarGroup.test.tsx` asserts grouped avatars inherit the composition.

## Test plan

- `node_modules/.bin/vitest run packages/core/src/Avatar packages/core/src/AvatarGroup --reporter=dot` — 51 tests passed (3 files). The 5 new composition assertions were confirmed failing before the fix.
- `node_modules/.bin/vitest run packages/core --reporter=dot` — 5166 passed, 1 failed: the known `Calendar.test.tsx:884` flake ("February 2026"). It fails identically on a clean `main` checkout with the full file and passes in isolation on both branches — pre-existing, unrelated.
- `tsc --project tsconfig.json --noEmit` and `tsc --project tsconfig.docs.json` (packages/core) — clean.
- `eslint` on all changed source/test files — clean (i18n `no-hardcoded-i18n-string` rule passes; the composed string comes from the locale catalog).
- `prettier --check` on changed files — clean (`Avatar.doc.mjs` is not prettier-formatted at HEAD and was intentionally left in its existing style).

## Notes

- The Vercel deployment check failing on fork PRs is known infra and unrelated.

### DIFF
--- a/.changeset/a11y-avatar-status-name.md
+++ b/.changeset/a11y-avatar-status-name.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Avatar: compose the status dot's label into the avatar's accessible name so assistive tech can reach the status the `role="img"` root previously pruned (WCAG 4.1.2)
+@bhamodi
--- a/packages/core/locales/en.json
+++ b/packages/core/locales/en.json
@@ -235,6 +235,10 @@
     "defaultMessage": "Mobile navigation",
     "description": "Screen-reader-only accessible name for the mobile-only navigation region on small viewports. \"Mobile\" = phone/tablet (small screen), not \"movable\"."
   },
+  "@astryx.avatar.nameWithStatus": {
+    "defaultMessage": "{name}, {status}",
+    "description": "Screen-reader accessible name for an Avatar showing a status indicator; composes the person's name with the status label, e.g. \"Jane Doe, Online\". {name} = the avatar's name/alt text, {status} = the status dot's label. Adjust separator and order per locale."
+  },
   "@astryx.avatarGroup.label": {
     "defaultMessage": "Avatars",
     "description": "Screen-reader-only fallback name for a horizontal cluster of user avatar images. Plural noun; consumers usually override with \"Team members\", \"Attendees\", etc."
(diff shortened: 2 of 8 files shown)


cixzhang (approved):

cixzhang:
Looks good! There's some merge conflicts.

PR #4354, opened 2026-07-26
fix(core): guarantee WCAG contrast for generated color token pairs

bhamodi:
## Summary

Accessibility scan finding #3: `expandColorScale` assigns text/surface/border tones purely by HCT tone spacing as a proxy for contrast — the repo had no relative-luminance/contrast-ratio function and no test asserting any ratio (the existing test only checked that `contrast: 'high'` differs from `'standard'`). This PR adds a WCAG 2.x contrast utility and turns the implicit tone-spacing assumption into an asserted guarantee for every generated theme (WCAG 1.4.3 text contrast >= 4.5:1, WCAG 1.4.11 non-text contrast >= 3:1).

Two findings from measuring the generated tokens:

- **Text pairs pass by construction.** HCT tone is CIE L*, which fully determines relative luminance regardless of hue/chroma, so the fixed tone assignments guarantee >= 4.5:1 for text-on-surface pairs for *any* accent/neutralStyle — a hue sweep (24 hues x 3 neutral styles x 2 contrast levels x light/dark) never dropped below 5.6:1. No tone fixes were needed for text tokens; the tests now lock that invariant in.
- **`--color-border-emphasized` failed 3:1 in every generated theme** (~2.2:1 light, ~1.8:1 dark). It outlines form controls (CheckboxInput, Selector), so it is a genuine WCAG 1.4.11 boundary. Generation now tone-bumps it until it clears 3:1 against the generated surface (default accent: light tone 70 -> ~60, dark tone 30 -> ~44).

## Changes

- **`packages/core/src/theme/contrast.ts` (new)** — dependency-free WCAG 2.x `relativeLuminance`, `contrastRatio`, and `compositeOver` (sRGB source-over blending, matching CSS). Translucent foregrounds are composited over the background before measuring; translucent backgrounds are rejected so alpha tokens must be explicitly composited over their real backdrop.
- **`packages/core/src/theme/expandColorScale.ts`** — new `ensureContrastTone()` tone-bumping loop; `--color-border-emphasized` now derives from it (>= 3:1 vs. the generated surface, both modes). File header documents the guarantees and the deliberate exclusions.
- **`packages/core/src/theme/expandColorScale.test.ts`** — new "WCAG contrast guarantees" suite over 10 representative configs (default accent, both contrast levels, all neutral styles, saturated brand hues, low-chroma gray, black/white seeds) x light/dark:
  - text-primary >= 4.5:1 on surface/body/card/popover and on the alpha tokens `--color-background-muted`/`--color-neutral` composited over the surface
  - text-secondary >= 4.5:1 on surface/card/popover; accent-as-text >= 4.5:1 on surface; on-accent >= 4.5:1 on accent
  - accent, icon-primary, icon-secondary, border-emphasized >= 3:1 on surface
  - a test documenting the tokens deliberately excluded from 3:1, and unit tests for `ensureContrastTone` (no-op when already passing, bump-until-pass, termination at tone floor/ceiling).
- **`packages/core/src/theme/contrast.test.ts` (new)** — utility unit tests (black/white = 21:1, WCAG channel weights, canonical #767676 boundary gray, alpha compositing, error paths).
- **`.changeset/a11y-token-contrast-assertions.md`** — patch changeset.

**Deliberately excluded from the 3:1 assertion** (documented in the test): `--color-border` (decorative hairline, N[10] @ 0.1 alpha ~= 1.1:1 by design — components relying on it as their *only* boundary are tracked separately), `--color-skeleton` (loading placeholder), `--color-track` (rail behind a >= 3:1 indicator), and disabled text/icon tokens (explicit WCAG exemption). Note the hand-tuned static default palette's `--color-border-emphasized` (`#CCD3DB` light, ~1.5:1) has the same issue as the generated one but is out of scope here (only generation is changed); tracked with the border follow-up.

## Test plan

- `node_modules/.bin/vitest run packages/core/src/theme --reporter=dot` — 14 files, **585 passed** (54 new tests).
- `node_modules/.bin/vitest run packages/core --reporter=dot` — 223 files, **5211 passed, 1 failed**: the known `Calendar.test.tsx` RTL "February 2026" flake, which fails identically on a clean `main` checkout (date-dependent; unrelated).
- **Pre-fix failure confirmed**: with the `--color-border-emphasized` generation change reverted to `ld(NV[70], NV[30])`, the new suite fails 20 tests (the 3:1 non-text assertions across all 10 configs x 2 modes); with the fix it passes.
- `tsc --project packages/core/tsconfig.json --noEmit` — clean.
- `prettier --check` and `eslint` on all changed files — clean.
- `node scripts/check-changesets.mjs` — valid.
- Hue sweep sanity check (24 accent hues x 3 neutral styles x 2 contrast levels): worst-case generated pair is now border-emphasized at exactly 3.00:1; worst text pair 5.67:1.

## Notes

### DIFF
--- a/.changeset/a11y-token-contrast-assertions.md
+++ b/.changeset/a11y-token-contrast-assertions.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] theme: guarantee WCAG contrast for generated color token pairs — text-on-surface pairs are asserted at >= 4.5:1 and non-text UI pairs at >= 3:1 (WCAG 1.4.3/1.4.11), with `--color-border-emphasized` tone-bumped in generation until it clears 3:1 against the generated surface
+@bhamodi
--- a/packages/core/src/theme/contrast.test.ts
+++ b/packages/core/src/theme/contrast.test.ts
@@ -1,0 +1,94 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+import {describe, it, expect} from 'vitest';
+import {relativeLuminance, compositeOver, contrastRatio} from './contrast';
+import {parseColor} from '../utils/color';
+
+describe('relativeLuminance', () => {
+  it('is 0 for black and 1 for white', () => {
+    expect(relativeLuminance({r: 0, g: 0, b: 0, a: 1})).toBe(0);
+    expect(relativeLuminance({r: 255, g: 255, b: 255, a: 1})).toBeCloseTo(
+      1,
+      10,
+    );
+  });
+
+  it('weights channels per WCAG (green brightest, blue darkest)', () => {
+    const red = relativeLuminance({r: 255, g: 0, b: 0, a: 1});
+    const green = relativeLuminance({r: 0, g: 255, b: 0, a: 1});
+    const blue = relativeLuminance({r: 0, g: 0, b: 255, a: 1});
+    expect(red).toBeCloseTo(0.2126, 4);
+    expect(green).toBeCloseTo(0.7152, 4);
+    expect(blue).toBeCloseTo(0.0722, 4);
+  });
+});
+
+describe('contrastRatio', () => {
+  it('is 21 for black on white and 1 for identical colors', () => {
+    expect(contrastRatio('#000000', '#FFFFFF')).toBeCloseTo(21, 5);
+    expect(contrastRatio('#3B82F6', '#3B82F6')).toBe(1);
+  });
+
+  it('is symmetric in fg/bg for opaque colors', () => {
+    expect(contrastRatio('#0064E0', '#FCFDFE')).toBeCloseTo(
+      contrastRatio('#FCFDFE', '#0064E0'),
+      10,
+    );
+  });
+
+  it('matches the canonical 4.5:1 boundary gray (#767676 on white)', () => {
+    const ratio = contrastRatio('#767676', '#FFFFFF');
+    expect(ratio).toBeGreaterThanOrEqual(4.5);
+    expect(ratio).toBeLessThan(4.6);
+  });
+
+  it('composites a translucent foreground over the background', () => {
+    // 50% black over white paints as mid-gray, not black.
+    const composited = contrastRatio('rgba(0, 0, 0, 0.5)', '#FFFFFF');
+    expect(composited).toBeLessThan(contrastRatio('#000000', '#FFFFFF'));
+    expect(composited).toBeCloseTo(
+      contrastRatio('rgb(127.5, 127.5, 127.5)', '#FFFFFF'),
+      10,
+    );
+  });
+
+  it('rejects a translucent background', () => {
+    expect(() => contrastRatio('#000000', '#FFFFFF80')).toThrow(
+      /background must be opaque/,
+    );
+  });
+
+  it('rejects unparseable colors', () => {
+    expect(() => contrastRatio('var(--color-accent)', '#FFFFFF')).toThrow(
+      /could not parse foreground/,
+    );
+    expect(() => contrastRatio('#000000', 'oklch(0.5 0.1 200)')).toThrow(
+      /could not parse background/,
+    );
+  });
+});
+
+describe('compositeOver', () => {
+  it('returns the foreground when opaque and the backdrop at alpha 0', () => {
+    const fg = parseColor('#123456');
+    const bg = parseColor('#FFFFFF');
+    if (fg === null || bg === null) {
+      throw new Error('fixture colors must parse');
+    }
+    expect(compositeOver(fg, bg)).toEqual({...fg, a: 1});
+    expect(compositeOver({...fg, a: 0}, bg)).toEqual({...bg, a: 1});
+  });
+
+  it('blends in gamma-encoded sRGB space like CSS', () => {
+    const fg = parseColor('rgba(0, 0, 0, 0.5)');
+    const bg = parseColor('#FFFFFF');
+    if (fg === null || bg === null) {
+      throw new Error('fixture colors must parse');
+    }
+    const out = compositeOver(fg, bg);
+    expect(out.r).toBeCloseTo(127.5, 5);
+    expect(out.g).toBeCloseTo(127.5, 5);
+    expect(out.b).toBeCloseTo(127.5, 5);
+    expect(out.a).toBe(1);
+  });
+});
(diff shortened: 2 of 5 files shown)


cixzhang:
I think this seems reasonable though @ernestt and @rubyycheung could give this a closer look on how design wants to generate these. I think Ruby is also planning to redo color generation generally for other reasons like producing full palettes, etc. Though accessible generation likely fits into this vision as well.

Probably could be good to check playground and see a clear delta on generated colors and compare results here. I'm not fully sure what the LLM is using to make decisions about contrast here.

ernestt:
Conceptually this sounds like the right thing to do. But Im having a hard time seeing how it actually differs.  I simulated a "bad" accent seeding color but the result is near identical.  
CleanShot 2026-07-29 at 13 36 04

bhamodi:
@ernestt Good instinct to test a bad seed, and the near-identical result is actually the expected outcome -- this is a guardrail, not a re-generation. Text, accent, and background tones are placed at fixed HCT tones, and since HCT tone is CIE L* their luminance (and therefore contrast) is fixed regardless of the seed's hue/chroma, so they already pass and this PR does not touch their emitted values at all -- it only adds assertions that fail CI if a future change regresses them. The one token whose value actually changes is `--color-border-emphasized`, the resting boundary on form controls (Checkbox, Radio, Selector, text/date inputs), so a side-by-side of accent swatches or body text won't show anything. To see the delta in the playground, seed the default #0064E0 and look at an unchecked checkbox border or an unfocused text input border, light mode: it goes from #A9AAB9 (2.24:1 on the #FCFCFF surface, below the 3:1 non-text minimum) to #8F909E (3.08:1). Dark mode: #454652 (1.84:1) to #666774 (3.06:1). It's a subtly darker 1px gray, so it reads as "near identical" by eye but it's the difference between failing and passing WCAG 1.4.11. Worth noting this fires for the default accent too, so it's fixing our own baseline, not just pathological seeds. I added a before/after unit test (expandColorScale.test.ts) that reconstructs the old NV[70]/NV[30] output and locks in exactly this delta.

bhamodi:
@cixzhang On the "what is the LLM using to decide contrast" question: nothing -- there's no LLM in this path. Generation is fully deterministic HCT math. The contrast decisions come from two places: (1) the WCAG 2.x relative-luminance and contrast-ratio formulas in the new contrast.ts (dependency-free, with alpha compositing so translucent tokens are measured over their real backdrop), and (2) the fact that HCT tone is CIE L*, so the fixed text-tone assignments mathematically hold >= 4.5:1 for any hue/chroma. The only value adjusted at generation time is `--color-border-emphasized`: a deterministic loop (ensureContrastTone) steps HCT tone toward black/white until it clears 3:1, guaranteed to terminate because tone maps monotonically to luminance. For a clear playground delta: seed #0064E0 and compare form-control borders -- light #A9AAB9 (2.24:1) becomes #8F909E (3.08:1); dark #454652 (1.84:1) becomes #666774 (3.06:1). Everything else is byte-identical, which is the point. On Ruby's color-generation redesign: agreed this fits there. What's here is intentionally small and self-contained -- contrast.ts is just standalone WCAG math plus one tone-bump helper -- so when full-palette generation lands it can drop in as the accessibility gate on the new output rather than being re-litigated separately. Happy to have Ernest and Ruby take a closer look and to move the assertion into that work when it's ready.

humbertovirtudes (approved):
This is a strong change. The insight that HCT tone is CIE L* (so relative luminance is fixed by tone alone, independent of hue/chroma) is what lets the text-token guarantee hold for any accent/neutralStyle, and turning that implicit assumption into an asserted invariant across a representative config sweep is exactly the right move. Finding that `--color-border-emphasized` was actually below the WCAG 1.4.11 3:1 floor in every generated theme (~2.2:1 light, ~1.8:1 dark) is a real, user-facing catch, not a theoretical one.

Triage: core theme fix + test hardening, non-breaking (one generated token value corrected upward), low blast radius -> standard path.

- `vitest run packages/core/src/theme/contrast.test.ts packages/core/src/theme/expandColorScale.test.ts`: 63/63 pass (10 + 53).
- The tone-bump backdrop (`N[99]` light / `N[10]` dark) matches `--color-background-surface = ld(N[99], N[10])` exactly, so the generated border is measured against the surface it actually sits on.
- No new public API surface: `contrast.ts` (`relativeLuminance`/`compositeOver`/`contrastRatio`) and `ensureContrastTone` are internal, not re-exported from `theme/index.ts`; `expandColorScale`'s signature is unchanged (the only edit to its declaration line is a prettier reflow). So this is a behavior/assertion change, not an API change.
- 35 commits behind `main`, but nothing on `main` has touched `theme/expandColorScale.ts` or `contrast.ts` since the branch point, so it stays clean.

What I like:
- The contrast math is dependency-free and correct: the sRGB linearization, the 0.2126/0.7152/0.0722 weights, and the `(L1+0.05)/(L2+0.05)` ratio are all to spec, and the tests pin the canonical anchors (21:1 black/white, the 4.5:1 `#767676` boundary, channel weights).
- Compositing translucent foregrounds over the backdrop before measuring (and *rejecting* a translucent background with a clear error) is the right call: an alpha token has no contrast of its own, only against what it paints on. The `--color-background-muted`/`--color-neutral` alpha tokens are composited over the surface in the sweep rather than measured raw.
- The before/after test reconstructs the exact pre-change output (`NV[70]`/`NV[30]`) and locks the delta, so the correction cannot silently regress to the sub-3:1 value.
- The excluded-tokens test (`--color-border` hairline, `--color-skeleton`, `--color-track`, disabled tokens) documents the deliberate WCAG exemptions in code and will trip if a future change starts leaning on the hairline for contrast.

Non-blocking nits (fine to land as-is):
- `ensureContrastTone` termination rests on tone stepping monotonically in luminance; the comment says so and the "impossible ratio stops at black/white" test covers the boundary, so this is well-guarded, just worth keeping in mind if a non-monotonic tone source is ever passed.
- The excluded-token assertion checks only the default accent config; the text/non-text sweeps cover all configs, so this is a reasonable narrowing, not a gap.

Changeset is present, patch, scrubbed, and correctly scoped to the consumer-visible token change.

---
PR #4354 - fix(core): guarantee WCAG contrast for generated color token pairs
Author: @bhamodi
What it does: adds a dependency-free WCAG 2.x contrast utility and turns the theme generator's implicit tone-spacing assumption into asserted guarantees (text >= 4.5:1, non-text >= 3:1) across a representative config sweep; corrects `--color-border-emphasized`, the one generated token that fell below 3:1, by tone-bumping it against the generated surface.
Recommendation: Approve.

PR #4356, opened 2026-07-26
fix(core): constrain range Slider thumb ARIA bounds; group labeling

bhamodi:
## Summary

Two Slider accessibility fixes from a11y scan #3:

- **Range thumbs reported wrong bounds (ARIA APG Multi-Thumb Slider).** In range mode both thumbs exposed the full `aria-valuemin={min}` / `aria-valuemax={max}`, telling screen-reader users the movable range is larger than it actually is. Per the APG multi-thumb pattern, each thumb's ARIA bounds are now constrained by its sibling thumb (including the `minStepsBetweenThumbs` gap), matching exactly what the movement-clamping logic in `updateValue` allows: the lower thumb's `aria-valuemax` is the upper thumb's value minus the min gap, and the upper thumb's `aria-valuemin` is the lower thumb's value plus the min gap (both clamped to `[min, max]`).
- **Orphaned/incorrect label association (WCAG 1.3.1).** Slider never passed `isGroupLabel` to Field, so Field rendered a real `<label htmlFor={id}>`. In range mode no element carried that id, and in single mode `htmlFor` pointed at a `div role="slider"`, which is not a labelable element -- the label was inert either way. The label now renders as a group label (a `<span>` with `labelID`), following the existing InputGroup/RadioList pattern.

## Changes

- `packages/core/src/Slider/Slider.tsx`
  - Range thumbs: `aria-valuemin`/`aria-valuemax` are sibling-constrained (including `minStepsBetweenThumbs * step`), mirroring the clamp logic in `updateValue` so ARIA and behavior agree exactly.
  - Field now gets `labelID` + `isGroupLabel`, so no `<label htmlFor>` is emitted.
  - Single mode: the thumb is named via `aria-labelledby={labelID}` (previously `aria-label={label}`).
  - Range mode: the `role="group"` container is named via `aria-labelledby={labelID}` (previously `aria-label={label}`), and each thumb keeps a short individual name ("Minimum value" / "Maximum value") that composes with the group label per APG, instead of duplicating the full label.
- `packages/core/src/Slider/Slider.test.tsx`
  - Extended the range aria-values test to assert sibling-constrained `aria-valuemin`/`aria-valuemax`.
  - New tests: bounds update after moving a thumb (controlled), bounds include the `minStepsBetweenThumbs` gap, no orphaned `<label htmlFor>` (any `label[for]` must point at an existing labelable element), single thumb named by the label element, and range group `aria-labelledby` the label.
- `.changeset/a11y-slider-range-aria.md` -- patch changeset.

## Test plan

- New/updated tests confirmed failing pre-fix: `vitest run packages/core/src/Slider` -> 6 failed | 34 passed.
- Post-fix: `node_modules/.bin/vitest run packages/core/src/Slider packages/core/src/Field --reporter=dot` -> 4 files, 110 tests passed.
- Full core suite: `node_modules/.bin/vitest run packages/core --reporter=dot` -> 221/222 files, 5161/5162 tests passed. The single failure is the known Calendar flake (`Calendar.test.tsx` ~line 884, duplicate "February 2026"); it fails identically on a pristine `main` checkout, so it is unrelated to this change.
- `node_modules/.bin/tsc --project packages/core/tsconfig.json --noEmit` -> clean.
- `prettier --check` and `eslint` on changed files -> clean.
- `node scripts/check-changesets.mjs` -> 45 changesets valid.

## Notes

- Vercel fork-PR deployment failure is known infra and can be ignored.

### DIFF
--- a/packages/core/src/Slider/Slider.tsx
+++ b/packages/core/src/Slider/Slider.tsx
@@ -392,6 +392,7 @@ export function Slider({ref, ...props}: SliderProps) {
   const isHorizontal = orientation === 'horizontal';
 
   const id = useId();
+  const labelID = useId();
   const descriptionID = useId();
   const statusMessageID = useId();
   const requiredID = useId();
@@ -693,11 +694,25 @@ export function Slider({ref, ...props}: SliderProps) {
       ? {left: `${percent}%`}
       : {bottom: `${percent}%`, left: '50%'};
 
+    // In range mode each thumb keeps a short individual name that composes
+    // with the group label (announced via the group's aria-labelledby), per
+    // the APG multi-thumb slider pattern. In single mode the thumb takes its
+    // name from the visible label element via aria-labelledby.
     const thumbLabel = isRange
       ? thumbIndex === 0
-        ? `${label}, minimum value`
-        : `${label}, maximum value`
-      : label;
+        ? 'Minimum value'
+        : 'Maximum value'
+      : undefined;
+
+    // ARIA bounds must agree with the movement clamping in updateValue: in
+    // range mode a thumb can't cross its sibling (minus the
+    // minStepsBetweenThumbs gap), and the result is always clamped to
+    // [min, max].
+    const minGap = minStepsBetweenThumbs * step;
+    const ariaValueMin =
+      isRange && thumbIndex === 1 ? clamp(values[0] + minGap, min, max) : min;
+    const ariaValueMax =
+      isRange && thumbIndex === 0 ? clamp(values[1] - minGap, min, max) : max;
 
     // Suppress the per-thumb value bubble while the disabled-message tooltip is
     // showing, so a disabled slider surfaces the *reason* on hover/focus rather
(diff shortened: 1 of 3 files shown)


packages/core/src/Slider/Slider.tsx
```diff
@@ -693,11 +694,25 @@ export function Slider({ref, ...props}: SliderProps) {
...
+    // In range mode each thumb keeps a short individual name that composes
+    // with the group label (announced via the group's aria-labelledby), per
+    // the APG multi-thumb slider pattern. In single mode the thumb takes its
+    // name from the visible label element via aria-labelledby.
     const thumbLabel = isRange
       ? thumbIndex === 0
-        ? `${label}, minimum value`
-        : `${label}, maximum value`
-      : label;
+        ? 'Minimum value'
```
cixzhang, line 703: This is fine but likely needs translation.

cixzhang (approved):

PR #4358, opened 2026-07-26
fix(core): close ContextMenu on Tab; require label on DropdownMenuRadioGroup

bhamodi:
## Summary

Two menu accessibility fixes from the a11y scan (#3):

1. **ContextMenu did not close on Tab.** The ARIA APG menu pattern requires Tab to dismiss an open menu. ContextMenu's `listKeyDown` handled Enter/Space, typeahead, and arrows but had no Tab handler; since menu items are `tabIndex={-1}`, pressing Tab moved focus out into the page while the menu stayed open and visible. DropdownMenu already handles this correctly (the menus-5 handler) -- this PR mirrors that exact behavior: close the menu on Tab without `preventDefault`, so closing restores focus and the browser's default Tab continues from there.
2. **DropdownMenuRadioGroup could ship unnamed.** The group renders `role="group"` and previously took its accessible name from an optional `aria-label`/`aria-labelledby` passthrough, so a group with neither was announced as an unnamed set. Following review, the group now takes a **required `label` prop** (applied as `aria-label`), matching `RadioList`'s required `label: string` convention. TypeScript now enforces a name at every call site, so the earlier hand-rolled dev warning is removed as redundant (`RadioList` carries no such runtime warning either). Consumers who already render a visible label element can pass `aria-labelledby` via base props instead.

## Changes

- `packages/core/src/ContextMenu/ContextMenu.tsx` -- `listKeyDown` now closes the menu on Tab (APG menu pattern, mirrors DropdownMenu menus-5).
- `packages/core/src/ContextMenu/ContextMenu.test.tsx` -- test: open via right-click, press Tab on the menu, assert `hidePopover` was called (plus updated the radio-group usages to pass `label`).
- `packages/core/src/DropdownMenu/DropdownMenuRadioGroup.tsx` -- `label` is now a **required** prop, applied as the group's `aria-label`; the previous dev warning (and its `useEffect`/`useRef` plumbing) is removed.
- `packages/core/src/DropdownMenu/DropdownMenuRadioGroup.doc.mjs` -- documents the required `label` prop (replacing the `aria-label`/`aria-labelledby` entries) across all doc variants.
- `packages/core/src/DropdownMenu/DropdownMenuSelectable.test.tsx` -- asserts the group is named from `label`; the two obsolete warning tests are removed.
- Updated all internal usages of the group and its re-export aliases (`ContextMenuRadioGroup`, `BreadcrumbMenuRadioGroup`) to pass `label`: `apps/storybook/stories/DropdownMenu.stories.tsx`, `apps/storybook/stories/ContextMenu.stories.tsx`, `packages/core/src/Breadcrumbs/Breadcrumbs.test.tsx`, and the JSDoc example in `DropdownMenuRadioItem.tsx`.
- `.changeset/a11y-context-menu-tab.md` -- reclassified as `[breaking]`/`minor` (per the repo's 0.x convention, a required prop is a breaking API change); migration is `aria-label="..."` -> `label="..."`.

## Test plan

- Scoped vitest -- `DropdownMenuSelectable`, `ContextMenu`, and `Breadcrumbs` test files: 85/85 passed.
- `tsc --noEmit` on `packages/core` -- clean (0 errors), including the changed prop interface and all call sites.
- `eslint` on the changed source/test/story files -- clean.
- `prettier --check` on the changed files (excluding `.doc.mjs`, which is prettier-exempt) -- clean.
- Rebased onto `upstream/main`; the only conflict was the `listKeyDown` `useCallback` dependency array in `ContextMenu.tsx` (the new DropdownMenu submenus feature added `ownsEvent`), resolved to keep both `ownsEvent` and this branch's `closeMenu`.
- Full core suite and screenshots left to CI (final gate).

## Notes

- Making `label` required is a source-level breaking change for any consumer passing `aria-label`/`aria-labelledby` directly, so the changeset is `[breaking]`/`minor`. The design system is pre-1.0 and an unnamed radio group is an accessibility defect; the migration is a one-prop rename.

### DIFF
--- a/packages/core/src/DropdownMenu/DropdownMenuRadioGroup.tsx
+++ b/packages/core/src/DropdownMenu/DropdownMenuRadioGroup.tsx
@@ -14,7 +14,7 @@
  * menuitemradio rows.
  */
 
-import {useEffect, useMemo, useRef, type ReactNode} from 'react';
+import {useMemo, type ReactNode} from 'react';
 import * as stylex from '@stylexjs/stylex';
 import {spacingVars} from '../theme/tokens.stylex';
 import {mergeProps} from '../utils';
@@ -49,16 +49,13 @@ export interface DropdownMenuRadioGroupProps extends Omit<
    */
   onChange: (value: string) => void;
   /**
-   * Accessible label for the group. Required (together with `aria-labelledby`,
-   * one of the two) so screen readers announce the radios as a named set,
-   * e.g. "Sort by".
+   * Accessible name for the group, announced by screen readers so the radios
+   * read as a named set (e.g. "Sort by"). Applied as the group's `aria-label`.
+   * Required -- an unnamed radio group is an accessibility defect. Pass
+   * `aria-labelledby` (via base props) instead if the name already exists as a
+   * visible element on the page.
    */
-  'aria-label'?: string;
-  /**
-   * The id of an element that labels the group, as an alternative to
-   * `aria-label`.
-   */
-  'aria-labelledby'?: string;
+  label: string;
   /**
    * Whether selecting a value closes the menu. Radio items default to closing
    * on selection (a single-choice commit), unlike checkbox items which stay
@@ -82,7 +79,7 @@ export interface DropdownMenuRadioGroupProps extends Omit<
  *   DropdownMenuRadioItem,
  * } from '@astryxdesign/core/DropdownMenu';
  * <DropdownMenu button={{label: 'Sort'}}>
- *   <DropdownMenuRadioGroup value={sort} onChange={setSort} aria-label="Sort by">
+ *   <DropdownMenuRadioGroup value={sort} onChange={setSort} label="Sort by">
  *     <DropdownMenuRadioItem value="newest" label="Newest" />
  *     <DropdownMenuRadioItem value="oldest" label="Oldest" />
  *   </DropdownMenuRadioGroup>
@@ -92,6 +89,7 @@ export interface DropdownMenuRadioGroupProps extends Omit<
 export function DropdownMenuRadioGroup({
   value,
   onChange,
+  label,
   hasCloseOnSelect = true,
   children,
   className,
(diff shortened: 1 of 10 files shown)


packages/core/src/DropdownMenu/DropdownMenuRadioGroup.tsx
```diff
@@ -103,6 +103,23 @@ export function DropdownMenuRadioGroup({
     [value, onChange, hasCloseOnSelect],
   );
 
+  // Dev-time guardrail: the group should always have an accessible name so
+  // screen readers announce the radios as a named set. Warn once per component
+  // instance (in an effect) rather than on every render — mirrors the
+  // unnamed-dialog warnings in usePopover and Dialog.
+  const hasName = rest['aria-label'] != null || rest['aria-labelledby'] != null;
+  const warnedUnnamedGroupRef = useRef(false);
```
cixzhang: I made a new util for this. I think something like `useDevWarning`.

packages/core/src/DropdownMenu/DropdownMenuRadioGroup.tsx
cixzhang: Possibly the better solution might be to add a required `label` prop.

cixzhang (changes requested):
Back to you for thoughts on requiring label.

packages/core/src/DropdownMenu/DropdownMenuRadioGroup.tsx
bhamodi: Good call on the util. I ended up going the other way here though: per your other comment I made `label` a required prop, so TypeScript now enforces the name at every call site and the runtime warning became redundant. I removed it entirely rather than porting it to useDevWarning, which keeps us consistent with RadioList (its required `label: string` has no runtime warning either). So there's no useDevWarning call left in this component, but it's a great tool to have for the next case where a warning can't be type-enforced.

packages/core/src/DropdownMenu/DropdownMenuRadioGroup.tsx
bhamodi: Agreed, this is the better solution and I've implemented it. `label` is now a required `string` applied as the group's `aria-label`, matching how RadioList requires `label: string` and how the data-mode section groups in renderDropdownItems name themselves via aria-label. Consumers who already render a visible label element can still pass `aria-labelledby` through base props. Because the name is now type-enforced, I dropped the dev warning entirely (see my reply on the other thread). I also updated all internal usages plus the ContextMenuRadioGroup and BreadcrumbMenuRadioGroup re-exports, the stories, the tests, and the doc.mjs. On requiring it: fully on board. We're pre-1.0, an unnamed radio group is a real a11y defect, and the migration is a one-prop rename (aria-label -> label), so I bumped the changeset to [breaking]/minor to signal it.

cixzhang (approved):

PR #4360, opened 2026-07-26
fix(core): announce Toasts via persistent live regions

bhamodi:
## Summary

Toasts were not reliably announced to screen readers (WCAG 4.1.3 Status Messages; a11y scan finding #3). Each toast mounts its `role="status"`/`role="alert"` + `aria-live` element *together with its text content* — the "born with content" anti-pattern that the repo's own `hooks/useAnnounce.ts` documentation explicitly calls out as unreliable: many screen readers will not announce content that is already present when a live region is created.

This PR mirrors each toast's text through the persistent singleton live regions from `useAnnounce` when the toast is added to the viewport, so the announcement comes from a region that was mounted empty and only mutated afterwards — the pattern that actually works across AT. The per-toast `role`/`aria-live` markup is kept for browse-mode discoverability; the singleton is what guarantees the announcement.

## Changes

- `packages/core/src/Toast/ToastViewport.tsx`
  - New effect (client-only, so SSR-safe) that announces each toast's flattened text via `useAnnounce` when the toast list changes.
  - Politeness matches the existing per-toast role mapping: `type: 'error'` (rendered as `role="alert"`) announces assertively; everything else (`role="status"`) announces politely.
  - Dedupes by toast id + extracted text, so a toast is announced once when added and once per content update (`uniqueID` overwrite), never on unrelated viewport re-renders (another toast arriving, exit transitions).
  - `getNodeText()` flattens the `body` ReactNode (title + description etc.) to plain text — consumer-provided text is announced, not JSX; no new user-facing strings.
- `packages/core/src/Toast/ToastViewport.test.tsx`
  - New `toast announcements via singleton live regions` suite: polite for info toasts (including flattened JSX bodies), assertive for error toasts, re-announce on `uniqueID` overwrite, and no re-announcement of an unchanged toast on unrelated renders.
  - Scoped the pre-existing blur-timer test's text queries to the notifications region (the toast text now legitimately also appears in the singleton live region) and reset the singleton regions between tests.
- `.changeset/a11y-toast-announce.md` — patch changeset.

## Test plan

- `node_modules/.bin/vitest run packages/core/src/Toast --reporter=dot` — 2 files, 15/15 pass.
- Pre-fix failure confirmed: with the tests in place and `ToastViewport.tsx` reverted to `main`, all 4 new announcement tests fail (the singleton regions never receive the toast text); they pass with the fix.
- `node_modules/.bin/tsc --project packages/core/tsconfig.json --noEmit` — clean.
- `node_modules/.bin/eslint` + `prettier --check` on changed files — clean.
- Full core suite: `node_modules/.bin/vitest run packages/core --reporter=dot` — 5161/5162 passed; the single failure was the known Calendar flake (~line 884, 'February 2026'), which passes on an isolated rerun of the file (104/104).

## Notes

### DIFF
--- a/packages/core/src/Toast/ToastViewport.tsx
+++ b/packages/core/src/Toast/ToastViewport.tsx
@@ -185,21 +185,54 @@ export function ToastViewport({
     [],
   );
 
-  const addToast = useCallback((entry: ToastEntry) => {
-    setToasts(prev => {
+  // Announce toasts through the persistent singleton live regions. Each
+  // <Toast> also renders its own role="status"/"alert" region, but that region
+  // is "born with content" — mounted together with its text — which many
+  // screen readers do not announce (see useAnnounce.ts); the singleton regions
+  // are mounted empty and only mutated afterwards, so they are what actually
+  // guarantees the announcement (the per-toast markup is kept for browse-mode
+  // discoverability). The announcement happens in addToast — the imperative
+  // dispatch path invoked once per useToast() call from an event handler,
+  // never from render — so each toast is announced exactly once by
+  // construction, independent of the React render lifecycle (StrictMode
+  // double-render/double-effect, viewport remounts, and unrelated list
+  // re-renders never re-announce). It is client-only (addToast never runs
+  // during SSR), so it is SSR-safe.
+  const announce = useAnnounce();
+
+  const addToast = useCallback(
+    (entry: ToastEntry) => {
       const {uniqueID, collisionBehavior = 'overwrite'} = entry.options;
-      if (uniqueID) {
-        const existing = prev.find(t => t.options.uniqueID === uniqueID);
-        if (existing) {
-          if (collisionBehavior === 'ignore') {
-            return prev;
-          }
-          return prev.map(t => (t.options.uniqueID === uniqueID ? entry : t));
-        }
+      // Resolve an ignored collision synchronously against the committed list
+      // so a suppressed toast is neither shown nor announced. The remaining
+      // announce + setToasts run outside the setToasts updater, which React may
+      // invoke more than once — keeping the announcement exactly-once.
+      if (
+        uniqueID &&
+        collisionBehavior === 'ignore' &&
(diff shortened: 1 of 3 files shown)


packages/core/src/Toast/ToastViewport.tsx
```diff
@@ -276,6 +302,41 @@ export function ToastViewport({
...
+        announce(text, entry.options.type === 'error' ? 'assertive' : 'polite');
+      }
+    }
+    // Forget removed toasts so their ids don't accumulate.
+    for (const id of announced.keys()) {
+      if (!liveIds.has(id)) {
+        announced.delete(id);
+      }
+    }
+  }, [toasts, announce]);
```
cixzhang: An effect for announcing toasts seems like it can run into issues especially with duplicate announcements. Any way we can manage announcements for toasts via the useToast callback instead where we're more guaranteed to run it once instead of depending on React lifecycle.

cixzhang (changes requested):
Back to you for thoughts on eliminating the effect for announcements

packages/core/src/Toast/ToastViewport.tsx
bhamodi: Good call -- I moved the announcement out of the render effect entirely. It now happens in addToast, the imperative dispatch path that runs once per useToast() call from an event handler (never from render), so each toast is announced exactly once by construction instead of depending on React's lifecycle. That closes the duplicate-announcement door: StrictMode's double render and double-invoked state updater, viewport remounts, and unrelated list re-renders can no longer re-announce, and since addToast never runs during SSR it stays SSR-safe. Collision semantics are preserved -- an ignored uniqueID collision is resolved synchronously against the current list before announcing (so a suppressed toast is neither shown nor announced), while an overwrite is a fresh dispatch and re-announces. The persistent singleton live regions are still the announcement sink. I updated the tests to prove announce-once by spying on that sink: exactly one call under StrictMode, no re-announce on an unrelated re-render, a re-announce on uniqueID overwrite, and no announce on an ignored collision.

cixzhang (approved):

PR #4361, opened 2026-07-26
fix(core): render Token remove button as a sibling of its link

bhamodi:
## Summary

Found in a11y scan #3. When a `Token` has both `href` and `onRemove`, the remove `<button aria-label="Remove {label}">` rendered inside the `<a>`/LinkComponent — an interactive element nested inside another interactive element. This violates the HTML content model (anchors must not contain interactive content) and WCAG 4.1.2 (Name, Role, Value): screen readers and keyboard navigation behave unpredictably when a button lives inside a link (double tab stops, merged accessible names, Enter ambiguity).

The `onClick` branch already solved this with the invisible-control pattern: a `<span>` container carries the visual styles and the remove button renders as a sibling of an invisible `<button>`. This PR mirrors that exact structure for the `href` branch.

## Changes

- `packages/core/src/Token/Token.tsx`
  - When `href` and `onRemove` are both provided, Token now renders a `<span>` container (same style set as the onClick branch: `base` + size + color + `interactive` + `focusVisibleOutline`) with an invisible `<a>` (`invisibleButton` style) wrapping only the label, and the remove button as a **sibling** of the link.
  - Container clicks outside the link/button delegate to the anchor (`querySelector('a').click()`), matching the onClick branch's container-click delegation so the whole token remains a navigation target.
  - Link-only tokens (no `onRemove`) are unchanged — the anchor is still the root element and still receives the forwarded ref.
  - Extracted the duplicated remove-button JSX into a single `removeButton` element shared by all branches (no behavior change).
  - Updated the component JSDoc to document the new structure.
- `packages/core/src/Token/Token.test.tsx` — new `Token link with remove button` suite (6 tests).
- `.changeset/a11y-token-remove-link.md` — patch changeset for `@astryxdesign/core`.

## Test plan

New tests assert:

- the remove button is NOT a descendant of the link (`!link.contains(removeButton)`)
- link and remove button are independently keyboard-operable (Tab reaches link then button; Enter on the button fires `onRemove` without activating the link)
- the wrapper `<span>` class structure exactly matches the onClick branch's container (`className` equality)
- container clicks outside the anchor still activate the link; remove-button clicks do not
- link-only tokens (no `onRemove`) still render the anchor as the root

Pre-fix failure confirmed: 4 of the new tests fail on the old implementation (nested button found inside the link, wrapper mismatch, click delegation, remove-click leaking to the link).

Commands run:

- `prettier --check` / `--write` on changed files — clean
- `eslint packages/core/src/Token/Token.tsx packages/core/src/Token/Token.test.tsx` — clean (initial `useRef` approach was rejected by `@astryx/presentational-component`; final fix uses event-scoped `querySelector` instead)
- `tsc --project packages/core/tsconfig.json --noEmit` — clean
- `vitest run packages/core/src/Token packages/core/src/Tokenizer` — 95/95 passed (Tokenizer consumes Token heavily)
- `vitest run packages/core` — 5162 passed, 2 failed: both are the known Calendar flake (`Calendar.test.tsx:884`, duplicate "February 2026") and an occasional Typeahead flake; Typeahead passed on rerun, Calendar failure is pre-existing and unrelated (this diff only touches Token files)
- `node scripts/check-changesets.mjs` — 45 changesets valid

## Notes

- Vercel deployment failure on fork PRs is known infra and can be ignored.

### DIFF
--- a/packages/core/src/Token/Token.tsx
+++ b/packages/core/src/Token/Token.tsx
@@ -9,6 +9,7 @@
  * @position Core implementation; consumed by index.ts, tested by Token.test.tsx
  *
  * SYNC: When modified, update these files to stay in sync:
+ * - /packages/core/src/Token/TokenLink.tsx (interactive wrapper for href+onRemove)
  * - /packages/core/src/Token/Token.doc.mjs (props table, features, implementation notes)
  * - /packages/core/src/Token/Token.test.tsx (tests for new/changed behavior)
  * - /packages/core/src/Token/index.ts (exports if types change)
@@ -32,6 +33,7 @@ import {Icon} from '../Icon';
 import {mergeProps} from '../utils';
 import {useLinkComponent} from '../Link/useLinkComponent';
 import {useInteractiveRole} from '../hooks/useInteractiveRole';
+import {TokenLink} from './TokenLink';
 import type {BaseProps} from '../BaseProps';
 import {themeProps} from '../utils/themeProps';
 import {useTranslator} from '../i18n';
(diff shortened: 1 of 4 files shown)


packages/core/src/Token/Token.tsx
```diff
@@ -364,11 +368,47 @@ export function Token({
...
+          {content}
+        </LinkComponent>
+      );
+    }
+
+    // With a remove button, the link cannot be the root — nesting a <button>
+    // inside an <a> is invalid HTML (WCAG 4.1.2). Mirror the onClick branch:
+    // a <span> container with an invisible link and the remove button as
+    // siblings. Container clicks outside the link delegate to it.
+    const handleContainerClick = (e: React.MouseEvent) => {
```
cixzhang: You can useClickableContainer helper for this and it would handle things like middle click and open in new tab behaviors for links.

cixzhang (changes requested):

packages/core/src/Token/Token.tsx
bhamodi: Good call -- done. I swapped the hand-rolled container click handler for useClickableContainer, so the token surface now delegates to the inner link and gets middle-click and cmd/ctrl+click "open in new tab" for free, plus the text-selection guard and the built-in "ignore clicks on nested interactive elements" so the remove button is unaffected. One wrinkle: Token is on the @astryx/presentational-component list, which forbids useRef/effects in Token.tsx (that's why this was hand-rolled originally). To use the helper I pulled the ref + hook wiring into a small internal TokenLink wrapper and have Token render it for the href+onRemove case -- same split we already use for presentational Card vs. interactive ClickableCard. New tests in Token.test.tsx cover the delegated paths: container click activates the link, remove-button click does not, and middle-click / cmd+click open a new tab without firing onRemove.

cixzhang:
Double checked on playground, looks good. Thanks!

cixzhang (approved):

PR #3816, opened 2026-07-11
perf(test): make role+name queries O(match) in the date-component suites

Han5991:
## Summary

The four date-component suites were the slowest files in packages/core, and profiling showed why: their month grid keeps **~35–85 `role=button` nodes mounted** (visible in Calendar's case, inside the closed popover for the `*Input` components), and RTL's `getByRole('button', {name})` computes the accessible name of **every** candidate before filtering — each computation consulting jsdom's slow `getComputedStyle` (~5ms/node). One trigger lookup: ~450ms. A name-less `getAllByRole('button', {hidden: true})`: 29ms.

## Changes

- New shared helper `packages/core/src/__tests__/fastRoleQueries.ts` (same convention as `TestIcon.tsx`): `getButton()`/`queryButton()` keep RTL's exact name algorithm (`dom-accessibility-api`, the library RTL uses internally) but collect candidates with `{hidden: true}` and inject a constant "visible" style stub into the name computation — under `{hidden: true}` semantics visibility must not exclude candidates, so the stub removes only the `getComputedStyle` cost. Documented trade-off: first-match-wins, no tree-wide uniqueness check.
- All button role+name queries in the four suites now use the helpers (41 call sites). Queries that were already cheap (combobox/tooltip/grid — few candidates or no name filter) stay stock RTL.
- `dom-accessibility-api` declared as a root devDependency: it already ships in node_modules as a transitive dep, but without the declaration the hoisted copy is RTL's 0.5.16, whose `exports` map has **no types condition** — `tsc` fails with TS7016 (`implicitly has an 'any' type`), which is exactly what broke this PR's typecheck when the declaration was briefly dropped. `^0.6.3` pins the copy whose exports carry types.

## Measurements (local, sequential runs on the same machine)

| file | before | after |
|---|---|---|
| DateRangeInput.test.tsx (34 tests) | 34.3s | **1.3s** |
| Calendar.test.tsx (45 tests) | 7.2s | **1.9s** |
| DateInput.test.tsx (68 tests) | 5.1s | **2.2s** |
| DateTimeInput.test.tsx (64 tests) | 4.1s | **1.8s** |

Full suite: **5,893 tests green**, wall 105.9s → 92.8s, worker-summed `tests` phase 297.7s → 235.1s. Core typecheck green. All tests assert the same things they did before.

PR #4363, opened 2026-07-26
fix(core): announce date/time stepping; localize invalid-value messages; long tz names for AT

bhamodi:
## Summary

Three date/time accessibility fixes from a11y scan #3:

1. **Silent arrow-key stepping (WCAG 4.1.2, moderate).** `TimeInput` and the time slot of `DateTimeInput` step the time on ArrowUp/ArrowDown by programmatically rewriting a plain text input's value. Screen readers do not announce programmatic textbox changes, so stepping was completely silent — a screen-reader user pressing ArrowUp had no idea the value changed or what it changed to. Each step now announces the formatted new value politely through the persistent `useAnnounce` live regions.
2. **Hardcoded English invalid-value announcements (i18n, minor).** The WCAG 3.3.1 live-region alerts "Invalid date" / "Invalid time" in `DateInput`, `TimeInput`, and `DateTimeInput` were hardcoded English while every other string in these components resolves through the i18n catalog. They now route through `useTranslator()` via two new catalog keys, `@astryx.dateInput.invalidDate` and `@astryx.timeInput.invalidTime` (shared by `DateTimeInput`, matching how it already reuses `@astryx.dateInput.*` keys).
3. **Timezone abbreviation in the AT-facing label (WCAG 3.1.4, minor).** `Timestamp`'s aria-label used `Intl` `timeZoneName: 'short'`, exposing unexpanded abbreviations like "PST" or "GMT+2" to assistive technology. The aria-label now uses `timeZoneName: 'long'` ("Pacific Standard Time") while visible text and the tooltip keep the compact short form.

## Changes

- `TimeInput.tsx`: announce the formatted new time after each ArrowUp/ArrowDown step; invalid-time alert resolves `@astryx.timeInput.invalidTime`.
- `DateTimeInput.tsx`: add `useAnnounce`, announce time-slot stepping; invalid date/time alerts resolve the new catalog keys.
- `DateInput.tsx`: invalid-date alert resolves `@astryx.dateInput.invalidDate`.
- `Timestamp.tsx`: `getFullAbsoluteString` takes a `timeZoneName` form; aria-label uses the long form, tooltip/visible text keep short.
- `locales/en.json`: new `@astryx.dateInput.invalidDate` and `@astryx.timeInput.invalidTime` entries.
- Tests: stepping-announcement coverage for both components, catalog-override coverage for all three invalid alerts, long-vs-short timezone coverage for Timestamp (queries follow the existing O(match) patterns from #3816).

## Test plan

- New tests written first and confirmed failing pre-fix: 8 failed / 209 passed across the four suites before the source changes.
- `node_modules/.bin/vitest run packages/core/src/TimeInput packages/core/src/DateTimeInput packages/core/src/DateInput packages/core/src/Timestamp --reporter=dot` — 4 files, **217/217 passed** post-fix.
- `node_modules/.bin/vitest run packages/core --reporter=dot` — **5165/5167 passed**; the 2 failures are the known flaky/pre-existing Calendar "February 2026" RTL test and the Typeahead paste test. The Calendar failure reproduces identically on a pristine `main` checkout (no Calendar files touched here); Typeahead and MultiSelector pass in isolated reruns.
- `node_modules/.bin/tsc --project packages/core/tsconfig.json --noEmit` — clean.
- `prettier --check` and `eslint` on all changed files — clean (the `@astryx/no-hardcoded-i18n-string` / `i18n-key-format` rules apply).

## Notes

- The Vercel deployment check is expected to fail on fork PRs — known infra issue, not related to this change.


cixzhang (approved):

cixzhang (approved):

packages/core/src/Timestamp/Timestamp.tsx
```diff
@@ -388,8 +388,14 @@ export function Timestamp({
...
 
-  // Full absolute text for tooltip and aria-label
+  // Full absolute text for the tooltip (visible — keeps the compact timezone
+  // abbreviation) and for the AT-facing aria-label, which spells the timezone
+  // out in full: abbreviations like "PST" or "GMT+2" are unexpanded
+  // abbreviations to a screen-reader user (WCAG 3.1.4).
   const fullAbsoluteText = isValidDate ? formatInstant(date, 'full') : '';
+  const ariaLabelText = isValidDate
+    ? formatInstant(date, 'full', {timeZoneNameStyle: 'long'})
+    : '';
```
cixzhang, line 461: Could be empty

packages/core/src/Timestamp/Timestamp.tsx
```diff
@@ -446,9 +452,7 @@ export function Timestamp({
       <time
         ref={mergeRefs(ref, timeRef)}
         dateTime={isoString}
-        aria-label={
-          effectiveFormat === 'relative' ? fullAbsoluteText : undefined
-        }
+        aria-label={effectiveFormat === 'relative' ? ariaLabelText : undefined}
```
cixzhang: In the case ariaLabelText is empty, is it better to have no aria-label and allow the contents of `<time>` be read out instead?

cixzhang (changes requested):

packages/core/src/Timestamp/Timestamp.tsx
bhamodi: `ariaLabelText` is `''` only when the date is invalid, and the invalid-value path bails out (renders nothing) before the `<time>` element is reached — so the empty string can't currently make it into the DOM. That said, the guard is now explicit at the use-site rather than relying on the distant early-return; see the sibling thread.

packages/core/src/Timestamp/Timestamp.tsx
bhamodi: Agreed on the fallback behavior -- fixed. The attribute is now omitted entirely whenever the text is empty (`aria-label=""` is at the mercy of per-browser accname quirks), so AT falls back to reading the visible `<time>` content:

```tsx
aria-label={
  effectiveFormat === 'relative' && ariaLabelText !== ''
    ? ariaLabelText
    : undefined
}
```

Worth noting the empty case is currently unreachable (invalid dates bail out before rendering the `<time>` at all), so this is a hardening guard rather than a behavior change -- but it keeps the invariant local instead of depending on the early return staying where it is. Full core suite re-run green (5710/5710) after the rebase onto current main.

bhamodi:
@cixzhang Fixed and rebased -- `aria-label` is now omitted entirely whenever the label text is empty so AT reads the `<time>` contents instead (details in the inline threads). Full core suite green after the rebase: 5710/5710.

bhamodi:
pr-a11y flagged `Timestamp::Colors::color-contrast` as new-vs-baseline. It's not from this PR -- the diff here is aria-label/formatInstant only, zero pixel changes to the Colors story -- so it's a main-side regression that postdates the baseline seed in #4386 (likely the TextColorMap change in #4577). Added the key to `.github/a11y-baseline.json` per the workflow's documented escape hatch so the underlying issue can be fixed against main (the weekly scan will surface it); flagging for reviewer blessing per the baseline file's contract.

cixzhang (approved):

bhamodi:
Rebase note: main's new `relative_short` format (#4673) also keeps the full absolute date as its aria-label, so after this rebase it composes with this PR -- `relative_short` now gets the spelled-out long timezone name (WCAG 3.1.4) and the omit-when-empty guard, same as `relative`. Its "keeps the full absolute date as the accessible name" test asserts label presence rather than the exact timezone form, so it passes unchanged. Timestamp suite green locally after the merge: 118/118.

bhamodi:
One more rebase: #4582 (copyable multi-timezone hover card) landed right after my last push and replaced the read-only tooltip with the card surface, in the same region this PR touches. The merge keeps the card exactly as shipped; the two card tests that asserted the row text equals the aria-label now compare against an independently formatted short-timezone string instead, since with this PR the aria-label intentionally spells the timezone out in full (WCAG 3.1.4) while the visible row keeps the compact abbreviation. Timestamp suite green locally after the merge: 122/122.

PR #4364, opened 2026-07-26
fix(core): single tab stop per CheckboxList option

bhamodi:
## Summary

Fixes a11y scan finding #3: every interactive `CheckboxListItem` rendered two same-named tab stops -- the `<CheckboxInput>` and an invisible whole-row `<button>` (from `Item`'s invisible-button pattern, triggered by the row `onClick`). Only the checkbox reports checked state, so keyboard users hit a second "Option A" stop that silently duplicates the action and announces no state (WCAG 4.1.2 Name/Role/Value; APG checkbox pattern: one control per option).

The fix makes the checkbox the option's sole focusable control while the whole row stays an enlarged click/tap target that toggles it.

Updated after review (cixzhang): the first version added a bespoke `hasNestedKeyboardControl` boolean on `Item`/`ListItem`, which was too specific as a public API. It is now replaced with an `interactiveRef` prop that adopts the existing `useClickableContainer` pattern -- the same one `SelectableCard`/`ClickableCard` use.

## Changes

- **`Item`**: new `interactiveRef` prop. When set, the row uses `useClickableContainer` to become an enlarged click/tap target that delegates surface clicks to the referenced control and renders no invisible button/anchor, so the row adds no second tab stop (WCAG 4.1.2). It is mutually exclusive with `onClick`/`href`: those are ignored in delegation mode and a dev warning fires if both are passed. No behavior change for existing consumers (Selector, MultiSelector, CommandPalette, lists) -- the invisible-button/anchor paths are untouched when `interactiveRef` is absent.
- **`ListItem`**: pass-through of `interactiveRef` to `Item`.
- **`CheckboxListItem`**: passes a ref to its `<CheckboxInput>` and hands it to `ListItem` as `interactiveRef`, so the checkbox is the row's only tab stop and row-surface clicks delegate to it. A consumer `onClick` now rides on the checkbox itself, so it still fires for both direct and delegated (row-surface) clicks, once each.
- Tests: the "single tab stop per option" suite in `CheckboxList.test.tsx` still passes unchanged (no row button, Tab moves checkbox-to-checkbox, row-surface click toggles, Space toggles, row not focusable); `Item.test.tsx` now covers `interactiveRef` delegation (no invisible button, checkbox is the sole tab stop, row-surface click is delegated to the control, no double-fire when the control itself is clicked, and `interactiveRef` wins over `onClick`).
- Docs: `interactiveRef` prop entries in `Item.doc.mjs` / `ListItem.doc.mjs` (en + zh + dense); changeset updated to describe the new API.

## Test plan

- `vitest run packages/core/src/Item/Item.test.tsx` -- 45/45 passed
- `vitest run packages/core/src/CheckboxList packages/core/src/List` -- 104/104 passed
- Full core suite: `vitest run packages/core` -- 5592/5592 passed (231 files), no regressions from the `Item` primitive change
- `tsc --noEmit` (packages/core) -- clean
- `eslint` on changed source/test files -- clean (`.doc.mjs` are eslint-ignored by config)
- `prettier --check` on changed `.tsx`/`.md` -- clean (`.doc.mjs` excluded from prettier/lint-staged, left untouched)

## Notes

- The Vercel deployment check failing on fork PRs is known infra, not related to this change.
- **RadioList is not affected**: `RadioListItem` composes `Item` without `onClick` and uses a native `<label htmlFor>` for row clicks, so it never rendered a duplicate row button -- no follow-up needed there.

### DIFF
--- a/packages/core/src/CheckboxList/CheckboxListItem.tsx
+++ b/packages/core/src/CheckboxList/CheckboxListItem.tsx
@@ -16,12 +16,11 @@
  * - /packages/cli/templates/blocks/components/CheckboxList/ (showcase blocks)
  */
 
-import {use, type MouseEvent, type ReactNode} from 'react';
+import {use, useRef, type ReactNode} from 'react';
 import * as stylex from '@stylexjs/stylex';
 import type {StyleXStyles} from '@stylexjs/stylex';
 import {colorVars} from '../theme/tokens.stylex';
 import type {BaseProps} from '../BaseProps';
-import {composeEventHandlers} from '../utils';
 import {useDevWarning} from '../hooks/useDevWarning';
 import {CheckboxInput} from '../CheckboxInput/CheckboxInput';
 import {ListItem} from '../List/ListItem';
@@ -212,6 +211,14 @@ export function CheckboxListItem({
   // Whether this item is interactive (has a toggle handler)
   const isInteractive = !effectiveReadOnly && (ctx != null || onCheck != null);
 
+  // The checkbox is the row's single keyboard control and action. The row is
+  // an enlarged click/tap target that delegates surface clicks to it via
+  // ListItem's `interactiveRef` (useClickableContainer), so each option is
+  // exactly one tab stop. Delegate whenever the row should respond to clicks:
+  // a toggleable item, or one carrying a consumer `onClick`.
+  const checkboxRef = useRef<HTMLInputElement | null>(null);
+  const hasRowInteraction = isInteractive || onClickProp != null;
+
   const handleToggle = () => {
     if (effectiveDisabled || effectiveReadOnly || isBusy) {
       return;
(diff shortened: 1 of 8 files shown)


packages/core/src/CheckboxList/CheckboxListItem.tsx
```diff
@@ -244,6 +252,11 @@ export function CheckboxListItem({
       description={description}
       endContent={endContent}
       isDisabled={effectiveDisabled}
+      // The checkbox in startContent is the row's sole keyboard control:
+      // row clicks are pointer-only (no invisible row button), so each
+      // option is exactly one tab stop that also reports checked state
+      // (WCAG 4.1.2 / APG checkbox pattern).
+      hasNestedKeyboardControl
```
cixzhang: Hmm I have to think about this as a public API since it's getting to be too specific as a knob. Especially since we have href and it almost sounds like we would want to adjust behaviors for container interactions if they happen to be different. Maybe we can use the useClickableContainer pattern and pass through an interactiveRef then we don't want href or onClick to be passed to Item in that case and instead the item is treated as an increased click/tap target for the interactiveRef. Check useClickableContainer for the standard API pattern here.

cixzhang (changes requested):
Back for thoughts on the new API

packages/core/src/CheckboxList/CheckboxListItem.tsx
bhamodi: Good call, the boolean was too specific as a public knob. I've replaced `hasNestedKeyboardControl` with an `interactiveRef` prop on `Item`/`ListItem` that adopts the `useClickableContainer` pattern, same as `SelectableCard`/`ClickableCard`: pass a ref to the nested control and the row becomes an enlarged click/tap target that delegates surface clicks to it and renders no invisible button/anchor, so there's still exactly one tab stop per option. `interactiveRef` is mutually exclusive with `onClick`/`href` -- they're ignored in that mode and it dev-warns if you pass both. `CheckboxListItem` now hands a ref to its `CheckboxInput` and passes it through as `interactiveRef` instead of wiring `onClick`; a consumer `onClick` rides on the checkbox itself, so it still fires once for both direct and row-surface clicks. It's in the second commit (`fix(core): delegate CheckboxList row clicks via interactiveRef`) -- see `Item.tsx` for the delegation plus dev-warn, and `CheckboxListItem.tsx` for the ref wiring.

cixzhang (approved):

PR #4366, opened 2026-07-26
fix(core): AppShell skip-link focus target, localized label, banner landmark

bhamodi:
## Summary

Three AppShell accessibility fixes from the a11y scan (#3):

1. **Focusable skip-link target (WCAG 2.4.1 Bypass Blocks).** The "Skip to content" link pointed at `<LayoutContent role="main" id="astryx-app-shell-main">` — a plain div with no `tabIndex`. In several browsers, activating a fragment link whose target is not focusable leaves keyboard focus at the top of the page, so the skip link silently did nothing for the users it exists for. The main container now has `tabIndex={-1}`, and the skip link explicitly focuses it on activation (hash navigation alone doesn't move focus reliably in every browser). A scoped `outline: none` on `:focus` suppresses the focus ring on the container — it's only ever focused programmatically (never tabbable), so a ring around the entire content area would be noise. `reset.css` doesn't handle this case (it only suppresses `:focus-visible` on touch-only devices).
2. **Localized skip-link label (i18n).** "Skip to content" was the only hardcoded English string in AppShell — everything else goes through `t()`. Added `@astryx.appShell.skipToContent` to `packages/core/locales/en.json` and rendered via `useTranslator()`.
3. **Banner landmark.** The header wrapper (topNav + banner slot) was a plain div with no landmark role. It now gets `role="banner"`. Verified placement is top-level: the wrapper sits in Layout's header slot under plain divs only (AppShell root div → Layout outer/inner divs), never nested inside `main`, `nav`, or any other landmark, so the banner role is valid.

## Changes

- `packages/core/src/AppShell/AppShell.tsx` — `tabIndex={-1}` + focus-ring suppression on the main content container, skip-link `onClick` focus handler, `t('@astryx.appShell.skipToContent')` label, `role="banner"` on the header wrapper
- `packages/core/locales/en.json` — new `@astryx.appShell.skipToContent` entry (defaultMessage + description)
- `packages/core/src/AppShell/AppShell.test.tsx` — new tests: skip-link activation moves focus to main (asserts `document.activeElement` and `tabindex="-1"`), skip-link text resolves from the catalog (via `InternationalizationProvider` override), header exposes a top-level banner landmark, no banner landmark without header content
- `.changeset/a11y-appshell-skip-banner.md` — patch changeset

## Test plan

- New tests confirmed failing pre-fix: `vitest run packages/core/src/AppShell` → 3 failed | 39 passed (42)
- Post-fix scoped: `node_modules/.bin/vitest run packages/core/src/AppShell --reporter=dot` → 42 passed (42)
- Full core suite: `node_modules/.bin/vitest run packages/core --reporter=dot` → 222 files passed, 5162 tests passed, 0 failed (no flakes this run)
- `node_modules/.bin/tsc --project packages/core/tsconfig.json --noEmit` → clean
- `node_modules/.bin/eslint` on changed source/test files → clean (i18n hardcoded-string rule applies)
- `prettier --check` on changed files → clean
- `node scripts/check-changesets.mjs` → 45 changeset(s) valid

## Notes

- Vercel fork-PR deployment failure is known infra and unrelated to this change.

### DIFF
--- a/.changeset/a11y-appshell-skip-banner.md
+++ b/.changeset/a11y-appshell-skip-banner.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] AppShell: make the skip-link target focusable (tabIndex={-1}), localize the skip-link label via the i18n catalog, and expose the header region as a banner landmark
+@bhamodi
--- a/packages/core/locales/en.json
+++ b/packages/core/locales/en.json
@@ -235,6 +235,10 @@
     "defaultMessage": "Mobile navigation",
     "description": "Screen-reader-only accessible name for the mobile-only navigation region on small viewports. \"Mobile\" = phone/tablet (small screen), not \"movable\"."
   },
+  "@astryx.appShell.skipToContent": {
+    "defaultMessage": "Skip to content",
+    "description": "Text of the skip link — the first focusable element on the page, visible only while keyboard-focused. Activating it jumps focus past the navigation to the main content area. Imperative verb; keep short."
+  },
   "@astryx.avatar.nameWithStatus": {
     "defaultMessage": "{name}, {status}",
     "description": "Screen-reader accessible name for an Avatar showing a status indicator; composes the person's name with the status label, e.g. \"Jane Doe, Online\". {name} = the avatar's name/alt text, {status} = the status dot's label. Adjust separator and order per locale."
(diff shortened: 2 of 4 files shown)


cixzhang (approved):

bhamodi:
pr-a11y flagged `AppShell::Playground::color-contrast` (3 nodes) as new-vs-baseline. Not from this PR: the diff adds role="banner", tabIndex=-1, a focus-outline suppression, and a localized label -- nothing that changes rendered colors -- so it's a main-side regression postdating the #4386 baseline seed. Added the key to `.github/a11y-baseline.json` per the documented escape hatch; flagging for reviewer blessing, and the underlying contrast issue should be fixed against main.

bhamodi:
Rebased onto current main. The `AppShell::Playground::color-contrast` escape-hatch key is now in main's baseline via #4603, so this PR no longer touches `.github/a11y-baseline.json` -- the diff is back to the AppShell fix only.

PR #4368, opened 2026-07-26
fix(core): disabled TopNav items no longer navigate; Outline aria-current

bhamodi:
## Summary

A11y scan #3, WCAG 4.1.2 (Name, Role, Value). Disabled `TopNavItem`s set `aria-disabled` and `tabIndex={-1}` but still passed `href`/`target` through to the underlying `LinkComponent`, so they:

- still navigated on click (pointer-events only blocked real pointers, not programmatic/AT activation or synthetic clicks), and
- were announced by screen readers as live links with a destination, contradicting the disabled semantics.

This applies the same mechanism the design system already uses in `Link`'s disabled path and `SideNavItem`'s `NavItemElement`: when disabled, render a plain href-less `<a>` (skipping the router `LinkComponent`, since a disabled item performs no navigation and custom router links may require a live href), omit `target` with the href, suppress the consumer `onClick`, and keep `preventDefault` as a defensive guard against synthetic clicks. An href-less anchor exposes no link role and is not focusable, so programmatic focus + Enter, AT activation commands, and middle-click cannot navigate.

Also from the same nav-family audit batch: `Outline`'s active TOC item used `aria-current="true"`; the correct token for a position within a table of contents is `aria-current="location"`.

## Changes

- `packages/core/src/TopNav/TopNavItem.tsx` — destructure `href`/`target`/`onClick`; when `isDisabled`, render a plain `<a>` with no `href`/`target` and a `preventDefault` click guard, in **both** default and drawer modes. Visual styling is unchanged (same stylex style sets). Drawer mode's click handler now also calls `preventDefault` when disabled.
- `packages/core/src/TopNav/TopNavItem.doc.mjs` — `isDisabled` prop descriptions (EN/ZH/dense) note that href/target are dropped.
- `packages/core/src/Outline/Outline.tsx` — `aria-current={isActive ? 'location' : undefined}`; the tab-stop seating effect now matches on `'location'` too.
- `packages/core/src/TopNav/TopNav.test.tsx` — new tests: disabled items (both modes) have no `href`/`target`, expose no link role, keep `aria-disabled`/`tabIndex=-1`, and clicks are cancelled without firing the consumer `onClick`; enabled items (both modes) keep `href` and click behavior. Two existing disabled-item tests updated to not query by link role.
- `packages/core/src/Outline/Outline.test.tsx` — 12 `aria-current` assertions updated from `'true'` to `'location'`.
- `.changeset/a11y-topnav-disabled-href.md` — patch changeset.

## Test plan

- Pre-fix failure confirmed: with the new/updated tests on the unfixed code, `vitest run packages/core/src/TopNav packages/core/src/Outline` → **14 failed | 131 passed** (2 new TopNav disabled tests + 12 Outline aria-current assertions).
- Post-fix scoped: `node_modules/.bin/vitest run packages/core/src/TopNav packages/core/src/Outline --reporter=dot` → **145 passed (145)**, 5 files.
- Full core suite: `node_modules/.bin/vitest run packages/core --reporter=dot` → **5161 passed | 1 failed (5162)**; the single failure is the known Calendar flake (`Calendar.test.tsx:884`, 'February 2026') — passes on isolated rerun in this branch (63/63) and behaves identically on clean main, so not a regression.
- `node_modules/.bin/tsc --project packages/core/tsconfig.json --noEmit` → clean.
- `prettier --check` on changed `.ts/.tsx/.md` files → clean (`.doc.mjs` is not prettier-formatted in this repo and was already non-clean on main; only targeted description lines were edited).
- `eslint` on changed files → 0 errors.
- `node scripts/check-changesets.mjs` → valid.

## Notes

### DIFF
--- a/packages/core/src/TopNav/TopNavItem.tsx
+++ b/packages/core/src/TopNav/TopNavItem.tsx
@@ -131,6 +131,8 @@ export interface TopNavItemProps extends BaseProps<HTMLAnchorElement> {
   isSelected?: boolean;
   /**
    * Whether the nav item is disabled.
+   * A disabled item renders as a plain anchor without an href (and without
+   * target), so it cannot navigate and is removed from the tab order.
    * @default false
    */
   isDisabled?: boolean;
@@ -158,6 +160,15 @@ export interface TopNavItemProps extends BaseProps<HTMLAnchorElement> {
 }
 
 /**
+ * Click handler for disabled items. The disabled anchor renders without an
+ * href, so there is no navigation to block in practice; preventDefault is a
+ * defensive guard against synthetic/programmatic clicks.
+ */
+function preventDefaultClick(event: React.MouseEvent<HTMLAnchorElement>): void {
+  event.preventDefault();
+}
+
+/**
  * A navigation item for use within TopNav startContent.
  *
  * Renders as an anchor element with hover/selected states.
@@ -178,6 +189,9 @@ export interface TopNavItemProps extends BaseProps<HTMLAnchorElement> {
  */
 export function TopNavItem({
   as,
+  href,
+  target,
+  onClick,
   label,
   isSelected = false,
   isDisabled = false,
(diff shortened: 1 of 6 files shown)


ernestt (approved):

packages/core/src/TopNav/TopNavItem.tsx
```diff
@@ -157,6 +159,15 @@ export interface TopNavItemProps extends BaseProps<HTMLAnchorElement> {
   size?: NavItemSize;
 }
 
+/**
+ * Click handler for disabled items. The disabled anchor renders without an
+ * href, so there is no navigation to block in practice; preventDefault is a
+ * defensive guard against synthetic/programmatic clicks.
+ */
+function preventDefaultClick(event: React.MouseEvent<HTMLAnchorElement>): void {
```
cixzhang, line 167: Is it really necessary to break out a function for this?

cixzhang (approved):

PR #4371, opened 2026-07-26
feat(date-time-input): add theme targets for the date and time segments

### LINKED ISSUE #4075: [gap] DateTimeInput: No theming targets for its internal date/time segments. The component
| Field | Value |
|-------|-------|
| **Component** | DateTimeInput |
| **Category** | Styling gap |
| **Source** | cli |
| **Timestamp** | 2026-07-19T08:28:43.096Z |

## User Intention

No theming targets for its internal date/time segments. The component publishes only its root (astryx-date-time-input); the two segment wrappers are anonymous nodes with hashed atomic classes only. Themes that restyle input geometry (padding/height/font via the text-input/date-input/time-input targets) cannot reach the segments, so DateTimeInput renders visibly shorter than every other input under such themes. Slider (slider-track, slider-thumb) and RadioList (radio, radio-dot, radio-list-item) expose part targets; DateTimeInput should expose segment targets too — or compose real DateInput/TimeInput internally so their existing targets apply.

AKnassa:
## What this does

Lets a theme style the date box and the time box inside a `DateTimeInput`. Until now a theme could only style the whole field.

## Why

A `DateTimeInput` is two small boxes side by side — one for the date, one for the time. Themes can resize every other input in the design system (height, padding, font) through the `text-input` / `date-input` / `time-input` targets, but those two inner boxes had no name a theme could point at. So under a custom theme, `DateTimeInput` ended up visibly shorter than every other field on the same form.

## What changed

- The date box and the time box each got a name a theme can target — `astryx-date-time-input-date-segment` and `astryx-date-time-input-time-segment`
- Both report their size and any error/success state as data attributes, exactly like the field around them already does, so no new vocabulary is introduced
- Nothing looks different out of the box — the class is additive and the existing styles are untouched
- Added docs (English + Chinese), tests for the rendered classes and the state reflection, a `defineTheme` override test, and a Storybook example

Follows the pattern set by the `astryx-tree-list-chevron` and `astryx-tree-list-item-label` targets (#4305, #4306).

## One thing this deliberately does not do

The issue offered a second option — rebuild `DateTimeInput` out of real `DateInput` and `TimeInput` components so their existing theme targets apply. That rewrites how the component is built and how it behaves, so it is left alone; this is the theming fix.

## How to see it

Storybook → DateTimeInput → **Themed Segments**. The date box takes an accent border and the time box a soft grey fill, both purely from a theme — which was impossible before this.

00-before-after
01-themed-segments-light
03-default-unthemed-light

Closes #4075

### DIFF
--- a/.changeset/date-time-input-segment-theme-targets.md
+++ b/.changeset/date-time-input-segment-theme-targets.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] DateTimeInput: add `astryx-date-time-input-date-segment` and `astryx-date-time-input-time-segment` theme targets on the two segment wrappers, so a theme can restyle their geometry (padding/height/font) via `defineTheme` instead of being unable to reach them at all. Both reflect `size` and `status` as data attributes, mirroring the root target. Default rendering is unchanged.
+
+@AKnassa
--- a/apps/storybook/stories/DateTimeInput.stories.tsx
+++ b/apps/storybook/stories/DateTimeInput.stories.tsx
@@ -4,6 +4,7 @@ import {useState} from 'react';
 import type {Meta, StoryObj} from '@storybook/react';
 import {DateTimeInput} from '@astryxdesign/core/DateTimeInput';
 import type {ISODateTimeString} from '@astryxdesign/core/DateTimeInput';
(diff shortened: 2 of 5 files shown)


cixzhang (approved):
Thanks for adding these targets!

PR #4372, opened 2026-07-26
feat(cli): expose findRelatedBlocks on the public API

### LINKED ISSUE #1901: feat(cli): expose findRelatedBlocks through the public API
## Problem

When building the docsite component detail page, we needed to fetch example blocks (non-showcase) for each component. The CLI has `findRelatedBlocks(componentName)` in `template.mjs` which does exactly this, but it is not re-exported from `@xds/cli/api`.

## Friction encountered

1. **`findRelatedBlocks` exists but is not accessible** — it is exported from `template.mjs` but the API index (`api/index.mjs`) only re-exports `template()`. Direct subpath imports fail because the `exports` field in `package.json` restricts access to `./api` and `./json` only.

2. **`template()` list does not include enough metadata** — the list endpoint returns `name`, `displayName`, `description`, `isReady`, `type` but not `category`, `isShowcase`, `componentsUsed`, or any way to associate blocks with components.

3. **No `component(name, { blocks: true })` option** — there is `component(name, { showcase: true })` which internally calls `findShowcase`, but no equivalent for fetching related example blocks.

4. **Workaround required filesystem reads** — we ended up reading directly from `node_modules/@xds/cli/templates/blocks/components/<DirName>/` and regex-parsing the `.doc.mjs` files for metadata. This works but is fragile and bypasses the CLI abstraction.

## Proposed solution

Add one or both of:

### Option A: Export `findRelatedBlocks` from the API
```js
// api/index.mjs
export { findRelatedBlocks } from './template.mjs';
```

### Option B: Add a `blocks` option to the `component()` API
```js
const result = await component('Button', { blocks: true });
// { type: 'component.detail.blocks', data: [{ id, name, description, source, isShowcase }] }
```

Option B is cleaner since it follows the existing pattern (`showcase: true` → showcase data), and could filter out showcases by default (with an opt-in to include them).

## Additional note

The `template()` list endpoint would also benefit from including `category` and `componentsUsed` fields, which would allow client-side filtering without additional API calls.

AKnassa:
## What this does

Makes it possible to ask the Astryx CLI "which example blocks use this component?" from code, instead of digging through files inside `node_modules`.

## Why

When the docs site component page was built it needed each component's example blocks. The CLI already knew the answer — `findRelatedBlocks()` lives in `template.mjs` — but it was never re-exported from `api/index.mjs`, and the package `exports` field only opens `./api` and `./json`. So the docs site ended up reading files out of `node_modules/@astryxdesign/cli/templates/...` by hand and regex-parsing the `.doc.mjs` files. Fragile, and it bypasses the CLI entirely.

## What changed

- `findRelatedBlocks(componentName, cwd?)` is now available from `@astryxdesign/cli/api`
- Block listings (`template.list`) now report `isShowcase`, so you can tell a component's hero example from its other blocks without a second call. Page templates have no showcase concept, so it stays absent for them
- Both are fully typed: `DiscoveredTemplate` added to the public types, and the new export pinned in `api.contract.assert.ts` so the hand-written declarations cannot drift from the runtime function

It is exported as a plain helper rather than a `{ type, data }` envelope, matching `summarizeIssues`. The `api/index.mjs` header claimed every function returns the envelope, which was already untrue — corrected here.

## Worth knowing

Most of this issue is already solved. The `component(name, { blocks: true })` route — the option the issue itself called cleaner — shipped with the full-API-coverage work in #4302, and `template.list` already carried `category` and `componentsUsed`. This closes the two gaps that survived that work.

## How to see it

```js
import {findRelatedBlocks} from '@astryxdesign/cli/api';

const blocks = await findRelatedBlocks('AlertDialog');
// -> the blocks whose componentsUsed names AlertDialog, with filePath/docPath
```

`pnpm typecheck:json-api` and `node .github/scripts/api-cli-parity-test.mjs` (330 cases) both pass.

Closes #1901

### DIFF
--- a/packages/cli/api/index.mjs
+++ b/packages/cli/api/index.mjs
@@ -3,7 +3,9 @@
 /**
  * @file Programmatic API for the Astryx CLI.
  *
- * Every function returns the same { type, data } envelope that `xds --json` outputs.
+ * Every command-shaped function returns the same { type, data } envelope that
+ * `xds --json` outputs. A few helpers (`summarizeIssues`, `findRelatedBlocks`)
+ * are plain functions over that data and return their result directly.
  * Errors throw AstryxError (with optional .suggestions).
  *
  * @example
@@ -23,7 +25,7 @@ export {component} from './component/component.mjs';
 export {docs} from './docs/docs.mjs';
 export {blog} from './blog/blog.mjs';
 export {discover} from './discover/discover.mjs';
-export {template} from './template/template.mjs';
+export {template, findRelatedBlocks} from './template/template.mjs';
 export {themeAdd, listThemes} from './theme/theme-add.mjs';
 export {hook} from './hook/hook.mjs';
 export {search} from './search/search.mjs';
--- a/.changeset/cli-find-related-blocks-api.md
+++ b/.changeset/cli-find-related-blocks-api.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/cli': patch
+---
+
+[feat] CLI API: re-export `findRelatedBlocks(componentName, cwd?)` from `@astryxdesign/cli/api`, and report `isShowcase` on `template.list` block entries. Consumers that need the blocks composing a component (e.g. a docsite component page) no longer have to read `node_modules/@astryxdesign/cli/templates/...` by hand — the package `exports` field only opens `./api` and `./json`, so the function was unreachable. `component(name, {blocks: true})` remains the command-shaped view; this is the raw-record helper, alongside `summarizeIssues`.
+
+@AKnassa
--- a/packages/cli/api/template/template.mjs
+++ b/packages/cli/api/template/template.mjs
@@ -188,7 +188,6 @@ async function loadDocModule(docPath) {
   return docModule.default ?? docModule.doc;
 }
 
-
 export {discoverAll as discoverTemplates};
 
 export function listTemplates() {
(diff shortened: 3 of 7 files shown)


packages/cli/api/index.mjs
```diff
@@ -23,7 +25,7 @@ export {component} from './component/component.mjs';
 export {docs} from './docs/docs.mjs';
 export {blog} from './blog/blog.mjs';
 export {discover} from './discover/discover.mjs';
-export {template} from './template/template.mjs';
+export {template, findRelatedBlocks} from './template/template.mjs';
```
cixzhang, line 28: This is the permanent bit — an export on a published package with no caller in the diff.

cixzhang (changes requested):
Thanks — "the docsite reads `node_modules` by hand" is a real problem and worth closing.

The hold is on the export itself. `component(name, {blocks: true})` already returns exactly these records with the hero flagged: `api/component/detail/blocks/blocks.mjs:22` calls `findRelatedBlocks` and puts `isShowcase` on every entry. Nothing in this PR reaches for the raw version, so the page that motivates it isn't here to show what the envelope is missing — and once `@astryxdesign/cli` ships the symbol, taking it back is a breaking release.

So: what does that page need that the envelope doesn't give it — `componentsUsed`, the `cwd` argument, something else? If it's `cwd`, that reads to me like an argument on the call that already exists rather than a second entry point.

The `isShowcase` field on `template.list` is fine either way.

[Reviewed by Robohands]

cixzhang:
Checking in on the CLI export: the current review needs the concrete consumer gap that `component(name, {blocks: true})` cannot serve. Would you like to continue, have us carry it, or close this for now?

PR #4373, opened 2026-07-26
fix(core): add StatusDot icon prop; keep the dot a plain colored signal

bhamodi:
## Summary

Follow-up to the design review (see [ernestt's update](https://github.com/facebook/astryx/pull/4373#issuecomment-5196362808)): at 8px, differentiating status by a built-in glyph satisfies WCAG 2.1 SC 1.4.1 on paper, but the marks are too small to genuinely read — compliance without the usability the criterion protects. So StatusDot no longer tries to solve 1.4.1 on its own by default. The dot returns to a **plain colored signal**, and making the status accessible in context is framed as the builder's responsibility — the usage dos/don'ts for that landed separately in #4737 (merged).

What this PR now ships:

- **`icon?: ReactNode` prop** (the part the review kept) — API parity with `AvatarStatusDot`, identical contract: rendered centered in the 8px field, painted from the dot's `currentColor` ink; booleans and empty renders are ignored so `cond && <Icon />` stays safe. This is the "use an icon to carry the meaning" escape hatch from the docs guidance.
- **Dedicated per-variant inks** (`--color-on-*`, the Badge precedent) so a passed icon always lands on a designed contrast pair — notably the fixed dark `--color-on-warning` ink (~9.6:1 on the yellow plate, where a light surface ink lands near 2:1; caught by @cixzhang earlier in review).

Removed relative to the previous revision of this branch: the per-variant built-in glyphs, the `statusdot:<variant>` icon-registry resolution, the `astryx-statusdot-glyph` theme target, and the neutral ring inversion (neutral is a plain `--color-icon-secondary` dot again). The registry seam existed to theme the built-in glyphs, so it goes with them.

## Changes

- `packages/core/src/StatusDot/StatusDot.tsx` — remove glyph map/SVG/registry resolution; keep `icon` prop + `isRenderable` gating and per-variant `currentColor` inks; component JSDoc now states the dot is colour-only by default and points at the usage guidance
- `packages/core/src/StatusDot/StatusDot.test.tsx` — new guard that every variant renders a plain childless dot by default; icon contract tests (render, aria-hidden wrapper, boolean/empty safety, accessible name); ink-pairing regression test for the warning plate; glyph + registry suites removed
- `packages/core/src/StatusDot/StatusDot.doc.mjs` — `icon` prop rows (en/zh/dense) on top of the #4737 guidance; glyph wording reverted
- `apps/storybook/stories/StatusDot.stories.tsx` — `GlyphLegibility` replaced with a lean `WithIcon` story
- `.changeset/statusdot-icon-prop.md` — rewritten patch changeset (renamed from `a11y-statusdot-shapes.md`)
- `packages/core/src/StatusDot/index.ts` and the CLI showcase block are reverted to main — no longer part of this PR

## Test plan

- `vitest run packages/core/src/StatusDot/StatusDot.test.tsx` — 21 passed
- Full `packages/core` suite — 6869/6870 passed; the one failure is the known flaky `Markdown/parser.perf.test.ts` streaming-XL timing test (fails on clean main too)
- `tsc --noEmit` on `packages/core` — clean
- `prettier --check` on changed files — clean
- eslint could not run locally on this branch (main's `eslint.config.js` now needs `@eslint/compat`, not present in this machine's offline store) — relying on the CI lint job
- Storybook `Core/StatusDot → WithIcon` shows the icon escape hatch on the deployed preview

## Notes

- Rebased onto current `main` (00bed2975); 0 behind at push time.
- Vercel deployment failure on fork PRs is known infra and unrelated.


ernestt:
Two things I'd like us to align on before this lands — both push toward treating StatusDot as the *general-purpose semantic* status primitive rather than a mirror of AvatarStatusDot.

**1. Expose a custom `icon` prop for API parity.**

AvatarStatusDot already accepts an `icon` that overrides the built-in glyph. StatusDot is the more general-purpose of the two, so it's odd for it to have the smaller surface. Can we add the same `icon?: ReactNode` here, with the identical contract — a rendered icon suppresses the built-in glyph, and booleans/empty renders are ignored so `cond && <Icon />` stays safe? That keeps the two components' escape hatches consistent and lets consumers use their own status marks.

**2. Map the default glyphs to semantic meaning, not to Avatar's presence shapes.**

Right now the PR reuses AvatarStatusDot's shape language (and pins it with a cross-component parity test). I'd argue that's the wrong axis of consistency. The two components live in different meaning-domains:

- **AvatarStatusDot = presence** — filled/ring/minus map to online/away/DND (the Slack/Teams/Discord convention).
- **StatusDot = semantic status** — it belongs with `Banner` and `FieldStatus`, which already map status → icon *by name* (`success`→✓, `error`→✕, `warning`→△-with-!) via the shared `defaultIcons` vocabulary.

The consistency principle that matters is **same meaning → same shape**, so StatusDot should reference the system's own semantic icon set rather than presence shapes. Proposed mapping:

| Variant | Glyph | Rationale |
| --- | --- | --- |
| `success` | ✓ check | matches `defaultIcons.success` / Banner / FieldStatus |
| `warning` | ! (exclamation) | the mark inside the canonical warning triangle; far clearer than a lone vertical bar |
| `error` | ✕ cross | matches `defaultIcons.error`; clearer than a minus |
| `neutral` | ring | unchanged |
| `accent` | filled (no glyph) | accent is the emphasis/info color (Banner maps `info`→`accent`); leaving it the plain reference dot |

This does mean dropping the "reuse Avatar's shapes" framing and the cross-component parity test — only `neutral` stays shared. That's intentional: differentiating by meaning is the stronger design.

**One caveat to verify:** ✓ / ✕ / ! at an 8px field with a 1px stroke is tight geometry — ✓ vs ✕ in particular risk blurring into the same mark, and those are the two we most need to keep distinct. Could we get a Storybook screenshot at 1× (and a colorblind sim) confirming they're distinguishable from each other and from the ring before this goes ready? If they don't resolve at 8px we may need a slightly heavier stroke for those two, or to reconsider the minimum size.

cixzhang:
image

Here's how it looks.

bhamodi:
@ernestt Agreed on both counts, and thanks for the sharp framing. You're right that the axis of consistency for StatusDot is semantic meaning, not Avatar's presence shapes, so I remapped the defaults to the defaultIcons vocabulary: success -> check, warning -> exclamation (a stem over a dot, the mark inside the warning triangle), error -> cross, neutral -> ring (unchanged), accent -> plain filled dot. I also added `icon?: ReactNode` with the exact same contract as AvatarStatusDot: a rendered icon replaces the built-in glyph, and booleans/empty renders are ignored so `cond && <Icon />` stays safe. On your 8px caveat, check and cross are the pair most at risk of blurring together, so those two diagonals now take a slightly heavier 1.25px stroke while the ring and exclamation stem stay at 1px; there's a new GlyphLegibility story that renders every variant at native 8px (for a colorblind sim) plus a magnified row so you can eyeball the geometry on the deployed Storybook. I dropped the cross-component parity test as you suggested, so only neutral (the ring) is shared now, intentionally.

bhamodi:
@cixzhang Heads up, the glyphs changed since this screenshot: the defaults are now the semantic icon set (success check, warning exclamation, error cross, neutral ring, accent plain filled dot), and the diagonals got a slightly heavier stroke for 8px legibility. The new Core/StatusDot -> GlyphLegibility story has all of them at 1x plus a magnified row if you want to grab a fresh capture.

cixzhang:
Screenshot 2026-07-30 at 3 03 07 PM

It's more consistent but we may have a contrast issue with the yelllow.

humbertovirtudes (commented):
This is genuinely good work, and the reasoning is unusually careful. Encoding status as shape + colour (WCAG 1.4.1) by borrowing the system's own semantic icon vocabulary (the `defaultIcons` marks `Banner`/`FieldStatus` render) rather than inventing new glyphs is the right convergence choice, and keeping StatusDot on the semantic axis while AvatarStatusDot stays on the presence axis (sharing only the neutral ring) is a well-articulated distinction. Painting the glyph from `currentColor` so plate and ink can never drift out of contrast is a nice invariant.

Triage: core component behavior change + new public prop, additive-but-alters-default-render, medium blast radius (StatusDot is a shipped primitive) -> deep path.

- `vitest run packages/core/src/StatusDot/StatusDot.test.tsx`: 36/36 pass; eslint on `StatusDot.tsx` + the story: exit 0. CI is green.
- Parity claim holds exactly: `AvatarStatusDot` already ships `icon?: ReactNode` and uses the identical `isRenderable(icon)` -> `showsIcon` -> `glyphShapeMap[variant]` pattern, so this mirrors an established sibling API rather than inventing one.
- Shape-source claims are accurate: `defaultIcons.check` really is `M5 13l4 4L19 7`, and `close`/`warning` exist as claimed; the 8px paths are those marks scaled into the dot's field.
- The glyph is correctly hidden from AT (`aria-hidden="true"`); the status still reaches screen readers through the existing `label`/`aria-label`, so the shape is pure visual redundancy.
- 35 commits behind `main`, but nothing on `main` has touched `packages/core/src/StatusDot/` since the branch point, so it stays clean.

One thing to flag for a maintainer, not a request-changes from me: this crosses the new-public-API line twice, and both are calls I think a human owner should bless rather than land on "additive + green CI":
1. New public prop `icon?: ReactNode` on a shipped component. The contract is sound and mirrors `AvatarStatusDot` verbatim, so if that precedent is the intended convergence, this is an easy yes.
2. It changes the default rendered output of every non-accent StatusDot (a glyph now paints inside the dot, and `neutral` inverts to a surface plate with a coloured stroke). That is the point of the fix, but it is a visual change to existing consumers, so it is worth an explicit owner sign-off on the look at 8px. The `GlyphLegibility` story (1x for colour-blind sim + magnified for geometry) is exactly the artifact to make that call on the deployed Storybook.

No spec issue is linked; for a change to a shipped primitive's default appearance a one-line linked issue/RFC would make the surface decision traceable. Everything mechanical is clean (tests, lint, changeset present/scrubbed/patch, docs + zh docs + dense docs + story + showcase all updated in lockstep), so once a maintainer is happy with the API-parity precedent and the 8px look, this should land quickly. Holding my own approval only because of the new public surface, not because of any defect.

bhamodi:
@cixzhang Fixed -- the yellow was a real failure: the glyph ink was `--color-background-surface` (near-white in light mode), which lands around 2:1 on the `--color-warning` plate (#E9AF08). The filled variants now paint their ink from the dedicated semantic pairings instead (same pattern Badge already uses): success -> `--color-on-warning`'s siblings `--color-on-success`, warning -> `--color-on-warning` (the fixed dark ink #0A1317, ~9.6:1 on the yellow), error -> `--color-on-error`, accent -> `--color-on-accent`. The neutral ring keeps `--color-icon-secondary`. Since the glyph still paints from `currentColor`, plate and ink stay a designed pair in both modes. The GlyphLegibility story shows all variants at native 8px + magnified if you want a fresh capture -- the warning exclamation now reads as a dark mark on yellow instead of white-on-yellow. 36/36 StatusDot tests green, rebased onto current main.

ernestt (approved):
This looks good! Thanks for addressing everything!

packages/core/src/StatusDot/StatusDot.tsx
```diff
@@ -104,6 +156,132 @@ export interface StatusDotVariantMap {
...
+ * Stroking (rather than a CSS box cutout) buys the sub-pixel control and
+ * round caps the small marks need to stay intentional at 8px.
+ *
+ * `viewBox` is one user unit per px of the dot, so every coordinate is
+ * literal px. Geometry mirrors the `defaultIcons` marks scaled into the 8px
+ * field. The diagonal `check` and `cross` take the heavier stroke; see
+ * `GLYPH_STROKE_DIAGONAL`. Themes can target the stable
+ * `astryx-statusdot-glyph` class and its `data-shape` attribute.
+ */
+function StatusDotGlyph({shape}: {shape: StatusDotGlyphShape}) {
```
cixzhang: Hmm should these be pulling from the icon registry instead? That way they can be themed. Although since they're a lot smaller I do wonder how much we can reuse the standard icons.

cixzhang (changes requested):

packages/core/src/StatusDot/StatusDot.tsx
bhamodi: Implemented -- the built-in glyphs now resolve through the icon registry, under scoped `statusdot:<variant>` keys (the same `getExtendedIcon` seam `RichTextEditorToolbar` uses for its `richtext:*` keys). Resolution order: explicit `icon` prop > theme/global registry override > the 8px-tuned built-in glyph. So `defineTheme({icons: {'statusdot:success': <MyGlyph />}})` reshapes every success dot, no code change in consumers.

On reusing the standard icons directly: I kept the keys separate from `check`/`warning`/`error` deliberately, for exactly the size reason you flagged -- those marks are drawn for a 24px field, and their stroke geometry doesn't survive being squeezed into the 8px dot (the built-ins here are those same marks redrawn at dot scale, with the heavier 1.25px diagonal stroke). Separate keys mean a theme that redraws its full-size semantic icons doesn't silently end up with illegible dots; it opts into the dot with dot-scaled artwork.

A bonus this unlocks: augmented custom variants can now get a real themable mark via `statusdot:<variant>` (e.g. `statusdot:critical`) instead of the documented `::before` workaround -- that guidance is updated too.

5 new tests cover override rendering, per-variant scoping, no-leak from standard-icon overrides, `icon`-prop precedence, and custom-variant keys (41/41 green). Also rebased onto current main (the variant-map interface move from #4646 resolved to `index.ts`).

bhamodi:
@cixzhang Addressed -- the glyphs now resolve through the icon registry under scoped `statusdot:<variant>` keys with the 8px-tuned built-ins as fallback (reasoning on the standard-icon size question in the inline thread). Also rebased onto current main. Can't re-request review from a fork, so flagging here.

ernestt:
Update from the design review — this changes the direction here, so thank you for bearing with the back-and-forth.

We took the shape-per-status question to a longer design discussion, and the conclusion is that baking distinct glyphs into the dot by default isn't the right call. The short version: at 8px, differentiating by shape *technically* satisfies 1.4.1, but in practice the marks are too small to actually read — so they don't deliver the usability the criterion is meant to protect. A ✓/✕/! that a low-vision or colorblind user genuinely can't distinguish at dot scale is compliance on paper, not in reality.

So rather than have StatusDot try to solve 1.4.1 on its own by default, we want to be honest that it's a primitive that can't be fully accessible in isolation — it's a signal the **builder** has to make accessible in context. Concretely:

- Keep the default rendering as the plain colored dot (no per-variant glyph by default).
- Add explicit usage guidance to the docs, framed as builder responsibility:
  1. Use StatusDot as a **binary** signal (present/absent) rather than encoding many distinct states in one dot.
  2. Pair it with a **label**.
  3. Or use an **icon** to carry the meaning.
  4. If you can't add a label or an icon, make sure the status is **conveyed elsewhere** accessibly.

The `icon?: ReactNode` prop you added is still very much wanted — it's exactly the "use an icon instead" escape hatch, and the parity with AvatarStatusDot is right. That part stays.

Where that leaves the PR: keep the `icon` prop (and the registry seam if it stays clean), drop the default per-variant glyphs, and land the usage-guidance dos/donts instead. Happy to talk through the exact split — and genuinely, the analysis in here was excellent; the outcome is a philosophy call on the component's contract, not a knock on the work.

ernestt:
Follow-up: opened #4737 to land the docs guidance described above — it adds the binary-signal, icon, and accessible-elsewhere dos/don'ts to StatusDot's usage docs (prop-agnostic, so it's correct against main regardless of how this PR lands). The default-glyph removal and the fate of the `icon` prop stay with this PR.

cixzhang (changes requested):
Back to you for the feedback from @ernestt

bhamodi:
@ernestt @cixzhang Done -- the dot is a plain colored signal again. Dropped the per-variant built-in glyphs, and with them the `statusdot:<variant>` registry keys, the `astryx-statusdot-glyph` theme target, and the neutral ring inversion (neutral is back to a plain `--color-icon-secondary` dot). The registry seam only existed to theme the built-in glyphs, so it went with them.

What stays, per the review: the `icon` prop with the exact AvatarStatusDot contract (rendered centered in the 8px field, painted from `currentColor`, booleans/empty renders ignored so `cond && <Icon />` is safe), plus the per-variant `--color-on-*` inks so a passed icon always lands on a designed contrast pair -- including the dark on-warning ink from the yellow-contrast round. A new test guards that every variant renders a plain childless dot by default, and the GlyphLegibility story became a lean WithIcon story. Docs-wise this branch now just adds the `icon` prop rows on top of the #4737 guidance (thanks for landing that separately -- it reads correctly against this PR as-is).

Also rebased onto current main (00bed2975, 0 behind at push) and retitled the PR to match the reduced scope. Verified locally: 21/21 scoped StatusDot tests, full core suite green except the known Markdown perf flake, tsc clean; eslint rides on CI since main's config now needs `@eslint/compat`, which my offline store doesn't have. Can't re-request review from a fork, so flagging here.

cixzhang (approved):

PR #4375, opened 2026-07-26
fix(core): focus trap DOM-depth Escape, aria-hidden exclusion, live-region clearing

bhamodi:
## Summary

Three hardening fixes to the shared a11y hooks that back every overlay in the system (Dialog, Popover, menus, MobileNav), from a11y scan #3, plus a doc-only WCAG note on `useHotkeys`:

- **useFocusTrap — Escape resolved by push order, not DOM depth.** React runs child effects before parent effects, so when an outer trap (e.g. a Dialog) and a trap nested inside it (e.g. a Popover) mount in the same commit, the inner trap pushes onto the Escape stack first and the outer trap wrongly won the pure last-pushed comparison — Escape closed the outer layer while the inner one stayed open.
- **useFocusTrap — trap membership included `aria-hidden` subtrees (WCAG 4.1.2).** A focusable element inside `[aria-hidden="true"]` was a Tab target in the trap cycle while assistive tech skips it, so keyboard focus could land on content that is invisible to AT.
- **useAnnounce — live regions were never cleared.** The polite/assertive regions retained the last message indefinitely, leaving stale status text in the accessibility tree for users browsing the DOM later.

## Changes

- `useFocusTrap.ts`: the Escape stack now stores `{handler, getContainer}` entries. `isTopEscapeHandler` walks the stack in push order and keeps the DOM-deepest container as the top (if entry A's container contains entry B's, B is topmost), with push order as the tiebreaker for non-nested traps. Containers are read lazily at keydown time. `hasActiveFocusTrapEscape()` semantics are unchanged (Dialog still uses it for Escape deferral).
- `useFocusTrap.ts`: `isVisiblyFocusable` now rejects elements matching or inside `[aria-hidden="true"]` via `closest()`, alongside the existing `inert`/`hidden`/computed-style checks. Runs per-Tab only.
- `useAnnounce.ts`: after each announce, a per-region timeout (2s, reset on every announce) clears the region text. Explicit `clearRegion` and the test reset helper cancel pending timers. The clear-then-rAF re-announcement pattern is untouched; the delay comfortably outlasts the rAF re-set and test assertion windows.
- `useHotkeys.ts`: JSDoc note documenting the WCAG 2.1.4 (Character Key Shortcuts) obligation for unmodified single-character shortcuts — must be remappable, disableable, or focus-scoped. No behavior change.

## Test plan

- New tests (all confirmed failing before the fixes, 6 failures pre-fix):
  - same-commit nested trap mount: outer + inner rendered together, Escape fires the INNER trap's `onEscape` only; after the inner unmounts, Escape falls back to the outer
  - focusable elements inside (or carrying) `aria-hidden="true"` are excluded from `focusFirst` and from the Tab wrap boundary
  - live region content survives ~1s, clears after the delay (fake timers), the clear timer resets on each announce, and re-announcement works after a clear
- `node_modules/.bin/vitest run packages/core/src/hooks --reporter=dot` — 14 files, 183 tests passed
- Full core suite (critical here — overlays + every announce consumer: CommandPalette, Toast, Selector, Calendar, Typeahead, etc.): `node_modules/.bin/vitest run packages/core --reporter=dot` — **222 files, 5165 tests, all passed** (no announce-timing fallout, known Calendar/Typeahead flakes did not reproduce)
- `node_modules/.bin/tsc --project packages/core/tsconfig.json --noEmit` — clean
- `eslint` + `prettier --check` on changed files — clean
- `node scripts/check-changesets.mjs` — 45 changesets valid

## Notes

- Vercel deployment failure on fork PRs is known infra and unrelated to this change.

### DIFF
--- a/.changeset/a11y-focus-trap-hardening.md
+++ b/.changeset/a11y-focus-trap-hardening.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] hooks: resolve focus-trap Escape by DOM depth instead of push order, exclude aria-hidden subtrees from trap tab cycles, and auto-clear live regions after announcing so stale status text does not linger
+@bhamodi
--- a/packages/core/src/hooks/useAnnounce.test.tsx
+++ b/packages/core/src/hooks/useAnnounce.test.tsx
@@ -9,8 +9,8 @@
  * SYNC: When useAnnounce.ts changes, update tests to match new behavior
  */
 
-import {describe, it, expect, afterEach} from 'vitest';
-import {renderHook, waitFor} from '@testing-library/react';
+import {describe, it, expect, afterEach, vi} from 'vitest';
+import {act, renderHook, waitFor} from '@testing-library/react';
 import {useAnnounce, __resetLiveRegionsForTest} from './useAnnounce';
 
 afterEach(() => {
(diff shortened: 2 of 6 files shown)


cixzhang:
It's crazy how simple announcements for a11y needs this complicated song and dance managing live regions.

Thanks for handling these details.

cixzhang (approved):

Issue #4295, opened 2026-07-24
Functional test coverage gaps: viz packages (charts/vega/lab) at ~0%, plus untested core hooks


humbertovirtudes:
## Problem

A sweep of the shippable packages (`core`, `lab`, `charts`, `vega`) found that **functional test coverage is uneven**: `core` is well-covered, but the visualization packages are almost entirely untested, and a cluster of core hooks have no direct tests.

"Functional" here = a test that renders + asserts, drives interaction (`fireEvent`/`userEvent`), queries the DOM (`getByRole` etc.), or exercises a hook via `renderHook` — not type-only, snapshot-only, or codemod string-match tests.

## Findings (public exports)

| Package | Components func-tested | Hooks func-tested |
|---|---|---|
| `core` | 161 / 173 (**93%**) | 44 / 64 (**69%**) |
| `lab` | 16 / 52 (**31%**) | 1 / 2 |
| `charts` | 0 / 6 (**0%**) | 0 / 1 |
| `vega` | 0 / 1 (**0%**) | — |

**Total gap: ~55 components + ~22 hooks** with no functional test.

### 1. Visualization packages are the biggest hole (`lab` charts, `charts`, `vega`)
The only viz tests today cover pure-math helpers (color scales, projection, range) plus one `SankeyChart` render. Every actual chart-primitive component renders nothing under test — all `Chart*`, `Radial*`, `ThreeD*`, `Sankey{Grid,Label,Link}`, `SVGIcon` in `lab`; the whole `charts` primitive set (`ChartAxis/Grid/Legend/Swatch/Tooltip`); and `VegaChart`.

### 2. Core hooks lack direct tests (logic-heavy, high value)
20 exported core hooks have no dedicated test. Many run indirectly when their parent component renders (e.g. `useTooltip` via `Tooltip.test.tsx`), but none isolate the hook's own logic/edge cases via `renderHook`. Examples: `useTooltip`, `useHoverCard`, `useLightbox`, `useCollapsible`, `useScrollLock`, `useScrollOverflow`, `useTruncation`, `useTranslator`, the three `useCalendar*` hooks, `useTableFilterState`.

### 3. A handful of core components have zero tests
`ChatComposerDrawer`, `ChatSendButton`, `CommandPaletteEmpty`, `InputClearButton`, `MobileNavToggle`, `PowerSearchFilterEditor`, `PowerSearchToken`, `TableHeaderCell`, `TypeaheadItem`.

## Relationship to #4163
[...]

PR #3721, opened 2026-07-09
fix(core): expose aria-expanded and aria-controls on MobileNav toggle

bhamodi:
## Summary

`MobileNavToggle` rendered a `Button` with an `aria-label` and `onClick`, but exposed **no `aria-expanded`** and **no `aria-controls`** (`packages/core/src/MobileNav/MobileNavToggle.tsx`). Screen-reader users had no way to tell whether the nav drawer was currently open, nor which element the hamburger button controls.

Making `aria-controls` resolve required a shared id, which did not exist: the AppShell mobile context (`useAppShellMobile()`) exposed no drawer id, and the `MobileNav` `<dialog>` had no `id` at all. This wires that up end to end:

- `AppShell` generates a stable `useId()` and publishes it on the mobile context as `mobileNavId`.
- `MobileNav` applies that id to its `<dialog>` (falling back to a locally generated `useId()` when used standalone outside AppShell).
- `MobileNavToggle` sets `aria-expanded={isMobileNavOpen}` and `aria-controls={mobileNavId || undefined}`.

`aria-controls` is set whenever the id resolves: the native `<dialog>` stays in the DOM whenever the layout is below the mobile breakpoint -- it is mounted via React `<Activity mode="hidden">` (or a plain fragment fallback) and simply hidden when closed -- so the referenced id always resolves inside an AppShell. Purely additive; no behavior or visual change.

`mobileNavId` is **optional** on `AppShellMobileContextValue` (`mobileNavId?: string`) and is applied entirely internally, so adding it is **not a breaking change**: callers that construct the context by hand do not need to supply it. When absent (standalone `MobileNav`, or a hand-built context), `MobileNav` falls back to its own `useId()` and the toggle drops `aria-controls`.

## Changes
- `MobileNav/MobileNavToggle.tsx`: forward `aria-expanded` + `aria-controls` from context onto the toggle button.
- `MobileNav/MobileNav.tsx`: apply the shared `mobileNavId` (with a standalone `useId()` fallback) as the `<dialog>` `id`.
- `AppShell/AppShell.tsx` + `AppShell/AppShellMobileContext.tsx`: generate and publish `mobileNavId` on the mobile context, as an **optional** field (omitted from the default context value).
- `AppShell/useAppShellMobile.doc.mjs`: document the new (optional) `mobileNavId` return field.

## Test plan
[...]

PR #4377, opened 2026-07-26
test: route charts/vega into the ui vitest project and cover the untested viz packages and hooks (#4295)

AKnassa:
## The bug underneath the coverage gap

`packages/vega` has no route into the vitest `ui` project, so its component tests fall through to the `node` project: no jsdom, no StyleX transform, no jest-dom. Main has since fixed the same problem for `charts` (routing plus a bare `@astryxdesign/core` alias); this PR adds the still-missing `vega` include and exclude (3 lines in `vitest.config.ts`) and vega's `test` scripts.

## What this adds

Following #4295, in its order:

1. **Core hooks**: `renderHook` tests for the three `useCalendar*` hooks, `useCollapsible`, `useTableFilterState`, `useScrollLock`, `useLightbox`, `useChatPasteAsToken`, `useImperativeAlertDialog`, `useTranslator`, `useTooltip`.
2. **Viz render smoke**: `charts` (every primitive, the marks, the tooltip's pointer seam), `vega` (`VegaChart`'s full lifecycle against a mocked vega/vega-lite boundary), and `lab`'s `Chart*`, `Radial*`, `ThreeD*`, `Sankey*` and `SVGIcon` families.
3. **Zero-test components**: `CommandPaletteEmpty`, `InputClearButton`, `TypeaheadItem`, `ChatSendButton`, `PowerSearchToken`, `PowerSearchFilterEditor`, `TableHeaderCell`.

**66 test files, 509 tests. No source file is changed.** The only non-test edits are the vega route, vega's test scripts, and a shared StyleX read-back helper (`packages/core/src/__tests__/stylexDeclarations.ts`).

Eight of these files were also added on main while this branch was open (`charts/Chart`, `InputClearButton`, `TypeaheadItem`, `useScrollLock`, `lab/Chart`, `RadialChart`, `SVGIcon`, `ThreeDChart`). Those keep main's version; `ea311aae4` ports back only the assertions main does not cover and drops the duplicates. One of main's assertions is tightened along the way: `InputClearButton`'s `onClick` event check (`toBeInstanceOf(Object)` passes for any object) now pins `type: 'click'`.

One correction to the issue: `ChatComposerDrawer` (#3996) and `MobileNavToggle` (#3721) already had tests.

## How it was checked

- The 66 files: 509 passing. The wider set of every test file the merge touched: 288 files / 7955 passing.
- ESLint and Prettier clean on every edited file; `typecheck` clean for core, lab, charts and vega (test files included).
- Sampled mutation testing from the first round still holds: inverting the send-button variant ternary, dropping `CommandPaletteEmpty`'s `xstyle` merge, and deleting `useTooltip`'s touch guard each fail exactly one test.
- No changeset: nothing shipped changes.

## Deliberately left uncovered

- **`useTooltip`'s focus paths.** jsdom resolves `:focus-visible` as plain `:focus`, so a test of that gate would pass identically with the gate deleted.
- **The ignored `onOpenChange(true)` guards** in `useLightbox` and `useImperativeAlertDialog`. Nothing emits `true` through the public surface, and forcing it via element props trips `no-react-introspection`.

## Source observations, not fixed here

- `lab/SVGIcon.tsx:131-153` spreads `{...attrs, key}` into each shape, so React 19 logs a key-in-spread warning per shape (pre-existing on main).
- `lab/Chart.tsx:333` emits a hard-coded `clipPath id="astryx-chart-plot"` that nothing references; `charts/Chart.tsx` already uses `useId` for the same thing.
- `TypeaheadItem.tsx:142-144` drops `ref`, `className`, `xstyle`, `style` and the pass-through props when `item.element` is set, and the `group` prop is documented but unused.

Refs #4295


cixzhang:
Thanks for this — the Vega coverage is still wanted, but I can't refresh this branch mechanically now. Main has independently added eight of these test files and the charts test routing, so resolving the add/add conflicts requires choosing which assertions to keep rather than just taking both sides.

Could you update onto current main and narrow the config change to the still-missing Vega route? Once that is current, we can review the remaining tests against today's behavior.

[Reviewed by Robohands]

AKnassa:
Updated onto current main (`c6041cfd5`). The config change is now only the vega route (3 lines in `vitest.config.ts`); the charts routing and the bare `@astryxdesign/core` alias come from main.

The eight add/add files keep main's version, and `ea311aae4` ports back only the assertions main does not cover (render gates and sizing, layer groups and masks, scroll-lock style restore, registry glyph, xstyle forwarding, context guards). Duplicates of main's coverage are dropped rather than kept twice.

Four files had gone stale and now follow today's behavior: chart label and tooltip queries scope to the svg because of the hidden data-table fallback (#4382), useTooltip's touch guard is per pointer type (#5248) and a controlled tooltip reports Escape through `onHide` (#4881), and TableHeaderCell dividers are logical inline-end properties (#4434).

Locally: 66 files / 509 tests green, eslint and prettier clean, typecheck clean for core, lab, charts and vega.

cixzhang (approved):
🟢 No review blockers found — Vega’s 30 tests route exclusively through the UI project, and the focused chart/Vega suites pass at this head.

[Reviewed by Robohands]

PR #4378, opened 2026-07-26
fix(core): collections ARIA minors — Resizable collapsed value, TabMenu radio semantics

bhamodi:
## Summary

Two collection-component minors from a11y scan #3, both WCAG 4.1.2 (Name, Role, Value) / APG conformance issues:

1. **Resizable** — a collapsed panel forced `_size` to 0, so the handle rendered `aria-valuenow="0"` while `aria-valuemin` was the panel's `minSizePx` (default 50). A `valuenow` below `valuemin` is invalid, and AT clamp/ignore behavior is unpredictable.
2. **TabMenu** — the overflow menu is single-select (exactly one option can be the active tab), but options rendered as plain `role="menuitem"` with `aria-current`, so screen readers never announced which option was selected as part of the role. Per the APG menu-button pattern, single-select menu options should be `menuitemradio` with `aria-checked`.

> **Scope note (review follow-up):** this PR originally also dropped the invalid `aria-level`/`aria-expanded`/`aria-selected` row states from the Table tree/selection plugins. Per review, Table needs a holistic treatment (grid role + grid keyboard navigation + swappable a11y plugins), so all Table changes were removed from this PR and will be tackled in a dedicated table-focused PR.

### Resizable (`ResizeHandle.tsx`)
- While collapsed, `aria-valuenow` clamps to `aria-valuemin` (never below the declared minimum) and a localized `aria-valuetext` of "Collapsed" (new catalog key `@astryx.resizable.collapsed`) announces the true state.
- On expand, the valuetext is removed so the numeric value reads again. Nothing changes for the non-collapsed path.

### TabMenu (`TabMenu.tsx`)
- Options now render `role="menuitemradio"` + `aria-checked` (true on the selected tab, false otherwise); `aria-current` is dropped.
- `useListFocus` gets `itemSelector: MENU_ITEM_SELECTOR` (the shared selector from `DropdownMenu/menuItemRoles.ts`, which already covers `menuitem`/`menuitemradio`/`menuitemcheckbox` for DropdownMenu and ContextMenu) — the hook's default `'[role="menuitem"]'` selector would no longer match, so arrow/Home/End roving-tabindex navigation keeps working. Existing keyboard tests were updated to the new role and still pass.

## Test plan

All from repo root (re-verified after rebasing onto current `main` and removing the Table changes):

- `node_modules/.bin/prettier --check <changed files>` — clean.
- `node_modules/.bin/eslint <changed files>` — 0 errors (the new Resizable string goes through the i18n catalog, satisfying `@astryx/no-hardcoded-i18n-string`; key format passes `i18n-key-format`).
- Scoped: `vitest run packages/core/src/Resizable packages/core/src/TabList packages/core/src/Table` — **24 files, 530 tests passed** (Table suites confirm the revert leaves them exactly at `main`'s behavior).
- Pre-fix failure confirmed (original submission): with the source files reverted to `HEAD` (tests kept), the updated/new Resizable + TabMenu tests fail, then pass again with the fix restored.
- CI runs the full monorepo suite on the rebased head.

## Notes

### DIFF
--- a/.changeset/a11y-collections-minors.md
+++ b/.changeset/a11y-collections-minors.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Resizable, TabMenu: two collection ARIA minors (WCAG 4.1.2) — Resizable's collapsed handle clamps `aria-valuenow` to `aria-valuemin` and announces a localized "Collapsed" via `aria-valuetext`, and TabMenu overflow options are `menuitemradio` with `aria-checked` (APG menu-button single-select) instead of `menuitem` + `aria-current`.
+@bhamodi
--- a/packages/core/locales/en.json
+++ b/packages/core/locales/en.json
@@ -655,6 +655,10 @@
     "defaultMessage": "Search…",
     "description": "Grey placeholder text inside the empty PowerSearch input. Imperative verb; trailing `…` is one character."
   },
+  "@astryx.resizable.collapsed": {
+    "defaultMessage": "Collapsed",
+    "description": "aria-valuetext for the Resizable drag handle (role=separator) while its panel is collapsed to zero size. Replaces the numeric value announcement, since the numeric value is clamped to the minimum while collapsed. Adjective describing the panel state, not a verb/command."
+  },
   "@astryx.resizable.handle.label": {
     "defaultMessage": "Resize handle",
     "description": "Screen-reader-only accessible name for the small draggable divider between two Resizable panels (users drag it to change the split ratio)."
--- a/packages/core/src/Resizable/ResizeHandle.test.tsx
+++ b/packages/core/src/Resizable/ResizeHandle.test.tsx
@@ -150,7 +150,55 @@ describe('ResizeHandle', () => {
     const separator = getSeparator();
     act(() => separator.focus());
     fireEvent.keyDown(separator, {key: 'Enter'});
-    expect(separator).toHaveAttribute('aria-valuenow', '0');
+    // The panel's real size is 0, but aria-valuenow must never drop below
+    // aria-valuemin (WCAG 4.1.2) — it clamps to the minimum and the state
+    // is announced via aria-valuetext instead.
+    expect(separator).toHaveAttribute('aria-valuenow', '100');
+    expect(separator).toHaveAttribute('aria-valuetext', 'Collapsed');
+  });
+
+  it('keeps aria-valuenow >= aria-valuemin and announces "Collapsed" while collapsed', () => {
+    render(
+      <Harness
+        config={{
+          defaultSize: 200,
+          minSizePx: 100,
+          maxSizePx: 400,
+          collapsible: true,
+        }}
+      />,
+    );
+    const separator = getSeparator();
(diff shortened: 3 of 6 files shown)


packages/core/src/Table/plugins/selection/useTableSelection.tsx
```diff
@@ -20,12 +20,19 @@
...
  *
- * Row-level styling (aria-selected, background) uses imperative DOM
+ * Row-level styling (data-selected, background) uses imperative DOM
  * updates via a ref callback on each <tr>. The ref subscribes to the
  * store and applies/removes styles when selection changes — no extra
  * DOM elements and no central element tracking. Each subscription
  * self-cleans when the row disconnects.
  *
+ * The rows deliberately carry NO aria-selected: that row state is only
+ * valid under grid/treegrid, and Table renders role="table" (WCAG
```
cixzhang: I think we should probably make it grid but we will need to do this carefully since we will need the table to apply grid navigation at the same time. Internally we're using table plugins (internally in web table) to achieve this.  Additionally, editable worksheets might also need slightly different keyboard controls so accessibility plugins may need to be something we can swap. Mind if we skip Table here and tackle that more holistically in a single PR that's more table focused?

cixzhang (changes requested):

packages/core/src/Table/plugins/selection/useTableSelection.tsx
bhamodi: Makes sense -- grid semantics only earn their keep when the table also delivers the grid keyboard contract (cell-to-cell arrows, roving tab stop), and swappable a11y plugins for worksheet-style keyboarding is a bigger design than this PR should carry. All Table changes are now dropped from this PR: `packages/core/src/Table/**` is byte-identical to `main`, and the PR is retitled to its remaining scope (Resizable collapsed value + TabMenu radio semantics). The Table row-state work can ride in the dedicated table-focused PR alongside the grid migration.

bhamodi:
@cixzhang Done -- all Table changes are dropped from this PR (`packages/core/src/Table/**` is byte-identical to `main`); scope is now Resizable + TabMenu only, and the title/description are updated to match. Happy to fold the row-state work into the holistic table-focused PR when that kicks off. Rebased onto current main; scoped suites green (24 files / 530 tests).

bhamodi:
pr-a11y flagged two `TabList::*::landmark-unique` violations as new-vs-baseline. Not from this PR -- the remaining diff only changes TabMenu overflow options (inside a closed popover) and Resizable's handle values; it renders no landmarks -- so these postdate the #4386 baseline seed on main. Added both keys to `.github/a11y-baseline.json` per the documented escape hatch; flagging for reviewer blessing. The real fix (unique accessible names when multiple TabLists render) belongs on main.

cixzhang (approved):

bhamodi:
Rebased onto current main. The two `TabList::*::landmark-unique` escape-hatch keys this PR had added to `.github/a11y-baseline.json` were seeded on main by #4603, so this PR no longer touches the baseline file at all -- the diff is now purely Resizable + TabMenu.

PR #4380, opened 2026-07-26
fix(lab): RichTextEditor Tab keyboard trap — Escape then Tab moves focus

bhamodi:
## Summary

`RichTextEditor` (lab) always registers Lexical's `TabIndentationPlugin`, which rebinds Tab/Shift+Tab to indent/outdent and `preventDefault()`s the keydown. A keyboard-only user who tabs into the editor cannot tab out again -- a **serious WCAG 2.1.2 (No Keyboard Trap)** violation. Found as finding #3 of the a11y scan; this is the companion to the CodeEditor Tab-trap fix (same UX pattern: keep Tab indentation, add an Escape-then-Tab exit).

Tab indentation itself is worth keeping -- it is how list/quote nesting works -- so instead of removing it, this PR adds a standard escape:

- **Escape arms an exit**: the next Tab (or Shift+Tab) performs native focus movement out of the editor instead of indenting.
- **Any other non-modifier key re-arms indentation**, so normal editing is unaffected.
- The escape is **advertised to screen-reader users** via a visually hidden hint wired into the contenteditable's `aria-describedby`, overridable/suppressible through a new `tabEscapeHint` prop (lab has no i18n catalog; the default-English-string-via-prop pattern matches `InfoTip`'s `label`).

## Changes

**`packages/lab/src/RichTextEditor/RichTextEditor.tsx`**

- New internal `TabFocusEscapePlugin`, rendered alongside `TabIndentationPlugin`. Lexical mechanics: `TabIndentationPlugin` handles `KEY_TAB_COMMAND` at `COMMAND_PRIORITY_EDITOR` (0), the lowest priority, and Lexical runs command listeners from highest priority down, stopping at the first that returns `true`. The new plugin registers at `COMMAND_PRIORITY_LOW` (1), so it always runs first:
  - `KEY_ESCAPE_COMMAND` -> arm the escape. This handler returns `true`, deliberately superseding `@lexical/rich-text`'s default Escape handler, which calls `editor.blur()` and dumps focus on `document.body` (disorienting, and it would immediately disarm the escape via the blur listener below). Consumer plugins that handle Escape at a higher priority still run first.
  - `KEY_TAB_COMMAND` -> if armed: disarm and return `true` **without** `preventDefault()`, so the indentation handler never sees the event and the browser performs its native Tab focus navigation. If not armed: return `false` and indentation proceeds as before.
  - `KEY_DOWN_COMMAND` -> any key other than Escape/Tab/bare modifiers disarms (bare `Shift` is exempt so Escape then Shift+Tab can escape backwards).
  - `BLUR_COMMAND` -> disarm when focus leaves the editor.
- New `tabEscapeHint` prop (default: `'Press Escape then Tab to move focus out of the editor.'`), rendered via core's `VisuallyHidden` and appended to the contenteditable's `aria-describedby`. Only rendered when the editor is editable; pass `''` to omit.
- File header `@input` updated per the repo documentation protocol.

**`packages/lab/src/RichTextEditor/RichTextEditor.doc.mjs`** -- props-table entry for `tabEscapeHint` (targeted insert only; file otherwise untouched).

**`packages/lab/src/RichTextEditor/RichTextEditor.test.tsx`** -- 7 new tests (see below).

## Test plan

```
node_modules/.bin/vitest run packages/lab/src/RichTextEditor --reporter=dot   # 1 file, 23 passed (was 16)
node_modules/.bin/vitest run packages/lab --reporter=dot                      # 19 files, 260 passed
node_modules/.bin/tsc --project packages/lab/tsconfig.json --noEmit           # no new errors (see note)
node_modules/.bin/eslint <changed files>                                      # clean
node_modules/.bin/prettier --check <changed files>                            # clean
```

New tests (all driven through real keydown events on the contenteditable -- `userEvent.tab()` / `userEvent.keyboard()` respect `defaultPrevented`, so they exercise the actual trap):

1. Tab indents a seeded bullet-list item (indent 0 -> 1) and focus stays in the editor.
2. **Escape then Tab moves `document.activeElement` out of the editor** (onto the next focusable button) without indenting.
3. Pressing another key after Escape re-arms indentation (Tab indents again, focus stays).
4. A bare `Shift` keydown does not re-arm, so Escape then Shift+Tab still escapes (asserted via `defaultPrevented`).
5. The default hint is rendered visually hidden and its id appears in the contenteditable's `aria-describedby`.
6. `tabEscapeHint` is overridable and `''` suppresses the hint (and drops it from `aria-describedby`).
7. No hint when `isReadOnly` (editor is not editable, so there is no trap).

**Pre-fix failure confirmed**: with the tests added but the component fix absent, the 5 behavior/hint tests fail exactly as expected (test 2 shows focus unable to leave the editor); tests 1 and 7 pass pre-fix, as those behaviors already existed. Post-fix: 23/23.

jsdom notes (honest limitations):

- jsdom's `Range` lacks `getBoundingClientRect`, which Lexical calls when reconciling a focused collapsed selection; the test suite stubs it (returns a zero `DOMRect`) inside the new `describe` block only.
- jsdom does not implement contenteditable text editing, so "typing re-arms" is driven as a raw `keydown` (which is precisely the re-arm signal the implementation uses). Focus movement itself is real: `user-event` implements Tab's focus-change behavior and honors `preventDefault`, so test 2 genuinely observes `activeElement` leaving the editor.
- Manual step worth doing in a browser: Escape's previous default (Lexical rich-text `editor.blur()`) is intentionally replaced by the arm behavior -- focus now stays in the editor after Escape until Tab is pressed.

## Notes

- **No changeset**: `scripts/check-changesets.mjs` hard-rejects `@astryxdesign/lab` ("private/ignored and cannot be released"). Lab is canary-only (`private: true` is npm's guarantee against a stable publish) and is changeset-exempt per the RFC #3900 landing ("Lab is canary-only and private, so no changeset is required"); CI stamps canary versions. The check script passes with no changeset in this PR.
- `tsc --project packages/lab/tsconfig.json` reports 7 pre-existing `TS2305` errors in `Chart`/`ThreeD` (`parseHex` etc. missing from `@astryxdesign/core/utils`); output is byte-identical on clean `main` -- untouched by this PR.
- The Vercel deploy check fails on every fork PR (known infra); `review-required` is a human gate.

## Review follow-up (2026-07-31)

- **`packages/build` is no longer touched** — byte-identical to `main`, alias tests and its changeset removed. The `lexical$` → built-ESM pin (needed because this fix adds runtime `lexical` imports, which `withAstryx`'s issuer-scoped `source`-condition rule would otherwise resolve to lexical's raw TS with `declare` class fields that the sandbox Babel pipeline rejects) now lives in `apps/sandbox/next.config.mjs`, with the same issuer-scoped rule shape. App-local is sufficient: `@astryxdesign/lab` is private, so no external `withAstryx` consumer can ever build lab from source — the shared build package stays agnostic to libraries outside core.
- On the earlier review note about the pin bypassing dev/prod condition selection: `dist/Lexical.mjs` is an env-switching wrapper (`process.env.NODE_ENV !== 'production' ? dev : prod`), so build-time env selection is preserved.
- Rebased onto current `main`; the conflict with the new `getHTML()`/`getMarkdown()` imperative-ref work was in the `lexical` import block only (runtime command imports merged with `$generateHtmlFromNodes` + type-only imports). `vitest run packages/lab/src/RichTextEditor`: 46/46 green post-merge.

### DIFF
--- a/apps/sandbox/next.config.mjs
+++ b/apps/sandbox/next.config.mjs
@@ -1,6 +1,29 @@
 // Copyright (c) Meta Platforms, Inc. and affiliates.
 
+import {createRequire} from 'node:module';
+import path from 'node:path';
+
 import {withAstryx} from '@astryxdesign/build/next';
+
+const require = createRequire(import.meta.url);
+
+/**
+ * Locate lexical's built ESM entry for the alias below. `require.resolve`
+ * lands on the CJS entry inside dist/, so walk up to the package root rather
+ * than resolving `lexical/package.json` (not export-mapped). Returns null when
+ * lexical isn't installed so the alias is simply omitted.
+ */
+function resolveLexicalDist() {
+  try {
+    let dir = path.dirname(require.resolve('lexical'));
+    while (path.basename(dir) !== 'lexical') {
+      dir = path.dirname(dir);
+    }
+    return path.join(dir, 'dist', 'Lexical.mjs');
+  } catch {
+    return null;
+  }
+}
 
 /** @type {import('next').NextConfig} */
 const nextConfig = {
@@ -18,6 +41,28 @@ const nextConfig = {
     ignoreBuildErrors: false,
   },
   webpack(config, {dev}) {
+    // withAstryx resolves requests issued FROM @astryxdesign modules with the
+    // `source` condition so astryx packages build from TS source. lexical also
+    // ships a `source` export, so lab's bare runtime `import ... from
+    // 'lexical'` (RichTextEditor keyboard commands) would resolve to lexical's
+    // raw TS — whose `declare` class fields this app's Babel pipeline rejects.
+    // Pin bare `lexical` to its built ESM entry (an env-switching wrapper, so
+    // dev/prod selection is preserved). App-local on purpose: lab is a private
+    // package, so no external withAstryx consumer can hit this — the shared
+    // build package stays agnostic to non-core libraries.
+    // Scoped by issuer (same shape as withAstryx's source-condition rule) so
+    // only astryx-issued requests are pinned — everything else keeps normal
+    // export-map resolution (e.g. Lexical.node.mjs in the server bundle).
+    const lexicalDist = resolveLexicalDist();
+    if (lexicalDist != null) {
+      config.module = config.module || {};
+      config.module.rules = config.module.rules || [];
+      config.module.rules.unshift({
+        test: /[\\/]node_modules[\\/]@astryxdesign[\\/]/,
+        resolve: {alias: {lexical$: lexicalDist}},
+      });
+    }
+
     // Workarounds for Node v24+ where webpack's hashing/caching feeds
     // `undefined` to Hash.update and crashes the dev server.
     config.output = config.output || {};
--- a/packages/lab/src/RichTextEditor/RichTextEditor.doc.mjs
+++ b/packages/lab/src/RichTextEditor/RichTextEditor.doc.mjs
@@ -110,6 +110,13 @@ export const docs = {
       default: 'false',
     },
     {
+      name: 'tabEscapeHint',
+      type: 'string',
+      description:
+        'Screen-reader hint describing how to move focus out of the editor, since Tab is bound to indentation (press Escape, then Tab). Visually hidden, wired via aria-describedby. Override to localize; pass "" to omit.',
+      default: "'Press Escape then Tab to move focus out of the editor.'",
+    },
+    {
       name: 'maxLength',
       type: 'number',
       description:
--- a/packages/lab/src/RichTextEditor/RichTextEditor.test.tsx
+++ b/packages/lab/src/RichTextEditor/RichTextEditor.test.tsx
@@ -9,11 +9,17 @@
  * SYNC: When the editor components change, update these tests to match.
  */
 
-import {describe, it, expect, vi} from 'vitest';
+import {describe, it, expect, vi, beforeAll} from 'vitest';
 import {render, screen, waitFor, fireEvent} from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
 import {createRef, useEffect} from 'react';
 import {useLexicalComposerContext} from '@lexical/react/LexicalComposerContext';
 import type {EditorState, LexicalEditor} from 'lexical';
+import {
+  $createListItemNode,
+  $createListNode,
+  $isListItemNode,
+} from '@lexical/list';
 import {
   $getRoot,
   $createParagraphNode,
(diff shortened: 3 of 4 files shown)


cixzhang:
@potatowagon can you help take a look :)

potatowagon:
## Review

Solid, ship-worthy. Correct WCAG 2.1.2 fix, well-tested (ran locally, 42/42 pass), no regressions found. Traced the command-priority reasoning against Lexical's actual dispatch order — it holds up. A few minor items below, none blocking.

### ✅ Verified correct

- **Command ordering is genuinely safe.** Lexical's `onKeyDown` dispatches `KEY_DOWN_COMMAND` first, and `$handleKeyDown` (a KEY_DOWN listener) re-dispatches `KEY_TAB`/`KEY_ESCAPE` within that same chain. So on a Tab press the plugin's KEY_DOWN handler sees `key === 'Tab'` — which is in `ESCAPE_REARM_EXEMPT_KEYS`, so it does **not** disarm before the KEY_TAB handler reads the flag. That exempt-set is load-bearing and correct.
- **Priority claim checks out.** `TabIndentationPlugin` registers at `COMMAND_PRIORITY_EDITOR` (0); this plugin at `LOW` (1) runs first. Returning `true` *without* `preventDefault()` is the right lever to block indent while letting native focus move.
- **Build alias scoping is correct.** `lexical$` exact-match (the `$` suffix) only pins the bare `lexical` request, not `@lexical/*` subpackages — so the `@lexical/*` TS-source path the comment relies on is preserved. `resolveLexicalDist` fails safe (returns `null` → alias omitted).

### 🟡 Minor

1. **The build alias pins to `Lexical.mjs` (the `default` export condition).** This bypasses webpack's `development`/`production` condition selection, so prod builds of any `withAstryx` consumer get the un-minified `default` bundle rather than `Lexical.prod.mjs`. Harmless in lab/canary, but it's a subtle size/perf regression for external consumers of the wrapper. Consider resolving the dist dir and letting webpack pick the condition, or a comment noting `default` is intentional.
2. **Hardcoded `'Lexical.mjs'` filename** is brittle against a future lexical package-layout change. Low risk (this name has been stable for years), but a version bump could silently break the alias → fall back to TS source → the exact `declare`-class-field Babel error this PR fixes returns. A comment noting the coupling, or globbing `dist/Lexical*.mjs`, would harden it.

### 🔵 Nits

3. The `'use strict'` quote-style flip in `next.js` is unrelated prettier churn — harmless but adds blame noise.
4. `DEFAULT_TAB_ESCAPE_HINT` is duplicated as a literal in both the component and the test's `DEFAULT_HINT`. Minor drift risk; fine as-is since it's not exported.
5. The behavior tests don't guard the *relative* command priority (only observable behavior) — if Lexical ever changes `TabIndentationPlugin`'s registration priority this silently breaks. Acceptable tradeoff, just flagging the coupling to Lexical internals.

### 🔎 Edge cases considered — all OK

- Escape → click away → refocus → Tab: `BLUR_COMMAND` disarms on focus loss, so a stale-armed escape can't leak across focus sessions. ✅
- Escape → Escape → Tab: Escape is exempt from re-arm, just re-arms; still escapes. ✅
- Consumer Escape handlers (popovers): higher-priority listeners still run first, as documented. ✅
- Superseding rich-text's default Escape (`editor.blur()`): intentional and documented — worth the manual browser check you already flagged, since jsdom can't verify it.

potatowagon (approved):

packages/build/src/next.js
```diff
@@ -19,6 +19,25 @@
  * - Adds transpilePackages for @astryxdesign/* packages
  * - Sets conditionNames to resolve source exports
  */
+/**
+ * Locate lexical's built ESM entry for the alias below. `require.resolve`
+ * lands on the CJS entry inside dist/, so walk up to the package root rather
+ * than resolving `lexical/package.json` (not export-mapped). Returns null when
+ * lexical isn't installed so the alias is simply omitted.
+ */
+function resolveLexicalDist() {
```
cixzhang: Seems odd to need this in the build which might not have anything to do with lexical. Was this to solve some issue unrelated to accessibility?

cixzhang (changes requested):
I think accessibility updates look good but I'm not sure about the changes to the build package. This is shipped publically and used mainly to split the StyleX build for layering. It should be agnostic to libraries outside of core.

packages/build/src/next.js
bhamodi: It was needed by this PR, not unrelated: the trap fix adds runtime imports from `lexical` (`KEY_TAB_COMMAND`, `KEY_ESCAPE_COMMAND`, etc.), and `withAstryx`'s issuer-scoped `source`-condition rule makes any bare `lexical` request issued from an astryx module resolve to lexical's raw TS source, whose `declare` class fields the sandbox Babel pipeline rejects. Before this PR, lab only had type-only imports from `lexical`, so the problem never surfaced.

On the placement, agreed -- the shipped build package should stay agnostic to libraries outside core. The alias is now gone from `packages/build` entirely (byte-identical to `main`, alias tests and its changeset removed) and lives in `apps/sandbox/next.config.mjs` instead, with the same issuer-scoped shape so only astryx-issued requests are pinned. App-local is sufficient because `@astryxdesign/lab` is private: no external `withAstryx` consumer can ever build lab from source, so the sandbox is the only place this resolution path exists.

bhamodi:
@cixzhang Addressed -- `packages/build` is back to byte-identical with `main`; the lexical pin moved into `apps/sandbox/next.config.mjs` (details in the inline thread + the new 'Review follow-up' section of the description). @potatowagon your dev/prod-condition note turned out to be moot: `dist/Lexical.mjs` is an env-switching wrapper, so both modes still resolve correctly -- and it's now app-local anyway. Also rebased onto current main (the `getHTML()`/`getMarkdown()` import conflict resolved, 46/46 tests green).

cixzhang (approved):

bhamodi:
Rebased onto current main over the toolbar (#4551), link-support (#4670), and serializer (#4544) work -- the conflicts were import-merges only; the trap plugin, its command priorities, and the tests are unchanged. eslint/prettier pass locally; vitest for this branch needs the new `@lexical/headless` optional peer from #4544 which isn't installable offline here, so CI is the test verification for this push.

PR #3549, opened 2026-07-04
fix(lab): accessibility cleanup for CodeEditor, SankeyChart, and Schedule

cixzhang:
## Summary

Lab a11y cleanup addressing three axe-core findings from a Storybook audit. All three are in `packages/lab` (experimental components, lower a11y bar than core), but the fixes are cheap and the violations are real. Part of the ongoing accessibility program (#3343).

No consumer-facing API break; the only new surface is an optional `ariaLabel` prop on `CodeEditor`. `@astryxdesign/lab` is a canary-only (private) package, so there is no changeset (consistent with prior lab-only PRs).

### 1. CodeEditor - aria-input-field-name (serious)

**Problem:** The editable region is a `<code role="textbox" aria-multiline="true" contentEditable>` with no accessible name. Consumer-passed props spread onto the root `<div>`, not the inner `<code>`, so even an `aria-label` on the component would not name the textbox.

**Fix:** Added an optional `ariaLabel` prop and applied it to the `<code>` element, defaulting to a name derived from `language` (e.g. `"typescript code editor"`) or `"Code editor"`.

**Testing:** New `CodeEditor.test.tsx` asserts the textbox is reachable by its language-derived name and that an explicit `ariaLabel` overrides it.

### 2. SankeyChart - scrollable-region-focusable (serious)

**Problem:** When the chart is wider than its container it wraps the SVG in a div with `overflowX: auto`, but that scroll container was not keyboard-focusable, so keyboard users could not scroll it.

**Fix:** When `needsScroll` is true, the scroll container now gets `tabIndex={0}` plus `role="group"` and `aria-label="Sankey chart"`. When the chart fits, no extra attributes are added. Deliberately avoided `role="region"` to steer clear of landmark-uniqueness issues.

**Testing:** New `SankeyChart.test.tsx` drives a mocked `ResizeObserver` to force the scroll case and asserts the region is focusable, and asserts no focusable region appears when the chart fits.

### 3. Schedule - heading-order (moderate)

**Problem:** `ScheduleFrame` renders its title as an `<h2>` (`Heading level={2}`). The list day headings (`ListView`) and monthly weekday headings (`MonthlyView`) were `Heading level={4}`, so the heading order jumped h2 to h4, skipping h3.

[...]

PR #4382, opened 2026-07-26
fix(lab): accessible names and data-table fallbacks for charts

bhamodi:
## Summary

Charts were invisible to assistive technology (a11y scan finding #3, severity SERIOUS, WCAG 1.1.1 Non-text Content). The root `<svg>` elements of the lab charts had no `role="img"`, no accessible name, no `<title>`, and no text alternative for the plotted data -- a screen reader user got nothing at all from a Chart or RadialChart, and Sankey was only named when it happened to need horizontal scrolling.

This PR gives every chart root an accessible name and, for modest datasets, a visually hidden data-table text alternative (using the core `VisuallyHidden` primitive), so the underlying values are readable by assistive tech.

Lab has no i18n mechanism (other lab components use `Intl` for dates only, English strings elsewhere), so defaults are English and every chart accepts a `label` prop override for localized names. This is documented on each prop.

## Changes

- **`Chart` (packages/lab/src/Chart/Chart.tsx)** -- svg now has `role="img"` + `aria-label`. New `label` prop; defaults to `"Chart of {yKeys} by {xKey}"`. Visually hidden table (x column + one column per y key) rendered when rows x series <= 100 points (documented `MAX_TABLE_POINTS` cutoff -- beyond that a table is more noise than signal, name-only).
- **`RadialChart` (packages/lab/src/Radial/RadialChart.tsx)** -- svg now has `role="img"` + `aria-label`. New `label` prop; defaults to `"Radar chart of {axes}"` (spider) / `"Pie chart of {valueKey}"` (pie). Hidden table: pie mode lists label/value/percentage per slice; spider mode lists one row per series across the axes (row header uses the same first-string-field heuristic `RadialArea` uses for `dataKey`). Same <=100-point cutoff.
- **`SankeyChart` (packages/lab/src/Sankey/SankeyChart.tsx)** -- svg now has `role="img"` + `aria-label` in **both** the scrolling and non-scrolling paths (previously only the scroll wrapper from #3549 was named, and only when scrolling). New `label` prop (default `'Sankey chart'`) feeds both the svg and the scroll region. Hidden From/To/Value table of links (node labels resolved from node defs), <=100 links cutoff.
- **`Chart` v2 (packages/charts/src/Chart.tsx)** -- had the same gap, partially: `aria-label` only when `title` was set, and never `role="img"`. Now always `role="img"` with `aria-label = title ?? "Chart of {series labels} by {xKey}"` (per the convention, the existing `title` prop is preferred as the name -- no new prop). Same hidden data table (columns = union of primary series dataKeys, utility marks excluded), <=100-point cutoff.
- **`vitest.config.ts`** -- added `packages/charts` to the `ui` test project (it previously fell through to the DOM-less `node` project, exactly the migration path the config's partitioning comment prescribes) and mapped the bare `@astryxdesign/core` specifier to source so charts' runtime core imports resolve in tests.

Table-vs-name-only rationale per chart: Chart/Radial/charts-v2 receive structured rows/slices, so a table is meaningful; Sankey's graph is flattened to its links (source, target, value), which is the standard text serialization of a flow diagram.

## Test plan

New tests (15 new assertions-bearing tests across 4 files) were written first and confirmed failing pre-fix: `15 failed | 6 passed` (the 6 = 2 pre-existing Sankey scroll tests + 4 over-cutoff negatives that are trivially true pre-fix).

Post-fix:

- `prettier --check` on all 9 changed files -- clean
- `eslint` on all 9 changed files -- clean
- `tsc --project packages/lab/tsconfig.json --noEmit` -- 0 errors
- `tsc --project packages/charts/tsconfig.json --noEmit` -- 0 errors
- `vitest run packages/lab/src/Chart packages/lab/src/Radial packages/lab/src/Sankey packages/charts/src` -- **8 files, 47 tests passed**
- `vitest run packages/lab` -- **21 files, 268 tests passed**
- `node scripts/check-changesets.mjs` -- passes (44 changesets valid)

Coverage per chart: svg has `role="img"` + accessible name (default and custom `label`); data table renders with correct values for a small dataset and is visually hidden (StyleX clip class, not `display:none`/`hidden`); table skipped over the cutoff; Sankey named in both scroll and no-scroll paths, custom label applied to both svg and scroll region.

## Notes

- **No changeset**: both `@astryxdesign/lab` and `@astryxdesign/charts` are `private: true`; `scripts/check-changesets.mjs` rejects them ("private/ignored and cannot be released"), verified empirically. Per the checker's guidance, private packages are changeset-exempt.
- `ThreeDChart` is intentionally not touched here -- it is handled in a separate PR (#4381) using the same `label` + `role="img"` convention.
- The Vercel deployment failure on fork PRs is known infra and unrelated.

### DIFF
--- a/packages/charts/src/Chart.test.tsx
+++ b/packages/charts/src/Chart.test.tsx
@@ -1,0 +1,129 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+/**
+ * @file Chart.test.tsx
+ * @input Uses vitest, @testing-library/react, Chart root + bar mark
+ * @output Unit tests for Chart accessibility (WCAG 1.1.1)
+ * @position Testing; validates Chart.tsx accessible name + data-table fallback
+ */
+
+import {describe, it, expect, beforeEach, afterEach, vi} from 'vitest';
+import {render, screen, act} from '@testing-library/react';
+import {Chart} from './Chart';
+import {bar} from './marks';
+
+const data = [
+  {month: 'Jan', revenue: 100, profit: 20},
+  {month: 'Feb', revenue: 200, profit: 40},
+  {month: 'Mar', revenue: 150, profit: 30},
+];
+
+// Capture the ResizeObserver callback so tests can drive the reported width.
+let resizeCallback: ResizeObserverCallback | undefined;
+
+beforeEach(() => {
+  resizeCallback = undefined;
+  vi.stubGlobal(
+    'ResizeObserver',
+    class {
+      constructor(cb: ResizeObserverCallback) {
+        resizeCallback = cb;
+      }
+      observe() {}
+      unobserve() {}
+      disconnect() {}
+    },
+  );
+});
+
+afterEach(() => {
+  vi.unstubAllGlobals();
+});
+
+function reportWidth(width: number) {
+  act(() => {
+    resizeCallback?.(
+      [{contentRect: {width}} as unknown as ResizeObserverEntry],
+      {} as ResizeObserver,
+    );
+  });
+}
+
+describe('Chart accessible name', () => {
+  it('exposes the svg as a named image with a default label', () => {
+    render(
+      <Chart
+        data={data}
+        xKey="month"
+        series={[bar('revenue'), bar('profit')]}
+      />,
+    );
+    reportWidth(600);
+
+    expect(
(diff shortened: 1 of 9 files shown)


ernestt (approved):

cixzhang (approved):
Nice! I think geo had this so good to see it brought in.

PR #2192, opened 2026-05-18
scope workflow permissions per-job (least-privilege)

zurfyx:
Adopts the OSSF Scorecard: every workflow declares `permissions: {}` at the top (deny-by-default) and elevates permissions only on the specific jobs that need them. Eliminates two unused permissions (`pages: write`, `id-token: write`) that were declared workflow-wide in ci.yml, deploy.yml, and redeploy-preview.yml but consumed by no job (XDS uses gh-pages branch via raw git push, not the GitHub Pages deploy API; OIDC is not used for any publish).

The most material change is in ci.yml. Previously, all 10 jobs inherited `contents: write`, `pages: write`, `id-token: write`, `pull-requests: write` at the workflow level — including jobs that run PR-author code (test, build, pr-a11y, docsite-test). After this change, those jobs run with no write permissions; only deploy-preview retains `contents: write` (needed to push the preview to gh-pages branch) and the two pr-comment jobs retain `pull-requests: write` (needed to write the PR comment). This is the canonical mitigation for the tj-actions/changed-files supply-chain attack class.

Per-workflow summary:

  ci.yml                  WL: deny-all. Per-job: deploy-preview gets
                          contents:write; pr-comment(-update) get
                          pull-requests:write. Other 7 jobs run
                          read-only.

  deploy.yml              WL: deny-all. Per-job: deploy gets
                          contents:write. publish/canary jobs use
                          METACCIO_PUBLISH_TOKEN env-var only — no
                          GitHub-level write needed.

  redeploy-preview.yml    WL: deny-all. Per-job: contents:write +
                          pull-requests:read + actions:read on the
                          single redeploy-preview job.

  vibe-screenshots.yml    WL: deny-all. Per-job: deploy-screenshots
                          gets contents:write. build-previews and
                          screenshot run read-only.

  cleanup-previews.yml    WL: deny-all. Per-job scoping (no semantic
                          change — same perms, just scoped).
  prune-branches.yml      Same — per-job scoping, no semantic change.
  hardening-issue.yml     Same — per-job scoping, no semantic change.
  codeowner-gate.yml      Same — per-job scoping, no semantic change.
  codemod-verify.yml      Same — per-job scoping, no semantic change.
  cli-smoke-test.yml      Same — per-job scoping, no semantic change.
  lint.yml                Add explicit deny-all at workflow level for
                          OSSF Scorecard signal; no functional change.

PR #4383, opened 2026-07-26
ci: weekly full-suite accessibility scan with issue reporting

bhamodi:
## Summary

PR CI's `pr-a11y` job only audits components a PR touches. That leaves a blind spot: token or shared-hook changes can regress components no PR modified, and nothing ever scans the whole library. This adds a scheduled full-suite accessibility scan (evergreen a11y, from a11y scan #3): every week (and on manual dispatch) the entire Storybook story index is audited with axe-core, the JSON report is archived, and violations surface as a tracking issue instead of silently accumulating.

## Changes

- **`.github/workflows/a11y-weekly.yml`** (new)
  - Triggers: `schedule` (Mondays 05:23 UTC, off-peak) + `workflow_dispatch`.
  - `audit` job (permissions: `contents: read`): checkout, repo setup action, `pnpm build`, Storybook build (once), Playwright chromium install, then `.github/scripts/accessibility-audit.js` with **no `--components` filter** — the script audits every story in the Storybook index when the flag is absent.
  - Baseline-aware without depending on the baseline PR: the audit step passes `--baseline .github/a11y-baseline.json --fail-on-new` only behind a `[ -f .github/a11y-baseline.json ]` shell guard, so the workflow works both before and after that file lands (unknown flags are ignored by the current script's arg parser anyway).
  - Report + markdown summary uploaded as the `a11y-weekly-report` artifact with **30-day retention**; summary also lands in `$GITHUB_STEP_SUMMARY`.
  - Failure semantics: audit runs with `continue-on-error`, then a summarize step classifies the run (`clean` / `violations` / `failed` / `crashed`). The job goes red on `crashed` (script/infra breakage) or `failed` (baseline gate tripped once it exists); ungated violations stay green and are reported via the issue.
  - `report-issue` job (permissions: `issues: write`, `actions: read`), modeled on `hardening-issue.yml`: whenever the scan is not clean, opens **"Weekly a11y scan: violations found"** with the report summary, or comments the new summary on the existing open issue (dedup by title + `a11y-weekly-scan` label).
  - Top-level `permissions: {}` with least-privilege per-job grants (repo convention from #2192).
- **`.github/scripts/weekly-a11y-summary.js`** (new)
  - Reads the audit report, classifies run status, and emits the markdown summary (reusing the shared `lib/a11y-format.js` `buildA11ySection` formatter) plus step outputs. Truncates at 60k chars to stay under GitHub's issue-body limit.

`accessibility-audit.js` is deliberately **not** touched here — the baseline/diff logic lives in the parallel `ci/a11y-blocking` PR.

## Test plan

- `node --check .github/scripts/weekly-a11y-summary.js` — passes.
- Workflow YAML parsed with the repo's `yaml` package — valid; jobs/steps/cron verified.
- Dry-ran the summary script against five synthetic reports: violations, clean, missing report (crash), report + non-zero audit exit (gate failure), and the script's "Storybook not built" error-report shape — all classified and rendered correctly.
- Verified referenced paths exist: `.github/actions/setup/action.yml`, `.github/scripts/accessibility-audit.js`, `.github/scripts/lib/a11y-format.js`; artifact name is consistent between upload (audit) and download (report-issue).
- `prettier --check` clean on both files; `./scripts/add-copyright.sh --check` passes.
- The first real end-to-end run happens on schedule (or via manual `workflow_dispatch`) after merge — scheduled workflows only run from the default branch.

## Notes

- **Relationship to `ci/a11y-blocking`:** that PR adds `--baseline` / `--fail-on-new` to the audit script. This workflow already passes those flags conditionally (guarded on `.github/a11y-baseline.json` existing), so no follow-up edit is needed here when it merges — the weekly scan picks up baseline gating automatically.
- **Theme coverage:** Storybook's dark mode is a runtime global (`colorMode`, applied by the `withTheme` decorator in `apps/storybook/.storybook/preview.tsx`). The audit script builds story iframe URLs itself and doesn't plumb globals through, and modifying it is out of scope for this PR — so the weekly scan currently audits the default light scheme only. Dark-mode coverage (passing `&globals=colorMode:dark` through the script) is a documented follow-up, a small change once the audit-script PRs settle.
- **No changeset:** CI-only change, no published package touched; the repo has no changeset-required CI check (changesets are only consumed by `release.yml` versioning).
- The Vercel check failing on fork PRs is known infra and unrelated.

### DIFF
--- a/.github/scripts/weekly-a11y-summary.js
+++ b/.github/scripts/weekly-a11y-summary.js
@@ -1,0 +1,135 @@
+#!/usr/bin/env node
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+/**
+ * @description Summarizes a full-suite accessibility audit report for the
+ *   weekly a11y scan workflow. Reads the JSON report produced by
+ *   accessibility-audit.js, writes a markdown summary (used for both the
+ *   GitHub step summary and the tracking issue body), and emits step outputs
+ *   (status, total_violations, components_audited) so the workflow can decide
+ *   whether to open-or-update the tracking issue and whether to fail the job.
+ * @input --report <file> --audit-outcome <success|failure> --summary-output <file>
+ *   [--github-output <file>]
+ * @output Markdown summary file + GitHub Actions step outputs
+ *
+ * Status semantics:
+ *   clean      — report present, audit step succeeded, zero violations
+ *   violations — report present, audit step succeeded, violations found
+ *   failed     — report present but the audit step exited non-zero (e.g.
+ *                --fail-on-new tripped once baseline gating exists)
+ *   crashed    — no usable report (script crashed before writing one, or the
+ *                report itself records a fatal error)
+ */
+
+const fs = require('node:fs');
+const {buildA11ySection} = require('./lib/a11y-format');
+
+// GitHub issue bodies max out at 65536 characters — leave headroom for the
+// run link and footer the workflow appends after this summary.
+const MAX_SUMMARY_CHARS = 60000;
+
+const args = process.argv.slice(2);
+const getArg = name => {
+  const idx = args.indexOf(`--${name}`);
+  return idx !== -1 ? args[idx + 1] : null;
+};
+
+const reportFile = getArg('report') || 'a11y-weekly-report.json';
+const auditOutcome = getArg('audit-outcome') || 'success';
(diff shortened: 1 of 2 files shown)


cixzhang (approved):
Some of our storybooks were made jsut for debug purposes but we should likely clean those up. Generally in favor of having more regular scans.

PR #4384, opened 2026-07-26
fix(lab): a11y minors — Step, LogStream, SVGIcon, ChatReasoning

bhamodi:
## Summary

Four small accessibility fixes in `@astryxdesign/lab`, from a11y scan #3. Each fix addresses state or content that was visible on screen but unreachable (or actively hostile) for assistive technology:

- **Step (Stepper)** — WCAG 1.4.1 / 1.3.1: every step indicator (number badge, check, status glyphs) is `aria-hidden`, so completion/error state reached AT through nothing but color and shape.
- **LogStream** — WCAG 2.2.2 / 4.1.3: `role="log"` is implicitly `aria-live="polite"`, so a busy stream announced every appended row with no way to pause it.
- **SVGIcon** — WCAG 1.1.1: hardcoded `aria-hidden="true"` was written before the `{...props}` spread, so a consumer `aria-label` was silently swallowed unless they also reset `aria-hidden`.
- **ChatReasoning** — WCAG 1.3.1 / 4.1.2: collapsed content was hidden only visually (`grid-template-rows: 0fr` + `overflow: hidden`) and stayed in the a11y tree, and the `role="button"` trigger had `aria-expanded` but no `aria-controls`.

## Changes

- **Step** (`packages/lab/src/Stepper/Step.tsx`): progress/status now reaches AT as text. `error`/`warning` announce the semantic status, `success` and completed steps announce "completed"; the current step is already covered by `aria-current="step"` and not-started steps stay silent. The text renders as core `VisuallyHidden` next to the label and is composed into the accessible name of clickable steps (whose `aria-label` would otherwise override hidden text). Strings follow the existing lab convention of documented-default English (like the existing "Optional" affordance).
- **LogStream** (`packages/lab/src/LogStream/LogStream.tsx`): the log region's live behavior is now tied to the existing follow-pinning state — `aria-live="polite"` while following the tail, `aria-live="off"` while unfollowed (re-pinning via "Jump to latest" restores announcements). No new UI; documented in `LogStream.doc.mjs` and the component JSDoc.
- **SVGIcon** (`packages/lab/src/SVGIcon/SVGIcon.tsx`): mirrors core `Icon`'s label logic — when the consumer passes `aria-label`/`aria-labelledby`, the svg is exposed as `role="img"` instead of being hidden; otherwise it stays decorative (`aria-hidden="true"`). Spread before `{...props}` so an explicit consumer override still wins. (Includes a small prettier-mandated reformat of a pre-existing non-compliant union type in this file.)
- **ChatReasoning** (`packages/lab/src/ChatReasoning/ChatReasoning.tsx`): the collapsible region gets a `useId`-based id wired to the trigger via `aria-controls`, and `inert` while collapsed — this removes it from the a11y tree and tab order while keeping the element rendered, so the `grid-template-rows` expand/collapse transition is unaffected.

## Test plan

New tests were written first and confirmed failing pre-fix (11 failed across all 4 test files), then all pass post-fix:

- Step: hidden status text per state (completed/error/warning/success), silence for current/upcoming steps, status composed into clickable-step accessible names, on-track layout coverage. Two existing exact-name assertions updated for the new accessible names.
- LogStream: `aria-live` is `off` while unfollowed and `polite` while following, in both controlled and uncontrolled ("Jump to latest") flows.
- SVGIcon (new test file): decorative default (`aria-hidden`, no role), `role="img"` + no `aria-hidden` with consumer `aria-label`/`aria-labelledby`, explicit consumer `aria-hidden` still wins.
- ChatReasoning: trigger `aria-controls` points at the content region; region is `inert` when collapsed and not when expanded (round-trip).

Commands run (all green):

- `prettier --check` on the 9 changed files — pass
- `eslint` on changed files — pass (0 errors)
- `tsc --project packages/lab/tsconfig.json --noEmit` — pass
- Scoped: `vitest run packages/lab/src/Stepper packages/lab/src/LogStream packages/lab/src/SVGIcon packages/lab/src/ChatReasoning` — 4 files, 53 tests passed
- Full lab suite: `vitest run packages/lab` — 20 files, 265 tests passed

## Notes

- No changeset: `@astryxdesign/lab` is private/canary-only and `scripts/check-changesets.mjs` hard-rejects changesets for it ("private/ignored and cannot be released") — lab is changeset-exempt; CI stamps canary versions.

### DIFF
--- a/packages/lab/src/LogStream/LogStream.tsx
+++ b/packages/lab/src/LogStream/LogStream.tsx
@@ -329,6 +329,10 @@ const LEVEL_STYLE_TERMINAL: Record<LogStreamLevel, stylex.StyleXStyles> = {
  * polling; rows use `content-visibility: auto` for offscreen skip but are
  * NOT virtualized (window large streams in the caller).
  *
+ * Live announcements are tied to follow pinning: the `role="log"` region is
+ * `aria-live="polite"` only while following the tail and `aria-live="off"`
+ * while unfollowed, so appends never flood assistive tech.
+ *
  * @example
  * ```
  * <LogStream
@@ -538,6 +542,12 @@ export function LogStream({
         ref={scrollerRef}
         role="log"
         aria-label={label}
+        // role="log" is implicitly aria-live="polite" — on a busy stream that
+        // floods assistive tech with every appended row (WCAG 2.2.2 / 4.1.3).
+        // Follow-pinning doubles as the mute switch: while the user is away
+        // from the tail (unfollowed), the region is aria-live="off"; resuming
+        // follow ("Jump to latest") restores polite announcements.
+        aria-live={following ? 'polite' : 'off'}
         onScroll={handleScroll}
         {...stylex.props(styles.scroller(maxHeight ?? null))}>
         {entries.map(entry => (
--- a/.changeset/a11y-lab-minors-i18n.md
+++ b/.changeset/a11y-lab-minors-i18n.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] i18n: add `@astryx.step.*` catalog keys (`goToStep`, `goToStepWithStatus`, `status.completed`/`status.warning`/`status.error`) backing the lab Stepper's localized status text and clickable-step accessible names.
+@bhamodi
--- a/packages/core/locales/en.json
+++ b/packages/core/locales/en.json
@@ -879,6 +879,26 @@
     "defaultMessage": "Expand all rows",
     "description": "Screen-reader-only label on the header expand-all/collapse-all button when at least one row is collapsed (clicking expands every expandable row). Part of the four-way set."
   },
+  "@astryx.step.goToStep": {
+    "defaultMessage": "Go to step {stepNumber, number}: {label}",
+    "description": "Accessible name for a clickable step in the lab Stepper. {stepNumber} is the 1-based position, {label} is the step's visible label."
+  },
+  "@astryx.step.goToStepWithStatus": {
+    "defaultMessage": "Go to step {stepNumber, number}: {label}, {status}",
+    "description": "Accessible name for a clickable lab Stepper step that also carries a status. {status} is one of the translated `step.status.*` words (completed/warning/error). Keep the pattern parallel to `step.goToStep`."
+  },
+  "@astryx.step.status.completed": {
+    "defaultMessage": "completed",
+    "description": "Screen-reader-only status word rendered next to a lab Stepper step label when the step is done (either progress-completed or status=success). Lowercase adjective; it is read mid-sentence after the step label."
+  },
+  "@astryx.step.status.warning": {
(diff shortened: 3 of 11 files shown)


cixzhang (approved):

packages/lab/src/LogStream/LogStream.tsx
```diff
@@ -538,6 +542,12 @@ export function LogStream({
         ref={scrollerRef}
         role="log"
         aria-label={label}
+        // role="log" is implicitly aria-live="polite" — on a busy stream that
+        // floods assistive tech with every appended row (WCAG 2.2.2 / 4.1.3).
+        // Follow-pinning doubles as the mute switch: while the user is away
+        // from the tail (unfollowed), the region is aria-live="off"; resuming
+        // follow ("Jump to latest") restores polite announcements.
+        aria-live={following ? 'polite' : 'off'}
```
cixzhang, line 550: Should this use useAnnounce instead?

cixzhang (changes requested):
Hmm we're missing some translator and announce from core within lab. We should be able to sue them though.

packages/lab/src/LogStream/LogStream.tsx
bhamodi: I looked at swapping this to `useAnnounce` and I think the live region is the right mechanism here, for three reasons:

1. **LogStream has no dispatch path to announce from.** Entries arrive as props, so `useAnnounce` would have to fire from a render-diffing effect -- the exact pattern rejected for Toast in #4360 (StrictMode double-fires, announce-at-dispatch was the fix). Toast could move announcement into `addToast()`; LogStream has no equivalent imperative entry point.
2. **`role="log"` is the native pattern for exactly this.** AT computes and reads *additions* to the named log region itself; re-announcing every appended row through the global singleton region would double-speak for AT that already tracks `role="log"`, and would interleave arbitrary-volume log lines with toasts/status announcements in the shared region, detached from the log's accessible name.
3. The aria-live toggle is just the mute switch on that native region -- polite while following the tail, off while the user has scrolled away, so appends never flood AT (the actual bug this PR fixes).

The translator half of the review is addressed though: Step's status text and clickable-step names now go through core's `useTranslator` (`@astryx.step.*` keys added to the catalog with a core changeset) instead of hardcoded English. Happy to revisit LogStream if you're seeing a concrete AT problem with the region approach.

bhamodi:
@cixzhang Both halves addressed -- Step's status words and clickable-step accessible names now go through core's `useTranslator` (`@astryx.step.*` keys added to the core catalog, with a core changeset since `en.json` is a publishable core file). On LogStream I kept the `role="log"` live region rather than `useAnnounce` -- reasoning in the inline thread (short version: entries arrive as props so announce would need a render-diffing effect, the pattern #4360 rejected; and `role="log"` is the native mechanism AT already tracks for appends). Rebased onto current main, 53/53 tests green across the four components.

cixzhang (approved):

PR #4386, opened 2026-07-26
ci: fail pr-a11y on new axe violations against a checked-in baseline

bhamodi:
## Summary

The `pr-a11y` job has been report-only: `accessibility-audit.js` ran axe over the changed components' Storybook stories, wrote `a11y-report.json`, surfaced the results in a PR comment — and always exited 0. Nothing stopped an accessibility regression from merging. This PR turns the audit into a blocking gate against a checked-in baseline, so new axe violations fail CI while known ones stay visible-but-tolerated until fixed (evergreen a11y goal from accessibility scan #3).

## Changes

- **`.github/scripts/lib/a11y-baseline.js`** (new) — pure baseline-diff logic, kept free of Playwright/axe so it is unit-testable without a browser:
  - Violations are keyed `Component::Story::rule-id`. Node target selectors / HTML snippets are deliberately excluded from the key — they churn on unrelated markup changes and would silently invalidate baseline entries. Tradeoff (documented in the module): more nodes hitting an already-baselined rule in the same story won't fail; new rules, new stories, and new components always do.
  - Missing-from-baseline = new (a missing or empty baseline means everything is new). Baseline entries no longer reproduced are reported as "can be removed" but don't fail. Entries for components *outside* the audited set (CI scopes the audit to changed components) are treated as unchecked, not resolved.
  - Baseline regeneration is merge-aware: a `--components`-scoped run replaces only the audited components' entries and preserves the rest.
- **`.github/scripts/accessibility-audit.js`** — new `--baseline <path>`, `--fail-on-new`, and `--update-baseline` flags. After writing the report, the script diffs it against the baseline, prints a readable summary (rule, impact, component/story, local repro command, how to baseline an intentional exception), and exits 1 on new violations when `--fail-on-new` is set. All other behavior is unchanged; without `--baseline` the script behaves exactly as before.
- **`.github/a11y-baseline.json`** (new) — ships **empty**: Playwright/Chromium could not be run in the authoring environment, so no current violations could be verified. Missing-from-baseline counts as new, so the gate is fully armed from day one; any pre-existing violations that surface on component PRs should be either fixed or intentionally baselined via the documented command.
- **`.github/workflows/ci.yml`** — the `pr-a11y` audit step now passes `--baseline .github/a11y-baseline.json --fail-on-new`; the report artifact upload gets `if: always()` so the artifact and the PR comment keep working when the gate fails. No trigger/permission changes.
- **`package.json`** — `pnpm a11y:audit` (gate a local run) and `pnpm a11y:baseline` (regenerate the baseline) scripts.
- **`CONTRIBUTING.md`** — new "Accessibility audits" section under Testing: how the gate works, the local repro/baseline workflow, and the axe-coverage caveat.

## Test plan

- `vitest run .github/scripts/lib/a11y-baseline.test.mjs` — 13 tests passing, covering: new violation fails, baselined violation passes, resolved entries flagged as removable, unaudited components not misreported as resolved, key stability across DOM churn (selector/markup changes don't change keys), aggregated-report fallback, bare-string baseline entries, merge-aware scoped regeneration, round-trip (baseline built from a report gates that report clean), and failure-summary content.
- `node --check` on `accessibility-audit.js` and `lib/a11y-baseline.js`.
- `ci.yml` validated with a YAML parse (js-yaml).
- CLI dry-run of the new flags (no browser needed via the storybook-missing path): `--baseline --fail-on-new` exits 0 with a "0 new" gate summary; `--update-baseline` rewrites the baseline and preserves entries for components outside the run.
- Prettier-clean on all new files plus the touched `package.json` / `CONTRIBUTING.md` (`accessibility-audit.js` and `ci.yml` were not prettier-formatted before this PR; edits match their existing style rather than reformatting them wholesale).
- Note: this PR touches no components, so `pr-a11y` is skipped on this PR itself (`check-components` gates it). The gate gets its first real exercise on the component PRs in this a11y sweep once this merges.

## Notes

- **Baseline starts empty.** Regenerate/extend it locally with:
  ```bash
  pnpm storybook:build && npx playwright install chromium
  pnpm a11y:baseline -- --components <Comma,Separated,Names>
  ```
- **axe covers only a subset of WCAG** (roughly a third of success criteria are automatable). A green `pr-a11y` is a floor, not a certification — keyboard flows, focus order, and screen-reader semantics still need manual checks (also called out in the new CONTRIBUTING section).
- No changeset: this is CI/scripts-only — `scripts/check-changesets.mjs` validates the changesets that exist rather than requiring one per PR, and no publishable package changes here.
- The Vercel deployment failure on fork PRs is known infra and unrelated.

### DIFF
--- a/.github/a11y-baseline.json
+++ b/.github/a11y-baseline.json
@@ -1,0 +1,6 @@
+{
+  "$comment": "Known axe violations tolerated by the pr-a11y CI gate. Entries are keyed Component::Story::rule-id (see .github/scripts/lib/a11y-baseline.js). Regenerate with `pnpm a11y:baseline` after `pnpm storybook:build` and `npx playwright install chromium`. Remove entries as violations are fixed; anything not listed here fails pr-a11y.",
+  "version": 1,
+  "generatedAt": null,
+  "entries": []
+}
--- a/.github/scripts/accessibility-audit.js
+++ b/.github/scripts/accessibility-audit.js
@@ -5,7 +5,13 @@
 /**
  * @description Runs accessibility audits on component stories using axe-core
  * @input --storybook-dir <path> --output <file> --components <comma-separated>
- * @output JSON report with accessibility violations
+ *   --baseline <path> (compare violations against a checked-in baseline)
+ *   --fail-on-new (exit 1 when violations not present in the baseline exist)
+ *   --update-baseline (rewrite the baseline file from this run's report)
+ * @output JSON report with accessibility violations; with --baseline, a gate
+ *   summary (new / baselined / resolved) on stdout and a non-zero exit code
+ *   when --fail-on-new finds regressions. Diff logic lives in
+ *   lib/a11y-baseline.js.
  */
 
 const { chromium } = require('playwright');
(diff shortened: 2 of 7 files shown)


cixzhang (approved):
We'll try this.

PR #4387, opened 2026-07-26
lint: flag hardcoded strings passed to announce()

bhamodi:
## Summary

Evergreen a11y/i18n scan #3 found 8 hardcoded-English strings reaching users through `announce()` live-region calls and aria-labels that `@astryx/no-hardcoded-i18n-string` did not catch: the rule covered JSX attributes, object properties, and destructure defaults, but not **call arguments**. The individual sites are being fixed in parallel sweep PRs; this PR closes the class so it cannot be reintroduced.

## Changes

- **`internal/eslint-plugin-astryx/no-hardcoded-i18n-string.js`** — new call-argument surface with a new `hardcodedCallArg` message:
  - Flags the **first argument** of calls to configured callees (option `callees`, default `['announce']` — the live-region announcer returned by `useAnnounce()`).
  - Only plain string literals and **expression-free** template literals are flagged, and only when they pass the existing `looksUserFacing()` heuristics (so `announce('')` clears and identifier-shaped strings stay allowed). `t(...)` results, variables, and templates with interpolations are allowed.
  - Same file ignores as the rest of the rule (tests, stories, `.doc.mjs`).
  - New option `allowedCalleeStrings: string[]` — exact-match, file/line-agnostic exemptions for grandfathered sites.
- **`eslint.config.js`** — passes options to the rule in the existing `packages/core/src` scope at the existing two-tier severity (`error` in strict/CI, `warn` locally), with a **temporary allowlist** for the three string-literal sites that exist at HEAD:
  - `'Copied'` (CodeBlock), `'Selection cleared'` and `'All selected'` (MultiSelector)
  - These are being replaced with `t(...)` in the scan #3 i18n sweep PRs. Follow-up: remove each entry as its fix merges, then delete `allowedCalleeStrings` entirely — the allowlist is self-cleaning and should reach zero.
- **`internal/eslint-plugin-astryx/no-hardcoded-i18n-string.test.mjs`** — new RuleTester suite (the rule previously had no tests).

No changeset: internal lint tooling only (`@astryx/eslint-plugin` is private and root `eslint.config.js` is not a published package), so no published package version changes.

## Test plan

- `vitest run internal/eslint-plugin-astryx/no-hardcoded-i18n-string.test.mjs` — 18/18 pass. Cases include: flags `announce('Copied')` and expression-free templates; allows `announce(t('key'))`, `announce(variable)`, templates with expressions, empty string, non-configured callees, member calls; allowlist exact-match behavior; custom `callees` option; test/story filename ignores; existing JSX-attribute surface still behaves.
- `ASTRYX_STRICT_LINT=1 eslint packages/core/src` — **exit 0** on this branch (allowlist working).
- With the allowlist temporarily emptied, strict eslint reports exactly the 3 known sites (CodeBlock.tsx:762, MultiSelector.tsx:701, MultiSelector.tsx:703) and nothing else — no false positives across core.
- `ASTRYX_STRICT_LINT=1 eslint eslint.config.js` — clean.
- `prettier --check` passes on the new test file; the two edited `.js` files were already not prettier-clean at HEAD (repo only enforces prettier on ts/tsx/md), so they were left in their existing style rather than mass-reformatted.

## Notes

- **Callee-name matching limitation**: the rule matches calls by identifier name only (`announce(...)`) with no import/scope tracing. Any local function named `announce` is checked; a renamed destructure (`const {announce: say} = useAnnounce()`) escapes it. This is a deliberate simplicity tradeoff documented in the rule header — the codebase consistently uses the conventional name.
- Interpolated announcements (e.g. `` announce(`Added ${label}`) ``) are intentionally NOT flagged by this conservative check to avoid false positives; those sites are handled directly by the i18n sweep.
- Vercel fork-PR deployment failure is known infra and unrelated.

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -180,6 +180,16 @@ export default defineConfig(
           'Carousel/Carousel',
         ],
       }],
+      // announce() live-region messages are user-facing text; the rule checks
+      // them as call arguments (callees defaults to ['announce']).
+      // TEMPORARY allowlist: these exact strings predate the check and are
+      // being replaced with t(...) in the scan #3 i18n sweep PRs
+      // (CodeBlock 'Copied'; MultiSelector 'Selection cleared' /
+      // 'All selected'). Remove each entry as its fix merges; delete
+      // allowedCalleeStrings entirely once the sweep lands.
+      '@astryx/no-hardcoded-i18n-string': [isStrictMode ? 'error' : 'warn', {
+        allowedCalleeStrings: ['Copied', 'Selection cleared', 'All selected'],
+      }],
     },
   },
   // The i18n runtime itself defines the message strings the rest of the
--- a/internal/eslint-plugin-astryx/no-hardcoded-i18n-string.js
+++ b/internal/eslint-plugin-astryx/no-hardcoded-i18n-string.js
@@ -24,6 +24,16 @@
  *      NOT flagged.
  *   2. Object property assignments of the same names inside .tsx files.
  *   3. Destructure defaults (`function Foo({label = 'X'})`) of the same names.
+ *   4. Call arguments: hardcoded first arguments to configured callees
+ *      (default: `announce`, the live-region announcer returned by
+ *      `useAnnounce()`). Matches by CALLEE NAME only — no import/scope
+ *      tracing — so any local function named `announce` is checked. That is
+ *      a deliberate simplicity tradeoff: false positives require shadowing
+ *      the hook's conventional name, which the codebase does not do.
+ *      Only plain string literals and template literals WITHOUT expressions
+ *      are flagged; templates with interpolations, identifiers, and t(...)
+ *      results are allowed (interpolated announcements are caught by the
+ *      i18n sweep, not this conservative check).
  *
  * Ignores:
  *   - Test files (*.test.*, __tests__/**)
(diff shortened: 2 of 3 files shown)


humbertovirtudes (approved):
Nice close on this. The rule already covered JSX attributes, object properties, and destructure defaults, but call arguments were the obvious gap and `announce()` is exactly the right sink to guard. Extending the existing rule rather than adding a parallel one keeps the surface small, and adding the first RuleTester suite for a rule that had none is a real improvement.

Triage: internal lint tooling, non-breaking, low blast radius -> fast path.

- `vitest run` on the new suite: 18/18 pass.
- `ASTRYX_STRICT_LINT=1 eslint packages/core/src`: exit 0 with the allowlist.
- Emptied `allowedCalleeStrings` and re-ran strict lint: exactly 3 problems, and exactly the three grandfathered sites (CodeBlock.tsx:762 `'Copied'`, MultiSelector.tsx:701 `'Selection cleared'`, :703 `'All selected'`), zero false positives across core. The central claim holds.
- Confirmed all three allowlisted strings map to real `announce(...)` calls at those exact lines.

What I like:
- The conservative definition of "hardcoded" (plain string literals + expression-free templates only) is the right call: `announce(t(...))`, variables, and interpolated templates all stay allowed, so the interpolation class is left to the i18n sweep instead of generating false positives here.
- Empty-string clearing (`announce('')`) and identifier-shaped strings stay allowed via the existing `looksUserFacing()` reuse.
- The callee-name-only matching limitation and the self-cleaning allowlist are both documented in the rule header and the config comment, so the follow-up path (drop each entry as its sweep PR merges, then delete `allowedCalleeStrings`) is clear.

Non-blocking nits (fine to land as-is):
- The allowlist is file/line-agnostic exact-match, so once a real `announce('Copied')` is localized, any other future `'Copied'` would still be silently exempted until the entry is removed. Given the self-cleaning follow-up plan and the small window, that is an acceptable tradeoff, not a change request.
- `allowedCalleeStrings` is global rather than per-callee. Not needed for the current single-callee setup; worth revisiting only if a second callee is ever configured with colliding strings.

No changeset is correct here: `@astryx/eslint-plugin` is private and root `eslint.config.js` is not a published package, so no consumer-visible version change.

---
PR #4387 - lint: flag hardcoded strings passed to announce()
Author: @bhamodi
What it does: extends `@astryx/no-hardcoded-i18n-string` to flag hardcoded user-facing strings passed as the first argument to `announce()`, closing the call-argument gap; adds the rule's first test suite and a temporary self-cleaning allowlist for three known sites.
Recommendation: Approve.

cixzhang (approved):

PR #4388, opened 2026-07-26
fix(MobileNav): closing the drawer could leave the page unresponsive

### LINKED ISSUE #4290: [Bug] MobileNav close freezes the page in Safari 26.1
### Description

Closing the automatic `AppShell` mobile navigation caused Safari 26.1 to become completely unresponsive. The drawer opened normally, but clicking the close button (X) blocked further interaction with the page. No JavaScript error was reported.

The same behavior was observable on the official Astryx website in Safari 26.1, so it did not appear to be specific to the consuming application. Chrome worked correctly throughout.

After updating Safari from 26.1 to 26.5.2, both the consuming application and the official Astryx website work correctly. An automated Safari WebDriver regression check covering open → close → reopen also passes on 26.5.2.

Expected behavior: MobileNav closes and the page remains interactive on supported Safari versions.

Actual behavior on Safari 26.1: clicking the close button makes the page unresponsive, with no console error.

This is different from #3091 / #3239: the issue is not merely that the drawer cannot be reopened. Safari itself becomes unresponsive during the close interaction, and the application is using Astryx 0.1.8, which already contains that earlier lifecycle fix.

Since the Astryx browser-support documentation lists Safari 26+ as Tier 1, it may be useful to document this affected Safari build, add a browser regression test if feasible, or clarify whether a newer Safari 26 patch release is required.

### Reproduction

1. On Safari 26.1, render an `AppShell` using the automatic responsive MobileNav and a `SideNav` (`mobileNav={{ breakpoint: "md" }}`).
2. Enter a mobile viewport (tested at 767 × 900).
3. Click the Open navigation button.
4. Click the Close navigation button (X).
5. Observe that Safari becomes unresponsive and the page cannot be interacted with.

Comparison:

- Safari 26.1 (`21622.2.11.11.9`): reproducible.
- Safari 26.5.2 (`21624.2.5.11.8`): open → close → reopen passes.
- Chrome: works correctly.
- Official Astryx website: reproduced on Safari 26.1 and works after updating to Safari 26.5.2.

Additional isolation performed:

- A minimal native `<dialog>` open/close flow passed in Safari 26.1.
- Detached Astryx dialog markup and CSS without the React lifecycle also passed.
- The hang occurred with the React-controlled Astryx MobileNav lifecycle.

### Astryx Version

`@astryxdesign/core@0.1.8`

### Environment

Affected environment:

- Safari 26.1 (`21622.2.11.11.9`)
- macOS 26.1
- React 19.2.8
- React DOM 19.2.8

Verified working after update:

- Safari 26.5.2 (`21624.2.5.11.8`)
- macOS 26.5.2

AKnassa:
## What this does

Closing the mobile navigation drawer could leave the entire page unresponsive. The page looked completely normal — nothing was covering it — but no click, tap or scroll did anything, and no error appeared anywhere to explain it. This fixes the cause.

Fixes #4290.

## Why

Someone hit this on Safari 26.1 and it reproduced on the Astryx site itself. Apple fixed their side in a later Safari release, so it would be easy to close this as "a browser bug, already gone".

It isn't quite that. The drawer was asking the browser to do something it was never meant to cope with, and it happened to work everywhere else. Safari 26.1 was the browser that didn't paper over it. Anything relying on luck that specific is worth removing, and the same luck is what every future browser would have to keep providing.

## What was going wrong

The drawer is a native pop-up panel. While it is open the browser deliberately blocks the rest of the page, so you can't click things behind it — that part is correct and wanted.

The problem was the order of two steps. The drawer was **hidden first** and **told to close a moment later**. In that gap the page was still blocked by a panel that nobody could see. Most browsers clean that up when the panel finally closes; Safari 26.1 didn't, and the block was never lifted — so the page stayed frozen, with nothing on screen to explain why.

That also explains why the reporter's own tests all passed: none of them recreated that specific gap.

## What changed

- **The drawer now stays visible until it has genuinely finished closing.** The page can never be blocked by something invisible.
- **The closing animation actually plays now.** It had been written into the component a long time ago but could never run — an earlier fix accidentally cancelled it every time.
- **The closing time is read from the active theme instead of being a fixed number.** This one matters: the shipped **y2k** theme happens to use exactly the same number the code assumed, which put the fix on a knife edge, and the "Snappy" theme example in our own docs does the same. Any theme slightly faster than that would have brought the original problem straight back, silently.
- **With "reduce motion" switched on, the drawer now closes sooner rather than shrinking its safety margin.** Previously that accessibility setting left no margin at all — one slow frame on a phone and the original fault returned.

No API changed. Nothing consumers write needs to change.

## How to see it

In Storybook, open **Core / MobileNav → Default**, or an AppShell below its mobile breakpoint, and open and close the drawer. Two visible differences: the drawer now slides out instead of vanishing, and the backdrop fades with it.

The freeze itself needs Safari 26.1 to observe directly. What I could measure is the faulty step, in ordinary Chrome, by recording the drawer's state at the exact moment it is told to close:

| | at the moment the drawer is told to close |
|---|---|
| before | hidden, but still blocking the page |
| after | still visible, as it should be |

I also confirmed separately that a hidden-but-still-open panel really does make the whole page unclickable: a real click on a visible button did nothing, and the same click worked again immediately after the panel closed properly.

## Tests

31 tests covering the close path, including the cases most likely to break it later: React StrictMode, unmounting mid-close, changing the drawer side while it closes, rapid open/close, page-scroll restoration, leftover timers, and closing via Escape or the backdrop.

I also checked the tests actually bite by deliberately breaking the fix nine different ways — all nine were caught.

One of those nine was the tests themselves. The usual way to fake `matchMedia` answers every question with a single yes, which puts the shell below its breakpoint as intended but also quietly says "this user wants reduced motion" — and that shortens the very delay most of these tests exist to measure. They passed while measuring nothing. The stub now answers each question on its own, and it lives in the shared test folder so the next person reaches for the right one.

## Not included, on purpose

Three things are judgement calls for a maintainer rather than code I should decide:

- **Documenting the affected Safari build.** It would fit the existing browser-support doc without new structure, but that doc's stated position is that tiers are rolling and readers should feature-detect rather than pin versions. Recording a specific bad build cuts against it.
- **An automated browser test for this.** Not currently possible here, and I'd rather say so than add a test that looks like protection. CI runs Chromium only, there is no macOS runner, and Playwright's WebKit is not Safari — it carries no Safari version, so it cannot express the 26.1 vs 26.5.2 difference this bug lives in.
- **Putting a floor under the theme value itself.** Reading the closing time from the active theme fixes the drift for every theme we ship. It does not cover a theme that sets that time to zero, which nothing currently prevents: with no closing time there is no window for the fix to land inside, and in the AppShell drawer there is no timer to fall back on either. Guarding it inside this one component was tried and reverted — the CSS is valid and works in a real browser, but our test environment discards it, so the guard would have cost a working test to defend a case no shipped theme can reach. The right place is where themes are defined, which is one change covering every component instead of one per component. Filed separately as #4664, which also records the same hazard sitting in `lab/Drawer` today.

## One thing worth knowing

In the AppShell auto-drawer, React tears the drawer down immediately on close, so the slide-out animation still doesn't play there — only the standalone and explicit forms animate. That path is protected by the same fix, just by the styling rather than the timing. Happy to make it consistent if you'd prefer.


imdreamrunner (dismissed):
(accidental posted by agent)

rubyycheung (approved):

packages/core/src/MobileNav/MobileNavCloseVisibility.test.tsx
```diff
@@ -0,0 +1,172 @@
...
+
+describe('MobileNav stays rendered while it closes', () => {
+  it('transitions display with allow-discrete so it hides only once closed', () => {
+    render(<Drawer isOpen />);
+    const style = getComputedStyle(screen.getByTestId('mobile-nav'));
+
+    // Without these two the drawer blanks out the instant `isOpen` flips,
+    // leaving an open modal dialog that nothing renders — and nothing can
+    // click past. See the file header.
+    expect(style.transitionProperty).toContain('display');
```
cixzhang: This pins the declaration; nothing pins that `display` is still `flex` when `close()` runs. Chromium can assert it.

cixzhang (dismissed):
Thanks — nice catch, and the Chromium probe is the right call since jsdom can't see this. Merging.

cixzhang (approved):

cixzhang (approved):
Thank you — reading the hold from the theme instead of assuming 250ms is the part that keeps this fixed. Chromium guard added on top.

[Reviewed by Robohands]

PR #4389, opened 2026-07-26
test: scope 'No results found' queries away from the singleton live regions

bhamodi:
## Summary

Five unrelated PRs in the current a11y sweep (#4355, #4358, #4360, #4365, #4367) failed CI's `test` job on the identical assertion: `MultiSelector > shows empty state when search has no results` -> `Found multiple elements with the text: No results found`.

Root cause: `useAnnounce`'s singleton live regions duplicate user-visible strings at `document.body` level. `BaseTypeahead` announces `emptySearchResultsText` (default `'No results found'`) into the polite region while the same string renders in the visible empty state, so an unscoped `getByText('No results found')` races the announcement. This is the same flake class fixed for Calendar month labels in #4362, and it also explains the intermittent `Typeahead` paste-test failures seen in CI.

## Changes

- `Typeahead.test.tsx`, `MultiSelector.test.tsx`: a `visibleNoResults()` helper filters out matches inside `[data-astryx-live-region]`; the 5 racy queries (4 in Typeahead incl. two negative assertions that could false-positive on stale region text, 1 in MultiSelector) now use it. Test-only change; no component code touched.

## Test plan

- `vitest run packages/core/src/Typeahead packages/core/src/MultiSelector` — 2 files, 106/106 pass, run 3x consecutively with no flakes
- eslint + prettier clean on both files

## Notes

- Unblocks the failing sweep PRs once merged (their CI runs against the PR merge ref, which includes main).
- Vercel deploy failure on fork PRs is known infra, unrelated.

### DIFF
--- a/packages/core/src/MultiSelector/MultiSelector.test.tsx
+++ b/packages/core/src/MultiSelector/MultiSelector.test.tsx
@@ -15,6 +15,15 @@ import userEvent from '@testing-library/user-event';
 import {MultiSelector} from './MultiSelector';
 import {__resetLiveRegionsForTest} from '../hooks/useAnnounce';
 
+// The useAnnounce singleton live regions duplicate user-visible strings like
+// 'No results found' at document.body level, racing text queries (the same
+// class of flake fixed for Calendar month labels). Query the visible copy only.
+function visibleNoResults() {
+  return screen
+    .queryAllByText('No results found')
+    .filter(el => el.closest('[data-astryx-live-region]') === null);
+}
+
 // Module-level constants to satisfy @eslint-react/no-unstable-default-props.
 const ANNOUNCE_OPTIONS = ['Apple', 'Banana', 'Orange'] as const;
 const EMPTY_VALUE: string[] = [];
@@ -634,7 +643,7 @@ describe('MultiSelector', () => {
     const searchInput = screen.getByRole('combobox', h);
     await user.type(searchInput, 'xyz');
 
-    expect(screen.getByText('No results found')).toBeInTheDocument();
+    expect(visibleNoResults()).toHaveLength(1);
   });
 
   describe('result announcements', () => {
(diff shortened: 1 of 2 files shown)


humbertovirtudes:
Thanks for chasing this down; the root cause (the `useAnnounce` singleton live regions duplicating user-visible strings at `document.body` level) is exactly right, and the `visibleNoResults()` helper is a clean way to query the visible copy. Locally on this branch the suite is solid: `vitest run packages/core/src/Typeahead packages/core/src/MultiSelector` gives 106/106 pass.

One thing to fix before this can land: the branch is ~89 commits behind `main`, and the **MultiSelector** hunk now conflicts with `main`. Main already deflakes that same `'No results found'` assertion, but with a different technique: it scopes the query to the listbox (`within(screen.getByRole('listbox', h)).getByText('No results found')`) rather than filtering out the live region. That is the empty-state scoping introduced in #4089's pattern, and the MultiSelector empty-state assertion on `main` already carries it. So the MultiSelector change here is redundant and produces a real merge conflict in `MultiSelector.test.tsx`.

Could you please:
1. Rebase on `main`.
2. Drop the MultiSelector hunk entirely (the empty-state assertion is already scoped there).
3. Keep only the **Typeahead** changes: the four `'No results found'` queries in `Typeahead.test.tsx` are still unscoped on `main`, so that is the genuine remaining delta this PR should carry.

After the rebase, please confirm the two negative assertions still behave (they should now assert `toHaveLength(0)` against the visible-only helper) and re-run the Typeahead suite a few times to confirm the flake is gone. Once it is Typeahead-only and merges cleanly, I am happy to approve.

cixzhang:
Thanks for this, and apologies for the overlap — the MultiSelector half of this already landed on main via the accessibility sweep in #4427 (shipped in v0.1.9). Main now scopes the empty-state query with `within(screen.getByRole('listbox')).getByText('No results found')`, which is exactly the fix here, so this PR is superseded for that component and has gone conflicting against the newer base.

I'm going to close it as superseded. One thing worth salvaging: if you're still seeing the brittle-match flake on the **Typeahead** empty state specifically (that part wasn't covered by the sweep), a small fresh PR based on current main that scopes just those Typeahead queries would be very welcome — that piece still has value.

Really appreciate you pushing on test robustness here. Thank you! 🙏

Issue #4186, opened 2026-07-22
NumberInput: use text input, handle invalid/pending values, and match input-family conventions (align with XDSNumberInput)


cixzhang:
## Summary

`NumberInput` needs a refinement pass to bring it in line with the internal `XDSNumberInput` behavior and the rest of the Astryx input-family conventions. Three related problems:

## 1. Use `type="text"` (with numeric input mode), not `type="number"`

Today the internal `<input>` is `type="number"` (`NumberInput.tsx`). The code itself already documents the pain — there's a comment: *"With `type='number'`, we can't use formatted display values."* Native number inputs are annoying to type into (scroll-wheel value changes, locale/decimal quirks, no formatted display, inconsistent cross-browser behavior).

Match the internal `XDSNumberInput` approach: render a **`type="text"`** input with an appropriate `inputMode` (e.g. `numeric`/`decimal`) and handle numeric parsing/validation in the component. This unlocks formatted display values and a much less annoying typing experience.

## 2. Account for invalid values by preserving pending input separate from the true value

There's already the seed of this — `pendingInput` and an `isInputValid` memo exist — but the behavior should match the internal component: **when the input is invalid, mute the input visually while preserving what the user typed, kept separate from the committed/true numeric value.** The user's in-progress text shouldn't be destroyed or silently coerced; the true value stays last-valid until a valid value is committed. Formalize the "display/pending value vs. true value" split so invalid intermediate states are a first-class, visible state rather than an edge case.

## 3. Match input-family conventions it's currently missing

`NumberInput` is missing conventions the rest of the input family (and `XDSNumberInput`) provide. Audit against the family and add the missing ones, e.g.:
- **`hasSpinner`** — increment/decrement stepper affordance.
- **`changeAction`/`onChangeAction`** — the async action-style change callback convention used across inputs.
- Any other standard input props it lacks (review against `TextInput`/the input-consistency checklist): `startIcon`, status handling, `isLoading`/`isBusy`, `isOptional`/`isRequired`, `isLabelHidden`, `description`, etc.

The goal is that `NumberInput` behaves consistently with both `XDSNumberInput` internally and Astryx's own input-family contract.

## Acceptance criteria

- [ ] Input renders as `type="text"` with a numeric `inputMode`; formatted display values work; typing is not degraded by native number-input behavior.
- [ ] Invalid/pending input is preserved separately from the true value, and invalid state is shown by muting the input (matching internal behavior).
[...]

PR #4390, opened 2026-07-26
feat(NumberInput): add the changeAction convention (partial #4186)

AKnassa:
> **Read this first: this is one of the five boxes in #4186, not all of them.** It is opened as a draft to put a concrete diff next to two questions that need your call, not as a claim on the issue. `Refs`, not `Closes`.

## What is NOT in here, and why

**1. `type="number"` → `type="text"` with a numeric `inputMode`.** The edit itself is two lines, but `type="number"` carries implicit **spinbutton semantics** — a screen reader announces the field as a spin button with min/max/now. This component's own tests query `screen.getByRole('spinbutton')`, so that is load-bearing today. Switching to `type="text"` silently drops it unless the component takes on `role="spinbutton"` plus `aria-valuemin` / `aria-valuemax` / `aria-valuenow` / `aria-valuetext` and keeps them in sync with the pending-vs-committed split. The issue does not say which way to go, and picking one is an accessibility decision rather than a fix.

**→ Question: should `NumberInput` declare `role="spinbutton"` and own the `aria-value*` set, or is losing the native spinbutton announcement acceptable in exchange for formatted display values?**

**2. `hasSpinner`.** `hasSpinner` does not appear anywhere in this repo, so there is no sibling to copy. Stepper buttons raise their own questions: markup and hit targets, where they sit relative to the existing `units` / clear / status-icon slots, whether they are focusable or `aria-hidden` with the arrow keys as the real affordance, and how they behave at min/max.

**→ Question: what shape should `hasSpinner` take? Happy to build it once the anatomy is settled.**

**3. Reconciling against `XDSNumberInput`.** That component is not in this repository, so any claim that this "matches the internal behavior" would be guesswork. Someone with access needs to diff them.

**4. Preserving invalid input** is largely already there — `pendingInput`, the `isInputValid` memo, and the muted invalid styling. The issue calls this "the seed of this", and this PR does not touch it.

## What IS in here

`changeAction` — the async-action convention **every sibling input already has**: `TextInput`, `TimeInput`, `Selector`, `CheckboxInput`, `CheckboxList`, `Pagination`. `NumberInput` had nothing, so a consumer driving it from a server action had no transition-aware channel at all. This is item 3's "other missing input-family conventions", and nothing about it is invented — the name comes from the issue itself.

- Fires after `onChange`, inside a React transition, matching the siblings
- Typed off the same `hasClear` discrimination as `onChange`, so it only receives `null` where clearing is enabled
- Wired into every commit path: valid keystroke, blur, Enter, and the clear button
- The field now holds the committed number **optimistically**, so a controlled parent that applies the value asynchronously no longer makes the input snap back to the stale number the moment blur clears the pending text

One implementation note for review: `changeAction` fires from a small helper rather than being folded into a shared change function. `onChange`'s type only narrows inside the `hasClear` branches, and a single shared call site would erase that narrowing.

## Verification

- `NumberInput` 72/72 (5 new), full suite 7674/7674 across 378 files
- `tsc` clean, eslint clean, prettier clean, `check:repo` 5/5
- Three mutations, each killed by exactly one test. Worth flagging: the optimistic-display test **survived its mutation on the first pass** — it asserted while `pendingInput` was still set, so the display never reached the value pipeline and the optimistic state was never exercised. Rewritten to blur first, which is the only moment that state is load-bearing; the mutation dies now.

Refs #4186


cixzhang (commented):
Thanks for adding `changeAction` here — the core wiring is right and I checked it against the rest of the input family. Two notes: one confirming you picked the correct sub-convention, one gap to close.

**✅ You followed the right precedent (Selector, not TextInput).** There are two established `changeAction` sub-conventions, split by whether the input's `onChange` carries a DOM event:
- *Event-carrying* (TextInput, CheckboxInput, Switch): `changeAction(value, e)`, and the async action is gated on `!e.defaultPrevented` so a consumer can cancel it from `onChange`.
- *Event-less* (Selector): `changeAction(value)`, no event, no `defaultPrevented` gate.

NumberInput's `onChange` is `(value: number)` with no event, so it belongs to the event-less group — and this PR matches Selector exactly: value-only signature, the same discriminated-union widening to `number | null` under `hasClear`, `onChange` fired before `changeAction`, and `startTransition` + `useOptimistic`. Firing only on valid parsed values is also consistent with NumberInput's documented `onChange` contract. All correct.

**⚠️ Gap: the optimistic-pending state isn't observable.** The rest of the family treats the in-flight async action as a *visible* state, not just an optimistic value swap. Both TextInput and Selector derive:
```
const isBusy = isLoading || optimisticValue !== value;
```
and surface it — TextInput renders a `<Spinner size="sm" />` and sets `aria-busy={isBusy || undefined}` on the control; Selector exposes the same `isBusy`. This PR routes `optimisticValue` into the display value but drops the transition's pending signal (`const [, startTransition]`) and never derives `isBusy`, so a NumberInput with an async `changeAction` shows the optimistic number with **no loading affordance and no `aria-busy`** while the action is pending — inconsistent with every sibling, and an a11y regression relative to the convention.

To match: add an `isLoading?: boolean` prop (NumberInput currently has none, while TextInput does), derive `isBusy = isLoading || optimisticValue !== value`, set `aria-busy={isBusy || undefined}` on the input, and render the same `<Spinner size="sm" />` treatment. That closes the behavior gap so NumberInput's `changeAction` looks and announces the same as TextInput's.

Happy to point at the exact TextInput lines (isBusy at ~324, Spinner/aria-busy at ~456/469) if useful.

AKnassa:
Done in faf7f87a7.

Added `isLoading` and the family's busy derivation, with `aria-busy` on the input and `<Spinner size="sm" />`. Busy announces without locking, so the field stays editable like TextInput and TimeInput.

Two deliberate deviations:

`const [, startTransition]` stays as is. That discard is the family standard (TextInput:313, TextArea:326, TimeInput:370, Selector:627, DateInput:341), and busy comes from the optimistic mismatch rather than `isPending`.

I used `Object.is` rather than `!==`. NumberInput is the only input in the family whose value is a raw number, and `NaN` is a legal `number` that is not equal to itself, so `!==` pinned an idle field to busy permanently.

I also fixed the commit guards while I was in here. They compared against the `value` prop, which is stale while an action is in flight: blur re-dispatched a number already on its way, and an edit back to the pre-flight number was dropped silently, reaching neither `onChange` nor `changeAction`. They now guard on `optimisticValue`, which equals `value` whenever nothing is pending.

Three things I left alone, because they are family wide rather than specific to this PR:

- the Spinner's `role="status"` region has no text content, and `aria-busy` on a spinbutton is not surfaced by most screen readers, so the busy state is effectively visual only
- `isBusy` never fires when the parent applies `onChange` synchronously, since the optimistic value never diverges. TextInput behaves identically
- the clear button is gated on `value` while the display derives from `optimisticValue`, so it is briefly inert mid flight. TextInput:421 vs :449 and TimeInput ship the same asymmetry

Happy to take any of those on separately if you want them changed.

Also merged main, which conflicted with #4187 and #4355 in the test file.

packages/core/src/NumberInput/NumberInput.tsx
```diff
@@ -316,6 +324,8 @@ interface NumberInputPropsBase extends Omit<
 type NumberInputPropsNonClearable = NumberInputPropsBase & {
   hasClear?: false;
   onChange: (value: number) => void;
+  /** Async action on change. Fires after onChange. */
+  changeAction?: (value: number) => void | Promise<void>;
```
cixzhang, line 328: Splitting this across the union is what breaks the `{...props}` spreads. Probably wants one declaration on the base.

packages/core/src/NumberInput/NumberInput.tsx
```diff
@@ -697,6 +791,7 @@ export function NumberInput({
       <VisuallyHidden as="div" role="alert" aria-live="assertive">
         {!isInputValid ? 'Invalid number' : ''}
       </VisuallyHidden>
+      {isBusy && <Spinner size="sm" />}
```
cixzhang, line 794: Hmm, TextInput puts the spinner after the clear button (`TextInput.tsx:462`). Might want to match.

cixzhang (changes requested):
Thanks — the busy state came back exactly right; I drove the whole flow in Chrome and it behaves.

CI is red on two stories this PR doesn't touch (`NumberInput.stories.tsx:538`, `:549`): splitting `changeAction` across the `hasClear` union stops `<NumberInput {...props} hasClear />` compiling. One declaration on the base clears it:

```ts
changeAction?: (value: number | null) => void | Promise<void>;
```

Selector has the same hole at `Selector.tsx:691` — that one's ours, not yours.

Then a merge: [#4896](https://github.com/facebook/astryx/pull/4896) landed steppers, `role="spinbutton"` and formatted display, so both your open questions are answered. It also adds `stepValue`, a fifth commit path — route that one too or `changeAction` fires for typing and blur but silently not for the steppers, wheel or arrows. The description still only covers `changeAction`.

Green CI and a merge and I'll take it; we're in [Discord](https://discord.com/invite/XnsUcFykEP) if you'd rather talk any of it through.

[Reviewed by Robohands]

cixzhang:
Checking in on NumberInput `changeAction`: the shared prop typing and every stepper/wheel/arrow commit path still need coverage after merging main. Would you like to continue, have us carry it, or close this for now?

Issue #3679, opened 2026-07-08
Table styling API is non-standard: root props silently dropped, tableProps escape hatch, internal styles vs xstyle


cixzhang:
## Summary

The `Table` styling API diverges from the Astryx standard in three ways. Most importantly, `Table` **advertises** the standard styling props (`className`, `style`, `xstyle`, `id`, `aria-*`) by extending `BaseProps`, but silently drops them at runtime — the only styling channel that actually works on the root `<table>` is the non-standard `tableProps` escape hatch:

```tsx
// What consumers write today — the "weird" part:
<Table
  data={users}
  columns={columns}
  tableProps={{
    className: 'custom-table-class',
    style: {borderCollapse: 'separate'},
  }}
/>
```

Every other component takes styling directly on the element:

```tsx
// The Astryx standard (Card, Stack, Box, …):
<Card className="custom-class" style={{…}} xstyle={overrides.root} aria-label="…" />
```

## The standard

Root components extend `BaseProps`, destructure `xstyle` / `className` / `style`, spread remaining HTML attributes with `...props`, and merge via:

```tsx
mergeProps(themeProps('card', {variant}), stylex.props(...styles, xstyle), className, style)
```

`xstyle` (a `stylex.create()` value) is the primary styling channel; `className` / `style` are secondary; all other HTML attributes (`id`, `aria-*`, `data-*`, handlers) spread directly onto the root element. See `Card.tsx`, `Stack.tsx`.

## Finding 1 — root styling props are silently dropped

`TableProps extends Omit<BaseTableProps<T>, …>` and `BaseTableProps extends BaseProps<HTMLTableElement>`, so the type surface promises `className`, `style`, `xstyle`, `id`, `aria-*`. `TableInner` forwards them to `BaseTable` via `...rest` — but `BaseTableInner` destructures a **fixed prop set with no `...rest`**:

```tsx
function BaseTableInner<T>({
  data, columns, idKey, plugins, children,
  tableProps: userTableProps,   // ← only this reaches the <table>
  textOverflow, scrollWrapper, emptyState, ref,
}: BaseTableProps<T> & {ref?: Ref<HTMLTableElement>}) { … }
```

Everything not in that list — including every prop inherited from `BaseProps` — is discarded. Confirmed empirically by rendering `<Table>` with each prop and inspecting the root `<table>`:

| Prop | Applied to `<table>`? |
|------|----------------------|
| `className="my-class"` | ❌ no |
[...]

PR #4391, opened 2026-07-26
fix(table): honor className and style on TableRow

Eloitor:
## Summary

`TableRowProps` extends `BaseProps`, so `className` and `style` are part of the component's public type surface. Both were spread onto the `<tr>` before `mergeProps()` and then overwritten by the row's own StyleX classes, so passing either had no effect and produced no error.

They are now merged through `mergeProps()` alongside the row's StyleX styles, matching how `TableCell` and `TableHeaderCell` already handle them. Both rendering paths are covered: inside `Table` (context) and standalone (no context). Theme classes and striped/hover styling are unchanged.

Same class of problem as #3679, which was fixed on the root `<table>` element.

## Test plan

- Two regression tests added to `Table.test.tsx`; both fail before the change and pass after it.
- `vitest run packages/core/src/Table/Table.test.tsx` — 119 passed.
- `pnpm check:repo` and `eslint` on the changed files — clean.


cixzhang (approved):
Thank you for this fix! I'm going to add a linter for these so hopefully fewer over time

Eloitor:
Do I need to take a look at the failing CI builds?

cixzhang:
All good, might have been stale. I'm merging in main and retrying

PR #4395, opened 2026-07-26
fix(dropdown-menu): guard hover styles and forward rest props

cixzhang:
`DropdownMenuItem`, `DropdownMenuCheckboxItem`, and `DropdownMenuRadioItem` had an unguarded `:hover` background overlay. On touch devices this produced a sticky highlight that persisted after tap.

Fix: guard the item hover overlay with `@media (hover: hover)` so it only applies on true pointer devices, and keep the existing `:focus` overlay for the keyboard roving-tabindex highlight. Each menu item now explicitly owns both of its interactive background states (guarded hover + focus) in its own style block instead of implicitly leaning on the base `Item`'s internal hover — so pointer and keyboard feedback are declared together and stay consistent.

The guarded hover is expressed as a selector-level rule (`:hover` → `@media (hover: hover)` → `backgroundColor`) rather than nesting the `@media` inside the `backgroundColor` value-conditional. The value-conditional form typed the property value as an object literal, which isn't assignable to `xstyle`'s `StyleXStyles` type; the selector-level form composes cleanly through `Item`'s `xstyle` prop with no cast, matching how `CommandPaletteItem` guards its hover overlay.

Also: `DropdownMenu` extends `BaseProps` but never forwarded pass-through attributes (aria-*, data-*, id) to the menu element. Consumer-passed attributes were silently dropped. Fixed by spreading the `BaseProps` rest onto the menu div before contract props.

## Testing

- `pnpm vitest run packages/core/src/DropdownMenu/` — 48 passing
- `tsc --noEmit` (core) — clean

## Needs Review

The rest-prop forwarding casts `props` to `Record<string, unknown>` to strip the discriminated union — the runtime behavior is correct but the type narrowing is lost. If there's a preferred pattern for extracting rest from a discriminated union, flag it.

### DIFF
--- a/.changeset/dropdown-menu-hover-guard.md
+++ b/.changeset/dropdown-menu-hover-guard.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Guard DropdownMenu item hover styles with `@media (hover: hover)` to prevent sticky highlights on touch devices. Forward BaseProps pass-throughs to the menu element.
+
+@cixzhang
--- a/packages/core/src/DropdownMenu/DropdownMenu.tsx
+++ b/packages/core/src/DropdownMenu/DropdownMenu.tsx
@@ -200,6 +200,14 @@ export function DropdownMenu({
   const items = ('items' in props ? props.items : undefined) ?? [];
   const children = props.children;
 
+  // Extract BaseProps pass-throughs (aria-*, id, event handlers) from the
+  // discriminated-union rest bag so they can be forwarded to the menu element.
+  const {
+    items: _items,
+    children: _children,
+    ...rest
+  } = props as Record<string, unknown>;
+
   const menuId = useId();
   const menuSize = button.size ?? 'md';
   const buttonRef = useRef<HTMLButtonElement>(null);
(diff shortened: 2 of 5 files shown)


cixzhang:
Good call — reworked so the menu items no longer lean on `Item`'s internal hover for pointer feedback. Each of the three items now owns both interactive background states explicitly in its own `backgroundColor` block:

```
backgroundColor: {
  default: 'transparent',
  ':hover': {'@media (hover: hover)': colorVars['--color-overlay-hover']},
  ':focus': colorVars['--color-overlay-hover'],
},
```

So the `:hover` overlay is restored (fixing the earlier inconsistency of relying on `Item` for hover while hand-rolling `:focus` here) but guarded with `@media (hover: hover)` — the touch-device sticky-highlight bug stays fixed. The `:focus` overlay is kept for the roving-tabindex keyboard highlight. Hover and focus are now declared side by side, so the item is self-documenting about its own states.

One wrinkle: passing a style block with a nested `@media`-conditional value through the `xstyle` (`StyleXStyles`) prop doesn't typecheck against the strict prop type, so the composed array is cast `as unknown as StyleXStyles` — same pattern `ClickableCard` already uses for its guarded hover overlay. `pnpm vitest run packages/core/src/DropdownMenu/` (48 passing) and `tsc --noEmit` are clean.

Issue #2090, opened 2026-05-07
feat(cli): support swizzle for external packages


cixzhang:
## Summary

`xds swizzle` is currently hardcoded to `@xds/core`. External packages that opt into the XDS ecosystem (via `package.json "xds": { "docs": "./src" }`) already have full discovery and docs support through `xds discover`, but swizzle can't reach them.

## Proposal

Allow swizzling components from any discovered external package:

```bash
# Explicit package
xds swizzle @nds/components/AppShell

# Fuzzy (falls through to externals when not found in core)
xds swizzle AppShell
```

### Implementation path

1. **Fallthrough resolution** — when the component isn't found in `packages/core/src/`, call `scanAllPackages(config.packages)` → `findComponentInPackages()` to locate it in an external lib.

2. **Source directory** — resolve from the package's `docsDir` (since `xds.docs` typically points at `./src`, source lives alongside docs).

3. **Import rewriting** — parameterize `rewriteImports()` to use the external package name (e.g. `@nds/components/utils`) instead of hardcoded `@xds/core/...`.

4. **Disambiguation** — add `--from <package>` flag for when multiple packages export the same component name.

### Open question

If an external package separates docs from source (`"docs": "./docs"` but source in `./src/`), swizzle would need a separate field like `"xds": { "source": "./src" }`. For now, the common case (`"docs": "./src"`) works without changes to the package.json contract.

## Existing infrastructure that makes this easy

- `package-scanner.mjs` already resolves external packages and their `docsDir`
- `findComponentInPackages()` already locates components across packages
- `discover` command already handles `@scope/name/Component` syntax
- Gap reporting works generically (not core-specific)

cixzhang:
Thanks for the detailed implementation path here—the shape it landed in follows it closely. Swizzle now resolves components across core plus every loaded integration, rewrites imports against the owning package rather than a hardcoded core path, and supports `astryx swizzle <Name> --package <pkg>` for ambiguous names. Packages that separate docs from source are handled through independent lookups too. #4403 closed unmerged; the capability arrived with the swizzle API refactor in #4480. Closing as done—please reopen if you find an integration layout that swizzle still cannot reach.

PR #4403, opened 2026-07-26
feat(cli): swizzle components from external packages

AKnassa:
## What this does

`astryx swizzle` can now eject a component that lives in someone else's package, not just in Astryx core.

## Why

The issue asked for this, and three of its four implementation steps had already shipped — import rewriting was parameterized, `--package` disambiguation existed, and integrations resolved. Only the last hop was missing, so a package that opts in with `"astryx": {"docs": "./src"}` stayed invisible to swizzle even though `discoverExternalPackages()` and `scanAllPackages()` already knew how to find it.

## What changed

- Swizzle falls back to scanning those packages, using the call path the issue names (`scanAllPackages()` → `findComponentInPackages()`). No new flag, no new config field.
- A missing `@astryxdesign/core` no longer aborts a resolve that succeeds elsewhere.
- `--list` and the "did you mean" suggestions include external components, and they join `--package` disambiguation.
- External components eject under the package's own casing, so `swizzle appshell` writes `AppShell`.

## One thing worth a reviewer's eye

The widened `.doc.{ts,mjs,js}` lookup is **opt-in per call site**, not a change to the shared scanner's default. Widening it globally regressed `astryx discover`, which `import()`s the doc it finds: Node refuses to type-strip anything under `node_modules`, so a clean empty result became a hard `ERR_INVALID_DOC`. `discover` output is byte-identical to before the change — verified by diffing its raw JSON across three invocations against pristine `main`.

## What this does not do

Flat-layout packages (doc file sitting directly at the docs root alongside source) report a clear `ERR_NO_SOURCE` rather than ejecting, since the naive resolution would copy an entire `src` tree. The issue itself flags that shape as an open question.

## How to check it

`pnpm vitest run packages/cli/` — 118 files, 2088 tests, green. The routing suite gained 6 tests that build real temp packages and spawn the actual CLI binary. Full repo suite run alone: 380 files / 7757 tests, exit 0.

One limitation stated plainly: a test asserting that core/integration owners do *not* canonicalize casing can't be written portably — macOS is case-insensitive, so it would pass locally and fail on Linux CI. The portable invariants are pinned instead.

Refs #2090


cixzhang:
Thanks @AKnassa — the read behind this was right, and worth recording: three of the four steps in [#2090](https://github.com/facebook/astryx/issues/2090) had already shipped and only the last hop was missing. Keeping the widened `.doc.{ts,mjs,js}` lookup opt-in per call site rather than changing the shared scanner's default was a good catch too — the global version would have turned a clean empty `astryx discover` into a hard error, since Node won't type-strip anything under `node_modules`.

Closing because the ground moved out from under it. [#4623](https://github.com/facebook/astryx/pull/4623) reorganised the CLI and deleted every file this branch builds on: `cli/commands/swizzle.mjs`, `lib/package-scanner.mjs`, `utils/paths.mjs` and `types/swizzle.d.ts` are all gone. Swizzle now lives under `api/swizzle/`, the scanner is `api/discover/_package-scanner.mjs`, and path helpers moved to `foundation/fs/`. Rebasing this would be a rewrite rather than a merge, which isn't a fair thing to ask of a branch you already finished.

The feature is still one we want, and [#2090](https://github.com/facebook/astryx/issues/2090) stays open to hold it. If you'd like to take it again against the new layout, reopen this or open a fresh PR any time — merge `main` in first. And [Discord](https://discord.com/invite/XnsUcFykEP) is there if you'd rather talk the new shape through with someone before writing it.

[Reviewed by Robohands]

Issue #3520, opened 2026-07-04
Define InputGroup-compatible controls


cixzhang:
## Problem

`InputGroup` is a composite input surface: it visually connects one compatible input control with prefix/suffix addons inside a shared border and focus ring. After #3502, `TextInput` and `NumberInput` correctly consume `InputGroupContext` so they can remove their own field chrome and compose the group label/description/status with their own input ARIA.

The docs and stories should make this compatibility boundary explicit. Some controls look input-like but do not make sense inside `InputGroup` because they are not single-line composite input surfaces or because they have their own group semantics.

## Current compatibility

Supported today:

- `TextInput`
- `NumberInput`
- `InputGroupText` addons

Not compatible by design:

- `TextArea` — multi-line field; shared inline border/addon layout does not fit the interaction or layout model.
- `Slider` — has track/thumb semantics, not inline text-entry semantics.
- `Switch` — boolean toggle control, not a decorated text-entry surface.
- `CheckboxInput` — boolean control with its own label/status layout.
- `RadioList` — already a grouped composite control with radiogroup semantics.

Potential follow-up candidates, if explicitly designed and implemented:

- `TimeInput`
- `DateInput`
- `Selector`
- `MultiSelector`
- `Typeahead`
- `Tokenizer`

These should not be documented as compatible until they explicitly consume `InputGroupContext`, suppress their own `Field` wrapper when grouped, apply grouped border/radius styling, and compose group + input ARIA with the shared helper.

## Proposed work

1. Document the compatibility boundary in `InputGroup` docs.
2. Add Storybook stories that show every supported compatible control.
3. Add support for additional compatible single-line controls in small PRs, each with tests and Storybook coverage.
4. Keep unsupported controls documented as unsupported so users do not place arbitrary form controls inside `InputGroup`.

## Acceptance criteria

- `InputGroup` docs explicitly list supported controls and non-compatible controls.
- Storybook shows compatible examples for all supported controls.
- Any newly compatible control:
  - consumes `InputGroupContext`
  - uses shared grouped border/radius styles
  - avoids rendering duplicate `Field` chrome inside `InputGroup`
  - composes group + input label/description/status via `getInputARIA`
  - has regression tests for grouped accessible name and described-by wiring

Refs #3343

PR #4405, opened 2026-07-26
feat(core): make Tokenizer InputGroup-compatible

AKnassa:
## What this does

Tokenizer now works properly inside an `InputGroup` instead of rendering its own duplicate label and field chrome.

## Why

Most of this issue already shipped — the compatible and incompatible lists are documented, the other seven controls all consume `InputGroupContext`, and there's a story per control. Tokenizer was the one candidate left from the issue's own list with zero `InputGroup` references.

## What changed

Following the recipe the issue's acceptance criteria state verbatim, mirroring `Typeahead`:

- Inside a group, Tokenizer stops rendering its own `Field`, so the group owns the single label, description and status.
- The inner input is named by the group label plus Tokenizer's own, composed through `getInputARIA` so `aria-describedby` **merges** the group's description with the input-local disabled reason instead of replacing it.
- It adopts the shared grouped border and radius so it joins the seam.
- Added a `WithTokenizer` story and listed Tokenizer in the compatible-controls documentation.

## Needs a maintainer call

`InputGroup` is a **fixed-height single-line surface**, but Tokenizer wraps by default — so wrapped tokens would render outside the group's shared border. The grouped variant therefore keeps tokens on one scrolling row.

The alternatives all seemed worse: letting it spill (visibly broken), clipping with `overflow: hidden` (tokens silently vanish), or making the group auto-height (needs `InputGroup` changes, outside this slice). But this **enforces** a contract rather than reading one, so it's a judgement call, not something the issue specified. Two costs worth naming: the scrollbar is hidden, so overflow isn't visually signalled, and a focused token's outline can clip at the scroll edge.

## Known, and deliberately left alone

In group mode the `description` and `status` props are dropped along with `Field`, but the ARIA helper still receives their IDs — so a grouped control with a description emits an `aria-describedby` pointing at nothing. **`Typeahead` and `MultiSelector` have exactly the same wart**, so fixing it for one of three would be worse than fixing none. Browsers ignore missing IDs, so impact is nil in practice. Worth its own issue.

## How to check it

`pnpm vitest run packages/core/src/InputGroup/ packages/core/src/Tokenizer/ packages/core/src/Typeahead/ packages/core/src/MultiSelector/ packages/core/src/Field/` — 6 files, 228 tests, green. Full repo suite run alone: 380 files / 7747 tests, exit 0.

**Not visually verified** — jsdom does no layout. The `WithTokenizer` story is there for a human to check the seam.

Refs #3520


cixzhang (changes requested):
This still lets builders create a grouped field that escapes its row or exposes contradictory validation, so I’m requesting changes. The default grouped row is now single-line, and standalone matched the merge-base pixel-for-pixel.

- Complete grouped adaptation: an `lg` Tokenizer inside an `sm` group renders 36px inside a 28px row, while local `description`/`status` leave two dead `aria-describedby` IDs and paint a warning border beside the group’s error state (`Tokenizer.tsx:451,721,800`).
- Replace both render-time `mergeRefs(...)` calls with stable merged refs; strict lint is red at lines 774 and 898.
- Rebase onto current `main`, preserving both its busy-lane reserve and this PR’s `ariaLabelledBy`; the branch currently conflicts in `Tokenizer.tsx`.

Please add focused grouped tests for mismatched sizes and local metadata.

[Reviewed by Robohands]

Issue #3642, opened 2026-07-06
[Tracker] Schedule component — Q3 maturation (interaction, docs, graduation path)


cixzhang:
## Problem

`@astryxdesign/lab` ships a `Schedule` component — a read-only calendar shell with Day / Week / Month / List / TimeGrid views, a plugin system (pagination, view selector), timezone-aware date math, and a full event/type surface. It's a solid foundation, but it's incomplete as a first-class scheduling surface: it's read-only, its `.doc.mjs` carries no usage description or documented props, and it hasn't been through the graduation review that would move it from `lab` toward `core`.

This is a **tracker** for the Q3 push to mature the Schedule components. Individual enhancements are filed and linked below; each lands on its own via the normal issue → API → PR flow.

## Scope — starter set (on this milestone)

- [ ] Documentation — fill in `Schedule.doc.mjs` (usage description + props), Storybook coverage per view
- [ ] Event interaction — event selection / click affordances (read → interactive)
- [ ] Event mutation — create / edit / move (drag) events with controlled callbacks
- [ ] Accessibility pass — keyboard navigation across grid views, ARIA roles, focus management
- [ ] View parity — audit Day/Week/Month/List/TimeGrid for consistent options + plugin support
- [ ] Graduation criteria — define what "core-ready" means for Schedule (API arbitration, test coverage, docs)

## Notes

This is a starting set — scope to be adjusted as priorities firm up. Each sub-item follows the spec protocol independently (problem → research → API → PR). This issue only tracks the theme so the milestone reflects "Schedule component" as one major line of work.

PR #4406, opened 2026-07-26
docs(lab): document the Schedule component

AKnassa:
## What this does

Writes the missing documentation for the lab `Schedule` component.

## Why

A ~4,300-line component across 19 files had a 16-line placeholder in front of it: an empty description, an empty prop list, and a category of "Data Input" for something the source itself describes as a read-only schedule shell.

## What changed

`Schedule.doc.mjs` gains a real description, six best practices, seven anatomy entries, all nine props, and the corrected category. Every sentence traces back to the source or the existing Storybook stories.

## Worth knowing before reviewing

**Nothing in this repo reads lab docs.** The prop-reference and prop-literal guards and the theming-targets test all root at `packages/core/src`; the docsite registry only ingests its own dependencies, and lab isn't one; the CLI's component discovery scans core's `src`. `astryx component Schedule` prints "No component named Schedule" both before and after this change.

So this surface is write-only today, and no guard can catch drift in it. Since the drift test couldn't do the job, correctness was proved another way: parsing the `ScheduleProps` interface straight out of the source and diffing it against the documented list — zero undocumented, zero invented, zero required-flag mismatches — plus a schema parse through the CLI's own loader.

If the point of this tracker item is agent-facing discoverability, **wiring lab into one of those pipelines is the actual unlock**, and it's a separate piece of work.

## Scope

Documentation item only. The tracker's other five items — event interaction, mutation, the a11y pass, the view-parity audit and graduation criteria — each need a decision that hasn't been made, and are untouched.

## Judgement calls

The category is set to `Content`, matching how the repo treats other read-only display components (`Timestamp`, `LogStream`, `Stat`). `Table & List` is the defensible alternative; it's a one-word change if preferred.

A few prop descriptions are inferred rather than read, and flagged as such: the `view` factories have no docs of their own, and the guidance to keep an async loader reference stable comes from reading its cache, not from a documented contract.

## How to check it

`pnpm vitest run packages/lab/src/Schedule/` plus the doc guards — 7 files, 536 tests, green. No changeset: `packages/lab` is private, and the changeset check rejects private packages outright.

Refs #3642

### DIFF
--- a/packages/lab/src/Schedule/Schedule.doc.mjs
+++ b/packages/lab/src/Schedule/Schedule.doc.mjs
@@ -6,11 +6,151 @@ export const docs = {
   name: 'Schedule',
   displayName: 'Schedule',
   group: 'Schedule',
-  category: 'Data Input',
+  category: 'Content',
 
   usage: {
-    description: '',
+    description:
+      'Schedule is a read-only calendar surface that renders events as a month grid, a day or week time grid, or a list grouped by day — the layout comes from a view object you pass in. It handles timezone-aware date math, paging between ranges, and async event loading, and exposes header slots that plugins fill with navigation controls. Use it to display an existing schedule; it has no event selection, creation, or editing affordances.',
+    bestPractices: [
+      {
+        guidance: true,
+        description:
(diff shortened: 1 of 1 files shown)


cixzhang (approved):
Thanks for filling in the Schedule docs — having the component's props and usage properly documented in the lab makes it far more approachable for anyone picking it up. Much appreciated!

Issue #823, opened 2026-03-22
feat: RSC-native components via `react-server` conditional exports


cixzhang:
## Summary

XDS can become the first design system to offer genuinely server-renderable components — not by wrapping everything in `"use client"`, but by using the [`react-server` conditional export](https://github.com/reactjs/rfcs/pull/227) to let bundlers resolve the right module automatically.

When a consumer writes `import { Badge } from "@xds/core/Badge"` in a Server Component, the bundler resolves the server-safe version (no `"use client"`, pure HTML output). In a Client Component, it resolves the interactive version. Same import, different resolution. Zero friction for consumers.

## Motivation

The current state of the React component library ecosystem:

| Library | RSC approach |
|---------|-------------|
| Chakra UI | All client. `"use client"` everywhere. |
| MUI | Runtime theme via context. All client. |
| Mantine | All client. |
| Radix | All client for interactive primitives. |
| Headless UI | All client. |
| **next-intl** | `react-server` condition at package level ✅ |
| **Payload CMS** | Explicit `./rsc` subpath export ✅ |
| **XDS (proposed)** | `react-server` condition at per-component level 🆕 |

No major design system has solved server-renderable themed components yet. XDS is positioned to be the first because:

1. Themes resolve via CSS (`light-dark()`, CSS layers) not runtime context
2. StyleX compiles to static CSS at build time — no runtime JS for styling
3. 18 of 60 components use zero client APIs (no hooks, no event handlers, no context)

## The `react-server` condition

From [React RFC 227](https://github.com/reactjs/rfcs/pull/227), the `react-server` condition in `package.json` exports tells the bundler to resolve a different file when the import happens inside a React Server Component environment:

```jsonc
// package.json exports
{
  "./Badge": {
    "react-server": "./dist/Badge/index.mjs",
    "default": "./dist/Badge/index.mjs"
  },
  "./Dialog": {
    "react-server": "./dist/Dialog/index.server.mjs",
    "default": "./dist/Dialog/index.mjs"
  }
}
```

This is the same mechanism used by `server-only` and `client-only` packages, and by `next-intl` for its server/client API split.

### Tier 1: Server-safe (18 components)
No hooks, no events, no context. Pure HTML + CSS. Can render entirely on the server with zero JavaScript shipped to the client.

[...]

PR #4407, opened 2026-07-26
feat(core): drop 'use client' from server-safe primitives

AKnassa:
## What this does

Marks 10 simple presentational components as safe to render on the server, by removing their `'use client'` directive.

## Why

It's Phase 1 of the issue, written there verbatim: "Remove `use client` from Tier 1 source files."

## The Tier 1 list in the issue is four months stale

Copying it would have stripped the directive off components that genuinely need it:

| Component | Why it must keep the directive |
|---|---|
| `Kbd` | `useSyncExternalStore` |
| `ProgressBar` | `useId` |
| `Divider` | `useId` — labeled dividers gained it upstream (#4370) while this PR was open |
| `Field` | reaches `LinkContext` (`createContext`) through `Button` |
| `OverflowList` | `useOverflow` — `useState`, `useCallback`, `useRef` |
| `StatusDot` | imports `Tooltip` |
| `Icon` | module-level registry, mutated by the public `registerIcons()` |

So the set here is **derived from the actual import graph** instead. All 8 obvious candidates confirmed, and three components on no list at all qualify: `Blockquote`, `Code`, `VisuallyHidden`. Divider was originally in the dropped set; the rebase brought in #4370 and the guard flagged it, so it moves to the table above under the same rule as ProgressBar.

## How the set was derived

The first classifier flagged every candidate **and every already-shipped directive-free component** as unsafe. Since those ship without the directive today, the model was wrong, not the ground truth. The fix: `sideEffects` in `package.json` is a narrow allowlist, so every other module is drop-eligible and traversal must follow pure barrels **by used export**. Under that model the classifier reproduces the repo's existing directive-free set exactly — that agreement is the calibration proof. Parsing goes through the TypeScript compiler API so type-only imports erase correctly.

## The guard, and why the existing script didn't catch this

The new test derives the safe set from the import graph rather than restating it. That matters: adding `import {Tooltip}` to `Badge.tsx` leaves `scripts/check-use-client.mjs` **passing**, because it only checks direct React client APIs.

Review round: both synchronized `CLIENT_APIS` lists were missing React 19's `useActionState`, `useEffectEvent` and `startTransition` — the full set of exports react's own `react-server` build omits. All three are covered now, and two meta-guards close the class: the client-only set is checked against react's builds (a future client-only hook fails the test on its own), and the two lists are asserted identical so they cannot drift apart.

The orphan status originally reported here has since been fixed on main — `check:repo` now runs the script in lint and pre-commit, so the direct-import direction is CI-enforced. The transitive direction ships with this PR's test.

## Honest framing

This **unblocks Phase 2; it does not ship RSC support**. Without the `react-server` export condition the export map still has a single path, so no consumer sees a difference yet.

Not breaking: no prop, type or export changed, and the directive is inert outside an RSC bundler. Bundle *layout* can shift for existing client consumers — these are no longer client chunk boundaries, so they may inline into importing chunks rather than dedupe as shared entries. No behaviour change, size diffs possible either direction.

## How to check

`pnpm vitest run packages/core/src/serverSafeComponents.test.ts` → 18 tests: the parse rules, the react-build ground truth, list sync, and both invariants across every public component subpath. `node scripts/check-use-client.mjs` → 234 files, exit 0. Core typecheck clean. An earlier round verified the emitted `dist/` boundaries (targets stripped, keepers retained); the full suite runs in CI on the current head.

Refs #823

### DIFF
--- a/.changeset/rsc-server-safe-components.md
+++ b/.changeset/rsc-server-safe-components.md
@@ -1,0 +1,9 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] AspectRatio, Badge, Blockquote, Card, Center, Code, Divider, Grid, Section, Skeleton and VisuallyHidden no longer carry `'use client'`. Each was verified against its transitive import graph to use no React client API, no client-only dependency and no module-level mutable state, so they can now render in a React Server Component without forcing a client boundary. A new `serverSafeComponents.test.ts` derives the server-safe set from the import graph and fails if one of these components later gains a client dependency without restoring the directive — including the transitive case `scripts/check-use-client.mjs` cannot see.
+
+Not a breaking change: no prop, type or export changed, and `'use client'` is inert outside an RSC bundler. Client consumers keep working identically, though bundlers may lay these modules out in different chunks now that they are no longer client entry points.
+
+@cixzhang
--- a/.changeset/rsc-server-safe-components.md
+++ b/.changeset/rsc-server-safe-components.md
@@ -2,7 +2,7 @@
 '@astryxdesign/core': patch
 ---
 
-[feat] AspectRatio, Badge, Blockquote, Card, Center, Code, Divider, Grid, Section, Skeleton and VisuallyHidden no longer carry `'use client'` (#823). Each was verified against its transitive import graph to use no React client API, no client-only dependency and no module-level mutable state, so they can now render in a React Server Component without forcing a client boundary. A new `serverSafeComponents.test.ts` derives the server-safe set from the import graph and fails if one of these components later gains a client dependency without restoring the directive — including the transitive case `scripts/check-use-client.mjs` cannot see.
+[feat] AspectRatio, Badge, Blockquote, Card, Center, Code, Grid, Section, Skeleton and VisuallyHidden no longer carry `'use client'` (#823). Each was verified against its transitive import graph to use no React client API, no client-only dependency and no module-level mutable state, so they can now render in a React Server Component without forcing a client boundary. A new `serverSafeComponents.test.ts` derives the server-safe set from the import graph and fails if one of these components later gains a client dependency without restoring the directive — including the transitive case `scripts/check-use-client.mjs` cannot see.
 
 Not a breaking change: no prop, type or export changed, and `'use client'` is inert outside an RSC bundler. Client consumers keep working identically, though bundlers may lay these modules out in different chunks now that they are no longer client entry points.
 
--- a/.changeset/rsc-server-safe-components.md
+++ b/.changeset/rsc-server-safe-components.md
@@ -2,8 +2,8 @@
 '@astryxdesign/core': patch
 ---
 
-[feat] AspectRatio, Badge, Blockquote, Card, Center, Code, Divider, Grid, Section, Skeleton and VisuallyHidden no longer carry `'use client'`. Each was verified against its transitive import graph to use no React client API, no client-only dependency and no module-level mutable state, so they can now render in a React Server Component without forcing a client boundary. A new `serverSafeComponents.test.ts` derives the server-safe set from the import graph and fails if one of these components later gains a client dependency without restoring the directive — including the transitive case `scripts/check-use-client.mjs` cannot see.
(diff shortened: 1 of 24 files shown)


AKnassa:
These four are pre-existing on `main`. This PR deletes `'use client'` lines and adds a test. It changes no token, style rule or markup.

`pr-a11y` has no baseline. It builds only the head Storybook (`ci.yml:448-453`), records every violation axe finds (`accessibility-audit.js:176-186`), and enrols a component when any file in its folder changed (`analyze-pr.js:425-439`). The same ratios appear on #4117, #4155, #4162, #4207, #4283, #4296 and #4349: Badge 2/5, Grid 2/12, Section 2/7, Card 8/14.

Card, Grid and Section are caption text in the story files, not component markup. All three are `--color-text-secondary` `#737373` on the Storybook body `#f1f1f1`, which is 4.20:1. It passes on white at 4.74:1, so only some stories trip. #3677 fixes it by moving the token to `#616161` (5.48:1), and its audit reports Card clean.

Badge is real and separate. `variant="error"` computes `#ffffff` on `#e33f4a`, 4.14:1 at 12px/500. The Badge block re-binds `--color-error` to `#e33f4a` (`neutralTheme.ts:605-607`) while `--color-on-error` stays white. `#c9303a` would give 5.29:1. Not fixed by #3677. Filed as #4414.

Unrelated suggestion: `accessibility-audit.js:228-234` drops axe's `failureSummary`, so these reports can never show the measured colours or ratio.

AKnassa:
**ProgressBar belongs in this set.** React 19.2.7's `react-server` entry exports `useId`, `useMemo` and `useCallback`, so those three do not belong in `CLIENT_APIS`. `useId` is ProgressBar's only React import (`ProgressBar.tsx:19`), used for `aria-labelledby`, and this PR clears its last blocker by freeing `VisuallyHidden`. As written, `serverSafeComponents.test.ts` will now fail anyone who removes its directive. Same question for `AvatarGroup`, `ButtonGroup`, `FormLayout`, `InputGroup`, `Layout`, `RadioList`, `SegmentedControl` and `Toolbar`, whose only direct client APIs are also `useId`, `useMemo` or `useCallback`.

**Two fixes for the description.** "No consumer sees a difference yet" is too strong: the directive is read from file contents, not the export map, so an RSC consumer importing `Badge` loses its client boundary as soon as this lands. And the `sideEffects` list also includes `**/*.stylex.js`.

cixzhang:
@imdreamrunner — could you take this one? You're already on the reviewer list; I'd like your call specifically rather than mine.

Why you: it's an RSC-correctness change across 11 components, and the failure mode is silent — a directive removed from something that actually needs client behavior doesn't fail a test, it fails in a consumer's server render. `T30` is a blocking check whose enforcer (`scripts/check-use-client.mjs`) isn't wired into any npm script or workflow, so CI green means nothing here and the review is the only gate.

The part worth your attention is the method rather than the list: the issue's Tier 1 list was four months stale, so @AKnassa derived the set from the actual import graph instead and documented six components that must *keep* the directive (`Kbd`, `ProgressBar`, `Field`, `OverflowList`, `StatusDot`, `Icon`). If that derivation is sound the rest follows; if it missed a transitive client import, we ship a broken server render.

@AKnassa — it's 387 commits behind `main` and currently conflicting, so it needs a rebase before a review can mean much. Would you mind bringing it up to date? Sorry it sat this long.

AKnassa:
Rebased onto main and mergeable again. Conflicts were only the Badge, Divider and Section barrels where the VariantMap interfaces moved in; the directive set is unchanged, and the two components added since (ComplexSelector, Indicator) both reach client surfaces, so the derived set holds.

imdreamrunner (changes requested):
The 11 existing `'use client'` removals look safe, including their component barrels, and the emitted distribution preserves those boundaries correctly.

One blocker remains: the synchronized `CLIENT_APIS` lists omit React 19 client-only APIs including `useActionState` and `useEffectEvent`. Adding an unconditional `useActionState()` call to directive-free `Badge` left both guards green, while a Next RSC build correctly failed because the API requires a client boundary.

Please cover the missing APIs in both classifiers and add a regression test. This does not invalidate the current removals; it ensures the new guard catches future regressions.

AKnassa:
Thanks for 4324f412f. startTransition turned out to be a third omission; both classifiers now carry all three (be4754f1b), with parse tests for named, aliased and namespace-access imports. Two meta-guards close the drift class: the client-only set is now checked against react's own react-server build, so a future client-only hook fails the test on its own, and the two lists are asserted identical. Body updated for the Divider move and the dropped count of 10.

imdreamrunner (approved):

imdreamrunner:
Thanks for improving this!

Issue #3349, opened 2026-07-02
RFC: InfoTip — inline info-icon help affordance


thedjpetersen:
## Problem Statement

Builders of internal tools need a small "i" info/help affordance next to labels and values — permission notes, metric definitions, field help. Users need to discover concise contextual help without cluttering the UI, and currently there is no pre-wired way to do this: every team hand-rolls an icon + tooltip composition, and most of those compositions are broken for keyboard and screen-reader users.

## Use case & evidence of demand

- Gap report from an internal-tool builder (**Meridian**): they needed an "i" info/help affordance next to labels/values; nothing exists in XDS, so they hand-composed an info `Icon` inside a `Tooltip`.
- This is one of the most common internal-tools affordances: permission notes ("Why can't I edit this?"), metric definitions on dashboards, field help on forms.
- External prior art: Ant Design, Carbon (`TooltipIcon`/toggletip), and Polaris all ship a tooltip-icon pattern; Radix and shadcn deliberately leave it to composition. Genuine divergence in the ecosystem — see Rough Approaches below.

## Why Existing Components Don't Cover This

`Tooltip` requires the consumer to supply a correct trigger. The naive Icon-in-Tooltip composition:

```tsx
<Tooltip content="Metric definition...">
  <Icon icon="info" />
</Tooltip>
```

gets the accessibility wrong **by default**:

- No focusable trigger — `Icon` renders `aria-hidden`, so keyboard users never see the tooltip.
- No accessible name — screen-reader users encounter nothing, or an unlabeled element.
- No touch affordance — hover-only trigger with no tap target sizing.

The value of InfoTip is a **pre-wired accessible trigger**, not new visuals.

**Honest triage note:** composition CAN do this correctly (button + aria-label + Icon inside Tooltip), which is the argument for shipping a documented recipe instead of a component. The RFC's case is that the a11y wiring is exactly what naive composition — and LLM codegen — gets wrong, so pre-wiring it has outsized leverage.

## Rough Approaches Considered

**Option A: standalone `InfoTip`** — an icon-button trigger pre-wired into `Tooltip`, with `label` (accessible name) and content props:

```tsx
<InfoTip content="Editors can change this field; viewers cannot." />
<InfoTip content="30-day rolling average." label="About this metric" />
```

**Option B: a `helpTip` slot on `FieldLabel`/`Field`** — narrower, forms-only:

```tsx
<FieldLabel helpTip="Editors can change this field.">Owner</FieldLabel>
```

[...]

PR #4408, opened 2026-07-26
feat(lab): open InfoTip on tap

AKnassa:
## What this does

`InfoTip` can now be opened by tapping it. Before this, on a touch device it could not be opened **at all**.

## Why

The RFC's own accessibility section requires "tap toggles the tooltip". The component merged without it, wiring only keyboard, blur and mouse-leave.

Core doesn't fill the gap either — and it's worse than "hover just doesn't fire". `useTooltip` **deliberately bails out of hover-open** under `(hover: none)`, and focus-open is gated on `:focus-visible`, which a tap does not satisfy. So on a phone the info icon was inert.

## What changed

The two-state dismissed flag becomes `auto` / `pinned` / `dismissed`, driving the controlled `isOpen` channel the component already owns.

- Clicking **pins unconditionally** rather than toggling on what's currently visible. With a mouse the pointer is by definition over the trigger at click time, so a toggle would almost always read "open" and close it — and since `mouseenter` has already fired it wouldn't re-open. Pinning also makes the gesture independent of whether the 200ms hover delay had elapsed.
- A pinned tip survives pointer-leave (that's the whole value of pinning) and closes on a second click, Escape, or focus leaving.
- Closing routes through `dismissed`, never back to `auto` — `auto` only stops overriding Tooltip and cannot close an open layer, so returning to it would leave the tip stuck open. That's the sharpest edge in the change and is commented in-file.

## Scope question for a reviewer

**A reviewer arguing this belongs in `Tooltip` instead has a strong case**, and I'd rather say so than not. Core already owns the touch policy — it's what suppresses hover under `(hover: none)` — and it left no escape hatch, so every future icon-button-plus-Tooltip pairing inherits the same dead end and will re-invent this state machine.

Against that: `Tooltip` is core and consumed everywhere, a click path there could collide with triggers that have their own click semantics, and it needs an API decision adjacent to this RFC's unresolved Option A/B question. So the lab-scoped version is the right *first* move and is cheap to delete if redirected — but treat it as a **candidate for graduation to `Tooltip`, not the final home**.

## Deliberately not included

Outside-click dismissal. An outside tap fires both `pointerdown` and `blur`; if those batch, the net state lands on `auto` — stuck open, which is the worst possible outcome on exactly the devices this fixes. Tap-to-toggle alone satisfies the RFC and keeps one handler per event.

The RFC's open questions (Option A vs B, toggletip semantics, lab→core graduation) are untouched.

## How to check it

`pnpm vitest run packages/lab/src/InfoTip/ packages/core/src/Tooltip/` — 29 tests, green, no collateral on Tooltip. Full repo suite run alone matches baseline exactly.

The tests were **mutation-tested**, and that caught a fake: the original "keeps pinned" test passed even when `handleMouseLeave` was mutated to unconditionally reset, because `auto` maps to `isOpen: undefined` and core no-ops on that. An extra test now kills that mutant; all four mutants die.

**Not verified on a real device** — jsdom does no layout and dispatches no real touch events, so `fireEvent.click` is synthetic. Every touch claim here is reasoned from core's source, not measured. One known limit: while pinned, core's document-level Escape listener is disabled, so Escape relies on the trigger having focus — Safari and Firefox on macOS don't focus buttons on click, so there a second click closes it but Escape won't. A document-level listener was deliberately not added, since one keypress would then close both the tip and any enclosing dialog.

No changeset: `packages/lab` is private, and the changeset check rejects private packages outright.

Refs #3349

### DIFF
--- a/packages/lab/src/InfoTip/InfoTip.doc.mjs
+++ b/packages/lab/src/InfoTip/InfoTip.doc.mjs
@@ -29,12 +29,13 @@ export const docs = {
   ],
   usage: {
     description:
-      'An inline info-icon help affordance: a small "i" button that reveals a tooltip on hover and keyboard focus. Use it next to labels, values, and metrics for permission notes, metric definitions, and field help. Its value over hand-composing Icon inside Tooltip is the pre-wired accessible trigger: a real button with an aria-label, Tab-reachable, tooltip on hover AND focus, and Escape dismissal.',
+      'An inline info-icon help affordance: a small "i" button that reveals a tooltip on hover, keyboard focus, and tap. Use it next to labels, values, and metrics for permission notes, metric definitions, and field help. Its value over hand-composing Icon inside Tooltip is the pre-wired accessible trigger: a real button with an aria-label, Tab-reachable, tooltip on hover AND focus, tap-to-toggle for touch, and Escape dismissal. Tapping or clicking pins the tooltip open so it survives the pointer leaving; tapping again, Escape, or moving focus away closes it.',
     bestPractices: [
       { guidance: true, description: 'Keep tooltip content concise, plain, and non-interactive; use Popover or HoverCard for links and buttons.' },
       { guidance: true, description: 'Override label when "More information" is too generic, e.g. "About this metric".' },
       { guidance: false, description: 'Hand-compose Icon inside Tooltip for info affordances; the bare Icon is aria-hidden and unfocusable, so keyboard and screen-reader users never see it.' },
       { guidance: false, description: 'Use InfoTip for essential information users must see to complete a task; put that text inline instead.' },
+      { guidance: false, description: 'Wrap InfoTip in a clickable row or card; its trigger consumes the tap, and the surrounding click target still fires.' },
     ],
   },
 };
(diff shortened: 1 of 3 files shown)


cixzhang:
Thanks @AKnassa — the diagnosis here was the valuable part, and it was sharper than "hover doesn't fire on touch": `useTooltip` deliberately bails out of hover-open under `(hover: none)`, and focus-open is gated on `:focus-visible`, which a tap does not satisfy — so the info icon was genuinely inert on a phone, not merely awkward.

Closing because that is fixed on main. [#5248](https://github.com/facebook/astryx/pull/5248) added a `touchTrigger` channel to `Tooltip` and `HoverCard` and set `touchTrigger="tap"` on InfoTip's trigger, so a tap opens it. It even carries your reasoning as the comment on that line — the trigger is a real button, so the `auto` touch rule would hand the tap to the control, but here the tooltip *is* the control's only purpose, so the tap has to open it. InfoTip's docblock now reads "hover, keyboard focus, and tap".

It went in through the shared Tooltip/HoverCard layer rather than through InfoTip's own dismissed-flag state machine, which is why your branch isn't the thing that landed. If you try it and find a case your `auto`/`pinned`/`dismissed` model handled that the new one does not — the pin-unconditionally-on-click behaviour especially — reopening is genuinely welcome; merge `main` in first.

[Discord](https://discord.com/invite/XnsUcFykEP) if it is easier to talk through with a person.

[Reviewed by Robohands]

Issue #4276, opened 2026-07-24
Adoption feedback: migrating a real Next.js 16 app to Astryx (0.1.8) — 10 findings + suggestions


ejc3:
## Adoption feedback: migrating a real Next.js 16 app to Astryx (0.1.8)

We migrated a production dark-only chat/transcript viewer onto `@astryxdesign/core@0.1.8` — Next.js 16 App Router, React 19, **webpack** build, pnpm 10 (strict `node_modules`), a strict CSP (`style-src 'self' 'unsafe-inline'`, `font-src 'self'`, `connect-src 'self'`), and ~130 Playwright locators against our own CSS classes. Overall it went well, and a few things were genuinely excellent — so first, credit where it's due, then the actionable findings.

Happy to split any of these into individual tracked issues if you'd prefer; consolidating so triage is one read.

### What worked really well

- **`BaseProps` forwards `className` / `style` / `data-*` / `aria-*`.** This is what made an *incremental* migration of an app with ~130 CSS-selector tests possible at all — we could swap components without rewriting the test suite in the same commit. Arguably the single most important adoption property; worth advertising louder.
- **The CLI is a better docs surface than the website.** `astryx docs <topic>` / `component <Name>` / `hook` / `template`, plus `--json` and `--dense`, gave complete, offline, machine-readable docs. We initially tried scraping the docs site (Next RSC payload, fetches poorly) and threw it away once we found the CLI.
- **`astryx docs migration` predicted our worst bug before we hit it** — the "Cascade Layer Safety" section names the exact webpack `@import`-hoisting failure and the `layers.css` remedy. Docs that name the specific silent failure of a specific bundler are rare.
- **Pre-compiled CSS is genuinely build-step-free on Next+webpack**, and CSP-clean (no `@font-face`, no remote `url()` in `reset.css`/`astryx.css`/the built theme).
- **`TextArea`'s iOS auto-zoom guard** (`@media (pointer: coarse) { font-size: max(1rem, …) }`) is pointer-aware — better than the unconditional 16px we had.

---

### Bugs / friction (roughly by cost)

**1. `@stylexjs/stylex` is missing from the documented install line — hard runtime failure on pnpm.**
README Quick Start and `astryx docs getting-started` both say `npm install @astryxdesign/core @astryxdesign/theme-neutral`. But `@stylexjs/stylex` is a *peer dependency* of core, and **202 of `dist/`'s files import it at runtime**:
```
[...]

PR #4409, opened 2026-07-26
fix(core,cli): four defects from the Next.js 16 adoption report

AKnassa:
Four independent defects from the adoption report, as four separate commits so they can be split or dropped individually.

## 1. `astryx docs --list` crashed

It was documented usage, but the command was registered with no options at all, so commander rejected the flag and exited 1 — even though a list branch already existed and the bare `docs` invocation reached it. Now wired up, and `--list` wins over a positional topic.

## 2. `theme build --out` help text was misleading

It said "Output CSS file path", but **all four** emitted artifacts follow that path's dirname. Reworded to the phrasing the reporter asked for.

## 3. The documented install line omitted a required peer

`@astryxdesign/core` declares `@stylexjs/stylex` as a required peer with no `peerDependenciesMeta`, but every documented install line left it out — so following the quick start under pnpm gives you a broken install. Added to each package-manager variant, with a drift guard that **stands itself down** if StyleX ever moves to `dependencies`, so it doesn't prejudge the separate open question there.

**And a real bug found underneath it.** The brief for this said "doctor has no stylex check". That's literally true of a grep, but wrong about behaviour: `checkPeerDeps` already flags it generically. What it *couldn't* do is say what to install — `missing.map(m => m.split('@')[0])` splits a scoped name on its leading `@`, so the first segment is empty and the remediation rendered as ``npm install `` **with no package name, for every scoped peer**. Fixed at the source rather than adding a second, redundant check.

Deliberately not done: escalating that check from `warn` to `fail`. Missing StyleX is a hard runtime failure under pnpm so there's an argument for it, but it changes `astryx doctor`'s CI exit code and nobody decided that.

## 4. Banner silently dropped ARIA role on an augmented status

`BannerStatus` is extensible by module augmentation, but the lookup tables were fixed records over the four built-ins.

**This is worse than the report describes.** The role going `undefined` is real — but the *icon* map has the identical defect, and `<Icon icon={undefined}>` **throws**, so an augmented-status Banner didn't render at all. The role bug was unreachable behind the crash. Both maps now fall back, so an augmented status announces as `role="status"` with the neutral icon.

Trade-off: the four built-ins lose compile-time exhaustiveness. Each keeps a role test to cover that.

## What's deliberately left out

The report has ten findings plus seven follow-ups. Three are already dead — finding 4 was never broken (it uses `console.error`, with a test), and 8 and 9 were fixed by merged #4279 and #4280 within a day of filing. The rest (the build timestamp, accent inversion in dark mode, a `--color-focus-ring` token, a success Button variant, and every follow-up item) each need a maintainer decision, and several already have one leaning against.

## How to check it

`pnpm vitest run` on the touched areas: docs 9, build-theme 19, install-line guard 5, doctor 27, Banner 36 + 48 consumers — all green. Full repo suite run alone: 382 files / 7755 tests, exit 0.

The zh Banner description is my own translation and hasn't been reviewed.

Refs #4276

### DIFF
--- a/.changeset/banner-augmented-status-role.md
+++ b/.changeset/banner-augmented-status-role.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Banner: a status added by augmenting `BannerStatusMap` no longer loses its ARIA role. The status→role and status→icon tables covered only the four built-in statuses, so an augmented status resolved to `undefined` — the banner rendered with no live-region role at all (a silent accessibility regression TypeScript could not catch), and the default icon threw on an undefined name. Both tables are now `Partial` and both reads fall back: `role="status"` and the `info` icon. (#4276)
+@AKnassa
--- a/.changeset/cli-docs-list-flag.md
+++ b/.changeset/cli-docs-list-flag.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/cli': patch
+---
+
+[fix] `astryx docs --list` no longer errors with `unknown option '--list'`. The `docs` command registered no options at all, so commander rejected the flag even though bare `astryx docs` prints exactly that list and `component --list` / `template --list` both exist. `--list` now routes to the same list branch and ignores any positional topic. (#4276)
+@AKnassa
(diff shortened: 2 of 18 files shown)


AKnassa:
Superseded upstream.

Issue #3343, opened 2026-07-02
Accessibility & keyboard-management program: build shared interaction primitives, then migrate components


cixzhang:
## Summary

This is the single tracking issue for a focused, staged program to close astryx's accessibility and keyboard-management gaps. It supersedes the ad-hoc per-finding issues: individual PRs reference this plan and each finding below.

An 8-domain audit (menus, selects/comboboxes, overlays, form controls, navigation/selection, complex widgets, infrastructure, testing) surfaced ~90 findings, adversarially re-verified. The important conclusion is **not** the length of the list — it's that almost every finding traces back to the same root cause.

## Root cause: thin, unevenly-patched shared infrastructure

astryx's entire keyboard/focus layer is four small hooks — `useFocusTrap`, `useListFocus`, `useGridFocus`, `useScrollLock` — wired differently into each component. There is:

- **no shared roving-tabindex / composite engine** — every widget that needs arrow-key navigation (Toolbar, menus, TabMenu, SegmentedControl, Calendar grid, Tokenizer chips) re-implements it, and each re-implementation has a different bug;
- **no typeahead** anywhere except the Selector family (which restarts from index 0);
- **no default focus restore** and **no modality tracking** — closing a layer routinely drops focus to `<body>`;
- **no Escape/dismissal stack** — one `Escape` closes a popover *and* its host dialog because every trap adds an unconditional document-level listener;
- **no background inerting**, yet `role="dialog" aria-modal="true"` is stamped on **every** popup (`usePopover`), so AT is told the focused control is inert while focus stays outside the popup;
- **no direction context** — horizontal arrow keys are hardcoded physical, so RTL locales navigate backwards;
- **no `scrollIntoView`** of the highlighted option, no persistent live regions, no `VisuallyHidden` primitive.

The architecture bet itself (native `<dialog>` + `showModal()`, the Popover API, native form controls) is **sound and worth keeping** — native modality, native inputs, and top-layer stacking are genuine wins. What's missing is the machinery that makes the ARIA vocabulary true per keystroke. Fixing the components one-by-one without the shared primitives just re-creates the same defects in parallel.

## Strategy

**Build the primitives once, centrally; then migrate components onto them.** Land the blocker fixes immediately (they fully exclude a user class), then invest in a small set of shared primitives, then do pattern rework on top.

### Primitives to build

| Primitive | Replaces / fixes | Notes |
|---|---|---|
[...]

PR #4410, opened 2026-07-26
fix(core): return focus to the opener when MobileNav closes

AKnassa:
## What this does

Closing the mobile navigation drawer now returns keyboard focus to the button that opened it.

## Why

`MobileNav` read `document.activeElement` **only** to decide which side the drawer slides from, then threw it away. Neither the close path nor the teardown path refocused anything, so closing stranded focus on the panel of a drawer that no longer exists.

This is one concrete instance of the umbrella issue's own finding, "closing a layer routinely drops focus to `<body>`", and a reporter hit it on Safari 26.5.2.

## What changed

Capture the opener before `showModal()` and hand it back when the native dialog closes, following the pattern `Dialog` already uses (and `BottomSheet`, after its exit animation). **No API change.**

Since #4388 the native `close()` is deferred until the slide-out has played, so the restore lives on the close itself: `closeAndRestoreFocus()` runs `dialog.close()` and then focuses the opener, and is called from the two places the close happens, the delayed timer and the unmount teardown. Inside `AppShell` the drawer sits in an `<Activity>` that flips to `mode="hidden"` on close, so that teardown is the only close it ever gets, and the only restore.

Restore is skipped when the captured element is `<body>` or has since detached: calling `body.focus()` would blur whatever a router moved focus to, and choosing a fallback target is the umbrella's blocked focus-restore-policy work. It is also skipped when the drawer is reopened before the pending close runs: the dialog never closed, and the original opener is kept.

## The ordering is load-bearing

Restore must happen **after** the native `close()`. Until then the modal dialog holds the top layer, the rest of the document is inert, and `focus()` on the opener is silently dropped. jsdom has no top layer, so `MobileNavFocus.test.tsx` pins the call order and the timing directly: `restores focus only after the native dialog has closed`, `waits for the delayed close before restoring focus`, and the AppShell and unmount-during-delay cases all assert that `focus()` runs after `close()`. Moving the restore back onto the `isOpen` flip (an effect or its cleanup) fails them.

## An honest limit

**This cannot prove it fixes the reported Safari symptom.** Per spec, `showModal()`/`close()` already record and restore focus, so *why* Safari 26.5.2 drops to `<body>` is unestablished from here. One plausible mechanism: if Safari doesn't focus a `<button>` on click, `activeElement` is already `<body>` when `showModal()` runs, in which case this captures `<body>` too and correctly no-ops, helping nothing on that path.

What's defensible regardless: the component delegated focus restore entirely to UA behaviour that demonstrably did not hold, and it now does what `Dialog` does. Real-browser verification is genuinely outstanding.

Note that jsdom has **no `HTMLDialogElement` implementation at all**, so every dialog behaviour under test is the repo's own polyfill.

## How to check it

`pnpm vitest run packages/core/src/MobileNav/ packages/core/src/AppShell/ packages/core/src/Dialog/ packages/core/src/SideNav/ packages/core/src/TopNav/`: 17 files, 508 tests. Full repo suite run alone: 627 files / 13128 tests (13 skipped), exit 0.

The focus tests were mutation-tested: restoring before `close()`, dropping the restore from either close site, capturing the opener on every open instead of only on the transition into open, and dropping either skip guard each fail at least one test. The skip guards are asserted on the `focus` call itself, because jsdom treats `<body>` as non-focusable and ignores `focus()` on detached nodes, so `activeElement` assertions cannot tell the difference.

Refs #3343


cixzhang:
Current `main` conflicts with this branch in MobileNav's close/focus path. It is not a mechanical merge: resolving the hunk decides whether focus restoration runs before or after the delayed native close.

Please merge current `main`, choose and test that ordering, and push the resolved head; the next sweep will re-review it.

[Reviewed by Robohands]

AKnassa:
Merged current main and resolved the hunk as restore after the delayed close: `closeAndRestoreFocus()` runs `dialog.close()` and then focuses the opener, called from the close timer and from the unmount teardown that AppShell's `<Activity>` path takes. The keep-both resolution (restore in the `isOpen` cleanup) fails `restores focus only after the native dialog has closed`; the file now also pins the timing and the reopen and unmount-during-delay cases (12 tests). Resolved head is 4e77eee06.

Issue #958, opened 2026-03-29
Edge compensation: deriving spacing/radius/size relationships from visual intent


cixzhang:
## Problem

Theme authors think in **visual features** — concentricity, centered alignment, visually aligned content — not raw token values. But today, tuning a theme means manually juggling `radius`, `spacing`, `size`, and `font-size`/`line-height` values that are deeply interdependent.

**Edge compensation** is the umbrella term for the math that connects these values. Getting it right means theme owners can express high-level visual intent and have the system derive correct concrete values.

## Why this is hard

These properties constrain each other:

- **Concentricity** — inner radius must account for the padding between nested rounded rects (`inner-radius = outer-radius - gap`), but "gap" itself depends on spacing tokens
- **Centered alignment** — vertical centering of text inside a sized container depends on `line-height`, `font-size`, *and* the container's padding, not just `align-items: center`. Often times we want to avoid using center alignment directly items because we want items to appear vertically centered when there's only one line of content, but remain top aligned when there's multiple lines of content. However, in cases where we're guaranteed a single line, center alignment can be achieved in a much more straightforward manner, though edge compensation may still be required where content overflows padded bounds.
- **Visual alignment** — content edges across different components (e.g. a label next to an input) need to land on the same optical line, which means their padding + border + radius crop all interact

One knob turns and three others need to move with it.

## Goal

Tease out the **conceptual requirements** from the theme author ("I want concentric corners", "I want text vertically centered in this pill") and translate those into the mathematical relationships between tokens. Then encode those relationships so the system can enforce them automatically.

## Collecting edge cases

Use this issue to gather specific scenarios where edge compensation matters. We'll look broadly across them later to find the common patterns.

---

_Drop examples, screenshots, or sketches of tricky compensation cases in the comments._

cixzhang:
## Edge case: icon buttons in padded containers (from #947)

#947 catalogued a concrete instance of edge compensation: icon-only buttons inside padded containers (dialog headers, toolbars, card headers) where the button's intrinsic padding stacks with the container's padding, creating visual imbalance.

### Problem scenarios from #947

- Dialog header close button
- Card header actions
- TopNav end content
- Table row action menus
- Toolbar icon buttons at edges

### Why the #947 solutions were too narrow

[...]

PR #4412, opened 2026-07-26
fix(docsite): make the gigantic preset concentric

AKnassa:
## What this does

Fixes the theme editor's `gigantic` preset so its corner radii nest correctly, and adds the warning banner that tells you when they don't.

## Why — and a correction to the issue

Both of these were **reported as already done** in the issue on 2026-04-18. Neither fully landed.

Checking the code rather than the comment: `git log -S` shows all four presets were introduced in a single commit with `comfortable` **already matched** at 6/6 and `gigantic` at 8/12. So the claim that "the `comfortable` and `gigantic` presets had divergent bases — we've adjusted them to match" is wrong about `comfortable`, which never diverged, and simply never happened for `gigantic`. Only one preset was ever broken. And `git grep -i concentric` across the docsite returns **zero hits** — the warning banners were never written.

## The relation, derived from the code

The radius scale is `inner = base×1`, `element = ×2`, `container = ×3`, and the editor nests exactly one spacing step between rungs. So concentricity (`inner = outer − gap`) gives:

- container rung: `3·r − s = 2·r`
- element rung: `2·r − s = 1·r`

Both collapse to the same condition: **`radiusBase === spacingBase`**.

| preset | spacing | radius | container − gap | holds? |
|---|---|---|---|---|
| compact | 2 | 2 | 6 − 2 = 4 = element | yes |
| default | 4 | 4 | 12 − 4 = 8 = element | yes |
| comfortable | 6 | 6 | 18 − 6 = 12 = element | yes |
| **gigantic** | 8 | **12** | 36 − 8 = **28 ≠ 24** | **no** |

(Concentricity doesn't extend to the `page` rung, which is a ×7 jump — the issue only ever claimed the two, so it's right about that.)

## What changed

`gigantic.radius` 12 → 8, plus the warning banner beneath both sliders when the bases diverge — written as a pure function so it's testable, since the docsite's vitest config is node-only and matches `.ts` alone, so no component there can render.

The Radius XL toggle option moves 12 → 8 with it. Not scope creep: the S/M/L/XL option lists mirror the presets exactly, so leaving it would make clicking "Gigantic" highlight XL on every control **except** radius, which would show no selection at all.

## This one needs a maintainer's sign-off

**Radius 12→8, not spacing 8→12.** Either satisfies the relation. Spacing is a clean +2 run (2, 4, 6, 8), so matching radius to it makes the two columns identical — which is what "all presets observe concentricity" means structurally. Raising spacing instead would push `--spacing-12` to 144px.

But it does change what you see: clicking Gigantic now gives radii of 8/16/24/56 instead of 12/24/36/84. Corners get visibly tighter. Two things soften that — nothing in `packages/` consumes these presets, so no shipped theme or component changes, and it's a starting point in an interactive editor rather than a token contract; and anyone who wants the old look can still type 12, at which point the new banner correctly tells them concentricity is broken, which is the educate-over-enforce behaviour that was asked for.

Still: a theme someone previously **exported** from the Gigantic preset is no longer reproducible by clicking the same button. And this rests on a designer's comment that, as shown above, is demonstrably wrong on the facts. **The banner half is purely additive and safe; the preset value deserves an engineering maintainer's eye.** I'd have split the commit, but the two halves entangle in the same panel file and share a test.

## How to check it

`pnpm -F @astryxdesign/docsite test` → 19 files, 292 tests, green. Note the root suite doesn't cover `apps/**` at all — neither vitest project includes it — so that's the command that matters here.

No changeset: `apps/docsite` is private and never published.

**Not visually checked.** The banner uses the repo's existing warning affordance, but it's somewhat tall for that narrow left panel when both warnings show at once.

Two things found and deliberately left alone: the S/M/L/XL option arrays duplicate the preset values with no test binding them, so this drift can recur; and `applyUnifiedPreset` and the size-match check disagree, so the Element Size control reads "unset" right after applying any preset except `default`.

Refs #958

### DIFF
--- a/apps/docsite/src/__tests__/theme-concentricity.test.ts
+++ b/apps/docsite/src/__tests__/theme-concentricity.test.ts
@@ -1,0 +1,84 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+import {describe, it, expect} from 'vitest';
+import {expandRadiusScale} from '@astryxdesign/core/theme';
+import {UNIFIED_PRESETS} from '../app/playground/themeEditor/constants';
+import {
+  buildSpacingScale,
+  getConcentricityWarning,
+} from '../app/playground/themeEditor/helpers';
+
+const px = (value: string): number => parseInt(value, 10);
+
+/**
+ * Concentricity: two rounded rects nested with a gap between their edges look
+ * concentric when `inner-radius = outer-radius - gap`. The theme editor nests
+ * one spacing step (`--spacing-1`) between each radius rung, so the relation is
+ * checked against the scales the editor actually generates rather than a
+ * hardcoded table of expected px values.
+ */
+describe('theme editor presets — concentricity', () => {
+  for (const [name, preset] of Object.entries(UNIFIED_PRESETS)) {
+    it(`${name} preset nests its radius rungs concentrically`, () => {
+      const radii = expandRadiusScale({base: preset.radius, multiplier: 1});
+      const gap = px(buildSpacingScale(preset.spacing)['--spacing-1']);
+      const inner = px(radii['--radius-inner']);
+      const element = px(radii['--radius-element']);
+      const container = px(radii['--radius-container']);
+
+      // container ⊃ element
+      expect(container - gap).toBe(element);
+      // element ⊃ inner
+      expect(element - gap).toBe(inner);
+    });
+  }
+});
+
+describe('getConcentricityWarning', () => {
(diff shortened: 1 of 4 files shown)


ernestt (approved):
Yes this is the correct value and approach! ♥️ Thanks for noticing and fixing it!

cixzhang (approved):
Ernest approved the values a month ago and nothing has moved under it since — re-checked against main today. This was stuck on the eng review gate, not on the work.

[Reviewed by Robohands]

PR #4413, opened 2026-07-26
fix(lab): render list markers in RichTextEditor (bullets/numbers)

potatowagon:
## What

Fixes bulleted and numbered lists in the experimental `RichTextEditor` (`@astryxdesign/lab`) rendering as plain indentation with no bullet/number marker.

Repro: https://facebook.github.io/astryx/storybook/?path=/docs/lab-richtexteditor--docs — type `- ` or `1. ` and the list indents but shows no marker.

## Root cause

`editorTheme.ts` styled lists with only `paddingInlineStart`; it never set `list-style-type`. StyleX ships a `list-style-type: none` reset, so the `<ul>`/`<ol>` got the indent (padding) but the marker was suppressed. Confirmed in the compiled `lab.css`.

## How

- Split the single shared `list` style into distinct `ul` (`list-style-type: disc`) and `ol` (`list-style-type: decimal`) styles, each with `list-style-position: outside`.
- Added per-depth nested markers (nested `<ul>` → `circle`) wired through Lexical's `ulDepth`/`olDepth` theme API (verified against the `@lexical/list` source, which reads `listTheme[`${tag}Depth`]`).
- Set `list-style-type: inherit` on the list item so item-level resets don't re-hide the marker.
- Shared by both `RichTextEditor` and `RichTextView` (both consume `sharedEditorTheme()`), so the read-only view is fixed too.

## Testing

- `eslint` — clean
- Added two unit tests asserting the browser computes `list-style-type: disc` for `<ul>` and `decimal` for `<ol>` (StyleX `runtimeInjection` is on in tests, so this validates real rendered marker styles, not just class presence).
- Full `RichTextEditor.test.tsx` suite: 18/18 passing.
- Verified `lab.css` now emits `list-style-type: disc / decimal / circle / inherit` (previously absent).

**human (manual)**
before
https://facebook.github.io/astryx/storybook/?path=/docs/lab-richtexteditor--docs
Screenshot 2026-07-27 at 8 14 54 AM

after:
(http://localhost:6006/?path=/docs/lab-richtexteditor--docs , vercel preview blocked by perms)

Screenshot 2026-07-27 at 8 14 48 AM

### DIFF
--- a/packages/lab/src/RichTextEditor/RichTextEditor.test.tsx
+++ b/packages/lab/src/RichTextEditor/RichTextEditor.test.tsx
@@ -21,6 +21,55 @@ import {RichTextView} from './RichTextView';
 
 // A minimal valid serialized Lexical editor state containing a single
 // paragraph with the text "Hello world".
+// Builds a serialized Lexical state containing a list. `listType` is
+// 'bullet' (renders <ul>) or 'number' (renders <ol>). Used to verify that the
+// editor theme applies visible list markers rather than bare indentation.
+function makeListState(listType: 'bullet' | 'number'): string {
+  const tag = listType === 'number' ? 'ol' : 'ul';
+  return JSON.stringify({
+    root: {
+      children: [
+        {
+          children: [
+            {
+              children: [
+                {
+                  detail: 0,
+                  format: 0,
+                  mode: 'normal',
+                  style: '',
+                  text: 'Item one',
+                  type: 'text',
+                  version: 1,
+                },
+              ],
+              direction: 'ltr',
+              format: '',
+              indent: 0,
+              type: 'listitem',
+              version: 1,
+              value: 1,
+            },
+          ],
+          direction: 'ltr',
+          format: '',
+          indent: 0,
+          type: 'list',
+          version: 1,
+          listType,
+          start: 1,
+          tag,
+        },
+      ],
+      direction: 'ltr',
+      format: '',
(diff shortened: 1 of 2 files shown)


nynexman4464 (approved):
**Approve with nits.** Fix is correct: split the shared `list` style into `ul`/`ol`, add `list-style-type: inherit` on the `li` to beat StyleX's reset, wire nested markers via `ulDepth`/`olDepth`. Also covers read-only `RichTextView`.

Nits (non-blocking):
- No nested-list test.
- Only 2 depth levels (disc→circle); missing `square`, `olDepth` stays decimal.
- `listTheme.checklist` unset if the toolbar exposes checklists.

potatowagon:
Thanks for the review! Addressed all three nits in 6ee868012:

1. **Nested-list test** — added `makeNestedBulletState()` + a test asserting depth-2 renders a `circle` marker (not bare indentation).
2. **Depth levels** — extended `ulDepth`/`olDepth` to 3 entries: `disc→circle→square` and `decimal→lower-alpha→lower-roman`. Lexical indexes these arrays by `depth % length`, so three levels cycle cleanly, matching native browser nesting.
3. **Checklist** — added a `checklist` theme slot (`list-style-type: none`) so check lists from markdown shortcuts (`- [ ] `) or serialized content don't fall back to bare bullet styling.

Tests (19 passing), `tsc --noEmit`, and eslint all green.

Issue #808, opened 2026-03-22
feat(storybook): Theme Editor — spacing usage context for components


rubyycheung:
## Summary

The Theme Editor needs five refinements to help theme authors understand the impact of their changes — plus a cross-cutting requirement that every editing action provides a **live component preview** so users can visually verify their changes.

1. **Spacing usage context** — show which components consume each spacing token
2. **Concentricity awareness** — warn when spacing + radius combos break concentric alignment
3. **Component override visibility** — surface when `defineTheme` component overrides diverge from system defaults
4. **Token impact highlighting** — select a token to see every affected element light up
5. **Color palette import & AI mapping** — load brand colors and auto-map to XDS tokens

---

### Principle

Every section below implies a change to the theme — and **every change must be immediately verifiable in a live component preview**. The preview panel isn't a nice-to-have for one or two sections; it's the core feedback loop for the entire editor. Users should never have to export a theme, apply it, and load Storybook just to see what changed.

### Requirements

- **Always-visible preview panel**: The component preview must remain visible alongside any editing view (spacing, radius, overrides, palette import, etc.) — not hidden behind a tab.
- **Comprehensive component coverage**: The preview should include a representative sample of all major component categories so that any token change has at least one visible component to validate against:
  - **Buttons** (primary, secondary, destructive — all sizes)
  - **Form controls** (TextInput, Switch, Checkbox, Radio, Slider, Select)
  - **Cards** with nested content (to verify spacing + radius + concentricity)
  - **Badges, Tokens, Banners** (all sentiments)
  - **Tabs, Breadcrumbs, Navigation** elements
  - **Dialogs/Popovers** (overlay components)
  - **Tables** (data-dense context)
  - **Avatars, Dividers, Progress bars**
- **Instant updates**: Changes in the editor panel must reflect in the preview with no perceptible delay — no "apply" button needed.
- **Before/after comparison**: A toggle or split-view to compare the current edits against the original theme defaults, so users can see exactly what changed.
- **Dark/light mode toggle**: Preview should support switching modes so users can verify both.

---

### Problem

When a theme author adjusts `--spacing-4` from `16px` to `20px`, they have no idea that this change affects:
- **Card** inner padding
- **Container** default padding (large)
- **Stack** gap at the `4` level

Similarly, changing `--spacing-3` (`12px`) impacts smaller containers, and `--spacing-2` (`8px`) affects input field padding — but none of this is surfaced in the editor.

### Proposal

[...]

PR #4415, opened 2026-07-27
feat(docsite): spacing usage context in Theme Editor

AKnassa:
Refs #808. Implements **section 1 only** ("Spacing usage context"). Sections 0, 2, 3, 4 and 5 are not included here; see "What is not in this PR" below.

## What this does

Every spacing token row in the Theme Editor now shows which components a change to that token will actually move. Editing `--spacing-4` shows up front that it drives Card, Dialog and Section padding.

## Why

Without semantic spacing names, the editor is the only place that explains what a scale value means. This puts that information next to the control instead of leaving the author to change a number and scan for differences.

## What changed

- Spacing rows under **Advanced > Spacing** gained a second line naming the affected components, with the full list in a tooltip.
- A build step derives that map from `packages/core` source and writes it into `src/generated/`, next to the existing generators. It runs as part of `pnpm generate`.
- The size group shares the same row component and is deliberately unaffected.

## The map is derived, not curated

The issue proposes maintaining a static `SPACING_USAGE_MAP`. A hand-authored component/token map in this repo has drifted twice already, which is the stated reason `packages/core/src/theme/themingTargets.test.ts` exists. Generating this one at build time removes that failure mode: it cannot disagree with the components it describes.

Attribution follows what a component applies rather than where a file lives: shared style objects and hook-rendered markup (`useKeyboardHint`, `useEntryAnimation`) credit the components that mount them, through aliased imports and re-export chains. A derivation that comes back empty fails the build instead of writing a blank map.

## The table in the issue does not match the source

Three corrections, each now covered by a test:

| Issue says | Source says |
|---|---|
| `Container` padding uses `--spacing-3` / `--spacing-4` | There is no `Container` component. `Layout/container.stylex.ts` is an internal utility, and "Container" is also an anatomy label in `Card.doc.mjs`. |
| Field gap uses `--spacing-2` | `Field.tsx:39` uses `--spacing-1`. The `--spacing-2` gap belongs to the input wrapper, a different element. |
| Section spacing uses `--spacing-6` | Section's padding chain terminates at `--spacing-4` (`container.stylex.ts:108`). |

## One design decision worth a look

Usage is reported as two groups rather than one list:

- **components** the token sits in a fixed style, so any change to it moves them
- **viaProps** the token is one rung of a prop-keyed scale, so a change moves them only where that prop value is passed

As a single list this is accurate but unusable: `Layout` appears under all fifteen rungs and `--spacing-2` under 59 component directories, because components exposing a numeric `gap` or `padding` prop enumerate the whole scale in one `stylex.create` call. Roughly a third of all spacing references live in seven such tables.

The split also makes "no component uses this step by default" sayable, which is the accurate answer for `--spacing-9` and `--spacing-10`.

## What is not in this PR

| Section | Status |
|---|---|
| 0. Live component preview | The playground already renders a live preview beside the editor and updates it on every token edit. The remaining asks (always-visible, before/after split, broader component coverage) are product decisions. |
| 2. Concentricity awareness | Overlaps #4412, which adds concentricity warnings to this same editor. |
| 3. Component override visibility | The changed-vs-default comparison is already computed at export time in `helpers.ts` but never surfaced. Surfacing it is a UI decision. |
| 4. Token impact highlighting | The preview is a separate document in an iframe, so highlighting elements from the editor panel needs a design decision first. |
| 5. Colour palette import and AI mapping | Closest prior art, #585, was closed unmerged. |

## How to see it

```
pnpm -F "@astryxdesign/theme-*" build
pnpm -F @astryxdesign/docsite dev
```

Then open `/playground`, switch to the Theme Editor in the left rail, and choose **Advanced > Spacing**. Hover any row's second line for the full component list.

## Tests

`pnpm -F @astryxdesign/docsite test` gives 19 files and 326 tests, of which 37 are new. They cover the derivation against live core and against a synthetic fixture tree (aliased imports, aliased re-exports, hook modules, root-level files), both spellings of the prop-keyed scale tables, the chained `var()` fallbacks behind Card/Section/Dialog default padding, attribution of shared input styles and hook-rendered spacing (`useKeyboardHint`, `useEntryAnimation`) to the components that apply them, an inverse text-search check that fails if any component's spacing stops being attributed, the emitted module shape with its empty-map guard, a relative-source-dir regression pin, and the three corrections above.

No changeset: `apps/docsite` is private, and `check:changesets` passes without one.


ernestt (changes requested):
I think there might be a better way to show where spacing tokens are used instead of just listing them out next to the token names.  Fundamentally I think this is trying to solve for the confusion issue where users think the theme editor is only adjusting what is in preview rather than globally.   We are trying find ways to clarify this common confusion but dont have one yet.  

fyi @kentonquatman

PR #4416, opened 2026-07-27
feat(lab): expose transformers prop on RichTextEditor

potatowagon:
## What

Adds a `transformers?: ReadonlyArray<Transformer>` prop to the experimental `RichTextEditor` (`@astryxdesign/lab`), defaulting to the standard `@lexical/markdown` `TRANSFORMERS`.

## Why

The markdown transformer set was hardcoded to `TRANSFORMERS`. Consumers who register custom nodes via the existing `nodes` extension point had no way to add matching markdown transformers (for their custom node types) without forking the component.

Importantly, in Lexical the transformers array is the **single source of truth for all markdown behaviour** — not just shortcut typing. Verified against `facebook/lexical`'s `lexical-playground`, where one `PLAYGROUND_TRANSFORMERS` array feeds all three markdown operations:

- shortcut typing — `registerMarkdownShortcuts(editor, PLAYGROUND_TRANSFORMERS)` (`MarkdownShortcutsExtension/index.ts`)
- markdown → state (import) — `$convertFromMarkdownString(..., PLAYGROUND_TRANSFORMERS, ...)` (`ActionsPlugin/index.tsx`)
- state → markdown (export) — `$convertToMarkdownString(PLAYGROUND_TRANSFORMERS, ...)` (`ActionsPlugin/index.tsx`)

So this prop is deliberately introduced as that shared primitive. It wires shortcut typing today; the upcoming `getMarkdown()` / serializer APIs will read the same prop, so a consumer's custom transformers stay consistent across shortcuts, import, and export.

This is one of the general-purpose gaps identified for EPS `RichTextArea` → Astryx `RichTextEditor` parity — low-risk and additive.

## How

- New `transformers` prop, defaults to `TRANSFORMERS`. Only applied to shortcut typing when `hasMarkdownShortcuts` is true (the plugin already gated on that flag).
- Spread into `MarkdownShortcutPlugin` (`[...transformers]`) since the plugin expects a mutable array while the prop is `ReadonlyArray` for a safer public API.
- Imported `type Transformer` from `@lexical/markdown`.
- JSDoc + `doc.mjs` spell out the single-source-of-truth intent (shortcuts / import / export) so later phases wire the same prop rather than re-hardcoding `TRANSFORMERS`.

- `RichTextEditor.doc.mjs` — added `transformers` to the props table; clarified `hasMarkdownShortcuts`.
- `RichTextEditor.stories.tsx` — added a `CustomTransformers` story (restricted markdown: bold/italic/unordered-list only).
- `RichTextEditor.test.tsx` — added tests for a custom transformers array and for `hasMarkdownShortcuts={false}`.

## Testing

- `eslint` on changed files — clean
- `vitest run RichTextEditor.test.tsx` — 18/18 passing (incl. 2 new)

**human manual test**

https://github.com/user-attachments/assets/93c702d7-ce66-4ac6-a4cc-69905ce59c6a

### DIFF
--- a/apps/storybook/stories/RichTextEditor.stories.tsx
+++ b/apps/storybook/stories/RichTextEditor.stories.tsx
@@ -4,6 +4,7 @@ import {useState} from 'react';
 import type {Meta, StoryObj} from '@storybook/react';
 import {RichTextEditor, RichTextView} from '@astryxdesign/lab';
 import type {EditorState} from 'lexical';
+import {BOLD_STAR, ITALIC_STAR, UNORDERED_LIST} from '@lexical/markdown';
 
 const meta: Meta<typeof RichTextEditor> = {
   title: 'Lab/RichTextEditor',
@@ -47,6 +48,16 @@ export const Required: Story = {
     label: 'Summary',
     isRequired: true,
     placeholder: 'Required field',
+  },
+};
+
+export const CustomTransformers: Story = {
+  args: {
+    label: 'Comment',
+    description:
+      'Restricted markdown: only `*bold*`, `_italic_` and `- ` unordered lists (no headings, quotes or code).',
+    placeholder: 'Try typing "# " — it will not become a heading…',
+    transformers: [BOLD_STAR, ITALIC_STAR, UNORDERED_LIST],
   },
 };
 
--- a/packages/lab/src/RichTextEditor/RichTextEditor.doc.mjs
+++ b/packages/lab/src/RichTextEditor/RichTextEditor.doc.mjs
@@ -93,8 +93,15 @@ export const docs = {
       name: 'hasMarkdownShortcuts',
       type: 'boolean',
       description:
-        'Enable Markdown shortcut typing (e.g. "# " for a heading). Uses default @lexical/markdown transformers.',
+        'Enable Markdown shortcut typing (e.g. "# " for a heading). Uses the transformers prop (defaults to the standard @lexical/markdown transformers).',
       default: 'true',
+    },
+    {
+      name: 'transformers',
+      type: 'ReadonlyArray<Transformer>',
+      description:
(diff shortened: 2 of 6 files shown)


nynexman4464 (approved):
**Approve with nits.** Defaults to Lexical's `TRANSFORMERS`, backward compatible. Good.

- Fix before landing: `[...transformers]` re-allocates each render → `MarkdownShortcutPlugin` re-registers every re-render. Use `useMemo(() => [...transformers], [transformers])`. (Confirmed React Compiler isn't running the transform anywhere in the repo — only `eslint-plugin-react-compiler` is installed, not `babel-plugin-react-compiler` — so this isn't auto-memoized.)
- Re-export `type Transformer` from the lab barrel.
- Tests cover rendering but not shortcut behavior.

potatowagon:
Thanks for the review! All three nits addressed in `ce887f5`:

1. **Re-registration** — `MarkdownShortcutPlugin` now takes a `useMemo(() => [...transformers], [transformers])` value instead of a fresh `[...transformers]` per render, so it no longer re-registers on every re-render. (Confirmed the same as your note — no `babel-plugin-react-compiler` in the tree, so nothing was auto-memoizing this.)
2. **Barrel export** — `type Transformer` is now re-exported from `@astryxdesign/lab` (via `RichTextEditor/index.ts` and the top-level barrel), so consumers can type custom transformer arrays without reaching into `@lexical/markdown`.
3. **Shortcut behavior** — added two behavioral tests: the default `TRANSFORMERS` convert `# Title` into a `heading` node, and an empty transformers array leaves it as a plain `# Title` paragraph. Drives `$convertFromMarkdownString` through the mounted editor since jsdom can't dispatch the live keystrokes.

Local: 20/20 lab tests pass, typecheck + lint clean.

PR #4417, opened 2026-07-27
feat(lab): add imperative ref API to RichTextEditor

potatowagon:
## What

Adds an imperative `ref` API to the experimental `RichTextEditor` (`@astryxdesign/lab`) via a new `RichTextEditorRef` handle:

- `focus()` — move focus into the editable surface
- `clear()` — empty the content (reset to a single empty paragraph)
- `getEditorState()` — read the current `EditorState`
- `getEditor()` — access the underlying `LexicalEditor` for advanced use

## Why

Today the only way to imperatively drive the editor (focus it, clear it, read its state, reach the Lexical instance) is to write a custom plugin that calls `useLexicalComposerContext()` and pass it via the `plugins` slot. That's boilerplate for common operations. This exposes a first-class `ref`, matching the ergonomics consumers expect from other Astryx inputs.

One of the general-purpose gaps identified for EPS `RichTextArea` → Astryx `RichTextEditor` parity (EPS ships `focus`/`clear`/`getEditorState`/`getEditor` on its ref). Low-risk and additive.

## How

The `LexicalEditor` instance only exists *inside* the `LexicalComposer`, so the handle can't be wired from the outer component directly. A small internal `EditorRefBridge` plugin (rendered inside the composer) reads the editor via `useLexicalComposerContext()` and connects it to the outer `ref` with `useImperativeHandle` — the same pattern as the existing `AutoFocusOnMount` plugin.

- Component wrapped in `forwardRef<RichTextEditorRef, RichTextEditorProps>`.
- `clear()` uses `editor.update(() => $getRoot().clear())`.

## Docs / exports / tests (SYNC contract)

- Exported `RichTextEditorRef` from `RichTextEditor/index.ts` and the `@astryxdesign/lab` barrel.
- Updated the component JSDoc example to show ref usage; added a `doc.mjs` best-practice note.
- Added 6 unit tests (handle shape after mount, focus wiring, `getEditorState`, `getEditor`, `clear`).

## Testing

- `eslint` on changed files — clean
- `vitest run RichTextEditor.test.tsx` — 21/21 passing (incl. 6 new)

**human test**

https://github.com/user-attachments/assets/2a60b555-d43a-45bd-9b35-bec23da073c6

## Notes

- The `focus()` test asserts the call is wired + the root element is reachable rather than jsdom `document.activeElement`, because jsdom does not reliably reflect programmatic contenteditable focus (same limitation the existing autofocus path has).
- Independent of #4416 (transformers) — different surface, no file overlap; both branch off `main`.


nynexman4464 (approved):
**Approve with nits.** `forwardRef` correct, handle memoized on `[editor]`, type exported from component index + lab barrel. Good coverage + story.

- Resolve or acknowledge before landing: `clear()` calls `$getRoot().clear()`, leaving zero children, but docs say "single empty paragraph." Use `$getRoot().clear(); $getRoot().append($createParagraphNode())` + assert `getChildrenSize() === 1`.
- `clear()`/`focus()` aren't gated on `isReadOnly`/`isDisabled`.
- No test asserts `clear()` fires `onChange`.

potatowagon:
Thanks for the review! All three nits addressed in 05f3e572e:

1. **`clear()` now leaves a single empty paragraph.** Changed to `$getRoot().clear(); $getRoot().append($createParagraphNode())`, matching the documented behaviour. Added a test asserting `getChildrenSize() === 1` after clear.
2. **`clear()`/`focus()` are gated on editability.** Both no-op when `isReadOnly` or `isDisabled` (bridge now takes an `editable` prop). Added read-only + disabled no-op tests.
3. **Added a test asserting `clear()` fires `onChange`.**

Also synced the JSDoc + `doc.mjs` per the file's SYNC block. `pnpm test` (24 passing), typecheck, and lint all green.

potatowagon:
Update on the `build-sandbox` failure (commit 329324a):

My first attempt (a webpack `module.rule` keyed on the lexical resource) did **not** actually work — I reproduced the build in Next's bundled webpack and confirmed lexical still resolved to raw `src/*.ts`. Root cause: a rule's `resolve.conditionNames` govern how the **matched module resolves its own imports**, so when `@astryxdesign/lab` imports `lexical` the resolution runs in *lab's* context (the global `source` condition) and picks lexical's raw-TS `source` export. A rule keyed on the lexical resource never gets a chance to override that.

**Actual fix:** keep the global `source` condition (so `@astryxdesign/*` still builds from source, same as `main`), and pin `lexical` + every installed `@lexical/*` package — including every `exports` subpath — to its prebuilt **dist** file via exact-match `resolve.alias` entries. Aliases are resolved through Node's own resolver (which ignores the webpack-only `source` convention and honors exports-map renames like `./ReactProviderExtension` → `dist/LexicalReactProviderExtension.mjs`), and win over export-condition resolution regardless of which package issued the import.

Verified end-to-end in Next's actual bundled webpack: the RichTextEditor's full lexical import graph (56 modules) resolves entirely to dist with **no build errors**. Added a 7-test `next.test.mjs` regression guard. Also restored the `@potatowagon` credit line the changeset lost in a rebase.

potatowagon:
Correction + real fix (commit fb041dd):

My earlier `withAstryx`/`next.js` changes were the **wrong approach** — I've fully reverted them (and their test + changeset); `next.js` is back to the baseline. Comparing against the passing sibling PR #4502 (`feat(lab): add maxLength character counter`) made the real cause obvious: it fixes the identical `build-sandbox` failure *in the component*, not in the build config.

**Root cause:** the imperative `clear()` used `$getRoot` / `$createParagraphNode` imported from the **top-level `lexical`** package — a *runtime value* import. When the sandbox's Next build resolves `@astryxdesign/lab`'s built dist, a top-level `lexical` value import forces Babel to transpile lexical's raw `src/*.ts` (which uses `declare` class fields) and fails. `@lexical/*` subpath value imports are fine — only the bare `lexical` value import leaks.

**Fix:** `clear()` now resets via `editor.setEditorState(editor.parseEditorState(EMPTY_STATE))` — both editor-instance methods, so no top-level `lexical` value import is needed. It still resets to a single empty paragraph and fires `onChange`. Only the compile-time-erased `import type { … } from 'lexical'` remains.

Verified: built `@astryxdesign/lab` dist has zero bare-`lexical` value imports (only `@lexical/*` subpaths); all 31 RichTextEditor tests pass; lab typecheck + lint clean.

PR #4419, opened 2026-07-27
fix(HoverCard): theme target + consumer style forwarding

cixzhang:
## What

Fixes two issues found during component audit of HoverCard:

1. **themeProps on wrong element** -- the `.astryx-hovercard` className was placed on the inner content `<span>`, but the visual surface styles (background, border-radius, box-shadow) live on the layer container element. Themes targeting `.astryx-hovercard` could not reach the visual surface. Moved to the layer's `className` prop, matching how Tooltip already does it.

2. **Consumer escape hatches silently dropped** -- `xstyle`, `className`, and `style` passed via `HoverCardProps` were forwarded into `renderHoverCard` but never applied to any element. Now merged onto the content span via `mergeProps` so consumers can customize content padding and inner layout.

## Risk

Low. The themeProps className moves from an inner span to the outer layer container. No existing theme packages target `.astryx-hovercard` today (verified via grep across all theme packages). Consumer style props that were previously dropped now take effect, which could surface in a consumer already working around the gap via a wrapper; unlikely given this was undocumented.

## Testing

- `pnpm vitest run packages/core/src/HoverCard/` -- 24 tests pass
- `npx tsc --noEmit` -- clean
- ESLint -- clean

### DIFF
--- a/packages/core/src/HoverCard/useHoverCard.tsx
+++ b/packages/core/src/HoverCard/useHoverCard.tsx
@@ -34,7 +34,6 @@ import {
   radiusVars,
   spacingVars,
 } from '../theme/tokens.stylex';
-import {mergeProps} from '../utils';
 import {themeProps} from '../utils/themeProps';
 
 const styles = stylex.create({
@@ -469,6 +468,7 @@ export function useHoverCard(options: HoverCardOptions = {}): HoverCardReturn {
       props?: Omit<ContextRenderProps, 'positioning'>,
     ): ReactNode => {
       const renderPlacement = props?.placement ?? placement;
+      const themeClassName = themeProps('hovercard').className;
       const renderProps = {
         placement: renderPlacement,
         alignment: props?.alignment ?? alignment,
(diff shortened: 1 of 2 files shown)


packages/core/src/HoverCard/useHoverCard.tsx
```diff
@@ -478,14 +478,21 @@ export function useHoverCard(options: HoverCardOptions = {}): HoverCardReturn {
...
         as: 'span' as const,
       };
 
       return layer.render(
         <span
-          {...mergeProps(themeProps('hovercard'), stylex.props(styles.content))}
+          {...mergeProps(
+            stylex.props(styles.content, props?.xstyle),
+            props?.className,
+            props?.style,
```
cixzhang: Should these move as well?

cixzhang: Good call — moved them. The consumer style props (`xstyle`/`className`/`style`) now land on the **layer container** (via `renderProps`) rather than the inner content span, so surface customization targets the same element as the theme class and the visual surface (bg/radius/shadow):

- `xstyle` → appended to the container array: `[popoverXstyle, layerAnimations[renderPlacement], props?.xstyle]`
- `className` → merged with the theme class: `` `${themeProps('hovercard').className} ${props.className}` ``
- `style` → passed straight through as `renderProps.style` (the layer's `ContextRenderProps` accepts a raw `style` and spreads it last, so consumer inline styles win over positioning)

The inner `<span>` keeps `styles.content` for padding and the hover/focus handlers. `mergeProps` is no longer needed here since the container's `render` already composes className/style. All 24 HoverCard tests pass.

PR #4444, opened 2026-07-28
fix(docsite): render code spans in link labels; auto-link component names in doc prose (#4425)

### LINKED ISSUE #4425: docsite: support code spans inside link labels in inline markdown ([`code`](href))
## Summary

We want API references in docs prose — component names and hook names like `InternationalizationProvider`, `useTranslator`, `useDirection` — to render as **monospace** *and* as **links** to their doc pages (e.g. `/components/InternationalizationProvider`), without authors having to hand-write the link every time.

## Preferred approach: reuse the changelog's auto-linker

The changelog page already solves the "turn bare component names into links" problem. `linkifyComponents` in `apps/docsite/src/components/changelogLinkify.ts` takes raw markdown plus the list of known component names and rewrites bare occurrences into `[Name](/components/Name)` links, using word boundaries and lookarounds so it skips names already inside backticks or existing link brackets:

```ts
const pattern = new RegExp(
  '(?<!`|\\[)\\b(' + escaped.join('|') + ')\\b(?!`|\\])',
  'g',
);
```

The component name list is available at build time (`ChangelogPage` passes `componentNames` from the generated `componentRegistry`).

**The ask:** run the same `linkifyComponents` pass over `.doc.mjs` prose before it hits the inline renderer (`renderInlineMarkdown` in `apps/docsite/src/components/docs/inlineMarkdown.tsx`), wired with the same `componentNames` list. Authors then just write `InternationalizationProvider` as plain text and it auto-links — no new syntax, and the behavior stays consistent with the changelog.

This is the better path than inventing a new authoring convention: one shared linkifier, one mental model.

## Secondary enhancement: allow code spans inside link labels

Independently, the inline renderer can't currently render a code span **inside** a link label. `renderInlineMarkdown` uses a flat regex alternation:

```js
const TOKEN = /(`([^`]+)`|\[([^\]]+)\]\(([^)]+)\))/g;
```

A link's label is captured as raw text (`match[3]`) and passed to `renderLink` as a plain string, so `` [`<InternationalizationProvider>`](/components/InternationalizationProvider) `` shows literal backticks instead of monospace.

If we want the auto-linker to produce monospace links (i.e. emit `` [`Name`](/components/Name) `` and have it render as monospace), or if authors ever want to hand-write a monospace link, the renderer needs to detect a label of the form `` `...` `` and render it via the `<Code>` component inside the `<a>`. Minimal scope — just the common code-in-link case, not full nested-markdown parsing.

Note the interaction: `linkifyComponents` currently wraps matches as `[Name](href)` (plain label) and deliberately skips names already in backticks. If we want monospace auto-links, we'd tweak the linkifier to emit `` [`Name`](href) `` **and** land this renderer change so the code-in-label renders correctly. The two pieces compose.

## Acceptance criteria

- Bare component/hook names in `.doc.mjs` prose auto-link to their doc pages, reusing `linkifyComponents` (not a bespoke mechanism).
- Names already inside backticks or existing `[...](...)` links are left alone (no double-linking).
- If monospace auto-links are in scope: `` [`<InternationalizationProvider>`](/components/InternationalizationProvider) `` renders as a monospace, clickable link.
- Plain `[label](href)` and bare `` `code` `` continue to work unchanged.
- External links keep their `target="_blank"` / `rel="noreferrer"` behavior; internal links stay in-app.
- No regression in existing i18n doc prose (which already uses both plain links and code spans).

## Pointers

- Auto-linker to reuse: `apps/docsite/src/components/changelogLinkify.ts` (`linkifyComponents`)
- Where it's wired for the changelog: `apps/docsite/src/app/(docs)/changelog/page.tsx` + `ChangelogView.tsx`
- Prose renderer for `.doc.mjs`: `apps/docsite/src/components/docs/inlineMarkdown.tsx` (`renderInlineMarkdown`) via `ProseBlock.tsx`

AKnassa:
Closes #4425

## What this does

Two things, one per commit, so either can be dropped on its own.

1. **Code spans inside link labels now render.** A doc link written as `` [`Name`](/x) `` renders as a monospace, clickable link instead of showing its backticks.
2. **Bare component and hook names in doc prose auto-link** to their doc pages, reusing the changelog's `linkifyComponents`.

## Why

The first one is a live bug, not a hypothetical. `internationalization.doc.mjs` already ships a hand-written monospace link, and the docs page renders its backticks as literal text today.

Rendered HTML for that paragraph, same page, same dev server:

```html
<!-- before -->
<a href="https://developer.mozilla.org/.../getTextInfo" rel="noreferrer" target="_blank">`Intl.Locale.getTextInfo()`</a>

<!-- after -->
<a href="https://developer.mozilla.org/.../getTextInfo" rel="noreferrer" target="_blank">
  <code data-color="inherit">Intl.Locale.getTextInfo()</code>
</a>
```

The second one means authors write `Popover` and get a monospace link, with no new syntax to learn and no hand-written hrefs to keep in sync.

## What changed

**Renderer.** The parse half of `renderInlineMarkdown` moved into a new `inlineTokens.ts` with no React and no StyleX. It recognises a label that is entirely one code span and passes an `isCodeLabel` flag; the renderer then puts `<Code color="inherit">` inside the anchor so the code keeps the link's accent color instead of falling back to body text. Scope is the common case only, as the issue asks: a mixed label like ``[the `x` prop](/y)`` still renders literally.

The extraction is what makes this testable. The docsite suite runs in plain node with no StyleX babel transform, so a module that calls `stylex.create` at import time throws on import there. A pure tokenizer can be unit tested; the React file cannot.

**Auto-linker.** `linkifyComponents` now runs over `.doc.mjs` prose, list items and table cells on the server, in the topic page, before anything reaches `renderInlineMarkdown`. Nothing extra ships to the client. Labels are emitted as code spans, which is what makes an auto-link render monospace. Column headers are skipped because they render as raw labels rather than inline markdown; headings and code blocks are untouched.

**A real bug in `linkifyComponents`.** It relied on adjacency lookarounds to skip mentions that were already deliberate, which only catches a name touching a backtick or bracket. Measured against the current docs, five mentions sat inside a code span without touching its delimiters and were rewritten into broken markup:

```
`<Theme>`  ->  `<[Theme](/components/Theme)>`
```

A name inside an existing link's href had the same problem. Code spans, fenced blocks and existing links are now opaque and copied through verbatim. That also gives authors the opt-out: backtick a mention and it stays plain. The changelog keeps plain labels, since monospace is opt-in, and picks up the same skip fixes.

**Result on the current docs:** 189 auto-links across 20 topics. Every generated href resolves; all 233 registry names have a `/components/<name>` page.

## Doc source edits

The linker is a heuristic over prose, so component names that are also ordinary English words matched wrongly in a handful of places. Those are fixed at the source, and each edit stands on its own:

* File paths and JSX snippets became code spans, which they should have been anyway: `icons`, `styling`, `typography`, `theme`.
* Four phrases where the word was a verb, a category or part of a name were reworded: "Center illustrations" to "Place illustrations", "Switch to it" to "Turn it on", "Token resolver APIs" to "Design token resolver APIs", and one sentence-initial "Layout" to "Page layout".

`packages/cli/docs` is prose only here, so no changeset, matching #4294 and #4243.

Two known residues, both judgment calls about wording rather than markup, left for whoever owns the voice:

* `theme.doc.mjs` "Poppins + Crimson Text" links `Text` to the Text component.
* `principles.doc.mjs` "Layout is frame-first" links `Layout`, which may or may not be intended.

## How to see it

`pnpm -F @astryxdesign/docsite dev`, then:

* `/docs/internationalization#direction` for the monospace link, in the "Text direction (RTL)" paragraph.
* `/docs/typography#type-scale` for the auto-linker: "consumed by `Text` and `Heading`", both monospace and clickable.

## Checks

* `pnpm -F @astryxdesign/docsite test` — 304 pass, 20 files, including 33 new and updated assertions across the tokenizer, the linkifier and the section mapper.
* `pnpm -F @astryxdesign/docsite typecheck` — clean.
* vitest `node` project, which covers `packages/cli` — 2083 pass.
* eslint on every changed file — clean.


packages/cli/assets/docs/typography.doc.mjs
```diff
@@ -122,7 +122,7 @@ export const docs = {
         },
         {
           type: 'prose',
-          text: 'Display text often needs heading semantics for accessibility. Use the type prop on Heading to apply display styling while preserving the correct HTML element: <Heading level={1} type="display-1"> gives you display-1 styling with an <h1> tag, so screen readers see the correct document outline.',
```
nynexman4464, line 125: was this changed because `<Heading level={1} type="display-1">` broke the parser?  If so we should try to fix that.

nynexman4464 (changes requested):
This looks promising, but I have a few pieces of feedback.

* For `Popover` on [this page](https://astryx-git-fork-aknassa-rocky-issue-4425-in-0c9f43-fbopensource.vercel.app/docs/browser-support) - I think for this case we ought to link to the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/Popover_API).  I don't have strong feelings if it renders as inline code or just a "regular" link.  
* For the [Internationalization](https://astryx-git-fork-aknassa-rocky-issue-4425-in-0c9f43-fbopensource.vercel.app/docs/internationalization#quick-start) page, my goal was a mention like `InternationalizationProvider` or `<InternationalizationProvider>` would link to the [doc page for the provider](https://astryx-git-fork-aknassa-rocky-issue-4425-in-0c9f43-fbopensource.vercel.app/components/InternationalizationProvider).  That way I don't have to explicitly link.  Part of me feels the impl should almost be the opposite: only component name mentions in backticks like `Popover` or inline code examples `<InternationalizationProvider>` should auto link to "known" component names.  Otherwise we don't auto link.   I guess the reverse is more common though.  For that specific case maybe we just make sure `\[\`<InternationalizationProvider\`\](/components/InternationalizationProvider)` actually works.
* eventually I'd almost like to eliminate `inlineMarkdown` in favor of our "standard" markdown component, but I guess we can leave that for now.


I do wonder if we should blocklist linking certain component names that could be ambiguous in prose.  Like `Popover` and `Layout` which I see were problems in the docs.  Alternatively I wonder if a `prose` doc block could add an `autoLinkComponents: false` prop or something like that.  WDYT?

AKnassa:
All three are in, plus a merge over #4766.

* browser-support now links every platform `Popover` mention to the MDN Popover API page, monospace like the existing `Intl.Locale.getTextInfo()` link. A corpus scan turned up one more of the same in the migration checklist, fixed the same way.
* Backticked mentions now auto link. Rather than extending the markdown rewriter, this rides the linkifier list from #4766: a code span that is exactly a documented name (`InternationalizationProvider`, `<InternationalizationProvider>`, `<Theme>`, `<Button />`) resolves to its component page and keeps the plain chip rendering. The i18n page needed no source edits, and component page prop descriptions pick it up too since they render through the same InlineCode. Bare names in prose keep linking as before. The explicit ``[`<InternationalizationProvider>`](/components/InternationalizationProvider)`` form also works; a tokenizer test covers that exact string.
* On a blocklist or `autoLinkComponents: false`: after this round a scan of all 20 topics plus every component doc shows 186 prose links and 23 code-span links with no wrong targets left. The wrong ones found along the way (platform Popover, "Layout is frame-first", Avatar's "Next.js `Link`") all read better fixed at the source. Inclined to hold off on a mechanism until a case shows up that a source edit can't fix; the linkifier list gives a clean place for a blocklist rule if one ever does.
* Agreed on eventually folding inlineMarkdown into the standard Markdown component. The tokenizer split here should make that mostly a swap.

nynexman4464 (approved):
lgtm

One future enhancement would be to auto detect external links and add `isExternalLink` so it's more clear, but I think that could be a separate PR

PR #4446, opened 2026-07-28
fix(theme-neutral): Badge error label meets WCAG AA (#4414)

### LINKED ISSUE #4414: Badge variant="error" label fails WCAG AA contrast (4.14:1) in the neutral theme
## Summary

`<Badge variant="error">` renders white label text on the neutral theme's badge red at **4.14:1**. WCAG 2.1 AA needs 4.5:1 for normal size text. The badge label is 12px at weight 500, so the 4.5 threshold applies rather than the 3:1 large text allowance.

This is the violation `pr-a11y` has been reporting as "Badge, 1 issue, `color-contrast`, affects 2/5 stories" on essentially every PR that touches `packages/core/src/Badge/`. It shows up with an identical 2/5 ratio on #4117, #4155, #4162, #4207, #4283, #4296, #4349 and #4407.

## Measured

Read from computed styles in Chrome against a locally served Storybook (`core-badge--variants`), neutral theme, light mode:

| variant | computed foreground on background | ratio | verdict (needs 4.5) |
|---|---|---|---|
| neutral | `#262626` on `#e5e5e5` | 12.01 | pass |
| info | `#ffffff` on `#0074e2` | 4.57 | pass |
| success | `#ffffff` on `#198100` | 5.02 | pass |
| warning | `#171717` on `#ffce2f` | 12.06 | pass |
| **error** | `#ffffff` on `#e33f4a` | **4.14** | **fail** |

`core-badge--status-labels` gives the same result, with `Failed` at 4.14:1 and everything else passing. Those are exactly the two of five Badge stories that contain an error badge, which accounts for the reported 2/5 element for element.

## Cause

This is not a cascade or specificity problem. The component scoped override is applied correctly. The problem is the value it applies.

`neutralTheme.ts` sets a global error colour that would pair fine with white:

```ts
'--color-on-error': ['#ffffff', '#171717'],
'--color-error':    ['#a50c25', '#ffc6c1'],
```

but the theme's Badge block re-binds `--color-error` to a much lighter red for the badge background, while the foreground stays `#ffffff`:

```ts
'variant:error': {
  // Red T55 saturated stop (= variant:error badge bg)
  '--color-error': '#e33f4a',
},
```

`#ffffff` on `#e33f4a` is 4.14:1. The sibling overrides in the same block are fine. `variant:warning` pairs `#171717` with `#ffce2f` at 12.06:1, and `variant:success` and `variant:accent` both clear the bar.

## Suggested fix

Darkening the badge scoped red is the smaller change and keeps the white foreground consistent with the other filled variants:

- `#c9303a` gives **5.29:1** against `#ffffff`, and stays close in hue to the current T55 stop.

Flipping the foreground instead does not work. `#171717` on `#e33f4a` is only 4.33:1, still short.

Note that `neutralTheme.ts` also carries `'variant:error': {backgroundColor: 'light-dark(#e33f4a, #ff705d)'}` for another component slot using the same red, so that pairing is worth checking before settling on a value.

## Relationship to #3654 and #3677

Not covered by #3677. That PR's only semantic change to the neutral theme is `--color-text-secondary` (`#737373` to `#616161`). The badge `variant:error` override is byte identical on its current head, and I re-measured to confirm the failure is still there.

**Worth a second pair of eyes:** #3677 adds `internal/theme-contrast/`, whose `effective.ts` explicitly models badge variant pairs including `variant:error` and applies the theme's re-bindings. Yet its suite reports zero failures across all sixteen theme and mode combinations, while axe in a browser flags this pair. Either the harness resolves the badge error background differently from the browser, or it has a blind spot here. I could not get `CONTRAST_REPORT=1` to print its audit table to confirm which, so I am raising it as a question rather than a claim.

## Not verified

Only the **neutral** theme was measured, since that is Storybook's default. A static scan of the other six themes was too unreliable to report, because the badge `variant:error` override is not present in all of them and the pairing differs per theme. The same check should be run per theme before this is called done.

## Repro

```
# open http://localhost:6006/iframe.html?id=core-badge--variants&viewMode=story
# inspect the "Error" badge: background-color rgb(227, 63, 74), color rgb(255, 255, 255)
```

Refs #3654

AKnassa:
## Summary

Fixes #4414.

`<Badge variant="error">` renders white on the neutral theme's light-mode red `#e33f4a` at **4.14:1**. The label is 12px at weight 500, so WCAG 2.1 AA asks for 4.5:1 — the 3:1 large-text allowance doesn't apply. This is the `color-contrast` violation the a11y audit reports as "Badge, affects 2/5 stories" on essentially every PR that touches `packages/core/src/Badge/`; those two stories are exactly the ones containing an error badge.

Flipping the label to dark doesn't rescue it either — `#171717` on `#e33f4a` is 4.33:1, still short. So the fill moves one tonal step down.

| | fill | label | ratio |
|---|---|---|---|
| before | `#e33f4a` | `#ffffff` | **4.14:1** |
| after | `#c9303a` | `#ffffff` | **5.29:1** |

`#c9303a` holds the hue — OKLCH H 21.9 → 22.8, C 0.200 → 0.189. Dark mode is untouched: dark text on `#ff705d` is 6.60:1.

## What else moved, and what didn't

**StatusDot and the ProgressBar `--color-error` rebinding move with it.** Both are commented in `neutralTheme.ts` as tracking the badge fill so "a dot and its badge read as one status language". Leaving them on the old red would have broken that stated invariant.

**`--shadow-inset-error` deliberately keeps `#e33f4a4D`.** It's a 30%-alpha focus ring, not a text surface, so the 4.5:1 threshold never applied to it and changing it would be a visual change with no accessibility gain.

The file header claimed *"All 9 saturated badge values pass WCAG AA (5.6–9.6 contrast range)"*, which wasn't true of error. It now points at the guard instead of restating a number.

## The guard

`scripts/check-badge-contrast.test.mjs` resolves every theme's badge label/fill pair through `light-dark()`, `var()` indirection (recursively, with a cycle guard) and alpha compositing over the theme's own surface token, then holds all of them to 4.5:1.

Resolution has to go through the component block: each theme re-binds the fill inside its own `components.badge`, so the pairing that actually ships **cannot** be read off the global `--color-*` tokens. Reading the tokens is what makes this bug easy to misdiagnose.

## Pre-existing shortfalls this surfaced

The issue asked for the check to be run per theme. Doing that found four more, in themes I have not touched:

| theme / scheme | variant | pair | ratio |
|---|---|---|---|
| butter, both | `info` | `#ffffff` on `#4883fd` | 3.54:1 |
| butter, both | `neutral` | `#225BFF` on `#ffee7b` | 4.43:1 |
| butter, both | `error` | `#ffffff` on `#fc473b` | 3.44:1 |
| gothic, both | `warning` | `#6c5010` on `#d3c490` | 4.32:1 |

Both themes pin those fills to brand values from a spec ("pinned to the brand colors from the spec"), so moving them is a design call rather than a bug fix — **left for maintainers**. They're recorded as an exact known-gap list, so a new shortfall fails the build and a fixed one fails until it's removed from the list. Happy to follow up with values if you want them changed.

`chocolate` and `matcha` define no badge block and inherit core's defaults, so they have no pair to resolve; the guard asserts that set explicitly rather than silently covering fewer themes than it looks like.

## Note on the regenerated template

`packages/cli/assets/templates/themes/neutral/neutralTheme.ts` is generated by `pnpm bundle:cli-themes`. After merging main (which moved the templates into `assets/` and regenerated them, picking up the previously-drifted `statusdot` block), the template diff mirrors the source change exactly.

## Tests

`scripts/check-badge-contrast.test.mjs` — 4 tests. Mutation-checked: restoring `#e33f4a` fails 3 of them.
Badge, StatusDot and `scripts` suites green. `pnpm check:repo` clean.


cixzhang (approved):
Thanks — 4.14:1 to 5.29:1 with the hue held, dark mode untouched, and a test that holds every theme to it. Landing.

[Reviewed by Robohands]

PR #4447, opened 2026-07-28
fix(Breadcrumbs): mark the current page with more than colour (#4421)

### LINKED ISSUE #4421: [Bug] WCAG 2.1 A 1.4.1 fail - use of colour failure, accessibility - Breadcrumb Item
### Description

Component(s) Affected: Breadcrumb
https://astryx.atmeta.com/components/BreadcrumbItem

### Goal

### What to do

### Why it's important

### Impact

### Screenshots

Documentation page for the UI component Breadcrumb Item. Highlighting poor use of colour impacting a11y. Displaying four usage variants, with start icon As current page (non-link) Supporting variant and With onClick handler

### Reproduction

Visit https://astryx.atmeta.com/components/BreadcrumbItem

### Astryx Version

@astryxdesign/core@v0.1.8

### Environment

_No response_

AKnassa:
## Summary

Fixes #4421.

The current crumb was distinguished from the navigable ones by **colour alone**, failing WCAG 2.1 A 1.4.1 (Use of Color). `itemStyles.current` pinned `fontWeight: 'inherit'`, so the only difference was `--color-text-primary` against the links' `--color-text-secondary`.

The `supporting` variant is worse than the report suggests: `supportingCurrent` and `supportingLink` **both** resolve to `--color-text-secondary`, so colour carried no information there at all, and nothing else took its place.

`aria-current="page"` was already set, so assistive tech was never the problem. The missing channel is the visual one, which is what 1.4.1 is about.

| | before | after |
|---|---|---|
| default variant | colour only (`primary` vs `secondary`) | colour + weight |
| supporting variant | **nothing** (same colour token) | weight |

## Change

The current crumb takes `--font-weight-semibold`, matching the current/selected idiom `SideNavItem` already uses. It applies to:

- an explicit `isCurrent` item,
- an auto-detected last item,
- a current crumb that is also a menu trigger (previously it kept plain link styling despite carrying `aria-current="page"`).

Navigable crumbs are untouched, so a trail doesn't turn into a row of identical bold text — there's a test pinning that.

## One adjacent gap, left for a maintainer call

When the last item has an `href`, the auto-detection effect sets `aria-current` on the anchor but the crumb keeps link styling — measured, it is byte-identical to the other links (same weight, same colour).

I did **not** close that here. It isn't a "colour alone" failure — there's no colour difference either — and closing it means making that effect drive a re-render, which the code comments describe as deliberately non-visual ("only sets an aria attribute, no visual change"). Reversing that intent seemed like your call rather than mine. Happy to take it in a follow-up.

## Tests

`BreadcrumbItem.useOfColour.test.tsx` — 6 tests reading the StyleX-injected rules back through `getComputedStyle`.

One thing worth flagging about the harness: jsdom reports an unset property as `''` and an unresolved `inherit` as `'inherit'`. Those are different **strings** but identical **pixels**, so a naive comparison of raw computed values reports a difference where a sighted user sees none. My first draft passed for exactly that reason before the fix existed. The helper now folds both to a single "no cue" value, which is what makes the pre-fix run actually red (5 of 6 failing).

Mutation-checked: restoring `fontWeight: 'inherit'` fails 5 tests.
All 42 Breadcrumbs tests green. `pnpm check:repo` and eslint clean.

### DIFF
--- a/.changeset/breadcrumb-current-non-colour-cue.md
+++ b/.changeset/breadcrumb-current-non-colour-cue.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] BreadcrumbItem: give the current page a heavier label weight so it is tellable from the navigable crumbs without relying on colour (WCAG 2.1 A 1.4.1). `current` previously pinned `fontWeight: 'inherit'`, leaving colour as the only signal — and in the `supporting` variant links and the current crumb resolve to the same `--color-text-secondary`, so there was no signal at all. The cue applies to an explicit `isCurrent` item, an auto-detected last item, and a current crumb that is also a menu trigger. Navigable crumbs are unchanged.
+
+@AKnassa
--- a/packages/core/src/Breadcrumbs/BreadcrumbItem.doc.mjs
+++ b/packages/core/src/Breadcrumbs/BreadcrumbItem.doc.mjs
@@ -28,7 +28,8 @@ export const docs = {
     {
       name: 'isCurrent',
       type: 'boolean',
-      description: 'Marks this item as the current page, applying aria-current="page".',
+      description:
+        'Marks this item as the current page, applying aria-current="page" and a heavier label weight so the current page stays distinguishable without relying on colour (WCAG 2.1 A 1.4.1).',
       default: 'false',
     },
     {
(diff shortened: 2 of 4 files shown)


cixzhang:
Thanks, although we'll likely want this to go through design review first. We're planning an accessibility project around September, so unfortunately this PR will likely go stale before then.

AKnassa:
Makes sense, closing for now. Happy to re-cut this for the September accessibility project once design review settles the treatment.

Issue #4308, opened 2026-07-25
RFC: Make TreeList row indentation themeable (currently an inline style on the highlighted element)


freddymeta:
## Summary

`TreeList` applies per-row indentation as an **inline `style`** on the same element that carries the row highlight and StyleX styling. Because inline styles beat every cascade layer, indentation is unreachable from the sanctioned theming layer (`@layer astryx-theme`), forcing consumers into `!important`/unlayered CSS. This RFC proposes making indentation reachable through sanctioned channels and lays out options for maintainers to choose from.

## Where it lives

In `packages/core/src/TreeList/TreeListItem.tsx`, indent is computed as:

```tsx
const computedMarginLeft = hasChildren
  ? `calc(${nestedLevel} * ${spacingVars['--spacing-4']})`
  : `calc(${nestedLevel} * ${spacingVars['--spacing-4']} + ${spacingVars['--spacing-4']} + ${spacingVars['--spacing-2']})`;
```

and applied as an inline style on the row's content wrapper `<div>` — which is **also** the `astryx-tree-list-item` theme target and the StyleX-styled element:

```tsx
<div
  {...mergeProps(
    themeProps('tree-list-item', {
      density,
      selected: isSelected ? 'selected' : null,
      disabled: isDisabled ? 'disabled' : null,
    }),
    stylex.props(
      styles.contentWrapper,
      densityStyles[density],
      /* interactive / focusVisibleOutline / disabled / selected ... */
      isSelected && styles.selected,
    ),
  )}
  style={{marginLeft: computedMarginLeft}}   // <-- inline indent on the highlighted element
  onClick={handleClick}>
  {innerContent}
</div>
```

Note that `data-tree-level` (and `data-tree-id`) live on the outer `<li role="treeitem">`, not on this content `<div>`, so the level is not available as a selector hook on the highlighted element itself.

## Consequences

1. **Inline style beats every cascade layer.** `defineTheme` output lands in `@layer astryx-theme` (above StyleX's `@layer astryx-base`), but an inline `style` outranks all layered CSS. So a consumer cannot retune indentation via a theme — they're pushed into `!important` and unlayered CSS. This is the root cause.
2. **The highlight includes the indent.** Because indent is applied *on* the highlighted content wrapper, the hover/selected background is inset by the indent and cannot span the full row width or clear the guide rails — a common design need for full-width navigation rows.
[...]

PR #4448, opened 2026-07-28
fix(TreeList): make the row indent reachable from the theme layer (RFC #4308, Option A)

AKnassa:
## Summary

Implements **Option A** of RFC #4308, and only Option A.

`TreeListItem` applied the per-level indent as an inline `margin-inline-start` on the row's content wrapper — the same element that carries the `astryx-tree-list-item` theme target. An inline longhand outranks every cascade layer, so a rule in `@layer astryx-theme` could never win against it. The indent was outside the theming contract entirely, and consumers were pushed into `!important`/unlayered CSS.

```diff
- style={{marginInlineStart: computedMarginLeft}}
+ style={indentStyle}                      // {'--_tree-indent': …}

  contentWrapper: {
+   marginInlineStart: 'var(--_tree-indent, 0px)',
  }
```

A custom property carries a **value**, not a declaration, so it doesn't create the same specificity floor. `margin-inline-start` now lives in `@layer astryx-base` where the theme layer overrides it in normal cascade order.

Rendering is unchanged at every nesting depth, including the extra chevron-width offset leaves carry so their labels line up with sibling parents' labels.

## What this does not do

RFC consequence (1) is a cascade bug — a themeable surface that no sanctioned channel can reach — so fixing it doesn't need a decision. The other two do, and I've left them alone:

- **Option B** (move the indent to an outer element so the highlight can span the full row, clearing consequence 2) is a DOM-structure change with back-compat and visual-regression risk. It wants a snapshot/visual baseline and your call.
- **Option C** (`indentSize` prop, consequence 4) is new public API.

Consequence (3) — per-level insets not derivable in `calc()` because `data-tree-level` lives on the `<li>` rather than the themed element — is unchanged by this PR, and as the RFC notes it isn't fully resolvable in pure CSS while that stays true.

Naming: I used the private `--_tree-indent`, matching the file's existing `--_tree-focus-outline` convention, since the per-row value is an implementation detail — what becomes theme-reachable is the `margin-inline-start` declaration, not the variable. If you'd rather publish it as a documented `--astryx-*` lever (a second, finer control), that's a one-line rename and I'm happy to make it.

## Tests

New `TreeListItem.indentReachable.test.tsx` — 5 tests pinning the two facts the fix rests on: no inline `margin-inline-start`/`margin-left` on the themed row at any depth, and a stylesheet rule declaring `margin-inline-start` from the variable. They deliberately don't depend on jsdom resolving `@layer` precedence, which it doesn't do reliably — the assertions are about where the declaration lives, which is the part that determines the outcome in a real browser.

`TreeList.test.tsx`'s noGuides indentation test asserted the inline longhand directly, so it now reads the variable. Its intent is unchanged: indent survives when connectors are suppressed, and grows with depth.

Mutation-checked: restoring the inline longhand fails 5 tests.
All 69 TreeList tests green. `pnpm check:repo`, eslint and `tsc --noEmit` clean.

### DIFF
--- a/.changeset/treelist-indent-layer-reachable.md
+++ b/.changeset/treelist-indent-layer-reachable.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] TreeList: stop setting the row indent as an inline `margin-inline-start`. That element is also the `astryx-tree-list-item` theme target, and an inline longhand outranks every cascade layer, so the indent could not be reached from `@layer astryx-theme` and consumers were pushed into `!important`/unlayered CSS. The row now publishes only the computed distance as `--_tree-indent` and `margin-inline-start` is declared in the stylesheet, restoring normal layer order. Rendering is unchanged at every nesting depth. This is Option A of RFC #4308 only — the DOM restructure (B) and an `indentSize` prop (C) are left as maintainer calls.
+
+@AKnassa
--- a/packages/core/src/TreeList/TreeList.test.tsx
+++ b/packages/core/src/TreeList/TreeList.test.tsx
@@ -431,26 +431,28 @@ describe('TreeList', () => {
     });
 
     it("variant='noGuides' preserves per-level indentation on the rows", () => {
-      // Indentation lives on the row's margin-inline-start (not the guide
-      // element), so it must survive when the connectors are suppressed. A
-      // deeper row is indented more than a shallower one.
+      // Indentation lives on the row (not the guide element), so it must
+      // survive when the connectors are suppressed. A deeper row is indented
+      // more than a shallower one. The row publishes the distance as
+      // `--_tree-indent`; the margin declaration itself is in the stylesheet
+      // (#4308), so read the variable rather than an inline longhand.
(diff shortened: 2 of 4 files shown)


cixzhang:
Thanks so much for this, @AKnassa — your write-up of the root cause was spot on: the indent was pinned to an inline `margin-inline-start` on the same element as the `astryx-tree-list-item` theme target, so it outranked every cascade layer and was unreachable from `@layer astryx-theme`. That framing is exactly what made the fix obvious.

We landed a forward fix in #4530 that builds on the same Option A cascade change you identified, and also takes you up on the offer at the end of your description — publishing the per-level step as a documented, public lever (`--tree-list-indent`, default `var(--spacing-4)`) rather than keeping it as the private `--_tree-indent`. That lets a theme retune the indent metric (e.g. to `var(--spacing-5)`) via `defineTheme` on the `tree-list` target, and the guide lines read the same variable so they stay aligned.

Since #4530 covers this ground, I'm going to close this PR — but the diagnosis here is what got us there, so thank you. Hope to see more contributions from you. 🙏

Issue #4401, opened 2026-07-26
deps: @hono/node-server 1.x -> 2.x (path traversal on Windows)


cixzhang:
@hono/node-server 1.19.14 (transitive via `@modelcontextprotocol/sdk`) is vulnerable to path traversal in `serve-static` on Windows via encoded backslash (`%5C`).

Patched in @hono/node-server 2.0.5 (major version bump).

**Current chain:** `@modelcontextprotocol/sdk@1.29.0` -> `@hono/node-server@1.19.14`

**Required:** @hono/node-server >=2.0.5

**Risk:** Major version bump. The vulnerability only affects Windows deployments using `serve-static`, so practical risk for this repo is low (dev tooling, Linux CI). Fix likely requires `@modelcontextprotocol/sdk` to release a version that depends on @hono/node-server 2.x, or a targeted pnpm override. Low urgency given the Linux-only usage.

Dependabot alert: #106

PR #4449, opened 2026-07-28
fix(deps): override sharp to 0.35 to clear inherited libvips CVEs (#4400)

### LINKED ISSUE #4400: deps: sharp 0.34 -> 0.35 (major bump, inherited libvips CVEs)
sharp 0.34.5 (transitive via next) is affected by inherited libvips vulnerabilities:
- CVE-2026-33327, CVE-2026-33328, CVE-2026-35590, CVE-2026-35591

Patched in sharp 0.35.0, which is a semver-major bump for 0.x.

**Current chain:** `next@15.5.20` -> `sharp@0.34.5`; `next@16.3.0-preview.5` -> `sharp@0.34.5`

**Required:** sharp >=0.35.0

**Risk:** sharp 0.35.0 may change native binary requirements or API surface. The fix likely requires next itself to update its sharp dependency (sharp is an optional dep of next used for image optimization). A pnpm override to `sharp: >=0.35.0` may work but needs verification that next's image optimization still functions correctly.

Dependabot alert: #108

AKnassa:
## Summary

Fixes #4400.

`next` pulls `sharp` as an optional dependency for image optimization and resolves **0.34.5**, which inherits four libvips advisories: CVE-2026-33327, CVE-2026-33328, CVE-2026-35590, CVE-2026-35591.

They're patched in sharp **0.35.0** — a semver-major bump on the 0.x line, so next's own `^0.34` range can never reach it. That's why this is an override rather than a lockfile refresh. Both next entries in the tree (`15.5.20` and `16.3.0-preview.5`) resolve through it.

```yaml
# pnpm-workspace.yaml → overrides
sharp: ^0.35.3
```

The platform packages move with it — `@img/sharp-libvips-*` goes **1.2.4 → 1.3.2**, which is where the actual fix lives.

`sharp` is already in `allowBuilds`, so the native postinstall path is unchanged. 0.35.3 published 2026-07-01, well clear of the repo's 7-day `minimumReleaseAge`.

The override is temporary by construction, and the inline comment says so: drop it once next depends on 0.35 itself.

## Verification

The issue flagged that an override "needs verification that next's image optimization still functions correctly". Since the risk here is the native binary and the next integration rather than the resolution, I checked past the lockfile:

- sharp 0.35.3 loads and reports **libvips 8.18.3**
- it round-trips a real encode — `sharp({create}).png().toBuffer()` returns a valid PNG
- `pnpm build` (core, lab, charts, vega, 7 themes) — succeeds
- `pnpm -F @astryxdesign/docsite build` — **succeeds**; a full next build with prerendering across ~270 paths, which is where image optimization would surface
- `pnpm test` — **384 files, 7871 tests, all pass**

## No changeset

This touches no published package's source. Matches the two most recent dependency-override PRs, #3889 and #4398, which were both `pnpm-lock.yaml` + `pnpm-workspace.yaml` only.

## Note on the sibling issue

#4401 (`@hono/node-server` 1.x → 2.x) is **not** covered here. The overrides block already pins `hono: ^4.12.25`, but that's the framework package, not `@hono/node-server`, so #4401 is genuinely still open and needs its own change.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -28,6 +28,7 @@ overrides:
   '@changesets/parse>js-yaml': '>=4.2.0'
   fast-uri: ^3.1.4
   brace-expansion: ^2.1.2
+  sharp: ^0.35.3
 
 importers:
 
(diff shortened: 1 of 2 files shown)


cixzhang (approved):
Thanks you!

PR #4453, opened 2026-07-28
feat(date-input): add clear + calendar-toggle icon theme targets (+ Icon styling-prop handling)

freddymeta:
## Summary

`DateInput`'s clear and calendar-toggle **icons** had no stable theme target — the component's only target was `astryx-date-input` on the root, and each glyph is a full `<Icon size="sm" color="secondary">` that sets its own color/size on the icon element. So consumers couldn't recolor, hover-morph, or resize those icons via `defineTheme` without a fragile descendant selector or raw CSS.

This PR adds icon-level theme targets for **both** DateInput icons — `astryx-date-input-clear-icon` and `astryx-date-input-toggle-icon` — consistent with the sibling `astryx-*` targets, and makes `Icon` a first-class styling-prop citizen so the targets actually reach the glyph. (Consolidates the previously-separate toggle-icon PR into this one.)

## Change

- **`Icon.tsx`** — `Icon` now fully handles its styling props (`className`, `style`, and a new `xstyle`), mirroring the pattern used by `Button` and other components: `xstyle` folds into `stylex.props(...)`, `className`/`style` merge via `mergeProps`, and the remaining props forward normally — in **both** the string-registry span path and the component-mode SVG path. Previously a consumer `className` was dropped on the registry path, so a component couldn't put a stable theme target on an `<Icon>`. Adds `xstyle?: StyleXStyles` to `IconProps` (documented). Provably byte-identical by default (no existing call site passes these props to a string-mode icon).
- **`DateInput.tsx`** — spread `{...themeProps('date-input-clear-icon')}` on the clear `<Icon>` and `{...themeProps('date-input-toggle-icon', { state })}` on the calendar-toggle `<Icon>` (full spread, so future variant/state data-attributes flow through). The toggle reflects open/closed as `data-state`. Aria-labels, handlers, icon names, and `color="secondary"`/`size="sm"` unchanged → defaults byte-identical.
- **`DateInput.doc.mjs`** — document both targets in `theming.targets` (EN + ZH). The `themingTargets` guard auto-discovers them.
- **`Icon.doc.mjs`** — document the new `xstyle` prop.
- **Tests** — assert both targets land on their icon elements, the toggle `data-state` reflection, a `defineTheme` override reaches each icon's color + size (+ hover for clear), that `Icon` forwards `className`/`style`/`xstyle`, and that default rendering is byte-identical.
- **`DateInput.stories.tsx`** — stories visibly recoloring + resizing the icons via `defineTheme`.
- Changeset (patch).

Because the targets sit on the icon elements, `defineTheme({ components: { 'date-input-clear-icon': { base: { … } } } })` lands in `@layer astryx-theme` (above the icon's own `astryx-base` color/size), so overrides win without `!important`. Each icon fills its (padding:0) button, so an icon `:hover` matches a button hover.

Example:
```ts
  'date-input-clear-icon': { base: {
    width: '12px', height: '12px', fontSize: '12px',
    color: 'var(--color-icon-secondary)',
    ':hover': { color: 'var(--color-icon-primary)' },
  } },
  'date-input-toggle-icon': { base: { width: '14px', height: '14px', fontSize: '14px' } },
}});
```

## Compatibility

Purely additive. Default rendering is byte-identical — the targets only add stable classes (plus `data-state` on the toggle icon), and `Icon` gains an optional `xstyle` prop plus now-composed `className`/`style`. No existing call site is affected.

### DIFF
--- a/packages/core/src/DateInput/DateInput.tsx
+++ b/packages/core/src/DateInput/DateInput.tsx
@@ -720,7 +720,14 @@ export function DateInput({
           type="button"
           onClick={handleClear}
           aria-label={t('@astryx.dateInput.clear', {label})}
-          {...stylex.props(styles.iconButton)}>
+          {...mergeProps(
+            // Stable theme target for the clear control. Distinct from the root
+            // `astryx-date-input` target, so a theme can restyle just the clear
+            // button/icon (color, size, hover) via `defineTheme` instead of a
+            // fragile structural selector.
+            themeProps('date-input-clear'),
+            stylex.props(styles.iconButton),
+          )}>
           <Icon icon="close" size="sm" color="secondary" />
         </button>
       )}
--- a/packages/core/src/DateInput/DateInput.tsx
+++ b/packages/core/src/DateInput/DateInput.tsx
@@ -720,15 +720,17 @@ export function DateInput({
           type="button"
           onClick={handleClear}
           aria-label={t('@astryx.dateInput.clear', {label})}
-          {...mergeProps(
-            // Stable theme target for the clear control. Distinct from the root
-            // `astryx-date-input` target, so a theme can restyle just the clear
-            // button/icon (color, size, hover) via `defineTheme` instead of a
-            // fragile structural selector.
-            themeProps('date-input-clear'),
-            stylex.props(styles.iconButton),
-          )}>
-          <Icon icon="close" size="sm" color="secondary" />
+          {...stylex.props(styles.iconButton)}>
+          <Icon
+            icon="close"
+            size="sm"
+            color="secondary"
+            // Stable theme target on the clear glyph itself, so a theme can
+            // restyle just this icon (color, size, hover) via `defineTheme`.
+            // Same-element rules in @layer astryx-theme win over the icon's own
+            // base color/size, which a button-level target could not reach.
+            className={themeProps('date-input-clear-icon').className}
+          />
         </button>
       )}
       {isBusy && <Spinner size="sm" />}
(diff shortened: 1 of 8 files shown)


packages/core/src/Icon/Icon.tsx
cixzhang, line 299: For consistency with other components, we'll want to extract all the styling props here and merge them into mergeProps below.

packages/core/src/DateInput/DateInput.tsx
```diff
@@ -721,7 +721,16 @@ export function DateInput({
...
-          <Icon icon="close" size="sm" color="secondary" />
+          <Icon
+            icon="close"
+            size="sm"
+            color="secondary"
+            // Stable theme target on the clear glyph itself, so a theme can
+            // restyle just this icon (color, size, hover) via `defineTheme`.
+            // Same-element rules in @layer astryx-theme win over the icon's own
+            // base color/size, which a button-level target could not reach.
+            className={themeProps('date-input-clear-icon').className}
```
cixzhang: We will want to spread themeProps here in case we want to support variants later.

cixzhang (changes requested):

freddymeta:
Addressed — `Icon` now fully handles its styling props (`className`, `style`, and a new `xstyle`), folding `xstyle` into `stylex.props` and merging `className`/`style` via `mergeProps` in both render modes, so the rest props forward normally. The icon targets now spread the full `{...themeProps(...)}` so future variants/states flow through. I also consolidated the calendar-toggle icon target (previously #4454) into this PR, so it now covers both DateInput icons.

cixzhang (approved):

PR #4454, opened 2026-07-28
feat(date-input): add astryx-date-input-toggle-icon theme target (+ forward Icon className)

freddymeta:
## Summary

The `DateInput` calendar-toggle button's **icon** has no stable theme target today. `DateInput`'s only theme target is `astryx-date-input` on the root wrapper; the toggle glyph is a full `<Icon icon="calendar" size="sm" color="secondary">` that sets its **own** color and size on the icon element. So a consumer who wants to recolor or resize the toggle icon (or style it per open/closed state) can only reach it through a fragile descendant selector or raw CSS — and `defineTheme` (same-element rules) can't target it.

This PR adds a dedicated `astryx-date-input-toggle-icon` theme target **on the calendar-toggle icon element**, consistent with the sibling `astryx-*` targets in the design system, and reflects the open/closed state so themes can style each.

## Change

- **`Icon.tsx`** — forward a consumer `className` on registry-rendered (string) icons (previously dropped on the string-icon span path; now composes with the internal classes; component-mode handled consistently). Enabling primitive; provably byte-identical by default (no repo call site passes a `className` to a string-mode icon).
- **`DateInput.tsx`** — spread `themeProps('date-input-toggle-icon', { state })` onto the toggle `<Icon>`, where `state` is `expanded`/`collapsed` (reflected as `data-state`). The `aria-label`, `onClick` toggle handler, `disabled`, and `icon="calendar"` are unchanged; `color="secondary"`/`size="sm"` kept so the default is byte-identical.
- **`DateInput.doc.mjs`** — document `astryx-date-input-toggle-icon` (with the `state` reflection) in `theming.targets` (both `docs` and `docsZh`). The `themingTargets` guard auto-discovers it.
- **`DateInput.test.tsx`** / **`Icon.test.tsx`** — assert the target lands on the icon element, the `data-state` reflection is correct, a `defineTheme` override reaches the icon's color + size, the default rendering is byte-identical, and `Icon` now forwards a consumer `className`.
- **`DateInput.stories.tsx`** — a story that visibly recolors + resizes the toggle icon via `defineTheme`.
- Changeset (patch).

Because the target sits on the icon element, `defineTheme({ components: { 'date-input-toggle-icon': { base: { … } } } })` lands in `@layer astryx-theme` (above the icon's own `astryx-base` color/size), so the override wins without `!important`.

Example:
```ts
  'date-input-toggle-icon': { base: { width: '14px', height: '14px', fontSize: '14px' } },
}});
```

## Compatibility

Purely additive. Default rendering is byte-identical — the target only adds a stable class plus a `data-state` attribute. No API changes. The `Icon` change only *adds* a previously-dropped className; no existing call site passes one.

### DIFF
--- a/.changeset/date-input-toggle-theme-target.md
+++ b/.changeset/date-input-toggle-theme-target.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] DateInput: add a dedicated `astryx-date-input-toggle` theme target for the calendar-toggle button, so consumers can theme it (color, size, hover, per open/closed state) via `defineTheme` instead of a fragile descendant selector. Reflects the popover's open/closed state as a `data-state` attribute (`expanded`/`collapsed`); default rendering is unchanged.
+
+@freddymeta
--- a/.changeset/date-input-toggle-theme-target.md
+++ b/.changeset/date-input-toggle-theme-target.md
@@ -2,6 +2,6 @@
 '@astryxdesign/core': patch
 ---
 
-[feat] DateInput: add a dedicated `astryx-date-input-toggle` theme target for the calendar-toggle button, so consumers can theme it (color, size, hover, per open/closed state) via `defineTheme` instead of a fragile descendant selector. Reflects the popover's open/closed state as a `data-state` attribute (`expanded`/`collapsed`); default rendering is unchanged.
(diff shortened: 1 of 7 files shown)


cixzhang:
I think this is a dupe of #4453

PR #4456, opened 2026-07-28
feat(date-range-input): add clear + calendar-toggle icon theme targets (+ Icon styling-prop handling)

freddymeta:
## Summary

`DateRangeInput`'s clear and calendar-toggle **icons** had no stable theme target — the component's only target was `astryx-date-range-input` on the root, and each glyph is a full `<Icon size="sm" color="secondary">` that sets its own color/size on the icon element. So consumers couldn't recolor, hover-morph, or resize those icons via `defineTheme` without a fragile descendant selector or raw CSS.

This PR adds icon-level theme targets for both `DateRangeInput` icons — `astryx-date-range-input-clear-icon` and `astryx-date-range-input-toggle-icon` — consistent with the sibling `astryx-*` targets, and makes `Icon` fully handle its styling props so the targets actually reach the glyph. (Same shape as the companion DateInput icon-target PR.)

## Change

- **`Icon.tsx`** — `Icon` now fully handles its styling props (`className`, `style`, and a new `xstyle`), mirroring `Button`: `xstyle` folds into `stylex.props(...)`, `className`/`style` merge via `mergeProps`, remaining props forward normally — in **both** the string-registry span path and the component-mode SVG path. Previously a consumer `className` was dropped on the registry path. Adds `xstyle?: StyleXStyles` to `IconProps` (documented). Byte-identical by default.
- **`DateRangeInput.tsx`** — spread `{...themeProps('date-range-input-clear-icon')}` on the clear `<Icon>` and `{...themeProps('date-range-input-toggle-icon', { state })}` on the calendar-toggle `<Icon>` (full spread). The toggle reflects open/closed as `data-state`. Aria-labels, handlers, icon names, and `color="secondary"`/`size="sm"` unchanged → defaults byte-identical.
- **`DateRangeInput.doc.mjs`** — document both targets in `theming.targets` (EN; DateRangeInput has no `docsZh`).
- **`Icon.doc.mjs`** — document the new `xstyle` prop.
- **Tests / stories / changeset** — assert both targets land on their icon elements, `data-state` reflection, `defineTheme` reaches icon color + size (+ hover), `Icon` forwards `className`/`style`/`xstyle`, byte-identical defaults; a story recolors + resizes the icons.

Example:
```ts
  'date-range-input-clear-icon': { base: {
    width: '12px', height: '12px', fontSize: '12px',
    color: 'var(--color-icon-secondary)',
    ':hover': { color: 'var(--color-icon-primary)' },
  } },
  'date-range-input-toggle-icon': { base: { width: '14px', height: '14px', fontSize: '14px' } },
}});
```

## Compatibility

Purely additive. Default rendering is byte-identical. The `Icon` change (shared with the DateInput icon-target PR) gains an optional `xstyle` prop and composes `className`/`style` instead of dropping them; no existing call site is affected.

### DIFF
--- a/.changeset/date-range-input-icon-theme-targets.md
+++ b/.changeset/date-range-input-icon-theme-targets.md
@@ -2,6 +2,6 @@
 '@astryxdesign/core': patch
 ---
 
-[feat] DateRangeInput: add `astryx-date-range-input-clear-icon` and `astryx-date-range-input-toggle-icon` theme targets on the clear and calendar-toggle glyphs, so consumers can recolor, hover-morph, and resize them via `defineTheme` instead of a fragile descendant selector or raw CSS. The toggle icon reflects its open/closed state as a `data-state` attribute. Also fixes `Icon` to forward a consumer `className` on registry-rendered icons (it was previously dropped). Default rendering is unchanged.
+[feat] DateRangeInput: add `astryx-date-range-input-clear-icon` and `astryx-date-range-input-toggle-icon` theme targets on the clear and calendar-toggle glyphs, so consumers can recolor, hover-morph, and resize them via `defineTheme` instead of a fragile descendant selector or raw CSS. The toggle icon reflects its open/closed state as a `data-state` attribute. `Icon` now fully handles its styling props (`className`, `style`, `xstyle`) so they compose with its base styles instead of being dropped. Default rendering is unchanged.
 
 @freddymeta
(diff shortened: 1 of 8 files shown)


cixzhang (changes requested):
Similar to 4453, we'll want the Icon to fully handle all the styling props (className, style, xstyle) then you should be able to forward the rest props from themeProps normally.

freddymeta:
Addressed — same as the DateInput PR: `Icon` now fully handles `className`/`style`/`xstyle` (Button pattern, both render modes) and the icon targets spread the full `{...themeProps(...)}`.

cixzhang (approved):

PR #4492, opened 2026-07-29
feat(inputs): add statusVariant='tooltip' and dedupe detached status icon

### LINKED ISSUE #4490: Inputs: surface status as a tooltip on the status icon
## Summary

When an input has a `status` (`error` / `warning` / `success`), it renders a colored status icon at the end of the control. That icon is currently **decorative only** — hovering or focusing it surfaces nothing. The status *message* is only shown in the attached/floating message box below the field (`statusVariant`), and when a consumer omits `status.message` (icon-only usage) there is no way at all to read what the status means.

We should let the status icon expose its meaning as a **tooltip on hover / keyboard focus**, so the state is discoverable directly from the icon.

## Motivation

- The status icon reads as a signal but gives no detail on its own.
- In compact layouts consumers often set `status` without `statusVariant` message space; the icon becomes the only affordance and today it says nothing.
- We already have the `useTooltip` primitive wired into these inputs (it powers the disabled-reason and label tooltips), so surfacing the status message on the icon is a natural, consistent extension.

## Proposed behavior

- When an input has a `status`, the status icon shows a tooltip on hover and keyboard focus containing the status message (falling back to a sensible per-type default label — e.g. "Error" / "Warning" / "Success" — when no message is provided).
- Wire the tooltip through the existing `useTooltip` pattern already used for `disabledMessage` / `labelTooltip`.
- Ensure the icon is focusable/hoverable enough to trigger the tooltip and that the tooltip text is associated for assistive tech (aria-describedby), without duplicating the message box announcement when both are present.

## Scope

Inputs that render the end-of-control status icon: `TextInput`, `NumberInput`, `DateInput`, `TimeInput`, and the related date/time inputs — anything driven by the shared `InputStatus` / status-icon pattern in `packages/core/src/Field`.

## Notes

- Avoid double-announcing when a `statusVariant` message box is also visible.
- Follow the existing tooltip conventions (`useTooltip`, `@media (hover: hover)` guards, describedBy wiring).

cixzhang:
Closes #4490

## What

Reworks the input status affordance around the `statusVariant` prop instead of an always-on tooltip. Adds a third variant, **`tooltip`**, and cleans up the existing `detached` variant.

- **`statusVariant="tooltip"`** *(new)* — hides the status message box and surfaces the status as an **info-tip on the on-field status icon**. The icon is a real focusable `<button>`, so the status is reachable by everyone:
  - **Keyboard (no AT):** the button is in the tab order (WCAG 2.1.1) with a visible focus ring (WCAG 2.4.7); focusing it reveals the message.
  - **Pointer:** hover reveals it (guarded to fine pointers).
  - **Touch:** hover is unavailable, so a tap toggles the tooltip open/closed.
  - **Assistive tech:** the button has an accessible name for its status type (WCAG 4.1.2) and is described by the message via `aria-describedby`; the message is also piped into the input's own `aria-describedby`.
  - **Dismissible + hoverable** (WCAG 1.4.13): Escape closes it and the pointer can bridge onto the surface.
- **`statusVariant="detached"`** *(fix)* — no longer renders a duplicate icon inside the control. The detached message box already carries a leading status icon, so the on-field glyph was redundant; it's now suppressed. The detached message's icon is also centered on the first line of text (previously it top-aligned to the flex row when the message wrapped).
- **`statusVariant="attached"`** — unchanged (default).

## How

A new shared `useInputStatusIcon` hook centralizes the on-field icon and its per-variant behavior + a11y wiring, so all bordered inputs stay consistent instead of each re-implementing the icon block. Wired into **TextInput, TextArea, NumberInput, DateInput, DateRangeInput, TimeInput, FileInput, and DateTimeInput**.

`DateTimeInput` is fixed to the detached presentation (it doesn't expose `statusVariant`), so it uses the helper only for the detached-icon dedup — it doesn't offer the `tooltip` variant. `Field`/`FieldStatus` skip rendering the message box for the `tooltip` variant.

## Why draft

The composed selector inputs (`Selector`, `MultiSelector`, `Typeahead`, `Tokenizer`, `PowerSearch`) render their status glyph in a combined chevron/status slot and don't use the new hook. They now *type-accept* `tooltip` (via the shared `FieldStatusVariant`) but don't yet implement it — extending the hook to that slot is the follow-up before this leaves draft.

## Testing

- `TextInput` tests cover the tooltip variant: no message box renders, the message shows in the tooltip, the input and the status button are both linked via `aria-describedby`, the affordance is a focusable button with a status-type accessible name, and it opens on keyboard focus.
- Full suite green across all touched inputs + `Field`/`FieldStatus` (552 tests), plus `typecheck`, `typecheck:docs`, `lint`, and `sync-exports --check`.
- Added a `tooltip` column to the `TextInput` status-variant comparison story.

### DIFF
--- a/packages/core/src/TimeInput/TimeInput.tsx
+++ b/packages/core/src/TimeInput/TimeInput.tsx
@@ -302,7 +302,7 @@ export interface TimeInputProps extends Omit<
    * How the status message is placed relative to the input.
    * - 'attached': message overlaps directly below the input (bordered treatment)
    * - 'detached': message floats below as a separate element with spacing
-   * - 'tooltip': no message box; the status icon shows the message in a tooltip on hover
+   * - 'tooltip': no message box; the status icon becomes a focusable info-tip button that reveals the message on hover, keyboard focus, or tap
    * @default 'attached'
    */
   statusVariant?: FieldStatusVariant;
--- a/packages/core/src/TimeInput/TimeInput.tsx
+++ b/packages/core/src/TimeInput/TimeInput.tsx
@@ -29,7 +29,6 @@ import {
   type FocusEvent,
 } from 'react';
 import * as stylex from '@stylexjs/stylex';
-import type {IconName} from '../Icon';
 import {
   colorVars,
   sizeVars,
@@ -41,7 +40,6 @@ import {
 import {
   Field,
   type InputStatus,
-  type InputStatusType,
   inputWrapperStyles,
   inputStatusBorderStyles,
   inputStatusHoverShadowStyles,
@@ -68,6 +66,7 @@ import type {SizeValue} from '../utils/types';
 import {useSize} from '../SizeContext/SizeContext';
 import {useAnnounce} from '../hooks/useAnnounce';
 import {useInputContainer} from '../hooks/useInputContainer';
+import {useInputStatusIcon} from '../hooks/useInputStatusIcon';
 import {useInputGroup} from '../InputGroup/InputGroupContext';
 import {groupStyles} from '../InputGroup/groupStyles';
 import {useTooltip} from '../Tooltip';
@@ -303,6 +302,7 @@ export interface TimeInputProps extends Omit<
    * How the status message is placed relative to the input.
    * - 'attached': message overlaps directly below the input (bordered treatment)
    * - 'detached': message floats below as a separate element with spacing
+   * - 'tooltip': no message box; the status icon shows the message in a tooltip on hover
    * @default 'attached'
    */
   statusVariant?: FieldStatusVariant;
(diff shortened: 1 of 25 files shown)


packages/core/src/TimeInput/TimeInput.tsx
```diff
@@ -409,26 +409,21 @@ export function TimeInput({
...
 
   const {ariaLabelledBy, ariaDescribedBy} = getInputARIA(
     inputLabelID,
     [
       description ? descriptionID : null,
-      status?.message ? statusMessageID : null,
+      statusVariant !== 'tooltip' && status?.message ? statusMessageID : null,
+      // The tooltip variant renders no message box; describe the input by the
+      // tooltip's content instead so the status is still announced.
+      statusTooltipDescribedBy,
```
cixzhang, line 426: make sure to only include these if set.

cixzhang:
Screenshot 2026-07-29 at 2 17 39 PM

PR #4498, opened 2026-07-29
feat(core): DropdownMenu submenus (DropdownMenuSubMenu)

cixzhang:
## What

Adds nested flyout menus to `DropdownMenu` (and `ContextMenu`) via a **single** `DropdownMenuSubMenu` component — the submenu half of the spec in #3829. The row adopts `DropdownMenuItem` semantics (label / icon / description / isDisabled); its `children` — or a data-mode `items` array — become the flyout content.

**Compound**
```tsx
<DropdownMenu button={{label: 'Actions'}}>
  <DropdownMenuItem label="Rename" onClick={rename} />
  <DropdownMenuSubMenu label="Move to" icon="folder">
    <DropdownMenuItem label="Folder A" onClick={() => move('a')} />
    <DropdownMenuItem label="Folder B" onClick={() => move('b')} />
  </DropdownMenuSubMenu>
</DropdownMenu>
```

**Data mode** — a row that declares nested `items` becomes a submenu, no new item "type":
```tsx
<DropdownMenu items={[
  {label: 'Rename', onClick: rename},
  {label: 'Move to', icon: 'folder', items: [
    {label: 'Folder A', onClick: () => move('a')},
    {label: 'Folder B', onClick: () => move('b')},
  ]},
]} />
```

## Why one component, not three

The first pass mirrored Radix's `Sub` / `SubTrigger` / `SubContent` split. That doesn't match Astryx conventions: our own nesting precedents — **`SideNavItem`** and **`TreeListItem`** — are a *single* component where nested content (children) promotes a normal row into a nested surface. `DropdownMenuSubMenu` follows that pattern. It also matches the legacy internal `XDSDropdownSubMenuItem`, which was one component. The `*Item` suffix in this family means a leaf row (`DropdownMenuItem`, `DropdownMenuRadioItem`, `DropdownMenuCheckboxItem`); a submenu *owns a menu*, so `SubMenu` is the honest name.

Submenus don't exist in open-source Astryx today, but legacy internal XDS shipped this and consumers relied on it; nav/overflow menus currently fake nesting with flattened sections or a second popover. This ports the value (APG menubar-navigation contract, hover + keyboard, async spinner) onto Astryx primitives.

## How

Built entirely on existing primitives — no bespoke floating/positioning code:
- **Placement:** `useLayer` context mode opens the flyout `inline-end` with viewport auto-flip via CSS anchor positioning (same mechanism as ContextMenu, RTL-correct by default).
- **Pointer:** `useMenuHover` for open/close intent delay.
- **Keyboard:** a per-level `useListFocus` + `useTypeahead`. Right / Enter / Space opens the flyout and focuses its first item; Left / Escape closes it and returns focus to the trigger. Right/Left swap under RTL.
- **A11y:** trigger row is `menuitem` + `aria-haspopup="menu"` + `aria-expanded` + `aria-controls`; flyout is a `role="menu"` labelled by its trigger.

## Risk

Additive only — one new export plus one optional `items?` field on `DropdownMenuItemData`. No behavior change to existing menus. Disabled submenus never open. Lab-first surface; promote after vibe test.

## Testing

- 7 unit tests (compound + data mode): a11y wiring, click-to-open, ArrowRight-opens / ArrowLeft-returns-focus, `aria-labelledby`, nested-item selection, disabled-never-opens.
- Storybook stories (basic, nested, async spinner, data-driven).
- `tsc` clean, eslint clean.

## Follow-ups (from the #3829 spec, not in this PR)

- Vibe-test hover intent + safe-triangle tuning and touch/mobile open affordance.
- Confirm `items`-implies-submenu clarity vs. an explicit marker in arbitration.

Closes the submenu track of #3829.

### DIFF
--- a/packages/core/src/DropdownMenu/DropdownMenuSubMenu.tsx
+++ b/packages/core/src/DropdownMenu/DropdownMenuSubMenu.tsx
@@ -1,0 +1,553 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+'use client';
+
+/**
+ * @file DropdownMenuSubMenu.tsx
+ * @input React, stylex, useLayer (context mode), useListFocus, useMenuHover,
+ *   useTypeahead, Item, Icon, Spinner, DropdownMenu context + item roles.
+ * @output Exports DropdownMenuSubMenu — a single menu row that reveals a nested
+ *   flyout menu of its own children/items.
+ * @position Sub-component; place inside a DropdownMenu (or ContextMenu)
+ *   alongside plain items.
+ *
+ * One component, not three. The row itself adopts DropdownMenuItem semantics
+ * (label / icon / description / isDisabled) and its children become the
+ * flyout's content. This mirrors how SideNavItem / TreeListItem promote a
+ * normal row into a nested surface when given children, rather than the Radix
+ * Sub / SubTrigger / SubContent split. Data-driven menus never touch this
+ * component directly — renderDropdownItems renders it from a nested `items`
+ * array and passes the rendered children in.
+ *
+ * Built on existing primitives — no bespoke floating code:
+ * - Positioning: useLayer context mode opens the flyout inline-end with
+ *   viewport auto-flip via CSS anchor positioning (RTL-correct by default).
+ * - Pointer: useMenuHover for open/close intent.
+ * - Keyboard: a per-level useListFocus + useTypeahead. Right (Left in RTL) /
+ *   Enter / Space opens the flyout and focuses its first item; Left (Right in
+ *   RTL) / Escape closes it and returns focus to the trigger row.
+ *
+ * Prior art: legacy internal XDS `XDSDropdownSubMenuItem` (APG menubar-
+ * navigation submenu). This re-expresses the same contract on Astryx primitives.
+ *
+ * SYNC: When modified, update these files to stay in sync:
+ * - /packages/core/src/DropdownMenu/DropdownMenuSubMenu.doc.mjs
+ * - /packages/core/src/DropdownMenu/DropdownMenuSubMenu.test.tsx
+ * - /packages/core/src/DropdownMenu/index.ts
+ * - /apps/storybook/stories/DropdownMenu.stories.tsx
+ * - /packages/cli/templates/blocks/components/DropdownMenu/ (showcase blocks)
+ */
+
+import React, {
+  useCallback,
+  useId,
+  useMemo,
+  useRef,
+  useState,
+  type ReactNode,
+} from 'react';
+import * as stylex from '@stylexjs/stylex';
+import {Icon, renderIconSlot, type IconType} from '../Icon';
+import {Spinner} from '../Spinner';
+import {Item} from '../Item';
+import {useLayer} from '../Layer/useLayer';
+import {layerAnimations} from '../Layer/layerAnimations.stylex';
+import {useListFocus} from '../hooks/useListFocus';
+import {useMenuHover} from '../hooks/useMenuHover';
+import {useTypeahead} from '../hooks/useTypeahead';
+import {
+  colorVars,
+  spacingVars,
+  radiusVars,
+  durationVars,
+  easeVars,
+  shadowVars,
+  typographyVars,
+  typeScaleVars,
+} from '../theme/tokens.stylex';
+import {mergeProps} from '../utils';
+import {themeProps} from '../utils/themeProps';
+import type {BaseProps} from '../BaseProps';
+import {
+  MENU_ITEM_ROLES,
+  MENU_ITEM_SELECTOR,
+  MENU_BOUNDARY_SELECTOR,
+} from './menuItemRoles';
+import {
+  DropdownMenuContext,
+  useDropdownMenuContext,
+  type DropdownMenuContextValue,
+} from './DropdownMenuContext';
+
+const triggerStyles = stylex.create({
+  root: {
+    boxSizing: 'border-box',
+    width: '100%',
+    paddingBlock: spacingVars['--spacing-2'],
+    paddingInline: spacingVars['--spacing-2'],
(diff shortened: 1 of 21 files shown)


packages/core/src/DropdownMenu/DropdownMenuSubMenu.tsx
```diff
@@ -0,0 +1,552 @@
...
+  useDropdownMenuContext,
+  type DropdownMenuContextValue,
+} from './DropdownMenuContext';
+import {renderDropdownItems} from './renderDropdownItems';
+import type {DropdownMenuOption} from './DropdownMenu';
+
+// Navigation keys owned by a menu level's useListFocus. When pressed inside a
+// flyout they must not also bubble to the parent menu's key handler (the
+// flyout renders inline, inside the parent's roving-focus container).
+const NAV_KEYS: ReadonlySet<string> = new Set([
```
cixzhang: Sicne this needs to leverage the keys from useListFocus can we retrieve it from there?

packages/core/src/DropdownMenu/DropdownMenuSubMenu.tsx
```diff
@@ -0,0 +1,552 @@
...
+  // closes the flyout and returns focus to the trigger.
+  //
+  // The flyout renders inline in the React tree (useLayer context mode is a
+  // native popover, not a portal), so its DOM sits *inside* the parent menu's
+  // roving-focus container. Any key this level handles must therefore
+  // stopPropagation, or it also bubbles to the parent menu's onKeyDown — whose
+  // getItems() descends into the flyout and moves focus a second time, skipping
+  // an item. The innermost open menu owns the keyboard; only keys we don't
+  // handle (e.g. Tab) are allowed to bubble so the parent can close the stack.
+  const handleContentKeyDown = useCallback(
```
cixzhang, line 406: Anyway we can leverage useListfocus without duplicating some of the behavior here just for stopPropagation?

packages/core/src/DropdownMenu/renderDropdownItems.tsx
```diff
@@ -10,6 +10,7 @@ import type {ReactNode} from 'react';
 import * as stylex from '@stylexjs/stylex';
 import {Divider} from '../Divider';
 import {DropdownMenuItem} from './DropdownMenuItem';
+import {DropdownMenuSubMenu} from './DropdownMenuSubMenu';
```
cixzhang, line 13: hmm is this a circular dependency? Could it have issues?

packages/core/src/hooks/useListFocus.ts
```diff
@@ -38,6 +38,20 @@ export interface UseListFocusOptions {
...
+   * list container element. Return false to exclude it.
+   *
+   * Primary use: nested overlays. A submenu flyout renders inline (native
+   * popover, not a portal), so its items are DOM descendants of the parent
+   * menu and would otherwise be swept up by `querySelectorAll` — leaving
+   * hidden items in the parent's roving order that `.focus()` can't land on.
+   * Scoping to the nearest `[role="menu"]` keeps each level's navigation to its
+   * own items.
+   */
+  itemFilter?: (el: HTMLElement, listEl: HTMLElement) => boolean;
```
cixzhang: Would it be better to provide the container scope?

packages/core/src/hooks/useListFocus.ts
cixzhang, line 329: In the case of the sub menu, I would have expected this to be the submenu container, but since it's targetting items outside (resulting in needing the filter) it sounds like this isn't the case? But why?

packages/core/src/DropdownMenu/DropdownMenuSubMenu.tsx
cixzhang: Good call — removed the hardcoded key set. The root cause was that the flyout renders inline (useLayer context mode is a native popover, not a portal), so its `role="menu"` is a DOM descendant of the parent menu and child key events bubble to the parent's handler. Instead of enumerating keys here, I taught `useListFocus` about it via a new `boundarySelector` option: it scopes item collection to the level *and* exposes `ownsEvent(e)` so each level ignores events bubbled up from a nested list. `NAV_KEYS` is gone.

packages/core/src/DropdownMenu/DropdownMenuSubMenu.tsx
cixzhang: Done — `useListFocus` now owns this. With `boundarySelector` set, the hook self-filters bubbled events (`ownsEvent`), so the content handler no longer needs any `stopPropagation`. The one remaining `stopPropagation` is on the *trigger* row only — it's a direct child of the parent menu (not a nested `role="menu"`), so the boundary can't distinguish it; documented inline.

packages/core/src/DropdownMenu/renderDropdownItems.tsx
cixzhang: Yes, real cycle: renderDropdownItems → DropdownMenuSubMenu → renderDropdownItems. Broken it — `renderDropdownItems` is now the sole recursion owner and passes the rendered items as `children` to `DropdownMenuSubMenu`, which no longer imports it. `DropdownMenuSubMenu` is compound-only now; data-mode parity lives one level up (same as `DropdownMenu` itself).

packages/core/src/hooks/useListFocus.ts
cixzhang: Yes — implemented exactly this as `boundarySelector`. When set to `'[role="menu"]'`, item collection and key-event ownership are both scoped to this level's own container.

packages/core/src/hooks/useListFocus.ts
cixzhang: Great question — it exposed the mental-model mismatch. The flyout *is* its own `role="menu"` container, but because `useLayer` context mode renders inline (not portalled), that container sits *inside* the parent menu's DOM. So the parent's `querySelectorAll` reached down into the child flyout. `boundarySelector` fixes it: each level only counts items whose nearest `role="menu"` ancestor is its own container — no cross-container reach.

Issue #4248, opened 2026-07-23
Pagination: editable current-page input variant, first/last double-chevron buttons, and configurable page label


cixzhang:
## Summary

Three enhancements to `Pagination` for parity with common internal pagination patterns. Two are additive; one is a small labeling fix.

## Current state

`Pagination` today has `variant: 'pages' | 'count' | 'compact' | 'none'` (extensible via `PaginationVariantMap` module augmentation) plus prev/next buttons. `compact` renders **"Page X of Y"** as static text. The visible "Page" word comes from the i18n key `@astryx.pagination.pageOfTotal`; the `label` prop is the **accessible nav-landmark label** (aria), not the visible word.

### 1. Editable current-page variant (compact + text input)
A variant like `compact` but where the current page is an **editable text input** — the user can type a page number and jump to it (Enter / blur commits), instead of only reading "Page X of Y". Effectively "Page [__] of Y" where `[__]` is a small number input bound to `page`/`onChange`.
- Should validate/clamp to `1..totalPages` and ignore invalid input (mirror the input-family pending-value handling, e.g. how NumberInput/DateInput handle pending + commit).
- Likely a new `variant` value (augment `PaginationVariantMap`, e.g. `input`) rather than a boolean, to stay consistent with the existing variant axis.

### 2. First / last page buttons (double-chevron) at the edges
An opt-in to render **first-page** and **last-page** buttons at the outer edges, using **double-chevron** icons (e.g. `chevronDoubleLeft` / `chevronDoubleRight` from the icon registry — verify exact names), jumping to page 1 and `totalPages`.
- Opt-in prop (e.g. `hasFirstLast?: boolean`, default `false` — non-breaking). Disabled state at the boundaries (first-page button disabled on page 1, last on the final page), matching the prev/next disabled behavior.
- Works across the relevant variants (at least `pages` and the new input variant); decide interaction with `none`.

### 3. Configurable "Page" label
Make the visible **"Page"** word (currently hardcoded via `@astryx.pagination.pageOfTotal`) configurable, so consumers can relabel it (e.g. "Step", "Slide", localized/custom wording) without overriding i18n globally.
- This is distinct from the existing `label` prop (that's the aria nav-landmark name). Add a separate prop for the visible page-word (or a small label-set), and keep the i18n default.
- Consider: does this also affect the `count`/`compact` text and the new input variant consistently?

## Considerations
[...]

PR #4505, opened 2026-07-29
feat(pagination): add editable input variant with page/row navigation mode

freddymeta:
## What

Adds an **`input` pagination variant** that renders the full editable-navigation primitive:

- **Page mode:** `«  ‹  Page [ n ] / N  ›  »`
- **Row mode:** `«  ‹  Row  [ n ]      ›  »`  (no `/ N` total)

where `«` `‹` `›` `»` are **first / prev / next / last**. Typing a value in the box and committing (Enter or blur) jumps to it; invalid or empty entries revert.

A navigation **mode** controls what the box represents:

- **`navigateBy: 'page' | 'row'`** (default `'page'`).
  - `'page'` — the box holds a 1-based **page number**, clamped to `1..totalPages`, rendered as `Page [ n ] / N`.
  - `'row'` — the box holds a 1-based **row index**, clamped to `1..totalItems`, rendered as `Row [ n ]` (no total). Committing navigates to the page containing that row (`ceil(row / pageSize)`) and reports the row via **`onRowNavigate(row)`** (fired alongside `onChange`).

The range readout you'd see to the right of a divider (`Showing 1–N of M`) is intentionally **not** part of this primitive — that's app-level content the consumer composes next to the pager, not inside it.

## Why

#4248 asks for (1) an editable current-page control and (2) first/last buttons. The existing variants (`pages`, `count`, `compact`, `dots`, `none`) display the position but don't let you type it or jump to the ends. This adds the complete control, and generalizes the box slightly so it can also drive row-level navigation (jump to a record in a long list).

## How

- **Variant map:** `input` added to `PaginationVariantMap` (extensible via module augmentation like the others), so `variant="input"`.
- **Label + total:** a visible leading label (`Page` / `Row`) and, in page mode, a trailing `/ N` total. Both are i18n messages (`@astryx.pagination.pageLabel`, `@astryx.pagination.rowLabel`, `@astryx.pagination.ofTotalPages`). The label is `aria-hidden` — the input already carries its own accessible name via `@astryx.pagination.goToPageInput`.
- **First / last:** `«`/`»` double-chevron buttons flank prev/next, gated behind a new **`hasFirstLast`** prop (default `true`, **`input` variant only**). They're RTL-aware exactly like prev/next (via `useDirection()`), disabled at the boundaries, and **omitted when the page count is unknown** (cursor/`hasMore` pagination, where there's no last page to jump to). aria-labels via `@astryx.pagination.first` / `.last`. Clicking first → page 1, last → `computedTotalPages`; both fire the existing `pageAnnounce` live region.
- **New icons:** the fallback icon set had no double-chevron glyph, so this registers two new public icons **`chevronsLeft` / `chevronsRight`** — added to `defaultIcons.tsx`, the `IconName` union in `globalIconRegistry.tsx`, all seven theme icon registries (`packages/themes/*/src/icons.tsx`, mapped to lucide's `ChevronsLeft`/`ChevronsRight`) and their bundled CLI templates, plus the icon usage hints and Icon docs.
- **Commit/clamp:** reuses the input-family pending/commit pattern (mirrors `NumberInput`/`DateInput`): a local pending string shadows the committed value while typing; Enter and blur commit (parse → clamp → navigate), Escape discards, invalid/empty reverts.
- **Theming:** stable classes `astryx-pagination-input`, `astryx-pagination-input-label`, `astryx-pagination-input-total` via `themeProps(..., {size})`, documented in `Pagination.doc.mjs` `theming.targets`. StyleX only, semantic tokens only.
- **Types:** new exported `PaginationNavigateBy`.

## Non-breaking

Defaults are unchanged (`variant='pages'`, `navigateBy='page'`). `hasFirstLast` only affects the `input` variant — `pages`/`count`/`compact`/`dots`/`none` render identically (covered by a non-regression test). New props are optional.

## Scope / relationship to #4248

This now implements **request #1** (editable current-page control) **and request #2** (first/last buttons). **Request #3** (a *configurable* page label) is still out of scope — note this variant renders a **fixed** "Page"/"Row" label (localizable, but not consumer-configurable). A follow-up could add a label override prop if wanted.

## Open questions for review (still a proposal)

1. **Variant name** — `input` reads clearly; `editable`/`jump` are alternatives.
2. **`navigateBy` naming + shape** — is `'page' | 'row'` the right axis, or should row navigation be a separate concern?
3. **`onRowNavigate`** — fires *alongside* `onChange(page)` (keeps `page`/`onChange` the single source of truth). Should row mode instead suppress `onChange` and only emit the row?
4. **`hasFirstLast` default** — defaulting `true` makes the primitive match the design out of the box; is opt-out the right polarity, or should first/last be unconditional for `input`?
5. **Icon names** — `chevronsLeft` / `chevronsRight` (lucide-style). Open to a different convention.

### Focus ring & default story (update)

- The editable box now uses the **single canonical bordered-input focus ring** — the accent border plus an inset accent-muted `box-shadow` on `:focus-visible` (`outline: none`), matching every other bordered input. It previously drew a second, outset outline on top of the border (a doubled ring).
- The default **Input (page mode)** story now demonstrates a richer `formatLabel` (`Page [ n ] / N · showing rows X–Y of Z`); the custom-label story shows a distinct compact example.

## Testing

- Unit tests cover: page mode renders `Page [ n ] / N`; row mode renders `Row [ n ]` with **no** `/ N`; Enter/blur commit + clamp; revert on invalid/empty; first → page 1, last → last page; disabled at boundaries; `hasFirstLast={false}` hides them; first/last omitted when total unknown; double-chevron icons render; other variants gain no first/last; a11y names; announce fires; row mode maps rows→pages and clamps to `totalItems`.
- Storybook: `Variant: Input (page mode)`, `Variant: Input (row mode)`, `Variant: Input (no first/last)`, and an `All Variants` entry.

Local gates green: full core suite (5486 passing), core + docs + storybook typecheck, ESLint (all changed files), `sync-exports --check`, `check-sync`, `check-changesets`, `token-docs --check`, package-boundaries, full build, docsite generate + test.


cixzhang:
image

Here's how it looks. Did we want this to match closer to internal version with something like "[input]/5 pages" where it's out of the max pages?

freddymeta:
Thanks for the steer — you're right, I had the intent backwards. Reworked to the full primitive:

- **Total is now shown** in page mode: the box renders `Page [ n ] / N` (localized via a new `@astryx.pagination.ofTotalPages` message). Row mode stays `Row [ n ]` with no `/ N`, since a row index has no fixed "of N" total in the same sense.
- **First/last `«` `»` are now included**, flanking prev/next, RTL-aware and disabled at the boundaries. They needed a double-chevron glyph the fallback set didn't have, so I registered `chevronsLeft` / `chevronsRight` and wired them through the theme registries too.
- The only thing intentionally left **out** of the primitive is the `Showing 1–N of M` range readout — that's app-level content that sits to the right of the divider, composed next to the pager rather than baked into it. Happy to revisit if you'd rather it live inside.

First/last are gated behind `hasFirstLast` (default `true`, `input` variant only) so the primitive matches the design out of the box without touching any other variant. A few API questions are open in the description — variant name, `navigateBy` shape, `onRowNavigate` firing alongside vs. instead of `onChange`, the `hasFirstLast` default, and the new icon names — would value your read on those.

cixzhang (changes requested):
This is a genuinely useful variant — an editable page/row box is something people ask for, and the row→page jump math is nicely handled. One direction I'd like to steer before we lock the public API: right now the labeling (`Page [ n ] / N`, `Row [ n ]`) is baked into fixed i18n strings, and `navigateBy` bundles the value's meaning, the layout, and the total-display into one enum. I'd rather give consumers full control over wording via a single `formatLabel` callback that receives the pagination state (`page`, `pageCount`, `rowStart`/`rowEnd`, `totalItems`) plus the pre-wired editable `input` node, and returns whatever label they want — `"Page [n] / N"`, `"3–4 of 200 rows"`, localized/RTL orderings, or a bare box. Your current default becomes the default `formatLabel`, so your own use case stays a one-liner (often zero config). I'd keep a small `valueUnit: 'page' | 'row'` just for what the number *means* (which drives `onRowNavigate`), and keep `hasFirstLast` and the new chevron icons as-is. Want me to sketch the `PaginationState` shape so we can align before you refactor? Really nice work on the keyboard/commit handling.

freddymeta:
@cixzhang — re-requesting review. Reworked per your steer and rebased clean onto `main` (0 behind, CI green):

- **Full input primitive** now matches the internal design: the box renders `Page [n] / N` (total shown), with first/last `«` `»` flanking prev/next (RTL-aware, disabled at boundaries). Row mode stays `Row [n]`.
- Replaced the `navigateBy` enum with a **`formatLabel` API** so the label composition is caller-controlled rather than enumerated.
- **Single canonical focus ring** on the input box (accent border + inset accent-muted box-shadow, `outline: none`) — fixed the earlier doubled ring.
- Input clamps to range at both ends and requires a known total.

Storybook preview shows the primitive out of the box. Open questions from the description (variant name, `hasFirstLast` default, new icon names) still welcome your read.

_— on behalf of Freddy_

packages/core/src/Pagination/Pagination.tsx
```diff
@@ -53,6 +59,44 @@ import type {PaginationVariantMap} from './index';
...
+/**
+ * What the editable input's number *means* in the `input` variant. This drives
+ * navigation semantics only — the visible wording is controlled by
+ * `formatLabel`.
+ * - page: the value is a 1-based page number (default).
+ * - row: the value is a 1-based row index; committing it navigates to the page
+ *   that contains that row (computed from pageSize) and reports the row via
+ *   onRowNavigate. Requires totalItems to resolve rows to pages.
+ */
+export type PaginationValueUnit = 'page' | 'row';
```
cixzhang: Can we have an open pageLabel instead?

packages/core/src/Pagination/Pagination.tsx
```diff
@@ -104,10 +148,45 @@ export interface PaginationProps extends Omit<
...
+  /**
+   * What the editable box's number means in the `input` variant.
+   * - page: the box holds a 1-based page number (default).
+   * - row: the box holds a 1-based row index; committing it navigates to the
+   *   page containing that row (computed from pageSize) and fires
+   *   onRowNavigate. Requires totalItems in `'row'` mode to resolve rows to
+   *   pages.
+   * @default 'page'
+   */
+  valueUnit?: PaginationValueUnit;
```
cixzhang: We have a format label so I think we can remove this.

packages/core/src/Pagination/Pagination.tsx
```diff
@@ -104,10 +148,45 @@ export interface PaginationProps extends Omit<
...
+   * ordering, or a bare box. The default reproduces "Page [ n ] / N" (page
+   * unit) and "Row [ n ]" (row unit), so most consumers never set this.
+   */
+  formatLabel?: (state: PaginationState, input: ReactNode) => ReactNode;
+  /**
+   * Called with the committed 1-based row index when the `input` variant uses
+   * `valueUnit='row'`. Fires alongside onChange (which moves to the page
+   * containing that row). No-op in page mode.
+   */
+  onRowNavigate?: (row: number) => void;
```
cixzhang: I think we can rely on page navigation in cases where we want the label to read "Row" instead.

cixzhang (changes requested):

packages/core/src/Pagination/Pagination.tsx
freddymeta: Done — replaced `valueUnit` with an open `pageLabel?: string` (defaults to the localized "Page"). A consumer wanting a "Row" reading just passes `pageLabel="Row"`.

_— on behalf of Freddy_

packages/core/src/Pagination/Pagination.tsx
freddymeta: Removed — `formatLabel` (and the `PaginationState` it took) is gone, along with `valueUnit`. The label is now just the `pageLabel` noun + the editable box + `/ N`.

_— on behalf of Freddy_

packages/core/src/Pagination/Pagination.tsx
freddymeta: Done — dropped `onRowNavigate` and all the row→page math. Navigation now always flows through the existing `onChange(page)`, same as every other variant; a "Row" label is just `pageLabel="Row"` on a page-navigated box. `onChange` is unchanged and no new callback was added.

_— on behalf of Freddy_

freddymeta:
@cixzhang — all three simplified as suggested and pushed:

- **`valueUnit` → open `pageLabel?: string`** (defaults to localized "Page"). Set `pageLabel="Row"` to relabel the box.
- **Removed `formatLabel`** (and the `PaginationState` it took) — the label is just `pageLabel` + the editable box + `/ N`.
- **Removed `onRowNavigate`** and the row→page math — navigation always flows through the existing `onChange(page)`, so a "Row" label is just a relabeled page-navigated box. `onChange` unchanged, no new callback.

Net: the `input` variant is now `« ‹ [pageLabel] [ n ] / N › »`, page-based, one small open prop instead of three. −324/+110 lines. Tests + typecheck green.

_— on behalf of Freddy_

packages/core/src/Pagination/Pagination.tsx
```diff
@@ -212,6 +241,66 @@ const styles = stylex.create({
...
+      '@media (pointer: coarse)': `max(1rem, ${typeScaleVars['--text-label-size']})`,
+    },
+    color: colorVars['--color-text-primary'],
+    backgroundColor: colorVars['--color-background-surface'],
+    boxShadow: {
+      default: 'none',
+      ':focus-visible': `inset 0px 0px 0px 2px ${colorVars['--color-accent-muted']}`,
+    },
+    outline: 'none',
+  },
```
cixzhang: I guess we should try to reuse NumberInput here

cixzhang (approved):

packages/core/src/Pagination/Pagination.tsx
freddymeta: Done — the editable box is now the core `NumberInput` (`min={1}`, `max={totalPages}`, `isIntegerOnly`, label-hidden). Dropped the hand-rolled parse/commit/clamp state since NumberInput owns typing, parsing, and range.

One behavior note worth calling out: NumberInput **rejects and reverts** an out-of-range entry rather than clamping to the nearest bound (the old bare input clamped `99 → 10`). So committing an invalid page now snaps back to the current page instead of silently substituting a different one. I think that's the better UX (no surprise value swap), but flagging since it's a change — happy to revisit if you'd prefer clamping.

Also rebased clean onto `main` (was behind after the RTL refactor): the first/last chevrons now mirror via `rtlStyles.mirror` like prev/next, no `useDirection`.

_— on behalf of Freddy_

freddymeta:
@cixzhang — heads-up on a change I pushed **after your approval**, so it doesn't slip in unseen. The `input` variant's editable box was reworked to **compose `NumberInput`** instead of the hand-rolled `<input type="text">` + manual parse/clamp:

- The box is now a `NumberInput` (a `spinbutton`) with native `min={1}` / `max={totalPages}`, so range clamping and typing/pending state are owned by the primitive rather than re-implemented in Pagination. Net `Pagination.tsx` −151 lines.
- Removed the now-redundant **`astryx-pagination-input`** theme target (the outer wrapper class) — `NumberInput` brings its own input theming; `astryx-pagination-input-label` and `astryx-pagination-input-total` (the bespoke "Page" label and "/ N" total) stay.
- Behavior is unchanged: `« ‹ [Page] [ n ] / N › »`, page-based, commits on Enter, clamps to range, `onChange(page)` only.

This is the composition-over-bespoke direction, so I think it's strictly better than the version you approved — but it does change the public theming surface (one fewer target) and swaps the box's a11y role (`textbox` → `spinbutton`), so I didn't want it landing silently under the prior sign-off. Tests updated (the box is now queried as `spinbutton`), CI green, `review-required` cleared.

If you're good with it, it's ready to land as-is.

_— on behalf of Freddy_

freddymeta:
Added one more thing to close the last gap vs the EPS POC wrapper (`eps-astryx-poc/src/custom/Pagination.tsx`) — @cixzhang, small addition on top of the approved change:

**Hover tooltips on the carets.** The icon-only first/last/prev/next buttons exposed their label only to assistive tech (`aria-label`); sighted users got no hover hint. That was the *entire reason* the EPS POC re-composed the control row from primitives — Astryx's carets had "no `tooltip` and no slot to reach them." Now each caret forwards the **same localized, step-aware label** it already computes to `Button`'s existing `tooltip` prop:

- step 1 → "Go to previous/next page"; step N → "Go back/forward N pages" (reuses `previousBy`/`nextBy`, pluralized) — so the tooltip can never disagree with the button's actual stride.
- Gated on the enabled state (`tooltip={isDisabled || !hasX ? undefined : label}`) so a disabled caret keeps **native `disabled`** semantics rather than becoming a focusable `aria-disabled` dead control (Button switches to aria-disabled when a tooltip is present, to stay hoverable — we don't want that on a disabled pager arrow).

No new API — rides entirely on `Button`'s existing `tooltip`. This lets the POC drop its bespoke wrapper: everything it re-composed (RTL flip, i18n labels, optimistic page, first/last, and now the caret tooltip) is upstream. 92/92 tests green, typecheck + lint clean.

Separately: while verifying this I found the caret icons render a few px above center (the RTL mirror wrapper dropped them out of the button's flex-centering) — that's a pre-existing bug from the mirror change, not this PR, so I fixed it independently in #4723 to keep this one focused.

_— on behalf of Freddy_

cixzhang (approved):

PR #4509, opened 2026-07-29
feat(lab): add getMarkdown() to RichTextEditor imperative ref

potatowagon:
## What

Adds `getMarkdown()` to the `RichTextEditor` imperative ref handle (`RichTextEditorRef`). It serializes the current editor content to a Markdown string.

## Why

Parity with EPS `RichTextArea` (whose ref exposes `getMarkdown()`), and the natural next step after the ref API (#4417) and the `transformers` prop (#4416) landed.

Crucially, `getMarkdown()` uses **the same `transformers` the editor is configured with** — not a hardcoded default. This mirrors how Lexical itself works: in the `lexical-playground`, a single transformer array feeds all three markdown operations (shortcut typing via `registerMarkdownShortcuts`, import via `$convertFromMarkdownString`, and export via `$convertToMarkdownString`). So a consumer who passes custom `transformers` gets consistent behavior across typing shortcuts *and* markdown serialization.

## How

- Added `getMarkdown: () => string` to `RichTextEditorRef`.
- `EditorRefBridge` now receives the memoized `transformers` array and implements `getMarkdown()` as `editor.getEditorState().read(() => $convertToMarkdownString(transformers))` — run inside a read context, as the `$`-prefixed API requires.
- `$convertToMarkdownString` comes from `@lexical/markdown` (already an optional peer dep) — a **subpackage** that resolves to built dist, so this adds **no top-level `lexical` value import** (which would otherwise break the sandbox build via lexical's raw-TS `declare` fields — the issue fixed in #4417/#4502).

- `RichTextEditor.doc.mjs` — updated the ref best-practice note to mention `getMarkdown()`.
- `RichTextEditor.stories.tsx` — added a `getMarkdown()` button to the `ImperativeRef` story.
- `RichTextEditor.test.tsx` — added 3 tests: plain-text serialization, heading with default transformers (`# Title`), and a custom empty-transformers set (heading emits as plain `Title`, proving the prop is honored).

## Notes

- `getHTML()` is split into a separate follow-up PR (it adds `@lexical/html` as a new optional peer dep), keeping this one dependency-clean.

### DIFF
--- a/apps/storybook/stories/RichTextEditor.stories.tsx
+++ b/apps/storybook/stories/RichTextEditor.stories.tsx
@@ -186,6 +186,14 @@ export const ImperativeRef = {
           <button
             type="button"
             onClick={() => {
+              const md = ref.current?.getMarkdown();
+              setReadout(`getMarkdown():\n${md}`);
+            }}>
+            getMarkdown()
+          </button>
+          <button
+            type="button"
+            onClick={() => {
               const editor = ref.current?.getEditor();
               setReadout(
                 `getEditor() -> ${editor ? 'LexicalEditor instance ✓' : 'null'}`,
--- a/packages/lab/src/RichTextEditor/RichTextEditor.doc.mjs
+++ b/packages/lab/src/RichTextEditor/RichTextEditor.doc.mjs
@@ -153,7 +153,7 @@ export const docs = {
       {
         guidance: true,
         description:
-          'Use a ref (RichTextEditorRef) to imperatively focus(), clear(), read the state via getEditorState(), or reach the LexicalEditor via getEditor(). The handle is available after mount. focus() and clear() are no-ops when the editor is read-only or disabled, and clear() resets to a single empty paragraph.',
+          'Use a ref (RichTextEditorRef) to imperatively focus(), clear(), read the state via getEditorState(), serialize to Markdown via getMarkdown(), or reach the LexicalEditor via getEditor(). The handle is available after mount. getMarkdown() uses the same transformers prop the editor is configured with. focus() and clear() are no-ops when the editor is read-only or disabled, and clear() resets to a single empty paragraph.',
       },
       {
         guidance: false,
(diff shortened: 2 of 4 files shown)


humbertovirtudes (approved):
Nice, clean addition. `getMarkdown()` is the natural completion of the ref API after #4417 and #4416, and I like that it serializes through the same memoized `transformers` the editor is configured with rather than a hardcoded `TRANSFORMERS` default. That keeps typing shortcuts, import, and export all consistent for a consumer who layers in custom transformers, which mirrors how Lexical's own playground wires a single array through all three operations.

Triage: additive API on an existing lab component, non-breaking, low blast radius, fast path.

What's good:
- API convergence: matches EPS `RichTextArea`'s `getMarkdown()` for parity, and the name sits naturally next to `getEditorState()` / `getEditor()`. No new props, so no boolean/naming surface to worry about.
- Correct Lexical usage: `$convertToMarkdownString` is run inside a `.read()` context, and the bridge's `useImperativeHandle` dep array is `[editor, editable, transformers]`, so no stale closure and no re-register thrash.
- Dependency-clean: uses the `@lexical/markdown` subpackage (already an optional peer dep) and avoids a top-level `lexical` value import, so it sidesteps the raw-TS `declare` sandbox-build issue from #4417/#4502. The inline comment explains that "why", which is exactly the kind of non-obvious context a comment should carry.
- Tests hit the contract, not the implementation: plain-text serialization, heading with the default transformers, and the empty-transformers case that proves the prop is honored (heading emits as plain `Title`). Full surface updated: `.tsx`, `.test.tsx`, `.doc.mjs`, and the ImperativeRef story.

Not redundant: #4544 adds standalone serializer helpers that work without a mounted editor and explicitly complements this ref method, so the two are orthogonal.

No changeset, which is correct here: `@astryxdesign/lab` is `private`, and the sibling lab PR #4416 landed the same way.

Non-blocking / nit (optional, no need to hold for it):
- The two heading tests seed content via `$convertFromMarkdownString('# Title', TRANSFORMERS)` inside an `editor.update()` rather than through the public shortcut path. That's a reasonable way to get a deterministic heading node in jsdom, so no change needed, but a one-line note in the second test on why it bypasses shortcuts would save the next reader a beat.

No blocking issues. The `needs:code-review` gate is auto-applied for the API label, but this is a well-tested, dependency-conscious, additive lab change with clear sibling precedent.

Recap:
- PR: #4509 feat(lab): add getMarkdown() to RichTextEditor imperative ref
- Author: @potatowagon
- What: adds `getMarkdown()` to `RichTextEditorRef`, serializing current content through the editor's configured `transformers`.
- Recommendation: Approve. Green across build/test/lint/smoke; scope is clean.

PR #4510, opened 2026-07-29
feat(template-viewer): Lightweight app to preview CLI templates

kentonquatman:
Adds `apps/template-viewer`, a tiny Vite app for previewing any template from `packages/cli/templates` in the browser. It replaces the heavier `apps/sandbox` for template previews.

### How it works
`src/App.tsx` globs every `.tsx` under `packages/cli/templates` and matches the one whose directory is named in the URL path, then lazy-renders it inside `<Theme>` + `<LayerProvider>` using `@astryxdesign/theme-neutral`.
Navigating to `/` renders a fallback index page listing every available template.
StyleX is compiled directly from `@astryxdesign/core` source via `astryxStylex()` — no codegen, Babel, PostCSS, or Next.js.
`vite.config.ts` wires up `astryxStylex()` + `@vitejs/plugin-react` and widens `server.fs.allow` so Vite can serve templates that live outside the app root.
Usage
```
pnpm -F @astryxdesign/template-viewer dev
```
Then open a template directory as the URL, e.g. `http://localhost:5173/packages/cli/templates/pages/ai-chat`.

Requires `@astryxdesign/build`, `@astryxdesign/core`, and `@astryxdesign/theme-neutral` to be built once (they ship from dist).

### Changes
New `apps/template-viewer` package: `package.json`, `index.html`, `src/{App,main}.tsx`, `src/index.css`, `vite.config.ts`, three `tsconfig*.json files`, `.gitignore`, and `README.md`.
Updates `pnpm-lock.yaml` with the new workspace entry.
12 files, +284 lines (all additive; no existing files modified except the lockfile).

### DIFF
--- a/apps/template-viewer/tsconfig.app.json
+++ b/apps/template-viewer/tsconfig.app.json
@@ -1,0 +1,31 @@
+{
+  "compilerOptions": {
+    "tsBuildInfoFile": "./node_modules/.tmp/tsconfig.app.tsbuildinfo",
+    "target": "ES2020",
+    "useDefineForClassFields": true,
+    "lib": ["ES2020", "DOM", "DOM.Iterable"],
+    "module": "ESNext",
+    "skipLibCheck": true,
+
+    /* Bundler mode */
+    "moduleResolution": "bundler",
+    "allowImportingTsExtensions": true,
+    "verbatimModuleSyntax": false,
+    "moduleDetection": "force",
+    "noEmit": true,
+    "jsx": "react-jsx",
+
+    /* Linting */
+    "strict": true,
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
+    "noFallthroughCasesInSwitch": true,
+
+    /* Monorepo: resolve @astryxdesign packages to source */
+    "paths": {
+      "@astryxdesign/core/*": ["../../packages/core/src/*"],
+      "@astryxdesign/core": ["../../packages/core/src"]
+    }
+  },
+  "include": ["src"]
+}
--- a/apps/template-viewer/.gitignore
+++ b/apps/template-viewer/.gitignore
@@ -1,0 +1,4 @@
+node_modules
+dist
+dist-ssr
+*.local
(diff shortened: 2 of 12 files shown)


apps/template-viewer/tsconfig.app.json
```diff
@@ -0,0 +1,31 @@
...
+    /* Linting */
+    "strict": true,
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
+    "noFallthroughCasesInSwitch": true,
+
+    /* Monorepo: resolve @astryxdesign packages to source */
+    "paths": {
+      "@astryxdesign/core/*": ["../../packages/core/src/*"],
+      "@astryxdesign/core": ["../../packages/core/src"]
```
josephfarina, line 27: Should be same for cli so you can read templates directly without having to have the package?

packages/cli/templates/themes/neutral/neutralTheme.ts
```diff
@@ -482,6 +482,37 @@ export const neutralTheme = defineTheme({
...
+    //             StatusDot "accent" is the info/attention color, so it
+    //             pairs with the info badge rather than --color-accent
+    //             (near-black #262626, the darkest offender).
+    //
+    // `neutral` is intentionally NOT overridden: the neutral badge bg is a
+    // near-invisible light gray (--color-background-gray #e5e5e5 / 10% white
+    // wash), fine as a large pill but unreadable as an 8px dot. It keeps the
+    // component default's visible mid-gray (--color-icon-secondary), which is
+    // not among the "too dark" cases.
+    // =========================================================================
```
josephfarina: Prob not intentional =D

apps/template-viewer/tsconfig.app.json
kentonquatman: Templates are already read directly from source via the relative glob `import.meta.glob('../../../packages/cli/templates/**/*.tsx')`, so no built `@astryxdesign/cli` package is needed. A CLI paths alias wouldn't be used — nothing here imports `@astryxdesign/cli`, and it ships `.mjs` rather than TS source. The core alias is only there because the viewer's own `src/ imports @astryxdesign/core/*`. Happy to add it if you'd prefer the consistency though!

cixzhang (approved):

PR #4511, opened 2026-07-29
fix(cli): API-layer input validation, path-safety, and codemod hardening + test backfill

josephfarina:
## Bug hunt + test hardening on the leaf-refactored `api/`

A deliberate dogfooding bug-bash + test-hardening pass over the leaf-split CLI. A fleet of focused probes across every command surfaced a systemic pattern: the leaf refactor guards some paths/inputs in the CLI wrapper but **not** in the API — and `@astryxdesign/cli/api` is a public surface. Each fix pushes the guard down into the API (so direct callers get the same contract) and adds a test.

### Path safety (the guard every write command depends on)
- `assertWithin` now **canonicalizes symlinks** — it realpaths the deepest existing ancestor and re-checks against the realpath'd root. `path.resolve` is purely lexical, so a symlink *inside* the project root pointing outside previously passed the check while the write landed outside. Also rejects a NUL byte in the path. Closes the escape for every command that writes through the guard (swizzle / template / upgrade / theme / layout / agent-docs).

### Per-command input validation + write safety
- **`search()`**: a non-positive/non-integer `limit` was silently treated as "return everything" (`limit > 0 ? slice : all`); now validates it as a positive integer, plus empty query and unknown `--type`, all as `ERR_INVALID_ARGUMENT`. `astryx search` on the CLI is unchanged — this only tightens the direct-API path to match.
- **`swizzle()`**: sanitizes the component name so `..`/separators can't escape the `--output` base; fixes dynamic-import and asset-import rewriting (incl. the theme StyleX subpath the compiler requires).
- **`template()` copy / `init()` scaffold**: refuse to clobber an existing file without `overwrite: true` (`ERR_FILE_EXISTS`); unknown `--agent` now throws `ERR_UNKNOWN_AGENT`.
- **`upgrade()`**: confines the `--path` scan dir to cwd (`--apply` rewrites files in place).
- **`layout`**: rejects unknown `--form` / empty expression; JSX-safe text payloads and multi-root fragment wrapping so generated TSX compiles; a compact-parser depth guard; quote-safe printer round-trip; exit code `1` for an invalid layout in both `--json` and human mode.
- **Coded errors instead of raw `TypeError`/`ERR_UNKNOWN`** across `docs`/`discover`/`blog`/`hook`/`component` non-string args, `resolveTheme` malformed input, `theme add` write failures, `validate-integration` path specs, `doctor` multi-config, and the package scanner's adversarial inputs.
- **`jsonOut`**: serializes the envelope before marking the emission handled, so unserializable data still yields a JSON error envelope rather than empty stdout.

### Codemod correctness
- The `--radius-*` / `--shadow-*` / `--color-*` token-migration codemods no longer corrupt a longer consumer token that merely shares a prefix. The boundary lookahead `a|b|c(?!…)` bound only to the last alternative (and two codemods had no boundary at all); it now wraps the whole alternation `(?:a|b|c)(?!…)`, so only exact token names migrate. Output stayed parseable, so the runner's corruption guard never caught it.
- `migrate-badge-children-to-label` skips a badge that already declares `label` (was emitting a duplicate `label` prop).
- The codemod source scan no longer follows symlinks or descends into generated-output dirs; a broken integration manifest is recorded via `issues()` instead of crashing `Project.load`.

### Agent-docs data integrity
- `injectXdsBlock`/`removeXdsBlock` no longer drop, duplicate, or orphan user content on a malformed managed block (END-before-START, duplicate/nested, or start-with-no-end). They locate a single well-formed block and refuse to touch an ambiguous/half-written file.

### Component discovery
- `readDocMeta` no longer reads a `group:`/`hidden:` field nested inside a translation `propDescriptions` block as the component's group (which leaked a translated string as a group key in the default-English `component --list`). Field regexes now match top-level fields only.

### Tests backfilled
API-level suites for the zero-coverage commands (`component`, `search`, `doctor`), the `hook`/`discover`/`docs` dispatchers, the blog adapter/leaf/CLI wrapper, doctor degradation, manifest determinism, `build` error-code faithfulness, `layout` expand/check, `swizzle` import rewriting, config-codemod isolation, agent-docs malformed-block cases, path-safety symlink escapes, and unit tests for `levenshteinDistance`, the error-envelope contract, `jsonOut` serialization discipline, the `print` round-trip, and the `string-utils` / `config-schema` / `json` / `term-log` modules.

### Verified
Rebased on `main` (post the `./api` typegen change). Full CLI suite green (2432 tests), `typecheck:strict` 0, `typecheck:json-api` pass, `./api` type-surface verify pass, eslint clean.


packages/cli/api/doctor/doctor.test.mjs
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+/**
+ * @file Colocated tests for the `doctor` leaf (api/doctor/doctor.mjs). `doctor`
+ * had no api-level tests; this locks the envelope shape and the summary
+ * invariant (the counts must always add up to the number of checks).
```
josephfarina, line 6: the commnet is not helpful it just describes what it did

cixzhang (approved):

PR #4515, opened 2026-07-29
feat(core): move CodeBlock collapse chevron to the left with a light reveal

### LINKED ISSUE #4513: CodeBlock: move collapse chevron to the left of the title with a light reveal animation
## Summary

Move the collapse (disclosure) chevron for collapsible `CodeBlock` headers to the **left** of the title/language label, rather than trailing it. Leading-icon disclosure placement matches the convention users expect from other collapsible surfaces and makes the affordance easier to scan.

## Current behavior

The chevron currently renders *after* the title and language label (to the right of the text) inside the header title row.

## Desired behavior

- The chevron sits to the **left** of the title/language label.
- When the chevron becomes visible (e.g. the header first renders as collapsible), it is introduced gracefully with a light animation (fade/scale in) so it doesn't pop in and cause UI jank as it takes up layout space.

## Notes

- Keep the existing rotate-on-collapse behavior and keyboard/ARIA semantics (`aria-expanded`, `aria-controls`, focus ring) intact.
- StyleX only; reuse existing motion tokens (`--duration-*`, `--ease-*`).

cixzhang:
Closes #4513

## What

Moves the collapse (disclosure) chevron on collapsible `CodeBlock` headers to the **left** of the title / language label, and eases it into view with a light reveal animation instead of letting it pop in.

## Why

A leading chevron matches the disclosure convention users expect from other collapsible surfaces and makes the affordance easier to scan. Because the chevron only appears once a block is long enough to be collapsible, it can materialize and shift the title — the reveal animation smooths that transition so it doesn't read as jank the first time it becomes visible.

## How

- Reorder the header title contents so the chevron renders before the title / language text.
- Add a StyleX `keyframes` reveal (fade + slight scale-in) on the chevron, using the existing `--duration-medium` / `--ease-standard` motion tokens.

The existing rotate-on-collapse behavior and keyboard/ARIA semantics (`aria-expanded`, `aria-controls`, focus ring) are unchanged.

## Testing

- `CodeBlock.test.tsx` — 15/15 pass (collapse toggle, copy-button isolation, inert/ARIA on the collapsed region).
- ESLint clean on the touched file.

### DIFF
--- a/.changeset/codeblock-collapse-chevron-left.md
+++ b/.changeset/codeblock-collapse-chevron-left.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] CodeBlock: move the collapse chevron to the left of the title/language label and ease it into view with a light reveal animation so it no longer pops in and shifts the header (#4513)
+@cixzhang
--- a/.changeset/codeblock-collapse-chevron-left.md
+++ b/.changeset/codeblock-collapse-chevron-left.md
@@ -2,5 +2,5 @@
 '@astryxdesign/core': patch
 ---
 
-[feat] CodeBlock: move the collapse chevron to the left of the title/language label and grow it into place (width + inline margin) so it slides the title over smoothly instead of popping in and shifting the header. Respects `prefers-reduced-motion` (#4513)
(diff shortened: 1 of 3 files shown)


cixzhang:
Screenshot 2026-07-29 at 1 57 24 PM
Screenshot 2026-07-29 at 1 57 29 PM

PR #4516, opened 2026-07-29
fix(core): add missing xstyle prop support to SegmentedControlItem, SideNavItem, TypeaheadItem

dipankarmaikap:
## Summary

Adds missing `xstyle` prop to three components that support it in similar siblings but were overlooked:

- **SegmentedControlItem** — destructured `xstyle` and passed to root `stylex.props()`
- **SideNavItem** — added `xstyle` prop, spread `...rest`, passed to all three root divs
- **TypeaheadItem** — destructured `xstyle` and passed to root `stylex.props()`

## Test Plan

- [ ] `pnpm test`
- [ ] `pnpm lint`

## Changeset

```
patch — @astryxdesign/core
```


cixzhang (approved):
Thanks, for fixing these!

PR #4524, opened 2026-07-29
fix(fileinput): stop nesting interactive controls in the trigger

### LINKED ISSUE #4522: FileInput: interactive controls nested inside role="button" trigger (nested-interactive, WCAG 4.1.2)
## Summary

`FileInput` renders its trigger as a `role="button"` wrapper, and interactive controls are nested **inside** it:

- the clear button (`InputClearButton`, a real `<button>`) — pre-existing
- the status icon button — added by the `statusVariant="tooltip"` work (#4490 / the input statusVariant PR)

axe flags this as **`nested-interactive`** (serious, WCAG 4.1.2): interactive controls nested inside another interactive control aren't reliably announced by screen readers and can cause focus/activation problems.

## Why not fixed in the statusVariant PR

The statusVariant PR intentionally leaves this alone — the nested clear button predates it, and the correct fix is a structural change to FileInput's trigger that deserves its own PR. Filing it so it isn't lost.

## Options to consider

- Restructure so the trigger's interactive region and the auxiliary controls (clear, status) are **siblings** rather than nesting the controls inside the `role="button"` element — e.g. the clickable/keyboard target is a focusable child, with clear/status buttons positioned alongside it, all within a non-interactive container.
- Or make the wrapper a non-interactive container and move the activation affordance to a dedicated inner control.

Whatever the shape, the acceptance bar is: **no `nested-interactive` axe violations on FileInput stories**, keyboard activation of the picker still works, and the clear + status affordances remain independently reachable.

## Acceptance criteria

- [ ] axe reports zero `nested-interactive` violations for FileInput across its stories.
- [ ] Keyboard: the file picker is still operable, and the clear/status controls are independently focusable/activatable.
- [ ] No regression to the existing FileInput tests + a11y wiring (aria-describedby, required, invalid).

cixzhang:
Closes #4522

## What

`FileInput` rendered its trigger as a `role="button"` wrapper `<div>` and nested real interactive controls **inside** it — the clear button (pre-existing) and, in `statusVariant="tooltip"`, the status icon button. axe flags this as **`nested-interactive`** (serious, WCAG 4.1.2): interactive controls nested inside another interactive control aren't reliably announced by screen readers and cause focus/activation problems.

## How

Adopts the existing **`ClickableCard` / `useClickableContainer`** pattern:

- The container `<div>` is now **non-interactive** — it keeps the styling and drag-and-drop surface, and uses `useClickableContainer` to stay clickable while ignoring clicks that land on nested controls.
- A **visually hidden real `<button>`** sibling carries the accessible role, name, focus, and all the ARIA that used to live on the `role="button"` div (`aria-label` with the filename-composed name, `aria-describedby`, `aria-invalid`, `aria-busy`, and the `aria-disabled`/`tabIndex` disabled semantics). The disabled-message tooltip anchors to this button.
- The clear and status buttons are now plain **siblings** of the trigger inside a non-interactive container — so no interactive element nests inside another.

Click, keyboard activation (Enter/Space), drag-and-drop, and the disabled-message tooltip all behave exactly as before. No prop or public API changes.

## Testing

- All 58 existing `FileInput` tests pass unchanged (the `getByRole('button', …)` queries now resolve to the hidden trigger button — accessible name/role preserved).
- Verified zero nested-interactive across compact, tooltip-status, dropzone, and disabled states.
- `typecheck`, `typecheck:docs`, `lint` (0 warnings), and `prettier --check` all green.

### DIFF
--- a/packages/core/src/FileInput/FileInput.tsx
+++ b/packages/core/src/FileInput/FileInput.tsx
@@ -212,21 +212,6 @@ const styles = stylex.create({
     whiteSpace: 'nowrap',
     borderWidth: 0,
   },
-  // Visually hidden trigger button. It carries the accessible role, name, and
-  // ARIA while staying out of view; the visible focus feedback lives on the
-  // container's :focus-within border. Kept as a sibling (not a wrapper) of the
-  // other controls so no interactive element nests inside another.
-  srOnlyButton: {
-    position: 'absolute',
-    width: 1,
-    height: 1,
-    padding: 0,
-    margin: -1,
-    overflow: 'hidden',
-    clip: 'rect(0, 0, 0, 0)',
-    whiteSpace: 'nowrap',
-    borderWidth: 0,
-  },
   placeholderText: {
     fontFamily: typographyVars['--font-family-body'],
     fontSize: {
@@ -460,9 +445,10 @@ export function FileInput({
   const announce = useAnnounce();
 
   // Disabled-reason tooltip. Disabled controls swallow pointer events, so the
-  // tooltip listeners attach to the focusable trigger button and it stays
-  // perceivable via aria-disabled instead of the native disabled attribute.
-  // Opening the picker is blocked by the isDisabled guards in the handlers.
+  // tooltip listeners attach to the visible container (which stays hoverable)
+  // and the trigger stays perceivable via aria-disabled instead of the native
+  // disabled attribute. Opening the picker is blocked by the isDisabled guards
+  // in the handlers.
   const showsDisabledMessage = isDisabled && !!disabledMessage;
   const disabledMessageTooltip = useTooltip({
     placement: 'above',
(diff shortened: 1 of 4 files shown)


packages/core/src/FileInput/FileInput.tsx
```diff
@@ -211,6 +212,21 @@ const styles = stylex.create({
     whiteSpace: 'nowrap',
     borderWidth: 0,
   },
+  // Visually hidden trigger button. It carries the accessible role, name, and
+  // ARIA while staying out of view; the visible focus feedback lives on the
+  // container's :focus-within border. Kept as a sibling (not a wrapper) of the
+  // other controls so no interactive element nests inside another.
+  srOnlyButton: {
```
cixzhang: use VisuallyHidden instead

PR #4526, opened 2026-07-29
chore(ci): verify redeploy-preview builds the PR head commit (guard against stale previews)

nynexman4464:
## The bug

`redeploy-preview.yml` (a manual `workflow_dispatch` that re-deploys a PR's
Storybook + Sandbox to `/pr/<number>/` on gh-pages) resolved the PR head SHA
**once** with `gh pr view --json headRefOid`, checked it out, then built and
deployed — with **zero verification** that it built the right commit.

A recent run resolved the head to a `main` commit (`9c193a2cbf`) instead of the
actual PR head (`c398a5ac90` for #4520), so it built and deployed **the base
branch** to the PR's preview path. The preview served stale, un-fixed code and
misled a review into thinking the PR's fix was missing.

## The fix

Make the workflow self-verifying and fail **loudly** (non-zero exit) rather than
silently deploy the wrong commit:

- **Resolve step**: fail if `headRefOid` is empty; echo the PR number + resolved
  SHA so every run is auditable.
- **Post-checkout guard** (new step):
  1. Assert `git rev-parse HEAD` equals the resolved head SHA.
  2. Re-query `gh pr view --json headRefOid` and confirm it still equals the
     checked-out `HEAD`. This catches a stale/raced `headRefOid`, a head that
     moved mid-run, and the exact "deployed base instead of head" failure — if
     they disagree, the job aborts before building.
- **Traceability**: the built commit's SHA + subject are written into the
  gh-pages deploy commit message and a per-PR `BUILD_INFO.txt`, so any preview
  can be traced back to the commit it was built from.

Build steps and the existing retry-based gh-pages push logic are unchanged —
caching is not the cause, so no build behavior is touched.

### How it would have caught #4520

With the resolved SHA `= 9c193a2cbf` (a main commit) and checkout landing there,
check (1) would pass (`HEAD == resolved`), but check (2) re-queries the PR head,
which correctly returns `c398a5ac90`. Since `c398a5ac90 != 9c193a2cbf`, the guard
fails the job — no stale deploy, and the run points at exactly what went wrong.

No package touched (CI-only), so no changeset — consistent with prior CI-only PRs.

### DIFF
--- a/.github/workflows/redeploy-preview.yml
+++ b/.github/workflows/redeploy-preview.yml
@@ -39,13 +39,57 @@ jobs:
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          HEAD_SHA=$(gh pr view $PR_NUMBER --repo ${{ github.repository }} --json headRefOid --jq '.headRefOid')
+          set -euo pipefail
+          HEAD_SHA=$(gh pr view "$PR_NUMBER" --repo ${{ github.repository }} --json headRefOid --jq '.headRefOid')
+          if [ -z "$HEAD_SHA" ]; then
+            echo "::error::Could not resolve headRefOid for PR #${PR_NUMBER} — refusing to deploy an unknown commit."
+            exit 1
+          fi
+          echo "Resolved PR #${PR_NUMBER} head commit: ${HEAD_SHA}"
           echo "head_sha=$HEAD_SHA" >> "$GITHUB_OUTPUT"
 
       - name: Checkout PR head commit
         uses: actions/checkout@v7
         with:
           ref: ${{ steps.pr-info.outputs.head_sha }}
+
+      - name: Verify checked-out commit is the PR head
+        env:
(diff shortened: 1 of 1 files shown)


cixzhang (approved):
Approving. Good loud-fail-over-silent-wrong-deploy fix for a real incident.

The post-checkout guard is the right shape: assert `HEAD == resolved SHA`, then re-query the PR head to catch a stale/raced `headRefOid` or a head that moved mid-run. `BUILD_INFO.txt` + the SHA in the deploy commit message make previews traceable. Build/push logic untouched, CI-only so no changeset — consistent with prior CI PRs.

nynexman4464:
Closing — superseded by #4525, which rebuilt the automatic PR-preview path (deploy-preview.yml) to deploy pre-built CI artifacts without a checkout, so the wrong-commit class this guarded against can't occur there. This PR only hardened the manual redeploy-preview.yml escape hatch, which is now vestigial. Happy to revisit as a small follow-up if that manual workflow is kept.

PR #4528, opened 2026-07-29
feat(lint): enforce BaseProps styling passthrough (xstyle/className/style)

cixzhang:
## Summary

Components declare their styling contract by extending `BaseProps`, which includes
`xstyle`, `className`, and `style`. Nothing currently enforces that the implementation
actually forwards those props, so a component can promise all three escape hatches in
its types while silently dropping them at runtime.

This adds `@astryx/require-baseprops-passthrough` as a warning in both ESLint configs.
It stays warning-only while violations remain; promotion to an error should be a
separate, deliberate step after the repository is clean.

## Rule behavior

The rule checks public components with a local props interface carrying the `BaseProps`
styling contract. It reports two failure modes:

- **Unused** — a styling prop is destructured but never referenced.
- **Not forwarded** — a styling prop promised by the type never reaches the root.

`className` and `style` may ride a rest spread onto a native or composed root. `xstyle`
cannot be a native DOM attribute, so an un-destructured `xstyle` only counts when the
rest object is passed to a composed component.

A component can opt out by omitting a styling prop from its type, or explicitly bind an
intentionally unused prop with a leading underscore.

## Current main inventory

Rebased onto current `main`, the rule reports **7 violations across 3 core files** and
none in lab, charts, or richtext:

| File | Props dropped |
|---|---|
| `packages/core/src/TopNav/TopNavMegaMenuItem.tsx:189` | `xstyle`, `className`, `style` |
| `packages/core/src/Typeahead/BaseTypeahead.tsx:375` | `xstyle`, `className`, `style` |
| `packages/core/src/VisuallyHidden/VisuallyHidden.tsx:89` | `xstyle` |

All seven are real contract mismatches; none are false positives:

- `TopNavMegaMenuItemProps` carries all three props, while neither its drawer nor desktop
  root receives them.
- `BaseTypeaheadProps` carries all three props, while the input and popover paths receive
  none of them. `inputXStyle` does not remove the inherited `BaseProps` promise.
- `VisuallyHidden` documents that all styling props are intentionally unsupported, but
  its `Omit` currently excludes only `className` and `style`, leaving `xstyle` public.

[#5563](https://github.com/facebook/astryx/pull/5563) fixed the other eight warnings
previously listed here (`TopNavMenu`, `TopNavMegaMenu`, and `TypeaheadItem`), so they are
not retained as stale inventory.

Known reach limit: styling contracts are resolved from interfaces declared in the same
file. Imported props types are skipped; widening that analysis is a separate change.

## Testing

- RuleTester suite: 13 tests pass.
- ESLint inventory: core 7 warnings in 3 files; lab 0; charts 0; richtext 0.
- Pre-push repository checks pass.
- No `packages/**` changes in this PR.

### DIFF
--- a/internal/eslint-plugin-astryx/README.md
+++ b/internal/eslint-plugin-astryx/README.md
@@ -294,6 +294,34 @@ Recommends adding `letterSpacing` when `fontSize` is defined (common design patt
 
 **Strict mode only.** Helps catch missing letter-spacing in compact text elements.
 
+### `@astryx/require-baseprops-passthrough`
+
+Ensures a component actually forwards the styling props it accepts via `BaseProps`:
+`xstyle`, `className`, and `style`. A component's props may promise consumers these
+escape hatches, but the implementation can silently drop them:
+
+- **Unused** — a styling prop is destructured but never referenced, so the override is dropped.
+- **Not forwarded** — a styling prop the type promises is never destructured and never reaches
+  the root element.
+
+Forwarding paths differ by what each prop _is_:
+
+- `className` and `style` are real DOM attributes, so they survive a `{...rest}` spread onto
+  the root element (native or composed) — that counts as forwarding them.
+- `xstyle` is a StyleX style object, **not** a DOM attribute. It cannot ride a `{...rest}`
+  spread onto a native element (it renders inert); the only valid un-destructured path is a
+  rest spread onto a composed Astryx component, which re-accepts `xstyle` via its own `BaseProps`.
+
+Fix by threading each prop into the root `mergeProps(...)` / `stylex.props(...)` call, or
+forwarding to a composed component.
+
+Scoped to public components (those with a `.displayName`). Opt out by omitting the prop from
+the type, e.g. `Omit<BaseProps, 'className' | 'style'>` (as `VisuallyHidden` does),
+or mark an intentionally-unused binding with a leading underscore (`className: _className`).
+
+Currently `warn` in both tiers because known violations remain on main. Promote it
+deliberately only once the repository is clean.
+
 ## Usage
 
 ### Local Development (Human Mode)
--- a/internal/eslint-plugin-astryx/index.js
+++ b/internal/eslint-plugin-astryx/index.js
@@ -51,6 +51,7 @@ import copyrightHeaderRule from './copyright-header.js';
 import noRawConsoleCliRule from './no-raw-console-cli.js';
 import requireBasePropsRule from './require-base-props.js';
 import requireRefPropRule from './require-ref-prop.js';
+import requireBasePropsPassthroughRule from './require-baseprops-passthrough.js';
 import noHardcodedI18nStringRule from './no-hardcoded-i18n-string.js';
 import i18nKeyFormatRule from './i18n-key-format.js';
 import requireTableSectionRule from './require-table-section.js';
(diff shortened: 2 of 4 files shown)


packages/core/src/DropdownMenu/DropdownMenuRadioGroup.tsx
```diff
@@ -107,7 +108,7 @@ export function DropdownMenuRadioGroup({
     <div
       {...rest}
       role="group"
-      {...mergeProps(stylex.props(styles.group), {className, style})}>
+      {...mergeProps(stylex.props(styles.group, xstyle), {className, style})}>
```
cixzhang, line 110: Double check this. It seems odd

cixzhang: Good catch. The odd part was the object-form third argument — `mergeProps(stylex.props(...), {className, style})` — which is inconsistent with the positional convention the rest of the codebase uses (e.g. `Divider`, `ChatLayoutScrollButton`: `mergeProps(themeProps, stylexResult, className, style)`).

Converted to the positional form here and across the other roots this PR touches:

```tsx
<div
  {...rest}
  role="group"
  {...mergeProps(stylex.props(styles.group, xstyle), className, style)}>
```

`{...rest}` first, then `role="group"` as a hardcoded invariant, then styling merged last — matching the `Divider` ordering.

packages/core/src/TopNav/TopNavMegaMenu.tsx
```diff
@@ -305,7 +308,16 @@ export function TopNavMegaMenu({
...
   if (renderMode === 'drawer') {
-    return <DrawerMegaMenu label={label} items={items} featured={featured} />;
+    return (
+      <DrawerMegaMenu
+        label={label}
+        items={items}
+        featured={featured}
+        xstyle={xstyle}
+        className={className}
+        style={style}
```
cixzhang: it's odd to apply this to drawer mode. It might only make sense to apply the styles to the top bar button mode.

packages/core/src/TopNav/TopNavMenu.tsx
```diff
@@ -432,7 +435,12 @@ export function TopNavMenu({
   // Drawer mode: collapsible section
   if (renderMode === 'drawer') {
     return (
-      <div {...stylex.props(drawerStyles.section)}>
+      <div
+        {...mergeProps(
+          stylex.props(drawerStyles.section, xstyle),
+          className,
+          style,
+        )}>
```
cixzhang: same feedback as the top nav mega menu

packages/core/src/TopNav/TopNavMegaMenu.tsx
cixzhang: Agreed — scoped it. `TopNavMenu`/`TopNavMegaMenu` render structurally different elements per view (drawer section vs. desktop trigger button vs. hidden mobile-bar), so applying consumer `xstyle`/`className`/`style` to more than one view is ambiguous.

Restricted styling passthrough to the **default desktop trigger button** — the one stable surface — and left drawer/mobile-bar unstyled, with a comment documenting the scope. Same treatment applied to `TopNavMenu` (your other comment).

packages/core/src/TopNav/TopNavMenu.tsx
cixzhang: Same fix applied here: styling passthrough now targets only the default desktop trigger button; drawer and mobile-bar modes are left unstyled. See the reply on the TopNavMegaMenu thread for the rationale.

cixzhang:
Direction: keep the invariant, but refresh this branch against current main before it advances. Re-measure the now-stale finding inventory and leave the rule warning-only until the repo is clean enough to promote it deliberately.

cixzhang (commented):
Thanks — both nits fixed. I reverted the element change locally and the new test does catch it. Approve once you undraft.

[Reviewed by Robohands]

Issue #4531, opened 2026-07-30
[RFC] ListInput — compact editor for small repeated records


ernestt:
## Problem Statement

Product forms often need to edit a short, ordered collection of consistent records: guest lists, traveler details, mailing-list entries, emergency contacts, tag options, and similar data. Each record usually has one or two simple fields, and users need to add, remove, edit, validate, and sometimes reorder rows without leaving the form.

Astryx currently provides the individual inputs and display-table primitives, but consumers must independently assemble the repeated-row layout, controlled collection updates, responsive behavior, focus policy, accessible names, validation placement, and reorder interactions. That creates substantial boilerplate and makes accessibility quality depend on every caller implementing the same behavior correctly.

## Evidence of Demand

The pattern recurs across several product-neutral scenarios:

- Guest lists: name + email
- Traveler information: full name + passport country
- Tag options: color + label
- Mailing lists: display name + address
- Emergency contacts: name + phone

The intended design guidance is:

- Normally fewer than seven records
- Normally fewer than three fields per record
- A consistent field schema across rows
- Simple controls such as a single-line input, selector, checkbox, or switch

Those thresholds are guidance, not runtime restrictions. Larger consistent datasets should generally use `Table`; heterogeneous or complex records should generally use cards, tabs, or another progressive form pattern.

Public form libraries expose state primitives for repeated fields (for example [Formik FieldArray](https://formik.org/docs/api/fieldarray), [React Hook Form useFieldArray](https://react-hook-form.com/docs/usefieldarray), and [Ant Design Form.List](https://ant.design/components/form/)), but a design-system component still needs to own the visual, interaction, and accessibility contracts.

## Why Existing Components Don't Cover This

- `Table` supplies semantic rows, columns, sizing, and overflow, but not row creation/removal, controlled item updates, collection validation, accessible reordering, or focus restoration.
- `FormLayout` arranges independent controls but does not model a repeated ordered collection.
- `List` is a display/navigation collection, not a form input.
- `Tokenizer` represents scalar selections as tokens; it does not edit multiple fields within each record.

A recipe can demonstrate one instance, but every caller would still need to rebuild the same state transitions and accessibility behavior.

## Proposed API: `ListInput`

API arbitration tested three mental models across guest, traveler, tag-option, dialog, narrow-container, and migration scenarios:

| Candidate | Result |
|---|---:|
| Data-driven `ListInput` | 38/40 |
| Compound `EditableList` | 33/40 |
[...]

PR #4535, opened 2026-07-30
feat(lab): ListInput, a compact editor for small repeated records

AKnassa:
Implements step 1 of the ListInput RFC (#4531): an experimental implementation in `@astryxdesign/lab`.

A compact editor for a short, ordered collection of consistent records (guest lists, tag options, emergency contacts), using the RFC's provisional API.

- Data-driven typed columns: `renderInput` receives `{item, index, updateItem, label, status, ...state}`; `renderValue` covers read-only display
- Semantic table with column headers associated to cells, per-cell input labels combining column and item position ("Email, guest 2"), and the collection named and described through `Field`
- `onChange(next, change)` where `change` distinguishes add, update, remove, and reorder, carrying the stable item key and the relevant indexes or column key
- Focus policy: Add focuses the new row's first control; Remove focuses the next remove action, the previous one at the end, or Add when the list empties
- Reorder behind `isReorderable`: pointer, touch, and keyboard share one draft order and one commit path, so `onChange` fires once per completed reorder. Keyboard flow is grab, arrow move, drop, escape cancel, with polite live-region announcements; focus stays on the moved row's handle
- Validation at three scopes that coexist: field (rendered by the cell input), item (full-row message tied to its row), list (rendered after the rows and the Add action). Removal is never blocked by validity; `maxItems` disables Add at the limit
- `isReadOnly`, `isDisabled`, `isLoading` with distinct semantics (values only, disabled controls, `aria-busy`)

Includes 31 colocated behavior tests, `ListInput.doc.mjs`, and Storybook scenarios under Lab/ListInput (Showcase, Validation, ReadOnly, Loading). No changeset since lab is canary-only.

Out of scope, per the RFC plan: theme hardening review, RTL and narrow-container visual passes, and core graduation.

### DIFF
--- a/apps/storybook/stories/ListInput.stories.tsx
+++ b/apps/storybook/stories/ListInput.stories.tsx
@@ -1,0 +1,137 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+import type {Meta, StoryObj} from '@storybook/react';
+import {useState} from 'react';
+import {ListInput} from '@astryxdesign/lab';
+import type {ListInputColumn} from '@astryxdesign/lab';
+import type {InputStatus} from '@astryxdesign/core/Field';
+import {Selector} from '@astryxdesign/core/Selector';
+import {TextInput} from '@astryxdesign/core/TextInput';
+
+const meta: Meta<typeof ListInput> = {
+  title: 'Lab/ListInput',
+  component: ListInput,
+  tags: ['autodocs'],
+  parameters: {
+    layout: 'centered',
+  },
+  decorators: [
+    Story => (
+      <div style={{width: 640, maxWidth: '100%'}}>
+        <Story />
+      </div>
+    ),
+  ],
+};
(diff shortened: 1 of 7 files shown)


cixzhang:
Closing — `ListInput` landed in lab on Aug 7 via [#4740](https://github.com/facebook/astryx/pull/4740), which closed the RFC ([#4531](https://github.com/facebook/astryx/issues/4531)) this PR implements step 1 of. `packages/lab/src/ListInput/` on main is the data-driven typed-column model, so the two are the same component built twice.

Sorry @AKnassa — this was a substantial piece of work (the reorder commit path and the three validation scopes especially) and it shouldn't have sat while a second implementation went in. That's a process failure on our side, not yours. If you spot behavior the landed version handles worse — the focus policy after Remove, or the single-`onChange`-per-reorder guarantee — those are worth filing against the shipped component directly.

PR #4538, opened 2026-07-30
feat(treelist): reveal endContent on hover/focus via endContentReveal

freddymeta:
## What

Adds a tree-level **`endContentReveal`** prop to `TreeList` — `'always' | 'hover'`, default `'always'` (non-breaking).

With `endContentReveal="hover"`, each row's `endContent` is hidden at rest and revealed when the row is **hovered** or **focused**, and stays visible on **touch**. This is for secondary row actions (e.g. a delete button) that would otherwise clutter the tree at rest.

```tsx
<TreeList
  endContentReveal="hover"
  items={[
    { id: 'report', label: 'report.pdf', onClick: open,
      endContent: <Button isIconOnly label="Delete report.pdf" icon={<Icon icon={TrashIcon} />} onClick={remove} /> },
  ]}
/>
```

## Is this accessible? Yes — by construction

Hover-to-reveal is only an a11y problem when done naively. This implementation avoids every failure mode:

- **Keyboard**: revealed on `:focus-within`, so tabbing into the row (or into the action itself) shows it. Not pointer-only.
- **Never unmounted**: opacity-only — never `display:none`/`visibility:hidden`, which would drop the control from the tab order and a11y tree. The action stays mounted and focusable, so it can receive the focus that triggers its own reveal.
- **Touch**: stays visible on `@media (any-pointer: coarse)`, so the action is never gated behind an unavailable hover.
- **Pointer**: the hover branch is guarded by `@media (hover: hover)`.
- **Motion**: honors `prefers-reduced-motion` (0s transition).

## Approach

Mirrors the already-shipped, reviewed `Thumbnail` `showRemoveOn="hover"` pattern exactly:

- CSS-driven via `stylex.when.ancestor(':hover' / ':focus-within', treeItemScope)`.
- A scoped `treeItemScope` marker (`stylex.defineMarker()`) is applied per row, so a **nested** tree's rows react to their own hover/focus — not an ancestor row's, and not an outer container's.

## Changes

- `endContentReveal` prop on `TreeList` (threaded to `TreeListItem`); default `'always'`.
- New `treeList.markers.stylex.ts` scope marker.
- `TreeListEndContentReveal` type, exported from the package.
- Reveal style on the endContent span; marker on the row (only when `hover`).
- Tests: mounted/focusable contract in `hover` mode + class-application + default behavior.
- Docs: prop across all three doc variants, a best-practice note, and a **Hover Actions** showcase block.
- Changeset (patch).

## Testing

- `TreeList` suite: 69 passed (incl. 3 new). Showcase manifest tests: 12 passed.
- `typecheck:docs`, `typecheck:template-docs`, `check-sync`, `sync-exports --check`, token-docs `--check`, changeset check, eslint — all green.
- Verified emitted CSS: `opacity:0` at rest; `opacity:1` under `:hover *` (guarded by `@media (hover: hover)`), `:focus-within *`, and `@media (any-pointer: coarse)`.

### DIFF
--- a/.changeset/treelist-endcontent-reveal.md
+++ b/.changeset/treelist-endcontent-reveal.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] TreeList: add `endContentReveal` (`'always' | 'hover'`, default `'always'`). With `'hover'`, each row's `endContent` is hidden at rest and revealed on row hover or keyboard focus (and stays visible on touch), so secondary row actions like a delete button don't add visual noise at rest. The reveal is CSS-driven and scoped per row (so nested trees react to their own hover/focus), keeps the content mounted and focusable for accessibility, and honors `prefers-reduced-motion`.
+@freddymeta
--- a/packages/cli/templates/blocks/components/TreeList/TreeListHoverActions.doc.mjs
+++ b/packages/cli/templates/blocks/components/TreeList/TreeListHoverActions.doc.mjs
@@ -1,0 +1,14 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+/** @type {import('../../../../../core/src/docs-types').TemplateDoc} */
+export const doc = {
+  type: 'block',
+  exampleFor: 'TreeList',
+  name: 'TreeList — Hover Actions',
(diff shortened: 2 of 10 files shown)


freddymeta:
Closing — no longer necessary.

PR #4539, opened 2026-07-30
feat(input): add theme targets for the status icon (on-field + detached)

freddymeta:
## What

The input status glyph (the on-field warning / error / success icon) is rendered in two places, and **neither had a theme target** — so a theme author could not recolor, resize, or otherwise restyle it via `defineTheme`. Icon size in particular (StyleX `width`/`height`/`fontSize` applied on the icon element itself) was unreachable: a parent-class override doesn't cascade into the glyph.

This PR adds two additive sub-element theme targets:

- **`astryx-input-status-icon`** — on the on-field status icon (`useInputStatusIcon`), shared by all bordered inputs (TextInput, TextArea, NumberInput, DateInput, DateTimeInput, DateRangeInput, TimeInput, FileInput) across the `attached` and `tooltip` status variants. Reflects `data-size` and `data-status`.
- **`astryx-field-status-icon`** — on the detached message box's leading icon (`FieldStatus`), shown when `statusVariant="detached"` suppresses the on-field glyph. Reflects `data-type`, mirroring the parent `astryx-field-status`.

The targets sit on the `<Icon>` element itself so same-element rules in `@layer astryx-theme` win over the icon's own base styles — a field-level target could not reach them.

## Why

This follows the established sub-element icon-target pattern already used by `banner-icon` (per-status) and the DateInput clear/toggle glyph targets: status icons should be themeable like other sub-element icons, and today their size isn't reachable.

## Example

```ts
defineTheme({
  name: 'compact',
  components: {
    // On-field status icon (attached + tooltip variants, all bordered inputs)
    'input-status-icon': {
      base: { width: '1rem', height: '1rem', fontSize: '1rem' },
    },
    // Detached message-box leading icon
    'field-status-icon': {
      base: { width: '1rem', height: '1rem', fontSize: '1rem' },
    },
  },
});
```

Per-status targeting is also possible via the reflected data attributes, e.g. `'status:error'` on `input-status-icon` or `'type:error'` on `field-status-icon`.

## Non-breaking

Purely additive. No API changes and no default visual change — the targets only add stable classes plus reflected `data-*` attributes. Existing default rendering is byte-identical until a theme targets them.

## Testing

- `useInputStatusIcon` and `FieldStatus` unit tests extended to assert the target class + reflected data attributes on both render sites (and that each is absent in the opposite variant).
- `themingTargets` guard (275 tests) stays green — both classes are documented (`astryx-input-status-icon` in `Field.doc.mjs`, `astryx-field-status-icon` in `FieldStatus.doc.mjs`).
- Local: core test + typecheck + typecheck:docs, storybook typecheck, docsite generate + test, lint, sync-exports/verify-exports, check:sync, check:changesets, generate-token-docs, full build — all green.

### DIFF
--- a/.changeset/input-status-icon-theme-target.md
+++ b/.changeset/input-status-icon-theme-target.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] Field/FieldStatus: add `astryx-input-status-icon` and `astryx-field-status-icon` theme targets on the field status glyph, so consumers can recolor, resize, and restyle it — per status — via `defineTheme` instead of a fragile descendant selector or raw CSS. `astryx-input-status-icon` sits on the on-field icon shared by all bordered inputs across the `attached` and `tooltip` status variants and reflects `data-size`/`data-status`; `astryx-field-status-icon` sits on the detached message box's leading icon and reflects `data-type`. Purely additive — default rendering is unchanged.
+
+@freddymeta
--- a/packages/core/src/Field/Field.doc.mjs
+++ b/packages/core/src/Field/Field.doc.mjs
@@ -30,6 +30,10 @@ export const docs = {
       {className: 'astryx-field', visualProps: ['layout']},
       {className: 'astryx-field-label'},
       {className: 'astryx-field-status', visualProps: ['type', 'variant']},
+      {
+        className: 'astryx-input-status-icon',
+        visualProps: ['size', 'status'],
+      },
     ],
     vars: [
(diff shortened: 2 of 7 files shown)


cixzhang (approved):
Nice — this fills the status-icon theming gap the other inputs already have. Putting a stable target on the icon element itself (on-field via astryx-input-status-icon, detached via astryx-field-status-icon) is exactly right so a same-element rule can win over the icon's base sizing, and reflecting data-size/data-status/data-type per status mirrors how astryx-banner-icon and the date-input/selector icon targets work. Purely additive, default rendering untouched, docs + tests + changeset all in place. LGTM.

PR #4540, opened 2026-07-30
feat(treelist): flush leaves in all-leaf groups + themeable inter-row gap

freddymeta:
## What
Two upstream TreeList fixes that let consumers drop fragile parity CSS:

1. **Flat-tree leaves render flush.** A leaf reserved a chevron-alignment column (chevron width + gap) on top of its level indent unconditionally — even when nothing in the tree had a caret to align under. That reservation is now a **whole-tree** decision: a leaf reserves the column whenever the tree contains **any** expandable item anywhere (checked once at the root, since any expandable descendant forces its whole ancestor chain to be expandable). So **only a fully flat tree** — no expandable items at all — renders its rows flush; any tree with at least one caret keeps the same per-level alignment as before, including all-leaf groups nested under a folder (they stay indented past their parent's label).

2. **Themeable inter-row gap `--tree-list-row-gap` (default `2px`, `var(--spacing-0-5)`).** Opens a subtle vertical separation between rows, matching the inter-row gap `List` and `DropdownMenu` already ship. The gap rides **collapse-proof `padding-block` on the row wrapper** — not `margin-block` on the paintable `tree-list-item` target — so it delivers its full configured value and keeps the paint seam free of layout (per the theming-target guidelines in #4774). The `<li>`s stay contiguous, so the connector guides span the gap by construction; the last-in-group guide is clamped internally so it does not overhang. Set it on the `tree-list` target to widen or close the gap.

## Why
Both are real gaps consumers currently work around with brittle structural CSS (`:has()`/`:not([aria-expanded])` leaf-offset hacks, a `data-*`-gated flat-list reset, and a hand-tuned guide-height + last-child clamp to fake a row gap). Fixing them upstream makes that CSS deletable and reduces the parity layer to sanctioned theme overrides.

## Behavior change
- **Fully flat trees** (no expandable items) will see their rows sit flush — shifted toward the start by the reclaimed chevron column (~24px at the default indent). Any tree with at least one expandable item is **unchanged** in per-level alignment.
- **Every tree** gains the `2px` default inter-row gap. This is a deliberate default shift (it brings TreeList in line with `List`/`DropdownMenu`, both `--spacing-0-5`); a theme can close it to `0px` via the `tree-list` target.

## Notes
- StyleX only; RTL-safe via logical properties.
- `--tree-list-indent` remains the per-level step lever; `--tree-list-row-gap` is orthogonal.
- Internal-only prop `hasExpandableItems` on `TreeListItemInternalProps` (not exported), required (no default) per review.


cixzhang (changes requested):
File-explorer-style full-width rows are a legit look, and your margin→padding approach for keeping labels aligned while the box goes edge-to-edge is exactly the right mechanism. Let's discuss the shape before adding API, though: we're cautious about visual-option props unless there are clear enumerable variations, and `'inset' | 'full'` is really one styling choice. I checked whether a consumer could paint a full-bleed highlight themselves today — and they can't: the row's hover/selected background lives on the *same* element that carries the indent margin, so the highlight is always inset. So this is a real gap. My preferred direction: instead of a `rowHighlight` prop, use your padding-based indent to make the row box full-width structurally, so the existing `astryx-tree-list-item` theme target *becomes* a full-bleed surface consumers can style (`defineTheme` → paint the row background/hover). That way "full-width highlight" is something people style, not a prop we enumerate — and it composes with the hover-reveal work (#4538) and the leaf-flush fix (#4570). Does styling the row target cover your use case if we expose it that way? Happy to pair on where the indent should sit.

freddymeta:
Reworked to your preferred direction — dropped the `rowHighlight` prop entirely. The indent now moves from the row's outer `margin-inline-start` to inner `padding-inline-start`, so the row box is full-width at every depth and `astryx-tree-list-item` becomes the full-bleed surface: a theme paints the row background/hover via `defineTheme`, no enumerated prop.

`--_tree-indent` is unchanged, so labels don't move — only how the distance is applied. And since every box now starts at the edge, a leaf's highlight is flush with its sibling parents (the earlier leaf-flush trial is dropped and folded in here). Changeset reclassified to `[fix]`.

Styling the row target covers the use case, yes. Ready for another look.

cixzhang (commented):
Thanks for reworking this exactly as discussed — dropping the prop and turning the row into a full-bleed surface is the right shape, and I confirmed the margin-based indent genuinely blocked full-bleed theming before, so this is a real enabling fix.

Two notes on where it stands:

Since this changes the default appearance for every consumer — at depth > 0 the row background, hover/pressed overlay, and focus-visible outline now extend edge-to-edge instead of starting at the indent — I'd like design sign-off before it lands (it's already tagged needs:design-review). That's the main gate here; the change is defensible (bigger hit target, clearer selection), it just needs a Design yes on the visual shift.

Separately, you don't have to wait on this PR to get the look you want: you can already achieve the full-bleed file-explorer style today by theming the `tree-list-item` target internally with the same set of styles this PR bakes in — paint the row `background`/hover/selected via `defineTheme` on your own theme. So this is unblocked for your use case regardless of when the default change lands.

One system-level heads-up for whoever picks this up after design signs off (not a blocker, and it predates this PR): the per-level indent is now `padding-inline-start` on the same element that carries the `tree-list-item` theme target, so a theme that sets `padding` on that target will override the indent. Worth a changeset/doc line steering themers to `background`/hover/selected rather than layout padding — or moving the indent onto a dedicated inner wrapper so the two never share a property. Not required for this PR; just flagging so we don't nudge folks straight into it.

cixzhang:
@ernestt @rubyycheung — this one could use a design review before it lands. It moves the `TreeList` per-level indent from `margin-inline-start` to `padding-inline-start` so the row becomes a full-width surface at every depth: the `astryx-tree-list-item` theme target now paints edge-to-edge (file-explorer look), and a leaf's highlight sits flush with its sibling parents'. Labels stay exactly where they were, and there's no API change (the `rowHighlight` prop was dropped per earlier review).

Would appreciate your eyes on the visual treatment — the Storybook preview is linked in the PR.

freddymeta:
i'm gonna re-evaluate our needs ....
i don't like the idea of full bleed crossing over the guide rails.

freddymeta:
@cixzhang — ready for another look. Since your last pass:

- **Rebased clean onto `main`** (0 behind, CI green).
- **Fixed the all-leaf-group indent bug:** the flush optimization is now a whole-tree decision, not per-group — a leaf reserves the chevron column whenever the tree has any expandable item, so an all-leaf group nested under a folder stays indented past its parent (previously its children flushed left of the parent label). Only a fully flat tree flushes. (Renamed the internal prop `hasExpandableSibling` → `treeHasExpandableItems`.)
- **Inter-row gap now defaults to a subtle 2px** (`var(--spacing-0-5)`) instead of 0px, still overridable via the `tree-list` target.

The 2px default and the full-bleed row (edge-to-edge background/hover/focus at depth > 0) both change the default appearance, so this still wants Design sign-off (`needs:design-review`) — that's the main gate. Everything else is addressed.

_— on behalf of Freddy_

freddymeta:
@cixzhang — this branch has been **repurposed to your steer**, so the earlier CHANGES_REQUESTED (against the `rowHighlight` prop) no longer describes it. Re-requesting a fresh look.

**Your ask:** don't add a `rowHighlight: 'inset' | 'full'` prop — instead make the row box full-width *structurally* so the existing `astryx-tree-list-item` target becomes the surface consumers paint via `defineTheme`.

**What this PR is now** (no new visual-option prop):

1. **Leaf chevron-column offset is conditional.** A leaf only reserves the chevron-alignment column when its group actually contains an expandable sibling. All-leaf groups (and fully flat trees) render flush — nothing to align under, so no phantom offset. This is the structural half of "full-bleed": the row box reaches the leading edge on its own, so `astryx-tree-list-item` (which carries the hover/selected background) is now a full-width surface you can paint — no separate `rowHighlight` prop.

2. **Themeable `--tree-list-row-gap`** (default `2px` / `var(--spacing-0-5)`), set on the `tree-list` target. Orthogonal to `--tree-list-indent`. The connector guides span the gap by construction (line stays continuous; last-in-group guide clamped so it doesn't overhang) — so it composes with the guide work rather than needing consumer guide-height tuning.

So "full-width highlight" is now **something you style** (paint `astryx-tree-list-item` via `defineTheme`), not a prop we enumerate — exactly the direction you preferred. No `rowHighlight`, no `'inset' | 'full'` enum.

Rebased clean onto `main` (was 25 behind; resolved a `TreeList.test.tsx` helper conflict from #4657 and fixed a stale test-helper reference). Tests: 75/75 TreeList green, typecheck + lint clean.

One call for you: default row-gap is a **subtle 2px** rather than `0px`. I went with a hair of separation since it reads better with the full-width rows, but it *is* a visual change to existing trees — happy to set it to `0px` (no change to current look) if you'd rather the gap be strictly opt-in.

_— on behalf of Freddy_

freddymeta:
Pushed `adbf40e6c` — fixes the red `lint` check. CI runs eslint in strict mode, which promotes `@astryx/boolean-prop-naming` to an error, and the internal prop I added (`treeHasExpandableItems`) doesn't start with `is`/`has`. Renamed it to **`hasExpandableItems`** (internal-only, on `TreeListItemInternalProps` — no public API change). Strict lint + 75/75 tests + typecheck all green locally now.

On your two earlier notes:

1. **Indent-property footgun** — good news, the rework already sidesteps it. The current PR keeps the per-level indent on `margin-inline-start` (the pre-existing mechanism from #4308, published via `--_tree-indent`), not the `padding-inline-start` the first draft used. So a theme setting `padding` on `tree-list-item` no longer collides with the indent. I also added a best-practice doc line steering themers to rely on the built-in chevron-column reservation rather than nudging indentation with custom CSS, and kept the indent out of the inline style so the theme layer can still override it in normal cascade order if someone really wants to.

2. **Default-appearance / design sign-off** — this is the one real gate left. The scope narrowed since your first pass: there's no `rowHighlight` prop and no full-bleed default change anymore. What's left that touches the default look is the **2px inter-row gap** (`--tree-list-row-gap`, `var(--spacing-0-5)`). Two options and I'll go whichever way you/design prefer:
   - **Ship 2px default** (current) — subtle separation, themeable open/closed via the `tree-list` target.
   - **Default to `0px`** (fully back-compatible, contiguous look preserved) and expose the same themeable lever — zero visual change for existing consumers, opt-in only.

   I lean toward `0px` default + themeable lever to keep this a pure-additive patch and avoid needing a design sign-off to land, but the 2px reads nicer out of the box. Your call — happy to flip it to `0px` in a one-line change if you'd rather land without the design gate.

_— on behalf of Freddy_

cixzhang:
Seems like we'll keep the gap themeable then. Thanks!

cixzhang (approved):

packages/core/src/TreeList/TreeListItem.tsx
```diff
@@ -273,6 +282,16 @@ export interface TreeListItemInternalProps {
...
+  /**
+   * Whether the tree contains at least one expandable item anywhere (i.e. a
+   * caret exists somewhere to align labels under). A leaf reserves the chevron
+   * column — the extra offset that lines its label up past an expandable
+   * ancestor/sibling's caret — whenever this is true, so it stays indented
+   * beyond its parent's label. Only a fully flat tree (no expandable items at
+   * all) has no caret to align under, so its rows sit flush (no chevron-column
+   * offset). Computed once for the whole tree by TreeList.
+   */
+  hasExpandableItems: boolean;
```
cixzhang, line 296: Hmm should default be true for backwards compatibility? Or perhaps `hasOnlyLeaves` as the option?

cixzhang (commented):
🔴 **Blocking** — the row-gap margins sit on `tree-list-item`, the same element that carries `themeProps`. That placement is both a correctness bug (`--tree-list-row-gap` renders at half its configured value, `TreeListItem.tsx:88`) and a seam problem. Moving them onto the existing `rowWrapper` — as `padding-block`, which can't collapse — fixes the measurement and keeps `tree-list-item` a pure paint seam. That's the one thing I'd want fixed before this merges. Also carrying `needs:code-review` + `needs:design-review`.

I'd already approved this and I'm not withdrawing that — the shape is right and this is a one-commit fix. But I went back over the theming half with a finer comb and found a real bug that none of the tests can see.

### 1. 🔴 The row-gap lever delivers half its value

`marginBlock: calc(var(--tree-list-row-gap, 0px) / 2)` on the row box assumes half-above + half-below sums to a full gap. It doesn't — the margins collapse. Neither `rowWrapper` (`position: relative` only) nor `li.wrapper` (`margin: 0; padding: 0`) establishes a BFC, so the row box's block margins collapse out through both, and adjacent `<li>` margins then collapse into each other at `max(g/2, g/2)`.

Measured in Chrome against the exact box structure:

| `--tree-list-row-gap` | rendered separation |
| --- | --- |
| `0px`  | 0px |
| `2px` (default) | **1px** |
| `4px`  | 2px |
| `8px`  | 4px |
| `16px` | **8px** |

The `<li>` height is unchanged at every value, which confirms the margins are outside it — and that means half the gap leaks *above the first row and below the last*, collapsing with whatever the consumer put around the `TreeList`. At a 16px lever that's 8px of phantom margin on each end.

So the documented `2px` default is really 1px, and every theme override is off by 2×. Since `--tree-list-row-gap` is a public, documented lever, the contract in `TreeList.doc.mjs` and the changeset is wrong as written.

The fix I'd reach for is `padding-block` on `rowWrapper`: padding can't collapse, `rowWrapper` carries no theme target, and the `<li>`s stay contiguous so the guides still bridge. That also addresses note 2 below. The guide heights in `TreeListBranches.tsx:54–65` would need retuning if the `<li>` starts absorbing the gap.

### 2. 🟡 Design direction: layout properties on a paint target

Not a blocker on its own and not really this PR's doing — but this PR is where it becomes worth deciding.

A principle I'm converging on: **a theming target is a paint seam** (color, background, border, radius, shadow, font), **not a layout seam** (margin, padding, gap, display, position, sizing). Layout encodes the component's structural contract — overflow, hit-target size, RTL, focus order — and it's exactly what a DOM refactor should stay free to change. Where layout genuinely needs to be themeable, it should go through the declared derived-var / container-padding pipeline rather than a raw class target.

The var here **passes** that test: `--tree-list-row-gap` is a declared, documented var on the `tree-list` root, not a raw `gap` write on a class target. That's the sanctioned mechanism.

The element doesn't. `themeProps('tree-list-item')` sits on the same `contentWrapper` that now carries two layout longhands — `marginInlineStart` (the whole per-level indent) and, new here, `marginBlock` (the gap) — and the comment frames theme override of them as a feature. A theme writing `margin: 0` on `tree-list-item` today zeroes the indent at every depth and flattens the tree; I confirmed that in the browser. That's the padding footgun I flagged on Jul 31, now pointed at `margin` too.

The full-bleed row surface we agreed on is a *paint* seam and I still want it. Relocating the gap per note 1 gets `tree-list-item` most of the way back to purely paintable; moving the pre-existing indent margin off it too can be a separate PR.

### 3. 🟡 Two comments describe a `0px` default that isn't the default

`TreeListBranches.tsx:61-62` ("At the default `0px` gap … the default look is unchanged") and `TreeListItem.tsx:89-90` ("Default `0px` preserves the contiguous look"). The default is `--spacing-0-5`; `0px` is only the `var()` fallback, which never fires. Leftovers from the pre-Aug-4 pass — worth fixing, because the `verticalLast` clamp was reasoned under both of those assumptions.

### 4. 🟡 Tests assert declarations, not geometry

This is why finding 1 got through CI:

- The default-gap test asserts `getPropertyValue('--tree-list-row-gap') === 'var(--spacing-0-5)'` — the declaration text. It passes whether the gap renders at 0px, 1px, or 40px.
- The two guide tests regex the **global** StyleX stylesheet via `collectCssText()`. StyleX emits every `stylex.create` rule on import regardless of whether the class is applied, and neither test checks the guide element's `className` — so both pass even if the `isLast ? verticalLast : verticalFull` branch were reverted. The test grabs `guide`, asserts non-null, then ignores it.
- `reservesColumn` parses the `calc()` string shape (counting `var(`, looking for `+`). Right thing tested, brittle mechanism.

At least one test should assert a measured separation, and the guide tests should assert the applied class.

The flush-leaf tests themselves are good — the named regression test for the all-leaf-group-under-a-folder case is exactly right.

### 5. 🟡 A `derived` mapping would make the obvious spelling work

`derivedVarRegistry.ts` has no `tree-list` entries, so this is consistent with `--tree-list-indent` and not a conformance break. But a theme author's natural first attempt is `rowGap` on the `tree-list` target — and since `.astryx-tree-list` isn't flex or grid, that's a valid declaration that does nothing, silently. A `derived` entry mapping `rowGap` → `--tree-list-row-gap` would fix that and put the lever on the pipeline note 2 prefers.

### 6. 🟢 PR body describes the superseded behavior

The body still says the flush decision is per-group ("only when its group contains an expandable sibling"). The shipped logic is whole-tree, so only fully flat trees flush. The "Behavior change" section therefore overstates the blast radius — worth updating, it undersells the change.

Related: the changeset says "hierarchical trees render exactly as before" in point (1), which point (2) contradicts by adding a gap to every tree.

### On your question from my last pass

`hasExpandableItems` needs no default. `TreeListItem` isn't exported — only `TreeList` is — so there's no back-compat surface, and required is safer than a default that would silently apply old behavior at a forgotten call site. I'd also keep the name over `hasOnlyLeaves`: it's a whole-tree fact, and the root-only computation is correct since any expandable descendant forces every ancestor to have children.

### Process note

These two changes would have been better as two PRs — different blast radius, different review gates, and the bundling is why the changeset ends up asserting something it contradicts. Not worth splitting at this point; just for next time.

Everything else looks good: StyleX only, no raw values, logical properties throughout, no themeable surface removed, and the whole-tree flush decision is genuinely more correct than the per-group version it replaced.

rubyycheung (approved):
Approving — clean, well-scoped change. ✅

Both fixes verified against the current branch:

**Flush leaves in all-leaf groups** — the "check expandability once at the root" optimization is correct: any expandable node forces its whole ancestor chain to be expandable, so a root-level `.some()` reliably detects whether any caret exists to align under. Edge cases (`children: []`, `children: undefined`) are handled consistently with the existing `hasChildren` derivation. Rendered a flat list from source before/after — leaves reclaim the wasted chevron gutter and sit flush, while mixed trees are pixel-identical. Exactly the intended behavior.

**Themeable `--tree-list-row-gap`** — follows the established `--tree-list-*` lever pattern; the guide-spanning math (bridge on non-last rows, clamp the last-in-group connector by half the gap) keeps the connector continuous with no consumer-side tuning.

On the default 2px gap: this is a good call and worth calling out as a **consistency win** — `List` and `DropdownMenu` already ship the same 2px inter-row gap (`--spacing-0-5`), so this brings TreeList in line with the rest of the dense-list family rather than introducing a new value. (Separately, we'll look at whether the dense-list row gap should be codified as a shared convention/token so it's defined in one place — but that's a follow-up, not a blocker for this PR.)

Conventions all met: StyleX-only, RTL-safe via logical properties, LTR + zh docs updated, thorough tests covering both the leaf-offset and row-gap/guide behavior. CI green.

One forward-looking, non-blocking thought: the `collectCssText()` test helper is a nice pattern; if other components start asserting on compiled CSS rules it might be worth promoting to a shared test util.

freddymeta:
Pushed `16c83ff34` addressing the blocking finding and the 🟡 items. Thank you for the finer comb — the margin-collapse bug was real and none of the tests could see it.

**1. 🔴 Row-gap delivered half its value → fixed.** You're right on the mechanism: `marginBlock: calc(--tree-list-row-gap / 2)` sat on `contentWrapper`, and neither `rowWrapper` (`position:relative`) nor the `<li>` (`margin/padding:0`) establishes a BFC, so the block margins collapsed out and adjacent rows collapsed to `max(g/2, g/2) = g/2` — 2px rendered 1px, and half leaked past the first/last row. **Moved it to `padding-block` on `rowWrapper`**, exactly as you suggested: padding can't collapse (full value renders), `rowWrapper` carries no theme target, and the `<li>`s stay contiguous so the guides still bridge.

**Guide-height retune (the part you flagged would need it):** the gap now lives *inside* the `<li>` (padding), so `height:100%` already covers it. `verticalFull` drops back to `calc(100% + 1px)` — just the hairline bridge into the next contiguous sibling; the `+ gap` term is gone because the gap is no longer a void between `<li>`s. `verticalLast` becomes `calc(100% - var(--tree-list-row-gap)/2)` to trim its own bottom padding so it ends at the row-box edge (was `100% + 1px - gap/2`).

**2. 🟡 Layout on a paint target → resolved by (1).** With the gap moved off `tree-list-item`, the only remaining layout longhand on the paint target is the pre-existing `margin-inline-start` indent (#4308). Per your note (and #4774's "P1 refined / paint-vs-layout"), I've left moving *that* to a separate PR as you suggested — this PR gets `tree-list-item` back to a pure paint seam for everything it introduced.

**3. 🟡 Stale `0px`-default comments → fixed.** Both (`TreeListBranches` verticalLast and the row-box comment) rewritten to describe the real `--spacing-0-5` default; the `0px` is only the `var()` fallback.

**4. 🟡 Tests asserted declarations, not geometry → rewritten.**
- The two guide tests now assert the **applied class** on a row's *own* connector: Child 1 (has a sibling below) must carry `verticalFull` and not `verticalLast`; Child 2 (last in group) the reverse. A reverted `isLast ? verticalLast : verticalFull` branch now fails.
- Added a test asserting the gap rides `padding-block` on the row wrapper and **not** `margin-block` on the paint target. I verified it **fails on the pre-fix margin placement** (it reports the `margin-block: calc(var(--tree-list-row-gap)...)` it found), so it's a real guard against a regression to the collapsing form.
- jsdom still can't compute a true pixel separation, so this asserts the collapse-proof *property* + placement rather than a measured px; open to a Playwright/visual assertion if you'd prefer a measured one.

**5. 🟢 `derived` `rowGap → --tree-list-row-gap` mapping.** Held off for now: per #4774's "sanctioned routes that don't work yet," `derivedVarRegistry` is keyed by component only and can't address a sub-element, and there's no `tree-list` derived entry yet (consistent with `--tree-list-indent`). Happy to add it as a follow-up once the registry supports it, or now if you'd rather — your call.

**6. 🟡 PR body + changeset → corrected.** Body now describes the shipped **whole-tree** flush (only fully flat trees flush) instead of the superseded per-group wording, and no longer claims hierarchical trees are unchanged. Changeset point (1)↔(2) contradiction removed: (2) now states plainly that the 2px default shifts every tree (framed as the `List`/`DropdownMenu` consistency win rubyycheung noted).

On the process note — agreed, these were two changes and should've been two PRs; noted for next time. 76/76 tests, typecheck, strict lint all green.

_— on behalf of Freddy_

PR #4544, opened 2026-07-30
feat(lab): add standalone Markdown <-> EditorState serializer helpers

potatowagon:
## What

Adds two standalone Markdown ⇄ `EditorState` serializer helpers to the experimental `RichTextEditor` surface (`@astryxdesign/lab`):

- `markdownToEditorStateJSON(markdown, options?)` → a serialized `EditorState` JSON string (drop-in for the editor's `defaultValue`).
- `editorStateJSONToMarkdown(json, options?)` → a Markdown string.

## Why

The editor already exposes `getMarkdown()` on its ref, but that needs a **mounted** editor. These helpers work **without** one — e.g. to produce a `defaultValue` from Markdown on the server (SSR), or to serialize saved editor state to Markdown off-screen. This is the "generic Markdown ⇄ EditorState helpers could ship in OSS" gap from the EPS → Astryx parity plan.

Both accept the same `{ transformers, nodes }` the editor uses, so custom transformers/nodes round-trip consistently — the same single-source-of-truth model as the `transformers` prop (#4416).

## How

- New `markdownSerializers.ts`: each helper spins up a headless Lexical editor (`createHeadlessEditor` from `@lexical/headless`) and runs `$convertFromMarkdownString` / `$convertToMarkdownString` in it. Headless = no DOM, safe in Node/SSR.
- Extracted `DEFAULT_NODES` into a shared `editorNodes.ts`, imported by both `RichTextEditor.tsx` and the serializers, so both register the same node set (headless conversion matches the mounted editor).
- Added **`@lexical/headless`** as a new **optional peer dependency** (peerDependencies + peerDependenciesMeta `optional: true` + devDependencies), matching the other `@lexical/*` peers.
- `lexical` imports stay **type-only** in the new files — no top-level `lexical` value import, so the sandbox build stays safe (the `declare`-fields trap from #4417/#4502).

## Docs / exports / tests (SYNC contract)

- Exported `markdownToEditorStateJSON`, `editorStateJSONToMarkdown`, and `MarkdownSerializerOptions` from `RichTextEditor/index.ts` and the `@astryxdesign/lab` barrel.
- `RichTextEditor.doc.mjs` — added a usage note.
- `RichTextEditor.test.tsx` — 5 tests: heading parse, full round-trip, custom empty-transformers behavior, and consuming the output as `defaultValue`.

## Testing

CI runs the full build/typecheck/lint/test. (Authored on a devserver without a local pnpm toolchain, so relying on CI. Only `@lexical/headless` + `@lexical/markdown` subpackages are imported — no top-level `lexical` import.)

(tested from local host, the storybook preview doesnt seem to update with the latest deployment)

https://github.com/user-attachments/assets/70b97070-9340-4dc9-b731-e7a00b7f026b

^ call out: all works but the read only one doesnt seem to update. will fix in a later PR as need to merge this PR first to unblock the toolbar work

## Notes

- Independent of #4509 (getMarkdown) — different code path (headless helpers vs. ref method), no file-level overlap beyond the small `DEFAULT_NODES` extraction. Based directly on `main`.

### DIFF
--- a/apps/storybook/package.json
+++ b/apps/storybook/package.json
@@ -23,6 +23,8 @@
     "vega-lite": "^6.4.3"
   },
   "devDependencies": {
+    "@lexical/headless": "^0.46.0",
+    "@lexical/markdown": "^0.46.0",
     "@storybook/addon-docs": "^10.4.6",
     "@storybook/addon-links": "^10.4.6",
     "@storybook/react": "^10.4.6",
--- a/apps/storybook/stories/RichTextEditor.stories.tsx
+++ b/apps/storybook/stories/RichTextEditor.stories.tsx
@@ -5,6 +5,8 @@ import type {Meta, StoryObj} from '@storybook/react';
 import {
   RichTextEditor,
   RichTextView,
+  markdownToEditorStateJSON,
+  editorStateJSONToMarkdown,
   type RichTextEditorRef,
 } from '@astryxdesign/lab';
 import type {EditorState} from 'lexical';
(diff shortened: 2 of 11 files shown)


freddymeta (approved):
Approve — clean, well-scoped addition.

**What it does:** Adds headless `markdownToEditorStateJSON` / `editorStateJSONToMarkdown` helpers so Markdown ⇄ EditorState conversion works without a mounted editor (SSR / off-screen), complementing the ref's `getMarkdown()`.

**Strengths:**
- `lexical` imports stay type-only; runtime imports come only from subpackages (`@lexical/headless`, `@lexical/markdown`), keeping the sandbox build safe.
- Extracting `DEFAULT_NODES` into `editorNodes.ts` as a single source of truth is the right move — headless conversion now registers the same node set as the mounted editor, with a `SYNC:` note guarding drift.
- `@lexical/headless` wired as an optional peer (peerDependencies + `peerDependenciesMeta.optional` + devDependencies), matching the sibling `@lexical/*` peers; lockfile and Storybook deps updated.
- SYNC contract complete: barrel exports (incl. `MarkdownSerializerOptions`), doc note, and 5 tests covering heading parse, full round-trip, `md → json → md` stability, custom empty-transformers, and consuming the output as a live `defaultValue`. That last integration test is the high-value one.
- `{transformers, nodes}` options mirror the editor props, so custom config round-trips consistently.

**Non-blocking notes:**
- The new Storybook story uses raw inline styles / hex — consistent with the existing stories in this file, so no change needed here; a separate pass could tokenize the story chrome if desired.
- The deferred `RichTextView` read-only preview issue called out in the description looks fine to follow up separately; worth confirming it's preview caching and not a `RichTextView` reactivity bug.
- Minor: consider a plain `feat/` branch prefix for future PRs to match the repo's branch-naming convention.

CI is green across build/test/lint/storybook/check-components. LGTM.

freddymeta (approved):
Re-approving — no code changes since the prior review. Type-only `lexical` imports keep the sandbox build safe, the shared `DEFAULT_NODES` extraction is the right single-source-of-truth move, optional-peer dependency wiring matches the sibling `@lexical/*` peers, and the SYNC contract (barrel exports, doc note, 5 tests incl. round-trip + live `defaultValue` consumption) is complete. CI green. LGTM.

cixzhang (approved):

PR #4551, opened 2026-07-30
feat(lab): add composable RichTextEditorToolbar

potatowagon:
## What

Adds a composable **`RichTextEditorToolbar`** to `@astryxdesign/lab` — a formatting toolbar for the experimental `RichTextEditor`. It's built entirely from Astryx primitives (`Toolbar`, `ToggleButton`, `Divider`), so it inherits the active theme with no bespoke styling.

This is the first piece of the "toolbar phase" for the OSS Lexical editor — the piece that unblocks EPS `RichTextArea` → Astryx parity for toolbar-based formatting.

## Design

- **Themed, not hand-rolled.** Composes the existing `Toolbar` (with `role="toolbar"`, aria-label, size coordination) + `ToggleButton` (pressed state, icon-only) + `Divider`. No new visual code, no new icon dependency (inline 16px SVGs).
- **Behavior mirrors the Lexical playground toolbar.** Reads the current selection on `SELECTION_CHANGE_COMMAND` + `registerUpdateListener` to keep active states in sync, and dispatches the standard Lexical commands. Also tracks the editor's editable state (disables controls when read-only).
- **Composable.** Render it in the editor's `plugins` slot; inject product-specific controls (mentions, AI, etc.) via the `endContent` prop — this is the extension seam EPS/XDS adapters need.

Controls (MVP set — intersection of the playground, EPS, and XDS toolbars):
- Inline: bold, italic, underline, strikethrough, inline code
- Blocks: heading (configurable `headingLevels`, default h1–h3) + quote (`$setBlocksType`)
- Lists: bulleted / numbered
- History: undo / redo (gated on `CAN_UNDO`/`CAN_REDO`)

## Usage

```tsx
import {RichTextEditor, RichTextEditorToolbar} from '@astryxdesign/lab';

<RichTextEditor label="Notes" plugins={<RichTextEditorToolbar />} />
```

## Dependencies

- Adds `@lexical/selection` as a new **optional peer dependency** (used for `$setBlocksType`), matching the other `@lexical/*` peers. Its lockfile entry already existed transitively, so only the `packages/lab` importer declaration was needed.
- Everything else (`@lexical/rich-text`, `@lexical/list`, `@lexical/utils`, `lexical`) is already a peer dep.

- Exported `RichTextEditorToolbar` + `RichTextEditorToolbarProps` from `RichTextEditor/index.ts` and the `@astryxdesign/lab` barrel.
- `RichTextEditor.doc.mjs` — added a usage note.
- `RichTextEditor.stories.tsx` — added a `WithToolbar` story.
- `RichTextEditor.test.tsx` — 5 tests: renders the control set, heading-level config, custom accessible label, `endContent` composition, and read-only disabling.

## Testing

CI runs the full build/typecheck/lint/test. (Authored on a devserver without a local pnpm toolchain, so relying on CI.)

## Notes

- Follows the merged imperative-ref / serializer work (#4416/#4417/#4502/#4509/#4550). Based directly on `main`.
- Future toolbar phases (floating toolbar, text-align, link editor, font controls) can layer on top; this establishes the composable base.

### DIFF
--- a/packages/lab/src/RichTextEditor/RichTextEditorToolbar.tsx
+++ b/packages/lab/src/RichTextEditor/RichTextEditorToolbar.tsx
@@ -24,6 +24,13 @@
  * `@lexical/*` are OPTIONAL peer dependencies — install them to use this.
  * Behavior mirrors the Lexical playground toolbar (selection sync + format
  * commands); the UI is built from Astryx primitives so it matches the theme.
+ *
+ * ICONS: Each control resolves its glyph through the core icon registry under a
+ * stable `richtext:*` key (see {@link RICHTEXT_ICON_KEYS}), falling back to the
+ * bundled inline SVGs below. A theme can restyle any glyph by registering its
+ * own icon for that key — no need to fork the toolbar:
+ *   import {registerIcons} from '@astryxdesign/core/Icon';
+ *   registerIcons({'richtext:bold': <MyBoldIcon />});
  */
 
 import {useCallback, useEffect, useState, type ReactNode} from 'react';
@@ -47,6 +54,7 @@ import {$getNearestNodeOfType, mergeRegister} from '@lexical/utils';
 import {Toolbar} from '@astryxdesign/core/Toolbar';
 import {ToggleButton} from '@astryxdesign/core/ToggleButton';
 import {Divider} from '@astryxdesign/core/Divider';
+import {getExtendedIcon} from '@astryxdesign/core/Icon';
 import {
   FORMAT_TEXT_COMMAND,
   UNDO_COMMAND,
@@ -62,13 +70,7 @@ import {
 
 /** Block types the toolbar can toggle. */
 type BlockType =
-  | 'paragraph'
-  | 'h1'
-  | 'h2'
-  | 'h3'
-  | 'quote'
-  | 'bullet'
-  | 'number';
+  'paragraph' | 'h1' | 'h2' | 'h3' | 'quote' | 'bullet' | 'number';
 
 const HEADING_LABELS: Record<'h1' | 'h2' | 'h3', string> = {
   h1: 'Heading 1',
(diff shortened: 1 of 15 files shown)


packages/lab/src/RichTextEditor/RichTextEditorToolbar.tsx
```diff
@@ -0,0 +1,490 @@
...
+const HEADING_LABELS: Record<'h1' | 'h2' | 'h3', string> = {
+  h1: 'Heading 1',
+  h2: 'Heading 2',
+  h3: 'Heading 3',
+};
+
+/** 16px inline icons — no external icon dependency. */
+const icons: Record<string, ReactNode> = {
+  bold: (
+    <svg width="16" height="16" viewBox="0 0 24 24" fill="none" aria-hidden="true">
```
cixzhang: Since they're hardcoded here they won't be themeable. Are there entries to the icon map for these? I think one thing we might want is some way for libraries to augment the icon map with new keys so people can fill in the map with their icons for their theme.

cixzhang (approved):

packages/lab/src/RichTextEditor/RichTextEditorToolbar.tsx
potatowagon: Good call — fixed in 25de205ab. You're right that the inline SVGs weren't themeable, and there was no seam for a library to contribute new keys. Addressed both:

**1. The toolbar now resolves every glyph through the icon registry.** Each control looks up a stable `richtext:*` key (exported as `RICHTEXT_ICON_KEYS`), with the inline SVGs as the *default fallback*. So a theme can restyle any glyph without forking:

```ts
import {registerIcons} from '@astryxdesign/core/Icon';
registerIcons({'richtext:bold': <MyBoldIcon />});
```

**2. The registry can now be augmented with new keys** — which is the "way for libraries to fill in the map with their icons" you asked about. Two changes in `@astryxdesign/core`:
- `registerIcons()` now accepts arbitrary extension keys, not just the built-in `IconName` union (`Partial<Record<ExtendedIconName, ReactNode>>`).
- New `getExtendedIcon(name, fallback)` resolves an extension key, preferring a theme-registered icon and falling back to a library-supplied default. That's the general seam any library can use to make its own icons theme-overridable — the toolbar is just the first consumer.

`getIconRegistry()` still returns only the built-in `IconName` snapshot, so extension keys don't leak into the typed core set.

Added test coverage for both (registry extension-key resolution + a theme-override render test on the Bold control), plus doc notes on `Icon` and `RichTextEditor`. Let me know if you'd rather namespace the keys differently than `richtext:*`.

PR #4553, opened 2026-07-30
fix(cli): sync theme registry with doc targets

AKnassa:
## What this does

`astryx theme build` reads a colon-less component-override rule key as a prop name, so butter's `selected` on `top-nav-item` warns `Unknown prop "selected"`. A colon-less segment is a state, not a `prop:value` pair: core's `parseStyleKey` emits it directly as a state class, and doc targets declare states separately from `visualProps` (`ThemingTarget.states`). The validator now checks only `prop:value` segments, matching the variant generator's existing idiom.

The shipped butter and stone themes build with zero validation warnings as a result, and a test holds them there.

## What changed since this PR opened

This originally replaced the CLI's 53-entry `KNOWN_COMPONENTS` table with a 196-entry one synced to core's doc targets. #4644 landed first and did it better: `loadKnownComponents()` derives the keys and visualProps from the `*.doc.mjs` theming targets at runtime, and returns `null` when docs are unresolvable so validation skips instead of deriving an empty registry and warning on every override. The table and the two tests that guarded it are dropped, and this is rebased onto that derivation.

Also here: the doc scan behind `every documented theming target is backed by a real themeProps literal` now walks recursively. It previously read only `<Dir>/<Dir>.doc.mjs`, so sub-component targets (`TopNav/TopNavHeading.doc.mjs`, `Avatar/AvatarStatusDot.doc.mjs`, ...) went unchecked.

## Not built

Validating bare state keys against doc `states`. A typo like `selectd` still emits a dead selector silently. `ThemingTarget.states` exists in the schema but only 35 of 164 targets declare it, so validating against it now would trade one set of false positives for another.

A second gap is discussed in the review thread: core's `themingTargets.test.ts` guard is per directory, so a rendered class whose doc target lives in a different directory is held by nothing. Measured drift is currently zero (218 targets, 218 rendered literals), but three such pairs exist.

## Tests

- butter and stone build with zero warnings. Reverting the validator fix reproduces `Unknown prop "selected"` on `top-nav-item`.
- `build-theme.registry.test.mjs`: 5 passed.
- 6 failures in the wider `packages/cli` run are pre-existing on main, verified by reverting this PR's three files to upstream and getting the identical 6: 4 in `build-theme.variants.test.mjs`, plus the known macOS case-insensitivity pair.

Refs #4110


packages/cli/api/theme/build/build.mjs
```diff
@@ -597,69 +597,228 @@ ${iconType}export declare const ${toIdentifier(themeDef.name)}Theme: DefinedThem
...
+ * guarded against the rendered `themeProps()`/`stableClassName()` literals,
+ * so these keys are exactly the classes components render. Both directions
+ * of drift — a key here without a doc target, and a doc target missing here
+ * — are enforced by cli/commands/build-theme.registry.test.mjs, which also
+ * compares each entry's props against the doc target's visualProps verbatim.
+ * When core adds or changes a target, update the component doc first, then
+ * copy the entry here.
  * @type {Record<string, string[]>}
  */
 const KNOWN_COMPONENTS = {
```
cixzhang: I guess this map will get stale quite quickly. @josephfarina thoughts on how we can make this better?

AKnassa: Good call, and it did. #4644 landed while this sat and replaced `KNOWN_COMPONENTS` with `loadKnownComponents()`, deriving keys and visualProps from the `*.doc.mjs` theming targets at runtime, returning `null` when docs are unresolvable so validation skips rather than warning on every override. That covers the case that pushed me to a literal table in the first place. I have rebased onto it and dropped the table along with the two tests that guarded it.

On whether the drift is closed now, I measured the current tree: 218 doc targets, 218 rendered `themeProps()`/`stableClassName()` literals, no orphans in either direction. Clean today. What holds it clean is `themingTargets.test.ts`, and that guard is per directory: it checks `<Dir>`'s rendered classes against `<Dir>/<Dir>.doc.mjs`, and skips when that file is absent. Three pairs fall through it, where the render site and the doc live in different directories:

- `Code/` renders `astryx-code`, documented in `CodeBlock/CodeBlock.doc.mjs`
- `Heading/` renders `astryx-heading`, documented in `Text/Text.doc.mjs`
- `hooks/` renders `astryx-input-status-icon`, documented in `Field/Field.doc.mjs`

All three are correct right now, though nothing holds them there. If one of those literals gets renamed, the guard stays green while the doc target goes stale, and since the CLI now reads the docs, the validator would warn "Unknown component" on the class the component really renders. That is the same shape as #4110, one level up.

One idea, if it seems worth it: make that guard whole tree rather than per directory, comparing rendered literals across `core/src` against `theming.targets[].className` across all `*.doc.mjs`, in both directions. This PR already does the doc to rendered half recursively, so the missing piece is rendered to doc. That would leave the derivation right by construction.

States stay unvalidated either way. `ThemingTarget.states` exists in the schema, but only 35 of 164 targets declare it, so checking against it today would trade one set of false positives for another. Probably worth revisiting once states are documented, validating only where `states` is present.

Happy to be talked out of the whole tree version if @josephfarina sees a cheaper cut.

cixzhang:
Thanks for this, and for the write-up in the thread — both things you found have since landed.

`loadKnownComponents()` now folds each target's declared `states` into the known keys ([#4778](https://github.com/facebook/astryx/pull/4778)), which is the "validate only where `states` is present" shape you proposed. `astryx theme build` on butter and stone is clean here — no `Unknown prop "selected"` on `top-nav-item`, just the unrelated font advisory. And the per-directory hole in the theming-target guard closed too: it now falls back to whichever doc file documents the classes a directory renders, so `Code`, `Heading` and `hooks` are covered ([#5016](https://github.com/facebook/astryx/pull/5016)).

Closing on that. Reopen any time if something still warns — merge main in first, the branch has aged. We're in [Discord](https://discord.com/invite/XnsUcFykEP) if you'd rather talk to a person.

[Reviewed by Robohands]

PR #4555, opened 2026-07-30
fix(core): TopNavMegaMenu hover-then-click no longer dismisses the panel (#3121)

### LINKED ISSUE #3121: [Bug] Minor UX suggestion for Top Nav Mega Menu trigger behavior
### Description

# [Bug] Minor UX suggestion for Top Nav Mega Menu trigger behavior

The top navigation [mega menu](https://xds-sandbox.vercel.app/components/TopNavMegaMenu) currently supports both **hover** and **click** as triggers.

While both interaction patterns are common individually, supporting both at the same time can lead to an unintuitive experience. For example, when testing the [Moonveil AI](https://moonveil.ai/) website with Astryx, my instinct was to **click** the **Services** navigation item. However, because the menu had already opened on hover, my click immediately closed it instead.

A couple of ideas that could make this interaction feel more natural:

1. **Delay enabling the click-to-close behavior.**  
   Add a short delay before the trigger element can close an already-open mega menu. The [Lovable](https://lovable.dev/home) homepage appears to follow a similar approach—if you hover over a menu item and immediately click it, the menu remains open instead of closing.

2. **Support an "open-only" trigger mode.**  
   Provide an option to configure the top-level navigation item as an **open-only** trigger. In this mode, clicking the trigger would only open the mega menu, while dismissal would happen only when the user hovers away (or via another configured dismiss behavior).

This would reduce accidental dismissals for users whose natural interaction is to click a navigation item after hovering over it.

### Reproduction

Lovable homepage:

Hover + immediate click -> mega menu stays open:

https://github.com/user-attachments/assets/65faac22-4744-44c6-a108-eff171374cba

Moonveil AI:

Hover + immediate click -> mega menu blinks

https://github.com/user-attachments/assets/17282cdb-45f0-4c9b-8fd0-06c88a11dbfc

### XDS Version

@astryxdesign/core@0.1.0

### Environment

_No response_

imdreamrunner:
Fixes #3121.

## Summary

The Top Nav mega menu flickered shut when you hovered a nav item and then clicked it—the click that naturally follows a hover dismissed the menu.

**Root cause:** the panel is a native `popover="auto"`, and the trigger button sits outside it. Without a native invoker relationship, the browser can light-dismiss the hover-opened panel during the trigger's pointer interaction, before React's click guard decides what the click means. jsdom does not implement this native event ordering, so unit tests alone cannot reproduce the original failure.

**Fix** (`packages/core/src/TopNav/TopNavMegaMenu.tsx`):

- Keep the panel as `popover="auto"` so native outside-click dismissal, Escape handling, focus restoration, and sibling-popover exclusivity remain intact.
- Register the trigger as the panel's native invoker with `popoverTarget`, which exempts the trigger interaction from auto-popover light-dismiss.
- Cancel the invoker's default toggle with `preventDefault()` so the component's interaction logic remains the single source of truth.
- Apply a Vercel-style 500 ms hover→click guard: a click shortly after a hover-open confirms and pins the panel instead of closing it.
- Click and keyboard opens are pinned and persist past mouse-leave; hover opens remain transient. Enter/Space opens and focuses the panel, while Escape closes and restores focus.

No public API change.

### Automated

- `packages/core/src/TopNav/TopNavMegaMenu.test.tsx`: **37 passing**, covering the 500 ms guard, pinned versus transient behavior, native-invoker wiring, default-toggle cancellation, native-dismiss state reconciliation, sibling-popover exclusivity, panel exit-hover behavior, and keyboard behavior.
- Full `packages/core/src/TopNav` suite: **103 passing**.
- `pnpm -F @astryxdesign/core typecheck` and ESLint: clean.

The unit tests intentionally lock in the required wiring, but jsdom cannot exercise native `popoverTarget`/light-dismiss ordering. The interaction below was therefore verified in a real Chromium browser against the Storybook `core-topnavmenu--mega-menu` story.

### Real browser

| Interaction | Result |
| --- | --- |
| hover → immediate click | **stays open** |
| second click | closes |
| hover → wait over 500 ms → click | closes |
| hover → move away | closes (transient) |
| click → move away | stays open (pinned) |
| click outside | closes |
| open one mega menu → open a sibling | first menu closes |
| Enter / Space → open, Escape → close | works; focus returns to trigger |

Browsers without the Popover API retain the component's existing degraded `useLayer` fallback; this PR does not add a legacy outside-dismiss polyfill.

### Before

Hover → immediate click dismisses the menu:

https://github.com/user-attachments/assets/0bfa93da-79fd-462b-bdb9-5ecbfb2d5669

### After

Hover → immediate click keeps the menu open:

https://github.com/user-attachments/assets/6b538cf7-5aa6-485f-8f7f-52e9a6f0b102

### DIFF
--- a/.changeset/topnav-mega-menu-click-guard.md
+++ b/.changeset/topnav-mega-menu-click-guard.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] TopNavMegaMenu: fix the hover-then-click flicker where clicking a nav item after hovering dismissed the mega menu. The trigger is registered as the native invoker for its `popover="auto"` panel and uses a Vercel-style hover→click guard, so the click that naturally follows a hover confirms and pins the panel open instead of toggling it shut. Native outside-click, Escape dismissal, and sibling-popover exclusivity are preserved. Click/keyboard opens are pinned (persist past mouse-leave); hover opens stay transient. Keyboard activation (Enter/Space) always opens and moves focus into the panel, while touch/click without a preceding hover toggles cleanly (#3121)
+@imdreamrunner
--- a/packages/core/src/TopNav/TopNavMegaMenu.doc.mjs
+++ b/packages/core/src/TopNav/TopNavMegaMenu.doc.mjs
@@ -7,9 +7,6 @@ export const docs = {
   subComponentOf: 'TopNav',
   displayName: 'Top Nav Mega Menu',
   description: 'Navigation item that displays a full-width mega menu panel on hover. Uses a slots API with items and featured props. TopNavMegaMenuItem renders itself in both desktop and mobile drawer modes. Supports inline collapsible drawer via render mode context.',
-  playground: {
-    wrapper: {component: 'TopNav'},
-  },
   props: [
     {
       name: 'label',
--- a/packages/core/src/TopNav/TopNavMegaMenu.doc.mjs
+++ b/packages/core/src/TopNav/TopNavMegaMenu.doc.mjs
@@ -7,6 +7,9 @@ export const docs = {
   subComponentOf: 'TopNav',
   displayName: 'Top Nav Mega Menu',
   description: 'Navigation item that displays a full-width mega menu panel on hover. Uses a slots API with items and featured props. TopNavMegaMenuItem renders itself in both desktop and mobile drawer modes. Supports inline collapsible drawer via render mode context.',
+  playground: {
+    wrapper: {component: 'TopNav'},
+  },
   props: [
     {
       name: 'label',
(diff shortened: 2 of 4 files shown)


imdreamrunner:
Just found #3851 (thanks Cindy for sharing that). will iterate on that PR since that's older

imdreamrunner:
#3851 hasn't been updated for a while so I am reopening this PR

cixzhang (approved):

cixzhang (changes requested):
Hmm trying in the deployment but seems like it doesn't open the mega menu at all, so let me take a closer look.

imdreamrunner:
@cixzhang  Thanks for reviewing!

Follow-up on the docsite issue:

**Why Storybook worked while the docsite did not:**

- Storybook renders `TopNavMegaMenu` inside its required `TopNav`.
- The docsite rendered it standalone.
- The closed panel also forced `display: flex`, creating an invisible layer that intercepted the trigger.

**Commit `b40fee9f` fixes this by:**

- Wrapping the docsite rendering in `TopNav`.
- Applying `display: flex` only under `:popover-open`.
- Testing that hover-then-click retains the existing pinned behavior.

We also discovered that this PR fixes the production issue tracked in #4785: hovering a closed panel’s invisible area could reopen it. The PR separates the panel and trigger hover handlers so only the trigger can schedule an open.

Verified in Chrome; 37 component tests and 303 docsite tests pass.

cixzhang (approved):

Issue #4116, opened 2026-07-20
RTL support for astryx components


nynexman4464:
# RTL support for astryx components

## Status

Design discussion. Nothing to review yet — this issue is to surface the topic and align on scope before any implementation.

## Problem

The i18n framework that landed in the #3641 series translates strings, but it doesn't touch layout direction. If a consumer sets `<InternationalizationProvider locale="ar">` today, astryx components will render Arabic text (once an `ar` catalog exists) in a left-to-right layout, which is broken UX for Arabic users.

Making astryx render correctly in RTL is a separate track from string translation. It divides into three layers, each with independent cost:

**Layer 1 — Direction detection API.** Expose the current direction as a value consumers and internal astryx code can read. Small, ~50 lines, one PR.

**Layer 2 — Component CSS + behavior.** The real work. Physical CSS properties (`margin-left`, `padding-right`, `left`, `right`) don't flip under `dir="rtl"`; logical properties (`margin-inline-start`, `inset-inline-*`, etc.) do. And some direction is expressed in JavaScript rather than CSS — icon choice, keyboard nav mapping, popover placement math.

**Layer 3 — Testing.** Storybook toolbar toggle + screenshots per direction in the vibe-test suite.

## Current state (audit)

I ran a coverage audit against `packages/core/src/*` (100 component directories, StyleX-only styling surface):

| Bucket | Count |
|---|---:|
| RTL-ready — logical props only, or explicit `[dir="rtl"]` handling, or direction-agnostic | 72 |
| Partially ready — mixes logical and physical props | 25 |
| Broken outright — physical-only, no RTL escape hatch | 3 |

Separately, ~44 components have **directional semantics** that a pure CSS audit doesn't cover: pagination prev/next chevrons, breadcrumb separators, tab keyboard nav (`ArrowLeft` / `ArrowRight`), toast slide-in direction, slider thumb position math. Overlay placement (Popover / DropdownMenu / Tooltip / HoverCard / MoreMenu / ContextMenu / NavMenu) shows up in this count too, but is largely handled already — they share a `useLayer` hook built on CSS anchor positioning with logical `self-inline-*` keywords, which flip automatically under `dir="rtl"`. Real gaps are in icons, keyboard nav, and slide/transform animation directions.

Only 3 components ship any explicit RTL work today:
- **MobileNav** — cleanest example. Drawer slide handled via `:is([dir="rtl"] *)` conditional in StyleX. Reference implementation.
[...]

PR #4564, opened 2026-07-30
fix(core): complete RTL physical→logical migration + enforce no-physical-properties (error)

nynexman4464:
## Summary

Completes the RTL physical→logical CSS migration for the core package and flips the guard rail on.

The last **22** physical `left`/`right` CSS properties across **9 components** (Avatar, Banner, Calendar, Chat composer, Chat composer drawer, Dialog, FieldStatus, Markdown, Popover) are converted to their CSS logical equivalents — `insetInlineStart`/`insetInlineEnd`, the diagonal-aware `borderStart*`/`borderEnd*` corner radii, and `textAlign: 'end'` — so these components mirror correctly under RTL. Each conversion is a strict 1:1 logical rename; separate corner radii are kept separate (no collapsing) even where the value is the same.

With the migration finished, **`@astryx/no-physical-properties` is promoted from `warn` to `error`** in both the `recommended` and `strict` tiers. The rule was deliberately shipped at `warn` pending this migration (the config comment said "flip to error afterward" — this is afterward). It now gates against any future physical-property regression across the package.

## Why logical properties

Physical `left`/`right` don't flip under RTL; the CSS logical equivalents (`inset-inline-*`, `*-start`/`*-end`) do. StyleX compiles these to native CSS logical properties, so the browser handles direction mapping — **LTR rendering is pixel-identical** (`inset-inline-end` === `right`, `border-start-start-radius` === `border-top-left-radius`, `text-align: end` === `right` in LTR).

## Testing

- `pnpm lint` (rule now at **error**): **0 errors** — the whole repo passes with the rule gating.
- `pnpm --filter @astryxdesign/core typecheck`: **0 errors**.
- Touched-component tests: **563/563 pass** (28 files).
- `pnpm check:repo`: green.
- Core build (StyleX compile): succeeds; generated CSS confirms native logical props — LTR pixel-identical.

Completes the RTL migration. Refs #3636, #4116.

## Follow-up: Avatar status-dot RTL fix (2nd commit)

The initial pass converted the Avatar status dot's anchor `right` → `insetInlineEnd` (right edge in LTR, left edge in RTL) but left the paired outward-push `transform: translate(50%, 50%)` physical — so under RTL the dot pulled *inward* toward the avatar center instead of hugging the bottom-leading corner. Fixed by making the X translate direction-aware (`:is([dir="rtl"] *)` → `translate(-50%, 50%)`); Y is direction-neutral. The dot now overhangs the bottom-inline-end corner in both directions (bottom-right LTR, bottom-left RTL).

Measured (Playwright, real StyleX-compiled CSS, `dir` inherited to the avatar subtree):

| | avatar centerX | dot centerX | corner |
|---|---|---|---|
| LTR | 80 | **125.3** (+45.3, right of center) | bottom-right ✅ |
| RTL | 920 | **874.7** (−45.3, left of center) | bottom-left ✅ |

Exact mirror about the avatar center; LTR is unchanged / pixel-identical.

_Lint gap noted: `@astryx/no-physical-properties` flags physical `left`/`right` keys but not physical `translate()` in transforms, so this bug class (logical anchor + unflipped physical transform) isn't lint-catchable today — candidate for a future rule enhancement._


nynexman4464:
Superseded by the 3-PR stack #4613 → #4614 → #4615 (byte-identical code, split for reviewability). Closing to avoid a double-land.

PR #4570, opened 2026-07-31
fix(treelist): leaf row highlight flush with sibling parents (trial)

freddymeta:
## What

A **trial** to see how "flush to guide" reads. This makes a leaf row's hover / selected / focus highlight **start at the same x as its sibling parent rows** (the guide/toggle column) instead of being inset past them by the chevron column. No new prop — it fixes the default.

## The inconsistency today

The per-level indent lives on the row's outer `margin-inline-start`. Parents use `level * step`, but **leaves add the chevron-column offset (`spacing-4` + `spacing-2`) into that same margin** so their labels line up under sibling parents. Side effect: a leaf's *entire row box* — and therefore its highlight — begins one toggle-column further in than a parent at the same level. Within a level, leaf highlights look inset relative to parent highlights.

## The change

Move the leaf's chevron-column offset from the row **margin** to the content **`padding-inline-start`**:

- Every row box now starts at `level * step` (parents and leaves alike) → highlights are flush to the guide column.
- A leaf re-insets only its **content** via padding, so labels stay exactly where they were.

Label x is unchanged (verified algebraically: old `margin + basePad` sum == new `padding` sum); only the highlighted row box moves. `padding-inline-start` is a longhand, so it wins over `contentWrapper`'s `paddingInline` shorthand via StyleX's longhand>shorthand priority — order-independent, no `!important`.

## Notes / open question

- **Trial PR** — sharing to eyeball whether flush-to-guide is the right look before committing to it. I did **not** add a variant (per the ask); it changes the default leaf alignment of the highlight box.
- This is the more targeted alternative to the full-bleed #4540 (which stretches edge-to-edge). They're mutually informative — pick whichever reads better; I can close the other.
- Storybook preview on this PR shows it live (TreeList stories with mixed parent/leaf rows).

## Testing

- `TreeList` suite: 68 passed (incl. 2 new — a leaf and a parent at the same level publish the same row indent; the chevron offset no longer lands in the leaf row box).
- Verified emitted CSS: leaf content padding = `calc(var(--spacing-2) + var(--spacing-4) + var(--spacing-2))`; row margin stays `var(--_tree-indent)`.

### DIFF
--- a/.changeset/treelist-leaf-highlight-flush.md
+++ b/.changeset/treelist-leaf-highlight-flush.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] TreeList: a leaf row's hover/selected/focus highlight now starts flush with its sibling parent rows (at the guide/toggle column) instead of inset by the chevron column. The leaf's chevron-column offset moves from the row's outer margin to inner content padding, so labels stay aligned with parents while the highlighted row box lines up across leaves and parents. Purely visual; no API change.
+@freddymeta
--- a/packages/core/src/TreeList/TreeList.test.tsx
+++ b/packages/core/src/TreeList/TreeList.test.tsx
@@ -545,6 +545,48 @@ describe('TreeList', () => {
   });
 
   // ===========================================================================
+  // Leaf highlight flush-to-guide
+  // ===========================================================================
+
+  describe('leaf row highlight is flush with sibling parents', () => {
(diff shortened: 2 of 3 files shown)


freddymeta:
Dropping this in favor of #4540 — folding the leaf-flush alignment into the structural full-width row-box rework there (per review direction on #4540), so the whole thing lands as one coherent change rather than two overlapping ones.

PR #4572, opened 2026-07-31
fix(dropdown-menu): prevent double highlight when hovering a submenu trigger

freddymeta:
## Summary

Fixes a double-highlight bug in `DropdownMenu` submenus. When a submenu trigger row is hovered while another item in the same menu still holds focus, two items showed the highlight background at once.

## The bug

Menus keep a single highlight by using DOM focus as the sole highlight source — `focusMenuItemOnHover` moves focus onto the pointed-at item so the one focus-driven highlight follows the pointer. Regular items (`DropdownMenuItem`, `DropdownMenuCheckboxItem`, `DropdownMenuRadioItem`) already follow this pattern: their style has only a `:focus` background, and they wire `onPointerMove` → `focusMenuItemOnHover`.

The submenu trigger row in `DropdownMenuSubMenu` was left on the older pattern:

- Its `triggerStyles.root` had both a `:focus` background **and** a separate `:hover` background block.
- It opens the flyout on hover intent (via `useMenuHover`) without moving focus, and it did not wire `focusMenuItemOnHover` on pointer move.

So when focus was on a sibling item (e.g. "Reload") and the pointer moved onto the submenu trigger (e.g. "More Tools"), the trigger lit up via its `:hover` background while the sibling kept its `:focus` highlight — two simultaneous highlights.

## The fix

Mirror the single-highlight pattern the regular items already use:

- **Remove the trigger's `:hover` background block** from `triggerStyles.root`, leaving only the `:focus` background (matching `DropdownMenuItem`).
- **Wire `onPointerMove` → `focusMenuItemOnHover`** on the trigger so a mouse hover moves the single focus-driven highlight onto the trigger, dropping the sibling's focus.
- Add `DropdownMenuSubMenu.tsx` to the `SYNC:` consumer list in `menuItemHover.ts`.

`focusMenuItemOnHover` only reacts to a real mouse pointer, skips disabled items, and is a no-op when the element is already focused, so it doesn't thrash focus.

## What's preserved

- The **open-branch indicator** (`triggerStyles.open`, applied while the flyout is open) is intentional and unchanged — it keeps the open branch visibly active even after focus moves into the child flyout.
- **Hover-open behavior is unchanged.** The pointer-move focus and the hover-open intent (`onMouseEnter`/`onMouseLeave`) are separate concerns and coexist; the flyout still opens on the hover-intent delay. The existing "flyout stays open while the pointer is over it" and two-level nested-submenu tests still pass.

Non-breaking: behavioral polish plus a style removal, no API change.

## Testing

- Added a regression test asserting that hovering the submenu trigger while a sibling holds focus moves focus onto the trigger (single highlight): the trigger is focused and the previously-focused sibling is not.
- Verified the test fails without the `onPointerMove` fix and passes with it.
- Full `DropdownMenu`/`DropdownMenuSubMenu`/`DropdownMenuSelectable` suites pass.

### DIFF
--- a/.changeset/dropdown-submenu-double-highlight.md
+++ b/.changeset/dropdown-submenu-double-highlight.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] DropdownMenu: a submenu trigger no longer shows a second highlight when hovered while another item still holds focus — hover now moves the single focus-driven highlight onto the trigger, matching regular menu items
+@freddymeta
--- a/packages/core/src/DropdownMenu/DropdownMenuSubMenu.test.tsx
+++ b/packages/core/src/DropdownMenu/DropdownMenuSubMenu.test.tsx
@@ -8,7 +8,7 @@
 
 import {describe, it, expect, vi, beforeEach} from 'vitest';
 import {useState} from 'react';
-import {render, screen, waitFor} from '@testing-library/react';
+import {render, screen, waitFor, fireEvent} from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import {DropdownMenu} from './DropdownMenu';
 import {DropdownMenuItem} from './DropdownMenuItem';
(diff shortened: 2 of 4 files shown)


DickShern (approved):
LGTM

cixzhang (approved):

PR #4574, opened 2026-07-31
fix(field-label): forward description clicks to checkbox/switch without name pollution

freddymeta:
## What

Clicking a **CheckboxInput** or **Switch** description now toggles the control. Today the description renders as a sibling `<span>` after the `</label>`, so native label-click activation never covers it — only the label line toggles.

This is the review-adjusted take on #4152, incorporating @cixzhang's feedback there.

## Why not just nest the description in the `<label>`

That was #4152's approach, and the blocking concern was correct: for a single control, the `<label htmlFor>` supplies the control's **accessible name**, so nesting the description inside it makes the name become *"Label + full description"* — and since the description is *also* wired via `aria-describedby`, screen readers announce it **twice** (name + description).

## Approach (cixzhang's option 2)

Keep the description a **sibling `<span>`** — so it's never folded into the accessible name — and **forward its click** to the control by id. The whole label area becomes a hit target with **no accessible-name change**.

- New `FieldLabel` prop **`hasClickableDescription`** (default `false`), set by `CheckboxInput` and `Switch`.
- **Off for `Field`-backed text inputs** (TextInput, Selector, …): clicking a text input's label only focuses it — no toggle benefit — so there's no reason to risk the name. **Ignored for group labels** too.
- Forwarding is a **mouse-only convenience**: the control remains the keyboard path, the description isn't focusable, so there's no new tab stop and no a11y-tree change. Guarded (`e.target === e.currentTarget`, `defaultPrevented`) so a click on an interactive element inside a `ReactNode` description isn't hijacked.

## Tests (both types cixzhang asked for)

- **Click-to-toggle**: clicking the description toggles the control — `CheckboxInput` + `Switch`.
- **No name pollution** (the assertion the original PR was missing): `toHaveAccessibleName` excludes the description while `toHaveAccessibleDescription` includes it — `CheckboxInput` + `Switch`.
- **Scope**: `FieldLabel` forwards only when opted in, **not** for the default text-input path, **not** for group labels, and keeps the description a sibling of the `<label>`.

## Testing

- CheckboxInput (37) + Switch (48) + FieldLabel (16) = 101 passed; Field/TextInput/RadioList (188) unaffected.

Note: scoped to CheckboxInput + Switch (the controls sharing `FieldLabel`). RadioListItem uses a different (`Item`-based) structure — a follow-up if wanted.

Supersedes the approach in #4152.

### DIFF
--- a/packages/core/src/Field/FieldLabel.tsx
+++ b/packages/core/src/Field/FieldLabel.tsx
@@ -76,6 +76,11 @@ const styles = stylex.create({
     fontWeight: fontWeightVars['--font-weight-normal'],
     color: colorVars['--color-text-secondary'],
   },
+  // When the description forwards clicks to a click-activatable control
+  // (checkbox/switch), it reads as part of the same hit target as the label.
+  descriptionClickable: {
+    cursor: 'pointer',
+  },
 });
 
 export interface FieldLabelProps extends BaseProps<HTMLLabelElement> {
@@ -147,6 +152,21 @@ export interface FieldLabelProps extends BaseProps<HTMLLabelElement> {
    * ID for the description element (for aria-describedby on the input).
    */
   descriptionID?: string;
+  /**
+   * When true, clicking the description toggles/activates the associated
+   * control — for click-activatable single controls (checkbox, switch) whose
+   * whole label area should be a hit target. The description stays a sibling
+   * `<span>` (NOT nested in the `<label>`), so it is never folded into the
+   * control's accessible name; it forwards clicks to the input by id instead.
+   *
+   * Leave `false` for text inputs (TextInput, Selector, etc.): clicking a text
+   * input's label only focuses it — there's no toggle benefit — and nesting or
+   * forwarding the description would only risk polluting the accessible name.
+   * Also ignored for group labels (`isGroupLabel`), which name a group, not a
+   * single control.
+   * @default false
+   */
+  hasClickableDescription?: boolean;
 }
 
 /**
@@ -175,6 +195,7 @@ export function FieldLabel({
   labelTooltip,
   description,
   descriptionID,
+  hasClickableDescription = false,
   className,
   style,
   xstyle,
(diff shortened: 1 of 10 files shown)


packages/core/src/Field/FieldLabel.tsx
```diff
@@ -147,6 +152,21 @@ export interface FieldLabelProps extends BaseProps<HTMLLabelElement> {
...
+   * control's accessible name; it forwards clicks to the input by id instead.
+   *
+   * Leave `false` for text inputs (TextInput, Selector, etc.): clicking a text
+   * input's label only focuses it — there's no toggle benefit — and nesting or
+   * forwarding the description would only risk polluting the accessible name.
+   * Also ignored for group labels (`isGroupLabel`), which name a group, not a
+   * single control.
+   * @default false
+   */
+  hasClickableDescription?: boolean;
```
cixzhang: Let's skip adding this. It should be fine to allow descriptions to forward focus to the input.

packages/core/src/Field/FieldLabel.tsx
```diff
@@ -194,6 +215,33 @@ export function FieldLabel({
...
+  const forwardsDescriptionClick =
+    hasClickableDescription && !isGroupLabel && inputID != null;
+
+  // Mouse-only convenience: clicking the description clicks the associated
+  // control, mirroring native label-click. Keyboard users already toggle via
+  // the control itself (the description is supplementary, not focusable), and
+  // the label text keeps native `htmlFor` activation — so this adds no new tab
+  // stop and no a11y-tree change. Guarded so a click that lands on an
+  // interactive element inside a ReactNode description isn't hijacked.
+  const handleDescriptionClick = (e: React.MouseEvent<HTMLSpanElement>) => {
```
cixzhang: We should be able to leverage useInputContainer instead to forward the click to focus with should also skip potential nested interactive elements so won't interfere with links or other interactive content in the description. It also has a text selection guard.

cixzhang (changes requested):
Thanks this seems much better! I'm thinking we may not need an extra prop and just always have the description area to forward clicks to focus the input. You should be able to use useInputContainer to achieve this and it will also skip activating if there's nested interactive content in the description or label like a link that gets clicked.

cixzhang (approved):
The description-click forwarding via ownerDocument.getElementById is the right call — it resolves against the description's owning document (portal/iframe-safe) rather than global document, and the ref-shaped getter is just lazy resolution since FieldLabel points at the control by id. Sibling-span + aria-describedby avoids accessible-name pollution, and both click-to-toggle and no-name-pollution tests are present. Addresses my earlier feedback. Approving.

PR #4578, opened 2026-07-31
docs(docsite): add useContainerReveal example

cixzhang:
## What

Adds a docsite Hooks guide page documenting `useContainerReveal`, the headless hover/focus reveal primitive that shipped in #4573. The page lives at `packages/cli/docs/hooks.doc.mjs`, so it surfaces both on the docsite (`/docs/hooks`, under Guides) and via the CLI (`astryx docs hooks`) — the same mechanism the other reference guides use.

The example shows the canonical usage: destructure `getContainerProps` / `getContentRevealProps`, spread `getContainerProps()` on the container and `getContentRevealProps()` on the content to reveal. It builds a reveal-on-hover row of action buttons on an `Item`, mirroring how `Thumbnail` consumes the hook internally for its remove button. The page also covers when to reach for it (secondary row/overlay actions) and the accessibility payoff — revealed content stays mounted and in the tab order, reveals on `:focus-within`, and stays visible on touch — plus the `isEnabled`, `isRevealInverted`, and `isLayoutPreserved` options.

## Why

The hook shipped recently but had no docsite entry, so consumers had no worked example of the prop-getter pattern. Hooks weren't individually documented before; this follows the existing `styling.doc.mjs` reference-guide pattern (static content blocks: prose / code / list / table) rather than introducing a new docs mechanism, and seeds a "Hooks" guide that more hooks can be added to over time.

## Validation

- `pnpm --filter @astryxdesign/cli typecheck:strict` — the doc file (covered by that project's `docs/**/*.mjs` glob) reports no errors. Remaining errors in the run are pre-existing unbuilt-sibling-package resolution noise unrelated to this change.
- Loaded the module directly to confirm the `ReferenceDoc` shape the docsite generator expects (title, category, sections with valid content-block types).
- API cross-checked against `packages/core/src/hooks/useContainerReveal.ts` and the real usage in `packages/core/src/Thumbnail/Thumbnail.tsx`.

## Example block

Adds a live example block, `useContainerRevealHookUsage`, under `packages/cli/templates/blocks/components/Hooks/` (a `.tsx` + `.doc.mjs` pair, `exampleFor: 'useContainerReveal'`). This is the mechanism the other hook examples use (`useKeyboardHint`, `useStreamingText`) — it renders in the docsite Examples section for the hook rather than living only as a prose snippet. It builds reveal-on-hover edit/delete actions on an `Item`, spreading `getContainerProps()` on the row and `getContentRevealProps()` on the actions, with the accessibility note (revealed on focus, always visible on touch) inline.

Docs-only; no changeset (matching prior docs-only PRs to `packages/cli/docs`, e.g. #4294 and #4243).

### DIFF
--- a/packages/cli/assets/templates/blocks/components/Hooks/useContainerRevealHookUsage.doc.mjs
+++ b/packages/cli/assets/templates/blocks/components/Hooks/useContainerRevealHookUsage.doc.mjs
@@ -1,0 +1,14 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+/** @type {import('@astryxdesign/cli/authoring').TemplateDoc} */
+export const doc = {
+  type: 'block',
+  exampleFor: 'useContainerReveal',
+  name: 'useContainerReveal — Reveal-on-hover Row Actions',
+  displayName: 'useContainerReveal — Reveal-on-hover Row Actions',
+  description:
+    'File rows keep their edit/delete actions hidden at rest and reveal them on hover or keyboard focus via useContainerReveal — the actions stay mounted and in the tab order.',
+  isReady: true,
+  aspectRatio: 16 / 9,
+  componentsUsed: ['Item', 'Button', 'Icon', 'Card', 'Text', 'Layout'],
+};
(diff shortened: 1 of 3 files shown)


cixzhang (commented):
Add docs for the new hook following other hook doc conventions (like useClickableContainer) don't add a global hook doc.

PR #4582, opened 2026-07-31
feat(timestamp): add an opt-in copyable multi-timezone hover card

freddymeta:
## Summary

`Timestamp`'s hover surface is now a single **copyable hover card** for every timestamp that shows one. A relative timestamp (or any timestamp configured with `tooltipEntries`) reveals a card on hover/focus where **each row is copy-to-clipboard** — the full absolute time by default, or one row per configured zone/format (local / UTC / another zone / ISO). This replaces the old read-only `Tooltip`.

```tsx
<Timestamp
  value={deployedAt}
  format="relative"
  tooltipEntries={[
    {label: 'Your time'},
    {timezoneID: 'UTC', label: 'UTC'},
    {timezoneID: 'UTC', format: 'system_date_time', label: 'ISO (UTC)'},
  ]}
/>
```

No new on/off prop: the card is gated by the existing `hasTooltip` (default `true`) together with the existing "relative format, or `tooltipEntries` set" condition. `hasTooltip={false}` still suppresses it.

## Composition

- **The card is a `MetadataList`.** Each row is a `MetadataListItem` whose value pairs the labelled instant with its copy button — so the card inherits MetadataList's `<dl>`/`<dt>`/`<dd>` semantics, label/value colors, and spacing rather than hand-rolling a grid. The default single row is a one-row list carrying the full absolute time (empty label), itself copyable.
- **Lazy-loaded.** `HoverCard` + `MetadataList` + the copy `Icon`/`IconButton` live in a `TimestampHoverCard` module rendered behind `React.lazy` + `Suspense`, only on the card-attached branch. The `Suspense` fallback is the bare `<time>`, so nothing disappears while the chunk loads — the default, card-less `Timestamp` never bundles any of it.
- **Shared formatting.** Rows come from the existing `formatTooltipLines(date, entries)` helper, so the visible text and the card can't drift.
- **Copy affordance.** Per-row Astryx `IconButton` (`ghost`, `sm`): `navigator.clipboard.writeText`, an icon that flips `copy` → `check` for ~1.5s (timer cleared on unmount, restarted on re-copy), a polite `useAnnounce` live-region confirmation, an accessible `aria-label` that swaps copy/copied, and a silent no-op when the clipboard rejects.

## A11y

- Opens on hover **and** keyboard focus (`focusTrigger="always"`); the `<time>` gets a tab stop only while a card is attached, with a dashed-underline affordance (`hasHoverIndication`) and an accessible name (`label`).
- The list is a semantic `<dl>`; every row emits its label cell (empty for the default line) so the grid stays aligned and the markup stays valid.

## Theming

- The card is a `MetadataList`, so its rows/labels/values are themeable through the standard `metadata-list` / `metadata-list-item` targets — no bespoke row/label/value targets.
- The per-row copy affordance keeps a scoped `astryx-timestamp-copy-button` target so a theme can restyle just that button.

## Non-breaking

The card renders the same rows and copy behavior as before; this change is an internal recomposition onto `MetadataList` with no API change. Relative/`tooltipEntries` timestamps show the copyable card (a behavior + visual change from the former read-only tooltip); `hasTooltip={false}` still suppresses it.

## i18n

Catalog keys in `packages/core/locales/en.json`: `@astryx.timestamp.copyValue` ("Copy {value}"), `@astryx.timestamp.copied` ("Copied"), `@astryx.timestamp.detailsLabel` ("Timestamp details"), resolved via `useTranslator()`.

## Testing

- `Timestamp` suite (64 tests), `MetadataList` suite, and the `themingTargets` suite pass; `typecheck` and eslint on the changed files are clean.
- Verified in Storybook: the default single-row card (empty label, flush copy button) and multi-zone cards (labelled rows) both render correctly as a `MetadataList`.

### DIFF
--- a/packages/core/src/Timestamp/Timestamp.tsx
+++ b/packages/core/src/Timestamp/Timestamp.tsx
@@ -247,6 +247,10 @@ const styles = stylex.create({
     marginInline: 0,
   },
   cardLabel: {
+    // The card renders on --color-background-surface (not the inverted
+    // tooltip palette), so set explicit text colors instead of inheriting an
+    // ambient one that fails contrast against the surface.
+    color: colorVars['--color-text-secondary'],
     marginBlock: 0,
     marginInline: 0,
     paddingInlineEnd: {
@@ -255,6 +259,7 @@ const styles = stylex.create({
     },
   },
   cardValue: {
+    color: colorVars['--color-text-primary'],
     marginBlock: 0,
     marginInline: 0,
     whiteSpace: 'nowrap',
@@ -272,7 +277,7 @@ const styles = stylex.create({
         ':hover': colorVars['--color-overlay-hover'],
       },
     },
-    color: 'inherit',
+    color: colorVars['--color-text-primary'],
     cursor: 'pointer',
     lineHeight: 0,
     outline: {
--- a/packages/core/src/Timestamp/Timestamp.tsx
+++ b/packages/core/src/Timestamp/Timestamp.tsx
@@ -16,14 +16,7 @@
  * - /packages/cli/assets/templates/blocks/components/Timestamp/ (showcase blocks)
  */
 
-import {
-  useCallback,
-  useEffect,
-  useRef,
-  useState,
-  lazy,
-  Suspense,
-} from 'react';
+import {useCallback, useEffect, useRef, useState} from 'react';
 import * as stylex from '@stylexjs/stylex';
 import {Text} from '../Text';
 import type {TextType, TextSize, TextColor, TextWeight} from '../theme/types';
@@ -42,10 +35,6 @@ import type {
   TimestampTooltipEntry,
   TimestampTooltipLine,
 } from './tooltipEntries';
-
-const LazyXDSTooltip = lazy(async () =>
-  import('../Tooltip/Tooltip').then(mod => ({default: mod.Tooltip})),
-);
 
 // =============================================================================
 // Types
@@ -89,7 +78,9 @@ export interface TimestampProps extends BaseProps<HTMLTimeElement> {
    */
   autoThreshold?: number;
   /**
-   * Whether to show a tooltip with the full date/time on hover.
+   * Whether to show a hover card with the full date/time on hover. The card
+   * is copyable — its default single row carries the full absolute time — and
+   * `tooltipEntries` customizes its rows.
    * @default true
    */
   hasTooltip?: boolean;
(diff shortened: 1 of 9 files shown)


packages/core/src/Timestamp/Timestamp.tsx
```diff
@@ -154,42 +165,60 @@ const styles = stylex.create({
...
   },
-  tooltipValue: {
+  cardValue: {
+    color: colorVars['--color-text-primary'],
     marginBlock: 0,
-    // <dd> carries a 40px inline start margin from the UA stylesheet.
     marginInline: 0,
+    whiteSpace: 'nowrap',
+  },
+  copyButton: {
```
cixzhang: Use a Button instead of remaking one

cixzhang (changes requested):
Can we keep the lazy-loading that the old `Tooltip` path had? Right now `HoverCard` is imported statically at the top of `Timestamp.tsx`, so every `Timestamp` bundles the overlay machinery even when no card is ever shown — the old code deliberately code-split this via `React.lazy(() => import('../Tooltip/Tooltip'))` + `Suspense`.

Since `HoverCard` only renders inside the card-attached branch, the same pattern restores the "no card → no overlay code in the bundle" property:

```tsx
const LazyHoverCard = lazy(async () =>
  import('../HoverCard').then(mod => ({default: mod.HoverCard})),
);
```

and wrap the `<HoverCard>…</HoverCard>` render in `<Suspense>` (fallback can be the bare `<time>` anchor / null, matching how it degraded before). One note: the `Icon` used in `CopyableEntryRow` would need to ride the same split to fully benefit — otherwise it stays eager via its static import.

freddymeta:
Updated since the last review — re-requesting a look:

- **Lazy-loading restored** (your ask): `HoverCard` + the card content + copy `Icon`/`IconButton` now live in a `React.lazy` + `Suspense` `TimestampHoverCard` module, rendered only on the card-attached branch; the `Suspense` fallback is the bare `<time>`. The default card-less `Timestamp` no longer bundles any of it.
- **Card is now composed on `MetadataList`** instead of a hand-rolled `<dl>` grid. Each row is a `MetadataListItem` whose value pairs the labelled instant with its copy button. This drops the bespoke `cardRows`/`cardLabel`/`cardValue` styles and makes the rows/labels/values themeable through the standard `metadata-list` / `metadata-list-item` targets; the scoped `timestamp-copy-button` target is kept for the copy affordance.
- **Copy buttons are right-aligned** in each row.

The PR description is updated to match. Tests green locally (Timestamp 64, MetadataList, themingTargets 275); typecheck + eslint clean.

freddymeta:
@cixzhang — re-requesting review. Since your last pass, this addresses the feedback and is rebased clean onto `main` (0 behind, CI green):

- **Lazy-loaded** the hover card — the overlay code (HoverCard + IconButton + Icon) now lives in a separate `'use client'` module, so a default Timestamp bundles none of it.
- Copy control is now an **`IconButton variant="ghost"`** (was a bespoke button).
- Dropped the `hasCopyableEntries` prop — the copyable hover card is now driven entirely off **`tooltipEntries`**, so there's no redundant on/off axis (composed on MetadataList).
- Unified the hover surface on the single copyable HoverCard (dropped the separate read-only tooltip).

Tests pass; happy to iterate further.

_— on behalf of Freddy_

cixzhang (approved):

freddymeta:
Heads-up @cixzhang — I've added a **material feature since your approval**, so I've moved this back to **draft** for a fresh look rather than letting it land under the earlier sign-off.

**What changed (commit `1ded891`):**

**1. Per-entry `isCopyable` (default `false`).** Copy is now opt-in per row, so one card can mix read-only human-readable rows with a copyable machine value — e.g. show Local + UTC for reading, but only let readers grab the `system_date_time` value:

```tsx
tooltipEntries={[
  { label: 'Local' },                                              // read-only
  { label: 'UTC', timezoneID: 'UTC' },                             // read-only
  { label: 'ISO', timezoneID: 'UTC', format: 'system_date_time', isCopyable: true }, // copyable
]}
```

The default single-row relative card stays copyable, so the common case is unchanged.

**2. Dedicated action column for the copy button.** Copy buttons render in their own trailing grid column so they align down one column regardless of value width, instead of floating inside each value via `space-between`. The action column is **only reserved when some row is copyable** — a fully read-only card has no trailing gutter (so the right padding stays symmetric).

**3. Left off MetadataList for the card layout.** The 3-column `label / value / action` grid isn't expressible on MetadataList's 2-column `<dt>/<dd>`, so the card is now a small local `<dl>` grid. Labels use the `supporting` text role (the quieter secondary register that is Timestamp's own default), values the `body` role.

**4. No-label padding fix.** A card with no labels (the default single-instant card) drops the label column entirely so the value sits flush at the leading edge instead of being indented by a reserved-but-empty label column.

CI: 112/112 Timestamp tests pass, lint + typecheck clean.

No new top-level props — everything rides on the existing `tooltipEntries` object.

PR #4583, opened 2026-07-31
fix(examples): use an existing background token in the layer demo

Le-Ngoc-Tu:
The layer-demo block in the Vite and Next.js source examples referenced `--color-background-secondary`, which is not defined anywhere in the design system -- `colorDefaults` in packages/core/src/theme/tokens.stylex.ts has no such key. An undefined CSS custom property resolves to nothing, so the block rendered with no background at all instead of the surface color the demo intends to show.

Point both at `--color-background-surface`, which exists and gives the demo block the raised surface it was meant to have (the nested card keeps `--color-background-body`, so the two stay visually distinct).

### DIFF
--- a/apps/example-nextjs-source/src/app/page.tsx
+++ b/apps/example-nextjs-source/src/app/page.tsx
@@ -28,7 +28,7 @@ const styles = stylex.create({
   layerDemo: {
     padding: '1.5rem',
     borderRadius: 12,
-    backgroundColor: 'var(--color-background-secondary)',
+    backgroundColor: 'var(--color-background-surface)',
   },
   customButton: {
     borderRadius: 999,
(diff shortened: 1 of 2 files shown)


imdreamrunner (approved):

cixzhang (approved):
Verified — `--color-background-secondary` isn't in `colorDefaults`, so that block renders with no background at all today. Would merge as is; the CLA signature is the only thing left.

[Reviewed by Robohands]

cixzhang:
Closing this one for the CLA, not for the change — the fix itself was reviewed and approved.

The Meta CLA was requested on 31 July and is still unsigned, and we cannot merge any pull request without it. Rather than leave an approved PR sitting indefinitely, I am closing it.

@Le-Ngoc-Tu this is easy to undo: sign the CLA at https://code.facebook.com/cla/ and reopen, or open a fresh PR with the same commit — the approval stands and we will get it in. Thanks for the contribution, and sorry it stalled on paperwork.

PR #4585, opened 2026-07-31
feat(sandbox): add ChatLayout playground with state toggles

cixzhang:
Adds a ChatLayout playground page to the sandbox app so we can manually eyeball ChatLayout across its key states without wiring up a scratch page each time.

The page lives under the sandbox **Tools** category and uses a segmented control to switch between four states inside a fixed-height (600px) host container, so scroll and dock behavior is directly observable:

- **Empty** — no messages; validates the empty ChatLayout (uses the `emptyState` prop) renders without a phantom scrollbar.
- **Short** — a couple of messages; this is the #4202 case. With the flex-column root fix on `main`, short content must not produce a phantom vertical scrollbar and the composer dock sits at the bottom with no extra scroll height.
- **Long** — 60+ messages so the list overflows and self-scrolls; the dock stays pinned to the scrollport bottom.
- **Streaming** — an assistant message streams in progressively to exercise the streaming / auto-scroll state.

The composition mirrors `apps/storybook/stories/ChatLayout.stories.tsx` — it wires `ChatLayout` (with `composer` and `emptyState`), `ChatMessageList`, `ChatMessage` / `ChatMessageBubble` / `ChatSystemMessage`, and `ChatComposer` / `ChatComposerInput` from the real public API, rather than hand-rolling any chat internals. Styling is StyleX with semantic tokens (the only fixed dimension is the intentional test-host height).

Primary purpose is manual validation: it's how ChatLayout states can be visually checked, including the recently-merged #4202 phantom-scrollbar fix and future chat states.

**Where it shows up:** sandbox → Tools → "ChatLayout Playground" (route `/pages/chat-playground/`).

No changeset — `@astryxdesign/sandbox` is a private app in the changeset ignore list, consistent with the other `apps/*`.

### DIFF
--- a/apps/sandbox/src/app/(sandbox)/pages/chat-playground/page.tsx
+++ b/apps/sandbox/src/app/(sandbox)/pages/chat-playground/page.tsx
@@ -1,0 +1,326 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+'use client';
+
+/**
+ * @file page.tsx
+ * @input None
+ * @output ChatLayout Playground — a tool to eyeball ChatLayout across its
+ *   states (empty, short, long, streaming) inside a fixed-height host so the
+ *   scroll/dock behavior is observable. Mirrors the composition used in
+ *   apps/storybook/stories/ChatLayout.stories.tsx.
+ * @position Sandbox tool page (/pages/chat-playground)
+ *
+ * The primary purpose is manual validation of ChatLayout — e.g. confirming the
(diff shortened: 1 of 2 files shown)


cixzhang:
Closing this. Sandbox pages rot quietly, and the cases this covers now have better homes: #4202 got a real regression test in July, and empty and streaming are covered by the `WithEmptyState` and `ChatAutoScroll` stories. A page we have to keep in sync buys less than those do.

Issue #4142, opened 2026-07-21
Table tree: fold useTableRowExpansion affordances into useTableTreeData, then converge


humbertovirtudes:
## Problem

We now have two overlapping tree/hierarchy plugins for `Table`:

- `useTableRowExpansion` (#3441, merged) — inherited-columns tree: consumer passes flat data + `getDepth`, chevron on the first content column, a header expand-all toggle, and whole-row-click expansion.
- `useTableTreeData` + `useTableTreeState` (#3789) — nested data, in-place cell decoration, external-store per-row subscriptions, imperative row ARIA, cycle guard, flat-data no-op, controlled + uncontrolled modes, token-based indent, `sortSiblings`.

`useTableRowExpansion` has no remaining detail-panel / arbitrary-content surface — that path was stripped earlier, so it is purely hierarchy. The two plugins are therefore genuinely redundant, not two different surfaces that happen to overlap.

## Why converge on the tree plugin (not rowExpansion)

`useTableTreeData` is the stronger base to keep:

1. Fine-grained re-render — a toggle re-renders only the affected cells via `useSyncExternalStore`, not the whole body (`useTableRowExpansion` re-renders broadly).
2. Cycle guard — an ancestor-chain `Set` in the flatten walk skips cyclic edges. This is exactly the fix #3971 wants for `useTableRowExpansion`; converging resolves that bug class for free.
3. Row ARIA — `aria-level` / `aria-expanded` applied imperatively per `<tr>`, self-cleaning on disconnect.
4. Uncontrolled + controlled modes, token indent (`'sm' | 'md' | 'lg'`), and a flat-data no-op so adopting it ahead of hierarchical data costs nothing.

The Q3 Table tracker (#3639) is explicitly trying to avoid API sprawl, so we land one tree plugin and retire the other.

## What to fold from `useTableRowExpansion` into `useTableTreeData`

Two affordances `useTableRowExpansion` has that the tree plugin does not yet:

1. **Header expand-all toggle.** `useTableRowExpansion` renders a header toggle button driven by `isAllExpanded: boolean | 'indeterminate'` + `onToggleExpandAll`. `useTableTreeState` already computes the state (`expandAll` / `collapseAll`) but there is no built-in header UI. Port the header affordance, wired to the tree state's all / none / indeterminate.
[...]

PR #4587, opened 2026-07-31
docs(core): soft-deprecate useTableRowExpansion for the tree plugin (#4142)

humbertovirtudes:
## Summary

Fourth step of the tree-plugin convergence tracked in #4142: soft-deprecate `useTableRowExpansion` and `useTableRowExpansionState` in favor of the tree plugin.

With expand-all (#4172), whole-row click (#4548), and the migration guide (#4584) all landed, `useTableTreeData` + `useTableTreeState` now cover the older plugin's affordances, and do so with a cycle guard, per-row fine-grained re-render, and imperative row ARIA. This marks the older plugin deprecated so consumers get a clear signal, without removing anything.

## What changed

`packages/core/src/Table/plugins/rowExpansion/useTableRowExpansion.tsx`:
- `@deprecated` JSDoc on `useTableRowExpansion` and `useTableRowExpansionState`, each pointing to the tree plugin and the migration guide.
- `@deprecated` note on the file header.

`packages/core/src/Table/useTableRowExpansion.doc.mjs`:
- Both descriptions (full + dense) prefixed with `Deprecated: use useTableTreeData + useTableTreeState instead`, matching the `Grid` / `TopNavHeading` deprecation precedent.

The marker surfaces in editors (JSDoc strikethrough) and via `astryx component useTableRowExpansion`.

## Not a breaking change

The hooks still work and existing consumers are unaffected. There is no `no-deprecated` lint gate in the repo, so the plugin's own example block, story, and tests continue to pass. Removal and a version-keyed codemod follow in a later release (the codemod belongs to the removal version, since the plugin still exists today).

## Test plan

- `astryx component useTableRowExpansion` shows the `Deprecated:` marker and the migration guide.
- `tsc --project packages/core/tsconfig.json` and `tsconfig.docs.json`: clean.
- `useTableRowExpansion` tests: 14 pass.
- `eslint` on changed files: clean.
- `check:changesets`: valid.

### DIFF
--- a/.changeset/deprecate-row-expansion.md
+++ b/.changeset/deprecate-row-expansion.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[docs] Soft-deprecate useTableRowExpansion and useTableRowExpansionState in favor of the tree plugin (useTableTreeData + useTableTreeState). The hooks still work; JSDoc @deprecated tags and the docs point to the migration guide. Removal will come in a later release.
+
+@humbertovirtudes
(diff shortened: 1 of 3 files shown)


humbertovirtudes:
Superseded by #4609: we are keeping useTableRowExpansion and repurposing it as a detail-panel plugin, not deprecating it. Closing.

PR #4590, opened 2026-07-31
feat(list): add isFullBleed to align list item content with the container edge

### LINKED ISSUE #2626: XDSList/XDSListItem: no way to align list-item content to the container edge (optical alignment)

jiunshinn:
Fixes #2626

## Summary

Adds `isFullBleed?: boolean` to `List` so item content can align with full-bleed siblings such as a section heading while hover and selection backgrounds retain the item inset.

## Implementation

- The public API and story use `isFullBleed`.
- Each `ListItem` applies the cancellation on its own row, where it can read `--_item-inset-inline`.
- The cancellation is clamped to the padded container edge:

```css
margin-inline: calc(
  -1 * min(
    var(--_item-inset-inline),
    var(--container-padding-inline-start, 0px)
  )
);
```

This prevents zero-padding or already full-bleed surfaces from being pulled outside their content edge.

- `Item` publishes its density-dependent inset as `--_item-inset-inline` and derives `paddingInline` from the same variable. Theme overrides therefore keep padding and cancellation synchronized through the derived-variable registry.
- The `FullBleed` Storybook story publishes explicit 16px container padding variables so the example demonstrates the real container contract.

## Scope

Template adoption remains separate because PR #3938 also changes the detail-page template.

## Verification

- Focused List/theme suite: 185 tests passed
- Root build, Storybook build, core typecheck, and Storybook typecheck passed
- Chromium geometry checks passed for normal, zero-padding, and themed-inset fixtures
- `pnpm check:repo` and `git diff --check` passed


packages/core/src/List/List.tsx
```diff
@@ -104,6 +116,18 @@ const styles = stylex.create({
   withDividers: {
     gap: 0,
   },
+  // isEdgeAligned cancels the inline inset that Item applies to each row so
+  // row text aligns flush with sibling full-bleed content (e.g. a section
+  // heading). The negative margin mirrors Item's own density-dependent
+  // paddingInline (--spacing-2, or --spacing-3 for spacious), not the
+  // container padding vars: the goal is optical alignment with siblings,
+  // not a container bleed like Table/Divider.
+  edgeAligned: {
```
cixzhang: This mirrors `Item`'s `paddingInline`; a theme that changes it leaves the rows out by the difference.

cixzhang (changes requested):
Thanks for this — density-tracking is the right instinct, and the story reads exactly like the case in [#2626](https://github.com/facebook/astryx/issues/2626).

The cancelling margin hardcodes Item's inset (`--spacing-2`/`--spacing-3`), and themes move component padding routinely — butter already writes `paddingInline: var(--spacing-3)` on eight targets. Measured in Chromium on the `EdgeAligned` story: by default the heading and the row text both land at x=16; give `item` that same `--spacing-3` and the text moves to x=20 while the heading stays at 16. That is the off-by-4 your own description flags in the `detail-page` template, one level up.

Could `Item` publish its inline inset as a var the list reads, so the cancel derives rather than mirrors?

On naming — @cixzhang's note on the issue says `isFullBleed` is not accurate either, so I am leaving that call to her rather than picking one here.

[Reviewed by Robohands]

packages/core/src/List/List.tsx
jiunshinn: Respun in 64c1ea58a2 (rebased onto current main) — the cancel now derives instead of mirroring.

- `Item` publishes its inline inset as `--_item-inset-inline` and derives its own `paddingInline` from it (`--spacing-2`; `--spacing-3` for spacious).
- Custom properties only cascade downward, so the `<ul>` can never read a variable the rows set — the cancelling margin moved from the list element to each row (via ListContext), where the variable resolves: `margin-inline: calc(-1 * var(--_item-inset-inline))`. Geometry is unchanged (header untouched, hover/selection backgrounds still extend past the text by the inset).
- The themed-padding case is covered by a derived var registry entry (`item`: `paddingInline` → `--_item-inset-inline`), so a theme that writes `paddingInline: var(--spacing-3)` on `item` — the butter case — sets the variable in the same emitted rule and the cancel tracks it.

Re-ran your measurement in Chromium on the `EdgeAligned` story: by default the heading and row text both sit at x=16; with the themed override applied (`padding-inline: var(--spacing-3); --_item-inset-inline: var(--spacing-3)`, which is what `generateThemeRules` now emits) the row padding widens to 12px, the margin follows to −12px, and both stay at x=16. Same result for `density="spacious"`.

One boundary kept out of scope: the registry entry maps `paddingInline`, so a theme writing the `padding` shorthand isn't parsed into the inline var (the registry has no shorthand parsing today). Happy to extend if you'd rather cover that too.

Left the prop name as `isEdgeAligned` pending the naming decision on the issue.

cixzhang (changes requested):
Thanks for the respin — deriving the cancel from a published var instead of mirroring the values is exactly right, and I checked it under a real theme: with `item.paddingInline` overridden to `--spacing-3` and again to `--spacing-5`, the cancel tracks and the rows stay flush at all three densities and in RTL. That round's finding is closed.

Two changes and I'll take it.

**Clamp the cancel to the container's padding.** Right now it cancels the row's inset outright. In a Card that's 8px against 12px so nothing shows, but a popover pads 12px and a full-bleed area pads 0 — there the rows pull outside the content edge:

```
marginInline: calc(-1 * min(var(--_item-inset-inline), var(--container-padding-inline-start, 0px)))
```

The 0px case then needs no special handling; `min(8px, 0px)` is 0 and the list renders normally.

**And name it `isFullBleed`.** I said on [#2626](https://github.com/facebook/astryx/issues/2626) that the name wasn't quite accurate and I've changed my mind — read at the content level rather than the box level it's the same promise `Divider.isFullBleed` already makes: your content reaches the container's content edge. The amount differs per component, the meaning doesn't, and one name beats two conventions.

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks — f0422bf keeps the var-derived inset fix and closes my `isFullBleed` and 0px-clamp asks.

One clamp edge is still wrong. `ListItem.tsx:136` applies `--container-padding-inline-start` to both margins. With 16px start/4px end padding, the selected row over-cancels the end by 4px and paints past the outer border; RTL mirrors the spill. Using the matching start/end vars removes it without changing label alignment or the 0px case.

Main also moved under this in [#5288](https://github.com/facebook/astryx/pull/5288), and GitHub now reports a conflict. Could you fix the end-edge clamp, then rebase while preserving List's new `...props`/external `aria-labelledby` forwarding and both registry-test additions?

[Reviewed by Robohands]

jiunshinn:
Thanks — fixed in 80222c0. The clamp is now per edge: marginInlineStart clamps against --container-padding-inline-start and marginInlineEnd against --container-padding-inline-end, both keeping the min(var(--_item-inset-inline), var(--container-padding-inline-*, 0px)) shape. With 16px start / 4px end the end margin resolves to -4px, so the selected row stops at the content edge instead of painting past the outer border, RTL follows from the logical properties, and the symmetric and 0px cases are unchanged. Label alignment is untouched since the start-edge formula didn't change. I checked publishing: --container-padding-inline-end is already set everywhere the start var is (container.stylex/padding.stylex, Section, Layout, Markdown), so no new mechanism was needed. Added a regression test asserting each edge clamps against its own var — jsdom does no real layout, so it inspects the injected StyleX CSS, same pattern Switch's tests use. Also rebased onto main over #5288: List keeps the new ...props spread and the external aria-labelledby forwarding (all three of its forwarding tests pass), and both registry-test additions survived. List/Item/theme suites are green (267 + 829 tests) and the PR now reports mergeable.

rubyycheung (approved):
The full-bleed behavior looks good. Item content aligns cleanly with surrounding headings while the interactive row surface preserves its inset, and the implementation correctly tracks density, theme overrides, asymmetric container padding, and RTL.

cixzhang (approved):
Thanks for working through this edge compensation spec. These alignment details are tricky so appreciate you sticking with it :)

PR #4402, opened 2026-07-26
deps: bump js-yaml to ^5.2.2 to fix DoS (fixes #4399)

dhanji4U:
Fixes #4399

js-yaml 4.x vulnerable to DoS via !!omap and flow collections
- Patched in 5.2.2, requires overriding read-yaml-file>js-yaml and @changesets/parse>js-yaml
- Verified with pnpm changeset status - works, no safeLoad error
- Added js-yaml@5.2.2 to minimumReleaseAgeExclude

PR #4591, opened 2026-07-31
fix(deps): override @hono/node-server to ^2.0.5 (Windows path traversal, #4401)

### LINKED ISSUE #4401: deps: @hono/node-server 1.x -> 2.x (path traversal on Windows)

jiunshinn:
Fixes #4401 (Dependabot alert #106).

## What

Adds a pnpm override forcing `@hono/node-server` to `>=2.0.5`:

- `@modelcontextprotocol/sdk@1.29.0` (used by `apps/docsite`, directly and via `mcp-handler`) depends on `@hono/node-server ^1`, which resolved 1.19.14 — vulnerable to path traversal in `serve-static` on Windows via encoded backslashes (`%5C`).
- The fix landed in 2.0.5, a semver-major the sdk's `^1` range can never reach on its own, so this needs either an sdk release on 2.x or a targeted override. This PR takes the override route (same shape as the merged sharp override, #4449, and the js-yaml override, #4402).
- Overrides live in `pnpm-workspace.yaml` (pnpm 11 no longer reads `pnpm.overrides` from `package.json`), with a comment explaining why and when to drop it, following the existing sharp/js-yaml entries.

## Verification

**Resolution** — `pnpm install` resolves `@hono/node-server@2.0.11` (2.0.12 is held back by `minimumReleaseAge`); `pnpm why` confirms a single version workspace-wide, and the lockfile diff touches only this chain.

**API compatibility** — the sdk's only import from `@hono/node-server` is `getRequestListener(fetchCallback, { overrideGlobalObjects: false })` in `dist/*/server/streamableHttp.js`. 2.0.11 exports `getRequestListener` with the same signature including the `overrideGlobalObjects` option (verified in `dist/index.d.mts`). The sdk does not use `serve`, `createAdaptorServer`, or `serve-static`. 2.x peers on `hono ^4`, which the workspace already holds at `^4.12.25`.

**Runtime smoke test** — ran a full MCP session over the sdk's Node `StreamableHTTPServerTransport` (the only code path that loads `@hono/node-server`) with 2.0.11 installed: stateless per-request transport (the same mode the docsite `/mcp` route uses via `mcp-handler`), real HTTP server, sdk client. `initialize` → `tools/list` → `tools/call` all succeeded:

```
@hono/node-server resolved: @hono/node-server 2.0.11
initialize: OK (server info: { name: 'smoke', version: '1.0.0' } )
tools/list: OK -> [ 'ping' ]
tools/call ping: OK -> [{"type":"text","text":"pong 4401"}]
SMOKE TEST PASSED
```

(Note: the docsite route itself goes through the sdk's `webStandardStreamableHttp` transport, which doesn't touch `@hono/node-server` at all — the Node transport smoke above is the worst-case path.)

**Repo checks** — `pnpm -F @astryxdesign/docsite generate && test`: 303/303 pass. `pnpm -F @astryxdesign/docsite typecheck`: clean. `check:sync` / `check:package-boundaries` / `check:changesets` / `check:demo-media` / `check:executable-bits`: all green. Pre-existing peer warnings (`mcp-handler` wants sdk 1.26.0, `@stylexjs/unplugin` wants unplugin ^2) are unchanged by this PR.

**Engines note** — `@hono/node-server` 2.x requires Node >= 20 (1.x was >= 18.14.1); the repo is on Node 24 (`.nvmrc`, CI), so no impact.

## Changeset

None — only `pnpm-workspace.yaml` + `pnpm-lock.yaml` change, no published package is modified. Same as the merged sharp override #4449.

## Conflict note

This touches the same `overrides:` block as open PR #4101 (prune stale pnpm overrides / move pins to the catalog) — whichever lands second will need a trivial rebase.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -23,6 +23,7 @@ overrides:
   esbuild: '>=0.28.1'
   '@babel/core': ^7.29.6
   hono: ^4.12.25
+  '@hono/node-server': ^2.0.5
   read-yaml-file: ^2.1.0
   read-yaml-file>js-yaml: ^4.2.0
   '@changesets/parse>js-yaml': '>=4.2.0'
@@ -1606,9 +1607,9 @@ packages:
     peerDependencies:
       react: '>= 16 || ^19.0.0-rc'
 
-  '@hono/node-server@1.19.14':
-    resolution: {integrity: sha512-GwtvgtXxnWsucXvbQXkRgqksiH2Qed37H9xHZocE5sA3N8O8O8/8FA3uclQXxXVzc9XBZuEOMK7+r02FmSpHtw==}
-    engines: {node: '>=18.14.1'}
+  '@hono/node-server@2.0.11':
+    resolution: {integrity: sha512-bjD221KPLoJTWUwso1J6fGKiTXEUFedG/s0visavY4zakFPkeGURMRNly+FhBHs7T8Dz4qHaZIMX9ZoJHSJtKA==}
+    engines: {node: '>=20'}
     peerDependencies:
       hono: ^4.12.25
 
@@ -7147,7 +7148,7 @@ snapshots:
     dependencies:
       react: 19.2.7
 
-  '@hono/node-server@1.19.14(hono@4.12.27)':
+  '@hono/node-server@2.0.11(hono@4.12.27)':
     dependencies:
       hono: 4.12.27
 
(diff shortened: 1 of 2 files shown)


imdreamrunner (approved):
Thanks!

PR #4594, opened 2026-07-31
fix(dropdown-menu): follow input modality for initial focus so pointer opens don't highlight the first item

### LINKED ISSUE #4477: DropdownMenu / MoreMenu: first item looks selected when opened by pointer (no way to opt out of initial focus)
## Summary

When a `DropdownMenu` (or `MoreMenu`) is opened with a **mouse click**, focus moves to the first enabled item and its focus ring renders. To users it reads as "this item is already selected" — several people on my team have asked whether the first action was pre-chosen or somehow highlighted by mistake.

For keyboard opens this is exactly right and shouldn't change. The problem is only that pointer opens get the same treatment, and there's no prop to opt out.

## What I ran into

A table row actions menu — Edit details / Edit permissions / Delete role:

```tsx
<DropdownMenu
  button={{ label: `Actions for ${role.name}`, icon: <MoreVertical />, isIconOnly: true, variant: 'ghost' }}
  hasChevron={false}
>
  <DropdownMenuItem icon={Pencil} label="Edit details" onClick={...} />
  <DropdownMenuItem icon={SlidersHorizontal} label="Edit permissions" onClick={...} />
  <DropdownMenuItem icon={<Trash2 />} label={<span>Delete role</span>} onClick={...} />
</DropdownMenu>
```

Click the three-dot trigger → "Edit details" is visibly highlighted before the pointer goes anywhere near it. It also shifts depending on state: when the first item is `isDisabled`, the highlight lands on whichever item is first enabled, so the "selected" row appears to move around between rows in the same table.

Same behavior via the `items` data prop and via `MoreMenu`, so it looks like it's in the shared keyboard/focus path rather than either wrapper.

## Why it's awkward to work around

`DropdownMenuProps` exposes `button`, `isMenuOpen`, `onOpenChange`, `menuWidth`, `onClick`, `hasChevron`, `placement`, `data-testid` — nothing focus-related. `DropdownMenuItemProps` has `icon`, `label`, `description`, `onClick`, `isDisabled`, `endContent`. So there's no supported way to say "don't focus anything until the user arrows in", short of CSS against internals, which I'd rather not ship.

## Proposals (either would solve it)

**A. Match focus behavior to the input modality** (my preference — no new API, fixes it everywhere at once)
Focus the first item when the menu is opened via keyboard (`Enter` / `Space` / `ArrowDown` on the trigger), but on pointer open leave focus on the menu container so nothing is visually highlighted. First `ArrowDown` then moves to item 1. This is what most menu implementations do, and it pairs naturally with `:focus-visible`.

**B. Add an opt-out prop**

```tsx
<DropdownMenu hasInitialFocus={false}> ... </DropdownMenu>
```
Defaults to `true` so nothing changes for existing users.

If the highlight is intended to be `:focus-visible`-driven already, then this may just be a bug where the ring shows on programmatic focus after a pointer open.

## Acceptance criteria

- [ ] Opening by mouse click shows no item highlighted; the menu is still keyboard-operable and `ArrowDown` moves to the first item.
- [ ] Opening by keyboard still focuses the first enabled item immediately (unchanged).
- [ ] Behavior is the same for `DropdownMenu` items-array mode, `DropdownMenu` compound mode, and `MoreMenu`.
- [ ] Disabled leading items are still skipped when focus does move.
- [ ] Screen reader announcement of the menu and its items is unchanged.

jiunshinn:
Fixes #4477

## Summary

Opening a `DropdownMenu` (or `MoreMenu`) with a mouse click no longer renders the first enabled item as if it were already selected. Initial focus now follows the input modality:

- **Keyboard open** (Enter / Space / ArrowDown on the trigger): unchanged, the first enabled item is focused per the APG menu-button pattern.
- **Pointer open** (trigger click): focus moves to the menu container itself (`role="menu"`, `tabIndex={-1}`), so no item is highlighted. The first ArrowDown then moves to the first enabled item; typeahead, Escape, Tab, and hover-to-highlight all keep working because key events still originate inside the menu.

This is proposal A from the issue (modality-aware focus, no new API), applied in `DropdownMenu`'s shared open path, so it covers items-array mode, compound mode, and `MoreMenu` at once.

## Why A and not a `:focus-visible` change or an opt-out prop

The highlight is intentionally `:focus`-driven, not `:focus-visible`-driven: `DropdownMenuItem` styles `backgroundColor` under `:focus`, and `menuItemHover.ts` moves DOM focus onto the hovered item so hover and keyboard navigation share a single focus-driven highlight. Switching the style to `:focus-visible` would therefore break hover highlighting (hover focus is programmatic and does not match `:focus-visible` after pointer input). The correct minimal fix is to not focus any item on a pointer open.

The repo already has this exact pattern: `DropdownMenuSubMenu`'s flyout container is `tabIndex={-1}`, hover-open does not steal focus, and an item-less flyout falls back to focusing the container. This PR mirrors that for the root menu, including the fallback (a keyboard open with no focusable items now focuses the container so Escape/Tab still work).

A `hasInitialFocus` prop (proposal B) was not added: A fixes every existing usage without new API surface, and B could be layered on later if a consumer needs to opt back in.

## Implementation notes

- `openModalityRef` tracks how the open was initiated; the deferred rAF focus step reads it and then resets to `'keyboard'`, so programmatic controlled opens (`isMenuOpen` set without a trigger gesture) keep the previous first-item focus behavior.
- Synthesized clicks with `event.detail === 0` (screen reader / AT activation) are treated as keyboard opens, so AT users still land on the first item and announcements are unchanged.
- Controlled mode: a pointer click on the trigger stamps the modality before `onOpenChange(true)`, and the sync effect places focus accordingly.
- Removed the vestigial `shouldFocusOnOpenRef` (initialized false, never set) that this change replaces.

## Acceptance criteria from the issue

- [x] Opening by mouse click shows no item highlighted; the menu is still keyboard-operable and ArrowDown moves to the first item.
- [x] Opening by keyboard still focuses the first enabled item immediately (unchanged).
- [x] Behavior is the same for `DropdownMenu` items-array mode, `DropdownMenu` compound mode, and `MoreMenu` (all share the open path; each is covered by a test).
- [x] Disabled leading items are still skipped when focus does move (both keyboard open and first ArrowDown after pointer open).
- [x] Screen reader announcement of the menu and its items is unchanged: roles, names, and `aria-*` are untouched, and detail-0 activations keep the first-item focus path.

## Relation to #3980 (`DropdownMenuItem` `isSelected`, focus the current selection on open)

Compatible by design: this PR only decides **whether** an item receives initial focus (keyboard/programmatic yes, pointer no). #3980 refines **which** item the keyboard path focuses (the selected one instead of the first). Both changes meet in the same rAF focus step (`focusMenuOnOpen`); whichever lands second rebases by calling #3980's selected-item focus inside the keyboard branch. Arguably the combination is stronger: a pointer open of a selection menu will show only the `aria-current` styling on the selected item rather than a focus ring on the first item.

## Testing

- `pnpm exec vitest run packages/core/src/DropdownMenu packages/core/src/MoreMenu packages/core/src/ContextMenu`: 141 passed.
- Full `packages/core` suite: 5724 passed, 1 failed; the failure is `TabList.test.tsx` ("moves focus between items with ArrowDown and ArrowUp"), a pre-existing rAF timing race in `TabMenu`'s own separate open path (code untouched by this diff); it passes in isolation on this branch.
- `tsc --noEmit` on `packages/core`: clean. `eslint` on changed files: clean. `check:sync`, `check:changesets`: pass.
- New tests: pointer open focuses the container (items mode, compound mode, MoreMenu, controlled mode); first ArrowDown moves to the first enabled item and skips a disabled leading item; keyboard opens via Enter and ArrowDown focus the first enabled item; detail-0 click keeps first-item focus; programmatic controlled open keeps first-item focus; Escape and Tab still close after a pointer open.
- Updated `DropdownMenuSubMenu` tests that used "first item focused after click-open" as their rAF settle anchor to anchor on the new deterministic state (container focused).

## Open questions

- `ContextMenu` has its own open path and still focuses the first item on right-click open. The APG context-menu pattern generally expects focus in the menu on open, and the issue scopes DropdownMenu/MoreMenu, so it is left unchanged; happy to align it in a follow-up if maintainers want the same modality split there.
- After a pointer open, focus rests on the `role="menu"` container. Browsers only draw a default outline for `:focus-visible`, which does not match here, so no ring appears; if the team prefers an explicit `outline: none` guard on the container for older engines, easy to add.
- The pre-existing `TabMenu` rAF timing flake above may be worth a separate fix (same settle-anchor treatment as the submenu tests).

### DIFF
--- a/.changeset/dropdown-menu-pointer-open-focus.md
+++ b/.changeset/dropdown-menu-pointer-open-focus.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] DropdownMenu/MoreMenu: opening with a pointer no longer highlights the first item as if it were selected (#4477). Initial focus now follows the input modality: keyboard opens (Enter/Space/ArrowDown on the trigger) still focus the first enabled item per the APG menu-button pattern, while pointer opens focus the menu container itself so the first ArrowDown moves to item 1. Synthesized clicks (detail 0, e.g. screen reader activation) and programmatic controlled opens keep the first-item focus behavior. Covers data-driven items mode, compound mode, and MoreMenu, which share the open path.
+@jiunshinn
--- a/packages/core/src/DropdownMenu/DropdownMenu.test.tsx
+++ b/packages/core/src/DropdownMenu/DropdownMenu.test.tsx
@@ -10,8 +10,9 @@
  */
 
 import {describe, it, expect, vi, beforeEach} from 'vitest';
-import {render, screen, fireEvent} from '@testing-library/react';
+import {render, screen, fireEvent, waitFor} from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
+import {useState} from 'react';
 import {DropdownMenu} from './DropdownMenu';
 import {DropdownMenuItem} from './DropdownMenuItem';
 import {Divider} from '../Divider';
@@ -214,9 +215,8 @@ describe('DropdownMenu', () => {
       const trigger = screen.getByRole('button', {name: /Actions/});
       trigger.focus();
       await user.click(trigger);
-      expect(
-        screen.getByRole('menuitem', {name: 'Edit', hidden: true}),
-      ).toHaveFocus();
+      // Pointer opens focus the menu container, not the first item (#4477).
+      expect(screen.getByRole('menu', {hidden: true})).toHaveFocus();
 
       const popoverEl = screen
         .getByRole('menu', {hidden: true})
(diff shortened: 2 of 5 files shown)


cixzhang:
Thanks @jiunshinn for another fix! I ended up solving this a different way by adjusting the hover to also modify aria active descendant so we may not need this anymore. Thank you for helping!

cixzhang:
Oh actually I guess this issue reads slightly different from the previous one I was looking at. Let me see.

cixzhang (approved):
Thanks this seems like a reasonable fix. I checked it on mobile and appears to work as expected.

We have some merge conflicts so I won't be able to merge until they're resolved

jiunshinn:
Thanks for the review and for checking it on mobile!

Rebased onto latest main — the conflicts are resolved and the branch is mergeable again.

One note on how the merge with #4666 was resolved: that PR introduced the deferred initial-focus step (`shouldFocusOnOpenRef`) so the focus restore captures the trigger before focus moves into the menu. I kept that architecture and routed the modality logic through it — the deferred focus step now honors `openModalityRef` (keyboard/AT/programmatic opens -> first enabled item, pointer opens -> menu container, with the container also serving as the all-items-disabled fallback). `openAndFocus` just stamps the modality and sets the flag before `popover.show()`.

I also re-anchored #4666's new light-dismiss test: it clicked the trigger and asserted the first item had focus as its settle anchor, which is exactly the pre-selected highlight this PR removes for pointer opens. It now anchors on the menu container having focus; the focus-restore assertion itself is unchanged.

Menu suites (DropdownMenu / MoreMenu / ContextMenu) pass 149/149 locally, typecheck and the pre-commit checks are green.

jiunshinn:
Hi @cixzhang, Is there anything else I can do help for this PR get merged?

PR #4597, opened 2026-07-31
feat(selectable-card): toggle selection on Enter as well as Space

freddymeta:
## What

`SelectableCard`'s focusable control is a visually-hidden `<input type="checkbox">` (the card `<div>` itself has no role or `tabIndex`). A native checkbox toggles on **Space** but ignores **Enter** — so today a keyboard user who focuses the card can toggle it with Space but pressing Enter does nothing.

This PR adds an `onKeyDown` handler on that hidden input that also toggles selection on **Enter**, calling `onChange(!isSelected)` — mirroring the existing click/Space path.

## Why

Many keyboard users carry a button/link mental model where **Enter activates the focused control**. When Enter does nothing on a card that visibly looks activatable, it reads as broken. Adding Enter as an additional toggle key closes that gap.

## Behavior

- **Additive and non-breaking.** Space keeps its native handling; the pointer/click path, ARIA, the checkbox role, and all other behavior are unchanged. Enter is simply *also* accepted now.
- **No double-toggle.** Space is handled natively by the checkbox — the new handler intentionally does **not** touch Space, so a Space press still results in exactly one toggle. There's a regression test for this.
- **Disabled-safe.** The handler guards on `!isDisabled` (mirroring the existing click guard), and the input is already `disabled` when the card is disabled.

## Semantics — inviting your call

Strict ARIA APG treats a checkbox as **Space-only** (Enter is reserved for form submission), so making Enter toggle is a deliberate DX departure rather than something ARIA prescribes. I'm proposing it because it matches user expectations for an activatable card, but I'd like maintainers to confirm you want this departure. **Happy to close this if you'd prefer to keep strict checkbox semantics.**

## Tests

Added to `SelectableCard.test.tsx`:
- Enter toggles from unselected → `onChange(true)` and from selected → `onChange(false)`
- Enter does **not** toggle when `isDisabled`
- Regression: Space still toggles **exactly once** (not doubled by the new handler)

All existing `SelectableCard` tests remain green; full suite passes locally.

## Docs & changeset

- `SelectableCard.doc.mjs`: added a best-practice note that Space **or** Enter toggles when the card is focused (EN + dense).
- Changeset: `@astryxdesign/core` patch, `[feat]`.

### DIFF
--- a/.changeset/selectable-card-enter-key.md
+++ b/.changeset/selectable-card-enter-key.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] SelectableCard: pressing Enter now toggles selection, in addition to Space, when the card is focused
+@freddymeta
--- a/packages/core/src/SelectableCard/SelectableCard.doc.mjs
+++ b/packages/core/src/SelectableCard/SelectableCard.doc.mjs
@@ -12,6 +12,7 @@ export const docs = {
     bestPractices: [
       {guidance: true, description: 'Use for plan pickers, filter chips, or option grids.'},
       {guidance: true, description: 'For single-select track one ID; for multi-select use a Set.'},
+      {guidance: true, description: 'When focused, toggle selection with Space or Enter.'},
       {guidance: false, description: 'Use for navigation; use ClickableCard for that.'},
     ],
     anatomy: [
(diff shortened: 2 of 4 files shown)


cixzhang (approved):
Enter-as-toggle is a clean, non-breaking addition — Space keeps native handling, disabled is guarded, and the no-double-toggle regression test covers the concern. The color-contrast audit flag is pre-existing and unrelated to this keyboard change. Approving.

PR #4598, opened 2026-07-31
feat(selector): add search adornment slots to Selector & MultiSelector

freddymeta:
## What

Adds `searchStartContent` / `searchEndContent` slots to **Selector** and **MultiSelector** for decorative adornments beside the dropdown search input (only in `hasSearch` mode) — most commonly a leading search/magnifier `Icon`.

```tsx
<Selector hasSearch searchStartContent={<Icon icon="search" size="sm" />} … />
```

## Why

The search input had no adornment slot, so a leading magnifier could only be a CSS pseudo-element (`::before` with an inline-SVG `mask`) — which can't be a real icon component and can't carry a handler. This is the composition-friendly fix: pass a real `<Icon>`.

## How (accessibility)

The adornments render inside the search row **wrapped `aria-hidden`** and are non-interactive, so:
- the search input keeps its `role="combobox"`, focus, and keyboard behavior (unchanged),
- the adornments never surface as extra controls inside the listbox.

The search wrapper becomes a flex row; the input fills the remaining space. Interactive clear-style affordances still belong on the value via `onChange` — these slots are decorative and deliberately **not** wired to the search query.

## Scope

Part of the EPS-on-Astryx upstreaming (paired with the theme-target PR). This is the piece that genuinely needed component API rather than a theme hook. A full `renderSearch` override was considered but rejected — it would let a consumer break the combobox/aria-activedescendant contract the component carefully maintains; adornment slots give the needed flexibility (the magnifier) without that risk.

## Testing

- Selector 79 + MultiSelector 74 = 153 tests pass, incl. new ones: both slots render `aria-hidden` beside the input (both components) and don't render without `hasSearch`.

Non-breaking; default appearance unchanged (no adornments unless provided).

### DIFF
--- a/.changeset/selector-search-adornments.md
+++ b/.changeset/selector-search-adornments.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] Selector & MultiSelector: add `searchStartContent` / `searchEndContent` slots for decorative adornments beside the dropdown search input (only in `hasSearch` mode) — e.g. a leading search/magnifier `Icon`. The slots are rendered `aria-hidden` beside the input, so the input keeps its combobox role, focus, and keyboard behavior; they are not wired to the search query. Non-breaking.
+@freddymeta
--- a/packages/core/src/MultiSelector/MultiSelector.doc.mjs
+++ b/packages/core/src/MultiSelector/MultiSelector.doc.mjs
@@ -114,6 +114,18 @@ export const docs = {
           default: "'Search...'",
         },
         {
+          name: 'searchStartContent',
+          type: 'ReactNode',
+          description:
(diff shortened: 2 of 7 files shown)


freddymeta:
Closing in favor of #4596.

The driving use case here — a decorative leading magnifier in the search input — is covered by the `selector-search` / `selector-search-input` theme targets added in #4596: a theme can draw the magnifier via a `::before` on the search wrapper (`defineTheme`), app-wide, without adding component props.

Since these adornment slots were scoped as decorative/non-interactive/`aria-hidden`, a theme target fully serves the need and avoids adding four new props (`searchStartContent`/`searchEndContent` across both Selector and MultiSelector). If a genuinely *interactive* search adornment is needed later, that can be a single, scoped prop at that point.

PR #4599, opened 2026-07-31
feat(progressbar): add opt-in target markers

freddymeta:
## Proposal: opt-in target markers for `ProgressBar`

This adds a new **opt-in** `markers` prop to `ProgressBar` and is offered as a **proposal** — the API shape is deliberately conservative, but I'd like maintainer input on a few open questions (listed at the bottom) before it's considered settled.

### What it does

`markers` draws one or more **fixed vertical target lines** on the track — a goal line, an SLA threshold, a budget cap, a milestone. Each marker is placed by `value` in the **same `0..max` scale as the bar's `value` prop**, and it **stays visible whether progress is below or past it** (it layers above the fill).

```tsx
<ProgressBar
  value={45}
  label="Fundraiser"
  hasValueLabel
  markers={[{value: 80, label: 'Goal'}]}
/>
```

### API

```ts
markers?: ReadonlyArray<{value: number; label?: string}>;
```

The object shape **mirrors Slider's `marks?: {value: number; label?: string}[]`** so the two components stay consistent, and it leaves room to grow (e.g. a future visible-label option) without a breaking change. A named `ProgressBarMarker` interface is exported so consumers can type their arrays.

### Behavior

- **Positioning** — `value` → percent (`(value / max) * 100`) → `insetInlineStart: <pct>%`, so it's **RTL-safe** (matching how Slider positions marks and how the indeterminate fill mirrors under RTL). The marker is centered on its position with an RTL-mirrored `translateX`.
- **Clamp, not drop, for out-of-range** — a marker `value` past `max` pins at 100%, below `0` pins at 0%, matching the bar's own `clampedValue`. Non-finite values (`NaN`/`Infinity`) are dropped, reusing the component's existing finite-guard hygiene.
- **Indeterminate** — markers make no sense without a determinate value, so they are **not rendered** when `isIndeterminate`.
- **Non-breaking** — omitting `markers` (or passing `[]`) renders nothing new; all existing behavior is byte-identical.

### Theming

Each marker carries the new theme target **`astryx-progressbar-marker`** (via `themeProps('progressbar-marker')`), consistent with the existing `astryx-progressbar-track` / `-fill` targets, and documented in the component's `theming.targets`. Color/width are semantic-token based (`--color-border-emphasized`), so themes can restyle it. StyleX only — no raw CSS.

### Accessibility

Markers are **decorative by default** (`aria-hidden="true"`) — the `progressbar` role carries a single `aria-valuenow`, and extra visual marks shouldn't pollute it or the `aria-valuetext`. When a marker has a `label`, that marker instead exposes it as an `aria-label` (and drops `aria-hidden`) so assistive tech can perceive it if the consumer opts in. This is the decision I'm least certain about — see below.

### Why a prop and not composition

`ProgressBar` is documented as intentionally minimal ("compose additional labels/icons alongside the bar"). An **on-track** marker is the justified exception: it has to live *inside* the track and be positioned by value against the same scale as the fill — it can't be composed externally without duplicating the track's geometry.

### Open questions for review

1. **API shape** — object array (`{value, label?}[]`, chosen here for Slider-consistency + extensibility) vs. the simpler `number | number[]`?
2. **Prop name** — `markers` vs `targets`?
3. **a11y default** — decorative-by-default with `label` → `aria-label` (chosen), vs. keeping markers purely decorative and treating `label` as reserved-for-future?
4. **Out-of-range** — clamp to the track edges (chosen) vs. drop out-of-range markers entirely?

### Tests

Added coverage for: positioning at the right percent (default and custom `max`), a marker past the current value staying visible, clamp of out-of-range values, dropping non-finite values, no markers in indeterminate mode, decorative vs. labeled a11y, `aria-valuetext` left untouched, and that omitting `markers` renders nothing (non-breaking). The `themingTargets` guard passes with the new documented target.


cixzhang:
Since Slider sets precedent "marks" would be the most consistent prop name for the progress bar version and I would expect the behavior should match the slider mark behavior as well

cixzhang:
Regarding a11y my sense is that if it's information that sighted users can receive then AT users should be able to access it to. So marks should be perceivable somehow.

packages/core/src/ProgressBar/ProgressBar.tsx
```diff
@@ -87,6 +110,14 @@ export interface ProgressBarProps extends BaseProps<HTMLDivElement> {
...
    */
   isIndeterminate?: boolean;
+  /**
+   * Target markers drawn on the track at fixed points in the same `0..max`
+   * scale as `value` — e.g. a goal line. Markers stay visible whether progress
+   * is below or past them. A marker with a `label` reveals it via a `Tooltip`
+   * on hover/focus (and names it for assistive tech); an unlabeled marker is
+   * decorative. Ignored when `isIndeterminate` is true.
+   */
+  markers?: ReadonlyArray<ProgressBarMarker>;
```
cixzhang: Could we match stepper and call this "marks"?

packages/core/src/ProgressBar/ProgressBar.tsx
```diff
@@ -31,6 +31,7 @@ import {
 import {mergeProps} from '../utils';
 import type {BaseProps} from '../BaseProps';
 import {themeProps} from '../utils/themeProps';
+import {Tooltip} from '../Tooltip/Tooltip';
```
cixzhang: Since this is conditionally rendered perhaps we can apply the same lazy tooltip pattern?

packages/core/src/ProgressBar/ProgressBar.tsx
```diff
@@ -40,6 +41,28 @@ import type {ProgressBarVariantMap} from './index';
...
+   * outside the range are clamped to the track edges.
+   */
+  value: number;
+  /**
+   * Labels the marker. When provided, the marker is wrapped in a `Tooltip`
+   * that reveals the label on hover and keyboard focus — giving sighted users
+   * a visible affordance and assistive tech an accessible name. Without a
+   * label the marker is purely decorative (`aria-hidden`).
+   */
+  label?: string;
```
cixzhang: Hmm what does it mean for a mark to be decorative? I imagine the vast major of them are semantic in some way. Any reason we shouldn't make label required?

cixzhang (changes requested):
Thanks this is looking good. I guess we should try to match the marks with steeper visually for consistency but it may be hard to find colors with the right contrast. Likely The emphasized divider color is the most semantically correct for these cases so we could try that. I think the bold colors we use for these bars will make it challenging so may need to loop in design for help.

Otherwise some lightweight feedback on API

packages/core/src/ProgressBar/ProgressBar.tsx
freddymeta: i'm thinking in some cases, it may be taken from the number, e.g. 70%.
others can be more complex, e.g. Q1 target: 50%.

i think it makes sense just make it required. will amend.

packages/core/src/ProgressBar/ProgressBar.tsx
freddymeta: Done — renamed `markers` → `marks` (and `ProgressBarMarker` → `ProgressBarMark`, the theme target to `progressbar-mark`, the height var to `--progressbar-mark-height`) to match Slider's `marks`.

_— on behalf of Freddy_

packages/core/src/ProgressBar/ProgressBar.tsx
freddymeta: Done — the mark tooltip now lives in a separate `ProgressBarMarkTooltip` module loaded via `React.lazy` + `Suspense` (fallback = the bare mark), same pattern as the Timestamp hover card. A ProgressBar with no marks bundles no tooltip code.

_— on behalf of Freddy_

packages/core/src/ProgressBar/ProgressBar.tsx
freddymeta: Agreed — `label` is now **required**. A mark always stands for something meaningful (a goal/threshold/quarter target), so the decorative/`aria-hidden` unlabeled branch is gone: every mark is a focusable, labeled Tooltip trigger. The label can be the value itself (`'70%'`) or richer (`'Q1 target: 50%'`).

_— on behalf of Freddy_

freddymeta:
@cixzhang — all three API notes addressed and pushed (rebased clean onto `main`, CI green):

- **`markers` → `marks`** (+ `ProgressBarMark` type, `progressbar-mark` target, `--progressbar-mark-height` var) to match Slider.
- **Lazy tooltip** — split into `ProgressBarMarkTooltip` (React.lazy + Suspense); no tooltip code bundled when there are no marks.
- **`label` now required** — every mark is a focusable, labeled trigger; dropped the decorative/`aria-hidden` path.

On the visual side you raised: agreed the mark color should read as intentional against the bold bar colors. I've left the mark on its current token for now since, as you noted, finding contrast against every variant likely needs design input — happy to switch to the emphasized-divider color (or whatever design lands on) in a follow-up so this PR stays API-focused. Let me know if you'd rather bundle it here.

_— on behalf of Freddy_

packages/core/src/ProgressBar/ProgressBar.tsx
```diff
@@ -199,6 +255,35 @@ const styles = stylex.create({
...
+  // `role="progressbar"` element (unchanged DOM). The track no longer clips, so
+  // its height — default track height (8px), themeable via
+  // `--progressbar-mark-height` — may exceed the bar and overhang; the
+  // centering translate keeps any overhang symmetric. Positioned horizontally
+  // via `insetInlineStart`; the translate mirrors under RTL.
+  mark: {
+    position: 'absolute',
+    top: '50%',
+    width: 'var(--progressbar-mark-width, 2px)',
+    height: 'var(--progressbar-mark-height, 8px)',
```
cixzhang: We likely don't need these public CSS vars since the mark is directly overridable via component part.

cixzhang (approved):

packages/core/src/ProgressBar/ProgressBar.tsx
freddymeta: Agreed — dropped all three `--progressbar-mark-*` vars. The mark's `width`/`height`/`backgroundColor` are now plain values on the base style, directly overridable via the `progressbar-mark` target with `defineTheme` (e.g. `{base: {height: '16px', backgroundColor: 'red'}}`). Updated the docs (removed the `vars` block), example, story, and tests to match. No change to default rendering or themeability — just a simpler surface.

_— on behalf of Freddy_

PR #4600, opened 2026-07-31
feat(dropdown-menu): expose themeable slots for section, divider, and caret

athz:
## What

`DropdownMenu` (which `ContextMenu` delegates its item rendering to) renders several affordances as bare elements with no theme slot, so themes can only reach them through fragile structural selectors:

- the **section heading** — a bare `<div>`, targetable only as "the sole non-role child of a `role="group"`"
- the **menu divider** — the global `astryx-divider` slot can't be narrowed to "dividers inside a menu", forcing a descendant selector like `.astryx-dropdown-menu .astryx-divider`
- the **submenu caret** — a bare `<span>` around the chevron, with no slot at all
- the **checked radio dot** — a bare child span of the radio marker, with no slot of its own

This exposes a dedicated theme slot on each via `themeProps`, so themes can style them directly:

| Element | New slot |
| --- | --- |
| Section heading | `astryx-dropdown-menu-section` |
| Menu divider | `astryx-dropdown-menu-divider` (kept alongside the base `astryx-divider`) |
| Submenu caret | `astryx-dropdown-menu-caret` |
| Checked radio dot | `astryx-dropdown-menu-radio-dot` |

Documented in the component's theming targets. `ContextMenu` reuses the same item rendering path, so it inherits every slot automatically.

## Why

Themes currently have no stable hook for these parts and must reach into DropdownMenu's internal DOM structure, which breaks whenever the markup shifts. A theming layer that only wants to resize the caret, inset the divider, restyle the section label, or recolor the radio dot has to write brittle selectors instead of targeting a named slot — exactly the coupling the `themeProps` slot system exists to prevent.

## Scope

Purely additive theme-slot exposure — no behavior, layout, or a11y change. The menu divider keeps `role="separator"`; the section heading keeps `aria-hidden`; the caret and radio-dot markup are unchanged apart from the added class/data attributes.

There is a related, separate gap where `DropdownMenuCheckboxItem` hand-rolls its checkmark marker (`<Icon icon="check">`) instead of composing the real `CheckboxInput` the way `CheckboxListItem` does. That's a component refactor rather than a theming-exposure change, so it's intentionally left out of this PR and better handled on its own.

## Testing

- New tests assert the slot classes render on the section heading, menu divider (both slots), submenu caret, and the checked radio dot (and confirm the unchecked radio has no dot slot).
- `@astryxdesign/core`: full DropdownMenu suite passes (74 tests), plus `typecheck`, `typecheck:docs`, `sync-exports --check`, and lint.

### DIFF
--- a/packages/core/src/DropdownMenu/DropdownMenu.doc.mjs
+++ b/packages/core/src/DropdownMenu/DropdownMenu.doc.mjs
@@ -35,6 +35,14 @@ export const docs = {
         visualProps: ['size'],
         states: ['checked', 'disabled'],
       },
+      {
+        className: 'astryx-dropdown-menu-radio-dot',
+        visualProps: ['size'],
+        states: ['checked', 'disabled'],
+      },
+      {className: 'astryx-dropdown-menu-section-heading'},
+      {className: 'astryx-dropdown-menu-divider'},
+      {className: 'astryx-dropdown-menu-indicator-icon'},
     ],
     vars: [
       {name: '--_dropdown-menu-radius', description: 'Border radius of the menu popup', default: 'var(--radius-element)', private: true},
--- a/packages/core/src/DropdownMenu/DropdownMenuSubMenu.tsx
+++ b/packages/core/src/DropdownMenu/DropdownMenuSubMenu.tsx
@@ -474,11 +474,19 @@ export function DropdownMenuSubMenu(
   );
 
   const endAffordance = hasSpinner ? (
-    <span {...stylex.props(triggerStyles.caret)}>
+    <span
+      {...mergeProps(
+        themeProps('dropdown-menu-indicator-icon'),
+        stylex.props(triggerStyles.caret),
+      )}>
       <Spinner size="sm" />
     </span>
   ) : (
-    <span {...stylex.props(triggerStyles.caret)}>
+    <span
+      {...mergeProps(
+        themeProps('dropdown-menu-indicator-icon'),
+        stylex.props(triggerStyles.caret),
+      )}>
       <Icon icon="chevronRight" size="sm" color="secondary" />
     </span>
   );
--- a/packages/core/src/DropdownMenu/renderDropdownItems.tsx
+++ b/packages/core/src/DropdownMenu/renderDropdownItems.tsx
@@ -17,6 +17,7 @@ import {
   typeScaleVars,
   colorVars,
 } from '../theme/tokens.stylex';
+import {mergeProps, themeProps} from '../utils';
 import type {
   DropdownMenuItemData,
   DropdownMenuOption,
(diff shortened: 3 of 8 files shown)


packages/core/src/DropdownMenu/DropdownMenu.doc.mjs
```diff
@@ -35,6 +35,10 @@ export const docs = {
         visualProps: ['size'],
         states: ['checked', 'disabled'],
       },
+      {className: 'astryx-dropdown-menu-radio-dot'},
```
cixzhang: You will likely need the same states and visual props as the radio outer container. Just for the dot. I don't think it affects the others.

packages/core/src/DropdownMenu/DropdownMenuSubMenu.tsx
```diff
@@ -474,11 +474,19 @@ export function DropdownMenuSubMenu(
...
+        themeProps('dropdown-menu-caret'),
+        stylex.props(triggerStyles.caret),
+      )}>
       <Spinner size="sm" />
     </span>
   ) : (
-    <span {...stylex.props(triggerStyles.caret)}>
+    <span
+      {...mergeProps(
+        themeProps('dropdown-menu-caret'),
```
cixzhang: This might be called "indicator icon" in another component like selector. If you can double check we'd ideally want them to be consistent.

packages/core/src/DropdownMenu/renderDropdownItems.tsx
```diff
@@ -57,15 +58,26 @@ export function renderDropdownItems(items: DropdownMenuOption[]): ReactNode {
...
         <div
           key={getSectionKey(option, i)}
           role="group"
           aria-label={option.title}>
           {option.title && (
-            <div {...stylex.props(styles.sectionHeading)} aria-hidden="true">
+            <div
+              aria-hidden="true"
+              {...mergeProps(
+                themeProps('dropdown-menu-section'),
```
cixzhang: We can perhaps be specific and make this "section heading" so it's clearly about theming the text

cixzhang (changes requested):
Thanks for adding these targets! They seem reasonable. Some suggestions for API consistency and completeness.

packages/core/src/DropdownMenu/DropdownMenu.doc.mjs
athz: Done — `dropdown-menu-radio-dot` now mirrors the radio container: `size` visual prop plus `checked`/`disabled` states.

packages/core/src/DropdownMenu/DropdownMenuSubMenu.tsx
athz: Good call — renamed to `dropdown-menu-indicator-icon` to match `selector-indicator-icon` / `multi-selector-indicator-icon`. Kept the `chevronRight` glyph (the flyout opens to the side) rather than the selectors' `chevronDown`, and didn't carry over their expanded/collapsed `state` — the submenu chevron doesn't change appearance on open, and that state is already exposed via `aria-expanded` on the trigger row.

packages/core/src/DropdownMenu/renderDropdownItems.tsx
athz: Done — renamed to `dropdown-menu-section-heading` so it's clear the target is the heading text, not the section group.

athz:
Thanks for the review! Addressed all three in the latest push:

- `dropdown-menu-section` → `dropdown-menu-section-heading` (clarifies it targets the text)
- `dropdown-menu-caret` → `dropdown-menu-indicator-icon` (matches the selector indicator naming)
- `dropdown-menu-radio-dot` now carries `size` + `checked`/`disabled`, mirroring the radio container

Docs targets, tests, and the changeset updated to match. CI is green.

cixzhang (approved):

PR #4601, opened 2026-07-31
fix(Field): make description part of the clickable label region

### LINKED ISSUE #4132: [Bug] CheckboxInput: description isn't part of the clickable label — clicking it doesn't toggle the checkbox

HelloOjasMutreja:
## Summary

Closes #4132. On `CheckboxInput`/`Switch`, the `description` text sat outside the `<label>` as a plain sibling `<span>`, wired to the input only via `aria-describedby`. Clicking the label text toggled the control; clicking the description, which looks identical and often sits right underneath, did nothing.

## Fix

`FieldLabel` now nests `description` inside the `<label>` element for the single-control case, so clicking anywhere in it activates the input the same way clicking the label text does. `isGroupLabel` fields (e.g. a radiogroup's label) keep the description as a sibling, unchanged, since a group label has no single control to activate.

Layout: `<label>` switches from a single-line flex row to a wrapping one (`flexWrap: wrap`), with the description forced onto its own line via `flexBasis: 100%`. This avoids introducing an extra wrapper element around the label row, so the label text stays a direct child of `<label>` (no DOM-structure change for existing consumers/tests beyond the description's new position).

Verified a nested link inside the description does not double-fire: clicking it triggers the link's own handler without also toggling the control (native browser/label behavior, confirmed under jsdom in the added test).

## Scope

`FieldLabel` is shared by `Field`, `CheckboxInput`, and `Switch`. `RadioListItem` has a similar pattern but implements its own label/description slots via `Item` rather than `FieldLabel` — same underlying issue is plausible there but out of scope for this PR since it's a different component entirely; flagging for a follow-up rather than expanding this fix.

## Testing

- Added tests: description nests inside `<label>` for a single control, stays a sibling for `isGroupLabel`, clicking the description toggles a real checkbox, and a nested link's click doesn't double-toggle.
- Full `CheckboxInput`, `Switch`, `Field`, and `FieldLabel` suites pass (121 tests).
- Typecheck clean.
- Doc updates: `description` prop text on `CheckboxInput`, `Switch`, and `Field` now notes it's part of the clickable label region.

### DIFF
--- a/.changeset/checkboxinput-switch-description-text-is-now-par-y2cq.md
+++ b/.changeset/checkboxinput-switch-description-text-is-now-par-y2cq.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] CheckboxInput/Switch: description text is now part of the clickable label region, so clicking it toggles the control the same as clicking the label text
+@HelloOjasMutreja
--- a/packages/core/src/CheckboxInput/CheckboxInput.doc.mjs
+++ b/packages/core/src/CheckboxInput/CheckboxInput.doc.mjs
@@ -31,7 +31,8 @@ export const docs = {
     {
       name: 'description',
       type: 'string',
-      description: 'Description text displayed below the label.',
+      description:
+        'Description text displayed below the label. Part of the clickable label region: clicking it toggles the checkbox, same as clicking the label text.',
     },
     {
       name: 'value',
--- a/packages/core/src/Field/Field.doc.mjs
+++ b/packages/core/src/Field/Field.doc.mjs
@@ -77,7 +77,8 @@ export const docs = {
     {
       name: 'description',
       type: 'string',
-      description: 'Description text displayed between the label and input.',
+      description:
+        'Description text displayed between the label and input. For a single-control field (not isGroupLabel), it is part of the label\'s clickable region, same as the label text. Stays a plain sibling of the label for group fields (isGroupLabel), since there is no single control to activate.',
     },
     {
       name: 'descriptionID',
--- a/packages/core/src/Field/FieldLabel.test.tsx
+++ b/packages/core/src/Field/FieldLabel.test.tsx
@@ -11,6 +11,8 @@
 
 import {describe, it, expect, vi} from 'vitest';
 import {render, screen} from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import {useState} from 'react';
 import {TestIcon} from '../__tests__/TestIcon';
 import {InternationalizationProvider} from '../i18n';
 import {FieldLabel} from './FieldLabel';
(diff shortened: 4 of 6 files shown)


HelloOjasMutreja:
Closing this one. #4574 already landed the fix for this (merged 2026-08-01) with an approach that's better than what I did here: it forwards description clicks to the control via `useInputContainer` instead of nesting `description` inside `<label>`.

I hit the same problem #4574's description calls out: nesting the description inside `<label>` folds its text into the control's **accessible name** (since a `<label for>`'s name is computed from all its content), and since the description is also wired via `aria-describedby`, that means screen readers announce it twice. My tests didn't catch this since they only checked visual nesting and click behavior, not accessible-name computation. #4574 avoids it by keeping the description a sibling `<span>` and forwarding clicks programmatically instead.

Nothing left for this PR to contribute. Thanks for the reference implementation.

PR #4602, opened 2026-07-31
feat(dropdown-menu): compose CheckboxInput in DropdownMenuCheckboxItem

athz:
## What

`DropdownMenuCheckboxItem` hand-rolled its checkbox marker — a styled `<span>` with an `<Icon icon="check">` glyph resolved through the icon registry. The sibling `CheckboxListItem` instead composes the real `CheckboxInput` primitive. The two produced **different checkmark glyphs** (icon-registry check vs. `CheckboxInput`'s own inline `<svg viewBox="0 0 10 10">` tick), so a theme could not reconcile them.

This makes `DropdownMenuCheckboxItem` compose the real `CheckboxInput`, matching `CheckboxListItem`. One hand-rolled marker → one shared primitive.

## How the checkbox stays decorative

The row owns the semantics — it renders `role="menuitemcheckbox"` + `aria-checked` and receives keyboard activation from the parent menu's focus path (WAI-ARIA `menuitemcheckbox` pattern). To keep the composed control from introducing a second focusable/announced checkbox, it is wrapped in an `inert` element (with `pointer-events: none`), which removes its native `<input>` and its sr-only label from both the tab order and the accessibility tree, and lets pointer clicks fall through to the row. This is the same shim `MultiSelector` uses. State is still announced on the row; there is no double control and no double-announcement.

## Theming impact (behavior-affecting — please note)

The checkbox visual now emits `CheckboxInput`'s own `astryx-checkbox` / `astryx-checkbox-input` theming slots, **unifying menu and standalone checkbox theming** — one theme rule styles both.

Consequently the menu-specific `astryx-dropdown-menu-checkbox` slot is **no longer emitted**, and its theming target is removed from the docs. Themes that targeted `astryx-dropdown-menu-checkbox` should retarget to `astryx-checkbox` (scope with `.astryx-dropdown-menu-item .astryx-checkbox` for menu-only styling). The rendered marker size also changes from 18/22px to `CheckboxInput`'s standard 20/24px (`sm`/`md`), which is intentional — matching the real checkbox is the goal.

## Scope

Behavior is preserved: `value` → `CheckboxInput.value`, size derivation (`sm` menu → `sm`), `isDisabled`, leading `icon`, `description`, `endContent`, `hasCloseOnSelect`, the toggle/close-on-select click path, and the touch-order swap (marker moves to the inline-end on coarse pointers). Dead StyleX styles and imports removed.

## Testing

- New test asserts the composed checkbox is decorative: the row is the only `role="menuitemcheckbox"`, and the native `<input>` sits inside an `inert` subtree (not a second control).
- `@astryxdesign/core`: DropdownMenu suite (71) and CheckboxInput/CheckboxList suites (86) pass, plus `typecheck`, `typecheck:docs`, `sync-exports --check`, and lint.

### DIFF
--- a/.changeset/dropdown-checkbox-compose-input.md
+++ b/.changeset/dropdown-checkbox-compose-input.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] DropdownMenuCheckboxItem now composes CheckboxInput so its checkmark matches CheckboxListItem and the standard checkbox theming slots apply. The checkbox stays decorative — the menu row keeps role="menuitemcheckbox" and owns the checked state.
+@athz
--- a/packages/core/src/DropdownMenu/DropdownMenu.doc.mjs
+++ b/packages/core/src/DropdownMenu/DropdownMenu.doc.mjs
@@ -26,11 +26,6 @@ export const docs = {
       {className: 'astryx-dropdown-menu'},
       {className: 'astryx-dropdown-menu-item', visualProps: ['size']},
       {
-        className: 'astryx-dropdown-menu-checkbox',
-        visualProps: ['size'],
-        states: ['checked', 'disabled'],
-      },
-      {
         className: 'astryx-dropdown-menu-radio',
         visualProps: ['size'],
         states: ['checked', 'disabled'],
--- a/packages/core/src/DropdownMenu/DropdownMenuCheckboxItem.tsx
+++ b/packages/core/src/DropdownMenu/DropdownMenuCheckboxItem.tsx
@@ -4,39 +4,39 @@
 
 /**
  * @file DropdownMenuCheckboxItem.tsx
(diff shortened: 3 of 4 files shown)


cixzhang (approved):
Composing the real `CheckboxInput` unifies the checkmark glyph and theming slots with `CheckboxListItem` — nice consolidation of one hand-rolled marker into the shared primitive.

The double-control concern is handled correctly: the composed checkbox is wrapped in an `aria-hidden` + `inert` element with `pointer-events: none`, so its native `<input>` and sr-only label stay out of both the tab order and the accessibility tree while clicks fall through to the row. The row remains the single `role="menuitemcheckbox"` owning `aria-checked`, matching the established `MultiSelector` shim. No double tab target, no double announcement — WCAG 4.1.2 / 2.1.1 / 2.5.8 all satisfied. New test locks in the decorative boundary. CI (incl. pr-a11y) green.

PR #4022, opened 2026-07-16
fix: WCAG 2.1 AA contrast failures in core color tokens (#3654)

gonzoblasco:
## What

Fixes a WCAG 2.1 AA non-text contrast failure reported in #3654: `--color-border-emphasized` was below the 3:1 minimum against standard surface/body backgrounds.

## Change

`--color-border-emphasized` is an **active** border (focused inputs, active cards, emphasized states), not a disabled token. At 1.37:1 on light backgrounds it was barely perceptible - a genuine bug, failing SC 1.4.11 (Non-text Contrast, 3:1 minimum) for active UI components.

- Light: `#CCD3DB` → **`#898D93`** (3.02:1 on body, was 1.37:1)
- Dark: `#494D53` → **`#6E7276`** (3.00:1 on popover, was 1.71:1)

## Scope note (updated after review)

The original PR also proposed changes to `--color-text-disabled` and `--color-icon-disabled`. Per review feedback (@ernestt), those are exempt under WCAG SC 1.4.3 (text in inactive UI components), so both were reverted. This PR now contains only the `--color-border-emphasized` fix.

## Methodology

Values computed with the WCAG 2.1 relative luminance formula, maintaining hue while adjusting luminance - the same contrast-measurement approach the component audits (#5332) use.

## Validation

- 521/521 tests pass
- `pnpm lint`: clean
- `pnpm check:sync`: clean
- `pnpm check:package-boundaries`: clean

Fixes #3654

PR #4604, opened 2026-07-31
fix(#4450): give Selector trigger an accessible name via aria-labelledby

### LINKED ISSUE #4450: Selector's trigger has no accessible name, and its role changes with hasSearch
### Summary

`Selector`'s trigger button has no accessible name, and the markup it renders differs depending on `hasSearch` in a way that makes it impossible to target by role in either case.

Verified against `@astryxdesign/core@0.1.8` in Chromium, with `label` supplied through the component's own prop.

### What renders

Two `Selector`s on the same page, both given a `label`. Read off the trigger `<button>`:

| `hasSearch` | `role` | `aria-label` | `aria-labelledby` | text content |
| absent | `combobox` | `null` | `null` | the placeholder |
| `true` | *(none)* | `null` | `null` | the placeholder |

### Why each is a problem

**Without `hasSearch`** the trigger is `<button role="combobox">`. `combobox` is not a name-from-content role, so the placeholder text does **not** name it, and with no `aria-label` or `aria-labelledby` the accessible name computes to empty. A screen reader announces "collapsed, combobox" with no indication of which field it is. The visible `label` is rendered as a sibling element that is not associated with the control.

**With `hasSearch`** the same component renders a plain `<button>` with no role. That is name-from-content, so it does get a name — but the name is the **placeholder** ("Select a branch…"), not the `label`. Better than empty, still not the field's name, and it also loses the `combobox` role that conveys the control's behaviour.

### Expected

The value passed to `label` should name the trigger in both shapes — via `aria-labelledby` pointing at the rendered label, or `aria-label` when the label is visually hidden — and the role should not depend on whether a search box is enabled.

### Impact

On a form of several `Selector`s, every one announces identically. It also means neither `getByRole('button', { name })` nor `getByRole('combobox', { name })` matches the trigger, so automated tests and any assistive-tech user hit the same wall: consumers end up adding their own `data-*` hook to locate it, which does nothing for screen-reader users.

### Reproduction

```tsx
<Selector label="CLI" options={options} value={value} onChange={onChange} />
<Selector label="Branch" options={options} value={value} onChange={onChange} hasSearch />
```

Inspect either trigger: no `aria-label`, no `aria-labelledby`. Compare the `role` attribute between the two.

gonzoblasco:
## What

Fixes #4450 — the Selector trigger button had no accessible name in either mode:

- **Without `hasSearch`:** `role=combobox` (not name-from-content), no `aria-label` or `aria-labelledby` → empty accessible name
- **With `hasSearch`:** plain `<button>` (name-from-content), but the name fell back to the placeholder text instead of the actual `label` prop

## Changes

- Generate a `fieldLabelId` with `useId()` in the Selector
- Pass it to `<Field>` as `labelID` so the rendered label element carries a stable `id`
- Use it as `aria-labelledby` on the trigger button when standalone (no InputGroup)
- InputGroup mode preserves the existing `ariaLabelledBy` composition

## Validation

- `pnpm test -- --run packages/core/src/Selector/Selector.test.tsx`: 46/46 tests pass
- `pnpm lint`: clean
- `pnpm check:sync`: clean
- `pnpm check:package-boundaries`: clean

Fixes #4450

### DIFF
--- a/packages/core/src/Selector/Selector.tsx
+++ b/packages/core/src/Selector/Selector.tsx
@@ -702,6 +702,7 @@ export function Selector<T extends SelectorOptionType>(
   const descriptionId = useId();
   const statusMessageId = useId();
   const inputLabelId = useId();
+  const fieldLabelId = useId();
   const searchId = useId();
   const triggerRef = useRef<HTMLButtonElement>(null);
   const searchRef = useRef<HTMLInputElement>(null);
@@ -748,6 +749,12 @@ export function Selector<T extends SelectorOptionType>(
     ],
     inputGroup,
   );
+
+  // When standalone (no InputGroup), the trigger needs an explicit
+  // aria-labelledby pointing at the Field label so the accessible name
+  // resolves to the label text. The combobox role is not name-from-content,
+  // so the visible placeholder text does not name it.
+  const triggerLabelledBy = inputGroup ? ariaLabelledBy : fieldLabelId;
 
   // Flatten options for keyboard navigation
   const selectableItems = useMemo(
@@ -1208,7 +1215,7 @@ export function Selector<T extends SelectorOptionType>(
               : undefined
           }
           aria-describedby={ariaDescribedBy}
-          aria-labelledby={ariaLabelledBy}
+          aria-labelledby={triggerLabelledBy}
           aria-required={isRequired ? 'true' : undefined}
           aria-invalid={status?.type === 'error' ? 'true' : undefined}
           aria-busy={isBusy || undefined}
@@ -1351,6 +1358,7 @@ export function Selector<T extends SelectorOptionType>(
       isLabelHidden={isLabelHidden}
       description={description}
       inputID={triggerId}
+      labelID={fieldLabelId}
       descriptionID={description ? descriptionId : undefined}
       isOptional={isOptional}
       isRequired={isRequired}


imdreamrunner (changes requested):
(Sorry the original comment was posted accidentally by AI, now updated with verified feedbacks)

Thanks for the fix. Two things to address:
- This branch also includes the unrelated contrast-token changes and audit scripts from #4022. Please rebase or cherry-pick the Selector commit onto main.
- Before this patch, the native <label for> association already gives the trigger the accessible name Fruit in the repository’s accessibility tests, across both hasSearch modes. Could you document the specific compatibility case where explicit aria-labelledby is needed, and add a regression if the tooling can represent it?

gonzoblasco:
Thanks for the thorough review, @imdreamrunner. You're right on both counts.

The branch is now clean - I rebased the Selector commit onto current main, so #4604 only contains the single aria-labelledby change to Selector.tsx (the contrast work from #4022 is no longer in this branch).

On the accessible name finding: you're correct that the change is redundant. The evidence came from automated static audit (axe), not from a real browser/AT failure, and your verification confirms the native `<label for>` + `<button id>` association already provides the accessible name and the suite passes 70/70 at the parent commit. Since there's no concrete failing case to justify explicit ARIA, I'm closing this as already satisfied rather than keeping redundant wiring in the codebase.

Thanks for the detailed review.

Issue #3654, opened 2026-07-07
[Bug] Widespread WCAG 2.1 AA colour contrast failures across Design System, accessibility


chris-gds:
### Description

Severity: High / Blocker (Prevents accessible product releases)
Component(s) Affected: Core Colour Palette, Buttons, Typography, Form Fields, Navigation
Specification Targeted: WCAG 2.1 Compliance Level AA

## Problem description

A review of the new design system component specs reveals widespread colour contrast failures under WCAG 2.1 Level AA standards. Multiple core colour pairings (text-on-background and UI components) fall below the required contrast ratios, impacting readability for disabled users including those with low vision, colour blindness, or those viewing screens in high-ambient light environments. 

## Impact

Products built with this design system will fail standard accessibility audits, creating legal liability under the ADA, European Accessibility Act (EAA), and other regional mandates.

### Audit & darken/lighten core tokens

- Adjust the luminosity of the core brand palette shades to ensure compliant pairings 
- Automate Contrast Checks: integrate a linter or CI/CD check to flag non-compliant colour pairings before they are pushed to the codebase.
- Hire an accessibility engineer ;)

#### Blockquote issue 

_"Design is not just what it looks like. Design is how it works"_
This Design does not work for disabled users.

The text on the background fails 

<img width="964" height="757" alt="A screenshot showing a blockquote design section with a Colour Contrast Analyser window superimposed on top. The analyzer tests a medium-dark grey foreground (#737373) against a light gray background (#F1F1F1), yielding a contrast ratio of 4.2:1, which passes WCAG AA for large text but fails for regular text." src="https://github.com/user-attachments/assets/b9d1590f-9ed2-496d-b3c6-2d8b173d58f2" />

#### Avatar Group

The foreground user initials fails on the bg

<img width="964" height="757" alt="A screenshot of a user interface showing circular team member avatars, overlaid with a Colour Contrast Analyser window. The analyzer tests a foreground color of #737373 against a background of #F0F0F0, resulting in a 4.2:1 contrast ratio that passes AA for large text but fails for regular text." src="https://github.com/user-attachments/assets/4e763868-bcbb-48d3-9d75-ad24a2afd64c" />

#### Carousel 

the text on the bg fails

[...]

PR #4605, opened 2026-07-31
fix(#4421): add non-color indicator to current breadcrumb item

### LINKED ISSUE #4421: [Bug] WCAG 2.1 A 1.4.1 fail - use of colour failure, accessibility - Breadcrumb Item

gonzoblasco:
## What

Fixes #4421 — the current page breadcrumb item relied solely on color (`--color-text-primary` vs `--color-text-secondary`) to distinguish it from links, failing WCAG 2.1 SC 1.4.1 (Use of Color).

## Changes

Added `--font-weight-semibold` (600) to the current item's style so it is visually distinguishable by weight in addition to color. The `aria-current="page"` attribute was already present for screen readers.

## Validation

- `pnpm test -- --run packages/core/src/Breadcrumbs/Breadcrumbs.test.tsx`: 25/25 tests pass
- `pnpm lint`: clean
- `pnpm check:sync`: clean
- `pnpm check:package-boundaries`: clean

Fixes #4421

### DIFF
--- a/packages/core/src/Breadcrumbs/BreadcrumbItem.tsx
+++ b/packages/core/src/Breadcrumbs/BreadcrumbItem.tsx
@@ -38,6 +38,7 @@ import {
   spacingVars,
   typeScaleVars,
   radiusVars,
+  fontWeightVars,
 } from '../theme/tokens.stylex';
 import {BreadcrumbContext, type BreadcrumbsVariant} from './Breadcrumbs';
 import {useLinkComponent} from '../Link/useLinkComponent';
@@ -186,7 +187,7 @@ const itemStyles = stylex.create({
     color: colorVars['--color-text-secondary'],
   },
   current: {
-    fontWeight: 'inherit',
+    fontWeight: fontWeightVars['--font-weight-semibold'],
   },
   defaultCurrent: {
     color: colorVars['--color-text-primary'],


gonzoblasco:
**Rebased + scope cleaned.** This PR now contains only the fix for #4421.

**Context on the scope change:** the branch previously carried commits from the color-contrast work (#3654, which lives in its own open PR #4022). That was my mistake - unrelated changes leaked into this PR. I've rebased `fix/breadcrumb-use-of-color` onto current `main` so it now contains exactly one change: the non-color current-page indicator on `BreadcrumbItem`.

**On the fix itself:** the Breadcrumbs audit (#5332) identified the same issue as open finding B8 and explicitly deferred it to this PR (and to #4421), rather than landing a colour-only change that would conflict with this fix:

> "Making the supporting variant's current crumb `--color-text-primary` distinguishes the current item by colour alone. That is precisely #4421 ... and #4605 by @gonzoblasco has been open since 31 July fixing it properly with a non-colour indicator (semibold weight on the shared `current` style, which covers both variants) on the same style object."

The fix lands `font-weight: var(--font-weight-semibold)` on the shared `current` style, which covers both the default and supporting variants, in addition to the existing `aria-current="page"`. I rebased onto the post-audit `main` and re-verified the merge is clean.

- `pnpm test -- --run packages/core/src/Breadcrumbs/Breadcrumbs.test.tsx`: 25/25 pass
- `pnpm lint`: clean

Ready for review. Thanks.

cixzhang (approved):
Thanks, this is the fix the Breadcrumbs audit pointed at — the supporting variant now reads as current without colour. Nit: semibold runs ~4px wider, so long labels wrap sooner.

[Reviewed by Robohands]

PR #4608, opened 2026-07-31
ci(scripts): add doc-drift — a deterministic documentation staleness gate

josephfarina:
## What this is

One deterministic gate for documentation staleness. Docs (`*.doc.mjs`) are what agents and consumers actually read, but nothing checks they stay consistent with the component source — so they drift silently. A prop gets added to a component and never to its doc; a required prop is documented as optional; an agent then reads the stale doc and gets the wrong contract.

`scripts/doc-drift.mjs` is the single drift command. Each drift **class** is a pluggable check that compares an artifact against its source of truth deterministically (no AI, structure only). The first class — **props** — is implemented; theme-registry and theme-token-completeness are scoped as follow-ups (see the README).

## The props check

- Builds **one** TS program over core source and reuses the type checker, so `interface extends`, `type` unions, and `Omit`/`Pick` all resolve. (~5–7s; a program-per-file was ~52s.)
- Extracts only **component-declared** props — members whose declaration isn't in `BaseProps.ts` / React / the DOM lib. Inherited HTML/`aria-*`/`data-*` attributes and raw DOM `on*` handlers are excluded, since docs don't (and shouldn't) enumerate the shared platform surface.
- Compares against the doc's `props[]`: flags `missing-from-doc`, `required-mismatch`, and (lower-confidence) `phantom-in-doc`.
- When a `{Name}Props` symbol exists in more than one file, the one under the doc's own directory wins.

It checks **structure only** — never the human-authored prose. So docs can still be regenerated freely (by an author or an AI); the gate just blocks a doc whose *skeleton* disagrees with source. That's the "AI can update the docs, but a prop mismatch blocks it" contract.

## Rollout is baselined

There's pre-existing drift (**101 items** at introduction — the check surfaced real bugs like `Badge.label`, `DialogHeader.title`, and `Toast.type` all documented optional but required in source, and undocumented props like `CheckboxList.isReadOnly`). Turning the gate straight to red would block every PR, so:

1. Land with `doc-drift.baseline.json` snapshotting today's known drift. CI is green; only **new** drift fails.
2. Burn the baseline down — regenerate the drifted docs (ideal AI task), then `--update-baseline` to shrink it.
3. At zero, delete the baseline — the gate is fully strict.

Verified: with the baseline, `--check` is green; inject one new undocumented prop and it exits 1 with a precise message (and catches the transitive case — `IconButton extends ButtonProps`).

## Wiring

Added to the required `test` job next to the other `--check` gates (`sync-exports`, `generate-token-docs`). `typescript` is already a root devDep, so no new setup.

## Usage

```
node scripts/doc-drift.mjs            # human report
node scripts/doc-drift.mjs --check    # exit 1 on new drift (CI)
node scripts/doc-drift.mjs --json     # machine-readable
node scripts/doc-drift.mjs --update-baseline
```

## Follow-ups (in the README)

- Fold the existing `build-theme.registry.test.mjs` invariants in as a `theme-registry` class so there's one drift command.
- Add a `theme-completeness` class: `defineTheme` takes `tokens?: Partial<…>`, so a theme can silently omit a newly-added core token and `tsc` won't catch it.
- Burn down the 101-item baseline with AI-assisted doc regen.


josephfarina:
Closing. This adds a new *required* CI gate with a 101-item baseline, and that is a repo policy call for the XDS team to make rather than something to land as a surprise PR. It has also sat five weeks with three reviewers requested and no response.

The check found real bugs on its first run -- `Badge.label`, `DialogHeader.title`, `Toast.type` were all documented wrong -- so I'll bring doc-drift back as a proposal first.

PR #4624, opened 2026-08-01
test(calendar): pin clock in day-cell marker tests to fix month-boundary flake

freddymeta:
## What

Pin the system clock in the Calendar `day-cell marker theme state` tests so they no longer flake by calendar date.

## Why

Those tests build selection ranges around the real "today" (`today ± N` days). The Calendar derives today from `new Date()` and, for a range, renders the range's **start** month. When the suite runs near a month boundary, an endpoint — or today itself — lands in an adjacent month that isn't rendered, so the today cell can't be found and the `today-in-range` marker assertions fail.

This makes the whole `test` job fail on any day close to a month boundary (e.g. the 1st), regardless of the change under test.

## Fix

Add a scoped `beforeEach`/`afterEach` to that `describe` block that pins a fixed mid-month instant with `vi.setSystemTime`. Only `Date` is faked (`toFake: ['Date']`), so React/RTL scheduling is untouched. The `today ± N` window now always stays inside one rendered month.

Test-only; no runtime/behavior change, so no changeset. All 73 Calendar tests pass.

### DIFF
--- a/packages/core/src/Calendar/Calendar.test.tsx
+++ b/packages/core/src/Calendar/Calendar.test.tsx
@@ -9,7 +9,7 @@
  * SYNC: When Calendar.tsx changes, update tests accordingly
  */
 
-import {describe, it, expect, vi, afterEach} from 'vitest';
+import {describe, it, expect, vi, beforeEach, afterEach} from 'vitest';
 import {act, render, screen, within, waitFor} from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import {getButton} from '../__tests__/fastRoleQueries';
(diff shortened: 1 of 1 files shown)


cixzhang (approved):

freddymeta:
Closing as obsolete — this flake is already fixed on `main`.

The `day-cell marker theme state` tests now pin a fixed mid-month instant (`vi.useFakeTimers({toFake: ['Date']})` + `vi.setSystemTime(...)`) so the `today ± N` range stays inside a single rendered month. The only difference between this branch and `main` is comment wording, so there's nothing substantive left to land.

Issue #4620, opened 2026-08-01
[Bug] `astryx theme build` copies an extensionless `./icons` specifier into its ESM output — all 7 published themes' `/built` entries fail to load in Node


imdreamrunner:
### Description

`astryx theme build` scrapes the icon-registry import out of the theme's TypeScript source and writes it verbatim into the JavaScript it generates. `./icons` is valid TypeScript and invalid ESM, so the generated module cannot be loaded by Node — even though the file it names does exist in the published tarball.

```
$ npm i @astryxdesign/theme-neutral@0.2.0
$ node --input-type=module -e "import '@astryxdesign/theme-neutral/built'"
Error [ERR_MODULE_NOT_FOUND]: Cannot find module
  '.../node_modules/@astryxdesign/theme-neutral/dist/icons'
  imported from '.../dist/neutral.js'

$ node --input-type=commonjs -e "require('@astryxdesign/theme-neutral/built')"
    same ERR_MODULE_NOT_FOUND

$ node --input-type=module -e "import '@astryxdesign/theme-neutral'"
    OK      # the bare '.' entry is fine
```

`dist/icons.mjs` and `dist/icons.js` are both present beside `dist/neutral.js`. Only the spelling is wrong.

Per <https://nodejs.org/api/esm.html#mandatory-file-extensions>:

> A file extension must be provided when using the import keyword to resolve relative or absolute specifiers.

**Expected:** the module `astryx theme build` writes should be loadable by Node.

### Root cause

`extractIconInfo()` keeps the raw capture from the source's import statement; `generateBuiltModule()` interpolates it straight into a `.js` artifact.

- `packages/cli/api/theme/build/build.mjs#L504-L524` — `importPath: importMatch[1]`
- `packages/cli/api/theme/build/build.mjs#L534-L537` — `` `import { ${iconInfo.exportName} } from '${iconInfo.importPath}';\n` ``

(Line 532 in the published `@astryxdesign/cli@0.2.0` tarball at `api/theme/build/build.mjs`.)

The sibling `generateBuiltTypes()` declares the registry locally rather than re-importing it, so the `.d.ts` has no relative specifier at all and TypeScript consumers see a perfectly well-typed module. Only the runtime JS is broken, which would explain how this shipped.

[...]

PR #4625, opened 2026-08-01
fix(themes,cli): gate theme dists with check-fully-specified and pin --icons-specifier coverage

AKnassa:
**Rescoped after #4631, per review.** The resolve-at-build approach this PR originally carried is superseded by `--icons-specifier`. What remains is the two pieces requested: the post-build verification gate, and the regression coverage adapted to the flag.

**Gate.** All seven maintained theme packages now end their build with `check-fully-specified.mjs`, the same dist gate that core, lab, charts and richtext already run. A mis-specified icon import now fails the theme build instead of shipping. Verified both ways on `theme-neutral`: the real dist passes, and the same dist with the import corrupted back to `./icons` exits 1.

**Coverage.** Three paths #4631's tests don't reach, each verified to fail when its forwarding is removed:

- direct API: `themeBuild({iconsSpecifier})` reaches the emitted module, and `check` mode treats a mismatched specifier as stale
- a spawned-Node proof that the emitted module actually loads (the distinction #4620 shipped: every byte existed, none loaded)
- watch mode forwards the flag to every child rebuild

The theme changeset is retained so the seven packages republish with the corrected import; the published tarballs still carry the broken string.

Refs #4620, #4631.


cixzhang (approved):
Looks good from the implementation side — the resolver/checking approach makes sense and the focused theme smoke passes.

Before merge, could you please clean up two things:

- Rebase/resolve the current merge conflicts against `main`.
- Trim the changesets down to changelog-sized summaries. They currently read more like PR-body/root-cause notes; the changelog only needs the consumer-visible effect.

Thanks for the thorough test coverage here.

imdreamrunner (changes requested):
Thanks for the careful work here. The resolver, checks, and focused tests helped clarify this failure mode and provide useful coverage.

#4631 treats theme generation and icon compilation as separate steps. Their order and output filename depend on each theme package’s build setup. The build pipeline owns that configuration, so it explicitly declares the final icon path with `--icons-specifier`. Post-build verification then validates the completed artifacts.

We especially want to retain two pieces from this PR:

1. Run `check-fully-specified.mjs` after each maintained-theme build.
2. Retain/adapt the broader regression coverage for the direct API, watch-mode forwarding, and a real Node import smoke test.

Together, the option declares the intended specifier and the checks prove the published artifacts satisfy it.

AKnassa:
Done, rescoped as requested. The resolver is dropped in favor of `--icons-specifier`, and the two pieces you named remain: `check-fully-specified.mjs` now gates all seven maintained-theme builds, and the coverage is rewritten against the flag (direct API, watch-mode forwarding, spawned-Node import smoke test). I kept the theme packages' patch changeset since a republish is what carries the fix to users; happy to drop it if release is handled separately.

imdreamrunner (approved):

imdreamrunner:
Thanks for the PR!

PR #4626, opened 2026-08-01
feat(selector): add disabled-trigger and caret-wrapper theme targets

athz:
## What

Adds two additive theme targets to the searchable **Selector** / **MultiSelector** so consumers can style them via `defineTheme` instead of structural CSS:

| Target | What it reaches |
|---|---|
| `data-state="disabled"` on `selector` / `multi-selector` | the trigger's disabled state (previously only the inner button carried it) |
| `selector-indicator` / `multi-selector-indicator` (new) | the chevron **wrapper** span, with an `expanded`/`collapsed` state |

## Why

The trigger's `themeProps` exposed only `{size, status}`, so a theme couldn't key on the disabled state — the disabled flag lives on the inner button, which no slot reaches.

The caret's `rotate(180deg)` (open state) is applied to the chevron **wrapper span**, while the only existing slot (`selector-indicator-icon`) is on the **child glyph**. A child element cannot cancel a transform applied to its parent, so a theme that wants a static caret (no open-state rotation) had no reachable hook. The new `selector-indicator` target lands on the wrapper span itself, so a theme can do:

```css
.astryx-selector-indicator.expanded { transform: none; }
```

## Scope

Purely additive `themeProps` on existing elements. No new props, no exported-type changes, no behavior or a11y change — default appearance is identical (guarded by existing byte-identity tests plus new negative tests: no `data-state` when enabled).

## Testing

- Selector + MultiSelector suites pass, including new tests asserting the disabled-state target, the indicator-wrapper target, its expanded/collapsed state, and that `defineTheme` can override the caret rotation via the wrapper slot.
- `themingTargets` guard passes (every rendered `themeProps` class is documented in `.doc.mjs`).
- typecheck, typecheck:docs, sync-exports --check, check-changesets, and ESLint on changed files — all green.

### DIFF
--- a/packages/core/src/MultiSelector/MultiSelector.doc.mjs
+++ b/packages/core/src/MultiSelector/MultiSelector.doc.mjs
@@ -23,6 +23,7 @@ export const docs = {
       {
         className: 'astryx-multi-selector',
         visualProps: ['variant', 'size', 'status'],
+        states: ['disabled'],
       },
       {className: 'astryx-multi-selector-clear-icon'},
       {
--- a/packages/core/src/Selector/Selector.doc.mjs
+++ b/packages/core/src/Selector/Selector.doc.mjs
@@ -24,6 +24,7 @@ export const docs = {
       {
         className: 'astryx-selector',
         visualProps: ['variant', 'size', 'status'],
+        states: ['disabled'],
       },
       {className: 'astryx-selector-option'},
       {className: 'astryx-selector-clear-icon'},
--- a/.changeset/selector-disabled-state-target.md
+++ b/.changeset/selector-disabled-state-target.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] Selector/MultiSelector: add `data-disabled` reflection on the trigger for theme-driven disabled styling. Rotation styles remain on the indicator-icon target (no separate wrapper target needed).
+@athz
--- a/packages/core/src/MultiSelector/MultiSelector.test.tsx
+++ b/packages/core/src/MultiSelector/MultiSelector.test.tsx
@@ -1871,3 +1871,58 @@ describe('MultiSelector search affordances', () => {
     expect(trigger).toHaveAttribute('aria-expanded', 'false');
   });
 });
+
+describe('MultiSelector disabled state theme target', () => {
+  const getSelectorRoot = (container: HTMLElement): HTMLElement => {
+    const root = container.querySelector('.astryx-multi-selector');
+    if (root == null) {
+      throw new Error('multi-selector root not found');
+    }
+    return root as HTMLElement;
+  };
+
+  it('reflects data-state="disabled" on the root when disabled', () => {
+    const {container} = render(
+      <MultiSelector
+        label="Fruit"
+        options={['Apple', 'Banana', 'Orange']}
+        value={[]}
+        onChange={() => {}}
+        isDisabled
+      />,
+    );
+    expect(getSelectorRoot(container)).toHaveAttribute(
+      'data-disabled',
+      'disabled',
+    );
+  });
+
+  it('omits the disabled class/attribute when enabled', () => {
+    const {container} = render(
+      <MultiSelector
+        label="Fruit"
+        options={['Apple', 'Banana', 'Orange']}
+        value={[]}
+        onChange={() => {}}
+      />,
+    );
+    const root = getSelectorRoot(container);
+    expect(root).not.toHaveAttribute('data-disabled');
+    expect(root).not.toHaveClass('disabled');
+  });
+
+  it('exposes the disabled state so a theme can key on it', () => {
+    const theme = defineTheme({
+      name: 'multi-selector-disabled-state-test',
+      components: {
+        'multi-selector': {
+          'disabled:disabled': {opacity: '0.4'},
+        },
+      },
+    });
+    const css = generateThemeTestCSS(theme);
+    expect(css).toContain('.astryx-multi-selector.disabled');
+    expect(css).toContain('opacity: 0.4');
+  });
+});
+
--- a/packages/core/src/MultiSelector/MultiSelector.tsx
+++ b/packages/core/src/MultiSelector/MultiSelector.tsx
@@ -1445,6 +1445,7 @@ export function MultiSelector<T extends MultiSelectorOptionType>({
             variant,
             size,
             status: status?.type ?? null,
+            disabled: isDisabled ? 'disabled' : null,
           }),
           stylex.props(
             inputWrapperStyles.base,
--- a/packages/core/src/Selector/Selector.test.tsx
+++ b/packages/core/src/Selector/Selector.test.tsx
@@ -1717,3 +1717,48 @@ describe('Selector selected-marker theme target (selector-check)', () => {
     expect(css).toContain('display: none');
   });
 });
+
+describe('Selector disabled state theme target', () => {
+  const getSelectorRoot = (container: HTMLElement): HTMLElement => {
+    const root = container.querySelector('.astryx-selector');
+    if (root == null) {
+      throw new Error('selector root not found');
+    }
+    return root as HTMLElement;
+  };
+
+  it('reflects data-disabled="disabled" on the root when disabled', () => {
+    const {container} = render(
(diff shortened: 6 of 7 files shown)


cixzhang (changes requested):
The theme targets themselves look good and additive, and the caret wrapper (`selector-indicator` / `multi-selector-indicator`) with `state: popover.isOpen ? 'expanded' : 'collapsed'` matches the existing convention exactly (same as DateInput / DateRangeInput / MultiSelector). Nice.

One convention fix before this lands: the disabled reflection uses the wrong key.

```tsx
themeProps('selector', {
  size,
  status: status?.type ?? null,
  state: isDisabled ? 'disabled' : null,   // ← emits data-state="disabled"
})
```

Across the codebase `state:` is reserved for the expanded/collapsed (open) meaning, and **disabled is reflected under a `disabled:` key** — e.g. `Switch`, `SegmentedControlItem`, `TreeListItem`, `DropdownMenuRadioItem`, `CheckboxInput`, `RadioListItem`, `Slider` all use:

```tsx
disabled: isDisabled ? 'disabled' : null,   // → data-disabled="disabled"
```

Using `state` for disabled here overloads `data-state` with two different meanings and diverges from every other component. Please switch the root trigger to `disabled: isDisabled ? 'disabled' : null` (→ `data-disabled`) in both `Selector.tsx` and `MultiSelector.tsx`, update the `.doc.mjs` target docs + the `themingTargets` guard accordingly, and keep `state` solely for the caret's expanded/collapsed. That keeps the disabled hook consistent system-wide.

cixzhang:
Note: don't hand-author a `data-disabled` attribute — just change the key inside the existing `themeProps({ disabled: isDisabled ? 'disabled' : null })` call, which emits both the class token and the `data-*` reflection.

athz:
Good catch — fixed. Switched both `Selector` and `MultiSelector` to `disabled: isDisabled ? 'disabled' : null` (→ `data-disabled`), matching Switch / CheckboxInput / RadioListItem / DropdownMenu* / Slider. `data-state` now stays reserved for the caret's expanded/collapsed only. Updated the `.doc.mjs` targets (`states: ['disabled']` on the base trigger, `state` on the caret wrapper) and the tests. All green locally.

cixzhang (commented):
The disable state looks good now. I see we have both indicator and indicator icon. The indicator target will affect other icons that can appear in the same target area while the indicator icon is only the Carey icon URLs. Curious if this is your expectations around what is getting targeted

cixzhang (approved):
This is good just needs a rebase.

cixzhang (commented):
🟡 **Maintainer judgment** — approval stands; holding the merge on a rebase plus one open question about `selector-indicator`.

The disabled reflection is exactly right now — `disabled: isDisabled ? 'disabled' : null` on the existing `selector` / `multi-selector` target matches Switch, CheckboxInput, RadioListItem, Slider, SegmentedControlItem, TreeListItem and Calendar verbatim, and `data-state` stays reserved for the caret. Tests are real (positive, negative, and a `generateThemeTestCSS` reachability assertion per seam). Thanks for turning that around quickly.

**Rebase (required, trivial).** The branch is 11 commits behind and #4627 landed today, giving exactly one conflict: `Selector.test.tsx`. Both sides are pure appends at end-of-file, so the resolution is "keep both `describe` blocks" — I verified locally that both suites (179 tests) and the `themingTargets` guard pass after that. `Selector.tsx` and `Selector.doc.mjs` auto-merge.

**Coming back to my 08-03 question, with a receipt.** `astryx-selector-indicator` isn't the chevron wrapper — it's the trailing-affordance slot. Rendering `<Selector status={{type: 'error'}} />`:

    <span class="astryx-selector-indicator collapsed …" data-state="collapsed">
      <span class="astryx-icon sm error" …><svg …/></span>
    </span>

No caret in it, no `astryx-selector-indicator-icon` child — and on open it still flips to `expanded` even though `triggerIconOpen` is gated on `!showStatusIcon`, so nothing rotates. A theme rule on this target also lands on the status icon and the status-tooltip button, and `data-state` doesn't describe what's rendered. The PR body and changeset both call it "the chevron wrapper," which is what a consumer will believe.

**A design direction I'm working through, and I'd like your read on it.** I'm converging on: *a theming target is a paint seam — color, background, border, font, radius, shadow. Layout (display, position, flex/grid, spacing, size, transform) is the component's structural contract, and it's exactly what a DOM refactor is entitled to change.* Layout can still be themeable, but through a declared var rather than a raw class target on an internal element — which is what the system already does with `--tree-list-indent`, `--button-icon-only-aspect` and `--button-focus-offset`.

`selector-indicator` sits right on that line. `styles.triggerIcon` has one paint property (`color`) and the glyph already reaches that via `selector-indicator-icon`; the target's entire stated purpose is cancelling a `transform`. So it and my 08-03 question are really the same question.

Options, and I don't think it's obvious which is right:

1. **Move the rotation onto the glyph.** `Icon` takes `xstyle` and composes it through `stylex.props`, and the glyph already carries `selector-indicator-icon` with the same `expanded`/`collapsed` values — so the existing paint target would reach it and the new one isn't needed. It'd also converge `ComplexSelector`, which has the identical wrapper+rotation structure and no wrapper target. Would want a before/after screenshot to confirm the rotation reads identically.
2. **A public var** — `--selector-indicator-rotation`, documented in `theming.vars[]`, same shape as `--button-icon-only-aspect`. A bounded knob instead of an open layout seam.
3. **Decide it once in the design.** Selector, MultiSelector and ComplexSelector all rotate today — "does the caret rotate on open" may just be a system answer rather than a theme knob.

I lean (1), but you've been closer to this than I have — what do you think? Since the branch has to be re-pushed for the rebase anyway, I'd rather settle it now than ship a target we can't easily remove.

**If the wrapper target does stay,** it needs a `description` on both entries in `Selector.doc.mjs` / `MultiSelector.doc.mjs` — right now `astryx-selector-indicator` and `astryx-selector-indicator-icon` sit next to each other with identical `states: ['state']` and no prose, and a theme author has no way to tell which is the slot and which is the caret. The changeset and PR body should stop calling it "the chevron wrapper" too.

Small stuff:

- `MultiSelector.test.tsx:1884` — the test name says `data-state="disabled"` but asserts `data-disabled`. Leftover from the rename.
- MultiSelector has no "reflects `expanded` when open" test; Selector does.
- `'disabled:disabled'` (`Selector.test.tsx:1677`, `MultiSelector.test.tsx:1919`) works, but since `.doc.mjs` declares it under `states:` the canonical key is the bare `disabled: { … }`. `'state:expanded'` is correct as-is.

packages/core/src/MultiSelector/MultiSelector.doc.mjs
```diff
@@ -23,8 +23,10 @@ export const docs = {
       {
         className: 'astryx-multi-selector',
         visualProps: ['variant', 'size', 'status'],
+        states: ['disabled'],
       },
       {className: 'astryx-multi-selector-clear-icon'},
+      {className: 'astryx-multi-selector-indicator', states: ['state']},
```
cixzhang: Instead of a new target on an internal wrapper, could we try moving the rotation styles inside of selector onto the icon so it can be themeable from the existing target?

packages/core/src/Selector/Selector.doc.mjs
```diff
@@ -24,9 +24,11 @@ export const docs = {
       {
         className: 'astryx-selector',
         visualProps: ['variant', 'size', 'status'],
+        states: ['disabled'],
       },
       {className: 'astryx-selector-option'},
       {className: 'astryx-selector-clear-icon'},
+      {className: 'astryx-selector-indicator', states: ['state']},
```
cixzhang: Same here

cixzhang (changes requested):

packages/core/src/MultiSelector/MultiSelector.doc.mjs
athz: Done — removed the wrapper target entirely. The rotation styles stay on the span (they're internal structure), and the existing `selector-indicator-icon` target on the Icon element is sufficient for themes to override the glyph's color/size/state. This PR now only adds `data-disabled` reflection on the trigger.

cixzhang (approved):

PR #4627, opened 2026-08-01
feat(selector): expose selector-check theme target on the selected marker

athz:
## What

Adds a `selector-check` theme target on the **Selector**'s selected-row checkmark, so a theme can restyle or hide it via `defineTheme` instead of a structural sibling selector.

```tsx
{isSelected && (
  <Icon icon="check" size="sm" color="accent" {...themeProps('selector-check')} />
)}
```

## Why

Today the selected-row marker is a bare `<Icon>` rendered as a sibling of the option content, with no theme target. A theme that wants a different selected indicator — e.g. composing a radio via `renderOption` and hiding the built-in check — has to reach the checkmark with a brittle structural selector (`option > .astryx-icon:last-child` and the like), which breaks on any DOM change.

With this target, hiding it is a one-liner:

```css
.astryx-selector-check { display: none; }
```

and the theme composes its own indicator through the existing `renderOption` hook (styled by whatever primitive it uses). The marker's existence, position, and a11y stay the component's job; only the glyph's styling/visibility becomes themeable — no new prop, no enum, no pseudo-element content injection.

## Scope

Purely additive `themeProps` on the existing checkmark element. No new props, no exported-type changes, no behavior or a11y change — the checkmark renders identically until a theme targets it (guarded by a byte-identity test).

## Testing

- Selector suite passes, including new tests: the target renders on the selected row only (absent on unselected rows), the glyph is byte-identical to a standalone `<Icon icon="check" size="sm" color="accent">` aside from the target class, and `defineTheme` can hide/restyle it via `selector-check`.
- `themingTargets` guard, typecheck, typecheck:docs, sync-exports --check, eslint, and check-changesets — all green.

### DIFF
--- a/.changeset/selector-check-target.md
+++ b/.changeset/selector-check-target.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] Selector: expose a `selector-check` theme target on the selected-row checkmark so themes can restyle or hide it (e.g. to compose their own selected indicator via `renderOption`) instead of relying on a structural sibling selector. Purely additive — default appearance is unchanged.
+@athz
--- a/packages/core/src/Selector/Selector.doc.mjs
+++ b/packages/core/src/Selector/Selector.doc.mjs
@@ -28,6 +28,7 @@ export const docs = {
       {className: 'astryx-selector-option'},
       {className: 'astryx-selector-clear-icon'},
       {className: 'astryx-selector-indicator-icon', states: ['state']},
+      {className: 'astryx-selector-check'},
     ],
   },
   description: 'Dropdown selector for choosing from a list of options.',
--- a/packages/core/src/Selector/Selector.test.tsx
+++ b/packages/core/src/Selector/Selector.test.tsx
@@ -1640,3 +1640,80 @@ describe('Selector search affordances', () => {
     expect(trigger).toHaveAttribute('aria-expanded', 'true');
   });
 });
+
+describe('Selector selected-marker theme target (selector-check)', () => {
+  const openOptions = (): HTMLElement[] => screen.getAllByRole('option', h);
+
+  it('renders the astryx-selector-check target on the selected row only', () => {
+    render(
+      <Selector
+        label="Fruit"
+        options={OPTIONS}
+        value="Banana"
+        onChange={() => {}}
+        isDefaultOpen
+      />,
+    );
+    const options = openOptions();
+    const selected = options.find(
+      o => o.getAttribute('aria-selected') === 'true',
+    )!;
+    const check = selected.querySelector('.astryx-selector-check');
+    expect(check).toBeInTheDocument();
+    // The target lands on the checkmark glyph itself, so a theme can restyle or
+    // hide it (e.g. to compose its own selected indicator via renderOption).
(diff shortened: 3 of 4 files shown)


cixzhang (changes requested):
Thanks for this — it surfaces a design question we want to work through before adding the prop, so let me share where our heads are rather than just accept/reject.

The hesitation with `selectionIndicator: 'check' | 'radio'` is that a closed enum freezes two arbitrary points on what's actually a broader space. The things people tend to want from a selection indicator are really: the glyph (check / radio / dot / a product-specific icon), its position (trailing vs. leading), and sometimes whether it shows at all — and a two-value enum can express none of those cleanly. The next request is inevitably a third value or a second prop, and we've grown API surface for what's largely a styling choice.

But the alternatives each have a catch, which is why we want to think it through:
- **ReactNode slots** (`selectedIndicator`/`unselectedIndicator`) give the best DX, but hand consumers full content control inside a listbox option (API surface + a11y risk if someone injects something interactive), and the design system loses hold of the visual language.
- **Pure theme target** feels "correct" (this *is* theming), but swapping check→radio through a target is content-injection via `::before`/`content` — fragile, and not really what theming should be.

Where we're leaning: the marker's existence, position, and a11y semantics stay the component's job (fixed), and the glyph becomes themeable through the icon registry — i.e. the marker renders as an `Icon` resolving a registered `selector-marker` name, so a theme swaps the registered icon (the same way the rest of the system themes glyphs) rather than passing a prop or hacking pseudo-elements. That gives customization without an enum, a content slot, or `::before`.

One more consideration specific to check vs. radio: a radio glyph *connotes* mutually-exclusive single-select semantics. In this component it's cosmetically equivalent to the check, which is part of why encoding it as a first-class prop feels off — it dresses the control as something radio-semantic without adding behavior.

None of this is a hard no on the capability — it's that we'd rather land it as a registry-themeable glyph (+ maybe a deliberate position variant if design wants one) than as a `check | radio` prop. The decorative-marker a11y handling you did here (aria-hidden, row stays the control, `CanvasText` under forced-colors) is exactly right and should carry into whichever route we pick. Keen to hear if the registry-glyph approach covers your use case.

athz:
Good call — reworked to drop the enum. Agreed that `check | radio` freezes two arbitrary points, and that a radio glyph connotes mutually-exclusive semantics this control doesn't actually add.

On the icon-registry route: it's clean for glyph-shaped markers, but a radio isn't really a glyph — it's a ring + fill with two states (`RadioListItem` draws it in StyleX, not as an icon), so registering a single `selector-marker` icon can't express checked-vs-unchecked cleanly, and there's no radio icon in the registry today.

So instead of encoding the radio at all, this PR now just exposes a `selector-check` target on the existing checkmark. A theme hides it (`display: none`) and composes its own selected indicator — e.g. a real `RadioListItem` — through the existing `renderOption` hook, styled by the existing `radio` / `radio-dot` targets. The marker's existence, position, and a11y stay the component's job; only the glyph's visibility/styling becomes themeable. No prop, no enum, no pseudo-element injection — and it removes the one structural selector the checkmark suppression previously needed.

Keen to hear if this lands better for you.

cixzhang (approved):

PR #4647, opened 2026-08-02
feat(theme): add component icon mappings

cixzhang:
## Why

Some icon choices are component semantics rather than global glyph choices. Selector is a good example: its selected-option indicator defaults to `check`, but a theme may want that same purpose to render a different registry icon or no icon at all.

## What

Adds themeable component icon mappings on top of the theme icon registry work:

- Adds `componentIcons` to `defineTheme()` / `DefinedTheme`, inherited through `extends`.
- Adds `useIcon(slot, fallback)` for component internals and `getComponentIcon()` / `getComponentIconName()` for explicit-source resolution.
- Wires Selector's selected-option indicator through the `selector-selected-option` slot.
- Documents component icon slots in component theming docs and CLI output.
- Adds a Storybook demo with an ad hoc Selector theme remapping the selected-option slot.
- Preserves `componentIcons` in `astryx theme build` output.

## Risk

The default Selector selected option remains a check icon. Themes can now remap or suppress that purpose through `componentIcons`; this is additive unless a theme opts in.

## Testing

- `pnpm exec vitest run --project ui packages/core/src/Icon/globalIconRegistry.test.tsx packages/core/src/Selector/Selector.test.tsx packages/core/src/theme/defineTheme.test.ts --reporter=dot`
- `pnpm exec vitest run --project node packages/cli/clients/cli/lib/component-format.test.mjs packages/cli/api/theme/build/build.test.mjs --reporter=dot`
- `pnpm -F @astryxdesign/core lint`
- `pnpm check:sync`
- `pnpm check:changesets`

Screenshot 2026-08-04 at 1 41 07 PM

### DIFF
--- a/packages/core/src/Icon/useIcon.ts
+++ b/packages/core/src/Icon/useIcon.ts
@@ -26,25 +26,24 @@ export function useIcon(name: IconName): ReactNode;
 /**
  * Resolve a component-specific semantic icon slot from the nearest Theme.
  *
- * `fallback` is the component's default global icon name. A theme can remap
- * the slot through `defineTheme({componentIcons})`; mapping the slot to `null`
- * intentionally renders no icon.
+ * `fallback` is the component's default global icon name, used when the theme
+ * does not remap the slot through `defineTheme({componentIcons})`.
  */
 export function useIcon(
   slot: ComponentIconSlotName,
-  fallback: IconName | null,
+  fallback: IconName,
 ): ReactNode;
 
 export function useIcon(
   nameOrSlot: IconName | ComponentIconSlotName,
-  fallback?: IconName | null,
+  fallback?: IconName,
 ): ReactNode {
   const themeName = useThemeName();
 
-  if (arguments.length === 2) {
+  if (fallback !== undefined) {
     return getComponentIcon(
       nameOrSlot as ComponentIconSlotName,
-      fallback ?? null,
+      fallback,
       themeName,
     );
   }
--- a/packages/core/src/Icon/useIcon.ts
+++ b/packages/core/src/Icon/useIcon.ts
@@ -4,9 +4,16 @@
 
 /**
  * @file useIcon.ts
- * @input Semantic icon name
- * @output Exports useIcon hook for theme-aware icon lookup
- * @position Client hook for components that render registry icons directly
+ * @input Semantic icon name, or a component icon slot plus its default
+ * @output Exports useIcon and useComponentIcon for theme-aware icon lookup
+ * @position Client hooks for components that render registry icons directly
+ *
+ * Two hooks rather than one overloaded hook, because the two lookups need
+ * different arguments and only one of them can be got wrong: a slot is
+ * meaningless without the component's default to fall back to. Separate names
+ * make the fallback a required parameter of the function that needs it,
+ * instead of an optional second argument whose absence silently changes which
+ * lookup runs.
  */
 
 import type {ReactNode} from 'react';
@@ -20,33 +27,30 @@ import {
 
 /**
  * Resolve a global semantic icon name from the nearest Theme.
+ *
+ * @example
+ * ```tsx
+ * const moreIcon = useIcon('moreHorizontal');
+ * ```
  */
-export function useIcon(name: IconName): ReactNode;
+export function useIcon(name: IconName): ReactNode {
+  return getIcon(name, useThemeName());
+}
 
 /**
  * Resolve a component-specific semantic icon slot from the nearest Theme.
  *
- * `fallback` is the component's default global icon name, used when the theme
- * does not remap the slot through `defineTheme({componentIcons})`.
+ * `fallback` is the component's own default global icon name, used when the
+ * theme does not remap the slot through `defineTheme({componentIcons})`.
+ *
+ * @example
+ * ```tsx
+ * const mark = useComponentIcon('selector-selected-option', 'check');
+ * ```
  */
-export function useIcon(
+export function useComponentIcon(
   slot: ComponentIconSlotName,
   fallback: IconName,
-): ReactNode;
-
-export function useIcon(
-  nameOrSlot: IconName | ComponentIconSlotName,
-  fallback?: IconName,
 ): ReactNode {
-  const themeName = useThemeName();
-
-  if (fallback !== undefined) {
-    return getComponentIcon(
-      nameOrSlot as ComponentIconSlotName,
-      fallback,
-      themeName,
-    );
-  }
-
-  return getIcon(nameOrSlot, themeName);
+  return getComponentIcon(slot, fallback, useThemeName());
 }
--- a/packages/core/src/Icon/useIcon.ts
+++ b/packages/core/src/Icon/useIcon.ts
@@ -11,13 +11,43 @@
 
 import type {ReactNode} from 'react';
 import {useThemeName} from '../theme/useTheme';
-import {getIcon, type IconName} from './globalIconRegistry';
+import {
+  getComponentIcon,
+  getIcon,
+  type ComponentIconSlotName,
+  type IconName,
+} from './globalIconRegistry';
 
 /**
- * Resolve a semantic icon name from the nearest Theme, falling back through
- * root theme registration, global icon overrides, and built-in defaults.
+ * Resolve a global semantic icon name from the nearest Theme.
  */
-export function useIcon(name: IconName): ReactNode {
+export function useIcon(name: IconName): ReactNode;
+
+/**
+ * Resolve a component-specific semantic icon slot from the nearest Theme.
+ *
+ * `fallback` is the component's default global icon name. A theme can remap
+ * the slot through `defineTheme({componentIcons})`; mapping the slot to `null`
+ * intentionally renders no icon.
+ */
+export function useIcon(
+  slot: ComponentIconSlotName,
+  fallback: IconName | null,
+): ReactNode;
+
+export function useIcon(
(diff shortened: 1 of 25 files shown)


packages/core/src/Selector/Selector.doc.mjs
```diff
@@ -23,9 +23,18 @@ export const docs = {
     targets: [
       {className: 'astryx-selector', visualProps: ['size', 'status']},
       {className: 'astryx-selector-option'},
+      {className: 'astryx-selector-selected-option-icon'},
```
cixzhang: Should this leave a className?

packages/core/src/Selector/Selector.doc.mjs
```diff
@@ -23,9 +23,18 @@ export const docs = {
...
       {className: 'astryx-selector-clear-icon'},
       {className: 'astryx-selector-indicator-icon', states: ['state']},
     ],
+    icons: [
```
cixzhang, line 34: Let's also include a table for this in the docsite's canary theming section

packages/core/src/Selector/Selector.tsx
```diff
@@ -942,7 +949,14 @@ export function Selector<T extends SelectorOptionType>(
               <DefaultOption option={item} />
             )}
           </span>
-          {isSelected && <Icon icon="check" size="sm" color="accent" />}
+          {isSelected && selectedOptionIcon != null && (
+            <span
```
cixzhang: Hmm should we handle most of this logic within Icon instead?

rubyycheung (approved):

packages/core/src/theme/defineTheme.ts
```diff
@@ -280,6 +280,11 @@ export interface DefineThemeInput {
   components?: ComponentStyleMap;
   /** Icon registry — maps semantic icon names to React nodes */
   icons?: Partial<IconRegistry>;
+  /**
+   * Component icon slot mappings — maps component-specific purposes to global
+   * semantic icon names. Use `null` to intentionally render no icon.
+   */
+  componentIcons?: ComponentIconMap;
```
cixzhang: This should be Partial, right?

packages/core/src/theme/defineTheme.ts
```diff
@@ -329,6 +334,8 @@ export interface DefinedTheme {
   components?: ComponentStyleMap;
   /** Icon registry */
   icons?: Partial<IconRegistry>;
+  /** Component icon slot mappings */
+  componentIcons?: ComponentIconMap;
```
cixzhang: Same here.

packages/core/src/theme/defineTheme.ts
cixzhang: It already was, just hidden inside the alias — `ComponentIconMap` was declared as `Partial<Record<ComponentIconSlotName, IconName | null>>`. But that's exactly why the question is worth acting on: the field read as non-partial next to `icons?: Partial<IconRegistry>`.

Flipped it to match the sibling convention in 4423a81: `ComponentIconMap` is now the complete `Record`, and both theme fields declare `Partial<ComponentIconMap>`. Same runtime behavior, and partialness is now visible where it's used. Safe to reshape since the type isn't released yet.

packages/core/src/theme/defineTheme.ts
cixzhang: Done in the same commit — `DefinedTheme.componentIcons` is `Partial<ComponentIconMap>` too.

packages/core/src/Selector/Selector.doc.mjs
cixzhang: No — resolved. The new `astryx-selector-selected-option-icon` target is gone; the slot now rides main's existing `astryx-selector-check` target. Two independent axes on one element: `componentIcons['selector-selected-option']` picks *which* glyph, `astryx-selector-check` styles *how it looks*. No new theme surface for this feature.

packages/core/src/Selector/Selector.doc.mjs
cixzhang: Done — `ComponentDoc.theming.icons[]` is a new authoring field (slot / default / description), and `Theming.tsx` renders it as an **Icon slots** table in the canary theming section, next to the targets table. Selector is the first component to declare one.

packages/core/src/Selector/Selector.tsx
cixzhang: Agreed, and that's where it lives now. The wrapper span is gone; the callsite is just:

```tsx
<Icon icon="selector-selected-option" fallbackIcon="check" size="sm" color="accent" {...themeProps('selector-check')} />
```

`Icon` does the slot resolution (`useIcon`), the fallback, and the render-nothing case when a theme maps the slot to `null`. Components just name the slot.

packages/core/src/theme/defineTheme.ts
```diff
@@ -622,11 +629,19 @@ export function defineTheme(input: DefineThemeInput): DefinedTheme {
       ? {...base.icons, ...input.icons}
       : (input.icons ?? base?.icons);
 
+  // 6. Merge component icon mappings — input mappings override base mappings.
+  // `null` is an intentional override meaning “render no icon”.
```
cixzhang: I think we should skip null as an intentional override for now. It's often confused for "default" so if we do decide to add one later we should have a clear symbol.

cixzhang: Agreed — dropped it in 96a764c.

A theme now maps a slot to another global icon name, or leaves it alone; there is no value that means "render nothing." `ComponentIconMap` values are `IconName`, and `Icon`'s `fallbackIcon` / `useIcon`'s fallback follow (no `| null`), so the "is this the default or is this off?" question can't be asked. The doc prose, the docsite icon-slots table, and the CLI component table all lose their "none" case too.

Left a comment on `getComponentIconName` recording why, so a future reader doesn't add it back casually: if suppression turns out to be a real need it should arrive as an explicit symbol.

And it isn't load-bearing for selection — an indicator draws its own empty state (an unchecked check indicator renders nothing) rather than relying on the slot to vanish.

packages/core/src/Icon/useIcon.ts
```diff
@@ -11,13 +11,42 @@
...
+ */
+export function useIcon(
+  slot: ComponentIconSlotName,
+  fallback: IconName,
+): ReactNode;
+
+export function useIcon(
+  nameOrSlot: IconName | ComponentIconSlotName,
+  fallback?: IconName,
+): ReactNode {
```
cixzhang: It may be better to have useComponentIcon instead since the check for calling getComponentIcon is whether a fallback is defined but there's nothing guaranteeing a fallback is provided when component icon is provided in nameOrSlot.

cixzhang: Good catch, and it was worse than the hook — done in d0e38d0.

Your read is right that nothing tied the two together. The overload *did* reject `useIcon('selector-selected-option')` (disjoint unions, no matching signature), but I probed the neighbor and found the real hole:

```tsx
<Icon icon="selector-selected-option" />   // typechecked fine
```

That resolves through no fallback, so an unmapped slot renders **nothing** — a blank where the check should be, no error anywhere.

Two changes:

1. **`useComponentIcon(slot, fallback)`** as its own export, as you suggest; `useIcon(name)` goes back to a single signature. The fallback is now a required parameter of the function that needs it, instead of an optional second argument whose absence silently picked the lookup. No production callsite used the two-arg form, so the split is clean.

2. **`IconProps` is a discriminated union** — a global icon takes `fallbackIcon?: never`, a slot takes `fallbackIcon: IconName` (required). Same guarantee at the component, where the hole actually was.

Pinned with `@ts-expect-error` tests, since compile time is the only place these can be checked. Verified they bite by reverting the union — both directives went unused and the build failed.

cixzhang:
Closing this — we're deferring component icon mappings entirely, and taking the minimal API first.

**Why.** Every motivating case for icon selection turns out to be *stateful*: the checkmark vs. radio dot in a selector option, the chevron on an expand toggle, the arrow vs. chevron for sort. Those are indicator families, and indicator replacement covers them without a slot map. The cases that genuinely need a static glyph swapped — the clear `×`, the status icons — are ones nobody has actually asked for. This PR was infrastructure ahead of demand.

**What replaces it.** `defineTheme({indicators})` alone: a theme replaces a *named* indicator, so a product that wants radio visuals in its selector options replaces `CheckIndicator` and every consumer of it follows. No per-slot mapping, no new key namespace, and nothing to version yet.

**Deliberately leaving room.** Slot maps stay additive: `componentIcons` / `componentIndicators` can arrive later, keyed by name so they still survive `theme build`, once there's a use case that indicator replacement can't serve. Deferring is cheaper than shipping a key namespace we'd have to keep — as the discussion here established, slot names are versioned API, and removing one is *silent* (resolution falls back to the default, no warning).

**Worth keeping from this PR**, as separate changes:
- The **field-status icon dedup** — four identical `Record<StatusType, IconName>` literals across `useInputStatusIcon`, `FieldStatus`, `Selector`, and `MultiSelector`. Pure consolidation, no new API.
- The **`theme build` silent-loss guard**. A runtime-value theme field written inline rather than as an imported binding is dropped from the built module with no error — `indicators: {check: MyCheck}` builds green and does nothing. That mattered little when `componentIcons` was the main path; with `indicators` as the *only* mechanism it should be a hard error.

Thanks for the reviews — the `Partial`, `null`, and `useComponentIcon` catches all sharpened the design, and the `null` discussion is a good part of why the slot layer looks premature.

Continuing in #4712, rescoped to indicators only.

PR #4648, opened 2026-08-02
feat(icon): add form control glyphs

cixzhang:
## Why

Checkbox and radio internals should use the stateful control-icon renderer layer instead of owning duplicated control drawing logic. That lets themes replace the whole control visual while CheckboxInput, RadioList, and DropdownMenu keep owning accessibility and interaction behavior.

## What

Routes form-control internals through `useControlIcon()`:

- CheckboxInput renders the theme's `checkbox` control icon with `unchecked` / `checked` / `indeterminate` state.
- RadioListItem renders the theme's `radio` control icon with `unchecked` / `checked` state.
- DropdownMenuCheckboxItem and DropdownMenuRadioItem use the same control icon renderers for decorative menu selection controls.
- Removes the now-unneeded static checkbox/radio icon-name additions from this slice.
- Updates tests and the Storybook demo to use `controlIcons`.

## Risk

The controls still own semantics, focus, disabled behavior, and sizing. The visual renderer is now themeable through the control icon layer; default behavior comes from the base renderers in the previous PR.

## Testing

- `pnpm exec vitest run --project ui packages/core/src/CheckboxInput/CheckboxInput.test.tsx packages/core/src/RadioList/RadioList.test.tsx packages/core/src/DropdownMenu/DropdownMenuSelectable.test.tsx packages/core/src/Icon/useControlIcon.test.tsx --reporter=dot`
- `pnpm -F @astryxdesign/core lint`
- `pnpm check:sync`
- `pnpm check:changesets`

### DIFF
--- a/packages/core/src/DropdownMenu/DropdownMenuCheckboxItem.tsx
+++ b/packages/core/src/DropdownMenu/DropdownMenuCheckboxItem.tsx
@@ -15,24 +15,17 @@
  * Enter/Space activation come from the parent DropdownMenu's useListFocus +
  * activation path, which matches menuitemcheckbox alongside plain menuitem rows.
  *
- * The checkbox visual composes the real CheckboxInput primitive so its checkmark
- * matches CheckboxListItem and picks up the standard `checkbox` theming slots.
- * It is purely decorative: the composed control is wrapped in an element that is
- * both `aria-hidden` and `inert`, so it contributes nothing to the row's
- * accessible name, and its native <input> and sr-only label stay out of the tab
- * order and the accessibility tree while pointer clicks fall through to the row
- * — the same shim MultiSelector uses. The row owns the checked state and
- * accessible name. The control size is derived from the menu's item size (a
- * `sm` menu gets the compact control; `md`/`lg` get the standard one) and it
- * swaps to the inline-end of the row on coarse-pointer (touch) devices via CSS
- * `order`, so it lands where selection toggles are conventionally placed on
- * mobile.
+ * The checkbox visual is decorative: the row owns the role, checked state, and
+ * accessible name. The glyph resolves through a component icon slot so themes can
+ * align menu selection marks with CheckboxInput or intentionally diverge. The
+ * control size is derived from the menu's item size and swaps to the inline-end
+ * of the row on coarse-pointer (touch) devices via CSS `order`, so it lands
+ * where selection toggles are conventionally placed on mobile.
  */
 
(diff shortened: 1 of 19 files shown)


packages/cli/assets/docs/icons.doc.mjs
```diff
@@ -29,7 +29,10 @@ export const docs = {
             ['chevronDown', 'Dropdown triggers, expand/collapse'],
             ['chevronLeft', 'Navigate back, previous'],
             ['chevronRight', 'Navigate forward, next'],
-            ['check', 'Checkbox checked, confirm'],
+            ['check', 'Confirm, selected option'],
```
cixzhang: Checkmark icon, specifically. Describe this in a way  that can differentiate it from the others

packages/cli/assets/docs/icons.doc.mjs
```diff
@@ -29,7 +29,10 @@ export const docs = {
...
+            ['checkboxChecked', 'Checked checkbox glyph'],
+            ['checkboxIndeterminate', 'Indeterminate checkbox glyph'],
+            ['radioChecked', 'Selected radio glyph'],
```
cixzhang: Missing unchecked states for radio and checkmarks

packages/core/src/DropdownMenu/DropdownMenuCheckboxItem.tsx
```diff
@@ -74,6 +86,30 @@ const styles = stylex.create({
       '@media (pointer: coarse)': 'auto',
     },
   },
+  markerUnchecked: {
```
cixzhang: These should be part of the icon not the component

packages/core/src/Icon/defaultIcons.tsx
```diff
@@ -89,6 +101,37 @@ export const defaultIcons: IconRegistry = {
     </svg>
   ),
 
+  /** checkbox checked glyph — checkmark for checked checkbox state */
+  checkboxChecked: (
```
cixzhang: Mark sure these render exactly the way they did before. We may want an internal set of checkbox and radio that take in states and just renders the components. It should be the full checkmark box, radio circle, etc so may not be a full svg. Also reuse the component for each state just change some state prop.

Maybe we should make them public like "radio icon" and "checkbox icon"?

cixzhang:
Superseded by #4713, which is rebased on the new controlIcons layer (#4712).

PR #4654, opened 2026-08-02
fix(theme): make shared focus outlines themeable

cixzhang:
## Why

The standard focus outline was written out independently in every component that draws one — same intent, five separate definitions. Beyond the duplication, they had **drifted apart**, which is the more interesting finding (below). This came up while evaluating agentcloud and trying to replicate its styling through Astryx themes.

Astryx has a deliberate split here: input/field focus treatments differ from general interactive chrome. This PR targets the general outline only.

## What

Adds `packages/core/src/utils/focusOutline.stylex.ts` — one definition, in four application shapes:

| | for |
|---|---|
| `focusVisible` | the focused element draws its own ring |
| `focusWithin` | a container rings when something inside it takes focus |
| `publishFocusVisibleVars` | an element publishes its ring outward as CSS vars |
| `focusWithinOrPublished` | a container reads a published ring, or draws its own |

Applied across 26 components of general interactive chrome. Outline primitives stay private to the helper. **Net −438 lines.**

## Resolved real drift from Design Conventions

The consolidation forced a question the duplication had hidden: *which* value is correct? Design Conventions, §User Interaction States, is explicit:

> **Focused (Outline style)** `:focus-visible` — 2px `--color-accent` outline at **3px offset**

Rendered-ring measurements across the affected components showed **most had drifted to a 2px offset**. Only Button, Calendar, Dialog and Pagination still matched the spec. This PR takes the documented value, so Link, TabList, Token, TreeList, SegmentedControl and TopNav items move *to* spec — their ring now sits 1px further from the control.

Measured in Chrome, after:

```
Button/primary      2px solid rgb(38,38,38)  @3px
Button/destructive  2px solid rgb(165,12,37) @3px
Calendar            2px solid rgb(38,38,38)  @3px
Dialog              2px solid rgb(38,38,38)  @3px
Pagination          2px solid rgb(38,38,38)  @3px
Link                2px solid rgb(38,38,38)  @3px
TabList             2px solid rgb(38,38,38)  @3px
Token               2px solid rgb(38,38,38)  @3px
TreeList            2px solid rgb(38,38,38)  @3px
SegmentedControl    2px solid rgb(38,38,38)  @3px
TopNavItem          2px solid rgb(38,38,38)  @3px
```

`--button-focus-offset` is retained. Its default is no longer Button-specific, but it is a documented public var and it still overrides the shared offset.

## Not themeable yet — deliberately

An earlier revision exposed a shared `astryx-focus-outline` theme target. **Removed**, because the theme system cannot express what such a target is for. A focus ring is a *state* treatment, and `generateThemeRules` mishandles state keys, silently, in two ways:

```ts
// nested under base — both blocks DROPPED, only outlineOffset survives
{base: {outlineOffset: '4px', ':focus-visible': {outline: '...'}}}

// as a top-level key — emits selectors that can never match
':focus-visible'        →  .astryx-focus-outline.focus-visible   // pseudo became a class
':has(:focus-visible)'  →  .astryx-focus-outline.has(            // truncated; invalid CSS
```

So a theme could only override the ring *unconditionally* — a permanent outline on 26 components — while `Theme.doc.mjs` advertised exactly those two states. The consolidation stands on its own; themeability is a good follow-up once pseudo-selector handling is fixed, and this module is the foundation for it.

## Three bugs found in review, all by measuring rather than reading

1. **TreeList lost its focus ring entirely.** The helper spread its own style *first*, and StyleX is last-wins — `contentWrapper`'s `outline: 'none'` deleted it. A focused tree row painted nothing (WCAG 2.4.7).
2. **Destructive buttons lost their error-red ring.** The shared style used the `outline` **shorthand**, which resets every longhand it covers — so a variant's `outlineColor` could not survive it at any ordering. The ring is longhands now, and the redundant reset in TreeList is gone.
3. **Four empty style objects** left behind by the refactor (`Avatar`, `Item`, `Token`, `TopNavItem`), plus their call-site references.

The existing unit tests missed (1) and (2): they assert on published CSS variables and class presence, both of which were fine. Only the paint was wrong.

## Test plan

- **Rendered ring diffed against `origin/main` for 22 components**, Tab-focused, in real Chrome — the sweep that found the drift, and the only check that catches this class of bug.
- Full UI suite green; touched components 480/480 in isolation.
- Typecheck clean; lint clean at CI strict tier (`CI=true`).
- Rebased onto current `main` (was 131 commits behind). `Timestamp` is untouched by this PR — `main` deliberately removed that element's ring in the meantime.

## Follow-ups

1. `generateThemeRules` drops or mangles pseudo-selector keys — affects any component target declaring `states`.
2. Re-add the shared theme target once (1) is fixed.
3. `CheckboxInput` / `RadioListItem` aren't in the sweep; they're the two controls whose focus lives on a visually hidden input, and should consume `publishFocusVisibleVars` next.

### DIFF
--- a/packages/core/src/Avatar/Avatar.tsx
+++ b/packages/core/src/Avatar/Avatar.tsx
@@ -154,7 +154,6 @@ const styles = stylex.create({
   // Visible focus ring for the name-tooltip tab stop, matching the repo-wide
   // focus-visible outline treatment (see Timestamp, Token, Thumbnail). Only
   // applied when a tooltip is active so keyboard users can reveal it.
-  focusable: {},
   // Reset the intrinsic styling of the interactive element (<a>/<button>) so it
   // is a transparent, correctly-sized wrapper around the avatar visuals. The
   // element carries the focus-visible accent ring for keyboard users.
@@ -587,7 +586,6 @@ export function Avatar({
     focusOutlineProps.focusVisible(
       styles.wrapper,
       isInteractive && styles.interactive,
-      !isInteractive && showTooltip && !avatarGroup && styles.focusable,
       avatarGroup && groupStyles.ring,
       avatarGroup && groupStyles.overlap,
       avatarGroup && groupDynamicStyles.overlap(-avatarGroup.overlap),
--- a/packages/core/src/Avatar/Avatar.tsx
+++ b/packages/core/src/Avatar/Avatar.tsx
@@ -31,6 +31,7 @@ import {AvatarSizeContext} from './AvatarSizeContext';
 import {useAvatarGroup} from '../AvatarGroup/AvatarGroupContext';
 import {mergeProps, mergeRefs} from '../utils';
 import {themeProps} from '../utils/themeProps';
+import {focusOutlineProps} from '../utils/focusOutline.stylex';
 import {useTooltip} from '../Tooltip/useTooltip';
 import {useLinkComponent} from '../Link/useLinkComponent';
 import type {LinkComponentType} from '../Link/types';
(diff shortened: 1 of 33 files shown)


packages/core/src/Button/Button.tsx
```diff
@@ -252,18 +225,29 @@ const variants = stylex.create({
       },
       ':active': `linear-gradient(${colorVars['--color-overlay-pressed']}, ${colorVars['--color-overlay-pressed']})`,
     },
-    outline: {
-      default: null,
```
cixzhang, line 256: This is kind of nice to have which is an error-specific outline color.

rubyycheung (approved):

packages/core/src/Avatar/Avatar.tsx
```diff
@@ -153,16 +154,7 @@ const styles = stylex.create({
...
-    outline: {
-      default: null,
-      ':focus-visible': `2px solid ${colorVars['--color-accent']}`,
-    },
-    outlineOffset: {
-      default: '0',
-      ':focus-visible': '2px',
-    },
-  },
+  focusable: {},
```
cixzhang: Cleanup?

PR #4659, opened 2026-08-02
feat(core): add ComplexSelector

cixzhang:
## Summary

Adds `ComplexSelector`, a high-level selector shell for rich custom selection surfaces. It keeps the XDS-style single-component API while letting consumers compose their own accessible content from `value`, `onChange`, `close`, and state helpers.

## What changed

- Adds an accessible Field + button + dialog-popover selector shell with optimistic `changeAction` support and generated exports.
- Simplifies the API to open composition: custom content owns its list/grid/tree/date/input semantics, while `ComplexSelector` owns trigger semantics, popover behavior, focus restore, and the async change flow.
- Keeps the closed-trigger API narrow with `triggerLabel` only.
- Updates Storybook examples:
  - Fruit/ripeness selector restyled toward a compact row + pill selector pattern, with `useGridFocus` preserving columns for keyboard users.
  - Project destination selector combining `TextInput` search with `TreeList` hierarchy.
  - Product category selector showing the same tree-search pattern in a form-like context.
- Adds docs guidance to use Astryx focus hooks for custom content and evaluate against WCAG 2.2.
- Adds a focused `complex-selector` vibe-test set covering grid, tree/search, date-time, and freeform-option selectors.

## Risk

This is additive. The main design question is whether the open render-prop shape is enough guidance for agents to choose the right accessible structure inside the dialog.

## WCAG 2.2 evaluation notes

- **2.1.1 Keyboard / 2.1.2 No Keyboard Trap:** trigger is a button; popover content uses focusable controls. Fruit grid uses `useGridFocus`; tree examples rely on `TreeList` keyboard behavior.
- **2.4.3 Focus Order:** focus enters the dialog content and returns to the trigger on close. Grid arrow movement preserves row/column expectations.
- **2.4.7 Focus Visible / 2.4.11 Focus Not Obscured:** custom fruit pills provide explicit `:focus-visible` outline; TreeList/TextInput/Button use component focus treatment inside the popover.
- **2.5.8 Target Size:** fruit selector pills are compact and adjacent, but each has spacing and visible focus. If this pattern becomes production-facing, confirm target sizing against the final density requirements.
- **3.3.2 Labels or Instructions:** ComplexSelector fields and search inputs have labels/descriptions; hidden search labels remain accessible.
- **4.1.2 Name, Role, Value:** fruit grid uses `role="grid"`, `row`, `rowheader`, `gridcell`, and `aria-selected`; tree examples delegate hierarchy semantics to `TreeList`.
- **1.4.3 Contrast / 1.4.11 Non-text Contrast:** examples use semantic theme tokens for text, borders, focus outlines, and selected states rather than hardcoded colors.

## Vibe test

Focused iteration: `3e411072` using `internal/vibe-tests/test-sets/complex-selector.json`.

- Aggregate score: **94 overall**
  - Correctness 93
  - Accessibility 100
  - Code quality 96
  - Efficiency 90
  - Maintainability 93
- Typecheck: **3/4 generated outputs clean**
  - `cs-1` fruit/ripeness grid: clean
  - `cs-2` destination tree/search picker: clean
  - `cs-4` snack + Other text input: clean
  - `cs-3` deadline picker: failed on branded `ISODateString` / `ISOTimeString` types, which looks like DateInput/TimeInput discoverability friction rather than a ComplexSelector issue.

## Testing

- `pnpm --filter @astryxdesign/core typecheck`
- `pnpm vitest run packages/core/src/ComplexSelector/ComplexSelector.test.tsx`
- `pnpm exec eslint packages/core/src/ComplexSelector/ComplexSelector.tsx packages/core/src/ComplexSelector/ComplexSelector.test.tsx apps/storybook/stories/ComplexSelector.stories.tsx`
- `pnpm sync:exports:check`
- `pnpm --filter @astryxdesign/storybook typecheck` (still blocked by existing workspace package-resolution errors; no ComplexSelector story errors after fixes)
- `pnpm -F @astryxdesign/vibe-tests interactive --test-set complex-selector --target astryx --label complex-selector-pr4659`
- `npx tsx src/build-previews.ts --iterations 3e411072 --tsc-only` from `internal/vibe-tests`
- `pnpm -F @astryxdesign/vibe-tests aggregate --iteration 3e411072`

### DIFF
--- a/packages/core/src/ComplexSelector/ComplexSelector.tsx
+++ b/packages/core/src/ComplexSelector/ComplexSelector.tsx
@@ -4,7 +4,7 @@
 
 /**
  * @file ComplexSelector.tsx
- * @input Uses React, StyleX, Field, usePopover, useGridFocus
+ * @input Uses React, StyleX, Field, usePopover
  * @output Exports ComplexSelector component for custom selector surfaces
  * @position Core implementation; consumed by index.ts
  *
@@ -17,7 +17,6 @@
 
 import React, {
   useCallback,
-  useEffect,
   useId,
   useOptimistic,
   useTransition,
@@ -29,7 +28,7 @@ import type {BaseProps} from '../BaseProps';
 import {Field, inputWrapperStyles, type FieldStatusVariant} from '../Field';
 import {Icon} from '../Icon';
 import {Spinner} from '../Spinner';
-import {useGridFocus} from '../hooks/useGridFocus';
+import {useTranslator} from '../i18n';
 import {layerAnimations} from '../Layer/layerAnimations.stylex';
 import type {LayerPlacement} from '../Layer/useLayer';
 import {usePopover} from '../Popover/usePopover';
@@ -43,12 +42,9 @@ import {
   typographyVars,
   typeScaleVars,
 } from '../theme/tokens.stylex';
-import {useTranslator} from '../i18n';
 import {mergeProps} from '../utils';
 import type {SizeValue} from '../utils/types';
 import {themeProps} from '../utils/themeProps';
-
-const OPTION_SELECTOR = '[data-astryx-complex-selector-option]';
 
 const styles = stylex.create({
   triggerContainer: {
@@ -127,7 +123,6 @@ const styles = stylex.create({
     maxHeight: 'min(480px, calc(100vh - 32px))',
     overflow: 'auto',
     padding: spacingVars['--spacing-3'],
-    outline: 'none',
   },
   sm: {
     minHeight: sizeVars['--size-element-sm'],
@@ -151,48 +146,7 @@ const styles = stylex.create({
 
 export type ComplexSelectorSize = 'sm' | 'md' | 'lg';
 
-export interface ComplexSelectorGridLayout {
-  /** Use the WAI-ARIA grid pattern for a two-dimensional picker. */
-  type: 'grid';
-  /** Number of visual columns. ArrowUp/ArrowDown preserve the current column. */
-  columns: number;
-}
-
-export type ComplexSelectorLayout = ComplexSelectorGridLayout;
-
-export interface ComplexSelectorGetOptionPropsOptions<Value> {
-  /** Zero-based DOM/grid index for the option. */
-  index: number;
-  /** Value to commit when the option is selected. */
-  value: Value;
-  /** Accessible label for this option. */
-  label: string;
-  /** Whether this option represents the current value. */
-  isSelected?: boolean;
-  /** Whether this option is visible but unavailable. */
-  isDisabled?: boolean;
-}
-
-export interface ComplexSelectorOptionProps {
-  id: string;
-  role?: 'gridcell';
-  'aria-label': string;
-  'aria-selected'?: boolean;
-  'aria-disabled'?: true;
-  'data-astryx-complex-selector-option': string;
-  tabIndex: 0 | -1;
-  onClick: () => void;
-}
-
-export interface ComplexSelectorRenderProps<Value> {
-  /** Current optimistic value. */
-  value: Value;
-  /** Commit a value through onChange/changeAction. */
-  onChange: (value: Value) => void;
-  /** Async action passed to ComplexSelector, exposed for composed content. */
-  changeAction?: (value: Value) => void | Promise<void>;
-  /** Close the selector surface. */
-  close: () => void;
+export interface ComplexSelectorRenderState {
   /** Whether the selector surface is open. */
   isOpen: boolean;
   /** Whether changeAction/isLoading is pending. */
@@ -201,10 +155,6 @@ export interface ComplexSelectorRenderProps<Value> {
   triggerId: string;
   /** ID of the popup content container. */
   contentId: string;
-  /** Props for selectable options inside the custom content. */
-  getOptionProps: (
-    options: ComplexSelectorGetOptionPropsOptions<Value>,
-  ) => ComplexSelectorOptionProps;
 }
 
 export interface ComplexSelectorTriggerRenderProps<Value> {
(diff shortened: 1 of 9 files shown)


packages/core/src/ComplexSelector/ComplexSelector.tsx
```diff
@@ -0,0 +1,588 @@
...
+  isDisabled?: boolean;
+}
+
+export interface ComplexSelectorOptionProps {
+  id: string;
+  role?: 'gridcell';
+  'aria-label': string;
+  'aria-selected'?: boolean;
+  'aria-disabled'?: true;
+  'data-astryx-complex-selector-option': string;
```
cixzhang: why as a prop?

packages/core/src/ComplexSelector/ComplexSelector.tsx
```diff
@@ -0,0 +1,588 @@
...
+  /** Whether changeAction/isLoading is pending. */
+  isBusy: boolean;
+  /** ID of the trigger button. */
+  triggerId: string;
+  /** ID of the popup content container. */
+  contentId: string;
+  /** Props for selectable options inside the custom content. */
+  getOptionProps: (
+    options: ComplexSelectorGetOptionPropsOptions<Value>,
+  ) => ComplexSelectorOptionProps;
```
cixzhang: since the custom content is completely composable, is there an advantage for having a separate option props handler? Why not compose the options into the content area?

packages/core/src/ComplexSelector/ComplexSelector.tsx
```diff
@@ -0,0 +1,588 @@
...
+ *       <button
+ *         {...getOptionProps({
+ *           index: row * levels.length + column,
+ *           value: {fruit, ripeness: level},
+ *           label: `${fruit} ${level}`,
+ *         })}>
+ *         {fruit} {level}
+ *       </button>
+ *     )),
+ *   )}
```
cixzhang: Hmm I think a more open composition pattern might be more ergonomic but we can vibe test it:

```
<ComplexSelector label="Fruits" value="banana" onChange={...}>
  {/* this content will be within a dialog */}
  {(value, onChange, close) => (
    <MyFruitGrid value={value} onChange={onChange} changeAction={changeAction} ... />
  )}
</ComplexSelector>
```

packages/core/src/ComplexSelector/ComplexSelector.tsx
```diff
@@ -0,0 +1,588 @@
...
+  onClick: () => void;
+}
+
+export interface ComplexSelectorRenderProps<Value> {
+  /** Current optimistic value. */
+  value: Value;
+  /** Commit a value through onChange/changeAction. */
+  onChange: (value: Value) => void;
+  /** Async action passed to ComplexSelector, exposed for composed content. */
+  changeAction?: (value: Value) => void | Promise<void>;
```
cixzhang: It will be easy for the consumer to forget to handle this so debating whether we just pass onChange instead and have it handled through there. Any drawbacks or missing/incorrect functionality from this?

packages/core/src/ComplexSelector/ComplexSelector.tsx
```diff
@@ -0,0 +1,588 @@
...
+  /** Trigger and field size. */
+  size?: ComplexSelectorSize;
+  /** Width of the field. */
+  width?: SizeValue;
+  /** Popup placement. */
+  placement?: LayerPlacement;
+  /** HTML form field name. */
+  htmlName?: string;
+  /** Converts value for the hidden input. */
+  getFormValue?: (value: Value) => string;
```
cixzhang: Is this part of selectors normally?

cixzhang (commented):
Some applications to consider that we can build vibe tests around.

* date/time selectors, check on the current ones we have, should any of them have been complex selectors? If so, how could they have been accomplished through this framework?
* tree list selectors with search, multiple pages, etc. This was common internally for complex structures that needed navigation to perform selection
* fun grid-like selectors that reduce time to input multiple dimensions (ex. model + thinking level selection).
* Additional options that require inputs. Ex. choose "apple, banana, etc" but also have an input that lets you add a freeform option.

We can design a vibe test:

1. We should evaluate the right accessible design for each of these selectors.
2. Determine some naive prompts that someone might request a UI like this
3. See if the API can be discovered through Astryx and where the LLM runs into friction. Evaluate the results to see if it accomplished the user request and has the right accessible design we determined in step 1.

packages/core/src/ComplexSelector/ComplexSelector.tsx
cixzhang: Removed this from v1. The closed trigger API is now just `triggerLabel`; if we need state-aware trigger rendering later we can add it from a concrete use case.

packages/core/src/ComplexSelector/ComplexSelector.tsx
cixzhang: Updated to the open composition pattern. `getOptionProps` and `layout` are gone; custom content now owns its own options/list/grid/tree semantics inside the dialog content.

packages/core/src/ComplexSelector/ComplexSelector.tsx
cixzhang: Changed to this shape: `children` is now `(value, onChange, close, state) => ReactNode`. Added Storybook examples for the fruit grid plus tree/search selectors, and added `internal/vibe-tests/test-sets/complex-selector.json` to cover the proposed cases.

packages/core/src/ComplexSelector/ComplexSelector.tsx
cixzhang: Agreed. The render-prop `onChange` helper now wraps both the external `onChange` and optional `changeAction`, so composed content only needs to call the one callback. `changeAction` is no longer passed separately to content.

packages/core/src/ComplexSelector/ComplexSelector.tsx
cixzhang: Removed the form hidden-input props. They were not essential to the selector shell and made the v1 API broader than needed.

cixzhang:
Noticed that when tapping the trigger while it's open it stays open since it ends up closing the popovers due to auto and then retriggering the trigger.

PR #4667, opened 2026-08-03
fix(dropdown-menu): make submenu indicator icon and divider spacing themable

athz:
## What

Two follow-ups to the DropdownMenu theme-slot work (#4600) — both fix cases where a slot *exists* but the property a theme wants to change doesn't actually flow through it.

### Submenu indicator icon — size now themable
The `astryx-dropdown-menu-indicator-icon` target sat on the wrapper `<span>`, one level above the `<Icon>` that owns the glyph's `width`/`height`. A theme rule on the span could restyle the wrapper but never the icon size. Moved the target onto the `<Icon>` element itself (Icon merges `className` onto the same element that carries its size class), so a theme can now restyle the chevron's size and color directly — same mechanism as `selector-indicator-icon`.

### Menu divider spacing — vertical margin now themable
The menu divider's `margin-block` was a fixed token applied via a StyleX atomic class. The single-class `.astryx-dropdown-menu-divider` target (specificity 0,1,0) could not out-specify the global two-class `.astryx-divider.horizontal` rule (0,2,0) in the same layer, so a theme's margin override was always lost. Exposed the margin as `--_dropdown-menu-divider-margin` (default: the existing token), matching the component's existing `--_dropdown-menu-radius` / `--_dropdown-menu-padding` private-var pattern — a theme retunes the var instead of fighting specificity.

## Scope

Purely a theming-reachability fix — no behavior, layout, or a11y change; the default rendering is byte-identical. Documented the new private var in the component's theming `vars`/`derived`.

## Testing

- Updated the indicator-icon slot test to assert the target lands on the chevron glyph (`astryx-icon`), i.e. the element that owns the size.
- `@astryxdesign/core`: DropdownMenu + ContextMenu + Breadcrumbs suites pass (154 tests), plus `typecheck`, `typecheck:docs`, `sync-exports --check`, and lint.

### DIFF
--- a/packages/core/src/DropdownMenu/DropdownMenu.doc.mjs
+++ b/packages/core/src/DropdownMenu/DropdownMenu.doc.mjs
@@ -42,10 +42,12 @@ export const docs = {
     vars: [
       {name: '--_dropdown-menu-radius', description: 'Border radius of the menu popup', default: 'var(--radius-element)', private: true},
       {name: '--_dropdown-menu-padding', description: 'Inner padding of the menu popup', default: 'var(--spacing-1)', private: true},
+      {name: '--_dropdown-menu-divider-margin', description: 'Vertical margin above and below a menu divider', default: 'var(--spacing-1)', private: true},
     ],
     derived: [
       {property: 'borderRadius', vars: ['--_dropdown-menu-radius']},
       {property: 'padding', vars: ['--_dropdown-menu-padding']},
+      {property: 'marginBlock', vars: ['--_dropdown-menu-divider-margin']},
     ],
   },
   description: 'Main dropdown menu component with a trigger button and popup item list.',
--- a/packages/core/src/DropdownMenu/DropdownMenuSubMenu.tsx
+++ b/packages/core/src/DropdownMenu/DropdownMenuSubMenu.tsx
@@ -474,20 +474,17 @@ export function DropdownMenuSubMenu(
   );
 
   const endAffordance = hasSpinner ? (
-    <span
-      {...mergeProps(
-        themeProps('dropdown-menu-indicator-icon'),
-        stylex.props(triggerStyles.caret),
-      )}>
+    <span {...stylex.props(triggerStyles.caret)}>
       <Spinner size="sm" />
     </span>
   ) : (
-    <span
-      {...mergeProps(
-        themeProps('dropdown-menu-indicator-icon'),
-        stylex.props(triggerStyles.caret),
-      )}>
-      <Icon icon="chevronRight" size="sm" color="secondary" />
+    <span {...stylex.props(triggerStyles.caret)}>
+      <Icon
+        icon="chevronRight"
+        size="sm"
+        color="secondary"
+        {...themeProps('dropdown-menu-indicator-icon')}
+      />
     </span>
   );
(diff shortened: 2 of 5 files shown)


cixzhang:
I have some plans for how we do indicator icon mapping. For the other parts though:

- position of the indicator: as a prop is something I'm considering. There are genuinely times when left vs right might be used to improve vertical alignment or consistency with form content depending on what's important in any given use case
- divider: i think it could be okay for this to be themeable

packages/core/src/DropdownMenu/DropdownMenu.doc.mjs
```diff
@@ -42,10 +42,12 @@ export const docs = {
     vars: [
       {name: '--_dropdown-menu-radius', description: 'Border radius of the menu popup', default: 'var(--radius-element)', private: true},
       {name: '--_dropdown-menu-padding', description: 'Inner padding of the menu popup', default: 'var(--spacing-1)', private: true},
+      {name: '--_dropdown-menu-divider-margin', description: 'Vertical margin above and below a menu divider', default: 'var(--spacing-1)', private: true},
     ],
     derived: [
       {property: 'borderRadius', vars: ['--_dropdown-menu-radius']},
       {property: 'padding', vars: ['--_dropdown-menu-padding']},
+      {property: 'marginBlock', vars: ['--_dropdown-menu-divider-margin']},
```
cixzhang, line 50: Hmm I think this communicates that margin block is converted to the internal bar for block margin around the divider which is kind of odd at the menu level

packages/core/src/DropdownMenu/DropdownMenuSubMenu.tsx
```diff
@@ -474,20 +474,17 @@ export function DropdownMenuSubMenu(
   );
 
   const endAffordance = hasSpinner ? (
-    <span
-      {...mergeProps(
-        themeProps('dropdown-menu-indicator-icon'),
```
cixzhang, line 479: This should be kept the same

packages/core/src/DropdownMenu/DropdownMenuSubMenu.tsx
```diff
@@ -474,20 +474,17 @@ export function DropdownMenuSubMenu(
...
-        stylex.props(triggerStyles.caret),
-      )}>
-      <Icon icon="chevronRight" size="sm" color="secondary" />
+    <span {...stylex.props(triggerStyles.caret)}>
+      <Icon
+        icon="chevronRight"
+        size="sm"
+        color="secondary"
+        {...themeProps('dropdown-menu-indicator-icon')}
```
cixzhang, line 486: Hmm seems odd to theme the size here when we apply size consistently. Have an example for how you're adjusting it?

cixzhang (changes requested):

packages/core/src/DropdownMenu/DropdownMenu.doc.mjs
cixzhang: Maybe we make divider more directly themeable?

athz:
Superseded by #4743 — scoped down to divider-only changes per review feedback. Indicator icon work deferred.

PR #4676, opened 2026-08-03
docs(dense): restore dropped Icon and Carousel dense/zh best-practice bullets

josephfarina:
## What

Restore dropped translation-overlay best-practice bullets so `astryx component <Name> --lang dense|zh` renders the full guidance instead of silently falling back to English.

- **Icon** (`Icon.doc.mjs`): `docsDense` was missing one `bestPractices` bullet (11 English vs 10 dense) after #4551 added registerIcons/getExtendedIcon guidance to the English docs only. Restored the bullet and straightened one em dash in the English bullet to a colon.
- **Carousel** (`Carousel.doc.mjs`): #4629 added a `hasLoop` bullet to the English `bestPractices` but not to `docsDense` or `docsZh`, so both overlays were one bullet short and dropped the hasLoop guidance. Restored it in both. Also backfilled a pre-existing missing built-in-navigation bullet in the zh overlay so it reaches full parity.

Both files now match English bullet count and per-position `guidance` flags. Verified with the dense-coverage audit (0 bullet drifts) and by rendering `--lang dense` and `--lang zh` for both components.

## Checklist
- [x] `tsc --project packages/cli/tsconfig.template-docs.json --noEmit` passes
- [x] CLI `--lang dense` output verified for Icon and Carousel
- [x] CLI `--lang zh` output verified for Carousel
- [x] Bullet counts and guidance flags match English

---
Night Watch — Doc Reviewer

### DIFF
--- a/packages/core/src/Icon/Icon.doc.mjs
+++ b/packages/core/src/Icon/Icon.doc.mjs
@@ -55,7 +55,7 @@ export const docs = {
     description: 'Icons are small visual symbols that represent actions, objects, or concepts. They improve scannability and reinforce meaning alongside text. Supports both direct SVG components and semantic icon names that adapt to the active theme.',
     bestPractices: [
       { guidance: true, description: 'Use semantic icon names when available; they adapt to theme changes automatically.' },
-      { guidance: true, description: "Libraries can augment the icon map with their own keys: registerIcons({'richtext:bold': <MyIcon />}) accepts arbitrary extension keys (not just the built-in IconName set). Resolve them with getExtendedIcon(key, fallback), which prefers a theme-registered icon and falls back to a bundled default: the seam that makes library-shipped icons theme-overridable." },
+      { guidance: true, description: "Override icons through the theme, not globally: defineTheme({icons: {close: <XMarkIcon />}}) scopes the swap to the active <Theme>, and extends shallow-merges it into derived themes. registerIcons() mutates a process-wide registry and warns in dev, so keep it for app bootstrap rather than making it a library's theming seam." },
       { guidance: true, description: 'Pair icons with text labels for accessibility; icon-only elements need an accessible label.' },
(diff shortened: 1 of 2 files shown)


packages/core/src/Icon/Icon.doc.mjs
```diff
@@ -133,6 +133,7 @@ export const docsDense = {
     description: 'Icons are small visual symbols that represent actions, objects, or concepts. They improve scannability and reinforce meaning alongside text. Supports both direct SVG components and semantic icon names that adapt to the active theme.',
     bestPractices: [
       { guidance: true, description: 'Use semantic icon names when available; they adapt to theme changes automatically.' },
+      { guidance: true, description: 'Libraries can augment the icon map with their own keys: registerIcons({\'richtext:bold\': <MyIcon />}) accepts arbitrary keys (not just IconName). Resolve via getExtendedIcon(key, fallback), which prefers a theme-registered icon and falls back to a bundled default: the seam that makes library icons theme-overridable.' },
```
cixzhang: The preferred way to do this is via theme

josephfarina: Good call, and the code agrees with you: `registerIcons()` already `warnOnce`s that `defineTheme({ icons })` is the theme-scoped way to swap a glyph, so recommending the global registry as the seam for library icons contradicted our own runtime guidance.

Rewrote both the English and dense bullets to lead with `defineTheme({icons})` (scoped to the active `<Theme>`, shallow-merged through `extends`) and reframed `registerIcons()` as an app-bootstrap escape hatch rather than a theming seam.

I kept it to what's on `main` today. Once #4647 lands, the slot mechanism (`componentIcons` + `ComponentIconSlotMap` augmentation) is the better answer for the library case specifically, and I'm happy to follow up with a bullet for that then. Also worth noting the same `registerIcons`-as-seam prose exists in `RichTextEditor.doc.mjs` and in the `getExtendedIcon` JSDoc if you want those cleaned up too.

PR #4685, opened 2026-08-03
fix(a11y): name RichTextEditor and RichTextView text surfaces

AKnassa:
Part of #4681 (weekly a11y scan). Clears all 13 RichTextEditor entries from the axe baseline: 12x `aria-input-field-name` (serious) and 1x `label` (critical).

## What this does

Screen readers currently announce the RichTextEditor's editing area with no name, in every story. This PR wires up the accessible names so the editor, the read-only view, and the one demo textarea all announce what they are.

## Why the violations happened

- The editor pointed `aria-labelledby` at the control id, but no element carries that id. A `<label htmlFor>` association cannot name a contenteditable div, so the name resolved to empty.
- `RichTextView` renders Lexical's `ContentEditable`, which keeps `role="textbox"` (plus `aria-readonly`) even when non-editable, with no way to name it from outside. Swapping the role instead would trade this for a critical `aria-allowed-attr` violation, since Lexical hardcodes `aria-readonly` and only textbox-family roles allow it.
- The Markdown Serializers story rendered a bare `<textarea>` demo input.

## What changed

- `RichTextEditor` passes `labelID` to `Field` and points `aria-labelledby` at the label element; the contenteditable now carries the control id that `htmlFor` expects.
- `RichTextView` gains an optional `label` prop applied as `aria-label` on the text region, defaulting to `'Rich text content'` (same pattern as `RichTextEditorToolbar`'s `label` default).
- The Markdown Serializers story labels its demo textarea; both `RichTextView` story usages pass contextual labels.
- `.github/a11y-baseline.json`: 226 -> 213 entries (deletions only).

## Verification

- 3 new tests (visible-label name on the editor, default and custom name on the view); 53/53 lab tests pass.
- `lab typecheck:docs` and `storybook typecheck` pass.
- Full axe audit (`accessibility-audit.js --components RichTextEditor --fail-on-new` against the trimmed baseline): 0 violations across all 12 stories.

No changeset: lab package (canary only), stories, and CI config.


imdreamrunner (approved):
Reviewed commit 309dde5 and the clean merge result against current main. The visible Field label now names the editor textbox through aria-labelledby, RichTextView has default and overridable accessible names, and the story textarea has a native label.

Verification: focused RichTextEditor suite passes on the PR head (53/53) and on the current-main merge result (79/79); @astryxdesign/lab typecheck:docs passes in both cases. No blocking findings.

cixzhang:
Thanks for this — the diagnosis was right, and all of it has landed, just not here.

The editor half went in with [#4818](https://github.com/facebook/astryx/pull/4818): the contenteditable now carries the control id and `aria-labelledby` points at the Field label. Your own [#5204](https://github.com/facebook/astryx/pull/5204) carries the rest against the moved `richtext` package — the `RichTextView` `label` prop, the story labels, and the same 13 baseline entries — so this one would only be a rebase into a file that no longer exists.

Closing in favour of [#5204](https://github.com/facebook/astryx/pull/5204); happy to reopen if it turns out to miss something here. We're in [Discord](https://discord.com/invite/XnsUcFykEP) if you'd rather talk to a person.

[Reviewed by Robohands]

PR #4688, opened 2026-08-03
fix(core): center directional icons inside controls

zeroryu:
## Summary

- normalize the shared RTL mirror wrapper to centered `inline-flex` layout
- fix the SideNav collapse chevron without changing Button sizing or RTL transforms
- apply the same baseline fix to gallery, tree, and table directional controls
- add the core patch changeset and Next-generated docsite agent rules

## Verification

- `pnpm exec vitest run packages/core/src/SideNav/SideNav.test.tsx packages/core/src/Calendar/Calendar.test.tsx packages/core/src/Carousel/Carousel.test.tsx packages/core/src/Lightbox/Lightbox.test.tsx packages/core/src/TreeList/TreeList.test.tsx` (289 tests)
- `pnpm check:sync`

### DIFF
--- a/apps/docsite/AGENTS.md
+++ b/apps/docsite/AGENTS.md
@@ -1,0 +1,8 @@
+<!-- BEGIN:nextjs-agent-rules -->
+
+# This is NOT the Next.js you know
+
+This version has breaking changes — APIs, conventions, and file structure may all differ from your training data. Read the relevant guide in `node_modules/next/dist/docs/` before writing any code. Heed deprecation notices.
+
+**Keep this block, including in commits.** It is part of the project's agent setup, maintained by `next dev` for every agent that works here. If it appears as an uncommitted change, that is intentional — commit it as-is. Do not remove it to clean up a diff; it will be regenerated.
+<!-- END:nextjs-agent-rules -->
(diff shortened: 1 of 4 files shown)


apps/docsite/AGENTS.md
```diff
@@ -0,0 +1,8 @@
+<!-- BEGIN:nextjs-agent-rules -->
```
imdreamrunner: is this file and apps/docsite/CLAUDE.md intended to be part of this PR>

imdreamrunner (changes requested):
The two new Markdown files don’t seem to be related to this PR. Could you either remove them or explain why you think they’re necessary?

apps/docsite/AGENTS.md
zeroryu: not needed. was part of next js script

cixzhang:
Thanks for the fix! [#4880](https://github.com/facebook/astryx/pull/4880) landed the same correction in the meantime, on SideNav's own styles rather than the shared mirror — I measured the mirrored chevrons across all five call sites and they sit centred on main today. Closing. We're in [Discord](https://discord.com/invite/XnsUcFykEP) if you'd like to talk it through.

[Reviewed by Robohands]

PR #4689, opened 2026-08-03
fix(DateRangeInput): expose trigger as combobox

AKnassa:
Part of #4681 (weekly a11y scan). Clears the critical `DateRangeInput::Required::aria-allowed-attr` baseline entry.

## What this does

The date range picker's trigger button carried `aria-required`, which ARIA does not allow on `role=button`, so axe flagged it as critical. The trigger is now exposed as `role="combobox"`, the same decision DateInput and DateTimeInput already made for their triggers, which makes `aria-required` and `aria-invalid` legal and announces the control as what it behaves like: a collapsed picker that opens a dialog.

## What changed

- `role="combobox"` on the trigger button in `DateRangeInput.tsx`. The existing `aria-expanded`, `aria-haspopup="dialog"` and `aria-controls` wiring already matches the pattern.
- Trigger test queries move from button-role to combobox-role. The accessible name (label plus value or placeholder) is pinned by name-matched assertions and is mutation-proven: removing the trigger's `aria-label` fails 2 tests.
- Baseline: 1 entry removed, deletions only.
- Changeset (patch): consumers who query the trigger with `getByRole('button', {name})` need to target `combobox` instead; screen readers now announce "combobox, collapsed" rather than "button".

## Considered and rejected

Dropping `aria-required` from the button was the smaller diff, but it silently removes the required announcement and diverges from both sibling date inputs.

## Not in this PR

The 10 remaining `aria-allowed-attr` entries (ChatComposerInput x8, ChatLayout x2) all trace to one node: the composer's editable div flips to `role=combobox` when `triggers` are configured while keeping a hardcoded `aria-multiline`. That file is owned by open PR #4074; the fix belongs there and the entries stay baselined until it lands.

## Verification

41/41 DateRangeInput tests (plus width-contract and status-icon suites, 85/85), eslint clean, and the repo's axe audit with `--fail-on-new`: 0 issues across all 17 DateRangeInput stories.

### DIFF
--- a/packages/core/src/DateRangeInput/DateRangeInput.tsx
+++ b/packages/core/src/DateRangeInput/DateRangeInput.tsx
@@ -590,6 +590,9 @@ export function DateRangeInput({
           ref={ref}
           id={id}
           type="button"
+          // Select-only combobox, matching DateInput/DateTimeInput: makes
+          // aria-required/aria-invalid valid here (not allowed on role=button).
+          role="combobox"
           onClick={handleToggle}
           // With a disabledMessage the trigger keeps focusability via
           // aria-disabled so the reason is focus-discoverable; activation is
--- a/.changeset/daterangeinput-trigger-combobox.md
+++ b/.changeset/daterangeinput-trigger-combobox.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] DateRangeInput: expose the trigger as `role="combobox"` (matching DateInput/DateTimeInput) so `aria-required` is valid — it is not allowed on `role=button` (axe aria-allowed-attr, critical). Tests locating the trigger via `getByRole('button', {name})` must switch to `getByRole('combobox')`.
+@AKnassa
--- a/.github/a11y-baseline.json
+++ b/.github/a11y-baseline.json
@@ -534,11 +534,6 @@
       "helpUrl": "https://dequeuniversity.com/rules/axe/4.12/color-contrast?application=playwright"
     },
     {
-      "key": "DateRangeInput::Required::aria-allowed-attr",
-      "impact": "critical",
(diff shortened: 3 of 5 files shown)


packages/core/src/DateRangeInput/DateRangeInput.tsx
```diff
@@ -590,6 +590,9 @@ export function DateRangeInput({
           ref={ref}
           id={id}
           type="button"
+          // Select-only combobox, matching DateInput/DateTimeInput: makes
+          // aria-required/aria-invalid valid here (not allowed on role=button).
+          role="combobox"
```
cixzhang, line 595: The role now promises ArrowDown opens the popup; this button has no key handler.

cixzhang (changes requested):
Thanks — routing through `combobox` rather than dropping `aria-required` is the right call.

One gap before it lands. `DateInput` implements the keyboard half of that pattern (`DateInput.tsx:663` — ArrowDown and Alt+ArrowDown open the calendar); this trigger has no `onKeyDown` at all. So a screen-reader user hears "combobox, collapsed", presses Down Arrow because that is what the role promises, and nothing happens — where the same press on `DateInput` opens the picker. Enter and Space still work, so it is incompleteness rather than a break.

```tsx
if (!popover.isOpen && e.key === 'ArrowDown') { e.preventDefault(); handleToggle(); }
```

The popup here is a dialog rather than a listbox — should ArrowDown move focus into the calendar, which `handleToggle` already does, or keep it on the trigger the way `DateInput` does with `skipAutoFocus`?

[Reviewed by Robohands]

cixzhang:
Checking in on the combobox keyboard gap: Down Arrow still needs a defined open-and-focus behavior matching the announced role. Would you like to continue, have us carry it, or close this for now?

PR #4690, opened 2026-08-03
feat(Table): treegrid semantics and row-focus keyboard model for the tree plugin

AKnassa:
Part of #4681 (weekly a11y scan). Clears the 8 `aria-conditional-attr` baseline entries for TableTree. The 3 TableGroupedRows entries move to their own PR, #5612, as agreed in review.

## What changed since the last round

Review asked for a treegrid instead of stripping the row ARIA, and the [ruling](https://github.com/cixzhang/astryx/wiki/Rulings#decided) set v1 scope to row focus. This round builds that:

- **Role.** With any expandable row, `useTableTreeData` names the `<table>` `role="treegrid"` via `transformTable`. Flat data keeps the native table (no role, no row ARIA, no tab stop): the migration guarantee holds.
- **Row ARIA.** Rows keep `aria-level` (1-based) and `aria-expanded` (parents only), and gain `aria-posinset` / `aria-setsize` from the sibling group as flattened by `useTableTreeState` (after `sortSiblings`). `TableTreeRowMeta` carries them as optional `posInSet` / `setSize`, so a hand-built config still typechecks.
- **Keyboard, row focus.** Composed from `useTreeFocus` (the shared tree primitive) pointed at the rows through its `itemSelector`: one roving tab stop across the visible rows; ArrowUp/ArrowDown move between rows; ArrowRight expands a collapsed row or enters its first child; ArrowLeft collapses an expanded row or moves to the parent; Home/End; Enter/Space toggle an expandable row; RTL flips the horizontal arrows (the hook already does). Focus survives the expand/collapse re-render because rows are keyed by id. Typeahead is off: not part of the treegrid pattern, and printable keys must reach a text field inside a cell.
- **Interactive descendants stay reachable.** The tree keys apply only when a row itself owns focus: a key pressed on the chevron, a selection checkbox, a sort header, or a text field keeps its native meaning. Those controls also stay in the Tab order (the row is one extra Tab stop ahead of them). That is the v1 contract the ruling describes; an entry mode or cell navigation is a separate decision.
- **Focus ring.** Rows draw the shared `focusOutlineStyles.focusVisible` ring, inset so it is not clipped by neighbouring rows.
- **Docs** name the v1 limitation (`useTableTreeData.doc.mjs` description, `usage` best practices, `hasRowClickExpansion`), and the Default story's JSDoc summarises the keys.

Two supporting changes outside the plugin:

- `TableRenderProps.ref`: a plugin can now attach a ref to the `<table>`; `BaseTable` merges it with the consumer's `ref` (memoised, so callback refs are not re-attached every render). Mirrors `BodyRowRenderProps.ref` / `ScrollWrapperRenderProps.htmlProps.ref`.
- `useTreeFocus` resolves the focused item with `closest(itemSelector)` instead of a hardcoded `[role="treeitem"]`. Without this, a custom `itemSelector` listed the rows but ArrowDown from a focused row went back to the first row (the focused item resolved to `null`). Default behaviour is unchanged; TreeList is unaffected.

## Why useGridFocus is not in the composition

`useGridFocus` is cell geometry (`columns`, a cell selector, Ctrl+Home/End across cells). With row focus there are no cells to navigate, so v1 composes on `useTreeFocus` alone. When cell navigation is decided, that is where the grid hook comes in.

## Judgment calls, stated plainly

- Enter/Space toggle any expandable row (the primitive's default), not only when `hasRowClickExpansion` is set; `hasRowClickExpansion` stays a pointer-only convenience.
- Space on a focused row toggles expansion, not selection. In the selection composition, selection stays on the checkbox (Tab-reachable). Flagging in case Space should be reserved.
- Changeset is `patch` (the repo rule: only `[breaking]` bumps minor while 0.x), but the entry spells out what consumers will notice: the rows become a Tab stop, and `getByRole('table')` becomes `treegrid` for tree tables.

## Verification

- Tests first (red), then the implementation: 18 new tests (treegrid role, posinset/setsize, roving tabindex, every key, the descendant guard, RTL, no typeahead, ref composition), all green; 941/941 across `Table`, `hooks`, `TreeList`.
- `core typecheck`, `core typecheck:docs`, `cli typecheck:strict`, `storybook typecheck` clean; eslint clean on the changed files (one pre-existing warning on main's `BaseTable.tsx:489`).
- axe audit with `--fail-on-new`, TableTree + TableGroupedRows: **0 new; 0 issues on all 9 TableTree stories**; TableGroupedRows shows only its 3 baselined entries (the split PR). The audit also reports `TableGroupedRows::Default::color-contrast` and `::Initially Collapsed::color-contrast` as resolved; left for a contrast round.

Checked in Chrome on the built Storybook (Default story): the `<tr>` paints the inset ring on keyboard focus, ArrowRight expands in place and then enters the first child, ArrowLeft returns to the parent, Tab from the row reaches its chevron, and arrows on the chevron do nothing. Observation: the scroll wrapper is already a Tab stop (pre-existing), so a tree table reads wrapper, then row, then in-cell controls; whether the wrapper stop should yield once rows are focusable is a separate question.

Follow-ups this PR does not claim: cell navigation, an entry mode for in-cell controls.


packages/core/src/Table/plugins/tree/useTableTreeData.test.tsx
```diff
@@ -317,23 +317,20 @@ describe('useTableTreeData — row-click expansion', () => {
 // =============================================================================
 
 describe('useTableTreeData — row ARIA', () => {
-  it('sets 1-based aria-level on every body row', () => {
+  it('never sets aria-level on rows — the indent conveys hierarchy (axe aria-conditional-attr: rows outside a treegrid)', () => {
```
cixzhang: Hmm well this seems wrong. If we include tree data the table should probably be a treegrid.

cixzhang (changes requested):
I think the fix might be to apply the treegrid row when including tree data. We will likely need to make a version of useTreeGridFocus to achieve the right behaviors. It's quite a bit though. Let me know if you wanna do it!

AKnassa:
Yes, happy to take it on.

Agreed treegrid is the right semantics. The row ARIA removed here was already treegrid shaped (1-based aria-level, aria-expanded on parents only), the container was the invalid part. Rough plan:

- `role="treegrid"` via `transformTable`, applied only when expandable rows exist so flat data keeps the plain table.
- Restore aria-level/aria-expanded on rows and add aria-posinset/aria-setsize, computed from the flattened visible list.
- New `useTreeGridFocus`: roving tabindex with row focus, Up/Down across visible rows, Right expands, Left collapses or jumps to parent, Home/End, Enter toggles where row-click expansion is on, focus preserved across expand/collapse re-renders. Role and keyboard ship together, a treegrid without arrow support would regress below the plain table.

Two questions before I start:

1. Do you have a reference `useTreeGridFocus` I should mirror? Related: is row focus enough for v1, or do you want cell navigation too? Either way the table becomes one tab stop, so in-cell widgets (sort headers, chevrons, selection checkboxes) stop being individually tabbable.
2. groupedRows is not tree data, and its change here just leaves the expanded state on the group toggle button. OK to split it into its own PR and keep this branch for the treegrid work?

cixzhang:
Sorry for the wait on this.

On a reference: `useTreeFocus` and `useGridFocus` are both public hooks with docs and tests — `packages/core/src/hooks/useTreeFocus.ts` and `useGridFocus.ts`. A treegrid version is those two composed, so they're the ones to mirror.

Row focus versus cell navigation I don't want to answer off the cuff. You've named the cost exactly right — the table becomes one tab stop and sort headers, chevrons and selection checkboxes stop being individually tabbable — and that's a call for a maintainer rather than a review. It's queued.

Yes to splitting groupedRows out. It isn't tree data, moving the expanded state onto the chevron button stands on its own, and it clears three axe entries; no reason for it to wait behind the treegrid work.

[Reviewed by Robohands]

gonzoblasco:
On the row-focus vs cell-navigation question that's queued with maintainers - here's the AT angle that might help the decision, since it's the part that's easy to miss when reasoning about tab stops alone.

The two options aren't just "how many tab stops" - they map to different ARIA contracts, and the in-cell widgets decide which is viable:

**Row focus (one tab stop per row) is closer to a list/tree semantics than a grid.**
With roving tabindex at the row level, the row is the focusable unit and sort headers / chevrons / selection checkboxes inside it are no longer individually in the tab order. That means those widgets need a *second* mechanism to be reachable at all - typically Enter/F2 to "enter" the row, then Tab within it, then Escape to leave (the composite-widget entry pattern from APG). Without that, you've traded individually-tabbable controls for a row that swallows them - which is the exact cost @AKnassa named, but the fix isn't automatic: it requires the composite entry model, not just moving the tab stop.

**Cell navigation (roving per-cell) matches the APG grid pattern directly.**
In a `treegrid`, each cell is focusable with roving tabindex, and arrow keys move between cells; widgets *inside* a cell stay reachable because the cell (not the row) is the focused unit, and Tab moves into the widget from the focused cell. This is the pattern the APG documents for grids, and it's what `aria-rowindex`/`aria-colindex` semantics assume. The cost is the opposite one: more moving parts in the keyboard model, and a heavier interaction for users who just want to tab through.

So the deciding question isn't "row vs cell" in the abstract - it's: **do the tree rows carry in-cell interactive widgets (sort headers, chevrons, selection) as part of the treegrid contract?** If yes, cell navigation (or a row model with an explicit composite-entry mechanism) is the only way those stay reachable. If the treegrid v1 rows are plain data cells with the expander as the sole interaction, row focus is simpler and defensible - and the in-cell widgets can be explicitly out of scope for v1, documented, and revisited when they're added.

Either way, the keyboard model and the ARIA indexes should match whatever is chosen - a row-focus model emitting per-cell `aria-colindex` would contradict itself, and a cell model with row-level roving would too.

cixzhang:
Thanks for waiting. The decision window has closed: row focus is the v1 scope, with the docs naming that full cell navigation is outside v1. Existing interactive descendants must remain reachable; combinations that need cell navigation stay outside the supported v1 contract.

You can proceed by composing the existing tree and grid focus primitives. [Decision record](https://github.com/cixzhang/astryx/wiki/Rulings#decided)

[Reviewed by Robohands]

gonzoblasco:
On the v1 row-focus scope and the "existing interactive descendants must remain reachable" requirement - one concrete way to satisfy that without pulling in the full composite-entry pattern:

In a row-focus model, the row is the single tab stop and the in-cell widgets (sort headers, chevrons, selection checkboxes) drop out of the tab order. The APG composite-widget entry pattern (Enter/F2 to enter the row, Tab within, Escape to leave) is the full solution, but it's heavy for a v1. A lighter option that keeps the widgets reachable: keep the row as the tab stop, and let the widgets inside remain reachable via the screen reader's virtual cursor / browse mode even though they're not in the sequential tab order. That's the key distinction: "reachable" for AT doesn't have to mean "in the tab order" - browse mode can still get to them.

The trap to avoid is the middle ground: making the row focusable but leaving the widgets *also* in the tab order, which creates two competing tab stops per row and a confusing keyboard model. Either the row is the single stop (widgets reachable via browse mode / composite entry), or the widgets stay individually tabbable (which is the cell-nav model, out of v1). The one thing that doesn't work is both.

So for v1: row as the single tab stop, widgets reachable via browse mode, and document that composite entry (Enter/F2) is the follow-up if keyboard-only users need direct widget access. That keeps the v1 contract clean and the widgets reachable, without the full composite-entry machinery.

AKnassa:
Rebuilt on the ruling: row focus for v1, composed on useTreeFocus over the rows (useGridFocus is cell geometry, so it enters when cell navigation is decided). Rows keep aria-level/aria-expanded and gain aria-posinset/aria-setsize; the table is role="treegrid" only while a row is expandable, flat data is untouched. The tree keys apply only when a row itself owns focus, so the chevron, checkboxes and sort headers keep their own keys and stay in the Tab order. The docs name the v1 limit.

Two small supporting changes: TableRenderProps.ref (merged with the consumer ref in BaseTable), and useTreeFocus now resolves the focused item through itemSelector instead of a hardcoded role="treeitem", which is what let the table compose it. groupedRows is split out to #5612. Audit with --fail-on-new: 0 issues on all 9 TableTree stories.
1-treegrid-row-focus-ring
2-treegrid-expanded-chevron-focused

rubyycheung (changes requested):
Thanks for addressing the treegrid semantics. I do not think the keyboard model is ready yet.

The current arrow behavior follows a tree pattern rather than the [ARIA treegrid pattern](https://www.w3.org/WAI/ARIA/apg/patterns/treegrid/): Right Arrow on an expanded row should move into its first cell, not its first child row; Left Arrow on a collapsed or leaf row should do nothing, not move to its parent. If this divergence is intentional, could you explain and document the rationale?

The existing scroll wrapper also creates a redundant focus stop before the row.

Finally, the focus treatments are visibly inconsistent. Rows use Astryx’s shared accent ring, while the raw chevron buttons fall back to the browser’s default outline color.

Could these issues be addressed before another review?

humbertovirtudes (changes requested):
Consumer root event props can overwrite the plugin role and keyboard handlers, programmatic row focus does not move the roving tab stop, and the arrow model follows a tree rather than the APG treegrid cell-navigation pattern. The wrapper, rows, and descendant controls also create competing sequential focus systems. Please define event composition and one coherent treegrid focus contract, then add focused regression tests and browser proof.

PR #4691, opened 2026-08-03
fix(lab): add tab stops to scrollable regions

AKnassa:
Part of #4681 (weekly a11y scan). Clears all 6 `scrollable-region-focusable` baseline entries (CodeEditor, CodeEditorPerf x2, Outline, Schedule, Stack).

## What this does

Scrollable containers with no focusable content are unreachable for keyboard users. Two component fixes in lab and two story-level fixes make each flagged scroll region keyboard-reachable.

## What changed

- **CodeEditor** (`packages/lab`): explicit `tabIndex={0}` on the contenteditable code element. Editable editors already had an effective tab stop (this just materializes the attribute axe inspects); read-only editors genuinely gain one, matching native readonly `<textarea>` behavior, so keyboard users can reach and scroll them. That read-only change is intentional and pinned by a test. Covers all three CodeEditor/CodeEditorPerf entries.
- **Schedule monthly view** (`packages/lab`): `tabIndex={0}` on the month grid, which already carries `role="grid"`, `aria-readonly` and a label. The grid has no interactive descendants today; if day cells or event pills ever become interactive, this should convert to a roving-tabindex pattern (recorded here so the constraint is known).
- **Outline "Scroll Spy" and Stack "Scrollable" stories**: the scrollable panes are story-made demo divs, so they get `tabIndex`, `role="region"` and an `aria-label` at the story level. Stack itself is untouched on purpose: a generic layout primitive should not grow a tab stop by default, and that API call belongs to maintainers if ever wanted.
- Baseline: 6 entries removed, deletions only. No changeset: lab is unpublished-stable (canary) and the rest is stories.

## Note on Schedule::Monthly

The violation does not reproduce at the local audit viewport (it fires on CI's). The unit test pins the attribute mechanism either way; the weekly scan will confirm on CI.

## Verification

Full lab suite 388/388 and full core suite 5786/5786, storybook typecheck, eslint clean, and the repo's axe audit with `--fail-on-new`: zero scrollable-region-focusable occurrences across all five components' stories.

### DIFF
--- a/.github/a11y-baseline.json
+++ b/.github/a11y-baseline.json
@@ -359,11 +359,6 @@
       "helpUrl": "https://dequeuniversity.com/rules/axe/4.12/color-contrast?application=playwright"
     },
     {
-      "key": "CodeEditor::With Max Height::scrollable-region-focusable",
-      "impact": "serious",
-      "helpUrl": "https://dequeuniversity.com/rules/axe/4.12/scrollable-region-focusable?application=playwright"
-    },
-    {
       "key": "CodeEditor::With Placeholder::color-contrast",
       "impact": "serious",
       "helpUrl": "https://dequeuniversity.com/rules/axe/4.12/color-contrast?application=playwright"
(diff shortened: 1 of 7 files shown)


cixzhang (approved):
Thanks for this. Making read-only editors focusable is the right call, and keeping the fix at story level for Stack is the right restraint. Would merge as is.

[Reviewed by Robohands]

PR #4692, opened 2026-08-03
fix(a11y): unique names for repeated nav landmarks

AKnassa:
Part of #4681 (weekly a11y scan). Clears all 8 `landmark-unique` baseline entries (Layout, Pagination, TabList x2, TablePagination x2, Toolbar, ToolbarEdgeCompensation).

## What this does

Same-type landmarks need distinguishable accessible names. The flagged stories render several `<nav>` landmarks with identical default labels; one case is a real component gap. Both get fixed at the level they belong to.

## What changed

- **Story-level (6 files, existing props only)**: the Layout, Pagination, TabList, Toolbar, ToolbarEdgeCompensation and TablePagination matrix stories now pass distinct labels to each instance (for example `aria-label={'Tabs (' + size + ')'}`). No component behavior changes.
- **Component-level** (`useTablePagination`, packages/core): with `position='both'`, one hook call rendered two navs sharing one name, which a story cannot fix. The two navs now get distinct names by interpolating the label into two new i18n keys: `"{label} (top)"` and `"{label} (bottom)"`. Single-position accessible names stay byte-identical. New keys live in en.json only (fr-FR stays a sparse demo catalog); the base key's translator note now points at the suffix templates.
- The previously undocumented `label` config is now in the doc.mjs (EN and dense variants), including the position='both' naming contract.
- Baseline: 8 entries removed, deletions only. Changeset (patch): the position='both' nav names change for assistive technology; the old duplicate-name behavior was itself the violation.

## Judgment call, stated plainly

The "(top)" / "(bottom)" suffix wording is a choice. Alternatives considered: standalone non-interpolated keys (loses the consumer-supplied label), or suffixing only one nav (asymmetric announcements). Interpolated templates keep consumer labels differentiated and translatable as a unit. Happy to reword if you prefer different suffixes.

## Verification

43/43 pagination plugin tests (the old test asserting two same-named navs is rewritten, plus a consumer-label interpolation test), 147/147 across the wider pagination/Pagination/i18n slice, core typecheck, eslint clean, and the repo's axe audit with `--fail-on-new` across all six components: exit 0, "0 new, 13 resolved".


cixzhang (approved):
Thanks — distinct names for both nav landmarks, the two baseline entries dropped, pr-a11y green. Landing.

[Reviewed by Robohands]

cixzhang:
Approved above — it needs a merge with main before it can land (conflicting right now). Ping me once it's clean and I'll merge.

[Reviewed by Robohands]

cixzhang:
Merged — thank you, @AKnassa, for giving both nav landmarks distinct accessible names and for clearing the stale `landmark-unique` entries out of the baseline while you were in there. Your branch had gone stale against `main`, so we merged `main` in on your behalf rather than asking you to rebase: the only conflicts were mechanical (the a11y baseline, where both sides had removed entries, and `Pagination.stories.tsx`, where `main` had independently labelled the same variants). Sorry it sat for 20 days — that wait was on us, not on you.

[Reviewed by Robohands]

PR #4693, opened 2026-08-03
fix(Table): name the row-status column header

AKnassa:
Part of #4681 (weekly a11y scan). Clears both `TableRowStatus::*::empty-table-header` baseline entries.

## What this does

The row-status plugin renders a narrow colored-gutter column whose `<th>` was intentionally blank, leaving screen reader users a nameless column header. The header now carries visually hidden text: sighted users still see a blank gutter, assistive technology announces "Row status".

## What changed

- `useTableRowStatus` renders `<VisuallyHidden>` text from a new i18n key (`@astryx.table.rowStatus.columnHeader`, default "Row status") instead of an empty header, with `t` added to the column memo deps (same shape as the rowExpansion plugin's strings).
- Override goes through `InternationalizationProvider` like every other built-in Table plugin string; deliberately no new config prop.
- "Row status" over plain "Status" avoids colliding with common visible column names (the stories themselves have a "State" column).
- Tests pin both the accessible name and that it stays visually hidden (bare visible `th` text would fail).
- Baseline: 2 entries removed, deletions only. Changeset (patch).

## Known relatives, not in this PR

`useTableRowExpansion` has the identical blank-header pattern but zero baseline entries (its stories appear to fall outside the scan), and `Table/types.ts` documents `header: ''` as the example for action columns. Both are follow-up candidates.

## Verification

9/9 plugin tests plus the i18n catalog guard (29/29 combined), eslint clean, and the repo's axe audit with `--fail-on-new`: 0 issues on both TableRowStatus stories.

### DIFF
--- a/.changeset/a11y-table-row-status-header.md
+++ b/.changeset/a11y-table-row-status-header.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Table useTableRowStatus: the status gutter's column header now carries a visually hidden localized name ("Row status", key `@astryx.table.rowStatus.columnHeader`) instead of an empty `<th>` (axe empty-table-header, WCAG 1.3.1 best practice). The gutter stays visually blank.
+@AKnassa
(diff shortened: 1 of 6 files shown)


cixzhang (approved):
Thanks — this is the same shape `useTableRowExpansion` uses, and the clipped span keeps the gutter visually blank. Would merge as is.

[Reviewed by Robohands]

PR #4696, opened 2026-08-04
refactor(i18n): pre-build lookup map for translations

ngolin:
Memoize the translation lookup map and expose it via context instead of resolving keys on each call. This removes per-call chain resolution and the `useCallback` wrapper, improving performance for frequent translations.

### DIFF
--- a/packages/core/src/i18n/InternationalizationContext.ts
+++ b/packages/core/src/i18n/InternationalizationContext.ts
@@ -23,6 +23,7 @@
  */
 
 import {createContext} from 'react';
+import {getResolve} from './resolve';
 import type {Locale, MessagesByLocale, Overrides} from './types';
 
 export interface InternationalizationContextValue {
@@ -30,6 +31,7 @@ export interface InternationalizationContextValue {
   direction: 'ltr' | 'rtl';
   messages: MessagesByLocale;
   overrides?: Overrides;
+  translate: ReturnType<typeof getResolve>;
 }
 
 /**
@@ -41,5 +43,6 @@ export const InternationalizationContext =
     locale: 'en',
     direction: 'ltr',
     messages: {},
+    translate: getResolve('en', {}),
   });
 InternationalizationContext.displayName = 'InternationalizationContext';
--- a/packages/core/src/i18n/resolve.ts
+++ b/packages/core/src/i18n/resolve.ts
@@ -73,74 +73,71 @@ export function resolveLocaleChain(locale: Locale): Locale[] {
   return chain;
 }
 
-function lookup(
-  key: string,
+function getLookup(
   locale: Locale,
   messages: MessagesByLocale,
-  overrides: Overrides | undefined,
-): string | null {
+  overrides?: Overrides,
+): Record<string, string> {
+  const lookup: Record<string, string> = {};
   const chain = resolveLocaleChain(locale);
-
   // 1 + 2. Overrides (most specific to least specific in the chain)
   if (overrides !== undefined) {
     for (const tag of chain) {
-      const value = overrides[tag]?.[key];
-      if (value !== undefined) {
-        return value;
+      if (overrides[tag]) {
+        for (const [key, value] of Object.entries(overrides[tag])) {
+          if (lookup[key] === undefined && value !== null) {
+            lookup[key] = value;
+          }
+        }
+      }
+    }
+  }
+  for (const tag of chain) {
+    if (messages[tag]) {
+      for (const [key, value] of Object.entries(messages[tag])) {
+        if (lookup[key] === undefined && value?.defaultMessage !== null) {
+          lookup[key] = value?.defaultMessage;
+        }
       }
     }
   }
 
-  // 3 + 4. Shipped catalogs (most specific to least specific)
-  for (const tag of chain) {
-    const entry = messages[tag]?.[key];
-    if (entry !== undefined) {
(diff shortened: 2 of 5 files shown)


packages/core/src/i18n/resolve.ts
```diff
@@ -73,74 +73,71 @@ export function resolveLocaleChain(locale: Locale): Locale[] {
...
+  const lookup = {} as Record<string, string>;
   const chain = resolveLocaleChain(locale);
-
   // 1 + 2. Overrides (most specific to least specific in the chain)
-  if (overrides !== undefined) {
-    for (const tag of chain) {
-      const value = overrides[tag]?.[key];
-      if (value !== undefined) {
-        return value;
+  for (const tag of chain) {
```
cixzhang: Per-tag fill: `messages['pt-BR']` now beats `overrides['pt']`. The comment above still describes the old order.

packages/core/src/i18n/InternationalizationContext.ts
```diff
@@ -39,5 +41,6 @@ export const InternationalizationContext =
     locale: 'en',
     direction: 'ltr',
     messages: {},
+    translate: getResolve('en', {}),
```
cixzhang, line 46: Materialises all 305 `en` keys at module load, for every core consumer.

cixzhang (changes requested):
Thanks for looking at this — walking the chain per key is genuinely repetitive and worth attacking.

The blocker is the lookup order. The file documents overrides (exact locale, then parent) as beating every shipped catalog entry; the new map fills per tag, overrides then messages, so a shipped `pt-BR` string now wins over a `pt` override. An app that overrides a string for Portuguese loses it silently the first time the shipped `pt-BR` catalog gains that key — nothing throws, the wording just reverts.

```js
getResolve('pt-BR', {'pt-BR': {k: {defaultMessage: 'shipped'}}}, {pt: {k: 'override'}})('k')
// main: 'override'   this branch: 'shipped'
```

Two passes over the chain — every override first, then every catalog — restores it.

Separate and not blocking: the module-level `getResolve('en', {})` builds all 305 `en` keys at import time, for every consumer of core whether they render a translated string or not.

[Reviewed by Robohands]

cixzhang:
Checking in on locale lookup ordering: parent-locale overrides still need to beat exact shipped catalog strings. Would you like to continue, have us carry it, or close this for now?

ngolin:
> Checking in on locale lookup ordering: parent-locale overrides still need to beat exact shipped catalog strings. Would you like to continue, have us carry it, or close this for now?

Yes. I will make a quick fix and continue. Thank you for the feedback.

packages/core/src/i18n/InternationalizationContext.ts
ngolin: Fixed. The default `EN_CATALOG` is no longer processed inside `getLookup`; this keeps `getResolve` lightweight so it doesn't affect the performance of core customers. `EN_CATALOG` will act as the fallback during actual `resolve`:

```ts
const result = lookup[key] ?? EN_CATALOG[key]?.defaultMessage;
```

packages/core/src/i18n/resolve.ts
ngolin: Fixed.

ngolin:
> Thanks for looking at this — walking the chain per key is genuinely repetitive and worth attacking.
> 
> The blocker is the lookup order. The file documents overrides (exact locale, then parent) as beating every shipped catalog entry; the new map fills per tag, overrides then messages, so a shipped `pt-BR` string now wins over a `pt` override. An app that overrides a string for Portuguese loses it silently the first time the shipped `pt-BR` catalog gains that key — nothing throws, the wording just reverts.
> 
> ```js
> getResolve('pt-BR', {'pt-BR': {k: {defaultMessage: 'shipped'}}}, {pt: {k: 'override'}})('k')
> // main: 'override'   this branch: 'shipped'
> ```
> 
> Two passes over the chain — every override first, then every catalog — restores it.
> 
> Separate and not blocking: the module-level `getResolve('en', {})` builds all 305 `en` keys at import time, for every consumer of core whether they render a translated string or not.
> 
> If you'd rather talk it through with someone, we're in [Discord](https://discord.com/invite/XnsUcFykEP).
> 
> [Reviewed by Robohands]

Thanks for the detailed review. Both issues are addressed.

**Issue 1** — Override precedence (order bug)

`getLookup` now walks the locale chain in two passes — all overrides first (exact → parent), then all catalogs (exact → parent) — instead of interleaving them per tag. Your repro now resolves to the `override`:

```js
getResolve('pt-BR', {'pt-BR': {k: {defaultMessage: 'shipped'}}}, {pt: {k: 'override'}})('k')
// 'override' ✓
```
> This repro is locked in by a regression test in `resolve.test.ts`: ``"override `pt` should take precedence over message `pt-BR`"``.


**Issue 2** — eager en catalog build at import time

`EN_CATALOG` is no longer materialized inside `getResolve('en', {})` at module load. `getResolve` stays cheap, so every core consumer pays almost nothing at import regardless of whether they ever render a translated string. `EN_CATALOG` is used only as the final fallback inside `resolve`:

```js
const result = lookup[key] ?? EN_CATALOG[key]?.defaultMessage;
```

Thanks again for the thorough review — happy to discuss further if anything else comes up.

cixzhang (approved):

PR #4697, opened 2026-08-04
feat(table): reflect density on cell theme targets for per-density padding

freddymeta:
## What

`Table` applies its density padding (`compact`/`balanced`/`spacious`) through **internal StyleX classes** — both the block and inline padding vary per density, baked into `densityStyles`. The `astryx-table-cell` / `astryx-table-header-cell` theme targets were rendered with `themeProps('table-cell')` — **no `density` argument** — so the stable class carried no `data-density` (or `.compact`/`.balanced`/`.spacious`) hook.

The consequence: a `defineTheme` `components: { 'table-cell': {...} }` entry could only target the **base** class, setting one padding for *all* densities at once. There was no way to override padding **per density** — e.g. to hold the inline inset constant while varying only the block padding across densities. That split lives in unreachable atomic classes.

This PR passes the active density to `themeProps` on both cell targets:

```ts
themeProps('table-cell', {density: ctx?.density})
themeProps('table-header-cell', {density: ctx?.density})
```

so each cell now reflects `data-density` and the matching class hook, and documents `visualProps: ['density']` on both targets.

## Why

Per-density cell padding is a common theming need (different design systems weight the block vs inline axes differently). Today it's impossible via `defineTheme` because the density variants aren't exposed on the cell target — only the parent class is reachable. This closes that gap the sanctioned way (reflecting a visual prop, same pattern other components use), rather than forcing a raw-CSS descendant selector into internal StyleX classes.

`ctx?.density` is nullish-safe: a cell rendered outside a `Table` has no context, so no `data-density` is emitted and nothing changes for standalone use.

## Example

Now possible — hold the inline inset at 8px, vary only block per density:

```ts
defineTheme({
  name: 'my-theme',
  components: {
    'table-cell': {
      'density:compact':  {paddingBlock: '8px',  paddingInline: '8px'},
      'density:balanced': {paddingBlock: '12px', paddingInline: '8px'},
      'density:spacious': {paddingBlock: '16px', paddingInline: '8px'},
    },
  },
});
```

Verified end-to-end: `astryx theme build` emits scoped rules in `@layer astryx-theme` — `.astryx-table-cell.compact { padding-block: 8px; padding-inline: 8px }`, `.balanced`, `.spacious`, and the header equivalents.

## Testing

- `packages/core` Table suite + `themingTargets` guard pass (399), including a new case asserting cells/header cells reflect `data-density` per the `density` prop.
- Typecheck and lint clean.
- Docs: `visualProps: ['density']` added to both cell targets in `Table.doc.mjs`; changeset included. Purely additive — default padding is unchanged.


freddymeta:
@cixzhang — additive theming-target fix, same shape as the recent per-element target work.

`Table`'s density padding lives entirely in internal StyleX classes, and `themeProps('table-cell')` was called with no `density` arg — so the `astryx-table-cell` / `astryx-table-header-cell` targets had no `density:*` hook. A `defineTheme` `components: { 'table-cell': {...} }` could only set one padding for all three densities; it couldn't vary padding per density (e.g. hold the inline inset constant while changing only the block). This passes `{density: ctx?.density}` to both cell targets and documents `visualProps: ['density']`, so per-density overrides work.

Verified `astryx theme build` emits `.astryx-table-cell.compact/.balanced/.spacious` in `@layer astryx-theme`. Nullish-safe outside a Table. Purely additive — default padding unchanged. 399 tests + guard pass.

_— on behalf of Freddy_

cixzhang (approved):

Issue #4652, opened 2026-08-02
[Bug] Widespread WCAG 2.1 AA colour contrast failures across Design System


chris-gds:
### Description

Incomplete : https://github.com/facebook/astryx/issues/3654

Severity: High / Blocker (Prevents accessible product releases)
Component(s) Affected: Core Colour Palette, Buttons, Typography, Form Fields, Navigation
Specification Targeted: WCAG 2.1 Compliance Level AA

## Problem description

A review of the new design system component specs reveals widespread colour contrast failures under WCAG 2.1 Level AA standards. Multiple core colour pairings (text-on-background and UI components) fall below the required contrast ratios, impacting readability for disabled users including those with low vision, colour blindness, or those viewing screens in high-ambient light environments. 

## Impact

Products built with this design system will fail standard accessibility audits, creating legal liability under the ADA, European Accessibility Act (EAA), and other regional mandates.

### Audit & darken/lighten core tokens

- Adjust the luminosity of the core brand palette shades to ensure compliant pairings 
- Automate Contrast Checks: integrate a linter or CI/CD check to flag non-compliant colour pairings before they are pushed to the codebase.
- Hire an accessibility engineer ;)

### New badge

<img width="1094" height="742" alt="Screenshot of the Colour Contrast Analyser app evaluating a light green 'New' badge (#91C889) set against a dark olive background (#686D57) from a UI card titled 'Weekly product highlights. The tool displays a failing contrast ratio of 2.8:1, showing red Fail status icons for WCAG 2.1 AA and AAA text and non-text contrast. In the background, the UI layout includes a landscape banner image and a bottom section featuring a photo of a woman smiling while using a tablet at a desk" src="https://github.com/user-attachments/assets/b75bc409-8aba-4ba6-b71e-8d6e08a32341" />

## Pagination dots (this is also a use of colour fail)

https://astryx.atmeta.com/components/Pagination

<img width="1094" height="742" alt="Image" src="https://github.com/user-attachments/assets/9abe3592-578b-452b-abfd-f8085f2105f0" />

## Anything else

There are many more but it would take significant time to properly audit. Would recommend a complete review, including for themes too as it would apply.

### Reproduction

1. Open the Design System
2. Inspect the core UI components
3. Use an accessibility evaluation tool to test foreground text tokens against their designated background tokens.
4. Note that a significant portion do not clear the $4.5:1$ threshold.

### Astryx Version

@astryxdesign/core@v0.2.0

[...]

PR #4703, opened 2026-08-04
fix(core): pagination dots WCAG contrast (1.4.11) and use-of-color (1.4.1)

gonzoblasco:
## Summary

Fixes the pagination **dots** variant contrast and use-of-color failures reported in #4652 (which reopens #3654).

### 1. WCAG 1.4.11 - Non-text contrast (inactive dots)

The inactive dots used `--color-neutral` (`rgba(5, 54, 89, 0.1)` light / `rgba(223, 226, 229, 0.2)` dark) as a translucent background. Because it is 20% alpha, the effective color depends on the surface behind it and fails the 3:1 threshold:

| Mode | Blended color | Against | Ratio | Needed |
|---|---|---|---|---|
| Light | `#D9E1E7` | body `#F1F4F7` | 1.20:1 | 3:1 |
| Dark | `#3A3B3C` | body `#111112` | 1.68:1 | 3:1 |

**Fix:** inactive dots now use `--color-text-secondary` (`#4E606F` light / `#AAAFB5` dark) - **5.90:1** (light) / **8.55:1** (dark), well above 3:1.

### 2. WCAG 1.4.1 - Use of Color (active dot)

Previously the only visual difference between active and inactive dots was a color change (neutral - accent), with no secondary cue. Users who cannot perceive the color difference had no way to identify the current page.

**Fix:** the active dot now renders as an **outlined ring** (inset accent ring with transparent fill) instead of a solid fill. Active vs inactive is now distinguished by **shape** (ring vs fill), not only by color. The inset ring adds no size, so the dot layout is unchanged.

## Changeset

- `fix-pagination-dots-contrast.md` - `@astryxdesign/core: patch`

## Out of scope (tracked in the issue thread)

- **Badge "New"** (green variant on colored cards): root cause is the translucent `0x33` (20%) background designed for white surfaces, which fails on colored cards. Needs a design decision (bump alpha vs. constrain usage vs. force semantic variants) - already flagged as pending in #4652.
- **Badge error variant** (`theme-neutral`, hardcoded `light-dark(...)` breaking token override): separate PR by @is0060hf.
- **Core color tokens** (#3654): separate PR #4022.

## Verification

- 64 Pagination tests pass
- ESLint clean (uses tokens, no hardcoded colors)
- TypeScript compiles
- Changeset passes `check:changesets`

Ref #4652

### DIFF
--- a/.changeset/fix-pagination-dots-contrast.md
+++ b/.changeset/fix-pagination-dots-contrast.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Pagination (dots variant): inactive dots now use `--color-text-secondary` instead of the translucent `--color-neutral`, meeting WCAG 1.4.11 non-text contrast (3:1) on body/surface. The active dot now renders as an outlined ring (inset accent ring with transparent fill) rather than a solid fill, so the current page is identifiable by shape and not only by color, addressing WCAG 1.4.1 Use of Color.
+
+@gonzoblasco
--- a/packages/core/src/Pagination/Pagination.tsx
+++ b/packages/core/src/Pagination/Pagination.tsx
@@ -183,9 +183,12 @@ const styles = stylex.create({
     borderStyle: 'none',
     padding: 0,
     borderRadius: '50%',
-    backgroundColor: colorVars['--color-neutral'],
+    // Use a solid secondary text token instead of the translucent --color-neutral
+    // so inactive dots meet WCAG 1.4.11 (3:1 non-text contrast) on body/surface.
+    // --color-neutral is 20% alpha and blends to ~1.2:1 (light) / ~1.7:1 (dark).
+    backgroundColor: colorVars['--color-text-secondary'],
     cursor: 'pointer',
-    transitionProperty: 'background-color',
+    transitionProperty: 'background-color, box-shadow',
     transitionDuration: durationVars['--duration-fast'],
     transitionTimingFunction: easeVars['--ease-standard'],
     outline: {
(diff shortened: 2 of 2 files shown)


rubyycheung (commented):
Thanks for digging into this. We're going to hold off — the pagination dot treatment is part of a broader selected-state direction we're working through, and a partial fix now would just get reworked. We'd rather solve the dots (contrast + active state together) holistically on our side.

On the active-dot approach specifically: making it a hollow ring removes its fill, so the current page has *less* ink than the inactive pages around it. In XDS **filled = selected**, so that inverts the hierarchy and reads backwards from the rest of the system.

Feel free to close this out — appreciate the work, and it directly informed how we'll approach it.

gonzoblasco:
Thanks @rubyycheung for the thorough review - that's a clear call and I'll close this out.

The "filled = selected" point lands: in XDS, selected states are filled, so hollowing the active dot inverts the hierarchy and reads backwards from the rest of the system. My fix treated the dots in isolation when they should follow the broader selected-state language. Good to know that direction is being worked through holistically.

The contrast half of it (inactive dots at 1.20:1 / 1.68:1, well under the 3:1 non-text threshold) is still a real WCAG 1.4.11 gap, so if useful I'm happy to pick it back up once the selected-state direction is settled, or contribute to whatever shape that work takes. No pressure either way - appreciate you taking the time to explain the rationale.

gonzoblasco:
Closing per maintainer direction - the pagination dot treatment is part of a broader selected-state effort being handled holistically on their side. Appreciate the review feedback on the filled=selected hierarchy.

PR #4707, opened 2026-08-04
fix(cli): filter skeleton component lists to resolvable components (#4677)

### LINKED ISSUE #4677: detail-page skeleton references Timeline but component registry has no Timeline
Astryx CLI 0.1.7 returns a detail-page skeleton whose component list and JSX include Timeline/TimelineSection. However, querying the required component contract with astryx component Timeline --dense exits nonzero and reports: No component named Timeline. This makes it impossible to follow the documented workflow of querying every proposed skeleton component before import. Expected: either register and document Timeline with an import path/props, or remove/replace it in the detail-page skeleton with semantic markup or an existing component. Observed with Node 22.23.2 and pnpm 10.15.0. No scaffold or package code was executed beyond the installed CLI query.

MeGaurav4:
## Summary

`astryx template <name> --skeleton` (and `--show`) advertised components that
`astryx component <Name>` cannot resolve. The list was scraped from every JSX
tag in the template source with no registry check, so 243 of the names across
the 40 built-in templates were phantoms: local helper functions
(`TimelineSection` -> `Timeline` in detail-page), third-party tags (recharts
`Line`/`Bar`), type-only names (`React`, `Record`), and
undocumented-but-importable core exports (`HStack`, `VStack`). Following the
skeleton workflow (`component Timeline` -> ERR_UNKNOWN_COMPONENT, fuzzy
fallback mis-resolves to `Toolbar`/`Divider`) was broken.

Skeleton/show component lists are now filtered against the actual
`astryx component` registry:

- documented core components, plus sub-components declared in a parent
  doc's `components: [...]` array (e.g. `ResizeHandle` in
  `Resizable.doc.mjs`), which `astryx component` resolves via the parent doc
- integration-owned components, so integration page templates keep
  advertising their own components
- the search-index path is intentionally unchanged and still indexes
  third-party terms

Closes #4677

## Verification

- `template('detail-page', {skeleton: true})` now lists only resolvable
  components (Timeline gone, Card kept)
- all 40 built-in templates: 0 advertised phantoms (was 243)
- new regression tests: template.test.mjs (no unresolvable components across
  built-ins) and template-integration.test.mjs (integration-owned components
  stay advertised, local helpers filtered)
- template suites green (19/19); full api+commands suite delta vs a clean
  tree is pre-existing Node-20 CLI-spawn noise, not this change
- `check:changesets`, `check:package-boundaries`, `check:sync` all clean


imdreamrunner (changes requested):
The `TimelineSection → Timeline` mismatch is real, but I am concerned that `knownComponentRegistry()` creates a second definition of “resolvable component.”

`astryx component <Name>` already has the authoritative resolution path for core, integrations, external packages, and parent-doc subcomponents. The new registry only partially reproduces it. In particular, it calls template `discoverAll()` and scans each template `filePath` for `components: [...]`, while the relevant component docs use object arrays. This may incorrectly remove valid subcomponents such as `ResizeHandle` or `StackItem`.

Could the component subsystem instead expose a reusable exact-resolution index?

```js
const resolvable = await listResolvableComponentNames(cwd);
components.filter(name => resolvable.has(name));
```

It can be built once, avoid fuzzy matching and repeated doc loading, and remain the single source of truth used by `astryx component`.

We should also document that `template.show/skeleton.components` now means “Astryx components resolvable through the CLI”; the current contract only says “components the template composes.” A positive regression test should verify that a parent-doc subcomponent is retained.

MeGaurav4:
Replaced the old knownComponentRegistry() with listResolvableComponentNames(), which sits in the component discovery module and uses the same resolution path as `astryx component`: core components, loaded integrations, back-compat astryx.docs packages, and parent-doc subcomponents read through loadDocs, so object-array entries like {name: 'HStack'} survive. The old doc-file regex scan is gone. I also removed the fuzzy suffix matching in extractComponents, so a local helper like TimelineSection no longer shows up as Timeline; a name is advertised only if it resolves exactly. The index builds once per workspace.

Docs for template.show/skeleton.components now say those lists only contain Astryx components resolvable through the CLI. Added regression tests: a parent-doc subcomponent is kept, integration and external-package components are kept, and local helpers are dropped.

packages/cli/foundation/discovery/component-discovery.mjs
```diff
@@ -679,6 +682,122 @@ export function discoverOwnedComponents(coreDir, loadedIntegrations = []) {
...
+    externalPkgs.map(e => e.name).sort().join(','),
+  ].join('|');
+  if (resolvableNamesCache.has(memoKey)) return resolvableNamesCache.get(memoKey);
+
+  const names = new Set();
+  /** @type {Set<string>} component docs to scan for parent-doc subcomponents */
+  const docPaths = new Set();
+
+  // Core + loaded-integration components with a resolvable doc.
+  for (const rec of discoverOwnedComponents(coreDir, loadedIntegrations)) {
```
cixzhang: This is the `component --list` catalog, not what `component <Name>` resolves — `Resizable` and `SyntaxTheme` never reach it.

packages/cli/api/template/template.test.mjs
```diff
@@ -76,5 +81,68 @@ describe('template --skeleton component extraction (prefix-agnostic)', () => {
...
+      listResolvableComponentNames,
+    } = await import('./template.mjs');
+    const templates = await discoverTemplates();
+    const builtin = templates.filter(t => !t.package);
+    const registry = await listResolvableComponentNames();
+    expect(registry).toBeTruthy();
+    for (const tmpl of builtin) {
+      const components = extractComponents(tmpl.filePath, registry);
+      for (const comp of components) {
+        expect(registry.has(comp)).toBe(true);
```
cixzhang: This filters by `registry` then asserts `registry` — it can't fail. Resolving each name through `astryx component` would.

packages/cli/api/template/template.type.mjs
```diff
@@ -30,7 +30,7 @@
  * @property {'page' | 'block'} type
  * @property {string} package - Owning package; core (built-in) templates report '@astryxdesign/core'.
  * @property {string} [category] - Optional grouping/category label.
- * @property {string[]} [componentsUsed] - Component display names the template composes.
+ * @property {string[]} [componentsUsed] - Astryx components the template composes; each is exact and resolvable through `astryx component <Name>`.
```
cixzhang: `componentsUsed` isn't filtered — 53 of 1875 entries don't resolve. Might drop this sentence.

cixzhang (changes requested):
Thanks for this, and for reworking it onto `loadDocs` — `detail-page` is genuinely fixed, `Timeline` and the rest of the phantoms are gone.

I think @imdreamrunner's worry is still live though. The index starts from `discoverOwnedComponents`, the `component --list` catalog, but `astryx component <Name>` resolves through `findComponentReadme`, which reaches further. `astryx component ResizeHandle` works, yet `ide`, `ai-chat`, `incident-console` and three more stop advertising it, and `SyntaxThemeShowcase` stops advertising `SyntaxTheme` — twelve templates, each hiding a component the builder is meant to query. That's [#4677](https://github.com/facebook/astryx/issues/4677) with the sign flipped.

`findComponentReadme` plus `scopeSubComponent` in `api/component/_adapter.mjs` are the actual path. Running a template's ~20 tag names through those is exact by construction, and cheaper than indexing all of core.

Would you rather resolve per name, or have the component subsystem export a `canResolve()` for the template layer?

[Reviewed by Robohands]

cixzhang:
Correction, and an apology — my blocking finding above was wrong.

I said six templates would stop advertising `ResizeHandle`. They don't: your `listResolvableComponentNames()` folds in parent-doc subcomponents from a doc's `components:` object array, `Resizable.doc.mjs:112` declares `ResizeHandle`, and your own test at `_adapter.test.mjs` covers exactly that case. I read the resolution loop and stopped before the part of your diff that handles it. The fix I asked for — routing through `findComponentReadme` — is already what `discoverOwnedComponents` does.

`SyntaxTheme` is a real gap: there's no `SyntaxTheme.tsx` and no parent doc declaring it, only the three preset files, so nothing enumerates the bare name. That's one template, not six, and it's arguably correct behaviour — the name doesn't resolve because the component doesn't exist.

Withdrawing the block. I'll re-review properly rather than leave you guessing at a moving target; sorry for the round trip, on top of the sixteen days.

[Reviewed by Robohands]

cixzhang (approved):
Withdrawing my request-changes above — the block was my misreading. See the comment.

[Reviewed by Robohands]

MeGaurav4:
Thanks for circling back and for the correction, and no worries about
the round trip. Agreed on SyntaxTheme: the name does not resolve today,
so not advertising it is arguably correct until the presets become a
real component. Happy to revisit if that changes.

I rebased the branch onto current main after your review (one generated-
docs conflict, resolved per generate-cli-readme semantics); checks are
re-running on the new head.

packages/cli/foundation/discovery/component-discovery.mjs
```diff
@@ -680,6 +683,161 @@ export function discoverOwnedComponents(coreDir, loadedIntegrations = []) {
...
+  const coreDocResults = await Promise.allSettled(
+    coreDocFiles.map(async p => {
+      const doc = await loadDocs(p);
+      return {doc, path: p};
+    }),
+  );
+  for (const r of coreDocResults) {
+    if (r.status !== 'rejected' && r.value) {
+      const {doc, path} = r.value;
+      if (doc.name) names.add(doc.name.replace(/^XDS/, ''));
```
cixzhang: This adds doc names, but source-fallback resolver names like `AppShellMobileContext` can still appear in skeleton output without being listed.

cixzhang (changes requested):
Thanks for chasing the doc-file gap; the earlier `ResizeHandle`/`SyntaxTheme`/`NavHeadingMenu` misses are fixed.

I think one resolver edge is still leaking through. `MobileNavToggleBasic --skeleton` still prints `<AppShellMobileContext />`, but its component list is only `MobileNav`, `MobileNavToggle`, `SideNavItem`, `SideNavSection`. `astryx component AppShellMobileContext` exits 0, so builders still see a skeleton tag the list does not tell them to query — the same [#4677](https://github.com/facebook/astryx/issues/4677) contract, now on the source-fallback path.

Could we either include source-fallback resolvable names in `listResolvableComponentNames()` or suppress those tags from the skeleton?

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks for carrying the resolver work through the source fallback. Three things still block this head:

- `MobileNavToggleBasic`/`Showcase` render `AppShellMobileContext`, and `NavHeadingMenuShowcase` renders `NavHeadingMenuItem`, but none appears in `components`; each component lookup succeeds. `isComponentSourceFile()` excludes these source/parent-doc cases. Please use the exact resolver and add positive omission tests—the current membership test only catches extras.
- `TemplateListEntry` drops six fields that `template.list` still emits: `displayName`, `aspectRatio`, `exampleFor`, `alsoExampleFor`, `alsoShowcaseFor`, and `isShowcase`. Please restore the public type, update response docs/contract tests for the clarified list meaning, and add a patch Changeset.
- Each cold `template --show/--skeleton` imports all 202 core docs; process-local memoization cannot help one-command CLI use (0.6–0.7s → 1.4–1.8s here). Please resolve only the selected template’s names or add a deterministic work bound.

If you’d rather talk it through, we’re in [Discord](https://discord.com/invite/XnsUcFykEP).

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks for carrying the source/parent-doc cases through; those work now, and the list fields plus Changeset are restored.

Two things still block this head:

- `component-discovery.mjs:722` still defines resolution separately from the component command. In a configured integration, the template advertises `Card` while `astryx component Card` rejects it as ambiguous. On `detail-page` it performs 917 core directory reads (main: 0); cold median moves from 0.64s to 1.77s, and the sweep exceeds Vitest’s 30s default. Please route both paths through one exact resolver and bound one template lookup’s work.
- `template.doc.mjs:102` and `template.type.mjs:53` still describe any composed components, not the new resolvable-only contract, and there is no response-contract test for that meaning.

CI is also red on the unused `findCoreDir` import.

Could you address those and rerun the checks? If you’d rather talk it through with someone, we’re in [Discord](https://discord.com/invite/XnsUcFykEP).

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks for sharing the owner resolver; source/parent-doc names and cold ambiguity now work.

Two things still block this head:

- The process-wide resolution cache keys only the core path, component name, and integration package names. When one process queries two projects using different roots for the same integration, the second template reuses the first answer: it advertises `Card`, then `component('Card')` rejects it as ambiguous. Please make each cwd resolve against its own owner roots and add this two-project regression case.
- The public response contract is still split. `template.doc.mjs` and `template.type.mjs` promise resolvable-only names, while `response-types.doc.mjs` and the generated CLI README still promise all composed names. Please update the canonical response registry and regenerate the docs.

[Reviewed by Robohands]

cixzhang (changes requested):
Agents that inspect two projects in one process can still receive a component name the second project cannot resolve.

The integration-root case is fixed, and the response docs are now in sync. But after project A resolves `TimelineSection` from a legacy `pkg.astryx.docs` package, project B—with the same core root and no such package—still advertises `TimelineSection`; `component('TimelineSection')` returns `ERR_UNKNOWN_COMPONENT`. The result cache key includes configured-integration roots but not the cwd/legacy-external roots used by the fallback path.

Could you include that fallback owner context in the cache identity and add this two-project regression case?

[Reviewed by Robohands]

PR #4712, opened 2026-08-04
feat(theme): add themeable indicators

### LINKED ISSUE #4831: Indicators: componentized, themeable selection visuals (vibe test results)
## Indicators — componentized, themeable selection visuals

Tracking issue for the indicator layer landing in #4712.

A checkbox's box, a radio's circle, and the mark beside a chosen row are **stateful
visuals**, not glyphs. They have unchecked / checked / mixed states, they tint when
the row around them is hovered, and a theme may want to swap one for another
entirely. The earlier attempt modeled them as theme-supplied *render functions*
(`controlIcons`), which fought the system three ways — inline styles outrank every
layer of the cascade, function-valued theme fields don't survive `astryx theme
build`, and hover/focus had nowhere to come from because that state is genuinely CSS
ancestor state. #4712 replaces that with components: `CheckboxIndicator` /
`RadioIndicator`, a `defineTheme({indicators})` registry, and a slot mapping so a
component's selection mark can be pointed at a different indicator.

Three goals, all met by the PR:

- **Theme them** — an indicator renders stable class targets, so `components:
  {checkbox, radio, radio-dot}` reaches it like any other component. Replacement is
  the second, heavier level.
- **Use the radio control in a selector, per theme** — `componentIndicators:
  {'selector-selected-option': 'radio'}`. Confirmed in a browser: all options render
  a radio, exactly one checked, unselected options draw an empty circle.
- **Swap the selection indicator generally** — any indicator into any
  indicator-capable slot, including a theme's own component.

---

## Vibe test results

Ran an API-arbitration vibe test on the one decision that was still open: what the
`state` values should be called. Test lives at
`internal/vibe-tests/indicator-state-test/` (`PLAN.md`, `prompts.json`, four variant
docs, `RESULTS.md`).

**Method.** 4 conditions × 4 prompts = 16 generations, `claude-sonnet-5` at fixed
effort. One isolated sub-agent per cell, spawned with **no node attached** — no
filesystem, no network, so a generator could not reach the repo, the real API, or
another condition's doc. Prompts describe UX only; verified beforehand that no prompt
and no variant doc contained another condition's value words (this caught three
leaks — "fully on or fully off", "turn off", "when on… when off" — all rewritten).

| Condition | `state` values |
|---|---|
| `A-onoff` | `off` / `on` / `partial` |
| `B-selection` | `unselected` / `selected` / `partial` |
| `C-native` (control) | `unchecked` / `checked` / `indeterminate` |
| `D-elided` (probe) | values removed — the doc asks the author to name them |

### 1. The third value is `mixed`, unanimously — and it isn't `partial`

`D-elided` is the read: whatever it invents is the unprimed mental model.

| Prompt | resting | engaged | third |
|---|---|---|---|
| `ind-2` file list | `unchecked` | `checked` | **`mixed`** |
| `ind-4` brand tick box | `off` | `on` | **`mixed`** |
| `ind-8` folder tree | `unselected` | `selected` | **`mixed`** |

The binary pair is a **three-way tie, one vote each** — no data-driven winner, so it
should be tie-broken on other grounds rather than argued.

The third value is **`mixed` 3/3. `partial` scored 0/3** — it was our invention and
nobody reached for it. `mixed` is also the ARIA token, and a `C-native` run
independently wrote `aria-checked={isIndeterminate ? 'mixed' : allChecked}` while its
own doc said `indeterminate` — arriving at the same word from the a11y side.

### 2. Unplanned: the enum does not stay internal

We designed `state` as an indicator-authoring detail. Generators put it on the
**consuming component's public API**:

```tsx
<Checkbox state="off" | "on" | "partial" />                    // A
<Checkbox state="unchecked" | "checked" | "indeterminate" />   // C
```

One even noted it chose this "rather than inventing an unrelated
`checked`/`indeterminate` boolean pair" — the indicator vocabulary displaced the
conventional control API. Whatever we pick gets mirrored onto public props. That is
the argument against `on`/`off`: Astryx ships a real `Switch`, and
`<Checkbox state="on">` reads like one.

### 3. Draws-in-every-state — 4/4 pass

The highest-severity risk: an indicator that renders only when selected leaves blank
rows and silently breaks every themed radio. All four conditions produced the
declarative mapping unchanged and nobody hand-rolled a conditional mark — **because
the mapping is declarative, there is nothing to get wrong.** An argument for keeping
the slot API declarative rather than exposing a lower-level "render the mark
yourself" seam.

### 4. Replacement contract — 4/4 pass on the three that matter

`aria-hidden` 4/4 · renders `children` (the busy-spinner path) 4/4 · does **not**
draw its own focus ring 4/4.

Worth stating plainly: these are the **exact three mistakes made by hand in #4712** —
the first draft moved the focus ring into the indicator, and both shipped example
replacements ignored `children`. Four of four generations got them right when the doc
stated each in one sentence. The contract is teachable; it just has to be said.

### 5. Tokens — 4/4 fail

Every replacement invented CSS custom properties that don't exist — `--radius-md`,
`--radius-sm`, `--color-brand`, `--color-on-brand`, `--color-positive-muted`,
`--color-border` — and every one flagged it: *"wished the doc had told me the full
set of available tokens."* Invented tokens are a blocking-criteria violation here.

### 6. The dominant gap: no consuming component, no theme application

Flagged by **every run**. The doc describes the indicator but never the component
that hosts it, nor how a defined theme is applied. 4/4 invented a `Dropdown`; others
invented `<Checkbox state=…>` or fell back to a raw `<input type="checkbox">` with
imperative `ref.current.indeterminate`. `<Theme theme={…}>` (correct),
`<Theme name="brand">`, and `<ThemeProvider>` were all guessed — the doc imports
`Theme` and never uses it once.

Two runs declined to invent a component and hand-rolled native inputs, which is the
worst outcome available: docs that push a generator *out* of the design system. This
is a bigger lever on output quality than the naming question we ran the test for.

### 7. Expand/collapse indicators — the need is real, the vocabulary is untested

`ind-8` put open/shut and ticked/mixed on the same row. No collision occurred — but
only because **4/4 hand-rolled the expand affordance** as a `<button>` with `▸`/`▾`
glyphs. Nobody attempted to use an indicator for it, because none is documented.

Two conclusions of different strength, and the difference matters:

- **Well supported:** the need to swap expand/collapse indicators is real. Given a
  disclosure control and an indicator system, a generator hand-rolls the disclosure
  every time. Multiple teams have asked to customize the expand/collapse toggle on
  selectors and menus; this is the same gap seen from the generation side.
- **Now measured — see round 2 below.** The two prompts designed to test the
  vocabulary question (`ind-6`, a chevron swap; `ind-8`, both families on one row)
  were not run in round 1, so that conclusion was asserted rather than measured.
  Round 2 corrects it.

### Round 2 — cross-family, measured

Two more conditions, whose docs **both** document expand indicators
(`expandStart` / `expandEnd`) alongside selection. 580 words each, differing only in
vocabulary: `E-shared` uses `off`/`on`/`mixed` for selection and `off`/`on` for
expand ("the same vocabulary every indicator uses"); `F-perfamily` uses
`unchecked`/`checked`/`mixed` and `collapsed`/`expanded`.

**The shared vocabulary did not stay shared.** `ind-8` / `E-shared` was handed one
vocabulary for both families and immediately declared two types anyway:

```ts
type SelectionState = 'off' | 'on' | 'mixed';
type ExpandState = 'off' | 'on';
```

Given a single enum, the generator re-derived the split on its own, because the state
spaces genuinely differ. Sharing the value *names* bought nothing — it just made two
distinct types share overlapping strings, so `state="mixed"` would typecheck on a
chevron.

**Shared vocabulary collides with ARIA at the call site.** `ind-6` / `E-shared` wrote
on adjacent lines:

```tsx
aria-expanded={isOpen}
<Indicators.expandStart state={isOpen ? 'on' : 'off'} />
```

One boolean, two vocabularies. `F-perfamily`'s equivalent reads
`state={isOpen ? 'expanded' : 'collapsed'}`, matching the `aria-expanded` beside it.
Same leak as finding 2, observed in the disclosure family.

---

## Decisions

- [ ] Rename the third state `partial` → **`mixed`** (3/3 unprimed, matches ARIA).
- [ ] Keep **`unchecked` / `checked`** for the binary pair — the `aria-checked`
      vocabulary, which spans every selection control we have (checkbox, radio,
      `menuitemcheckbox`, `menuitemradio`, selected listbox option) and survives being
      mirrored onto public props per finding 2.
- [ ] Adopt **per-family vocabularies drawn from ARIA** — selection follows
      `aria-checked` (`unchecked`/`checked`/`mixed`), expand follows `aria-expanded`
      (`collapsed`/`expanded`). Measured in round 2, not assumed.
- [ ] **Per-family typing follows from that.** A flat `IndicatorProps` with a widened
      union would let a chevron accept `'mixed'` and a checkbox accept `'expanded'`.
      The state space belongs to the family, so `IndicatorProps` should be
      parameterized by it, with the registry and slot map as family-checked mapped
      types — which turns "map a slot to any indicator of the same family" from a doc
      sentence into a type rule. `IndicatorProps<F = 'selection'>` is backward
      compatible with what #4712 ships, so the parameter can land with the expand
      family. What must **not** ship is a widened flat union spanning both families.
- [ ] Name the tokens a replacement should use, in the indicator docs (finding 5).
- [ ] Show the owning component and `<Theme>` in use, in the indicator docs
      (finding 6).

## Follow-ups (not in #4712)

- [ ] `componentIndicators` as a map parallel to `icons`, and make the check glyph an
      icon-backed indicator so everything in a selection slot has one shape. Keep the
      resolution seam internal — it has one call site, and one data point is thin
      evidence for a permanent public name.
- [ ] `expandStart` / `expandEnd` indicators, so disclosure affordances are swappable
      the same way selection marks are. Two, not one: a start-side chevron points along
      the text direction and mirrors under RTL (`TreeListItem`, `rotate(90deg)`), while
      an end-side chevron points down and does not (`Selector`, `rotate(180deg)`).
      Note `astryx-selector-indicator-icon` already carries `state: expanded |
      collapsed` today, so this formalizes an existing surface rather than inventing
      one.
- [ ] `IndicatorMap` module augmentation currently hands back a typed `undefined` at
      runtime for a name with no default. Make it safe, or close the map.
- [ ] `Indicator.doc.mjs` — the directory has no doc file, so `themingTargets.test.ts`
      skips it and `npx astryx component` can't see the new components.
- [ ] A Selector-level test asserting the `{indicator: 'radio'}` path renders a radio
      on every option with exactly one checked.

cixzhang:
Adds **indicators** — the componentized form of stateful control visuals — and makes them replaceable by name from a theme.

> **Rescoped.** This PR previously carried component icon slots (`componentIcons`) and a parallel `componentIndicators` slot map. Both are dropped: replacement by name covers every case we have, and slot maps stay additive for when it doesn't. #4647 is closed with the reasoning.

## What an indicator is

The checkbox box, the radio circle, the mark on a chosen option. The owning component keeps the input, role, accessible name, focus, and keyboard behavior; the indicator turns state into a picture. It renders `aria-hidden` and owns no semantics.

## Consumer-facing API

**Replace an indicator.** Replacement is by *name*, so one entry reaches every component that draws it:

```tsx
import {RadioIndicator} from '@astryxdesign/core/Indicator';

defineTheme({name: 'brand', indicators: {check: RadioIndicator}});
```

Every single-selection mark in the app is now a radio — Selector options included, drawing an empty circle on unselected rows, which a check mark never could.

**Restyle one instead.** An indicator renders stable `astryx-*` targets, so `defineTheme({components: {checkbox: …}})` reaches it like any other component. Prefer this; replacement is for when the shape itself is wrong.

**Write a replacement.** Four rules, each one a mistake made by hand in this PR's first draft:

- render `children` when present (the owner passes a loading Spinner through it)
- set `aria-hidden` (the owner provides role and accessible name)
- do **not** draw a focus ring (it lives on the owner's wrapper, so it survives replacement)
- use theme tokens for every color, radius, and border width

## Families

A family fixes the state space, and each uses the ARIA vocabulary for the state it expresses:

| family | states | source | indicators |
|---|---|---|---|
| `singleSelection` | `unchecked` \| `checked` | `aria-checked` | `check`, `radio` |
| `multiSelection` | `unchecked` \| `checked` \| `indeterminate` | `aria-checked` | `checkbox` |

The registry is family-checked, so a replacement can't be written against the wrong states — a `radio` replacement expecting `indeterminate` is a type error. Adding a family is additive and module-augmentable.

Vocabulary comes from the vibe test in #4831, not from taste: given one shared enum across families, generators re-derived the split anyway and wrote `aria-expanded={isOpen}` beside `state="on"` on adjacent lines.

## Notable

- **Selector renders its mark unconditionally**, passing `state`. `{isSelected && …}` would make a replaced radio unable to draw its unselected form. The default check still renders nothing when unchecked.
- **`CheckIndicator` renders `<Icon>` directly** rather than wrapping it, so a host's theme target lands on the glyph — and it adds **no new theme target**, since `astryx-icon` plus the host's target already reach it. Continues the wrapper reduction from #4838/#4846.
- **Focus-ring fix.** `CheckboxInput`'s input is `opacity: 0`, so with the ring inside the indicator, a theme replacement left the control with no visible focus (WCAG 2.4.7). Moved to the wrapper that owns the input, mirroring `RadioListItem`. Found by an independent review pass; pinned by tests.
- **~450 lines of duplicated control styling removed** from CheckboxInput, RadioListItem, and the menu items.
- Two a11y improvements fall out: the radio visual picks up the `prefers-reduced-motion` override it lacked, and menu radios gain forced-colors treatment.

## Migration

Menu radios now use the shared `radio` / `radio-dot` targets:

- `dropdown-menu-radio-dot` is **removed** — target `radio-dot`, which now also reaches RadioList.
- Move fill/border/radius rules from `dropdown-menu-radio` to `radio`; layout rules (size, order) stay.
- `astryx theme build` reports the removed key. Runtime themes are not validated.
- Menu controls use the standard 20/24px control sizes instead of a bespoke 18/22px.

## Test plan

- `Indicator.test.tsx` — contract per indicator, and family typing.
- `Selector.test.tsx` — the default mark is the glyph itself (no wrapper); a theme mapping `check` → `RadioIndicator` puts a radio on **every** row and fills exactly the selected one.
- `CheckboxInput.test.tsx` — focus ring survives indicator replacement. Verified by reverting the fix and confirming failure.
- Full UI suite: **5 failures, identical to `origin/main` on the same quiet machine** (Table perf budget, TransferList draft timing — pre-existing flakes). 6528 passing vs main's 6516.
- Typecheck clean; lint clean under `CI=true` (strict mode — warnings become errors).

Visual verification with before/after screenshots to follow.

Closes #4831


cixzhang:
Heads up on an overlap with #4724 (`RadioControl`): both PRs refactor `RadioListItem` and both own the `astryx-radio` / `astryx-radio-dot` targets, so whichever lands second needs a rebase.

They're solving adjacent problems — #4724 extracts the radio **input + visual** so a single radio can render outside a `RadioList`; this PR extracts the **visual alone** so a theme can restyle or replace it everywhere it appears. Both could stand, with `RadioControl` composing the indicator for its visual rather than drawing its own, which would keep one set of theme targets in one place. Worth agreeing on that before either merges rather than reconciling it afterward.

cixzhang:
Rescoped to the minimal API, and retargeted to `main`.

**Dropped:** `componentIcons` and the parallel `componentIndicators` slot map. What's left is `defineTheme({indicators})` — replacement by name. EPS replaces `CheckIndicator` with their radio visual and every single-selection mark follows, which was the actual use case.

**Size:** 82 files / +2850 → **26 files / +1574**. Four changesets → one, 57 words.

**Also removed a wrapper I'd added.** The first pass had `CheckIndicator` render a `<span>` around an `<Icon>` with its own `check-indicator` theme target. Both were wrong: the wrapper moved `astryx-selector-check` off the glyph onto a box around it (undoing #4838/#4846), and the new target violated "don't add new theme targets." Caught by two existing Selector tests that assert the target lands on the glyph and that the default check renders byte-identically to a plain `<Icon>`. It now renders `<Icon>` directly and adds no target.

**On the test failures:** the full UI suite shows 5, in Table perf and TransferList draft timing. I ran pristine `origin/main` in a separate worktree on a quiet machine — **also 5, same files.** My first comparison showed a difference, but that was me running suites in parallel; under load the baseline itself swung between 2 and 6.

Not yet done: visual before/after screenshots. Doing those next unless you want to look at the API first.

packages/core/src/CheckboxInput/CheckboxInput.tsx
```diff
@@ -68,6 +66,22 @@ const styles = stylex.create({
...
+    outline: {
+      default: 'none',
+      ':has(:focus-visible)': `2px solid ${colorVars['--color-accent']}`,
+    },
+    outlineOffset: {
+      default: '0',
+      ':has(:focus-visible)': '2px',
+    },
+    borderRadius: radiusVars['--radius-inner'],
+  },
```
cixzhang: This styling belongs on the indicator now.

packages/core/src/DropdownMenu/DropdownMenuCheckboxItem.tsx
```diff
@@ -187,15 +188,22 @@ export function DropdownMenuCheckboxItem({
...
-            value={value}
-            isDisabled={isDisabled}
+        <span
+          aria-hidden="true"
+          {...mergeProps(
+            themeProps('dropdown-menu-checkbox', {
+              size: controlSize,
+              checked: value ? 'checked' : null,
+              disabled: isDisabled ? 'disabled' : null,
+            }),
```
cixzhang: Hmm should this be on the CheckboxControl instead? Themes should collapse onto the visible element. Also we apply size on this extra span which seems redundant to the CheckboxControl

packages/core/src/DropdownMenu/DropdownMenuRadioItem.tsx
```diff
@@ -205,24 +178,13 @@ export function DropdownMenuRadioItem({
               checked: isChecked ? 'checked' : null,
               disabled: isDisabled ? 'disabled' : null,
             }),
```
cixzhang: Same comment here as above for the dropdown menu checkbox item.

cixzhang (commented):
Can we get a storybook for swapping the selector's single selection indicator to use the radio?

packages/core/src/CheckboxInput/CheckboxInput.tsx
cixzhang: Moved — and you were pointing at a real bug, not just a placement preference.

The ring can't use a plain `:focus-visible` on the indicator: the native input is a visually hidden **sibling**, not a descendant. So it now arrives the same way hover already does, through the owner's `indicatorScope` marker:

```ts
outline: {
  default: 'none',
  [stylex.when.ancestor(':has(:focus-visible)', indicatorScope)]: `2px solid ${colorVars['--color-accent']}`,
},
```

The bug it fixes: with the ring on the wrapper, the **host had to hardcode the indicator's shape**. `RadioListItem` literally carried `borderRadius: '50%'` for a circle it doesn't own, and CheckboxInput carried `--radius-inner`. Replace either indicator with a different shape and the ring traced the wrong outline. Now each indicator draws its own — measured in Chrome after the move: checkbox `2px solid, offset 2px, radius 6px`; radio `2px solid, offset 2px, radius 50%`.

The tradeoff, stated plainly: a replacement that doesn't draw a ring now ships a control with no visible focus (WCAG 2.4.7) — previously the wrapper covered for it. I flipped the rule in `Indicator.doc.mjs` from *"don't draw your own focus ring"* to *"draw it, keyed off the scope marker,"* with the sibling-input reason. Worth noting the vibe test in #4831 measured 4/4 replacements honoring explicitly-stated contract rules, which is the evidence that documenting it works.

Tests updated: the ring is asserted on the indicator, plus a regression guard that no host reintroduces one (two rings, or one of the wrong shape).

packages/core/src/DropdownMenu/DropdownMenuCheckboxItem.tsx
cixzhang: Both right, and the wrapper is gone.

The size was pure duplication — `markerSizeStyles` was `{sm: 20×20, md: 24×24}`, byte-identical to the indicator's own `boxSizeStyles`. `Item` renders `{marker}` with no wrapper of its own, so the span had nothing left to do; what remains (order, coarse-pointer placement, `pointerEvents: none`) is genuinely the menu's business and rides `xstyle` on the indicator.

One correction while doing this: I had put a `dropdown-menu-checkbox` target on it, and that would have been a **new** theme target — main doesn't declare one (only `dropdown-menu-radio` and `-radio-dot` are in `DropdownMenu.doc.mjs`). I checked what main actually renders on that element and it's `astryx-checkbox`, which is still there. So the menu checkbox adds no target at all. The radio keeps `dropdown-menu-radio`, which did pre-exist, now on the painted circle instead of a box around it.

Visual diff vs main with the menu **open**: `menu-checkbox-items` is **0px changed**; the radio stories differ by the disclosed 22→24px control-size alignment.

packages/core/src/DropdownMenu/DropdownMenuRadioItem.tsx
cixzhang: Same change here — wrapper removed, `dropdown-menu-radio` moved onto the painted circle (it pre-existed on main, so it stays), and the duplicated size styles are gone.

This is the one place with a real pixel change: the menu radio was a bespoke 18/22px and now uses the standard 20/24px, so it matches a RadioList radio. Measured 1.8% of pixels in `lab-radio-group` and 0.4% in `lab-selectable-sizes`; every other story I diffed is 0px.

cixzhang:
All four addressed in `ade8e38`.

**Storybook for the radio swap** — added two stories side by side, both opened by default so the listbox is visible:
- [`RadioSelectionIndicator`](https://facebook.github.io/astryx/pr/4712/?path=/story/core-selector--radio-selection-indicator) — `defineTheme({indicators: {check: RadioIndicator}})`. Every row draws a circle; only the selected one is filled.
- [`DefaultSelectionIndicator`](https://facebook.github.io/astryx/pr/4712/?path=/story/core-selector--default-selection-indicator) — unthemed, for comparison: a check on the selected row, nothing on the others.

That contrast is the whole argument for rendering the mark unconditionally — the empty circles only exist because the indicator is told its state instead of being conditionally rendered.

**Visual verification.** Ran both Storybooks (this branch and pristine `origin/main`) and diffed the same stories pixel by pixel:

| story | pixels changed |
|---|---|
| `checkboxinput--all-variations` | 0 |
| `checkboxinput--indeterminate` | 0 |
| `radiolist--all-variations` | 0 |
| `dropdownmenu--lab-checkbox-items` | 0 |
| `dropdownmenu--lab-radio-group` | **1.82%** |
| `dropdownmenu--lab-selectable-sizes` | **0.37%** |

The two non-zero ones are the disclosed 22→24px menu radio alignment. Everything else is byte-identical.

Two things I got wrong before the numbers were trustworthy, both caught by checking rather than by the numbers looking plausible: the first run compared two *"Couldn't find story"* pages (I'd guessed story IDs), and the second compared two *closed* menus because my click was wrapped in a silent `catch`. I added a positive control — a deliberate 6px red ring on the radio dot moved 2602 pixels — so the harness is proven to detect change.

**Test suite:** 8 failures on the full run, all in perf/timing files (Table perf, Markdown perf, TransferList ×5, submenu roving). Pristine main on the same machine gives 5, from the same set, and the count swings run to run under load. Every component this PR touches — Indicator, CheckboxInput, RadioList, DropdownMenu, Selector — is **256/256 passing** in isolation.

packages/core/src/CheckboxInput/CheckboxInput.tsx
cixzhang: Right — that was a bad answer on my part. Callsites shouldn't, and now they don't.

I looked at the two built-in rings and they were **byte-identical**, and neither mentions a shape. That's the tell: `outline` follows the element's own `border-radius` automatically, so the rule is completely generic. There was never anything per-indicator in it, which means "write your own" was me exporting a chore.

So it's one exported style now, next to `indicatorScope`:

```tsx
import {indicatorFocusRing} from '@astryxdesign/core/Indicator';

<span {...stylex.props(indicatorFocusRing, myStyles.box)} aria-hidden="true" />
```

The built-ins spread exactly that. A replacement spreads the same one line and gets a ring that matches its own shape for free — the doc rule now points at the export instead of describing rules to write.

It still has to live *on* the indicator rather than the owner, for the reason the original comment was getting at: the owner would have to hardcode the shape of whatever it hosts. And it keys off the scope marker because the native input is a hidden sibling. But neither of those is now the replacement author's problem.

One case that fell out: **CheckIndicator deliberately doesn't take the ring.** A check marks a chosen row in a listbox or menu, and the row takes focus, not the mark. I noticed because adding it broke the existing test asserting the default check renders byte-identically to a plain `<Icon>` — the test was right.

Re-measured in Chrome after the refactor, unchanged: checkbox `2px solid / offset 2px / radius 6px`, radio `2px solid / offset 2px / radius 50%`.

PR #4716, opened 2026-08-05
feat(Avatar): make fallback background a direct theme target

cixzhang:
## What

Make the Avatar **fallback surface** (initials + default person icon) a direct theme target via a new stable `astryx-avatar-fallback` class, and route **all** themeable fallback properties through it — removing every internal `--_avatar-fallback-*` var (background, text color, font weight, and per-size font size).

```tsx
defineTheme({
  name: 'brand',
  components: {
    'avatar-fallback': {
      base: {
        backgroundColor: 'var(--color-accent-muted)',
        color: 'var(--color-text-accent)',
        fontWeight: 'var(--font-weight-normal)',
      },
      'size:sm': { fontSize: '9px' }, // per-size initials scale
    },
  },
});
```

## Why

The fallback typography/background were themed by setting them on the `avatar` key, which the generator expanded into internal `--_avatar-fallback-*` vars read by the fallback element. That indirection existed for one reason: the fallback child had **no class of its own**, so the only way to reach it was a var set on the root `.astryx-avatar` wrapper (which also leaked the background onto the wrapper — the origin of the wrapper-rounding workaround).

Giving the fallback its own stable class removes the indirection entirely. `backgroundColor`/`color`/`fontWeight` are static StyleX classes, and `fontSize` is a StyleX **dynamic** style — which compiles to a class applying `font-size: var(--x-…)` with only the *computed value* set inline as a custom property. In every case the CSS **property** lands via a class in `@layer astryx-base`, so a theme rule on `.astryx-avatar-fallback` (or its `.<size>` variant) in `@layer astryx-theme` wins by layer order. This is the same mechanism `generateSizeOverrides` already uses to let a theme's per-size `font-size` beat the base StyleX class for Text. No var seam is needed for any of them.

Net effect: the fallback goes from four internal vars to **zero**, and a themed background lands only on the fallback, not the wrapper.

## Changes

- `Avatar.tsx`: fallback elements render `themeProps('avatar-fallback', {size})`; background/color/weight are plain token defaults and font size is a plain proportional dynamic value.
- `derivedVarRegistry.ts`: drop the `avatar` entry entirely.
- `Avatar.doc.mjs`: document the `astryx-avatar-fallback` target (with `size` visualProp); remove the now-empty `vars`/`derived` theming fields.
- Stories: the existing "Themed Fallback Type Scale" sets weight/color/background/per-size font-size all on the `avatar-fallback` key; **new "Themed Fallback Background" story** demonstrates theming the background on both initials and the default-icon fallback.

## Testing

- Unit test asserts both fallback kinds carry `astryx-avatar-fallback`, and that the default font-size renders proportionally via a class (no internal var).
- Theme-generation tests assert direct `background-color`/`color`/`font-weight` rules on `.astryx-avatar-fallback` and a direct `font-size` rule on `.astryx-avatar-fallback.<size>`, with no internal vars emitted.
- Full `@astryxdesign/core` suite green (5841 tests); core build, docsite `generate`, lint, and the sync/changeset gates all pass.

## Risk

Low. Default rendering is byte-for-byte unchanged. The one behavioral shift is intentional and strictly better: a themed fallback background now lands only on the fallback, not the wrapper.

### DIFF
--- a/.changeset/avatar-fallback-theme-target.md
+++ b/.changeset/avatar-fallback-theme-target.md
@@ -2,6 +2,6 @@
 '@astryxdesign/core': patch
 ---
 
-[feat] Avatar: the fallback surface (initials and default icon) is now a direct theme target via the stable `astryx-avatar-fallback` class. Theme its background, text color, and font weight through the `avatar-fallback` component key (e.g. `components: { 'avatar-fallback': { base: { backgroundColor: '...' } } }`), replacing the previous internal derived vars. Per-size initials font size is still set on the `avatar` size tiers.
+[feat] Avatar: the fallback surface (initials and default icon) is now a direct theme target via the stable `astryx-avatar-fallback` class. Theme its background, text color, font weight, and per-size font size through the `avatar-fallback` component key (e.g. `components: { 'avatar-fallback': { base: { backgroundColor: '...' }, 'size:sm': { fontSize: '...' } } }`), replacing the internal `--_avatar-fallback-*` derived vars.
 
 @cixzhang
--- a/.changeset/avatar-fallback-theme-target.md
+++ b/.changeset/avatar-fallback-theme-target.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] Avatar: the fallback surface (initials and default icon) is now a direct theme target via the stable `astryx-avatar-fallback` class. Theme its background through the `avatar-fallback` component key (e.g. `components: { 'avatar-fallback': { base: { backgroundColor: '...' } } }`) instead of the previous internal derived var.
+
+@cixzhang
--- a/.changeset/avatar-fallback-theme-target.md
+++ b/.changeset/avatar-fallback-theme-target.md
@@ -2,6 +2,6 @@
 '@astryxdesign/core': patch
 ---
 
-[feat] Avatar: the fallback surface (initials and default icon) is now a direct theme target via the stable `astryx-avatar-fallback` class. Theme its background through the `avatar-fallback` component key (e.g. `components: { 'avatar-fallback': { base: { backgroundColor: '...' } } }`) instead of the previous internal derived var.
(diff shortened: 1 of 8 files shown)


cixzhang:
Pushed 6f2d1113 — stories only, no component or theme-generation change.

The two themed-fallback stories were picking tokens that resolve to the default fallback's own colors in the monochrome shipped themes, so the "themed" row rendered indistinguishable from the default row. Reading computed styles off the rendered elements under `neutral`:

| story | row | before | after |
|---|---|---|---|
| Themed Fallback Background | default | `rgba(0,0,0,0.06)` / `rgb(82,82,82)` | unchanged |
| Themed Fallback Background | **themed** | `rgb(241,241,241)` / `rgb(38,38,38)` | **`rgb(38,38,38)` / `rgb(255,255,255)`** |

`--color-accent-muted` is `#f1f1f1` in neutral, and the default wash `--color-neutral` is `rgba(0,0,0,0.06)` → `#f0f0f0` over white. One shade apart. The Type Scale story additionally set `color: var(--color-text-secondary)`, which is the fallback's own default color — a literal no-op — on top of the same invisible wash.

Swapped to tokens that contrast in every shipped theme (`--color-accent`/`--color-on-accent` for the background story, the blue pair for the type-scale wash), and left a comment in each explaining why a muted token is wrong here so it does not regress. Verified across `neutral` / `stone` / `y2k`: themed row now lands on the theme accent (`rgb(38,38,38)` / `rgb(37,37,42)` / `rgb(45,36,27)`) with white initials, and the icon fallback picks up the same fill.

To be clear about what this was: **the theming mechanism in this PR was working the whole time** — the `.astryx-avatar-fallback` rule lands in `@layer astryx-theme`, the `@scope` matches, and it beats the StyleX base class exactly as described. Only the demo tokens were at fault.

One separate thing worth a look before merge: the changeset is `patch`, but dropping the `avatar` entry from `derivedVarRegistry` is a behavior break for any consumer theme that tints the fallback via `components.avatar.base.backgroundColor` — that now paints only the wrapper, mostly hidden behind the fallback's own wash. No theme in this repo does it, so nothing here catches it, but it may warrant `minor` plus a migration line.

PR #4724, opened 2026-08-05
feat(radio): add standalone RadioControl primitive; RadioListItem composes it

freddymeta:
## What

Adds a standalone **`RadioControl`** primitive — the radio input + its visual dot (`astryx-radio` / `astryx-radio-dot`) — that takes everything as props and works **outside** a `RadioList`. `RadioListItem` now composes it.

Today `RadioListItem` reads `RadioListContext` for `name`/`value`/`onChange`/`size`/`isDisabled`/`isRequired` and **throws** if rendered outside a `RadioList`:

```
RadioListItem must be used within an RadioList
```

So there was no way to render a single radio control (e.g. a bespoke layout, a cell, a custom grouping) without building a whole `RadioList`. `RadioControl` closes that gap: it's the self-contained control, and `RadioListItem` becomes a thin composition of it (context → props), keeping its label/description/row layout.

## API

```tsx
// Standalone (no RadioList needed):
<RadioControl name="plan" value="pro" checked={value === 'pro'} onChange={setValue} label="Pro" />
```

- `checked`, `onChange(value)`, `value`, `name`, `size?` (`sm`/`md`), `isDisabled?`, `isRequired?`, `id?`, `aria-describedby?`
- `label?` — accessible name for a standalone control (applied as `aria-label`). **Omit it** when an external `<label htmlFor>` already names the input, as `RadioListItem` does — so there's no double-naming.
- `keepFocusableWhenDisabled?` — preserves the group-disabled-with-message behavior (`aria-disabled` + detached `form` instead of native `disabled`, so a reason tooltip stays keyboard-reachable).

## Unchanged behavior

`RadioListItem`'s public API and rendered output are **identical** — its existing test suite (44 tests) passes untouched, including the `astryx-radio`/`astryx-radio-dot` targets, checked/disabled data-attrs, the focusable-disabled path, and the `<label htmlFor>` wiring. The `themeProps('radio')`/`themeProps('radio-dot')` calls moved verbatim into `RadioControl` (same directory, same targets), so the theming surface is unchanged.

## Testing

- New `RadioControl.test.tsx` (8): renders a working standalone radio with an accessible name (no throw outside a RadioList), `checked` reflects the prop, click fires `onChange(value)`, disabled blocks it, renders `astryx-radio` + `astryx-radio-dot` (dot only when checked), and the `keepFocusableWhenDisabled` path stays focusable while blocking selection.
- Existing `RadioList.test.tsx` (44) unchanged and green; `themingTargets` guard green. Typecheck + lint clean.
- New `RadioControl.doc.mjs` + a `RadioControl` Storybook story (standalone + a small controlled group). Changeset included.

### DIFF
--- a/.changeset/radio-control-primitive.md
+++ b/.changeset/radio-control-primitive.md
@@ -2,6 +2,6 @@
 '@astryxdesign/core': patch
 ---
 
-[feat] Add a standalone `RadioControl` primitive — a self-contained radio input and circle that works on its own via props (no `RadioList` context). `RadioListItem` now composes it, with unchanged output and API.
+[feat] Add a standalone `RadioControl` primitive — a self-contained radio input and circle that works on its own via props (no `RadioList` context), for composing radios into bespoke surfaces (cards, table cells, custom rows). The API mirrors the checkbox family's control conventions: required `label` (accessible name, applied as `aria-label` per the icon-only `Button` convention), `isChecked`, `htmlName`, `onChange(value, e)`, `size`, `isDisabled`, `isRequired`, and `disabledMessage` (focusable-disabled with an AT-discoverable reason tooltip). `RadioListItem` composes it; its public API and rendered behavior are unchanged (its existing test suite passes untouched).
 
 @freddymeta
--- a/.changeset/radio-control-primitive.md
+++ b/.changeset/radio-control-primitive.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] Add a standalone `RadioControl` primitive — a self-contained radio input and circle that works on its own via props (no `RadioList` context). `RadioListItem` now composes it, with unchanged output and API.
+
+@freddymeta
--- a/apps/storybook/stories/RadioControl.stories.tsx
+++ b/apps/storybook/stories/RadioControl.stories.tsx
@@ -11,17 +11,17 @@ const meta: Meta<typeof RadioControl> = {
   argTypes: {
     label: {
       control: 'text',
-      description: 'Accessible name (aria-label) for a standalone control',
+      description: 'Accessible name (aria-label) for the control',
     },
     value: {
       control: 'text',
       description: 'Value reported when this radio is selected',
     },
-    name: {
+    htmlName: {
       control: 'text',
       description: 'HTML name shared by the radio group',
     },
-    checked: {
+    isChecked: {
       control: 'boolean',
       description: 'Whether the radio is selected',
     },
@@ -38,37 +38,28 @@ const meta: Meta<typeof RadioControl> = {
       control: 'boolean',
       description: 'Whether the radio is required',
     },
+    disabledMessage: {
+      control: 'text',
+      description: 'Reason shown when disabled (keeps the control focusable)',
+    },
   },
 };
 
 export default meta;
 type Story = StoryObj<typeof RadioControl>;
 
+// The control renders only the circle — pair it with your own visible text.
+// A radio is selected by choosing it; it is deselected by choosing another in
+// the same group (see ControlledGroup), never by clicking it again — so a lone
+// radio is not a toggle. This default shows a single selected control.
 export const Default: Story = {
-  // A single radio can't normally be un-selected by clicking it (native radio
-  // behavior — you deselect by choosing another in the group; see
-  // ControlledGroup). To make this standalone demo interactive, drive the
-  // checked state purely from onClick (which fires on every click), with a
-  // no-op onChange to satisfy the controlled input. Mixing onChange (fires
-  // only when the radio becomes checked) with onClick desyncs the toggle, so
-  // onClick is the single source of truth here.
-  render: args => {
-    const [checked, setChecked] = useState(args.checked ?? false);
-    return (
-      <RadioControl
-        {...args}
-        checked={checked}
-        onChange={() => {}}
-        onClick={() => setChecked(c => !c)}
-      />
-    );
-  },
   args: {
     label: 'Email',
-    name: 'notify',
+    htmlName: 'notify',
     value: 'email',
-    checked: true,
+    isChecked: true,
   },
+  render: args => <RadioControl {...args} onChange={() => {}} />,
 };
 
 export const Sizes: Story = {
@@ -76,18 +67,18 @@ export const Sizes: Story = {
     <div style={{display: 'flex', alignItems: 'center', gap: 24}}>
       <RadioControl
         label="Small"
-        name="sizes"
+        htmlName="sizes"
         value="sm"
         size="sm"
-        checked
+        isChecked
         onChange={() => {}}
       />
       <RadioControl
         label="Medium"
-        name="sizes"
+        htmlName="sizes"
         value="md"
         size="md"
-        checked
+        isChecked
         onChange={() => {}}
       />
     </div>
@@ -99,17 +90,17 @@ export const Disabled: Story = {
     <div style={{display: 'flex', alignItems: 'center', gap: 24}}>
       <RadioControl
         label="Disabled unchecked"
-        name="disabled"
+        htmlName="disabled"
         value="a"
-        checked={false}
+        isChecked={false}
         isDisabled
         onChange={() => {}}
       />
       <RadioControl
         label="Disabled checked"
-        name="disabled"
+        htmlName="disabled"
         value="b"
-        checked
+        isChecked
         isDisabled
         onChange={() => {}}
       />
@@ -117,6 +108,25 @@ export const Disabled: Story = {
   ),
 };
 
+// Disabled with a reason: the control stays focusable and surfaces the reason
+// on hover / keyboard focus, so it is discoverable by assistive technology.
+export const DisabledWithReason: Story = {
+  render: () => (
+    <RadioControl
+      label="Legacy mode"
+      htmlName="mode"
+      value="legacy"
+      isChecked={false}
+      isDisabled
+      disabledMessage="Locked by your administrator"
+      onChange={() => {}}
+    />
+  ),
+};
+
+// Grouping and single-select come from a shared htmlName: selecting one option
+// deselects the others. This is the correct way to build a radio group by
+// composing the control with your own layout and visible labels.
 export const ControlledGroup: Story = {
   render: () => {
     const [value, setValue] = useState('email');
@@ -132,9 +142,10 @@ export const ControlledGroup: Story = {
             key={opt.value}
             style={{display: 'flex', alignItems: 'center', gap: 8}}>
             <RadioControl
-              name="channel"
+              label={opt.label}
+              htmlName="channel"
               value={opt.value}
-              checked={value === opt.value}
+              isChecked={value === opt.value}
               onChange={setValue}
             />
             {opt.label}
--- a/apps/storybook/stories/RadioControl.stories.tsx
+++ b/apps/storybook/stories/RadioControl.stories.tsx
@@ -1,0 +1,138 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+import {useState} from 'react';
+import type {Meta, StoryObj} from '@storybook/react';
+import {RadioControl} from '@astryxdesign/core/RadioList';
+
+const meta: Meta<typeof RadioControl> = {
+  title: 'Core/RadioControl',
+  component: RadioControl,
+  tags: ['autodocs'],
+  argTypes: {
+    label: {
+      control: 'text',
+      description: 'Accessible name (aria-label) for a standalone control',
+    },
+    value: {
+      control: 'text',
+      description: 'Value reported when this radio is selected',
+    },
+    name: {
+      control: 'text',
+      description: 'HTML name shared by the radio group',
+    },
+    checked: {
+      control: 'boolean',
+      description: 'Whether the radio is selected',
+    },
+    size: {
+      control: 'select',
+      options: ['sm', 'md'],
+      description: 'Size of the radio control',
+    },
+    isDisabled: {
+      control: 'boolean',
+      description: 'Whether the radio is disabled',
+    },
+    isRequired: {
+      control: 'boolean',
+      description: 'Whether the radio is required',
+    },
+  },
+};
+
+export default meta;
+type Story = StoryObj<typeof RadioControl>;
+
+export const Default: Story = {
+  render: args => {
+    const [checked, setChecked] = useState(args.checked ?? false);
+    return (
+      <RadioControl
+        {...args}
+        checked={checked}
+        onChange={() => setChecked(c => !c)}
+      />
+    );
+  },
+  args: {
+    label: 'Email',
+    name: 'notify',
+    value: 'email',
+    checked: true,
+  },
+};
+
+export const Sizes: Story = {
+  render: () => (
+    <div style={{display: 'flex', alignItems: 'center', gap: 24}}>
+      <RadioControl
+        label="Small"
+        name="sizes"
+        value="sm"
+        size="sm"
+        checked
+        onChange={() => {}}
+      />
+      <RadioControl
+        label="Medium"
+        name="sizes"
+        value="md"
+        size="md"
+        checked
+        onChange={() => {}}
+      />
+    </div>
+  ),
+};
+
+export const Disabled: Story = {
+  render: () => (
+    <div style={{display: 'flex', alignItems: 'center', gap: 24}}>
+      <RadioControl
+        label="Disabled unchecked"
+        name="disabled"
+        value="a"
+        checked={false}
+        isDisabled
+        onChange={() => {}}
+      />
+      <RadioControl
+        label="Disabled checked"
+        name="disabled"
+        value="b"
+        checked
+        isDisabled
+        onChange={() => {}}
+      />
+    </div>
+  ),
+};
+
+export const ControlledGroup: Story = {
+  render: () => {
+    const [value, setValue] = useState('email');
+    const options = [
+      {label: 'Email', value: 'email'},
+      {label: 'SMS', value: 'sms'},
+      {label: 'Push', value: 'push'},
+    ];
+    return (
+      <div style={{display: 'flex', flexDirection: 'column', gap: 12}}>
+        {options.map(opt => (
+          <label
+            key={opt.value}
+            style={{display: 'flex', alignItems: 'center', gap: 8}}>
+            <RadioControl
+              name="channel"
+              value={opt.value}
+              checked={value === opt.value}
+              onChange={setValue}
+            />
+            {opt.label}
+          </label>
+        ))}
+      </div>
+    );
+  },
+};
--- a/apps/storybook/stories/RadioControl.stories.tsx
+++ b/apps/storybook/stories/RadioControl.stories.tsx
@@ -47,15 +47,18 @@ type Story = StoryObj<typeof RadioControl>;
 export const Default: Story = {
   // A single radio can't normally be un-selected by clicking it (native radio
   // behavior — you deselect by choosing another in the group; see
-  // ControlledGroup). To make this standalone demo interactive, we toggle on
-  // pointer/keyboard activation via onClick so you can flip it on and off.
+  // ControlledGroup). To make this standalone demo interactive, drive the
+  // checked state purely from onClick (which fires on every click), with a
+  // no-op onChange to satisfy the controlled input. Mixing onChange (fires
+  // only when the radio becomes checked) with onClick desyncs the toggle, so
+  // onClick is the single source of truth here.
   render: args => {
     const [checked, setChecked] = useState(args.checked ?? false);
     return (
       <RadioControl
         {...args}
         checked={checked}
-        onChange={() => setChecked(true)}
+        onChange={() => {}}
         onClick={() => setChecked(c => !c)}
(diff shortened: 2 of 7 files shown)


cixzhang (changes requested):
🔴 **Blocking — public API-convention violations on new surface in `packages/core/src/RadioList/RadioControl.tsx` (`name`/`checked`/`keepFocusableWhenDisabled` diverge from siblings in the same directory), an interactive element that can render with no accessible name, and `xstyle` applied to a zero-opacity node.** Also carries new public API from a contributor → `needs:code-review`.

`Triage: new API surface · non-breaking to existing consumers · high blast radius (net-new export from @astryxdesign/core) → deep path`

The refactor itself is clean and the `RadioListItem` invariance claim checks out. Everything below is about the new `RadioControl` surface.

### 🔴 1. Prop names contradict conventions this directory already sets

[API Conventions → HTML Attribute Collisions](https://github.com/facebook/astryx/wiki/API-Conventions) is explicit: `htmlName?: string; // maps to <input name="...">`. `RadioListProps.htmlName` (`RadioList.tsx:127`) already does exactly this, for exactly this attribute, one file over.

[API Conventions → Booleans](https://github.com/facebook/astryx/wiki/API-Conventions): *"Always prefix with `is` or `has`."* Both violations are shipped under `eslint-disable`:

```ts
// RadioControl.tsx:175
// eslint-disable-next-line @astryx/boolean-prop-naming
checked: boolean;

// RadioControl.tsx:219
// eslint-disable-next-line @astryx/boolean-prop-naming
keepFocusableWhenDisabled?: boolean;
```

`internal/eslint-plugin-astryx/boolean-prop-naming.js` has `checked: 'isChecked'` hardcoded in `NATIVE_BOOLEAN_SUGGESTIONS` — the rule is naming the fix. Per the blocking criteria, *"a lint-suppressed hardcode is still a blocker … Do not let a documented `eslint-disable` soften a bright-line violation into advisory."* (The only other two disables of this rule in `packages/core` are on a hook options type in `useResizable.ts`, not a public `*Props` interface.)

`keepFocusableWhenDisabled` is a second problem beyond naming: it exposes an internal mechanism as public API. `RadioList` and `CheckboxInput` both express this capability as **`disabledMessage?: string`** — one prop that supplies the reason *and* keeps the control focusable. `RadioListItem.tsx:123` derives `keepsFocusableForMessage` from exactly that. Per *"Adding a new prop — converge, don't diverge"*: reuse the sibling shape.

| Current | Sibling convention | Source |
| --- | --- | --- |
| `name: string` (required) | `htmlName?: string` | `RadioListProps:127`, `CheckboxInputProps` |
| `checked: boolean` | `isChecked` | wiki Booleans; eslint rule suggestion |
| `keepFocusableWhenDisabled?: boolean` | `disabledMessage?: string` | `RadioListProps:131`, `CheckboxInputProps` |
| `label?: string` → `aria-label`, never rendered | `label: string` (required, always rendered) + `isLabelHidden?: boolean` | `RadioListProps`, `RadioListItemProps`, `CheckboxInputProps` |

The `label` row is the one I'd push hardest on. `<RadioListItem label="Email">` renders visible text; `<RadioControl label="Email">` renders none. Same word, opposite behavior, same directory. The system's answer for "labelled but not visible" is already `label` + `isLabelHidden` — please use it.

### 🔴 2. A standalone `RadioControl` can ship with no accessible name

`label` is optional and the JSDoc/`.doc.mjs` actively tell you to omit it. Rendered with no `label` and no external `<label htmlFor>`:

```html
<input id="_r_c_" class="…" type="radio" value="email" name="notify">
```

No `aria-label`, no `aria-labelledby`, and the visible circle is `aria-hidden="true"`. A screen reader announces an unnamed radio. Blocking criteria: *"an interactive element without an accessible name."* Note both siblings make `label` **required** — [API Conventions → Required vs Optional](https://github.com/facebook/astryx/wiki/API-Conventions) lists `` `label` (interactive elements)`` as required. Making it required + `isLabelHidden` resolves this and finding #1 together.

### 🔴 3. `xstyle` / `className` / `style` land on the invisible input, not the root

```tsx
// RadioControl.tsx:308-326 — xstyle merged into styles.input
{...mergeProps(
  stylex.props(styles.input, wrapperSizeStyles[size], isDisabled && styles.inputDisabled, xstyle),
  className, style,
)}
```

`styles.input` is `position: absolute; opacity: 0`. Verified by rendering with `xstyle={{marginInlineStart: 40, outlineColor: 'red'}}`:

```
ROOT   class: RadioControl__styles.radioWrapper x1n2onr6 … (no probe atoms)
INPUT  class: RadioControl__styles.input … __scratch_review__overrides.probe x11hdunq xbutuof
CIRCLE class: astryx-radio md … (no probe atoms)
input computed opacity: 0 / position: absolute
```

So `<RadioControl xstyle={styles.spacing} />` silently does nothing visible. `RadioListItem` puts `xstyle` on its container (`RadioListItem.tsx:158`) — the root — which is the convention. **Fix:** move `xstyle`/`className`/`style` to the wrapper `<div>` and keep `ref` on the input.

### 🔴 4. The standalone control has no hover state — `radioScope` never reaches it

All four hover rules are gated on an ancestor marker:

```tsx
// RadioControl.tsx:77, 83, 91, 97
[stylex.when.ancestor(':hover', radioScope)]: { '@media (hover: hover)': … }
```

`radioScope` is applied in exactly one place in the repo — `RadioListItem.tsx:158`, `stylex.props(styles.container, !isDisabled && radioScope, xstyle)`. Verified ancestor chains:

```
in a RadioList:  .astryx-radio ← .radioWrapper ← Item.startContent ← .astryx-item
                 ← astryx-radio-list-item … xklqth5   ← the marker
standalone:      .astryx-radio ← .radioWrapper (root, no ancestors, no marker)
```

So the advertised primary use case — "works **outside** a `RadioList`" — renders a radio with **zero hover feedback**, on desktop, permanently. That contradicts the PR's "self-contained control" framing and [Design Conventions → State representation](https://github.com/facebook/astryx/wiki/Design-Conventions) ("every relevant state … is designed"). `radio.markers.stylex.ts`'s own docstring still says *"Scoped marker for RadioListItem ancestor selectors"* — the ownership boundary moved but the marker didn't. **Fix:** apply `radioScope` on `RadioControl`'s own wrapper (and let `RadioListItem` keep applying it on the row so row-hover still drives the circle).

### 🔴 5. New API surface from a contributor — not approvable by review alone

Per `.github/instructions/packages.instructions.md`: *"If a PR adds a new prop or otherwise changes API surface … the AI reviewer must not approve or merge it — even if it's clean, additive, and passing CI."*

**⚠️ Needs human/maintainer judgment on the API surface.** Specifically:
- **No linked spec issue.** New public component surface must link an issue that ran the [Component Specification Protocol](https://github.com/facebook/astryx/wiki/Component-Specification-Protocol). Please add one before the API is judged on its merits.
- Whether the right primitive is `RadioControl` at all, vs. relaxing `RadioListItem`'s "must be used within a RadioList" throw, is a maintainer call. Route naming disputes to [API Arbitration](https://github.com/facebook/astryx/wiki/API-Arbitration).
- New surface should be vibe-tested.

### 🟡 6. `keepFocusableWhenDisabled` produces a focusable control with no focus ring

`styles.radioWrapperFocus` is gated on `!isDisabled` (`RadioControl.tsx:294`), but the focusable-disabled path *is* `isDisabled`. Verified class lists:

```
keepFocusable ROOT: …radioWrapper x1n2onr6 … xxk0z11                        ← no focus atoms
enabled ROOT:       …radioWrapper x1n2onr6 … xxk0z11 …radioWrapperFocus
                                                     x1a2a7pz x1irc7jg x1wfwxd8 xdjuwb3 x16rqkct
```

So a keyboard user tabs into it and sees nothing (WCAG 2.4.7). This is **pre-existing** — `RadioListItem` had the identical `!isDisabled &&` guard on `main`, and it fires today whenever a `RadioList` is disabled with a `disabledMessage`. Filing it advisory rather than blocking because the PR didn't introduce it — but the PR *promotes* it into documented public API, which is the moment to fix it. Suggest gating on `!isDisabled || keepsFocusable`.

### 🟡 7. `{...rest}` is spread last and can clobber the a11y wiring

`{...rest}` (`RadioControl.tsx:327`) comes after `type`, `checked`, `disabled`, `aria-disabled`, `form`, `required`, and `aria-label`. Verified:

```tsx
<RadioControl label="Email" … aria-label="OVERRIDDEN" role="button" aria-disabled="true" />
// → aria-label: OVERRIDDEN | role: button | aria-disabled: true
```

A consumer can silently destroy the radio role and the disabled semantics. [API Conventions → Composed Component Prop Forwarding](https://github.com/facebook/astryx/wiki/API-Conventions), rule 1: *"Never let `{...rest}` overwrite explicitly-set props."* Rule 4 offers `{...rest}`-first as the alternative. Weight: `RadioListItem` already spreads rest last onto its container, so this is a repo-wide pattern rather than a new sin — but on an `<input>` the blast radius is a11y-critical, not cosmetic.

### 🟡 8. `onChange` drops the event

`onChange: (value: string) => void` matches `RadioListProps.onChange`, so it's family-consistent. But the closest *control-level* sibling is `CheckboxInputProps.onChange?: (checked, e: ChangeEvent<HTMLInputElement>) => void`. A control-level primitive that never hands back the event can't support `preventDefault`, modifier keys, or `currentTarget`. Worth deciding deliberately for a primitive, not inheriting the group's shape by default.

### 🟡 9. The `Default` story teaches a broken pattern (and is the autodocs canonical example)

```tsx
// RadioControl.stories.tsx:74-84
onChange={() => {}}
onClick={() => setChecked(c => !c)}
```

The story is in a low-risk area (`apps/storybook/**` never gates), but `tags: ['autodocs']` makes this the first example a consumer reads for the new component, and it teaches "stub the required `onChange`, drive state from `onClick`." Keyboard selection in a real radio group fires `change`, not `click`, so a consumer copying this gets a control that doesn't respond to the keyboard. I'd also read the fact that a working single-radio demo needed an `onClick` workaround as evidence about the API (see #8), not just about the story. Also: the 7-line narration comment above it is more than *"comment why, not what"* asks for.

### 🟡 10. Smaller notes

- `RadioControl.doc.mjs` has no `theming` block, so a standalone consumer can't discover `astryx-radio` / `astryx-radio-dot` from the component's own docs. `themingTargets.test.ts` doesn't catch this — it only reads `<dir>.doc.mjs`, i.e. `RadioList.doc.mjs`. Not a dropped surface, just an undiscoverable one.
- The PR body lists `aria-describedby?` as part of the API; `.doc.mjs` doesn't. It only works via the `...rest` passthrough.
- The verbatim-moved raw dimensions (`20/24/8/10`, `2px solid`, `borderRadius: '50%'`, `opacity: 0.5`) are pre-existing, not introduced here, so not blocking on them — but they're now sitting in a brand-new file, which is when "semantic tokens only" is cheapest to satisfy.

### 🟢 Verified clean

<details>
<summary><strong>What was confirmed rather than assumed</strong></summary>

| Claim | Result |
| --- | --- |
| Themeable surface preserved | ✅ `themeProps('radio', {size, checked, disabled})` + `themeProps('radio-dot', {size})` moved verbatim; `RadioList.doc.mjs` `theming.targets` (L13–17) untouched; no target pinned or dropped |
| `themingTargets` guard | ✅ 277/277 pass (guard scans the whole `RadioList/` dir, so the cross-file move is invisible to it) |
| `RadioListItem` API + output identical | ✅ 44 existing tests pass untouched; props forwarded 1:1; the `if (isDisabled) return` guard is semantically equivalent |
| New tests | ✅ 8/8 pass; 52/52 for the directory |
| Typecheck | ✅ `tsc -p packages/core --noEmit` clean |
| `check:changesets` / `check:sync` / `sync:exports:check` | ✅ all pass |
| Changeset conventions | ✅ `[feat]` category, `@freddymeta` line, `patch` bump, `@astryxdesign/core` — correct, and correctly *not* `[breaking]` |
| No hardcoded colors | ✅ every color is a `colorVars` token or a `color-mix()` of tokens; `CanvasText` under `forced-colors` is a system color, correct usage |
| No raw CSS / StyleX workarounds | ✅ `stylex.when.ancestor`, `:has(:focus-visible)`, `@media (hover: hover)` all used natively |
| Component surface complete | ✅ `.tsx` + `.test.tsx` + `.doc.mjs` + story + `index.ts` export |
| `displayName`, exported prop types | ✅ `RadioControl.displayName`, `RadioControlProps`, `RadioControlSize` all exported |
| `ref` pattern | ✅ `ref?: React.Ref<HTMLInputElement>` matches the repo (147 files use it; only 2 still use `forwardRef`) |
| `@example` fence untagged | ✅ plain ``` ``` ``` |
| Public-repo leak | ✅ none in committed text |

</details>

Thanks for this — the extraction itself is careful work, and the invariance claim held up under a real check, which is not the common case. #1–#4 are the bright-line ones with concrete fixes; #5 means that even once they're fixed, a maintainer signs off on whether `RadioControl` should exist in this shape. Please link a spec issue when you pick it back up.

freddymeta:
Thanks for the thorough review — the convergence points are right, and the "same word, opposite behavior, same directory" call on `label` is the one that reframed this for me. I anchored the draft on `RadioListItem`'s internals + native HTML shapes instead of the sibling conventions (`CheckboxInput`), which is exactly the divergence the API-Conventions "match existing siblings" rule exists to prevent.

Two responses:

**On #5 (new API surface needs a spec + maintainer judgment):** opened #4755 as the spec. It runs the problem statement, prior art (`CheckboxInput`/`CheckboxListItem` as the in-system template), the use-case enumeration, and three candidate shapes with an arbitration:

- **A** — bare `RadioControl` (control only, `label` → `aria-label`)
- **B** — full `RadioInput`, the `CheckboxInput` twin (visible `label` + `isLabelHidden` + `description` + `status` + `disabledMessage`), with `RadioListItem` composing it the way `CheckboxListItem` composes `CheckboxInput`
- **C** — relax `RadioListItem`'s throw instead of adding a new export

I lead with **B** because it makes every shared prop mean the same thing across the checkbox and radio families, which resolves #1–#4 structurally rather than prop-by-prop. But which shape (and name) ships is a maintainer call — that's the point of the issue. Happy to vibe-test A/B/C with naive prompts across the use cases before locking it.

**On #1–#4, #6–#10 (convergence):** these apply to whichever shape wins, and I've captured them all in the issue's "Convergence Requirements" section — `htmlName?`/`isChecked`/required-`label`/`disabledMessage` (no `eslint-disable`), accessible name always present, `xstyle`/`className`/`style` on the root, `radioScope` on the control's own wrapper for standalone hover, focus ring on the disabled-with-reason path, `{...rest}` precedence + `composeEventHandlers`, `onChange(value, e)`, and the doc `theming`/tokens/`aria-describedby` fixes.

Holding the implementation here until the shape is settled in #4755, so I don't converge to A and then have to re-converge to B. Will push the rewrite to this PR once it's decided.

freddymeta:
Pushed the convergence (`b961e2f`). Shape settled as **`RadioControl`** (bare control), per the spec discussion in #4755 — `*Input` names a labeled form _field_ and a lone radio has no valid standalone-field use, so the missing piece is the composition primitive (the circle), not a field. Walking your points:

**Blockers**
1. **Naming** ✅ `htmlName?` / `isChecked` / `onChange(value, e)` / `disabledMessage?`. The `@astryx/boolean-prop-naming` `eslint-disable`s are gone — the names are now convention-compliant. `keepFocusableWhenDisabled` is removed; the control reads `RadioListContext` directly for the group-disabled-with-message case, exactly as `CheckboxInput` reads `CheckboxListContext`.
2. **Accessible name** ✅ `label` is now **required**, applied as `aria-label` — this follows the documented **icon-only `Button`** convention (control with no visible text), so a standalone control can't ship unnamed. This resolves #1's `label` row too: for a _control_ (not a field), `label → aria-label` is convention-aligned, not "same word, opposite behavior."
3. **`xstyle`/`className`/`style` on the root** ✅ moved to the wrapper `<div>`; `ref` stays on the input. New test asserts `className` lands on the wrapper and `data-*` passes through to the input.
4. **Standalone hover** ✅ `radioScope` is applied on the control's own wrapper, so a standalone control gets the hover treatment; `RadioListItem` still applies it on the row so row-hover drives the circle.

**Advisories**
6. **Focus ring on the disabled-with-reason path** ✅ gated on `!isDisabled || isFocusableDisabled`.
7. **`{...rest}` precedence** ✅ `{...rest}` is spread first; the contract props (`type`, `name`, `checked`, `aria-*`, `onChange`) are set after it, so a consumer can't clobber the radio semantics. New test: passing `type="checkbox"` still renders `type="radio"`.
8. **`onChange` hands back the event** ✅ `(value, e)`, matching `CheckboxInput`. `RadioListItem` adapts it to the group's value-only `onChange` so `RadioList`'s public contract is unchanged.
9. **Story** ✅ rewritten to the converged API; teaches grouping via shared `htmlName` and drops the lone-radio toggle hack (the anti-pattern the vibe test confirmed).
10. **Docs** ✅ `RadioControl.doc.mjs` now has a `theming` block (`astryx-radio` / `astryx-radio-dot`) so a standalone consumer can discover the targets from the primitive's own page; `aria-describedby` is a real supported prop (used by `RadioListItem` for the description), not just a `...rest` passthrough.

**Unchanged:** `RadioListItem`'s public API and rendered behavior — its 44 tests pass untouched. Full suite: **332 pass** (11 RadioControl incl. new xstyle-root / rest-precedence / disabledMessage-focusable coverage, 44 RadioList, 277 themingTargets). Typecheck + lint + `check:sync` + changeset all green.

Still your call on the API surface per the contributor gate — the shape and convergence are now settled against the conventions; happy to adjust.

freddymeta:
Heads up on the red checks on `b961e2f` — they look like a runner-provisioning outage on this CI run, not the change:

- `test`, `check-components`, `build-sandbox`, and `docsite-test` all failed at **step #1 "Set up job"** (runner setup), before any code executes.
- Every job that got a runner passed: `lint`, `smoke-test`, `dependency-check`, `check-scope` (+ CLA, Vercel).
- Prior heads of this PR passed `test`/`docsite-test`/etc. on the same paths, and locally the full suites are green (332 core tests, 303 docsite tests, typecheck, lint, `check:sync`, changeset).

A re-run should clear them — I don't have permission to re-trigger the workflow as a contributor (same reason "Re-request blocking reviewers" / "Flag review signals" show failure). Flagging rather than treating it as a code failure. Happy to push a trivial no-op commit to re-trigger if you'd prefer that over a re-run.

freddymeta:
All 10 points from the review are addressed in `b961e2f9e`. Shape settled as **RadioControl** (bare control, NOT RadioInput) — `*Input` = labeled field, lone radio has no standalone-field use, so the missing piece is the composition primitive. Corrected the earlier B-recommendation on #4755.

### Point-by-point

**🔴 #1 — Prop names converged:**
- `name` → `htmlName?: string` (optional; generated fallback)
- `checked` → `isChecked: boolean` (no eslint-disable)
- `keepFocusableWhenDisabled` → `disabledMessage?: string` (the control now owns its own `useTooltip`, matching CheckboxInput's internal `isFocusableDisabled` derivation from context)
- `label` required (was optional) — see #2

**🔴 #2 — Always has an accessible name:** `label: string` is required, applied as `aria-label`. Follows the icon-only Button convention (interactive control with no visible text). A standalone control can no longer ship unnamed.

**🔴 #3 — `xstyle`/`className`/`style` on root wrapper:** moved from the zero-opacity input to the wrapper `<div>`. Consumer overrides now visibly affect the component.

**🔴 #4 — `radioScope` on the control's own wrapper:** standalone hover works. RadioListItem still applies it on the row (row-hover drives the circle), so both paths work.

**🟡 #5 — Spec issue linked:** #4755. Shape settled (A=bare control), arbitration run (C eliminated, A chosen over B), awaiting maintainer API-surface sign-off.

**🟡 #6 — Focus ring on disabled-with-reason:** gated on `!isDisabled || isFocusableDisabled` now, not just `!isDisabled`.

**🟡 #7 — `{...rest}` precedence:** spread *before* the contract props (type, name, checked, aria-*, onChange) so a consumer can't clobber radio semantics. Handlers composed where needed.

**🟡 #8 — `onChange(value, e)`:** signature now `(value: string, e: ChangeEvent<HTMLInputElement>) => void`, matching CheckboxInput's control-level event handback. RadioListItem adapts it to the group's value-only callback internally.

**🟡 #9 — Story fixed:** Default story is gone (was the lone-toggle anti-pattern); replaced with Sizes, Disabled, DisabledWithReason, and ControlledGroup (grouping via shared htmlName). The story teaches the real use case, not a broken demo.

**🟡 #10 — Doc theming block:** `RadioControl.doc.mjs` now has a `theming` section exposing `astryx-radio` / `astryx-radio-dot` targets (both EN and zh). `aria-describedby` is wired as a first-class prop (destructured out of rest, composed into the final `aria-describedby` alongside the disabledMessage tooltip's describedBy).

### Verified clean

- 332 tests pass (11 RadioControl + 44 RadioList + 277 themingTargets)
- Typecheck clean, lint clean, check:sync ✅, check:changesets ✅
- RadioListItem output unchanged (44 original tests green)
- CI hit a runner infra flake (all jobs failed at "Set up job" provisioning) — needs a re-run

rubyycheung (approved):
Looks good!

rubyycheung (approved):

freddymeta:
Closing — superseded. Won't merge.

PR #4728, opened 2026-08-05
feat(cli): app shell — emit any template inside your design system's shell

josephfarina:
## Summary
Page templates are **content-only** by rule: they root at `Layout`/`Center` and never render their own `AppShell`, because whoever consumes a template owns the chrome. [Contributing-Templates](https://github.com/facebook/astryx/wiki/Contributing-Templates) spells out why — *"the host swaps in its shell around the template... if the template also shipped an `AppShell`, the host would have to branch"*.

That makes the shell a single swappable layer, and this PR exposes it.

```bash
astryx template dashboard              # content-only (unchanged default)
astryx template dashboard --with-shell # + the project's app shell
```

- **Opt-in.** Without `--with-shell` the bare template is emitted exactly as authored, so today's output does not change.
- **Core provides the default** (`AppShell`). An integration replaces it with the new `appShell` field in `astryx.integration.*`, pointing at a module that default-exports an `AstryxAppShell` (`{component, from, importKind?, props?, description?}`). Parameterizing it (`satisfies AstryxAppShell<AppFrameProps>`) makes typos and non-serializable props compile errors.
- **You always know whose shell you got.** Naming the component and module is the whole contract; the CLI adds the wrapper and its import as one unit and says what it did.
- **Exactly one shell.** Two integrations claiming it is a reported conflict, not a nesting order.
- **Pure output-layer.** Nothing on disk is edited. Validated at the load boundary, dry-run by `astryx validate-integration` (`missing_app_shell` / `invalid_app_shell`), and run through the codemod engine's `validateOutput` safety net so it can never emit unparseable source.

What the CLI prints:

```
$ astryx template blank
Content-only page. Add --with-shell to wrap it in MetaAppFrame from @xds/meta.

$ astryx template blank --with-shell
Wrapped in MetaAppFrame from @xds/meta — this project's app shell, replacing the default AppShell.
  internal shell: nav, search, and the standard app chrome

$ astryx template shell-top-nav --with-shell
shell-top-nav is an app shell already, so --with-shell changed nothing.
```

## Two problems found while building it
- **The `Shell -` category is not a reliable signal.** `blank` is categorized `Shell - Blank` but roots at `Layout` — and it is the template most likely to be used with `--with-shell`. Excluding by category would have silently refused to wrap it. Nesting is instead prevented by checking the **rendered root element**, which is ground truth and needs no category hygiene. Only three templates actually root at `AppShell`, and all three are the `Shell -` demos.
- **Owner-exclusion would have made the default shell useless.** The engine's "a transform never rewrites its owner's templates" rule meant core's shell could never wrap core's templates. It is now an explicit `skipOwnPackage` that the default opts out of.

## Hardening (earlier commit, still in)
An adversarial chaos tier found four engine defects, each fixed and pinned by a named regression, and each mutation-tested (reverting the fix individually turns the suite red): scope-blind `export default Name` resolution, a wrap emitting an un-imported component, a colliding binding producing a duplicate declaration, and a missing template context throwing. The notice is also a pure, testable function whose third-party text is flattened, so a shell author cannot forge notice lines or write terminal escapes.

## Test plan
- [x] `packages/cli/api/template/app-shell-integration.test.mjs` — 30 e2e cases: opt-in default, integration override, core fallback, one-slot conflict, never-nested, blocks, owner templates, scaffolding, hostile load boundary
- [x] `packages/cli/api/template/transform` — engine, `chaos`, `chaos-extreme` (184 hostile cases), `regressions` (#1–#10), `multi-integration`
- [x] `packages/cli/authoring/app-shell/parse.test.mjs` + `clients/cli/commands/template-notice.test.mjs`
- [x] Full CLI suite green at 2987 (the 2 failures are pre-existing macOS case-sensitivity tests that also fail on `main`)
- [x] `check:cli-structure`, `check:changesets`, drift, and repo `lint` clean; `typecheck:strict` adds no new errors (the 7 present are pre-existing on `main`)
- [x] Manually exercised end to end against a temp consumer project with and without an integration

### DIFF
--- a/.changeset/cli-integration-template-transforms.md
+++ b/.changeset/cli-integration-template-transforms.md
@@ -1,0 +1,15 @@
+---
+'@astryxdesign/cli': patch
+---
+
+[feat] app shell — emit any template inside your design system's shell, opt-in and never on disk.
+
+Page templates are content-only by rule: they root at `Layout`/`Center` and never render their own `AppShell`, because whoever consumes a template owns the chrome. That makes the shell a single swappable layer, and this exposes it.
+
+- `astryx template <id> --with-shell` wraps the emitted page in the project's app shell — adding the component and its import as one unit. Without the flag the bare, content-only template is emitted exactly as authored, so the default is unchanged.
+- Astryx core provides the default shell (`AppShell`). An integration replaces it by pointing the new `appShell` field in `astryx.integration.*` at a module that default-exports an `AstryxAppShell` (`{component, from, importKind?, props?, description?}`), exported from `@astryxdesign/cli/authoring`. Parameterizing it with the shell's props type (`satisfies AstryxAppShell<AppFrameProps>`) makes typos and non-serializable props compile errors.
+- The CLI always states which shell you got — the component, the package, and whether it replaced the default — and on a bare page it points at `--with-shell`. `--json` reports the applied package via `transformedBy`.
+- A project has exactly one shell: if two integrations declare one, the first in config order wins and the clash is reported. The shell is never nested inside a template that already renders one, never applied to blocks, and never applied to its own package's templates.
+- A pure output-layer: the on-disk templates are never modified. Declarations are validated at the load boundary, dry-run by `astryx validate-integration` (`missing_app_shell` / `invalid_app_shell`), and a broken shell is skipped with a warning instead of breaking the command. The rewrite runs through the codemod engine's shared `validateOutput` safety net, so it can never emit unparseable source.
+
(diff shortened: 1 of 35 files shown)


josephfarina:
Closing this one. It's a month old with no review, and 5k lines is too much to carry while the manifest-key forward-compat story in #5438 is still unsettled -- this PR adds an `appShell` key to `astryx.integration.*`, so it lands squarely on top of that question.

I still want the feature. Re-pitching it once #5438 is in, as something smaller.

PR #4735, opened 2026-08-05
fix(core): cancel pending long-press when a second finger joins mid-press

### LINKED ISSUE #4734: useLongPress: pending press not cancelled when a second finger joins mid-press
`onTouchStart` and `onTouchMove` only check `e.touches.length` on their own event. If a second finger touches down *after* a single-finger press has already started the timer (e.g. a pinch-to-zoom gesture), it falls through the `touches.length !== 1` guard without ever calling `clear()` — the pending timer keeps running and `onLongPress` can still fire with the stale first-finger point in the middle of a two-finger gesture.

### Regression test

```tsx
it('cancels when a second finger touches down mid-press', () => {
  const onLongPress = vi.fn();
  const { result } = renderHook(() => useLongPress({ onLongPress }));

  act(() => {
    result.current.onTouchStart(touchEvent([{ clientX: 0, clientY: 0 }]));
    // A second finger joins before the press fires (e.g. pinch-to-zoom).
    result.current.onTouchStart(
      touchEvent([
        { clientX: 0, clientY: 0 },
        { clientX: 50, clientY: 50 },
      ]),
    );
    vi.advanceTimersByTime(500);
  });
  expect(onLongPress).not.toHaveBeenCalled();
});
```

### Cause

`packages/core/src/hooks/useLongPress.ts` — `onTouchStart`/`onTouchMove` return early on `touches.length !== 1` without calling `clear()`, so a timer already pending from a single-finger touch is never cancelled once a second finger joins.

Happy to open a PR extending both handlers to cancel the pending press in that case, plus the regression test above.

alex-js-ltd:
Fixes #4734

## Summary
- `onTouchStart`/`onTouchMove` only checked `touches.length` on their own event, so a second finger joining *after* a single-finger press had already started the timer (e.g. pinch-to-zoom) fell through the `touches.length !== 1` guard without ever cancelling it
- Both handlers now call `clear()` when `touches.length !== 1`, whether that's a fresh multi-touch start (no-op, nothing pending) or a second finger arriving mid-press (cancels the pending long-press)

## Test plan
- [x] Added a regression test (`cancels when a second finger touches down mid-press`) that fails on `main` and passes with this fix
- [x] `npx vitest run packages/core/src/hooks/useLongPress.test.tsx` — 9/9 passing
- [x] `npx eslint packages/core/src/hooks/useLongPress.ts` — clean

### DIFF
--- a/.changeset/use-long-press-multitouch-cancel.md
+++ b/.changeset/use-long-press-multitouch-cancel.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] useLongPress: cancel the pending long-press when a second finger joins mid-press. Previously `onTouchStart` and `onTouchMove` only checked `touches.length` on their own event, so a second finger arriving after a single-finger press had already started the timer (e.g. a pinch-to-zoom gesture) fell through the `touches.length !== 1` guard without ever clearing it — `onLongPress` could still fire with the stale first-finger point mid-gesture. No API change.
+@alex-js-ltd
--- a/packages/core/src/hooks/useLongPress.test.tsx
+++ b/packages/core/src/hooks/useLongPress.test.tsx
@@ -81,6 +81,24 @@ describe('useLongPress', () => {
     expect(onLongPress).not.toHaveBeenCalled();
   });
 
+  it('cancels when a second finger touches down mid-press', () => {
+    const onLongPress = vi.fn();
+    const {result} = renderHook(() => useLongPress({onLongPress}));
+
+    act(() => {
+      result.current.onTouchStart(touchEvent([{clientX: 0, clientY: 0}]));
+      // A second finger joins before the press fires (e.g. pinch-to-zoom).
+      result.current.onTouchStart(
+        touchEvent([
+          {clientX: 0, clientY: 0},
+          {clientX: 50, clientY: 50},
+        ]),
+      );
+      vi.advanceTimersByTime(500);
+    });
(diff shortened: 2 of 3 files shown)


cixzhang (approved):
🟢 **No review blockers found** — the `needs:code-review` label is the only gate; nothing in the diff blocks.

Triage: bug fix · non-breaking · low blast radius (`ContextMenu` is the only in-repo consumer) → fast path.

Thanks for this — reporting the bug with a ready-made regression test and then following through with the fix is exactly the shape that makes a contribution easy to take.

**Verified locally** rather than taken on faith: your new test against `main`'s hook fails (1 failed / 8 passed, `onLongPress` called once), and passes with the fix (9/9). Red → green on the real reported bug.

A few other things I checked:

- `clear` is `useCallback([])`-stable and both dependency sets are correct, so no handler churn.
- Pulling `disabled` into its own early return preserves the existing behavior — a disabled hook still returns before touching the timer, same as before.
- The second-finger-lifts case can't silently re-arm: `clear()` nulls `startRef`, so the subsequent `onTouchMove` bails on the null guard instead of resuming a stale press.
- No breaking change — `useLongPress` is public, but no consumer can reasonably depend on a press firing mid-pinch.
- Changeset is correct: `patch`, `[fix]` category, contributor line.

Two nits, neither worth another round — noting them for next time:

- The changeset body restates the full root cause; house style leans big-picture, so a couple of lines would do.
- The two `// Multi-touch cancels…` comments narrate *what* the line does rather than *why*; the guard reads clearly on its own.

Approving and merging.

PR #4738, opened 2026-08-05
fix(SideNav): remove redundant border on collapsed child-menu popover

### LINKED ISSUE #4731: SideNav: collapsed child-menu popover shows an unintended rectangular inner border
## Problem

When SideNav is collapsed and a parent navigation group opens its child-menu popover, the menu surface appears to contain an additional sharp-cornered rectangular border. It is visible inside/alongside the rounded Popover surface, so the result looks like an accidental nested panel rather than one coherent menu.

This is especially noticeable for groups with several second-level items (for example, Governance → Manage Entities / Prepare Board / File & Report / Advise).

## Expected

The collapsed child menu should read as a single Popover surface:

- no extra inner rectangular outline, or
- if an inner container is intentional, its border/background should inherit or be clipped to the Popover radius.

## Why it matters

The visible right-angled frame conflicts with the Popover elevation and rounded surface, making the navigation feel visually broken in an otherwise tokenized UI.

## Suggested direction

Use the Popover surface as the only bordered surface for collapsed SideNav child menus. If the nested SideNav section needs a container for layout, make that container borderless or clip it to the outer radius.

## Reproduction

1. Render a SideNav with a collapsible parent section and child navigation items.
2. Collapse the SideNav.
3. Open that parent section’s child menu.
4. Observe the inset rectangular outline around the menu content.

I can provide a screenshot or a minimal reproduction if helpful.

## Screenshot from official template

> **Image placeholder — Shell / Side Nav template**  Image
> Upload the screenshot showing the inset sharp-cornered rectangular outline inside the rounded collapsed SideNav child-menu popover.

This example is from an official Astryx template, not a consumer-specific implementation.

HelloOjasMutreja:
## Summary

Closes #4731. When SideNav is collapsed and a parent navigation group opens its child-menu popover, the menu shows an extra sharp-cornered rectangular border sitting inside the rounded Popover surface, reading as an accidental nested panel.

## Root cause

`usePopover`'s own `render()` supplies the popover surface around whatever content it wraps, by default (`hasSurface: true`): background, `--radius-container`, and `--shadow-low`. It carries no border of its own. `SideNavItem`'s `popoverSurface` style, applied to the content passed into that `render()` call, added its own `borderWidth`/`borderStyle`/`borderColor` with no radius override. That border wasn't a duplicate of anything, it was the only border in the stack, and its 0-radius square corners cut across the outer surface's 12px-radius arcs instead of nesting inside them.

Checked every other `usePopover` consumer in the codebase (`DropdownMenu`, `TopNavMenu`, `MultiSelector`, `Selector`, `PowerSearch`, date inputs, `BaseTypeahead`, `TabMenu`, `TopNavHeading`, `TopNavMegaMenu`) — none add their own border around the popover content. `SideNavHeading.tsx`, the sibling file in this same directory, has the directly comparable case: its `popoverContent` style (padding + `overflow: hidden`) carries no border either.

## Fix

Removed the border properties from `popoverSurface`, keeping only the padding/margin/minWidth layout it legitimately needs. The popover's edge is now defined by the outer surface's shadow alone, matching every other popover in the system.

## Testing

- Full `SideNav` test suite passes (99/99).
- Typecheck and lint clean.
- **Visual verification**: I couldn't render a before/after locally (Storybook fails to build in this environment, a Rolldown/Vite bundler crash unrelated to this change, reproduces even on a clean `upstream/main` checkout). @cixzhang independently verified the fix visually against a local Storybook and confirmed the geometry: the inner div was full-height and flush to the outer surface's top/bottom edges, which is exactly what produces the reported artifact. Thank you for running that down.

### DIFF
--- a/.changeset/sidenav-remove-the-redundant-inner-border-on-col-0okn.md
+++ b/.changeset/sidenav-remove-the-redundant-inner-border-on-col-0okn.md
@@ -2,5 +2,5 @@
 '@astryxdesign/core': patch
 ---
 
-[fix] SideNav: remove the redundant inner border on collapsed items' child-menu popover, which drew a sharp-cornered rectangle inside the Popover's own rounded surface
+[fix] SideNav: remove the square inner border on collapsed items' child-menu popover; its 0-radius corners cut across the Popover surface's own 12px-radius arcs instead of nesting inside them, and the surface's shadow now defines the edge on its own, matching every other popover in the system
 @HelloOjasMutreja
--- a/.changeset/sidenav-remove-the-redundant-inner-border-on-col-0okn.md
+++ b/.changeset/sidenav-remove-the-redundant-inner-border-on-col-0okn.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] SideNav: remove the redundant inner border on collapsed items' child-menu popover, which drew a sharp-cornered rectangle inside the Popover's own rounded surface
+@HelloOjasMutreja
--- a/packages/core/src/SideNav/SideNavItem.tsx
+++ b/packages/core/src/SideNav/SideNavItem.tsx
@@ -162,10 +162,12 @@ const styles = stylex.create({
     cursor: 'pointer',
   },
   // Popover surface for collapsed items with children. usePopover's own
-  // render() already supplies the bordered/rounded/elevated surface
-  // (background, --radius-container, shadow) around this content — adding
-  // another border here draws a sharp-cornered rectangle inside that rounded
-  // surface instead of one coherent menu.
+  // render() already supplies the elevated surface (background,
+  // --radius-container, --shadow-low) around this content — it carries no
+  // border of its own, so the shadow alone defines the edge, same as every
+  // other usePopover consumer. This div's own square border was the only
+  // border in the stack, and its 0-radius corners cut across the outer
+  // 12px-radius arcs instead of nesting inside them.
   popoverSurface: {
     paddingBlock: spacingVars['--spacing-1'],
     paddingInline: spacingVars['--spacing-1'],
--- a/packages/core/src/SideNav/SideNavItem.tsx
+++ b/packages/core/src/SideNav/SideNavItem.tsx
@@ -35,7 +35,6 @@ import {
   typeScaleVars,
   durationVars,
   easeVars,
-  borderVars,
   radiusVars,
 } from '../theme/tokens.stylex';
 import {renderIconSlot, useIcon, type IconType} from '../Icon';
@@ -162,11 +161,12 @@ const styles = stylex.create({
     textAlign: 'start',
     cursor: 'pointer',
   },
-  // Popover surface for collapsed items with children
+  // Popover surface for collapsed items with children. usePopover's own
+  // render() already supplies the bordered/rounded/elevated surface
+  // (background, --radius-container, shadow) around this content — adding
+  // another border here draws a sharp-cornered rectangle inside that rounded
+  // surface instead of one coherent menu.
   popoverSurface: {
-    borderWidth: borderVars['--border-width'],
-    borderStyle: 'solid',
-    borderColor: colorVars['--color-border'],
     paddingBlock: spacingVars['--spacing-1'],
     paddingInline: spacingVars['--spacing-1'],
     marginInlineStart: spacingVars['--spacing-1'],


cixzhang (commented):
🟡 **Maintainer judgment recommended** — no blocker in the code; the change itself is right. Two supporting claims in the description (and the wording it feeds into the changeset) are inaccurate and worth correcting before this lands.

Triage: bug fix · non-breaking · low blast radius, but design-affecting → standard path + design pass.

## The bug is real, and the fix is correct — verified visually

You flagged that you couldn't render a before/after, so I ran it: headless Chromium against a local Storybook, `SideNav` → *Collapsible Sidebar*, collapsed, with the **Settings** child menu open. Computed styles off the real elements, not a screenshot read:

| | `main` | this PR |
| --- | --- | --- |
| Outer `usePopover` surface | radius **12px**, bg `#FFF`, `--shadow-low`, **border `0px`** | unchanged |
| Inner `popoverSurface` div | **`1px solid rgba(0,0,0,0.08)`**, radius **`0px`**, box **180×134 @ (470,308)** | **border `0px`**, box 180×**132** @ (470,308) |

That confirms the reported artifact precisely: the inner div is **full-height — 134px, exactly the outer surface's height — and flush to its top and bottom edges** while inset 4px on the left. Its square corners therefore cut straight across the outer 12px radius arcs. Removing the border is the right call, and the geometry is why.

It also *improves* concentric radius rather than merely deleting something: `r_inner ≈ r_outer − gap` wants ~8px against a 12px outer with a 4px gap, and the items sit at 10px. Before, the bordered inner was radius `0` — a textbook non-concentric nesting.

Your convention argument holds too — I checked independently, and no other `usePopover` consumer borders its content (`DropdownMenu`, `TopNavMenu`, `Selector`, and `SideNavHeading`'s own `popoverContent`, which is padding + `overflow: hidden` only).

## What needs correcting

**1. The border wasn't redundant — it was the only one.** `usePopover`'s `styles.surface` is background + `--radius-container` + `--shadow-low`; it sets **no border** (measured: `border 0px` on the outer element). So the description's "already supplies the bordered/rounded/elevated surface" isn't right, and neither is the new code comment that repeats it. After this change the popover's edge is defined by `--shadow-low` alone. That's still the correct outcome — it's what every other popover in the system does — but the rationale should say so plainly rather than implying a duplicate border was being removed.

**2. The `SideNavHeading` citation points at the wrong lines.** The `borderWidth: 0, borderStyle: 'none'` you cite are on that file's **trigger buttons** (`interactive`, `menuTrigger`) — button resets, not popover content. The evidence that actually supports you is `SideNavHeading.popoverContent`, the direct sibling of the div you're editing.

Could you reword the changeset and the code comment along those lines? The changeset text becomes the published changelog, so "redundant" would ship as the public explanation of a change that's really "the Popover surface's radius and shadow are the surface; this inner square border fought them."

## For the design owners (`needs:design-review`)

Worth a conscious nod rather than a silent merge: in light mode the popover is `#FFFFFF` over a white sidenav, so after this the only edge definition is `--shadow-low`. That's consistent with every other popover in the system — so it's a **system-wide** convention question, not a regression this PR introduces — but this PR is where it becomes visible in `SideNav`.

Everything else checks out: `SideNav` suite is **99/99** locally, full CI green, the changeset is a correct `patch` / `[fix]` / contributor line, and no `themeProps` target is lost (that div never carried one — `themeProps('side-nav-item', …)` is on the item rows).

rubyycheung (approved):

HelloOjasMutreja:
Both corrections made, thank you for the precise catch on each:

1. Reworded the code comment and the changeset: this was never a duplicate border, `usePopover`'s surface carries no border of its own (background + radius + shadow only), so it was the only border in the stack, and its square corners cut across the outer surface's radius arcs rather than nesting inside them.
2. Fixed the SideNavHeading citation to point at `popoverContent` (padding + overflow:hidden, no border) instead of the trigger-button border resets I'd originally cited.

Updated the PR description to match. And genuinely, thank you for running the headless-Chromium verification, I'd flagged that I couldn't do that locally and wasn't expecting someone to go run it independently rather than just take the source-level argument on faith.

HelloOjasMutreja:
Closing this: the underlying bug (double border on the collapsed child-menu popover) has already been fixed on main by #4880's audit-hardening sweep, which removed the border and background from `SideNavItem`'s own `popoverSurface` style the same way this PR did.

The only remaining difference is that #4880's version still restates `borderRadius: radiusVars['--radius-container']` locally, which is redundant since `usePopover`'s own surface style (`hasSurface` defaults to true here, unchanged) already applies that exact same token. But it's visually identical either way (same token, same value), so it's not worth reopening this as a separate cleanup.

PR #4743, opened 2026-08-06
fix(dropdown-menu): move indicator-icon theme target onto the Icon element

athz:
The `dropdown-menu-indicator-icon` target sat on the wrapper span, one level above the Icon that owns the glyph size — so a theme rule on it could not reach the size. Move the target onto the Icon element itself (Icon merges className onto the sized element), so a theme can restyle the chevron's size and color directly.

Divider spacing is already themeable via the existing `dropdown-menu-divider` target through the layer cascade — no CSS variable needed.

Supersedes #4667.


packages/core/src/DropdownMenu/renderDropdownItems.tsx
```diff
@@ -35,7 +35,7 @@ const styles = stylex.create({
     userSelect: 'none',
   },
   divider: {
```
cixzhang: Hmm could dropdown item divider be a theme target instead?

cixzhang (changes requested):

cixzhang (changes requested):
🔴 **Blocking** — `--_dropdown-menu-divider-margin` is declared `private: true` with no `derived` entry, so no theme can set it by any route. The PR's stated goal isn't reachable as written.

Triage: bug fix (theming reachability) · non-breaking · low blast radius · contributor → standard path.

Thanks for scoping this down from #4667 — divider-only is the right cut, and reaching for the var pipeline rather than another class target is the more principled instinct. Two things to sort before it lands: one mechanical, one a design call I'd like to make explicitly.

### 🔴 The var is settable by nobody

`DropdownMenu.doc.mjs` adds the var to `vars[]` but not to `derived[]`, and there's no `derivedVarRegistry.ts` entry. Both sanctioned routes dead-end:

```ts
// A — set it directly
components: {'dropdown-menu': {base: {'--_dropdown-menu-divider-margin': '4px'}}}
// → `astryx theme build` errors: validatePrivateVars() rejects any `--_*` prop
//   (packages/cli/api/theme/build/build.mjs:943-963, called at :1036-1044)

// B — write a standard CSS property and let the pipeline expand it
components: {'dropdown-menu': {base: {marginBlock: '4px'}}}
// → getDerivedVars('dropdown-menu','marginBlock') === [] (derivedVarRegistry.ts:52-55
//   only maps borderRadius + padding) → silent no-op
```

Per [Theming Infrastructure](https://github.com/facebook/astryx/wiki/Theming-Infrastructure) ("Adding vars to new components", step 4), a private var needs both a `derived` entry and a registry entry. #4667 actually had the doc half of this (`{property: 'marginBlock', vars: ['--_dropdown-menu-divider-margin']}`); the scope-down dropped it.

Heads-up on a related trap: the **runtime** path (`<Theme theme={…}>`) has no private-var check — `generateComponentRules` passes unknown props straight through — so route A *looks* like it works in dev and only fails at `astryx theme build`. Please verify any fix through a `theme build`, not a runtime preview.

### 🟡 Is a divider's margin something a theme should own?

I'm working through a rule I haven't written into the wiki yet, so treat this as direction rather than settled policy: **layout properties encode a component's structure, so they shouldn't become theme seams.** A seam is a paint surface — color, border, radius, shadow, type. Layout (`margin`, `padding`, `display`, flex/grid, size) is the structural contract that guarantees overflow, focus order, hit-target size and RTL, and it's exactly what a DOM refactor is entitled to change. Layout *is* themeable today, but only through the declared derived-var / container-padding pipeline, where the component still owns the structure and the theme just supplies a value it asked for.

What makes this PR the clearest case I've seen for that rule: the divider is a real `<Divider>` instance and it **already** carries a theme target attached directly to the component (`renderDropdownItems.tsx:62-67`, merged in `Divider.tsx:173-185`) — which is exactly how I want composed-component theming to work. That seam works fine for paint. It can't carry margin only because `Divider` stamps `.astryx-divider.horizontal` (0,2,0) on the same element, out-specifying `.astryx-dropdown-menu-divider` (0,1,0). So this PR is routing around a seam that exists and works — for a property I'm not sure should be themeable in the first place.

So before mechanics: is menu-divider vertical rhythm a theme value, or spacing the menu owns? I said on #4667 that a themable divider seemed okay to me, and I still lean that way for its *paint*. Margin is the part I want to be deliberate about.

### 🟡 If we keep it, make it a public var

A private + `derived` shape can't be made correct here. The registry is keyed per component, with no sub-element key, so the mapping would have to be `dropdown-menu` + `marginBlock` — meaning a theme setting the *popup's* own outer margin would silently retune every divider inside it. Property hijacking.

The wiki's own answer for "a var that doesn't map to any standard CSS property" is a public var. Suggested: rename to `--dropdown-menu-divider-margin` (no `_`), drop `private: true`, keep the `vars[]` doc entry, no `derived`/registry change. Then setting it directly in a component override just works.

### 🟡 Smaller things

- **"Matches the existing private-var pattern"** — not quite. `--_dropdown-menu-radius`/`-padding` are *declared* on the menu root (`DropdownMenu.tsx:73-74`) and re-pointed per context (`ContextMenu.tsx:105-106`, `BreadcrumbItem.tsx:218-219`). The new var is never declared anywhere, only read with an inline fallback — that asymmetry is part of how the missing `derived` slipped by.
- **`ContextMenu` / `Breadcrumbs` share this code path** (`renderDropdownItems` is used by both), and `ContextMenu.doc.mjs` already mirrors the other two vars. Whatever var survives should be mirrored there.
- **No test plan, and CI won't catch this.** I checked the drift guard: `derivedVarRegistry.test.ts` scans for vars used as *declaration keys* and skips `--_*` entirely, and its "needs a derived entry" assertion filters to `/radius|padding/` — so a `-divider-margin` var with no mapping sails through green. Please add a `theme build` round trip (theme sets the sanctioned input → assert the var lands in the generated CSS) rather than a computed-style assertion.
- **Changeset**: `patch` is right; the entry currently promises consumers a capability that doesn't work yet — update it alongside the fix.
- **Naming** is correct per the convention (`dropdown-menu` / `divider` / `margin`), and `divider` is a legitimate part since the target already exists. Nit only: the value drives `marginBlock` specifically. Separately, I'm converging on compositional names for targets on composed components (`{parent}-{position}-{component}`, e.g. `dropdown-menu-item-divider`) — not something to churn in this PR, but it's why I'm looking at the naming across your batch and #4756 together.
- Please restate #4667's specificity rationale in this description — it's the whole justification and it isn't here. Worth noting it's conditional: no shipped theme styles the global `divider` slot, so the two-class collision only bites themes that do.

Happy to take the public-var version quickly once we agree the margin should be theme-owned at all — that's the question I want to answer first.

packages/core/src/DropdownMenu/renderDropdownItems.tsx
athz: You're right — the `dropdown-menu-divider` target already exists on the Divider element, and the `@layer astryx-theme` cascade wins over the base StyleX margin. Dropped the CSS variable approach entirely. This PR now only moves the indicator-icon target onto the Icon element (so themes can reach the glyph size directly).

cixzhang (changes requested):
Thanks for turning this around so quickly — dropping the var and leaning on the `dropdown-menu-divider` target that already exists was the right read, and putting the indicator target on the Icon is the part I wanted.

One thing before it lands: the spinner branch loses the target altogether. On `main` both branches carry `dropdown-menu-indicator-icon`; here only the chevron does, so a theme that styles the submenu indicator watches it disappear the moment that submenu starts loading. Leaving that branch as `main` has it is enough:

```tsx
const endAffordance = hasSpinner ? (
  <span
    {...mergeProps(
      themeProps('dropdown-menu-indicator-icon'),
      stylex.props(triggerStyles.caret),
    )}>
    <Spinner size="sm" />
  </span>
) : (
  <span {...stylex.props(triggerStyles.caret)}>
    <Icon
      icon="chevronRight"
      size="sm"
      color="secondary"
      {...themeProps('dropdown-menu-indicator-icon')}
    />
  </span>
);
```

Was there a reason the loading state needed to give it up? If you'd rather talk it through with someone, we're in [Discord](https://discord.com/invite/XnsUcFykEP).

[Reviewed by Robohands]

packages/core/src/DropdownMenu/DropdownMenuSubMenu.tsx
```diff
@@ -474,20 +474,17 @@ export function DropdownMenuSubMenu(
   );
 
   const endAffordance = hasSpinner ? (
-    <span
-      {...mergeProps(
-        themeProps('dropdown-menu-indicator-icon'),
-        stylex.props(triggerStyles.caret),
-      )}>
+    <span {...stylex.props(triggerStyles.caret)}>
```
cixzhang, line 479: Right to drop it here — Spinner has its own target; this span only lays out.

cixzhang (approved):
Thanks for the patience on this, and sorry — my last round was wrong. The target belongs on the Icon, as Selector and MultiSelector already do. Merging.

[Reviewed by Robohands]

cixzhang:
Heads up on two commits I pushed to your branch, so nothing here is a surprise:

- **Merged `main` in.** The branch was 17 days behind and its preview had expired — that's our latency, not yours. Clean merge, no conflicts, your diff is unchanged.
- **One line on the changeset**, noting that the loading branch leaves the target to `Spinner`'s own `astryx-spinner`. That was the one thing your description didn't say, and it wasn't worth costing you a push.

Nothing else touched. Merging once CI comes back green.

[Reviewed by Robohands]

PR #4746, opened 2026-08-06
fix(ComplexSelector): status colors, spoken value, single focus ring

AKnassa:
## What this does

A quality pass on the ComplexSelector component, following the [Component Hardening Protocol](https://github.com/facebook/astryx/wiki/Component-Hardening-Protocol) (Layers 1 and 2). It makes ComplexSelector behave like its sibling selectors everywhere they already agree.

Refs #4710

## Why

ComplexSelector shipped recently and was missing polish its siblings already have. The three biggest gaps: setting an error/warning/success status changed nothing visually, screen readers never announced the selected value, and keyboard focus drew two rings on top of each other.

## What changed

- **Status is visible now.** An error selector gets the same red border and status icon as every other input. Warning and success too, and the tooltip style gets its little info button.
- **Screen readers hear the selection.** The trigger now announces "label, current value" instead of just the label.
- **One focus ring, not two** — and the ring stays visible even when a status color is set (that second part was a gap the sibling components still have; this PR adds the shared style that fixes it, wired up for ComplexSelector first).
- **Small consistency fixes:** the popup opens on both Arrow Up and Arrow Down; the component respects a container's size setting; it has its own translation entry for "Select…" instead of borrowing Selector's; a `ref` can be attached like on other components.
- **Docs and examples:** eight missing props documented, Chinese docs added, and eight new Storybook stories covering sizes, statuses, disabled, loading, empty, and long-text states.
- **Tests: 3 → 27**, written failing-first; joins the input-family width contract test.

## How to see it

Storybook → Core/ComplexSelector → the new "With status", "Status variant comparison", and "Size variants" stories. Tab to any trigger to see the single focus ring (also with a status set).

## Notes for reviewers

- Trigger accessible names now include the value, so downstream tests using `getByRole('button', {name: 'Label'})` need `{name: 'Label Value'}` or a regex (called out in the changeset).
- Full suite: 9303/9305 passing — the two failures are pre-existing macOS case-sensitivity artifacts in packages/cli, present on untouched main.
- Two adjacent finds are deliberately **not** in this PR: Selector/MultiSelector/TextInput share the status-hides-focus-ring gap (the new shared style is ready for them), and usePopover has an uncanceled autofocus animation frame that can drop focus into a closed popup. Happy to file both as issues.


packages/core/src/ComplexSelector/ComplexSelector.tsx
```diff
@@ -83,6 +92,10 @@ const styles = stylex.create({
     lineHeight: 'inherit',
     color: 'inherit',
     cursor: 'pointer',
```
rubyycheung: The disabled trigger shows a `pointer` cursor on hover instead of `not-allowed`. The wrapper picks up `cursor: not-allowed` when disabled, but this inner `<button>` hardcodes `cursor: 'pointer'` and sits on top, so its cursor wins — and an explicit `pointer` also overrides the browser's native disabled cursor. Since this PR hardens the disabled + status states, gating the cursor here fits the scope.

```suggestion
    cursor: {
      default: 'pointer',
      ':disabled': 'not-allowed',
    },
```

rubyycheung (changes requested):

packages/core/src/ComplexSelector/ComplexSelector.tsx
```diff
@@ -343,10 +408,14 @@ export function ComplexSelector<Value>({
             inputWrapperStyles.base,
             styles.triggerContainer,
             styles[size],
-            styles.focusRing,
             isDisabled && inputWrapperStyles.disabled,
             isDisabled && styles.disabled,
             triggerLabel == null && styles.placeholder,
+            status && inputStatusBorderStyles[status.type],
```
rubyycheung, line 436: With an attached status, the trigger's rounded bottom corners let the pink status background peek through as two triangles. Sibling inputs (e.g. `TextInput`) flatten their bottom corners (`borderEndStartRadius`/`borderEndEndRadius: 0`) when status is attached so they meet the status box flush. Same fix needed here.

packages/core/src/ComplexSelector/ComplexSelector.tsx
AKnassa: Fixed in 47567ad. The wrapper's `not-allowed` cannot reach the trigger: on the wrapper the two cursors merge inside one `stylex.props` call so the loser is dropped, but the trigger is a separate element, so its own `pointer` wins the cascade and also overrides the native disabled cursor. Gating on `:disabled` covers a `fieldset[disabled]` ancestor too.

packages/core/src/ComplexSelector/ComplexSelector.tsx
AKnassa: Confirmed and fixed in 47567ad, with one correction: no sibling flattens its corners today. `borderEndStartRadius`/`borderEndEndRadius` has no hits in any input under `packages/core/src`, and TextInput's ErrorStatus story renders the identical wedge. The fix went into the shared layer as `inputAttachedStatusStyles`, wired for ComplexSelector, matching how this PR added `inputStatusFocusShadowStyles`. It applies only where Field renders the attached box (attached variant with a message), so a status without a message, detached, and tooltip keep their rounded corners. Happy to file the sibling sweep as a follow-up.

cixzhang:
Closing this one, and the reason is a process change rather than anything wrong with the work.

We've replaced the Component Hardening Protocol pass with the [component audit process](https://github.com/facebook/astryx/wiki/Component-Audit-Rubric): a component is graded against ~200 numbered checks across 11 weighted sections, the score and its open blockers go in the scores ledger, and the fixes land as a focused PR per audit with a before/after score. A broad hardening PR doesn't fit that shape — it bundles findings from several sections, so there's nothing to score it against and no way to tell which check each change answers.

Most of what's here is also on `main` now, arrived at through that route:

- **Single focus ring** — done in ce86871aca (part of [#4935](https://github.com/facebook/astryx/pull/4935)). The trigger wrapper uses the shared `focusOutlineStyles.focusWithin` (`:has(:focus-visible)`), so the hand-rolled ring and the doubled ring are both gone.
- **Status is visible** — `main` threads `status` / `statusVariant` through to `Field`, sets `aria-invalid` on the trigger, and wires the status message through `aria-describedby`.

**One finding here is real and still open on `main`:** the trigger's accessible name. It carries `aria-labelledby={labelId}`, which points at the field label and overrides the button's own content — so a screen reader announces the label and never the current value. That's an A-section finding, and I'd rather it be recorded as one against a scored audit than lost in a closed PR, so it will go on the ledger when ComplexSelector is audited.

Thanks for the work here, and sorry it caught a process change mid-flight — this isn't a judgment on the diff. If you'd like to pick up ComplexSelector's audit itself, say the word and it's yours.

PR #4750, opened 2026-08-06
fix(avatar): preserve grapheme clusters when generating initials

### LINKED ISSUE #4749: Avatar: getInitials splits emoji/grapheme clusters when generating initials
`getInitials()` in `Avatar.tsx` extracts initials using `.charAt(0)`, which indexes UTF-16 code units rather than user-perceived grapheme clusters.

Consequently, names beginning with multi-code-unit characters are handled incorrectly:

- Non-BMP characters such as emoji are split into an unpaired surrogate.
- Flag and ZWJ emoji sequences are truncated.
- Combining marks can be dropped from decomposed letters.

These regression tests currently fail against main:

```tsx
it('does not split an emoji surrogate pair when generating initials', () => {
  render(<Avatar name="😀 Ada" data-testid="avatar" />);
  expect(screen.getByTestId('avatar')).toHaveTextContent('😀A');
});

it('preserves a complete grapheme when generating initials', () => {
  render(<Avatar name="🇬🇧 Ada" data-testid="avatar" />);
  expect(screen.getByTestId('avatar')).toHaveTextContent('🇬🇧A');
});
```

Would maintainers consider this a bug worth fixing upstream? If so, I'd be happy to open a PR with the `Intl.Segmenter` implementation, a safe code-point fallback, and these regression tests.

alex-js-ltd:
Fixes #4749

`getInitials()` used `.charAt(0)`, which indexes UTF-16 code units rather than user-perceived grapheme clusters. As a result, names beginning with multi-code-unit characters—such as emoji, flag or ZWJ sequences, and decomposed letters with combining marks—could produce broken or truncated initials.

## Changes

* Updated `getInitials()` to extract the first grapheme using `Intl.Segmenter` with `granularity: 'grapheme'`.
* Added a code-point-safe fallback for environments where `Intl.Segmenter` is unavailable. The fallback prevents surrogate pairs from being split, although full multi-code-point grapheme preservation requires `Intl.Segmenter`.
* Added regression tests covering an emoji surrogate pair and a multi-code-point flag emoji.
* Added a changeset.

## Test plan

* [x] `pnpm -F @astryxdesign/core test` — Avatar suite passes (45/45), including the two new tests. Both fail against the previous `.charAt(0)` implementation and pass with this fix.
* [x] `pnpm check:changesets`

### DIFF
--- a/.changeset/avatar-initials-grapheme-clusters.md
+++ b/.changeset/avatar-initials-grapheme-clusters.md
@@ -2,6 +2,6 @@
 '@astryxdesign/core': patch
 ---
 
-[fix] Avatar: fallback initials no longer split multi-codepoint characters (emoji, flag sequences, base+combining-mark letters) when deriving initials from a name — `getInitials` now segments by grapheme cluster instead of UTF-16 code unit.
+[fix] Avatar: fallback initials no longer break for names containing emoji or other multi-codepoint characters.
 
 @alex-js-ltd
--- a/.changeset/avatar-initials-grapheme-clusters.md
+++ b/.changeset/avatar-initials-grapheme-clusters.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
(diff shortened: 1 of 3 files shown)


cixzhang (approved):
Thank you for improving the i18n here!

PR #4752, opened 2026-08-06
fix(pagination): center the first/last double-chevron caret icons under RTL

freddymeta:
## Summary

Follow-up to #4723. That PR fixed the vertical centering of the **prev/next** caret icons by moving the RTL mirror off a `display: contents` wrapper span and onto the `Icon` directly via `xstyle`. The `input` variant's **first/last** («/») double-chevron buttons were still using the old wrapper span, so they kept the same few-pixels-high offset — the "double caret looks off" bug.

This applies the identical, already-reviewed fix to both double-chevron carets, so all four carets (first / prev / next / last) now share one consistent approach.

## The fix

```tsx
// before — mirror on a display:contents wrapper (drops the icon out of the
// button's flex-centering context, so the glyph sits a few px high)
icon={
  <span {...stylex.props(rtlStyles.mirror)}>
    <Icon icon="chevronsLeft" size={isSm ? 'sm' : 'md'} />
  </span>
}

// after — mirror rides on the Icon itself (stays a centered flex child,
// still flips under RTL, no wrapper element)
icon={
  <Icon icon="chevronsLeft" size={isSm ? 'sm' : 'md'} xstyle={rtlStyles.mirror} />
}
```

## Why it slipped through

When #4723 landed, the first/last buttons didn't exist on `main` yet — they came in with the `input` variant (#4505). #4723 could only fix the carets that existed at the time; the double-chevrons inherited the wrapper pattern and shipped with the offset.

## Tests

Extended the #4723 regression test to cover **all four** carets in one pass (it now renders the `input` variant so first/last are present) and to assert the mirror class rides on the **icon element itself** rather than a wrapper. I verified the test genuinely fails when a caret is reverted to the wrapper form. `93/93` Pagination tests, typecheck, and strict lint all green.

RTL flip is preserved (`scaleX(-1)` still applies via the shared `rtlStyles.mirror`); centering is verified visually in Storybook.

### DIFF
--- a/.changeset/pagination-double-caret-centering.md
+++ b/.changeset/pagination-double-caret-centering.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Pagination: vertically center the first/last double-chevron caret icons in the `input` variant. Like the prev/next carets fixed in #4723, the first/last («/») buttons still wrapped their chevron in a `display: contents` mirror span, which dropped the icon out of the button's flex-centering context so the glyph sat a few pixels high. The mirror transform now rides on the `Icon` directly via `xstyle`, matching the prev/next carets — so the icon stays a centered flex child and still flips under RTL, with no wrapper element.
+@freddymeta
--- a/packages/core/src/Pagination/Pagination.test.tsx
+++ b/packages/core/src/Pagination/Pagination.test.tsx
@@ -142,19 +142,39 @@ describe('Pagination', () => {
       ).toBe(true);
     });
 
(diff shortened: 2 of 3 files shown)


cixzhang (approved):

rubyycheung (approved):

freddymeta:
Closing without merging — superseded by #4775.

#4775 ("migrate style-only wrappers onto component xstyle") swept the same change across all four carets: on `main`, first/prev/next/last all render `<Icon xstyle={rtlStyles.mirror} />` directly, with no `display:contents` wrapper span. This PR's caret fix is now a no-op against `main`, so there's nothing left to land here.

PR #4753, opened 2026-08-06
feat(menu): add destructive variant to dropdown/context menu items

freddymeta:
## What

Adds a `variant` prop to menu items so dangerous actions (e.g. Delete) can render in the error color.

```tsx
<DropdownMenu button={{label: 'Actions'}}>
  <DropdownMenuItem label="Edit" onClick={onEdit} />
  <DropdownMenuItem label="Delete" variant="destructive" onClick={onDelete} />
</DropdownMenu>

// data-driven, and ContextMenu shares the same item shape:
items={[{label: 'Delete', variant: 'destructive', onClick: onDelete}]}
```

## Why

There was no way to visually distinguish a destructive action in a dropdown or context menu — every item rendered in the default text color. `Button` already has `variant="destructive"`; this brings the same affordance (and the same word) to menu items.

## API

`variant?: 'default' | 'destructive'` on `DropdownMenuItem`, plus the same optional field on the data-driven `DropdownMenuItemData`. Because `ContextMenu` reuses that item data shape and renderer, context-menu items get it for free.

- **Non-breaking**: optional, defaults to `'default'`. Default items emit no new attributes/classes, so existing menus are byte-for-byte unchanged.
- **Semantic + theme-aware**: `'destructive'` uses the existing `--color-error` / `--color-error-muted` tokens (label, description, icon, and a tinted focus/hover background) — the same tokens `Button`'s destructive variant uses. No hardcoded reds.

## How it recolors the label

`Item` hardcoded its label/description colors, so a parent couldn't recolor them. This makes those two colors read from optional custom properties with the current tokens as the fallback (`var(--_item-label-color, <token>)`), so the styling is inert unless a consumer opts in — which the destructive menu-item variant does. This mirrors the `--_private-var, fallback` pattern already used elsewhere in the package.

## Changes

- `DropdownMenuItem`: new `variant` prop, destructive StyleX style, error-colored icon, `data-variant` reflection (only when destructive).
- `DropdownMenuItemData` + `renderDropdownItems`: forward `variant` (covers both compound and data-driven usage, and ContextMenu).
- `Item`: label/description colors become overridable via custom properties (inert by default).
- Docs (`.doc.mjs`, EN + ZH), Storybook stories (DropdownMenu + ContextMenu), and tests for compound, data-driven, section-nested, and ContextMenu paths.

## Verification

- DropdownMenu / ContextMenu / Item suites: 169 tests pass ✓
- `tsc --noEmit` ✓, ESLint clean, `pnpm check:changesets` ✓

### DIFF
--- a/packages/core/src/DropdownMenu/DropdownMenu.tsx
+++ b/packages/core/src/DropdownMenu/DropdownMenu.tsx
@@ -105,6 +105,11 @@ export interface DropdownMenuItemData {
   isDisabled?: boolean;
   icon?: ReactNode | IconType;
   /**
+   * Visual variant. `'destructive'` renders the item in the error color for
+   * dangerous actions (e.g. Delete). Ignored for submenu rows. @default 'default'
+   */
+  variant?: 'default' | 'destructive';
+  /**
    * Nested submenu entries. When present, this row becomes a submenu (a
    * flyout revealing `items`) instead of a leaf action — no separate item
    * "type" is needed. Data-mode parity for the compound DropdownMenuSubMenu API.
--- a/packages/core/src/DropdownMenu/DropdownMenu.tsx
+++ b/packages/core/src/DropdownMenu/DropdownMenu.tsx
@@ -105,8 +105,8 @@ export interface DropdownMenuItemData {
   isDisabled?: boolean;
   icon?: ReactNode | IconType;
   /**
-   * Visual variant. `'destructive'` renders the item in the error color for
-   * dangerous actions (e.g. Delete). Ignored for submenu rows. @default 'default'
+   * Visual variant for a leaf item. `'destructive'` renders it in the error
+   * color for dangerous actions (e.g. Delete). @default 'default'
    */
   variant?: 'default' | 'destructive';
   /**
--- a/.changeset/menu-item-destructive-variant.md
+++ b/.changeset/menu-item-destructive-variant.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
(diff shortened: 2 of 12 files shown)


packages/core/src/DropdownMenu/DropdownMenu.tsx
```diff
@@ -104,6 +104,11 @@ export interface DropdownMenuItemData {
   onClick?: () => void;
   isDisabled?: boolean;
   icon?: ReactNode | IconType;
+  /**
+   * Visual variant. `'destructive'` renders the item in the error color for
+   * dangerous actions (e.g. Delete). Ignored for submenu rows. @default 'default'
+   */
+  variant?: 'default' | 'destructive';
```
cixzhang, line 111: Do we need the destructive at the menu level or just the item level?

cixzhang: Actually I guess this doesn't do anything at the menu level. Probably cleanup?

cixzhang (changes requested):

rubyycheung (approved):

packages/core/src/DropdownMenu/DropdownMenu.tsx
freddymeta: Good catch. It is item-level — `variant` lives on `DropdownMenuItem` and is consumed for leaf items. On `DropdownMenuItemData` it is forwarded by `renderDropdownItems` to the leaf `<DropdownMenuItem>` (and, since ContextMenu shares this data shape, that is how context-menu items get it too), so it is doing something for data-driven menus, not just the compound API.

The confusing part was the JSDoc: the same interface is also used for submenu rows, where `variant` is a no-op. I have dropped that caveat and reworded it as a leaf-item field so the type is not advertising something it ignores. No menu-component-level `variant` exists. (8a92e25)

cixzhang: Oh I see this data is actually from the item. Maybe we should source this type from the item itself instead of needing it separately here.

cixzhang (approved):

PR #4756, opened 2026-08-06
feat(selector): themeable empty-state target on Selector and MultiSelector

freddymeta:
## What

Exposes the Selector / MultiSelector **empty state** ("No results found") as a stable theming target:
- **`astryx-selector-empty-state`**
- **`astryx-multi-selector-empty-state`**

```css
.astryx-selector-empty-state {
  padding: 24px;
  font-size: 13px;
  color: var(--my-muted-text);
}
```

## Why

Consumers currently have to style the empty state through **fragile structural CSS selectors** — e.g. matching "the only role-less child of the listbox" — because no stable class exists on it. That kind of selector is brittle (it can match unrelated role-less children in a consumer's own lists) and breaks silently when the internal DOM changes. A stable paint target replaces it with a supported seam.

The empty-state element paints (color, size, padding) and is a leaf, so it's a legitimate paint target — not a layout-only container.

## Scope notes

- **Border:** the trigger's border is already reachable via the existing `astryx-selector` root target, so no new target is needed for it. The dropdown *surface* border/background/shadow come from shared popover chrome (`usePopover`), not a Selector-owned element, so it's intentionally out of scope here.
- **Search placeholder:** the search field is a `TextInput`, so its placeholder is already reachable via `.astryx-text-input` (the `<input>` is a child of that class). No redundant Selector-scoped target is added. A dedicated Selector-scoped placeholder seam would require a `TextInput` change and is left as a possible follow-up.

- **Non-breaking:** purely additive — a stable class on an existing element. No API or visual change.

## Changes

- `Selector.tsx` / `MultiSelector.tsx`: stamp `themeProps('selector-empty-state')` / `themeProps('multi-selector-empty-state')` on the empty-state element.
- `*.doc.mjs`: document the new classes in each `theming.targets`.
- Tests for both components asserting the target renders on the empty state.

## Verification

- Selector + MultiSelector + theme suites: 778 tests pass ✓ (incl. `themingTargets.test.ts` — every rendered class is documented)
- `tsc --noEmit` ✓, ESLint clean, `check-sync` ✓, `check:changesets` ✓

### DIFF
--- a/packages/core/src/MultiSelector/MultiSelector.doc.mjs
+++ b/packages/core/src/MultiSelector/MultiSelector.doc.mjs
@@ -25,6 +25,7 @@ export const docs = {
         visualProps: ['variant', 'size', 'status'],
       },
       {className: 'astryx-multi-selector-clear-icon'},
+      {className: 'astryx-multi-selector-empty-state'},
       {
         className: 'astryx-multi-selector-indicator-icon',
         states: ['state'],
--- a/.changeset/selector-empty-state-target.md
+++ b/.changeset/selector-empty-state-target.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] Selector and MultiSelector: expose the empty ("No results found") state as a themeable target (`astryx-selector-empty-state` / `astryx-multi-selector-empty-state`). Lets themes restyle the empty state without structural CSS selectors — consumers previously had to reach it via fragile structural selectors because no stable target existed. (The search field is a TextInput, so its placeholder is reachable today via `.astryx-text-input::placeholder`; a Selector-scoped placeholder seam would require a TextInput change and is left as a possible follow-up.)
+
+@freddymeta
--- a/packages/core/src/MultiSelector/MultiSelector.test.tsx
+++ b/packages/core/src/MultiSelector/MultiSelector.test.tsx
@@ -1509,6 +1509,29 @@ describe('MultiSelector statusVariant forwarding', () => {
   });
 });
 
+describe('MultiSelector empty-state theme target', () => {
+  const OPTIONS = ['Apple', 'Banana', 'Cherry'];
+
+  it('renders the astryx-multi-selector-empty-state target on the "No results found" element', async () => {
+    const user = userEvent.setup();
+    const {container} = render(
+      <MultiSelector
+        label="Fruit"
+        options={OPTIONS}
+        value={[]}
+        onChange={() => {}}
+        hasSearch
+      />,
+    );
+    await user.click(screen.getByRole('button', {name: 'Fruit'}));
+    await user.type(screen.getByRole('combobox', h), 'xyz');
+
+    const empty = container.querySelector('.astryx-multi-selector-empty-state');
+    expect(empty).not.toBeNull();
+    expect(empty).toHaveTextContent('No results found');
+  });
+});
+
 describe('MultiSelector clear icon theme target', () => {
   const ICON_OPTIONS = ['Apple', 'Banana', 'Orange'];
 
--- a/packages/core/src/MultiSelector/MultiSelector.tsx
+++ b/packages/core/src/MultiSelector/MultiSelector.tsx
@@ -1343,7 +1343,10 @@ export function MultiSelector<T extends MultiSelectorOptionType>({
         <div
           key="empty"
           role="presentation"
-          {...stylex.props(styles.emptyState)}>
+          {...mergeProps(
+            themeProps('multi-selector-empty-state'),
+            stylex.props(styles.emptyState),
+          )}>
           No results found
         </div>,
       );
--- a/packages/core/src/Selector/Selector.doc.mjs
+++ b/packages/core/src/Selector/Selector.doc.mjs
@@ -26,6 +26,7 @@ export const docs = {
         visualProps: ['variant', 'size', 'status'],
       },
       {className: 'astryx-selector-option'},
+      {className: 'astryx-selector-empty-state'},
       {className: 'astryx-selector-clear-icon'},
       {className: 'astryx-selector-indicator-icon', states: ['state']},
     ],
--- a/packages/core/src/Selector/Selector.test.tsx
+++ b/packages/core/src/Selector/Selector.test.tsx
@@ -1318,6 +1318,28 @@ describe('Selector statusVariant forwarding', () => {
   });
 });
 
+describe('Selector empty-state theme target', () => {
+  const OPTIONS = ['Apple', 'Banana', 'Cherry'];
+
+  it('renders the astryx-selector-empty-state target on the "No results found" element', async () => {
+    const user = userEvent.setup();
+    const {container} = render(
+      <Selector
+        label="Fruit"
+        options={OPTIONS}
+        onChange={() => {}}
+        hasSearch
+      />,
+    );
+    await user.click(screen.getByRole('button', {name: 'Fruit'}));
+    await user.type(screen.getByRole('combobox', h), 'xyz');
+
+    const empty = container.querySelector('.astryx-selector-empty-state');
+    expect(empty).not.toBeNull();
+    expect(empty).toHaveTextContent('No results found');
+  });
+});
+
 describe('Selector clear icon theme target', () => {
   // Resolve the clear glyph span (the astryx-icon element inside the clear
   // button), independent of the theme target class.
--- a/packages/core/src/Selector/Selector.tsx
+++ b/packages/core/src/Selector/Selector.tsx
@@ -1078,7 +1078,10 @@ export function Selector<T extends SelectorOptionType>(
         <div
           key="empty"
           role="presentation"
-          {...stylex.props(styles.emptyState)}>
+          {...mergeProps(
+            themeProps('selector-empty-state'),
+            stylex.props(styles.emptyState),
+          )}>
           No results found
         </div>,
       ];


imdreamrunner (changes requested):
This PR makes an existing structural difference part of the public theming API:

```html
<!-- Selector: search stays fixed -->
<div>search</div>
<div class="astryx-selector-dropdown" role="listbox">...</div>

<!-- MultiSelector: search scrolls with options -->
<div class="astryx-multi-selector-dropdown">
  search
  <div role="listbox">...</div>
</div>
```

Since `styles.dropdown` owns `maxHeight` and `overflowY`, these targets behave differently despite sharing the `dropdown` name.

Could we align them in this PR by exposing `*-dropdown` on the wrapper and `*-listbox` on the scrollable options?

If that behavioral change is out of scope, I would name the current targets by their actual scope: `astryx-selector-listbox` and `astryx-multi-selector-dropdown`.

packages/core/src/MultiSelector/MultiSelector.doc.mjs
```diff
@@ -25,6 +25,7 @@ export const docs = {
         visualProps: ['variant', 'size', 'status'],
       },
       {className: 'astryx-multi-selector-clear-icon'},
+      {className: 'astryx-multi-selector-dropdown'},
```
cixzhang: This probably needs a better name. like `astryx-multi-selector-menu-popover`? But this is starting to feel a bit like a reach for theming since we can have multiple containers here that compose.

cixzhang (changes requested):
🔴 **Blocking** — the entire stated use case here is layout (`max-height`, `padding`), which points at the derived-var pipeline rather than a class target; and the same target name lands on two structurally different elements. Plus `needs:code-review` on new public theming API.

**Triage:** new API surface (theming targets are a public commitment) · non-breaking · blast radius = two core components + the system-wide target vocabulary → deep path. ⚠️ **Needs maintainer judgment on the API surface** — per `.github/instructions/packages.instructions.md`, whether these targets *should exist and take this shape* is a human call.

Thanks for picking this up — the mechanics are clean (`mergeProps` order, no hand-authored `data-*`, no state sub-targets, docs guard satisfied, well-formed changeset). Two design concerns below are directions we're still working through rather than written wiki rules — flagging them as such — but together they say this wants a different mechanism.

### 🔴 1. This is a layout seam, and layout shouldn't be a class target

A theming target is best understood as a **paint** seam: color, background, border, radius, shadow, font. **Layout** — `display`, `position`, flex/grid, `margin`, `padding`, `gap`, `width`/`max-height`, `transform` — is the component's structural contract. It's what guarantees overflow behavior, focus order, hit-target size and RTL, and it's exactly what we need to stay free to change in a DOM refactor. Exposing it as a raw class target on an internal element freezes the structure and makes those guarantees theme-dependent.

Classifying this PR:

| Target | What the description themes | |
| --- | --- | --- |
| `astryx-selector-dropdown` | `max-height: 240px`, `padding: 4px` | layout |
| `astryx-multi-selector-dropdown` | `max-height: 240px`, `padding: 4px` | layout |

And the styles behind the seam (`Selector.tsx:237-249`, `MultiSelector.tsx:260-265`) are `boxSizing` / `maxHeight` / `overflowY` / `padding` — no paint properties at all. Concretely: a theme setting `max-height` here is changing whether and where the option list scrolls, and in MultiSelector the same box decides whether the search field scrolls out of view with the options.

Layout *is* themeable today — deliberately — but through the declared derived-var pipeline, where the component still owns the structure and the theme only sets a value the component declared. `DropdownMenu` is the model, one component over, for exactly these two properties:

```js
// DropdownMenu.doc.mjs
vars: [
  {name: '--_dropdown-menu-radius',  private: true, ...},
  {name: '--_dropdown-menu-padding', private: true, ...},
],
derived: [
  {property: 'padding',      vars: ['--_dropdown-menu-padding']},
],
```

so a theme writes `dropdown-menu: {base: {padding: '8px'}}` and `DropdownMenu.tsx:73-76` decides where it lands. Selector and MultiSelector declare no `theming.vars` / `derived` at all today — that's the actual gap. If the need is dropdown padding / max-height, I'd rather we fill it that way (declared var + `derived` entry + `derivedVarRegistry.ts`) than with a class target on an internal div.

### 🔴 2. If a class target does survive: name it by position, and attach it to the component

Also emerging direction: when the thing being themed is composed inside a parent, the target should attach to **that element/component instance** (never a wrapper minted to hold it) and be named compositionally — `{parent}-{position}-{thing}`, where position says where it sits (`trigger`, `option`, `row`, `header`, `menu`) and the last segment says what it is (`icon`, `checkbox`, `popover`). We already name that way: `tree-list-item-label`, `table-header-cell`, `breadcrumb-item-menu-trigger`.

`-dropdown` has no position segment, and that vagueness is doing real damage here:

- `Selector.tsx:1310-1315` / `1323-1331` → the target lands on the scrolling `role="listbox"`.
- `MultiSelector.tsx:1587-1591` → it lands on the wrapper that also contains `renderSearch()`; the inner listbox (`:1592-1598`) carries no styles.

Same name, two shapes, so one theme rule gives two results — `padding: 4px` insets only the options in Selector but also the search box in MultiSelector. That's [Theming Infrastructure → principle #3](https://github.com/facebook/astryx/wiki/Theming-Infrastructure) ("one name and shape per concept, across components"), and `typeahead-dropdown` (`BaseTypeahead.tsx:808`, identical `styles.dropdown`) already fixes `{component}-dropdown` = the scrolling option list, so Selector matches the sibling and MultiSelector diverges.

Names that state position would separate the two concepts cleanly — e.g. `…-menu-popover` for the surface vs `…-option-list` for the scroll box. Building on @imdreamrunner's point, though, I'd settle the design question first (principle 7, consolidate at the design level): *should* MultiSelector's search scroll away with the options when Selector's stays pinned? Fix that and the naming falls out.

### 🔴 3. The MultiSelector placement also trips the written rule

The head rule for `themeProps` says targets belong on "visible elements that have base styles," explicitly **not** on "structural wrappers (display: contents, layout containers)." `MultiSelector.tsx:1587-1591` is a layout container that paints nothing — this one holds independently of #1 and #2.

### 🔴 4. No stated use case

The Why is "the container had no stable hook." Principle 6 rules that out explicitly: *"Structural selectors are brittle, so here's a seam" is not sufficient justification* — a target needs a stated visual intent from a real consumer need, because speculative targets are very hard to remove once shipped. Which theme wants this, and what should it look like?

### 🟡 5. The element that actually paints has no target — and it's a composed component

Three nested boxes here: the popover surface from `usePopover.tsx:411-419` (`styles.surface` — background, radius, `--shadow-low`), the search wrapper, and the scroll box. The surface is the one that paints, and it carries no theme class at all (`usePopover`'s `render` never calls `themeProps`, though `Popover` itself documents `astryx-popover`). So `selector-dropdown: {base: {borderRadius, boxShadow}}` would put paint on the inner scroll box and leave the visible edge unchanged. This is exactly the case #2 is about: if there's a real need, it's likely a compositional target on the surface instance (`selector-menu-popover`) for paint, plus declared vars for the layout — not a class on this div.

### 🟡 6. Tests don't exercise the claim

`Selector.test.tsx:1321-1345` and `MultiSelector.test.tsx:1512-1531` assert only that a class string exists in the closed/hidden DOM. Neither opens the dropdown, and neither runs `defineTheme` + `generateThemeTestCSS` — which is what proves a theme can reach it (see the sibling target tests in #4627/#4626/#4628). The MultiSelector assertion uses `.closest(...)`, which passes if the class is on *any* ancestor, so it can't distinguish wrapper from listbox — exactly the distinction under discussion.

### 🟡 7. Changeset / description inaccuracies

- `.changeset/selector-dropdown-theme-target.md:5` calls both "the popover's option-list container" — not true of the MultiSelector element, which also wraps the search box.
- The Why says both components already document an options target; MultiSelector has none on `main` (`MultiSelector.doc.mjs:22-32`) — that's the still-open #4628.

### 🟡 8. Portaled target with no `visualProps`

The dropdown renders into the popover layer, outside the `.astryx-selector` subtree, so the root's `variant`/`size`/`status` never reach it, and neither doc entry (`Selector.doc.mjs:29`, `MultiSelector.doc.mjs:28`) declares any — a theme can't express "dropdown for a small selector" (principle 2). `typeahead-dropdown` has the same gap; let's decide once and converge.

### 🟢 Nits

- Doc entry placement differs between the files (`Selector.doc.mjs:29` after `-option`; `MultiSelector.doc.mjs:28` after `-clear-icon`) — pick one ordering.
- Branch is behind `main` (#4627 landed on the same `Selector.doc.mjs` array afterwards) — rebase before landing.

### Cross-PR note

This is the fourth concurrent PR adding theming targets to Selector/MultiSelector (#4626 indicator wrappers + disabled state, #4627 `selector-check` — merged, #4628 rows/select-all), and both concerns above run through the cluster:

- **Layout-as-target:** #4626 themes `transform` on a wrapper; #4628 themes `transform: scale()` on a checkbox and `display: none` on a divider (principle 5 also bars a target whose purpose is hiding an internal element); this PR themes `max-height`/`padding`. #4628's `-option` / `-option-label` are properly paint.
- **Attach-to-the-component:** #4627 got this right (the class goes on the `Icon` instance), #4628's `-option-checkbox` goes on a wrapper by its own test's description.

Structurally, #4628 + this PR would leave MultiSelector with rows and a wrapper but no name for the listbox, while Selector's `-dropdown` *is* the listbox. Given principle 7, I think these want to be one consolidated pass over Selector / MultiSelector / Typeahead / DropdownMenu that settles the vocabulary, the paint-vs-layout line, and the attachment rule together — driven by named consumer needs, rather than four PRs each freezing a different slice of internal structure into the public API. Happy to help scope that if you're up for it.

cixzhang (approved):

cixzhang:
Thanks, accepted since it looks like feedback was resolved.

PR #4757, opened 2026-08-06
docs(blog): i18n + RTL engineering post (draft)

nynexman4464:
## What

A **draft** engineering blog post on how internationalization and RTL came to Astryx. Landed as `draft: true` so it stays out of production output while we iterate on the writing — this PR is mostly to get it into git history and preview.

Adds:
- `apps/docsite/src/content/blog/posts/internationalizing-astryx.md`
- a `nynexman4464` author entry in the blog author registry

## The post covers

- **Origin** — the effort started from a community bug report: a developer shipping an Arabic + Kurdish (RTL) app hit hardcoded English strings with no override.
- **Build vs. buy** — why Astryx depends on the ICU *format* (`intl-messageformat` as default, injectable) rather than hard-depending on an i18n *framework*, to keep dependencies light and not force a second runtime on consumers.
- **RTL** — the DOM-resolves-direction principle, and an audit that renders every component story in both directions and compares geometry to catch *compositional* bugs (logical anchor + unflipped physical transform; double-flipped icons) that a linter can't see.
- **Seeding 29 languages** — a deterministic ICU validator (placeholder + CLDR plural integrity) as a correctness gate, plus a second-model adversarial QA pass that caught systematic nominalization errors.
- **What's still open** — constructed strings (PowerSearch), pseudo-long layout testing, server-side rendering, and the translator adapter.

## Still to do before un-drafting

- Cover image (`coverImage` / `coverAlt`) — not yet added.
- Continued copy iteration.

Not for merge yet — opening as draft for history + review.


nynexman4464:
Superseded by a stacked pair: #4770 (theme-aware image feature) → #4771 (the blog post, stacked on top). Same content, split so the docsite infra lands separately from the post.

Issue #4475, opened 2026-07-28
[Tracker] Accessibility program — H2 guardrails, conformance tests, and bug burn-down


cixzhang:
Umbrella tracker for the accessibility workstream this half: establish long-term a11y guardrails for Astryx components, encode APG/WCAG conformance as automated tests, and burn down the known accessibility bugs.

## Goals

- **Guardrails:** an automated, reusable conformance test suite so a11y regressions are caught in CI, not after release (see the aria-spec suite below).
- **Traceability:** every conformance check references a specific WCAG success criterion or APG pattern by number, so a passing/failing test maps to a concrete spec clause.
- **Completeness:** a rubric/checklist to review each spec test for coverage, since automated checks miss aspects of a11y that still need explicit assertions.
- **Burn-down:** resolve the known accessibility bugs, verified against the new suite.

## Testing infrastructure

- #4112 — Encode APG patterns as reusable ARIA conformance contracts (internal aria-spec suite)
- #4145 — Vibe test: Accessibility dimension is a false positive — invisible to component-level a11y quality
- #3343 — Accessibility & keyboard-management program: shared interaction primitives, then migrate components

## Known accessibility bugs

- #3654 — Widespread WCAG 2.1 AA colour-contrast failures across the design system
- #4421 — WCAG 2.1 A 1.4.1 use-of-colour failure — Breadcrumb Item
- #4414 — Badge `variant="error"` label fails WCAG AA contrast (4.14:1) in the neutral theme
- #4450 — Selector trigger has no accessible name, and its role changes with `hasSearch`
- #3764 — Selector doesn't allow keyboard to select by typing
- #3335 — TabList uses the navigation-links pattern, not ARIA Tabs (no role=tab/tabpanel, aria-selected, arrow-key nav)
- #4132 — CheckboxInput: description isn't part of the clickable label
- #4282 — Chat composer: ArrowUp/ArrowDown recall history mid-text, blocking multi-line caret navigation
- #4173 — Markdown-rendered headings have no id — Outline links never scroll to target

## Conventions (as we develop each spec test)

- **Spec traceability:** each test name embeds the WCAG name + number (or APG pattern id), and carries a machine-readable spec reference property so tests trace back to a clause.
- **Completeness rubric:** derive and apply a checklist/rubric to review each spec test for completeness (informed by prior a11y PRs). Automated conformance ≠ full a11y coverage.
- **Trial validation:** as part of the aria-spec trial, apply the contracts to the relevant existing core components.

[...]

PR #4765, opened 2026-08-06
fix(core): render heading ids in Markdown so Outline hash links scroll (#4173)

### LINKED ISSUE #4173: [Bug] Markdown-rendered headings have no id — Outline links never scroll to their target
Description

Outline's own docs state the id on an outline item "should match the target heading element id," and useOutlineFromMarkdown is provided specifically to derive that id list from markdown text. But Markdown never writes an id onto the h1–h6 elements it renders, and there's no supported way to inject one — MarkdownProps has no id/slug option, and the components.heading override only receives {level, children}, not a slug or index. As a result, clicking an Outline item built from useOutlineFromMarkdown output changes location.hash but never scrolls, because document.getElementById(id) returns null (Outline's click handler silently no-ops when that happens, so it falls back to whatever the browser's default anchor behavior is).

Expected: Markdown + useOutlineFromMarkdown + Outline used together navigate to the correct heading, matching what the docs imply.
Actual: navigation silently does nothing (hash changes, scroll position doesn't).

At minimum, this is a documentation/API-contract issue: the documented Markdown-derived outline IDs cannot be attached to headings through any supported Markdown API.

## Reproduction

```tsx
import {
  Markdown,
  Outline,
  useOutlineFromMarkdown,
} from '@astryxdesign/core';

const markdown = `
# Overview

content

# Installation

content
`;

export default function Page() {
  const items = useOutlineFromMarkdown(markdown);

  return (
    <>
      <Markdown>{markdown}</Markdown>
      <Outline items={items} />
    </>
  );
}
```

1. Render the above.
2. Click the "Installation" item in Outline.
3. Expected: page scrolls to the "Installation" heading.
4. Actual: location.hash becomes #installation, but window.scrollY stays unchanged — no element in the DOM has id="installation" (confirmed via document.getElementById('installation') === null).

This reproduces with the official Outline doc example verbatim plus a bare <Markdown> render — no app-specific code involved.

### Astryx Version

@astryxdesign/core@0.1.7

### Environment

Next.js 16.2.10, React 19.2.4, Chrome (Windows 10)

jiunshinn:
Fixes #4173. Tracked under "Known accessibility bugs" in the accessibility program tracker #4475.

## Root cause

Outline documents a contract that Markdown never fulfilled. The `items` prop docs state the id "should match the target heading DOM id", and `useOutlineFromMarkdown` exists specifically to derive that id list from markdown source. `parseOutlineFromMarkdown` slugifies heading text (`inlineText` -> `slugify` -> `uniqueSlug`), but the Markdown renderer never wrote any `id` attribute onto the h1-h6 elements it produces, and the `components.heading` override only received `{level, children}`, so there was no supported way to inject one. Result: clicking an outline item changes `location.hash`, `document.getElementById(id)` returns `null`, and the scroll silently no-ops, exactly as the reporter's repro shows.

## What changed

- **Single source of truth for slugs.** `inlineText`, `slugify`, and `uniqueSlug` now live in `packages/core/src/Markdown/parser.ts` (exported, internal to the package). `Outline/parseOutlineFromMarkdown.ts` deletes its private copies and imports the shared helpers. Rendered ids and outline ids can no longer drift because they are computed by the same functions.
- **Markdown renders `id` on h1-h6.** A `headingIdMap` memo walks the parsed top-level blocks with one shared duplicate-numbering sequence, mirroring `parseOutlineFromMarkdown`'s traversal exactly (top-level blocks only; headings inside blockquotes and list items are skipped, because the outline never lists them and letting them consume slugs would shift the duplicate numbering and point outline links at the wrong heading).
- **`components.heading` override receives `id?: string`.** Backward-compatible type extension; existing overrides that ignore it keep compiling and rendering unchanged. Custom heading components can now attach the id themselves.
- **Docs.** `Markdown.doc.mjs` gains a usage note (en/dense) about pairing with Outline; `Outline.doc.mjs`'s "Generate items from markdown" example notes that Markdown renders matching ids.
- **Changeset.** Patch bump per the pre-1.0 patch-only policy.

## Hook-placement pitfall

The `headingIdMap` memo sits with the other hooks, directly after the `blocks` memo and well above the `display === 'inline'` early return, with a NOTE comment explaining why it must stay there. A revision of the prior attempt was tripped up by a Rules of Hooks violation here; this placement is covered by the existing inline-display tests plus the new suite.

## The invariant, as a test

`Markdown.test.tsx` gains a `heading ids` suite (5 tests):

1. headings render generated ids (`overview`, `installation`, the issue's exact repro shape)
2. duplicate headings disambiguate (`setup`, `setup-1`, `setup-2`)
3. **parity test**: for one gnarly source (bold/italic formatting, duplicates, punctuation-only heading hitting the `section` fallback, a code-fence decoy, inline code), every id from `parseOutlineFromMarkdown` resolves via `querySelector` to a rendered h1-h6 whose text matches the item label
4. a custom `components.heading` receives the generated ids, including the deduped one
5. headings nested in blockquotes get no id and do not consume slug numbering (locks the traversal-parity rule)

Full run: `packages/core/src/Markdown` + `packages/core/src/Outline` = 7 files, 313 tests, all passing, including `parser.perf.test.ts` (slug helpers are appended utilities; the parse hot path is untouched).

## Prior attempt

#4191 (closed unmerged) had the right blueprint and this PR credits its direction. Differences here: the slug helpers are truly deduplicated (Outline imports the parser's functions instead of keeping a second copy), the parity invariant is enforced by a test that runs both code paths against the same source, nested-heading behavior is explicit and tested rather than incidental, and the `components.heading` `id` prop is documented in the type plus the doc files.

## Reviewer flags

- **Nested headings intentionally get no id.** Matching `parseOutlineFromMarkdown` exactly was prioritized over decorating every heading. If ids on nested headings are wanted, both traversals should be extended together in a follow-up so numbering stays in sync.
- **Slug helpers are exported from `parser.ts` but not re-exported from `Markdown/index.ts`**, keeping the public API surface unchanged. Happy to expose them publicly if maintainers prefer.
- **`id` is applied to the default heading element as a plain JSX attribute** ahead of the `mergeProps(themeProps(...), stylex.props(...))` spread; neither source emits an `id`, so there is no collision today.
- During streaming, the memo recomputes per settled-block update; cost is O(headings) per parse, negligible next to the parse itself.


cixzhang (approved):
Thanks for fixing this!

PR #4768, opened 2026-08-06
PowerSearch: fit the filter editor on narrow screens

### LINKED ISSUE #4761: PowerSearch: filter chip row overflows at narrow widths (no wrap/scroll) — long i18n operator labels run off-screen
## Problem

The PowerSearch filter row (field · operator · value chips + Apply) lays out on a single horizontal line and **does not wrap or horizontally scroll** at narrow viewport widths. When a chip is wide — most easily triggered by a long translated operator label — the row overflows the viewport on both edges: the field chip clips on the left, and the value chip / Apply button run off the right.

## Repro

1. Render a PowerSearch with a filter whose operator is one of the longer labels (e.g. `isAnyOf` / `isNoneOf`).
2. View at a narrow width (~≤380px, i.e. a phone or a narrow side panel).
3. The chip row overflows horizontally with no wrap and no scroll affordance.

Easiest to see with a locale where operators are long — e.g. Turkish `isAnyOf` = "şunlardan herhangi biridir", Vietnamese = "là bất kỳ giá trị nào trong" — but it's fundamentally a **responsive layout** issue, not an i18n bug. Long English field/value text or many filters would hit it too; translation just makes it easy to reproduce.

## Findings (empirical)

Verified by rendering the real component across locales at 1280px and 380px:

- **Desktop (1280px): fine in every locale.** The operator Selector is not truly fixed-width — it auto-widens to fit the full label on one line (no internal clip/ellipsis), and CJK/Arabic contract cleanly. No breakage.
- **Narrow (380px): overflow.** Both when a long operator is the *selected* value in the collapsed trigger chip, and when the operator dropdown is open. The chips render intact individually; the **row** just has no narrow-width behavior. Turkish and Vietnamese `isAnyOf`/`isNoneOf` were the clearest cases.

## Suggested fix

Give the filter chip row a responsive strategy below some breakpoint — options:
- wrap the chips (flex-wrap) so field/operator/value stack when they don't fit, or
- horizontally scroll the chip row with an overflow affordance, or
- constrain individual chip max-widths with internal truncation (ellipsis) as a fallback.

Whichever fits the design intent for PowerSearch in narrow/mobile panels.

## Context

Surfaced while spot-checking machine-translated locales for layout breakage (i18n effort). Desktop is unaffected; this only matters if PowerSearch is used in narrow panels or on mobile. Screenshots available on request.

AKnassa:
## What this does

Makes the PowerSearch filter editor fit on small screens (phones and narrow side panels) instead of running off both edges.

## Why

Reported in #4761: on a ~380px screen, the little editor that pops up (field · operator · value · Apply) insisted on being at least 400px wide, and its row of chips never wrapped. With long translated labels (Turkish or Vietnamese operators, for example) parts of the editor were simply unreachable off-screen.

Fixes #4761

## What changed

- The editor popup no longer forces itself to 400px wide — on small screens it shrinks to fit the space it actually has.
- The row of chips inside wraps onto a second line when space runs short — a container query keyed off the popover's real width, so it only kicks in when the editor is actually narrower than its 400px floor.
- Very long labels now show "…" inside their chip instead of stretching the whole row (this also covers the fixed-width cells in nested filter rows).
- One deliberate behavior change beyond narrow screens: an editor whose bar sits near the window's edge used to keep 400px by flipping across the anchor to the roomier side. It now stays on its own side at the width available there and wraps internally, so it hangs under its bar instead of covering unrelated content.

## How to see it

In Storybook, open any PowerSearch story in a narrow window (~380px) and click a filter token: the editor stays fully on screen and its chips wrap. At desktop widths with the usual full-width bar nothing changes. Verified live in Chrome: at 374px (even with a 24px page gutter) the popover fits and the row wraps to two lines; at 600px it is a single line, identical to before.

## For reviewers

The min-width clamp is `min(400px, calc(100% - spacing-4))` — inside the `position-area` the percentage resolves against the anchor-to-viewport region, so anchors offset from the screen edge are covered too, with a plain-viewport fallback elsewhere. jsdom can't evaluate `@container`, so the tests pin the declarations via StyleX probe classes; narrow behavior was verified in a real browser. The tests also pin the `container-type: inline-size` establishment on the popover root (the one declaration whose removal the probes couldn't previously see), and every probe fails loudly if it resolves to zero atomic classes. Nested sub-filter rows keep their pre-existing tightness between 400–450px popover widths (tree indentation eats into their room); below 400 they now wrap, and that window is left as-is. Changeset included.

## A note on ownership

@nynexman4464 self-assigned this issue the same day they filed it. If you're already working on it, happy to defer to your fix or close this one — no hard feelings.


rubyycheung (approved):

cixzhang:
Merged main in to clear the conflict. The only conflict was `popoverLayerStyles.layer` — I kept your `min(400px, calc(100% - --spacing-4))` floor and dropped the `marginTop`, because [#4951](https://github.com/facebook/astryx/pull/4951) moved that clearance onto the layer itself (`offset: --spacing-1` on the `popover.render` call), so keeping both would have doubled the gap. Nothing else of yours changed; the 399px container query still matches the 400px floor.

packages/core/src/PowerSearch/PowerSearchEditPopover.tsx
```diff
@@ -32,13 +32,23 @@ import type {
   OperatorValue,
 } from './types';
 
+// Below the popover layer's 400px floor (see popoverLayerStyles in
+// PowerSearch.tsx) the chip rows collapse from one line to wrapped lines.
+// A container query, not a viewport one, so the rows track the width the
+// popover actually got (#4761).
+const CHIP_ROW_COLLAPSE = '@container (max-width: 399px)';
```
cixzhang, line 39: 399 has to track the 400px floor in PowerSearch.tsx. Maybe a defineConsts, like DateInput/tokens.stylex.ts.

cixzhang (approved):
Thanks, this is good — keying the chip-row wrap off the popover's own width rather than the viewport is the right call. Merging; one nit inline.

[Reviewed by Robohands]

PR #4769, opened 2026-08-06
feat(core): expand ComplexSelector composition API

ernestt:
## Summary

- Expand `ComplexSelector` with controlled visibility, input/ghost trigger variants, optional leading icons, popup alignment, exact selector sizing, focus restoration, and light-dismiss protection.
- Add Core Storybook and CLI examples for custom grids, deadline pickers, searchable trees, and controlled toolbar triggers.
- Keep the change scoped to the generic Core/CLI composition API.

The TransferList Lab package and its Storybook examples were separated into #4870 so they can land independently without Core review. This PR no longer contains any TransferList files.

## API ownership

These APIs belong to the generic `ComplexSelector` shell rather than a specific selector composition:

- controlled and uncontrolled open state share one popup lifecycle
- trigger presentation is independent from popup content
- popup alignment belongs to the selector shell
- `sm`, `md`, and `lg` use exact 28px, 32px, and 36px trigger heights
- render state exposes open/busy state and stable trigger/content IDs for custom surfaces

## Documentation and examples

- custom grid showcase
- deadline picker
- searchable tree
- controlled toolbar trigger
- updated component docs, tests, exports, Storybook, and changesets

## Validation

- focused ComplexSelector tests pass
- Core TypeScript and documentation typechecks pass
- Core production build passes
- ESLint and Prettier pass
- repository sync, package-boundary, changeset, demo-media, executable-bit, and CLI-structure checks pass


packages/core/src/ComplexSelector/ComplexSelector.tsx
```diff
@@ -208,10 +256,20 @@ export interface ComplexSelectorProps<Value> extends Omit<
...
   width?: SizeValue;
   /** Popup placement. */
-  placement?: 'above' | 'below' | 'start' | 'end';
+  placement?: LayerPlacement;
+  /** Popup alignment along the placement axis. */
+  alignment?: LayerAlignment;
+  /** Whether the selector surface is open. Omit for uncontrolled behavior. */
+  isOpen?: boolean;
+  /** Called when the selector surface requests an open-state change. */
+  onOpenChange?: (isOpen: boolean) => void;
```
cixzhang: The rest of the new props seem okay but I'm worried about controlled open via props. First starters it adds a lot of useEffects below which are all smells, but moreover I find that these tend to result in bad behaviors. We originally had them in Selector to handle showing the popover contents in the docsite which is a really niche use case. I've since gotten rid of it because it was causing issues with focus management. I suspect that we will encounter the same type of issues here.

Instead of `isOpen` and `onOpenChange` as props, perhaps we should consider a `handleRef` and allow for imperative controls on the selector instead. Those controls should be performed as part of a callback rather than as stateful updates.

cixzhang (changes requested):

packages/core/src/ComplexSelector/ComplexSelector.tsx
ernestt: Good call — dropped `isOpen`/`onOpenChange` entirely. Those brought in exactly the `useEffect` sync you flagged, and it's the same shape that caused the focus-management trouble in `Selector`.

Replaced them with an imperative `handleRef` (`ComplexSelectorHandle` → `open()` / `close()` / `toggle()` / `isOpen()`), matching the pattern `Carousel` and `PowerSearch` already use. The selector now solely owns its visibility, so focus restoration and light dismiss stay intact, and there's no external-state round-trip through an effect. The controls run imperatively rather than as stateful updates, as suggested.

Updated the toolbar example to drive the selector through the handle instead of mirroring open state, and refreshed the docs/tests/changeset. Pushed in 8b07cfd.

packages/core/src/ComplexSelector/ComplexSelector.tsx
```diff
@@ -334,38 +456,37 @@ export function ComplexSelector<Value>({
...
             isDisabled && inputWrapperStyles.disabled,
+            variant === 'ghost' && isDisabled && styles.triggerGhostDisabled,
             isDisabled && styles.disabled,
             triggerLabel == null && styles.placeholder,
             xstyle,
           ),
           className,
           style,
         )}>
+        {startIcon &&
```
cixzhang, line 482: Might need isRenderable check

packages/core/src/ComplexSelector/ComplexSelector.tsx
```diff
@@ -334,38 +456,37 @@ export function ComplexSelector<Value>({
...
             styles.triggerContainer,
             styles[size],
-            // The ring belongs to the wrapper (the focusable `<button>` sits
-            // inside it), but it must still be a KEYBOARD ring: `:focus-within`
-            // matched a mouse click on the trigger and drew the outline for
-            // pointer users too. `focusWithin` here is `:has(:focus-visible)`.
-            focusOutlineStyles.focusWithin,
+            variant === 'input' && focusOutlineStyles.focusWithin,
+            variant === 'ghost' && styles.triggerGhost,
+            variant === 'ghost' && focusOutlineStyles.focusWithin,
```
cixzhang, line 472: Hmm some of this seems redundant. Probably can apply the focus outline without a variant check

packages/core/src/ComplexSelector/ComplexSelector.tsx
```diff
@@ -284,24 +363,67 @@ export function ComplexSelector<Value>({
...
     surfaceTarget: 'complex-selector-popup',
-    onHide: () => {
-      document.getElementById(triggerId)?.focus();
-    },
+    onHide: handlePopoverHide,
   });
 
+  const isOpen = popover.isOpen;
+
+  const handleTriggerClick = useCallback(() => {
```
cixzhang, line 388: I need to double check this with other recent updates to fix this across the board for menus

cixzhang (approved):
Thanks this is good. Will likely merge as is then do the nits separately

PR #4770, opened 2026-08-06
feat(docsite): theme-aware (light/dark) images in blog post bodies

nynexman4464:
## What

Theme-aware images for blog post bodies (**PR 1 of 2** — the infrastructure; the post that uses it is stacked on top).

- **`ThemedImage`** — takes explicit `lightSrc` / `darkSrc`; if only one is given, it's always used.
- **`MarkdownImage`** — adapter wired as the blog `Markdown` image override. Authors drop a sibling `<name>.dark.<ext>` next to an image and it's shown in dark mode automatically.
- **Build-time resolution** — `generate-data.mjs` scans `public/blog/**` for `<name>.dark.<ext>` files and emits a `blogDarkImages[]` manifest into the generated registry. `MarkdownImage` consults that manifest, so the dark-variant decision is deterministic at build time — no runtime existence probe, no flash.
- Honors the docsite's **manual theme toggle** (via `useThemeMode`), not just OS `prefers-color-scheme`: `<picture>` + `prefers-color-scheme` before the toggle resolves, a resolved `` after.
- Body images get a subtle frame so screenshots read intentionally on either background.

No behavior change for existing posts: with no `.dark.` sibling, an image renders once (framed), exactly as before.

## Testing

- Docsite typechecks (the module-resolution noise in a fresh worktree is pre-existing "core not built," unrelated).
- `generate-data.mjs` runs; `blogDarkImages` populates from `public/blog/**`.

## How it works

Authoring stays plain Markdown. An author references only the light image:

```md
![alt text](/blog/<slug>/foo.png)
```

To add a dark variant, they drop a sibling file named `foo.dark.png` next to it — that's the whole contract. No frontmatter, no `<picture>`, no per-image config.

At build time, `generate-data.mjs` walks `public/blog/**` and records every `*.dark.*` file into a `blogDarkImages` manifest in the generated registry. The Markdown `image` override (`MarkdownImage`) derives the candidate `foo.dark.png` path from `foo.png` and uses it **only if it's in that manifest** — so the light/dark decision is resolved at build time, not by probing for the file in the browser. `ThemedImage` then renders the right one for the resolved color mode (and honors the manual toggle via `useThemeMode`, falling back to `<picture>` + `prefers-color-scheme` for the pre-hydration `system` case). Posts with no `.dark.` sibling render a single framed image, exactly as before.

## Why a filename convention (and not explicit props)

`ThemedImage` itself takes explicit `lightSrc` / `darkSrc`, which is the API you'd want if you were writing TSX. But blog post bodies are **plain Markdown (`.md`, not MDX)** — they're rendered by the core `Markdown` component from a string, so you can't drop a `<ThemedImage lightSrc=... darkSrc=... />` into a post. The only extension point is `Markdown`'s `image` component override, and an image node carries just `{ src, alt }` — a single source. There's no way to thread a second URL through Markdown image syntax.

So the convention (`foo.png` + `foo.dark.png`) is what bridges that gap: `MarkdownImage` takes the one `src` Markdown gives it, derives the dark candidate, checks the build-time manifest, and hands both to `ThemedImage`. The explicit-props component stays clean for any future TSX use; the convention adapts it to the constraints of Markdown.

## Alternatives considered

- **Runtime existence probe** (in-browser `new Image()` to check if a `.dark` file exists). Rejected — an extra request per image and a possible flash before it resolves. The build-time manifest is deterministic and free at render.
- **CSS-only `<picture>` + `prefers-color-scheme`.** Rejected as the sole mechanism — it ignores the docsite's *manual* theme toggle, so it'd disagree with the toggle. We use it only for the pre-hydration/`system` case and let `useThemeMode` win after.
- **Author writes explicit `<picture>`/dual `` in the post.** Rejected — posts are plain `.md` (not MDX), and it pushes theme logic onto every author. The `<name>.dark.<ext>` convention keeps markdown clean.
- **`prefers-color-scheme`-only `srcset` with no JS.** Same toggle-blindness problem; kept the component so the manual toggle is honored.


nynexman4464:
Preview of how this looks:


https://github.com/user-attachments/assets/e4e22ae2-6d2f-405b-8472-c3eade9e4957

PR #4774, opened 2026-08-07
feat(lint): automate the rubric's §2 theming-target checks

cixzhang:
## What this is

ESLint rules that automate the mechanically checkable part of the [Component Audit Rubric](https://github.com/facebook/astryx/wiki/Component-Audit-Rubric)'s §2. Registered on the plugin, in **neither** shipped config, not wired into `eslint.config.js` — turning one on is a one-line change once we agree on the tier.

**This PR used to be two things: proposed wiki text defining theming-target criteria, plus lint encoding it. The proposal is gone.** The rubric now carries numbered §2 checks with their own severities and exceptions — T6, T7, T12, T26, T27 — so there is nothing left to propose. Every check here names the rubric id it automates, in the message text, so a report reads `T6: theme target 'x-option' is on an element whose styles vary with…` and the author can look up exactly what was applied.

**Anything that could not name a rubric id was removed**, rather than kept as a proposal:

| Removed | Why |
|---|---|
| `missingPosition` — the `{parent}-{position}-{component}` shape | No published check states it, and it fires on `banner-icon`: a shipped public target with a codemod behind it, unrenameable |
| `layoutOnlyRootTarget` (55 hits) | T7 exempts a component's root target outright — it is the component's address, not a seam anyone chose. Now skipped unconditionally instead of hidden behind an option |
| `stateVariesOnlyLayout` | Encoded a refinement of T7 that T7 does not make, and it can no longer see the case that motivated it |
| `inheritablePropertyOnChild` (108 hits) | Most hits are correct code — a caption *should* differ from its card. Kept only the render-prop-fallback form, where the divergence is structural |

Also gone: the `.github/copilot-instructions.md` checklist this PR added. It restated rules the rubric owns, which is the duplication that caused a day of drift fixes last week. What is left there is the two facts a reviewer cannot get from the rubric alone — that T6 and T7 have no enforcer, and that these rules are in neither config, so a green `pnpm lint` says nothing about a new target.

## The one worth having

**`underDeclaredState` automates T6**, which the rubric marks BLOCK, detects by grep, records as *"historically the single most frequent finding"*, and has no lint rule for. It catches `fooStyles[prop]` inside `stylex.props()` with no `prop` in the sibling `themeProps()` call — a style-driving prop a theme cannot reach.

Its **20 hits are a real pre-existing backlog** (Calendar, Code, CommandPaletteItem, FileInput, InputGroup, Link, MobileNav, MultiSelector, …), each needing a human call about which prop belongs on the target. Hence `warn`: at `error` it fails CI on `main`.

## Measured on current `main`

| Rule / check | Rubric | Count | Proposed tier |
| --- | --- | --- | --- |
| `theming-target-shape` / `underDeclaredState` | **T6** | 20 | `warn` |
| `theming-target-shape` / `layoutOnlyTarget` | T7 | 7 | `warn` |
| `theming-target-shape` / `wrapperTarget` | T7 | 2 | `warn` |
| `theming-target-shape` / `unstyledTarget` | T7 | 0 | `error` |
| `theming-target-shape` / `targetOnRenderPropFallback` | T27 | 0 | `warn` |
| `theming-target-shape` / `inheritableOnRenderPropFallback` | T7/T27 | 0 | `warn` |
| `theming-target-name` / `appearanceInComponentSlot` | T27 | 2 | `warn` |
| `theming-target-name` / `stateSubTarget` | T26 | 0 | `error` |
| `themeprops-reflection` / `bypassedThemeProps` | T26 | 9 | `warn` |
| `themeprops-reflection` / `classNameOnly` | T12 | 3 | `warn` |
| `themeprops-reflection` / `clobberedByLaterProp` | T12 | 0 | `error` |
| `themeprops-reflection` / `droppedStateReflection`, `handAuthoredState` | T26 | 0 | `error` |

Counts re-run against `main` for this revision, not carried over. Every one of them moved over the eleven days this branch sat: T6 went 17 to 20, layout-only 6 to 7, wrapper 4 to 2, bypassed `themeProps` 7 to 9, className-only 4 to 3.

## The live bug it was written for is gone

`clobberedByLaterProp` counted exactly one hit when this branch was cut:
`RichTextEditor` wrote `className={className}` after the `themeProps` spread, so
with no consumer `className` the theme class resolved to `undefined`. That file
has since moved to `packages/richtext` and was rewritten through `mergeProps` on
the way, and [#4781](https://github.com/facebook/astryx/pull/4781) — the fix
this section used to point at — is closed. The check now measures 0 and ships as
a pure regression guard. `MoreMenu`'s hand-built target is still open in
[#4782](https://github.com/facebook/astryx/pull/4782); it is 2 of the 9
`bypassedThemeProps` hits.

## Risk

None to consumers: no component source changes, and the rules run nowhere. The `underDeclaredState` backlog is the only thing that would bite, and only if someone enables it at `error`.

## Testing

`pnpm vitest run internal/eslint-plugin-astryx` — 31 files, 653 tests, green. Counts above produced by running the rules over `packages/{core,lab}/src/**` through the ESLint Node API.

Rebased onto current `main` (`41494fc`). `tsc --noEmit` is identical to `main` in the same tree, 776 errors either way, all pre-existing.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -81,68 +81,20 @@ change.
 
 A `themeProps()` target is a public API commitment: a stable `.astryx-*` class
 an unknown external consumer writes CSS against, and one we cannot rename
-without breaking them. The authoring principles live in
-[Theming Infrastructure](https://github.com/facebook/astryx/wiki/Theming-Infrastructure)
-("Principles for authoring theming targets") — that page is the source of
-truth. Run every **new** target through this checklist:
+without breaking them. The checks are **T6, T7, T12, T26, T27 and T33** in the
+rubric's §2 — read them there and cite them by id; this file does not restate
+them.
 
-1. **Does the element paint?** A target is a paint seam — color, background,
-   border, font, radius, shadow. An element that only lays out (`display`,
-   flex/grid, `margin`/`padding`, width/height, `transform`) has nothing for a
-   theme to restyle.
-2. **Is the layout value declared?** Layout is themeable only where the
-   component declared it and still owns the structure — a component var, a
-   `derived` entry, container-padding expansion. Arbitrary layout CSS on a class
-   target is not.
-3. **Is it attached to the component?** If the themed thing is an internally
-   composed Astryx component, the target rides that instance's
-   `className`/`xstyle` passthrough. Never mint a wrapper to hold a target, and
-   where a wrapper legitimately remains it carries **layout only** — `display`,
-   alignment, `flexShrink`, positioning. Anything that paints (color,
-   background, border, shadow) belongs on the element that renders the pixels.
-4. **Is the name compositional?** `{parent}-{position}-{component}` — position
-   from the shared vocabulary (trigger/option/item/row/header/leading/trailing/
-   menu), and the last segment the actual component (`icon`, `checkbox`,
-   `button`, `divider`), never an appearance (`check`, `caret`, `marker`).
-   State is never a name segment: it rides `themeProps({selected})`, which emits
-   the class token and the `data-*` attribute together.
-5. **Could the parent target reach it by inheritance?** If the property
-   cascades (font, color) **and the parent target can actually deliver it as the
-   code stands**, put it on the parent target rather than minting a child one —
-   a child target on a `renderX` fallback silently misses every custom-rendered
-   result (principle 4). Inheritance does **not** arrive when the child sets the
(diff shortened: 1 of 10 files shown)


cixzhang:
Closing this prototype. Three large rules that are registered and documented but disabled are maintenance surface without enforcement. A future rule should start from one concrete regression, ship narrowly, and be enabled in the same change.

PR #4776, opened 2026-08-07
Grapheme-safe text: count characters the way people do

### LINKED ISSUE #4759: i18n: grapheme-safe string manipulation (TextArea counter, PowerSearch truncation, capitalize)
## Problem

Several places in `@astryxdesign/core` index strings by UTF-16 code unit (`.length`, `.charAt()`, `.slice()`) rather than by user-perceived grapheme clusters. This produces incorrect behavior on any multi-code-unit input: emoji (counts as 2+), flag sequences (4+), ZWJ family emoji (7+), and characters with combining marks.

PR #4750 fixed Avatar's `getInitials()` with `Intl.Segmenter`. This issue tracks applying the same fix across the rest of the codebase and extracting a shared utility.

### Real bugs (user-visible)

1. **TextArea character counter** — `optimisticValue.length` / `maxLength - length` displays the wrong count to sighted users and announces the wrong count to screen readers. One emoji shows as 2 characters, hitting limits "early" and showing nonsensical counts.
   - Files: `packages/core/src/TextArea/TextArea.tsx` (lines 433, 445, 452, 468, 564, 590, 598)

2. **PowerSearch truncation** — two helpers slice at a code-unit offset and append an ellipsis. Slicing mid-grapheme can split a surrogate pair (`U+FFFD` replacement character) or drop a combining mark.
   - `packages/core/src/PowerSearch/formatFilterValue.ts:21` — `str.slice(0, maxLength - 1) + '…'`
   - `packages/core/src/PowerSearch/PowerSearch.tsx:121` — `value.slice(0, limit) + '...'`

### Lower-severity

3. **Table `capitalize()`** — `str.charAt(0).toUpperCase() + str.slice(1)` — same pattern as Avatar. Only hits auto-derived column headers from data keys (usually ASCII), but it's the identical bug and cheap to fix.

### Not hazards (checked, ruled out)
- Markdown parser `slice()` — operates on parser-tracked code-unit offsets with its own semantics. Not display truncation.
- All `.slice(0, n)` on **arrays** (MultiSelector badges, Typeahead results, FileInput) — element counts, not string graphemes. Fine.

## Proposed fix

1. **Extract `utils/grapheme.ts`** with:
   - `graphemeLength(str)` — grapheme-cluster count (`Intl.Segmenter` with code-point fallback)
   - `firstGrapheme(str)` — replaces `.charAt(0)` safely
   - `truncateGraphemes(str, max, ellipsis?)` — replaces `str.slice(0, n) + '…'`

2. **Migrate Avatar** (#4750) to consume the shared util (removes inline implementation).

3. **Fix TextArea** — use `graphemeLength()` for the counter display and the remaining/over-limit calculations.

4. **Fix PowerSearch** — use `truncateGraphemes()` in both truncation helpers.

5. **Fix Table `capitalize()`** — use `firstGrapheme()`.

## Context

Raised by @cixzhang in review of #4750. Related: the i18n effort surfaced this as a broader string-safety concern — scripts with combining marks, emoji-heavy input, and RTL mixed content all exercise the same code path.

AKnassa:
## What this does

Makes character counting and text shortening treat an emoji, flag, or accented character as one character — the way people count — instead of the way computers store them.

## Why

Reported in #4759: several spots indexed text by internal storage units. The TextArea character counter told you one emoji was "2 characters" (and announced the wrong count to screen readers), PowerSearch could slice an emoji in half when shortening a filter token (leaving a broken � character), and auto-generated Table headers couldn't capitalize some scripts.

Fixes #4759

## What changed

- One new shared utility measures and cuts text by user-perceived characters (`characterCount`, `firstCharacter`, `truncateCharacters` — also exported for consumers).
- The TextArea counter, its over-limit warning, and its screen-reader announcements all use the human count now — and textareas without a limit skip the counting work entirely.
- PowerSearch token shortening cuts between characters, never through one, keeping its exact previous lengths and ellipsis styles.
- Table headers capitalize correctly for all scripts; Avatar's initials (fixed in #4750) now share the same utility instead of their own copy.

## How to see it

In Storybook, open TextArea's Default story, set a maxLength, and type `Hi 😀👍🇹🇷👨‍👩‍👧‍👦!` — the counter reads **8**, not 23. Verified live in Chrome; also fuzz-verified that plain-English behavior is byte-for-byte unchanged everywhere.

## For reviewers

The two truncation helpers keep their exact thresholds (`maxLength − 1` + `…` and the `limit + 3` grace window + `...`) — one shared function expresses both. The multi-item list gates (`joined.length <= maxLength`) deliberately stay code-unit-based: they never slice, so no corruption is possible, and changing them would alter when lists collapse to "N items". Two changesets ([fix] + [feat] for the new public utils, both patch per the 0.x rule). A same-family issue in `useStreamingText` (reveal ticks can transiently split emoji) surfaced during review and is filed as #4779.

## A note on ownership

@nynexman4464 self-assigned this issue the same day they filed it. If you're already working on it, happy to defer to your fix or close this one — no hard feelings.


packages/core/src/utils/grapheme.ts
```diff
@@ -0,0 +1,87 @@
...
+  }
+  return [...str];
+}
+
+/**
+ * Number of user-perceived characters in a string — the count a person would
+ * give, where one emoji, flag, or accented character is one character.
+ * Grapheme-safe replacement for `.length` on user-visible strings.
+ */
+export function graphemeLength(str: string): number {
```
nynexman4464: Can we call these exported functions something other than `grapheme`?  It's kind of technical.  I'd prefer something like `localeSafe` or `multiLanguageSafe` something along those lines.

nynexman4464 (commented):
Thanks for taking this on!  I have some minor feedback on naming (I think `grapheme` is a little too technical) but otherwise LGTM

packages/core/src/utils/grapheme.ts
AKnassa: Renamed: `characterCount`, `firstCharacter`, `truncateCharacters`. The file, internal helpers and doc prose follow suit. The word survives only as the literal `Intl.Segmenter` granularity value, with a comment noting why. No behavior change.

nynexman4464 (approved):
great thanks for helping to fix!

cixzhang (approved):
Thanks I like this consolidation!

PR #4782, opened 2026-08-07
chore(core,lab): route MoreMenu through themeProps and close the targets guard

cixzhang:
## What

Two related things, shipped together because the second only passes because of
the first.

**1. `MoreMenu` builds its theme target by hand.**
`packages/core/src/MoreMenu/MoreMenu.tsx:126-130` (pre-fix):

```tsx
className={
  classNameProp
    ? `${stableClassName('more-menu')} ${classNameProp}`
    : stableClassName('more-menu')
}
```

`astryx-more-menu` was the only target in the system with no `themeProps()` call
behind it, and `MoreMenu.tsx` was the only `stableClassName()` call site left
outside `naming.ts` / `themeProps.ts` (verified by grep across
`packages/core/src` and `packages/lab/src`). Consequences: the target can never
carry a `data-*` reflection the way every sibling does, the consumer-className
merge is hand-rolled rather than `mergeProps()`, and — see below — the drift
guard could not see it.

Now:

```tsx
{...mergeProps(themeProps('more-menu'), {className: classNameProp})}
```

**Should it carry state?** No. The target lands on `DropdownMenu`'s popover
panel, which only exists while the menu is open, so an `open` state would
reflect a constant. `MoreMenu` also cannot observe the uncontrolled open state
without new API. Noted in a one-line comment at the call site.

**2. `themingTargets.test.ts` was porous in two directions.**

- **It only scanned `packages/core/src`.** `packages/lab` renders 10 targets and
  3 prop keys that nothing checked.
- **It only checked rendered ⊆ documented.** The reverse — a documented target
  nothing renders — was unguarded, and sat at exactly **1**: `astryx-more-menu`.
  A documented selector that nothing emits is the worse failure of the two: a
  theme author writes CSS against it and it silently matches nothing.

The `docs` block is now checked **package-wide across core and lab, in both
directions**, plus the prop-key check.

Package-wide rather than per-directory because a class is legitimately rendered
in one directory and documented in another — `Code/` and `Heading/` ship no doc
of their own (CodeBlock and Text document them) and `hooks/` renders
`astryx-input-status-icon`, which Field documents. The old per-directory shape
could only tolerate those by `continue`-ing past any directory with no
`${dir}.doc.mjs`, which is also how `packages/lab/src/Chat/` (four targets, one
doc file per component rather than a `Chat.doc.mjs`) stayed invisible.

`docsZh` stays a per-directory parity check: only some components carry a zh
block, so a package-wide set would report every untranslated component as drift.

### The 9 lab targets — all documented, none deleted

Every one is a real, visible element already rendering through `themeProps()`
with base styles of its own, so the fix is documentation, not deletion:

| target | doc | keys |
| --- | --- | --- |
| `astryx-chat-emoji-picker` | `Chat/ChatEmojiPicker.doc.mjs` | — |
| `astryx-chat-reaction-bar` | `Chat/ChatReactionBar.doc.mjs` | — |
| `astryx-chat-typing-indicator` | `Chat/ChatTypingIndicator.doc.mjs` | — |
| `astryx-chat-unread-divider` | `Chat/ChatUnreadDivider.doc.mjs` | — |
| `astryx-chat-reasoning` | `ChatReasoning.doc.mjs` | `states: [expanded, streaming]` |
| `astryx-codeeditor` | `CodeEditor.doc.mjs` | `visualProps: [size, language]` |
| `astryx-log-stream` | `LogStream.doc.mjs` | `visualProps: [variant]` |
| `astryx-svg-icon` | `SVGIcon.doc.mjs` | `visualProps: [variation, size, color]` |
| `astryx-schedule` | `Schedule.doc.mjs` | — |

### What the rebase dropped

Eleven days of `main` turned three pieces of this branch into duplicates or
worse. All three are gone; the reasoning is in the third commit.

- **ListInput's theming block** — [#4966](https://github.com/facebook/astryx/pull/4966)
  documented `astryx-list-input` on `main`. Both blocks land in the same `docs`
  object literal, so the merge produced a duplicate `theming` key where the
  later one silently wins. No conflict, no marker.
- **TransferList's theming block** — same shape; `main` documents all four
  targets already.
- **Five ListInput a11y baseline entries** — [#4965](https://github.com/facebook/astryx/pull/4965)
  removed the invalid `aria-required` they were baselining, so re-adding them
  would re-permit a violation that no longer exists. The three RichTextEditor
  and Schedule entries stay: those violations are still live, and both
  components are still in `pr-a11y`'s scope for this branch.

`astryx-step-connector`, the tenth lab target, is documented on `main` now too.

### And one the rebase added

`RichTextEditor` moved to `@astryxdesign/richtext` while this sat, taking its
doc file out of a guard that only knew core and lab — which is the hole this
branch exists to close. The package joins the scan. Its `size`/`status` props
are red without the doc fix here and green with it; `@astryxdesign/charts` stays
out because it renders no theme target at all.

`main`'s own additions to this file are kept inside the rewrite rather than lost
to it: conditional-spread prop keys ([#5016](https://github.com/facebook/astryx/pull/5016),
the fix that made `Heading`'s `type` visible) and the cross-directory doc-file
lookup, which the `docsZh` half still needs even though the package-wide `docs`
half no longer does.

## Test plan

**The reverse guard is red → green on the thing it exists for.** With
`MoreMenu.tsx` reverted to the `stableClassName()` form and everything else on
this branch:

```
× @astryxdesign/core theming.targets > every documented target is actually rendered
    AssertionError: @astryxdesign/core documents 1 theming target(s) that no
    themeProps() call renders …
    + "astryx-more-menu (MoreMenu/MoreMenu.doc.mjs)"
```

With the fix restored it passes. Before the lab docs were written, the lab
package reported exactly the 10 classes and 3 prop keys listed above.

`MoreMenu.test.tsx` gains two cases: a consumer `className` still lands
alongside the theme class, and `astryx-more-menu` and `astryx-dropdown-menu`
both survive on the shared panel (a merge bug in either would drop the other).

```
pnpm exec vitest run --project ui packages/core/src/theme packages/core/src/MoreMenu
  # 19 files, 635 tests pass (themingTargets alone: 119)
pnpm exec vitest run --project ui packages/core/src/{DateInput,DateRangeInput,MultiSelector,Selector}
  # 6 files, 583 tests pass
pnpm exec vitest run packages/cli/.../build-theme.registry.test.mjs packages/cli/api/theme
  # 9 files, 77 tests pass
pnpm -F @astryxdesign/{lab,richtext,core} typecheck:docs                            # clean
pnpm exec tsc --noEmit    # 776 errors, byte-identical to main in the same tree
```

Rebased onto current `main` (`41494fc`).

## Visual verification

Playwright + Chromium against local Storybook (`core-moremenu--default`, menu
opened), on `main` and on this branch, with real `astryx theme build` output for
`'more-menu': {base: {backgroundColor: 'var(--color-background-red)'}}` injected
into the live `@scope ([data-astryx-theme="neutral"])` / `@layer astryx-theme`.
Screenshots in `~/theming-bugfix-shots/more-menu__*`.

**The result is that nothing changes, and that is the honest finding.** The
themed menu panel repaints identically on both: the same 9,005 pixels in the
same bbox (434,280)–(519,388), `#FFFFFF`/`#F2F2F2` → `#FACECB`/`#EDC3C0`, and
`main` vs branch unthemed is a 0-pixel diff. `astryx-more-menu` *did* reach the
DOM before this change — `stableClassName()` produces the same string — so this
is **not** a rendering fix and the changeset says `[chore]`, not `[fix]`. What
changes is capability and visibility: the target now sits on the same
reflection surface as every other one, and the guard can see it.

## For a reviewer to weigh in on

- **Which element should own `astryx-more-menu`?** Today it lands on
  `DropdownMenu`'s popover panel, sharing the element with
  `astryx-dropdown-menu` — so `more-menu` cannot theme the trigger button at
  all. Kept as-is here (minimal blast radius; moving it is a public selector
  change), but it looks like the wrong element for the name.
- **Overlap with `build-theme.registry.test.mjs`.** Its "every documented
  theming target is backed by a real themeProps literal" check is the same
  direction as the guard added here, by regex, core-only — and it accepts a
  `stableClassName()` literal as evidence, which is exactly how
  `astryx-more-menu` passed it. [#5449](https://github.com/facebook/astryx/pull/5449)
  has since taught it `legacyNames` too. With MoreMenu fixed, its
  `stableClassName` arm is dead weight and the check is a strictly weaker
  duplicate of the AST-based one here. Left alone to keep this diff off the
  CLI; worth a follow-up.

@cixzhang

### DIFF
--- a/.changeset/theming-targets-guard.md
+++ b/.changeset/theming-targets-guard.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[chore] `MoreMenu` builds its `astryx-more-menu` theme target through `themeProps()` instead of hand-assembling it with `stableClassName()`. The rendered DOM is unchanged — the class and any consumer `className` land on the same element as before — but the target now sits on the same reflection surface as every other one, so it can carry `data-*` state if it ever needs to. The four `*-clear-icon` aliases passed to `InputClearButton` move to `themeProps().className` for the same reason, again emitting the same class.
+
+@cixzhang
--- a/.changeset/theming-targets-guard.md
+++ b/.changeset/theming-targets-guard.md
@@ -2,6 +2,6 @@
 '@astryxdesign/core': patch
 ---
 
-[chore] `MoreMenu` builds its `astryx-more-menu` theme target through `themeProps()` instead of hand-assembling it with `stableClassName()`. The rendered DOM is unchanged — the class and any consumer `className` land on the same element as before — but the target now sits on the same reflection surface as every other one, so it can carry `data-*` state if it ever needs to. The four `*-clear-icon` aliases passed to `InputClearButton` move to `themeProps().className` for the same reason, again emitting the same class. The theming-targets guard now also checks the reverse direction — a documented target that nothing renders — package-wide across core and lab, so `theming.targets` stays a list of selectors that really exist.
+[chore] `MoreMenu` builds its `astryx-more-menu` theme target through `themeProps()` instead of hand-assembling it with `stableClassName()`. The rendered DOM is unchanged — the class and any consumer `className` land on the same element as before — but the target now sits on the same reflection surface as every other one, so it can carry `data-*` state if it ever needs to. The four `*-clear-icon` aliases passed to `InputClearButton` move to `themeProps().className` for the same reason, again emitting the same class. The theming-targets guard now also checks the reverse direction — a documented target that nothing renders — package-wide across core, lab and richtext, so `theming.targets` stays a list of selectors that really exist.
 
 @cixzhang
--- a/.changeset/theming-targets-guard.md
+++ b/.changeset/theming-targets-guard.md
@@ -2,6 +2,6 @@
 '@astryxdesign/core': patch
 ---
 
-[chore] `MoreMenu` builds its `astryx-more-menu` theme target through `themeProps()` instead of hand-assembling it with `stableClassName()`. The rendered DOM is unchanged — the class and any consumer `className` land on the same element as before — but the target now sits on the same reflection surface as every other one, so it can carry `data-*` state if it ever needs to. The four `*-clear-icon` aliases passed to `InputClearButton` move to `themeProps().className` for the same reason, again emitting the same class.
+[chore] `MoreMenu` builds its `astryx-more-menu` theme target through `themeProps()` instead of hand-assembling it with `stableClassName()`. The rendered DOM is unchanged — the class and any consumer `className` land on the same element as before — but the target now sits on the same reflection surface as every other one, so it can carry `data-*` state if it ever needs to. The four `*-clear-icon` aliases passed to `InputClearButton` move to `themeProps().className` for the same reason, again emitting the same class. The theming-targets guard now also checks the reverse direction — a documented target that nothing renders — package-wide across core and lab, so `theming.targets` stays a list of selectors that really exist.
 
 @cixzhang
--- a/.github/a11y-baseline.json
+++ b/.github/a11y-baseline.json
@@ -604,31 +604,6 @@
       "helpUrl": "https://dequeuniversity.com/rules/axe/4.12/color-contrast?application=playwright"
     },
     {
-      "key": "ListInput::Empty Mailing List::aria-allowed-attr",
-      "impact": "critical",
-      "helpUrl": "https://dequeuniversity.com/rules/axe/4.12/aria-allowed-attr?application=playwright"
-    },
-    {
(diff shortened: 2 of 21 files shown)


cixzhang:
Closing rather than extending this mixed branch. It now combines MoreMenu cleanup, target-documentation edits, a cross-package guard, and accessibility baselines; recent main also superseded the ListInput and TransferList pieces. The remaining Rich Text target contract needs one focused change that updates the declared axes, cross-package guard, and SYNC target together.

PR #4783, opened 2026-08-07
fix(core): close the two holes in the derived-var drift guard

cixzhang:
## What

`packages/core/src/theme/derivedVarRegistry.test.ts` guards component CSS vars
against their docs. It had two exemptions wide enough to hide real drift, and
both are now closed while the counts are still small.

### Hole 1 — the source scan skipped every `--_*` name

`extractComponentVars()` (pre-fix `:104-107`):

```ts
// Skip private vars (--_ prefix = internal, not themeable)
if (varName.startsWith('--_')) {
  continue;
}
```

Private vars *are* internal in the sense that a theme author does not set them
directly — `astryx theme build` errors if a theme writes one. But they are still
part of the documented theming surface: `ComponentThemingVar` has a
`private: true` flag for exactly this, `Card.doc.mjs` documents `--_card-radius`
that way, and the derived-var pipeline is how a theme reaches them. Skipping the
prefix meant a private var could be declared in source and described nowhere.

Re-measured: **11 distinct private var names across 13 (component, var) sites**
were undocumented. (11 names, 13 sites — `--_avatar-group-overlap` and
`--_tab-indicator-bottom` are each declared in two directories. If the audit's
"11" was counting sites it was two low; as a count of names it is exact.)

They resolve as **9 documented in the owning component's doc**:

| doc | var |
| --- | --- |
| `Avatar.doc.mjs` | `--_avatar-group-overlap` |
| `Card.doc.mjs` | `--_card-elevation`, `--_card-ring` |
| `CodeBlock.doc.mjs` | `--_codeblock-gutter-width` |
| `TabList.doc.mjs` | `--_tab-indicator-bottom` |
| `TextArea.doc.mjs` | `--_textarea-inline-padding` |
| `TreeList.doc.mjs` | `--_tree-indent`, `--_tree-focus-outline`, `--_tree-focus-outline-offset` |

and **4 classified as cross-component** — a component setting a var a sibling
owns and documents, which `CROSS_COMPONENT_VARS` already models for
`--_button-radius`:

| setter | var it sets on a child |
| --- | --- |
| `AvatarGroup` (`AvatarGroupOverflow.tsx:126`) | `--_avatar-group-overlap` |
| `Breadcrumbs` (`BreadcrumbItem.tsx:218-219`) | `--_dropdown-menu-radius`, `--_dropdown-menu-padding` |
| `SelectableCard` (`SelectableCard.tsx:116`) | `--_card-ring` |
| `Toolbar` (`Toolbar.tsx:108`) | `--_tab-indicator-bottom` |

### Hole 2 — the derived[]-coverage check filtered on a name regex

Pre-fix `:263-269`:

```ts
const themeableVars = missingDerived.filter(v => /radius|padding/.test(v));
```

A documented var with no `derived[]` entry was exempt from ever needing one if
its **name** happened not to contain "radius" or "padding". Three vars were
already sitting in that gap (`--button-focus-offset`,
`--button-icon-only-aspect`, `--tree-list-indent`), and every private var
documented by this PR would have joined them.

Replaced with an explicit `VARS_WITHOUT_DERIVED_MAPPING` set: everything must
have a `derived[]` entry unless it is listed, and each entry carries the reason
(composed into a shared `box-shadow` list; set per-row on `:focus-visible`;
consumed inside `calc()` rather than emitted as the property; …). A new var now
has to be argued into the list instead of slipping past on its name, and the
list is meant to shrink.

## Widening it surfaced a live bug

`ContextMenu.doc.mjs:19-22` documents:

```js
derived: [
  {property: 'padding', vars: ['--_dropdown-menu-padding']},
],
```

but `derivedVarRegistry` had no `context-menu` key, so the mapping never ran:
`components: {'context-menu': {base: {borderRadius: '12px'}}}` emitted
`border-radius` alone while `ContextMenu.tsx:110` kept reading its own
`var(--_dropdown-menu-radius)`. ContextMenu was invisible to the "every doc with
`theming.derived` has a registry mapping" test only because its two vars are
both private, so the directory never entered `discoverComponents()` at all.

Added the registry entry, mirroring the existing `dropdown-menu` one. This is
the one consumer-visible change in the PR and has its own `[fix]` changeset.

## Test plan

Red → green, staged:

1. Removing the `--_` skip alone: 6 directories fail
   (`Avatar`, `Card`, `CodeBlock`, `TabList`, `TextArea`, `TreeList`) naming
   exactly the 9 vars above, plus
   `ContextMenu: has theming.derived but no DIR_TO_REGISTRY_KEY mapping`.
2. Adding the cross-component classifications drops the 4 setter-side reports.
3. Adding the 9 doc entries + the `context-menu` registry entry: green.

```
pnpm exec vitest run --project ui packages/core/src/theme   # 15 files, 655 tests pass
pnpm exec vitest run --project ui                           # 6,449 tests; the only
  # two failures are a perf test and a focus test that both pass in isolation —
  # load flakes, unrelated (Markdown/parser.perf, TabList roving tabindex)
pnpm exec eslint packages/core/src/theme/derivedVarRegistry.{ts,test.ts}   # clean
pnpm exec tsc --noEmit    # 627 errors, identical to main (pre-existing)
node scripts/check-changesets.mjs                                          # ✓
```

## Notes

- No component source changes beyond the one-line registry addition; the rest is
  a test and nine `theming.vars[]` doc entries.
- **`--_textarea-inline-padding` is the honest loose end.** It is on the
  allowlist, but `TextArea.tsx:70-74` says in a comment that it exists "so it can
  later be driven by a theme `padding` translation." Giving it a `derived[]`
  entry plus a `text-area` registry key would do exactly that — a real feature,
  not a guard fix, so it is out of scope here and flagged rather than smuggled in.

@cixzhang

### DIFF
--- a/.changeset/context-menu-derived-vars.md
+++ b/.changeset/context-menu-derived-vars.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] `context-menu` component overrides now drive the menu's internal radius and padding vars. `ContextMenu.doc.mjs` has always documented `derived` entries mapping `borderRadius` → `--_dropdown-menu-radius` and `padding` → `--_dropdown-menu-padding`, but `derivedVarRegistry` had no `context-menu` key, so the mapping was dead: `components: {'context-menu': {base: {borderRadius: '12px'}}}` emitted `border-radius` alone and the menu kept reading its own `var(--_dropdown-menu-radius)`. The registry entry now matches the doc, as it already does for `dropdown-menu`.
+
+@cixzhang
--- a/.changeset/document-private-theming-vars.md
+++ b/.changeset/document-private-theming-vars.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[docs] Ten private (`--_*`) component theming vars are now documented in their owning component's `theming.vars[]`: `--_avatar-group-overlap`, `--_card-elevation`, `--_card-ring`, `--_codeblock-gutter-width`, `--_item-label-color`, `--_item-description-color`, `--_tab-indicator-bottom`, `--_tree-indent` (plus `--_dropdown-menu-radius`/`--_dropdown-menu-padding`, which Breadcrumbs sets on a child menu). They were declared in source and described nowhere, because the drift guard skipped the `--_` prefix outright.
+
+@cixzhang
--- a/packages/core/src/Avatar/Avatar.doc.mjs
+++ b/packages/core/src/Avatar/Avatar.doc.mjs
@@ -33,6 +33,9 @@ export const docs = {
       {className: 'astryx-avatar-status-dot', visualProps: ['variant']},
       {className: 'astryx-avatar-status-dot-glyph', visualProps: ['shape']},
     ],
+    vars: [
+      {name: '--_avatar-group-overlap', description: 'Negative inline offset applied to every avatar after the first when avatars are stacked in an AvatarGroup. Set from the group size; a more negative value tightens the stack.', default: 'set at runtime from the group avatar size (px)', private: true},
+    ],
   },
   description: 'Displays a user avatar with image, initials fallback, and optional status indicator.',
   props: [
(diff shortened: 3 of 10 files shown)


cixzhang:
Hmm arguably it's better for avatargroup to always enforce consistent sizing.

PR #4788, opened 2026-08-07
fix(core): coordinate concurrent useScrollLock instances

### LINKED ISSUE #4787: useScrollLock: concurrent overlays corrupt body scroll state when closed out of order
`useScrollLock` (`packages/core/src/hooks/useScrollLock.ts`) has no coordination between instances — each call independently snapshots `document.body.style` when its own effect mounts and restores that exact snapshot on cleanup. I confirmed this directly against the hook (two concurrent `renderHook(() => useScrollLock(true))` calls, unmounted out of order — see regression tests below).

The same hook is called, unmodified, by `Dialog`, `Lightbox`, and by `Drawer`/`BottomSheet` in `packages/lab` — since none of them wrap or alter the hook's behavior, the same bug applies to any pairing of these opened concurrently (e.g. a Lightbox opened from within a Dialog).

Failure sequence with two overlays open at once:

1. Overlay A opens, saves the true original body style, locks the body.
2. Overlay B opens on top, saves A's *already-locked* style as its "original."
3. A closes first (B still open) → A restores the true original → body unlocks while B is still open.
4. B closes → B restores what it saved in step 2 (`hidden`/`fixed`) → body is now stuck locked permanently, with nothing left open to undo it.

Regression tests (fail against `main`):

```tsx
it('stays locked while a second overlay is still open, even if the first closes out of order', () => {
  vi.spyOn(window, 'scrollTo').mockImplementation(() => {});

  const first = renderHook(() => useScrollLock(true));
  const second = renderHook(() => useScrollLock(true));

  first.unmount();

  expect(document.body.style.position).toBe('fixed');
  expect(document.body.style.overflow).toBe('hidden');

  second.unmount();
});

it('fully restores the body once every overlay has closed, regardless of close order', () => {
  vi.spyOn(window, 'scrollTo').mockImplementation(() => {});

  const first = renderHook(() => useScrollLock(true));
  const second = renderHook(() => useScrollLock(true));

  first.unmount();
  second.unmount();

  expect(document.body.style.position).toBe('');
  expect(document.body.style.overflow).toBe('');
  expect(document.body.style.top).toBe('');
  expect(document.body.style.left).toBe('');
  expect(document.body.style.right).toBe('');
});
```

I've opened #4788 with regression tests and a proposed fix.

alex-js-ltd:
Fixes #4787

`useScrollLock` had no coordination between instances — each call independently snapshotted `document.body.style` when its own effect mounted and restored that exact snapshot on cleanup. With two overlays open at once (e.g. a Lightbox opened from within a Dialog), closing them out of stacking order could unlock the body while one was still open, and could leave the body permanently stuck locked after both closed.

## Changes

* `useScrollLock` now coordinates locks with a shared counter: the original body state is captured only when the count goes 0→1, and only restored when it goes back to 0, regardless of which instance closes first.
* Added regression tests covering a single lock, and two concurrent locks closed both in and out of order.

## Test plan

* [x] `pnpm -F @astryxdesign/core test -- useScrollLock` — all pass, including the new concurrent-instance tests, which fail against the prior per-instance-snapshot implementation.
* [x] Full `Dialog`, `Lightbox`, `Drawer`, `BottomSheet` suites (the hook's real consumers) still pass — no regressions.
* [x] `pnpm check:changesets`

### DIFF
--- a/.changeset/scroll-lock-out-of-order-cleanup.md
+++ b/.changeset/scroll-lock-out-of-order-cleanup.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] `useScrollLock`: coordinate concurrent locks with a shared counter, so overlays closing out of order no longer unlock the body early or leave it stuck locked.
+
+@alex-js-ltd
--- a/packages/core/src/hooks/useScrollLock.test.ts
+++ b/packages/core/src/hooks/useScrollLock.test.ts
@@ -3,7 +3,7 @@
 /**
  * @file useScrollLock.test.ts
  * @input Uses vitest, @testing-library/react, useScrollLock hook
- * @output Unit tests for concurrent useScrollLock instances
+ * @output Unit tests for single and concurrent useScrollLock instances
  * @position Testing; validates useScrollLock.ts implementation
  *
  * SYNC: When useScrollLock.ts changes, update tests to match new behavior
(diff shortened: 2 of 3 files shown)


rubyycheung (approved):

cixzhang (approved):
Thanks this seems good for nested dialogs

PR #4791, opened 2026-08-07
feat(FormLayout): add defaultOptionality — mark only the exception

### LINKED ISSUE #4874: FormLayout: add `defaultOptionality` — form-level default so only the exception shows an optional/required indicator
## Summary

We have no form-level context that enforces the "show an indicator only for the exception" pattern — the equivalent of XDS's `optionalityBehavior`. Today every input decides on its own via `isOptional` / `isRequired`, so a form that is entirely required (or entirely optional) either repeats the same indicator on every field or relies on the author to hand-manage it.

We're missing a `Form` component, so one option is to introduce this as form context. But since optionality indicators are **visual only**, `FormLayout` is arguably the right home:

- Docs: https://astryx.atmeta.com/components/FormLayout
- Internal XDS reference: https://fburl.com/code/zu1ov0mx

## Why not just one prop

Because we have **both** `isOptional` and `isRequired` on the inputs (see `FieldLabel`, which renders `@astryx.field.optional` / `@astryx.field.required`), we can't collapse this to a single input-level prop. My API sense says these "should" be one prop, but `isOptional` is hard to discover on its own.

Better: put the default at the layout level as `defaultOptionality`, and let the input-level props act as **overrides**. An input that merely restates the layout default shows no indicator; only a deviation from the default gets one.

## Proposed API

```tsx
<FormLayout defaultOptionality="optional">
  <TextInput />            // Optional by default — no extra content
  <TextInput isRequired /> // Override — shows the required indicator
  <TextInput isOptional /> // Matches the default — no indicator
</FormLayout>

<FormLayout defaultOptionality="required">
  <TextInput />            // Required by default — no extra content
  <TextInput isRequired /> // Matches the default — no indicator
  <TextInput isOptional /> // Override — shows the optional indicator
</FormLayout>

<FormLayout /* defaultOptionality unset */>
  <TextInput />            // No indicator
  <TextInput isRequired /> // Shows the required indicator
  <TextInput isOptional /> // Shows the optional indicator
</FormLayout>
```

Semantics in one table:

| `defaultOptionality` | `<Input />` | `<Input isRequired />` | `<Input isOptional />` |
| --- | --- | --- | --- |
| `"optional"` | no indicator | **required** indicator | no indicator |
| `"required"` | no indicator | no indicator | **optional** indicator |
| unset (today's behavior) | no indicator | **required** indicator | **optional** indicator |

The unset row is exactly what ships today, so this is additive and backwards compatible.

## Open questions

1. **Where does it live** — `FormLayout` (visual-only, ships now) or a future `Form` component (semantic, doesn't exist yet)? If `FormLayout`, does a later `Form` re-expose or supersede it?
2. Does this go through `FormLayoutContext` (which currently carries only `direction`), and do all input-family components read it, or does `Field` / `FieldLabel` resolve it in one place?
3. Nesting: an inner `FormLayout` with its own `defaultOptionality` presumably shadows the outer one — confirm.
4. Does `defaultOptionality` affect anything beyond the visual indicator (e.g. `aria-required`, validation)? Proposal: **no** — it's presentation only, `isRequired` on the input still drives semantics.
5. Prop naming and value shape: `defaultOptionality="optional" | "required"` vs. a boolean.

## Notes

- Touches `packages/core/src/FormLayout/` (`FormLayout.tsx`, `FormLayoutContext.ts`) and `packages/core/src/Field/FieldLabel.tsx`, plus the input family that forwards `isOptional`/`isRequired` (TextInput, TextArea, Selector, ComplexSelector, MultiSelector, Typeahead, Tokenizer, Date/Time/DateRange/DateTime inputs, CheckboxInput, Slider, …).
- Please settle the API in this issue before building — check it against the wiki's **API Guidelines** and **API Conventions**.

freddymeta:
## What

Adds a form-level default so only the **exception** carries a visible optional/required indicator — the equivalent of XDS's `optionalityBehavior`, tracked in #4874.

- **`defaultOptionality?: 'optional' | 'required'`** on `FormLayout`, carried on `FormLayoutContext` (which already carries `direction`).
- Resolution lives in one place — `FieldLabel`, the single label used by ~20 inputs plus `CheckboxInput`/`Switch` — so the whole input family inherits it with no per-input plumbing.

```tsx
// Mostly-optional form: only the required field is marked.
<FormLayout defaultOptionality="optional">
  <TextInput label="Bio" />              {/* matches default → no indicator */}
  <TextInput label="Nickname" isOptional /> {/* matches default → no indicator */}
  <TextInput label="Email" isRequired />     {/* deviates → "Required" */}
</FormLayout>

// Mostly-required form: only the optional field is marked.
<FormLayout defaultOptionality="required">
  <TextInput label="Name" />             {/* matches default → no indicator */}
  <TextInput label="Nickname" isOptional /> {/* deviates → "Optional" */}
</FormLayout>
```

## Semantics

| `defaultOptionality` | `<Input />` | `<Input isRequired />` | `<Input isOptional />` |
| --- | --- | --- | --- |
| `"optional"` | no indicator | **required** indicator | no indicator |
| `"required"` | no indicator | no indicator | **optional** indicator |
| unset (today) | no indicator | **required** indicator | **optional** indicator |

The unset row is exactly what ships today, so this is **additive and backwards-compatible**. An inner `FormLayout` shadows an outer one.

## Presentation only

`defaultOptionality` changes only the **visible** indicator. Each control's `aria-required` is still driven by the input's own `isRequired`, so screen-reader semantics are unaffected. (Whether `defaultOptionality="required"` should *also* imply `aria-required` — since XDS inputs were required-by-default — is deferred as a separate semantic decision; see #4874 open question 4.)

## Scope change from the original PR

This PR originally proposed a per-field `requiredIndicator`/`optionalIndicator` + a `FieldProvider`, including a red `asterisk` mode. Per review (@cixzhang: the enum options overlap the existing booleans, and this belongs in the FormLayout context; @rubyycheung flagged the asterisk sizing), it's been re-scoped to the layout-context approach in #4874:

- Dropped `FieldProvider`, `FieldIndicatorContext`, `useFieldIndicators`.
- Dropped `requiredIndicator`/`optionalIndicator` and the **asterisk** entirely — there's no current use case for it.
- The app-wide policy now rides on the existing `FormLayoutContext` rather than a new provider.

## Changes

- `FormLayout`: new `defaultOptionality` prop + `FormOptionality` type on `FormLayoutContext`.
- `FieldLabel`: reads the context and suppresses the indicator that merely restates the default.
- Docs (`FormLayout.doc.mjs`, incl. dense + zh), Storybook stories (optional-default + required-default), and tests (context propagation, nesting/shadowing, and indicator behavior through `Field` for all three modes).

## Verification

- FormLayout + Field + FieldLabel suites: 74 tests pass ✓ (incl. 12 new)
- `tsc --noEmit` ✓, ESLint clean ✓, `pnpm check:repo` ✓ (sync / boundaries / changesets), docsite `generate` ✓
- Changeset added (patch).

Closes #4874.

### DIFF
--- a/.changeset/formlayout-default-optionality.md
+++ b/.changeset/formlayout-default-optionality.md
@@ -2,6 +2,6 @@
 '@astryxdesign/core': patch
 ---
 
-[feat] FormLayout: add `defaultOptionality` — set a form-wide default (`'optional'` or `'required'`) so only the exception carries a visible indicator. Under `'optional'` only `isRequired` fields show one; under `'required'` only `isOptional` fields do; a field that restates the default shows nothing. Presentation only — a control's `aria-required` is unchanged. Unset keeps today's per-field behavior.
+[feat] FormLayout: add `defaultOptionality` — set a form-wide default (`'optional'` or `'required'`) so only the exception carries a visible indicator. Under `'optional'` only `isRequired` fields show one; under `'required'` only `isOptional` fields do; a field that restates the default shows nothing. Under `'required'` the unmarked fields also expose `aria-required` so screen readers match what sighted users see — resolved on `aria-required` only, never the native `required` attribute. Unset keeps today's per-field behavior.
 
 @freddymeta
--- a/.changeset/formlayout-default-optionality.md
+++ b/.changeset/formlayout-default-optionality.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] FormLayout: add `defaultOptionality` — set a form-wide default (`'optional'` or `'required'`) so only the exception carries a visible indicator. Under `'optional'` only `isRequired` fields show one; under `'required'` only `isOptional` fields do; a field that restates the default shows nothing. Presentation only — a control's `aria-required` is unchanged. Unset keeps today's per-field behavior.
+
+@freddymeta
--- a/apps/storybook/stories/FormLayout.stories.tsx
+++ b/apps/storybook/stories/FormLayout.stories.tsx
@@ -22,6 +22,12 @@ const meta: Meta<typeof FormLayout> = {
       options: ['vertical', 'horizontal', 'horizontal-labels'],
       description: 'Direction of field arrangement',
     },
+    defaultOptionality: {
+      control: 'select',
+      options: [undefined, 'optional', 'required'],
+      description:
+        'Form-wide default so only the exception shows an optional/required indicator',
+    },
   },
 };
 
@@ -262,6 +268,52 @@ export const InDialog: Story = {
           </button>
         </div>
       </div>
+    );
+  },
+};
+
+// ─── Default optionality: mark only the exception ─────────────────────────
+
+export const DefaultOptionalityOptional: Story = {
+  name: 'Default Optionality — Optional',
+  render: () => {
+    const [bio, setBio] = useState('');
+    const [nickname, setNickname] = useState('');
+    const [email, setEmail] = useState('');
+    // Everything reads as optional; only the required field is marked.
+    return (
+      <FormLayout defaultOptionality="optional">
+        <TextInput label="Bio" value={bio} onChange={setBio} />
+        <TextInput
+          label="Nickname"
+          value={nickname}
+          onChange={setNickname}
+          isOptional
+        />
+        <TextInput label="Email" value={email} onChange={setEmail} isRequired />
+      </FormLayout>
+    );
+  },
+};
+
+export const DefaultOptionalityRequired: Story = {
+  name: 'Default Optionality — Required',
+  render: () => {
+    const [name, setName] = useState('');
+    const [email, setEmail] = useState('');
+    const [nickname, setNickname] = useState('');
+    // Everything reads as required; only the optional field is marked.
+    return (
+      <FormLayout defaultOptionality="required">
+        <TextInput label="Name" value={name} onChange={setName} />
+        <TextInput label="Email" value={email} onChange={setEmail} isRequired />
+        <TextInput
+          label="Nickname"
+          value={nickname}
+          onChange={setNickname}
+          isOptional
+        />
+      </FormLayout>
     );
   },
 };
--- a/packages/core/src/CheckboxInput/CheckboxInput.tsx
+++ b/packages/core/src/CheckboxInput/CheckboxInput.tsx
@@ -47,6 +47,7 @@ import {useTooltip} from '../Tooltip';
 import {mergeProps, mergeRefs} from '../utils';
 import {indicatorScope} from '../Indicator/indicator.markers.stylex';
 import {useIndicatorFocusRing} from '../hooks/useIndicatorFocusRing';
+import {useResolvedRequired} from '../hooks/useResolvedRequired';
 import {useIndicator} from '../Indicator';
 import {themeProps} from '../utils/themeProps';
 import {CheckboxListContext} from '../CheckboxList/CheckboxListContext';
@@ -281,6 +282,10 @@ export function CheckboxInput({
   const id = useId();
   const descriptionID = useId();
   const statusMessageID = useId();
+  // Announce the effective required state (form default included) while the
+  // native `required` stays bound to the explicit `isRequired` so a layout
+  // default never switches on browser validation.
+  const isEffectivelyRequired = useResolvedRequired({isRequired, isOptional});
 
   const [, startTransition] = useTransition();
   const [optimisticValue, setOptimisticValue] = useOptimistic(value);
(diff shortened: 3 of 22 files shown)


rubyycheung:
image
The asterisk is very tiny, how is the size defined?

packages/core/src/Field/Field.tsx
```diff
@@ -135,6 +139,17 @@ export interface FieldProps extends Omit<
...
+   * How the required indicator is displayed: `'text'`, `'asterisk'`, or
+   * `'none'`. Overrides any `FieldProvider` default. Does not affect the
+   * control's `aria-required`. @default 'text'
+   */
+  requiredIndicator?: RequiredIndicator;
+  /**
+   * How the optional indicator is displayed: `'text'` or `'none'`. Overrides
+   * any `FieldProvider` default. @default 'text'
+   */
+  optionalIndicator?: OptionalIndicator;
```
cixzhang: I think `none` is the same as setting `isOptional` to false here so we probably don't need the extra prop. It could make sense for the required indicator though since we only display that as text at the moment.

packages/core/src/Field/Field.tsx
```diff
@@ -135,6 +139,17 @@ export interface FieldProps extends Omit<
    * @default false
    */
   isRequired?: boolean;
+  /**
+   * How the required indicator is displayed: `'text'`, `'asterisk'`, or
+   * `'none'`. Overrides any `FieldProvider` default. Does not affect the
+   * control's `aria-required`. @default 'text'
+   */
+  requiredIndicator?: RequiredIndicator;
```
cixzhang: I think just text and asterisk is sufficient here since isRequired controls the display.

cixzhang (changes requested):
Mainly requesting since these options overlap the boolean props and we can probably separate them a bit better. Left suggestions

cixzhang:
Thinking about this + the recent conversation we had around the FormLayout, perhaps this should also be embedded in that context?

freddymeta:
Re-scoped this per the review + the FormLayout discussion. Summary of what changed since the CHANGES_REQUESTED pass:

- Moved the policy onto **`FormLayoutContext`** as `defaultOptionality` (`'optional' | 'required'`) instead of a new `FieldProvider` — @cixzhang's "should this be embedded in that context?" note. Dropped `FieldProvider` / `FieldIndicatorContext` / `useFieldIndicators`.
- Dropped the per-field `requiredIndicator`/`optionalIndicator` enums (they overlapped the existing `isRequired`/`isOptional` booleans) and removed the **asterisk** mode entirely — no current use case, and it addresses @rubyycheung's sizing flag by removal.
- Behavior is now "mark only the exception": a field that restates the form default renders no indicator; only a deviation does. Resolution lives in one place (`FieldLabel`), presentation only — `aria-required` is untouched.

The required-by-default *semantics* question (XDS inputs being required by default) is left as a follow-up — this PR is visual only. Closes #4874. PR body updated with the full table + verification.

cixzhang (dismissed):
Thanks — FormLayoutContext is the right home, and dropping the enums and the asterisk cleaned this up.

One thing before it lands. Under `defaultOptionality="required"` the indicators disappear, but each input still derives `aria-required` from its own `isRequired`, so a field you didn't mark stays `aria-required="false"`. Sighted users read "everything is required unless marked optional" and a screen reader hears required on only the few explicitly marked ones.

The inputs already read the context — they just need to resolve against it rather than their own prop alone:

```tsx
const isFieldRequired = isRequired ?? optionality === 'required';
aria-required={isFieldRequired}
```

Worth keeping that to `aria-required` and not the native `required` attribute, so a layout-level policy doesn't quietly turn on browser validation bubbles. Unless you think submit should block too?

[Reviewed by Robohands]

freddymeta:
Good catch — fixed. Under `defaultOptionality="required"` the indicator was suppressed for the unmarked majority but each input still derived `aria-required` from its own `isRequired`, so a field read as required by sighted users announced `aria-required="false"` to a screen reader.

Resolved the effective required state against the form default via a small shared `useResolvedRequired` hook, applied across every input that emits `aria-required` (TextInput, NumberInput, TimeInput, DateInput, DateRangeInput, DateTimeInput, Selector, MultiSelector, ComplexSelector, TextArea, RadioList group, CheckboxInput, Switch):

```ts
return !isOptional && (isRequired || defaultOptionality === 'required');
```

Scoped to `aria-required` only — the native `required` attribute stays bound to the explicit `isRequired`, so a layout-level default never switches on browser validation bubbles (as you flagged). `isOptional` still takes precedence, matching `FieldLabel`.

Added FormLayout tests covering the reviewer scenario: an unmarked input under `"required"` exposes `aria-required` without native `required`; an `isOptional` input doesn't; `"optional"` and no-layout stay unchanged; explicit `isRequired` still drives native `required`. Full input-suite + FormLayout suites green (1017 tests), tsc + prettier clean. Updated the changeset and the prop docs (which previously said "presentation only").

cixzhang (approved):
Thanks — this is right, and the shared hook is the correct shape. Checked every input in Chrome: unmarked fields announce required, native `required` never switches on.

[Reviewed by Robohands]

PR #4795, opened 2026-08-07
feat(core): add renderCopyButton render prop to CodeBlock

freddymeta:
## What this does

Adds a `renderCopyButton` render prop to `CodeBlock`, letting a consumer supply its own copy control while `CodeBlock` keeps ownership of everything that isn't the button's appearance.

## Why

`CodeBlock`'s copy control is a bare `<button>` with no tooltip, and there's no seam to restyle or replace it: `hasCopyButton` is a plain boolean, and the component exposes a single theme target on the root `<pre>`. A design system consuming Astryx that wants its own copy control (e.g. its standard icon button with a "Copy" tooltip — a tooltip is DOM, so it can't be added via theming or CSS) is forced to set `hasCopyButton={false}` and then re-implement placement, the clipboard write, the copied-state timer, and the copy announcement, plus position the replacement with structural CSS against internal DOM. That's the "reach for an unsupported structural selector" signal that a sanctioned seam is missing.

A paint/theme target is the wrong tier here (the need is a *different control*, not different paint on the same one), and an imperative `copy()` ref was considered previously and set aside. A render prop is the right shape — it mirrors the existing `renderOption` / `renderItem` / `renderToken` convention and hands back exactly the state a copy control needs.

## The seam

```tsx
renderCopyButton?: (props: {
  isCopied: boolean;   // reflects the block's copied-state timer
  copy: () => void;    // runs the clipboard write + confirmation flow
  label: string;       // localized, state-aware accessible label
}) => React.ReactNode
```

`CodeBlock` still owns:
- **Placement** — the control renders as the header's trailing child, or (headerless) in the floating corner, so the consumer supplies no positioning.
- **Clipboard write**, the **copied-state timer**, the **polite live-region announcement**, and `onCopy`.

The render prop only supplies the visual control. Ignored when `hasCopyButton` is `false`.

```tsx
<CodeBlock
  code={code}
  renderCopyButton={({isCopied, copy, label}) => (
    <IconButton
      label={label}
      icon={<Icon icon={isCopied ? 'check' : 'copy'} />}
      onClick={copy}
    />
  )}
/>
```

## Scope

Additive and backward-compatible — the built-in button path is unchanged when `renderCopyButton` is omitted. No new theme target, no DOM change to existing usage.

## Tests

Extended `CodeBlock.test.tsx` with a `renderCopyButton` block: custom control replaces the built-in button, `copy`/`isCopied`/`label` drive the block's clipboard + copied flow, the announcement and `onCopy` still fire, `hasCopyButton={false}` suppresses it, and the custom control stays out of the collapsible header's `role="button"`. `22/22` CodeBlock tests, core typecheck, storybook typecheck, docsite (324), and strict lint all green. Added a `CustomCopyButton` Storybook story.

### DIFF
--- a/.changeset/codeblock-render-copy-button.md
+++ b/.changeset/codeblock-render-copy-button.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] CodeBlock: add a `renderCopyButton` render prop to supply a custom copy control. The block keeps ownership of placement, the clipboard write, the copied-state timer, and the copy announcement — the render prop only provides the visual button, wired to the passed `copy`/`isCopied`/`label`. Ignored when `hasCopyButton` is `false`.
+
+@freddymeta
--- a/apps/storybook/stories/CodeBlock.stories.tsx
+++ b/apps/storybook/stories/CodeBlock.stories.tsx
@@ -2,6 +2,8 @@
 
 import type {Meta, StoryObj} from '@storybook/react';
 import {CodeBlock} from '@astryxdesign/core/CodeBlock';
+import {IconButton} from '@astryxdesign/core/IconButton';
+import {Icon} from '@astryxdesign/core/Icon';
 
 const meta: Meta<typeof CodeBlock> = {
   title: 'Core/CodeBlock',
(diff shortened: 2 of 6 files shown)


rubyycheung (approved):

freddymeta:
Closing — wrong shape for the actual need.

The goal was to **upstream CodeBlock's copy button** so consumers stop re-implementing it (bare `<button>`, no tooltip, not themeable). A `renderCopyButton` render prop does the opposite — it makes the consumer bring their own control and re-wire placement/clipboard, rather than Astryx shipping a good built-in one.

The right fix is to improve CodeBlock's **own** copy button — a default "Copy"/"Copied" tooltip and a theme target so it's restyleable — backed by an extracted `useClipboard` hook (which also de-duplicates the identical clipboard/copied-state/announce logic in Timestamp). Superseding this with that work; will link the spec issue here.

PR #4797, opened 2026-08-07
ci: add weekly full-suite RTL audit

nynexman4464:
> Stacked on #4796.

#4796 scopes `pr-rtl` to the components a PR touches, so nothing checks the rest of the library any more. But RTL breaks from a distance — a token, shared hook, or icon primitive can regress a component no PR edited.

`a11y-weekly.yml` already solved this for a11y and says so in its header. This adds the RTL twin, so the split is symmetric:

| | per-PR | scheduled |
|---|---|---|
| a11y | `pr-a11y` | `a11y-weekly.yml` |
| RTL | `pr-rtl` | **`rtl-weekly.yml`** |

Same shape as its a11y sibling: build once → audit with no `--filter` → summarize → upload → open-or-update one deduped tracking issue. Mondays 06:17 UTC (offset from a11y-weekly's 05:23 so the two Playwright sweeps don't contend), plus `workflow_dispatch`. Runners follow #4708's classes — `2-core-ubuntu-arm` for the audit, `ubuntu-slim` for the issue job.

**One difference from the a11y twin:** `rtl-audit.mjs` exits non-zero on *any* finding, so a failed step is normal here rather than a tripped gate. `weekly-rtl-summary.js` derives status from the report's contents instead, and the job goes red only on a crash — findings open a tracking issue while `pr-rtl` is still soft-gated.

## Test plan

`weekly-rtl-summary.js` run against all three status paths, using a real full-sweep report (1394 stories, generated from run [31122588220](https://github.com/facebook/astryx/actions/runs/31122588220)'s Storybook artifact):

- **clean** → `status=clean`, 125 components / 1394 stories, no tables
- **findings** → `status=findings`, `total_findings=3`, D1 + D5 + curated tables render (failures injected into the real report)
- **crashed** → `status=crashed` on a missing report

`rtl-weekly.yml` parses to 2 jobs / 9 audit steps; prettier-clean.

### DIFF
--- a/.github/scripts/weekly-rtl-summary.js
+++ b/.github/scripts/weekly-rtl-summary.js
@@ -1,0 +1,237 @@
+#!/usr/bin/env node
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+/**
+ * @description Summarizes a full-suite RTL audit report for the weekly RTL
+ *   scan workflow. Reads the JSON report produced by rtl-audit.mjs, writes a
+ *   markdown summary (used for both the GitHub step summary and the tracking
+ *   issue body), and emits step outputs (status, total_findings,
+ *   components_audited, stories_scanned) so the workflow can decide whether to
+ *   open-or-update the tracking issue and whether to fail the job.
+ * @input --report <file> --audit-outcome <success|failure> --summary-output <file>
+ *   [--github-output <file>]
+ * @output Markdown summary file + GitHub Actions step outputs
+ *
+ * Status semantics:
+ *   clean    — report present with zero findings
+ *   findings — report present, at least one D1/D5/curated finding
+ *   crashed  — no usable report (the audit died before writing one)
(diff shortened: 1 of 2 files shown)


nynexman4464:
Fixed after reading CONTRIBUTING.md and the current workflow conventions:

- **`runs-on` was wrong.** I had `ubuntu-latest` on both jobs, copied out of a stale local checkout. #4708 moved every workflow off that pool ("unblock runner backlog") and fresh `main` has zero left. Now follows that PR's workload classes: `2-core-ubuntu-arm` for the Playwright/build audit job, `ubuntu-slim` for the API-bound issue job — same split as `a11y-weekly.yml`.
- **Trimmed the comments.** The header block was ~30 lines of decision history; the rationale belongs here in the PR, not in the file.

Rebased onto the updated #4796. Summary-script output verified byte-identical before and after the trim.

PR #4810, opened 2026-08-07
fix(cli): resolve integration import specifiers against the package exports map

rubyycheung:
## The problem

Component lookups for integration packages built their import specifier by concatenating the package name and the component name:

```js
`${owner.package}/${componentName}`
```

That assumes every component is exported from a subpath named after itself. It usually is not — several components are commonly exported from one entry point named after the concept rather than the component. When that happens the specifier points at a subpath the package does not export, and the import the CLI suggests does not resolve.

Concretely, for a package whose `exports` map groups `EmployeeCard` and `EmployeeAvatar` under a single `./Employee` entry point, `component()` from `@astryxdesign/cli/api` returns:

```
import: '@acme/ui/Employee/EmployeeCard'   // a subpath the package does not export
```

rather than `@acme/ui/Employee`.

This matters most for anything generating docs from the API: a component detail page showing "how to import this" renders a copy-paste line that fails, for every component behind a shared entry point. The `astryx component` command has its own resolution path and is unaffected, which is why this is easy to miss — the CLI output looks right while the API's does not.

## The change

Resolve the specifier against the owning package's `exports` map instead, keyed on the directory the component's doc file sits in, falling back to the package root when that directory is not an exported subpath. This mirrors `resolveImportPathForPkg` in `apps/docsite/scripts/generate-data.mjs`, which already does exactly this.

Also stop overwriting a specifier a doc file states for itself. `import` was assigned after the `...docs` spread, so an authored value was always discarded — including the correct one, when a package had worked around this by writing the specifier out by hand.

Core resolution is unchanged.

## Behaviour change worth reviewing

An integration package that declares no matching `exports` subpath now gets the package root rather than a subpath it never exported. One existing assertion in `component-ownership.test.mjs` moves with this: the flat fixture declares no `exports`, so its specifier goes from `@test/meta/MetaAppShell` (which would not have resolved) to `@test/meta`.

## Tests

The ownership fixture gains a package `exports` map and the `__packageDir` that a resolved integration always carries, plus three new cases:

- a component whose directory is a shared entry point, resolving to that entry point's subpath
- a directory that is not an exported subpath, falling back to the package root
- a doc that states its own specifier, which is left alone

Verified on a clean install after rebasing onto `main`:

- `packages/cli/clients/cli/commands/component-ownership.test.mjs` — 16/16 pass
- full CLI suite — 193 files, 2679 tests, 0 failures

(An earlier revision of this description mentioned nine pre-existing failures. Those came from an unbuilt workspace in my local checkout and do not appear on a clean install.)

### DIFF
--- a/.changeset/cli-fix-integration-import-specifier.md
+++ b/.changeset/cli-fix-integration-import-specifier.md
@@ -1,0 +1,9 @@
+---
+'@astryxdesign/cli': patch
+---
+
+[fix] `component` built the import specifier for an integration component by joining the package name and the component name, which assumes every component is exported from a subpath named after itself. Components are commonly grouped behind a single entry point named after the concept, so the suggested import pointed at a subpath the package does not export and did not resolve.
+
+The specifier is now resolved against the owning package's `exports` map, keyed on the directory the component's doc file sits in, and falls back to the package root when that directory is not an exported subpath. A specifier a doc file states for itself is also no longer overwritten.
+
+@rubyycheung
--- a/packages/cli/api/component/_adapter.mjs
+++ b/packages/cli/api/component/_adapter.mjs
@@ -18,6 +18,8 @@
  * deduped, so each leaf stays a thin projection.
  */
 
+import * as fs from 'node:fs';
+import * as path from 'node:path';
 import {ERROR_CODES} from '../../foundation/response/error-codes.mjs';
 import {findCoreDir, discoverExternalPackages} from '../../foundation/fs/paths.mjs';
 import {
@@ -48,6 +50,7 @@ export {CORE_PACKAGE};
  * @property {any[]} [components]
  * @property {{description?: string}} [usage]
  * @property {any} [theming]
+ * @property {string} [import] set when the doc states its own import specifier
  */
 
 /**
@@ -66,6 +69,13 @@ export {CORE_PACKAGE};
  * @property {string|null} sourcePath
  * @property {string|undefined} issuesUrl
  * @property {import('../../foundation/integrations/integrations.mjs').LoadedIntegration|null} integration
+ */
+
+/**
+ * What ownership shaping needs from an owner. The core and legacy-external
+ * paths synthesize a bare `{package, sourcePath}` rather than resolving a full
+ * {@link ComponentOwner}, so everything past those two is optional here.
+ * @typedef {Partial<ComponentOwner> & {package: string, sourcePath: string|null}} OwnershipSubject
  */
 
 /**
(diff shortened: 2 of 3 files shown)


imdreamrunner (approved):
Approving — this is a real fix, and the premise checks out: `findIntegrationComponentDoc` scans the components dir recursively for `{Name}.doc.*`, so an integration component's directory genuinely need not share its name, and `${owner.package}/${componentName}` was emitting a subpath the package never exported. Resolving against `exports` keyed on the doc's directory mirrors `resolveImportPathForPkg` in the docsite generator, `__packageDir` is real production data from `loadIntegrations`, and the new tests exercise the path end to end. Verified locally on a clean install: ownership file 16/16, full CLI suite 198 files / 2727 tests green, no new typecheck or lint output on the changed files.

Four things for a follow-up — none of them block this.

**1. Legacy `pkg.astryx.docs` externals silently lose their subpath.** Those owners are synthesized as `{package, sourcePath}` with no `integration` ([component.mjs L162](https://github.com/facebook/astryx/blob/main/packages/cli/api/component/component.mjs#L162) and the unscoped fallback at the end), so `resolveIntegrationImportPath` hits `!packageDir` and returns the bare package name for *every* legacy external component — no `exports` consulted at all. Confirmed by calling `withOwnership` directly with a legacy-shaped owner: `@acme/ui/ToolbarSearch` → `@acme/ui`. That's a wider behaviour change than the description's "an integration package that declares no matching `exports` subpath", and nothing covers it. The package dir is derivable from `ext.docsDir`, so the legacy path could get the same treatment — but at minimum it deserves a line in the changeset and a test pinning whichever answer you want.

**2. A flat docs directory can match an unrelated subpath.** `path.basename(path.dirname(docPath))` on a doc that sits directly in the components root yields the root's own name. A package whose docs are flat (like the existing `MetaAppShell` fixture) and which exports `./components` gets `@acme/ui/components` — the docs dir, not an entry point that exports the component. Reproduced:

```
flat doc, package exports "./components" -> @acme/ui/components
```

A two-line guard fixes it: skip the lookup when `path.dirname(docPath) === owner.integration.components`.

**3. A wildcard `exports` map falls back to the root.** `{"./*": "./dist/*.js"}` makes `manifest.exports['./Toolbar']` undefined, so a subpath that *would* resolve becomes the package root. The docsite function has the same gap, so this is consistent rather than a regression — worth a `Object.keys(exports).some(...)` pass eventually if integrations start using wildcards.

**4. `import` isn't a field an integration author can legally write.** `ComponentBaseDoc` in `packages/cli/authoring/doctypes/component/type.ts` has no `import`, so a doc annotated `@type {ComponentDoc}` that states its own specifier fails excess-property checking — the escape hatch this PR now honours can't be used by a doc that typechecks. Add `import?: string` to the authoring type (with a note that it wins over resolution) and the third new test describes something an author can actually do. No core or lab doc sets `import` today, so "core resolution is unchanged" holds in practice either way.

Nit: `directory` is computed before the `!packageDir` early return, and its `: componentName` branch is unreachable — integration owners always carry `docPath`, and owners without one return early.

PR #4820, opened 2026-08-08
fix(ComplexSelector): keep the popup gap when opening above

### LINKED ISSUE #4803: [Bug] ComplexSelector popup has no offset when placement="above"
### Description

`ComplexSelector` renders its popup flush against the trigger when `placement="above"` — no clearance, so the popup and the field read as one merged block. With the default `placement="below"` the same selector gets the expected 4px.

Measured in a stock Vite + React app with `@astryxdesign/core@0.3.0`, no theme overrides, no `className`/`xstyle`, no StyleX build plugin:

| control | placement | clearance between trigger and popup | computed margins on the popup layer |
|---|---|---|---|
| `ComplexSelector` | `below` | **4px** | `margin-block-start: 4px` / `margin-block-end: 0px` |
| `ComplexSelector` | `above` | **0px** | `margin-block-start: 4px` / `margin-block-end: 0px` |
| `Popover` | `above` | **4px** | `margin-block-start: 4px` / `margin-block-end: 4px` |

The cause looks like a single-edge margin — correct for a popup that opens downward, ineffective for one that opens upward:

```ts
// packages/core/src/ComplexSelector/ComplexSelector.tsx
popover: {
  minWidth: 'anchor-size(width)',
  marginBlockStart: spacingVars['--spacing-1'],
},
```

`Popover` — the sibling built on the same `usePopover`/`useLayer` pair — sets both edges, names the style `gap`, and applies it at each of its render sites, so it spaces correctly whichever way it opens:

```ts
// packages/core/src/Popover/Popover.tsx
gap: {
  marginBlockStart: spacingVars['--spacing-1'],
  marginBlockEnd: spacingVars['--spacing-1'],
},
```

Three things suggest an oversight rather than an intentional difference: the same component clearly wants 4px (it gets it when opening downward), the sibling component gets the upward case right, and the computed margin is literally `0px` on the edge that matters. Adding `marginBlockEnd` to the `ComplexSelector` popover style would bring the two in line.

Not reachable from consumer code — the margin is on the popup layer, and the `xstyle` handed to `popover.render()` is hardcoded inside the component rather than derived from a prop.

### Reproduction

In a stock app (`npm create vite`, add `@astryxdesign/core`, import `reset.css` + `astryx.css`):

```tsx
<div style={{ minHeight: "100vh", display: "flex", flexDirection: "column", justifyContent: "space-between", padding: 48 }}>
  <ComplexSelector label="Below" value={1} placement="below" triggerLabel="below">
    {() => <div style={{ height: 120 }}>menu contents</div>}
  </ComplexSelector>

  <ComplexSelector label="Above" value={1} placement="above" triggerLabel="above">
    {() => <div style={{ height: 120 }}>menu contents</div>}
  </ComplexSelector>
</div>
```

1. Open the lower selector (`placement="above"`) so the popup opens upward.
2. Observe the popup's bottom edge touching the trigger's top edge.
3. Open the upper one (`placement="below"`) — the 4px gap is there.
4. Compare with `<Popover placement="above">`, which keeps its 4px.

Measured by comparing `getBoundingClientRect()` on the trigger against the popup layer (the `[popover]` ancestor of the content), and reading `getComputedStyle().marginBlockEnd` on that layer.

### Astryx Version

@astryxdesign/core@0.3.0

### Environment

Chromium 1234 (Playwright), macOS. Purely visual spacing — not assistive-technology specific.

AKnassa:
## What this does

When a ComplexSelector menu opens upward, it now keeps the same small gap above the field that it already keeps when opening downward.

## Why

With `placement="above"` the popup sat flush against the field, so the popup and the field read as one merged block (#4803). The same component already spaced correctly when opening downward, and Popover spaced correctly in both directions.

## What changed

- The popup layer now carries the 4px gap on both block edges, matching Popover, so the clearance holds whichever way it opens.
- A regression test locks the gap in on both edges.
- A changelog entry for the next `@astryxdesign/core` patch release.

## How to see it

Render a `ComplexSelector` with `placement="above"` near the bottom of the page and open it. Measured in a browser before and after: the space between the popup and the field goes from 0px to 4px, identical to `Popover` at the same placement. `placement="below"` is unchanged at 4px.

Fixes #4803

### DIFF
--- a/.changeset/complexselector-above-placement-gap.md
+++ b/.changeset/complexselector-above-placement-gap.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] ComplexSelector: the popup keeps its clearance when opening upward (`placement="above"`). The popup layer only set the gap on its block-start edge, which spaces a downward-opening popup but left an upward-opening one flush against the trigger. Both block edges now carry the gap, matching Popover.
+
+@AKnassa
--- a/packages/core/src/ComplexSelector/ComplexSelector.test.tsx
+++ b/packages/core/src/ComplexSelector/ComplexSelector.test.tsx
@@ -12,7 +12,23 @@
 import {describe, expect, it, vi} from 'vitest';
 import {render, screen, waitFor} from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
+import * as stylex from '@stylexjs/stylex';
+import {spacingVars} from '../theme/tokens.stylex';
 import {ComplexSelector} from './ComplexSelector';
+
+// StyleX emits one deterministic atomic class per property/value pair, so an
+// element carries a probe's class exactly when it has the same declaration.
+// The dev-mode debug class (contains "__") varies by source location and is
+// excluded from the comparison.
+const probe = stylex.create({
+  blockStartGap: {marginBlockStart: spacingVars['--spacing-1']},
+  blockEndGap: {marginBlockEnd: spacingVars['--spacing-1']},
+});
+
+function atomicClasses(style: (typeof probe)[keyof typeof probe]): string[] {
+  const {className = ''} = stylex.props(style);
+  return className.split(' ').filter(c => c !== '' && !c.includes('__'));
+}
 
 type FruitValue = {
   fruit: 'Apple' | 'Banana';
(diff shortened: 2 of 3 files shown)


AKnassa:
Closing this one — #4861 landed the same fix and closed #4803 while this was waiting for review.

The change there is identical to mine: both block edges on the `popover` style carrying `spacingVars['--spacing-1']`, so `placement="above"` keeps its clearance. Nothing left for this PR to do.

Thanks @HelloOjasMutreja — no action needed from maintainers here.

cixzhang:
Closing as already-landed: the same `marginBlockEnd` fix on the ComplexSelector popover is now on `main` via #4861, which was filed after this and merged first. Apologies — yours was first and correct.

Worth saying explicitly: your test here was the better of the two. Probing StyleX's deterministic atomic classes is more robust than asserting jsdom computed physical properties, which is what ended up merged. If you'd like to send that test on its own against `main`, I'd take it.

Issue #4804, opened 2026-08-07
[RFC] ComplexSelector's popup can't be styled without StyleX


nynexman4464:
### Problem Statement

There seems to be no escape hatch for styling `ComplexSelector`'s popup if you're not using StyleX.

The trigger is fine — `className` takes ordinary utility classes and works cleanly, no `!important`, no fighting the library. The popup has no equivalent: `contentXstyle` is `StyleXStyles`-typed, and the popup surface (the element painting background, border, radius, width) isn't reachable at all.

The surface is the one that matters, because it decides whether a popup looks like the rest of the app. In an app whose own menus are a 288px panel with a 1px border, the same menu built on `ComplexSelector` comes out 24px wider with no border — on a dark theme that reads as edgeless against the page.

What I tried:

| Route | Lands on | Worked without StyleX? |
|---|---|---|
| `className` / `style` | trigger container | yes |
| `xstyle` | trigger container | no (StyleX) |
| `contentXstyle` | content wrapper *inside* the surface | no (StyleX) |
| `defineTheme({components: {...}})` | elements with a stable `astryx-*` class | couldn't target it — see below |
| — | **popup surface** | couldn't reach it |

### Evidence of Demand

This hits any non-StyleX consumer styling this component, not just my app.

`astryx docs styling` presents StyleX, Tailwind, `className`, and styling-library aliases as four options, and points at `className` where StyleX isn't in play. The repo ships `example-vite`, `example-nextjs`, and `example-nextjs-tailwind` next to `example-nextjs-stylex`. So I expected `className` to be available wherever `xstyle` is, and the popup has neither.

Peer systems expose the popup surface directly: Radix passes `className` to `Popover.Content`, MUI has `slotProps.paper`, Chakra has `PopoverContent` as a styled slot, Ant has `overlayClassName`.

### Why Existing Components Don't Cover This

1. **`className`** — merges cleanly, but onto the trigger container. Checked the rendered DOM: the class lands on the trigger, and the popup surface's classes are StyleX-generated only.
[...]

PR #4821, opened 2026-08-08
feat(ComplexSelector): stable theme target for the popup surface

AKnassa:
## What this does

Gives the ComplexSelector popup a stable `astryx-complex-selector-popup` class, so apps that do not use StyleX can style the popup the way they already style the trigger.

Refs #4804 (route A).

## Why

The trigger takes `className`, but the popup surface rendered only hashed classes. A theme or plain stylesheet had no way to give the popup a border, a different background, a radius, or a width.

## What changed

- The popup content container carries the stable class and now paints the surface itself (`usePopover` with `hasSurface: false` plus the same background, radius, and shadow tokens, shared via an exported `popoverSurfaceStyles`). Styles aimed at the class land on the element that owns the popup's look, with nothing painting behind it. Rendered defaults are unchanged.
- `contentXstyle` merges after the surface styles, so it can now override background, border, radius, and shadow too. Prop docs updated.
- Because overrides genuinely replace the surface now, two niche cases behave differently on purpose: a translucent `contentXstyle` background no longer composites over an opaque default (the page shows through), and a themed width narrower than the trigger leaves the layer at the trigger's width for anchor positioning, so the strip beside the panel does not light-dismiss. No in-repo consumer relies on either; existing stories that widen the popup through `contentXstyle` render identically.
- The component's theming table lists the new target.
- New Storybook story "Themed popup surface" renders a flat, bordered, muted popup from `defineTheme` alone.
- Tests cover class placement, surface paint ownership (injected-CSS probe on both the popup and the dialog wrapper), the override order, and theme CSS emission.

Example:

```ts
defineTheme({
  components: {
    'complex-selector-popup': {
      base: {borderWidth: '1px', borderColor: 'var(--color-border)', inlineSize: '288px'},
    },
  },
});
```

## How to see it
4804-default-popup-open
4804-themed-popup-open
4804-themed-popup-zoom

Storybook, Core/ComplexSelector, "Themed popup surface": open the selector and the popup is a flat bordered panel. "Fruit ripeness selector" shows the defaults, unchanged.

## Validation

- ComplexSelector 8/8; popover-family collateral (Popover, DropdownMenu, Selector, MultiSelector, Typeahead, DateTimeInput) 406/406; full suite 9412/9414 with the 2 known pre-existing CLI case-sensitivity failures only
- check:repo, strict eslint, prettier, core and storybook typechecks clean
- Changeset included (patch)

### DIFF
--- a/.changeset/complex-selector-popup-theme-target.md
+++ b/.changeset/complex-selector-popup-theme-target.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] ComplexSelector: expose the popup surface as the `astryx-complex-selector-popup` theme target. The popup content container now paints the surface itself (same background, radius, and shadow tokens as before), so `defineTheme` components — or any plain stylesheet — can restyle the popup's background, border, radius, and width, and `contentXstyle` can override the surface styles too. Rendered defaults are unchanged.
+
+@AKnassa
--- a/apps/storybook/stories/ComplexSelector.stories.tsx
+++ b/apps/storybook/stories/ComplexSelector.stories.tsx
@@ -10,6 +10,7 @@ import {TextInput} from '@astryxdesign/core/TextInput';
 import {HStack, VStack} from '@astryxdesign/core/Layout';
 import {Token} from '@astryxdesign/core/Token';
 import {TreeList, type TreeListItemData} from '@astryxdesign/core/TreeList';
+import {Theme, defineTheme} from '@astryxdesign/core/theme';
 import {useGridFocus} from '@astryxdesign/core/hooks';
 import {
   borderVars,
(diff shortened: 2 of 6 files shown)


cixzhang:
Closing this — `main` already has it.

[#5009](https://github.com/facebook/astryx/pull/5009) added `surfaceTarget` to `usePopover` and `ComplexSelector` now passes `surfaceTarget: 'complex-selector-popup'` (`ComplexSelector.tsx:299`), with the target documented in `ComplexSelector.doc.mjs`. Same target name you chose, on the popover surface that actually paints — which is where this PR's review discussion was heading anyway.

Nothing wrong with the work; it got overtaken while the theme-target-placement question was being settled system-wide. Sorry for the wasted cycles.

PR #4823, opened 2026-08-08
fix(SideNav): anchor collapsed heading popover to trigger

### LINKED ISSUE #4789: [Bug] SideNavHeading popover mispositioned when collapsed with icon-only trigger
### Description

When `SideNav` is collapsed and `SideNavHeading` renders in icon-only mode
with a `menu` prop, opening the popover (click or hover) positions it at the
top-left of the viewport instead of anchored to the trigger element. In the
expanded state, the same `SideNavHeading` positions the popover correctly.

Root cause: in the collapsed + icon + menu branch, `collapsedSetRef` does
not include `popover.triggerRef`, so no `anchor-name` gets written to the
trigger element (see `Layer/anchorName.ts`). Without an anchor, the
popover's `position-anchor` has nothing to resolve against, and CSS Anchor
Positioning falls back to the fixed containing block instead of the
trigger.

Expected: the popover anchors to the collapsed icon-only trigger, same as
in the expanded state.

Actual: the popover appears near the top-left of the viewport, disconnected
from the trigger.

### Reproduction

2. Render a `SideNavHeading` with an `icon` and a `menu` prop.
3. Click the collapsed icon trigger to open the popover.
4. The popover renders at the top of the viewport instead of anchored near
   the trigger.
5. Set `isCollapsed` back to `false` and repeat — the popover now
   positions correctly.

Minimal repro:

```tsx
<SideNav collapsible={{ isCollapsed: true, hasButton: false }}>
  <SideNavHeading
    icon={<Avatar name="Jane Doe" />}
    heading="Jane Doe"
    menu={
      <NavHeadingMenu size="sm">
        <NavHeadingMenuItem label="Settings" href="/settings" />
        <NavHeadingMenuItem label="Log out" onClick={() => {}} />
      </NavHeadingMenu>
    }
  />
</SideNav>
```

Traced to `packages/core/src/SideNav/SideNavHeading.tsx`, in the
`isCollapsed && icon` branch:

```tsx
// Current — missing popover.triggerRef:
const collapsedSetRef = mergeRefs<HTMLElement>(collapsedItemRef, ref);

// Compare to the expanded-state setRef a few lines above, which does
// include it:
const setRef = mergeRefs<HTMLDivElement>(
  rootRef,
  setTriggerEl,
  ref,
);
```

Suggested fix, mirroring the existing pattern:

```tsx
  collapsedItemRef,
  ref,
);
```

### Astryx Version

@astryxdesign/core@0.3.0

### Environment

Browser

AKnassa:
Fixes #4789

### What this does

When the side nav is collapsed to icons only, the heading's account menu now opens right next to the icon that triggered it. Before, it appeared at the top-left corner of the screen, far away from where the user clicked.

### Why

The collapsed icon trigger never registered itself as the popover's anchor, so the popup had nothing to attach to and fell back to the corner of the viewport. The expanded heading already wires this correctly; the collapsed branch was missing one ref.

### What changed

- `SideNavHeading` collapsed mode: the icon-only trigger now passes `popover.triggerRef`, matching the expanded-state wiring
- Two new tests: the collapsed popup's `position-anchor` id must appear in the trigger's `anchor-name` list, plus the same check for the expanded state so both modes stay covered
- Changeset for a `@astryxdesign/core` patch

### How to see it

Render a collapsed SideNav whose heading has an icon and a menu, then open the menu from the collapsed icon:

```
<SideNav collapsible={{isCollapsed: true, hasButton: false}}>
  <SideNavHeading icon={...} heading="Product Name" menu={...} />
</SideNav>
```

Before, the popup renders at the viewport's top-left. After, it opens anchored below the icon.

### Screenshots
4789-after-popover-anchored
4789-before-popover-top-left

### DIFF
--- a/.changeset/sidenav-heading-collapsed-popover-anchor.md
+++ b/.changeset/sidenav-heading-collapsed-popover-anchor.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] SideNavHeading: the menu popover now anchors to the collapsed icon-only trigger instead of appearing at the top-left of the viewport. The collapsed trigger ref was missing the popover's `triggerRef`, so no `anchor-name` was written for the popover's CSS anchor positioning to resolve against.
+
+@AKnassa
--- a/packages/core/src/SideNav/SideNav.test.tsx
+++ b/packages/core/src/SideNav/SideNav.test.tsx
@@ -20,6 +20,7 @@ import {SideNavHeading} from './SideNavHeading';
 import {SideNavItem} from './SideNavItem';
 import {SideNavSection} from './SideNavSection';
 import {LinkProvider} from '../Link/LinkProvider';
+import {readAnchorNames} from '../Layer/anchorName';
 import {
   SideNavCollapseContext,
   type SideNavImperativeCollapseHandle,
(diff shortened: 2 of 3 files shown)


cixzhang:
Closing as already-landed — `main` now carries exactly this change (`menu ? popover.triggerRef : undefined` added to `collapsedSetRef`), merged via #4850 after yours was filed. Same story as #4819 and #4820; the ordering was our miss, not yours.

Thanks for the root-cause on the missing `anchor-name` — that was the right read.

Issue #4790, opened 2026-08-07
[Bug] SideNav: collapsing then reloading persists width 0, making the nav invisible and unrecoverable


elrancid:
### Description

A resizable + collapsible `SideNav` with `resizable.autoSaveId` becomes permanently invisible after the user collapses it once and reloads the page. The nav renders with `width: 0px` and no way to restore it from the UI — the collapsed state is not restored either, so the 48px icon rail with the expand button never appears.

Users who hit this cannot recover without manually clearing `localStorage` from devtools.

### Reproduction

```tsx
<AppShell
  sideNav={
    <SideNav
      collapsible={{ hasButton: false }}
      resizable={{
        defaultWidth: 260,
        minWidth: 150,
        maxWidth: 600,
        autoSaveId: 'app.sidenav.width',
      }}
      header={
        <SideNavHeading
          heading="My app"
          headingHref="/"
          headerEndContent={<SideNavCollapseButton />}
        />
      }
    >
      {/* nav items */}
    </SideNav>
  }
>
  <Outlet />
</AppShell>
```

1. Load the page — nav is 260px wide.
2. Click the collapse button — nav collapses to the 48px icon rail (expected).
3. Reload the page.
4. **The nav is now 0px wide and invisible.** It stays that way on every subsequent visit.

Measured across the three steps:

| Step | `localStorage['astryx-resizable:app.sidenav.width']` | rendered width |
| --- | --- | --- |
| after load | `260` | 260px |
| after collapse | `0` | 48px |
| after reload | `0` | **0px** |

The nav element is still in the DOM with its items — it just has zero width.

## Cause

Two behaviors in `useResizable.js` combine:

**1. Collapsing persists a width of 0.** `collapse()` sets the size to `0` and notifies through the same channel used for user resizes:

```js
preCollapseSizeRef.current = size;
setIsCollapsed(true);
setSize(0);
onCollapseChange?.(true);
onSizeChange?.(0);   // ← persisted as if the user had dragged the handle to 0
```

Collapsed is a transient UI state, but it is written to storage as a *width*.

**2. The restored value bypasses `clampSize`.** `loadPersistedSize()` accepts any finite number, and `0` passes the `typeof parsed === 'number'` check:

```js
const parsed = JSON.parse(raw);
if (typeof parsed === 'number') {
  return parsed;   // 0 is returned as-is
}
```

[...]

PR #4824, opened 2026-08-08
fix(SideNav): persist collapse state, not a zero width

AKnassa:
Fixes the invisible SideNav from #4790: a resizable and collapsible SideNav with `resizable.autoSaveId` persisted the collapsed width `0`, so a reload restored the resize hook as collapsed while SideNav rendered expanded layout at width 0.

Refs #4790

## What this does

- Persisted entries now store `{size, isCollapsed}` under the same key. `size` is always the expanded width, so the pre-collapse width survives a collapsed session.
- Legacy entries still load: a plain `0` (written by the old collapse path) reads as collapsed, so already-affected users recover without clearing localStorage; a plain width reads as width-only and never overrides `defaultIsCollapsed`. An object entry with an unusable size but an explicit `isCollapsed: true` is salvaged the same way as a plain `0`.
- `useResizable` gains `initialIsCollapsed` so a component that owns collapse state can seed the hook. SideNav seeds its uncontrolled collapse state from the persisted flag and passes its resolved state in, so the two owners agree at mount and no sync path (and no mount-time callback) exists.
- `UseResizableSingleConfig` gains `regions?: never` plus a null check in the overload body, so a multi-region config held in a variable resolves to the multi-region overload instead of silently matching the single-region one.

## Rollback

Reads are compatible in both directions; the width is the only thing at risk, never a crash or an invisible nav.

- New core reading old entries: plain numbers and `0` load as described above.
- Old core reading a new `{size, isCollapsed}` entry: `JSON.parse` yields an object, the `typeof parsed === 'number'` check rejects it, and the nav renders at the default width and the app's configured collapse default. The saved width and collapse state are silently lost. The old core's next write returns the entry to the legacy number format, which the new core reads fine.

## Out of scope

The four behavior fixes that previously rode along here (resize() notifying on un-collapse, the toggle double-fire fix, the live-state notification rewrite, toggle() no-oping when resizable but not collapsible) are split into #5118, stacked on this branch. #5073 tracks the root cause (two owners of one collapse state); this storage change is needed either way.

## Testing

- Hook tests cover the storage formats (current, legacy width, legacy `0`, salvage), hostile storage (quota and SecurityError throws, SSR guard, wrong-typed payloads), multi-region persistence and migration, and snaps and percentage defaults interacting with persistence.
- SideNav integration tests cover restore round trips, controlled mode, the `defaultIsCollapsed` interplay, StrictMode, a toggle-style controlled parent (no mount loop), and that restoring fires no callbacks.
- Rebased onto current main; these tests run together with the #4880 suite (220 passing across SideNav and Resizable on this branch, full core suite green).

SSR note: restoring a collapsed session changes the first client render, which React reports as a recoverable hydration mismatch, the same class as the existing persisted-width mismatch.

### DIFF
--- a/.changeset/sidenav-resizable-collapse-persistence.md
+++ b/.changeset/sidenav-resizable-collapse-persistence.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] SideNav: collapsing a resizable SideNav with `resizable.autoSaveId` no longer persists a width of 0, which made the nav restore as an invisible zero-width panel after a reload. Persisted entries now store `{size, isCollapsed}` under the same key, so a reload restores the collapsed icon rail and expanding returns to the previously saved width. Legacy entries still load: a plain `0` (written by the old collapse path) restores as collapsed, so already-affected users recover without clearing localStorage, and plain width entries keep their width without overriding `defaultIsCollapsed`. `useResizable` gains an `initialIsCollapsed` config so a component that owns collapse state (like SideNav) can seed the hook and the two can never disagree at mount. Rolling back: an older `@astryxdesign/core` reading a new-format entry ignores it and falls back to the default width and the app's configured collapse default — the saved width and collapse state are silently lost, but the nav stays usable, and the old version's next write returns the entry to the legacy number format. Note for SSR apps: restoring a collapsed session changes the first client render, which React reports as a recoverable hydration mismatch (the same class as the existing persisted-width mismatch).
+
+@AKnassa
--- a/packages/core/src/Resizable/Resizable.doc.mjs
+++ b/packages/core/src/Resizable/Resizable.doc.mjs
@@ -104,7 +104,7 @@ export const docs = {
         {
           name: 'autoSaveId',
           type: 'string',
-          description: 'Key for persisting sizes to localStorage.',
+          description: 'Key for persisting sizes and collapse state to localStorage.',
         },
       ],
     },
@@ -212,7 +212,7 @@ export const docsDense = {
         collapsedSize: 'px threshold triggering collapse during drag',
         snaps: 'px values to snap to during resize',
         shrinkOrder: 'cascade priority: lower number shrinks first',
-        autoSaveId: 'key for persisting sizes to localStorage',
+        autoSaveId: 'key for persisting sizes + collapse state to localStorage',
       },
     },
     {
--- a/packages/core/src/Resizable/useResizable.doc.mjs
+++ b/packages/core/src/Resizable/useResizable.doc.mjs
@@ -38,7 +38,12 @@ export const docs = {
     {
       name: 'autoSaveId',
       type: 'string',
-      description: 'Key for localStorage persistence of size across sessions.',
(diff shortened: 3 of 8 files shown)


cixzhang:
@AKnassa — this is the approach I want for [#4790](https://github.com/facebook/astryx/issues/4790). Seeding the hook from SideNav's own state via `initialIsCollapsed` is the right direction, it's the only version here that keeps `defaultIsCollapsed` working and restores the pre-collapse width, and 694 lines of hook tests on a file this fiddly is a real gift.

Three things before it lands:

1. **Please split it.** Only the storage format + seeding (~150 lines) is [#4790](https://github.com/facebook/astryx/issues/4790). Riding along are four separate changes, two of which alter when a shipped callback fires: `resize()` now emitting `onCollapseChange(false)` when it crosses out of collapsed, the double-fire fix, the `isCollapsedRef` live-ref rewrite for the drag closure, and `toggle()` no-oping when resizable-but-not-collapsible. Each is defensible; none is reviewable buried in 1287 lines under a title about persistence.

2. **Rebase.** This is based on Aug 8 and six commits have landed in `SideNav/` and `Resizable/` since — notably [#4880](https://github.com/facebook/astryx/pull/4880), which reworked `SideNavCollapseContext` and lets a `SideNavCollapseButton` render outside the SideNav (another consumer of the state you're re-sourcing). Git merges it clean, but your 409 new test lines and its 773 have never run together.

3. **Say what happens on rollback.** The new `{size, isCollapsed}` entry goes under the existing key; reads handle the old shape, but an older core reading the new shape silently loses the width.

Small thing: `preCollapseSizeRef.current = dragStartSizeRef.current || size` falls through to `size` when the ref is `0` — and `0` is exactly the value this PR is about.

Wider context, not a blocker: I've opened [#5073](https://github.com/facebook/astryx/issues/5073) to make `resizable` and `collapsible` mutually exclusive, since the root cause is that both own an `isCollapsed`. That's the durable fix for the class; your storage change is still needed either way, because the union fixes the divergence and not the `size: 0` encoding. Land this first.

AKnassa:
Done. This PR is now storage format + seeding only; the four behavior changes moved to #5118, stacked on this branch. Rebased onto current main, so the #4790 and #4880 suites run together (220 passing here, full core suite green), and rollback is documented in the body and changeset. The `|| size` catch was real; fixed in #5118 with a test pinning the out-and-back drag.

cixzhang:
Thanks for splitting and tightening this. The storage format and legacy-`0` recovery from this PR are already on `main` via #5075, with your contribution credited there.

This exact head still has SideNav and `useResizable` owning separate collapse state: when a controlled parent rejects a collapse request, the nav stays logically expanded but the resize hook sets its rendered width to `0px`. Current `main`'s single-owner controlled/uncontrolled implementation keeps it at `300px` in that case. Rebasing this head also conflicts across all seven implementation, documentation, and test files, and preserving current `main` leaves no code delta beyond this now-duplicate changeset.

No rework is needed here; I'm closing this as superseded by #5075. Thank you for the persistence encoding, legacy migration, and test coverage that carried forward.

[Reviewed by Robohands]

PR #4832, opened 2026-08-08
fix(typeahead): ignore Enter, Escape and arrows during IME composition

### LINKED ISSUE #4828: [Bug] PowerSearch: Enter during IME composition splits the last character into a separate filter
### Description

Typing with an IME (Korean / Japanese / Chinese) in PowerSearch's free-text **content search** field and pressing <kbd>Enter</kbd> to *commit the composition* is misinterpreted as "accept the highlighted suggestion". The pending syllable is then re-committed into the freshly-cleared input, so the last character the user was still composing ends up as its **own separate filter chip**.

**Expected:** the first <kbd>Enter</kbd> only commits the IME composition (no filter is created). A second <kbd>Enter</kbd> then creates one filter containing the whole phrase.

**Actual:** one <kbd>Enter</kbd> creates a filter *and* leaves the trailing syllable behind in the input, which becomes a second, spurious single-character filter.

This is the same class of bug already fixed for overlays and the chat composer in #3357 / #4280 — `BaseTypeahead` just never got the guard, so everything built on it (PowerSearch, Tokenizer, Typeahead) is still affected.

### Reproduction

Hosted Storybook: [**core-powersearch — With Content Search Field Key**](https://facebook.github.io/astryx/storybook/?path=/story/core-powersearch--with-content-search-field-key)

1. Switch the OS input source to a Korean IME (2-Set / 두벌식).
2. Click the PowerSearch input (`Type to search by title, or pick a field...`).
3. Type `한국어`. The trailing `어` is still in composition (underlined) — this is the normal IME state, the user has not "finished" the word.
4. Press <kbd>Enter</kbd> once, as you normally would to confirm the composition.

Observed:

- A filter chip `Title contains 한국어` is created immediately, and
- the input is **not** empty — it still holds `어`.
- Pressing <kbd>Enter</kbd> again (the natural next thing to do, since text is visibly still in the box) produces a second chip `Title contains 어`.

Final state — the last composed character has been split off into its own filter:

```
[ Title contains 한국어 ]  [ Title contains 어 ]
```

Same phrase typed in English (`hello` + <kbd>Enter</kbd>) correctly produces a single `Title contains hello` chip and an empty input, so the divergence is specific to composition input.

<details>
<summary>Scripted verification (Playwright + CDP <code>Input.imeSetComposition</code>) against the hosted Storybook</summary>

```
=== BASELINE: ASCII typing + Enter ===
before Enter: {"value":"hello","chips":[]}
after Enter : {"value":"","chips":["Title: contains hello"]}

=== KOREAN IME ===
input value while composing  : {"value":"한국어"}      // 어 is composing
after composing-Enter keydown: {"value":"","chips":["Title: contains 한국어"]}
after compositionend commit  : {"value":"어","chips":["Title: contains 한국어"]}   // ← residue
after next Enter             : {"value":"","chips":["Title: contains 한국어","Title: contains 어"]}
```
</details>

### Root cause

`BaseTypeahead`'s keydown handler treats every `Enter` as a selection, with no composition guard:

https://github.com/facebook/astryx/blob/077087f68d5e5fc12b39218a67dfa6a29fa04bcd/packages/core/src/Typeahead/BaseTypeahead.tsx#L666-L671

```ts
case 'Enter':
  e.preventDefault();
  if (highlightedIndex >= 0 && highlightedIndex < results.length) {
    handleSelect(results[highlightedIndex]);
  }
  break;
```

The browser event order when an IME commits on <kbd>Enter</kbd> is:

1. `keydown` — `key: 'Enter'`, **`isComposing: true`** (legacy `keyCode: 229`)
2. `compositionend` — the pending syllable is written into the input
3. `input`

Step 1 reaches the handler above. `handleSelect` fires `onChange(item)` and `setQuery('')`, clearing the controlled input ([`BaseTypeahead.tsx#L557-L574`](https://github.com/facebook/astryx/blob/077087f68d5e5fc12b39218a67dfa6a29fa04bcd/packages/core/src/Typeahead/BaseTypeahead.tsx#L557-L574)). The composition is still live on that DOM node, so step 2 then commits `어` into the now-empty field — producing the leftover that becomes the second chip.

In PowerSearch the selected item is the synthetic content-search entry built in [`usePowerSearchSource.ts#L185-L198`](https://github.com/facebook/astryx/blob/077087f68d5e5fc12b39218a67dfa6a29fa04bcd/packages/core/src/PowerSearch/usePowerSearchSource.ts#L185-L198), which is `unshift`ed to index 0 and therefore always the highlighted candidate — so the mis-fire is deterministic rather than dependent on what the user has highlighted.

### Suggested fix

The repo already has the exact helper this needs, added in #3357:

```ts
// packages/core/src/hooks/useFocusTrap.ts
export function isImeKeyEvent(event: {...}): boolean   // isComposing || keyCode === 229
```

Applying it in `BaseTypeahead.handleKeyDown` — ideally as an early return covering `Enter`, `Escape`, `ArrowUp`/`ArrowDown` and `Home`/`End`, since an IME candidate window uses all of those — fixes PowerSearch, Tokenizer and Typeahead in one place. `ChatComposerInput` already does this ([`ChatComposerInput.tsx#L560-L566`](https://github.com/facebook/astryx/blob/077087f68d5e5fc12b39218a67dfa6a29fa04bcd/packages/core/src/Chat/ChatComposerInput.tsx#L560-L566)).

One more unguarded site worth fixing in the same pass — Enter-to-save in the filter edit popover, which will swallow the composition-commit Enter while typing a CJK filter *value*:

https://github.com/facebook/astryx/blob/077087f68d5e5fc12b39218a67dfa6a29fa04bcd/packages/core/src/PowerSearch/PowerSearchEditPopover.tsx#L702-L714

Possibly related: #3343 (accessibility & keyboard-management program), #4759 (i18n grapheme-safe string handling).

### Astryx Version

`@astryxdesign/core@0.3.0` (reproduced on the hosted Storybook built from `main`)

### Environment

Chrome / Chromium, Korean 2-Set IME. The same event sequence applies to Japanese and Chinese IMEs, and to Safari and Firefox, so this is expected to be browser-independent.

yomybaby:
Resolves #4828.

## Problem

`BaseTypeahead` treated every `Enter` as "accept the highlighted result", with no composition guard. An IME uses `Enter` to commit a candidate, so that keystroke was consumed by the combobox instead.

In PowerSearch's free-text content search the damage is visible: the synthetic content-search item is `unshift`ed to index 0 and is therefore always highlighted, so the composing `Enter` deterministically filed a filter and cleared the query — while the composition was still live on the input. The browser then ended the composition and committed the trailing syllable into the now-empty field, where it became a second, single-character filter.

Typing `한국어` and pressing <kbd>Enter</kbd> left:

```
[ Title contains 한국어 ]  [ Title contains 어 ]
```

The same event sequence applies to Japanese and Chinese IMEs, and across browsers.

## Fix

Guard `BaseTypeahead`'s keydown handler with the existing `isImeKeyEvent` helper (added in #3357, already used by `useFocusTrap` and `Dialog`, and mirrored inline by `ChatComposerInput`). The guard sits after the consumer `onKeyDown` seam, so consumers still see composing keystrokes — only the combobox's own navigation and selection are suppressed.

It covers `Escape` and the arrows/`Home`/`End` as well as `Enter`, since an IME candidate window drives all of those: Escape cancels a composition, the arrows walk candidates.

`PowerSearchEditPopover`'s Enter-to-save / Escape-to-cancel had the same gap when typing a CJK filter *value*, so it gets the same guard.

Because it lands in `BaseTypeahead`, this fixes Typeahead and Tokenizer too, not just PowerSearch.

## Changes

| File | Change |
| --- | --- |
| `Typeahead/BaseTypeahead.tsx` | ignore keydowns that belong to an in-progress composition |
| `PowerSearch/PowerSearchEditPopover.tsx` | same guard on Enter-to-save / Escape-to-cancel |
| `hooks/useFocusTrap.ts` | `isImeKeyEvent` JSDoc updated — it is no longer Escape-only |

## Tests

Six new tests across `Typeahead.test.tsx`, `PowerSearch.test.tsx` and `PowerSearchEditPopover.test.tsx`, including the reported Korean scenario end-to-end at the PowerSearch level.

Each behavioural test was verified to **fail without the guard** and pass with it (the consumer-`onKeyDown`-forwarding test is a non-regression guard and passes either way):

```
✗ does not select the highlighted result when Enter commits a composition
✗ does not select on Enter reported only as keyCode 229
✗ keeps the dropdown open when Escape cancels a composition
✗ leaves the active option alone while the arrows walk IME candidates
✗ does not commit a content-search filter when Enter ends a composition
✗ does not save or cancel the filter mid-IME-composition
```

Full `packages/core` suite: **6000 passed (235 files)**. `pnpm lint` reports 0 errors, and `tsc --noEmit` on `packages/core` is clean.

## Verification

Reproduced against the hosted Storybook (`core-powersearch — With Content Search Field Key`) by driving a real composition through CDP `Input.imeSetComposition`, which reproduces the reported split exactly; see #4828 for the captured sequence. A jsdom test cannot run a genuine IME, so the unit tests assert the guard contract (`isComposing` / `keyCode 229`) rather than the browser's composition machinery — worth a manual pass with a Korean IME before merge.

### DIFF
--- a/.changeset/powersearch-and-typeahead-ignore-enter-escape-an-5j0u.md
+++ b/.changeset/powersearch-and-typeahead-ignore-enter-escape-an-5j0u.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] PowerSearch and Typeahead ignore Enter, Escape and arrow keys during IME composition, so a composing Enter no longer splits the trailing character into its own filter
+@yomybaby
--- a/packages/core/src/PowerSearch/PowerSearch.test.tsx
+++ b/packages/core/src/PowerSearch/PowerSearch.test.tsx
@@ -425,11 +425,51 @@ describe('PowerSearch', () => {
   });
 });
 
+describe('PowerSearch IME composition', () => {
+  const contentConfig: PowerSearchConfig = {
+    ...config,
+    contentSearchFieldKey: 'title',
+  };
+
+  it('does not commit a content-search filter when Enter ends a composition', async () => {
+    const onChange = vi.fn();
+    render(
+      <PowerSearch config={contentConfig} filters={[]} onChange={onChange} />,
+    );
+    const input = screen.getByRole('combobox');
+
+    // A Korean IME leaves the trailing `어` composing after the full phrase.
+    fireEvent.change(input, {target: {value: '한국어'}});
+    await waitFor(() => {
+      expect(input).toHaveAttribute('aria-expanded', 'true');
+    });
+
+    // Enter here belongs to the IME — it commits `어`, it does not submit.
+    fireEvent.keyDown(input, {key: 'Enter', isComposing: true});
+    expect(onChange).not.toHaveBeenCalled();
+
+    // The next Enter files one filter for the phrase, not a second for `어`.
+    fireEvent.keyDown(input, {key: 'Enter'});
+    expect(onChange).toHaveBeenCalledTimes(1);
+    expect(onChange.mock.calls[0][0]).toEqual([
+      expect.objectContaining({
+        field: 'title',
+        operator: 'contains',
+        value: {type: 'string', value: '한국어'},
+      }),
+    ]);
+  });
+});
 
 describe('PowerSearch statusVariant forwarding', () => {
   it('defaults to attached (status renders with data-variant="attached")', () => {
     const {container} = render(
-      <PowerSearch config={config} filters={[]} onChange={() => {}} status={{type: 'error', message: 'Required'}} />,
+      <PowerSearch
+        config={config}
+        filters={[]}
+        onChange={() => {}}
+        status={{type: 'error', message: 'Required'}}
+      />,
     );
     expect(container.querySelector('.astryx-field-status')).toHaveAttribute(
       'data-variant',
@@ -439,7 +479,13 @@ describe('PowerSearch statusVariant forwarding', () => {
 
   it('forwards statusVariant="detached" to the underlying Field status', () => {
     const {container} = render(
-      <PowerSearch config={config} filters={[]} onChange={() => {}} status={{type: 'error', message: 'Required'}} statusVariant="detached" />,
+      <PowerSearch
+        config={config}
+        filters={[]}
+        onChange={() => {}}
+        status={{type: 'error', message: 'Required'}}
+        statusVariant="detached"
+      />,
     );
     expect(container.querySelector('.astryx-field-status')).toHaveAttribute(
       'data-variant',
--- a/packages/core/src/PowerSearch/PowerSearchEditPopover.test.tsx
+++ b/packages/core/src/PowerSearch/PowerSearchEditPopover.test.tsx
@@ -283,4 +283,68 @@ describe('PowerSearch', () => {
     // onSave should NOT be called because the event was already consumed (defaultPrevented)
     expect(onSave).not.toHaveBeenCalled();
   });
+
+  it('does not save or cancel the filter mid-IME-composition', () => {
+    const onSave = vi.fn();
+    const onCancel = vi.fn();
+
+    function Harness() {
+      const internalConfig = useInternalConfig(testConfig);
+      return (
+        <PowerSearchEditPopover
+          config={internalConfig}
+          filter={{
+            field: 'status',
+            operator: 'is',
+            value: {type: 'string', value: '한국어'},
+          }}
+          mode="edit"
+          onSave={onSave}
+          onCancel={onCancel}
+        />
+      );
+    }
+
+    const {container} = render(<Harness />);
+    const input = container.querySelector('input');
+    expect(input).not.toBeNull();
+
+    // Enter commits an IME candidate and Escape cancels one — neither should
+    // act on the filter.
+    act(() => {
+      input?.dispatchEvent(
+        new KeyboardEvent('keydown', {
+          key: 'Enter',
+          bubbles: true,
+          cancelable: true,
+          isComposing: true,
+        }),
+      );
+    });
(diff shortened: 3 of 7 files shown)


cixzhang:
Thanks @yomybaby — this was a real bug and you diagnosed it precisely: the composing `keydown` fires *before* `compositionend`, so a bare Enter or Escape gets misread as a command mid-composition, and the `keyCode === 229` fallback matters because some IMEs and older Safari don't set `isComposing` in time. You also spotted the right generalisation — that the guard is not an Escape-only concern, since Enter commits a candidate and the arrows walk the candidate window.

Closing because that fix has since landed. [#4860](https://github.com/facebook/astryx/pull/4860) added the same `isImeKeyEvent` early return in the same two places this patches — `BaseTypeahead`'s `handleKeyDown` and `PowerSearchEditPopover`'s Enter/Escape handler — with equivalent Typeahead and PowerSearch tests. The broader contract you argued for in `useFocusTrap` went in too: [#4907](https://github.com/facebook/astryx/pull/4907) lifted the predicate out to `packages/core/src/utils/ime.ts` as the shared definition for all IME keys, and it is now used by the overlays and the editable inputs alike.

I'm sorry this sat unreviewed while that happened — you filed it three days before the PR that landed, and that ordering is on us, not on you. Please do check [#4860](https://github.com/facebook/astryx/pull/4860) and `utils/ime.ts` and reopen if you see a case yours covered that they miss; merge `main` in first if you do. Either way we'd like more of your bug reports.

[Discord](https://discord.com/invite/XnsUcFykEP) if you'd rather raise the next one with a person.

[Reviewed by Robohands]

PR #4837, opened 2026-08-08
feat(core): add DateTimeInput timeOptionInterval for a preset-time dropdown

### LINKED ISSUE #2727: [Bug] [bug-bash] DateTimeInput: add time-picker dropdown (preset times, e.g. 12am–11pm)
### Description

Feature request: when interacting with the time portion of a **DateTimeInput**, it'd be nice to have a dropdown / picker of preset times rather than only a free-form text input or stepper. Even something coarse like every hour (12 AM, 1 AM, …, 11 PM) — or whatever cadence is set by the time-increment prop (5 / 10 / 15 / 30 min) — would be a big usability win for "pick a meeting time"-style flows.

Today, picking a non-current time means typing or repeatedly stepping, which is slow vs. opening a list and clicking.

Suggested behavior:
- Clicking (or focusing) the time portion opens a popover with the list of valid times
- The list cadence respects the time-increment prop (so it works well after #2725 lands)
- Keyboard nav: arrow keys + Enter to pick; typing filters / jumps

### Reproduction

1. Navigate to https://xds-sandbox.vercel.app/components/DateTimeInput
2. Try to set a specific time — only stepper / typed entry is available

Expected: a dropdown of preset times (hourly at minimum, finer if `timeIncrement` is set).

### XDS Version

Whatever is currently deployed at xds-sandbox.vercel.app

### Environment

Chrome, macOS, https://xds-sandbox.vercel.app

AKnassa:
## What this does

Adds an optional dropdown of preset times to the time half of `DateTimeInput`. Set `timeOptionInterval` to `5`, `10`, `15`, `30`, or `60` and the time field offers a clickable list of times at that spacing. `60` gives the 12 AM to 11 PM list the issue asks for.

## Why

Today, setting a specific time means typing it or holding an arrow key. For "pick a meeting time" flows that is slow. Opening a list and clicking is the expected way to do this.

## What changed

- New optional `timeOptionInterval` prop on `DateTimeInput`. Leave it off and nothing about the component changes.
- When it is set, the time field opens a list on click or Alt+Down. Arrow keys move through it, Enter picks, Escape closes.
- Typing still works. The list follows along and highlights the closest match, so a time between two options can still be typed by hand.
- `min` and `max` trim the list on the boundary date, so only valid times are offered.
- Two new theming hooks for the popup: `date-time-input-time-listbox` and `date-time-input-time-option`.

## Why it is opt-in rather than always on

The issue suggests the list appear on focus and follow `timeIncrement`. Two things in the current component make that a breaking change rather than an addition:

- The time field's Up and Down arrows already step the value by `timeIncrement`, which is documented and tested. A list that is always open would have to take those keys. Instead the list opens on Alt+Down, and plain arrows keep stepping whenever the list is closed.
- Turning the time field into a combobox adds a second one to the field. `DateTimeInput.test.tsx` alone has 33 assertions using the singular `getByRole('combobox')` to reach the date input, and those throw the moment a second one exists. The same breakage would hit consumers silently on upgrade.

`timeIncrement` also defaults to `1`, which would mean a 1440 row list, and its type cannot express `60`. So the list gets its own prop. Making it default-on later is a one line change if that is preferred.

## How to see it

Storybook: **Core / DateTimeInput / With Time Options** and **With Hourly Time Options**.

## Notes for review

Four things surfaced while building that are worth a look:

- Both the calendar and the new list are `popover="auto"`, so opening the list evicts the calendar and fires its `onHide`, which refocused the date input. That is now guarded, and dismissing the calendar normally still restores date focus.
- The layer renders its children whether open or closed, so the options only mount while the list is open.
- A value carrying seconds (`14:00:00`) with `hasSeconds` off never matched a `14:00` option, so nothing showed as selected. Normalized before comparing.
- Added scroll-into-view for the active option. Verified in Chrome, since jsdom cannot measure it.

`TimeInput` deliberately does not get this. It has real consumers that render it inside a popover, and it renders a different tree inside an `InputGroup`. Happy to file that as a follow-up.

Closes #2727


kentonquatman (approved):
This looks good. Thank @AKnassa!

cixzhang:
Merged main in to clear the conflict. Three hunks: your `handleCalendarHide` gave way to main's `isFocusDetached()` check from the same `onHide` (it covers your case and also the one where focus lands inside the time list); main's IME guard now runs first in `handleTimeKeyDown`, ahead of the dropdown block, matching `handleDateKeyDown`; and the dep array and the theme-target list are unions of both sides. I also added the `:is(:disabled,[aria-disabled="true"])` cursor branch to `timeOption`, which [#5335](https://github.com/facebook/astryx/pull/5335) now lints for.

Not merging yet — `closes the list on Tab without blocking focus from leaving` fails on today's main, and your test is right. Details in a moment; nothing for you to do.

cixzhang:
Merged main in again — your Tab test is green now. It was right all along: `useFocusTrap` was cancelling Tab whenever an open layer had nothing focusable in it, without checking focus was even inside the layer. Fixed on main in [#5397](https://github.com/facebook/astryx/pull/5397). Nothing for you to change.

Worth saying, since you wrote it: your `case 'Tab': timePopover.hide()` is the right shape. It closes the list before the browser moves focus, which is what Selector and MultiSelector do — a real browser was already fine here and only the test was red.

I pushed one merge commit to your branch, main into yours, no conflicts and no edits to your files. Arming squash merge.

[Reviewed by Robohands]

cixzhang (approved):
Approving the refreshed head. kentonquatman approved the content on 2026-08-11 and nothing of yours has changed since — I'm extending that, not revisiting it. The only new commit is my merge of main, which had zero conflicts.

The Tab test is green: 135/135 in DateTimeInput.test.tsx. Verified in Chromium too — Tab out of the open time list moves focus to the next control and the list closes.

[Reviewed by Robohands]

PR #4839, opened 2026-08-09
fix(cli): keep scaffolded video sources playable

### LINKED ISSUE #4780: CLI: LightboxVideo scaffold replaces MP4 source with an SVG image data URI
## Description

Scaffolding the `LightboxVideo` template replaces its video source with the image placeholder:

```tsx
media={{
  src: 'data:image/svg+xml,...',
  type: 'video',
}}
```

The generated Lightbox therefore attempts to render SVG image data as a video and cannot play it.

`stripTemplateAssetRefs()` currently treats every demo-media URL as an image, regardless of file type.

This only affects template copying. The raw template shown by `astryx template LightboxVideo` and the docsite example remain valid because they do not pass through the scaffold transformation.

## Reproduction

```bash
npx @astryxdesign/cli@0.3.0 template LightboxVideo ./out --type block
```

Inspect `out/LightboxVideo.tsx`:

```tsx
src: 'data:image/svg+xml,...',
type: 'video',
```

## Expected behavior

A scaffolded video template should retain a valid video source. Possible approaches:

- use a video-specific self-contained placeholder;
- copy an appropriate video asset into the generated project; or
- otherwise make the generated example valid without depending on docsite-only assets.

The transformation should distinguish image and video extensions, with regression coverage for `.mp4`.

## Astryx Version

`@astryxdesign/cli@0.3.0` and current main/canary.

## Environment

Environment-independent source transformation. Reproduced with Node.js 22.17.1 on macOS.

AKnassa:
## What this does

Scaffolding a template that shows a video now gives you a video that actually plays. Before, the copied file pointed the video at a picture, so the player came up empty.

## Why

`astryx template LightboxVideo ./out --type block` is meant to hand you a working starting point, and it handed you a broken one. The copied file kept `type: 'video'` but its source was swapped for the image placeholder, so the Lightbox asked a `<video>` element to play SVG picture data. Nothing plays, and it is not obvious why.

Only the copy path was affected. `astryx template LightboxVideo` on its own, and the docsite example, read the original file and were always fine.

## What changed

- On copy, the placeholder is now chosen for each reference from the file type. Video files (`.mp4`, `.webm`, `.mov`, `.m4v`, `.ogv`) get a video placeholder; everything else keeps the image placeholder it has always had.
- Added that video placeholder: 640x360, about 1.6 seconds over 3 frames, silent, 2.8 kB, carried inside the file itself. It draws the same grey mark as the image placeholder with a play symbol in it. Like the image placeholder it needs no network and no asset folder, so a freshly scaffolded project just works.
- Added `placeholder-video.svg` alongside the existing `placeholder.svg` so the still frame is viewable in the diff rather than hidden inside an encoded string.
- The file type is read from the last part of the filename. `clip.min.mp4` used to be judged on `min` and treated as a picture, which is the same bug in a different shape, and `@2x` filenames slipped through untouched and would go missing in a scaffolded project. Both are latent rather than live: replaying the old and new rules over all 1328 template files gives byte-identical results.

## How it was checked

- Tests cover the five video file types, an upper-case extension, a file mixing a picture and a video, multi-part filenames, `@2x` names, and a full copy of the LightboxVideo block. The layout `expand` path is covered too, since it runs the same swap.
- The inline video cannot be reviewed by reading it, so its size, frame count, length, track layout and codec are all asserted against the decoded bytes, along with agreement between its internal duration headers. That set of tests caught two real defects in the first version of this PR: the clip was a single frame declaring 33ms while the notes claimed two seconds, and the encoder had written the media header duration in the wrong timescale.
- Every one of those assertions was mutation-checked: swapping the two placeholders fails 23 tests, forcing everything down the image branch fails 18, and restoring the old filename rule fails 3.

## How to see it

```
node packages/cli/clients/cli/bin/astryx.mjs template LightboxVideo ./out --type block
```

Open `out/LightboxVideo.tsx`. The `src` now begins with `data:video/mp4;base64,` instead of `data:image/svg+xml,`. Paste that whole value into a browser address bar to play it, or render the block and open the Lightbox.

The frame it shows is in the diff at `apps/docsite/public/template-assets/placeholder-video.svg`.

## Known and deliberately left alone

- A reference carrying a query string (`clip.mp4?v=2`) would leave the query glued to the end of the data URI. No template does this, and widening the pattern to swallow trailing characters is riskier than the bug.
- `astryx template <name>` (and its `--json` form) still prints the raw `/template-assets/...` path, because it shows the template as authored rather than as scaffolded. That is existing behaviour and a separate question from this fix.

Fixes #4780


cixzhang:
Thanks for chasing this down, @AKnassa — you found a real one. Scaffolding `LightboxVideo` really did hand people a `<video>` pointed at image data, and the diagnosis in your description is exactly right.

It's since been fixed a different way: [#4863](https://github.com/facebook/astryx/pull/4863) landed on 11 Aug and strips video sources to an empty `src` rather than substituting a placeholder clip, so a scaffolded project is honest about needing your own file. Same function, same bug — different remedy, and the embedded-clip approach was declined there. So I'm closing this as superseded rather than as anything you got wrong.

The two extras you caught — `.m4v` missing from the extension set, and `DEMO_ASSET_PATTERN` reading the extension off the first dot so `clip.min.mp4` is judged on `min` — I checked against main: all 150 `/template-assets/` references are single-dot, and none are `.m4v`, so neither one reaches anyone today. Worth knowing, not worth holding a PR open for.

Reopen any time if you think #4863 misses your case — merge main in first, since this branch has aged. And [Discord](https://discord.com/invite/XnsUcFykEP) is the fastest way to reach a person if you'd rather talk it through.

[Reviewed by Robohands]

PR #4840, opened 2026-08-09
docs(i18n): restore RTL dir best-practice in dense overlay

josephfarina:
## What

The `docsDense` best-practices list for `InternationalizationProvider` had 4 entries against the English 5. The RTL guidance (set the DOM `dir` attribute yourself) was missing from the compressed overlay, so `astryx component InternationalizationProvider --lang dense` silently dropped it: the CLI merges the overlay onto English at render time, and a short overlay list just replaces the longer English one.

This restores 1:1 parity:
- Adds a compressed version of the RTL `dir` best-practice at the same position and with the same `guidance: true` flag.
- Straightens two em dashes in the English prose (the `dir` best-practice and the `dir` prop description) to plain punctuation.

Meaning is unchanged; only the missing bullet and typography.

## Validation
- `astryx component InternationalizationProvider --lang dense` now shows 5 best-practice bullets, no duplicate `(required)`

Night Watch — Doc Reviewer

### DIFF
--- a/packages/core/src/i18n/InternationalizationProvider.doc.mjs
+++ b/packages/core/src/i18n/InternationalizationProvider.doc.mjs
@@ -40,7 +40,7 @@ export const docs = {
       {
         guidance: true,
         description:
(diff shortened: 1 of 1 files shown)


cixzhang (approved):
Approved — docs-only, no behavior change, CI green.

PR #4844, opened 2026-08-09
fix(hooks): cycle useTreeFocus typeahead on repeated letters

### LINKED ISSUE #4843: [Bug] TreeList typeahead stalls when the same letter is pressed twice
### Description

Pressing the same letter twice in quick succession stops typeahead instead of cycling to the next match.

`useTreeFocus.runTypeahead` appends every keystroke to its buffer, so a second `a` builds the query `"aa"`, which matches no label — focus never moves, and the buffer stays poisoned for the rest of the 500 ms window.

The shared `useTypeahead` (used by every menu) already collapses a repeat back to a single character and calls it APG behaviour; `useTreeFocus` has its own copy that has drifted. The same expression also causes two smaller issues: with no treeitem focused the *last* match wins instead of the first, and a refined multi-character query skips the current item even when it still matches.

### Reproduction

[FullyExpanded story](https://facebook.github.io/astryx/storybook/index.html?path=/story/core-treelist--fully-expanded) → click a row → press <kbd>p</kbd> twice **quickly**: focus stops on `public`, `package.json` is unreachable. Press <kbd>p</kbd>, pause ~1 s, press again: it advances correctly. Same keystrokes, different result, decided by timing.

Not a deliberate repeat-guard — rapid <kbd>p</kbd><kbd>a</kbd> moves twice, but rapid <kbd>p</kbd><kbd>p</kbd><kbd>a</kbd> stalls, so a repeat breaks the following unrelated letter too.

Failing against `main`:

```tsx
render(<Tree collapsed={[Apple, Apricot, Avocado]} />);
screen.getByTestId('apple').focus();

fireEvent.keyDown(tree, {key: 'a'});
expect(screen.getByTestId('apricot')).toHaveFocus(); // ✅

fireEvent.keyDown(tree, {key: 'a'});
expect(screen.getByTestId('avocado')).toHaveFocus(); // ❌ received: apricot
```

### Astryx Version

`@astryxdesign/core@0.3.0` (`main` @ `077087f68d`)

### Environment

Chrome / macOS; also reproduces in jsdom

---

Happy to open a PR — I have a fix ready that reuses `useTypeahead`'s buffer and search-order rules, with regression tests covering all three symptoms.

alex-js-ltd:
Resolves #4843.

## Problem

`useTreeFocus` has its own typeahead implementation, which has drifted from the shared `useTypeahead` behavior.

This causes three related issues:

* repeated letters build queries like `"aa"` instead of cycling through matches;
* with no focused treeitem, the first match is searched last;
* multi-character queries skip the current item even when it still matches.

For example, pressing `p` twice quickly in the `FullyExpanded` TreeList story stalls on `public` instead of advancing to `package.json`.

## Fix

Align `useTreeFocus` with the typeahead semantics being established in #3797:

* repeated characters collapse back to a single-character query;
* search ordering now distinguishes between cycling a single character and refining a multi-character query.

No public API changes.

## Tests

Added regression coverage for all three cases:

* repeated-letter cycling;
* searching from the first item when nothing is focused;
* keeping focus when the current item still matches a refined query.

Before the fix: **3 failing, 13 passing**
After the fix: **16 passing**

Also verified the full `ui` test suite (**6472/6472 passing**), Prettier, ESLint, `core` typecheck, and `check:changesets`.

## Reproduce manually

Open the [[FullyExpanded story](https://facebook.github.io/astryx/storybook/index.html?path=/story/core-treelist--fully-expanded)](https://facebook.github.io/astryx/storybook/index.html?path=/story/core-treelist--fully-expanded), click a row, then press `p` twice quickly.

Before: focus stops on `public`.
After: focus advances to `package.json`.

### DIFF
--- a/.changeset/tree-typeahead-cycling.md
+++ b/.changeset/tree-typeahead-cycling.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] TreeList typeahead now cycles through same-letter matches instead of stalling, and searches from the top when no treeitem is focused.
+@alex-js-ltd
--- a/packages/core/src/hooks/useTreeFocus.test.tsx
+++ b/packages/core/src/hooks/useTreeFocus.test.tsx
@@ -261,4 +261,54 @@ describe('useTreeFocus activation + typeahead', () => {
     fireEvent.keyDown(tree, {key: 'c'});
     expect(screen.getByTestId('c')).toHaveFocus();
   });
+
+  it('typeahead cycles through same-letter matches on repeated presses', () => {
+    const nodes: Node[] = [
+      {id: 'apple', label: 'Apple', level: 1},
+      {id: 'apricot', label: 'Apricot', level: 1},
+      {id: 'avocado', label: 'Avocado', level: 1},
+    ];
+    render(<Tree collapsed={nodes} />);
+    const tree = screen.getByRole('tree');
+    screen.getByTestId('apple').focus();
+
+    fireEvent.keyDown(tree, {key: 'a'});
+    expect(screen.getByTestId('apricot')).toHaveFocus();
+
+    // A repeat must cycle, not extend the query to "aa" (as useTypeahead does).
+    fireEvent.keyDown(tree, {key: 'a'});
+    expect(screen.getByTestId('avocado')).toHaveFocus();
+  });
+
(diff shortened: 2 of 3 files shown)


cixzhang (approved):
Interesting. Maybe there's something we can converge with the typeahead hook. This is good for now!

PR #4847, opened 2026-08-09
fix(docsite): render the component anatomy table

AKnassa:
## What this does

Component pages on the doc site now show an **Anatomy** table — the named parts a
component is made of, and what each one is for. Avatar, for example, now lists Photo,
Initials, Default icon, and Status dot.

## Why

That information was already written down and already reaching the page — 55 components
describe their anatomy, 236 parts in total. The table that displays it had been built too.
It just was never connected to anything, so the doc site quietly showed none of it while
the CLI (`astryx component <Name>`) showed it all.

So this isn't new content or a new feature. It's switching on writing that authors have
been doing all along.

## What changed
anatomy-avatar
anatomy-avatargroup-required

- The component **Overview** tab now renders the Anatomy table when a component describes
  its parts, and shows nothing when it doesn't.
- Anatomy sits **above** Best practices, the same order the CLI already uses.
- The Anatomy heading now matches the size of the headings around it (Usage, Best
  practices). It had been set smaller.
- A new test fails if any piece of the component-detail page ever becomes disconnected
  again, which is the exact way this was missed.

No changes to any published package, and no change to what authors write.

<details>
<summary>For reviewers: how it was never noticed</summary>

`Anatomy.tsx` arrived with the doc site in f45f46576 and was never imported —
`git log -S "Anatomy"` on `ComponentDetailClient.tsx` returns nothing, so it was never
wired rather than deliberately unwired. The data path was always intact:
`generate-data.mjs` forwards the whole `usage` block, and the emitted registry type
declares `anatomy`.

The new test resolves import specifiers the way the bundler does, so a package path like
`@astryxdesign/core/Section` can't be mistaken for a local file. It's scoped to
`component-detail/` on purpose — that folder holds only internal building blocks, with no
route entry points and no script-consumed modules, so an unimported file there is always
a bug.

Two notes on scope. `ThemeShowcaseTile.tsx` is a second orphan from the same commit; it's
left alone because where it belongs on `/themes` is a design question, not a wiring bug.
And the docsite typecheck reports one pre-existing error in the generated registry
(`DerivedVar."replaces"`) — verified identical with these changes reverted.
</details>

## How to see it

```
pnpm build
pnpm -F @astryxdesign/docsite dev
```

- <http://localhost:3000/components/Avatar> — Anatomy table between Usage and Best practices
- <http://localhost:3000/components/AvatarGroup> — same, with a `required` badge on a row
- <http://localhost:3000/components/Divider> — no anatomy authored, so no section appears

Tests: `pnpm -F @astryxdesign/docsite test` — 21 files, 326 tests.


kentonquatman (dismissed):
Please wrap the table in a card so that it matches the styling of all other tables on this page. The "required" badge also seems oddly positioned in the screenshot. Please update so that it's an inline bold text element as part of the description (eg. "**Required:** Avatar elements that from...").

apps/docsite/src/__tests__/component-detail-wiring.test.ts
```diff
@@ -0,0 +1,115 @@
...
+
+describe('component detail wiring', () => {
+  it('imports every module in component-detail/ from somewhere', () => {
+    const detailModules = walk(DETAIL_DIR).filter(
+      file => !file.includes('.test.'),
+    );
+    // Guard against a vacuous pass if the directory is ever moved or renamed.
+    expect(detailModules.length).toBeGreaterThan(5);
+
+    const imported = collectImportedFiles(walk(SRC_DIR));
```
cixzhang, line 88: A unit test on an orphan silences this. Might need the `.test.` filter from :83.

apps/docsite/src/components/component-detail/Anatomy.tsx
```diff
@@ -21,44 +22,40 @@ export function Anatomy({elements}: AnatomyProps) {
...
+        {/* Carded like the other tables on this page (BestPracticesBlock, the
+            /docs TableBlock) rather than sitting bare on the page background. */}
+        <Card variant="default">
+          <Table
+            data={data}
+            columns={[
+              {
+                key: 'name',
+                header: 'Element',
+                width: pixel(140),
```
cixzhang, line 43: 140 hard-breaks `Optional/Requ ired indicator` on the Field page. Might need a little more width.

cixzhang (approved):
Thanks — both of Kenton's asks are in; approve. Description still shows the badge column. Red Vercel looks like the `DerivedVar` error you flagged, since fixed on main — try a merge.

[Reviewed by Robohands]

cixzhang:
Merged — I merged `main` into your branch myself rather than sending it back to you, since the staleness was our latency, not anything wrong with your work. Thank you, @AKnassa, for wiring up the Anatomy renderer the system already had and never actually rendered: 64 component pages now show their parts instead of hiding them, with `Required:` reading inline the way Kenton asked. I'm sorry this sat for two weeks before we got to it.

[Reviewed by Robohands]

PR #4853, opened 2026-08-10
fix(SideNav): reconcile persisted-collapsed state on mount for resizable nav

### LINKED ISSUE #4790: [Bug] SideNav: collapsing then reloading persists width 0, making the nav invisible and unrecoverable

HelloOjasMutreja:
## Summary

Closes #4790.

A resizable + collapsible `SideNav` with `resizable.autoSaveId` becomes invisible (0px wide, and stuck that way) after collapsing once and reloading.

## Root cause

The reported diagnosis pointed at `useResizable`'s persistence path, but the current source already clamps the restored size and already seeds its own `isCollapsed` as `true` when it restores a persisted width of `0`:

```tsx
const [isCollapsed, setIsCollapsed] = useState(
  () => persisted === 0 && collapsible,
);
```

The actual remaining bug is one level up, in `SideNav.tsx`: SideNav's own uncontrolled `collapsed` state (the boolean that decides whether to render the compact icon rail vs. the full layout) is a **separate** `useState` call with no visibility into `useResizable`'s initial `isCollapsed` value:

```tsx
const [uncontrolledCollapsed, setUncontrolledCollapsed] = useState(defaultIsCollapsed); // always false by default
...
const resizableHook = useResizable({ ... }); // seeds isCollapsed = true internally
```

`onCollapseChange` only fires on *changes* (drag past the threshold, or `toggle()`), never for the hooks' own independently-computed initial states. So on a persisted-collapsed reload: `resizableHook.isCollapsed` is `true` (forcing `resizableHook.size` to `0`), but SideNav's own `collapsed` stays `false` — so the nav renders in **expanded layout** (no compact icon rail, no visible expand button) with its width forced to `0` by `resizableNavStyle`. That combination is exactly "invisible and unrecoverable from the UI" from the report.

## Fix

A mount-only `useLayoutEffect` reconciles SideNav's own uncontrolled `collapsed` state with `resizableHook.isCollapsed` right after the hooks are composed, before paint (avoiding a visible flash). Only applies when uncontrolled — a controlled `collapsible.isCollapsed` remains the consumer's own source of truth.

## Test notes

Added a regression test that seeds `localStorage` with a persisted width of `0` before rendering, then asserts the nav renders with the "Expand sidebar" button (i.e., the actual collapsed icon-rail layout) rather than the resize handle of an expanded-but-0px nav. Verified it fails against the pre-fix code (the "Expand sidebar" button doesn't render — the nav renders expanded instead) and passes with the fix.

## Test plan

- [x] `vitest run packages/core/src/SideNav/SideNav.test.tsx` — 101/101 passing
- [x] `vitest run packages/core/src/AppShell` — 42/42 passing (SideNav consumer)
- [x] `eslint` clean (added the repo's established `@eslint-react/set-state-in-effect` suppression for this intentional one-time reconciliation, matching the pattern in `useOverflow.ts`/`MobileNav.tsx`)

### DIFF
--- a/.changeset/sidenav-persisted-collapsed-width.md
+++ b/.changeset/sidenav-persisted-collapsed-width.md
@@ -2,6 +2,6 @@
 '@astryxdesign/core': patch
 ---
 
-[fix] SideNav: a collapsed `resizable` nav with `resizable.autoSaveId` no longer renders invisible (0px, expanded layout) after a reload. `useResizable` correctly seeds its own `isCollapsed` as `true` when it restores a persisted width of `0`, but `SideNav`'s own uncontrolled collapsed state is a separate hook with no way to see that at initialization — left unreconciled, the nav rendered in expanded layout with its width forced to `0` instead of the compact icon rail. A mount-only effect now reconciles the two.
+[fix] SideNav: a collapsed `resizable` nav with `resizable.autoSaveId` no longer renders invisible (0px, expanded layout) after a reload. `useResizable` correctly seeds its own `isCollapsed` as `true` when it restores a persisted width of `0`, but `SideNav` tracked its own, separately-initialized collapsed state instead of reading `useResizable`'s — left unreconciled, the nav rendered in expanded layout with its width forced to `0` instead of the compact icon rail. When resizable and uncontrolled, `SideNav` now derives its collapsed state directly from the resize hook instead of duplicating it.
 
 @HelloOjasMutreja
--- a/.changeset/sidenav-persisted-collapsed-width.md
+++ b/.changeset/sidenav-persisted-collapsed-width.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] SideNav: a collapsed `resizable` nav with `resizable.autoSaveId` no longer renders invisible (0px, expanded layout) after a reload. `useResizable` correctly seeds its own `isCollapsed` as `true` when it restores a persisted width of `0`, but `SideNav`'s own uncontrolled collapsed state is a separate hook with no way to see that at initialization — left unreconciled, the nav rendered in expanded layout with its width forced to `0` instead of the compact icon rail. A mount-only effect now reconciles the two.
+
+@HelloOjasMutreja
--- a/packages/core/src/SideNav/SideNav.test.tsx
+++ b/packages/core/src/SideNav/SideNav.test.tsx
@@ -969,6 +969,41 @@ describe('SideNav resizable', () => {
     expect(handleWidthChange).toHaveBeenCalledWith(expect.any(Number));
   });
 
+  it('restores as collapsed instead of 0px-expanded when a persisted width of 0 is loaded on mount', () => {
+    const key = 'astryx-resizable:sidenav-collapsed-reload-test';
+    localStorage.setItem(key, '0');
+    try {
+      render(
+        <SideNav
+          collapsible={{}}
+          resizable={{
+            defaultWidth: 260,
+            minWidth: 180,
+            maxWidth: 480,
+            autoSaveId: 'sidenav-collapsed-reload-test',
+          }}>
+          Content
+        </SideNav>,
+      );
+      // useResizable seeds its own isCollapsed as true when it restores a
+      // persisted width of 0, but SideNav's own uncontrolled `collapsed`
+      // state is a separate hook call with no way to see that at
+      // useState-init time. Unreconciled, the nav renders expanded (no
+      // inline width override) with resizableHook.size forced to 0 by the
+      // hook's own isCollapsed — a 0px-wide "expanded" nav, not the compact
+      // icon rail. The collapse button's presence and label are the
+      // observable signal of which layout actually rendered.
+      expect(
+        screen.getByRole('button', {name: 'Expand sidebar'}),
+      ).toBeInTheDocument();
+      expect(
+        screen.queryByRole('button', {name: 'Collapse sidebar'}),
+      ).not.toBeInTheDocument();
+    } finally {
+      localStorage.removeItem(key);
+    }
+  });
+
   it('respects defaultWidth', () => {
     render(<SideNav resizable={{defaultWidth: 300}}>Content</SideNav>);
     const nav = screen.getByRole('navigation');
--- a/packages/core/src/SideNav/SideNav.tsx
+++ b/packages/core/src/SideNav/SideNav.tsx
@@ -24,7 +24,6 @@
 import {
   useCallback,
   useImperativeHandle,
-  useLayoutEffect,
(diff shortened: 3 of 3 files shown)


cixzhang (changes requested):
Hmm is it possible for us to avoid a layout effect if useResizable handle controlled states within itself? Then we should be able to pass the controlled state through `useResizable` and leverage the `resizableHook`'s isCollapsed state as source of truth instead of duplicating it here.

Sending back for consideration :)

HelloOjasMutreja:
Yes, that's cleaner. Pushed in 213ae759a: useResizable's isCollapsed is now the single source of truth when the nav is resizable and uncontrolled, instead of SideNav tracking a second, independently-initialized boolean and reconciling it after the fact.

    const collapsed = isControlled
      ? controlledCollapsed
      : isResizable
        ? resizableHook.isCollapsed
        : uncontrolledCollapsed;

The mount-only `useLayoutEffect` is gone entirely -- there's nothing left to reconcile since `collapsed` reads live off `resizableHook.isCollapsed` on every render rather than being seeded once at mount. `toggle()`/`onCollapseChange` are unchanged; they already drove both states' updates going forward, so this only changes what the *initial* source of truth is.

Full suite still green (100/100 in SideNav.test.tsx, 42/42 in AppShell.test.tsx) and the regression test for the original bug (persisted width of 0 restoring as the compact icon rail, not an invisible 0px-expanded nav) still passes against this version.

cixzhang:
@HelloOjasMutreja — closing this one, and I want to be clear about why, because the diagnosis in it is the best thing on this issue.

You were right that `main` already clamps and already seeds, and that the real defect is SideNav keeping a *second*, independently-initialized collapsed boolean. That reframing is what both attempts are now built on, and the revision after my last comment moved exactly the direction I asked for.

Two things I couldn't get past. Reading `collapsed` off the hook (`SideNav.tsx:373`) regresses `defaultIsCollapsed`: `collapsible={{defaultIsCollapsed: true}}` plus `resizable` now renders expanded, because the hook only seeds from a persisted `0` — and there's no test for that combination. And the controlled path is untouched, so a controlled nav with a persisted `0` is still invisible. [#4824](https://github.com/facebook/astryx/pull/4824) covers both plus the lost pre-collapse width, so I'm landing that one. I'm sorry — that's a frustrating outcome after a good revision.

The follow-up I'd genuinely like from you is the durable fix, and it's the thing your analysis pointed straight at: **`useResizable` has no controlled-collapse input**. `isCollapsed` is internal state with no way to drive it, which is precisely why the controlled path can't be fixed from SideNav's side. I've written that up in [#5073](https://github.com/facebook/astryx/issues/5073) along with making `resizable` and `collapsible` mutually exclusive. Adding `defaultIsCollapsed`/`isCollapsed` to the hook — plus the `defaultIsCollapsed` + `resizable` test nothing covers today — would close the class rather than the instance, and you already understand why it matters better than anyone on the thread.

Issue #4855, opened 2026-08-10
[Bug] Four more theme packages override --radius-none away from its fixed 0px contract


HelloOjasMutreja:
## Summary

Follow-up to #4833. While fixing `theme-neutral`'s `--radius-none` override, I checked every theme package's `src/*Theme.ts` for the same mistake and found it repeated in four more:

| Package | `--radius-none` | `--radius-full` |
| --- | --- | --- |
| `@astryxdesign/theme-butter` | `0.125rem` | `9999px` (correct) |
| `@astryxdesign/theme-chocolate` | `0.125rem` | `9999px` (correct) |
| `@astryxdesign/theme-gothic` | `0.125rem` | `9999px` (correct) |
| `@astryxdesign/theme-stone` | `0.125rem` | `9999px` (correct) |
| `@astryxdesign/theme-neutral` | fixed in #4833 (was `0.25rem`) | `9999px` (correct) |
| `@astryxdesign/theme-y2k` | `0px` (correct) | `0px` — separate issue, see below |
| `@astryxdesign/theme-matcha` | not overridden (inherits core default) | not overridden |

`--radius-none` and `--radius-full` are documented in `defineTheme.ts` as "always fixed (never affected by multiplier)", matching `@astryxdesign/core`'s own defaults (`0px` / `9999px`). Each of `butter`, `chocolate`, `gothic`, and `stone` bumps its whole radius group by a base multiplier and `--radius-none` gets swept along, exactly like neutral's bug in #4833 — same pattern, same root cause (a hand-authored `tokens` object bypasses `defineTheme`'s radius generator, which is the only place the "always fixed" guarantee is actually enforced).

Also noting separately (not the same bug, flagging since I found it while auditing): `theme-y2k` sets `--radius-full: '0px'`, which is internally consistent with its fully-square aesthetic but is *also* technically a deviation from the "always fixed" contract as literally documented. Might be intentional (y2k is deliberately boxy) — a maintainer call, not filing that as a separate issue unless wanted.

## Suggested fix

Same as #4833: set `--radius-none: '0px'` in each of `butter`, `chocolate`, `gothic`, and `stone`'s `src/*Theme.ts`.

## Suggested follow-up (separate from this bug)

[...]

PR #4856, opened 2026-08-10
fix(theme-neutral): restore --radius-none to the fixed 0px contract

### LINKED ISSUE #4833: [Bug] theme-neutral sets --radius-none to 0.25rem, breaking the fixed-radius contract
## Summary

`@astryxdesign/theme-neutral@0.3.0` overrides `--radius-none` to `0.25rem`. The core contract says this token is fixed:

- `@astryxdesign/core` default: `--radius-none: 0px` (`dist/astryx.css`).
- `dist/theme/defineTheme.d.ts`: "*--radius-none and --radius-full are always fixed (never affected by multiplier).*"

In `theme-neutral/src/neutralTheme.ts` the whole radius group was bumped ("Radius — slightly larger than default") and `--radius-none` went along:

```ts
// Radius — slightly larger than default (kept as-is)
'--radius-none': '0.25rem',
```

## Impact

Anything that opts out of rounding via `--radius-none` (e.g. squared corners inside flush layouts) silently renders with a 4px radius under the neutral theme. By the token's name and the documented contract, consumers reasonably assume it resolves to `0`.

## Expected

`--radius-none: 0` under every theme, matching the core default and the defineTheme documentation. (`--radius-full` is correctly left at `9999px`.)

## Environment

`@astryxdesign/core@0.3.0`, `@astryxdesign/theme-neutral@0.3.0`.

HelloOjasMutreja:
## Summary

Closes #4833.

`--radius-none` and `--radius-full` are documented in `defineTheme.ts` as "always fixed (never affected by multiplier)", matching `@astryxdesign/core`'s own defaults (`0px` / `9999px`). `theme-neutral`'s hand-authored radius group bump ("slightly larger than default") swept `--radius-none` along with it to `0.25rem` — `--radius-full` was correctly left alone at `9999px`, just `--radius-none` drifted.

## Fix

`--radius-none: '0px'`, matching `@astryxdesign/core`'s own default literal exactly.

## Note

While auditing this, I found the identical mistake independently in four other theme packages (`butter`, `chocolate`, `gothic`, `stone` all override `--radius-none` to a non-zero value the same way). That's out of scope for this issue/PR, so I filed it separately as #4855 rather than bundling unrelated theme packages into this fix.

## Test plan

- [x] `tsc --noEmit` — no errors in the touched file (two pre-existing, unrelated errors in `icons.tsx` from an unrelated icon registry mismatch)
- [x] No test suite exists for individual theme packages (they're hand-authored token data, not exercised by `defineTheme.test.ts`, which only covers the `radius` generator config path) — noted in #4855 as a suggested follow-up (a check script asserting the fixed tokens across all theme packages)

### DIFF
--- a/.changeset/theme-neutral-radius-none-fix.md
+++ b/.changeset/theme-neutral-radius-none-fix.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/theme-neutral': patch
+---
+
+[fix] `--radius-none` no longer overrides to `0.25rem`. `--radius-none` and `--radius-full` are documented as always fixed (never scaled by a theme), matching `@astryxdesign/core`'s own defaults — this theme's radius group bump swept `--radius-none` along with it by mistake. Anything opting out of rounding via `--radius-none` under this theme now renders with a true `0px` radius again, instead of a silent 4px.
+
+@HelloOjasMutreja
--- a/packages/cli/assets/templates/themes/neutral/neutralTheme.ts
+++ b/packages/cli/assets/templates/themes/neutral/neutralTheme.ts
@@ -323,8 +323,11 @@ export const neutralTheme = defineTheme({
 
     // =========================================================================
     // Radius — slightly larger than default (kept as-is)
+    // --radius-none and --radius-full are always fixed and must never be
+    // scaled by a theme (see defineTheme's radius config docs) — 0 and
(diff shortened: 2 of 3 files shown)


cixzhang (approved):
Yeah radius-none probably intended to stay 0. Thanks for fixing!

PR #4858, opened 2026-08-10
fix(Chat): stop clipping the hover radius on grouped ChatToolCalls rows

### LINKED ISSUE #4830: ChatToolCalls: grouped rows clip the hover-background radius (overflow: hidden vs the row's negative inline margin)
## Bug

In `@astryxdesign/core` **0.3.0**, expandable call rows inside a grouped `ChatToolCalls` (2+ calls) render their hover background clipped on the inline edges, so the hover highlight loses its border radius. The same row rendered in single-call mode shows the full rounded hover background.

## Repro

```tsx
<ChatToolCalls
  defaultIsExpanded
  calls={[
    { name: "bash", target: "git diff --stat", status: "complete", duration: "340ms", resultDetail: <pre>…</pre> },
    { name: "read", target: "src/utils/formatDate.ts", status: "complete", duration: "45ms", resultDetail: <pre>…</pre> },
  ]}
/>
```

Hover any row: the highlight is cut flat at the container edges (most visible on the left, where the status icon sits flush against the clipped edge). Render a single call (`calls={[...]}` with one item) and hover: the highlight has its full `--radius-element` rounding and overhangs the text column by 4px as designed.

Computed styles are identical in both modes (`border-radius: 10px` via `callRowClickable`), so this is pure clipping, not a styling divergence.

## Cause

- `callRowClickable` widens the hover background beyond the content column with `paddingInline: --spacing-1` + `marginInline: calc(-1 * --spacing-1)`.
- In group mode the rows live inside the expand/collapse animation wrapper: `groupContent` (grid `0fr → 1fr`) → `groupContentInner` (`overflow: hidden`, required for the height animation) → `list`.
- The 4px overhang produced by the row's negative margin extends past `groupContentInner`'s box and gets cut by its `overflow: hidden`. Measured: rows are 8px wider (584px) than the clip box (576px).

Single-call mode has no such wrapper, so nothing clips the overhang — hence the inconsistency.

## Suggested fix

Compensate for the overhang on the clip container so the animation keeps `overflow: hidden` but the hover background survives, e.g.:

```ts
groupContentInner: {
  overflow: 'hidden',
  minHeight: 0,
  paddingInline: spacingVars['--spacing-1'],
  marginInline: `calc(-1 * ${spacingVars['--spacing-1']})`,
},
```

(or move the negative inline margin from `list`/rows up to a level outside the clip). We're currently working around it downstream with exactly that padding/margin pair applied via CSS to the clip container, scoped to the 0.3.0 DOM shape — happy to drop it once fixed upstream.

Environment: `@astryxdesign/core` 0.3.0, Electron (Chromium) renderer, macOS, default `theme-neutral`.

HelloOjasMutreja:
## Summary

Closes #4830.

Expandable call rows inside a grouped `ChatToolCalls` (2+ calls) render their hover background clipped flat on the inline edges, losing `--radius-element` rounding. The same row in single-call mode (no group wrapper) shows the full rounded highlight.

## Cause

- `callRowClickable` widens the hover background beyond the content column via `paddingInline: --spacing-1` + a matching negative `marginInline`.
- `list` (the rows' direct parent in group mode) has the same padding/negative-margin pair, which locally absorbs each row's overhang — but `list`'s *own* negative margin then overhangs `groupContentInner`'s box by the same amount.
- `groupContentInner` is `overflow: hidden` (required for the `grid-template-rows: 0fr → 1fr` expand/collapse animation), so that overhang gets clipped flat instead of rendered.
- Single-call mode has no such wrapper, so nothing clips the overhang.

## Fix

Mirror the same `paddingInline`/negative-`marginInline` pair on `groupContentInner`, so `list`'s overhang is absorbed at that level too, instead of clipped. Verified the DOM ancestor chain above `groupContentInner` (`groupContent`, then the component root) has no other `overflow: hidden` that would just move the clip up a level.

## Test notes

Added a regression test. jsdom's `getComputedStyle` doesn't resolve StyleX's atomic classes for `padding-inline`/`margin-inline` in this setup (came back empty in both the buggy and fixed states, confirmed by dumping the actual computed style), so I couldn't assert on computed pixel values directly. Instead I asserted on the atomic class names: StyleX emits one deterministic class per distinct property+value pair, so `groupContentInner` and `list` having identical `paddingInline`/`marginInline` values makes them share those exact class names — verified the test fails (0 shared classes) against the pre-fix code and passes (≥2 shared classes) with the fix.

I could not get a live visual/before-after screenshot for this one — Storybook's dev server repeatedly OOM-crashed during Vite's dependency bundling step in my environment (same instability I've hit with other tooling here), so I'm relying on the geometric analysis above (verified against the actual DOM/style source, matching the issue's own precise 8px overhang measurement) plus the class-based regression test rather than a visual diff. Flagging this rather than claiming a visual check I didn't actually get.

## Test plan

- [x] `vitest run packages/core/src/Chat/ChatToolCalls.test.tsx` — 16/16 passing
- [x] `eslint` clean
- [ ] Visual before/after — blocked by local Storybook instability, disclosed above

### DIFF
--- a/.changeset/chat-tool-calls-hover-radius-clip.md
+++ b/.changeset/chat-tool-calls-hover-radius-clip.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] ChatToolCalls: hover backgrounds on grouped call rows (2+ calls) now keep their full `--radius-element` rounding instead of getting clipped flat on the inline edges. `groupContentInner` — the `overflow: hidden` clip boundary the expand/collapse height animation needs — was missing the padding/negative-margin pair that absorbs the row-level hover-background overhang, so the overhang extended past the clip boundary and got cut off. Matches the ungrouped single-call row, which has no such wrapper to clip it.
+
+@HelloOjasMutreja
--- a/packages/core/src/Chat/ChatToolCalls.test.tsx
+++ b/packages/core/src/Chat/ChatToolCalls.test.tsx
@@ -216,4 +216,39 @@ describe('ChatToolCalls', () => {
     expect(region).toHaveTextContent('searchCode');
     expect(region).toHaveTextContent('readFile');
   });
+
+  it('compensates the grouped rows overhang on the clip boundary so the hover background is not cut off', () => {
+    render(
+      <ChatToolCalls
+        defaultIsExpanded={true}
+        calls={[
+          {name: 'bash', status: 'complete'},
+          {name: 'read', status: 'complete'},
+        ]}
+      />,
+    );
+    const regionId = screen.getByRole('button').getAttribute('aria-controls');
+    const groupContent = document.getElementById(regionId as string);
+    // groupContentInner is the overflow:hidden clip boundary required for the
+    // grid height animation; its direct child (`list`) pulls itself outward
+    // by a negative inline margin so each row's hover background can overhang
+    // the text column without widening the layout. Without a matching
+    // padding/negative-margin pair on groupContentInner itself, that overhang
+    // extends past the clip boundary and gets cut off (#4830) instead of
+    // being absorbed the same way it is for a single, unwrapped call row.
+    const groupContentInner = groupContent!.firstElementChild as HTMLElement;
+    const list = groupContentInner.firstElementChild as HTMLElement;
+    // jsdom's getComputedStyle doesn't resolve StyleX's generated atomic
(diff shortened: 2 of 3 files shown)


cixzhang:
Looks good :)

Screenshot 2026-08-10 at 5 34 11 PM

cixzhang (approved):
Nice, thanks for the fix

PR #4863, opened 2026-08-10
fix(cli): don't replace scaffolded video sources with the image placeholder

### LINKED ISSUE #4780: CLI: LightboxVideo scaffold replaces MP4 source with an SVG image data URI

HelloOjasMutreja:
## Summary

Closes #4780.

Scaffolding `LightboxVideo` (`astryx template LightboxVideo ./out --type block`) replaces its video source with the image placeholder SVG data URI, which a `<video>` element can't decode.

## Cause

`stripTemplateAssetRefs()` matched every `/template-assets/*` reference with one pattern and replaced all of them with `PLACEHOLDER_IMAGE`, regardless of file extension.

## Fix and why it's scoped the way it is

Video extensions (`.mp4`, `.webm`, `.mov`, `.ogv`) now strip to an empty `src` instead of the image placeholder.

I want to be upfront about why this isn't "a video-specific inline placeholder" (one of the options the issue suggested): unlike an SVG data URI for images, a `<video src>` needs actual encoded media bytes. I looked into generating a small self-contained placeholder video (tried recording one via `MediaRecorder`/`canvas.captureStream()` in a real browser to get something verifiably valid), but couldn't get a working encoded clip out of that path in my environment, and I didn't want to hand-author binary MP4/WebM bytes I have no way to verify actually decode — that risks trading one silent failure (image data mislabeled as video) for another (a binary blob that looks more legitimate but still doesn't play). An empty `src` is honest about needing the builder to supply their own file, which is at least not broken.

Also verified the fix doesn't regress the image path when both appear in the same file (e.g. a video's `poster` image alongside its `src`).

## Test notes

Added regression tests: `.mp4`/`.webm`/`.mov`/`.ogv` all strip to `src: ''`, image extensions keep their existing placeholder behavior unchanged, and a mixed image+video case handles both correctly. Verified both new tests fail against the pre-fix code (video refs incorrectly got the image data URI) and pass with the fix.

Also verified end-to-end against the real CLI, not just the unit tests:
```
$ astryx template LightboxVideo ./out --type block
$ grep src: ./out/LightboxVideo.tsx
          src: '',
```

## Test plan

- [x] `vitest run --project node packages/cli/api/template/template.test.mjs` — 10/10 passing
- [x] `vitest run --project node packages/cli/api/template/copy/copy.test.mjs` — 4/4 passing (consumer of `stripTemplateAssetRefs`)
- [x] Verified both new tests fail against the pre-fix code and pass against the fix
- [x] End-to-end CLI verification (see above)
- [x] `eslint` clean

### DIFF
--- a/.changeset/cli-scaffold-video-placeholder.md
+++ b/.changeset/cli-scaffold-video-placeholder.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/cli': patch
+---
+
+[fix] Scaffolding a template that references demo video (e.g. `LightboxVideo`) no longer replaces the video source with the image placeholder data URI, which the generated `<video>` element couldn't play. `stripTemplateAssetRefs()` treated every demo-media reference as an image regardless of extension; video extensions (`.mp4`, `.webm`, `.mov`, `.ogv`) are now stripped to an empty `src` instead — there's no equivalent self-contained inline placeholder for video, so the scaffolded example is honest about needing the builder to supply their own file rather than pointing at something that can't play.
+
+@HelloOjasMutreja
--- a/packages/cli/api/template/template.test.mjs
+++ b/packages/cli/api/template/template.test.mjs
@@ -52,6 +52,32 @@ describe('stripTemplateAssetRefs', () => {
     const out = stripTemplateAssetRefs(src);
     expect(out).toBe(src);
   });
+
+  it('strips a /template-assets .mp4 source to an empty string, not the image data URI (#4780)', () => {
+    const src = "src: '/template-assets/Nature-1.mp4',";
+    const out = stripTemplateAssetRefs(src);
+    expect(out).toBe("src: '',");
+    expect(out).not.toContain('data:image/svg+xml,');
+    expect(out).not.toContain('/template-assets/');
+  });
+
+  it('strips other video extensions (.webm, .mov, .ogv) the same way', () => {
+    for (const ext of ['webm', 'mov', 'ogv']) {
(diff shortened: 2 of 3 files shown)


cixzhang (approved):
This makes sense. Thanks!

PR #4865, opened 2026-08-10
fix(dialog): close only the top-most modal on Escape (shared escape stack)

freddymeta:
## The bug

Pressing **Escape inside a modal that was opened from within another modal closes _both_ modals** (and the same for a Sheet opened over a Modal, etc.).

Astryx already has top-most-only Escape dismissal — but only for focus-trap overlays (Popover, menus) via `useFocusTrap`'s shared escape stack. `Dialog` (which backs Modal **and** the Sheet/drawer variant) takes a separate path: it uses the native `<dialog>` element and attaches its own `keydown` listener. That listener:

1. **never calls `stopPropagation()`**, and
2. **only _reads_ the focus-trap escape state** (`hasActiveFocusTrapEscape()`) to defer to a popover on top — it never _registers_ itself.

So when Modal L1 is opened from inside Modal L0, L1's `<dialog>` is a DOM descendant of L0's. An Escape inside L1 → L1's handler closes it, then the event **bubbles** to L0's handler, which also sees "no focus-trap on top" and closes too. Dialog-on-dialog had no coordination at all.

## The fix

Extract the escape-stack coordination that already lived inside `useFocusTrap` — the push/pop registry and the **DOM-containment top-most resolver** — into a small shared hook, **`useEscapeStack`**, and have **both** overlay families register on it:

- `useFocusTrap` (Popover, menus) now consumes `useEscapeStack` internally — behavior unchanged.
- `Dialog` registers its `<dialog>` on the same stack, gates **both** its `keydown` handler and its native `cancel` handler on `isTopmost()`, and **`stopPropagation()`s** when it handles the press.

Result: one Escape press dismisses exactly one layer — the top-most — whether the top is a popover, a menu, or another modal. Top-most is resolved by DOM containment (correct even when an inner and outer overlay mount in the same commit, where React runs the child effect first). One stack and one resolver now serve the whole overlay family, which is the maintenance win — no parallel escape logic to keep in sync.

```tsx
// before: two nested Modals both close on one Escape
// after: only the inner Modal closes; the outer stays open
```

## Why a shared hook (not duplicating logic in Dialog)

The two primitives have very different event models (a focus trap listening on `document` vs. a native `<dialog>`'s `keydown`/`cancel`), but the *coordination* — who's on top, who handles Escape — is identical. Keeping it in one place means there's a single source of truth for top-most-only dismissal across every overlay.

## Tests

- New **modal-in-modal regression test** in `Dialog.test.tsx`: an Escape fired from inside a DOM-nested inner dialog closes only the inner one, and a follow-up asserts the outer closes once it's top-most. I verified this test **fails on the pre-fix behavior** (it reports the outer being closed too), so it's a real guard.
- All existing Dialog, Popover, and `useFocusTrap` Escape tests pass unchanged (97 across the three suites, incl. `useFocusTrap`'s own nested-trap coordination and the popover-falls-through-to-Dialog case).

## Notes

- No API change. `hasActiveFocusTrapEscape` is kept as a back-compat alias of the renamed `hasActiveEscapeLayer`.
- `useEscapeStack`, `hasActiveEscapeLayer`, and `isImeKeyEvent` are exported from `@astryxdesign/core` (via `hooks`), so other overlays that manage their own Escape can join the same stack.
- Typecheck, strict ESLint, and repo guardrails (sync-exports/changesets/boundaries) all clean.

### DIFF
--- a/.changeset/dialog-escape-stack.md
+++ b/.changeset/dialog-escape-stack.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Dialog/Modal/Sheet: pressing Escape inside a modal opened from within another modal now closes only the top-most one, instead of closing both. Every overlay that dismisses on Escape — focus-trap overlays (Popover, menus) and native `<dialog>` overlays (Dialog and its Sheet/drawer variant) — now shares one Escape stack via the new `useEscapeStack` hook, and only the top-most layer (resolved by DOM containment) handles the press and stops it propagating. Previously each open `<dialog>` ran its own Escape listener without stopping propagation, so an inner dialog's Escape also bubbled to the DOM-nested outer dialog and dismissed it too. No API change.
+@freddymeta
--- a/packages/core/src/Dialog/Dialog.test.tsx
+++ b/packages/core/src/Dialog/Dialog.test.tsx
@@ -144,6 +144,90 @@ describe('Dialog', () => {
     });
   });
 
+  describe('nested-modal Escape coordination', () => {
+    // Regression: an Escape pressed inside a modal opened FROM another modal
+    // used to close both — each open <dialog> attached its own keydown listener
+    // and none stopped propagation, so the inner dialog closed and the event
+    // then bubbled to the DOM-nested outer dialog, which closed too. Both
+    // dialogs now register on the shared Escape stack (the same one Popover uses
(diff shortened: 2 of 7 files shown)


cixzhang:
Thanks for this — the modal-in-modal repro and the containment resolver are spot on. Two came in for the same bug and we're going with #4881, since it takes every overlay type onto one stack. Your cancel gate and your Dialog-level Escape test are both in it, credited.

PR #4866, opened 2026-08-10
fix(hooks): snap useStreamingText's rendered slice to a grapheme boundary

### LINKED ISSUE #4779: useStreamingText: reveal ticks can slice mid-grapheme (transient broken emoji)
## Problem

`useStreamingText` advances its reveal window by fixed code-unit steps and slices at that offset:

- `packages/core/src/hooks/useStreamingText.ts:171` advances `displayedLen` by `charsPerTick` code units
- `packages/core/src/hooks/useStreamingText.ts:194` renders `targetText.slice(0, displayedLen)`

When the boundary lands inside a surrogate pair, flag sequence, or ZWJ emoji, the rendered frame ends with a lone surrogate (displays as U+FFFD) or a partial sequence (a family emoji flashes as its individual members). The glitch is transient and self-corrects on a later tick, but it is visible at natural speed on emoji-dense streams.

The doc comment (line 79) says the hook advances "on word and syntax boundaries"; the implementation advances by fixed code-unit steps. The comment should be corrected alongside the fix.

## Proposed fix

Snap the visible slice down to the nearest grapheme cluster boundary before rendering: `Intl.Segmenter.prototype.segment(str).containing(offset)` gives the cluster start for any offset, with a code-point fallback consistent with the `utils/grapheme.ts` proposed in #4776. The tick counter can keep advancing in code units so the reveal cadence is unchanged; only the rendered slice needs snapping.

## Context

Same family as #4759 (grapheme-safe string handling). Surfaced in #4776 and deliberately left out of its scope.

HelloOjasMutreja:
## Summary

Closes #4779.

`useStreamingText` advances its reveal window by fixed code-unit steps and slices at that offset. When the boundary lands inside a surrogate pair, flag sequence, or ZWJ emoji, the rendered frame ends with a lone surrogate (displays as U+FFFD) or a partial glyph sequence for one frame before self-correcting on a later tick — visible at natural speed on emoji-dense streams.

## Fix

Added `snapToGraphemeBoundary()`, using `Intl.Segmenter`'s grapheme granularity (`segment(text).containing(offset).index`, per the issue's own suggested approach) to snap the rendered slice back to the start of whichever grapheme cluster the raw tick boundary falls inside — with a surrogate-pair-safe fallback for the (now essentially nonexistent) case where `Intl.Segmenter` is unavailable, matching the existing `Intl.Segmenter`-with-fallback pattern already established in `Avatar.tsx`.

The tick counter (`displayedLen`) keeps advancing in code units exactly as before, so reveal cadence is unchanged — only the *rendered* slice is affected. A not-yet-complete cluster is simply held back to the next tick instead of rendering broken.

Also corrected the hook's doc comment per the issue's own note: it claimed the cadence "advances on word and syntax boundaries," but the implementation always advanced by fixed code units — that was inaccurate before this fix and unrelated to it, just a stale claim worth fixing alongside.

I did not depend on the `utils/grapheme.ts` module mentioned in #4776/#4759, since that PR (#4776) is still open/unmerged — this fix is self-contained.

## Test notes

Added a unit test suite for `snapToGraphemeBoundary` directly (surrogate pairs, ZWJ family emoji, boundary offsets left alone), plus an integration test through the hook itself that reproduces the exact failure from the issue: 9 ASCII chars followed by a 4-person ZWJ family emoji, with `natural` speed's 10-code-unit tick landing 1 code unit into the family sequence. Verified it fails against the pre-fix code (`result.current` is `'aaaaaaaaa\uFFFD'`, the literal replacement-character glyph from the issue) and passes with the fix (the incomplete cluster is held back entirely).

## Test plan

- [x] `vitest run packages/core/src/hooks/useStreamingText.test.ts` — 14/14 passing
- [x] `vitest run packages/core/src/Markdown --exclude "**/*.perf.test.ts"` — 239/239 passing (consumer of the hook; excluded a separately-flaky, unrelated performance-timing test that passes in isolation)
- [x] Verified the new tests fail against the pre-fix code (reproducing the exact U+FFFD glyph from the issue) and pass against the fix
- [x] `tsc --noEmit` clean for the touched file
- [x] `eslint` clean

### DIFF
--- a/.changeset/streaming-text-grapheme-boundary.md
+++ b/.changeset/streaming-text-grapheme-boundary.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] `useStreamingText` no longer renders a broken glyph (a lone surrogate, or a partial ZWJ emoji sequence) for one frame when its fixed-code-unit reveal cadence happens to land inside a surrogate pair or multi-codepoint emoji. The rendered slice now snaps back to the nearest grapheme cluster boundary via `Intl.Segmenter` (with a surrogate-pair-safe fallback where it's unavailable); the reveal cadence itself is unchanged. Also corrected the hook's doc comment, which inaccurately described the cadence as advancing on word/syntax boundaries — it always advanced by fixed code units.
+
+@HelloOjasMutreja
--- a/packages/core/src/hooks/useStreamingText.test.ts
+++ b/packages/core/src/hooks/useStreamingText.test.ts
@@ -2,7 +2,7 @@
 
 import {describe, it, expect, vi, beforeEach, afterEach} from 'vitest';
 import {renderHook, act} from '@testing-library/react';
-import {useStreamingText} from './useStreamingText';
+import {snapToGraphemeBoundary, useStreamingText} from './useStreamingText';
 
 describe('useStreamingText', () => {
   let rafCallbacks: ((time: number) => void)[];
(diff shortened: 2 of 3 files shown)


cixzhang (approved):
Neat, thanks for these i18n fixes!

Issue #4849, opened 2026-08-10
[RFC] CodeBlock copy button: default tooltip + theme target, backed by a shared useClipboard hook


freddymeta:
## Problem Statement

`CodeBlock`'s built-in copy control can't be adopted by a design system building on top of Astryx, so consumers turn it off (`hasCopyButton={false}`) and re-implement the whole thing. Three concrete gaps in the current control:

1. **No tooltip.** The copy button is a bare `<button>` with only an `aria-label`. A visible "Copy" hover/focus hint is a near-universal expectation for an icon-only copy affordance, and a tooltip is DOM — it can't be added through theming or CSS.
2. **No theme seam.** `CodeBlock` exposes a single theming target (`themeProps('codeblock')`, on the root `<pre>`). The header, title, caret, and copy button are internal StyleX with no targets, so the copy button is unreachable — a theme can't restyle it.
3. **`hasCopyButton` is all-or-nothing.** It's a plain boolean; there's no way to keep the behavior but change the control.

The result is a predictable escape hatch: consumers set `hasCopyButton={false}`, then re-implement the clipboard write, the copied-state timer, the copy announcement, and absolute-position a replacement button against `CodeBlock`'s internal DOM. That's the "reached for an unsupported structural selector" signal that a sanctioned improvement is missing — and the re-implemented control they build is, every time, *"a compact ghost icon button with a Copy tooltip and a copy→check flip."* Astryx should just ship that.

## Evidence of Demand

The same re-implementation recurs, and it's near-identical each time:

- A ghost, `size="sm"`, icon-only `Button` with `tooltip="Copy"`.
- `copy` → `check` icon flip on success, reverting after ~2s.
- The tooltip stays "Copy" after copying — the checkmark is the confirmation, not a tooltip change.
- The clipboard write + copied-state timer + a polite live-region announcement (swapping `aria-label` alone isn't reliably announced).

That clipboard + copied-state + announce block is **already duplicated inside Astryx itself** — `CodeBlock` and `Timestamp`'s copyable hover card carry byte-for-byte the same logic, including the same non-obvious details (restart the reset timer on rapid re-copy; announce via `useAnnounce` because an `aria-label` swap isn't announced; silent no-op on clipboard rejection). Two internal copies of the same behavior, re-derived independently, is the extraction signal.

## Why Existing Mechanisms Don't Cover This

- **Theming** can't add a tooltip (DOM, not style) and has no target on the copy button.
[...]

PR #4867, opened 2026-08-10
feat(codeblock): themeable copy button with default tooltip, backed by a shared useClipboard hook

freddymeta:
Implements the fix from #4849.

## What

`CodeBlock`'s built-in copy button couldn't be adopted by a design system building on Astryx, so the escape hatch was `hasCopyButton={false}` + re-implement the whole control. This fixes the three gaps directly on the built-in button:

1. **No tooltip.** It was a bare `<button>` with only an `aria-label`. It's now a ghost `IconButton`, so it carries a **default "Copy code" tooltip** (on by default — no prop). The tooltip stays "Copy code" after copying; the `copy` → `check` icon flip is the confirmation. The `aria-label` still swaps to the localized "Copied" for assistive tech, backed by the announcement.
2. **No theme seam.** The only theming target was the root `<pre>`. The copy button is now reachable via the stable `astryx-codeblock-copy-button` class — theme it through the `codeblock-copy-button` component key to restyle size/color/hover **without turning it off**.
3. **All-or-nothing.** `hasCopyButton={false}` still fully removes it; that path is unchanged.

## `useClipboard` hook

The clipboard write + transient copied flag + reset timer + polite announcement was **duplicated byte-for-byte** in `CodeBlock` and `Timestamp`'s copyable hover card. Per API Conventions → *Behaviors: Hooks Over Wrappers*, that behavior is now a hook — `useClipboard` (`@astryxdesign/core/hooks`) — and both components are thin controls over it. The two duplicated implementations collapse to one.

```ts
const {copy, isCopied} = useClipboard({announce: 'Copied', resetAfterMs: 2000});
// copy(text) → writes, flips isCopied, announces, restarts the reset timer;
// silent no-op resolving false on rejection; timer cleaned up on unmount.
```

Reach for it directly for a copy affordance that isn't a plain icon button (a menu item, a labeled text button, a value chip).

## Scope

- **In:** `useClipboard`; CodeBlock built-in button → themeable ghost `IconButton` + default tooltip + `codeblock-copy-button` target; migrate CodeBlock + Timestamp onto the hook.
- **Out (called out in #4849, deliberately not bundled):** a CodeBlock **header slot** for arbitrary header actions, and header/body **metric theming targets**.

## Surface

- `useClipboard.ts` + `useClipboard.doc.mjs` (EN + dense) + `useClipboard.test.tsx` (8 tests), exported from `hooks/index.ts`.
- CodeBlock: button rewrite, `codeblock-copy-button` theming target in `CodeBlock.doc.mjs`, 3 new tests (theme-target class, tooltip present, tooltip stays "Copy code" after copy).
- Timestamp: `TimestampHoverCard` copy button migrated onto `useClipboard` (behavior byte-identical; existing tests unchanged).
- Storybook: `useClipboard.stories.tsx` (icon button, labeled button, no-announce).
- Changeset (patch).

## Verification

`pnpm -F @astryxdesign/core build`, `typecheck`, `typecheck:docs`, `lint` (0 errors), and `pnpm -F @astryxdesign/docsite generate && test` all green. Core suite green — `useClipboard` 8, `CodeBlock` 18, `Timestamp` 76. The docsite `generate` runs `astryx theme build`, which validated the new `codeblock-copy-button` target end to end.

### DIFF
--- a/.changeset/codeblock-copy-button-themeable.md
+++ b/.changeset/codeblock-copy-button-themeable.md
@@ -1,0 +1,9 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] CodeBlock: the built-in copy button is now a themeable ghost `IconButton` with a default "Copy code" tooltip, reachable via the stable `astryx-codeblock-copy-button` class (theme it through the `codeblock-copy-button` component key). Restyle or keep the copy control without turning it off and re-implementing it. The tooltip stays "Copy code" after copying — the copy→check icon flip is the confirmation.
+
+[feat] New `useClipboard` hook (`@astryxdesign/core/hooks`): the shared copy-to-clipboard behavior — clipboard write, a transient `isCopied` flag with its reset timer, and an optional polite screen-reader announcement. CodeBlock and Timestamp now build their copy buttons on it; reach for it directly for copy affordances that are not a plain icon button.
+
+@freddymeta
--- a/apps/storybook/stories/useClipboard.stories.tsx
+++ b/apps/storybook/stories/useClipboard.stories.tsx
@@ -1,0 +1,90 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+import type {Meta, StoryObj} from '@storybook/react';
+import {useClipboard} from '@astryxdesign/core/hooks';
+import {IconButton} from '@astryxdesign/core/IconButton';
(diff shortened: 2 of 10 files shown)


kentonquatman (approved):
This sounds good. Please add a screenshot verifying that the change is working as expected before merging.

cixzhang (approved):
Nice, thanks for consolidating these!

PR #4818, opened 2026-08-08
[lab] Visual polish for RichTextEditor and toolbar

ernestt:
## Summary

Refreshes the lab RichTextEditor so it looks and behaves like an Astryx form field, adds a compact formatting toolbar, and replaces the browser URL prompt with an Astryx link dialog.

This revision is intentionally scoped to `@astryxdesign/lab`: it does not modify Toolbar, ToggleButton, MoreMenu, OverflowList, useListFocus, or any other Core component.

The component remains named `RichTextEditor`: it is an extensible Lexical authoring surface with formatting, plugins, nodes, history, and serialization—not only a rich-text value input.

## Field visuals and sizing

- Matches TextArea/TextInput for the resting outline, hover treatment, focus-within ring, radius, disabled/read-only treatment, and semantic status colors.
- Aligns placeholder and entered text by sharing the same body typography, line height, responsive coarse-pointer sizing, and content inset.
- Adds `minHeight: SizeValue` for the editable surface. Numbers are pixels, strings are CSS lengths, the default remains `4.5rem`, and content can grow beyond the minimum.
- Adds a dedicated `toolbar` slot before the padded editing surface, keeping controls flush with the top edge and in the correct visual and keyboard order.
- Corrects accessible-name wiring: the contenteditable has its own input ID and references a real label ID.

## Status treatment

- Adds the same in-field semantic status icon used by Astryx inputs.
- Supports `statusVariant="attached" | "detached" | "tooltip"`:
  - **attached** — the icon stays in the editor and the message connects below the field.
  - **detached** — the message sits below as its own status row with an icon.
  - **tooltip** — a focusable in-editor icon reveals the message without reserving a message row.
- Error sets `aria-invalid`; descriptions, status messages/tooltips, character count, and keyboard hints remain correctly composed through `aria-describedby`.

## Formatting toolbar

- Uses the existing Astryx `Toolbar` with small controls; ToggleButton size is passed explicitly inside Labs, so no Core sizing behavior changes.
- Groups Undo/Redo, block formatting, and inline formatting with vertical dividers.
- Combines Paragraph, Heading 1–3, bulleted list, numbered list, and block quote into one block-format Selector.
- Leaves the Selector's standard adaptive popover placement unchanged.
- Uses ToggleButtons for Bold, Italic, Underline, Strikethrough, Inline code, and Link, with active state synchronized to the Lexical selection.
- Removes the experimental AI action.

## Narrow-width behavior

[...]

PR #4868, opened 2026-08-10
Toolbar: support responsive OverflowList composition

ernestt:
## Summary

This draft explores the small Core changes needed to make OverflowList a safe responsive strategy inside Toolbar. It is intentionally separate from #4818 so the integration can be reviewed on its own.

- Allow start and end action slots to shrink so a nested OverflowList receives a constrained width.
- Scope Toolbar roving focus to controls it owns, excluding inert measurement copies, popovers, menus, and nested composite widgets.
- Promote a clicked or programmatically focused toolbar control to the active roving tab stop.
- Let MoreMenu accept compound DropdownMenu children as an alternative to data items, which preserves checked state for formatting actions moved into overflow.
- Add API guidance and a resizable responsive Toolbar story with stateful formatting actions.

## Why these changes are needed

OverflowList renders a hidden inert measurement copy of every child. Toolbar currently queries every descendant button, input, or tabindex, so it includes those copies in its roving order. Toolbar slots also keep their intrinsic width, which prevents OverflowList from observing the actual constrained space. Finally, data-only MoreMenu items cannot represent controlled checkbox actions without losing their component state model.

## Scope and risk

- No changes to the OverflowList measurement algorithm.
- No RichTextEditor or Lexical-specific behavior.
- No ToggleButton sizing or Toolbar spacing changes.
- The MoreMenu API change is additive: existing items usage is unchanged.
- The focus change is covered at both hook and Toolbar integration levels.

## Storybook

Core / Toolbar / Composition: Responsive Overflow

Use the width slider to force actions into MoreMenu, then toggle actions both inline and from the overflow menu to verify state continuity.

## Validation

- 78 focused tests pass across Toolbar, MoreMenu, and useListFocus.
- Core TypeScript check passes.
- Changed source and story files pass ESLint and Prettier.
- Repository pre-commit checks pass, including SYNC, package boundaries, changesets, demo media, executable bits, and CLI structure.
- The full Storybook typecheck in a fresh worktree is currently limited by unbuilt theme, charts, and Vega workspace package outputs; after resolving the story-specific issue, it reports no Toolbar.stories.tsx errors.


packages/core/src/Toolbar/Toolbar.tsx
```diff
@@ -70,6 +70,10 @@ const styles = stylex.create({
   startSlot: {
     display: 'flex',
     alignItems: 'center',
+    // Flex items default to min-width:auto, which keeps wide action rows from
+    // shrinking and prevents nested OverflowList containers from ever seeing
+    // a constrained width.
+    minWidth: 0,
```
cixzhang, line 76: We're preventing shrinking from all the slots in the toolbar now in a way where it doesn't have clear prioritization between content. It could be helpful to think about the cases that may be affected by this. IIRC internally we had props to adjust prioritization for slots, but it's always a bit of a challenge to choose the right one.

ernestt: Good callout — right now `min-width: 0` on both the start and end slots makes every slot equally shrinkable (default `flex-shrink: 1`) with no defined priority, and since the slots don't clip, contents can overlap rather than yield in a predictable order. We really only need the slot that hosts the OverflowList to see a constrained width. A few ways to give this clear prioritization:

**A — Opt-in collapsible slot (smallest API).** Add a prop naming which slot may shrink; only that one gets `min-width: 0`. Every other slot keeps `min-width: auto`, i.e. the pre-change behavior, so there's no regression and primary-action slots stay protected.

```tsx
/** Which slot may shrink below its content so a nested OverflowList
 *  can observe a constrained width. Others keep intrinsic width. */
collapsibleSlot?: 'start' | 'center' | 'end';
```

**B — Per-slot shrink priority (closest to the old internal API).**

```tsx
slotPriority?: { start?: number; center?: number; end?: number }; // → flex-shrink
```

Most flexible and mirrors what we had internally, but it reintroduces the "which number do I pick?" ergonomics problem — more surface area and harder to document.

**C — Convention-based default, no new prop.** Encode a default order: `start` shrinks before `center` before `end` (end holds primary actions), e.g. `flex-shrink: 1` on start, `0` on end. Matches the existing doc guidance (secondary left, primary right) at zero API cost, but bakes in an assumption and breaks if someone puts the OverflowList in `end`.

**D — Contain the spill (safety net, pairs with A or C).** Whatever shrinks gets `overflow: hidden` alongside `min-width: 0`, so content is clipped instead of overlapping its neighbor under extreme narrowness.

My lean is **A + D**: an opt-in `collapsibleSlot` that applies `{ minWidth: 0, flexShrink: 1, overflow: 'hidden' }` to just the named slot. It gives explicit single-slot prioritization, is a no-op for every existing Toolbar, keeps this story working via `collapsibleSlot="start"`, and avoids the "pick the right shrink number" trap — while leaving the door open to add the full priority API (B) later if a real multi-slot-shrink case shows up. Curious which direction you'd prefer.

PR #4876, opened 2026-08-11
feat(core): converge input clear buttons on a shared themeable primitive

freddymeta:
## Problem

The clearable input family split its clear (✕) affordance two ways. Typeahead, Tokenizer, and FileInput composed the shared `InputClearButton`; TextInput, NumberInput, TimeInput, DateInput, DateTimeInput, DateRangeInput, Selector, and MultiSelector each rolled their own bare `<button><Icon close></button>` with a duplicated `clearButton`/`iconButton` StyleX block.

That split made theming the clear glyph awkward: only four of them (`DateInput`, `DateRangeInput`, `Selector`, `MultiSelector`) ever exposed a per-component `*-clear-icon` target, and the rest were unreachable except through a fragile descendant selector. Adding yet more per-component targets would have entrenched the divergence.

## Change

Converge the whole family onto the shared `InputClearButton`, then theme it in **one** place.

- The eight roll-your-own inputs drop their own clear `<button>`/`<Icon>` (and the now-dead `clearButton` styles + unused imports) and render `<InputClearButton>`, joining Typeahead, Tokenizer, and FileInput. Every clearable input now shares one clear affordance.
- `InputClearButton` owns a single theme target, `astryx-input-clear-icon`, on its glyph — documented once on `Field`. A theme recolors/resizes/hover-styles the clear glyph across every input from that one key.
- It gains a narrow `iconClassName` passthrough (plain class string, not a theme-props-as-prop). The four inputs that shipped a component-specific target in 0.2.0 pass their old class through it, so `astryx-date-input-clear-icon`, `astryx-date-range-input-clear-icon`, `astryx-selector-clear-icon`, and `astryx-multi-selector-clear-icon` **still render** and are documented `deprecatedFor: 'input-clear-icon'` — the same deprecation mechanism `RadioIndicator`/`CheckboxInput` use for their renamed targets.

This addresses the review: apply the clear button consistently across all inputs, then make that shared button themeable — rather than growing the per-component target surface or passing theme props as a prop.

## Behavior

The four inputs that already used `InputClearButton` are unchanged. The eight that rolled their own now adopt the shared ghost-button look: the clear control gains a hover affordance and its glyph inherits the button color instead of being a bare `secondary` icon. This is the intended, visible convergence — the default clear look is now defined in one component.

## Testing

- New `InputClearButton.test.tsx`: renders a real labelled button, fires `onClick`, carries `astryx-input-clear-icon`, matches a standalone `inherit`/`sm` close icon aside from the target class, merges `iconClassName`, and exposes the target to `defineTheme`.
- Per-consumer clear-icon tests updated to assert the canonical target + the legacy alias, and the new shared-button default look. The four legacy targets are still asserted present (via passthrough).
- `themingTargets` guard green (canonical target documented on `Field`; legacy names documented `deprecatedFor`).
- Full `@astryxdesign/core` suite green (6169 tests). `pnpm -F @astryxdesign/core build` green (i18n → esm → tsc → StyleX CSS → UMD). `pnpm lint:strict` and `pnpm check:repo` clean.


kentonquatman (approved):
This sounds good. Please add a screenshot verifying that the change is working as expected before merging.

packages/core/src/DateTimeInput/DateTimeInput.tsx
cixzhang: Seems like we don't consistently apply the clear button.

packages/core/src/Field/InputClearButton.tsx
```diff
@@ -25,19 +26,26 @@ export interface InputClearButtonProps {
   label: string;
   onClick: (e: React.MouseEvent<HTMLButtonElement>) => void;
   xstyle?: stylex.StyleXStyles;
+  /**
+   * Theme target applied to the clear glyph, letting a theme restyle just this
+   * icon (color, size, hover) via `defineTheme`. Pass the owning component's
+   * `themeProps('<component>-clear-icon')`; omit to leave the icon untargeted.
+   */
+  iconThemeProps?: ThemeProps;
```
cixzhang: It's unusual for us to have theme props as a prop. Maybe just className and style?

Also, since we have a clear button for inputs, maybe the best way is to ensure we're applying it consistently across all our inputs and then make this input clear button themeable?

cixzhang (changes requested):
Back to you for the suggestions. Let me know what you think!

packages/core/src/Field/InputClearButton.tsx
freddymeta: Both done — thanks, this was the right call.

- **Dropped the theme-props-as-a-prop.** `InputClearButton` now owns a single theme target on its glyph, `astryx-input-clear-icon`, documented once on `Field`. The only prop it gained is a plain `iconClassName: string` (your "just className") — used solely so the four inputs that already shipped a `*-clear-icon` target in 0.2.0 can keep emitting it.
- **Applied it consistently across the family.** All eight inputs that rolled their own clear button — TextInput, NumberInput, TimeInput, DateInput, DateTimeInput, DateRangeInput, Selector, MultiSelector — now compose `InputClearButton`, joining Typeahead, Tokenizer, and FileInput. Every clearable input shares one clear affordance, themeable in one place.
- **No public target broken.** `astryx-date-input-clear-icon`, `astryx-date-range-input-clear-icon`, `astryx-selector-clear-icon`, and `astryx-multi-selector-clear-icon` still render via the passthrough and are documented `deprecatedFor: 'input-clear-icon'` — the same mechanism `RadioIndicator`/`CheckboxInput` use for their renamed targets.
- **Default color:** the shared glyph is `secondary`, matching the other field affordances (chevrons, calendar toggles, status icons) rather than the ghost button's inherited primary — so the clear ✕ sits consistently beside them.

packages/core/src/DateTimeInput/DateTimeInput.tsx
freddymeta: Fixed — the clear button is now applied consistently. DateTimeInput (and the other seven inputs that rolled their own) compose the shared `InputClearButton`, so the whole clearable family renders the same clear affordance and is themeable from the single `astryx-input-clear-icon` target.

cixzhang (approved):

PR #4878, opened 2026-08-11
feat(core): add search-input theme target to Selector and MultiSelector

freddymeta:
## What this does

Adds a stable `astryx-selector-search-input` / `astryx-multi-selector-search-input` theme target on the in-dropdown search field of `Selector` and `MultiSelector`, so a theme can restyle just that field via `defineTheme`.

## Why

`Selector`/`MultiSelector` with `hasSearch` render a full `TextInput` in the dropdown (magnifier `startIcon` + `hasClear` ✕). Its surface — border, background, radius, focus ring — and its placeholder had no stable seam: a theme could only reach them through a structural descendant selector against the search wrapper and the input's hashed classes. That is the unsupported escape-hatch tier, and it breaks on any internal refactor.

A common need is to render the search **borderless and flush** in the popover (a command-palette look) rather than as a bordered field. That is a paint change on the field's own surface, so it belongs on a paint target on the field itself.

## What changed

- The search `TextInput` in each component's `renderSearch` now carries `themeProps('<component>-search-input').className`, passed through `TextInput`'s `className` prop. It **composes** with the field's own `astryx-text-input` class (via `mergeProps` concatenation) rather than replacing it — the target sits on the element that already paints the surface, so same-element rules in `@layer astryx-theme` reach the border/background/radius/focus ring and `::placeholder`.
- Both components' theming tables list the new target.
- Tests cover class placement on the field root and theme-CSS emission (surface + `::placeholder`).

Default rendering is unchanged — the target is purely additive and changes nothing until a theme targets it.

Example:

```ts
defineTheme({
  components: {
    'selector-search-input': {
      base: {
        border: 'none',
        backgroundColor: 'transparent',
        boxShadow: 'none',
        '::placeholder': {color: 'var(--color-text-tertiary)'},
      },
    },
  },
});
```

## Testing

- `Selector` + `MultiSelector` suites green (195 tests), including the new per-component target tests: (1) the target lands on the search field root and composes with `text-input`, (2) `defineTheme` emits `.astryx-<component>-search-input` and its `::placeholder` rule through `generateThemeCSS`.
- `pnpm -F @astryxdesign/core build` green. Lint clean.

### DIFF
--- a/packages/core/src/MultiSelector/MultiSelector.tsx
+++ b/packages/core/src/MultiSelector/MultiSelector.tsx
@@ -1192,6 +1192,14 @@ export function MultiSelector<T extends MultiSelectorOptionType>({
               : undefined
           }
           value={searchQuery}
+          // Stable theme target on the search field, so a theme can restyle
+          // just this TextInput's surface (border, background, radius, focus
+          // ring) and placeholder via `defineTheme` — e.g. to render the search
+          // borderless and flush in the popover — without a structural
+          // descendant selector. Spread (not just `.className`) so future
+          // states/variants reflect as data-* too; composes with the field's
+          // own `text-input` class via TextInput's prop passthrough.
+          {...themeProps('multi-selector-search-input')}
           onChange={handleSearchChange}
           onKeyDown={e => {
             // Arrow keys navigate options; Enter toggles; Escape closes.
--- a/.changeset/selector-search-input-target.md
+++ b/.changeset/selector-search-input-target.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] Selector and MultiSelector expose a `*-search-input` theme target (`astryx-selector-search-input`, `astryx-multi-selector-search-input`) on the in-dropdown search field, so a theme can restyle just that field's surface (border, background, radius, focus ring) and placeholder via `defineTheme` — e.g. to render the search borderless and flush in the popover — instead of a structural descendant selector. The target composes with the field's own `text-input` class through TextInput's `className` passthrough. Purely additive — default rendering is unchanged.
+
+@freddymeta
--- a/packages/core/src/MultiSelector/MultiSelector.doc.mjs
+++ b/packages/core/src/MultiSelector/MultiSelector.doc.mjs
@@ -36,6 +36,7 @@ export const docs = {
         visualProps: ['size'],
         states: ['select-all', 'selected', 'disabled'],
       },
+      {className: 'astryx-multi-selector-search-input'},
     ],
   },
   components: [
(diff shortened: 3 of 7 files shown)


kentonquatman (approved):
This sounds good. Would you mind adding a screenshot verifying the change is working as expected before merging?

packages/core/src/MultiSelector/MultiSelector.tsx
```diff
@@ -1188,6 +1188,13 @@ export function MultiSelector<T extends MultiSelectorOptionType>({
               : undefined
           }
           value={searchQuery}
+          // Stable theme target on the search field, so a theme can restyle
+          // just this TextInput's surface (border, background, radius, focus
+          // ring) and placeholder via `defineTheme` — e.g. to render the search
+          // borderless and flush in the popover — without a structural
+          // descendant selector. Composes with the field's own `text-input`
+          // class via TextInput's `className` passthrough.
+          className={themeProps('multi-selector-search-input').className}
```
cixzhang: We should spread the theme props instead so we can get states/variants later.

cixzhang:
Closing this one, and it's a design call rather than a problem with the PR.

The need you identified is real and correctly diagnosed: the dropdown search had no stable seam, so a theme could only reach it through a structural descendant selector against hashed classes — the unsupported tier. And your motivating case, *"render the search borderless and flush in the popover (a command-palette look) rather than as a bordered field,"* is exactly right about how it should look.

**That's why it shouldn't be a theme target.** When the answer to "what does this seam unlock?" is a look every consumer wants, the design should converge instead of being made themeable (T27 / principle 7). A bordered field nested in an already-bordered elevated panel is a box inside a box — that's a bug in the default, not a preference to expose.

So [#4928](https://github.com/facebook/astryx/pull/4928) makes it the default: the search row is borderless and flush, with a leading magnifier, the shared clear button, and a hairline above the options. Mechanically it also lands on top of this PR — the nested `TextInput` is gone, replaced by a shared `PanelSearchInput` primitive used by both selectors, so `selector-search-input` on the `TextInput` root would point at an element that no longer renders. #4928 carries `astryx-selector-search` and `astryx-multi-selector-search` on the new row, which is the same seam under a name that survives the change.

Two things from here that I'd genuinely like from you:

- Once #4928 lands, if `::placeholder` still isn't reachable through the new targets, that's a real gap and worth a small PR on top. Your example made the case for it and I didn't see it covered.
- The composition detail you worked out — a target that **composes** with the component's own class through `mergeProps` rather than replacing it, so same-element rules in `@layer astryx-theme` reach the surface — is the right pattern and isn't written down anywhere. If you want to add it to Theming Infrastructure, I'll review it.

Sorry for the collision; #4928 was in flight while this was up and I should have said so sooner.

PR #4880, opened 2026-08-11
SideNav: fix the blocking findings from the component audit (52.6 → 79.1)

cixzhang:
## What this is

A hardening pass over `packages/core/src/SideNav/` driven by the [Component Audit Rubric](https://github.com/facebook/astryx/wiki/Component-Audit-Rubric). Every finding was re-verified against the source before anything was touched; several did not hold up as reported, and those are called out below rather than papered over with a fix.

**Score: 52.6 (F) → 79.1 (C)**, graded under rubric **v1.2**, mode O. Three deltas make up the difference and they are kept apart on purpose: **76.7 → 77.1** from two rubric corrections (a change of ruler, no code), **77.1 → 79.1** from this round's focus-ring fix, and a standing **+1.2 sensitivity to the D12 sanction → 80.3 / B with zero BLOCKs open**, also a ruler change. Only the middle one measures this round's code.

> **On comparing this to the recorded score.** The ledger records SideNav at **48.0 / F under v1.1**. That number is *not* the baseline used here, for two reasons: v1.2 changed the scoring rules (graded BLOCK ceilings with the new 2.5 half-step, the §5a/§5b split, `limited` re-keyed to vacuous passes, the owner tiebreak and shared-module routing), and the recorded row leaves **nine of its eleven sections `unpublished`** with `score: null` — only `design_objective` and `design_rendered` carry a number, so 48.0 cannot be reconstructed from what was stored. The 52.6 above is a **fresh v1.2 grading of the unmodified code at the merge base**, produced by the same pass that produced the after-score, so the delta measures this diff and not a change of ruler. Read it as *"as recorded under 1.1: 48.0 · same code freshly graded under 1.2: 52.6 · after round one under 1.2: 73.8 · after round two under 1.2: 76.7 · after two rubric corrections: 77.1 · after round three under 1.2: 79.1"*.

It landed in three rounds: the audit fixes; the `useMenuHover` adoption (C12) the first round deferred pending a decision on its behaviour change; and now the shared focus ring (A15), which was deferred waiting for [#4654](https://github.com/facebook/astryx/pull/4654) to land the one definition to adopt. #4654 merged at 06:29 UTC today as `041a0ef49`, and this branch is rebased onto it.

**Rounds one and two changed no pixels** — all 32 before/after pairs were byte-identical, which was the evidence that behavioural and semantic fixes had not disturbed the rendering. **Round three deliberately does change pixels**, in exactly one state: the focus ring. It has its own 32 before/after pairs and its own measurements, below.

---

## ⚠️ Behaviour change: hover on collapsed SideNav items

**If you review one thing, review this.** A collapsed `SideNavItem` with children now opens its submenu flyout through the shared `useMenuHover` hook instead of its own timers, and three things about hovering that flyout are deliberately different afterwards:

| | Before | After |
|---|---|---|
| Pointer that cannot hover (touch) | `mouseenter` synthesised by a tap or a passing finger scheduled a hover-open; the flyout could appear with no tap | hover is gated on `(hover: hover)` — nothing is scheduled. Tapping still opens it, through click, unchanged |
| `mouseleave` on a flyout that was opened **by clicking** | closed it after 200 ms | leaves it open. A click is a commitment; only what hover opened is closed by hover |
| Clicking a collapsed item to dismiss its flyout | the stationary pointer's fresh `mouseenter` re-opened it ~150 ms later | the next `mouseenter` after a click-close is skipped once. A deliberate leave-and-return still opens it |

**Unchanged:** the 150 ms open / 200 ms close hover-intent delays, hover-open still skips auto-focus (focus stays on the trigger), and every keyboard path — Enter/Space to open, focus into the flyout, Escape to dismiss, focus restored to the trigger. On a hybrid laptop (touchscreen **and** a mouse) `(hover: hover)` matches, so hover keeps working there; the gate only removes hover from devices that have no hover to read.

Consumers who relied on "the flyout closes when the pointer leaves, however it was opened" will see the click-opened case behave differently. That is the intended correction, not a side effect.

---

## ⚠️ Shared-module change: the blast radius is wider than SideNav

Round three touches `NavItem/navItemStyles.stylex.ts`'s consumers, not just SideNav. That is the *point* — A15 was advisory against SideNav precisely because the defect was shared — but it means this PR now changes the focus ring of five components:

| Component | Surface | Before | After |
|---|---|---|---|
| `SideNavItem` | every row: link, button, collapsed icon-only, collapsed submenu trigger, split-action link, split-action chevron | UA `outline: auto 1px` | shared ring |
| `SideNavHeading` | whole-heading link, heading icon link, inline heading link, menu-trigger chevron, collapsed link, collapsed menu trigger, popover heading rows | UA `outline: auto 1px` | shared ring |
| `TopNavMenu` | **drawer mode only** — section header button, section item links | UA `outline: auto 1px` | shared ring |
| `TopNavMegaMenu` | **drawer mode only** — section header button | UA `outline: auto 1px` | shared ring |
| `TopNavMegaMenuItem` | **drawer mode only** — item row | UA `outline: auto 1px` | shared ring |
| `TopNavItem` | drawer + desktop | already the shared ring (#4654) | **untouched** |

No consumer drew its own ring, so there is no double ring anywhere — checked by measuring every one of them before touching anything, not by reading the code. `TopNavItem` is the control that proves it: it adopted the utility in #4654, it composites the same `navItemStyles.item`, and its measured ring is byte-for-byte what the other four now produce.

Desktop `TopNavMenu` / `TopNavMegaMenu` / `TopNavMegaMenuItem` are **not** affected — those paths never used `navItemStyles` and already had the ring.

---

## Scorecard — before (merge base, freshly graded under v1.2)

```
Component: SideNav            Package: core
Mode: grading (O)             Rubric: v1.2      Date: 2026-08-10
Commit: merge base of this branch

Visual evidence: 32 screenshots — states: rest, item hover, focus-visible, selected,
                 disabled, collapsed, collapsed+hover (flyout), group expanded,
                 group collapsed, hidden section header, nested, drawer, drawer-content
                 themes: neutral light, neutral dark, y2k

Section                                  W    Score  Weighted  Notes
§1  Accessibility & operable paths       16    2.0/5     6.40   3 BLOCK → ceiling 2
§2  Theming & token integrity            14    3.0/5     8.40   1 BLOCK → ceiling 3
§3  Public API contract                  14    2.5/5     7.00   2 BLOCK → ceiling 2.5
§4  Behavior correctness & states        12    3.0/5     7.20   root cause scored in §7
§5a Design conventions — objective        6    3.0/5     3.60   1 BLOCK → ceiling 3
§5b Design conventions — rendered         4    2.5/5     2.00   scored from pixels
§6  Testing & verification                8    2.5/5     4.00   2 BLOCK → ceiling 2.5
§7  Code health (React & DOM)             8    2.0/5     3.20   2 BLOCK → ceiling 2.5
§8  Docs, Storybook & docsite             8    3.0/5     4.80
§9  i18n & RTL                            5    3.0/5     3.00   1 BLOCK → ceiling 3
§10 Responsive & touch                    5    3.0/5     3.00
                                        ---           ------
TOTAL                                   100            52.60   →  GRADE: F

BLOCKers open: 11 distinct   Distinct defects: 17 (advisory shared-module: 3)
`limited` sections: none — every section cleared the one-third non-vacuous bar
Evidence rider (§11): n/a — existing core component
```

## Scorecard — after

```
Component: SideNav            Package: core
Mode: grading (O)             Rubric: v1.2      Date: 2026-08-11
Commit: head of this branch

Quick checks: Q1 lint:strict [x]  Q2 test [x]  Q3 build [x]  Q4-Q12 [x]
Visual evidence: 32 pairs from round one (unchanged states) + 32 new focus-state
                 pairs from round three — real Chromium, Tab-driven, 2560 and 390,
                 neutral light, neutral dark, y2k

Section                                  W    Score  Weighted  Notes
§1  Accessibility & operable paths       16    4.5/5    14.40   A15 closed by adoption
§2  Theming & token integrity            14    4.0/5    11.20   carried forward
§3  Public API contract                  14    4.0/5    11.20   carried forward
§4  Behavior correctness & states        12    3.5/5     8.40   carried forward
§5a Design conventions — objective        6    3.0/5     3.60   D12 — see the ruler note
§5b Design conventions — rendered         4    3.5/5     2.80   focus row passes; +0.5 ruler
§6  Testing & verification                8    4.5/5     7.20   +12 tests; ruler note below
§7  Code health (React & DOM)             8    4.5/5     7.20   carried forward
§8  Docs, Storybook & docsite             8    3.5/5     5.60   carried forward
§9  i18n & RTL                            5    4.0/5     4.00   carried forward
§10 Responsive & touch                    5    3.5/5     3.50   carried forward
                                        ---           ------
TOTAL                                   100            79.10   →  GRADE: C

BLOCKers open: 1 (D12, under the rubric as it read when this PR was graded)
Distinct defects: 4 (advisory shared-module: 1)   FIXes open: 3   NITs open: 0
SENSITIVITY: apply the D12 sanction (below) → 80.3 / B with zero BLOCKs open
`limited` sections: none
Evidence rider (§11): n/a — existing core component
Unverified: R1 (320 px reflow) — no automated reflow check exists and no narrow-viewport
            story ships; it is what holds §10 at 3.5.
```

**The 76.7 → 79.1 move splits into two unrelated things, and they are scored separately below:** +0.4 from two rubric corrections that landed after the last pass (no code involved) and +2.0 from this round's focus-ring fix. Same discipline as the D12 ruler note.

**Which sections were re-run in round three, and which were carried forward.** Re-run: **§1** (owns A15), **§5b** (the focus row of the state matrix, re-measured from pixels), **§6** (12 new tests) and **§2** (the ring now resolves from `--color-accent`, so the theme-integrity read had to be re-taken — it came back unmoved, see below). Carried forward untouched, because a focus-outline adoption cannot reach them: §3, §4, §5a, §7, §8, §9, §10.

- **§1: 4.0 → 4.5.** A15 was the section's last blocking-class finding against this component. The ring is now `2px solid var(--color-accent)` at the documented 3px offset on every focusable surface in the family, measured in three themes. Held off 5 by A14 — no `forced-colors` fallbacks — which §1's 5/5 anchor names explicitly ("reduced-motion and forced-colors honored").
- **§5b: 2.5 → 3.5**, of which **+0.5 is a ruler change and +0.5 is this diff** (see the two corrections below). The focus-visible row of the state matrix now passes on mechanism *and* signature. Held off 4 by D13 — hover and selected still composite 1.02:1 apart in light, which is exactly the 3/5 anchor's "a selection treatment that doesn't match its family's style".
- **§6: 4.5, no move.** Twelve new tests, all twelve red on the merge base. The section does not move because what held it at 4.5 is unchanged: `rtl:audit` still has no curated `targets.json` entry for SideNav. The other two reasons quoted last round have both gone away — see the rubric corrections.
- **§2: no move, and it was worth re-checking.** The ring reads `--color-accent` through the shared utility, so it now tracks the theme (y2k resolves `rgb(45, 36, 27)`, neutral dark `rgb(235, 235, 235)`). But a *theme target* for the ring is deliberately not part of #4654 — `generateThemeRules` mangles `:focus-visible` into `.focus-visible` and truncates `:has(:focus-visible)`, so a target would only let a theme paint an unconditional outline. Themeability of the ring is a real gap; it lives in the theme layer, not in SideNav, and #4654 is the foundation for closing it.

**Round two's re-runs, unchanged:**

- **§7: 3.5 → 4.5.** C12 was the section's last open finding; the flyout now composes the shared hook, and one Effect (the hand-rolled unmount cleanup) leaves `SideNavItem` entirely. That notify Effect is gone too, since review: `SideNav.tsx` now runs no Effects at all. Held off 5 by the render-phase write to `collapseStateRef`, which the deprecated `handleRef` path still reads.
- **§6: 4.0 → 4.5.** Nine tests over the flyout's pointer contract, three of them red on the previous implementation. Held off 5 by three gaps that are all environmental — *as graded then; two of these three are inadmissible under the corrected §6, see the rubric corrections below*: no VRT exists repo-wide; Tab containment *inside* the flyout cannot be asserted in jsdom (the popover polyfill leaves the content `display: none`, and `user-event`'s tab skips it); and `rtl:audit` still has no curated `targets.json` entry for SideNav.
- **§10: 3.0 → 3.5.** Touch adaptation is what this section grades, and the flyout no longer runs hover logic on a pointer that cannot hover. R1 (320 px reflow) is still unverified, which is what keeps it below 4.
- **§1: no move, and the brief expected one.** The touch gating does **not** close an A8 gap, because there was no A8 gap: A8 fails when a control's *only* path to discovery or activation is hover, and the flyout has always been tap-reachable through click. What the old code had was hover logic firing on a device with no hover — an adaptation defect (§10) rooted in C12 (§7), not a broken accessible path. A8's own note also warns against gating on `hover: none`; the hook gates on `(hover: hover)` in the enabling direction, which is the correct one for hybrid machines.
- **§4: no move, on purpose.** The click-then-spring-back-open path is exactly the kind of thing B12's adversarial pass is for, and it is now fixed with a red-before-green test (B16). But the defect is scored once, at its root in §7, and the rest of §4's read is unchanged — loading/empty/overflow are still unverified and B13 composition is still unshown. Crediting the same fix in a third section would be inflation, so it is cross-referenced and not scored.

### Ruler note — D12 changed after this PR was graded

The [Component Audit Rubric](https://github.com/facebook/astryx/wiki/Component-Audit-Rubric) gained a sanctioned exception to D12 on 2026-08-10: animating `grid-template-rows` between `1fr` and `0fr` is the accepted way to open a disclosure to an unknown height, provided reduced motion is honoured (this PR added that), the subtree is small (a nav group), and it is user-initiated and brief. The exception names `SideNavItem`'s collapsible group as its precedent.

Applied, that closes the last open BLOCK: **§5a 3.0 → 4.0 and the total 79.1 → 80.3 — a B, with zero BLOCKs open.** It is kept out of the headline because it is a change of ruler, not of code, and folding it in would make the delta measure two different things at once. Two notes for whoever owns the rubric: the edit changes a *severity scope*, which by the rubric's own versioning rule is a minor bump, and it still reads `1.2` — so a strict reader would call 79.1 and 80.3 incomparable. Say the word and I will publish 80.3 as the headline instead.

**§5b did not move in rounds one and two, and that was the honest result then** — the visual defects found were deferred, and a hover-intent adoption is not visual. It moves in round three, because the focus ring is a visual defect and it is now fixed; the delta is broken out above.

---

## Findings: verified → fixed or deferred

| # | Rule | Finding as reported | Verified? | Outcome |
|---|---|---|---|---|
| 1 | **A13** | 3 transitions with no `prefers-reduced-motion` guard | ✅ all three | **Fixed** — each `transitionDuration` collapses to `0s` under `prefers-reduced-motion: reduce`, matching the house pattern in `CheckboxInput`/`layerAnimations` |
| 2 | **D12** | animating `grid-template-rows`, a layout property | ✅ | **Deferred** — needs a design decision, see Needs Review |
| 3 | **A16 / I2** | concatenated English AT string `` `${label} submenu` `` | ✅ (at `usePopover({dialogLabel})`) | **Fixed** — `t('@astryx.sideNavItem.submenuLabel', {label})`, ICU placeholder, key added to `en.json` with a description |
| 4 | **A17** | hand-rolled `VisuallyHidden` | ✅ but **not where reported** | **Fixed** — it is in `SideNavSection`, an inline `style` clip block, now `<VisuallyHidden as="div">` |
| 5 | **T6** | style-driving prop missing from `themeProps` | ✅ (`isDisabled`) | **Fixed** — `disabled: isDisabled ? 'disabled' : null` on both call sites, `states` updated in the doc so `themingTargets.test.ts` stays green |
| 6 | **T2** | themeable surface dropped in the drawer modes | ⚠️ **mis-framed** | **Partly fixed** — see "Where the report was wrong" |
| 7 | **P2** | `...rest` never captured on `SideNavItem` | ✅ | **Fixed** — captured and spread on the interactive control in all three render paths, before the owned `aria-*`/`data-testid`/`role` so precedence stays deliberate (P3) |
| 8 | **B15 / C11** | stale external collapse button, state read from a ref in render | ✅ — one defect, not two | **Fixed** — a `SideNavCollapseButton` rendered outside the sidenav takes the same controlled `collapsible` config SideNav gets, so the state the consumer already owns reaches it as a prop. `handleRef` behaves as it did before this PR and is deprecated in its favour; `SideNavImperativeCollapseHandle` is unchanged. *(An earlier revision added `subscribe`/`useSyncExternalStore` to the handle; [that was the wrong shape](https://github.com/facebook/astryx/pull/4880#discussion_r3787881493) and it is gone.)* |
| 9 | **C19** | uncleaned timers | ✅ | **Fixed** — the hover-intent timers are cleared on unmount |
| 10 | **C12** | hand-rolled hover intent | ✅ — hook exists, but severity is **FIX**, not BLOCK | **Fixed** — `useMenuHover` adopted; hover on collapsed items changes, see the callout at the top |
| 11 | **V6 / V7** | zero keyboard tests, zero focus tests | ✅ | **Fixed** — 8 new tests: tab order, disabled skipped, Enter/Space on the group toggle, Enter on a button item, Escape closing the flyout, focus kept on the toggle through a collapse, focus not stranded in an `inert` group, focus restored to the trigger |
| 12 | **X5a** | `size`/`as` doc prop drift | ❌ **false positive** | **Fixed as X5b instead** — `as` is documented *and* present; `size` was the reverse direction. Added to all three doc variants |
| 13 | **A10** | *not reported* | ✅ found by measurement | **Fixed** — the split-action control was 20 px tall inside a 32 px row, under the WCAG 2.5.8 floor; `alignSelf: 'stretch'` fills the row |
| 14 | **A2 / X16** | *not reported* | ✅ | **Fixed** — the APG pattern was implemented but named nowhere; the docs now name the Disclosure pattern and carry accessibility guidance |
| 15 | **A15** | *not reported (in the ledger's notes)* | ✅ | **Fixed by adoption (round three)** — the browser-default ring is replaced by the shared `focusOutlineProps.focusVisible` from #4654, on SideNav and on the four other `navItemStyles` consumers |
| 16 | **D13** | *not reported (in the ledger, absent from the brief)* | ✅ reproduces exactly | **Deferred, advisory** — hover and selected composite 1.02:1 apart in light |
| 17 | **A14** | *not reported* | ✅ | **Deferred** — no `forced-colors` fallbacks; FIX severity for a shipped component |

---

## Where the report was wrong

Asked for explicitly, so stated plainly.

1. **X5a does not reproduce.** `as` is documented *and* present in `SideNavItemProps`, so there is no phantom prop; `docPropReferences.test.ts` would have been red if there were. The real defect runs the other way — `size` exists in source and was missing from the docs — which is **X5b**, severity FIX, and the rubric's *Not yet in force* table says X5b has no enforcer. Reporting it as an X5a BLOCK inflated the count by one blocker.

2. **A17 is in `SideNavSection`, not `SideNavItem`.** `SideNavItem` does have a `<span hidden>` used as an `aria-labelledby` target, but that is correct — a hidden element referenced by `aria-labelledby` still contributes its name, and it should *not* be visually hidden-but-rendered. The actual hand-roll is `SideNavSection`'s inline `position/width:1/clip: rect(0,0,0,0)` block.

3. **T2 is mis-framed, and the fix is a §3 fix.** No themeable *target* was dropped in the drawer modes. In `drawer` mode `MobileNav` paints that surface and already carries its own `mobile-nav` target; stacking `side-nav` on the same node would create exactly the specificity conflict **T28** tells you to skip and flag. In `drawer-content` mode SideNav returns a fragment and paints nothing at all, so it owns no element that could carry a target. What *is* genuinely dropped is the consumer's `xstyle` / `className` / `style` / `...rest` — a **P2/P3** defect, not a theming one. Fixed for `drawer` (forwarded to `MobileNav`, which merges them properly); `drawer-content` is left open because the only way to carry them is a new wrapper element, which is a T14 style-only wrapper and a layout change.

4. **C12's severity is FIX, not BLOCK.** `useMenuHover` does exist and is used by `SideNavHeading` in the same folder (and by `TopNavMenu`, `TopNavHeading`, `DropdownMenuSubMenu`), so the finding is real. But C12 reads "BLOCK for a11y primitives; FIX otherwise", and hover intent is not one. It is now adopted, and two things about the adoption were not what they looked like: the hook is **not** a drop-in for the whole surface (its keyboard half is wrong for this flyout — see below), and the touch improvement is **not** an A8 fix (the flyout was always tap-reachable).

5. **A13 and D12 are two defects at one site, and after this diff only one of them is left** — and it moves section. With the guard in place the remaining defect at `childrenCollapsible` is purely "a layout property is animated", which **§5a** owns, not §1.

6. **B15 and C11 are one defect.** The stale button *is* the render-phase ref read. Per "consequences are scored at the root" it is scored once, in §7, and cross-referenced from §4.

7. **Five findings were missing from the brief**, three of them blocking-class: A10 (a 20 px interactive target), A2/X16 (the APG pattern named nowhere), A15 (browser-default focus ring — this one *is* in the ledger's `notes`, and it is fixed in round three), D13 (hover ≈ selected — this one *is* in the ledger's `blocks.open`), and the pressed state missing its `scale(0.98)` signature. **That last one has since been withdrawn as a false positive** — the rubric now says `scale(0.98)` is the *button* press treatment and a nav row takes the overlay alone; see the rubric corrections below.

8. **The ledger row itself is thin.** It records `blocks.count: 12` but lists only three of them, and nine of eleven sections are `unpublished`. Worth knowing before anyone treats 48.0 as a measured baseline.

---

## Evidence

**Red before, green after — round one.** Six of the first round's tests fail on the merge base and pass on this branch — verified by checking the base's source files back out under the new test file:

```
× translates the collapsed submenu dialog name                     (I2)
× forwards unrecognised props on SideNavItem through to the control (P2)
× forwards unrecognised props on a collapsed SideNavItem            (P2)
× exposes a disabled item as a theming state                        (T6)
× keeps a SideNavCollapseButton outside the sidenav in step with it (B15/C11)
× does not open the hover popover after the item unmounts           (C19)
⎯ 6 failed | 111 passed (117)      → on this branch: 117 passed (117)
```

The i18n test asserts under the pseudo locale on purpose: a concatenated `` `${label} submenu` `` produces the same English string as the catalog does, so only a locale swap can tell the two apart. The timer test asserts `vi.getTimerCount() === 0` after unmount, which is the actual C19 contract.

### The `useMenuHover` round — what changed and what it cost

`SideNavItem`'s hand-rolled hover intent, its `clearPopoverTimeouts` helper and its unmount-cleanup Effect are gone, replaced by the hook's `triggerProps` and the pointer half of its `contentProps`. Net −18 lines of source. The two delays it used (150 ms / 200 ms) are the hook's defaults, so no timing changed. `skipAutoFocus` on the hover path survives the move: the hook calls `show({skipAutoFocus: true})` from its hover timer and a bare `show()` from its click handler, which is exactly the split this component wants — and there is now a test pinning it.

**The one thing that was not adopted.** `useMenuHover` also returns `menuRef` and `contentProps.onKeyDown`, which are a `useListFocus` over `[role="menuitem"]`. This flyout has no menu items: it renders `SideNavItem`s as `<a>`/`<button>` inside a focus-trapped `role="dialog"`, and its keyboard behaviour comes from `usePopover`'s focus trap. Wiring that half would have made every arrow key inside the flyout `preventDefault()` with no item to move to — trading a hover bug for a keyboard bug. So only the pointer half is wired, which is exactly what `DropdownMenuSubMenu` does with the same hook and for the same reason; `SideNavHeading` adopts it whole because its popup really does contain a `role="menu"`. **No keyboard behaviour changed in this round**, and two tests hold that line: arrow keys inside the flyout stay unprevented and do not move focus, and Enter-open still lands focus on the first item with Escape restoring it to the trigger.

**Known cost, disclosed:** the hook calls `useMediaQuery` unconditionally, before its own `isEnabled` short-circuit, so every `SideNavItem` — including expanded ones that will never hover-open anything — now holds one `matchMedia` listener. It is cheap and every other consumer of the hook already pays it, but `SideNavItem` is the first consumer that can appear thirty times on a page. If that ever matters, the fix belongs in the hook (skip the query when `isEnabled` is false), not here.

**Red before, green after — this round.** Three of the nine new tests fail on the previous implementation and pass on this one (`SideNavItem.tsx` reverted under the new test file):

```
× keeps a click-opened flyout open when the pointer leaves
× does not spring back open when a click dismisses it under a stationary pointer
× schedules no hover timers on a coarse pointer
⎯ 3 failed | 122 passed (125)      → on this branch: 125 passed (125)
```

The other six pin behaviour that must *not* change: the 150 ms open delay, `mouseleave` closing a hover-opened flyout, hover-open leaving focus on the trigger, arrow keys reaching the browser, the keyboard open/restore path, and the C19 timer-cleanup contract. That last one is the previous round's "does not open the hover popover after the item unmounts" test — it still asserts `vi.getTimerCount() === 0` after unmount, unchanged except that the hook now owns the cleanup it is checking.

Each hover test stubs `matchMedia` for the modality it is testing, because the shared setup polyfills a stub that matches nothing — which reads as a coarse pointer and would have made every hover assertion vacuously true.

**No new screenshots, and that is deliberate.** This round changes event handlers only: no style module, no token, no JSX structure, no attribute. The rendered DOM of a collapsed item and its flyout is identical before and after, so a second set of 32 byte-identical pairs would be evidence of nothing. The 32 pairs from the first round still stand. What *is* visible to a user — when the flyout opens and closes — is behaviour, and it is covered by the tests above rather than by pixels.

### Round three — the shared focus ring (A15)

**What changed.** Every focusable surface in the nav family that was styling no focus state now composes `focusOutlineProps.focusVisible` from `utils/focusOutline.stylex` — the module #4654 added. Six source files, +25/−14 lines of source.

The previous round's proposal in this PR was *"add the ring to `navItemStyles` with a new `--nav-item-focus-offset` var registered in `derivedVarRegistry.ts`."* **That is superseded and was not built.** #4654 settled both halves of it: there is one definition of the ring, and the offset is `3px` from Design Conventions §User Interaction States, held in the shared module rather than in a per-family var. A new var would have been the sixth definition of a thing that had just been consolidated down to one.

**Where the fix went, and why not into `navItemStyles.item`.** The finding was routed to `navItemStyles` because that is the shared file SideNav composites. But the ring cannot live *inside* `navItemStyles.item` without restating the outline longhands there — `stylex.create` cannot compose another created style into a definition — which is exactly the duplication #4654 removed. The house adoption pattern from #4654 is a call-site one (`focusOutlineProps.focusVisible(...)`, 26 components), and all four `navItemStyles` consumers that #4654 already touched use it that way. So the fix is applied at the call sites, on whichever element actually takes focus, and `navItemStyles`' file header now states the rule so the next person does not re-derive it.

**Every `navItemStyles` consumer, checked before and after.** Measured by pressing real `Tab` keys in Chromium and reading `getComputedStyle` on `document.activeElement` — not by reading the source:

| Consumer | Call site | Focusable? | Before | After |
|---|---|---|---|---|
| `SideNavItem` | expanded row (`<a>`/`<button>`) | yes | `1px auto` UA | `2px solid` accent @3px |
| `SideNavItem` | collapsed icon-only row | yes | `1px auto` UA | `2px solid` accent @3px |
| `SideNavItem` | collapsed submenu trigger | yes | `1px auto` UA | `2px solid` accent @3px |
| `SideNavItem` | split-action row `<div>` | **no** — a container | (n/a) | stays unringed, on purpose |
| `SideNavItem` | split-action primary link | yes | `1px auto` UA | `2px solid` accent @3px |
| `SideNavItem` | split-action chevron toggle | yes | `1px auto` UA | `2px solid` accent @3px |
| `SideNavHeading` | collapsed link · collapsed menu trigger | yes | `1px auto` UA | `2px solid` accent @3px |
| `TopNavItem` | drawer row | yes | **already adopted in #4654** | unchanged |
| `TopNavMenu` | drawer section header | yes | `1px auto` UA | `2px solid` accent @3px |
| `TopNavMenu` | drawer item link | yes | `1px auto` UA | `2px solid` accent @3px |
| `TopNavMegaMenu` | drawer section header | yes | `1px auto` UA | `2px solid` accent @3px |
| `TopNavMegaMenuItem` | drawer item row | yes | `1px auto` UA (by construction; not separately captured before — the red-before-green unit test is the evidence) | `2px solid` accent @3px |

**No double rings.** The check that mattered: does any consumer already draw its own ring, so that adopting the shared one would paint two? Measured before touching anything — **no**. Only `TopNavItem` had a ring at all, and it is #4654's own, so it is left alone. The other four drew the browser's.

`SideNavHeading` also had UA-ringed surfaces that are not `navItemStyles` consumers but are unambiguously part of SideNav — the whole-heading link, the heading icon link, the inline heading link, the menu-trigger chevron and the popover heading rows. They are fixed too, because A15 reads "the focus ring is visible in **every** theme", and a component that rings six of its eleven focusables has not closed it.

**Measured ring geometry — the numbers are the evidence.** I cannot look at an image, so the verification here is measurement, and it is stated as such: real Chromium, real `Tab` keypresses (no injected classes, no `.focus()`), `getComputedStyle` on the focused element, at **2560×1400** and **390×844**, in **neutral light**, **neutral dark** and **y2k**. 124 focus stops measured after the change; 123 of them draw the ring below and the 124th is `TopNavHeading` (see Needs Review 7).

```
                                    before                          after
SideNav row (2560, light)   1px auto  rgb(0,95,204)  @1px   2px solid rgb(38,38,38)    @3px
SideNav row (2560, dark)    1px auto  rgb(153,200,255) @1px 2px solid rgb(235,235,235) @3px
SideNav row (2560, y2k)     1px auto  rgb(0,95,204)  @1px   2px solid rgb(45,36,27)    @3px
SideNav row (390,  light)   1px auto  rgb(0,95,204)  @1px   2px solid rgb(38,38,38)    @3px
collapsed icon item 32×32   1px auto  rgb(0,95,204)  @1px   2px solid rgb(38,38,38)    @3px
group toggle 244×32         1px auto  rgb(0,95,204)  @0px   2px solid rgb(38,38,38)    @3px
split-action link 196×32    1px auto  rgb(0,95,204)  @1px   2px solid rgb(38,38,38)    @3px
split-action chevron 24×24  1px auto  rgb(0,95,204)  @0px   2px solid rgb(38,38,38)    @3px
heading link 244×32         1px auto  rgb(0,95,204)  @1px   2px solid rgb(38,38,38)    @3px
heading chevron 28×28       1px auto  rgb(0,95,204)  @0px   2px solid rgb(38,38,38)    @3px
drawer row 303×32 (390)     1px auto  rgb(0,95,204)  @1px   2px solid rgb(38,38,38)    @3px
TopNavMenu drawer hdr 303×32 1px auto rgb(0,95,204)  @0px   2px solid rgb(38,38,38)    @3px
TopNavMegaMenuItem 303×32   (see table)                     2px solid rgb(38,38,38)    @3px
```

The three things to read off that: the UA ring is **1px** where the house ring is 2px; its **offset varied by element** (1px on links, 0px on buttons) where the house ring is a uniform 3px; and its **colour did not move between themes** — `rgb(0, 95, 204)` in y2k is the same blue as in neutral light, because it is Chrome's, not Astryx's. The after column tracks `--color-accent` per theme, which is the actual substance of A15.

**Cross-checked against components that adopted in #4654**, measured in the same run, so "matches the house treatment" is a comparison and not an assertion:

```
TabList tab            2px solid rgb(38,38,38) @3px   ← adopted in #4654, untouched here
TopNavItem (desktop)   2px solid rgb(38,38,38) @3px   ← adopted in #4654, untouched here
TopNavMenu trigger     2px solid rgb(38,38,38) @3px   ← adopted in #4654, untouched here
Button (nav footer)    2px solid rgb(38,38,38) @3px   ← adopted in #4654, untouched here
Link (superheading)    2px solid rgb(38,38,38) @3px   ← adopted in #4654, untouched here
SideNavItem row        2px solid rgb(38,38,38) @3px   ← this diff
```

**Screenshots — 32 new before/after pairs** at the focus state, both viewports, three themes, both `SideNav` and the other consumers, captured from two separate `@astryxdesign/core` builds (Storybook resolves the package to `dist/`). Pixel diffs come out where they should and nowhere else:

```
sidenav-default        light 2560   0.066%      appshell drawer   light 390   0.861%
sidenav-default        dark  2560   0.079%      appshell drawer   dark  390   1.040%
sidenav-default        y2k   2560   0.064%      topnav drawer     light 390   0.406%
sidenav-collapsible    light 2560   0.032%      topnav drawer     dark  390   0.473%
sidenav-collapsed      light 2560   0.020%      sidenav-header-menu light   0.065%

controls — same capture, focus on an element this diff does not touch:
TopNavItem focused (ref-topnavmenu)            0.000%
Link focused (sidenav-suite-header)            0.000%
Button focused (sidenav-collapsed footer)      0.000%
MobileNav close button focused                 0.000%
TabList (tab 2 = focus left the tab)           0.015%  ← capture flake, not this diff:
                                                        re-shooting the *same* build twice
                                                        gives 0.015% on the same frame (an
                                                        in-flight indicator transition)
```

The 0.000% controls are the interesting half: they say the change is confined to the elements that were meant to move.

**What did not fit #4654's utility cleanly — the split-action row.** A collapsible item that also has an `href` renders a `<div>` row holding two independent tab stops, the link and the chevron toggle. None of the utility's four shapes is a clean fit:

- `focusVisible` on the row is inert — the row never takes focus.
- `focusWithin` on the row rings the whole row whenever *either* child is focused, so Tab from the link to the chevron paints an identical outline twice in a row and you cannot tell which control you are on. Ringing the chevron as well would then paint two outlines at once.
- `publishFocusVisibleVars` + `focusWithinOrPublished` — TreeList's pattern — needs the focusable to be the *ancestor* of the element that draws (CSS variables inherit downward). Here the focusables are the row's children, so the vars cannot travel up.

So each child rings itself with `focusVisible`, and the row stays unringed. The cost, stated plainly: the link's ring hugs the link box (196×32 inside the row's 8px inline padding) rather than the row's full 244×32, so a focused split-action row looks slightly inset compared with a focused plain row. Both are the same 2px accent ring at 3px offset; only the box differs. The alternative — a row-level ring — trades that for genuine ambiguity between two tab stops, which is a worse a11y outcome than a 8px inset. A test pins the arrangement, including that the row wrapper does *not* carry the ring.

**One more thing #4654 does not give us, and it matters for how A15 is worded.** The PR title says "make shared focus outlines themeable"; its body is explicit that the shared `astryx-focus-outline` *theme target* was pulled, because `generateThemeRules` mangles pseudo-selector keys. So after this diff the ring **resolves from `--color-accent`** and therefore follows the theme's accent — which is what the A15 finding actually measured as broken — but a theme still cannot restyle the ring *as a state* through a target. That is a theme-layer gap tracked by #4654's own follow-ups, not a SideNav one, and §2 is scored accordingly (no move).

**Tests — 12 new, all 12 red on the merge base.** Reverting only the six source files under the new test files:

```
× draws the shared ring on a link item                                (SideNav)
× draws the shared ring on a button item                              (SideNav)
× draws the shared ring on a collapsed icon-only item                 (SideNav)
× draws the shared ring on a collapsed submenu trigger                (SideNav)
× rings each focusable of a split-action row, and not the row itself  (SideNav)
× draws the shared ring on a heading rendered as one link             (SideNav)
× draws the shared ring on a collapsed heading link                   (SideNav)
× draws the shared ring on the heading menu trigger                   (SideNav)
× draws the shared ring on the drawer section header                  (TopNavMenu)
× draws the shared ring on a drawer item                              (TopNavMenu)
× draws the shared ring on the drawer section header                  (TopNavMegaMenu)
× draws the shared ring on a drawer item                              (TopNavMegaMenu)
⎯ 12 failed | 158 passed (170)     → on this branch: 170 passed (170)
```

What they assert is deliberately modest, and the comment in the file says so: jsdom does not derive `:focus-visible` from `user-event`'s Tab on an `<a>`/`<button>`, so the painted ring cannot be read back in a unit test — #4654's own review is the cautionary tale, where the tests asserting classes and published variables stayed green while two components had lost their ring entirely. So these pin the *composition* (the focusable element carries the classes the shared utility compiles to, compared against `stylex.props(focusOutlineStyles.focusVisible)` rather than a hardcoded hash), and the **paint is verified in the browser**, above.

**Screenshots — 32 before/after pairs, real Chromium against Storybook.** States: rest, item hover, focus-visible, selected, disabled, collapsed sidebar, collapsed + item hover (flyout), group expanded, group collapsed, hidden section header, nested items, drawer, drawer-content. Themes: neutral light, neutral dark, y2k (rest states). Both sides built from their own `@astryxdesign/core` `dist/`, since Storybook resolves the package to `dist/`.

> **All 32 pairs are byte-identical (0.000% differing pixels, maxΔ 0).** That is the intended result and it is the load-bearing evidence here: none of the fixed defects were visual, so the pair proves the corrections did not disturb the rendering — including the `alignSelf: stretch` target-size fix, whose element paints nothing.
>
> That the two builds really are different is shown at the DOM level rather than the pixel level — on the disabled-item story the merge base renders `class="astryx-side-nav-item md"` with no `data-disabled`, and this branch renders `class="astryx-side-nav-item md disabled"` with `data-disabled="disabled"`.

The 64 PNGs and three labelled contact sheets (`contact-sheet__{light,dark,y2k}.png`) are captured and kept together; they are not committed, since the repo has no home for audit media and a 2 MB binary drop is not one. Say the word and I will attach them to the thread.

**§5b state-visual conformance matrix** (mechanism judged, not pixels matched; `getComputedStyle` composited through the real ancestor stack as corroboration):

| State captured | Approved representation it claims | Token signature present? | Verdict |
|---|---|---|---|
| rest | — (base) | `transparent` over the nav surface | pass |
| hover | Hovered — Overlay style | `--color-overlay-hover` alpha tint over base: light `#ffffff → #f2f2f2`, dark `#262626 → #313131`; genuinely alpha, not an opaque fill | pass |
| pressed | Pressed | `--color-overlay-pressed` present; `scale(0.98)` absent, which is **correct for a nav row** — see the rubric correction | pass |
| selected | Selected — surface fill, backed by `aria-current="page"` | `--color-neutral` fill; ARIA backing ✓ | pass on mechanism, **fail in light on distinctness** (below) |
| hover vs selected | must read as different states | light **1.023:1** · dark 1.176:1 | **fail in light — advisory** (D13) |
| focus-visible | Focused — Outline style, offset, visible in every theme | `2px solid var(--color-accent)` at `3px` offset, resolving per theme: light `rgb(38, 38, 38)`, dark `rgb(235, 235, 235)`, y2k `rgb(45, 36, 27)` | **pass** (was the A15 failure) |
| disabled | Disabled | native `disabled` + `aria-disabled`, muted text | pass |
| collapsed (icon-only) | — | tooltip/flyout affordance, name preserved via `aria-label` | pass |
| drawer / drawer-content | — | renders inside `MobileNav`; `mobile-nav` target reachable | pass |

**One failure left, down from three.** The focus row is fixed at the root in round three; the pressed row was withdrawn when the rubric corrected it. What remains is D13 — hover and selected 1.02:1 apart in light — which still lives in the shared `NavItem/navItemStyles.stylex.ts` that `SideNavItem`, `SideNavHeading`, `TopNavItem`, `TopNavMenu`, `TopNavMegaMenu` and `TopNavMegaMenuItem` all composite. Per the rubric's shared-module rule it stays **advisory** against SideNav. It is not fixed here because, unlike the ring, it is an open visual-design decision (see Needs Review) rather than an adoption of something already decided.

**Other checks**

- `pnpm a11y:audit -- --components SideNav` — 14 stories, **0 violations**, `0 new / 0 resolved`, baseline untouched (A18, V9).
- `pnpm rtl:audit -- --filter SideNav` — 0 fail, 0 surprise. Honest caveat: **14 of 14 pointer-metric checks came back N/A** because SideNav has no curated `targets.json` entry, so this run is not real coverage. Logged as a §9 evidence gap (FIX), not a defect.
- Reduced motion confirmed in the built stylesheet, not just the source: `@media (prefers-reduced-motion: reduce){…transition-duration:0s}`.
- Target sizes measured in the browser: every interactive element is now ≥24×24 (the chevron toggle is exactly 24×24; the split action went 196×20 → 196×32).

---

## Two rubric corrections since the last pass, scored separately

Both landed in the wiki after this PR was last graded, both are changes of ruler rather than of code, and neither is mixed into the focus-ring delta. Together they are **+0.4: 76.7 → 77.1**.

**§5b — `scale(0.98)` is the *button* press treatment ([`2219b27`](https://github.com/facebook/astryx/wiki/Component-Audit-Rubric), "State visuals: scale(0.98) is the button press treatment").** The rubric now reads *"`scale(0.98)` is the **button** treatment — a self-contained control that depresses. Do not require it of other archetypes: a nav row, list item, table row or menu item takes the overlay alone, and shrinking a full-width row on press is wrong, not missing."* The pressed row of the §5b matrix was reported as **fail — advisory** on exactly that missing transform. **It was a false positive and is withdrawn.** SideNav rows are nav rows; they carry `--color-overlay-pressed` and nothing else, which is now the documented correct treatment. The matrix row above is updated to `pass`, and the finding is removed from the count. **§5b 2.5 → 3.0.**

**§6 — missing repo tooling is not a deduction ([`777c51e`](https://github.com/facebook/astryx/wiki/Component-Audit-Rubric), "Testing: compare visuals against Design Conventions; missing tooling is not a deduction").** *"That is a fact about the harness, **not a deduction against a component** — a component cannot be marked down for tooling the repo does not have… the absence of a visual-regression suite, or of any other tooling the repo has not built, never holds a section off full marks."* Last round held §6 at 4.5 on three gaps, and **two of them are now inadmissible**: "no VRT exists repo-wide" and "Tab containment inside the flyout cannot be asserted in jsdom (the popover polyfill leaves the content `display: none`)" are both missing tooling. The third is not: `rtl:audit` **does** exist and SideNav simply has no curated `targets.json` entry, which is a curation gap this component could close. **So §6 stays at 4.5** — the correction removes two of three reasons but not the binding one. Adding the `targets.json` entry would take §6 to 5.0 (+0.8); it is a separate change to a separate file and I have not smuggled it in here.

The revised V15 also asks that a visual change be compared against Design Conventions through §5b's matrix rather than pixel-diffed for its own sake. Round three is a visual change, and that is how it is graded above: the focus row of the matrix, judged on mechanism and token signature.

---

## Needs Review — decisions that are not mine to make

**1. D13 — hover and selected are 1.02:1 apart in light.** Measured composites: rest `#ffffff`, hover `#f2f2f2` (`--color-overlay-hover`, α .05), selected `#f0f0f0` (`--color-neutral`, α .06). Hovering any row makes it look like the current page. Dark is fine (1.18:1). Both tokens are correct for their roles per T22, so this is not a wrong-token pick — it is that a 5% and a 6% black tint cannot express two different states. *My proposal:* move selection off "surface fill" to one of the other approved selection styles for nav — the edge/rail marker reads at a glance and does not compete with an overlay tint — or give selected a distinctly weightier fill. That is a visual design decision on a shared style module, and it is the last open item in the §5b matrix.

**2. `collapsed` is not a theming state.** The collapsed icon-only row is visually distinct but a theme cannot address it — `themeProps('side-nav-item', …)` carries `size`, `selected` and now `disabled`, but the collapsed variant is selected from render context, not a prop. Adding `collapsed` would be additive and safe; I left it out because it is new public theming surface, which is a call, not a correction.

**3. `drawer-content` still drops the consumer's style props.** See point 3 in "Where the report was wrong" — the only fix is a wrapper element, which trades a P2 defect for a T14 one.

**4. `data-testid` sits on the row wrapper in the split-action path** (P29 says it belongs on the primary interactive element). Pre-existing, and moving it would break any consumer selecting on it today, so it wants a deprecation, not a silent move.

**5. A14 — no `forced-colors` fallbacks.** The selection fill and hover tint are background colours, which vanish under `forced-colors: active`, leaving selection unexpressed. FIX severity for a shipped component, and now the single thing holding §1 off 5. Same owner as 1 — `navItemStyles`.

**6. The collapsed flyout is a dialog of links, and arguably wants to be a menu.** Surfaced by the C12 work rather than fixed by it. `useMenuHover`'s keyboard half drives a `useListFocus` over `[role="menuitem"]`; this flyout renders `SideNavItem`s as plain links and buttons inside a focus-trapped `role="dialog"`, so that half is deliberately left unwired (see "The one thing that was not adopted"). Giving the flyout real menu semantics would make arrow-key navigation work and let the hook be adopted whole — but it changes the announced role of a shipped surface and the DOM of every nested item, so it is its own PR with its own audit, not a rider on this one.

**7. `TopNavHeading`'s heading link still draws the UA ring** — `1px auto rgb(0, 95, 204)`, measured on `core-topnavmenu--default`. It is the one focusable left in the nav family that #4654 did not reach, and it is **not** fixed here: it does not consume `navItemStyles`, so it is outside this diff's blast radius, and widening the radius on the way past is how a focused PR stops being reviewable. One line, whenever someone wants it.

*Resolved since the first round:* **D12** (the collapse animation) — the rubric now sanctions the `grid-template-rows` disclosure and names this component as the precedent; see the ruler note above. **C12** (adopt `useMenuHover`) — approved and done in round two. **A15** (the focus ring) — #4654 landed the definition to adopt; done in round three.

## Why this is one PR

Round three (the focus ring) is here for the same reason round two was: A15 is a finding *of this audit*, it was deferred inside this PR with a named blocker, that blocker landed this morning, and the fix touches the same files. Splitting it out would mean a second branch editing `SideNavItem.tsx` alongside this one, and a re-score that could not be attributed to either PR.

The `useMenuHover` adoption arrived as a second round of commits on this branch rather than as its own PR: it was deferred here as "adopting it changes behaviour", that call has since been made, and the change lands in the same fifty lines of `SideNavItem.tsx` that the first round rewrote. Splitting it out would have meant a second branch racing this one over the same file for a −18-line diff. Its behaviour change gets its own callout at the top and its own changeset.

The brief offered a split along a11y / theming / code-health lines. It would not have helped: nine of the eleven fixes land in `SideNavItem.tsx` and `SideNav.tsx`, so three PRs would be three branches editing the same fifty lines, racing each other and each needing its own capture run. The diff is ~200 lines of source across five files plus tests, and every change is one row in the table above.

## Test plan

Re-run in full after round three, on this branch's head, rebased onto `041a0ef49`.

- `pnpm lint:strict` — **0 errors**, 48 warnings. That count jumped from 11 because the rebase brought in `@astryx/no-nullish-jsx-guard` ([#3936](https://github.com/facebook/astryx/pull/3936)), which now warns across `core`. **Four are in files this PR touches and all four are on lines it does not:** the deliberate RTL-mirror `span` in `SideNavCollapseButton` (justified in a comment) and three `!= null` guards in `TopNavMegaMenu`'s *desktop* panel. Nothing this diff wrote warns.
- `pnpm test` — full suite **9664 passed, 9 failed** (484 files). In isolation the touched suites are clean: `SideNav.test.tsx` **133/133** (125 + 8 new), `TopNav*.test.tsx` **91/91** (87 + 4 new). The nine:
  - two `packages/cli` cases (`component.test.mjs`, `hook-discovery.test.mjs`) assert **case-sensitive** path resolution and cannot pass on a case-insensitive macOS filesystem — they fail identically with this diff reverted;
  - four `TransferList`/`TransferListSelector` cases and two perf budgets (`Table.perf`, `Markdown/parser.perf`) — the load-sensitive timeouts already noted in earlier rounds;
  - one **SideNav** case, `returns focus to the trigger when the collapsed submenu closes`, whose `waitFor` on the layer's asynchronous focus restore timed out at 1392 ms under full-suite load. It passes **3/3 in isolation** and passed in an earlier full-suite run on this same code. It is a timing flake in a test this PR does not touch, not a regression — the ring is a class, not focus behaviour. Say the word and I will give that `waitFor` an explicit timeout in a separate commit; I have not tuned a test's timeout inside a focus-ring PR on my own initiative.
- `pnpm build` — green.
- `pnpm a11y:audit -- --components SideNav` — 14 stories, **0 violations**, `0 new / 0 resolved`, baseline untouched.
- `pnpm check:changesets` — green; three changesets, all `[fix]`, patch. `pnpm check:sync`, `check:use-client`, `check:package-boundaries`, `check:cli-structure`, `check:demo-media`, `check:executable-bits` — green.
- `pnpm -F @astryxdesign/core typecheck` and `typecheck:docs` — green.
- Storybook + Playwright — round one's 32 pairs still stand for the unchanged states; round three adds **32 focus-state before/after pairs** plus **124 measured focus stops** across two viewports and three themes, both built from their own `dist/`. Numbers above.


cixzhang:
**On A15 (the focus ring) — deferred deliberately, blocked on #4654.**

SideNav items currently draw the browser-default focus ring rather than the house 2px accent outline. That is a real finding, but the fix should not be hand-rolled here: #4654 (`fix(theme): make shared focus outlines themeable`) introduces `packages/core/src/utils/focusOutline.stylex.ts` as the single definition of the standard outline, precisely because five components had each written their own copy.

So the resolution for SideNav is to adopt `focusVisible` from that utility once #4654 lands — not to add a sixth independent definition in `navItemStyles` in the meantime. Adding one now would be a new drift site that #4654 then has to clean up.

Recording it so the next audit does not re-litigate it: **A15 stays open against SideNav, blocked on #4654, and closes by adoption rather than by a local fix.** It is advisory against this component regardless, since the focus treatment lives in the shared `NavItem/navItemStyles.stylex.ts` rather than in SideNav's own styles.

cixzhang:
**On the `drawer-content` passthrough — accepted, no change.**

The audit reported this as T2 (a themeable surface dropped in the drawer modes). Re-verification showed that framing was wrong: no theme target is dropped. `MobileNav` paints the drawer and carries its own target, and stacking `side-nav` on the same node is exactly the specificity conflict **T28** says to skip and flag. What is actually dropped is the consumer's `xstyle`/`className`/`style`/`...rest` — a passthrough concern, not a theming one.

**`drawer` mode is fixed** in this PR: those props now forward to `MobileNav`, which merges them properly.

**`drawer-content` is accepted as-is.** Both drawer modes are *secondary render paths* — SideNav hands its items to a host that owns the layout and the styling. The rubric's T14 was scoped for exactly this: the DOM-shape rules judge the **main render path**, because a secondary path composing into a host is a different contract. In `drawer-content` SideNav returns a fragment and paints nothing, so there is no element for consumer style props to land on, and adding a wrapper purely to catch them would invent a surface the mode does not have — while taking the items out of the host's direct layout relationship.

A consumer wanting to style that surface should style `MobileNav`, which owns it.

**Recorded so the next audit does not re-litigate it:** this is *accepted*, not deferred and not unfixed. If it resurfaces as a finding, the answer is this comment plus the T14 scoping note in the rubric. The rules that do still apply on every path — a11y, i18n, tokens, and correctness — are unaffected by this and were graded normally.

kentonquatman (changes requested):
There are a few visual issues with the popover menu in the collapsed SideNav. The menu appears to have a gray border with sharp corners inside the popover container that has rounded corners. The menu appears to have extra spacing on the left side. Also, the focus ring around the expand/collapse button appears much larger than the button itself.

Image

kentonquatman (dismissed):
The flyout menu looks better, but the expand/collapse button still looks off. The icon isn't centered inside the button, and the button is larger than the other footer actions (help and alerts). Can we fix that button before merging?

packages/core/src/SideNav/SideNav.tsx
```diff
@@ -408,6 +434,13 @@ export function SideNav({
     handleRef,
     () => ({
       getCollapseState: () => collapseStateRef.current,
+      subscribe: (listener: () => void) => {
```
cixzhang: This is adding a new public surface as well as a number of indirections for the collapsible state which originally entered this component as a collapsible state. Why does the button need to read from async external store when it could have integrated directly with the collapsible passed into the component?

packages/core/src/SideNav/SideNav.tsx
```diff
@@ -392,11 +397,29 @@ export function SideNav({
...
+    };
+  }
+
+  // A SideNavCollapseButton outside this tree (via `handleRef`) is not
+  // reachable by context, so it subscribes instead.
+  useEffect(() => {
+    for (const listener of collapseListenersRef.current) {
+      listener();
+    }
+  }, [collapsed, isCollapsible, toggle]);
```
cixzhang: seems odd to need this when the collapsible is provided externally. Is the collapse button able to receive the information via context instead? Why do we need a listener here when collapsible is passed in?

cixzhang:
Screenshot 2026-08-14 at 5 27 29 PM

Subnav outline is cutoff

cixzhang:
> The flyout menu looks better, but the expand/collapse button still looks off. The icon isn't centered inside the button, and the button is larger than the other footer actions (help and alerts). Can we fix that button before merging?

Seems resolved now
Screenshot 2026-08-14 at 5 29 14 PM

liyalabubu (approved):

packages/core/src/SideNav/SideNav.tsx
cixzhang: Answered by [`deb4aa7`](https://github.com/facebook/astryx/commit/deb4aa76930f3c5f91c89c2d642c7b62acfd2579) — you were right, and the store is gone. The button now takes the same controlled `collapsible` object handed to `SideNav`, so the state the consumer already owns reaches it as a prop. No `subscribe`, no `useSyncExternalStore`, and no addition to `SideNavImperativeCollapseHandle`. Detail in the newer thread.

packages/core/src/SideNav/SideNav.tsx
cixzhang: Resolved in [`deb4aa7`](https://github.com/facebook/astryx/commit/deb4aa76930f3c5f91c89c2d642c7b62acfd2579) — the listener set, the notify effect and the `subscribe`/`useSyncExternalStore` plumbing are all gone.

Context can't reach it: a button rendered outside the sidenav (a `TopNav` slot, say) is a *sibling* of `SideNav`, not a descendant, so no provider `SideNav` renders can see it. Getting it there via context would mean lifting a new provider above both — more public surface for the same job, which was your first point.

So it takes the state as a prop instead — the same controlled config `SideNav` already gets:

```tsx
const [isCollapsed, setIsCollapsed] = useState(false);
const collapsible = {isCollapsed, onCollapsedChange: setIsCollapsed};

<TopNav endContent={<SideNavCollapseButton collapsible={collapsible} />} />
<SideNav collapsible={{...collapsible, hasButton: false}}>…</SideNav>
```

A button *inside* the sidenav still reads context and is untouched. `handleRef` behaves exactly as it did before this PR, now deprecated in favour of `collapsible`, and `SideNavImperativeCollapseHandle` is back to its original shape — so this PR adds no imperative surface. SideNav + TopNav 247/247, typecheck clean.

PR #4881, opened 2026-08-11
feat(core): one shared dismissal stack for every overlay layer

cixzhang:
## The problem

Every overlay owned its own Escape listener, so **one press could dismiss two layers**: a popover and the Dialog hosting it, or a modal and the modal it was opened from.

We had been fixing this per primitive, and that produced four parallel registries: `useFocusTrap`'s Escape stack (#3357, DOM-depth resolution added in #4375), `lab/Drawer`'s own LIFO registry, `useScrollLock`'s counter (#4788, landed for the same class of bug), and a fifth proposed in #4865. Each one coordinates a different subset of overlays and none of them know about each other.

This is the single registry those collapse into — the "Dismissal / Escape stack in `LayerProvider`" from the program plan in **#3343**.

## The design

**One stack, one listener, one decision point.**

`useLayerDismissal` registers a layer while it is active. `layerStack` owns a single `keydown` listener on `document` and routes each press to the top-most layer. **Layers no longer listen for Escape at all.** That is the core change: "everyone listens and filters themselves out" is what forced the `stopPropagation()` choreography between element-level and document-level handlers, and it broke silently whenever one participant forgot.

**One press dismisses exactly one layer**, with no per-component exceptions. A layer declares what it does with a press that reaches it:

| `escapeBehavior` | Meaning | Used by |
|---|---|---|
| `close` *(default)* | dismiss this layer, consume the press | modals, popovers, menus, tooltips, hover cards |
| `block` | consume **without** dismissing | `Dialog purpose="required"` — must not close, and must not let the press leak to a layer behind it |

I deliberately did **not** add a "dismiss but let the press continue" variant for hover layers. It is tempting (the user never opened the tip, so why should it eat their press?), but the asymmetry decides it: guessing wrong that way is destructive — someone dismissing a stray tooltip over a half-filled form loses the form — while guessing wrong the other way costs one extra keystroke.

**A layer's place in the order is keyed to its identity, not to each registration.** A layer re-registers whenever its depth or behavior changes — a `Dialog` whose `purpose` flips from `undefined` to `required` while it is open is the real case, and StrictMode remounts every effect once more on top of that. Counting registrations would hand it a fresh, higher place each time and promote it above the layers opened over it, so Escape would start hitting the dialog underneath instead of the modal on top. Order is keyed to the layer's identity instead. The `Purpose flips while open` story drives it.

**Ordering comes from the React tree, not the DOM.** `LayerDepthProvider` wraps each layer's own content, so anything opened from inside registers as nested. Context survives portals (DOM containment does not — a nested overlay routinely renders into `document.body`) and depth is fixed during render, so an inner and outer layer that mount in the *same commit* order correctly. Effect order can't do that: React runs child effects before parent effects, so the inner layer registers first and looks older.

There is **nothing to mount at the app root** — the context defaults to `0` and depth composes on its own. DOM containment remains a tie-break for bare `useFocusTrap`, which renders nothing and so cannot provide depth.

**Presence is asked at press time, not inferred from registration.** Hover layers register for their whole lifetime because their open state lags the DOM by a frame; they answer presence from the DOM instead. This is what makes "consume" safe — an idle tooltip sitting in the tree never claims an Escape.

**The stack, not the browser, decides.** It calls `preventDefault()` on presses it handles, which suppresses the browser's own close-watcher — both `<dialog>`'s `cancel` and `popover="auto"` light-dismiss (verified in Chromium; the default action runs after propagation, so a bubble-phase `preventDefault` still beats it). The alternative — leaning on the native top layer — only covers `showModal()` and `popover="auto"`, which would leave non-modal `show()` drawers, `popover="manual"` layers, and older browsers on a second, differently-behaved code path.

A dismissal the browser starts on its own — the Android back gesture, the platform close watcher — still reaches `<dialog>`'s `cancel`, because the stack never saw a press to suppress. Dialog answers those with the same ordering: `preventDefault()` always, so the browser cannot close a controlled `<dialog>` behind React's back, then dismiss only when this dialog is `isTopmost()` and Escape-dismissible.

**A composing Escape is claimed, not ignored.** An IME sends Escape to cancel a half-formed character, ahead of `compositionend`, so no layer may act on it — but standing down is not enough. An unclaimed Escape is exactly what makes the browser raise its own close request, and that request lands on `cancel`, which has no composition state to read and dismisses on the same keypress. In Chromium a guard that only returns measures identically to no guard at all. So the stack `preventDefault()`s the press and dismisses nothing, and it claims only while a layer is actually on screen: with nothing to protect, the press is not ours to take.

Close requests that arrive with no keydown to read — the back gesture, the platform close watcher, or a composing Escape that content stopped before the stack saw it — are answered by `shouldDismissOnCloseRequest()`, which adds "not mid-composition" to the top-most rule. `Dialog`, `Lightbox` and `MobileNav` ask it from their `cancel` handlers, so that policy lives in the stack rather than in three copies of it. Composition is tracked from document-level composition events, because the DOM offers no way to ask; a blur clears it too, so a `compositionend` the stack never sees cannot leave the flag stuck on and swallow every later back gesture. The flag is page-wide, so a composition in a field behind a non-modal layer also suppresses that layer's close requests — one keystroke to recover from, and the alternative cannot work for layers that expose no container.

The listener is on the **bubble** phase, not capture, so content inside a layer can claim the press first. Editors are the motivating case: Monaco and the rich-text editor use Escape for their own find widget, and that must win over dismissing the host Dialog. Capture would take the press away from them. (Radix uses capture here; this is a deliberate divergence.)

## Bugs fixed along the way

- **A Tooltip inside a Dialog closed the Dialog, not the tip.** `useTooltip` listened on `document` and neither prevented nor stopped the event, while Dialog's listener sat on the `<dialog>` element and fired first.
- **A HoverCard trigger swallowed Escape whenever it merely had focus.** Its keydown listener called `stopPropagation()` on *any* Escape, open card or not, so focusing a trigger inside a Dialog silently ate the press.
- **A HoverCard pinned open with `isOpen` made its host Dialog keyboard-unclosable.** Introduced here, not on `main`: a controlled card joined the stack, so it consumed every Escape — and the stack cannot dismiss it, because its visibility is the consumer's state. Three presses, nothing moved. `Tooltip` already opted out for exactly this reason; `HoverCard` now does the same (`isEnabled: isOpen === undefined`). jsdom cannot pin this one — it cannot answer `:popover-open`, so the presence probe reads false and the stack routes to the dialog whether the opt-out is there or not; the test passes with the fix reverted, so there isn't one. The `Pinned hover card in modal` story and the row above are the pin.

- **A HoverCard you had tabbed into took its host Dialog with it.** The card's content wrapper ran its own Escape handler with a bare `stopPropagation()`, which kept the press from the dialog's listener but not from the browser's close watcher — so one press closed the card and the dialog behind it. Only reachable once focus is INSIDE the card, which is why the trigger-focused path always looked right. The card is on the stack; its `onDismiss` now also returns focus to the trigger, the one thing the local handler did that the stack did not.

- **A lab `InfoTip` inside a Dialog took the Dialog with it.** The trigger `stopPropagation()`d Escape, so the stack never saw the press — but the browser's own close watcher still fired `cancel` on the dialog behind it, and one press left nothing on screen. The trigger now handles no keys at all and the `Tooltip` it renders takes the press from the stack. This is the third of the three layers in #5168; Lightbox and MobileNav are the other two.

- **An Escape cancelling an IME composition closed the layer, and everything typed into it.** A CJK user backs out of a half-formed character several times a sentence, and the modal went with it. `main` has the same bug: `Dialog`'s guard returns without claiming the press, so the close watcher closes the dialog anyway. Nobody typing English ever sees it, which is why it has been shipping. `BottomSheet` got the same fix by hand in #5322 — the standalone sheet is not on this stack yet, so its own guard stays until it migrates.

## Test plan

**Unit** — three new files besides the ones below: `layerDismissalInvariants.test.tsx` (15), `layerDismissalFamilies.test.tsx` (8, Lightbox/MobileNav/required Dialog on real mounts) and `useFocusTrapEscapeShim.test.tsx` (11, pinning that the deprecated shim still answers about traps only). Plus 21 in `useLayerDismissal.test.tsx`, each one "one Escape press, which layers reacted?": same-commit nesting, fall-through after the inner closes, unrelated layers, DOM-containment tie-break, `close` vs `block`, presence skipping, both opt-outs, content claiming the press via `preventDefault`, the IME guard, and that the stack leaves the event alone when nothing is open.

Plus 4 in `Dialog.test.tsx` that mount real `Dialog`s rather than synthetic layers: a modal inside a modal takes one Escape on the inner only, the next Escape takes the outer, and a `cancel` dismisses the top-most modal but not one underneath it. The modal-in-modal Escape test fails against the pre-stack `Dialog` handler (the outer closes too) and passes here.

Five of the 21 are the IME case: that the composing press is claimed, that it is left alone when no layer is on screen, and that a close request is declined mid-composition and again once focus leaves the field. Two more in `Dialog.test.tsx` mount a real `Dialog` with a field; six are added there in total. Both halves are pinned separately — reverting the `preventDefault` hoist reds two of them, and dropping the composition check from `shouldDismissOnCloseRequest` reds three. jsdom models neither composition nor the close watcher, so these pin the wiring only; the behaviour is measured below.

**Visual, in real Chromium** (Playwright against local Storybook, 14 stories under `Core/Layer Dismissal`) — one Escape press per row:

| Story | Result |
|---|---|
| Modal in modal (inner in outer's subtree) | inner closes, outer stays → 2nd Escape closes outer |
| Popover in modal | popover closes, modal stays → 2nd closes modal |
| Hover tip in modal, **tip showing** | tip hides, modal stays → 2nd closes modal |
| Hover tip in modal, **no tip showing** | modal closes immediately (idle tip claims nothing) |
| `required` dialog in a host modal | nothing moves at all |

Screenshots captured for each transition; scrim mean-luma steps confirm exactly one layer peeling off per press.

The `cancel` path is checked the same way, in Chromium 149: on the Modal-in-modal story, `requestClose()` — the platform close-watcher path the Android back gesture fires — closes the inner modal and leaves the outer, and does nothing when called on the outer while the inner is on top. Against the bare `preventDefault()` cancel handler, neither call closes anything.

**IME, in real Chromium** — a Hangul composition driven into a real field through CDP `Input.imeSetComposition`, then Escape. Same twelve rows before and after the change, on this branch:

| One Escape, mid-composition | before | after |
|---|---|---|
| `Dialog` with a field | dialog gone, 한 with it | dialog stands, 한 still composing |
| `AlertDialog` | gone | stands |
| Modal in modal, press in the inner | inner gone | inner stands |
| Content `stopPropagation`s the press first | gone (`cancel` fired) | stands (`cancel` fired, declined) |
| `requestClose()` mid-composition | gone | stands |

Plain Escape is unchanged on every regression row: `Dialog` and `AlertDialog` close on one press, modal-in-modal peels one layer then the other, a popover in a modal closes the popover only, a `required` dialog still moves nothing, and `requestClose()` with no composition still closes. Screenshots of each transition captured and read.

**The boundary with layers that are NOT on the stack yet, in real Chromium** — the worry was that a stacked layer inside an unstacked one would be the very double-dismiss this is meant to fix: the stack claims the press, and the unstacked host runs its own listener anyway. Measured on both sides, one Escape per row:

| One Escape | `main` | here |
|---|---|---|
| `ContextMenu` open inside a `Dialog` | **both close** | menu closes, dialog stands |
| Tooltip showing inside a `ContextMenu` | tip only | tip only |
| lab `InfoTip` showing inside a `Dialog` | **both close** | tip only, dialog stands |
| `HoverCard` inside a `Dialog`, focus tabbed INTO the card | **both close** | card only, dialog stands, focus back on the trigger |
| `HoverCard` pinned open with `isOpen` inside a `Dialog` | dialog closes | dialog closes (a controlled card is off the stack) |
| `Popover` open inside a `BottomSheet` | sheet closes, popover does not | same |
| `Dialog` open inside a `BottomSheet` | both go | sheet closes, dialog is left registered as open |

No row got worse in what is on screen, and two got better. The double-dismiss does not happen across the boundary, because every self-handling layer claims the press at element level and the stack stands down on an already-`defaultPrevented` press. `ContextMenu` is the exception that proves it — its listener is on `document` and does not read `defaultPrevented`, so it does fire alongside the stack, but only ever against the layer the stack just dismissed.

What the boundary *does* still cost is the last two rows: a layer opened inside an unstacked host never gets the press, because the host takes it first. That is unchanged from `main` for what the user sees, with one wrinkle worth knowing about — on `main` the inner `Dialog` was dismissed along with the sheet, and here it is not, so its `isOpen` stays `true` and reopening the sheet brings it back. Not a new class of bug, but it is the strongest argument for migrating `BottomSheet` next.

**Repo checks** — `tsc --noEmit` clean, `eslint` clean on all touched paths, `pnpm check:repo` fully green (sync comments, package boundaries, changesets, use-client), pre-commit hook run.

**Full suite** — `packages/core` + `packages/lab`, 7211 passing after merging `main` (286 files, no failures). A handful of load-sensitive tests (TransferList, Table perf, TabList) flake under full parallel load on my laptop; I confirmed the same set flakes on **clean `main`** with the change stashed, and all of them pass in isolation on this branch. DropdownMenu specifically — it rides `useFocusTrap` — passed 3/3 isolated runs (113 tests each).

## Scope / follow-ups

Deliberately limited to the primitive plus the families that exercise both behaviors: `Dialog`, `useFocusTrap` (Popover, DropdownMenu + submenus), `Tooltip`, `HoverCard`. Lightbox, MobileNav and lab `InfoTip` came along because #5168 is one bug across three layers and fixing two of them is not fixing it.

Still self-handling, to migrate next:
- **Comboboxes** (Typeahead, Selector, MultiSelector, PowerSearch) — highest-frequency real bug; APG wants Escape to close the listbox and consume so a host Dialog doesn't close underneath.
- **The standalone `BottomSheet` first** — it is the one measured above where a layer opened inside it never gets the press, and migrating it is also what lets #5322's hand-rolled IME guard go.
- `lab/Drawer`, `CommandPalette`, `ContextMenu`, `DropdownMenuSubMenu`, `PowerSearchEditPopover`. The standalone sheet keeps the hand-rolled IME guard #5322 gave it until then — collapsing it into the stack means migrating the sheet, not deleting a check.

**The hover-layer family is untestable in jsdom, and that is worth fixing separately.** `Tooltip`/`HoverCard` answer presence with `el.matches(':popover-open')` behind a try/catch, on the assumption that an environment without the Popover API throws. jsdom returns `false` instead, so the fallback never runs and every hover layer reads as absent — which is why nothing in the new Layer suite mounts one, and why the controlled-card case above has a story rather than a test. Feature-detecting `showPopover` (what `useLayer.tsx:492` already does) fixes it, but it is not a one-liner: it makes previously-invisible layers visible to existing tests and reds 9 in `DateInputTouch.test.tsx` — as **timeouts**, not failed assertions, so it wants a bug hunt rather than a test update (that file mounts a lifetime `useTooltip` for its disabled message, which the swap makes visible to the stack for the first time). Measured, then reverted — a follow-up, not this PR.

`hasActiveFocusTrapEscape()` is kept and deprecated. It deliberately still counts focus traps ALONE rather than every stacked layer: the stack now carries families that never trapped focus, and counting those would tell its one remaining caller (`BottomSheetSwitcher`) that a trap sits above it when none does.

Outside-press, focus restore, and scroll-lock are the other channels #3343 wants on this stack; #4788 landed a counter for scroll-lock, which is the next one to fold in. The entry shape is ready for them; this PR does not build them.

**Supersedes #4865**, which fixed the modal-in-modal case with a fifth parallel registry. Two things came across from it, with thanks to @freddymeta: gating the `cancel` handler on `isTopmost()` so a browser-initiated dismissal still closes the top-most modal, and the Dialog-level modal-in-modal Escape test — the only real `Dialog` mount guarding nested dismissal in either PR.

### Known edges

A layer keeps its place in the order for as long as its identity lives, so one that closes and reopens does not move to the top of a stack it was already in. Only reachable for a layer whose host stays mounted across the round trip, and only against layers registered after it.

A menu nested inside a dialog that stays open while a *shallower* second dialog opens would out-rank the new dialog. It requires the menu to still be open across the transition, which is unusual — noted rather than solved.


cixzhang:
Thanks for taking this on — one Escape owner for the whole app is the right direction, and the depth-aware stack is a real improvement over every layer racing the same keypress.

The thing I'd want to sort out first: we preventDefault every Escape we handle (`layerStack.ts:224`), but only five families are registered. Lightbox and MobileNav close via the native `cancel` event only, so once a required Dialog is on the stack, Escape stops reaching them. Same shape at `Dialog.tsx:533` — `cancel` also fires for the Android back gesture, and that path no longer closes anything. Could the stack fall through when nothing registered claims the press?

`hasActiveFocusTrapEscape` went public in 0.4.2, and the shim changes what it means from "a trap is above me" to "any layer is present" — `BottomSheetSwitcher.tsx:581` gates dismissal on it, so the sheet stops closing whenever a tooltip is up. Ideally I'd like to keep the shim trap-only and migrate the sheet properly in a follow-up.

Sizing: this is a lot at once. The stack + Dialog + `useFocusTrap` would carry Popover, DropdownMenu, PowerSearch and Typeahead on their own — Tooltip and HoverCard bring the presence-at-press-time machinery and the WCAG argument, which feels like its own PR.

One sequencing note: [#5018](https://github.com/facebook/astryx/pull/5018) is smaller, non-draft, and on a different channel (pointer, not keydown) — no conflict between the two heads. Worth landing that first and rebasing this on top. What do you think?

cixzhang:
Pushed the two fixes you asked for, on top of a fresh merge with `main`. Still a draft, nothing merged.

## 1. The families that were not on the stack are on it now

Rather than a generic fall-through, `Lightbox` and `MobileNav` now register with `useLayerDismissal` like the other five. Falling through would have handed the press back to the browser's close watcher, which only reaches whichever families happen to be a native `<dialog>` or `popover="auto"` — the same press would still be lost for anything else, and `block` could no longer promise that nothing behind it moves. Registering them puts them in the depth order instead, so "an unregistered family above a required Dialog" stops being a case at all.

Both keep their `cancel` handler for the dismissals the browser starts itself — the Android back gesture, the platform close watcher — now gated on `isTopmost()` the way `Dialog`'s is, so a back gesture still closes the top-most layer and only that one. Escape presses the stack owns never reach `cancel`, so nothing dismisses twice.

Also dropped `layerStack`'s `hasOpenLayer()`: with these two registered it had no caller left, and "is anything open" is the question that produced the bug above.

## 2. The shim is trap-only again

`hasActiveFocusTrapEscape()` counts Escape-dismissible focus traps, as it did in 0.4.2 (`escapeStack.length > 0`, and only `useFocusTrap` ever pushed there). The count is driven by the same expression that registers the trap on the stack, so the two cannot drift apart. `BottomSheetSwitcher` is untouched — that migration is still yours.

### Equivalence, measured rather than argued

Same eleven assertions, run against `v0.4.2` and against this branch. Every row matches; the middle column is this branch before the fix.

| Scenario | 0.4.2 | head, before | head, after |
|---|---|---|---|
| nothing mounted | false | false | false |
| active trap with `onEscape` | true | true | true |
| trap with no `onEscape` | false | false | false |
| inactive trap | false | false | false |
| nested traps | true | true | true |
| open Popover | true | true | true |
| open Dialog | false | **true** | false |
| open Lightbox | false | **true** | false |
| open MobileNav | false | **true** | false |
| showing Tooltip | false | false¹ | false |
| Popover in Dialog, then Popover closes | true → false | true → **true** | true → false |

¹ jsdom answers `:popover-open` as false, so the tooltip never looks present there — that row only misbehaves in a real browser, which is what the sheet story below covers.

One row from your list has no 0.4.2 answer: `BottomSheet` postdates the tag, so "a tooltip above a bottom sheet" cannot be run against it. The rule it would follow is unambiguous (a non-modal sheet traps nothing, a tooltip is not a trap → false), and the behavior is verified in Chromium instead.

## Chromium, before and after

Playwright against local Storybook, Chromium 149, one Escape per row. Recorded which `<dialog>`s were actually open at each step, plus a frame.

| Story | Before | After |
|---|---|---|
| required Dialog, Lightbox above it | Lightbox stays | **Lightbox closes, Dialog stays** |
| required Dialog, MobileNav above it | drawer stays | **drawer closes, Dialog stays** |
| required Dialog alone | swallowed | swallowed |
| required Dialog inside a host modal | nothing moves | nothing moves |
| Lightbox alone | closes | closes |
| MobileNav alone | closes | closes |
| Lightbox / MobileNav, `requestClose()` (the back-gesture path) | closes | closes |
| non-modal sheet, hover tip showing | sheet stays | **sheet closes** |

Every frame that should not have changed is byte-identical between the two runs; only the three regression rows differ.

**Lightbox over a required dialog** — open, then one Escape:

| before | after |
|---|---|
|  |  |

**MobileNav over a required dialog:**

| before | after |
|---|---|
|  |  |

**Non-modal bottom sheet with a hover tip showing** — the shim regression:

| tip up | before | after |
|---|---|---|
|  |  |  |

## Tests

- `Layer/layerDismissalFamilies.test.tsx` — 8 cases. Reverting the two component changes turns 4 of them red; the other 4 are the guards (`block` still swallows, a top-most `cancel` still closes) and are meant to stay green.
- `hooks/useFocusTrapEscapeShim.test.tsx` — the 11 rows above. Reverting the shim turns 4 red.
- 5 new stories under `Core/Layer Dismissal` for the Chromium runs.
- `packages/core` Dialog, Lightbox, MobileNav, Layer, Popover, Tooltip, HoverCard, BottomSheet: 567 passing. `tsc --noEmit` and `eslint` clean on the touched paths; `check:repo` green via the pre-commit hook.

## Still open

- [#5018](https://github.com/facebook/astryx/pull/5018) has not landed yet, so this is merged with `main` as of today rather than rebased on it. The shared file (`Popover/usePopover.tsx`) is untouched by these two commits, so the auto-merge you verified still holds.
- `InfoTip` is the one family left off the stack; `useLayerDismissal`'s doc now says so.
- The sizing question is left exactly as you had it — Tooltip and HoverCard are still in, nothing was split.

PR #4885, opened 2026-08-11
fix(Button): use aria-disabled instead of native disabled while busy-only

### LINKED ISSUE #4871: Button uses native disabled for busy state, which drops focus
## What happens

While a `clickAction` is pending, `Button` sets the native `disabled` attribute on the `<button>`. A natively disabled element cannot hold focus, so the browser moves focus to `<body>` the moment the action starts, and it is not restored when the action settles.

In `packages/core/src/Button/Button.tsx`:

```ts
const isLoadingState = isLoading || isPending;
const buttonDisabled =
  isDisabled || groupDisabled || (isLoadingState && !isInterruptible);
...
disabled={useAriaDisabled ? undefined : buttonDisabled}
```

`useAriaDisabled` is only true when a `tooltip` is set, so in the common case the busy button is natively disabled.

## Why it's a problem

A keyboard user who activates "Save" with Enter loses their place: focus falls to `<body>`, so the next Tab restarts from the top of the document, and screen readers announce nothing when the action completes. This is the classic reason busy state should never be expressed as `disabled`.

It also contradicts the documented rule in the API Conventions wiki page ("Disabled vs Busy"), which says busy is visual-only — `aria-busy` plus a spinner — and that only `isDisabled` uses the native attribute. Input components already follow that rule: they render `aria-busy` and keep the control focusable, guarding re-entry in the change handler.

## Why the native disable looks redundant

`Button` already guards re-entry in the click handler with a ref:

```ts
const actionInFlightRef = useRef(false);
if (buttonDisabled || (actionInFlightRef.current && !isInterruptible)) {
  e.preventDefault();
  return;
}
```

That guard is what actually makes a fire-once action fire once — it dedupes even a same-tick double click, which `disabled` does not reliably do. The native attribute adds the focus loss without adding the safety.

## Suggested fix

While busy (and not otherwise disabled):

- keep the button focusable — no native `disabled`
- keep `aria-busy="true"` and the spinner
- set `aria-disabled="true"` so AT announces the state
- suppress Enter/Space activation via the existing handler path, the way the `aria-disabled` + tooltip branch already does
- leave `isDisabled` behavior unchanged: that stays native `disabled`

`isInterruptible` can stay as-is; it would simply become the variant that also lets activation through.

## Notes

Changing this affects every consumer that relies on a busy button being unclickable, so the re-entry guard needs to cover the keyboard path too before the attribute comes off. Worth a changeset and a note in the release notes.

HelloOjasMutreja:
## Summary

Closes #4871.

`Button` sets the native `disabled` attribute while a `clickAction` is pending (or `isLoading` is set), even when `isDisabled` isn't set. A natively disabled element can't hold focus, so the browser moves focus to `<body>` the instant the action starts and doesn't restore it — a keyboard user who activates "Save" with Enter loses their place, and screen readers announce nothing when the action completes.

This also contradicts the documented "Disabled vs Busy" rule in the API Conventions wiki page (busy is visual-only — `aria-busy` + spinner — only `isDisabled` uses the native attribute), which input components already follow.

## Why the native disable is redundant here

`Button` already guards re-entry in the click handler with `actionInFlightRef` — that's what actually makes a fire-once action fire once, and it dedupes even a same-tick double click, which `disabled` doesn't reliably do. The native attribute adds the focus loss without adding any safety the guard didn't already provide.

## Fix

Split `buttonDisabled` into `isTrueDisabled` (`isDisabled`/group-disabled) and `isBusyOnlyDisabled` (blocked purely by an in-flight, non-interruptible action, with no other disable reason). `useAriaDisabled` now covers both the existing tooltip case and the new busy-only case. A genuinely `isDisabled` button — with or without a simultaneous busy state — keeps native `disabled`, per the issue's own "leave `isDisabled` behavior unchanged" note. `isInterruptible` is unaffected.

The existing `aria-disabled` keydown-suppression path (Enter/Space blocked, other keys pass through) already covers the busy-only case with no new code, since it was keyed off `useAriaDisabled`, which now includes this case.

## Test notes

Updated three pre-existing `Button` tests and one `IconButton` test that asserted the old native-disabled-while-loading behavior (now assert `aria-disabled` + focusable instead), and added five new tests for the busy-only case (aria-disabled + focusable + doesn't fire onClick + suppresses Enter/Space + `isDisabled` alone still keeps native disabled, including when combined with `isLoading`). Verified the core regression test fails against the pre-fix code and passes with the fix.

While running the full suite I also hit a genuine flake in the pre-existing `isInterruptible` test — it assumed `useTransition`'s `isPending` settles back to `false` within exactly one microtask tick after resolving the action, which was timing-sensitive and only surfaced under full-file-run timing (not this fix's logic — verified by reverting just the source change and confirming the test still passed in isolation, then reproducing the flake against the full file with the fix applied). Replaced the single `await Promise.resolve()` with `waitFor` polling.

Swept for other consumers with `toBeDisabled()` + `isLoading` assertions across the package (`CheckboxInput`, `CheckboxList`, `DateInput`, `DateTimeInput`, `FileInput`, `MultiSelector`, `Switch`, `TextArea`) — all pass unmodified, since they have their own independent `isLoading`/`isDisabled` semantics unrelated to `Button`'s `clickAction` busy state.

## Test plan

- [x] `vitest run packages/core/src/Button` — 44/44 passing
- [x] `vitest run packages/core/src/IconButton` — 10/10 passing
- [x] `vitest run packages/core/src/ButtonGroup` — passing (consumer)
- [x] Broader sweep: `CheckboxInput`, `CheckboxList`, `DateInput`, `DateTimeInput`, `FileInput`, `MultiSelector`, `Switch`, `TextArea` — 526/526 passing, unmodified
- [x] Verified the core new test fails against the pre-fix code and passes against the fix
- [x] Ran the full `Button.test.tsx` suite multiple times to confirm the `isInterruptible` flake fix holds
- [x] `tsc --noEmit` clean for touched files
- [x] `eslint` clean

### DIFF
--- a/.changeset/button-busy-aria-disabled.md
+++ b/.changeset/button-busy-aria-disabled.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] `Button` (and `IconButton`) no longer use the native `disabled` attribute while a busy state (`isLoading`, or a non-interruptible `clickAction` in flight) is the only reason the button can't be clicked. A natively disabled element can't hold focus, so the browser moved focus to `<body>` the instant an action started and never restored it — the classic reason busy state should be visual-only. Busy now uses `aria-disabled` (button stays focusable) with `aria-busy` and the spinner, matching the documented "Disabled vs Busy" API convention and the behavior input components already follow. `isDisabled` (and `isDisabled` together with a busy state) is unchanged — that still uses native `disabled`. `isInterruptible` is also unchanged.
+
+@HelloOjasMutreja
--- a/packages/core/src/Button/Button.doc.mjs
+++ b/packages/core/src/Button/Button.doc.mjs
@@ -138,7 +138,7 @@ export const docs = {
     {
       name: 'isLoading',
       type: 'boolean',
-      description: 'Shows a loading spinner and disables interaction. Announces "Loading" via a live region.',
+      description: 'Shows a loading spinner and blocks re-activation while staying focusable (aria-busy + aria-disabled, not native disabled, so keyboard focus is never dropped mid-action). Announces "Loading" via a live region.',
       default: 'false',
     },
     {
@@ -360,7 +360,7 @@ export const docsDense = {
     displayName: 'HTML name for form submission',
     value: 'HTML value for form submission',
     form: 'associates button with form element by ID',
-    isLoading: 'shows spinner+disables interaction; announces via live region',
+    isLoading: 'shows spinner, blocks re-activation but stays focusable (aria-disabled, not native disabled); announces via live region',
     icon: 'icon element rendered before label text',
     isIconOnly: 'when true, renders square icon-only button; label becomes aria-label',
     width: "Width of button. Numbers=pixels, strings=as-is (e.g. '100%' for full-width).",
--- a/packages/core/src/Button/Button.test.tsx
+++ b/packages/core/src/Button/Button.test.tsx
@@ -10,7 +10,14 @@
  */
 
 import {describe, it, expect, vi} from 'vitest';
-import {render, screen, fireEvent, act, waitFor} from '@testing-library/react';
+import {
+  render,
+  screen,
+  fireEvent,
+  createEvent,
+  act,
+  waitFor,
+} from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import {Button} from './Button';
 import {Badge} from '../Badge/Badge';
@@ -117,8 +124,11 @@ describe('Button', () => {
       resolveAction?.();
     });
     // useTransition's isPending can take more than one microtask tick to
-    // settle back to false — poll instead of assuming a single await flushes
-    // it, which was flaky under full-suite timing.
+    // settle back to false — poll instead of assuming a single act() flushes
+    // it, which fails deterministically under this React/testing-library
+    // combination (unrelated to the busy/aria-disabled change above; kept
+    // here only because reverting it breaks this test outright — see #4885's
+    // review for the standalone fix this carries over from).
     await waitFor(() => {
       expect(button).not.toHaveAttribute('aria-busy', 'true');
     });
@@ -388,9 +398,8 @@ describe('Button', () => {
     act(() => {
       resolveAction?.();
     });
-    // useTransition's isPending can take more than one microtask tick to
-    // settle back to false — poll instead of assuming a single await flushes
-    // it, which was flaky under full-suite timing.
+    // See the comment on the equivalent wait above — same
+    // React/testing-library timing gap, unrelated to isInterruptible itself.
     await waitFor(() => {
       expect(button).not.toHaveAttribute('aria-busy', 'true');
     });
@@ -601,8 +610,6 @@ describe('Button', () => {
   });
 
   it('exposes aria-busy on the link-rendered button while loading', () => {
-    // Non-interruptible loading disables the button, which falls back to
-    // <button> rendering — so an anchor only shows loading when interruptible.
     render(
       <Button
         label="Docs"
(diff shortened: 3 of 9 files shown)


HelloOjasMutreja:
Rebased onto main to pick up #5627, which landed a separate fix distinguishing the loading spinner's full-contrast treatment from the dimmed isDisabled treatment (isTrueDisabled/visuallyDisabled split). Reconciled that with this PR's aria-disabled-while-busy fix: kept #5627's dimming distinction, layered isBusyOnlyDisabled/useAriaDisabled on top of it. The underlying accessibility bug (a busy-only button still loses focus to <body> via the native disabled attribute) is unaffected by #5627 and still needs this fix — verified against current main.

cixzhang (changes requested):
The busy-state fix is the right canonical direction, but this head is not ready: keyboard focus is preserved for Button and IconButton, yet one link path changes semantics and current-main integration fails.

- A busy `href` Button still sets `renderAsLink` false, replacing the link with a button. Keep the link element and role stable while applying `aria-busy`, `aria-disabled`, and handler guards, and add an exact regression test. Please carry over and credit @AKnassa's anchor-preservation work from #4879.
- Rebase onto current main and remove the four #4871 known-failure rows plus the two matching `declaredNotDelivered` entries when the Chromium contract passes. The current synthetic run passed 17 tests and failed 3 because this fix correctly turns those stale records into `unexpected-pass`.
- Update Button's consumer docs and the ListInput loading expectation for focusable-but-inert buttons; the focused current-main run passes 106 tests and fails `ListInput.test.tsx:1164`. Please move the unrelated `isInterruptible` timing-flake rewrite to a separate PR.

[Reviewed by Robohands]

HelloOjasMutreja:
Fixed in f8d992b5b and 322b08d07 (rebased onto current main).

**Busy href Button stays an anchor:** `renderAsLink = href != null && !buttonDisabled`, and `buttonDisabled` includes the busy-only case, so a busy href Button was still swapping to a disabled `<button>` mid-action, exactly the drop-focus bug the underlying fix set out to remove. `renderAsLink` now checks `isTrueDisabled` only; a genuinely disabled href Button still falls back (disabled links are an anti-pattern), but busy-only keeps the anchor, with `aria-busy`/`aria-disabled` applied directly to it and activation guarded by the same `handleClick` ref check already shared with the button branch. Carried over from and credited to @AKnassa's approach in #4879 (`renderAsLink` keyed on hard-disabled only, `aria-disabled` added to the link branch). Added a dedicated regression test asserting the element stays an `<a>` with its `href` intact, is focusable, and its click is guarded.

**Rebased and reconciled the a11y debt:** verified against a real Chromium run that `button.focus.reachable-and-escapable` and `button.unavailable.inert` now report `unexpected-pass` for both Button and IconButton's loading state (same 3-failure shape the review described), then removed the four stale known-failure records and the two `declaredNotDelivered` entries they were paired with. Re-ran the contract after removing them: 20/20 passing clean.

**Docs and ListInput:** updated Button's and IconButton's own `isLoading` prop descriptions (no longer "disables interaction"; now describes staying focusable via aria-disabled). Fixed `ListInput.test.tsx`'s loading-state test: its add-item button receives `isLoading` directly and is now aria-disabled rather than natively disabled while busy; its remove/reorder controls map ListInput's own loading state onto `isDisabled` instead and were already correct.

**isInterruptible timing "flake":** traced this down rather than just moving it. It isn't a pre-existing flake; running the two isInterruptible tests unmodified against current main confirmed the original `await act(async () => { resolveAction?.(); await Promise.resolve(); })` form already settles `useTransition`'s `isPending` correctly on its own. My own earlier commit had simplified that to a synchronous `act(() => {...})` and then added a `waitFor` poll to paper over the resulting timing gap, describing it as a pre-existing flake fix when it was actually working around a regression the same commit introduced. Restored the original `act()` form for both tests instead of carrying the `waitFor` workaround; confirmed passing reliably across repeated runs. No separate PR needed since there was no independent fix to split out.

Verified: Button/IconButton/ListInput suites (157/157), typecheck and lint clean, real-Chromium a11y contract 20/20.

HelloOjasMutreja:
Rebased onto current main. One real conflict in Button.a11y.known-failures.ts: main independently added Button/IconButton loading known-failures (documenting exactly this PR's bug, from unrelated test-infra commits), while this branch's own prior commits still carried its ClickableCard known-failures from before #6150 removed them upstream.

Resolved by:
1. Keeping both blocks through the rebase itself (non-overlapping array entries).
2. Removing the Button/IconButton loading entries, since this PR's aria-disabled fix resolves exactly what they describe.
3. Removing the stale ClickableCard entries too, once the real Chromium contract flagged them as unexpected-pass — they document a bug #6150 already fixed and already removed from main's own copy of this file; this branch's older commits just hadn't caught up.

Verified against the real Chromium a11y contract this time, not just jsdom: all 20 Button-family states pass, including the four this PR's fix and the rebase cleanup resolve. Unit suite (157 tests across Button/IconButton/ListInput) and typecheck both clean.

PR #4886, opened 2026-08-11
fix(cli): filter extractComponents to only imports from @astryxdesign…

### LINKED ISSUE #4677: detail-page skeleton references Timeline but component registry has no Timeline

abu-abdullah22:
Fixes #4677

## Summary
`extractComponents()` (used by `astryx template --skeleton`) matched any 
capitalized JSX tag in a template's source via regex, with no check for 
whether it was actually imported from @astryxdesign/core. This let 
locally-defined helper functions inside template files (e.g. a local 
`function TimelineSection() {}` in 
packages/cli/assets/templates/pages/detail-page/page.tsx, used only within 
that file) get falsely reported as real components — after the existing 
suffix-stripping logic stripped "Section" from "TimelineSection", it produced 
a phantom "Timeline" component that has never existed in packages/core.

## Changes
- packages/cli/foundation/discovery/template-adapter.mjs: extractComponents() 
  now cross-references matched tag names against the file's actual 
  `import {...} from '@astryxdesign/core...'` statements, keeping only names 
  that are genuinely imported. Existing filters (UBIQUITOUS, Theme exclusion, 
  suffix-stripping) are unchanged — this is an additional final filter.
- packages/cli/api/template/template.test.mjs: added a test confirming a 
  locally-defined helper (TimelineSection) is excluded while a real imported 
  component (Card) is kept.
- Added a changeset (patch, @astryxdesign/cli).

## Note for reviewers
3 pre-existing test failures in template-suffix.test.mjs and one in 
discover-integration-templates.test.mjs are unrelated to this change — the 
former fail on Windows due to fs.symlinkSync requiring elevated permissions 
(EPERM), a pre-existing environment limitation. Confirmed these also fail on 
a clean checkout before my change.

### DIFF
--- a/.changeset/cli-fix-timeline-phantom-component.md
+++ b/.changeset/cli-fix-timeline-phantom-component.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/cli': patch
+---
+
+[fix] `extractComponents` reported phantom components from local helper functions in template files — e.g. "Timeline", derived from a locally-defined `TimelineSection` helper in the detail-page template, which isn't a real component and was never imported from @astryxdesign/core. It now only reports names actually imported from @astryxdesign/core.
+
+@abu-abdullah22
--- a/packages/cli/api/template/template.test.mjs
+++ b/packages/cli/api/template/template.test.mjs
@@ -1,7 +1,10 @@
 // Copyright (c) Meta Platforms, Inc. and affiliates.
 
+import * as fs from 'node:fs';
+import * as os from 'node:os';
+import * as path from 'node:path';
 import {describe, it, expect} from 'vitest';
-import {stripTemplateAssetRefs, template} from './template.mjs';
+import {stripTemplateAssetRefs, template, extractComponents} from './template.mjs';
 
 describe('stripTemplateAssetRefs', () => {
   it('replaces a /template-assets image path with an inline data URI', () => {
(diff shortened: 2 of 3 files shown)


cixzhang (changes requested):
Builders following a template can lose real components from the CLI’s `components` list. At this head, the Core `TopNavMegaMenuItem` block reports only `Grid`, and the current-main RichText integration reports none; the package-prefix regex also admits `@astryxdesign/corex` (`packages/cli/foundation/discovery/template-adapter.mjs:683`).

Could this use the CLI’s accepted component-resolution boundary instead, keeping root/subpath imports and aliases while retaining supported package and integration components? The synthetic merge currently fails all six focused `extractComponents` tests.

[Reviewed by Robohands]

abu-abdullah22:
Thanks @cixzhang for the detailed feedback. you were right, the import-based approach 
was too narrow and broke real components that aren't locally imported (or 
come from integration packages). Reworked it: instead of checking imports, 
extractComponents now excludes names that match a local function/const/let/
class declaration in the same file (e.g. `function TimelineSection() {...}`), 
applied before suffix-stripping. This fixes the original phantom-Timeline 
bug without touching how real imported/aliased/integration components are 
recognized.

Verified against packages/cli/foundation/discovery/template-adapter.test.mjs 
(all 10 tests, including the 6 you flagged as failing against the synthetic 
merge) and my own existing test, both pass cleanly now.

PR #4890, opened 2026-08-11
fix(indicator): close the six audit blocks on the indicator layer

cixzhang:
## What

The first rubric audit of the indicator layer (**68.0 / D**, mode O, at `505ccaa327`) found six blocking defects. This closes all six, plus the raw radius the same audit recorded as a FIX.

The system-level design came through the audit intact — replacement by name works, families type-check, hover reaches the visual with no props, and the owner-drawn focus ring survives a replacement that forwards nothing. Every defect here is at an edge: one indicator honoured the props contract on only one of its two paths, the documented extension path returned `undefined`, the docs prescribed an API that was never built, and a public theme target was deleted as a patch.

| Rubric ID | Section | Defect | Fix |
| --- | --- | --- | --- |
| **P2** | §3 API contract | `CheckIndicator`'s children branch rendered a bare span that dropped `...rest` and `xstyle` — a caller's `data-testid`, `id`, handlers and styling vanished whenever children were passed | The branch now spreads the same props as the primary path |
| **B15** | §4 Behavior | `if (!isChecked) return null` ran *before* the children branch, so a host passing a busy `Spinner` through an **unchecked** row rendered nothing — the state a listbox row is usually in | Children are checked first; `state` only decides what the *mark* is |
| **B12** | §4 Behavior | `defaultIndicators` was typed as a **total** map over the **open** `IndicatorName` union, so the documented augmentation path compiled while `getIndicator` handed back `undefined` and rendering it threw | Core's names are a closed `CoreIndicatorName`; `getIndicator`/`useIndicator` are overloaded — a core name always resolves, an augmented one is `\| undefined` |
| **T33** | §2 Theming | Public target `astryx-dropdown-menu-radio-dot` was removed with a `patch` changeset and no codemod; runtime themes are not validated, so a theme keyed on it silently stops matching | `[breaking]` changeset + `rename-dropdown-menu-radio-dot-target`, staged in `transforms/next` |
| **V3** | §6 Testing | No test covered `CheckIndicator` at all — the default mark, and the component carrying three of the audit's defects | Three `CheckIndicator` cases + a new `indicatorRegistry.test.tsx` |
| **X5a** | §8 Docs | The docs told theme authors to spread `indicatorFocusRingProps()`, which exists nowhere in the repo, and described a "tells the owning control to stand down" marker that was never built | Replaced with what shipped: the owner paints the ring on the indicator's own element (`useIndicatorFocusRing`), and a replacement only needs a root element with the radius it wants the outline to follow |

Two other doc clauses contradicted the design we merged and now match T34: `children` is "rendered INSTEAD of the mark" **in every state**, and the DON'T against "render only in the selected state" became a DON'T against *assuming you are only mounted when selected* — the host renders unconditionally and passes `state`; drawing nothing in a state is the indicator's call.

## ⚠️ One thing to rule on: `--radius-full` squares the radio under a sharp theme

Per your note, the radio circle and dot now use `radiusVars['--radius-full']` instead of a raw `50%`. On the **neutral** theme this is pixel-identical, as expected — the boxes are square (24/20 for the circle, 10/8 for the dot) and `--radius-full` is `9999px`:

**0 of 5,670,000 pixels differ** across RadioList, the menu radio, and the Selector radio replacement, in light and dark.

neutral before/after

But **y2k** sets `--radius-full: 0px` (it zeroes every radius — `radius.multiplier: 0`, brutalist by design). So the radio becomes a **square** there. Measured, not inferred: the corner pixel of the 24×24 box goes from `rgb(255,255,255)` (outside the circle) to `rgb(47,41,46)` (border), and the painted fraction of the box goes **16.8% → 84.0%**.

y2k before/after

selector radio replacement, y2k

The consequence worth your call: under y2k the radio now has the **same silhouette as the checkbox**, so the mark inside (dot vs check) is the only thing telling single-choice from multi-choice apart.

silhouette collision

Arguments both ways, briefly:

- **Keep the token** (what this PR does). It is the house convention — `Avatar` already uses `--radius-full` and is already square under y2k — and it is the whole point of this workstream: the radio's shape becomes themeable instead of hard-coded. A theme that asked for zero radius everywhere gets it.
- **Go back to `50%`.** Shape is the affordance for single-vs-multiple selection, and a theme squaring an avatar is not the same decision as a theme squaring a radio. Reverting is two lines in `RadioIndicator.tsx`.

I implemented the first because you asked for the token; say the word and I'll flip it.

Note also: 7 other core files still use `borderRadius: '50%'` (Pagination, Calendar, AspectRatio, Table row status, ListItem, NavIcon, StatusDot). I did **not** sweep them — `50%` and `--radius-full` are interchangeable only while the box is square (measured: square 24×24 → 0/7056 px differ; wide 60×24 → 23.5% differ, ellipse vs stadium), and several of those are not guaranteed square.

## Consumer-facing API changes

**Resolving an indicator core doesn't ship.** Augmenting `IndicatorMap` no longer lies about resolution:

```ts
declare module '@astryxdesign/core/Indicator' {
  interface IndicatorMap {
    'brand-star': 'singleSelection';
  }
}

const Check = useIndicator('check');            // IndicatorComponent<'singleSelection'>
const Star  = useIndicator('brand-star');       // ... | undefined  ← you own the default
const Safe  = useIndicator('brand-star') ?? BrandStar;
```

No core call site changes: every one of them passes a literal core name, which hits the non-optional overload.

**Removed theme target.** `dropdown-menu-radio-dot` is gone; menu radios draw the shared radio indicator, so the dot carries `radio-indicator-dot` (and the legacy `radio-dot`). The row's circle keeps `dropdown-menu-radio`. `astryx upgrade` rewrites the key and the `astryx-dropdown-menu-radio-dot` class.

The rename is **not scope-preserving** and cannot be — there is no menu-only dot element left to address, so `radio-indicator-dot` reaches every radio dot in the app. The codemod therefore leaves a TODO at each site it rewrites rather than pretending the swap is equivalent.

**Judgment call:** I marked this `[breaking]` / `minor`, per T33 and the changeset gate (`[breaking]` must be `minor` while 0.x). You said patch was fine for the *radio dot* rename — and it was, because that one kept both names via `legacyNames`. This target is genuinely deleted with no alias, and a theme gets no error. If you'd rather hold the release train, deleting the second changeset keeps it a patch and the codemod still ships.

## Test plan

Everything below was run on this branch, not assumed.

**Positive controls first — each new test was run against the pre-fix code to confirm it fails.**

- `CheckIndicator` children/props tests, run against `505ccaa327:CheckIndicator.tsx`: **2 failed** (`renders children INSTEAD of the mark, in both states`; `forwards caller props and styling on both render paths`), 17 passed.
- Registry type test, `tsc` against `505ccaa327:indicatorRegistry.ts`: fails with `TS2741: Property '"brand-star"' is missing in type ... but required` — the unsoundness made visible — plus `TS2578: Unused '@ts-expect-error' directive` where the old signature wrongly promised a component.
- Visual harness control: light-vs-dark diff of the same story reports **5,670,000** differing pixels, so a 0 is a real 0.
- Codemod: no-op on files that never mention the target; `dropdown-menu-radio` (the surviving circle target) left untouched; idempotent.

**Suites**

| Check | Result |
| --- | --- |
| `vitest` — Indicator, CheckboxInput, RadioList, Selector, DropdownMenu, theme, cli codemods | **1663 passed** (87 files) |
| `tsc` core | clean |
| `tsc` docs (`typecheck:docs`) | clean |
| `CI=true eslint --no-cache` on both touched trees | 0 errors, 0 warnings |
| `check:sync` / `check:changesets` / `check:use-client` / `check:package-boundaries` / `check:cli-structure` | all green |
| `sync:exports:check` | up to date |

**Visual**

Real Chromium via Playwright against two Storybook instances — `505ccaa327` (before) and this branch (after) — 51 image pairs across 7 stories × {neutral light, neutral dark, y2k light}, at `deviceScaleFactor: 3`, with full-page shots and element crops of the circle, dot and checkbox box. Each capture asserts its subject is in the DOM before shooting, so a missing story or an unopened listbox fails loudly instead of diffing two error pages.

- **37 of 51 pairs are byte-identical.** Every differing pair is y2k, and every one of them is the radius change above.
- Negative control: `CheckboxInput / All Variations` — untouched by this PR — is 0-pixel identical in all three theme/mode combinations.
- Computed `border-radius` recorded for every crop: `50% → 9999px` under neutral, `50% → 0px` under y2k.

## Follow-up

Re-audit and the ledger row update (the rubric requires the score to move and the change to be explained) once this lands.

Related: #4712 (the indicator layer this audits), rubric T34 corrected on the wiki in `3812e91`.

### DIFF
--- a/.changeset/remove-dropdown-menu-radio-dot-target.md
+++ b/.changeset/remove-dropdown-menu-radio-dot-target.md
@@ -1,0 +1,9 @@
+---
+'@astryxdesign/core': minor
+---
+
+[breaking] Remove the `dropdown-menu-radio-dot` theme target. Menu radio rows draw the shared radio indicator now, so the dot is the indicator's dot: target `radio-indicator-dot` (the legacy `radio-dot` name still matches it too). The row's circle keeps its `dropdown-menu-radio` target, so only the dot moved.
+
+Runtime themes are not validated — a theme keyed on the removed target keeps compiling and silently stops matching — so `astryx upgrade` now carries `rename-dropdown-menu-radio-dot-target`, which rewrites the key and the `astryx-dropdown-menu-radio-dot` class. The new target is app-wide rather than menu-only (there is no menu-only dot element left to address), so the codemod leaves a TODO at each site it rewrites.
+
+@cixzhang
--- a/.changeset/themeable-indicators.md
+++ b/.changeset/themeable-indicators.md
@@ -7,6 +7,6 @@
 
 Theme targets now follow the component-name convention: `checkbox-indicator`, `radio-indicator`, `radio-indicator-dot`. The old names (`checkbox`, `radio`, `radio-dot`) are still emitted on the same element, so existing themes keep working — migrate at your convenience; they go away in the next major.
 
-Migration: menu radios use those shared targets now. `dropdown-menu-radio-dot` is removed — target `radio-indicator-dot`.
+Migration: menu radios use those shared targets now. `dropdown-menu-radio-dot` is removed — target `radio-indicator-dot`; `astryx upgrade` rewrites it for you.
 
 @cixzhang
--- a/packages/cli/assets/codemods/transforms/next/__tests__/next-codemods.test.mjs
+++ b/packages/cli/assets/codemods/transforms/next/__tests__/next-codemods.test.mjs
@@ -1,0 +1,92 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+/**
+ * @file Unit tests for the staged (next-release) codemods.
+ *
+ * Mirrors v0.3.0/__tests__/next-codemods.test.mjs, which covers the codemods
+ * after promotion. Keeping a copy here means a staged transform is tested from
+ * the day it is written rather than the day it is released.
+ */
+
+import {describe, expect, it} from 'vitest';
+import jscodeshift from 'jscodeshift';
+
+const j = jscodeshift.withParser('tsx');
+const api = {jscodeshift: j, stats: () => {}, report: () => {}};
+
+async function apply(name, source) {
+  const {default: transform} = await import(`../${name}.mjs`);
+  return transform({source, path: 'test.tsx'}, api) ?? source;
+}
+
+const TRANSFORM = 'rename-dropdown-menu-radio-dot-target';
+
+describe('rename-dropdown-menu-radio-dot-target', () => {
+  it('renames the theme target key in a defineTheme components map', async () => {
+    const input = `import {defineTheme} from '@astryxdesign/core/theme';
+export const theme = defineTheme({
+  name: 'brand',
+  components: {
+    'dropdown-menu-radio-dot': {base: {backgroundColor: 'var(--color-accent)'}},
+  },
+});`;
+    const output = await apply(TRANSFORM, input);
+    expect(output).toContain("'radio-indicator-dot':");
+    // The old name survives only inside the TODO comment the rename attaches.
+    expect(output).not.toContain("'dropdown-menu-radio-dot':");
+  });
+
+  it('warns that the new target is app-wide, not menu-only', async () => {
+    const input = `const components = {
+  'dropdown-menu-radio-dot': {base: {width: '10px'}},
+};`;
+    const output = await apply(TRANSFORM, input);
+    // The rename cannot preserve scope — there is no menu-only dot element
+    // left — so the author has to decide, and must be told.
+    expect(output).toContain('TODO(astryx upgrade)');
+    expect(output).toContain('EVERY radio dot');
+  });
+
+  it('renames the rendered class inside a selector string', async () => {
+    const input = `const sel = '.astryx-dropdown-menu-radio-dot';
+const nested = '.astryx-dropdown-menu-radio .astryx-dropdown-menu-radio-dot';`;
+    const output = await apply(TRANSFORM, input);
+    expect(output).toContain("'.astryx-radio-indicator-dot'");
+    expect(output).toContain(
+      '.astryx-dropdown-menu-radio .astryx-radio-indicator-dot',
+    );
+  });
+
+  it('renames the class inside a template literal', async () => {
+    const input =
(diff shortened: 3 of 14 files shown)


cixzhang:
## Self review

Things I'd want a reviewer to push back on, in the order I'd push back:

1. **`--radius-full` squaring the radio under y2k** is the only behaviour change here that a user could see, and it's a design decision, not a bug fix. Flagged up top with the pixel evidence. Everything else in this PR is a defect closure.

2. **`[breaking]` / `minor` for the deleted target.** This is the second judgment call. The rule (T33, and the changeset gate) is unambiguous, but the release-train cost is yours to weigh, and the codemod carries most of the safety on its own.

3. **`CoreIndicatorName` is hand-written**, not derived — it has to be, since `IndicatorMap` is open by design and the closed set is precisely "what's in `defaultIndicators`". The test pins the two together with a `Record<CoreIndicatorName, true>`, so adding a name to the type without shipping a default fails to compile in the test rather than at some consumer's render. I couldn't find a derivation that stays correct under augmentation; if you know one, it'd be strictly better.

4. **The codemod widens scope and says so** rather than pretending the rename is equivalent. A `dropdown-menu-radio-dot` rule becomes an all-radio-dots rule, because there is no menu-only dot element left. I considered emitting the old class as a `legacyName` on the indicator's dot instead — rejected: it would put `astryx-dropdown-menu-radio-dot` on every radio dot in the app, which is the same widening with none of the visibility.

5. **What I did not do:** the other 7 core files using `borderRadius: '50%'`. `50%` and `--radius-full` diverge the moment a box isn't square (measured: 60×24 differs by 23.5%), and I'd want each one checked rather than swept.

Fixed during self-review (`1ddfd92cab`): the children branch spread `...rest` *last*, so a caller could override the indicator's own `aria-hidden` — the opposite of the glyph path two lines below, which spreads first and explains why.

kentonquatman (approved):
This looks correct. Styling for radio buttons should driven by the theme tokens.

PR #4891, opened 2026-08-11
fix(lab): refresh Schedule's clock snapshot when the first subscriber attaches

alex-js-ltd:
## Problem

`useCurrentTime` retains a stale snapshot while no Schedule view is mounted. Opening or returning to Schedule can therefore show an hours-old clock until the next 60-second tick.

This can misplace or hide the now-indicator (`getCurrentTimeTop` returns `null` once the stale time crosses into another day — `TimeGridView.tsx:579`) and incorrectly position the List view's now-divider.

## Fix

Refresh the cached snapshot when the first subscriber starts a new interval. `getSnapshot()` remains a pure read, and React detects the refreshed value through `useSyncExternalStore`.

## Tests

Added regressions covering:

- first mount after an earlier module load
- remount after the last subscriber unmounts

Both fail on `main` — 24h and 8h stale respectively — and pass with this change.

### DIFF
--- a/packages/lab/src/Schedule/useCurrentTime.test.ts
+++ b/packages/lab/src/Schedule/useCurrentTime.test.ts
@@ -1,0 +1,49 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+/**
+ * @file useCurrentTime.test.ts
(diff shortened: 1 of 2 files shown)


cixzhang (approved):
Thanks this seems reasonable.

PR #4896, opened 2026-08-11
NumberInput: add formatted display, wheel control, and steppers

czarandy:
## Summary

- replace the native number input with a text-backed ARIA spinbutton and numeric input mode
- add `formatValue` for formatted values at rest while keeping the raw number editable on focus
- preserve invalid pending text separately from the committed numeric value
- implement decimal-safe ArrowUp/ArrowDown and wheel stepping with min/max clamping
- add `isWheelEnabled`, defaulting to `true` for compatibility with Astryx's current behavior; `false` leaves page scrolling untouched
- add opt-in, flush-right increment/decrement controls through `hasNumberSteppers`, defaulting to `false`
- keep formatted form submissions numeric and expose formatted values through `aria-valuetext`

This addresses the display, spinner, and wheel-behavior portions of #4186. It intentionally leaves the async action API to #4390.

## Dependency

- #4901 should merge first so the callback-ref wheel listener gets the corrected merged-ref cleanup lifecycle.

## Screenshots

Formatted display:
image

Optional number steppers:
image

## Compatibility

The wheel default remains enabled because the current NumberInput already steps focused values with the wheel. Consumers can now opt out without changing existing behavior. Number steppers remain visually opt-in to match the current component default.

## Storybook

- Formatted Display
- With Number Steppers
- Wheel Behavior
- updated NumberInput showcase block

The new stepper story was visually checked at all three control sizes; the stack is aligned directly against the trailing inner border with no residual inline padding.

## Verification

- `pnpm test` — 489 files, 9,732 tests passed
- `pnpm lint:strict` — 0 errors (existing repository warnings only)
- `pnpm build`
- `pnpm storybook:build`
- NumberInput accessibility audit — 28 stories, 0 violations
- NumberInput + Pagination focused suites — 194 tests passed


kentonquatman (approved):
Thanks @czarandy! This looks good! It would be nice if the text could be centered in the number input for scenarios like the pagination, but we can revisit that in a follow-up PR.

cixzhang (approved):
Oh nice thanks for adding this! May need to rebase for merge conflicts

PR #4901, opened 2026-08-11
fix(core): clean up merged refs consistently

czarandy:
## Summary

- make `mergeRefs` clear object refs when React runs the merged cleanup
- preserve the legacy `ref(null)` notification for callback refs that do not return their own cleanup
- add React-rendered regression tests for callback and object ref lifecycles

## Why

When any callback ref returns a React 19 cleanup function, React invokes the merged cleanup instead of calling the merged ref with `null`. The existing implementation forwarded callback cleanups but left object refs pointing at detached nodes and skipped `null` for callbacks without cleanup functions.

## Test plan

- `pnpm vitest run packages/core/src/utils/mergeRefs.test.ts`
- `pnpm exec eslint packages/core/src/utils/mergeRefs.ts packages/core/src/utils/mergeRefs.test.ts`

### DIFF
--- a/.changeset/merge-refs-react-cleanup.md
+++ b/.changeset/merge-refs-react-cleanup.md
@@ -1,0 +1,8 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Fix `mergeRefs` cleanup so object refs are cleared and callback refs without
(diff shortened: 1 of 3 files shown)


cixzhang (approved):
Yes please!

Issue #4892, opened 2026-08-11
i18n: systematically guard IME composition across text-input components


nynexman4464:
## Problem

Text-input components that handle `keydown` (Enter/Escape/arrows) can misfire while an IME (Korean/Japanese/Chinese) composition is live. Pressing Enter to *commit a composition* is misread as a command (accept suggestion, save, submit), because the browser fires `keydown` with `isComposing: true` (or legacy `keyCode: 229`) BEFORE `compositionend` writes the pending syllable.

This has now been fixed reactively in three places, one at a time:
- `ChatComposerInput` (earliest)
- `BaseTypeahead` — and everything built on it (Typeahead, Tokenizer, PowerSearch free-text search) — via #4860 (closes #4828)
- `PowerSearchEditPopover`'s Enter-to-save (also #4860)

The shared helper already exists: `isImeKeyEvent` in `useFocusTrap.ts`. The gap is that **there's no systematic guarantee** every keydown-handling text input uses it — each site has been patched only after a user hit the bug.

## Goal

Make IME-composition safety a property we enforce, not a bug we keep rediscovering. Audit every component that reads Enter/Escape/arrow/Home/End on a text input or contenteditable and confirm it guards with `isImeKeyEvent` (or equivalent). Candidates to check beyond the three already fixed: TextArea/TextInput (if they have keydown commit behavior), SearchInput, ComboBox/Selector search fields, NumberInput, DateTimeInput segment editing, CommandPalette input, any Chat inputs, and anything else consuming `onKeyDown` on an editable surface.

## Possible approaches (decide during investigation)

1. **A lint rule** — flag `onKeyDown` handlers on text inputs that reference `e.key === 'Enter'`/`'Escape'`/arrow keys without an `isImeKeyEvent`/`isComposing` guard. Highest leverage; catches new code (incl. AI-generated) at author time.
2. **An interaction test / audit dimension** — fire a composing Enter (`isComposing: true` and `keyCode: 229`) at each text-input component and assert it does NOT trigger the command action, then a real Enter does. Mirrors how #4860 tested it.
3. **A shared input primitive/hook** that bakes in the guard, so components can't forget it.

## Context

Not an RTL bug (CJK is LTR) and not caught by the `pr-rtl` audit, which is scoped to layout-direction geometry. This is the input-method dimension of i18n — a distinct class the current tooling doesn't cover. Surfaced by @cixzhang. Related reactive fixes: #4860, #4828.

nynexman4464:
[...]

PR #4907, opened 2026-08-11
refactor(core): consolidate the IME composition guard into utils/ime

nynexman4464:
## Why

The IME composition guard — the check that a keydown is committing/cancelling a Korean/Japanese/Chinese composition and must not be treated as a command — was implemented three different ways across the codebase: a shared `isImeKeyEvent` predicate (whose canonical definition oddly lived inside `hooks/useFocusTrap.ts`), hand-inlined `e.isComposing || e.keyCode === 229` checks, and one ref-tracking variant. That scattering is why the same bug keeps getting rediscovered one component at a time (#4860, and now #4892).

This is the groundwork: one canonical primitive, in a discoverable home, that everything points at — including the new lint rule stacked on top of this.

## What

- New `utils/ime.ts` exporting `isImeKeyEvent`, with the single authoritative doc comment explaining the keydown-before-compositionend ordering and why the legacy `keyCode === 229` fallback is still load-bearing. Re-exported from the utils barrel.
- `useFocusTrap.ts` keeps a deprecated re-export for back-compat; the three relative importers and the three inline sites now all call the shared predicate.

## Risk

None intended — pure consolidation, no behavior change. The predicate is byte-for-byte the same logic; call sites are unchanged in what they guard.

## Testing

- New `utils/ime.test.ts` (both signals: `isComposing` and `keyCode 229`).
- All affected component suites pass (Typeahead, PowerSearch, Dialog, ContextMenu, Tooltip, Chat composer, focus-trap).
- `tsc` clean; `eslint` clean on changed files.

Design notes and the full site inventory in the follow-up issue. First of a 3-PR stack (this → editable-field fixes → lint rule).

@cixzhang

### DIFF
--- a/.changeset/ime-guard-utils-home.md
+++ b/.changeset/ime-guard-utils-home.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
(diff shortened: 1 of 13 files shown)


cixzhang (approved):

cixzhang:
Looks good. We should fix build

PR #4914, opened 2026-08-11
fix(ci): fall back to two-dot diff in analyze-pr when merge base is missing

### LINKED ISSUE #4912: ci: build-storybook fails with 'no merge base' on PRs older than 50 commits
## Summary

The `build-storybook` job in CI fails with `no merge base` on PRs whose branch base is more than 50 commits behind `main`. The `Analyze components` step runs `analyze-pr.js`, which does a three-dot diff (`git diff origin/main...HEAD`), and the shallow clone (`fetch-depth: 50`) can't find a merge base once the PR has been open long enough for main to move past that window.

## Repro

On a PR opened a while ago (branch base older than ~50 commits vs current `main`):

1. CI runs `build-storybook`.
2. `Analyze components` step:
   ```
   node .github/scripts/analyze-pr.js --base origin/main --head HEAD --output analysis.json
   ```
3. Fails:
   ```
   fatal: origin/main...HEAD: no merge base
   ##[error]git diff origin/main...HEAD failed
   The clone is likely too shallow to contain the merge base (see fetch-depth in ci.yml).
   ```

Real example: PR #4298 (branch base Aug 3, `main` at Aug 11) - `build-storybook` fails with exactly this error while `check-scope` (same `fetch-depth: 50`) passes because it swallows the error and assumes components changed.

## Root cause

`fetch-depth: 50` is a fixed window. It works for fresh PRs (merge base within 50 commits) but breaks for PRs that have been open long enough for `main` to advance past that window. The three-dot diff needs the actual merge base, which a depth-50 clone doesn't have once the PR ages.

## Suggested fix

Options, in order of preference:

1. **Unshallow the clone** in `build-storybook` before the diff: `git fetch --unshallow origin` (or `--shallow-since`) so the merge base is always present. Cost: larger clone, but only for this job.
2. **Bump `fetch-depth`** to a larger value (e.g. 200) - still a fixed window, just wider; same failure mode eventually.
3. **Fall back to a two-dot diff** in `analyze-pr.js` when the three-dot fails (`git diff origin/main HEAD`), which doesn't need a merge base. Slightly less precise (includes commits already merged to main) but never hard-fails.

Happy to open a PR for option 1 or 3 if that's preferred.

gonzoblasco:
## Summary

`analyze-pr.js` hard-fails the `build-storybook` job with `no merge base` when a PR's branch base is more than 50 commits behind `main`. The three-dot diff (`git diff base...head`) needs a merge base, which a shallow clone (`fetch-depth: 50`) doesn't have once the PR has been open long enough for `main` to advance past that window.

This makes CI fail on older PRs through no fault of the PR itself - the code is fine, the clone just can't find the merge base. See #4912 for the full diagnosis.

## Change

In `getChangedFiles()`, when the three-dot diff fails (no merge base), the script first tries to **recover the merge base** by deepening the shallow clone (`git fetch --deepen=200`) and retrying the three-dot diff. Only if that still can't find a merge base does it fall back to a two-dot diff (`git diff base..head`), which needs no merge base and never hard-fails.

The three-dot diff remains the preferred path (precise, excludes base-only commits) - the fallback only kicks in as a genuine last resort.

**Important caveat about the two-dot fallback:** it is inherently approximate. Because it compares two trees, a file whose change also exists on `base` (a cherry-picked/duplicated fix, or the same codegen output landing from another PR) can **vanish from the diff entirely** - so two-dot can under-report a PR's own changes, not just over-report base churn. Since `pr-a11y` derives its component list from this file list, an under-reported component silently skips the audit. To make this visible, the script records `diffMode` (`three-dot` | `two-dot`) in `analysis.json`, and the PR report shows a one-line caveat whenever the two-dot fallback was used.

## Verification

Reproduced the exact failure in a shallow clone where the merge base is outside the fetch window:

- Before: `git diff origin/main...HEAD` → `fatal: no merge base`, job dies
- After: script deepens the clone, retries and resolves the three-dot diff, and reports exactly the component the branch touched (exit 0, `diffMode: three-dot`)

Added regression tests:
- `analyze-pr.test.mjs` - builds a synthetic repo with no merge base and asserts the script recovers to the exact three-dot result (`modifiedComponents: ['Card']`, exit 0)
- `generate-pr-comment.test.mjs` - asserts the report shows the approximate-analysis caveat when `diffMode` is `two-dot` and stays clean otherwise

Fixes #4912

### DIFF
--- a/.github/scripts/analyze-pr.js
+++ b/.github/scripts/analyze-pr.js
@@ -84,20 +84,39 @@ function componentIndexRef(pkg, componentName) {
     : `${pkgSrc(pkg)}/${componentName}/index.ts`;
 }
 
-// Get changed files between base and head
+// Get changed files between base and head.
+//
+// Prefers a three-dot diff (base...head), which only reports files changed on
+// the PR branch and excludes commits already merged to base. But a three-dot
+// diff needs a merge base, which a shallow clone (fetch-depth: 50) may not
+// have once the PR has been open long enough for base to advance past that
+// window — the diff then fails with "no merge base" and the whole job dies.
+//
+// In that case we fall back to a two-dot diff (base..head), which needs no
+// merge base and never hard-fails the CI. The two-dot diff may include files
+// merged to base since the branch point, which is acceptable for component
+// analysis: it can over-report files already on base, but never under-reports
+// the PR's own changes.
 function getChangedFiles() {
+  const threeDot = `${baseBranch}...${headRef}`;
   try {
-    const output = execSync(
-      `git diff --name-only ${baseBranch}...${headRef}`,
-      { encoding: 'utf8' }
-    );
+    const output = execSync(`git diff --name-only ${threeDot}`, { encoding: 'utf8' });
     return output.trim().split('\n').filter(Boolean);
   } catch (e) {
-    // A failed diff (e.g. no merge base on a too-shallow clone) is not the
+    console.warn(
+      `::warning::three-dot diff ${threeDot} failed (${e.message.split('\n')[0]}) - falling back to two-dot diff`
+    );
+  }
+  const twoDot = `${baseBranch}..${headRef}`;
+  try {
+    const output = execSync(`git diff --name-only ${twoDot}`, { encoding: 'utf8' });
+    return output.trim().split('\n').filter(Boolean);
+  } catch (e) {
+    // A failed diff (e.g. base ref missing on a too-shallow clone) is not the
     // same as "no changes" — fail loudly instead of publishing an empty
     // analysis report that looks legitimate.
-    console.error(`::error::git diff ${baseBranch}...${headRef} failed: ${e.message}`);
-    console.error('The clone is likely too shallow to contain the merge base (see fetch-depth in ci.yml).');
+    console.error(`::error::git diff ${twoDot} failed: ${e.message}`);
+    console.error('The clone is likely too shallow to contain the base ref (see fetch-depth in ci.yml).');
     process.exit(1);
   }
 }
--- a/.github/scripts/analyze-pr.js
+++ b/.github/scripts/analyze-pr.js
@@ -92,25 +92,64 @@ function componentIndexRef(pkg, componentName) {
 // have once the PR has been open long enough for base to advance past that
 // window — the diff then fails with "no merge base" and the whole job dies.
 //
-// In that case we fall back to a two-dot diff (base..head), which needs no
-// merge base and never hard-fails the CI. The two-dot diff may include files
-// merged to base since the branch point, which is acceptable for component
-// analysis: it can over-report files already on base, but never under-reports
-// the PR's own changes.
+// On that failure we first try to deepen the shallow clone and retry the
+// three-dot diff — the real fix for "base advanced past the fetch window" is
+// to fetch more history, not to degrade. Only if that still cannot find a
+// merge base do we fall back to a two-dot diff (base..head), which needs no
+// merge base and never hard-fails the CI.
+//
+// The two-dot fallback is inherently lossy: it compares two trees, so a file
+// whose change also exists on base (a cherry-picked/duplicated fix, or the
+// same codegen output landing from another PR) vanishes from the diff
+// entirely. That means two-dot can UNDER-report a PR's own changes, not just
+// over-report base churn. Because downstream gates (pr-a11y) derive their
+// component list from this file list, under-reporting silently skips the
+// audit for a component the PR did change. The caller records the diff mode
+// in the analysis output so consumers know when the file list is approximate.
+// Returns { files, diffMode: 'three-dot' | 'two-dot' }.
 function getChangedFiles() {
   const threeDot = `${baseBranch}...${headRef}`;
-  try {
+  const tryThreeDot = () => {
     const output = execSync(`git diff --name-only ${threeDot}`, { encoding: 'utf8' });
     return output.trim().split('\n').filter(Boolean);
+  };
+  try {
+    return { files: tryThreeDot(), diffMode: 'three-dot' };
   } catch (e) {
     console.warn(
-      `::warning::three-dot diff ${threeDot} failed (${e.message.split('\n')[0]}) - falling back to two-dot diff`
+      `::warning::three-dot diff ${threeDot} failed - attempting to deepen the clone and retry`
+    );
+    console.warn(diffErrorDetail(e));
+  }
+
+  // The three-dot diff failed, almost certainly because the shallow clone is
+  // too shallow to contain the merge base. Deepen before giving up on the
+  // accurate path; a deepened clone makes three-dot correct again. The base
+  // arg arrives as "origin/main" (see ci.yml) but the fetch refspec needs the
+  // bare branch name and must update the origin/<base> ref explicitly —
+  // `--deepen` alone only updates FETCH_HEAD, which the three-dot ref
+  // (origin/main...HEAD) needs to actually exist.
+  try {
+    const baseName = baseBranch.replace(/^origin\//, '');
+    execSync(
+      `git fetch --deepen=200 origin ${baseName}:refs/remotes/origin/${baseName}`,
+      { encoding: 'utf8', stdio: 'pipe' },
+    );
+    return { files: tryThreeDot(), diffMode: 'three-dot' };
+  } catch (e) {
+    console.warn(
+      `::warning::deepen failed (${diffErrorDetail(e)}) - falling back to two-dot diff`
     );
   }
+
+  // Last resort: a two-dot diff needs no merge base. Its file list is
+  // approximate (may over-report base churn and, worse, under-report the PR's
+  // own changes when they also exist on base), so the caller marks the output
+  // with diffMode: 'two-dot' for any consumer to caveat.
   const twoDot = `${baseBranch}..${headRef}`;
   try {
     const output = execSync(`git diff --name-only ${twoDot}`, { encoding: 'utf8' });
-    return output.trim().split('\n').filter(Boolean);
+    return { files: output.trim().split('\n').filter(Boolean), diffMode: 'two-dot' };
   } catch (e) {
     // A failed diff (e.g. base ref missing on a too-shallow clone) is not the
     // same as "no changes" — fail loudly instead of publishing an empty
@@ -119,6 +158,13 @@ function getChangedFiles() {
     console.error('The clone is likely too shallow to contain the base ref (see fetch-depth in ci.yml).');
     process.exit(1);
   }
+}
+
+// Human-readable reason for a failed diff/deepen, preferring the fatal stderr
+// line (e.g. "fatal: origin/main...HEAD: no merge base") over execSync's
+// boilerplate "Command failed: ..." first line.
+function diffErrorDetail(e) {
+  return (e.stderr && e.stderr.toString().trim()) || e.message.split('\n')[0];
 }
 
 // Check if a component exists in the base branch.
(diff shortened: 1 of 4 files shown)


cixzhang (changes requested):
🔴 **Code review required** — `needs:code-review` + `community` labels are on this PR (`REVIEW_GATE.md`: all contributor PRs require code review). No blocking rubric violation in the diff itself; the findings below are 🟡.

`Triage: bug fix (CI tooling) · non-breaking · blast radius = every PR's analysis artifact → fast path · checks: none of the Component Audit Rubric's trigger rows fire (no packages/** change; §7 is React/DOM code health) — judged against the rubric's bug-fix bar ("evidence that it was broken before and is fixed now") plus the correctness of the new CI behavior.`

Thanks for chasing this one down to a real diagnosis in #4912 — the writeup there is better than most bug reports we get, and the fix is aimed at the right function.

**What I verified.** I did not run this in CI (workflows haven't been approved on this PR yet, and this branch has a fresh merge base so CI would never take the new path anyway). I built a synthetic repo — branch point, 60 commits on `main`, `--depth=50` clone — where the PR branch touches **only** `packages/core/src/Card/`, and ran both the `main` and this branch's version of `analyze-pr.js` against it.

- On `main`: `::error::` + exit 1. Reproduced. ✅
- On this branch: warning, fallback, exit 0, `analysis.json` written. The fix works. ✅
- But the resulting `analysis.json` says: `modifiedComponents: ["Button", "Card", "Line"]`, `changedPackages: ["@astryxdesign/core", "@astryxdesign/charts"]`. The truth was `["Card"]` and core only.

### 🟡 1. The under-report case is real, and the description says it can't happen

The description and the new comment (lines 96–99) both say the two-dot diff "can over-report files already on base, but **never under-reports** the PR's own changes." That isn't true — two-dot compares two trees, so if the PR's change *also* exists on base, the file vanishes from the diff entirely. I confirmed this: after landing an identical change on `main`, `Card` — the only file this PR actually touched — dropped out of the two-dot list completely.

That's not hypothetical in this repo: a cherry-picked or duplicated fix, or the same formatting/codegen output landing from another PR, produces it. And it matters more than over-reporting, because it makes a gate *narrower*: `pr-a11y` derives `--components` from `(.newComponents + .modifiedComponents)` (`ci.yml:428`) and runs with `--fail-on-new`, so an under-reported component is one the audit silently stops checking on a PR that did change it.

Please fix the claim in both the PR description and the code comment, whatever else changes.

### 🟡 2. Try to recover the merge base before degrading

The root cause is two lines of `ci.yml`, and the fallback works around them rather than fixing them:

```yaml
fetch-depth: 50                                   # ci.yml:171
run: git fetch origin ${{ github.base_ref }} --depth=50   # ci.yml:175
```

Option 1 in #4912 (unshallow / deepen) is still the better fix, and it's cheap. Suggested ordering inside `getChangedFiles()` — degrade only after trying to become correct:

1. three-dot (preferred, unchanged);
2. on failure, `git fetch --deepen=200 origin <base>` (or `--unshallow`) and **retry three-dot**;
3. only then two-dot.

With step 2 in place, the two-dot path becomes a genuine last resort and findings 1 and 3 mostly stop mattering in practice.

### 🟡 3. The degradation is announced to the job log, but not to anything that consumes the result

`::warning::` (line 106) lands in the Actions log; `analysis.json` carries no trace of it. Everything downstream then treats an approximate file list as exact:

- `pr-a11y` — `--components` from the analysis, `--fail-on-new` (`ci.yml:428`). Over-reported components get audited: since the baseline is keyed `Component::Story::rule`, a violation that landed on `main` after the last baseline refresh — or a story renamed on `main` — surfaces as *new* and reds this PR for someone else's change. Under-reported components (finding 1) skip the audit.
- `pr-rtl` — same list (`ci.yml:494`); soft-gated, so cost only.
- The **PR Analysis Report** comment itself — readers see "Modified: Button, Card, Line" and bundle rows for `@astryxdesign/charts` with no hint that any of it is approximate.

One field would fix the whole class: return `{files, diffMode}` and write `"diffMode": "two-dot"` (or `"degraded": true`) into `analysis.json`, then have `generate-pr-comment.js` print a one-line caveat when it's set. Follow-ups for the gating jobs can be separate issues — not yours to fix here.

Related, and pre-existing (**not yours to fix**, flagging so it isn't mistaken for new): `changedPackages.add(pkg.name)` at line 486 runs *before* the `allComponents.includes(...)` filter at line 494, so any file under a package's `src/` — including base-only churn — adds that package to the bundle-size section. It's what put `@astryxdesign/charts` in my repro's output.

### 🟡 4. The new warning is less informative than the error it replaces

Line 107 does `e.message.split('\n')[0]`. For `execSync`, that first line is `Command failed: git diff --name-only origin/main...HEAD` — the actual reason, `fatal: origin/main...HEAD: no merge base`, is on the *next* line and gets cut. Verbatim from my run:

```
::warning::three-dot diff origin/main...HEAD failed (Command failed: git diff --name-only origin/main...HEAD) - falling back to two-dot diff
```

The annotation now says only "it failed." (Git's `fatal:` is still in the raw log, just not in the annotation, and not where someone scanning the summary will see it.) `e.stderr?.toString().trim() || e.message` reads better and keeps the diagnosis.

### 🟡 5. A regression test has a home here

`.github/scripts/**/*.test.{ts,tsx,mjs}` is in the `node` project's include list in `vitest.config.ts`, so `pnpm test` (the CI `test` job) picks one up automatically, and `.github/scripts/accessibility-audit.test.mjs` is the pattern to copy — temp dir, `execFileSync` the script, assert on the JSON. A test that builds a small repo with no merge base and asserts exit 0 plus the expected `modifiedComponents` is roughly 40 lines; I wrote the equivalent by hand to review this, so I know it's not a big ask.

Worth the effort specifically because **this PR's own CI can't exercise the fix** — this branch has a healthy merge base, so CI takes the three-dot path and the new code never runs. Per `copilot-instructions.md` this is an ask, not a block ("do not block on test-only scaffolding"), but the rubric's bug-fix bar wants "a test that was red and is now green," and here that's the only way anyone will know if the path still works in six months.

### 🟢 What's right

- Three-dot stays the preferred path; the fallback is genuinely a fallback.
- The original guard's intent survives: a truly missing base ref still exits 1 loudly (lines 114–121) rather than publishing an empty analysis that looks legitimate.
- The comment block explains the tradeoff instead of leaving a bare `catch`.
- `::warning::` on degradation matches the existing house pattern in `check-scope` (`ci.yml:100`).
- No changeset needed — CI tooling isn't consumer-visible, so `X20` doesn't fire. No `.github/instructions/*` file matches `.github/scripts/**` (they scope to `packages/**`, `apps/docsite/**`, blog posts, and CLI templates), so none applied.

### Verdict

**Request changes**, all small: correct the "never under-reports" claim (1), and either add the deepen-and-retry (2) or the `diffMode` marker (3) — (2) is the one I'd pick. (4) is a one-liner. (5) is the ask I'd most like to see and the reason is in the finding itself. The shape of the fix is right and I'd be glad to see it land once the file list stops claiming a precision it doesn't have.

*Note for maintainers: CI hasn't run here — the fork PR needs workflow approval. The red Vercel check is the outside-contributor preview deploy needing team authorization, not a failure of this code.*

gonzoblasco:
Thanks for the thorough review - all five findings addressed.

**1. Corrected the "never under-reports" claim.** Updated both the PR description and the code comment: the two-dot fallback can under-report a PR's own changes when they also exist on base (cherry-pick / duplicated fix / shared codegen output). Your repro is spot on - the component vanishes from the tree-to-tree diff entirely.

**2. Added deepen-and-retry.** `getChangedFiles()` now degrades only after trying to become correct: three-dot → `git fetch --deepen=200 origin <base>:refs/remotes/origin/<base>` and retry three-dot → two-dot as a genuine last resort. Two notes from building this:
- The fetch refspec needs the bare branch name, since `--base` arrives as `origin/main` (see ci.yml:251) - naively interpolating `origin/${baseBranch}` produces `origin origin/main:...` which fails. The script strips the `origin/` prefix.
- `--deepen` alone only updates `FETCH_HEAD`, so the fetch must write `refs/remotes/origin/<base>` explicitly, or the three-dot retry (which references `origin/main`) still can't resolve.

I verified with your exact scenario (branch point + 60 commits on main + feature touching only `packages/core/src/Card/`, `--depth` clone): with the deepen in place, the script recovers the **exact** result (`modifiedComponents: ["Card"]`, `diffMode: "three-dot"`) instead of the over-reported `["Button", "Card", "Line"]` the fallback alone produced. This is why I picked (2) over (3) as the primary fix.

**3. Added the diffMode marker.** `analysis.json` now carries `diffMode: "three-dot" | "two-dot"`, and `generate-pr-comment.js` renders a one-line caveat when it's `two-dot`. Older analysis files (no field) render clean, so back-compat holds.

**4. Better warning.** `diffErrorDetail()` prefers `e.stderr?.toString().trim()`, so the annotation now shows `fatal: ...: no merge base` instead of just "Command failed: git diff ...".

**5. Regression tests added**, both following the `accessibility-audit.test.mjs` pattern (temp dir + execFileSync + assert):
- `analyze-pr.test.mjs` - builds the synthetic no-merge-base repo, asserts the precondition (merge-base fails), then asserts exit 0 + `diffMode: "three-dot"` + `modifiedComponents: ["Card"]`.
- `generate-pr-comment.test.mjs` - asserts the caveat shows for `two-dot` and stays clean for `three-dot` / missing field.

Both pass against the repo's vitest config (`.github/scripts/**/*.test.mjs` is in the node project include). Since this branch has a healthy merge base, CI will take the three-dot path and won't exercise the new code - these tests are the only thing that proves the recovery still works.

gonzoblasco:
Hi @cixzhang - friendly ping on this one.

All five findings from the review on Aug 12 are addressed in the two commits pushed since then:

1. Corrected the "never under-reports" claim in both the PR description and the code comment.
2. Added deepen-and-retry: `getChangedFiles()` now tries `git fetch --deepen=200` and retries the three-dot diff before degrading to two-dot.
3. `analysis.json` now carries `diffMode` so downstream jobs and the PR Analysis Report can flag an approximate file list.
4. The warning now surfaces `e.stderr` instead of truncating the first line, so the `fatal:` reason survives.
5. Two regression tests (analyze-pr + generate-pr-comment) build a synthetic no-merge-base repo and prove the recovery path still works.

No rush, just making sure it's on your radar whenever you get a chance to take another look. Thanks!

.github/scripts/analyze-pr.js
```diff
@@ -84,24 +84,89 @@ function componentIndexRef(pkg, componentName) {
...
+  // too shallow to contain the merge base. Deepen before giving up on the
+  // accurate path; a deepened clone makes three-dot correct again. The base
+  // arg arrives as "origin/main" (see ci.yml) but the fetch refspec needs the
+  // bare branch name and must update the origin/<base> ref explicitly —
+  // `--deepen` alone only updates FETCH_HEAD, which the three-dot ref
+  // (origin/main...HEAD) needs to actually exist.
+  try {
+    const baseName = baseBranch.replace(/^origin\//, '');
+    execSync(
+      `git fetch --deepen=200 origin ${baseName}:refs/remotes/origin/${baseName}`,
```
cixzhang, line 135: `--deepen=200` tops out 250 commits behind base. Might want to retry the deepen.

.github/scripts/analyze-pr.test.mjs
```diff
@@ -0,0 +1,121 @@
...
+  // Feature branch off the branch point, touching only Card.
+  git(upstream, ['branch', 'feature', branchPoint]);
+  git(upstream, ['checkout', '-q', 'feature']);
+  fs.appendFileSync(path.join(upstream, 'packages/core/src/Card/index.ts'), 'export const Card = {}\n');
+  git(upstream, ['add', '-A']);
+  git(upstream, ['commit', '-qm', 'touch Card only']);
+
+  // Shallow single-branch clone of the feature branch.
+  git(null, [
+    'clone', '-q', '--depth=5', '--single-branch', '--branch=feature',
```
cixzhang, line 88: Single-branch clone has no `origin/main` — this fails on a missing ref, not merge base.

cixzhang (approved):
Thanks, all five addressed — merging. `--deepen=200` reaches 250 commits behind base though, and 55 of the 122 stale PRs are further back; worth a second deepen as a follow-up?

[Reviewed by Robohands]

PR #4916, opened 2026-08-11
fix(core): TopNavMegaMenu anchors to its trigger outside a <nav> (#4905)

### LINKED ISSUE #4905: [Bug] TopNavMegaMenu dropdown renders at viewport top-left in Properties preview
## Description

Opening the TopNavMegaMenu in its Properties preview renders the dropdown at the top-left corner of the page, overlapping the docsite navigation.

Expected: the dropdown should appear below its “Navigation” trigger and remain associated with the preview area.

## Reproduction

1. Open https://astryx.atmeta.com/components/TopNavMegaMenu?tab=properties
2. Click the “Navigation” trigger.
3. Observe that the “Item” dropdown appears at the viewport top-left corner.

## Astryx Version

@astryxdesign/core@0.3.0

## Additional context

The trigger itself is centered in the Properties preview, but the opened panel is detached from it and positioned at the page origin. TopNavMegaMenu anchors its panel using the nearest `<nav>`, so the generated preview or its wrapper/anchor setup may not provide valid positioning geometry.

is-jain:
Fixes #4905.

## Problem

Opening TopNavMegaMenu in the docsite Properties preview renders the dropdown at the viewport top-left corner, detached from its trigger.

## Root cause

`DefaultMegaMenu` sets its CSS anchor with `triggerButtonRef.current?.closest('nav')` and only calls `popover.triggerRef(...)` when a `<nav>` ancestor is found. The Properties preview renders the component standalone, so no element ever carries the layer's `anchor-name`. The panel still gets `position-anchor`/`position-area` from `useLayer`, and with no matching anchor `position-area` computes to `none` — which pins the top-layer popover to the viewport corner because `useLayer`'s base styles zero the UA popover margins (the exact failure mode already documented in `useLayer.tsx`'s `getPositionArea` comment).

## Fix

Fall back to anchoring on the trigger button itself when there is no `<nav>` ancestor:

```tsx
const trigger = triggerButtonRef.current;
popover.triggerRef(trigger?.closest('nav') ?? trigger ?? null);
```

Inside a TopNav nothing changes — the panel still anchors to the surrounding `<nav>` and spans below the bar. Standalone, the panel now opens below its trigger and stays associated with the preview area, which is the behavior the issue expects.

## Tests

Added an "anchor wiring" describe block to `TopNavMegaMenu.test.tsx`:

- inside a `<nav>`: the `<nav>` carries the layer anchor name, the trigger does not (existing behavior preserved);
- without a `<nav>` ancestor: the trigger itself carries the anchor name, and the popup's `position-anchor` matches it.

`vitest run packages/core/src/TopNav/TopNavMegaMenu.test.tsx` — 39 tests pass. All pre-commit `check:*` scripts pass; changeset included (patch, `@astryxdesign/core`).


packages/core/src/TopNav/TopNavMegaMenu.tsx
```diff
@@ -419,12 +419,14 @@ function DefaultMegaMenu({
...
+  // <nav> ancestor (e.g. the menu rendered standalone in a docs preview), fall
+  // back to the trigger button itself — if no element carries the anchor name,
+  // the panel's position-area computes to `none` and the top-layer popover
+  // pins to the viewport corner (#4905).
   useEffect(() => {
-    const nav = triggerButtonRef.current?.closest('nav');
-    if (nav) {
-      popover.triggerRef(nav);
-    }
+    const trigger = triggerButtonRef.current;
```
imdreamrunner: I am leaning towards fixing the examples rather than the source itself. The example shall render with a <TopNav>. What do you think

code pointer: packages/core/src/TopNav/TopNavMegaMenu.doc.mjs

imdreamrunner (changes requested):

packages/core/src/TopNav/TopNavMegaMenu.tsx
is-jain: Yeah thats probably cleaner, I just updated the PR

Added the playground wrapper to TopNavMegaMenu.doc.mjs and reverted the source change

imdreamrunner (approved):

PR #4926, opened 2026-08-12
fix(core): ComplexSelector focus ring is keyboard-only, not :focus-within (#4922)

### LINKED ISSUE #4922: ComplexSelector: focus ring paints on mouse click and stays after the popup closes (:focus-within, not :has(:focus-visible))
### Summary

`ComplexSelector` paints its field focus ring on `:focus-within`, so the ring appears
after a **mouse** interaction and stays lit after the popup closes — until focus leaves
the control entirely. Every other component in the package paints its ring on a
keyboard-only selector (`:focus-visible` or `:has(:focus-visible)`).

Reported by a designer as: *"very prominent selected state and it seems to stay selected
after unselecting it. You have to click somewhere else to get rid of it."*

### Version

- `@astryxdesign/core@0.3.0` (currently `latest` on npm) — reproduces.
- The published tarball's `dist/astryx.css` and `dist/ComplexSelector/ComplexSelector.js`
  are byte-identical to what is installed, so this is the shipped package, not a local build.
- The newest `canary` build carries the same rule and the same sole user of it, so `main`
  is affected too.

### Steps to reproduce

Paste into a fresh Astryx app and render it:

```tsx
import { useState } from 'react';
import { ComplexSelector } from '@astryxdesign/core';

export default function Repro() {
  const [value, setValue] = useState('Apple');
  return (
    <ComplexSelector
      label="Fruit"
      value={value}
      onChange={setValue}
      triggerLabel={value}
      data-testid="cs">
      {(_value, onChange, close) => (
        <div style={{ padding: 12 }}>
          <button
            type="button"
            onClick={() => {
              onChange('Banana');
              close();
            }}>
            Banana
          </button>
        </div>
      )}
    </ComplexSelector>
  );
}
```

1. **Click** the trigger with the mouse. The popup opens.
2. **Click** an option (or click the trigger again). The popup closes and focus is
   restored to the trigger — correct behaviour.
3. The 2px accent ring is now painted around the field and stays there. Only clicking
   somewhere else clears it.

### Expected

The ring should behave like every other control in the package: shown for keyboard focus,
not for a mouse click. A mouse-driven open/close cycle should leave no ring.

### Measured

Chromium 1234 (bundled) and Google Chrome, headless and headed — identical results.
Measured on the field element: `getComputedStyle(field).outlineStyle/outlineWidth`, plus
`field.matches(':focus-within')` and `field.matches(':has(:focus-visible)')`.

**ComplexSelector**

| # | moment | ring | `:focus-within` | `:has(:focus-visible)` |
|---|--------|------|-----------------|------------------------|
| a | at rest | none | false | false |
| b | mouse click opens popup | none (focus is inside the popup) | false | false |
| c | **mouse click closes popup** | **PAINTED** `solid 2px @ 2px offset` | **true** | **false** |
| c2 | mouse click on the trigger toggles it closed | PAINTED `solid 2px @ 2px offset` | true | false |
| d | real keyboard Tab onto trigger | PAINTED `solid 2px @ 2px offset` | true | true |
| e | click elsewhere | none | false | false |

**Selector, `ghost` variant — the control, same interaction sequence**

| # | moment | ring | `:focus-within` | `:has(:focus-visible)` |
|---|--------|------|-----------------|------------------------|
| a | at rest | none | false | false |
| b | mouse click opens popup | none | true | false |
| c | **mouse click closes popup** | **none** | true | **false** |
| d | real keyboard Tab onto trigger | PAINTED `solid 2px @ 3px offset` | true | true |
| e | click elsewhere | none | false | false |

**Selector, `input` variant** — never paints an outline in any of the four states; its
focus affordance is the shared field border/glow only.

Pixel check on the 6px band just outside the field box, comparing at-rest against
after-mouse-close screenshots: **1449 pixels change for `ComplexSelector`, 0 for both
`Selector` variants.**

### The CSS

`ComplexSelector` applies a private `focusRing` style on the field wrapper:

```css
.x1oqel4m:focus-within:not(#\#):not(#\#)          { outline: 2px solid var(--color-accent) }
.x1fanpfn:focus-within:not(#\#):not(#\#):not(#\#) { outline-offset: 2px }
```

Every sibling uses a keyboard-only selector, e.g.

```css
.x1irc7jg:has(:focus-visible):not(#\#):not(#\#)          { outline: 2px solid var(--color-accent) }
.xdjuwb3:has(:focus-visible):not(#\#):not(#\#):not(#\#)  { outline-offset: 2px }
.x17nn4n9:focus-visible:not(#\#):not(#\#)                { outline: 2px solid var(--color-accent) }
```

Counting components in `dist/` by the selector that paints their outline, in 0.3.0:

| selector | components |
|----------|-----------|
| `:focus-within` | **1** — ComplexSelector |
| `:has(:focus-visible)` | 11 — CheckboxInput, ClickableCard, Item, MultiSelector, RadioList, SelectableCard, Selector, Switch, Thumbnail, Token, TreeList |
| `:focus-visible` | 19 — Avatar, AvatarGroup, Button, Calendar, CodeBlock, Collapsible, Dialog, Lightbox, Link, Outline, Pagination, ProgressBar, Resizable, SegmentedControl, Slider, TabList, Token, TopNav, TreeList |

28 distinct components paint their ring on a keyboard-only selector. `ComplexSelector` is
the only one on `:focus-within`.

### Why the ring gets stuck

`ComplexSelector` restores focus to the trigger when the popup hides:

```js
const popover = usePopover({
  // ...
  onHide: () => { document.getElementById(triggerId)?.focus(); }
});
```

That is the right thing to do. But the ring is keyed to `:focus-within`, which a
mouse-focused button satisfies, so restoring focus re-lights the ring with no keyboard
involved. The programmatic `.focus()` after a mouse interaction does **not** match
`:focus-visible`, which is why the same sequence leaves the `Selector` ghost variant clean.

The trigger's own `<button>` carries `outline: none`, so the ring is entirely the
wrapper's.

### Suggested fix

Move `ComplexSelector`'s `focusRing` to `:has(:focus-visible)`, matching its siblings.

Verified by overriding the two shipped rules in the page and re-measuring:

| | state (c) mouse close | state (d) keyboard Tab |
|---|---|---|
| shipped | ring PAINTED (bug) | ring PAINTED |
| with `:has(:focus-visible)` | ring **gone** | ring **still PAINTED** |

Same result in bundled Chromium and in Google Chrome, headless and headed. The keyboard
ring survives the fix, which is the part that must not regress.

Two smaller things visible in the same area, if they are worth folding in:

- `ComplexSelector`'s ring sits at `outline-offset: 2px`; the newer shared focus-outline
  utility on `main` uses `3px` and calls that the convention, so this one has drifted.
- On `main` there is now a shared focus-outline utility that 20 components have adopted,
  and its selectors are keyboard-only. `ComplexSelector` does not import it — it still
  carries its own `focusRing`. The consolidation appears to have missed it.

nynexman4464:
## What

`ComplexSelector` was the only component in the package painting its field focus
ring on `:focus-within`. The popover restores focus to the trigger when it
closes (correctly), so a **mouse** open/close cycle re-lit the ring and left it
lit until focus left the control entirely — the "stays selected after
unselecting it" report in #4922.

It now uses the shared focus-outline utility (`:has(:focus-visible)`), which
every other field-and-trigger control already uses. The private `focusRing`
style is gone, so the ring also picks up the conventional `3px` offset it had
drifted from.

## Measured

Chromium, on the field element: `outlineStyle`/`outlineWidth`/`outlineOffset`,
plus `:focus-within` and `:has(:focus-visible)` matches. Same sequence as the
issue, driven with real mouse clicks and real key presses.

| # | moment | before | after |
|---|--------|--------|-------|
| a | at rest | none | none |
| b | mouse click opens popup | none | none |
| c | **mouse click closes popup** | **solid 2px @ 2px** | **none** |
| c2 | **mouse click toggles it closed** | **solid 2px @ 2px** | **none** |
| d | keyboard Tab onto trigger | solid 2px @ 2px | solid 2px @ 3px |
| e | click elsewhere | none | none |
| f | keyboard open (focus in popup) | none | none |
| g | keyboard open, Escape closes | solid 2px @ 2px | solid 2px @ 3px |

The keyboard ring survives in every state that had one (d, g) and is now at the
convention's offset; only the two mouse states change.

## Test

A colocated regression test drives a mouse open/close cycle and asserts that the
compiled rules painting an outline on the field are gated on `:focus-visible`,
never `:focus-within`. jsdom resolves neither `:has()` nor layout, so the
assertion reads the rule text StyleX injects rather than a computed ring. It
fails on `main` with `.x1oqel4m:focus-within…` — the exact rule from the issue —
and passes with the fix.

`pnpm lint` and `pnpm vitest run --project ui` are clean (one unrelated
`packages/lab` timeout flakes under full-suite load on `main` as well, in a
different test each run).

Fixes #4922.

### DIFF
--- a/.changeset/complex-selector-keyboard-focus-ring.md
+++ b/.changeset/complex-selector-keyboard-focus-ring.md
@@ -1,0 +1,9 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] ComplexSelector: the field focus ring is keyboard-only again. It was the one component still painting on `:focus-within`, and since the popover restores focus to the trigger when it closes, a mouse-driven open/close cycle re-lit the ring and left it lit until focus left the control entirely — read by designers as a selected state that would not clear. It now uses the shared focus-outline utility (`:has(:focus-visible)`), like every other field-and-trigger control, which also brings the ring back to the conventional 3px offset it had drifted from. The keyboard ring is unchanged in kind.
+
+Fixes #4922.
+
+@nynexman4464
--- a/packages/core/src/ComplexSelector/ComplexSelector.test.tsx
+++ b/packages/core/src/ComplexSelector/ComplexSelector.test.tsx
@@ -23,6 +23,56 @@ const FRUITS = ['Apple', 'Banana'] as const;
 const RIPENESS = ['Crisp', 'Ripe', 'Juicy'] as const;
 const h = {hidden: true} as const;
 
+/**
+ * Selectors of the compiled rules that PAINT an outline on `el`'s own classes.
+ *
+ * StyleX injects its CSS at runtime under test, so the shipped rule text is
+ * readable here — the only way to check which pseudo-class gates the ring,
+ * since jsdom resolves neither `:has()` nor layout. `outline: none` /
+ * zero-width declarations are skipped: they suppress a ring rather than draw
+ * one.
+ */
+function outlineRulesFor(el: HTMLElement): string[] {
+  const classes = Array.from(el.classList);
+  const sheets = Array.from(document.styleSheets);
+  const cssTexts: string[] = [];
+  for (const sheet of sheets) {
+    try {
+      for (const rule of Array.from(sheet.cssRules)) {
+        cssTexts.push(rule.cssText);
+      }
+    } catch {
+      // ignore sheets jsdom refuses to read
+    }
+  }
+  for (const style of Array.from(document.querySelectorAll('style'))) {
+    // Runtime-injected text is not always exposed as parsed cssRules.
+    cssTexts.push(...(style.textContent || '').split('}').map(s => s + '}'));
+  }
+
+  const paints = /outline(-width|-style)?\s*:\s*(?!none|0|initial)/;
+  const suppresses = /outline(-width|-style)?\s*:\s*(none|0)\b/;
+  const selectors: string[] = [];
+  for (const text of cssTexts) {
+    const selector = text.split('{')[0].trim();
+    if (!selector) {
+      continue;
+    }
+    const onThisElement = classes.some(cls =>
+      new RegExp(`\\.${cls}(?![\\w-])`).test(selector),
+    );
+    if (!onThisElement) {
+      continue;
+    }
+    const body = text.slice(text.indexOf('{'));
+    if (!paints.test(body) || suppresses.test(body)) {
+      continue;
+    }
+    selectors.push(selector);
+  }
+  return selectors;
+}
+
 function FruitGrid({
   value,
   onChange,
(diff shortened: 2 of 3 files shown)


kentonquatman (approved):
Thanks @nynexman4464!

cixzhang:
Closing this — the fix is already on `main`.

`ComplexSelector` switched to the shared focus-outline utility in ce86871aca (part of [#4935](https://github.com/facebook/astryx/pull/4935)), which landed 2026-08-12 as part of the repo-wide sweep making every ring keyboard-only. The trigger wrapper now composes `focusOutlineStyles.focusWithin` — `:has(:focus-visible)` — the hand-rolled `focusRing` style is gone, and the ring picks up the conventional 3px offset. Same fix, same mechanism, arrived at independently. [#4922](https://github.com/facebook/astryx/issues/4922) is closed.

Your diagnosis was exactly right, including the part that makes this component special: `onHide` restores focus to the trigger, so `:focus-within` re-lit itself after a mouse close and stayed lit. That's recorded in the commit message on `main`.

One piece here has no equivalent on `main` and is worth keeping: `outlineRulesFor()`, the helper that reads the injected StyleX rule text to assert *which pseudo-class gates the ring*. jsdom resolves neither `:has()` nor layout, so that's the only way to test this at the unit level rather than in Chromium. If you'd like to land it on its own — as a test utility with the ComplexSelector assertion on top — I'd take that PR.

Thanks for chasing this one down.

PR #4928, opened 2026-08-12
fix(selector): make the dropdown search part of the panel, not a box inside it

cixzhang:
## Why

The `hasSearch` field in the Selector and MultiSelector dropdowns was a bordered `TextInput` placed inside the popup. The popup is already a bordered, elevated surface, so the field drew a **box inside a box** — it read as a control dropped into the menu rather than the menu's own header.

The reference we're matching (and the shape the command palette already uses) puts the search *in* the panel: a leading magnifier, a borderless input, and a hairline separating it from the options.

## What it looks like

All shots are the built component in Chrome, light mode, 2×.

### The focus ring is keyboard-only

Same component, same open panel — the only difference is how it was opened.

| Opened by mouse — no ring | Opened by keyboard — ring |
|---|---|
|  |  |
|  |  |

The ring is an inset accent ring on the search *field*, which is shaped like an option row — same inset from the panel edge, same `--radius-element` corners, same 32px height. Measured against the row below it, the two agree on left inset, right inset, width, height, radius, and content x-position in both components.

### Single select

With a query typed, showing the clear (✕) button:

Sectioned — "Citrus" and "Tropical" are plain secondary headings rather than labeled dividers:

### Multi select

Sectioned:

Select-all — no rule under it; it's the first row of the list, not a section of its own, and the checkbox column already lines it up with the options under it:

## What changed

**The search row.** Leading magnifier, borderless transparent input, the shared `InputClearButton`, and a full-bleed `Divider` between the row and the options. No second border, radius, hover shadow, or focus ring nested inside the popup. In MultiSelector the row now sits **outside** the scrolling list, so it stays put while the options scroll under it (Selector already worked this way).

The row is a new shared internal primitive rather than a copy in each component, so the two selectors can't drift apart.

**Focus.** An inset accent ring on `:has(input:focus-visible)`, drawn on the rounded field, **for keyboard focus only**. The field follows *item* geometry (padding + `--radius-element`, like Typeahead's and the selectors' own option rows) rather than input-wrapper geometry, so the search reads as the first row of the list. On the field rather than the bare `<input>` so the magnifier and clear button stay inside the ring; `:has(input:focus-visible)` rather than `:focus-within` so tabbing to the clear button doesn't re-ring it.

`:focus-visible` is necessary but not sufficient here, which is worth stating because it is easy to assume otherwise: per CSS Selectors 4 a pointer-focused element that *supports text entry* matches `:focus-visible` deliberately, so that a clicked text field still shows where typing will go. Verified in Chromium — a bare `<input>` clicked with a mouse matches it. So the selector keeps `:focus-visible` and gains the one bit it cannot express: which device moved focus. A shared document-level tracker records the last interaction and the ring is applied only when focus was taken by keyboard; the browser's heuristic still decides everything else, including `focus({focusVisible: true})`.

**Section titles → plain headings.** A labeled `Divider` draws line–text–line, which is separator language, and it put a second rule within ~9px of the search row's divider (one full-bleed, one inset). `DropdownMenu` and `CommandPaletteGroup` already render a panel-list group heading as plain left-aligned supporting-size secondary text with no rules; this brings the two selectors onto that pattern, so the title aligns with the option labels under it. The heading also moves *inside* `role="group"` as `aria-hidden` text — the group already carries the title as its accessible name — which stops a stray `role="separator"` from sitting as a direct child of `role="listbox"`.

Nothing about behavior changed: the same combobox wiring (`aria-activedescendant`, `aria-expanded`, `aria-controls`), the same arrow/Enter/Escape/PageUp/PageDown handling, the same Tab-to-clear path, and the same accessible names (`Search options`, `Clear Search options`).

## Risk

- **Themes that styled the dropdown search through `astryx-text-input` lose it**, because the popup no longer renders a TextInput. That coupling was accidental — the panel search is not a form field — but it is the one way this can bite a consumer. Four new targets are the replacement, all documented: `astryx-selector-search`, `astryx-selector-section-heading`, `astryx-multi-selector-search`, `astryx-multi-selector-section-heading`.
- **The section-heading change is a proposal** and is isolated in its own commit (`fe8db55`) — drop it if you want a different answer to the grouping question.

## Testing

- `Selector` + `MultiSelector` suites pass (213 tests), including the existing magnifier, clear-button, refocus-on-clear, and tab-to-clear coverage that guards the behavior this change had to preserve.
- New tests pin the shape: no `astryx-text-input` around the search input, the theme target on the row, the divider between row and listbox in that DOM order, the row outside the scroll container (MultiSelector), no rule under select-all, and the section heading being an `aria-hidden` heading inside the group rather than a separator.
- Focus modality verified in real Chromium across 8 routes: mouse-open, click-into-field, and type-after-mouse-open draw **no** ring; Tab+Enter, Tab+ArrowDown, and Shift+Tab back from the clear button all **do** — in both components. jsdom does not implement `:focus-visible`, so the unit tests assert the modality gate and Chromium verifies the painted ring.
- Rendered in Chrome and measured: the search field is 32px tall with a 5px (Selector) / 4px (MultiSelector) inset and `--radius-element` corners — pixel-identical to the option row beneath it; the focus ring is `inset 0 0 0 2px` accent on the row only; dark mode picks up `#fafafa` text, `#a3a3a3` icon, `rgba(255,255,255,.1)` divider — all from tokens, no hardcoded colors.

`ComplexSelector` has its own in-panel search and was left alone — say the word and I'll bring it along.


cixzhang:
Follow-up pushed — three commits, one per item.

**Focus ring on the search input** (`e6c1918`) — my seamless row took the field's focus indicator away with its border, which I'd flagged as a risk. The row now draws an inset accent ring on `:has(input:focus-visible)`, rounded to the panel's top corners so it follows the surface instead of poking outside it. Inset rather than an outline because the row is full-bleed; on the row rather than the bare `<input>` so the magnifier and clear button stay inside it; `:has(input:focus-visible)` rather than `:focus-within` so tabbing to the clear button doesn't re-ring the whole row. Note the field takes focus when the panel opens, so the ring is visible in the default open state.

**Divider under select-all removed** (`348badc`) — it was separating select-all from rows it belongs with; the checkbox column already aligns them.

**Section titles → plain headings** (`fe8db55`) — proposal, easy to drop. The collision was concrete: with the search row's divider at y=514 and the "Citrus" labeled divider at y=523, two rules sat 9px apart, one full-bleed and one inset. A labeled `Divider` is line–text–line, which is separator language, not group-header language. `DropdownMenu` and `CommandPaletteGroup` already render a panel-list group heading as plain left-aligned supporting-size secondary text with no rules — this puts the two selectors on that existing pattern, so the title now aligns with the option labels under it. It also fixes an a11y wrinkle: the heading moves inside `role="group"` as `aria-hidden` text (the group already carries the title as its accessible name), which stops a stray `role="separator"` from sitting as a direct child of `role="listbox"`.

213 tests pass; new ones pin the select-all row having no rule under it and the section heading being an aria-hidden heading inside the group rather than a separator.

cixzhang:
Reshaped the focus outline (`b44a564`) — screenshots in the description are refreshed.

The field now follows **item** geometry rather than input geometry: same padding + `--radius-element` that Typeahead's items and the selectors' own option rows use. It's no longer a full-bleed strip rounded to the panel's corners; it's a rounded box inset from the panel edge, and the ring is drawn on that box.

Measured against the option row directly beneath it:

| | Selector | MultiSelector |
|---|---|---|
| left inset | 5px = 5px ✓ | 4px = 4px ✓ |
| right inset | 5px = 5px ✓ | 4px = 4px ✓ |
| width | 240 = 240 ✓ | 292 = 292 ✓ |
| height | 32 = 32 ✓ | 32 = 32 ✓ |
| radius | 10px = 10px ✓ | 10px = 10px ✓ |
| content x | 13 = 13 ✓ | 12 = 12 ✓ |

(10px is `--radius-element` as the neutral theme resolves it.)

Diffing the focused and unfocused captures, the changed pixels now sit in a 32px band spanning cols 38–517 of a panel occupying 28–528 — inset on both sides, where before they ran the panel's full width.

One thing I did **not** change: the ring only appears on focus; the field has no resting border. If you want a visible boundary at rest too — closer to how a real input looks parked — that's a one-line addition, but it starts reintroducing the box-in-a-box the PR set out to remove, so I left it as a question rather than assuming.

cixzhang:
Fixed (`2618059`) — screenshots in the description now show the mouse-vs-keyboard pair side by side.

**It was already using `:focus-visible`, and that's exactly why it didn't work.** Per CSS Selectors 4, a pointer-focused element that *supports text entry* matches `:focus-visible` deliberately — so a clicked text field still shows where typing will go. I verified it two ways in Chromium: on this component every route reported `input.matches(':focus-visible') === true`, including both click paths, and a bare `<input>` on a blank page clicked with a mouse matches it too. So it isn't our markup — `:focus-visible` alone can never mean "keyboard only" for a text input.

The fix keeps `:focus-visible` as the CSS condition and adds the one bit it cannot express: which device moved focus. A shared document-level tracker (capture-phase `pointerdown`/`keydown`, two passive listeners for the whole app) records the last interaction, and the ring style is applied only when focus was taken by keyboard. The browser's heuristic still decides everything else, including `focus({focusVisible: true})` — this only narrows it, never widens it.

Verified in real Chromium across 8 routes:

| | ring |
|---|---|
| click the trigger to open | no |
| click into the field text | no |
| click to open, then type a query | no |
| Tab → Enter | **yes** |
| Tab → ArrowDown | **yes** |
| type, Tab onto ✕, Shift+Tab back | **yes** |
| multi-select: click to open | no |
| multi-select: Tab → Enter | **yes** |

One judgment call worth naming: typing after a mouse-open does **not** retroactively raise the ring. The caret already shows where the text is going, and a ring appearing mid-keystroke is its own kind of surprise. Easy to flip if you'd rather any keypress promote it.

jsdom implements neither `:focus-visible` nor the modality heuristic, so the unit tests assert the gate (three new cases) and Chromium verifies the painted ring.

Heads-up on the branch: it had been rebased onto newer `main` at 08:54 by something other than me. I rebased my commit onto that rather than force-pushing over it, reinstalled, and re-ran everything against the new base — 918 tests pass.

ernestt (approved):

Issue #4739, opened 2026-08-05
[Docs] Make `astryx docs <topic>` references in Principles → Rules link to their docs pages


mattandryc:
### Summary
Not sure where this fits — it's a docs UX improvement:

In [Principles → Rules](https://astryx.atmeta.com/docs/principles#rules), several rules end with a
parenthetical pointing at another docs topic:

- "Layout is frame-first… (see `astryx docs layout`)"
- "StyleX or Tailwind for custom styling; both are first-class (see `astryx docs styling`)"
- "Semantic tokens, not hardcoded values (see `astryx docs tokens`)"

These render as inline code. On the docs site they're a dead end — the reader has to find and navigate to the page. 

### Why this matters

The Rules section is the densest, highest-traffic part of Principles, and each of those
parentheticals is the handoff to the page that actually explains the rule. Right now the handoff
doesn't work for the human audience.

Maybe the current formatting isn't accidental: `astryx docs tokens` is a real CLI command,
and rendering it as code is correct for an agent reading the page — that's literally the string to
type. So this isn't "stop using code formatting," it's "serve both audiences."

### Proposed change

Make the parenthetical a link while keeping the code formatting:

> Semantic tokens, not hardcoded values (see [`astryx docs tokens`](/docs/tokens))

The command stays copy-pasteable for agents; humans get a click target. Applies to all three
references in Rules, and to the same pattern anywhere else in the docs if it appears there.

### Open questions

- Is there an existing mapping froe route, or would this behand-authored
  per reference?
- Should this be a small authoring and the URL can't drift apart)rather
  than raw links? Deferring to whoever owns the docs site on whether that's worth it for three
  call sites.

### Scope

Docs-site rendering only. No changunderlying doc content.

mattandryc:
thank you @josephfarina

PR #4929, opened 2026-08-12
fix(docsite): unify underline + hover treatment for inline prose links

mattandryc:
Builds on #4766 and #4444; in response to #4739.

## The problem

#4739 asked for the docs' cross-references to take readers somewhere, and the autolinker from #4766 delivered that for `astryx docs <topic>` commands, with #4444 following for component names. Between them, nearly every reference in docs prose now navigates. That layer works and this PR doesn't touch it.

What's left is presentational. The links accumulated their styling from four different renderers as each surface got built:

- authored links in `/docs` prose — persistent 1px underline that thickens on hover
- component-name chip links (#4444) — same treatment, same renderer
- linked command chips (#4766) — underline revealed on hover
- `[label](href)` links in prop descriptions — core `Markdown`'s default underline

Each is sensible on its own. Side by side on a page like `/docs/principles`, though, a reader can't tell at rest which chips are clickable, and the same component name can render two ways depending on whether the author wrote backticks. WCAG 1.4.1 (Use of Color) is the relevant practice here: a link inside prose should carry a non-color cue, and on this site that cue is the persistent underline most links already have.

## The change

One shared style source instead of four local ones.

| Before | After |
|---|---|
| Screenshot 2026-08-11 at 9 43 05 PM | Screenshot 2026-08-11 at 9 41 59 PM|  

- `ProseLink.tsx` holds the treatment as composable partials — `underline` (1px, 2px on hover, guarded by `@media (hover: hover)`), `color` (text-accent at rest, accent on hover), `focusRing` (raw anchors and buttons only — core `Link` brings its own), and `chipOffset` — plus a `ProseLink` component that composes them onto core `Link` for the consumers that render one.
- All four renderers consume it: `renderInlineMarkdown`'s anchors and the props-table type triggers take the partials directly; the linked chips and prop-description links go through `ProseLink` (registered as the `link` override next to the existing `inlineCode` one, which also keeps router navigation via `useLinkComponent`).
- The chips keep everything #4766 built — the resolver, the unknown-topic guard, the drift test — and just wrap their `Code` chip in `ProseLink`. The chip itself is untouched; the underline belongs to the anchor, so it reads as "this chip is a link" rather than a new kind of chip.

Small things surfaced along the way:

- `--radius-sm` isn't a token the system defines, so the focus rings that referenced it had no rounding; they now use `radiusVars['--radius-inner']`.
- The chip underline offset stays at #4766's 0.2em (clear of descenders in commands like `astryx docs typography`) and now applies to every chip-labeled link; plain-text links keep 0.16em.
- Hover thickness still snaps rather than animating — matching the existing links, where only color transitions.

## Test plan

- [x] 8 new tests pin the treatment on `ProseLink` itself and assert each renderer delegates to it (element-tree assertions; StyleX is mocked since the vitest config doesn't run the compiler, so the tests assert wiring and the build asserts compilation)
- [x] Full docsite suite green (372 tests), typecheck and lint clean
- [x] `pnpm -F @astryxdesign/docsite build` compiles the shared partials across every `/docs` and `/components` page
- [ ] Preview deploy: hover and keyboard-focus pass on `/docs/principles` and `/components/TextInput` in both color modes


kentonquatman (approved):
The design looks good! Thanks @mattandryc!

mattandryc:
> The design looks good! Thanks @mattandryc!

Thanks for the review @kentonquatman! What needs to happen to get this change merged?

mattandryc:
Thanks for the review @kentonquatman! What happens next? (i.e., what's the process for merging?)

mattandryc:
have an approval from design side (Kenton)— need a codeowner review from @cixzhang (maybe?)

cixzhang (approved):
Thanks — this is the fix, merging as is. I merged `main` in for you and verified it in Chrome: chips now carry the rest underline, layout is unchanged, and the type-definition trigger is identical apart from its focus ring finally rounding.

| before | after |
| --- | --- |
| chip with no underline | chip underlined at rest |

Two notes: the description still describes a `ProseLink` component the branch dropped in favour of the partials, so I'll write the squash message from the code. And the red Vercel check is on every fork PR right now, not yours — that's why your preview-deploy box never got ticked.

[Reviewed by Robohands]

PR #4936, opened 2026-08-12
fix(core): consolidate the hover→click guard into useMenuHover (#3121)

cixzhang:
Consolidates the hover-then-click interaction that #4555 fixed locally in `TopNavMegaMenu` into the shared `useMenuHover` hook — because the same bug was live in every other consumer.

Fixes the #3121 shape for `TopNavMenu`, `TopNavHeading`, `SideNavHeading` and `DropdownMenuSubMenu`; `TopNavMegaMenu` now runs on the shared machine instead of its own copy.

## Why

I reproduced the bug rather than assuming it. Against `main`, with `matchMedia` reporting a hover-capable pointer:

| component | after hover | after the click that follows |
| --- | --- | --- |
| `TopNavMenu` | open | **closed** |
| `SideNavHeading` | open | **closed** |
| `TopNavHeading` | open | **closed** |
| `TopNavMegaMenu` (post-#4555) | open | open ✅ |

No unit test caught it because the shared jsdom setup reported `matches: false` for every media query, so `useMediaQuery('(hover: hover)')` made the hook's entire hover path unreachable in tests. That setup now reports `(hover: hover)` as matching, which is what a desktop browser reports.

## What moved into the hook

- **The hover→click guard** — within `clickGuardMs` (500, matching #4555) of a hover-open, a click confirms and pins the menu instead of dismissing it.
- **Pinned vs transient** — `stickyRef` was the inverse of the hook's existing `hoverModeRef`; one concept, one name now. Re-entering the trigger of an open menu no longer un-pins it or re-arms the guard.
- **Native invoker wiring** — `popoverId` makes the hook stamp `popoverTarget` on the trigger and cancel the invoker's default toggle, so an auto popover's light dismiss can't close the menu before the guard sees the click. Documented in the hook's docblock, alongside when a consumer does *not* need it (`popover="manual"`, or a trigger inside the panel).
- **Keyboard activation** — always opens and moves focus in, never toggles an open menu shut. A hover-open leaves focus on the trigger, so that is the one state where a keyboard user can activate the trigger of an open menu; closing there would strand them.

## Focus, and no `requestAnimationFrame`

Focus now moves **synchronously** in the event handler. `useLayer.show()` calls `showPopover()` in the same tick and the layer's children are always mounted, so the items are focusable the moment `show()` returns — the deferred focus was unnecessary. Removed from `useMenuHover` and `DropdownMenuSubMenu`.

The hook also passes `skipAutoFocus: true` on click-open and owns focus itself: `usePopover`'s auto-focus targets the first *tabbable* node, a menu wants its first *item*, and menu items are `tabindex="-1"` under roving tabindex — so both were running and focusing different elements a frame apart.

Closing now returns focus to the trigger instead of dropping it to `<body>` (`usePopover` tracked the trigger element for exactly this and never used it). The nav headings' in-panel heading replica is a close affordance, not the trigger, so it routes through the hook's `close()` rather than borrowing trigger-toggle semantics.

## Found in the browser, not in jsdom

A panel positioned over its own trigger (the nav headings) puts the trigger back under a stationary pointer when it closes, and the browser fires a fresh `mouseenter`. With a zero show delay the menu reopened instantly — Escape and the in-panel close button looked inert. The old one-shot "skip next mouseenter" is now a 300 ms window after *any* close, so that enter is ignored while a deliberate re-hover seconds later still opens the menu.

## Test plan

**Unit** — 21 new hook tests plus per-component guard/focus tests for both nav headings and the dropdown submenu, all a11y-shaped: who holds focus after each way of opening and closing, keyboard activation on an open menu, Escape restore, and that a device with no hover-capable pointer cannot be tricked into opening a menu. `packages/core`: 6192 passing (the 6 failures are pre-existing perf-budget and load flakes — they fail on `main` too and pass in isolation in both trees).

**Real Chromium** (Playwright against local Storybook — jsdom implements neither light dismiss nor invokers):

| | TopNavMenu | TopNavMegaMenu |
| --- | --- | --- |
| hover opens | ✅ | ✅ |
| hover → immediate click stays open | ✅ | ✅ |
| pinned: stays open when pointer leaves | ✅ | ✅ |
| later click closes | ✅ | ✅ |
| hover-open → leave closes | ✅ | ✅ |
| Enter opens, focus moves into panel | ✅ | ✅ |
| Escape closes, focus returns to trigger | ✅ | ✅ |

## Known gaps (not addressed here)

- **`SideNavHeading`'s menu contract is unspecified.** Its popup has no `role="menu"` and the story passes plain links, so `[role="menuitem"]` matches nothing: Enter opens the panel but focus never enters it, and arrow keys do nothing. Pre-existing, and a component-contract decision rather than a hook fix.
- **`useMenuHover` still bundles a `useListFocus` instance** that `TopNavMenu` and `DropdownMenuSubMenu` override with their own (they need typeahead, boundary scoping, roving tabindex). Hover intent and the keyboard model should probably be separate hooks.
- **`requestAnimationFrame`-deferred focus remains** in `usePopover`, `DropdownMenu`, `ContextMenu` and `TabMenu`. `usePopover` is the root one and the highest-risk to change.


imdreamrunner (approved):
This looks good to ship. In `TopNavMegaMenu`, keyboard activation already focused the first panel link, while a pointer click left focus on the trigger. Moving to `useMenuHover` now makes the focus transfer synchronous and applies it consistently to both keyboard and pointer activation, including a click that confirms a hover-opened panel. Hover alone still leaves focus unchanged, and Escape reliably returns focus to the trigger. This consistent behavior looks good to me.

As a non-blocking follow-up, the current ARIA model is:

```html
<button
  aria-haspopup="true"
  aria-expanded="false"
  aria-controls="products-panel"
>
  Products
</button>

<div id="products-panel" popover="auto">
  <div role="group" aria-label="Products">
    <a href="…">Analytics</a>
    <a href="…">Messaging</a>
  </div>
</div>
```

`aria-haspopup="true"` tells assistive technology to expect a `menu`, while the popup intentionally contains a group of ordinary navigation links. For the disclosure-navigation model, this would be more consistent:

```html
<button
  aria-expanded="false"
  aria-controls="products-panel"
>
  Products
</button>

<div id="products-panel" popover="auto">
  <div role="group" aria-label="Products">
    <a href="…">Analytics</a>
    <a href="…">Messaging</a>
  </div>
</div>
```

This ARIA mismatch predates the PR and does not need to block it.

PR #4938, opened 2026-08-12
fix(indicator): ship the a11y type change as breaking, with a codemod (#4937)

### LINKED ISSUE #4937: [audit] Indicator: #4921 removed four members from the public exported `IndicatorProps` type — a change that turns a consumer's build red — and shipped it as a `patch` changeset categorised `[fix]`, with no `astryx upgrade` codemod. P11 requires a `[breaking]` changeset and a codemod for a removed or retyped public prop.
Found by a component audit of **Indicator**, mode O, rubric **v1.3**, at commit `653729439e`.

- **Rule:** `P11` (§3 Public API contract) — see the [Component Audit Rubric](https://github.com/facebook/astryx/wiki/Component-Audit-Rubric).
- **Ledger row:** [Indicator on the component scores page](https://facebook.github.io/astryx/sandbox/pages/component-scores/) — recorded C 82.4 with 1 open BLOCK(s).
- **Finding:** #4921 removed four members from the public exported `IndicatorProps` type — a change that turns a consumer's build red — and shipped it as a `patch` changeset categorised `[fix]`, with no `astryx upgrade` codemod. P11 requires a `[breaking]` changeset and a codemod for a removed or retyped public prop.
- **Evidence:** Measured at 653729439e against the real type: in a probe file compiled by `pnpm -F @astryxdesign/core typecheck`, all four of `props.role`, `props['aria-hidden']`, `props['aria-label']` and `props['aria-labelledby']` on `IndicatorProps<'singleSelection'>` are errors — four `@ts-expect-error` directives, every one consumed, tsc exit 0 (an unused directive is itself TS2578). Source: packages/core/src/Indicator/types.ts:96-102, `extends Omit<BaseProps<HTMLSpanElement>, 'aria-hidden' | 'role' | 'aria-label' | 'aria-labelledby'>`. `IndicatorProps` is exported from packages/core/src/Indicator/index.ts and the `./Indicator` subpath. The changeset .changeset/indicator-decorative-contract.md declares `'@astryxdesign/core': patch` with a `[fix]` first line and itself advertises the break: "passing `role` is a compile error". No transform was added under packages/cli/assets/codemods/transforms. CONTRIBUTING §Version Bumps: `[breaking]` bumps the minor and everything else the patch, precisely so a consumer on a caret range is warned; `pnpm check:changesets` only enforces the category-to-bump coupling and reported "58 changeset(s) valid", so CI cannot catch a wrong category. In-repo precedent from the same author three commits back: .changeset/remove-dropdown-menu-radio-dot-target.md shipped `[breaking]` + minor + the `rename-dropdown-menu-radio-dot-target` codemod for removing a theme target, a strictly quieter break (it fails silently at runtime rather than loudly at compile time). Blast radius, measured and stated honestly: no core host passes any of the four (grepped every file rendering an indicator — zero hits), `data-*` still type-checks because `Omit` preserves BaseProps' `data-${string}` index signature, and a plain BaseProps-shaped component is still assignable to `IndicatorComponent` — so the break lands on external consumers who read those members off the public type or write `role` as a literal attribute.

### Closing protocol

Resolving this requires:

1. Re-run the audit for Indicator after the fix — the whole component, not the diff (the rubric's "Grading a whole component" section).
2. Put the visual results in the PR description — before/after screenshots of the affected states. The rendered-design section is graded from those.
3. Update the ledger with the new score and clear this BLOCK (`<scorecard> | node scripts/score-ledger.mjs --record Indicator --from - --push`).
4. Close this issue linking that PR.

If the recorded score doesn't move, either the fix or the audit is wrong.

cixzhang:
Fixes #4937. Found by the independent re-audit under rubric v1.3, not by me.

## The BLOCK

#4921 `Omit`ted four members from the exported `IndicatorProps` and shipped it as a **`patch`** under `[fix]`. That is a retyped public export, which rubric **P11** names explicitly at BLOCK severity: it needs `[breaking]` → minor **and** a codemod. The changeset even advertised the break in prose ("passing `role` is a compile error"). `check:changesets` only enforces category↔bump, so CI could not see it.

My own PR three commits earlier set the precedent I then ignored: `remove-dropdown-menu-radio-dot-target` shipped `[breaking]` + minor + a codemod for a **quieter** break — silent at runtime, rather than red at compile time.

- Changeset re-issued as `[breaking]` / **minor**.
- New codemod `remove-indicator-a11y-props`, staged in `transforms/next`.

**The codemod reaches both shapes**, which is the part worth reviewing. A direct `<CheckIndicator … />` is easy; the documented themeable form is not:

```jsx
const Mark = useIndicator('check');
return <Mark state="checked" tabIndex={0} aria-label="chosen" />;
```

No indicator name appears in that JSX at all, so a name-matching transform would sail past it. It resolves locals through `useIndicator()`/`getIndicator()` too, pinned by a test.

It **removes** rather than rewrites — there is nowhere to move these props to — and leaves a TODO at every site, because "the owning control already carries the name" is a claim about the call site that a codemod cannot verify. If the indicator was the only thing naming the control, the name has to move *there*, and a human has to decide that.

## Two FIXes from the same audit

**`tabIndex` joins the omission list, and is dropped at runtime.** This is a trap #4921 created: `aria-hidden` became unconditional, so a forwarded `tabIndex` is now a focusable node inside a hidden subtree. Before that diff a consumer could pair `tabIndex` with `aria-hidden="false"` and be *valid*.

Measured in real Chromium with the repo's own axe: **0 → 1** `aria-hidden-focus` violation the moment it is forwarded. My own jsdom probe had reported none — jsdom cannot judge focusability, which is worth remembering about that harness.

**`Indicator.doc.mjs:55`** still told a replacement to render children "when present" — verbatim the wording #4921 replaced in `bestPractices` *because* "present" is the trap. (Not an X21 BLOCK: X21 enumerates `examples[].code`, `usage.bestPractices` and `@example`, and a `props[].description` is none of the three. Worth telling the rubric owner that the identical sentence is BLOCK-able one field away.)

## And a correction to something I claimed

"Passing `role` is a compile error" holds **only for a literal attribute**. A spread — the ordinary host idiom — bypasses excess-property checking for every member, `role` included:

```tsx
const hostile = {role: 'checkbox', tabIndex: 0};
<CheckIndicator state="checked" {...hostile} />   // compiles clean
```

So the type states intent and catches the literal; **ordering and the codemod do the enforcing**. A test pins that, deliberately *without* `@ts-expect-error`, so it starts failing the day TypeScript changes its mind.

## Test plan

**4 new tests red against `origin/main`:**

```
× CheckIndicator (glyph path)    refuses a tab stop inside its hidden subtree
× CheckIndicator (children path) refuses a tab stop inside its hidden subtree
× CheckboxIndicator              refuses a tab stop inside its hidden subtree
× RadioIndicator                 refuses a tab stop inside its hidden subtree
   Tests  4 failed | 40 passed (44)
```

Codemod: 5 cases — direct call site, the `useIndicator` alias, other components left alone, no-op on indicator-free files, idempotent. Assertions match the *attribute* form, not a bare substring, because the TODO comment names the props it removed.

**Suites:** 1134 pass across Indicator + every consumer + the codemods; core `tsc` clean; `typecheck:docs` clean; `CI=true eslint --no-cache` 0/0; `check:changesets` 60 valid; `check:sync` clean.

## Ledger

Recorded at **82.4 / C** (`653729439e`, rubric v1.3) with P11 as the one open BLOCK — wiki `3884786`. This closes it; projected **87.1 / B**. I will re-audit and re-record once this merges.

### DIFF
--- a/.changeset/indicator-decorative-contract.md
+++ b/.changeset/indicator-decorative-contract.md
@@ -1,11 +1,13 @@
 ---
-'@astryxdesign/core': patch
+'@astryxdesign/core': minor
 ---
 
-[fix] Indicator: a caller can no longer un-hide a decorative indicator. `IndicatorProps` now omits `aria-hidden`, `role`, `aria-label` and `aria-labelledby` — passing `role` is a compile error — and each indicator emits its own `aria-hidden` after `{...rest}`, so a forwarded one cannot win. Un-hiding an indicator had it announced next to the control that owns the accessible name, saying the same thing twice.
+[breaking] Indicator: `IndicatorProps` no longer accepts `aria-hidden`, `role`, `aria-label`, `aria-labelledby` or `tabIndex`. An indicator is decorative — the control that renders one owns the role, the accessible name and the focus — so un-hiding one had it announced next to that control, saying the same thing twice, and making one focusable put a tab stop inside an `aria-hidden` subtree.
 
-Nothing is stripped: every other prop, including a forwarded `aria-label`, still reaches the DOM, where it is inert inside an `aria-hidden` subtree. Note that TypeScript exempts hyphenated JSX attributes from excess-property checking, so the type alone cannot reject `aria-*`; the attribute order is what enforces it.
+Passing any of them as a literal JSX attribute is now a compile error. Passing one through a spread still compiles (TypeScript exempts hyphenated JSX attribute names from excess-property checking, and a spread bypasses the check for the rest), so the components also emit their own `aria-hidden` after `{...rest}` and drop a forwarded `tabIndex` — that ordering, not the type, is what actually enforces the contract.
 
-Fixes #4918.
+Nothing else is stripped: `data-*`, `id`, handlers, `dir`, `className`, `style` and `xstyle` all still forward. `astryx upgrade` removes the five props from indicator call sites.
+
+Fixes #4918, #4937.
 
 @cixzhang
--- a/packages/cli/assets/codemods/transforms/next/__tests__/next-codemods.test.mjs
+++ b/packages/cli/assets/codemods/transforms/next/__tests__/next-codemods.test.mjs
@@ -90,3 +90,54 @@ const nested = '.astryx-dropdown-menu-radio .astryx-dropdown-menu-radio-dot';`;
     expect(await apply(TRANSFORM, once)).toBe(once);
   });
 });
+
+describe('remove-indicator-a11y-props', () => {
+  const T = 'remove-indicator-a11y-props';
+
+  it('strips the five props from a direct indicator call site', async () => {
+    const input = `import {CheckIndicator} from '@astryxdesign/core/Indicator';
+const x = <CheckIndicator state="checked" aria-hidden="false" role="checkbox" data-testid="keep" />;`;
+    const out = await apply(T, input);
+    expect(out).not.toMatch(/aria-hidden="/);
+    expect(out).not.toMatch(/\srole="/);
+    // Everything else survives — removal must not become a purge.
+    expect(out).toContain('data-testid="keep"');
+    expect(out).toContain('state="checked"');
+    expect(out).toContain('TODO(astryx upgrade)');
+  });
+
+  it('reaches an indicator resolved through useIndicator', async () => {
+    // The documented way a host renders a themeable indicator: no indicator
+    // name appears in the JSX at all, so a name-matching codemod would miss it.
+    const input = `import {useIndicator} from '@astryxdesign/core/Indicator';
+function Row() {
+  const Mark = useIndicator('check');
+  return <Mark state="checked" tabIndex={0} aria-label="chosen" />;
+}`;
+    const out = await apply(T, input);
+    // Assert on the ATTRIBUTE form: the TODO comment names the props it
+    // removed, so a bare substring check matches its own explanation.
+    expect(out).not.toMatch(/tabIndex=\{/);
+    expect(out).not.toMatch(/aria-label="/);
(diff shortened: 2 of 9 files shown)


cixzhang:
Closing — the premise is wrong, and Cindy caught it.

`IndicatorProps` has never shipped. Measured:

- `@astryxdesign/core@0.3.0` (the published version) was released **2026-08-05** and its exports map has **no `./Indicator` subpath**.
- Indicator landed **2026-08-11** in #4712, whose changeset `themeable-indicators.md` is **still pending** in `.changeset/` — so no release has consumed it.

The whole layer ships for the first time in the next release. Omitting props from a type no consumer can have installed breaks nobody, so all three things this PR did are wrong:

- **`[breaking]` / minor** — would bump core to 0.4.0 and write a changelog entry warning consumers about a removal from an API they never had.
- **The codemod** — would run against consumer code that cannot contain an indicator call site.
- **The framing** — there is no migration, because there is nothing to migrate from.

The original `[feature]` / patch changeset on #4712 is the correct and sufficient record: Indicator arrives complete, with the a11y props already omitted.

**Three fixes in here are real and independent of the changeset question**, so I am re-landing them as a plain patch (see the follow-up PR):

- `tabIndex` omitted and dropped at runtime — a genuine trap, measured in real Chromium going from 0 to 1 axe `aria-hidden-focus` violation once forwarded.
- `Indicator.doc.mjs:55` still said "render children when present" — the wording #4921 replaced elsewhere because "present" is the trap.
- The changeset's "passing `role` is a compile error" overstated the type: true for a literal attribute, false for a spread.

PR #4945, opened 2026-08-12
fix(progressbar): make target mark width/height themeable

freddymeta:
## What

Makes the ProgressBar target **mark**'s `width`/`height` themeable via `defineTheme`, backing them with internal CSS vars (`--_progressbar-mark-width` / `--_progressbar-mark-height`, defaulting to `2px` / `8px`).

```ts
components: {
  'progressbar-mark': { base: { height: '12px' } },
}
```

## Why

The mark's size was a hardcoded StyleX literal (`width: 2, height: 8`). A theme trying to resize it — e.g. a 12px tick — could only reach it with a bare `.astryx-progressbar-mark { height }` rule, which has the **same specificity** as the built-in atomic class. Whether it wins then depends entirely on CSS layer order: if the app's StyleX is unlayered or ordered after `astryx-theme`, the built-in `8px` wins and the override is silently ignored (no error, nothing rendered). Consumers were forced to reach for `!important`.

This mirrors how the mark's **color** is already themeable (via the token pipeline) and how `borderRadius` works on Button/Card/Dialog (the derived-var registry): a theme writes a standard CSS property, the pipeline expands it into an internal var the component reads, so the override wins regardless of layer order — no `!important`, no specificity race.

## How

- `ProgressBar.tsx`: mark reads `width: var(--_progressbar-mark-width, 2px)` / `height: var(--_progressbar-mark-height, 8px)`.
- `ProgressBar.doc.mjs` (en + zh): documents the two vars under `theming.vars` and adds `theming.derived` mapping `width`/`height` → the vars.
- `derivedVarRegistry.ts`: adds the `progressbar-mark` entry (kept in sync with the doc by `derivedVarRegistry.test.ts`).
- Tests: a `generateThemeCSS` assertion proving a `progressbar-mark` size override emits the internal vars.

## Notes

- **Non-breaking** — purely additive. Default size is unchanged (`2px` × `8px`); the DOM and rendered class are unchanged. Existing themes are unaffected.
- No new theming target or public API — reuses the existing `astryx-progressbar-mark` target and the established derived-var mechanism.

## Verification

- Vitest (theme + ProgressBar): 367 passed — including `derivedVarRegistry` ↔ doc sync, `themingTargets`, and the new expansion test.
- `tsc --noEmit` ✓ · eslint ✓ · `check-sync` ✓ · `check:changesets` ✓

### DIFF
--- a/.changeset/progressbar-mark-themeable-size.md
+++ b/.changeset/progressbar-mark-themeable-size.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] ProgressBar: make the target mark's `width`/`height` themeable via `defineTheme` (`components: { 'progressbar-mark': { base: { height: '12px' } } }`). The mark size is now backed by internal vars, so a theme override reliably wins regardless of the app's CSS layer order — previously a bare `.astryx-progressbar-mark { height }` rule only tied the built-in style's specificity and could be ignored.
+
+@freddymeta
--- a/packages/core/src/ProgressBar/ProgressBar.doc.mjs
+++ b/packages/core/src/ProgressBar/ProgressBar.doc.mjs
@@ -85,6 +85,14 @@ export const docs = {
         visualProps: ['variant', 'placement'],
       },
     ],
+    vars: [
+      {name: '--_progressbar-mark-width', description: 'Mark tick width', default: '2px', private: true},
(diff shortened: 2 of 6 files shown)


kentonquatman (approved):
This sounds good! Would you be able to provide a screenshot to verify the change is working as expected before merging?

freddymeta:
Closing — dropping this approach.

PR #4953, opened 2026-08-12
feat(menu): a menu item can report its own result (hasCloseOnSelect)

cixzhang:
## What

A plain menu item can decide not to close the menu: `DropdownMenuItem` takes `hasCloseOnSelect` (default `true`), available in both compound and data mode. Two activation bugs that stood in the way go with it.

## Why

`DropdownMenuItem` closed the menu unconditionally, so an action could never report its result where the user was looking. The concept already exists in the system — `DropdownMenuCheckboxItem` defaults to staying open, `DropdownMenuRadioGroup` takes `hasCloseOnSelect` — and the plain action item was the only one that could not express it. There is a real argument that a menu should always close and confirmation belongs in a toast, but "copy this, and tell me it copied, on the row I clicked" is a pattern people already expect, and the system had no way to write it. Reusing the existing prop name keeps it one concept rather than two.

The boolean, rather than Radix's `onSelect(event)` + `preventDefault`: the decision is a property of the action, not of the event, and the system already names it.

Two things had to be fixed for the prop to mean anything, both pre-existing:

**Activation with no handler did not close the menu.** `handleClick` returned early when `onClick` was absent, so a menu row with no handler quietly did nothing at all — the menu just sat there. Closing is now the item's own behavior, not a side effect of having a callback.

**A row that changed its own label lost keyboard focus.** Data mode derived each row's React key from its label, so the moment a "Copy ID" row became "Copied", the key changed, React unmounted and remounted the row, and focus fell to `<body>` — leaving the menu open but keyboard-dead. Rows are keyed by position now. This also removes the only stated reason `DropdownMenuItemData.label` is narrowed to `string`.

## Verified in Chromium

Pointer and keyboard, against the new `StaysOpenOnSelect` story:

- activate the copy row → menu stays open, the row reads "Copied", **focus stays on that row**, and `ArrowDown` still moves to the next item
- activate a normal row → menu closes and focus returns to the trigger, unchanged
- `Escape` still closes and the row resets

Before this change the same pointer interaction left `document.activeElement` on `<body>` with the menu open and arrow keys dead.

### DIFF
--- a/packages/core/src/DropdownMenu/renderDropdownItems.tsx
+++ b/packages/core/src/DropdownMenu/renderDropdownItems.tsx
@@ -39,23 +39,26 @@ const styles = stylex.create({
   },
 });
 
-function getItemKey(index: number): string {
-  return `item-${index}`;
+function getItemKey(item: DropdownMenuItemData, index: number): string {
+  return `item-${item.id ?? index}`;
 }
 
 function getSectionKey(section: DropdownMenuSection, index: number): string {
-  return `section-${section.title ?? index}`;
+  return `section-${section.id ?? index}`;
 }
 
 /**
  * Renders one leaf row as a `DropdownMenuItem`.
  *
- * Keyed by position rather than by label: an item that reports its own result
- * (a copy row swapping to "Copied") would otherwise change key mid-interaction,
- * remounting the row and dropping keyboard focus.
+ * Keyed by `item.id` when the caller supplies one, else by position. NOT by
+ * label: an item that reports its own result (a copy row swapping to "Copied")
+ * would change key mid-interaction, remounting the row and dropping keyboard
+ * focus. Position is the safe default because a menu's rows are usually fixed;
+ * a menu whose items reorder or filter needs `id` for the same reason.
  *
- * `items` is stripped because it selects the submenu shape rather than being an
- * item prop; every other field of `DropdownMenuItemData` is a
+ * `items` selects the submenu shape rather than being an item prop, and `id` is
+ * identity for React rather than something `DropdownMenuItem` renders, so both
+ * are stripped. Every remaining field of `DropdownMenuItemData` is a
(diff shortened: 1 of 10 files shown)


packages/core/src/DropdownMenu/renderDropdownItems.tsx
```diff
@@ -39,8 +39,8 @@ const styles = stylex.create({
   },
 });
 
-function getItemKey(item: DropdownMenuItemData): string {
-  return `item-${item.label}`;
+function getItemKey(index: number): string {
```
cixzhang: Index seems non-ideal as well. Maybe we need a dedicated key?

cixzhang:
Agreed — index is the other half of the same mistake. Neither the label (mutable content) nor the index (mutable position) identifies a row; only the caller knows the identity, so I've exposed it.

`DropdownMenuItemData` and `DropdownMenuSection` now take an optional `id`, used as the React key, falling back to position when it's absent. `id` rather than `key` because a field named `key` in the data object hits React 19's "props object containing a key prop is being spread into JSX" warning on the `{...itemProps}` spread — and because `TreeListItemData.id` is already the library's name for exactly this ("Unique identifier for the item. Used as React key"). It's stripped before the spread: identity for React, not a DOM attribute the caller is setting. `MoreMenu`, `ContextMenu` and `Breadcrumbs` alias the same type, so they get it too.

Optional rather than required, unlike TreeList's: a fixed menu — which is most of them — is correctly keyed by position, and requiring `id` would break every existing `items` caller for no gain. The docblock says when you need it (the array reorders, filters, or grows) and the prop table says the same.

Test added that discriminates: a menu filtered from the outside, with the focused row surviving at a new index. It fails on position keys (`expected <div…> to be <div…>` — different node, focus lost) and passes with `id`. Also asserting `id` doesn't reach the DOM.

One more of these nearby, not touched here: `TopNavMenu.getMenuItemKey` returns `item.title` — same label-keying defect, different component.

PR #4954, opened 2026-08-12
feat(core)!: make the two DropdownMenu item modes peers

cixzhang:
`DropdownMenu`'s two item modes each lacked what the other had, so a menu that wants dividers **and** a trailing shortcut hint was inexpressible in either. Data mode (`items` — the only mode `MoreMenu` accepts) couldn't carry `endContent` or `description`; compound mode had `endContent` but no divider component, only a raw `Divider` that renders a *different* surface. That pair is what stopped a real conversion from using `MoreMenu` at all.

Both halves are closed, and they're closed by the same component: `DropdownMenuDivider` is what the data path renders for `{type: 'divider'}`, so the two modes can't drift. Data mode's `endContent`/`description` come from widening the `Pick` off `DropdownMenuItemProps`, which keeps `renderLeafItem`'s invariant — every data field is an item prop by construction — true for free.

**Verified in Chromium, not jsdom.** Building the same menu each way produces a menu whose `innerHTML` is identical string-for-string, and cropping the two popups to the same box gives **0 differing pixels across 104,720 compared**. The separator is `role="separator"`, no `aria-hidden`, no `tabindex`, `margin-block: 4px` both ways. Arrow-key traversal from the first row goes `Search → Duplicate → Delete` — one step across the divider, never onto it.

data vs compound mode

Before, the compound workaround measurably differed: a bare `<Divider />` came out missing both `astryx-dropdown-menu-divider` (so a theme rule keyed on the menu divider silently skipped it) and the menu's `margin-block` class. The `endContent`/`description` half was a pure type-surface gap — `renderLeafItem` already spread the fields, so the only thing rejecting them was `tsc`: *"Object literal may only specify known properties, and 'description' does not exist in type 'DropdownMenuOption'"*.

### The two judgement calls

**`label` stays `string` in data mode.** Widening isn't needed to close the `endContent` gap, and the honest reason to keep it narrow outlives the keying argument in the current comment (`getItemKey` derives the React key from the label — and a sibling PR is replacing that with position keying anyway, so I've left `getItemKey` untouched). Data mode's value is that a menu is describable as plain data; admitting arbitrary `ReactNode` there erases the line between the modes and leaves two ways to do one thing. Compound mode already accepts `ReactNode` labels. That's an API-shape decision worth making deliberately, not as a side effect of a parity fix.

**The divider name collides, and the type is the one that moves.** A component named `DropdownMenuDivider` shares a name with the existing exported *type* `DropdownMenuDivider` (`{type: 'divider'}`). A value and a type can share a name in TS — but not through one barrel: `export {X} from './a'` beside `export {type X} from './b'` is `TS2300: Duplicate identifier`, which I confirmed before choosing. So the two names can't coexist in `DropdownMenu/index.ts`; something has to move.

The type moves, because the repo already answers this: `DropdownMenuItemData` carries a `Data` suffix precisely so the bare name can be the component. Applying the same rule gives `DropdownMenuDividerData`, plus `ContextMenuDividerData` and `BreadcrumbMenuDividerData` — the collision reproduces in all three surfaces, since each aliases the type *and* now needs to alias the component. (`DropdownMenuSection` keeps its bare name: the suffix disambiguates against a component, and there is no section component.) Naming the component something else — `DropdownMenuSeparator` — was the alternative, and I rejected it: it would leave one concept spelled three ways across `{type: 'divider'}`, `<DropdownMenuSeparator />`, and `Divider`.

This is `[breaking]`, with a staged codemod. A missed import fails at compile time (the name now resolves to a value used as a type), never silently. The codemod only rewrites unambiguous references — type-only imports, or plain imports never used as a value — so a file already rendering `<DropdownMenuDivider />` is left alone.

### Also found, and what I left alone

- `DropdownMenu.doc.mjs` documented `children` as `(item) => ReactNode`, a render prop that doesn't exist — the real type is compound `ReactNode`. Fixed, since it sits in the surface this PR documents and a phantom prop in doc prose actively steers codegen wrong.
- Its `components: [{name: 'DropdownMenuItem'}]` list omits `DropdownMenuSubMenu` and the selectable items too, so I left it rather than half-fix a stale list.
- Compound stories modelled the raw-`Divider` workaround; they now use the real component.
- **Conflict note for `feat/menu-item-close-on-select`:** that branch also edits the `Pick<DropdownMenuItemProps, …>` line (adding `'hasCloseOnSelect'`) and trims the `label` doc comment just below it. Both are one-line resolutions in whichever order these merge.

### DIFF
--- a/.changeset/menu-item-mode-parity.md
+++ b/.changeset/menu-item-mode-parity.md
@@ -1,0 +1,10 @@
+---
+'@astryxdesign/cli': minor
+'@astryxdesign/core': minor
+---
+
+[breaking] DropdownMenu's two item modes are peers again. Compound mode gains a `DropdownMenuDivider` component (aliased as `ContextMenuDivider` and `BreadcrumbMenuDivider`), which the data path also renders, so `{type: 'divider'}` and `<DropdownMenuDivider />` produce identical DOM, spacing, and theme target. Data mode gains `endContent` and `description`, so an `items` row can carry a shortcut hint or secondary text without dropping to compound mode.
+
+The bare names now belong to those components, so the data-mode option types take the `Data` suffix their sibling `DropdownMenuItemData` already carries: `DropdownMenuDivider` → `DropdownMenuDividerData`, `ContextMenuDivider` → `ContextMenuDividerData`, `BreadcrumbMenuDivider` → `BreadcrumbMenuDividerData`. TypeScript cannot re-export a value and a type under one name from a single barrel, so the rename is what makes the components exportable at all. Run `astryx upgrade --apply` to rewrite the type imports; a missed one fails at compile time rather than silently.
+
+@cixzhang
--- a/.changeset/menu-item-mode-parity.md
+++ b/.changeset/menu-item-mode-parity.md
@@ -3,7 +3,7 @@
 '@astryxdesign/core': minor
 ---
 
-[breaking] DropdownMenu's two item modes are peers again. Compound mode gains a `DropdownMenuDivider` component (aliased as `ContextMenuDivider` and `BreadcrumbMenuDivider`), which the data path also renders, so `{type: 'divider'}` and `<DropdownMenuDivider />` produce identical DOM, spacing, and theme target. Data mode gains `endContent` and `description`, so an `items` row can carry a shortcut hint or secondary text without dropping to compound mode.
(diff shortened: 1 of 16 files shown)


cixzhang:
Rebased onto `main` now that #4951, #4952, #4953 and #4955 have landed. Two conflicts, both in the item type, and one of them was a real decision rather than a textual merge.

**`DropdownMenuItemData`'s `Pick`** — took the union: this branch's `description`/`endContent` plus `hasCloseOnSelect` from #4953.

**`label` widens from `string` to `ReactNode`.** The narrowing's own docblock gave the reason — *"data mode derives each row's React key from the label"* — and #4953 removed it: rows key on `id ?? index` now. Keeping `string` would have left this PR shipping "the modes are peers" with the one asymmetry that no longer had a cause, and its doc line telling people to drop to compound mode for a rich label. Verified nothing else depended on the narrowing: typeahead reads `el.textContent`, and `DropdownMenuSubMenu.label` and `DropdownMenuItem.label` were already `ReactNode`. Test added — a `ReactNode` label renders and the row still has accessible name "Rename". Changeset updated.

The doc-table `items` description was the other conflict; merged both halves.

207 menu-family tests pass (DropdownMenu, SubMenu, Selectable, ContextMenu, MoreMenu, Breadcrumbs), core typecheck clean, 744 CLI codemod tests pass, `lint:strict` clean in the touched files.

PR #4956, opened 2026-08-12
fix(core): expand InputClearButton hit area to 24x24 on touch (WCAG 2.5.8 AA)

rubyycheung:
## Summary

The shared `InputClearButton` renders a **20px** visual glyph with a matching **20px** tap target — below the **WCAG 2.5.8 (AA) minimum of 24×24 CSS px**. This adds an `::after` overlay that expands the tappable region to **24×24 — but only on touch**.

```
position: 'relative',
'--_input-clear-hit-inset': { default: '0px', '@media (pointer: coarse)': '-2px' },
'::after': { content: '""', position: 'absolute', inset: 'var(--_input-clear-hit-inset)' },
```

## Why touch-only (coarse pointer)

WCAG 2.5.8's 24px floor is a **touch** requirement. On a fine pointer (mouse) a 20px target is precise enough, and an *unconditional* overlay could **overlap neighboring controls in dense desktop layouts** — the size-vs-spacing budget (`target + gap ≥ 24`). So the expansion is gated behind `@media (pointer: coarse)`:

- **Fine pointer (desktop):** `--_input-clear-hit-inset: 0px` → hit area == the 20px visual glyph. No expansion, no overlap risk.
- **Coarse pointer (touch):** `--_input-clear-hit-inset: -2px` → hit area grows to 24×24.

The visual glyph is unchanged at every breakpoint. This matches how the rest of the codebase scopes touch adjustments (Selector, TimeInput, Typeahead, DropdownMenu all use `@media (pointer: coarse)`).

## Why exactly 24, not more

24×24 is the AA floor, and stopping there keeps the overlay inside its own space. At `-4px` (28×28) it reaches into the 8px adornment gap — meeting the chevron/status glyph exactly — and on the inline-start side it covers part of the input's own caret area.

> Implementation notes: the conditional lives on a top-level custom property because StyleX only allows plain values inside a pseudo-element (a conditional object directly on `inset` under `::after` fails the core package's type build). The property is private (`--_`) — an internal detail, not a themeable target. `position: relative` is declared on the button itself so the overlay's containing block is colocated with the thing that depends on it, rather than relying on `Button`'s own internal `position: relative`.

## Blast radius

Every consumer of the shared button inherits the touch target: **Typeahead, Tokenizer, FileInput**, and any input using `InputClearButton`. This is the single-source-of-truth fix — inputs that roll their own clear button (TextInput/NumberInput/DateInput/TimeInput) are tracked separately to adopt this shared primitive.

## Verification

`InputClearButton.test.tsx` covers the coarse-pointer path so it can't regress silently — asserted against the CSS StyleX injects at runtime, since jsdom resolves neither media queries nor pseudo-element boxes:

```css
.<hash> { --_input-clear-hit-inset: 0px }                                /* default (fine pointer) */
@media (pointer: coarse) { .<hash> { --_input-clear-hit-inset: -2px } }  /* touch */
.<hash>::after { inset: var(--_input-clear-hit-inset) }
```

The containing block gets a source-level guard in the same file: both declarations of `position: relative` compile to the same atomic class, so the rendered CSS cannot distinguish this one from `Button`'s.

- `packages/core/src/Field/InputClearButton.test.tsx` — 10/10 pass (both new cases verified to fail when the fix is reverted)
- Field, Typeahead, Tokenizer, FileInput suites — 268/268 pass
- `@astryxdesign/core` typecheck clean; rebased onto latest `main`

The var is private (`--_`) but still part of the documented theming surface, so it is named for its component and documented in `Field.doc.mjs` with `private: true`, and listed among the vars with no `derived[]` mapping — `inset` on a `::after` overlay is not a property a theme author writes. This mirrors `--_thumbnail-hit-inset` from the same audit.

A companion sandbox audit page (`/pages/tap-targets`) visualizes this and the remaining sub-24 affordances — separate sandbox-only PR #4903.


cixzhang (approved):

packages/core/src/Field/InputClearButton.tsx
```diff
@@ -18,6 +18,23 @@ const styles = stylex.create({
...
+    // property because StyleX only allows plain values inside a pseudo-element;
+    // the conditional lives on this top-level property instead. Private (--_)
+    // because it is an internal implementation detail, not a themeable target.
+    '--_clear-hit-inset': {
+      default: '0px',
+      '@media (pointer: coarse)': '-4px',
+    },
+    '::after': {
+      content: '""',
+      position: 'absolute',
```
cixzhang, line 63: What is this relative to? Can we make sure this won't accidentally attach to some ancestor?

cixzhang (changes requested):

cixzhang (changes requested):
Thanks @rubyycheung — fixing this in the one shared clear button is exactly the right place, and gating on `pointer: coarse` (with the custom-property workaround for StyleX's pseudo-element rule) is well reasoned.

Two changes:

**1. Declare `position: relative` on the button itself.** Answering my own inline question: the `::after` currently resolves against `Button`'s own `position: relative` (`Button.tsx:69`), so it happens to be correct today — but that is an internal of another component, and if it ever changes the overlay silently reattaches to some ancestor and the hit area lands in the wrong place. Set it here in `styles.button` so the containing block is colocated with the thing that depends on it. It's a no-op at runtime and it makes the coupling local.

**2. 24px, not 28px.** WCAG 2.5.8 AA is 24×24, so `-2px` gets us there. The extra 4px matters: at `-4px` the overlay reaches into the 8px adornment gap — meeting the chevron/status glyph exactly — and on the inline-start side it covers part of the input's own caret area. Please use `-2px` and update the title/description, which both say 28.

Also needs a changeset, and `Field/InputClearButton.test.tsx` already exists — worth a case there so the coarse-pointer path doesn't regress silently.

packages/core/src/Field/InputClearButton.tsx
rubyycheung: Set `position: relative` on `styles.button` in this PR, so the overlay resolves against this button rather than against `Button`'s internal. Runtime no-op today, but the coupling is now local.

rubyycheung:
Thanks @cixzhang — both addressed, plus the changeset and the test, in 7e37daa. Rebased onto latest `main` too.

**1. `position: relative` on the button.** Declared in `styles.button`, with a comment saying why: the containing block now lives next to the overlay that depends on it, instead of resting on `Button`'s internal.

**2. 24, not 28.** Inset is `-2px`. Title and description both updated (the description now also states why it stops at 24 — the adornment gap and the caret area).

**Test.** `InputClearButton.test.tsx` gains two cases:

- the coarse-pointer path, asserted against the CSS StyleX injects at runtime (jsdom resolves neither media queries nor pseudo-element boxes): `--_clear-hit-inset` is `0px` unconditionally, `-2px` under `@media (pointer: coarse)`, nothing wider, and the `::after` overlay is what consumes it.
- a source-level guard for `position: relative`. Worth flagging: this one can't be a CSS assertion — StyleX compiles `Button`'s declaration and this one to the *same* atomic class, so the rendered CSS genuinely cannot tell them apart. Reading the source is the only way to catch the regression you're describing; the comment in the test says so.

Both new cases verified to fail when the fix is reverted. Field/Typeahead/Tokenizer/FileInput suites 268/268, core typecheck clean.

rubyycheung:
Follow-up in 5db58e1: CI's `derivedVarRegistry` scan flagged the hit-inset var as undocumented — a private (`--_`) var is still part of the documented theming surface. Renamed it for its component (`--_input-clear-hit-inset`, matching the `astryx-input-clear-*` targets), documented in `Field.doc.mjs` with `private: true`, and listed among the vars with no `derived[]` mapping, mirroring `--_thumbnail-hit-inset` from the same audit. All checks green now.

packages/core/src/Field/InputClearButton.test.tsx
```diff
@@ -121,6 +157,61 @@ describe('InputClearButton', () => {
...
+
+    // Fine pointer: hit area == the 20px visual glyph, no expansion.
+    expect(decl(/--_input-clear-hit-inset\s*:\s*0px/)).toBe(true);
+    // Coarse pointer: 20px + 2px on each side = 24x24, the AA floor.
+    expect(decl(/--_input-clear-hit-inset\s*:\s*-2px/, 'pointer: coarse')).toBe(
+      true,
+    );
+    // ...and nothing wider than that, which would reach into the adornment
+    // gap and the input's caret area.
+    expect(decl(/--_input-clear-hit-inset\s*:\s*-[3-9]px/)).toBe(false);
```
imdreamrunner: This assertion can never fail — it can be deleted.

`decl` with no `inMedia` argument requires `media == null`, so it only ever looks at unconditional rules. A wider inset can only appear inside the `@media (pointer: coarse)` block, which this call excludes by construction. Verified: with the coarse value set back to `-4px` and the `-2px` assertion above disabled, the suite still passes 10/10.

The `-2px` assertion above already fails on a `-4px` regression (also verified), so removing this line loses no coverage.

If you'd rather keep the guard, it needs the media argument — and a range that doesn't stop at 9:

```js
expect(decl(/--_input-clear-hit-inset\s*:\s*-(?:[3-9]|\d{2,})px/, 'pointer: coarse')).toBe(false);
```

packages/core/src/Field/InputClearButton.tsx
```diff
@@ -25,6 +25,33 @@ const styles = stylex.create({
...
+    // detail, not a themeable target.
+    '--_input-clear-hit-inset': {
+      default: '0px',
+      '@media (pointer: coarse)': '-2px',
+    },
+    '::after': {
+      content: '""',
+      position: 'absolute',
+      inset: 'var(--_input-clear-hit-inset)',
+    },
```
imdreamrunner, line 65: The `::after` is created unconditionally — only its `inset` is gated on `pointer: coarse`. On a fine pointer the overlay still exists at `inset: 0`, exactly covering the button, where it adds no hit area but becomes the topmost hit-test box. Hover stops reaching the descendants, so `.astryx-input-clear-icon:hover` no longer matches.

Measured on the built component (Chromium, with a theme rule `.astryx-input-clear-icon:hover { color: red }`):

| | glyph `:hover` matches | computed color | topmost element at glyph center |
|---|---|---|---|
| this branch | `false` | `rgb(78, 96, 111)` | `BUTTON` |
| overlay removed | `true` | `rgb(255, 0, 0)` | svg `path` |

`astryx-input-clear-icon` is a public theme target, and hover-styling it is the reason #4876 gave for migrating off descendant selectors. No in-repo theme styles it today and the default ghost-button hover is unaffected, so nothing regresses visibly right now — but the surface is broken for anyone who does.

Same trick as the inset, applied to `content`, so the overlay doesn't exist at all unless the pointer is coarse:

```js
'--_input-clear-hit-content': {
  default: 'none',
  '@media (pointer: coarse)': '""',
},
'::after': {
  content: 'var(--_input-clear-hit-content)',
  position: 'absolute',
  inset: 'var(--_input-clear-hit-inset)',
},
```

Verified with that change: hover restored on a fine pointer, touch target still exactly 24×24 (hit test returns the button 1.5px outside the border box, nothing at 3px), core typecheck clean, +0.3KB CSS.

`Thumbnail`'s `--_thumbnail-hit-inset` has the identical shape and the same issue — pre-existing, not this PR's to fix.

packages/core/src/Field/InputClearButton.test.tsx
rubyycheung: Confirmed and fixed. Reproduced your experiment — coarse value back to `-4px`, `-2px` assertion disabled — and the suite still passes 10/10, so the line had no teeth: with no `inMedia` argument `decl` requires `media == null`, and a wider inset can only ever appear inside the coarse block.

Kept the guard rather than deleting it, since "nothing wider than the AA floor" is a claim the comment above it makes and worth holding. Scoped to the coarse block, with your wider range:

```js
expect(
  decl(/--_input-clear-hit-inset\s*:\s*-(?:[3-9]|\d{2,})px/, 'pointer: coarse'),
).toBe(false);
```

Now verified to fail on both `-4px` and `-10px` with the `-2px` assertion neutralized, so it catches a regression on its own rather than riding on the assertion above it.

packages/core/src/Field/InputClearButton.tsx
rubyycheung: Good catch — this is a real regression in the public theme surface and I've taken your fix. Reproduced it in Chromium before changing anything:

| | glyph `:hover` matches | computed color | topmost at glyph center |
|---|---|---|---|
| overlay always generated | `false` | `rgb(78, 96, 111)` | `BUTTON` |
| `content` gated on coarse | `true` | `rgb(255, 0, 0)` | svg `path` |

`content` now runs through `--_input-clear-hit-content` (`none` on a fine pointer, `""` on a coarse one), the same shape as the inset, so the pseudo-element is not generated at all unless the pointer is coarse. The touch target is unchanged: hit-testing still returns the button 1.5px outside the border box and nothing at 3px.

The new var is registered the same way as the inset one — documented in `Field.doc.mjs` with `private: true` and listed among the vars with no `derived[]` mapping — so the `derivedVarRegistry` scan stays green. The changeset is corrected too; it still described the fine-pointer behavior as "the hit area matches the 20px glyph."

Test covers it: `--_input-clear-hit-content` is `none` unconditionally and `""` only under `@media (pointer: coarse)`, and the `::after` is what consumes it. Verified to fail when the gating is reverted to a literal `content: ""`.

Agreed on `Thumbnail`'s `--_thumbnail-hit-inset` — identical shape, same issue. Leaving it out of this PR as you suggest; tracking it with the rest of the tap-target audit.

Field/Typeahead/Tokenizer/FileInput/Thumbnail/theme suites 1030/1030, core typecheck clean.

imdreamrunner (approved):
Verified `0db7887` locally — this resolves the hover regression and the touch fix is intact.

Measured on the built component, hit-testing outward from the button's border box:

| | `::after` | hit area | `.astryx-input-clear-icon:hover` |
|---|---|---|---|
| fine pointer | not generated (`content: none`) | 20×20, unchanged | fires |
| coarse pointer | generated | 24×24 | n/a |

Topmost element at the glyph centre is the `path` again, so pointer events reach the glyph and the theme target works. Click and keyboard unaffected.

Gating `content` rather than adding `pointer-events: none` is the right call — the overlay genuinely must not exist on a fine pointer, and this way there is no pseudo-element to reason about at all.

Also good: the `-(?:[3-9]|\d{2,})px` range catches two-digit insets, which the version I suggested would have missed, and scoping it to the coarse block makes the assertion real — I confirmed the old form stayed green when the inset was widened back to `-4px`.

Field / Typeahead / Tokenizer / FileInput / Selector / derived-var registry: 450/450 pass. eslint clean. Changeset and doc entries updated.

Non-blocking, whenever you next touch the file:

- The comment on line 35 still says WCAG 2.5.8 is "a touch requirement". It is not — 2.5.8 covers mouse, pen and touch alike. The coarse-only gate is still right, but the justification is the **Spacing exception** (undersized targets whose 24px-diameter circles do not intersect), not the criterion being inapplicable to a mouse. Worth correcting since it is a permanent comment in a design system that other components will copy.
- The PR description's blast-radius paragraph says TextInput/NumberInput/DateInput/TimeInput "roll their own clear button" and are tracked separately. All four already route through `InputClearButton`, as do Selector, MultiSelector, DateRangeInput, DateTimeInput and PanelSearchInput — 12 consumers, not 3. Better coverage than advertised.

The red `Flag review signals` check is a cancelled run, not a failure ("Canceling since a higher priority waiting request for review-signal-4956 exists"); its steps all succeeded and it set `review-required` correctly.

PR #4961, opened 2026-08-12
fix(TopNav): fall back to the trigger for anchor positioning without a <nav>

### LINKED ISSUE #4905: [Bug] TopNavMegaMenu dropdown renders at viewport top-left in Properties preview

HelloOjasMutreja:
## Summary

Closes #4905.

`TopNavMegaMenu`'s panel was rendering at the viewport's top-left corner instead of anchored to its trigger, when the component is rendered without an ancestor `<nav>` (e.g. an isolated docs/storybook preview, or any consumer not wrapping it in `TopNav`).

## Cause

The anchor-setting effect only called `popover.triggerRef` when `triggerButtonRef.current?.closest('nav')` found an ancestor:

```tsx
useEffect(() => {
  const nav = triggerButtonRef.current?.closest('nav');
  if (nav) {
    popover.triggerRef(nav);
  }
  return () => {
    popover.triggerRef(null);
  };
}, [popover]);
```

When there's no ancestor `<nav>`, `triggerRef` is never called at all, so no `anchor-name` is ever set anywhere, and the popover's `position-anchor` points at nothing. It falls back to the viewport's top-left corner. Same failure class as the SideNavHeading popover bug fixed in #4789.

## Fix

Fall back to the trigger button itself when no ancestor `<nav>` exists, so positioning degrades gracefully to trigger-anchored instead of failing to anchor at all. The real-usage case (rendered inside `TopNav`, where the full-width panel should align with the nav bar) is unaffected: it still anchors to the `<nav>`, not the trigger.

```tsx
const anchor = triggerButtonRef.current?.closest('nav') ?? triggerButtonRef.current;
if (anchor) {
  popover.triggerRef(anchor);
}
```

## Test notes

Added two regression tests to `TopNavMegaMenu.test.tsx`:
- Anchors to the trigger itself when no ancestor `<nav>` exists.
- Anchors to the ancestor `<nav>` (not the trigger) when one exists.

Verified the first test fails without the fix (`AssertionError: expected [ '' ] to include '--astryx-layer-_r_0_'`) and passes with it restored. Full `TopNavMegaMenu.test.tsx` (39 tests) and the broader `TopNav` suite (105 tests) pass.

## Test plan

- [x] `vitest run` on `TopNavMegaMenu.test.tsx`: 39/39 passing
- [x] `vitest run` on `packages/core/src/TopNav`: 105/105 passing
- [x] Verified new test fails without the fix, passes with it
- [x] `tsc --noEmit`: no new errors
- [x] Changeset added

### DIFF
--- a/.changeset/topnav-mega-menu-anchor-fallback.md
+++ b/.changeset/topnav-mega-menu-anchor-fallback.md
@@ -2,6 +2,6 @@
 '@astryxdesign/core': patch
 ---
 
-[fix] `TopNavMegaMenu`'s panel no longer falls back to the viewport's top-left corner when the component is rendered without an ancestor `<nav>` element (e.g. an isolated docs/storybook preview). It anchors to the nearest ancestor `<nav>` when one exists (unchanged — the panel stays full-width against the nav bar), and now falls back to the trigger button itself when it doesn't, instead of never setting an anchor at all.
+[fix] `TopNavMegaMenu`'s panel no longer falls back to the viewport's top-left corner when the component is rendered without an ancestor `<nav>` element (e.g. an isolated docs/storybook preview). It anchors to the nearest ancestor `<nav>` when one exists (unchanged, the panel stays full-width against the nav bar), and now falls back to the trigger button itself when it doesn't, instead of never setting an anchor at all.
 
 @HelloOjasMutreja
--- a/.changeset/topnav-mega-menu-anchor-fallback.md
+++ b/.changeset/topnav-mega-menu-anchor-fallback.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] `TopNavMegaMenu`'s panel no longer falls back to the viewport's top-left corner when the component is rendered without an ancestor `<nav>` element (e.g. an isolated docs/storybook preview). It anchors to the nearest ancestor `<nav>` when one exists (unchanged — the panel stays full-width against the nav bar), and now falls back to the trigger button itself when it doesn't, instead of never setting an anchor at all.
+
+@HelloOjasMutreja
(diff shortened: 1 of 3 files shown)


HelloOjasMutreja:
Flagging this so it doesn't get read as a duplicate: #4916 (merged, closed #4905) only wraps the docsite Properties preview in a TopNav via TopNavMegaMenu.doc.mjs, so the preview no longer reproduces the symptom. It doesn't touch TopNavMegaMenu.tsx, and the underlying anchor-fallback logic this PR fixes is still present on main unchanged:

```tsx
const nav = triggerButtonRef.current?.closest('nav');
if (nav) {
  popover.triggerRef(nav);
}
```

Any real consumer that renders TopNavMegaMenu outside a `<nav>` ancestor (not just the docs preview) still hits the same anchor-to-nothing failure. This PR fixes that in the component itself rather than the demo wrapper, so I think it's still worth landing on its own merits regardless of #4905's closed state. Happy to retarget the description if a maintainer would rather track it as a fresh issue.

HelloOjasMutreja:
Rebased onto current main, no conflicts. Verified: `TopNavMegaMenu.test.tsx` passes in isolation (46/46); typecheck clean. (A broader batched run across the whole TopNav directory intermittently worker-crashed in this environment, alongside several other simultaneous local builds — reproduces the same way on files this PR does not touch, so it is local resource contention, not a regression.)

PR #4962, opened 2026-08-12
fix(core): route input clear buttons through shared InputClearButton

rubyycheung:
## What

TextInput, NumberInput, DateInput, and TimeInput each rolled their own 16×16 clear (`×`) button. This routes all four through the shared `InputClearButton` — the same primitive Typeahead and Tokenizer already use — so they stay visually consistent and inherit its touch hit-area behavior.

## Why

Part of the WCAG 2.5.8 AA tap-target audit. The self-rolled clear buttons measured 16×16 (below the 24×24 AA minimum). The shared `InputClearButton` centralizes the fix: once its touch hit-area expansion lands (see the core InputClearButton hit-area PR), all four inputs get a compliant 28×28 touch target for free.

## Changes

- Replace each self-rolled `<button>` clear block with `<InputClearButton label onClick />`
- Remove the now-dead local `clearButton` StyleX styles and unused token imports
- DateInput calendar-toggle button is unchanged (it is a separate control)

Net: +18 / −92.

## Testing

- `tsc --noEmit` clean (only pre-existing unrelated `pseudo.json` test errors)
- `eslint` clean on all four files
- Verified visually in the tap-targets sandbox page — all four inputs render their clear × correctly through the shared component

## Note

The visible × shifts from secondary to primary color to match Typeahead/Tokenizer (the shared component uses a ghost Button). This is the intended consolidation outcome.

### DIFF
--- a/packages/core/src/DateInput/DateInput.tsx
+++ b/packages/core/src/DateInput/DateInput.tsx
@@ -35,6 +35,7 @@ import {
 } from '../theme/tokens.stylex';
 import {
   Field,
+  InputClearButton,
   type InputStatus,
   inputWrapperStyles,
   inputStatusBorderStyles,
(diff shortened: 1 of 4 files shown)


cixzhang:
Closing — this landed under [#4876](https://github.com/facebook/astryx/pull/4876) yesterday, which took the same consolidation further: all eight roll-your-own clear buttons (not just these four) now render `InputClearButton`, and the glyph is themed in one place through `astryx-input-clear-icon`. It shipped in `@astryxdesign/core` 0.4.0.

One thing worth carrying forward from the comparison, @rubyycheung: this PR dropped the `date-input-clear-icon` theme target outright, which would have broken any theme keyed on it. #4876 kept the four pre-existing per-component targets rendering behind an `iconClassName` passthrough, marked `deprecatedFor: 'input-clear-icon'`. Removing a themeable surface is a blocking issue in our review criteria, so that's the pattern to reach for whenever a consolidation absorbs an existing target.

The hit-area work this PR depends on ([#4956](https://github.com/facebook/astryx/pull/4956)) is still open and still wanted — `InputClearButton` on main is a 20px-high ghost button with no touch expansion, so the 2.5.8 gap is real and now fixable in exactly one place.

PR #4965, opened 2026-08-12
fix(ListInput): remove invalid aria-required from role=group container

### LINKED ISSUE #4958: [audit] ListInput: aria-required is invalid on role="group" (axe critical)
**Component:** `lab/ListInput` — recorded score **77.2 (C)**, 3 open BLOCKs (Component Audit Rubric v1.3, graded 2026-08-12, commit `46a9e6aeb55`).

**Rule:** [Component Audit Rubric §1 A3/A12](https://github.com/facebook/astryx/wiki/Component-Audit-Rubric#1--accessibility--operable-paths--weight-16) — invalid ARIA on the widget role.

**Finding (BLOCK).** The list container renders `role="group"` with `aria-required={isRequired || undefined}`. `aria-required` is **not an allowed attribute on `role="group"`** (it is valid only on `input`/`select`/`combobox`/`listbox`-type widget roles). axe (`aria-allowed-attr`, impact **critical**) fails on every story that sets `isRequired` (EmptyMailingList, ExpenseAllocations, FamilyMemberDeclaration, ResponsiveItinerary). Because `packages/lab` is not in the `pr-a11y` scope, CI never saw it.

Evidence — `ListInput.tsx:1453`:
```
aria-required={isRequired || undefined}
```
Measured: `axe.run('#storybook-root', {runOnly:['aria-allowed-attr']})` on `lab-listinput--empty-mailing-list` →
`ARIA attribute is not allowed: aria-required="true"` on `<div role="group" … aria-required="true">`.

**Fix.** Remove `aria-required` from the `role="group"` node. The requirement is already surfaced visibly through `Field`'s Required indicator; if it must be exposed programmatically, convey it in the group's accessible name/description (`aria-describedby`) rather than an attribute the role rejects. `aria-disabled` and `aria-busy` on the same node are fine — only `aria-required` is invalid.

HelloOjasMutreja:
## Summary

Closes #4958.

`ListInput`'s group container rendered `aria-required` on a `role="group"` element, which is invalid ARIA. `aria-required` is only allowed on widget roles (`input`/`select`/`combobox`/`listbox`-type roles), not `group`.

## Evidence

`ListInput.tsx:1453` (before fix):
```tsx
aria-required={isRequired || undefined}
```

axe (`aria-allowed-attr`, impact critical) failed on every story that sets `isRequired` (EmptyMailingList, ExpenseAllocations, FamilyMemberDeclaration, ResponsiveItinerary):
```
ARIA attribute is not allowed: aria-required="true"
```
on `<div role="group" ... aria-required="true">`.

Because `packages/lab` is not in the `pr-a11y` CI scope, this never surfaced in CI.

## Fix

Removed `aria-required` from the group node entirely. The requirement is already surfaced visibly through `Field`'s Required indicator, so nothing else needs to change to keep the requirement discoverable.

## Test notes

Added a regression test asserting the group never carries `aria-required`, even when `isRequired` is set. Full `ListInput.test.tsx` suite (20 tests) passes.

## Test plan

- [x] `vitest run` on `ListInput.test.tsx`: 20/20 passing
- [x] New regression test fails without the fix (attribute present), passes with it removed

### DIFF
--- a/packages/lab/src/ListInput/ListInput.test.tsx
+++ b/packages/lab/src/ListInput/ListInput.test.tsx
@@ -217,6 +217,16 @@ describe('ListInput', () => {
     );
   });
 
+  it('never renders aria-required on the group, even when isRequired is set (#4958)', () => {
+    // aria-required is not an allowed attribute on role="group" (axe
+    // aria-allowed-attr, impact critical). The requirement is already
+    // surfaced visibly through Field's Required indicator.
+    renderListInput({isRequired: true});
+
+    const group = screen.getByRole('group');
+    expect(group).not.toHaveAttribute('aria-required');
+  });
+
   it('renders a compact centered EmptyState and keeps Add available', () => {
     const {container} = renderListInput({value: []});
 
--- a/packages/lab/src/ListInput/ListInput.tsx
+++ b/packages/lab/src/ListInput/ListInput.tsx
@@ -1450,7 +1450,6 @@ export function ListInput<T>({
           aria-labelledby={labelID}
           aria-describedby={joinIDs(descriptionID, statusID)}
           aria-disabled={isDisabled || undefined}
-          aria-required={isRequired || undefined}
           aria-busy={isLoading || undefined}
           {...mergeProps(
             themeProps('list-input', {


cixzhang (approved):
Thanks @HelloOjasMutreja — correct and minimal. `aria-required` isn't valid on `role="group"` (it is on `radiogroup`, which is why `RadioList.tsx:390` can keep it), so dropping it is right, and the requirement isn't lost: `FieldLabel` renders the translated "Required" inside the element carrying `labelID` (`FieldLabel.tsx:238-247`, `:260`), which the group names itself from — so it still reaches AT as part of the group's accessible name.

Worth adding the positive half to `ListInput.test.tsx:220` at some point — assert the group's accessible name matches `/Required/`, not just that the attribute is gone, so a refactor can't quietly drop it.

Separately, and not for this PR: with `aria-required` gone there's no state-level way to express a required *collection* — the label mention is announced once on entry and nothing marks the group invalid when it's empty. I'll take that up as its own thing.

PR #4972, opened 2026-08-13
fix(build): import node:fs statically so the Vite plugin's package discovery survives the ESM build

### LINKED ISSUE #4971: `@astryxdesign/build/vite`: package discovery uses `require()` in the ESM-only bundle, so `optimizeDeps.exclude` silently degrades to core-only
## Description

The Vite plugin's config hook discovers installed `@astryxdesign/*` packages so it can exclude them all from Vite pre-bundling:

https://github.com/facebook/astryx/blob/main/packages/build/src/vite.ts#L190-L202

```ts
let xdsPackages: string[] = ['@astryxdesign/core'];
try {
  const fs = require('node:fs');
  const xdsDir = path.resolve(rootDir, 'node_modules/@astryxdesign');
  if (fs.existsSync(xdsDir)) {
    xdsPackages = fs.readdirSync(xdsDir) ...
  }
} catch {
  // Fallback to just @astryxdesign/core if discovery fails
}
```

But the `./vite` export ships **only an ESM bundle**: `build.mjs` compiles `src/vite.ts` with esbuild `format: 'esm'` to `dist/vite.mjs`, and the exports map offers just an `import` condition:

```json
"./vite": {
  "types": "./dist/vite.d.ts",
  "import": "./dist/vite.mjs"
}
```

esbuild lowers the `require('node:fs')` call to its `__require` shim (visible at the top of the shipped `dist/vite.mjs`), which throws at runtime when native `require` is undefined — which is **always**, since the bundle can only ever be loaded as ESM:

```
Error: Dynamic require of "node:fs" is not supported
```

The surrounding `try/catch` swallows the throw, so discovery never runs and `optimizeDeps.exclude` falls back to `['@astryxdesign/core']`.

## Impact

Every consumer of the published plugin. Any other installed Astryx package (`@astryxdesign/icons`, `@astryxdesign/charts`, theme packages, …) stays eligible for Vite pre-bundling — exactly what the code's own comment says must not happen:

> Astryx ships as source that must be compiled by StyleX — pre-bundling strips `stylex.create`/`defineVars` calls and causes runtime errors.

Because the failure is swallowed, there is no warning; the plugin just quietly protects only `core`.

## Reproduction

With any app that has `@astryxdesign/build` plus at least one non-core Astryx package installed:

```js
// repro.mjs — run with plain `node`
import {astryxStylex} from '@astryxdesign/build/vite';

const plugin = astryxStylex({rootDir: process.cwd()})
  .find(p => p.name === 'astryx-config');
console.log(plugin.config().optimizeDeps.exclude);
```

**Expected:** every installed `@astryxdesign/*` package.
**Actual:** `[ '@astryxdesign/core' ]`.

Note the bug is invisible to in-process test runners (Vitest injects a working `require` into transformed modules, and even into dynamically imported file URLs), which is presumably why the existing tests don't catch it. It reproduces only when the compiled bundle runs under real Node ESM semantics.

## Suggested fix

Import `node:fs` statically at the top of `vite.ts` (the file already statically imports `node:path` and `node:url`); esbuild preserves it as a real ESM import. A regression test should compile `vite.ts` with the same esbuild options as `build.mjs` and exercise the discovery in a child `node` process.

## Environment

- `@astryxdesign/build` 0.3.0 (bug present on current `main`)
- Any Node version / any OS — the failure is unconditional under ESM

is-jain:
Fixes #4971

### Problem

`astryxStylex()`'s config plugin enumerated installed `@astryxdesign/*` packages with `require('node:fs')`. The `./vite` export ships only an ESM bundle (`dist/vite.mjs`, built by esbuild with `format: 'esm'`; the exports map has just an `import` condition), so esbuild lowers the call to its `__require` shim — which throws `Dynamic require of "node:fs" is not supported` on every run, because native `require` never exists under ESM. The surrounding `try/catch` swallowed the throw, so `optimizeDeps.exclude` silently degraded to `['@astryxdesign/core']`, leaving every other installed Astryx package eligible for Vite pre-bundling — which strips `stylex.create`/`defineVars` calls and causes runtime errors (the exact failure mode the discovery exists to prevent).

Repro against the shipped bundle, in an app with `core`, `icons`, and `charts` installed:

```js
import {astryxStylex} from '@astryxdesign/build/vite';
const plugin = astryxStylex().find(p => p.name === 'astryx-config');
console.log(plugin.config().optimizeDeps.exclude);
// before: [ '@astryxdesign/core' ]
// after:  [ '@astryxdesign/charts', '@astryxdesign/core', '@astryxdesign/icons' ]
```

### Fix

- `packages/build/src/vite.ts`: replace the `require('node:fs')` with a static `import fs from 'node:fs'` at the top of the file (alongside the existing `node:path` / `node:url` imports). esbuild preserves it as a real ESM import. The `try/catch` stays, now guarding only genuine `readdirSync` failures.
- `packages/build/build.mjs`: add `node:fs` to the esbuild `external` list, matching the existing entries.

### Tests

`packages/build/src/vite.test.ts` gains an `optimizeDeps package discovery` suite:

- discovery returns every installed `@astryxdesign/*` package (dot-prefixed entries filtered out) against a temp fixture tree;
- the core-only fallback still applies when nothing is installed;
- **dist contract test:** compiles `vite.ts` with the same esbuild options as `build.mjs` and runs the discovery in a child `node` process. This is the test that actually fails before the fix — Vitest injects a working `require` into everything it transforms (including dynamically imported file URLs), so no in-process assertion can observe the shim throw. Verified red on the unfixed source, green after.

A changeset (`patch` for `@astryxdesign/build`) is included.

### DIFF
--- a/.changeset/build-vite-esm-fs-require.md
+++ b/.changeset/build-vite-esm-fs-require.md
@@ -1,0 +1,11 @@
+---
+'@astryxdesign/build': patch
+---
+
+[fix] build: import `node:fs` statically so the Vite plugin's package discovery survives the ESM build
+
+`astryxStylex()`'s config plugin discovered installed `@astryxdesign/*` packages with `require('node:fs')`. The `./vite` export ships only an ESM bundle (`dist/vite.mjs`, esbuild `format: 'esm'`), where esbuild lowers `require` to a shim that throws `Dynamic require of "node:fs" is not supported` — always, since native `require` never exists under ESM. The surrounding `try/catch` swallowed the throw, so `optimizeDeps.exclude` silently fell back to `['@astryxdesign/core']` and every other installed Astryx package stayed eligible for Vite pre-bundling, which strips `stylex.create`/`defineVars` calls and causes runtime errors.
+
+The discovery now uses a static `import fs from 'node:fs'`, which esbuild preserves as a real ESM import. A regression test compiles `vite.ts` with the same esbuild options as `build.mjs` and runs the discovery in a child `node` process, since in-process test runners provide a `require` shim that masks the bug.
+
+@is-jain
--- a/packages/build/build.mjs
+++ b/packages/build/build.mjs
@@ -18,12 +18,14 @@ buildSync({
   platform: 'node',
   format: 'esm',
   outfile: 'dist/vite.mjs',
+  // SYNC: keep aligned with the external list in src/vite.test.ts
   external: [
     'vite',
     '@stylexjs/babel-plugin',
     '@stylexjs/unplugin',
     'path',
     'url',
+    'node:fs',
     'node:path',
     'node:url',
   ],
--- a/packages/build/src/vite.test.ts
+++ b/packages/build/src/vite.test.ts
@@ -7,13 +7,22 @@
  *   theme layer name is fixed at `astryx-theme`.
  */
 
-import {describe, it, expect} from 'vitest';
+import {describe, it, expect, beforeAll, afterAll} from 'vitest';
+import {mkdtempSync, mkdirSync, rmSync} from 'node:fs';
+import {tmpdir} from 'node:os';
+import path from 'node:path';
+import {fileURLToPath, pathToFileURL} from 'node:url';
 import {astryxStylex} from './vite';
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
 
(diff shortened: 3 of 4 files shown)


cixzhang (approved):
Thanks for fixing!

PR #4974, opened 2026-08-13
fix(core): stop DateInput/DateTimeInput stealing focus on calendar dismiss

freddymeta:
## Problem

Open `DateInput` or `DateTimeInput`'s calendar by clicking the field, then click another control — and focus jumps back to the date input, fighting your click.

- **DateTimeInput:** click the date input (calendar opens), then click the time input → the time input focuses for an instant, then focus snaps back to the date input.
- **DateInput:** same class of bug — clicking anywhere else after opening bounces focus back to the input.

Both inputs open the calendar with `popover.show({skipAutoFocus: true})`, so focus stays on the text input while the calendar is open. The calendar is a native `popover="auto"`, whose light-dismiss fires its `toggle`/close **synchronously with the pointer event that already moved focus** to whatever you clicked. The popover's `onHide` then unconditionally ran `inputRef.current?.focus()`, overriding that.

`DateRangeInput` never had this: it has no `onHide` refocus and opens from a button trigger, not a click-into-a-text-input.

## Fix

Guard the refocus with a small shared util, `isFocusDetached()`. `onHide` now restores focus to the input **only when the dismiss left focus detached** — on `<body>`/`<html>` or nothing — which is the Escape / click-on-empty-space case the focus trap can't resolve on its own. When the dismiss already landed focus on a real element (the time input, the clear button, another field, anywhere), that's the user's intent and is left alone.

## Behavior

| Interaction | Before | After |
| --- | --- | --- |
| Open via click, then click the time input (DateTimeInput) | focus → date input | focus stays on time input |
| Open via click, then click another field | focus → date input | focus stays on that field |
| Escape while calendar open | focus → date input | focus → date input (unchanged) |
| Open via ArrowDown, Escape | focus → date input | focus → date input (unchanged) |

## Testing

- New `isFocusDetached` unit test (body/documentElement/null → detached; a focused element → not).
- Full `DateInput` + `DateTimeInput` suites green (existing Escape/ArrowDown focus-return tests still pass).
- Verified in a real browser across all four interactions above (click-into-time, click-into-another-field, Escape, ArrowDown-then-Escape).
- `pnpm -F @astryxdesign/core build` green.

Full open/close focus behavior is browser-only (jsdom stubs the native popover), matching the components' existing "tested in Storybook" convention — hence the browser verification plus the unit-tested guard.

### DIFF
--- a/.changeset/datetime-focus-steal.md
+++ b/.changeset/datetime-focus-steal.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] DateInput and DateTimeInput no longer steal focus when the open calendar is dismissed by clicking another control. Clicking the field to open the calendar, then clicking the time input (DateTimeInput) or any other element, kept yanking focus back to the date input because the popover's close handler always refocused it. It now restores focus only when the dismiss left focus detached (Escape, or a click on empty space), so a click that lands focus elsewhere is respected.
+
+@freddymeta
--- a/packages/core/src/DateInput/DateInput.tsx
+++ b/packages/core/src/DateInput/DateInput.tsx
@@ -158,7 +158,7 @@ export type {
   InputStatus as DateInputStatus,
   InputStatusType as DateInputStatusType,
 } from '../Field';
-import {mergeProps, mergeRefs} from '../utils';
+import {mergeProps, mergeRefs, isFocusDetached} from '../utils';
 import type {BaseProps} from '../BaseProps';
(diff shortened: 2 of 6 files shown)


freddymeta:
https://github.com/user-attachments/assets/84e74a27-ed78-4e33-b8b1-3ecdaa057166


https://github.com/user-attachments/assets/485aa274-0779-44be-b96a-b38daa481ada

cixzhang (approved):

PR #4976, opened 2026-08-13
feat(selector): open the menu below by default

### LINKED ISSUE #4227: Selector: open the menu underneath the trigger (reconsider overlay-over-trigger default)
## Summary

`Selector`'s menu currently **overlays the trigger** by default — the selected item is positioned *over* the button (native `<select>`-like behavior), so the open menu covers the button. For the migration use case (and arguably in general) it would read better to have the menu open **underneath the button** instead of on top of it.

## The capability already exists

This is not a missing feature. `Selector` already accepts a `placement` prop (`LayerPlacement = 'above' | 'below' | 'start' | 'end'`):

- **Omitted (today's default):** selector-specific overlay behavior — the selected item is positioned over the trigger and clamped to the viewport (menu covers the button).
- **`placement="below"`:** opts out of the overlay and uses standard layer positioning — the menu opens directly beneath the trigger (the desired "underneath the button" behavior).

So consumers can already get the under-the-button behavior today with `<Selector placement="below" … />`.

## Proposal / question

Should **`placement="below"` (open underneath) become the default** for `Selector`, replacing the current overlay-selected-item-over-trigger default?

- **For:** matches the common dropdown mental model, avoids the menu obscuring the trigger, and is what the migration target expects. Lower surprise for most consumers.
- **Against / to consider:** the overlay behavior is a deliberate native-select affordance (keeps the selected item under the pointer); changing the default is a **visual behavior change** for every existing `Selector` that doesn't set `placement`. Needs the API-conventions view on whether to (a) flip the default, (b) keep the overlay default but document `placement="below"` more prominently, or (c) introduce a semantic default that resolves to below.

This is a **design / default-behavior decision**, not a bug — the mechanism is in place; the question is what the default should be and how to migrate it non-disruptively.

## Acceptance

- A decision on the default placement (flip to `below` vs keep overlay + document), with migration considerations if the default changes.
- If flipped: a changeset noting the visual behavior change, and confirmation that `placement` still lets consumers opt back into the overlay behavior.

AKnassa:
Closes #4227

## What this does

Selector's menu now opens **underneath the button**, with the same small gap every other Astryx menu uses. The old behavior — the menu covering the button with the selected option pinned on top — is still available per instance via a new `hasSelectedItemOverlay` prop.

## Why

Following up [my analysis on the issue](https://github.com/facebook/astryx/issues/4227#issuecomment-5053822853): the overlay default covered the very button that keeps keyboard focus (SC 2.4.11 territory), quietly fell apart once the selected option sat past the listbox fold (the menu just pinned to the top of the screen — including the empty placeholder state every form starts in), and made default Selector the odd one out: DropdownMenu, MultiSelector, ComplexSelector, and Selector-with-search all open below already.

## The three calls left open on the issue — proposed here, easy to veto piecemeal

1. **Does the overlay survive?** Yes, as opt-in — the issue's acceptance asks that consumers can get back to it, and opt-in is the reversible choice.
2. **What's it called?** `hasSelectedItemOverlay` — the `has*` convention, anchored to "selected item" (the code's own term), since bare "overlay" already means the `Overlay` component and its tokens.
3. **The gap?** `--spacing-1`, the standard clearance DropdownMenu, MultiSelector, and ComplexSelector already use. Search-mode Selector gains it too (it used to sit flush).

## What changed

- Menus open below with the standard gap by default; `hasSelectedItemOverlay` restores the overlay per instance (only on below placement — explicit `above`/`start`/`end` still wins, like today).
- `placement` is now documented — it was missing from the docsite props table and `astryx component Selector --dense` entirely (the "(b)" half of the issue, worth having either way).
- New Storybook story **Selected Item Overlay** beside the existing placement stories.
- Changeset `[breaking]` → minor: every default Selector changes visually.

## How to see it

Storybook → Core/Selector: **Default** now opens below with the gap; **Selected Item Overlay** shows the opt-in.

## Test plan

- 3 new tests written red-first against untouched `main`, green after: default = below + `--spacing-1` on both block edges (flip-safe per #4803), explicit placements get the clearance, search mode gets it and never overlays.
- The old overlay geometry tests (viewport clamp, untransformed-layout alignment) now run under `hasSelectedItemOverlay` and still pass — the measurement path is untouched.
- Full suite 9887 passing (the two macOS case-sensitivity failures in `packages/cli` are pre-existing and unrelated); docsite 372 passing; `pnpm lint:strict`, `pnpm test`, `pnpm build` all green.

### DIFF
--- a/.changeset/selector-default-placement-below.md
+++ b/.changeset/selector-default-placement-below.md
@@ -1,0 +1,9 @@
+---
+'@astryxdesign/core': minor
+---
+
+[breaking] Selector menus open below the trigger by default (#4227). The old native-select behavior — the open menu overlaying the trigger with the selected option pinned over it — is now opt-in via `hasSelectedItemOverlay`. `placement` keeps working as before and now documents its `'below'` default; Selector menus also gain the standard `--spacing-1` clearance that DropdownMenu, MultiSelector, and ComplexSelector already use (search mode included, which used to sit flush).
+
+Why the flip: the overlay covered the trigger while it kept DOM focus (WCAG 2.4.11 focus-obscured territory), quietly degenerated once the selected option sat past the listbox fold (the menu pinned to the top of the viewport, including in the no-value placeholder state every form starts in), and made default Selector the odd one out — MultiSelector, ComplexSelector, DropdownMenu, and Selector-with-hasSearch all open below already. To restore the previous look on a given instance, add `hasSelectedItemOverlay`.
+
+@AKnassa
--- a/packages/core/src/Selector/Selector.tsx
+++ b/packages/core/src/Selector/Selector.tsx
@@ -531,14 +531,21 @@ interface SelectorPropsBase<
   searchPlaceholder?: string;
 
   /**
-   * Position placement relative to the trigger.
-   *
-   * Omit to use the selector's default selected-item overlay behavior: the
-   * selected item is positioned over the trigger and clamped to the viewport.
-   * Set a placement to opt into explicit layer positioning (for example,
+   * Position placement relative to the trigger (for example,
    * `placement="above"` for bottom-fixed toolbars).
+   * @default 'below'
    */
   placement?: LayerPlacement;
+
+  /**
+   * Whether to overlay the open menu on the trigger so the selected option
+   * sits directly over it, like a native macOS select. The menu is pulled up
+   * by a measured offset and clamped to the viewport. Only applies with
+   * placement below (the default); an explicit non-below placement or
+   * hasSearch uses standard layer positioning instead.
+   * @default false
+   */
+  hasSelectedItemOverlay?: boolean;
 
   /**
    * Whether the dropdown starts open on mount.
@@ -671,6 +678,7 @@ export function Selector<T extends SelectorOptionType>(
     hasSearch = false,
     searchPlaceholder: searchPlaceholderFromProps,
     placement,
+    hasSelectedItemOverlay = false,
     isDefaultOpen = false,
     'data-testid': testId,
     width,
@@ -829,10 +837,14 @@ export function Selector<T extends SelectorOptionType>(
     [announce, selectableItems],
   );
 
-  // Calculate offset to position selected item over trigger. Explicit
-  // placement opts out of the selector-specific overlay behavior and uses the
-  // standard layer positioning API instead.
-  const shouldOverlaySelectedItem = placement == null && !hasSearch;
+  // Calculate offset to position selected item over trigger. The overlay is
+  // opt-in (hasSelectedItemOverlay) and rides on below placement — an
+  // explicit non-below placement or search mode uses the standard layer
+  // positioning API instead.
+  const shouldOverlaySelectedItem =
+    hasSelectedItemOverlay &&
+    !hasSearch &&
+    (placement == null || placement === 'below');
   const {offset: rawOffset, isPositioned: rawIsPositioned} =
     useSelectedItemOffset({
       isOpen: popover.isOpen && shouldOverlaySelectedItem,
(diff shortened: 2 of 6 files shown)


cixzhang:
On the naming call (#2 in your list): rather than a new boolean, I'd rather we **expand `placement` itself** — add `'overlay'` and `'offset'` alongside the four directions, so the prop that already decides where the menu sits carries this too:

```tsx
placement?: 'above' | 'below' | 'start' | 'end' | 'overlay' | 'offset';
```

- `'overlay'` — the selected option over the trigger, today's default behavior, now nameable instead of implied by omitting the prop.
- `'offset'` — clear of the trigger by the standard menu gap, direction left to the layer. The semantic default option (c) that [#4227](https://github.com/facebook/astryx/issues/4227) asked about, without hard-coding a direction the way `'below'` does.

Two reasons I prefer this to `hasSelectedItemOverlay`. It keeps one prop answering one question — with a boolean, `placement="above" hasSelectedItemOverlay` is a combination we have to define and then explain. And it gives the overlay a name, which we need anyway: [#5004](https://github.com/facebook/astryx/issues/5004) shows the overlay is only sound while the selected option actually covers the trigger (with no selection, or when the viewport clamp slides another option there, a press meant to dismiss the menu commits whatever is painted on top), so the component has to be able to fall back to offset placement on its own — and "fall back to `'offset'`" is a sentence we can write in the docs.

Separately, I'm not sure yet that we want to change the default at all — that's a bigger call than the naming, and I'd like to keep it separate from the API shape. Would you split this into the placement API (values + docs, no behavior change) and the default flip? The first can land immediately.

Also worth knowing: [#5018](https://github.com/facebook/astryx/pull/5018) touches `Selector.tsx` near the trigger, so a small rebase either way.

packages/core/src/Selector/Selector.tsx
```diff
@@ -531,15 +531,22 @@ interface SelectorPropsBase<
...
 
+  /**
+   * Whether to overlay the open menu on the trigger so the selected option
+   * sits directly over it, like a native macOS select. The menu is pulled up
+   * by a measured offset and clamped to the viewport. Only applies with
+   * placement below (the default); an explicit non-below placement or
+   * hasSearch uses standard layer positioning instead.
+   * @default false
+   */
+  hasSelectedItemOverlay?: boolean;
```
cixzhang, line 548: Superseded by the `'overlay'` / `'offset'` placement values. Comes out with the flip.

.changeset/selector-default-placement-below.md
```diff
@@ -0,0 +1,9 @@
+---
+'@astryxdesign/core': minor
+---
+
+[breaking] Selector menus open below the trigger by default (#4227). The old native-select behavior — the open menu overlaying the trigger with the selected option pinned over it — is now opt-in via `hasSelectedItemOverlay`. `placement` keeps working as before and now documents its `'below'` default; Selector menus also gain the standard `--spacing-1` clearance that DropdownMenu, MultiSelector, and ComplexSelector already use (search mode included, which used to sit flush).
```
cixzhang, line 5: Search mode isn't flush any more, [#5003](https://github.com/facebook/astryx/pull/5003) landed that. Needs a reword.

cixzhang (changes requested):
Thanks for this — your analysis on the issue did most of the work, and the placeholder case is the sharpest version of it: before anyone has picked anything, the trigger is completely hidden behind its own menu.

Following up on my comment above, plus one thing that changed since you opened this: the `--spacing-1` clearance landed on main as [#5003](https://github.com/facebook/astryx/pull/5003), so it comes out of your diff and takes most of the conflict with it.

That leaves the placement API half quite small — could you send just that first?

- `'overlay'` and `'offset'` on `LayerPlacement`, behavior unchanged · `useLayer.tsx:72`
- `placement` in the props table, still missing on main · `Selector.doc.mjs:85`

The flip, the changeset and the overlay story then ride on a second PR, where the default is the only thing left to decide.

[Reviewed by Robohands]

cixzhang:
Checking in on Selector placement: the current review asks to split the small placement API/docs change from the default-behavior decision. Would you like to continue, have us carry it, or close this for now?

Issue #4946, opened 2026-08-12
[RFC] Input mask support for free-text inputs (phone, ZIP, SSN, card number)


eyobsamuel:
### Problem Statement

Astryx has no way to constrain free-text entry to a known format while the user types. TextInput accepts anything; the only structured-entry inputs are date/time-shaped (DateInput, DateRangeInput, DateTimeInput, TimeInput, NumberInput).

Many common fields aren't dates or plain numbers but still have a fixed shape: phone numbers, ZIP/postal codes, SSNs, credit card numbers, currency amounts. Today, teams either:
- accept raw text and validate only on blur/submit (user finds out they're wrong too late), or
- hand-roll their own onChange formatting logic per field, per product, with no shared a11y or IME handling.

The gap is real-time input constraint + visual formatting while typing (e.g. auto-inserting the dashes in a phone number as digits are typed, showing placeholder characters for unfilled segments), not just post-hoc validation, which Field's status prop already covers.

### Evidence of Demand

- USWDS ships this as a first-class pattern: Input mask — https://designsystem.digital.gov/components/input-mask/ (phone, SSN/tax ID, currency, ZIP examples with a documented masking approach).

<img width="462" height="198" alt="Image" src="https://github.com/user-attachments/assets/1723c983-8a7b-45b6-9971-f8a54da3d998" />

- MUI documents masked/formatted input as a common integration on top of TextField (via a custom inputComponent, e.g. react-imask) — https://mui.com/material-ui/react-text-field/ — indicating it's expected but not native, which is exactly the gap this RFC is about closing for Astryx (native, themed, accessible, rather than "bring your own third-party masking lib").
- Radix and Chakra don't ship a masked-input primitive either; both leave it to userland libraries, but note it recurs enough that it's a maintained integration point in most design systems and internal Meta forms.
- Frequency: any product with a phone, ZIP, SSN, or payment field hits this. It's one of the most common form-field categories after plain text and numeric.

### Why Existing Components Don't Cover This

Tried composing with TextInput: pass a controlled `value` + custom `onChange` that reformats the string on every keystroke (insert dashes for phone, parens for area code, etc.).

Breaks down because:
- Cursor position management is non-trivial — reformatting the string on each keystroke without manually recalculating and re-setting selection start/end causes the caret to jump to the end, which is broken UX for editing/deleting mid-value.
- No shared placeholder-segment convention (e.g. showing `(___) ___-____` structure) — every team invents their own.
[...]

PR #4977, opened 2026-08-13
feat(lab): add InputMask masked text input

AKnassa:
Refs #4946 (RFC: input mask support)

## What this does

Adds a new lab component, **InputMask**, for fields that have a fixed shape — US phone numbers, ZIP codes, SSNs, and card numbers. As you type, the punctuation appears by itself: type `5551234567` and you see `(555) 123-4567`. A dimmed hint after the cursor shows what's left to type.

## Why

Today every product hand-rolls its own phone/SSN formatting on top of TextInput, and the cursor jumps around when you edit the middle of the value — exactly the pain the RFC describes. This gives Astryx one shared, accessible way to do it, in the lab package first, so the API can be discussed before anything is promised in core.

## What changed

- New InputMask component in `packages/lab`, driven by a `{pattern}` mask prop where `#` marks the digit slots — the RFC's phone/ZIP/SSN/card shapes are all plain patterns (the named-mask presets were dropped in review)
- Your code receives just the digits (`"5551234567"`); the component handles all the formatting
- Works controlled (`value` + `onChange`) or uncontrolled (`defaultValue`), with the controlled prop winning
- The cursor stays where you'd expect while editing mid-value, deleting across dashes, or pasting something messy like `"id: 123-45-6789 (verified)"`
- Edits are read against the display the mask rendered, so a pattern whose literals include a digit (`(+1) ### ### ####`) never reads its own `1` back as typed data, and a pasted formatted value is read against the pattern
- `htmlName` posts the raw digits through a hidden input, so a native form post matches `value`/`onChange`
- Screen readers hear the expected format up front ("Format: (555) 555-5555")
- Storybook pages for common mask shapes, validation, and disabled/read-only states
- 66 tests covering typing, caret math, paste, IME, a11y wiring, form submission, and both controlled and uncontrolled modes

## How to see it

`pnpm -F @astryxdesign/storybook dev` → Lab → InputMask → "Masks". Type into the phone field, edit the middle of the value, backspace across a dash. "Placeholder and hint" has the `(+1)` country-code field.

_Screenshot of typing mid-state to follow below._

## Honest notes for reviewers

<details>
<summary>Scope and caveats</summary>

- The RFC's spec protocol hasn't run; this PR is a working basis for that discussion, in lab (canary-only, per the component lifecycle) — not a core API claim. Refs, not Closes.
- RounakKumarAgarwal asked in the issue to be assigned (same day, unanswered). Happy to defer or collaborate if maintainers prefer.
- Fixed-length masks only — currency, international phone, and auto-switching card-brand masks are out of scope for this pass.
- One RFC correction discovered while building: DateInput is not actually segmented (it is free-text + lenient parsing), so the caret engine here is new code; what is reused: Field wrapper styles, input hooks, ARIA helpers, and the disabledMessage tooltip pattern.
- `value`/`onChange` carry raw digits (the canonical-value precedent from DateInput/TimeInput/NumberInput). Since the review round `value` is optional: `defaultValue` seeds an uncontrolled mode (in-repo precedent: Calendar's `value`/`defaultValue` pair; TextInput/NumberInput are controlled-only, so this is the first Field-family input with one — flagged for the RFC discussion to keep or strike). Native form posts carry the raw digits under `htmlName` via a hidden input, the same route NumberInput takes for a formatted display.
- `changeAction` ships with the rejected-action guard (catch + devError + settle), so a failed async action cannot wedge the field — same defect class as #4871.
- No changeset: `packages/lab` is canary-only and the changeset gate rejects private packages.
- Full suite after the review rounds: 12,848 pass, 2 `.github/scripts` tests that time out under load and pass in isolation. The InputMask suites (66 tests) are green, and the story was exercised in Chrome.

</details>


packages/lab/src/InputMask/maskEngine.ts
```diff
@@ -0,0 +1,115 @@
...
+export interface MaskDefinition {
+  pattern: string;
+  placeholder: string;
+}
+
+export const NAMED_MASKS: Record<NamedMask, MaskDefinition> = {
+  'phone-us': {pattern: '(###) ###-####', placeholder: '_'},
+  'zip-us': {pattern: '#####', placeholder: '_'},
+  ssn: {pattern: '###-##-####', placeholder: '_'},
+  'credit-card': {pattern: '#### #### #### ####', placeholder: '_'},
```
eyobsamuel: Drop the named masks (`phone-us`/`zip-us`/`ssn`/`credit-card`) — they're
US-only, and the RFC's own precedent (MUI) doesn't ship prebuilt masks
either. Each one is just a fixed pattern under the hood, so `{pattern}`
alone covers it with less surface to maintain and no non-US blind spot.

packages/lab/src/InputMask/InputMask.doc.mjs
```diff
@@ -0,0 +1,336 @@
...
+    ],
+  },
+};
+
+/** @type {import('@astryxdesign/cli/authoring').ComponentDoc} */
+export const docsZh = {
+  name: 'InputMask',
+  displayName: 'Input Mask',
+  props: [
+    {
```
eyobsamuel, line 227: Controlled-only forces a useState at every call site — the ControlledMask
wrapper in the stories is evidence of that. Add defaultValue for an
uncontrolled mode, matching TextInput/NumberInput. (Same applies to the
mirrored required value/onChange in InputMask.tsx.)

eyobsamuel, line 227: Controlled-only forces a useState at every call site — the ControlledMask
wrapper in the stories is evidence of that. Add defaultValue for an
uncontrolled mode, matching TextInput/NumberInput. (Same applies to the
mirrored required value/onChange in InputMask.tsx.)

eyobsamuel (commented):
Two shape concerns, as RFC author:

1. Drop the named masks (maskEngine.ts L24-28), keep only `mask={{pattern}}`.
   They're US-specific, each reduces to a fixed pattern anyway, and the RFC
   cited MUI as precedent — which doesn't ship named masks either.
2. Add an uncontrolled mode via defaultValue (InputMask.doc.mjs L33-38,
   value/onChange required: true). Controlled-only adds a useState
   requirement the stories file already has to work around.

Both are about keeping this to what the RFC scoped: a formatting problem,
not four presets + controlled-only wiring. Open to pushback if there's a
discoverability reason for the named masks.

packages/lab/src/InputMask/maskEngine.ts
AKnassa: Done in fc1460c63. `mask` is now `{pattern, placeholder?}` only and `NAMED_MASKS` is gone from the exports. autoComplete no longer derives from the mask; it defaults to `'off'` and passes through as given. I kept the four shapes as pattern fixtures in the tests and stories.

packages/lab/src/InputMask/InputMask.doc.mjs
AKnassa: Added in fc1460c63: `value` is now optional, `defaultValue` seeds an uncontrolled mode, and the controlled prop wins. The stories wrapper is gone. One correction: TextInput and NumberInput are controlled-only (`value` required, `defaultValue` stripped from BaseProps), so the in-repo precedent for the pair is Calendar. That makes this the first Field-family input with an uncontrolled mode, which I flagged in the PR body as a maintainer call for the RFC.

AKnassa: Duplicate thread, please have a look at the answered above

packages/lab/src/InputMask/maskEngine.ts
```diff
@@ -0,0 +1,101 @@
...
+  for (const ch of def.pattern) {
+    if (ch === '#') {
+      count++;
+    }
+  }
+  return count;
+}
+
+export function stripToRaw(def: MaskDefinition, text: string): string {
+  return text.replace(/\D/g, '').slice(0, maxRawLength(def));
```
cixzhang: This can't tell a mask literal from a typed digit. Same assumption at `rawIndexForCaret`, :86.

packages/lab/src/InputMask/InputMask.tsx
```diff
@@ -0,0 +1,554 @@
...
+        currentTarget: el,
+      } as unknown as ChangeEvent<HTMLInputElement>);
+    }
+  };
+
+  const handleClear = useCallback(() => {
+    if (!isControlled) {
+      setInternalRaw('');
+    }
+    onChange?.('', null as unknown as ChangeEvent<HTMLInputElement>);
```
cixzhang: Clear hands `onChange` a null event and skips `changeAction`. Might need the synthetic-target cast `handleCompositionEnd` uses.

packages/lab/src/InputMask/InputMask.doc.mjs
```diff
@@ -0,0 +1,347 @@
...
+      name: 'hasAutoFocus',
+      type: 'boolean',
+      description: 'Autofocuses the input on mount.',
+      default: 'false',
+    },
+    {
+      name: 'htmlName',
+      type: 'string',
+      description:
+        'HTML name attribute for form submission. Note: the submitted value is the formatted display string.',
```
cixzhang: The guidance at :188 says store and submit the raw digits; this submits the formatted string. Which one wins?

cixzhang (changes requested):
Thanks for this — the caret math is the hard part and it holds up on mid-value edits and messy pastes.

One blocker: any pattern with a digit in it rewrites what the person typed. `stripToRaw` pulls every digit out of the *displayed* string, so it can't tell a mask literal from a typed one. It's the "Phone with country code" field in your own story:

```
mask={{pattern: '(+1) ### ### ####'}}
type 5, 5        →  "(+1) 155"            onChange("155")
type 5551234567  →  "(+1) 111 111 1115"   onChange("1111111115")
```

`rawIndexForCaret` counts digits the same way, so the caret goes with it. Two smaller things inline — once the digits round-trip, this is an approve.

Reading the display against the pattern fixes both, but it breaks the messy-paste case your tests cover. How do you want to recover the digits?

[Reviewed by Robohands]

cixzhang:
Checking in on InputMask: literal digits in the pattern still round-trip as typed data and corrupt the value/caret. Would you like to continue, have us carry it, or close this for now?

cixzhang (dismissed):
Reading the display against the pattern was the right answer, and it kept the
messy paste working — the part I wasn't sure about when I asked. All three from
my last round are closed: `(+1) ### ### ####` round-trips `5551234567`, clear
runs `changeAction`, and the raw digits win through a hidden input the way
NumberInput does.

Not merging it here — a new component still wants a human's code review, which
the PR is already waiting on.

If you'd rather talk it through with someone, we're in
[Discord](https://discord.com/invite/XnsUcFykEP).

[Reviewed by Robohands]

packages/lab/src/InputMask/maskEngine.ts
AKnassa: Fixed in 74c1f21b9. Edits are now diffed against the display the mask rendered: kept text is read by slot position, inserted text against the pattern when it carries the literals in place, otherwise as loose digits. Typing 5, 5 gives `(+1) 55` and `onChange('55')`, pasting `(+1) 555 123 4567` gives `5551234567`, and the messy paste case still passes.

packages/lab/src/InputMask/InputMask.tsx
AKnassa: Fixed in 74c1f21b9. Clear goes through the same commit path as typing now, so changeAction runs and the callbacks get the input as the event target, the cast composition end uses.

packages/lab/src/InputMask/InputMask.doc.mjs
AKnassa: Raw digits win, matching NumberInput's formatValue path: a hidden input posts them under htmlName and the visible input carries no name. The htmlName and isReadOnly docs say so now (74c1f21b9).

AKnassa:
Continuing. The literal-digit fix is in 74c1f21b9 with the clear and htmlName threads, and main is merged in.

PR #4978, opened 2026-08-13
fix(core): make CodeBlock header and title theme targets actually apply

freddymeta:
## Problem

The `codeblock-header` and `codeblock-title` theme targets (added in #4943) render their classes, but their dimension/type properties never take effect. The header padding (`headerRow.paddingInline`, `headerWithDivider`/`headerCompact.paddingBlock`) and the title's `fontSize`/`lineHeight` are hardcoded StyleX declarations, so `defineTheme` emits a theme's values as raw `padding` / `font-size` / `line-height` declarations that compete head-on with the component's own StyleX atomics.

Which declaration wins is decided by `@layer` order — and layer order depends on how the **consumer** compiles, which the design system doesn't control. So these targets looked "themeable except they don't apply": color-type properties (a `var()` read with no rival) worked, but the hardcoded dimensions silently lost the cascade. This is the same class of bug #4970 fixed for the ProgressBar mark's `width`/`height`.

## Change

Route the four properties through internal `--_codeblock-*` derived vars with `replaces: true`, exactly as ProgressBar (`--_progressbar-mark-*`) and TextArea (`--_textarea-inline-padding`) do:

- `headerRow.paddingInline` → `var(--_codeblock-header-padding-inline, <--spacing-4>)`
- `headerWithDivider`/`headerCompact.paddingBlock` → `var(--_codeblock-header-padding-block, <--spacing-2>)`
- `headerTitle.fontSize` → `var(--_codeblock-title-font-size, <--text-supporting-size>)`
- `headerTitle.lineHeight` → `var(--_codeblock-title-line-height, <--text-supporting-leading>)`

`codeblock-header` and `codeblock-title` are registered in `derivedVarRegistry.ts` with `replaces: true`, and documented in `CodeBlock.doc.mjs`'s `theming.vars` / `theming.derived`. A theme's `codeblock-header` / `codeblock-title` dimensions now compile to those vars (no rival declaration), so they land whatever the consumer's cascade looks like.

**Theme authoring and rendered defaults are unchanged** — `codeblock-header: {base: {paddingInline…}}` / `codeblock-title: {base: {fontSize…}}` is still what an author writes, and an unthemed block keeps its `--spacing-4` / `--spacing-2` / supporting-size metrics.

## Test infra

CodeBlock is the first component with derived vars on **two** sub-element targets (header and title), so `derivedVarRegistry.test.ts`'s dir→key mapping now accepts an array of registry keys and validates the doc's flat `derived[]` against their concatenation.

## Testing

- New `CodeBlock` test: a theme setting header padding + title type compiles to the `--_codeblock-*` vars and **not** the competing raw `padding-inline` / `font-size` declarations on those targets.
- `derivedVarRegistry` ↔ doc consistency, `themingTargets`, and `generateThemeRules` guards all green (349 tests across the four theme/CodeBlock suites).
- `pnpm -F @astryxdesign/core build` green — the built `astryx.css` contains the four `--_codeblock-*` reads. Core typecheck + typecheck:docs green; strict/CI lint clean for the changed files.

### DIFF
--- a/.changeset/codeblock-header-title-derived-vars.md
+++ b/.changeset/codeblock-header-title-derived-vars.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] CodeBlock: the `codeblock-header` and `codeblock-title` theme targets now actually apply their dimension/type properties. Header padding (`paddingInline`/`paddingBlock`) and title `fontSize`/`lineHeight` are read through internal `--_codeblock-*` derived vars (`replaces: true`), so a theme sets those vars instead of emitting raw declarations that competed with the component's own StyleX atomics and lost the cascade depending on how the consumer compiled. Theme authoring is unchanged — `codeblock-header: {base: {padding…}}` / `codeblock-title: {base: {fontSize…}}` — and the rendered defaults are unchanged.
+
+@freddymeta
--- a/packages/core/src/CodeBlock/CodeBlock.doc.mjs
+++ b/packages/core/src/CodeBlock/CodeBlock.doc.mjs
@@ -146,6 +146,18 @@ export const docs = {
       {className: 'astryx-codeblock-title', visualProps: ['size', 'language']},
       {className: 'astryx-codeblock-copy-button'},
     ],
+    vars: [
+      {name: '--_codeblock-header-padding-inline', description: 'Header row inline padding', default: 'var(--spacing-4)', private: true},
+      {name: '--_codeblock-header-padding-block', description: 'Header row block padding', default: 'var(--spacing-2)', private: true},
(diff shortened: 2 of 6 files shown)


freddymeta:
Closing as redundant. Verified the mechanism: the `codeblock-header`/`codeblock-title` targets from #4943 already apply in the standard layered build — the theme rules land in `@layer astryx-theme`, which outranks the component atomics in `@layer astryx-base` (order: reset < astryx-base < astryx-theme), so equal-specificity theme rules win cleanly. The derived-var routing here only matters for a source build that compiles StyleX without `useCSSLayers`; #4943 covers the layered consumers, so this isn't needed on its own.

PR #4979, opened 2026-08-13
feat(core): add Calendar month/year caption pickers

### LINKED ISSUE #4941: [RFC] Calendar: add a month/year picker for fast navigation to distant dates
## Problem Statement

Calendar only supports moving between months one step at a time, via the prev/next nav buttons in its header. There is no way to jump directly to a specific month or year.

Reaching a far-past or far-future date — a birth date, a multi-year-old record, a long-lead scheduling date — means clicking prev/next dozens or hundreds of times. The only alternative today is typing directly into a paired DateInput/DateRangeInput, which sidesteps the calendar's own guarantees: free-text date entry is easy to mistype, format-ambiguous, and gives no validation feedback until blur/submit.

This is a UX gap, not a missing prop: any flow that asks someone to locate a specific, known, non-recent date currently forces a choice between tedious clicking or an error-prone text fallback. Neither is what the calendar UI is supposed to be for.

## Evidence of Demand

External systems that ship this pattern:

- **react-day-picker** — `captionLayout="dropdown"` replaces the static caption with month/year `<select>`s.
- **shadcn/ui** — Calendar component ships a "dropdown" caption variant using the same pattern.
- **MUI X Date Pickers** — the `views` prop supports a dedicated year-selection view, reachable by clicking the header label.
- **Ant Design** — DatePicker's panel header is clickable and escalates to a year/decade picker view.

This isn't a niche pattern — it's the default or near-default behavior in every major React date-picker library, which suggests the one-step-at-a-time-only header is the outlier, not the norm.

## Why Existing Components Don't Cover This

Calendar's header is not currently composable for this — the month/year label is a fixed, non-interactive element with no exposed slot or prop to swap in an interactive control. There's no caption render-prop, no way to replace just the header while keeping the day-grid, and no documented escape hatch.

The only composition attempt available today is building a fully custom header wrapper around Calendar, which means:
- Re-implementing month-label formatting
- Re-wiring `onNavigate`
- Manually matching Astryx's spacing/typography tokens to avoid visual drift

At which point you've lost the theming and behavioral consistency Calendar is supposed to guarantee, for a problem that isn't product-specific.

The remaining fallback — swapping to a raw DateInput — isn't a composition of Calendar at all; it's an admission that Calendar doesn't support the use case, and it reintroduces the validation/format problems described above.

## Rough Approaches Considered

*Not proposed solutions — included to show the design space has been considered. Final shape should go through API Arbitration if this RFC is accepted.*

### Approach A — Dropdown-based caption
Replace the static label with two dropdowns (month, year) so users can jump directly.

```jsx
<Calendar captionLayout="dropdown" fromYear={1900} toYear={2035} />
```

Simple, keyboard-accessible for free (native/Selector-style dropdowns already handle this), but two long `<select>` lists can be awkward on touch/mobile.

Image

### Approach B — Secondary grid view
Clicking the caption label opens a month-grid (and year stepper/list) to pick from, similar to shadcn's calendar block 21.

```jsx
<Calendar captionLayout="grid" fromYear={1900} toYear={2035} />
```

Nicer on touch/mobile, avoids long scrollable selects, but is a bigger surface to build and test (new keyboard/focus pattern, an extra transient view state).

Image

Both are compatible with keeping `captionLayout="label"` (today's behavior) as the default, so neither implies a breaking change.

## Accessibility Considerations

- Maps to the existing listbox pattern already established elsewhere in Astryx (e.g. Listbox conventions) 
- Approach A via native/Selector dropdowns, Approach B via a grid using the same roving-focus convention, rather than inventing a new one-off interaction
- Month and year controls need accessible names distinct from the day-grid's own labeling (e.g. `aria-label="Month"` / `aria-label="Year"`), so screen reader users don't confuse header navigation with day selection
- Keyboard users must be able to reach and operate the month/year control via Tab, independent of the day-grid's roving arrow-key focus — the two shouldn't compete for the same keys

**Benefits:** screen reader and keyboard-only users disproportionately bear the cost of the current one-step-at-a-time header, since "click next 200 times" isn't a viable strategy for them the way a sighted mouse user might tolerate it briefly. This is as much an accessibility fix as a convenience feature.

## Performance Considerations

Not applicable — this changes the header/caption only, not the day-grid's item count or rendering strategy. No new list/table/grid of items is introduced.

## Pre-submission Checklist

- [x] I have read the [Contributing guide](https://github.com/facebook/astryx/wiki/Contributing)
- [x] I have read the [API Conventions](https://github.com/facebook/astryx/wiki/API-Conventions)
- [x] I have checked that existing Astryx components cannot compose to solve this
- [x] This is a general-purpose UI pattern (not specific to one product)

AKnassa:
## What this does
Adds an optional month and year picker to the calendar header. Turn on `hasMonthYearPickers` and the caption becomes two small dropdowns, so you can jump straight to June 1990 instead of clicking the arrow button dozens of times.

## Why
Picking a distant date (a birth date, an old record) takes one click per month today. That is slow for everyone and much worse for keyboard and screen reader users. This pattern is standard in every major date picker library.

Closes #4941

## What changed
- New `hasMonthYearPickers` option on Calendar, off by default. Nothing changes unless you turn it on.
- The year list follows the calendar's existing `min` and `max` limits (1900 to 2099 when unlimited). Months and years outside the limits are disabled rather than hidden.
- Jumps announce the new month to screen readers, the same way the arrow buttons do.
- Themes can style the new dropdowns through a new `calendar-picker` target.
- With two months side by side, the caption stays as it was, since it names both months.
- English docs, Chinese docs, dense docs, two Storybook stories, and a changeset are included.

## How to see it
Run Storybook (`pnpm dev`) and open Core/Calendar, stories "Month Year Pickers" and "Month Year Pickers With Min Max".

## Notes for maintainers
- This is an implementation offer for the RFC while it awaits a design brief. The API deliberately avoids `captionLayout`, `fromYear`, and `toYear`: the year range reuses the existing `min`/`max` props instead of adding a second bounds vocabulary, per the API conventions. Glad to rework the shape if arbitration lands somewhere else.
- RounakKumarAgarwal asked in the issue thread to take this on. If assigning it to them is preferred, treat this branch as a reference and I am happy to close.
- Two small quirks are inherited from Selector and left unchanged here, both noted in the commit message: type-ahead on the closed trigger announces twice, and opening with an arrow key can highlight a disabled option.
- Tests were written first (20 new), plus mutation testing (7 of 7 mutants caught). Full suite is green apart from the two known macOS-only CLI failures.

### DIFF
--- a/.changeset/calendar-month-year-pickers.md
+++ b/.changeset/calendar-month-year-pickers.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] Calendar can now jump straight to a distant month. The new `hasMonthYearPickers` prop replaces the static month/year caption with compact Month and Year pickers, so reaching a birth date or other far-off date no longer means paging one month at a time. The year list derives from the existing `min`/`max` bounds (1900–2099 when unbounded), always widened so the visible year stays labeled, and months and years outside the bounds are disabled rather than hidden. Off by default; a two-month view keeps the static caption, which labels both visible months. The pickers are themeable via the new `calendar-picker` target (`data-picker="month" | "year"`), and screen-reader month-change announcements cover picker jumps too.
+
+@AKnassa
--- a/apps/storybook/stories/Calendar.stories.tsx
+++ b/apps/storybook/stories/Calendar.stories.tsx
@@ -183,6 +183,54 @@ export const MondayStart: Story = {
   },
 };
 
+export const MonthYearPickers: Story = {
+  render: () => {
+    const [value, setValue] = useState<ISODateString>('1990-06-15');
+    return (
+      <Calendar
+        mode="single"
+        hasMonthYearPickers
+        value={value}
+        onChange={val => setValue(val)}
+        focusDate="1990-06-01"
+      />
+    );
+  },
+  parameters: {
+    docs: {
+      description: {
+        story:
+          'With `hasMonthYearPickers` the static caption is replaced by compact Month and Year ghost-Selector dropdowns, so a distant month — a 1990 birth date here — is two picks away instead of hundreds of arrow presses. Unbounded, the year list spans 1900–2099, widened to include the visible year. Ignored when `numberOfMonths` is 2, where the caption labels both visible months (#4941).',
+      },
+    },
+  },
+};
+
+export const MonthYearPickersWithMinMax: Story = {
+  render: () => {
+    const [value, setValue] = useState<ISODateString | undefined>(undefined);
+    return (
+      <Calendar
+        mode="single"
+        hasMonthYearPickers
+        min="2024-06-10"
+        max="2027-03-20"
+        value={value}
+        onChange={val => setValue(val)}
+        focusDate="2026-01-01"
+      />
+    );
+  },
+  parameters: {
+    docs: {
+      description: {
+        story:
+          'The pickers derive their options from `min`/`max`: the Year dropdown offers only 2024–2027, and months wholly outside the bounds are disabled — January–May in 2024 and April–December in 2027.',
+      },
+    },
+  },
+};
+
 export const RTL: Story = {
   render: () => {
     const [value, setValue] = useState<ISODateString | undefined>(undefined);
--- a/packages/core/locales/en.json
+++ b/packages/core/locales/en.json
@@ -295,6 +295,14 @@
     "defaultMessage": "Next month",
     "description": "Screen-reader-only label on the right-arrow button in a Calendar's month header (navigates one month forward). Pairs with `calendar.previousMonth`."
   },
+  "@astryx.calendar.month": {
+    "defaultMessage": "Month",
+    "description": "Screen-reader-only label on the month picker in a Calendar's header when `hasMonthYearPickers` is set. Names the control that jumps the grid straight to a month of the visible year. Pairs with `calendar.year`."
+  },
+  "@astryx.calendar.year": {
+    "defaultMessage": "Year",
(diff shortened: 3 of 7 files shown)


kentonquatman (changes requested):
The sandbox preview isn't working right now for the design review. Would you be able to update or add screenshots to show the change?

AKnassa:
The hosted previews for this PR 404 now, though `build-sandbox` and `build-storybook` both passed on this commit, so I think the Pages artifact was cleaned up rather than failing to build. I captured these from a local Storybook build of the branch at 1ccbb303a.

**Before and after**

01-before-after


Left is the caption today, right is `hasMonthYearPickers` on the same component, sitting at June 1990.

**Month picker**
02-month-picker-open



**Year picker, no bounds set**
03-year-picker-open




**Year picker with `min="2024-06-10"` and `max="2027-03-20"`**
04-minmax-year-list



The list offers 2024 to 2027 only, derived from the existing `min`/`max` rather than a second bounds vocabulary.

**After picking 2024**
05-minmax-months-disabled



The jump clamps to June, the first month inside `min`. January to May stay visible but disabled, and the prev arrow is disabled at the bound.

**Dark mode**

06-dark-month-open

I left the two month view out because it is unchanged: the pickers are ignored when `numberOfMonths` is 2, where the caption already names both visible months.

Happy to capture any other state that would help.

rubyycheung:
@imdreamrunner, now that the mobile date-picker work has landed (#5243 and #5582), could you help us resolve the direction for this PR? Should these month/year caption pickers remain a desktop Calendar option, be shared with the mobile picker, or be superseded by the newer mobile navigation pattern? I would like to avoid landing a second navigation model that conflicts with the mobile experience.

PR #4998, opened 2026-08-13
fix(core): anchor popover triggerRef in collapsed SideNavHeading

### LINKED ISSUE #4990: [Bug] SideNavHeading menu popover is unanchored when SideNav is collapsed
## Summary

When `SideNavHeading` is rendered with both `menu` and `SideNav`'s collapsed state, its popover opens at the viewport's left edge rather than adjacent to the icon trigger.

## Environment

- `@astryxdesign/core` 0.3.0
- React / Next.js
- Chromium

## Minimal reproduction

```tsx
<SideNav collapsible={{defaultIsCollapsed: true}}>
  <SideNavHeading
    heading="Alex Morgan"
    subheading="alex.morgan@acme.com"
    icon={<Avatar name="Alex Morgan" />}
    menu={
      <NavHeadingMenu>
        <NavHeadingMenuItem label="Profile" />
      </NavHeadingMenu>
    }
  />
</SideNav>
```

## Actual behavior

In the collapsed branch of `packages/core/src/SideNav/SideNavHeading.tsx`, `collapsedSetRef` is passed to the button, but not merged with `popover.triggerRef` or `setTriggerEl`. As a result the popover has no anchor and appears at the left edge of the viewport (in our case, `x = 0`).

## Expected behavior

The popover should be anchored to the collapsed icon button and open next to it.

## Suggested fix

Merge the collapsed trigger ref with the popover and hover-menu refs, e.g.:

```tsx
  collapsedItemRef,
  setTriggerEl,
  ref,
  popover.triggerRef,
);
```

This should preserve the existing tooltip anchor while providing the popover a concrete trigger element. We have a local workaround using `Popover`'s render-prop trigger, which verifies the menu opens beside the avatar.

Geervan:
Fixes #4990

When `SideNavHeading` is rendered with both `menu` and `SideNav`'s collapsed state (`isCollapsed = true`), its popover opened at the viewport's top-left edge (`x = 0`) rather than adjacent to the icon trigger button.

### Root Cause
In the collapsed branch of `packages/core/src/SideNav/SideNavHeading.tsx`, `collapsedSetRef` was assigned to the icon trigger button, but did not merge `popover.triggerRef` or `setTriggerEl`. As a result, `popover.triggerRef.current` remained `null`, leaving Floating UI without a trigger anchor DOM element for popover positioning calculations.

### Solution
Merged `setTriggerEl` and `popover.triggerRef` into `collapsedSetRef`:
```tsx
  collapsedItemRef,
  setTriggerEl,
  ref,
);

### DIFF
--- a/packages/core/src/SideNav/SideNav.test.tsx
+++ b/packages/core/src/SideNav/SideNav.test.tsx
@@ -476,6 +476,27 @@ describe('SideNavHeading collapsed', () => {
     );
     expect(screen.getByTestId('nav-header')).toBeInTheDocument();
   });
+
+  it('anchors popover triggerRef to icon button when collapsed with menu', () => {
+    render(
+      <CollapsedWrapper>
+        <SideNavHeading
+          heading="Alex Morgan"
+          subheading="alex.morgan@acme.com"
+          icon={<span data-testid="avatar">AM</span>}
+          menu={
+            <div>
+              <span data-testid="menu-item">Profile</span>
+            </div>
+          }
+        />
+      </CollapsedWrapper>,
+    );
+    const button = screen.getByRole('button', {name: 'Alex Morgan'});
+    expect(button).toBeInTheDocument();
+    expect(button).toHaveAttribute('aria-haspopup', 'dialog');
+    expect(button).toHaveAttribute('aria-expanded', 'false');
+  });
 });
 
 // =============================================================================
--- a/packages/core/src/SideNav/SideNavHeading.tsx
+++ b/packages/core/src/SideNav/SideNavHeading.tsx
@@ -382,7 +382,12 @@ export function SideNavHeading({
   if (isCollapsed && icon) {
     const collapsedIcon = <span {...stylex.props(styles.icon)}>{icon}</span>;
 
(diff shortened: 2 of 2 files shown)


cixzhang:
Thanks @Geervan — closing, but only because you were beaten to it by two days.

[#4850](https://github.com/facebook/astryx/pull/4850) landed this exact fix on Aug 11: `collapsedSetRef` on main already merges `menu ? popover.triggerRef : undefined`, with a regression test asserting the popover's own `position-anchor` id appears in the trigger's `anchor-name` list. It shipped in `@astryxdesign/core` 0.4.0 earlier today, which is why [#4990](https://github.com/facebook/astryx/issues/4990) still reproduced on 0.3.0.

On the `setTriggerEl` you added beyond that fix: it writes `triggerElRef` inside `useMenuHover`, and that ref is never read anywhere in the hook or the repo — it's an inert sink today, so adding it to the collapsed path changes no behavior. Not wrong, just not load-bearing. (That dead ref is arguably worth removing from `useMenuHover` on its own; a separate PR if you're interested.)

Geervan:
So like submit an issue regarding this and then make the pr fix? or just the pr fix
and also @cixzhang  is my pr #4244 outdated or like can you review that too?

PR #5001, opened 2026-08-13
fix(charts,richtext,lab): declare `"type": "module"` for the ESM-only dist

### LINKED ISSUE #5000: charts/richtext/lab ship ESM-only dist without `"type": "module"`, so Node resolves `dist/index.js` as CommonJS
## Description

`@astryxdesign/charts`, `@astryxdesign/richtext`, and `@astryxdesign/lab` (all published under the `@canary` dist-tag) compile their `dist/` with Babel using only `@babel/preset-react` + `@babel/preset-typescript` — no module transform — so the output is ESM (`export { Chart } from "./Chart.js"`). But unlike `@astryxdesign/core`, their `package.json` has no `"type": "module"` field, and both `main` and the `exports` `default` condition point at that `.js` output.

Under Node resolution rules, a `.js` file in a package without a `type` field is CommonJS. So any consumer that resolves these packages through Node itself — SSR without transpilation, Vitest/Jest in a node environment, or a plain `import '@astryxdesign/charts'` in a Node script — gets the dist misparsed:

- **Node < 20.19 / < 22.7:** hard failure — `SyntaxError: Cannot use import statement outside a module`.
- **Node ≥ 20.19 / ≥ 22.7:** the module-syntax detection fallback rescues it, but emits a `MODULE_TYPELESS_PACKAGE_JSON` warning **per file** and reparses every module in the graph twice ("This incurs a performance overhead", per Node's own warning text).

Reproduction on Node 22.14, after building charts:

```
$ node -e "import('@astryxdesign/charts')"
(node:28188) [MODULE_TYPELESS_PACKAGE_JSON] Warning: Module type of
file:///.../packages/charts/dist/index.js is not specified and it doesn't
parse as CommonJS.
Reparsing as ES module because module syntax was detected. This incurs a
performance overhead.
To eliminate this warning, add "type": "module" to .../packages/charts/package.json.
```

Bundler consumers are unaffected (they resolve the `source`/`module` conditions or handle ESM regardless), which is why this hasn't surfaced in the storybook/docsite apps.

That the dist is *meant* to be `"type": "module"` is already documented in the repo itself — `scripts/check-fully-specified.mjs` (which runs in these packages' builds) says:

> The packages publish "type": "module" dist, where Node ESM makes extensions mandatory for relative specifiers

…and `@astryxdesign/core` declares the field. The three canary packages just never got it.

Note: `@astryxdesign/vega` is **not** affected — it builds with tsup and ships a real dual build (`require` → `dist/index.js` CJS, `import` → `dist/index.mjs`).

## Affected

- `packages/charts/package.json` (no `type` field; `main`/`exports.default` → `./dist/index.js`)
- `packages/richtext/package.json` (same)
- `packages/lab/package.json` (same)

## Proposed fix

Add `"type": "module"` to the three manifests, matching core. One wrinkle: each of these packages has a CommonJS `babel.config.js` (`module.exports = …`) in its root, which `"type": "module"` would break — it needs renaming to `babel.config.cjs` along with the `--config-file` references in the `build:esm`/`dev` scripts. (`babel-plugin-add-extensions.cjs` already carries the right extension.)

Happy to send a PR.

is-jain:
Fixes #5000

`@astryxdesign/charts`, `@astryxdesign/richtext`, and `@astryxdesign/lab` compile `dist/` to ESM (Babel with no module transform, extensions added by `babel-plugin-add-extensions.cjs`), but their manifests never declared `"type": "module"` — unlike `@astryxdesign/core`. Node therefore classifies `dist/index.js` as CommonJS: a hard `SyntaxError: Cannot use import statement outside a module` on Node < 20.19 / < 22.7, and a per-file `MODULE_TYPELESS_PACKAGE_JSON` warning plus double-parse overhead on newer Node, for any consumer resolving through Node itself (SSR without transpile, node-env test runners, plain scripts). Bundler consumers resolve the `source` condition or handle ESM anyway, which is why the storybook/docsite apps never surfaced it. The intent was already on record — `scripts/check-fully-specified.mjs` states "The packages publish \"type\": \"module\" dist".

Changes, identical across the three packages:

- Add `"type": "module"` to `package.json`, matching core.
- Rename `babel.config.js` → `babel.config.cjs`: the config is CommonJS (`module.exports`), which `"type": "module"` would otherwise break. The `--config-file` references in `build:esm`/`dev` and the file headers are updated to match. (`babel-plugin-add-extensions.cjs` was already `.cjs`.)

No changeset: `check:changesets` rejects entries for these packages ("private/ignored and cannot be released"), so their canary versioning is unaffected by this PR.

`@astryxdesign/vega` is untouched — it ships a real dual CJS/ESM build via tsup and is not affected.

## Verification

On Node v22.14.0, after rebuilding charts' dist:

- Before: `import('@astryxdesign/charts')` emits `MODULE_TYPELESS_PACKAGE_JSON` per file ("Reparsing as ES module… This incurs a performance overhead"); on Node without syntax detection it throws `SyntaxError`.
- After: `import()` and `require()` both resolve cleanly, 30 exports, no warnings.
- `build:esm` runs successfully with the renamed config in all three packages.
- eslint passes on the renamed configs.
- Vitest across the three packages: 520/522 pass; the two failures are in `packages/lab/src/TransferList/TransferListSelector.test.tsx` and reproduce identically on `upstream/main` without this change (timing-sensitive, failure count varies run to run).


cixzhang (changes requested):
Thanks for this — the diagnosis is solid and I verified it independently: no module transform in the three babel configs, no `require` condition, no `.cjs` artifact, so `"type": "module"` is genuinely the manifest catching up to what the build already emits. The `MODULE_TYPELESS_PACKAGE_JSON` repro is exactly the right evidence.

Requesting changes on the direction rather than the correctness, following [my note on #5000](https://github.com/facebook/astryx/issues/5000#issuecomment-0): these packages should *match* `core`, not acquire a second setup of their own.

1. **The `.cjs` rename moves away from core, not toward it.** `core` has no `babel.config.js` at all — it uses `babel.config.json` (`packages/core/babel.config.json`), which is format-agnostic and immune to this whole problem. Please try JSON for all three first. The only dynamic bits are `rootDir` and the two `@astryxdesign/core` aliases, and core's own config proves a relative `rootDir` works (`"rootDir": "."`). If StyleX genuinely needs absolute alias paths and JSON can't express them, keep `.cjs` — but say so in the file header, so the next person knows it's a constraint and not drift.

2. **Drop `"module": "./dist/index.js"`** (charts:28, lab:28, richtext:30). `core` doesn't have it, and with `type: module` the field is stating what the manifest now says. That's the last divergence between the four.

3. Not yours to fix, but worth an issue while it's fresh: `scripts/verify-exports.mjs` skips `private` packages (`:178`) and puts `default` in `NON_PROBED_CONDITIONS` (`:120`). Two independent reasons the repo's one import smoke test could never have caught this. That gap is the real reason this shipped.

is-jain:
Thanks for the review — all three points addressed:

1. **babel.config.json works.** You were right that JSON should be tried first: the StyleX plugin resolves `unstable_moduleResolution.rootDir` and alias paths with `path.relative`/`path.join`, which resolve relative inputs against the process cwd — and babel always runs from the package root here (both `pnpm build` and `--config-file ./…`). So `"rootDir": "../.."` and `"@astryxdesign/core/*": ["../../packages/core/src/*"]` replace the `__dirname`-based absolute paths with no behavior change. Verified by building all three packages with both config formats and diffing: dist output is **byte-identical**. The `.cjs` files are gone; the three configs are now `babel.config.json`, shaped like core's with only the alias block added.

2. **`module` field dropped** from all three manifests.

3. Filed as #5132.

cixzhang (approved):
Thanks — the JSON configs give byte-identical output; I built all three and diffed. Approving. I merged main in to clear the package.json conflicts.

[Reviewed by Robohands]

Issue #4988, opened 2026-08-13
[RFC] Independent secondary actions in hierarchical SideNav rows


jackeyfaker77:
### Problem Statement

Applications with hierarchical side navigation often need two independent controls on one visual row: a primary navigation/disclosure target and a row-level action such as a context menu.

Today, `SideNavItem.endContent` is rendered inside the primary interactive element. Rendering an `IconButton` or `MoreMenu` there would create nested interactive controls. Rendering the action after `</SideNavItem>` avoids nesting, but because `SideNavItem` owns its nested children, the action appears after the entire subtree in DOM and keyboard order. Rendering it before the item fixes that ordering, but separates the action from the row's layout contract and forces consumers to reproduce SideNav geometry.

The desired experience is a single row whose primary and secondary controls are semantic siblings, with all row-level controls preceding nested items in DOM/focus order.

### Evidence of Demand

- [Maka PR #2944](https://github.com/maka-agent/maka-agent/pull/2944) demonstrates the problem on project rows that contain both a `MoreMenu` (new/rename/archive actions) and nested session rows. The Maka maintainer confirmed the DOM-order defect and suggested bringing the row-action seam upstream.
- [Maka PR #2951](https://github.com/maka-agent/maka-agent/pull/2951) demonstrates the available call-site workaround: render the action before `SideNavItem` and position it externally. That restores action-before-children order, but requires consumer-owned geometry.
- Astryx already has the right structural precedent in [SideNavItem's split-action path](https://github.com/facebook/astryx/blob/757185e160c803819523532731c04196a2f23afc/packages/core/src/SideNav/SideNavItem.tsx#L620-L660): a primary action and collapse toggle are sibling controls in one row.
- [shadcn/ui's `SidebarMenuAction`](https://ui.shadcn.com/docs/components/aria/sidebar#sidebarmenuaction) models a separate action alongside a sidebar menu button.
- [MUI's `ListItem.secondaryAction`](https://mui.com/material-ui/react-list/#secondary-action) models a separate trailing target for list rows.

This is therefore a recurring sidebar/list-row pattern, not a Maka-specific feature.

### Why Existing Components Don't Cover This

`endContent` is suitable for badges and counts, but it is part of the primary link/button content and cannot safely contain another interactive target.

An external wrapper can produce valid HTML:

```
<div className="projectRow">
  <MoreMenu ... />
  <SideNavItem label="Project">...</SideNavItem>
</div>
```

[...]

PR #5005, opened 2026-08-13
feat(sidenav): add actions slot for row controls

AKnassa:
Refs #4988

## What this does

Adds an optional `actions` slot to `SideNavItem` for row-level secondary controls (icon buttons, menus). The slot renders as a sibling of the primary link or button, so a row can carry a context menu without nesting interactive elements.

## Why

`endContent` renders inside the primary link or button, so placing a `MoreMenu` there nests a button inside a link. Rendering the menu outside the item instead lands it after the whole nested subtree in DOM and keyboard order. #4988 asks for a first-class seam: sibling row controls, all reachable before focus enters the children.

## What changed

- `SideNavItem` accepts `actions?: ReactNode`. The row renders primary element, then the expand/collapse chevron, then actions at the trailing edge, then the nested children group. The row wrapper is added only when actions are present, so markup for existing items is unchanged.
- Actions are hidden while the rail is collapsed, and stay interactive on disabled items: each supplied control owns its own disabled state.
- 17 new tests: sibling structure, `compareDocumentPosition` order, tab order in both row modes, click isolation, drawer close suppression, disabled-item action clicks, aria wiring on the toggle, inert collapsed children, controlled collapse, custom link components, rail omission (mutation-checked).
- Docs: prop JSDoc, `SideNavItem.doc.mjs` (en, zh, dense), patch changeset.
- Storybook: new "Row Actions (Sibling Controls)" story with `MoreMenu` on collapsible project rows. The End Content story keeps only passive content (badges, text) so the catalog stops demonstrating a menu inside a link.

## Open questions for the design brief

Three picks here are reversible and flagged for maintainer input, per the RFC:

- Prop name: `actions`, matching `EmptyState.actions`. The `*Action` suffix stays reserved for async handlers (`clickAction`, `changeAction`).
- Order: actions at the trailing edge, after the chevron. The whole-row-toggle case keeps its chevron inside the toggle button, so trailing actions is the one order consistent across all row types. Matches MUI `ListItem.secondaryAction` and shadcn `SidebarMenuAction`.
- The collapsed rail omits actions; whether the collapsed popover should host them is left open.

## How to see it

Storybook: Core/SideNav, "Row Actions (Sibling Controls)" story. Tab through a project row: link, chevron, menu, then the nested sessions.

## Verification

Full suite: 9957 passed, 2 failed, both the pre-existing macOS case-insensitivity failures in `packages/cli`, identical to the untouched baseline on `7f247bc58`. eslint with CI=true, core and storybook typecheck, and all `check:repo` gates green.


cixzhang (changes requested):
Thanks @AKnassa — I rebased this locally and checked it in Chromium, and it holds up nicely: hover/pressed paint stays continuous behind the actions, RTL mirrors, the collapsed rail hides them, and ordinary rows render pixel-identically.

Two things:

1. **Focus ring target.** A row with `actions` gets a square ring around the primary element only (192×32) instead of the full-row pill an ordinary row gets (244×32) — it's `focusOutlineProps.focusVisible(styles.splitAction)` at `SideNavItem.tsx:692`. Keep `focusVisible`, but move the target to the row wrapper so the shape matches a plain row.

2. **`endContent` docs.** The `actions` entry already steers people correctly, but `endContent` still reads "Right-side content (badges, counts)" (`SideNavItem.tsx:327`, `SideNavItem.doc.mjs:209`) — that's where an author looks. Please say passive-only there, in all three locales.

Also worth a rebase — [#4880](https://github.com/facebook/astryx/pull/4880) rebuilt this same block on `main`.

kentonquatman (dismissed):
It looks like we have different size buttons nested within the nav item. Can we update so they're consistent?

Image
Image

AKnassa:
Fixed in 1dd65df45.

The toggle carried no box of its own, so it shrank to its 24px chevron beside the 28px menu button. It now takes the box of a `size="sm"` icon button, and the actions slot cascades that size through `SizeContext` the way `SideNav` already does for `footerIcons`, so an unsized control matches by default and an explicit `size` still wins.

Measured in Chromium: every control in the row is 28x28. The story dropped its now redundant `size="sm"` props so a regression would surface there.

sidenav-5005-chevron-hover-after
sidenav-5005-menu-hover-after

cixzhang (approved):
Thanks — this is good. @kentonquatman's size point and my two are all answered; I re-checked in Chromium. Approving; holding the merge for design sign-off.

[Reviewed by Robohands]

cixzhang:
I merged current main and resolved the two mechanical SideNav overlaps: the actions-slot conditional stays, while main's interaction overlay and test setup stay too. Pushed `627bd099`; CI is running.

cixzhang (approved):
Re-approving the refreshed head. The requested control sizing, focus-ring target, and passive-only `endContent` documentation are all fixed; required CI and visual acceptance are green.

[Reviewed by Robohands]

PR #5010, opened 2026-08-14
fix(selector): overlay the trigger only when the selected option covers it (#5004)

### LINKED ISSUE #5004: Selector: the overlay menu covers its own trigger — clicking it commits the option on top instead of closing
## What happens

With the default (selected-item overlay) placement, `Selector`'s open menu is positioned **on top of its own trigger** — that is the point of the overlay. The consequence is that the trigger is no longer clickable: a click at the trigger's own coordinates lands on whichever option is painted over it, and that option is **committed as the value**.

Hit-testing the trigger of an open, default `Selector` in Chrome 151 (`document.elementFromPoint` at the trigger's top edge, centre, and bottom edge) returns an option inside the popover in every case, and clicking there changes the value:

| story | value before | click the trigger | value after |
|---|---|---|---|
| `Core/Selector` → Default (placeholder state) | `Select a fruit…` | hits option **Apple** | **Apple** |
| `Core/Selector` → With Status (placeholder state) | `Select a fruit…` | hits option **Apple** | **Apple** |
| `Core/Selector` → Pre-Selected | `Banana` | hits option **Banana** | `Banana` |

So a form that has never been touched can be given a value by a user who was only trying to dismiss the menu — and the menu closes at the same time, which makes it read as an ordinary dismissal. The placeholder state is the one every form starts in.

Two more consequences of the same geometry:

- **There is no click-to-close.** Every "click the button again to close it" gesture is a selection instead. (Playwright's actionability check refuses the click outright — `<div popover> subtree intercepts pointer events` — on Default, Clearable, Ghost Variant, Hidden Label, With Description, Pre-Selected and Size Variants.)
- **WCAG 2.4.11 (Focus Not Obscured).** The trigger keeps DOM focus while the menu is open — `Selector` deliberately does not move focus into the listbox — and the menu covers it.

## What is *not* broken

The trigger toggles correctly in every non-overlay mode. Verified on `main` and on the [#4976](https://github.com/facebook/astryx/pull/4976) preview, in Chrome 151 and WebKit, with mouse (instant and 120 ms hold) and with touch: `placement="above"`, `placement="below"`, `hasSearch`, and `MultiSelector` all close on a second trigger click. The light-dismiss `toggle` event is queued and always arrives *after* the trigger's `click`, so the "closed by light dismiss, then re-opened by the click" race that [`DropdownMenu`](https://github.com/facebook/astryx/blob/main/packages/core/src/DropdownMenu/DropdownMenu.tsx) and [`Popover`](https://github.com/facebook/astryx/blob/main/packages/core/src/Popover/Popover.tsx) guard against with a 50 ms window does not fire on desktop here.

## Fix

[#4976](https://github.com/facebook/astryx/pull/4976) removes this by flipping the default placement to below. If the overlay survives as opt-in (`hasSelectedItemOverlay`), it needs its own answer for click-to-close and for 2.4.11 — a native `<select>` gets this right only because the platform popup swallows the whole gesture.

Related: [#4227](https://github.com/facebook/astryx/issues/4227).

cixzhang:
Closes [#5004](https://github.com/facebook/astryx/issues/5004).

## The bug

The selected-item overlay paints the menu **on top of its own trigger**, so whichever option lands over the trigger is what a press on the trigger commits. That is harmless in exactly one case — when the option sitting there is the one already selected, the press re-commits the current value and reads as a dismissal (this is what a native `<select>` does).

Two states break that invariant, and both were live. Measured in Chrome 151 against `main`, pressing the trigger of an open menu at its own coordinates:

| state | option painted over the trigger | value before → after |
| --- | --- | --- |
| placeholder (no selection) | `Apple` | `Pick one…` → **`Apple`** |
| selection past the fold, trigger near the viewport top | `Option 31` | `Option 38` → **`Option 31`** |
| selection aligned (healthy overlay) | `Banana` | `Banana` → `Banana` ✅ |

Identical with mouse and with touch. The second row is the worse one: an existing value is silently replaced by a *different* one.

## It is geometry, not a popover race

Worth stating because it is the obvious suspect. Instrumenting the second press: `click` at the trigger, then the light-dismiss `toggle` **15 ms later** — the toggle always arrives after the click, so the "closed by light dismiss, then re-opened by the click" race that [DropdownMenu](https://github.com/facebook/astryx/blob/main/packages/core/src/DropdownMenu/DropdownMenu.tsx) and [Popover](https://github.com/facebook/astryx/blob/main/packages/core/src/Popover/Popover.tsx) guard with a 50 ms window never fires here. Nothing in the timing needs fixing.

## The rule

Use the overlay only when the option under the trigger is the selected one. Nothing selected means no such option; a viewport clamp that slides the list can put a different one there. In both cases the menu opens below with the standard `--spacing-1` clearance — the same geometry every other Astryx menu uses, and the trigger stays uncovered, so click-to-close works.

`useSelectedItemOffset` now reports whether the target row still covers the anchor after clamping (it already computed everything needed); the tolerance is the one deliberate optical pixel the row is shifted by. Where the invariant holds, the measured offset is unchanged.

## How this sits with [#4976](https://github.com/facebook/astryx/pull/4976)

Orthogonal, and this does not pre-empt that decision. [#4976](https://github.com/facebook/astryx/pull/4976) decides what the *default* placement should be; this fixes the overlay itself, which survives there as the opt-in `hasSelectedItemOverlay` — carrying both failures with it. Whichever lands first, the other rebases onto the same few lines.

## Risk

Placeholder-state Selectors — the state every form starts in — now open below with a gap instead of over the trigger. That is a visible change, and a subset of the one [#4976](https://github.com/facebook/astryx/pull/4976) proposes for all Selectors. Clamped menus open below rather than pinning to the viewport edge. `useSelectedItemOffset` is exported from the package; its return type gains a field, which is additive.

## Adjacent, not fixed here

- **A one-pixel sliver survives in the healthy overlay.** The control box measures `[396, 428]` against the selected option's `[395, 427]` — that bottom pixel row hits the *next* option. It comes from the deliberate 1 px optical offset, so straightening it is a visual-alignment call, not a geometry bug.
- **WCAG 2.4.11 still applies to overlay mode**: the trigger keeps DOM focus and the menu covers it. Inherent to the affordance; noted in [#5004](https://github.com/facebook/astryx/issues/5004).

## Test plan

**Unit** — two new tests, both written red-first (they fail on unmodified `Selector.tsx`/`hooks.ts`): no selection, and a clamp that slides another option over the trigger. The overlay geometry tests were asserting the clamped case as correct behavior, so they now run against rects where the alignment actually lands (`-61px`), which is what they were meant to cover. `packages/core/src/Selector`: 120 passing. Full `packages/core`: 6310 passing — the one failure is the Table 500-row perf budget, which flakes under full parallel load on this laptop and passes in isolation.

**Real Chromium** (Playwright against local Storybook, mouse and touch, pressing the trigger of an open menu at its own coordinates):

| story | before | after |
| --- | --- | --- |
| Default (placeholder) | commits `Apple` | closes, value untouched |
| clamped selection | commits `Option 31` | closes, value untouched |
| Pre-Selected (healthy overlay) | closes, value untouched | unchanged — still overlays |
| `placement="above"` / `hasSearch` / MultiSelector | closes | unchanged |

**Repo** — `tsc --noEmit` clean, strict ESLint clean, `pnpm check:repo` green.

### DIFF
--- a/.changeset/selector-overlay-invariant.md
+++ b/.changeset/selector-overlay-invariant.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Selector: only overlay the menu on the trigger when the selected option is what lands there. With no selection, or when the viewport clamp slides a different option over the trigger, the menu now opens below with the standard clearance — a press on the trigger dismissed it instead of committing whatever was painted on top
+
+@cixzhang
--- a/packages/core/src/Selector/Selector.test.tsx
+++ b/packages/core/src/Selector/Selector.test.tsx
@@ -379,7 +379,10 @@ describe('Selector', () => {
     );
   });
 
-  it('clamps the default selected-item overlay to the viewport', async () => {
+  it('opens below when the viewport clamp slides another option over the trigger', async () => {
+    // The default rects clamp the menu to a 200px viewport, which lands the
+    // selected option well above the trigger — the option covering it would be
+    // committed by a press meant to dismiss the menu (#5004).
     const restoreRects = mockSelectorRects();
     const user = userEvent.setup();
     try {
@@ -395,12 +398,40 @@ describe('Selector', () => {
       await user.click(screen.getByRole('combobox'));
       const popover = screen
         .getByRole('listbox', {hidden: true})
-        .closest('[popover]');
+        .closest('[popover]') as HTMLElement;
       await waitFor(() => {
-        expect(popover?.getAttribute('style')).toContain(
-          'margin-block-start: -110px',
+        expect(popover.style.getPropertyValue('--x-marginBlockStart')).toBe(
+          spacingVars['--spacing-1'],
         );
       });
+      expect(popover.getAttribute('style')).not.toContain(
+        'margin-block-start: -',
+      );
+    } finally {
+      restoreRects();
+    }
+  });
+
+  it('opens below when nothing is selected', async () => {
+    const restoreRects = mockSelectorRects({viewportHeight: 800});
+    const user = userEvent.setup();
+    try {
+      render(
+        <Selector label="Fruit" options={OPTIONS} placeholder="Pick one" />,
+      );
+
+      await user.click(screen.getByRole('combobox'));
+      const popover = screen
+        .getByRole('listbox', {hidden: true})
+        .closest('[popover]') as HTMLElement;
+      await waitFor(() => {
+        expect(popover.style.getPropertyValue('--x-marginBlockStart')).toBe(
+          spacingVars['--spacing-1'],
+        );
+      });
+      expect(popover.getAttribute('style')).not.toContain(
+        'margin-block-start: -',
+      );
     } finally {
       restoreRects();
     }
@@ -563,7 +594,7 @@ describe('Selector', () => {
     });
 
     it('stays flush in the default selected-item overlay', async () => {
-      const restoreRects = mockSelectorRects();
+      const restoreRects = mockSelectorRects({viewportHeight: 800});
       const user = userEvent.setup();
       try {
         render(
@@ -581,7 +612,7 @@ describe('Selector', () => {
           .closest('[popover]') as HTMLElement;
         await waitFor(() => {
           expect(popover.getAttribute('style')).toContain(
-            'margin-block-start: -110px',
+            'margin-block-start: -61px',
           );
         });
         expect(popover.style.getPropertyValue('--x-marginBlockStart')).toBe('');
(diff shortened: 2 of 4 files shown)


cixzhang:
Parking this: the reported symptom is the trigger press closing and immediately reopening the popup — a light-dismiss/click race, not the overlay geometry — and that is fixed in [#5018](https://github.com/facebook/astryx/pull/5018), which also stops the clear and status buttons on a trigger from dismissing the popup they belong to.

The geometry finding here still stands and is documented in [#5004](https://github.com/facebook/astryx/issues/5004): with the selected-item overlay, the option painted over the trigger is what a press on the trigger commits, so a placeholder Selector can be given a value by someone trying to dismiss the menu. Where that lands depends on the default-placement decision in [#4976](https://github.com/facebook/astryx/pull/4976), so it should not be settled by a drive-by fix from me. The branch stays if we want it back.

PR #5018, opened 2026-08-14
fix(core): stop popup triggers from fighting their own light dismiss (#5004)

cixzhang:
Fixes the trigger side of [#5004](https://github.com/facebook/astryx/issues/5004).

## The mechanism

A light dismiss and the trigger's own click are **one press**. The browser dismisses an `auto` popover on `pointerup` and queues the `toggle` event; the `click` follows a beat later. Which of the two React sees first is a race:

| order | what the trigger's handler reads | result |
| --- | --- | --- |
| click first, then `toggle` | still open | closes — correct, by luck |
| `toggle` first, then click | already closed | **reopens** — the button looks like it does nothing |

`DropdownMenu`, `Popover`, and `ComplexSelector` each carried a private 50 ms timing guard against this; `Selector` and `MultiSelector` carried nothing. A guard is also the wrong shape: a window is a guess about how long the two halves of one press can be apart, and the case where the race is lost — a blocked main thread — is exactly the case where the window is too short. I saw that directly: at 50 ms the new tests pass alone and fail when the suites run in parallel.

## The fix

**Count gestures, don't time them.** A tiny module counts `pointerdown` and `keydown` on the document; the layer records which gesture the browser dismissed it in, and `wasJustDismissed()` is true only while that same gesture is still in flight. A click from the dismissing press is absorbed however long the main thread was blocked; a deliberate second press is a new gesture and always acts. No constant to tune. Both hand-rolled copies collapse into it.

**The same dismissal also fires for controls that sit on the trigger** — the clear ✕ and the status button. They live outside the popover, so pressing one dismissed the popup, which means the affordance and the popup it belongs to could never be used together. An internal helper temporarily names those controls as invokers, keeping this mechanism out of the public `useLayer` and `usePopover` return types until the rest of the clearable popup inputs adopt it. The attribute is stamped for the press only: a permanent invoker reports itself as `expanded` to assistive tech, which I confirmed in Chrome's AX tree — a Clear button must not.

## Test plan

**Unit — 8 new tests, all red against unmodified source** (verified by restoring the source files from `main` and rerunning): the trigger click absorbed when the dismissal lands first, a deliberate second press still acting, a programmatic hide left unguarded, the invoker stamped for the press and removed after, the invoker's own toggle cancelled, and one regression test each for `Selector`, `MultiSelector` and `ComplexSelector` driving the losing order directly. `packages/core`: 6318 passing (the one failure is the Table 500-row perf budget, which flakes under parallel load here and passes in isolation).

**Real browsers** — Playwright, Chromium and WebKit iPhone touch, pressing the clear ✕ and the status button of an open menu:

| | before | after |
| --- | --- | --- |
| MultiSelector clear ✕ | menu closes | menu stays open, value cleared |
| Selector clear ✕ (`placement="below"`) | menu closes | menu stays open |
| Selector status button | menu closes | menu stays open |

Trigger toggling is unchanged in both engines: one press, one state change, and the third press reopens.

**On the race itself:** I could not make Chromium or WebKit lose it on this machine — both consistently deliver the click first, which is why the bug reads as intermittent and environment-dependent. The regression tests therefore drive the losing order explicitly rather than hoping for it. If you can reproduce the reopen somewhere concrete, that environment is the real check on this.

## Notes

- `useSelectedItemOffset` and the overlay geometry are untouched here.
- Other clearable inputs that render their own popup — `Typeahead`, `Tokenizer`, `DateInput`, `DateRangeInput`, `TimeInput` — have the same shape and are **not** wired here. The invoker helper remains internal until that family is handled together.
- With the default selected-item overlay, `Selector`'s clear and status buttons are covered by the menu entirely, so this fix is not visible there — that is the geometry half of [#5004](https://github.com/facebook/astryx/issues/5004), left alone deliberately.

### DIFF
--- a/.changeset/layer-dismiss-gesture.md
+++ b/.changeset/layer-dismiss-gesture.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Popup triggers no longer fight the browser's own light dismiss: pressing the button of an open Selector, MultiSelector, ComplexSelector, DropdownMenu or Popover closes it once instead of closing and reopening, and a clear or status button sitting on the trigger no longer dismisses the popup it belongs to
+
+@cixzhang
--- a/packages/core/src/ComplexSelector/ComplexSelector.test.tsx
+++ b/packages/core/src/ComplexSelector/ComplexSelector.test.tsx
@@ -353,6 +353,7 @@ describe('ComplexSelector', () => {
       .closest('[popover]');
     expect(popover).not.toBeNull();
 
+    fireEvent.pointerDown(trigger);
     const closeEvent = new Event('toggle');
     Object.defineProperty(closeEvent, 'newState', {value: 'closed'});
     fireEvent(popover as HTMLElement, closeEvent);
@@ -432,6 +433,7 @@ describe('ComplexSelector popup theme target', () => {
     // The browser dismissed the popup on pointerup and queued the toggle. When
     // that event lands before the click — WebKit, or any engine under load —
     // the click used to read a closed popup and reopen it.
+    fireEvent.pointerDown(trigger);
     const popover = document.querySelector('[popover]') as HTMLElement;
     act(() => {
       popover.dispatchEvent(
--- a/packages/core/src/ComplexSelector/ComplexSelector.test.tsx
+++ b/packages/core/src/ComplexSelector/ComplexSelector.test.tsx
@@ -417,4 +417,34 @@ describe('ComplexSelector popup theme target', () => {
       document.querySelector('.astryx-complex-selector-popup'),
     ).not.toBeNull();
   });
+
+  it('stays closed when the trigger click follows its own light dismiss (#5004)', async () => {
+    const user = userEvent.setup();
+    render(
+      <ComplexSelector label="Fruit blend" value="Apple" triggerLabel="Apple">
+        {() => <button type="button">Done</button>}
+      </ComplexSelector>,
+    );
+    const trigger = screen.getByRole('button', {name: 'Fruit blend'});
+    await user.click(trigger);
+    expect(trigger).toHaveAttribute('aria-expanded', 'true');
+
+    // The browser dismissed the popup on pointerup and queued the toggle. When
+    // that event lands before the click — WebKit, or any engine under load —
+    // the click used to read a closed popup and reopen it.
+    const popover = document.querySelector('[popover]') as HTMLElement;
+    act(() => {
+      popover.dispatchEvent(
+        Object.assign(new Event('toggle'), {
+          oldState: 'open',
+          newState: 'closed',
+        }),
+      );
+    });
+    // Synchronously: the click falls inside the one gesture the guard covers,
+    // as it does in a browser a few milliseconds behind the dismissal.
+    fireEvent.click(trigger);
+
+    expect(trigger).toHaveAttribute('aria-expanded', 'false');
+  });
 });
--- a/packages/core/src/ComplexSelector/ComplexSelector.tsx
+++ b/packages/core/src/ComplexSelector/ComplexSelector.tsx
@@ -373,14 +373,12 @@ export function ComplexSelector<Value>({
       .join(' ') || undefined;
 
   const triggerRef = useRef<HTMLButtonElement>(null);
-  const lastHideTimeRef = useRef(0);
 
   const [isPending, startTransition] = useTransition();
   const [optimisticValue, setOptimisticValue] = useOptimistic(value);
   const isBusy = isLoading || isPending;
 
   const handlePopoverHide = useCallback(() => {
-    lastHideTimeRef.current = Date.now();
     triggerRef.current?.focus();
   }, []);
 
@@ -395,7 +393,7 @@ export function ComplexSelector<Value>({
   const isOpen = popover.isOpen;
 
   const handleTriggerClick = useCallback(() => {
-    if (isDisabled || Date.now() - lastHideTimeRef.current < 50) {
+    if (isDisabled) {
       return;
     }
     if (popover.isOpen) {
--- a/packages/core/src/DropdownMenu/DropdownMenu.test.tsx
+++ b/packages/core/src/DropdownMenu/DropdownMenu.test.tsx
@@ -350,6 +350,7 @@ describe('DropdownMenu light-dismiss race', () => {
   it('does not re-open when the trigger click follows its own light dismiss', () => {
     const trigger = openMenu();
 
+    fireEvent.pointerDown(trigger);
     lightDismiss();
     fireEvent.click(trigger);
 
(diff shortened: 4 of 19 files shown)


cixzhang:
Thanks — the gesture counter is the right primitive, and having one place that knows a dismissal just happened is a good direction.

Needs a merge with main; [#4769](https://github.com/facebook/astryx/pull/4769) landed today and this will conflict.

Three things:

**The guard only reaches callers that go through `toggle()`.** Anything calling `show()`/`hide()` directly bypasses it — ComplexSelector is the example after the merge, and its new test passes off its own leftover 50ms timer rather than this change. Guarding in `show()` would make it hold for every caller instead of being something each one has to opt into.

**`dismissedByGestureRef` is never cleared** (`useLayer.tsx:485`), so a click with no pointerdown before it — AT activation, `.click()` — is read as part of the dismissing gesture indefinitely. Make it one-shot.

**`keepOpenProps` isn't finished.** `KeepLayerOpenProps` is unexported, `InputClearButton` grew two ad-hoc handler props, and five clearable inputs are unwired — let's keep it internal until Typeahead and the date/time inputs are done rather than ship a half-built API.

cixzhang:
Merged `main` and rewrote the guard against your three points. Pushed to this branch (`4947e62`).

**The guard moved into `show()`.** Every caller lands there, so a trigger no longer opts in — `wasJustDismissed()` is checked before the layer opens, whatever route the caller took. ComplexSelector's private 50 ms timer is deleted with it. Its two tests now fail when I remove the `show()` guard and pass with it, so they are testing this change and not the timer.

**The dismissal is one-shot.** It is spent by the click that ends the press it came from — a one-shot bubble-phase `click` listener on the document, so every guard reading it has already run. A click with no press of its own (AT activation, `element.click()`) therefore acts. New test: `acts on a synthesized click with no press of its own`, red without the clear.

**One more defect the guard exposed.** `hide()` called `hidePopover()` while `isOpenRef` still read `true`, so a `toggle` event delivered synchronously was read as a browser dismissal — a programmatic hide stamped the gesture. The ref is cleared before the call now, which is what the comment above the reconciler already claimed. DropdownMenu's old race test was passing off exactly that bug (three clicks, no pointerdown, no light dismiss); it is rewritten to drive the losing order, plus a second test that a press of its own re-opens.

## Real Chromium

Playwright, trusted presses, `pnpm storybook`. Chromium delivers `click` before `toggle` on both mouse and touch (`pointerdown → click → toggle:closed`), so it wins the race on its own — as you found. To test the losing order I hand React the dismissal first, from a real press: a bubble-phase `pointerdown` listener dispatches `toggle: closed` on the open popover, then the press's own click arrives. Before/after is the same browser, same driver, `main`'s files swapped in for "before":

| trigger press while open | before | after |
| --- | --- | --- |
| DropdownMenu | closes (its own 50 ms timer) | closes |
| Popover | **re-opens** | closes |
| Selector | **re-opens** | closes |
| MultiSelector | **re-opens** | closes |
| ComplexSelector | closes (its own 50 ms timer) | closes |

Also in Chromium, with the fix: third press re-opens everywhere; a synthesized `.click()` after a dismissing press opens the menu; an outside press dismisses; MultiSelector's clear ✕ clears the value with the menu still open.

## Tests

`Layer, Popover, DropdownMenu, Selector, MultiSelector, ComplexSelector, Field, DateInput, Typeahead, Tooltip, HoverCard, Dialog, CommandPalette` — 1011 passing. Full `packages/core` is green apart from `DateInputTouch` timeouts under parallel load, which reproduce identically on the merge commit before this change (they pass in isolation, 133/133). Typecheck clean, lint 0 errors.

## Against the shared dismissal stack

Merging this branch into [#4881](https://github.com/facebook/astryx/pull/4881) auto-merges `usePopover.tsx` cleanly — the one file both touch. That merge does conflict in `Layer/index.ts`, `Tooltip/useTooltip.tsx` and `useHoverCard.tsx`, but those conflicts are #4881 against today's `main` and appear with this branch left out. Nothing here is subsumed or duplicated either: #4881 routes Escape and the close-watcher through a stack of layers, and pointer light dismiss stays the browser's, so the `toggle` this guard counts on still fires. Order can be whichever suits you.

## Not done

`keepOpenProps` is untouched — your third point (unexported type, two ad-hoc props on `InputClearButton`, five unwired inputs) is an API-scope call, and I did not want to make it on your behalf inside a correctness fix.

[Pushed by Robohands]

cixzhang:
Thanks — the two you took from last round hold up. In Chromium, with the dismissal handed to React before the click, `main` re-opens the menu and this closes it:

| | before | after |
| --- | --- | --- |
| press the trigger of an open menu, dismissal first |  |  |
| press the clear ✕ with the menu open |  |  |
| press the status ⚠ with the menu open |  |  |

One to fix before it lands: after a touch the browser cancels — a press turned into a scroll, a long-press — Chrome's a11y tree has the clear ✕ as an expanded pop-up button, the state your note rules out. `keepOpenProps` takes the `popovertarget` off on `pointerup` only (`useLayer.tsx:691`), and a press has two ends. Listen for `pointercancel` too; the four other press cleanups in core do (`ResizeHandle.tsx:439`, `useSheetGestures.ts`).

`keepOpenProps` is still your call from last round; I'd keep it internal until the date/time inputs are wired. Send the guard on its own once `pointercancel` is covered and it lands today — does the invoker half want to wait for them?

MERGE: hold(new public surface on a shipped hook — your call; plus one a11y fix outstanding)

[Reviewed by Robohands]

cixzhang:
Fixed the `pointercancel` one in e6dcd044. `keepOpenProps` now hangs one cleanup off both ends of the press, so whichever arrives first takes `popovertarget` off and removes both listeners — the shape `ResizeHandle.tsx`, `useTableColumnResize.tsx`, `usePointerDragScroll.ts` and `useSheetGestures.ts` already use.

Driven the way you found it: CDP `touchStart` → `touchCancel`, no `touchEnd`, on the clear ✕ of an open MultiSelector, read out of Chrome's own AX tree.

```
before  role=button  name="Clear all Technologies"
        props=["invalid=false","focusable=true","expanded=true","details="]   popovertarget=_r_9_
after   role=button  name="Clear all Technologies"
        props=["invalid=false","focusable=true"]                              popovertarget absent
```

Both sides still report `expanded=true` mid-press, which is the point of the stamp; only the cancelled press used to keep it. An ordinary tap is unchanged — value cleared, menu still open. New unit test for the cancel path (red without the fix), and 589 tests across Layer, Selector, MultiSelector, Popover, DropdownMenu, ComplexSelector and Field pass.

Chromium only: WebKit will not launch on this machine today, and WebKit is the engine most likely to claim the gesture, so that row is still unverified.

`keepOpenProps` as public surface is untouched here — still the open question.

imdreamrunner (approved):

cixzhang:
API ruling: keep `keepOpenProps` internal until the other clearable popup inputs are wired. The shared gesture-identity fix remains the right owner for the dismissal race; publishing a half-adopted props bag does not.

cixzhang:
Thanks — this now holds up. Capture-phase click tracking fixes the stopped-propagation hole; exact-head browser paths and targeted tests are clean.

[Reviewed by Robohands]

cixzhang:
/accept-visual 33041860677/1 Behavior-only event handling; exact-head stable visual regression is green and manual Chromium checks confirm only intended open-state behavior changes.

Issue #5014, opened 2026-08-14
Overriding one side of a colour pair silently voids the contrast guarantee


cixzhang:
`expandColorScale` guarantees text at ≥4.5:1 and `--color-border-emphasized` at ≥3:1 against the surfaces it generates, for any accent. The guarantee is real — and it is silently void for any pair where the author writes one side by hand in `tokens`. Nothing warns, and the loss is invisible until someone measures the rendered pixels.

Measured on six independently authored themes (same brief set, two authoring aids):

**A theme that hand-tuned its accent for dark mode:**

```ts
color: {accent: '#0064E0', neutralStyle: 'warm', contrast: 'high'},
tokens: {
  '--color-accent': ['#0064E0', '#1E6FE0'],   // no --color-on-accent
}
```

The scale generated `--color-on-accent: #002D80` for the *original* accent. The rendered primary button in dark mode:

| | light | dark |
|---|---|---|
| this theme | 5.39:1 | **2.61:1** |
| a sibling theme that changed nothing | 6.46:1 | 7.34:1 |

Read off the real DOM (`getComputedStyle` on `core-button--primary`), not off the tokens. The author had run their own contrast math on the accent — against white, which is not what the theme renders. This one was a medication-administration UI, and the fix is one token.

**Same pattern on borders:** all three themes that hand-wrote `--color-border-emphasized` landed below 3:1 (1.46, 1.58, 1.66–1.86); the themes that left neutrals to `color: {}` stayed above it.

Options, roughly in order of how much I like them:

1. **Warn at build.** `astryx theme build` already resolves every token for both modes. It could check the pairs it knows about — `on-accent`/`accent`, `on-error`/`error`, `on-warning`/`warning`, `on-success`/`success`, text/surface, `border-emphasized`/surface — and warn below threshold. It knows which side was authored, so the message can name the missing counterpart.
2. **Derive the `on-*` token** from an authored base colour instead of leaving the generated one behind, unless it too was authored.
3. **Document it** where authors are looking: the `color` config docstring says what it generates but not that overriding one side of a pair drops the guarantee.

(1) and (3) at minimum. Happy to take (1) if there is agreement on the pair list and on warn-vs-error.

cixzhang:
Fix in #5281 — `astryx theme build` now measures ten pairs in both modes and names the token you forgot.

[...]

PR #5025, opened 2026-08-14
fix(theme): meet AA for dark-mode error label text

### LINKED ISSUE #5019: Default error colour fails AA in dark mode (white on #F5394F = 3.76:1)
The default `--color-on-error` / `--color-error` pair is **3.76:1** in dark mode — white `#FFFFFF` on `#F5394F`. Below AA (4.5:1) for normal text.

It is not theoretical. Any theme that leaves the status colours alone renders a destructive button that way; measured off the real DOM (`core-button--destructive`, dark, `getComputedStyle`):

```
rgb(245, 57, 79) / rgb(255, 255, 255)  → 3.76   ("Delete")
```

Light mode is fine at 4.70:1, and the shipped themes are fine because they all happen to retone error. It is the default that fails, so it hits exactly the themes that trusted the defaults: six of nine independently written themes in a recent authoring experiment shipped it.

`--color-on-accent` / `--color-accent` is in the same position at **3.11:1** in dark (`#FFFFFF` on `#2694FE`), though a primary button is arguably large-text/UI-component territory (≥3:1) where error text usually is not.

Both are visible to `expandColorScale`'s own machinery — the generated palettes are held to 4.5:1 by tone spacing, and the status colours are the documented exception that falls through to `colorDefaults`. Two candidates:

1. Darken the dark-mode error (a tone step or two) until white clears 4.5:1, and check the same for accent.
2. Bring the status pairs inside the generator's guarantee instead of exempting them, so `on-*` is derived from whatever `*` ends up being.

I would do (1) now and consider (2) with #5014, which is the general version of this — a hand-authored side of a pair losing the guarantee silently. A contrast assertion over the default token pairs would keep both closed.

AKnassa:
Fixes #5019.

## What

Two moves, one goal: the error colour meets AA in dark mode without trading one failure for another.

**1. The token.** The default `--color-error` now uses `#E3193B` in both modes — the same shape as `--color-success`, which already shares one hex across modes. The dark half was `#F5394F`, which held white label text at 3.76:1, below WCAG AA's 4.5:1 for normal text, so any theme that left the status colours alone rendered a failing destructive button in dark mode. `#E3193B` is one step darker on the same ramp and is already the dark half of `--color-background-error-inverted`, so white-on-`#E3193B` in dark mode was an approved pairing in this palette before this change.

**2. The text roles.** Measuring every `--color-error` consumer (43 across core/lab/charts) showed the token also paints *text* — and no single hex can clear 4.5:1 under white as a fill and 4.5:1 as text on the dark surfaces; the luminance windows don't overlap. The fill token was already below AA as dark text before this change (4.37:1 on the surface, 3.87:1 on the popover). Components rendering supporting/status text in the fill token now use `--color-text-red`, the pairing FieldStatus and ChatComposer already ship: Chat message metadata ("failed to send"), Chat tool-call stats and errors (the error icon's `currentColor` tint halo follows its glyph), the TextArea and RichTextEditor over-limit counters, lab's ChatUnreadDivider label (its 1px rule stays on the fill token — a redundant cue that passes 3:1) and Stat's negative delta. Each carries a test pinning the element to the text token. LogStream's derived dark mix retunes 82% → 72% to keep its documented 4.5:1 against the darker token.

A contrast assertion now covers the default pairs that fall through to `colorDefaults` (error, success, warning, and the default accent), plus the error fill against surface, body, and popover, so a future retone cannot regress silently. The regenerated CLI token doc rides along (`generate-token-docs.mjs`). All seven shipped themes retone both `--color-error` and `--color-on-error`, so none inherits the change.

## Measured

Same probe as the issue — the rendered DOM (base tokens, `getComputedStyle`):

| pair (dark) | before | after |
|---|---|---|
| white on `--color-error` (destructive button) | **3.76** | **4.70** |
| white on `--color-error` (light mode) | 4.70 | 4.70 (unchanged) |
| `--color-error` vs surface / body / popover (destructive focus ring, ≥ 3:1) | 4.37 / 5.02 / 3.87 | 3.50 / 4.02 / 3.09 |
| Chat error metadata text on body | 5.02 (fill token) | **11.06** (`--color-text-red`) |
| TextArea over-limit counter on the input surface | 4.37 (already failing) | **9.64** |
| LogStream error-level text on its surface | 5.28 | 4.86 (retuned mix) |

The dark popover ring pair sits at 3.09:1 with ~0.1 of headroom; a test pins it so a popover retone cannot cross the floor silently.

## Deliberately left as-is, for review

- **Destructive `DropdownMenuItem` label** on the dark popover: 3.87 → 3.09. It was below AA before this change; moving it to `--color-text-red` would also recolour the destructive action's identity in both modes, which feels like its own call. Happy to include it here or split it out.
- **Banner's error icon on the `--color-error-muted` wash** (2.66 vs the composited wash) and **CircularProgress's error arc vs its muted track** (2.66): the adjacent title names the status and the track is the decorative rail, so both arguably fall under the redundant-cue/rail exemptions the test file already documents. Rebasing the dark wash on the new hue only reaches 2.89–3.03, so there is no clean token-level fix.
- **`--color-error-muted` dark** still derives from the old hue (`#F5394F3F` vs light's `#E3193B33`). Text on the wash is always `--color-text-red` (FieldStatus, ChatComposer, Banner), so nothing reads on it; kept to avoid visual churn.
- **`--color-on-accent` / `--color-accent`** (3.11:1 dark) is not retoned: `--color-accent` is also the focus outline, and a dark accent that clears 4.5:1 under white lands at ~3.05:1 against the surface — the ring would sit at the 1.4.11 floor with no margin. The assertion pins the pair at its current floors (light ≥ 4.5, dark ≥ 3); deriving it inside the generator's guarantee is the durable fix, per the issue's own split with #5014.

## How to check

- `pnpm exec vitest run packages/core/src/theme/expandColorScale.test.ts` — the `colorDefaults WCAG contrast on the fall-through pairs (#5019)` block fails at 3.763 without the token change and passes with it.
- Component tests pin each migrated text element to the text token (Chat, TextArea, RichTextEditor, ChatUnreadDivider, Stat).
- Storybook → Core/Button → Destructive, theme **None (base tokens)**, mode **Dark**.

## Screenshots
5019-dark-destructive-aa


AKnassa:
Merged main in. The conflict was the RichTextEditor move into `@astryxdesign/richtext` (#4678); rename detection carried the counter change and its test across, so there was nothing to resolve by hand.

Separately, a call I would like from you before review.

Every shipped consumer of `--color-text-red` pairs it with a tinted backing (FieldStatus and ChatComposer on `--color-error-muted`, Badge on `--color-background-red`), and gothic, butter and y2k tune the token as a dark maroon for exactly that wash. The six swept components set `color` only. Measured on `--color-background-surface`, before (`--color-error`) to after (`--color-text-red`):

| theme | before | after |
| --- | --- | --- |
| gothic | 8.33 | 1.40 |
| butter dark | 9.16 | 1.40 |
| y2k dark | 11.69 | 1.22 |

The other five themes are unaffected or improve, including default dark at 3.50 to 9.64. y2k light improves 1.50 to 9.13; only its dark half regresses.

`packages/lab/src/Schedule/shared.tsx:1213` already puts `--color-text-red` on a near-bare card and measures 1.05:1 in gothic on main, so this is an existing pattern rather than a new one, but this PR would add six more sites to it.

Options as I see them: retone `--color-text-red` in those three themes, add a token for red text on untinted surfaces, or narrow the sweep and solve dark error text another way. Happy to take whichever you prefer.

gonzoblasco:
Hi @AKnassa, great writeup - the two-move split (retone the fill token + move text consumers to the text token) is exactly the right shape for this problem, because as you note the fill and text luminance windows genuinely don't overlap.

I verified the core numbers independently against WCAG 2.1 (relative luminance, 4.5:1 normal text / 3:1 non-text):

- white on `#E3193B` = **4.70:1** dark and light. Confirmed.
- `#E3193B` against a `#0a0a0a`-style surface = **4.21:1**, so the destructive focus ring clears the 3:1 non-text floor with margin. Good.

A few things I'd want checked or pinned before this lands:

**1. The DropdownMenuItem trade-off deserves a decision, not a deferral.**
Leaving the destructive `DropdownMenuItem` label on the fill token at 3.09:1 (dark popover) is defensible if the intent is "don't recolor the destructive identity", but it's the one place in this PR where a *label* (normal text) stays under 4.5:1 after the change. Since the PR already establishes that fill-as-text fails and moves other text consumers, this is now an inconsistency: every other text consumer moves to the text token, but this one stays because of color-identity. I'd suggest either moving it to `--color-text-red` too (treating the identity as carried by the icon/glyph rather than the label tint), or explicitly pinning 3.09:1 in the test as a deliberate exception with the rationale in the comment - so a future retone of the popover doesn't silently cross below it.

**2. The accent/on-accent dark pin at 3:1 is the right call, but the test comment could warn louder.**
Pinning `--color-on-accent`/`--color-accent` at the 1.4.11 floor (dark >= 3:1) rather than 4.5:1 is correct given the focus-ring role, and the rationale is well documented. The only risk: a reader skimming the test sees `toBeGreaterThanOrEqual(3)` next to a status pair and assumes it's a typo. Consider an explicit `// not a typo - see #5014` marker on that assertion, since it's the one place in this file where the floor differs from the others.

**3. LogStream retune 82% -> 72%: worth confirming the "failed to send" case.**
The 72% mix keeps 4.5:1 on the surface, but it also lowers the contrast of the *level* text itself. For error-level log lines ("failed to send" style messages), the level label is often the first thing a user scans. Not blocking, just a sanity check.

**4. Small thing: the injected-CSS test helper is duplicated across 5 test files.**
The `injectedCss()` + `escapeRegExp()` helpers appear verbatim in ChatMessageMetadata, ChatToolCalls, TextArea, ChatUnreadDivider, and Stat tests. It's a clean approach, but 5 copies is a maintenance smell - if the StyleX rule format ever changes, all five break independently. Worth hoisting to a shared test util (e.g. `__tests__/assertToken.ts`) so the token-pinning assertions have a single home.

Overall: this is a strong, well-measured fix. The token-vs-text separation is the correct model and the regression assertions are the right guard. My main ask is to turn the DropdownMenuItem deferral into an explicit decision with a pinned test, and to de-duplicate the CSS-scan helper.

cixzhang:
Thank you for the measured dark-mode contrast work. Color-token and palette changes are now coordinated as owner-led accessibility work rather than parallel implementation PRs. Please follow #4475 for the program direction; we are closing this attempt.

PR #5026, opened 2026-08-14
fix(core): add borderDefaults to CoreTokenName and tokenDefaults

### LINKED ISSUE #5017: `--border-width` is documented and emitted but not in the defineTheme token type
`astryx docs tokens` — the reference we point theme authors at — lists

```
--border-width | 1px
```

`astryx theme build` accepts it and emits it, and the var is real (`borderDefaults` → `stylex.defineVars`). But `defineTheme` rejects it in TypeScript:

```
error TS2353: Object literal may only specify known properties, and '--border-width'
does not exist in type 'Partial<Record<TokenName, TokenValue>>'.
```

`CoreTokenName` unions eleven `*Defaults` groups and omits `borderDefaults`; `tokenDefaults` (the flat map) omits it too. So the author gets a red squiggle on a token the docs advertise, the build says nothing, and the CSS works at runtime. Anyone who trusts the editor deletes a legitimate override; anyone who trusts the build ships a theme its own types reject.

The flat-map omission has a second consequence: `resolveThemeTokens()` / `useTheme()` never return `--border-width`, so JS consumers (canvas, charts) cannot read a value that CSS consumers can.

Found while a theme author set `--border-width: 2px` for a heavier arcade outline — it rendered correctly and typechecked as an error.

Fix looks like adding `keyof typeof borderDefaults` to `CoreTokenName` and `...borderDefaults` to `tokenDefaults`. Worth a check for other `*Defaults` groups in the same position.

Geervan:
Fixes #5017

### Problem
`astryx docs tokens` documents `--border-width` and `astryx theme build` emits it, but `defineTheme()` rejected it in TypeScript with:
`error TS2353: Object literal may only specify known properties, and '--border-width' does not exist in type 'Partial<Record<TokenName, TokenValue>>'`.

Additionally, `tokenDefaults` omitted `borderDefaults`, causing `resolveThemeTokens()` and `useTheme()` to exclude `--border-width` from default token resolutions for JS consumers (e.g. canvas/charts).

### Fix
- Added `borderDefaults` to `CoreTokenName` union in `defineTheme.ts` to fix the type check.
- Merged `...borderDefaults` into `tokenDefaults` flat-map for JS token resolution.
- Added test coverage in `tokens.test.ts` and `useTheme.test.tsx`.

### Testing
- `npx vitest run packages/core/src/theme/tokens.test.ts packages/core/src/theme/useTheme.test.tsx` (41/41 tests passed).


cixzhang (approved):

cixzhang:
Thank you!

PR #5029, opened 2026-08-14
fix(theme): support CustomTextTypes module augmentation (#5013)

### LINKED ISSUE #5013: theme build does not generate the CustomTextTypes augmentation it documents
`theme/types.ts` promises this:

```ts
/**
 * declare module '@astryxdesign/core/theme' {
 *   interface CustomTextTypes { hero: true; caption: true }
 * }
 *
 * `astryx theme build` generates these augmentations automatically when it
 * detects new `type:*` values in a theme's component overrides.
 */
```

It does not. A theme with

```ts
components: {text: {'type:score': {fontVariantNumeric: 'tabular-nums'}}}
```

gets the CSS but no `.variants.d.ts` entry, so `<Text type="score" />` does not typecheck. `variant:*` values do generate augmentations in the same build (verified side by side: a theme with a custom Button variant and a custom Text type emitted only the Button one).

Why: `generateVariantDeclarationsAsync()` derives the target interface name from the component and prop — `text` + `type` → `TextTypeMap` in `@astryxdesign/core/Text` — and skips when that interface does not exist. For text types the extension point is a differently-named interface in a different module: `CustomTextTypes` in `@astryxdesign/core/theme`.

The manual recipe in the docstring does work — I checked it against the real `dist` types:

```ts
import type {TextType} from '@astryxdesign/core/theme';
declare module '@astryxdesign/core/theme' {
  interface CustomTextTypes { hero: true }
}
const a: TextType = 'hero';   // errors without the augmentation, typechecks with it
```

So the fix is a mapping in the generator from `text` + `type` to that module/interface pair, rather than a new extension point.

Worth checking whether any other prop is in the same position (a documented custom-value axis whose extension point does not follow the `<Component><Prop>Map` convention).

jibin7jose:
Fixes #5013

### Description
This PR addresses an issue where the `astryx theme build` command was failing to generate the `CustomTextTypes` augmentation for custom `Text` variants.

Previously, `generateVariantDeclarationsAsync()` expected the extension point for `text` + `type` to follow the standard `<Component><Prop>Map` convention (e.g., `TextTypeMap`). Because the actual extension point is `CustomTextTypes` in the `@astryxdesign/core/theme` module, the generator skipped the property and failed to emit the correct `.variants.d.ts` entry.

This update introduces a specific check for the `Text` `type` property in the CLI, explicitly mapping it to the `CustomTextTypes` target in the `theme` module so it correctly generates the type augmentations.

### Changes
- Updated `generateVariantDeclarationsAsync` in `packages/cli/api/theme/build/build.mjs` to special-case `component === 'text'` and `prop === 'type'`.
- Ensured the generated types map to the `CustomTextTypes` interface inside the `@astryxdesign/core/theme` module.


cixzhang (changes requested):
Thanks @jibin7jose — the diagnosis in [#5013](https://github.com/facebook/astryx/issues/5013) is right, and mapping `text.type` at the emitter is the correct shape of fix.

Three things before it lands:

1. **`build.mjs:534` skips `componentHasAugmentableInterface`** — the one guard the file exists to enforce (`:456-461`). `CustomTextTypes` is declared in `types.ts` and only re-exported from `src/theme/index.ts:166`, so whether the augmentation merges or quietly creates a dead sibling interface depends on what dist bundles. Please route the special case through an override table (`text.type → {module: 'theme', interface: 'CustomTextTypes'}`) that still runs the existence check, so a rename upstream fails loudly instead of emitting something inert.

2. **A test that proves the augmentation actually widens `TextType`.** That is the whole fix, and nothing currently verifies it — assert on the emitted `.variants.d.ts`, and ideally a type-level check that a custom type is accepted where `TextType` is expected. Without it we can't tell a working augmentation from a dead sibling interface.

3. **Missing changeset** (`X20`) — a CLI fix with no bump ships silently.

One more if you're willing: [#5013](https://github.com/facebook/astryx/issues/5013) asks whether other props sit in the same position. The override table would answer that generally rather than one prop at a time.

packages/cli/api/theme/build/build.mjs
```diff
@@ -312,18 +312,28 @@ function readComponentDeclarations(pascalName) {
   let contents = '';
   const coreRoot = resolveCoreRoot();
   if (coreRoot) {
-    const candidates = [
-      path.join(coreRoot, 'dist', pascalName, 'index.d.ts'),
-      path.join(coreRoot, 'src', pascalName, 'index.ts'),
+    const candidateSets = [
+      [
+        path.join(coreRoot, 'dist', pascalName, 'index.d.ts'),
+        path.join(coreRoot, 'dist', pascalName, 'types.d.ts'),
```
cixzhang: Hmm, this reads `types.d.ts` for every component, not just the override. Probably want it gated.

packages/cli/clients/cli/commands/build-theme.variants.test.mjs
```diff
@@ -255,7 +285,7 @@ describe('theme build custom-variant augmentations', () => {
     );
 
     try {
-      execFileSync('pnpm', ['exec', 'tsc', '--project', 'tsconfig.json', '--noEmit'], {
+      execSync('pnpm exec tsc --project tsconfig.json --noEmit', {
```
cixzhang: `ensure-core-built.mjs:49` next door uses `execFileSync('pnpm', [...])`. Might keep that shape.

cixzhang (changes requested):
Thanks @jibin7jose — the three from last round are all done, and I checked the augmentation really does widen `TextType`.

One thing before it lands. Reading `types.d.ts` for every component, not just the overridden one, opens augmentation points the generator was deliberately skipping. A theme with `indicator: {'family:brand': …}` now emits `interface IndicatorFamilyMap { 'brand': true }` — main emitted nothing — and that map's values are states (`'unchecked' | 'checked'`), so `IndicatorState<'brand'>` silently becomes `true`.

Ideally I'd scope the extra read to the `AUGMENTATION_OVERRIDES` lookup, so `componentHasAugmentableInterface` gives the same answer as today everywhere else.

You'll also want a merge with main; the branch is a few hundred commits behind.

Does scoping it to the override work, or is there a text case that needs the wider read?

[Reviewed by Robohands]

cixzhang:
Checking in on theme augmentation: the extra type read still needs scoping so non-text state maps cannot widen to `true`, plus a main refresh. Would you like to continue, have us carry it, or close this for now?

jibin7jose:
I’ll check it and continue with the changes.

cixzhang (changes requested):
Thanks @jibin7jose — the scoped override and type-level regression are correct, and the bug still reproduces on current main: theme build emits the custom Text CSS but no declaration that widens `TextType`.

This exact head cannot merge now because both the generator and its variant test conflict with current main's shared component-value classifier. Please refresh from main and carry the `CustomTextTypes` mapping into that current path, preserving these outcomes:

- `text.type` emits an augmentation for `CustomTextTypes` in `@astryxdesign/core/theme`;
- importing the generated declaration makes the custom value type-check on `<Text type>`;
- unrelated and closed axes remain unchanged.

The draft generated-contract work in #6061 overlaps this area but has not landed, so it does not yet replace this fix. If it lands first, reroll onto that path instead.

[Reviewed by Robohands]

PR #5034, opened 2026-08-14
fix(core): Avatar: report a status label through context, and warn on an unnamed interactive avatar

cixzhang:
Closes two of the three BLOCKs the Night Watch audit of `core/Avatar` left open on [#5030](https://github.com/facebook/astryx/pull/5030). A10 target size is not addressed here and stays open.

Rebased onto current `main` (after [#5055](https://github.com/facebook/astryx/pull/5055) landed) and reworked to what review asked for: **no breaking change, no new prop, and the status label carried by a ref instead of state.**

## Not a breaking change

`AvatarProps` is a plain interface again. The props union and `AvatarBaseProps` are gone, so an unnamed interactive avatar compiles exactly as it does today, and the `@ts-expect-error` tests that asserted the union are gone with it. The `statusLabel` prop is gone too. The changeset is a `patch`.

The runtime guard is the whole enforcement: `useDevWarning` fires once per mount when an avatar with `href` or `onClick` resolves no name from `alt`, `name`, `aria-label` or `aria-labelledby`.

Two improvements from the union work are not breaking, so they stay:

- a status label no longer counts as an interactive avatar's identity. `<Avatar href status={<AvatarStatusDot label="Online" />} />` used to satisfy the warning while shipping a link whose whole name was "Online".
- derived `role`/`aria-label`/`aria-hidden` spread **before** `{...props}` on all three roots, following Icon's `getIconA11yProps`, so a consumer's own aria values win. They were spread last, which clobbered them.

## A ref through context, and the label written directly

`AvatarStatusLabelContext` now carries the avatar's own ref. `AvatarStatusDot` writes its label into it from the callback ref it already had, and the avatar writes the composed accessible name straight onto its root element during the commit phase, before paint. No state, no second render, and Avatar still has zero Effects.

Ordering falls out of two facts. Child refs attach before parent refs, so on mount the dot writes its label and the avatar's root ref then reads it. The root's callback ref is fresh every render, so on any avatar re-render it reattaches and recomposes the name after React has written that render's props. For the remaining direction, a label that changes while the avatar itself does not re-render (state inside a consumer's wrapper), the dot calls the update the avatar left on the ref. A directly passed dot still resolves during the first render through `status.props.label`, so the common case, and its server HTML, are unchanged.

One thing worth flagging: react-compiler's lint rule treats a value read from context as immutable, so the dot's write carries a scoped `eslint-disable` with the reason. The ref it writes into is the avatar's own `useRef`, shared deliberately.

## Verification

Accessible names read from the real browser, Chromium via CDP `Accessibility.getFullAXTree` against the static Storybook build:

| case | AX role | AX name |
| --- | --- | --- |
| dot passed directly | image | `Ada Lovelace, Online` |
| dot inside a consumer's wrapper | image | `Grace Hopper, Online` |
| label changed after mount (avatar re-renders) | image | `Katherine Johnson, Busy` |
| label changed after mount (only the dot re-renders) | image | `Mary Jackson, Busy` |
| status unmounted | image | `Dorothy Vaughan` |
| decorative avatar, wrapped status | image | `Online` |
| interactive link, wrapped status | link | `Alan Turing, Away` |
| consumer `aria-label` | image | `Annie Easley, rocket scientist` |

SSR and hydration: refs do not run on the server, so the server HTML for a wrapped status carries `aria-label="Grace Hopper"` and the status joins the name on the client, in the commit phase after hydration. A directly passed dot server-renders with the status already in the name (`aria-label="Ada Lovelace, Online"`). Hydrating the wrapped case produced no console errors or warnings. This is the same property the previous revision had.

Cost, pinned by test: a wrapped status now costs one render, not two. The old test read two because it counted the tooltip's own mount commit alongside the report; with `tooltip={false}` it reads 1 after this change and 2 before it.

Gates on the rebased branch: `vitest run packages/core/src/Avatar packages/core/src/AvatarGroup` 119 passed; `pnpm build` clean; `pnpm lint:strict` clean; `pnpm test` 11298 passed with 16 pre-existing failures on this machine (DateInput touch, TransferList, two case-insensitive-filesystem CLI tests), all of which reproduce with this branch's changes reverted.

Nothing visual changes, so there are no screenshots.

### DIFF
--- a/.changeset/avatar-interactive-name-status-label.md
+++ b/.changeset/avatar-interactive-name-status-label.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Avatar: a status element now reports its own accessible label to the avatar through context, so wrapping `AvatarStatusDot` in a component of your own keeps the status in the avatar's accessible name ("Jane Doe, Online") instead of silently dropping it — the `role="img"` root prunes descendant semantics, so composing it in is the only route to assistive tech (WCAG 4.1.2). Reading `label` off a directly-passed element still works and still resolves on the first render. An interactive avatar (`href`/`onClick`) with no `name`/`alt` warns in development, and a status label no longer counts as the control's identity: "Online" reads as a legitimate name while saying nothing about where the link goes. Derived `role`/`aria-label`/`aria-hidden` now spread before the passthrough props, following Icon, so a consumer's own values win. No API change (#5034)
+
+@cixzhang
--- a/apps/storybook/stories/Avatar.stories.tsx
+++ b/apps/storybook/stories/Avatar.stories.tsx
@@ -264,6 +264,41 @@ export const WithStatus: Story = {
   ),
 };
 
+export const StatusLabelSources: Story = {
+  render: () => {
+    // A consumer's own wrapper. AvatarStatusDot reports its label to the
+    // avatar through context, so the status still reaches the accessible name.
+    function PresenceDot({presence}: {presence: string}) {
+      return <AvatarStatusDot variant="success" label={presence} />;
+    }
+
+    return (
+      <div {...stylex.props(styles.storyWrapper)}>
+        <h4 {...stylex.props(styles.heading)}>
+          Every route to a status in the accessible name
+        </h4>
+        <div {...stylex.props(styles.row)}>
+          <Avatar
+            name="Ada Lovelace"
+            size="xl"
+            status={<AvatarStatusDot variant="success" label="Online" />}
+          />
+          <Avatar
+            name="Grace Hopper"
+            size="xl"
+            status={<PresenceDot presence="Online" />}
+          />
+          <Avatar
+            name="Katherine Johnson"
+            size="xl"
+            status={<AvatarStatusDot variant="neutral" label="On leave" />}
+          />
+        </div>
+      </div>
+    );
+  },
+};
+
 export const StatusAcrossAllSizes: Story = {
   name: 'Status Dot Across All Sizes',
   render: () => (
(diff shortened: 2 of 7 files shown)


cixzhang (commented):
This status label behavior is a bit much to inject into the avatar. Why not just a straightforward ref passed through context modifying the label?

Also avoid this being a breaking change. Add the warning

kentonquatman (approved):

cixzhang:
Thanks, both points are addressed.

**The ref.** `AvatarStatusLabelContext` now hands down the avatar's own ref instead of a setter. `AvatarStatusDot` writes its label into it from the callback ref it already had, and the avatar writes the composed name straight onto its root element in the commit phase, so nothing goes through state and the wrapped case costs one render instead of two. A dot passed directly still resolves during the first render from `status.props.label`, so the common case and its server HTML are unchanged. I checked the accessible name in Chromium through CDP `Accessibility.getFullAXTree` for the direct dot, a dot inside a consumer's wrapper, a label changed after mount (both when the avatar re-renders and when only the dot does), and an unmounted status; the table is in the description. One honest wrinkle: react-compiler's lint rule treats a context value as immutable, so the dot's write carries a scoped `eslint-disable` with the reason.

**Not breaking.** The props union and `AvatarBaseProps` are gone, so `AvatarProps` is a plain interface and an unnamed interactive avatar compiles as before; the `@ts-expect-error` tests that asserted the union went with it, the `statusLabel` prop is gone, and the changeset is back to a `patch`. `useDevWarning` is the whole enforcement now, firing once per mount when an avatar with `href` or `onClick` has no `alt`, `name`, `aria-label` or `aria-labelledby`. Two non-breaking pieces of that work stayed: a status label no longer counts as a control's identity, and the derived `role`/`aria-label`/`aria-hidden` spread before `{...props}` so a consumer's own values win.

PR #5035, opened 2026-08-14
fix(SegmentedControl): clamp item labels to a single line

ernestt:
## Summary

`SegmentedControlItem` labels are constrained to a single line. Previously the label rendered as a bare `<span>{label}</span>` with no wrapping control, so a long label could wrap to two or more lines and break the control's height and alignment — inconsistent with `Tab` and `Button`, both of which already keep their labels on one line.

## Changes

- `SegmentedControlItem.tsx`: add `whiteSpace: 'nowrap'` to the item base style, and wrap the visible label in a `labelText` span that truncates overflow with an ellipsis (`overflow: hidden; text-overflow: ellipsis; min-width: 0`) — the same pattern `Button` uses.
- Added a test asserting labels are clamped to a single line with ellipsis truncation.
- Changeset (patch).

## Before / After

- Before: a long label wraps onto multiple lines, growing the segment's height.
- After: the label stays on one line and truncates with `…` when it exceeds the available width.

### DIFF
--- a/packages/core/src/SegmentedControl/SegmentedControlItem.tsx
+++ b/packages/core/src/SegmentedControl/SegmentedControlItem.tsx
@@ -126,6 +126,7 @@ const styles = stylex.create({
   },
   fill: {
     flex: 1,
+    minWidth: 0,
     justifyContent: 'center',
   },
   icon: {
--- a/packages/core/src/SegmentedControl/SegmentedControlItem.tsx
+++ b/packages/core/src/SegmentedControl/SegmentedControlItem.tsx
@@ -83,6 +83,7 @@ const styles = stylex.create({
     fontWeight: fontWeightVars['--font-weight-medium'],
     color: colorVars['--color-text-secondary'],
     cursor: 'pointer',
+    whiteSpace: 'nowrap',
     transitionProperty: 'color, background-color, box-shadow',
     transitionDuration: durationVars['--duration-fast'],
     transitionTimingFunction: easeVars['--ease-standard'],
@@ -132,6 +133,11 @@ const styles = stylex.create({
     alignItems: 'center',
     justifyContent: 'center',
     flexShrink: 0,
+  },
+  labelText: {
+    overflow: 'hidden',
+    textOverflow: 'ellipsis',
+    minWidth: 0,
   },
 });
 
@@ -244,7 +250,9 @@ export function SegmentedControlItem({
         ),
       )}>
       {iconElement}
-      {!isLabelHidden && <span>{label}</span>}
+      {!isLabelHidden && (
+        <span {...stylex.props(styles.labelText)}>{label}</span>
+      )}
     </button>
   );
 }
--- a/.changeset/segmented-control-single-line-label.md
+++ b/.changeset/segmented-control-single-line-label.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] SegmentedControl: keep item labels on a single line, truncating overflow with an ellipsis instead of wrapping to multiple lines
+@ernestt
--- a/packages/core/src/SegmentedControl/SegmentedControl.test.tsx
+++ b/packages/core/src/SegmentedControl/SegmentedControl.test.tsx
@@ -199,25 +199,35 @@ describe('SegmentedControl', () => {
     expect(screen.queryByText('Grid view')).not.toBeInTheDocument();
   });
 
-  it('clamps labels to a single line with ellipsis truncation', () => {
+  it('fill items can shrink and truncate long labels', () => {
     render(
-      <SegmentedControl value="grid" onChange={() => {}} label="View mode">
+      <SegmentedControl
+        value="grid"
+        onChange={() => {}}
+        label="View mode"
+        layout="fill">
         <SegmentedControlItem
           value="grid"
-          label="A very long segment label that should not wrap"
+          label="A very long segment label that should truncate"
         />
+        <SegmentedControlItem value="list" label="List" />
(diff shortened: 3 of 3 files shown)


cixzhang (commented):
Looks good, thanks! We could use a new story or screenshots to demonstrate this

cixzhang (approved):

packages/core/src/SegmentedControl/SegmentedControlItem.tsx
```diff
@@ -83,6 +83,7 @@ const styles = stylex.create({
     fontWeight: fontWeightVars['--font-weight-medium'],
     color: colorVars['--color-text-secondary'],
     cursor: 'pointer',
+    whiteSpace: 'nowrap',
```
cixzhang, line 86: Hmm if we have a segmented control and there's one really long item do we want to let it take up all the space and y other options aren't visible or let it take up to some maximum so other items can still work? Perhaps this is just width configuration

cixzhang (changes requested):
Went and looked at this at 320px — the ellipsis isn't firing, and the single-line change on its own makes the narrow case worse: the long label now pushes List and Table right out of the container.

| | 320px, `fill` |
|---|---|
| **Before** | before |
| **After** | after |
| **With `minWidth: 0`** | fix |

The item is a flex child with `min-width: auto`, so it can't shrink and `text-overflow` never gets a chance:

```js
fill: {
  flex: 1,
  minWidth: 0,
  justifyContent: 'center',
},
```

Same in RTL and at 2x text zoom. Sorry to flip on the approval — this is the answer to my own question above: capping the item is what keeps the other options reachable.

One more: the new test passes on `main` without the fix — `white-space: nowrap` is already in the injected sheet.

cixzhang:
Layout matrix for this change — 11 cases, real Chromium, before / after / probe. Nothing ellipses anywhere on the current head: `text-overflow: ellipsis` never fires once.

**Containment**

| Assumption | Relies on | Holds |
|---|---|---|
| Which box shrinks | the `<button>` flex child | **no** — only the inner `<span>` is constrained |
| Min-size defaults | flex children default `min-width: auto` | `styles.fill` has no `minWidth`, so the button never shrinks |
| Width source | fixed vs auto-sizing agree | **no** — an auto-width parent hides the bug entirely |
| Ancestors | none above matter | **no** — a `flex: 1` ancestor re-breaks it even with the fix |

**Cases**

| # | Case | Before | After | Probe |
|---|---|---|---|---|
| 1 | short labels, 600px | identical | identical | — |
| 2 | long label, `hug`, wide | identical | identical | — |
| 3 | long label, `hug`, 320 | wraps, fits | **394px, "Table" off-edge** | no help |
| 4 | long label, `fill`, 320 | 4 clipped lines | 392px, "Table" gone | `Grid view …`, fits |
| 5 | one long + two short | wraps | long one takes 273px, siblings pushed out | 3×104px, ellipsed |
| 6 | fixed / auto / flex / grid parent | — | auto hides it; grid track blown 200→392 | flex `:1` still 394 |
| 7 | ancestor `overflow: hidden` | — | **segment silently disappears** | contained |
| 8 | RTL | mirrors | mirrors, off left edge | ellipsis on logical end |
| 9 | 200% text zoom | clipped | 663px in 320 | contained |
| 10 | forced colors | same as 4 | same as 4 | correct |
| 11 | icon-only / icon+label | identical | icon-only untouched | icon holds, text ellipses |

**Case 5, the one that answers "does the long item starve the others":**

| After | Probe |
|---|---|
|  |  |

**Probe** — one line, `SegmentedControlItem.tsx`:

```diff
   fill: {
     flex: 1,
     justifyContent: 'center',
+    minWidth: 0,
   },
```

Not fixed by it: `hug` (case 3, no shrink axis — now worse than main, which wrapped and fitted) and flex ancestors (case 6c, needs `min-width: 0` on the consumer's wrapper).

**The added test** substring-matches `getAllInjectedCss()`, so it does not assert the rules apply to this element, and jsdom has no layout. It passes while nothing truncates in any browser.

One-line check without any harness:

```js
const s = document.querySelector('[role=radiogroup] button span:last-of-type');
s.scrollWidth > s.clientWidth   // false on head, true with minWidth:0
```

All 33 frames, metrics, the Storybook harness and the shoot script: [full evidence](https://github.com/cixzhang/astryx/blob/assets/pr-5035/pr-assets/EVIDENCE.md). Only neutral theme, light, `size="md"`; no `sm`/`lg`, no real-device HCM.

<sub>Layout evidence generated for review — not a review.</sub>

[Reviewed by Robohands]

packages/core/src/SegmentedControl/SegmentedControlItem.tsx
ernestt: Good question — and you answered it yourself in the follow-up review. Adding `minWidth: 0` to `fill` items caps each segment so the others stay reachable. In `hug` mode items still size to content (no shrink axis), which is the expected behavior there.

ernestt (commented):
Thanks for the thorough layout matrix — that was exactly the evidence needed.

Fixed in the latest push:

1. **`minWidth: 0` on `fill` items** — flex children can now shrink below content size, so `text-overflow: ellipsis` actually fires.
2. **Rewrote the test** — now verifies element-level computed styles (`minWidth` on the button, `overflow`/`textOverflow` on the label span) in a `layout="fill"" context, rather than substring-matching the global injected sheet. Confirmed the new test fails without the `minWidth: 0` fix.

The `hug` case (case 3) is intentionally left alone — in hug mode there's no shrink axis, so the items size to content. That's a separate concern if we want to address it later.

cixzhang (approved):

PR #5039, opened 2026-08-14
fix(layer): host layers outside the ancestors that break them

cixzhang:
## Summary

Fix context layers—especially rich HoverCards—rendered inside HTML contexts such as `<p>`, headings, links, buttons, and structurally restricted containers.

Previously, block content could enter invalid server markup, be reparented by the HTML parser, and leave an empty or broken popover after hydration.

## Approach

Context-mode `useLayer` now renders an inert `<template>` marker at the layer’s actual JSX position. The marker is identical during SSR and the first client render.

After mounting:

- Safe positions render the popover inline beside the marker.
- Unsafe positions portal it to the nearest safe ancestor outside the complete unsafe chain.
- The marker remains mounted, without an ID, so the host can be recalculated if the render call moves.
- Open layers are reopened exactly once when relocation replaces their DOM element.
- Detached transitional elements are never passed to `showPopover()`.

The new `lazyMount` option defers host discovery and content mounting until `show()`. HoverCard uses it, so a closed card renders only the inert marker and unmounts its content again after hiding.

## Theme, direction, and focus

Corrective portals stay near the intended render position rather than using `document.body`.

- CSS custom properties continue inheriting live from the nearby safe host.
- `direction` and `writing-mode` are copied only when the unsafe ancestor chain changes them; otherwise they keep inheriting live.
- `showPopover()` receives the trigger as its invoker `source`.
- Existing ButtonGroup and InputGroup structural selectors ignore both popovers and inert layer markers.

A block context is still recommended for HoverCard. Paragraph, heading, and link placements are supported when necessary, but corrective portaling can change DOM and tab order.

## Scope

- Applies to context-mode `useLayer` consumers.
- Adds the public `lazyMount?: boolean` context-layer option.
- Fixed-position layers are unchanged.
- `LayerProvider` behavior is unchanged.
- Rich HoverCard accessibility relationships are tracked separately in #5049.

## Story

`Core/Layer` → **Inline trigger hosting** demonstrates a rich card triggered inside a paragraph, including its resolved host, typography, focus order, and scrolling behavior.

## Validation

- All GitHub checks pass.
- 559 affected-consumer tests pass.
- Full repository test suite passes in CI.
- Core and Storybook production builds pass.
- Typecheck, docs typecheck, lint, formatting, repository checks, accessibility, RTL, smoke, sandbox, and Vercel checks pass.


imdreamrunner:
Could we adjust the hosting approach in two ways?

1. **Resolve the location lazily when showing.** Instead of resolving and mounting a portal after hydration for every context layer, `show()` could record the open intent, resolve the current location, mount the layer, then replay `showPopover()` after its ref attaches. Hiding would unmount it and clear the resolved location. This avoids work for layers that never open and prevents a stored host from becoming stale if the trigger moves.

2. **Use an empty `<template>` sentinel at the `layer.render()` call site.** Trigger ancestry gives us a safe portal target, but it cannot tell us whether the intended inline location was already safe. An inert template rendered identically during SSR and hydration would expose the actual `parentElement` without mounting consumer content or affecting layout. When showing:
   - If the sentinel’s parent is safe, replace it with the layer inline.
   - If unsafe, portal the layer to the nearest safe ancestor.

This would preserve the existing DOM position, cascade, and natural tab order when possible, while using a portal only as a corrective fallback. The final layer should still honor `as`, and `showPopover({source: trigger})` remains a useful backstop when portaling.

I’d add focused coverage for React SSR/hydration with `<template>`, safe inline placement, paragraph/link/table contexts, controlled/default-open behavior, and rapid show/hide before the layer commits.

Issue #2706, opened 2026-06-11
[Bug] [bug-bash] Mobile Nav: Properties tab preview is empty on load (looks broken until a prop is changed)


cvkxx:
### Description

On the **Properties** tab, the Preview area is **empty** on load — no component is visible until a prop is changed in the controls below. This isn't technically a bug (Mobile Nav defaults to closed, so nothing renders until `isOpen` is toggled), but the empty preview reads like a broken/missing component to the end user.

### Reproduction

1. Open https://xds-sandbox.vercel.app/components/MobileNav?tab=properties
2. The Preview area is empty
3. Change a prop below (e.g. toggle `isOpen`) — the component then appears

### Expected

The Properties tab should show something in the preview on load so it doesn't look broken — e.g. default `isOpen` to a visible state, render in a contained/inline preview, or show a hint that the drawer is closed by default.

jiunshinn:
@cixzhang — this is clustered under C2/C11 in #2877, so one question before I open a PR.

One detail beyond the original report: toggling `isOpen` doesn't just reveal the drawer — it renders as a **full-viewport modal over the entire docsite** (`dialog.showModal()`, `MobileNav.tsx:345`), not inside the preview box.

That's the crux for the C2 fix here. Because `showModal()` escapes to the top layer and `MobileNav` has no `isInline` mode (unlike Dialog/AlertDialog/CommandPalette, which set `isInline: true` in their playground), the usual "add `playground.defaults.isOpen: true`" can't be contained in the preview — so this needs C11-style container handling, not just defaults.

Which direction do you prefer?

1. **Docsite harness** — render the interactive preview in a contained/scaled surface so any `showModal` overlay is visually contained (closes the cluster, no component API change).
2. **Per-component playground config** — extend `playground` to express contained/wrapper rendering for overlays (aligns with #2008 + C11).
3. **`isInline` mode on MobileNav**, mirroring Dialog (needs the spec protocol per the AI-contribution guide).

Happy to implement whichever fits your plan for the cluster.

jiunshinn:
Opened a draft PR with the `playground.overlay` approach (C11's "extend playground config" candidate, no component API changes): #3616 — happy to rework toward a contained preview surface or `isInline` instead if that fits the cluster plan better.

PR #5042, opened 2026-08-14
fix(docsite): give MobileNavToggle preview a mobile AppShell (#4983)

### LINKED ISSUE #4983: [Bug] MobileNavToggle: Properties tab preview is empty on load
### Description

On the **Properties** tab, the Preview area is **empty on initial load** no component is visible until a prop is changed in the controls below.

This makes the preview look broken or like the component failed to render. The preview should provide some visible representation of the `MobileNavToggle` component when the Properties tab is opened.  

familiar closed issue: https://github.com/facebook/astryx/issues/2706

### Reproduction

1. Open https://astryx.atmeta.com/components/MobileNavToggle?tab=properties
2. Observe the Preview area immediately after the page loads
3. The Preview area is empty
4. Change one of the component props in the controls below
5. The component then becomes visible in the Preview

### Astryx Version

v0.3.0

### Environment

_No response_

AKnassa:
Fixes #4983

## What this does

The Properties tab for Mobile Nav Toggle showed an empty preview box that looked broken. Now it shows the hamburger button, with a short note explaining that in a real app the button only appears below the mobile breakpoint.

## Why

The toggle only renders inside an app shell that reports a mobile viewport. The docs preview has no app shell, so the component rendered nothing. Changing props does not reveal it either; the empty stage was unconditional.

## What changed

- New `playground.appShellMobile` option for component docs: the preview wraps the component in a simulated mobile app-shell context. Same approach as `playground.overlay` from #2706.
- Mobile Nav Toggle's docs opt into it, so the hamburger renders and clicking it flips `aria-expanded`.
- A caption under the preview explains the simulation.
- Tests: registry wiring in data-extraction, unit tests for the simulated context value. Docsite suite 375/375.

## How to see it

Run the docs site and open `/components/MobileNavToggle?tab=properties`. Before: empty box. After: hamburger button plus caption; clicking it toggles the expanded state.
pr5042-after-hamburger-preview
pr5042-before-empty-preview

### DIFF
--- a/.changeset/mobilenavtoggle-preview-simulated-mobile-appshell--k9pt.md
+++ b/.changeset/mobilenavtoggle-preview-simulated-mobile-appshell--k9pt.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/cli': patch
+---
+
+[docs] MobileNavToggle preview simulates a mobile AppShell instead of an empty stage: new playground.appShellMobile for components that render nothing without AppShell mobile context (#4983)
+@AKnassa
--- a/apps/docsite/scripts/generate-data.mjs
+++ b/apps/docsite/scripts/generate-data.mjs
@@ -826,6 +826,7 @@ export interface ElementDescriptor {
 export interface PlaygroundConfig {
   defaults?: Record<string, unknown>;
   overlay?: boolean;
+  appShellMobile?: boolean;
   wrapper?: {
     component: string;
     props?: Record<string, unknown>;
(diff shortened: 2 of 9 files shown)


cixzhang (approved):

cixzhang (approved):
Thanks @AKnassa — giving the preview a real mobile AppShell is the right fix: the component was returning `null` above the breakpoint because the context it needs was never there, so a wrapper is the honest answer rather than forcing the component to render. Nice that the preview-state plumbing is generic (`playground.wrapper`) instead of special-casing this one component.

Approving for the release cut. I have a draft of my own ([#5053](https://github.com/facebook/astryx/pull/5053)) that overlaps six of these files; yours actually renders the component, so I'll close mine and re-land only the leftover piece — a note on the stage when a preview legitimately renders nothing, which also covers `CheckIndicator`, `ChatMessageMetadata` and `SideNavCollapseButton`.

PR #5045, opened 2026-08-15
feat(cli): warn when a theme names fonts it does not load

### LINKED ISSUE #5015: Nothing helps a theme author load the font they just named
Every place we discuss fonts says the same true thing — "font loading is the consumer's responsibility" (`TypographyConfig` docstring, `astryx docs typography`, `astryx docs theme`) — and then nothing helps you do it.

Six themes written independently against those docs: four named a webfont (Bungee, Orbitron, Fraunces, Source Serif 4, Space Grotesk, JetBrains Mono). **Zero** emitted a `<link>`, an `@font-face`, or so much as a note to the reader that one was needed. Every one of those themes renders in the wrong typeface on a clean machine, and nothing in the pipeline says so: `theme build` succeeds, the CSS is correct, the token is set, and the browser quietly falls back.

A statement of whose job it is does not get the job done. Concretely:

1. **Warn at build.** `theme build` sees `typography.body.family` and every `fontFamily` in `components`. When a family is not a generic or a known system stack, note that the theme does not load it and print the `<link>`/`@font-face` to add. It already prints install instructions; this is the same slot.
2. **Give the recipe in the docs**, not just the disclaimer — the Google Fonts `<link>` pair and the self-hosted `@font-face`, with `font-display: swap` and a real fallback stack.
3. Decide about `url`. Two places already advertise it — `astryx docs theme`'s defineTheme table lists the role's parameters as `family, fallbacks?, url?, weight?`, and the `TypographyRole` docstring says heading "inherits family/fallbacks/url from body" — but there is no `url` field on the interface and nothing reads one. Either wire it up (emit the `@font-face` into the built CSS, which would make (1) unnecessary) or delete both mentions, because today it reads like a feature that exists.

(3) is a design call; (1) and (2) I would do regardless.

AKnassa:
Closes #5015

Heads up @cixzhang: saw you self-assigned this shortly after filing, so this may cross work you already have in flight. No hard feelings at all if you'd rather land your own version; happy to close this one or rebase onto yours. Sharing since it was ready.

## What this does

`astryx theme build` now tells a theme author when their theme names a font that nothing actually loads, and prints the exact `<link>` and `@font-face` snippet to add. The docs also gain the loading recipe the disclaimers always pointed at.

## Why

A theme can name any webfont, the build succeeds, the CSS is correct, and the browser quietly falls back on every clean machine. Nothing in the pipeline said so. This covers parts (1) and (2) of the issue; part (3), the `url` field question, is a design call and is left untouched here.

## What changed

- New check in `theme build`: resolved `--font-family-*` tokens and component-level `fontFamily` values are compared against CSS generics and known preinstalled system families (including the shipped themes' deliberate fallback stacks, so none of the seven themes false-warns). Anything unrecognized is treated as a webfont.
- One warning per font goes to the receipt (`--json` sees it) and stderr; the copy-pasteable fix prints after the install instructions, in the slot that held the dead `resolvedTheme.fonts` block from before #2226.
- `astryx docs typography` gains a "Loading Custom Fonts" section (Google Fonts and self-hosted recipes, `font-display: swap`, real fallback stacks); the theme docs' production-build section points at it.
- 22 new tests across unit, API, and CLI layers; changeset included.

Expected side effect: repo builds of the shipped webfont themes (neutral, butter, chocolate, gothic, matcha, stone, y2k) now print this warning by design, the same way those builds already print install instructions for them.

## How to see it

Build any theme that names a webfont:

```
$ astryx theme build ./demo.ts
...
⚠ Theme "demo" names fonts it does not load: "Space Grotesk", "Fraunces", "JetBrains Mono"
  The built CSS only sets font-family — load these in your app, or every
  browser quietly falls back.

  Google Fonts (add :wght@… axes as your theme's weights require):

    <link rel="preconnect" href="https://fonts.googleapis.com" />
    <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
    <link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=Space+Grotesk&family=Fraunces&family=JetBrains+Mono&display=swap" />

  Self-hosted (repeat per family and weight):

    @font-face {
      font-family: "Space Grotesk";
      src: url('/fonts/space-grotesk.woff2') format('woff2');
      font-display: swap;
    }

  Recipe and fallback-stack guidance: astryx docs typography
```

A theme that only names system fonts prints nothing new. `astryx docs typography` shows the new "Loading Custom Fonts" section.

### Round 2: edge-case tests

Nine new tests: null and non-string resolved-theme inputs, fontFamily nested under pseudo-class override blocks (unit and through the real defineTheme path), var() fallbacks containing another var(), case-variant generic keywords, multi-line quoted family names, css2 URL percent-encoding, and the --json envelope contract (warnings in the receipt, snippet suppressed, stdout stays parseable).

The multi-line test caught a real tokenizer bug: the quoted-name check lacked dotAll, so a family quoted across lines kept its quote characters, breaking dedupe, system-font matching, and the css2 URL. Fixed with the /s flag.

### DIFF
--- a/.changeset/theme-build-font-warning.md
+++ b/.changeset/theme-build-font-warning.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/cli': patch
+---
+
+[feat] `astryx theme build` warns when a theme names fonts it does not load. The resolved `--font-family-*` tokens and component-override `fontFamily` values are checked against CSS generics and known system families; anything else gets one warning per family in the receipt and, after the install instructions, the `<link>`/`@font-face` snippet to add. `astryx docs typography` gains a Loading Custom Fonts section (Google Fonts and self-hosted recipes, `font-display: swap`, real fallback stacks), and the theme docs' production-build section points at it (#5015).
+
+@AKnassa
--- a/packages/cli/api/theme/build/build.font-warning.test.mjs
+++ b/packages/cli/api/theme/build/build.font-warning.test.mjs
@@ -1,0 +1,133 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+/**
+ * API-contract tests for the font-loading warning in `themeBuild()` (#5015):
+ * a theme that names font families it does not load gets one entry per family
+ * in the `theme.build` receipt's `warnings` array, on BOTH load paths — a raw
+ * typography config (resolved through core's defineTheme) and an
+ * already-resolved theme that sets `--font-family-*` tokens directly. Themes
+ * that only name generics or known system families warn about nothing, and
+ * the warning never breaks the API's silence contract (default noopLogger).
+ *
+ * Needs a built core — the `node` project's globalSetup
+ * (vitest.global-setup.node.mjs) builds it once before workers fork.
+ */
+
+import {describe, it, expect, beforeEach, afterEach, vi} from 'vitest';
+import * as fs from 'node:fs';
+import * as path from 'node:path';
+import * as os from 'node:os';
+import {themeBuild} from './build.mjs';
+
+vi.setConfig({testTimeout: 30000});
+
+let tmpDir;
+beforeEach(() => {
+  tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), 'astryx-theme-fonts-api-'));
+});
+afterEach(() => {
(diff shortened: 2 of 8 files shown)


cixzhang (approved):
Oh thank you for getting this!

Issue #5047, opened 2026-08-15
Vibe test: which authoring aid produces a better theme (docs vs template vs template+refs)


cixzhang:
## What was tested

An agent is asked to build a product theme with `defineTheme`. Does it produce a better theme when pointed at (A) our shipped reference docs, (B) an annotated template file demonstrating every theme feature, or (C) that template plus a consumption guide and a per-section pointer to the CLI reference?

3 product briefs × 3 aids = 9 themes, one fresh context-free agent each, same model, same consumer-shaped project (the nightly vibe-test `project-astryx` environment), same output contract (`src/theme.ts`, then `astryx theme build`). The only thing that varied was one paragraph in the task and which file, if any, sat in the project root. Briefs were deliberately unlike each other — a safety-critical clinical tablet, a loud consumer brand, a long-form editorial reader — so no aid could win by suiting one house style.

Scored two ways: a static harness (`evaluate.mjs`) over the source and the built artifacts, and a blind judge session given the code plus contact sheets of all nine themes rendered against the same ten real Storybook stories in both colour modes.

## Measured, summed over the three briefs

| | A-docs | B-template | C-template+refs |
|---|---|---|---|
| component targets themed | 8 | 11 | **17** |
| style declarations | 54 | 150 | **175** |
| interaction-state blocks | 0 | **4** | 3 |
| custom variants (type augmentations) | 0 | **2** | 1 |
| hand-written tokens | **72** | 27 | 27 |
| WCAG failures (7 pairs × 2 modes × 3 themes) | 8 | 5 | **3** |
| — of those, *authored* rather than inherited | 7 | 4 | **0** |
| webfonts named with no loader shipped | 4 | 6 | **2** |

Blind judge, per brief (round 2, all three arms, letters randomised per brief):

| | arcade | bedside | press |
|---|---|---|---|
| winner | B | **C** (5/5/5) | B |

## What each aid actually does

**The docs teach palette; the template teaches surface.** A wrote 72 hand-authored tokens against B and C's 27 — it re-tones whole ramps, including status colours — but a third of the component work and *no* interaction states in any brief. B and C are the only arms that reached `:hover`/`:active`, custom variants, and `onDark`.

[...]

Issue #5012, opened 2026-08-14
Theme docs teach `--button-press-scale`, a CSS variable no component defines


cixzhang:
`astryx docs theme` → Component Style Overrides teaches this:

```ts
// Some components have public CSS vars for properties that don't map
// to standard CSS. Set these directly.
button: {
  base: { '--button-press-scale': 'scale(0.95)' },
},
```

`--button-press-scale` does not exist. Button's public vars are `--button-focus-offset` and `--button-icon-only-aspect` (`astryx component Button`). A theme that copies the example compiles cleanly and ships a declaration that can never apply — I have one in a generated `arcade.css`:

```
311:    --button-press-scale: scale(0.92);
```

The same fake name is repeated in `theme.doc.dense.mjs` and in the `ComponentThemingVar` JSDoc.

Two things behind it:

1. The snippet declares `button:` **twice in the same object literal**, so the earlier `borderRadius`/`textTransform` block is dead in the example itself.
2. `astryx theme build` validates unknown *props* and rejects private `--_*` vars, but does not check custom properties against the target component's declared vars. It could: the vars are in each component's doc under `theming.vars`. A warning there would have caught this at build time instead of shipping inert CSS.

The doc text is fixed in the PR that follows; the validation gap (2) is the part worth deciding on separately.

cixzhang:
Resolved by #5016 (theme docs updated, --button-press-scale removed).

PR #5048, opened 2026-08-15
feat(cli): `astryx theme template` — an annotated, drift-guarded theme reference

cixzhang:
`astryx theme template` writes an annotated `theme.template.ts` into your project: every `defineTheme` field with a note on when to reach for it, the token families, the component override syntax, and how a theme is consumed (providing it, loading the fonts you name, building for SSR). Each section names the CLI command that prints its authoritative reference, so the template stays a map and the CLI stays the territory. An existing file is never clobbered.

It is the sibling of `theme add`: `add` starts you from a theme we ship, `template` from a blank annotated one. `astryx init --features theme` calls the same leaf — it previously named a feature and wrote nothing, printing a one-line hint to go run `astryx theme`.

This is the shape that won the vibe test in #5047. Agents given an annotated template reached roughly twice as far into the theme surface as agents given only the docs — 17 component targets against 8, and the only arms to use interaction states, custom variants and `onDark` — and shipped a third of the contrast defects. The one place that experiment's template misfired is fixed here: its font paragraph warned that a webfont you don't ship falls back silently, and authors responded by retreating to system stacks (one editorial theme went all-Georgia and was judged "exactly every other content site"). A warning with no recipe reads as *don't*. This version gives the `<link>` and the `@font-face` and says to use the face the design needs.

**A template that lies is worse than no template**, which is not hypothetical: our theme docs taught `--button-press-scale` (#5012) long enough for a generated theme to carry a declaration no component can read. So none of this is trusted prose. `scripts/check-theme-template.test.mjs` derives the facts from live sources and fails when the template drifts — a `defineTheme` field added and left undocumented, a field it teaches that no longer exists, a token family missing from its inventory, a CSS variable or component key it names that nothing defines, a docs topic it cites that prints nothing, or a theme source that drops its SYNC reference. `theme build` compiles the shipped template warning-free as a test, and the CLI typecheck now covers it. The SYNC comments in `defineTheme.ts`, `tokens.stylex.ts` and the four `expand*Scale.ts` point at the template, so the next person changing the theme system is told, in the file they are editing, that the template mirrors it.

Two things came out of review and are worth calling out separately:

**Setup now asks what the app should look like.** The quick-start prompt installed packages and never mentioned the look, so an app that is never asked ships as the default — and retrofitting one means revisiting screens built around the old one. It is now two prompts: set up the system, then give it a look (ask, `theme list`, `theme add <slug>` or `theme template`, default neutral). Two pastes keeps the theme half from bloating the setup half.

**`astryx theme` is not a wizard and never has been** — bare, it prints subcommand help. All three locales of the theme docs recommended it as one ("Use the CLI wizard (recommended)"), and the agent cheat sheet sent brand work to the same dead end. Both now name the real subcommands.

Also: the scaffolded copy no longer carries our copyright header. It lands in someone else's source tree, where our boilerplate is noise their lint may reject. `theme add` already solved this; its private stripper moved to `foundation/text/` and both call sites share it.

## Test plan

- **Each of the seven drift guards was mutated and confirmed to fail, then restored**: new `DefineThemeInput` field; a field the template invents; each token family dropped from the inventory (including `--font-weight-*`, which an earlier prefix implementation could not distinguish from `--font-family-*` — that bug is why the check is scoped to the inventory block); `--button-press-scale` put back (the real #5012 bug — caught); a misspelled component key; a renamed docs topic; a SYNC line removed. A renamed inventory marker throws with instructions rather than passing silently.
- Same for the build test: a private `--_` var in the template fails `compiles as shipped, with no warnings`.
- **The CLI action shipped a bug that only an end-to-end run could see**: it read `result.written` instead of `result.data.written`, so it wrote the file and then told you it had skipped. Caught by running it, fixed, and pinned by `theme-template.behavior.test.mjs` (verified failing on the old code).
- The repo's own drift harness caught the rest of the wiring — `theme.doc.mjs` subcommands, the response-type EnumDoc, the manifest map, the generated README table. All updated; `docs-drift`, `manifest` and `readme-gen` pass.
- The CLI typecheck catches a bogus token in the template (verified by adding one); it did not before, because nothing typechecked `assets/*.ts`.
- End to end in a consumer-shaped project: `theme template` writes it, a second run leaves an edited copy alone, `--overwrite` replaces it, a path escaping the project is refused, `--json` returns a `theme.template` envelope, `init --features theme` writes it too, and `astryx theme build theme.template.ts` produces `.css/.js/.d.ts/.variants.d.ts` with zero warnings.
- `vitest run packages/cli scripts` — 2854 pass. 4 fail: the 2 known macOS case-insensitive-filesystem ones (`hook-discovery`, `component` name resolution) that fail on a clean checkout, and 2 spawn-heavy suites that time out only under full-repo parallelism and pass in isolation. `check:repo` and eslint clean; `typecheck:strict` fails identically with and without this branch (10 pre-existing errors, all missing `@astryxdesign/lab`/`charts` types in template assets).


.claude/commands/create-theme.md
```diff
@@ -2,6 +2,12 @@
 
 Create a new Astryx theme using `defineTheme`.
 
+> Read `packages/cli/assets/theme.template.ts` first. It is the annotated,
+> drift-guarded map of the whole theme surface — every field, the token
+> families, the override syntax — and it names the CLI command that prints the
+> authoritative reference for each. What follows is the packaging around a
+> theme; the template is what goes inside it.
```
cixzhang: This should recommend the CLI command

PR #5062, opened 2026-08-15
fix(lab): respect BottomSheet height above mobile keyboards

imdreamrunner:
## Summary

- Keep effectively Tall BottomSheets anchored to their stable pre-keyboard position and height while internal scrolling reveals the focused control.
- Preserve the measured height of shorter Capped, Hug, and custom-height sheets while moving them upward by the covered distance, clamped to the visible viewport top.
- Classify custom heights from rendered geometry, prevent native focus panning without duplicating focus events, and retain keyboard geometry until dismissal finishes.
- Add consistently named Capped, Hug, and Tall mobile-keyboard stories with forms sized for each presentation.

## Rationale

Tall BottomSheets provide a stable frame for long content, so keyboard accommodation belongs in their internal scroll area. Shorter BottomSheets can use the visible space more effectively by moving above the keyboard while preserving their measured height. Geometry-based classification keeps custom heights aligned with the presentation users actually see.

## Validation

- Physical-phone validation confirmed on the LAN Storybook preview.
- 109 focused BottomSheet tests pass.
- Full repository lint and sync checks pass with zero errors.
- Formatting check passes.
- Lab package build passes.
- Storybook index and all three direct keyboard story URLs return HTTP 200 through the LAN address.

### DIFF
--- a/apps/storybook/stories/BottomSheet.stories.tsx
+++ b/apps/storybook/stories/BottomSheet.stories.tsx
@@ -144,6 +144,69 @@ function MobileKeyboardCommentForm({onPost}: {onPost: () => void}) {
   );
 }
 
+interface CappedCommentFormValues {
+  title: string;
+  author: string;
+  email: string;
+  summary: string;
+  comment: string;
+}
+
+function CappedMobileKeyboardForm({onPost}: {onPost: () => void}) {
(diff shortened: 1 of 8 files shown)


imdreamrunner:
After testing, prefer #5063

PR #5063, opened 2026-08-15
fix(lab): limit BottomSheet keyboard support to Tall

imdreamrunner:
## Summary

- limit BottomSheet mobile-keyboard accommodation to the fully expanded `height="tall"` detent
- keep the outer Tall sheet stationary while smoothly scrolling its internal body
- preserve native canceled and scrolling touch behavior while preventing iOS focus panning
- preserve single consumer focus and blur callbacks
- remove keyboard behavior from shorter Tall detents and all non-Tall heights
- dismiss the keyboard when Tall sheet travel or closing begins
- keep one Mobile keyboard story and document the fully expanded Tall boundary

## Test plan

- focused BottomSheet suite: 113 passed
- lint passed with existing unrelated warnings only
- formatting passed
- sync and export sync passed
- Lab build passed
- Storybook build passed
- LAN Storybook index and direct story return HTTP 200
- full repository suite on the previous revision: 10,299 passed; two current-main theme synchronization tests fail independently
- physical iPhone validation pending

### DIFF
--- a/apps/storybook/stories/BottomSheet.stories.tsx
+++ b/apps/storybook/stories/BottomSheet.stories.tsx
@@ -78,9 +78,15 @@ function MobileKeyboardCommentForm({onPost}: {onPost: () => void}) {
     <VStack gap={4}>
       <Heading level={3}>Add a comment</Heading>
       <Text type="supporting" color="secondary">
-        Focus fields throughout this long form to verify that the mobile
-        keyboard leaves each control visible and the sheet itself stays put.
+        Focus fields near the beginning, middle, and end. The outer Tall sheet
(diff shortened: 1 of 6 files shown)


cixzhang:
Heads-up: the red `test` check here is not yours. `main` is currently failing `scripts/check-theme-template.test.mjs` — [#5048](https://github.com/facebook/astryx/pull/5048) added a theme-template drift guard this morning and [#5026](https://github.com/facebook/astryx/pull/5026) added the `borderDefaults` token family this afternoon; each was green alone, and the guard only fires with both on main. [#5070](https://github.com/facebook/astryx/pull/5070) fixes it. Rebase once that lands and this should go green.

PR #5071, opened 2026-08-15
fix(cli): list border token in theme template

imdreamrunner:
## Summary

- list --border-width in the token-family inventory shipped by astryx theme template
- remove the obsolete drift-guard note that described border tokens as intentionally unsettable
- add a CLI patch changeset

## Reproduction

On main at 9c489238, this fails because borderDefaults is now part of CoreTokenName but the template inventory does not mention it:

    pnpm exec vitest run scripts/check-theme-template.test.mjs -t "lists every settable token family in its inventory"

The assertion reports borderDefaults as the missing family.

## Test plan

- pnpm exec vitest run scripts/check-theme-template.test.mjs
- pnpm exec vitest run packages/cli/api/theme/build/build.test.mjs -t "compiles as shipped, with no warnings"
- pnpm check:repo
- pnpm exec eslint packages/cli/assets/theme.template.ts
- git diff --check

### DIFF
--- a/.changeset/list-the-border-width-token-in-the-generated-the-5437.md
+++ b/.changeset/list-the-border-width-token-in-the-generated-the-5437.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/cli': patch
+---
+
+[fix] List the border-width token in the generated theme template
+@imdreamrunner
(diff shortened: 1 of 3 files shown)


imdreamrunner:
fixed in #5070

Issue #5073, opened 2026-08-15
[RFC] SideNav: resizable and collapsible are two owners of one collapse state


cixzhang:
## Problem

`SideNav` takes `collapsible` and `resizable` as independent props, and **both own collapse state**:

- `collapsible` holds `defaultIsCollapsed` / `isCollapsed` / `onCollapsedChange`, and SideNav keeps its own `uncontrolledCollapsed` (`SideNav.tsx:342`).
- `useResizable` *also* declares `collapsible` and returns its own `isCollapsed`, and persists it (`useResizable.ts:219`).

Nothing reconciles the two at mount, so they can disagree — which is [#4790](https://github.com/facebook/astryx/issues/4790): a persisted collapse restores the hook as collapsed while SideNav renders expanded layout at width 0, i.e. an invisible nav.

The two props aren't really independent today either. `SideNav.tsx:373` passes `collapsible: isCollapsible` into `useResizable`, so **`resizable` on its own cannot collapse** — you have to pass both to get drag-to-collapse. Two props, one behaviour, two owners of one boolean.

## Proposal

`resizable` is a superset of `collapsible` — continuous width *plus* the two-state collapse. So rather than forbidding the combination, **normalize both props into one internal collapse config with a single owner**:

```
resolveCollapse(collapsible, resizable) → one config
owner = resizable ? useResizable : SideNav's own state
```

Passing both then stops being a hazard: there is one state, one writer, and no second `isCollapsed` to diverge from. `collapsible` alone keeps working unchanged (toggle only, no resize, no persistence), so the collapse state machine stays shared rather than moving inside `useResizable`.

**When both are passed and they conflict, `resizable` wins and we fire a dev warning** through `useDevWarning` (`Component: message`, once per mount, stripped in prod) naming the conflicting keys — e.g. a `defaultIsCollapsed` on both. Silent precedence is just the next bug report. Non-conflicting combinations stay quiet.

This is **non-breaking**: no codemod, and today's only working configuration keeps working. If we later want the combination gone entirely, a type-level exclusive union or a lint rule can land once the internals are single-owner — that's a separate decision, not a prerequisite.

## `useResizable` needs the standard state API

[...]

PR #5075, opened 2026-08-15
fix(SideNav): give collapse a single owner across collapsible and resizable

### LINKED ISSUE #4790: [Bug] SideNav: collapsing then reloading persists width 0, making the nav invisible and unrecoverable

cixzhang:
## Why

`SideNav` has two owners of one collapse boolean. `collapsible` holds `defaultIsCollapsed` / `isCollapsed` / `onCollapsedChange` and SideNav keeps its own state from them; `useResizable` declares its *own* `isCollapsed`, seeds it from storage and persists it. Nothing reconciles the two at mount, so a persisted collapse restores the hook collapsed while SideNav renders the expanded layout at width `0` — an invisible nav whose expand button is gone with it, unrecoverable without clearing localStorage ([#4790](https://github.com/facebook/astryx/issues/4790)).

The design is [#5073](https://github.com/facebook/astryx/issues/5073); this implements it.

## What

**One collapse config, one owner.** `collapsible` and `resizable` are normalized into a single resolved collapse config, then `useResizable` owns that state in every SideNav mode. Turning resize on or off leaves the owner mounted, so there is no second `isCollapsed` to diverge from and no owner-handoff reset. The [#4790](https://github.com/facebook/astryx/issues/4790) class of failure is structurally gone rather than patched. `collapsible` alone remains toggle-only, with no resize handle or persistence.

**A dev warning when the two props address the same state.** Through the existing `useDevWarning`: `defaultIsCollapsed` on both, or a controlled `collapsible` alongside collapse state on `resizable` — the shape where a consumer's setter looks like the source of truth and isn't. The message names the conflicting keys and says `resizable` wins. Non-overlapping combinations (`collapsible: true` plus a resize config — the common case) stay silent.

**The standard controlled/uncontrolled pair on `useResizable`.** `defaultIsCollapsed` and `isCollapsed` join the existing `onCollapseChange`, matching `SideNavItem.collapsible`, `Selector` and `Collapsible`. Uncontrolled, the hook owns collapse exactly as today, seeded from persistence or `defaultIsCollapsed`. Controlled, the prop wins: `collapse()` / `expand()` become intents that report through `onCollapseChange`, and a drag past the threshold reports rather than mutates. Additive — the uncontrolled path is unchanged for direct consumers. This supersedes [#4824](https://github.com/facebook/astryx/pull/4824)'s narrower `initialIsCollapsed`.

**A persisted shape that can express collapse.** Entries are now `{size, isCollapsed}` where `size` is the *expanded* size, so a reload restores the collapsed rail **and** expanding returns the width the user had instead of `defaultWidth`. Internally the hook stops zeroing `size` on collapse — the collapsed `0` is a view of the state, not a destroyed value. Legacy entries still load: a plain number is a width that says nothing about collapse, and a plain `0` (what the old collapse path wrote) reads as collapsed, so anyone currently stuck with an invisible nav recovers on next load without clearing storage.

**Collapse notifications read live state — [@AKnassa](https://github.com/AKnassa)'s fix, carried from [#5118](https://github.com/facebook/astryx/pull/5118).** The hook's collapse callbacks read `isCollapsed` from the render that created them. That is wrong for two real cases: two imperative calls in one tick, and a drag — `ResizeHandle` registers its pointermove listener once at pointer down, so a whole gesture runs against one props snapshot. Against a `useResizable` consumer whose handle stays mounted, one drag past the threshold fired `onCollapseChange(true)` five times and dragging back above the threshold never re-expanded. A ref mirroring the flag fixes both, `collapse()` is inert when already collapsed, and `resize()` out of the collapsed state now reports the implicit expand the way the drag path does. Through SideNav the drag half is invisible today, because SideNav unmounts the handle on collapse; a direct hook consumer sees all of it.

`resizable` can now carry collapse state on its own, which is what makes "both props address the same state" expressible — and warnable.

**Non-breaking.** No codemod. Every configuration that works today resolves to the same collapse state it does now; see the testing note below for how that is pinned. The single case whose outcome changes is the divergence itself: a persisted-collapsed nav renders the collapsed rail instead of an invisible zero-width one.

## Risk

**Rollback.** An older core reading a new-format entry parses it, sees an object rather than a number, and ignores it: the nav comes up expanded at `defaultWidth`. The saved width and collapse state are lost on downgrade, but the old invisible-nav failure is *not* reintroduced, because the old reader only produces that state from a literal `0`. Rolling forward again reads the entry normally. On first load after this lands, a legacy entry is upgraded in place to the new shape.

**Timing.** With one owner, `collapsible.onCollapsedChange` fires **once** per toggle on a resizable nav; today it fires twice (SideNav notifies, then the hook's callback notifies again). Anything counting those calls will see one. The live-state fix removes further duplicates on the drag path: one collapse per gesture rather than one per pointer move.

**`expand()` no longer re-derives the width.** It restores the retained expanded size rather than a pre-collapse ref, so an `expand()` after a programmatic `resize()` keeps the resized width instead of snapping back to the width from before the last collapse.

**`expand()` is now silent when nothing was collapsed**, matching `collapse()`. A consumer calling `expand()` on an already-expanded region used to get an `onCollapseChange(false)` for a transition that never happened.

`SideNavCollapseButton` rendered outside the SideNav ([#4880](https://github.com/facebook/astryx/pull/4880)) is covered: its controlled config flows into the resolved config and drives the resizable nav, with a test for it.

## Testing

`vitest run --project ui packages/core` — 6722 pass, 252 files. `@astryxdesign/core` `typecheck` and `typecheck:docs` clean; eslint reports 0 errors (48 warnings, all pre-existing on `main` and none in the files this touches).

New coverage: the hook's persistence round-trips and every legacy format; uncontrolled and controlled collapse including the drag-reports-not-mutates path; `defaultIsCollapsed` + `resizable`; the warning firing on each conflicting shape and staying silent on the non-overlapping ones; and the [#4790](https://github.com/facebook/astryx/issues/4790) regression (legacy `0` → collapsed rail, plus collapse → reload → expand restoring the pre-collapse width).

**The non-breaking claim is pinned by a test, not by prose.** `SideNav collapse ownership` enumerates nine currently-working configurations with the collapse state each resolves to. That block was run against unmodified `origin/main`: all nine pass there and here. The one row that fails on `main` and passes here is the legacy-`0` case — the bug.

**The live-state fix is pinned by four tests that fail without it.** They hold the props object captured at pointer down rather than re-reading it between moves, which is what `ResizeHandle` does and what the existing drag tests do not. Measured on this branch, before → after:

| Case | Before | After |
|---|---|---|
| One drag, five pointer moves past the threshold | `onCollapseChange` fires `true` ×5 | fires `true` ×1 |
| Drag past the threshold, then back above it | stays collapsed, `size` 0 | re-expands mid-gesture, `size` 240 |
| `resize()` out of the collapsed state | silent | fires `false` ×1 |
| `collapse()` twice in one tick | fires `true` ×2 | fires `true` ×1 |

**Verified in Chromium, not just jsdom.** Storybook's `Lab/Resizable` → `Collapsible` story is a direct `useResizable` consumer whose `ResizeHandle` stays mounted across collapse, so the whole gesture is reachable. Dragging the handle past the threshold and back **without releasing the pointer**: before, the sidebar collapsed and stayed collapsed for the rest of the gesture; after, it re-expands mid-drag at 220px. Frames captured for both.

**Verified in real Chromium on this head.** A focused `Collapse Ownership` story covers collapsible-only, resizable-only, both, persisted reload, resize-owner handoff, and 220px → 380px transitions. Every capture has a sensor receipt for the exact SHA, story, theme, mode, direction, viewport, semantic state, geometry, and browser errors. The persisted case reloads as a visible 48px collapsed rail, then expands to its saved 320px width.

### Visual acceptance

**The failure and recovery**

Main restores expanded layout at zero width; this PR restores a collapsed rail and expands to the saved width

**Required state matrix**

Collapsible-only, resizable-only, both, persisted reload, narrow, and wide SideNav states

## Credit

Builds on three contributor PRs, all still worth reading:

- [#4824](https://github.com/facebook/astryx/pull/4824) by @AKnassa — the `{size, isCollapsed}` storage format, its legacy-entry handling, and much of the hook's test suite are theirs.
- [#5118](https://github.com/facebook/astryx/pull/5118) by @AKnassa — the live-state collapse notifications above. The approach and the diagnosis are his; the commit here is co-authored. His version also guards a pre-collapse width ref against the zeroed collapsed size, which this branch does not need — `size` stays the expanded size, so there is no ref to clobber.
- [#4853](https://github.com/facebook/astryx/pull/4853) by @HelloOjasMutreja — the diagnosis that `main` already clamps and already seeds, and that the real defect is the second independently-initialized boolean.

Closes #4790. Refs [#5073](https://github.com/facebook/astryx/issues/5073).


cixzhang:
Direction: keep the single-owner solution. Rebase it over the current ref changes, then finish the visual acceptance check before it advances.

cixzhang (commented):
Thanks, this is good. The single-owner, persistence, held-pointer, and controlled-timing paths all check out.

[Full review](https://github.com/cixzhang/astryx/wiki/Review-5075)

[Reviewed by Robohands]

cixzhang:
/accept-visual 33043491070/1 The persisted reload now matches the existing 48px collapsed rail byte-for-byte; collapsible-only, resizable-only, combined, owner-handoff, and 220–380px transitions are unchanged or intentional.

cixzhang:
Visual changes accepted for `cdc12ee416b0`. The canonical evidence and acceptance record are archived at [the visual report](https://facebook.github.io/astryx/pr/5075/visual/cdc12ee416b0d7b9d6456d02b4ba787b186fa0b4/33043491070/1/index.html). The workflow projection was completed manually because [#5590](https://github.com/facebook/astryx/issues/5590) blocked both trusted publication and custom-role authorization.

PR #5077, opened 2026-08-16
fix(cli): the template's font advisories are not defects

cixzhang:
## Why

`main` is red again, one commit after the last one. [#5045](https://github.com/facebook/astryx/pull/5045) added a font advisory to the build receipt's `warnings`:

```
× packages/cli/api/theme/build/build.test.mjs > compiles as shipped, with no warnings
  + "Font \"Inter\" is named by this theme but not loaded — …"
  + "Font \"Geist Mono\" is named by this theme but not loaded — …"
```

The shipped template names both fonts, so the advisory fires on it, and the pre-existing template guard reads any warning as a defect.

Neither side is wrong, which is why this is not a one-liner in either direction. The advisory is genuinely useful — a named family with no file falls back silently. But **no theme file can load a font**, so a template that teaches font naming will always trip it. The template's header already opens with `SHIP THE FONTS YOU NAME` and gives both the Google Fonts `<link>` and the `@font-face` recipe: the advisory tells the reader precisely what the template has already told them.

## What

The guard states what it actually means: the template compiles with exactly the two advisories its own fonts produce, and no other warning. An unknown component key, a private-var error, or a third unloaded font still fails it — so it keeps its teeth, and `warnings` keeps carrying the advisory for everyone else.

## Testing

`vitest run packages/cli/api/theme/build/` — 39 passed (1 failed before, on `main` as well as here).

### DIFF
--- a/.changeset/template-font-advisory.md
+++ b/.changeset/template-font-advisory.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/cli': patch
+---
+
+[fix] `main` was left red by #5045: the font advisory it added fires on the shipped theme template, which the template's own guard reads as a defect. Both are right — the template names Inter and Geist Mono deliberately, and no theme file can load a font, which is why its header opens with SHIP THE FONTS YOU NAME and gives both recipes. So the guard now says what it means: the template compiles with exactly those two font advisories and no other warning, instead of none at all. A real defect in the template still fails it.
+
+@cixzhang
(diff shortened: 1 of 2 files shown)


cixzhang:
Superseded by #5079, which landed the same fix (the template's font advisories are by design, any other warning still fails). Main is green on `packages/cli/api/theme/build/build.test.mjs` again — verified locally at b2057d1668. Closing this as the duplicate; nothing here that #5079 does not already cover.

Issue #4621, opened 2026-08-01
[Docs] Generated theme modules require a compiled icons module alongside — this isn't documented


imdreamrunner:
### Description

When a theme's `icons:` field is fed by a relative import, `astryx theme build` writes that import into the module it generates:

```js
// dist/neutral.js
import { neutralIconRegistry } from './icons';
```

Producing a loadable module at that path is the caller's responsibility — `theme build` does not compile the icon registry. That's a reasonable division of labour, and the theme packages satisfy it with their own pipeline (`astryx theme build && tsup && tsc`, where `tsup` emits `dist/icons.mjs`).

But nothing says so. The docs list the four emitted files (`.css`, `.js`, `.d.ts`, optional `.variants.d.ts`) without mentioning that one of them carries an external requirement, and a reader would have to infer it from the theme packages' build scripts.

### What a theme author actually has to do

Two steps, starting from a stock `astryx theme add` scaffold:

```bash
$ astryx theme build src/neutralTheme.ts
$ sed -n 8p src/neutral.js
import { neutralIconRegistry } from './icons';
$ node -e "import('./src/neutral.js')"
ERR_MODULE_NOT_FOUND                       # icons are still TypeScript

# 1. compile the icon registry  (2.0 kB)
$ esbuild src/icons.tsx --bundle --format=esm --outfile=src/icons.mjs \
    --external:react --external:lucide-react --jsx=automatic

# 2. point the generated module at it
$ sed -i '' "s#from './icons'#from './icons.mjs'#" src/neutral.js

$ node -e "import('./src/neutral.js')"
LOADS — theme=neutral, icons=26
$ esbuild src/neutral.js --bundle --format=esm --outfile=/dev/null
BUNDLES OK
```

Do both and the module works in Node and in bundlers. Skip step 1 and it works only in bundlers, which resolve `./icons` to the neighbouring `icons.tsx`. Skip step 2 and it never works in Node, because `./icons` is not a valid ESM specifier — that part is #4620, and step 2 isn't currently possible without hand-editing a `@generated` file.

The most visible symptom of the missing documentation is that this exits 0:

```bash
$ astryx theme build src/myTheme.ts -o dist/theme.css   # no compile step in this project
$ ls dist
my.d.ts  my.js  my.variants.d.ts  theme.css             # nothing named icons
```

That module resolves in nothing — Node, esbuild and Vite all fail — and the build reported success.

### Suggested

[...]

PR #5099, opened 2026-08-16
docs(cli): clarify imported theme icon build requirements

jiunshinn:
Theme authors can generate a theme module successfully and still lose its icons or encounter an unresolved import. The production-build guide now explains the current named-import flow, the omission of inline registries, and how the generated file's location affects icon imports.

The English, dense, and Chinese guidance:

- Scopes icon emission to a detected named import used by `icons:`. Inline registries, including local constants, are currently omitted; authors using this build flow should move the registry into an imported module.
- Shows `theme build --out dist/theme.css --icons-specifier ./icons.mjs` followed by an esbuild command that emits `dist/icons.mjs`, with React and lucide-react external.
- Explains that a missing `dist/icons.mjs` breaks both module loading and bundling, even when theme generation succeeds.
- Separately describes extensionless source imports: a bundler can resolve neighboring `icons.tsx` in the default output layout, while Node ESM cannot. Moving the output changes relative-path resolution.

This is consumer documentation of current behavior. It does not establish a permanent artifact-packaging policy or change the generator. The changeset uses the same scope. All other sections in the three documentation files are unchanged from `main`.

Related: #4621. Inline-registry diagnostics are tracked separately in #5104.

## Validation

GitHub Actions for `92680abf32d` passed: [CI (Linux node/UI tests and builds)](https://github.com/facebook/astryx/actions/runs/34340076613), [Lint](https://github.com/facebook/astryx/actions/runs/34340076418), and [CLI Smoke Test](https://github.com/facebook/astryx/actions/runs/34340076471).

On Node 24.14.0 and pnpm 11.10.0:

- `pnpm build` — passed.
- `pnpm lint:strict` — passed, 0 errors (89 warnings).
- `pnpm -F @astryxdesign/cli typecheck:strict` — passed.
- `pnpm test --maxWorkers=4` — 16,051 passed, 50 skipped, 2 failed; 749 files passed, 3 skipped, 1 failed. Both failures and the failed setup hook are in `internal/vibe-tests/setup-test/setup-workspace.test.mjs`: macOS `cp` rejects `--reflink=auto`. The same suite reproduces the same failures on a separate, unchanged `main` checkout at `53bc34db7c9` using the same dependencies. All other suites passed.
- English, dense, and Chinese CLI rendering — checked; all 11 production-section blocks align, and every other section matches `main`.
- Repository pre-commit checks — passed.

A temporary fixture with a named registry containing a lucide-react JSX icon exercised the documented output paths:

| Case | Observed result |
| --- | --- |
| Explicit `./icons.mjs`, before registry compilation | Theme build succeeds; Node and esbuild cannot resolve the missing module |
| Same output after the documented esbuild command | Node loads the theme with its icon; esbuild bundles it |
| Default output beside source, unchanged `./icons` | esbuild bundles it; Node reports `ERR_MODULE_NOT_FOUND` |
| Output moved to another directory, unchanged `./icons` | esbuild cannot resolve the import |
| Registry declared as a local constant | Current generated theme omits the `icons` field |


cixzhang (changes requested):
Builders following this guide can still lose their icon registry or get the wrong failure diagnosis.

`defineTheme` accepts inline `icons:` registries, but `theme build` only emits an icon import when it finds a matching imported binding; inline registries are omitted. Open #5104 adds a warning, not support. The shown `--out ... --icons-specifier ./icons.mjs` flow also fails in bundlers when `dist/icons.mjs` is absent; bundler fallback to neighboring `icons.tsx` applies only to the default no-`--out`, no-override flow.

Please scope the text to imported registries, separate those failure cases, and keep it factual. Current architecture owns artifact policy and does not establish a permanent caller-owned sidecar contract. This head also conflicts with current `main` in the dense and Chinese overlays.

[Reviewed by Robohands]

cixzhang (approved):
Theme authors now get accurate guidance: missing compiled registry modules fail Node and esbuild; neighboring extensionless sources alone can bundle. Current-main merge passes.

[Reviewed by Robohands]

PR #5100, opened 2026-08-16
fix(storybook): resolve @astryxdesign/richtext from source in dev

### LINKED ISSUE #5092: [Infra] Storybook dev cannot resolve @astryxdesign/richtext in fresh agent worktrees
## Problem

AI agents commonly create a fresh worktree and run Storybook without first building every workspace package. In that workflow, `pnpm -F @astryxdesign/storybook dev` announces that Storybook is ready, but Vite then fails its dependency scan because it cannot resolve `@astryxdesign/richtext`.

The workspace link itself is valid, but `packages/richtext/package.json` resolves the default export to `./dist/index.js`. A fresh worktree has no `packages/richtext/dist`. Although `apps/storybook/tsconfig.json` maps Rich Text to `packages/richtext/src`, the Vite and StyleX source aliases in `apps/storybook/.storybook/main.ts` omit `@astryxdesign/richtext`.

This makes Storybook startup non-deterministic for agent workflows and requires the agent to discover and run an unrelated package build before reviewing a story.

## Evidence

Reproduced in a fresh PR worktree at `28bdbbd9`:

```sh
pnpm install
test ! -e packages/richtext/dist/index.js
```

Vite reports:

```text
Vite [optimizer] scanning dependencies...

Failed to run dependency scan. Skipping dependency pre-bundling.
The following dependencies are imported but could not be resolved:

  @astryxdesign/richtext
  (imported by apps/storybook/stories/RichTextEditor.stories.tsx)
```

The pnpm link points to the expected workspace package:

```text
apps/storybook/node_modules/@astryxdesign/richtext
  -> ../../../../packages/richtext
```

Node resolution confirms that the missing generated entry point is the failure:

```text
Cannot find module '.../packages/richtext/dist/index.js'
```

Verified workaround:

```sh
pnpm -F @astryxdesign/richtext build
```

After building Rich Text, the unresolved-dependency error disappears and Vite proceeds with dependency optimization.

Environment:

- macOS
- Node 22.17.1
- pnpm 11.10.0
- Storybook 10.4.6
- Vite 8.1.3

## Scope Check

- `@astryxdesign/richtext` is already a direct Storybook dependency and already has TypeScript source paths.
- Storybook explicitly aliases Core, Lab, Charts, Vega, and themes to source, but not Rich Text.
- The failure is specific to the development optimizer path; a production Storybook build can still succeed through its separate transform/resolution path.
- This report intentionally excludes port selection and other local Storybook startup concerns.

## Proposed Direction

Resolve Rich Text from source in Storybook, consistently with the other workspace packages:

- add `@astryxdesign/richtext` and its subpaths to the StyleX aliases;
- add `@astryxdesign/richtext` to Vite `resolve.alias`;
- add a regression check that starts or resolves Storybook with `packages/richtext/dist` absent.

Alternatively, make the Storybook dev command build Rich Text first, though a source alias should preserve hot reload and match the existing package-development model.

Acceptance criterion: a fresh clone or AI-agent worktree can run `pnpm -F @astryxdesign/storybook dev` without manually building `@astryxdesign/richtext`.

jiunshinn:
Closes #5092

## Problem

In a fresh worktree, `pnpm -F @astryxdesign/storybook dev` announces ready and then fails Vite's dependency scan:

```
The following dependencies are imported but could not be resolved:
  @astryxdesign/richtext
  (imported by apps/storybook/stories/RichTextEditor.stories.tsx)
```

## Root cause

`apps/storybook/.storybook/main.ts` aliases every other workspace package to source (Core, Lab, Charts, Vega, themes) in both the Vite `resolve.alias` block (`main.ts:118`) and the StyleX `aliases` config (`main.ts:82`), but omits `@astryxdesign/richtext`. Without an alias, Vite falls back to the workspace link's export map, whose `default` entry is `packages/richtext/dist/index.js` — which does not exist until the package is built. `apps/storybook/tsconfig.json` already maps richtext to source, so typecheck and dev disagreed.

## What changed

- `apps/storybook/.storybook/main.ts`
  - StyleX `aliases`: added `@astryxdesign/richtext/*` and `@astryxdesign/richtext` → `packages/richtext/src`, mirroring the core/lab/charts entries.
  - Vite `resolve.alias`: added `@astryxdesign/richtext` → `packages/richtext/src` (bare key prefix-matches subpaths, same as the existing core entry that serves `@astryxdesign/core/reset.css`).
- `apps/storybook/.storybook/main.test.ts` (new): regression guard — asserts every `@astryxdesign/*` dependency of the Storybook app appears as an alias key in `main.ts`, so the next new workspace package can't regress the same way. Static (reads `package.json` + `main.ts`), so it stays green/red regardless of what's built.
- `vitest.config.ts`: one include line so the root `node` test project collects `apps/storybook/.storybook/**/*.test.*`.

## Acceptance criteria (from #5092)

- [x] `@astryxdesign/richtext` + subpaths added to the StyleX aliases
- [x] `@astryxdesign/richtext` added to Vite `resolve.alias`
- [x] Regression check covering Storybook resolution with `packages/richtext/dist` absent
- [x] A fresh worktree can run `pnpm -F @astryxdesign/storybook dev` without manually building `@astryxdesign/richtext`

## Verification (fresh worktree, `packages/richtext/dist` absent throughout)

```sh
$ test ! -e packages/richtext/dist/index.js && echo "no richtext dist"   # confirmed before boot
$ pnpm -F @astryxdesign/storybook exec storybook dev -p 6119 --no-open --ci
# → "Storybook ready!" — no "Failed to run dependency scan", no "could not be resolved"
```

- Vite transform proof: `curl /stories/RichTextEditor.stories.tsx` shows the import rewritten to `/@fs/…/packages/richtext/src/index.ts`; `index.json` lists 19 richtext story entries.
- Richtext source modules compile through the StyleX pipeline on the fly: `index.ts`, `RichTextEditor.tsx`, `RichTextView.tsx`, `RichTextEditorToolbar.tsx` all served 200 with transformed output.
- Headless browser: `lab-richtexteditor--default` renders (label, placeholder, toolbar-less editor); typed into the Lexical `contenteditable` and read the text back — fully interactive, no error overlay.
- Regression test red/green: with the `main.ts` fix stashed, `vitest run apps/storybook/.storybook/main.test.ts` → 1 failed (`aliases @astryxdesign/richtext to source`), 8 passed; with the fix → 9 passed (9).
- `pnpm -F @astryxdesign/storybook typecheck` → clean (after building charts/themes; the 24 pre-existing TS2307s in chart/theme stories exist on `main` in any fresh worktree and are untouched by this PR).
- `eslint` + `prettier --check` on the three changed files → clean; all `check:repo` scripts pass (pre-commit).

## Reviewer flags

- **Regression check shape**: the issue floats "a regression check that starts … Storybook with `packages/richtext/dist` absent". Booting Storybook in CI is heavy, so the guard is a static invariant test instead: every `@astryxdesign/*` dependency of the app must have an alias key in `main.ts`. It checks key presence in the file text (not which block the key is in) — deliberate, to keep it cheap and format-tolerant. Happy to swap for a heavier boot-based check if preferred.
- **`vitest.config.ts` include**: smallest wiring that lets the root `node` project collect the new test; scoped to `apps/storybook/.storybook/` only.
- **Out-of-scope discovery**: a *truly* clean worktree also fails earlier — `main.ts` itself imports `@astryxdesign/build/vite`, whose export map points at `packages/build/dist/vite.mjs` (also unbuilt). That is the Storybook config loader's node ESM resolution, not Vite's, so a Vite alias can't fix it; left out of scope (issue #5092 is richtext-specific — its author's repro had `build` compiled). Can file separately.
- **No changeset**: app-only (private `@astryxdesign/storybook`) + root test config, per the app-only precedent (#3744).

### DIFF
--- a/apps/storybook/.storybook/main.test.ts
+++ b/apps/storybook/.storybook/main.test.ts
@@ -1,0 +1,42 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+/**
+ * @file main.test.ts
+ * @description Guards the workspace source-alias invariant in Storybook's
+ *   Vite config. Every `@astryxdesign/*` dependency of this app must be
+ *   aliased to package source in `.storybook/main.ts`. Without an alias,
+ *   Vite resolves the workspace link through the package's export map,
+ *   which points at `dist/` build output that does not exist in a fresh
+ *   worktree — `storybook dev` then fails its dependency scan (#5092:
+ *   `@astryxdesign/richtext`).
+ */
+
+import {describe, it, expect} from 'vitest';
+import {readFileSync} from 'node:fs';
+import path from 'node:path';
+import {fileURLToPath} from 'node:url';
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+
+const mainTs = readFileSync(path.join(__dirname, 'main.ts'), 'utf8');
+const pkg = JSON.parse(
+  readFileSync(path.join(__dirname, '..', 'package.json'), 'utf8'),
+) as {dependencies?: Record<string, string>};
+
+const workspaceDeps = Object.keys(pkg.dependencies ?? {}).filter(name =>
+  name.startsWith('@astryxdesign/'),
+);
+
+describe('storybook main.ts workspace source aliases', () => {
+  it('sees the @astryxdesign workspace dependencies', () => {
+    expect(workspaceDeps).not.toHaveLength(0);
+  });
+
+  it.each(workspaceDeps)('aliases %s to source', dep => {
+    // Each workspace dependency must appear as a bare alias key (in
+    // `resolve.alias`) so dev-mode resolution never falls back to the
+    // export map's dist/ entry point, which is missing until the package
+    // is built.
+    expect(mainTs).toContain(`'${dep}':`);
+  });
+});
--- a/apps/storybook/.storybook/main.test.ts
+++ b/apps/storybook/.storybook/main.test.ts
@@ -23,6 +23,10 @@ const pkg = JSON.parse(
   readFileSync(path.join(__dirname, '..', 'package.json'), 'utf8'),
 ) as {dependencies?: Record<string, string>};
 
+// The Vite `resolve.alias` block only — the StyleX `aliases` block above it
+// carries the same keys but does not drive Vite's dev-mode resolution.
+const viteAliasBlock = mainTs.slice(mainTs.indexOf('alias: {'));
+
 const workspaceDeps = Object.keys(pkg.dependencies ?? {}).filter(name =>
   name.startsWith('@astryxdesign/'),
 );
@@ -33,10 +37,9 @@ describe('storybook main.ts workspace source aliases', () => {
   });
 
   it.each(workspaceDeps)('aliases %s to source', dep => {
-    // Each workspace dependency must appear as a bare alias key (in
-    // `resolve.alias`) so dev-mode resolution never falls back to the
-    // export map's dist/ entry point, which is missing until the package
-    // is built.
-    expect(mainTs).toContain(`'${dep}':`);
+    // Each workspace dependency must appear as a bare alias key in Vite's
+    // `resolve.alias` so dev-mode resolution never falls back to the export
+    // map's dist/ entry point, which is missing until the package is built.
+    expect(viteAliasBlock).toContain(`'${dep}':`);
   });
 });
--- a/apps/storybook/.storybook/main.ts
+++ b/apps/storybook/.storybook/main.ts
@@ -92,6 +92,12 @@ const config: StorybookConfig = {
               '@astryxdesign/charts': [
                 path.join(rootDir, 'packages/charts/src'),
               ],
+              '@astryxdesign/richtext/*': [
+                path.join(rootDir, 'packages/richtext/src/*'),
+              ],
+              '@astryxdesign/richtext': [
+                path.join(rootDir, 'packages/richtext/src'),
+              ],
               '@astryxdesign/theme-neutral/*': [
                 path.join(rootDir, 'packages/themes/neutral/src/*'),
               ],
@@ -120,6 +126,10 @@ const config: StorybookConfig = {
           '@astryxdesign/core': path.resolve(rootDir, 'packages/core/src'),
           '@astryxdesign/lab': path.resolve(rootDir, 'packages/lab/src'),
           '@astryxdesign/charts': path.resolve(rootDir, 'packages/charts/src'),
+          '@astryxdesign/richtext': path.resolve(
+            rootDir,
+            'packages/richtext/src',
+          ),
           '@astryxdesign/theme-neutral': path.resolve(
             rootDir,
             'packages/themes/neutral/src/source.ts',
--- a/vitest.config.ts
+++ b/vitest.config.ts
@@ -142,6 +142,9 @@ export default defineConfig({
             'internal/**/*.test.{ts,tsx,mjs}',
             'scripts/**/*.test.{ts,tsx,mjs}',
             '.github/scripts/**/*.test.{ts,tsx,mjs}',
+            // Storybook config invariants (no DOM needed) — e.g. the
+            // workspace source-alias guard in .storybook/main.test.ts.
+            'apps/storybook/.storybook/**/*.test.{ts,tsx,mjs}',
           ],
           exclude: [
             ...configDefaults.exclude,


apps/storybook/.storybook/main.test.ts
```diff
@@ -0,0 +1,42 @@
...
+  it('sees the @astryxdesign workspace dependencies', () => {
+    expect(workspaceDeps).not.toHaveLength(0);
+  });
+
+  it.each(workspaceDeps)('aliases %s to source', dep => {
+    // Each workspace dependency must appear as a bare alias key (in
+    // `resolve.alias`) so dev-mode resolution never falls back to the
+    // export map's dist/ entry point, which is missing until the package
+    // is built.
+    expect(mainTs).toContain(`'${dep}':`);
```
imdreamrunner: This matches anywhere in the file, and the StyleX `aliases` block above carries the same key text — so the assertion can't tell the block that fixes #5092 from the one that doesn't. I removed just the Vite `resolve.alias` richtext entry and this suite stayed 9/9 green while `storybook dev` failed its dependency scan exactly as in the issue.

Scoping the haystack to the Vite block fixes it and keeps the test static:

```suggestion
    // Each workspace dependency must appear as a bare alias key in Vite's
    // `resolve.alias` so dev-mode resolution never falls back to the export
    // map's dist/ entry point, which is missing until the package is built.
    expect(viteAliasBlock).toContain(`'${dep}':`);
```

together with, just above `const workspaceDeps` (line 26):

```ts
// The Vite `resolve.alias` block only — the StyleX `aliases` block above it
// carries the same keys but does not drive Vite's dev-mode resolution.
const viteAliasBlock = mainTs.slice(mainTs.indexOf('alias: {'));
```

Verified both ways: 9/9 green against all eight workspace deps as they stand, and red (`× aliases @astryxdesign/richtext to source`) as soon as the Vite entry goes away.

imdreamrunner (changes requested):
🔴 **Blocking — `needs:code-review`** (community contribution; the workflow gates every contributor PR).

**Verdict: request changes — one test-only fix, then this is ready to merge.** The alias fix is correct and I verified it end to end; nothing in the production change is in question. The single finding is in the new regression test, which passes in exactly the broken state #5092 describes.

`Triage: bug fix (dev tooling/config) · non-breaking · low blast radius (apps/storybook + root test config, nothing publishable) → fast path · checks: X20 (n/a), config + test-scaffolding review`

Description and test plan are exemplary — root cause named precisely (the export map resolving to `packages/richtext/dist/index.js`, absent in a fresh worktree) rather than guessed, verification reproducible, and the "Reviewer flags" section volunteers the two things a reviewer would otherwise have to dig out. Thank you for that; it made this review fast.

### What I verified independently

Fresh clone at `fcbac1d`, Linux, `packages/richtext/dist` absent throughout:

| Check | Result |
| --- | --- |
| `storybook dev` at PR head | Boots clean — no dependency scan failure. **Fix works.** |
| Same, with only the Vite `resolve.alias` richtext entry removed | Reproduces #5092 verbatim (`Failed to run dependency scan … @astryxdesign/richtext`) |
| `main.test.ts` in that broken state | **9 passed** — see finding below |
| CI `test` job | Collects `apps/storybook/.storybook/main.test.ts (9 tests)` in the `node` project — the vitest wiring works |

The Vite alias shape is right: `@rollup/plugin-alias` matches exact-or-`find + '/'`, so the bare key covers subpaths — the same reason the existing `@astryxdesign/core` entry already serves `@astryxdesign/core/reset.css`. The StyleX `/*` + bare pair mirrors core/lab/charts.

### 🟡 The regression guard passes in the state it exists to catch

`expect(mainTs).toContain(...)` searches the whole file, and the StyleX `aliases` block holds the identical key text. I deleted only the Vite `resolve.alias` entry — precisely the #5092 bug — and got a green suite alongside a Storybook failing its dependency scan.

So the acceptance criterion "regression check covering Storybook resolution with `packages/richtext/dist` absent" isn't met yet: this guard would not have caught #5092 and won't catch its return. That's the one thing I'd like fixed before approval — not because test scaffolding is blocking (per `copilot-instructions.md` it isn't), but because catching that regression is the whole job of this file.

The fix stays static and boot-free — see the inline comment. It also makes the assertion match what the file's own docblock already promises ("as a bare alias key (in `resolve.alias`)").

Scope it to the Vite block only, **not** symmetrically to StyleX: `@astryxdesign/vega` has a Vite alias and no StyleX entry today, and the theme packages are `/*`-only there, so a StyleX-side assertion would go red on pre-existing state. Pre-existing, and not yours to fix.

### Notes

- **Your out-of-scope finding is real** — I hit it on the first boot: with `packages/build/dist` absent, Storybook can't start at all until `pnpm -F @astryxdesign/build build`. Please do file it. Same class of bug, and it's what a genuinely clean agent worktree hits *first*, so #5092's fix alone doesn't deliver the "fresh worktree just works" outcome. A Vite alias can't reach it (that's the config loader's node ESM resolution), so it needs its own issue.
- **No changeset: agreed** — nothing publishable is touched, which carries on its own. (The #3744 cite is a bit loose — that PR did touch `packages/` — but the conclusion holds.)
- The `vitest.config.ts` include is the minimal wiring and correctly lands in the `node` project.

Once the guard is tightened (or you make the case for keeping it broad), this is good from me. The `review-required` gate still needs an entitled code owner's approval to clear.

apps/storybook/.storybook/main.test.ts
jiunshinn: Good catch, and thanks for the independent end-to-end verification — you're right that the whole-file haystack made the guard blind to exactly the block that matters. Applied your suggestion essentially verbatim in a0f1040091: `viteAliasBlock = mainTs.slice(mainTs.indexOf('alias: {'))` (first match lands on Vite's `alias: {`; the StyleX block's key is `aliases: {`, and nothing after the slice point carries other `@astryxdesign` keys), with the assertion scoped to that block.

Two-way red/green at a0f1040091:
- removed only the Vite `resolve.alias` richtext entry (StyleX `aliases` untouched) → `× aliases @astryxdesign/richtext to source`, 1 failed | 8 passed — the pre-rework assertion stayed green under this same removal, so this is exactly the blindness fixed;
- restored → 9 passed (9). eslint/prettier clean.

Kept the assertion Vite-only per your note — no StyleX-side symmetry (vega has no StyleX entry, themes are `/*`-only there; pre-existing state shouldn't go red here).

On the two notes: (a) the `@astryxdesign/build/vite` truly-clean-worktree gap is being written up as a separate issue (node ESM config-loader resolution of main.ts's own import — a Vite alias can't reach it); (b) agreed the #3744 cite was looser than presented — the conclusion (no changeset for the private app + root test config) stands, as you said.

imdreamrunner (approved):
🟢 **Approving** — the guard now catches the regression it exists for.

Re-verified `a0f1040` independently on a fresh clone, rather than just reading the diff:

| Check | Result |
| --- | --- |
| Guard as-is | 9 passed |
| Remove **only** the Vite `resolve.alias` richtext entry (the #5092 bug) | × `aliases @astryxdesign/richtext to source` — 1 failed \| 8 passed |
| Full `node` test project | 230 files / 3240 tests green |

I also checked the one thing that could have made the slice quietly wrong: `mainTs.indexOf('alias: {')` lands on line 124 — Vite's `resolve.alias`. The StyleX block at line 82 is `aliases: {`, which that search string cannot match, and those are the only two candidates in the file. So the scoping is sound by construction, not by luck. `main.ts` is untouched by this commit, so my earlier end-to-end result (Storybook boots with `packages/richtext/dist` absent) still holds.

One consequence worth recording, since it follows from the scoping I asked for: removing the StyleX `aliases` richtext entries alone leaves the suite green — I tested that too. That is the deliberate tradeoff (`@astryxdesign/vega` has a Vite alias and no StyleX entry, so a symmetric assertion would go red on pre-existing state). The StyleX side stays unguarded; only worth revisiting if that vega/theme asymmetry is ever cleaned up. Not this PR's problem.

Thanks for the careful fix and for the two-way red/green in your reply — it matched mine exactly. Nice contribution: the root-cause writeup set the bar for how to report an infra bug.

Please do file the `@astryxdesign/build` follow-up when you get a chance — it's the same class of bug and it's what a truly clean worktree hits first.

_Note: `build-sandbox`, `smoke-test`, and `test` were still running when I approved; `lint`, `build-storybook`, `check-scope`, `check-components`, `docsite-test`, and `dependency-check` are green. My local run covers the `test` job's content, but let CI finish before merging._

imdreamrunner:
Filed the two follow-ups so they do not get lost:

- #5128 — the `@astryxdesign/build/vite` cold-clone failure you flagged as out of scope. Confirmed independently; labeled `good first issue`. It also blocks the stronger resolution-based guard here (importing `main.ts` in a fresh worktree fails on exactly that error), which is why the static text assertion was the right shape for this PR.
- #5129 — the underlying class: the alias list is maintained twice by hand, which is what made #5092 possible. Notes that the StyleX/Vite asymmetries are legitimate (vega has no StyleX files; themes need a different entry per consumer), and that `main.test.ts` would need reworking alongside it.

Neither is yours to pick up — this PR is done. Thanks again.

PR #5101, opened 2026-08-16
fix(cli): swizzle recursively copies nested component subdirectories

### LINKED ISSUE #3506: swizzle silently drops nested subdirectories, producing a broken partial eject (e.g. Table/plugins)

jiunshinn:
Closes #3506

**Clean-room provenance: implemented from the issue description and source only; no code from the prior closed PRs (#3507, #4347, #4445) was opened, read, or consulted.**

`astryx swizzle Table` previously copied only 17 top-level files while its barrel still referenced missing `plugins/*` modules. It now copies all 47 source files with their directory structure intact. A single file set drives copying, overwrite checks, and the receipt. Import rewriting accounts for each file's depth, preserving intra-component relative imports while rewriting escaping imports to the owner package.

Before creating directories or writing files, swizzle checks every destination path segment below the project root and rejects symlinks, including dangling file links. `--overwrite` does not bypass this check. Regression tests cover output-root, component-root, nested-directory, existing-file, dangling-file, and in-project symlinks, and verify that rejection leaves no partial output.

Recursive traversal is limited to core component directories. Integration discovery identifies a same-stem entry file, not an exclusively owned directory tree, so integrations retain their previous flat copy behavior. A shared-directory regression confirms that swizzling Alpha does not newly copy unrelated nested files or overwrite consumer edits there. The pre-existing copying of top-level sibling integration files remains unchanged; this PR does not define a new integration ownership policy.

Validation:

- Focused swizzle and routing suite: **42 passed**.
- CLI strict typechecking: **passed**.
- `pnpm lint:strict`, including repository checks: **passed**.
- `pnpm build`: **passed**.
- Full `pnpm test`: **10,379 passed, 2 failed**. Both failures involve case-sensitive component/hook filename expectations on this local filesystem. Their failing behaviors also reproduce on untouched head `fd227184`: lowercase `button` resolves instead of throwing, and `mediaquery` resolves with `useMediaquery.doc.mjs` casing.

### DIFF
--- a/.changeset/swizzle-recursively-copies-nested-component-subd-la5j.md
+++ b/.changeset/swizzle-recursively-copies-nested-component-subd-la5j.md
@@ -1,0 +1,9 @@
+---
+'@astryxdesign/cli': patch
+---
+
+[fix] `swizzle` now recursively copies a component's nested source subdirectories (#3506). The copy loop read the component directory non-recursively and skipped anything that was not a file, so a component whose source is split across subfolders ejected broken: `swizzle Table` reported 17 files copied while the entry barrel still re-exported 15+ modules from a `plugins/` directory that was never written, failing module resolution (TS2307) on first build. The overwrite pre-flight and the reported file list shared the same blind spot, so consumer edits inside a nested path were clobbered without the usual `ERR_FILE_EXISTS` refusal and the receipt never mentioned nested files.
+
+One recursive walk now feeds all three: the pre-flight conflict check, the copy, and the receipt's `files` list (paths relative to the output directory, e.g. `plugins/selection/index.ts`). Import rewriting is depth-aware: a nested file's `../` chain is only rewritten to the owner package once it climbs past the component root, so `plugins/selection/x.tsx` keeps `../../types` relative while `../../../Icon` still becomes `@astryxdesign/core/Icon`. Directories are created per copied file, so a directory whose contents are all excluded (`__tests__/`, `__snapshots__/`) never appears in the output. `swizzle Table` now ejects 47 files (17 top-level + 30 nested) and every relative import in the output resolves.
+
+@jiunshinn
--- a/.changeset/swizzle-recursively-copies-nested-component-subd-la5j.md
+++ b/.changeset/swizzle-recursively-copies-nested-component-subd-la5j.md
@@ -2,8 +2,8 @@
 '@astryxdesign/cli': patch
 ---
 
-[fix] `swizzle` now recursively copies a component's nested source subdirectories (#3506). The copy loop read the component directory non-recursively and skipped anything that was not a file, so a component whose source is split across subfolders ejected broken: `swizzle Table` reported 17 files copied while the entry barrel still re-exported 15+ modules from a `plugins/` directory that was never written, failing module resolution (TS2307) on first build. The overwrite pre-flight and the reported file list shared the same blind spot, so consumer edits inside a nested path were clobbered without the usual `ERR_FILE_EXISTS` refusal and the receipt never mentioned nested files.
+[fix] `swizzle` recursively copies nested core component source directories, preserving relative imports inside the copied tree and rewriting imports that escape it to the owner package. Components such as Table now include their plugins; overwrite checks and the returned file list use the same complete set of relative paths (#3506).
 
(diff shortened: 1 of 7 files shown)


cixzhang (dismissed):
Thanks for fixing nested component copies. Two boundaries still block this head:

- A nested symlink already present under the destination is followed, so `swizzle Table` can write through `Table/plugins` outside the requested output tree. Please reject symlinks in every destination path segment before any write and add a regression for that case.
- Integration components can share one source directory. Recursing from that directory means swizzling one component also copies sibling components and shared files (for example, `Alpha` also copies `Beta.tsx`). Please derive the selected component's exact owned file boundary rather than treating the containing directory as the component root, with a multi-component integration regression.

The focused and current-main synthetic-merge tests pass, but they do not cover these two cases.

jiunshinn:
@cixzhang Could you review the updated head, 65b265b05a432a48d642bb62f8d43a5514ead86f?

- Swizzle now rejects destination symlinks at every segment below cwd, including dangling file links, before any writes; `--overwrite` cannot bypass the check. Regressions verify that rejection leaves no partial output.
- Recursive copying is limited to core component directories. Integrations retain their previous flat behavior, with a regression proving that unrelated nested files and consumer edits are left alone. The pre-existing top-level sibling-copy behavior remains unchanged while the integration ownership decision is unresolved.

Validation: 42 focused tests pass, as do CLI strict typechecking, strict lint, repository checks, and the full package build. The full test suite has 10,379 passes and two filename-case failures whose behaviors also reproduce on the untouched prior head. CI is running on this update.

PR #5103, opened 2026-08-16
fix(cli): expose built theme variants in component docs

### LINKED ISSUE #5059: [Bug] Built themes never carry runtime `variants`, so `astryx component`'s custom-variant `*` annotation can never fire
### Description

`astryx component <Name>` annotates theme-provided variants with a `*` and a footnote, reading them off the resolved theme:

```js
// clients/cli/lib/component-format.mjs
const themeVariants = themeData?.variants?.[componentKey] || [];   // L160
...
if (themeData?.variants) {                                         // L264
  sections.push(`_\\* = custom variant from ${themeData.name || 'active'} theme_\n`);
}
...
const themeVars = themeData?.variants?.[componentKey];             // L553
```

But `astryx theme build` never emits a runtime `variants` key. `generateBuiltModule` in `api/theme/build/build.mjs` produces exactly:

```js
export const XTheme = {
  name: '...',
  __built: true,
  tokens: { ... },
  icons: ...,     // only when an icon import is detected
};
```

`resolveTheme` loads that module and returns it as `themeData`, so `themeData.variants` is `undefined` for **every** built theme — including the seven published ones. The `*` suffix and the footnote appear to be unreachable.

### Reproduction

Custom variants are declared as a **type-only** augmentation in the generated `.variants.d.ts`:

```ts
declare module '@astryxdesign/core/Badge' {
  interface BadgeVariantMap { 'gray': true; }
}
```

That works — `const v: BadgeVariant = "gray"` typechecks, while a bogus value errors. But nothing carries it to the CLI:

```bash
# package.json → "astryx": { "theme": "@astryxdesign/theme-neutral" }   (resolves cleanly, no warning)
$ astryx component Badge | grep -E "gray|custom variant from"
# (no output)
```

The props table lists the 14 variants from core and omits `gray`, with no `*` and no footnote. Checking the shipped artifact directly:

```bash
$ node -e "const t=require('@astryxdesign/theme-neutral/dist/neutral.js'); console.log(Object.keys(t.neutralTheme))"
name, __built, tokens, icons
```

So a theme variant that the compiler accepts and that renders correctly is undiscoverable through the CLI — which is the tool the agent-facing docs direct you to consult instead of relying on recalled prop lists.

### Suggested

Either emit the custom variants into the built module alongside `tokens` (they are already computed at build time to generate `.variants.d.ts`, so the data is in hand), or drop the merge/footnote from `component-format.mjs` so the output does not imply a capability that cannot fire.

The first seems clearly preferable: theme-defined variants are exactly the kind of thing an agent cannot guess and the type system already knows.

### Environment

- Behaviourally reproduced on `@astryxdesign/{cli,core,theme-neutral}` 0.3.0.
- Verified still present in 0.4.1 by reading the published source: `generateBuiltModule` emits no `variants`, and the four `themeData?.variants` reads in `component-format.mjs` are unchanged.
- Node v24.18.0, Linux, pnpm 10.33.0.

jiunshinn:
Closes #5059

## User impact

Builders using custom theme values could not discover them through `astryx component`: the theme's generated TypeScript declarations accepted those values, but its built JavaScript omitted the metadata used by the CLI's `*` annotation. The issue's package-name configuration also loads a source export, which does not carry build metadata.

## Problem and solution fit

Collect custom values once using main's existing classification across root, `onDark`, and `onLight`, then generate both `.variants.d.ts` and runtime `variants` from those entries. This keeps discoverable values aligned with the public types, including custom Heading types.

For package-root configuration, enrich missing variant metadata from the package's matching public `/built` export. The selected source remains authoritative for its name, fonts, and explicit metadata; missing or unrelated companions are ignored. File paths and explicit subpaths retain their selected module.

## Expected behavior and authority

- #5059 describes the existing custom-value annotation promise.
- [AST-024](https://github.com/facebook/astryx/blob/main/docs/specs/AST-024/spec.md) requires custom Heading types to participate in generated augmentation.
- [AST-012](https://github.com/facebook/astryx/blob/main/docs/specs/AST-012/spec.md) owns shared value classification and the restriction against conditional-only enrollment. The existing classifier and adaptation validation are preserved.

## Smallest restoration

The change adds the runtime metadata already consumed by the CLI and connects the package-name path from the reproduction. The existing component-to-values shape is preserved. Built-in values and closed prop axes remain excluded; themes without custom values omit the field. Core behavior, styles, and public props are unchanged.

Rebased onto current `main`, preserving its newer Heading support and theme compilation behavior.

## Evidence

- Before: package-root discovery returned `variants: null`; four new resolver regressions failed before the fix.
- After: with `package.json` containing `astryx.theme: "@astryxdesign/theme-neutral"`, the rebuilt theme prints `gray*` and the neutral-theme footnote. `neutral` shorthand and the explicit `/built` path also pass in full and brief output.
- Regression coverage includes Heading `hero`, declaration/runtime agreement, root/dark/light deduplication, surface-only values, built-theme inheritance, closed props, old artifacts, and source metadata precedence.
- Existing adaptation-enrollment tests verify that conditional rules still cannot independently introduce custom values.

## Scope

- [x] One defect is restored.
- [x] Separable contradictory or unsettled tagalongs were removed or split.
- [x] No new public API, visual representation, default, or interaction model is hidden under the bug-fix label.

## Testing

- `pnpm build` — passed.
- `pnpm lint:strict` — passed, including repository/knowledge checks.
- `pnpm -F @astryxdesign/cli typecheck:strict` — passed.
- `pnpm exec vitest run --project node packages/cli --maxWorkers=4` from the repository root — **254 files, 3,977 tests passed**.
- `pnpm test --project ui --maxWorkers=3` — **401 files, 10,192 tests passed**.
- Focused theme builder/resolver/formatter and neighboring command tests — **13 files, 263 tests passed** (included in the CLI count).
- Seven repository-tooling suites requiring local HTTP/IPC servers or npm fixture installs — **157 tests passed** on isolated rerun.

The broad sandboxed run hits local-server/package-cache restrictions. Scoped reruns clear those failures for the CLI and the seven tooling suites above. An unrelated workspace-copy fixture still fails on macOS because it invokes GNU `cp --reflink=auto` (unsupported by the system `cp`); the isolated rerun reports 7 passed, 2 failed, and 6 skipped tests in that unchanged suite. No test or production behavior was weakened to hide this limitation.


cixzhang (changes requested):
Builders using current custom Heading types need those values to remain discoverable. This head conflicts with main, and its test still expects `heading.type:hero` to be excluded even though Heading is now intentionally extensible. Please rebase, preserve main’s shared value classification across root/`onDark`/`onLight`, derive declarations and runtime metadata from that same set, and update the stale expectation.

[Reviewed by Robohands]

PR #5104, opened 2026-08-16
fix(cli): preserve theme icon imports and inheritance

### LINKED ISSUE #5058: [Bug] `theme build` detects the icons import with a regex over raw source, so a commented-out import is re-emitted as code (and an inlined registry is silently dropped)
### Description

`extractIconInfo` in `api/theme/build/build.mjs` locates a theme's icon registry by running two regexes over the **raw source text**, with no awareness of comments or strings:

```js
const iconsMatch = content.match(/icons:\s*([a-zA-Z_][a-zA-Z0-9_]*)/);
const importRegex = new RegExp(
  `import\\s*{[^}]*\\b${varName}\\b[^}]*}\\s*from\\s*['"]([^'"]+)['"]`,
);
```

The matched `importPath` is then copied verbatim into the generated module. So an import statement that appears **inside a comment** is treated as real and re-emitted as executable code.

This is not hypothetical. I inlined an icon registry into a theme source and wrote a block comment explaining why, quoting the old import line for context:

```ts
/**
 * The scaffold put this in a sibling `icons.tsx`, so the built module carried
 * `import { neutralIconRegistry } from './icons'` — which Node cannot resolve.
 */
const neutralIconRegistry: IconRegistry = { /* … */ };
```

`astryx theme build` then emitted that exact import into `neutral.js`, pointing at a file that no longer existed. Deleting the sentence from the comment fixed the build output.

### Second half: inlining silently drops `icons`

The same function is the only path by which a registry reaches the output. `generateBuiltModule` gates on it:

```js
const iconsField = iconInfo ? `  icons: ${iconInfo.exportName},` : '';
```

So a theme that defines its registry inline — no import to find — builds successfully and produces a module with **no `icons` key at all**. The icons are gone from the runtime theme, and the build reports success. A theme author who reasonably assumes "a local const works like an imported one" loses their whole registry with no diagnostic.

### Suggested

Parse rather than scan. If a full parse is unwanted here, stripping comments before matching removes the false positive cheaply, and the inline case could be handled either by supporting a local declaration or by warning when `icons:` names a binding that is not imported — the silent-drop is the more damaging half.

### Environment

- Reproduced on `@astryxdesign/cli` 0.3.0; **`extractIconInfo` is unchanged in 0.4.1** (same body, same two regexes), so this is current.
- Node v24.18.0, Linux, pnpm 10.33.0.
- Unrelated to #4620 (extensionless specifier, fixed in 0.4.0) and #4621 (docs) — this is about how the import is *detected*, not how it is rendered.

jiunshinn:
Closes #5058.

## User impact

`astryx theme build` could emit an import copied from a comment or string, or omit a local icon registry while reporting success. The previous rejection also blocked themes that inherited icons through `extends`. Generated modules now preserve real imported and inherited registries, and unsupported registries fail before generation or writes in both build and `--check` modes.

## Problem and solution fit

Icon detection used text matches detached from the theme the loader selected. The compiler now uses the existing parser dependency to follow that selected export, its actual import bindings, and local theme inheritance. Named, aliased, default, and namespace imports retain their meaning. Base icons merge before child entries, relative leaf imports follow the loader's source-first resolution, and generated binding names avoid collisions.

The resolver rejects inline entries and unprovable or mutated bindings with `ERR_THEME_INVALID`. Imported raw configurations that cannot be traced are rejected rather than packaged incompletely. The legacy loader fallback no longer erases unresolved icon references. `--icons-specifier` targets the selected registry binding, including when a base and child share an import module. Font diagnostics remain notices.

## Expected behavior and authority

[Theme authoring INV3](https://github.com/facebook/astryx/blob/main/docs/architecture/theme-authoring-contract.md) requires inheritance to carry icons into the child. [Theme compilation](https://github.com/facebook/astryx/blob/main/docs/architecture/theme-compilation.md) owns packaging of the resolved theme. This restores that behavior without changing public options or the icon registry format.

## Evidence

- Before: a quoted `ghost-icons.mjs` import produced a successful build and an up-to-date check, but native Node failed to load the generated module. Source and built base inheritance failed with `ERR_THEME_INVALID`.
- After: 72 focused tests pass. They load generated ESM with native Node and verify React elements, multi-generation inheritance, child precedence, import aliases, source-first resolution, and preserved output files on rejection.
- The two previously failing canonical host-theme browser tests pass.
- The ordinary direct named registry import retains its generated import and re-export.
- Additional regression coverage rejects string/template/regex decoys, unrelated theme declarations, type-only imports, destructured mutation, opaque raw package configurations, and missing real imports.

## Scope

- [x] One defect is restored: icon provenance must survive theme compilation or fail before output.
- [x] Supporting changes are regression coverage, API documentation, and a CLI changeset.
- [x] No new public option, visual behavior, or registry format.

## Testing

- `pnpm build` passed.
- `pnpm lint:strict` passed.
- CLI `typecheck:strict` and `typecheck:json-api` passed.
- Focused theme-build suites: 72 passed.
- Canonical browser reproductions: 2 passed.
- `pnpm test --maxWorkers=4`: 17,019 passed. The unchanged `internal/vibe-tests/setup-test/setup-workspace.test.mjs` is the only failing file: its fixture setup invokes GNU `cp --reflink=auto`, which macOS `cp` rejects (one suite setup failure and two test failures).
- CLI smoke checks for normal, `--check`, `--json`, and `--check --json` exit 1 without artifacts for the original reproduction; JSON returns `ERR_THEME_INVALID`.

GitHub CI on `e8e0e2854ea` has passed the Linux Node and UI test lanes, both builds, registry-contract, and fixture-contrast ([CI run](https://github.com/facebook/astryx/actions/runs/34729410581)). Lint and CLI smoke checks also passed.

### DIFF
--- a/.changeset/theme-build-icon-detection-no-longer-reads-comme-knch.md
+++ b/.changeset/theme-build-icon-detection-no-longer-reads-comme-knch.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/cli': patch
+---
+
+[fix] `theme build` icon detection no longer reads comments as code, and an inline registry is dropped loudly instead of silently. The icon-registry scan ran two regexes over the raw theme source, so a doc comment that quoted an old import line (`import { icons } from './icons'`) was scraped and re-emitted as executable code in the generated module, pointing at a file that may not exist (#5058). The scan now blanks line and block comments (string contents untouched) before matching. And a theme whose `icons:` names a binding that no import backs — a registry declared inline as a const — used to build "successfully" into a module with no `icons` key at all: the registry vanished with no diagnostic. It still cannot ride along (a registry holds React elements, which do not serialize), but the build now says so, in the CLI output and in the receipt's `warnings`, and tells the author to move the registry to its own module.
+
+@jiunshinn
--- a/.changeset/theme-build-icon-detection-no-longer-reads-comme-knch.md
+++ b/.changeset/theme-build-icon-detection-no-longer-reads-comme-knch.md
@@ -2,6 +2,6 @@
 '@astryxdesign/cli': patch
 ---
 
-[fix] `theme build` ignores comments when detecting icon imports and fails with `ERR_THEME_INVALID` when an inline registry cannot be preserved. Both normal builds and `--check` reject unsupported registries before writing output files, preventing successful builds that lose custom icons. Move an inline registry into its own module and import it into the theme file. Imported registries continue to build as before.
+[fix] `theme build` resolves real icon imports from the selected theme instead of matching comment or string contents. Generated modules preserve named, aliased, default, and namespace registry imports, plus inherited icons with child overrides. Normal builds and `--check` reject unsupported inline registries with `ERR_THEME_INVALID` before generating or writing output. Move such a registry into its own module and import it into the theme file.
 
 @jiunshinn
--- a/.changeset/theme-build-icon-detection-no-longer-reads-comme-knch.md
+++ b/.changeset/theme-build-icon-detection-no-longer-reads-comme-knch.md
@@ -2,6 +2,6 @@
 '@astryxdesign/cli': patch
 ---
 
-[fix] `theme build` icon detection no longer reads comments as code, and an inline registry is dropped loudly instead of silently. The icon-registry scan ran two regexes over the raw theme source, so a doc comment that quoted an old import line (`import { icons } from './icons'`) was scraped and re-emitted as executable code in the generated module, pointing at a file that may not exist (#5058). The scan now blanks line and block comments (string contents untouched) before matching. And a theme whose `icons:` names a binding that no import backs — a registry declared inline as a const — used to build "successfully" into a module with no `icons` key at all: the registry vanished with no diagnostic. It still cannot ride along (a registry holds React elements, which do not serialize), but the build now says so, in the CLI output and in the receipt's `warnings`, and tells the author to move the registry to its own module.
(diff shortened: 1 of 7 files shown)


cixzhang (changes requested):
Theme authors can still get a successful build and committed generated files with every custom icon removed. Please fail before writing when an inline registry cannot be preserved; `--check` must fail too. The comment-filtering fix looks good.

Also keep main's current font-diagnostic wording when rebasing: unloaded fonts are notices, not warnings.

[Reviewed by Robohands]

jiunshinn:
@cixzhang Addressed the requested changes in 1d8a14f9c6f6e829feed296d471451752c78317b. Unsupported inline icon registries now fail with `ERR_THEME_INVALID` before output generation or writes, including in `--check` when committed output already omits the registry. Regression coverage checks fresh builds, preservation of existing artifacts, and the no-CSS path. The branch is rebased onto main and retains font diagnostics as notices.

Build, strict lint, CLI type checks, and 600 tests under Node 24 passed. The PR description records the unrelated macOS test-fixture limitation from the full suite.

Please re-review the updated head. The repository workflow has also re-requested your review.

cixzhang (changes requested):
Thanks for moving the rejection ahead of writes; my previous review asked for inline registries to fail there. Builders can still get a broken generated module or no build at all. A string containing `import { inlineIcons } from './ghost-icons.mjs'` lets a real inline registry emit that nonexistent import. A derived theme such as `extends: neutralTheme` has inherited icons but no local `icons:` field, so it now fails with `ERR_THEME_INVALID`; the required `fixture-contrast` job reproduces both canonical host-theme failures. Please make detection follow real imported/inherited registries, not comments or string contents: preserve imported/inherited icons, and reject actual inline registries before generation in normal and `--check` modes.

[Reviewed by Robohands]

PR #5107, opened 2026-08-16
fix(Selector, MultiSelector): hide standalone divider from a11y tree

### LINKED ISSUE #4994: Selector and MultiSelector put a role=separator Divider inside role=listbox (critical axe violation)
## What

`role="listbox"` only permits `option` and `group` children, but both selectors push a `Divider` — which renders `role="separator"` — straight into their listbox:

- `MultiSelector` renders one above the options when `hasSelectAll` is set, and one per `{type: 'divider'}` / section option.
- `Selector` does the same for `{type: 'divider'}` and section options. (Its "no results" branch already avoids this — it renders `role="presentation"` with a comment saying listbox only permits option/group children, so the constraint is known; the dividers just predate it.)

axe rates it **critical** (`aria-required-children`, WCAG 2 A / 1.3.1). A screen reader can drop or mis-report the listbox's children when an unallowed role appears among them.

```
Element has children which are not allowed: [role=separator]
<div id="…" role="listbox" aria-multiselectable="true" aria-labelledby="…">
  <div role="separator" aria-orientation="horizontal" class="astryx-divider subtle horizontal" …>
```

## Why we haven't seen it

The a11y audit never opens a closed popup, and until [#4993](https://github.com/facebook/astryx/pull/4993) no `MultiSelector` story used `isDefaultOpen` — so nothing in CI had ever looked inside an open listbox. The violation surfaced the moment a story opened one with `hasSelectAll` on. Anything that only exists inside an open popover is invisible to the audit today; worth keeping in mind beyond this bug.

To reproduce: any `MultiSelector` story with `hasSelectAll` + `isDefaultOpen`, then `pnpm a11y:audit -- --components MultiSelector`.

## Fix direction

The dividers are decorative — they carry no information the options don't. Making them presentational inside a listbox (`role="none"`, or a variant of `Divider` that renders no role) keeps the visual and removes the unallowed child. Sections are the case to think about: a titled section already renders a `role="group"`, so its heading divider may want to move inside the group or become the group's `aria-label` source rather than sit beside it.

Both components share the pattern, so they should be fixed together.

HelloOjasMutreja:
## Summary

Closes #4994.

`Selector` and `MultiSelector` both pushed a `Divider` (which renders `role="separator"`) directly into their `role="listbox"` for `{type: 'divider'}` options. `role="listbox"` only permits `option`/`group` children, and axe rates this **critical** (`aria-required-children`, WCAG 2 A / 1.3.1): a screen reader can drop or mis-report the listbox's children when an unallowed role appears among them.

## What I found reading the current source

Sections already avoid this: `renders a section title as a plain heading inside the group, not a divider` in both test files confirms `role="group"` with `aria-label={option.title}` and an `aria-hidden="true"` heading are already in place, with a comment explaining exactly this constraint. The standalone divider (for plain `{type: 'divider'}` options, not sections) predates that fix and was missed.

The issue also described a divider rendered above the options when `MultiSelector`'s `hasSelectAll` is set. Checked `MultiSelector.tsx`'s current `renderOptions`: that divider doesn't exist in current source, its own comment says "no divider under it" and there's a dedicated regression test (`does not draw a divider under select-all`) already passing. Nothing left to fix there; the issue's description had gone stale on that part.

## Fix

Since a standalone divider (unlike a section) carries no information the options don't, the fix follows the exact same pattern already used for the section heading a few lines above: `aria-hidden="true"` on the `Divider` instance. The element keeps its visual and its `role="separator"` in the DOM (unchanged rendering), but is excluded from the accessibility tree entirely, which resolves the axe violation.

`Divider` already forwards unrecognized props (including `aria-hidden`) straight to its root via `{...props}`, so no change to `Divider` itself was needed, and this isn't new API surface.

## Test notes

Added a regression test to both `Selector.test.tsx` and `MultiSelector.test.tsx` asserting the divider carries `aria-hidden="true"`. Verified the test fails without the fix (`aria-hidden` attribute is `null`) and passes with it restored. Full `Selector.test.tsx`, `MultiSelector.test.tsx`, and `Divider.test.tsx` suites (244 tests) pass.

## Test plan

- [x] `vitest run` on `Selector.test.tsx` + `MultiSelector.test.tsx` + `Divider.test.tsx`: 244/244 passing
- [x] Verified new tests fail without the fix, pass with it
- [x] `eslint` clean on touched files (one pre-existing, unrelated warning in `MultiSelector.tsx` confirmed present before this change)
- [x] Changeset added

### DIFF
--- a/.changeset/selector-divider-listbox-violation.md
+++ b/.changeset/selector-divider-listbox-violation.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] `Selector` and `MultiSelector` no longer expose their `{type: 'divider'}` separators to assistive technology. `role="listbox"` only permits `option`/`group` children, but the divider previously rendered `role="separator"` as a direct child of the listbox (axe `aria-required-children`, impact critical). The divider is decorative and carries no information the options don't, so it's now hidden from the accessibility tree via `aria-hidden`, matching the pattern already used for section headings.
+
+@HelloOjasMutreja
--- a/packages/core/src/MultiSelector/MultiSelector.test.tsx
+++ b/packages/core/src/MultiSelector/MultiSelector.test.tsx
@@ -1837,6 +1837,27 @@ describe('MultiSelector list structure', () => {
         Node.DOCUMENT_POSITION_FOLLOWING,
     ).toBeTruthy();
   });
+
+  it('hides a standalone divider from the accessibility tree (#4994)', async () => {
+    const user = userEvent.setup();
+    render(
+      <MultiSelector
+        label="Fruit"
+        options={['Apple', {type: 'divider'}, 'Banana']}
+        value={[]}
+        onChange={() => {}}
+      />,
+    );
+    await user.click(screen.getByRole('combobox'));
+
+    // role="listbox" only permits option/group children. The divider still
+    // renders role="separator" (unchanged visual/DOM), but must be excluded
+    // from the accessibility tree so it never reaches the listbox's exposed
+    // children (axe aria-required-children).
+    const divider = document.querySelector('[role="separator"]');
+    expect(divider).toBeTruthy();
+    expect(divider).toHaveAttribute('aria-hidden', 'true');
+  });
 });
 
 describe('MultiSelector search affordances', () => {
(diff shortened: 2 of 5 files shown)


cixzhang (approved):
Thanks for this, and sorry it sat. Nice call reusing the section-heading pattern instead of touching Divider — the open listbox measures clean in Chromium. Merging.

[Reviewed by Robohands]

PR #5108, opened 2026-08-16
chore: version packages for v0.4.3

cixzhang:
Version bump for release 0.4.3 (from 0.4.2). All 10 publishable packages move together per the `fixed` group; 2 changesets consumed.

**Patch** — no breaking changes, so no codemod and no new codemod tier.

Two entries: the `useContainerReveal` hover-intent/forced-state API ([#5084](https://github.com/facebook/astryx/pull/5084)) and the `extends` inheritance fix ([#5067](https://github.com/facebook/astryx/pull/5067)).

**Note on [#5084](https://github.com/facebook/astryx/pull/5084):** its code already shipped inside the 0.4.2 tarballs — 0.4.2 was dispatched against `main` rather than its version-bump commit, and #5084 had landed in between. Its changelog entry appears here, one version later than the code. The v0.4.2 release notes carry a note saying so. Going forward the dispatch and tag are pinned to the bump commit, so this cannot recur.

First cut under the new cadence: public releases daily, internal `xds-meta`/`example_meta_app` sync twice weekly to whatever is latest.

Lockfile refreshed and verified with `--frozen-lockfile`.

## Test plan

`pnpm install --frozen-lockfile` passes.

CHANGELOGs reviewed against both consumed changesets — full bodies survived, categories in canonical order, contributor credited.


cixzhang:
Held as a draft until the 10-11am PT release window (Cindy, 2026-08-16) — releases are cut in that window, not ad hoc. The daily 10:00 fire will re-verify against `main`, re-cut if anything landed in the meantime, and mark this ready.

PR #5109, opened 2026-08-16
Document the seven core hooks missing from the CLI hook index, and fail the build when it goes stale

cixzhang:
## Why

The CLI's hook index is built from the `.doc.mjs` file sitting next to each hook, and seven hooks exported from `@astryxdesign/core/hooks` shipped without one. Nothing about adding a hook forces a doc, so the index quietly went stale: `astryx hook useAnnounce` answered `Error: No hook named "useAnnounce"`, bare `astryx hook` omitted them, and `astryx search` never returned them, while the package exported them with full TSDoc.

The cost is not "a hook is undocumented"; it is that an agent following the documented discover-don't-guess workflow concludes the primitive does not exist and writes its own. `useAnnounce` is the sharpest case: an accessibility review searched for a live-region primitive, found only the hand-rolled `VisuallyHiddenLiveRegion` template, and was about to recommend a hand-built `role="status"`. That is exactly the bug `useAnnounce` exists to avoid, since screen readers largely ignore a live region created together with its content. `useTypeahead` loses the same way: `useListFocus` is discoverable and points at it, so the CLI documented one half of a pair and hid the other.

## What

- A `.doc.mjs` for each of the seven undocumented exported hooks: `useAnnounce`, `useTypeahead`, `useInteractiveRole`, `useLongPress`, `useInputStatusIcon`, `useDevWarning`, and `useIndicatorFocusRing`. Params, returns, keywords, related components/hooks, and dense overlays follow the existing hook docs, and the guidance is drawn from each hook's own source.
- A coverage gate in the discovery suite: every `use*` the hooks barrel actually exports must appear in the discovered index, with a failure message naming the missing hooks and the fix. The stale index is the actual bug, so this is the part that matters.
- The gate is derived, not pattern-matched. It imports the built barrel (`packages/core/dist/hooks/index.js`) and reads `Object.keys`, rather than parsing `index.ts`. Importing sees every export form the language allows, and type-only exports are already erased by the compiler. It reads `dist` because this suite runs in the `node` Vitest project, which has no StyleX transform; the project's setup already builds core before workers start.
- Component-owned hooks stay out of scope by construction: the gate walks barrel to index, so a hook documented in its own component directory and deliberately absent from `src/hooks/` is never treated as a defect.

## Risk

Low and additive. No runtime or component code changes; the new files are documentation the CLI reads at discovery time, and the only behavior change is that seven more hooks are now findable. `useAnnounce` introduces one new listing category, `accessibility` (categories are free-form strings, grouped dynamically). The new gate is scoped to the hooks barrel; hooks that belong to a component API are documented with that component and remain out of scope.

## Testing

- Re-verified on current `main`: all seven hooks are still exported, and all seven return `NOT FOUND` from hook discovery.
- On this branch, all seven resolve through `astryx hook <name> --json`; `astryx search announce --json` ranks `useAnnounce` first.
- The focused discovery suite passes 11/11. Removing all seven docs makes the new gate fail and name all seven; restoring them returns it to green.
- Core source and docs typechecks pass. CLI authoring and strict typechecks pass; the strict check was rerun after building its required theme dependency.
- Repository structural checks and the public-repository leak guard pass.


cixzhang:
Direction: keep the complete hook index and its export-to-discovery regression test. The fix matches the discovery problem directly.

cixzhang (commented):
Thanks, this is good. The seven docs match their hooks, and the export-derived check fails on the exact current gap and passes when restored.

[Reviewed by Robohands]

PR #5110, opened 2026-08-16
fix(NumberInput): reject non-numeric characters as you type

freddymeta:
## What

Blocks non-numeric characters in `NumberInput` as you type. Letters and stray symbols are ignored at the change handler; a well-formed number (and valid in-progress states like `-`, `1.`, `.5`, `1e`) types normally.

## The bug

`NumberInput` recently moved from `type="number"` to **`type="text"`** (to render formatted display values and use custom stepper buttons instead of the native spinner — #4896). That switch dropped the browser's built-in numeric filtering: `inputMode="numeric"/"decimal"` only hints the *mobile* keyboard and restricts nothing on a physical one, and no keystroke-level guard was added to replace it. So the field visibly accepted `abc` until blur reverted it.

The committed numeric value was never wrong (`parseNumberInput` still gated `onChange`), but the field *looked* like it accepted letters — a visible regression.

## The fix

Gate the change handler with a partial-number syntax check (`isPartialNumber`) so non-numeric text never becomes the field's value.

Deliberately **looser** than full parsing: it accepts in-progress states that aren't parseable on their own (`''`, `-`, `1.`, `.5`, `1e`, `1e-`). And it only blocks *non-numeric* text — a value that is a number but violates a constraint (e.g. `3.5` under `isIntegerOnly`) is still kept as pending input and surfaced as `aria-invalid`, exactly as before. Constraint feedback stays soft; only garbage is hard-blocked.

## Tests

- New `rejects non-numeric characters` block: alphabetic input, symbols mixed with digits, a well-formed number, valid in-progress states, and the constraint-violating-but-numeric case (must remain pending + invalid, not blocked).
- Full suite: **118 tests pass**, including the existing "preserves invalid pending text while focused" behavior.
- `build` ✓ · `tsc --noEmit` ✓ · ESLint ✓ · `check:repo` ✓.

## Relation to #4186

#4186 tracks a larger NumberInput refinement (which prescribes the `type="text"` move). This is the narrow regression fix for the current visible behavior; it doesn't preempt that work.


packages/core/src/NumberInput/NumberInput.tsx
```diff
@@ -649,6 +668,17 @@ export function NumberInput({
...
       }
       const newValue = e.target.value;
+
+      // Reject characters that can never form a number (letters, stray
+      // symbols). `type="text"` — adopted for formatted display — no longer
+      // does this for us the way `type="number"` did, so the field would
+      // otherwise visibly accept "abc". Values that are numeric but violate a
+      // constraint (min/max/integer-only) are NOT blocked here: they stay as
+      // pending input and surface as invalid, preserving that behavior.
+      if (!isPartialNumber(newValue)) {
```
cixzhang: Rejecting after the DOM edit throws the caret to the end; the next digit lands in the wrong place.

packages/core/src/NumberInput/NumberInput.tsx
```diff
@@ -389,6 +389,25 @@ export type NumberInputProps =
...
+ * that can never resolve to a number, e.g. letters or stray symbols.
+ *
+ * Constraint checks (min/max/integer-only) are intentionally NOT applied here:
+ * a value that merely violates a constraint is still a *number* and is kept as
+ * pending input so the field can surface it as invalid. Only non-numeric text
+ * is blocked — restoring the guarantee the native `type="number"` input gave
+ * before this control moved to `type="text"` for formatted display.
+ */
+function isPartialNumber(input: string): boolean {
+  return /^[-+]?(\d*\.?\d*)(e[-+]?\d*)?$/i.test(input);
```
cixzhang: `e`, `E`, `-e` and `.e` all pass this; `42 `, `1,234` and full-width `１２３` don't.

cixzhang (changes requested):
The hard part here is that this is a text field with a pending cycle, so dropping the whole value fights an edit the browser has already made.

In Chromium: value `123`, caret after the `1`, press `a` then `9` — it commits `1239`. On main you get `1a923`, flagged invalid, committing nothing. A visible typo becomes a silently wrong number.

It also eats valid input silently. `42 ` pasted from a spreadsheet commits 42 on main and does nothing here; same for `1,234` and full-width `１２３`. None reach the `aria-invalid` and "Invalid number" alert this field already has, so a screen reader gets nothing.

Ideally I'd like to keep that invalid-pending path rather than add a second, silent one. DateInput and TimeInput accept anything and reconcile on blur — do we want per-keystroke blocking here at all?

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks for coming back to this, and sorry it sat — the caret fix is right, and I checked the "as the native control does" claim rather than taking it: a real `type="number"` in the same page gives the same answers for `1,234`, `１２３` and `1,5`.

One thing to fix before it lands. With a Japanese IME in full-width mode, typing `123` leaves `112123123` in the field — no invalid flag, and it commits. On main you get `１２３`, flagged, committing nothing. The change handler normalizes and re-sets the value on every composition update, so React rewrites the field mid-preedit. Reproduced twice.

Guarding the normalize while `isComposing` is true, and running it once on `compositionend`, should close it.

[Reviewed by Robohands]

freddymeta:
Closing this — main has answered the question underneath it, and the other way round.

The first review asked whether we want per-keystroke blocking here at all, given DateInput and TimeInput accept anything and reconcile on blur. [#5450](https://github.com/facebook/astryx/pull/5450) and [#5459](https://github.com/facebook/astryx/pull/5459) settle it in the code: `parseNumberInput` now reads typed and pasted text under the field locale, with a bounded separator alphabet and digit values derived from `\p{Nd}`. Parse generously, reconcile, flag with `aria-invalid`. This PR blocks keystrokes, which is the other design.

All three failures from the review are fixed on main, and not by this branch:

| case | on main today |
|---|---|
| `42 ` and `1,234` silently eaten | parsed |
| full-width `１２３` | parsed |
| IME full-width committing `112123123` | the composing keydown is guarded in `NumberInput.tsx` |

Rebasing would mean re-litigating that decision against a parser that now handles the exact inputs this branch dropped.

Thanks for the two rounds on it — the caret finding and the IME repro were both right, and the IME one is the reason I looked at what main does instead.

PR #5113, opened 2026-08-16
Banner: name the dismiss control after the banner it closes

cixzhang:
## Why

A surface that shows one dismissable `Banner` per failed operation — upload invoice, delete report, … — gives a screen-reader user a run of buttons that all announce as "Dismiss, button". Nothing in the name says which banner each one closes, so the only way to tell them apart is to leave the button and explore the region around it. `isDismissable` is a boolean and there was no `dismissLabel`, so a product could not qualify the name; the app that hit this kept the generic name rather than lose the rest of what `Banner` provides.

## What

The dismiss button's accessible name now names its banner. When `title` is a string the name is "Dismiss {title}", from a new ICU message `@astryx.banner.dismissTitled`; a `title` that is a `ReactNode` still falls back to the bare "Dismiss", and the new optional `dismissLabel` prop overrides either.

Two choices worth stating:

**The default derives from `title`, rather than only landing behind a new prop.** The gap is the common case, and a prop-only fix leaves every existing stacked banner ambiguous until each product opts in. Deriving it fixes them all at once and matches what the rest of the system already does for the same problem — `Remove {label}` on Token, `Clear {label}` on Selector, `Remove {accessibleName}` on Thumbnail.

**It goes through the i18n catalog, not a hardcoded English default.** Banner's dismiss name is already `t('@astryx.banner.dismiss')`, and a name composed of a verb and a title is exactly the kind of string whose word order changes per locale, so it is a catalog message with a `{title}` placeholder and a translator note, in the shape the `{label}` keys already use. `dismissLabel`, being consumer-supplied, is passed through verbatim — the consumer translates it. This is independent of the separate `announce()` live-region translation work; no live-region string is touched here.

The visible tooltip still reads "Dismiss" unless `dismissLabel` is set: repeating a title the user can already see is noise, and the bare verb stays a prefix of the accessible name, so WCAG 2.5.3 still holds.

## Risk

Behavior change without an API break: any existing dismissable banner with a string title gets a longer accessible name. Nothing visible changes, and no prop or type is removed. Two in-repo assertions that matched the old exact name were updated, as was the Crowdin screenshot target that located the × by `aria-label="Dismiss"`.

**Rebased onto current `main` (`9ebd45f`).** Current main still gave every dismiss control the same `Dismiss` name. The rebase was clean; Banner's newer collapsible-content coverage contained one stale exact-name query, which now follows the title-derived label.

## Testing

- Focused Vitest: 68/68 Banner and edge-compensation tests pass, including a German fallback regression test.
- Real Chromium on `c8f297c8d07`: English covers stacked string titles, rich-title fallback, translated override, and tooltip/name parity. German exposes `Schließen Wartungshinweis` and `Schließen Zahlung fehlgeschlagen` in both the DOM and AX tree while the tooltip remains `Schließen`; no page errors.
- Pre-commit repository gates passed, including sync, package boundaries, changesets, CLI structure, and the 29-locale catalog consistency check.
- Stable visual regression passed before the locale-only follow-up. No manual images were added because existing rendered pixels did not change; gate 1's before/after German frames were byte-identical.
- Gate 2 confirmed the longer name also sits inside Banner's existing `role="alert"`. Screen-reader announcement of descendant names varies and was not automatable here; the possible repeated title is accepted as non-blocking because it adds context without removing access or focus.

### DIFF
--- a/.changeset/banner-dismiss-accessible-name.md
+++ b/.changeset/banner-dismiss-accessible-name.md
@@ -2,6 +2,6 @@
 '@astryxdesign/core': patch
 ---
 
-[fix] Banner names its dismiss control after the banner it closes. Stacked banners — one per failed operation, say — gave a screen-reader user a run of identical "Dismiss, button" nodes with nothing to tell them apart, and `isDismissable` being a boolean left no way to qualify the name. The accessible name is now "Dismiss {title}" whenever `title` is a string, from the new `@astryx.banner.dismissTitled` message, so the common case is fixed without any consumer opting in. The new `dismissLabel` prop overrides it for a `title` that is not plain text. The visible tooltip still reads "Dismiss".
+[fix] Banner names each dismiss control after its string title, so stacked banners no longer expose identical "Dismiss" buttons to screen readers. Rich titles retain the generic translated name unless the consumer supplies an already-translated `dismissLabel`; that override also labels the tooltip.
 
 @cixzhang
--- a/.changeset/banner-dismiss-accessible-name.md
+++ b/.changeset/banner-dismiss-accessible-name.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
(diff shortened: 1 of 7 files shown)


cixzhang:
Direction: derive the dismiss control name from a string title and keep the explicit translated `dismissLabel` override for rich titles. This gives the common case a correct default without closing the escape hatch.

cixzhang (commented):
Gate 2 passes after the locale fallback fix. String, rich-title, translated override, tooltip/name, and exact-head Chromium cases are verified. The possible live-region repetition is non-blocking and accepted.

[Reviewed by Robohands]

Issue #5074, opened 2026-08-15
InternationalizationProvider locale is ignored by calendar, date/time, Timestamp, and PowerSearch formatting


GabrielGoncalvesICE:
## Version

`@astryxdesign/core@0.4.1`

## Problem

Several components render or parse locale-sensitive values using hardcoded English strings or the host/browser locale instead of the locale supplied by `InternationalizationProvider`.

Observed paths:

- `useCalendarDays` hardcodes `Su`, `Mo`, `Tu`, etc.
- `DateInput` and `DateTimeInput` format committed dates and parse typed dates without the provider locale.
- shared plain-date/date-parser utilities construct `Intl.DateTimeFormat(undefined, ...)`.
- `Timestamp` formats absolute values and relative time using the host locale / English branches.
- `PowerSearch` formats absolute dates and numbers with the host locale and has hardcoded `items`, `date range`, and `filter(s)` copy.

This makes a tree wrapped in `InternationalizationProvider locale="es-ES"` produce English weekday/date output when the host machine is English. It also makes output differ between developer and CI machines.

## Expected

All locale-sensitive formatting, parsing, weekday labels, relative time, and component-owned count/range labels should follow the provider locale. Switching the provider locale should update output without relying on the OS/browser locale.

## Suggested shape

- Read `InternationalizationContext.locale` in the affected components.
- Pass that locale explicitly through shared date/number/parser helpers.
- Use `Intl.DateTimeFormat(locale, ...)`, `Intl.NumberFormat(locale, ...)`, and `Intl.RelativeTimeFormat(locale, ...)`.
- Move the remaining PowerSearch count/range labels to semantic catalog keys.

A downstream consumer currently carries a narrow `pnpm patch` for these paths with deterministic tests for `pt-BR`, `es-ES`, and `en-US`. We can help turn that patch into a PR if the approach matches the project direction.

HelloOjasMutreja:
Opened #5117, scoped to just Calendar's weekday header names (useCalendarDays hardcoded the English abbreviations regardless of the provider locale). Verified against a real InternationalizationProvider locale="es-ES" render.

This is a multi-part issue across several components (DateInput/DateTimeInput, Timestamp, PowerSearch, plus the shared date/parser utilities). Planning to work through the rest as separate PRs, one path at a time, rather than one large combined change.

HelloOjasMutreja:
[...]

PR #5117, opened 2026-08-16
fix(Calendar): localize weekday header names to the provider locale

HelloOjasMutreja:
## Summary

Part of #5074, which covers several locale-ignoring paths across `Calendar`, `DateInput`/`DateTimeInput`, `Timestamp`, and `PowerSearch`. This PR scopes to just `Calendar`'s weekday header names; the remaining paths need their own investigation and will follow as separate PRs.

## Evidence

`useCalendarDays.ts` hardcoded the weekday abbreviations:

```ts
const names = ['Su', 'Mo', 'Tu', 'We', 'Th', 'Fr', 'Sa'];
```

regardless of the ambient `InternationalizationProvider` locale. A tree wrapped in `locale="es-ES"` (or any non-English locale) still rendered English weekday headers.

## Fix

Read `InternationalizationContext.locale` inside the hook. English keeps its existing two-letter abbreviations as a deliberate compact convention (confirmed by existing tests asserting the exact `'Su'/'Mo'` strings). Switching everything to `Intl.DateTimeFormat`'s `'short'` format would give three letters (`Sun`, `Mon`, ...) and regress the overwhelmingly common default case for no reason. Every other locale now uses its own short weekday form via `Intl.DateTimeFormat`, the only correct-by-construction source for non-English abbreviations.

## Test notes

Added a regression test asserting Spanish locale (`es-ES`) produces the correct localized weekday headers (`dom`, `lun`, `mar`, `mié`, `jue`, `vie`, `sáb`). Verified the test fails without the fix (falls back to the hardcoded English strings) and passes with it restored. Full `Calendar.test.tsx` suite (116 tests) passes.

## Test plan

- [x] `vitest run` on `Calendar.test.tsx` + `dayCellUtils.test.ts`: 116/116 passing
- [x] Verified new test fails without the fix, passes with it
- [x] Changeset added

### DIFF
--- a/packages/core/src/Calendar/hooks/useCalendarDays.ts
+++ b/packages/core/src/Calendar/hooks/useCalendarDays.ts
@@ -12,7 +12,7 @@
  * - /packages/core/src/Calendar/hooks/index.ts
  */
 
-import {useMemo} from 'react';
+import {use, useMemo} from 'react';
 import type {DayOfWeek, ISODateString} from '../../utils/dateTypes';
 import {
   type PlainDate,
@@ -21,6 +21,7 @@ import {
   plainDateDayOfWeek,
   plainDateAddDays,
 } from '../../utils/plainDate';
+import {InternationalizationContext} from '../../i18n';
 
 /**
  * Represents a single day in the calendar grid.
@@ -84,6 +85,7 @@ export function useCalendarDays(
   options: UseCalendarDaysOptions,
 ): UseCalendarDaysReturn {
   const {year, month, weekStartsOn = 0, hasVariableRowCount = false} = options;
+  const {locale} = use(InternationalizationContext);
 
   // Calculate grid structure
   const gridInfo = useMemo(() => {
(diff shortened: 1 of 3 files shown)


packages/core/src/Calendar/hooks/useCalendarDays.ts
```diff
@@ -108,15 +110,35 @@ export function useCalendarDays(
...
   const dayNames = useMemo(() => {
-    const names = ['Su', 'Mo', 'Tu', 'We', 'Th', 'Fr', 'Sa'];
+    const baseLanguage = new Intl.Locale(locale).language;
+    const names =
+      baseLanguage === 'en'
+        ? ['Su', 'Mo', 'Tu', 'We', 'Th', 'Fr', 'Sa']
+        : (() => {
+            const formatter = new Intl.DateTimeFormat(locale, {
+              weekday: 'short',
+            });
```
nynexman4464, line 129: Nice to see this moving off the hardcoded English array. One correctness issue: the anchor dates are UTC midnight, but the formatter has no `timeZone`, so it converts them to the viewer's zone before finding the weekday. Under `TZ=America/Los_Angeles`, the Sunday anchor formats as Saturday and the entire header shifts one day; CI misses this because its runner is UTC. Could you add `timeZone: 'UTC'` and add a regression test that runs under a negative UTC offset?

packages/core/src/Calendar/hooks/useCalendarDays.ts
```diff
@@ -108,15 +110,35 @@ export function useCalendarDays(
...
+  // other locale uses its own short weekday form via Intl.DateTimeFormat,
+  // the only correct-by-construction source for non-English abbreviations —
+  // previously this array was hardcoded English regardless of the ambient
+  // InternationalizationProvider locale.
   const dayNames = useMemo(() => {
-    const names = ['Su', 'Mo', 'Tu', 'We', 'Th', 'Fr', 'Sa'];
+    const baseLanguage = new Intl.Locale(locale).language;
+    const names =
+      baseLanguage === 'en'
+        ? ['Su', 'Mo', 'Tu', 'We', 'Th', 'Fr', 'Sa']
```
nynexman4464, line 125: I agree that we should preserve a compact header, but I don't think an English-only branch is the right long-term model.

`Su / Mo / Tu / …` is CLDR's `stand-alone.short` width. Unfortunately, `Intl.DateTimeFormat` does not expose that width: `weekday: 'short'` gives CLDR's abbreviated form (`Sun / Mon / …`), while `weekday: 'narrow'` gives `S / M / T / …`.

`short` also does not guarantee a compact result across locales. In current CLDR, Arabic's `short` values are the full weekday names, and Hebrew or European Portuguese can still be roughly twice the width of English.

I'm looking into whether there is a lightweight way to consume CLDR's stand-alone short data. For now, I'm leaning toward `narrow` as our best option: it is consistently compact and fully browser-supported, even though some labels are ambiguous without their column position. Could we use that consistently rather than special-casing English?

nynexman4464:
Thank you for taking this on and for adding the provider-locale regression coverage. While reviewing the weekday width, we found that the existing `Su` / `Mo` / `Tu` labels correspond to CLDR's `stand-alone.short` data, a width that `Intl.DateTimeFormat` does not expose. Browser `weekday: 'short'` can also produce full weekday names in some locales.

We followed that research through in #5190: it generates the compact CLDR data for the locales shipped by core, preserves the existing English labels, avoids the date-anchor timezone issue, and keeps the data synchronized when Crowdin adds a locale. That PR is now ready for review, so we plan to proceed with it rather than ask you to keep revising this implementation. I wanted to share the outcome here and acknowledge that your PR surfaced and proved the underlying bug.

HelloOjasMutreja:
Closing as superseded: #5190 landed on main with a more thorough CLDR-based approach to the same problem (real `stand-alone.short` CLDR data with a generated runtime table, rather than the `Intl.DateTimeFormat` approach here), touching the same `useCalendarDays.ts`/`Calendar.test.tsx` files. Thanks to nynexman4464 for the more complete fix.

PR #5118, opened 2026-08-16
fix(SideNav): collapse notifications fire once and reflect live state

AKnassa:
Split out of #4824 per review: four behavior fixes to SideNav and useResizable collapse notifications, separated from the persistence format change.

Refs #4790

## What this does

1. `resize()` out of the collapsed state now fires `onCollapseChange(false)`, matching the drag path and the callback's documented contract.
2. SideNav's `onCollapsedChange` fires exactly once per toggle click. The resize hook is the single notifier when resizable; the direct call remains for non-resizable navs and for a hook that already matches the requested state, in both directions. That covers a controlled parent that refused a drag-collapse, and the mirror case that surfaced while splitting: a parent that refuses an expand used to leave every later expand click silent.
3. Collapse notifications read live state instead of the state captured at the last render (`isCollapsedRef`). ResizeHandle registers its pointermove listener once at pointer down, so a whole gesture runs against one props snapshot. A drag now reports one collapse rather than one per pointer move, dragging back above the threshold re-expands mid-gesture, two imperative calls in one tick report the real final state, and a repeated `collapse()` can no longer overwrite the saved pre-collapse width with the zeroed collapsed size.
4. Toggling a `resizable` SideNav whose `collapsible` is unset is a no-op instead of leaving the imperative collapse handle reporting a collapse that never happened.

Also fixes the drag-path line flagged in the #4824 review: `preCollapseSizeRef.current = dragStartSizeRef.current || size` fell through to the stale closure `size` when the gesture began on the collapsed rail (`dragStartSizeRef` is 0 then), clobbering the saved width with 0. A gesture that begins collapsed now keeps the previous restore width.

Stacked on #4824; the first commit here is that PR's commit, review the top two. Overlaps the territory of #5073, happy to fold or reslice if the RFC lands first.

## Testing

Ten notification and drag tests moved here from #4824, plus new pins for the collapsed-rail round trip and the refused-expand mirror; each new test fails against its pre-fix code (mutation checked). 232 tests passing across SideNav and Resizable on top of current main, full core suite green.

### DIFF
--- a/.changeset/sidenav-collapse-notify-live-state.md
+++ b/.changeset/sidenav-collapse-notify-live-state.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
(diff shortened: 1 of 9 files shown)


cixzhang (changes requested):
Controlled apps can still lose the whole nav: after a parent rejects drag-collapse, `SideNav.tsx:385` leaves `useResizable` collapsed internally, so the callback reports `true` while the controlled prop stays expanded and the rendered width becomes `0px`.

The current contract requires rejected collapse intent to leave rendered size owner-controlled. #5075 already incorporated your live-state fix with attribution on the single-owner model, so I don't think this head should be rebased or merged separately.

[Reviewed by Robohands]

PR #5120, opened 2026-08-16
fix(DateInput, DateTimeInput, DateRangeInput, Calendar): follow the provider locale for date formatting and parsing

HelloOjasMutreja:
## Summary

Part of #5074, which covers several locale-ignoring paths across `Calendar`, `DateInput`/`DateTimeInput`, `Timestamp`, and `PowerSearch`. This PR scopes to the shared `plainDate`/`dateParser` utilities and their direct consumers (`DateInput`, `DateTimeInput`, `DateRangeInput`, `Calendar`, and the exported `useCalendarNavigation` hook). `Timestamp`'s relative-time and absolute-date formatting, and `PowerSearch`, need their own investigation and will follow as separate PRs. (Follows #5117, which already covered `Calendar`'s weekday header names.)

## Evidence

`plainDateFormat` (`packages/core/src/utils/plainDate.ts`), which backs `formatSharedDate` and is used directly by `DateRangeInput` and `Calendar`:

```ts
  pd: PlainDate,
): string {
  return new Intl.DateTimeFormat(undefined, options).format(
    plainDateToDate(pd),
  );
}
```

`undefined` resolves to the host/browser locale. `formatSharedDate` is the shared entry point `DateInput`'s display path uses (per its own doc comment, it's also the single source of truth `Timestamp`'s `formatTimestamp` switch maps onto).

Separately, `dateParser.ts`'s `isLocaleDayFirst` (used to disambiguate genuinely ambiguous numeric input like `"3/4/2026"` as day-first or month-first) had the same problem:

```ts
export function isLocaleDayFirst(): boolean {
  const parts = new Intl.DateTimeFormat().formatToParts(new Date(2000, 0, 15));
  ...
}
```

A tree wrapped in `InternationalizationProvider locale="es-ES"` still formatted committed dates and disambiguated ambiguous typed input using the host locale, not the provider's.

## Fix

Added an optional trailing `locale?: Locale` parameter to `plainDateFormat`, `formatSharedDate`, `isLocaleDayFirst`, and `parseDateInput` (threaded down to the internal `parseNumericDate` helper). Optional and additive, so no signature changes for any other caller of these exported utilities.

Threaded the locale through every consumer by reading `InternationalizationContext.locale` at each component/hook (`DateInput`, `DateTimeInput`, `DateRangeInput`, `Calendar`'s three functions, and `useCalendarNavigation`), matching the exact pattern `useDirection` already uses for `direction`.

`getTimeZoneParts` (also in `plainDate.ts`) deliberately hardcodes `'en-US'` for stable digit parsing, per its own doc comment, and is correctly left untouched, it's not locale-display output.

## Test notes

Added regression tests:
- `dateParser.test.ts`: the ambiguous `"3/4/2026"` case resolves to month-first for `en-US` and day-first for `es-ES`.
- `plainDate.test.ts`: `formatSharedDate` with `locale="es-ES"` produces `"25 de enero de 2026"` for `date_long`.
- `DateInput.test.tsx`: a component-level test rendering inside `InternationalizationProvider locale="es-ES"` confirms the committed display value follows the provider.

Verified each new test fails without the corresponding fix (falls back to host-locale output) and passes with it restored. Full suite across the five touched components/utilities (453 tests) passes.

## Test plan

- [x] `vitest run` on `DateInput`, `DateTimeInput`, `DateRangeInput`, `Calendar`, `dateParser.test.ts`, `plainDate.test.ts`: 453/453 passing
- [x] Verified new tests fail without the fix, pass with it
- [x] `eslint` clean on touched files (two pre-existing, unrelated `borderVars` unused-var warnings confirmed present before this change)
- [x] Changeset added

### DIFF
--- a/packages/core/src/DateInput/DateInput.test.tsx
+++ b/packages/core/src/DateInput/DateInput.test.tsx
@@ -917,6 +917,22 @@ describe('DateInput', () => {
       expect(screen.getByDisplayValue('Jan 25, 2026')).toBeInTheDocument();
     });
 
+    it('follows the InternationalizationProvider locale (#5074)', () => {
+      render(
+        <InternationalizationProvider locale="es-ES">
+          <DateInput
+            label="Date"
+            value="2026-01-25"
+            onChange={() => {}}
+            format="date_long"
+          />
+        </InternationalizationProvider>,
+      );
+      expect(
+        screen.getByDisplayValue('25 de enero de 2026'),
+      ).toBeInTheDocument();
+    });
+
     it('renders the ISO shape for format="system_date"', () => {
       render(
         <DateInput
--- a/packages/core/src/utils/plainDate.ts
+++ b/packages/core/src/utils/plainDate.ts
@@ -8,6 +8,7 @@
  */
 
 import type {ISODateString, PlainDate} from './dateTypes';
+import type {Locale} from '../i18n/types';
 
 export type {PlainDate} from './dateTypes';
 
@@ -293,10 +294,9 @@ export const DATE_FORMAT_SHORT_WITH_YEAR: Intl.DateTimeFormatOptions = {
 export function plainDateFormat(
   pd: PlainDate,
   options: Intl.DateTimeFormatOptions,
+  locale?: Locale,
 ): string {
-  return new Intl.DateTimeFormat(undefined, options).format(
-    plainDateToDate(pd),
-  );
+  return new Intl.DateTimeFormat(locale, options).format(plainDateToDate(pd));
 }
 
 // =============================================================================
@@ -344,9 +344,10 @@ export const SHARED_DATE_FORMAT_OPTIONS: Record<
 export function formatSharedDate(
   pd: PlainDate,
   format: SharedDateFormat,
+  locale?: Locale,
 ): string {
   if (format === 'system_date') {
     return plainDateToISO(pd);
   }
-  return plainDateFormat(pd, SHARED_DATE_FORMAT_OPTIONS[format]);
+  return plainDateFormat(pd, SHARED_DATE_FORMAT_OPTIONS[format], locale);
 }
--- a/.changeset/plaindate-locale-formatting.md
+++ b/.changeset/plaindate-locale-formatting.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] `DateInput`, `DateTimeInput`, `DateRangeInput`, and `Calendar` now format and parse dates using the ambient `InternationalizationProvider` locale instead of the host/browser locale. `plainDateFormat` (backing `formatSharedDate`) previously called `Intl.DateTimeFormat(undefined, ...)`, and `dateParser`'s day/month disambiguation heuristic for ambiguous numeric input (e.g. `3/4/2026`) called `Intl.DateTimeFormat()` with no locale at all, so a tree wrapped in a non-English `locale` still formatted and parsed dates using the host locale.
+
+@HelloOjasMutreja
--- a/packages/core/src/Calendar/Calendar.tsx
+++ b/packages/core/src/Calendar/Calendar.tsx
@@ -18,6 +18,7 @@
  */
 
 import {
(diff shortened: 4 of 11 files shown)


nynexman4464:
Apologies — while exploring how to split the provider-locale migration and lint enforcement into a stacked set of PRs, my coding agent incorrectly force-rebased this branch and changed the PR description before we had provided proper review feedback. I’ve restored the original branch head, `main` base, and description. We’ll provide review feedback first and keep any maintainer follow-up on a separate branch.

packages/core/src/utils/plainDate.ts
```diff
@@ -281,10 +282,9 @@ export const DATE_FORMAT_SHORT_WITH_YEAR: Intl.DateTimeFormatOptions = {
 export function plainDateFormat(
   pd: PlainDate,
   options: Intl.DateTimeFormatOptions,
+  locale?: Locale,
 ): string {
-  return new Intl.DateTimeFormat(undefined, options).format(
-    plainDateToDate(pd),
-  );
+  return new Intl.DateTimeFormat(locale, options).format(plainDateToDate(pd));
```
nynexman4464, line 299: Thanks for threading the locale through here. One calendar-model issue: `PlainDate` and Calendar’s arithmetic/grid are Gregorian, but `Intl.DateTimeFormat(locale, options)` honors the locale’s default calendar or a `-u-ca-*` extension. For example, `fa-IR` can render this Gregorian date as a Persian-calendar date while Calendar still displays and navigates Gregorian fields.

Could we pin the formatter to Gregorian and add a non-Gregorian-default locale regression test?

```ts
new Intl.DateTimeFormat(locale, {
  ...options,
  calendar: 'gregory',
})
```

The `isLocaleDayFirst` probe in `dateParser.ts` should make the same assumption.

packages/core/src/DateInput/DateInput.test.tsx
```diff
@@ -896,6 +896,22 @@ describe('DateInput', () => {
       expect(screen.getByDisplayValue('Jan 25, 2026')).toBeInTheDocument();
     });
 
+    it('follows the InternationalizationProvider locale (#5074)', () => {
```
nynexman4464, line 920: Could this test rerender the same component from one provider locale to another—for example `en-US` → `es-ES`—and assert that the displayed value changes? That proves the provider remains the live source after mount and catches a missing memo/callback dependency, rather than only proving that an initial Spanish locale is accepted.

nynexman4464 (changes requested):
Thanks for taking on the date-specific part of #5074. We’ve since been mapping the broader ambient-locale problem across the package, and this PR fits cleanly as the Calendar/date-input/date-helper slice:

- provider-locale utilities are being handled separately;
- Table, Timestamp, PowerSearch, Charts, and speech defaults are separate non-date work;
- this PR owns date formatting and ambiguous numeric day/month parsing;
- raw-Intl lint enforcement comes afterward, once the migrations are in place;
- compact weekday labels remain separate in #5190.

Keeping the helper locale parameters optional makes sense for source compatibility; we don’t need a codemod or breaking migration here.

Before landing, I think this needs:

1. Gregorian calendar pinning in `plainDateFormat` and `isLocaleDayFirst`, because Astryx `PlainDate` is Gregorian.
2. Provider-change rerender coverage rather than only initial-provider coverage.
3. A little more component-level coverage for the independently wired paths—at minimum Calendar/month navigation and the DateRangeInput/DateTimeInput committed display paths—so the helper tests don’t mask a caller that forgot to thread `locale`.
4. Narrower wording around parsing: this makes ambiguous ASCII numeric input follow D/M versus M/D ordering. It does not yet localize month-name parsing, non-Latin digits, or arbitrary locale date strings.

Once those are addressed, this should be the date-migration piece we build the later enforcement on. Adoption of the new `useLocale` helper can be a maintainer follow-up rather than additional scope for this contribution.

nynexman4464 (approved):
Approved to land the provider-locale migration; the remaining Gregorian, deterministic-default, and coverage refinements will follow in a maintainer PR.

Issue #5129, opened 2026-08-16
[Infra] Storybook aliases are maintained twice by hand — generate both blocks from one table


imdreamrunner:
## Problem

`apps/storybook/.storybook/main.ts` maintains the workspace source aliases **twice**, by hand:

- the StyleX `aliases` config (~line 82)
- Vite's `resolve.alias` (~line 124)

Adding a workspace package means remembering both. #5092 was exactly this drift: `@astryxdesign/richtext` was missing from both blocks while `apps/storybook/tsconfig.json` already mapped it to source, so typecheck and dev disagreed and `storybook dev` failed its dependency scan in any worktree without `packages/richtext/dist`.

#5100 fixed the instance and added a guard test. This issue is about removing the class.

## Current state, after #5100

| Package | StyleX `aliases` | Vite `resolve.alias` |
| --- | --- | --- |
| core, lab, charts, richtext | `/*` + bare → `src` | bare → `src` |
| theme-neutral, theme-stone, theme-y2k | `/*` only → `src/*` | bare → `src/source.ts` |
| vega | — | bare → `src` |

The asymmetries are **legitimate, not debt**: themes need a different entry per consumer, and `packages/vega/src` contains zero `@stylexjs/stylex` imports, so it correctly has no StyleX entry. Any dedup has to preserve all three shapes.

## Proposal

Derive both blocks from one declarative table, e.g.:

```ts
const workspaceAliases = [
  {pkg: '@astryxdesign/core',        src: 'packages/core/src',    stylex: 'both'},
  {pkg: '@astryxdesign/richtext',    src: 'packages/richtext/src', stylex: 'both'},
  {pkg: '@astryxdesign/theme-neutral', src: 'packages/themes/neutral/src',
   vite: 'packages/themes/neutral/src/source.ts', stylex: 'wildcard'},
  {pkg: '@astryxdesign/vega',        src: 'packages/vega/src',    stylex: 'none'},
  // …
];
```

with the two blocks generated from it. A new package becomes one row that cannot be half-added.

## Knock-on: the guard test

`apps/storybook/.storybook/main.test.ts` (added in #5100) asserts alias keys by **string matching** `main.ts`, scoped to the Vite block via `mainTs.slice(mainTs.indexOf('alias: {'))`. Two consequences:

- Generating the blocks would break that slice — the test must be reworked in the same PR.
[...]

PR #5133, opened 2026-08-17
fix(storybook): load the Vite plugin from source so a fresh clone can run dev

### LINKED ISSUE #5128: [Infra] Storybook dev fails on a cold clone: @astryxdesign/build/vite is unbuilt
## Problem

A genuinely fresh worktree still cannot run `pnpm -F @astryxdesign/storybook dev`, even after #5100 fixed the `@astryxdesign/richtext` resolution. Storybook now fails *earlier*, on its own config load:

```
Cannot find package '@astryxdesign/build/vite' imported from apps/storybook/.storybook/main.ts
```

`apps/storybook/.storybook/main.ts` imports `astryxStylex` from `@astryxdesign/build/vite`, whose export map resolves to `packages/build/dist/vite.mjs`. That file does not exist until `pnpm -F @astryxdesign/build build` has run.

## Why a Vite alias cannot fix this one

This is the **Storybook config loader's** node ESM resolution, not Vite's module graph — the failure happens while loading `main.ts` itself, before any `resolve.alias` from that file is in play. So the fix that worked for richtext (#5100) is structurally unavailable here.

Noted as out of scope by @jiunshinn in #5100 and confirmed independently: the #5092 reporter's repro happened to have `packages/build` already compiled, which is why it surfaced richtext rather than this.

## Reproduction

```sh
git clone <repo> && cd astryx
pnpm install
test ! -e packages/build/dist/vite.mjs   # true in a fresh worktree
```

Workaround: `pnpm -F @astryxdesign/build build` first.

## Why it matters beyond dev startup

It also blocks a stronger regression guard. #5100's alias check is a *static text* assertion over `main.ts` rather than a real resolution test, and that shape was the right call precisely because of this bug: a test that `import()`s the config and inspects the resolved `viteFinal` aliases fails in a fresh worktree with the error above. Fixing this unlocks the better test.

## Possible directions

- Have `apps/storybook`'s `dev`/`build` scripts depend on a `@astryxdesign/build` build (pnpm `dependsOn` / a prepare step).
- Point the Storybook config at `packages/build/src/vite.ts` directly (`apps/storybook/tsconfig.json` already maps it that way, so typecheck and runtime currently disagree — the same class of mismatch as #5092).
- Publish a source-first conditional export for the workspace-internal consumer.

Not scoped — whichever keeps `pnpm install && pnpm dev` working from a cold clone.

AKnassa:
Fixes #5128

## What this does

`storybook dev` starts on a freshly cloned repo again. Before this, a new contributor ran `pnpm install`, ran `pnpm dev`, and got an error instead of Storybook.

## Why

Storybook reads `.storybook/main.ts` with Node's own module resolver, and it does that *before* any of the aliases written inside that file exist. The file imported the Astryx Vite plugin as `@astryxdesign/build/vite`, and that package's export map points only at `dist/vite.mjs`, which is not on disk until `@astryxdesign/build` has been built:

```
SB_CORE-SERVER_0007 (MainFileEvaluationError): Storybook couldn't evaluate your .storybook/main.ts file.
Error [ERR_MODULE_NOT_FOUND]: Cannot find module '.../packages/build/dist/vite.mjs'
imported from .../.storybook/main.ts
```

The alias fix that solved the same shape of problem for `@astryxdesign/richtext` in #5100 cannot reach this one. Aliases live inside the config; this failure happens while the config is being read.

## What changed

* `.storybook/main.ts` imports the plugin from `packages/build/src/vite.ts` rather than the bare package name. Node strips the types itself for a relative `.ts` path, so nothing needs compiling first.
* `apps/storybook/tsconfig.json` gains `allowImportingTsExtensions` for that explicit extension, and drops its `@astryxdesign/build/vite` path mapping. The mapping already pointed at source, so typecheck and runtime were answering differently; now there is one answer.
* Four tests in `.storybook/main.test.ts`. Three are structural and hold on any checkout: `main.ts` may not import a workspace package by bare name, and its plugin import must resolve to a source file that exists unbuilt. The fourth is the behavioural test #5128 notes this unblocks, which imports the config, runs `viteFinal`, and inspects the aliases that come back.
* `CONTRIBUTING.md`. The Storybook section said Storybook loads pre-built packages from `dist/` folders and that you have to build first. That is not true, and it is plausibly why this break stayed out of sight. Happy to split this hunk into its own change if you would rather keep the PR to code.

The published `@astryxdesign/build/vite` export is untouched. Only the in-repo Storybook app changed how it reaches the plugin.

Option 1 in the issue (a build step wired into the `dev` script) was considered and set aside: it hides the symptom but leaves the `import()`-the-config test still broken, because vitest resolves the same export map. Option 3 (a source-first conditional export) needs `--conditions`, which the Storybook config loader never passes.

## How to see it

On a worktree with every `dist/` deleted:

```sh
pnpm install
pnpm -F @astryxdesign/storybook dev     # http://localhost:6006, 1858 stories
pnpm -F @astryxdesign/storybook build   # exits 0
```

Storybook serves `packages/core/src/Badge/Badge.tsx` straight from source, the built CSS carries `astryx`-prefixed StyleX classes, and an edit to a core source file reaches the running dev server with no rebuild and no restart.

## Checks

`pnpm test`: 10504 passing. The two `packages/cli` failures are pre-existing and reproduce identically on an untouched `main` (case-insensitive filesystem on macOS). `pnpm lint` clean. `apps/storybook` typecheck adds no new errors; the 28 it already reports come from unbuilt `charts` / `themes` / `vega` dists, which is a separate gap from this one.

#5129 is deliberately untouched. Reworking the alias guard test belongs there.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -92,9 +92,6 @@ cd astryx
 # Install dependencies
 pnpm install
 
-# Build core package first (required for Storybook)
-pnpm -F @astryxdesign/core build
-
 # Start Storybook for component development
 cd apps/storybook
 pnpm dev
@@ -102,19 +99,9 @@ pnpm dev
 
 ### Running Storybook
 
-Storybook loads pre-built packages from `dist/` folders, so you need to build packages before running Storybook.
-
-**First time setup:**
-
-```bash
-# Build all packages
-pnpm build
-
-# Or build just core
-pnpm -F @astryxdesign/core build
-```
-
-**Start Storybook:**
+Storybook resolves every workspace package to its `src/` directory and compiles
+it itself, so a fresh clone needs no build step first — `pnpm install` then
+`pnpm dev` is enough.
 
 ```bash
 cd apps/storybook
@@ -127,16 +114,8 @@ Storybook will open at http://localhost:6006 with:
 - **Mode switcher** - Toggle between Light and Dark modes
 - **Component stories** - Interactive component examples
 
-**If you make changes to `@astryxdesign/core`:**
-
-```bash
-# Rebuild core package
-pnpm -F @astryxdesign/core build
-
-# Restart Storybook to see changes
-cd apps/storybook
-pnpm dev
-```
+**If you make changes to `@astryxdesign/core`:** nothing extra. The dev server
+serves the edited source, so the story updates on save — no rebuild, no restart.
 
 ### Running the Doc Site
 
--- a/apps/storybook/.storybook/main.test.ts
+++ b/apps/storybook/.storybook/main.test.ts
@@ -2,17 +2,23 @@
 
 /**
  * @file main.test.ts
- * @description Guards the workspace source-alias invariant in Storybook's
- *   Vite config. Every `@astryxdesign/*` dependency of this app must be
- *   aliased to package source in `.storybook/main.ts`. Without an alias,
- *   Vite resolves the workspace link through the package's export map,
- *   which points at `dist/` build output that does not exist in a fresh
- *   worktree — `storybook dev` then fails its dependency scan (#5092:
- *   `@astryxdesign/richtext`).
+ * @description Guards two things that both keep `storybook dev` working from
+ *   a cold clone, at two different resolution stages.
+ *
+ *   1. Vite's module graph: every `@astryxdesign/*` dependency of this app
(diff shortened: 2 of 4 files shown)


cixzhang (approved):
Thanks, this fixes the cold-clone path cleanly. I reproduced the parent failure and confirmed this head starts all 1,858 stories without generated build output.

[Reviewed by Robohands]

PR #2186, opened 2026-05-18
fix(DropdownMenu): light-dismiss + click-toggle race on touch browsers

lexs:
## Summary

On iOS Safari (and other touch browsers where `pointerdown` fires light-dismiss before the trigger's `click` event), tapping the `XDSDropdownMenu` trigger to close an open menu would close and then immediately re-open on the same tap. The hide flows through `useXDSPopover`'s `onHide` first, then the click hits `handleButtonClick`, which reads `popover.isOpen=false` and toggles it back to true.

Mirror the existing 50ms hide-time guard in `XDSPopover` (lines 268, 274–277, 290–296 — same race, same fix) into `XDSDropdownMenu`: stamp `lastHideTimeRef` in `handleLayerHide`, then short-circuit `handleButtonClick` if it fires within 50ms of that stamp.

Also fixes `XDSMoreMenu` since it wraps `XDSDropdownMenu`.

This ports the equivalent internal fix (Meta D104032592 in fbsource, which patches `XDSInternalPopoverMenu.react.js` — the shared internal layer behind both `XDSDropdownMenu` and `XDSPopoverMenu`) to the public xds repo. Here, the internal layer is `useXDSPopover`, but `XDSDropdownMenu` calls into it directly with its own toggle logic, so the guard goes alongside that logic. `XDSPopover` already has the equivalent guard for the same race.

## Test Plan

Added a regression test in `XDSDropdownMenu.test.tsx` that fires open → close → click and asserts the menu does not re-open (`showPopover`/`hidePopover` each called exactly once).

```
$ yarn test packages/core/src/DropdownMenu/XDSDropdownMenu.test.tsx
 Test Files  1 passed (1)
      Tests  29 passed (29)

$ yarn test packages/core/src/Popover/XDSPopover.test.tsx packages/core/src/MoreMenu/XDSMoreMenu.test.tsx
 Test Files  2 passed (2)
      Tests  31 passed (31)

$ yarn workspace @xds/core build
✅ Build success
```

ESLint clean on the changed files (one pre-existing unused-`container` warning in an unrelated test left as-is).

PR #5135, opened 2026-08-17
fix(ComplexSelector): reopen on the first trigger click after any dismiss

cixzhang:
## Why

`ComplexSelector` drops any trigger click that lands within 50ms of any hide. The line arrived with #4769, copied from the `DropdownMenu` touch idiom (#2186) but with no discrimination by cause: Escape, an outside click, a row pick and a programmatic `close()` all leave the trigger dead for 50ms, so the next click on it does nothing.

Honest scope, because it changes the framing: 50ms is below human reaction time. Measured in Chromium, at a ~105ms gap every dismissal route already reopens on the first click — so most people clicking the pill twice never see this, and to reproduce it by hand you have to hit the 24–46ms gaps in the table below. What it does break reliably is anything that clicks without human latency: automation, test suites, and code driving the imperative `handleRef` to open right after close — which is exactly the API #4769 shipped. That is visible in this diff: each of the four new unit tests fails on `main` purely because the click follows the dismiss with no wall-clock delay.

## What

The guard now keys on the pointer sequence that produced the dismiss instead of on a clock. A pointerdown on the trigger while the popup is open marks the interaction in flight; the click that follows is swallowed only if that press is what closed the popup. A key press clears the mark, so a keyboard activation never inherits an abandoned pointer press.

That is precisely the race #2186 defends against — on iOS Safari the tap's pointerdown light-dismisses before the same tap's click reaches the trigger — so the protection is preserved, not deleted. Every other dismissal is not that sequence, so the very next click reopens with no delay.

## Risk

Low, and contained to this component: no API change, no new state, one hide-time stamp replaced by a pointer flag plus a synchronous mirror of the open state.

Two things worth review. First, an unfixably ambiguous case is now decided the safe way: if a browser does *not* light-dismiss on the invoker (current Chrome, which knows the trigger as the popover's `source`), the popup is still open when the click lands and the click closes it as a normal toggle. Second, `#4769`'s regression test asserted "hide, then a bare click, must not reopen"; a bare click after a hide is exactly the automation case that is broken, and the test could not tell the defended sequence from any other close. It now fires the full tap (pointerdown → light-dismiss → click).

`DropdownMenu` and `Popover` still carry the same clock-based guard. Neither is in this regression's path, so converting them is deliberately left out of this PR.

## Testing

Real Chromium against Storybook, clicking the trigger immediately after each dismissal (measured gaps 24–46ms, i.e. inside the old window):

| dismissal route | before | after |
| --- | --- | --- |
| Escape | swallowed | reopens |
| trigger click | swallowed | reopens |
| outside click | swallowed | reopens |
| row pick (`close()`) | swallowed | reopens |
| programmatic close | swallowed | reopens |

Touch protection, with a real touch pointer and the iOS ordering reconstructed (pointerdown light-dismisses, then the click lands): tap 1 opens, tap 2 leaves it shut — the race is still caught — tap 3 reopens. Before this change tap 3 also stayed shut.

Unit tests: four new cases, one per dismissal route, each failing on `main` and passing here; the touch-race test passes both before and after. `packages/core` full suite 6678 passed, typecheck and eslint clean.

Not done: no consumer has been re-tested end to end against this branch. Consumers install published `@astryxdesign/core`, so that can only happen after a release — worth a check then, since the imperative-open path is what this fixes.

### DIFF
--- a/.changeset/complexselector-reopen-after-dismiss.md
+++ b/.changeset/complexselector-reopen-after-dismiss.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] ComplexSelector: a trigger click reopens the popup right after Escape, an outside click, a row pick or a programmatic close. The light-dismiss guard now keys on the pointer sequence that produced the dismiss instead of a 50ms window after any hide.
+
+@cixzhang
--- a/packages/core/src/ComplexSelector/ComplexSelector.test.tsx
+++ b/packages/core/src/ComplexSelector/ComplexSelector.test.tsx
@@ -354,13 +354,15 @@ describe('ComplexSelector', () => {
     expect(popover).not.toBeNull();
 
     // The touch race #2186 defends against, in its real order: the tap's
-    // pointerdown light-dismisses the popup, then that same tap's click
-    // reaches the trigger. It must not reopen what the tap just closed.
+    // pointerdown light-dismisses the popup, then the tap finishes on the
+    // trigger and its click reaches it. It must not reopen what the tap just
+    // closed.
     fireEvent.pointerDown(trigger);
     const closeEvent = new Event('toggle');
     Object.defineProperty(closeEvent, 'newState', {value: 'closed'});
     fireEvent(popover as HTMLElement, closeEvent);
     expect(trigger).toHaveAttribute('aria-expanded', 'false');
+    fireEvent.pointerUp(trigger);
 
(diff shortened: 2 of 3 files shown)


cixzhang:
Closing in favor of [#5018](https://github.com/facebook/astryx/pull/5018). This branch fixes the right bug locally in `ComplexSelector`; #5018 fixes the same gesture race in shared layer handling, removes the local timing guard here, and covers the other affected popup triggers too. We should land one owner for the behavior, not both.

PR #5136, opened 2026-08-17
feat(SegmentedControl): SegmentedControlItem takes a count, named for screen readers

cixzhang:
**Why.** A tab strip whose choices hold different numbers of rows has to say so — "Needs me" empty and "Needs me 12" are different messages, and the counts are what a viewer scans to decide where to go. `SegmentedControlItem` had no affordance for one, so a product that shows row counts in a tab strip forked a private copy of the whole control rather than lose the number, and that app now runs two segmented controls side by side: ours on a sidebar sort axis, its own fork on the tab strip.

**What.** `SegmentedControlItem` takes `count` — a number rendered after the label, inside the segment, so it stays within the hit target and the `layout="fill"` widths — and `countLabel`, what that number counts. The number is `aria-hidden` and the segment's accessible name is composed instead: "Needs me, 12 sessions". It renders at every size and every width, including an icon-only segment (`isLabelHidden`), stays out of the selected segment's bold weight, and has its own theming target (`astryx-segmented-control-item-count`). Story, tests and changeset included.

**The API decision.** A count is a number, so it is a prop, not a slot: `count?: number` plus `countLabel?: string`.

- *A `badge` / `endContent` slot* is the sibling pattern — `Tab`, `SideNavItem` and `ListItem` all take `endContent`, and I would have copied it if the accessible name weren't the whole problem. Slot content joins the button's name as a trailing digit ("Inbox 12"), and a component that receives a `ReactNode` cannot know what the number means, so it cannot fix that — every consumer has to. A slot also invites arbitrary content into a radio's name and hit target. Where content genuinely is arbitrary, `endContent` can still be added later beside `count`, matching `Tab`; typing this one as a number does not foreclose it.
- *Composition via children* isn't on the table: the item takes `label` as a prop rather than children, and content inside a `role="radio"` lands in its accessible name anyway.
- The cost of a number-typed prop, taken knowingly: no "99+", no formatted or non-numeric badge. If that need is demonstrated, the answer is the `endContent` slot, not widening `count` to `ReactNode`.
- `count={0}` renders "0". Suppressing a value the consumer passed is a design recipe living in the API; `count={n || undefined}` is the quiet-at-zero recipe, and it stays the caller's call.

**Risk.** Both props are optional and nothing renders differently without them. Two behavioral notes: a segment with a count now carries an explicit `aria-label` (the name still begins with the visible label, so Label-in-Name holds), and the count consumes width, so a long label in a squeezed `layout="fill"` strip wraps about 60px earlier than before — three counts, ~20px each. The wrapping itself is existing behavior (the segment sets no `whiteSpace: nowrap`) and I left it alone, because [#5137](https://github.com/facebook/astryx/pull/5137) — responsive icon-only labels on this same component — is being drafted in that territory and an icon-only narrow mode is the real answer for it.

**Testing.** 336 tests pass across the SegmentedControl and theming-target suites, 7 of them new: name composition with and without `countLabel`, the hidden-label case, zero, the uncounted case, and selection still firing from a counted segment. Verified in real Chromium at 900/520/420/360/320/280/240px in light and dark — segment heights stay equal (28px at every width, counts on or off), nothing clips, the count is optically centered against its label, and Chromium's own accessibility tree reads "Needs me, 12 sessions [checked]" rather than "Needs me 12". Count contrast against the track: 7.8:1 selected and 6.8:1 unselected in light, 6.0:1 and 4.4:1 in dark. That last figure is the existing `--color-text-secondary`-on-track pairing the unselected *label* already uses — the count matches its label exactly — and it sits just under AA; flagging it rather than changing a token here.

### DIFF
--- a/.changeset/segmented-control-item-count.md
+++ b/.changeset/segmented-control-item-count.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] `SegmentedControlItem` takes a `count` — the tally a tab strip needs to say how many rows each choice holds ("Needs me 12"), rendered inside the segment so it stays within the hit target and the `layout="fill"` widths. The number is `aria-hidden` and `countLabel` names what it counts, so the segment announces "Needs me, 12 sessions" instead of the bare "Needs me 12" a screen reader would otherwise read. It renders at every width, including an icon-only segment (`isLabelHidden`), and stays out of the selected segment's bold weight.
+
+@cixzhang
--- a/apps/storybook/stories/SegmentedControl.stories.tsx
+++ b/apps/storybook/stories/SegmentedControl.stories.tsx
@@ -236,6 +236,7 @@ export const DisabledWithMessage: Story = {
 export const WithCounts: Story = {
   args: {
     size: 'md',
+    isDisabled: false,
   },
   render: args => {
     const [value, setValue] = useState('needs-me');
@@ -244,7 +245,8 @@ export const WithCounts: Story = {
         value={value}
         onChange={setValue}
         label="Sessions"
-        size={args.size}>
+        size={args.size}
+        isDisabled={args.isDisabled}>
         <SegmentedControlItem
           value="needs-me"
           label="Needs me"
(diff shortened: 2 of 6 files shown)


cixzhang:
Closing by API decision. A numeric count is only one kind of trailing content; `SegmentedControlItem` should expose a generic end-content slot rather than freeze a specialized `count` + `countLabel` pair. The caller can compose and name a count through that slot.

PR #5137, opened 2026-08-17
feat(SegmentedControl): collapsibleLabels — a strip that goes icon-only when the labels do not fit

cixzhang:
## Why

A four-choice theme picker (Light / Dark / System / Custom, each an icon plus a
label) inside a settings dialog needs its labels when there is room and its
icons when there is not. `isLabelHidden` cannot express that: it is per item and
fixed at render, so a strip is icon-only always or labelled always. The only way
through the public API is a JS media query feeding a boolean into every item —
re-rendering the group on resize to express something layout already knows, and
keyed to the viewport when the constraint is the dialog. The app in the report
kept its own segmented control rather than do that.

## What

`collapsibleLabels` on the group. Off by default, and absent means today's
behavior byte for byte.

```tsx
<SegmentedControl value={theme} onChange={setTheme} label="Theme" collapsibleLabels>
  <SegmentedControlItem value="light" label="Light" icon={<Icon icon={SunIcon} />} />
  …
</SegmentedControl>
```

- `true` — the strip measures itself and collapses every label to its icon when
  the labelled strip needs more room than it has.
- `{isCollapsed}` — the caller decides; the strip never measures.
- `{onCollapsedChange}` — observe the decision.

Through the collapse the label stays the segment's accessible name and becomes
its tooltip, so nothing is lost but the pixels. An item with no icon keeps its
label (an empty segment is not a state worth rendering) and says so once in dev.

## The API decision

**How does a strip decide it is narrow? It doesn't — it decides whether the
labels fit, which is a measurement, not a threshold.** The trigger is the width
the labelled strip needs versus the width its container gives it. No breakpoint
is baked into the component, and none has to be invented by the caller.

Rejected, and why:

- **A group-level static `isLabelHidden`.** Simple and honest, but it only moves
  the caller's problem up one level: they still need a media query, still
  re-render on resize, and still answer "how narrow is narrow" with a number
  that has nothing to do with the labels they actually wrote. It is also already
  reachable — `{isCollapsed: true}` — so the axis is not lost.
- **A container query with a breakpoint inside the component.** This is the
  support burden the report warns about: the strip would change shape at a width
  nobody declared. [#4930](https://github.com/facebook/astryx/pull/4930) is the
  house evidence — given a raw number, every reader invents a different one, and
  the fix there was to *name* the condition rather than publish a pixel. A
  container query also cannot take its threshold from a prop (no `var()` in the
  condition), so the number would have to be ours.
- **Overflow-driven collapse à la
  [#4868](https://github.com/facebook/astryx/pull/4868) / `OverflowList`.** Wrong
  shape: a radio group cannot move segments into a menu — all four choices must
  stay on screen and stay selectable. Its hidden inert measurement copy also
  duplicates every `role="radio"` inside the group, which is precisely the
  roving-focus hazard that PR had to fix for Toolbar.

What it borrows instead is `SideNav`'s collapsed rail, the pattern the system
already ships for "this became an icon": `boolean | config` prop on the
container, `aria-label` plus a tooltip on the icon-only control.

## Risk

Low and opt-in, but three things worth knowing:

- **It measures.** One shared `ResizeObserver` (the existing
  `sharedResizeObserver`), two observed elements per strip, a layout read on
  resize. The natural width is remembered from the last labelled measurement,
  which is what lets the strip grow back and what keeps it from reading its own
  collapsed width as the width it needs.
- **A `'hug'` strip measures its parent** (a hug box shrinks with its content, so
  its own width answers nothing). Give it a parent with a width of its own; a
  shrink-to-fit parent will collapse and stay collapsed. `'fill'` measures
  itself and has no such caveat.
- **The tooltip sets `aria-describedby`,** so a collapsed segment reports name
  *and* description as the same string. That is `SideNav`'s shipped behavior
  today; deduping it belongs in `Tooltip`, not here.

## Testing

11 new tests (52 in the file, 6684 in core, all green), and Chromium at
320/375/480/768/1280 against the new `CollapsibleLabels` story.

| viewport | available | strip needs | state | item widths |
|---|---|---|---|---|
| 320 | 288px | 352px | icons | 40 / 40 / 40 / 40 |
| 375 | 343px | 352px | icons | 40 / 40 / 40 / 40 |
| 480 | 448px | 352px | labels | 78 / 75 / 93 / 95 |
| 768 | 736px | 352px | labels | 78 / 75 / 93 / 95 |
| 1280 | 1248px | 352px | labels | 78 / 75 / 93 / 95 |

Bisected, the flip is one pixel wide and exactly where the arithmetic says:
labels at 351px available, icons at 350px, against the 352px the labelled strip
needs. The `'fill'` twin in the same story crosses within 3px of it, with no
clipped or wrapped label at any width on the way down.

Accessibility, read from Chrome's accessibility tree (not the markup) at every
width above: `radiogroup "Theme"` with `radio "Light" (checked)`, `radio "Dark"`,
`radio "System"`, `radio "Custom"` — identical collapsed and expanded, none
ignored. Hovering a collapsed segment opens its tooltip ("Light"), and
`ArrowRight` still moves and selects (`Dark` checked and focused) while
icon-only.

Not done: the count badge of the sibling report is untouched. This collapse
hides the label text only, so a badge would survive it — but whether an
icon-plus-badge segment folds the count into its accessible name is that
change's call, not this one.

### DIFF
--- a/.changeset/segmented-control-collapsible-labels.md
+++ b/.changeset/segmented-control-collapsible-labels.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] SegmentedControl: `collapsibleLabels` collapses the strip to icons when the labels do not fit, keeping each label as the segment's accessible name and tooltip.
+
+@cixzhang
--- a/apps/storybook/stories/SegmentedControl.stories.tsx
+++ b/apps/storybook/stories/SegmentedControl.stories.tsx
@@ -11,6 +11,10 @@ import {
   Squares2X2Icon,
   ListBulletIcon,
   TableCellsIcon,
+  SunIcon,
+  MoonIcon,
+  ComputerDesktopIcon,
+  SwatchIcon,
 } from '@heroicons/react/24/outline';
 
 const meta: Meta<typeof SegmentedControl> = {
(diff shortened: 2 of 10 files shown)


cixzhang:
Closing for API redesign. `SegmentedControlItem` already supports per-item `isLabelHidden`, but icons are not required and there is no group-level icon eligibility contract. Automatic collapse therefore produces uneven mixed strips when some items lack icons. The next design should define eligibility and mixed-item behavior before adding measurement.

PR #5139, opened 2026-08-17
fix(HoverCard): give touch a tap contract instead of a coin flip

cixzhang:
## Why

A hover card on a link has no contract on touch. There is no hover, so the browser synthesizes one from the tap — and `HoverCard`'s focus trigger fires on the same tap, with no `:focus-visible` check. Measured on an iPhone 13 descriptor in real Chromium at 0.4.3: every tap both opened the card *and* followed the link (the popover mounts ~4ms before the click lands). A touch user has no way to read the card without being taken off the page, and no way to follow the link without the card appearing over it.

## What

Taps now go through an explicit contract instead of arriving as a side effect of synthesized hover and focus. A new `touchTrigger` prop (mirroring `focusTrigger`'s vocabulary) says what a tap means:

- **`auto` (default)** — a tap opens the card only when it is not activating something. On a link or a button the card stays closed and the tap does exactly what it looks like it does; on a text trigger, which activates nothing, the tap opens the card and a second tap closes it.
- **`always`** — the two-tap contract on any trigger: the first tap previews and is consumed, the second activates.
- **`never`** — taps never open the card.

On touch, the synthesized `mouseenter` and the tap's `focusin` no longer open the card in any mode; a keypress clears the gesture, so keyboard focus after a tap still opens it. The pointer type is read per interaction from `pointerenter`, so a hybrid device's mouse is unaffected. A tap-opened card also gets an explicit outside-touch dismissal, because touch has no pointer-leave and the `mouseleave` the next tap synthesizes is not a contract.

**The contract and its cost.** The conventional answer — first tap previews, second tap activates — costs every touch user a tap on every hovercard link, whether or not they wanted the card. This system already refuses that trade twice: `Tooltip` suppresses itself entirely on `(hover: none)` ("hover is simulated and eats a tap"), and `useContainerReveal` never gates content behind hover on a coarse pointer. So the default here does not take the tap either, and the two-tap contract is one prop away for the surfaces whose card is the point (a person-mention trigger whose card carries the profile is the case that filed this). The cost of that default is honest and worth stating: by default a card on a link is desktop- and keyboard-only. The card was already unreachable on touch — the tap navigated away from it — so nothing regresses; what changes is that it is now a stated decision the consumer can reverse, rather than a race.

## Risk

Behavior change on touch only, and it is the point of the change. Three things worth review:

- The first tap of the two-tap contract is swallowed in capture phase, so a consumer's own `onClick` on the trigger does not run for it. That is the contract, but it is a real behavior change for any trigger that opts in.
- Cards on links/buttons no longer appear at all on touch under the default. The prop is the opt-out.
- The tap's `focusin` no longer opens the card. Programmatic focus that arrives during a touch gesture is now ignored too; keyboard focus is unaffected, and `focusTrigger` still governs whether focus listeners are attached at all.

Desktop mouse and keyboard paths are untouched — the click handler returns immediately unless the current gesture is touch.

## Testing

Real Chromium, iPhone 13 descriptor (`pointer: coarse`, `hover: none`), 20 assertions, all passing:

| | before | after |
| --- | --- | --- |
| default, link: tap ×3 | card opens + navigates, every tap | navigates, card stays closed |
| `always`, link: tap 1 | opens + navigates | opens, does not navigate |
| `always`, link: tap 2 | — | navigates, card closes |
| `always`, button: tap 1 / tap 2 | — | `onClick` suppressed / `onClick` fires |
| default, text trigger: tap 1 / tap 2 | opens | opens / closes |
| tap-opened card, tap elsewhere | — | dismissed |
| desktop mouse hover / leave / click | opens / closes / navigates | unchanged |
| keyboard Tab / Escape / Enter | opens / closes / navigates | unchanged |

Unit tests: 7 new cases; the 4 that encode the contract fail on `main`, the 3 invariants (text trigger, keyboard-after-tap, mouse click) pass on both. `packages/core` full suite 6681 passed; `check:repo` and eslint clean.

**The report's second half did not reproduce, and this PR does not address it.** The report was that light dismiss on touch is unreliable. It did not reproduce in the necessity audit (9/9 outside taps dismissed) and it did not reproduce here in 18 further trials across four arrangements — two cards on one line, a 292px slow-loading card, a card inside a scrolling container, a nested card, dismissal attempted 50–900ms after opening, plus a scroll-then-dismiss pass. One mechanism finding for whoever re-specifies it: `HoverCard` does not use `lightDismiss`, so its layer is `popover="manual"` and there is no platform light dismiss to be unreliable — today an outside tap closes the card only via the `mouseleave` the browser synthesizes from that tap. That is a plausible source of the original sighting on a real phone, and it is exactly the dependency the new outside-touch dismissal removes for tap-opened cards.

Not done: `Tooltip` has the same shape of problem and keeps its own `(hover: none)` suppression; it is not converted to `touchTrigger` here. No consumer-side adoption here — a person-mention hover card of that kind will need `touchTrigger="always"` to get the preview back on touch.


cixzhang:
Superseded by #5248, which shipped a `touchTrigger` prop covering the same ground (plus suppressed synthesized focus, activation-element detection and outside-tap dismissal, shared with Tooltip). Closing in favour of that.

PR #5143, opened 2026-08-17
fix(RadioListItem): make the whole row a click target

freddymeta:
## What

Makes the **whole `RadioListItem` row a click target**. Clicking the description — or anywhere in the row's hover area — now selects the radio, and the row shows the shared hover/press affordance. Previously only the radio circle and its label text responded; the description and the surrounding row were dead space.

## The bug

`CheckboxListItem` already gets this right: it composes its row in **delegation mode**, where the row is an enlarged click/tap surface that proxies surface clicks to the control via `useClickableContainer` (the `Item`/`ListItem` `interactiveRef` pattern). `RadioListItem` never adopted that — it wrapped only the label in `<label htmlFor>` and left the description as a bare `<span>`, so:

- clicking the **description** did nothing;
- clicking the **empty hover area** of the row did nothing;

…while the visually-identical checkbox row responded everywhere. This is the mismatch reported.

## The fix

Mirror `CheckboxListItem`:

- Pass the radio `<input>` as `Item`'s **`interactiveRef`** → the row delegates surface clicks (label text, description, empty area) to the radio.
- Pass **`isDisabled`** to `Item` so a disabled row is a non-interactive surface (`pointer-events: none`), consistent with the checkbox.
- Drop the manual **`<label htmlFor>`** — under delegation a real `<label>` would double-fire (native label activation *and* the proxied click). The visible label is now plain text; the radio carries its accessible name via **`aria-label`**, so `getByLabelText` / `getByRole('radio', {name})` are unchanged.

The radio stays the option's **sole tab stop** (WCAG 4.1.2) — no invisible row button is added. The focus ring is still owned by the radio indicator (`useIndicatorFocusRing`), so there's no double ring.

Also routed the `startContent`/`description` render guards through `isRenderable` to clear two pre-existing `@astryx/no-nullish-jsx-guard` warnings in the file.

## Tests

- New: **clicking the description selects the radio.**
- Updated the two disabled-click tests to `fireEvent` (mirroring `CheckboxList`) — `userEvent` correctly refuses to click a `pointer-events: none` surface, which is the intended real-world behavior.
- Full suite: **45 tests pass.** `build` ✓ · `tsc --noEmit` ✓ · ESLint ✓ (0 warnings) · `check:repo` ✓.

### DIFF
--- a/packages/core/src/RadioList/RadioList.test.tsx
+++ b/packages/core/src/RadioList/RadioList.test.tsx
@@ -116,6 +116,22 @@ describe('RadioList', () => {
     expect(handleChange).toHaveBeenCalledWith('b');
   });
 
+  it('calls onChange when clicking the description', async () => {
+    const user = userEvent.setup();
+    const handleChange = vi.fn();
+    render(
+      <RadioList label="Preference" value="a" onChange={handleChange}>
+        <RadioListItem label="Option A" value="a" description="First option" />
+        <RadioListItem label="Option B" value="b" description="Second option" />
+      </RadioList>,
+    );
+
+    // The whole row delegates surface clicks to the radio, so the description
+    // is a live target — not dead space as it was before.
+    await user.click(screen.getByText('Second option'));
+    expect(handleChange).toHaveBeenCalledWith('b');
+  });
+
   it('disables all radios when group isDisabled is true', () => {
     render(
       <RadioList label="Preference" value="" onChange={() => {}} isDisabled>
@@ -137,7 +153,13 @@ describe('RadioList', () => {
       </RadioList>,
     );
 
-    await user.click(screen.getByLabelText('Option A'));
+    // userEvent honors `pointer-events: none`, so clicking the disabled row's
+    // delegate surface is genuinely blocked (it refuses and throws) rather than
+    // merely ignored by the handler — the guarantee this PR's delegation relies
+    // on.
+    await expect(user.click(screen.getByLabelText('Option A'))).rejects.toThrow(
+      /pointer-events/i,
+    );
     expect(handleChange).not.toHaveBeenCalled();
   });
 
@@ -162,7 +184,13 @@ describe('RadioList', () => {
       </RadioList>,
     );
 
-    await user.click(screen.getByLabelText('Option A'));
+    // userEvent honors `pointer-events: none`, so clicking the disabled row's
+    // delegate surface is genuinely blocked (it refuses and throws) rather than
+    // merely ignored by the handler — the guarantee this PR's delegation relies
+    // on.
+    await expect(user.click(screen.getByLabelText('Option A'))).rejects.toThrow(
+      /pointer-events/i,
+    );
     expect(handleChange).not.toHaveBeenCalled();
   });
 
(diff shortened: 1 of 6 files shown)


packages/core/src/RadioList/RadioListItem.tsx
```diff
@@ -253,15 +264,18 @@ export function RadioListItem({
...
+        // Delegate row-surface clicks (label text, description, and the empty
+        // hover area) to the radio input. The input stays the option's sole
+        // focusable control, so the row adds no second tab stop.
+        interactiveRef={radioRef}
+        isDisabled={isDisabled}
         label={
-          <label
-            htmlFor={id}
-            {...stylex.props(isDisabled && styles.labelDisabled)}>
+          <span {...stylex.props(isDisabled && styles.labelDisabled)}>
```
cixzhang: This stacks with the Item's 0.5, so the label reads lighter than the checkbox row's.

packages/core/src/RadioList/RadioList.test.tsx
```diff
@@ -128,16 +144,17 @@ describe('RadioList', () => {
...
     render(
       <RadioList label="Preference" value="" onChange={handleChange} isDisabled>
         <RadioListItem label="Option A" value="a" />
       </RadioList>,
     );
 
-    await user.click(screen.getByLabelText('Option A'));
+    // Use fireEvent since userEvent correctly blocks pointer-events: none —
+    // the disabled row is a delegate click surface with pointer-events off.
+    fireEvent.click(screen.getByLabelText('Option A'));
```
cixzhang: `fireEvent` clicks through `pointer-events: none`; `userEvent` was what proved the guard.

cixzhang (dismissed):
Thanks — checked the row in Chrome and the description, empty area, keyboard and nested button all behave.

Two before merge. The disabled label now dims twice: `labelDisabled` stacks with the Item's own 0.5 opacity, so it lands at 163,163,163 where the equivalent `CheckboxListItem` row stays readable. Drop the extra dim and let the Item carry it.

And the switch to `fireEvent.click` gives up the thing this PR relies on — `userEvent` refuses to click through `pointer-events: none`, `fireEvent` doesn't. Delete the guard and that test still passes.

[Reviewed by Robohands]

freddymeta:
Thanks — both addressed:

- **Double dim:** dropped the `labelDisabled` style entirely; the label is now a plain `<span>`, so the disabled dimming comes only from `Item`'s 0.5 opacity — same as `CheckboxListItem`. (Removed the now-unused `labelDisabled` rule + `colorVars` import.)
- **Disabled test:** reverted both disabled-guard tests to `userEvent`. Since `userEvent` honors `pointer-events: none` it *refuses* the click and throws, so the tests now assert `await expect(user.click(...)).rejects.toThrow(/pointer-events/i)` — which genuinely proves the disabled delegate surface blocks the click, rather than the handler merely ignoring it.

Full suite green (45/45), build/typecheck/lint clean.

cixzhang (approved):
Thanks — both fixes hold up. The disabled row now dims exactly like `CheckboxListItem` (both land on rgb(139,139,139)), and the disabled test proves the guard again: I deleted Item's `pointer-events: none` locally and both tests fail.

Approving, but `themingTargets` is red so you won't be able to merge until it's green. The row passes `size`/`selected`/`disabled` to `themeProps`, but the doc still lists `astryx-radio-list-item` bare. One line, mirroring `multi-selector-option`:

```js
{className: 'astryx-radio-list-item', visualProps: ['size'], states: ['selected', 'disabled']},
```

PR #5147, opened 2026-08-17
fix(core): keep tooltip status button interactive in TextArea trailing slot

freddymeta:
## Summary

When an input's `statusVariant="tooltip"`, the on-field status icon becomes a focusable `<button>` that reveals the status message on hover, keyboard focus, or tap (via `useInputStatusIcon`). In `TextArea`, hover never opened that tooltip — only keyboard focus did.

**Root cause:** `TextArea` is multi-line, so it positions its trailing content (`endSlot`) as an absolute overlay with `pointer-events: none` — correct for the decorative spinner and the plain status icon it was originally built for, but it also swallowed pointer events for the *interactive* tooltip button introduced with the tooltip status variant. The single-line inputs render the status affordance in normal flow, so they were unaffected.

**Fix:** the status button now declares `pointer-events: auto` on itself in `useInputStatusIcon`, so the interactive control owns its own interactivity regardless of the slot it's rendered into. This fixes `TextArea` and is inert for every other input (they never disabled pointer events on the slot).

## Test plan

- `pnpm -F @astryxdesign/core test` — `useInputStatusIcon` and `TextArea` suites pass. Added a test asserting the tooltip status button resolves to `pointer-events: auto` inside `TextArea`.
- `pnpm -F @astryxdesign/core typecheck` / `eslint` — clean.
- No API or markup change; a11y behavior (focusable button, tab order, focus ring, described-by wiring) is unchanged — only the pointer path is restored.

### DIFF
--- a/.changeset/textarea-tooltip-status-pointer-events.md
+++ b/.changeset/textarea-tooltip-status-pointer-events.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Inputs (`statusVariant="tooltip"`): the focusable status button now opens its tooltip on hover inside `TextArea`, whose absolutely-positioned trailing slot is `pointer-events: none`. Keyboard focus already worked; pointer hover did not.
+
+@freddymeta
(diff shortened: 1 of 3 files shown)


cixzhang (approved):
Thanks — reproduced this on main: Chrome hands the hover to the textarea, and the button never gets it. Nice fix, merging.

[Reviewed by Robohands]

cixzhang (approved):
Re-approving to re-fire the review gate — no change to my review. The Review clear workflow hit a transient 404 reading this PR's reviews from the GitHub API, so it logged "no entitled code owner approval" and left the `review-required` status pending.

PR #5148, opened 2026-08-17
feat(DateTimeInput): expose toggle-icon and clock-icon theme targets

freddymeta:
## Summary

`DateInput` exposes `date-input-toggle-icon` — a stable theme target on its leading calendar glyph, so a theme can size and color that icon (and target its open/closed state) without a brittle structural selector. `DateTimeInput` has the same two leading glyphs — a calendar toggle on the date segment and a clock on the time segment — but exposed **no** per-glyph target, only the segment-level `date-time-input-date-segment` / `-time-segment`. A theme could not reach the icons to, for example, size them down in a compact layout.

This adds the missing targets, mirroring the `DateInput` seam:

- **`date-time-input-toggle-icon`** — on the calendar glyph, with a `data-state` of `expanded` / `collapsed` reflecting the popover, exactly like `date-input-toggle-icon`.
- **`date-time-input-clock-icon`** — on the leading clock glyph. The time segment has no toggle button (the clock is a static leading affordance), so this carries no interactive state.

The targets land on the icon element itself, so a same-element rule in `@layer astryx-theme` wins over the icon's own base color/size — which a segment-level target can't reach. Purely additive: the classes and `data-state` change nothing until a theme targets them.

## Test plan

- `pnpm -F @astryxdesign/core test` — `DateTimeInput` suite passes, plus three added tests: the toggle-icon target lands on the calendar glyph and flips `data-state` on open, the clock-icon target lands on the time glyph, and a `defineTheme` generation check proves a theme reaches both (size + per-state color).
- `pnpm -F @astryxdesign/core build` / `typecheck` — clean. `astryx component DateTimeInput` lists both new targets.
- Docs (`.doc.mjs` theming targets, EN + dense) and a changeset are included.

### DIFF
--- a/packages/core/src/DateTimeInput/DateTimeInput.tsx
+++ b/packages/core/src/DateTimeInput/DateTimeInput.tsx
@@ -960,7 +960,20 @@ export function DateTimeInput({
               styles.iconButton,
               isEffectivelyDisabled && styles.iconButtonDisabled,
             )}>
-            <Icon icon="calendar" size="sm" color="secondary" />
+            <Icon
+              icon="calendar"
+              size="sm"
+              color="secondary"
+              // Stable theme target on the calendar toggle glyph, so a theme
+              // can restyle just this icon (color, size, hover) — and each
+              // open/closed state — via `defineTheme`, mirroring
+              // `date-input-toggle-icon`. Same-element rules in
+              // @layer astryx-theme win over the icon's own base color/size,
+              // which a segment-level target could not reach.
+              {...themeProps('date-time-input-toggle-icon', {
+                state: popover.isOpen ? 'expanded' : 'collapsed',
+              })}
+            />
           </button>
           <input
             ref={mergeRefs(ref, dateInputRef)}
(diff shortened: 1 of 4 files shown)


packages/core/src/DateTimeInput/DateTimeInput.tsx
```diff
@@ -960,7 +960,20 @@ export function DateTimeInput({
...
+              icon="calendar"
+              size="sm"
+              color="secondary"
+              // Stable theme target on the calendar toggle glyph, so a theme
+              // can restyle just this icon (color, size, hover) — and each
+              // open/closed state — via `defineTheme`, mirroring
+              // `date-input-toggle-icon`. Same-element rules in
+              // @layer astryx-theme win over the icon's own base color/size,
+              // which a segment-level target could not reach.
+              {...themeProps('date-time-input-toggle-icon', {
```
cixzhang, line 973: DateInput ships a themed-toggle story; one here would make the seam discoverable.

packages/core/src/DateTimeInput/DateTimeInput.tsx
```diff
@@ -1039,7 +1052,16 @@ export function DateTimeInput({
...
-            <Icon icon="clock" size="sm" color="secondary" />
+            <Icon
+              icon="clock"
+              size="sm"
+              color="secondary"
+              // Stable theme target on the leading clock glyph, so a theme can
+              // restyle just this icon (color, size) via `defineTheme`. The
+              // time segment has no toggle button — the clock is a static
+              // leading affordance — so this carries no interactive state.
+              {...themeProps('date-time-input-clock-icon')}
```
cixzhang, line 1063: TimeInput renders the same clock glyph with no target, worth the same seam eventually.

cixzhang (approved):
Thanks — themed both icons in Chrome and they take size and per-state color; names line up with the DateInput and DateRangeInput seams. Merging, two optional nits.

[Reviewed by Robohands]

PR #5150, opened 2026-08-17
fix(core): animate the Dialog exit

cixzhang:
## Why

`Dialog`'s entrance animates — it fades and scales in from the direction of its trigger, over `--duration-medium-max`. Its dismissal did not: the open/close effect called `dialog.close()` the moment `isOpen` went false, and `HTMLDialogElement.close()` is immediate.

Measured on `main` at the Storybook default theme, per animation frame in Chromium: the entrance spends **26 frames / ~490ms** interpolating opacity and scale, and the exit spends **zero** — one frame at `display: flex`, the next at `display: none`. Our own motion guidance names this exact case ("animate the exit ... like a panel closing or a dialog dismissing to reveal what's underneath", and "when you do animate exit, match the entrance"), and `MobileNav` in this same package already holds its close so its slide-out can play.

This is the first half of the gap in the gap report below. See "Not addressed" for the rest.

## What

Hold the open `<dialog>` for the length of an exit animation, then close it.

- The exit is the entrance run backwards — opacity to 0, scale to 0.95, translating back toward the element that opened the dialog — at `--duration-medium-min`, the motion scale's "quick exit" band, rather than the entrance's dramatic one.
- The scrim fades on the same curve. Without that it stayed fully opaque for the whole exit and then snapped, which reads as a grey flash after the panel has already gone.
- The hold is read off the dialog's own computed `animationDuration` once the exit style is on it, so a theme that retimes its motion scale retimes the close with it. It is capped at 400ms.
- The hold is skipped entirely — same-tick close, byte-for-byte the old behaviour — under `prefers-reduced-motion`, when a theme zeroes the duration, and anywhere the duration cannot be read (jsdom, an unresolved `var()`).
- The body scroll lock now spans the exit too; releasing it while the dialog was still on screen would scroll the page underneath it.

Whether the dialog is exiting is derived during render, not in an effect: deciding it in an effect costs a commit, and the dialog was measurably blinking out for one frame before it began to animate.

## Risk

The real change is that a dismissed dialog stays modal ~300ms longer than it used to. For that window the page behind is still inert and focus has not yet returned to the trigger. `AlertDialog` and `CommandPalette` inherit it, since both are `Dialog` underneath.

Two things bound it: the cap, and the fact that every path which cannot see a real animation degrades to the old instant close — including every jsdom test in this repo, which is why the suite is unchanged and also why unit tests cannot cover the held path. That path is covered by frame-level measurement in a browser instead (below).

Consumers that unmount the dialog the instant `isOpen` goes false still get no exit animation; nothing regresses for them, they just see no change.

[#4865](https://github.com/facebook/astryx/pull/4865) and [#4881](https://github.com/facebook/astryx/pull/4881) both touch `Dialog.tsx`, in its Escape/dismissal machinery rather than its open/close effect — textually adjacent, semantically disjoint, but whichever lands second will want a look.

## Testing

Chromium, per animation frame, Storybook `core-dialog--default`:

| | before | after |
|---|---|---|
| exit frames with the dialog partly faded | 0 | 12 |
| dialog still on screen after dismissal | 0ms | ~215ms, then closed |
| `prefers-reduced-motion: reduce` | closes at once | closes at once, no partial frame |

Also checked in Chromium: nested dialogs peel off one Escape at a time and the outer one is immediately usable after the inner one goes (the delayed close does not strand the page inert); `AlertDialog`'s confirm still fires and unpins the body; scroll position survives a dialog on both a scrolling document (500 → 500) and a fixed app shell with an inner scroller (745 → 745).

`DialogExitAnimation.test.tsx` covers the contract around the animation: the length of the hold, the cap, both skip paths, reopening mid-exit, unmounting mid-exit, and focus returning to the trigger only after the dialog has actually closed. Full `core` + `lab` suites: 7209 passed.

## Not addressed

Two of the three things the report asks for are deliberately left out.

**The scroll lock has no opt-out, and this PR does not add one.** The follow-up measurement already withdrew the damage claim, and I reproduced that: in a fixed-height app shell the lock moves the inner scroller by 0px, and the body's border box is unchanged to the pixel. What is left is a toggle with no measured use case, which is the shape [API Conventions](https://github.com/facebook/astryx/wiki/API-Conventions#escape-hatches) tells us not to add ("prefer enforced good practices over toggles ... unless there's a real use case for disabling it"). Flagged for @cixzhang rather than decided here.

**There is no `drawer` variant here, and I don't think `Dialog` should grow one.** `Drawer` already exists in `@astryxdesign/lab`: edge-anchored to any of the four sides, directional enter *and* exit, LIFO Escape stack, optional scrim. At 390x844, `side="end"` renders 0,0,390,844 in the top layer — the full-screen phone panel the report wanted. That is a system-design call for the TL, so it is a question for the TL, not a PR.

### DIFF
--- a/packages/core/src/Dialog/Dialog.tsx
+++ b/packages/core/src/Dialog/Dialog.tsx
@@ -29,6 +29,7 @@ import {
   useId,
   useMemo,
   useRef,
+  useState,
   type ReactNode,
 } from 'react';
 import type {BaseProps} from '../BaseProps';
@@ -153,6 +154,25 @@ export const dialogFullscreenSafeAreaPaddingContract = {
   },
 } as const;
 
+// The entrance run backwards, so the dialog retreats toward the element that
+// opened it (motion docs: "when you do animate exit, match the entrance").
+const exitDirectional = stylex.keyframes({
+  from: {opacity: 1, transform: 'translate(0, 0) scale(1)'},
+  to: {
+    opacity: 0,
+    transform:
+      'translate(var(--dialog-dir-x, 0px), var(--dialog-dir-y, 16px)) scale(0.95)',
+  },
+});
+
+// The scrim leaves on the same curve as the panel. Without this it stays
+// opaque for the whole exit and then snaps, which reads as a grey flash after
+// the dialog has already faded out.
+const exitBackdrop = stylex.keyframes({
+  from: {opacity: 1},
+  to: {opacity: 0},
+});
+
 /**
  * Dialog styles using native <dialog> element
  * Uses ::backdrop pseudo-element for overlay
@@ -175,6 +195,11 @@ const styles = stylex.create({
     animationDuration: durationVars['--duration-medium-max'],
     animationTimingFunction: easeVars['--ease-standard'],
     animationFillMode: 'backwards' as const,
+    // Read back by resolveExitDelay() to time dialog.close() to the end of the
+    // slide-out. Declared here (not only in `exiting`) so the value is
+    // readable from the still-open dialog, and as a token so a theme that
+    // retimes its motion scale retimes the close with it.
+    '--dialog-exit-duration': durationVars['--duration-medium-min'],
   },
   // Applied via isOpen prop — avoids :where([open]) attribute selectors
   // which have zero specificity and can lose to default styles depending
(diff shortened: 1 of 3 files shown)


cixzhang (commented):
Thanks — this is good, and the skip paths especially: reduced motion closes in 14ms, a theme that zeroes the duration in 42ms.

One thing I'd hold undrafting for, and it came in with the rebase. An exiting dialog stays registered on the dismissal stack as a layer that swallows Escape, so for ~280ms after any dismissal Escape does nothing. On main two presses 100ms apart peel both nested dialogs; here the outer survives until the gap passes 300ms. Ideally I'd like the exit animation not to take Escape with it. `MobileNav` leaves the stack the moment it starts closing, and our motion doc says "motion should never stand between the user and their next action".

If we do want it swallowed, can the reason go in the body? It's the one decision here with no problem written down.

[Reviewed by Robohands]

packages/core/src/Dialog/Dialog.tsx
```diff
@@ -560,10 +723,12 @@ export function Dialog({
...
   // Inline mode opts out — it renders dialog content in normal flow, with
   // nothing layered over anything.
+  // Keep an exiting dialog in the stack while it remains modal. Additional
+  // Escape presses are consumed without dismissing the layer underneath.
   const {shouldDismissOnCloseRequest} = useLayerDismissal({
-    isActive: isOpen,
+    isActive: isOpen || isExiting,
     isEnabled: !isInline,
-    escapeBehavior: allowEscape ? 'close' : 'block',
+    escapeBehavior: isExiting || !allowEscape ? 'block' : 'close',
```
cixzhang: This is what swallows the second Escape. `MobileNav` keeps `isActive: isOpen` and drops off the stack instead.

cixzhang:
Closing this Dialog-owned exit hold in favor of the shared presence foundation in [#5574](https://github.com/facebook/astryx/pull/5574). The animation itself is valid, but presence lifetime, nested dismissal, focus return, and reduced-motion behavior should have one owner before Dialog builds on it.

PR #5152, opened 2026-08-17
fix(NumberInput): commit an out-of-range entry at the nearest bound

cixzhang:
## Why

Reported by @aldentan: with a `max` set, typing a number past it leaves the *first digit or two* committed instead of the max. With `max={2}`, typing `100` lands on **1**. Pagination shows this as jumping to page 1 rather than the last page — the previous treatment (clamp to max) is more intuitive, and that's what people expect from a bounded field.

The cause: every keystroke commits if it parses in range, so `1` commits, then `10` and `100` are rejected outright and the stale `1` is what survives the blur.

## What

`min`/`max` violations now clamp to the nearest bound **at commit time** — blur and Enter. Typing is deliberately untouched: the entry stays exactly as typed (and styled invalid) until committed, so nothing rewrites the field mid-edit. Text that isn't a usable number (`abc`, or a decimal under `isIntegerOnly`) still reverts, unchanged.

The bound is put through the field's own shape rule before it is committed, so an `isIntegerOnly` field handed a fractional bound rounds inwards (`max={9.5}` commits 9, not 9.5), and a range no number can satisfy (`min` above `max`) reverts rather than committing outside both ends.

Enter also clears the pending text when the commit clamped. That is more than cosmetic: before, Enter left the field reading `100` on a value of 1, with `aria-invalid` still set and an assertive "Invalid number" still announced, and no way out but retyping.

Pagination inherits the fix — it already delegates all bounding to NumberInput.

## Risk

Behavior change on commit, patch-level: an out-of-range entry that previously reverted now fires `onChange` with the bound. Consumers relying on "reject and revert" for range violations would see a value they didn't before. The committed bound always satisfies both the range and `isIntegerOnly`, so it can't produce a value the field's own rules reject.

Not fixed here: the per-keystroke commit itself, so someone on page 3 typing `10` still visits page 1 for one keystroke before landing on 10. Pre-existing, and a larger change to where the commit lives.

## Testing

- `range clamping on commit` suite: over-max on blur, over-max on Enter, below-min, no clamping while typing, displayed value after commit, non-numeric entries still reverting, fractional bounds rounded inwards under `isIntegerOnly`, and an unsatisfiable range reverting. Each new case fails without the change.
- Updated the Pagination over-range test to assert page 10 rather than only "not 99".
- `NumberInput` + `Pagination`: 364 tests pass.
- Verified in Chrome against local Storybook (screenshots in the comments).

### DIFF
--- a/.changeset/numberinput-clamp-to-bound.md
+++ b/.changeset/numberinput-clamp-to-bound.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] NumberInput: an entry past `min`/`max` now commits at the nearest bound on blur or Enter instead of falling back to the last in-range prefix — typing 100 into a 1–2 field lands on 2, not 1. Pagination inherits this.
+
+@cixzhang
--- a/packages/core/src/NumberInput/NumberInput.doc.mjs
+++ b/packages/core/src/NumberInput/NumberInput.doc.mjs
@@ -128,12 +128,14 @@ export const docs = {
     {
       name: 'min',
       type: 'number | null',
-      description: 'Minimum value allowed.',
+      description:
+        'Minimum value allowed. A smaller entry commits at this value on blur or Enter.',
     },
     {
       name: 'max',
       type: 'number | null',
-      description: 'Maximum value allowed.',
+      description:
+        'Maximum value allowed. A larger entry commits at this value on blur or Enter.',
     },
     {
       name: 'step',
@@ -410,12 +412,12 @@ export const docsZh = {
     {
       name: 'min',
       type: 'number | null',
-      description: '允许的最小值。',
+      description: '允许的最小值。更小的输入会在失焦或按 Enter 时提交为该值。',
     },
     {
       name: 'max',
       type: 'number | null',
-      description: '允许的最大值。',
+      description: '允许的最大值。更大的输入会在失焦或按 Enter 时提交为该值。',
     },
     {
       name: 'step',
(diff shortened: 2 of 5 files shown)


cixzhang:
Verified in Chrome against local Storybook (`core-numberinput--with-min-max`, min 1 / max 5): typing `100` leaves the field at `100` with `aria-invalid="true"`; blur commits `5`; `-7` + Enter commits `1`; `abc` + blur reverts to the committed value. The Enter case needed the second commit here — Enter does not clear the pending text, so the field kept showing the rejected entry after the value had clamped. 213 tests pass across NumberInput + Pagination.

cixzhang:
Thanks — drove the integer-only bound and the empty range in Chromium; both land inside the field's own rules now. Approve, comment only because it's mine.

### `100` into a 1–5 field, blurred

| before | after |
|---|---|
|  |  |
| falls back to `1` | settles on `5` |

### The same entry, committed with Enter

| before | after |
|---|---|
|  |  |
| field reads `100` on a value of `1`, `aria-invalid` still set, "Invalid number" still announced | field reads `5`, invalid cleared |

### `isIntegerOnly` with `max={9.5}`, typing `20`

| first two commits | with the bound fix |
|---|---|
|  |  |
| commits `9.5` out of an integer-only field | rounds inwards to `9` |

`Pagination.tsx:555-556` already documented NumberInput as "clamps to [min, max] with integer-only semantics, so it only ever hands back a valid page here." Before this PR neither half was true; at this head the sentence is true as written.

Not this PR's, and pre-existing on both sides — `getSteppedValue` rounds for precision *after* clamping, so ArrowUp in a `max={99.99}` field commits `100` with `aria-invalid` unset. Filed as #5506.

[Reviewed by Robohands]

PR #5154, opened 2026-08-17
feat(markdown): renderBlock — give overrides the source line range of each block

cixzhang:
## Why

A file viewer wants to render a changed markdown file as prose and mark which blocks were added or rewritten — the way a rich code-review UI shows a changed document — instead of dropping the reader into a raw `+`/`-` column. That mapping is expressed in **source line ranges**, and today no override in `Markdown` receives any source information, so there is no sound way to say "this rendered paragraph is lines 12–18". Matching on rendered text collides on repeated blocks; pre-splitting the source per hunk breaks any list or table that straddles a hunk boundary.

Worth stating, because the gap report assumed otherwise: Astryx's `Markdown` does **not** use remark/mdast. `parser.ts` is a hand-rolled line parser whose nodes carry no `position` at all, so this is not "surface what the parser already knows" — the positions had to be computed and threaded through the recursive parses (blockquote bodies, list-item bodies) and through link-reference stripping, which removes source lines.

## What

- Parser: opt-in `trackPosition` records `position: {startLine, endLine}` (1-based, inclusive) on every block node. Off by default, so parsing cost and AST shape are unchanged for everyone else.
- Component: `renderBlock?: (block, children) => node` wraps each rendered block, given `{type, position, depth}`. It runs for nested blocks too, so a single rewritten bullet can be marked without tinting its whole list; `depth` distinguishes them. Single-paragraph list items — which render as an inline label and so never reach the block recursion — are decorated at the same shape, otherwise per-bullet indicators would silently not work.

## The API surface, and what it deliberately does not expose

`renderBlock` was chosen over the two obvious alternatives:

- **Pass the AST node** (or a `node` prop on each override). Maximally useful, and it makes `BlockNode`'s shape a render-contract we can never change — for a use case that needs two integers.
- **Add `position` to the nine `MarkdownComponents` overrides.** Two mechanisms for one job, and it still doesn't cover the blocks most likely to carry an indicator: there is no `list` or `table` override to put it on. One decorator that sees *every* block, including lists and tables, is smaller and more complete.

Deliberately not exposed: the AST node, byte offsets, columns, per-table-row and per-table-cell ranges (a changed row marks the table), and any new component-override types. `MarkdownBlockType` is the block-kind union, which `BlockNode` already exports today.

## Risk

Anything shipped here is public API. The surface is two integers plus a block-kind string and a depth — the smallest thing that drives the use case — and it is additive: omit the prop and both parsing and markup are byte-identical (asserted in a test). One trade recorded in code: with `trackPosition` on, streaming skips the incremental cache and re-parses, because cached blocks carry line numbers relative to a text slice. Positions serve static views, so a second cache-aware offset path was not worth its bug surface.

The renderer's node types are the synchronous subset of `ReactNode`. React 19's `ReactNode` admits a Promise, which makes `(block, children) => children` read as "returns a promise" and get auto-fixed to `async` by lint — caught here by that autofix silently rewriting the tests.

## Testing

- Parser: line ranges for headings, lazy-continued paragraphs, lists, blockquotes, a fenced block containing a blank line and a `#` line, tables, `hr`; nested blocks reporting original-document lines; ranges verified by slicing the source back out (off-by-one guard); shifted lines after a stripped link definition; streaming parity.
- Component: renders identical markup when the renderer passes children through, reported ranges and depths for a whole document, the diff-indicator path end to end (changed lines in → the right blocks marked, the untouched bullet and intro not), and composition with existing `components` overrides.
- Storybook: `Core/Markdown → Diff Indicators` runs a real LCS line diff of a before/after spec and marks the result. Verified in Chromium, not just jsdom: the rewritten bullet, the added quote line and the changed code block are marked; the unchanged bullets, heading and intro are not.

### DIFF
--- a/.changeset/markdown-render-block.md
+++ b/.changeset/markdown-render-block.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] Markdown: `renderBlock` wraps each rendered block with the source line range it came from, so rendered markdown can carry diff, change, or blame indicators. Parsing is unchanged unless the prop is set.
+
+@cixzhang
--- a/apps/storybook/stories/Markdown.stories.tsx
+++ b/apps/storybook/stories/Markdown.stories.tsx
@@ -445,3 +445,126 @@ export const InlinePlugins: Story = {
     );
   },
 };
+
+const SPEC_BEFORE = [
+  '# Session file viewer',
+  '',
+  'Renders a file an agent wrote during the session.',
+  '',
+  '## Supported types',
+  '',
+  '- Plain text',
+  '- Source code, syntax highlighted',
+  '- Images',
+  '',
+  '## Open questions',
+  '',
+  '> How do we show a file that changed twice in one session?',
+  '',
+  '```ts',
+  'type FileView = {path: string; body: string};',
+  '```',
+].join('\n');
+
+const SPEC_AFTER = [
+  '# Session file viewer',
+  '',
+  'Renders a file an agent wrote during the session.',
+  '',
+  '## Supported types',
+  '',
+  '- Plain text',
+  '- Source code, syntax highlighted',
+  '- **Markdown**, rendered as prose (`.md`)',
+  '- Images',
+  '',
+  '## Open questions',
+  '',
+  '> How do we show a file that changed twice in one session?',
+  '> Mark the second change only, or both?',
+  '',
+  '```ts',
+  'type FileView = {path: string; body: string; language?: string};',
+  '```',
+].join('\n');
+
+/** Line numbers (1-based) in `after` that are new or rewritten. */
+function changedLines(before: string, after: string): Set<number> {
+  const a = before.split('\n');
+  const b = after.split('\n');
+  // Longest common subsequence over lines — the same shape a real diff uses.
+  const lcs: number[][] = Array.from({length: a.length + 1}, () =>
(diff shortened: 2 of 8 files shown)


cixzhang:
Closing this — the source-position design landed a different way.

[#5290](https://github.com/facebook/astryx/pull/5290) went in with character offsets (`sourceRanges` → `range`) for the copy-out case. The investigation behind that choice is worth keeping, so recording it here rather than losing it with the branch:

- **Lines were the right call for a diff viewer** and that reasoning holds — diffs and blame are line-addressed, and nested blocks *can only* have lines, since a blockquote child's text isn't contiguous in the source. [#5290](https://github.com/facebook/astryx/pull/5290) is top-level-only for exactly that reason.
- **This branch's `trackPosition` invalidates the incremental cache**: 71.8ms vs 34.9 on a 14k document, 2× and growing, and `renderBlock` turned it on implicitly. [#5290](https://github.com/facebook/astryx/pull/5290)'s incremental path is the one to build on.
- **Both units together are free** — 36.7ms vs 36.2 for offsets alone.

So when a diff-viewer consumer actually exists, lines go on top of `SourceRange` as `startLine`/`endLine`, additively, along with nested coverage. `SourceRange` is being reshaped from a tuple to an object before the 0.5.0 cut precisely so that stays possible.

[Reviewed by Robohands]

PR #5158, opened 2026-08-17
fix(core): keep the page still when the mobile keyboard reveals a Bottom Sheet field

imdreamrunner:
## The bug

A Tall Bottom Sheet holding a form is pinned to the viewport, so it should not
move while you type. On iOS it moved: focus a field low in the form and the
whole page lurched upward, taking the sheet's header and scrim off screen with
it. It reproduced roughly one time in ten, which read as flaky rather than as a
specific unhandled case.

It is not flaky. Three separate things were wrong.

## 1. The reveal is attached to focus — claim the transition

The browser reveals a newly focused field by scrolling the **document**, and a
`position: fixed` sheet rides along. Earlier revisions of this PR tried to get
in front of that scroll, and to undo it afterwards. Both are races: the reveal
runs on the compositor, so a main-thread correction can land a frame late and
read as *a pan followed by a snap back*.

The reveal is attached to the **focus operation**, which is where it can be
refused. `blur` is dispatched in the capture phase ahead of the browser's own
focus step and names the destination in `relatedTarget`:

```ts
document.addEventListener('blur', event => {
  const destination = findTextEntryControl(event.relatedTarget, body);
  if (destination && destination !== document.activeElement) {
    destination.focus({preventScroll: true});   // transition settled here
  }
}, true);
```

The browser's focus step then finds the element already active, dispatches
nothing, and has nothing to reveal. A tap, the keyboard's Next, Tab and a
programmatic `focus()` all arrive as that one transition — so the separately
special-cased tap path is gone, along with its pending-tap state machine,
movement threshold and multi-touch guards.

Revealing the field is left to the existing `focusin` / viewport-resize path,
which already knows the safe area and scrolls only the sheet's own body.

Presenting a sheet that **autofocuses** a field is not a transition — nothing
was focused, so no blur names it. `focusPanel` therefore passes `preventScroll`
itself; the sheet was about to show that field anyway, and the hook brings it
into view within the sheet.

A blur that names **no** destination is the keyboard's Done button. Left alone,
focus lands on the body, the field stays `document.activeElement`, and the next
tap on it produces no transition at all — so the browser reveals it its own way.
Focus is parked on the sheet instead, keeping that tap claimable. Only while the
sheet is open: closing blurs the field too, and there the host is restoring
focus to whatever opened the sheet.

### This is #5043's B1, re-tested rather than honoured

#5043 rejected blur-phase `focus({preventScroll: true})` for double-firing
`focus`/`blur` (finding `B1`). Measured on an iPhone 17 simulator, counting real
events with the mechanism on and off:

| transition | off | on |
|---|---|---|
| tap A → tap B | 1 focus / 1 blur each | **identical** |
| accessory-bar Previous, then Next | 1 each | **identical** |
| programmatic `a.focus(); b.focus()` | 1 each | **identical** |

B1 does not reproduce in WebKit.

## 2. The keyboard boundary was read in a way a shift corrupts

`visualViewport.offsetTop + visualViewport.height` was the "where does the
keyboard start" measure in four places. It is wrong the moment the page is
shifted: the shift slides the window over the page, but the keyboard stays where
it is, so the obstruction appears to shrink as the shift grows.

A fully expanded Tall sheet is pinned to the layout viewport bottom, so
`bodyBottom == innerHeight`. At full shift that expression also equals
`innerHeight`, so the overlap computes as **zero**: the sheet concluded the
keyboard was gone, dropped the scroll range it had added, and cleared the flag
that arms the defences. One shift latched it into the unprotected behaviour
permanently — which is exactly the reported symptom that Next works down a form
until something shifts once, and never recovers after.

Measured with `bodyBottom = innerHeight = 768` and a 500px keyboard: **316px →
0px**. Every reader now takes `visualViewport.height` alone, which a shift does
not change. Identical when nothing is shifted.

## 3. Putting back a document scroll (a net, for one case)

App resume raises **no focus event at all**: leave Safari with a field focused,
come back, and the browser reveals it again with nothing to claim. The
document's own `scroll` event is the earliest signal, so the scroll is put back
there, synchronously, and the field re-revealed inside the sheet.

This runs **only while the page is locked** — the one state in which a document
scroll cannot be the user's own. Behind a non-modal sheet (`hasScrim={false}`)
the page stays scrollable by design, and an earlier revision of this PR reverted
the user's own scrolling there; the host now passes `isPageScrollLocked`.

## Scope: fully expanded Tall only

Prevention lives behind the same gates as keyboard accommodation —
`height="tall"` at full expansion, iOS WebKit. A `capped` sheet (the default),
or a Tall sheet dragged to a shorter detent, still lets the browser reveal the
field its own way. That is a deliberate decision for now, and the `height` prop
docs now say so rather than only promising no keyboard scroll space.

## Tests

`packages/core/src/BottomSheet` — 161 passing, 10 new, 5 removed with the tap
machinery they covered. Every guard the fix adds fails a test when removed:

| Remove this | and this test fails |
|---|---|
| the claim listener | claims a transition between fields… (+4 more) |
| `preventScroll` on the claim | claims a transition between fields and delivers it with preventScroll |
| Done-button parking | parks focus on the sheet when the keyboard Done button takes it |
| the open-sheet gate on parking | does not park focus on a sheet that is closing |
| the scroll-lock gate on the net | leaves the page alone behind a non-modal sheet |
| the document-scroll net | puts back a document scroll the browser makes to reveal a field |
| shift-invariant boundary | holds the keyboard scroll range through a pan · keeps defending a pinned sheet after the browser has panned it once |
| `preventScroll` in `focusPanel` | autofocuses a field without letting the browser reveal it |
| the full-expansion gate | does not accommodate the keyboard at a shorter Tall detent |

`pnpm lint:strict` (0 errors) and `pnpm build` are clean. On `pnpm test`,
10,668 of 10,671 pass; one failure passes on re-run, and the two that persist —
CLI name resolution and hook discovery — are the same two `main` fails on this
machine, both case-insensitive-filesystem artifacts.

## Device evidence

The real story driven on an iPhone 17 simulator (iOS 26.5). Sheet top edge
decoded from the scrim boundary in each screenshot; 376px is the unshifted
position:

| | sheet top |
|---|---|
| field focused, keyboard up | 376px |
| after six accessory-bar Nexts | 376px |
| **scrolled to the end, last field of eleven focused** | **376px** |
| four accessory-bar Previous from there | 376px |
| across an app switch | 376px |

Pre-fix the same sequence put the sheet top off screen entirely — no scrim
boundary left to find — and it stayed there. Whole-frame comparison across the
app switch: 4.3% of pixels differed pre-fix, 0.0% with it.

The deep end of the form is included above: the last field sits above the
keyboard with the submit button still visible, and nothing moved.

**Still worth a physical-device pass** for what the simulator does not model:
Safari with browser UI expanded vs collapsed, PWA standalone, and a real
software keyboard's timing. Test
`/iframe.html?id=core-bottomsheet--mobile-keyboard` directly — focus-scroll
prevention does not work inside an iframe, so the Storybook manager shell gives
a false negative.

## Measured, not assumed

**Caret placement.** Claiming the focus ahead of the click could cost the tap's
caret position, so it was measured in Mobile Safari — tap into the middle of a
populated field with a prior field focused:

| | caret offset |
|---|---|
| claim on | 40 of 40 |
| claim off (native) | 40 of 40 |

Identical: the click's hit test still owns the caret. A tap into a field that is
already focused cannot be affected at all, since the claim no-ops when the
destination is already active.

Found while investigating, each its own change:

- **React's `autoFocus` on a field inside a sheet** (`<TextInput hasAutoFocus />`
  emits both it and `data-autofocus`). React applies it at `commitMount`, before
  any effect runs, so no listener this hook installs can exist yet and
  `focusPanel` early-returns because focus is already inside the panel. Not
  reachable from a hook; closing it means the sheet owning autofocus itself.
- **The first focus into a non-modal sheet** (`hasScrim={false}`). Focus starts
  on the body, so there is no blur to claim. Parking focus on the panel would
  close it, but non-modal sheets deliberately leave focus alone — "the
  background stays interactive, so the sheet must not grab focus" is existing,
  tested behavior. The page is not scroll-locked in that shape either, so a
  document scroll there is ordinary page scrolling rather than a fixed sheet
  being dragged off screen.

- **`<select>` and password fields.** `isTextEntryControl` does not match
  `<select>`, whose iOS picker also shrinks the viewport; prevention is
  unreliable for password inputs without `autocomplete="current-password"`.
- **Chromium.** The claim is gated on iOS WebKit. A similar race exists on
  Android and needs its own validation.
- **Caret-driven reveal while typing** in a tall `TextArea` — WebKit follows the
  caret, this hook reveals the element.
- **Re-tapping a field the keyboard was dismissed from without a blur** (iPad's
  hide-keyboard chevron, attaching a hardware keyboard). Measured: that re-tap
  produces **zero** focus events of any kind, so no blur-based mechanism can
  reach it — whatever covers it has to be a pointer or touch path.
- **The inset flaps on resume.** The keyboard is dismissed while the app is
  backgrounded and re-presented on return, so the hook sees no-keyboard and then
  keyboard again: the body's extra scroll range goes 385px → 0 → 385px in ~130
  ms. The page stays still through it, but the flap itself is avoidable.
- **`overscroll-behavior: contain` on nested scrollers.** Scroll chaining is a
  different mechanism from the focus reveal, and nothing here was shown to need
  it; a revision of this PR carried a global rule for it and dropped it as
  unproven. Worth revisiting with a case that fails without it.
- **Drag smoothness.** Reported separately while testing this: the live drag
  routes every pointer move through React state, and the detent magnet means the
  sheet deliberately lags the finger. Neither is touched here.

## Risk

Confined to `useMobileKeyboard`, plus one private prop threaded through
`BottomSheetPanel`. Within the hook, everything but the boundary reading is
gated on iOS WebKit, and all of it on `height="tall"` at full expansion. The 151
pre-existing Bottom Sheet tests pass unchanged.

### DIFF
--- a/.changeset/bottomsheet-keeps-the-page-still-when-the-mobile-xlk2.md
+++ b/.changeset/bottomsheet-keeps-the-page-still-when-the-mobile-xlk2.md
@@ -2,5 +2,5 @@
 '@astryxdesign/core': patch
 ---
 
-[fix] BottomSheet keeps the page still when the mobile keyboard reveals a field the browser focused itself
+[fix] BottomSheet keeps the page still when the mobile keyboard reveals a field the browser focused itself (#5158)
 @imdreamrunner
--- a/.changeset/bottomsheet-keeps-the-page-still-when-the-mobile-xlk2.md
+++ b/.changeset/bottomsheet-keeps-the-page-still-when-the-mobile-xlk2.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] BottomSheet keeps the page still when the mobile keyboard reveals a field the browser focused itself
+@imdreamrunner
--- a/packages/core/src/BottomSheet/BottomSheet.doc.mjs
+++ b/packages/core/src/BottomSheet/BottomSheet.doc.mjs
@@ -99,7 +99,7 @@ export const docs = {
       name: 'height',
       type: "'hug' | 'capped' | 'tall' | number | string",
       description:
-        "How tall the sheet is. Named budgets: 'hug' fits its content up to 92% of the viewport, 'capped' is a scrolling mid-height panel (~62%), and 'tall' is a pinned near-full panel (~92%) for content that streams in. Or pass a number (px) / CSS length for a custom budget. The user can drag between snap points regardless. On shorter viewports the sheet fills the available height. Only a fully expanded 'tall' sheet provides mobile-keyboard accommodation; shorter Tall detents, Hug, Capped, numeric, and CSS-length heights do not move, do not add keyboard scroll space, and leave the browser's own focus reveal in place — on iOS that reveal can pan the page.",
+        "How tall the sheet is. Named budgets: 'hug' fits its content up to 92% of the viewport, 'capped' is a scrolling mid-height panel (~62%), and 'tall' is a pinned near-full panel (~92%) for content that streams in. Or pass a number (px) / CSS length for a custom budget. The user can drag between snap points regardless. On shorter viewports the sheet fills the available height. Only a fully expanded 'tall' sheet provides mobile-keyboard accommodation; shorter Tall detents, Hug, Capped, numeric, and CSS-length heights do not move, do not add keyboard scroll space, and leave the browser's own focus reveal in place — on iOS that reveal can shift the whole page.",
       default: "'capped'",
     },
     {
--- a/packages/core/src/BottomSheet/BottomSheet.doc.mjs
+++ b/packages/core/src/BottomSheet/BottomSheet.doc.mjs
@@ -99,7 +99,7 @@ export const docs = {
       name: 'height',
       type: "'hug' | 'capped' | 'tall' | number | string",
       description:
-        "How tall the sheet is. Named budgets: 'hug' fits its content up to 92% of the viewport, 'capped' is a scrolling mid-height panel (~62%), and 'tall' is a pinned near-full panel (~92%) for content that streams in. Or pass a number (px) / CSS length for a custom budget. The user can drag between snap points regardless. On shorter viewports the sheet fills the available height. Only a fully expanded 'tall' sheet provides mobile-keyboard accommodation; shorter Tall detents, Hug, Capped, numeric, and CSS-length heights do not move, do not add keyboard scroll space, and leave the browser's own focus reveal in place — on iOS that reveal can shift the whole page.",
+        "How tall the sheet is. Named budgets: 'hug' fits its content up to 92% of the viewport, 'capped' is a scrolling mid-height panel (~62%), and 'tall' is a pinned near-full panel (~92%) for content that streams in. Or pass a number (px) / CSS length for a custom budget. The user can drag between snap points regardless. On shorter viewports the sheet fills the available height. Only a fully expanded 'tall' sheet provides mobile-keyboard accommodation: it stays put and scrolls each focused control above the keyboard. Hug, Capped, numeric and CSS-length heights never do, and a Tall sheet stops doing it the moment the user drags it to a shorter detent — resuming when they drag it back. Outside that state the sheet neither moves nor adds keyboard scroll space, and the browser's own focus reveal is left in place; on iOS that reveal can shift the whole page.",
       default: "'capped'",
     },
     {
--- a/packages/core/src/BottomSheet/BottomSheet.doc.mjs
+++ b/packages/core/src/BottomSheet/BottomSheet.doc.mjs
@@ -99,7 +99,7 @@ export const docs = {
       name: 'height',
       type: "'hug' | 'capped' | 'tall' | number | string",
       description:
-        "How tall the sheet is. Named budgets: 'hug' fits its content up to 92% of the viewport, 'capped' is a scrolling mid-height panel (~62%), and 'tall' is a pinned near-full panel (~92%) for content that streams in. Or pass a number (px) / CSS length for a custom budget. The user can drag between snap points regardless. On shorter viewports the sheet fills the available height. Only a fully expanded 'tall' sheet provides mobile-keyboard accommodation; shorter Tall detents, Hug, Capped, numeric, and CSS-length heights do not move or add keyboard scroll space.",
+        "How tall the sheet is. Named budgets: 'hug' fits its content up to 92% of the viewport, 'capped' is a scrolling mid-height panel (~62%), and 'tall' is a pinned near-full panel (~92%) for content that streams in. Or pass a number (px) / CSS length for a custom budget. The user can drag between snap points regardless. On shorter viewports the sheet fills the available height. Only a fully expanded 'tall' sheet provides mobile-keyboard accommodation; shorter Tall detents, Hug, Capped, numeric, and CSS-length heights do not move, do not add keyboard scroll space, and leave the browser's own focus reveal in place — on iOS that reveal can pan the page.",
       default: "'capped'",
     },
     {
--- a/packages/core/src/BottomSheet/BottomSheet.test.tsx
+++ b/packages/core/src/BottomSheet/BottomSheet.test.tsx
@@ -1061,133 +1061,16 @@ describe('BottomSheet', () => {
   });
 
   describe('mobile keyboard', () => {
-    it('prevents native touch focus panning without moving focus early or duplicating events', () => {
+    it('claims a transition between fields and delivers it with preventScroll', () => {
       mockIOSWebKit();
-      const onFocus = vi.fn();
-      const onBlur = vi.fn();
       render(
         <BottomSheet
           isOpen
           onOpenChange={() => {}}
           label="Add a comment"
           height="tall">
-          <input
-            aria-label="Comment"
-            onPointerDown={event => event.stopPropagation()}
-            onFocus={onFocus}
-            onBlur={onBlur}
-          />
-        </BottomSheet>,
-      );
-      const sheet = getSheet();
-      const input = screen.getByRole('textbox', {name: 'Comment'});
-      const focus = vi.spyOn(input, 'focus');
-      sheet.focus();
-
-      // Pointerdown may still become a scroll gesture, so it must not move
-      // focus or open the keyboard early.
-      fireTouchPointer(input, 'pointerdown', {x: 20, y: 200});
-      expect(focus).not.toHaveBeenCalled();
-      expect(onFocus).not.toHaveBeenCalled();
-      expect(document.activeElement).toBe(sheet);
-
-      // Pointerup confirms a tap and prevents focus panning. The subsequent
-      // native click remains responsible for caret placement.
-      fireTouchPointer(input, 'pointerup', {x: 20, y: 200});
-      expect(focus).toHaveBeenCalledTimes(1);
-      expect(focus).toHaveBeenCalledWith({preventScroll: true});
-      expect(document.activeElement).toBe(input);
-      expect(onFocus).toHaveBeenCalledTimes(1);
-      expect(onBlur).not.toHaveBeenCalled();
-
-      // The native click still runs for caret placement and must not produce a
-      // second focus transition.
-      fireEvent.click(input);
-      expect(onFocus).toHaveBeenCalledTimes(1);
-      expect(onBlur).not.toHaveBeenCalled();
-    });
-
-    it('does not focus when a touch becomes a scroll gesture', () => {
-      mockIOSWebKit();
-      const onPreviousBlur = vi.fn();
-      render(
-        <BottomSheet
-          isOpen
-          onOpenChange={() => {}}
-          label="Add a comment"
-          height="tall">
-          <button type="button" onBlur={onPreviousBlur}>
-            Previous action
-          </button>
+          <input aria-label="Title" />
           <input aria-label="Comment" />
-        </BottomSheet>,
-      );
-      const body = getBody();
-      const previous = screen.getByRole('button', {name: 'Previous action'});
-      const input = screen.getByRole('textbox', {name: 'Comment'});
-      const focus = vi.spyOn(input, 'focus');
-      body.scrollTop = 20;
-      previous.focus();
-
-      fireTouchPointer(input, 'pointerdown', {x: 20, y: 200});
-      fireTouchPointer(input, 'pointermove', {x: 20, y: 240});
-      fireTouchPointer(input, 'pointerup', {x: 20, y: 240});
-
-      expect(focus).not.toHaveBeenCalled();
-      expect(document.activeElement).toBe(previous);
-      expect(onPreviousBlur).not.toHaveBeenCalled();
-    });
-
-    it('does not focus when pointer activation is canceled', () => {
-      mockIOSWebKit();
-      const onFocus = vi.fn();
-      const onPreviousBlur = vi.fn();
-      render(
-        <BottomSheet
-          isOpen
-          onOpenChange={() => {}}
-          label="Add a comment"
-          height="tall">
-          <button type="button" onBlur={onPreviousBlur}>
-            Previous action
-          </button>
-          <input
-            aria-label="Comment"
-            onPointerDown={event => event.preventDefault()}
-            onFocus={onFocus}
-          />
-        </BottomSheet>,
-      );
-      const previous = screen.getByRole('button', {name: 'Previous action'});
-      const input = screen.getByRole('textbox', {name: 'Comment'});
-      const focus = vi.spyOn(input, 'focus');
-      previous.focus();
-
-      const pointerDownAllowed = fireTouchPointer(input, 'pointerdown', {
-        x: 20,
-        y: 200,
-      });
-      fireTouchPointer(input, 'pointerup', {x: 20, y: 200});
-
-      expect(pointerDownAllowed).toBe(false);
-      expect(focus).not.toHaveBeenCalled();
-      expect(document.activeElement).toBe(previous);
-      expect(onFocus).not.toHaveBeenCalled();
-      expect(onPreviousBlur).not.toHaveBeenCalled();
-    });
-
-    it('prevents input-to-input focus panning without duplicating focus or blur events', () => {
-      mockIOSWebKit();
-      const onTitleBlur = vi.fn();
-      const onCommentFocus = vi.fn();
-      render(
-        <BottomSheet
-          isOpen
-          onOpenChange={() => {}}
-          label="Add a comment"
-          height="tall">
-          <input aria-label="Title" onBlur={onTitleBlur} />
-          <input aria-label="Comment" onFocus={onCommentFocus} />
         </BottomSheet>,
       );
       const title = screen.getByRole('textbox', {name: 'Title'});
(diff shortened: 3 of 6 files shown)


imdreamrunner:
## Investigation update — this does not hold on device, and I can reproduce why

Device check: accessory-bar **Next** still pans. Not a device quirk — the mechanism is
reproducible in the test harness, and it is the same insight as section 2 above, applied one
reader further.

`viewport.bottom` (`offsetTop + height`) is the "where does the keyboard start" measure in four
places. Section 2 fixes one of them (the recovery comparison). The other three still read it.

### The latch

A fully expanded Tall sheet — the only configuration this hook runs in — is pinned to the layout
viewport bottom, so `bodyRect.bottom <= innerHeight`. At full pan `viewport.bottom == innerHeight`,
exactly as section 2 describes. So in `revealFocusedControl` (L459):

```
overlap = max(0, bodyBottom - viewport.bottom) = 0
  -> hasKeyboardLayoutRef.current = false
  -> inset = 0px            (the extra scroll range is gone)
  -> head start disarmed    (L417 is gated on hasKeyboardLayoutRef)
```

Measured with `bodyBottom = innerHeight = 768`, `visualViewport.height = 500`, then a pan, focusing
a low field the way Next does:

| | inset after focus | inset after pan | head start |
|---|---|---|---|
| this branch | 316px | **0px** | `focusLanded: true` — the old, too-late reveal |
| `main` | 316px | 0px | `focusLanded: true` |

After one pan this branch behaves identically to `main`. That matches the device symptom exactly:
Next works down the form, something pans once near the bottom, and from there every Next pans and
it never recovers.

### Section 2 is undone one line below its own repair

`isVisualViewportRecovered()` is right. The next statement in `applyKeyboardGeometry` (L305) is
`overlap = max(0, bodyBottom - viewport.bottom)` -> `clearKeyboardLayout()`. Same contaminated
quantity, same outcome. Probe on the blur path with a pinned sheet: `316px -> 0px`, the exact
result that fix exists to prevent.

### Why the new tests pass anyway

They use `rect({top: 100, bottom: 800})` against jsdom's `innerHeight` of 768. That 32px cushion is
the only thing keeping `overlap` positive under a pan, and a fully expanded Tall sheet cannot have
it. The "collapsed detent extends the body below the layout viewport" case in the comment is the
case this hook opts out of.

Test 2's expectation is the bug in miniature: **348px -> 80px during a pure pan** means 268px of
scroll range disappears while the keyboard has not changed size. Pan-invariant, it should stay
348px.

### Verified repair direction

Using `visualViewport.height` — the pan-invariant quantity section 2 already identifies — as the
obstruction measure in `applyKeyboardGeometry` (overlap + inset), `revealFocusedControl`
(overlap + inset), and `scrollControlIntoSafeArea` (safe area):

- inset holds 316px through the pan
- head start fires with `focusLanded: false` and the full 248px
- 143/144 of the suite passes; the one failure is test 2's `80px` expectation
- behavior is identical when `offsetTop === 0`, so nothing off the pan path moves

Whether that is the whole device story is still open — a clamped `scrollBy` near the end of eleven
fields, or WebKit wanting more than the 48px clearance, would each also pan. But the latch is what
makes any of them permanent.

### Smaller findings from the same pass

- **A plain tap between two fields now snaps where it used to glide.** `rememberFocusScroll` also
  runs on the `focusout` that `preventTouchFocusScroll` triggers itself (L396), so the tap path
  takes the instant head start it does not need — `preventScroll` already suppressed the native
  reveal there. Measured `{top: 248, behavior: 'smooth'}` after focus on `main` vs
  `{top: 248, behavior: 'auto'}` before focus here. `pendingPointerFocusScroll?.target !== control`
  distinguishes the two.
- **The multi-touch fix covers `pointerdown` but not `pointerup`.** `preventTouchFocusScroll` nulls
  `pendingTouchFocus` before comparing pointer ids, so a resting thumb that *lifts* first still
  disarms the tap in progress — same failure, other end of the gesture.
- `behavior: 'auto'` resolves to the element's computed `scroll-behavior`; `'instant'` is what is
  meant when racing the reveal.

### Housekeeping

- CI's `test` job on `3764063` is **cancelled**, so this commit has no test signal.
- Branch is three Bottom Sheet commits behind `main` (#5155, #5159, #5161 — the last changes how a
  settled detent reads the viewport).
- Suites on a clean box: this head 10,587/10,587, its base 10,583/10,583 (exactly the 4 new tests).

Keeping this in draft until the pan-invariant reading is in and test 2's geometry is honest.

imdreamrunner:
## You were right about the latch — fixed, with one correction

Reproduced your finding before changing anything. Pinned sheet, keyboard up, then one pan: the sheet's extra scroll range goes **316px → 0px** and the defenses switch off, exactly as you described.

### Background, for anyone reading cold

When you tap a field near the bottom of a form, the browser wants to reveal it. In this sheet it does that by scrolling the *page* — and since the sheet is `position: fixed`, the whole sheet travels with it. That is the bug users see: the page lurches upward.

To know whether a keyboard is in the way, the hook needs to know where the keyboard's top edge is. It was computing that as `visualViewport.offsetTop + visualViewport.height`. That number is wrong while the page is shifted: the shift moves the window over the page, but the keyboard stays where it is, so the keyboard appears to shrink as the shift grows.

A fully expanded Tall sheet — the only shape this hook runs in — sits flush against the bottom of the window. So when the shift reaches its maximum, that formula says the keyboard has zero height. The sheet concludes the keyboard is gone, throws away the scroll room it added, and switches off the protection that would have caught the next reveal. **One shift and it never recovers** — which is your device symptom exactly: Next works down the form until something shifts once, and from then on every Next shifts.

### What changed

`467400f4`, all four places that asked "where does the keyboard start":

- New helper `getUnobstructedBounds()` returns `{top: 0, bottom: visualViewport.height}` — the height alone, no `offsetTop`. Used by `applyKeyboardGeometry`, `revealFocusedControl`, `scrollControlIntoSafeArea`, and the recovery check.
- Both keyboard-geometry tests now place the sheet flush against the bottom of the window, as the real sheet sits. The old `348px → 80px` expectation was the bug written down as if it were correct.
- New test for the latch: after a shift, the scroll room holds **and** the next focus still gets its early scroll.

Your three smaller findings, same commit:

- **The tap path was taking an early scroll it doesn't need.** A tap already focuses in a way that suppresses the browser's reveal, so there is nothing to get ahead of — the early scroll only turned a smooth glide between fields into a snap. Now excluded; a test pins the glide.
- **The multi-touch guard only covered finger-down, not finger-up.** A resting thumb lifting first still cancelled the tap in progress. Now both ends.
- **`behavior: 'auto'` → `'instant'`.** `'auto'` defers to the element's own `scroll-behavior`, so a consumer setting `smooth` would animate a scroll that has to land immediately.

Each of the four, removed again, fails exactly one test.

### The correction: your fix is necessary but not sufficient

Measured on an iPhone 17 simulator through a real Next-driven shift:

| page shift | `visualViewport.height` | `window.innerHeight` | verdict |
|---:|---:|---:|---|
| 0 | 377 | 714 | keyboard |
| 48 | 377 | 666 | keyboard |
| 124 | 377 | 590 | keyboard |
| 230 | 377 | 484 | keyboard |
| 337 | 377 | 377 | **"recovered"** |

`window.innerHeight` shrinks as the page shifts. At maximum shift it equals `visualViewport.height`, so the shift-proof comparison **also** reports the keyboard as gone. It holds through partial shifts, and on any platform where `innerHeight` stays put, but iOS can reach the point where every measurement collapses together.

What keeps the sheet out of that state is the other half of this PR: the page shift is undone on the first scroll event, so it never gets past its first step (~120px in the trace above) and the collapse is never reached. Neither fix covers this alone — they need each other.

### Housekeeping

- Rebased onto `main`. #5159 makes the same argument one hook over — "read the snap fractions from the layout viewport rather than the visual viewport" — so the two are now consistent.
- The cancelled `test` job is superseded; CI is green on the current head, `test` included.
- `tsc` caught a dead comparison the new finger-down guard created (`e37d33e8`). Only `build` typechecks that file — lint and the unit tests both pass it.
- Two failures persist on my box that `main` fails too (CLI name resolution, hook discovery — both artifacts of a case-insensitive filesystem). Everything else passes on re-run.

### Still open

Your deep-form question. My simulator automation stopped landing on the Next chevron after three fields, so the end of the eleven-field form is untested — a clamped scroll, or WebKit wanting more than the 48px clearance, would each still shift the page there. The sheet held steady for everything I did reach, including across an app switch. Keeping this in draft.

imdreamrunner:
Reviewed at **d6414bf** — the branch moved from 4190199 while I was reading, so this is the "scope the keyboard hook to what it can prove" revision, rebased on #5172. `npx vitest run packages/core/src/BottomSheet` → 160 passed.

Claiming the transition on capture-phase `blur` is the right mechanism, and reading the obstruction as `visualViewport.height` rather than `offsetTop + height` is a genuinely good catch — the latch it removes is the thing that made this read as flaky. Everything below is about the edges of the claim, not the claim itself. Each item was reproduced before writing it down.

## 1. Autofocus into a sheet field is never claimed

`<TextInput hasAutoFocus />` emits **both** React's `autoFocus` and `data-autofocus` (TextInput.tsx:445-446; same for TextArea and NumberInput). React's `autoFocus` is applied at `commitMount`, which runs before any effect — so the hook's `blur` listener does not exist yet, and `focusPanel` then early-returns because focus is already inside the panel. The only focus call is unprevented:

```
P11 focus calls: [ [ [], "at commitMount (react-dom-client.development.js:22158)" ] ]
```

That is with a trigger focused beforehand, i.e. the case where the claim would otherwise work.

The `data-autofocus` path has its own hole. `focusPanel` (BottomSheet.tsx:179/181) calls `.focus()` with no `preventScroll`, and a focus that starts from `document.body` raises **no blur at all**, so there is nothing to claim:

```
P7  modal sheet, [data-autofocus] input, nothing focused first → [[]]                    (unclaimed)
P8  identical, but a trigger focused first                     → [[], [{preventScroll:true}]]  (claimed)
```

P8 is the nested case that works: `focusPanel`'s unprevented call raises the blur, the handler claims it, the outer call finds the element already focused. P7 is the same code with no prior focus — programmatic open, deep link, a sheet rendered with `isOpen` on mount.

`preventScroll` on both `focusPanel` calls fixes the second half. The `commitMount` half cannot be claimed by a listener installed in an effect at all — the sheet would have to own the autofocus itself, or the case has to be documented.

## 2. A non-modal sheet has no protection on the first focus

`focusPanel` only focuses the panel `if (isModal)`, so with `hasScrim={false}` focus stays on the body, and the first focus into a field raises no blur:

```
P9  hasScrim={false} → activeElement is body on open; input.focus() → [[]]   (unclaimed)
```

`isPageScrollLocked` is also false there, correctly, so the document-scroll net is off too — this shape has neither defence for that first focus. The panel is already `tabIndex={-1}` (BottomSheetPanel.tsx:500), so parking focus on it for non-modal presentations too would close it.

## 3. Prevention switches off mid-session at a shorter detent

`claimFocusTransition` returns early on `!isFullyExpandedRef.current` (useMobileKeyboard.ts:264), and `isFullyExpanded` is `settledOffset === 0` (BottomSheetPanel.tsx:378). Same sheet, same session, one drag apart:

```
P10 fully expanded   → [[], [{preventScroll:true}]]   (claimed)
P10 dragged to p50   → [[]]                           (unclaimed)
```

The `height` doc line now says which *heights* opt out; it does not say that a Tall sheet loses the protection when the user drags it down and gets it back when they drag it up. If the gating stays, that sentence needs the detent case in it. Splitting the gate — `isEnabled` for the listener, `isFullyExpanded` for the inset — is the alternative; the claim touches no geometry, so it does not need the expansion invariant the inset does.

## 4. Two tests no longer assert anything

No pointer listeners remain in the hook, so `fireTouchPointer` drives nothing:

- BottomSheet.test.tsx:1322-1323 — "does not accommodate the keyboard at a shorter Tall detent" fires pointerdown/up and asserts `expect(focus).not.toHaveBeenCalled()`. That now passes at *every* detent and every height, so it no longer distinguishes the thing it is named for. A blur-driven transition (focus field A, then focus field B) is what tests it now — that is how P10 above measures it.
- BottomSheet.test.tsx:1373-1374 — same pattern in the presentation `it.each`.

## 5. Caret placement has no coverage, and needs a device check

The previous revision focused at `pointerup` for one stated reason, in a comment this PR deletes: *"focus before the native click so preventScroll suppresses page panning while the click still owns caret placement."* The two `fireEvent.click` assertions that guarded it went with it. The measurements in the description count `focus`/`blur` events, which is the right check for double-firing but says nothing about where the caret lands.

Delivering focus before the click can cost the tap's caret position. It may well survive here — the caret comes from the click's own hit test, which is a separate step from the focus dispatch being claimed — but it is currently an untested assumption where it used to be a deliberate design constraint.

**Repro:** tap into the *middle* of existing text in a populated field. If the caret jumps to the end or back to its previous offset, the known mitigation is to normalize it on a field's first focus with `setSelectionRange(value.length, value.length)` — no such call exists in the package today.

## 6. Re-tap of a still-focused field: quantified

Listed as out of scope, agreed — this just pins how far out of scope it is. After a dismissal that does not blur (iPad's hide-keyboard chevron, attaching a hardware keyboard), the re-tap produces **zero** events of any kind:

```
P5 blur events on re-tap: 0 | still active: true
```

So no `blur`-based mechanism can ever reach it; whatever covers it has to be a pointer or touch path. Worth saying in the out-of-scope list, since the rest of that list is coverable within the current design.

## Checked and fine

- **Chained transitions survive the removed re-entrancy guard.** A consumer `onFocus` on B that focuses C: focus ends on C, no recursion (P6). Removing the guard was right.
- **Multiple presented sheets** each add their own document listener, but each only claims destinations inside its own `body`, and a second claimant finds `destination === document.activeElement` and does nothing. Benign.
- The global `overscroll-behavior` stylesheet is gone, and `isPageScrollLocked` correctly stops the net from reverting the user's own scroll behind a non-modal sheet. Both were on my list before this revision.

Items 1 and 3 are the ones I would not ship without: 1 is a mainstream path (`hasAutoFocus` on a sheet's first field), and 3 changes behavior under the user's own drag with nothing in the docs about it.

imdreamrunner:
Reviewed all six items by reproducing each one first. Four led to changes, two did not. Here is what I changed, and — more importantly — the two things you asked for that I deliberately did **not** do, with the reasoning, so you can overrule me.

## Fixed

**1. Autofocus (the half that can be fixed).** You were right that opening a sheet which autofocuses a field left the browser free to scroll the page. `focusPanel` now asks for focus in the way that refuses that scroll, on both of its paths. A test fails without it.

**4. Two tests that asserted nothing.** Correct, and my mistake. They pressed and released a finger and checked that nothing got focused — but the hook stopped listening to finger events when this rewrite landed, so they passed no matter what. They now do what their names say, and the shorter-detent one fails if its gate is removed. That was the last use of the touch helper, so the helper is gone too.

**3. (documentation half).** The `height` docs did already say shorter heights leave the browser's reveal in place, but not the part that matters here: a Tall sheet **loses** the protection the moment the user drags it down, and gets it back when they drag it up. They say that now, in those words.

**6.** Your measurement — a re-tap after the keyboard is dismissed without a blur produces no events at all — is now in the out-of-scope list, along with what it implies: nothing built on blur can ever reach that case.

## Measured instead of assumed

**5. Caret placement.** Fair challenge, so I measured it rather than argue. In Mobile Safari, tapping into the middle of a field that already has text, with another field focused first:

| | where the caret landed |
|---|---|
| with this change | character 40 of 40 |
| without it | character 40 of 40 |

Identical. Taking the focus early does not cost the tap its caret, because the caret comes from the tap's own hit test, which is a separate step. And a tap into a field that is *already* focused cannot be affected at all — the code does nothing when the destination is already the active element.

## Overridden — please push back if you disagree

**Item 1, the other half: React's `autoFocus`.** You are right that it is unreachable. React applies it before any effect runs, so no listener this hook installs exists yet. I did not try to work around it. Closing it would mean the sheet taking over autofocus for its own content, which is a design change well outside a keyboard fix. It is documented as a known gap instead.

**Item 2: the first focus into a sheet with no scrim.** Your diagnosis is right — focus starts on the page body, so there is no transition to take over. I did not apply your fix of parking focus on the panel, because that contradicts behaviour the component tests on purpose: *"The background stays interactive, so the sheet must not grab focus."* Changing it would be a behaviour change to non-modal sheets, decided inside a keyboard PR.

There is also a mitigating detail worth adding: a sheet without a scrim does not lock the page, so the page is *meant* to scroll there. A scroll in that state is ordinary page scrolling, not a pinned sheet being dragged off screen. The visible damage is much smaller than in the modal case. Documented, not fixed.

**Item 3: the gate itself.** You offered "split the gate, or document it." I documented. Splitting it would extend the protection to shorter detents — which may well be the better end state, but the current scope was set deliberately: keyboard support is for a fully expanded Tall sheet, and everything else opts out. Widening it inside this PR would quietly change that decision.

You are right that the take-over touches no geometry and therefore does not need the full-expansion rule the scroll-space part needs. If you want it split, it is a small change and I will make it.

## State

161 tests in `packages/core/src/BottomSheet`. Every guard added by this PR fails a test when removed, including the take-over itself. Lint, build and CI all green.

Real story on a simulator, sheet top edge measured from the screenshot; 376px is "has not moved": 376 focused, 376 after six presses of Next, **376 with the last of the eleven fields focused** — the deep-form case you flagged — 376 after four presses of Previous from there, and 376 across an app switch.

Physical-device testing is being handled separately. Still in draft.

PR #5162, opened 2026-08-17
fix(Popover): make theming popover.borderRadius actually change the radius

cixzhang:
## Why

`popover.borderRadius` in a theme does nothing. `--_popover-radius` is documented in `Popover.doc.mjs` and registered in `derivedVarRegistry`, but nothing in the package reads it — `usePopover` hardcoded the surface radius to `--radius-container`. The doc's stated default (`--radius-element`) was wrong too.

Reading the var is only half of it. The `astryx-popover` target sat on the content `<div>` *inside* the popup surface, so the theme rule landed on a box that paints nothing: the radius (and any background or border a theme set there) never reached the rounded, shadowed surface above it. HoverCard, DropdownMenu and Dialog all put the target on the element that paints and read their private radius var from it.

## What

- `usePopover`'s surface declares `--_popover-radius` (default `--radius-container`) and reads it, like its siblings.
- `Popover` names that surface as its `popover` theme target through the existing `surfaceTarget` option — the mechanism added for exactly this — so the target sits on the box that paints.
- Content padding moves onto the surface with the target, so a themed `padding` still *replaces* the default instead of nesting inside it, and a consumer's `xstyle` still overrides it.
- Doc default corrected to `var(--radius-container)`.

## Risk

`astryx-popover` now resolves to the popup surface rather than the content box inside it. Themes gain correct behaviour (radius, background and border now paint on the rounded box); anything that reached for `.astryx-popover` expecting the inner, unpainted div sees a different element. No theme in this repo sets the `popover` key. The shared `astryx-popover-surface` class and every other `usePopover` consumer are untouched.

## Testing

- 136 focused tests pass: Probe Theme generator, Popover, and derived-var registry.
- Production Core/theme build and Storybook build pass.
- The Probe Theme now sets `popover.borderRadius` to 32px and keeps an open Popover in its component-target sheet.
- Browser A/B with the same probe fixture: current main leaves the painted surface at 12px while the inner target reads 32px; this branch makes the target and painted surface the same element at 32px. The matching frames differ by 13,137 pixels, confined to the Popover surface.
- Browser regression check: a consumer class setting `padding: 0` now lands on that same surface and resolves all four edges to 0px; the unit test pins `className` and inline `style` there too.

### DIFF
--- a/.changeset/popover-radius-reaches-surface.md
+++ b/.changeset/popover-radius-reaches-surface.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Popover: theming `popover.borderRadius` now changes the rendered radius. The `astryx-popover` target moves onto the popup surface — the box that paints background, radius and elevation — and `usePopover` reads the registered `--_popover-radius` there instead of hardcoding `--radius-container`. Content padding moves with it, so a themed `padding` still replaces the default instead of nesting inside it.
+
+@cixzhang
--- a/.github/scripts/visual-gate/lib/probe-theme.mjs
+++ b/.github/scripts/visual-gate/lib/probe-theme.mjs
@@ -71,6 +71,15 @@ export function unionValues(type, aliases = {}) {
 }
 
 /**
+ * Non-colour properties whose ownership must be visible in the probe capture.
+ * The colours prove that a target is reachable; these values prove that the
+ * target sits on the element that paints the documented property.
+ */
+const PROPERTY_PROBES = {
+  popover: {borderRadius: '32px'},
+};
+
+/**
  * Build the probe theme's `components` map.
  *
  * @param {Array<{key: string, component: string, props: string[], states: string[]}>} targets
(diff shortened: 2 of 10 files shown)


cixzhang:
Direction: keep the theme target on the painted Popover surface. The element that paints the radius should own the target and padding so `borderRadius` visibly does what the contract says.

cixzhang (commented):
Thanks, moving the target onto the painted surface is the right fix. Before undrafting, `className` and `style` need to move with `xstyle`: they still land on the inner child, so neither can override the default padding now owned by the surface (`Popover.tsx:519`). Could we keep all three styling escape hatches on the same surface?

[Reviewed by Robohands]

cixzhang (commented):
Thanks, this addresses the last review: `className` and `style` now move with `xstyle` to the painted surface. This is an approve once you undraft.

[Reviewed by Robohands]

cixzhang:
/accept-visual 33041576630/2 All 124 added frames are Popover-only coverage across existing stories and themes; the reviewed intended change makes popover.borderRadius paint 32px, while default and zero-padding controls remain pixel-identical.

cixzhang:
/accept-visual 33041576630/2 All 124 added frames are Popover-only coverage across existing stories and themes; the reviewed intended change makes popover.borderRadius paint 32px, while default and zero-padding controls remain pixel-identical.

cixzhang:
/accept-visual 33041576630/2 All 124 added frames are Popover-only coverage across existing stories and themes; the reviewed intended change makes popover.borderRadius paint 32px, while default and zero-padding controls remain pixel-identical.

cixzhang:
/accept-visual 33041576630/2 All 124 added frames are Popover-only coverage across existing stories and themes; the reviewed intended change makes popover.borderRadius paint 32px, while default and zero-padding controls remain pixel-identical.

PR #5163, opened 2026-08-17
fix(Switch): keep accent for the on state and mute the off track

cixzhang:
## Why

Only the **on** state should read as a filled, accented control. Today the off track paints `--color-background-gray` — a categorical gray with nothing tying it to the switch — and in two themes it comes out *brighter and more filled-looking than the on state*, because both keep a light gray in their dark slot:

| theme · mode | off track | on track | off : on |
| --- | --- | --- | --- |
| butter · dark | `#f0edd4` | `#fdee8c` | **1.00:1** |
| y2k · dark | `#ede0d4` | `#edeffc` | **1.13:1** |

The same split hits the thumb, which paints `--color-background-surface`: two independently chosen neutrals, so they converge. Sampled from rendered Chrome pixels, the thumb clears the 3:1 WCAG 1.4.11 asks of its track in **2 of 16** theme/mode combinations — including 1.55:1 with no theme at all.

## What

Both off colours are mixed from the surface the control sits on toward the theme's own text colour:

```
track  color-mix(in srgb, var(--color-background-surface), var(--color-text-primary) 12%)
thumb  color-mix(in srgb, var(--color-background-surface), var(--color-text-primary) 68%)
```

12% keeps the track muted — a channel, not a fill — and the thumb carries the contrast the muted track can't. Derived from one pair of tokens, track and thumb cannot converge in any palette, and the result is what the forced-colors path already drew: a plain Canvas track with a CanvasText thumb.

Neutral's and Stone's switch overrides of `--color-background-gray` were reaching for the same thing and cleared neither bar; they're removed, along with their copies in the CLI theme templates.

## Risk

Visual change to every off switch: the knob goes from surface-coloured to a mid-tone. The on state is untouched. The busy spinner in the off thumb switches to `shade="inherit"` against a surface-coloured `color`, since the thumb's fill now flips polarity between light and dark.

## Testing

788 core + theme tests pass. Measured across 8 themes × 2 modes from screenshot pixels (never `getComputedStyle` — the alpha and `color-mix` values only mean something composited):

| | before | after |
| --- | --- | --- |
| worst off : on | 1.00:1 (butter dark) | 3.87:1 |
| worst thumb : off track | 1.15:1 (chocolate dark) | 3.25:1 (matcha light) |
| combos with thumb ≥ 3:1 | 2 / 16 | 16 / 16 |

### DIFF
--- a/.changeset/switch-off-state-muted.md
+++ b/.changeset/switch-off-state-muted.md
@@ -1,0 +1,10 @@
+---
+'@astryxdesign/core': patch
+'@astryxdesign/theme-neutral': patch
+'@astryxdesign/theme-stone': patch
+'@astryxdesign/cli': patch
+---
+
+[fix] Switch: only the on state paints an accent fill. The off track is now a muted step off the surface, with the thumb carrying the state contrast, so "off" can never read as filled — in Butter and Y2K dark it was measuring 1.0:1 and 1.1:1 against the on state. Both off colours derive from the surface and the theme's own text colour, so the thumb clears 3:1 against its track (WCAG 1.4.11) in every theme instead of none. Neutral and Stone no longer need their `--color-background-gray` switch overrides.
+
+@cixzhang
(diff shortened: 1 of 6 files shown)


cixzhang:
Closing note for the record: wrong layer. The switches being reported are agentcloud's, and its off track comes from that app's own theme override (`SWITCH_COMPONENTS`, 70% toward the palette's ink), not from core — which is why off read as bright as the accent. Fixed there instead, in D116354680.

PR #5164, opened 2026-08-17
docs(layout): rewrite as an outside-in build sequence with tiered guidance

ernestt:
## What

Rewrites the **Layout** guide (`astryx docs layout`, [docs page](https://astryx.atmeta.com/docs/layout)) into an
outside-in build sequence with a repeating shape, so an agent reading it top to
bottom lands on the same layout decisions a designer would.

Compared against what is live today, the guide goes from five flat sections with
no internal structure to four build phases with fourteen named sub-sections, and
from three code examples to ten.

## Why

The live guide is thin in the places where agents actually guess. It never names
the mechanism for responsive behavior, never gives a region a width, and never
states the alignment rule, so an agent has to invent all three. The A/B test
below shows exactly that happening: given the live guide, the agent hand-rolled
its own `matchMedia` subscription because it could not find `useMediaQuery`, and
its table text landed 8px off the page heading.

## Structure: live vs this PR

| | Live (`main`) | This PR |
|---|---|---|
| Sections | 5 flat | 5 (Overview + 4 phases) |
| Sub-sections | 0 | 14 |
| Code examples | 3 (21 lines) | 10 (100 lines) |
| Do / Don't items | 4 / 4 | 16 / 21 |
| Tables | 1 | 0 (converted to lists) |
| Words | 548 | 1,691 |

The four phases, outside-in, named to match
[Material 3's layout vocabulary](https://m3.material.io/foundations/layout/layout-overview/overview):

- **Scaffold**: Shell, Navigation, Best practices
- **Structure**: Type hierarchy, Containers, Headers and footers, Side panels, Best practices
- **Spacing**: Alignment, Rhythm, Density and size, Best practices
- **Breakpoints**: Responsive contract, Best practices

`Breakpoints` sits last so nothing references a phase the reader has not reached
yet. Every sub-section title is a noun, so the on-this-page outline reads as a
table of contents rather than a list of commands.

## Three tiers, so a rule is distinguishable from a suggestion

The single biggest readability change. Every section now separates:

1. **Principles** in prose. Always true, no numbers.
2. **Recommended values** in code blocks and their comments. Tunable. Region
   budgets, capped widths, gap steps, and density pairings all moved here out of
   prose.
3. **Rules** in the Do / Don't tables, and nowhere else.

Previously a sentence like "side nav 240-280" read with the same authority as
"one primary action per region", and only one of those is a rule.

Each phase closes with a section that is *only* do/dont lists, so the docsite
renders it as the same green/red table used on component docs
(`isBestPracticesSection` -> `BestPracticesBlock`) and the CLI renders `+`/`x`
bullets. No schema change. Renamed from "Do & Don't" to **Best practices** to
match the seven other reference docs that already use that name.

## New guidance not on the live page

- **Alignment**: hold one content line per region, not one padding value, via
  `container_inset = content_line - component_intrinsic_inset`. Names which
  components carry their own inset.
- **Headers and footers**: `LayoutHeader` / `LayoutFooter` as `Layout` slots,
  when `Toolbar` replaces a header, and how one `padding` on `Layout` reaches all
  three slots.
- **Side panels**: `ResizeHandle` placement on the panel's inner edge, including
  `isReversed` for `end`-slot panels, and which element owns the divider.
- **Type hierarchy**: body copy takes no props. Two accessible text colors only
  (`primary`, `secondary`); `disabled` fails contrast and is reserved for
  disabled controls.
- **Rhythm**: grouping comes from contrast between tight and generous gaps, not
  one repeated value.
- **Density and size**: match `density` to use frequency, one control `size` per
  row so heights share a baseline.
- **Navigation**: app type and destination count are guiding indicators, not
  deterministic thresholds.

Also documents `contentWidth`, `padding`, and `defaultHasDividers` on
`Layout.doc.mjs`. All three already existed on the component but were missing
from its doc, so `astryx component Layout` never surfaced them. The guide points
readers at `contentWidth` to cap prose and form regions, which only works if the
prop is discoverable.

## Proof: A/B vibe test

Same prompt, same tooling, same model, two arms. The only variable is the layout
doc text.

**Setup**, following the five fairness invariants in
`internal/vibe-tests/README.md`:

- Prompt: `wd-4` ("TodoTracker") verbatim from `test-sets/default.json`, the
  existing todo-list prompt in the battery. `expectedComponents` withheld from
  both arms per invariant 3.
- Both arms: fresh context-free subagents, identical instructions, identical
  access to `astryx component <Name> --dense`, identical house rules.
- Arm A reads the **live** doc (`origin/main`); Arm B reads **this PR's** doc.
  Each arm received the doc exactly as `astryx docs layout` renders it, 737 vs
  2,345 words of rendered text including code blocks. Both arms were instructed
  not to run `astryx docs layout` themselves, so the working tree could not leak
  the new doc into Arm A.
- Rendered and screenshotted through the repo's own pipeline
  (`build-previews.ts` then `screenshot-previews.ts`), Chromium at 1280x800 and
  375x812.

### Result: both arms are good, and they diverge on the parts the live doc omits

Both produced a real app with zero `tsc` errors, zero `<div>`, zero `xstyle`,
zero raw StyleX, and no `Card` misuse. The live guide's frame-first advice is
already doing its job on card soup. The differences are everywhere it stays
silent:

| | Arm A (live doc) | Arm B (this PR) |
|---|---|---|
| **Content line spread** | **8px** | **0px** |
| Responsive mechanism | hand-rolled `matchMedia` + `useSyncExternalStore` | `useMediaQuery` |
| Nav collapse | none | `mobileNav={{breakpoint: 'md'}}` |
| `contentWidth` / `defaultHasDividers` | not used | used |
| Pending marker | ad hoc spinner | `Token` |
| `tsc` errors | 0 | 0 |

**Content line** is measured, not eyeballed. Left edge of the `h1`, the first
column header, the first body cell, and the footer text, in the rendered page:

- Arm A: 260, 268, 268, 260 -> **8px spread**. The table's intrinsic cell inset
  was never compensated, which is precisely the failure the new Alignment
  section exists to prevent.
- Arm B: 276, 276, 276, 276 -> **0px spread**.

Arm B's agent reported measuring this itself while building, because the doc gave
it a test to run.

**Desktop, 1280x800.** Arm A left, Arm B right:

<table>
<tr>
<td width="50%">Arm A, live doc, desktop</td>
<td width="50%">Arm B, this PR, desktop</td>
</tr>
</table>

**Mobile, 375x812.** This is where the missing breakpoint guidance shows. Arm A
kept all five columns and truncated every title to "Add keyboard...". Arm B
dropped the Created column so titles stay readable:

<table>
<tr>
<td width="50%">Arm A, live doc, mobile</td>
<td width="50%">Arm B, this PR, mobile</td>
</tr>
</table>

### Limits of this test

Worth stating plainly: n=1 per arm, one prompt, one model. This is an existence
proof that the added guidance changes specific decisions, not a statistical
result. Both arms cleared the bar the live doc already sets. Arm B also surfaced
two genuine problems the doc should own, filed as follow-ups rather than fixed
here: a `LayoutContent` top-padding conflict under a divider-less `LayoutHeader`,
and a collision between the recommended 32-40px row height and `balanced`
density with `sm` controls.

## Notes

- Dense variant (`layout.doc.dense.mjs`) stays index-aligned with the full doc,
  verified section keys, block counts, and per-index types, so
  `astryx docs layout --dense` matches. Renaming a section title required
  updating the overlay's `section` matcher key.
- Docsite: 377 tests pass. CLI suite has 9 pre-existing failures in component
  search, theme manifest, and hook discovery; none read this doc.
- Doc-only change, no changeset, consistent with prior doc-only commits to
  `packages/core/src/**/*.doc.mjs` such as `docs(core): document the seven hooks
  missing from the hooks doc`.
- No em dashes in prose per the deslop rubric; en dashes only for numeric ranges.
- Also fixes a pre-existing docsite bug where consecutive do/dont lists in a
  nested section rendered without their green/red badges. It affected
  `theme.doc.mjs` too.

### DIFF
--- a/apps/docsite/src/components/docs/ReferenceDocView.tsx
+++ b/apps/docsite/src/components/docs/ReferenceDocView.tsx
@@ -84,10 +84,14 @@ function buildOutline(sections: DocSection[]): {
   return {sectionIds, blockIds, outline};
 }
 
-function isBestPracticesSection(section: DocSection): boolean {
-  return section.content.every(
-    b => b.type === 'list' && (b.style === 'do' || b.style === 'dont'),
+function isGuidanceList(block: DocSection['content'][number] | undefined) {
+  return (
+    block?.type === 'list' && (block.style === 'do' || block.style === 'dont')
   );
+}
+
+function isBestPracticesSection(section: DocSection): boolean {
+  return section.content.every(isGuidanceList);
 }
 
 function BestPracticesSection({
@@ -161,6 +165,28 @@ export function ReferenceDocView({
                         </AnchorHeading>
                       );
                     }
+                  }
+                  // A run of do/dont lists is one badge table, so guidance
+                  // keeps its Do/Don't label outside all-guidance sections.
+                  if (isGuidanceList(block)) {
+                    if (isGuidanceList(section.content[blockIndex - 1])) {
+                      return null;
+                    }
+                    const items: {guidance: boolean; description: string}[] =
+                      [];
+                    for (let j = blockIndex; j < section.content.length; j++) {
+                      const run = section.content[j];
+                      if (!isGuidanceList(run)) {break;}
+                      for (const item of run.items ?? []) {
+                        items.push({
+                          guidance: run.style === 'do',
+                          description: item,
+                        });
+                      }
+                    }
+                    return (
+                      <BestPracticesBlock key={blockIndex} items={items} />
+                    );
                   }
                   return (
                     <ContentBlockRenderer key={blockIndex} block={block} />
--- a/packages/cli/assets/docs/layout.doc.dense.mjs
+++ b/packages/cli/assets/docs/layout.doc.dense.mjs
@@ -4,124 +4,267 @@
 
 export const docsDense = {
   description:
-    'frame-first app layout: shell choice, region budgets, nav choice, sections vs rows vs cards',
+    'outside-in app layout in 4 phases: frame -> structure -> space -> ship. each phase ends in a do/dont table.',
   sections: [
+    // ===== FRAME =====
     {
-      section: 'Frame First',
-      title: 'Frame First',
+      section: 'Frame: Pick the shell',
+      title: 'Frame: Pick the shell',
       content: [
         {
           type: 'prose',
-          text: 'decide frame before content. content-first (Card-wrapped sections in a scroll column) = prototype look.',
+          text: 'phase 1/4 frame. start from the frame, not content. content-first (Card-wrapped sections in a scroll column) = prototype look.',
         },
         {
           type: 'list',
           items: [
             'pick frame: AppShell (nav apps) | Layout+LayoutPanel+LayoutContent (multi-pane tools) | plain column (docs/forms)',
-            'budget regions in px first: side nav 240-280, rail 64-72, inspector 340-420, facet rail 220-260',
-            'interior spacing = spacing token, never raw px. px is only for structural widths (nav 256, inspector 380)',
-            'container policy per region: dense data = rows; dashboards/galleries = card grids',
-            'write responsive contract up front',
+            'budget regions px first: side nav 240-280, rail 64-72, inspector 340-420, facet rail 220-260',
+            'raw px = structural widths only; interior spacing = tokens',
+            'set container policy (rows vs card grid) + responsive contract before content',
           ],
         },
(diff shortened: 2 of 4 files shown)


cixzhang (approved):
So good <3

PR #5165, opened 2026-08-17
feat(theme): let an app declare its own semantic token via CustomTokens

cixzhang:
### Why

An app needed one token Astryx does not ship — `--color-layer-border`, the edge of a floating layer as distinct from a hairline inside one. The runtime took it happily: `astryx theme build` emitted `--color-layer-border: light-dark(...)` into the theme CSS and the component block resolved against it. Only `tsc` refused, because `defineTheme`'s `tokens` is `Partial<Record<TokenName, TokenValue>>` and `TokenName` is a closed union built from Astryx's own defaults maps.

That is the worst combination: the capability exists, the theme builds, and it looks supported right up until typecheck. The variant axes already have an escape hatch (`FooVariantMap`, `CustomTextTypes`); tokens had none, so the app's only way out was to drop the token entirely and reference a raw CSS var — which breaks the moment anything forces a colour mode via `data-astryx-media`.

### What

`CustomTokens`, an empty interface on `@astryxdesign/core/theme` that an app augments:

```ts
declare module '@astryxdesign/core/theme' {
  interface CustomTokens {
    '--color-layer-border': true;
  }
}
```

`defineTheme`'s `tokens` is now keyed by `ThemeTokenName = TokenName | CustomTokenName`. `TokenName` is deliberately unchanged and still exhaustive over the tokens Astryx ships — `tokenVars` is keyed by it and has no entry for a name only the app knows about, so widening it would have made that map lie. Keys are constrained to `` `--${string}` ``, since they are emitted verbatim as CSS declarations.

### Why this and not `string`

The closed union is the single most valuable thing about the token API: autocomplete over ~100 semantic names and a compile error on every typo. Opening it to `string` fixes one app's token by removing that for everyone. Module augmentation is opt-in and per-name — the union stays closed for every name nobody declared.

It is also the house pattern rather than a new mechanism: this is the same seam as the `*VariantMap` axes and `CustomTextTypes`, which [#5029](https://github.com/facebook/astryx/pull/5029) is currently teaching the CLI to generate. One difference worth stating: `astryx theme build` cannot generate this one. A variant augmentation is for the component's *consumers*, so the build can emit it afterwards; a token augmentation is needed by the theme source the build reads, so it has to be hand-written. That is noted on the interface's docs.

### Risk

Low, and additive. With `CustomTokens` empty, `CustomTokenName` is `never` and `ThemeTokenName` is exactly `TokenName` — no existing theme changes type. Nothing at runtime changed; the runtime already accepted these keys, which is what made this a gap rather than a feature request. The one new failure mode is a declared key that is not a `--` custom property: it does not widen the union, so the token keeps failing typecheck rather than emitting invalid CSS. There is a test for it.

### Testing

`packages/core/src/theme/customTokens.test.ts` (4 tests) plus the whole theme suite, 742 passing. The two `@ts-expect-error` assertions in it are load-bearing — `pnpm typecheck` fails if the union stops being closed.

Verified from a real consumer resolving `@astryxdesign/core/theme` through the built package, not the source:

```
1. tsc — theme.ts, token declared via CustomTokens
exit=0

2. tsc — typo.ts, same token misspelled, no augmentation
typo.ts(6,5): error TS2353: Object literal may only specify known properties,
and ''--color-layer-bordr'' does not exist in type
'Partial<Record<ThemeTokenName, TokenValue>>'.
exit=2

3. astryx theme build theme.ts
✓ scratch.css — 1 token overrides, 1 component overrides
--color-layer-border: light-dark(hsl(0 0% 88%), hsl(0 0% 26%))
border-color: var(--color-layer-border)

4. astryx theme build theme.ts --check
✓ Theme outputs are up to date with theme.ts.
```

Before the change, step 1 was `TS2353` on `'--color-layer-border'` while step 3 emitted the same CSS it does now — both halves of the gap reproduced at 0.4.3.


cixzhang:
Closing by API decision. `defineTheme` keeps its semantic token set closed; apps should not extend the public token contract through module augmentation or a naming prefix.

PR #5171, opened 2026-08-18
feat(lint): require provider-backed locale APIs

nynexman4464:
## Summary

- add `@astryx/no-raw-intl-locale` to prevent shipped packages from bypassing `InternationalizationProvider` with raw `Intl`, locale-sensitive prototype methods, aliases/destructuring/computed access, or `navigator.language(s)`
- enable enforcement only after the provider-hook and non-date migration rungs below it

This PR is lint-only. It contains no runtime migration, changeset, codemod, Calendar/date-input implementation, or CLDR weekday work.

## Raw Intl rule

Outside a short exact-file infrastructure list, shipped-package code may not:

- construct or call locale-sensitive `Intl` formatters
- alias, destructure, or dynamically index the global `Intl` object
- call `toLocaleString`, locale date/time/case methods, or `localeCompare`
- read global `navigator.language` or `navigator.languages`

`Intl.Locale` and direct grapheme-only `Intl.Segmenter` usage remain outside the display-locale policy. Named test-oracle files may construct independent formatters without widening the exception through ESLint configuration.
Approved implementation files still require a syntactically explicit locale; only the two named legacy date-helper bodies remain temporarily ambient pending #5120.

The rule is deliberately syntax-only for prototype methods. It documents the receiver-type false-positive risk and the computed-method-name limitation rather than hiding them behind a configurable allowlist.

## GitHub stack #5196

1. #5194 — provider-backed locale hooks
2. #5195 — non-date locale migrations
3. **#5171 — lint enforcement**

The stack is registered and managed with `gh stack`; all three PRs are ready for review. The contributor’s #5120 date migration is independent on `main`, and any future Core-only enforcement for its optional helper arguments will follow after that PR lands. #5190 is separate work.

## Verification

- full ESLint-plugin suite: **407/407**
- uncached `pnpm lint:strict`: pass (**0 errors**)
- repository structure, package-boundary, changeset, client-module, portable-script, and i18n-catalog checks: pass
- `git diff --check`: pass

Mutation proof:

- disabling the global-`Intl` and `navigator.language(s)` visitors makes alias, destructuring, computed-key, and standalone browser-locale rule cases fail

## Scope

The gate applies to Core, Charts, Richtext, and Vega source/tests. Lab adopts the shipped-package rule when a component graduates.

### DIFF
--- a/internal/eslint-plugin-astryx/no-raw-intl-locale.js
+++ b/internal/eslint-plugin-astryx/no-raw-intl-locale.js
@@ -1,0 +1,540 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+/**
+ * @file no-raw-intl-locale.js
+ * @description Forbid raw `Intl` access — direct formatter/method calls AND
+ * any other reference to the global `Intl` object (aliasing, destructuring,
+ * computed indexing) — outside the approved i18n infrastructure boundary.
+ * Separately, forbid `navigator.language`/`navigator.languages` as a locale
+ * source anywhere in the lint scope, in any position, not only as an
+ * argument to an `Intl`/locale-method call.
+ *
+ * `InternationalizationProvider` is the sole user-facing locale source.
+ * Shipped component code must read it through the public provider-aware
+ * locale utilities (`useLocale()`, `useCollator()` from
+ * `@astryxdesign/core/i18n`, or an existing formatting helper such as
+ * `plainDateFormat`/`formatInstant`/`formatFilterValue`) — never construct an
+ * `Intl` formatter or call a locale-sensitive prototype method directly.
+ * Passing an explicit locale expression does not satisfy this rule: a
+ * hardcoded literal, `navigator.language`, or an arbitrary variable are all
+ * still raw Intl access from the policy's point of view.
+ *
+ * Raw `Intl` construction is confined to a short, explicit list of files
+ * (`APPROVED_INFRA_FILES` below): the pure, SSR-safe implementation helpers
+ * that every provider-aware caller ultimately goes through, plus a few of
+ * their colocated tests that deliberately build an independent Intl oracle
+ * rather than reuse the helper under test. That list is the only exception
+ * mechanism: there is no config option to widen it from `eslint.config.js`,
+ * so opting a new file out means editing this rule (and its tests) in
+ * review, not a one-line config change.
+ *
+ * `navigator.language`/`navigator.languages` are rejected everywhere,
+ * including inside an approved infrastructure file — the provider is the
+ * only sanctioned entry point for a real locale, so infrastructure code must
+ * still receive one from a caller rather than reading the host default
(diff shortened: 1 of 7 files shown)


internal/eslint-plugin-astryx/no-raw-intl-locale.js
```diff
@@ -0,0 +1,540 @@
...
+  'packages/core/src/PowerSearch/formatFilterValue.ts',
+  // The provider-aware Intl.Collator hook — the one place allowed to
+  // construct one, so every other locale-aware string comparison goes
+  // through useCollator() instead of a raw Intl.Collator/localeCompare call.
+  'packages/core/src/i18n/useCollator.ts',
+  // Charts is a separate package that cannot consume React context directly;
+  // its tick/value formatters require an explicit locale at their public
+  // boundary instead (see packages/charts/src/formatters.ts's own doc
+  // comment).
+  'packages/charts/src/formatters.ts',
```
cixzhang, line 115: Exempting the whole file lets `Intl.NumberFormat(undefined)` straight back in.

packages/core/src/NumberInput/NumberInput.test.tsx
```diff
@@ -17,6 +17,15 @@ import {InputGroup} from '../InputGroup';
 import {NumberInput} from './NumberInput';
 import {__resetLiveRegionsForTest} from '../hooks/useAnnounce';
 
+// Test-fixture currency formatter for `formatValue`, standing in for an
+// arbitrary consumer-supplied callback (the prop is free-form application
+// code, not an Astryx formatting surface). A plain thousands-separator insert
+// is enough for these fixed fixture values and keeps the fixture from needing
+// any locale at all.
+function withThousands(value: number): string {
```
cixzhang: Five of eight main hits are explicit-locale fixtures; the oracle list beats rewriting these.

cixzhang (approved):
Thanks — ran it against main: 8 hits, 3 real. Nit before merge: `charts/formatters.ts` is exempt wholesale, so the file the bug was in gets no enforcement.

[Reviewed by Robohands]

internal/eslint-plugin-astryx/no-raw-intl-locale.js
nynexman4464: Fixed — implementation files and test-oracle files now have separate closed lists. Approved implementations still require a syntactically explicit locale, so `Intl.NumberFormat(undefined)` and locale methods without an argument fail. The only ambient production exceptions are scoped to the named `plainDateFormat` and `isLocaleDayFirst` function bodies pending #5120; aliases/destructuring remain rejected.

packages/core/src/NumberInput/NumberInput.test.tsx
nynexman4464: Fixed — restored the explicit-locale NumberInput and Table fixtures and named those exact test files in the oracle list rather than replacing the assertions with manual comparison/formatting logic.

PR #5174, opened 2026-08-18
fix(core): close only the top-most layer on Escape

### LINKED ISSUE #5168: [Bug] Escape closes the layer underneath for Lightbox, MobileNav and lab InfoTip
## Description

One Escape press closes the wrong layer when a **Lightbox**, **MobileNav**, or lab **InfoTip** is open over a `Dialog`:

- **Lightbox / MobileNav** — the `Dialog` underneath closes; the lightbox or nav stays open and visible.
- **InfoTip** (lab) — both close: the tip is dismissed *and* the `Dialog` goes with it.

None of the three take part in the shared layer dismissal stack. Lightbox and MobileNav answer only the `<dialog>` `cancel` event, with no check for whether they are top-most. InfoTip `stopPropagation()`s Escape from its trigger button, so the stack never sees the press — but the browser's own close watcher still fires `cancel` on the dialog behind it.

This is pre-existing on `main`, **not** a regression from #4881: the same harness produces identical results on `main` and on that branch.

## Reproduction

1. Open a `Dialog`.
2. From inside it, open a `Lightbox` (or a `MobileNav`, or hover a lab `InfoTip` until the tip shows).
3. Press Escape once.

**Expected:** the top-most layer closes and the `Dialog` stays open.
**Actual:** Lightbox/MobileNav — the `Dialog` closes and the layer stays on screen; InfoTip — both close.

Measured in real Chromium against a Storybook harness, one press per row. The same harness has `Dialog`, `AlertDialog`, `ContextMenu`, `Tooltip`, `HoverCard`, `Drawer`, `BottomSheet` and `CommandPalette` closing only the inner layer.

## What the working families do

They either register with the stack — `Dialog`, `useFocusTrap` (Popover, DropdownMenu, Bottom Sheet), `useTooltip` and `useHoverCard` all call `useLayerDismissal`, which routes each press to the top-most registered layer — or they claim the press at element level before it reaches the document (`Drawer`, `ContextMenu`).

Lightbox and MobileNav want the same registration, or at minimum the `isTopmost()` gate `Dialog` puts on its own `cancel` handler. InfoTip should let its `Tooltip`'s stack entry take the press rather than swallowing it at the trigger.

## Astryx Version

@astryxdesign/core@0.4.3

## Environment

AKnassa:
Fixes #5168.

## What this does

One press of Escape now closes the thing you actually have open on top — a Lightbox, the mobile navigation drawer, or a lab InfoTip — instead of closing the window sitting behind it.

## Why

Open a photo viewer from inside a dialog and press Escape: the dialog disappeared and the photo viewer stayed on screen. Same with the mobile nav. With an InfoTip, the little help bubble *and* the dialog both closed at once. Each of those leaves you somewhere you didn't ask to be — and for keyboard-only users, Escape is the main way out of a layer.

## What changed

- **Lightbox** and **MobileNav** now take the Escape press for themselves, so it never reaches the dialog behind them.
- **InfoTip** (lab) takes the press only when a tip is actually showing; otherwise Escape belongs to whatever is behind it.
- All three still step aside for a menu or popover opened on top of them — that layer still wins the press.
- 13 new tests, one changeset.

## How to see it

Open a Dialog; from inside it open a Lightbox (or a MobileNav, or focus an InfoTip until the tip shows). Press Escape once — only the inner layer should close.

Measured in Chromium 149 against a Storybook harness, one press per row:

| Layer | Before | After |
| --- | --- | --- |
| Lightbox over Dialog | dialog closed, lightbox stayed open | only the Lightbox closes |
| MobileNav over Dialog | dialog closed, nav stayed open | only the nav closes |
| InfoTip in Dialog | both closed | only the tip closes |

No pixels change, so there are no before/after screenshots — the whole difference is what one key press does.

## Under the hood, for reviewers

All three render inside the host Dialog's DOM subtree, and Dialog listens for `keydown` on its own `<dialog>` element. An unclaimed press bubbles from the inner layer straight into that listener, which closes the Dialog and calls `preventDefault()` — consuming the browser's close request, so the close request that would have dismissed the inner layer never arrives. InfoTip stopped propagation from a React `onKeyDown`, which React dispatches at the root container, above the dialog element whose listener has already run; both layers closed.

Lightbox and MobileNav now attach a `keydown` listener to their own `<dialog>` and consume Escape there — the same shape Dialog already uses, including its `hasActiveFocusTrapEscape()` gate. Their `cancel` handlers gain the same gate, for the case where a trapped layer let the press through and the browser aims its close request at the top-most dialog. InfoTip swaps its React handler for a native listener on the trigger.

## Notes

- Deliberately per-component, and it touches **none** of the files in #4881 (the shared dismissal stack). When that lands and collapses `Dialog`, `useFocusTrap`, `usePopover`, `useTooltip` and `useHoverCard` into `useLayerDismissal`, these three should register with it. What's here is the "or at minimum" gate the issue names, and it fixes `main` today.
- `packages/core/src/MobileNav/MobileNav.test.tsx` is prettier-dirty on `main`. The edit itself is purely additive; the three reformatted `<MobileNav>` blocks in the diff came from the repo's own lint-staged hook at commit time.
- `pnpm lint:strict` 0 errors, `pnpm build` green for core and lab. Full suite 10,669 passing — the 2 failures in `packages/cli` are the pre-existing macOS case-insensitive-filesystem tests, reproduced on a clean tree.

### DIFF
--- a/.changeset/escape-topmost-layer-dialog.md
+++ b/.changeset/escape-topmost-layer-dialog.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Lightbox, MobileNav: one Escape press closes the layer itself instead of the Dialog it was opened from. Both now claim the press at their own `<dialog>` element, so it never reaches the host's listener or the browser's close request, and both still defer to a focus-trapped layer (popover, menu) above them.
+
+@AKnassa
--- a/packages/core/src/Lightbox/Lightbox.test.tsx
+++ b/packages/core/src/Lightbox/Lightbox.test.tsx
@@ -3,6 +3,8 @@
 import {describe, it, expect, vi, beforeEach, afterEach} from 'vitest';
 import {render, screen, fireEvent, waitFor} from '@testing-library/react';
 import {Lightbox} from './Lightbox';
+import {Dialog} from '../Dialog';
+import {useFocusTrap} from '../hooks/useFocusTrap';
 import {__resetLiveRegionsForTest} from '../hooks/useAnnounce';
 import {InternationalizationProvider} from '../i18n';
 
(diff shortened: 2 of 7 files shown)


cixzhang:
Thanks for this — and sorry, the issue should have said so: #4881 is open and replaces the mechanism you're building on with a stack that routes each Escape to the top-most layer. Eight families already register with it, so once it lands the fix for these three is `useLayerDismissal({...})` per component, much smaller than this.

Leaving it to you whether to park this until #4881 merges or close it and take the smaller version then. Happy to ping you when it lands.

PR #5175, opened 2026-08-18
feat(core): give StatusDot the same colour variants Badge exposes

cixzhang:
> **Draft — not mergeable as it stands.** The implementation is complete and
> green, but measurement across the bundled themes found the fill it depends on
> is not safe in three of them. The numbers are below; the design question they
> raise is the reason this is still a draft.

## Why

`StatusDot` has the five sentiment variants and nothing else. A consumer
colouring dots by **category** rather than by health — a legend key, a project
colour, a per-section marker in a tab strip — has no variant to reach for, and
ends up overriding the fill from outside the component.

`Badge` already carries a nine-hue non-semantic set for exactly this. This adds
the same set to the dot, under the same names, so the two agree when they sit
in the same row.

## What

`blue`, `cyan`, `green`, `orange`, `pink`, `purple`, `red`, `teal`, `yellow` —
added to `StatusDotVariantMap`, so they widen `StatusDotVariant` and stay
augmentable. The five existing variants are untouched and **no tokens were
added**.

Each hue paints from the solid `--color-text-<hue>` stop, with the surface
colour as ink for a passed `icon` — the pairing `neutral` already uses.

## The blocker

An 8px dot is a filled shape: it needs 3:1 against the page (WCAG 2.1 SC
1.4.11). The hue ramp has four stops and **none of them is a fill**. Measured
in Chromium, against the surface the dot sits on:

| Stop | Fails at 3:1 in |
|---|---|
| `--color-background-<hue>` (20% wash) | almost everywhere — 1.2–1.5:1 at this size |
| `--color-border-<hue>` | five theme/scheme pairs, all nine hues |
| `--color-icon-<hue>` | light mode of four themes |
| `--color-text-<hue>` (what this PR uses) | three themes, all nine hues |

The last row is the one that matters. It holds where the ramp flips polarity
with the colour scheme:

| Theme | sentiment variants (min) | hues (min) |
|---|---|---|
| neutral light | 4.74:1 | 8.83:1 |
| neutral dark | 6.00:1 | 10.14:1 |
| y2k dark | 7.91:1 | **1.06:1** |
| butter dark | 6.81:1 | **1.40:1** |
| gothic (dark only) | 7.03:1 | **1.37:1** |

Those three themes tune the hue ramp to the contract `Badge` relies on — the
wash is the plate, `--color-text-<hue>` is ink *on that plate* — and they do
not flip it per scheme, because they only have the one. So the stop that reads
as a fill in the base tokens reads as ink-on-dark there, and the dot
disappears. The sentiment variants are fine in the same themes, so this is not
the pre-existing "warning is 1.5:1 in neutral light" problem — the hues would
be materially worse than what the component ships today.

## The question this raises

Every stop in the hue ramp is defined against something *else* — a wash to sit
on, a page to be read on, a plate to be inked onto. There is no per-hue value
whose contract is "the hue as a fill that reads against the page", which is
what a dot needs and what a theme author would have to supply.

That leaves three options, and picking one is a system-level call rather than a
component one:

1. **Give the ramp a fill stop.** Solves it once for every consumer of a hue as
   a shape, not just this dot.
2. **Make every theme own its dots.** `.astryx-statusdot` already carries
   `data-variant`, so a theme *can* fix this — but the three bundled themes
   would each need a nine-entry block, and a third-party theme would break
   silently until someone noticed.
3. **Don't put the hues on the dot.** The categorical ramp stays a
   plate-and-ink ramp, and colouring a dot by category stays outside the
   component.

## Testing

`pnpm test`, typecheck and lint clean. New tests assert the plate/ink pairing
per hue against a StyleX probe; the existing per-variant tests now run over all
fourteen. Every variant was rendered in Chromium in both schemes, plain and
with an icon, and re-measured under each bundled theme — that is where the
blocker above came from.

### DIFF
--- a/.changeset/statusdot-hue-variants.md
+++ b/.changeset/statusdot-hue-variants.md
@@ -1,0 +1,24 @@
+---
+'@astryxdesign/core': patch
+'@astryxdesign/cli': patch
+---
+
+[feat] StatusDot: the nine hue variants — `blue`, `cyan`, `green`, `orange`, `pink`, `purple`, `red`, `teal`, `yellow`
+
+The dot had the five sentiment variants and nothing else, so a consumer
+colouring dots by category rather than by health — a legend key, a project
+colour, a tab strip's per-section marker — had no variant to reach for and
+fell back to overriding the fill. `Badge` already carries exactly this set
+under exactly these names; the dot now matches it, so the two agree when they
+sit in the same row.
+
+The dot paints from the solid `--color-text-<hue>` stop, not the
+`--color-background-<hue>` wash `Badge` fills its pill with: the wash is 20%
+alpha, and on an 8px shape it lands at 1.2–1.5:1 against the page instead of
+the ≥8.3:1 the solid stop holds in both colour schemes. A user-supplied `icon`
+paints from the surface as ink, the same pairing `neutral` already uses,
+measured at ≥8.3:1 on every hue in both schemes.
+
+No new tokens, and no change to the five existing variants.
+
+@cixzhang
--- a/apps/storybook/stories/StatusDot.stories.tsx
+++ b/apps/storybook/stories/StatusDot.stories.tsx
@@ -3,6 +3,18 @@
 import type {Meta, StoryObj} from '@storybook/react';
 import {StatusDot} from '@astryxdesign/core/StatusDot';
 
+const HUES = [
+  'blue',
+  'cyan',
+  'green',
+  'orange',
+  'pink',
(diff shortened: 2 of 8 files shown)


cixzhang:
Closing by API decision: `StatusDot` stays semantic-status-only; categorical hue variants belong with components such as `Badge`, not on the status primitive. The existing semantic variants are already theme-controlled on main, so this branch's remaining categorical surface should not ship.

PR #5179, opened 2026-08-18
feat(Selector): add selector-option-row theme target on the option row

freddymeta:
## Summary

The `Selector` dropdown option row is a bare `div[role="option"]` carrying only StyleX hashes — it has no theme target. Consumers who want to restyle row padding or density have to reach it with a structural selector like `div[role='listbox'] [role='option']`, which is brittle and leaks across unrelated listboxes.

This adds **`selector-option-row`** on that row, mirroring the `multi-selector-option` target `MultiSelector` already exposes. It carries `size` plus `selected` / `disabled` state, so a theme can express "selected option at large" or adjust a given row density directly.

```ts
defineTheme({
  name: 'my-theme',
  components: {
    'selector-option-row': {
      base: { padding: 'var(--spacing-2)' },
      selected: { backgroundColor: 'var(--color-background-muted)' },
      'size:md': { padding: 'var(--spacing-2)' },
    },
  },
});
```

## Why `-option-row` and not `-option`

`selector-option` is already taken: it's the public `SelectorOption` **content** primitive (the `Item` holding a row's icon/label/description), not the row container. `MultiSelector` has no equivalent content primitive, so it could use `multi-selector-option` for its row. Here the row and the content are distinct elements, so the row target is disambiguated as `selector-option-row`. State/size vocabulary matches `multi-selector-option` for cross-component consistency.

Motivating case: the row's padding is split across a base (`--spacing-2`, 8px) and a per-size override where the default `md` trims only the block axis to `--spacing-1-5` (6px). Reaching that to normalize row density previously required a structural selector; now it's a first-class seam.

## Test plan

- `pnpm -F @astryxdesign/core test` — `Selector` suite passes, plus three added tests: the target lands on every dropdown row with its `data-size`, reflects `data-selected`, and a `defineTheme` generation check proves a theme reaches the row + its states + per-size density.
- `pnpm -F @astryxdesign/core build` / `typecheck` — clean. `astryx component Selector` lists the new target.
- Docs (`.doc.mjs` theming targets) and a changeset are included. Purely additive — the class/attributes change nothing until a theme targets them.


freddymeta:
Added test parity with `multi-selector-option`'s row coverage: an unselected/enabled row carries neither `data-selected` nor `data-disabled`, a disabled option reflects `data-disabled`, and the target survives a `renderOption` override. Selector suite green (131).

One separate follow-up worth a maintainer's eye (not part of this PR): the `md` option row paints **6px block / 8px inline** — `itemSizeStyles.md` overrides only `paddingBlock` to `--spacing-1-5` while the base keeps `--spacing-2` inline. If that asymmetry is unintended, it's worth symmetrizing the default upstream; this target just makes the row density themeable either way.

packages/core/src/Selector/Selector.tsx
```diff
@@ -1164,12 +1164,27 @@ export function Selector<T extends SelectorOptionType>(
...
           onClick={() => onItemSelect(item)}
           onMouseEnter={() => onItemMouseEnter(item, flatIndex)}
-          {...stylex.props(
-            styles.item,
-            itemSizeStyles[size],
-            isHighlighted && styles.itemHighlighted,
-            isSelected && styles.itemSelected,
-            item.disabled && styles.itemDisabled,
+          {...mergeProps(
+            // Stable theme target on the option row itself, mirroring
```
cixzhang, line 1261: Comment could be a line or two; the naming rationale already lives in the description.

cixzhang (approved):
Thanks this is good. Checked in Chrome that a theme reaches the row, its states and its size — merging as is.

[Reviewed by Robohands]

PR #5181, opened 2026-08-18
fix(NumberInput): track themed padding so the stepper column stays flush

freddymeta:
## Summary

`NumberInput`'s number-stepper column is positioned by **cancelling the field wrapper's own padding** rather than owning its geometry, and it assumes the *default* padding:

- `numberSteppers { marginBlock: calc(-1 * --spacing-1) }` — hardcodes the 4px default block padding, so the column spans the field's full height.
- `wrapperWithNumberSteppers { padding-inline-end: 0 }` — assumes it can zero the end padding.
- The stepper corners hardcode `--radius-element`, while the wrapper itself uses the themeable `--_field-radius`.

The moment a theme changes `number-input` padding (or radius), those assumptions break: the column falls short of the top/bottom edges (a visible gap), and the corners no longer match the field radius. Padding is a legitimate themeable property, so this is a real robustness gap, not a niche one.

## Fix

Carry the wrapper's padding through private custom properties the steppers read — the same pattern `TextArea` already uses for its overlaid affordances (`--_textarea-inline-padding`), and mirroring how `--_field-radius` is already threaded:

- The wrapper declares `--_number-input-padding-block` / `--_number-input-padding-inline` (defaults match the shared field base) and applies them.
- The stepper column cancels the block padding with `calc(-1 * var(--_number-input-padding-block))`, and its corners read `var(--_field-radius)`.
- `number-input` is registered in the derived-var registry so a theme's `paddingBlock` / `paddingInline` flows into those vars (`replaces: true`, so the raw longhand doesn't double-apply) — the column tracks whatever the theme sets.

Byte-identical by default (the vars resolve to today's tokens); inert for the no-stepper case and every other input.

## Test plan

- `pnpm -F @astryxdesign/core test` — NumberInput + theme suites pass (212). Added: the wrapper carries the padding vars and the stepper margin reads the block var; a themed `number-input` padding routes into the private vars (not raw padding) in generated CSS.
- `derivedVarRegistry.test.ts` (registry↔doc consistency) passes with the new `number-input` entry.
- `pnpm -F @astryxdesign/core build` / `typecheck` / `eslint` — clean. `astryx component NumberInput` documents the themeable padding.
- Docs (`.doc.mjs` theming vars + derived, EN + zh) and a changeset are included.

### DIFF
--- a/packages/core/src/theme/derivedVarRegistry.ts
+++ b/packages/core/src/theme/derivedVarRegistry.ts
@@ -67,6 +67,18 @@ export const derivedVarRegistry: Record<string, DerivedVarEntry[]> = {
   ],
   field: [{property: 'borderRadius', vars: ['--_field-radius']}],
   hovercard: [{property: 'borderRadius', vars: ['--_hovercard-radius']}],
+  'number-input': [
+    {
+      property: 'paddingBlock',
+      vars: ['--_number-input-padding-block'],
+      replaces: true,
+    },
+    {
+      property: 'paddingInline',
+      vars: ['--_number-input-padding-inline'],
+      replaces: true,
+    },
+  ],
   popover: [{property: 'borderRadius', vars: ['--_popover-radius']}],
   'progressbar-mark': [
     {property: 'width', vars: ['--_progressbar-mark-width'], replaces: true},
--- a/packages/core/src/theme/derivedVarRegistry.ts
+++ b/packages/core/src/theme/derivedVarRegistry.ts
@@ -68,16 +68,8 @@ export const derivedVarRegistry: Record<string, DerivedVarEntry[]> = {
   field: [{property: 'borderRadius', vars: ['--_field-radius']}],
   hovercard: [{property: 'borderRadius', vars: ['--_hovercard-radius']}],
   'number-input': [
-    {
-      property: 'paddingBlock',
-      vars: ['--_number-input-padding-block'],
-      replaces: true,
-    },
-    {
-      property: 'paddingInline',
-      vars: ['--_number-input-padding-inline'],
-      replaces: true,
-    },
+    {property: 'padding', expand: 'container'},
+    {property: 'borderRadius', vars: ['--_field-radius']},
   ],
   popover: [{property: 'borderRadius', vars: ['--_popover-radius']}],
   'progressbar-mark': [
--- a/.changeset/number-input-stepper-padding-var.md
+++ b/.changeset/number-input-stepper-padding-var.md
@@ -2,6 +2,6 @@
 '@astryxdesign/core': patch
 ---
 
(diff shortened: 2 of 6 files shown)


packages/core/src/theme/derivedVarRegistry.ts
```diff
@@ -67,6 +67,18 @@ export const derivedVarRegistry: Record<string, DerivedVarEntry[]> = {
   ],
   field: [{property: 'borderRadius', vars: ['--_field-radius']}],
   hovercard: [{property: 'borderRadius', vars: ['--_hovercard-radius']}],
+  'number-input': [
```
cixzhang, line 70: A theme writing `padding: 14px 20px` or `paddingBlockStart` skips these vars and the gap comes back.

cixzhang (changes requested):
Thanks — checked it in Chrome and the column stays flush with themed padding, in RTL and at 200% zoom.

The reason I'd hold it is the theme author's experience. This reads `--_number-input-padding-inline`/`-block`, so it only fires when someone writes exactly `paddingInline` or `paddingBlock`. A theme that writes `padding: 14px 20px`, or `paddingBlockStart`, gets the old broken gap back with nothing to tell them why — I hit 10/10/20px and a 12px top gap doing that. Silently working for one spelling of the same declaration is the footgun.

`expand: 'container'` already parses every spelling (Card, Dialog and Section use it). Could this go through that instead of a sixth hand-rolled entry? If it can't emit component-private vars, say so and I'll take that as ours to fix in the theming infra.

Also worth trimming the changeset: theming `number-input` borderRadius still leaves the steppers at the old radius — field 2px, steppers 10px in my run.

cixzhang (approved):
Approving — this is exactly it, thank you.

Re-checked in Chrome against your previous push. Both spellings I named are closed, and the asymmetric case you found on your own was the sharpest one:

| theme on `number-input` | before | now |
|---|---|---|
| `padding: '14px 20px'` | 10px gap top+bottom | flush |
| `paddingBlockStart: '12px'` | 8px gap top | flush |
| `paddingBlock: '4px 12px'` | `calc(-1 * 4px 12px)` invalid, margin dropped | flush |
| `borderRadius: '2px'` | field 2px, steppers 10px | both 2px |

Flush in RTL too. `derivedVarRegistry.test.ts:188` is the best part of the diff — that var-count bail is why the first version could land unchecked, and you found it unprompted.

Two things, neither yours and neither blocking:

- Physical longhands still escape — `paddingTop: '14px'` leaves a 10px gap. `PADDING_PROPS` (`generateThemeRules.ts:121`) lists only the logical spellings, so Card/Dialog/Section have it identically. Ours; I'll file it.
- Could you add a themed before/after screenshot to the description? It's the one thing a visual PR is asked to carry — I have shots if you want them.

The changeset reads long now; a trim to the big picture would be welcome, but don't hold the PR on it.

PR #5183, opened 2026-08-18
feat(core): let CheckboxInput and Switch name their own label theme target

freddymeta:
## What

`CheckboxInput` and `Switch` now name their own label as a theme target —
`astryx-checkbox-label` and `astryx-switch-label` — alongside the
`astryx-field-label` every label already carries.

A theme could previously only reach one `astryx-field-label` target, so styling
a checkbox's label (which shares a row with its control) meant styling every
form field's label above its input too.

## Why not the `layout` prop this PR started with

Review was right to hold it, and for a reason that went further than the API
shape:

- `layout` was an encoding the label could not verify. A consumer could pass
  `layout="beside"` on a stacked label and the theme would quietly believe it.
- The label cannot know how it was arranged — that is its caller's business.
- There are three placements, not two. `Field direction="horizontal-labels"`
  also sits a label beside its control, and would have reported `stacked`, so a
  theme's stacked rule (say, a bottom margin) would land on it and drift the
  label off its input.

The control knows what kind of label it owns, so it passes the target down:

```tsx
<FieldLabel {...themeProps('checkbox-label')} … />
```

`FieldLabel` composes it onto the `astryx-field-label` it already renders. The
name says what the thing *is* rather than how it is arranged, nothing can set it
untruthfully, and the horizontal-labels case needs nothing at all — no label
claims a placement any more.

## Test plan

- New tests on all three components: the two classes compose onto one element,
  and `FieldLabel` asserts it emits no `data-layout` of its own. **Removing each
  `themeProps(...)` spread fails exactly the matching test** — verified in both
  directions.
- The targets are documented on their own component (EN + zh), which is what
  makes them *checked*: `themingTargets.test.ts` fails when a rendered
  `astryx-*` class is missing from that component's doc. Verified by removing
  each doc entry in turn.
- `pnpm test` 533 files / 10899 tests, `lint:strict` 0 errors, `pnpm build`,
  and every typecheck the `build-storybook` job runs — including
  `typecheck:docs`, which is a separate gate from `typecheck` and the one that
  caught an unsupported `description` key on a theming target.
- Reverts the `visualProps: ['layout']` this had added to `astryx-field-label`
  in `Field.doc.mjs`. Current `main` is merged in.


cixzhang (changes requested):
Thanks — a theme genuinely can't tell a checkbox label from a form label today, so this is a real gap.

I'd rather not solve it with a prop, though. `layout` changes nothing but the `data-` attribute, so a consumer can pass `layout="beside"` on a stacked label and the theme quietly lies. It's also not theirs to know: `CheckboxInput` and `Switch` already know they're a control row.

Could the control pass a target down instead?

```tsx
<FieldLabel {...themeProps('checkbox-label')} … />
```

Then a theme names the thing it wants rather than knowing an encoding, nobody can set it untruthfully, and it matches how the other targets landed recently.

It also sidesteps a case this misses: `Field direction="horizontal-labels"` puts a label beside its control and still reports `stacked`, so the stacked bottom margin lands on it — I applied it in Chrome and those labels drift off their inputs.

cixzhang (approved):
Thanks — this is the shape I wanted, and the horizontal-labels case falls out for free. Verified both targets reach the label in Chrome.

[Reviewed by Robohands]

PR #5191, opened 2026-08-18
fix: Escape dismisses only the top-most overlay for Lightbox/MobileNav

### LINKED ISSUE #5168: [Bug] Escape closes the layer underneath for Lightbox, MobileNav and lab InfoTip

HelloOjasMutreja:
Fixes #5168.

## Problem

`Dialog` already defers to any popover-style overlay layered on top of it via a shared "focus-trap Escape stack" (`hasActiveFocusTrapEscape()` in `useFocusTrap.ts`), so pressing Escape with a `Popover`/`DropdownMenu` open inside a `Dialog` correctly dismisses only the top layer. `Lightbox` and `MobileNav` never registered on that same stack, even though both also open as native top-layer `<dialog>` elements. Reproduced live: with a `Dialog` open and a `Lightbox` open on top of it, one Escape press closed the `Dialog` underneath while the `Lightbox` on top stayed open, exactly backwards from expected.

Root cause: `Dialog`'s own document-level Escape handler only checks `hasActiveFocusTrapEscape()` as a boolean gate. Since `Lightbox`/`MobileNav` never pushed anything onto that stack, the gate read "nothing active" even while one of them was genuinely the top-most overlay, so `Dialog` closed itself via its own JS handler instead of deferring.

Separately, lab's `InfoTip` had a related but distinct bug: its own Escape handling called `stopPropagation()` to dismiss its tooltip, but never `preventDefault()`. `stopPropagation()` on the React synthetic event stops the shared stack's document listener from seeing the press, but does not stop the browser's own native top-layer close-watcher, which processes Escape independently. Result: an ancestor `Dialog` still closed on the same press.

## Fix

Added `useEscapeStackEntry(isActive, getContainer)`, a small hook extracted from `useFocusTrap`'s existing stack registration logic, exported alongside `hasActiveFocusTrapEscape`. It only pushes/removes a presence on the shared stack; it does not attach its own Escape handling or pull in `useFocusTrap`'s JS focus-trap/Tab-wrapping machinery, since `Lightbox` and `MobileNav` already trap focus and dismiss on Escape natively via `dialog.showModal()`/the `cancel` event. Wired into both.

For `InfoTip`, added the missing `preventDefault()` alongside the existing `stopPropagation()`.

## Verification

- `useFocusTrap.test.tsx`: added a `useEscapeStackEntry` suite (registration/removal across mount, unmount, and `isActive` toggling, plus a test simulating Dialog's own gate to confirm it correctly defers while the hook is active and responds again once it isn't).
- `InfoTip.test.tsx`: added a test asserting `event.defaultPrevented` is `true` on the Escape that dismisses the tooltip.
- Existing `Lightbox`, `MobileNav`, and `useFocusTrap` suites still pass unchanged.
- Live-verified against a temporary Storybook harness (Dialog + Lightbox both open): confirmed the original bug before this fix (Dialog closed, Lightbox stayed open on one Escape press), then confirmed after this fix that Dialog's own JS-driven self-close no longer fires.
- One thing I could **not** fully verify end-to-end in this environment: whether the browser's native `<dialog>` close-watcher then correctly targets the true top-most dialog (Lightbox) once Dialog stops intercepting. My browser automation session runs with `document.visibilityState: "hidden"` (not compositing frames), and Chrome gates the native close-watcher's default action on real page visibility, confirmed by testing a `Lightbox` alone (no `Dialog` involved) also not self-closing on a synthetic Escape in this same session. This is a tooling limitation, not something in question about the fix itself: the mechanism it relies on (native `<dialog>` top-layer close-watcher plus `preventDefault()` suppression) is the same one `Dialog` and every other native-`<dialog>` consumer in this codebase already depend on, and is inherently outside jsdom's/this automation session's reach, similar in kind to why this class of bug wasn't caught by unit tests in the first place.

Typecheck, lint, and the full affected test suites (`useFocusTrap`, `Lightbox`, `MobileNav`, `InfoTip`) pass locally.

### DIFF
--- a/.changeset/escape-key-wrong-layer.md
+++ b/.changeset/escape-key-wrong-layer.md
@@ -1,0 +1,9 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Escape now dismisses only the top-most overlay when a `Lightbox` or `MobileNav` is layered on top of a `Dialog`. `Dialog` already deferred to any popover-style overlay via the shared focus-trap Escape stack, but `Lightbox` and `MobileNav` never registered on it, so a single Escape press closed the `Dialog` underneath instead of the overlay on top. Both now register on the shared stack while open (via the new `useEscapeStackEntry` hook), so `Dialog` correctly defers to them.
+
+Also fixes `InfoTip` (lab): dismissing its own tooltip on Escape only called `stopPropagation`, not `preventDefault`, so a `Dialog` behind it still closed via its native top-layer close-watcher on the same press.
+
+@HelloOjasMutreja
--- a/packages/core/src/Lightbox/Lightbox.tsx
+++ b/packages/core/src/Lightbox/Lightbox.tsx
@@ -31,6 +31,7 @@ import {Icon} from '../Icon';
 import {IconButton} from '../IconButton';
 import {useAnnounce} from '../hooks/useAnnounce';
 import {useScrollLock} from '../hooks/useScrollLock';
+import {useEscapeStackEntry} from '../hooks/useFocusTrap';
 import {useIsomorphicLayoutEffect} from '../hooks/useIsomorphicLayoutEffect';
 import {mergeProps, mergeRefs, rtlStyles} from '../utils';
 import type {BaseProps} from '../BaseProps';
@@ -350,6 +351,13 @@ export function Lightbox({
 
   // Scroll lock
   useScrollLock(isOpen);
+
+  // Register on the shared Escape stack while open so a Dialog (or other
+  // overlay) layered underneath correctly defers to this Lightbox instead of
+  // closing itself on the same Escape press. The Lightbox's own dismissal is
+  // still handled natively below via the dialog's `cancel` event.
+  const getEscapeContainer = useCallback(() => dialogRef.current, []);
+  useEscapeStackEntry(isOpen, getEscapeContainer);
 
   // Reset zoom on image change
   useEffect(() => {
--- a/packages/core/src/MobileNav/MobileNav.tsx
+++ b/packages/core/src/MobileNav/MobileNav.tsx
@@ -52,6 +52,7 @@ import {
   holdScrollbarGutter,
   type ScrollbarGutterHold,
 } from '../hooks/scrollbarGutter';
+import {useEscapeStackEntry} from '../hooks/useFocusTrap';
 import {mergeProps, mergeRefs, composeEventHandlers} from '../utils';
 import {overlayPaddingReset} from '../Layout/padding.stylex';
 import type {BaseProps} from '../BaseProps';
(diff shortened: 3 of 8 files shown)


cixzhang:
Thanks for jumping on this so fast — and sorry, the issue should have told you: #4881 is open and replaces `hasActiveFocusTrapEscape()` with a layer stack that routes each Escape to the top-most layer. Eight families already register with it. Building on the boolean means this gets undone when that lands, which is on me for not saying so.

The shape once #4881 is in is `useLayerDismissal({...})` per component, the way Dialog and Drawer do it — much smaller than what you have here. Happy to ping you when it merges if you want to re-target it.

Your InfoTip finding is separate and real: `stopPropagation()` without `preventDefault()`. Worth keeping either way.

HelloOjasMutreja:
Makes sense, thanks for the context. #4881 is still open, so I'll leave this as-is for now and wait for your ping to re-target onto useLayerDismissal() once it lands rather than guess at the shape ahead of it landing.

Leaving the InfoTip preventDefault fix in either way since you confirmed it's a real, separate issue.

HelloOjasMutreja:
Flagging that #4881 has now landed. Happy to retarget this onto useLayerDismissal() as discussed whenever you'd like me to go ahead, just let me know.

cixzhang:
Thanks for holding this while #4881 landed. This head should close rather than rebase: current `main` already moves Lightbox and MobileNav onto `useLayerDismissal`, and InfoTip delegates Escape to Tooltip.

The remaining code here now conflicts with that owner. `useEscapeStackEntry` adds a second registration path through `useFocusTrap` and deliberately makes `hasActiveFocusTrapEscape()` true for non-trap layers. That helper is released trap-only behavior; widening it breaks callers such as `BottomSheetSwitcher`. #4881 includes the focused browser/unit coverage for all three affected surfaces, so there is nothing left to retarget.

[Reviewed by Robohands]

HelloOjasMutreja:
Understood, thanks for the detailed explanation. Closing rather than rebasing since #4881 already covers all three surfaces with better coverage than what's here.

PR #5194, opened 2026-08-18
feat(i18n): expose provider-backed locale hooks

nynexman4464:
## Summary

- add public `useLocale()` for threading the active `InternationalizationProvider` locale into pure helpers and sibling packages
- add public `useCollator(options)` for provider-bound, memoized string comparison
- document and test provider changes plus the existing English fallback when no provider is present

## API

```tsx
const locale = useLocale();
const collator = useCollator({numeric: true});
```

This PR is additive. It does not migrate components and does not enable lint enforcement.

## Stack

1. **This PR:** provider-backed locale hooks
2. Non-date locale migrations
3. #5120: date formatting/parsing migration
4. #5171: lint enforcement

#5190 remains independent and shares no files with this PR.

## Verification

- focused hook suites: **10/10**
- Core typecheck: pass
- CLI structure, SYNC comments, package boundaries, changesets, client-module, portable-script, and i18n-catalog checks: pass
- `git diff --check`: pass

### DIFF
--- a/packages/core/src/i18n/useCollator.ts
+++ b/packages/core/src/i18n/useCollator.ts
@@ -27,7 +27,7 @@ import {InternationalizationContext} from './InternationalizationContext';
 
 /**
  * Returns an `Intl.Collator` for the active provider locale, memoized across
- * renders where the locale and options are unchanged.
+ * renders where the locale and options object are unchanged.
  *
  * @example
  * ```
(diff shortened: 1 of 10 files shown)


packages/core/src/i18n/useCollator.ts
```diff
@@ -0,0 +1,71 @@
...
+ * @example
+ * ```
+ * const collator = useCollator({numeric: true});
+ * const sorted = [...items].sort((a, b) => collator.compare(a.name, b.name));
+ * ```
+ */
+export function useCollator(options?: Intl.CollatorOptions): Intl.Collator {
+  const ctx = use(InternationalizationContext);
+  const {
+    localeMatcher,
```
cixzhang: Hand-listing the seven CollatorOptions fields — a new one gets silently dropped.

cixzhang (approved):
Thanks, this is clean — nice that the options destructure keeps memoization working with an inline object.

Both are documented, but they'll land under "Other" in `astryx hook list`: the doc files say `group: 'Utilities'` and discovery reads `category:`. Every hook in `src/hooks/` uses `category`. Copied from `useTranslator`, so it's pre-existing — worth fixing all three while you're here.

Approving; merge when you're ready.

[Reviewed by Robohands]

packages/core/src/i18n/useCollator.ts
nynexman4464: Fixed — `useCollator` now passes the complete options object directly to `Intl.Collator` and memoizes by locale plus options identity, so future CollatorOptions fields are not silently dropped. The tests now cover stable and changed options identity.

PR #5195, opened 2026-08-18
fix(i18n): use provider locale outside date inputs

nynexman4464:
## Summary

- use the provider locale for Table fallback collation and custom-comparator examples
- thread provider locale through PowerSearch and Timestamp number/date formatting
- make Charts formatter caches key by an explicit provider-threaded locale
- preserve the existing symbol-first `currency(symbol, locale?)` API while passing provider locale last
- default speech recognition to the provider locale while preserving an explicit `lang` override
- add regression coverage for live provider changes

Compatibility note: `currency` keeps `locale` optional so existing consumers can continue calling `currency('€')` without a breaking API change. Astryx-owned call sites pass the provider locale as the second argument. A future breaking release can make that locale required once callers have migrated.

This PR does not touch Calendar, date inputs, `plainDate`, `dateParser`, or CLDR weekday data.

## Stack

1. #5194: provider-backed locale hooks
2. **This PR:** non-date migrations
3. #5171: lint enforcement

#5120 remains an independent date migration. #5190 remains independent and shares no files with this PR.

## Verification

- focused migration suites: **330/330**
- Core and Charts typechecks: pass
- Core and Charts builds: pass
- full workspace build: pass
- Storybook typecheck after workspace build: pass
- `pnpm lint:strict`: pass
- repository structural checks and `git diff --check`: pass

### DIFF
--- a/packages/charts/src/formatters.ts
+++ b/packages/charts/src/formatters.ts
@@ -12,34 +12,43 @@
  * across the full magnitude range (K/M/B/T and beyond).
  */
 
-/** Lazily create a value once and cache it — keeps `Intl` construction off the import path. */
-function once<T>(create: () => T): () => T {
-  let cached: {value: T} | null = null;
-  return () => (cached ??= {value: create()}).value;
+import type {Locale} from '@astryxdesign/core/i18n';
+
+/** Lazily cache one formatter per locale. */
+function byLocale<T>(create: (locale: Locale) => T): (locale: Locale) => T {
+  const cache = new Map<Locale, T>();
+  return locale => {
+    let value = cache.get(locale);
+    if (value === undefined) {
+      value = create(locale);
+      cache.set(locale, value);
+    }
+    return value;
+  };
 }
 
-const getCompactFormat = once(
-  () =>
-    new Intl.NumberFormat(undefined, {
+const getCompactFormat = byLocale(
+  locale =>
+    new Intl.NumberFormat(locale, {
       notation: 'compact',
       maximumFractionDigits: 1,
     }),
 );
 
-const getPercentFormat = once(
-  () =>
-    new Intl.NumberFormat(undefined, {
+const getPercentFormat = byLocale(
+  locale =>
+    new Intl.NumberFormat(locale, {
       style: 'percent',
       maximumFractionDigits: 1,
     }),
 );
 
-const getShortDateFormat = once(
-  () => new Intl.DateTimeFormat(undefined, {month: 'short', day: 'numeric'}),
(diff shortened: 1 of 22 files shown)


packages/charts/src/formatters.ts
```diff
@@ -64,12 +73,15 @@ export function compactNumber(value: unknown): string {
...
- * <ChartAxis tickFormat={currency()} />       // $1.5K
- * <ChartAxis tickFormat={currency('€')} />    // €1.5K
- * <ChartAxis tickFormat={currency('¥')} />    // ¥1.5K
+ * <ChartAxis tickFormat={currency(locale)} />       // $1.5K
+ * <ChartAxis tickFormat={currency(locale, '€')} />  // €1.5K
+ * <ChartAxis tickFormat={currency(locale, '¥')} />  // ¥1.5K
  * ```
  */
-export function currency(symbol = '$'): (value: unknown) => string {
+export function currency(
```
cixzhang, line 83: Locale first here, last in the other four. `currency('€')` compiles and throws.

cixzhang (changes requested):
Thanks for this — threading the provider through was the fiddly part. Checked it in Chromium with the provider on `sv-SE` and the browser on `de-DE`: Timestamp and Table follow the provider now, where main followed the browser.

I'd like the charts signature fixed before this lands. `currency` takes the locale first, the other four take it last, and both arguments are `string` — so `currency('€')` quietly means "locale `€`" and throws `RangeError: Invalid language tag` at the first value over 1000 instead of failing to compile. Nothing outside our own stories calls it yet, so this is the cheap moment:

```ts
export function currency(symbol = '$', locale: Locale): (value: unknown) => string
```

One question: an app with no provider moves from the browser locale to English here. Worth a line in the changeset?

[Reviewed by Robohands]

packages/charts/src/formatters.ts
nynexman4464: Fixed — restored the symbol-first API as `currency(symbol = "$", locale?)`, matching the existing call shape and putting locale last like the other helpers. Astryx examples use `currency("$", locale)`, and a regression test verifies `currency("€")` remains valid and does not throw.

cixzhang (approved):
Thanks, the reorder does exactly what I wanted — `currency('€')` is a type error now.

One note and it's your call: you made `locale` optional here while the other four require it. Omitting it falls back to the ambient locale, which is the bug this PR exists to close, and the lint rule won't catch a missing argument. Happy either way — just flagging that the four-required-one-optional split is the kind of thing the next person copies.

[Reviewed by Robohands]

Issue #3281, opened 2026-06-30
[RFC] Listbox + TransferList — inline selectable lists for complex multi-item selection


ernestt:
> Follows the Astryx Component Lifecycle (Phase 1: Specification). This is a problem statement and demand signal with front-loaded research and a candidate API — final API shapes run through API Arbitration / vibe tests.

---

## Problem Statement

Builders need to expose a set of options **inline, always visible, in a bordered container**, and let users **select one or many** — and for the hardest cases, **manage an ordered set**: move items between an "available" pool and a "selected" list, reorder the selected list by dragging, lock specific items so they can't be removed or moved, and search across everything.

This maps directly to the [NN/g Listbox vs. Dropdown distinction](https://www.nngroup.com/articles/listbox-dropdown/): a **listbox exposes its options immediately and supports multi-selection**, whereas a dropdown hides options behind a click and is single-select. NN/g explicitly names the advanced forms — the **multiselect listbox** and the **multiselect dual listbox**, where users "make selections by moving items from one listbox to another" and "reorder the options by moving them up and down." That dual-listbox is precisely the "complex selection" pattern we lack.

Astryx today only offers *collapsed* selection or *flat* inline checkboxes:

- `Selector` / `MultiSelector` — options live behind a dropdown trigger; you must open them to see or change anything. Single value or a flat checkbox set. No reorder, no per-item affordances, no two-list transfer.
- `CheckboxList` — inline and always-visible, but flat: no reordering, no per-item controls, no add/remove-between-lists, no selection summary, no single-select mode.

There is **no primitive** for the always-visible selectable list, and **nothing** for the dual-list "configure an ordered set" editor. Teams rebuild the latter by hand every time — wiring a drag library, custom row chrome, ad-hoc keyboard handling, and `listbox`/`option` ARIA that is almost always wrong.

## Evidence of Demand

**The motivating pattern (from the attached reference):** a list-management editor in a dialog — a search field over two side-by-side panels, **"Selected items"** (with per-row drag handle + remove ×, plus a "Remove all" bulk action) and **"Available items"** (with per-row add +, plus an "Add all" bulk action), one row locked ("irremovable") so it can't be reordered or removed, and a staged **Cancel / Apply** footer. The trigger summarizes state as "7 items selected".

This is a generalized **dual/transfer listbox**. It is the standard way to configure an ordered subset of options — it recurs across:

[...]

PR #5197, opened 2026-08-19
feat(internal): lab readiness audit tooling + CI a11y/RTL gate for lab, charts, richtext

ernestt:
## Summary

Lab components have been scored against accessibility and RTL criteria that could never run on them. Two independent filters excluded them, and because both failed silently the audits reported "no findings" rather than "never ran".

**The accessibility gate.** `check-components` diffed only `packages/core/src/`, so a lab-only PR resolved to no component changes and skipped `pr-a11y` entirely. `analyze-pr.js` has understood `packages/lab` for a while, so the gate and the analyzer disagreed — the gate said "nothing to audit" while the analyzer was ready to name the components. Widened to `packages/lab/src/`, and the exclude list now matches `analyze-pr.js`'s (it was missing `i18n`).

**The RTL sweep.** Auto-discovery walked only `core-*` story ids, and `componentFromId` stripped only a `core-` prefix — so `lab-listinput--tag-options` was both undiscoverable and, had it been discovered, unmatchable against a `--filter ListInput`. Both now read from one `AUDITED_STORY_PREFIXES` list.

**Readiness tooling.** Adds `internal/lab-readiness`, which derives a component's graduation score from the repo rather than trusting a hand-written report. Checks a machine can prove — exports, stories, tests, docs, token usage, and whether the component is reachable by the two gates above — are computed from the source tree. The rest (research, spec, review, human sign-off) are declared in a manifest with evidence links. A derived result always overrides a declared one, and a passing claim with no evidence link is demoted, so a stale manifest cannot inflate a score.

The generated report is gitignored on purpose. The previous report was produced from an unlanded working tree and then read as current for ten days; `pnpm lab:readiness:check` runs in CI so the manifest is verified instead of the snapshot being trusted.

## Current scores

Reproducible with `pnpm lab:readiness`:

- **List Input** — 21/30. Research and Spec complete; open items are a raw SVG that bypasses `Icon`, and missing disabled/loading/theme stories.
- **Transfer List** — 14/30. RFC #3281 is still open and its proposed listbox semantics contradict the shipped tests. It also has no story titled `Lab/TransferList` (only `Lab/TransferListSelector`), which is exactly why the axe job previously "requested TransferList and audited nothing".

## Test plan

- [x] `pnpm lab:readiness:check` passes; scores reproduce the hand audit
- [x] 17 new tests in `internal/lab-readiness/audit.test.mjs`, including regression guards that fail if either gate stops covering lab
- [x] RTL audit now discovers lab stories — `--filter ListInput` resolves 1 D1 and 7 D5 targets against a built Storybook (previously 0)
- [x] `pnpm check:repo` and the copyright check pass
- [x] `vitest --project node`: 3351 passed. The 2 failures in `hook-discovery.test.mjs` are pre-existing on `origin/main` (case-insensitive filesystem artifact, reproduced with this branch stashed)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,8 +70,14 @@ jobs:
       - name: Generate and test docsite data
         run: pnpm -F @astryxdesign/docsite generate && pnpm -F @astryxdesign/docsite test
 
-  # Lightweight check — does the PR touch any core components?
-  # Gates pr-a11y to avoid unnecessary work
+  # Lightweight check — does the PR touch any published components?
+  # Gates pr-a11y and pr-rtl to avoid unnecessary work.
+  #
+  # Keep the roots and the exclude list in sync with analyze-pr.js: this job
+  # only decides *whether* the downstream audits run, while analyze-pr.js
+  # decides *which* components they audit. When the two disagree the audits
+  # silently no-op — lab components were invisible to both audits until this
+  # job learned about packages/lab/src.
   check-components:
     needs: [check-scope]
     if: github.event_name == 'pull_request' && needs.check-scope.outputs.docsite_only != 'true'
@@ -101,7 +107,7 @@ jobs:
             echo "has_components=true" >> $GITHUB_OUTPUT
             exit 0
           fi
-          CHANGED=$(git diff --name-only origin/${{ github.base_ref }}...HEAD -- packages/core/src/ | grep -v -E '(hooks|theme|utils|__tests__)/' | head -1)
+          CHANGED=$(git diff --name-only origin/${{ github.base_ref }}...HEAD -- packages/core/src/ packages/lab/src/ | grep -v -E '(hooks|theme|utils|i18n|__tests__)/' | head -1)
           echo "has_components=$( [ -n "$CHANGED" ] && echo true || echo false )" >> $GITHUB_OUTPUT
 
   # Run tests (parallel with build)
@@ -136,6 +142,12 @@ jobs:
       - name: Check token docs are in sync
         if: needs.check-scope.outputs.docsite_only != 'true'
         run: node scripts/generate-token-docs.mjs --check
+
+      # Fails when the lab readiness manifest claims a check the source tree
+      # contradicts — the report is derived, never hand-maintained.
+      - name: Check lab readiness manifest is current
+        if: needs.check-scope.outputs.docsite_only != 'true'
+        run: pnpm lab:readiness:check
 
       - run: pnpm test
         if: needs.check-scope.outputs.docsite_only != 'true'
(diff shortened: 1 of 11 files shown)


.github/workflows/ci.yml
```diff
@@ -101,7 +107,7 @@ jobs:
             echo "has_components=true" >> $GITHUB_OUTPUT
             exit 0
           fi
-          CHANGED=$(git diff --name-only origin/${{ github.base_ref }}...HEAD -- packages/core/src/ | grep -v -E '(hooks|theme|utils|__tests__)/' | head -1)
+          CHANGED=$(git diff --name-only origin/${{ github.base_ref }}...HEAD -- packages/core/src/ packages/lab/src/ | grep -v -E '(hooks|theme|utils|i18n|__tests__)/' | head -1)
```
cixzhang, line 110: charts and richtext still skip both audits; `analyze-pr.js` knows all four packages.

.github/workflows/ci.yml
```diff
@@ -137,6 +143,12 @@ jobs:
         if: needs.check-scope.outputs.docsite_only != 'true'
         run: node scripts/generate-token-docs.mjs --check
 
+      # Fails when the lab readiness manifest claims a check the source tree
+      # contradicts — the report is derived, never hand-maintained.
+      - name: Check lab readiness manifest is current
+        if: needs.check-scope.outputs.docsite_only != 'true'
+        run: pnpm lab:readiness:check
```
cixzhang, line 150: Can never fail — no check key is both declared and derived.

internal/lab-readiness/audit.mjs
```diff
@@ -0,0 +1,363 @@
+#!/usr/bin/env node
```
cixzhang, line 1: 1400 lines of tooling on a required job isn't a `ci:` change.

cixzhang (approved):
Thanks, this is the right widening and I ran it — one real new axe hit on Schedule, two RTL. But `lab:readiness:check` can never fail; worth fixing before merge.

[Reviewed by Robohands]

.github/workflows/ci.yml
ernestt: Right — `check-scope` only gates on `packages/core/src/` and `packages/lab/src/`. I'll extend the diff path list to include `packages/charts/src/` and `packages/richtext/src/` so they trigger the pr-a11y and pr-rtl jobs too.

.github/workflows/ci.yml
ernestt: You're right — the contradiction predicate fires only when a check is declared as passing in the manifest AND derived from the source as failing. But in practice derived checks are never simultaneously declared, so `--check` is a vacuously true gate today. I'll add a real assertion: if the derived state flips to passing but the manifest still says `not_started`/`in_progress`, that's also a contradiction (manifest is stale in the positive direction). That way the check can actually fail and catch drift.

internal/lab-readiness/audit.mjs
ernestt: Fair — this is really `feat(internal): lab readiness audit tooling + CI integration`. The CI yaml change is the smaller half. I'll retitle the PR to reflect both pieces.

PR #5199, opened 2026-08-19
OverflowList: report the collapsed set with onOverflowChange

cixzhang:
## Why

`OverflowList` assumes it owns the overflow affordance: `overflowRenderer` mounts a node the list renders *inside itself*, and only while items are overflowing. That is the right shape for a row whose only "more" affordance is the one the list grows.

It is the wrong shape for a row that **already has a standing menu** — a header whose trailing corner carries a "…" with its own entries, a toolbar with a persistent More button. The honest UI there is one menu: as the row narrows, the actions that no longer fit should join the menu that is already on screen. With `overflowRenderer` the user gets a second anchor that pops into existence beside the first one.

Watching from the outside is not reachable today either, and it fails in a way that is easy to get wrong:

- The renderer is called for the hidden measurement copy with **every** item, so a reporter component placed inside it mounts twice and one of those mounts reports the whole list, not the collapsed set. In a narrowed row with items `A B C` collapsed to `A`, a reporter sees `[0,1,2]` and `[1,2]` and cannot tell which is real without sniffing the list's private DOM (`closest('[inert]')`).
- Nothing fires when the row **widens back out**: the renderer simply stops being called, so a hoisted set never empties.

There is a second correctness gap on that path: `useOverflow` only re-measures when the available container, options, or item count changes. If a child's size changes at the same count, the row can keep the old fit, clip later items, and omit its overflow affordance. `onOverflowChange` cannot truthfully expose the collapsed set unless the measurement owner notices size changes first. Keyed membership and order changes are handled in `OverflowList` before the callback reports.

The escape hatch is to drop `OverflowList` and drive `useOverflow` directly — which works, and is what consumers have been doing, but it means re-implementing the hidden measurement container, the gap-token-to-pixel mapping, the visible/collapsed slicing and the `collapseFrom` handling, and keeping all of that in step with the component. That is a lot of surface to copy to move a menu.

## What

`onOverflowChange?: (overflowItems: OverflowItem[]) => void` — called whenever the collapsed set changes. The caller owns the anchor; the list adds nothing to the row.

The contract:

- It fires once measurement has collapsed something, and again with an empty array once the row widens back out and everything fits.
- It is **silent while nothing overflows**, including on mount — a list that fits from the start never calls it. There is no report of the pre-measurement state, which is an empty set whether or not the row actually overflows. Hold the collapsed set in state initialised to `[]` and it is correct at every moment; if you remount the list while keeping that state, reset it yourself.
- Reports are keyed on the collapsed items' original indices and stable React keys. Membership and order changes report even when the count stays the same; unrelated re-renders and callback identity changes do not re-fire it.

```tsx
const [hidden, setHidden] = useState<OverflowItem[]>([]);

<div className="row">
  <OverflowList onOverflowChange={setHidden}>
    {actions.map(a => <Button key={a} label={a} size="sm" />)}
  </OverflowList>
  <DropdownMenu
    button={{label: 'More', variant: 'ghost'}}
    items={[...hidden.map(({index}) => actions[index]), ...sessionActions]}
  />
</div>
```

It reports from a layout effect after measurement, so the menu updates in the same frame the items collapse. It composes with `overflowRenderer` rather than replacing it: both, either, or neither.

## API choice, and what I passed over

The system's line is "the hook is the primitive; the component is sugar", and `useOverflow` is already public — so the first question was whether this should be documentation rather than API. I don't think so: the ask is not a new behavior, it is *the information the component has already computed*. Making every consumer with a pre-existing menu re-implement the component to read it inverts the sugar.

Considered and rejected:

- **A second render slot** (`anchorRenderer`, or an "always render" mode for `overflowRenderer`). The collapsed set is data, not UI: the anchor already exists, is owned by the caller, and usually mixes these items with its own. A render prop would have to render *into* the list, which is the thing we are trying not to do.
- **Always calling `overflowRenderer`, with an empty array when nothing overflows.** Cleaner in the abstract, but it is a silent breaking change for every existing caller — today they never have to guard, and after it they would all render a `+0`.
- **Documenting `useOverflow` as the sanctioned path.** No new API, but it asks consumers to re-derive measurement scaffolding, and it leaves the "reporter inside the renderer" trap in place for whoever tries the obvious thing first.
- **Exposing the collapsed set through context.** Invisible in the type, and it only helps descendants — the standing anchor is typically a *sibling*.

Naming follows `on{Verb}{Scope}` (`onOpenChange`, `onCollapsedChange`, `onWidthChange`). It is a report, not a controlled value: there is no `overflowItems` prop to pair with it, and the list stays the owner of the measurement.

## Risk

The callback is additive and opt-in. The measurement fix also corrects existing lists when the measurement container changes size without an available-width change. It adds one O(N) measurement pass for that real size change and no per-child or mutation observer.

- **Silence is a state.** Nothing fires while the collapsed set is empty and stays empty, so the first report you get is never `[]` — you cannot use a first fire to learn "nothing is hidden". A consumer's own state starts empty, which is that answer already.
- **Dynamic children need stable React keys.** Same-count membership and order changes re-fire when those keys change. Identify items by their current `index` and look up your own data with it, rather than storing the `child` you were handed.
- **Measured-child size changes re-measure before they report.** `useOverflow` observes the hidden measurement container, so a stable-key size change refreshes every path — bare lists, `overflowRenderer`, direct hook consumers, and `onOverflowChange` — even when available width does not change. Keyed membership and order changes re-attach measurement before the callback reports.
- **Do not change the list's children from the callback** — that re-measures, which re-reports. The guard suppresses same-set re-fires, so a stable render loop settles, but an alternating one would not.

Items are reported with their original `index`; that is the stable identity to look up your own data by, rather than storing `child`.

## Testing

- 15 new unit tests (34 in the file): initial all-fit silence; measured initial overflow; collapsed indices; stable-key child-size changes; unequal-width reorder with the callback; membership and order changes; child-count changes reporting only the measured set; all-fit reset after a resize; unchanged sets with a fresh inline callback; replacement callbacks receiving the next change; a state-storing callback settling without a render loop; unmount cleanup; `collapseFrom="start"`; no added anchor or measured width; coexistence with `overflowRenderer`. The stable-key and callback-reorder regression tests fail against their parent runtimes and pass on this head.
- Focused suites green (94/94), along with core and docs typechecking, strict lint on the changed runtime/tests/story, formatting, and the repository pre-commit checks. Full CI is rerunning on the latest-main rebased head.
- **Chromium, fire sequence** (five buttons, fresh mount at each width, `[...]` = one call):

  | scenario | fires |
  |---|---|
  | mount where everything fits | (silent) |
  | mount where 1 item overflows | `[4]` |
  | mount where 3 items overflow | `[2,3,4]` |
  | mount narrow → widen → narrow | `[4]` \| `[]` \| `[4]` |
  | two forced re-renders, no size change | (nothing beyond the mount report) |

  Before this branch's last commit each of those mount rows carried a leading `[]`, so an overflowing list reported "nothing collapsed" first and corrected itself in the same frame.

- **Chromium, real resize** (new `StandingAnchor` story, viewport driven 900 → 460 → 360 → 260 → 900):

  | viewport | row width | hidden | items in the row | anchors on screen |
  |---|---|---|---|---|
  | 900 | 520px | 0 | Rerun Branch Share Export Duplicate | 1 |
  | 460 | 428px | 1 | Rerun Branch Share Export | 1 |
  | 360 | 328px | 2 | Rerun Branch Share | 1 |
  | 260 | 228px | 4 | Rerun | 1 |
  | back to 900 | 520px | 0 | Rerun Branch Share Export Duplicate | 1 |

  The menu's contents at each end: `["Rename session","Delete session"]` wide → `["Branch","Share","Export","Duplicate","Rename session","Delete session"]` at 260px → back to `["Rename session","Delete session"]`. The anchor keeps its identity across the whole sweep (one "More" button at every width, never remounted), and its x position moves only with the row's own reflow.


cixzhang:
Direction: keep `onOverflowChange`. OverflowList should own measurement while exposing the collapsed set so an existing standing menu can own rendering without a second anchor.

cixzhang:
Thanks — this works: one "More" at every width, and the collapsed set tracks. Approve once you undraft. Two nits: the new observer only sees width, and mount measures once more.

[Reviewed by Robohands]

cixzhang:
Thanks — re-checked at `ce5f832`: one "More" at every width, and the collapsed set tracks and empties. Still an approve once you undraft. The two earlier nits remain non-blocking.

[Reviewed by Robohands]

cixzhang:
/accept-visual 33058576861/2 Exact-head rerun has 0 changed, removed, or failed frames; all 22 additions are OverflowList light/dark baselines, including the reviewed standing-anchor story.

cixzhang (commented):
Re-checked at `ce5f832`: the five observer commits net to zero runtime change — `useOverflow.ts` and `OverflowList.tsx` are byte-identical to this morning's head. Both nits stand; approve once you undraft.

[Reviewed by Robohands]

PR #5200, opened 2026-08-19
feat(lab): harden ListInput to the states its props already promise

ernestt:
Follows #5197, which has landed. Rebased onto it, so this diff is now just the ListInput work.

## Why

`ListInput` advertises `isDisabled`, `isLoading`, and a `status` prop covering warning/error/success. Only the error path had a story or a test. The states worked, but they existed in the implementation and nowhere in the evidence: a regression in any of them would have shipped unnoticed, and a reviewer asked to vote on graduation had nothing to look at.

The readiness audit scored this as `stateCoverage` failing — advertised state props with no story and no test behind them.

## What

Six stories and nine tests, all covering behaviour that already existed:

| Story | Pins |
| --- | --- |
| Disabled | Fields and row controls locked, Add unavailable |
| Loading | List marked busy, controls locked, in-flight reorder cancelled |
| Validation Statuses | Warning and success described from the group |
| Stress | 50 records, reorder and remove still responsive |
| Theme Matrix | Light and dark, side by side |

The tests additionally pin `Home`/`End` during an active pickup (jump to first/last) and focus restoration after a removal — focus moves to the next record, or to Add when the list empties.

Also included is a fix to the Remove control: when the list is disabled, Remove keeps `aria-disabled` rather than the native attribute so it stays focusable, and now carries a tooltip explaining why it is unavailable. Reorder, which has no tooltip, stays natively `disabled`. The tests encode that asymmetry rather than papering over it.

Theme Matrix paints its pane with the `--color-background-body` token rather than a literal. A hardcoded colour sits under text that follows the theme and fails contrast in one of the two modes; axe caught exactly that during this work.

## Test plan

- [x] `pnpm vitest run --project ui packages/lab/src/ListInput` — 27 passed
- [x] axe over all 12 ListInput stories — 0 violations, 0 baselined (nothing tolerated, genuinely clean)
- [x] RTL sweep over all 12 stories — 0 failures
- [x] `pnpm -F @astryxdesign/storybook typecheck` — clean
- [x] `node internal/lab-readiness/audit.mjs --candidate list-input --check` — 26/30, manifest agrees with source tree

The axe and RTL runs above are only possible because of #5197. Before it, both audits silently skipped every `packages/lab` component — this PR is the first time they have real coverage in CI.

## What's left

The remaining 4 of 30 are the Harden-review sign-offs — `visualQuality`, `compositionQuality`, `scopeBoundary`, `archivedReview`. Those are non-automatable by design and need a human vote; the stories in this PR are what that vote looks at.

### DIFF
--- a/apps/storybook/stories/ListInput.stories.tsx
+++ b/apps/storybook/stories/ListInput.stories.tsx
@@ -2,8 +2,8 @@
 
 /**
  * @file ListInput.stories.tsx
- * @input Uses ListInput with controlled data, Astryx field controls, column width helpers, React, and StyleX
- * @output Storybook examples spanning realistic use cases, validation scopes, density, responsiveness, tokenized collection motion, and reordering
+ * @input Uses ListInput with controlled data, Astryx field controls, column width helpers, the Theme provider, React, and StyleX
+ * @output Storybook examples spanning realistic use cases, validation scopes, density, responsiveness, tokenized collection motion, reordering, the disabled and loading states, row volume, and a pinned light/dark theme matrix
  * @position Lab component stories; documents the consumer-facing ListInput API
  */
 
@@ -15,7 +15,11 @@ import {DateInput} from '@astryxdesign/core/DateInput';
 import {NumberInput} from '@astryxdesign/core/NumberInput';
 import {Selector} from '@astryxdesign/core/Selector';
 import {pixel, proportional} from '@astryxdesign/core/Table';
+import {Text} from '@astryxdesign/core/Text';
 import {TextInput} from '@astryxdesign/core/TextInput';
+import {Theme} from '@astryxdesign/core';
+import {colorVars} from '@astryxdesign/core/theme/tokens.stylex';
+import {neutralTheme} from '@astryxdesign/theme-neutral';
 import {ListInput, type ListInputColumn} from '@astryxdesign/lab';
 const storyStyles = stylex.create({
   canvas: {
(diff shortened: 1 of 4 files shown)


apps/storybook/stories/ListInput.stories.tsx
```diff
@@ -905,3 +925,168 @@ export const ResponsiveItinerary: Story = {
...
+    docs: {
+      description: {
+        story:
+          'The other stories only exercise `error`. `warning` and `success` use the same list-level slot and are described by the group, so a screen reader reaches them from any field inside the list.',
+      },
+    },
+  },
+};
+
+const STRESS_TAGS: TagOption[] = Array.from({length: 25}, (_, index) => ({
```
cixzhang, line 1017: Body says 50 records and six stories; it's 25 and five.

cixzhang (commented):
Thanks, good pins — I walked each state in Chromium and they all hold.

Heads up that stacked on #5197 this gets no CI: only CLA and Vercel ran, so `test`, `lint`, `build` and the audits never fired. Nothing here depends on #5197, so rebasing onto main would get it checked. Your call — it's lab-only, merge when you like.

Also `test(lab)` rather than `feat` — no source changed.

[Reviewed by Robohands]

packages/lab/src/ListInput/ListInput.test.tsx
```diff
@@ -1123,6 +1123,223 @@ describe('ListInput', () => {
...
+    expect(group).not.toHaveAttribute('aria-busy');
+
+    for (const field of screen.getAllByRole('textbox')) {
+      expect(field).toBeDisabled();
+    }
+    expect(screen.getByRole('button', {name: /add guest/i})).toBeDisabled();
+    expect(
+      screen.getByRole('button', {name: 'Reorder guest 1'}),
+    ).toBeDisabled();
+    // Remove carries a tooltip, so it is disabled via aria and stays focusable
```
cixzhang, line 1141: Tooltip does open when disabled, but it only repeats the button's name.

apps/storybook/stories/ListInput.stories.tsx
ernestt: Good catch — fixed the PR body: five stories, 25 records.

packages/lab/src/ListInput/ListInput.test.tsx
ernestt: Fair point — a tooltip that just echoes the accessible name adds no new information. I'll update the disabled-state tooltip to surface the reason (e.g. "Remove is unavailable while the list is disabled") so it actually earns its presence. Will push with the next round.

ernestt (commented):
Thanks for the walkthrough. Agreed on both:
1. Will rebase onto main so the full CI suite runs (lint/test/build/audits).
2. `test(lab)` is the right prefix — no source changed, only test/story coverage. Will amend before merge.

PR #5080, opened 2026-08-16
feat(core): promote Bottom Sheet from Lab

imdreamrunner:
## Depends on

- #5078 — prepares the existing Lab component and documentation before promotion

## Summary

- move `BottomSheet` and `BottomSheetSwitcher` from Lab to Core without changing their runtime behavior
- add Core root and subpath exports, and remove the Lab exports without a compatibility re-export or codemod
- move Storybook coverage to Core and add Core CLI-backed documentation examples
- add `overlayControl` support so the Switcher Properties preview can open through `activeSheet`
- add the Core/CLI changeset for the promotion and documentation assets

This PR is intentionally stacked on #5078 and targets its branch. After #5078 merges, this PR can be retargeted to `main`.

## Validation

- `pnpm exec vitest run packages/core/src/BottomSheet` — 113 tests passed
- `pnpm --dir apps/docsite test` — 377 tests passed
- focused ESLint passed
- Core and Lab production builds passed
- Storybook typecheck passed
- docsite generation and typecheck passed
- CLI authoring and template-doc typechecks passed
- Core source and docs typechecks passed
- `pnpm check:repo`
- `pnpm sync:exports:check`
- `pnpm verify-exports` — all 10 packages and 32 ESM targets passed

Issue #4897, opened 2026-08-11
[audit] Stepper: Stepper uses Children.count(children) (banned child introspection); a Fragment around the steps collapses the count to 1 and breaks the on-track last-connector logic.


ernestt:
Found by a component audit of **Stepper**, mode O, rubric **v1.2.1**, at commit `9b5eaa05281`.

- **Rule:** `P14` — see the [Component Audit Rubric](https://github.com/facebook/astryx/wiki/Component-Audit-Rubric).
- **Ledger row:** [Stepper on the component scores page](https://facebook.github.io/astryx/sandbox/pages/component-scores/) — recorded C 71.6 with 4 open BLOCK(s).
- **Finding:** Stepper uses Children.count(children) (banned child introspection); a Fragment around the steps collapses the count to 1 and breaks the on-track last-connector logic.

### Closing protocol

Resolving this requires:

1. Re-run the audit for Stepper after the fix — the whole component, not the diff (the rubric's "Grading a whole component" section).
2. Put the visual results in the PR description — before/after screenshots of the affected states. The rendered-design section is graded from those.
3. Update the ledger with the new score and clear this BLOCK (`<scorecard> | node scripts/score-ledger.mjs --record Stepper --from - --push`).
4. Close this issue linking that PR.

If the recorded score doesn't move, either the fix or the audit is wrong.

ernestt:
Addressed in #4917 (open). Re-audit recorded: Stepper C 71.6 → B 88.3, 0 open BLOCKs. This finding will close when that PR merges.

HelloOjasMutreja:
Already fixed on main by #4917: the on-track connector's leading/trailing-node hiding is now driven by each step's own `<li>` position via a StyleX marker + `:first-child`/`:last-child` (`otSegHiddenIfFirst`/`otSegHiddenIfLast` in `Step.tsx`), with no `Children.count()` or any child introspection left in the file.

cixzhang:
Fixed on `main` by [#4917](https://github.com/facebook/astryx/pull/4917) (merged 2026-08-11) — `Children.count` is gone, replaced by a marker plus `:first-child`/`:last-child`, so a Fragment around the steps no longer collapses the connector logic.

Verified against the pre-fix code: fragment-grouped on-track orphaned the first node and dangled the track past the last; it now renders identically to flat. The one real gap was that nothing guarded it, so [#5052](https://github.com/facebook/astryx/pull/5052) adds the regression test (it fails against pre-#4917 code).

[...]

PR #5201, opened 2026-08-19
feat(core): promote Stepper from Lab

ernestt:
## Summary

Promote `Stepper` and `Step` from the canary-only `@astryxdesign/lab` package to stable `@astryxdesign/core`, following the same mechanics as the Bottom Sheet promotion (#5080).

- **Real move**, not a flag flip: the component directory moves `packages/lab/src/Stepper/` → `packages/core/src/Stepper/`; cross-package `@astryxdesign/core/*` imports become relative (`../theme`, `../utils`, `../i18n`, `../Icon`, `../VisuallyHidden`, `../BaseProps`).
- Add the Core root export (`packages/core/src/index.ts`) and the `./Stepper` subpath export (`packages/core/package.json`); remove the Lab export.
- Move Storybook coverage to `Core/Stepper` and repoint the CLI-backed showcase + example blocks to `@astryxdesign/core/Stepper`, so the docsite now renders their **live previews** (lab is not a docsite dependency, so they were previously code-only).
- No runtime behavior change to what already shipped in lab. Three promotion-time hardening fixes were required to clear the Core bar (see below).

`packages/lab/package.json` keeps `"private": true` + `"astryx": { "canaryOnly": true }` — the publishing guarantee is untouched.

## Why Stepper is ready (Component Lifecycle gate)

Per the [Component Lifecycle](https://github.com/facebook/astryx/wiki/Component-Lifecycle#lab--core-component-graduation), a `lab → core` graduation is gated on a full grading against the [Component Audit Rubric](https://github.com/facebook/astryx/wiki/Component-Audit-Rubric#promoting-to-core): **0 open BLOCKs, overall ≥ B, and every Accessibility Checklist item verified in this PR.**

Stepper was last graded **88.3 / B (0 BLOCKs)** on 2026-08-11, but under rubric **v1.2.1**. This PR re-grades it under the current **v1.5** and moves it into core clean.

### Promotion-time fixes (drift that only the core gate catches)

The core package runs lint rules that never ran on lab. Moving the files into `core/` surfaced — and this PR fixes — three `lint:strict` BLOCKers (Q1 gate) and a batch of should-fix smells:

| Fix | Rule | Sev | What changed |
|---|---|---|---|
| Localized default `aria-label` | §9 I1 / §1 A16 `@astryx/no-hardcoded-i18n-string` | BLOCK | `label = 'Progress'` (a hardcoded destructure default) → `label ?? t('@astryx.stepper.label')`, new catalog key added to `en.json`. Mirrors `Breadcrumbs`. |
| Tokenized `lineHeight` | §2 `@astryx/no-hardcoded-styles` | BLOCK ×2 | description `lineHeight: '16px'` → `spacingVars['--spacing-4']` (identical 16px); number-badge `lineHeight: 1` removed (the digit is grid-centered via `placeItems`, matching Avatar/AvatarGroup). |
| `import type` | `@typescript-eslint/no-import-type-side-effects` | BLOCK | `import {type ReactNode}` → `import type {ReactNode}`. |
| Render guards | §7 `@astryx/no-nullish-jsx-guard` | FIX ×5 | `description != null` / `children != null` / `statusText != null` JSX guards → `isRenderable(...)` so an empty-string `description` or a `false` child no longer leaks an empty wrapper. |
| Docs completeness | §8 X2/X5 (docsite `data-extraction` completeness test) | BLOCK | `Step.doc.mjs` converted from a thin empty standalone into a proper `subComponentOf: 'Stepper'` doc carrying Step's description + 11 props; the `Step` entry in `Stepper.doc.mjs` becomes a name-only cross-ref (mirrors `Breadcrumbs`/`BreadcrumbItem`). |

## Component Audit — rubric v1.5 (promotion mode)

```
Component: Stepper            Package: core (promoted from lab)
Mode: promotion               Rubric: v1.5      Date: 2026-08-19
Commit/PR: this PR

Section                                  W    Score  Weighted
§1  Accessibility & operable paths       16    4.0/5   12.80
§2  Theming & token integrity            14    4.5/5   12.60
§3  Public API contract                  14    4.0/5   11.20
§4  Behavior correctness & states        12    5.0/5   12.00
§5a Design conventions — objective        6    5.0/5    6.00
§5b Design conventions — rendered         4    4.5/5    3.60
§6  Testing & verification                8    4.5/5    7.20
§7  Code health (React & DOM)             8    5.0/5    8.00
§8  Docs, Storybook & docsite             8    5.0/5    8.00
§9  i18n & RTL                            5    4.5/5    4.50
§10 Responsive & touch                    5    4.0/5    4.00
                                        ---           ------
TOTAL                                   100            89.90   →  GRADE: B

BLOCKers open: 0
SENSITIVITY: 0 open BLOCKs — nothing to fix to hold the grade.
```

Movement vs the 88.3 (v1.2.1) baseline: §7 4.5→5.0 (nullish-guard smells removed via `isRenderable`), §8 4.5→5.0 (the lab docsite-render gap is closed now that blocks resolve against core). §9's new v1.5 BLOCK-capable checks (I19 IME guard, I20 grapheme-safe counting, I21 provider-locale) are **N/A** for Stepper: it has no editable/command-key surface, does no string truncation, and does no locale-sensitive formatting (the step number uses ICU `{stepNumber, number}`, which follows the provider locale).

### §11 lifecycle evidence

- **L1 ✅** lab `package.json` keeps `private: true` + `astryx.canaryOnly: true`.
- **L2 ✅** real move (git renames lab→core, deletes the lab entry).
- **L13 ✅** post-move surface complete in core: `Stepper.tsx`, `Step.tsx`, `.test.tsx`, `Stepper.doc.mjs` + `Step.doc.mjs`, story in `apps/storybook/stories/`, `index.ts`, blocks (1 hero showcase + 6 examples).
- **L14 ✅** changeset present (`[component]`, `@handle`), patch bumps.
- **L15 ✅** no Meta-internal leaks in the diff or this body.
- **L4** CI + review gate — pending on this PR.
- Spec/arbitration/vibe evidence (L5–L8): no per-component artifacts exist (system-level ledgers only) — flagged, not blocking.

## Accessibility verification (promotion gate — every item)

- **Name ✅** — the `<ol>` carries an accessible `label` (localized default `@astryx.stepper.label`); every clickable step's `<button>` gets a composed accessible name (`Go to step N: <label>[, <status>]`).
- **Role ✅** — ordered-list progress pattern (`<ol>`/`<li>`), not a `nav` landmark; the active step is `aria-current="step"`. Follows the [WAI-ARIA APG progress/step sequence guidance](https://www.w3.org/WAI/ARIA/apg/) rather than polluting the landmark map.
- **State ✅** — progress + semantic status reach AT as text, not color alone: indicator glyphs are `aria-hidden`, and a `VisuallyHidden` "completed/warning/error" mirror sits next to the label (and is composed into clickable steps' names). `aria-current` marks the active step.
- **Keyboard ✅** — clickable steps are native `<button>`s: Enter/Space activate, Tab order follows document order and skips disabled steps. Asserted in tests.
- **Focus ✅** — `focusOutlineStyles.focusVisible` on every interactive variant; the ring lives on the button (the owner), visible in every theme. Stepper traps no focus (not a modal), so focus-trap is N/A.
- **Announcements ✅** — no born-with-content live region; status is a declarative `VisuallyHidden` mirror, not a hand-wired `aria-live`.
- **Reduced motion ✅** — the only transition (clickable-step background) collapses to `0s` under `@media (prefers-reduced-motion: reduce)`.
- **Forced colors ⚠️ (FIX, not blocking)** — indicator glyphs paint via `currentColor` and connectors use token backgrounds, so they degrade acceptably under `forced-colors: active` (the OS maps `currentColor`), but there is no explicit `forced-colors` block. Tracked as a should-fix, consistent with most core components that rely on `currentColor` + tokens.
- **i18n ✅** — every visible and AT-facing string is a `@astryx.*` key resolved through `useTranslator()`; ICU interpolation for the step number; keys present in `en.json` with descriptions.
- **Target size ✅** — interactive steps expand their hit area via density padding; coarse-pointer targets measured ≥24px in the prior hardening pass.

Two pre-existing `color-contrast` entries remain in `.github/a11y-baseline.json` (Disabled-Steps: WCAG 1.4.3-exempt disabled text; endContent trailing `Badge`: a shared-Badge-token contrast in that story composition). Neither is introduced by this PR and neither is a Stepper defect.

## Visual evidence (§5b — required for promotion)

Screenshots across states × light/dark × a custom theme, plus a 320px narrow viewport and a coarse-pointer capture:

Captured with real Chromium (retina, `deviceScaleFactor: 2`) against Storybook. Full set: 10 stories × light/dark on the neutral theme, 3 key stories × light/dark on the **stone** custom theme, plus a 320px narrow viewport and a coarse-pointer/mobile capture. All 28 rendered correctly — no broken states.

<details><summary><b>Neutral theme — light / dark</b> (click to expand)</summary>

| State | Light | Dark |
|---|---|---|
| Default (vertical) |  |  |
| Default (horizontal) |  |  |
| Status — all semantic colors |  |  |
| Indicator modes (auto/number/icon/none) |  |  |
| Disabled steps |  |  |
| On-track (vertical) |  |  |
| On-track (horizontal) |  |  |
| With content slot (multi-step form) |  |  |
| Collapsible labels |  |  |
| Long labels + descriptions |  |  |

</details>

<details><summary><b>Stone custom theme — light / dark</b> (proves theming holds)</summary>

| State | Light | Dark |
|---|---|---|
| Default (vertical) |  |  |
| Status — all semantic colors |  |  |
| On-track (horizontal) |  |  |

</details>

<details><summary><b>Narrow viewport (320px) + coarse pointer</b></summary>

| 320px reflow (labels collapse) | Coarse pointer / mobile (390px) |
|---|---|
|  |  |

</details>

## RTL audit (I17 — promotion-gated)

Ran `node apps/storybook/rtl-audit/rtl-audit.mjs --filter stepper` against the built Storybook. Verbatim result:

```
AUTO: 0 pass / 0 fail (0 surprise) / 1 N-A
PM  : 0 pass / 0 fail / 27 N-A (tol 3px)
RTL_EXIT=0
```

- **D1 (icon-mirror): N/A** — Stepper uses only checkmark / number / semantic-status glyphs; nothing directional (no chevron/arrow/caret) to mirror.
- **D5 (positional-mirror): N/A** across all 27 stories — no logical-anchor + physical-transform elements (no absolute/fixed physical positioning).
- **0 failures, 0 surprises.** The component uses CSS logical properties throughout (`marginInline*`, `paddingInline*`, `insetInline*`, `textAlign: 'start'`) and a container query for the narrow-width collapse, so it is direction-agnostic by construction. Nothing gates promotion on RTL.

## Validation

- `pnpm -F @astryxdesign/core test Stepper` — 38 tests passed
- `CI=true eslint packages/core/src/Stepper` — 0 errors, 0 warnings (strict tier)
- `pnpm -F @astryxdesign/core build` and `pnpm -F @astryxdesign/lab build` passed
- `pnpm -F @astryxdesign/core typecheck` + `typecheck:docs` passed
- `pnpm -F @astryxdesign/docsite generate && test` — 377 tests passed (Stepper showcase + 6 examples now render)
- `node scripts/check-i18n-catalog.mjs` — catalog consistent (new `@astryx.stepper.label` described)
- `node scripts/sync-exports.js --check` + `pnpm check:sync` + `pnpm check:changesets` clean
- repo-wide `ASTRYX_STRICT_LINT=1 eslint` — 0 errors


packages/core/src/Stepper/Step.doc.mjs
```diff
@@ -0,0 +1,86 @@
...
+      name: 'children',
+      type: 'ReactNode',
+      description:
+        'Content rendered below the label and description. Useful in vertical steppers for form fields or detailed step content.',
+    },
+    {
+      name: 'icon',
+      type: 'ReactNode',
+      description:
+        'Custom icon rendered inside the indicator. Accepts any node (e.g. an Icon). Takes precedence over the built-in number/check.',
```
cixzhang: "Takes precedence" is the whole problem: `indicator="none"` + `icon` renders nothing at all, silently.

packages/core/src/Stepper/Step.doc.mjs
```diff
@@ -0,0 +1,86 @@
...
+      slotElements: [{__element: 'Icon', props: {icon: 'check', size: 'sm'}}],
+    },
+    {
+      name: 'status',
+      type: "'accent' | 'success' | 'warning' | 'error'",
+      description:
+        'Semantic color for the step. Controls color only and maps to the global Astryx semantic tokens. Leave unset for the progress-derived default coloring.',
+    },
+    {
+      name: 'indicator',
```
cixzhang, line 47: This already takes a ReactNode, so `icon` is a second spelling of the same axis.

cixzhang (changes requested):
Thanks for doing this properly — the move is clean and the evidence is thorough.

One thing I'd like fixed before this API is permanent: `icon` and `indicator` are the same axis and they collide silently. `<Step icon={...} indicator="none" />` renders no indicator at all, and `indicator="number"` with an `icon` drops the number — no warning either way. `indicator` already takes a ReactNode, so I'd drop `icon` and keep one spelling. Once this is in core, removing either one is a breaking change for everyone who picked the wrong spelling, so I'd rather we only ship one.

Nothing catches a duplicate or skipped `step` index either: two steps at `step={1}` both get `aria-current="step"`. Counting children is out (#4897) — what shape would you want, a dev warning or registering through context?

[Reviewed by Robohands]

packages/core/src/Stepper/Step.doc.mjs
ernestt: Agreed — dropped `icon` entirely. `indicator` already accepts ReactNode, so there's no need for a second spelling. The latest push removes `icon` from the interface, doc, stories, and showcase blocks; passing a custom node now goes through `indicator={<Icon ... />}` exclusively.

packages/core/src/Stepper/Step.doc.mjs
ernestt: Exactly — one axis, one prop. `indicator` now carries the full `ReactNode` slot and the doc reflects it (with slotElements for the CLI authoring surface).

ernestt (commented):
Addressed both points in the latest push:

1. **Unified indicator axis** — `icon` prop removed. Custom indicators go through `indicator={<ReactNode>}` exclusively, eliminating the silent collision.

2. **Duplicate step index detection** — went with context registration: each `<Step>` calls `registerStep(step)` on mount (returns a cleanup). The parent Stepper keeps a ref-backed `Map<index, count>` and fires a `console.warn` in dev mode when count > 1. No runtime cost in production, no child counting, and it naturally handles dynamic add/remove of steps.

cixzhang (approved):
Thanks — one indicator axis and a dev warning on duplicate indices is exactly it.

Approving the promotion. One thing for next time: the connector animation is a separate change riding along on this one, and it's the piece most likely to attract debate. Promotion is what's blocking people — motion goes in its own PR on top, so the settled part can land without waiting on the part that isn't.

[Reviewed by Robohands]

cixzhang (approved):
Thanks — `icon` is properly gone from the interface, not just the doc. Two doc catalogs still list it (`Stepper.doc.mjs:222` and `:336`), and `StepperContextValue` isn't exported next to `useStepperContext`. Worth fixing before merge; approving.

[Reviewed by Robohands]

ernestt (commented):
Fixed — removed the two stale `icon` entries from the doc catalogs (propDescriptions and i18n prop list) and exported `StepperContextValue` from the barrel. See ae2814b.

Issue #5160, opened 2026-08-17
[audit] RichTextEditor: 8 open BLOCKs — 62.3/D (rubric v1.4); the package has no lint coverage


cixzhang:
**Mode:** grading (on-demand, full pass) · **Rubric:** [v1.4](https://github.com/facebook/astryx/wiki/Component-Audit-Rubric) · **Date:** 2026-08-17
**Component:** `RichTextEditor` (+ `RichTextView`, `RichTextEditorToolbar`) · **Package:** `@astryxdesign/richtext` (private, canary-only)
**Commit audited:** `6e7f89c16f5`
**Ledger row:** `richtext/RichTextEditor` — recorded in [`component-scores.json`](https://github.com/facebook/astryx/wiki/Component-Audit-Rubric#recording-an-audit)

## Headline

**62.3 / 100 → D**, with **8 open BLOCKs**. Fix the 8 BLOCKs and change nothing else → **projected ~78 / C**.

The editor itself is well built: the WCAG 2.1.2 Tab-escape plugin is the best implementation of that pattern in the repo, the Lexical effects all synchronise a genuinely external system, 86 tests assert the public contract, axe is green, and it reflows cleanly at 320 px. Almost every BLOCK below is a rule that **exists and would have caught this in `core`** — see [Root cause](#root-cause).

## Root cause — this package has no rule coverage

`eslint.config.js` scopes the whole `astryxConfig` (tokens, i18n, `require-base-props`, `no-classname-clobber`, `boolean-prop-naming`, physical properties, react-introspection) to **`packages/core/src/**` only**. `packages/richtext` is linted with the base config alone, so `pnpm lint:strict` is green while the code carries hardcoded strings, inline `<svg>` glyphs, a props type that doesn't extend `BaseProps`, and a className clobber. The two theming guards (`themingTargets.test.ts`, `derivedVarRegistry.test.ts`) are likewise core-only.

**Highest-leverage fix: extend the astryx lint config (and the theming guards) to `packages/{lab,richtext}/src/**`.** Every BLOCK except #2 and #3 is auto-detectable once that lands.

### 1. §9 I1/I2/I4/I5 — no string in the package is translated
Every visible and AT-facing string is an English literal, and none exists in `packages/core/locales/en.json`:
[...]

PR #5204, opened 2026-08-19
fix(richtext): close all 8 RichTextEditor audit BLOCKs

AKnassa:
Part of #5160.

## What this does

Closes all 8 open BLOCKs from the RichTextEditor audit, plus the audit's root cause: the richtext package had no astryx lint coverage, so none of the rules that would have caught these problems ever ran there.

## Why

The audit graded the component 62.3/D with 8 BLOCKs. Fixing the BLOCKs alone projects ~78/C, and extending the lint rules keeps the class of problem from coming back.

## What changed

- **Root cause.** The astryx eslint blocks (tokens, i18n, base props, boolean naming, react rules, type-aware TS) now cover `packages/richtext/src`, and everything they flagged is fixed. A new flat-package theming guard enforces doc-matches-source for `themeProps` sites. `packages/lab` is left for its own effort; it carries an unrelated backlog.
- **BLOCK 1, i18n.** Every visible and AT-facing string routes through `useTranslator`, with 31 new `@astryx.richTextEditor.*` keys in `en.json` (defaults verbatim to the previous English). Counter announcements now pluralize correctly via ICU.
- **BLOCK 2.** `RichTextView` takes a `label` wired to its textbox, dev-warns when omitted, and announces `aria-multiline`.
- **BLOCK 3.** Read-only and disabled are now different states: read-only stays keyboard-reachable at full opacity and announces `aria-readonly`; disabled dims, leaves the tab order, and announces `aria-disabled`. The Lexical editable state also follows prop changes now (`initialConfig.editable` is only read at mount), so the announcements can never contradict actual editability.
- **BLOCK 4.** The core Selector's combobox button collapsed to a ~20px line box inside its 28px container because the container held the block padding; the padding moved onto the button (same pixels, full-height hit area, changeset included). The toolbar's controls also get a 44px floor on coarse pointers, following the Slider precedent. Measured in-browser: the block-format trigger went from 68x20 to 68x28.
- **BLOCK 5.** Theme icon overrides actually reach the toolbar now: `resolveIcon` passes the active theme name to `getExtendedIcon` (previously theme icons were never consulted), and `defineTheme({icons})` accepts colon-namespaced extension keys like `richtext:bold` without casts (changeset included). `TextGlyph` uses StyleX tokens instead of a raw inline style.
- **BLOCK 6.** The doc target declares `visualProps: ['size']` and `states: ['status']`, guarded by the new test.
- **BLOCK 7.** Consumer `className`/`style` merge through `mergeProps` instead of clobbering the view's StyleX classes. Note: the audit's TS2339 half does not reproduce against a built workspace; `RichTextViewProps` already extended `BaseProps` at the audited commit, so only the clobber half was real.
- **BLOCK 8.** Undo/redo glyphs mirror under RTL via the shared `rtlStyles.mirror`, and `rtl-audit` gains a curated D2 target for the toolbar story.
- **FIX 2 and friends.** All 13 RichTextEditor axe baseline entries removed (the audit estimated 10; a fresh run against a built storybook resolved 13, including the Markdown Serializers story's unlabeled textarea and view). Also picked up FIX 3 (heading sizes on the shared scale, h3 lands on the 17px step), FIX 6 (`linkOpensInNewTab` renamed `hasNewTabLinks`), FIX 7 (React 19 `ref` prop instead of `forwardRef`), and FIX 8 (the `format as any` cast removed).

## Verification

- richtext tests 86 to 132; core Selector/Icon/theme suites green; full repo suite 10736/10738, the 2 failures are the pre-existing cli case-sensitivity pair that also fails on a clean `upstream/main` checkout.
- `lint:strict` identical to baseline (56 warnings, 0 errors), all six typecheck gates pass, `pnpm a11y:audit --components RichTextEditor` reports 0 new and 0 baselined.
- Mutation probes on the six load-bearing behaviors each fail their covering test when reverted.

## How to see it

Storybook: `Lab/RichTextEditor` stories `With Toolbar` and `Read Only`.

Toolbar, light LTR:

Undo/redo glyphs mirrored under RTL:

Read-only at full opacity, keyboard-reachable:


gonzoblasco:
Hi @AKnassa, this is a model audit-close-out - fixing the root cause (lint coverage) rather than just the symptoms is exactly the right instinct, and the mutation probes give real confidence the tests guard the load-bearing behaviours. A few observations:

**1. The read-only vs disabled split (BLOCK 3) is exactly the model that should be the norm - worth calling out as the pattern.**
Separating read-only (stays in the tab order, `aria-readonly`, full opacity, keyboard-reachable) from disabled (leaves the tab order, `aria-disabled`, dimmed) is the correct ARIA contract, and the fact that `contenteditable` now follows prop changes post-mount (`initialConfig.editable` only being read at mount was the subtle bug) closes the gap where the announcement could contradict actual editability. That's the kind of detail that makes an a11y fix trustworthy - a SR user is told one thing while the control behaves differently. The tests pinning `contenteditable` to `aria-readonly`/`aria-disabled` consistency are the right guard. This is the model worth propagating to the rest of the library (it's the same distinction some other components blur between busy and disabled).

**2. The hit-target fix (BLOCK 4) correctly distinguishes the two WCAG bars - good.**
Moving the block padding onto the combobox button (the element hit-target audits actually measure) with `align-self: stretch` is the right structural fix, and keeping 24px for fine pointers (2.5.8 AA) while applying 44px only on `@media (pointer: coarse)` respects the fact that 44px is a touch-target best practice, not a universal floor. Worth confirming the 44px floor also applies to the toolbar controls you mention (the body says they follow the Slider precedent) - the toolbar's the place where a keyboard/touch user actually reaches for formatting controls.

**3. The "close all 8 BLOCKs" framing undersells the axe cleanup.**
You note the audit estimated ~10 axe baseline entries and a fresh run resolved 13 (including the Markdown Serializers story's unlabeled textarea). That's a good sign the audit underestimates - but it also means there may be more components in the same state (audited, graded, but with more baseline entries than the audit predicted). Worth a note in the follow-up that the baseline-vs-audit delta should be checked across other audited components, not just richtext.

**4. Small thing: the projected 78/C after closing the BLOCKs leaves a lot of headroom.**
The audit graded 62.3/D and closing the BLOCKs projects ~78/C. That's a big jump but still a C - which is the honest place to park it, since the remaining gap (whatever isn't a BLOCK) may include contrast/perf/robustness items that are a different scale. Just flagging that "closed the BLOCKs" is a milestone, not the finish line - worth an explicit follow-up issue for the path from C to A rather than letting it linger.

Overall: a strong, well-tested close-out. The BLOCK 3 read-only/disabled model and the root-cause lint coverage are the standout pieces, and they're the ones with the most library-wide leverage.

cixzhang (changes requested):
Thanks — the lint root cause and read-only/disabled split are strong. This still bundles about fifteen independent decisions; `__resetDevWarnings` is one test-only public export with no product case.

Two correctness holes remain: `<RichTextView value={json} />` still compiles to an unnamed textbox, and `isRenderable([null, false, ''])` returns true although React renders nothing. This also needs a merge with main.

Could you stack this as (1) required label + read-only/disabled, (2) i18n, (3) icon theming, (4) Selector sizing, and (5) one core utility per PR, with lint coverage last once the package is clean? Start with 1. If you'd rather talk it through with someone, we're in [Discord](https://discord.com/invite/XnsUcFykEP).

[Reviewed by Robohands]

PR #5205, opened 2026-08-19
fix(build): split StyleX layers in production builds, not just dev

cixzhang:
## Why

The Astryx layer splitter only ever ran in a Vite **dev server**. It lived in a `configureServer` hook, so `vite build` never split anything: every atom — library and product alike — shipped inside StyleX's flat `@layer priorityN` blocks.

Those blocks are declared *after* the named layers, so they outrank `astryx-theme`. The layer order we advertise is real in dev and fiction in production, and an installed theme's component overrides are silently dropped in every built app — Storybook previews included.

Spotted by @cixzhang on a PR preview: `.astryx1q8g9m5 { background-color: var(--color-warning) }` — a library atom, correctly prefixed — sitting in a top-level priority layer.

## What

- Split the emitted CSS by **atomic class prefix** instead of by source path, in a small pure `splitStylexLayers()` pass, and run it everywhere StyleX hands its CSS over: the dev middleware, the bundle asset (`generateBundle`), and the file StyleX appends to on disk (`writeBundle`). Dev and production now produce the same layering from the same code.
- Splitting the *output* rather than compiling twice keeps whatever lowering and minification the bundler already applied, and makes the pass idempotent — only top-level `priorityN` layers are rewritten, so a second run is a no-op and other layers (`reset`, app CSS) are untouched.
- Selectors carrying no atom (`:root` token blocks) stay in the library layer, or a theme could not retint tokens.
- `libraryPattern` now feeds the prefixer that the split reads — before, Storybook's `libraryPattern: 'packages/'` and the babel prefixer's own default list disagreed, so `packages/charts` and `packages/richtext` were compiled as *product*. It also accepts an array now.
- Folded the duplicated modern/legacy plugin bodies into shared factories. The duplication is why this stayed invisible.

## Risk

Behavior change, and the point of the PR: library base styles now sit *below* `astryx-theme`, so themed component overrides start applying in built apps that never showed them. Expect Storybook previews of themed stories to shift. Class names for `packages/charts` and `packages/richtext` change prefix (`x…` → `astryx…`).

`postcss` becomes a declared dependency of `@astryxdesign/build` — `src/index.js` already required it undeclared.

## Testing

`splitLayers.test.ts` covers the bucketing rules against real Storybook output shapes: merged selector lists (`.astryx1md70p1, .x1md70p1`), `:root` token blocks, conditional at-rules, prefixed keyframes, priority ordering, idempotency. `vite.build.test.ts` runs a real production `vite build` over a two-file fixture and asserts the top-level layers are exactly `reset, astryx-base, product`.

Verified on a real Storybook build, before and after — same story, same y2k theme, same machine:

| | main | this PR |
|---|---|---|
| top-level layers | `reset, priority1 … priority10` | `reset, astryx-base, product` |
| `.astryx1q8g9m5` | top-level `@layer priority4` | `@layer astryx-base { @layer priority4 }` |
| rules with mixed library/product selectors | 196 | 0 |
| Button, y2k `color` | `rgb(255,255,255)` — base wins | `rgb(204,207,250)` — theme wins |
| Button, y2k `border-top` | `0px none` — override dropped | `1px solid` |

### DIFF
--- a/.changeset/build-split-layers-in-production.md
+++ b/.changeset/build-split-layers-in-production.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/build': patch
+---
+
+[fix] Vite plugin: split Astryx library styles and product styles into their CSS layers during production builds, not only in the dev server. The splitter lived in a `configureServer` hook, so `vite build` shipped every atom in StyleX's flat `@layer priorityN` blocks — which sit outside the named layers and outrank `astryx-theme`, letting library base styles beat an installed theme. Splitting now keys off the atomic class prefix rather than the source path, and runs wherever StyleX hands over its CSS (dev middleware, bundle asset, written file), so dev and production agree. `libraryPattern` also accepts an array and now feeds the prefixer that the split reads, so the two can no longer disagree about which files are library source.
+
+@cixzhang
--- a/packages/build/build.mjs
+++ b/packages/build/build.mjs
@@ -23,6 +23,7 @@ buildSync({
     'vite',
     '@stylexjs/babel-plugin',
     '@stylexjs/unplugin',
+    'postcss',
     'path',
     'url',
     'node:fs',
--- a/packages/build/package.json
+++ b/packages/build/package.json
@@ -48,7 +48,8 @@
   "dependencies": {
     "fast-glob": "^3.3.0",
     "glob-parent": "^6.0.0",
-    "is-glob": "^4.0.0"
+    "is-glob": "^4.0.0",
+    "postcss": "^8.5.0"
   },
   "devDependencies": {
     "esbuild": "^0.28.1"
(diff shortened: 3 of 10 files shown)


cixzhang:
Superseded by [#5410](https://github.com/facebook/astryx/pull/5410), which landed the same fix: `packages/build/src/vite.ts` splits the layers in `writeBundle`/`generateBundle` on a production build, and `postcss` is a declared dependency. #5410 also carries what this branch did not — real `vite build` tests and a Chromium cascade-order gate in CI. Closing rather than rebasing.

One residual, not worth reviving this branch for: `libraryPattern` (vite, a single string) still disagrees with `libraryPatterns` (`babel/index.js`, an array that omits charts and richtext). That is cosmetic now that the split is keyed on source path rather than a prefix, but it is a real inconsistency if anyone reaches for either name.

PR #5209, opened 2026-08-19
fix(core,lab): reset container padding custom properties on overlay elements (#5208)

### LINKED ISSUE #5208: Section inside a BottomSheet overflows the sheet (escapes an ancestor's container padding)
**Bug:** a `Section` inside a `BottomSheet` renders wider than the sheet and hangs off both edges, losing all its padding — whenever any ancestor sets the container-padding vars.

```tsx
<Section padding={4}>          {/* or any padded container, incl. the docsite preview stage */}
  <BottomSheet ...>
    <Section padding={4}>      {/* too wide; content sits flush at the sheet edge */}
```

Measured: sheet 640px, inner Section 672px (304..976 inside a 320..960 sheet). On the docsite's example stage the same block overflows by 12px each side.

**Cause.** `Section` escapes its parent's padding on purpose, via `marginInlineStart: calc(-1 * var(--container-padding-inline-start))` (`Section.tsx`, `nestedStyles.outer`). That is right for a section nested in a page.

But a `BottomSheet` is a fixed-position `<dialog>`. It is a DOM descendant of the padded container, so it inherits `--container-padding-inline-start` — and the inner Section escapes padding that is not visually there. The var crosses a boundary the layout does not.

**Fix idea:** the sheet panel should reset `--container-padding-*` to 0, the way `Section`'s own inner wrapper already does for its descendants. Any overlay that leaves its parent's visual box likely needs the same — Dialog, Popover.

**Workaround:** use a padded `VStack` (`<VStack gap={4} padding={4}>`) instead of `Section` inside the sheet. Stack has no escape behavior.

Found while writing the snap-points example in #5203, which uses the workaround.

Geervan:
Fixes #5208

### Problem
When a `<Section>` is rendered inside an overlay (`Drawer`, `Dialog`, `MobileNav`) that sits inside a page `<Section padding={4}>`, the inner `<Section>` calculates `-16px` negative margins on both sides and overflows the overlay by 32px.

`Section` uses `margin-inline-start: calc(-1 * var(--container-padding-inline-start, 0px))` to escape parent section padding. Because fixed overlay elements inherit CSS custom properties from ancestor DOM elements, the inner `Section` escapes padding that isn't visually there on the overlay panel.

### Fix
Reset container padding CSS variables to `0px` on root overlay styles in `Drawer.tsx`, `Dialog.tsx`, and `MobileNav.tsx`:

```ts
'--container-padding-inline-start': '0px',
'--container-padding-inline-end': '0px',
'--container-padding-block-start': '0px',
'--container-padding-block-end': '0px',

### CODE REVIEW

**Comment by imdreamrunner** [comment]
(this is an AI agent comment)

Not a review, just findings — I filed #5208, so I built this branch and measured it against the repro. **The mechanism is right.** Zeroing `--container-padding-*` at the overlay root is exactly the fix. Two things stop it from closing the issue as-is.

### 1. BottomSheet isn't covered — and it's the component in the issue

The branch is based on **2026-07-21, 634 commits behind main**, which predates the BottomSheet promotion into core. `packages/core/src/BottomSheet/` doesn't exist on it, so a rebase is needed before the component can be touched at all.

BottomSheet renders its own `<dialog>` (`BottomSheet.tsx`, `styles.dialog`) rather than going through `Dialog`, so it doesn't inherit this fix.

### 2. The Dialog change is a no-op today

I measured the variable at each level, using a 40px page padding so the leaked value is distinguishable from Dialog's own 16px:

```
page Section        40px
<dialog>            40px   ← the leak arrives (this PR sets it to 0)
  content wrapper   16px   ← Dialog's own padding already shadows it
    inner Section   16px   ← reads 16px either way
```

The escape margin is `-16px` with the fix and `-16px` without it. Same layout both ways.

That is also **why BottomSheet is the one that breaks**: its scroll body is `padding: 0`, so nothing shadows the leaked value and it reaches the `Section` intact — 672px of content in a 640px sheet.

Worth keeping the Dialog change as defence in depth (it is correct, and a future Dialog with an unpadded body would need it), but it isn't what closes #5208.

### Smaller things

- **No changeset.** Nothing under `.changeset/`; the rubric treats a missing changeset as blocking for a consumer-visible change (X20). CI seems not to flag the absent-file case.
- **The Drawer test doesn't test the fix.** It's titled "resets container padding custom properties on the root dialog element" but only asserts `expect(dialog).toBeInTheDocument()` — it passes unchanged with the fix reverted. jsdom does no layout, so the assertable thing is the computed custom property:
  ```js
  expect(
    getComputedStyle(dialog).getPropertyValue('--container-padding-inline-start'),
  ).toBe('0px');
  ```
- **Popover** is also non-portaling and has the same exposure; not covered here, and fine as a follow-up.

### Suggested shape

1. Rebase onto main.
2. Add the reset to BottomSheet's `<dialog>` (and its panel).
3. Add a changeset.
4. Make the test assert the property.

Repro I used, if it helps — the inner `Section` overflows the sheet by 16px each side:

```tsx
<Section padding={4}>
  <BottomSheet isOpen label="Sheet" height="tall" onOpenChange={() => {}}>
    <Section padding={4}>content</Section>
  </BottomSheet>
</Section>
```

Thanks for picking this up — the diagnosis is right, it just needs the newer base to reach the component that shows it.

Geervan:
Thanks for the detailed review feedback, @imdreamrunner, I dont know how i didnt realise the stale branch

### Updates in this push:

1. **Rebased onto latest `upstream/main`:**  
   Rebased onto HEAD commit (`33ba2b6a81b083568bf46cd579fb50e5b36285da`) so the branch is 0 commits behind `facebook/astryx:main`.

2. **Core Fix Applied to `BottomSheetPanel`:**  
   Added `--container-padding-*` resets (`0px`) directly to `styles.sheet` in `packages/core/src/BottomSheet/BottomSheetPanel.tsx`—the actual visual sheet surface container in `@astryxdesign/core`. Also added resets to `usePopover.tsx` (`styles.surface`) and preserved `Dialog.tsx` for defense-in-depth.

3. **Updated Unit Tests with Computed Style Assertions:**  
   Replaced passive `toBeInTheDocument()` checks with explicit `window.getComputedStyle(element)` assertions in `BottomSheet.test.tsx`, `Dialog.test.tsx`, and `Drawer.test.tsx`, asserting that `--container-padding-inline-start`, `inline-end`, `block-start`, and `block-end` directly evaluate to `'0px'`.

4. **Added Changeset:**  
   Added `.changeset/reset-overlay-container-padding.md` targeting `@astryxdesign/core`.

---

### Verification Summary
- `node scripts/check-changesets.mjs`: `✓ 1 changeset(s) valid`
- `BottomSheet` unit test suite (`BottomSheet.test.tsx`): 71/71 passed.
- `Drawer` unit test suite (`Drawer.test.tsx`): 39/39 passed.

imdreamrunner (approved):
(this is an AI agent review, run by @imdreamrunner)

Approving — this closes #5208. Verified on a build of `8d0c34e`.

**The repro is fixed.** Section inside a 640px sheet: 672px before (32px overflow), **640px now**. The sheet panel reads `--container-padding-inline-start: 0px` instead of the leaked 16px.

**The tests are real now.** I stripped the reset from `BottomSheetPanel.tsx`, `Dialog.tsx` and `Drawer.tsx` and all three new tests failed with `expected '' to be '0px'` — they discriminate, unlike the earlier `toBeInTheDocument()` version. Thanks for reworking those.

**No harm from the reset.** The four patched elements have no padding of their own to escape (`Drawer`/`MobileNav` dialogs are `padding: 0`, the sheet has only its safe-area `paddingBlockEnd`, and the popover surface keeps padding on the inner content div). Every consumer reads these as `var(--container-padding-*, 0px)`, so with no ancestor leak they already computed 0px — the reset changes a computed value only where an ancestor leaked one, which was the bug. Legitimate escapes still work: a Section inside a Dialog still bleeds to the dialog edge (640px in a 640px dialog, `-16px` margins), because Dialog's inner wrapper sets real values via `container()` and shadows the root zero.

Full suites for the five touched components: 351/351. `pnpm lint` clean. Core-only changeset frontmatter is correct — `@astryxdesign/lab` is `private: true`.

Two things I'm deliberately **not** blocking on; I'll send a follow-up PR for both:

1. **The other half of the leak.** `--astryx-section-padding` and `--layout-padding-inner-x/y` still cross the overlay boundary, so an *unpadded* `Section` in an overlay under `<Section padding={10}>` renders at 40px instead of the 16px theme default. The overflow is gone; this isn't.
2. **Three overlays still uncovered** — `Lightbox` (own `<dialog>`), `ContextMenu` and `HoverCard` (they call `useLayer` directly, so the `usePopover` surface reset misses them). Putting the reset in `useLayer`'s base style covers the last two, and an exported `containerPaddingReset` would stop the block being hand-copied — it's in six places now, which is why each new overlay keeps forgetting it.

One nit worth a quick push before merge: `pnpm exec prettier --write` on the three test files. The `expect(...)` lines exceed print width and `BottomSheet.test.tsx` / `Drawer.test.tsx` each end with a stray blank line. No CI job runs a format check, so it's green either way — it'd just land as drift.

imdreamrunner:
Thanks for the fix! Looks good!

PR #5210, opened 2026-08-19
docs: replace em dashes in DateRangeInput and FormLayout prop descriptions

josephfarina:
## What

Two AI-slop typographic tells (check 17, sub A1) in component prop descriptions, replaced with plain punctuation. Prose only; meaning preserved.

- `DateRangeInput.doc.mjs` (maxRangeSpan): `Rolling window relative to the start — for fixed calendar bounds` becomes `... relative to the start; for fixed calendar bounds`. Two independent clauses, so a semicolon.
- `FormLayout.doc.mjs` (defaultOptionality): `match the visual default — aria-required only, never` becomes `match the visual default; aria-required only, never`.

Both props exist on current main (added by #5145 and #4791). Renders verified via `astryx component <Name> --props`.

## Checklist
- [x] CLI `--props` output verified
- [x] Prose strings only; meaning preserved

Night Watch — Doc Reviewer

### DIFF
--- a/packages/core/src/DateRangeInput/DateRangeInput.doc.mjs
+++ b/packages/core/src/DateRangeInput/DateRangeInput.doc.mjs
@@ -102,7 +102,7 @@ export const docs = {
       name: 'maxRangeSpan',
       type: 'number',
       description:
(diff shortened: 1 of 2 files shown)


cixzhang (approved):
Punctuation only, reads the same. Thanks for grinding through these.

[Reviewed by Robohands]

PR #5211, opened 2026-08-19
feat(ComplexSelector): report open state with onOpenChange

cixzhang:
## Why

`ComplexSelector` owns a popup layer but could not tell a consumer when that layer opened or closed. A hover-to-open consumer can drive it through the existing `handleRef`, but without a callback it cannot learn when Escape, outside click, selection, or native light dismiss closes the popup without querying private DOM and risking stale state.

Astryx's visibility convention uses one callback for controllable layers:

```ts
onOpenChange?: (isOpen: boolean) => void;
```

The imperative handle remains the visibility-control API. This does not add a declarative `isOpen` prop.

## What

Adds `onOpenChange` to `ComplexSelector`, reports each real open or close once, and documents the callback alongside `handleRef`.

The callback is wired through the existing popover lifecycle, so trigger clicks, ArrowDown, Escape, light dismiss, content `close()`, and imperative `open`/`close`/`toggle` all use the same reporting path. Focus restoration runs before the close callback so a consumer may deliberately move focus from the callback.

## Integration with the dismissal fix

This is rebased onto [#5018](https://github.com/facebook/astryx/pull/5018). Its earlier local timing guard and overlapping `useLayer` change are gone; the PR now preserves the systemic gesture-aware dismissal behavior from current `main`.

A regression test covers the combined path: the browser's light-dismiss toggle reports `false` once, and the trigger click from that same gesture does not reopen the popup or report another state.

## Risk

Additive optional callback; no default behavior, rendering, styling, or declarative state change. Runtime work is limited to the callback when visibility actually changes.

## Testing

On the rebased head:

```text
vitest run ComplexSelector.test.tsx useLayer.test.tsx
2 files passed · 76 tests passed
```

Focused cases verify exactly-once reporting for trigger open/close, ArrowDown/open lifecycle, Escape, browser light dismiss/outside click, the [#5018](https://github.com/facebook/astryx/pull/5018) gesture path, content selection calling `close()`, and imperative `open`/`close`/`toggle`. Repeated requests for the current state report nothing.

## Not done

No declarative `isOpen` prop. Consumers continue to control visibility through `handleRef` and observe it through `onOpenChange`.

### DIFF
--- a/.changeset/complexselector-onopenchange.md
+++ b/.changeset/complexselector-onopenchange.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] ComplexSelector: `onOpenChange` reports every open and close of the popup — the trigger, the keyboard, a light dismiss, Escape, content calling `close()`, or the imperative handle. Paired with the existing `handleRef`, a consumer can drive and observe the surface without reading the shell's DOM.
+
+@cixzhang
--- a/apps/storybook/stories/ComplexSelector.stories.tsx
+++ b/apps/storybook/stories/ComplexSelector.stories.tsx
@@ -764,6 +764,7 @@ export const ControlledToolbarTrigger: Story = {
   name: 'Controlled toolbar trigger',
   render: () => {
     const [density, setDensity] = useState<ViewDensity>('Comfortable');
+    const [isOpen, setIsOpen] = useState(false);
     const selectorRef = useRef<ComplexSelectorHandle>(null);
 
     return (
@@ -779,6 +780,7 @@ export const ControlledToolbarTrigger: Story = {
           isLabelHidden
           value={density}
           onChange={setDensity}
+          onOpenChange={setIsOpen}
           triggerLabel={`Density: ${density}`}
           variant="ghost"
           startIcon="viewColumns"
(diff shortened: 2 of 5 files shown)


cixzhang:
API ruling: keep imperative visibility control through `handleRef`, but rename the observation callback from `onOpenChange` to `onToggle`. This is a visibility event, not a controlled data update; do not add declarative `isOpen`.

cixzhang:
Follow-up after checking the repository convention: keep this callback named `onOpenChange`, not `onToggle`. Astryx uses `onOpenChange?: (isOpen: boolean) => void` consistently for layer visibility (40 declarations across Dialog, Popover, Tooltip, HoverCard, Typeahead, menus, and others); `onToggle` is used for item/row expansion actions. The API guideline and audit rubric explicitly require `onOpenChange` for layer components. The earlier tentative naming direction is withdrawn.

cixzhang (commented):
Thanks, this is good and an approve once undrafted.

[Reviewed by Robohands]

cixzhang:
/accept-visual 33045632593/1 New ComplexSelector callback story; gate 2 verified the intended readout only.

cixzhang:
/accept-visual 33045632593/1 New ComplexSelector callback story; gate 2 verified the intended readout only.

cixzhang:
/accept-visual 33045632593/1 New ComplexSelector callback story; gate 2 verified the intended readout only.

PR #5213, opened 2026-08-19
fix(cli): reject malformed ASTRYX_LATEST_VERSION values

gwagjiug:
### Summary                                                    
                                                                
 getLatestVersion() accepted any ASTRYX_LATEST_VERSION value    
 that started with a MAJOR.MINOR.PATCH prefix. For example,     
 0.0.8garbage passed validation and could appear in an invalid  
 update hint.                                                   
                                                                
 This PR validates the complete environment variable value with 
 the existing shared isValidSemver() utility.                   
                                                                
 ### Why this approach                                          
                                                                
 The previous inline regular expression only checked the        
 beginning of the string:                                       
                                                                
 ```js                                                          
/^\d+\.\d+\.\d+/
 ```                                                            
                                                                
 Reusing isValidSemver() avoids maintaining a second, less      
 strict SemVer implementation and preserves support for valid   
 prerelease and build metadata.                                 
                                                                
 ### Changes                                                    
                                                                
 - Replace the prefix-only regular expression in                
   getLatestVersion() with isValidSemver().                     
 - Add a regression assertion for                               
   ASTRYX_LATEST_VERSION=0.0.8garbage.                          
 - Add a patch changeset for @astryxdesign/cli.                 
                                                                
 ### Validation                                                 
                                                                
 - node_modules/.bin/vitest run                                 
   packages/cli/clients/cli/lib/update-check.test.mjs — 11      
   tests passed                                                 
 - Manual reproduction: 0.4.5garbage now produces               
   {"latest":null,"hint":null}                                  
 - pnpm lint:strict                                             
 - pnpm build                                                   
 - pnpm test — 524 of 526 test files passed. Two failures       
   remain in untouched case-sensitivity tests:                  
   - packages/cli/api/component/component.test.mjs              
   - packages/cli/foundation/discovery/hook-discovery.test.mjs  
                                                                
 ### Scope                                                      
                                                                
 This PR only tightens validation of the latest-version         
 environment signal. It does not change version comparison or   
 installed-version detection.


cixzhang (dismissed):
Thanks — this routes the environment hint through the CLI’s shared version validator, and the regression test proves trailing garbage is ignored. [Reviewed by Robohands]

gwagjiug:
> Thanks — this routes the environment hint through the CLI’s shared version validator, and the regression test proves trailing garbage is ignored. [Reviewed by Robohands]

@cixzhang 
Thanks for the review! I've completed the Meta CLA signing process and am waiting for the CLA signed label to be applied.

PR #5214, opened 2026-08-19
feat(Spinner): make the ring geometry and colors themeable

freddymeta:
## Summary

Makes the Spinner's ring **themeable** — its per-size geometry and its two colors — without changing either prop's type, and without changing anything a consumer can observe today. `size` and `shade` stay fixed enums; what each named value *resolves to* becomes a theme's to set.

Four public custom properties on the `spinner` target:

| var | what it sets | default |
|---|---|---|
| `--spinner-diameter` | diameter of the drawn ring | 10 / 14 / 18 / 28px per size |
| `--spinner-stroke-width` | stroke width of **both** circles — the arc and the track | 2 / 3 / 3 / 4px per size |
| `--spinner-color` | the moving arc | the shade's token |
| `--spinner-track-color` | the track behind it | the shade's token |

```ts
spinner: {
  'size:xl': {'--spinner-diameter': '2.5rem', '--spinner-stroke-width': '0.375rem'},
  'shade:subtle': {'--spinner-track-color': 'transparent'},
  base: {'--spinner-color': 'var(--color-brand)'},
}
```

Any length and any color notation works: `rem`, `em` and `calc()` resolve into the radius and stroke the ring is drawn with, and colors accept `var()`, `color-mix()` and `currentColor`.

## How the SVG honours the vars

`r` and `stroke-width` are real CSS properties on an SVG shape, and a CSS declaration outranks the presentation attribute of the same name. So the attributes stay as the size's defaults — what a server render and a no-CSS render draw — and the cascade takes over the moment a theme has a value. The colors ride `stroke` directly. The dash pattern is composed from the same resolved diameter (`diameter x pi x arc-fraction`), so a themed ring keeps the same *fraction* of arc rather than the same absolute dash: every themed row below measures 135.000°.

The box is sized from the same two vars (`calc(diameter + 2 x stroke)`), so the box and the ring are one measurement by construction rather than by a listener keeping them in step — and that size is applied as an inline `width`/`height` written after the caller's `style`, which is exactly where the number used to go.

## Why public vars, and not derived or private ones

`--_*` means "the pipeline sets this, not you" — `validatePrivateVars` reports every theme that writes one — so the first revision of this PR advertised an API the build refuses. These are public, set directly, following `--button-focus-offset` and `--tree-list-indent`.

Derived vars are the other route, and they are right when the thing being themed genuinely is a CSS property of the element ([#5181](https://github.com/facebook/astryx/pull/5181)'s padding). Mapping `width` here would name the box rather than the ring, and mapping `color` would take the label's text color with it. A public var says what it is.

`sizeStyles` and `shadeStyles` declare the four vars on the element carrying the `spinner` theme target — the span, or the wrapper when a label moves the target there, never both. The box and the ring read them through a still-private registered `<length>` pair, so a themed stroke of `0` stays a zero-width stroke instead of poisoning `calc(28px + 0 * 2)`. The four public vars stay **unregistered** on purpose: a registered property has an `initial-value`, so every element reports a value for it, and `theme-var-reachability.js` finds a var's declaring element by exactly that test.

## Review changes in the latest revision

**`--spinner-rail-width` is now `--spinner-stroke-width`** ([@rubyycheung](https://github.com/rubyycheung)). The old name reads as the stationary track — and `--spinner-track-color` beside it made that the natural reading — while it is the stroke width of both circles, so a themed `0` takes the moving arc with it. Deliberately **not** split into two widths: nothing has asked to set the arc and the track independently, and a var per circle is a wider public contract than the demand for it. The private resolved var follows the name, and the doc entry in both locales now leads with the one-width-drives-both fact.

**The sizing precedence is unchanged, and the `[breaking]` changeset is gone** ([@rubyycheung](https://github.com/rubyycheung)). Sizing the box from the vars had moved the declaration out of the inline style and into a rule, which would have handed a caller's `style={{width}}` a precedence over the box that it has never had. Instead the size is composed into an unregistered `--_spinner-box-size` and applied inline after the caller's `style`, with the size's own frame as the `var()` fallback for a render where no stylesheet has declared it. So there is no behaviour change left to separate out or to describe:

| | `main` | this PR |
|---|---|---|
| `<Spinner size="xl">` | 36 × 36 | 36 × 36 |
| `<Spinner size="xl" style={{width: 999, height: 999}}>` | 36 × 36 | 36 × 36 |
| `<Spinner size="xl" style={{opacity: .5, margin: 7}}>` | 36 × 36, opacity .5, margin 7px | same |
| `<Spinner size="xl" label="…" style={{width: 999}}>` | 36 × 36 | 36 × 36 |

**`pathLength` is gone** — found while re-measuring, and it was mine. Normalizing the arc to the default circumference rescales the dash against the path length the UA measures on *its own approximation* of the circle: 87.398 against the 87.965 of `pi x 28`. That shortened the default arc by 0.64% and moved the cap by half a pixel — **191 changed pixels** on an `xl` spinner against `main`, in a PR whose whole claim is that defaults do not move. Composing the dash from the resolved diameter is exact at the default and scales identically when a theme moves it.

**Registration is at module scope** ([@cixzhang](https://github.com/cixzhang)) — pushed in `42ae22fc`, about half an hour after the review that asked for it, so that block was written against the previous head. Registering an inherited property with an `initial-value` invalidates style for the whole document, and a spinner mounts onto a page that has already rendered.

## What a stroke width of `0` does — asked in review

**It paints nothing, and that is what it should do.** One `stroke-width` drives both circles, so zeroing it is a zero-width stroke on each: `box 28×28, r 14px, stroke-width 0px` — an empty box that still announces "Loading". That is what `stroke-width: 0` means everywhere else in CSS, and the case people actually want — an arc with no track behind it — is `--spinner-track-color: transparent`, independent of geometry. The doc entry says so, and the story that promised a trackless arc is a 1px hairline over a transparent track, which is what it was trying to show.

## Verified in Chromium 149

**Defaults are untouched.** Every size × every shade × the labelled case × an inline-in-text case, at four frozen rotation angles: **72 of 72 element screenshots byte-identical** between a build of `origin/main` and a build of this branch, with box geometry and position identical.

| size | box | `r` | `stroke-width` |
|---|---|---|---|
| sm | 14 × 14 | 5px | 2px |
| md | 20 × 20 | 7px | 3px |
| lg | 24 × 24 | 9px | 3px |
| xl | 36 × 36 | 14px | 4px |

**A theme moves them**, through a built Storybook so the values travel the real `@scope` + `@layer astryx-theme` path:

| themed | diameter / stroke written | box | `r` | `stroke-width` | arc |
|---|---|---|---|---|---|
| sm | `1rem` / `0.125rem` | 20 × 20 | 8px | 2px | 135.000° |
| md | `1.5rem` / `0.25rem` | 32 × 32 | 12px | 4px | 135.000° |
| lg | `2rem` / `0.3125rem` | 42 × 42 | 16px | 5px | 135.000° |
| xl | `calc(2rem + 8px)` / `0.375rem` | 52 × 52 | 20px | 6px | 135.000° |
| hairline xl | default / `1px` | 30 × 30 | 14px | 1px | 135.000° |

**Colors**: a themed arc and wash resolve (`rgb(0,69,140)` over `rgb(196,221,251)`), and `shade:subtle` with `transparent` gives `rgba(0,0,0,0)` on that spinner only.

**Reachability guard**, against a full production build:

```
✓ Spinner --spinner-color — .astryx-spinner sets it (light-dark(#262626, #ebebeb) → 9987px)
✓ Spinner --spinner-diameter — .astryx-spinner sets it (14px → 9987px)
✓ Spinner --spinner-stroke-width — .astryx-spinner sets it (3px → 9987px)
✓ Spinner --spinner-track-color — .astryx-spinner sets it (light-dark(#ccd3db, #5a5e66) → 9987px)
```

## Split out of this PR, at review request

- **The flex-shrink fix is [#5484](https://github.com/facebook/astryx/pull/5484)**, against `main`. Its cause turned out to be the box's leftover `overflow: hidden`, which zeroes a flex item's automatic minimum size; the PR has the measurements.
- **Inline size → CSS size** is no longer a change at all — see above. Its `[breaking]` changeset is deleted.

## Test plan

- `pnpm build`, `pnpm lint:strict` (0 errors; the same 80 warnings `origin/main` reports), `pnpm check:repo`, `pnpm -F @astryxdesign/core typecheck` — all clean.
- `pnpm exec vitest run packages/core/src/Spinner packages/core/src/theme packages/cli/api/theme`: 28 files / 927 tests pass.
- Unit tests cover what jsdom can hold honestly: the theme *routing* (a `size:xl` key has to land on `.astryx-spinner.xl`, asserted on the whole emitted rule), the authored dash staying absolute with no `pathLength`, the registration timing, and the box-sizing precedence (`style={{width: 999}}` leaves the composed value in place, other style properties pass through). The cascade itself has no jsdom equivalent — the Chromium numbers above are its evidence.
- `build.public-component-vars.test.mjs` is doc-driven: it scans every `*.doc.mjs`, builds one theme naming every non-private, non-derived var, and asserts the builder reports no private-var warning, so a new component earns the check for free.

@freddymeta


cixzhang (changes requested):
🔴 **Blocking** — the theming API this adds is rejected by `astryx theme build`, plus `needs:code-review` / `needs:design-review`.

`Triage: new API surface (theming) · non-breaking by default · low blast radius → deep path · checks: T2, P1, X20, §derived-vars`

Thanks — this is a real gap, and the care shows. I checked the "byte-identical" claim in Chromium against main's Storybook with the animation frozen, and it holds exactly: `Sizes` and `WithLabel` come out pixel-for-pixel the same. The `size:xl` key form is right, the specificity reasoning is right, the labeled-case cascade is right, EN + zh docs and a correct `[feat]`/patch changeset are all there, and folding the second `getComputedStyle` into `canvasStyle` (`Spinner.tsx:217`) is a nice cleanup.

Three things block. The first is a tier call — I've made it below, so this is a rework you can start on.

**1. Use public vars, not private ones.** `--_` + `private: true` means "themes never set this directly" ([Theming Infrastructure → Private vars](https://github.com/facebook/astryx/wiki/Theming-Infrastructure)), and `validatePrivateVars` in `packages/cli/api/theme/build/build.mjs:1025` enforces it. Running the exact snippet from the description:

```
✗ Component "spinner" (size:xl) sets private var "--_spinner-diameter". Private vars (--_*)
  are internal — use standard CSS properties instead. The pipeline expands them automatically.
✗ Component "spinner" (size:xl) sets private var "--_spinner-rail-width". …
2 private var error(s).
```

That snippet is also in the changeset and in `Spinner.doc.mjs:47-48` / `:103-104`. (It still emits CSS after logging `✗`, which is probably why the throwaway build read as a pass.)

Please go with **`--spinner-diameter` / `--spinner-rail-width`** — no underscore, `private: false`, set directly, the `--button-press-scale` precedent. The other route, derived vars (`width` → `--_spinner-diameter`, the `progressbar-mark` shape at `derivedVarRegistry.ts:71-73`, which is what you did for padding in [#5181](https://github.com/facebook/astryx/pull/5181)), is the right answer when the thing being themed genuinely *is* a CSS box property. Here it isn't: the ring is painted on a canvas, so `width` and `borderWidth` would name properties the element doesn't really have, and a theme author reading `borderWidth` on a spinner would reasonably expect a border. A public var says what it is. That does mean an entry in `VARS_WITHOUT_DERIVED_MAPPING` (`derivedVarRegistry.test.ts:286`) — fine, that list is public vars and self-set vars, which is exactly what these are.

**2. Only `px` overrides survive.** `parseFloat(canvasStyle.getPropertyValue(…))` at `Spinner.tsx:218-223` reads an unregistered custom property, so it gets the specified token and drops the unit — while the box at `:73-74` goes through real `calc()`. Measured on this PR's Storybook, `size:xl`:

| theme sets | CSS box | drawn ring |
|---|---|---|
| `40px` / `6px` | 52px | 46px ✓ |
| `2.5rem` / `0.375rem` | 52px | **3.25px** |
| `calc(2rem + 8px)` | 52px | 36px (`NaN` → fallback) |
| rail `0px` | **28px** | 36px canvas, **clipped** by `overflow: hidden` |

The last row is the `||` at `:219`/`:222` reading `0` as absent. Registering the properties (`@property { syntax: '<length>'; inherits: true }`) makes the computed value a used px length and fixes the rem/em/calc rows in one move; `Number.isFinite(x) ? x : default` fixes the `0` row.

**3. Box and ring only agree at mount.** Deps are `[shade, size, themeTokens]` (`Spinner.tsx:295`), so a var change that isn't a size/shade/token change never redraws. Changing the var at runtime: box went 36px → 52px, canvas stayed 36px. Worth narrowing the description's "stay in sync … after a theme switch" to mount-time overrides — or observing the var.

Non-blocking:

- The six new tests all pass against `main`'s `Spinner.tsx` unchanged (I swapped the file and ran vitest — 23/23). They assert `themeProps` output that already existed and that four sizes hash to four distinct class strings. jsdom has no canvas and no layout, so nothing here can reach the actual contract; a Storybook check is the way, and `needs:design-review` wants screenshots regardless.
- `Spinner.tsx:69-73`, `:93-97`, `:211-217`, `:321-328` restate the description in-file — a line or two each is plenty.
- Dropping the inline `width`/`height` also flips precedence: a consumer's `style={{width}}` now wins where `frameSize` used to override it. Fine by me, worth a line in the changeset.
- Pre-existing, not yours: `SpinnerProps.size` JSDoc says xl is 36px, but `SIZES.xl.diameter` is 28 (36 is the frame), and it still says "three sizes". Happy to take that separately.

cixzhang:
[#5408](https://github.com/facebook/astryx/pull/5408) landed the Spinner as SVG, which changes the ground under this PR — sorry, that's ours arriving mid-review, not you.

The ring now paints `stroke: currentColor`, so **the arc colour needs no var at all: set `color` on the host and the ring follows**, including on hover. `Button` and `CheckIndicator` already work this way. That should remove `--spinner-color` and the second `getComputedStyle` with it.

Geometry is the half worth keeping. Diameter and rail have no `currentColor` equivalent, but on SVG they're plain CSS properties plus `pathLength` — no vars, no JS, and the mount-only redraw and the `--spinner-rail-width: 0` hairline both go away, since both were canvas artifacts.

Could you rebase onto main and see what's left? I think it's a lot smaller.

One question that decides whether any var survives: **should a theme set the track independently of the arc, or is the track always 30% of it?** `color` gives you one channel. If 30% is the design, nothing else is needed.

[Reviewed by Robohands]

cixzhang (dismissed):
Thanks — verified in Chrome: rem, calc and a 0 rail all reach the ring now, labelled case included. Description still describes the canvas version though.

[Reviewed by Robohands]

cixzhang (dismissed):
Thanks, and apologies — I approved this an hour ago and have retracted it. Two things I should have caught.

The earlier request-changes on this PR was still open, and that review superseded it without reconciling. Of its three blockers, the public-var one is genuinely fixed — `--spinner-diameter` / `--spinner-rail-width`, no underscore, exactly the shape asked for.

The bigger one: [#5408](https://github.com/facebook/astryx/pull/5408) landed on main and draws the ring in SVG rather than canvas. This PR themes the canvas — `getComputedStyle` plus `parseFloat` off a canvas element, and a redraw effect keyed to it. It merges cleanly because the texts do not collide, but the implementation it is theming is gone.

Worth rebasing onto the SVG ring and re-checking what the new drawing path actually needs; a good part of the measurement work should carry over.

[Reviewed by Robohands]

cixzhang (commented):
Correcting my own review — the one below this is wrong, and I'm sorry for the noise.

I asked you to rebase onto [#5408](https://github.com/facebook/astryx/pull/5408). You'd already done it, six hours before I asked: the 08:37 merge reimplements the theming on the SVG ring, no canvas, no `getComputedStyle`, no redraw effect. I reviewed a stale head and wrote a block that had nothing behind it. The merge note reconciling the two is the clearest I've read.

All three blockers from 20 Aug are genuinely closed — public var names, `rem`/`em`/`calc()` all reaching the ring, box and ring in step by construction. The CLI test guarding the doc-declared vars is a better answer than the one I asked for.

One real thing the rebase changed: **`--spinner-rail-width: 0` now paints nothing.** One `strokeWidth` drives both circles (`Spinner.tsx:243`), so zeroing the rail zeroes the arc with it. The story's own caption promises "an arc with no track"; what renders is an empty 28px box that still announces "Loading".

| a themed rail of 0 | the same spinner, default |
|---|---|
| rail 0 | default |

Both circles compute `stroke-width: 0px`; the element screenshot is a blank 116-byte PNG.

What should a 0 rail do — hide the ring entirely, or only the track?

Separately, and not yours to resolve: the two colour vars are a maintainer's call on whether the system wants per-component paint overrides when the arc already follows the shade tokens. Cindy has it. The geometry half lands either way.

[Reviewed by Robohands]

packages/core/src/Spinner/Spinner.tsx
```diff
@@ -140,30 +232,83 @@ const styles = stylex.create({
...
     strokeLinecap: 'round',
+    // The geometry the ring is actually drawn at. `r` and `stroke-width` are
+    // CSS properties on an SVG shape, and a CSS declaration outranks the
+    // presentation attribute of the same name — so the attributes below stay
+    // as the size's default (and as what a server render and a no-CSS render
+    // draw), and these take over the moment the cascade has a themed value.
+    // The arc's dash pattern needs no rule of its own: `pathLength` normalizes
+    // it to the default circumference, so it scales with whatever `r` becomes.
+    r: `calc(var(${RESOLVED_DIAMETER}) / 2)`,
+    strokeWidth: `var(${RESOLVED_RAIL})`,
```
cixzhang: One `strokeWidth` for both circles, so a themed rail of 0 hides the arc as well as the track.

cixzhang:
Direction settled on our side, so you're not rebasing into a moving target: **spinner theming comes through this PR.** [#5250](https://github.com/facebook/astryx/pull/5250), which was solving the neighbouring half by passing `shade="inherit"` per host, is closed in favour of it.

So the rebase onto the SVG ring is the whole remaining ask, and the hosts that own their own surface needing a `color` becomes part of this rather than a separate PR.

[Reviewed by Robohands]

cixzhang:
Correction, and it reverses what I said above — my earlier comment was wrong, so please go by this one.

**Keep all four vars.** `--spinner-color`, `--spinner-track-color`, `--spinner-diameter`, `--spinner-rail-width`, as you proposed them.

I argued `currentColor` made `--spinner-color` unnecessary. It doesn't, and [#5250](https://github.com/facebook/astryx/pull/5250) is the proof I should have read that way the first time: `currentColor` resolves against whatever the host's cascade happens to carry, and six of our own input hosts set no `color` at all. So "set `color` on the host" is not an interaction a theme author can rely on — it works or doesn't depending on a host they don't control and can't see. An explicit var is a guaranteed seam; `currentColor` is a hope.

So the rebase onto the SVG ring is mechanical rather than a redesign: the vars survive, `stroke` reads them, and the geometry pair loses the canvas artifacts (the mount-only redraw and the `--spinner-rail-width: 0` hairline) without losing the properties.

The track question I asked earlier is answered by keeping `--spinner-track-color`: independently settable, not pinned to a ratio of the arc.

Two things in here that aren't theming and are worth separating so they can land on their own merits:

- **The flex-shrink fix** — a Switch thumb at 14px compressing the box while the ring drew at full size is a real bug, and it's independent of the vars.
- **Inline size → CSS size.** This flips precedence: a consumer's `style={{width}}` now wins where the inline size used to. That's a behaviour change on shipped surface, and it should say so in the changeset rather than riding inside a `[feat]`.

Sorry for the reversal — that one is on me, not you.

[Reviewed by Robohands]

freddymeta:
Thanks — and no problem about the reversal; the vars were still in the branch, so keeping all four cost nothing to undo. The rebase onto the SVG ring was already in at `1a41110` (the 24 Aug merge), so what was left was your two separations. Both are done, plus an answer to the rail question and one thing I found while re-measuring.

**1. The flex-shrink fix is out, as [#5484](https://github.com/facebook/astryx/pull/5484) against `main`.** One property, its own changeset, before/after screenshots.

Correcting my own claim while moving it, because it did not survive measurement: **on `main` today the internal hosts do not reproduce it.** A Switch thumb is 14px around a 14px spinner box — no compression, nothing clipped, `sm`/`md` and CheckboxInput all measured. The numbers in my old description were taken with a theme enlarging the ring, which I should have said. What does reproduce with no theme anywhere is a consumer's own flex host narrower than the spinner: a 36px ring in a 16px host comes out as a 16×36 box with 20px of ring cut off by `overflow: hidden`. So it is still a real bug on `main` and still lands on its own merits — just not the one the description described.

**2. Inline size → CSS size has its own changeset now**, tagged `[breaking]`: a caller's inline `style` width wins over the box where the component's inline size used to win. It stays in this PR because the box has to read the same vars as the ring, but it is out of the `[feat]` and says what flips and what to do about it. I picked `[breaking]` because it is the tier that surfaces in Breaking Changes — if you would rather not bump the minor for a precedence flip nobody could have been relying on, it is a one-word edit to `[fix]`.

**3. A rail of `0` paints nothing, and I think that is right.** `box 28×28, r 14px, stroke-width 0px` — an empty box that still announces "Loading". One `stroke-width` drives both circles, and a zero-width stroke painting nothing is what `stroke-width: 0` means everywhere else; the thing people want, an arc with no track, is `--spinner-track-color: transparent` and is independent of geometry. So no code change: the doc entry says this now instead of promising a trackless arc, and the story that promised it is a 1px hairline rail over a transparent track (`box 30×30, r 14px, stroke-width 1px`), which is what it was trying to show. If you would rather `0` hid only the track, that needs the two circles to take separate widths — happy to, but it reads as magic to me.

**4. The "Default —" rows in the themed stories were not default.** A `Theme` with no parent `Theme` syncs its name onto the document root so its `@scope`'d rules reach portals — which also means they reach spinners rendered outside the provider. In a page holding both rows, the *unthemed* row measured 20 / 30 / 40 / 52: the themed numbers. Removed rather than relabelled, one `Theme` per story, unthemed reference named. Not specific to Spinner — `Avatar.stories.tsx` pairs "Default fallback" with a sibling `<Theme>` the same way — so if that wants a guard or a doc note, it is its own change and I will take it.

Defaults are unchanged and re-measured against a build of `origin/main`: 14 / 20 / 24 / 36 boxes, `r` 5/7/9/14, `stroke-width` 2/3/3/4, same arc and track colours. Themed geometry, `rem`, `calc()` and both colour vars all reach the ring; the tables are in the description, which is rewritten for the SVG implementation — the canvas sections are gone.

One housekeeping thing: the PR still reads **blocked** on the 24 Aug `CHANGES_REQUESTED`. You corrected it in the comment below it, but a comment does not clear the review state — could you dismiss it or re-review when you get a moment?

freddymeta:
Follow-up: rebased onto current `main`, and **the mechanism this PR was built around is obsolete** — [#5410](https://github.com/facebook/astryx/pull/5410) fixed it. Worth writing down, because it retires a claim I made twice in this thread.

The premise was that a component cannot declare its own public var: StyleX emitted custom-property declarations at priority 0, outside its layers, and unlayered beats `@layer astryx-theme`. #5410 splits the compiled CSS into `astryx-base` and the product layer, so a component declaration loses to a theme rule now. **The four "pre-existing casualties" I listed — `--button-focus-offset`, `--button-icon-only-aspect`, `--tree-list-indent`, `--tree-list-row-gap` — are not inert any more.** They pass the new reachability guard in the same run as ours. No separate PR needed; that item is closed by your build fix.

Your `theme-var-reachability` guard is what surfaced it, and it was right to fail us: the fallback shape declares the var nowhere, so there is no element for the guard to check, and all four Spinner vars read as unreachable. So `d18ef06` moves them onto the sanctioned shape — `sizeStyles` and a new `shadeStyles` declare all four on the element carrying the theme target (the span, or the wrapper when a label moves the target there, never both), and the box, the ring and the circles read them. The private registered `<length>` pair stays, so a themed rail of `0` is still a zero-width stroke rather than a poisoned `calc()`. The public vars stay unregistered deliberately: a registered property has an `initial-value`, every element then reports a value, and your guard would resolve the declaring element to `<html>`.

Re-measured against a full production build, everything is where it was before the reshape:

| | box | `r` | `stroke-width` |
|---|---|---|---|
| defaults sm/md/lg/xl | 14 / 20 / 24 / 36 | 5 / 7 / 9 / 14px | 2 / 3 / 3 / 4px |
| themed `rem` + `calc()` | 20 / 30 / 40 / 52 | 8 / 12 / 16 / 20px | 2 / 3 / 4 / 6px |
| labelled, themed through the wrapper | 52 | 20px | 6px |

Defaults are byte-for-byte the numbers a build of `origin/main` produces. Colours: `--spinner-color: #d93025` → `rgb(217,48,37)`, a `color-mix()` track resolves, `shade:subtle` + `transparent` → `rgba(0,0,0,0)` on that spinner only. Guard: `✓ Spinner --spinner-diameter — .astryx-spinner sets it (14px → 9987px)`, and the same for the other three.

One practical note for anyone else hitting the guard locally: it needs a full `pnpm build` before the Storybook build. Against a partially rebuilt tree the layer split is missing and *everything* fails, Button and TreeList included — which reads exactly like a real cascade regression.

cixzhang (changes requested):
Thanks — the rail answer is right, and the doc entry is the right home for it. Checked at `d18ef06`: box, radius, stroke width and both colours are unchanged from main at every size, every shade, and the labelled case.

| main | this PR |
|---|---|
| main | head |

I've cleared the stale request-changes — sorry that took a day; it was wrong when it was written.

One thing to fix. `registerSpinnerVars()` runs from the ring's ref callback, so the two registrations land at the **first spinner mount** — and registering an inherited property with an `initial-value` invalidates style for the whole document. A spinner is the loading indicator, so that lands on a page that is already rendered, with someone already waiting: mounting one costs 10.3 ms of style recalc on an 8,400-element page against 0.8 ms with the call removed, and 73 ms on a heavily-styled one.

At module scope it costs 0.7 ms and a themed rail of `0` still draws its 28×28 box. Your own `typeof CSS === 'undefined'` guard already covers the server render. Could you move it?

The two colour vars are still Cindy's call, and the geometry lands either way.

[Reviewed by Robohands]

packages/core/src/Spinner/Spinner.tsx
```diff
@@ -78,12 +144,24 @@ function pinRingsToTimelineOrigin(): void {
...
+ * so it lands before the first paint. It reads nothing back — the geometry is
+ * resolved by the cascade, not in JS.
+ */
 function syncRotationPhase(
   svg: SVGSVGElement | null,
 ): (() => void) | undefined {
+  if (svg == null) {
+    return undefined;
+  }
+  registerSpinnerVars();
```
cixzhang: This lands the registration at the first spinner mount, so the document-wide recalc hits a full page. Module scope instead?

cixzhang (changes requested):
Thanks — I previously asked for the SVG rebase, all four vars, rail-0 semantics, the flex-shrink split, and a separate sizing-precedence changeset; those are resolved.

One thing still blocks: first-mount registration adds a 21.7 ms style recalculation on a populated page while its loading indicator appears. The same registration takes 0.7 ms at module scope with zero changed pixels. Could you move it there?


[Reviewed by Robohands]

rubyycheung (dismissed):
The theming capability is useful, but I have two concerns with the public contract before approval.

First, `--spinner-rail-width` sounds like it controls the stationary track, while it actually controls the stroke width of both the track and the animated arc. Setting it to `0` therefore makes the entire Spinner disappear. Since this becomes durable public API—and `--spinner-track-color` already uses “track” specifically—please rename it to something that reflects both circles, such as `--spinner-stroke-width`, or split the widths if independent control is intended.

Second, this also changes inline `width`/`height` precedence, allowing those values to override the calculated Spinner box and potentially desynchronize or clip the ring. Please preserve the existing sizing precedence or move that behavior change into a separately reviewed change rather than coupling it to the theming API.

freddymeta:
Both of @rubyycheung's points are in, and one of them turned out to delete a change rather than move it.

**1. `--spinner-rail-width` → `--spinner-stroke-width`.** You are right that the name pointed at the track while the value drives both circles, and that `--spinner-track-color` next to it makes that reading the natural one — the `0` behaviour is surprising precisely because of the name. Renamed, and the doc entry in both locales now leads with "the stroke width of both circles the ring is drawn from — the moving arc and the track behind it" instead of mentioning the rail at all.

I did **not** split it into two widths. Independent control is a wider public contract, and nothing has asked for it — the case people want from a zero width is an arc with no track, which is already `--spinner-track-color: transparent`. Happy to add the split if you have a use for it; it is additive either way.

**2. The sizing precedence is no longer changed at all** — I preferred your first option to the second. The box is now sized by an unregistered `--_spinner-box-size` applied as an inline `width`/`height` written *after* the caller's `style`, which is exactly where the number used to be written, so the component's own size keeps winning; the fallback in the `var()` is the size's own frame, for a render where no stylesheet has declared it. The box and the ring still come from the same two vars, so themed geometry still moves them together. Measured against a build of `main`:

| | `main` | this PR |
|---|---|---|
| `<Spinner size="xl">` | 36 × 36 | 36 × 36 |
| `<Spinner size="xl" style={{width: 999, height: 999}}>` | 36 × 36 | 36 × 36 |
| `<Spinner size="xl" style={{opacity: .5, margin: 7}}>` | 36 × 36, opacity .5, margin 7px | same |
| `<Spinner size="xl" label="…" style={{width: 999}}>` | 36 × 36 | 36 × 36 |

The `[breaking]` changeset is deleted — there is no behaviour change left to review separately. There is a unit test on the precedence now, so it cannot drift back.

**3. Something I found while re-measuring, and it was mine.** `pathLength` rescales the dash against the path length the UA measures on its own approximation of the circle — 87.398 against the 87.965 of `pi × 28` — so the default arc was rendering **0.64% short**, 191 changed pixels on an `xl` spinner against `main`. In a PR whose claim is that defaults do not move, that is exactly the kind of thing the claim should have caught. `pathLength` is gone; the dash is composed from the resolved diameter instead, which is exact at the default and scales the same way when a theme moves it. Every themed row measures 135.000° of arc.

Re-measured after all three: every size × shade × the labelled case × an inline-in-text case, at four frozen rotation angles — **72 of 72 element screenshots byte-identical** to a build of `origin/main`.

@cixzhang — the module-scope registration you asked for was already pushed in `42ae22fc`, about half an hour after that review, so it was written against the previous head; it has been at module scope since. The four vars still pass the reachability guard after the rename.

cixzhang (approved):
Thanks — our module-scope ask was fixed. Ruby's two are now in too: the name matches both circles, box sizing keeps its old precedence, and defaults are pixel-identical to main.

[Reviewed by Robohands]

freddymeta:
@rubyycheung re-requesting — both of your asks are in `d673c7dcb`, which came after your review.

1. **`--spinner-rail-width` → `--spinner-stroke-width`.** You were right that "rail" reads as the stationary track, and `--spinner-track-color` sitting beside it made that the natural reading. The docs now say plainly that one stroke width drives both circles, and that `0` is honoured as a zero-width stroke — it paints nothing rather than falling back to the default. For an arc with no track behind it the answer is `--spinner-track-color: transparent`, which is now documented on the var itself so nobody has to discover it by setting `0` and losing the spinner.

2. **Sizing precedence is preserved.** The inline `width`/`height` precedence change is out; the calculated box wins as it did before, so the ring cannot be desynchronized or clipped by an author's inline size.

cixzhang has approved since. Yours is the one still open.

cixzhang (approved):

cixzhang (approved):

[Reviewed by Robohands]

cixzhang (approved):

[Reviewed by Robohands]

cixzhang:
/accept-visual 33045918232/1 These 12 frames add Spinner coverage; no existing pixels changed, and sensor-backed review verified unchanged defaults plus the intentional themed geometry.

freddymeta:
/accept-visual 33045918232/1 These 12 frames add Spinner coverage; no existing pixels changed, and sensor-backed review verified unchanged defaults plus the intentional themed geometry.

cixzhang:
/accept-visual 33045918232/1 These 12 frames add Spinner coverage; no existing pixels changed, and sensor-backed review verified unchanged defaults plus the intentional themed geometry.

PR #5217, opened 2026-08-19
feat(ci): check ICU plural categories in the i18n catalog gate

nynexman4464:
`check:i18n-catalog` compared key sets only. A plural branch naming a category the locale never selects passed unnoticed — and such a branch is dead code: `intl-messageformat` asks `Intl.PluralRules` which category a number falls in and drops to `other` when the message has no branch for it, so the translator's wording never renders.

Crowdin is about to make this concrete. On 2026-08-26 it drops the `many` category for Hebrew to match current CLDR (one/two/many/other → one/two/other). Our Hebrew catalog is already one/two/other and current ICU already agrees, so nothing changes for us — but nothing was watching either, and the next language CLDR revises would land the same way.

## What the gate does now

Every message in every catalog is parsed with the same ICU parser the runtime uses, and each plural branch is checked against the categories `Intl.PluralRules` reports for that locale.

Categories are read at check time, never hardcoded — a CLDR table in the repo is exactly what breaks when CLDR moves, and it would then fail catalogs that are correct. The check also inherits the ICU data of whatever Node runs it, which is the same data the app will use.

Using the parser rather than a regex is what makes the rest correct: `{who, select, many {…}}` is not a plural, `selectordinal` is scored against the ordinal categories (Hebrew has three cardinal categories and one ordinal), `=0` is an exact match rather than a category, quoted braces are not syntax, and a plural nested inside another branch is still found.

## Severity

- **Unreachable branch in `en.json` — error.** We author it, and every translation is modelled on it.
- **Unreachable branch in a translation — note.** Crowdin owns that content and it degrades to `other`; failing here would red every translations PR, the same reasoning that makes untranslated keys a note.
- **A message that does not parse — error, anywhere.** It throws when formatted.

Not included: placeholder parity between a translation and its English source. That is a different comparison and worth its own change.

## Test plan

- `pnpm check:i18n-catalog` passes on all 29 shipped locales plus `en` — 7,404 messages parse, zero dead branches.
- Each new rule was removed in turn to confirm its test fails and nothing else does: cardinal/ordinal split, the `=N` exemption, the unsupported-locale guard, nested descent, and plural-vs-`select`.
- Injected into real catalogs, then reverted: a `many` branch in Hebrew reports the dead branch and exits 0; an unbalanced brace in Hebrew reports `EXPECT_ARGUMENT_CLOSING_BRACE` and exits 1; a `few` branch in `en.json` exits 1.
- `pnpm lint:strict` clean. Full `pnpm test` matches `main`'s failures on this machine (all in `packages/cli`, timeouts under load); the one difference passes on its own.

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,6 +51,7 @@
     "@eslint-react/eslint-plugin": "^5.10.4",
     "@eslint/compat": "^2.1.0",
     "@eslint/js": "^10.0.1",
+    "@formatjs/icu-messageformat-parser": "^3.5.12",
     "@playwright/test": "^1.61.1",
     "@testing-library/user-event": "^14.5.0",
     "@types/node": "^25.9.3",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -52,6 +52,9 @@ importers:
       '@eslint/js':
         specifier: ^10.0.1
         version: 10.0.1(eslint@10.6.0(jiti@2.7.0))
+      '@formatjs/icu-messageformat-parser':
+        specifier: ^3.5.12
+        version: 3.5.12
       '@playwright/test':
         specifier: ^1.61.1
         version: 1.61.1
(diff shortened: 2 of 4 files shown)


nynexman4464:
Thanks — the plural-category validation from this draft has been folded into #5188, so I’m closing this duplicate.

PR #5218, opened 2026-08-19
fix(core): slide the Mobile Nav drawer in, not just out

imdreamrunner:
## The bug

`MobileNav` has an exit animation but no entry animation. Opening it puts the drawer straight into place; closing it slides out over `--duration-medium`.

## Why — two causes, and neither fix works alone

**1. The transition has no before-change value.** The `<dialog>` is `display: none` until `isOpen`, so nothing inside it is rendered while closed. React commits the open `display` and the open `transform` in the same pass, so the first frame the drawer is ever rendered in already holds the on-screen transform — and a transition with nothing to run from does not run. Same for the `::backdrop`, which only exists once `showModal()` has put the dialog in the top layer, already opaque. `@starting-style` supplies that value.

**2. The dialog was a scroll container.** It clipped the off-screen drawer with `overflow: hidden`. A scroll container in the top layer whose subtree holds *another* scroller — here the drawer's own content area — does not paint a `@starting-style` entry transition for its descendants in Chromium. The transition ticks in the CSSOM while every painted frame shows the end value. `overflow: clip` clips exactly as `hidden` did without creating a scroll container.

Minimal repro of (2): dialog `overflow: hidden` + a scrolling child in the drawer → snaps. Either one alone → animates.

Closing was always fine: by then the drawer has been painted, so the transform has a real previous value and no starting style is involved.

## Verified in Chromium — on painted frames, not computed style

**Computed style is the trap in this bug**, and I fell in it: `getComputedStyle` reports a perfect `320 → 268 → 221 → …` interpolation on open while the screen shows the drawer already in place. `getBoundingClientRect()` and `page.screenshot()` are also unreliable for compositor-driven animations. The only instrument that tells the truth is a CDP screencast of painted frames.

| | before | after |
|---|---|---|
| painted intermediate positions, open | **0** | **70–98** |
| painted intermediate positions, close | 61–67 | 61–67 (unchanged) |

Checked at both edges in LTR and RTL — each slides in from its own edge — and under `prefers-reduced-motion`, where it collapses to a 10ms transition. Drawer content still scrolls, the dialog still clips (no overflow bleed), and scroll containment is identical to `main`.

## Tests

`MobileNavEntryAnimation.test.tsx` — 8 tests. jsdom has no top layer, no transitions, no `@starting-style` and no compositor, so the animation itself is not observable there; the tests pin the three declarations the behaviour rests on, and each half of the fix is guarded independently (reverting either one fails tests).

The `overflow: clip` assertion is the one worth keeping: `hidden` looks like a pure clipping choice and clips just as well, so swapping it back reads as a harmless tidy-up. It isn't — it silently removes the entry animation while every CSSOM-based check still says the transition runs.

Still true, and not addressed here: there is no browser-based animation test venue in this repo, so nothing can catch a *behavioural* regression of this kind — only the declarations are pinned.

### DIFF
--- a/.changeset/mobile-nav-entry-animation.md
+++ b/.changeset/mobile-nav-entry-animation.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] MobileNav: the drawer now slides in when it opens, instead of only sliding out when it closes. Two things were needed, and each is useless without the other. First, the dialog is `display: none` while closed, so the drawer's first rendered frame already holds the on-screen transform and a transition has no before-change value to run from — `@starting-style` supplies it, for the drawer's transform and for the `::backdrop`'s opacity. Second, the dialog clipped the off-screen drawer with `overflow: hidden`, which makes it a scroll container; a scroll container in the top layer whose subtree holds another scroller (the drawer's content area) does not paint a `@starting-style` entry transition for its descendants in Chromium — the transition ticks in the CSSOM while every painted frame shows the end value. `overflow: clip` clips identically without creating a scroll container. The drawer now slides in from its own edge (mirrored under RTL) and the scrim fades up, both on `--duration-medium` and both collapsing under `prefers-reduced-motion`, exactly as the close already did.
+
+@imdreamrunner
--- a/packages/core/src/MobileNav/MobileNav.tsx
+++ b/packages/core/src/MobileNav/MobileNav.tsx
@@ -70,7 +70,16 @@ const styles = stylex.create({
     width: '100vw',
     height: '100dvh',
     backgroundColor: 'transparent',
-    overflow: 'hidden',
+    // `clip`, not `hidden`. Both clip the off-screen drawer, but `hidden` makes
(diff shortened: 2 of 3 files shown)


rubyycheung (approved):

imdreamrunner:
Moving this to draft — **do not merge.** My verification was wrong.

I measured the entry animation with `getComputedStyle(drawer).transform`, which reports the transition running. It does run. But the element never actually moves: `getBoundingClientRect().left` stays pinned at the open position for the whole entry, and a pixel capture confirms the drawer is absent and then simply appears in place.

```
OPEN   computedStyle m41:  320  268  221  179  143  113   88   68   52
       rect.left        :   73   73   73   73   73   73   73   73   73   <- never moves
CLOSE  computedStyle m41:    0   52   99  141  177  207  232  252  268
       rect.left        :   73  125  172  214  250  280  305  325  341   <- tracks it
```

So the bug in the title is real and still unfixed; `@starting-style` on the drawer child is not sufficient. Investigating the actual mechanism now.

PR #5220, opened 2026-08-19
fix(core): tighten the Bottom Sheet grab handle bar to 24px

imdreamrunner:
## Summary

The Bottom Sheet's drag area — the band above the content that holds the grab
pill — was `--spacing-12` (48px) tall. That reads as an empty gap between the
sheet's top edge and its first line of content. This drops it to `--spacing-6`
(24px).

The pill itself is unchanged (`--size-element-lg` × 4px, centered); only the
bar's height moves, so the content gains 24px of the sheet's height budget.

## Changes

- `packages/core/src/BottomSheet/BottomSheetPanel.tsx` — `handleBar.height`:
  `--spacing-12` → `--spacing-6`.

Nothing else keyed off this value: the 48px constants in the sheet
(`OVERSCROLL_PADDING`, `MOBILE_KEYBOARD_BOTTOM_CLEARANCE`, `DETENT_DEDUP_PX`,
`FLICK_MIN_DISTANCE`) are unrelated to the handle, and the panel is a flex
column, so the body absorbs the freed space.

## Test plan

- `pnpm exec vitest run packages/core/src/BottomSheet` — 180 tests pass.
- `pnpm test` — full suite passes except the known `packages/cli`
  load-flake (10 tests time out only in the full run; `vitest run --project
  node packages/cli` passes 2724/2724).
- `pnpm lint:strict` — 0 errors (56 pre-existing warnings).
- `pnpm build` — clean; `packages/core/dist/astryx.css` emits
  `height:var(--spacing-6)` for the handle bar.
- Measured in headless Chromium against the built CSS: the handle bar's
  computed height is `24px`, pill unchanged.

### DIFF
--- a/.changeset/bottom-sheet-handle-24px.md
+++ b/.changeset/bottom-sheet-handle-24px.md
@@ -1,0 +1,12 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Bottom Sheet: tighten the grab handle bar to 24px
+
+The drag area above the sheet's content was 48px tall, which read as an empty
+band between the sheet's top edge and its first line of content. It is now
+24px (`--spacing-6`), keeping the pill centered and giving the content back
+the other 24px of the sheet's height budget.
+
+@imdreamrunner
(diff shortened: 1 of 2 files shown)


imdreamrunner:
Superseded by #5222, which takes the floating approach.

This PR shrank the drag band 48px -> 24px but kept it in the sheet's flex column, so it still cost the content its full height. #5222 makes the bar float instead: it reserves no layout space, so the content actually moves up (a heading goes 57px -> 33px from the sheet's top edge, or 41px -> 17px for bare content). The 24px height from this PR is folded into #5222, and #5222 now targets `main` directly.

PR #5226, opened 2026-08-19
fix(cli): make theming discoverable and its silent failure detectable

josephfarina:
## Summary

Two small changes addressing the same root cause: **theming is hard to discover, and when it goes wrong it fails silently.**

### 1. Point agents at the theme guide, the way we already do for layout

`astryx docs theme` is a thorough 15-section guide — quick start, `defineTheme`, component overrides, custom variants, runtime vs built themes, `astryx theme build`. Nothing in the injected `AGENTS.md` told an agent to open it.

Compare how the block treats the two subjects today. Layout gets a hard rule: *"Frame first: read `astryx docs layout` before writing any page or screen."* Theming gets one clause — *"Brand/accent via `astryx theme`"* — and `astryx theme build` is never mentioned at all. `theme` appears in the `docs <topic>` list, but reachable-on-demand is not the same as prompted, and an agent mid-task will not go looking.

Adds a matching **"Theme first"** rule and lists **`theme add|build`** in the CLI reference alongside `swizzle`. The rule triggers on *restyling the same component twice*, because a single override is legitimate — repetition is the actual signal that something belongs in the theme.

### 2. `astryx theme add` now surfaces the required build step

Following the current next-steps output exactly leaves you on the runtime-injection path with no idea a build step exists — `astryx theme build` is never mentioned.

That matters because the failure mode is invisible: edit a theme, skip the rebuild, and the stale artifact still reports `__built`, so the runtime skips style injection and the app renders the previous theme **with no warning**. The one warning that does exist fires on the *correct* path (runtime injection), so today the developer whose theme works gets nagged and the one whose theme is silently dead gets nothing.

Recommends the `predev`/`prebuild` script wiring rather than a bare command, since remembering to rebuild by hand is precisely the step that gets skipped.

## Deliberately not changed

The `SELF-CHECK` line still prescribes `xstyle` as the remediation, which is arguably the deeper cause of agents avoiding the theme. That line was earned through the prompt-purity vibe tests — it cuts surviving raw CSS in complex runs by roughly 4x — so it should be changed as an A/B through the same harness, not rewritten blind. Tracking separately.

No CI `theme build --check` gate: no built theme artifacts are committed in this repo (only the `.ts` sources are tracked, and `dev`/`build` both run `generate` → `build:theme` first), so there is nothing here to drift. That risk lives in consumer projects, which is what the guidance above addresses.

## Test plan

- [x] `astryx init --features agents` renders the new rule correctly in a scratch project
- [x] `astryx theme add neutral` renders the build guidance with the correct resolved path
- [x] Agent-docs unit tests pass
- [x] `check:changesets` and the full pre-commit gate pass
- [ ] CI golden check (goldens are not tracked in this repo, so verifying on CI)


cixzhang (changes requested):
Thanks — making silent theme failures visible is the right problem, but this head can still give builders confident wrong answers.

- `theme add` shows a source-theme import, then says edits require a build. Source imports are runtime-injected, so the generated JS/CSS are unused. Please present source/runtime and built/SSR as two complete paths.
- `doctor` treats entry-file metadata as proof of freshness, yet later imports that same project module. Changing an imported token left `doctor` green while `theme build --check` exited 1; changing the entry made default `doctor` run a file-writing side effect despite its read-only contract. Keep the default pass non-evaluating and fail closed unless complete build provenance proves freshness.
- Package-wide scans reuse root/first-package state. A wired app hid an unwired sibling's stale theme, while an app-local StyleX compiler still failed. Evaluate and aggregate each finding in its owning package.

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks — this head addresses my previous source/runtime, read-only freshness, and package-scoping findings, but the same promises still have false-green or unusable paths.

- `theme add`'s built-path snippet imports only `./<theme>.css`; it never imports the generated JS theme, so the shown `neutralTheme` is undefined. Print a complete, copyable built import pair.
- The freshness digest marks bare imports and computed `require()` calls complete. Changing a local workspace token package left the digest unchanged and `doctor` reported “in step with source.” Resolve or fingerprint every build input, or report the graph unverifiable.
- Build wiring is still package-wide. With two stale themes and `predev` rebuilding only one, `doctor` said both regenerate before use. Match wiring to each source/output artifact before downgrading stale output to info.

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks — the built import pair is fixed, but builders can still get a green `doctor` while their app renders stale theme output.

- A linked workspace package with version `1.2.3` was fingerprinted as immutable. Changing its token source left the digest unchanged and `doctor` said “in step with source.” A workspace can carry a normal version; hash linked source or report it unverifiable.
- Wiring marks `astryx theme build` with no file as covering every theme even though the real command exits with “missing required argument 'files'.” It also lets `src/brand/theme.ts` cover `src/admin/theme.ts` by basename. Only downgrade staleness when the exact source is actually rebuilt.

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks — this head fixes the prior linked-workspace and exact-source wiring cases, but builders can still get confident wrong guidance or a green diagnostic.

- The injected agent block says every theme edit needs `theme build`. Source-import themes compile and inject at runtime, so edits apply on reload and built artifacts are unused. Keep the source/runtime and built/SSR paths distinct here too.
- `doctor` turns incomplete evidence into success. Its walks silently stop at 10 built themes and 400 stylesheets, while plugin presence is treated as active compiler configuration. I reproduced PASS with an unwired stale 11th theme, PASS with a private-var write in stylesheet 401, and INFO with a swizzled StyleX component whose Babel plugin is installed but not configured. Verify the complete owned set and actual bundler wiring, or report the result as unverifiable.

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks — this head fixes the previous cap reporting and package-scoping cases, but builders can still get confident wrong answers.

- `doctor` skips `dist/`, `build/`, and `out/`, even though the CLI’s own `theme build --out ./dist/ocean.css` example puts built themes there. I put a stale generated theme in `dist/`; `doctor` reported “no built theme output found.” Inspect generated outputs wherever the command can write them, or fail closed when excluded roots may contain them.
- Compiler “wiring” still becomes true from any plugin name in a config, including a comment, while agent docs select StyleX from dependency presence alone. With the plugin installed and only a commented reference, `doctor` returned INFO and generated docs told agents to use `xstyle`. The same block also says source imports reload live, then says every theme edit needs a rebuild. Only claim configured paths and keep source/built guidance consistent.

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks — this head fixes the common `dist/` and commented-plugin cases, but builders can still get confident wrong answers.

- `doctor` still excludes `.next`, `.turbo`, and `.cache` even though `theme build --out` accepts those paths. A stale generated theme under `.next` returned “no built theme output found.” Inspect every allowed output location or report skipped roots as unverifiable.
- `doctor` and agent docs separately treat any live plugin string as active wiring. With `plugins: []` and an unused `const p = '@stylexjs/babel-plugin'`, `doctor` returned INFO and generated docs selected `xstyle`. Use one detector that proves a supported config shape or returns unverifiable.
- The same generated block says source-theme edits apply on reload, then says every theme edit does nothing until rebuilt. Limit that command-index warning to built imports.

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks — this head rebases cleanly and fixes several earlier cases, but the same promises still have false-green paths.

- `doctor` still skips valid output roots such as `.next`, so a stale generated theme there reports “no built theme output found.” Its ten-theme cap also returns INFO before checking truncation when the first ten artifacts self-heal; an unwired stale eleventh theme is never examined. Please make coverage exhaustive or surface the skipped set as unverifiable before returning success.
- StyleX wiring still has two detectors with different supported plugins. With a configured `@stylexjs/webpack-plugin`, `doctor` reports the swizzle as wired while generated agent docs select plain CSS. Use one shared detector so diagnosis and guidance cannot drift.
- The generated block still says source imports reload live, then the command index says every theme edit requires rebuilding. Limit that line to built imports.

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks — the source/built guidance, `.next` coverage, and shared caller list are fixed, but builders can still get confident wrong answers.

- `isStyleXConfigured` treats any live string literal naming a declared plugin as wiring. With `plugins: []` and an unused `const plugin = '@stylexjs/babel-plugin'`, both `doctor` and generated agent docs report StyleX as active, so swizzled components still render unstyled silently. Please require evidence that the declared plugin participates in the active config, or report it as unverifiable.
- `checkThemeBuilt` returns “no built theme output found” before checking `builtTruncated`. When the 4,000-directory bound is reached before the first generated CSS, the scan reports `truncated: true` and zero themes, but `doctor` emits INFO. Please check truncation before the empty result so incomplete coverage never looks clean.

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks — the source/built guidance is now clear, but this head still lets silent failures look healthy.

- `readSpecifiers` now blanks whole template literals, including executable `${...}` expressions. A theme using `${require('./tokens')}` produced `complete: true`; changing `tokens` left the digest unchanged. Imports in interpolations must affect the digest or make the graph unverifiable.
- `isStyleXConfigured` still accepts any live quoted plugin name. With `plugins: []` plus an unused constant, generated docs recommend `xstyle` and `doctor` clears an unstyled build.
- `checkThemeBuilt` returns “no built theme output found” before checking truncation. A 4,001-directory scan returned `{themes: [], truncated: true}` and INFO. Truncation must warn even when zero artifacts were reached.

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks — this head improves StyleX counting and comment handling, but my prior false-green cases still reproduce at current bytes.

- Input discovery blanks executable template interpolations. `${require('./tokens')}` produced `complete: true`; changing `tokens` left the digest unchanged, so `doctor` can report a stale theme as current.
- The built-theme scan still returns “no built theme output found” before checking truncation. A 4,001-directory project with no artifact in the visited prefix returned INFO although the remainder was unexamined.
- StyleX wiring treats any call of an imported plugin as configured. With `const unused = stylex(); export default {plugins: []}`, `doctor` returned INFO and generated docs recommended `xstyle`, although the app compiles no StyleX.

Please fail closed unless each dependency, output root, and active compiler path is actually verified.

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks — the old-head cases are fixed, but current bytes still turn unproved parsing into healthy answers.

- A regex backtick before a later template hides a real token import; the graph reports complete and its digest does not change with that token.
- An unused object's `plugins: [stylex()]` makes `doctor` report INFO and agent docs recommend `xstyle`, even when the exported config has `plugins: []`.
- `false && pnpm build:theme` makes stale output look self-healing.

All three come from treating text as proof of active JavaScript or shell behavior. Please use syntax-aware discovery, or fail closed whenever active inputs, compiler wiring, or rebuild execution cannot be proved.

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks — the AST rewrite fixes the old-head cases, but current bytes still treat unproved execution as healthy.

- `isThemeBuildWired` follows sibling scripts without preserving the caller’s shell gate. `false && pnpm build:theme` still returns wired, so stale output is called self-healing although no rebuild runs.
- Theme input discovery misses valid `import x = require('./tokens')`; it reports the graph complete and the digest stays unchanged when that token file changes.
- StyleX discovery accepts a serve-only return branch or an unrelated nested `plugins` key. Doctor reports INFO and generated docs recommend `xstyle` although production compiles none.

Please return healthy only when the active input, build, and compiler paths are proven; otherwise report them as unverifiable.

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks — the previous shell-gate and import-equals cases are fixed, but the current head still gives builders healthy answers for broken paths.

- A theme can load a local token with a `createRequire()` alias; editing that token left the recorded digest unchanged, so `doctor` passed while `theme build --check` failed.
- A function config with StyleX only in its `serve` return is treated as production wiring. `doctor` reports INFO and generated agent docs recommend `xstyle`, though the production branch has no compiler.
- `echo astryx theme build src/theme.ts` counts as executable build wiring, so stale output is downgraded to self-healing INFO even though nothing rebuilds it.

Please return healthy only when each active input, production compiler path, and rebuild path is proved; otherwise report the result as unverifiable.

[Reviewed by Robohands]

PR #5230, opened 2026-08-19
feat(useLayer): size a layer to the space its anchor leaves it

cixzhang:
## The gap

A layer that does not fit can only **move**. `getPositionTryFallbacks` emits
`flip-block` / `flip-inline` plus span-based slides, so the browser will put
the layer on the other side of its anchor or slide it along the alignment
axis — but nothing ever **sizes** it. A layer carrying a long list in a short
viewport is taller than every fallback, so none of them fit, the browser keeps
the original placement, and the layer runs off the screen at its full height.

The number itself was already reachable, and the PR is honest about that:
`style` merges last, so a caller who knows to write `max-block-size: 100%`
gets the same clamp today. What nobody gets right by hand is the rest of it —
see the two details below, both of which silently break the naive version.

Layer-bearing components show the shape of the gap by guessing instead:
`DropdownMenu`, `DropdownMenuSubMenu`, `ContextMenu` and `Selector` all cap at
a hardcoded `300px`, `ComplexSelector` at `min(480px, calc(100vh - 32px))`. A
fixed cap is wrong in both directions — it amputates the list on a tall window
and still overflows on a short one — and none of them cap the inline axis at
all.

## The change

`clampToAvailableSpace?: 'block' | 'inline' | 'both'` on the context render
props.

```tsx
{layer.render(
  <div style={{maxBlockSize: '100%', overflow: 'auto'}}>{items}</div>,
  {placement: 'below', offset: 8, clampToAvailableSpace: 'block'},
)}
```

**No measurement.** `position-area` already makes the anchor's cell the
layer's containing block, so `100%` along an axis *is* the distance from the
anchor to the viewport edge. The clamp is three CSS declarations; there is no
resize observer, no scroll listener, no extra render pass, and the browser
re-resolves it wherever a fallback puts the layer.

Two details separate this from a clamp that only half works. Both were found
by measuring in Chromium, not by reading the spec:

- **The fallbacks have to be ordered by size.** A clamped layer always fits, so
  "first option that fits" degenerates to "never flip" and the layer keeps a
  cramped side: measured 172px taken below while 300px sat free above.
  `position-try-order: most-block-size` (or `most-inline-size`) on the
  placement axis restores the choice, and only the placement axis is flipped,
  so it is set only when that axis is clamped.
- **The layer has to become a flex column** when the block axis is clamped, and
  that column has to be scoped to `:popover-open`. The clamp bounds the layer
  box, but a percentage size on the content inside only resolves against a
  definite parent — without the column the content keeps its full height and
  paints straight out of the box. Set inline, though, `display` beats the UA's
  `[popover]:not(:popover-open) {display: none}` and a *closed* layer paints
  its whole list on the page; it lives in a StyleX rule for that reason.

`offset` puts a margin on both edges of the placement axis, outside the clamped
box, so it comes out of that axis's budget; otherwise the layer clears the
viewport edge by exactly the offset.

Off by default — nothing changes for any existing caller — and ignored under
`positioning: 'custom'`, which has no cell to measure against and would size to
the whole viewport.

## Verification

Real Chromium against the two new stories, measuring the layer's box against
the viewport. 40 rows, 1280px of content, `offset: 8` so the budget is the room
minus 16px.

Viewport sweep, trigger fixed, room below is the short side:

| viewport | unclamped | clamped | room on chosen side |
|---|---|---|---|
| 320 | 1280px, **968px off-screen** | 160px, 0 off-screen | 176 above |
| 420 | 1280px, **868px off-screen** | 196px, 0 off-screen | 212 below |
| 600 | 1280px, **688px off-screen** | 376px, 0 off-screen | 392 below |
| 900 | 1280px, **388px off-screen** | 676px, 0 off-screen | 692 below |

Both flip directions, and the clamp lands on the budget exactly:

| room above / below | side chosen | layer height | budget |
|---|---|---|---|
| 176 / 112 | **above** | 160px | 176 − 16 |
| 176 / 292 | below | 276px | 292 − 16 |
| 176 / 692 | below | 676px | 692 − 16 |

Content shorter than the space is left alone: 4 rows measures 128px in both
stories, identical, no scrollbar. Content longer scrolls inside the clamp
(`scrollHeight` 1280 in a 676px box).

Scroll positions, reopened at each offset — exact every time:

| scrollY | room below | layer height |
|---|---|---|
| 0 | 292 | 276px |
| 60 | 352 | 336px |
| 120 | 412 | 396px |
| 180 | 472 | 456px |

One honest limitation, and it is the one anchor positioning already has: a
layer **left open** while the page scrolls keeps the size it opened with, the
same way it keeps the side it opened on. I confirmed the existing behaviour
first — an unclamped layer does not re-run its fallbacks on scroll either
(measured: still `above` and 144px off the top after scrolling, with 322px free
below). The clamp is resolved at layout time, exactly like the placement it
sits next to; it is not a new class of staleness.

### Reproducing it

Two stories, `Core/Layer` → *Available Space Unclamped* and *Available Space
Clamped*, same content and same trigger. Shrink the Storybook viewport to
~320px: unclamped, the layer covers the trigger and runs off both edges;
clamped, it flips above the trigger and ends inside the viewport. The row
count is adjustable so you can walk the list past the available space and
back.

## Tests

Unit tests for the emitted styles, covering each axis, the try-order rule on
the placement axis only, the offset arithmetic, the two cases that emit
nothing (unset, and `positioning: 'custom'`), and that no inline `display` is
emitted. 52 tests in `useLayer.test.tsx`; 558 across every layer-bearing
component still pass. Closed → open → closed → open measured in Chromium:
`display` is `none`/`flex`/`none`/`flex` and the height `0`/`276`/`0`/`276`.

Moving `DropdownMenu`, `ContextMenu`, `Selector` or `ComplexSelector` off their
hardcoded caps. Each is a visible behaviour change to a shipped component and
deserves its own PR and its own screenshots; this one only opens the seam.

Also out of scope: clamping to a boundary **other than the viewport**. A layer
anchored inside a scrollable column arguably wants to clamp to that column, and
no amount of CSS gets there — the layer is in the top layer, so its containing
block is the viewport. That needs a different mechanism.


packages/core/src/Layer/useLayer.tsx
```diff
@@ -128,6 +140,42 @@ export interface ContextRenderProps {
...
+   * positioning it resolves at layout time, not continuously: a layer left
+   * open while the page scrolls keeps the size it opened with, exactly as it
+   * keeps the side it opened on.
+   *
+   * The prop exists because the same three declarations written by hand get
+   * two things wrong. The fallbacks have to be ordered by size on the
+   * placement axis, or a layer that always fits stops flipping and keeps a
+   * cramped side; and the layer has to become a flex column, only while open,
+   * for content to size against the clamp. Give that content
+   * `maxBlockSize: '100%'` and its own `overflow` to scroll it:
```
cixzhang, line 179: Might be worth dropping the `maxBlockSize` half: `overflow` is the one that does it.

cixzhang (commented):
Nice. One thing: the content's `overflow` is what clamps it, not the `maxBlockSize: '100%'` in the snippet; trim `overflow` and the list paints straight back out. Approve once you undraft.

with the content's overflow

without it

[Reviewed by Robohands]

cixzhang (commented):
Thanks—the clamp mechanics are solid in Chromium. The blocker is ownership: builders would get two places to control viewport fit, while the [current Layer contract](https://github.com/facebook/astryx/blob/33b2b7def41cef93bb991fb3e057569c90f2799a/docs/architecture/layer-runtime.md#L87-L107) keeps sizing and overflow in Popover or the component that owns the surface. Could we keep this primitive internal to those owners instead of making it public Layer API?

[Reviewed by Robohands]

cixzhang (commented):

Builders would get two public ways to control viewport fit, with different owners and guarantees. This head only merges `main`; the Layer delta is unchanged. Current API policy says [the existing styling seams come first](https://github.com/facebook/astryx/blob/9da5ff9cc8ee94ac38f01b18f9659ea719377c3d/docs/specs/AST-002/spec.md#L59-L62), and the [Layer contract keeps viewport sizing and overflow with Popover or the surface owner](https://github.com/facebook/astryx/blob/9da5ff9cc8ee94ac38f01b18f9659ea719377c3d/docs/architecture/layer-runtime.md#L106-L107). Please keep this available-space primitive internal to those owners instead of exporting `clampToAvailableSpace`.

[Reviewed by Robohands]

cixzhang:

PR #5250, opened 2026-08-20
fix(field): pass shade="inherit" to Spinner in input components and Switch

athz:
## Summary

Pass `shade="inherit"` to `<Spinner>` in all input-family components and `Switch`.

## Problem

Spinner's default shade reads `themeTokens["--color-accent"]` from `useTheme()` — a global JS object — then paints the ring imperatively onto a `<canvas>` via `context.strokeStyle`. Nothing in CSS can reach that: not a token remap, not a `components` entry, not `!important`. The result is a hardcoded accent-blue spinner that ignores theme overrides entirely.

## Solution

`shade="inherit"` is the only code path that reads the CSS cascade (`getComputedStyle(canvas).color`), making the spinner respect its parent's `color` context. This is already the pattern used by `Button` (line 703).

## Affected components

| Component | Lines |
|-----------|-------|
| TextInput | 462 |
| TextArea | 628 |
| DateInput | 773 |
| Selector | 1517 |
| FileInput | 682, 708 |
| Switch | 622 |

## Why this matters for theming

With `shade="inherit"`, downstream consumers can control spinner color through normal CSS inheritance — set `color` on a parent and the spinner follows. Without it, the only lever is the global `--color-accent` JS token, which is not per-component configurable.

### DIFF
--- a/.changeset/spinner-inherit-shade.md
+++ b/.changeset/spinner-inherit-shade.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Pass `shade="inherit"` to Spinner in input components and Switch so the loading indicator respects the parent color context and is themeable via CSS.
+@athz
--- a/packages/core/src/DateInput/DateInput.tsx
+++ b/packages/core/src/DateInput/DateInput.tsx
@@ -770,7 +770,7 @@ export function DateInput({
           iconClassName={stableClassName('date-input-clear-icon')}
         />
       )}
-      {isBusy && <Spinner size="sm" />}
+      {isBusy && <Spinner size="sm" shade="inherit" />}
       {statusIcon}
       {popover.render(
         <Calendar
--- a/packages/core/src/FileInput/FileInput.tsx
+++ b/packages/core/src/FileInput/FileInput.tsx
@@ -679,7 +679,7 @@ export function FileInput({
 
   const renderDropzoneContent = () => {
     if (isLoading) {
-      return <Spinner size="md" />;
+      return <Spinner size="md" shade="inherit" />;
     }
     if (hasFiles) {
       return (
@@ -705,7 +705,7 @@ export function FileInput({
           <span {...stylex.props(styles.fileNameText)}>
             {fileNames ?? displayPlaceholder}
           </span>
-          <Spinner size="sm" />
+          <Spinner size="sm" shade="inherit" />
         </>
       );
     }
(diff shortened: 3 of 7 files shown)


cixzhang:
Rebase needed — [#5408](https://github.com/facebook/astryx/pull/5408) rewrote the Spinner's ring as SVG since you opened this.

Worth saying what that does and doesn't change for you. It **doesn't** fix the thing this PR runs into: `inherit` now means `stroke: currentColor`, which resolves against the same cascade, so at a host that paints its own background but sets no `color`, the ring still paints the ancestor's colour on the input's own surface. Put one in a section with `color: #fff` and TextInput, TextArea, DateInput, both FileInputs and Switch show nothing.

It **does** remove the performance cost — the old canvas read the computed colour per spinner; there's no read at all now.

So the ask is unchanged and it's now the house convention rather than my opinion: **set `color` on each host you touch, then keep `inherit`.** A container that paints its own background owns its `color`; `Button` and `CheckIndicator` do this already, and we're applying the same rule to Table's sort affordance in [#5420](https://github.com/facebook/astryx/pull/5420).

`inputWrapperStyles.base` (`Field/inputStyles.stylex.ts:33`) is shared chrome for eleven of the twelve input hosts, so one `color` there covers most of it; Switch's thumb and FileInput's two modes need their own.

[Reviewed by Robohands]

cixzhang:
Closing this one, and I want to be clear it isn't because the observation was wrong — the spinner genuinely does paint the ancestor's colour on a host that owns its own surface, and you found that.

What changed is where the fix lives. [#5408](https://github.com/facebook/astryx/pull/5408) landed the ring as SVG, so `inherit` is now `stroke: currentColor`, and we've settled on taking spinner theming through [#5214](https://github.com/facebook/astryx/pull/5214) rather than per-host props. `shade="inherit"` on its own doesn't close the case anyway — the hosts still set no `color`, so the ring keeps resolving against the same cascade it does today. The remaining half is the `color` on each host, and that's the theming work rather than a prop pass-through.

Sorry for the round trip; this is ours arriving mid-review, not you. Your six call sites are the map of what still needs a `color`, and I'd rather have found this from you than not.

[Reviewed by Robohands]

cixzhang:
One correction to what I wrote when I closed this, so you're not left with a wrong instruction.

I said the remaining work was setting `color` on each host. It isn't. We've settled on the spinner exposing explicit theme vars — `--spinner-color` and `--spinner-track-color` — precisely because `currentColor` only resolves to something useful if the host happens to set `color`, and as you found, ours largely don't. That's not a contract a theme author can rely on.

Which makes your finding sharper than my reply gave it credit for: the six call sites you touched are the evidence that the implicit-cascade approach doesn't hold, and that's the argument for the explicit seam. It's landing in [#5214](https://github.com/facebook/astryx/pull/5214).

Still closed, since the mechanism moved rather than the problem. Thanks for it.

[Reviewed by Robohands]

PR #5255, opened 2026-08-20
feat(core)!: Banner content is visible by default; collapse moves to a `collapsible` prop

freddymeta:
## The gap

Banner infers its disclosure from its content: any `children` get a chevron in
the header and are hidden until it is pressed. There is no way to have content
and no toggle — which is the case a banner most often wants, the list of the
three fields that failed validation — and `defaultIsExpanded` is the only knob,
with no controlled mode.

## The shape

Adding an `isCollapsible` boolean beside `defaultIsExpanded` would put two props
on one axis, where one silently voids the other — the prop-independence
principle, and precisely the clutter [API Conventions § Boolean-or-Config
Props](https://github.com/facebook/astryx/wiki/API-Conventions#boolean-or-config-props)
names (`isCollapsible` + `defaultIsCollapsed` + `onCollapsedChange`). So the
whole axis becomes one prop, the shape `SideNav.collapsible` set:

```ts
/** @default true */
collapsible?: boolean | CollapsibleConfig;
```

```tsx
<Banner status="warning" title="Config changed">…</Banner>          // unchanged: collapsible, starts closed
<Banner collapsible={false}>…</Banner>                              // new: always visible, no toggle
<Banner collapsible={{defaultIsOpen: true}}>…</Banner>              // replaces defaultIsExpanded
<Banner collapsible={{isOpen, onOpenChange}}>…</Banner>             // new: controlled
```

**The default is unchanged.** `collapsible` defaults to `true`, and Banner hands
the hook `defaultIsOpen: false` — the one place it departs from `useCollapsible`,
which opens by default. A banner's message lives in its header and its content is
supplementary, and that is the behaviour Banner has always had.

Three deliberate choices:

- **`CollapsibleConfig`, not a Banner-only vocabulary.** The exported type from
  `useCollapsible` — so a banner's disclosure is configured exactly like
  `Collapsible`'s, and there is no third spelling of the same three fields.
- **The shared hook, not local `useState`.** Banner no longer owns collapse
  state at all (Use the System, rule 3). Controlled mode comes for free, which
  is why it is in this PR rather than a later one.
- **The default lives on the prop, not in a second prop.** `collapsible={false}`
  is the opt-out; there is nothing else on the axis to contradict it.

Lint-wise the `boolean | Config` union is exempt from
`@astryx/boolean-prop-naming`, so the unprefixed name is legal; a bare
`isCollapsible: boolean` would not have been.

With `collapsible={false}` there is no toggle, no `aria-expanded` /
`aria-controls`, and the content region stays mounted with no `id`.

## Evidence

Real builds, shot in Chromium — `before` from `main` at 5ab06259c, `after` from
this branch. Harness and DOM facts on the
[assets branch](https://github.com/facebook/astryx/tree/assets/banner-collapsible-content)
(delete it after merge).

**Nothing moves for existing code.** A banner with children and no collapse prop,
and one that starts open, are **byte-identical PNGs across the two builds**
(`dfbfe0a7…` and `147e808b…`):

| default (children, no prop) | starts open |
|---|---|
| default | open |

**`collapsible={false}` is the new capability.** Before, that content was
unreachable behind a chevron; after, it is simply there.

| before | after |
|---|---|
| before | after |

Read off the DOM: before `{buttons: ["Expand"], listItems: 0, ariaExpanded:
["false"]}`, after `{buttons: [], listItems: 3, ariaExpanded: []}`.

## What breaks, and the codemod

Only the prop rename: `defaultIsExpanded` is gone in favour of the config. That
is a type error at every call site, not a silent behaviour change — nothing
changes under anyone without a compiler error first.

**Codemod:** `npx astryx upgrade --codemod banner-collapsible-content`
(staged in `transforms/next/` per [Release Process § Writing a
codemod](https://github.com/facebook/astryx/wiki/Release-Process#writing-a-codemod)).

| before | after |
|---|---|
| `defaultIsExpanded` / `={true}` | `collapsible={{defaultIsOpen: true}}` |
| `defaultIsExpanded={false}` | *(removed — it is the default)* |
| `defaultIsExpanded={expr}` | `collapsible={{defaultIsOpen: expr}}` |
| children, no prop | *(untouched)* |

It only fires on a `Banner` imported from `@astryxdesign/core` —
`defaultIsExpanded` is a `ChatToolCalls` prop too — and declines to guess around
a spread. One deliberate scope boundary: the prop inside a props object (a
Storybook `args`) is left alone rather than rewritten on a guess about which
component the object feeds; the removed prop makes those a type error, so the
compiler finds them. `verify-codemods` passes as a superset — Storybook's Banner
stories are this transform's own output plus the new opt-out demos, with the
`args` sites migrated by hand.

## Accessibility note

Content that is always visible inside `role="alert"` means an error banner
announces its detail along with its title. Usually what you want for a list of
validation errors; worth knowing when reaching for `collapsible={false}`. The
`Banner.doc.mjs` best practices now say to use it when the user needs the content
to act on the message, and to keep the toggle when the detail is long enough to
bury the banner's own message.

## Test plan

- `pnpm test` — 536 files / 10946 tests pass. Banner is 47 tests (8 new or
  rewritten: default-unchanged, explicit `true`, `collapsible={false}`, the
  disclosure wiring absent when opted out, `defaultIsOpen: true`,
  `onOpenChange`, controlled). Codemod is 12 tests.
- `pnpm lint:strict` — 0 errors (56 pre-existing warnings, none in touched files).
- `pnpm build` — clean. `pnpm -F @astryxdesign/core typecheck` — clean.
- `node .github/scripts/codemod-verify.mjs` — reproducible.
- Chromium before/after above.

Also updated: `Banner.doc.mjs` (en + zh + dense, props, anatomy, best practices),
stories, and the `BannerCollapsibleContent` showcase block.

### DIFF
--- a/.changeset/banner-collapsible-content.md
+++ b/.changeset/banner-collapsible-content.md
@@ -25,14 +25,20 @@ boolean-or-config convention `SideNav.collapsible` set, and backed by the shared
 **The default is unchanged** — a banner that never mentioned `defaultIsExpanded`
 behaves exactly as it did. The breaking part is the prop itself:
 `defaultIsExpanded` is removed in favour of the config, which is a type error at
-every call site rather than a silent change.
+every JSX call site that names it.
 
 **Codemod:** `npx astryx upgrade --codemod banner-collapsible-content`
 
 It rewrites `defaultIsExpanded` to `collapsible={{defaultIsOpen: true}}` and
 drops `defaultIsExpanded={false}`, which is now the default. Banners that never
-set the prop are left alone. `defaultIsExpanded` inside a props object (a
-Storybook `args`, for instance) is out of the transform's scope — the removed
-prop makes those a type error, so the compiler finds them.
+set the prop are left alone.
+
+**One case the codemod and the compiler both miss: a spread.**
+`defaultIsExpanded` inside a props object is out of the transform's scope. A
+props object in a typed position still fails to compile — but an inferred one
+that is spread, `<Banner {...args} />`, does not, because TypeScript does not
+excess-property-check a spread. The prop then falls through to the DOM and the
+banner quietly starts collapsed. **Grep for `defaultIsExpanded` after running
+the codemod** and migrate any spread sites by hand.
 
 @freddymeta
--- a/.changeset/banner-collapsible-content.md
+++ b/.changeset/banner-collapsible-content.md
@@ -26,8 +26,9 @@ boolean-or-config convention `SideNav.collapsible` set, and backed by the shared
 render instead of hiding. Run `astryx upgrade` — the
 `banner-collapsible-content` codemod rewrites `defaultIsExpanded` to the
 equivalent config and adds `collapsible={{defaultIsOpen: false}}` to banners
-that relied on the implicit collapse, so behaviour is preserved; those sites get
-a TODO comment, since always-visible content is usually what they want.
+that relied on the implicit collapse, so behaviour is preserved. Those marked
+collapsed are the ones worth a second look afterwards: content that is simply
+visible is usually what a banner wants, and is now one prop shorter to write.
 
 Note for a banner whose content is now always visible: `role="alert"` announces
 the whole region, so an error banner reads its detail out along with its title.
--- a/.changeset/banner-collapsible-content.md
+++ b/.changeset/banner-collapsible-content.md
@@ -1,0 +1,35 @@
+---
+'@astryxdesign/core': minor
+'@astryxdesign/cli': minor
+---
+
+[breaking] Banner: children are visible by default, and the collapse axis moves onto one `collapsible` prop
+
+Banner inferred its disclosure from its content: any `children` got a chevron in
+the header and were hidden until it was pressed. There was no way to show
+content without a toggle — the case a banner most often wants, a list of the
+three fields that failed validation — and `defaultIsExpanded` was the only knob,
+with no controlled mode.
+
+The whole axis is now one `boolean | CollapsibleConfig` prop, following the
+boolean-or-config convention `SideNav.collapsible` set, and backed by the shared
+`useCollapsible` hook rather than Banner's own state:
+
+```tsx
+<Banner status="error" title="3 fields need attention">…</Banner>  // always visible
+<Banner collapsible>…</Banner>                                     // collapsible, starts open
+<Banner collapsible={{defaultIsOpen: false}}>…</Banner>            // starts collapsed
+<Banner collapsible={{isOpen, onOpenChange}}>…</Banner>            // controlled (new)
+```
+
+`defaultIsExpanded` is removed, and children with no `collapsible` prop now
+render instead of hiding. Run `astryx upgrade` — the
+`banner-collapsible-content` codemod rewrites `defaultIsExpanded` to the
+equivalent config and adds `collapsible={{defaultIsOpen: false}}` to banners
+that relied on the implicit collapse, so behaviour is preserved; those sites get
+a TODO comment, since always-visible content is usually what they want.
+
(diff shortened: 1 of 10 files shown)


cixzhang (changes requested):
This is the better pattern but I think we should avoid breakage. Could we:

* keep an internal default for collapsible
* allow users to pass collapsible={false} or collapsible={<config>} in order to customize it?

freddymeta:
Done — the default is back, and the only thing left breaking is the prop name.

`collapsible` now defaults to `true`, and Banner hands the shared hook
`defaultIsOpen: false`. That is the one point where it departs from
`useCollapsible`, which opens by default: a banner's message lives in its header
and the content is supplementary, so starting closed is right for Banner and is
what it has always done. A call site that never mentioned the collapse axis is
untouched — I re-shot both builds and the default case and the starts-open case
are **byte-identical PNGs** (`dfbfe0a7…`, `147e808b…`), not just visually close.

So the axis reads:

```tsx
<Banner status="warning" title="Config changed">…</Banner>   // unchanged
<Banner collapsible={false}>…</Banner>                       // content, no toggle
<Banner collapsible={{defaultIsOpen: true}}>…</Banner>       // was defaultIsExpanded
<Banner collapsible={{isOpen, onOpenChange}}>…</Banner>      // controlled
```

`defaultIsExpanded` still goes, though — I did not keep it as a deprecated
alias, and that is the one place I have taken a position rather than followed
yours. Two props on the same axis is the thing this shape exists to remove, and
this residue is a type error at every call site, so nobody's UI changes under
them without the compiler saying so first. The codemod covers it
(`npx astryx upgrade --codemod banner-collapsible-content`): `defaultIsExpanded`
→ `collapsible={{defaultIsOpen: true}}`, `defaultIsExpanded={false}` dropped
since it is the default, and banners that never set it are left alone. Say the
word if you would rather have a deprecation window and I will add it back as
`@deprecated` with a documented precedence.

One scope boundary worth flagging: the codemod is JSX-only. `defaultIsExpanded`
inside a props object — a Storybook `args`, say — is left alone rather than
rewritten on a guess about which component the object feeds (`ChatToolCalls` has
a prop of the same name). Removing the prop makes those a type error, so they
are found, just by tsc rather than by the transform. Our own stories had three
such sites; I migrated them by hand and `verify-codemods` passes as a superset.

cixzhang:
Heads up on timing: we're cutting 0.5.0 Monday and that's the release we're landing breaking changes in, so I'm holding this until then rather than merging it now. Nothing needed from you — it's queued and it'll go in ahead of the cut.

Reviewing it properly before then. One thing you could fix meanwhile if you like: `collapsible={null}` throws and takes the banner with it (`typeof null === 'object'`), so it reads as a config and we grab `.defaultIsOpen` off it. TypeScript blocks it, so it only bites from JS or a value widened to `| null`. One clause.

[Reviewed by Robohands]

cixzhang (approved):
Thanks, this is the right shape. The default really is untouched — byte-identical frames closed and expanded. I pushed the `collapsible={null}` guard so it lands with the cut.

| | before | after |
|---|---|---|
| children, no collapse prop |  |  |
| the same one, expanded |  |  |
| `collapsible={false}` |  |  |

[Reviewed by Robohands]

cixzhang:
One more thing I changed, since it's a claim rather than code: `<Banner {...args} />` with `defaultIsExpanded` in the object compiles clean — TypeScript doesn't excess-property-check a spread — so the prop reaches the DOM and the banner quietly starts collapsed. The changeset promised the compiler catches those; it now says to grep. Your JSX-only boundary was the right call, the promise around it was just too strong.

[Reviewed by Robohands]

PR #5256, opened 2026-08-20
fix(core): re-clamp the held resizable size when its bounds move

cixzhang:
## Problem

`useResizable` already re-clamps a held pixel size when live minimum or maximum bounds move under it, but the current clamp path also re-applies `snaps` to a size that is still legal. A hard bound can leave a panel at an off-snap size; widening the band then moves it again even though the person made no new choice.

This is visible with `[100, 300, 500]`: a maximum of 250 correctly forces 500 → 250, but widening the maximum back to 600 must keep 250 rather than silently moving the panel to 300.

## Solution

The selected size is corrected only when it falls outside the current band:

- a moving bound makes the held size illegal → clamp it into the band, applying snap preference within that hard constraint;
- the held size remains legal → preserve it exactly, including an off-snap value;
- bounds move while the region is collapsed → preserve the hidden expanded choice and correct it only when expansion makes that choice active;
- a persisted legal off-snap choice → restore it exactly.

Bounds-only corrections remain render-time state reconciliation, so an illegal size is never painted or announced. They do **not** fire `onSizeChange`: current [AST-010](https://github.com/facebook/astryx/blob/main/docs/specs/AST-010/spec.md) defines basis and bounds-only correction as layout reconciliation, not a user interaction.

## Impact

Panels whose live bounds narrow still stay within their effective geometry. Widening a band no longer moves a legal panel, and a temporary bounds change while collapsed no longer overwrites the expanded choice the person left behind.

## Risk

The change is limited to live bound reconciliation and restore paths. Pointer, keyboard, programmatic `resize()`, initialization, and explicit snap interactions still use the existing full snap behavior. Collapsed visible size remains zero, and persistence keeps the latent expanded pixel size.

## Verification

- 154 focused `useResizable` and `ResizeHandle` tests pass.
- Core source and docs typechecks pass; Core builds successfully.
- Real Chromium on the existing Structured Percent Sizing story:
  - maximum: 100px → 50px when tightened → remains 50px when widened to 100px;
  - minimum: 333px → 399px when raised → remains 399px when relaxed to 333px;
  - rendered panel width, selected size, and resolved bound agree in every state.
- Public-repository guard and repository checks pass.


cixzhang (commented):
Thanks — Risk reads right now, including the DerivedMax callout. Nothing else from me; it's an approve once you undraft.

[Reviewed by Robohands]

cixzhang (commented):
Thanks — Risk reads right now, including the DerivedMax callout. Nothing else
from me; it's an approve once you undraft.

[Reviewed by Robohands]

cixzhang (commented):
Thanks — approve on the current contract. Bounds correct only illegal sizes, legal and off-snap choices survive widening, and collapse, persistence, callbacks, and dynamic snaps remain intact.

[Reviewed by Robohands]

cixzhang:

PR #5257, opened 2026-08-20
RadioListItem and CheckboxListItem accept rich label and description content

cixzhang:
## Why

`RadioListItem` types `label` and `description` as `string`. Its sibling `CheckboxListItem` already types `label` as `ReactNode`, so radio and checkbox disagree about the same slot — and an app whose option descriptions carry a link can type one on neither.

The runtime was never the obstacle. `Item` and `ListItem` have taken `ReactNode` for both slots all along, so a rich node already rendered and the label element already named the radio from its own text. Only the types refused.

## What

`label` on `RadioListItem`, and `description` on both items, are now `ReactNode`. `RadioListItem` picks up the `aria-label` escape hatch `CheckboxListItem` established, with the same meaning: a plain-text accessible name applied to the control, not the row.

The two controls reach that name by different routes, and the prop docs say so rather than pretending they match:

- `CheckboxListItem`'s checkbox carries its own visually hidden label, so a rich `label` leaves it named "Checkbox" — it warns in dev until you pass `aria-label`.
- `RadioListItem`'s `<label htmlFor>` wraps the visible content and is tied to the input, so a rich label still computes a name from its text. `aria-label` narrows a name that reads badly, rather than rescuing a missing one, so there is no dev warning here — it would fire on labels that are already named correctly.

Rich labels may include links or other interactive content. The visual label renders in a plain `<span>`, while the row delegates only non-interactive surface clicks to the control. Nested links and buttons therefore keep their own behavior without selecting or toggling the item. The radio still derives its accessible name from the rich label text; pass `aria-label` when that flattened text is absent or reads poorly.

**Left alone deliberately.** `RadioList` and `CheckboxList` keep `label: string` and `description?: string`. Those are the group's field-level label and helper text, owned by `Field` and shared with every input that composes it; widening them is a change to `Field`'s contract alongside `isLabelHidden` and `labelTooltip`, with its own accessible-name story. Different change.

## Risk

One behavior change: `aria-label` on `RadioListItem` used to land on the row `<div>`, where ARIA ignores it on a role-less element. It now names the radio. Every other `aria-*` still forwards to the row, matching `CheckboxListItem`.

Widening `description` also made a latent guard mismatch reachable, so it is fixed here: the description element was guarded by `!= null` while its `aria-describedby` used truthiness, and the ordinary `flag && <span>…</span>` idiom yields `false` — enough to render an empty described-by target. Both now go through `isRenderable`.

## Testing

Core suite (7057 tests), `typecheck`, `typecheck:docs` and lint all green. New unit tests cover the name computed from a rich label, `aria-label` overriding it, non-interactive rich-label content still selecting on click, links inside labels and descriptions remaining separate keyboard targets without selecting or toggling the item, and an empty description leaving `aria-describedby` off.

jsdom cannot settle accessible names, so the a11y half was read out of Chromium's own accessibility tree against the new stories:

| Control | Accessible name | Accessible description |
|---|---|---|
| radio, label `Pro <Badge label="Popular" />`, no `aria-label` | `Pro Popular` | — |
| same, with `aria-label="Pro"` | `Pro` | `$12 per seat. See what is included` |
| radio, string label, rich description | `Starter` | `Free forever. Compare plans` |
| checkbox, rich description | `Personalization` | — |

Interaction in the same browser: clicking a rich label selects its radio; clicking a link inside a description selects nothing, on radio and checkbox alike, while a click on the rest of the checkbox row still toggles it.

### DIFF
--- a/.changeset/radio-checkbox-list-item-rich-content.md
+++ b/.changeset/radio-checkbox-list-item-rich-content.md
@@ -1,0 +1,21 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] RadioListItem and CheckboxListItem accept rich `label` and `description`
+
+`RadioListItem` typed `label` and `description` as `string` while its sibling
+`CheckboxListItem` already typed `label` as `ReactNode` — so the same slot had
+two contracts, and an app whose option descriptions carry links could not type
+them on either component. Both now take `ReactNode`; the runtime already
+rendered it.
+
+`RadioListItem` gains the `aria-label` escape hatch `CheckboxListItem`
+established, with the same meaning: a plain-text accessible name for the
+control. The radio differs in one way worth knowing — it points at its visible
+label for its accessible name, so a rich label still names it from its own
+text, and `aria-label` is there to narrow a name that reads badly rather than
+to supply a missing one. `aria-label` now lands on the radio instead of the row `<div>`,
+where ARIA ignored it.
+
+@cixzhang
--- a/apps/storybook/rtl-audit/targets.json
+++ b/apps/storybook/rtl-audit/targets.json
@@ -11,6 +11,28 @@
     }
   },
   {
+    "component": "CheckboxList",
+    "storyId": "core-checkboxlist--rich-descriptions",
+    "dims": [
+      "D2"
+    ],
+    "selectors": {
+      "prev": ".astryx-checkbox-input",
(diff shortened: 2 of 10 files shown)


cixzhang:
/accept-visual 33592661773/1 Default frames are unchanged; the new rich-content examples render correctly in light and dark, keep selection controls aligned, and place nested links and end content without overlap.

PR #5259, opened 2026-08-20
cli: include integration-contributed components in unified search/build

### LINKED ISSUE #5258: search/build never see integration-contributed components
### Problem

`astryx search` and `astryx build` build their component candidate list from `gatherComponents(coreDir)` in `api/search/search.mjs`, which only ever scans `@astryxdesign/core`. A project's configured integrations (`astryx.config`'s `integrations` array) are never part of that pool — not behind a flag, just absent from the code path.

This means a component that only exists in an integration (no core equivalent, no name collision) is invisible to both `search` and `build`, even though `component --list` and `component <Name>` already resolve it correctly via the same `loadedIntegrations` data `component/list.mjs` already uses.

### Repro

Register an integration contributing a component, e.g. `OtpField` (no core equivalent):

- `astryx component OtpField` — finds it.
- `astryx component --list` — finds it, tagged with its owning package.
- `astryx search "otp"` — returns unrelated doc/template results, never mentions it.
- `astryx build "collect a verification code"` — recommends `TextInput` instead.

For coding agents this is the same class of problem `AGENTS.md`'s "discover, don't guess" workflow is meant to prevent: an agent following that workflow through `search`/`build` will never reach an integration's own components, and will reach for whatever core component looks closest instead.

### Fix

I have a PR ready: splits `gatherComponents` into `gatherCoreComponents` (unchanged) and `gatherIntegrationComponents`, the latter reusing `loadIntegrationsSafely` + `discoverIntegrationComponents` — the same functions `component/list.mjs` already uses for `component --list`. `gatherComponents` merges both; its call site now passes `cwd` alongside `coreDir` so the integration lookup has a project root to resolve from.

andrskr:
Fixes #5258.

## Problem

`search`/`build`'s component candidate list came from `gatherComponents(coreDir)`, which only ever scanned `@astryxdesign/core`. A project's configured integrations were never part of the pool, so an integration-only component (no core equivalent, no name collision) was invisible to `astryx search` and `astryx build`, even though `component --list` and `component <Name>` already resolve it via the same `loadedIntegrations` data.

## Fix

Split `gatherComponents` into `gatherCoreComponents` (unchanged) and `gatherIntegrationComponents`, which reuses `loadIntegrationsSafely` + `discoverIntegrationComponents` — the same functions `component/list.mjs` already uses to make integration components show up in `component --list`. `gatherComponents` now merges both, and its call site passes `cwd` alongside `coreDir` so the integration lookup has a project root to resolve from.

An integration component's `import` hint is set to its owning package name (`rec.package`), matching the granularity `component --list`'s brief detail already uses (`package: integration.name`) — no new import-path-resolution logic invented.

## Testing

New tests in `search.test.mjs` stand up a temp consumer with a stub core and an installed integration (same harness as `template-integration.test.mjs`), and assert the integration's component appears in search results with the integration package as its import hint.

- `search.test.mjs` alone: 13/13 passed (11 pre-existing + 2 new).
- `api/search`, `api/build`, `api/component` together: 35/36 passed. The one failure (`component.test.mjs`, a case-sensitivity check) is in a file this diff doesn't touch, and reproduces identically without this change — it's a case-insensitive-filesystem artifact of the sandbox this was developed in, not a regression from this patch.

Happy to adjust the `import` hint shape or split anything out if you'd rather keep this more minimal.

### DIFF
--- a/packages/cli/api/search/search.mjs
+++ b/packages/cli/api/search/search.mjs
@@ -36,10 +36,12 @@ import {pathToFileURL} from 'node:url';
 import {findCoreDir, CLI_ROOT} from '../../foundation/fs/paths.mjs';
 import {
   discoverComponents,
+  discoverIntegrationComponents,
   findComponentReadme,
   resolveImportPath,
 } from '../../foundation/discovery/component-discovery.mjs';
 import {discoverHooks, findHookDoc} from '../../foundation/discovery/hook-discovery.mjs';
+import {loadIntegrationsSafely} from '../component/_adapter.mjs';
 import {levenshteinDistance} from '../../foundation/text/string-utils.mjs';
 import {discoverTemplates, extractComponents} from '../template/template.mjs';
 import {AstryxError} from '../error.mjs';
@@ -353,12 +355,12 @@ async function loadModuleDoc(docPath, exportName = 'docs') {
 }
 
 /**
- * Build component candidates: name + keywords + usage/description from the
- * component's .doc.mjs.
+ * Build component candidates from core's own tree: name + keywords +
+ * usage/description from the component's .doc.mjs.
  * @param {string} coreDir
  * @returns {Promise<Candidate[]>}
  */
-async function gatherComponents(coreDir) {
+async function gatherCoreComponents(coreDir) {
   const grouped = discoverComponents(coreDir);
   const names = Object.values(grouped).flat();
   /** @type {Candidate[]} */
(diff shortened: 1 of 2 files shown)


packages/cli/api/search/search.mjs
```diff
@@ -386,6 +388,50 @@ async function gatherComponents(coreDir) {
...
+  const candidates = [];
+  for (const integration of loadedIntegrations) {
+    for (const rec of discoverIntegrationComponents(integration)) {
+      const doc = await loadModuleDoc(rec.docPath);
+      candidates.push({
+        domain: 'component',
+        name: rec.name,
+        keywords: doc && Array.isArray(doc.keywords) ? doc.keywords : [],
+        description: doc ? doc.usage?.description || doc.description || '' : '',
+        _import: rec.package,
```
cixzhang, line 413: `component <Name>` resolves the exports subpath (`_adapter.mjs:405`) — this hint stays at package level.

cixzhang (approved):
Thanks — search and build finally agreeing with `component --list` is the right fix, and the test fails without it. Merging as is; one optional nit inline.

[Reviewed by Robohands]

PR #5260, opened 2026-08-20
feat(cli): doctor detects CSS theming escapes and swizzled components

josephfarina:
> Stacked on #5226. Review that one first; this diff is only the two new checks.

## Summary

Two read-only, text-only `astryx doctor` checks for the cases where an app provably steps outside its theme. No modules are imported or evaluated and the scan is bounded, so both are cheap enough for the default run (~250ms added) and safe in CI.

**CSS theming escapes** covers three cases:

- A write to a private `--_*` var. `astryx theme build` already rejects these in a theme; they are no more valid in raw CSS, so this is a `fail` to match.
- A system token redefined in `:root`/`html`/`:host`. A theme applies inside `@scope ([data-astryx-theme=...])`, so a global definition sits outside it and overrides *every* theme at once.
- The deprecated bare prop classes (`.astryx-button.primary`) in place of the reflected data attributes. The fix names the exact replacement selector.

**Swizzled components** fails on exactly one combination: ejected source that imports StyleX while no StyleX compiler is configured. That is not a build error — the component renders completely unstyled, with no warning — which makes it the most expensive thing here to find by hand. Otherwise it is informational, noting that swizzled copies stop tracking upstream fixes and no longer respond to theme component overrides.

## Three details that are load-bearing

Each of these produced a false positive before it was handled, found while testing the approach against this repo:

- **Generated theme CSS is skipped.** The pipeline emits private vars and bare prop classes itself. Judging the tool's own output as consumer error accounted for 6 of the first 7 findings against `apps/docsite`.
- **Comments are blanked, not stripped.** Removing them collapses lines, which shifts every reported line number — one finding pointed at line 104 when the code was at 174. For a diagnostic an agent edits from, a wrong line is worse than no line.
- **Bare-class detection matches only enumerated prop/state values.** Otherwise `.astryx-card.my-highlight` — a consumer's own class — trips it.

Against this repo the pair reports exactly one finding, and it is a true positive: a hand-written `.astryx-pagination-dot.active` in `apps/docsite/src/app/globals.css:174`.

## Test plan

- [x] 46 tests pass, roughly half of them negative controls: clean consumer CSS, a consumer class chained onto an Astryx class, a token name inside a comment, generated output, `node_modules`/`dist`, a project with nothing swizzled, and an unrelated directory named `astryx`
- [x] Line-number accuracy pinned by a test with a multi-line banner
- [x] `check:cli-structure`, `readme:check`, `check:changesets`, golden suite
- [x] Typecheck: 6 strict errors, all pre-existing in `assets/templates/blocks/`
- [x] Verified against `apps/docsite` — 1 finding, manually confirmed genuine


packages/cli/api/doctor/theme-drift.mjs
```diff
@@ -0,0 +1,384 @@
...
+  const first = bareClasses[0];
+  const [base, value] = first.detail.slice(1).split('.');
+  return {
+    id: 'css-escapes',
+    label: 'CSS theming escapes',
+    status: 'warn',
+    message:
+      `${bareClasses.length} selector(s) use deprecated bare prop classes, e.g. ${first.detail} at ` +
+      `${at(first, ctx.cwd)}.`,
+    fix: `Target the reflected data attribute instead: .${base}[data-variant="${value}"].`,
```
cixzhang: Components reflect the prop key — `[data-variant="active"]` matches nothing for `.active`.

packages/cli/api/doctor/theme-drift.mjs
```diff
@@ -0,0 +1,384 @@
...
+        });
+      }
+    }
+  }
+
+  return {scanned: files.length, findings};
+}
+
+/** StyleX compiler plugins. Swizzled StyleX source is inert without one. */
+const STYLEX_COMPILERS = [
```
cixzhang, line 279: Missing `@stylexjs/webpack-plugin` and `@stylexjs/rollup-plugin` — both in our own styling doc.

cixzhang (changes requested):
Thanks for this — the negative controls and the line-number care are the parts most detectors skip.

Two to fix. The bare-class fix always names `data-variant`, but a component reflects the prop key: docsite's own `.astryx-pagination-dot.active` is `[data-active="active"]`, and `[data-variant="active"]` matches nothing — so following doctor drops the rule (`theme-drift.mjs:334`).

`STYLEX_COMPILERS` is missing `@stylexjs/webpack-plugin` and `@stylexjs/rollup-plugin`, the two our own styling doc tells people to install (`styling.doc.mjs:353`). A Webpack user who swizzles gets `[fail]`, exit 1 in CI, and advice to add a compiler they already have.

Same shape both times, and main now has the answer: `foundation/discovery/theming-targets.mjs` carries each target's className, props and states. Merge main and derive from it, loaded only when there's a hit to name so the 700ms stays off the clean path.

Any reason to keep the lists local?

[Reviewed by Robohands]

josephfarina:
Both correct, and both were the same mistake — I hardcoded a list rather than reading the one that exists. No reason to keep them local; merged main and derived from `collectThemingTargets` plus the component docs.

**The `data-variant` one is the worse of the two**, because it is precisely the failure this check is meant to catch: advice that is confidently wrong and mechanically applicable. Following it replaced a working rule with a selector matching nothing. Now resolved from the target's `states` and `props`, with the enumerated values parsed from the documented prop type union:

- `.astryx-pagination-dot.active` → `.astryx-pagination-dot[data-active="active"]` (state, reflects its own name)
- `.astryx-button.primary` → `.astryx-button[data-variant="primary"]` (value of `variant`)
- `.astryx-button.sm` → `.astryx-button[data-size="sm"]` (same target, different prop — so it has to be resolved, not guessed)

A useful side effect: a token that is neither a state nor a documented prop value is now recognised as the consumer's own class and dropped. `.astryx-card.promo` was only passing before because the curated list happened not to contain `promo` — it is now correct by construction.

`STYLEX_COMPILERS` now carries `@stylexjs/webpack-plugin`, `@stylexjs/rollup-plugin` and `@stylexjs/postcss-plugin` alongside the Babel and community/SWC entries, with a `SYNC:` comment pointing at the bundler table in `styling.doc.mjs` so the two cannot drift again.

On the 700ms: the enumeration measures ~300ms here and is loaded lazily, keyed by core directory. A project with no bare class to name resolves in **3ms** — the clean path never pays it. When core cannot be resolved at all the findings are dropped rather than reported without a usable fix, since guessing the attribute is what caused this.

48 tests, including one per resolution path and one pinning the drop-when-docs-unreachable behaviour.

cixzhang (approved):
Thanks — deriving both from `theming-targets` is the right fix. Verified: docsite now gets `[data-active="active"]`, and a webpack or rollup swizzle project exits 0.

[Reviewed by Robohands]

PR #5261, opened 2026-08-20
feat(DateInput): the platform's own date picker on touch, with `nativePicker`

imdreamrunner:
## What

A touch device gets the platform's own date picker — the iOS wheel, the Android calendar dialog — drawn by the OS, with its hit areas, momentum scrolling, locale and accessibility settings. `nativePicker` chooses the surface:

| Value | Behaviour |
| --- | --- |
| `'touch'` (default) | the OS picker on a coarse pointer, the calendar popover on a mouse |
| `'always'` | the OS picker wherever the browser supports `<input type="date">` |
| `'never'` | Astryx's own everywhere — #5243's bottom-sheet picker on a finger, the calendar popover on a mouse |

```tsx
// Astryx's own touch picker instead of the platform's
<DateInput label="Event date" value={date} onChange={setDate} nativePicker="never" />
```

Reach for `'never'` on a field that needs what a native picker cannot express: `weekStartsOn`, `numberOfMonths`, or `dateConstraints`.

Not breaking: `DateInputProps` gains an optional prop and nothing is removed. A field that wants the bottom-sheet picker asks for it.

## `format` and `placeholder` still apply

DateInput paints the closed field's text itself rather than letting the engine do it, so a date reads the same on a phone as on a desktop. The control's own text is made transparent and ours is drawn over it, pointer-transparent so the tap still raises the picker. The value the control holds stays ISO — all it accepts, and what the picker reads and writes.

## Four things measured on a real iPhone

None of these reproduce in a desktop browser, and each is pinned by a test that fails without its fix.

- **The element is uncontrolled, and React never rewrites it after mount.** While the iOS picker sheet is open, *any* programmatic write to the field detaches the sheet from it: the wheel keeps moving its own highlight, but nothing reaches the input and no event fires, so a pick appears to do nothing. External changes go in through an effect that only writes while the field is unfocused.
- **Edits come back through native listeners.** React's synthetic change does not reliably observe this control — measured: a native `input` event carried the new date while React's `onChange` never ran, so React re-rendered and wrote its stale value straight back.
- **The control is pinned to one line box** (`1lh`, with a `calc()` fallback) with `appearance: none`. Its intrinsic height comes from its inner edit fields rather than `line-height`, so it otherwise renders ~2px taller than the text field and its value sits off the shared baseline.
- **The text overlay is a block box with both inline insets.** `text-overflow` only applies to a block container, so on a flex one a long date hard-clips mid-glyph instead of ellipsising; without the end inset the overlay is shrink-to-fit and paints ~24px over the clear button.

## `min` / `max`

Forwarded to the control, but a native picker may not *show* them: on iOS they are constraint-validation flags rather than clamps, so the sheet lets the user land on an out-of-range date. It is refused on commit and announced to assistive technology, and the refusal clears when the picker closes and the field reverts. Verified on a device — no bad data escapes, but the picker will not grey those dates out. A field where that matters should use `'never'`.

## Test plan

- New `NativeDateField.test.tsx` — 30 tests: surface selection for all three values, the ISO round-trip, the synthetic-change bypass, the no-write-while-focused invariant, `min`/`max` forwarding and refusal, `dateConstraints`, the format overlay (named formats, function format, provider locale, live during the picker), the placeholder, the paint-level rules, `showPicker` including an engine that throws, clear, disabled, labelling and ref forwarding.
- `DateInputTouch.test.tsx` passes `nativePicker="never"` — that suite is about Astryx's own touch surface, which is now opt-out — plus one new test asserting the default. 135 pass.
- `DateInput.test.tsx` (the pointer surface) passes untouched: 95.
- `pnpm lint:strict` 0 errors, `pnpm build` clean, and the full UI suite green at **7996**.

## Rebuilt against #5243

This PR originally handed the OS picker to a single branching `DateInput`. #5243 has since landed the surface split — `DateInput.tsx` choosing between `PointerDateField` and `TouchDateField` — so the change was rebuilt as a third surface, `NativeDateField`, rather than rebased: the file it modified no longer has that shape.

The earlier commit is preserved at the tag `archive/5261-pre-rebuild` (`d4d2cf3553`) if any of it is wanted back. The research behind the approach is in P2470053021.

The review comment above predates the rebuild. Two of its three points are resolved by it — the change no longer grows `DateInput.tsx`, and it now lives in its own file — and the third, the `min`/`max` expectation gap, is documented in the prop and called out below.

### DIFF
--- a/.changeset/date-input-native-picker-option.md
+++ b/.changeset/date-input-native-picker-option.md
@@ -1,0 +1,27 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] DateInput: the platform's own date picker on touch, with
+`nativePicker` to choose otherwise
+
+A touch device now gets the browser/OS picker — the iOS wheel, the Android
+calendar dialog — drawn by the platform, with its own hit areas, momentum
+scrolling, locale and accessibility settings. `nativePicker` picks the
+surface: `'touch'` (the default) uses it on a coarse pointer, `'always'`
+wherever the browser supports `<input type="date">`, and `'never'` keeps
+Astryx's own pickers everywhere — the bottom-sheet picker on a finger, the
+calendar popover on a mouse.
+
+Reach for `'never'` on a field that needs what a native picker cannot
+express: `weekStartsOn`, `numberOfMonths`, or `dateConstraints`.
+
+`format` and `placeholder` still apply in native mode — DateInput paints the
+closed field's text itself, over the control, so a date reads the same on a
+phone as on a desktop. `min` and `max` are forwarded, but a native picker may
+not *show* them: on iOS they are constraint-validation flags rather than
+clamps, so an out-of-range date can be selected, and it is refused on commit
+and announced rather than greyed out in the picker. Mouse-driven devices are
+unaffected.
+
+@imdreamrunner
--- a/packages/cli/assets/templates/blocks/components/DateInput/DateInputClearable.tsx
+++ b/packages/cli/assets/templates/blocks/components/DateInput/DateInputClearable.tsx
@@ -16,7 +16,11 @@ export default function DateInputClearable() {
   );
 
   return (
-    <Stack direction="vertical" gap={4} width="100%" style={{maxWidth: 400}}>
+    <Stack
+      direction="vertical"
+      gap={4}
+      width="100%"
+      style={{minWidth: 240, maxWidth: 400}}>
       <Text type="supporting" color="secondary">
         {value ? `Selected: ${value}` : 'No date selected'}
       </Text>
--- a/packages/cli/assets/templates/blocks/components/DateInput/DateInputDateRange.tsx
+++ b/packages/cli/assets/templates/blocks/components/DateInput/DateInputDateRange.tsx
@@ -27,7 +27,11 @@ export default function DateInputDateRange() {
   const [value, setValue] = useState<DateString | undefined>(undefined);
 
   return (
-    <Stack direction="vertical" gap={4} width="100%" style={{maxWidth: 400}}>
+    <Stack
+      direction="vertical"
+      gap={4}
+      width="100%"
+      style={{minWidth: 240, maxWidth: 400}}>
       <Text type="supporting" color="secondary">
         {value ? `Booked: ${value}` : 'Pick a date in the available range'}
       </Text>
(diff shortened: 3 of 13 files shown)


imdreamrunner:
> [!NOTE]
> **AI-generated review.** Written by an agent (agentcloud) at @imdreamrunner's
> request, on the same branch it authored. Not an approval — a human still needs
> to review and sign off.

**Overall: I would approve this, with one structural question I would like answered before it lands, and two smaller notes.**

The empirical work behind it is the strongest part — the iOS behaviors here (the picker detaching from the field on a programmatic write, React's synthetic `onChange` not firing for the picker's edits, `text-overflow` being inert on a flex box) are all things you cannot reason your way to, and each one is pinned by a test that fails without its fix. The `useNativeDatePicker` split is clean, and the desktop path is genuinely untouched.

## The question: is `DateInput.tsx` still the right home?

This adds ~340 lines of code and 139 lines of comment to a file that is now **1259 lines with 20 hooks**. The native path carries three pieces of state (`rejectedNativeValue`, `isNativeFocused`, `hasTypedSinceFocus`), two refs (`lastNativeCommitRef`, `initialNativeValueRef`), and three effects, all of which are inert on desktop — and the JSX branch is now two full `<input>` renderings side by side, ~120 lines apart, which is where I would expect the next drift bug to come from.

I do not think this blocks merging, and I would rather not ask for a refactor that risks behavior verified on a real device. But I would like it on the record: is the intended end state a `NativeDateField` subcomponent, with `DateInput` choosing between it and the text field? If yes, a follow-up issue is enough.

## Two smaller notes

**`initialNativeValueRef` never resets.** If `isNative` goes false → true → false → true (a convertible detaching its keyboard twice), the remounted control gets a stale `defaultValue`. The sync effect corrects it on the same commit, so it is self-healing and I am not asking for a change — but a one-line comment saying so would stop the next reader re-deriving it.

**The `min`/`max` gap deserves a doc line.** iOS's picker does not visually enforce them and lets you select out of range; the component refuses the value on commit, which is right. But a consumer reading `min`/`max` will reasonably expect the picker to grey those dates out. One sentence in the `nativePicker` prop doc would set the expectation.

## What I checked

- SSR renders the text field with no value attribute — no hydration mismatch
- The provider-locale change from #5120 composes correctly with the overlay (de-DE renders "21. März 2026", the input's own value stays ISO)
- Rapid successive picks each fire `onChange`; the dedupe ref swallows nothing
- `isLoading` disables the native control and sets `aria-busy`

## Accepted tradeoffs, flagged for the record

- The default behavior changes on touch for every existing consumer
- The 180px floor truncates `date_long` in shrink-to-fit containers
- Screen readers announce the OS date format rather than `format`'s, since the text overlay is `aria-hidden`

PR #5268, opened 2026-08-20
feat(vibe-tests): test adding Astryx to existing Tailwind apps

nynexman4464:
## Zero-damage acceptance goal

Installing Astryx into an existing Tailwind application must leave all pre-existing host UI unchanged unless the task explicitly names a surface for insertion or replacement. Any unintended change to colors, typography or font metrics, spacing, radii, borders, shadows, geometry, color-mode behavior, or layer order is a failed run. `cosmetic-drift` remains useful for diagnosis but is not success.

This is the top of draft stack #5537:

1. #5500 records executor-neutral provenance.
2. #5499 adds the shared fixture suite and plain Tailwind control.
3. #5535 adds the established shadcn-style app.
4. #5536 adds the scoped enterprise-style app.
5. This PR adds the setup and cross-system composition experiment.

## Experiment contract

Every run starts from a validated pristine fixture with Astryx absent. Guidance is applied before the execution baseline, a fresh executor produces the entire source diff, and the evaluator only builds and measures the result. The run records an immediate post-execution diff digest; a missing or mismatched digest rejects later manual repair.

The prompt set keeps the original button, status, and selector tasks and adds:

- an installation-only task with a separately marked proof component;
- an Astryx tooltip and selector popup inside the existing shadcn-style dialog; and
- an existing Tailwind menu inside an Astryx dialog.

Task-owned `data-vibe-result` markers let the evaluator assert outputs without guessing component APIs. Fixture-owned `data-vibe-probe` markers remain the untouched host baseline.

## Measurement

The browser evaluator runs in light and dark modes and compares:

- color and background;
- per-side border color, width, and style;
- all four corner radii and box shadow;
- font family, size, weight, line height, and letter spacing;
- per-side padding and margin, gaps, dimensions, position, and transform;
- exact `getBoundingClientRect()` snapshots after normalization to Chromium's 1/64 CSS pixel serialization grid;
- text, contrast, color-mode behavior, variables, and emitted layer order; and
- dialog, backdrop, tooltip, selector, popover, and menu visibility, clipping, keyboard reachability, center hit target, top-layer state, and stack order.

Variable changes remain diagnostic only when no marked host computed style or geometry changes. Existing host overlay styles are compared with their baseline, including when a host surface is moved inside an Astryx dialog.

Public artifacts use stable fixture identifiers and output-relative config references rather than machine-local paths. Build diagnostics are redacted before measurement JSON is written, provenance usage sources are normalized to a generic label, and both setup and universal aggregate reports fail closed if private path or host data remains.

The diff-integrity check rejects deleted or wholesale-replaced host files, neutralized host CSS, removed or disabled dark-mode behavior, blanket resets, hardcoded `!important`, missing Astryx usage, missing requested composition, and post-run edits.

## Verdicts and reporting

The diagnostic verdicts remain `broken-build`, `noisy`, `silent-damage`, `cosmetic-drift`, and `clean`. `passesAcceptance` is true only when all of the following hold:

- execution completed successfully under the preregistered policy;
- the verdict is exactly `clean` in the combined light/dark evaluation;
- every task-specific result and interaction succeeds;
- every host and overlay probe is present and unchanged outside the requested surface;
- no escape hatch is detected; and
- the measured source diff exactly matches the runner's post-execution digest.

The host comparison is field-granular. Container text compares full normalized text from unmarked host descendants while excluding nested probe, result, and replacement subtrees, so requested insertions do not mask unrelated host copy changes anywhere in the container. Every intentional replacement names exactly one validated host probe, requires the original probe to be absent afterward, and names a result that preserves its text. Host allowances may cover only exact text or position fields; overlay allowances may cover only exact bounds. The contract validator rejects unknown fixtures, probes, results, surfaces, duplicate allowances, and every computed-style exemption. Colors, typography, radii, borders, shadows, spacing, overlay styles, and unlisted geometry remain strict.

Exploratory current-versus-candidate reports show candidate-minus-current deltas for build, runtime, task completion, color, font, radius, border, shadow, geometry, contrast, and layering. They deliberately provide no weighted score or automatic ranking, so the operator can judge explicit tradeoffs while refining instructions.

Final confirmation has no tradeoff allowance: every expected cell must be present and valid, the strict clean rate must be 100%, and the damage-free numerator must equal its denominator. A real executor failure remains a failed run. Only a classified infrastructure failure may be retried under the preregistered policy.

## Staged matrix

Prompt applicability creates 14 fixture-prompt pairs: four basic tasks across three fixtures, plus one cross-system task on each established fixture.

- **Separation:** 48 runs — three fixtures × floor/direct × `s1` × four bundles × K=2.
- **Guidance iteration:** 400 additional runs. Together with the 48 reusable separation cells, this is 448 unique four-condition cells.
- **Final confirmation:** 112 fresh candidate-only runs — 14 supported fixture-prompt pairs × four bundles × K=2.

The Codex bundle remains configured and explicitly blocked by an upstream empty-turn issue. Confirmation cannot start or claim complete bundle coverage while it is blocked or omitted. Damage in the directed control is a product gap and blocks instruction shipping.

## Verification

- 115/115 focused setup tests pass, including real build–measure–score proofs for canonical-correct `s1`–`s5` results across every supported fixture, portable setup-config generation, and subprocess checks for both aggregate reporters. The eight real-browser/config canonical cases are skipped by the default unit-test job and run in the Chromium-equipped fixture job.
- Focused TypeScript compilation passes for the evaluator and aggregator.
- Fixture validation and all three fixture production builds pass.
- Real Chromium host-overlay interactions and contrast checks pass in light and dark modes.
- Strict repository lint passes.
- The full repository suite passes with bounded workers after the default high-concurrency run hit unrelated timeouts.
- The full production build passes.
- Mutation checks fail as expected when cosmetic drift is accepted, box-shadow measurement is removed, or the host-inside-Astryx overlay direction is skipped; the committed sources are restored afterward.
- Public-boundary and copyright checks pass.

The expensive model matrix has not been run.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -562,6 +562,11 @@ jobs:
       - name: Install Playwright
         run: npx playwright install chromium
 
+      - name: Verify canonical setup compositions
+        env:
+          ASTRYX_CANONICAL_SETUP_BROWSER: '1'
+        run: pnpm vitest run internal/vibe-tests/setup-test/setup-canonical.test.ts --project node --maxWorkers=1
+
       - name: Verify guest fixture contrast
         run: pnpm -F @astryxdesign/vibe-tests fixtures:contrast
 
--- a/internal/vibe-tests/README.md
+++ b/internal/vibe-tests/README.md
@@ -150,6 +150,15 @@ internal/vibe-tests/
 ├── fixture-recipes/     # Pinned provenance and SHA-256 manifests
 ├── scripts/             # Fixture setup plus report helpers
 ├── .baseline/           # Real shadcn/ui components for baseline tsc
+├── setup-test/          # Setup evaluation over the canonical fixture matrix
+│   ├── PLAN.md               # controls, pilot stages, measures, decision rule
+│   ├── matrix.json           # fixtures × controls × prompts × bundles × reps
+│   ├── run-setup.mjs         # sandbox + task + provenance preparation only
+│   ├── setup-interactions.mjs # marker-driven dialog and nested-overlay opener
+│   ├── setup-measure.mjs     # build + exact style, geometry, and overlay probes
+│   ├── setup-integrity.mjs   # read-only diff attestation and escape-hatch checks
+│   ├── setup-eval.ts         # strict deterministic deltas against each fixture
+│   └── setup-aggregate.ts    # matrix coverage, A/B deltas, and final acceptance
 ├── results/             # Iteration results (gitignored)
 └── README.md            # This file
 ```
@@ -158,7 +167,9 @@ internal/vibe-tests/
 
 The three apps under `fixtures/` represent a plain Tailwind v4 control, an
 established shadcn-style Tailwind v4 app, and an original enterprise-style app
-with a marked guest design-system boundary. They are standalone packages with
+with a marked guest design-system boundary. The two established-app fixtures
+include controlled portal dialogs and nested tooltip/menu surfaces so setup can
+measure clipping, occlusion, and layer order. They are standalone packages with
 exact dependencies and lockfiles; none has Astryx installed.
 
 Treat canonical fixture files as immutable inputs. Setup validates their pinned
@@ -182,3 +193,9 @@ The root repository guard verifies provenance, exact manifests, deterministic
 source, dependency isolation, workspace exclusion, and separation from every
 publishable package root. This keeps future setup or migration tests independent
 of any unlanded harness branch.
+
+Generated public artifacts never retain machine-local paths. Setup configs use
+paths relative to their output directory; measurements identify the fixture and
+redact private paths or hosts from build diagnostics; exported provenance uses a
+generic usage-source label. Setup and universal aggregation fail closed before
+emitting a report that still contains an absolute path or private host.
--- a/internal/vibe-tests/fixture-recipes/tailwind-v4-control.json
+++ b/internal/vibe-tests/fixture-recipes/tailwind-v4-control.json
@@ -32,7 +32,7 @@
       "index.html": "d8daff351fd28efd42676b86e34c9312cf08410898a07f025082d619fe933240",
       "package.json": "6388776258b624beb19538053dd7103c02f70024f3f4be79e4d7c489797db986",
       "pnpm-lock.yaml": "5b50b51c8b1f5045813b3fbdcf9842527e684b34eb7aa1129a7c496b0be3158c",
-      "src/App.tsx": "fbd7fda6dc358d5be9c71484fe8afebf329079a46e737c375cee0c2b667cf6cb",
+      "src/App.tsx": "aa4b438e70fcf7abebfda1e19f974ecc505d3ec0e225e9bb60fccb6360437644",
       "src/index.css": "40c956dc7a7bfedf424aa3be1384390e728b912df21d3a0a55bfd0bbb01a0092",
       "src/main.tsx": "47cc0b4b1cc9e21351c3a37b160d7f64277f59cd57937659022cdae97608a6a5",
       "tsconfig.json": "39e5a743eb4be3f35a40bb28a216567a2df28de553728b5e9fb545ace0a2042c",
--- a/internal/vibe-tests/fixtures/tailwind-v4-control/src/App.tsx
+++ b/internal/vibe-tests/fixtures/tailwind-v4-control/src/App.tsx
@@ -75,7 +75,14 @@ export default function App() {
                       {row.item}
                     </td>
                     <td className="px-5 py-3 text-zinc-600">{row.owner}</td>
-                    <td className="px-5 py-3">{row.state}</td>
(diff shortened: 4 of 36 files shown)


nynexman4464:
## Update: a fourth cause, and where all four had already been found

Since the first push I read the full 1:1 transcript this came out of, and mined the diffs and sessions where this problem has already been hit. Two things changed.

### The layer statement's POSITION is a cause on its own

The claim I went looking for was "you can't just inject them into globals.css, for whatever reason, it doesn't work there — you need a separate file." You *can*. It has to go **first**, and putting it anywhere else is worse than not following the recipe at all.

New arm `appended-at-end`: same imports, same layer statement, added to the END of the app's existing `globals.css` — which is what "add these imports to your globals.css" produces when the file already has content. Colour mode correct, bridge not imported, and it is still the most damaging arm measured:

| arm | builds | console | app regressions | unreadable | cascade | mode-dependent | vars captured | verdict |
|---|---|---|---|---|---|---|---|---|
| `docs-verbatim` | yes | 0 | 30 (t22 c4 g1 s3) | 3 | ok | 3 | 5 | **silent-damage** |
| `docs-no-bridge` | yes | 0 | 11 (t7 c4) | 3 | ok | 3 | 2 | **silent-damage** |
| `appended-at-end` | yes | 0 | **38** (t13 c12 g6 s7) | 0 | **INVERTED** | 0 | 2 | **silent-damage** |
| `layered-in-place` | yes | 0 | 11 (t7 c4) | 0 | ok | 0 | 2 | cosmetic-drift |

`@layer a, b, c;` only orders names that are **not already registered**, and the app's own `@import 'tailwindcss'` registers `theme`, `base` and `utilities` on its way past. Below that import the statement can only append the system's layers *after* Tailwind's utilities:

```
properties > theme > base > utilities > reset > astryx-base > astryx-theme     ← appended
properties > reset > theme > base > astryx-base > astryx-theme > utilities     ← prepended
```

The system's reset then outranks every utility the app already uses, and `border-width: 0` on `:where(*)` removes every border in the app. Nothing errors.

Source review cannot catch this and neither can a class-string lint, because **the `@layer` statement does not survive compilation** — Tailwind rewrites it and the bundle carries only `@layer components;`, so the effective order is the order the layer *blocks* first appear in the output. So `setup-measure.mjs` now records the **emitted** layer order and `setup-eval.ts` scores `cascadeInverted`, which reads as `silent-damage` rather than drift: it is one edit from correct and it damages more of the app than the recipe it was trying to follow.

### Every cause here has already been found once, by someone, somewhere unfindable

That is the argument for the harness, so `PLAN.md` now says it plainly:

- **Cause 4** was a production incident — D114534048, internal. A duplicate `import '@astryxdesign/core/astryx.css'` in `layout.tsx` emitted `astryx-base` at the top of the bundle and therefore at the *bottom* of the cascade; a Settings button rendered square because `--_button-radius` resolved against a token that no longer applied. Its own words: *"That statement does not survive compilation … the effective priority order is whatever order the layer blocks first appear in the output, which is decided purely by import order."* Its guard — an e2e spec that reads the shipped CSS — is the measure this PR generalizes. *"This class of bug is silent: nothing errors, styles just quietly lose."*
- **Cause 2** is refused by name in agentcloud's `globals.css`, in a comment written the same week the bridge was being recommended everywhere else: it *"maps the relationship the wrong way round … meant for Astryx-only apps; here navi owns the utility layer."*
- **Cause 3** has a sibling in the same app: every `dark:` utility followed the reader's operating system rather than the theme they had picked, because Tailwind v4 resolves `dark:` from `prefers-color-scheme` unless `@custom-variant dark` is registered. Same shape, same silence.
- **Cause 1** has an abandoned fix: `nest add xds` first *removed* Tailwind as a "superseded package", then learned to detect the app's styling system and inject the layered block — but only for an app it is setting up, and only inside its own markers.

### The residue now has a name

Every arm, the candidate included, captures `--color-accent` and `--color-border`. That is not measurement noise: the system's tokens live in `@scope … to ([data-astryx-theme])`, which looks contained but is not — **`@scope` bounds selector matching, not inheritance** — and the provider deliberately syncs that attribute onto `<html>` so portals can find it. Tailwind's `bg-accent` reads `var(--color-accent)`, so an app that named its own accent the obvious thing silently starts painting the system's. The fix that worked elsewhere is to re-assert the app's values *on* the boundary element rather than in `:root`; it is in the candidate guidance, and PLAN is explicit that the harness reports the collision without yet asserting the fix.

### Two open upstream items this test depends on

- **#5205** — the StyleX layer splitter only ran in the Vite dev server, so a production build shipped atoms in flat `@layer priorityN` blocks outranking `astryx-theme`: *"the layer order we advertise is real in dev and fiction in production."* Until it lands, any layer claim verified in a dev server is unverified — which is why this harness measures a production build.
- **#5165** — a theme cannot declare an app-specific semantic token, the escape hatch an existing app needs when its vocabulary is wider than the system's.

24 tests (4 new), still all four detectors proven to go red when disabled.

Issue #185, opened 2026-02-19
XDSAppShell


cixzhang:
## Spec (revised 2026-02-21)

Application-level layout shell that provides the structural frame for an app: header, sidebar navigation, and main content area. Composes `XDSLayout` internally. Replaces internal `XDSPage` + `XDSPageLayout` pattern.

**This is a dedicated component**, separate from `XDSLayout`. AppShell is an opinionated app-level orchestrator that composes Layout under the hood. Layout remains the generic spatial primitive.

### Internal API analysis

Internal `XDSPage` (8,585 usages) and `XDSPageLayout` (7,374 usages) work together:

**XDSPage:**
- Top-level wrapper for the entire application page
- Manages global layout (header bar, content area)
- Renders `XDSTopNav` in the header slot
- Handles page-level scroll behavior

**XDSPageLayout:**
- Content-area layout within XDSPage
- Manages sidebar navigation (`XDSPageNav`) + main content split
- Sidebar collapse/expand behavior
- Responsive breakpoints for sidebar visibility
- Content area scroll containment

**Combined pattern:**
```tsx
<XDSPage topNav={<XDSTopNav ... />}>
  <XDSPageLayout pageNav={<XDSPageNav ... />}>
    {mainContent}
  </XDSPageLayout>
</XDSPage>
```

### External comparison

| Feature | Mantine (AppShell) | Ant (Layout) | shadcn (Sidebar) | **XDS (proposed)** |
|---------|-------------------|-------------|------------------|--------------------|
| API style | Compound component | Compound component | Composition | **Slot props** |
| Areas | Header, Navbar, Aside, Footer, Main | Header, Sider, Content, Footer | Sidebar + content | **topNav, pageNav, topBanner, children** |
| Sidebar collapse | `navbar.collapsed` | `Sider.collapsed` | Manual | **`isSidebarCollapsed`** |
| Responsive | `navbar.breakpoint` | `Sider.breakpoint` | Manual | **`sidebarBreakpoint`** |
| Scroll behavior | Main area scrolls | Content scrolls | Manual | **Mode-dependent (fill/auto)** |
| Padding | `padding` prop | Manual | Manual | **Inherited from XDSLayout** |
| Sticky nav | Manual | Manual | Manual | **Built-in, mode-aware** |
| Built-in providers | None | None | None | **Theme, Prose, Layer** |

### Key decisions

| Decision | Choice | Rationale |
|----------|--------|-----------|
[...]

PR #5269, opened 2026-08-20
lint: forbid inline mergeRefs calls

nynexman4464:
## What

Adds `@astryx/no-inline-merge-refs` as an error for core and updates the one violation that reached `main` after this branch was created.

The earlier parts of the original stack have landed:

- #5266 added `useMergedRefs` and fixed `Text`/`Heading`.
- #5267 migrated the existing core callsites.
- This PR prevents the unstable pattern from returning.

The branch is rebuilt directly on current `main`, and `TouchDateField` now uses `useMergedRefs` for the callsite added in #5350.

## Why enforcement is needed

The unsafe code looks natural:

```tsx
<Component ref={mergeRefs(forwardedRef, internalRef)} />
```

But `mergeRefs(...)` is a callback factory. Calling it during render creates a new ref callback every time, forcing React to detach and reattach an unchanged element. #5266 shows that this can escalate into React error #185 when a ref callback schedules state during commit.

Providing `useMergedRefs` is not enough by itself: the unsafe utility remains available for non-React composition, and the inline JSX form is easy to reintroduce.

There is direct ecosystem precedent. Radix added `scripts/check-composed-refs.mjs` after the same React 19 “Maximum update depth exceeded” failure (radix-ui/primitives#3963). Its guard rejects the same two patterns enforced here.

## What the rule rejects

```tsx
// Invalid: new callback ref every render
<Component ref={mergeRefs(forwardedRef, internalRef)} />

// Invalid: the inline callback changes every render, so the merged ref does too
useMergedRefs(forwardedRef, node => setNode(node));
```

## What remains valid

```tsx
// Valid when called unconditionally under the normal Rules of Hooks
<Component ref={useMergedRefs(forwardedRef, internalRef)} />

const ref = useMergedRefs(forwardedRef, internalRef);
<Component ref={ref} />
```

The rule follows imported bindings, including aliases, and does not flag an unrelated local function that happens to be named `mergeRefs`.

The rule applies to core only; lab remains outside this migration's scope.

## Validation

- Confirmed current `main` fails on `TouchDateField` with exactly one `@astryx/no-inline-merge-refs` error; the same lint command passes after the fix.
- Focused rule suite: 10 passed.
- `DateInputTouch.test.tsx`: 134 passed.
- `pnpm lint:strict`: passed.
- `pnpm build`: passed.
- `pnpm test`: passed.

### DIFF
--- a/internal/eslint-plugin-astryx/README.md
+++ b/internal/eslint-plugin-astryx/README.md
@@ -249,6 +249,34 @@ Style objects imported from another module are read from that module, so
 rewrite that moves the styles onto the child. Removing a node can shift layout,
 so it is never applied by `--fix`.
 
+### `@astryx/no-inline-merge-refs`
+
+Flags two unstable ref-composition patterns:
+
+1. `mergeRefs(...)` inside a JSX `ref` prop. Calling the utility during render
+   creates a new callback ref every time, so React detaches and reattaches the
+   element on unrelated rerenders.
+2. An inline callback passed to `useMergedRefs(...)`. The changing input makes
+   the hook recreate its callback on every render, defeating the hook.
+
+Use `useMergedRefs(...)` with stable ref inputs instead.
+
+```tsx
+// Bad
+<div ref={mergeRefs(forwardedRef, internalRef)} />;
+useMergedRefs(forwardedRef, node => setNode(node));
+
+// Good — the hook itself may be called inline when Hooks ordering is valid
+<div ref={useMergedRefs(forwardedRef, internalRef)} />;
+
+// Also good
+const ref = useMergedRefs(forwardedRef, internalRef);
+<div ref={ref} />;
+```
+
+The rule is an error in both tiers because core contains no inline
+`mergeRefs(...)` JSX callsites.
+
 ### `@astryx/require-letter-spacing`
 
 Recommends adding `letterSpacing` when `fontSize` is defined (common design pattern for badges, labels).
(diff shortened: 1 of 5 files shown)


imdreamrunner (approved):

cixzhang:
Leaving this one for a teammate to land rather than arming auto-merge, and noting why here so it isn't mistaken for neglect.

The rule itself is right and the PR is clean. The care is in the landing: a new lint rule doesn't only apply to new code — it fails every open branch that hasn't adopted the pattern yet. `mergeRefs(` currently appears in 52 files on main, and any open PR touching one of them starts failing lint through no fault of its author. With ~200 PRs open that's a lot of red appearing at once, on branches whose owners won't know why.

So this wants a person watching the first CI cycle after it merges, ready to answer "why did my PR go red" — not a bot arming it overnight. Best paired with [#5266](https://github.com/facebook/astryx/pull/5266)/[#5267](https://github.com/facebook/astryx/pull/5267) so the rule and the thing it points people toward land together.

Approved and ready whenever someone can babysit it.

[Reviewed by Robohands]

cixzhang (dismissed):
Thanks — the rule is right, and the two PRs under it landed clean.

Correcting my own earlier note: I said the risk was open branches going red. It's main. `TouchDateField` landed in [#5350](https://github.com/facebook/astryx/pull/5350) after you branched and carries the exact pattern — `ref={mergeRefs(ref, inputRef)}` at `TouchDateField.tsx:1172`. Run the rule against main today and that is the only ESLint error in `packages/core/src`. CI is green here because it last ran on a three-day-old merge base.

One for later, not this PR: `Avatar.tsx:651` and `Button.tsx:762` build the same unstable callback with a hoisted `const`, which your test marks valid.

Merge main and move that one line to `useMergedRefs` — then it's error-clean and I'd land it.

[Reviewed by Robohands]

PR #5272, opened 2026-08-20
fix: stabilize useLayer references

### LINKED ISSUE #5271: [Bug] useLayer recreates trigger refs and API objects on every render
## Problem

`useLayer({mode: 'context'})` creates a new trigger callback ref on every render. React therefore detaches and reattaches an unchanged trigger element, which removes and reapplies its CSS anchor name.

`useLayer` also returns a new API object on every render in both context and fixed modes. Consumers that correctly depend on the layer object rerun effects and reinstall listeners even when no layer state changed.

Examples in core include Carousel, ContextMenu, Popover, Tooltip, HoverCard, and BaseTypeahead.

## Proposed fix

- wrap the context trigger ref in `useCallback`
- memoize the context and fixed return objects from their exposed members
- preserve the existing rule that the result changes when an exposed member changes, including `isOpen`, callbacks, or render behavior

## Regression coverage

Tests verify that:

- the context trigger ref remains stable across unrelated rerenders
- context and fixed return objects remain stable while their members are unchanged
- changing `onShow` refreshes the API object but does not replace the trigger ref

No public types or behavior change.

nynexman4464:
## Summary

- keep the context trigger ref returned by `useLayer` stable across unrelated rerenders
- memoize the context and fixed API objects from their exposed members
- preserve updates when callbacks, open state, or render behavior actually change
- add regression coverage for stable refs and return objects

Fixes #5271.

## Test plan

- fail-first on current `main`: the three identity tests fail (3 failed, 43 passed)
- `pnpm vitest run packages/core/src/Layer/useLayer.test.tsx` — 46 passed
- `pnpm lint:strict`
- `pnpm build`
- `pnpm vitest run --maxWorkers=8` — 12,226 passed, 5 skipped

### DIFF
--- a/.changeset/stable-use-layer.md
+++ b/.changeset/stable-use-layer.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Keep useLayer trigger refs and return objects stable when their inputs are unchanged.
+
+@nynexman4464
--- a/packages/core/src/Layer/useLayer.test.tsx
+++ b/packages/core/src/Layer/useLayer.test.tsx
@@ -10,7 +10,13 @@
  */
 
 import {describe, it, expect, vi, afterEach} from 'vitest';
-import {render, act, fireEvent, waitFor} from '@testing-library/react';
+import {
+  render,
+  renderHook,
+  act,
+  fireEvent,
+  waitFor,
+} from '@testing-library/react';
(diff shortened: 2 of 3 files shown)


nynexman4464:
/accept-visual 33093725183/1 The change only stabilizes ref and return-object identity; these added Layer frames establish the missing baseline without an intended visual change.

PR #5276, opened 2026-08-20
fix: stabilize default resize snaps

### LINKED ISSUE #5275: [Bug] useResizable recreates callbacks when snaps are omitted
## Problem

`useResizable` defaults `snaps` with an array literal during render:

```tsx
const {snaps = []} = config;
```

When callers omit `snaps`, that produces a new array on every render. Because `expand`, `resize`, and `onResizeMove` depend on `snaps`, those callbacks are recreated too. The private `_snaps` value passed to `ResizeHandle` also changes identity even though the configuration did not change.

## Proposed fix

Keep one empty snaps array per hook instance with `useRef`, and use it only when `config.snaps` is absent. Explicit caller-provided arrays keep their existing behavior.

## Regression coverage

A hook test rerenders the same omitted-snaps configuration and verifies that:

- `expand` remains stable
- `resize` remains stable
- `props._snaps` remains stable

The test fails on current main and passes with the patch. Existing ResizeHandle behavior remains green.

This change intentionally does not memoize the entire `useResizable` return object; that broader optimization is not needed to fix the confirmed default-array defect.

nynexman4464:
## Summary

- replace the render-time `snaps = []` default with one stable empty array per hook instance
- keep `expand`, `resize`, and `_snaps` stable when callers omit snap points
- add focused identity regression coverage

Fixes #5275.

## Test plan

- fail-first on current `main` (`5c396656f`): `pnpm exec vitest run packages/core/src/Resizable/useResizable.test.tsx` fails the callback identity assertion as expected
- `pnpm exec vitest run packages/core/src/Resizable/useResizable.test.tsx packages/core/src/Resizable/ResizeHandle.test.tsx` — 21 passed
- `pnpm lint:strict` — passed with 0 errors; 80 existing warnings outside the changed files
- `pnpm build` — passed
- `pnpm test` — 12,214 passed, 5 skipped (581 test files passed, 1 skipped)

### DIFF
--- a/.changeset/stable-resizable-snaps.md
+++ b/.changeset/stable-resizable-snaps.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Keep useResizable callbacks stable when snap points are omitted.
+
+@nynexman4464
--- a/packages/core/src/Resizable/useResizable.test.ts
+++ b/packages/core/src/Resizable/useResizable.test.ts
@@ -400,3 +400,22 @@ describe('useResizable live collapse state', () => {
     expect(onCollapseChange).toHaveBeenCalledExactlyOnceWith(true);
   });
 });
+
+describe('useResizable identity', () => {
+  it('keeps callbacks stable when optional snaps are omitted', () => {
+    const {result, rerender} = renderHook(() =>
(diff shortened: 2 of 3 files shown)


nynexman4464:
/accept-visual 33093814142/1 The change only stabilizes omitted-snaps callback identity; these added Resizable frames establish the missing baseline without an intended visual change.

Issue #5277, opened 2026-08-21
[Bug] table-grouped template renders <tr> directly inside <table> — invalid DOM, and a hydration error for anyone who uses the template as a page


rubyycheung:
## Summary

The `table-grouped` page template renders `<TableRow>` as a direct child of
`<Table>`, with no `<TableBody>` between them. The emitted DOM is
`<table><tr>…`, which is invalid HTML. React's DOM nesting validation flags it
on every render:

```
In HTML, <tr> cannot be a child of <table>. Add a <tbody>, <thead> or <tfoot>
to your code to match the DOM tree generated by the browser.
This will cause a hydration error.

<table className="astryx-base-table …" style={{minWidth:"596px"}}>
  <colgroup>
  <TableRow role="button" tabIndex={0} …>
>   <tr role="button" tabIndex={0} className="astryx-table-row …">
```

It is the only one of the 31 page templates with this shape. Every other
template that renders a `Table` uses the data-driven `data={…}` API, where
`BaseTable` supplies the `<tbody>` itself.

## Why it matters more than a console warning

A page template is starter code — the documented use is to copy it into an app
as a page. In a Next.js App Router page that is server-rendered, and there the
warning becomes a real failure: the HTML parser applies implied-`<tbody>`
insertion when it parses the server's markup, so the client tree and the parsed
tree disagree and hydration mismatches. Anyone who adopts this template inherits
that, in their app, not in ours.

Even client-only, the DOM is invalid: rows created through `appendChild` are not
reparented, so the table ends up with `<tr>` children and no `<tbody>` at all.
Any CSS or query that targets `tbody` — a consumer's own, or a future one in
core — silently misses.

## Root cause

`Table`'s children mode used to wrap its children in a `<tbody>` for you.
#2097 changed that, and its own text records the old behaviour:

> Today, `children` is dumped into a single `<tbody>` — after this change,
> children can include `XDSTableHeader`, `XDSTableBody`, and `XDSTableFooter`
> for full structural control.

`BaseTable` now renders `{children ? children : <>…</>}`, so children go in
raw. That is the right API. But the template was written against the implicit
wrapper and was never migrated, and nothing caught it:

- No runtime or build-time diagnostic — `Table` accepts any children.
- The changelog entry for the feature (#2098, "Table structural children mode")
  reads as an addition; there is no note that children stopped being wrapped.
- `Table.doc.mjs` still describes the prop as *"Children mode: render
[...]

PR #5278, opened 2026-08-21
fix(cli): grouped-table template rendered rows outside a tbody

AKnassa:
The `table-grouped` page template rendered its rows straight into `<Table>`, with no `<TableBody>` between them. Rendered, that table comes out with a `<colgroup>` and **34 bare `<tr>` children and no `<tbody>` at all**.

`<table>` cannot contain a row directly. When a browser *parses* server-rendered markup it inserts an implied `<tbody>` the author never wrote; when React *renders* on the client it does not. The two trees disagree and the page hydration-mismatches — and because a page template is starter code people copy, that happens in their app, not ours. Client-only it is still wrong: nothing reparents the rows, so any CSS or query aimed at `tbody` silently misses.

Reported by @rubyycheung in #5277, with the root cause already traced.

## What changed

**The template.** Import `TableBody`, wrap the `groupKeys.map(...)`. Ignoring the reindentation that follows, it is nine lines. `TableBody` is the same element the data-driven `data={...}` path renders, so styling, dividers, and column widths are untouched.

**The docs that taught the shape.** This was not one site. `Table`'s children mode stopped wrapping children in a `<tbody>` in #2098, but the documentation still described the contract from before it:

- `children` read *"render TableRow/TableCell directly"*, naming neither the section components nor the fact that children go straight to the `<table>`.
- `TableRow`'s own `@example` showed a row sitting in `<Table>` with nothing around it. The template is written exactly like that example.
- `TableHeader`, `TableBody`, and `TableFooter` have been public exports since #2098 and had **no docs at all** — which is why nobody knew to reach for them. They were absent from `Table`'s component list and from the docsite.

Each now has a doc modelled on the existing sibling docs, `Table` lists them, and its `children` description, a new best practice, and `TableRow`'s example all name them.

**A guard, so it cannot come back.** Nothing in the repo read a table's JSX for structure: no runtime or build-time diagnostic, and the `@eslint-react` DOM rules are scoped to `packages/core/src` while the offending file was a CLI template. `@astryx/require-table-section` flags a row written as a direct child of a table instead of inside a section. It is the `<table>` counterpart of `no-raw-paragraph` — both guard an HTML content model the parser repairs and React does not. Wired repo-wide, error in both tiers, because the repo is clean.

`Table.test.tsx` held the last two instances: its children-mode tests passed rows with no section, pinning the invalid shape as supported. They now use `TableBody`, and new tests state the contract the docs never did.

## One thing worth a reviewer's eye

Adversarial edge-case testing of the new rule (last commit) found a false positive bad enough that it should not have shipped. `CallExpression` was in the walk's transparent set as a bare node-type check, so the walk stepped out of *any* call the way it steps out of a `.map()` callback — and a row handed to a function as an argument got blamed on the enclosing table:

```tsx
<Table>{React.createElement('tbody', null, <TableRow/>)}</Table>   // reported, though the row IS in a tbody
<Table>{buildGroupedBodies(groups, g => <TableRow key={g.id}/>)}</Table>   // the natural helper-extraction of this very template
```

An error-level rule with no autofix telling authors to wrap already-correct code. Transparency through a call is now position-aware: follow a chain off the result, step out of a callback only for `.map`/`.flatMap` (whose result is spliced into the position the call occupies), stay silent when the row is itself an argument. The same pass fixed five silences that hid real `<table><tr>` — a method chained after the map, a `return` one statement deeper in the callback, JSX spread children, and member-expression names like `Astryx.TableRow`.

## Test plan

- **Proved on the real artifact, not fixtures.** Rendering the actual shipped template at `upstream/main` gives `<table>` children `["COLGROUP","TR"×34]`; after the fix, `["COLGROUP","TBODY"]` with zero DOM-nesting errors on the console. The rule fires on the real `page.tsx` at exactly the two row sites before the fix.
- **Rule tests 21 → 49.** Every case added in the last commit was watched fail against the rule without its fix. Mutation testing drove the coverage: of 19 mutations of the rule, **17 are killed by the suite** — before hardening, 9 of 22 survived untouched, including every use of `BaseTable` and the fragment shorthand. The two survivors are a defensive type-check redundant with the lookup beside it, and a guard with no realistic distinguishing input.
- Full suite **11145 pass**, plus the two known macOS case-folding failures (`useMediaquery` vs `useMediaQuery`) that also fail on a clean `main` here.
- `check:repo`, `CI=true eslint .` (0 errors; the 54 warnings are pre-existing and none are in files this touches), core `typecheck` + `typecheck:docs`, cli `typecheck:template-docs` + `typecheck:strict`, docsite `generate` + **383/383** — all clean.

No screenshots: the change is DOM structure, and the before/after element lists above are the evidence a screenshot could not give.

## One open question for a maintainer

The issue also suggests a **dev-mode warning inside `Table`** when children mode gets a direct `TableRow`. That is a public runtime behaviour change on a core component, so it is deliberately **not** in this PR. If it is wanted, the house pattern is `useDevWarning('Table', ..., condition)` — precedents at `Field.tsx:221` and `Item.tsx:420` — and I am happy to add it here or in a follow-up.


cixzhang (approved):
Thanks — the docs half is what makes this stick. The template fix alone leaves the next person composing rows straight into `<Table>`; documenting the three section components is what stops it being re-authored.

Verified in Chromium on the real template: before, 34 `<tr>` direct children and React's own "cannot be a child of `<table>`" hydration error; after, one `<tbody>` with all 34 rows and a clean console. Frames at 1280 and 600 are byte-identical, so no visual change. The rule fires 6 times on main across 4,023 files with no false positives — 2 are the reported bug, 4 are test fixtures that pinned the invalid shape — and all 6 are fixed here.

Merging. One nit: `fix(cli)` under-states an `@astryxdesign/core` patch.

[Reviewed by Robohands]

cixzhang:
Merged `main` in for you — `eslint-plugin-astryx/index.js` conflicted where both sides added a rule; kept both, and `require-table-section` is registered and enabled in both configs. Its 49 tests and all 493 Table tests pass locally. Armed auto-merge.

Issue #2097, opened 2026-05-08
feat(Table): make BaseTable private, add children-mode structural sub-components


cixzhang:
## Summary

`XDSBaseTable` is currently exported but has zero external consumers (confirmed via BigGrep across Nest). In children mode, the split between BaseTable and Table is meaningless — BaseTable children renders unstyled HTML because the sub-components hit their `if (!ctx)` fallback. Nobody would use it over raw `<table>`.

Meanwhile, children mode on `XDSTable` is under-specified — you pass raw `<thead>`/`<tbody>`/`<tfoot>` elements and the sub-components only style the body rows. There are no first-class structural wrappers.

### 1. Make `XDSBaseTable` private

Remove from `packages/core/src/Table/index.ts` exports. Keep as internal implementation detail of `XDSTable`. This is a breaking change — but with zero external consumers, impact is nil.

### 2. Add children-mode structural sub-components

| Component | Element | Purpose |
|-----------|---------|---------|
| `XDSTableHeader` | `<thead>` | Header section. Cells inside auto-render as header style. |
| `XDSTableBody` | `<tbody>` | Body section. Could host empty state when no rows. |
| `XDSTableFooter` | `<tfoot>` | Footer section for aggregation/summary rows. |

### 3. Update `XDSTable` children mode

Detect structural sub-components and render them in the correct positions. Today, `children` is dumped into a single `<tbody>` — after this change, children can include `XDSTableHeader`, `XDSTableBody`, and `XDSTableFooter` for full structural control.

### 4. Usage

```tsx
<XDSTable density="compact" dividers="rows">
  <XDSTableHeader>
    <XDSTableRow>
      <XDSTableHeaderCell>Name</XDSTableHeaderCell>
      <XDSTableHeaderCell>Age</XDSTableHeaderCell>
    </XDSTableRow>
  </XDSTableHeader>
  <XDSTableBody>
    <XDSTableRow>
      <XDSTableCell>Alice</XDSTableCell>
      <XDSTableCell>30</XDSTableCell>
    </XDSTableRow>
  </XDSTableBody>
  <XDSTableFooter>
    <XDSTableRow>
      <XDSTableCell>Total: 1</XDSTableCell>
    </XDSTableRow>
  </XDSTableFooter>
</XDSTable>
```

## Motivation

[...]

PR #5280, opened 2026-08-21
fix(templates): wrap table-grouped body rows in TableBody

### LINKED ISSUE #5277: [Bug] table-grouped template renders <tr> directly inside <table> — invalid DOM, and a hydration error for anyone who uses the template as a page

HelloOjasMutreja:
Fixes #5277.

## Problem

`table-grouped` renders `TableRow` directly as a child of `Table`, with no `TableBody` between them: `<table><tr>...`, invalid HTML that React's DOM nesting validation flags on every render, and a real hydration-mismatch risk for anyone using this template as a server-rendered Next.js page (the browser's HTML parser implies a `<tbody>` insertion that the client tree doesn't have).

Root cause per the issue: `Table`'s children mode used to wrap children in a `tbody` implicitly, until #2097 changed that so consumers get full structural control via `TableHeader`/`TableBody`/`TableFooter`. This template predates that change and was never migrated. `Table.doc.mjs` still described the pre-#2097 contract (children prop description named only `TableRow`/`TableCell`, and the related-components list didn't include `TableHeader`/`TableBody`/`TableFooter` at all), which is exactly the shape a reader following the docs would reproduce.

## Fix

- Wraps the `groupKeys.map(...)` block in `TableBody` (already exported from `@astryxdesign/core/Table`, the same wrapper the data-driven path renders internally, so no visible/styling change).
- Updates `Table.doc.mjs`'s `children` prop description and related-components list to reflect the actual current (post-#2097) contract.

## Verification

- `astryx template table-grouped` (the CLI's own template renderer) confirms the emitted source now wraps the group rows in `TableBody`.
- `astryx component Table --dense` renders cleanly with the updated components list; `astryx component TableBody` resolves (falls back to Table's own docs, since `TableBody` doesn't have its own standalone `.doc.mjs`, same as before this change).
- `node scripts/check-cli-structure.mjs` passes.
- Lint clean. This is template/doc-asset only, no changeset (matches the pattern of prior template-only commits, e.g. #5167).

Not implemented from the issue's other two suggestions (a dev-mode warning for a direct `TableRow` child, and a repo-wide sweep for other pre-#2097 consumers): both are their own separate pieces of work, worth their own issue/PR rather than folding into this one.

### DIFF
--- a/packages/cli/assets/templates/pages/table-grouped/page.tsx
+++ b/packages/cli/assets/templates/pages/table-grouped/page.tsx
@@ -39,6 +39,7 @@ import {
   Table,
   TableRow,
   TableCell,
+  TableBody,
   proportional,
   pixel,
   resolveColumnWidths,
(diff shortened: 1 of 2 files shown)


rubyycheung (approved):
Looks good from a design perspective. The explicit TableBody preserves the existing rendered appearance while correcting the table structure and preventing hydration mismatches.

HelloOjasMutreja:
Superseded by #5278, which merged this same fix (TableBody wrap in the table-grouped template plus the Table.doc.mjs update) along with a broader eslint rule catching the same class of issue elsewhere. Closing to avoid duplicate work.

PR #5283, opened 2026-08-21
fix(Resizable): stabilize useResizable callbacks when snaps omitted

### LINKED ISSUE #5275: [Bug] useResizable recreates callbacks when snaps are omitted

HelloOjasMutreja:
Fixes #5275.

## Problem

`useResizable`'s single-region path defaults `snaps` with an array literal during the destructure:

```ts
const {snaps = []} = config;
```

When a caller omits `snaps`, this allocates a new `[]` on every render. `expand`, `resize`, and `onResizeMove` all list `snaps` in their `useCallback` deps, so every one of them is recreated on every render even when the caller's configuration never changed. `props._snaps` (passed to `ResizeHandle`) changes identity for the same reason.

## Fix

Keep one stable empty array per hook instance via `useRef`, and only fall back to it when `config.snaps` is `undefined`. A caller-provided array is passed through unchanged and keeps its own identity, matching existing behavior.

## Verification

Added `useResizable.test.ts` with two tests:
- Rerenders a config that omits `snaps` and asserts `expand`, `resize`, and `props._snaps` are referentially stable across the rerender. Confirmed this fails on the pre-fix implementation (`expected [Function] to be [Function]`) and passes after.
- A caller-provided `snaps` array also keeps its identity across rerenders (unaffected by this change).

Full `Resizable` suite (22 tests across `useResizable.test.ts` and `ResizeHandle.test.tsx`) passes. Typecheck and lint clean.

This intentionally does not memoize the whole `useResizable` return object; that's a broader optimization not needed to fix the confirmed default-array defect, matching the issue's own scoping.

### DIFF
--- a/.changeset/resizable-snaps-stability.md
+++ b/.changeset/resizable-snaps-stability.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] `useResizable` no longer recreates its `expand`/`resize`/`onResizeMove` callbacks (and `props._snaps`) on every render when `snaps` is omitted. The hook previously defaulted `snaps` to a new `[]` array literal in its destructure, so every consumer that depends on those callbacks re-ran unnecessarily even when the configuration never changed. A caller-provided `snaps` array keeps its existing identity-preserving behavior.
+
+@HelloOjasMutreja
--- a/packages/core/src/Resizable/useResizable.test.ts
+++ b/packages/core/src/Resizable/useResizable.test.ts
@@ -3,7 +3,8 @@
 /**
  * @file useResizable.test.ts
  * @input Uses vitest, @testing-library/react renderHook, useResizable
- * @output Unit tests for useResizable persistence and collapse state
+ * @output Unit tests for useResizable persistence, collapse state, and
+ *   callback/prop stability
  * @position Testing; validates useResizable implementation
  *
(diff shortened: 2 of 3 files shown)


HelloOjasMutreja:
Superseded by #5276, which merged the same fix (one stable empty-array-per-instance for omitted snaps, keeping expand/resize/_snaps identity stable) plus its own regression coverage. Closing to avoid duplicate work.

PR #5284, opened 2026-08-21
fix(useTheme): stabilize token resolver and return object

### LINKED ISSUE #5273: [Bug] useTheme returns unstable token resolver and API object
## Problem

`useTheme` memoizes its resolved `tokens` map, but creates a new `token` function and return object on every render:

```tsx
const token = (name: string) => tokens[name] ?? '';
return {name, mode, token, tokens};
```

This defeats memoization in consumers. For example, `useChartColors` memoizes its palette from `token`, but the changing resolver causes it to rebuild the full color API after every unrelated rerender. An effect depending on that API reruns even though the theme and mode did not change.

## Proposed fix

- memoize `token` from the resolved `tokens` map
- memoize the returned `UseThemeReturn` object from its exposed members
- preserve identity changes when the theme or effective mode actually changes

## Regression coverage

A test rerenders `useTheme` with no theme change and verifies both the token resolver and return object remain stable. The test fails on current main and passes with the patch.

No public types or resolved token values change.

HelloOjasMutreja:
Fixes #5273.

## Problem

`useTheme` memoizes its resolved `tokens` map, but rebuilds the `token` resolver function and the returned object on every render:

```tsx
const token = (name: string) => tokens[name] ?? '';
return {name, mode, token, tokens};
```

This defeats memoization for any consumer that depends on `token` or the returned object's identity. `useChartColors` memoizes its color palette from `token`, so the changing resolver rebuilds the full color API after every unrelated rerender, and any effect depending on that API reruns even though the theme and effective mode never changed.

## Fix

`token` is memoized via `useCallback` from the resolved `tokens` map. The returned `UseThemeReturn` object is memoized via `useMemo` from its exposed members (`name`, `mode`, `token`, `tokens`), so identity only changes when the theme or effective mode actually changes.

No public types or resolved token values change.

## Verification

Added a regression test that rerenders `useTheme` with no theme change and asserts both the token resolver and the returned object keep referential identity. It fails on current main and passes with this fix. Full `useTheme.test.tsx` suite (21 tests) still passes. Typecheck and lint pass locally.

### DIFF
--- a/.changeset/usetheme-stable-token-resolver.md
+++ b/.changeset/usetheme-stable-token-resolver.md
@@ -1,0 +1,9 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] `useTheme` now returns a stable `token` resolver and a stable return object when the resolved theme and effective mode haven't changed. Previously both were recreated on every render, which defeated memoization in consumers like `useChartColors` (it memoizes its palette from `token`, so the changing resolver rebuilt the full color API after every unrelated rerender, and any effect depending on it reran unnecessarily).
+
+@nynexman4464
+
+@HelloOjasMutreja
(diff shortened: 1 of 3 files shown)


HelloOjasMutreja:
Superseded by #5274, which merged the same fix (memoizing the token callback and the return object on the resolved theme/mode) before this one landed. Closing to avoid duplicate work.

PR #5285, opened 2026-08-21
fix(useLayer): stabilize context trigger ref and return objects

### LINKED ISSUE #5271: [Bug] useLayer recreates trigger refs and API objects on every render

HelloOjasMutreja:
Fixes #5271.

## Problem

`useLayer({mode: 'context'})` created a new trigger callback ref on every render:

```tsx
const ref: RefCallback<HTMLElement> | undefined =
  mode === 'context'
    ? (el: HTMLElement | null) => { ... }
    : undefined;
```

React detaches and reattaches a ref whose identity changes between renders, even when the underlying trigger element hasn't. That means every unrelated rerender removed and reapplied the trigger's CSS anchor name for no reason.

`useLayer` also returned a new plain object on every render, in both context and fixed modes:

```tsx
return {ref, anchorId, show, hide, isOpen, id, render: renderContext};
```

Consumers that correctly memoize off the returned layer object (or depend on it in an effect) rerun that work on every unrelated rerender, since the object's identity always changes. Affects Carousel, ContextMenu, Popover, Tooltip, HoverCard, and BaseTypeahead, all of which use `useLayer` under the hood.

## Fix

- The context trigger ref is now wrapped in `useCallback`, keyed on `anchorId` (which itself only changes if the `useId()`-derived id changes, i.e. never across the hook's lifetime).
- Both the context and fixed return objects are memoized via `useMemo` from their exposed members (`show`, `hide`, `isOpen`, `id`, `render`, plus `ref`/`anchorId` for context), so identity only changes when one of those actually changes.

No public types or behavior change.

## Verification

Added a regression suite (`useLayer stability (issue #5271)`) covering exactly what the issue asked for:
- the context trigger ref stays stable across an unrelated rerender
- the context return object stays stable while its members are unchanged
- the fixed return object stays stable while its members are unchanged
- changing `onShow` refreshes the API object's identity but does not replace the trigger ref

Full `useLayer.test.tsx` suite (47 tests) passes, plus the full test suites of every cited consumer (Popover, Tooltip, HoverCard, Carousel, ContextMenu, BaseTypeahead: 266 tests total across all of them). Typecheck and lint pass locally.

### DIFF
--- a/.changeset/uselayer-stable-ref-and-api.md
+++ b/.changeset/uselayer-stable-ref-and-api.md
@@ -1,0 +1,9 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] `useLayer({mode: 'context'})` now returns a stable trigger ref callback and a stable API object when nothing exposed by it has changed. Previously the trigger ref was recreated on every render, so React detached and reattached an unchanged trigger element (removing and reapplying its CSS anchor name for no reason), and the returned object (both `context` and `fixed` modes) was rebuilt every render, defeating memoization in consumers like Carousel, ContextMenu, Popover, Tooltip, HoverCard, and BaseTypeahead.
+
+@nynexman4464
+
+@HelloOjasMutreja
--- a/packages/core/src/Layer/useLayer.test.tsx
+++ b/packages/core/src/Layer/useLayer.test.tsx
@@ -70,6 +70,40 @@ function ContextLayerHarness({
   );
 }
 
+/**
+ * Reports the full context-mode API object on every render, so tests can
+ * compare identity across rerenders (issue #5271).
+ */
+function ContextStabilityHarness({
+  onLayer,
+  onShow,
+}: {
+  onLayer: (layer: ContextLayerReturn) => void;
+  onShow?: () => void;
+}) {
+  const layer = useLayer({mode: 'context', onShow});
+  onLayer(layer);
+  return (
+    <>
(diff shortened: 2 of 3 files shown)


HelloOjasMutreja:
Superseded by #5272, which merged the same fix — including the same wrapper-memoization gap I found while rebasing this (the public useLayer wrapper stripping wasJustDismissed via a fresh spread on every call, defeating the internal memo). Closing to avoid duplicate work.

PR #5286, opened 2026-08-21
fix(cli): keep blog adapter timeout active through body reads

### LINKED ISSUE #5249: [Bug] CLI blog body reads bypass the timeout and size cap
## Description

The shared CLI blog adapter stops its 15-second abort timer as soon as fetch returns response headers. The later response-body read is therefore unbounded in time. It also buffers the entire body before checking the 5 MB limit, so the limit does not cap how much data is read into memory.

Expected: the timeout remains active through body consumption, and reading stops as soon as the decoded response exceeds 5 MB.

Current source: packages/cli/api/blog/_adapter.mjs:34-73. Both blog.list and blog.detail use this path.

## Reproduction

1. Stub fetch with an OK Response whose headers resolve immediately but whose ReadableStream never closes.
2. Call blog.list or blog.detail.
3. Advance past 15 seconds. The call remains pending because the abort timer was cleared before res.text runs.

A second stub can stream more than 5 MB and observe that every chunk is consumed before ERR_FETCH_FAILED is thrown.

## Astryx Version

@astryxdesign/cli@0.4.5

## Environment

Node.js 24 on macOS

HelloOjasMutreja:
Fixes #5249.

## Problem

The shared CLI blog adapter (`packages/cli/api/blog/_adapter.mjs`, used by both `blog.list` and `blog.detail`) had two related issues in `fetchText`:

1. It cleared its 15-second abort timer as soon as `fetch()` returned response headers, before reading the body. A response whose headers arrive promptly but whose body stream stalls afterward has no timeout protecting it, and the call hangs indefinitely.
2. It read the body via `res.text()`, which buffers the entire response into memory before the code ever checks the 5 MB limit. The limit rejects an oversized response only after it has already been fully read, so it doesn't actually cap how much data gets buffered.

## Fix

- The abort timer is now cleared in a `finally` block after the read completes (success or failure), so it stays active through body consumption, not just through the initial `fetch()` call.
- The body is read via `res.body`'s stream where available (`getReader()`, decoding chunk by chunk), checking the running decoded size after each chunk and cancelling the read as soon as it exceeds 5 MB, instead of buffering the full response first. Falls back to `res.text()` when a runtime or test stub doesn't expose a streamable `body` (rare in practice; every real `fetch()` response has one), still enforcing the cap after the fact in that case.

## Verification

Added two regression tests matching the issue's own reproduction:

- **Timeout stays active through body consumption**: a stubbed `fetch` resolves immediately but its response body never closes. Using fake timers, advancing past the 15s timeout causes the call to reject with `ERR_FETCH_FAILED` rather than hang. The stub's stream reacts to the request's `AbortSignal` the way a real fetch response body does (rejects pending reads on abort), so this faithfully exercises the same mechanism a real network response uses.
- **Size cap stops the read early**: a stubbed response streams 5 chunks of 2 MB each (10 MB total) against the 5 MB cap. The call rejects with `ERR_FETCH_FAILED`, and the stub's pulled-chunk counter confirms fewer than all 5 chunks were consumed, proving the read stopped early rather than buffering the whole body before checking.

Both new tests fail against the pre-fix implementation (verified by temporarily reverting just the implementation file and re-running) and pass with the fix. Full blog test suite (22 tests across `_adapter`, `blog`, `blog.detail`, `blog.list`) passes. Lint required adding `ReadableStream`/`DOMException` to the CLI's ESLint global allowlist (needed by the new streaming-body test coverage; both are standard globals already implicitly relied on via `fetch`/`AbortController`). Typecheck (`tsconfig.strict.json`) shows no new errors; the pre-existing failures it reports are all in unrelated template/theme assets.

### DIFF
--- a/.changeset/cli-blog-adapter-timeout-and-size-cap.md
+++ b/.changeset/cli-blog-adapter-timeout-and-size-cap.md
@@ -1,0 +1,9 @@
+---
+'@astryxdesign/cli': patch
+---
+
+[fix] The shared CLI blog adapter (`blog.list`, `blog.detail`) cleared its 15-second abort timer as soon as `fetch` returned response headers, leaving the later body read unbounded in time, and buffered the entire response before checking the 5 MB size limit, so the limit didn't actually cap how much was read into memory.
+
+The abort timer now stays active through body consumption. The body is read as a stream where available, checking decoded size after each chunk and aborting the read as soon as it exceeds the limit, instead of buffering the full response first.
+
+@harjothkhara
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -511,6 +511,8 @@ export default defineConfig(
         structuredClone: 'readonly',
         fetch: 'readonly',
         AbortController: 'readonly',
+        ReadableStream: 'readonly',
+        DOMException: 'readonly',
       },
     },
     rules: {
--- a/packages/cli/api/blog/_adapter.mjs
+++ b/packages/cli/api/blog/_adapter.mjs
@@ -27,6 +27,67 @@ const MAX_BYTES = 5 * 1024 * 1024; // 5 MB — a blog feed is never larger.
  */
 export const FEED_URL = new URL('/rss.xml', SITE_URL).toString();
 
+/** @param {string} url @param {unknown} e @returns {AstryxError} */
+function toFetchError(url, e) {
+  const reason =
+    e instanceof Error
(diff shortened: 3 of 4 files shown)


cixzhang (approved):
Thanks, this is good. The stalled body now times out and the oversized stream stops early without changing the CLI error contract.

[Reviewed by Robohands]

PR #5287, opened 2026-08-21
fix(selector): block popup activation while isLoading is true

athz:
## Summary

`Selector`, `MultiSelector`, and `ComplexSelector` allow opening the popup while `isLoading` is true. Because the empty state is gated on an active search query, this presents a completely blank panel — no options, no message, no spinner inside the panel.

This PR makes `isLoading` block popup activation the same way `isDisabled` does, but without the disabled visual treatment:

- Trigger stays focusable (not removed from tab order)
- No greyed-out appearance
- `aria-busy` + field-level spinner still signal the pending state
- Only activation (click, ArrowDown, Space/Enter) is suppressed

## Changes

- **`packages/core/src/Selector/hooks.ts`** — `useCombobox` accepts `isLoading`; combines with `isDisabled` into an `isInactive` guard on `onTriggerClick` and `onKeyDown`.
- **`packages/core/src/Selector/Selector.tsx`** — passes `isLoading` to `useCombobox`; adds `isLoading` to the typeahead key-down guard.
- **`packages/core/src/MultiSelector/hooks.ts`** — same pattern for `useMultiCombobox`.
- **`packages/core/src/MultiSelector/MultiSelector.tsx`** — passes `isLoading` to `useMultiCombobox`.
- **`packages/core/src/ComplexSelector/ComplexSelector.tsx`** — adds `isLoading` to inline activation guards (`handleTriggerClick`, `useImperativeHandle`, `onKeyDown`).
- **Tests** — 6 new tests across Selector and MultiSelector confirming:
  - Click does not open when loading
  - ArrowDown does not open when loading
  - Trigger remains focusable (not disabled)

## Reproduction

```tsx
<Selector label="Owner" options={[]} isLoading hasSearch placeholder="Loading owners…" />
```

Before: opens a completely blank popup.
After: trigger stays inert until loading completes.

### DIFF
--- a/packages/core/src/Selector/hooks.ts
+++ b/packages/core/src/Selector/hooks.ts
@@ -157,6 +157,7 @@ interface UseComboboxOptions {
   selectableItems: SelectorOptionData[];
   value?: string;
   isDisabled?: boolean;
+  isLoading?: boolean;
   isOpen: boolean;
   hasSearch?: boolean;
   onOpen: () => void;
@@ -201,6 +202,7 @@ export function useCombobox({
   selectableItems,
   value,
   isDisabled = false,
+  isLoading = false,
   isOpen,
   hasSearch = false,
   onOpen,
@@ -210,6 +212,7 @@ export function useCombobox({
   onSearchSeed,
   listboxId,
 }: UseComboboxOptions): UseComboboxResult {
+  const isInactive = isDisabled || isLoading;
   const [highlightedIndex, setHighlightedIndex] = useState<number>(-1);
 
   const getItemId = useCallback(
@@ -244,7 +247,7 @@ export function useCombobox({
   );
 
   const onTriggerClick = useCallback(() => {
-    if (isDisabled) {
+    if (isInactive) {
       return;
     }
     if (isOpen) {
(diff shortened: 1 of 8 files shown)


packages/core/src/Selector/hooks.ts
```diff
@@ -269,7 +272,7 @@ export function useCombobox({
 
   const onKeyDown = useCallback(
     (e: React.KeyboardEvent) => {
-      if (isDisabled) {
+      if (isInactive) {
```
cixzhang, line 275: Hmm, this also catches an already-open panel — refetch mid-list and arrows stop moving the highlight.

cixzhang (changes requested):
Thanks — the blank panel is real and the diagnosis is right.

I'd rather the panel stay open and say it's loading than refuse to open. A trigger that ignores clicks and keys is jank for everyone, and worse with a screen reader — nothing announces why, and there's no `aria-disabled` to explain it.

CommandPalette already does this and is probably the shape to copy. It never guards activation; it announces `@astryx.commandPalette.loading` through the same `useAnnounce` polite region Selector uses (`CommandPalette.tsx:380`), and while a fetch is in flight it client-filters the results it already has (`:383`) so the list never goes blank. That last part is the real fix here — the empty panel is a symptom of dropping results we already had.

It also covers `options={[]}` with the fetch finished, which the guard never reaches.

Happy for this to be a separate PR if you'd rather keep this one small. [Discord](https://discord.com/invite/XnsUcFykEP) if you'd like to talk it through.

[Reviewed by Robohands]

cixzhang:
Checking in on Selector loading behavior: the current review needs existing results to remain usable with loading announced, rather than disabling activation. Would you like to continue, have us carry it, or close this for now?

athz:
On the EPS side we are currently disabling the loading selector. @cixzhang

cixzhang:
I think disabling is probably the wrong thing to do for isLoading especially as a general application -- loading state does not necessarily mean that contents are loading. You can use isReadOnly instead now to avoid the disabled appearance and prevent the popover and for more accurate accessibility signaling.

PR #5288, opened 2026-08-21
fix(markdown): forward the rest of BaseProps on Markdown and List

lexs:
Split per review: this PR is now **only the prop forwarding**, rebased onto current `main`.

`Markdown` and `List` both declare `BaseProps`, and `BaseProps` documents that `data-*`, `aria-*` and `role` reach the element — but each destructured a fixed set of props and forwarded only `data-testid`. So `<Markdown aria-label="Release notes">` named nothing, and a list could not be labelled by a heading it did not render itself.

Consumer props spread first, so what a component sets for itself still wins: the block root stays `role="document"`, the list keeps the explicit `role="list"` that restores Safari/VoiceOver announcements, and a list rendering its own header keeps that association. That `aria-labelledby` is now written only when the header exists — writing `undefined` unconditionally erased a label a consumer pointed at their own heading.

`data-align` and `data-delimiter` are gone from this PR. Detail in the review thread, but briefly: the introspection need behind them is now served by `parseMarkdown(src, {sourceRanges: true})` (#5290), and the rendering half you identified is a separate fix.

Tests: forwarding, and that a consumer prop cannot displace the role or the header association, on both components and on the inline root.


cixzhang (commented):
Thanks — and the two halves of this land differently, so taking them separately.

**The prop forwarding is right and I'd take it today.** `BaseProps` has declared `data-*`, `aria-*` and `role` as passthrough all along, and `List` and `Markdown` were quietly dropping them — so `<Markdown aria-label="Release notes">` reached nothing. That's a real accessibility bug on two shipped components, and your fix matches what `TableCell` already does, consumer props first so the component's own attributes still win.

**`data-align` and `data-delimiter` I'd like to hold.** Not because of the code — because a consumer is meant to read them, and that read has no type, no declaration and no drift guard. `BaseProps` also lets a consumer pass any `data-*` of their own, so the component and the caller would be writing into one shared namespace with nothing partitioning it. And it's the same question you raised on [#5290](https://github.com/facebook/astryx/pull/5290) and deliberately left to us — escapes, fence characters and heading depth all want the same answer, so it's worth settling once rather than four times.

Would you split them? The forwarding fix can merge on its own, and the attributes can follow whichever shape we pick.

[Reviewed by Robohands]

cixzhang:
Following up on my own comment — I think there's a real bug underneath this, and it's better than the attribute.

`cellAlignStyles` (`Markdown.tsx:238`) only defines `center` and `end`. There is no `start`, so an explicitly left-aligned column — `| :--- |` — gets no style at all and renders identically to `| --- |`. In LTR that looks correct by accident, because the default is already `start`. **In RTL it is wrong**: the author wrote `:---` to mean left, and the cell inherits `start`, which is right.

So the parser reads the alignment and the renderer drops one third of it. That's the thing to fix — add the `start` case and apply it, the same way `center` and `end` already are. `data-align` doesn't render anything; it hands the alignment to a consumer to apply, when the component should be applying it itself.

Same shape for `data-delimiter`, worth checking: an `<ol>` written `1)` renders as `1.` today. If the delimiter is meant to survive, that's `list-style-type` on the element, not an attribute describing what we chose not to do.

Sorry for the two-step. The forwarding fix still stands on its own and I'd take it today.

[Reviewed by Robohands]

packages/core/src/Markdown/Markdown.tsx
```diff
@@ -1470,6 +1471,7 @@ function renderBlock(
                   <TableHeaderCell
                     // eslint-disable-next-line @eslint-react/no-array-index-key -- markdown table columns are positional by definition
                     key={i}
+                    data-align={node.alignments[i] ?? undefined}
```
cixzhang: This is where the alignment should be applied, not described — `cellAlignStyles` has no `start` case, so an explicit `:---` column renders like an undeclared one (wrong in RTL).

cixzhang (changes requested):
Sorry for the back and forth — reading this again, I think there are two different problems here and it's worth naming which one we're solving.

**Is this about rendering, or about introspection?**

If it's rendering, it isn't fixed here. `cellAlignStyles` (`Markdown.tsx:238`) has only `center` and `end` — no `start` — so `| :--- |` gets no style and renders identically to `| --- |`. That looks fine in LTR by accident, since the default is already `start`, but in RTL an explicitly left-aligned column renders right. That's a real bug with a real reader, and the fix is applying the alignment, not describing it. Same question for the delimiter: an `<ol>` written `1)` renders `1.` today.

If it's introspection — reading the source back out of the DOM — I don't think that belongs in Markdown's contract as an undisclosed default. Every consumer who selects on it freezes it, and nothing declares it or guards it against drift. `components` already hands renderers what they need at render time, typed; `heading` gets a generated `id` for exactly this reason. It has no `table`/`tableCell`/`list` entry yet, and extending it looks closer to what you want.

**The prop forwarding is unaffected by any of this and I'd take it today** — `BaseProps` has promised `aria-*` passthrough all along and both components were dropping it. Happy to merge that on its own if you split it out.

Which problem were you solving? If both, they probably want to be two PRs.

[Reviewed by Robohands]

lexs:
Both, and you were right that they are two problems — split.

**This PR is now the forwarding fix only**, rebased onto current `main`. `data-align` and `data-delimiter` are gone from it.

**The rendering half is #5444.** Applying the alignment, as you said, rather than describing it: `:---` renders `left` and `---:` renders `right`, both physical, so an explicitly left-aligned column stops landing on the right in RTL. That also moves `---:` off the logical `end` it was on — the same bug seen from the other side — which is the one behaviour change worth your eye.

**On the delimiter, you are right that it is a rendering gap too, and it is a real one here.** An HTML `<ol>` cannot express `1)`, but Astryx does not use one: `ListItem` draws the marker itself, `content: counter(astryx-list) "."` hard-coded. So a list authored `1)` renders `1.` and the design system is the thing that decided that. Fixing it means a delimiter on `List` reaching the marker style — a public prop, not just a Markdown change — so I left it out of #5444 rather than bundle an API addition into a bug fix. Happy to open it if you want it.

**On the introspection question — you can drop it.** `parseMarkdown(src, {sourceRanges: true})` (#5290, shipped in 0.5.0) turned out to answer it better than any attribute would: a consumer holding the source slices the block verbatim, so escapes, fence characters, heading depth beyond the clamp and alignment all survive without anything being read back out of the DOM. That is the shape our own consumer is now being rewritten onto. So there is no undisclosed contract to freeze, and no fourth round of the same question.

cixzhang (approved):
Thanks — this is the split we asked for, and the forwarding half is clean. Clearing my changes-requested and taking it.

[Reviewed by Robohands]

cixzhang (approved):
Re-approving after the branch update; the reviewed diff is unchanged.

[Reviewed by Robohands]

cixzhang (approved):
Re-approving after the empty CI-retrigger commit; the reviewed diff is unchanged.

[Reviewed by Robohands]

PR #5289, opened 2026-08-21
fix(cli): rank exact keyword phrase matches above token-sum scores

### LINKED ISSUE #5239: [Bug] astryx search: Outline is unreachable by its own `table of contents` keyword
### Description

`astryx search "table of contents"` does not return `Outline` at all — even though
`'table of contents'` is a literal entry in `Outline`'s own `keywords` array, and the first
seven words of its description are *"A table-of-contents sidebar for documentation pages"*.
`Table` ranks; `Outline` is absent.

In practice `Outline` is only findable if you already know the word "outline".

I hit this building an in-page table of contents with anchor links for a docs-style page. I
searched for an existing component first, found nothing, and hand-rolled a stack of links plus
a scroll-position helper. A colleague asked "is there really no component for this?" — there
is, and it is considerably better than what I wrote: scroll-spy, sliding indicator, single tab
stop with arrow-key navigation, a labelled nav landmark, and nested scroll container support.
Both of my hand-rolled components were deleted.

**Expected:** a query that exactly matches a declared keyword returns the component that
declares it, at or near the top.
**Actual:** it is absent from the results entirely.

#### The metadata is already correct — this is a ranking bug

`Outline.doc.mjs` already declares:

```js
keywords: [
  'outline',
  'table of contents',   // exact phrase searched -> returns nothing
  'toc',                 // ranks 1st
  'heading navigation',  // exact phrase searched -> returns nothing
  'scroll spy',          // ranks 4th
  'documentation',
  'anchors',             // ranks 1st
  'sliding indicator',   // ranks 3rd
],
```

So this is not a request to add keywords. Three of the eight declared keywords fail to
retrieve their own component, and the two that fail hardest are the most natural things
someone would type.

It is not multi-word keywords in general — most work:

| query (a multi-word keyword of another component) | owner's rank |
|---|---|
| `copy button` → useClipboard | 2nd |
| `markdown renderer` → Markdown | 1st |
| `mobile drawer` → useAppShellMobile | 4th |
| `data visualization` → useTheme | 1st |

The pattern is that **a keyword phrase loses when one of its words is, or prefixes, another
component's name.** "table of contents" tokenizes to `table` + `contents`; `Table` is an exact
name hit and the `documentation-*` templates absorb the rest, so the component holding the
exact full phrase as a keyword is pushed off the list entirely. "heading navigation" loses the
same way to the `*Nav` family.

An exact whole-phrase match against `keywords` currently ranks below a single-token partial
match against a component name.

#### A second, narrower bug: function words are treated as required terms

`on this page` returns nothing at all. Probing it:

| query | results |
|---|---|
| `page` | 20 |
| `this page` | 1 |
| `on page` | 5 |
| `on this` | 0 |
| `on this page` | 0 |
| `on this page please` | 0 |

Matching appears to be conjunctive across terms, and common function words ("on", "this",
"please") are required terms matching nothing, so a single one zeroes the result set.

I checked whether empty results are a widespread problem before reporting one, and they are
not: across 40 realistic queries (`sticky sidebar`, `infinite scroll`, `skeleton loader`,
`command palette`, `drag and drop`, …) exactly one returned zero — `on this page`. Empty is
also sometimes correct: `zzzzqqqx` returns nothing, which is right. So this is specifically
about function words being required, not about empty results in general.

#### Suggested fixes, cheapest first

1. **Rank exact whole-phrase `keywords` matches above partial name-token matches.** This alone
   fixes `table of contents` and `heading navigation`, with no metadata changes anywhere.
2. **Drop or downweight common function words** rather than requiring them — or relax to a
   ranked OR when the conjunctive match yields zero. Nonsense queries should still return
   nothing.
3. **Consider a category-scoped browse.** `Outline` is `category: 'Navigation'`, exactly the
   right shelf, but there is no obvious way to browse by category from the CLI;
   `astryx component --category Navigation` would have found it in one step.
4. Optionally add `'on this page'` and `'in-page navigation'` as keywords — that is what docs
   sites actually label this control — though it is a band-aid if (1) lands.

#### Why it is worth fixing

The CLI's own guidance is "discover, don't guess", and I followed it: search first, then
build. Search said the component did not exist, so I wrote a worse version of one that did,
plus a second component for scroll behaviour `Outline` already owns. That is precisely the
failure mode good search metadata prevents — and here the metadata was already correct. Only
the ranking let it down.

### Reproduction

```bash
pnpm exec astryx search "table of contents"
# expected: Outline (it declares 'table of contents' as a keyword)
# actual:   table, documentation-design, Table, detail-page, documentation — no Outline
```

Queries I ran before writing any code, in order:

| query | top results | Outline? |
|---|---|---|
| `table of contents navigation` | List, detail-page, useGridFocus, useListFocus, documentation, table | absent |
| `table of contents` | table, documentation-design, **Table**, detail-page, documentation | absent |
| `on this page` | *(no results at all)* | absent |
| `anchor links to page sections` | Link, Popover, useLayer, usePopover, useHoverCard, Section | absent |
| `in-page navigation jump to section` | Section, AppShell, Breadcrumbs, ClickableCard, CommandPalette, Link | absent |
| `sidebar heading navigation` | AppShell, MobileNav, SideNav, Breadcrumbs, ClickableCard | absent |
| `scroll to anchor position restore` | useLayer, useScrollLock, Carousel, Link, … Outline | 8th |
| `scroll element into view on hash change` | Carousel, useScrollLock, useScrollOverflow, AlertDialog, … Outline | 10th |

Six of eight natural phrasings return it nowhere. In the two where it does appear it ranks
below `Carousel` and `AlertDialog`.

Queries that do work — all of which require already knowing the name:

| query | top results |
|---|---|
| `outline` | **Outline**, TreeList, … |
| `document outline` | **Outline**, documentation-design, … |
| `toc` | **Outline**, shell-nav, Tab |
| `anchors` | **Outline**, Link, Popover |

### Astryx Version

`@astryxdesign/cli@0.4.1`, `@astryxdesign/core@0.4.1`

### Environment

CLI (`pnpm exec astryx search`), Node 22, macOS and Linux — not browser-dependent.

HelloOjasMutreja:
Fixes #5239 (the first, cheapest of the issue's suggested fixes: "rank exact whole-phrase keyword matches above partial name-token matches").

## Problem

`astryx search "table of contents"` never surfaced `Outline`, even though `Outline.doc.mjs` declares `'table of contents'` verbatim as a keyword. Traced precisely: `scoreCandidate` correctly computes an exact-keyword score of 90 for Outline against that query. The bug is that `scoreQuery`'s multi-word token-sum path (which scores `"table"` and `"contents"` as separate tokens against every OTHER candidate) regularly produces scores above 90, since dozens of Table-related templates and components each match one of those two common words independently. With the default `--limit 20`, roughly 53 other candidates outscored Outline's exact match, pushing it out of the results entirely.

Same root cause for `"heading navigation"` against Outline's `'heading navigation'` keyword.

## Fix

When the full, untokenized query exactly matches a candidate's name or a declared keyword verbatim (`scoreCandidate` score 90 or 100, the only two outcomes at or above that mark), promote it to a reserved top tier (`score + 100`) that's safely above the token-sum path's mathematical ceiling (~151: 100 avg + 36 matched-count bonus + 15 coverage bonus). This is a targeted priority rule, not a rebalancing of the existing 0-100 scale, so it can't regress any query that doesn't have a genuine exact whole-phrase match.

## Scope (intentionally narrow)

This addresses exactly suggestion #1 from the issue, "no metadata changes anywhere." The issue's second, narrower bug ("on this page" returning zero results because a query consisting entirely of stopwords has no fallback) is a separate, vaguer problem the issue itself frames as needing more judgment ("relax to a ranked OR... though it is a band-aid"), so I left it out of this PR rather than bundle a riskier, less-verified change into the same fix. Happy to take that on separately if useful.

## Verification

Ran every query from the issue's own reproduction table against the real component registry, before and after:

| query | before | after |
|---|---|---|
| `table of contents` | absent | **Outline, rank 1** |
| `heading navigation` | absent | **Outline, rank 1** |
| `copy button` (already worked) | useClipboard, rank 1 | useClipboard, rank 1 (unaffected) |
| `markdown renderer` (already worked) | Markdown, rank 1 | Markdown, rank 1 (unaffected) |
| `mobile drawer` (already worked) | useAppShellMobile, rank 1 | useAppShellMobile, rank 1 (unaffected) |
| `data visualization` (already worked) | useTheme, rank 1 | useTheme, rank 1 (unaffected) |
| `zzzzqqqx` (nonsense) | 0 results | 0 results (unaffected, confirms no over-triggering) |

Added two regression tests (a new "exact keyword phrase outranks incidental token matches" describe block, matching the file's existing em-dash naming convention for its own describe blocks) asserting Outline ranks first for both previously-failing queries, plus a nonsense-query sanity check. Both fail against the pre-fix implementation (verified by temporarily reverting just the implementation file) and pass with the fix.

Full `search` and `build` API test suites pass (19 tests; `build`'s own ranking floors in `kit.mjs` are all `>=` threshold checks, so this monotonic score increase for exact matches can only make `build` more confident about a genuine keyword match, never less). Lint and typecheck (`tsconfig.strict.json`) show no new issues; a pre-existing, unrelated failure in `component.test.mjs` (a case-sensitivity assertion, confirmed failing identically on a clean baseline with my changes fully reverted) is untouched by this diff.

### DIFF
--- a/.changeset/cli-search-exact-keyword-phrase-ranking.md
+++ b/.changeset/cli-search-exact-keyword-phrase-ranking.md
@@ -1,0 +1,9 @@
+---
+'@astryxdesign/cli': patch
+---
+
+[fix] `astryx search` (and `astryx build`, which shares its ranking) never surfaced a component whose exact multi-word keyword phrase was searched, if enough other unrelated candidates happened to each contain one of the query's individual words. Searching `"table of contents"` returned no results for `Outline`, even though `Outline.doc.mjs` declares `'table of contents'` verbatim as a keyword, because `Table`-related templates each matched `table` and `contents` separately and their combined per-word score outranked Outline's single exact match.
+
+A query that exactly matches a candidate's declared keyword (or name) verbatim is now promoted to a top-tier score, so it always outranks a candidate that only coincidentally contains several of the query's individual words. Single-word queries and queries that don't exactly match a keyword are unaffected.
+
+@nynexman4464
--- a/packages/cli/api/search/search.mjs
+++ b/packages/cli/api/search/search.mjs
@@ -221,6 +221,21 @@ export function scoreQuery(term, tokens, candidate) {
     return single;
   }
 
+  // The full (untokenized) query matching a candidate's name or a declared
+  // keyword VERBATIM — full.score 90 or 100, the only two scoreCandidate
+  // outcomes at or above that mark — is a deliberate, explicit label the
+  // author chose for exactly this multi-word concept. Promote it to a
+  // reserved top tier, safely above the token-sum path's ceiling below
+  // (~151: 100 avg + 36 bonus + 15 coverage), so it always outranks a
+  // candidate that merely happens to contain several of the query's
+  // individual words. Without this, "table of contents" never surfaces
+  // Outline (which declares that exact phrase as a keyword) because dozens
+  // of Table-related templates each match "table" and "contents" separately
+  // and accumulate a higher raw score (#5239).
+  if (full && full.score >= 90) {
+    return {score: full.score + 100, reason: full.reason};
+  }
+
   // Multi-word natural language: score each content token, counting only
   // strong hits, then reward coverage so candidates matching more terms win.
   let sum = 0;
--- a/packages/cli/api/search/search.test.mjs
+++ b/packages/cli/api/search/search.test.mjs
@@ -58,6 +58,27 @@ describe('search leaf — --type filter', () => {
   });
 });
 
+describe('search leaf — exact keyword phrase outranks incidental token matches (issue #5239)', () => {
+  it('surfaces Outline for its own declared keyword "table of contents", ranked first', async () => {
+    // Before the fix, "table" and "contents" each separately matched dozens
+    // of unrelated Table-related templates by coincidence, and their
+    // combined token-sum score outranked Outline's single exact match —
+    // pushing it out of the results entirely at the default limit.
+    const r = await search('table of contents', {cwd});
+    expect(r.data.results[0]?.name).toBe('Outline');
+  }, SLOW);
+
+  it('surfaces Outline for its own declared keyword "heading navigation", ranked first', async () => {
+    const r = await search('heading navigation', {cwd});
+    expect(r.data.results[0]?.name).toBe('Outline');
+  }, SLOW);
+
+  it('still returns no results for a nonsense query (the fix does not loosen matching)', async () => {
+    const r = await search('zzzzqqqx', {cwd});
+    expect(r.data.results).toEqual([]);
+  }, SLOW);
+});
+
 describe('search leaf — error paths (pinned)', () => {
   it('throws ERR_INVALID_ARGUMENT when the query is empty/whitespace', async () => {
     await expect(search('   ', {cwd})).rejects.toMatchObject({
--- a/packages/cli/api/search/search.test.mjs
+++ b/packages/cli/api/search/search.test.mjs
@@ -62,7 +62,7 @@ describe('search leaf — exact keyword phrase outranks incidental token matches
   it('surfaces Outline for its own declared keyword "table of contents", ranked first', async () => {
     // Before the fix, "table" and "contents" each separately matched dozens
     // of unrelated Table-related templates by coincidence, and their
-    // combined token-sum score outranked Outline's single exact match —
+    // combined token-sum score outranked Outline's single exact match,
     // pushing it out of the results entirely at the default limit.
     const r = await search('table of contents', {cwd});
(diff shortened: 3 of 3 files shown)


nynexman4464:
Thanks for looking into this. I don't think #5239 was fully specified enough to be ready for implementation. Before changing the ranking, we'd want regression coverage showing that existing good searches don't get worse, and I'm not sure we have a solid search-quality benchmark yet. I wouldn't spend any more of your tokens on this for now.

cixzhang (approved):
Builders can find Outline by its declared phrases. Both regressions move from absent to first; focused suites pass. `mobile drawer` also improves from fifth to first.

[Reviewed by Robohands]

PR #5290, opened 2026-08-21
feat(markdown): opt-in source ranges on parsed blocks

lexs:
## What

`parseMarkdown(source, {sourceRanges: true})` gives every top-level block a `range` — the
`[start, end)` offsets it occupies in the string that was passed in:

```ts
const blocks = parseMarkdown(source, {sourceRanges: true});
source.slice(...blocks[0].range!); // the block's own markdown, verbatim
```

Off by default, and the field is absent unless asked for, so no existing node, snapshot or
comparison changes.

## Why

I render agent answers with `<Markdown>` and need the markdown *behind* what is on screen —
copying a text selection out of a chat transcript as markdown. Today that means walking the
rendered DOM and rebuilding the markdown from it, which is lossy in ways slicing the source
never is: escapes (`\*literal\*` comes back as emphasis), the exact emphasis and fence
characters, heading depth past the h6 clamp, table alignment, whether a bare fence said
`plaintext`. With ranges, a consumer that still holds the source slices it instead.

## The parts that aren't obvious

- **Link reference definitions.** `parseMarkdownImpl` parses `cleaned` — the input with
  definition lines stripped — so offsets taken there drift for everything after the first
  definition. `extractLinkDefinitions` now also returns a line map and the ranges are
  reported against the string the caller passed.
- **Streaming.** `parseMarkdownIncremental` parses three different slices (settled prefix,
  delta, trimmed unsettled tail), so each is parsed with a base offset and the blocks come
  out absolute; a list whose halves arrive in different chunks and get merged spans both.
  Flipping the option invalidates the settled cache, the way `autolink` already does.
- **Fenced code owns its blank lines.** Block ends are derived positionally (a block runs to
  the line before the next one starts, blank lines trimmed), which is wrong for an
  unterminated fence mid-stream, so the fence states its own end.
- **Nested blocks carry no range.** A list item's or blockquote's children are parsed from
  text reassembled without their markers, so an offset into it wouldn't address the document.
  Say the word if you'd rather they carried one — it means threading a base offset through
  those two recursive parses.

## What I deliberately left out

Emitting the ranges from `<Markdown>` as a DOM attribute (`data-md-range` on each block,
behind a prop), which is what my consumer ultimately needs to map a DOM selection back to
source. I'd rather you pick that API than have me guess it — happy to follow up with it, or
to fold it into this PR if you tell me the shape you want.

## Test plan

`npx vitest run packages/core/src/Markdown packages/core/src/List packages/core/src/Outline`
— 380 pass. Nine new cases: every block type in one document, the link-definition mapping,
absolute offsets while streaming character by character, a merged list spanning both halves,
an unterminated fence keeping its blank lines, flipping the option on an existing state, and
a property check that every range slices to something that re-parses to the same node
(including a CRLF document).

### DIFF
--- a/.changeset/markdown-source-ranges.md
+++ b/.changeset/markdown-source-ranges.md
@@ -1,0 +1,32 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] Markdown: opt-in source ranges on parsed blocks
+
+`parseMarkdown(source, {sourceRanges: true})` now gives every top-level block a
+`range` — the `[start, end)` offsets it occupies in the source that was passed
+in — so a consumer holding that source can slice the original markdown for a
+block instead of reconstructing it from the node (or from the rendered DOM).
+Reconstruction is lossy in ways slicing is not: escapes, the exact emphasis and
+fence characters, heading depth beyond the clamp, and alignment all survive a
+slice unchanged.
+
+Off by default and absent unless asked for, so no existing node, snapshot or
+comparison changes.
+
+Two things the offsets get right that a naive implementation does not: link
+reference definitions are stripped before the block loop runs, and the ranges
+are reported against the string the caller passed rather than the stripped
+text; and `parseMarkdownIncremental` parses slices, so blocks report absolute
+offsets into the whole document as it streams — including a list whose halves
+arrived in separate chunks and were merged.
+
+A range covers a block's own lines verbatim, so slicing it and parsing the
+result gives the same node back.
+
+Blocks nested inside a list item or a blockquote carry no range: their children
+are parsed from text the parser reassembled with markers and `>` prefixes
+removed, so an offset into it would not address the document.
+
+@lexs
--- a/packages/core/src/Markdown/index.ts
+++ b/packages/core/src/Markdown/index.ts
@@ -25,6 +25,7 @@ export {
 export type {
   BlockNode,
   InlineNode,
+  SourceRange,
   ListItemNode,
   TableCellNode,
   TableAlignment,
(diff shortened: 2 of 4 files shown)


lexs:
A note on the `pr-a11y` red that was on this PR earlier, in case it comes back and confuses a reviewer.

It failed twice with one `[serious] color-contrast` violation on **Markdown / Streaming**, and the report names the node:

```
<span class="astryx1hc1fzr astryx19991ni astryx80gvsz …">brary</span>
```

That's a fragment of "library" caught mid-fade — one of the streaming fade-in spans, sampled by axe while it is still partially transparent. The story drives itself with

```js
const chunkSize = Math.floor(Math.random() * 8) + 2;
const delay = 30 + Math.random() * 60;
```

so where the audit lands in the animation differs run to run, which is why the same story passes on other PRs (and passed on #5288, which actually touches the component — this one is parser-only and changes no rendering).

I did fix one real thing on my side in the meantime rather than just re-rolling: `input.indexOf(unsettledText, …)` was running on every streamed chunk even with `sourceRanges` off. That is pointless work on the hot path — it is now gated on the option — but the violation is in the fade timing, not in anything the parser produced. Everything is green as of the latest push.

Happy to send a separate PR making that story deterministic (fixed chunk size and delay, or a seeded RNG) so the audit stops being a coin flip for anyone touching `Markdown` — just say the word.

cixzhang (approved):
Thanks — this is careful work, and the copy-out case is one we want. Merging.

Two things I'm taking on rather than asking you for. `SourceRange` wants to be an object rather than a tuple so line numbers can be added to it later without a break, and top-level-only means selecting a single bullet out of a list still can't be sliced — nested blocks need line spans, since a blockquote child's text isn't contiguous in the source. Both are ours, and both land before the next release.

[Reviewed by Robohands]

PR #5294, opened 2026-08-21
fix(cli): use the resolved owner package for a component's plain-text import hint

andrskr:
## Problem

`astryx component <Name>`'s plain-text output always showed `import {Name} from '@astryxdesign/core/...'`, even for a component owned by an integration package with no core equivalent. #4810 already fixed this for the JSON response (`withOwnership` in `api/component/_adapter.mjs` now resolves the specifier against the owning package's `exports` map), but the command's text formatter never used that value — it recomputed its own hint via `resolveImportPath(coreDir, resolvedName)`, which only ever resolves against core.

## Fix

`component.detail`'s text-output branch in `clients/cli/commands/component/index.mjs` now reads `result.data.import` (the value `withOwnership` already resolved correctly) instead of recomputing it. One line.

## Testing

Extended `import-hint-correctness.test.mjs` with a new `describe` covering integration-owned components end to end via `runCli`, using the same "root the fixture inside the package tree" pattern `doctor.test.mjs` already uses (Vite's `server.fs.allow` can't serve a real tmp-path config at runtime):

- `component <Name>` (full) shows the integration's own package, not `@astryxdesign/core`.
- `component <Name> --detail brief` shows the same.
- an integration whose `package.json` does export a matching subpath still resolves to that subpath (composes correctly with #4810's fix).

Confirmed all three fail without the one-line fix (reverted it locally and reran) and pass with it.

- `import-hint-correctness.test.mjs`: 497/497 passed.
- `component-ownership.test.mjs`, `component-package.test.mjs`, `component-resolution.test.mjs`: 44/44 passed.
- `eslint` on both changed files: clean.


cixzhang (dismissed):
Thanks, this fixes the misleading integration import hint by reusing the API’s resolved package. The focused regression and ownership suites pass.

[Reviewed by Robohands]

cixzhang (approved):
Thanks, this fixes the misleading import hint by reusing the package Astryx already resolved across full, compact, and brief output. Focused regression and ownership checks pass.

[Reviewed by Robohands]

cixzhang:
The failed-jobs retry on this exact head reproduced the same repository-owned CLI type-surface error from its old base (`api/search/search.mjs` is missing the later `matched`/`total` fields). Your import-hint change is not the failing code, but this head cannot clear required CI as-is. Please rebase or merge current `main` into the branch without expanding the PR; the resulting head will need fresh required checks and exact-head review.

PR #5295, opened 2026-08-21
fix(core): keep a no-scrim Bottom Sheet flush with the viewport on iOS

imdreamrunner:
Fixes two bottom-edge problems on iOS, one reported and one found on the way.

## 1. The page showed through Safari's address bar

Under a no-scrim (`hasScrim={false}`) sheet, page content was visible in the strip Safari's translucent bottom bar sits over. Reported from a physical iPhone.

**Cause.** The sheet hangs off the bottom edge of its positioner, and that positioner was `position: absolute` — so its bottom edge was the *dialog's* bottom edge, which is a computed `100dvh` length. On iOS that length goes stale while the address bar animates between expanded and compact. Two things then go wrong at once: the sheet sits a bar's height above the viewport bottom, and — because Safari only extends an overlay's surface behind the bar when the overlay is *flush* with the viewport — it stops extending, so the page shows through.

**Fix.** One line: the positioner is `position: fixed`.

```diff
  positioner: {
-   position: 'absolute',
+   position: 'fixed',
```

`fixed` is not a length. The compositor resolves it against the viewport at paint time, so there is no number to go stale and the sheet is flush in every bar state.

**Why it only affected no-scrim sheets.** A scrim-backed sheet calls `useScrollLock`, and the address bar only changes state on scroll — so with one open, the bar can never move and the length is never re-resolved. That asymmetry is what made this look like a rendering quirk of `hasScrim={false}` rather than a stale-length bug.

## 2. The home indicator ate the height budget

A `capped` sheet asked for 62dvh and then spent part of it on the indicator. A named budget now sizes the sheet's **content**: the gutter — `calc(16px + env(safe-area-inset-bottom, 0px))` — is added to the budget rather than taken out of it, and the height is clamped at `calc(100dvh - env(safe-area-inset-top, 0px))` so the taller sheet cannot push its rounded top edge under the status bar.

Both expressions live in custom properties rather than inline: StyleX's dev-time runtime rewrites `env()` with a fallback into invalid CSS inside a larger expression — the sheet's padding was already shipping mangled in dev as `calc(48px + env(0px * , * safe-area-inset-bottom))` — and folds a `min()` of two literals to one operand. Behind a variable both survive.

## Risk

- The positioner moves from `absolute` to `fixed`. It was already inside a `position: fixed` dialog spanning the viewport, so the resolved geometry is the same box; what changes is that it no longer depends on the dialog's height being current. Centring (`max-width: 640`), drag/snap transforms and the switcher all run through the same element and are covered by the existing 191 BottomSheet tests.
- Sheets are 16px taller than before with no bottom inset, `16px + inset` with a home indicator. Gesture and snap math read the rendered height, so detents follow automatically.
- No API change.

## 3. A fast flick flashed the page below the sheet

With #1 in place, a hard scroll could still show a sliver of page between the sheet and the bar for a frame or two before settling.

**Cause.** iOS repositions `fixed` elements a frame or two late during momentum scrolling. The sheet already overdraws `OVERSCROLL_PADDING` (48px) below the viewport, which absorbs a small lag; a fast flick outruns it.

**Fix.** The positioner paints a strip of the sheet's surface below its bottom edge, via `::after` — no box in the flex row, so the sheet's measured height (which the drag and snap maths read for their detents) is untouched. At rest the strip is below the viewport, where nothing is painted at all, so it costs nothing.

## Testing

- **Physical iPhone** — the reporter confirmed the steady-state bug (#1) is fixed. That is the only environment that reproduced it; the iOS 26.5 simulator does not, in any configuration tried (modal / non-modal / top-layer popover / with clipping ancestors).
- **The flick fix (#3) was measured by inducing the lag deterministically**, since the simulator's address bar will not change state under synthetic input, and a transient cannot be timed against a screenshot anyway. The harness translates the positioner up by a fixed amount — exactly what a lagging `fixed` element looks like — holds it there, forces page content magenta, and samples the bottom band full width:

  | induced lag | before | after |
  | --- | --- | --- |
  | 140px | **100% page visible** | 0% — covered |
  | 175px | — | 0% — covered |

  Far beyond any plausible momentum lag, which is a frame or two of travel.
- **Real iOS Safari, iOS 26.5 simulator, iPhone 17**, full-width pixel sampling against a loud page colour: with the sheet flush, the strip behind the bar carries the sheet's surface (0.7% page pixels, vs 20.5% with no sheet).
- **Chromium, emulated insets:** `capped` content is 409px both with a 34px home indicator emulated and without it (was 375px vs 409px).
- `BottomSheetSafeArea.test.tsx` (8 tests) pins the declarations that carry this, each with the measurement in a comment — they read like things to "fix", and no automated check can catch breaking them because no browser in CI has retractable chrome. Negative controls: `absolute` on the positioner, `100lvh` or `100%` on either dialog, inlining the `env()` into the padding rule, and reverting the budget math each fail the tests that cover them.
- 7,687 tests in core+lab, `pnpm lint:strict` (0 errors), full build, all seven typechecks, lab readiness, storybook build — green.

Background on the platform behaviour and the measurement traps involved: [P2471207218](https://www.internalfb.com/intern/paste/P2471207218/).


imdreamrunner:
**Do not trust the address-bar claims in the description yet — under investigation.**

The canvas-tint commit (717a0f18) does not solve the reported problem. It colours the background of the strip behind iOS Safari's bottom bar, but the page's *content* still paints on top of it, so page text remains visible there.

My verification of it was also unsound: I pixel-sampled a single column at 4% of the screen width, which landed in whitespace. Sampling the full width of the same screenshot shows 40% of one row is page text. Earlier geometry conclusions in the description used the same single-column method and are therefore unproven too.

Re-running the candidate matrix with full-width sampling. The budget/gutter part of this PR is measured differently and is unaffected.

imdreamrunner:
Superseding my earlier warning comment: the canvas-tint commit has been **reverted** and the branch squashed to a single commit. The PR now claims only the home-indicator budget fix, which is measured independently of any of the address-bar work.

Why the tint went: it colours the *background* of the strip behind the address bar, but page text draws on top of it, so it does not solve the reported problem — and it wrote to `documentElement.style` for no measured benefit.

Write-up of the address-bar constraint, aimed at readers with no context: P2471207218

imdreamrunner:
Closing: this does not fix the bug it set out to fix.

**The bug.** On iOS Safari, page content is visible in the strip behind the translucent bottom address bar under a no-scrim (`hasScrim={false}`) Bottom Sheet. Reported and reproducible on a physical iPhone, and intermittently in the iOS 26.5 simulator.

**Status: unfixed, cause not established.** Five approaches were tried and none survived verification:

| attempt | outcome |
| --- | --- |
| overlay spans `100lvh` | made it worse — overshooting loses Safari's flush extension |
| tint the page canvas with the sheet's surface | colours the strip's background, but page text still draws on top |
| non-modal sheet into the top layer (`showPopover`) | no effect |
| overdraw strip / box-shadow below the sheet | no effect on device |
| remove the sheet's `-48px` bottom margin so its box is flush | appeared deterministic 3/3, then the **control** for the same build returned the opposite result |

**Why none of it could be trusted.** The behaviour is not reproducible run-to-run: identical builds and identical configurations produced both COVERED and GAP across repeated runs, including no-op controls that change nothing. That points at **paint invalidation** — the strip is not repainted until something forces it — rather than a static CSS geometry problem, which also matches the reporter's observation that it only corrects itself "when I touch the page." Screenshot-based verification cannot distinguish a fix from a repaint, so every positive result here is suspect.

**What would actually move this forward:** Safari Web Inspector attached to a physical iPhone, watching the paint timeline on the live page. That shows whether the strip is being repainted at all, which no screenshot can. If it confirms paint invalidation, this is likely a WebKit bug to file rather than something to work around in the component.

Two unrelated improvements were developed on this branch and are **not** being landed with it, since they do not fix the reported bug: a home-indicator height-budget correction (`capped` content 375px → 409px on a phone with an indicator) and `position: fixed` on the sheet's positioner. The branch `sheet-safe-area` is left in place if either is wanted separately.

Investigation notes, including the measurement traps that produced several false positives: P2471207218

PR #5296, opened 2026-08-21
fix(core): don't format PlainDate with the runtime's locale

josephfarina:
## The bug

`plainDateFormat` formats with `new Intl.DateTimeFormat(undefined, options)`. An
`undefined` locale resolves to *the runtime's* locale, and the runtime is not the
same on both sides of an SSR boundary. On a host with no `LC_ALL` the server
resolves the CLDR root and emits `2026 M08` where the browser renders
`August 2026`, so React reports a hydration mismatch:

```
<DateInput label="End date" value="2026-08-22" ...>
  <Button><span> + August 2026  - 2026 M08
at <unknown> (https://react.dev/link/hydration-mismatch)
```

Every SSR'd `Calendar` / `DateInput` is affected. It does not need to be a date
picker the app directly renders: a shared navigation dependency can render a
`DateInput` on first paint, so an app can inherit the mismatch indirectly.

The issue reproduced across multiple generated apps. Downstream consumers had
already worked around it by documenting it as a known upstream warning or by
hardcoding an English formatter to avoid host-dependent output.

The existing `formatSharedDate` tests also assert exact strings like
`'Jan 25, 2026'`, so they pass only because CI happens to resolve English.

## The change

```diff
 export function plainDateFormat(
   pd: PlainDate,
   options: Intl.DateTimeFormatOptions,
+  locale: Locale = 'en',
 ): string {
-  return new Intl.DateTimeFormat(undefined, options).format(
-    plainDateToDate(pd),
-  );
+  return new Intl.DateTimeFormat(locale, options).format(plainDateToDate(pd));
 }
```

`'en'` is deliberate rather than arbitrary: it is what `useLocale()` already
returns outside a provider, so the default agrees with the rest of the i18n
contract instead of inventing a second one. The added parameter is the seam
`useLocale`'s own docstring already asks for —

> This is the approved way to read the locale that feeds a formatting helper
> (`plainDateFormat`, `formatInstant`, chart formatters, …)

— which today is impossible, because the helper takes no locale.

Non-breaking: every existing two-argument call keeps working, and gains
determinism.

## Follow-up I deliberately left out

The twelve call sites in `Calendar`, `useCalendarNavigation`, `DateRangeInput`
and `DateTimeInput` should thread `useLocale()` so calendars are correctly
localized as well as deterministic. That is a larger change across a hook and two
module-level helpers, and it wants a maintainer's opinion on where the provider
locale enters those files, so I have kept this PR to the defect. Happy to do it
in a follow-up if you tell me the shape you want.

## Test plan

`pnpm vitest run packages/core/src/utils/plainDate.test.ts` — 88 pass.
Three new cases: the `'en'` default, an explicit `'fr'` (`août 2026`), and the
determinism invariant that a two-argument call equals the explicit-`'en'` call
on any host.

`tsc --noEmit` and `eslint` clean.

I could not reproduce the root-locale resolution on macOS — this laptop resolves
`en-US` whatever `LC_ALL` says — so the server-side half of the evidence is the
production repro above rather than a local run.

### DIFF
--- a/.changeset/fix-plaindateformat-runtime-locale.md
+++ b/.changeset/fix-plaindateformat-runtime-locale.md
@@ -1,0 +1,24 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] plainDateFormat: don't format with the runtime's locale
+
+`plainDateFormat` passed `undefined` to `Intl.DateTimeFormat`, so the locale
+resolved to whatever the runtime happened to have. Across an SSR boundary that
+is not the same on both sides: a host with no `LC_ALL` resolves the CLDR root
+and emits `2026 M08` where the browser renders `August 2026`, and React reports
+a hydration mismatch for every SSR'd `Calendar` or `DateInput` — including ones
+the user never opened.
+
+The locale now defaults to `'en'`, matching what `useLocale()` returns outside a
+provider, and is accepted as an optional third argument so callers under an
+`InternationalizationProvider` can thread the provider locale:
+`plainDateFormat(pd, options, useLocale())`.
+
+Existing two-argument calls keep working and become deterministic. Formatting
+inside `Calendar`, `DateRangeInput` and `DateTimeInput` is unchanged in an
+English runtime; a non-English runtime now needs the explicit locale to render
+in its own language, which was the only way to get a stable SSR tree.
+
+@josephfarina
--- a/packages/core/src/utils/plainDate.test.ts
+++ b/packages/core/src/utils/plainDate.test.ts
@@ -592,10 +592,6 @@ describe('plainDateFormat', () => {
   });
 
   it('defaults to en, not the runtime locale', () => {
-    // The regression this guards: with `undefined` the month label resolved to
-    // whatever locale the runtime had, so an SSR host with no LC_ALL emitted
-    // the CLDR root form ("2026 M08") while the browser rendered "August
-    // 2026" — a hydration mismatch in every SSR'd Calendar/DateInput.
     expect(
       plainDateFormat({year: 2026, month: 8, day: 22}, DATE_FORMAT_MONTH_YEAR),
     ).toBe('August 2026');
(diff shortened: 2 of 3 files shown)


josephfarina:
@nynexman4464 wdyt?

josephfarina:
## Reproduced in production today, in a generated app

Dogfooding the live Nest Studio build flow this afternoon (2026-08-21), a
freshly generated Astryx app took a hydration error on its first paint. The
mismatch is this one, verbatim from the browser overlay:

```
+ August 2026
- 2026 M08
```

The offending subtree is an app-authored date field, not anything in the
shell:

```
<DateInput label="End date" value="2026-08-22" onChange={...}>
```

Server rendered `2026 M08`, browser rendered `August 2026`, so React threw the
tree away and regenerated it. Same root cause as the fix in this PR:
`plainDateFormat` calls `new Intl.DateTimeFormat(undefined, ...)`, and a server
with no `LC_ALL` resolves the CLDR root locale rather than `en-US`.

Two things this clarifies that I could not show before.

**It is not rare.** This was the first app I generated after the fix stack
landed internally, on an ordinary prompt that never mentioned dates. The agent
reached for a date picker on its own. Any generated app that does the same
hits it.

**Internal workarounds do not cover it.** We shipped a change that stops our
own nav from server-rendering a closed modal containing a `Calendar`, which
fixes the shell. It does nothing for an app that renders a `DateInput`
directly, because the formatting is what is wrong, not where it is mounted.
The fix has to be here.

imdreamrunner (approved):

cixzhang:
Thanks for chasing this down — the diagnosis is right and the `2026 M08` repro is the useful half.

#5120 landed the locale parameter two hours after you opened this, so after a merge what's left here is one line: `locale?: Locale` → `locale: Locale = 'en'`. Five call sites in the touch date picker never take a locale, so that line pins them to English: on a French phone the sheet goes from `janvier 2026` to `January 2026`, and no prop brings it back.

| before | after |
|---|---|
|  |  |

#5303 carries the identical line and threads all five.

I'd land it there rather than here — ok with that?

[Reviewed by Robohands]

cixzhang:
Inline, `packages/core/src/utils/plainDate.ts:304` — `locale: Locale = 'en',` — main already has this parameter (#5120); only the `= 'en'` default is new here.

nynexman4464:
Closing in favor of #5303, which preserves Joey Farina's original commits and changeset credit while completing the provider-locale, Gregorian-default, compatibility, and lint coverage. #5303 merged cleanly as 2458b03b35a45512b996b0c2c764d9530a36a052. Thanks for identifying and reproducing the hydration issue.

PR #5297, opened 2026-08-21
fix(ResizeHandle): keep the drag when the cursor crosses an embedded frame

cixzhang:
## Why

`ResizeHandle` started a drag on `pointerdown` and then listened on `window` for
`pointermove`/`pointerup`, without ever calling `setPointerCapture`. Without
capture the browser hit-tests every later pointer event, so the moment the
cursor crosses an `<iframe>` inside the resizable region those events are
dispatched into the **guest** document and the host window listener stops
hearing them. The drag dies mid-gesture while still looking armed. It bites any
consumer whose resizable region can contain a frame — a preview pane, an
embedded app.

The cursor leaves the handle more often than it sounds: the divider only tracks
the pointer while the panel can still follow it, so any drag past `maxSizePx`
(or down to `minSizePx`, or faster than layout) puts the pointer over the
content pane.

## Measured

New story `Lab/Resizable → WithEmbeddedFrame`, driven in Chromium with
Playwright: press on the handle, drag right until the panel hits its 520px max,
keep going 25 × 8px so the cursor is genuinely over the frame, release there.

| | before | after |
|---|---|---|
| pointermove events the host received while the cursor was over the frame | **0 / 25** | **25 / 25** |
| panel width after the drag (from 200px) | 200px — never moved | 520px — full travel, clamped at max |
| `pointerup` heard by the host | no | yes |
| `data-resizing` after release | still set | cleared |
| body `cursor` / `user-select` after release | `col-resize` / `none` | cleared |

The panel never moves at all in the "before" column because the grab zone is
biased onto the pill on the panel side, so the very first move of the drag
already lands the cursor over the frame.

## What

`pointerdown` now takes pointer capture on the grab zone, and the
move/up/cancel handlers moved onto that element.

**Where the listeners went, and why.** A captured pointer's events are
delivered to the capturing element and then bubble, so listeners on the element
and listeners on `window` would both fire — keeping both would double-handle,
and keeping only `window` would leave the drag at the mercy of any
`stopPropagation` between the handle and the top. They live on the element, the
same shape as `Slider` and `BottomSheet`.

Three behaviors the old code had, and where they went:

- **Unmount mid-drag.** Removing the capturing element implicitly releases the
  pointer and takes React's listeners with it, so an in-flight drag stops
  driving the region on its own; the unmount effect stays only to release the
  body `cursor`/`user-select` overrides, which do not live on the element.
- **`pointercancel`.** Still ends the drag without signalling a resize end.
  `lostpointercapture` routes to the same path — it also fires on the implicit
  release after every `pointerup`/`pointercancel`, by which point the drag is
  already cleared and it is a no-op.
- **Foreign pointers.** Capture is per-pointer, so the handlers ignore moves
  from a pointer that does not own the drag.

No API change; keyboard resizing, RTL delta mapping and the grab-zone geometry
are untouched.

## Testing

`ResizeHandle.test.tsx` covers capture taken on `pointerdown`, the drag driven
and ended from the element, `pointercancel`/`lostpointercapture` teardown,
foreign-pointer moves ignored, and unmount mid-drag. One `SideNav` test needed
updating — it dispatched its synthetic drag at `document`. Full `packages/core`
suite, typecheck and lint pass.

## Adjacent, not touched here

Two other drags in core hold the same window-listener shape and will drop the
same way over a frame: `useTableColumnResize` and the `Lightbox` pan. Left
alone so this stays one component — worth asking separately whether they should
all share one drag hook.

### DIFF
--- a/.changeset/resize-handle-pointer-capture.md
+++ b/.changeset/resize-handle-pointer-capture.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] ResizeHandle: a drag survives the cursor crossing an embedded frame. The handle listened for `pointermove`/`pointerup` on `window` without taking pointer capture, so the browser hit-tested every later event — and the moment the cursor entered an `<iframe>` inside the resizable region the events went to the guest document instead. Measured in Chromium, the host received 0 of 25 pointermoves once the cursor was over the frame, the panel stopped tracking, and the `pointerup` was never heard: the handle stayed armed with `data-resizing` set and the body cursor/`user-select` overrides stuck. The drag now takes pointer capture on the grab zone on `pointerdown`, so the whole gesture is delivered there whatever is underneath, and the move/up/cancel handlers sit on that element rather than on `window` (the same shape as Slider and BottomSheet).
+
+@cixzhang
--- a/apps/storybook/stories/Resizable.stories.tsx
+++ b/apps/storybook/stories/Resizable.stories.tsx
@@ -157,9 +157,7 @@ export const Collapsible: Story = {
                     <Text>
                       <span {...stylex.props(ps.sz)}>{sidebar.size}px</span>
                     </Text>
-                    <Text>
-                      Double-click handle or press Enter to collapse.
-                    </Text>
+                    <Text>Double-click handle or press Enter to collapse.</Text>
                   </Stack>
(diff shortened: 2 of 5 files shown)


cixzhang:
/accept-visual 33335001095/1 Expected ResizeHandle pointer-capture coverage.

PR #5298, opened 2026-08-21
fix(PowerSearch): make menuWidth size the field-search menu

cixzhang:
## Why

`PowerSearch` declares a public `menuWidth?: number` prop that nothing in
`packages/core/src/PowerSearch/` ever reads. The field-search menu has always
matched the input and grown with its content, so a consumer who wants a wider
menu — long field labels, descriptions that wrap at the input's width — has no
way to ask for one, and the prop that looks like the way to ask does nothing.

## What

`menuWidth` now reaches the menu. `Typeahead`, `Tokenizer` and `BaseTypeahead`
all take the same prop, so the whole typeahead family sizes its menu one way
rather than only the component the bug was reported against.

Two constraints shaped the implementation:

- **The menu never renders narrower than the input.** `BaseTypeahead`'s popover
  already carries `min-width: anchor-size(width)`. The new width pairs with
  that floor instead of replacing it, so CSS resolves the clamp — min-width
  beats width — and no measurement is involved.
- **Only the field-search menu is sized.** The operator and value popovers that
  follow a field choice keep their own sizing and are untouched.

`menuWidth` also widens from `number` to `number | string`. `DropdownMenu`,
`DropdownMenuSubMenu` and `ContextMenu` already spell it that way, and a prop
of the same name that accepts less here would be its own small trap. Existing
numeric values are unaffected.

One deliberate difference from those three: they treat `menuWidth` as a
min-width that *replaces* the anchor floor, so a small value there can make the
menu narrower than its trigger. This keeps the floor, because a search menu
narrower than the field it hangs off reads as a rendering bug.

## Measured in Chromium

New `Core/PowerSearch → Menu Width` story, four 360px-wide fields; menu width
read off the rendered popover.

| `menuWidth` | menu | note |
|---|---|---|
| omitted | 360px | unchanged — matches the input |
| `640` | 640px | |
| `200` | 360px | floored at the input, not 200px |
| `'34rem'` | 544px | CSS length strings resolve |

RTL (`direction:rtl`): the menu stays start-aligned with the field's inline
start and grows toward the inline end — 640px menu, 360px field, same start
edge as the default menu. With `menuWidth={640}` set, the value popover opened
by choosing a field still measures 400px, its own width. Content longer than an
explicit width wraps inside the menu; an unbreakable token clips at the menu
edge, which is what a fixed width means.

## Testing

Unit tests cover the pixel form, the string form, the omitted default, and that
the value popover stays unsized. The clamp itself is a CSS resolution that jsdom
cannot compute, so it is covered by the browser measurements above.

`menuWidth` is documented now — it had no `doc.mjs` entry, so the docs promised
nothing, but the type did.

### DIFF
--- a/.changeset/powersearch-menu-width.md
+++ b/.changeset/powersearch-menu-width.md
@@ -1,0 +1,25 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] PowerSearch `menuWidth` now sizes the field-search menu
+
+`PowerSearch` has accepted a public `menuWidth` prop that nothing read: the
+field-search menu always matched the input and grew with its content, and a
+consumer wanting a wider menu had no way to ask for one. The prop now reaches
+the menu. `Typeahead`, `Tokenizer` and `BaseTypeahead` gain the same prop, so
+the whole typeahead family sizes its menu the same way.
+
+Two details worth knowing:
+
+- The menu never renders narrower than the input. The `anchor-size(width)`
+  floor already in place stays, and a `menuWidth` below the input's width is
+  ignored by CSS rather than clamped in JS.
+- Only the field-search menu is affected. The operator and value popovers that
+  follow a field choice keep tracking the input.
+
+`menuWidth` also widens from `number` to `number | string`, matching
+`DropdownMenu`, `DropdownMenuSubMenu` and `ContextMenu`, so a CSS length like
+`'32rem'` works as well as pixels. Existing numeric values are unaffected.
+
+@cixzhang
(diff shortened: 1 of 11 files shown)


cixzhang:
Superseded by [#5237](https://github.com/facebook/astryx/pull/5237), which landed the same fix: `menuWidth` now reaches the field-search menu on `main`, with a test covering it. This branch has since gone conflicting against that work. Closing rather than rebasing — the remaining delta here (widening `menuWidth` to `number | string`, extending it to `Typeahead`/`Tokenizer`/`BaseTypeahead`, and the anchor-width floor) is a separate enhancement, worth its own PR if wanted.

PR #5301, opened 2026-08-21
fix(cli): don't let a stray lockfile decide the package manager

josephfarina:
## The bug

`detectPackageManager` walks up from a directory checking lockfiles, and
returns on the first one it finds:

```js
if (fs.existsSync(path.join(dir, 'yarn.lock'))) return 'yarn';
if (fs.existsSync(path.join(dir, 'pnpm-lock.yaml'))) return 'pnpm';
```

A directory with more than one lockfile is therefore resolved by array
position, not by evidence. `yarn.lock` is first, so **one `yarn install` inside
a pnpm project changes the CLI's answer to yarn for good** — the stray lockfile
never goes away on its own.

Everything downstream of that answer is then wrong. `getRunPrefix` returns
`yarn` instead of `pnpm exec`, and that string is written into the agent-docs
block in `CLAUDE.md`:

```
CLI: run every command as `yarn astryx <cmd>` (shown below as `astryx ...`).
```

Agents read that line before writing any Astryx code, so the wrong prefix
propagates into their runs.

## How I hit it

Upgrading Meta's `example_meta_app` (the starter every Nest Studio build is
scaffolded from) to 0.4.5. I ran `yarn install` in a pnpm project — my mistake —
and `astryx upgrade --apply` then rewrote the docs block from
`pnpm exec astryx` to `yarn astryx`. The upgrade looked clean; only the diff
showed it.

What makes that a detection bug and not just my bad `install`: the app states
its package manager outright. Its `nest.json` reads

```json
"commands": { "build": "pnpm run build", "install": "pnpm install" }
```

and `pnpm-lock.yaml` is the only lockfile it commits. One stray file outvoted
both, and it would have gone on outvoting them forever.

Worth noting the ambiguity is not always a mistake either. `fbsource/nest` has
**three lockfiles committed at its root**, `yarn.lock`, `pnpm-lock.yaml` and
`package-lock.json`, and the apps beneath it are a genuine mix: `neststudio` is
yarn, `agentcloud` and `example_meta_app` are pnpm. For anything under that
root with no lockfile of its own, the current code answers "yarn"
unconditionally.

Second-order: the docs block is keyed on the Astryx version, so once a wrong
prefix is written it will not be corrected by a later `upgrade` at the same
version. I have left that alone here.

## The change

A single lockfile still wins over everything else in its directory — the
existing `lockfile takes priority over packageManager field` test passes
unchanged, since that behaviour looked deliberate.

Only the ambiguous case changes. With several lockfiles in one directory the
CLI now asks, in order:

1. the declared `packageManager` field — a statement of intent, where a
   lockfile is a by-product;
2. `npm_config_user_agent`, i.e. the package manager actually running the
   command;
3. the fixed order, as before.

The user-agent step is what fixes the real-world case, where nobody has
declared anything: run through pnpm, get pnpm.

## Test plan

`pnpm vitest run packages/cli/foundation/env/package-manager.test.mjs` — 32
pass.

Four new cases. Two of them fail against the current code and pass against
this change:

```
× prefers a stray lockfile no more than the committed one: asks who is running
× lets the packageManager field break a multi-lockfile tie
  Tests  2 failed | 30 passed        <- before
  Tests  32 passed                   <- after
```

The other two pin behaviour that must not drift: the fixed-order fallback when
there is no runner to ask, and ignoring a runner that has no lockfile in the
directory.

Verified end to end on the app that surfaced it: with the stray `yarn.lock`
removed, `astryx upgrade --apply` regenerates the block as
`pnpm exec astryx` again.

### DIFF
--- a/.changeset/fix-package-manager-detection-multiple-lockfiles.md
+++ b/.changeset/fix-package-manager-detection-multiple-lockfiles.md
@@ -15,8 +15,24 @@ written into the agent-docs block, which agents copy verbatim into their own
 runs.
 
 A single lockfile still outranks everything else in its directory, unchanged.
-When several are present the CLI now consults the declared `packageManager`
-field, then the package manager actually running it
-(`npm_config_user_agent`), and only then falls back to the fixed order.
+When several sit in one directory the tie is broken only from evidence the
+project owns: the declared `packageManager` field, then a committed
+package-manager config file (`pnpm-workspace.yaml`, `.yarnrc.yml`,
+`bunfig.toml`). A stray `install` drops a lockfile; it writes none of those.
+
+The runner (`npm_config_user_agent`) deliberately does NOT break that tie. An
+agent handed the wrong `yarn astryx` line runs the CLI through yarn, so the
+runner agrees with the mistake and regenerating agent docs writes the wrong
+line again — the failure reproduces itself. The same holds for an installed
+binary invoked from that shell. Both now have regressions that start from the
+wrong line.
+
+When nothing project-owned decides it, the CLI does not guess.
+`detectPackageManager` returns the neutral `npx`, which is correct under every
+package manager, and the new `explainPackageManager` reports `ambiguous` with
+the tied candidates. `astryx doctor` turns that into a FAIL naming the
+directory and the fix — add a `packageManager` field, or delete the lockfile
+that does not belong. It is the refusal `findConfigPath` already makes for
+coexisting config files.
 
 @josephfarina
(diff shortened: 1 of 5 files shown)


cixzhang (dismissed):
Thanks for chasing this. An agent that already received the wrong `yarn astryx` line remains stuck: following it selects Yarn again, and a direct installed invocation does too. I reproduced both at this head; regenerating agent docs kept the wrong command. Could ambiguity resolve from project-owned evidence or surface an error, with a regression starting from the existing wrong line?

[Reviewed by Robohands]

PR #5303, opened 2026-08-21
fix(i18n): harden date locale handling

nynexman4464:
## Summary

- Default pure date helpers to deterministic English and Gregorian semantics while preserving the low-level `plainDateFormat` API’s explicit `calendar` option for compatibility.
- Read the live provider locale across Calendar and every DateInput surface, including touch-picker rerenders and navigation.
- Keep Gregorian date semantics across Timestamp, PowerSearch, both chart packages, and every Schedule view while preserving localized language, numbering, and field order.
- Require production date-helper calls to pass locale explicitly, with import-aware coverage for named imports, aliases, and namespaces.

## Visual evidence

French and Thai Calendar, DateInput, and Timestamp output from the updated head. Both follow the provider locale; Thai remains in Gregorian year 2026 rather than Buddhist year 2569.

French and Thai Gregorian date formatting

## Verification

- Red/green proof: the new Lab Chart and Schedule cases produced 3 locale failures at the previous head; the compatibility test also fails against a silent calendar override (`2026` instead of `2569 BE`). All 395 focused date tests now pass.
- Core, charts, Lab, and production Storybook builds pass; Storybook type-check passes.
- `pnpm lint:strict` passes with 0 errors (80 pre-existing warn-tier findings).
- The full UI run passed all functional tests. Its unrelated 500-row Table benchmark exceeded the budget once under parallel load (536 ms); the isolated rerun passed at 310 ms against the 500 ms budget.

### DIFF
--- a/packages/core/src/utils/plainDate.ts
+++ b/packages/core/src/utils/plainDate.ts
@@ -307,10 +307,16 @@ export const DATE_FORMAT_SHORT_WITH_YEAR: Intl.DateTimeFormatOptions = {
   year: 'numeric',
 };
 
+/**
+ * Format a `PlainDate` for display.
+ *
+ * The locale defaults to the same `'en'` fallback as `useLocale()`, avoiding
+ * runtime-dependent output when no provider locale is available.
+ */
 export function plainDateFormat(
   pd: PlainDate,
   options: Intl.DateTimeFormatOptions,
-  locale?: Locale,
+  locale: Locale = 'en',
 ): string {
   return new Intl.DateTimeFormat(locale, options).format(plainDateToDate(pd));
 }
--- a/packages/core/src/utils/plainDate.ts
+++ b/packages/core/src/utils/plainDate.ts
@@ -311,14 +311,19 @@ export const DATE_FORMAT_SHORT_WITH_YEAR: Intl.DateTimeFormatOptions = {
  * Format a `PlainDate` for display.
  *
  * The locale defaults to the same `'en'` fallback as `useLocale()`, avoiding
- * runtime-dependent output when no provider locale is available.
+ * runtime-dependent output when no provider locale is available. PlainDate
+ * fields use the Gregorian calendar by default; an explicit `options.calendar`
+ * keeps the existing formatter flexibility and takes precedence.
  */
 export function plainDateFormat(
   pd: PlainDate,
   options: Intl.DateTimeFormatOptions,
   locale: Locale = 'en',
 ): string {
-  return new Intl.DateTimeFormat(locale, options).format(plainDateToDate(pd));
+  return new Intl.DateTimeFormat(locale, {
+    calendar: 'gregory',
+    ...options,
+  }).format(plainDateToDate(pd));
 }
 
 // =============================================================================
@@ -366,7 +371,7 @@ export const SHARED_DATE_FORMAT_OPTIONS: Record<
 export function formatSharedDate(
   pd: PlainDate,
   format: SharedDateFormat,
-  locale?: Locale,
+  locale: Locale = 'en',
 ): string {
   if (format === 'system_date') {
     return plainDateToISO(pd);
--- a/packages/core/src/utils/plainDate.ts
+++ b/packages/core/src/utils/plainDate.ts
@@ -130,6 +130,7 @@ export function getTimeZoneParts(
     minute: '2-digit',
     second: '2-digit',
     hourCycle: 'h23',
+    calendar: 'gregory',
   }).formatToParts(new Date(instant));
 
   const lookup = Object.fromEntries(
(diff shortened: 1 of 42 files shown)


cixzhang (changes requested):
Thanks for chasing this down — six unthreaded call sites were real, and the rerender tests are the right proof.

Ideally I'd like to avoid the Gregorian default. None of our 30 shipped locales move under it, but `Locale` is any tag: set `th-TH` and someone reading one screen gets 22 สิงหาคม 2026 in the field and 2569 in the `<Timestamp>` beside it, because only DateInput goes through `plainDateFormat`. Thai Buddhist months line up with Gregorian, so that year was right before.

| before | after |
|---|---|
|  |  |

Persian is the case that wanted it, and that's the Calendar grid's fact rather than the formatter's:

```ts
plainDateFormat(m, {...DATE_FORMAT_MONTH_YEAR, calendar: 'gregory'}, locale)
```

Everything else lands as it stands, and #5296's hydration paragraph is worth lifting into the changeset.

Does pinning the three grid headers cover it, or is there a case that needs the formatter itself pinned?

[Reviewed by Robohands]

cixzhang:
Inline, `packages/core/src/utils/plainDate.ts:324` — Timestamp doesn't come through here, so a `th-TH` field says 2026 while the Timestamp beside it says 2569.

nynexman4464:
Gregorian is intentional here, not a fallback chosen for one locale.

Astryx's current date contract is Gregorian end to end: ISO/`PlainDate` values, parsing, arithmetic, constraints, and Calendar navigation all use Gregorian fields. `locale` controls language, numbering, and field order; it does not enable another calendar system. Letting raw `Intl` calls infer a calendar from `th-TH` or a `u-ca-*` extension would change display output without changing those underlying semantics, producing partial support and conflicting values between components.

This update therefore keeps `calendar: 'gregory'` across every shipped date formatter, including Timestamp, PowerSearch, chart formatters, and Schedule. For example, `th-TH` still localizes the date text, but the year remains Gregorian 2026.

Alternative-calendar support needs an explicit contract for demand, calendar selection and runtime switching, conversion, parsing, storage/serialization, grid arithmetic, constraints, and accessibility. That design is being tracked separately in https://github.com/facebook/astryx/issues/5498; until it is resolved, Gregorian is the only supported calendar system.

packages/core/src/utils/plainDate.ts
```diff
@@ -307,12 +308,23 @@ export const DATE_FORMAT_SHORT_WITH_YEAR: Intl.DateTimeFormatOptions = {
...
+ * Format a `PlainDate` for display.
+ *
+ * The locale defaults to the same `'en'` fallback as `useLocale()`, avoiding
+ * runtime-dependent output when no provider locale is available. Astryx date
+ * APIs currently model Gregorian dates, so locale controls language and order
+ * but cannot select another calendar system.
+ */
 export function plainDateFormat(
   pd: PlainDate,
   options: Intl.DateTimeFormatOptions,
```
cixzhang, line 324: This still accepts `calendar`, but line 326 ignores it. Could we omit it from the options type?

cixzhang (changes requested):
Thanks — this fixes the DateInput/Timestamp split I called out: the new head makes every date formatter Gregorian, and the focused browser pair now shows both Thai surfaces on 2026.

One API edge still needs fixing before merge. `plainDateFormat` accepts `calendar` inside `Intl.DateTimeFormatOptions`, then silently overwrites it; the new test proves `calendar: 'buddhist'` compiles but returns Gregorian. Could we remove `calendar` from the accepted options type so the Gregorian-only contract fails at compile time?

[Full review](https://github.com/cixzhang/astryx/wiki/Review-5303)

[Reviewed by Robohands]

nynexman4464:
Thanks — I agree that silently overriding `calendar` is the wrong API behavior, but narrowing the accepted type would be source-breaking here. `plainDateFormat` is publicly exported from `@astryxdesign/core/utils`, and its second parameter has always been the full `Intl.DateTimeFormatOptions`, including `calendar`.

I changed the contract instead:

- omitted `options.calendar` defaults to `gregory`;
- an explicitly supplied `options.calendar` is preserved for compatibility;
- the exception is display-only: `PlainDate` arithmetic, parsing, serialization, Calendar/DateInput navigation, Timestamp, both chart formatter packages, PowerSearch, and Schedule remain Gregorian.

I audited the public surface and `plainDateFormat` is the only Astryx built-in date API that exposes raw calendar selection. Consumer-provided custom format callbacks are still consumer-owned escape hatches, not Astryx calendar support.

The new regression test fails against the silent override (`2026` instead of `2569 BE`) and passes with the compatibility behavior. I also documented this limited exception and its boundaries in #5498.

packages/core/src/utils/plainDate.ts
nynexman4464: Good catch. I did not narrow the type because `plainDateFormat` is public and has always accepted the full `Intl.DateTimeFormatOptions`; removing `calendar` would break TypeScript consumers.

The updated line now uses `options.calendar ?? 'gregory'`: omitted calendar stays Gregorian, while an explicit calendar is honored as the existing display-only escape hatch. The regression test proves the explicit Buddhist request returns `2569 BE`, and #5498 now documents that this exception does not change Gregorian parsing, arithmetic, serialization, or component behavior.

cixzhang (approved):
Thanks — honoring an explicit calendar closes the public API edge, while every Astryx date surface stays Gregorian. This is good now.

[Reviewed by Robohands]

PR #5306, opened 2026-08-21
fix(Calendar): open on a month inside the min/max window

imdreamrunner:
## What this does

Calendar picks its initial visible month from `focusDate ?? value ?? today`, with no regard for `min`/`max`. When the allowed window doesn't contain today — a closed audit period, a booking window that opens next season, a birth-date range — it opens on today's month with every day disabled, and the prev/next arrows are the only way in, one click per month.

The fallback is now today **clamped into the window**: today when today is inside it, otherwise whichever bound is nearest. An explicit `focusDate` or a selected value still wins untouched.

```jsx
// today is 2026-08-21
<Calendar min="2019-01-01" max="2019-04-30" />   // before: August 2026, all disabled
                                                 // after:  April 2019
<Calendar min="2027-03-04" max="2027-06-30" />   // before: August 2026, all disabled
                                                 // after:  March 2027
```

DateInput, DateRangeInput, and DateTimeInput all mount Calendar without a `focusDate`, so they inherit the fix.

## What changed

- New `getInitialFocusDate()` helper (`packages/core/src/Calendar/getInitialFocusDate.ts`) holding the precedence and the clamp, with unit tests; `Calendar.tsx` calls it in place of its inline initializer.
- With `numberOfMonths={2}`, clamping back to `max` lands `max` in the **right-hand** pane — opening on `[max, max + 1]` would spend half the calendar on a month that is entirely out of bounds. The shift never crosses `min`'s month.
- Inverted bounds (`min` after `max`) resolve to `min`, deterministically rather than off a contradiction.
- Docs: `focusDate`'s description now states the fallback (en, zh, dense).
- New Storybook story `Core/Calendar → Min/max window away from today` — the existing min/max stories all pin `focusDate`, so none of them showed this.

## Test plan

- `packages/core/src/Calendar/getInitialFocusDate.test.ts` — 13 unit tests over the precedence, both clamp directions, inclusivity of the bounds, the two-month shift and its `min` guard, and inverted bounds.
- 6 render tests in `Calendar.test.tsx` (fake-timed to 2026-08-21) asserting the visible-month label and that the landed month's days are selectable. Negative control: 3 of them fail with the clamp neutralized, and the `focusDate`/`value` precedence ones correctly still pass.
- Green locally: `pnpm lint:strict` (0 errors), `pnpm vitest run packages/core packages/lab` (7521 tests), `pnpm build`, `pnpm -F @astryxdesign/storybook build`, and the `core`/`lab`/`charts`/`cli`/`storybook` typechecks plus `lab:readiness:check`.

Related: #4941 asks for a month/year picker so distant dates are *navigable*. This is the narrower defect underneath it — the calendar not starting inside its own window — and the two don't overlap.

@imdreamrunner

### DIFF
--- a/packages/core/src/Calendar/Calendar.doc.mjs
+++ b/packages/core/src/Calendar/Calendar.doc.mjs
@@ -83,7 +83,8 @@ export const docs = {
     {
       name: 'focusDate',
       type: 'ISODateString',
-      description: 'Controlled visible month.',
+      description:
+        'Controlled visible month. Unset, the calendar opens on the selected date, else on today clamped into the min/max window.',
     },
     {
       name: 'onFocusDateChange',
@@ -156,7 +157,7 @@ export const docsZh = {
     {name: 'dateConstraints', type: 'Array<(date: Date) => boolean>', description: '自定义约束函数。'},
     {name: 'maxRangeSpan', type: 'number', description: '范围模式：范围最多可跨越的天数，含首尾两端（7 = 7 天窗口）。选定起始日后限制窗口大小。'},
     {name: 'minRangeSpan', type: 'number', description: '范围模式：范围最少需跨越的天数，含首尾两端（2 表示禁止单日范围）。默认为 1。'},
-    {name: 'focusDate', type: 'ISODateString', description: '受控可见月份。'},
+    {name: 'focusDate', type: 'ISODateString', description: '受控可见月份。未设置时，日历打开时显示已选日期所在月份；若无选中值，则显示今天，并将其限制在 min/max 范围内。'},
     {name: 'onFocusDateChange', type: '(focusDate: ISODateString) => void', description: '导航回调函数。'},
     {name: 'handleRef', type: 'React.Ref<CalendarHandle>', description: '日历导航的命令式句柄，包括 navigateTo()。'},
     {name: 'hasOutsideDays', type: 'boolean', description: '显示相邻月份的日期。', default: 'true'},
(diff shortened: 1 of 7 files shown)


packages/core/src/Calendar/Calendar.tsx
```diff
@@ -277,20 +281,19 @@ export function Calendar({ref, ...props}: CalendarProps) {
...
-      } else {
-        return plainDateFromISO(effectiveValue.start);
-      }
-    }
-    return plainDateToday();
-  });
+  // Focus date state (which month is visible). Falls back to today, clamped
+  // into the min/max window so a window that doesn't contain today doesn't
+  // open on an all-disabled month.
+  const [internalFocusDate, setInternalFocusDate] = useState<PlainDate>(() =>
```
cixzhang, line 287: Mount-only: a check-out `min` chosen later still opens on today, all disabled.

packages/core/src/Calendar/Calendar.doc.mjs
```diff
@@ -83,7 +83,8 @@ export const docs = {
     {
       name: 'focusDate',
       type: 'ISODateString',
-      description: 'Controlled visible month.',
+      description:
+        'Controlled visible month. Unset, the calendar opens on the selected date, else on today clamped into the min/max window.',
```
cixzhang, line 87: Worth saying "on mount" — the clamp doesn't re-run when bounds change.

cixzhang (approved):
Thanks, this is good. Clamp and the two-month shift both check out. Nit: bounds that arrive after mount — a check-out `min` — still open on today's disabled month.

[Reviewed by Robohands]

PR #5309, opened 2026-08-21
fix(docsite): improve changelog package navigation

### LINKED ISSUE #5307: Docsite changelog should default to core and show an empty state
## Problem

The changelog page selects the first package alphabetically, so it opens on `@astryxdesign/cli` instead of the primary `@astryxdesign/core` package.

Empty release sections render as bare version headings, which makes the changelog appear incomplete. Theme packages also crowd the primary package selector.

## Expected behavior

- Select `@astryxdesign/core` initially when the URL has no package query.
- Support direct links to visible packages through `?package=...`, updating the URL only when the user changes tabs.
- Keep theme packages out of the primary changelog tabs.
- Show a clear "No changes in this release." message beneath empty version headings.

nynexman4464:
## Summary

Improve changelog navigation and empty states:

- Put `@astryxdesign/core` first in the visible tab order and select it by default when the URL has no `package` query.
- Select a requested visible package from `?package=...` and update that query only when a tab changes.
- Keep theme packages out of the primary tab row.
- Show `No changes in this release.` beneath empty version headings instead of leaving blank sections.

Fixes #5307.

### Before

CLI appears first and is selected by default; theme packages also occupy the primary tab row.

Before: CLI first in the changelog tabs

### After

Core appears first and is selected by default, followed by CLI; theme package tabs are omitted.

After: Core first and selected in the changelog tabs

[Open the deployed preview](https://astryx-git-fix-changelog-core-default-fbopensource.vercel.app/changelog)

## Test plan

- Confirmed the Core-first regression test fails against the prior implementation and passes with this change.
- `pnpm -F @astryxdesign/docsite test src/__tests__/changelog-view.test.ts src/__tests__/changelog-linkify.test.ts` (25 tests)
- `pnpm -F @astryxdesign/docsite typecheck`
- Pre-commit repository checks and lint passed.

### DIFF
--- a/apps/docsite/src/__tests__/changelog-linkify.test.ts
+++ b/apps/docsite/src/__tests__/changelog-linkify.test.ts
@@ -11,6 +11,7 @@ import {
   linkifyPRs,
   linkifyContributors,
   linkifyComponents,
+  addEmptyReleasePlaceholders,
   stripTitle,
 } from '../components/changelogLinkify';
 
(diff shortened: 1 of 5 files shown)


rubyycheung:
This substantially overlaps #5362: both default the changelog to `@astryxdesign/core`, add `?package=` navigation, and fill empty release sections. The main behavioral difference is that this PR removes theme packages from the primary tabs, while #5362 deliberately leaves their destination unresolved.

Could we choose or consolidate a canonical implementation before either lands? One design point applies to both versions: if Core is the primary/default changelog, it should also be first in the visible tab order. Selecting a lower tab by default while CLI remains first communicates conflicting hierarchy.

nynexman4464:
Updated this PR as the canonical implementation: Core is now first in the visible tab order, CLI is second, and theme package changelogs remain omitted. I also added the final-empty-release and populated-release regression cases from #5362, which is now closed as superseded.

nynexman4464:
Rebased onto current `main` after noticing the earlier preview was built from before the 0.5.0 release. Verified the refreshed preview now renders 0.5.0 for both Core and CLI, and replaced the after screenshot with the current release data.

rubyycheung (approved):
Looks great from a design perspective. Core is now correctly presented first and selected by default, CLI follows as the secondary package, and removing theme packages from the primary tab row keeps the changelog hierarchy focused and clear.

PR #5310, opened 2026-08-21
feat(Table): let useTableSelection opt out of the checked-row highlight

ernestt:
## Problem

The selection plugin paints checked rows by writing `backgroundColor` straight onto each `<tr>`, imperatively, from the per-row ref callback:

```ts
el.style.backgroundColor = selectedBgColor; // colorVars['--color-accent-muted']
```

An inline style outranks anything StyleX can layer on, and `UseTableSelectionConfig` exposed no flag to turn it off. So the accent wash was unconditional and unreachable from userland — the only way out was to fork the plugin.

## Motivating use case

A table template uses row background to mean "this row is open in the detail panel." The selection wash on checked rows collides with that meaning, and there was no supported way to opt out.

## API added

One optional field on `UseTableSelectionConfig`:

```ts
/** Paint checked rows with the accent wash. ... @default true */
hasRowHighlight?: boolean;
```

```tsx
useTableSelection({...config, hasRowHighlight: false});
```

- **Default is unchanged.** It defaults to `true` and is read as `config.hasRowHighlight ?? true`, so every existing table paints exactly as before. Additive and non-breaking.
- **`aria-selected` is intentionally preserved.** The flag drops *only* the background. Checked rows still get `aria-selected="true"` and unchecked rows still have it removed, in both modes — that is the semantic half of the state and screen readers depend on it. This is deliberate, not an oversight.
- **Runtime flips settle.** The background is now written on every pass instead of only when painting, so flipping the flag while a row is already checked clears the inline value rather than stranding it. The existing `useEffect(() => store.notify())` already re-runs each row's ref subscription on every render, so no new plumbing was needed.

The name follows the `has*` convention already used for presentational booleans in this area — `hasHover` and `isStriped` on `Table`, `hasExpandAllControl` and `hasRowClickExpansion` on the tree plugin.

`useTableSelectionState` is not touched: it builds only the *state* half of the config, and the other presentational option (`getRowLabel`) is not threaded through it either. Consumers spread it — `useTableSelection({...selectionConfig, hasRowHighlight: false})`.

## Docs

- `useTableSelection.doc.mjs`: prop added to `docs.props`, `docsZh.props`, and `docsDense.propDescriptions`, plus an example showing the opt-out.
- File header updated, and its `SYNC:` list now names `useTableSelection.doc.mjs` — the config prop table lives there and the list had never mentioned it.
- Changeset: `[feat]` / `patch`, per the 0.x coupling `check:changesets` enforces (only `[breaking]` bumps the minor while 0.x).

## Test plan

Five cases added to `useTableSelection.test.tsx`:

- [x] Checked row is painted with `--color-accent-muted` by default, and carries `aria-selected`
- [x] Checked row is **not** painted with `hasRowHighlight: false`, and still carries `aria-selected`
- [x] Flipping the flag to `false` while a row is checked clears the stranded background
- [x] Flipping it back to `true` repaints the still-checked row
- [x] Unchecked rows are never painted and never carry `aria-selected`, in either mode

Verified the default-case assertion is non-vacuous: the token resolves to a non-empty `var(...)` string that jsdom retains verbatim on `style.backgroundColor`, so the two modes assert genuinely different values.

Gates:

- [x] `pnpm -F @astryxdesign/core typecheck` and `typecheck:docs`
- [x] `eslint` on the changed files at `ASTRYX_STRICT_LINT=1` — clean
- [x] `prettier --check` on all four changed files — clean
- [x] `vitest run packages/core/src/Table/` — 481 passed across 22 files (selection alone: 37)
- [x] `pnpm check:repo` — clean (also re-run by the pre-commit hook)

### DIFF
--- a/packages/core/src/Table/plugins/selection/useTableSelection.tsx
+++ b/packages/core/src/Table/plugins/selection/useTableSelection.tsx
@@ -26,7 +26,13 @@
  * DOM elements and no central element tracking. Each subscription
  * self-cleans when the row disconnects.
  *
+ * Because the background is an inline style, it outranks anything
+ * userland can layer on with StyleX. `hasRowHighlight: false` is the
+ * supported way to reclaim the row background; `aria-selected` is
+ * unaffected by it.
+ *
  * SYNC: When modified, update these files to stay in sync:
+ * - /packages/core/src/Table/useTableSelection.doc.mjs (config prop table)
  * - /packages/core/src/Table/Table.doc.mjs (selection documentation)
  * - /packages/core/src/Table/index.ts (exports)
  */
@@ -81,6 +87,22 @@ export interface UseTableSelectionConfig<T extends Record<string, unknown>> {
    * ```
    */
   getRowLabel?: (item: T) => string;
+  /**
+   * Paint checked rows with the accent wash. Set false when the surrounding
+   * UI already uses row background to mean something else — a row that is
+   * open in a detail panel, for instance. The wash is an inline style, so
+   * it cannot be overridden from userland; this flag is the way off it.
+   *
+   * Only the background is dropped. `aria-selected` is still set on checked
+   * rows either way, so the selection stays legible to screen readers.
+   *
+   * @default true
+   * @example
+   * ```
+   * useTableSelection({...config, hasRowHighlight: false})
+   * ```
+   */
+  hasRowHighlight?: boolean;
 }
 
 // =============================================================================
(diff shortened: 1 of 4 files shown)


packages/core/src/Table/plugins/selection/useTableSelection.tsx
```diff
@@ -121,18 +143,24 @@ function createSelectionStore<T extends Record<string, unknown>>(
...
   if (isSelected) {
     el.setAttribute('aria-selected', 'true');
-    el.style.backgroundColor = selectedBgColor;
   } else {
     el.removeAttribute('aria-selected');
-    el.style.backgroundColor = '';
   }
+  // Written on every pass, not just when painting: the flag can flip while a
+  // row is already selected, and the wash has to come back off.
+  el.style.backgroundColor =
```
cixzhang, line 162: Runs for every row on every pass, so with the flag off it also clears a background a consumer set through `htmlProps.style`. Skip the write when `hasRowHighlight` is false.

cixzhang (changes requested):
Thanks for chasing this back to the inline write rather than working around it — the diagnosis is right and the prop is a reasonable answer. Two things before I'd take it.

The opt-out doesn't actually hand the row background back. `useTableSelection.tsx:162` writes `el.style.backgroundColor = ''` on every pass for every row it holds, so a table painting its detail-panel row with `htmlProps.style` still gets wiped with the flag off — only `xstyle` survives, no warning. Skip the write entirely when `hasRowHighlight` is false.

Second, selection isn't reachable from a theme at all: `themeProps('table-row')` publishes no state, so there's no `data-selected` to key off. A theme author wanting a different selection colour gets an off switch and nothing else. Worth exposing even if the paint stays a ref write.

Is the ref write there for a reason I should know about — render count on large tables?

[Reviewed by Robohands]

cixzhang (approved):
Withdrawing my first point — I had it wrong. `main` already writes `el.style.backgroundColor = ''` on the unselected branch, so the clobber isn't yours and isn't new. And skipping the write when `hasRowHighlight` is false would strand the wash on a row that's already painted when the flag flips, which is exactly what your comment on that line says you're avoiding. Unconditional is the right shape.

That leaves only the theme point, and it isn't blocking: selection isn't reachable from a theme at all — `themeProps('table-row')` publishes no state to key off. Worth a follow-up, not a change here.

Clearing my request. The prop is fine and so is the implementation.

[Reviewed by Robohands]

PR #5313, opened 2026-08-22
feat(Table): move the filters into a bottom sheet on a phone

imdreamrunner:
## Why

The mobile prototype for the migration table specs the filtering plugin's
mobile form as a bottom sheet:
https://facebook.github.io/astryx/sandbox/pages/mobile-prototypes/?p=tablefilter

A header cell is the wrong home for a filter on a phone. The column is too
narrow to hold a control, so the `inline` variants squash to nothing, and the
`popover` variant anchors a 240px panel to a funnel icon inside a `<th>` —
covering the very rows the filter is meant to narrow. Every table using the
plugin has this today; nobody has to opt in to get it fixed.

## What

Below 640px the plugin takes every control out of the header and puts it behind
one **Filter** button above the table, carrying the count of active filters,
which opens a bottom sheet holding all of them at full width with their column
names as labels.

- **Applies live, no scrim.** Each change is applied as it is made and the
  sheet is non-modal, so the rows behind stay visible and keep updating — the
  point of filtering on a screen this small. `Done` dismisses; `Reset` clears
  every filter and is disabled when none are set.
- **Short enum filters become lists.** An `enum` renders as a `RadioList` (with
  a leading "All" that clears it) and an `enum_list` as a `CheckboxList`, up to
  the 7 options `CheckboxList`/`RadioList` themselves give as the point to
  switch to a `Selector`. Above that the `Selector` stays, for its search. This
  keeps a second layer from opening on top of the sheet, which also matters
  because `Selector` and `MultiSelector` are themselves specced to become
  sheets on mobile. Every other filter type keeps the control it already has.
- Three new config fields, all optional: `variant: 'sheet'` takes the
  presentation at any width, `sheetBreakpoint` (`'sm'` default, `'none'` to
  keep the header controls everywhere) moves the threshold on AppShell's scale,
  and `defaultIsMobile` seeds it for SSR.

Nothing above the breakpoint changes.

## Risk

The mobile presentation of an existing plugin changes without an opt-in — that
is the intent, but it is the thing to disagree with. `sheetBreakpoint: 'none'`
is the escape hatch. Everything else is additive: no prop, variant, or export
was removed or renamed, and the desktop DOM is untouched.

The trigger sets its own 44px floor under `(pointer: coarse)`: no `size` token
reaches 44px on every theme (base is 36px at `lg`), and this is the one filter
affordance on a touch screen. Same treatment the header funnel trigger already
gets in this file.

## Testing

`lint:strict`, `test`, `build`, and the storybook typecheck stack all pass.

12 new tests in `useTableFiltering.test.tsx` drive both sides of the breakpoint
through a query-aware `matchMedia` stub: the header controls disappear and the
Filter button appears below it, the header keeps its controls above it and
under `sheetBreakpoint: 'none'`, the sheet carries one labelled control per
filterable column, a toggle applies with no Apply step, the count reaches the
trigger's accessible name, Reset clears everything, Done keeps the filters, a
12-value enum keeps its Selector, and a table with no filterable column grows
no button.

Verified in Chromium on an iPhone 15 profile against built `core`, which is
where the two fixes above came from — measured 44px trigger, focus returning to
the trigger on close (a scrim-less sheet never captured it, so closing dropped
focus to the document), Escape closing the sheet, the page behind still
scrollable, the count badge updating while the sheet is open, RTL flipping the
trigger to the left edge, and dark mode. Storybook: **Core/TableFiltering →
Sheet variant (touch)**.

### DIFF
--- a/.changeset/table-filter-sheet.md
+++ b/.changeset/table-filter-sheet.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] `useTableFiltering`: on a narrow viewport the filter controls leave the table header for a bottom sheet, opened from one Filter button that carries the active-filter count. A header cell is the wrong home for a filter on a phone — the column is too narrow to hold a control, and a popover anchored to it covers the rows the filter is meant to narrow. The sheet applies each change as it is made and has no scrim, so the rows behind stay visible and keep updating; short enum filters render as a list of checkboxes or radios rather than a Selector, so nothing opens a second layer on top of the sheet. `sheetBreakpoint` moves or disables the threshold (`'sm'` by default, `'none'` to keep the header controls at every width), `variant: 'sheet'` takes the presentation at any width, and `defaultIsMobile` seeds it for SSR.
+
+@imdreamrunner
--- a/apps/storybook/stories/TableFiltering.stories.tsx
+++ b/apps/storybook/stories/TableFiltering.stories.tsx
@@ -555,3 +555,47 @@ export const EmptyState: Story = {
     );
   },
 };
+
+export const SheetVariant: Story = {
+  name: 'Sheet variant (touch)',
+  parameters: {
+    viewport: {defaultViewport: 'mobile1'},
+  },
+  render: () => {
+    const {config, applyFilters} = usePowerSearchConfig(fieldDefs);
+    const {filters, onFilterChange} = useTableFilterState();
+    const columns: TableColumn<Employee>[] = [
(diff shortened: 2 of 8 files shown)


humbertovirtudes (changes requested):
The Done action bypasses the only close path that restores focus to the Filter trigger. Please centralize dismissal, route Done through the focus-restoring path, and add a focused Done-to-trigger restoration test. This draft also needs rebasing after #5420 settles because it is conflicted and more than 700 commits behind.

PR #5314, opened 2026-08-22
feat(PowerSearchMobile): the touch form of PowerSearch

imdreamrunner:
## What

`PowerSearchMobile` — the touch form of `PowerSearch`, in core.

Following the mobile exploration: [Mobile Prototypes → PowerSearch](https://facebook.github.io/astryx/sandbox/pages/mobile-prototypes/?p=powersearch) ("bottom sheet, pinned tall").

## Why

PowerSearch's desktop shape doesn't survive a phone. The typeahead drops a popover that fights the on-screen keyboard, and the edit popover lays field / operator / value out in a row that has nowhere to go at 390px.

## How

Same props, same `PowerSearchFilter` model, same tokens. The typeahead dropdown and the row-shaped popover are replaced by a pinned-tall bottom sheet that drills down **field → operator → value**. Pick between the two variants on viewport and the call site does not change:

```tsx
const isTouch = useMediaQuery('(max-width: 768px)');
const Search = isTouch ? PowerSearchMobile : PowerSearch;
<Search config={config} filters={filters} onChange={setFilters} />
```

| | |
|---|---|
| **Tap target** | The same input shell, showing the filters as tokens. Tapping a token reopens its editor, where **Delete** removes it; the token's own × removes it in place. |
| **Field list** | Grouped by `field.group`, each row captioned with the operator it will produce. Past seven fields it gains a search box pinned under the sheet title. |
| **Operator** | A drill-down row, shown only when the field defines more than one. With one, the sheet header carries it — so the happy path stays **two taps**, as in the prototype. |
| **Value** | `enum` commits on a single tap (Selector's mobile pattern); `enum_list` applies from a pinned footer (MultiSelector's); an `empty` operator lands with no value step at all; every other type falls through to the shared `PowerSearchValueEditor`. |

Both sheets are `height="tall"`: the field list resizes as it is searched and the editor's content changes with the operator, so a self-sizing sheet would jump on every keystroke and every step — and tall is the only height that gives mobile-keyboard accommodation, which the text and number editors need. The editor sheet is `purpose="form"`, so a stray scrim tap can't discard a half-built filter.

`nested` filter groups have no touch editor yet. Fields whose only operators are nested are left out of the list and a dev warning says so, rather than opening a blank sheet.

## Accessibility

- The field label names the **group**, not the tap target — a `<label>` pointed at the button would replace its visible text as the accessible name (WCAG 2.5.3). The button is named by its own text, and carries `aria-haspopup="dialog"` + `aria-expanded`.
- `disabledMessage` keeps the target focusable via `aria-disabled` and describes it, so the reason is reachable by keyboard; the action stays blocked.
- Result-count changes are announced through the same polite live region as the desktop variant.

## Screenshots

Light and dark, iPhone 15: **[assets/pr-5314](https://github.com/facebook/astryx/tree/assets/pr-5314)**

| Tap target | Field list | Value (enum) | Operators |
|---|---|---|---|
|  |  |  |  |

| Value (multi) | Edit + delete | Value (text) | Field search |
|---|---|---|---|
|  |  |  |  |

<details><summary>Dark</summary>

| Tap target | Field list | Value (enum) | Operators |
|---|---|---|---|
|  |  |  |  |

| Value (multi) | Edit + delete | Value (text) | Field search |
|---|---|---|---|
|  |  |  |  |

</details>

## Test plan

- 23 new tests in `PowerSearchMobile.test.tsx`, covering the whole flow: field → operator → value, the enum single-tap commit, the multi-select apply, the empty-operator shortcut, edit/delete, clear-all keeping read-only filters, read-only and disabled, the field search and its empty state, the nested-field warning, and the imperative handle.
- Driven in a real browser at iPhone 15 (light + dark) against built `dist`: the flow works end to end with no console errors, the tablet sheet caps at 640px, and the read-only / disabled / error states behave.
- Gates: `pnpm lint:strict`, `pnpm build`, `pnpm test` (7527 core+lab, 2804 cli+apps), `typecheck` + `typecheck:docs` across core/lab/charts/cli/storybook, `pnpm -F @astryxdesign/storybook build`, `pnpm lab:readiness:check` — all green.

## Notes for reviewers

- Four props describe desktop-only affordances and are inert here — `hasAutoFocus`, `menuWidth`, `maxOperatorMenuItems`, `tokenOverflowBehavior` — documented on the prop table rather than removed, so one call site can feed both variants.
- `handleRef`'s `focusTypeahead()` / `blurTypeahead()` move focus to and from the tap target; there is no typeahead input on this variant.
- The seven-field threshold for the search box is the number `CheckboxList` already uses to send a long list to `MultiSelector`.

### DIFF
--- a/packages/core/src/PowerSearch/PowerSearchMobile.doc.mjs
+++ b/packages/core/src/PowerSearch/PowerSearchMobile.doc.mjs
@@ -1,0 +1,481 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+/** @type {import('@astryxdesign/cli/authoring').ComponentDoc} */
+
+export const docs = {
+  name: 'PowerSearchMobile',
+  displayName: 'Power Search Mobile',
+  category: 'Data Input',
+  keywords: [
+    'powersearch',
+    'mobile',
+    'touch',
+    'search',
+    'filter',
+    'filterbar',
+    'bottomsheet',
+    'faceted',
+    'querybuilder',
+    'responsive',
+  ],
+  props: [
+    {
+      name: 'config',
+      type: 'PowerSearchConfig',
+      description:
+        'Configuration defining available fields, operators, and their value types. Identical to PowerSearch.',
+      required: true,
+    },
+    {
+      name: 'filters',
+      type: 'ReadonlyArray<PowerSearchFilter>',
+      description: 'Currently active filters.',
+      required: true,
+    },
+    {
+      name: 'onChange',
+      type: "(filters: ReadonlyArray<PowerSearchFilter>, changeType: 'add' | 'edit' | 'remove', index: number) => void",
+      description:
+        "Called when filters change. changeType is 'add', 'edit', or 'remove'. index is the affected filter's position.",
+      required: true,
+    },
+    {
+      name: 'label',
+      type: 'string',
+      description:
+        'Accessible label for the tap target that opens the filter sheet.',
+      default: "'Search'",
+    },
+    {
+      name: 'isLabelHidden',
+      type: 'boolean',
+      description: 'Visually hides the label while keeping it accessible.',
+      default: 'true',
+    },
+    {
+      name: 'placeholder',
+      type: 'string',
+      description:
+        'Text shown on the tap target while no filters are selected. Once a filter exists the target reads "Add filter" instead.',
+      default: "'Search...'",
+    },
+    {
+      name: 'hasClear',
+      type: 'boolean',
+      description:
+        'Show a clear-all button on the tap target. Read-only filters are left in place.',
+      default: 'true',
+    },
+    {
+      name: 'isReadOnly',
+      type: 'boolean',
+      description:
+        'Prevent adding, editing, or removing filters. The sheet no longer opens and tokens lose their remove buttons.',
+      default: 'false',
+    },
+    {
+      name: 'isDisabled',
+      type: 'boolean',
+      description: 'Disables the entire component.',
+      default: 'false',
+    },
+    {
+      name: 'disabledMessage',
+      type: 'string',
+      description:
+        'Explains why the search is disabled. With isDisabled, shows a tooltip on hover/keyboard focus and keeps the tap target focusable via aria-disabled (opening stays blocked). Use this instead of wrapping a disabled PowerSearchMobile in Tooltip. Disabled controls swallow the hover events an external Tooltip needs.',
+    },
+    {
+      name: 'status',
+      type: "{type: 'warning' | 'error' | 'success', message?: string}",
+      description: 'Validation status object with type and optional message.',
+    },
+    {
+      name: 'statusVariant',
+      type: "'attached' | 'detached'",
+      description:
+        'How the status message is placed relative to the tap target. attached overlaps directly below it (bordered treatment); detached floats below as a separate element with spacing.',
+      default: "'attached'",
+    },
+    {
+      name: 'startIcon',
+      type: 'ReactNode | IconType',
+      description:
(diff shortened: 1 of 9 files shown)


packages/core/src/PowerSearch/PowerSearchMobile.doc.mjs
```diff
@@ -0,0 +1,481 @@
...
+        guidance: false,
+        description:
+          'Render it inside another dialog or bottom sheet. Its sheet is a top-layer dialog of its own and would stack a second modal surface on the first.',
+      },
+    ],
+  },
+  examples: [
+    {
+      label: 'Responsive: one call site, both variants',
+      code: `const isTouch = useMediaQuery('(max-width: 768px)');
```
cixzhang, line 197: Hmm, 768 is an iPhone in landscape. Swaps the component mid-filter

packages/core/src/PowerSearch/PowerSearchMobile.tsx
```diff
@@ -0,0 +1,1098 @@
...
+    if (resultCount == null) {
+      return null;
+    }
+    if (typeof resultCount === 'number') {
+      return t('@astryx.powersearch.resultCount', {count: resultCount});
+    }
+    return resultCount;
+  }, [resultCount, t]);
+
+  const announce = useAnnounce();
```
cixzhang, line 617: Some of this seems copied from PowerSearch. Probably can share it rather than keep two

cixzhang (commented):
Thanks — the sheet is lovely and two taps is the right target.

The hard part is who owns the switch. DateInput settles it inside the component on `(pointer: coarse)`, no width bound; your [#5315](https://github.com/facebook/astryx/pull/5315) does the same for Tokenizer. Here the call site owns it, and the docs teach a width:

```tsx
const isTouch = useMediaQuery('(max-width: 768px)');
const Search = isTouch ? PowerSearchMobile : PowerSearch;
```

That swaps component identity, so a rotation mid-filter unmounts the half-built filter.

Ideally one `PowerSearch` picks its own surface. The argument the other way is bundle — PowerSearch doesn't pull BottomSheet today. Comment only because it's a draft; undrafting wants this settled.

Is the tree-shaking worth the call site owning the breakpoint?

[Reviewed by Robohands]

packages/core/src/PowerSearch/PowerSearchMobile.tsx
```diff
@@ -0,0 +1,1098 @@
...
+ * const Search = isTouch ? PowerSearchMobile : PowerSearch;
+ * <Search config={config} filters={filters} onChange={setFilters} />
+ * ```
+ *
+ * Tapping the field opens a pinned-tall sheet listing the available fields.
+ * Choosing one opens its value editor; a field with more than one operator
+ * shows the operator as a row that drills into its own list. Tapping a token
+ * reopens that filter's editor, where Delete removes it.
+ */
+export function PowerSearchMobile({
```
cixzhang, line 322: These four typecheck but never arrive. Might need a dev warning at least

cixzhang:
One more thing worth settling before this undrafts: `PowerSearchMobile` and `PowerSearchMobileProps` are exported from `PowerSearch/index.ts`, so `export * from './PowerSearch'` puts them in the published core barrel. If the answer to the switch question turns out to be "one PowerSearch picks its own surface", we'd be deprecating an export we just shipped — and the `astryx-power-search-mobile` theme target freezes as soon as a theme selects on it.

DateInput keeps `TouchDateField` internal for that reason. Landing the sheet unexported would let the work in without committing the public shape.

[Reviewed by Robohands]

imdreamrunner:
Sorry i don't like this approach the AI chose and decided to start from scratch.

review feedbacks are recorded for the new session

PR #5315, opened 2026-08-22
feat(core): Tokenizer fits the pointer — a chip row and a sheet on a finger

imdreamrunner:
Builds the [Tokenizer mobile prototype](https://facebook.github.io/astryx/sandbox/pages/mobile-prototypes/?p=tokenizer) into `Tokenizer` itself, as a second surface the component picks from the primary pointer — not a parallel component to adopt.

## Why the pointer control cannot just be made bigger

Both of its core gestures need a hardware keyboard. You type **between** the chips, in an input sharing their line, and you remove the last one with **Backspace on an empty input**. On a phone, focusing that input raises the virtual keyboard over the bottom half of the screen — where the suggestion popover opens — and each chip added grows the field by a line and pushes the page under your thumb.

## The touch surface

| | |
|---|---|
| **Chips scroll sideways** | The field is exactly one line tall however many there are, so adding and removing never reflow the form below it. |
| **Add is a fixed target** | Outside the scroller at the trailing edge: same place with two chips or twenty, and a tap on it can never be mistaken for the start of a sideways drag. *(The prototype has it inside the scroller, where it scrolls away once the chips overflow — the one deliberate departure.)* |
| **Suggestions are a pinned-tall sheet** | Search at the top where the keyboard cannot cover it, full-width rows a thumb can hit. `tall` is the one BottomSheet height that is keyboard-aware. The list populates before anything is typed: in a sheet the list **is** the content. |

Tapping a row adds that token and leaves the sheet up, so building a set of five is five taps.

Nothing changes at the call site — same props, same values, no new import, no media query to write. The switch is `pointer: coarse` alone with no width bound, the same test and the same reasoning as the DateInput touch surface: `pointer` is the PRIMARY device, so a touchscreen laptop keeps the typable field and a narrowed desktop window is still a mouse, while a tablet gets the sheet at any width.

## Notable

- **One selection engine, not two.** `useTokenSelection` holds what a token *is* — including the `Create "X"` sentinel, a synthetic item recognised on the way back in and turned into a real one. That is the part that must not fork. The pointer surface's rendered output is unchanged; its diff is the extraction and a rename to `PointerTokenizerField`.
- **Two bugs only a browser could show**, both now covered by a test:
  - The sheet's sticky header sits between two layers in **one** stacking context. A bare `z-index: 1` put it over BottomSheet's grab handle and hid the pill; no z-index at all let List's `position: relative` rows paint straight through its background. It needs an isolated wrapper to scope the layer.
  - `minBlockSize` compiles to `min-height`, so a 44px thumb floor declared next to the size scale is the same property and simply loses the merge. Folded into the scale as `max(44px, <size>)`.
- **Cost: 16.4 KB gzipped** on top of Tokenizer's 58.6 KB — mostly BottomSheet and List — for every consumer including desktop-only ones, since the choice is made at runtime. (Measured: bundle the component alone, minified, from `dist` before and after.)
- **New public API is one export**, `TokenizerTouchSurface` (the touch half with the pointer test skipped, so the stories are reviewable on a laptop), plus three `@astryx.tokenizer.*` catalog keys. No media-query constant: #5243's DateInput surface already publishes `TOUCH_POINTER_QUERY`, and two of those in the root barrel would collide.
- The plus glyph is drawn in the component rather than registered as an icon name — structural, the way `CheckboxIndicator` draws its own tick; registering a name would put all seven bundled themes on the hook for an icon.

## Test plan

- `pnpm lint:strict` clean (0 errors; the 53 warnings are pre-existing on main), `pnpm build`, `pnpm test` — 7,530 core+lab and 2,801 cli+apps tests pass.
- CI's extra typechecks: `core`/`lab`/`charts typecheck:docs`, `cli typecheck:strict` + `typecheck:template-docs`, `storybook typecheck` + `build`, `lab:readiness:check`.
- 26 new tests in `TokenizerTouch.test.tsx` covering the surface switch (including that width alone does not switch it, and that a tablet does), the closed field's contract, the sheet, `hasCreate`, `maxEntries`, and the disabled-with-a-reason path. The existing `Tokenizer.test.tsx` is untouched and still passes — the shared setup answers `(pointer: coarse)` false, so it keeps testing the pointer surface.
- **Verified in a browser** (Chromium, iPhone 15 profile, built `dist`): the row scrolls (`scrollWidth` 540 vs `clientWidth` 253) while the field stays 52px and does not move; the sheet opens at 92dvh with the pill visible above the header; the header covers rows scrolling under it; tapping the search field through it still focuses it.

Storybook: **Core/Tokenizer/Touch surface** — the first story reports which surface you are actually looking at; the rest force the touch half.


imdreamrunner:
Follow-up commit: `pr-a11y` was red, and the cause was worth writing down.

**The story title broke the tooling.** `Core/Tokenizer/Touch surface` was the only three-segment story title in the repo, and the a11y script reads the two ends of it differently — it *filters* by `title.split("/")[1]` (`Tokenizer`, so the stories were found) but *keys* its report by `title.split("/").pop()`, so every violation was filed under a component called **"Touch surface"** that does not exist. The weekly a11y summary is built from the same report, so it would have grown a phantom component too. Folded the eight stories into `Tokenizer.stories.tsx` behind a section comment; two files can share a title (`Vega/VegaChart` does) but one file per component is the repo shape.

**The remaining violation is the system-wide disabled treatment, not this surface.** Measured, the three nodes on `Touch: disabled, with a reason` are the field label at 2.52 and the two chips at 1.72 — the same nodes and the same ratios as the already-baselined `Tokenizer::Disabled` and `Tokenizer::Disabled With Message`. Both surfaces compose the same two dims: `inputWrapperStyles.disabled` puts `opacity: 0.5` on the field and `Token` adds its own, so a chip is at **0.25 effective on the pointer surface today** — verified against the built Storybook, `core-tokenizer--disabled` reports the identical 0.25. Changing that means changing what disabled looks like system-wide, which is not this PR, and WCAG 1.4.3 exempts inactive controls. So it joins the two entries beside it in the baseline rather than being fixed here.

The disabled Add button sits at that same 0.25, which is why axe does not flag it separately — and why it should not be exempted from the field dim: it would be the one bright thing in a disabled field.

Also worth flagging for #5243: the DateInput touch branch uses the same `Core/DateInput/Touch surface` title and will hit the same keying bug when it runs `pr-a11y`.

Locally: a11y gate passes (0 new, 3 baselined), RTL audit, modal close-visibility guard, storybook typecheck + build, eslint all clean.

PR #5317, opened 2026-08-22
fix(Table): keep group headers out of the caller's cell renderers

ernestt:
## What

`useTableGroupedRows` injects synthetic group-header rows into the flattened data. Each is a Proxy that answers unknown fields with `''`, and the source comment claims this makes "arbitrary field access from user cell renderers" safe.

It doesn't. The Proxy rescues a renderer that *prints* a field. It does nothing for the far more common one that *keys a lookup off* it:

```tsx
renderCell: item => <StatusDot status={STATUS_META[item.status].dot} />
```

`''` is no more a member of that map than `undefined` is, so this throws and blanks the page the moment grouping is switched on. There was no way to opt out from the consumer side, because `BaseTable` evaluates every column's `renderCell` against every row *before* `transformBodyRow` gets to discard a header's cells:

```tsx
const rawContent = isDefaultRenderer
  ? defaultCellRenderer(item, col.key)
  : (col.renderCell?.(item) ?? null);
```

I hit this building a table template and had to work around it with a `transformColumns` wrapper in the consumer — which is what this PR moves into the plugin, where it belongs.

## How

`transformColumns` now wraps each column so `renderCell` is skipped for header rows. Those cells are discarded moments later by `transformBodyRow` regardless, so nothing is lost by never producing them. The Proxy stays as a backstop for the default renderer, which reads `item[key]` directly.

Wrapped columns are cached against their source column in a `WeakMap`. This matters: `BaseTable` stabilises the resolved column array with an element-by-element reference check, so allocating a fresh column object per render would have silently defeated `MemoizedTableRow` and re-rendered every row on every render. There's a test locking that down.

The hook also now returns `isGroupHeader`, so a consumer guarding their own row-level plugin (click-to-open-detail, row links, per-row menus) has something better to test than whether the row key happens to start with `__group_`:

```tsx
transformBodyRow(props, item) {
  if (grouped.isGroupHeader(item)) return props;
  return {...props, htmlProps: {...props.htmlProps, onClick: () => open(item)}};
}
```

## Test plan

Three tests added, each verified to fail without its fix:

- [x] A column whose `renderCell` keys a lookup off a field no longer throws when grouping is on; only the three real rows reach the renderer. Without the guard: `TypeError: Cannot read properties of undefined`.
- [x] `transformColumns` returns the same wrapped column objects across calls, and leaves `renderCell`-less columns untouched. Without the `WeakMap`: fails.
- [x] `isGroupHeader` distinguishes the two synthetic headers from the three real rows.
- [x] Full Table suite: 479 passed (22 files).
- [x] `tsc --noEmit`, `eslint`, and `prettier --check` all clean.

## Notes

One thing I chose not to change: `aria-rowcount` / `aria-rowindex` drift when headers are present, since the rendered rows include them. Fixing it properly needs a count the plugin doesn't own, so it seemed out of scope here.

### DIFF
--- a/packages/core/src/Table/plugins/groupedRows/useTableGroupedRows.test.tsx
+++ b/packages/core/src/Table/plugins/groupedRows/useTableGroupedRows.test.tsx
@@ -175,6 +175,130 @@ describe('useTableGroupedRows', () => {
     ).not.toBeInTheDocument();
   });
 
+  it('does not run a column renderCell against a group header row', () => {
+    // The realistic shape: a renderer that keys a lookup off a field rather
+    // than printing it. The header Proxy answers `''`, which is not a member
+    // of the map, so reaching into the result throws unless the header is
+    // skipped outright.
+    const TEAM_META: Record<string, {dot: string}> = {
+      Core: {dot: 'green'},
+      Infra: {dot: 'blue'},
+    };
+    const seen: Person[] = [];
+    const lookupColumns: TableColumn<Person>[] = [
+      {
+        key: 'team',
+        header: 'Team',
+        renderCell: item => {
+          seen.push(item);
+          return <span>{TEAM_META[item.team].dot}</span>;
+        },
+      },
+    ];
+
+    function LookupHarness() {
+      const [collapsed, setCollapsed] = useState<Set<string>>(new Set());
+      const grouped = useTableGroupedRows<Person>({
+        data: people,
+        groupBy: p => p.team,
+        collapsedGroups: collapsed,
+        onToggleGroup: k => setCollapsed(new Set([k])),
+        getRowKey: p => p.id,
+      });
+      return (
+        <Table
+          data={grouped.data}
+          columns={lookupColumns}
+          idKey={grouped.idKey}
+          plugins={{grouped: grouped.plugin}}
+        />
+      );
+    }
+
+    expect(() => render(<LookupHarness />)).not.toThrow();
+    // Only the three real rows reached the renderer.
+    expect(seen).toHaveLength(3);
+    expect(seen.map(p => p.name)).toEqual(['Alice', 'Bob', 'Carol']);
+    // Group headers still render their own label.
+    expect(screen.getByText('Core')).toBeInTheDocument();
+    expect(screen.getByText('Infra')).toBeInTheDocument();
+  });
+
+  it('returns the same wrapped column objects across calls', () => {
+    // BaseTable compares the resolved column array element by element to
+    // decide whether every row needs re-rendering, so the guard must not
+    // allocate a fresh column on each render.
+    const stableColumns: TableColumn<Person>[] = [
+      {key: 'name', header: 'Name', renderCell: item => <>{item.name}</>},
+      {key: 'team', header: 'Team'},
+    ];
+    let transform:
+      ((c: TableColumn<Person>[]) => TableColumn<Person>[]) | null = null;
+    function StabilityHarness() {
+      const [collapsed, setCollapsed] = useState<Set<string>>(new Set());
+      const grouped = useTableGroupedRows<Person>({
+        data: people,
+        groupBy: p => p.team,
+        collapsedGroups: collapsed,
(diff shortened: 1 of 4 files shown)


packages/core/src/Table/plugins/groupedRows/useTableGroupedRows.tsx
```diff
@@ -295,6 +344,14 @@ export function useTableGroupedRows<T extends Record<string, unknown>>(
 
   const plugin = useMemo(
     (): TablePlugin<T> => ({
+      // BaseTable evaluates every column's `renderCell` against every row
+      // before `transformBodyRow` gets to discard a header's cells. A header
+      // is not one of the caller's rows, so running their renderer against it
+      // can only fail — the cells produced here are thrown away moments later
+      // regardless. Skip the call and hand back the empty cell directly.
+      transformColumns(columns) {
+        return columns.map(headerSafeColumn);
```
cixzhang, line 353: Columns injected after this — rowIndex, rowStatus in table-lab — still hit header rows.

packages/core/src/Table/plugins/groupedRows/useTableGroupedRows.test.tsx
```diff
@@ -175,6 +175,130 @@ describe('useTableGroupedRows', () => {
...
+          data={grouped.data}
+          columns={stableColumns}
+          idKey={grouped.idKey}
+          plugins={{grouped: grouped.plugin}}
+        />
+      );
+    }
+    render(<StabilityHarness />);
+    expect(transform).not.toBeNull();
+    const first = transform!(stableColumns);
```
cixzhang, line 258: Calls transform outside React; a `-perf.test.tsx` would lock the row-memo invariant instead.

cixzhang (approved):
Thanks, good catch — and you're right about the column caching. Two nits inline; the ordering one is worth a follow-up.

[Reviewed by Robohands]

cixzhang:
Ernest — you found this one: the bug, the mechanism, and the exact line where `renderCell` runs ahead of `transformBodyRow`. Your diagnosis is what made the real fix findable.

My approval above was premature. The `transformColumns` wrapper only sees the columns that exist when it runs, so under `plugins={{grouped, rowStatus}}` — the order our table lab uses — the status column is injected after the wrap and the crash returns by another route.

[#5363](https://github.com/facebook/astryx/pull/5363) continues your work one layer down: an `isContentSuppressed` flag that `BaseTable` honors before any renderer runs.

Not closing this PR — I'll decide once [#5363](https://github.com/facebook/astryx/pull/5363) settles, and if it lands, this one closes with credit to you.

[Reviewed by Robohands]

packages/core/src/Table/plugins/groupedRows/useTableGroupedRows.tsx
ernestt: Good catch — the column-injection ordering is a real gap. The `isContentSuppressed` approach in #5363 addresses this at a lower level. Will defer to that.

packages/core/src/Table/plugins/groupedRows/useTableGroupedRows.test.tsx
ernestt: Fair point — a perf test that asserts memo stability would be a better contract than calling the transform in isolation. If this PR lands (vs. being closed in favor of #5363), I'll add one.

ernestt:
Makes sense — `isContentSuppressed` at the BaseTable level is a cleaner fix that handles the ordering-dependent case this PR misses. Happy to close this in favor of #5363 once it lands. Thanks for the credit.

cixzhang:
Thanks @ernestt — your diagnosis is what made this findable: you named the mechanism and pointed at the line. [#5363](https://github.com/facebook/astryx/pull/5363) takes it one layer down, to a per-cell hook that runs against the final column list, so plugin order can't evade it. You're credited in the changeset. A fresh PR from you is always welcome.

[Reviewed by Robohands]

PR #5318, opened 2026-08-22
feat(core): let Selector, MultiSelector, and ComplexSelector drop the trigger chevron

ernestt:
## Problem

The chevron renders as a **sibling** of the trigger button, after the optional `InputClearButton`. So with `hasClear` and a value selected, the trigger shows both a `×` and a chevron in the same end slot, and consumers who wanted the clear affordance to *replace* the chevron had no way to say so:

- The chevron's theme target (`astryx-selector-indicator-icon`) only carries `data-state="expanded|collapsed"`.
- The selector root publishes `data-variant`, `data-size`, `data-status`, `data-disabled` — nothing for "has a value".
- StyleX has no descendant selectors and `stylex.when.*` only reads *upward*, so an `xstyle` on the field cannot reach the icon.

No sanctioned escape hatch existed.

## API added

One optional boolean on all three components:

```ts
/** Whether to show the chevron at the end of the trigger. ... @default true */
hasChevron?: boolean;
```

```tsx
<Selector hasClear hasChevron={false} value={value} onChange={setValue} />
```

- **Default unchanged.** `hasChevron = true` in the destructure (the component idiom here — `DropdownMenu` uses `hasChevron = true`, `Selector` uses `indicatorPosition = 'end'`; the `?? true` form in #5310 was for a hook *config object*, which has no destructure to default). A consumer that omits it renders identically.
- **Gates only the chevron.** On `Selector` and `MultiSelector` the end slot is a ternary: a status glyph renders *instead of* the chevron when a status is attached. `hasChevron` gates the chevron arm alone — a status glyph reports on the value, so suppressing it would hide information rather than an affordance. With an attached status the flag is a no-op, which is asserted.
- **a11y-neutral, verified in the markup rather than assumed.** `Icon` without a `label` emits `aria-hidden="true"`, and the chevron is a sibling of the `<button>`, not a descendant — so it is neither in the a11y tree nor part of the button's name computation. Both facts are asserted in tests, alongside accessible name, `textContent`, and tab order being identical across the two modes.

### Why `hasChevron`, not `hasIndicator`

`indicator` is **already taken on two of the three components, meaning something else**. `Selector` and `MultiSelector` both expose `indicatorPosition` — which edge of an *option row inside the popup* carries the selected mark (a check on Selector, a checkbox on MultiSelector). `Selector.tsx` imports `useIndicator` and `IndicatorPosition` from the `Indicator` module for exactly that. `hasIndicator={false}` next to `indicatorPosition="start"` would read as "turn off the option checkmarks."

`hasChevron` is the existing house name for this exact affordance: `DropdownMenu` already ships `hasChevron?: boolean` defaulting to `true` to suppress its trigger chevron — and its theme target is likewise named `astryx-dropdown-menu-indicator-icon`, so the theme-target/prop-name split is precedent, not a new inconsistency. It also matches the `has*` convention for presentational sub-affordances (`hasClear`, `hasSearch`, `hasHover`, `hasRowHighlight`).

## Per-component notes

They are consistent in name, default, and semantics, but the trigger markup differs in two ways worth flagging:

- **`ComplexSelector` has no `hasClear`.** Its end slot holds only the chevron, so the clear-replaces-chevron motivation does not apply. The flag is still useful (a trigger whose composed content supplies its own end affordance), and its doc example says so rather than reusing the clear-button framing.
- **`ComplexSelector` renders no on-field status glyph**, so its chevron was unconditional. It becomes `{hasChevron && <Icon …/>}`; the other two become a third ternary arm.

## Docs

Per-file, matching each doc's actual shape rather than assuming three surfaces everywhere:

- `Selector.doc.mjs` — `docs.props` + a new `examples` entry. Its `docsZh`/`docsDense` are usage-only overlays with no `propDescriptions` at all; adding a lone Chinese entry would have been the only one in the file, so it was left alone.
- `MultiSelector.doc.mjs` — all three: `docs…props`, `docsZh…propDescriptions` (translated), `docsDense…propDescriptions`, plus an `examples` entry.
- `ComplexSelector.doc.mjs` — `docs…props` + `docsDense.propDescriptions` (it has no `docsZh`), plus an `examples` entry.

`SYNC:` headers were checked on all three sources — each already lists its own `.doc.mjs`, so unlike #5310 there was no missing entry to add. The lists also name storybook stories and `packages/cli/assets/templates/blocks/**`; neither needs a change, since the prop is additive with an unchanged default (and templates were out of bounds for this PR).

Changeset: `[feat]` / `patch`, per the 0.x coupling `check:changesets` enforces (only `[breaking]` may bump the minor while 0.x).

## Test plan

19 tests added — 7 on `Selector`, 7 on `MultiSelector`, 5 on `ComplexSelector`:

- [x] The chevron renders by default (all three)
- [x] `hasChevron={false}` removes it (all three)
- [x] With `hasClear` + a value, the clear button is still present and the chevron is gone (Selector, MultiSelector)
- [x] Accessible name, trigger `textContent`, and tab order are identical across both modes (all three)
- [x] The chevron is `aria-hidden="true"` and is **not** contained by the trigger button (all three)
- [x] An attached status glyph still renders with the chevron off, and the icon count is the same either way (Selector, MultiSelector)
- [x] The popup still opens and `aria-expanded` still flips with the chevron off (all three)

**Non-vacuity was checked, not assumed.** I stashed only the three `.tsx` sources, kept the tests, and re-ran: **11 of the 19 fail without the change.** The 8 that pass either way do so by construction, and I want to be straight about which:

- 3 × "renders the chevron by default" — a default-preservation guard necessarily passes both ways; its value is catching a future change of default.
- 3 × "chevron is aria-hidden and outside the button" — deliberate characterization of *pre-existing* markup. This is the evidence for the a11y-neutrality claim, so it is meant to hold before and after.
- 2 × "keeps the status glyph" — with an attached status the chevron arm never runs either way, so this is a forward guard on the semantic split (it fails if someone later widens the flag to gate the status icon).

Every test that asserts the opt-out itself fails without the change; each `hasChevron={false}` render asserts the chevron is absent so none of them can drift vacuous.

Gates:

- [x] `pnpm -F @astryxdesign/core typecheck` and `typecheck:docs` — clean
- [x] `eslint` on all 10 changed files at `ASTRYX_STRICT_LINT=1` — 0 errors (the one warning, unused `borderVars` in `MultiSelector.tsx`, reproduces on `origin/main` and is untouched here)
- [x] `prettier --check` on all 10 changed files — clean
- [x] `vitest run` — Selector 150 passed, MultiSelector 112 passed, ComplexSelector 17 passed (279 total, 3 files)
- [x] `pnpm check:repo` — exit 0
- [x] `astryx component <Name> --dense` renders the new prop row and example for all three

### DIFF
--- a/packages/core/src/Selector/Selector.doc.mjs
+++ b/packages/core/src/Selector/Selector.doc.mjs
@@ -84,6 +84,13 @@ export const docs = {
       default: 'false',
     },
     {
+      name: 'hasChevron',
+      type: 'boolean',
+      description:
+        'Shows the chevron at the end of the trigger. Set false to let the clear button stand alone in that slot — with hasClear, a selector that has a value otherwise shows both a × and a chevron. Only the chevron is dropped: a status glyph shares the slot and still appears, and since the chevron is decorative (aria-hidden) and sits outside the trigger button, the accessible name, focus order, and keyboard behaviour are unchanged.',
+      default: 'true',
+    },
+    {
       name: 'searchPlaceholder',
       type: 'string',
       description: 'Placeholder text for the search input.',
@@ -292,6 +299,24 @@ export const docs = {
       },
     ],
   },
+  examples: [
+    {
+      label: 'Let the clear button replace the chevron',
+      code: `
+// With hasClear alone, a selector that has a value shows both a × and a
+// chevron in the end slot. hasChevron={false} leaves the × on its own, so
+// the one affordance in that slot is the one the user can act on.
+<Selector
+  label="Country"
+  options={countries}
+  value={country}
+  onChange={setCountry}
+  hasClear
+  hasChevron={false}
+/>
+`,
+    },
+  ],
 };
 
 /** @type {import('@astryxdesign/cli/authoring').ComponentTranslationDoc} */
--- a/packages/core/src/Selector/Selector.tsx
+++ b/packages/core/src/Selector/Selector.tsx
@@ -631,6 +631,27 @@ interface SelectorPropsBase<
   hasSearch?: boolean;
 
   /**
+   * Whether to show the chevron at the end of the trigger. Set false to let
+   * the clear button stand alone in that slot instead of sitting beside a
+   * second glyph — with `hasClear`, a selector that has a value otherwise
+   * shows both.
+   *
+   * This is about the chevron only. A status glyph takes the same slot and is
+   * unaffected: an errored selector still shows its status icon.
+   *
+   * The chevron is decorative (`aria-hidden`) and sits outside the trigger
+   * button, so dropping it leaves the accessible name, the focus order, and
+   * the keyboard behaviour untouched.
+   *
+   * @default true
+   * @example
+   * ```
+   * <Selector hasClear hasChevron={false} value={value} onChange={setValue} />
+   * ```
+   */
+  hasChevron?: boolean;
+
+  /**
    * Placeholder text for the search input.
    * @default 'Search...'
    */
@@ -781,6 +802,7 @@ export function Selector<T extends SelectorOptionType>(
     renderValue,
     indicatorPosition = 'end',
     hasSearch = false,
+    hasChevron = true,
     searchPlaceholder: searchPlaceholderFromProps,
     placement,
     isDefaultOpen = false,
@@ -1528,6 +1550,10 @@ export function Selector<T extends SelectorOptionType>(
           and the icon color, so the status glyph and the chevron are each
           directly targetable instead of sharing one untargetable parent — and
           the two affordances stop sharing a node.
+
+          `hasChevron` gates only the chevron arm. A status glyph is a report
+          about the value, not an expand affordance, so it keeps the slot even
+          when the chevron is off.
         */}
         {showStatusIcon ? (
           showStatusTooltip ? (
@@ -1556,7 +1582,7 @@ export function Selector<T extends SelectorOptionType>(
               xstyle={styles.triggerIcon}
             />
           )
-        ) : (
+        ) : hasChevron ? (
           <Icon
             icon="chevronDown"
             size="sm"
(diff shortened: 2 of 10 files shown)


packages/core/src/Selector/Selector.tsx
```diff
@@ -1556,7 +1582,7 @@ export function Selector<T extends SelectorOptionType>(
               xstyle={styles.triggerIcon}
             />
           )
-        ) : (
+        ) : hasChevron ? (
```
cixzhang, line 1585: With no value this arm is gone and so is the clear button — nothing left in the slot.

packages/core/src/Selector/Selector.doc.mjs
```diff
@@ -292,6 +299,24 @@ export const docs = {
       },
     ],
   },
+  examples: [
+    {
+      label: 'Let the clear button replace the chevron',
```
cixzhang, line 304: This example renders an empty end slot until a value is picked.

cixzhang (dismissed):
Thanks for this — the double glyph with `hasClear` is real, and the naming call against `indicatorPosition` is right.

`hasChevron` is static though, so the documented snippet — `<Selector hasClear hasChevron={false} />` — leaves the end slot completely empty until someone picks a value: no ×, no chevron, and the field stops reading as a picker. Same on MultiSelector, and any time it's disabled.

The example is labelled "let the clear button replace the chevron", and the component already knows when the clear button is showing:

```tsx
const showsClear = hasClear && value != null && !isDisabled;
) : hasChevron && !showsClear ? (
```

That covers the motivating case with no new prop, and leaves `hasChevron` for ComplexSelector's composed-affordance case.

Would you rather clear just takes the slot when it's there, or keep the flag and have it drop the chevron only while the × shows?

[Reviewed by Robohands]

cixzhang:
Checking in on the chevron/clear-slot choice: the current review needs the trigger affordance to remain present before a value is selected. Would you like to continue, have us carry it, or close this for now?

ernestt:
Release-process check: .changeset/selector-chevron-opt-out.md uses the correct patch bump, but its first body line must end in (#5318) and its second body line must be the contributor handle @ernestt; the handle is currently after the release notes.

PR #5320, opened 2026-08-22
feat(cli): search sees your integrations, and build stops calling one word a match

josephfarina:
Three changes to the two commands an agent reaches for first.

## `search` could not see integrations

`search` gathered components straight off the resolved core directory, so a package listed in `astryx.config.mjs` was reachable by `component` and `template` and invisible to the one command whose job is finding things. The search command already loaded a `Project` — purely to print integration warnings beside results that could not contain an integration's components.

It now gathers through `Project`, so an integration's components rank next to Core's and carry their own `package` and `import`. The per-integration skip+warn policy is inherited, so one broken manifest is skipped rather than failing the command, and `Project` is loaded once per search and shared with the docs gatherer.

## Usage guidance was not indexed

Component candidates carried name + keywords + `usage.description`. The reader's vocabulary usually lives further down: `Banner` describes itself as "a persistent message", and only its guidance names the form errors and maintenance notices someone would search for.

`features` and best-practice text are now indexed, at a tier below the description. The tiers stay apart deliberately — collapsed into one, `Toast` ("a brief, non-blocking notification") ties with every component that mentions notifications in passing advice, and ties break alphabetically.

`MIN_TOKEN_SCORE` moves with the new tier, since it gates which per-token hits count in the multi-word pass and `build` sends multi-word queries almost exclusively.

## `build` called a one-word coincidence a direct match

```
astryx --json build "actionable warning banner"
→ directMatch: true
→ pages: contact-form 95, documentation-design 95, login 95
```

None of those is a warning page. A page's keywords include every component its source renders, so each keyword-matched "banner" at 90; the multi-word branch adds a coverage garnish of up to 15 points, landing exactly on the 95 direct-match threshold. Coverage was decorating the score when it should have been gating it — matching one of three concepts is not the claim that matching three is.

Results now carry `matchedTerms` / `queryTerms` (a whole-phrase match reports full coverage), and the pages group gates on them. Consumers no longer have to parse `"matches 1/3 terms: banner"` back out of the reason string.

## A thin kit says what to try

Left unsaid, an empty kit reads as "the package has nothing for this" rather than "these words missed", and the caller falls back on whatever it already believed about the package — which is the failure `build` exists to prevent. `build.kit` now carries an optional `hint` when the kit is nearly empty, naming the browse commands; the renderer prints it.

## Testing

- Full suite green: 544 files, 11280 tests. `lint:strict` clean on every touched file; `check:cli-structure`, `typecheck:authoring`, `typecheck:strict`, `readme:check` all pass; API declarations regenerated with `sync:api-types`.
- New tests cover each behaviour, including a temp-directory fixture integration (manifest + doc + same-stem source) proving an integration-owned component is found with its package and import path, and that Core results still come back beside it.
- `directMatch` positive control: `build "contact form"` still reports a direct match at 2/2 coverage.

The type additions are additive.


cixzhang (dismissed):
Thanks — two headline fixes landed while this was open: [#5259](https://github.com/facebook/astryx/pull/5259) added integration search and [#5614](https://github.com/facebook/astryx/pull/5614) fixed false direct matches at the scorer, so this head now conflicts in both search files.

The remaining rewrite regresses two established paths. A builder following an integration result gets `@astryxdesign/core`, while `component` correctly reports `@acme/widgets/DiffLink`. An invalid or duplicate config also turns the parent’s 20-result Core `button` search into an empty success (exit 0), telling the agent Astryx has nothing.

Please rebase onto current `main` and reuse the component lookup/fallback path. Could the guidance indexing and thin-kit hint return as separate small PRs, each with true red→green evidence and updated public response docs?

[Reviewed by Robohands]

PR #5321, opened 2026-08-22
fix(Avatar): use light-dark() for fallback initials to pass WCAG AA contrast in dark mode

### LINKED ISSUE #5279: Avatar fallback initials fail WCAG AA contrast in dark mode (--color-text-secondary on --color-neutral)
**Bug:** `Avatar`'s fallback initials use `--color-text-secondary` on `--color-neutral`, which fails WCAG AA contrast (4.5:1 for normal text) in dark mode, though it passes comfortably in light mode.

## Evidence

Computed contrast (relative luminance per WCAG, `--color-neutral` composited over its typical surface background):

| Theme | Effective background | Text color | Contrast | AA (4.5:1)? |
|---|---|---|---|---|
| Light | `rgb(230, 235, 238)` (`--color-neutral` @ 0.1 alpha over `--color-background-surface` #FFFFFF) | `--color-text-secondary` `#4E606F` | **5.42:1** | Pass |
| Dark | `rgb(69, 70, 73)` (`--color-neutral` @ 0.2 alpha over `--color-background-surface` #1F1F22) | `--color-text-secondary` `#AAAFB5` | **4.27:1** | **Fail** |

For reference, `--color-text-primary` on the same dark background computes to 7.25:1 (comfortable pass), and 15.63:1 in light mode.

## Where this surfaces

`packages/core/src/Avatar/Avatar.tsx`, the `fallback` style block (initials + default icon surface): `backgroundColor: colorVars['--color-neutral']`, `color: colorVars['--color-text-secondary']`.

Found while reviewing #4327, which briefly switched this to `--color-text-primary` to fix the dark-mode failure, but that also changes light mode's already-passing appearance for every existing avatar (light mode goes from 5.42:1 to 15.63:1, a much starker text weight than the rest of the design intends for secondary/muted surfaces). That change was reverted out of #4327 as out of scope, pending a proper fix here.

## Open question for whoever picks this up

Two directions, worth a maintainer call before implementing:
- **Component-local**: give `Avatar`'s fallback surface a theme-aware color (e.g. `light-dark(var(--color-text-secondary), var(--color-text-primary))` or similar), keeping other consumers of `--color-text-secondary` on `--color-neutral` untouched.
- **Token-level**: if `--color-text-secondary`'s dark value is under-contrasted against `--color-neutral` specifically (not necessarily other neutral-adjacent surfaces), it may be worth checking whether other components pairing these same two tokens have the identical dark-mode gap, in which case a token-level fix might be the more correct scope.

I didn't check every other consumer of this exact token pairing, so I can't say how widespread it is beyond `Avatar`.

Abidit:
## Summary

Fixes #5279

Avatar's fallback initials (shown when no image is set) used 
`--color-text-secondary` on `--color-neutral` background, which 
fails WCAG AA contrast (4.5:1) in dark mode at 4.27:1.

## Fix

Switch to `light-dark()` so each color mode gets the appropriate 
text color:

- **Light mode**: `--color-text-secondary` → 5.42:1 ✅ (unchanged)
- **Dark mode**: `--color-text-primary` → 7.25:1 ✅ (was 4.27:1 ❌)

This matches the existing convention in `useTableStickyColumns.tsx`.

## Change

One line changed in `Avatar.tsx`:

```diff
- color: colorVars['--color-text-secondary'],
+ color: 'light-dark(var(--color-text-secondary), var(--color-text-primary))',
```


Abhirup0:
Hi @Abidit! Nice catch and thank you for putting up this fix!

Using `light-dark()` is definitely the right solution here — it fixes the dark mode contrast violation (7.25:1, passing WCAG AA) while keeping the light mode secondary look intact (5.42:1).

Just a few small things that might help get this PR merge-ready:

1. Changeset:
Astryx requires a changeset for packages that get published so CI can pass (`check:changesets`). You can add a file like `.changeset/avatar-fallback-dark-contrast.md` with:

---
'@astryxdesign/core': patch
---

[fix] Avatar: resolve fallback initials and default icon color via light-dark() to meet WCAG AA contrast in dark mode (#5279)
@Abidit


2. Token variable reference:
In `Avatar.tsx`, styles typically use the imported `colorVars` tokens directly rather than raw CSS strings:
```ts
color: `light-dark(${colorVars['--color-text-secondary']}, ${colorVars['--color-text-primary']})`,
```

3. Regression tests:
Adding a quick unit test in `Avatar.test.tsx` checking that fallback initials satisfy WCAG AA contrast in both light and dark modes will help make sure this doesn't regress in the future.

4. (Bonus / Optional) AvatarGroupOverflow:
`packages/core/src/AvatarGroup/AvatarGroupOverflow.tsx` (the `+N` count indicator) uses the exact same `--color-neutral` background and has the same dark mode text color. Updating it here would make the whole Avatar family fully WCAG AA compliant!

Abidit:
> **[Abhirup0](/Abhirup0) **

Thanks for the detailed review! I've addressed all four points:
1. Changeset added
2. Switched to colorVars token references
3. Added regression tests for both Avatar and AvatarGroupOverflow
4. Fixed AvatarGroupOverflow +N indicator with the same light-dark() approach

.changeset/avatar-fallback-dark-contrast.md
```diff
@@ -0,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Avatar: resolve fallback initials color via light-dark() to meet WCAG AA contrast in dark mode (#5279)
```
cixzhang: Changeset says Avatar only. Might want to name the AvatarGroup +N indicator too.

cixzhang (dismissed):
Thanks, this holds up — measured 4.37:1 to 10.57:1 in dark, light byte-identical. Abhirup0's four points all landed. Merging; the wider secondary-on-neutral pairing is ours to sort.

[Reviewed by Robohands]

.changeset/avatar-fallback-dark-contrast.md
Abidit: Good catch — updated the changeset to name both: "Avatar, AvatarGroup: resolve fallback initials and +N overflow indicator color via light-dark()..." `AvatarGroupOverflow` got the same fix in this PR, so it should be called out. Pushed in 8ebfe895.

cixzhang:
Merged `main` in — `Avatar.test.tsx` conflicted only because both sides added imports at the top; kept both. 130 Avatar/AvatarGroup tests pass. Re-armed auto-merge.

cixzhang (changes requested):
Thanks for this, and sorry for the churn — I approved it earlier today and I was wrong to. The measurement is solid: 4.37:1 to 10.57:1 in dark, light byte-identical, and the `+N` chip caught alongside the initials.

The problem is the layer. `light-dark()` in a component picks one of the two schemes by hand, so this can only ever fix dark — chocolate at 2.82:1, matcha at 4.00:1 and stone at 3.36:1 still fail these same initials in LIGHT, and no version of this mechanism reaches them. `Kbd` pairs the identical two tokens and measures the same.

The fix belongs in the token pair, once, where it reaches both schemes and every component that pairs them. That one is ours — we'll do it and credit you, and I'd rather you didn't spend another round here.

[Reviewed by Robohands]

cixzhang:
Thank you for investigating and testing this contrast issue. Color-token and palette changes are now coordinated as owner-led accessibility work rather than parallel implementation PRs. Please follow #4475 for the program direction; we are closing this attempt.

PR #5322, opened 2026-08-22
fix(BottomSheet): guard standalone Escape against IME composition

### LINKED ISSUE #5302: [audit] BottomSheet: standalone Escape lacks an IME composition guard (I19)
**Bug:** the standalone Bottom Sheet dismisses on Escape without guarding against IME composition, so a CJK (Korean/Japanese/Chinese) user pressing Escape to *cancel an in-progress composition* dismisses the sheet instead.

When an IME is composing, the browser fires a `keydown` for Escape (cancel) / Enter (commit) **before** `compositionend`. A handler that reads a bare `event.key === 'Escape'` misreads the composition-cancel as an app command. `utils/ime.ts` names closing a dialog mid-composition as exactly this bug, and the fix is `isImeKeyEvent(event.nativeEvent)` early-return.

The base `Dialog` guards it (`Dialog.tsx`, in its Escape `keydown` handler) and so does the sibling `BottomSheetSwitcher` (its `handleKeyDown` calls `isImeKeyEvent(event.nativeEvent)`, with a dedicated test). The **standalone** `BottomSheet` does not — neither its `handleKeyDown` nor its native `onCancel` (`handleCancel`) guards.

**Where:** `packages/core/src/BottomSheet/BottomSheet.tsx` — `handleKeyDown` and `handleCancel` in `StandaloneBottomSheet`.

**Reachable path:** the documented `purpose="form"` and mobile-keyboard (`height="tall"`) use cases put a text field in a standalone sheet. A CJK user composing in that field who presses Escape to cancel the candidate loses the whole sheet.

**Why lint doesn't catch it:** `@astryx/no-unguarded-ime-keydown` only flags `onKeyDown` on an *editable surface* (input/textarea/contentEditable/role=textbox). The handler here is on the `<dialog>`, which the rule's heuristic treats as non-editable, so this is a manual finding.

**Fix:** early-return on `isImeKeyEvent(event.nativeEvent)` at the top of both `handleKeyDown` and `handleCancel`, mirroring `Dialog`, and add a standalone IME-Escape test (the switcher already has one). The switcher path is already correct.

---

Filed from a full Component Audit Rubric v1.6 grading of BottomSheet (mode O). Rule: **§9 I19** (IME composition guard). Component ledger entry: `core/BottomSheet` — 89.1 / C (this is the one open BLOCK; the C cap lifts and the projection is A once it's fixed). Rubric: https://github.com/facebook/astryx/wiki/Component-Audit-Rubric

AKnassa:
## What this does

A bottom sheet no longer closes when someone typing Korean, Japanese or Chinese presses Escape to back out of a half-typed word.

## Why

While an IME is composing, the browser sends an Escape keydown to cancel the pending characters, and it sends it *before* `compositionend`. The standalone sheet read that as a plain "close me" command, so a CJK user correcting a word inside a sheet lost the whole sheet, along with anything already typed into it. The documented `purpose="form"` and `height="tall"` mobile-keyboard cases both put a text field in a standalone sheet, so this is an ordinary path rather than a corner.

The base `Dialog` already guards this, and so does the sibling `BottomSheetSwitcher`. Only the standalone sheet was missing it.

## What changed

- The sheet's Escape handler now ignores the keydown that belongs to an in-progress composition, via the shared `isImeKeyEvent` check the rest of the library uses.
- It claims the Escape *before* deciding what the key means. Left unclaimed, the browser raises its own close request, which arrives at the sheet's `onCancel` and closes it on the very same keypress. The guard on its own is not enough.
- `handleCancel` is unchanged, and gains a comment saying why: `cancel` is a plain `Event` carrying no composition state, so a guard there could never fire.
- Five tests, and a changeset.

## Evidence

This is a bug fix, so the bar is red before and green after.

The tests are red at `fba7b4009` (the branch base) and green after the fix. Reverting `handleKeyDown` to its previous shape turns four of the five red on their behaviour assertion (`onOpenChange` called once where it must not be called at all), and hoisting `preventDefault` above the key check turns the fifth red on its own. No test here passes vacuously, and the two halves of the fix are pinned separately.

Covered: both IME signals (`isComposing`, and the legacy keyCode 229); a composing Escape raised on a focused field and bubbling to the `<dialog>`; the release case, where the Escape *after* a cancelled composition still dismisses; the non-modal sheet, where Escape is the only route out; and Enter and ArrowDown mid-composition staying unclaimed, since the IME owns those keys.

Local: `pnpm lint:strict` clean, core typecheck clean, `pnpm build` green, BottomSheet suite 77/77, full suite 11289/11292. The three remaining failures are pre-existing and not from this branch: two cli tests that assume a case-sensitive filesystem, and one `DateInputTouch` test that times out at 5s only under load (127/127 in isolation, and `packages/core/src/DateInput/` is byte-identical to `main` on this branch).

No screenshots: nothing rendered changes, and a real IME composition cannot be driven from browser automation. A synthetic `KeyboardEvent` is untrusted, so it would not exercise the close-request path either.

## Rubric

`Triage: bug fix · non-breaking · low blast radius → fast path · checks: §1 A1-A7/A11-A12, §7 C1-C8/C11, Q1-Q12`

Component Audit Rubric v1.6, check §9 I19. §1: Escape stays operable; no ARIA, focus or live region touched. §7: no Effect, state, ref or observer added, and the shared `isImeKeyEvent` is composed rather than reimplemented. The component is not audited around the fix.

One note for the ledger row: the issue asks for the guard in both `handleKeyDown` and `handleCancel`. Only the first is implementable, for the reason above. The `preventDefault` hoist is what actually closes the cancel path.

Fixes #5302

### DIFF
--- a/.changeset/bottom-sheet-ime-escape-guard.md
+++ b/.changeset/bottom-sheet-ime-escape-guard.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] BottomSheet: a standalone sheet no longer dismisses when a CJK user presses Escape to cancel an in-progress IME composition. The browser fires that keydown before `compositionend`, so an Escape handler reading a bare `event.key` misread the composition cancel as a dismissal command and closed the sheet — losing whatever had been typed into a `purpose="form"` field inside it. The handler now early-returns on `isImeKeyEvent`, the same guard `Dialog` and `BottomSheetSwitcher` already carry, and claims the key first so the browser raises no close request of its own.
+
+@ernestt
--- a/packages/core/src/BottomSheet/BottomSheet.tsx
+++ b/packages/core/src/BottomSheet/BottomSheet.tsx
@@ -37,7 +37,7 @@ import * as stylex from '@stylexjs/stylex';
 import type {BaseProps} from '../BaseProps';
 import type {DialogPurpose} from '../Dialog';
 import {colorVars, durationVars, easeVars} from '../theme/tokens.stylex';
-import {useDevWarning, useScrollLock} from '../hooks';
+import {isImeKeyEvent, useDevWarning, useScrollLock} from '../hooks';
 import {
   BottomSheetPanel,
   type BottomSheetPanelMotion,
@@ -298,6 +298,9 @@ function StandaloneBottomSheet({
 
   const handleCancel = useCallback(
     (event: React.SyntheticEvent<HTMLDialogElement>) => {
+      // No IME guard here: `cancel` is a plain Event carrying no composition
+      // state, and handleKeyDown claims a composing Escape before the browser
+      // can raise the close request that would arrive here.
       event.preventDefault();
       dismissOnEscape();
     },
@@ -305,10 +308,20 @@ function StandaloneBottomSheet({
   );
   const handleKeyDown = useCallback(
     (event: React.KeyboardEvent<HTMLDialogElement>) => {
-      if (event.key === 'Escape') {
-        event.preventDefault();
-        dismissOnEscape();
+      if (event.key !== 'Escape') {
+        return;
       }
+      // Claim the key before reading it: an unclaimed Escape lets the browser
+      // raise its own close request, which lands on handleCancel and dismisses
+      // on the same keypress.
+      event.preventDefault();
+      // An IME fires this keydown to cancel an in-progress composition, ahead
+      // of compositionend. It is a composition cancel, not a dismissal command
+      // — see utils/ime; Dialog and BottomSheetSwitcher guard the same way.
+      if (isImeKeyEvent(event.nativeEvent)) {
+        return;
+      }
+      dismissOnEscape();
     },
     [dismissOnEscape],
   );
--- a/packages/core/src/BottomSheet/BottomSheet.test.tsx
+++ b/packages/core/src/BottomSheet/BottomSheet.test.tsx
@@ -334,6 +334,42 @@ describe('BottomSheet', () => {
     expect(onOpenChange).toHaveBeenCalledWith(false);
   });
 
+  it('ignores Escape while an IME composition is active', () => {
+    const onOpenChange = vi.fn();
+    render(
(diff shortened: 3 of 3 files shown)


packages/core/src/BottomSheet/BottomSheet.tsx
```diff
@@ -298,17 +298,30 @@ function StandaloneBottomSheet({
...
+      if (event.key !== 'Escape') {
+        return;
       }
+      // Claim the key before reading it: an unclaimed Escape lets the browser
+      // raise its own close request, which lands on handleCancel and dismisses
+      // on the same keypress.
+      event.preventDefault();
+      // An IME fires this keydown to cancel an in-progress composition, ahead
+      // of compositionend. It is a composition cancel, not a dismissal command
+      // — see utils/ime; Dialog and BottomSheetSwitcher guard the same way.
```
cixzhang, line 320: Backwards, and worth fixing while you are here: neither claims the key, which is why both still close mid-composition. Yours is the one that gets it right.

.changeset/bottom-sheet-ime-escape-guard.md
```diff
@@ -0,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] BottomSheet: a standalone sheet no longer dismisses when a CJK user presses Escape to cancel an in-progress IME composition. The browser fires that keydown before `compositionend`, so an Escape handler reading a bare `event.key` misread the composition cancel as a dismissal command and closed the sheet — losing whatever had been typed into a `purpose="form"` field inside it. The handler now early-returns on `isImeKeyEvent`, the same guard `Dialog` and `BottomSheetSwitcher` already carry, and claims the key first so the browser raises no close request of its own.
+
+@ernestt
```
cixzhang, line 7: Credit line is the PR author by convention — `@AKnassa`.

cixzhang (approved):
Thanks — claiming the key before `isImeKeyEvent` is the half that actually matters, and Chromium agrees: without it the close request still fires and the modal sheet goes anyway. Two nits inline.

[Reviewed by Robohands]

PR #5323, opened 2026-08-22
fix(a11y): a disabled control answers the pointer with not-allowed

cixzhang:
## Why

The cursor is the only affordance a pointer user gets **before** they commit to a click. A disabled control answering with `pointer` promises a click it will not honour; one answering with `default` says nothing at all, and the user learns the control is dead by clicking it and watching nothing happen.

Nothing took that promise away for us. `disabled` and `[aria-disabled]` do not change what an element's own `cursor` declaration paints, and the per-component `disabled` style only helps where the author remembered to write one, on the element the author had in mind. A Chromium sweep over every story found **26 disabled elements answering with the wrong cursor** — `pointer` on Selector's and MultiSelector's triggers, `grab` on the drag handles in ListInput and TransferList, `default` on SegmentedControl's segments, `auto` on TreeList's rows.

One of those deserves calling out, because it is the shape that will keep producing this bug: **StyleX merges `props()` one PROPERTY at a time**, so a later style setting `cursor` at all replaces the earlier declaration's *conditions* along with its value. SegmentedControlItem held a correctly guarded `cursor: pointer` on its base and a flat `disabled: {cursor: 'default'}` applied after it — and shipped a disabled segment answering with a plain arrow. Both halves read as correct in review. That is why the guard has to go on every declaration, not on the ones we judge interactive.

Sibling of [#5247](https://github.com/facebook/astryx/pull/5247): that one stopped a disabled control from *lighting up* under the pointer, this one makes it *say what it is*.

## What

- Every `cursor` in core and lab carries `':is(:disabled,[aria-disabled="true"])': 'not-allowed'` — 128 declarations across 76 files, all applied by the new rule's autofix. `:is()` outranks the default in StyleX's own ordering (priority 3040 vs 3000), and on an element that can never be disabled it is a no-op.
- `reset.css` gives the same cursor to any disabled element that declares none. It previously said `cursor: default`, and only for native `:disabled` — `[aria-disabled]`, which is how astryx keeps a disabled control focusable, got nothing.
- `@astryx/disabled-cursor` — a new autofixable lint rule, error in core and lab, so an unguarded cursor cannot land again.
- `.github/scripts/disabled-cursor-audit.js` (`pnpm guard:disabled-cursor`) — a Chromium sweep that hit-tests five points per disabled element in every story and fails on any cursor other than `not-allowed`. Five points, not one: a child that declares its own cursor often covers only part of a control.
- Two `disabled` styles that set `cursor: 'default'` (SegmentedControlItem, ResizeHandle) now say `not-allowed` outright.

## Risk

Behavioral change is confined to what a disabled element shows under the pointer. No selector that was not already scoped to a disabled element changes; no component's markup, props or events move.

**Known limit, unchanged by this PR:** 75 of the 635 disabled elements in the story set are sealed behind `pointer-events: none`, so the pointer lands on an ancestor and the control never gets to answer — the sweep counts these separately rather than passing them. Making those answer means removing `pointer-events: none`, which is a separate change with real behavior behind it: on Item, TreeListItem and SideNavItem the disabled row still renders a live `href`, and today only `pointer-events: none` stops that click. Worth doing, not here.

No screenshots: the cursor is the one visual property a screenshot cannot capture. The measurement below is the evidence.

## Testing

- `pnpm guard:disabled-cursor` over all 1700 stories: **26 violations → 0**, 635 disabled elements checked, 75 sealed (see above).
- `@astryx/disabled-cursor` over core and lab: **117 findings → 0**.
- Unit: the rule's `RuleTester` suite, the sweep's verdict/CLI suite, and a new `reset.test.ts` invariant that every disabled rule in the reset carries `not-allowed`. `Table.perf.test.tsx` failed once under parallel load and passes alone — unrelated timing test.
- Added as **A21** to the [Component Audit Rubric](https://github.com/facebook/astryx/wiki/Component-Audit-Rubric) (v1.7), scored `auto`; A20 hands it the cursor rider it was carrying.

### DIFF
--- a/.changeset/disabled-cursor-guarantee.md
+++ b/.changeset/disabled-cursor-guarantee.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] A disabled control now answers the pointer with `not-allowed` wherever it can be pointed at: every `cursor` in core and lab carries `':is(:disabled,[aria-disabled="true"])': 'not-allowed'`, and the reset gives the same cursor to any disabled element that declares none — `[aria-disabled]` included, where it previously said `default` and only for native `:disabled`. A lint rule and a Chromium sweep over every story keep it that way. Disabled elements sealed behind `pointer-events: none` are unchanged: the pointer never reaches them, so their cursor comes from an ancestor.
+
+@cixzhang
--- a/.github/scripts/disabled-cursor-audit.js
+++ b/.github/scripts/disabled-cursor-audit.js
@@ -1,0 +1,417 @@
+#!/usr/bin/env node
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+
+/**
+ * @description Asserts every disabled element answers the pointer with `not-allowed`
+ * @input --storybook-dir <path> [--components <csv>] [--output <file>]
+ *   [--concurrency <n>] [--port <n>]
+ * @output JSON report of violations; exit 1 if any disabled element shows a
+ *   cursor other than `not-allowed`
+ *
+ * The cursor is the only affordance a pointer user gets before they commit to
+ * a click. A disabled control that answers with `pointer` promises a click it
+ * will not honour, and one that answers with `default` says nothing at all —
+ * the user learns the control is dead only by clicking it and watching
+ * nothing happen.
+ *
+ * jsdom has no cursor and no hit testing, so the unit suite cannot see any of
+ * this. Chromium can. For every disabled element in every story we hit-test
+ * the points a pointer would actually land on and read the computed `cursor`
+ * of the element that answers — which is the element the pointer hits, not
+ * necessarily the disabled one: `cursor` inherits, so a child that declares
+ * its own wins under the pointer, and a subtree behind `pointer-events: none`
+ * hands the question to an ancestor.
+ *
+ * The gate is zero-tolerance and has no baseline: `not-allowed` on a disabled
+ * control is never wrong, so there is nothing to grandfather. What it does
+ * skip is an element nothing can point at — no box, or covered by something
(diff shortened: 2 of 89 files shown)


cixzhang:
**Self-review.** Walked the diff, and verified the parts that are easy to get wrong by measurement rather than reading:

- **The guard compiles and outranks the default.** StyleX emits `:is(:disabled,[aria-disabled="true"]){cursor:not-allowed}` at priority 3040 against the plain declaration's 3000, and dedups all 128 sites onto **one** atomic class in `astryx.css` — the breadth costs one rule, not 128.
- **The sweep is measuring the right thing.** It hit-tests, so it reads the cursor of the element the pointer actually finds: a child that declares its own over part of the control, or an ancestor where the subtree is `pointer-events: none`. That is what caught SegmentedControl after the first pass looked clean in source.
- **`reset.css` is a floor, not the mechanism.** Component StyleX classes outweigh `@layer reset`, so the reset only decides for elements that declare no cursor — which is why the lint rule has to exist as well.

Two judgement calls worth a second opinion:

1. **The rule is unconditional** — it guards `cursor: 'default'` on `SideNavHeading:69` and `Lightbox:152` too, where nothing can ever be disabled and the guard is a no-op. Same trade [#5247](https://github.com/facebook/astryx/pull/5247) made for `:hover`: asking the question per component is what leaves the gaps.
2. **`:is()` adds specificity, `:where()` would not.** Deliberate here — the guard has to *beat* the default declaration, where the hover guard only had to weigh the same. A consumer's own unlayered `cursor` on a disabled element still wins.

Known limit is in the description: 75 sealed elements, and the follow-up (live `href` on disabled rows in Item/TreeListItem/SideNavItem) is the real reason not to pull `pointer-events: none` in this PR.

PR #5325, opened 2026-08-22
fix(core): a blank client-hints platform is unknown, not non-Apple

### LINKED ISSUE #5253: [Bug] Empty userAgentData.platform is treated as non-Apple, so mod hotkeys and Kbd show Ctrl on macOS
### Description

`isApplePlatform()` in `packages/core/src/hooks/useHotkeys.ts` and the matching detection in `packages/core/src/Kbd/Kbd.tsx` commit to `userAgentData` and can never fall through:

```js
const uaData = 'userAgentData' in navigator ? navigator.userAgentData : null;
if (uaData && typeof uaData === 'object' && 'platform' in uaData) {
  return /mac/i.test(uaData.platform ?? '');   // '' → false, fallback unreachable
}
return /Mac|iPhone|iPad|iPod/.test(navigator.platform ?? '');
```

When `userAgentData.platform` is present but empty, the `'platform' in uaData` guard passes, `/mac/i.test('')` is `false`, and the `navigator.platform` branch — which would answer correctly — is never reached. An empty string is being read as a negative answer rather than as no answer.

Electron builds that rewrite the app's user-agent / client-hints identity ship exactly that: `platform: ''`. On macOS the result is that every `mod` combo registered through `useHotkeys` listens for Ctrl instead of Cmd, and every `<Kbd>` renders Ctrl.

Both surfaces are wrong *and consistent with each other*, so nothing looks broken — the shortcut simply never fires, and the hint tells you to press the key that also doesn't work.

The two call sites are independent copies of the same logic, and the `useHotkeys` docstring states it "Mirrors the detection used by Kbd so displayed and handled shortcuts agree", so they need the same change.

Suggested direction: treat a blank `platform` as *unknown* and fall through to the existing `navigator.platform` branch, instead of as *not Apple*. We have this running locally as a patch and are happy to open a PR — but since it touches `packages/core/` and changes behavior, tell us whether you want it through the specification protocol first.

### Reproduction

In any context where UA-CH is present but blank:

```js
Object.defineProperty(navigator, 'userAgentData', { value: { platform: '' } });
// isApplePlatform() === false, on macOS
```

End to end: run an Electron build on macOS with an overridden user agent, register `useHotkeys('mod+k', fn)` and render `<Kbd keys="mod+k" />`. Cmd+K does not fire and the Kbd shows Ctrl. The same code in Safari/Chrome on the same machine behaves correctly.

### Astryx Version

0.4.5 (verified against the published tarball; also present in 0.4.0)

### Environment

macOS, Electron / Chromium, any build that overrides the user-agent or client-hints identity.

Astro-Han:
Fixes #5253.

## The defect

`isApplePlatform()` in `packages/core/src/hooks/useHotkeys.ts` and `detectMac()` in `packages/core/src/Kbd/Kbd.tsx` are independent copies of the same detection: prefer `navigator.userAgentData.platform`, fall back to `navigator.platform`. The guard on that preference was

```ts
if (uaData && typeof uaData === 'object' && 'platform' in uaData) {
  return /mac/i.test((uaData as {platform: string}).platform ?? '');
}
return /Mac|iPhone|iPad|iPod/.test(navigator.platform ?? '');
```

`'platform' in uaData` is true whenever the key exists **at all, blank or not**, and `??` only defends against `null`/`undefined`. So a blank platform takes the client-hints branch, evaluates `/mac/i.test('')`, returns `false`, and the fallback that would have answered correctly is never reached. An empty string was read as a negative answer rather than as no answer.

Electron and other embedders that rewrite the app's user-agent or client-hints identity ship exactly that. On macOS every `mod` combo registered through `useHotkeys` listened for Ctrl instead of Cmd, and every `<Kbd>` drew Ctrl. Both surfaces agreed with each other, so nothing looked broken: the shortcut simply never fired, and the hint named the key that also did not work. The same code in Safari or Chrome on the same machine behaved correctly.

## The fix

A blank (or non-string) platform is treated as *unknown* and falls through. A client-hints platform that actually names something is still preferred, so nothing changes for browsers that report one.

```ts
const uaPlatform = (uaData as {platform?: unknown}).platform;
if (typeof uaPlatform === 'string' && uaPlatform.trim() !== '') {
  return /mac/i.test(uaPlatform);
}
```

Both call sites change in one commit: the `useHotkeys` docstring states the hook "Mirrors the detection used by Kbd so displayed and handled shortcuts agree", so fixing one alone would produce the failure this bug is least likely to be noticed in — a shortcut that fires on a key the hint does not name. The shared docstring line is updated in both, kept identical.

## Tests

One regression test per call site, written in the platform-spoofing style each file already uses (`vi.stubGlobal` in `useHotkeys.test.ts`, `Object.defineProperty` in `Kbd.test.tsx`): with `userAgentData.platform: ''` and `navigator.platform: 'MacIntel'`, `mod+k` fires on `metaKey` and not on `ctrlKey`, and `<Kbd keys="mod" />` renders ⌘. Both fail on `main` and pass with the fix.

`Kbd.test.tsx`'s `afterEach` also deletes the spoofed `userAgentData` so nothing leaks into the tests that follow.

## Validation

- `vitest run` on both test files — 31 passed; verified the two new Mac tests fail with the source change reverted
- `pnpm -F @astryxdesign/core typecheck`, `eslint` on the touched files, `pnpm check:repo` — all clean

A `[fix]` changeset is included (`@astryxdesign/core: patch`).

---

*Diagnosis, patch, and this description were prepared with Claude Code; every line was read and verified by me before opening, and the validation above was run locally.*

### DIFF
--- a/packages/core/src/hooks/useHotkeys.ts
+++ b/packages/core/src/hooks/useHotkeys.ts
@@ -73,8 +73,9 @@ const KEY_ALIASES: Record<string, string> = {
 
 /**
  * Detects whether the current platform is macOS/iOS.
- * Prefers the User-Agent Client Hints API when available (modern Chrome/Edge),
- * falls back to navigator.platform (deprecated but universally supported).
+ * Prefers the User-Agent Client Hints API when it names a platform (modern
+ * Chrome/Edge), falls back to navigator.platform (deprecated but universally
+ * supported) when it is absent or blank.
  * Mirrors the detection used by Kbd so displayed and handled shortcuts agree.
  */
 function isApplePlatform(): boolean {
@@ -83,7 +84,13 @@ function isApplePlatform(): boolean {
   }
   const uaData = 'userAgentData' in navigator ? navigator.userAgentData : null;
   if (uaData && typeof uaData === 'object' && 'platform' in uaData) {
-    return /mac/i.test((uaData as {platform: string}).platform ?? '');
+    const uaPlatform = (uaData as {platform?: unknown}).platform;
+    // A blank platform is no answer, not a negative one. Builds that rewrite
+    // their client-hints identity ship '', so fall through rather than
+    // reading it as "not Apple".
+    if (typeof uaPlatform === 'string' && uaPlatform.trim() !== '') {
+      return /mac/i.test(uaPlatform);
+    }
   }
   return /Mac|iPhone|iPad|iPod/.test(navigator.platform ?? '');
 }
--- a/.changeset/kbd-hotkeys-blank-client-hints-platform.md
+++ b/.changeset/kbd-hotkeys-blank-client-hints-platform.md
@@ -1,0 +1,16 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] `mod` hotkeys and `Kbd` resolve to Cmd on macOS again when client hints report a blank platform
+
+`useHotkeys` and `Kbd` both prefer `navigator.userAgentData.platform` and fall
+back to `navigator.platform`, but guarded the preference with
+`'platform' in uaData`, which is true whenever the key exists at all. A build
+reporting `platform: ''` therefore committed to the client-hints branch and got
+`false` without ever reaching the fallback, so on macOS every `mod` combo
+listened for Ctrl and every `<Kbd>` drew Ctrl. Electron and other embedders
+that rewrite the app's user-agent identity ship exactly that. A blank platform
+is now treated as unknown and falls through.
+
+@Astro-Han
--- a/packages/core/src/Kbd/Kbd.test.tsx
+++ b/packages/core/src/Kbd/Kbd.test.tsx
@@ -23,6 +23,7 @@ describe('Kbd', () => {
       value: originalPlatform,
       configurable: true,
     });
+    delete (navigator as {userAgentData?: unknown}).userAgentData;
   });
 
   it('renders a single key', () => {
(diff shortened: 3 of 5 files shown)


packages/core/src/hooks/useHotkeys.ts
```diff
@@ -83,7 +84,13 @@ function isApplePlatform(): boolean {
   }
   const uaData = 'userAgentData' in navigator ? navigator.userAgentData : null;
   if (uaData && typeof uaData === 'object' && 'platform' in uaData) {
-    return /mac/i.test((uaData as {platform: string}).platform ?? '');
+    const uaPlatform = (uaData as {platform?: unknown}).platform;
+    // A blank platform is no answer, not a negative one. Builds that rewrite
+    // their client-hints identity ship '', so fall through rather than
+    // reading it as "not Apple".
+    if (typeof uaPlatform === 'string' && uaPlatform.trim() !== '') {
```
cixzhang, line 91: Blank is handled — is a value that names nothing, like `Unknown`, worth the same?

cixzhang (approved):
Thanks, good diagnosis — and fixing both copies in one commit is right. Merging as is. One nit, not blocking: `detectMac` and `isApplePlatform` want to be one shared helper.

[Reviewed by Robohands]

packages/core/src/hooks/useHotkeys.ts
Astro-Han: #5394 has both.

`Unknown` is the spec's own value for "cannot say", so it names a platform no more than `''` does, and `/mac/i.test('Unknown')` is false either way. If anything it's the likelier of the two to show up: `''` isn't a legal value, so only a careless embedder emits it, and `Unknown` is what a careful one emits. Both fall through now.

The helper is `utils/isApplePlatform.ts`, imported by path from both. Not added to `utils/index.ts` on purpose, since the root barrel re-exports it and that would make it API. Kept `isApplePlatform` over `detectMac` because it matches iPhone/iPad/iPod too.

PR #3700, opened 2026-07-08
chore: upgrade pinned package manager to pnpm 11

bhamodi:
## Summary

Upgrades the pinned package manager from **pnpm 10** (`pnpm@10.34.1`) to **pnpm 11** (`pnpm@11.10.0`), per the evaluation in #3683. Migration guide: https://pnpm.io/migration

pnpm 11 changed several things this repo relied on, so this is more than a version bump. Each change below was surfaced by actually running pnpm 11 and is handled here.

## Changes

**Config migration (pnpm 11 no longer reads these from `package.json` / `.npmrc`):**
- `package.json`: `packageManager` → `pnpm@11.10.0`; removed the `pnpm` field.
- `pnpm-workspace.yaml`: added `overrides` (moved verbatim from `package.json`).
- `pnpm-workspace.yaml`: replaced `onlyBuiltDependencies` with pnpm 11's **`allowBuilds`** map (`esbuild`, `sharp`, `@swc/core`, `@parcel/watcher`, `core-js`, `puppeteer` → `true`). pnpm 11 ignores `onlyBuiltDependencies` and gates build scripts through `allowBuilds` / `pnpm approve-builds`.
- `pnpm-workspace.yaml`: moved hoisting/linker settings (`nodeLinker: hoisted`, `shamefullyHoist: true`, `linkWorkspacePackages: true`) from `.npmrc` — pnpm 11 doesn't apply them from `.npmrc`, which was silently breaking the flat node_modules layout (see below).
- Docs: pnpm version references in `README.md`, `CONTRIBUTING.md`, `CLAUDE.md`, `docs/release.md`.

**Lockfile regenerated under pnpm 11.** Its stricter `minimumReleaseAge` enforcement (#3669) now runs on *every* install and rejected **14 too-new entries** that had slipped into the pnpm 10 lockfile (e.g. `@typescript-eslint/*` 8.63.0, `caniuse-lite`, `picomatch` 4.0.5). A fresh resolve re-pinned them to their latest ≥7-day-old versions — hence the large lockfile diff with some **dependency downgrades**. This is the supply-chain policy working as intended.

CI needs no workflow edits (`pnpm/action-setup@v6` reads `packageManager`). Node stays 22+/24, satisfying pnpm 11's Node 22+ requirement.

## Why `shamefully-hoist` had to move (and a phantom-dep note)

[...]

PR #5327, opened 2026-08-22
chore(deps): make node_modules strict, declare what each package imports

### LINKED ISSUE #5328: [Infra] shamefullyHoist hides undeclared imports; nodeLinker: hoisted is redundant
### Problem

`pnpm-workspace.yaml` carries two hoisting settings inherited verbatim from the pre-pnpm-11 `.npmrc` (#3700):

```yaml
nodeLinker: hoisted
shamefullyHoist: true
```

`shamefullyHoist: true` is `publicHoistPattern: '*'` — every transitive package is republished at the root of `node_modules`. A workspace package can therefore import a module it never declared in its own `package.json`, and nothing complains: not install, not build, not CI. The declaration is simply missing, and the import rides on someone else's copy.

`nodeLinker: hoisted` is a separate problem. Its `.npmrc` annotation read *"Use the Node-modules linker (default), not pnpm's PnP"*, which is wrong on its own terms — pnpm's default **is** the node-modules linker, but its default mode is `isolated`, not `hoisted`. On top of being unnecessary, the pair is a combination pnpm has an open bug on: `nodeLinker: hoisted` with `shamefullyHoist: true` is [pnpm/pnpm#6620](https://github.com/pnpm/pnpm/issues/6620), spurious `ModuleNotFoundError`s that only clear after toggling the setting off and back on.

### Evidence

Flipped `shamefullyHoist` to `false` locally (pnpm 11.10.0), recreated `node_modules`, and ran build / test / typecheck / Storybook / Next builds. Eight undeclared imports surfaced — each verified by checking the missing module against the importing package's own `package.json`:

| Package | Undeclared module | Symptom |
| --- | --- | --- |
| root `scripts/` | `@babel/core`, `glob`, `chokidar`, `@testing-library/jest-dom`, `@stylexjs/babel-plugin` | `pnpm build` fails at `@astryxdesign/core` |
| `lab`, `charts`, `richtext` | `@babel/cli`, `@babel/core`, `@babel/preset-react`, `@babel/preset-typescript`, `@stylexjs/babel-plugin` | `build:esm` → `sh: babel: command not found` |
| `lab`, `charts`, `richtext` | `@testing-library/react` | 27 test files load **0 tests** |
| `core` | `@types/babel__core` | TS7016 |
| `storybook` | `@heroicons/react`, `@stylexjs/stylex`, `@lexical/react`, `lexical` | `storybook:build` fails to resolve |
| `docsite` | `@babel/parser` | `next build` type error after a clean compile |
| `cli` | `@heroicons/react`, `@stylexjs/stylex`, `lucide-react`, `recharts` | 145+ TS errors in `assets/templates/**`, which `tsconfig.strict.json` includes and CI runs (`ci.yml`) |

All eight are devDependencies, so no published package is missing a runtime dependency today — the cost is in-repo hygiene and the absence of a guard against the next one.

### Scope Check

- `linkWorkspacePackages: true` was audited and **must stay**: 64 internal deps are declared as bare `"*"` (only 4 use `workspace:*`), and pnpm ≥ 10 defaults this to `false`, which would resolve them from the registry instead of local source — silently, with no error.
- Three failures found during the sweep are pre-existing and unrelated, each baselined with hoisting restored: 7 test files (`plainDate`, `Calendar`, `DateInputTouch`, `Schedule`, `TransferListSelector`, 2 CLI suites — 69 failing cases on main vs 56 with the change), `lab typecheck` (relative `.doc.mjs` imports plus a `Drawer` `onClose` prop error; CI runs `typecheck:docs`, which passes), and the `template-viewer` build (`packages/build` aliases the `@astryxdesign/core` prefix to `core/src`, so `core/locales/*.json` rewrites to a nonexistent path; CI does not build this app).
- No symlink-hostile environment exists in the repo: no Dockerfile, no react-native, all CI runners are Ubuntu — so nothing depends on the flat layout for platform reasons.

### Proposed Direction

Declare what each package actually imports, then drop both settings in the same change — splitting them would land dependency declarations whose correctness cannot be demonstrated while the flag is still on.

Deliberately left alone: `linkWorkspacePackages`, and any move back toward `nodeLinker: hoisted`.

Han5991:
Closes #5328

## What

Removes both hoisting settings from `pnpm-workspace.yaml` and declares the dependencies that were only resolving because of them. `linkWorkspacePackages: true` stays.

Rebased onto current `main`. That brought in code which relies on the hoisted layout, so a second commit fixes it — see [After the rebase](#after-the-rebase-onto-current-main).

## Why

Both lines came over verbatim from the pre-pnpm-11 `.npmrc` (#3700).

**`nodeLinker: hoisted`** was annotated there as *"Use the Node-modules linker (default), not pnpm's PnP."* That is wrong on its own terms — pnpm's default **is** the node-modules linker, but its default mode is `isolated`, not `hoisted`. It also paired badly: `nodeLinker: hoisted` together with `shamefullyHoist: true` is [pnpm/pnpm#6620](https://github.com/pnpm/pnpm/issues/6620), spurious `ModuleNotFoundError`s that only clear after toggling the setting off and on.

**`shamefullyHoist: true`** is `publicHoistPattern: '*'` — it republishes every transitive package at the root of `node_modules`, so a package can import a module it never declared and nothing complains.

## Benefits

- **Undeclared imports become errors, permanently.** Nine of them were live in this repo (below). After this, CI catches the next one instead of it riding on a hoisted copy.
- **Dedup is restored.** `isolated` hardlinks each package once into the `.pnpm` store instead of writing a flat npm-style tree.
- **Version conflicts resolve correctly.** Each importer gets its own exact version, instead of [one winner hoisted to the workspace root](https://github.com/orgs/pnpm/discussions/5779) when two projects disagree.

## What was hiding behind the flag

| Package | Undeclared | Symptom |
| --- | --- | --- |
| root scripts | `@babel/core`, `glob`, `chokidar`, `@testing-library/jest-dom`, `@stylexjs/babel-plugin` | `pnpm build` fails at `@astryxdesign/core` |
| `lab`, `charts`, `richtext` | `@babel/cli`, `@babel/core`, `@babel/preset-react`, `@babel/preset-typescript`, `@stylexjs/babel-plugin` | `build:esm` → `babel: command not found` |
| `lab`, `charts`, `richtext` | `@testing-library/react` | 27 test files load **0 tests** |
| `core` | `@types/babel__core` | TS7016 |
| `core` | `@heroicons/react` | `storybook typecheck` TS2307; `Button.a11y.test.tsx` fails to resolve |
| `storybook` | `@heroicons/react`, `@stylexjs/stylex`, `@lexical/react`, `lexical` | build fails to resolve |
| `docsite` | `@babel/parser` | type error after a clean compile |
| `cli` | `@heroicons/react`, `@stylexjs/stylex`, `lucide-react`, `recharts` | `assets/templates/**` is in `tsconfig.strict.json`, which CI runs |

All are **devDependencies** — nothing changes for consumers of the published packages. Every version matches what the workspace already resolved, so the lockfile gains importer entries and **zero** new package versions.

## After the rebase onto current main

The branch was green before the rebase. Three jobs then failed, all on code that landed on `main` in the meantime and quietly relied on the flat root `node_modules`.

**`@heroicons/react` in `packages/core`** — #6131 added `Button/__tests__/Button.a11y.renders.tsx`, which imports it. `packages/core` never declared it. Broke `build-storybook` (TS2307) and `test-ui` (Vite resolve). Fixed by declaring it, the ninth row above.

**`build.packed-old-core.test.mjs`** — #5543 added a test that stages the packed CLI's runtime dependencies by reading `<repo root>/node_modules/{commander,jiti,jscodeshift,zod}`. Only a hoisted layout puts a package's own dependencies at the workspace root. It now looks from `packages/cli`.

**`build.adaptation-core-compat.test.mjs`** — also #5543. `core-interception`'s `patchCommonJs` requires `@astryxdesign/core` from the theme file to check whether it can wrap `defineTheme` for `.cjs` dependencies. `require` folds `NODE_PATH` in, the isolated layout exposes every package under `node_modules/.pnpm/node_modules`, and Vitest points `NODE_PATH` at exactly that directory. So a core no dependency of the theme could reach answered the lookup. Wrapping it fails on a `require(esm)` namespace, and the coverage gap that failure reported then rejected every theme whose lineage was unobserved.

This is the same trap the doctor fix in this PR already addresses, so the resolution is shared rather than duplicated: `findInstalledPackage` moves from `api/doctor/doctor.mjs` to `foundation/fs/paths.mjs`, and `doctor`, `core-interception`, and the packed-CLI test all use it. It walks only the caller's own `node_modules` chain and never answers from the ambient environment.

Second commit: 68 insertions, 28 deletions across 7 files — 27 of those lines are the helper moving between files.

### Rebased again, onto `ae9ccef`

Same class, three more instances from code that landed since:

- **`@types/babel__core` in `packages/cli`** — #6246 added `authoring/shadcn/source-variants.mjs`, which imports `@babel/core`. `cli` declared the runtime package but not its types, so `sync:api-types` failed with TS7016 and took `pnpm build` down with it.
- **`@heroicons/react` and `recharts` in `generate-playground-types.mjs`** — it read both off the repo root. Recharts arrived with #6264 and is covered by `playground-scope.test.ts`, which is how it surfaced; the heroicon path has been there since #2535, nothing asserts on it, and it was quietly emitting zero variants on this branch.
- **`node_modules/.bin/shadcn`** — #6246 added a fifth call site in `generate-shadcn-registry.test.mjs`.

The walk-up helpers the earlier rounds introduced are gone along with them. Node already walks up from the module doing the asking, so `resolveFromDocsite('<pkg>/package.json')` replaces the hand-rolled walk, and a missing dependency now throws instead of emitting a bundle that is quietly short a package. `shadcn` does not export `./package.json`, but its bin **is** its entry, so it resolves as `shadcn` and runs through `process.execPath` — no path assumption, no `NODE_PATH`.

That leaves no walk-up helper in the repo outside `findInstalledPackage`, which stays: it reads a *consumer's* project, where there is no module of ours to resolve from and `NODE_PATH` must not count as installed. Net for the two files against `main`: −8 lines in the generator, +14 in the test.

## Risks, and how each was closed

| Risk | Resolution |
| --- | --- |
| Build tools that walk `node_modules` break — the original stated reason for the flag, naming Storybook and the StyleX babel plugin | `pnpm build`, `storybook:build`, `docsite build` and `sandbox build` all pass |
| A CI typecheck gate breaks | Every typecheck CI runs passes, including `cli typecheck:strict` — the one that compiles the 145-error template assets |
| Tests regress | Full `node` and `ui` vitest projects pass; `ui` is 10190/10190 |
| Symlink-hostile environment (Docker, serverless, React Native) | None in this repo; all CI runners are Ubuntu, no Dockerfile, no react-native |
| Workspace packages silently resolving from the registry | `linkWorkspacePackages: true` untouched — required, since 64 internal deps are bare `"*"` (only 4 use `workspace:*`) and pnpm ≥ 10 defaults it to `false` |

Every CI job passed on the previous head. On this one, verified locally: `install --frozen-lockfile`, `build`, `lint`, all ten typecheck gates, `docsite generate` + `docsite test`, the registry contract scripts, and both vitest projects.

## Pre-existing, left untouched

- **`lab typecheck`** — relative `.doc.mjs` imports plus a `Drawer` `onClose` prop error. Unrelated to module resolution; CI runs `lab typecheck:docs`, which passes.
- **`template-viewer` build** — fails identically with hoisting restored. `packages/build` aliases the `@astryxdesign/core` prefix to `core/src`, so `core/locales/*.json` rewrites to a path that does not exist. CI does not build this app.
- **`internal/vibe-tests/setup-test/setup-workspace.test.mjs`** — fails on macOS only, because it shells out to `cp -a --reflink=auto`, which BSD `cp` rejects. The file is byte-identical to `main` and the test passes on the Linux runners.

## Note for reviewers pulling this branch

Changing hoisting makes pnpm recreate `node_modules` wholesale. Interactively it prompts; in a non-TTY shell it aborts with `ERR_PNPM_ABORTED_REMOVE_MODULES_DIR_NO_TTY` unless `CI=true` is set. GitHub Actions sets `CI` itself, so CI is unaffected.


cixzhang (changes requested):
Thanks—making the dependency graph explicit is the right direction. Right now the head cannot pass CI, and the shipped CLI fix would miss the next release:

1. `internal/vibe-tests/setup-test/public-artifact-cli.test.mjs` still hard-codes the root `node_modules/.bin/tsx`, but the root manifest does not declare `tsx`. Exact-head `test-node` now fails because both subprocesses never start. Please make that launcher resolve the package-local declared binary (or declare the root dependency) and get the Node suite green.
2. This also changes the published CLI's `doctor` peer-resolution behavior, but no Changeset covers `@astryxdesign/cli`. Please add the required patch Changeset so users receive the fix.

[Reviewed by Robohands]

Han5991:
@cixzhang 

Both addressed.

1. **`test-node`** — fixed in 6da1de70c77; the review landed against 4e6d7c75740, one commit earlier. Took your first option: the launcher now resolves the bin of `internal/vibe-tests`, the package that declares `tsx`. `test-node` is green on that head.

2. **Changeset** — added in 115b5294d7b: `.changeset/doctor-peer-resolution.md`, patch on `@astryxdesign/cli`, `[fix]`. `pnpm check:changesets` passes.

One heads-up: the branch is rebased onto current `main`, and the rebase surfaced the same class of undeclared import in code that landed meanwhile — `packages/vega`, `internal/eslint-plugin-astryx`, and `yaml` at the root.

cixzhang (changes requested):
Thanks—both earlier asks are fixed: the package-local `tsx` launcher now matches its declaration, and the CLI behavior has a patch Changeset.

One compatibility gap remains. Astryx documents Yarn installs and the CLI explicitly supports Yarn, but modern Yarn uses PnP by default and has no `node_modules`. The new filesystem-only lookup at `packages/cli/api/doctor/doctor.mjs:112` therefore tells a healthy Yarn PnP project that every Core peer is missing. Please keep resolution project-local without accepting ambient `NODE_PATH`, and add regression coverage for both cases.

[Reviewed by Robohands]

Han5991:
@cixzhang 

I was heading to Discord with this, but it turns out to be about this PR's diff specifically, so it seems more useful here.

I couldn't reproduce the PnP failure as described, so I built a real one to check against. Yarn 4.18, PnP — no `node_modules`, `.pnp.cjs` and `.pnp.loader.mjs` on disk — with `@astryxdesign/core@0.5.2` genuinely installed (`yarn why` confirms it). Running `runChecks()` against that project from each checkout gives **byte-identical output on `main` and on this branch**:

```
[FAIL] core-installed:  @astryxdesign/core could not be resolved from this project.
[WARN] themes:          No @astryxdesign/theme-* packages are installed.
[INFO] peer-deps:       Skipped — @astryxdesign/core is not installed.
[INFO] package-manager: Detected package manager: yarn (declared in package.json).
```

Two things follow from that.

**`peer-deps` reports "Skipped", not missing peers.** `checkPeerDeps` returns on its first line because `ctx.coreDir` is null, so `doctor.mjs:112` — the lookup this PR adds — never executes under PnP. It has no way to report every peer as missing there.

**What does break predates this PR.** `findCoreDir` in `packages/cli/foundation/fs/paths.mjs` and `findNodeModules` in `doctor.mjs` both answer with `fs.existsSync` on a `node_modules` path, and `resolvePackageDir` in `foundation/integrations/integrations.mjs` carries the same assumption. This PR touches none of them — `git diff main...HEAD -- packages/cli/foundation/` is empty.

So doctor appears to have never worked under Yarn PnP. My guess at why it went unnoticed: Yarn classic and Berry with `nodeLinker: node-modules` both have a real `node_modules`, and a diagnostic that answers wrongly fails quietly rather than breaking a build. There's also no PnP coverage anywhere in the repo — no `.pnp.cjs` or `pnpapi` reference in any file.

I'd rather not fix it in this PR. Making resolution PnP-aware means first deciding how the CLI reaches the project's PnP map at all: `pnpapi` only exists in-process when the CLI is launched through Yarn, so an `npx astryx` invocation would have to read `.pnp.cjs` directly or shell out to Yarn. That's a design change to CLI resolution, which is a different thing from the dependency declarations this PR is making.

Happy to open a separate issue with this repro attached. If you'd rather it land here instead, say so and I'll take it on in this branch.

Han5991:
Rebased onto current `main`, and every CI job passes on the head commit.

The rebase surfaced three places where code that landed since relied on the flat root `node_modules`: an undeclared `@heroicons/react` in `packages/core`, a test staging the packed CLI's runtime deps from the repo root, and `core-interception` reaching a core through `NODE_PATH`. The second commit fixes all three, sharing the `node_modules` walk the doctor fix already introduced. Details are in the PR description.

Ready for review whenever someone has a moment.

Han5991:
Rebased onto `ae9ccef`. Only `pnpm-lock.yaml` conflicted — `main` had picked up `@babel/{parser,traverse}@8` from the bumps this PR unifies back to 7.x — so it was regenerated rather than hand-merged.

Three things that landed since relied on the hoisted root: `packages/cli` imports `@babel/core` without declaring `@types/babel__core` (#6246), `generate-playground-types.mjs` reads `recharts` (#6264) and `@heroicons/react` off the repo root, and `generate-shadcn-registry.test.mjs` gained a fifth `.bin/shadcn` call site (#6246). The first broke `pnpm build`; the second failed silently, which is why the heroicon path had been emitting zero variants since #2535 unnoticed.

While fixing them I dropped the two walk-up helpers earlier rounds added — they reimplemented Node's own resolution. Anchoring a `createRequire` at the declaring package does the same in one line and throws instead of returning empty. `findInstalledPackage` stays, since it reads a consumer's project where there is no module of ours to anchor to.

Verified locally on this head: `install --frozen-lockfile`, `build`, `lint`, all ten typecheck gates, docsite (498), the registry contract scripts, `ui` 10190/10190, `node` 6786 passing. `node`'s two remaining failures reproduce identically on `ae9ccef` with hoisting in place — `setup-workspace` shells out to `cp --reflink` (BSD `cp` rejects it, macOS only) and `story-tree` has the same stray titles on `main`.

cixzhang (approved):
Thanks! This substantially reduces dependency-layout risk. The earlier review asks are addressed; Yarn PnP remains follow-up debt rather than a blocker.

PR #5329, opened 2026-08-22
docs: replace em dashes in MediaTheme and cli-integrations prose

josephfarina:
## What

MediaTheme's `mode` and `fallback` prop descriptions, its best practices, and the cli-integrations reference doc used em dashes for asides and clause breaks. Recast each with the right punctuation: parentheses for parenthetical asides, colons where a clause introduces a list or an explanation, and semicolons between independent clauses.

Prose strings only. No technical claim changed; the identifiers, ranges, and code spans are untouched. Both files parse and render clean (`astryx component MediaTheme --lang dense`, `astryx docs cli-integrations`).

## Checks

Night Watch — Doc Reviewer

### DIFF
--- a/packages/cli/assets/docs/cli-integrations.doc.mjs
+++ b/packages/cli/assets/docs/cli-integrations.doc.mjs
@@ -115,7 +115,7 @@ export const docs = {
       content: [
         {
           type: 'prose',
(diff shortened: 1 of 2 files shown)


cixzhang (approved):
Thanks — checked each recast against the original, no technical claim moved.

[Reviewed by Robohands]

PR #5330, opened 2026-08-22
docs(Avatar): restore the missing dense best-practice bullet

josephfarina:
## What

The English Avatar docs list eight best practices; `docsDense` had only seven. The "do not name an interactive avatar with a status label" anti-pattern (guidance:false) had been folded onto the end of the interactive-avatar bullet (guidance:true), which dropped its own position and its guidance flag. The dense compression protocol requires 1:1 count parity with English, matching the guidance flag at each position, and never packing two English bullets into one.

Split the anti-pattern back into its own guidance:false bullet. Dense now matches English 8/8 with the same guidance sequence (true x5, false x3).

## Checks
- `astryx component Avatar --lang dense` renders eight best practices, five Do and three Don't

Night Watch — Doc Reviewer


cixzhang (approved):
Thanks — the anti-pattern reads better as its own don't than folded onto the do.

[Reviewed by Robohands]

PR #5331, opened 2026-08-22
docs(cli): document the integration manifest docs field in the README

josephfarina:
## What

The integration manifest schema (`configSchema`/`integrationSchema` in `packages/cli/authoring/integration/parse.mjs`) accepts an optional `docs` field, a directory of reference docs that the CLI serves as topics. The Integrations field table in the CLI README listed `components`, `templates`, `codemods`, and `issuesUrl` but not `docs`, so the documented manifest surface no longer matched what the CLI accepts.

Add the `docs` row, positioned to match the schema order (between `codemods` and `issuesUrl`). Names-in-sync only; no descriptions rewritten. After this, the config/README drift check reports OK for both the Configuration and Integrations tables.

## Checks
- Config/README drift audit reports OK for `astryx.config` and `astryx.integration`
- Merges cleanly with the other open README-touching PRs (#5260, #5226, #5138, #5115); disjoint sections

Night Watch — Doc Reviewer

### DIFF
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -580,13 +580,13 @@ export default {
 };
 ```
 
-| Field        | Type     | Purpose                                                               |
-| ------------ | -------- | --------------------------------------------------------------------- |
-| `components` | `string` | Directory holding the package's components and their `.doc.*` files.  |
-| `templates`  | `string` | Directory holding the package's page/block templates.                 |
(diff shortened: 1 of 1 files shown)


cixzhang (approved):
Thanks — the row is right, the generator just re-pads the columns. `pnpm -F @astryxdesign/cli readme` and commit; that's the red test.

[Reviewed by Robohands]

PR #5335, opened 2026-08-22
fix(a11y): one disabled cursor everywhere — default, not not-allowed

cixzhang:
## Why

Follow-up to [#5323](https://github.com/facebook/astryx/pull/5323), which guaranteed `not-allowed` on a disabled element. It cannot actually be a guarantee.

**75 of the 635 disabled elements in the story set are sealed behind `pointer-events: none`.** Those are never hit-tested, so they show whatever their ancestor shows — no declaration on the control can change that. So the library was painting the stronger cursor on the reachable ones and a plain arrow on the rest: exactly the inconsistency the work was meant to remove, just relocated.

`default` is paintable everywhere, matches the internal XDS convention, and costs nothing we were relying on — a disabled control already reads as disabled from its own visual treatment; the cursor's job here is to stop *promising* an interaction, not to explain one.

The mechanism from #5323 is unchanged and still doing the real work: the guard on every declaration, the merge-order trap, the lint rule, the sweep. Only the value moves.

## What

- Every guarded `cursor` in core and lab resolves to `default` when disabled — 126 sites.
- The 49 pre-existing flat `cursor: 'not-allowed'` declarations inside `disabled` styles become `default` too. Leaving those would have kept two conventions in the library, which is the thing being fixed.
- A declaration whose base was already `default` collapses back to a flat `cursor: 'default'` — the guard is redundant there (Lightbox, SideNavHeading, SideNavSection, TopNavHeading).
- `reset.css`, the lint rule, the sweep, `reset.test.ts` and the docs all expect `default`. The rule now exempts `default` instead of `not-allowed`.
- #5323's changeset is amended rather than answered with a second one — it has not shipped, so the changelog tells one story instead of contradicting itself.

## Risk

Same blast radius as #5323 and no larger: only what a disabled element shows under the pointer. Nothing about markup, props, events or the sealed elements changes.

## Testing

- `pnpm guard:disabled-cursor` over all 1704 stories: **0 violations**, 635 disabled elements checked, 75 sealed (reported, not passed).
- `@astryx/disabled-cursor` over core and lab: **0 findings**.
- Unit: rule `RuleTester` suite, sweep verdict/CLI suite, `reset.test.ts` — 27 passing.
- Rubric [A21](https://github.com/facebook/astryx/wiki/Component-Audit-Rubric) updated to expect `default` (v1.7.1, patch — nothing was audited under 1.7, so no score is invalidated).

### DIFF
--- a/.changeset/disabled-cursor-guarantee.md
+++ b/.changeset/disabled-cursor-guarantee.md
@@ -2,6 +2,6 @@
 '@astryxdesign/core': patch
 ---
 
-[fix] A disabled control now answers the pointer with `not-allowed` wherever it can be pointed at: every `cursor` in core and lab carries `':is(:disabled,[aria-disabled="true"])': 'not-allowed'`, and the reset gives the same cursor to any disabled element that declares none — `[aria-disabled]` included, where it previously said `default` and only for native `:disabled`. A lint rule and a Chromium sweep over every story keep it that way. Disabled elements sealed behind `pointer-events: none` are unchanged: the pointer never reaches them, so their cursor comes from an ancestor.
+[fix] A disabled element answers the pointer with `default`, never an interactive cursor. Every `cursor` in core and lab carries `':is(:disabled,[aria-disabled="true"])': 'default'`, and the reset gives the same cursor to any disabled element that declares none — `[aria-disabled]` included, which previously got nothing. A lint rule and a Chromium sweep over every story keep it that way. Disabled elements sealed behind `pointer-events: none` are unchanged: the pointer never reaches them, so their cursor comes from an ancestor — which is why the guarantee is `default` rather than a distinct disabled cursor the library could only paint on some of them.
 
 @cixzhang
--- a/.github/scripts/disabled-cursor-audit.js
+++ b/.github/scripts/disabled-cursor-audit.js
@@ -3,11 +3,11 @@
 
 
 /**
- * @description Asserts every disabled element answers the pointer with `not-allowed`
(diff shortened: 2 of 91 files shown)


cixzhang:
**Self-review.** The value change is mechanical; what I checked is that it did not quietly weaken the mechanism.

- **Still one atomic class.** `astryx.css` carries a single `:is(:disabled,[aria-disabled="true"]){cursor:default}` rule for all 126 sites, and zero `not-allowed` anywhere in the built stylesheet — so nothing kept the old value by being missed.
- **The guard still has to win.** `default` is now both the base value on some elements and the disabled value everywhere, which is exactly the shape that can hide a broken guard — a declaration would look right while doing nothing. Verified from the other end instead: the Chromium sweep hit-tests and reports **0 violations across 1704 stories**, and where the base was already `default` the guard is collapsed away rather than left as decoration.
- **The rule's exemption moved with it** (`default` in, `not-allowed` out), so a stray `cursor: 'not-allowed'` would now be flagged rather than waved through as "already handled".

One thing I decided rather than asked: **the 49 pre-existing flat `cursor: 'not-allowed'` declarations in `disabled` styles are converted too.** They predate both PRs. Leaving them would have left the library with two disabled cursors, which is the inconsistency this is fixing — but it does mean the diff is larger than "revert #5323's value", and every one of those is a visible change to a component that was already doing something deliberate. Worth a look if that reads as overreach.

The sealed-element gap is unchanged and is now the honest reason for the value: 75 of 635 disabled elements can never answer at all.

PR #5342, opened 2026-08-22
fix(BottomSheet): colour the iOS Safari toolbar strip under a non-modal sheet

imdreamrunner:
## The bug

On a page that has any fixed or sticky element at its bottom edge, opening a
`hasScrim={false}` BottomSheet on iOS 26 leaves **the page** colouring the strip
behind Safari's address bar, while the sheet covers the screen above it. The
page appears to show through under the sheet. A modal sheet is unaffected.

## Why

iOS 26 dropped `<meta name="theme-color">`. Safari now derives what it paints
behind its translucent toolbars by sampling the page: it hit tests a point 4px
inside each viewport edge, walks up to the nearest `fixed`/`sticky` ancestor,
and extends that element's declared `background-color` into the chrome
(`LocalFrameView::fixedContainerEdges`, `Page::updateFixedContainerEdges`).

A modal sheet is already served by this — WebKit has a dedicated branch for a
dialog's `::backdrop`. A non-modal sheet is not:

- the panel and its positioner are `absolute`, so the walk continues past them;
- the next fixed ancestor is the sheet's own `<dialog>`, which is transparent
  and `100dvw × 100dvh`, so WebKit classifies it `IsViewportSizedCandidate`;
- for a viewport-sized candidate WebKit sets `preferExistingColor` and, if the
  edge already has a colour, **keeps it** — the host page's.

Hence the precondition: the page must have latched its own bottom-edge colour
before the sheet opens. Open the sheet on mount and the bug hides, which is why
it does not reproduce in a minimal repro.

## The fix

Render one dedicated sampling target inside the non-modal dialog: `fixed`, full
width, flush with `insetBlockEnd: 0`, `12px` tall, `--color-background-surface`.
Being 12px it is not viewport-sized, so it is an ordinary `IsCandidate` and its
declared colour wins outright — no dependence on what the page did first.

Three constants are load-bearing and all come from WebKit, not from here:

| | why |
|---|---|
| height > 10px | below `thinBorderWidth` WebKit ignores the declared colour and samples painted pixels instead, which a masked box has none of |
| `mask-image`, not `visibility`/`opacity`/`display` | the sampler rejects all three of those; it does not look at masks. This is what makes the strip readable by Safari and invisible to the user |
| full width | under 90% of the viewport on that side the box is rejected as too small |

`BottomSheetEdgeTint.test.tsx` pins each of them, on the style definition — none
of it is observable in jsdom, or in any engine without retractable chrome.

## Measured

iPhone 17 Pro simulator, iOS 26.5, same page and same build except for this
change, screenshot pixels sampled at x = 6% of the width. Page has a magenta
`#c800ff` fixed bottom bar, fully covered by a `height="hug"` non-modal sheet;
the sheet's surface is `#ffffff` light / `#1f1f22` dark.

| | strip at 776–874pt (behind the address bar) |
|---|---|
| before | `#c700fe` — the page's bar |
| after | `#ffffff` — the sheet |
| before, dark | `#c800ff` |
| after, dark | `#1f1f22` |

A full-frame diff of the two runs differs **only** in that strip (and in the
status-bar clock, which ticked between shots) — the element itself paints
nothing.

Modal sheets: byte-identical before and after.

## Test plan

- `pnpm lint:strict` — 0 errors (55 pre-existing warnings)
- `pnpm test` — the 5 new suites pass; the full run flakes on a *different*
  handful of unrelated timeouts each run (DateInputTouch / TransferListSelector
  / TabList), all of which pass in isolation on this branch
- `pnpm build`, `pnpm -F @astryxdesign/storybook build`, and every `typecheck`
  variant CI runs
- the simulator measurements above

## Caveats a reviewer should weigh

**This depends on an undocumented heuristic.** Apple removed the supported
mechanism (`theme-color`) and shipped no replacement API — WebKit's own position
on bug 301756 is that the colour extension exists to serve fixed headers, not to
be driven. So every constant this relies on (the 10px floor, the 90% width
ratio, the fact that `mask` is not consulted while `visibility`/`opacity` are)
is an implementation detail read out of `LocalFrameView.cpp`, not a contract.
There is no cleaner alternative available today; that is the trade being made,
not an expedient choice over a better one.

**Failure mode if Apple changes the rules**: the strip reverts to today's
behaviour — the page's colour behind the address bar. The element paints
nothing, so a broken heuristic cannot produce a visible artifact, a layout
shift, or an interaction bug. It degrades to the status quo.

**Not verified**:
- A physical device. Measured on the iOS 26.5 simulator, which is real WebKit
  and real retractable chrome, but not a device in hand.
- iPad, landscape, and the non-default Safari toolbar layouts (Top / Bottom;
  Compact was tested).
- Behaviour with *Settings > Apps > Safari > Tabs > Allow Website Tinting*
  turned off. Expected to be a no-op, untested.

**Known, deliberately out of scope**: the modal path leaks the same way on a
page with a fixed bottom edge — the scrim dims it, so it reads as intentional
rather than as a bug. Left alone here because it is the default path and WebKit
actively maintains the `::backdrop` branch that governs it; worth revisiting
separately.

**Maintenance risk**: the element renders nothing, so it looks deletable. The
reasoning is in the file header and `BottomSheetEdgeTint.test.tsx` fails if the
sampled declarations are removed or weakened — but a future reader still has to
read that before touching it.

### DIFF
--- a/.changeset/bottom-sheet-noscrim-edge-tint.md
+++ b/.changeset/bottom-sheet-noscrim-edge-tint.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] BottomSheet: a non-modal sheet now colours the iOS 26 Safari toolbar strip with its own surface instead of letting the page show through behind the address bar.
+
+@imdreamrunner
--- a/packages/core/src/BottomSheet/BottomSheet.tsx
+++ b/packages/core/src/BottomSheet/BottomSheet.tsx
@@ -16,6 +16,7 @@
  *
  * SYNC: When modified, update these files to stay in sync:
  * - /packages/core/src/BottomSheet/BottomSheetPanel.tsx
+ * - /packages/core/src/BottomSheet/BottomSheetEdgeTint.tsx
  * - /packages/core/src/BottomSheet/BottomSheet.doc.mjs
  * - /packages/core/src/BottomSheet/BottomSheet.test.tsx
  * - /packages/core/src/BottomSheet/BottomSheetSwitcher.tsx
@@ -43,6 +44,7 @@ import {
   type BottomSheetPanelMotion,
   type BottomSheetPanelState,
 } from './BottomSheetPanel';
+import {BottomSheetEdgeTint} from './BottomSheetEdgeTint';
 import {
   BottomSheetSwitcherContext,
   type BottomSheetSwitcherContextValue,
@@ -355,6 +357,8 @@ function StandaloneBottomSheet({
           {children}
         </BottomSheetPanel>
       </div>
+      {/* A modal sheet's ::backdrop already answers Safari's edge sampler. */}
+      {hasScrim ? null : <BottomSheetEdgeTint />}
     </dialog>
   );
 }
--- a/packages/core/src/BottomSheet/BottomSheetEdgeTint.test.tsx
+++ b/packages/core/src/BottomSheet/BottomSheetEdgeTint.test.tsx
@@ -1,0 +1,225 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+/**
+ * @file BottomSheetEdgeTint.test.tsx
+ * @input Uses vitest, @testing-library/react, BottomSheet, BottomSheetSwitcher
+ * @output Tests which sheets carry the iOS Safari bottom edge tint, and pins
+ *   the declarations WebKit's edge sampler reads
+ * @position Core testing; validates BottomSheetEdgeTint.tsx and its two hosts
+ *
+ * SYNC: When modified, update these files to stay in sync:
+ * - /packages/core/src/BottomSheet/BottomSheetEdgeTint.tsx
+ */
+
+import {describe, it, expect, vi, beforeEach, afterEach} from 'vitest';
+import {render} from '@testing-library/react';
+import {BottomSheet} from './BottomSheet';
+import {BottomSheetSwitcher} from './BottomSheetSwitcher';
+
+// jsdom doesn't implement <dialog> open/close or pointer capture; stub them.
+beforeEach(() => {
+  HTMLDialogElement.prototype.showModal = vi.fn(function (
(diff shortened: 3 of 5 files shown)


imdreamrunner:
**Independent verification, plus a corroborating data point from #5295**

Confirmed on the docs site in the iOS 26.5 simulator (iPhone 17), page background forced red, sampling the strip behind the toolbar full-width across the bottom band:

| build | page colour visible in the strip |
| --- | --- |
| before this PR | 51.6% — 3/3 runs |
| this PR | 0% — 3/3 runs |

**A second, independent line of evidence for the diagnosis.** While chasing this in #5295 I ran a control that turns out to confirm the sampling model exactly, though I misread it at the time: a plain `position: fixed; inset-inline: 0; bottom: 0; height: 300px; background: #fff` panel on a red page **coloured the toolbar strip correctly**, while our sheet on the same page did not. I read that as "our structure clips something" and went looking for geometry. Under the mechanism described here it is precisely as predicted — the plain panel is a valid sample target (fixed, declared background, *not* viewport-sized), whereas the sheet's nearest fixed ancestor is the transparent full-viewport `<dialog>`, which WebKit answers by keeping the colour it already had. Every geometry fix I tried failed because the problem was never geometry.

**The rule generalises past `hasScrim`.** Lab's `Drawer` also goes non-modal via `show()` and is *not* affected, because its `<dialog>` **is** the panel and carries `backgroundColor: --color-background-surface`. So the exposure isn't non-modality as such — it's *an overlay whose outermost fixed element is transparent and viewport-sized*. BottomSheet has that shape (`dialog` → `positioner` → panel); Drawer doesn't. Might be worth stating in that form somewhere durable, since it's the property a future overlay would need to avoid rather than a fact about this component.

**One small question:** the tint hard-codes `--color-background-surface`, which matches the panel token, so themes stay in sync. A consumer overriding the panel's background through `xstyle` would desync the sampled colour from the visible sheet. Narrow enough to be fine — just flagging in case it's worth reading the panel's computed background instead.

PR #5343, opened 2026-08-22
fix(AlertDialog): keep actions readable on narrow surfaces

rubyycheung:
## Summary

- rebase the AlertDialog branch onto latest `main` so it relies on the merged Dialog responsive width clamp from #5352
- remove AlertDialog's small-screen width replacement and pass the consumer's preferred `width` through unchanged (default 400)
- keep AlertDialog's width-only action-order breakpoint for its destructive/cancel semantics, with narrow DOM/visual/tab order aligned and Cancel still the autofocus target
- keep Button labels single-line at the normal size-token height; wide footers may move whole buttons onto another row, while narrow footers stack full-width actions
- update responsive stories, docs, tests, and changeset credit

### Responsive layout

| Check | Result | Evidence |
| --- | --- | --- |
| Dialog-owned surface geometry | Pass | AlertDialog passes `width` through to Dialog. Dialog preserves the requested width and clamps it with `max-width: min(100%, calc(100dvw - var(--spacing-4) - var(--spacing-4)))` from #5352. |
| Wide viewport actions | Pass | Above 640px, AlertDialog keeps Cancel before the destructive action and lets whole buttons move onto another row when both do not fit. Labels retain standard single-line Button behavior. |
| Narrow viewport actions | Pass | At 640px and below, AlertDialog stacks the destructive action above Cancel and stretches both actions to the row while preserving standard single-line Button sizing and labels. |
| Reference states and responsive tests | Pass | The three stories are reference states only unless the reviewer configures viewport/input. Tests cover Dialog's clamp, wide whole-button row movement without changing Button sizing, narrow stacking, pointer/hover independence, and tab-order/autofocus behavior. |

### Touch, pointer, and hover

| Check | Result | Evidence |
| --- | --- | --- |
| Pointer/hover independence | Pass | AlertDialog still queries only `(max-width: 640px)` for semantic action order; it does not query pointer or hover capability and does not reintroduce pointer/touch coupling. |
| Wide touch/coarse/no-hover viewport | Pass | Because action order is width-only, a wide coarse/no-hover viewport stays on the wide horizontal branch. |
| Narrow fine-pointer viewport | Pass | A narrow fine-pointer viewport uses the same stacked destructive-above-Cancel order as mobile. |
| Gestures | N/A | AlertDialog has no swipe dismissal path, and this diff does not add one. |

### Accessibility and interaction contracts

| Check | Result | Evidence |
| --- | --- | --- |
| Keyboard/focus order | Pass | Narrow visual order, DOM order, and tab order are aligned as destructive then Cancel. Cancel remains the `data-autofocus` least-destructive target when the dialog opens. |
| Role/semantics | Pass | The diff does not change AlertDialog's `role="alertdialog"`, modal behavior, or title/description wiring. |
| Action semantics ownership | Pass | Destructive-above-Cancel remains AlertDialog-specific. Generic Dialog owns the surface clamp and footer multi-row affordance; Button retains its standard single-line label behavior; consumers own footer composition, semantics, and ordering. |
| Motion/reduced motion | N/A | This diff does not change Dialog animation or reduced-motion behavior. |

### Mobile viewport constraints

| Check | Result | Evidence |
| --- | --- | --- |
| Width and dynamic viewport clamp | Pass | Dialog #5352 owns `max-width: min(100%, calc(100dvw - var(--spacing-4) - var(--spacing-4)))`; AlertDialog no longer sets `100dvw` or replaces `width` at small viewports. |
| Software keyboard | N/A | This diff changes action-row layout only; it does not change software-keyboard handling. |
| Safe area/dynamic viewport ownership | Pass | AlertDialog delegates viewport geometry to Dialog and only owns the confirmation action layout. |

## Test plan

- `pnpm exec vitest run packages/core/src/AlertDialog/AlertDialog.test.tsx packages/core/src/Dialog/Dialog.test.tsx`
- `pnpm -F @astryxdesign/theme-butter build`
- `pnpm -F @astryxdesign/theme-gothic build`
- `pnpm -F @astryxdesign/theme-matcha build`
- `pnpm -F @astryxdesign/docsite generate && pnpm -F @astryxdesign/docsite typecheck`
- `pnpm exec prettier --check packages/core/src/AlertDialog/AlertDialog.tsx packages/core/src/AlertDialog/AlertDialog.test.tsx packages/core/src/AlertDialog/AlertDialog.doc.mjs apps/storybook/stories/AlertDialog.stories.tsx .changeset/alert-dialog-responsive-actions.md`
- `pnpm check:repo`
- `pnpm exec eslint packages/core/src/AlertDialog/AlertDialog.tsx packages/core/src/AlertDialog/AlertDialog.test.tsx apps/storybook/stories/AlertDialog.stories.tsx`

### DIFF
--- a/.changeset/alert-dialog-responsive-actions.md
+++ b/.changeset/alert-dialog-responsive-actions.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] AlertDialog action labels now wrap, and the action row stacks based on the dialog's own width so destructive decisions remain fully readable on narrow surfaces and at increased zoom.
+
+@rubycheung
--- a/.changeset/alert-dialog-responsive-actions.md
+++ b/.changeset/alert-dialog-responsive-actions.md
@@ -2,6 +2,6 @@
 '@astryxdesign/core': patch
 ---
 
-[fix] AlertDialog lets Dialog preserve and clamp its preferred width, stacks the destructive action above Cancel at 640px and below, and wraps complete decision labels in both horizontal and stacked action layouts regardless of pointer type.
+[fix] AlertDialog lets Dialog preserve and clamp its preferred width, lets whole actions move to another row when needed, and stacks full-width destructive and Cancel actions at 640px and below while preserving standard single-line Button sizing and labels regardless of pointer type.
 
 @rubyycheung
--- a/.changeset/alert-dialog-responsive-actions.md
+++ b/.changeset/alert-dialog-responsive-actions.md
@@ -2,6 +2,6 @@
 '@astryxdesign/core': patch
 ---
 
-[fix] AlertDialog preserves its >640px centered layout and, at 640px and below, fills available width with token gutters, stacks the destructive action above Cancel, and wraps complete decision labels regardless of pointer type.
+[fix] AlertDialog lets Dialog preserve and clamp its preferred width, stacks the destructive action above Cancel at 640px and below, and wraps complete decision labels in both horizontal and stacked action layouts regardless of pointer type.
 
-@rubycheung
+@rubyycheung
--- a/.changeset/alert-dialog-responsive-actions.md
+++ b/.changeset/alert-dialog-responsive-actions.md
@@ -2,6 +2,6 @@
 '@astryxdesign/core': patch
 ---
 
-[fix] AlertDialog action labels now wrap, and the action row stacks based on the dialog's own width so destructive decisions remain fully readable on narrow surfaces and at increased zoom.
+[fix] AlertDialog preserves its >640px centered layout and, at 640px and below, fills available width with token gutters, stacks the destructive action above Cancel, and wraps complete decision labels regardless of pointer type.
 
 @rubycheung
--- a/packages/core/src/AlertDialog/AlertDialog.test.tsx
+++ b/packages/core/src/AlertDialog/AlertDialog.test.tsx
@@ -167,7 +167,11 @@ describe('AlertDialog', () => {
   });
 
   describe('responsive actions', () => {
-    it('preserves the desktop fine-pointer layout above the small breakpoint', () => {
+    const longCancelLabel = 'Keep this workspace and all of its dashboards';
+    const longActionLabel =
+      'Permanently delete this workspace and all dashboards';
+
+    it('preserves the preferred width and horizontal order above the small breakpoint', () => {
       stubAlertDialogMedia({
         smallScreen: false,
         coarsePointer: false,
@@ -180,15 +184,49 @@ describe('AlertDialog', () => {
       expect(window.matchMedia).not.toHaveBeenCalledWith('(hover: none)');
       const dialog = screen.getByRole('alertdialog');
       const buttons = within(dialog).getAllByRole('button');
+      const footerStack = buttons[0].parentElement!;
       expect(dialog.getAttribute('style')).toContain('--x-width: 400px');
+      expect(dialog.getAttribute('style')).toContain(
+        '--x-maxWidth: min(100%, calc(100dvw - var(--spacing-4) - var(--spacing-4)))',
+      );
       expect(buttons.map(button => button.textContent)).toEqual([
         'Cancel',
         'Delete',
       ]);
-      expect(getComputedStyle(buttons[0]).whiteSpace).toBe('nowrap');
+      expect(getComputedStyle(footerStack).flexWrap).toBe('wrap');
     });
 
-    it('stacks destructive above cancel on a narrow fine-pointer screen', () => {
+    it('wraps long action labels in the wide horizontal branch', () => {
+      stubAlertDialogMedia({
+        smallScreen: false,
+        coarsePointer: false,
+        noHover: false,
+      });
+      render(
+        <AlertDialog
+          {...defaultProps}
+          cancelLabel={longCancelLabel}
+          actionLabel={longActionLabel}
+        />,
+      );
+
+      const dialog = screen.getByRole('alertdialog');
+      const buttons = within(dialog).getAllByRole('button');
+      const footerStack = buttons[0].parentElement!;
+      expect(buttons.map(button => button.textContent)).toEqual([
+        longCancelLabel,
+        longActionLabel,
+      ]);
+      expect(getComputedStyle(footerStack).flexWrap).toBe('wrap');
+      for (const button of buttons) {
+        const computed = getComputedStyle(button);
+        expect(computed.whiteSpace).toBe('normal');
+        expect(computed.height).toBe('auto');
+        expect(computed.maxWidth).toBe('100%');
+      }
+    });
+
+    it('stacks long destructive action labels above cancel on a narrow fine-pointer screen', () => {
       stubAlertDialogMedia({
         smallScreen: true,
         coarsePointer: false,
@@ -197,8 +235,8 @@ describe('AlertDialog', () => {
       render(
         <AlertDialog
           {...defaultProps}
-          cancelLabel="Keep this workspace"
-          actionLabel="Permanently delete workspace"
+          cancelLabel={longCancelLabel}
+          actionLabel={longActionLabel}
         />,
       );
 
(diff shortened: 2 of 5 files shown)


packages/core/src/AlertDialog/AlertDialog.tsx
```diff
@@ -147,19 +166,48 @@ export function AlertDialog({
...
+
   return (
     <Dialog
       {...rest}
       ref={ref}
       isOpen={isOpen}
       isInline={isInline}
       onOpenChange={onOpenChange}
-      width={width}
+      width={isSmallScreen ? SMALL_SCREEN_DIALOG_WIDTH : width}
```
cixzhang: This replaces the consumer's `width` rather than clamping it, and `100dvw` ignores the container.

packages/core/src/AlertDialog/AlertDialog.tsx
```diff
@@ -185,24 +233,14 @@ export function AlertDialog({
...
-              />
-              <Button
-                variant={actionVariant}
-                label={actionLabel}
-                onClick={onAction}
-                isLoading={isActionLoading}
-              />
-            </HStack>
+            <Stack
+              direction={isSmallScreen ? 'vertical' : 'horizontal'}
```
cixzhang, line 240: The horizontal branch needs `wrap` — long labels still spill out of the surface above 640px.

.changeset/alert-dialog-responsive-actions.md
```diff
@@ -0,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] AlertDialog preserves its >640px centered layout and, at 640px and below, fills available width with token gutters, stacks the destructive action above Cancel, and wraps complete decision labels regardless of pointer type.
+
+@rubycheung
```
cixzhang: Handle typo — your other changesets credit `@rubyycheung`.

cixzhang (changes requested):
Thanks for this — the narrow stacking is right, and on a phone it's a real fix: the cancel label gets sliced in half today.

`100dvw` doesn't know its container, so the inline docs preview now renders 468px inside a 458px column at a 500px viewport and runs off the page, and dragging a window across 640px snaps the dialog between 400px and 608px. [#5352](https://github.com/facebook/astryx/pull/5352) already clamps this in Dialog, keeps the consumer's `width`, and needs no breakpoint:

```
max-width: min(100%, calc(100dvw - var(--spacing-4) - var(--spacing-4)))
```

So I'd let Dialog own the width and drop the override here.

Separately, and pre-existing: the clipping is still there above 640px — at 1440px that same cancel label is cut by the surface edge, because the footer row can't reflow. `wrap` on the horizontal Stack, like [#5352](https://github.com/facebook/astryx/pull/5352)'s footers, covers both sides.

Should destructive-above-Cancel be a Dialog footer behavior too, or is that AlertDialog's alone?

[Reviewed by Robohands]

packages/core/src/AlertDialog/AlertDialog.tsx
rubyycheung: Fixed in bb81a6a. I removed AlertDialog’s width override and rebased onto the merged Dialog sizing change, so Dialog now preserves the requested 400px width and applies its container + dynamic-viewport clamp.

packages/core/src/AlertDialog/AlertDialog.tsx
rubyycheung: Fixed in bb81a6a. The wide horizontal action row now wraps, with coverage for long labels above 640px; the narrow branch remains stacked with the destructive action above Cancel.

.changeset/alert-dialog-responsive-actions.md
rubyycheung: Fixed in bb81a6a. The changeset now credits @rubyycheung.

packages/core/src/AlertDialog/AlertDialog.tsx
```diff
@@ -17,18 +17,38 @@
...
+import {useMediaQuery} from '../hooks';
+import {sizeVars} from '../theme/tokens.stylex';
+
+const SMALL_SCREEN_QUERY = '(max-width: 640px)';
+
+const styles = stylex.create({
+  action: {
+    maxWidth: '100%',
+    minWidth: 0,
+    height: 'auto',
```
imdreamrunner: 8px padding plus 20px line-height is 36, so `--size-element-md` stops reaching these buttons.

packages/core/src/AlertDialog/AlertDialog.tsx
```diff
@@ -185,24 +238,19 @@ export function AlertDialog({
...
-            </HStack>
+            {/* Generic Dialog footers should wrap, but their consumers own
+                action semantics and order. AlertDialog keeps this
+                destructive-above-Cancel narrow order because confirmation
+                actions have component-specific meaning here. */}
+            <Stack
+              direction={isSmallScreen ? 'vertical' : 'horizontal'}
+              gap={2}
+              hAlign={isSmallScreen ? 'stretch' : 'end'}
+              wrap={isSmallScreen ? 'nowrap' : 'wrap'}>
```
imdreamrunner, line 243: Wrapping belongs in Button — a plain Dialog footer clips the same label. Separate PR?

packages/core/src/AlertDialog/AlertDialog.test.tsx
```diff
@@ -109,6 +166,158 @@ describe('AlertDialog', () => {
...
+      const footerStack = buttons[0].parentElement!;
+      expect(buttons.map(button => button.textContent)).toEqual([
+        longCancelLabel,
+        longActionLabel,
+      ]);
+      expect(getComputedStyle(footerStack).flexWrap).toBe('wrap');
+      for (const button of buttons) {
+        const computed = getComputedStyle(button);
+        expect(computed.whiteSpace).toBe('normal');
+        expect(computed.height).toBe('auto');
```
imdreamrunner: Pins `height: auto`, so it moves with the height fix.

imdreamrunner (approved):
Thanks — the wrapping holds at 1440 and in RTL, not just on phones. One thing to fix before you merge, and one question.

[Reviewed by Robohands]

imdreamrunner:
Re-checked `bb81a6a` against @cixzhang's review — all three code points are in, and the ownership question is answered:

- **Width override is gone.** AlertDialog passes the consumer's `width` (default 400) straight through to Dialog, and Dialog clamps with `min(100%, calc(100dvw - var(--spacing-4) - var(--spacing-4)))`. No breakpoint width swap, so no 400↔608 snap dragging across 640px and no overflow of the 458px docs column.
- **Horizontal branch wraps.** `wrap={isSmallScreen ? 'nowrap' : 'wrap'}` on the footer Stack, with a test for long labels above 640px, so the 1440px clipping is covered too.
- **Changeset credits `@rubyycheung`.**
- **Destructive-above-Cancel stays AlertDialog's alone** — generic Dialog footers own wrapping, consumers own action order and semantics. That's stated in the footer comment and the PR description.

My own notes from yesterday still stand.

[Reviewed by Robohands]

packages/core/src/AlertDialog/AlertDialog.tsx
rubyycheung: Fixed in 6b2ec1f19b2: removed AlertDialog’s `height: auto` override. The actions now retain Button’s normal `--size-element-md` height and single-line label behavior, and the regression test checks that token height.

packages/core/src/AlertDialog/AlertDialog.tsx
rubyycheung: Agreed on the ownership boundary. I removed label wrapping from AlertDialog rather than broadening Button behavior in this PR. AlertDialog now only lets whole buttons move to another row above 640px and stacks full-width buttons below it; Button labels stay single-line.

packages/core/src/AlertDialog/AlertDialog.test.tsx
rubyycheung: Updated in 6b2ec1f19b2: this test no longer pins `height: auto`; it now verifies the normal medium Button token height and single-line label behavior.

PR #5344, opened 2026-08-22
fix(selector): render in-listbox dividers as presentation, not separator (#4994)

### LINKED ISSUE #4994: Selector and MultiSelector put a role=separator Divider inside role=listbox (critical axe violation)

gonzoblasco:
## What

Fixes #4994 - `Selector` and `MultiSelector` placed a `role="separator"` divider inside their `role="listbox"` popup. A `listbox` only permits `option`/`group` children, so a separator among them was a **critical** axe violation (`aria-required-children`, WCAG 1.3.1/4.1.2) that could drop or mis-report the listbox children in a screen reader.

## Changes

- **`Divider`** now accepts an explicit `role` prop (default `'separator'`), so consumers can override it for placement inside a restricted-role container.
- **`Selector`** renders its in-listbox option dividers as `role="presentation"`, keeping the visual rule out of the accessibility tree.
- **`MultiSelector`** does the same for its in-listbox option dividers.

The standalone divider between the search row and the listbox keeps its `role="separator"` — it is a legitimate separator between two sections, not a child of the listbox.

## Validation

- New tests in both `Selector` and `MultiSelector` assert the in-listbox divider renders `role="presentation"`.
- Existing tests that assert the search-row divider stays `role="separator"` still pass (that divider is outside the listbox).

Fixes #4994

### DIFF
--- a/.changeset/selector-listbox-divider-role.md
+++ b/.changeset/selector-listbox-divider-role.md
@@ -1,0 +1,5 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] `Selector` and `MultiSelector` no longer place a `role="separator"` divider inside their `role="listbox"` popup. A `listbox` only permits `option`/`group` children, so a separator among them was a critical axe violation (`aria-required-children`, WCAG 1.3.1/4.1.2) that could drop or mis-report the listbox children in a screen reader. The visual dividers between options now render `role="presentation"`, keeping them out of the accessibility tree while preserving the visual rule. A standalone divider between the search row and the listbox keeps its `role="separator"`. `Divider` now also accepts an explicit `role` prop (default `separator`) so consumers can override it.
--- a/packages/core/src/Divider/Divider.tsx
+++ b/packages/core/src/Divider/Divider.tsx
@@ -150,6 +150,7 @@ export function Divider({
   ref,
   'aria-label': ariaLabel,
   'aria-labelledby': ariaLabelledBy,
+  role = 'separator',
   ...props
 }: DividerProps) {
   const isHorizontal = orientation === 'horizontal';
@@ -166,7 +167,7 @@ export function Divider({
     <div
       ref={ref}
       {...props}
-      role="separator"
+      role={role}
       aria-orientation={orientation}
       aria-label={ariaLabel}
       aria-labelledby={resolvedLabelledBy}
--- a/packages/core/src/MultiSelector/MultiSelector.test.tsx
+++ b/packages/core/src/MultiSelector/MultiSelector.test.tsx
@@ -1981,6 +1981,29 @@ describe('MultiSelector list structure', () => {
         Node.DOCUMENT_POSITION_FOLLOWING,
     ).toBeTruthy();
   });
+
+  it('renders a divider inside the listbox as presentation, not a separator', async () => {
+    // A role="separator" is not a permitted child of role="listbox" (axe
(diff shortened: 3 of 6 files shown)


cixzhang:
Thanks for chasing this down — your diagnosis is right, and leaving the search-row divider alone was exactly the right call.

Two PRs came in for [#4994](https://github.com/facebook/astryx/issues/4994) and we've gone with [#5107](https://github.com/facebook/astryx/pull/5107), now merged, which hides the divider with `aria-hidden` instead of changing `Divider` itself.

The measurement that decided it: `Divider` always writes `aria-orientation`, and `presentation` doesn't allow it — so axe trades `aria-required-children` for `aria-allowed-attr` on the same element. Critical ×1 becomes critical ×2. That's not obvious from reading it, and it's the kind of thing only an axe run turns up.

Sorry it sat a while before we said so. Please do send more — you found a real bug and read it correctly.

[Reviewed by Robohands]

gonzoblasco:
@cixzhang, thanks for the detailed explanation - that's a genuinely better call, and I appreciate you spelling out the reasoning.

The aria-orientation point is exactly the kind of thing that's invisible until an axe run surfaces it: Divider always writes it, presentation doesn't allow it, so my approach would have just traded aria-required-children for aria-allowed-attr on the same element. Critical ×1 becoming critical ×2 is a real regression, not a wash. aria-hidden on the divider keeps the visual and the role="separator" in the DOM while taking it out of the accessibility tree entirely - cleaner, no new API surface, and it follows the pattern you already use for section headings. #5107 is the right fix.

Glad the diagnosis and the call to leave the search-row divider alone were on point. I'll keep an eye out for more a11y issues in the selector space. Thanks again for the thorough review.

PR #5349, opened 2026-08-22
feat(core): TabList reads its role and speaks the ARIA tabs pattern

cixzhang:
## What

`TabList` can speak the WAI-ARIA [tabs pattern](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/) when a caller asks for it — and there is no new prop for the asking. `TabList` declares `role?: AriaRole` and **reads** it, the way `LayoutHeader`, `LayoutContent` and `LayoutPanel` already declare and document theirs:

| `role` | what the strip is |
|---|---|
| `"tablist"` | the tabs pattern: `role="tablist"` on the strip, `role="tab"` and `aria-selected` on the tabs, `aria-controls` from a new `panelId` on `Tab` |
| *unset* | **unchanged** — the `<nav>` landmark with `aria-current` it has always been |
| anything else | passed through to the element, untouched, exactly as before |

Stacked on #5348 — review that one first; this branch contains it.

## Why no new prop

`role` already reaches the DOM: `{...restProps}` spreads it onto the wrapper, so a caller can pass `role="tablist"` today and get a tablist whose children are still `<button>`s with `aria-current` — invalid markup, no `aria-selected`, nav-style arrow keys, and no warning. Reading the role turns that silent breakage into the correct behaviour, and declaring it explicitly is what puts it in the type, the prop table and the docs, where a behaviour keyed on an anonymous passthrough would be invisible.

## Nothing changes for anyone who does not ask

A caller who passes no `role` gets exactly what they got before: same element, same classes, same `aria-current`, same both-axis arrow keys, same pixels. Measured rather than asserted — seven configurations of a no-role strip (plain, divider + `lg` + `fill`, with a menu, with `href` tabs, icon-only `sm`, `overflow="visible"`, with action buttons) serialise **byte-identically** against this PR's base, 31,326 bytes each. Against `main` the DOM differs only where #5348 changed it, and the role and ARIA attribute set is identical there too.

Under the asserted role the strip takes only the horizontal arrows and leaves ArrowUp and ArrowDown to the page, because a tablist reports itself as horizontal. ArrowLeft, ArrowRight, Home, End, Tab and the roving tab stop are unchanged.

## Two things the asserted role tells you about

**A tab with an `href` stops navigating**, and a development warning says so. Activating a tab swaps a panel in place, so a tab that navigates is a false statement — but it is also the caller's link, so it is only ever ignored where the pattern was asked for by name.

**A tab that controls nothing is asked for a `panelId`.** Either `panelId` or an `aria-controls` you wrote yourself satisfies it — `Tab` already spreads yours to the DOM, so both spellings count and a hand-written one is never overwritten. Nothing validates that the target exists: a panel may mount later.

`aria-controls` is emitted **only** when an id is supplied. Generating one would ship a dangling reference, which axe rates **critical** (`aria-valid-attr-value`); measured in Chromium on three hand-built tablists, dangling → 1 critical violation, resolved → clean, **absent → clean**.

A menu or any other non-tab in a tablist strip is invalid markup, and warns too — read off the rendered DOM in development, so a menu behind a conditional, inside a `.map`, or wrapped in a component of your own is still caught.

## Testing

Accessibility trees read from Chromium's own tree (CDP), not inferred from the DOM:

| | landmark | widget | wrapper | tabs |
|---|---|---|---|---|
| no `role` | `navigation("Tabs")` | none | `<nav>` | 3 buttons, `aria-current="true"` on the selected one |
| `role="tablist"` + `panelId` | none | `tablist("Project views")` | `<div>` | 3 `tab`s, `aria-selected`, `aria-controls`, 3 `tabpanel`s |
| `role="tablist"`, no `panelId` | none | `tablist("Project views")` | `<div>` | 3 `tab`s, no `aria-controls`, one warning each |
| `role="navigation"` | `navigation("Project sections")` | none | `<nav role="navigation">` | 3 buttons, `aria-current="true"` |

The no-`role` row is identical, row for row, to the same story on `main`.

axe is clean on every one of them, discounting `region`, which the repo's own audit disables for story isolation (`.github/scripts/accessibility-audit.js:54`).

86 unit tests green across `TabList` and the edge-compensation suite; every pre-existing TabList test is untouched — the test file's diff against the base is additions only. Typecheck and lint clean.

### DIFF
--- a/packages/core/src/TabList/Tab.tsx
+++ b/packages/core/src/TabList/Tab.tsx
@@ -7,7 +7,8 @@
  * @input Uses React, StyleX, TabListContext
  * @output Exports Tab component and TabProps type
  * @position Core tab item; renders as button or anchor in navigation with a
- *   divider-overlay selected indicator
+ *   divider-overlay selected indicator. Where the TabList speaks the tabs
+ *   pattern it is a button with role="tab".
  *
  * SYNC: When modified, update:
  * - /packages/core/src/TabList/TabList.doc.mjs
@@ -35,6 +36,7 @@ import {tabScope} from './tab.markers.stylex';
 import {useLinkComponent} from '../Link/useLinkComponent';
 import type {LinkComponentType} from '../Link/types';
 import {mergeProps} from '../utils';
+import {useDevWarning} from '../hooks/useDevWarning';
 import {EDGE_COMP_ATTR} from '../Layout/edgeCompensation.stylex';
 import {themeProps} from '../utils/themeProps';
 import {focusOutlineProps} from '../utils/focusOutline.stylex';
@@ -64,8 +66,20 @@ export interface TabProps extends BaseProps<HTMLButtonElement> {
   isLabelHidden?: boolean;
   /**
    * URL to navigate to. When provided, renders as an anchor element.
+   *
+   * Ignored in a TabList given an explicit `role="tablist"`: activating a tab
+   * there swaps a panel in place, so a tab that navigates would be a false
+   * statement.
    */
   href?: string;
+  /**
+   * Id of the panel this tab controls, wired up as `aria-controls` where the
+   * TabList speaks the tabs pattern. Put the same id on the panel element.
+   *
+   * Has no effect under the navigation pattern, where there is no panel to
+   * associate — a development warning says so.
+   */
+  panelId?: string;
   /**
    * Icon element shown when tab is not selected.
    */
@@ -228,6 +242,7 @@ export function Tab({
   label,
   isLabelHidden = false,
   href,
+  panelId,
   icon,
   selectedIcon,
   endContent,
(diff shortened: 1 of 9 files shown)


packages/core/src/TabList/TabList.tsx
```diff
@@ -349,12 +436,24 @@ export function TabList({
...
   const hasScroll = overflow !== 'visible';
 
+  const stripRef = useRef<HTMLDivElement | null>(null);
+  // `role="tablist"` is an assertion: the consumer says these are tabs, and
+  // the strip is one whatever it holds. Any other role is theirs to place, so
+  // it rules the auto pick out too — a strip that was told it is something
+  // else does not get to be a tablist behind the caller's back.
+  const isTabListAsserted = role === 'tablist';
+  const isPatternAuto = role === undefined;
+  const canAutoTabs = useAutoPattern(stripRef, isPatternAuto);
```
cixzhang: Hmm, this is the default change — every caller passes no role today. Probably wants its own PR.

packages/core/src/TabList/Tab.tsx
```diff
@@ -242,13 +259,43 @@ export function Tab({
...
+
+  // Only where the tabs pattern was asked for: a strip left to pick reaches it
+  // through tabs that were written for the navigation pattern, and scolding
+  // the consumer for a panel they were never asked for is noise.
+  useDevWarning(
+    'Tab',
+    'a tab in a role="tablist" TabList controls nothing: pass panelId with ' +
+      'the id of the panel it opens, so assistive technology can associate ' +
+      'the two.',
+    isTabRole && !tabListCtx.isPatternAuto && controls == null,
```
cixzhang: Suppressed here, so an auto-picked tablist never asks for a panel. Those are the strips that have none.

cixzhang (commented):
Thanks — reading `role` is the right shape, and that half is an approve from me.

I'd rather land it alone. Picking the pattern when no role is given is a second decision, and not the review's to settle: every caller passes no role, so they all become tablists — and none pass `panelId`, so a screen reader announces a selected tab with no panel to move to. The explicit role answers a problem the body states; I can't find one written down for the default change, and the 20 moved tests and the `[breaking]` are all its.

Send the explicit-role half on its own and it lands; the auto pick wants its own PR, with that reason in the body.

[Reviewed by Robohands]

cixzhang (commented):
Thanks — this is the split the earlier review asked for, and the explicit-role half stands on its own. Approve from me once you undraft.

[Reviewed by Robohands]

packages/core/src/TabList/Tab.tsx
```diff
@@ -242,13 +257,36 @@ export function Tab({
...
+  // A consumer who wired aria-controls by hand already said which panel this
+  // is, so panelId is the sugar, not the only way in.
+  const controls = panelId ?? restProps['aria-controls'];
+
+  useDevWarning(
+    'Tab',
+    'a tab in a role="tablist" TabList controls nothing: pass panelId with ' +
+      'the id of the panel it opens, so assistive technology can associate ' +
+      'the two.',
+    isTabRole && controls == null,
```
cixzhang, line 287: Hmm, panelId with no role is dropped silently. The other three mistakes here warn.

cixzhang: Taken — it warns now: `panelId` outside a `role="tablist"` strip says so and points at the role. `Tab.tsx:295`.

PR #5351, opened 2026-08-23
docs: add responsive and interaction readiness rubric

rubyycheung:
## Latest refinement

Popover review exposed two reusable gaps, now added without expanding the rubric:

- anchored/layered surfaces must test explicit, trigger-derived, intrinsic, and consumer-overridden sizing on both axes, including both outcomes: oversized content scrolls and fitting content remains unclipped;
- responsive stories must use the real Storybook viewport/container and preserve normal trigger, outside-dismiss, keyboard-dismiss, and focus behavior. Stories may auto-open through interaction code rather than being permanently controlled open; simulator/device evidence remains separate.

## Summary

- narrow the rubric scope around Pass/Fail/Blocked/N/A status semantics, applicable scenario evidence, conditional adaptive/transient appendices, and interaction-affecting motion while moving generic hardening guidance out of scope.
- address reviewer feedback in #5351 by adding a platform/browser evidence layer: Storybook and Playwright cannot prove iOS Safari/shell behavior, iOS touch/WebKit claims need Simulator/device evidence, unavailable platform evidence is Blocked rather than N/A, and diagnostics remain development-only.
- add a reusable `astryx docs responsive-interaction-readiness` rubric for reviewing new or changed components across viewport space, pointer precision, hover availability, gestures, transient and queued UI, mobile viewport constraints, and existing interaction contracts
- add an Adaptive presentation decision gate so authors decide from task semantics/product intent before using width or input capability as implementation signals
- add conditional Transient and queued UI checks covering queue/stack policy, timing, announcement semantics, gesture alternatives, and viewport obstruction evidence for toasts, snackbars, notifications, transient banners, and similar surfaces
- add generalized viewport obstruction and placement guidance for fixed overlays, transient feedback, floating actions, and other viewport-anchored surfaces
- wire Responsive and Interaction Readiness into lab readiness as a new evidenced human-review check so component reviews can record Pass/Fail/N/A plus linked evidence instead of relying on an isolated checklist
- add a copyable component-PR reporting template with compact `Check | Result | Evidence` subsections that preserve the four viewport/input independence scenarios
- link the rubric from principles and layout guidance so authors can discover it while authoring responsive or input-sensitive behavior

## Why

Responsive component work was mixing viewport width with touch capability. This rubric keeps those axes separate, preserves the four reference scenarios that prove width independence and input independence, and organizes review outcomes into four scannable categories: responsive layout; touch, pointer, and hover; accessibility and interaction contracts; and mobile viewport constraints.

The adaptive presentation gate captures the Dialog vs BottomSheet lesson generally: responsive pressure usually means reflow/resize, while presentation changes must be explicit, opt-in, documented, and backed by evidence for changed placement, motion, dismissal, focus, scrolling, gesture, or announcement contracts.

The transient UI guidance captures the Toast lessons generally: queued or auto-dismissing surfaces need explicit stack/queue/timing/announcement/gesture evidence, and unrelated components can mark those checks N/A.

The viewport obstruction guidance keeps placement decisions tied to actual obstructions and content priority, not touch/coarse-pointer inference; safe-area support alone is not treated as keyboard or app-chrome avoidance.

## WCAG AA boundary

The rubric is scoped to WCAG 2.2 Level AA. For target size, it uses Success Criterion 2.5.8: at least 24x24 CSS px or a permitted exception.

## Validation

- `pnpm exec prettier --write .changeset/responsive-interaction-readiness.md internal/lab-readiness/catalog.mjs internal/lab-readiness/audit.test.mjs packages/cli/assets/docs/responsive-interaction-readiness.doc.mjs packages/cli/foundation/discovery/docs-discovery.test.mjs`
- `pnpm exec prettier --check .changeset/responsive-interaction-readiness.md internal/lab-readiness/catalog.mjs internal/lab-readiness/audit.test.mjs packages/cli/assets/docs/responsive-interaction-readiness.doc.mjs packages/cli/foundation/discovery/docs-discovery.test.mjs`
- `pnpm exec vitest run internal/lab-readiness/audit.test.mjs packages/cli/foundation/discovery/docs-discovery.test.mjs packages/cli/api/docs/integrationDocs.test.mjs`
- `pnpm -F @astryxdesign/cli astryx docs responsive-interaction-readiness --dense`
- `pnpm check:changesets`
- `pnpm check:repo`
- `pnpm lab:readiness:check`

## Relationship to #5343

This is intentionally separate from the AlertDialog implementation PR #5343. AlertDialog is only used as a concise worked example of the rubric; this branch is based on current `main` and is not stacked on #5343.


light-merlin-dark:
A field note from adopting this rubric's spirit downstream, offered because it cost us a real debugging session this week:

The mobile-readiness rubric marks "Touch, pointer, and hover: Pass" on evidence from Storybook and Playwright. Both are structurally blind to WebKit-on-iOS: Playwright's WebKit is macOS WebKit with no iOS shell and no top-layer `<dialog>` behaviour to match. A component can pass every current check and still behave differently on a real iPhone.

Recommend adding an iOS Simulator verification step to the rubric for touch-interaction claims. A cheap technique that works without a console: render an on-page, `pointer-events: none` forensics overlay behind a query flag, so one screenshot reads the whole event/dispatch story. One React-specific trap worth documenting for anyone instrumenting this: React freezes props in development, so any handler probe must survive commits (intercept the props property and return a wrapped copy) — mutating the props object silently no-ops and makes a working dispatch look broken.

For what it's worth, we ran exactly this suspicion against `@astryxdesign/core@0.4.0`'s Dialog on an iOS 26.5 simulator with real touch injection: dispatch is clean, top-layer propagation is clean. The rubric gap is about what the current instruments *can't* see, not about a defect they missed.

By Merlin ([rbeckner.com](https://rbeckner.com)) and his AI agent (Claude Code, Opus 5 (1M context), thinking on: high).

rubyycheung:
@light-merlin-dark Thanks so much for your feedback! Will definitely include, let me know if you have any other thoughts on the rubric. 🙂

imdreamrunner (commented):
Thanks — the tighter scope since yesterday reads much better, and the stale AlertDialog example is gone.

One thing before this undrafts: where the rubric lives. I framed this as a preference earlier; it isn't one. [#5368](https://github.com/facebook/astryx/pull/5368) landed after you branched and settles it — `packages/cli/assets/docs/` is for people building *with* Astryx, and its README closes with this exact rubric as the worked example: grading criteria go to the Component Audit Rubric, **as a section on the page that already covers it, not a new page**.

One finding, three places: move the rubric body there, drop the Principles rule, and trim the trailing `astryx docs` pointer off the `layout.doc.mjs` sentence — both references dangle once the topic goes. Keep the rest of that sentence, and the lab-readiness check; those are right as they are.

Do that and undraft, and I'll approve. Worth rebasing first; `CLAUDE.md` carries the rule now too.

[Reviewed by Robohands]

packages/cli/assets/docs/responsive-interaction-readiness.doc.mjs
```diff
@@ -0,0 +1,356 @@
...
+          text: 'Non-normative aid: when a remote console is unavailable, a query-flagged, on-page forensics overlay can make iOS evidence reviewable in one screenshot. Keep it development-only, disable or remove it in production, set pointer-events: none so it cannot change the interaction under test, and record only the event, pointer, dispatch, focus, dismissal, viewport, and top-layer details needed for diagnosis.',
+        },
+      ],
+    },
+    {
+      title: 'Scope example',
+      content: [
+        {
+          type: 'prose',
+          text: 'AlertDialog illustrates the boundary: constrained-width evidence should show content fit, wrapping, target size, focus/dismissal continuity, and platform evidence if the dialog relies on platform-specific top-layer behavior. General copy alignment, visual polish, or unchanged roles belong in broader component-hardening review unless the responsive or adaptive behavior changes them.',
```
ernestt, line 337: Worth adding Stepper here as the counter-example to AlertDialog — it draws the line on the other side, which is exactly the edge people get wrong about adaptive presentation.

Stepper's horizontal → collapsed change is **not** a responsive adaptation candidate. It never asks a media query (or a container query) anything about the viewport. It measures its own frame with a ResizeObserver, divides that by the number of registered steps, and collapses when a step's share drops below a fixed `MIN_STEP_WIDTH` (112px). So the threshold follows step count, not a breakpoint: four steps hold out to ~448px, seven need ~784px. And what it collapses *to* is configuration the tool builder sets (`hasCollapsedControls`, `hasCollapsedLabel`), not a presentation the system infers on its own.

Same surface, same interaction contract, width-and-content-fit driven — that belongs under "Available space, content fit, and overflow," with the adaptive presentation appendix marked N/A. Adaptation is for opting into a *different* presentation by task semantics; a component reflowing itself because its own content stopped fitting is not that.

Suggested prose to append to this section:

```
{
  type: 'prose',
  text: 'Stepper illustrates the opposite edge: its collapse is not adaptive presentation. It reads no media or container query — it measures its own frame, divides by the number of registered steps, and collapses when a step\'s share falls below a fixed minimum step width, so the threshold follows step count rather than a breakpoint. The collapsed result is configured by the consumer rather than inferred, and the surface, semantics, and interaction contract are unchanged. Record it under available space, content fit, and overflow, and mark the adaptive presentation appendix N/A. Reserve adaptive presentation for an intentional change of presentation, not for a component reflowing itself when its own content stops fitting.',
},
```

PR #5353, opened 2026-08-23
fix(Toast): improve responsive layout and motion

rubyycheung:
## Problem

Toast had five responsive-layout, motion, and accessibility rough edges:

1. A 400px Toast could exceed a narrow viewport, and the fixed edge padding did not account for device safe areas or RTL logical edges.
2. A long or localized message could overflow. When it wrapped, Undo and dismiss controls could sit below the first line.
3. The 12px stack gap was loose, and removal needed to collapse the wrapper rather than move the stack in one step.
4. Top and bottom placements used the same motion direction.
5. An empty `ToastViewport` exposed a redundant Notifications landmark; nested viewports could therefore expose duplicate landmarks.

## Solution

- Span the available inline viewport and put the 400px cap on an explicitly `border-box` wrapper. The Toast itself is also `border-box`, so containment does not depend on a consumer loading Astryx reset CSS.
- Reserve `max(16px, env(safe-area-inset-*))` on every viewport edge. Because safe-area insets are physical while the padding is logical, RTL swaps which left/right inset `padding-inline-start/end` reads.
- Let the body shrink and break anywhere while keeping the trailing controls in the same flex row. The trailing control group is one body line box tall, so controls remain centered on the first line even when the body wraps or a consumer supplies a taller control.
- Tighten only the inter-Toast gap from 12px to 8px. The wrapper nearest the configured viewport edge drops that gap; the selector flips for the reversed top stack.
- Keep the existing opacity/timing contract and add one private inherited block-axis offset: bottom placements start/end at `+8px`, top placements at `-8px`. No scale or corner-origin behavior is added.
- Publish `role="region"`, the Notifications label, and `tabIndex={-1}` only while Toasts are present. In a nested composition, only the viewport that actually receives a Toast becomes a landmark.

## Scope

- No public API change.
- `position`, `maxVisible`, custom `inset`, auto-hide defaults, announcement dispatch, focus handoff, collision behavior, and dismissal reasons are unchanged.
- Queue policy is unchanged; `maxVisible` still renders the newest N Toasts.
- Browser-toolbar/chrome behavior is deliberately unchanged and not claimed by this PR.
- Toast block padding remains the existing 16px.

## Evidence

Measured in Chromium at 320×640, DPR 2, against the built package:

- all Toasts stayed inside the viewport with and without Astryx reset CSS;
- an unbroken URL had `0px` horizontal overflow;
- stack gaps were `8px`, with the viewport-edge gap remaining `16px`;
- a four-line message, 36px Undo button, and 28px dismiss button shared the same first-line center: `26px / 26px / 26px`;
- bottom entry began at `translateY(8px)` and top entry at `translateY(-8px)`, both settling at zero.

Four focused Storybook references cover narrow localized copy, RTL placement, top/bottom stack motion, and nested landmark behavior.

## Validation

- 110 focused Toast/derived-variable tests pass (30 ToastViewport, 4 useToast, 76 derived-variable).
- Full UI run: 12,365 tests passed; the combined run hit the repo's known CLI temporary-directory race, then the CLI/scripts node suite passed serially: 3,265/3,265.
- `pnpm lint:strict`
- `pnpm build`
- Core, Lab, Charts, CLI, and Storybook typechecks
- Production Storybook build
- Toast accessibility audit: 13 stories, 0 violations
- Toast RTL audit: 0 failures

### DIFF
--- a/packages/core/src/Toast/Toast.tsx
+++ b/packages/core/src/Toast/Toast.tsx
@@ -2,6 +2,21 @@
 
 'use client';
 
+/**
+ * @file Toast.tsx
+ * @input Uses React timers, Toast options, Button/Icon, MediaTheme, tokens, and
+ *   placement-derived motion variables inherited from ToastViewport
+ * @output Exports the rendered Toast surface and its pause/dismiss behavior
+ * @position Core implementation; rendered by ToastViewport and documented by Toast.doc.mjs
+ *
+ * SYNC: When Toast layout, timer pause, media theme, or dismissal behavior changes,
+ *   update these files to stay in sync:
+ * - /packages/core/src/Toast/ToastViewport.test.tsx
+ * - /packages/core/src/Toast/Toast.doc.mjs
+ * - /apps/storybook/stories/Toast.stories.tsx
+ * - /packages/cli/assets/templates/blocks/components/Toast/ (showcase blocks)
+ */
+
 import {useCallback, useEffect, useRef} from 'react';
 import type {ReactNode} from 'react';
 import * as stylex from '@stylexjs/stylex';
@@ -24,13 +39,15 @@ import type {ToastType, ToastDismissReason} from './types';
 import {themeProps} from '../utils/themeProps';
 import {useTranslator} from '../i18n';
 
+const TOAST_EDGE_DRIFT = spacingVars['--spacing-2'];
 const styles = stylex.create({
   root: {
     paddingBlock: spacingVars['--spacing-4'],
     paddingInline: spacingVars['--spacing-4'],
     borderRadius: radiusVars['--radius-container'],
+    boxSizing: 'border-box',
     width: 400,
-    maxWidth: 'min(100%, calc(100vw - 32px))',
+    maxWidth: '100%',
     boxShadow: shadowVars['--shadow-med'],
     opacity: 1,
     fontFamily: typographyVars['--font-family-body'],
@@ -45,15 +62,17 @@ const styles = stylex.create({
     transitionTimingFunction: easeVars['--ease-standard'],
     '@starting-style': {
       opacity: 0,
-      transform: 'translateY(8px)',
+      transform: `translateY(var(--_toast-slide-y, ${TOAST_EDGE_DRIFT}))`,
     },
   },
   variantDefault: {
     backgroundColor: colorVars['--color-background-inverted'],
   },
+
   inner: {
     display: 'flex',
     alignItems: 'flex-start',
+    flexWrap: 'nowrap',
     gap: spacingVars['--spacing-3'],
     width: '100%',
   },
@@ -63,17 +82,22 @@ const styles = stylex.create({
   content: {
     flex: 1,
     minWidth: 0,
+    overflowWrap: 'anywhere',
   },
   exiting: {
     opacity: 0,
-    transform: 'translateY(-8px)',
+    transform: `translateY(var(--_toast-slide-y, ${TOAST_EDGE_DRIFT}))`,
   },
   endContent: {
     flexShrink: 0,
     display: 'flex',
     alignItems: 'center',
     gap: spacingVars['--spacing-2'],
-    marginBlock: `calc(${spacingVars['--spacing-1']} * -1)`,
+    // Keep every trailing control centered on the first 20px body line, even
+    // when the body wraps or a consumer supplies a control taller than the
+    // built-in 28px dismiss button. The action label should still stay short;
+    // the wrappers above let it break rather than widen the Toast.
+    blockSize: `calc(${typeScaleDefaults['--text-body-size']} * ${typeScaleDefaults['--text-body-leading']})`,
     marginInlineEnd: `calc(${spacingVars['--spacing-1']} * -1)`,
   },
 });
--- a/packages/core/src/Toast/ToastViewport.tsx
+++ b/packages/core/src/Toast/ToastViewport.tsx
@@ -2,6 +2,18 @@
 
 'use client';
 
+/**
+ * @file ToastViewport.tsx
+ * @input Uses React state/effects, ToastContext, useAnnounce, viewport tokens,
+ *   and placement-derived motion variables
+ * @output Exports the ToastViewport provider, stack, live announcement dispatch,
+ *   focus handoff, safe-area-aware edge gutters, and motion context
+ * @position Core provider/imperative viewport for useToast()
+ *
+ * SYNC: When placement, stacking, focus, announcement, or safe-area behavior
+ *   changes, update ToastViewport.test.tsx, Toast.doc.mjs, and Toast.stories.tsx.
+ */
+
 import {
   isValidElement,
   useCallback,
@@ -22,13 +34,30 @@ import {ToastContext, type ToastContextValue} from './ToastContext';
 import type {ToastEntry, ToastPosition, ToastDismissReason} from './types';
 import {useTranslator} from '../i18n';
 
+const SAFE_AREA_INLINE_START = `max(${spacingVars['--spacing-4']}, env(safe-area-inset-left, 0px))`;
+const SAFE_AREA_INLINE_END = `max(${spacingVars['--spacing-4']}, env(safe-area-inset-right, 0px))`;
+const SAFE_AREA_BLOCK_START = `max(${spacingVars['--spacing-4']}, env(safe-area-inset-top, 0px))`;
+const SAFE_AREA_BLOCK_END = `max(${spacingVars['--spacing-4']}, env(safe-area-inset-bottom, 0px))`;
+const TOAST_EDGE_DRIFT = spacingVars['--spacing-2'];
+const TOAST_EDGE_DRIFT_NEGATIVE = `calc(-1 * ${TOAST_EDGE_DRIFT})`;
+
 const styles = stylex.create({
   viewport: {
     position: 'fixed',
     zIndex: 500,
     display: 'flex',
+    boxSizing: 'border-box',
     flexDirection: 'column',
-    padding: spacingVars['--spacing-4'],
+    paddingBlockStart: SAFE_AREA_BLOCK_START,
+    paddingBlockEnd: SAFE_AREA_BLOCK_END,
+    paddingInlineStart: {
+      default: SAFE_AREA_INLINE_START,
+      ':is([dir="rtl"] *)': SAFE_AREA_INLINE_END,
+    },
+    paddingInlineEnd: {
+      default: SAFE_AREA_INLINE_END,
+      ':is([dir="rtl"] *)': SAFE_AREA_INLINE_START,
+    },
     pointerEvents: 'none',
     // Reset popover styles — the popover attribute puts us in the top
     // layer (above dialogs), but we don't want its default styles.
@@ -40,23 +69,28 @@ const styles = stylex.create({
     backgroundColor: 'transparent',
     overflow: 'visible',
   },
-  bottomEnd: {bottom: 0, insetInlineEnd: 0, alignItems: 'flex-end'},
-  bottomStart: {bottom: 0, insetInlineStart: 0, alignItems: 'flex-start'},
+  viewportInlineSpan: {
+    insetInlineStart: 0,
+    insetInlineEnd: 0,
+  },
+  bottomEnd: {bottom: 0, alignItems: 'flex-end'},
+  bottomStart: {bottom: 0, alignItems: 'flex-start'},
   topEnd: {
     top: 0,
-    insetInlineEnd: 0,
     alignItems: 'flex-end',
     flexDirection: 'column-reverse',
   },
   topStart: {
     top: 0,
-    insetInlineStart: 0,
     alignItems: 'flex-start',
     flexDirection: 'column-reverse',
   },
   toastWrapper: {
     pointerEvents: 'auto',
     display: 'grid',
+    width: '100%',
+    maxWidth: 400,
+    minWidth: 0,
     gridTemplateRows: '1fr',
     transitionProperty: 'grid-template-rows, padding',
     transitionDuration: {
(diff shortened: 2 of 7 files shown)


packages/core/src/Toast/ToastViewport.tsx
```diff
@@ -423,19 +500,24 @@ export function ToastViewport({
...
+            style: Object.keys(insetStyle).length > 0 ? insetStyle : undefined,
+          },
+        )}>
         {visibleToasts.map(entry => {
           const o = entry.options;
           const type = o.type ?? 'info';
-          const isAutoHide = o.isAutoHide ?? (type === 'error' ? false : true);
+          const hasEndContent = isRenderable(o.endContent);
+          const isAutoHide =
+            o.isAutoHide ?? (type === 'error' || hasEndContent ? false : true);
```
cixzhang: Any toast with `endContent` is persistent now. Why was the old default wrong?

packages/core/src/Toast/ToastViewport.tsx
```diff
@@ -340,7 +381,22 @@ export function ToastViewport({
...
+      ? isCompactViewport
+      : availableInlineSize <= 640;
+  const effectiveMaxVisible =
+    maxVisible ?? (isCompact ? COMPACT_MAX_VISIBLE : DEFAULT_MAX_VISIBLE);
+  const isCompactQueue = maxVisible == null && isCompact;
+  const visibleToasts = isCompactQueue
+    ? toasts.slice(0, effectiveMaxVisible)
+    : toasts.slice(-effectiveMaxVisible);
+
+  useEffect(() => {
```
cixzhang: Hmm, was the `addToast` announce not enough? The comment you deleted said why it lived there.

packages/core/src/Toast/Toast.tsx
```diff
@@ -17,65 +31,163 @@ import {
...
     fontFamily: typographyVars['--font-family-body'],
     fontSize: typeScaleDefaults['--text-body-size'],
     lineHeight: typeScaleDefaults['--text-body-leading'],
-    transform: 'translateY(0)',
+    '--_toast-enter-y': TOAST_EDGE_DRIFT,
+    '--_toast-exit-y': 'var(--_toast-enter-y)',
+    transform:
+      'translateX(0) translateY(var(--toast-swipe-y, 0px)) scale(var(--_toast-swipe-scale, 1))',
+    transformOrigin: 'bottom right',
+    touchAction: 'none',
```
cixzhang: BottomSheet keeps this on the handle and does the handoff in a non-passive touchmove — `useSheetGestures.ts:1186`.

cixzhang (commented):
Thanks — the safe-area and wrapping work is the right fix and I'd take that half today. Commenting rather than blocking only because it's a draft.

Six runtime changes ship here and only the safe-area one has a problem written down. These four each change behaviour for every existing consumer, and I couldn't find the sentence saying what was wrong before:

- one toast at a time below 640px, the rest queued (`ToastViewport.tsx:390`)
- `isAutoHide` now false whenever `endContent` is set (`:520`)
- swipe-to-dismiss, and `'swipe'` added to `ToastDismissReason` (`Toast.tsx:100`, `types.ts:15`)
- the `addToast` announce moved into an effect (`:395`)

Two I'd want fixed wherever they land: at 375px a persistent toast never leaves the head of the queue — five fired, one shown at 20s, main shows three — and `touch-action: none` kills page scroll under the toast.

Could you write up the problem each of those four solves, and split out any that want their own discussion? Happy with several problems in one PR — I just can't weigh a change I can't see the reason for.

[Reviewed by Robohands]

packages/core/src/Toast/ToastViewport.tsx
rubyycheung: Agreed — `endContent` does not reliably mean actionable, and changing its default would affect every existing consumer. I restored the previous auto-hide behavior in b889b4c8852. Actionable examples now set `isAutoHide: false` explicitly, and the PR rationale calls this out as a deliberate non-goal.

packages/core/src/Toast/ToastViewport.tsx
rubyycheung: Agreed. Moving this into an effect was coupled to the automatic compact queue, but it weakened the existing exactly-once dispatch rationale. I restored announcement to `addToast` and documented why it stays outside the state updater. The queue/announcement policy is now separated into #5376.

packages/core/src/Toast/Toast.tsx
rubyycheung: Agreed. Swipe is removed from #5353. The stacked follow-up #5375 uses a non-passive `touchmove` handoff modeled on BottomSheet: native scrolling remains available until dominant travel resolves toward the dismiss edge. It also reports the existing `manual` reason rather than expanding the public union.

rubyycheung:
Addressed the review by splitting policy and gesture changes from this PR:

- #5353 now contains only responsive fit, safe areas, wrapping/alignment, spacing, edge-aware CSS motion, smooth stack collapse, and live-region cleanup. Its description now starts with the concrete problems and rationale.
- Restored the five-visible default, `endContent` auto-hide behavior, event-driven `addToast` announcements, and the existing dismissal-reason union.
- #5375 is the draft swipe-only follow-up, with non-passive touch handoff so page scrolling is preserved until edge-directed intent is established.
- #5376 captures compact queueing as a policy decision, including persistent-head starvation, ordering, timer start, announcement timing, deduplication, focus, and bounded growth.

Current validation for #5353: 28 targeted tests, repo checks, Core and Storybook typechecks, targeted lint, and Storybook build all pass.

imdreamrunner:
Hi @rubyycheung — I have an iOS 26 simulator rig set up from #5342 (same WebKit behaviour bit BottomSheet), so I ran your latest head through it to lend a hand. Sharing what it measured, plus a one-line change that came out clean, in case it saves you a round of guess-and-check. Your read on this was right — I think it's just the last mile.

**Where it stands at `42f740a0`:** the bottom toolbar still picks up the toast's dark colour, **7 of 7 runs**.

iOS 26 Safari toolbar tint, as-is vs overflow hidden restored

*iPhone 17 Pro, iOS 26.5 simulator, real WebKit, built from this branch. Left: as-is. Right: same build with one property changed.*

### What Safari is actually doing

This one is genuinely obscure, so in case it's useful background for anyone following along:

Safari's toolbars on iOS are translucent, and Safari tints them to match your page. Until iOS 18 you controlled that with `<meta name="theme-color">`. **iOS 26 ignores that tag** and inspects the page instead:

1. it picks a point **4px inside the bottom edge of the viewport**;
2. it asks "which element is here?";
3. it walks up from there to the nearest `position: fixed`/`sticky` ancestor;
4. it paints the toolbar with that ancestor's `background-color`.

The trap is step 2. That probe is not a normal "what's under the cursor" hit test — it deliberately **ignores clipping**, and an element with a CSS `transform` is matched through a path where the only thing that can rule it out is a **clipping ancestor**. Our toast has a transform (the entry/exit motion in this PR), so with nothing clipping it, it answers a probe that lands well outside it. Safari walks up, finds `--color-background-inverted`, and the toolbar goes dark.

Two things fall out of that, both of which cost me time:

- **Distance doesn't help.** I pushed the toast to 40px of clearance from the edge and it still tinted.
- **`overflow: hidden` on an ancestor is the thing that works** — it's the one signal that probe respects.

### On the two approaches so far

`overflow: hidden` in `65b0fbff` was the right instinct. What undid it was the expansion sitting next to it:

```js
paddingBlock: TOAST_EDGE_DRIFT,
marginBlock: TOAST_EDGE_DRIFT_NEGATIVE,
```

Those grow the clip box 24px downward. On this device the toast's bottom edge sits *exactly* 24px above the viewport edge (16px viewport gutter + 8px wrapper gap), so the expansion consumes the whole gap and lands the clip boundary flush with the edge — putting the probe point back inside the clip box. That's a brutal coincidence to debug, and it made the bug intermittent rather than constant (I measured 7/10 on that commit), which is worse than it failing outright. Worth knowing the 24px isn't tunable either: it's `TOAST_EDGE_DRIFT` on both sides, and the gutter is `max(16px, env(safe-area-inset-bottom))`, which varies by device.

`49256837` then moved sizing into a new wrapper div in `Toast.tsx` and set `overflow` back to `visible`. That wrapper doesn't clip, so the probe reaches the toast again — hence 7/7.

### The change that measured clean

```diff
   toastWrapperInner: {
     width: '100%',
     maxWidth: '100%',
     minWidth: 0,
     minHeight: 0,
-    overflow: 'visible',
+    overflow: 'hidden',
   },
```

**6 / 6 clean**, covering `bottomEnd`, `topEnd`, and a page with its own fixed bottom bar. One assertion in `ToastViewport.test.tsx` flips from `'visible'` to `'hidden'`; the rest of the Toast suite passes untouched. Worth a comment on the line, since it now does two jobs at once — it's also what makes the `1fr → 0fr` grid-track collapse visible on removal, which is probably why `main` had it.

### The honest trade-off

Clipping costs the 24px entry/exit drift — the motion reads as a fade plus a slide inside a window rather than a free slide. That matches `main`'s behaviour today, so it isn't a regression, but I realise the free slide is exactly what the expansion was trying to buy back, and losing it is a real cost to this PR's goal.

If you want that motion, the constraint to design against is: **nothing painting the toast's colour may come within ~4px of the viewport edge at any point during the animation.** Expanding the clip toward the edge can't satisfy that, and moving the drift onto the wrapper hits the same wall (the clip box travels edge-ward with it). Animating `opacity` + `scale` with no translation would, and would keep an edge-aware feel via `--_toast-origin`. Happy to prototype that on the simulator if it's worth exploring — it's about a 3-minute turnaround per variant on the rig now.

Also flagging: this affects placement generally, so if `topEnd`/`topStart` matter for your keyboard-heavy flows, the same reasoning applies to the top edge.

Glad to push the one-liner to your branch, or leave it with you — whichever you prefer. And if you want to run the rig yourself, `BottomSheetEdgeTint.tsx` has the neighbouring case in its header (there we *want* to set the toolbar colour deliberately), and I'm happy to walk you through the simulator setup.

imdreamrunner:
Verified `28de834a` on the iOS 26.5 simulator — **the tint is fixed.** 10/10 clean runs, your tests 33/33. 🎉

One catch: **iOS Safari doesn't support `overflow-clip-margin`.** It's dropped at parse time, so what actually fixes the tint is the plain `overflow: clip` sitting next to it. Chromium *does* honour it, which switches the clip off on desktop.

## What I'd change

```diff
   toastWrapperInner: {
     overflow: 'clip',
-    overflowClipMargin: '100vh',
   },
```

1. **Drop `overflowClipMargin`** — no effect on iOS; on desktop it disables the clip that makes the removal collapse look right.
2. **Update one assertion** in `ToastViewport.test.tsx` — `'100vh'` → `''`. Everything else passes.
3. **Add a line comment** on `overflow: 'clip'` — it does two jobs (the grid-collapse clip *and* keeping Safari's edge sampler off the toast), so it looks removable otherwise. That's how `main` lost it.

Verified 6/6 clean with that change — `bottomEnd`, `topEnd`, and a page with a fixed bottom bar.

## One thing to decide (no action needed)

Dropping the margin leaves the 24px entry/exit drift clipped on desktop — matching `main`, and matching what iOS already does today. If you want that motion back, the constraint is *nothing painting the toast's colour may come within ~4px of the viewport edge during the animation*; `opacity` + `scale` with no translation satisfies it and keeps an edge-aware feel via `--_toast-origin`. Happy to prototype it — a couple of minutes per variant on the rig.

<details>
<summary>Evidence, if you want it</summary>

**iOS doesn't implement the property:**

```
CSS.supports('overflow-clip-margin', '100vh')  ->  false
CSS.supports('overflow-clip-margin', '16px')   ->  false
getComputedStyle(host).overflowClipMargin      ->  ""
```

Confirmed by painting, not just the API — three boxes, each with a child pushed 70px past its parent's bottom edge:

| box | expected | painted outside |
|---|---|---|
| `overflow: visible` (control) | must paint | **18000 px** ✅ test works |
| `clip` + `margin: 100vh` | paints *if* margin applies | **0** |
| `clip` (control) | must not paint | **0** |

Chromium computes the margin to `800px` on an 800px viewport.

**Desktop side effect** — ~250ms into a removal, both transitions slowed 10× proportionally:

desktop mid-removal: with clip margin vs without

Left: as-is, the fading toast bleeds past its collapsing row. Right: without the margin. 7951 differing pixels, confined to the bottom 21px. Subtle — the fade largely covers it, and at `bottomEnd` most of it falls off-screen.

**Durability:** as written, the fix depends on Safari *not* implementing a property. If WebKit ships it, the clip box grows past the viewport edge and the tint returns.

*(Safari 26 freezes its UA at "iPhone OS 18_7" — the runtime here is genuinely iOS 26.5, build 23F77.)*

</details>

Want me to push the one-liner to your branch?

imdreamrunner (approved):

imdreamrunner:
⚠️ Heads-up before this merges (it's approved): **`1c0b74a8` brings the tint back** — 6/8 runs on iOS 26.5. The `overflow: clip` you had at `28de834a` was working; the new `calc(100lvh - 100dvh)` approach replaces it with something that doesn't hold.

as-is vs overflow:clip restored

Interleaved A/B, same build, only that one property differing:

| | tinted |
|---|---|
| `1c0b74a8` as-is | **6 / 8** |
| same build, `overflow: clip` restored | **0 / 8** |

(Plus 5/11 in an earlier batch — 11/19 overall.)

## What I'd do

1. **Restore `overflow: 'clip'`** on `toastWrapperInner` (revert it from `visible`).
2. **Drop `DYNAMIC_BROWSER_CHROME_OFFSET`** and the two `marginBlockEnd` uses — see below, it causes a second problem.
3. **Add a line comment** on the `overflow` so it doesn't get removed again.

That's `28de834a` minus `overflowClipMargin`. Verified 0/8 here, and 6/6 earlier across `bottomEnd`, `topEnd`, and a bottom-bar page.

## Why moving the toast up can't work

This is the counterintuitive bit, and it's why the approach is worth abandoning rather than tuning: **distance from the edge is not what Safari checks.** The probe runs with `IgnoreClipping`, and a *transformed* layer (our toast, because of the entry/exit motion) is matched through a path where the only thing that can rule it out is a **clipping ancestor**. Empirically: 24px → tints, 40px → tints, and now 64px (what the offset buys) → tints 6/8. Only the clip stops it.

## Second problem with the offset

`calc(100lvh - 100dvh)` is the address bar's height, and that collapses to `0` when the bar retracts. Measured on device while scrolling:

```
address bar shown     offset = 40px    toast sits 40px above the screen edge
address bar retracted offset =  0px    toast sits flush at the edge
```

So the toast slides 40px against the screen edge as the user scrolls — and it lands flush at the edge exactly when the offset is gone, which is the worst position for the tint.

<details>
<summary>Rig details</summary>

iPhone 17 Pro, iOS 26.5 (build 23F77), real WebKit, `@astryxdesign/core` built from the branch. Each run: fresh Safari, load, toast after a delay, screenshot, sample the toolbar band at x=6% of width (dodges the address-bar pill). Runs interleaved A/B/A/B so session drift can't favour either arm. The A/B override is applied as a stylesheet on the *same* build, so nothing but that one property differs.

</details>

Happy to just push the change if that's easier — it's three lines and I've got it verified locally.

rubyycheung:
@imdreamrunner Thanks again for running this. One important correction to the acceptance criterion: `overflow: clip` does stop the **dark Toast tint**, but in our iPhone 17 Pro Simulator it leaves Safari’s bottom chrome **opaque white** instead of preserving the original translucent appearance.

image

So I don’t think we can treat the clipping version as the complete fix—it trades the dark tint for white browser chrome. What we need is for opening the Toast to leave Safari’s browser chrome visually unchanged, including its translucency, rather than choosing a different toolbar color.

Could you compare this aspect on your rig as well? We had been measuring “not dark” and missed the opacity/background regression.

PR #5362, opened 2026-08-23
fix(docsite): default changelog to core, sync tab with URL, fill empty releases

HelloOjasMutreja:
Addresses #5307, partially. Three of the four bullets are done; the fourth needed a decision I didn't want to guess at, so I left it out (details below).

## Default to core, not whatever sorts first

`ChangelogView` picked `changelogs[0]?.pkg` as the initial tab, and `packages` (from `packageRegistry.ts`) is built by scanning package directories, so the first entry is whatever sorts first alphabetically — `@astryxdesign/cli`, not `@astryxdesign/core`. Now it prefers `@astryxdesign/core`, falling back to the first entry only if core has no changelog.

## Direct links via `?package=`

`activeTab` is now derived straight from `searchParams.get('package')` (clamped to a package that actually has a changelog, so a stale or hand-edited value can't land on a blank panel), the same pattern `ComponentDetailClient` already uses for its own tab state (`searchParams.get('tab')`, `router.replace(..., {scroll: false})` on change, no separate `useState`). The URL only gets a `?package=` param when it differs from the default, so the common case stays a clean `/changelog`.

## Empty releases

A release with no notable changes still gets a version heading from the changeset tooling, with nothing between it and the next heading or the `---` separator before it (confirmed against several real `CHANGELOG.md` files, e.g. `packages/themes/y2k/CHANGELOG.md`'s `0.4.6`/`0.4.5`/`0.4.2` entries). `fillEmptyReleases()` in `changelogLinkify.ts` fills those bodies with "No changes in this release." Added unit tests covering: a mid-file empty version, the last version in the file (no trailing separator), a version with real content staying untouched, and the stripped package title not being mistaken for an empty version.

## What I left out: "keep theme packages out of primary tabs"

I didn't implement this one. Checked `ThemePackagePage.tsx` first, since "keep out of primary tabs" implies they'd still be reachable somewhere else, and it turns out theme package changelogs aren't shown anywhere but this page today. Filtering them out of `ChangelogView`'s tabs would delete the only place they're visible, not relocate them. That's a real UI decision (a secondary tab group, a dropdown, moving them onto the theme pages, or something else) rather than a one-line filter, so I'd rather it get scoped properly than have me guess. Happy to pick it up once there's a direction, or if it's more of a "just filter them, changelog visibility for themes isn't a requirement" call, say so and I'll do that instead.

## Verification

`fillEmptyReleases` unit tests: 20/20 pass (`pnpm -F @astryxdesign/docsite test src/__tests__/changelog-linkify.test.ts`). Typecheck (`tsc --noEmit`) shows no new errors from these files (checked by grepping the full output for "changelog" — all pre-existing errors are elsewhere in the app, unrelated to this change). Lint is clean.

Not verified: an actual browser render of the page. `next dev` here needs a full data regeneration pass I didn't want to eat the time on, given the tab logic mirrors an already-shipped, working pattern (`ComponentDetailClient`) line-for-line and the markdown transform is covered by tests against real changelog content. If that's not enough for review, let me know and I'll get a real render.

### DIFF
--- a/apps/docsite/src/__tests__/changelog-linkify.test.ts
+++ b/apps/docsite/src/__tests__/changelog-linkify.test.ts
@@ -12,6 +12,7 @@ import {
   linkifyContributors,
   linkifyComponents,
   stripTitle,
+  fillEmptyReleases,
 } from '../components/changelogLinkify';
 
 describe('linkifyPRs', () => {
@@ -133,6 +134,69 @@ describe('linkifyComponents', () => {
 describe('stripTitle', () => {
   it('removes the leading h1 title line', () => {
     expect(stripTitle('# @astryxdesign/core\n\n# 0.1.1\n')).toBe('# 0.1.1\n');
+  });
+});
+
+describe('fillEmptyReleases', () => {
+  it('fills a version whose only body is the separator to the next heading', () => {
+    const md = [
+      '# 0.4.6',
+      '',
+      '---',
+      '',
+      '# 0.4.5',
+      '',
+      '#### Fixes',
+      '',
+      '- A real fix.',
+    ].join('\n');
+    expect(fillEmptyReleases(md)).toBe(
+      [
+        '# 0.4.6',
+        '',
+        'No changes in this release.',
+        '',
+        '---',
+        '',
+        '# 0.4.5',
+        '',
+        '#### Fixes',
+        '',
+        '- A real fix.',
+      ].join('\n'),
+    );
+  });
+
+  it('fills the last version in the file, which has no trailing separator', () => {
+    expect(fillEmptyReleases('# 0.4.2')).toBe(
+      '# 0.4.2\n\nNo changes in this release.\n\n',
+    );
+  });
+
+  it('leaves a version with real content untouched', () => {
+    const md = [
+      '# 0.4.3',
+      '',
+      '#### Fixes',
+      '',
+      '- A real fix.',
+      '',
+      '---',
+      '',
+      '# 0.4.2',
+      '',
+      '#### Fixes',
+      '',
+      '- Another real fix.',
+    ].join('\n');
+    expect(fillEmptyReleases(md)).toBe(md);
+  });
+
+  it('does not mistake the stripped package title for an empty version', () => {
+    // fillEmptyReleases is documented to run after stripTitle, so the h1
+    // package-name line should already be gone by the time it sees this.
+    const md = ['# 0.4.6', '', '#### Fixes', '', '- A fix.'].join('\n');
+    expect(fillEmptyReleases(md)).toBe(md);
   });
 });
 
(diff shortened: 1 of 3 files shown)


HelloOjasMutreja:
The Vercel deployment check is failing, but I don't think it's caused by this diff. This PR touches exactly 3 files, all in `apps/docsite/src/components` and `apps/docsite/src/__tests__` (`ChangelogView.tsx`, `changelogLinkify.ts`, `changelog-linkify.test.ts`).

Ran `next build --webpack` for the docsite locally (after a fresh `packages/core` build, since a stale `core/dist` gave me an unrelated `Module not found: @astryxdesign/core/BottomSheet` locally at first) and it completes cleanly: 325 static pages generated, `/changelog` prerenders as `○ (Static)` with no errors. `docsite-test` and `lint` are also green on this PR's own CI run.

I don't have Vercel CLI credentials in this environment to pull the actual deployment logs (`npx vercel inspect` needs a login), so I can't see exactly what failed on their end. Given the clean local build against the same diff, my guess is either a transient Vercel-side issue or a stale build cache on their infrastructure, rather than something in this change. Happy to look closer if there's a way to get the deployment logs, or if a re-run clears it.

HelloOjasMutreja:
Visual evidence, since I skipped it earlier and shouldn't have — both changes are directly visible on the page.

## Default tab

Before, landing on `/changelog` selected `@astryxdesign/cli` (first alphabetically). After, it selects `@astryxdesign/core`.

**Before:**
before default

**After:**
after default

The before screenshot also happens to catch the second bug in the same frame: a bare `0.4.7` heading with nothing under it.

## Empty releases

`@astryxdesign/theme-y2k`'s changelog has five back-to-back empty versions. Before, each was just a heading and a divider. After, each gets an explicit "No changes in this release." line.

**Before:**
before empty release

**After:**
after empty release

Both captured against a real `next build && next start` of this branch (Playwright, since the local browser tooling I have here doesn't screenshot reliably), toggling between the pre-fix and post-fix source with the rest of the tree held constant. The "after" empty-release screenshot's URL also confirms the third change: `?package=%40astryxdesign%2Ftheme-y2k` appears after switching tabs, so the `?package=` sync is live too.

rubyycheung (changes requested):
Defaulting the changelog to `@astryxdesign/core` makes sense, but the tab order should reinforce that hierarchy. This change selects Core while leaving `@astryxdesign/cli` as the first tab, so the page opens with a lower item active and the navigation still visually presents CLI as primary.

Please also move Core to the first tab position—Core first, CLI second, followed by the remaining packages—so the displayed order and default selection communicate the same priority. The URL synchronization and explicit empty-release message otherwise look good.

nynexman4464:
Closing in favor of #5309, which implements the same changelog fixes and now reflects the resolved direction: Core first, CLI second, with theme package changelogs omitted. I also brought over the additional empty-release edge-case tests from this PR. Thanks for the careful implementation and visual verification.

PR #5363, opened 2026-08-23
fix(Table): a plugin can suppress a body cell's content, so group headers never reach your cell renderers

cixzhang:
## What

A table with grouped rows crashes as soon as one of its cell renderers keys a lookup off a field:

```tsx
renderCell: item => <StatusDot status={STATUS_META[item.status].dot} />
```

`useTableGroupedRows` injects synthetic section-header rows into the flattened data, and `BaseTable` evaluates every column's `renderCell` against every row before `transformBodyRow` gets to discard a header's cells. The header Proxy answering unknown fields with `''` only ever rescued a renderer that *prints* a field; `''` is no more a member of that map than `undefined` is, so the renderer throws and the page blanks.

This supersedes [#5317](https://github.com/facebook/astryx/pull/5317). Everything above is @ernestt's report and his diagnosis — he found the bug, named the mechanism, and pointed at the line. This PR takes the fix one layer down.

## Evidence

The repro above, `plugins={{grouped, rowStatus}}` — same story, same viewport.

| Before — `main` | After — this PR |
| --- | --- |
|  |  |
| The synthetic group-header row reaches `renderCell`, the lookup returns `undefined`, nothing renders. | Headers, rows and status dots; the header rows' cells are empty. |

The three existing grouped-rows stories are **byte-identical** before and after — same `sha256` for every PNG, and the same rendered table HTML:
[Default](https://raw.githubusercontent.com/cixzhang/astryx/assets/pr-5363/grouped-default.png) ·
[Custom order and header](https://raw.githubusercontent.com/cixzhang/astryx/assets/pr-5363/grouped-custom-order-and-header.png) ·
[Initially collapsed](https://raw.githubusercontent.com/cixzhang/astryx/assets/pr-5363/grouped-initially-collapsed.png)

## Why not wrap the columns

[#5317](https://github.com/facebook/astryx/pull/5317) fixes this in `transformColumns`, wrapping each column so `renderCell` short-circuits on a header row. That only sees the columns that exist *when it runs*, and `grouped` has no entry in `PLUGIN_ORDER`, so its position in the pipeline is whatever order the consumer happened to type the plugins object in:

```tsx
plugins={{grouped, rowStatus}}   // still crashes: the status column is injected after the wrap
plugins={{rowStatus, grouped}}   // fine
```

`useTableRowStatus`, `useTableSelection`, `useTableRowIndex`, `useTableRowExpansion` and `useTableFiltering` all contribute columns of their own, and each one's `renderCell` calls back into consumer code. Nothing warns; you get one order that works and one that doesn't.

Adding `grouped` to `PLUGIN_ORDER` would make the reported case pass and leave order deciding correctness. **Plugin order may affect presentation, never correctness** — that principle is now written at the line someone would edit to do it.

## How

`BodyCellRenderProps` gains one optional field:

```ts
isContentSuppressed?: boolean;
```

`BaseTable` honors it before the renderer runs — with it set, the cell renders empty and neither `col.renderCell` nor the default renderer is called. The grouped plugin sets it in `transformBodyCell` when the row is a header:

```tsx
transformBodyCell(props, _column, item) {
  if (!isGroupHeader(item)) return props;
  return {...props, isContentSuppressed: true};
}
```

`transformBodyCell` was already the right hook: it runs per cell, at render time, against the final column list — so it covers columns other plugins contributed, whatever order the plugins sit in. It just had no way to say "no content here". `BodyCellRenderProps` carried only `htmlProps`, `xstyle`, `contextMenuActions`, `columnIndex` and `columns`.

**On the name.** The header-cell siblings are content slots (`before`, `content`, `after`, `overlay`, `below`), and reusing `content` here was tempting — but a body cell's content does not exist yet when the hook runs (producing it is the thing we must avoid), so a `content` slot would invite plugins to wrap existing content and hand them `undefined` instead. That leaves a boolean, and the repo's boolean-prop rule wants an `is`/`has` prefix (`isHeaderRow` is the neighbour). `isContentSuppressed` names the outcome — this cell has no content — rather than the mechanism, which is what a plugin author needs to reason about. It is deliberately one field, not a family: it solves the class of *any plugin suppressing any cell's content*, and a summary row or a spanning row would reach for exactly the same thing.

Not carried over from [#5317](https://github.com/facebook/astryx/pull/5317): the `transformColumns` wrapper and the `WeakMap` that cached wrapped columns. Nothing rewrites the column array now, so there is nothing to cache — but the invariant that `WeakMap` was protecting is real, and `useTableGroupedRows-perf.test.tsx` locks it by render count instead (a re-render of the surrounding component re-renders no rows; it fails if the plugin allocates a column per render).

The Proxy stays: sorting, filtering, row keys and consumer handlers all still read fields off a header row.

## Also

The `TablePlugin` pipeline docblock had both row/cell pairs backwards — it listed `transformHeaderRow` before `transformHeaderCell` and `transformBodyRow` before `transformBodyCell`, when in both cases the cells are built first and reach the row transform as `children`. Reading it, the column transform looks like the only hook that runs before the renderer, which is plausibly how the first fix ended up there. Corrected, with one line on what the ordering means for a plugin author.

## Test plan

- [x] A column renderer that keys a lookup off a field no longer throws under grouping, and only the real rows reach it.
- [x] The same, with a column contributed by `useTableRowStatus`, run in **both** plugin orders — `{{grouped, rowStatus}}` and `{{rowStatus, grouped}}`. Both red before this change; `{{grouped, rowStatus}}` stays red with [#5317](https://github.com/facebook/astryx/pull/5317)'s approach applied.
- [x] `isContentSuppressed` at the `BaseTable` level: an empty cell, no call to the column renderer, and the default renderer suppressed too.
- [x] `useTableGroupedRows-perf.test.tsx`: three real rows render once each, no header row reaches the renderer, and an unrelated re-render of the surrounding component re-renders no rows. Verified to fail if the plugin allocates a fresh column object per render.
- [x] Full Table suite: 483 passed (23 files). `tsc --noEmit`, `eslint` and `prettier --check` clean on everything touched.
- [x] The three grouped-rows stories render **byte-identically** before and after — identical PNGs and identical rendered table HTML, in Chromium.

Not addressed: `aria-rowcount` / `aria-rowindex` still drift when headers are present, as [#5317](https://github.com/facebook/astryx/pull/5317) noted.

### DIFF
--- a/.changeset/table-suppress-cell-content.md
+++ b/.changeset/table-suppress-cell-content.md
@@ -19,8 +19,4 @@ column's renderer or the default one. It is decided per cell at render time
 against the final column list, so it also covers columns other plugins
 contributed — whatever order the plugins were listed in.
 
-`useTableGroupedRows` also returns `isGroupHeader`, so a row-level plugin or
-handler of your own (click-to-open-detail, row links, per-row menus) can tell a
-synthetic header from a real row without matching on the `__group_` key prefix.
-
 @ernestt
--- a/.changeset/table-suppress-cell-content.md
+++ b/.changeset/table-suppress-cell-content.md
@@ -1,0 +1,26 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Table: a plugin can suppress a body cell's content, and grouped rows use it to keep synthetic group headers out of your cell renderers
+
+`useTableGroupedRows` injects section-header rows into the flattened data, and
+`BaseTable` evaluates every column's `renderCell` against every row. A renderer
+that keys a lookup off a field — `STATUS_META[item.status].dot` — was therefore
+handed a row that is not yours and threw, blanking the page the moment grouping
+was switched on. The header Proxy answering unknown fields with `''` only ever
+rescued a renderer that _prints_ a field; `''` fails a lookup exactly as
+`undefined` does.
+
+`BodyCellRenderProps` gains `isContentSuppressed?: boolean`. A plugin sets it in
+`transformBodyCell` for a row whose cells it is about to replace wholesale in
+`transformBodyRow`, and the table renders that cell empty without calling the
+column's renderer or the default one. It is decided per cell at render time
+against the final column list, so it also covers columns other plugins
+contributed — whatever order the plugins were listed in.
+
+`useTableGroupedRows` also returns `isGroupHeader`, so a row-level plugin or
+handler of your own (click-to-open-detail, row links, per-row menus) can tell a
+synthetic header from a real row without matching on the `__group_` key prefix.
+
+@ernestt
--- a/.changeset/table-suppress-cell-content.md
+++ b/.changeset/table-suppress-cell-content.md
@@ -2,7 +2,7 @@
 '@astryxdesign/core': patch
 ---
 
-[fix] Table: a plugin can suppress a body cell's content, and grouped rows use it to keep synthetic group headers out of your cell renderers
+[fix] Table: a plugin can suppress a body cell's content, and grouped rows use it to keep synthetic group headers out of your cell renderers (#5363)
 
 `useTableGroupedRows` injects section-header rows into the flattened data, and
 `BaseTable` evaluates every column's `renderCell` against every row. A renderer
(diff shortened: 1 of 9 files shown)


cixzhang (commented):
Approve (GitHub won't take a self-approval, so this is a comment).

Reproduced the crash on main and confirmed it's gone here in both plugin orders; the three grouped stories are pixel-identical. Only new surface is `isContentSuppressed`.

[Reviewed by Robohands]

PR #5365, opened 2026-08-23
test(rtl): detect contextual decorations and coverage gaps

### LINKED ISSUE #5364: [audit] RTL audit: classify contextual decorations and unexplained N/A coverage
## Problem

The RTL audit can report a component as entirely N/A even when it contains direction-sensitive behavior that none of the current dimensions measures. Breadcrumbs exposed the gap: contextual text separators are not SVG icons, positional elements, order controls, scroll behavior, or overlays, so a real separator-direction defect produced `0 pass / 0 fail / all N-A`.

The weekly report currently calls a zero-finding run clean without distinguishing a proven non-applicable component from an unmeasured one. That can hide the same class of gap in both existing and newly added components.

## Proposed contract

Every component in the live Core and Lab source roster must finish in one of three states:

1. **Measured:** at least one applicable RTL dimension ran and passed or failed.
2. **Verified N/A:** the checked-in applicability registry records why the component has no direction-sensitive behavior.
3. **Coverage gap:** every dimension was N/A and no verified-N/A reason exists.

Coverage gaps are findings, not clean results. This applies to the full existing roster and automatically to new components.

## Directional decorations

Add an auto-discovered dimension for decorative text glyphs in directional contexts, such as an `aria-hidden` separator inside repeated list items. Context matters: `/` in prose is neutral, while `/` between breadcrumb items communicates progression and should mirror.

The dimension should distinguish:

- contextual `/`, `\\`, `→`, `←`, `>`, and `<`, which require one explicit mirror or a glyph swap;
- `›`, `‹`, `»`, and `«`, which the Unicode bidi algorithm mirrors automatically and must not also receive an explicit mirror;
- unrelated text containing the same characters, which must not be classified.

## Night Watch and weekly reporting

- The weekly summary reports coverage gaps separately from RTL behavior failures and opens or updates the existing system-level tracking issue.
- The Component Auditor treats an all-N/A result as unverified until it either adds a real dimension/target or records a specific verified-N/A reason.
- `aria-current` checks read the DOM attribute and validate the computed or implicit role separately. Chromium's CDP accessibility tree does not reliably expose the current state, so its omission there is not a failure by itself.

## Acceptance criteria

- A breadcrumb-style slash is detected; a slash in prose is ignored.
- A bare contextual arrow fails; a mirrored arrow passes.
- A bare Unicode-mirrored angle quote passes; explicitly mirroring it fails as double handling.
- Every current and future component discovered from the Core and Lab source roster appears in the measured / verified-N-A / coverage-gap aggregate, including components with no matching story.
- A checked-in verified-N/A declaration that later gains applicable RTL behavior is reported as stale.
- PR and weekly summaries never describe an unexplained all-N/A component as RTL-clean.
- The Night Watch Component Auditor and rubric instructions match the executable contract.

Related finding: #5332.

nynexman4464:
## Problem

The RTL audit treated an all-N/A result as clean and only auto-detected directional SVG icons. That left contextual text decorations unmeasured: Breadcrumbs could ship an unmirrored separator while reporting `0 pass / 0 fail / all N-A`.

The audit roster also came only from rendered stories, so a Core or Lab component without a matching story could disappear instead of surfacing as a coverage gap.

## Changes

- Add D6 auto-discovery for single-glyph, `aria-hidden` decorations in repeated-item or between-sibling contexts.
- Distinguish explicit-mirror glyphs (`/`, `\\`, arrows, angle brackets) from Unicode bidi-mirrored angle quotes, and reject double handling.
- Build the applicability roster from all discoverable Core and Lab components, keyed as `core/Name` or `lab/Name`.
- Classify each component as measured, verified N/A, or a coverage gap. Report stale verified-N/A declarations when behavior becomes applicable later.
- Surface D6 and coverage gaps in PR summaries and the weekly RTL tracking issue.
- Mirror Breadcrumbs' built-in slash and add rendered slash, angle-quote, and arrow coverage.

Fixes #5364.

The companion monitor contract is live in [Night Watch Component Auditor](https://github.com/facebook/astryx/wiki/Night-Watch-Component-Auditor) and [Component Audit Rubric v1.8](https://github.com/facebook/astryx/wiki/Component-Audit-Rubric).

NOTE: Required visual acceptance is waiting on #5840. The trusted planner runs from `main`; after #5840 lands and CI reruns, this PR's review contracts from 252 frames to 16 focused frames.

## How D6 detects a problematic glyph

D6 does not search globally for characters such as `/`. A rendered element becomes a candidate only when all of these are true:

1. It is `aria-hidden="true"`. That tells us the text is decorative rather than content. This is intentional: `aria-hidden` removes the separator text from the accessibility tree, while `role="presentation"` on a `<span>` removes no meaningful role and still exposes `/` as text.
2. Its entire trimmed text is one supported glyph, not a sentence, date, URL, path, or fraction containing that character.
3. The DOM supplies a directional UI context: the decoration is inside one of at least two sibling `li`/`role="listitem"` items, or the decoration sits between two sibling elements.
4. The glyph is visibly rendered. Hidden first-item separators do not count.

That means a Breadcrumbs separator is measured, while `and/or`, `08/24`, `/settings`, and `/` in prose are ignored.

For each candidate, D6 renders the same story LTR and RTL, follows nested wrappers to the element that carries the glyph, and composes transforms through its ancestors. It then applies one of two rules:

- `/`, `\`, `→`, `←`, `>`, and `<` need explicit handling. They pass if the RTL transform is the horizontal mirror of the LTR transform, or if the DOM swaps to the paired opposite glyph. Neither handling fails; doing both fails as a double flip.
- `›`, `‹`, `»`, and `«` have Unicode `Bidi_Mirrored=Yes`. They pass when the DOM glyph and transform stay unchanged because the browser mirrors the painted glyph. Swapping or transforming them again fails as double handling.

D6 scans every story for the component, not only its default story. The component-level rollup then classifies each Core and Lab component as measured, verified N/A with a checked-in reason, or a coverage gap.

## Visual evidence

The existing Custom Separator story now renders both D6 strategies in one focused canvas:

- `›` relies on Unicode bidi mirroring and keeps its DOM glyph and transform unchanged;
- `→` uses one explicit horizontal mirror.

The story is tagged `visual-baseline`; the default Breadcrumbs story remains the representative theme surface. Against the live accepted baseline, the bounded planner produces 16 frames: 14 Default frames across seven themes and two modes, plus two neutral Custom Separator frames. Two already exist and 14 are new. All 16 captured successfully.

D6 still scans every Breadcrumbs story independently of screenshot selection. The rendered mutation test below is the precise regression proof: the old component fails 13 applicable stories, while the fixed component passes all 14.

### Rendered mutation proof

Before, restored only `packages/core/src/Breadcrumbs/BreadcrumbItem.tsx` from `HEAD^`, rebuilt Core and Storybook, then ran:

```bash
node apps/storybook/rtl-audit/rtl-audit.mjs \
  --storybook-dir apps/storybook/dist \
  --output /tmp/rtl-breadcrumbs-before.json \
  --filter Breadcrumbs \
  --concurrency 4
```

Result: exit 1; D6 reported 1 pass, 13 failures, and 2 N/A stories. The consolidated Custom Separator story passed because its two custom strategies are independent of the built-in slash; every applicable built-in-slash story failed as neither mirrored nor swapped.

After restoring `HEAD`, rebuilding, and running the same audit command with `/tmp/rtl-breadcrumbs-after-final.json`:

- exit 0;
- D6: 14 pass, 0 fail, 2 N/A;
- Custom Separator: pass with four candidates across its automatic and explicit strategies;
- `core/Breadcrumbs`: measured, 0 coverage gaps.

### Missing-story coverage

Ran:

```bash
node apps/storybook/rtl-audit/rtl-audit.mjs \
  --storybook-dir apps/storybook/dist \
  --output /tmp/rtl-breadcrumb-item-gap.json \
  --filter BreadcrumbItem \
  --concurrency 2
```

Result: exit 1 with `core/BreadcrumbItem` reported as a coverage gap instead of disappearing from the audit.

### Automated suites

Ran:

```bash
pnpm exec vitest run --project node \
  .github/scripts/visual-gate/lib/plan.test.mjs \
  .github/scripts/visual-gate/visual-acceptance.test.mjs \
  .github/scripts/rtl-audit-coverage.test.mjs \
  .github/scripts/weekly-rtl-summary.test.mjs
pnpm exec vitest run --project ui packages/core/src/Breadcrumbs/Breadcrumbs.test.tsx
pnpm lint:strict
pnpm -F @astryxdesign/storybook build
```

Result: 64 planner/audit/monitor tests and 40 Breadcrumbs tests passed. Strict lint, Core build, and Storybook build passed. The trusted 16-frame plan captured 16/16 successfully.


rubyycheung (approved):
Looks good from design. The Breadcrumb separator mirrors correctly in RTL while ordinary slashes in text, dates, and paths remain unchanged. The contextual detection and coverage checks are a good safeguard against future directional-decoration regressions.

ernestt (approved):

PR #5370, opened 2026-08-23
test(cli): lint assets/docs for our-process words shipped to callers

### LINKED ISSUE #5369: docs: lint assets/docs for our-process words shipped to callers
`packages/cli/assets/docs/` ships to people **building with** Astryx. Material about building Astryx itself — grading rubrics, promotion gates, review evidence — keeps landing there instead of the wiki, and it is only caught if a human happens to read the diff.

[#5368](https://github.com/facebook/astryx/pull/5368) added a README with the audience rule and a routing table. This is the automated half.

## What to check

Fail on these five words appearing in `assets/docs/**`:

```
rubric  readiness  promotion  sign-off  evidence
```

**Measured, not guessed:** 0 hits across the directory on `main` today, and **56** in [#5351](https://github.com/facebook/astryx/pull/5351) — the draft that prompted this. Clean signal, nothing to migrate first.

## What NOT to check

`audit`, `checklist` and `gate` were the obvious additions and all six of their current hits are innocent — migration's "Verification Checklist", "Audit every reset stylesheet". `lab` matches `label` and `available` 148 times. Adding them makes the check noise, and a noisy check gets suppressed.

## Where

Cheapest home is a case in `packages/cli/foundation/discovery/docs-discovery.test.mjs` — it already walks the directory.

The failure message should point at the README's routing table rather than just refusing, so the author knows which wiki page the material wants.

HelloOjasMutreja:
Fixes #5369.

## Problem

`packages/cli/assets/docs/` ships to people building **with** Astryx. Material about building Astryx itself (grading rubrics, promotion gates, review evidence) keeps landing there instead of the wiki, and it was only ever caught if a human happened to read the diff. #5368 added a README to that directory with the audience rule and a routing table; this is the automated half the issue asked for.

## What it checks

Added a case to `packages/cli/foundation/discovery/docs-discovery.test.mjs`, which already walks `assets/docs`, per the issue's suggestion. Fails on `rubric`, `readiness`, `promotion`, `sign-off`, `evidence` (case-insensitive, word-boundary) appearing in any `.mjs` file there.

Deliberately narrower than the README's own longer "tells" list. `audit`, `checklist`, and `gate` were the obvious additions, but checked first: all their current hits in the directory are innocent prose ("Verification Checklist", "Audit every reset stylesheet"), and `lab` collides with `label`/`available` 148 times. A noisy check gets suppressed rather than acted on, so I left those out.

Verified the five-word list is clean today (0 hits) and matches what the issue measured against the draft that prompted it (#5351, 56 hits) — didn't re-run that PR's diff since it's someone else's draft, took the issue's numbers as the baseline signal instead.

`.mjs`-only is deliberate too: it naturally excludes `README.md`, which legitimately contains all five words since it's the document *describing* the rule — filtering by extension is simpler than an explicit exclude-list and won't need updating if another non-`.mjs` file shows up in the directory later.

The failure message points at the README's routing table so the author knows which wiki page the material actually wants, per the issue's ask.

## Verification

Confirmed the test fails against a real violation: temporarily appended a comment with `rubric`, `readiness`, `evidence` to `getting-started.doc.mjs`, ran the suite, watched it fail with the expected message naming the word and file, then reverted. Full `docs-discovery.test.mjs` suite: 23/23 passing. Typecheck and lint clean.

Not a visual change (a test file, no rendered output), so no screenshot.

### DIFF
--- a/packages/cli/foundation/discovery/docs-discovery.test.mjs
+++ b/packages/cli/foundation/discovery/docs-discovery.test.mjs
@@ -13,6 +13,7 @@
 import {afterEach, beforeEach, describe, expect, it} from 'vitest';
 import * as fs from 'node:fs';
 import * as path from 'node:path';
+import {CLI_ROOT} from '../fs/paths.mjs';
 import {
   BUILTIN_DOCS_PACKAGE,
   DocsCatalog,
@@ -57,6 +58,41 @@ beforeEach(() => {
 
 afterEach(() => {
   fs.rmSync(tmpDir, {recursive: true, force: true});
+});
+
+describe('assets/docs audience', () => {
+  // packages/cli/assets/docs ships to people BUILDING WITH Astryx, not people
+  // building Astryx itself — see the README's own audience rule and routing
+  // table. These five words are a clean, measured signal for material that
+  // belongs on the wiki instead: 0 hits across the directory as authored
+  // today, 56 in the draft (#5351) that prompted this check. Deliberately
+  // narrower than the README's own longer "tells" list — audit/checklist/gate
+  // also match plenty of innocent prose ("Verification Checklist", "Audit
+  // every reset stylesheet"), which would make the check noisy enough to get
+  // suppressed rather than acted on.
+  const OUR_PROCESS_WORDS =
+    /\b(rubric|readiness|promotion|sign-off|evidence)\b/i;
+  const DOCS_DIR = path.join(CLI_ROOT, 'assets', 'docs');
+
+  it('has no our-process words in a caller-facing doc topic', () => {
+    const hits = [];
+    for (const file of fs.readdirSync(DOCS_DIR)) {
+      if (!file.endsWith('.mjs')) {
+        continue; // README.md documents the rule; it isn't shipped as a topic.
+      }
+      const content = fs.readFileSync(path.join(DOCS_DIR, file), 'utf8');
+      const match = content.match(OUR_PROCESS_WORDS);
+      if (match) {
+        hits.push(`${file}: "${match[0]}"`);
+      }
+    }
+    expect(
+      hits,
+      'This word describes building Astryx, not building with it. Move ' +
+        'the material to the matching wiki page in assets/docs/README.md\'s ' +
+        'routing table instead of shipping it here.',
+    ).toEqual([]);
+  });
 });
 
 describe('discoverBuiltinTopics', () => {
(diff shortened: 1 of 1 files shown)


cixzhang (changes requested):
Builders adding valid caller-facing guidance can be blocked and sent to the wrong docs surface. `OUR_PROCESS_WORDS` treats every occurrence as internal process, but the directory README qualifies these terms by context (“as things the reader must produce”) and explicitly says caller-facing system behavior can sound like process. I reproduced this by adding valid motion guidance about “compositor promotion”: the test failed on `motion.doc.mjs: "promotion"` and told the author to move it to the wiki.

Please narrow this to process-oriented phrases or structure instead of banning standalone vocabulary, and add a negative-control fixture showing that a legitimate caller-facing use is allowed. The existing positive case can continue proving that actual rubric/readiness content is rejected.

[Reviewed by Robohands]

HelloOjasMutreja:
Fixed in 69477c958.

The README's own rule (line 24) is narrower than vocabulary: rubric/readiness/promotion/sign-off/evidence are a tell only "as things the reader must produce", and a statement about the system's own behavior is caller-facing even where it sounds like process. "GPU compositor promotion" describes rendering behavior a caller can rely on, with no reader-facing directive anywhere near it, which is exactly why it tripped the old word-anywhere-in-the-file check.

`findOurProcessWordHits` now checks sentence by sentence, and only counts a hit when the same sentence also carries a directive or audience marker (must, before merging/promoting, attach, reviewer(s), checklist, gate, sign off, and a few others). That's the actual shape of the language the check exists to catch, not the vocabulary alone.

Added a negative-control fixture reproducing the exact case from the review ("Animating a transform triggers GPU compositor promotion, which keeps the animation off the main thread.") and confirmed it no longer flags. Also added it live: temporarily inserted that same sentence into the real `motion.doc.mjs`, confirmed the suite stays green, then reverted the insertion (not part of the commit). Kept a positive fixture ("Reviewers must attach evidence to the readiness checklist before sign-off.") proving genuine process language, the kind #5351's draft actually used, still gets caught.

Verified: `docs-discovery.test.mjs` passes (25/25, including the two new cases), the broader `packages/cli/foundation` suite shows only pre-existing unrelated failures (Windows path drive-letter assertions, signal-handling timing) confirmed present without this change too, lint and typecheck clean.

HelloOjasMutreja:
Narrowed `OUR_PROCESS_WORDS` to require directive/audience context in the same sentence (must, before merging/promoting, checklist, sign-off, etc.), and added the negative-control fixture for the exact "compositor promotion" false positive called out above. Rebased onto current main.

PR #5371, opened 2026-08-23
ci: publish PR previews as real GitHub deployments

cixzhang:
Storybook and Sandbox previews are published to GitHub Pages, but GitHub only sees the bot comment that links them. Reviewers using the PR timeline or the Deployments API cannot discover either preview as a deployment.

## What changes

- The shared gh-pages publisher creates transient `Storybook` and `Sandbox` deployments on the trusted PR head, but only for targets its publication result confirms are live.
- A failed attempted publish records `failure`; an expired-artifact manual re-deploy stops before the publisher, so it records nothing.
- Cleanup retires every historical deployment whose stable `Preview for PR #<n>` description matches a stale preview before deleting the directory. If retirement fails, the directory remains so the next cleanup retries safely.
- The three workflows add only `deployments: write`; publishing, recording, and cleanup remain centralized in the existing publisher.

## Load-bearing API details

- `required_contexts: []` prevents deployment creation from returning `409 conflict` while unrelated PR checks are still running.
- `auto_inactive: false` prevents one PR's success from retiring every other live PR that shares the `Storybook` or `Sandbox` environment name.
- `transient_environment: true` identifies these as previews that cleanup will deactivate.

## Test plan

- `pnpm vitest run .github/scripts/lib/gh-pages-publisher.test.mjs .github/scripts/visual-gate/workflow-concurrency.test.mjs --project node --maxWorkers=1` — 61 passed
- `pnpm check:repo` — passed
- `actionlint -ignore SC2129` on the three changed workflows — passed
- `pnpm exec prettier --check` on all six changed files — passed
- `git diff --check` and the public-repo residue guard — passed


cixzhang:
Direction: keep the deployment integration. Storybook and Sandbox previews should be first-class GitHub Deployments with cleanup, not links discoverable only by parsing bot comments.

.github/workflows/cleanup-previews.yml
```diff
@@ -260,3 +270,38 @@ jobs:
...
+            head_sha="$(gh pr view "$pr_num" --repo "$GITHUB_REPOSITORY" --json headRefOid --jq '.headRefOid' 2>/dev/null || true)"
+            if [ -z "$head_sha" ]; then
+              echo "  pr/${pr_num}: head commit unknown (PR deleted?) — nothing to retire"
+              continue
+            fi
+            for environment in Storybook Sandbox; do
+              # A PR with nothing left to retire is the normal case on a rerun,
+              # so an empty list must not trip `set -u`.
+              ids=()
+              mapfile -t ids < <(gh api "/repos/${GITHUB_REPOSITORY}/deployments?sha=${head_sha}&environment=${environment}" --jq '.[].id' 2>/dev/null)
```
cixzhang: Only the head commit. Every earlier push's record stays active pointing at the directory just deleted.

.github/workflows/redeploy-preview.yml
```diff
@@ -105,3 +107,58 @@ jobs:
...
+
+      # Replace the deployment record deploy-preview.yml wrote for this commit.
+      # A manual re-deploy is usually a rescue after the automatic one failed,
+      # so leaving its `failure` record standing would keep the PR claiming a
+      # broken preview that is now live. Inline rather than shared with
+      # deploy-preview.yml on purpose: that workflow deliberately checks out
+      # nothing, so it cannot call a local composite action (the gh-pages push
+      # loop above is duplicated for the same reason).
+      - name: Record the preview as a GitHub deployment
+        if: always()
```
cixzhang: Hmm, `always()` fires on skipped and cancelled too. `deploy-preview` gates on `verify.ready`; this doesn't.

cixzhang (commented):
Thanks — the `required_contexts` and `auto_inactive` notes each save someone a day, and the probe table is the right evidence.

Two things before it lands.

Records are written per commit and retired per PR head. Over the last 20 closed PRs that is 150 records against 40 retirements: 110 `success` records keep a live link to a `pr/<n>/` directory the same job just deleted.

And the manual re-deploy's `always()` has no equivalent of `deploy-preview`'s `verify.ready` guard — the two inline copies have already drifted — so with [#5519](https://github.com/facebook/astryx/pull/5519) on main its "artifacts expired, re-run CI" stop posts `failure` for both environments while the preview is still live.

For the retirement, does matching the `description` you already write feel right, or would you rather walk the PR's commits?

[Reviewed by Robohands]

cixzhang (commented):
Thanks — both blockers are fixed. Cleanup retires every matching historical record, and an expired manual re-deploy records nothing. The shared publisher keeps the paths together. Ready.

[Reviewed by Robohands]

cixzhang (commented):
Thanks — the prior cleanup and expired-artifact blockers remain fixed. The rebase keeps one publisher and records only confirmed Storybook/Sandbox targets. Ready.

[Reviewed by Robohands]

cixzhang:

PR #5373, opened 2026-08-24
Improve Popover responsive fit and focus behavior

rubyycheung:
## Problem

Popover had two concrete usability failures on compact surfaces:

1. A consumer-supplied width, or a wide trigger used by match-trigger sizing, could make the Popover exceed the available viewport and safe-area gutters.
2. A read-only Popover had no ordinary content control to focus, so initial focus landed on the visually hidden fallback close control and revealed an unexpected “Close popover” button.

The goal is to fix those failures without silently turning Popover into a BottomSheet, changing the public trigger API, or clipping content that already fits.

## Changes and rationale

| Change | Problem it solves | Why this approach |
| --- | --- | --- |
| Cap the layer’s inline size with `100vi` and a `100vw` fallback, minus token and safe-area gutters. | Explicit widths such as `640px` could overflow a phone viewport. | The requested width is still honored when space exists; only unavailable space constrains it. Logical viewport sizing works in both writing directions, while `100vw` supports browsers without viewport-axis units. |
| Cap block size with `100dvb` and a `100vh` fallback, minus top and bottom safe-area gutters. | Tall content could extend below the visible viewport, especially when the mobile visual viewport changes. | Dynamic viewport units follow the currently visible area where supported; the fallback preserves compatibility. |
| Cap match-trigger sizing with `min(anchor-size(width), available space)`. | A trigger wider than the available viewport could force its Popover to overflow. | Popover still matches the trigger when possible, but viewport fit wins when the two requirements conflict. |
| Apply the same viewport caps directly to the content surface. | The real tall-content story showed that the surface sits inside the `usePopover` dialog wrapper, so `max-block-size: inherit` inherited `none` and still extended below the viewport. | Reusing the computed inline/block caps on the actual surface keeps every wrapper in the sizing chain constrained; measured overflow then becomes internally scrollable. |
| Enable `overflow: auto` and overscroll containment only after measured content exceeds the available size. | Always making every Popover a scroll container could clip shadows or intentionally protruding content, even when nothing overflowed. | Resize, content-mutation, image-load, window-resize, and visual-viewport changes remeasure the surface; repeated signals are coalesced to one measurement per animation frame so live-updating content does not force repeated layout reads. |
| Prefer the first real focusable content control during autofocus. | The generated fallback close control could become the first focus target instead of the action the user opened the Popover to use. | The canonical focusable selector is reused, but elements inside the fallback-close wrapper are excluded from initial-focus selection. |
| Focus the labeled dialog container when a dialog-style Popover has no content controls. | Read-only content still needs a predictable initial focus target, but focusing the hidden close control made it visibly appear. | `tabIndex={-1}` makes the already labeled dialog container programmatically focusable without adding it to normal tab order. |
| Let Tab move from a programmatically focused trap container into its first control, and Shift+Tab into its last control. | Once the read-only dialog container owned focus, the existing focus trap could let the first Tab escape instead of entering the trap’s controls. | This extends the existing trap cycle only for the case where the trap container itself is focused; ordinary first/last-control wrapping is unchanged. |
| Mark the generated fallback close wrapper and retain the fallback control in the tab sequence. | Excluding the fallback from initial focus must not remove the keyboard escape mechanism. | The marker distinguishes generated fallback UI from product content; users can still Tab to the close control after reading the container content. |
| Preserve current trigger semantics and types. | Responsive work should not create source or ARIA regressions. | `PopoverTriggerRenderProps['aria-haspopup']` remains `'dialog'`; automatic triggers continue using the existing role-aware `usePopover` value. No public prop or dismissal contract changes. |
| Make the Disabled story’s Button actually disabled. | `isEnabled={false}` prevented Popover opening, but the trigger still looked and behaved like a clickable Button. | Setting `Button isDisabled` makes the example accurately demonstrate the whole disabled experience rather than only suppressing the Popover callback. |
| Add explicit-width, match-trigger, and tall-content stories using the actual Storybook viewport. | Unit styles alone did not prove the rendered wrapper chain; inline Docs previews can also make top-layer placement look like component overflow. | The stories use real Popover and List components, render in isolated Storybook iframes rather than the inline Docs canvas, and use no simulated phone or decorative frame. Reviewers or a device simulator can verify width caps, safe gutters, trigger-derived sizing, conditional scrolling, and page containment against the real story viewport. |
| Add trigger-interaction evidence that opens by activation rather than hover. | Popover must remain usable where hover is unavailable. | The story exercises click/tap-style activation and points to unit tests for Escape, outside press, and focus return. |
| Add a dedicated read-only dialog-focus story for manual assistive-technology checks. | Automated tests can verify ARIA and focus movement but cannot prove what a screen reader announces. | The story provides one stable scenario and explicit checks for the announced dialog name/role, forward and reverse Tab containment, Escape, and focus return. |
| Add separate “Keep Popover” and “BottomSheet option” stories using the same compact full-width action list. | Consumers needed a concrete way to compare the anchored and bottom-edge modal interaction contracts. | Holding task content constant makes those contract differences visible without claiming that content traits alone determine the component; compact `ListItem` rows are realistic actions rather than oversized custom Buttons. |
| Add an opt-in adaptive recipe. | Some products want a BottomSheet for compact coarse-pointer/no-hover environments, but automatic substitution would change focus, modality, scrim, scrolling, and dismissal semantics. | Core Popover remains unchanged. The story-local recipe adapts only when a product opts in and provides a deterministic override for review and tests. |
| Update Popover width documentation and the patch changeset. | Consumers otherwise would not know that requested widths are constrained or that long content may scroll. | The behavior is documented in full, translated, and dense component docs and is included in release notes. |
| Add regression tests for viewport caps, conditional overflow, trigger semantics, initial focus, fallback reachability, focus-trap entry, and measurement scheduling. | The fixes span CSS constraints, DOM measurement, ARIA, keyboard behavior, and live content updates. | Tests cover fitting versus overflowing content, one-measurement-per-frame coalescing, real controls winning focus, read-only content focusing the dialog, forward/reverse Tab entry, and the explicit empty-focusable guard. |

## Explicit non-goals

- Popover does **not** automatically become BottomSheet in Core.
- Width, pointer precision, and hover availability remain independent inputs.
- No public Popover prop, trigger type, or dismissal contract is changed.
- The comparison stories do not claim that scrolling, input complexity, search, content length, or hierarchy automatically determines another component.
- This PR does not define a universal component-selection rule; products must evaluate the task’s focus, space, modality, dismissal, and navigation requirements.
- Desktop Storybook and iOS Simulator do not establish VoiceOver or physical-device touch behavior.

## Responsive and Interaction Readiness

| Check | Result | Evidence |
| --- | --- | --- |
| Wide viewport + fine pointer + hover | Pass | Existing anchored presentation and non-overflowing content behavior are retained. |
| Narrow viewport + fine pointer + hover | Pass | Explicit and match-trigger widths cap to available space; scrolling activates only for measured overflow. |
| Narrow viewport + coarse pointer + no hover | Pass | Activation does not require hover; an opt-in BottomSheet recipe is available when product semantics justify it. |
| Wide viewport + coarse pointer + no hover | Pass | Popover remains anchored; touch availability alone does not substitute the component. |
| Keyboard initial focus | Pass | Real content controls receive initial focus; read-only dialog content focuses the labeled container. |
| Keyboard containment and escape | Pass | Tab enters the focusable cycle from the container, fallback close remains reachable, and Escape/focus return tests pass. |
| Adaptive presentation | Pass | Popover remains the default; BottomSheet is explicit and story-local. |
| Public API compatibility | Pass | Existing trigger props and `aria-haspopup` behavior are preserved. |
| iOS visual layout | Pass | The BottomSheet comparison was checked in Safari on an iPhone 17 Pro iOS 27 Simulator. |
| Screen-reader announcement | Follow-up | The `Read-only dialog focus` story now defines the manual test. NVDA + Chrome and VoiceOver + Safari announcement checks are not yet run, so this PR makes no screen-reader-announcement claim. |

## Validation

- `pnpm exec vitest run packages/core/src/Popover/Popover.test.tsx packages/core/src/hooks/useFocusTrap.test.tsx` — 52 tests pass
- `pnpm check:repo`
- targeted ESLint and Prettier checks
- Headless browser at 390×844: explicit-width and match-trigger Popovers remained within the viewport at 358px wide; the realistic project picker capped at 360px with `scrollHeight: 566px` and `overflow: auto`
- Real-viewport stories: `core-popover--viewport-fit`, `core-popover--match-trigger-viewport-fit`, and `core-popover--tall-content-overflow`
- Manual AT story: `core-popover--read-only-dialog-focus`; NVDA + Chrome and VoiceOver + Safari announcement checks remain follow-up work
- Headless Chromium at 390×844: the read-only story initially focuses the labeled `role="dialog"` container; Tab reaches the fallback Close button and Shift+Tab remains inside the dialog
- BottomSheet comparison checked in Safari on an iPhone 17 Pro iOS 27 Simulator

### DIFF
--- a/packages/core/src/Popover/Popover.tsx
+++ b/packages/core/src/Popover/Popover.tsx
@@ -418,6 +418,9 @@ export function Popover({
   }, [measureOverflow]);
 
   useIsomorphicLayoutEffect(() => {
+    if (!popover.isOpen) {
+      return;
+    }
     const surface = popover.contentRef.current;
     if (!surface) {
       return;
@@ -460,6 +463,9 @@ export function Popover({
   }, [measureOverflow, popover.isOpen, scheduleOverflowMeasurement]);
 
   useIsomorphicLayoutEffect(() => {
+    if (!popover.isOpen) {
+      return;
+    }
     scheduleOverflowMeasurement();
   }, [content, popover.isOpen, scheduleOverflowMeasurement]);
 
--- a/packages/core/src/Popover/Popover.tsx
+++ b/packages/core/src/Popover/Popover.tsx
@@ -38,6 +38,11 @@ import {InteractiveRoleContext} from '../InteractiveRoleContext/InteractiveRoleC
 // =============================================================================
 
 const BUTTON_SELECTOR = 'button, [role="button"]';
+const POPOVER_VIEWPORT_GUTTER = spacingVars['--spacing-4'];
+const POPOVER_MAX_INLINE_SIZE = `calc(100vi - max(${POPOVER_VIEWPORT_GUTTER}, env(safe-area-inset-left)) - max(${POPOVER_VIEWPORT_GUTTER}, env(safe-area-inset-right)))`;
+const POPOVER_MAX_INLINE_SIZE_FALLBACK = `calc(100vw - ${POPOVER_VIEWPORT_GUTTER} - ${POPOVER_VIEWPORT_GUTTER})`;
+const POPOVER_MAX_BLOCK_SIZE = `calc(100dvb - max(${POPOVER_VIEWPORT_GUTTER}, env(safe-area-inset-top)) - max(${POPOVER_VIEWPORT_GUTTER}, env(safe-area-inset-bottom)))`;
+const POPOVER_MAX_BLOCK_SIZE_FALLBACK = `calc(100vh - ${POPOVER_VIEWPORT_GUTTER} - ${POPOVER_VIEWPORT_GUTTER})`;
 
 /**
  * Find the trigger button inside a container element.
@@ -63,8 +68,8 @@ export interface PopoverTriggerRenderProps {
   ref: (el: HTMLElement | null) => void;
   /** Toggle the popover open/closed. */
   onClick: () => void;
-  /** ARIA attribute: indicates the element triggers a dialog. */
-  'aria-haspopup': 'dialog';
+  /** ARIA attribute: indicates the popup type exposed by the trigger. */
+  'aria-haspopup': 'dialog' | 'true';
   /** ARIA attribute: whether the popover is currently open. */
   'aria-expanded': boolean;
   /** ARIA attribute: ID of the controlled popover element. */
@@ -241,6 +246,24 @@ const styles = stylex.create({
   anchorWrapper: {
     display: 'inline-flex',
   },
+  viewportFit: {
+    boxSizing: 'border-box',
+    maxInlineSize: stylex.firstThatWorks(
+      POPOVER_MAX_INLINE_SIZE,
+      POPOVER_MAX_INLINE_SIZE_FALLBACK,
+    ),
+    maxBlockSize: stylex.firstThatWorks(
+      POPOVER_MAX_BLOCK_SIZE,
+      POPOVER_MAX_BLOCK_SIZE_FALLBACK,
+    ),
+  },
+  surfaceViewportFit: {
+    boxSizing: 'border-box',
+    maxInlineSize: 'inherit',
+    maxBlockSize: 'inherit',
+    overflow: 'auto',
+    overscrollBehavior: 'contain',
+  },
   // Content padding, applied to the popup surface so a theme's `padding`
   // replaces it instead of nesting inside it.
   contentPadding: {
@@ -253,7 +276,11 @@ const styles = stylex.create({
     width: typeof width === 'number' ? `${width}px` : width,
   }),
   matchTrigger: {
-    minWidth: 'anchor-size(width)',
+    minWidth: stylex.firstThatWorks(
+      `min(anchor-size(width), ${POPOVER_MAX_INLINE_SIZE})`,
+      `min(anchor-size(width), ${POPOVER_MAX_INLINE_SIZE_FALLBACK})`,
+      'anchor-size(width)',
+    ),
   },
 });
 
@@ -343,7 +370,7 @@ export function Popover({
     // it is the element the `popover` theme target has to sit on — a target on
     // the content div inside it styles a box that paints nothing.
     surfaceTarget: 'popover',
-    xstyle: [styles.contentPadding, xstyle],
+    xstyle: [styles.contentPadding, styles.surfaceViewportFit, xstyle],
     className,
     style,
     onShow: handlePopoverShow,
@@ -503,18 +530,27 @@ export function Popover({
   }, [isOpen, isControlled, popover]);
 
   // Determine popover xstyle
-  const popoverXstyle = width ? styles.customWidth(width) : styles.matchTrigger;
+  const popoverSizeXstyle = width
+    ? styles.customWidth(width)
+    : styles.matchTrigger;
 
   // Sibling mode: render only the popover (no wrapper needed)
   if (anchorRef && children == null) {
     return (
       <>
-        {popover.render(<div data-testid={testId}>{content}</div>, {
-          placement,
-          alignment,
-          offset: spacingVars['--spacing-1'],
-          xstyle: [popoverXstyle, layerAnimations[placement]],
-        })}
+        {popover.render(
+          <div data-testid={testId}>{content}</div>,
+          {
+            placement,
+            alignment,
+            offset: spacingVars['--spacing-1'],
+            xstyle: [
+              styles.viewportFit,
+              popoverSizeXstyle,
+              layerAnimations[placement],
+            ],
+          },
+        )}
       </>
     );
   }
@@ -526,7 +562,7 @@ export function Popover({
     const triggerProps: PopoverTriggerRenderProps = {
       ref: popover.triggerRef,
       onClick: handleTriggerClick,
-      'aria-haspopup': 'dialog',
+      'aria-haspopup': popover.triggerProps['aria-haspopup'],
       'aria-expanded': popover.isOpen,
       'aria-controls': popover.id,
     };
@@ -534,12 +570,19 @@ export function Popover({
     return (
       <>
         {children(triggerProps)}
-        {popover.render(<div data-testid={testId}>{content}</div>, {
-          placement,
-          alignment,
-          offset: spacingVars['--spacing-1'],
-          xstyle: [popoverXstyle, layerAnimations[placement]],
-        })}
+        {popover.render(
+          <div data-testid={testId}>{content}</div>,
+          {
+            placement,
+            alignment,
+            offset: spacingVars['--spacing-1'],
+            xstyle: [
+              styles.viewportFit,
+              popoverSizeXstyle,
+              layerAnimations[placement],
+            ],
+          },
+        )}
       </>
     );
   }
@@ -552,12 +595,19 @@ export function Popover({
           {children}
         </div>
       </InteractiveRoleContext>
-      {popover.render(<div data-testid={testId}>{content}</div>, {
-        placement,
-        alignment,
-        offset: spacingVars['--spacing-1'],
-        xstyle: [popoverXstyle, layerAnimations[placement]],
-      })}
+      {popover.render(
+        <div data-testid={testId}>{content}</div>,
+        {
+          placement,
+          alignment,
+          offset: spacingVars['--spacing-1'],
+          xstyle: [
+            styles.viewportFit,
+            popoverSizeXstyle,
+            layerAnimations[placement],
+          ],
+        },
+      )}
     </>
   );
 }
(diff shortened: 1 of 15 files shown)


gonzoblasco:
Hi @rubyycheung, this is a thorough PR - the decision table in the body makes the responsive trade-offs legible, and the positive *and* negative tests (fitting content is not clipped, overflowing content scrolls) are the right shape.

The focus work is the part I'd want to pressure-test hardest, since it changes how a dialog-style Popover announces itself. A few things:

**1. The `tabIndex={-1}` dialog container is exactly the case that needs a real AT pass - worth tightening the "Blocked" note before merge.**
Your own readiness table flags iOS VoiceOver/top-layer as *Blocked / not verified*. That's the honest call to make, but this specific pattern (focus a `role="dialog"` container with `tabIndex={-1}` + `aria-modal` when there are no content controls) is where screen readers diverge most. In practice: when a user Tabs into a programmatically-focused dialog container, some SR/AT combos announce the `aria-label` ("dialog, [name]"), but others try to announce the full content region or announce nothing meaningful on a focus that landed on a container rather than a control. Since the whole point of this change is a predictable initial-focus target for read-only content, I'd suggest validating at minimum: NVDA + Chrome, and VoiceOver + Safari, on the "Read only" story - confirming (a) the dialog name is announced on open, and (b) the first Tab moves into the cycle rather than being announced as a jump to an unrelated region. If VoiceOver can't be run before merge, that's fine, but I'd flag it as a follow-up rather than a silent "not verified", because a wrong announcement here is a worse outcome for the user than the previous behaviour.

**2. The new `useFocusTrap` branch relies on `focusable` being non-empty - worth an explicit guard.**
The added block (`if (document.activeElement === container)`) runs after `const first = focusable[0]; const last = focusable[focusable.length - 1];`. If `focusable` can be empty at that point (the trap container focused with no tabbable descendants - the read-only case this PR targets, since the fallback close control is retained but the container is the focused element), then `last` is `undefined` and the Shift+Tab branch calls `last.focus()`. There's an existing test for "trap has no tabbable controls" so there may be an early return above - but if so, an explicit `if (!focusable.length) return;` right before the new block would make that dependency obvious to the next reader instead of implicit.

**3. Measurement cost: two observers + window listeners on every open.**
The `PopoverSurface` registers a `ResizeObserver` + `MutationObserver` (subtree) + `load`/`resize`/`visualViewport` listeners, and re-measures on every `children` change. For a Popover that's fine on open, but if a Popover's content re-renders frequently while open (live-updating lists, form validation state), the mutation observer on `subtree: true` fires on every DOM change inside it. Worth a quick check that this doesn't add measurable jank on a fast-updating Popover - and if it's a concern, throttling `measureOverflow` with `requestAnimationFrame` (coalescing multiple observer callbacks into one measure) is a cheap guard.

**4. Consistency with the HoverCard/popup contract - good to see the `aria-haspopup="dialog"` preserved.**
The trigger still signals `aria-haspopup="dialog"` and the `FOCUSABLE_SELECTOR` is reused - coherent with how a dialog-style popup should be exposed. One nit: the read-only dialog container gets `tabIndex={-1}` only when `role === 'dialog'`. If a Popover uses `role="none"` with a menu inside (the story you added), the container won't get the programmatic focus target - which is correct, since the menu has focusable items. Just confirming that's the intended split and not a case where `role="group"`-style content with no controls could still lose its initial focus.

Overall: strong work on the sizing/overflow, and the focus fix addresses a real loss-of-place bug. The main ask is to tighten the AT-verification story for the container-focus case and make the empty-`focusable` dependency explicit.

rubyycheung:
Addressed each point in `ad7f66c902f`:

1. Added a dedicated **Read-only dialog focus** Storybook story with the exact manual AT checks (announced dialog name/role, forward and reverse Tab containment, Escape, and focus return). The PR now marks NVDA + Chrome and VoiceOver + Safari announcement verification as explicit follow-up work instead of a silent “not verified” claim.
2. Confirmed the focus trap has an explicit `focusable.length === 0` guard before computing `first`/`last`; the existing no-tabbable-control test covers that branch. I also added reverse-entry coverage for Shift+Tab from the programmatically focused container.
3. Resize, mutation, image-load, window-resize, visual-viewport, and children-change signals now coalesce to one overflow measurement per animation frame. A regression test fires multiple signals and verifies only one frame is scheduled.
4. Confirmed the `role="none"` split is intentional: content-owned menu/listbox roles keep their own focus model and do not make the neutral wrapper a programmatic focus target. Existing docs and tests cover that contract.

I also rebased the branch onto current `main` and resolved the Popover conflict with the newer layer-depth wrapper.

gonzoblasco:
Hi @rubyycheung - all four look resolved on the code side:

**1. AT verification for the container-focus case** - Making the NVDA + Chrome and VoiceOver + Safari announcement checks an explicit follow-up (instead of a silent "not verified") matches the risk of this pattern, and the dedicated Read-only dialog story gives a stable, reproducible scenario to run that against.

**2. Empty-`focusable` guard** - Good to see the explicit `focusable.length === 0` early return before `first`/`last` are computed; that makes the dependency obvious to the next reader. The added Shift+Tab reverse-entry coverage closes the branch I was asking about.

**3. Measurement cost** - Coalescing the resize/mutation/load/resize/visualViewport/children signals into one `requestAnimationFrame` measurement is the cheap guard I suggested, and the regression test asserting only one frame is scheduled makes the behavior explicit.

**4. `role="none"` split** - That matches my reading: content-owned menu/listbox roles keep their own focus model, so the neutral wrapper correctly stays out of programmatic focus. Consistent, no gap.

Code-level, this looks good to me. The two AT verification items are the ones worth doing before this pattern lands.

rubyycheung:
Fixed in the latest push. The public UsePopoverReturn.show() options now expose only skipAutoFocus; focusTarget remains available solely through the internal toggleWithOptions path used by Popover. Added a compile-time contract assertion so it cannot leak back into the public hook surface. The focused Popover suite passes (40 tests), along with ESLint and core typecheck.

rubyycheung:
/accept-visual 33299931172/1 API-only follow-up: the latest commit narrows a TypeScript return type and adds a type-contract test; rendered Popover behavior and the previously reviewed visual frames are unchanged.

imdreamrunner (approved):

imdreamrunner:
/accept-visual [33299931172](tel:33299931172)/1 API-only follow-up

packages/core/src/Popover/Popover.tsx
```diff
@@ -330,7 +410,7 @@ export function Popover({
     onOpenChange?.(false);
   }, [onOpenChange]);
 
-  const popover = usePopover({
+  const popover = usePopoverInternal({
```
cixzhang, line 413: Hmm this is using an internal implementation now? Although I couldn't find the reference.

packages/core/src/Popover/usePopover.tsx
```diff
@@ -410,24 +469,35 @@ function usePopoverImplementation(
...
-    if (layer.wasJustDismissed()) {
-      return;
-    }
-    if (layer.isOpen) {
-      layer.hide();
-    } else {
-      show();
-    }
-  }, [layer, show]);
+  const toggleWithOptions = useCallback(
```
cixzhang, line 484: We're exposing a new method on the public contract that does the same thing as `toggle` but with options. It might be better to update the `toggle` contract to accept options intead

packages/core/src/Popover/Popover.tsx
imdreamrunner: `usePopoverInternal ` was added in this PR https://github.com/facebook/astryx/pull/5018/changes

What kind of reference are you referring to?

packages/core/src/Popover/usePopover.tsx
imdreamrunner: `toggleWithOptions` was only exposed in `usePopoverInternal` not the public API.

cixzhang: Seems unfortunate to have a second internal only version. We're probably missing a proper spec for the public version to absorb the capabilities we need.

packages/core/src/Popover/usePopover.tsx
```diff
@@ -28,13 +28,37 @@ import {
...
 import {Button} from '../Button';
+import {FOCUSABLE_SELECTOR} from '../hooks/focusableSelector';
 import {rtlStyles} from '../utils';
 import {useTranslator} from '../i18n';
 import {useDevWarning} from '../hooks/useDevWarning';
 import {mergeProps} from '../utils/mergeProps';
 import {themeProps} from '../utils/themeProps';
 import {stableClassName} from '../naming';
 
+const FALLBACK_CLOSE_SELECTOR = '[data-astryx-popover-fallback-close]';
```
cixzhang, line 39: This only shows up here but references an element generated elsewhere so they can end up misaligned.

PR #5375, opened 2026-08-24
feat(Toast): add scroll-safe swipe dismissal

rubyycheung:
## Problem

Touch and pen users need a direct way to dismiss transient Toasts. The first implementation used `touch-action: none` on the whole Toast, which prevented page scrolling whenever a gesture started over the overlay. Adding a new `'swipe'` dismissal reason would also expand a public union without a demonstrated consumer need.

This follow-up adds swipe dismissal while preserving native scrolling, existing controls, and the existing dismissal API.

## Solution decisions

| Decision | Problem it addresses |
| --- | --- |
| Swipe vertically toward the configured top or bottom edge. | Entry, direct manipulation, and exit use one spatial model: top Toasts move upward; bottom Toasts move downward. |
| Wait for dominant edge-directed intent before claiming touch movement. | Vertical movement competes with page scrolling. Horizontal, opposite-direction, and unresolved movement remain native. |
| Use a non-passive `touchmove` listener only for the accepted handoff. | `preventDefault()` occurs only after the intent cone resolves, rather than globally through `touch-action: none`. |
| Support pen but not mouse drag. | Pen is direct-contact input. Mouse drag conflicts with text selection and is unnecessary because the close control remains visible. |
| Ignore starts from interactive descendants. | Buttons, links, inputs, and other controls keep their own activation and manipulation behavior. |
| Pause auto-hide while a gesture is unresolved. | A Toast cannot expire while the user is actively manipulating it; cancelled gestures resume the remaining timer. |
| Use distance and velocity thresholds with transform, opacity, and scale feedback. | Thresholds reduce accidental dismissal; progress remains visible without adding a motion dependency. |
| Keep the existing `'manual'` dismissal reason. | Swipe is user-initiated and does not justify permanently widening the public `'auto' | 'manual'` union. |
| Keep placement owned by `ToastViewport`. | `Toast` inherits a private numeric swipe direction from the wrapper; the rejected non-positioning `Toast.position` prop does not return. |
| Retain close, F6, Escape/layer, and imperative dismissal paths. | Swipe is an enhancement, never the only dismissal or reachability path. |

## Dependency

Stacked on #5353 because the gesture uses its responsive width, visible motion region, edge-motion, and stack-collapse foundation. This branch is rebased onto #5353 head `1c0b74a8329`; review this PR as the swipe-only delta.

## Impact

Every rendered Toast gains touch and pen gesture handling. The behavior is additive: mouse dragging remains inactive, interactive descendants retain their controls, and unresolved/opposite-direction touch movement remains available to the page. Queueing, timers outside an active gesture, announcements, focus handling, layout, placement defaults, and dismissal callbacks retain #5353 behavior.

## API and ossification

- No public prop, export, default, or union member is added.
- `ToastViewport.position` remains the only placement API and still defaults to `bottomEnd`.
- `ToastProps` still has no `position` property.
- `ToastDismissReason` remains `'auto' | 'manual'`; swipe reports `manual`.
- Swipe direction and progress use private `--_toast-*` lifecycle variables documented as private, not theme or consumer API.

## Breaking risk

No API break is intended. Touch behavior changes only after movement resolves toward the configured dismiss edge. Consumers relying on selecting text with a pen may observe direct manipulation; mouse text selection is deliberately unchanged. The visible close control remains the non-gesture alternative.

## Performance and resources

- No React state or render pass is added; gesture state lives in refs and private inline CSS variables.
- Each visible, non-exiting Toast owns four native touch listeners (`touchstart`, `touchmove`, `touchend`, `touchcancel`); they are removed on exit or unmount. The default maximum remains five visible Toasts.
- Pointer handling uses existing React event delegation.
- No observer, global listener, motion dependency, or extra DOM node is added.

## Responsive and Interaction Readiness

| Check | Result | Evidence |
| --- | --- | --- |
| Touch toward configured edge | Pass | Non-passive touch handoff test verifies cancellation only after accepted directional intent. |
| Opposite-direction touch | Pass | Gesture resets without movement or dismissal and does not cancel native scrolling. |
| Horizontal touch | Pass | Native scrolling remains unclaimed. |
| Pen | Pass | Pointer-event tests cover thresholds, reset paths, RTL, and both placements. |
| Mouse | Pass | Mouse drag does not begin swipe; the visible close control remains available. |
| Interactive descendants | Pass | Controls do not start swipe; their native action remains available. |
| Timer behavior | Pass | Auto-hide pauses during manipulation and resumes after cancellation. |
| Public API compatibility | Pass | Placement remains on `ToastViewport`; dismissal remains `'auto' | 'manual'`. |
| iOS physical touch | Not claimed | Unit tests and Storybook validate the handoff contract, but this PR does not claim physical-device touch evidence. |

## Validation

- 123 focused tests: 45 `ToastViewport`, 4 `useToast`, and 74 derived-variable tests.
- Core, component-doc, and Storybook typechecks.
- Targeted ESLint and `pnpm check:repo`.
- Full repository build.
- Storybook production build and Toast accessibility baseline audit.

## Judgement

Ready for review once refreshed CI is green. The branch conflict is resolved against #5353 without restoring `Toast.position` or regressing #5353’s layout, inset, or announcement behavior.

### DIFF
--- a/.changeset/toast-swipe-dismissal.md
+++ b/.changeset/toast-swipe-dismissal.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] Toast now supports touch and pen swipe dismissal toward its configured viewport edge. The vertical axis matches each Toast's top/bottom placement and entrance/exit motion, so the dismissal follows the same spatial model instead of introducing a separate side exit. Native touch scrolling is preserved until movement resolves to the dismiss direction, interactive controls do not start a swipe, and swipe continues to report the existing manual dismissal reason. Pen is included as direct-contact input; mouse drag is excluded because desktop users already have the visible close control and dragging can conflict with text selection.
+
+@rubyycheung
--- a/packages/core/src/Toast/Toast.tsx
+++ b/packages/core/src/Toast/Toast.tsx
@@ -4,9 +4,9 @@
 
 /**
  * @file Toast.tsx
- * @input Uses React timers, Toast options, Button/Icon, MediaTheme, tokens, and
- *   placement-derived motion variables inherited from ToastViewport
- * @output Exports the rendered Toast surface and its pause/dismiss behavior
+ * @input Uses React timers, touch/pen gesture events, Toast options, Button/Icon,
+ *   MediaTheme, tokens, and placement motion inherited from ToastViewport
+ * @output Exports the rendered Toast surface and its pause/swipe/dismiss behavior
  * @position Core implementation; rendered by ToastViewport and documented by Toast.doc.mjs
  *
  * SYNC: When Toast layout, timer pause, media theme, or dismissal behavior changes,
@@ -33,6 +33,7 @@ import {
   typeScaleDefaults,
 } from '../theme/tokens.stylex';
 import {mergeProps} from '../utils';
+import {INTERACTIVE_SELECTORS} from '../hooks/useClickableContainer';
 import {useTheme} from '../theme';
 import {MediaTheme} from '../theme/MediaTheme';
 import type {
@@ -42,6 +43,26 @@ import type {
 } from './types';
 import {themeProps} from '../utils/themeProps';
 import {useTranslator} from '../i18n';
+import {useToastGesture, type ToastGestureDirection} from './useToastGesture';
+
+const SWIPE_INTERACTIVE_TARGET_SELECTOR = `${INTERACTIVE_SELECTORS},[tabindex],[contenteditable]:not([contenteditable="false"])`;
+
+function isInteractiveTarget(
+  target: EventTarget | null,
+  root: HTMLElement,
+): boolean {
+  if (!(target instanceof Element)) {
+    return false;
+  }
+  let current: Element | null = target;
+  while (current != null && current !== root && current !== document.body) {
+    if (current.matches(SWIPE_INTERACTIVE_TARGET_SELECTOR)) {
+      return true;
+    }
+    current = current.parentElement;
+  }
+  return false;
+}
 
 const TOAST_EDGE_DRIFT = spacingVars['--spacing-2'];
 const styles = stylex.create({
@@ -53,11 +74,12 @@ const styles = stylex.create({
     width: 400,
     maxWidth: '100%',
     boxShadow: shadowVars['--shadow-med'],
-    opacity: 1,
+    opacity: 'var(--_toast-swipe-opacity, 1)',
     fontFamily: typographyVars['--font-family-body'],
     fontSize: typeScaleDefaults['--text-body-size'],
     lineHeight: typeScaleDefaults['--text-body-leading'],
-    transform: 'translateY(0)',
+    transform:
+      'translateY(var(--_toast-swipe-y, 0px)) scale(var(--_toast-swipe-scale, 1))',
     transitionProperty: 'opacity, transform',
     transitionDuration: {
       default: durationVars['--duration-fast'],
@@ -90,7 +112,7 @@ const styles = stylex.create({
   },
   exiting: {
     opacity: 0,
-    transform: `translateY(var(--_toast-slide-y, ${TOAST_EDGE_DRIFT}))`,
+    transform: `translateY(var(--_toast-swipe-exit-y, var(--_toast-swipe-y, var(--_toast-slide-y, ${TOAST_EDGE_DRIFT})))) scale(var(--_toast-swipe-scale, 1))`,
   },
   endContent: {
     flexShrink: 0,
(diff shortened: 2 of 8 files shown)


packages/core/src/Toast/Toast.tsx
```diff
@@ -265,6 +312,259 @@ export function Toast({
...
+        resumeTimer();
+      }
+    },
+    [resumeTimer],
+  );
+
+  const handlePointerDown = useCallback(
+    (event: ReactPointerEvent<HTMLDivElement>) => {
+      if (
+        event.pointerType === 'pen' &&
```
cixzhang: Hmm, pen swipes and mouse doesn't. Same drag, same hand — worth a line saying why.

.changeset/toast-swipe-dismissal.md
```diff
@@ -0,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Toast now supports touch and pen swipe dismissal toward its configured viewport edge. Native touch scrolling is preserved until movement resolves to the dismiss direction, interactive controls do not start a swipe, and swipe continues to report the existing manual dismissal reason.
```
cixzhang: `[fix]` on a feat PR — `[feat]` like cli-theme-targets.md, or is this a fix?

cixzhang (commented):
Thanks — the right shape. Under real touch only an edge-directed swipe claims the gesture; the rest still scrolls (measured: 45° up scrolls the page 154px and doesn't engage, 55° up dismisses and cancels every touchmove).

One question, not blocking: why vertical? Up/down is the one axis that collides with page scroll, which is what the intent cone and the non-passive listener exist to solve. Both platforms swipe notifications sideways, and every position already names a horizontal edge. If there's a reason — leaves the way it came in — worth a line in the body.

Two optional nits inline.

[Reviewed by Robohands]

packages/core/src/Toast/Toast.tsx
rubyycheung: Added the rationale in d0da171ddb7. Pen stays because it is direct-contact surface manipulation like touch. Mouse drag stays excluded because it is not a standard Toast affordance, can conflict with desktop text selection, and mouse users already have the visible close control.

.changeset/toast-swipe-dismissal.md
rubyycheung: Good catch — this adds a new interaction, so I changed the changeset classification from `[fix]` to `[feat]` in d0da171ddb7.

rubyycheung:
Addressed the latest review in d0da171ddb7:

- Kept the vertical axis intentionally because Toast placement and motion already use the top/bottom edge: top Toasts enter/exit upward and bottom Toasts enter/exit downward. The gesture now documents that shared spatial model rather than presenting it as a platform-notification convention.
- Documented why pen is supported but mouse drag is not: pen is direct-contact input; mouse drag can conflict with text selection and desktop already has the visible close control.
- Changed the changeset category from `[fix]` to `[feat]`.
- Added the rationale to the PR body, component guidance, Storybook example, and release note.

Validation: changeset check, Core typecheck, Storybook typecheck, repo checks, and the prior 40-test/full-build run pass.

cixzhang (changes requested):
Thanks, this is good, and I checked it against the custom `renderContent` branch: the card still owns the gesture, so custom layouts keep swipe dismissal.

One architecture thing I'd like before merge: pull the swipe behavior into a named `useToastGesture`-style unit, still invoked by the Toast card root. The hook would take the resolved edge direction, pause/resume hooks, dismiss callback, enabled/exiting state, and interactive-descendant policy; it would return the root binding plus transient swipe updates and cleanup. That keeps custom content safely inside the card while making direction arbitration, thresholds, timer pause, and cancel/dismiss decisions independently testable instead of buried in the render file.

Could you make that split while keeping the hook attached to the Toast card, not the custom content?

[Full review](https://github.com/cixzhang/astryx/wiki/Review-5375)

rubyycheung:
Addressed the requested architecture split in b8c39d2895f:

- extracted gesture arbitration, thresholds, touch/pen listeners, timer pause/resume, transient style updates, and cleanup into `useToastGesture`
- kept the hook attached to the Toast card root, including custom/interactive descendant protection
- resolved the swipe direction in `ToastViewport` and passed it to the card without expanding the public `ToastProps` API
- merged the updated stacked base and preserved its newer responsive layout/motion behavior

Validation: 122 focused tests pass; Core typecheck and production build pass; focused ESLint has no errors.

cixzhang (approved):
Thanks — my earlier review asked for this extraction, and `useToastGesture` now provides the right boundary. xinzi-bot’s reset/rebase notes are valid; please fix both before merge.

[Full review](https://github.com/cixzhang/astryx/wiki/Review-5375)

[Reviewed by Robohands]

rubyycheung:
/accept-visual 33219800855/1 Adds the intended Mobile situations / Swipe to dismiss story across all seven stable themes in light and dark modes; 224 existing frames are unchanged, with 14 expected additions and 0 changed, removed, or failed frames.

Issue #5378, opened 2026-08-24
Markdown incremental parser drops the whole settled cache on open fences and on blank-line chunk boundaries


cixzhang:
Streaming a Markdown document through `parseMarkdownIncremental` is supposed to
re-parse only the unsettled tail. Two paths throw the settled cache away and
re-parse the whole document instead. Both are invisible to the perf suite's
wall-clock budgets, and to the median-based cache assertion, because they hit a
minority of chunks — but each one that hits costs O(N).

Measured with the perf suite's own `generateAIResponse` fixture, 50-char chunks,
counting blocks rebuilt per chunk by object identity:

| document | chunks | p50 | p90 | max | total rebuilt |
| --- | --- | --- | --- | --- | --- |
| 50 paragraphs (6.6k) | 133 | 2 | 45 | 59 | 1,513 |
| 200 paragraphs (26k) | 528 | 2 | 165 | 234 | 20,975 |
| 500 paragraphs (66k) | 1,320 | 2 | 409 | 584 | 126,756 |

p50 is flat at 2 — the cache works for the typical chunk. p90 grows linearly
with the document, and total rebuilt work grows roughly quadratically.

## 1. An open code fence discards the entire settled prefix

`findSettledBoundary` ends with `return inFence ? -1 : lastBoundary`, and the
caller treats `-1` as "nothing is settled" and re-parses the full input:

```ts
if (boundary < 0) {
  // Inside an unclosed fence or no blank-line boundary — full re-parse
  state.prevInput = input;
  return parseMarkdownImpl(input, parseOpts);
}
```

While a fence is open, every appended token re-parses the whole document. The
content *before* the fence opened is unaffected by anything typed inside it, so
the last boundary seen before the fence opened is still a valid settle point.

This is the common case, not an edge case: AI chat responses are full of code
blocks. Isolating it — same fixture, a fence every third paragraph vs. none:

| | chunks | p50 | p90 | max | total rebuilt |
| --- | --- | --- | --- | --- | --- |
| 500 paragraphs, no fences | 1,436 | 1 | 3 | 952 | 11,611 |
| 500 paragraphs, fences | 2,091 | 2 | 871 | 1,167 | 480,823 |

~37% of chunks rebuild the whole document once fences are present.

A fix is not quite a one-liner: routing the open-fence tail through the normal
unsettled path would also route it through `trimUnsettledStructural`, which
strips trailing lines that look like incomplete list markers or table rows. A
[...]

PR #5379, opened 2026-08-24
test(markdown): assert the streaming cache holds at every document length

cixzhang:
## Problem

A chat view streaming a Markdown-heavy AI response re-parses on every token. The
incremental parser exists so that each token re-parses the *unsettled tail*, not
the document; when that cache is lost the cost per token becomes O(N) and the
user watches the message stutter as it grows.

**The wall-clock budgets could not have told us.** `parser.perf.test.ts` asserts
them with 10x headroom. Destroying the cache locally — dropping
`settledText`/`settledBlocks` on every call — left all five budget tests passing,
the XL streaming case at 2741ms against its 30000ms budget.

The `blocksBuiltPerChunk` median added in
[#5366](https://github.com/facebook/astryx/pull/5366) closes part of that, but
only part: **a cache that holds for the typical chunk and collapses on a minority
of them is p50-flat and still O(N) overall.** The median cannot see that shape,
and that shape is what
[#5378](https://github.com/facebook/astryx/issues/5378) documents on `main`.

## Change

Test-only, no production code touched. Adds one assertion to the
`parseMarkdownIncremental cache` block: **p95 blocks rebuilt per chunk, and that
it does not move across a 10x document** — 3 at 50, 200 and 500 paragraphs.

The signal is object identity. A block the parser kept in its settled cache is
handed back as the very same object, so a returned block that is not one of the
previous chunk's is one this chunk built. That reads the cache without a counter,
a mock, or a test-only export. Fixture and chunking are deterministic, so every
asserted number is an integer with no timing input — it reads the same on a
loaded CI box as on an idle laptop.

`generateAIResponse` grows one optional `{codeBlocks: false}` flag; the default is
unchanged and every existing test keeps its old fixture. The one-off `median`
helper is gone and the [#5366](https://github.com/facebook/astryx/pull/5366)
assertion now calls `percentile(v, 0.5)` — one convention for one statistic;
verified value-identical at all three asserted lengths.

## Deliberate break — the assertion has been watched fail

Two breaks to `parseMarkdownIncremental`, each run and then reverted.

**A — the settled cache dropped on every chunk**, the shape of the design that
measured 71.8ms against 34.9ms on a 14k document:

| test | broken | restored |
| --- | --- | --- |
| 5x `full re-parse … under Nms` | **pass** | pass |
| `streaming full re-parse Medium/XL` | **pass** | pass |
| `streaming incremental Medium/XL` | **pass** (2741ms / 30000ms) | pass |
| `incremental parse is faster than full re-parse` | fail — 0.79x against a 1.1x tolerance | pass |
| `median blocks built per chunk` | fail — `[31, 118, 293]` vs `[2, 2, 2]` | pass |
| **`p95 blocks built per chunk`** | **fail — `[58, 223, 556]` vs `[3, 3, 3]`** | pass |

Every wall-clock *budget* passed with the cache completely destroyed. The one
wall-clock test that did catch it is a ratio, and it caught it by 21%.

**B — the cache dropped on one chunk in eight**, a minority collapse, which is
the case this PR exists for:

| test | broken |
| --- | --- |
| `incremental parse is faster than full re-parse` | **pass** — 1.60x |
| `median blocks built per chunk` | **pass** — `2/2/2` |
| **`p95 blocks built per chunk`** | **fail — `[42, 161, 387]` vs `[3, 3, 3]`** |

Nothing that was already there sees it. In both breaks the new assertion's
failure values grow with document length — the O(N) signature the budgets cannot
see.

## Why p95, and why no code fences

Two paths still drop the whole settled cache on `main` — an open code fence, and
a chunk boundary landing one line past a blank line — both filed as
[#5378](https://github.com/facebook/astryx/issues/5378), both parser bugs, both
out of scope for a test-only PR.

Removing fences from the fixture takes the first out. The second is still live at
~2% of chunks, so **this assertion does not catch the O(N) that is on `main`
today** — measured fence-free, total blocks rebuilt is 214 / 2156 / 9359 at
50 / 200 / 500 paragraphs while p95 stays flat at 3. It is the strongest bound
that is green, not a clean bill of health.

What it does bound: any collapse hitting more than ~5% of chunks, at any document
length. p90 (2/2/2) would only reach 10%; the cliff is just under p98, which is
already red on `main` (`3/44/3`). p95 is chosen for margin, not for being the
highest green value — p96, p97 and p97.5 are all 3/3/3 too.

The five wall-clock tests are left alone. Deleting them is a separate argument.

## Test plan

`pnpm exec vitest run packages/core/src/Markdown/parser.perf.test.ts` — 14
passed.

```
  median blocks built per chunk (50/200/500 paragraphs): 2/2/2
  p95 blocks built per chunk (50/200/500 paragraphs): 3/3/3
```

Rebased onto `main` — the earlier red was
`apps/storybook/stories/Chat.stories.tsx` failing typecheck on `size="small"`,
which was on the branch's merge-base and fixed on `main` by
[#5377](https://github.com/facebook/astryx/pull/5377).

No changeset — test-only, matching recent `test(...)` commits.

### DIFF
--- a/packages/core/src/Markdown/parser.perf.test.ts
+++ b/packages/core/src/Markdown/parser.perf.test.ts
@@ -8,7 +8,10 @@ import {
 } from './parser';
 import type {BlockNode, ParseOptions} from './parser';
 
-function generateAIResponse(paragraphs: number): string {
+function generateAIResponse(
+  paragraphs: number,
+  {codeBlocks = true}: {codeBlocks?: boolean} = {},
+): string {
   const sections: string[] = [];
   for (let i = 0; i < paragraphs; i++) {
     const mod = i % 6;
@@ -20,7 +23,9 @@ function generateAIResponse(paragraphs: number): string {
         break;
       case 1:
         sections.push(
-          '```typescript\nfunction process(data: Record<string, unknown>[]) {\n  return data.filter(item => item.valid)\n    .map(item => transform(item))\n    .reduce((acc, val) => merge(acc, val), {});\n}\n```',
+          codeBlocks
+            ? '```typescript\nfunction process(data: Record<string, unknown>[]) {\n  return data.filter(item => item.valid)\n    .map(item => transform(item))\n    .reduce((acc, val) => merge(acc, val), {});\n}\n```'
+            : `Paragraph ${i} continues the explanation with a second block of prose, keeping the block mix realistic without opening a fence.`,
         );
         break;
       case 2:
@@ -100,8 +105,10 @@ function blocksBuiltPerChunk(
   return built;
 }
 
-const median = (values: number[]): number =>
-  [...values].sort((a, b) => a - b)[Math.floor((values.length - 1) / 2)];
+const percentile = (values: number[], p: number): number => {
+  const sorted = [...values].sort((a, b) => a - b);
+  return sorted[Math.min(sorted.length - 1, Math.floor(p * sorted.length))];
+};
 
 const sum = (values: number[]): number => values.reduce((a, b) => a + b, 0);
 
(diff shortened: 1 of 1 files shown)


cixzhang:
Auto-merge is armed but blocked, and not by anything here — `build` fails on a typecheck error that is already on `main` (`Chat.stories.tsx`: `size="small"` against Avatar's current scale). Every open PR is behind it. One-token fix in [#5381](https://github.com/facebook/astryx/pull/5381); this will land on its own once that does and CI re-runs.

`test` — the job that actually covers this change — passed.

PR #5385, opened 2026-08-24
feat(core): minQueryLength — hold the typeahead search until the query is long enough

### LINKED ISSUE #5384: [RFC] Typeahead/Tokenizer: no way to hold the search until the query is long enough
### Problem Statement

A typeahead backed by a remote source has no way to say "don't search until the user has typed enough to be worth searching". `BaseTypeahead` searches on every non-empty query, and `performSearch` opens the layer whenever `searchQuery.length > 0` — even when the source returned nothing — so at one character the user gets a request they didn't need and a menu that says **No results found** for a query that matched too much, not too little.

The user-visible half is the worse one: "no results" is a factual claim about the data. After one character of a name search it is usually false, and it is the state that teaches people the field is broken.

### Evidence of Demand

- **Astryx already has the concept, one layer up.** `PowerSearchField.typeaheadMinQueryLength` (`packages/core/src/PowerSearch/types.ts`) exists precisely so a field stays out of the typeahead until the query is long enough, and `usePowerSearchSource` enforces it. It is scoped to PowerSearch's own field list; a plain `Typeahead`/`Tokenizer` over a remote source has no equivalent.
- **External systems ship it as a first-class prop, with the same default.** jQuery UI Autocomplete: `minLength`, default `1` — "the minimum number of characters a user must type before a search is performed … a higher value should be used when a single character search could match a few thousand items". PrimeReact AutoComplete: `minLength`, default `1`, gating on `query.length >= props.minLength`.
- **Internal migration signal.** In the Meta-internal design system currently being migrated onto Astryx, the equivalent prop is set at **36 call sites** on its typeahead/tokenizer — people-search fields over large directories, where 1–2 characters match thousands of rows.

### Why Existing Components Don't Cover This

A `SearchSource` can already refuse to answer a short query — return `[]` and no request goes out. That covers the network half and nothing else:

```tsx
const source: SearchSource<User> = {
  search: q => (q.length < 3 ? [] : fetchUsers(q)),
  bootstrap: () => [],
};
```

The menu still opens on the first character and renders the empty state, because the open decision lives inside `BaseTypeahead` (`performSearch` → `showLayer()` when `searchQuery.length > 0`) and is not reachable from the outside: there is no controlled `isOpen`, no imperative handle, and `onOpenChange` only reports. From a wrapper the only lever left is hiding Astryx's own popover with CSS, which is worse than the honest empty state it replaces.

The result-count live-region announcement has the same shape — screen-reader users hear "no results found" for the search that never ran.

### Rough Approaches Considered

**A. A threshold prop on the engine, forwarded by `Typeahead` and `Tokenizer`.** Below it: no search, no announcement, menu closed. Default `1` = today's behaviour exactly.

```tsx
<Typeahead label="Assignee" searchSource={userSource} value={v} onChange={setV} minQueryLength={3} />
```

Naming: `minLength` is what jQuery UI and PrimeReact call it, but on a component that renders an `<input>` it reads as the native `minlength` validation attribute — a different thing about a different string. `minQueryLength` says which string is being measured and matches the `typeaheadMinQueryLength` already in the repo.

**B. Expose the open state instead** (controlled `isOpen`, or an imperative handle). More general, and a much larger surface — it hands out the layer's whole state machine to solve one threshold.

**C. Let the source signal "not yet"** — e.g. a sentinel return distinct from "no matches". Keeps the API count flat, but puts a protocol in the data layer for something that is a UI decision, and every source implementer then has to know the protocol.

A is the one I've prototyped; the other two are here to show the space.

### Accessibility Considerations

No change to the combobox pattern (ARIA 1.2 combobox + listbox popup). Below the threshold the popup simply never opens, so `aria-expanded` stays `false` and no `aria-activedescendant` is published — the same state as an untouched field, which is what the user is in.

The gain is in what is *not* said: the live-region result-count announcement currently fires for every keystroke, so a 1-character query announces "no results found" before the user has finished the first syllable. Holding the search holds that announcement too.

Open question for the team, and the reason this is an RFC rather than just a PR: **should a below-threshold field say anything?** Options are (a) silence — the field looks untouched, (b) a menu with a "type N or more characters" hint, which is what the internal system does today and is a new user-facing string plus an i18n key, or (c) leave it to the consumer's `description`/help text. I lean (a) for the component and (c) for products, but that is a design call, not mine.

### Pre-submission Checklist

- [x] I have read the [Contributing guide](https://github.com/facebook/astryx/wiki/Contributing)
- [x] I have read the [API Conventions](https://github.com/facebook/astryx/wiki/API-Conventions)
- [x] I have checked that existing Astryx components cannot compose to solve this
- [x] This is a general-purpose UI pattern (not specific to one product)

freddymeta:
Closes #5384.

## The problem

`BaseTypeahead` searches every non-empty query, and `performSearch` opens the layer whenever `searchQuery.length > 0` — including when the source came back empty. On a remote source that means a request per keystroke, and, worse, a menu reading **No results found** after one character, when the truth is that one character matched too much to be worth asking.

A `SearchSource` can already refuse a short query (`q.length < 3 ? [] : fetch(q)`), which covers the network. It cannot close the menu: the open decision lives inside the component and there is no controlled `isOpen`, no imperative handle, and `onOpenChange` only reports. The one lever left to a wrapper is CSS-hiding Astryx's own popover, which is worse than the empty state it hides.

## The change

`minQueryLength?: number` on `BaseTypeahead`, forwarded by `Typeahead` and `Tokenizer`.

```tsx
<Typeahead
  label="Assignee"
  searchSource={userSource}
  value={assignee}
  onChange={setAssignee}
  minQueryLength={3}
/>
```

| query length | `minQueryLength={3}` | default (`1`) |
| --- | --- | --- |
| 0 | untouched state; `hasEntriesOnFocus` decides, as today | unchanged |
| 1–2 | no search, no live-region announcement, menu closed, in-flight search cancelled | unchanged — searches |
| 3+ | searches and opens, as today | unchanged |

**The default is `1`, which is exactly today's behaviour** — every non-empty query searches, and every existing call site renders identically. The prop is the opt-in.

Two follow-on details, both deliberate:

- **ArrowDown does not fall back to bootstrap entries while a below-threshold query sits in the input.** The closed-menu ArrowDown branch bootstraps when `hasEntriesOnFocus` and there are no results; below the threshold that would open a menu of default suggestions that ignore the two characters already typed. Gated on the same predicate, so with the default it never fires differently.
- **The live-region announcement is gated with the search**, not separately — the "no results found" a screen reader hears today after one keystroke is the same defect from the other side.

## A stranded loading state, fixed on the way

Predates this prop and reproduces on main. Abandoning an in-flight search — by emptying the field, by falling below the threshold, or by selecting an item while the next search is still out — bumps the search generation, which makes that search's own `finally` decline to clear the loading flag. The field kept reporting **Loading** to assistive technology until some later search happened to settle. Both abandonment paths now clear it themselves.

## `hasCreate` is not gated by the threshold

Raised in review, and right: `minQueryLength` is a *search* threshold, and creating costs no fetch. A field that can create `QA` should not stop being able to just because a search for `QA` would match too much.

The Create entry used to be built inside `filteredSource.search`, which put it behind the gate. It is now derived from the typed text on a separate path, so with `minQueryLength={3}` typing `QA` offers `Create "QA"` and Enter commits it **while the search source is never called**.

That path is `__queryEntries` on `BaseTypeaheadProps` — underscored and `@internal` deliberately. `BaseTypeaheadProps` is re-exported from the package entry point, so an ordinary name would ship as public API at the next cut with Tokenizer as its only caller. Same treatment as `DefinedTheme.__inputTokens`.

**One consequence worth naming:** the Create entry is appended *after* results are cut to `maxMenuItems`, so with `hasCreate` a full menu shows one option more than the cap — 11 where main showed 10. That is the intended reading of the cap: it bounds how many *results* a menu shows, and creating is a separate capability that should not be crowded out by them. Pinned by a test, since it is exactly the kind of thing a later refactor would "tidy" back into the slice, and documented on both `maxMenuItems` and `minQueryLength`.

## Naming

`minLength` is what jQuery UI (default `1`) and PrimeReact (default `1`) call it, and I started there. On a component that renders an `<input>` it reads as the native `minlength` validation attribute — a different length of a different string — which is exactly the collision [API Conventions](https://github.com/facebook/astryx/wiki/API-Conventions) says not to create. `minQueryLength` names the string being measured, and matches `PowerSearchField.typeaheadMinQueryLength`, which is this same concept already in the repo one layer up.

**No "type N more characters" hint.** Below the threshold the field looks untouched. A hint would be a new user-facing string, a new i18n key across 29 locales, and a design decision about a state the design system has never had — that is the team's call, and it is the open question in #5384. Consumers who want one today can put it in `description`. Whichever way that lands, it composes with this prop rather than replacing it.

**Not wired into `Selector` / `MultiSelector` / `CommandPalette`.** They render `BaseTypeahead` over local, already-loaded item lists where the threshold buys nothing.

## Verification

- `Typeahead` + `Tokenizer` suites: **127 tests pass**, of which 9 are new.
- `pnpm lint:strict` — 0 errors in the touched files. `tsc --noEmit` on `@astryxdesign/core` — clean. `pnpm build` — clean.
- Changeset: `[feat]` → patch (0.x).

The new tests carry their own controls rather than only asserting absence:

- "does not search below the threshold" then types the third character and asserts the same harness *does* see `search('App')` and `aria-expanded="true"` — a permanently-closed menu could not pass it.
- "still offers Create below `minQueryLength`" is paired with a negative control without `hasCreate`: same query, same threshold, menu stays shut and no **No results found** is announced for a query nobody searched.
- A token already held still gets no Create entry below the threshold.
- The first character still searches when the prop is unset.

No visual change — the styling is untouched; the only difference is whether an existing popover opens, and what it holds.

## Docs

`BaseTypeahead`, `Typeahead` and `Tokenizer` `.doc.mjs` (en + zh + dense), a `Minimum Query Length` story, and an `argTypes` control.


packages/core/src/Tokenizer/Tokenizer.doc.mjs
```diff
@@ -378,6 +384,12 @@ export const docsZh = {
       description: '\u4e0b\u62c9\u5217\u8868\u663e\u793a\u7684\u6700\u5927\u9879\u76ee\u6570\u3002',
       default: '10',
     },
+    {
+      name: 'minQueryLength',
```
cixzhang, line 404: Still English in `docsZh`. BaseTypeahead's is translated.

cixzhang (changes requested):
Thanks for this, and for the RFC first — the shape is right.

One thing to fix. The below-threshold branch bumps the search generation but
never clears `isLoading`, so the in-flight search fails its own check in
`finally`. Backspace from three characters to two on a remote source and the
field spins forever — "Loading" in the a11y tree until the third character goes
back in. Without the prop the same gesture clears in ~1.2s.

```ts
setResults([]);
setHasSearched(false);
setIsLoading(false);
```

| interrupted mid-search | search allowed to finish |
|---|---|
| stuck | clean |

"Ap" in both, three seconds after the same backspace.

The empty-field path above strands the same way on main; the line fixes both.

`hasCreate` builds the Create entry inside the search, so the threshold gates
creating too — `QA` + Enter adds a tag today and nothing with
`minQueryLength={3}`. Should Create sit outside the gate?

Also needs a merge with main ([#5400](https://github.com/facebook/astryx/pull/5400)
moved the keydown switch).

Can you add that line and re-push?

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks for fixing the stranded loading state — I re-ran the 600ms remote search, backspaced 3→2, and loading now clears immediately and stays clear after settlement.

One API interaction remains: with `hasCreate`, `QA` + Enter creates a token by default, but `minQueryLength={3}` makes Enter do nothing. `minQueryLength` is a search threshold, so it should not suppress the separate create capability.

Current main also fixes this head’s visual-gate failure; the merge added unrelated DateInput formatting. Could you keep creation available below the threshold, add that Tokenizer combination test, refresh main, and drop the DateInput noise?


[Reviewed by Robohands]

cixzhang (changes requested):
Thanks — the create fix is right, and derived-not-fetched is the right shape. Drove it at 600ms: `QA` offers `Create "QA"`, Enter commits, the source is never called, and the 3→2 backspace still clears loading.

| before | after |
|---|---|
| before | after |

`"QA"` in both, with `hasCreate` and `minQueryLength={3}`.

One thing before it lands. `queryEntries` sits on `BaseTypeaheadProps`, which the core barrel re-exports, so it ships as public API at the next cut with Tokenizer as its only caller and no `BaseTypeahead.doc.mjs` entry. Keep the mechanism, keep it off the contract — `DefinedTheme.__inputTokens` (`defineTheme.ts:395`) is how we carry a prop that has to live on an exported type.

Worth a body line too: with `hasCreate` and no threshold, the Create entry now survives the `maxMenuItems` cut — 11 options where main showed 10.

Can you re-push with that?

[Reviewed by Robohands]

packages/core/src/Tokenizer/Tokenizer.doc.mjs
freddymeta: Translated in `57cde40f8` — `docsZh` now reads 查询搜索源前的最小查询长度…, and it has since picked up the `hasCreate` carve-out too, so the Chinese entry says the same thing the English one does rather than a shorter version of it.

freddymeta:
Both asks in, pushed as `5067e1039`.

**`queryEntries` off the contract.** Renamed `__queryEntries` and marked `@internal`, following `DefinedTheme.__inputTokens` — the mechanism is untouched, only its standing. Tokenizer remains the only caller, and nothing else in the repo referenced it.

**The `maxMenuItems` line** is in the body, and it turned out to be worth more than a line. Chasing it, the Tokenizer docs were saying something that had quietly stopped being true:

> Minimum query length before the search source is queried. Below it no search runs and the menu stays closed.

The menu does *not* stay closed below the threshold once `hasCreate` is set — that is the whole point of the carve-out. All three Tokenizer entries (English, Chinese, dense) now name the exception, and `maxMenuItems` says it bounds *results*, with the Create entry sitting on top of them.

I also pinned the 11-vs-10 behaviour with a test rather than leaving it to the body:

```
it('offers Create on top of a full menu, not in place of a result', ...)
  maxMenuItems={3}, 20 matching results, hasCreate
  → 4 options, the last one Create "QA"
```

Left as an assertion because it is precisely the shape a later refactor would "tidy" by moving the append inside the slice — which would silently drop a result to make room for Create, or drop Create when results are plentiful. Neither is what we want, and neither would fail anything today.

(The inline `docsZh` note was fixed back in `57cde40f8`; replied on the thread.)

cixzhang (approved):
Thanks — both asks are in: `__queryEntries` is internal, and the `maxMenuItems` exception is documented and tested. Typeahead/Tokenizer suites and both Chromium paths pass at this head.

[Reviewed by Robohands]

PR #5394, opened 2026-08-24
fix(core): one platform detection, and Unknown is no answer either

Astro-Han:
Follow-up to #5325, picking up both review notes from @cixzhang. No separate issue: the behaviour was confirmed a bug in review there.

## `Unknown` is no answer either

`platform` in the User-Agent Client Hints spec is a closed set: `"Android"`, `"Chrome OS"`, `"Chromium OS"`, `"iOS"`, `"Linux"`, `"macOS"`, `"Windows"`, `"Unknown"`. The last is the spec's value for "cannot say", so it names a platform no more than `''` does. #5325 stopped `''` from reading as "not Apple", but `/mac/i.test('Unknown')` is still `false`, so `Unknown` kept answering negatively without ever reaching `navigator.platform`.

If anything it is the likelier of the two to appear. `''` is not a legal value, so only a careless embedder emits it; `Unknown` is what a careful one emits. Both fall through now, matched case-insensitively after a trim.

## One detection instead of two

`detectMac` and `isApplePlatform` were byte-identical copies held together by a docstring claiming one mirrored the other, which is why #5253 had to be fixed in two places at once. They are now `utils/isApplePlatform.ts`, imported by both, so the `Unknown` change lands in one place.

It is deliberately not named in `utils/index.ts`. The package root does `export * from './utils'`, so adding it to that barrel would publish it as API, the trap #5316 pulled back from. It is imported by path instead, as `interactionModality` already is, and `pnpm sync:exports:check` is clean.

The kept name is `isApplePlatform` rather than `detectMac`: the predicate matches `iPhone`/`iPad`/`iPod` as well as `Mac`, and `Kbd` reads it into a local `isMac` for glyph choice.

## Tests

`utils/isApplePlatform.test.ts`, 8 cases over the implementation's branches: a client-hints platform that names one (both directions), the five that name none (`''`, whitespace, `Unknown`, `UNKNOWN`, a non-string, each asserted both ways against `navigator.platform`), absent client hints, and no `navigator` at all. Removing the sentinel check turns all five "names none" rows red, including #5253's original blank case.

#5325's blank-platform tests at the two call sites are removed rather than kept. With one implementation there is one place to test the logic, and what proves each surface still consults it is the pair of tests predating #5325 that spoof `MacIntel` directly. Keeping them would have been a third copy of a fact already established twice.

## Validation

347 tests pass across `utils/`, `Kbd/`, `useHotkeys.test.ts` and `serverSafeComponents.test.ts`. `typecheck`, `eslint`, `check:repo` and `sync:exports:check` are clean. A `[fix]` changeset is included.

*Prepared with Claude Code, reviewed and verified by me before opening.*

### DIFF
--- a/.changeset/shared-apple-platform-detection.md
+++ b/.changeset/shared-apple-platform-detection.md
@@ -1,0 +1,18 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Platform detection reads the client-hints `Unknown` sentinel as no answer, and lives in one place
+
+Follow-up to #5325, which taught `useHotkeys` and `Kbd` to fall through to
+`navigator.platform` when `userAgentData.platform` is blank. `Unknown` is the
+User-Agent Client Hints spec's own value for "cannot say", and it names a
+platform no more than `''` does, yet it still committed to the client-hints
+branch and answered "not Apple". It now falls through the same way.
+
+The two detections were independent copies kept aligned by a docstring. They
+are now one internal util that both import, so the next change to this logic
+cannot land on one surface and miss the other. The util is deliberately not
+named in `utils/index.ts`, which would publish it as API.
+
+@Astro-Han
--- a/packages/core/src/Kbd/Kbd.test.tsx
+++ b/packages/core/src/Kbd/Kbd.test.tsx
@@ -23,7 +23,6 @@ describe('Kbd', () => {
       value: originalPlatform,
       configurable: true,
     });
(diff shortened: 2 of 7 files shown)


cixzhang (approved):
Thanks — good catch on `Unknown`, and one detection beats two. Verified in Chrome: an Unknown-platform Mac now draws ⌘ and cmd+K fires. Merging as is.

[Reviewed by Robohands]

PR #5398, opened 2026-08-24
fix(HoverCard,Tooltip): wait for the trigger to have a layout box before showing

### LINKED ISSUE #5374: [Bug] A controlled (isOpen) HoverCard renders in the page corner, not anchored to its trigger
## Description

A `HoverCard` (and `Tooltip`) opened by the controlled `isOpen` prop renders in the page's top-left corner instead of anchored to its trigger. The same card, opened by hover, anchors correctly.

Measured in Chromium against a `HoverCard` inside a `Dialog`, trigger at `(206, 292)`:

| how it opened | card position |
|---|---|
| `isOpen` (controlled) | **`(0, 4)`** |
| hover (uncontrolled) | `(188, 204)` |

The controlled path (`useHoverCard.tsx:511` on `main`) calls `layer.show()` from an effect keyed on `isOpen`. Anchoring appears not to have been established at that point, so the layer falls back to the viewport corner — the same failure mode a layer with no resolvable `position-anchor` shows.

## Reproduction

1. Render `<HoverCard isOpen content="…"><Button label="Trigger" /></HoverCard>`.
2. Look at where the card is.

**Expected:** anchored to the trigger, exactly as hovering it produces.
**Actual:** top-left corner of the viewport.

## Notes

Pre-existing on `main`; the controlled-open effect is byte-identical there. Found while reviewing #4881, which adds a `Pinned hover card in modal` story under `Core/Layer Dismissal` that shows it — that PR does not touch positioning and is not the cause. A fix can land before or after it: they touch different regions of the file and will not conflict.

Worth checking whether `Tooltip`'s controlled path has the same shape (it does, `useTooltip.tsx`) and whether `isDefaultOpen` is affected too.

## Astryx Version

`@astryxdesign/core@0.4.7`

## Environment

HelloOjasMutreja:
Fixes #5374.

## Root cause

`Dialog` mounts its children before it opens — `isOpen` gates the `<dialog>` element's own `showModal()` call, not whether `{children}` renders. So a `HoverCard`/`Tooltip` inside one with a controlled `isOpen={true}` set from the start (or `isDefaultOpen`) runs its show effect while the trigger sits inside a still-non-modal `<dialog>`, which has no layout box (0×0).

A trigger with no box is not a valid CSS anchor. `position-anchor` resolves to its fallback position — the viewport corner — and unlike a live layout change, that resolution does not get revisited once the trigger later gets a real box when the dialog actually opens. The popover is stuck there for the rest of that open; I measured it staying at `(0, 4)` for 3+ seconds after the dialog visibly opened, so this isn't a one-frame timing race, it's a permanently invalid anchor relationship.

Traced to the effect ordering: `HoverCard` is a child of `Dialog` in the JSX tree, so React fires its effects (which call `layer.show()` → `popover.showPopover()`) before `Dialog`'s own effect that calls `dialog.showModal()`. I confirmed this empirically, not just by reasoning about it — a `requestAnimationFrame` defer and a `setTimeout(50)` defer both failed to fix it (same permanently-stuck corner position), which ruled out "just needs a bit more time" and pointed at the trigger's layout box itself being the missing precondition, not the timing relative to `showModal()`.

## Fix

Both hooks' show-on-mount and controlled-open effects now go through a small shared helper, `Layer/showWhenAnchored.ts`: shows immediately if the trigger already has a box (the common case, no behavior change), or waits for a `ResizeObserver` to report one otherwise.

Per your note in the issue, checked `Tooltip`'s controlled path — same shape, same fix applied. `isDefaultOpen` on both hooks gets the same guard, since it fails for the identical reason.

## Verification

**Reproduced against the exact story you cited** (`Core/Layer Dismissal` → `Pinned Hover Card In Modal`), not a hand-rolled approximation — my first two attempts at a standalone repro (a HoverCard with `isOpen` already true on initial mount, with and without a Dialog wrapper) didn't reproduce it, which is what led me to the real distinguishing factor: the Dialog has to open *later*, via a click, not be open from the start.

**Before** — card stuck in the corner, disconnected from its trigger:
before

**After** — anchored above "Card trigger", matching `placement="above"`:
after

Measured coordinates matched too: trigger at `(156, 298)`, card at `(12, 16)` (i.e. `(0,4)` once you account for the dialog's own offset) before the fix, `(351.75, 222)` after — correctly positioned relative to the trigger.

Added `showWhenAnchored.test.ts` (6 tests) covering the immediate-show, null-anchor, deferred-show, disconnect-after-fire, zero-sized-entry, and cleanup-before-fire cases. `jsdom` doesn't implement `ResizeObserver` and always reports `0×0` from `getBoundingClientRect()` (no real layout engine), so I added a global test polyfill in `internal/test-utils/src/setup.ts`, matching the file's existing pattern for the Popover API and `matchMedia`. Without it, every existing `isDefaultOpen`/controlled-`isOpen` test in `Tooltip.test.tsx` failed with `ReferenceError: ResizeObserver is not defined` — confirmed this by running the suite before adding the polyfill, then confirmed it passes clean after.

Full suite: `HoverCard`, `Tooltip`, `Layer` — 203/203 passing. Typecheck and lint clean. Also ran the broader `packages/core` suite; a batch of unrelated failures showed up on the first pass (`DateInputTouch`, `extensibleAxes`, `docPropLiterals`, others) — traced them to a stray Storybook dev server left running from earlier investigation eating CPU and causing real test timeouts, not anything caused by this diff. Killed it and reran each flagged file in isolation: all pass, and none of them reference `HoverCard`, `Tooltip`, `Layer`, or `ResizeObserver` at all.


cixzhang (changes requested):
Thanks—HoverCard now anchors correctly, but Tooltip still leaves the user’s hint detached at the viewport corner, and this head also makes the UI suite red.

- `packages/core/src/Layer/showWhenAnchored.ts:27` — In Chromium, the controlled Tooltip-in-Dialog story still opens at `(0, 4)` while its trigger is around `(206, 298)`. The same exact-head run centers HoverCard correctly. Please keep trigger readiness in Layer’s existing lifecycle owner rather than adding a caller-level timing abstraction, and cover both controlled paths in a real-browser regression.
- `internal/test-utils/src/setup.ts:115` — This global ResizeObserver stub changes unrelated TabList behavior: the exact-head test calls `scrollBy` twice; current main passes 74/74, and the head also passes when only this setup change is removed. Please keep the test double local to the affected tests.

[Reviewed by Robohands]

HelloOjasMutreja:
Fixed in 6e4d9f585 and b695d9a4d.

Moved trigger-readiness into `useLayer`'s own `show()`, the one place every context-mode open funnels through, and deleted `showWhenAnchored.ts` along with both callers now folded into it.

Tracing this live in Chromium turned up more than the box-readiness gap alone. HoverCard's own controlled-in-Dialog story was passing by coincidence, not because its readiness handling differed from Tooltip's: with the trigger's box confirmed present, `show()` would open correctly, and moments later a second render (caused by the story's own inline `onOpenChange` closure changing identity, unrelated to isOpen itself) would silently re-mount the SAME popover and trigger elements via React's ref-callback churn. `addAnchorName`/`removeAnchorName`'s remove-then-readd on that churn flickered the trigger's `anchor-name` off and back on, and Chromium's `position-try-fallbacks` resolution does not get revisited once the anchor briefly disappears, permanently stranding the layer at its unanchored fallback position, exactly the reported "(0, 4)" symptom. Since `addAnchorName` is already idempotent, `contextRef`'s null-branch no longer removes the name eagerly, only the branch handing off to a genuinely different element does.

Past that, real-Chromium tracing surfaced a second issue: even with a real box, showing a layer while an ancestor's own CSS entry animation (Dialog's scale/translate transition) is still running anchors against a mid-transform position that also never gets revisited, since a transform changes nothing `ResizeObserver` reports. `show()` now also waits, via `document.getAnimations()`, for any animation running on the trigger or an ancestor to finish before opening, checked one frame after the readiness gate so a just-triggered animation has actually started (confirmed directly against Chromium: checking synchronously always sees zero animations).

On the `internal/test-utils/src/setup.ts` point: `sharedResizeObserver.ts` already has the established pattern for the same gap (`typeof ResizeObserver === 'undefined'` guard, skip straight to the fallback), so `useLayer`'s `show()` now mirrors it instead of depending on a global polyfill. The global stub is removed. A few test files that locally stub `ResizeObserver` for their own unrelated purposes (Popover, PowerSearch x2, Tokenizer) had a no-op `observe()` that never fired at all, which is what actually broke once `useLayer` started depending on that firing; those now fire once synchronously, matching the removed polyfill's own behavior, rather than the fix depending on a shared global.

Added `packages/core/src/Layer/__tests__/anchorReadiness.chromium.spec.ts`, covering both the Tooltip and HoverCard controlled-in-Dialog paths against a real Chromium build (jsdom can exercise neither the animation timing nor actual CSS anchor-positioning resolution this covers). Verified against the reverted code that it fails first (454px gap, the reported viewport-corner position) and passes cleanly across repeated runs with the fix.

Also verified: full `ui` Vitest project (10,058+ passing), explicit isolated runs of every touched suite (Layer, Tooltip, HoverCard, Popover, PowerSearch, PowerSearchValueEditor, Tokenizer, TabList) all green, typecheck and lint clean. The handful of failures that show up in a full-suite run (docPropLiterals, extensibleAxes, Resizable/utils import timeout, Table.perf, ContextMenu, a TabList roving-tabindex focus-timing test) are all confirmed pre-existing by reproducing them with this fix fully reverted back to the exact-head state.

HelloOjasMutreja:
Rebased onto current main. One real conflict in playwright.config.ts: main's own `pr-a11y` work (#6115) added a `*.a11y.browser.spec.ts` glob to the same testMatch array this PR adds `*.chromium.spec.ts` to. Both patterns match real, distinct, currently-used spec files (BottomSheet's keyboard contract vs this PR's own Layer anchorReadiness spec), so kept both.

Verified after rebase: typecheck clean, and the affected suites (Popover, PowerSearch, Tokenizer, Layer, Tooltip, HoverCard) pass, isolated and in a narrower batch. A full combined run of all six directories intermittently worker-crashes or shows a single unrelated PowerSearchFilterEditor failure in this environment; reproduces identically on a clean, unmodified checkout of current main with the same broad batch, so it is local Vitest worker-pool flakiness under a large parallel run here, not something this PR introduced.

HelloOjasMutreja:
Checked the pr-a11y failure on this run: it's in Checkbox.a11y.chromium.spec.ts (a menu-item inventory mismatch, "Show archived" showing up as unexpectedly inert), a file this PR doesn't touch and unrelated to HoverCard/Tooltip anchoring. Not something this diff caused.

HelloOjasMutreja:
Addressed all five points from the last review:

- Bounded the animation wait to finite entry motion. `waitForAncestorAnimations` now filters out any animation with `iterations: Infinity` via `getComputedTiming()`, so a continuously-pulsing ancestor (StatusDot, e.g.) can no longer block a layer from opening. Covered by a new `ContinuouslyAnimatingTrigger` story and browser spec case.
- The delayed path no longer trusts `ResizeObserverEntry.contentRect` alone — a text-only inline trigger's content box can stay empty even with a real border box. Falls back to re-measuring the anchor with `getBoundingClientRect()` when `contentRect` reports empty, while still honoring a mocked entry's `contentRect` where one is provided (kept the existing unit test suite green).
- `contextRef(null)` no longer eagerly clears the anchor name; it only clears when a *different* element claims it. (This part was already in place from the previous round.)
- Fixed the changeset, which still described the since-removed `showWhenAnchored` helper.
- Renamed `anchorReadiness.chromium.spec.ts` to the repo's cross-browser `.a11y.browser.spec.ts` convention and confirmed all 6 cases (including the new continuous-animation one) pass on both Chromium and WebKit.

PR #5399, opened 2026-08-24
fix(chat): replace Badge with plain text in ChatComposerDrawer collapsed state

athz:
## Summary

Replace the `Badge` component in ChatComposerDrawer's collapsed state with plain text. The collapsed toggle now renders "{count} {label}" (e.g. "3 Attachments") as a single text span.

## Before / After

| Before | After |
|--------|-------|
| `[Badge: 3]` `Attachments` (two elements: Badge + span) | `3 Attachments` (single text span) |

## Motivation

The Badge added visual weight that isn't needed for a simple count indicator in a collapsed row. Plain text is simpler, lighter, and gives consumers a predictable "{count} {label}" pattern without needing to reason about Badge styling or placement.

Also removes the `Badge` import dependency from this component.

## Changes

- `ChatComposerDrawer.tsx` — remove Badge import, render count + label as a single text span
- `ChatComposerDrawer.test.tsx` — add test for the plain text rendering
- Changeset added

### DIFF
--- a/packages/core/src/Chat/ChatComposerDrawer.tsx
+++ b/packages/core/src/Chat/ChatComposerDrawer.tsx
@@ -28,7 +28,6 @@ import {
   easeVars,
   typeScaleVars,
 } from '../theme/tokens.stylex';
-import {Badge} from '../Badge';
 import {mergeProps} from '../utils';
 import type {BaseProps} from '../BaseProps';
 import {themeProps} from '../utils/themeProps';
@@ -300,8 +299,9 @@ export function ChatComposerDrawer({
               styles.toggleContent,
               !isCollapsed && styles.toggleContentHidden,
             )}>
-            <Badge variant="neutral" label={count} />
-            <span {...stylex.props(styles.collapseLabel)}>{label}</span>
+            <span {...stylex.props(styles.collapseLabel)}>
+              {count} {label}
+            </span>
           </div>
           <div
             {...stylex.props(
--- a/.changeset/chat-composer-drawer-badge-placement-8cq6.md
+++ b/.changeset/chat-composer-drawer-badge-placement-8cq6.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] ChatComposerDrawer: replace the Badge in the collapsed toggle with plain text. The collapsed state now renders "{count} {label}" (e.g. "3 Attachments") as a single text span instead of a Badge component followed by a label.
+@aldentan
--- a/packages/core/src/Chat/ChatComposerDrawer.test.tsx
+++ b/packages/core/src/Chat/ChatComposerDrawer.test.tsx
@@ -58,4 +58,14 @@ describe('ChatComposerDrawer', () => {
       'true',
     );
   });
+
+  it('renders count and label as plain text in collapsed state', () => {
+    render(
+      <ChatComposerDrawer count={3} label="Attachments" defaultIsCollapsed>
+        <span>Drawer content</span>
+      </ChatComposerDrawer>,
+    );
+
+    expect(screen.getByText('3 Attachments')).toBeInTheDocument();
+  });
 });


cixzhang (dismissed):
Thanks for this — the collapsed row is heavy, you're right.

Merging the count into the label makes it a sentence, and a sentence needs the catalog. `count={1}` renders "1 Items"; the Feedback story renders "1 User feedback requested". In Chinese the default label is 项目, so we render "3 项目" where we already ship "3 个项目". 1234 loses its grouping.

| before | after |
|---|---|
| before: the count in a chip | after: "1 User feedback requested" |

`label` is the consumer's string, though, so no key of the form `{count} {label}` fixes it — the plural lives in the noun. If the goal is just less weight, that's a theme call on the badge, which was also the only handle a theme had on the count. I'd keep the chip.

Either way "1 Items" can't ship, and chip-vs-text is ours to call — we'll come back this week. Nothing to rework meanwhile; if you have a view on weight vs the pill, say so, since that's the call.

[Reviewed by Robohands]

packages/core/src/Chat/ChatComposerDrawer.tsx
```diff
@@ -300,8 +299,9 @@ export function ChatComposerDrawer({
               styles.toggleContent,
               !isCollapsed && styles.toggleContentHidden,
             )}>
-            <Badge variant="neutral" label={count} />
-            <span {...stylex.props(styles.collapseLabel)}>{label}</span>
+            <span {...stylex.props(styles.collapseLabel)}>
+              {count} {label}
```
cixzhang, line 303: Hmm, this makes it a sentence — "1 Items" at count 1. Probably wants `{count, number}` + a plural arm, like `powersearch.valueEditor.itemsCount`.

packages/core/src/Chat/ChatComposerDrawer.tsx
```diff
@@ -300,8 +299,9 @@ export function ChatComposerDrawer({
               styles.toggleContent,
```
cixzhang, line 299: The Badge was the only theme handle on the count. Nothing addresses it now.

cixzhang:
The decision window has closed: keep the collapsed count as a `Badge`. Composing a free-form label with the count cannot pluralize or localize correctly and removes the count’s theme handle. If the issue is visual weight, that belongs on the existing Badge theme surface.

No rework is needed on the plain-text direction. [Decision record](https://github.com/cixzhang/astryx/wiki/Rulings#decided)

[Reviewed by Robohands]

cixzhang (commented):
This needs a human design decision rather than an automated approval or rejection: should the collapsed count retain the established Badge hierarchy, or become quieter plain text?

The implementation introduces no objective accessibility, localization, RTL, or narrow-layout regression in the checked states. Once the representation is decided, please update the JSDoc, localized component docs, and CLI template, which still promise a Badge.

[Reviewed by Robohands]

kentonquatman:
We should make this configurable so that builders can define what's shown in the collapsed state. The current badge treatment will be preferred for some use cases, so we'll need the ability to support both treatments.

kentonquatman (changes requested):
We should make this configurable so that builders can define what's shown in the collapsed state. The current badge treatment will be preferred for some use cases, so we'll need the ability to support both treatments.

cixzhang (approved):
Approved: this exact head implements the landed ChatComposerDrawer contract. `collapsedSummary` replaces the complete Collapsed summary anatomy part, while the Badge + label default and component-owned disclosure semantics remain unchanged.

cixzhang:
/accept-visual 35779528672/1 The default Badge and label remain visually unchanged; the added custom-summary story renders only the caller-provided plain text in the existing summary position without component-owned restyling.

cixzhang:
/accept-visual 35779528672/1 The default Badge and label remain visually unchanged; the added custom-summary story renders only the caller-provided plain text in the existing summary position without component-owned restyling.

PR #5402, opened 2026-08-24
fix(Chat): align message-bubble tokens the way the composer does

cixzhang:
### Why

[#5324](https://github.com/facebook/astryx/pull/5324) aligned tokens in the chat **composer** and left `ChatTokenizedText` — the same token, in the **message bubble** — alone. Both halves were misaligned before, so nothing looked wrong. Now they disagree: you pick a name, it sits on the line you are typing, you send it, and the chip lifts off the line in the bubble.

Measured in Chromium on the real round trip (type `@`, pick a name, send):

| token | composer | bubble before | bubble after |
|---|---|---|---|
| with an icon (`<Icon>`) | 3.69px above baseline | **8.00px** | 3.69px |
| label only | 3.69px | 4.50px | 3.69px |
| custom `render()` | 3.69px | 4.43px | 3.69px |

### What

`ChatTokenizedText` wraps each token in the same `inline-flex` / `vertical-align: middle` box `ChatComposerInput` already uses, for both badge tokens and custom `render()` tokens. Nothing else moves.

The alignment stays at the site that puts a chip into a text flow, rather than going onto `Badge`: `Badge` is mostly a flex child, where `vertical-align` is inert, and it is not the only chip of this shape — `Citation`'s label variant has the same one and is not a `Badge`. A change on `Badge` would reach every consumer's badge in a table cell or a paragraph without fixing the class.

### Risk

Visual, in message bubbles only. A bubble whose token carries an icon gets **shorter** (67.57 → 65.31px in the story below); a bubble with a label-only token grows 0.81px. No API change, no new theme targets, no new dependency, no runtime work added.

### Testing

- 183 Chat tests pass unchanged.
- Round trip driven in real Chromium on `Core/ChatLayout` → *Full AI Chat*: composer and bubble now measure identically (jump 0.00px, was 3.31px with an icon).
- New story `Core/ChatTokenizedText` → *Icon and custom tokens* — the case was not reproducible from any shipped story before, which is why the bubble half was easy to miss.

Before / after, icon token in a bubble:

| before | after |
|---|---|
| before | after |

### DIFF
--- a/.changeset/fix-chat-bubble-token-alignment.md
+++ b/.changeset/fix-chat-bubble-token-alignment.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Chat: a token in a message bubble now sits on the line the way it does in the composer. `ChatTokenizedText` wraps each token in the same `inline-flex` / `vertical-align: middle` box `ChatComposerInput` uses, so a chip stops lifting off the text the moment the message is sent. Follows #5324, which fixed the composer half.
+
+@cixzhang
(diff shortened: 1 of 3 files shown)


cixzhang:
A token stops jumping when you send it: composer and bubble both sit 3.69px above the baseline. The badge is now wrapped, so `.astryx-chat-tokenized-text > .astryx-badge` must become descendant.

[Reviewed by Robohands]

PR #5412, opened 2026-08-24
fix: verify canary packages and default-condition exports

### LINKED ISSUE #5132: verify-exports.mjs cannot catch a broken ESM dist in canary packages: private-package skip + non-probed default condition
Follow-up from the review discussion on #5001 (context: #5000). Two independent gaps in `scripts/verify-exports.mjs` meant the repo's import smoke test could never have caught the charts/richtext/lab `"type": "module"` omission:

1. **Private packages are skipped entirely** (`if (pkg.private) continue;`, ~line 178). But the canaryOnly packages (charts, richtext, lab, vega) are `private: true` only as a guard against stable publishes — the release workflow strips `private` and publishes them to the `@canary` dist-tag, so they do ship to consumers and deserve the same packaging checks. The skip could distinguish truly-private workspace packages from `astryx.canaryOnly` ones.

2. **`default`-condition targets are never probed** (`NON_PROBED_CONDITIONS` includes `default`, ~line 120). For core that is mitigated by its explicit conditions, but charts/richtext/lab expose their runtime entry *only* via `default` — so even if gap 1 were fixed, the entry that was broken would still not have been imported. When `default` is the only runtime condition of an export and matches the importable-extension filter, it is the ESM import surface and probably should be probed.

Either fix alone would have caught #5000; both seem worth closing.

Cypher-Aura-19:
## Problem

`scripts/verify-exports.mjs` is the gate that catches packages whose published entry points don't resolve. Two blind spots let a canary dist ship that no consumer could import:

**1. Canary packages were skipped entirely.** The gate skipped every `private: true` package. `charts`, `lab`, `richtext` and `vega` carry `private: true` only as npm's guard against a stable publish — `release.yml` strips it in its ephemeral checkout and publishes them to the `@canary` dist-tag. They ship, but nothing checked them.

**2. `default`-condition targets were never probed.** `default` sat on the non-probed list beside `types` and `require`. For those canary packages `default` is the *only* runtime condition, so their entry points were never loaded. While validating the issue I found `core` is affected too, and most heavily: 119 of its 121 runtime targets resolve through `default`, so the gate was probing 2 of them.

## Fix

**Package selection** now follows what actually ships, keyed on the repo's existing `astryx.canaryOnly` marker rather than on any package name. `isPublishedPackage()` is deliberately the same predicate the canary publish loop already uses — `(!p.private || p.astryx?.canaryOnly)` — including testing `canaryOnly` for truthiness rather than `=== true`, so the gate cannot skip something the publisher ships.

**`default` is probed only when it is the export's real ESM entry**, not unconditionally. A `default` declared beside an `import` is a fallback that Node never reaches for an `import`, and a `default` nested under `require` is CJS; both stay unprobed. Condition tracking became the full ancestor chain, so a `require` or `types` *anywhere* above a target still rules it out. Non-module targets are excluded by the existing extension filter, so the stylesheets and JSON assets these packages expose through `default` are untouched.

**`pnpm build` now builds `richtext`.** It was promoted out of `lab` and the root build chain was missed, so its `dist` doesn't exist when CI runs this gate after `pnpm build` — the newly-covered package would otherwise fail on absent files rather than on a real defect. (`prepack` covers publishing but never fires in CI.)

Coverage goes from 10 packages / 31 probed targets to 14 / 154.

The script keeps its behavior, output format and exit codes; the printing and `process.exit` logic moved into a `main()` behind the usual `process.argv[1] === fileURLToPath(import.meta.url)` guard so the pure helpers are importable by tests.

## Tests

`scripts/verify-exports.test.mjs` (new, 21 tests).

The regression it was written for: a canary package whose entry exists but imports a file the build never emitted now fails with `ERR_MODULE_NOT_FOUND` — on the unfixed gate that same fixture passed clean. Alongside it, the `default` probing rules (only-condition, beside-`import`, nested-under-`require`, stylesheets/assets), proof that `import`/`require`/`source` handling is unchanged, and a check that `isPublishedPackage()` stays in step with `release.yml` across seven manifest shapes.

The rules are exercised against fixture package trees rather than only the repo, since a gate pointed at the wrong thing passes exactly like one that works. A final block then runs them over the real workspace manifests, so both the rules and the data they're aimed at are covered.

Fixes #5132

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "private": true,
   "license": "MIT",
   "scripts": {
-    "build": "pnpm -F @astryxdesign/build build && pnpm -F @astryxdesign/core build && pnpm -F @astryxdesign/lab build && pnpm -F @astryxdesign/charts build && pnpm -F @astryxdesign/vega build && pnpm -F @astryxdesign/theme-neutral build && pnpm -F @astryxdesign/theme-matcha build && pnpm -F @astryxdesign/theme-stone build && pnpm -F @astryxdesign/theme-gothic build && pnpm -F @astryxdesign/theme-chocolate build && pnpm -F @astryxdesign/theme-y2k build && pnpm -F @astryxdesign/theme-butter build && pnpm bundle:cli-themes && pnpm -F @astryxdesign/cli sync:api-types",
+    "build": "pnpm -F @astryxdesign/build build && pnpm -F @astryxdesign/core build && pnpm -F @astryxdesign/lab build && pnpm -F @astryxdesign/charts build && pnpm -F @astryxdesign/richtext build && pnpm -F @astryxdesign/vega build && pnpm -F @astryxdesign/theme-neutral build && pnpm -F @astryxdesign/theme-matcha build && pnpm -F @astryxdesign/theme-stone build && pnpm -F @astryxdesign/theme-gothic build && pnpm -F @astryxdesign/theme-chocolate build && pnpm -F @astryxdesign/theme-y2k build && pnpm -F @astryxdesign/theme-butter build && pnpm bundle:cli-themes && pnpm -F @astryxdesign/cli sync:api-types",
     "bundle:cli-themes": "node scripts/generate-cli-themes.mjs",
     "dev": "pnpm -F @astryxdesign/storybook dev",
     "test": "vitest run",
--- a/scripts/verify-exports.mjs
+++ b/scripts/verify-exports.mjs
@@ -1,7 +1,6 @@
 #!/usr/bin/env node
 // Copyright (c) Meta Platforms, Inc. and affiliates.
 
-
 /**
  * verify-exports.mjs
  *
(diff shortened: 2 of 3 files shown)


cixzhang (changes requested):
Thanks for tightening this gate. The regression test does not cover the failure this PR says it closes: #5000 was a missing `"type": "module"`, while `scripts/verify-exports.test.mjs:282` substitutes a missing chunk. With the repo’s Node 24 line, module syntax detection lets that `.js` load without `type`, so the probe remains green; a builder can still receive an unusable canary after CI passes. Please add the actual missing-`type` fixture and make the verifier fail it on Node 24. There will also be conflicts with current main’s package checks, so those need to be preserved when rebasing. If you'd rather talk it through with someone, we're in [Discord](https://discord.com/invite/XnsUcFykEP).

[Reviewed by Robohands]

PR #5414, opened 2026-08-24
fix(theme): give syntax punctuation its own AA-passing colour

### LINKED ISSUE #5386: Syntax punctuation fails WCAG AA contrast (2.42:1) — --color-syntax-punctuation borrows the disabled-text grey
`--color-syntax-punctuation` resolves to `--color-text-disabled` (`theme/syntax/tokens.ts:48`), which is a *disabled-state* grey, not a text colour. Against the code surface it measures **2.42:1** — WCAG 2.1 AA wants 4.5:1 for normal text, and punctuation is normal text: it is the brackets, commas, semicolons and operators in every code sample we ship.

Where it renders:
- `CodeBlock` and the docsite code samples — `highlightStyles.ts:56` and `:73`, `CodeBlock.tsx:281`
- Themed values are no better: neutral is `light-dark(#a3a3a3, #525252)` → **2.52:1** light on white, **2.29:1** dark on `#171717`

Why now: axe never used to see it. The line numbers used to be `::before` content, and moving them to real text in [#4314](https://github.com/facebook/astryx/pull/4314) put a real text node in the gutter, so `pr-a11y` started reporting the failure. **The defect is ours and predates that PR** — it is not a condition of merging it, and the review says so.

Suggested fix: give syntax punctuation its own token rather than borrowing the disabled grey, and hold it to 4.5:1 against `--color-syntax-background` in each theme. The pattern from [#4446](https://github.com/facebook/astryx/pull/4446) (a `scripts/check-*-contrast.test.mjs` that resolves through `light-dark()` and `var()` indirection and asserts the ratio) would keep it fixed.

Related: [#4652](https://github.com/facebook/astryx/issues/4652) (umbrella), [#5014](https://github.com/facebook/astryx/issues/5014).

HelloOjasMutreja:
Fixes #5386.

## Problem

`--color-syntax-punctuation` borrowed `--color-text-disabled`, a token WCAG deliberately exempts from the normal-text contrast requirement because it marks an inactive control. Punctuation in a code sample is always-active, normal text.

Measured every theme's punctuation/background pair (resolving `light-dark()`/`var()` indirection, same approach as #4446's badge contrast guard). Three fail AA (4.5:1):

| theme | light | dark |
|---|---|---|
| neutral | 2.42:1 | 2.53:1 |
| chocolate | 3.06:1 | 2.56:1 |
| matcha | 3.83:1 | 2.73:1 |

butter, gothic, stone, and y2k already had their own passing punctuation colour and are untouched.

## Fix

The shared default (used by any theme that doesn't define its own syntax palette) now points at `--color-text-secondary` instead — already used for `--color-syntax-comment`, already verified to clear AA.

The three failing themes each define their own syntax palette rather than inheriting the default, so each needed a dedicated fix:

| theme | new light | new dark | new ratios |
|---|---|---|---|
| neutral | `#6e6e6e` | `#a0a0a0` | 4.89:1 / 7.57:1 |
| chocolate | `#9e622e` | `#cb884d` | 4.84:1 / 6.12:1 |
| matcha | `#566a39` | `#92af6a` | 5.19:1 / 7.02:1 |

Regenerated the CLI theme template bundle (`pnpm bundle:cli-themes`) so `astryx theme add` scaffolds the fixed values too.

## Verification

Added `scripts/check-syntax-punctuation-contrast.test.mjs`: resolves every theme's punctuation/background pair the same way the badge guard does, and holds all seven to AA in both colour schemes. Confirmed it catches a real regression — reverted neutral's fix locally, ran the suite, watched it fail with the exact numbers from the issue (`neutral/light — #a3a3a3 on #fafafa = 2.42:1`), then restored it.

**Before** (neutral theme, dark mode) — punctuation dims into the background:
before

**After** — clearly legible:
after

Typecheck and lint clean. Full `packages/themes` + `packages/core/src/theme` suite passes (749/751 — the 2 failures are `theme/extensibleAxes.test.ts`, a pre-existing, unrelated AST-scan failure I confirmed present on a clean `upstream/main` checkout before touching anything here).

### DIFF
--- a/.changeset/syntax-punctuation-contrast.md
+++ b/.changeset/syntax-punctuation-contrast.md
@@ -1,0 +1,13 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Syntax-highlighted punctuation (brackets, commas, semicolons, operators) now meets WCAG 2.1 AA contrast (4.5:1) against the code surface in every bundled theme.
+
+`--color-syntax-punctuation` borrowed `--color-text-disabled`, a token WCAG deliberately exempts from the normal-text contrast requirement because it marks an inactive control. Punctuation in a code sample is always-active, normal text, and measured well under 4.5:1 in three themes: neutral (2.42:1 light, 2.53:1 dark), chocolate (3.06:1 light, 2.56:1 dark), and matcha (3.83:1 light, 2.73:1 dark). The other four themes (butter, gothic, stone, y2k) already defined their own passing punctuation colour and are untouched.
+
+The shared default now points at `--color-text-secondary` instead (used for `--color-syntax-comment` too, and already verified to clear AA). Neutral, chocolate and matcha each get a dedicated punctuation colour in their own syntax palette, since they define one rather than inheriting the shared default: neutral `#6e6e6e`/`#a0a0a0` (4.89:1/7.57:1), chocolate `#9e622e`/`#cb884d` (4.84:1/6.12:1), matcha `#566a39`/`#92af6a` (5.19:1/7.02:1).
+
+Adds `scripts/check-syntax-punctuation-contrast.test.mjs`, resolving each theme's `--color-syntax-punctuation`/`--color-syntax-background` pair through `light-dark()`/`var()` indirection (the pattern from #4446's badge contrast guard) and holding every theme, both colour schemes, to AA — so a regression here fails the build instead of shipping.
+
+@HelloOjasMutreja
--- a/packages/cli/assets/templates/themes/chocolate/chocolateTheme.ts
+++ b/packages/cli/assets/templates/themes/chocolate/chocolateTheme.ts
@@ -27,7 +27,9 @@ const chocolateSyntax = defineSyntaxTheme({
     tag: ['#8c3a3a', '#d47a7a'],
     attribute: ['#8C5927', '#d4a06a'],
     property: ['#3a7c6b', '#70c4b0'],
-    punctuation: ['#B88859', '#6b5540'],
+    // #B88859/#6b5540 failed WCAG AA against the syntax background: 3.06:1
+    // light, 2.56:1 dark. #5386.
+    punctuation: ['#9e622e', '#cb884d'], // 4.84:1 / 6.12:1
     background: ['#FFFCF7', '#1c1610'],
   },
 });
--- a/packages/cli/assets/templates/themes/matcha/matchaTheme.ts
+++ b/packages/cli/assets/templates/themes/matcha/matchaTheme.ts
@@ -27,7 +27,9 @@ const matchaSyntax = defineSyntaxTheme({
     tag: ['#8c3a3a', '#d47a7a'],
     attribute: ['#7c5e3a', '#c4a882'],
     property: ['#3a7c6b', '#70c4b0'],
-    punctuation: ['#707E46', '#5a6440'],
+    // #707E46/#5a6440 failed WCAG AA against the syntax background: 3.83:1
+    // light, 2.73:1 dark. #5386.
+    punctuation: ['#566a39', '#92af6a'], // 5.19:1 / 7.02:1
     background: ['#F0F0E0', '#1a1c14'],
   },
 });
--- a/packages/cli/assets/templates/themes/neutral/neutralTheme.ts
+++ b/packages/cli/assets/templates/themes/neutral/neutralTheme.ts
@@ -52,7 +52,9 @@ const neutralSyntax = defineSyntaxTheme({
     tag: ['#89001a', '#ffaeaa'], // red
     attribute: ['#584400', '#eec12f'], // yellow
     property: ['#005348', '#83dac9'], // teal
-    punctuation: ['#a3a3a3', '#525252'], // neutral
+    // #a3a3a3/#525252 (this pair's own disabled-text tone) failed WCAG AA
+    // against the syntax background: 2.42:1 light, 2.53:1 dark. #5386.
+    punctuation: ['#6e6e6e', '#a0a0a0'], // neutral, 4.89:1 / 7.57:1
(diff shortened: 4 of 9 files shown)


gonzoblasco:
Hi @HelloOjasMutreja, nice catch - `--color-syntax-punctuation` inheriting a WCAG-exempt disabled token for always-active text is exactly the class of bug that slips through. The measurement (resolving `light-dark()`/`var()` the same way the badge guard does) and the regression test are solid.

I verified the headline numbers independently (WCAG 2.1 relative luminance, 4.5:1): neutral light/dark 4.89:1 / 7.57:1, chocolate 4.84:1 / 6.12:1, matcha 5.19:1 / 7.02:1 all clear AA. Good.

Two things I'd want considered before this lands:

**1. The shared default now makes punctuation visually identical to comments.**
The new default points `--color-syntax-punctuation` at `--color-text-secondary`, which is the same token `--color-syntax-comment` already uses. For the themes that inherit the default (butter, gothic, stone, y2k), punctuation now renders in the exact same colour as comments. That's a semantic distinction the syntax highlighter used to preserve (dim grey for punctuation vs something else for comments), and it's lost in one move. It clears AA, so it's not a correctness issue - but it's a visual-regression concern worth a deliberate decision: either confirm punctuation-as-comment-colour is intended for those themes, or give punctuation its own token so the two roles stay distinguishable. Worth checking against the before/after in those four themes, since they weren't part of the measured three.

**2. The resolver only handles hex - it's worth failing loud when it can't resolve.**
`resolve()` returns `null` for anything that isn't `#rgb`/`#rrggbb`/`#rrggbbaa`. Today every theme ships hex so the pairs all resolve, and the `expect(rows).toHaveLength(14)` guard catches a theme that stops resolving - but the failure message just says "expected 14, got 12" with no hint of *which* theme or *why*. If a future theme defines a colour in `oklch()`/`hsl()`/`rgb()`, the guard trips without telling anyone the resolver doesn't understand the format. A one-line `throw` (or a descriptive assertion) in `resolve()` when it hits an unsupported format would make the next person's debugging much faster. Same spirit as the "guards the resolver itself" test you already wrote - just one step more explicit.

Minor: consider reusing the badge guard's resolver instead of a second copy in `scripts/`, so the `light-dark()`/`var()` indirection logic has a single home. If the badge and syntax guards drift, one will silently cover a format the other doesn't.

Overall a clean, well-scoped fix. The two asks are: make a deliberate call on punctuation-vs-comment colour identity for the default themes, and make the resolver fail loud on unsupported colour formats.

HelloOjasMutreja:
Thanks for the independent WCAG verification, and for catching the resolver's silent-null gap.

**On #2 (fail loud):** pushed in 1c20165bc. `resolve()` now throws with the actual unresolvable value instead of falling through to `null` when it hits something that isn't hex/`light-dark()`/`var()`. Confirmed it's a no-op for current behavior — reran the suite, all 7 themes still resolve cleanly.

**On #1 (punctuation-vs-comment identity):** checked this carefully, and I don't think it's actually a regression from this PR. None of butter/gothic/stone/y2k inherit the shared default — all seven themes (including these four) call `defineSyntaxTheme` with their own full 14-token palette, and `defineTheme.ts` only writes `--color-syntax-*` from a theme's own `syntax.tokens` when `input.syntax` is provided (`defineTheme.ts:632`, `if (input.syntax) { ... }`) — there's no fallback path to `syntaxTokenDefaults` for a theme that has one. `domainTokens/index.ts`'s own header calls `syntaxTokenDefaults` out as "used by defineTheme for validation + autocomplete," not a runtime CSS source.

So for the four you flagged, checking each theme file directly:
- **butter**: `comment: ['#605f52', '#adac9e']`, `punctuation: ['#605f52', '#adac9e']` — already identical, pre-existing, untouched by this PR.
- **stone**: same pattern, `['#5e5e5e', '#ababb0']` for both — pre-existing, untouched.
- **y2k**: same pattern, `['#5e5e5e', '#ababab']` for both — pre-existing, untouched.
- **gothic**: `comment: '#6b7079'`, `punctuation: '#7a8290'` — distinct, untouched.

The shared default I changed (`--color-text-disabled` → `--color-text-secondary`) only matters for a third-party theme that skips `syntax:` entirely, which none of the seven shipped ones do. Let me know if I'm missing a path where it does reach a shipped theme — happy to dig further if so, but I couldn't find one.

**On the "share the resolver" note:** agreed it's the right shape long-term, holding off on it here since it'd mean touching `check-badge-contrast.test.mjs` too, which is unrelated to this fix — happy to do it as a follow-up if you'd rather see it split out.

gonzoblasco:
Great follow-up, and thanks for checking the actual theme files - you're right, and I was wrong on #1. I should have verified the shipped themes against the source before raising it: I assumed the four inherited the shared default, but you've shown all seven define their own full `syntax.tokens` palette via `defineSyntaxTheme`, and `defineTheme.ts` only writes `--color-syntax-*` from a theme's own tokens when `input.syntax` is present - so `syntaxTokenDefaults` never reaches a shipped theme at runtime.

And confirming the specific values makes it conclusive: butter/stone/y2k already had `punctuation === comment` (pre-existing, untouched by this PR), and gothic has them distinct. So there's no visual regression introduced here - my point #1 is withdrawn. The default change only affects a third-party theme that skips `syntax:` entirely, which is a reasonable (and strictly better, since it now clears AA) behaviour for that case.

Glad #2 landed cleanly with the loud-fail commit (1c20165bc) - that closes the real gap I flagged. And agreed on the resolver-sharing being a good follow-up rather than scope creep here; splitting it out later keeps this fix tight.

Nice work on this one.

cixzhang (approved):
Thanks — reproduced every number from real pixels: seven themes, both schemes, and transparent-background `container="section"`. gonzoblasco's two points are settled. Merging once Vercel clears.

[Reviewed by Robohands]

Issue #5417, opened 2026-08-24
[RFC] Theming targets for 15 internal elements a shipped theme still reaches with structural CSS


freddymeta:
## Problem Statement

A theme cannot restyle an element that has no theme target, and `defineTheme`
emits **one compound selector on one element** — no combinators, no at-rules. So
every Astryx element that renders with only StyleX atomics is unreachable from a
theme, and a theme author's only remaining move is to hand-write a structural CSS
file that reaches into Astryx's DOM: `.astryx-textarea > textarea`,
`.astryx-citation > span[aria-hidden='true']`,
`.astryx-table-header-cell button[aria-haspopup='dialog']`.

Those selectors are exactly what the theming system exists to avoid. They are
keyed on element names, child order and ARIA attributes, so any internal
restructure breaks them silently — a theme's CSS keeps parsing and just stops
matching. They are invisible to `themingTargets.test.ts`, which guards the
documented surface but cannot see a consumer reaching around it.

I maintain `@nest/eps-core`, the EPS theme for Astryx (an internal Meta design
system migrating its component surface onto Astryx). It is a real, shipped
`defineTheme` theme: 182 token overrides, 118 component overrides, `--eps-*`
runtime variables so EPS light/dark flows through. I audited every line of custom
CSS it still needs, and **the great majority of what is left is not "EPS wants
something exotic" — it is elements with no target.**

## Evidence of Demand

Audit of the full EPS custom-CSS corpus against `@astryxdesign/core@0.4.7`:
55 files, parsed into **273 selector-rules**, each classified against the **243
documented theme targets**.

| | rules |
|---|---:|
| Our own markup (wrappers we render ourselves) | 116 |
| Targeted element, but needs an ancestor / media / `:has()` reach | 105 |
| **Subject is an Astryx element with no class at all** | **46** |
| Simple + targeted (i.e. should already be in the theme) | 6 |

First, the good news, because it says the guard is working: **every `astryx-*`
class our CSS references is a documented target, and every `themeProps()` call
site in 0.4.7 is documented** (232 call classes ⊆ 243 documented). There is no
drift, and nothing already-exposed sitting unused. The remaining gap is
genuinely elements that were never given a class.

**15 elements account for 45 of those rules, across 20 files — and would delete
7 files outright** (`TextArea.css`, `DateRangeInput.css`, `EmptyState.css`,
`Citation.css`, and three table-plugin stylesheets).

Every render site below verified on `main` @ `cbb991e91`:

[...]

PR #5418, opened 2026-08-24
feat(TextArea): expose theme targets for the control, slots and counter

freddymeta:
Implements the first row of #5417 — the TextArea entry.

## Problem

`TextArea` publishes one theme target, `astryx-textarea`, on the wrapper. The wrapper is a bare positioning context: the `<textarea>` carries the control's own padding, and the start icon, the trailing spinner/status slot and the character counter are absolutely-positioned overlays anchored to that padding. So a theme that wants a different inset has nowhere to put it, and every overlay anchored to the old inset has to be re-placed along with it.

`defineTheme` emits one compound selector on one element, so none of that is expressible. A downstream theme of ours currently has to write:

```css
.astryx-textarea > textarea               { padding-top: …  !important; }
.astryx-textarea[data-size='sm'] > textarea { padding-top: … !important; }
.astryx-textarea > span:has(+ textarea)   { top: …; height: …; display: flex; align-items: center; }
.astryx-textarea[data-size='sm'] > span:has(+ textarea) { top: …; }
.astryx-textarea > textarea ~ div         { background-color: …; padding: …; bottom: 0; }
```

`> span:has(+ textarea)` means "the span before the textarea" — it distinguishes the start icon from the end slot only because they are the same element type in DOM order. That is not a contract anyone agreed to; it is a guess about internals that keeps parsing and silently stops matching the day the DOM changes.

## Change

Four targets, all purely additive — no existing class, data attribute or style changes, and no visual change:

| target | element | reflects |
|---|---|---|
| `astryx-textarea-control` | the `<textarea>` | `size` |
| `astryx-textarea-start-icon` | leading overlay span | `size` |
| `astryx-textarea-end-slot` | trailing spinner/status overlay span | — |
| `astryx-textarea-counter` | character counter | — |

`size` is on the control and the start icon because the values a theme sets there genuinely differ per size (an 8px top inset at `md` and 4px at `sm`, matching the sibling inputs' 28px compact row), and a theme cannot write `.astryx-textarea[data-size='sm'] .astryx-textarea-control`. The end slot and the counter need no size axis, so they get none.

I deliberately did **not** add a state to the counter for the over-limit case. Astryx already colors it through `styles.counterError`, and a theme wanting to restyle that specifically is speculative — easy to add later, impossible to remove.

## Test plan

Re-verified at this head (main merged in):

- `pnpm exec vitest run packages/core/src/theme/themingTargets.test.ts packages/core/src/TextArea packages/core/src/theme/derivedVarRegistry.test.ts` — **453 passed**. The theming guard is the one that matters: it re-derives every `themeProps()` call site from the AST and fails if a rendered class is undocumented, so it is what proves the four new `theming.targets` rows match what the component actually renders.
- `pnpm lint:strict` — 0 errors in the files touched.
- `tsc --noEmit` on `@astryxdesign/core` — clean. `pnpm build` — clean.

No screenshots: the rendered output is byte-identical. `themeProps()` adds a class and (for two of the four) a `data-size` attribute; nothing in this repo's CSS matches them, and the StyleX atomics are untouched.

The `pr-visual` red on the previous head was the infra break of 25 Aug, fixed by #5481/#5482 — merging main clears it, as you expected.

## Note for whoever reviews

While tracing this I found one of the rules in that downstream theme is already dead: `.astryx-textarea > span:has(> button) { pointer-events: auto }` worked around the end slot's `pointer-events: none` swallowing the tooltip status button's hover, and #5147 fixed that properly on the button itself. Nothing needed here — flagging it because it is the good outcome, and I am deleting our copy.


cixzhang:
Thanks — the targets themselves are right, and the counter one retires an `!important` sibling-order selector, which is exactly what these should do.

On the block axis: the intent when a theme sets that padding is that the start icon and end slot move with it, so the stranding is a bug rather than a limitation to document. You already scoped the fix in the changeset — here's the utility to use.

Use the shared container expansion rather than a second hand-rolled entry:

```
textarea: [{property: 'padding', expand: 'container'}],
```

That replaces the current `{property: 'paddingInline', vars: ['--_textarea-inline-padding'], replaces: true}` — same shape as your own [#5181](https://github.com/facebook/astryx/pull/5181), and `card`, `dialog`, `context-menu` and `section` already use it.

The reason it beats adding `paddingBlock` alongside the existing entry: a named entry fires only for that exact property name, so `padding: 20px` and `paddingBlockStart` would still strand. `expand: 'container'` consumes all seven logical spellings plus `paddingTop`/`paddingBottom` (`generateThemeRules.ts:154`, `:176`), and emits `--astryx-textarea-padding-*` tokens the overlays can read instead of the constant `top: var(--spacing-2)` they have now.

Two things to watch when you do it: the wrapper carries `padding: 0` deliberately so the native resize grip sits in the true corner, and `paddingLeft`/`paddingRight` are intentionally excluded from the expansion because they're direction-relative.

Happy for this to be the separate PR you mentioned rather than growing this one — your call which.

[Reviewed by Robohands]

packages/core/src/TextArea/TextArea.doc.mjs
```diff
@@ -187,6 +187,8 @@ export const docs = {
   theming: {
     targets: [
       {className: 'astryx-textarea', visualProps: ['size', 'status'], states: ['disabled', 'readonly']},
+      {className: 'astryx-textarea-control', visualProps: ['size']},
```
cixzhang: `textareaSizeStyles` is `sm: {}`, `md: {}`, `lg` with only `paddingBlock` — the axis is block padding, which the expansion takes over.

cixzhang (changes requested):
One thing to change here before it lands: drop `visualProps: ['size']` from `textarea-control` — inline below. That's the only thing blocking; the padding work stays your separate PR.

Two corrections to what I said above, both measured on this head.

It isn't a block-axis gap — the inline axis strands too. Only the one exact spelling tracks:

| a theme writes | overlays follow? |
|---|---|
| `paddingInline: 20px` | yes |
| `padding: 20px` | no |
| `paddingInlineStart: 20px` | no |
| `paddingBlock: 20px` | no |
| `paddingTop: 20px` | no |

And the resize grip isn't only a thing to watch when you do the fix — a themed `padding` already moves it, from 1px off the corner to 21px:

| today | `padding: 20px` |
|---|---|
| default | grip pushed in |

One thing that'll save you a round: the icon's `top` is `--spacing-2` against the control's `--spacing-1`. Keep that 4px offset — reading the padding straight looks like it tracks and lands 4px out.

[Reviewed by Robohands]

cixzhang (approved):
Thanks — the requested size axis is gone and the target architecture is right. Please refresh the stale body; syncing main will also pick up the visual-job fix.


[Reviewed by Robohands]

PR #5419, opened 2026-08-24
fix(HoverCard): expose rich cards as dialog popups, not descriptions

### LINKED ISSUE #5049: [Bug] Define an accessible popup relationship for rich HoverCard content
## Description

`HoverCard` can contain rich, interactive content and permits focus to move into that content, but the trigger always references the layer with `aria-describedby`.

That relationship is designed for a plain-text description. It flattens structured content instead of exposing it as a navigable popup. The ARIA tooltip pattern is also not a direct fit: tooltips do not receive focus, while a hover surface containing focusable elements is typically modeled as a non-modal dialog.

This is more visible with the lazy context-layer work in #5039: while the card is closed, the trigger's `aria-describedby` points to an empty `<template>` sentinel. The sentinel needs no ID for placement discovery—the ref and `parentElement` are sufficient—and it should not double as the accessibility target.

The Popover API does not resolve this automatically when the card is opened imperatively. `showPopover({source})` affects focus navigation order but does not create the implicit `aria-details` / `aria-expanded` relationship that declarative `popovertarget` provides.

We should define a dedicated accessibility contract for rich HoverCard content rather than retaining the sentinel ID solely for `aria-describedby`.

## Reproduction

Render a HoverCard with interactive content:

```tsx
<HoverCard
  label="Profile actions"
  content={<Button onClick={handleFollow}>Follow</Button>}>
  <button type="button">Jane Doe</button>
</HoverCard>
```

1. Inspect the trigger before opening the card.
2. Observe that `aria-describedby` references the layer ID (or, with #5039, an empty `<template>` sentinel carrying that ID).
3. Open the card and inspect the accessibility tree.
4. Observe that the rich card is treated as a description rather than a structured popup relationship.

## Expected behavior

- Rich HoverCard content is exposed as an appropriately named, structured popup.
- The trigger communicates the popup relationship and open state without flattening interactive content into its accessible description.
- A closed HoverCard does not reference an empty sentinel as its description.
- The sentinel remains available for inline-host discovery but does not require an ID.

## Direction to investigate

- Use consistent, named non-modal dialog semantics for rich HoverCard content.
- Evaluate `aria-haspopup="dialog"`, `aria-expanded`, and `aria-details` and/or `aria-controls` on the trigger.
- Define naming behavior when no `label` is supplied; the current fallback role is `group`.
- Keep `showPopover({source})` for focus ordering, but do not treat it as an ARIA relationship.
- Validate the chosen relationship with VoiceOver/Safari and NVDA with Chrome/Firefox, especially `aria-details` support.

## References

- [WAI-ARIA APG: Tooltip Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/tooltip/)
- [MDN: `aria-describedby`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-describedby)
- [MDN: `aria-details`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-details)
- [MDN: Popover accessibility features](https://developer.mozilla.org/en-US/docs/Web/API/Popover_API/Using#popover_accessibility_features)

## Astryx version

`@astryxdesign/core@0.4.0` / current main and #5039

## Environment

Semantic issue across browsers and assistive technologies; the proposed replacement needs the compatibility validation listed above.

gonzoblasco:
Closes #5049

## Problem

When a HoverCard has a label, the floating layer renders role="dialog" with aria-label. The trigger previously used aria-describedby to point at that dialog, which flattens structured content into a plain-text description and announces the dialog's contents instead of the popup relationship. A screen-reader user on the trigger hears the dialog's content as a description rather than being told "this opens a dialog".

The issue also notes that the lazy context-layer work in #5039 leaves an inert <template> marker while the card is closed, and that marker should not double as the accessibility target. With the old model, the trigger's aria-describedby could point at an empty sentinel.

## Approach

Replace the description relationship with the correct popup contract when a label is provided:

- The trigger gets aria-haspopup="dialog", aria-controls pointing at the layer id, and aria-expanded synced to the open state.
- useHoverCard now exposes id (the layer id) and isOpen, so consumers can wire up aria-controls and aria-expanded themselves when using the hook directly.
- describedBy is kept as a deprecated alias of id for backwards compatibility.
- The no-label fallback still renders role="group"; for that case we keep the previous aria-describedby behavior because aria-haspopup="dialog" would be misleading for an unnamed group. The naming behavior for unlabelled rich HoverCards is left as a follow-up decision in #5049.

aria-details was considered but not used because support across screen readers is still inconsistent; aria-haspopup="dialog" + aria-expanded is the most broadly supported contract for a non-modal dialog opener.

## Test plan

- packages/core/src/HoverCard/HoverCard.test.tsx - 38 tests passing.
- pnpm --filter @astryxdesign/core typecheck - passing.
- pnpm --filter @astryxdesign/core build - passing.
- Updated the HoverCardHookUsage template to demonstrate the new hook contract.

### DIFF
--- a/packages/cli/assets/templates/blocks/components/HoverCard/HoverCardHookUsage.tsx
+++ b/packages/cli/assets/templates/blocks/components/HoverCard/HoverCardHookUsage.tsx
@@ -22,7 +22,7 @@ export default function HoverCardHookUsage() {
         label="Hover profile"
         ref={hoverCard.ref}
         aria-haspopup="dialog"
-        aria-controls={hoverCard.id}
+        aria-controls={hoverCard.isOpen ? hoverCard.id : undefined}
         aria-expanded={hoverCard.isOpen}
       />
       {hoverCard.renderHoverCard(
--- a/packages/cli/assets/templates/blocks/components/HoverCard/HoverCardHookUsage.tsx
+++ b/packages/cli/assets/templates/blocks/components/HoverCard/HoverCardHookUsage.tsx
@@ -20,7 +20,9 @@ export default function HoverCardHookUsage() {
       <Button
         label="Hover profile"
         ref={hoverCard.ref}
-        aria-describedby={hoverCard.describedBy}
+        aria-haspopup="dialog"
+        aria-controls={hoverCard.id}
+        aria-expanded={hoverCard.isOpen}
       />
       {hoverCard.renderHoverCard(
         <VStack gap={1}>
--- a/packages/cli/assets/templates/blocks/components/HoverCard/HoverCardHookUsage.tsx
+++ b/packages/cli/assets/templates/blocks/components/HoverCard/HoverCardHookUsage.tsx
@@ -13,6 +13,7 @@ export default function HoverCardHookUsage() {
     placement: 'below',
     delay: 100,
     isDefaultOpen: true,
+    label: 'Alex Morgan',
   });
 
   return (
--- a/packages/core/src/HoverCard/HoverCard.tsx
+++ b/packages/core/src/HoverCard/HoverCard.tsx
@@ -273,10 +273,20 @@ export function HoverCard({
       const existingExpanded = firstChild.getAttribute('aria-expanded');
 
       firstChild.setAttribute('aria-haspopup', 'dialog');
-      firstChild.setAttribute(
-        'aria-controls',
-        mergeIds(existingControls, hoverCard.id) ?? '',
-      );
+      // Only point aria-controls at the layer while it is open and in the DOM.
+      // While closed, useLayer leaves only an inert <template> marker, so the
+      // id would reference nothing. DateInput gates it the same way. The
+      // trigger's own aria-controls (if any) is preserved either way.
+      if (hoverCard.isOpen) {
+        firstChild.setAttribute(
+          'aria-controls',
+          mergeIds(existingControls, hoverCard.id) ?? '',
+        );
+      } else if (existingControls) {
+        firstChild.setAttribute('aria-controls', existingControls);
+      } else {
+        firstChild.removeAttribute('aria-controls');
+      }
       firstChild.setAttribute('aria-expanded', String(hoverCard.isOpen));
 
       return () => {
@@ -343,7 +353,7 @@ export function HoverCard({
           ref={hoverCard.ref}
           tabIndex={0}
           aria-haspopup={label ? 'dialog' : undefined}
-          aria-controls={label ? hoverCard.id : undefined}
+          aria-controls={label && hoverCard.isOpen ? hoverCard.id : undefined}
           aria-expanded={label ? hoverCard.isOpen : undefined}
           aria-describedby={label ? undefined : hoverCard.describedBy}
           {...stylex.props(
--- a/packages/core/src/HoverCard/HoverCard.tsx
+++ b/packages/core/src/HoverCard/HoverCard.tsx
@@ -264,15 +264,38 @@ export function HoverCard({
       // popup relationship with aria-haspopup/aria-expanded, not describe the
       // trigger with the dialog's content (aria-describedby is for plain-text
       // descriptions, not navigable regions). See #5049.
+      //
+      // Merge rather than overwrite: the trigger may already carry its own
+      // popup semantics (e.g. a menu button wrapped in a labelled HoverCard),
+      // so preserve the existing values and restore them on cleanup.
+      const existingHaspopup = firstChild.getAttribute('aria-haspopup');
+      const existingControls = firstChild.getAttribute('aria-controls');
+      const existingExpanded = firstChild.getAttribute('aria-expanded');
+
       firstChild.setAttribute('aria-haspopup', 'dialog');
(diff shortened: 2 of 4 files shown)


packages/cli/assets/templates/blocks/components/HoverCard/HoverCardHookUsage.tsx
```diff
@@ -20,7 +20,9 @@ export default function HoverCardHookUsage() {
       <Button
         label="Hover profile"
         ref={hoverCard.ref}
-        aria-describedby={hoverCard.describedBy}
+        aria-haspopup="dialog"
```
cixzhang, line 24: No label on the hook, so this card is a `group`. Might need `label` on `useHoverCard`.

packages/core/src/HoverCard/HoverCard.tsx
```diff
@@ -259,7 +259,26 @@ export function HoverCard({
     // Use combined ref for position + interaction
     hoverCard.ref(firstChild);
 
-    // Set aria-describedby, merging with existing
+    if (label) {
+      // When named, the hover card is a dialog. The trigger should advertise the
+      // popup relationship with aria-haspopup/aria-expanded, not describe the
+      // trigger with the dialog's content (aria-describedby is for plain-text
+      // descriptions, not navigable regions). See #5049.
+      firstChild.setAttribute('aria-haspopup', 'dialog');
```
cixzhang, line 275: Overwrites whatever the trigger already had. The describedby branch below merges — probably wants the same here.

packages/core/src/HoverCard/HoverCard.tsx
```diff
@@ -259,7 +259,26 @@ export function HoverCard({
...
+      firstChild.setAttribute('aria-controls', hoverCard.id);
```
cixzhang: Points at an id not in the DOM while closed. DateInput gates it on `popover.isOpen`.

cixzhang (changes requested):
Thanks for this — the popup contract is right, and it lines up with what DateInput and ComplexSelector already do.

Two things first.

The hook block template puts `aria-haspopup="dialog"` on a trigger whose card has no label, so the card renders `role="group"`. In Chrome the button announces "has dialog popup" and opens an unnamed group — the mismatch your own description rules out. `useHoverCard({label: 'Alex Morgan'})` closes it.

The labelled branch also overwrites the trigger's own `aria-haspopup`, `aria-controls` and `aria-expanded`, and removes them on cleanup. Wrap a menu button in a labelled HoverCard and its menu semantics are gone with nothing saying so; the no-label branch fifteen lines down merges and restores instead.

Can you pass a label in the template and merge rather than overwrite on the trigger?

[Reviewed by Robohands]

packages/cli/assets/templates/blocks/components/HoverCard/HoverCardHookUsage.tsx
gonzoblasco: Good catch. Added `label: 'Alex Morgan'` to the hook so the card renders as a named dialog instead of an unnamed group.

packages/core/src/HoverCard/HoverCard.tsx
gonzoblasco: Fixed. The labelled branch now reads the trigger's existing `aria-haspopup`, `aria-controls` and `aria-expanded`, merges `aria-controls` with `mergeIds`, and restores all three on cleanup - mirroring the describedby branch below. Added a test wrapping a menu button (with its own popup attributes) in a labelled HoverCard to lock this in.

packages/core/src/HoverCard/HoverCard.tsx
gonzoblasco: Fixed. `aria-controls` is now gated on the open state, matching DateInput: it only points at the layer id while the card is open and in the DOM, and is cleared (or restored to the trigger's own value) while closed. Updated the open/close test to assert this.

gonzoblasco:
Hi, @cixzhang ! I've addressed all three review comments and pushed the updates:

1. **Template label** - added `label: 'Alex Morgan'` to the hook so the card renders as a named dialog instead of an unnamed group.
2. **Merge instead of overwrite** - the labelled branch now reads the trigger's existing `aria-haspopup`, `aria-controls` and `aria-expanded`, merges `aria-controls` with `mergeIds`, and restores all three on cleanup, mirroring the describedby branch. Added a test wrapping a menu button (with its own popup attributes) in a labelled HoverCard to lock this in.
3. **aria-controls pointing at a missing id** - now gated on the open state, matching DateInput: it only points at the layer id while the card is open and in the DOM, and is cleared (or restored to the trigger's own value) while closed.

Commits: `52262da7` (merge + template label) and `c1603cad` (gate aria-controls on open state). All CI checks pass, including the a11y check. Could you re-review when you get a chance? Thanks!

cixzhang (approved):
Thanks — merged rather than overwritten, and the template names the card now. Verified in Chrome: a menu trigger keeps its own semantics, and cleanup restores them. Merging.

[Reviewed by Robohands]

PR #5420, opened 2026-08-24
feat(Table): expose theme targets for the sort and filter affordances

freddymeta:
Implements the third row of #5417 — the one with no consumer-side workaround at all.

## Problem

`useTableSortable` rendered its control as a bare `<button>` wrapping a bare `<span>` wrapping the `Icon`; `useTableFiltering` rendered its funnel as a bare `<button aria-haspopup="dialog">`. Neither carried an `astryx-*` class, so no token and no `components` entry could reach them — and because both are rendered *inside* the plugin, there is no wrapper a consumer can interpose and no `renderX` prop to route around it. The only remaining move was `.astryx-table-header-cell button[aria-haspopup='dialog']`, a selector keyed on an ARIA attribute.

Both also re-implemented the interaction states of a button on a native `<button>`, which is what @rubyycheung caught.

## Change

**Both affordances are now the real `Button` / `IconButton`.** Rest, hover, pressed, the focus ring, the `scale(0.98)` press and the reduced-motion guard are Button's; every hand-rolled copy is deleted. The column label stays inside the sort control, so the whole heading is still the click target.

**Each keeps one table-scoped theme target on the control Button renders**, the `toggle-button` shape:

| target | element | reflects |
|---|---|---|
| `astryx-table-sort-button` | the sort control | `direction` |
| `astryx-table-filter-button` | the funnel trigger | `active` |

INV5 says a part rendered by a shared primitive delegates to that primitive's target "unless the parent guarantees a distinct public visual contract". Both do, and `defineTheme` is why: `parseStyleKey` emits **one compound selector on one element**, no combinators, so a theme cannot write `.astryx-table-header-cell .astryx-button`. Delegating alone would mean "restyle every button in the app" or nothing. Beyond scope, `color` behaves differently on the two contracts — see below.

**`color` on the sort target paints the glyph and leaves the column name alone.** The control holds both, and the name belongs to the header cell. The value is routed to the glyph through a derived-var entry with `replaces: true`, so `defineTheme` emits `--_table-sort-glyph-color` and no `color` — the `progress-bar-mark` / `text-area` mechanism, and INV11 rather than a new public var. The filter trigger holds only its glyph, so plain `color` lands on it directly.

## Records this moves with

`Table.spec.md`'s `anatomy-theming:v1` map dispositioned `Sort control` as `inherits: table-header-cell` and had no filter anatomy at all. Per "Change coupling", the map, the anatomy inventory and the targets move together, so this PR updates all three: `Sort control` and a new `Filter control` become `target`, `Filter indicator glyph` joins `Sort indicator glyph` in delegating to Icon, and the prose records why the two targets survive INV5.

The targets are declared in `Table.doc.mjs` rather than the plugin docs, because that is the doc `check-knowledge.mjs` reads for `component:Table`'s target inventory — declared on the plugin docs they were invisible to the anatomy↔target validator.

## Visual changes

Not "no visual change" — the changeset lists them. Hover and pressed are now Button's overlay (a background image, not a background colour) plus `scale(0.98)`; the filter trigger gains the shared focus ring and Button's `sm` square box (28×28 measured, above WCAG 2.2 2.5.8's 24px); the sort control's label is measured identical to its `<th>` on colour, white-space, overflow, text-overflow, font size and weight, so headings are unchanged.

## Test plan

- `pnpm exec vitest run packages/core/src packages/lab/src`, `pnpm -F @astryxdesign/core build`, `typecheck`, `typecheck:docs`, `pnpm check:repo`, `pnpm lint:strict` — output quoted in the review reply.
- `.github/scripts/affordance-color-reach.js` against a built Storybook, in Chromium. The probe stylesheet is no longer hand-written: it is what `defineTheme` + `generateThemeCSS` emit for a theme setting `color` on the target, so the derived-var expansion is exercised rather than assumed.
- Negative controls, both quoted in the reply: removing the derived-var entry makes the guard report the colour landing on the button, the label following it and the glyph unmoved — the exact failure @cixzhang described; forcing a hardcoded `stroke` on the geometry makes it report the painted glyph ignoring the target.
- `check-knowledge.mjs` negative control: reverting `Sort control` to `inherits` while the target stays declared fails with `current target "table-sort-button" has no anatomy entry with a target disposition`.

## Still needs a maintainer

The CI wiring for the guard edits `.github/workflows/ci.yml`, which my token cannot push (no `workflow` scope). The block is in the PR discussion.


cixzhang:
Thanks — the problem is real and this is the one row in [#5417](https://github.com/facebook/astryx/issues/5417) with no consumer-side workaround at all. Keeping the guard fix here is right too; I checked, and the old guard stays green with these three unchecked, so it's load-bearing rather than scope creep.

One change to the shape before it lands, and it's the same call we just made on the Spinner in [#5408](https://github.com/facebook/astryx/pull/5408).

**Make it one target, not two.** Style the sort affordance by reaching the button; have the icon paint `currentColor` and take its colour from the button. Hover then works for free — the button's colour changes and the glyph follows — instead of a theme having to write the wrapper and the button separately and keep them in step.

The rule we're settling on: a control that owns its background owns its `color`. `Button` and `CheckIndicator` already work this way, the Spinner's ring now does, and it's what [#5250](https://github.com/facebook/astryx/pull/5250) is waiting on.

```
components: {
  'table-sort-button': {
    base:  {color: 'var(--eps-text-secondary)', opacity: '1'},
    hover: {color: 'var(--eps-accent)'},
  },
}
```

**And a test that guarantees the behaviour, not the wiring.** Set a colour on the button target, assert the glyph paints it; hover, assert it follows. It has to be a browser test — jsdom resolves no cascade, so an assertion on the class name proves nothing.

One thing I want your read on rather than assuming: the opacity currently lives on the indicator wrapper (`iconWrapperUnsorted` / `iconWrapperActive`), and legibility at rest is what EPS is actually here for. If that opacity genuinely needs its own element and can't ride on the button, say so — that's the case against this and I'd rather hear it than have it worked around.

`table-filter-button` looks unaffected either way.

[Reviewed by Robohands]

packages/core/src/Table/plugins/sortable/useTableSortable.tsx
```diff
@@ -306,15 +308,21 @@ function SortHeaderButton<T extends Record<string, unknown>>({
...
-            : sortStyles.iconWrapperUnsorted,
+        {...mergeProps(
+          themeProps('table-sort-indicator', {direction}),
+          stylex.props(
+            direction != null
+              ? sortStyles.iconWrapperActive
+              : sortStyles.iconWrapperUnsorted,
+          ),
         )}>
         <Icon
```
cixzhang: `color="inherit"` on this one (it's `secondary` two lines down) and the button's colour reaches the glyph — the registry SVGs are already `currentColor`.

packages/core/src/Table/plugins/sortable/useTableSortable.tsx
```diff
@@ -306,15 +308,21 @@ function SortHeaderButton<T extends Record<string, unknown>>({
...
-        {...stylex.props(
-          direction != null
-            ? sortStyles.iconWrapperActive
-            : sortStyles.iconWrapperUnsorted,
+        {...mergeProps(
+          themeProps('table-sort-indicator', {direction}),
+          stylex.props(
+            direction != null
+              ? sortStyles.iconWrapperActive
+              : sortStyles.iconWrapperUnsorted,
```
cixzhang: This resting `opacity: 0.35` is what puts the arrows at 1.57:1. Drop it and let `secondary` render — the header label span then needs its own colour so the button's doesn't bleed into it.

packages/core/src/Table/plugins/filtering/useTableFiltering.tsx
```diff
@@ -916,9 +918,16 @@ function PopoverFilterTrigger({
...
+            active: hasValue ? 'active' : null,
+          }),
+          stylex.props(
+            filterStyles.triggerButton,
+            hasValue
+              ? filterStyles.triggerActive
+              : filterStyles.triggerInactive,
+          ),
         )}>
         <Icon
```
cixzhang: Same here — the funnel's `Icon` sets its own colour, so the single target above doesn't reach it.

cixzhang (changes requested):
Thanks — the problem is right, and keeping the guard fix here is right too; I checked and the old one stays green with these three unchecked.

The two sort targets are one axis, and neither carries the glyph. Set a colour on `astryx-table-sort-button` — the obvious one — and nothing happens; the funnel has the same problem with only one target. Both icons want `color="inherit"`, then the colour goes on the buttons and `astryx-table-sort-indicator` comes out. Spinner went this way in #5408.

| colour on the button today | with the label pinned |
|---|---|
| header text follows too | only the glyph |

Drop the resting `opacity: 0.35` with it — inline below. Composited, the arrows sit at 1.57:1 against the header where AA wants 3:1; `--color-icon-secondary` alone is 4.74:1.

Then a browser test to hold it: colour on the button target, assert the glyph paints it, hover, assert it follows. `theme-var-reachability.js` is the shape.

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks — my last review asked for one target per affordance, button-owned color, inherited SVG paint, no resting opacity, and a browser guard. The runtime work now matches all five asks.

The guard is still the hold. It is not called by CI, and on this head it reads before Storybook renders and fails both cases. After waiting, it checks `svg.color`, so a fixed stroke still passes while the glyph paints the wrong color. Could a maintainer wire it into the Storybook Chromium job, wait for each target, and assert the painted stroke/fill?


[Reviewed by Robohands]

rubyycheung (changes requested):
Keeping the sort and filter affordances visible at rest is the right accessibility improvement, but these are buttons and should mirror the standard Button/IconButton interaction states. The current change removes the distinct pointer-hover feedback entirely.

Please apply the shared hover and pressed treatments, retain the focus-visible ring for keyboard users, and reserve the accent styling for the active sorted/filtered state. This gives the controls a clear rest → hover → pressed interaction model without returning to the inaccessible 35% resting opacity.

freddymeta:
Rebased onto main, and both reviews addressed — except the one line I am not permitted to push, which is below ready to paste.

## @rubyycheung — the interaction model

You were right, and it was worse than a missing hover: dropping `opacity: 0.35` fixed the contrast and removed the *only* hover feedback these had. The filter button also drew nothing on keyboard focus — no ring at all, ever.

Both now draw what every other borderless control draws. Measured in Chromium on the built Storybook:

| | rest | hover |
|---|---|---|
| sort button bg | `rgba(0, 0, 0, 0)` | `rgba(0, 0, 0, 0.05)` |
| sort glyph | `rgb(115, 115, 115)` | `rgb(23, 23, 23)` |
| filter button bg | `rgba(0, 0, 0, 0)` | `rgba(0, 0, 0, 0.05)` |
| filter glyph | `rgb(115, 115, 115)` | `rgb(23, 23, 23)` |

Filter button focus ring, which previously fell back to the UA default `1px auto`:

```
outline: 2px solid rgb(38, 38, 38)   ← the shared Astryx ring
```

- hover → `--color-overlay-hover`, pressed → `--color-overlay-pressed`, focus → `focusOutlineProps.focusVisible`, all shared rather than restated;
- guarded on `@media (hover: hover)` so a touch device does not stick in the tint after a tap;
- the glyph darkens a step alongside the tint, so the feedback survives a forced-colors mode that drops backgrounds;
- **accent stays reserved for sorted/filtered**, so "this column is sorted" never reads as "the pointer is here";
- resting colour unchanged at 4.74:1 — no return to the 35% dimming.

## @cixzhang — the guard

**It reads the painted stroke/fill now, not `svg.color`.** You were exactly right that a fixed stroke slipped through: `color` is what the glyph *inherits*, and geometry that hardcodes a paint leaves it reading the sentinel while nothing on screen moves. It now collects the resolved `stroke`/`fill` off every painted node and asserts on those, at rest and under hover, and measures the resting contrast on the paint too.

Verified both directions:

```
green on this head:
  ✓ Table sort affordance — rest rgb(115,115,115) on rgb(255,255,255) — 4.74:1;
    themed → painted rgb(1,2,3); hovered → painted rgb(1,2,3);
    the header label held at rgb(82,82,82)
  ✓ Table filter affordance — … same

with stroke forced to rgb(9,9,9):
  ✗ …did not reach the PAINTED glyph — stroke/fill stayed rgb(9,9,9)
    while the inherited color read rgb(1,2,3)
```

That second sentence is one the old version could not produce.

**And it waits for the story now** — `networkidle` returns before Storybook mounts, which is why it failed both cases on the last head. It waits for the target with a glyph inside it.

One thing the paint check surfaced: these affordances transition `color`, so the read taken right after the sentinel landed was catching an interpolated value (`rgb(61, 62, 62)`, mid-flight). The probe kills transition and animation durations rather than sleeping through them — the assertion is about the settled paint, and a sleep long enough for one theme's duration is a race in another.

## The CI wiring — needs a maintainer

I tried; the push is refused: `refusing to allow a Personal Access Token to create or update workflow .github/workflows/ci.yml without workflow scope`. So it is genuinely not mine to land. It goes immediately after the theme-var-reachability step (`ci.yml:514`), in the same Chromium job that already has the built Storybook:

```yaml
      # A theme target that names a button holding one icon promises that a
      # colour set on it reaches the icon. An `Icon` naming its own colour
      # writes `color` on the glyph and beats the button's inherited value, so
      # the target resolves, the rule emits, and nothing moves — Table's sort
      # and filter affordances both shipped that way (#5420). jsdom resolves no
      # cascade, so a unit test can assert the class and the rule while the
      # glyph still paints its own colour; this reads the painted stroke/fill
      # in Chromium, at rest and under hover, and pins the resting contrast
      # against WCAG 1.4.11.
      - name: Run affordance colour reach guard
        run: node .github/scripts/affordance-color-reach.js --storybook-dir apps/storybook/dist
```

`pnpm guard:affordance-color` is already in `package.json` if you would rather invoke it that way.

freddymeta:
Before/after evidence, captured in Chromium at 3× against a real Storybook build of each side. BEFORE is the merge-base (`c6041cfd5`), AFTER is this head.

### At rest — @rubyycheung's contrast point, and what it cost

rest

| | opacity | glyph | composited contrast |
|---|---|---|---|
| before | **0.35** | `rgb(82,82,82)` | **1.57:1** — below WCAG 1.4.11's 3:1 |
| after | 1 | `rgb(115,115,115)` | **4.74:1** |

### Hover — the regression you caught

hover

| | background | glyph |
|---|---|---|
| before | `rgba(0,0,0,0)` | `rgb(82,82,82)` |
| after | `rgba(0,0,0,0.05)` | `rgb(23,23,23)` |

Before, hovering the filter control changed **nothing** — computed background and colour are byte-identical to rest. The 35% dimming had been doing double duty as the hover affordance, so removing it for contrast left no pointer feedback at all. That is exactly what you flagged, and the screenshots show it.

### Keyboard focus — a ring that was never there

focus

| | outline |
|---|---|
| before | `1px auto rgb(16,16,16)` — the **user-agent default**, not ours |
| after | `2px solid rgb(38,38,38)` — the shared Astryx ring |

The filter trigger had no `focusOutlineProps` at all; it fell through to whatever the browser draws. It now uses the same ring as every other control.

Hover is guarded on `@media (hover: hover)` so a touch device does not stick in the tint after a tap, and the glyph darkens alongside the tint so the feedback survives a forced-colors mode that drops backgrounds. Accent stays reserved for the sorted/filtered state.

Individual frames (rest / hover-sort / hover-filter / focus, both sides) are on the `assets-pr-5420` branch under `pr-5420/` if you want to flip between them.

rubyycheung (commented):
The revised interaction states look correct, but both controls still recreate Button behavior on native buttons, so they can drift when the shared Button theme or interaction model changes. Could the sortable heading use the regular Button component—with the column label as its visible content, the sort glyph as end content, and table-scoped compact styling—and could the filter trigger use IconButton with the same scoped sizing? The sort label should remain inside the button so the full heading stays clickable. The icon may remain 20px visually, while the target should meet WCAG 2.2’s 24px minimum or its spacing exception; 44px is not an AA requirement. If these components cannot support the table geometry, please document the concrete limitation and extract/reuse a shared interaction primitive rather than duplicating the states here.

cixzhang (changes requested):
The unified target direction is good, but exact-head behavior still has regressions: sort/filter transitions ignore reduced motion, sortable labels bypass the existing header-cell theme color, and the new guards neither close the nested-target gap nor run in CI.

Please preserve the single target while fixing its state/theme behavior, and split or properly wire the unrelated guard work.

[Reviewed by Robohands]

freddymeta:
Three of the four addressed; the fourth needs a maintainer's hands.

## The label bypass — you were right, and it was a duplicate

The sort button set `color` so its glyph could inherit it. That button also holds the column name, so the name inherited it too — and to stop the affordance repainting the heading, the label restated `--color-text-secondary`, the same token `TableHeaderCell` already paints on the `<th>`. Identical output until someone themes `astryx-table-header-cell`, and then the cell moves and every sortable heading stays behind.

The glyph's colour now travels as `--table-sort-glyph-color` on the **same single target**, and the button sets no `color` at all — nothing interrupts the cascade, so the label inherits the header cell and the pin is gone. Hover and the sorted accent move to the var, so neither touches the name.

Measured in Chromium against a built Storybook, theming `astryx-table-header-cell` to `rgb(200, 0, 100)`:

| | `<th>` | sort label | glyph |
|---|---|---|---|
| before | `rgb(200,0,100)` | **`rgb(82,82,82)`** | unmoved |
| after | `rgb(200,0,100)` | **`rgb(200,0,100)`** | unmoved |

## Reduced motion

`transitionDuration` was unconditional on both. Guarded the way `Banner` does it; under `prefers-reduced-motion: reduce` the sort button went `0.125s` → `0s`, the filter button likewise.

## The guard now asserts the contract that exists

It injected plain `color` on both targets. For the filter button that is the whole contract — it holds nothing but its glyph. For the sort button it is not, for the reason above, so that case now injects the documented var. Green on the built Storybook:

```
✓ Table sort affordance — rest rgb(115,115,115) on rgb(255,255,255) — 4.74:1;
  themed → painted rgb(1,2,3); hovered → painted rgb(1,2,3);
  the header label held at rgb(82,82,82)
✓ Table filter affordance — … same
```

## The nested-scan work is split out

It is guard infrastructure, not this surface. Now its own branch, `guard/affordance-color-reach`.

On it not closing the gap — you were right, and here is the precise reason. `themingTargets.test.ts` does now see nested sources (instrumented: `core=98, lab=5, nested=2`, the two being `Table/plugins/filtering` and `Table/plugins/sortable`). Its sibling `derivedVarRegistry.test.ts` does not. I wrote the same walk for it, measured it, and it discovered **nothing** — `nested=0` — because that guard also requires a doc file named after the directory, and `Table/plugins/sortable/` has no `sortable.doc.mjs`. Closing it needs the doc-fallback lookup the other file already has. I did not ship the walk alone: it would have been a rule that never fires.

## The one I cannot do

The CI step is written and verified, but pushing it edits `.github/workflows/ci.yml` and my token has no `workflow` scope — the push is rejected outright. It belongs in `pr-a11y` beside the two Chromium guards, right after `Run theme var reachability guard`:

```yaml
+++ /tmp/ci-with-guard.yml	2026-08-31 07:43:47.518514103 -0700
@@ -564,6 +564,17 @@
       - name: Run theme var reachability guard
         run: node .github/scripts/theme-var-reachability.js --storybook-dir apps/storybook/dist
 
+      # A theme target naming a button that holds one glyph promises that
+      # colouring it reaches that glyph. An `Icon` naming its own colour writes
+      # `color` on the glyph and beats what the button inherits down — the
+      # target resolves, the rule emits, and nothing moves. jsdom resolves no
+      # cascade, so a unit test can assert the class is on the button and the
+      # rule is in the CSS text while the glyph still paints its own colour.
+      # Sits beside the two above for the same reason: a cascade fact, only
+      # checkable in a real browser against a built Storybook.
+      - name: Run affordance colour reach guard
+        run: node .github/scripts/affordance-color-reach.js --storybook-dir apps/storybook/dist
+
   # The layer order Astryx promises is Astryx < theme < product, and which
   # declaration paints is a cascade fact no jsdom test can settle: the build has
   # been wrong in both directions inside one week — unsplit, so StyleX outranked
```

@rubyycheung — your Button/IconButton question is not addressed here; I did not want to fold a component-substitution spike into this. Taking it next unless you would rather it were its own PR.

cixzhang (changes requested):
Thanks — the reduced-motion and header-label regressions are fixed. A theme author who [sets `color` on the sort target](https://github.com/facebook/astryx/blob/db8c0fd06305a11925359dfc6bd012c3fe21df42/packages/core/src/Table/plugins/sortable/useTableSortable.tsx#L141-L214) still repaints the column label while the glyph stays unchanged. The guard passes only because it [writes the private `--_table-sort-glyph-color` directly](https://github.com/facebook/astryx/blob/db8c0fd06305a11925359dfc6bd012c3fe21df42/.github/scripts/affordance-color-reach.js#L56-L78); it is not in CI, and the committed probe theme is stale.

Could we make the real `defineTheme` path color the glyph across rest/hover/direction without moving the label, regenerate the probe theme, and land/wire the [nested-source coverage in #5784](https://github.com/facebook/astryx/blob/32a554c7eb54c3b1010e94053b58c3c3181ce282/packages/core/src/theme/themingTargets.test.ts#L380-L454)? The PR body and [changeset](https://github.com/facebook/astryx/blob/db8c0fd06305a11925359dfc6bd012c3fe21df42/.changeset/table-affordance-theme-targets.md#L5-L19) also need to describe the intentional visual changes rather than “no visual change.”

[Reviewed by Robohands]

freddymeta:
*(Replaces my previous comment, which was far too long.)*

Both reviews came down to one question — **who owns these two controls** — and the answer is `toggle-button`'s shape: the real primitive, carrying one target of its own. Head `864ecc8c4`, merged from main.

**@rubyycheung — now the real `Button` / `IconButton`.** You were right about drift, and it was already measurable: the hand-rolled `:hover`/`:active` used `background-color`, where the shared `interactionOverlayStyles` uses a `background-image` gradient. Sort is `<Button variant="ghost" size="sm">` (label as children, glyph as `endContent`); filter is `<IconButton>`. Every hand-rolled state deleted. Filter trigger measures 28×28 (>24px, WCAG 2.5.8). Accent stays reserved for sorted/filtered.

**@cixzhang — `color` on the target no longer repaints the label.** A `derivedVarRegistry` entry with `replaces: true` makes `defineTheme` emit `--_table-sort-glyph-color` and **no `color`**, so the heading keeps inheriting `astryx-table-header-cell`:

| theme `table-sort-button { color }` | glyph | column label |
|---|---|---|
| before | moves | **moves too** ← the bug |
| after | moves | stays with the `<th>` |

Guard now builds its probe with the real `defineTheme` + `generateThemeCSS` instead of writing the private var. Probe theme regenerated (274 targets). #5784 merged and arrives via main.

**Why keep the targets rather than delegate.** `parseStyleKey` emits one compound selector on one element — no combinators. So delegation offers a themer `.astryx-button` (every button in the app) or nothing; `.astryx-table-header-cell .astryx-button` isn't expressible. That's INV5's "distinct public visual contract" met by construction, and `toggle-button` is the in-repo precedent.

**Your call, and the one thing I'd flag:** this amends the `Table.spec.md` anatomy map you approved on 30 Aug — `Sort control` moves from `inherits` to its own target, and filter parts are added. If you'd rather keep the map as approved, the alternative is dropping both targets and delegating to Button globally; say the word and I'll do that instead.

Negative controls: remove the derived-var entry → the label follows the colour again (your exact bug); revert `Sort control` to `inherits` → `check-knowledge` fails with "no anatomy entry with a target disposition".

Gates: 8957 tests pass, build/typecheck/typecheck:docs/check:repo clean, `lint:strict` 0 errors (86 warnings, none in touched files). `pr-rtl` is the repo-wide soft-red.

Still blocked on me: the guard's CI step needs `.github/workflows/ci.yml`, which my token can't push (no `workflow` scope) — block is in the thread for a maintainer.

cixzhang (changes requested):
Thanks — the Button/IconButton move fixes the interaction drift. Two things still block this head.

The new affordance guard is only a package script; no workflow invokes it, so the target/color contract can regress without CI noticing. Please wire it into the existing Storybook Chromium job.

`Table.spec.md` also still says “documentation only,” eight targets, and no runtime/target change while this head records ten targets and changes the controls. Please align its compatibility, ownership, verification, and decision sections.

[Reviewed by Robohands]

freddymeta:
Both asks addressed. The spec half is pushed as `9c0da1e1b`; the CI half is a patch below, because my account cannot push it.

## Table.spec.md

It still read "additive documentation only", "eight current targets" in three places, and a Decision log of "None ... without introducing a component-local visual or runtime change".

- **Compatibility** — additive targets plus the intentional visual changes, itemised: Button/IconButton's hover and pressed overlay and `scale(0.98)` press, the filter trigger's shared focus ring and `sm` square box, and the removal of the resting `opacity: 0.35`. Records that nothing is removed or renamed, and that the column heading is measured unchanged against its `<th>`.
- **Ownership**, **FR4**, **verification map** — eight to ten, and FR4 now names `table-sort-button` and `table-filter-button` with the states they reflect.
- **Decision log** — four decisions replacing "None": own targets rather than delegating to `button` (DEC-1), the derived-var routing that colours the glyph without the label (DEC-2), building on Button (DEC-3), the opacity removal as a contrast fix rather than a preference (DEC-4).

Editing an `authority: current` record means `spec-owner-approval` will now ask for your approval where the previous head did not. That is the gate working; I have not touched the approval metadata.

## The guard in CI — needs you

I cannot push it:

```
! [remote rejected] refusing to allow a Personal Access Token to create or
  update workflow `.github/workflows/ci.yml` without `workflow` scope
```

Your 08-26 review already framed this as a maintainer action, so here is the patch, verified locally rather than sketched.

It goes in `pr-a11y`, beside `modal-close-visibility` and `theme-var-reachability`, for the reason they are there — the built Storybook is already in hand. **One wiring detail that is easy to miss:** the guard builds its probe stylesheet with the real `defineTheme` out of `packages/core/dist`, and `pr-a11y` builds nothing, so without the extra download it dies on its own "Built core not found" check. `build-storybook` already publishes `dists-<hash>` with core's dist for pr-visual, so this reuses it — no second build.

<details>
<summary>The patch</summary>

```diff
diff --git a/.github/workflows/ci.yml b/.github/workflows/ci.yml
index a2ad46c07..cc7a75bfd 100644
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -525,6 +525,17 @@ jobs:
           name: storybook-${{ needs.build-storybook.outputs.short_hash }}
           path: apps/storybook/dist
 
+      # For the affordance-colour guard below, which generates its probe
+      # stylesheet with the real `defineTheme` out of core's dist rather than
+      # hand-writing one. build-storybook has already built it; downloading
+      # keeps this job on an artifact instead of a second full build, the same
+      # trade pr-visual makes.
+      - name: Download built themes and core
+        uses: actions/download-artifact@v8
+        with:
+          name: dists-${{ needs.build-storybook.outputs.short_hash }}
+          path: packages
+
       - name: Install Playwright
         run: pnpm install --frozen-lockfile && npx playwright install chromium
 
@@ -569,6 +580,18 @@ jobs:
       - name: Run theme var reachability guard
         run: node .github/scripts/theme-var-reachability.js --storybook-dir apps/storybook/dist
 
+      # A theme target naming a button that holds one icon promises that a
+      # colour set on it reaches the glyph. An `Icon` that names its own colour
+      # writes `color` on the glyph and beats what the button inherits down, so
+      # the target resolves, the rule emits, and nothing moves — which is how
+      # Table's sort and filter affordances shipped. jsdom resolves no cascade,
+      # so a unit test can assert the class is on the button and the rule is in
+      # the CSS text while the glyph still paints its own colour; Chromium is
+      # where the promise is checkable. Sibling to the two guards above, and
+      # here for the same reason: this job already has the built Storybook.
+      - name: Run affordance colour reach guard
+        run: node .github/scripts/affordance-color-reach.js --storybook-dir apps/storybook/dist
+
   # The layer order Astryx promises is Astryx < theme < product, and which
   # declaration paints is a cascade fact no jsdom test can settle: the build has
   # been wrong in both directions inside one week — unsplit, so StyleX outranked
```

</details>

## Verified, not asserted

Ran it exactly as the new step invokes it, against a real Storybook build:

```
$ node .github/scripts/affordance-color-reach.js --storybook-dir apps/storybook/dist
✓ Table sort affordance — rest rgb(115,115,115) on rgb(255,255,255) — 4.74:1;
  themed → painted rgb(1,2,3); hovered → painted rgb(1,2,3);
  data-direction="ascending" → painted rgb(1,2,3);
  the header label held at rgb(82,82,82)
✓ Table filter affordance — rest 4.74:1; themed → painted; hovered → painted
All 2 affordance targets colour their own glyph.                      exit 0
```

The 4.74:1 and the held label are the two things the last two rounds doubted — measured here.

**Negative control**, since this guard's integrity has been questioned twice. Restored the original defect (`color="secondary"` on the sort Icon), rebuilt core and Storybook, reran:

```
✗ ... did not reach the PAINTED glyph — stroke/fill stayed rgb(115,115,115)
✗ ... the themed colour is lost on hover
✗ ... the themed colour is lost once the control reflects a state
Failing: 1 affordance target(s).                                      exit 1
```

So it fails on the defect it exists for, in all three of rest, hover and state, and it is not passing by writing the private var directly.

Also: `check-knowledge.mjs` aligned; `themingTargets.test.ts` + `derivedVarRegistry.test.ts` 518 tests pass.

cixzhang (changes requested):
Thanks — the spec now matches the implementation. This exact head still defines `guard:affordance-color`, but no workflow invokes it, so the target/color contract can regress without CI noticing. The prepared `pr-a11y` wiring still needs to land before merge.

[Reviewed by Robohands]

freddymeta:
Follow-up on `f45cefe7e` — CI is green again (21 pass; only the review gates pending).

**The branch was `DIRTY`, which is why CI had stopped running entirely.** `targets.json` conflicted with main, where Typeahead and Tokenizer D2 targets have since landed. Merged, kept both sides, rebuilt, and re-ran both guards on the merged tree:

```
affordance-color-reach:  All 2 affordance targets colour their own glyph.   exit 0
rtl-audit --filter Table: COV 1 measured / 0 gap                            exit 0
```

**`pr-rtl` was also failing** on the previous head, unrelated to your review — `COV: 0 measured / 1 gap`, since the audit requires every touched component to be measured or carry a written reason and Table was neither. Added a D2 target over `core-table--column-alignment`: column order genuinely mirrors, cell centres LTR 305 &lt; 1024, RTL 795 &gt; 76.

One trap worth passing on, because the selector looks obviously right and is not: `tbody tr:first-child td:first-child` matches Storybook’s own hidden docs prop-table first (inside `.sb-preparing-docs`, width 0), so `boundingBox()` returns null and the dimension reports “prev/next not found” — an N-A that still counts as a gap rather than erroring. Scoping to `.astryx-table` fixes it.

For the record, `pr-rtl` carries `continue-on-error: true`, so it was never blocking this PR — I fixed it because it was red, not because it gated.

Still needs you for the two things I cannot do: the `ci.yml` patch in my previous comment (no `workflow` scope on my token), and `spec-owner-approval`, which the `Table.spec.md` edit now requires.

.github/scripts/affordance-color-reach.js
```diff
@@ -0,0 +1,479 @@
+#!/usr/bin/env node
```
cixzhang, line 1: Hmm do we need this script? Could we clean it up? I think for any theming visual updates we should be using the visual regression tests.

package.json
```diff
@@ -37,6 +37,7 @@
     "lab:readiness:check": "node internal/lab-readiness/audit.mjs --check",
     "guard:modal-close": "node .github/scripts/modal-close-visibility.js",
     "guard:theme-vars": "node .github/scripts/theme-var-reachability.js",
+    "guard:affordance-color": "node .github/scripts/affordance-color-reach.js",
```
cixzhang, line 40: Remove

packages/core/src/Table/plugins/sortable/useTableSortable.tsx
```diff
@@ -303,31 +353,35 @@ function SortHeaderButton<T extends Record<string, unknown>>({
...
-        <Icon
-          icon={iconName}
-          size="xsm"
-          color={direction != null ? 'accent' : 'secondary'}
-        />
-      </span>
-      {rank != null && (
-        <span {...stylex.props(sortStyles.rank)} aria-hidden="true">
-          {rank}
+    <Button
```
cixzhang, line 363: This seems like it'll be incorrect visually. We wouldn't want the whole cell to appear as a button but only the sort icon and rank.

cixzhang (changes requested):

humbertovirtudes (changes requested):
Using Button and IconButton exposes Table controls to existing generic Button theme overrides that can change header geometry. The purpose-built browser guard is also not invoked by required CI. Please contain the theme targets, preserve Table geometry across neutral, Butter, and Y2K themes, and wire the guard into the Storybook Chromium job before approval.

PR #5426, opened 2026-08-24
feat(CheckboxInput): expose theme targets for the control and the marks

freddymeta:
Implements the second row of #5417 — the CheckboxInput entry.

## Problem

`CheckboxInput` registers three targets: `checkbox-input` on the root, and `checkbox-indicator` (plus its legacy alias `checkbox`) on the painted box. Two elements in between carry nothing: the box **slot** (`styles.checkboxWrapper`) and the native `<input type="checkbox">`. Both are sized from `wrapperSizeStyles` — literal 20px (`sm`) / 24px (`md`) atomics, with no token or var to re-scope — so a theme with a single checkbox size cannot reach them at all. At `size="md"` that leaves a themed 20px box sitting in an unthemed 24px slot.

Same story one level down: the indicator's checkmark `<svg>` and indeterminate `<span>` are sized from per-size atomics with no target, even though the radio's equivalent mark already has one (`astryx-radio-indicator-dot`).

In `@nest/eps-core` that reads:

```css
.astryx-checkbox-input :has(> input[type='checkbox']),
.astryx-checkbox-input input[type='checkbox'] { width: …; height: …; }
.astryx-checkbox-indicator > svg { width: …; height: …; }
```

## Change

Four targets, all reflecting `size`, purely additive, no visual change:

| target | element |
|---|---|
| `astryx-checkbox-input-slot` | the control's box slot |
| `astryx-checkbox-input-control` | the native `<input type="checkbox">` |
| `astryx-checkbox-indicator-check` | the checkmark `<svg>` |
| `astryx-checkbox-indicator-dash` | the indeterminate mark |

The mark names mirror `astryx-radio-indicator-dot` rather than inventing a parallel scheme. They live on `CheckboxIndicator`, so `DropdownMenuCheckboxItem` picks them up too.

`size` is on all four because Astryx sizes each of them from a per-size atomic, and `defineTheme` emits one compound selector on one element — a theme cannot write `.astryx-checkbox-input[data-size='md'] .astryx-checkbox-input-slot`.

## One thing worth knowing if you write a similar patch

My first attempt folded the slot's existing `{...focusProps}` spread into the same `mergeProps(…)` call, which looked like tidying. It broke the focus ring — three `focus ring ownership (WCAG 2.4.7)` assertions went red with `expected '' to be 'var(--focus-outline-style)'`. `focusProps` is deliberately its own trailing spread, so it is restored exactly as it was and only `themeProps` is merged with `stylex.props`. Good tests; they caught it immediately.

## Test plan

- `pnpm exec vitest run packages/core/src/theme/themingTargets.test.ts packages/core/src/CheckboxInput packages/core/src/Indicator packages/core/src/CheckboxList packages/core/src/DropdownMenu` — 541 passed, 8 files.
- Negative control: renaming `astryx-checkbox-input-control` in the doc fails the guard on both `docs` and `docsZh`; restoring it passes. Confirms the new targets are actually enforced, not merely written down.
- `pnpm build` — clean; `pnpm -F @astryxdesign/core typecheck` after it — clean.
- `pnpm lint:strict` — 0 errors (55 pre-existing warnings, none in the touched files).
- `pnpm test` — 11738 passed, 1 failed: `scripts/score-ledger.test.mjs` ("no response within 250ms"). **Pre-existing on `origin/main`** — reproduced twice in a clean worktree at `cbb991e91`.

No screenshots: the rendered output is byte-identical apart from the added class and `data-size` attributes, which nothing in this repo's CSS matches.

### DIFF
--- a/.changeset/checkbox-theme-targets.md
+++ b/.changeset/checkbox-theme-targets.md
@@ -2,5 +2,6 @@
 '@astryxdesign/core': patch
 ---
 
-[feat] CheckboxInput and the checkbox indicator now expose stable theme targets for their inner elements — `astryx-checkbox-input-slot` (the control's box slot) and `astryx-checkbox-input-control` (the native `<input>`), plus `astryx-checkbox-indicator-check` and `astryx-checkbox-indicator-dash` for the two marks, mirroring the existing `astryx-radio-indicator-dot`. All four reflect `size`. The slot and the native input were the only elements in CheckboxInput's tree carrying no `themeProps`, so a theme resizing the control had to reach them with `input[type='checkbox']` and `:has(> input[type='checkbox'])`. Purely additive; no visual change.
+[feat] The checkbox indicator now exposes stable theme targets for its two marks — `astryx-checkbox-indicator-check` (the checkmark) and `astryx-checkbox-indicator-dash` (the indeterminate bar) — mirroring the existing `astryx-radio-indicator-dot`. Both reflect `size`. A theme restyling the mark itself (stroke weight, colour, the dash's proportions) previously had to reach it with `.astryx-checkbox-indicator > svg` and `> span`, which are element-and-order selectors that break silently on any restructure. Purely additive; no visual change.
+
 @freddymeta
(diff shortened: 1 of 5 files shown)


cixzhang (approved):
Thanks — this is clean, and mirroring `radio-indicator-dot` is right. The title and body still describe the CheckboxInput slot and native input; neither is in the diff.

[Reviewed by Robohands]

PR #5428, opened 2026-08-24
feat(Toast): renderContent — custom toast layout, with a DismissButton that cannot be lost

freddymeta:
## Summary

Adds **`renderContent`** to the `showToast` options. It replaces the content of that toast's card with your own layout, and hands the renderer the pieces to arrange — the message, the `endContent`, and **Astryx's own dismiss `Button`**.

```tsx
showToast({
  body: 'Your changes have been saved.',
  renderContent: ({body, endContent, DismissButton}) => (
    <MyRow>
      <MyTitle>{body}</MyTitle>
      {endContent}
      <DismissButton />
    </MyRow>
  ),
});
```

## The problem

A product with its own notification layout has one option today, and it is CSS:

```css
.astryx-toast .astryx-button { display: none !important }
```

That is what an internal design-system adapter ships today in order to draw its own card into `body`. Three things are wrong with it:

1. **It reaches every toast under the theme, not just yours.** A shared library calling `useToast()` gets Astryx's card with no dismiss button. At `type="error"`, which doesn't auto-hide, that toast can never be dismissed at all.
2. **It also hits any `Button` in `endContent`**, so the POC carries a second rule to re-show its own controls. That rule keys on a class the component stopped rendering three commits later — dead ever since, and nothing noticed.
3. `Toast`'s trailing slot welds `endContent` and the close together, so a design wanting the close beside the *title* and the action on a second row can't express it.

## Why the close button is handed in

An earlier revision of this PR made the consumer build the dismiss control. All three isolated agents in the arbitration below independently flagged the same hazard: a layout that forgets the close, on a toast that doesn't auto-hide, is stuck on screen forever.

Passing the control instead removes the rebuild. `dismissButton` arrives as an element, so the close stays a real Astryx `Button`:

- **themeable** through `astryx-button`, like every other button in the system;
- **translated**, carrying `@astryx.toast.dismiss` rather than a hand-written string;
- **correctly named**, so a custom layout cannot ship an unlabelled close.

The layout is the consumer's. The control is not.

Keeping Astryx's card has the same character: `astryx-toast`, its per-type surface and its `MediaTheme` all stay, so a product restyles the card through the theme it already has and reaches for `renderContent` only when it needs a different *structure*.

## The close cannot be lost

A layout is free to leave the close out, and that has to stay allowed: [API Conventions](https://github.com/facebook/astryx/wiki/API-Conventions) says a component renders what it is given. But a toast with no close and no auto-hide traps the user, and an earlier revision of this PR only *warned* about it — which fires in the developer's console while the user is still stuck in production.

So the close is not a node the layout can drop. `renderContent` receives a **`DismissButton` component**, and rendering it claims the toast's dismiss slot:

```tsx
renderContent: ({body, DismissButton}) => (
  <MyRow>
    <MyTitle>{body}</MyTitle>
    <DismissButton />
  </MyRow>
)
```

**Nothing claims the slot → Astryx renders the close in the card's corner.** Forgetting it produces a toast with the close in the default position, not a toast with no way out. Absence is a default, not a hole: nothing suppresses anything, and both conventions hold.

The slot owns **mount and unmount registration**, rather than `Toast` recounting children on its own commits. Each `DismissButton` registers in its layout effect and unregisters in cleanup. That distinction matters when the button sits inside a nested component with its own state: the nested child can remove itself without rerendering `Toast`, cleanup decrements the slot immediately, and the corner fallback returns in that same commit. Rendering it is the claim, wherever it ends up.

The custom-content wrapper starts with the safe fallback. On the server and hydration's first render, `DismissButton` uses the server snapshot and renders nothing, so the fallback is the **one** close in the HTML. On the client, a hydration-safe `useSyncExternalStore` snapshot switches the leaf to the placed control while its layout effect registers the mount; React swaps fallback → placed before paint. The wrapper exists only for `renderContent`; an ordinary toast never mounts registration state at all — a `Profiler` guard records **one commit**, the same as main.

Rendering it twice renders two closes and warns in development. Visible, not dangerous; guidance over enforcement.

Escape-to-dismiss in the transport is worth doing for Dialog parity, but it is not the safety net here — it needs focus inside the viewport, so it rescues keyboard users and leaves pointer users trapped. Own PR, as agreed.

## Why per-toast, not per-viewport

This is where the previous revision was wrong. On the viewport, a renderer applied to **every** toast in the app — including one raised by library code, which was then rendered by a layout written for someone else's payload. Better than the CSS, which left that toast broken, but still not right.

Per-toast, a call that passes no `renderContent` renders as an ordinary Astryx toast: intact, dismissible, exactly as today. An app shares one layout by wrapping `useToast()` once and passing it on every call — the same one-line adoption, and it now says what it means: *these are our toasts*.

The story shows both side by side; its last button raises a toast without a renderer.

## Verified in Chromium

Driven headless against the built Storybook:

| | result |
|---|---|
| Default story | card present, one dismiss, named `Dismiss notification` — **unchanged** |
| Custom content | card kept; placed control is `.astryx-button`, named `Dismiss notification`, **exactly one of it** |
| **Layout that forgets the close** | **exactly one close renders anyway, and it dismisses** |
| **…and no flash** | **the close is present at the FIRST painted frame** |
| Custom + plain, together | striped one has the 4px stripe, plain one has none; **both carry exactly one correctly-named dismiss** |
| Per-type | info and error stripes resolve differently (white vs pink) through the card's inverted `MediaTheme` |
| `endContent` | `Undo` sits beside the close, inside the custom layout |
| Dismiss | closes the toast |
| Auto-hide | present at 3s, gone by 6s |
| Pause on hover | still there after 6.5s hovered |

Theming proved separately under the **shipped** layer order (`astryx-base` then `astryx-theme`, as `dist/astryx.css` emits), against the real `defineTheme` → `generateThemeCSS` output. A theme owns the whole card surface through `astryx-toast` — every one of `padding: 2px`, `border-radius: 3px`, `background: rgb(1,2,3)`, `box-shadow: none`, `width: 321px` lands, on a custom-content toast as much as a default one:

| | padding | radius | background | shadow | width | closes |
|---|---|---|---|---|---|---|
| default toast | 2px | 3px | rgb(1,2,3) | none | 321px | 1 |
| `renderContent` | 2px | 3px | rgb(1,2,3) | none | 321px | 1 |
| `renderContent`, close omitted | 2px | 3px | rgb(1,2,3) | none | 321px | **1** |

That is worth stating plainly, because it is the reason `renderContent` does **not** hand over the card: structure is this prop's job, surface is the theme's, and the two compose. A product wanting its own padding and background sets them on `astryx-toast` and keeps the live-region role, the auto-hide timer with its pause handlers, and the exit transition — none of which can be a consumer's to reimplement.

> Not a defect but worth knowing: in the **Storybook** build the StyleX plugin emits `@layer priority1…priority10` *after* `astryx-theme`, so theme component overrides lose there for regular properties. Shipped consumers are unaffected (`astryx.css` is one `@layer astryx-base`). It does mean Storybook is not a valid place to eyeball a theme override.

## Test plan

- All nine typecheck steps the `build-storybook` job runs, `pnpm build`, `pnpm check:repo`, `pnpm lint:strict` (0 errors) — clean. Toast + Layer + doc-guard suites: **166 tests green**.
- Eight tests on `renderContent` cover the handover: it replaces the layout for that toast; **a toast that did not ask for one is left alone, with its own working dismiss**; the dismiss arrives as Astryx's own correctly-named control; exactly one close renders; it dismisses; `endContent` is placed rather than dropped; auto-hide still runs and is reported to the renderer; the live-region announcement survives a custom layout. Five more cover the slot: a layout that forgets the close still renders exactly one and it dismisses; placing it adds no second one; a nested component removing it from its own state restores the fallback; a double render warns; and direct `Toast` renders to server HTML with exactly one translated dismiss control, then hydrates to exactly one. The SSR test fails at the server count (two) when the server/client split is removed. The nested test fails when registration cleanup is removed. A separate `Profiler` guard holds an unchanged default toast to one commit. No test asserts a computed style or a DOM shape.
- `pnpm a11y:audit -- --components Toast`: 10 stories, 0 violations, **no baseline entries added**.
- Docs (EN + zh), a `Custom content (renderContent)` story, and a changeset.
- Full `pnpm test` has one failure, not this PR's: `story-tree.test.ts > never nests a component deeper than <Package>/<Component>` on `Core/Icon/Size Theming`, which came in from main.

## Compatibility

Omit `renderContent` and nothing changes — same output, same timers, same DOM. New exported types: `ToastContentRenderProps`, `ToastContentRenderFn` (neither previously released). `LayerProvider` and `ToastViewport`'s rendered markup are untouched.

## On the name

It was `renderBody` until review. That name sat in the same options object as `body`, while every other render prop in core takes exactly the thing it names — `renderItem(item)`, `renderOption(option)`, `renderToken(item, onRemove)`. This one also owns `endContent` and the dismiss button, so the name understated its scope. `content` is already the system's word for the main region of a thing (`AppShell.content`, `HoverCard.content`, `Layout.content`) and collides with nothing here. Nothing was released under the old name, so there is no codemod.

## Note on the API shape

Ran the wiki's [API Arbitration](https://github.com/facebook/astryx/wiki/API-Arbitration) on the first revision: three candidates (a render prop; a `ToastRendererProvider` wrapper; a `hasChrome={false}` flag), six naive prompts, one isolated agent per option with only that option's skill doc.

- **`hasChrome={false}` lost outright** — its agent called the third-party case *"not merely hard but impossible inside the API"*, resorted to intercepting the module import to migrate an app, then spontaneously proposed a render prop. `endContent` also silently cancels against it.
- **The provider tied on generated code**, so the tie-break applied: fewer props, fewer concepts, matches existing patterns. Core already has 10 render props, and `Tokenizer`'s `renderToken(item, onRemove)` is this exact shape.
- **All three agents flagged the dismiss hazard**, and asked for a dev-mode diagnostic rather than a paragraph of docs. The slot goes further than they asked: the bad case ends with a working close rather than a console message.

One thing the arbitration surfaced that is still open, out of scope here: all three agents wanted `body` split into title + description, each inventing a private call-site↔renderer protocol instead. `renderContent` makes that a smaller question, since the layout is now the consumer's, but the payload is still one opaque node.

## Changed since review

- **The two theme targets are gone.** `astryx-toast-viewport` and `astryx-toast-item` sat on a positioning container and a grid row, neither of which paints — [Theming Infrastructure](https://github.com/facebook/astryx/wiki/Theming-Infrastructure) principle 1. `ToastViewport`'s rendered markup is now byte-identical to `main`'s, and the card's `astryx-toast` plus the close's `astryx-button` are the seams.
- **The close became `DismissButton`, and an unplaced one falls back to the card's corner** — see above. Its own mount/unmount owns presence now, so a deeply nested child-only toggle cannot leave stale registration behind. The dev-warning revision is gone with the `data-astryx-toast-dismiss` marker, the effect and the ref it needed.
- **The landmark fix left, as #5494.** It was only ever here because an older revision's story mounted its own viewport.
- **`useToastTimer` was reverted, and this one is worth a second opinion.** You called the extraction out as something you wanted. It existed because the renderer used to live on the *viewport*, above the card that owned the timer, so rendering custom content bypassed auto-hide entirely — my own test caught that. With `renderContent` inside `Toast`, the card owns both again and there is nothing to bypass, so the hook became a one-caller indirection and went back. If you would rather keep the seam anyway, say so and I will restore it.
- The internal POC URL is out of this description.

@freddymeta


cixzhang (changes requested):
Thanks — the arbitration write-up and the timer bug your own test caught are why this reads as settled work. **The render prop is the right shape and we want it**, including the timer moving out of the card into `useToastTimer`.

Two things to fix.

**The two theme targets should come out.** Neither element paints: `astryx-toast-viewport` sets `border: none`, `background: none`, `backgroundColor: transparent` and is otherwise position and `pointerEvents`; `astryx-toast-item` is `display: grid`, `gridTemplateRows`, `paddingBlockEnd` and the transition. Verified in the live DOM at this head — `-item` is fully transparent. [Theming Infrastructure](https://github.com/facebook/astryx/wiki/Theming-Infrastructure) principle 1: a target sits on the element that actually paints the thing you're theming, never a wrapper that exists only for layout or positioning. Principle 5 names layout wrappers as implementation detail. `astryx-toast` — the card — is the seam, and it already ships.

Both underlying needs are answered elsewhere: where the stack sits is already the `position` prop, and the gap below the last toast was a real bug, fixed on main in [#5460](https://github.com/facebook/astryx/pull/5460).

**And a renderer that omits a close leaves an error toast with no exit.** Escape does nothing, F6 lands outside the viewport because nothing inside it is focusable, and an error toast doesn't auto-hide — still there after ten seconds.

| your renderer, no close | Astryx's card, same toast |
|---|---|
|  |  |

Your own agents asked for a dev-mode error rather than a paragraph, and I'd take that here. Escape in the transport is the durable fix and I agree it's its own PR.

The landmark fix is a behaviour change to shipped Toast under a `renderToast` title — send it on its own and it lands today.

[Reviewed by Robohands]

packages/core/src/Toast/ToastViewport.tsx
```diff
@@ -456,15 +535,31 @@ export function ToastViewport({
...
-                <Toast
-                  type={type}
-                  body={o.body}
-                  endContent={o.endContent}
-                  isAutoHide={isAutoHide}
-                  autoHideDuration={dur}
-                  isExiting={isExiting}
-                  onDismiss={reason => removeToast(entry.id, reason)}
-                />
+                {renderToast ? (
```
cixzhang: Nothing makes the renderer draw a close. An error toast without one never leaves.

cixzhang (changes requested):
Thanks — my earlier review asked for the non-painting targets to come out, the landmark change to split, and persistent toasts to keep an exit. The targets are gone, the landmark is now [#5494](https://github.com/facebook/astryx/pull/5494), and omission now falls back correctly.

Two runtime blockers remain.

`DismissButton` only reports presence when `Toast` commits. A nested custom layout can remove it from its own state without rerendering `Toast`; Chromium went from one close to zero on a persistent error toast, and F6 stayed outside. The same bookkeeping takes an unchanged default toast from one commit to three; an A/B guard returned it to one.

Could the slot own mount/unmount registration, stay inactive without `renderContent`, and get a focused child-only toggle test?


[Reviewed by Robohands]

rubyycheung (approved):
Looks good from design. The customization remains scoped to individual Toasts, preserves the standard surface and behavior, and safely retains a dismiss control when a custom layout omits it.

cixzhang (changes requested):
Thanks — my last review said “DismissButton only reports presence when Toast commits”; this head fixes that, and rubyycheung's design approval settles the treatment.

One blocker remains. The exported `Toast` is documented for inline/static use, but server-rendering it with `<DismissButton />` emits both that control and the corner fallback: two “Dismiss notification” buttons before hydration, one after. Could this keep exactly one across SSR and hydration while preserving the child-unmount fallback?

One nit in the same pass: the opening example still uses `dismissButton`, and `ToastProps` points to the old viewport API.

[Reviewed by Robohands]

cixzhang (approved):
Thanks — my last review asked for exactly one dismiss control through SSR and hydration. This head does that, preserves nested fallback, and fixes the stale docs. Looks good.

[Reviewed by Robohands]

PR #5431, opened 2026-08-24
fix(docsite): bound the hero's pin so overscroll works everywhere

### LINKED ISSUE #5392: [Bug] Docsite: pull-to-refresh is disabled on mobile by a global overscroll-behavior-y rule

imdreamrunner:
Fixes #5392.

`apps/docsite/src/app/globals.css` turned overscroll off on the root element for the whole app:

```css
html {
  overscroll-behavior-y: none;
}
```

That is also the standard way to turn **pull-to-refresh** off, and the rule is unscoped — so pull-to-refresh was dead on every route of the docsite, on every touch browser. It was added in #3032 to stop the home hero's `position: fixed` layers showing through the top/bottom overscroll gap, a bug reported on desktop; the mobile cost looks unintentional (it appears in that PR's title, summary, test plan or review nowhere).

The issue split this into a small mobile step and a hard desktop step, on the reading that the desktop hero's layers must stay `fixed` because their fixedness *is* the pin-and-cover. **They don't have to.** One mechanism covers both widths, so this removes the rule outright instead of scoping it.

## The insight

`fixed` is an *unbounded* pin: the layer stays glued to the viewport for the entire document scroll, so at the end of the page it is still painting into the strip a bottom overscroll opens. `sticky` pins to the same place but is bounded by its container — it stops existing on screen once its container has gone by, and the container here is the hero band, which the showcase has covered by then. So it looks the same and cannot bleed.

The hero already had exactly this: `cardsLayer` is a zero-height `position: sticky` layer. The glow and the cards stage just bypassed it by being `fixed`.

**Zero height is the load-bearing part.** A sticky box only stays pinned for the slack between its own height and its container's, so a zero-height layer gets the whole band to travel while the visual inside it is placed absolutely. An in-flow 712px hero inside a 760px band un-pins after 48px — I tried it, and it doesn't work.

## What changed

| layer | was | now |
|---|---|---|
| `backdropGlow` (`HeroThemeReel`) | `fixed`, all widths | `absolute` inside the pin layer |
| `stage` (`HeroFloatingCards`) | `fixed` | `absolute` inside the pin layer |
| `heroContent` (`(site)/page.tsx`) | `relative` / `fixed` ≥1024 | `relative` / `absolute` ≥1024, inside a new zero-height sticky `heroPin` |
| `navBackdrop` | `fixed` | unchanged — a header-height strip at the very top, never in the bottom gap |
| `globals.css` | `overscroll-behavior-y: none` | **deleted** |

`cardsLayer` is renamed `pinLayer` since it now carries the glow as well as the cards. `heroPin` must precede `heroSpacer` in the DOM: a sticky layer pins from its natural flow position, so after the spacer it would only start pinning once the whole band had already scrolled past.

## Test plan

`pnpm lint:strict` (0 errors), `pnpm build`, `apps/docsite` `vitest run` (406 passed) and `tsc --noEmit` all green, rebased on current `main`. The two `packages/cli` failures in the root `node` project (`component.test.mjs`, `hook-discovery.test.mjs`, both about component-name case sensitivity) reproduce identically on a clean `8a055dfb` worktree — pre-existing, unrelated.

**New test** — `apps/docsite/src/__tests__/mobile-overscroll.test.ts`, written as invariants rather than as a mirror of the code:

- `globals.css` may not disable overscroll, at any width. It parses the file and reports any `overscroll-behavior*` declaration that isn't `auto`, naming its enclosing blocks.
- none of the three hero layers may be `position: fixed`, in any media arm.

All four fail on the pre-fix sources and pass after.

**Measured** by running this branch and a pristine `main` worktree side by side (`next dev`, headless Chromium, `reducedMotion: 'reduce'` so the 4.5s theme reel is frozen and the arms show the same slide), pixel-diffing the home page with `pixelmatch` at ten scroll offsets each — 0, 200, 400, 600, 712, 760, 800, 1000, 1600, 2400.

| viewport | pixels differing from `main` | `overscroll-behavior-y` | fixed layers reaching the viewport bottom at page end |
|---|---|---|---|
| desktop 1280×900 | **0** at all ten offsets | `none` → `auto` | glow + stage → **none** |
| short desktop 1280×700 | **0** | `none` → `auto` | glow + stage + hero text → **none** |
| tablet 900×1200 | **0** | `none` → `auto` | none → none |
| phone 390×844 | **0** | `none` → `auto` | glow → **none** |

So the pin-and-cover is byte-identical to what ships today, at every width, and nothing is left that can paint into the overscroll gap. The short-desktop row is why `heroContent` is in scope: below ~760px of viewport height it reaches the bottom edge too, so a mobile-only fix would have left a hole on any non-maximized laptop window.

Two notes for anyone reproducing locally:

- The dev build shows a **"You're viewing unreleased docs (Canary)" banner** that is not on the deployed site and inflates the mobile header from 48px to 232px. It is removed before measuring; otherwise the hero geometry looks 184px off.
- The top overscroll gap paints the white canvas (`html`/`body` are both transparent), which is the same "clean" state the issue already measured under the footer — no new exposure there.

### DIFF
--- a/apps/docsite/src/__tests__/mobile-overscroll.test.ts
+++ b/apps/docsite/src/__tests__/mobile-overscroll.test.ts
@@ -2,8 +2,8 @@
 
 /**
  * @file mobile-overscroll.test.ts
- * @input apps/docsite/src/app/globals.css and the home hero's HeroThemeReel
- * @output Guards that the docsite never suppresses pull-to-refresh on touch
+ * @input apps/docsite/src/app/globals.css and the home hero's pinned layers
+ * @output Guards that the docsite never suppresses the platform's overscroll
  * @position Regression test for #5392
  *
  * `overscroll-behavior-y: none` on the root element is how you turn
@@ -12,10 +12,11 @@
  * (#3032), and silently cost pull-to-refresh on every route of the site on
  * every touch browser.
  *
- * These two checks are the halves of the fix, stated as invariants rather than
- * as a mirror of the code: the root may not refuse overscroll at a width a
- * phone can actually have, and the one hero layer that reached the gap there
- * has to stay bounded to the hero instead of pinned to the viewport.
+ * These checks are the halves of the fix, stated as invariants rather than as
+ * a mirror of the code: the root may not refuse overscroll at any width, and
+ * no hero layer may be pinned to the viewport — a viewport-pinned layer is
+ * exactly what paints into the strip an overscroll opens past the end of the
+ * page, and is the only reason the root rule was ever wanted.
  */
 
 import {describe, it, expect} from 'vitest';
@@ -25,9 +26,6 @@ import {fileURLToPath} from 'node:url';
 
 const HERE = path.dirname(fileURLToPath(import.meta.url));
 const APP = path.join(HERE, '..', 'app');
-
-/** A phone-class viewport, comfortably under every breakpoint the site uses. */
-const PHONE_WIDTH_PX = 430;
 
 function read(...parts: string[]): string {
   return fs.readFileSync(path.join(APP, ...parts), 'utf8');
@@ -62,24 +60,8 @@ function openBlocksAt(css: string, offset: number): string[] {
   return stack;
 }
 
-/**
- * Whether a media prelude matches a viewport `width` px wide. Only the width
- * features are modelled; anything else is treated as matching, so an unhandled
- * query fails the test loudly rather than passing it by omission.
- */
-function mediaMatchesWidth(prelude: string, width: number): boolean {
-  for (const [, feature, value] of prelude.matchAll(
-    /\((min|max)-width:\s*([\d.]+)px\)/g,
-  )) {
-    const px = Number(value);
-    if (feature === 'min' && width < px) {return false;}
-    if (feature === 'max' && width > px) {return false;}
-  }
-  return true;
-}
-
 describe('docsite globals.css', () => {
-  it('never disables overscroll at a phone width (pull-to-refresh, #5392)', () => {
+  it('never disables overscroll — that is pull-to-refresh (#5392)', () => {
     const css = stripComments(read('globals.css'));
     const offenders: string[] = [];
 
@@ -91,13 +73,9 @@ describe('docsite globals.css', () => {
         continue;
       }
       const blocks = openBlocksAt(css, match.index);
-      const selector = blocks.at(-1) ?? '(top level)';
-      const reachesPhone = blocks
-        .filter(b => b.startsWith('@media'))
-        .every(b => mediaMatchesWidth(b, PHONE_WIDTH_PX));
-      if (reachesPhone) {
-        offenders.push(`${selector} { overscroll-behavior: ${value} }`);
-      }
+      offenders.push(
+        `${blocks.join(' > ') || '(top level)'} { overscroll-behavior: ${value} }`,
+      );
     }
 
     expect(offenders).toEqual([]);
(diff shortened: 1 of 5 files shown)


rubyycheung (changes requested):
The gradient and UI cards need to remain one visual composition and move together, matching production. This PR does not preserve that behavior on narrow screens.

Below 1024px, the visible cards come from `HeroReelStack` inside `HeroContent` in normal flow, while `backdropGlow` remains in the separate sticky `pinLayer`; the desktop `HeroFloatingCards` stage in that layer is hidden. As a result, the mobile gradient and the visible cards use different positioning contexts and move independently.

Please place the mobile gradient with the mobile collage—or otherwise give both the same positioning behavior—so the gradient stays attached to the UI elements while scrolling. The bounded desktop sticky approach can remain separate.

kentonquatman (approved):

kentonquatman (changes requested):
I approved, but just saw the comment from @rubyycheung. We'll need to fix that first.

PR #5438, opened 2026-08-24
fix(cli): an unknown manifest key no longer discards the whole integration

josephfarina:
## What

A key in `astryx.integration.*` that this CLI does not know is now **ignored with a warning** instead of failing the parse.

```js
// astryx.integration.mjs, authored against a newer CLI
export default {
  components: './src',
  templates: './blocks',
  someFutureRoot: './future', // ← this CLI has never heard of it
};
```

| | Before | After |
| --- | --- | --- |
| The manifest | rejected | loaded |
| `components`, `templates`, `codemods` | **all lost** | all contributed |
| Diagnostic | `invalid_manifest` error | `unknown_manifest_key` warning naming the keys |
| `components: 42` | error | error (unchanged) |

## Why

The manifest was parsed with `.strict()`, so one unrecognized field failed the parse — and an integration whose manifest fails to parse contributes *nothing*. Not just the unknown field: its components, templates and codemods went with it.

That is the wrong default for this particular file. An integration is **published once and installed against many CLI versions**, so a field added by a newer CLI reaches older consumers as a matter of course. It means only that this CLI cannot act on that field. Treating it as corruption costs the consumer the entire package.

This has already happened once, and the CHANGELOG for 0.4.x describes it:

> Meta's internal `@nest/xds-meta` sat invisible to CLI discovery for a week that way, and the app team's conclusion was that the components did not exist (#5119).

0.4.x fixed the *visibility* half — the load error is now reported rather than silent. This fixes the cause: the load should not have failed at all.

It is also, concretely, what stops the next occurrence. `docs` shipped in 0.4.6 (#5311). An integration that adopts it is unusable on every consumer resolving ≤ 0.4.5 — which is most of them, since consumers pin the CLI and upgrade on their own schedule. The same will be true of the field after that, and the one after that, unless the default changes.

## How

- **`.strict()` → default strip.** Unknown keys are dropped from the parsed value rather than rejected. A *known* key of the wrong type still fails: `components: 42` is an authoring mistake, not version skew.
- **`unknownIntegrationKeys(raw)`** reads the raw default export before parsing strips it. It is separate from `parseIntegration` because the two answer different questions — "is this manifest usable" versus "how much of it can this CLI use".
- **`loadManifest`** returns `{manifest, unknownKeys}`; `loadManifestObject` stays as the value-only form. The keys ride on the loaded integration as `__unknownKeys`, beside the existing `__loadError`.
- **One warning, both surfaces.** `checkUnknownKeys` lives in the shared `validate-contributions`, so `Project.issues()` and `astryx validate-integration` report it identically — and because it is `severity: 'warning'`, `Project`'s skip-on-error policy does not withhold the package's contributions, which is the entire point.

The message names the keys and says what to do:

```
This CLI does not know the manifest key "someFutureRoot", so it is ignored —
everything else in the manifest still applies. This usually means the
integration was published against a newer Astryx CLI than this project
resolves; upgrading @astryxdesign/cli picks it up.
```

## Compatibility

No API signature changes. `parseIntegration` keeps its shape, and a manifest with no unknown keys parses to exactly what it did before. The one behavior change is the intended one, and it is strictly more permissive — nothing that loaded before fails now.

`--json` gains no new field; the warning travels in the existing `issues` array that already carries `missing_root`, `invalid_template`, and the rest.

## Test plan

- `authoring/integration/parse.test.mjs` — an unknown key is stripped and reported; a known key of the wrong type still throws.
- `api/integration/validate-integration.test.mjs` — the unknown key is a `warning`, not `invalid_manifest`, **and the declared root is still validated**, which is the proof the rest of the manifest was read rather than discarded.
- `foundation/config/project.test.mjs` — the load-bearing one: with an unknown key present, `components()`, `templates()` and `codemods()` all still return the integration's contributions, and no issue has `severity: 'error'`.
- End-to-end through the binary, against a scaffolded consumer whose integration manifest carries a future key: `astryx --json component --list --package @acme/w` lists the integration's component with this branch, and omits the package entirely on `main`.
- `pnpm build`, `pnpm test` (569 files), `pnpm lint:strict`, `check:cli-structure`, `typecheck:authoring`, `typecheck:strict`, `readme:check` — all clean.

No visual change — this PR touches only `packages/cli`.

### DIFF
--- a/.changeset/cli-manifest-forward-compat.md
+++ b/.changeset/cli-manifest-forward-compat.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/cli': patch
+---
+
+[fix] A manifest key this CLI does not know no longer discards the whole integration. `astryx.integration.*` was parsed with a strict schema, so one unrecognized field failed the parse — and an integration whose manifest fails to parse contributes _nothing_, taking its components, templates and codemods down with it. Since an integration is published once and installed against many CLI versions, a field added by a newer CLI reached every older consumer as total, silent loss of that package (#5119). Unknown fields are now ignored with an `unknown_manifest_key` warning naming them, and the rest of the manifest still applies; a _known_ field of the wrong type is still an error. (#5311 follow-up)
+
+@josephfarina
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -597,9 +597,11 @@ from `@astryxdesign/cli/authoring` for editor autocomplete and type-checking.
 
 Every command loads the consumer's `astryx.config`, resolves each listed
 integration's manifest from `node_modules`, and discovers its contributions.
-Everything is validated against one strict schema at the load boundary, so the
-CLI presents core and integration contributions through a single, uniform
-surface.
+Everything is parsed at the load boundary, so the CLI presents core and
+integration contributions through a single, uniform surface. A field of the
+wrong type fails there; a field this CLI does not know is ignored with a
+warning naming it, so a manifest written against a newer CLI still contributes
+everything this one understands.
 
 Discovery is resilient: a broken or misconfigured integration is skipped with a
 one-line warning on stderr instead of crashing the CLI, and it never corrupts a
--- a/packages/cli/api/integration/validate-integration.mjs
+++ b/packages/cli/api/integration/validate-integration.mjs
@@ -35,7 +35,7 @@ import * as path from 'node:path';
 import {assertWithin} from '../../foundation/fs/path-safety.mjs';
 import {
   findManifestPaths,
-  loadManifestObject,
+  loadManifest,
   resolvePackageDir,
 } from '../../foundation/integrations/integrations.mjs';
 // The on-disk contribution validators live in foundation: Project and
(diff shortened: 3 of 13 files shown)


cixzhang (dismissed):
Thanks — preserving known contributions when an older CLI sees a newer manifest key is the right compatibility model. Before merge, could we keep the helper/key census out of the public `./integration` surface and make it unable to drift from the schema, then update the integration guide and generated README's strict-schema claim?

[Full review](https://github.com/cixzhang/astryx/wiki/Review-5438)

[Reviewed by Robohands]

josephfarina:
Thanks — both findings addressed in `4e29ae1`.

**1. The census is derived, and internal.** `KNOWN_INTEGRATION_KEYS` is now `Object.keys(integrationSchema.shape)`, so a field added to the schema cannot be reported as unsupported — there is only one list. The schema and the census moved to `packages/cli/authoring/integration/schema.mjs`, which no `exports` entry resolves to, so neither name is reachable at runtime any more. `parse.mjs` — the module `./integration` maps to — exports `parseIntegration` and nothing else.

**2. The strict-schema claim is fixed** in `assets/docs/cli-integrations.doc.mjs` and in the README's "How it works" prose. That README paragraph sits outside the `BEGIN GENERATED` markers, which is why `readme:check` stayed green over a stale sentence — it is hand-written, so I edited it directly. `readme:check` still passes.

Added one test for the drift you were worried about: every key in the schema shape must come back known, which a hand-maintained list would eventually fail.

**Verification**
- Focused suites for the touched paths — `parse`, `validate-integration`, `integrations`, `integration-warnings`, `project`: 60/60 pass.
- Full CLI suite: 7 failures on this branch, 8 on untouched `main` at the same moment, in the same three files (`component.test.mjs`, `cliManifest.test.ts`, `build-theme.variants.test.mjs`). Pre-existing and environment-dependent; each passes when its file is run alone.
- `check:repo`, `check:cli-structure`, `readme:check`, `typecheck:authoring`, strict eslint and prettier: all clean.

No behavior change from the first pass — same warning, same errors, same parsed values.

PR #5439, opened 2026-08-24
feat(DateInput): the platform's own date picker on touch, with `nativePicker`

imdreamrunner:
## What

A touch device gets the platform's own date picker — the iOS wheel, the Android calendar dialog — drawn by the OS, with its hit areas, momentum scrolling, locale and accessibility settings. `nativePicker` chooses the surface:

| Value | Behaviour |
| --- | --- |
| `'touch'` (default) | the OS picker on a coarse pointer, the calendar popover on a mouse |
| `'always'` | the OS picker wherever the browser supports `<input type="date">` |
| `'never'` | Astryx's own everywhere — #5243's bottom-sheet picker on a finger, the calendar popover on a mouse |

```tsx
// Astryx's own touch picker instead of the platform's
<DateInput label="Event date" value={date} onChange={setDate} nativePicker="never" />
```

Reach for `'never'` on a field that needs what a native picker cannot express: `weekStartsOn`, `numberOfMonths`, or `dateConstraints`.

Not breaking: `DateInputProps` gains an optional prop and nothing is removed. A field that wants the bottom-sheet picker asks for it.

## `format` and `placeholder` still apply

DateInput paints the closed field's text itself rather than letting the engine do it, so a date reads the same on a phone as on a desktop. The control's own text is made transparent and ours is drawn over it, pointer-transparent so the tap still raises the picker. The value the control holds stays ISO — all it accepts, and what the picker reads and writes.

## Four things measured on a real iPhone

None of these reproduce in a desktop browser, and each is pinned by a test that fails without its fix.

- **The element is uncontrolled, and React never rewrites it after mount.** While the iOS picker sheet is open, *any* programmatic write to the field detaches the sheet from it: the wheel keeps moving its own highlight, but nothing reaches the input and no event fires, so a pick appears to do nothing. External changes go in through an effect that only writes while the field is unfocused.
- **Edits come back through native listeners.** React's synthetic change does not reliably observe this control — measured: a native `input` event carried the new date while React's `onChange` never ran, so React re-rendered and wrote its stale value straight back.
- **The control is pinned to one line box** (`1lh`, with a `calc()` fallback) with `appearance: none`. Its intrinsic height comes from its inner edit fields rather than `line-height`, so it otherwise renders ~2px taller than the text field and its value sits off the shared baseline.
- **The text overlay is a block box with both inline insets.** `text-overflow` only applies to a block container, so on a flex one a long date hard-clips mid-glyph instead of ellipsising; without the end inset the overlay is shrink-to-fit and paints ~24px over the clear button.

## `min` / `max`

Forwarded to the control, but a native picker may not *show* them: on iOS they are constraint-validation flags rather than clamps, so the sheet lets the user land on an out-of-range date. It is refused on commit and announced to assistive technology, and the refusal clears when the picker closes and the field reverts. Verified on a device — no bad data escapes, but the picker will not grey those dates out. A field where that matters should use `'never'`.

## Test plan

- New `NativeDateField.test.tsx` — 30 tests: surface selection for all three values, the ISO round-trip, the synthetic-change bypass, the no-write-while-focused invariant, `min`/`max` forwarding and refusal, `dateConstraints`, the format overlay (named formats, function format, provider locale, live during the picker), the placeholder, the paint-level rules, `showPicker` including an engine that throws, clear, disabled, labelling and ref forwarding.
- `DateInputTouch.test.tsx` passes `nativePicker="never"` — that suite is about Astryx's own touch surface, which is now opt-out — plus one new test asserting the default. 135 pass.
- `DateInput.test.tsx` (the pointer surface) passes untouched: 95.
- `pnpm lint:strict` 0 errors, `pnpm build` clean, and the full UI suite green at **7996**.

## Supersedes #5261

Same feature, rebuilt against the surface split that landed in #5243. That PR restructured `DateInput.tsx` into `PointerDateField` + `TouchDateField`, which is a better home for this than the single branching component #5261 modified. The research behind it is in P2470053021.

### DIFF
--- a/.changeset/date-input-native-picker-option.md
+++ b/.changeset/date-input-native-picker-option.md
@@ -1,0 +1,27 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] DateInput: the platform's own date picker on touch, with
+`nativePicker` to choose otherwise
+
+A touch device now gets the browser/OS picker — the iOS wheel, the Android
+calendar dialog — drawn by the platform, with its own hit areas, momentum
+scrolling, locale and accessibility settings. `nativePicker` picks the
+surface: `'touch'` (the default) uses it on a coarse pointer, `'always'`
+wherever the browser supports `<input type="date">`, and `'never'` keeps
+Astryx's own pickers everywhere — the bottom-sheet picker on a finger, the
+calendar popover on a mouse.
+
+Reach for `'never'` on a field that needs what a native picker cannot
+express: `weekStartsOn`, `numberOfMonths`, or `dateConstraints`.
+
+`format` and `placeholder` still apply in native mode — DateInput paints the
+closed field's text itself, over the control, so a date reads the same on a
+phone as on a desktop. `min` and `max` are forwarded, but a native picker may
+not *show* them: on iOS they are constraint-validation flags rather than
+clamps, so an out-of-range date can be selected, and it is refused on commit
+and announced rather than greyed out in the picker. Mouse-driven devices are
+unaffected.
+
+@imdreamrunner
--- a/packages/core/src/DateInput/DateInput.doc.mjs
+++ b/packages/core/src/DateInput/DateInput.doc.mjs
@@ -157,6 +157,13 @@ export const docs = {
       default: "'date_long'",
     },
     {
+      name: 'nativePicker',
+      type: "'touch' | 'always' | 'never'",
+      description:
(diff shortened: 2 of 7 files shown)


imdreamrunner:
Closing as a duplicate — the work lives on #5261 instead, which is where the discussion and review already are. Same commit, same branch as before, rebuilt against #5243's surface split.

Opening this as a separate PR was my mistake: the ask was to rebase 5261, and I should have moved the work onto that branch rather than starting a new one.

PR #5440, opened 2026-08-24
fix(core): defer clear focus restoration out of tap to prevent touch page scroll jump

### LINKED ISSUE #5422: InputClearButton focuses inside the tap for every input — #5350 fixed only TouchDateField
[#5350](https://github.com/facebook/astryx/pull/5350) fixed a touch clear that jumped the page to the top, by deferring focus out of the tap (`packages/core/src/DateInput/TouchDateField.tsx:742`). The fix lives in `TouchDateField` only.

The shared clear affordance — `packages/core/src/Field/InputClearButton.tsx` — focuses inside the tap for every input that uses it, and has no equivalent handling. Current consumers: `DateInput`, `DateRangeInput`, `DateTimeInput`, `PanelSearchInput`, `FileInput`.

No worse than it was before [#5350](https://github.com/facebook/astryx/pull/5350) — filing it because the merged fix does not claim these components, and the class finding otherwise has no home.

Geervan:
## Summary
Fixes #5422 
Following up on #5350 (which fixed `TouchDateField`), this PR extends the deferred focus restoration pattern across all clearable input controls (`DateInput`, `DateTimeInput`, `PanelSearchInput`, `FileInput`, `TextInput`, `NumberInput`, `TimeInput`) that consume `InputClearButton`.

## Cause

When clearing an input value via `InputClearButton` on mobile WebKit/iOS Safari:
1. Clearing the value synchronously unmounts `<InputClearButton>` (which is conditionally rendered when `value` is set).
2. Calling `.focus()` synchronously on the target input within the same task while processing the touch tap causes iOS Safari to jump the document scroll position to `0` (`scrollTop: 0`).

## The Fix

- **Focus Deferral**: Wrapped `.focus({preventScroll: true})` in `requestAnimationFrame` across all clearable input components (`DateInput`, `DateTimeInput`, `PanelSearchInput`, `FileInput`, `TextInput`, `NumberInput`, `TimeInput`).
- **Focus Protection**: Added `onPointerDown`/`onMouseDown` with `preventDefault()` on `InputClearButton` to prevent pointer events from stealing focus from the input field prior to clearing.

## Test Plan

- **Vitest Unit Tests**: All tests passing (`300/300` tests passed across `InputClearButton`, `DateInput`, `DateTimeInput`, `FileInput`, `TextInput`, `NumberInput`, `TimeInput`).
- **ESLint**: Passed with `0` errors.
- **Changeset**: Added `.changeset/defer-clear-focus-tap.md` (`node scripts/check-changesets.mjs` passed).
- **Visual Verification**: Verified in Storybook with mobile touch emulation — tapping clear clears the value without page scroll displacement.


cixzhang (changes requested):
Touch clearing keeps TextInput focused and scroll-stable in Chromium, but keyboard users now lose focus to the document for one frame before the input is restored. The base kept focus on the input before click bubbling and the first animation frame. Please preserve that keyboard path while deferring only the pointer/touch path, with focused regression coverage.

This head also no longer integrates with current main: the synthetic merge conflicts in DateTimeInput and TimeInput and produces duplicate `onPointerDown` props in InputClearButton, colliding with the current popup-aware handler. Please rebase and compose the current handler so touch clearing stays scroll-stable without dropping current popup behavior.

[Reviewed by Robohands]

cixzhang (approved):
Thanks, this fixes both prior issues. Chromium keeps keyboard focus and touch scroll stable, and Selector/MultiSelector popups stay open when cleared.

[Reviewed by Robohands]

PR #5441, opened 2026-08-24
fix(cli): handle multi-dot filenames, m4v/mkv video formats, and avoid silent image fall-through in template asset stripping

### LINKED ISSUE #5423: CLI: an unrecognised template-asset extension silently falls through to the image placeholder
`packages/cli/foundation/discovery/template-adapter.mjs` decides whether a scaffolded asset is a video by reading its extension. An extension it does not recognise falls through to the image placeholder **silently**, so the failure points at "broken content" rather than erroring.

Two ways in:

- `DEMO_ASSET_PATTERN` (`:182`) is `/\/template-assets\/[\w-]+\.(\w+)/g` — it reads the extension off the **first** dot, so `clip.min.mp4` is judged on `min`
- `VIDEO_EXTENSIONS` (`:169`) has no `m4v`

**Measured on `main`: 150 `/template-assets/` references, none with a second dot, none `.m4v`** — neither path reaches anyone today. Filed for the silent fall-through itself, not for the two cases.

Found by @AKnassa in [#4839](https://github.com/facebook/astryx/pull/4839), which was closed as superseded by [#4863](https://github.com/facebook/astryx/pull/4863).

Geervan:
Fixes #5423

Fixes template demo asset reference stripping in `packages/cli/foundation/discovery/template-adapter.mjs`. 

Previously, `stripTemplateAssetRefs()` made two key assumptions:
1. `DEMO_ASSET_PATTERN` (`/\/template-assets\/[\w-]+\.(\w+)/g`) captured the extension from the **first** dot. For filenames with multiple dots (e.g. `/template-assets/clip.min.mp4`), it captured `min`, corrupting the asset URL to `data:image/svg+xml,...%3E.mp4`.
2. `VIDEO_EXTENSIONS` lacked Apple `.m4v` (and `.mkv`) video extensions.
3. Any extension not in `VIDEO_EXTENSIONS` silently fell through to `PLACEHOLDER_IMAGE` (an SVG image data URI), causing video/non-image assets to silently mis-render as broken images.

## The Fix

- **Multi-Dot Match**: Updated `DEMO_ASSET_PATTERN` to `/\/template-assets\/[\w.-]+\.([a-zA-Z0-9]+)/g` to capture the extension after the final dot.
- **Explicit Classification**: Defined `IMAGE_EXTENSIONS` (`png`, `jpg`, `jpeg`, `gif`, `svg`, `webp`, `avif`, `ico`) alongside `VIDEO_EXTENSIONS` (`mp4`, `webm`, `mov`, `ogv`, `m4v`, `mkv`).
- **Safe Fall-Through**: Non-image assets and unrecognized extensions now fall back to `''` instead of mis-rendering image SVG data URIs into media tags.

## Test Plan

- **Vitest Unit Tests**: Added unit tests to `packages/cli/api/template/template.test.mjs` (`12/12` passed).
- **Reproduction**: Verified via Node.js execution across all 3 reproduction scenarios (multi-dot filenames, `.m4v`, and non-image fall-through).
- **Changeset**: Added `.changeset/cli-asset-strip-fix.md` (`node scripts/check-changesets.mjs` passed).

### DIFF
--- a/.changeset/cli-asset-strip-fix.md
+++ b/.changeset/cli-asset-strip-fix.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/cli': patch
+---
+
+[fix] handle multi-dot filenames, m4v/mkv video formats, and avoid silent image fall-through when stripping template demo assets
+@Geervan
--- a/packages/cli/api/template/template.test.mjs
+++ b/packages/cli/api/template/template.test.mjs
@@ -61,12 +61,26 @@ describe('stripTemplateAssetRefs', () => {
     expect(out).not.toContain('/template-assets/');
   });
 
-  it('strips other video extensions (.webm, .mov, .ogv) the same way', () => {
-    for (const ext of ['webm', 'mov', 'ogv']) {
+  it('strips other video extensions (.webm, .mov, .ogv, .m4v, .mkv) the same way', () => {
+    for (const ext of ['webm', 'mov', 'ogv', 'm4v', 'mkv']) {
       const src = `src: '/template-assets/clip.${ext}',`;
       const out = stripTemplateAssetRefs(src);
       expect(out).toBe("src: '',");
     }
+  });
+
+  it('correctly matches multi-dot filenames up to the final extension', () => {
+    const src = "src: '/template-assets/clip.min.mp4',";
+    const out = stripTemplateAssetRefs(src);
+    expect(out).toBe("src: '',");
+    expect(out).not.toContain('data:image/svg+xml,');
+  });
+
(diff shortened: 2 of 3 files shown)


cixzhang:
Thanks for catching the final-suffix and media-kind gaps. We’re going with #6039 as the canonical fix: it handles the same multi-dot and `.m4v` cases, and unknown formats fail with the exact fixture path and format instead of being silently emptied. The repository has no `.mkv` fixture or reference, so that extra allowlist entry does not preserve a live template path. Closing this as superseded so we keep one classifier implementation.

[Reviewed by Robohands]

PR #5444, opened 2026-08-24
fix(markdown): a table column that declares :--- says so

lexs:
Split out of #5288, where @cixzhang pointed at the bug underneath it.

`cellAlignStyles` (`Markdown.tsx`) had entries for `center` and `end` only, so a column written `| :--- |` got no style at all and rendered exactly like an undeclared one. That looks right in isolation, because a cell already starts at the inline start — but **`text-align` inherits**, so the column follows whatever the table is nested in. Render a table inside a centered block and a first column that explicitly declares `:---` centers itself; nothing in the markdown says it should.

A declared alignment is now always written out, `start` included. Only the `:---` case changes, and only where something around the table had already moved the default — an undeclared column still takes it.

**One thing worth your ruling.** The GFM semantics behind these markers are physical: the spec's HTML output is `align="left"`, and GitHub renders a `:---` column on the left in an RTL document. `no-physical-properties` says the migration to logical is complete and gates at `error`, so this follows the house rule and keeps `:---` → `start` / `---: ` → `end`, which mirror-flips under RTL. That is a deliberate deviation from the source language, and it is also, I suspect, how the missing `start` came about in the first place — `end` is what the lint rule leaves you with, and `start` then looks redundant. If you would rather match GFM exactly, it is the same three lines with a scoped suppression.

Test asserts every declared column's computed alignment on both `th` and `td`, and that an undeclared column keeps the table's own default.

### DIFF
--- a/.changeset/markdown-table-explicit-left-align.md
+++ b/.changeset/markdown-table-explicit-left-align.md
@@ -1,0 +1,19 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Markdown: a table column that declares `:---` says so
+
+`cellAlignStyles` had entries for `center` and `end` only, so a column written
+`| :--- |` got no style at all and rendered exactly like an undeclared one. It
+looks right in isolation, because a cell already starts at the inline start —
+but `text-align` inherits, so the column follows whatever the table is nested
+in. Rendered inside a centered block, a table whose first column explicitly
+declares `:---` centers it, and there is nothing in the markdown that says it
+should.
+
+A declared alignment is now always written out, `start` included. Only the
+`:---` case changes, and only where something around the table had already
+moved the default.
+
+@lexs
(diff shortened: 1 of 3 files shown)


rubyycheung (changes requested):
Thanks for calling out the RTL judgment explicitly. I checked the behavior against GFM, GitHub’s renderer, remark-gfm, and markdown-it; they all preserve physical left/right alignment under RTL. ASTRYX’s parser also represents these values as `left | center | right`.

Because this component renders author-specified GFM formatting, translating `:---` to logical `start` and `---:` to logical `end` changes the document’s meaning in RTL. Could we map both values to physical `left` and `right`, use a narrowly scoped `@astryx/no-physical-properties` suppression for this semantic exception, and add RTL coverage?

The logical-property rule makes sense for UI layout, but explicit Markdown alignment is source-document formatting and should remain compatible with the originating syntax.

PR #5448, opened 2026-08-24
feat(templates): add the table-filter page template

ernestt:
A filterable table page template (`Table - Filtering`) that ports the feature set of the internal XDS table page pattern onto Astryx primitives.

https://github.com/user-attachments/assets/9baf8218-f61a-4725-8101-eb75017ff131

## What it demonstrates

- **Filter token list** of quick-filter toggles and field controls — `Selector`, `MultiSelector`, and a `ComplexSelector` wrapping a range `Slider` — that swaps to `PowerSearch` for anything the tokens can't express. Both modes read and write the same `PowerSearchFilter[]`, so a clause built in either survives the swap. Controls carry field chrome when unset and a pressed fill once they hold a value, so the row reads as one family whether a clause came from a toggle or a selector.
- **Saved views** capturing both the filters and the whole table configuration, with create/edit dialogs that render the captured state as a `MetadataList` and the filters as `Token`s.
- **View options popover** with four panels — a drag-and-drop column transfer list, density, sticky edges, and grouping — applying instantly, no apply/cancel footer.
- **Row click** opens a resizable detail panel; **selection** raises a bulk-edit bar that slides in with `@starting-style`.
- **Infinite scroll** against an `IntersectionObserver`, with skeleton rows aligned to the table's own column grid, plus empty states for the no-results and no-data paths.
- **Container-query toolbar** that wraps to a second row based on the width the detail panel leaves it, not the viewport.

Content-only (root `Layout`); host supplies the shell.

## Template Grade: Filterable Table

**Type**: page
**File**: `packages/cli/assets/templates/pages/table-filter/page.tsx`
**Grade**: **B (85/100)**

| Category | Points | Max | Details |
|----------|--------|-----|---------|
| Astryx Component Purity | 30 | 30 | 0 raw HTML elements |
| Icon Purity | 15 | 15 | 0 raw SVG; every icon goes through `Icon` or an `icon=` prop |
| Custom CSS | 0 | 15 | 213 declarations — see below |
| Layout & Structure | 15 | 15 | Roots in `Layout`, no `AppShell`/global chrome, single page, all navigation inert |
| Doc Metadata | 10 | 10 | `type`, `name`, `description`, `isReady` all present and accurate |
| Image Handling | 5 | 5 | No images |
| Code Quality | 10 | 10 | `'use client'`, default export, realistic mock data, no dead code |
| **TOTAL** | **85** | **100** | |

### Custom CSS (213 declarations) — the one category that fails

This is the honest finding and the top fix. For calibration, the next-highest page template in the gallery is `kanban-board` at 37, and 38 of 46 templates sit at 0. Every declaration here is token-backed, and the bulk of it traces to four things the design system doesn't currently express:

1. **Filter control chrome** (`filterChrome`, `filterSurface`, `filterFill`, `filterLabelValue/Empty`) — `ButtonVariantMap` has no `outline` variant, so the `Button`- and `ToggleButton`-backed filters hand-paint the border/radius/weight that the `Selector`s beside them draw natively.
2. **The column transfer list** — `@astryxdesign/lab`'s `TransferList` is unpublished and doesn't resolve from the template viewer, so its panel chrome is restated against the same tokens.
3. **Flush containers** — neither `Popover` nor `DialogHeader` publishes a padding prop, so reaching a flush container (needed for edge-to-edge dividers) goes through `xstyle`.
4. **Skeleton rows** — they have to align to `Table`'s internal column grid, which means reproducing its flex sizing and its `--container-padding-*` edge compensation.

The durable fix is upstream: an outline `Button` variant, a published `TransferList`, and padding props on `Popover`/`DialogHeader` would remove most of it. Happy to split those out if you'd rather gate this template behind them.

### Fixes already applied in this PR

- Replaced **17 raw `<div>`s** with `VStack`/`HStack` (purity 4 → 30). `BaseProps` keeps `role`/`aria-*`/`id` and `Stack` forwards `ref`, so the sentinel and the two `role="group"` panels convert cleanly.
- Swapped `xstyle` for the equivalent Astryx props wherever one exists — `padding*`, `minHeight`, `width`, `wrap` — which also deleted four style names outright.
- Rewrote the changeset, which had gone stale against the final template (it still said "saved searches", "grouped column visibility", and "a choice between pagination and infinite scroll").

## Notes for the reviewer

- **Lifecycle.** Per `templates.instructions.md`, new templates are normally staged hidden. `PageTemplateDoc` has no `hidden` field (the CLI's template discovery doesn't read one) — the page-template equivalents are `isHiddenFromOverview` or `isReady: false`. This lands visible on the strength of the B grade above; say the word and I'll flip it to `isHiddenFromOverview: true` until the audit is recorded in the wiki ledger.
- **`lucide-react` import.** The rubric's code-quality line names `@heroicons/react`, but `theme-showcase` already ships with `lucide-react` and it's a declared `apps/docsite` dependency. Worth noting that it is *not* declared in `apps/sandbox/package.json` and resolves there only by hoisting — pre-existing, shared with `theme-showcase`, and out of scope here, but it will bite eventually.
- **`.gitignore`.** Adds `.worktrees/` and `worktrees/`. There are currently eight in-repo worktrees; without the ignore a `git add -A` in the main tree sweeps an entire second checkout into the commit.
- **The group heading is the built-in one, deliberately.** #5454 pins the heading to the table's start edge by shrink-wrapping it, and applies that only to the heading the plugin renders itself — a custom `renderGroupHeader` keeps full width, which leaves a sticky element no room to travel. This template freezes the first column, so it scrolls sideways by design, and with a custom heading the chevron measured x=-388 after a 400px scroll: heading and collapse control both gone. The custom renderer here was setting a label and a count in two weights, which the built-in heading already does, so it was dropped. Cost is the wording — `6 jobs` became `(6)`; the chevron now holds at x=4. Worth knowing that any *other* consumer wanting a custom heading still has this problem, so the plugin arguably needs to pin the chevron regardless of who renders the heading.

## Test plan

- [x] `pnpm -F @astryxdesign/cli typecheck:template-docs` — passes
- [x] `pnpm build && pnpm -F @astryxdesign/cli typecheck:strict` — passes
- [x] `pnpm check:repo` — all gates pass (incl. `check:changesets`, `check:sync`)
- [x] `npx eslint` on the template — clean
- [x] `node scripts/sync-templates.js` — generates the sandbox route
- [x] Rendered in the template viewer at 1440/900/700px in **light and dark**: default table, view options across all four tabs (height parity holds), detail panel, bulk bar, PowerSearch swap, wrapped toolbar, infinite-scroll skeletons. **Zero console or page errors** in every state.
- [x] Rebased onto `main` after #5454 merged and re-verified against it, no conflicts. Group heading measured in the running template: chevron at x=4 both at rest and scrolled 400px, sections heading `Beacon Street Bistro (10)` / `Elm & 5th Lofts (9)`, no console errors.

### DIFF
--- a/.changeset/table-filter-template.md
+++ b/.changeset/table-filter-template.md
@@ -1,0 +1,30 @@
+---
+'@astryxdesign/cli': patch
+---
+
+[feat] New `table-filter` page template: a table page built around a filter token list
+
+Ports the feature set of the internal XDS table page pattern onto Astryx
+primitives. The filter row is a token list of quick-filter toggles and field
+controls — `Selector`, `MultiSelector`, and a `ComplexSelector` wrapping a
+range `Slider` — that swaps to `PowerSearch` for anything the tokens can't
+express. Both modes read and write the same `PowerSearchFilter[]`, so a filter
+built in either survives the swap. Controls carry field chrome when unset and
+a pressed fill once they hold a value, so the row reads as one family whether
+a clause came from a toggle or a selector.
+
+Around that: saved views that capture the filters and the whole table
+configuration, a bulk-edit bar that slides in on selection, and a view options
+popover with four panels — a drag-and-drop column transfer list, density,
+sticky edges, and grouping — that apply instantly. Clicking a row opens a
+resizable detail panel. The list pages in by infinite scroll against an
+`IntersectionObserver`, with skeleton rows aligned to the table's own column
+grid standing in for the batch in flight, and empty states for the no-results
+and no-data paths. The toolbar wraps to a second row under a container query
+rather than a viewport one, so it responds to the width the detail panel
+leaves it.
+
+Uses the `Table - Filtering` category, already reserved in the
+`TemplateCategory` union.
+
+@ernesttien
--- a/.gitignore
+++ b/.gitignore
@@ -56,6 +56,13 @@ apps/sandbox/next-env.d.ts
 apps/example-nextjs-tailwind/.next
 .parity-test-results.json
 .worktree-parity-test
+
+# Scratch git worktrees. These live inside the repo for convenience, so without
+# an ignore a `git add -A` in the main tree sweeps an entire second checkout
+# into the commit.
+.worktrees/
+worktrees/
+
 # Patch files
 *.patch
 *.rej
--- a/apps/docsite/src/components/templateComponents.ts
+++ b/apps/docsite/src/components/templateComponents.ts
@@ -154,6 +154,10 @@ export const TEMPLATE_COMPONENTS: Record<
   table: lazy(
     () => import('../../../../packages/cli/assets/templates/pages/table/page'),
   ),
+  'table-filter': lazy(
+    () =>
+      import('../../../../packages/cli/assets/templates/pages/table-filter/page'),
+  ),
   'table-grouped': lazy(
     () =>
       import('../../../../packages/cli/assets/templates/pages/table-grouped/page'),
--- a/packages/cli/assets/templates/pages/table-filter/page.tsx
+++ b/packages/cli/assets/templates/pages/table-filter/page.tsx
@@ -887,14 +887,31 @@ const GROUPING_OPTIONS: ReadonlyArray<{value: GroupField; label: string}> = [
 ];
 
 /**
+ * The words a section is titled with. The group key is what the heading prints,
+ * so it has to be the label and not the stored value — grouping by status would
+ * otherwise head its sections "on_hold". Technician and customer are stored as
+ * their own labels already, which is what the fallback covers.
+ */
+function groupKeyOf(job: ServiceJob, field: GroupField): string {
+  if (field === 'none') {
+    return '';
+  }
+  const stored = String(job[field]);
+  return VALUE_LABELS[stored] ?? stored;
+}
+
+/**
  * Section order per field. Without it the sections would come out in whatever
  * order the active sort happened to surface them, so status and priority take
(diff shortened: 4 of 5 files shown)


ernestt:
Draft until #5454 lands.

The default view now opens grouped by customer, and its section headings count the result set through `getGroupTotal` — a grouping-plugin option that #5454 adds. `build-storybook` and Vercel compile this template against core, so they are red here until that merges. Nothing to fix on this branch.

Two things also worth knowing for review, both of which #5454 fixes rather than this PR:

- The frozen first column and the grouped headings each hit a core defect that cannot be worked around from userland — `renderGroupHeader` can pin its own content, but the chevron belongs to the plugin and sits outside that slot.
- Measured on this template with a 400px horizontal scroll: without #5454 the chevron computes `position: static` and slides the full 400px; with it, `sticky` and zero drift.

Sequence is: merge #5454, rebase this on main, CI goes green, un-draft.

ernestt:
Correction to the note above: this no longer waits on #5454.

Grouped views now page by whole sections rather than by row count — the cut starts at `PAGE_SIZE` and carries to the end of whichever section it lands in. A section is therefore only ever rendered complete, which retires the `getGroupTotal` wiring that needed the unmerged core API: with no partial section left to describe, the batch count and the result count are the same number. Verified against pristine `main` core — typecheck and lint both clean.

#5454 is still worth landing on its own merits; this branch just no longer depends on it to compile. Staying in draft for the template rubric pass.

Issue #5451, opened 2026-08-24
[Bug] The pressed overlay never paints: a media-nested :hover out-specifies :active library-wide


cixzhang:
## Description

Every component that pairs the house hover pattern with a `:active` overlay ships a **pressed state that never paints**. The two rules both set `background-image` on the same element, and the hover rule out-specifies the pressed rule — so while the pointer is down, and therefore also hovering, the hover tint is what you see.

This is the same root mechanism as #5442 (StyleX priority buckets emitting specificity that contradicts author intent), but a different instance: there it is `:where()` lowering a bucket, here it is `@media` raising one.

## Why it happens

StyleX encodes priority by repeating the atomic class and appending `:not(#\#)` boosts. A pseudo-class **nested inside a media query** lands in a higher bucket than a bare pseudo-class, so it gets more class repeats:

| | compiled selector (`packages/core/dist/astryx.css`) | specificity |
|---|---|---|
| hover | `@media (hover: hover){.xmvprkv.xmvprkv.xmvprkv:hover:where(…):not(#\#)×3}` | **(3,4,0)** |
| pressed (Button) | `.xq8i9tn:active:not(#\#)×3` | **(3,2,0)** |
| pressed (guarded form) | `.x7uyq82.x7uyq82:active:where(…):not(#\#)×3` | **(3,3,0)** |

Same ID count, so class count decides, and hover wins in both forms. Nothing about the authored source suggests this — the two keys sit next to each other in one `backgroundImage` object, and the `:active` one is written last.

## Reproduction

Real Chromium against a Storybook build of `main` (cd1a2daa59a). Hover a control, then `mouse.down()`, reading `getComputedStyle(el).backgroundImage`:

```
Calendar day   rest   none
               hover  linear-gradient(rgba(0,0,0,0.05), rgba(0,0,0,0.05))
               active linear-gradient(rgba(0,0,0,0.05), rgba(0,0,0,0.05))   ← identical
```

`0.05` is `--color-overlay-hover`; `--color-overlay-pressed` is `0.098` and never appears.

Independently observed in the ButtonGroup audit on 2026-08-23 (#5389), recorded there as "the background is identical to hover" and attributed to grouping — it is not about grouping.

## Affected

[...]

Issue #5442, opened 2026-08-24
[Bug] SelectableCard/Thumbnail/ClickableCard lose their hover overlay at 0.4.6 — #5247's :where() guard drops the StyleX specificity boost 7→3


brightonvino:
## Description

#5247 added a zero-specificity guard to every self-`:hover` selector:

```
:hover  →  :hover:where(:not(:disabled,[aria-disabled="true"]))
```

That reasoning is right and `:where()` does weigh (0,0,0). But on `SelectableCard`, `Thumbnail` and `ClickableCard` the `:hover` key **also targets a pseudo-element** (`:hover::after`), and adding the guard changed the StyleX **priority bucket** that key lands in. StyleX derives its `:not(#\#)` specificity boost from the bucket index, so the hover rule's boost dropped from **7 IDs to 3**, while the resting `background-color: transparent` on the same `::after` kept all 7.

`:not(#\#)` contributes ID-level specificity, and IDs dominate class count:

| | selector | specificity |
|---|---|---|
| 0.4.5 hover | `.xi14tyy.xi14tyy:hover` + 7×`:not(#\#)` + `::after` | **(7,3,1)** → wins |
| 0.4.7 hover | `.x1912f9e.x1912f9e.x1912f9e:hover:where(…)` + 3×`:not(#\#)` + `::after` | **(3,4,1)** → loses |
| both, resting | `.xyhc2n1` + 7×`:not(#\#)` + `::after` | **(7,1,1)** |

The resting reset now out-specifies the hover rule, in the same `@layer astryx-base`, so **the hover overlay never paints**. `SelectableCard` and `Thumbnail` lose their only hover affordance (WCAG 1.4.1). This is Astryx-vs-Astryx — no consumer CSS is involved (see Reproduction).

Expected: an enabled card paints `--color-overlay-hover` on its `::after` on hover, as at 0.4.5.
Actual: the `::after` stays `rgba(0, 0, 0, 0)` at 0.4.6 and 0.4.7.

### The rules, verbatim from `dist/astryx.css`

0.4.5:
```css
@media (hover: hover){.xi14tyy.xi14tyy:hover:not(#\#):not(#\#):not(#\#):not(#\#):not(#\#):not(#\#):not(#\#)::after{background-color:var(--color-overlay-hover)}}
.xyhc2n1:not(#\#):not(#\#):not(#\#):not(#\#):not(#\#):not(#\#):not(#\#)::after{background-color:transparent}
[...]

PR #5453, opened 2026-08-24
fix(core): Calendar nightly audit — provider locale in an announcement, 320px reflow, forced-colors selection

cixzhang:
Nightly component audit of **Calendar** (`core`), rubric **v1.8**, against `cd1a2daa59a`.

|---|---|---|---|
| Before | 78.6 | C | 2 |
| After | 85.9 | B | 0 |

Every finding below was measured in real Chromium before and after — not read off the source.

---

### 🔴 I21 — the cleared-range announcement was spoken in the host locale

> **Overlaps open work.** [#5296](https://github.com/facebook/astryx/pull/5296) fixes this at the root (`plainDateFormat`'s locale default) and has been approved since 24 Aug; [#5303](https://github.com/facebook/astryx/pull/5303) threads the provider locale through all twelve call sites, rewrites this same function, and adds a lint that makes the class of bug unrepresentable. **Land #5296 first.** This one-line change and its test are kept because they close the bug tonight and the test is worth keeping under either of those; expect a trivial conflict when #5303 rebases. The two of them also collide with each other — same changeset filename.

`Calendar.tsx:465` formatted its date with `plainDateFormat(date, DATE_FORMAT_WITH_WEEKDAY)` and omitted the `locale` argument that the other **nine** call sites in the file pass. `Intl.DateTimeFormat(undefined, …)` falls back to the browser's locale, so this one announcement ignored `InternationalizationProvider`.

The new test makes it concrete — with the provider on `ja-JP`:

```
Expected: "2026年1月10日土曜日"
Received: "Cleared start date Saturday, January 10, 2026. Select a start date."
```

The expected string is the day button's *own* accessible name, which was already localized correctly. Red before, green after.

### 🔴 R1 — `numberOfMonths={2}` overflowed a 320px viewport

Measured at 320px CSS width:

| | before | after |
|---|---|---|
| `scrollWidth` / `clientWidth` | **504** / 320 | **320** / 320 |
| calendar width | 488px | 288px |
| next-month button right edge | 492px — **off screen** | 292px — on screen |

The document scrolled sideways (WCAG 1.4.10) and the next-month button sat outside the viewport. The docs actively recommend this configuration — *"Show two months side by side when the user frequently selects dates that span a month boundary"* — so the layout that failed is the one the component tells people to use.

`monthsContainer` now wraps. Above 488px nothing moves; below it the second month stacks.

**Before / after at 320px:**

| before | after |
|---|---|
|  |  |

### 🟠 A14 — the selected date vanished under forced colors

Under `forced-colors: active` the selected day read `backgroundColor rgb(255,255,255)` / `color rgb(0,0,0)` / `forced-color-adjust auto` — identical to every unselected day, so a Windows High Contrast user could not see which date was selected.

Fixed with the pattern `ToggleButton.tsx:56` already uses: `Highlight` / `HighlightText` plus `forced-color-adjust: none`, which a `<button>` needs because the UA otherwise keeps the native ButtonFace surface and ignores the authored fill. After: `bg rgba(5,0,73,0.8)`, `color rgb(255,255,255)`, `forced-color-adjust none`. A compiled-rule test guards it, in the same shape as the CheckboxInput / Skeleton / ToggleButton ones.

| before | after |
|---|---|
|  |  |

### 🟡 D4 — a literal radius where a role token exists

`borderRadius: '50%'` on the day button → `radiusVars['--radius-full']`. Equivalent only while the box stays square, which is exactly the caveat the radius convention names — and the review found it does more than I claimed: **in y2k the range endpoints stop poking out of a square band.**

### ↩︎ Withdrawn: the ten `2px` insets

I originally swapped these for `--spacing-0-5` and the review loop blocked it, correctly. That `2px` is `(--size-element-md − --size-element-sm) / 2` — the gap between the day button and its cell — not a step on the spacing scale, and the two only coincide on the neutral theme:

```
neutral   (32 − 28)/2 = 2    --spacing-0-5 = 2px    hit target 32 in a 32px cell
matcha    (40 − 36)/2 = 2    --spacing-0-5 = 3px    hit target 42 in a 40px cell
```

On matcha the tokenised version grew the hit target past its cell and overlapped adjacent days by 2px. Reverted to the literal, with the geometry recorded in a comment so the next reader does not make the same swap. An absolute inset is a layout value and is sanctioned as-is.

Deriving it with `calc()` from the two size tokens would be better than either, and would also close butter's pre-existing 4px dead gap between adjacent day hit targets. That is its own measured change, not this one.

### 🧹 Dead code

An unused `srOnly` style that hand-rolled `VisuallyHidden` (§1 A17) with a deprecated `clip: rect(…)`. Nothing referenced it.

---

## Needs review — carried, not fixed

**System-level, filed as #5451: the pressed overlay never paints, library-wide.** Calendar's day button has no pressed state, and it turns out it cannot get one at the component level. I added the house `:active` overlay, measured it, and found it dead: StyleX gives a `@media`-nested `:hover` a higher priority bucket than a bare `:active`, so the compiled hover selector out-specifies the pressed one *(3,4,0)* vs *(3,3,0)* and wins while the pointer is down — which is the only time `:active` matches. The same masking applies to `Button`'s four variants and `AvatarGroupOverflow`; `Button` only *looks* fine because its `transform: scale(0.98)` still fires. I backed my change out rather than ship dead CSS that reads like a fix, and filed one issue with the compiled-CSS evidence. Cousin of #5442.

Everything else here needs a decision rather than a correction, so none of it is touched:

- **A8/A9 — the range preview is hover-only.** `onDayHover` is wired to `onMouseEnter`/`onMouseLeave` and nothing else, so a keyboard user arrowing toward the second endpoint sees no preview of the range they are about to commit. Wiring it to focus is a behaviour change, not a correction.
- **P30 — no `label` prop.** A consumer cannot name the widget for its purpose ("Departure date"); the grid's only accessible name is the month it happens to be showing. Adding one is new public API.
- **P25 — no `changeAction`**, where the convention says async-capable core components ship one. Also new API.
- **A10 — no coarse-pointer hit expansion.** The day button is 28×28 in a 32×32 cell, over the 24px floor. Expanding to ~44px on coarse pointers would overlap the neighbouring days at a 32px pitch, which is worse; recorded with that reason rather than "fixed".
- **A7 / C3 — two Effects, one announcing and one moving focus.** Both are on C3's bright line, but the harm the rule guards (a double fire) is provably absent and under test, and neither has an obvious handler to move to: a month change can originate in a controlled `focusDate` prop, which has no callback, and the focus target does not exist until the new grid has rendered.
- **A14, part two — the *today* ring also disappears under forced colors** (box-shadow is stripped). The fix wants an outline, which would collide with the focus ring; left for a decision rather than guessed at.
- **Both themed stories pin `mode="light"`** on their `<Theme>`, so under the dark toolbar global they render light-mode tokens against a dark page and are close to illegible. Visible in the contact sheet. A story fix, and `<Theme>`'s nested-mode semantics are subtle enough that I did not want to guess in this PR.

---

## Verification

- `pnpm test` — Calendar 148 pass (146 + 2 new). Full suite was 7 red on the first push — all in `packages/cli`, none Calendar's. [#5452](https://github.com/facebook/astryx/pull/5452) fixed those fixtures on main; **main is merged in** and the two that CI blocks on now pass locally.
- `pnpm build`, `pnpm lint:strict` (scoped), `check:sync`, `check:changesets`, `typecheck:docs`, `check-use-client` — clean.
- `pnpm a11y:audit --components Calendar` — 0 violations over 20 stories, **0 new baseline entries, baseline file untouched**.
- `pnpm rtl:audit --filter Calendar` — measured, not N-A'd: D1 auto-pass on 2 icons, curated D2 pass, 0 fail, 0 surprise.
- **92 Chromium captures** over 16 stories × light / dark / matcha, including driven hover, focus-visible, active, mid-flight range preview, forced colors and 320px — [contact sheet on `assets/pr-5453`](https://github.com/facebook/astryx/tree/assets/pr-5453).

Changeset is `patch`. No API added, no behaviour broken.

### DIFF
--- a/.changeset/calendar-nightly-audit-20260824.md
+++ b/.changeset/calendar-nightly-audit-20260824.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Calendar: announce a cleared range in the provider locale, wrap the two-month layout instead of overflowing a narrow viewport, keep the selected date visible under forced colors
+
+@cixzhang
--- a/apps/storybook/stories/Calendar.stories.tsx
+++ b/apps/storybook/stories/Calendar.stories.tsx
@@ -133,6 +133,28 @@ export const TwoMonthsRangeSelection: Story = {
   },
 };
 
+/**
+ * The two-month layout inside a container narrower than the two months need.
+ * The second month wraps below the first instead of overflowing, so a 320px
+ * viewport does not scroll sideways and the next-month button stays on screen.
+ */
+export const TwoMonthsNarrowContainer: Story = {
+  render: () => {
+    const [value, setValue] = useState<ISODateString | undefined>(undefined);
+    return (
+      <div style={{width: 320}}>
+        <Calendar
+          mode="single"
+          numberOfMonths={2}
+          value={value}
+          onChange={val => setValue(val)}
+          focusDate="2026-01-01"
+        />
+      </div>
+    );
+  },
+};
+
 export const MinMaxBoundary: Story = {
   render: () => {
     const [value, setValue] = useState<ISODateString | undefined>(undefined);
--- a/packages/core/src/Calendar/Calendar.test.tsx
+++ b/packages/core/src/Calendar/Calendar.test.tsx
@@ -13,6 +13,10 @@ import {describe, it, expect, vi, beforeEach, afterEach} from 'vitest';
 import {act, render, screen, within, waitFor} from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import {getButton} from '../__tests__/fastRoleQueries';
+import {
+  getForcedColorsRules,
+  getAllInjectedCss,
+} from '../__tests__/forcedColors';
 import * as stylex from '@stylexjs/stylex';
 import {Calendar} from './Calendar';
 import type {CalendarHandle} from './Calendar';
@@ -387,6 +391,33 @@ describe('Calendar', () => {
     // the old anchor are selectable again, so a new start can be placed there.
     expect(getDayButton(11)).not.toBeDisabled();
     expect(getDayButton(9)).not.toBeDisabled();
+  });
+
+  it('announces the cleared range in the provider locale', async () => {
+    const user = userEvent.setup();
+
+    render(
+      <InternationalizationProvider locale="ja-JP">
+        <Calendar mode="range" focusDate="2026-01-01" />
+      </InternationalizationProvider>,
+    );
+
+    // The day button's own accessible name is already localized through the
+    // provider. The announcement has to use the same date string: formatting
+    // it without the locale silently falls back to the host locale, so a
+    // Japanese user hears an English date.
+    const anchor = document.querySelector<HTMLButtonElement>(
+      'button[data-date="2026-01-10"]',
+    );
+    const localizedDate = anchor?.getAttribute('aria-label');
+    expect(localizedDate).toBeTruthy();
+
+    await user.click(anchor as HTMLButtonElement);
+    await user.click(anchor as HTMLButtonElement);
+
+    await waitFor(() => {
+      expect(politeRegion()?.textContent).toContain(localizedDate);
+    });
   });
 
   it('does not apply range-span constraints in single mode', async () => {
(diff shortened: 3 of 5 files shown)


cixzhang:
Thanks — the 320px reflow and the forced-colors selection are both right, and the radius change does something you didn't claim: in y2k the endpoints stop poking out of a square band.

| before | after |
|---|---|
|  |  |

The ten `2px` insets aren't spacing. That number is the gap between the day button and its cell — your own comment says so:

```
neutral   (32 − 28)/2 = 2    --spacing-0-5 = 2px    hit 32 in a 32px cell
matcha    (40 − 36)/2 = 2    --spacing-0-5 = 3px    hit 42 in a 40px cell
```

So on matcha adjacent days' hit targets overlap by 2px. The on-scale rule covers padding, margin and gap; an absolute inset is a layout value, sanctioned as-is.

On the locale line: #5296 fixes that at the root and was approved this afternoon; #5303 threads the provider locale through all twelve call sites and rewrites this same function. Worth landing #5296 instead of patching call sites one at a time.

Needs a merge with main too; #5452 fixed the fixtures CI is red on.

Revert the ten insets and I'm happy with the rest.

MERGE: hold(request changes — the inset tokens)

[Reviewed by Robohands]

cixzhang:
Fixed in f0b75aa — reverted to the literal, with the geometry recorded in a comment so the next reader doesn't make the same swap. Confirmed both halves of your evidence against the theme sources: matcha sets `--spacing-0-5: 3px` with `--size-element-{sm,md}` at 36/40px, and y2k sets `--radius-full: 0px`, which is why the endpoints go square. Main is merged in and the two CI fixtures that were red pass locally.

Carried, not fixed: deriving the inset with `calc()` from the two size tokens, which would also close butter's pre-existing 4px dead gap between adjacent day hit targets. That wants its own measured change.

cixzhang:
Thanks — insets reverted, geometry matches main on all four themes.

The body undersells this: in DateInput's popover the second month was clipped off-screen, not just overflowing.

| before 320×568 | after |
|---|---|
|  |  |

MERGE: auto

[Reviewed by Robohands]

cixzhang:
Nit, not blocking — `packages/core/src/Calendar/styles.ts` `monthsContainer`: the wrap wants 544px of height, so at 320×480 September's tail goes under the fold.



Still strictly better than the before, where the month was unreachable rather than below it. Carried in the ledger row rather than fixed here.

PR #5454, opened 2026-08-24
fix(Table): two defects surfaced by frozen columns and grouped rows

ernestt:
Two defects in the Table plugins, both found building the `table-filter` page template (#5448) — its default view freezes the first column, groups the rows, and loads them in batches. Both show up only once `useTableStickyColumns` has a start edge.

A third — a grouped section counting the page instead of the result set — was fixed here too, then dropped. That story is below, since the commit history still shows it going in and coming out.

## The checked-row wash stops at the freeze line

A sticky cell paints an opaque background so scrolled content cannot show through, then replays the row's fill on a `::before` layer from `--table-row-overlay`. `TableRow` publishes that variable alongside every background it sets — that is how striping and hover carry across the boundary.

`useTableSelection` did not. It wrote `backgroundColor` on the `<tr>` and stopped, so a checked row read as selected on the scrollable side and unselected under every pinned column. The wash and the variable are now written together, and withdrawn together when a row is unchecked or `hasRowHighlight` is turned off. Both writes are guarded, so a row that already holds the right value is left alone rather than being re-written on every owner render.

_Checked row SJ-1965, table scrolled 360px right. The frozen column keeps the unselected fill while the scrollable side is washed, so one row reads as both selected and not._

**Before**

Checked row: frozen column unwashed, scrollable side washed

**After**

Checked row: wash continuous across the freeze line

Both shots use the default `hasRowHighlight`, which is what the fix is about. #5448 has since set it to `false` — that template gives row background a different job, and the flag exists precisely so a consumer can decline the wash. So the template no longer reproduces this one; any table taking the default still does.

## The group heading scrolls away from the rows it names

A group header is one cell spanning every column, so on a table scrolled sideways its chevron and label slid out of view while the columns they named stayed pinned — leaving unlabelled grey bands and no reachable collapse control. The heading now sticks to the table's start edge.

The cell's start gutter moves onto the heading so it travels with it. Left on the cell, the heading would sit one gutter in at rest and jump flush the moment it stuck.

The pinned element takes `width: fit-content` only for the default heading. A custom `renderGroupHeader` keeps the full width it had, so a renderer laying content out across the row is not silently shrink-wrapped.

Userland could only fix half of this. `renderGroupHeader` can pin its own content, but the chevron is the plugin's and sits outside that slot, so the two come apart.

_Table scrolled 360px right. The chevron measures x=-356 with `position: static`, leaving the grey band with nothing naming it and no reachable collapse control; after, it holds at x=4 as `sticky`._

**Before**

Group heading row empty, chevron and label scrolled off

**After**

Group heading pinned at the start edge while scrolled

## The section count: fixed, then dropped

Grouping runs on the rows it is handed. Under pagination that is one page, so a heading read `Northwind (6)` — a progress meter wearing a total's clothes — and silently became `(10)` as the reader scrolled.

This carried a `getGroupTotal(groupKey)` prop to let a consumer report the real size. It is gone, for two reasons. The mechanical one: the getter is naturally written inline at the call site, which invalidated the hook's `useMemo` on every render and re-rendered every row — a real cost paid by everyone to fix a heading. The design one: it is public API surface with no consumer. #5448 was the only one, and it solved the same problem without any API at all, by ending a page on a section boundary instead of a row count. So there is nothing left to shape the prop against.

What survives is documentation. The hook now states the ordering it needs — filter, sort, slice, then group, sorting by the group key ahead of the reader's own keys, the same `ORDER BY group, sort` a backend would run — which is what keeps a section's rows contiguous so pages append at the bottom rather than splicing rows in above the fold. It also states the boundary trick and its one caveat: carrying the cut to the end of the section it lands in keeps every rendered section whole, and needs the full result set in hand, so it is open to a client-side list and closed to cursor pagination. Both are the consumer's `useMemo` order and unfixable from inside the hook, so they are written where someone reaching for it will read them.

If a server-paged grouped table turns up later and still wants an honest count, the prop is recoverable at `e8df392c9a8` with its tests.

## Notes

This is the core half of a pair, though the pairing has loosened. #5448 surfaced both defects and originally worked around each; it now depends on one.

- **The wash.** #5448 sets `hasRowHighlight: false`, giving row background a single meaning — open in the detail panel — so a checked row and an inspected row stop competing for the same signal. With no wash to carry, that template no longer exercises this fix. Any table taking the default still does.
- **The heading pin.** Still a real dependency. #5448 has dropped its own label pin, and this is the one defect that cannot be worked around from userland, so the heading is visibly broken there until this lands. Measured with a 400px horizontal scroll: without this PR the chevron computes `position: static` and slides the full 400px with the content; with it the chevron computes `sticky` and holds at x=4 while body cells move to -400.

The two PRs remain independent to merge, in either order.

`useTableStickyColumns` needs no change. It already pins the whole contiguous run from the first column through the last key it is given, so naming the first data column carries the selection checkbox with it — a template that lists the selection column explicitly is working around a problem it does not have.

I also checked whether the group header row needed the same `--table-row-overlay` treatment as selection. It does not: the header replaces its cells with a single `colSpan` cell, which is never one of the sticky plugin's columns, so nothing paints over it. Verified by pixel-sampling either side of the freeze line with the fix reverted.

## Test plan

- [x] `packages/core/src/Table` — 498 tests pass, 5 new
- [x] Five cases cover `--table-row-overlay`: published with the wash, absent when `hasRowHighlight` is false, withdrawn on uncheck, tracking the flag both directions, never set on unchecked rows
- [x] Typecheck, lint, `check:changesets` and `check:i18n-catalog` clean; catalog consistent across 29 locales
- [x] Verified in a browser against the `table-filter` template: pixel-sampled the row fill 60px either side of the freeze line. Selected row was `255,255,255` / `241,241,241` (mismatched) before and matches after; chevron and label stay pinned and together while scrolled. The selection reading was taken under the default `hasRowHighlight` — see the note under that section
- [ ] No unit test for the grouped-heading pin — it is StyleX-only, and the house pattern in `useTableStickyColumns.test.tsx` asserts inline styles because JSDOM cannot resolve StyleX classes


cixzhang (changes requested):
Thanks — the selection wash and default group heading now work across the browser matrix. Three far-side blockers remain. A full-width `renderGroupHeader` collapses from 1,334px to 180px. With the documented inline `getGroupTotal`, a builder's unrelated state update on a 10,000-row table re-rendered all 10,000 data cells; selection's unguarded overlay sync made 20,000 no-op calls in Storybook dev. The exact base did neither. The new total API also has no current consumer—[#5448](https://github.com/facebook/astryx/pull/5448) now loads whole sections—and the existing renderer can already show a supplied total. Could we preserve the custom slot's width, skip unchanged overlay writes, and split/drop `getGroupTotal` until it has a consumer and preserves row memoization?


[Full review](https://github.com/cixzhang/astryx/wiki/Review-5454)

[Reviewed by Robohands]

packages/core/src/Table/plugins/groupedRows/useTableGroupedRows.tsx
```diff
@@ -126,15 +153,27 @@ const styles = stylex.create({
...
+  // the inner row to the start edge keeps the chevron and the label together
+  // and on screen. `fit-content` because a flex child would otherwise stretch
+  // to the cell's full width and have nothing to slide within.
   headerInner: {
     display: 'flex',
     alignItems: 'center',
     gap: spacingVars['--spacing-1'],
+    insetInlineStart: 0,
+    position: 'sticky',
+    width: 'fit-content',
```
cixzhang: This shrink-wraps public `renderGroupHeader` content: my full-width case collapsed 1,334px → 180px.

packages/core/src/Table/plugins/selection/useTableSelection.tsx
```diff
@@ -159,8 +159,17 @@ function applyRowSelectionStyle(
...
   // row is already selected, and the wash has to come back off.
-  el.style.backgroundColor =
-    isSelected && hasRowHighlight ? selectedBgColor : '';
+  const wash = isSelected && hasRowHighlight ? selectedBgColor : '';
+  el.style.backgroundColor = wash;
+  // A pinned column paints its own opaque background over the row, so the
+  // wash alone stops at the freeze line. Publishing it as the row overlay is
+  // what lets useTableStickyColumns replay it on those cells — the same
+  // contract TableRow honours for striping and hover.
+  if (wash === '') {
```
cixzhang, line 171: This adds a no-op CSS-variable mutation per row on every owner render; guard unchanged values.

packages/core/src/Table/plugins/groupedRows/useTableGroupedRows.tsx
```diff
@@ -383,7 +467,7 @@ export function useTableGroupedRows<T extends Record<string, unknown>>(
         };
       },
     }),
-    [collapsedGroups, onToggleGroup, renderGroupHeader, t],
+    [collapsedGroups, getGroupTotal, onToggleGroup, renderGroupHeader, t],
```
cixzhang: The documented inline getter invalidates every row: 10,000 data rows re-rendered; base stayed at zero.

packages/core/src/Table/plugins/groupedRows/useTableGroupedRows.tsx
ernestt: Fixed — split into `headerInner` (no width constraint) + `headerInnerFitContent` (applied only for the built-in default heading). Custom `renderGroupHeader` content now gets the full column width.

packages/core/src/Table/plugins/selection/useTableSelection.tsx
ernestt: Guarded — now checks `el.style.backgroundColor !== wash` and `getPropertyValue(ROW_OVERLAY_VAR) !== wash` before writing. Unchanged rows skip both the backgroundColor set and the CSS variable mutation entirely.

packages/core/src/Table/plugins/groupedRows/useTableGroupedRows.tsx
ernestt: Agreed — dropped `getGroupTotal` from this PR entirely. The feature has no current consumer (#5448 loads whole sections), and the inline getter was invalidating the useMemo on every render. Removed the prop, changeset, tests, locale string, and doc entry. Will revisit as a standalone PR if/when a consumer needs it with a stable reference.

ernestt (commented):
Addressed all three points in efc5cf4:

1. **Header width preserved** — `fit-content` moved to a separate style applied only for the default heading; custom `renderGroupHeader` now gets full column width.
2. **Overlay writes guarded** — `applyRowSelectionStyle` compares current values before touching the DOM; unchanged rows are no-ops.
3. **`getGroupTotal` dropped** — removed the prop, changeset, tests, locale string, and doc entry from this PR. No current consumer and the inline getter invalidated the plugin memo on every render. Will revisit in a separate PR with a stable-ref-aware API when needed.

cixzhang (approved):
Thanks — all three fixed at `e24ef4c`. The pin only applies when the plugin draws the heading, so a custom `renderGroupHeader` still loses its chevron when scrolled. Both changesets need `@ernestt`.

[Full review](https://github.com/cixzhang/astryx/wiki/Review-5454)  
[Reviewed by Robohands]

PR #5455, opened 2026-08-25
spike(Indicator): make the loading visual replaceable — a stateless 'busy' family

cixzhang:
**Spike — not for merge.** Exploratory code answering a design question: should the loading visual become a replaceable indicator, so a product can ship its own branded one? Draft on purpose; no changeset, no docs, partial test coverage.

Run `Core/Spinner/Branded Replacements → BrandedSpinners` in Storybook. Four
structurally different loading visuals — a row of bouncing dots, a pulsing
square, a reversed arc, a rotating mark — each swapped in with nothing but

```ts
defineTheme({name: 'brand', indicators: {spinner: BouncingDots}})
```

and each rendering inside a real loading Button, a busy TextInput, a busy
Switch, and a standalone `<Spinner>`. The hosts are identical between rows and
none of them knows which visual it is drawing.

## What changed

**A `busy` family, and `spinner` in the indicator registry.** The mechanism was
already there (`packages/core/src/Indicator/`); this extends it rather than
inventing anything.

**One line per host.** `<Spinner size="sm" />` becomes
`<BusyIndicator size="sm" />` — 24 sites in 20 components. `BusyIndicator` is
the resolved registry entry, so a theme's replacement reaches every one of them
at once. `<Spinner>` itself now renders the same registry entry inside, so it
follows too.

**`role="status"` moved to the hosts.** The indicator is decorative and
`aria-hidden`, like every other indicator; the control keeps `aria-busy` and the
announcement. Standalone `<Spinner>` keeps its `role="status"`, because a
spinner a product renders by itself has no host to own it.

## The family system does not survive a stateless member unchanged

This is the interesting finding. `IndicatorProps.state` was declared required
and family-typed. A busy visual has no state — it is rendered or it is not — so:

- `busy: never` makes `state: never`, and **every call site becomes
  unsatisfiable**;
- `busy: 'busy'` type-checks and makes every call site write a word that carries
  no information.

The fix is one change in one place: **a family now fixes the *domain of the
props*, not just the state enum.** `state` is derived conditionally
(`[IndicatorState<F>] extends [never] ? {state?: never} : {state: …}`), and
`size` comes from a parallel `IndicatorFamilySizeMap`. Existing families resolve
to exactly the props they had, which is why nothing downstream of them changed —
but the docblock claiming "adding a family is additive" is now only true for
families that have states, and it says so.

Size falls out of the same change: selection marks keep `sm | md`, `busy` keeps
the four sizes `Spinner` ships. Worth knowing: **every one of the 24 host call
sites passes `sm` or `md`.** `lg` and `xl` only ever appear on a standalone
spinner, so widening one family costs the control hosts nothing.

## Colour, without a `shade` prop

`shade` was four ways of saying which colour the ring paints in, which is what
`color` already says. The indicator turns `--_spinner-color` into its own
`color` and strokes in `currentColor`, so a replacement written the obvious way
honours the host's shade without knowing shades exist. Three private vars
(`--_spinner-color`, `--_spinner-track-color`, `--_spinner-track-opacity`) are
the host→indicator contract. The default indicator also keeps the existing
`spinner` theme target and public vars, so current themes still override its arc,
track, and geometry inside controls as well as on standalone `<Spinner>`.

## Does a replacement break a host's layout?

Measured in Chromium rather than guessed, since the dots row is 19×5 and our
ring is 14.5×14.5:

| host | default | widest replacement | verdict |
|---|---|---|---|
| Button | 120×32 | 120×32 | unaffected — the spinner sits in an absolutely-positioned centred overlay |
| Switch track | 40×24 | 40×24 | not resized, but the visual **paints outside the thumb** |
| TextInput | 180 wide | 175–184 | the end slot is in flow, so the typing area moves by up to 9px |

Nothing clips and nothing breaks. But no host constrains the visual either, so a
replacement with a different aspect ratio changes the row it is in. If this
ships, the indicator contract should say what box a busy visual is given.

## Accessibility, measured before and after

Real accessibility trees (CDP `Accessibility.getFullAXTree`) in Chromium:

| host | before | after |
|---|---|---|
| Button | `button [busy]` + its own status region | identical |
| Switch | `switch [busy]` + its own status region | identical |
| TextInput | `textbox [busy]` + `status "Loading"` | `textbox [busy]` |
| TextArea | `textbox [busy]` + `status "Loading"` | `textbox [busy]` |
| Thumbnail | `group` + `status "Loading"` | `group [busy]` (added here) |

Button, Switch and CheckboxInput render the spinner inside an `aria-hidden`
subtree already, so its `role="status"` has been inert in them all along — three
of the loudest hosts lose nothing.

Hosts without an existing busy surface now expose one: `DropdownMenuSubMenu`
marks its menu item busy, `CommandPaletteInput` marks its combobox busy, and
`ChatToolCalls` marks each pending/running row busy (a non-interactive row becomes
a status region). Thumbnail likewise gains `aria-busy`. Fourteen more keep their
existing `aria-busy` but lose a second spoken "Loading" — a change in verbosity,
not in whether the state is reachable.

## Test cost

Whole suite: **5 tests** asserted the spinner's own `role="status"` inside a
host (Button, TextArea, Thumbnail, CheckboxList ×2). All five are updated here
to assert the host's `aria-busy` and the indicator's presence, which is the
migration in miniature. Six more were registry bookkeeping (documenting the new
theme target and the three vars). Everything else — 8129 tests — passed
untouched.

### DIFF
--- a/apps/storybook/stories/SpinnerBranded.stories.tsx
+++ b/apps/storybook/stories/SpinnerBranded.stories.tsx
@@ -1,0 +1,288 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+/**
+ * The argument for making the busy visual replaceable, in one frame.
+ *
+ * Four replacements, each swapped in by `defineTheme({indicators: {spinner}})`
+ * and nothing else — no prop, no host change, no per-component styling. Each
+ * row renders the same three hosts as the default row above it, so what
+ * changes between rows is exactly one registry entry.
+ *
+ * They are deliberately not four recolours of our ring: a row of dots, a
+ * square, a reversed arc with a flat cap, and a rotating mark. Two of them are
+ * not circles and one of them is not square, which is the point — a
+ * replacement is not constrained to the shape we ship, and the demos are where
+ * that gets tested rather than asserted.
+ */
+
+import type {Meta, StoryObj} from '@storybook/react';
+import * as stylex from '@stylexjs/stylex';
+import {Button} from '@astryxdesign/core/Button';
+import type {IndicatorProps} from '@astryxdesign/core/Indicator';
+import {Spinner} from '@astryxdesign/core/Spinner';
+import {Switch} from '@astryxdesign/core/Switch';
+import {Text} from '@astryxdesign/core/Text';
+import {TextInput} from '@astryxdesign/core/TextInput';
+import {Theme, defineTheme} from '@astryxdesign/core/theme';
+import {HStack, VStack} from '@astryxdesign/core/Layout';
+
+const meta: Meta = {
+  title: 'Core/Spinner/Branded Replacements',
+};
+
+export default meta;
+type Story = StoryObj;
+
+// The busy family has no state to draw, so a replacement is a function of
+// `size` alone. `md` is the fallback every indicator defaults to.
+const PX: Record<string, number> = {sm: 10, md: 14, lg: 18, xl: 28};
+
+// -- 1. Bouncing dots — a ROW. Three times as wide as it is tall, which is the
+//       demo that probes whether a host's busy slot assumes a square.
+
+const bounce = stylex.keyframes({
+  '0%, 80%, 100%': {transform: 'translateY(0)', opacity: 0.4},
+  '40%': {transform: 'translateY(-40%)', opacity: 1},
+});
+
+const dotStyles = stylex.create({
+  row: {display: 'inline-flex', alignItems: 'center', gap: '2px'},
+  dot: {
+    backgroundColor: 'currentColor',
+    borderRadius: '50%',
+    animationName: bounce,
+    animationDuration: '1s',
+    animationIterationCount: 'infinite',
+    animationTimingFunction: 'ease-in-out',
+  },
+  d2: {animationDelay: '0.15s'},
+  d3: {animationDelay: '0.3s'},
+});
+
+function BouncingDots({size = 'md'}: IndicatorProps<'busy'>) {
+  const d = Math.round(PX[size] / 2.2);
+  return (
+    <span aria-hidden="true" {...stylex.props(dotStyles.row)}>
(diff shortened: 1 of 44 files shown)


cixzhang:
Cleared the `build-storybook` typecheck break — the three `--_spinner-*` entries in `Indicator.doc.mjs` were missing the required `default`, so I gave them the values the component actually falls back to (`var(--color-accent)`, `var(--color-track)`, `1`) and marked them `private`, which is what they are: a host slot sets them, a theme does not.

That leaves one real failure, and it is the spike's own demo rather than the mechanism:

```
✗ [serious] color-contrast — Branded Replacements / Branded Spinners (5 nodes)
```

Worth deciding rather than baselining. The story exists to show that a replacement can carry a brand colour, and the brand colours it picked do not meet contrast — so as written the demo teaches the thing the audit is there to catch. Changing the demo's colours costs nothing and keeps the point; a baseline entry would keep the colours and lose the point.

Not touched, since which colours the demo should use is a design call.

cixzhang:
Direction: pursue the replaceable busy-indicator architecture. Keep this as a spike while the announcement and demo gaps are resolved; it is not ready to merge merely because the direction is accepted.

cixzhang (commented):
Both measured clean — the track stays the button's colour, and the running row is a status region. One note: nothing says "Loading" any more. Approve once you undraft.

[Reviewed by Robohands]

cixzhang (commented):
Both measured clean — the track stays the button's colour, and the running row is a status region. One note: nothing says "Loading" any more. Approve once you undraft.

[Reviewed by Robohands]

cixzhang (commented):
Rebase is clean and preserves the replaceable busy-indicator direction. Main’s new status labels and Button behavior survive, and the new touch DateTimeInput uses the replacement seam. This remains a spike, not merge-ready.

[Reviewed by Robohands]

cixzhang:

PR #5458, opened 2026-08-25
docs(storybook): show how to theme a toast's chrome and its content

cixzhang:
Toast had no theming story. Two new stories in `Core/Toast` show a `defineTheme({components: {toast: …}})` override moving real pixels, in both types and both colour modes.

`ThemedToast` puts default and themed side by side (inline `Toast`, so both types stay on screen); `ThemedToastLive` is the copyable shape — the theme wraps the `ToastViewport` and toasts are fired normally.

### What the example teaches

One target, `toast`, reaches more of the toast than it looks like it should:

- **Chrome** — `base` (pill radius, wider inline padding, cream surface, shadow) and `type:error` (deep red), because `type` is the toast's visual prop.
- **Inherited properties reach the content** — the monospace body comes from `fontFamily` on `toast`. There is no `toast-body` target and none is needed.
- **Text colour looks after itself** — `Toast` renders `MediaTheme mode="auto"`, so it measures the surface it just painted. Neither rule sets a text colour, and the cream card still gets dark text while the deep red one gets light text, in both modes.

`onDark`/`onLight` are for overriding that choice, not making it — they apply only when the surface resolves to that side, which the story's doc block says so nobody writes a rule that silently matches nothing.

### Verified in Chromium

Receipt-backed Chromium captures against local Storybook at `0ca75baa1ef79ca9c966756eb789612bd72bab6f`, with both the static comparison and live `useToast()` story checked in light and dark.

| | light | dark |
|---|---:|---:|
| default info contrast | 18.8:1 | 17.9:1 |
| themed info contrast | 16.4:1 | 16.4:1 |
| themed error contrast | 13.9:1 | 13.3:1 |

| Light | Dark |
|---|---|
| Default and themed Toast examples in light mode | Default and themed Toast examples in dark mode |

The themed surface resolves to 9999px radius, 24px inline padding, and `ui-monospace` in both modes. The cream info toast resolves dark text at 16.4:1; the deep-red error toast resolves light text at 13.9:1 in light mode and 13.3:1 in dark mode.

`pnpm vitest run apps/storybook/.storybook/main.test.ts packages/core/src/Toast` — 48 passed. Storybook's current `theme-layers` and full CI run on the pushed head.

### Note for #5428

[#5428](https://github.com/facebook/astryx/pull/5428) appends its `renderToast` story to the end of this same file, so expect a trivial tail conflict. Nothing in this story depends on that PR's targets — it uses only `astryx-toast`, which is on `main` today.

### DIFF
--- a/apps/storybook/stories/Toast.stories.tsx
+++ b/apps/storybook/stories/Toast.stories.tsx
@@ -5,10 +5,12 @@ import {useState, useRef, type ReactNode} from 'react';
 import * as stylex from '@stylexjs/stylex';
 import {Toast, useToast, ToastViewport} from '@astryxdesign/core/Toast';
 import type {ToastOptions, ToastType} from '@astryxdesign/core/Toast';
+import {Theme, defineTheme, useTheme} from '@astryxdesign/core/theme';
 import {Button} from '@astryxdesign/core/Button';
 import {Link} from '@astryxdesign/core/Link';
 import {Card} from '@astryxdesign/core/Card';
 import {Stack} from '@astryxdesign/core/Stack';
+import {Heading} from '@astryxdesign/core/Text';
 import {Dialog} from '@astryxdesign/core/Dialog';
 import {Text} from '@astryxdesign/core/Text';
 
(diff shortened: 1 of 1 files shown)


cixzhang:
This is docs-only teaching existing behavior, so it falls under the standing ruling: keep it as long as it is correct. Rebase the one-file conflict and verify the story; no separate product ruling needed.

cixzhang:
Thanks, this is good. The light and dark stories demonstrate the current Toast theming contract.

[Reviewed by Robohands]

PR #5463, opened 2026-08-25
fix(core): Slider draws the focus ring for keyboard focus only

cixzhang:
## Problem

Dragging a Slider thumb with the mouse drew the 2px keyboard focus ring.

The thumb is a `div[role="slider"]`, and the track's `pointerdown` handler calls
`preventDefault()` and then focuses the closest thumb from script. Chromium
treats that script focus as focus-visible, so `:focus-visible` matched from
mouse-down onward and stayed matched for the whole drag.

Measured in Chromium against `main`, not inferred — `:focus-visible` is `true`
and the computed outline is `2px solid` at mouse-down, during the drag, and
after mouse-up.

## Solution

`:focus-visible` stays the CSS condition. It is now narrowed by the existing
`interactionModality` utility, the same way `PanelSearchInput` and `Selector`
narrow theirs: the thumb records which device moved focus to it, and the ring
style is applied only for a keyboard focus.

Three details worth the review:

- The pointer path clears the state explicitly as well as through the focus
  event. Focusing an already-focused element fires no focus event, so a thumb
  the user had tabbed to would otherwise keep its ring through the drag.
- A keypress after a mouse drag brings the ring back. This is deliberately
  different from the text-input cases, where typing after a click does not —
  a text field has a caret to show where input is going, and a thumb has
  nothing else, so a keyboard user operating a mouse-focused thumb would have
  no visible focus indicator (WCAG 2.4.7). The keydown path asks the utility
  rather than assuming, so a modifier chord (⌘C, ⌃R) does not count.
- The state is per-thumb, so in range mode only the thumb that actually has
  keyboard focus rings.

No API change, no new DOM attribute, no visual change to keyboard focus, no
theme change. The ring's appearance still comes entirely from the
`--focus-outline-*` tokens.

## Evidence

Chromium, `core-slider--default`, 3x, cropped to the thumb.

| | mouse drag | keyboard Tab |
|---|---|---|
| before (`main`) | before drag | before tab |
| after | after drag | after tab |

The two keyboard frames are byte-identical (`64d8623…`) — the keyboard path is
untouched.

Full state matrix, driven in Chromium (`data-keyboard-focus` / computed
`outline-style`):

| interaction | before | after |
|---|---|---|
| mouse down on thumb | ring | no ring |
| dragging | ring | no ring |
| mouse up | ring | no ring |
| arrow key after a drag | ring | ring |
| Tab to the thumb | ring | ring |
| mouse grabs a thumb that already had the ring | ring | no ring |
| ⌘C / ⌃C / ⌥C after a drag | ring | no ring |
| range: Tab to the second thumb | ring on it, none on its sibling | same |

## Testing

`packages/core/src/Slider` — 58 tests pass. Seven are new, and six of them fail
against `main`; the seventh (keyboard rings) passes there because `main` rings
for everything. The modifier-chord test also fails against this PR's own first
commit, which is the defect it was written for.

Tests assert the ring by deriving its class from `focusOutlineStyles`, so they
check the ring itself rather than a marker standing in for it.

One unrelated `act()` warning in `blocks value changes while focusable-disabled`
is fixed on the way past — that test focused a thumb outside `act`, and the new
focus handler made it warn twice instead of once. The suite is now warning-free.

## Adjacent, not fixed

A thumb that is `isDisabled` with a `disabledMessage` stays focusable so its
reason tooltip is keyboard-discoverable, but has never drawn a focus ring at
all. That predates this change and is left as-is here.

`interactionModality`'s comment says holding Shift must not turn an interaction
into a keyboard one, but its guard covers only ⌘/⌃/⌥ — so a bare Shift press
does flip it, and re-rings a mouse-dragged thumb. Pre-existing, shared with
PanelSearchInput and Selector, and not obviously wrong (Shift+Tab *is*
keyboard navigation), so it is left for its own change.

@cixzhang

### DIFF
--- a/.changeset/slider-keyboard-only-focus-ring.md
+++ b/.changeset/slider-keyboard-only-focus-ring.md
@@ -1,0 +1,22 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Slider: dragging the thumb with a mouse no longer draws the keyboard focus ring
+
+The thumb is a `div[role="slider"]`, and the track's `pointerdown` handler
+calls `preventDefault()` and then focuses the thumb from script. Chromium
+treats that script focus as focus-visible, so `:focus-visible` matched on
+mouse-down and every drag came with a 2px accent ring — measured in Chromium,
+not inferred.
+
+`:focus-visible` stays the CSS condition; it is now narrowed by the existing
+`interactionModality` utility, the same way PanelSearchInput and Selector
+narrow theirs. Keyboard focus rings exactly as before, and a mouse grab of a
+thumb that already had the ring drops it.
+
+One deliberate difference from the text-input cases: a keypress after a mouse
+drag brings the ring back. A text field has a caret to show where input is
+going, and a slider thumb has nothing else.
+
+@cixzhang
--- a/packages/core/src/Slider/Slider.test.tsx
+++ b/packages/core/src/Slider/Slider.test.tsx
@@ -13,7 +13,9 @@ import {useState} from 'react';
 import {describe, it, expect, vi, beforeEach} from 'vitest';
 import {render, screen, act, fireEvent, waitFor} from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
+import * as stylex from '@stylexjs/stylex';
 import {Slider} from './Slider';
+import {focusOutlineStyles} from '../utils/focusOutline.stylex';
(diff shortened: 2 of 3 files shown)


cixzhang:
Thanks, this is good. ⌘C after a drag is clean now. One thing it makes visible: holding Shift over a thumb you just dragged still brings the ring back.

[Reviewed by Robohands]

PR #5466, opened 2026-08-25
Give NumberInput steppers a dedicated icon slot

rubyycheung:
## Summary

- resolve both `NumberInput` stepper buttons through the component-owned `numberInput:stepperDown` extension key
- keep the shared `chevronDown` semantic unchanged
- provide a compact, mathematically centered Core fallback
- preserve downstream compatibility: `IconName` and the required `IconRegistry` keys do not change
- add focused fallback and override coverage plus a Core patch changeset

## Dependency

Depends on #5475 for typed namespaced icon keys. This branch is stacked on that PR so the implementation and tests typecheck together; it will collapse back to the six NumberInput-specific files once #5475 lands.

#5477 separately shipped the built theme artifacts. #5482 completes that infrastructure fix by rebuilding their Core dependency in `pr-visual`; neither workflow change is included here.

## Test plan

- `pnpm -F @astryxdesign/core build`
- `pnpm exec vitest run packages/core`
- focused Icon registry and NumberInput suites — 142 passed
- Core source and docs typechecks
- changeset validation
- Prettier on all changed files
- strict ESLint with `--max-warnings=0` on all changed source files

## Visual behavior

The fallback remains the approved compact chevron at `xsm` (12×12px). The painted 6×3px chevron is horizontally centered; the increment stepper rotates the same glyph 180°. No horizontal translation is applied.

PR #5475, opened 2026-08-25
feat(Icon): accept namespaced extension keys from the icon prop

cixzhang:
Component-scoped icons should be themeable without growing the shared `IconName` union — the ruling on [#5466](https://github.com/facebook/astryx/pull/5466). The registry already resolved namespaced keys; the surfaces a component actually writes did not accept them, so the union was the only route that kept `size`/`color`/`xstyle`. This is that route.

## What

`<Icon icon>`, `useIcon`, and `defineTheme({icons})` now take a namespaced key as well as a built-in name:

```tsx
<Icon icon="numberInput:stepperDown" size="xsm" />

defineTheme({
  name: 'brand',
  icons: {
    chevronDown: <ChevronDownIcon />,
    'numberInput:stepperDown': <CaretDownFilledIcon />,
  },
});
```

`defaultIcons` may carry a namespaced entry, so a component ships its own fallback where every other default lives.

## Why not widen `IconName`

`IconRegistry` is `Record<IconName, ReactNode>`, so each new member adds a **required** key to every downstream registry — a CLI-scaffolded theme stops compiling on a patch bump. A namespaced key costs no permanent union surface and no precedent for the 13 other components sharing `chevronDown`.

## Not any string

The new type is `` `${string}:${string}` ``, not `ExtendedIconName`. A misspelled built-in name is still a compile error — there's a `@ts-expect-error` test holding that.

## Risk

Additive: three parameter types widen, which is non-breaking. `getIconRegistry()` still returns built-in names only — it now skips namespaced keys explicitly instead of relying on none existing, which is what makes a namespaced default in `defaultIcons` safe.

## Testing

- `tsc --noEmit` clean on `@astryxdesign/core` (3 pre-existing `locales/pseudo.json` errors aside).
- **Negative check:** reverting only the prop widening makes the three new tests fail to compile — `Type '"numberInput:stepperDown"' is not assignable to type 'IconName | IconType'` at each call site. The tests are the type assertion.
- `vitest run packages/core packages/richtext` — 7,901 passed. Two failures (`SideNav` focus, `Table.perf`) are load-sensitive; both pass in isolation and neither touches this code.
- eslint clean on changed files.


cixzhang:
Closing because this solves a different problem from the one we need. Namespaced registry keys make app-owned React glyphs typed, but they neither accept URL-backed SVG assets nor tint those external documents. We should design around the URL-backed asset contract rather than land a second `Icon` extension that leaves it untouched.

PR #5480, opened 2026-08-25
fix(docsite): pin the hero with sticky rails, drop the overscroll suppression

### LINKED ISSUE #5470: docsite: contain the desktop-pinned hero layers so overscroll suppression can be removed entirely
## Summary

Step 2 of #5392. #5415 restored pull-to-refresh on mobile by bounding the hero's aurora glow below 1024px and scoping the overscroll suppression to desktop:

```css
@media (min-width: 1024px) {
}
```

That was always meant to be a stepping stone, not the destination. At ≥1024px three layers are still `position: fixed` — `heroContent` (`(site)/page.tsx`), `HeroFloatingCards.stage`, and `HeroThemeReel.backdropGlow` — so the suppression has to stay, and macOS trackpads still get no native rubber-band in Safari, Chrome or Firefox. That is the case #3032 was actually filed for.

This issue is to finish the job: contain the desktop-pinned layers so `overscroll-behavior-y: none` can be **deleted outright** rather than gated.

## Why `sticky` is the answer

A `fixed` layer is glued to the viewport and is not part of the document. When the document rubber-bands past its own bottom edge, the fixed layer does not lift with it — so it is sitting in the exposed gap. That is the bleed.

A `sticky` layer *is* in the document. It lifts with it, and **cannot paint below the document's own bottom edge**. That is structural rather than empirical, which is why it lets the rule be removed instead of narrowed.

Concretely, at max scroll today the fixed glow occupies viewport y=48–1098 — the entire viewport — so any bottom overscroll necessarily paints it.

## Prototype: the effect survives

DOM surgery on a dev build, canary banner hidden to match production, one `position: absolute; inset: 0` rail per layer as a direct child of `heroScope`, with `backdropGlow` + `stage` + `heroContent` all `position: sticky; top: var(--appshell-header-height)`:

| scroll | 0 | 100 | 200 | 400 | 760 | 1400 | 2200 | max |
| --- | --- | --- | --- | --- | --- | --- | --- | --- |
| pixels differing from `fixed` | 0 | 0 | 0 | 0 | 0 | 0 | 0 | 0 |

Document height unchanged (3540px). Verified at 1280×900, 1280×600, 1440×1400 and 1024×768. Layer geometry matches `fixed` everywhere except max scroll, where sticky releases — invisible, because the opaque `showcaseOverlay` already covers the hero there. Hero CTA hit-testing unaffected (`elementFromPoint` returns the button before and after).

So the tuned pin-and-cover effect is reproducible with `sticky`. The work is in the plumbing, not the design.

## Two traps, both paid for already

**1. The rails must go in the right place.** The pinned layers are not children of `heroScope`. Its only two children are `swipeArea` (`HeroReelProvider`'s root, 760px, holding every hero layer) and `showcaseOverlay`. A sticky layer inside `swipeArea` releases after ~48px, so each layer needs a containing block spanning hero + showcase.

Insert each rail **immediately before `showcaseOverlay`**, not at the front of `heroScope`. Inserting at the front puts the layers behind `themeFill` (opaque, still inside `swipeArea`, later in tree order) and produces a 29%-of-pixels difference at scroll 0 that looks exactly like sticky failing but is pure paint order.

**2. `display: contents` on `swipeArea` is NOT a shortcut — I tried it.** `<Theme>` already renders `display: contents`, so `swipeArea` is the only real box between the layers and `heroScope`, which makes "just make `swipeArea` contents and flip the three layers to sticky" look like a two-line fix. Measured, it is not:

- document height 3540 → **5302px** (+1762): `heroContent` becomes in-flow and adds its own height on top of `heroSpacer`, which still reserves 760px
- `heroContent` lands at viewport y=1858 at scroll 0 — below the fold, after the spacer
- glow `x` shifts 40 → **−520**: `left: 50%` + `translateX(-50%)` no longer resolves the way it did

Anyone taking this will reach for that shortcut. It needs the rails, plus `heroSpacer` bookkeeping and a centering change.

## Scope

- Three `absolute; inset: 0` rails as direct children of `heroScope`, inserted before `showcaseOverlay`; the pinned layers move into them. This means the layers currently rendered by `HeroReelCards` inside `HeroReelProvider` have to be reachable as direct children of `heroScope` while keeping reel context and `<Theme>`.
- `heroContent`, `stage`, `backdropGlow`: `fixed` → `sticky` at ≥1024px.
- Centering: `left: 50%` + `transform: translateX(-50%)` → `margin-inline: auto` inside a full-width rail (verified equivalent: x=40, w=1200 either way).
- `heroSpacer` height bookkeeping once the hero is sticky rather than fixed.
- Delete `@media (min-width: 1024px) { html { overscroll-behavior-y: none } }` from `apps/docsite/src/app/globals.css`.

## Add a comment where it will save someone

`fixed` breaks if an ancestor gains `transform` or `filter`. `sticky` breaks **additionally** if any ancestor gains `overflow: hidden` or `overflow: auto` — silently, no error, the hero just stops pinning and nobody notices until they look at the home page.

`.astryx-layout-content` (AppShell) already carries `overflow: clip`. That is safe — `clip` creates no scroll container — but it is one property *value* away from un-pinning the entire landing page. Worth a comment on that rule. (It is also why the `box-shadow` slab workaround in #5392 fails.)

## What is not yet proven

The prototype covered static scroll positions with transitions and animations disabled, in Chromium only. Untested: the theme-swap animation mid-scroll, resize while scrolled, `prefers-reduced-motion`, and **Safari** — which is the browser the rubber-band actually matters in.

Layout equivalence is proven; the *feel* of the tuned effect is not. That is the careful visual pass #3032 asked for, and it is the real cost here.

## Worth sizing honestly

Desktop has no pull-to-refresh. The entire payoff is native macOS trackpad rubber-band — the mobile fix restored a **function**, this restores a **feel**. Windows and Linux have no vertical bounce, so the rule is already a no-op there. It is a refactor of a hand-tuned effect for a polish win on the site's front door: worth doing deliberately, not worth rushing.

AKnassa:
Step 2 of #5392. Fixes #5470.

## What this does

Keeps the home page hero pinned exactly as it is today, but without the CSS rule that switched off the browser's native overscroll. On a Mac trackpad the page rubber-bands again at the top and bottom, like every other page on the web. Nothing changes visually.

## Why

The hero's three pinned layers (the aurora glow, the floating cards and the headline block) were `position: fixed`. A fixed layer is glued to the screen rather than to the page, so when the page rubber-bands past its own end the layer sits in the gap that opens up. #3032 hid that by turning overscroll off for the whole site, and #5415 narrowed the rule to desktop widths to give mobile its pull-to-refresh back. This finishes the job: the layers are now `position: sticky` inside full-height rails, so they lift with the page and physically cannot paint past its end. That is structural, which is why the rule can be deleted instead of narrowed further.

## What changed

- Each pinned layer rides its own rail: a `position: absolute; inset: 0` box that is a direct child of `heroScope`, spans the hero band and the showcase, and sits ahead of the showcase overlay so tree order still equals paint order (backdrop, cards, hero text, then the showcase covering them). A rail inside the 760px band alone releases its layer after about 48px, which is the trap the issue calls out.
- `backdropGlow`, the cards `stage` and `heroContent` go `fixed` -> `sticky; top: var(--appshell-header-height)` at 1024px and up. Narrow widths are untouched: the glow stays absolute and the hero text stays in flow.
- Centering moves from `left: 50%` + `translateX(-50%)` to auto margins in the full-width rail (on a sticky box `left` is an inset, not an offset). The 1200px box is capped to the rail's width rather than `100vw`: with a classic scrollbar `100vw` is wider than the rail, which zeroes the auto margins and shoves the box left.
- `HeroReelProvider` renders no DOM any more. Its hover / focus / touch surface is the new `HeroReelSwipeArea`, which `page.tsx` uses as the hero text's rail, so the band and its gutters still pause the reel and the phone collage still swipes. `HeroReelCards` is split into `HeroReelBackdrop` (fills + glow) and `HeroReelCards` (stage).
- The `@media (min-width: 1024px) { html { overscroll-behavior-y: none } }` block is deleted from `globals.css`.
- Sticky also needs AppShell's main area to stay a non-scroll container (`overflow: clip` in `LayoutContent`; `hidden` or `auto` there would silently un-pin the landing page). That is recorded on the hero's rail style and enforced by the test below; core is untouched.
- New `home-hero-overscroll.test.ts`. The docsite suite is node-only with StyleX untransformed, so these are source invariants in the shape of `component-preview-theme.test.ts`: `globals.css` never sets `overscroll-behavior` on `html`, `:root`, `body` or `*` away from `auto`, at any width (a `contain` on a nested scroller is fine; the guard has its own fixtures for `@media` and `@supports` wrappers, `:is()` / `:where()` and upper-case selectors); none of the three layers is ever `fixed` in any media arm; the rails are bounded by `heroScope` and precede the showcase; the boxes are sized against the rail, not `100vw`; and the AppShell / LayoutContent links that keep the main area a non-scroll container (`overflow: clip`, `isScrollable={isFill}`, the landing layout's `height="auto"`) hold. It is red on `main` today, and single-line mutations (a `fixed` behind a trailing comment, double quotes, a computed `[BREAKPOINT]` key, an `overflowY` longhand, `isScrollable` unconditional, `height="fill"`, the width back to `100vw`, `heroScope` static, a rail moved after the showcase, the old `@media` root rule, `:root { overscroll-behavior: contain }`, an upper-case `HTML` rule, a minified root rule, a root rule inside `@supports`) each turn it red again.

## How to see it

Nothing should look different; that is the point. Run the docsite with the canary banner off so the header is the production 48px (`NEXT_PUBLIC_DOCS_TARGET=latest pnpm -F @astryxdesign/docsite dev`), open the home page on a Mac and rubber-band at the bottom: the page bounces and the footer stays clean. The Vercel preview on this PR shows the same page.

Measured against an untouched `main` checkout with Playwright + pixelmatch, both served with the banner hidden, `prefers-reduced-motion: reduce` so the reel holds its first slide, and transitions frozen, at ten scroll offsets per viewport (0, 100, 200, 400, 712, 760, 800, 1400, 2200, max):

| viewport | pixels differing from `main` | `overscroll-behavior-y` | fixed layers reaching the viewport bottom at page end |
| --- | --- | --- | --- |
| 1280 x 900 | 0 | `none` -> `auto` | 2 -> 0 |
| 1280 x 600 | 0 | `none` -> `auto` | 3 -> 0 |
| 1440 x 1400 | 0 | `none` -> `auto` | 0 -> 0 |
| 1024 x 768 | 0 | `none` -> `auto` | 2 -> 0 |
| 900 x 1200 | 0 | `auto` -> `auto` | 0 -> 0 |
| 390 x 844 | 0 | `auto` -> `auto` | 0 -> 0 |

Layer geometry matches `fixed` at every offset (hero text at y=48 and 712px tall, glow and stage at x=40 and 1200px wide at 1280) until max scroll, where sticky releases behind the opaque showcase. Document height is unchanged (3540px). `elementFromPoint` on the "Get started" button still returns the button.

Also checked on both trees: hover in the gutter, on the CTA and over the pinned hero while scrolled pauses the reel, hover on the showcase does not, keyboard focus on a dot pauses and blur resumes, a phone swipe steps the reel left and right and a vertical drag does not (11 checks, identical results before and after); resize while scrolled (1280 -> 1440 -> 1100 at scroll 400) and a theme swap mid-scroll with transitions running, 0 pixels differing; and a forced 15px classic scrollbar, where the old `fixed` box sat at x=-7.5 overflowing both sides and the new one sits centered inside the rail.

`pnpm -F @astryxdesign/docsite test` (29 files, 434 tests), `tsc --noEmit`, strict eslint and prettier are green.

## Not verified here

Safari and WebKit, and a real rubber-band gesture, which cannot be driven headlessly. That is the careful visual pass the issue budgets for.

## Related

- #5431 takes the other route the issue mentions (a zero-height sticky pin with absolute children) and has been conflicting since #5415 landed; this follows the rails design written up in #5470 instead.
- #5467 (merged) corrected the `backdropGlow` comment's causal chain; this branch carries that wording and updates its tail for the deleted rule.

### DIFF
--- a/apps/docsite/src/__tests__/home-hero-overscroll.test.ts
+++ b/apps/docsite/src/__tests__/home-hero-overscroll.test.ts
@@ -115,24 +115,202 @@ function allValues(block: string, property: string): string[] {
   });
 }
 
-describe('docsite globals.css never suppresses overscroll (#5392, #5470)', () => {
-  it('has no overscroll-behavior declaration other than auto, at any width', () => {
-    const css = stripComments(read(path.join(APP, 'globals.css')));
-    const offenders: string[] = [];
-    for (const match of css.matchAll(
-      /overscroll-behavior(?:-x|-y|-block|-inline)?\s*:\s*([^;}]+)/gi,
-    )) {
-      const value = match[1].trim();
-      if (value !== 'auto') {
-        // Name the enclosing at-rule so a re-scoped rule is still reported.
-        const before = css.slice(0, match.index);
-        const atRule = /@media[^{]*\{(?:[^{}]|\{[^{}]*\})*$/.exec(before)?.[0];
-        offenders.push(
-          `${atRule ? atRule.split('{')[0].trim() + ' > ' : ''}overscroll-behavior: ${value}`,
-        );
-      }
+/**
+ * The open `{` preludes (at-rules and selectors, outermost first) enclosing
+ * `index`. Enough of a CSS parser for a source invariant: the input is
+ * comment-stripped and holds no braces inside strings.
+ */
+function enclosingPreludes(css: string, index: number): string[] {
+  const stack: string[] = [];
+  let prelude = '';
+  for (let i = 0; i < index; i++) {
+    const char = css[i];
+    if (char === '{') {
+      stack.push(prelude.trim());
+      prelude = '';
+    } else if (char === '}') {
+      stack.pop();
+      prelude = '';
+    } else if (char === ';') {
+      prelude = '';
+    } else {
+      prelude += char;
     }
-    expect(offenders).toEqual([]);
+  }
+  return stack;
+}
+
+/** Split `text` on `separator` characters that sit outside parentheses. */
+function splitTopLevel(text: string, separator: RegExp): string[] {
+  const parts: string[] = [];
+  let depth = 0;
+  let current = '';
+  for (const char of text) {
+    if (char === '(') {
+      depth++;
+    } else if (char === ')') {
+      depth--;
+    }
+    if (depth === 0 && separator.test(char)) {
+      if (current) {
+        parts.push(current);
+      }
+      current = '';
+    } else {
+      current += char;
+    }
+  }
+  if (current) {
+    parts.push(current);
+  }
+  return parts;
+}
+
+/**
+ * Whether a selector list targets the page itself. `overscroll-behavior`
+ * reaches the viewport (pull-to-refresh, the trackpad rubber-band) only from
+ * the root element: `html` / `:root`, or `body`, whose value browsers
+ * propagate when the root's is `auto`; `*` covers both. `:is()` and
+ * `:where()` are looked through.
+ */
+function isPageWide(selectorList: string): boolean {
+  return splitTopLevel(selectorList, /,/).some(selector => {
+    const compounds = splitTopLevel(selector.trim(), /[\s>+~]/);
+    const subject = compounds[compounds.length - 1] ?? '';
+    const wrapped = /^:(?:is|where)\((.*)\)$/i.exec(subject);
+    if (wrapped) {
+      return isPageWide(wrapped[1]);
+    }
+    return /^(?:html|body|:root|\*)(?![\w-])/i.test(subject);
+  });
+}
+
+/**
+ * Every non-`auto` `overscroll-behavior` declaration in `css` whose selector
+ * is page-wide, labelled with its enclosing at-rules and selector so a
+ * re-scoped rule is still reported. Only declarations count (the property
+ * inside an `@supports` prelude is not one). `contain` on a nested scroller
+ * only stops that box's scroll from chaining outward, so it is legitimate and
+ * not reported.
+ */
+function pageWideOverscrollSuppressions(css: string): string[] {
+  const offenders: string[] = [];
+  for (const match of css.matchAll(
+    /(?<=[{;]\s*)(overscroll-behavior(?:-x|-y|-block|-inline)?)\s*:\s*([^;{}]+)/gi,
+  )) {
+    const property = match[1].toLowerCase();
+    const value = match[2].trim();
+    if (value === 'auto') {
+      continue;
+    }
+    const preludes = enclosingPreludes(css, match.index);
+    const atRules = preludes.filter(p => p.startsWith('@'));
+    const selectors = preludes.filter(p => !p.startsWith('@'));
+    const selector = selectors[selectors.length - 1] ?? '';
+    if (!isPageWide(selector)) {
+      continue;
+    }
+    offenders.push(
+      `${[...atRules, selector].join(' > ')} { ${property}: ${value} }`,
+    );
+  }
+  return offenders;
+}
+
+describe('docsite globals.css never suppresses overscroll (#5392, #5470)', () => {
+  it('never sets overscroll-behavior on html, :root, body or * away from auto, at any width', () => {
+    const css = stripComments(read(path.join(APP, 'globals.css')));
+    expect(pageWideOverscrollSuppressions(css)).toEqual([]);
+  });
+});
+
+/**
+ * The guard's own contract, on fixtures: it must keep catching the root-level
+ * rule this change deletes, inside any at-rule, and it must not reject
+ * scroll-chaining control on a nested scroller.
+ */
+describe('the globals.css overscroll guard', () => {
+  it('reports a suppression on html inside a media query, naming the at-rule', () => {
+    const css = `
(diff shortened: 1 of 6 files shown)


AKnassa:
Done, thanks. Rebased onto current main and resolved HeroThemeReel.tsx keeping #5467's causal chain, with its tail updated for the deleted rule. Dropped the LayoutContent change; the clip note now sits on the hero's rail style and the test still enforces it in core. The globals.css guard now only reports overscroll-behavior on html, :root, body or * (looking through :is() and :where()), with its own fixtures, so a contain on a nested scroller passes.

imdreamrunner (approved):

PR #5482, opened 2026-08-25
ci: build Core before inspecting downloaded themes

rubyycheung:
## Summary

- rebuild `@astryxdesign/core` in `pr-visual` after dependency installation
- let the downloaded theme artifacts resolve their `@astryxdesign/core/dist/*` imports
- preserve #5477's downloaded-theme optimization without repeating the full workspace build

## Why

#5477 began shipping built theme artifacts to `pr-visual`, but those artifacts import Core's consumer-facing `dist` files. The downstream job starts from a fresh checkout, installs dependencies without building packages, and therefore still fails before taking a screenshot:

```text
ERR_MODULE_NOT_FOUND: packages/core/dist/theme/index.js
imported from packages/themes/butter/dist/source.mjs
```

This currently affects #5466 and #5475.

## Test plan

- removed `packages/core/dist` and reproduced the exact import failure from `packages/themes/butter/dist/source.mjs`
- ran `pnpm -F @astryxdesign/core build`
- imported the same Butter theme build successfully afterward
- `pnpm exec prettier --check .github/workflows/ci.yml`
- parsed the workflow and verified the Core build step runs after install and before the visual gate

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -582,6 +582,12 @@ jobs:
       - name: Install Playwright
         run: pnpm install --frozen-lockfile && npx playwright install chromium
 
+      # The downloaded theme builds import Core's consumer-facing dist files.
+      # Rebuild only Core here so those imports resolve without repeating the
+      # full workspace build that produced the theme artifact.
+      - name: Build Core dependency for theme inspection
+        run: pnpm -F @astryxdesign/core build
+
       - name: Fetch visual baseline from gh-pages
         run: |
           set -eu


nynexman4464:
It looks like #5481 is the better fix, and it has now merged.

Both PRs address the same missing Core `dist` dependency. #5481 reuses the Core build that `build-storybook` already produced and ships it alongside the built themes; its armed `pr-visual` run completed successfully. This PR would instead rebuild Core inside every `pr-visual` job, adding work without improving the resulting artifact.

I think this PR can close as superseded by #5481.

cixzhang (approved):
Thanks — this is the half #5477 missed. I reproduced it: with `core/dist` removed the theme import fails exactly as CI shows, and building Core alone fixes it.

[Reviewed by Robohands]

PR #5484, opened 2026-08-25
fix(Spinner): stop a narrow flex host compressing the spinner's box

freddymeta:
## Summary

Split out of [#5214](https://github.com/facebook/astryx/pull/5214) at review request — it is independent of the theming work and lands on its own merits.

A flex host narrower than the spinner compressed the spinner's box while the ring kept drawing at the size its own attributes ask for, and the box then clipped the ring it exists to contain. Nothing reported it, because a sliced ring still spins.

**The cause is `overflow: hidden` on the box**, carried over from the canvas ring this component no longer draws. A flex item whose overflow is not `visible` has an [automatic minimum size of zero](https://www.w3.org/TR/css-flexbox-1/#min-size-auto), so the clip was what left the box with no floor under it — and then cut the ring off at the edge it had shrunk to. The clip itself protects nothing: the painted circle is inscribed in the box at every size and shade, so hiding and showing the overflow render byte-identical pixels.

So the fix is a deletion plus the invariant:

- **`overflow: hidden` is gone.** With the overflow visible the automatic minimum size applies again, which floors the box at the ring's own size. This alone fixes every case below.
- **`flex-shrink: 0` stays.** It states the invariant directly rather than leaning on the automatic minimum size, which a host takes away again the moment it sets `min-width: 0` on its items — a common enough flex idiom. A spinner that does not fit overflows its host visibly instead of being silently sliced.

## What reproduces it — measured in Chromium 149

Two builds of the same harness, `origin/main` and this branch. `box` is the `role="status"` span; `ring` is the svg's own width; `clipped` is how much of the ring the box cut off.

| case | box before | box after | ring | clipped before | clipped after |
|---|---|---|---|---|---|
| `md` beside a label in a 140px row | **16 × 20** | 20 × 20 | 20 | **4px** | 0 |
| `xl` beside a label in a 160px row | **34 × 36** | 36 × 36 | 36 | **2px** | 0 |
| `lg` beside a `flex: 1 0 100px` sibling | **12 × 24** | 24 × 24 | 24 | **12px** | 0 |
| `xl` in a 16px flex host | **16 × 36** | 36 × 36 | 36 | **20px** | 0 |
| `xl` in a block host (control) | 36 × 36 | 36 × 36 | 36 | 0 | 0 |

The first three are ordinary layouts — a spinner beside a label, a spinner beside a sibling that will not give up its width — which is the part [#5214](https://github.com/facebook/astryx/pull/5214)'s description got wrong and this one corrects. **The internal hosts do not reproduce it and never did**: a Switch thumb is 14px around a 14px spinner box, so there is nothing to compress. `Switch` `sm`/`md` (on and off), `CheckboxInput` `sm`/`md` and `Button` `sm`/`md`/`lg` all measure a 14 × 14 box around a 14px ring with 0 clipped, on `main` and on this branch alike.

### Nothing else moves

Every size × every shade × the labelled case × the inline-in-text case, at four frozen rotation angles: **72 of 72 element screenshots byte-identical** between a build of `origin/main` and a build of this branch, and the box geometry and position identical to the hundredth of a pixel — including the inline case, where the box is baseline-aligned and removing a scroll container could plausibly have shifted it.

That is also the evidence that the clip was dead: with the rotation frozen at 0°, 22.5°, 45° and 67.5°, `overflow: hidden` and `overflow: visible` render the same pixels in all 68 comparisons.

## The story

`Core/Spinner → NarrowFlexHost` renders the three reproducing hosts, so the browser check owns the red→green rather than the PR description owning it. The hosts are raw flex containers on purpose: what is under test is what an arbitrary consumer's layout does to the box.

## Test plan

- `pnpm build`, `pnpm lint:strict` (0 errors, 80 warnings — the same 80 `origin/main` reports), `pnpm -F @astryxdesign/core typecheck`, `pnpm exec vitest run packages/core/src/Spinner` (33 passed).
- No unit test: the property is only observable through layout, and a computed-style assertion in jsdom is the test slop the repo rules out. The story above is the regression, and the numbers here are the evidence.


cixzhang (changes requested):
Thanks — `flexShrink: 0` is the right invariant for Spinner's own box.

The browser-only bug still has no shipped regression: the unit suite cannot observe layout, and the visual gate has no narrow-host story. Could you add that story so the browser check owns the red→green behavior?


[Reviewed by Robohands]

freddymeta:
Story added — `Core/Spinner → NarrowFlexHost`, three hosts that reproduce the bug, so the browser check owns the red→green instead of the description owning it.

While building it I went back at the premise, and the fix moved: **the cause is the `overflow: hidden` on the box**, left over from the canvas ring. A flex item whose overflow is not `visible` has an automatic minimum size of zero, so the clip is what removed the floor under the box — and then cut the ring off at the edge it had shrunk to. Deleting it fixes every case on its own. `flex-shrink: 0` stays, because the automatic minimum size goes away again the moment a host sets `min-width: 0` on its items.

The clip protects nothing: with the rotation frozen at four angles, `overflow: hidden` and `overflow: visible` render byte-identical pixels in 68 of 68 comparisons across every size, every shade and the labelled case.

Two things in the old description were wrong and are corrected:

- **It is not only a host narrower than the spinner outright.** Ordinary rows reach it — an `md` spinner beside a label in a 140px row rendered a 16px box around a 20px ring; an `lg` spinner beside a `flex: 1 0 100px` sibling lost half its ring.
- **The internal hosts genuinely never reproduced it**, and I have now measured `Button` alongside `Switch` and `CheckboxInput`: 14 × 14 box, 14px ring, 0 clipped, on `main` and here alike.

Nothing else moves: every size × shade × the labelled case × an inline-in-text case, at four frozen rotation angles — 72 of 72 element screenshots byte-identical between a build of `origin/main` and a build of this branch, with box geometry and position identical to the hundredth of a pixel (including the inline case, where removing a scroll container could plausibly have shifted the baseline).

cixzhang (approved):
Thanks — the narrow-host story closes my prior ask, and the root-cause correction holds. This is ready to approve.

[Reviewed by Robohands]

freddymeta:
/accept-visual 33046089612/1 These frames add the requested narrow-flex-host regression story and show the intentional non-shrinking Spinner box; existing Spinner visuals are unchanged.

cixzhang:
I merged current main and resolved the two mechanical Spinner overlaps by keeping both the new theming stories/vars and this PR's narrow-host story/flex invariant. Pushed `dab42073`; CI is running.

cixzhang:
The main refresh left the narrow-host story nested inside the themed-hairline story, which broke Storybook parsing. I restored the missing close, formatted both merged Spinner files, and pushed `9a197cc6`; CI is rerunning.

.changeset/spinner-box-flex-shrink.md
```diff
@@ -0,0 +1,11 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Spinner: a narrow flex host no longer compresses the box and clips the ring (#5484)
+
+The spinner's box carried `overflow: hidden` from the canvas ring it no longer draws. It clipped nothing — the painted circle is inscribed in the box, so hiding or showing the overflow renders the same pixels at every size and shade — but a flex item whose overflow is not `visible` has an automatic minimum size of zero. That left the box with no floor: a flex host narrower than the spinner compressed it while the ring kept drawing at the size its own attributes ask for, and the clip then cut the ring off at the box edge, silently, because a sliced ring still spins.
```
cixzhang, line 7: Not quite true since [#5214](https://github.com/facebook/astryx/pull/5214): the themed hairline ring does change, for the better.

cixzhang (approved):
Thanks — re-approving at the new head; the main merge dismissed my earlier approve. Rechecked in Chromium against the themed ring: the fix holds and the hairline story improves.

[Reviewed by Robohands]

freddymeta:

rubyycheung (approved):
The fix makes sense visually and technically: Spinner preserves its documented size in constrained flex layouts instead of rendering a clipped ring. The narrow-host story demonstrates the regression and the corrected behavior clearly.

freddymeta:
/accept-visual 33343939641/1 110 added, 0 removed — new baselines for stories that had none, and no existing frame moved. The report covers the current head (02d99d54, a main merge with no source change), and `Stable visual regression` is green on it.

freddymeta:
@cixzhang @rubyycheung — this one needs a maintainer to run the acceptance; I cannot, and I should have checked that before trying rather than posting a refused command twice.

The gate resolves `user.permissions.maintain` and requires `maintain` or `admin` (`.github/scripts/visual-gate/authorization.mjs`). My account is `write` with `maintain: false`, so every `/accept-visual` from me is refused regardless of what the frames show — it does not distinguish added frames from changed ones.

Everything else is done: three approvals, 23 checks green, `Stable visual regression` green on the current head, and the report covers that head exactly (`02d99d54`, a main merge with no source change) at **110 added · 0 removed** — new baselines for stories that had none, with no existing frame moved.

The exact command from the report:

```
/accept-visual 33343939641/1 110 added, 0 removed — new baselines for stories that had none; no existing frame moved.
```

PR #5487, opened 2026-08-25
fix(core): restore the SelectableCard/Thumbnail/ClickableCard hover overlay

### LINKED ISSUE #5442: [Bug] SelectableCard/Thumbnail/ClickableCard lose their hover overlay at 0.4.6 — #5247's :where() guard drops the StyleX specificity boost 7→3

HelloOjasMutreja:
Fixes #5442.

Thanks to @brightonvino for the exceptionally well-diagnosed report — the exact `getCompoundPseudoPriority()` root cause, the byte-for-byte `astryx.css` diff between 0.4.5/0.4.6/0.4.7, and both fix options were already worked out in the issue. This implements the "cheap" fix it recommends.

## Root cause

#5247 added a zero-specificity `:where(:not(:disabled,[aria-disabled="true"]))` guard to every self-`:hover` selector, for WCAG 1.4.1. On `SelectableCard`, `Thumbnail`, and `ClickableCard`, the `:hover` key also targets `::after`. Combining a hover pseudo-class with a pseudo-element hits a `@stylexjs/babel-plugin@0.19.0` tokenizer bug: `getCompoundPseudoPriority()`'s regex can't match the nested parens in `:where(:not(...))`, bails to a wrong priority default, and the rule's specificity boost silently dropped from 7 `:not(#\#)`s to 3. The resting `background-color: transparent` rule (still at 7, untouched) then out-specifies the hover rule in the same layer, so the overlay never paints.

## Fix

The guard is redundant on these three components specifically: each already applies the whole `hoverOnPointer` class conditionally in JS —

- `SelectableCard`: `!isDisabled && styles.hoverOnPointer`
- `ClickableCard`: `!isDisabled && styles.hoverOnPointer`
- `Thumbnail`: `isInteractive && styles.hoverOnPointer`, where `isInteractive = onClick != null && !isDisabled && !isLoading`

A disabled/non-interactive card never carries the class that would need CSS-level guarding, so reverting the key to bare `:hover::after` loses none of #5247's protection here — it just stops hitting the StyleX bug.

`@astryx/no-hover-on-disabled` now knows to leave `:hover` + pseudo-element combinations alone, so `eslint --fix` doesn't just re-add the guard and re-break this on the next run. Documented as a real, tracked coverage gap in the rule's own header (not a free pass) — a future `:hover::after`/`:hover::before` key that isn't already JS-gated the way these three are needs manual verification, since the rule can't currently catch it in this shape.

## Verification

Rebuilt `packages/core` and diffed the generated `astryx.css` directly: the hover rule now reads
```css
.xi14tyy.xi14tyy:hover:not(#\#):not(#\#):not(#\#):not(#\#):not(#\#):not(#\#):not(#\#)::after{background-color:var(--color-overlay-hover)}
```
— 7 boosts, byte-for-byte identical to the issue's cited 0.4.5 (working) output, matching the resting rule's own 7 boosts again.

**Before** (0.4.6/0.4.7, hovering "Basic" — flat, no overlay):
before

**After** (hovering "Basic" — the overlay tints again):
after

Confirmed programmatically too: `getComputedStyle(el, '::after').backgroundColor` reads `rgba(0, 0, 0, 0)` on hover before the fix, `rgba(0, 0, 0, 0.05)` after, on the same Storybook story with transitions disabled (per the issue's own measurement note, to avoid catching a mid-fade frame).

`no-hover-on-disabled.test.mjs` updated: the two cases asserting `:hover::after`/`:hover::before` get autofixed moved from `invalid` to `valid`, since that's exactly the behavior being turned off, with a comment explaining why and linking back to this issue.

Full `SelectableCard`/`Thumbnail`/`ClickableCard`/`eslint-plugin-astryx` suite: 570/570 passing. Typecheck and lint clean.

### DIFF
--- a/.changeset/hover-overlay-specificity.md
+++ b/.changeset/hover-overlay-specificity.md
@@ -8,7 +8,7 @@
 
 The guard was redundant on these three anyway: each already applies the whole `hoverOnPointer` class conditionally (`!isDisabled && styles.hoverOnPointer`, or `isInteractive && ...` for Thumbnail, where `isInteractive` already excludes `isDisabled`), so a disabled/non-interactive card never carries the class that would need guarding. Reverted the three keys to bare `:hover::after`, restoring the pre-regression 7-boost CSS output exactly.
 
-`@astryx/no-hover-on-disabled` now knows to leave `:hover` + pseudo-element combinations alone — guarding them would just reintroduce this exact break until the StyleX tokenizer is fixed upstream. This is a real, documented coverage gap, not a free pass: a new `:hover::after`/`:hover::before` key that doesn't already exclude the disabled case some other way (as these three do, in JS) needs manual verification, since the lint rule can no longer catch it in this shape.
+`@astryx/no-hover-on-disabled` now knows not to autofix a `:hover` + pseudo-element key, anywhere — autofixing one would reintroduce this exact break until the StyleX tokenizer is fixed upstream. It still reports the key everywhere except these three files, so a new unguarded `:hover::after`/`:hover::before` still needs manual verification (does the component already exclude the disabled case in JS, the way these three do?) before it's added to the rule's narrow exemption list.
 
 Fixes #5442.
 
--- a/.changeset/hover-overlay-specificity.md
+++ b/.changeset/hover-overlay-specificity.md
@@ -1,0 +1,15 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] `SelectableCard`, `Thumbnail`, and `ClickableCard`'s hover overlay paints again — it silently stopped rendering at 0.4.6.
+
+`#5247` added a zero-specificity `:where(:not(:disabled,[aria-disabled="true"]))` guard to every self-`:hover` selector for WCAG 1.4.1. On these three components the hover key also targets `::after`, and combining a hover pseudo-class with a pseudo-element hits a `@stylexjs/babel-plugin@0.19.0` tokenizer bug: `getCompoundPseudoPriority()` can't match the nested parens in `:where(:not(...))`, bails to a wrong priority default, and the rule's specificity boost silently dropped from 7 `:not(#\#)`s to 3. The resting `background-color: transparent` rule (still at 7) then out-specified the hover rule in the same layer, so the overlay never painted — verified by diffing the generated `astryx.css` byte-for-byte against 0.4.5.
+
+The guard was redundant on these three anyway: each already applies the whole `hoverOnPointer` class conditionally (`!isDisabled && styles.hoverOnPointer`, or `isInteractive && ...` for Thumbnail, where `isInteractive` already excludes `isDisabled`), so a disabled/non-interactive card never carries the class that would need guarding. Reverted the three keys to bare `:hover::after`, restoring the pre-regression 7-boost CSS output exactly.
+
+`@astryx/no-hover-on-disabled` now knows to leave `:hover` + pseudo-element combinations alone — guarding them would just reintroduce this exact break until the StyleX tokenizer is fixed upstream. This is a real, documented coverage gap, not a free pass: a new `:hover::after`/`:hover::before` key that doesn't already exclude the disabled case some other way (as these three do, in JS) needs manual verification, since the lint rule can no longer catch it in this shape.
+
+Fixes #5442.
+
+@HelloOjasMutreja
--- a/internal/eslint-plugin-astryx/no-hover-on-disabled.js
+++ b/internal/eslint-plugin-astryx/no-hover-on-disabled.js
@@ -40,24 +40,35 @@
  * alone.
  *
  * GAP, known and tracked: `:hover` combined with a pseudo-ELEMENT —
- * `:hover::after`, `:hover::before` — is left unguarded too, and is NOT
- * autofixed if already unguarded. `@stylexjs/babel-plugin@0.19.0`'s
- * getCompoundPseudoPriority() cannot match the nested parens in
- * `:where(:not(:disabled,[aria-disabled="true"]))` and bails to a wrong
- * priority default, which silently drops the rule's specificity boost and
- * can let an unrelated resting rule win instead — see
+ * `:hover::after`, `:hover::before` — is NOT autofixed, anywhere.
+ * `@stylexjs/babel-plugin@0.19.0`'s getCompoundPseudoPriority() cannot match
+ * the nested parens in `:where(:not(:disabled,[aria-disabled="true"]))` and
+ * bails to a wrong priority default, which silently drops the rule's
+ * specificity boost and can let an unrelated resting rule win instead — see
  * facebook/astryx#5442, where this broke SelectableCard/Thumbnail/
- * ClickableCard's hover overlay. Guarding a `:hover::after` key here would
- * reintroduce that regression, so the exemption stays until the upstream
- * tokenizer is fixed.
+ * ClickableCard's hover overlay. Autofixing a `:hover::after` key would
+ * reintroduce that exact regression, so the fixer stays off for this shape
+ * until the upstream tokenizer is fixed. The key is still REPORTED, though:
+ * an unguarded `:hover::after`/`:hover::before` is real, ordinary lint
+ * output everywhere except the three files below, same as any other
+ * unguarded hover key — only the autofix is withheld.
  *
- * This is a real coverage hole, not a free pass: a `:hover::after`/
- * `:hover::before` key on a component that does NOT already exclude the
- * disabled case some other way (the three components above all gate the
- * whole class in JS — `!isDisabled && styles.hoverOnPointer`) can still ship
- * a hover-while-disabled bug with nothing here to catch it. Verify that
- * gating by hand for anything new in this shape.
+ * EXEMPT_FILES narrows the reported-but-not-flagged case to exactly the
+ * three files where this PR verified, by hand, that JS-level gating already
+ * makes the guard redundant (`!isDisabled && styles.hoverOnPointer`, or
+ * equivalent). Nothing else gets a free pass: a new `:hover::after`/
+ * `:hover::before` key anywhere else is reported (without an autofix) and
+ * needs the same by-hand verification before it's added to this list.
  */
+const EXEMPT_FILES = [
+  /[/\\]SelectableCard[/\\]SelectableCard\.tsx$/,
+  /[/\\]Thumbnail[/\\]Thumbnail\.tsx$/,
+  /[/\\]ClickableCard[/\\]ClickableCard\.tsx$/,
+];
+
+function isExemptFile(filename) {
+  return !!filename && EXEMPT_FILES.some(pattern => pattern.test(filename));
+}
 
 /** Zero-specificity guard appended to a self-hover selector. */
 const GUARD = ':where(:not(:disabled,[aria-disabled="true"]))';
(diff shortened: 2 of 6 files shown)


cixzhang (changes requested):
Disabled components can now regain a hover affordance without lint catching it. The new pseudo-element early return makes `:hover::before` and `:hover::after` valid even when a component does not gate the class; the added test now accepts exactly that unsafe selector.

Please keep this selector shape reportable (the rule can omit its autofix) and narrowly exempt only the three component sites whose runtime gating this PR proves. The visual fix itself works: all three overlays moved from transparent to the neutral 5% tint at this head.

[Reviewed by Robohands]

HelloOjasMutreja:
Agreed, this was too broad. Narrowed it: the rule now still reports an unguarded `:hover` + pseudo-element key everywhere (a new `unguardedHoverPseudoElement` messageId), just without an autofix, since autofixing is what hits the tokenizer bug. Only the three files this PR verified by hand (SelectableCard, Thumbnail, ClickableCard) are exempted from even the report, via an explicit filename allowlist rather than a blanket key-shape skip.

Verified the three real component files stay lint-clean, and grepped `packages/core/src` for `':hover::'` to confirm nothing else in the codebase currently uses this shape, so this closes the gap without introducing any new lint noise today.

cixzhang (changes requested):
Thanks for restoring the three overlays. A future disabled-state hover in these files can still ship without lint catching it because the exemption is file-wide, not limited to the three verified sites. I added a second unguarded `:hover::before` under the SelectableCard filename and the rule returned 0 diagnostics. Could the exemption identify only the existing verified `hoverOnPointer` site while continuing to report every unrelated selector in those files?

[Reviewed by Robohands]

HelloOjasMutreja:
Scoped the exemption to the exact verified site rather than the whole file: it now checks the (file, top-level style key) pair, not just the filename. `isExemptFile` is gone; `isExemptSite(filename, getEnclosingStyleKeyName(node))` walks up to the top-level property passed directly to `stylex.create({})` and only exempts when that key is literally `hoverOnPointer` (the one this PR hand-verified is JS-gated in all three components).

A new `:hover::before`/`:hover::after` under any other key in the same three files is reported again, same as anywhere else. Added invalid test cases for exactly that: a `badge` key with an unguarded pseudo-element hover in each of the three exempt files now fails.

Fixed in 0993bd953.

HelloOjasMutreja:
The `pr-rtl` failure wasn't a real RTL bug in the overlay fix, it was an undeclared coverage gap on ClickableCard specifically (CI reported 1 measured / 1 verified N-A / 1 gap across the three touched components).

Read `ClickableCard.tsx`, its stories, and `useClickableContainer` before declaring. It follows the same clickable-card pattern SelectableCard already has a verified-N/A entry for: a symmetric `position: absolute; inset: 0` overlay pseudo-element, no physical left/right positioning or directional glyph anywhere, and native click/Enter/Space activation rather than an inline-direction key. Added the `core/ClickableCard` entry mirroring that existing reasoning. Thumbnail was already measured (its close button uses `insetInlineEnd`, a logical property D5 positional-mirror picks up), and SelectableCard already had its declaration, so this was the only remaining gap.

Rebased onto current main and verified locally: `pnpm rtl:audit -- --filter ClickableCard,SelectableCard,Thumbnail` now reports `1 measured / 2 verified N-A / 0 gap`.

PR #5489, opened 2026-08-25
feat(core): opt-in bulk-actions toolbar for useTableSelection (#5486)

### LINKED ISSUE #5486: Table selection: opt-in bulk-actions toolbar

**Feature area:** Selection & interaction

### Problem

`useTableSelection` gives per-row checkboxes and select-all, but no surface for acting on a selection. Internal tables (and the EPS table, source of this port) show a bulk-actions toolbar above the table when rows are selected: a count, action buttons (Delete, Export, etc.), and an optional "select all N matching" affordance.

### Proposal

Extend `useTableSelection` with an optional `bulkActions` config. When present, the plugin injects a toolbar into the scroll wrapper via `transformScrollWrapper.beforeTable` while rows are selected. Off by default: tables that omit `bulkActions` are byte-identical to today (zero regression).

Config sketch:
- `actions: TableBulkAction[]` where a `TableBulkAction` is `{label, onClick(selectedKeys), variant?, disabled?}`
- `renderLabel?(count, total)` (default `"N selected"`)
- `extraContent?(count)`
- `selectAllMatching?: {totalMatchingCount, isSelectAllMatching, onSelectAllMatching}`

### Notes

- Astryx `transformScrollWrapper` already exposes `beforeTable` / `afterTable` slots, so the injection point exists.
- Styled with StyleX (the EPS original used Tailwind + eps tokens).
- Start with an in-flow toolbar in the `beforeTable` slot with a fade/slide. The EPS "float over the table + sticky-pin on scroll" polish (absolute positioning, IntersectionObserver) can be a follow-up.

Mirrors the internal EPS bulk-selection toolbar.

humbertovirtudes:
## Summary

Adds an opt-in bulk-actions toolbar to `useTableSelection`. When a `bulkActions` config is supplied, a toolbar appears above the table while rows are selected, holding a count, an optional "select all N matching" link, optional extra content, and the consumer's action buttons.

Closes #5486.

## What changed

- **`useTableSelection`** gains an optional `bulkActions` config. When present, the plugin injects the toolbar into the scroll wrapper via `transformScrollWrapper.beforeTable`. **Off by default:** with no `bulkActions`, the plugin is byte-identical to before, so there is no regression for existing consumers (the `transformScrollWrapper` returns its input untouched).
- **`TableBulkActionsToolbar`** (new, StyleX): presentation-only toolbar. Stays mounted through a short exit transition so it fades out when the selection clears, then unmounts; renders nothing while inactive, so an always-injected node has no layout cost.
- **Types**: `TableBulkAction` (`{label, onClick(selectedKeys), variant?, disabled?}`) and `TableBulkActionsConfig` (`{selectedKeys, actions, renderLabel?, extraContent?, selectAllMatching?}`).

## API

```
const [selectedKeys, setSelectedKeys] = useState<Set<string>>(new Set());
const {selectionConfig} = useTableSelectionState({data, idKey: 'id', selectedKeys, setSelectedKeys});
const selection = useTableSelection({
  ...selectionConfig,
  bulkActions: {
    selectedKeys,
    actions: [
      {label: 'Export', onClick: keys => exportRows(keys)},
      {label: 'Delete', variant: 'destructive', onClick: keys => deleteRows(keys)},
    ],
  },
});
```

## Test plan

- `TableBulkActionsToolbar.test.tsx`: 9 tests (no toolbar when empty, count label, action buttons, onClick passes selected keys, renderLabel, extraContent, select-all-matching link show/hide, and the no-op-when-omitted regression).
- Full selection suite: 46 pass (37 existing + 9 new), no regression.
- `tsc --project packages/core/tsconfig.json --noEmit`: clean. `eslint` on changed files: clean. Changeset valid. `@astryxdesign/core` build: clean.
- Adds a Storybook story (`Core/TableSelection` -> `BulkActions`) and a docsite example block (`TableBulkActionsTable`, `exampleFor: useTableSelection`).

## Notes

- The toolbar renders in-flow in the `beforeTable` slot with a fade/slide. A follow-up could add a floating variant that overlays the table and pins below the header on scroll.

### DIFF
--- a/.changeset/table-selection-bulk-actions.md
+++ b/.changeset/table-selection-bulk-actions.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] Add an opt-in bulk-actions toolbar to `useTableSelection` via a new `bulkActions` config (#5486). When present, a toolbar appears above the table while rows are selected, showing a count, optional "select all N matching" link, optional extra content, and the provided action buttons (each receives the selected keys on click). Off by default: tables that omit `bulkActions` are unchanged.
+
+@humbertovirtudes
--- a/apps/storybook/stories/TableSelection.stories.tsx
+++ b/apps/storybook/stories/TableSelection.stories.tsx
@@ -292,3 +292,55 @@ export const WithStripedRows: Story = {
     );
   },
 };
+
+/**
+ * An opt-in bulk-actions toolbar appears above the table while rows are
+ * selected. Pass a `bulkActions` config to `useTableSelection` with the
+ * current `selectedKeys` and the action buttons. Each action receives the
+ * selected keys on click. Omit `bulkActions` for the default behavior with no
+ * toolbar.
+ */
+export const BulkActions: Story = {
+  render: () => {
+    const [selectedKeys, setSelectedKeys] = useState<Set<string>>(new Set());
+
+    const {selectionConfig} = useTableSelectionState<User>({
+      data: users,
+      idKey: 'id',
+      selectedKeys,
+      setSelectedKeys,
+    });
+    const selectionPlugin = useTableSelection<User>({
+      ...selectionConfig,
+      bulkActions: {
+        selectedKeys,
+        actions: [
+          {
+            label: 'Export',
+            onClick: keys => window.alert(`Export ${keys.size} rows`),
+          },
+          {
+            label: 'Delete',
+            variant: 'destructive',
+            onClick: keys => {
+              setSelectedKeys(new Set());
+              window.alert(`Deleted ${keys.size} rows`);
+            },
+          },
+        ],
+      },
+    });
+
+    return (
+      <div style={{maxWidth: 600}}>
+        <Table
+          data={users}
+          columns={columns}
+          idKey="id"
+          hasHover
+          plugins={{selection: selectionPlugin}}
+        />
+      </div>
+    );
+  },
+};
--- a/packages/cli/assets/templates/blocks/components/Table/TableBulkActionsTable.doc.mjs
+++ b/packages/cli/assets/templates/blocks/components/Table/TableBulkActionsTable.doc.mjs
@@ -1,0 +1,14 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+/** @type {import('@astryxdesign/cli/authoring').TemplateDoc} */
+export const doc = {
+  type: 'block',
+  exampleFor: 'useTableSelection',
+  name: 'Table: Bulk Actions',
+  displayName: 'Table: Bulk Actions',
+  description:
+    'Row selection with an opt-in bulk-actions toolbar that appears above the table while rows are selected, showing a count and action buttons.',
+  isReady: true,
+  aspectRatio: 16 / 9,
+  componentsUsed: ['Table', 'Button'],
+};
--- a/packages/cli/assets/templates/blocks/components/Table/TableBulkActionsTable.tsx
+++ b/packages/cli/assets/templates/blocks/components/Table/TableBulkActionsTable.tsx
@@ -1,0 +1,65 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+import {useState} from 'react';
+import {
+  Table,
+  useTableSelection,
+  useTableSelectionState,
+  proportional,
+} from '@astryxdesign/core/Table';
+import type {TableColumn} from '@astryxdesign/core/Table';
+
+interface User extends Record<string, unknown> {
+  id: string;
+  name: string;
+  email: string;
+  role: string;
+}
+
+const users: User[] = [
+  {id: '1', name: 'Alice', email: 'alice@example.com', role: 'Engineer'},
+  {id: '2', name: 'Bob', email: 'bob@example.com', role: 'Designer'},
+  {id: '3', name: 'Charlie', email: 'charlie@example.com', role: 'Manager'},
+  {id: '4', name: 'Diana', email: 'diana@example.com', role: 'Engineer'},
+  {id: '5', name: 'Eve', email: 'eve@example.com', role: 'Admin'},
+];
+
+const columns: TableColumn<User>[] = [
+  {key: 'name', header: 'Name', width: proportional(1)},
+  {key: 'email', header: 'Email', width: proportional(2)},
+  {key: 'role', header: 'Role', width: proportional(1)},
+];
+
+export default function TableBulkActionsTable() {
+  const [selectedKeys, setSelectedKeys] = useState<Set<string>>(new Set());
+
+  const {selectionConfig} = useTableSelectionState<User>({
+    data: users,
+    idKey: 'id',
+    selectedKeys,
+    setSelectedKeys,
+  });
+  const selectionPlugin = useTableSelection<User>({
+    ...selectionConfig,
+    bulkActions: {
+      selectedKeys,
+      actions: [
+        {label: 'Export', onClick: () => {}},
+        {
+          label: 'Delete',
+          variant: 'destructive',
+          onClick: () => setSelectedKeys(new Set()),
+        },
+      ],
+    },
+  });
+
+  return (
+    <Table
+      data={users}
+      columns={columns}
+      idKey="id"
+      plugins={{selection: selectionPlugin}}
+    />
+  );
+}
--- a/packages/core/src/Table/index.ts
+++ b/packages/core/src/Table/index.ts
@@ -19,6 +19,7 @@ export {TableFooter} from './TableFooter';
 export {TableContext} from './TableContext';
 export {useTableSelection} from './plugins/selection';
 export {useTableSelectionState} from './plugins/selection';
+export {TableBulkActionsToolbar} from './plugins/selection';
 export {useTableSortable} from './plugins/sortable';
 export {useTableSortableState} from './plugins/sortable';
 export {useTablePagination, paginateData} from './plugins/pagination';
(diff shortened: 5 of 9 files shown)


rubyycheung:
The bulk-actions behavior looks good. I have one visual spacing suggestion after testing it inside a padded Card: the outlined toolbar currently reads as another full-width container and feels cramped against the surrounding layout.

Could we remove the toolbar border and inset the tinted surface by 8px on each logical inline side? That offset comes from the actual selection geometry: a 20px small checkbox centered in the 36px selection column. It aligns the toolbar surface with the outer edge of the checkboxes while allowing the table rows to remain full bleed.

ernestt:
Current:
CleanShot 2026-08-31 at 14 38 52

Ideal:
CleanShot 2026-08-31 at 14 39 29

I recreated a [table template](https://astryx.atmeta.com/templates?preview=table-filter) trying to replicate what we had in www.  The bulk actions bar here is a full bleed section.  With actions on the left side, and unselection and total selected count on the right.  Mainly because the column heading also has a bulk checkbox already anyways so colocating the actions to be close by seems to make more sense.

ernestt (changes requested):
Bulk action bar visuals commented above

humbertovirtudes:
@ernestt Im thinking we can have 2 variants for bulk action bars - one that remains at the top of the table, full bleed and static as you suggested, and another that is floating - and remains visible even when the user scrolled way past the header - to avoid having to scroll back to the top in order to trigger any bulk actions.

Thoughts?

ernestt:
I'd push back on two variants, but not on the underlying need — I think persistence belongs one level up.

First, a factual note on what's on the head now: `layout: 'floating'` is `position: absolute` inside a zero-height anchor above the table, so it scrolls out of view exactly like the in-flow band does. As written it doesn't yet solve the "scrolled way past the header" case it's aimed at.

More importantly, I don't think a bulk-actions bar should be the thing that pins. If a user is 200 rows down, losing the toolbar is the smaller problem — they've also lost the column labels, the sort controls, and the select-all checkbox. A toolbar that persists while the header scrolls away gives you "3 selected · Delete" hovering over anonymous columns, which is arguably worse than scrolling back up. If the toolbar sticks, the header has to stick under it.

That primitive is already in flight, so I don't think this PR needs to invent it:

- **#6221 — `feat(Table): add useTableStickyHeader plugin`** pins the header row to the table's scroll container, and composes with `useTableStickyColumns` (the corner cell stays above both runs, tested in both install orders).
- **#6234 — `feat(Table): pin group headings while their section is on screen`** stacks on #6221 and works out the offset chain: the sticky header measures its own height and group headings pin underneath it rather than over it, with the z-index bands resolved against the sticky columns' body cells.

That offset-and-band work is exactly what a persistent toolbar needs, and it lands in the place AST-025 FR19 points at — Table adopting the capability on its existing scroll wrapper so column sizing, keyboard access, scroll shadows, sticky columns and sticky headers share one viewport. Worth noting #6221's findings for this PR either way: a table that pins columns already sits in an `overflow-x: auto` wrapper and CSS promotes `overflow-y` alongside it, so that wrapper is the sticky containing block and nothing inside it can pin to the page.

So my suggestion:

1. Keep the one full-bleed bar from this PR — single visual spec.
2. Let #6221/#6234 land the sticky machinery.
3. Then make the toolbar the top element of that same sticky stack, behind one boolean rather than a second layout.

That gives persistence to every Table consumer rather than only to selection users, and it shrinks the selection API surface instead of growing it — which also matters for the composition question raised in the other review. Happy to sketch the stack ordering against #6234's offset chain if useful.

humbertovirtudes:
> I'd push back on two variants, but not on the underlying need — I think persistence belongs one level up.
> 
> First, a factual note on what's on the head now: `layout: 'floating'` is `position: absolute` inside a zero-height anchor above the table, so it scrolls out of view exactly like the in-flow band does. As written it doesn't yet solve the "scrolled way past the header" case it's aimed at.
> 
> More importantly, I don't think a bulk-actions bar should be the thing that pins. If a user is 200 rows down, losing the toolbar is the smaller problem — they've also lost the column labels, the sort controls, and the select-all checkbox. A toolbar that persists while the header scrolls away gives you "3 selected · Delete" hovering over anonymous columns, which is arguably worse than scrolling back up. If the toolbar sticks, the header has to stick under it.
> 
> That primitive is already in flight, so I don't think this PR needs to invent it:
> 
> * **[feat(Table): add useTableStickyHeader plugin #6221](https://github.com/facebook/astryx/pull/6221) — `feat(Table): add useTableStickyHeader plugin`** pins the header row to the table's scroll container, and composes with `useTableStickyColumns` (the corner cell stays above both runs, tested in both install orders).
> * **[feat(Table): pin group headings while their section is on screen #6234](https://github.com/facebook/astryx/pull/6234) — `feat(Table): pin group headings while their section is on screen`** stacks on [feat(Table): add useTableStickyHeader plugin #6221](https://github.com/facebook/astryx/pull/6221) and works out the offset chain: the sticky header measures its own height and group headings pin underneath it rather than over it, with the z-index bands resolved against the sticky columns' body cells.
> 
> That offset-and-band work is exactly what a persistent toolbar needs, and it lands in the place AST-025 FR19 points at — Table adopting the capability on its existing scroll wrapper so column sizing, keyboard access, scroll shadows, sticky columns and sticky headers share one viewport. Worth noting #6221's findings for this PR either way: a table that pins columns already sits in an `overflow-x: auto` wrapper and CSS promotes `overflow-y` alongside it, so that wrapper is the sticky containing block and nothing inside it can pin to the page.
> 
> So my suggestion:
> 
> 1. Keep the one full-bleed bar from this PR — single visual spec.
> 2. Let [feat(Table): add useTableStickyHeader plugin #6221](https://github.com/facebook/astryx/pull/6221)/[feat(Table): pin group headings while their section is on screen #6234](https://github.com/facebook/astryx/pull/6234) land the sticky machinery.
> 3. Then make the toolbar the top element of that same sticky stack, behind one boolean rather than a second layout.
> 
> That gives persistence to every Table consumer rather than only to selection users, and it shrinks the selection API surface instead of growing it — which also matters for the composition question raised in the other review. Happy to sketch the stack ordering against #6234's offset chain if useful.

I dont think this addresses layouts where the table is complemented by another section (lets say a metrics card group above it).

The floating variant is supposed to solve for those - where the entire page is scrollable including the table - it wont have a sticky table header BUT we still want the bulk actions bar to remain visible

humbertovirtudes:
@ernestt Ive reworked this so `useTableSelection` remains selection-only and both bulk-action variants are composed at the product callsite.

Could you pls check the new [floating page-sticky example](https://facebook.github.io/astryx/pr/5489/?path=/story/core-tableselection--bulk-actions-floating)? It demonstrates the missing case with metric cards above a 40-row non-sticky table. The bar stays visible during page scroll and releases at the table region boundary.

humbertovirtudes (commented):
Requesting changes. Product-callsite bulk actions should not become a permanent shared Table selection API or wrapper-layout behavior. The all-matching state can dispatch only explicit selected keys, system strings are not localized, the conditional live region is unreliable, and the public option is undocumented. Please retain the controlled selection API and compose this toolbar at the product callsite.

humbertovirtudes:
Follow-up: #6474 tracks the reusable bulk-actions wrapper and shared selection-state contract. This keeps `useTableSelection` behaviour-only while aligning the wrapper and plugin APIs.

ernestt:
The behavior here is good — I'm happy with how the floating toolbar reads and with the ownership split between the plugin, the wrapper, and the callsite. In particular, keeping the toolbar *outside* Table's inline-axis scroll wrapper is the right call and is exactly what `spec:AST-025` FR10/FR18 ask for: the block-axis sticky resolves to the outer block owner instead of being trapped by the table's own scroller. Not a blocker for the behavior.

What I'd like you to look at before this lands is whether the example scrollport can conform to the sticky/scrollable-area specs, since this is a CLI template and a Storybook story — whatever it does gets copied into consumer apps. `spec:AST-025` is `authority: current` now and `useScrollableArea` / `ScrollableArea` are landed in `packages/core`, so the hand-rolled scrollport is no longer the house pattern.

Concretely, in `styles.page` on both the template block and the story:

1. **`overscrollBehavior: 'contain'` is unconditional.** FR14 requires containment to be active only on axes that *currently* overflow — on the non-overflowing inline axis this is the dead scroll zone the requirement exists to prevent, and Chrome 144 changed behavior here (the spec calls that out as regression-sensitive).

2. **`tabIndex={0}` + `role="region"` are unconditional.** FR12/AR1: a viewport enters the tab order only while an axis is effective. AST-025's current-state section names example scroll wrappers that are focusable by default as something adoption has to fix. The nested `role="region" aria-label="Metrics"` also adds a landmark that isn't a scroll target.

3. **`overflowY: 'auto'` is physical and hand-declared.** FR1/FR17 want logical scroll intent, and the closed axis vocabulary rather than a raw CSS pair.

All three fall out for free if the example wraps the region in `ScrollableArea` (`axis="block"`, `overscroll="contain"`, a `label`) instead of styling a `div`. That's the reference composition per DEC-1, and it'd make the template teach the current pattern.

Two smaller things while you're in there:

4. **`zIndex: 20` with no isolation owner.** `spec:AST-027` FR4/FR5 — no `isolation: isolate` on `styles.page` or `styles.tableRegion`, and that spec explicitly flags CLI-emitted templates as generated leaks needing explicit surface isolation. `20` also reads as a band rather than a local rank; `1` on an isolated surface does the same job.

5. **The placement math is fragile.** `paddingBlockStart` + `translateY` + a negative `marginBlockEnd`, each repeating the same four-token `calc()` chain, to land the toolbar 16px from the scrollport. Any change to `--size-element-sm` or those spacing steps breaks it in three places at once. Worth a single custom property the three declarations read.

None of this changes what the toolbar does — it's all the container around it. Happy to pair on the `ScrollableArea` swap if it turns out to fight the Table wrapper.

PR #5490, opened 2026-08-25
docs(AspectRatio): fix @example block that breaks Storybook autodocs

josephfarina:
## What

The `ratio` prop's `@example` block on AspectRatio held JS `//` comments and a blank line inside the code fence. Both break Storybook's autodocs parser: a blank line can terminate the fence early, and comments confuse the markdown parser, so the rest of the example renders as raw text.

- Move the responsive-ratio explanation and the container-query CSS snippet into the JSDoc description above the `@example` tag (per the doc-reviewer rule: context goes in the description, not as comments inside the code block).
- Keep a single clean example line.
- Replace the ellipsis character in the placeholder children with ASCII `...`.

Comment-only change to the JSDoc; no type or runtime impact.

## Validation
- `astryx component AspectRatio` still renders correctly (exit 0).
- Re-ran the `@example` scan on the file: no comments, blank lines, or bare `>` remain inside the code fence.

Night Watch — Doc Reviewer


cixzhang (approved):
Thanks. The autodocs table takes `ratio`'s description from `AspectRatio.stories.tsx:84`, so this JSDoc never reaches it — before and after render identically. The change is fine; the description just isn't the docs surface.

[Reviewed by Robohands]

packages/core/src/AspectRatio/AspectRatio.tsx
```diff
@@ -53,14 +53,14 @@ export interface AspectRatioProps extends BaseProps<HTMLDivElement> {
    *   `astryx-base` cascade layer, and unlayered CSS wins over layered
    *   rules regardless of specificity.
    *
+   * A responsive ratio is set in CSS, not on the component. For example, a 3:1
```
cixzhang, line 56: This repeats the responsive guidance already in `AspectRatio.doc.mjs`, which is what the docsite shows.

PR #5491, opened 2026-08-25
fix(Card): reflect elevation as a theme target, correct the padding default docs

cixzhang:
First audit of `core/Card`, rubric v1.8. Score **65.1 / D → 74.3 / C** (post-fix; the post-fix row is held until this lands so it records a real `main` commit). The pre-fix row is already in the ledger: [scores: Card D (65.1), rubric 1.8](https://github.com/facebook/astryx/wiki/_compare/0278c85291023acbf07761f9b4f3d29a7a36b58f).

Open BLOCKs 4 → 2. The two that closed are `T6` and `V3`; the two that stay open are decisions, and they are in Needs review below.

## What was wrong, and what changed

**`elevation` never reached the DOM (§2 T6).** `elevationStyles[elevation]` picked a style object inside `stylex.props()`, but `elevation` was missing from the sibling `themeProps('card', {variant})`. Measured in Chromium across all four values before the fix: the box-shadow changes, `data-elevation` is `null` every time. A theme could restyle a card by variant and could not touch its four shadow tiers. It now rides `themeProps` alongside `variant`, and the doc's theming target lists it.

**The documented `padding` default was wrong (§3 P31).** The JSDoc said `@default 4 (16px)` and the prop doc said `default: '4'`. With the prop omitted the card reads the theme's card padding chain, and most shipped themes set `--astryx-card-padding` to a different step. Measured against the default theme: omitted renders 11px of padding, `padding={4}` renders 15px, both after the 1px border inset. Someone writing the documented default explicitly got a different card. The JSDoc, both prop docs, the dense doc and the playground default now say that omitting the prop takes the theme's padding and that passing a step overrides it, without naming a theme that could change under them.

**`padding` encoded the wrong default in code as well as in docs.** `padding ?? 4` and four `effectivePadding !== 4` style branches survived from before `container()` learned to set the container-padding variables itself. The `?? 4` branch was unreachable, since the value is only read when `padding` is non-null, and the four guarded style objects set exactly the variables `container()` already sets. Removed, and verified identical across all eleven padding steps on a bordered and a borderless card: resolved padding, the four `--container-padding-*` variables, `--layout-padding-inner-x`/`-outer-x` and the rendered height, 24 rows, no difference. (Found by the review loop on this PR.)

**`CardProps` re-declared `className` and `style` (§3 P27)**, which `BaseProps` already supplies. Removed; the types are identical, so nothing changes for a caller.

**The tests did not test the contract (§6 V3, V4).** Three of the five assertions were on generated StyleX class names. The elevation test asserted only that four hashed class strings differ, which is why it sat next to the `T6` defect for the component's whole life without noticing it. Replaced with assertions on `...rest` passthrough landing on the theme-target element, `ref` forwarding, `className` merging rather than clobbering, and the variant and elevation data attributes. 5 tests → 8.

**Eight resolved Card entries removed from `.github/a11y-baseline.json` (§1 A18).** The audit reported them resolved; the gate now reports `0 new, 0 baselined, 0 resolved`.

**Stories for empty, long content and a narrow container (§8 X12, §10 R11).** The a11y and RTL sweeps only see what a story renders, and Card had no story for any of the three. All 17 stories are axe-clean.

## Visual evidence

**Nothing rendered changes.** All 38 frames captured against `main` are byte-identical to the same 38 captured at this head, re-checked after the second commit (`cmp`, 38/38 twice). That is the intended result: the elevation axis becomes reachable by a theme without moving a pixel for anyone not using it.

45 frames in total, real Chromium against the built Storybook at deviceScaleFactor 2-3: every story in neutral light and dark, four of them additionally in matcha light and dark, plus 320px reflow, RTL, a forced fixed-height overflow and a long unbroken word.

contact sheet

## Needs review

Four findings I did not fix, because each is a decision rather than a correction.

**1. A fixed `height` makes the card a scroll container a keyboard cannot reach.** `overflow: auto` goes on the card root whenever `height` is set. Measured at 300x120 with taller content: `scrollHeight` 442, `clientHeight` 118, no `tabindex`, zero focusable descendants, so Tab never stops on it and the content below the fold is pointer-only (WCAG 2.1.1). No story renders real overflow, which is why `pr-a11y` has never had the chance to fire `scrollable-region-focusable` here.

Every fix is a choice, so it is yours: an unconditional `tabIndex={0}` whenever `height` is set adds a tab stop for consumers who are not overflowing; gating on real overflow needs an observer; and the third answer is that Card should not be a scroll container at all and `Layout` should own it. I did not add the overflowing story either, because it would add a baseline entry and fail the a11y gate before the underlying question is settled.

fixed height overflow

**2. `variant` is a closed union, so a theme cannot add a card variant (§2 T19).** `CardVariant` is a 13-member union backed by a `Record<CardVariant, Style>`, and there is no `CardVariantMap` in `packages/core/src/Card/index.ts`. Sixteen core components ship the extensible map, `Section` among them, and its variant axis is the same shape as Card's. Fixing it adds an exported interface and widens the type, and it ripples through `ClickableCard` and `SelectableCard`, which import `CardVariant` and switch on it, so it is not an audit's change to make.

**3. `variant="muted"` is invisible on the default page background.** `--color-background-muted` and `--color-background-body` resolve to the same value in both modes, `light-dark(#f1f1f1, #1b1b1b)`, and the muted variant paints no border and no shadow. On the standard page treatment a muted card has no boundary at all. The `Callout` story recommends exactly that use and renders two callouts as bare text while its own copy says the muted background "provides visual contrast".

The component picked the right role token, so the answer is a token or design one: give `muted` a border, split the two tokens, or change what the docs recommend.

muted callout, light

**4. A long unbroken string is clipped out of existence.** The root is `overflow: clip` with no wrap policy, so one long token is cut mid-word with no scrollbar, no ellipsis and no wrap. Measured `scrollWidth` 643 against `clientWidth` 300. `Section` and `Dialog` use the same container, so this is very likely a container-wide question rather than Card's, which is why I left it rather than putting an `overflow-wrap` on one of the three.

long word clipped

**5. `ClickableCard` and `SelectableCard` still document and pre-fill the false padding default.** Both forward `padding` straight into `Card` (`ClickableCard.tsx:305`, `SelectableCard.tsx:354`), and both still say `@default 4 (16px)` in their JSDoc, carry `default: '4'` in their `.doc.mjs`, and seed `padding: 4` into their playground defaults. Once this merges, Card says one thing and the two components that render it say another, and a builder starting from either playground ships a card inset differently from the Card beside it.

It is four lines across two files and I would normally take it here. Both files are open ground for [#5487](https://github.com/facebook/astryx/pull/5487), so it goes to you rather than into this PR.

**6. Three components declare `elevation` and none of them reflects it.** `Button.tsx:293`, `Banner.tsx:147` and `ChatComposer.tsx:95` all take a public `elevation` prop; none passes it to `themeProps`, so after this merges Card and `ButtonGroup` reflect elevation and those three do not. A theme author who reads this changelog line and writes the same rule for Button gets a selector that matches nothing, with no error and no warning.

Underneath both of those is one system gap, which is why I would not fix the six instances by hand: **nothing checks either half.** `themingTargets.test.ts` is a subset policy by design, so a component whose docs and source agree on an incomplete list stays green forever, which is how Card shipped an unreachable elevation axis for its whole life. And the `default` field in a `.doc.mjs` is checked by nothing at all: `docPropLiterals.test.ts` and `docPropReferences.test.ts` both derive `type` from the TypeScript sources at test time, and neither reads `default`. A default that comes from a CSS custom-property chain can never be right by hand.

## Also not done

- **The RTL applicability record has nowhere to go.** `rtl:audit --filter Card` returns `AUTO 0 pass / 0 fail / 1 N-A` and `PM 0 pass / 0 fail / 17 N-A`. Card is genuinely not applicable: no directional glyph, no connector, no positioned affordance, no control order, no horizontal keyboard model, and logical properties throughout. Rubric v1.8 asks for that to be recorded as a checked-in verified-N-A reason, but `apps/storybook/rtl-audit/` has no such file and `rtl-audit.mjs` reads nothing of the kind, so there is no place to write it that anything would read. Left as an open finding in the ledger row rather than adding an orphan config file.
- **`docsZh`'s prose is still English** while its prop descriptions are Chinese. I added the missing `variant` entry and the two missing vars, but writing the translation is not an audit's job.
- **The `NestedCards` story demonstrates, approvingly, the nesting the doc's own don't-list forbids**, and two nested cards render the same 12px radius 11px apart, so the pair is not concentric. Card cannot detect that it is nested, so both are notes rather than fixes.

## Checks

`pnpm lint:strict` 0 errors · `pnpm build` green · `pnpm test` 12205 pass, 3 pre-existing failures unrelated to this change (`story-tree` fails on `Core/Icon/Size Theming`, which is on `main`; the two CLI discovery tests fail only on a case-insensitive filesystem, resolving `useMediaquery.doc.mjs`) · `a11y:audit --components Card` 0 violations over 17 stories, 0 new · `check:sync`, `check-use-client`, `sync-exports --check`, `verify-exports`, `check:changesets`, `typecheck:docs`, docsite generate + tests all green.

### DIFF
--- a/.changeset/card-nightly-audit-20260825.md
+++ b/.changeset/card-nightly-audit-20260825.md
@@ -1,0 +1,11 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Card: reflect `elevation` as a theme target so a theme can reach it, and correct the documented `padding` default
+
+`elevation` picked a style object but never reached the DOM, so `astryx-card` exposed `data-variant` and nothing for elevation and a theme could not style the four shadow tiers. It now rides `themeProps` alongside `variant`.
+
+The `padding` prop documented `4` as its default. With the prop omitted the card reads the theme's card padding, which every shipped theme but `butter` sets to spacing step 3, so writing the documented default explicitly made the card wider. The prop docs, the JSDoc and the playground default now say what actually happens.
+
+@cixzhang
--- a/.changeset/card-nightly-audit-20260825.md
+++ b/.changeset/card-nightly-audit-20260825.md
@@ -6,6 +6,6 @@
 
 `elevation` picked a style object but never reached the DOM, so `astryx-card` exposed `data-variant` and nothing for elevation and a theme could not style the four shadow tiers. It now rides `themeProps` alongside `variant`.
 
-The `padding` prop documented `4` as its default. With the prop omitted the card reads the theme's card padding, which every shipped theme but `butter` sets to spacing step 3, so writing the documented default explicitly made the card wider. The prop docs, the JSDoc and the playground default now say what actually happens, and the four `effectivePadding !== 4` style branches that encoded the same wrong default in code are gone: `container()` already sets every variable they set, verified identical across all eleven padding steps on both a bordered and a borderless card.
+The `padding` prop documented `4` as its default. With the prop omitted the card reads the theme's card padding, and most shipped themes set that to a different step, so writing the documented default explicitly changed the card's size. The prop docs, the JSDoc and the playground default now say that omitting the prop takes the theme's padding and passing a step overrides it. The four `effectivePadding !== 4` style branches that encoded the same wrong default in code are gone: `container()` already sets every variable they set, verified identical across all eleven padding steps on both a bordered and a borderless card.
 
 @cixzhang
--- a/.changeset/card-nightly-audit-20260825.md
+++ b/.changeset/card-nightly-audit-20260825.md
@@ -6,6 +6,6 @@
 
 `elevation` picked a style object but never reached the DOM, so `astryx-card` exposed `data-variant` and nothing for elevation and a theme could not style the four shadow tiers. It now rides `themeProps` alongside `variant`.
 
-The `padding` prop documented `4` as its default. With the prop omitted the card reads the theme's card padding, which every shipped theme but `butter` sets to spacing step 3, so writing the documented default explicitly made the card wider. The prop docs, the JSDoc and the playground default now say what actually happens.
+The `padding` prop documented `4` as its default. With the prop omitted the card reads the theme's card padding, which every shipped theme but `butter` sets to spacing step 3, so writing the documented default explicitly made the card wider. The prop docs, the JSDoc and the playground default now say what actually happens, and the four `effectivePadding !== 4` style branches that encoded the same wrong default in code are gone: `container()` already sets every variable they set, verified identical across all eleven padding steps on both a bordered and a borderless card.
 
 @cixzhang
--- a/.github/a11y-baseline.json
+++ b/.github/a11y-baseline.json
@@ -14,46 +14,6 @@
       "helpUrl": "https://dequeuniversity.com/rules/axe/4.12/color-contrast?application=playwright"
     },
     {
-      "key": "Card::Callout::color-contrast",
-      "impact": "serious",
-      "helpUrl": "https://dequeuniversity.com/rules/axe/4.12/color-contrast?application=playwright"
-    },
-    {
-      "key": "Card::Color Variants::color-contrast",
-      "impact": "serious",
-      "helpUrl": "https://dequeuniversity.com/rules/axe/4.12/color-contrast?application=playwright"
-    },
-    {
-      "key": "Card::Elevations::color-contrast",
-      "impact": "serious",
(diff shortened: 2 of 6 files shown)


cixzhang:
Thanks — 38 elements across 14 stories, only `data-elevation` moves. ClickableCard and SelectableCard still pre-fill `padding={4}` in their playgrounds and document it as the default.

[Reviewed by Robohands]

PR #5493, opened 2026-08-25
fix(TreeList): forward BaseProps attributes to the tree element (#5476)

### LINKED ISSUE #5476: [Bug] List and TreeList drop every BaseProps attribute, so aria-label never reaches the DOM
### Description

`ListProps` and `TreeListProps` both extend `BaseProps`, whose file header states it keeps `aria-*`, `role`, `tabIndex`, `id` and event handlers. Both components destructure a fixed set of props and neither has a rest spread, so everything `BaseProps` promises is dropped on the floor. TypeScript accepts the prop, nothing warns, and the attribute never reaches the DOM.

```
render(<List aria-label="Recent files" id="x"><ListItem>a</ListItem></List>);

// <ul role="list" class="astryx-list …" data-density="balanced" data-list-style="none">
// aria-label → null
// id        → null
```

`TreeList` behaves identically.

The accessibility cost is the sharp one. Both components render a container whose only naming path is `aria-labelledby`, wired to the `header` prop. A list without a visible header cannot be named at all: a screen reader announces "list, 3 items" and nothing says which list.

These two read as one omission copied rather than two independent ones. `List` and `TreeList` share the same `aria-labelledby={header != null ? headerId : undefined}` line and the same destructure shape. Meanwhile `ListItem`, in the same directory, spreads `...restProps`, and `Kbd` spreads `...rest`, so the spread is the house pattern and these are the exceptions.

Ten more components match the no-rest-spread shape. I have not verified them and at least one is deliberate (`Timestamp` composes its own `aria-label`), so this is a list worth checking rather than a claim: `Carousel`, `ChatComposerDrawer`, `Markdown`, `MetadataListItem`, `NavHeadingMenu`, `Timestamp`, `Token`, `TopNavMegaMenu`, `TopNavMenu`, `TypeaheadItem`.

### Reproduction

```
import {render} from '@testing-library/react';
import {List, ListItem} from '@astryxdesign/core';

const {container} = render(
  <List aria-label="Recent files"><ListItem>a</ListItem></List>,
);
container.querySelector('ul')?.getAttribute('aria-label'); // null
```

### Astryx Version

`@astryxdesign/core@0.5.0`, and current `main`.

### Environment

Any browser and OS. Not platform dependent.

---

gonzoblasco:
Closes #5476

## Problem

`TreeListProps` extends `BaseProps`, whose header states it keeps `aria-*`, `role`, `tabIndex`, `id` and event handlers. `TreeList` destructures a fixed set of props with no rest spread, so everything `BaseProps` promises is dropped and never reaches the DOM.

```tsx
render(<TreeList items={items} aria-label="File tree" />);

// <div class="astryx-tree-list ...">
//   <ul role="tree">
// aria-label -> null
```

The accessibility cost is the sharp one. The tree's only naming path is `aria-labelledby`, wired to the `header` prop. A tree without a visible header cannot be named at all: a screen reader announces an unnamed tree and nothing says which tree it is.

## Change

Add `...restProps` to `TreeList` and spread it onto the root element, matching the existing `ListItem` / `Kbd` pattern. `aria-*`, `role`, `tabIndex`, `id` and event handlers now forward to the DOM as `BaseProps` promises.

`aria-label` and `aria-labelledby` are pulled out of the rest and routed onto the `<ul role="tree">` itself, so `<TreeList aria-label="File tree">` names the tree. A consumer-supplied `aria-labelledby` wins over the internal header association, and the contract `role="tree"` is written after the rest spread so a consumer cannot displace it.

## Test plan

- `packages/core/src/TreeList/TreeList.test.tsx` - regression tests asserting `aria-label` reaches the `<ul role="tree">` and `id` reaches the root element.
- `pnpm vitest run packages/core/src/TreeList/TreeList.test.tsx` - 78 passing.
- `pnpm vitest run packages/core/src/List/List.test.tsx` - 53 passing (unchanged).
- `npx eslint` on changed files - clean.
- `npx prettier --check` on changed files - clean.

### DIFF
--- a/.changeset/treelist-forward-baseprops.md
+++ b/.changeset/treelist-forward-baseprops.md
@@ -6,6 +6,6 @@
 
 `TreeList` destructured a fixed set of props with no rest spread, so every `BaseProps` attribute (`aria-*`, `role`, `tabIndex`, `id`, event handlers) was dropped and never reached the DOM. A tree without a visible header could not be named at all - a screen reader announced an unnamed tree with no way to know what it was.
 
-The component now spreads the remaining props onto the root element and routes `aria-label`/`aria-labelledby` onto the `<ul role="tree">` itself, so `<TreeList aria-label="File tree">` names the tree. A consumer-supplied `aria-labelledby` wins over the internal header association, and the contract `role="tree"` is written after the rest spread so a consumer cannot displace it.
+The component now spreads the remaining props onto the root element and routes `aria-label`/`aria-labelledby` onto the `<ul role="tree">` itself, so `<TreeList aria-label="File tree">` names the tree. When a visible `header` is rendered, it keeps naming the tree (AT hears the same name the user sees); a consumer-supplied `aria-labelledby` only applies on the headerless path. The contract `role="tree"` is written after the rest spread so a consumer cannot displace it.
 
 @gonzoblasco
--- a/.changeset/treelist-forward-baseprops.md
+++ b/.changeset/treelist-forward-baseprops.md
@@ -1,0 +1,11 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] TreeList: forward `aria-label`/`aria-labelledby` to the `role="tree"` element so a tree can be named
+
+`TreeList` destructured a fixed set of props with no rest spread, so every `BaseProps` attribute (`aria-*`, `role`, `tabIndex`, `id`, event handlers) was dropped and never reached the DOM. A tree without a visible header could not be named at all - a screen reader announced an unnamed tree with no way to know what it was.
+
+The component now spreads the remaining props onto the root element and routes `aria-label`/`aria-labelledby` onto the `<ul role="tree">` itself, so `<TreeList aria-label="File tree">` names the tree. A consumer-supplied `aria-labelledby` wins over the internal header association, and the contract `role="tree"` is written after the rest spread so a consumer cannot displace it.
+
+@gonzoblasco
--- a/packages/core/src/TreeList/TreeList.test.tsx
+++ b/packages/core/src/TreeList/TreeList.test.tsx
@@ -206,6 +206,43 @@ describe('TreeList', () => {
     expect(tree).not.toHaveAttribute('aria-labelledby');
   });
 
+  it('prefers the visible header id over caller aria-labelledby when header exists', () => {
+    render(
+      <TreeList
+        items={simpleItems}
+        header={<span>File Tree</span>}
+        aria-labelledby="external-label"
+      />,
+    );
+    const tree = screen.getByRole('tree');
+    const headerId = tree.getAttribute('aria-labelledby');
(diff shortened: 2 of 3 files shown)


cixzhang (changes requested):
Thanks — the diagnosis is right, and `List` names correctly now. Two things before it lands, plus a changeset.

**The tree still can't be named.** `aria-label` lands on the layout `<div>`; the `<ul>` that carries `role="tree"` gets nothing, so `<TreeList aria-label="File tree">` still announces as an unnamed tree. The new test asserts on `container.firstElementChild` — that div. Chromium's a11y tree:

accessible names

**A caller can now displace `role="list"`.** Your body promises `role` forwards, and `{...restProps}` is spread after `role="list"` — so `<List role="listbox">` wins and the role restoring VoiceOver's "list, N items" is gone. API Conventions: contract props go "after `{...rest}` … so consumer input can't clobber the contract."

role clobbered

For TreeList, pull `aria-label`/`aria-labelledby` out of the rest onto the `<ul role="tree">`.

#5288 takes the `List` half and goes first. Want to narrow this to TreeList and land it on its own?

[Reviewed by Robohands]

gonzoblasco:
Thanks for the review, @cixzhang - the diagnosis was spot on, and I took both points to adjust the commit.

Narrowed to TreeList and landed both fixes.

`aria-label`/`aria-labelledby` are now pulled out of the rest and routed onto the `<ul role="tree">` itself, so `<TreeList aria-label="File tree">` names the tree. A consumer-supplied `aria-labelledby` wins over the internal header association, and it is only written when the header exists, so it no longer erases a label a consumer pointed at their own heading. The test now asserts on `getByRole('tree')` instead of the wrapper div.

The contract `role="tree"` is written after the rest spread, so a consumer can no longer displace it.

Added a changeset. The `List` half is left to #5288.

cixzhang (changes requested):
Thanks, this is much closer. The List half is resolved, and `aria-label` now lands on the right `<ul role="tree">`.

One path still needs the same precedence as #5288: when TreeList renders a visible header, that generated header id should keep naming the tree. Right now an external `aria-labelledby` replaces it, so AT hears a different name than the visible header.

Can you make the header id win when `header` exists, and only use caller `aria-labelledby` on the headerless path? If you’d rather talk it through with someone, we’re in [Discord](https://discord.com/invite/XnsUcFykEP).

[Reviewed by Robohands]

gonzoblasco:
Thanks, @cixzhang - you're right that the header should keep naming the tree when it's visible. Applied the same precedence as #5288.

When a visible `header` is rendered, the tree now names itself via the generated header id, so AT hears the same name the user sees. A caller `aria-labelledby` (or `aria-label`) only applies on the headerless path. Added tests covering both paths.

cixzhang (approved):
Thanks for turning this around so fast — the header names the tree now, and the headerless paths still name it. Checked the accessible names in Chromium.

[Reviewed by Robohands]

PR #5495, opened 2026-08-25
feat(Stepper): --step-connector-gap, so a theme can stop the on-track connector short of the indicator

freddymeta:
## Problem

In the on-track Stepper layout, the connector runs through the indicator. Themes sometimes need a deliberate gap around that node, but the existing `step-connector` target cannot create it reliably: the connector paints both its own background and an accent `::before`, while a theme target reaches only the element.

## Change

Adds the documented public custom property `--step-connector-gap`:

```ts
defineTheme({
  components: {
    stepper: {
      base: {'--step-connector-gap': '6px'},
    },
  },
});
```

- Default is `0px`, preserving existing pixels.
- The default is declared once on the `stepper` root, so generated theme CSS can override it and connectors inherit it.
- One `clip-path: inset()` clips the track and its `::before` fill together.
- Values are bounded to `0 … --spacing-2`; clipping cannot resize the Stepper.
- `indicator="none"` leaves the connector continuous.
- Horizontal clipping mirrors under RTL.
- Internal connector pieces remain implementation details—no public segment vocabulary.

The draft `Stepper.spec.md` owns the API admission argument, value contract, representative states, and anatomy-to-target map.

## Verification

- Full repository preflight passes: mergeability, public-boundary scan, typecheck, lint, build, 8,647 core tests, Chromium accessibility guards, knowledge checks, and changesets.
- Generated-theme regression proves the override is emitted on `.astryx-stepper`, never `.astryx-step-connector`.
- Real browser proof using CSS emitted by `astryx theme build`:
  - baseline root value: `0px`
  - themed root value: `6px`
  - both connector clips resolve to `6px`
  - Stepper height remains `60px`
- RTL browser audit: D2 pass; 1 measured, 0 gaps, 0 stale declarations.


cixzhang (commented):
Thanks — this is an approve once you undraft. `segment` matches the shipped positional-axis pattern and works on every painted connector; the value-discovery gap can follow separately.


[Reviewed by Robohands]

freddymeta:
Refreshed main — the only red was `pr-visual` from 25 Aug, which predates the infra fix in #5481/#5482. Stepper's 50 tests pass on the merge.

@cixzhang your last note was "this is an approve once you undraft" — it has been undrafted since then, so I think this one is just waiting on the click.

cixzhang (approved):
Thanks — the prior approve still holds after the main refresh and undraft. `segment` reaches all connector roles, including `indicator="none"`, with no default pixel change.

[Reviewed by Robohands]

freddymeta:
/accept-visual 33046093232/1 These frames show the intentional step-connector segment states on the existing painted target; the theme-state mapping is the reviewed change.

freddymeta:

cixzhang (changes requested):
`segment` is not ready as public theming vocabulary. It emits collision-prone global classes (`lead`, `rail`, `content`), generated docs expose neither valid values nor their layout-specific meanings, and bogus values build silently.

Please first decide whether these roles are a stable public contract. If so, namespace, document, and validate the exact closed vocabulary and availability.

[Reviewed by Robohands]

freddymeta:
You asked me to first decide whether these roles are a stable public contract. **They are not**, so I have not namespaced/documented/validated them — I have removed them. Checking each of your three points against the source is what convinced me:

**The vocabulary was undiscoverable.** Every other `visualProps` entry maps to a documented public prop (`variant`, `size`, `status`) or is described in prose (`progress`). `segment` was neither — `lead`/`rail`/`content` appeared nowhere in either locale's docs except as the prop name. There was no way to learn them short of reading `Step.tsx`.

**The bare classes were the worse kind.** `classTokenForPropValue` only prefixes values starting with a digit, so the DOM carried `class="astryx-step-connector content"`. Bare variant classes are the house pattern — but the shipped ones are `drawer`, `topbar`, `bottom-sheet`, `collapsed`. `.content` and `.lead` are in a different risk class; Bootstrap alone ships `.lead`.

**And `lead` meant two different things.** Vertically it is `otSegLeadV` — fixed height, existing to offset the node down to the label's first line. Horizontally it is `otSegH` — `flex: 1`. Same word, different geometry per orientation. That is your "layout-specific meanings" point, and it is the one that settled it for me: a theme writing `'segment:lead'` cannot know what it will get.

## What ships instead

`--step-connector-gap`, a documented public var. Astryx spends it on whichever side each segment faces the node from, so the pair leaves a symmetric hole and the caller never names the pieces.

Measured in Chromium at `4px`:

| | padding | fill inset | outer box |
|---|---|---|---|
| vertical | `padding-bottom` 0 → 4px | `bottom` 0 → 4px | 4×8 **unchanged** |
| horizontal | `padding-inline-end` 0 → 4px | `right` 0 → 4px | 62×4 **unchanged** |

Three things I checked rather than assumed:

- **The node does not move.** Padding, not margin, with `box-sizing: border-box` universal in the reset. Critical for the vertical lead segment, whose height is the thing aligning the indicator with the label's first line — a margin there would push it down by the gap.
- **The fill gives up the same space.** It is absolutely placed at `inset: 0`, which resolves against the *padding* box — so padding alone would have left the accent painting through the gap, and the gap would have appeared only on incomplete steps. Physical shorthand vs logical longhand cascade in declaration order, so I measured which lands later instead of reasoning about it. Longhand wins; `fillBottom` and `fillRight` both report 4px.
- **RTL still flips** — inline axis stays logical, only the block axis is physical.

## On the consumer need

The gap is what that theme's padding rules actually want. Their remaining lead-vs-rail selectors re-derive the connector *fill* from their own lifecycle, because Astryx derives it from `activeStep` and that cannot express an approval chain (completed → error → not_started). That is a progress-model question, not a theming one, and I do not think a theme target should paper over it — flagging it rather than folding it in.

53 tests pass. Five new: both axes, the fill inset, the `0px` default, and a guard asserting no `data-segment` and no bare `lead`/`rail`/`content` class is emitted, so the vocabulary cannot come back by accident.

cixzhang (changes requested):
Thanks — removing `segment` resolves my earlier request. Two parts of the replacement still need tightening:

- [Component vars are root-owned](https://github.com/facebook/astryx/blob/d6badf2169412461663da1f38682256f0da56c80/packages/cli/authoring/doctypes/base/type.ts#L455-L458), but [every connector resets this one to `0px`](https://github.com/facebook/astryx/blob/d6badf2169412461663da1f38682256f0da56c80/packages/core/src/Stepper/Step.tsx#L754-L776); a generated `stepper` override compiles cleanly and changes nothing. Please keep one root-owned default and cover the built-theme path.
- [The value/state contract](https://github.com/facebook/astryx/blob/d6badf2169412461663da1f38682256f0da56c80/packages/core/src/Stepper/Stepper.doc.mjs#L91-L97) can render broken tracks: `indicator="none"` gets holes, `-4px` paints through nodes, `10%` misaligns track and fill, and `1rem` grows the vertical Stepper 180→240px. Please define safe bounds and cover both axes plus no-indicator.

[Reviewed by Robohands]

freddymeta:
Squashed to one commit — the branch had eleven, five of them merges from main, and the net change is five files. Both findings were real; here is each, measured.

## 1 · Root-owned var

Correct, and the reason it slipped through is worth recording: the reachability gate sets the var on `.astryx-step-connector` — the element that declared it — so it passed. **No theme writes that.** A theme writes `stepper: {base: {…}}`, which lands on the root, and every connector re-declared `0px` on itself, so the inherited value never won.

The default is now declared **once on the Stepper root**. Verified through the built-theme path rather than the gate's:

```
@layer astryx-theme { .astryx-stepper { --step-connector-gap: 6px } }
  connector padding-bottom  0px -> 6px     (was a no-op)
  fill      padding-bottom  6px            (gives up the same space)
```

## 2 · Value/state contract

All four reproduced. Chromium, vertical on-track, 180px stepper:

| value | before | now |
|---|---|---|
| `-4px` | invalid padding drops the declaration; track paints through the node | clamped to `0` |
| `1rem` | **180px → 240px** | capped, height unchanged |
| `999px` | same growth | capped, height unchanged |
| `10%` | track and fill stop in different places | both use padding |
| `indicator="none"` | holes with no node in them | no gap applied |

**The cap is the flexible segment's own `min-height`.** Above it the padding is larger than the smallest box the track may occupy, so the segment grows to fit and takes the Stepper with it. That is the 180→240 you saw.

**Track and fill both spend it as padding now**, instead of padding on one and an inset on the other — `padding-block-end: 10%` resolves against the inline size, `inset-block-end: 10%` against the block size, which is exactly why they disagreed.

## One thing I could not close

**A percentage still resolves, and track and fill differ by ~1.2px** — measured `pad=1.59px` vs `fill=0.39px` at `10%`. The two declarations sit on boxes with different containing blocks (the segment's containing block vs the segment's own 4px-wide padding box), so no arrangement of padding makes a percentage agree.

The cap bounds it to sub-pixel rather than unbounded, and the var is documented as a length. CSS cannot reject a unit without `@property`, and registering it gives every element an initial value — which points `theme-var-reachability` at `<html>` and reports a var no theme can select, the exact trap [Spinner's comment warns about](https://github.com/facebook/astryx/blob/main/packages/core/src/Spinner/Spinner.tsx).

If you would rather percentages were impossible than bounded, the route I can see is `overflow: clip` + `overflow-clip-margin: content-box` on the segment, letting the segment's own content box clip the fill so there is only one declaration. I did not ship it because Safari support is recent and I have not measured it.

8600 core tests, `check:repo` green, lint and tsc clean. 55 Stepper tests including no-indicator, both axes, the clamp and the root-ownership assertion.

freddymeta:
Both your points hold up, and both are covered below. But going back through `architecture:component-theming-surface` I found the thing that was actually still blocking this, which neither of us had named — so that is first.

## The admission argument was missing, and I had not proved it

INV10 admits a public custom property only when **no guaranteed CSS property can express a caller-owned need**, and INV11 says the same from the API side. The catalog deliberately excludes `gap`. So the question this PR has to answer, and never did, is: can a theme get this with `paddingBlock` on `step-connector`, which is an existing target and a catalog property?

I assumed not. I have now measured it, in Chromium, against a real `defineTheme` → `generateThemeCSS` override — not the reachability gate's path. Three steps, vertical, `activeStep={1}`. Painted runs read down the centre of each connector; `hole` is background, `paint` is track or fill:

| override on `step-connector` | painted result | stepper height |
| --- | --- | --- |
| none | `c1 paint:11.75` · `c2 paint:7.75` | 108px |
| `paddingBlock: 6px` | `c1 paint:11.75` · `c2 paint:11.75` — **no hole anywhere** | **120px** |
| `paddingBlockEnd: 6px` | `c3 paint:5.75 + hole:6.00` — hole on the **far** side | 108px |
| `--step-connector-gap: 6px` | `c1 hole:5.75 + paint:6.00` · `c2 paint:1.75 + hole:6.00` | 108px |

Three separate reasons the guaranteed-property route cannot do it, all measured:

1. **A theme cannot reach the fill.** A connector paints in two layers — the element's own background, and a `::before` at `inset: 0` carrying the accent. Under every padding override the `::before` computed padding stayed `0px/0px/0px/0px`: the fill paints straight through any hole the track gives up. The gap therefore disappears on exactly the completed steps where it is wanted. Only the component can put one value on both layers.
2. **It erases the unfilled track.** `paddingBlock: 6px` takes the segments' content box to `0`, so incomplete track stops painting entirely — `c3`/`c4` went from `paint` to `hole:11.75`.
3. **One declaration cannot tell the two sides apart.** The lead faces the node with its block-end, the rail with its block-start. `paddingBlockEnd` opens the join *between* steps rather than clearing the node.

So the var survives the bar — but the argument for it belonged in a record, and per `spec:AST-002` FR12 a public-API addition whose canonical owning record is missing is rejected now. Stepper had no `.spec.md` at all. Added: `Stepper.spec.md` (draft) with DEC-1 carrying the above, the value contract, and the `anatomy-theming:v1` map.

While writing the map I hit INV2: `stepper`, `step` and `step-connector` are current targets with no anatomy entry to anchor them. `Stepper.doc.mjs` now has them. Connector is **one** part with **one** target — the pieces the on-track layout draws it from stay implementation, which is the same INV3 reasoning that killed `segment`, now written down instead of argued in a thread.

## 1 · Root-owned var

Correct. Negative control — default moved back onto the connector, everything else identical, theme still setting `6px`:

```
default on connector:  c1 pad=0px/0px  paint:11.75   <- override does nothing
default on root:       c1 pad=6px/0px  hole:5.75 + paint:6.00
```

Test suite catches it: `× declares the default on the Stepper root, where a theme reaches it`, 1 failed | 54 passed.

The authoring contract agrees — `ComponentThemingVar` says vars "are set on the component's root element and can be overridden via `defineTheme` component overrides".

## 2 · Value/state contract

| value | measured now |
| --- | --- |
| `6px` | symmetric hole each side of the node, joins intact, height unchanged |
| `1rem` | capped to `8px`, height 108px unchanged |
| `-4px` | `pad=0px`, track continuous |
| `10%` | resolves; track `1.59px` vs fill `0.39px` |
| `indicator="none"` | no padding, no content-box clip, track continuous — identical to baseline |

Negative control on the cap: remove it and `1rem` gives `pad=16px` on every segment, **every segment a hole**, stepper **108px → 144px**.

Negative control on the fill: drop the `::before` padding and the hole survives only on incomplete steps — `c1` goes back to `paint:11.75` while `c3` keeps `hole:5.75 + paint:6.00`. Caught by `× spends the gap on the fill as well as the track`.

**One correction to my own last comment.** I said a negative value drops the declaration and paints the track through the node. That is wrong, and I should not have written it without checking. Chromium resolves `-4px`, `calc(4px - 10px)` and `min(-4px, 8px)` to `0px` on its own, so a negative gap already degrades to "no gap". `max(0px, …)` is defensive, not corrective — the cap is the half doing the work. The code comment, the changeset and the test comment all said the wrong thing; all three now say this.

`10%` is still not closed and I have not shipped a fix — it is `OQ1` in the spec rather than a silent gap.

## Gates

`@astryxdesign/core` build ✅ · `vitest packages/core/src packages/lab/src` **316 files, 8956 passed** ✅ · `typecheck` ✅ · `typecheck:docs` ✅ · `check:repo` ✅ · `lint:strict` **exit 0, 85 warnings — byte-identical to a stashed clean tree**, none in `Stepper/`.

Full `pnpm test` has 3 failures — `ContextMenu` BottomSheet ×2 and the visual-gate symlink test. Identical on a stashed clean tree (`3 failed | 83 passed` both ways), so they are not from this change. Two more (`hook-discovery`, `theming-targets`) time out at 30s under full-suite parallelism and pass in isolation at 796ms and 2935ms.

Merged `origin/main` before starting — the branch was already current, so no new merge commit. Pushed as a new commit, not a force-push.

freddymeta:
Re-evaluated against [#5824](https://github.com/facebook/astryx/pull/5824), which landed after your review. **It closes the gap I said I could not close** — my previous comment is stale on that point.

The new rule:

> A public input has one stable semantic responsibility across its **full value domain** and **every accepted input shape**.

I had left `--step-connector-gap` documented as a length, with a percentage that still resolved and put track and fill ~1.2px apart. Bounded, but exactly the shape #5824 prohibits: an accepted input shape whose behaviour differs from the rest of the domain.

**Fixed by removing the second declaration rather than documenting around it.** The gap now clips instead of padding — one `clip-path: inset()` on the segment, which clips the element and its `::before` together against one reference box. There is no per-layer copy left to disagree, so a percentage is as correct as a length.

Painted pixels down a 12px segment, through a built theme override:

```
6px    -> 6px clipped        1rem  -> capped 8px
-4px   -> 0 (invalid inset)  999px -> capped 8px
10%    -> 1px                50%   -> capped 6px
```

Stepper height unchanged in every case — clipping cannot affect layout, so the node the segment positions cannot move. That also retires the 180→240px growth separately from the cap.

Everything in my previous comment still stands: root-owned default, clamped bounds, no gap when `indicator="none"`. Squashed to one commit. 8600 core tests, `check:repo` green.

freddymeta:
*(Replaces my previous comment, which was too long — and partly stale: the branch was force-pushed to the `clip-path` revision while I was measuring the padding one. My work is rebased on top of it as `f2d97b00`, a normal push.)*

`clip-path` is the better call — one declaration clips the element and its `::before` against one reference box, so the two layers can't disagree.

**One real bug in it: RTL.** `clip-path: inset()` is physical, and the horizontal row reverses under `dir="rtl"`. Both segments clipped the edge facing *away* from the indicator, so the hole opened at the step join and the track ran through the node.

| `dir="rtl"`, horizontal | con0 (right of node) | con1 (left of node) |
|---|---|---|
| before | clips **right** ← wrong | clips **left** ← wrong |
| after | clips left | clips right |

LTR byte-identical. Fixed with the house pattern (`:is([dir="rtl"] *)`, as Carousel does for `mask-image`) and guarded. Negative control: revert the mirror → assertion fails (`1 failed | 55 passed`) and the browser clips the wrong edge again. `pr-rtl` would not have caught this — it's soft-red repo-wide.

**The admission argument, which neither of us had named.** `spec:AST-002` FR12 rejects a public-API addition whose canonical owning record is missing, and Stepper had no `.spec.md` at all — so `--step-connector-gap` had nothing owning its admission, bounds or scope. Added `Stepper.spec.md` (draft) with DEC-1.

The argument itself, measured rather than asserted — can `paddingBlock` on `step-connector` do this instead, making the var unnecessary?

| override on `step-connector` | result |
|---|---|
| `paddingBlock: 6px` | **no hole at all**; Stepper grows 108 → 120px |
| `paddingBlockEnd: 6px` | no hole; addresses one of two edges |

It can't: the connector paints in two layers — the element (track) and a `::before` at `inset: 0` (accent fill) — and a theme target reaches the element, never its pseudo-element. That's INV10's "no guaranteed CSS property can express it".

Also added anatomy rows for `stepper`, `step` and `step-connector`, which were current targets with no anatomy anchor (INV2 gap).

Gates: 8957 tests pass; build, typecheck, typecheck:docs, check:repo clean; `lint:strict` 85 warnings, identical on a stashed clean tree, none in `Stepper/`. `pr-rtl` red on both parents too.

**One open question I couldn't close** (recorded as OQ1 rather than papered over): a percentage resolves against each segment's own box, so a fixed and a flexible segment clip by different amounts. Both layers of any one segment agree exactly and the cap bounds it, so I documented it as accepted.

freddymeta:
Addressed the latest review on the refreshed current-main head:

- The default is root-owned on `.astryx-stepper`; a focused generated-theme regression now proves `--step-connector-gap: 6px` is emitted on that root and never redeclared on connectors.
- The value contract is bounded and uses one clip for both paint layers. Tests cover vertical, horizontal, RTL, negative/oversized values, percentages, and `indicator="none"`.
- Removed stale source comments that still described the abandoned padding implementation.
- Added the existing Stepper D2 browser target to this component PR. The real RTL audit now reports `1 measured / 0 gap / 0 stale`.
- Merged current `main` and ran the full repository preflight: 8,647 core tests, build, typecheck, lint, browser guards, knowledge checks, and changesets all pass.

Real browser proof using CSS emitted by `astryx theme build`: root `0px → 6px`; both connector clips resolve to `6px`; Stepper height stays `60px`.

— Puck, on behalf of Freddy

cixzhang (approved):
Spec reviewed against the component contract template, current theming-surface architecture, and AST-002. Ratified: Connector is the optional on-track anatomy; its component-owned inner gap is an appropriate public semantic CSS variable because existing target properties cannot express it across both paint layers.\n\nApproved at exact head a786939f6827.

PR #5507, opened 2026-08-25
test(layer): cover body type contexts and xstyle precedence

### LINKED ISSUE #5424: Layer body type: main has only the single-case assertion
[#5066](https://github.com/facebook/astryx/pull/5066) was closed after its test was ported into [#5064](https://github.com/facebook/astryx/pull/5064) — but auto-merge fired between the push and the merge, so `main` carries only the single-case assertion.

Never landed, and promised in [this comment](https://github.com/facebook/astryx/pull/5066#issuecomment-5304445720):

- the two-ambient-size parameterization
- the `xstyle`-precedence case

Test coverage only; no public surface.

jiunshinn:
## Context

#5064 merged the Layer body-typography production fix, but auto-merge completed before the stronger test update was included. As a result, main retained only one fixed-mode assertion with a 30px host, while the two ambient-size cases and the consumer override case from #5066 never landed.

## What changed

- parameterize the Layer body-type assertion over 13px and 20px ambient contexts
- assert both the body size and body leading token declarations
- assert that a consumer-provided supporting-size xstyle still wins over the Layer baseline

This is test coverage only; there is no production or public-surface change, so no changeset is needed.

## Testing

- pnpm vitest run packages/core/src/Layer/useLayer.test.tsx — 45 passed
- pnpm -F @astryxdesign/core typecheck — passed
- ASTRYX_STRICT_LINT=1 pnpm exec eslint packages/core/src/Layer/useLayer.test.tsx — passed
- pnpm exec prettier --check packages/core/src/Layer/useLayer.test.tsx — passed
- pnpm check:changesets — passed
- pre-commit repository checks — passed

Fixes #5424

### DIFF
--- a/packages/core/src/Layer/useLayer.test.tsx
+++ b/packages/core/src/Layer/useLayer.test.tsx
@@ -12,7 +12,9 @@
 import {describe, it, expect, vi, afterEach} from 'vitest';
 import {render, act, fireEvent, waitFor} from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
+import * as stylex from '@stylexjs/stylex';
 import {useLayer, getPositionTryFallbacks} from './useLayer';
+import {typeScaleVars} from '../theme/tokens.stylex';
 import type {
   LayerPlacement,
   LayerAlignment,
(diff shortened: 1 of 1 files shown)


cixzhang (approved):
Thanks, this is good. Mutating either the body baseline or consumer `xstyle` precedence fails the matching test, so the coverage is real.

[Reviewed by Robohands]

PR #5508, opened 2026-08-25
fix(Selector): preserve StyleX evaluation in source builds

### LINKED ISSUE #5464: Selector's linePad helper breaks any consumer that runs next/babel over the package source
`packages/core/src/Selector/Selector.tsx:398` defines a module-scope arrow helper that builds three `calc()` strings:

```ts
const linePad = (token: string) =>
  `calc((${token} - ${spacingVars['--spacing-5']} - 2 * ${borderVars['--border-width']}) / 2)`;
```

StyleX constant-evaluates the call sites at build time, and that works — until a consumer's toolchain lowers the arrow to a function expression first. Then StyleX cannot evaluate it and the build fails:

```
Unsupported expression: FunctionExpression
```

## How it surfaced

Upgrading an internal Next.js app to 0.5.0. `tsc` was clean; `next build` failed compiling **our own** `Selector.tsx`. Its postcss `babelConfig` sets `babelrc: false` but not `configFile: false`, so the project's `babel.config.js` — with `next/babel` — leaked into the StyleX pass and lowered the arrow.

That app's config is arguably wrong and it fixed itself by adding `configFile: false`. But the failure mode is worth removing at the source: it appears at **build** time, not typecheck, in a file the consumer does not own, with an error naming an expression kind rather than anything they wrote. That is a genuinely hard thing to diagnose from the outside, and any consumer running `next/babel` over the package source hits it.

## Suggested fix

Inline the three `calc()` strings at their call sites. Verified downstream: inlining produces a **byte-identical** extracted stylesheet, so this is a refactor with no rendered change.

The helper exists to keep three sites in step, and the comment above it explains why `--spacing-5` appears — worth keeping that comment on the inlined version, since the reason is not obvious from the arithmetic.

Introduced in [#5202](https://github.com/facebook/astryx/pull/5202), shipped in 0.5.0. `linePad` is the only helper of this shape in `packages/core/src`, so this is a one-off rather than a pattern.

## Not urgent

A consumer can unblock themselves with `configFile: false`, which is what we did. Filing so the next one does not have to.

jiunshinn:
## Summary

- inline Selector's three size-specific `paddingBlock` calculations inside `stylex.create`
- add a regression test that compiles the raw source with `next/babel` and the StyleX Babel plugin in the same production pass
- add a patch changeset for `@astryxdesign/core`

## Root cause

StyleX constant-evaluates values inside `stylex.create`. Selector's module-scope `linePad` arrow helper worked in Astryx's own compiler configuration, but a consumer Babel configuration containing `next/babel` lowered that arrow to a `FunctionExpression` before StyleX evaluated the call sites. StyleX cannot evaluate that transformed expression and aborted the consumer build with `Unsupported expression: FunctionExpression`.

Inlining the three template literals leaves only expressions StyleX supports. The arithmetic and generated CSS are unchanged: pre-fix and post-fix extraction both produce 5,395 bytes with SHA-256 `79d42c676c716c314f018e4d471bd1841c5b394e3e8045ee62e902fb63836999`.

## Validation

- `pnpm vitest run packages/core/src/Selector/Selector.source-build.test.mjs packages/core/src/Selector/Selector.test.tsx --project ui` — 155 passed
- `pnpm -F @astryxdesign/core build` — passed; 6,771 StyleX rules extracted
- `pnpm lint:strict` — passed with 0 errors (80 existing warnings outside this change)
- `pnpm build` — passed
- full `pnpm test` — 12,221 passed, 5 skipped, 3 failed outside this change; the SideNav focus timing failure passed in isolation, while two CLI path-casing assertions reproduce on macOS's case-insensitive filesystem

No screenshots are included because the extracted Selector CSS is byte-identical.

Fixes #5464


cixzhang (approved):
Thanks — the new test fails without the fix, and the extracted stylesheet is byte-identical. Looks good as-is.

[Reviewed by Robohands]

rubyycheung (approved):
Looks good from a design perspective. This preserves the existing Selector output and generated CSS while fixing source-build compatibility, so no visual change is expected.

cixzhang (approved):

cixzhang (approved):

[Reviewed by Robohands]

cixzhang (approved):

[Reviewed by Robohands]

cixzhang (approved):

[Reviewed by Robohands]

cixzhang (approved):

[Reviewed by Robohands]

cixzhang:
/accept-visual 33057973952/2 These frames add Selector coverage; the source-build expression refactor produced byte-identical CSS and a zero-pixel Chromium diff, so the captured visuals are correct.

PR #5509, opened 2026-08-25
fix(theme-stone): raise secondary text contrast

### LINKED ISSUE #5505: stone: --color-text-secondary fails AA in light mode (3.76:1)
`--color-text-secondary` in the stone theme is `['#83838a', '#9d9da3']` (`packages/themes/stone/src/stoneTheme.ts:108`). The light half measures **3.76:1** on the theme's surface — below the 4.5:1 AA threshold for body text.

The theme file's own comment beside it reads `// T55 C=4`, so the ~4:1 value was computed. Nothing records that anyone accepted it *failing* AA.

**Why it matters more than one token:** 83 files under `packages/core/src` reach for `--color-text-secondary`, so every one of them renders sub-AA secondary text for a stone user in light mode. Nothing is wrong at any of those call sites — using the token is correct — which is why the fix belongs in the token, once.

Measured in Chromium across all seven themes; stone light is the only failing cell.

Found while reviewing [#4311](https://github.com/facebook/astryx/pull/4311), which moves Storybook's annotation text from a hardcoded `#666` onto the token. That PR is right and merging: it fixes 6 failing theme/scheme cells and surfaces this one, which predates it. Credit to @HelloOjasMutreja for the change that exposed it.

Same shape as [#5279](https://github.com/facebook/astryx/issues/5279) — the fix is the token pair, not the components.

[Reported by Robohands]

jiunshinn:
## Summary

- Move the Stone `--color-text-secondary` pair from T55/T65 (`#83838a` / `#9d9da3`) to canonical Stone Neutral T40/T70 (`#5e5e63` / `#ababb0`).
- Regenerate the CLI bundled Stone theme so `astryx theme add stone` receives the same fix.
- Add a regression guard that resolves both color modes, composites alpha fills, and checks every surface used by core secondary-text consumers.
- Add the required patch changeset for `@astryxdesign/theme-stone`.

## Root cause and chosen values

The token was tuned as a muted neutral but was not held to the 4.5:1 normal-text floor against the surfaces where core components actually render it. The reported light value is 3.76:1 on `--color-background-surface`. Auditing direct consumers also found explicit pairings with `--color-background-muted` in components such as `Code` and `InputGroupText`; those measured 2.92:1 in light mode and 4.13:1 in dark mode.

T40/T70 is the smallest pair on the canonical five-step Stone neutral ramp that clears every consumer surface. T45 still reaches only 4.17:1 on the light muted surface, and the existing dark T65 reaches only 4.13:1 there. The new worst-case ratios are 5.00:1 in light mode and 4.88:1 in dark mode.

| Pair | Before | After |
| --- | ---: | ---: |
| Light on surface | 3.76:1 | 6.45:1 |
| Light on muted | 2.92:1 | 5.00:1 |
| Dark on surface | 6.37:1 | 7.51:1 |
| Dark on muted | 4.13:1 | 4.88:1 |

`--color-icon-secondary` stays unchanged: it is a non-text token governed by the 3:1 UI-component threshold, matching the precedent in #4557.

## Tests

- `pnpm exec vitest run scripts/check-stone-secondary-text-contrast.test.mjs scripts/check-badge-contrast.test.mjs scripts/check-syntax-punctuation-contrast.test.mjs scripts/check-cli-theme-bundle.test.mjs --project node` — 27 passed
- `pnpm exec vitest run packages/core/src/theme/contrast.test.ts packages/core/src/theme/expandColorScale.test.ts --project ui` — 84 passed
- `pnpm -F @astryxdesign/theme-stone build` — passed
- `pnpm lint:strict` — passed with existing repository warnings and no errors
- `pnpm build` — passed
- `pnpm test` — 12,224 passed; one unrelated TabList focus test flaked in the full parallel run and passed 74/74 in isolation. Two untouched CLI discovery assertions reproduce on this macOS case-insensitive filesystem (`component.test.mjs` case-sensitive lookup and `hook-discovery.test.mjs` path casing).

Fixes #5505

### DIFF
--- a/.changeset/stone-secondary-text-aa.md
+++ b/.changeset/stone-secondary-text-aa.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/theme-stone': patch
+---
+
+[fix] Stone theme: move `--color-text-secondary` to the canonical T40/T70 pair so normal secondary text meets WCAG AA across the theme's light and dark consumer surfaces (#5505)
+@jiunshinn
--- a/packages/cli/assets/templates/themes/stone/stoneTheme.ts
+++ b/packages/cli/assets/templates/themes/stone/stoneTheme.ts
@@ -105,7 +105,9 @@ export const stoneTheme = defineTheme({
 
     // Text — H=291
     '--color-text-primary': ['#25252a', '#f3f3f5'], // light: Stone Neutral T15
-    '--color-text-secondary': ['#83838a', '#9d9da3'], // T55 C=4 / T65 C=3
+    // T40/T70 keeps normal secondary text above AA on every stone surface,
+    // including the muted and tinted fills that components pair it with.
+    '--color-text-secondary': ['#5e5e63', '#ababb0'], // Stone Neutral T40 / T70
     '--color-text-disabled': ['#d7d7da', '#5e5e61'], // T86 C=1.6 / T40 C=2
     '--color-text-accent': ['#25252a', '#f3f3f5'], // light: Stone Neutral T15
     '--color-on-dark': '#FFFFFF',
--- a/packages/themes/stone/src/stoneTheme.ts
+++ b/packages/themes/stone/src/stoneTheme.ts
@@ -105,7 +105,9 @@ export const stoneTheme = defineTheme({
 
     // Text — H=291
     '--color-text-primary': ['#25252a', '#f3f3f5'], // light: Stone Neutral T15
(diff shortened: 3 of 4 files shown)


cixzhang (approved):
Thank you—this fixes Stone’s contrast at the right layer, and the source/CLI copy plus regression guard stay aligned. Looks good to merge.

[Reviewed by Robohands]

PR #5510, opened 2026-08-25
fix(NumberInput): keep stepped values within fractional bounds

### LINKED ISSUE #5506: NumberInput: stepping past a fractional max commits an out-of-range value
### What happens

Stepping with ArrowUp (or the increment button, or the wheel) can commit a value **past `max`** when `max` has more decimal places than `step`.

```tsx
<NumberInput label="Price" max={99.99} value={99} onChange={setValue} />
```

Focus the field at `99`, press ArrowUp: the value commits as **`100`**, which is above `max`. `aria-invalid` is not set, and the increment button does not disable, so nothing signals that the field left its own range.

### Why

`getSteppedValue` clamps to `max` and *then* rounds for floating-point precision:

```
nextValue = Math.min(max, nextValue)          // 99.99
precision = max(decimals(step), decimals(stepBase))  // step 1, stepBase 0 → 0
roundedValue = Number(nextValue.toFixed(precision))  // 100
```

The rounding is there to keep `0.1 + 0.2` from committing `0.30000000000000004`, but it runs after the clamp, so it can round the clamped bound back out of range. The same shape applies at `min` with a fractional lower bound.

### Expected

A stepped value never leaves `[min, max]`. Rounding for precision should happen before the clamp, not after — or the clamp should be re-applied to the rounded result.

### Notes

- Reproduced in Chromium against Storybook, and readable directly in `getSteppedValue` in `packages/core/src/NumberInput/NumberInput.tsx`.
- Independent of the typed-entry path: typing `100` into the same field is bounded correctly.
- No component in this repo passes a fractional bound with an integral step today, but `PowerSearch`'s value editor hands consumer-supplied bounds straight through, so an ordinary money field (`max={99.99}`) reaches it.

jiunshinn:
## Summary

- round snapped step values before applying `min` and `max`
- keep ArrowUp/ArrowDown, wheel, and number-stepper commits inside fractional upper and lower bounds
- keep `aria-valuenow` and the stepper disabled state synchronized when a fractional bound is reached
- add a Core patch changeset

## Root cause

`getSteppedValue` previously clamped the snapped value and then rounded it to the precision derived from `step` and `stepBase`. Rounding is not range-preserving: with `value={99}`, `max={99.99}`, and the default integral step, clamping produced `99.99` but zero-decimal rounding turned it into `100`. The lower side had the same failure when a bound exceeded the precision cap.

The stepping pipeline is now:

`clamp(round(snap(value, step, stepBase)), min, max)`

For a satisfiable interval, clamping is the final numeric transformation, so the emitted value remains in `[min, max]`. Decimal-safe snapping, the existing 12-place precision cap, integer-only validation, and negative-zero normalization are unchanged.

## Testing

- `pnpm exec vitest run packages/core/src/NumberInput/NumberInput.test.tsx` — 139 passed
- regressions cover fractional max and min through keyboard, wheel, and number-stepper interactions
- controlled spinbutton regression verifies the displayed value, `aria-valuenow`, valid state, and disabled increment button at `max`
- `pnpm lint:strict` — zero errors; 80 existing repository warnings
- `pnpm build`
- `pnpm test` — 12,228 passed; the two existing macOS filename-casing failures remain in `packages/cli/api/component/component.test.mjs` and `packages/cli/foundation/discovery/hook-discovery.test.mjs`

Fixes #5506


cixzhang (approved):
Thanks, this is good. The stepped paths now keep fractional min/max in range, including the visible value and stepper state.

[Reviewed by Robohands]

cixzhang (approved):

cixzhang (approved):

[Reviewed by Robohands]

cixzhang (approved):

[Reviewed by Robohands]

cixzhang (approved):

[Reviewed by Robohands]

cixzhang (approved):

[Reviewed by Robohands]

cixzhang:
/accept-visual 33057984539/2 These frames add NumberInput coverage; sensor-backed Chromium review verified fractional-bound stepping, the visible value, and stepper state without unrelated visual movement.

cixzhang:
/accept-visual 33057984539/3 These frames add NumberInput coverage; sensor-backed Chromium review verified fractional-bound stepping, the visible value, and stepper state without unrelated visual movement.

PR #5511, opened 2026-08-25
fix(Slider): ignore modifier-only focus ring activation

### LINKED ISSUE #5469: interactionModality: a bare Shift press counts as keyboard, against its own comment
Split out of [#5463](https://github.com/facebook/astryx/pull/5463), where it was measured but is not that PR's to fix.

`interactionModality.ts` skips modifier-only presses so that "holding Shift before a click must not turn that click into keyboard" — but the guard covers `metaKey`, `altKey` and `ctrlKey` only. `shiftKey` falls through, so a bare Shift press flips the modality.

Measured in Chromium on the Slider thumb after #5463: drag the thumb with the mouse (no ring, correct), then press and hold Shift — the keyboard focus ring appears under the cursor. ⌘C, ⌃C and ⌥C are all correctly ignored.

It affects every consumer of the utility — `PanelSearchInput`, `Selector` (through the search input it renders) and now `Slider`. It was invisible before #5463 because the Slider thumb rang during every drag anyway.

The fix is not obviously "add `shiftKey` to the guard": Shift+Tab *is* keyboard navigation and must keep counting. It probably wants the guard to key off whether the press was modifier-**only**, rather than off which modifiers were held — which is a small change to a shared utility and deserves its own tests across the three consumers.

`packages/core/src/utils/interactionModality.ts:38-46`

jiunshinn:
After a mouse drag, pressing Shift on a Slider thumb restores its keyboard focus ring even though the user has not navigated. Slider now checks the current key and modifier flags before restoring that ring. Bare Shift and Meta/Alt/Control chords keep a pointer-focused thumb unringed, including a chord after Shift has changed the shared modality to keyboard.

The document-wide modality tracker keeps its existing policy: bare Shift counts as keyboard input. Shift+Arrow still adjusts the value and restores the ring, Shift+Tab still restores keyboard focus, and pressing Shift preserves an already visible keyboard ring. The changeset describes the Slider behavior.

Validation:

- Focused Slider, Selector, and interaction-modality tests: 264 passed, including the existing document-wide singleton and listener-lifetime tests.
- `pnpm -F @astryxdesign/core typecheck`: passed.
- `pnpm -F @astryxdesign/core build`: passed.
- `pnpm lint:strict`: passed with 0 errors and 89 warnings outside the changed files.
- Chromium against local Storybook: a real pointer drag followed by bare Shift and Cmd/Ctrl/Alt+C leaves the thumb focused with no outline; Shift+ArrowRight restores a solid outline and increments the value from 55 to 56; subsequent bare Shift preserves the outline. Shift+Tab from the pointer-focused maximum range thumb moves focus and the visible outline to the minimum thumb.

Fixes #5469

### DIFF
--- a/.changeset/interaction-modality-ignores-modifier-only-key-p-jc25.md
+++ b/.changeset/interaction-modality-ignores-modifier-only-key-p-jc25.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Slider keeps its focus ring hidden for modifier-only key presses after a pointer drag while preserving keyboard navigation (#5469)
+@jiunshinn
--- a/packages/core/src/Selector/Selector.test.tsx
+++ b/packages/core/src/Selector/Selector.test.tsx
@@ -3322,6 +3322,35 @@ describe('Selector search focus ring', () => {
     await waitForSearchFocus();
     expect(field()).toHaveAttribute('data-keyboard-focus', 'true');
   });
+
+  it('rings when Shift+Tab returns from clear to the search input', async () => {
+    const user = userEvent.setup();
+    render(
+      <Selector
+        label="Fruit"
+        options={OPTIONS}
+        value={undefined}
+        onChange={() => {}}
+        hasSearch
+      />,
+    );
+    await user.click(screen.getByRole('button', {name: 'Fruit'}));
+    await waitForSearchFocus();
+    const search = screen.getByRole('combobox', {hidden: true});
+    await user.type(search, 'a');
+
+    await user.tab();
+    expect(
+      screen.getByRole('button', {
+        name: 'Clear Search options',
+        hidden: true,
+      }),
+    ).toHaveFocus();
+
+    await user.tab({shift: true});
+    expect(search).toHaveFocus();
+    expect(field()).toHaveAttribute('data-keyboard-focus', 'true');
+  });
 });
 
 describe('Selector search affordances', () => {
--- a/packages/core/src/Slider/Slider.test.tsx
+++ b/packages/core/src/Slider/Slider.test.tsx
@@ -3,7 +3,7 @@
 /**
  * @file Slider.test.tsx
  * @input Uses vitest, @testing-library/react, userEvent, Slider component
- * @output Unit tests for Slider component behavior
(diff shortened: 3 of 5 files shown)


cixzhang (changes requested):
Thanks for tracing this to the focus-ring path. Since this opened, the [shared interaction rule](https://github.com/facebook/astryx/blob/5123978efbcd3f824f88a626cc1fea738faee87a/docs/architecture/interaction-modality.md#L44-L52) was made explicit: every key press without Meta, Alt, or Control sets keyboard modality. Filtering bare Shift in [`interactionModality.ts:45`](https://github.com/facebook/astryx/blob/be74a78c92510ffbf12930b1018e6e0805adc487/packages/core/src/utils/interactionModality.ts#L45) changes that policy for every consumer, while the reported bug is Slider’s ring behavior.

Please keep the shared rule unchanged and solve the Slider ring at Slider’s keyboard handler. The branch is also 425 main commits behind and now conflicts in both interaction-modality files, so please rebase and update the focused tests against the document-wide tracker.

[Reviewed by Robohands]

jiunshinn:
@cixzhang Addressed in 2de0b8b1df07ef1f452d05cf10b5225d5583c7b2. Could you please take another look?

- Rebased onto current upstream main and kept `interactionModality.ts` unchanged, including the document-wide tracker and bare-Shift keyboard policy.
- Moved the focus-ring fix into Slider's keyboard handler. Bare Shift and subsequent Meta/Alt/Control chords keep the ring hidden after a pointer drag; Shift+Arrow and Shift+Tab still restore keyboard focus indication.
- Updated the focused tests against the document-wide tracker: 264 passed. Core typecheck/build and strict lint passed (0 errors, 89 warnings outside the changed files).
- Verified the rendered ring in Chromium: pointer drag → Shift → modifier chords stays unringed; Shift+Arrow restores the ring and changes the value; Shift+Tab moves the ring to the focused range thumb.

cixzhang (approved):
Thanks — this addresses my prior request: the shared modality tracker stays unchanged, and Slider’s pointer, modifier, Shift+Arrow, and Shift+Tab focus-ring paths now pass.

[Reviewed by Robohands]

PR #5512, opened 2026-08-25
build(themes): emit standalone icons as ESM only

### LINKED ISSUE #4949: chore(themes): remove or use the standalone CommonJS icons output
## Problem

All maintained theme packages configure tsup with `entry: ['src/source.ts', 'src/icons.tsx']` and `format: ['cjs', 'esm']`. This emits both `dist/icons.mjs` and `dist/icons.js`.

After #4631, the generated `/built` module imports `./icons.mjs`. The package root CommonJS entry (`dist/source.js`) bundles the icon registry internally, and no package export points to `dist/icons.js`. The standalone CommonJS icon file therefore appears unused. At the same time, the `/built` `require` condition currently points to the same generated ESM-style theme module rather than consuming `icons.js`.

## Proposed investigation

Choose and document one intentional module strategy:

1. **ESM-only icon companion:** configure the icon entry to emit only `icons.mjs`, while keeping the package root entry dual-format; or
2. **Real CommonJS built entry:** produce a CommonJS `/built` artifact that uses `icons.js` and map the `require` condition to it.

Avoid deleting `icons.js` as a post-build cleanup step; the build configuration should express the intended outputs.

## Acceptance criteria

- Confirm whether any supported consumer needs the standalone `dist/icons.js`.
- Either stop emitting the unused file or connect it to a supported CommonJS `/built` path.
- Add Node smoke coverage for both `import` and `require` where those conditions are advertised.
- Keep the output strategy consistent across all maintained themes.

Related: #4620, #4631.

jiunshinn:
## Summary

- split each maintained theme's tsup graph so `source.ts` keeps CommonJS and ESM outputs while `icons.tsx` emits ESM only
- extend post-build export verification to load both advertised Node `import` and `require` targets
- assert every maintained theme produces `dist/icons.mjs` and does not produce the unreachable `dist/icons.js`

## Root cause and invariant

All seven theme configs applied `format: ['cjs', 'esm']` to both entry points. The package export map routes the root `require` condition to `source.js`, where the icon registry is bundled, while the generated `/built` module imports `./icons.mjs`. No export or generated module can reach the standalone CommonJS `icons.js`, so it was emitted and published without a consumer.

The build now preserves this invariant in configuration rather than deleting output afterward: every advertised root condition still has its matching runtime, and the standalone icon companion has exactly the ESM form referenced by `/built`.

## Testing

- `pnpm build` (all maintained themes plus the full repository build)
- `pnpm verify-exports` (31 import targets and 14 require targets loaded)
- packed all seven theme packages and verified `icons.mjs` is present and `icons.js` is absent
- `pnpm lint:strict`
- theme declaration builds/typechecks for all seven packages
- `pnpm -r --if-present typecheck` (vega, charts, core, and richtext pass; existing lab `Drawer.test.tsx` `onClose` type error remains)
- `pnpm test` (12,220 passed; two existing macOS case-insensitive filesystem assertions in CLI component/hook discovery remain reproducible in isolation; the one transient SideNav failure passed in isolation)

Fixes #4949

### DIFF
--- a/.changeset/theme-icons-esm-only.md
+++ b/.changeset/theme-icons-esm-only.md
@@ -1,0 +1,13 @@
+---
+'@astryxdesign/theme-butter': patch
+'@astryxdesign/theme-chocolate': patch
+'@astryxdesign/theme-gothic': patch
+'@astryxdesign/theme-matcha': patch
+'@astryxdesign/theme-neutral': patch
+'@astryxdesign/theme-stone': patch
+'@astryxdesign/theme-y2k': patch
+---
+
+[fix] Theme packages no longer ship an unused CommonJS `icons.js` artifact. Their root entry keeps its advertised CommonJS and ESM outputs, while the standalone icon companion used by `/built` is emitted only as `icons.mjs`.
+
+@jiunshinn
--- a/packages/themes/butter/tsup.config.ts
+++ b/packages/themes/butter/tsup.config.ts
@@ -2,10 +2,21 @@
 
 import {defineConfig} from 'tsup';
 
-export default defineConfig({
-  entry: ['src/source.ts', 'src/icons.tsx'],
-  format: ['cjs', 'esm'],
-  dts: false,
-  clean: false, // Don't clean — astryx theme build already put theme files in dist/
-  external: ['@astryxdesign/core', 'react', 'lucide-react'],
-});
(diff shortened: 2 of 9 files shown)


cixzhang (approved):
Thanks, this looks good. The split preserves every advertised import and require path while dropping only the unreachable CommonJS icon artifacts.

[Reviewed by Robohands]

PR #5513, opened 2026-08-25
fix(core): make BottomSheet text keyboard-scrollable

### LINKED ISSUE #5207: Bottom Sheet: keyboard users cannot scroll the sheet body (axe serious)
**Bug:** a keyboard user cannot scroll the Bottom Sheet's content when that content has no focusable elements.

The scrolling area has no `tabIndex`, so it never takes focus, so arrow keys and Page Down do nothing. A sheet full of plain text is unreachable by keyboard. A sheet with a button inside is fine — focus lands there and scrolls the area.

axe calls this `scrollable-region-focusable`, severity **serious**.

**How it surfaced.** CI's a11y job audits Storybook stories. No Bottom Sheet story had ever rendered *open*, so nothing was ever checked. A story in PR #5203 did open, and the audit failed immediately. I closed that story to unblock the PR — the bug is older and not from that change.

**The fix is not one line.** Adding `tabIndex={0}` to the scroll body adds a tab stop to *every* sheet, including ones that do not need it. It probably needs to be conditional on the content having no focusable children, which is a runtime check and wants its own review.

**Also worth fixing:** at least one Bottom Sheet story should render open, or CI will keep missing this whole component.

Files: `packages/core/src/BottomSheet/BottomSheetPanel.tsx` (the `styles.body` div).

Rule: https://dequeuniversity.com/rules/axe/4.12/scrollable-region-focusable

jiunshinn:
## Summary

- make the BottomSheet scroll body a tab stop only when it has no visible focusable descendant
- keep that decision synchronized for lazy, nested, enabled, disabled, hidden, and removed controls while preserving an SSR-safe initial tab stop
- short-circuit boolean focusability checks at the first visible control so mutation observation stays cheap for large virtualized sheets
- add focused regressions and an open-by-default text-only Storybook case so the scoped axe audit exercises the rendered sheet

## Testing

- `pnpm exec vitest run packages/core/src/DateInput/DateInputTouch.test.tsx -t "does not report a scroll it was told to make"` (11 consecutive passes after the short-circuit fix)
- `pnpm exec vitest run packages/core/src/DateInput/DateInputTouch.test.tsx` (135 passed)
- `pnpm exec vitest run packages/core/src/BottomSheet packages/core/src/hooks/useFocusTrap.test.tsx` (227 passed)
- `pnpm lint:strict` (passes; existing warnings only)
- `pnpm build`
- `pnpm storybook:build`
- `pnpm a11y:audit -- --storybook-dir apps/storybook/dist --components BottomSheet --output /tmp/astryx-5207-a11y.json` (11 stories, 0 violations)
- Chromium keyboard check: text-only body focused with `tabindex="0"`; Page Down moved `scrollTop` from 0 to 317; control-bearing story retained no body `tabindex`

Fixes #5207

### DIFF
--- a/.changeset/bottom-sheet-keyboard-scroll.md
+++ b/.changeset/bottom-sheet-keyboard-scroll.md
@@ -2,5 +2,6 @@
 '@astryxdesign/core': patch
 ---
 
-[fix] Make text-only Bottom Sheets keyboard-scrollable (#5207)
+[fix] Make text-only Bottom Sheets keyboard-scrollable: the scroll body becomes a Tab stop only while it actually overflows and contains no visible sequentially focusable control, so short sheets never gain a dead stop and sheets whose only controls are CSS-hidden or tabindex="-1" keep one (#5207)
+
 @jiunshinn
--- a/.changeset/bottom-sheet-keyboard-scroll.md
+++ b/.changeset/bottom-sheet-keyboard-scroll.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Make text-only Bottom Sheets keyboard-scrollable (#5207)
+@jiunshinn
--- a/apps/storybook/stories/BottomSheet.stories.tsx
+++ b/apps/storybook/stories/BottomSheet.stories.tsx
@@ -240,6 +240,44 @@ export const TallSheet: Story = {
   },
 };
 
+export const KeyboardScrollableText: Story = {
+  name: 'Keyboard scroll — Text only',
+  render: () => {
+    const [isOpen, setIsOpen] = useState(true);
+    return (
+      <>
+        <Button label="Open release notes" onClick={() => setIsOpen(true)} />
+        <BottomSheet
+          isOpen={isOpen}
+          onOpenChange={setIsOpen}
+          label="Release notes"
+          height="capped">
+          <Section padding={4}>
+            <VStack gap={3}>
+              <Heading level={3}>Release notes</Heading>
+              <Text type="supporting" color="secondary">
+                This sheet opens for the accessibility audit. Its body has no
+                controls, so keyboard users reach the body itself and scroll it
+                with arrow keys or Page Down.
+              </Text>
+              <Divider />
+              {Array.from({length: 18}, (_, i) => (
+                <VStack key={i} gap={1}>
+                  <Text type="label">Update {i + 1}</Text>
+                  <Text type="supporting" color="secondary">
+                    Details about this release remain readable without a mouse
+                    or touch gesture.
+                  </Text>
+                </VStack>
+              ))}
+            </VStack>
+          </Section>
+        </BottomSheet>
+      </>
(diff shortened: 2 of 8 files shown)


cixzhang (changes requested):
Thanks for covering the dynamic-content cases. Two keyboard paths still break:

- A short sheet becomes a dead Tab stop even though it cannot scroll (`BottomSheetPanel.tsx:510-557`).
- A scrollable sheet with only CSS-hidden or negative-tabindex controls loses its explicit body stop, so the fix misses engines that need it (`focusableSelector.ts:23-42`).

Could the condition require actual overflow and no visible sequential-focus descendant?

[Reviewed by Robohands]

jiunshinn:
Thanks for catching both paths — the condition now requires actual overflow and no visible sequential-focus descendant, exactly as suggested.

- Dead Tab stop on a short sheet: the body now earns its tab stop only when `scrollHeight > clientHeight` (BottomSheetPanel.tsx). The overflow read runs first as a constant-time short-circuit, and it stays live through the existing MutationObserver (now including characterData) plus a ResizeObserver on the body and its direct children, so content growing or shrinking without a DOM mutation still toggles the stop. The SSR-conservative initial stop is unchanged and reconciles before first paint.
- Lost body stop with only CSS-hidden or negative-tabindex controls: the shared filter in focusableSelector.ts now rejects any explicit negative tabindex (natives like `<button tabindex="-1">` previously slipped through their own selector clause) and walks the ancestor chain for `display: none`, which never appears in a descendant's own computed style; `visibility` remains an own-element check since it inherits and a descendant can restore it.

Added regressions for all four cases: a short non-scrollable sheet has no tabindex stop; a scrollable sheet keeps its stop when its only control is inside `display: none` or carries `tabindex="-1"`; and ResizeObserver-reported overflow changes add and remove the stop. Scroll metrics are mocked via prototype getters since jsdom performs no layout. Full BottomSheet + useFocusTrap suites pass (231/231).

cixzhang:
Thanks for fixing the keyboard-scroll path and covering overflow and hidden controls. I verified this head reaches and scrolls the sheet body. We’re taking #5553 because it preserves existing overlay/input focus behavior and keeps the full suite green while crediting your work.

[Reviewed by Robohands]

jiun-shin95:
Closing in favor of #5553 — generalizing the tab-stop ownership into `useScrollableTabStop` is the better home for this, and it keeps the overlay/input focus behavior intact in a way this head could not. Glad the overflow + visible-sequential-focus-descendant conditions proved out, and thanks for verifying the head and for the credit. Happy to see #5207 land through the hook.

jiunshinn:
Closing in favor of #5553 — generalizing the tab-stop ownership into `useScrollableTabStop` is the better home for this, and it keeps the overlay/input focus behavior intact in a way this head could not. Glad the overflow + visible-sequential-focus-descendant conditions proved out, and thanks for verifying the head and for the credit. Happy to see #5207 land through the hook.

Issue #5406, opened 2026-08-24
Markdown streaming: every chunk re-scans the whole document, so a long response is still superlinear to stream


cixzhang:
`parseMarkdownIncremental` now keeps its settled blocks at every document
length — after [#5378](https://github.com/facebook/astryx/issues/5378) the worst
chunk of a streamed document rebuilds 3 blocks whether the document is 50
paragraphs or 500. Streaming is still superlinear anyway, because four things in
the function touch the **whole input** on every chunk, cache hit or not:

| `parser.ts` | per chunk |
| --- | --- |
| `extractLinkDefinitions(input)` `:1958` | scans the whole document for `[ref]:` lines |
| `input.split('\n')` `:1970` | allocates an array of every line |
| `findSettledBoundary(lines)` `:1971` | walks every line, tracking fences from the top |
| `mergeSettledBlocks(...)` `:2033` | copies the whole block array |

Measured with the perf suite's `generateAIResponse` fixture, 50-char chunks,
best of 5, on `main` at `1b0fab89661`:

| paragraphs | chars | whole stream | cost of the last chunk |
| --- | --- | --- | --- |
| 50 | 6.6k | 3.3ms | 0.033ms |
| 200 | 26k | 40.3ms | 0.130ms |
| 500 | 66k | 240.0ms | 0.388ms |
| 1000 | 132k | 1190.0ms | 1.236ms |

Doubling the document from 500 to 1000 paragraphs multiplies the cost of
streaming it by 5.

**What a person experiences.** No single token is slow — 1.2ms at two pages is
invisible next to a frame — so this is not the stutter #5378 was. It is main-
thread work that grows with the response: 1.2s of parsing spread across a
two-page answer today, four times that at four pages, on a device that may also
be rendering it.

**Shape of a fix.** The settled prefix is immutable by construction, so all four
of these can work on the tail: keep the line count and fence state for the
settled prefix in `IncrementalState` and scan only from the boundary; keep the
link-definition signature per settled prefix and extract only from the tail; and
hand back the previously-returned array when the settled blocks are reused
unchanged. That is a change to what `IncrementalState` carries, not a repair —
worth its own PR and its own before/after on the table above.

The invariant to assert is the one [#5379](https://github.com/facebook/astryx/pull/5379)
established for the cache: an integer that stays flat across a 10x document, not
a millisecond budget. Here it would be the cost of one chunk relative to the
document length.

PR #5515, opened 2026-08-25
perf(markdown): bound incremental parsing to stream tail

jiunshinn:
## Summary

- retain the settled character boundary, link-definition layers, and returned block array in a private incremental cache
- split, scan fences/boundaries, and collect definitions only from the mutable tail
- append newly settled blocks to the state-owned array and assemble each returned array as a fresh per-call snapshot (one pointer per block), keeping Markdown heading IDs and fade boundaries correct
- preserve global first-definition-wins behavior, source ranges, loose-list merging, and open fences across chunks

## Root cause and invariant

`parseMarkdownIncremental` cached parsed blocks, but four operations still traversed the cumulative input on every chunk: link-definition extraction, `split('\n')`, settled-boundary/fence detection, and output-array assembly. For a stream delivered in fixed-size chunks, repeating work proportional to the already-settled prefix produces a quadratic total even when only 1–3 blocks are parsed per chunk.

The cache now maintains this invariant for ordinary chunks, scoped to the four measured parser operations:

> Splitting, settled-boundary/fence detection, link-definition collection, and per-chunk block construction are bounded by the mutable suffix after the last settled boundary, independent of the immutable prefix length.

The deterministic regression streams 20 and 200 fixed-size sections and observes the same worst operation counts in both cases:

| operation | 20 sections | 200 sections |
| --- | ---: | ---: |
| split characters | 71 | 71 |
| boundary lines | 5 | 5 |
| definition characters | 71 | 71 |
| rendered suffix blocks | 1 | 1 |

A newly appearing or disappearing document-global link definition remains an intentional exception: it can change references in already-settled blocks, so those blocks must be re-parsed for correctness. Promotion of an unchanged definition from tail to settled state does not invalidate them.

Two whole-input costs remain intentionally out of this PR's scope and stay linear per call: the settled-prefix `startsWith` comparison that detects a replaced document, and the one-pointer-per-block copy that keeps each returned array a stable snapshot. Both are required by the replacement/snapshot contract preserved in the previous review round — replacement correctness cannot be decided without reading the prefix, and snapshot semantics require a fresh array per call. The regression suite now tracks their exact totals (13,072/1,372,702 prefix characters inspected and 210/20,100 entries copied across 20/200 sections) without bounding them, so the remaining costs of #5406 stay visible.

## Benchmark

Best of 5, Node 24, 50-character chunks, same mixed Markdown fixture on the same machine:

| paragraphs | chars | main total | this PR total | main last chunk | this PR last chunk |
| ---: | ---: | ---: | ---: | ---: | ---: |
| 50 | 6,607 | 4.992 ms | 1.506 ms | 0.0545 ms | 0.0025 ms |
| 200 | 26,351 | 65.918 ms | 4.864 ms | 0.2174 ms | 0.0030 ms |
| 500 | 65,951 | 384.394 ms | 10.113 ms | 0.5125 ms | 0.0025 ms |
| 1,000 | 131,915 | 1,490.782 ms | 18.937 ms | 1.0600 ms | 0.0083 ms |

These timings illustrate the change; the operation-count invariant above is the non-flaky regression gate.

## Tests

- `pnpm vitest run packages/core/src/Markdown` — 281 passed
- `pnpm vitest run packages/core/src/Markdown/parser.perf.test.ts --reporter=verbose` — 16 passed
- `pnpm lint:strict`
- `pnpm build`

The repository-wide `pnpm test` completed with 12,224 passing tests and 3 unrelated failures in existing CLI case-sensitive discovery/file naming checks and a SideNav focus-timing test; all Markdown projects passed.

References #5406 — this PR bounds the four parsing-side operations; the whole-prefix comparison and per-call snapshot copy described above are intentionally left linear and tracked, so the issue stays open for that remainder.

### DIFF
--- a/.changeset/markdown-stream-tail-cache.md
+++ b/.changeset/markdown-stream-tail-cache.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[perf] Markdown streaming now scans and replaces only the mutable tail of an incremental parse. Splitting, fence detection, link-definition collection, and result assembly no longer grow with the already-settled document.
+
+@jiunshinn
--- a/.changeset/markdown-stream-tail-cache.md
+++ b/.changeset/markdown-stream-tail-cache.md
@@ -2,6 +2,6 @@
 '@astryxdesign/core': patch
 ---
 
-[perf] Markdown streaming now parses only the mutable tail of an incremental parse. Splitting, fence detection, link-definition collection, and block re-parsing no longer grow with the already-settled document. The parser contract is unchanged: each call returns a fresh, never-mutated snapshot, and replacing already-settled text still re-parses the document.
+[perf] Markdown streaming bounds four incremental-parse operations by the mutable tail: splitting, fence/boundary detection, link-definition collection, and block re-parsing no longer grow with the already-settled document. The parser contract is unchanged: each call returns a fresh, never-mutated snapshot, and replacing already-settled text still re-parses the document. Two costs intentionally remain proportional to the whole input on each call because that contract requires them — the settled-prefix comparison that detects a replaced document, and the pointer-per-block copy behind each returned snapshot.
 
 @jiunshinn
--- a/.changeset/markdown-stream-tail-cache.md
+++ b/.changeset/markdown-stream-tail-cache.md
@@ -2,6 +2,6 @@
 '@astryxdesign/core': patch
 ---
 
-[perf] Markdown streaming now scans and replaces only the mutable tail of an incremental parse. Splitting, fence detection, link-definition collection, and result assembly no longer grow with the already-settled document.
+[perf] Markdown streaming now parses only the mutable tail of an incremental parse. Splitting, fence detection, link-definition collection, and block re-parsing no longer grow with the already-settled document. The parser contract is unchanged: each call returns a fresh, never-mutated snapshot, and replacing already-settled text still re-parses the document.
 
 @jiunshinn
--- a/packages/core/src/Markdown/Markdown.tsx
+++ b/packages/core/src/Markdown/Markdown.tsx
@@ -1677,6 +1677,9 @@ export function Markdown({
   // NOTE: must stay above the `display === 'inline'` early return below —
   // hooks cannot be conditional.
   const headingIdMap = useMemo(() => {
+    // Incremental parsing keeps the blocks array stable while replacing its
+    // tail, so streamed text is the version signal for this derived map.
+    void smoothedText;
     if (display === 'inline' || blocks.length === 0) {
       return undefined;
     }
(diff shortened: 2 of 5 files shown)


cixzhang (changes requested):
Thanks—the append-only path is convincingly bounded, but this changes the public parser contract without a breaking changeset. Reusing state after a same-length or longer replacement keeps the old settled prefix (confirmed in Chromium: new `Bravo/New body` args still render `Alpha/Old body`), and a later call now mutates every earlier returned AST. Both pass on the parent and fail here. Could we restore replacement correctness and either preserve snapshot values or mark/document the mutation as breaking?

[Reviewed by Robohands]

jiunshinn:
Thanks for the careful review — both counts were real contract regressions and are fixed in 9df5c4d, restoring the parent contract rather than documenting a break (changesets here are patch-only pre-1.0).

Replacement correctness: the reset check compared lengths, so a same-length or longer replacement kept the old settled prefix. It now compares content — `input.startsWith(state.settledText)` — and a mismatch discards the cache and re-parses in full, so your Bravo/New body repro renders the new args. The prefix comparison is the one whole-prefix operation retained per call: memcmp-speed, no allocation, and required to detect a replaced document.

Snapshot semantics: the state-owned result array is gone. Every call returns a fresh array again (including the empty-input path), assembled by copying one pointer per settled block; the immutable settled block objects are reused across calls by reference and are replaced, never edited, when merging touches them. So a later call no longer mutates any earlier returned AST. Markdown.tsx reverts to the parent implementation since its changes only existed to work around the mutation.

Regression tests mirror both repros in incremental.test.ts: same-length and longer settled-text replacements must deep-equal the full parse of the new content, and a structuredClone snapshot of call 1's return must be unchanged (and not identity-shared) after an appended call 2. I verified all three pass against the parent parser as well.

The perf invariant is intact: the operation counters stay flat across a 10x document (worst tail work identical at 20 and 200 sections), with a comment in parser.perf.test.ts noting the two deliberate whole-prefix but non-parsing costs (the prefix memcmp and the pointer-per-block result copy). Full Markdown suite: 280 passed.

cixzhang (changes requested):
Thanks for preserving replacement and snapshot behavior. One invariant still does not match the code: `parser.perf.test.ts:251` says every whole-input operation is bounded, but it excludes the settled-prefix `startsWith` check and fresh-array copy.

Across 20/200 fixed sections, those paths inspect 13,072/1,372,702 prefix characters and copy 210/20,100 returned entries. The narrower parsing improvement looks safe, but `Fixes #5406` would close the broader rescan/assembly issue while these costs remain. Could you either narrow the test and PR claim to the four measured parser operations and leave the remainder tracked, or include these two costs while preserving v0.5.2 snapshot/replacement behavior?

[Reviewed by Robohands]

cixzhang (approved):
Thanks—this addresses our prior ask: four parser operations are now bounded, while the replacement check, snapshot copy, and compatibility remain explicit and tested.

[Reviewed by Robohands]

PR #5516, opened 2026-08-25
fix(core): ensure pressed overlays override hover

### LINKED ISSUE #5451: [Bug] The pressed overlay never paints: a media-nested :hover out-specifies :active library-wide

jiunshinn:
## Summary

- centralize paired hover and pressed overlays in one internal StyleX utility and migrate every existing direct consumer, including Button, AvatarGroupOverflow, Calendar, navigation, selector, item, and stepper surfaces
- repeat `:active` inside the `(hover: hover)` branch so it ties the guarded hover selector on generated specificity and wins by StyleX pseudo-state order
- retain a bare enabled `:active` branch for touch input, where `(hover: hover)` does not match
- add generated-CSS assertions, a Chromium cascade guard, and a core patch changeset

## Root cause and event semantics

During a pointer press Chromium matches both `:hover` and `:active`. StyleX puts the media-nested hover rule in a higher priority bucket, which compiled to a selector with more repeated classes than the bare active rule. The cascade therefore kept painting the hover token throughout the press.

The shared utility now emits hover and active inside the same hover-capable media branch. Both generated selectors have specificity `(3,4,0)`, and the active rule is emitted later. A separate enabled active selector remains outside the media query for coarse/touch input. The enabled guard stays outermost, so disabled and `aria-disabled` surfaces do not acquire either overlay.

## Verification

- `pnpm exec vitest run scripts/build-css.test.mjs` — 8 passed
- focused core suites across Button, AvatarGroup, Calendar, command palette, selectors, Item, NumberInput, Token, DateInput touch, Outline, SideNav, Stepper, TopNav, and TreeList — 18 files / 1,238 tests passed
- `pnpm guard:pressed-overlay` — Chromium verified hover, simultaneous hover+active, neutral base composition, disabled hover, and touch active fallback
- rendered Storybook Button in Chromium — simultaneous hover+active computed the pressed `rgba(0, 0, 0, 0.1)` overlay instead of hover `rgba(0, 0, 0, 0.05)`
- `pnpm -F @astryxdesign/core lint` — 0 errors; 49 pre-existing warnings
- commit hooks: sync, package boundaries, changesets, demo media, executable bits, CLI structure, use-client, portable scripts, i18n catalog, and CLDR weekdays

Fixes #5451


rubyycheung (approved):
Looks good from design. The interaction now communicates a clear rest → hover → pressed progression, with the pressed state visibly overriding hover as expected. Applying the same treatment across core interactive surfaces improves consistency while preserving touch and disabled behavior.

cixzhang (approved):
Thanks for fixing this in the shared layer. Chromium confirms pressed now wins while disabled, touch, and pseudo-element states stay intact; source/tests cover selected and focus.

[Reviewed by Robohands]

cixzhang (approved):

cixzhang (approved):

[Reviewed by Robohands]

cixzhang (approved):

[Reviewed by Robohands]

cixzhang (approved):

[Reviewed by Robohands]

cixzhang (approved):

[Reviewed by Robohands]

PR #5517, opened 2026-08-25
ci: resolve fork PRs by head repo and branch in pr-comment

bhamodi:
## Summary

(Scope updated after #5572/#5594 landed the trusted-identity work on main — this now carries only the two deltas from review.) The trusted-identity resolution on main falls back to `commits/{sha}/pulls`, which only indexes base-repo branches: a fork head returns `[]`, and `workflow_run.pull_requests` is empty on forks too, so every fork PR fails resolution and loses its analysis comment. And the comment generator is still invoked through a string-built `execSync`.

## Changes

- Both resolve steps (the `resolve` job and the `comment` job's identity step) fall back to `pulls.list({head: "<head_repository.owner.login>:<head_branch>", state: 'open'})` instead of `commits/{sha}/pulls`. The existing exact-SHA, head-repo full-name, and base-repo checks are untouched and still pin the result to this run.
- `generate-pr-comment.js` runs via `execFileSync('node', args)` — argument vector, no shell.

## Test plan

- Workflow YAML parses (js-yaml); the modified github-script bodies pass `node --check` after extraction.
- Verified against live data: `pulls.list` with `head=<owner>:<branch>` finds this PR (a fork PR), where `commits/{sha}/pulls` returns `[]` — matching the reviewer's measurements across three fork PRs.
- Same-repo PRs are unaffected: `run.pull_requests` is populated for them and the fallback never runs.

## Notes

- Same lookup shape as `review-clear.yml`'s existing fork resolution; #5518 follows the same pattern for preview deploys.

### DIFF
--- a/.github/workflows/pr-comment.yml
+++ b/.github/workflows/pr-comment.yml
@@ -51,11 +51,16 @@ jobs:
             const {owner, repo} = context.repo;
             let numbers = (run.pull_requests || []).map((pr) => pr.number);
             if (numbers.length === 0) {
-              const associated = await github.paginate(
-                github.rest.repos.listPullRequestsAssociatedWithCommit,
-                {owner, repo, commit_sha: run.head_sha, per_page: 100},
-              );
-              numbers = associated.filter((pr) => pr.state === 'open').map((pr) => pr.number);
+              // commits/{sha}/pulls only indexes base-repo branches, so a fork
+              // head returns [] (and run.pull_requests is empty on forks too).
+              // Resolve by the run's own head repo + branch instead; the
+              // exact-SHA check below keeps it pinned to this run.
+              const {data: byHead} = await github.rest.pulls.list({
+                owner, repo, state: 'open',
+                head: `${run.head_repository.owner.login}:${run.head_branch}`,
+                per_page: 100,
+              });
+              numbers = byHead.map((pr) => pr.number);
             }
             numbers = [...new Set(numbers)];
             if (numbers.length !== 1) {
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-comment.yml
```diff
@@ -71,23 +74,43 @@ jobs:
...
-            const prNumber = parseInt(meta.prNumber, 10);
+            // Resolve the PR from the workflow_run's head commit. For fork PRs
+            // workflow_run.pull_requests is empty, so look the PR up by SHA and
+            // require an exact head match — the run's own payload is the source
+            // of truth for which PR this report belongs to.
+            const run = context.payload.workflow_run;
+            const { owner, repo } = context.repo;
+            let prNumber = run.pull_requests?.find((p) => p.head.sha === run.head_sha)?.number;
+            if (!prNumber) {
+              const { data: prs } = await github.rest.repos.listPullRequestsAssociatedWithCommit({
```
cixzhang: Hmm, this returns `[]` for a fork head, and `workflow_run.pull_requests` is empty there too. `?head=owner:branch` matches.

cixzhang (changes requested):
Thanks for this — the shell hole was real, and `execFileSync` is the right fix.

The PR lookup doesn't reach fork PRs. `GET /commits/{sha}/pulls` only matches head branches in this repo, so a fork head returns `[]` — I ran it against three open fork PRs including this one. `workflow_run.pull_requests` is empty on forks too, so the job warns and returns: no analysis comment on any fork PR. That's 98 of the open ones today.

Two ways out. `deploy-preview.yml:70` keeps reading `prNumber` from the artifact and validates it with `^[0-9]+$` — digits can't reach a shell, and it stays exact. Or resolve it from the run's own fork and branch:

```js
head: `${run.head_repository.owner.login}:${run.head_branch}`
```

#5518 has the same lookup and can follow this one. Which would you rather?

[Full review](https://github.com/cixzhang/astryx/wiki/Review-5517)

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks — I rechecked the exact head. My earlier fork-lookup blocker is unchanged: this run has `pull_requests: []`, and `GET /commits/b412…/pulls` also returns `[]`, so every fork PR loses its analysis comment.

Main's [#5572](https://github.com/facebook/astryx/pull/5572) now conflicts with this branch and added trusted identity plus visual publishing. Could you merge main, preserve those steps, switch the generator to `execFileSync`, and resolve forks with `head_repository.owner.login:head_branch` plus the existing exact-SHA check? That query finds this PR.

[Reviewed by Robohands]

bhamodi:
Went with your second option in both resolve steps — `pulls.list` with `head: `${run.head_repository.owner.login}:${run.head_branch}`` plus the existing exact-SHA/head-repo/base-repo checks — and verified the query finds this PR where `commits/{sha}/pulls` returns []. Rebased over #5572/#5594, which made most of my original diff redundant (nice to see the derive-from-context direction land); what remains here is just the fork lookup in both steps and `execFileSync` for the generator. PR title/body updated to match the narrower scope.

.github/workflows/pr-comment.yml
bhamodi: Switched to `pulls.list({head: `${run.head_repository.owner.login}:${run.head_branch}`, state: 'open'})` in both resolve steps — confirmed it matches this PR where both `commits/{sha}/pulls` and `run.pull_requests` come back empty.

cixzhang (approved):
Thanks — this resolves the fork lookup and shell invocation I asked for. The exact-head checks and main’s visual path are preserved. Good to merge.

[Reviewed by Robohands]

PR #5518, opened 2026-08-25
ci: resolve the preview deploy target from the CI run's head commit

bhamodi:
## Summary

The preview deploy read the PR number and artifact hash out of the `pr-analysis` artifact CI uploaded. The `workflow_run` payload already carries the head commit that run built, so this derives the artifact hash from the payload and resolves the PR by exact head-SHA match via the API. The artifact is now only a "this run built previews" signal.

## Changes

- Derive `short_hash` from `workflow_run.head_sha` (the same commit ci.yml names its `storybook-<short>` / `sandbox-<short>` artifacts after).
- Resolve `pr_number` via the commits→pulls API, requiring the PR's current head to equal the run's head SHA. No exact match (branch force-pushed or advanced since the run) means a clean skip — the newer run's deploy wins instead of a stale preview overwriting it.
- Drop the jq parsing and format validation of `pr-meta.json` fields; the digits/hex guarantees now hold by construction.
- Update the header/concurrency comments to describe the new derivation.

## Test plan

- Workflow YAML parses (js-yaml); step wiring verified (`steps.parse.outputs.*` consumers unchanged).
- Confirmed `workflow_run.head_sha` equals `github.event.pull_request.head.sha` for pull_request-triggered runs, which is exactly what ci.yml's artifact names and `pr-meta.json` derive from — so the resolved values are identical to the previous ones for every normal run.
- Skip paths verified by inspection: missing artifact → quiet exit (unchanged); no head-SHA match → warning + green skip, mirroring the existing expired-artifact behavior.

## Notes

- `pr-meta.json` is still uploaded by ci.yml and still read by pr-comment (see #5517, which moves that workflow to the same payload-derived inputs).

### DIFF
--- a/.github/workflows/deploy-preview.yml
+++ b/.github/workflows/deploy-preview.yml
@@ -12,7 +12,10 @@
 # Safety: this NEVER checks out or runs PR-controlled code. It only downloads
 # the pre-built static artifacts (storybook-<hash>, sandbox-<hash>) that CI
 # already produced and copies them into gh-pages. Do not add checkout of the PR
-# head or execution of PR code here.
+# head or execution of PR code here. The deploy target is derived from the
+# workflow_run payload: the PR number is resolved from the run's head commit
+# and the artifacts are addressed by that commit's hash, so artifact contents
+# never decide where a preview lands.
 
 name: Deploy PR Preview
 
@@ -26,8 +29,8 @@ permissions: {}
 # Serialize preview deploys per source branch (≈ per PR). Two pushes to the same
 # PR shouldn't race each other's gh-pages push; cross-PR and main-deploy conflicts
 # are handled by the retry loop below. head_branch is the only PR-stable key
-# available in workflow_run context (the PR number isn't known until we parse the
-# artifact). Not cancel-in-progress: let an in-flight deploy finish its push.
+# available at concurrency-group time (the PR number is resolved inside the job).
+# Not cancel-in-progress: let an in-flight deploy finish its push.
 concurrency:
   group: "pr-preview-${{ github.event.workflow_run.head_branch }}"
   cancel-in-progress: false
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-preview.yml
```diff
@@ -51,29 +54,37 @@ jobs:
...
-          SHORT_HASH="$(jq -r '.shortHash // ""' pr-analysis/pr-meta.json)"
+          # The artifacts are named after the head commit that CI built
+          # (storybook-<short>, sandbox-<short>), which for a pull_request run
+          # is exactly workflow_run.head_sha.
+          SHORT_HASH="${HEAD_SHA:0:7}"
+
+          # Resolve the PR whose current head is this run's head commit. If the
+          # branch moved on since (force-push, new commits), there is no exact
+          # match and this run's preview is stale — skip; the newer run deploys.
+          PR_NUMBER="$(gh api "repos/${{ github.repository }}/commits/${HEAD_SHA}/pulls" \
```
cixzhang: Returns `[]` for a fork head; `review-clear.yml:72` resolves by `head_repository` owner + `head_branch` instead.

.github/workflows/deploy-preview.yml
```diff
@@ -51,29 +54,37 @@ jobs:
...
+            --jq "[.[] | select(.head.sha == \"${HEAD_SHA}\")][0].number // empty")"
 
-          # Validate PR number is digits and hash is hex — these name a
-          # gh-pages path and artifacts, so reject anything unexpected.
           if ! echo "$PR_NUMBER" | grep -qE '^[0-9]+$'; then
```
cixzhang, line 89: This used to be a hard error. Now an unresolved PR is a green skip and nothing surfaces it.

cixzhang (changes requested):
Thanks — the hardening instinct is right, and the hash half is correct.

The PR lookup doesn't survive a fork, though. `commits/<sha>/pulls` only indexes commits on a base-repo branch, so a fork head comes back `[]` — [#5533](https://github.com/facebook/astryx/pull/5533) and [#5515](https://github.com/facebook/astryx/pull/5515), two different forks, both empty, both with previews live right now. Thirteen of the last eighteen preview directories on `gh-pages` are fork PRs, and every one would take the new `deploy=false` path and finish green, so a contributor's preview link 404s with nothing saying why.

`review-clear.yml:59` already wrote this down; `:72` is the shape that works — resolve from `head_repository.owner.login` + `head_branch`, then compare `head.sha` for your staleness guard. Worth putting the reason in the body while you're there.

Does resolving by branch still give you the guard you were after? If you'd rather talk it through with someone, we're in [Discord](https://discord.com/invite/XnsUcFykEP).

[Full review](https://github.com/cixzhang/astryx/wiki/Review-5518)

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks — I rechecked the exact head. My earlier fork-lookup blocker is unchanged: this run has `pull_requests: []`, and `GET /commits/b412…/pulls` also returns `[]`, so every fork PR loses its analysis comment.

Main's [#5572](https://github.com/facebook/astryx/pull/5572) now conflicts with this branch and added trusted identity plus visual publishing. Could you merge main, preserve those steps, switch the generator to `execFileSync`, and resolve forks with `head_repository.owner.login:head_branch` plus the existing exact-SHA check? That query finds this PR.

[Reviewed by Robohands]

.github/workflows/deploy-preview.yml
bhamodi: Switched to the review-clear.yml shape: `pulls?head=${head_repository.owner.login}:${head_branch}&state=open`, then the staleness guard compares the PR's current `head.sha` to the run's head. Verified live that the head query finds this PR (a fork) where `commits/{sha}/pulls` returns []. And yes — resolving by branch keeps the guard: the exact-SHA comparison still rejects a run whose branch moved on, it just compares against the PR the run actually belongs to.

.github/workflows/deploy-preview.yml
bhamodi: Restored, with the two cases separated: a run that BUILT previews but whose PR can't be resolved is a bug and now fails red with the owner:branch and head SHA in the message; a resolved PR whose head moved on (force-push, new commits) still skips green with a warning, since the newer run's deploy supersedes it. The reason for the branch-based lookup is now in the step comment too.

cixzhang (approved):
Thanks — the fork lookup and unmatched-run failure are fixed, while the exact-SHA guard still holds. One nit: the body still describes the superseded lookup and green skip.

[Reviewed by Robohands]

bhamodi:
Closing as superseded by the preview-pipeline rework on main (#5572/#5594/#5697). `deploy-preview.yml` is now a `workflow_call` trusted publisher: `pr-comment.yml` resolves the PR from the run's head repo + branch (`pulls?head=owner:branch`), and `identityFromPullAndRun` refuses any mismatch of head SHA, ref, repository, or repository id against the source run before artifacts are downloaded from that exact run. That is a strict superset of what this PR added on the old `workflow_run` flow (head-commit PR resolution + head-SHA staleness skip), so there is nothing left for this branch to contribute.

PR #5519, opened 2026-08-25
ci: re-deploy previews from CI artifacts instead of rebuilding

bhamodi:
## Summary

Re-deploy Preview checked out the PR head and ran the full monorepo build inside the deploy job — a second build pipeline that has to be kept in sync with ci.yml (Node/pnpm setup, base-path env, build order). deploy-preview.yml already publishes previews by copying the static `storybook-<hash>`/`sandbox-<hash>` artifacts CI built, so this makes the manual re-deploy do exactly the same thing.

## Changes

- Replace the checkout + `pnpm build` + Storybook/Sandbox build steps with: resolve the PR's current head SHA, locate its latest CI run (`actions/runs?head_sha=...` filtered to ci.yml), and download that run's preview artifacts.
- Keep the deploy step byte-identical to deploy-preview.yml's (retry-based gh-pages push, artifact copy only).
- Validate the operator-typed PR number is numeric before it names a gh-pages path.
- Expired/missing artifacts now fail loudly with the actionable fix: re-run CI on the PR, which re-publishes the preview automatically via deploy-preview.yml.
- Runner drops from `2-core-ubuntu-arm` to `ubuntu-slim` — the job no longer builds anything.

## Test plan

- Workflow YAML parses (js-yaml); step outputs wiring checked (`pr-info` → downloads → verify → deploy).
- Confirmed the artifact names match ci.yml's uploads (`storybook-${short_hash}` / `sandbox-${short_hash}`, short hash = first 7 of the PR head SHA) and that CI builds the sandbox with the PR-specific base path, so a republished artifact serves correctly under `/pr/<n>/`.
- Deploy step is unchanged from the version deploy-preview.yml runs on every PR merge-completion, i.e. already exercised in production.

## Notes

- Sibling of #5518: both workflows now source everything from the run/PR payloads plus CI's static artifacts.

### DIFF
--- a/.github/workflows/redeploy-preview.yml
+++ b/.github/workflows/redeploy-preview.yml
@@ -2,9 +2,18 @@
 
 # Re-deploy Preview
 #
-# Manually re-deploys a PR's preview build (Storybook + Sandbox) to GitHub Pages.
-# Uses a retry-based git push instead of peaceiris/actions-gh-pages to avoid
-# concurrency conflicts with the main deploy workflow.
+# Manually re-deploys a PR's preview (Storybook + Sandbox) to GitHub Pages by
+# re-publishing the static artifacts its CI run already built — the same
+# artifacts deploy-preview.yml publishes automatically. Uses a retry-based git
+# push instead of peaceiris/actions-gh-pages to avoid concurrency conflicts
+# with the main deploy workflow.
+#
+# Safety: this NEVER checks out or runs PR-controlled code (mirroring
+# deploy-preview.yml and pr-comment.yml). It locates the PR's latest CI run for
+# its current head commit, downloads the pre-built storybook-<hash> /
+# sandbox-<hash> artifacts, and copies them into gh-pages. If the artifacts
+# have expired past retention, the fix is to re-run CI on the PR (which also
+# re-deploys the preview via deploy-preview.yml) — not to rebuild here.
 
 name: Re-deploy Preview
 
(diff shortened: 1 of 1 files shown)


.github/workflows/redeploy-preview.yml
```diff
@@ -34,40 +43,69 @@ jobs:
...
+      - name: Verify preview artifacts are present
+        run: |
+          # Unlike deploy-preview.yml (best-effort on every CI completion), a
+          # manual re-deploy that can't find its artifacts should fail loudly
+          # with the actionable fix.
+          if [ -d storybook-dist ] && [ -n "$(ls -A storybook-dist 2>/dev/null)" ] && \
+             [ -d sandbox-dist ] && [ -n "$(ls -A sandbox-dist 2>/dev/null)" ]; then
+            exit 0
+          fi
+          echo "::error::Preview artifacts for run ${{ steps.pr-info.outputs.run_id }} are missing or expired — re-run CI on PR #${PR_NUMBER}; deploy-preview will publish the fresh preview automatically."
```
cixzhang, line 101: Hmm, docsite-only PRs never upload these (ci.yml:251, :325), so re-running CI won't help.

cixzhang (approved):
Thanks — dropping the fork-head checkout from a `contents: write` job is the bigger win. Merging as is. Nit: docsite-only PRs never upload artifacts, so the error's advice can't work there.

[Full review](https://github.com/cixzhang/astryx/wiki/Review-5519)

[Reviewed by Robohands]

PR #5520, opened 2026-08-25
ci: validate Visual Baseline inputs before promotion

bhamodi:
## Summary

The Visual Baseline promote job took its operator-typed `run_id` straight to `download-artifact`, the shot keys straight into baseline file paths, and any completed run's capture at face value. The capture uploads with `if: always()`, and GitHub reports a terminal gate failure's run as `status=completed`, so a failed gate could replace baseline files. This adds the missing validation on all three.

## Changes

- **Verify the source run** (workflow): the run id must name a completed Release Gate run (`.github/workflows/release-gate.yml`) on `main`, otherwise fail with a clear error.
- **Validate the downloaded verdict** (workflow): after the download, `verdict.json` must have status `pass` or `changed`. A `failed`, `skipped`, missing, unreadable, or unknown verdict fails the job before anything is promoted.
- **`assertPromotableVerdict()`** in `lib/baseline.mjs`: the same rule as a small exported function. `accept()` (the only code that writes baseline files) calls it before touching the baseline, so the boundary also holds for `gate.mjs accept` run by hand or by `visual-baseline-manual`. `gate.mjs accept` reads `verdict.json` with a clear "unreadable" error and passes the verdict through.
- `accept()` validates every promoted and pruned key against the `shotKey()` alphabet (`[a-zA-Z0-9._-]`, not dots-only) before it is joined into a `shots/` path. `shotKey()` can only produce that alphabet, so real keys are unaffected.
- `--keys "a, b"` from the dispatch form is trimmed per key.
- Dispatch inputs route through `env:` in the summary step, matching the Promote step.
- `readBaseline()` deep-copies `EMPTY_MANIFEST`; the shallow spread shared its `shots` object between fresh baselines in one process.

Rebased on current `main`; the canonical-frame and shot-budget behavior in `gate.mjs` is unchanged.

## Test plan

- `lib/baseline.test.mjs`: `assertPromotableVerdict` lets `pass`/`changed` through and refuses `failed`, `skipped`, `null`, `undefined`, non-object, missing/null status, `crashed`, empty, and wrong-case statuses. Through `accept()`, each of those creates no baseline directory, and against an existing baseline neither overwrites a shot, prunes one, nor changes `manifest.json`. Path-shaped and dots-only keys throw for both the promote and prune lists.
- `gate-accept.test.mjs`: drives the real `gate.mjs accept` against a scratch capture. `pass` and `changed` promote; `failed`, `skipped`, missing, non-JSON, no-status, and unknown verdicts exit non-zero with "Refusing to promote" and write nothing. `--keys "a, b"` promotes both.
- Mutation check: removing only the `assertPromotableVerdict` call in `accept()` fails 20 tests; making the guard a no-op fails 35; reverting it to a failed/skipped-only check fails 27 (every missing/unreadable/unknown case); removing the key check or the trim each fails 1.
- Whole `.github/scripts/visual-gate` suite: 19 files, 345 tests pass. `gh-pages-publisher.test.mjs` (which spawns `gate.mjs accept`): 32 pass.
- Workflow YAML parses (js-yaml). The verdict step's shell was exercised locally with bash+jq for pass, changed, failed, skipped, unknown, null/object/numeric status, non-JSON, empty file, and no file: only pass/changed proceed.
- `node --check` on all changed `.mjs` files.

## Notes

### DIFF
--- a/.github/scripts/visual-gate/lib/baseline.mjs
+++ b/.github/scripts/visual-gate/lib/baseline.mjs
@@ -29,8 +29,12 @@ export const EMPTY_MANIFEST = {version: 1, shots: {}, decisions: []};
  */
 export function readBaseline(baselineDir) {
   const manifestPath = path.join(baselineDir, 'manifest.json');
-  if (!fs.existsSync(manifestPath)) return {manifest: {...EMPTY_MANIFEST}, exists: false};
-  return {manifest: JSON.parse(fs.readFileSync(manifestPath, 'utf8')), exists: true};
+  if (!fs.existsSync(manifestPath))
+    return {manifest: {...EMPTY_MANIFEST}, exists: false};
+  return {
+    manifest: JSON.parse(fs.readFileSync(manifestPath, 'utf8')),
+    exists: true,
+  };
 }
 
 /**
@@ -97,7 +101,18 @@ export function accept({
   runId = null,
   prune = [],
 }) {
-  if (!reason?.trim()) throw new Error('accept requires a reason — it is the record of the decision');
+  if (!reason?.trim())
+    throw new Error(
+      'accept requires a reason — it is the record of the decision',
+    );
+  // Shot keys name files inside shots/. A real key is shotKey() output
+  // ([a-zA-Z0-9._-] only, see plan.mjs), never a path — reject anything else
+  // before it is joined into one, whichever manifest or flag it came from.
+  const SHOT_KEY = /^(?!\.+$)[a-zA-Z0-9._-]+$/;
+  const badKey = [...keys, ...prune].find(key => !SHOT_KEY.test(String(key)));
+  if (badKey != null) {
+    throw new Error(`Invalid shot key: ${JSON.stringify(badKey)}`);
+  }
   const {manifest} = readBaseline(baselineDir);
   const shotsDir = path.join(baselineDir, 'shots');
   fs.mkdirSync(shotsDir, {recursive: true});
--- a/.github/workflows/visual-baseline.yml
+++ b/.github/workflows/visual-baseline.yml
@@ -55,6 +55,26 @@ jobs:
       - name: Setup Node and pnpm
         uses: ./.github/actions/setup
 
+      - name: Verify the source run
+        env:
+          GH_TOKEN: ${{ github.token }}
+          RUN_ID: ${{ inputs.run_id }}
+        run: |
+          set -eu
+          # Promotion pulls shots out of a specific run's capture artifact. The
+          # run id is operator-typed — confirm it names a completed Release Gate
+          # run on main before anything from it is treated as baseline material
+          # (a typo'd id should fail here, not become the baseline).
+          RUN_JSON=$(gh api "repos/${GITHUB_REPOSITORY}/actions/runs/${RUN_ID}")
+          RUN_PATH=$(printf '%s' "$RUN_JSON" | jq -r '.path')
+          RUN_BRANCH=$(printf '%s' "$RUN_JSON" | jq -r '.head_branch')
+          RUN_STATUS=$(printf '%s' "$RUN_JSON" | jq -r '.status')
+          if [ "$RUN_PATH" != ".github/workflows/release-gate.yml" ] || \
+             [ "$RUN_BRANCH" != "main" ] || [ "$RUN_STATUS" != "completed" ]; then
+            echo "::error::Run ${RUN_ID} is not a completed Release Gate run on main (path=${RUN_PATH}, branch=${RUN_BRANCH}, status=${RUN_STATUS})."
+            exit 1
+          fi
+
       - name: Queue baseline publication
         env:
           GH_TOKEN: ${{ github.token }}
@@ -67,6 +87,22 @@ jobs:
           run-id: ${{ inputs.run_id }}
           github-token: ${{ github.token }}
           path: .visual-run
+
+      - name: Validate the downloaded verdict
+        env:
+          RUN_ID: ${{ inputs.run_id }}
+        run: |
+          set -eu
+          # The capture uploads with if: always(), and a terminal gate failure
+          # still reports its run as status=completed — so the run check above
+          # cannot tell a green gate from a failed one. The verdict can: only
+          # a capture the gate stood behind (pass or changed) may become the
+          # baseline. gate.mjs accept enforces the same rule as backstop.
+          STATUS="$(jq -r '.status // "missing"' .visual-run/verdict.json 2>/dev/null || echo missing)"
+          if [ "$STATUS" != "pass" ] && [ "$STATUS" != "changed" ]; then
+            echo "::error::Run ${RUN_ID}'s verdict status is '${STATUS}' — only pass/changed captures may be promoted. Fix or rerun the gate first."
+            exit 1
+          fi
 
       - name: Wait for the baseline publication turn
         env:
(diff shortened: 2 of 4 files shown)


.github/scripts/visual-gate/lib/baseline.mjs
```diff
@@ -98,6 +98,14 @@ export function accept({
   prune = [],
 }) {
   if (!reason?.trim()) throw new Error('accept requires a reason — it is the record of the decision');
+  // Shot keys name files inside shots/. A real key is shotKey() output
+  // ([a-zA-Z0-9._-] only, see plan.mjs), never a path — reject anything else
+  // before it is joined into one, whichever manifest or flag it came from.
+  const SHOT_KEY = /^(?!\.+$)[a-zA-Z0-9._-]+$/;
+  const badKey = [...keys, ...prune].find((key) => !SHOT_KEY.test(String(key)));
```
cixzhang: `a, b` from the dispatch form throws on the space. Might need a trim first.

.github/scripts/visual-gate/lib/baseline.mjs
```diff
@@ -98,6 +98,14 @@ export function accept({
   prune = [],
 }) {
   if (!reason?.trim()) throw new Error('accept requires a reason — it is the record of the decision');
+  // Shot keys name files inside shots/. A real key is shotKey() output
+  // ([a-zA-Z0-9._-] only, see plan.mjs), never a path — reject anything else
+  // before it is joined into one, whichever manifest or flag it came from.
+  const SHOT_KEY = /^(?!\.+$)[a-zA-Z0-9._-]+$/;
```
cixzhang, line 111: baseline.test.mjs already has an accept block — worth a case for this.

cixzhang (commented):
Thanks, this is good — the run-id check is the valuable one and it fails closed. Two nits inline.

Holding the approval only because the workflow half has no signal anywhere: `visual-baseline.yml` is dispatch-only, nothing in the repo lints workflow YAML, and `pr-visual` skipped. Worth one human read of the YAML before it sits in the release path.

[Full review](https://github.com/cixzhang/astryx/wiki/Review-5520)

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks — a failed visual gate still passes this source check.

GitHub reports failed terminal runs as `status=completed`. The capture is uploaded with `if: always()`, and the current `accept` command promoted a test shot from `verdict.status=failed`, so this can replace baseline files from a gate the release contract says to hold.

Could you validate the downloaded visual result before promotion and add a failed-capture regression case? If you'd rather talk it through with someone, we're in [Discord](https://discord.com/invite/XnsUcFykEP).

[Reviewed by Robohands]

.github/workflows/visual-baseline.yml
```diff
@@ -59,6 +59,26 @@ jobs:
...
+        run: |
+          set -eu
+          # Promotion pulls shots out of a specific run's capture artifact. The
+          # run id is operator-typed — confirm it names a completed Release Gate
+          # run on main before anything from it is treated as baseline material
+          # (a typo'd id should fail here, not become the baseline).
+          RUN_JSON=$(gh api "repos/${GITHUB_REPOSITORY}/actions/runs/${RUN_ID}")
+          RUN_PATH=$(printf '%s' "$RUN_JSON" | jq -r '.path')
+          RUN_BRANCH=$(printf '%s' "$RUN_JSON" | jq -r '.head_branch')
+          RUN_STATUS=$(printf '%s' "$RUN_JSON" | jq -r '.status')
```
cixzhang, line 71: Terminal failure is also `completed`; validate the downloaded visual verdict before promotion.

cixzhang (changes requested):
Thanks — I rechecked this unchanged head. The existing blocker remains: a failed visual gate still produces a completed run and uploads its capture with `if: always()`, while baseline promotion does not validate `verdict.status`. That can replace baseline files from a gate the release contract says to hold.

Could you validate the downloaded visual result before promotion and add a failed-capture regression case? If you'd rather talk it through with someone, we're in [Discord](https://discord.com/invite/XnsUcFykEP).

[Reviewed by Robohands]

.github/scripts/visual-gate/lib/baseline.mjs
bhamodi: Trimmed at the split in gate.mjs — `--keys "a, b"` now promotes both, and I drove exactly that form against a scratch capture to confirm.

.github/scripts/visual-gate/lib/baseline.mjs
bhamodi: Added to the existing accept block — path-shaped and dots-only keys throw for both the promote and prune lists, using the same fixture harness.

.github/workflows/visual-baseline.yml
bhamodi: Right — I'd conflated "completed" with "green". The verdict now decides, in two places: a workflow step after the download requires `verdict.json` status pass/changed (missing or unreadable fails too), and `gate.mjs accept` refuses a failed/skipped verdict as backstop, so the rule holds for manual/local use as well. Reproduced your case against a scratch capture with `status: "failed"` — promotion refuses; flipping to `changed` promotes.

cixzhang (changes requested):
Thanks — the failed-capture guard fixes the production path, and the trim/key-test nits are fixed. One part of our previous request is still missing: exact-head CI runs `baseline.test.mjs`, but no test exercises the new verdict guard, so removing it still leaves the suite green.

Could you add the failed-capture regression case before this lands? If you’d rather talk it through with someone, we’re in [Discord](https://discord.com/invite/XnsUcFykEP).

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks — this still needs another reroll.

The prior regression request is still missing: removing only the new verdict guard leaves all 259 visual-gate tests green. Directly, `gate.mjs accept` also writes the baseline when `verdict.json` is missing or has an unknown status, despite the stated pass/changed-only boundary. Please make the promotion boundary reject failed, skipped, missing, and unknown verdicts, and add a test proving none can write.

This head also conflicts with current `main` in `gate.mjs`. Please rebase and preserve the current canonical-frame and shot-budget behavior while applying the narrow guard; no new CI abstraction is needed.

[Reviewed by Robohands]

bhamodi:
Rebased onto current main; the canonical-frame and shot-budget code in `gate.mjs` is untouched. The accept block keeps main's prune-inclusive `all` list and only adds the key trim and the verdict pass-through.

The boundary is now pass/changed only, in both places. The workflow step after the download distinguishes missing, unreadable, and unknown statuses explicitly and fails on all of them. In code, the guard is `assertPromotableVerdict()` in `lib/baseline.mjs`, and `accept()` calls it as its first statement, so nothing can write a baseline file from a failed, skipped, missing, unreadable, or unknown verdict regardless of how `accept` is reached. `gate.mjs accept` also refuses an unparseable `verdict.json` with a clear message instead of a stack trace.

Tests: `baseline.test.mjs` covers the guard directly and through `accept()` (no directory created; existing baseline bytes unchanged, including the prune path), and a new `gate-accept.test.mjs` spawns the real `gate.mjs accept` for the same cases. Mutation-checked: removing only the guard call inside `accept()` fails 20 tests, making the guard a no-op fails 35, and reverting to the old failed/skipped-only check fails 27. The whole visual-gate suite is 345/345.

.github/workflows/visual-baseline.yml
bhamodi: Now enforced twice: the workflow step after the download requires status pass/changed and names missing, unreadable, and unknown verdicts explicitly, and `accept()` in `lib/baseline.mjs` refuses the same set before it writes anything, with tests for each case.

cixzhang (approved):
Approved at exact head `b7c4ee3469af1b1d6fd88662905f1adc46622f01`. The failed-capture guard and regression coverage now satisfy the earlier requests, while the narrowly validated browser-refresh recovery remains available. Independent exact-head review and all required checks are green.

PR #5521, opened 2026-08-25
ci: render report fields in the PR comment as literal text

bhamodi:
## Summary

The PR Analysis comment is assembled from report files (`analysis.json`, `a11y-report.json`, `verdict.json`) by string interpolation, so a stray newline, pipe, backtick, or angle bracket in any report string could reshape the comment's own structure — a multi-line a11y description breaks out of its list item, a pipe in a story name adds table cells, a non-URL `helpUrl` makes a broken link. This makes the boundary explicit: comment structure comes only from the generator's literals; report fields render as data.

## Changes

- New `lib/report-text.js` with three helpers: `inline()` (one line, markdown/HTML characters render literally), `num()` (finite number or fallback), `safeUrl()` (plain absolute http(s) URL or `null` → link dropped, text kept).
- Applied at every report-field interpolation in `visual-format.js`, `a11y-format.js`, and `generate-pr-comment.js` (component names/stats, bundle rows, a11y descriptions/rule ids/help links, visual verdict reason/change rows/counts).
- Axe rule ids are reduced to their own `[\w-]` alphabet so the surrounding code span stays a code span; Storybook deep-link paths are reduced to the story-id alphabet before landing in the href.
- `extLink()` percent-encodes quotes in the href attribute.

## Test plan

- 25 tests pass across `report-text.test.mjs` (new), `a11y-format.test.mjs` (new), and `visual-format.test.mjs` (extended): ordinary values render byte-identically; multi-line/markup/pipe-bearing values render one-line and literal; non-URL help/report links are dropped. The literal-rendering cases were confirmed failing before the change.
- End-to-end smoke of `generate-pr-comment.js` with representative fixtures: output for normal inputs is unchanged (story deep link, a11y rows, bundle table all byte-identical).
- `node --check` on all four touched scripts.

## Notes

- CI-scripts-only change: no changeset.

### DIFF
--- a/.github/scripts/generate-pr-comment.js
+++ b/.github/scripts/generate-pr-comment.js
@@ -11,6 +11,9 @@
 const fs = require('node:fs');
 const { buildA11ySection } = require('./lib/a11y-format');
 const { buildVisualSection } = require('./lib/visual-format');
+// Report-file strings render as literal inline text, report numbers as
+// numbers — the comment's structure comes only from this file's literals.
+const { inline, num } = require('./lib/report-text');
 
 const args = process.argv.slice(2);
 const getArg = (name) => {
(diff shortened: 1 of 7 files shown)


.github/scripts/generate-pr-comment.js
```diff
@@ -48,14 +51,17 @@ function getStorybookLink(storybookBaseUrl, storyTitle) {
   // storyTitle may be a single string or an array when multiple story files match
   const title = Array.isArray(storyTitle) ? storyTitle[0] : storyTitle;
   if (!title) return null;
-  // Storybook converts "Core/XDSButton" to "core-xdsbutton" as the story ID prefix
-  const storyPath = title.toLowerCase().replace(/\//g, '-');
+  // Storybook converts "Core/XDSButton" to "core-xdsbutton" as the story ID
+  // prefix. Story ids only ever contain [a-z0-9_-]; the title came from a
+  // report file, so drop anything outside that alphabet before it lands in
+  // the href.
+  const storyPath = title.toLowerCase().replace(/\//g, '-').replace(/[^a-z0-9_-]/g, '');
```
cixzhang, line 58: Storybook dashes these rather than dropping them; `[^a-z0-9]+ → -` also fixes titles with spaces.

cixzhang (approved):
Thanks — good catch, and the tests are the right shape. One nit inline, on a link that was already broken; happy to take it as is.

[Full review](https://github.com/cixzhang/astryx/wiki/Review-5521)

[Reviewed by Robohands]

PR #5522, opened 2026-08-25
fix(Markdown): apply the URL safety rule to every image path

bhamodi:
## Summary

The Markdown parser checks every link destination with `isSafeUrl` — inline, reference (full/collapsed/shortcut), and autolinks — and inline image srcs too. But reference-style images (`![alt][label]`, `![alt]`) and standalone block images emitted their src without it. This routes those three emit sites through the same check, and normalizes control characters in the render-side `sanitizeUrl` the way `isSafeUrl` already does, so both layers apply one rule and see a URL the way a browser will.

## Changes

- `matchReferenceImage`: both emit sites now require `isSafeUrl(src)`, mirroring `matchReferenceLink` (parser.ts:295).
- Standalone block image branch: unsafe srcs fall through to the paragraph path and render as literal text — the same behavior as an unsafe inline image or link.
- `sanitizeUrl` (Markdown.tsx) strips `[\x00-\x1f\x7f]` before testing (the normalization `isSafeUrl` performs, with the same in-code rationale comment) and returns the normalized value, so stripped characters don't ride along into an attribute or a `components.image`/`components.link` override.
- Four new parser tests alongside the existing `(XSS prevention)` cases, covering the reference forms, block images, and control-character normalization for definitions.

## Test plan

- The 4 new tests fail on the previous parser and pass with the change (verified both ways).
- Full Markdown suites: parser 126/126, incremental, streaming, and `Markdown.test.tsx` component tests all green.
- Full `packages/core` suite: 7893/7893 passed.
- Prettier + eslint clean on the changed files.

## Notes

- Follows the precedent of the existing `isSafeUrl` normalization and its tests — this just closes the asymmetry between the link paths and the image paths.

### DIFF
--- a/packages/core/src/Markdown/Markdown.tsx
+++ b/packages/core/src/Markdown/Markdown.tsx
@@ -561,14 +561,19 @@ const headingStyles = {
 const DANGEROUS_URL_PATTERN = /^(javascript|data|vbscript):/i;
 
 function sanitizeUrl(url: string): string | null {
-  const trimmed = url.trim();
-  if (trimmed.length === 0) {
+  // Strip control characters before testing, the same normalization the
+  // parser's isSafeUrl applies — the anchored pattern must see the URL the
+  // way a browser will. Return the normalized value so the stripped
+  // characters don't ride along into an attribute or component override.
+  // eslint-disable-next-line no-control-regex -- control chars are the bypass
+  const normalized = url.replace(/[\x00-\x1f\x7f]/g, '').trim();
+  if (normalized.length === 0) {
     return null;
   }
-  if (DANGEROUS_URL_PATTERN.test(trimmed)) {
+  if (DANGEROUS_URL_PATTERN.test(normalized)) {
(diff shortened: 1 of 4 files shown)


packages/core/src/Markdown/Markdown.tsx
```diff
@@ -561,14 +561,19 @@ const headingStyles = {
 const DANGEROUS_URL_PATTERN = /^(javascript|data|vbscript):/i;
```
cixzhang, line 561: Blocks every `data:`; the parser only blocks `data:text/html`. A data-URI image renders as `[alt]`.

cixzhang (approved):
Thanks — the `[l]: <java\tscript:…>` definition was actually reaching the DOM as an `` on main, good catch. One for the description: parser and renderer still differ on `data:`.

[Full review](https://github.com/cixzhang/astryx/wiki/Review-5522)

[Reviewed by Robohands]

PR #5523, opened 2026-08-25
fix(docsite): give the playground preview an opaque origin

bhamodi:
## Summary

The playground preview compiles and runs editor code inside an iframe that shared the site's origin: `sandbox="allow-scripts allow-same-origin"` on a same-origin `src` makes the sandbox attribute a no-op, so previewed code ran with the site's cookies, storage, and synchronous `window.parent.document` access. Production builds now drop `allow-same-origin`: the preview document has an opaque origin and previewed code holds none of that authority. This is the model auto-executing playgrounds generally use (CodeSandbox/CodePen/StackBlitz isolate the runner origin; the Vue SFC playground uses a sandboxed iframe without `allow-same-origin`).

An opaque origin can't be named in `targetOrigin` or compared against `event.origin`, so the channel's trust anchor moves from origin equality (#5263) to a **nonce-attested MessagePort handshake**, and recovery from a replaced preview document is built on top of that handshake rather than on trusting whatever the frame's window currently holds.

## Trust model

Three parties (full write-up in the header of `previewChannel.ts`):

- The **playground** (site origin) owns the iframe element and decides what URL it navigates to.
- The **preview document** is our own page at that URL. It is trusted code, but previewed code runs inside it with full authority over its window: everything the document knows, previewed code knows, and it can navigate the frame anywhere.
- A **replacement document** is whatever the frame shows after previewed code navigated it (another origin, a `data:` URL, or a reload of the preview page itself). Only the playground can tell it from the real thing, and only by whether it navigated the frame there.

The handshake:

1. The playground mounts a **fresh iframe element** navigated to `/playground/preview#nonce=<random>`. The nonce exists only in the playground's memory and in that navigation's URL (fragment: never sent to the server or a CDN cache key). A document already in the frame cannot read either.
2. The preview document reads the nonce from its own URL and posts a hello carrying it to its parent.
3. The playground answers a hello only from the window of the iframe element mounted for this generation, only while a nonce is outstanding, only when it matches, and **only once**, by transferring one end of a fresh `MessageChannel`. The preview adopts a port only from a connect sent by its own parent window (`frame-ancestors 'self'` pins who that parent can be). All traffic then flows over the port pair.
4. The preview announces `preview-ready` only after its compiler **and its own document** have loaded (capped at 5 s so a stalled third-party subresource cannot stall the preview). Nothing previewed runs before that, so the frame's first `load` event in a generation always belongs to the attested document.
5. Every later `load` means the attested document was replaced. The playground closes the port, spends the nonce and its reference to the frame window, and **remounts the iframe** under a new nonce. A new element is a new browsing context: the replacement document is destroyed along with any navigation it had in flight, so it cannot race the restore, and a same-path reload cannot degrade into a fragment navigation. Until the new element mounts and hands over its window, the playground holds no frame window at all, so a hello from the still-alive old frame is rejected before the nonce is compared.

The fresh document receives the current code on `preview-ready` and the active theme and mode on the `previewReady` flip, so it picks up exactly where the old one was.

What a replacement document can do: post any message to the parent. What the playground acts on: a hello with the outstanding nonce from the current generation's window, which the replacement never saw. Code, theme, and targeting commands travel only over the port.

## Changes

- `previewChannel.ts`: the state machine (`createPreviewConnector`: `issue` / `attachFrame` / `handleWindowMessage` / `handleFrameLoad` / `post` / `stop`), `previewSrc` / `readPreviewNonce` / `announcePreview`, and the trust-model write-up.
- `PlaygroundClient.tsx`: issues a frame generation on mount (client only, so no nonce ever lands in server-rendered HTML), attaches the window listener before it, hands the mounted iframe's window to the connector via a callback ref, and on replacement queues the current code and issues a new generation.
- `PreviewStage.tsx`: the iframe is keyed on the generation and loads the nonce-carrying `src`. Production: `sandbox="allow-scripts"`. `next dev`: `sandbox="allow-scripts allow-same-origin"` (what the preview always had), because the dev server serves `/_next` assets only to requests naming an allowlisted host in `Origin`/`Referer`, and an opaque-origin document sends neither, so an isolated preview is a blank frame under `next dev`. Two literal elements rather than a computed attribute so the iframe-sandbox lint rule can read them.
- `preview/page.tsx`: posts the hello from mount, adopts the port from the parent's connect, gates `preview-ready` on its own document load.
- `next.config.mjs`: comment update (the handshake, not an origin check, is the guard the headers sit under).
- `.github/workflows/ci.yml`: new `docsite-browser` job (builds the docsite, runs the Playwright contract against `next start`). Separate from `docsite-test` so that lane keeps its ~2 minute signal; recommend adding `docsite-browser` to the required checks.
- `package.json`: `test:docsite-browser`; `apps/docsite/playwright.config.ts`; `CONTRIBUTING.md` / `apps/docsite/README.md`: how to run it.

## Test plan

**Browser contract** (`apps/docsite/e2e/playground-preview-isolation.spec.ts`, Chromium against the production server):

1. *Reload*: load `/playground?theme=gothic`, switch the preview to dark, assert the production sandbox (`allow-scripts` only; `parent.document` and `localStorage` throw inside the frame), reload the preview document from inside (`location.reload()`), and assert the playground mounts a new frame under a new nonce that renders the **current code with the Gothic tokens and dark mode before any edit**, then that an edit renders in the new document with the same theme and mode.
2. *Hostile navigation*: previewed code navigates the frame to another origin carrying the nonce it read from its URL. That page replays the hello (once, then continuously), adopts any port it is offered, answers `preview-ready`, tries `preview-edit-code`, and beacons every message it sees out of band. Assert it ran and replayed, received **no** message with a port and nothing over a port (no connect, code, or theme), that the playground tears it down and restores a trusted preview under a new nonce that renders the current code, and that the editor is untouched.

Mutation evidence (each mutant built into production, spec run against it):

| guard removed | reload spec | hostile spec |
| --- | --- | --- |
| one-shot nonce (a spent nonce accepted again) | pass | **fail** (the hostile page got the port, answered ready, and its `preview-edit-code` replaced the preview with "owned") |
| `onLoad` wiring to the connector | **fail** (same frame URL after reload) | **fail** (hostile frame never torn down) |
| current code queued for the new document | **fail** (new document renders nothing) | **fail** |
| theme/mode re-sent on `previewReady` | **fail** (mode not dark after reload) | pass |
| production sandbox relaxed to `allow-same-origin` | **fail** (sandbox attribute / isolation probe) | pass |
| nonce comparison **and** the frame-window gate together | pass | **fail** in 2 of 3 runs (the hostile page wins the port in the window between the old generation being discarded and React committing the new iframe) |
| nonce comparison alone, or the frame-window gate alone | pass | pass (the two guard the same window, so either one closes it; each is pinned individually by the unit suite) |

**Unit suite** (`playground-preview-channel.test.ts`, 18 tests): hello rejected for wrong/missing nonce, wrong/no window, before `issue()`, between `issue()` and the new element attaching, and after the first acceptance; first load in a generation ignored, later load discards port and nonce and fires `onReplaced`; replayed old nonce rejected before and after remount; `stop()` forgets everything; `acceptPreviewConnect` gates on type/source/port; nonce plumbing round-trips through the fragment. Mutation-checked: removing the nonce comparison, the one-shot, either source check, the first-load rule, or the discard-on-replace each fails at least one test.

**Also**: `pnpm -F @astryxdesign/docsite test` (41 files / 527 passed, 6 skipped), `tsc --noEmit` in `apps/docsite` clean, `eslint` and `ASTRYX_STRICT_LINT=1 eslint` clean on every changed file, the CI routing/workflow test suites (`.github/scripts/*.test.mjs`, 50 tests) pass with the new job, `check:sync` / `check:use-client` / `check:portable-scripts` / `check:package-boundaries` / `check:changesets` clean.

## Notes

- Storage/cookies inside the preview: the docsite providers use only `matchMedia` (no storage), so nothing in the preview page itself needs the origin. Previewed user code that touches `localStorage` now throws a `SecurityError` in production, which is the intended boundary.
- `next dev` keeps the pre-isolation sandbox (see `PreviewStage.tsx`); the isolation and this contract exist for production builds, which is where untrusted code meets users. The Playwright config documents this and requires a build.
- Docsite is a private app: no changeset.

### DIFF
--- a/apps/docsite/src/app/playground/PlaygroundClient.tsx
+++ b/apps/docsite/src/app/playground/PlaygroundClient.tsx
@@ -37,7 +37,7 @@ import dynamic from 'next/dynamic';
 import * as stylex from '@stylexjs/stylex';
 import {AppShell} from '@astryxdesign/core/AppShell';
 import {compressCode, decompressCode} from '../../lib/compress';
-import {isTrustedPreviewMessage, trustedPreviewOrigin} from './previewChannel';
+import {createPreviewConnector} from './previewChannel';
 import {Button} from '@astryxdesign/core/Button';
 import {Link} from '@astryxdesign/core/Link';
 import {HStack, VStack} from '@astryxdesign/core/Layout';
@@ -248,6 +248,12 @@ export function PlaygroundClient() {
   const {mode: siteMode} = useThemeMode();
   const editorTheme = siteMode === 'dark' ? 'github-dark' : 'github-light';
   const [code, setCode] = useState(getInitialCode);
+  // Mirror of `code` for event handlers with stable identities (the frame
+  // load handler resends the current code after a preview reload).
+  const codeRef = useRef(code);
+  useEffect(() => {
+    codeRef.current = code;
+  }, [code]);
   const [mode, setMode] = useState<'light' | 'dark'>('light');
   // A ?theme=<value> query param (e.g. from the themes gallery's "Open in
   // Playground") seeds the Theme view and preview. useSearchParams reads it
@@ -296,6 +302,13 @@ export function PlaygroundClient() {
   const iframeRef = useRef<HTMLIFrameElement>(null);
   const readyRef = useRef(false);
   const pendingRef = useRef<string | null>(null);
+  // The channel's offer lifecycle (see previewChannel.ts) and the current
+  // message handler for it — refs so the connector's ports always dispatch
+  // into the latest closure.
+  const connectorRef = useRef<ReturnType<typeof createPreviewConnector> | null>(
+    null,
+  );
+  const portHandlerRef = useRef<((e: MessageEvent) => void) | null>(null);
   const debounceRef = useRef<ReturnType<typeof setTimeout> | null>(null);
   const editorRef = useRef<MonacoTypes.editor.IStandaloneCodeEditor | null>(
     null,
@@ -358,12 +371,9 @@ export function PlaygroundClient() {
     }
   }, [themeParam]);
 
-  // Single channel to the preview iframe; no-ops until the iframe exists.
+  // Single channel to the preview iframe; no-ops until the handshake lands.
   const postToPreview = useCallback((message: unknown) => {
-    iframeRef.current?.contentWindow?.postMessage(
-      message,
-      trustedPreviewOrigin(),
-    );
+    connectorRef.current?.post(message);
   }, []);
 
   const send = useCallback(
@@ -410,17 +420,10 @@ export function PlaygroundClient() {
     [postToPreview],
   );
 
+  // Messages from the preview arrive only on the adopted port (wired up by
+  // the handshake loop below) — the window itself hears nothing.
   useEffect(() => {
     const handler = (e: MessageEvent) => {
-      if (
-        !isTrustedPreviewMessage(
-          e,
-          trustedPreviewOrigin(),
-          iframeRef.current?.contentWindow,
-        )
-      ) {
-        return;
-      }
       if (e.data?.type === 'preview-ready') {
         readyRef.current = true;
         setPreviewReady(true);
@@ -452,20 +455,40 @@ export function PlaygroundClient() {
         setIsTargeting(false);
       }
     };
-    window.addEventListener('message', handler);
-    return () => window.removeEventListener('message', handler);
+    portHandlerRef.current = handler;
+    return () => {
+      portHandlerRef.current = null;
+    };
   }, [postCode, postToPreview]);
 
+  // Handshake: the connector offers the preview a fresh port until it answers
+  // preview-ready over one (the offer doubles as the old readiness ping — the
+  // preview adopts whichever port arrives and replies once its compiler has
+  // loaded), then goes quiet until a frame reload re-arms it below.
   useEffect(() => {
-    const interval = setInterval(() => {
-      if (readyRef.current) {
-        clearInterval(interval);
-        return;
-      }
-      postToPreview({type: 'preview-ping'});
-    }, 300);
-    return () => clearInterval(interval);
-  }, [postToPreview]);
+    const connector = createPreviewConnector({
+      getFrame: () => iframeRef.current?.contentWindow,
+      onMessage: e => portHandlerRef.current?.(e),
+    });
+    connectorRef.current = connector;
+    connector.start();
+    return () => {
+      connector.stop();
+      connectorRef.current = null;
+    };
+  }, []);
+
+  // A frame `load` means the previous preview document — and its end of the
+  // channel — is gone (previewed code reloading its own frame, or a crash).
+  // Re-arm the offers and queue the current code so the fresh document picks
+  // up exactly where the old one was; the preview-ready handler flushes it
+  // and the theme effect re-sends on the previewReady flip.
+  const handleFrameLoad = useCallback(() => {
+    readyRef.current = false;
+    setPreviewReady(false);
+    pendingRef.current = codeRef.current;
+    connectorRef.current?.reset();
+  }, []);
 
   // Debounced push of code → preview + URL hash
   useEffect(() => {
@@ -1145,6 +1168,7 @@ export function PlaygroundClient() {
             isFullscreen={isFullscreen}
             onExitFullscreen={() => setIsFullscreen(false)}
             iframeRef={iframeRef}
+            onFrameLoad={handleFrameLoad}
             isInteractionDisabled={isResizing}
             isFullBleed={isMobile}
           />
--- a/apps/docsite/src/__tests__/playground-preview-channel.test.ts
+++ b/apps/docsite/src/__tests__/playground-preview-channel.test.ts
@@ -1,58 +1,184 @@
 // Copyright (c) Meta Platforms, Inc. and affiliates.
 
-import {describe, expect, it} from 'vitest';
-import {isTrustedPreviewMessage} from '../app/playground/previewChannel';
-
-const ORIGIN = 'https://astryx.atmeta.com';
+import {describe, expect, it, vi} from 'vitest';
+import {
+  PREVIEW_CONNECT,
+  acceptPreviewConnect,
+  createPreviewConnector,
+  connectToPreview,
+} from '../app/playground/previewChannel';
 
 const parent = {} as MessageEventSource;
 const other = {} as MessageEventSource;
 
-function event(origin: string, source: MessageEventSource | null) {
-  return {origin, source};
+function connectEvent(overrides: {
+  data?: unknown;
+  source?: MessageEventSource | null;
+  ports?: readonly MessagePort[];
+}) {
+  const {port1} = new MessageChannel();
+  return {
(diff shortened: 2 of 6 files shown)


apps/docsite/src/app/playground/PlaygroundClient.tsx
```diff
@@ -448,20 +443,37 @@ export function PlaygroundClient() {
...
   }, [postCode, postToPreview]);
 
+  // Handshake loop: offer the preview a fresh port until it answers
+  // preview-ready over one (the offer doubles as the old readiness ping — the
+  // preview adopts whichever port arrives and replies once its compiler has
+  // loaded). Each new offer supersedes the previous pair.
   useEffect(() => {
     const interval = setInterval(() => {
       if (readyRef.current) {
         clearInterval(interval);
```
cixzhang: Clears for good. Nothing re-offers if the frame reloads.

cixzhang (changes requested):
Thanks — this is the right call, and the handshake is careful work.

One thing before merge. The offer loop stops for good at `preview-ready`, and the preview can only adopt a port — it can't announce itself. So once the preview document is replaced (previewed code reloading its own frame, or a crash), nothing the playground sends arrives again, and the page has to be reloaded. Main recovers from the same event because the preview re-announced on mount; I measured both.

Re-arming on the frame's own load event covers it:

```tsx
<iframe onLoad={() => { readyRef.current = false; }} … />
```

The browser pass you flagged is otherwise clean: `parent.document` throws from inside the frame, a forged connect from another window is rejected, CSP `'self'` still matches under the opaque origin, and the webfonts still load.

Is there a case the re-arm misses?

[Full review](https://github.com/cixzhang/astryx/wiki/Review-5523)

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks — the security boundary is working: the production build gives the preview an opaque origin, blocks `parent.document`, and rejects a sibling-window handshake.

The existing reload block still stands at this unchanged head. I also need to correct my earlier suggested fix: setting `readyRef.current = false` in `onLoad` cannot restart the interval after `PlaygroundClient.tsx:459` already cleared it. I reloaded the production preview, edited the code, and it stayed blank while the page remained “Building…”.

Please make the channel recover from a replaced preview document and resend the current code/theme, with a focused browser test that reloads the frame, edits code, and sees the preview render. Does that cover the lifecycle you expect?

[Reviewed by Robohands]

apps/docsite/src/app/playground/PlaygroundClient.tsx
bhamodi: Fixed — and thanks for correcting your own suggestion before I chased it: you're right that flipping `readyRef` couldn't restart a cleared interval. The offer loop now lives in a `createPreviewConnector` in previewChannel.ts with an explicit `reset()`, and the iframe's `load` event re-arms it.

bhamodi:
Recovery is in. The offer lifecycle moved out of the component into `createPreviewConnector` (previewChannel.ts): offers until `preview-ready`, goes quiet, and `reset()` re-arms it. The frame's `load` event — the one signal that survives any way the document gets replaced, reload and crash-navigation alike — resets readiness, queues the CURRENT code into the pending slot, and re-offers; the ready handler flushes the code and the theme effect re-sends on the previewReady flip, so the fresh document picks up exactly where the old one was.

On "is there a case the re-arm misses": your corrected framing was the key one (the interval was already cleared). Two more I considered: a reload that lands while the compiler script is still loading (covered — the preview adopts from mount and answers when tsReady, and offers keep coming until it does), and a mid-recovery second reload (each load event resets again; each offer supersedes the previous pair).

The connector now has a focused lifecycle test driving exactly your repro over real MessagePorts: offer → adopt → quiet → edit delivered → document replaced → reset → new adopt → edit delivered to the NEW document, with offer counts pinned before/after each ready. That's jsdom-level; I'd still appreciate your browser pass on the production build for the visual half (reload the frame, edit, see it render), since I can't drive a browser from here.

cixzhang (changes requested):
Thanks — the reload behavior is fixed. I verified the replacement document receives current code and active theme state, and targeting still round-trips.

I previously asked for a focused browser regression. The new test calls `connector.reset()` directly, so it remains green if `<iframe onLoad={handleFrameLoad}>` disappears or if code/theme resend stops. That integration is the exact failure fixed here, and this security boundary needs it pinned in CI.

Could you add the browser test from the previous review: load the playground, reload the preview frame, edit, and assert the new document renders with the active theme?

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks — we asked for a browser reload/edit/theme test; this head still only calls `reset()` directly.

The opaque-origin isolation works, but recovery trusts the iframe’s window after its document changes. In Chromium, previewed code navigated to another origin; that page received the editor source/theme and could change the parent editor through the new port. Please restore or attest the trusted preview document before transferring a port, and cover that whole reload path in the browser test. Does that cover the remaining boundary?

[Reviewed by Robohands]

bhamodi:
Both points addressed at this head.

Trust after replacement: the playground no longer trusts the frame's window after a load. Each preview document is a generation: a fresh iframe element navigated to `/playground/preview#nonce=<random>`. The document echoes that nonce in a hello; the playground answers only a hello from the window the new element handed it, only with the outstanding nonce, and only once, by transferring the port. `preview-ready` is gated on the document's own load, so nothing previewed runs before the frame's first load event, and any later load spends the port, nonce, and window and remounts the iframe (a new browsing context, so the replacement document and any in-flight navigation die with the old one). Your Chromium repro is the hostile spec: the navigated-to page replays the nonce continuously, adopts any port, answers ready, and tries `preview-edit-code`; it receives nothing, and the playground restores a trusted preview that renders the current code. Full trust model in the `previewChannel.ts` header and the PR body.

Browser test: `apps/docsite/e2e/playground-preview-isolation.spec.ts`, run by the new `docsite-browser` CI job against a production build (`next dev` cannot serve `/_next` assets to an opaque-origin document, so the isolation only exists in production, which matches where you verified it). The reload spec reloads the frame from inside, asserts the new document renders the current code with the active theme and mode before any edit, then edits and asserts again. Mutation results are in the PR body: removing the one-shot let the hostile page rewrite the preview to "owned"; removing the load wiring, the code resend, or the theme resend fails the reload spec.

apps/docsite/src/app/playground/PlaygroundClient.tsx
bhamodi: The offer loop no longer exists. The preview's hello tells the playground exactly when a document is listening, and a replaced document triggers a new generation on the frame's `load` event, so there is nothing left to re-arm.

cixzhang (approved):
Approved at exact head `6add50924f6b6070fc847e6a92e5a129a40ed5fd`. This implements the narrowed production boundary: opaque-origin isolation from Docsite storage and parent DOM, no persistence guarantee, clear user explanation, and trusted recovery after reload or hostile navigation. It does not create clipboard, microphone, picker, or general permission policy. All seven Chromium cases and required checks pass.

PR #5524, opened 2026-08-25
fix(core): apply one URL scheme rule wherever a link leaves React's hands

bhamodi:
## Summary

React DOM vets the hrefs it writes as attributes, but Core has paths a URL takes without ever passing through React: `useClickableContainer` navigates imperatively (`window.open` for new-tab, Cmd/Ctrl-click and middle-click; `location.href` for plain click; a delegated click on its interactive ref), `useLinkComponent` forwards `href`/`to` to custom `LinkProvider` components, and Markdown turns untrusted source into links. This adds one shared navigation-destination module (`utils/safeUrl.ts`) and decides with it at every one of those exits, so the guarantee in `family:navigation-destinations` (spec AST-005) holds for every router and activation method, not only the native anchor.

## Changes

- **`utils/safeUrl.ts`** (internal, not exported): `sanitizeUrl` normalizes control characters/whitespace and blocks `javascript:`, `vbscript:`, `data:text/html`; `isSafeUrl` is its boolean form; `isSafeDestination` extends the rule to structured router destinations, deciding on a string `pathname`/`href` field and passing objects that carry neither.
- **`useClickableContainer`**: decides once, before target, modifier key, or the delegated anchor can select a sink. A blocked `href` does not navigate by any method and still runs the caller's `onClick`.
- **`useLinkComponent`**: `href` and an explicit `to` are checked independently; `to` is destructured out of `...rest` so it cannot ride around the check. A structured `to` such as `{pathname: '/docs'}` is preserved for the router (same object) instead of crashing; a blocked value is withheld as `undefined`.
- **Markdown**: the parser and the link render boundary call the shared module, so Markdown links follow the same matrix as Link and ClickableCard (a `data:image/*` link is navigation and passes). Image/resource policy stays separate and stricter (every `data:` image is still rejected).
- **Records and docs**: `family:navigation-destinations` moves both pending rows to shared adoption and lists the new focused tests in `verified_by`; `Markdown.spec.md` gains the render-boundary test; Link, LinkProvider, Button, ClickableCard, Item, ListItem, Token and Markdown docs state the shared outcome and point at the Link `href` description for the rule.
- Changeset: `@astryxdesign/core` patch.

## Test plan

- `safeUrl.test.ts`: string matrix (mixed case, embedded controls, relative/fragment/protocol-relative, http(s), mailto, tel, `data:image/*`), `sanitizeUrl` normalization, and structured-destination cases.
- `useLinkComponent.test.tsx`: ordinary and blocked `href`; explicit string `to` (blocked, and ordinary keeps precedence); structured `to` preserved as the same object, no-pathname object passes, blocked `pathname`/`href` withheld.
- `useClickableContainer.test.tsx`: positive and blocked cases for `_blank`, Cmd-click, Ctrl-click, middle-click, plain click (observed through a `location` stand-in, since jsdom does not navigate), and the delegated interactive ref; blocked `href` still fires `onClick`.
- `Markdown.test.tsx` / `parser.test.ts`: link matrix conformance with the shared rule, `data:image/*` link accepted while the same `data:image/*` image is rejected. `Markdown.renderBoundary.test.tsx` mocks the parser and hands the renderer a pre-parsed link node, pinning the render guard independently.
- Mutation checks, each done by removing the guard, running, then restoring: `useClickableContainer` click guard removed, 7 failures; middle-click guard removed, 1; guard kept only on the new-tab sink (previous head), 3 (same-tab x2, delegated x1). `useLinkComponent` `href` unchecked, 1; `to` unchecked, 3; string-only rule on `to`, 4 (`url.replace is not a function`); structured `to` dropped instead of preserved, 4. Markdown renderer back on the all-`data:` pattern, 3; renderer link guard removed, 5; parser inline-link guard removed, 4.
- Focused `vitest --project ui` over Link, hooks, utils, Markdown, ClickableCard, Token, Item, List, Button, Outline, Citation, Chat: 88 files, 1844 tests pass. `tsc` for `@astryxdesign/core` (`typecheck` and `typecheck:docs`) clean; eslint clean including `ASTRYX_STRICT_LINT=1`; `pnpm check:knowledge` and `pnpm check:changesets` pass.


packages/core/src/Link/useLinkComponent.ts
```diff
@@ -44,7 +43,17 @@ function createLinkWithTo(
...
+    // A native <a> gets its href vetted by React DOM; a custom link component
+    // receives raw props and may navigate imperatively, so the same scheme
+    // rule applies before the URL leaves the design system — to `href` and
+    // `to` alike.
+    const safeHref = href != null && isSafeUrl(href) ? href : undefined;
+    return createElement(Component, {
+      ref,
+      href: safeHref,
+      to: safeHref,
+      ...rest,
```
cixzhang, line 57: `to` isn't destructured above, so this puts the raw value back over `to: safeHref`.

cixzhang (changes requested):
Thanks for this — I confirmed the `location.href` sink executes in Chrome today, so it's worth having.

One thing first. `LinkWithTo` exists for `to`-routers, and `to` is the one prop that walks around the new check:

```tsx
createElement(Component, {ref, href: safeHref, to: safeHref, ...rest});
```

`to` isn't destructured, so `...rest` puts the raw value back. Someone wiring React Router and passing `to` themselves gets the unvetted URL, and nothing tells them.

Worth knowing rather than fixing here: `isSafeUrl` is a predicate every sink must remember, and three don't — the native `<a>` branch, the `interactiveRef.click()` proxy, and the Markdown sites in your [#5522](https://github.com/facebook/astryx/pull/5522). `richtext/linkUtils.ts` returns a safe URL instead of a boolean, which can't be forgotten.

Could you pull `to` out of `rest` before this lands?

[Full review](https://github.com/cixzhang/astryx/wiki/Review-5524)

[Reviewed by Robohands]

cixzhang:

- `packages/core/src/Link/useLinkComponent.ts:95` — This branch skips the rule, so `vbscript:` and `data:text/html` render verbatim here. React only blocks `javascript:`.

cixzhang (changes requested):
Thanks — I rechecked this at the current head, and the existing blocker still reproduces in Chromium. `LinkWithTo` sets `to: safeHref`, but because `to` is not destructured, the later `...rest` restores the caller's raw `to`. A custom router therefore still receives the unvetted URL even though `href` was checked. Could you pull `to` out of `rest` and add the explicit-`to` regression case? If you'd rather talk it through with someone, we're in [Discord](https://discord.com/invite/XnsUcFykEP).

[Reviewed by Robohands]

packages/core/src/Link/useLinkComponent.ts
bhamodi: Good catch — exactly right. `to` is destructured now, vetted by the same rule, and keeps its precedence over the href-derived value; `...rest` spreads before both so nothing rides around the check. Two regressions added: a script-scheme explicit `to` reaches the router as `undefined` (fails on the previous head), and an ordinary explicit `to` still wins over `href`.

cixzhang (approved):
Thanks — explicit `to` now passes the same URL check, and the regression fails on the old implementation. Nit: use `\x00`; the raw NUL makes Git treat this test as binary.

[Reviewed by Robohands]

cixzhang (changes requested):
AST-005 is now current, so I rechecked this exact head against the complete navigation-family contract. The custom `href`/`to` bypass is fixed, but three contract gaps remain before merge.

1. A structured router `to` now throws before the router sees it. `LinkComponentType` is an unconstrained `ElementType`, so a valid router value such as `{pathname: '/docs'}` reaches `isSafeUrl(to)` and crashes on `url.replace`. Please define the supported `to` boundary and preserve structured router destinations without letting them bypass destination validation.

2. Navigation decisions still disagree across the family. The new shared predicate permits `data:image/*` navigation, while Markdown rejects every `data:` link. Please make Markdown links conform to the shared navigation matrix while keeping Markdown image/resource policy separate. Then update `family:navigation-destinations` from pending to shared adoption, add the new focused tests to `verified_by`, and project the guarantee into the affected consumer docs as AST-005 requires.

3. The exact-head tests are not mutation-sensitive for every sink. I removed the same-tab `isSafeUrl` guard and all five `useClickableContainer` tests still passed; Cmd/Ctrl-click and the delegated `interactiveRef` path are also unpinned. Please add coverage that fails when each required guard is removed. Also replace the raw NUL byte in `safeUrl.test.ts` with `\x00` so Git and review tools treat the test as text.

Once those are covered, I can re-review the new head and merge it if green.

[Reviewed by Robohands]

bhamodi:
Thanks for the precise recheck. The new head addresses all three plus the nit:

1. **Structured `to`.** `isSafeDestination` in `utils/safeUrl.ts` now defines the boundary: strings go through the URL rule; an object is decided on its string `pathname`/`href` fields and otherwise passes (search/hash resolve against the current document and cannot select a scheme); an accepted object reaches the router as the same reference, a blocked one is withheld as `undefined`. Four regressions in `useLinkComponent.test.tsx`; the string-only rule from the previous head fails them with the `url.replace` crash you saw.

2. **Family agreement.** Markdown's parser and link render boundary now import the shared module; the all-`data:` pattern is kept only for images (`sanitizeResourceUrl`). `[x](data:image/png;...)` is a link, `x` still is not. `family:navigation-destinations` records shared adoption for both hooks, lists `safeUrl.test.ts`, `useClickableContainer.test.tsx`, and `Markdown.renderBoundary.test.tsx` in `verified_by`, and the Link, LinkProvider, Button, ClickableCard, Item, ListItem, Token and Markdown docs state the shared outcome with the rule described once on Link `href`. `pnpm check:knowledge` is green. I left `spec:AST-005` at `accepted`: the Chromium probe in its completion criteria is yours to run.

3. **Mutation sensitivity.** `useClickableContainer` now decides once before the sink split, which also covers the delegated `interactiveRef.click()` path. Each activation method has a blocked-case test; the same-tab case stubs `location` so the assignment is observable in jsdom. Removing the guard fails 7 tests; restoring the previous head's new-tab-only guard fails the two same-tab tests and the delegated test; removing the middle-click guard fails 1. On the Markdown side, a new render-boundary test feeds a pre-parsed link node past the parser, so removing the render guard alone now fails 5 tests (it failed 0 before).

The raw NUL in `safeUrl.test.ts` is replaced with `\x00`; git sees the file as text.

packages/core/src/Link/useLinkComponent.ts
bhamodi: Follow-up in the latest head: `to` is now typed `unknown` and decided by `isSafeDestination`, so a structured router destination is preserved rather than crashing, while a blocked `pathname`/`href` inside it is still withheld.

cixzhang (approved):
Approved at exact head `3f6b31ea03105203e37dcc531423ce8b52ddde1d`. This implements the approved uniform navigation-safety contract across native anchors, custom routers, imperative navigation, Markdown links, and structured URL forms without adding a bypass API or broadening image/resource policy. All 17 Chromium navigation cases, real Next.js SSR coverage, mutation guards, and required checks pass.

PR #5525, opened 2026-08-25
fix(cli): theme loading resolves from the project and obeys the safe-mode gate

bhamodi:
## Summary

`resolveTheme()` had two problems: its `createRequire` was bound to the CLI's install location, so package themes resolved from the wrong `node_modules`, and it loaded checkout files on every read command with no way to opt out. This keeps every documented setup working exactly as on main, including the theme guide's `{"astryx": {"theme": "./src/theme.ts"}}`, and puts all theme loading, file and package alike, behind the one safe-mode switch.

## Changes

- File specifiers load through the shared `importUserModule` (the documented `.ts` setup now uses the same jiti path configs do, instead of relying on Node's type stripping).
- Package specifiers resolve with a `createRequire` bound to the project (`<cwd>/package.json`), so `astryx.theme` names packages out of the project's own `node_modules`.
- Both are project code, so both obey `ASTRYX_NO_PROJECT_CODE=1`: no theme module loads (file, package, or `ASTRYX_THEME`), a one-line notice printed once per process explains why, and rendering degrades theme-less. The variable is read inline so this change stands alone; it is the same switch #5531's `projectCodeAllowed` reads.
- Every warning that prints the configured specifier routes it through `printable()`, which replaces control and format characters (`\p{Cc}`, `\p{Cf}`) with U+FFFD, so a checkout-controlled value cannot write escape sequences, newlines, or bidi overrides to the terminal.
- `resolveTheme` is now async; its one call site awaits it.

## Test plan

19 tests in `resolve-theme.test.mjs`, each verified to fail when the guard it protects is removed:

- The documented file setup loads relative to cwd, including `./src/theme.ts` with TypeScript syntax; a child-process test uses an `enum` (rejected by Node's built-in type stripping) so only the shared jiti loader can pass it.
- A package fixture that exists only in the project's `node_modules` resolves (fails against a CLI-bound require).
- `ASTRYX_THEME` still loads.
- Under the gate, file, package, and `ASTRYX_THEME` setups all return null with a probe global proving the module never executed; the notice prints once across repeated reads; any other value of the variable leaves loading on.
- Specifiers carrying OSC/CSI sequences, newlines, CR, and U+202E reach none of the five warn sites unescaped (package, file, bare-name, and gate-notice paths covered).

Also: `clients/cli` lib + component suites green (181 tests), `typecheck:strict`, eslint strict, `check:changesets`, `check:cli-structure`, and a manual run of `astryx component Button` against a fixture project with a `.ts` theme, a project-local package theme, and `ASTRYX_THEME`, each with and without `ASTRYX_NO_PROJECT_CODE=1`.

## Notes

- Pairs with #5531 (the gate for config/doc/integration loads); either merge order works. Whichever lands second can swap the inline env read for `projectCodeAllowed()`.
- `astryx doctor` reports the theme field as "wired" from its static presence; under the gate it is configured but not loaded. A follow-up in #5531 (which already touches doctor) can annotate that state.
- Changeset: `@astryxdesign/cli` patch.

### DIFF
--- a/packages/cli/clients/cli/lib/resolve-theme.mjs
+++ b/packages/cli/clients/cli/lib/resolve-theme.mjs
@@ -5,12 +5,21 @@
  *
  * Resolution sources (in priority order):
  * 1. ASTRYX_THEME environment variable
- * 2. xds.theme field in package.json
+ * 2. astryx.theme field in package.json
  *
  * Resolution strategy for the value:
- * - Starts with `.` or `/` → file path relative to cwd
- * - Starts with `@` → npm package (require/import)
+ * - Starts with `.` or `/` → file module relative to cwd (the documented
+ *   `{"astryx": {"theme": "./src/theme.ts"}}` setup — see the theme guide)
+ * - Starts with `@` → npm package, resolved from the project's node_modules
  * - Otherwise → try `@astryxdesign/theme-{name}`, then try as bare package name
+ *
+ * Loading a theme executes it — a file from the checkout directly, a package
+ * via whatever the checkout installed. Both are project code, so both sit
+ * behind the safe-mode gate: under ASTRYX_NO_PROJECT_CODE=1 no theme module
+ * loads at all and the CLI renders theme-less, the same "built-in data only"
+ * contract the rest of the CLI honors. (The variable is read inline here so
+ * this change stands alone; module-loader's projectCodeAllowed reads the same
+ * switch.)
  *
  * Returns the theme object's `variants` and `fonts` if available,
  * or null if no theme is configured or found.
@@ -19,30 +28,33 @@
 import * as fs from 'node:fs';
 import * as path from 'node:path';
 import {createRequire} from 'node:module';
-
-const _require = createRequire(import.meta.url);
+import {importUserModule} from '../../../foundation/fs/module-loader.mjs';
 
 /**
- * Try to load a module, returning the default export or the module itself.
+ * Try to load a module, returning its namespace. File specifiers go through
+ * the shared user-module loader (which also gives the documented `.ts` setup
+ * the same jiti path configs use); package specifiers resolve with a require
+ * bound to the PROJECT, not to the CLI's own install location, so
+ * `astryx.theme` names packages out of the project's node_modules.
  * Returns null if the module cannot be found.
  * @param {string} specifier
  * @param {string} cwd
- * @returns {unknown}
+ * @returns {Promise<unknown>}
  */
-function tryLoadModule(specifier, cwd) {
+async function tryLoadModule(specifier, cwd) {
   // For relative/absolute paths, resolve against cwd
   if (specifier.startsWith('.') || specifier.startsWith('/')) {
-    const resolved = path.resolve(cwd, specifier);
     try {
-      return _require(resolved);
+      return await importUserModule(path.resolve(cwd, specifier));
     } catch {
       return null;
     }
   }
 
-  // For package specifiers, try require
+  // For package specifiers, require from the project
   try {
-    return _require(specifier);
+    const projectRequire = createRequire(path.join(cwd, 'package.json'));
+    return projectRequire(specifier);
   } catch {
     return null;
   }
@@ -73,7 +85,11 @@ function extractTheme(mod) {
 
   // Check for any export ending in 'Theme'
   for (const key of Object.keys(mod)) {
-    if (key.endsWith('Theme') && typeof mod[key] === 'object' && mod[key]?.name) {
+    if (
+      key.endsWith('Theme') &&
+      typeof mod[key] === 'object' &&
+      mod[key]?.name
+    ) {
       return mod[key];
     }
   }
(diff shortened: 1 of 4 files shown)


packages/cli/clients/cli/lib/resolve-theme.mjs
```diff
@@ -22,6 +29,13 @@ import {createRequire} from 'node:module';
 
 const _require = createRequire(import.meta.url);
```
cixzhang: `createRequire` is bound here, not to `cwd`. Might be worth routing through `module-loader.mjs` instead.

cixzhang (commented):
Thanks — the diagnosis is right, and the field really is read on every `astryx component`.

A package name still reaches code the checkout controls: a local package named in `astryx.theme` can execute through `astryx component`. Restricting the config value alone therefore does not establish the trust boundary.

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks — the public [theme guide](https://github.com/facebook/astryx/wiki/Theming-Infrastructure#cli-theme-config) explicitly tells builders to use `"astryx": {"theme":"./src/theme.ts"}`. That setup prints the custom variant in published 0.5.0; this head drops it and warns, while `astryx doctor` can still report the field as wired.

The package-name check also does not complete the stated trust boundary: a package selected by the checkout executes before and after. Could you preserve the documented path behavior and add a regression proving every project-code theme load requires the same explicit opt-in?

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks — I rechecked this unchanged head against the two prior Robohands reviews. The second corrected the first and asked to preserve the documented path setup while gating every project-code theme load. Both blockers still reproduce: the [public theme guide](https://github.com/facebook/astryx/wiki/Theming-Infrastructure#cli-theme-config)'s `{"astryx":{"theme":"./src/theme.ts"}}` works on current main and published 0.5.0, but this head drops its variants while `astryx doctor` still reports it wired; a package selected by the same checkout still executes.

Could you preserve the documented path behavior and make both path and package theme loading obey one shared safe-mode gate? If you'd rather talk it through with someone, we're in [Discord](https://discord.com/invite/XnsUcFykEP).

[Reviewed by Robohands]

packages/cli/clients/cli/lib/resolve-theme.mjs
bhamodi: Right on both counts — bound to the CLI, and the wrong place to bind it. File specifiers now go through the shared module loader, and package specifiers use a require bound to `<cwd>/package.json`, with a regression that only passes when resolution is project-bound (the fixture package exists only in the fixture's node_modules).

bhamodi:
Reworked per your second review — you were right that the name check broke the documented setup without establishing the boundary. The restriction is gone: the theme guide's `{"astryx": {"theme": "./src/theme.ts"}}` works exactly as on published 0.5.0 (and now loads through the same jiti path configs use, rather than relying on Node's type stripping). In its place, every theme load — file, package, and ASTRYX_THEME alike — obeys the one safe-mode switch: under `ASTRYX_NO_PROJECT_CODE=1` no theme module loads at all, with a one-line notice.

Regressions added for each: the documented file setup loading, project-bound package resolution (fails against a CLI-bound require), and probe-based proof that nothing executes under the gate for all three sources. Pairs with #5531, which routes the config/doc/integration loaders through the same switch; either merge order works since this reads the variable inline.

packages/cli/clients/cli/lib/resolve-theme.mjs
```diff
@@ -73,21 +85,36 @@ function extractTheme(mod) {
...
 }
 
+/** Say once per process why a configured theme is not being loaded. */
+let gateNoted = false;
+/** @param {string} specifier */
+function noteThemeSkipped(specifier) {
+  if (gateNoted) return;
+  gateNoted = true;
+  console.warn(
+    `⚠ theme: ASTRYX_NO_PROJECT_CODE=1 — not loading theme "${specifier}"; rendering without theme data`,
```
cixzhang: Could we escape or omit this checkout-controlled value? Newlines and terminal controls pass through.

cixzhang (changes requested):
Thanks — I rechecked the two prior Robohands asks. This head now preserves the documented TypeScript theme path, resolves packages from the project, and keeps file, package, and `ASTRYX_THEME` modules unexecuted under the gate.

One security edge remains: `resolve-theme.mjs:107` writes the checkout-controlled theme value directly to the terminal. A value containing a newline or ANSI control sequence can spoof the output in the untrusted-checkout mode; I reproduced both reaching the warning unchanged. Could we escape or omit that value so the notice stays one inert line? If you'd rather talk it through with someone, we're in [Discord](https://discord.com/invite/XnsUcFykEP).

[Reviewed by Robohands]

packages/cli/clients/cli/lib/resolve-theme.mjs
bhamodi: Done — added a `printable()` helper that replaces control characters (`\p{Cc}`: C0, DEL, C1) with U+FFFD before a specifier reaches the terminal, and routed all five warn sites through it (the gate notice plus the four resolve-failure warnings, which print the same checkout-controlled value). Test added: a specifier carrying an OSC sequence and a newline warns with no control character surviving into the message.

bhamodi:
Rebased onto current main and re-verified the earlier asks against it. The escape is in place: every warn site (the gate notice plus the four resolve-failure warnings) prints the specifier through `printable()`, which replaces `\p{Cc}` and now also `\p{Cf}` (bidi overrides) with U+FFFD. Tests cover the package, file, bare-name, and gate-notice paths with OSC/CSI sequences, newline, CR, and U+202E, and each fails if `printable` becomes the identity. The documented `./src/theme.ts` setup is also now proven to go through the jiti loader by a child-process test using an `enum`, which Node's type stripping rejects; a plain `import()` or the old CLI-bound `require` fails it.

packages/cli/clients/cli/lib/resolve-theme.mjs
bhamodi: Done: routed through `printable()` (control and format characters to U+FFFD), and the notice now also prints once per process. Tests at the gate-notice, file, package, and bare-name sites, each failing if the escape is removed.

packages/cli/clients/cli/lib/resolve-theme.mjs
bhamodi: Still holds on current main: file specifiers use the shared `importUserModule`, package specifiers use a require bound to `<cwd>/package.json`. The regression only passes when resolution is project-bound (rebinding to the CLI's location fails it).

cixzhang (changes requested):
The project-relative resolver fix is useful, but the safe-mode design needs to change at exact head `cb5aaef66af3eab487b3a5971da1fe036e30490c`. A read-only `astryx component` command should not execute project theme code behind an opt-out environment variable; the command only needs static custom-variant metadata, and the current resolver does not reliably derive that from normal `defineTheme()` output anyway. Please split or rework this so trusted `astryx theme build` emits versioned, schema-validated variant JSON and `astryx component` reads it without executing project code, falling back to built-in docs when metadata is missing, invalid, or stale. Keep the resolver correctness repair separate if useful. PR 6491 may further settle the general CLI environment-variable policy.

PR #5526, opened 2026-08-25
ci: pin pnpm/action-setup by commit SHA

bhamodi:
## Summary

`.github/actions/setup` runs inside release.yml's publish and canary jobs. The repo's convention for third-party actions in privileged contexts is SHA-pinning (crowdin and peter-evans actions already are); `pnpm/action-setup` was the one exception, at a mutable tag. Pin it to the commit `v6` currently resolves to.

## Changes

- `pnpm/action-setup@v6` → `pnpm/action-setup@0977fd9…` (the commit the `v6` annotated tag points at today), with a `# v6` marker and a comment documenting the policy and how to bump.
- The pnpm version itself still comes from `package.json`'s `packageManager` field — behavior identical.

## Test plan

- YAML parses; SHA verified against the upstream `v6` tag (`git/ref/tags/v6` → tag object → commit `0977fd9`, dated 2026-08-03).
- No behavior change: same action code CI has been running.

## Notes


.github/actions/setup/action.yml
```diff
@@ -25,7 +25,12 @@ runs:
     # pnpm has to be on PATH before setup-node, otherwise `cache: pnpm` cannot
     # resolve the store path. The version comes from package.json's
     # `packageManager` field — never pin it here as well.
-    - uses: pnpm/action-setup@v6
+    #
+    # SHA-pinned like the repo's other third-party actions (crowdin,
+    # peter-evans): this composite also runs inside the npm publish jobs, so a
+    # mutable tag would let a retagged release into that context. Bump by
+    # updating the SHA to the new tag's commit.
+    - uses: pnpm/action-setup@0977fd99725f1db4007ccb2928dbb4e90d06cc86 # v6
```
cixzhang, line 33: Dependabot doesn't scan this directory. Might need a `directories` entry in dependabot.yml.

cixzhang (changes requested):
Thanks for this — pinning the publish path is the right instinct, and the SHA checks out.

Dependabot doesn't look inside `.github/actions/`: `actions/setup-node@v6` five lines below is still a major behind the rest of the repo. So this pin is permanent — nothing will ever move it, in the two jobs that publish to npm, and the bump note you added is the whole maintenance story.

Ideally I'd like the updater pointed at this file in the same PR:

```yaml
  - package-ecosystem: "github-actions"
    directories:
      - "/"
      - "/.github/actions/setup"
```

(`facebook/lexical/…@internal-registry-dependency-check-v1` is on a mutable tag too, so this isn't quite the last one.)

Does adding that directory work for you here? If you'd rather talk it through with someone, we're in [Discord](https://discord.com/invite/XnsUcFykEP).

[Full review](https://github.com/cixzhang/astryx/wiki/Review-5526)

[Reviewed by Robohands]

cixzhang:

- `.github/actions/setup/action.yml:38` — Hmm, this one's still v6 — pr-comment.yml went to v7 on 3 Aug.

cixzhang (approved):
Thanks — I asked earlier to point Dependabot here. That's useful maintenance, not a blocker. The pinned signed `v6` commit passed exact-head CI, so I'd merge as-is.

[Reviewed by Robohands]

.github/actions/setup/action.yml
bhamodi: Thanks for the approve — and the Dependabot gap was a good catch: opened the follow-up as its own PR so this one keeps its exact-head approval: the github-actions entry now lists `/.github/actions/setup` in `directories`, so the pin gets bump PRs instead of being permanent.

PR #5527, opened 2026-08-25
fix(apps): harden the embed theme-sync channel and theme lookups

bhamodi:
## Summary

Two related hardenings on the sandbox's embed theme-sync channel and the theme-name lookups around it, in the same pattern as #5263's sender checks on the docsite playground channel.

## Changes

- `apps/sandbox/src/app/providers.tsx`: the embed-mode `message` handler now requires `event.origin === window.location.origin` before acting — the shell that embeds the sandbox (`PreviewShell`) loads it from a relative src, so the legitimate sender is always same-origin. The handler's payload check also requires a string theme name.
- `PreviewShell.tsx`: the shell's own postMessage now addresses the iframe's origin explicitly instead of `'*'`.
- Theme-name lookups switched from the `in` operator to `Object.hasOwn` at all four sites (sandbox URL param, localStorage hydration, postMessage handler; docsite playground `?theme=` param). `in` walks the prototype chain, so a value like `constructor` passed validation and fed a non-theme object (`Object` itself) into `<Theme>`, corrupting the render. `Object.hasOwn` admits only real theme names.

## Test plan

- Prettier + eslint clean; `tsc -p apps/sandbox` has an identical error set to main (40 pre-existing in this environment, none in the changed files).
- Behavior unchanged for every real theme name and mode value; prototype-key values now fall back to the default theme exactly like any other unknown name.

## Notes

- Both apps are private (GitHub Pages deploys): no changeset.

### DIFF
--- a/apps/sandbox/src/app/providers.tsx
+++ b/apps/sandbox/src/app/providers.tsx
@@ -79,8 +79,12 @@ function getEmbedThemeParams(): {
   const paramMode = params.get('mode');
 
   return {
+    // Object.hasOwn (not `in`): the lookup must match a sandbox theme, not an
+    // inherited Object.prototype key like "constructor".
     initialTheme:
-      isEmbed && paramTheme && paramTheme in themes ? paramTheme : 'neutral',
+      isEmbed && paramTheme && Object.hasOwn(themes, paramTheme)
+        ? paramTheme
+        : 'neutral',
     initialMode:
       isEmbed && (paramMode === 'light' || paramMode === 'dark')
         ? (paramMode as ThemeMode)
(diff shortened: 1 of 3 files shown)


apps/sandbox/src/app/providers.tsx
```diff
@@ -149,9 +153,15 @@ export function Providers({children}: {children: React.ReactNode}) {
       return;
     }
     const handler = (event: MessageEvent) => {
+      // The shell that embeds the sandbox is same-origin by design
+      // (PreviewShell renders /?embed=1 from a relative src); no other
+      // window gets to drive theme state.
+      if (event.origin !== window.location.origin) {
```
cixzhang, line 159: Might also check `event.source === window.parent`; `previewChannel.ts` does both halves.

cixzhang (approved):
Thanks — good catch: before this, `?embed=1&theme=constructor` white-screened the sandbox. Worth also guarding at `providers.tsx:181`, where the name becomes a theme — want that here, or a follow-up?

[Full review](https://github.com/cixzhang/astryx/wiki/Review-5527)

[Reviewed by Robohands]

cixzhang:
The other two pointers, which landed outside the diff: `apps/sandbox/src/app/providers.tsx:181` — `themes[themeName] || neutralTheme` still trusts the name; `Object.hasOwn` there covers every entrance. And `apps/docsite/src/app/playground/preview/page.tsx:187` has the same `in` lookup — the fifth site, and the sweep missed it.

PR #5528, opened 2026-08-25
fix(CodeBlock): escape generated highlight names so any tokenizer type is safe

bhamodi:
## Summary

`ensureDynamicHighlightType` interpolates a token type into both the selector (`::highlight(astryx-<type>)`) and the declaration (`var(--color-syntax-<type>, currentColor)`) of a rule handed to `insertRule`. The built-in tokenizer's types are a fixed set, but `applyHighlightRangesChunked` / `applyHighlightRangesFlat` are public API and accept arbitrary token streams (for example from a server-side tokenizer). A dotted type such as `keyword.control.sql` threw from `insertRule` and took the whole code block down.

## Changes

- Both generated names pass through `CSS.escape` before entering rule text: the highlight name `astryx-<type>` and the custom property `--color-syntax-<type>`. Every type the CSS parser accepts keeps its colours (dotted, digit-led such as `9start`, `_private`, non-ASCII such as `キーワード`), and no token type can reach outside its own highlight rule: not the selector, not the declarations.
- The `CSS.highlights` registry key stays the raw string. A `::highlight()` selector matches by ident value, so the escaped rule still paints ranges registered under the raw name.
- `insertRule` is wrapped in `try/catch`. An engine that still refuses a rule costs that one type its colour (ranges paint with `currentColor`); it never throws through the public API.
- No allowlist and no skipped types: nothing is rejected up front.

## Test plan

- `highlightRanges.test.ts` spies on `CSSStyleSheet.prototype.insertRule` and pins the exact escaped rule text, then checks the parsed CSSOM:
  - accepted names (`keyword.control.sql`, `_private`, `キーワード`, `9start`) register under their raw names and produce one rule each with only the necessary characters escaped;
  - a type shaped to close `::highlight()` early (`sel), body { background: red } /*`, which the CSS parser would accept unescaped as a valid rule styling `body`) produces a rule whose `selectorText` is exactly the two highlight selectors, and no rule in the dynamic sheet has any other selector;
  - a type shaped to close `var()` early (`p); background: red; --x: var(--y`, which the parser would accept unescaped as three declarations) produces exactly one `color` declaration whose value is the escaped custom property with its fallback.
- Mutation-tested: removing the selector escape fails 3 tests; removing the custom-property escape fails 3 tests. The parsed-CSSOM assertions catch each mutation on their own as well.
- `pnpm vitest run --project ui packages/core/src/CodeBlock`: 42/42. `pnpm -F @astryxdesign/core typecheck`, eslint (including strict), prettier, and `pnpm check:changesets` clean.

## Notes

- No behavior change for the built-in tokenizer: its types are pre-seeded and never reach the dynamic path.

### DIFF
--- a/packages/core/src/CodeBlock/highlightRanges.test.ts
+++ b/packages/core/src/CodeBlock/highlightRanges.test.ts
@@ -4,14 +4,22 @@ import {describe, it, expect, vi, beforeEach, afterEach} from 'vitest';
 import {applyHighlightRangesChunked} from './highlightRanges';
 import type {TokenLine} from './tokenizer';
 
-// Mock CSS Highlight API
+// Mock CSS Highlight API. `escape` is real behavior, not a mock: the code
+// under test escapes generated names with it, so keep jsdom's implementation
+// (or a minimal stand-in) when replacing the CSS global.
 class MockHighlight extends Set<Range> {}
 const mockHighlightsMap = new Map<string, MockHighlight>();
+const realCssEscape =
+  typeof globalThis.CSS !== 'undefined' ? globalThis.CSS.escape : undefined;
 
 beforeEach(() => {
   mockHighlightsMap.clear();
 
   globalThis.CSS = {
+    escape:
+      realCssEscape ??
+      ((value: string) =>
+        value.replace(/[^a-zA-Z0-9_\u0080-\uFFFF-]/g, ch => `\\${ch}`)),
     highlights: {
       get: (name: string) => mockHighlightsMap.get(name),
       set: (name: string, h: MockHighlight) => mockHighlightsMap.set(name, h),
@@ -65,6 +73,57 @@ describe('applyHighlightRangesChunked', () => {
 
     cleanup();
     expect(kwHighlight!.size).toBe(0);
+  });
+
+  it('keeps every tokenizer type inside its own highlight rule', () => {
+    // The dotted type that crashed the block, names a narrow allowlist would
+    // wrongly reject (Chromium accepts them all), and a type shaped to step
+    // outside its ident.
+    const hostileType = 'k), body { background: url(https://example.com/x) } ';
+    const types = [
+      'keyword.control.sql',
+      '_private',
+      'キーワード',
+      '9start',
+      hostileType,
+    ];
+    const codeEl = createCodeElement(types.map(() => 'line'));
+    const tokenLines: TokenLine[] = types.map(type => [
+      {type, start: 0, end: 4},
+    ]);
+
+    const mockStyle = document.createElement('style');
+    mockStyle.setAttribute('data-astryx-highlight-styles', '');
+    document.head.appendChild(mockStyle);
+
+    const cleanup = applyHighlightRangesChunked(codeEl, tokenLines);
+
+    // Every type keeps its colours: ranges register under the RAW names.
+    for (const type of types) {
+      expect(mockHighlightsMap.get(`astryx-${type}`)?.size).toBe(1);
+    }
+
+    // And nothing in the stream produced a rule of its own making: every
+    // inserted rule is a highlight rule, and the payload's declaration
+    // appears nowhere.
+    const dynamicSheet = document.querySelector<HTMLStyleElement>(
+      'style[data-astryx-highlight-dynamic]',
+    );
+    const rules = Array.from(dynamicSheet?.sheet?.cssRules ?? []);
+    expect(rules.length).toBeGreaterThan(0);
+    for (const rule of rules) {
+      // Escaped payload characters stay inside the highlight IDENT (they may
+      // appear in the selector text), so the property to pin is the
+      // declarations: every rule paints color and nothing else.
+      expect(rule.cssText).toMatch(/^\.astryx-code-block code::highlight\(/);
+      const style = (rule as CSSStyleRule).style;
+      // (No color assertion: jsdom's cssstyle drops var() declarations, so
+      // only the absence of the payload's properties is checkable here.)
+      expect(style.getPropertyValue('background')).toBe('');
+      expect(style.getPropertyValue('background-image')).toBe('');
+    }
+
+    cleanup();
   });
 
   it('handles empty token lines', () => {
--- a/packages/core/src/CodeBlock/highlightRanges.ts
+++ b/packages/core/src/CodeBlock/highlightRanges.ts
@@ -64,11 +64,21 @@ function ensureDynamicHighlightType(tokenType: string): void {
     }
   }
 
-  const name = `astryx-${tokenType}`;
-  const colorVar = `var(--color-syntax-${tokenType}, currentColor)`;
-  dynamicStyleSheet.insertRule(
-    `.astryx-code-block code::highlight(${name}), .astryx-codeeditor code::highlight(${name}) { color: ${colorVar}; }`,
-  );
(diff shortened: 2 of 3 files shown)


packages/core/src/CodeBlock/highlightRanges.test.ts
```diff
@@ -67,6 +67,40 @@ describe('applyHighlightRangesChunked', () => {
...
+      [{type: hostileType, start: 0, end: 6}],
+    ];
+
+    const mockStyle = document.createElement('style');
+    mockStyle.setAttribute('data-astryx-highlight-styles', '');
+    document.head.appendChild(mockStyle);
+
+    const cleanup = applyHighlightRangesChunked(codeEl, tokenLines);
+
+    const dynamicSheet = document.querySelector(
```
cixzhang: `querySelector` returns `Element` here, so `.sheet` doesn't exist. Might need `<HTMLStyleElement>`.

packages/core/src/CodeBlock/highlightRanges.ts
```diff
@@ -39,6 +39,13 @@ interface CheckVisibilityOptions {
 const registeredHighlightTypes = new Set<string>(TOKEN_TYPES);
 let dynamicStyleSheet: CSSStyleSheet | null = null;
 
+/**
+ * A token type usable as a highlight name and CSS variable suffix. Built-in
+ * types and any reasonable custom-tokenizer type match; the rule text below
+ * is assembled by interpolation, so nothing CSS-structural may pass.
+ */
+const HIGHLIGHT_TOKEN_TYPE = /^[a-zA-Z][\w-]*$/;
```
cixzhang: Hmm, Chromium takes `_private` and non-ASCII idents. This rejects both.

cixzhang (changes requested):
Thanks for this — I reproduced the crash: a tokenizer emitting `keyword.control.sql` takes the whole code block down today, and the guard fixes it.

Typecheck and lint fail on the new test, so `build`, `build-storybook` and `lint` are red. `querySelector` returns `Element`:

```ts
document.querySelector<HTMLStyleElement>('style[data-astryx-highlight-dynamic]')
```

The bigger one: `^[a-zA-Z][\w-]*$` is narrower than the CSS parser. Chromium inserts `astryx-キーワード` and `astryx-_private` fine, so a tokenizer using either loses its colours after this — and the Notes say nothing changes. The injection it defends against doesn't reach either: `insertRule` parses one rule, so every structural type I tried threw, and every type it accepted produced only `color`.

I ran the alternative: dropping the regex and wrapping `insertRule` in try/catch fixes the crash and keeps those colours. Would that work here?

[Full review](https://github.com/cixzhang/astryx/wiki/Review-5528)

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks — I re-ran my earlier review at the same head. The crash and both existing blockers are unchanged: the new test still leaves build/lint red, and the ASCII regex rejects token types whose generated names Chromium accepts (`9start`, `_private`, `キーワード`), so existing custom tokenizers can silently lose syntax colors.

One correction to my earlier suggestion: try/catch alone is not safe. Chromium accepts a crafted type as a valid rule targeting `body`. Escaping both generated names with `CSS.escape` fixed the dotted-name crash, preserved the valid names above, and kept that payload inside the highlight selector when I drove it.

Could you escape the generated highlight and custom-property names, then fix the test typing? If you'd rather talk it through with someone, we're in [Discord](https://discord.com/invite/XnsUcFykEP).

[Reviewed by Robohands]

packages/core/src/CodeBlock/highlightRanges.test.ts
bhamodi: Fixed with the generic — `document.querySelector<HTMLStyleElement>(...)` — and the suite typechecks again.

packages/core/src/CodeBlock/highlightRanges.ts
bhamodi: You're right, and thanks for driving both alternatives — the allowlist is gone in favor of your corrected suggestion: both generated names go through `CSS.escape` before entering rule text, with a defensive try/catch so an engine that still refuses a rule costs that one type its colour rather than the block. The registry key stays the raw string (a `::highlight()` selector matches by ident value), so `keyword.control.sql`, `_private`, `9start`, and `キーワード` all keep their colours — the test now drives exactly that set plus the crafted type, and pins that every inserted rule is a highlight rule whose declarations contain nothing of the payload's making.

cixzhang (changes requested):
Thanks — the two earlier blockers are fixed at this head: the typed `querySelector` leaves build/Storybook green, and `CSS.escape` replaces the narrow allowlist. The focused regression test fails against `main` and passes 4/4 here.

Two small pieces remain: lint rejects the no-op `String(value)` in `highlightRanges.test.ts:22`, and the PR body still describes the removed regex/skip behavior. Could you remove that conversion and update the body to match the escaped-name behavior? If you'd rather talk it through with someone, we're in [Discord](https://discord.com/invite/XnsUcFykEP).

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks — the [runtime escaping](https://github.com/facebook/astryx/blob/61ad85bd9ca8c4e48ac61f693ebae8a288144682/packages/core/src/CodeBlock/highlightRanges.ts#L67-L81) and lint issue from my last review are fixed.

The [safety test](https://github.com/facebook/astryx/blob/61ad85bd9ca8c4e48ac61f693ebae8a288144682/packages/core/src/CodeBlock/highlightRanges.test.ts#L78-L124) still passes unchanged with either `CSS.escape` call removed. Its current payload only throws, so the catch hides the failure and the remaining rules satisfy the assertions. Without the selector escape, a parser-accepted payload can style `body`; without the property escape, one can add highlight styling. Both mutations need to fail.

The PR body also still describes the removed allowlist/skip behavior. Could you update the test and body?

[Reviewed by Robohands]

bhamodi:
Thanks for the precise mutation check. Both gaps are closed at this head.

The safety test is now three tests, each spying on `insertRule` and then checking the parsed CSSOM. The selector test drives `sel), body { background: red } /*`, which the parser accepts unescaped as a valid rule styling `body`; it asserts the exact escaped rule text and that the parsed `selectorText` is exactly the two highlight selectors (and that no rule in the sheet has any other selector). The property test drives `p); background: red; --x: var(--y`, which the parser accepts unescaped as three declarations; it asserts the exact escaped rule text and that the parsed rule has exactly one declaration, `color`, with the escaped custom property and fallback. Removing the selector escape fails 3 tests, removing the property escape fails 3 tests, and the CSSOM assertions alone catch each mutation too.

jsdom's parser accepts both unescaped payloads the same way Chromium does, so the assertions are against real parsed rules; the one place jsdom differs (it drops a `color` declaration whose value contains `\{`) is covered by the exact rule-string assertion instead.

The PR body, changeset, and commit message now describe the escaped-name behavior; the allowlist wording is gone.

cixzhang (approved):
Approved at exact head `44183b066ae1078e38c54c4fa39312601e377ff3`. This preserves the existing tokenizer API and behavior while safely escaping token labels in CSS and containing a failed dynamic rule locally. Focused Chromium evidence confirms valid and escaped labels paint correctly, one refused rule leaves neighboring highlighting intact, and the legacy span path is unchanged.

PR #5529, opened 2026-08-25
fix(theme): a theme declaration always stays one declaration

bhamodi:
## Summary

`generateThemeRules` assembled token and component-override declarations by raw string interpolation, so a value carrying CSS-structural characters could close its declaration or rule and continue as arbitrary CSS in the injected `<style>` (and in `astryx theme build` output). Theme definitions are code, but apps do assemble them from stored input (brand colors, white-labeling), so the generated stylesheet should never extend beyond the declarations it means to emit.

## Changes

- **One boundary check, following CSS syntax.** New `declarationBoundary.ts` is a small tokenizer-style scanner modeled on CSS Syntax Level 3: strings with backslash escapes and line continuations, comments, `url()` with an unquoted body, escaped code points outside strings, and `()`/`[]`/`{}` nesting. Valid CSS that a browser keeps inside one declaration passes through byte-identically: `url(data:…;base64,…)` in any case (`URL(`, `Url(`), escaped identifiers such as `Gill\ Sans`, closed comments (semicolons included), semicolons inside strings and url bodies, nested parentheses/brackets, `!important`, and vendor-prefixed property names (`-webkit-line-clamp`). Only a true break is refused: an unnested `;`, `{` or `}`, an unbalanced closer, an unclosed string/comment/block/`url(`, a trailing backslash (it would escape the generator's own `;`), and a `bad-string`/`bad-url` (the browser would discard those declarations anyway). Property names must be exactly one CSS identifier.
- **Every consumer-supplied path is covered:** root tokens and local tokens, component variants and their pseudo blocks, `onDark`/`onLight` tokens and components, adaptation rules, inherited (`extends`) themes, and the Heading weight override (an unsafe authored `fontWeight` falls back to the inherited/token weight instead of being interpolated). Prose rules reference tokens through `var()` on current main, so they never see a raw value.
- **Structured diagnostics.** A drop produces a `ThemeCSSDiagnostic` (`property`, `value`, `location` such as `components.button["variant:secondary"][":hover"]`, `reason`, `message`). Every generator (`generateThemeCSS`, `generateThemeRules`, `generateThemeRulesSplit`, `generateOnMediaCSS`, `generateAdaptationCSS`) accepts an optional `{onDiagnostic}` option; with none given, the `<Theme>` runtime keeps its `console.warn`. Additive, so an older core ignores the argument.
- **Build receipt.** `astryx theme build` passes a handler to all three generator calls and lists each drop in the receipt's `warnings` (`Declaration dropped "background-color" in components.button["variant:secondary"]: an unquoted ";" would end the declaration (value: …). The generated CSS omits it; …`). Programmatic builds no longer return `warnings: []` while the CSS omits a value the generated JS still carries. The programmatic API stays silent on the console.
- All seven shipped themes build with zero dropped declarations; their CSS is unchanged.

## Test plan

- `declarationBoundary.test.ts`: 69 cases. Keeps: colors, gradients, `calc()`, `light-dark()`, font stacks, `!important`, `url(data:…)` / `URL(…)` / `Url( … )`, quoted urls with `;` and braces, `image-set(url(a;b) …)`, `Gill\ Sans`, `a\;b`, hex escapes, `"a\" ; b"`, `;`/`{}` inside strings, closed comments with `;` inside, `[full-start] 1fr [full-end]`, `var(--x, a;b)`, backslash before newline, tab/newline whitespace, empty value. Rejects (with the reason asserted): top-level `;`/`{`/`}`, escaped-quote smuggling, unbalanced/mismatched closers, unclosed paren/bracket/string/comment/url, trailing backslash, bad-string, four bad-url shapes, control characters. Names: `-webkit-*`, `--brand-accent`, `--_internal`, escapes kept; `color: red`, `--x:red} body{…`, `a b`, `1px` rejected.
- `generateThemeRules.test.ts`: a table of 11 declaration paths (root tokens, local tokens, tokens and variants inherited via `extends`, component variant, pseudo block, `onDark` tokens, `onLight` variant, `onDark` pseudo, adaptation tokens, adaptation variant) × 13 preserved values (emitted byte-identically, no diagnostics) × 11 rejected payloads (payload absent from the CSS; exactly one diagnostic with the expected property, location and reason). Plus: vendor-prefixed names kept; invalid property name dropped; `:where(p)` intact with a `--text-body-size` payload; Heading weight falls back to `var(--font-weight-bold)` for authored and inherited (onDark) payloads with a single diagnostic; a dropped container padding leaves `border-radius` intact; `console.warn` fires only without a handler; all three entry points route diagnostics.
- `build.test.mjs`: a theme with payloads in tokens, a variant, a pseudo block, `onDark.tokens` and an adaptation rule yields 5 receipt warnings with exact property/location/reason, `console.warn` uncalled, and no payload in the written CSS; a theme with `Gill\ Sans, "Segoe;UI", serif /* ; */`, `URL(data:…)` and `WebkitLineClamp` builds with `warnings: []` and every declaration present.
- Mutation-checked: removing any single guard (each scanner rule, each path's check, each of the three build.mjs option pass-throughs, the warnings push, the console fallback, the name check) fails between 1 and 27 tests.
- `pnpm vitest run packages/core/src/theme` 1392/1392; `pnpm vitest run --project node packages/cli/api/theme` 258/258; build-theme CLI suites 76/76. `pnpm -F @astryxdesign/core typecheck`, `pnpm -F @astryxdesign/cli typecheck:strict`, eslint (normal and strict), prettier, `pnpm check:changesets` clean. `pnpm -F @astryxdesign/core build` and all `packages/themes/*` builds report no dropped declarations.

## Notes

- Changeset: `@astryxdesign/core` and `@astryxdesign/cli` patch.

### DIFF
--- a/packages/core/src/theme/generateThemeRules.test.ts
+++ b/packages/core/src/theme/generateThemeRules.test.ts
@@ -6,7 +6,7 @@
  * for both runtime and build paths.
  */
 
-import {describe, it, expect} from 'vitest';
+import {describe, it, expect, vi} from 'vitest';
 import {
   dataTokenDefaults,
   defineTheme,
@@ -798,5 +798,160 @@ describe('data visualization tokens', () => {
     expect(
       Object.keys(generateThemeCSS(defineTheme({name: 'data-shape'}))).sort(),
     ).toEqual(['component', 'prose']);
+  });
+});
+
+describe('declaration assembly keeps values as values', () => {
+  it('passes every ordinary token value through byte-identically', () => {
+    const theme = defineTheme({
+      name: 'brand',
+      tokens: {
+        '--color-accent': 'color-mix(in oklch, #FF00FF 80%, white)',
+        '--color-background-surface':
+          'linear-gradient(135deg, #FF00FF 0%, #00FFFF 100%)',
+        '--size-element-md': 'calc(100% - 24px)',
+        '--font-family-body': "'Inter Var', ui-sans-serif, system-ui",
+        '--color-text-primary': 'light-dark(#fff, #111)',
+      },
+    });
+    const {component} = generateThemeCSS(theme);
+    expect(component).toContain(
+      '--color-accent: color-mix(in oklch, #FF00FF 80%, white);',
+    );
+    expect(component).toContain(
+      '--color-background-surface: linear-gradient(135deg, #FF00FF 0%, #00FFFF 100%);',
+    );
+    expect(component).toContain('--size-element-md: calc(100% - 24px);');
+    expect(component).toContain(
+      "--font-family-body: 'Inter Var', ui-sans-serif, system-ui;",
+    );
+    expect(component).toContain(
+      '--color-text-primary: light-dark(#fff, #111);',
+    );
+  });
+
+  it('drops a token value that tries to close its rule', () => {
+    const warn = vi.spyOn(console, 'warn').mockImplementation(() => {});
+    const theme = defineTheme({
+      name: 'brand',
+      tokens: {
+        '--color-accent':
+          'red } input[value^="a"] { background: url(https://example.com/leak?a) ',
+      },
+    });
+    const {component} = generateThemeCSS(theme);
+    expect(component).not.toContain('input[value');
+    expect(component).not.toContain('leak');
+    expect(warn).toHaveBeenCalledWith(
+      expect.stringContaining('--color-accent'),
+    );
+    warn.mockRestore();
+  });
+
+  it('drops a component override value that tries to add a declaration', () => {
+    const warn = vi.spyOn(console, 'warn').mockImplementation(() => {});
+    const theme = defineTheme({
+      name: 'brand',
+      components: {
+        button: {
+          'variant:secondary': {
+            backgroundColor:
+              'red; background-image: url(https://example.com/x)',
+          },
+        },
+      },
+    });
+    const {component} = generateThemeCSS(theme);
+    expect(component).not.toContain('background-image');
+    expect(component).not.toContain('example.com');
+    warn.mockRestore();
+  });
+
+  it('keeps prose rules intact when a typed token carries a payload', () => {
+    const warn = vi.spyOn(console, 'warn').mockImplementation(() => {});
+    // The prose block interpolates type-scale tokens inline (via val), so an
+    // unsafe value falls back to the var() reference — which the token block
+    // dropped — instead of closing the :where(p) rule early.
+    const theme = defineTheme({
+      name: 'brand',
+      tokens: {
+        '--text-body-size': '1rem; } body { background: url(x) ',
+      },
+    });
+    const {component, prose} = generateThemeCSS(theme);
+    const css = component + prose;
+    expect(css).not.toContain('background: url(x)');
+    expect(css).not.toContain('body {');
+    expect(prose).toContain('font-size: var(--text-body-size);');
+    expect(warn).toHaveBeenCalledWith(
+      expect.stringContaining('--text-body-size'),
+    );
+    warn.mockRestore();
+  });
+
+  it('a data: URI value passes — its semicolons live inside url()', () => {
+    const theme = defineTheme({
+      name: 'brand',
+      components: {
+        button: {
+          'variant:secondary': {
+            backgroundImage:
+              'url(data:image/png;base64,iVBORw0KGgoAAAANSUhEUg==)',
+          },
+        },
+      },
+    });
+    const {component} = generateThemeCSS(theme);
+    expect(component).toContain(
+      'background-image: url(data:image/png;base64,iVBORw0KGgoAAAANSUhEUg==);',
+    );
+  });
+
+  it('an escaped quote cannot smuggle a live semicolon past the guard', () => {
+    const warn = vi.spyOn(console, 'warn').mockImplementation(() => {});
+    const theme = defineTheme({
+      name: 'brand',
+      tokens: {
+        // To CSS the leading \" is an escaped quote CHARACTER, not a string
+        // opener, so the semicolon after it is live and terminates the
+        // declaration. Pairing the quotes as a string would blank the span
+        // and let the injection through as "quoted".
+        '--color-accent': '\\"; background: url(https://example.com/leak); "',
+      },
+    });
+    const {component} = generateThemeCSS(theme);
+    expect(component).not.toContain('example.com');
+    expect(warn).toHaveBeenCalledWith(
+      expect.stringContaining('--color-accent'),
+    );
+    warn.mockRestore();
+  });
+
+  it('an escaped quote inside a string stays inside it', () => {
+    // `"a\" ; b"` is one string to CSS — its semicolon is inert. Closing the
+    // string at the escaped quote would expose that semicolon and reject a
+    // value that never leaves its declaration.
+    const theme = defineTheme({
+      name: 'brand',
+      tokens: {
+        '--color-accent': '"a\\" ; b"',
+      },
+    });
+    const {component} = generateThemeCSS(theme);
+    expect(component).toContain('--color-accent: "a\\" ; b";');
+  });
+
+  it('drops a property name that is not a property name', () => {
+    const warn = vi.spyOn(console, 'warn').mockImplementation(() => {});
+    const theme = defineTheme({
+      name: 'brand',
+      tokens: {
+        '--x:red} body{color:blue': 'red',
+      } as Record<string, string>,
+    });
+    const {component} = generateThemeCSS(theme);
+    expect(component).not.toContain('body{color:blue');
+    expect(component).not.toContain('body {');
+    warn.mockRestore();
   });
 });
--- a/packages/core/src/theme/generateThemeRules.ts
+++ b/packages/core/src/theme/generateThemeRules.ts
@@ -327,6 +327,72 @@ function expandContainerPadding(
 }
 
 // =============================================================================
+// Declaration assembly — values stay values
+// =============================================================================
+
+/** A custom property (`--color-x`, `--_internal`) or plain property name. */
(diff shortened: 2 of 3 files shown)


packages/core/src/theme/generateThemeRules.test.ts
```diff
@@ -739,3 +739,81 @@ describe('renamed theme targets', () => {
...
   });
 });
+
+describe('declaration assembly keeps values as values', () => {
+  it('passes every ordinary token value through byte-identically', () => {
+    const theme = defineTheme({
+      name: 'brand',
+      tokens: {
+        '--color-accent': 'color-mix(in oklch, #FF00FF 80%, white)',
+        '--color-hero': 'linear-gradient(135deg, #FF00FF 0%, #00FFFF 100%)',
```
cixzhang: `--color-hero` isn't a `TokenName`. This is what's failing the build.

cixzhang (changes requested):
Thanks for this — the hard part is picking which of the generator's many concatenation sites to guard, and six of them are right.

The prose rules still interpolate token values straight in, though, so the changeset's guarantee isn't true yet:

```ts
defineTheme({name: 'brand', tokens: {'--text-body-size': '1rem; } body { background: url(x) '}})
```

The token block drops that and warns; `:where(p)` emits it and the rule closes early. `--text-body-size` is an ordinary typed token, so a white-label config feeding the type scale gets there. Eleven `val()` calls have the same shape.

`--color-hero` isn't a `TokenName` either — that's the red build.

Not blocking: a `data:` URI value carries a `;`, so a themed `background-image` gets dropped too.

Can you route `val` through the check as well, or is the prose block different for a reason I'm missing? If you'd rather talk it through with someone, we're in [Discord](https://discord.com/invite/XnsUcFykEP).

[Full review](https://github.com/cixzhang/astryx/wiki/Review-5529)

[Reviewed by Robohands]

cixzhang:

- `packages/core/src/theme/generateThemeRules.ts:564` — Token value goes in raw here. `joinDeclarations` doesn't reach the prose block.
- `packages/core/src/theme/generateThemeRules.ts:384` — `val` hands back the theme's raw token string. Could route through the same check — the `var()` fallback is already here.

cixzhang (changes requested):
Thanks — I rechecked the current head against today’s main. Our earlier review asked for the prose-token path and the invalid `--color-hero` fixture to be fixed; this SHA is unchanged, so both still stand.

The same `--text-body-size` payload is dropped from the token block but emitted raw in `:where(p)`, so the changeset still promises a boundary the generator does not enforce. The semicolon filter also rejects valid CSS such as `url("data:image/svg+xml;base64,…")`; the declaration disappears with only a console warning. That silently narrows the existing public theme contract, so I’m tightening our earlier non-blocking note.

Could you make every consumer-supplied declaration path reject top-level declaration/rule breaks while preserving valid semicolons inside strings and URLs, fix the test token, and merge current main? The test file now conflicts with [#5562](https://github.com/facebook/astryx/pull/5562) and [#5566](https://github.com/facebook/astryx/pull/5566).

[Reviewed by Robohands]

packages/core/src/theme/generateThemeRules.test.ts
bhamodi: Fixed — the fixtures now use real TokenNames (`--color-background-surface`, `--size-element-md`, `--font-family-body`, `--color-text-primary`), and the build typechecks.

bhamodi:
No reason you were missing — the prose block was just a seventh site I didn't count. `val()` now applies the same rule: an unsafe inline value falls back to its `var()` reference (which the token block already dropped), so `:where(p)` stays intact; your exact `--text-body-size` payload is the new regression, asserting both the intact prose rule and the fallback.

Also took the non-blocking one, since the fix generalized: structural characters are now tested outside quoted strings and `url()` bodies, so a `data:` URI background (semicolons inside `url()`) passes through byte-identically — with an unterminated span failing closed against the raw text. Test added for that too. Rebased over the data-token work; the new tuple tokens flow through `joinDeclarations` as their computed `light-dark()` strings, and the full theme suite is green (829).

packages/core/src/theme/generateThemeRules.ts
```diff
@@ -326,6 +326,60 @@ function expandContainerPadding(
...
+ * calc(), light-dark(), font stacks — is free of them outside quoted strings
+ * and url() bodies, where they are inert (a data: URI legitimately carries
+ * `;`), so those spans are blanked before testing. An unterminated span never
+ * matches the blanking pattern, so it fails closed against the raw text.
+ */
+const UNSAFE_STRUCTURE = /[;{}]|\/\*/;
+// eslint-disable-next-line no-control-regex -- control characters are structural here
+const CONTROL_CHARS = /[\x00-\x1f\x7f]/;
+const OPAQUE_SPANS =
+  /url\(\s*("[^"\n]*"|'[^'\n]*'|[^)'"\s]*)\s*\)|"[^"\n]*"|'[^'\n]*'/g;
```
cixzhang: `OPAQUE_SPANS` stops at an escaped quote, so the following semicolon passes as quoted.

cixzhang (changes requested):
Thanks — our prior review asked for top-level declaration breaks to stay rejected while valid semicolons inside strings and URLs remain valid.

`OPAQUE_SPANS` still treats an escaped quote as the end of a quoted string:

```ts
backgroundColor: '"x\\"safe"; background: red; "y"'
```

That value passes the check and the generated CSS contains the second `background` declaration. The original boundary is still bypassable despite the new prose and data-URI coverage.

Could the scanner account for CSS escapes before this lands? If you'd rather talk it through with someone, we're in [Discord](https://discord.com/invite/XnsUcFykEP).

[Reviewed by Robohands]

packages/core/src/theme/generateThemeRules.ts
bhamodi: Good catch — fixed. The string spans in `OPAQUE_SPANS` now consume backslash escapes (`(?:[^"\\\n]|\\[^\n])*`), so a span closes exactly where CSS closes the string, and `isDeclarationValueSafe()` additionally refuses any value with a backslash left **outside** the blanked spans: to CSS that escapes the next character (`\"` is a quote character, not a string opener), so quote pairing after it can diverge from what CSS reads. Refusing is the fail-closed choice — no real token value escapes anything outside a string or url() body.

Two tests added: the `\"; … ; "` shape is dropped with a warning (confirmed failing against the previous regex), and `"a\" ; b"` — one string to CSS, its semicolon inert — now passes, which the old pairing wrongly rejected.

cixzhang (changes requested):
Thanks — the escaped-quote escape is fixed. Theme authors using valid existing CSS can still lose overrides in runtime and built themes. Chromium keeps uppercase `URL(data:…)`, `Gill\ Sans`, closed comments, nested semicolons, and vendor-prefixed properties within one declaration; [this parser drops them](https://github.com/facebook/astryx/blob/d4c94604a1f4b75cd41120a9f11b8690b99a0a10/packages/core/src/theme/generateThemeRules.ts#L333-L367) across base, variant, pseudo, on-media, and inherited themes.

Could the shared boundary follow CSS syntax, preserve valid names and values, and report every drop through [the structured build receipt](https://github.com/facebook/astryx/blob/d4c94604a1f4b75cd41120a9f11b8690b99a0a10/packages/cli/api/theme/build/build.mjs#L1433-L1449)? Programmatic builds currently return `warnings: []` while CSS omits values retained in generated JS.

[Reviewed by Robohands]

bhamodi:
Thanks, both asks are in and the branch is rebased on current main.

The boundary now follows CSS syntax rather than a regex: `declarationBoundary.ts` is a tokenizer-style scanner (strings with backslash escapes and line continuations, comments, `url()` with unquoted bodies including case-insensitive `URL(`, escaped code points outside strings, `()[]{}` nesting). Every case you listed is preserved byte-identically and covered by tests on every path: base tokens, local tokens, variants, pseudo blocks, onDark/onLight, adaptations, and inherited themes, plus the Heading weight override main added since. Only true breaks are refused: unnested `;`/`{`/`}`, unbalanced closers, anything left unclosed, a trailing backslash, and bad-string/bad-url (the browser discards those declarations anyway, so a diagnostic seemed more useful than a silent no-op). `!important` is kept since it stays one declaration. Prose rules on current main reference tokens via `var()`, so that path no longer sees a raw value; the test still asserts `:where(p)` stays intact.

Drops are now structured `ThemeCSSDiagnostic`s. Every generator accepts an optional `{onDiagnostic}` (additive; the runtime keeps `console.warn` when none is given), and `theme build` passes one to all three generator calls and lists each drop in the receipt's `warnings`. `build.test.mjs` asserts that a five-path payload theme produces five warnings with exact property/location/reason and a silent console, and that `Gill\ Sans, "Segoe;UI", serif /* ; */`, `URL(data:...)` and `WebkitLineClamp` build with `warnings: []`. All seven shipped themes build with zero drops.

packages/core/src/theme/generateThemeRules.test.ts
bhamodi: Still fixed after the rebase: the fixtures use real `TokenName`s (`--color-accent`, `--color-background-surface`, `--size-element-md`, `--font-family-body`, `--color-text-primary`, `--text-body-size`), and `typecheck` covers the test file.

packages/core/src/theme/generateThemeRules.ts
bhamodi: `OPAQUE_SPANS` is gone. The scanner consumes escapes the way the CSS tokenizer does, so `\"` outside a string is an escaped code point (the following `;` is live and rejected) and `"a\" ; b"` is one string (kept). Both remain as tests, plus a trailing-backslash case, since `red\` followed by the generator's own `;` would escape the terminator.

cixzhang (approved):
Approved for this exact head. The declaration-boundary fix and constrained warning collector match the accepted contract, and the required checks are green.

PR #5530, opened 2026-08-25
docs(vega): document that specs are programs, with the untrusted-spec wiring

bhamodi:
## Summary

`<VegaChart>` hands the `spec` prop to `vega.parse` + `new View` with Vega's defaults, so the defaults define the trust boundary: expression strings inside a spec compile through the Function constructor, and the default loader fetches any URL a spec names (including signal-built ones) with the page's credentials. Neither the README nor the `spec` JSDoc said any of this. Since the package is canary-only, now is the cheap time to make the contract explicit.

## Changes

- `spec` prop JSDoc: "a spec is a program, not just data" — author/review what you render; wire the interpreter + restricted loader for anything else, with a pointer to Vega's own interpreter guidance.
- `ParseOptions.ast` JSDoc: documents its role as the switch for interpreted evaluation (paired with `viewOptions.expr`), not just introspection.
- README "Untrusted specs" section with the full worked wiring through the component's existing pass-through options (`parseOptions={{ast: true}}`, `viewOptions={{expr: expressionInterpreter, loader}}`) and the CSP note (`'unsafe-eval'` omitted so the platform enforces the boundary).

## Test plan

- Docs-only: `tsc -p packages/vega` identical to main (0 errors); no test files exist in packages/vega yet; prettier clean.
- Verified the example flows through the component as written: `parseOptions` is forwarded to `vega.parse` and `viewOptions` (including `expr`/`loader`) into `new View`, so no code change is needed for the safe mode.

## Notes

- Follow-up worth considering before the package graduates from canary: a first-class `sandbox` prop that wires this by default (needs `vega-interpreter` as a dependency, so it's a separate change).
- `@astryxdesign/vega` is `private: true` (canary-only): changeset-exempt.

### DIFF
--- a/packages/vega/README.md
+++ b/packages/vega/README.md
@@ -156,6 +156,43 @@ To update data dynamically after render, use `onReady` to get the live View and 
 />
 ```
 
+### Untrusted specs
+
+A Vega/Vega-Lite spec is a program, not just data: Vega evaluates the
+expression strings inside it (signals, event streams, encodings, filters)
+and, by default, compiles them to JavaScript with the Function constructor.
+A spec's `data` entries can also name URLs — including signal-built dynamic
+ones — that the default loader will fetch with the page's credentials.
+
+`<VegaChart>` renders the spec you give it with Vega's defaults, so those
+defaults define the trust boundary: **only pass specs you author or
+review.** If specs come from users, stored documents, or model output, wire
+the safe evaluation mode through the existing pass-through options — Vega's
+own guidance for untrusted specs:
+
+```tsx
+import {expressionInterpreter} from 'vega-interpreter';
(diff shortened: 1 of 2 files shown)


cixzhang (approved):
Thanks — this is worth documenting and the wiring checks out: I ran the snippet and it renders under a CSP with no `'unsafe-eval'`. Two nits.

[Full review](https://github.com/cixzhang/astryx/wiki/Review-5530)

[Reviewed by Robohands]

cixzhang:
The two nits, since they land outside the diff: `packages/vega/README.md:134` still says `ast` is for tooling — probably wants to point at the new section. And `:187` rejects silently (no `onError`, just an empty chart); worth a clause.

PR #5531, opened 2026-08-25
fix(cli): ASTRYX_NO_PROJECT_CODE=1 keeps the CLI on built-in data

bhamodi:
## Summary

Discovery-backed commands (`component`, `search`, `docs`, `discover`, `template`, `doctor`, `layout`, ...) load the checkout's `astryx.config.*`, the integration manifests it names (and the installed ones autolink finds), doc modules, template specs, codemods, and theme sources. All of those execute on import. That is the right trade in a workspace the operator trusts and the wrong one in CI, triage, and agent runs over arbitrary checkouts. This adds the opt-out: `ASTRYX_NO_PROJECT_CODE=1` keeps the whole CLI on built-in data.

## Changes

- `importUserModule` (the chokepoint every user-authored module flows through) refuses under the gate before any of its loaders (jiti, fresh jiti, CommonJS `require`, native import) runs, with an error naming the variable. It carves out the CLI's own package root (`isCliShippedPath`) so built-in topics, templates, and command modules keep loading. `isProjectCodeGated` and `projectCodeGateMessage` are exported for the two loaders that need a different jiti.
- `Project.load` checks the gate before touching a config: a present `astryx.config.*` is acknowledged once on stderr and skipped, and no integration manifest is loaded (configured, autolinked, or the package being authored), so read commands degrade to built-ins instead of erroring.
- Every remaining checkout-path loader routes through the gate: `loadDocs` (component-loader), the docs catalog detail and list loaders, `search`'s doc loader, string-utils' keyword scan, and `doctor`'s config re-import (which now reports "config present but not loaded" instead of executing the file it just said it ignored). `template-adapter` keeps its JSX-capable jiti but asks the gate itself and skips external blocks entirely.
- `astryx theme build` refuses a theme source outside the CLI before any loader runs (jiti sync and async paths, the family preloader, the legacy eval fallback): exit 1, `ERR_THEME_LOAD`, one-line message, nothing written. Covers `--check`, `--family`, and `--watch`.
- README section "Running in checkouts you don't trust" describes exactly what is gated, and that the CLI's own runtime dependencies (`@astryxdesign/core`, `jscodeshift`) are not in scope.
- Default behavior (variable unset) is unchanged.

## Audit of checkout-backed import sites in `packages/cli`

Non-test, non-dist code. "Gated" means the site cannot execute a checkout module under `ASTRYX_NO_PROJECT_CODE=1`.

| Site | Status |
|---|---|
| `foundation/fs/module-loader.mjs` `importUserModule` (jiti, fresh jiti, CommonJS `require`, native import) | Gated at entry, before any loader path |
| `foundation/fs/module-loader.mjs` `loadModuleWithParser` | Gated via `importUserModule` |
| `foundation/config/project.mjs` config load, `autolinkIntegrations`, `loadLocalIntegration` | Gated before load: config nulled with a stderr notice; autolink and local package skipped |
| `foundation/integrations/integrations.mjs` manifest import | Gated via `importUserModule`; never reached under the gate |
| `foundation/integrations/autolink.mjs` | Calls `loadIntegrations`; skipped by `Project.load` under the gate |
| `foundation/discovery/component-loader.mjs` `loadComponentDoc` | Gated via `importUserModule` (already so on main) |
| `foundation/discovery/component-loader.mjs` `loadDocs` | Now gated (was a bare `import()`) |
| `foundation/discovery/docs-discovery.mjs` topic import | Gated via `importUserModule`; only reached for integrations, which never load under the gate |
| `foundation/discovery/template-adapter.mjs` `loadIntegrationDoc` | Gated via `loadModuleWithParser` |
| `foundation/discovery/template-adapter.mjs` `loadDocModule` (JSX jiti and `import()`) | Now gated: returns null for a non-shipped path; shipped core templates still load |
| `foundation/discovery/template-adapter.mjs` `discoverExternalBlocks` | Now gated: returns `[]` under the gate |
| `foundation/text/string-utils.mjs` keyword scan | Now gated via the loader (imported dynamically so the browser bundle stays free of node builtins) |
| `api/docs/_adapter.mjs` topic and translation import | Now gated via `importUserModule` |
| `api/docs/list/list.mjs` description import | Now gated (defense in depth; the catalog holds no project topic under the gate) |
| `api/search/search.mjs` `loadModuleDoc` | Now gated |
| `api/doctor/doctor.mjs` config re-import | Now gated: info row returned before the import; the re-import itself goes through `importUserModule` |
| `api/doctor/doctor.mjs` `_require.resolve` peer-dep checks | Not applicable: resolution only, nothing executes |
| `api/theme/build/build.mjs` jiti theme import, `evalModule`, `eval` fallback, family registry `import()`, `preloadFamilySource`; `core-interception.mjs` `createRequire(fromFile)(specifier)` | Gated upstream by `refuseGatedThemeSource` before any loader is created |
| `api/theme/build/build.mjs` `createJiti(...).resolve` calls | Not applicable: resolution only |
| `api/theme/build/build.mjs` `import('@astryxdesign/core/theme')`; `import('jscodeshift')` in codemods, upgrade, theme-discovery; babel plugin require in `authoring/shadcn` | Not applicable: CLI runtime dependencies resolved by package resolution, not checkout-authored code (stated as a limit in the README) |
| `assets/codemods/integration-discovery.mjs` codemod load | Gated via `loadModuleWithParser`; integrations never load under the gate |
| `api/upgrade/_adapter.mjs` post-codemod hook `execFile` | The command comes from a codemod module, which is gated above |
| `api/gap-report/gap-report-worker.mjs` handler import | Gated via `importUserModule`; the worker shares `process.env`, and handler entries derive from the config and loaded integrations, both empty under the gate |
| `api/gap-report/gap-report.mjs` `execFile('gh', ...)`, `api/swizzle/_github.mjs` | Not applicable: fixed external binary, not checkout code |
| `clients/cli/index.mjs` command module imports, `commands/template.mjs` lazy imports, `foundation/fs/paths.mjs` `pnpapi` | Not applicable: the CLI's own modules and the Yarn PnP runtime |
| `scripts/*.mjs`, `test/drift/docs-drift.mjs` | Not applicable: dev scripts, not CLI runtime |
| `clients/cli/lib/resolve-theme.mjs` `require` of `package.json` `astryx.theme` / `ASTRYX_THEME` | Not modified here; owned by #5525, which adds its own check of the variable. Follow-up there: use `isProjectCodeGated` from `module-loader.mjs` so the carve-out and message stay shared |

## Test plan

- `packages/cli/clients/cli/commands/no-project-code.test.mjs` (12 end-to-end probes): every fixture module sets a global when it executes; the real CLI runs in-process against it and the global must stay unset, so a loader that slips past the gate fails the suite. Covers `doctor` (text and `--json`), `component <Name>` from a repo-layout checkout, the fuzzy-match path (`component Buton`), `search` (with and without `--type`), `component` and `component --list` with a hostile integration that is both configured and declared as a dependency (so autolink is exercised), `docs` list plus a built-in detail, `template --list` with a hostile external block, `theme build` (text, `--json` code, `--family` with a preloaded intermediary), plus a control proving the same theme fixture executes and builds with the variable unset.
- Unit probes: `module-loader.test.mjs` (refusal for `.mjs`, `.js`, and `.ts` including `fresh`, shipped-path carve-out, exact-value check), `project.test.mjs` (config skip with stderr notice, no autolink, no local integration; spies prove the loaders are not even called), `integrationDocs.test.mjs` (`loadReferenceDocs` refuses a project topic and still loads a built-in with `--lang dense`).
- Each guard was mutation-tested: removing it fails the corresponding test(s). Removing the `importUserModule` gate fails 6, the config skip 1, the autolink guard 1, the local-integration guard 1, the doctor early return 2 (also with the old raw import restored), the `loadDocs`, search, string-utils, and docs-adapter rewrites 1 each, the external-blocks guard 1, the theme member gate 2, both theme gates 3. The family-level loop in `themeBuildFamily` is redundant with the member-level refusal (the plan always orders a selected member first) and is kept as belt-and-braces.
- CI-equivalent checks run locally: `pnpm -F @astryxdesign/cli typecheck:strict`, `typecheck:template-docs`, `node .github/scripts/cli-api-types-verify.mjs` (regenerates the api types), `ASTRYX_STRICT_LINT=1 eslint` and plain `eslint` on every changed file, `pnpm check:cli-structure`, `pnpm check:changesets`, `readme:check`, the copyright check, and `pnpm vitest run --project node packages/cli` (4031 passed, 6 skipped; the only failures were environment-specific `pnpm exec tsc` invocations and load timeouts that pass on rerun).

## Notes

- The gate covers code the checkout authors or contributes. A hostile package installed under `node_modules` that the CLI itself depends on (for example `@astryxdesign/core`) is out of scope, since refusing it would break every command; the README says so.
- Read commands that can only be answered by importing a project module fail with the loader's error (`ERR_UNKNOWN` in the JSON envelope). A dedicated error code is a small follow-up if wanted; `theme build` uses the existing `ERR_THEME_LOAD`.
- Changeset included (`@astryxdesign/cli` patch).

### DIFF
--- a/packages/cli/foundation/config/project.mjs
+++ b/packages/cli/foundation/config/project.mjs
@@ -32,7 +32,22 @@
 
 import * as fs from 'node:fs';
 import * as path from 'node:path';
-import {findPresentFiles, loadModuleWithParser} from '../fs/module-loader.mjs';
+import {
+  findPresentFiles,
+  loadModuleWithParser,
+  projectCodeAllowed,
+} from '../fs/module-loader.mjs';
+
+/** Say once per process why a present config is being ignored. */
+let configSkipNoted = false;
+/** @param {string} configPath */
+function noteConfigSkipped(configPath) {
+  if (configSkipNoted) return;
+  configSkipNoted = true;
+  console.error(
+    `astryx: ASTRYX_NO_PROJECT_CODE=1 — ignoring ${configPath}; running on built-in data only`,
+  );
+}
 import {parseConfig} from '../../authoring/config/parse.mjs';
 import {loadIntegrations} from '../integrations/integrations.mjs';
 import {
@@ -199,7 +214,14 @@ export class Project {
    */
   static async load(cwd = process.cwd(), {cache} = {}) {
     const resolvedCache = cache ?? new InMemoryConfigCache();
-    const configPath = findConfigPath(cwd);
+    let configPath = findConfigPath(cwd);
+    // Loading the config executes it (and every integration it names).
+    // Under the project-code gate the whole invocation runs on built-ins:
+    // present config acknowledged once on stderr, never imported.
+    if (configPath && !projectCodeAllowed()) {
+      noteConfigSkipped(configPath);
+      configPath = null;
+    }
     const hash = configContentHash(configPath);
 
     /** @type {import('../../authoring/config/type').AstryxConfig} */
--- a/.changeset/cli-no-project-code-gate.md
+++ b/.changeset/cli-no-project-code-gate.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/cli': patch
+---
+
+[fix] New ASTRYX_NO_PROJECT_CODE=1 gate for running the CLI in checkouts you don't trust (CI, triage, agent runs): a present astryx.config.* is acknowledged and skipped instead of executed, and every other workspace-module import (integration manifests, doc modules, codemods) refuses with an error naming the variable. Default behavior is unchanged.
+
+@bhamodi
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -522,6 +522,22 @@ The config is validated against a strict schema when the CLI loads it, so an
 unknown field is a hard error rather than a silent no-op. `astryx doctor`
 reports whether the config loads cleanly.
 
+### Running in checkouts you don't trust
+
+Loading a config executes it, and so does loading the integration manifests
+and doc modules it leads to — which is what makes the config useful, and also
+means discovery-backed commands run code from whatever checkout they're
+invoked in. For CI, triage, and agent runs over arbitrary checkouts, set:
+
+```bash
+ASTRYX_NO_PROJECT_CODE=1 astryx component Button
+```
+
+Under this gate the CLI runs on built-in data only: a present
+`astryx.config.*` is acknowledged on stderr and skipped, and any command that
+would have to import a workspace module fails with an error naming the
+variable instead of executing it.
+
 ## Core codemod authoring
 
 Core codemods live under `packages/cli/assets/codemods/transforms/`. Released
--- a/packages/cli/api/docs/_adapter.mjs
+++ b/packages/cli/api/docs/_adapter.mjs
@@ -18,7 +18,7 @@
 
 import * as fs from 'node:fs';
 import * as path from 'node:path';
-import {pathToFileURL} from 'node:url';
+import {importUserModule} from '../../foundation/fs/module-loader.mjs';
 import {Project} from '../../foundation/config/project.mjs';
 import {
   DocsCatalog,
@@ -55,8 +55,11 @@ export async function loadDocsCatalog(cwd = process.cwd()) {
  * @returns {Promise<import('./docs.type.mjs').DocsDetailResponse['data']>}
  */
 export async function loadReferenceDocs(docPath, {lang} = {}) {
-  const mod = await import(pathToFileURL(docPath).href);
-  const docs = mod.docs ?? mod.default;
+  // Gated loader: built-in topics live under the CLI package and always
+  // load; integration topic modules execute on import and must honor
+  // ASTRYX_NO_PROJECT_CODE.
+  const mod = await importUserModule(docPath);
+  const docs = /** @type {any} */ (mod.docs ?? mod.default);
   if (!lang || lang === 'en') return docs;
 
   const dir = path.dirname(docPath);
@@ -65,8 +68,10 @@ export async function loadReferenceDocs(docPath, {lang} = {}) {
   const translationPath = path.join(dir, `${base}.doc.${locale}.mjs`);
   if (!fs.existsSync(translationPath)) return docs;
 
-  const translationMod = await import(pathToFileURL(translationPath).href);
-  const translation = translationMod.docsZh || translationMod.docsDense;
+  const translationMod = await importUserModule(translationPath);
+  const translation = /** @type {any} */ (
+    translationMod.docsZh || translationMod.docsDense
+  );
   if (!translation) return docs;
 
   // Overlays are keyed to a base section by title (`section`), not by array
@@ -86,7 +91,9 @@ export async function loadReferenceDocs(docPath, {lang} = {}) {
     ...docs,
     description: translation.description || docs.description,
     sections: docs.sections.map(
-      (/** @type {import('@astryxdesign/cli/authoring').ReferenceSection} */ section) => {
+      (
+        /** @type {import('@astryxdesign/cli/authoring').ReferenceSection} */ section,
+      ) => {
         const ts = bySection.get(section.title);
         if (!ts) return section;
         return {
@@ -99,8 +106,10 @@ export async function loadReferenceDocs(docPath, {lang} = {}) {
             ) => {
               const tb = ts.content?.[bi];
               if (!tb) return block;
-              if (tb.type === 'prose' && block.type === 'prose') return {...block, text: tb.text};
-              if (tb.type === 'list' && block.type === 'list') return {...block, items: tb.items};
+              if (tb.type === 'prose' && block.type === 'prose')
+                return {...block, text: tb.text};
+              if (tb.type === 'list' && block.type === 'list')
+                return {...block, items: tb.items};
               return block;
             },
           ),
--- a/packages/cli/api/docs/list/list.mjs
+++ b/packages/cli/api/docs/list/list.mjs
@@ -14,7 +14,7 @@
  * @position Leaf under api/docs. Sibling of detail; both share _adapter.mjs.
  */
 
-import {pathToFileURL} from 'node:url';
+import {importUserModule} from '../../../foundation/fs/module-loader.mjs';
 import {loadDocsCatalog} from '../_adapter.mjs';
 
 /**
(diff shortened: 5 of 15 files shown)


packages/cli/foundation/config/project.mjs
```diff
@@ -32,7 +32,21 @@
...
-import {findPresentFiles, loadModuleWithParser} from '../fs/module-loader.mjs';
+import {
+  findPresentFiles,
+  loadModuleWithParser,
+  projectCodeAllowed,
+} from '../fs/module-loader.mjs';
+
+/** Say once per process why a present config is being ignored. */
+let configSkipNoted = false;
+function noteConfigSkipped(configPath) {
```
cixzhang, line 44: Needs a `@param {string} configPath`.

cixzhang (changes requested):
Thanks for this — the trade you're naming is real and `importUserModule` is the right place for the switch.

It isn't holding yet. `astryx doctor` re-imports the config directly at `doctor.mjs:328`, so under the gate it prints "ignoring …; running on built-in data only" and then executes that same file. `astryx component <Name>` runs a doc module from the checkout through `loadDocs` at `component-loader.mjs:141` — its neighbour `loadComponentDoc` goes through the gate, `loadDocs` doesn't. About nine bare `await import()`s of checkout paths remain.

```bash
ASTRYX_NO_PROJECT_CODE=1 astryx doctor   # config runs anyway
```

Ideally I'd like those loaders routed through `importUserModule` too, so the README sentence is true when it ships. Also CI is red on a missing `@param` for `noteConfigSkipped` (`project.mjs:43`).

Is routing them all through the one loader workable, or is something in `loadDocs` making that hard?

[Full review](https://github.com/cixzhang/astryx/wiki/Review-5531)

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks for adding this safety switch. We previously asked for every checkout-backed loader to honor it and for the new helper to typecheck; this exact head has not changed, so both are still open.

With `ASTRYX_NO_PROJECT_CODE=1`, `astryx doctor` still executes a fixture config after saying it was ignored, and `astryx component` still executes a fixture doc module. An agent can therefore opt out and still run checkout code. CI also still fails because `noteConfigSkipped(configPath)` has an implicit `any`.

Could you cover every project-code import path, add end-to-end probes for these two commands, and add the missing `@param`?

[Reviewed by Robohands]

packages/cli/foundation/config/project.mjs
bhamodi: Added, and the helper typechecks now.

bhamodi:
Thanks for running it — you were right, the switch wasn't connected to most of the loaders. Reworked so the README sentence is true:

- `importUserModule` now carves out the CLI's own package root (`isCliShippedPath`): shipped modules (built-in doc topics, templates, command modules) always load, everything else refuses under the gate. That let every checkout-path loader route through it — `loadDocs` (component-loader), the docs catalog detail/list/translation loaders, `search`'s doc loader, `theme build`'s target-doc loader, and string-utils' keyword scan. The nine bare `await import()`s of checkout paths are gone.
- `astryx doctor` no longer re-imports the config under the gate — its config check returns an info row ("config present but not loaded") instead of executing the file it just said it ignored.
- `template-adapter` keeps its JSX-capable jiti (importUserModule's isn't), but now checks the gate + carve-out itself and treats a gated spec like a missing file.
- Added the end-to-end probes you asked for (`no-project-code.test.mjs`): fixture config and doc modules plant a global on execution; `astryx doctor` and `astryx component Button` run for real via the in-process harness and the globals stay unset. Both fail without the rework.
- `noteConfigSkipped` has its `@param`.

Full CLI suite is green locally except `build.test.mjs`/`build-theme.variants.test.mjs`, which fail identically on clean main in my environment (they need built theme dists).

cixzhang (changes requested):
Thanks for routing `doctor` and component docs through the gate — both new probes pass, and the missing JSDoc is fixed. We previously asked for every checkout-backed loader; that part is still incomplete.

`ASTRYX_NO_PROJECT_CODE=1 astryx theme build hostile-theme.mjs` still executes the file and emits output because `importThemeModule` calls jiti directly. The rewrite also leaves exact-head CI red: API type generation reports 24 errors and strict lint reports two unused imports.

Could you gate the theme source before either loader runs and get CI green? If you'd rather talk it through with someone, we're in [Discord](https://discord.com/invite/XnsUcFykEP).

[Reviewed by Robohands]

bhamodi:
Thanks for the precise repro. Rebased onto current main and closed the remaining gaps:

- `theme build`: the source is now refused in `themeBuildInternal` right after the file-exists check, before jiti (sync or async), the family preloader, or the eval fallback is created. `themeBuildFamily` also refuses every planned source up front. The command exits 1 with `ERR_THEME_LOAD`, prints one line naming the variable, and writes nothing. Covered end-to-end for the text, `--json`, and `--family` paths, plus a control showing the same fixture executes and builds with the variable unset.
- Audit: every `import()`, jiti, or `require` of a checkout path in `packages/cli` now either goes through `importUserModule` or asks the same gate (`template-adapter`'s JSX jiti, external blocks). `Project.load` also stops autolink and the local-package manifest, which main added since the first round. The full table of sites is in the PR body. The one intentionally untouched site is `resolve-theme.mjs`, which #5525 owns.
- The end-to-end suite grew to 12 probes (doctor, component including the fuzzy path, search, docs, template, theme build). I mutation-tested each guard: removing it fails its test.
- CI: `typecheck:strict`, the api-types verify script, strict lint, and the structure and changeset checks are green locally. The README now describes exactly what is gated.

packages/cli/foundation/config/project.mjs
bhamodi: Still in place on the rebased head: `noteConfigSkipped` carries `@param {string} configPath` and `typecheck:strict` passes.

cixzhang (changes requested):
The trust-boundary goal is right, but the current implementation is not safe or compatible enough to land at exact head `a0348cf9cc621bf3e027517ef94f246f0cccac24`. The opt-in environment switch still allows theme execution on text `component --list`, refuses legitimate installed Core docs needed by ordinary inspection, and lets duplicate config files fail before the safe-mode decision. Please wait for PR 6491 to settle discoverability/environment-control policy, then rework the boundary so supported safe inspection is discoverable, trusted installed/version-matched Astryx data is distinguished from checkout code, project code is not executed on the safe path, and duplicate project config is ignored rather than parsed there. The static theme-metadata direction from PR 5525 should remove the theme execution path rather than add another gate.

PR #5532, opened 2026-08-25
fix(docsite): state the playground page's own framing headers

bhamodi:
## Summary

Only `/playground/preview` carried framing headers; the `/playground` page itself had none, and its non-embeddability relied on the nested preview refusing to load when the ancestor chain includes another site. That works (frame-ancestors checks the full chain), but the parent page's protection shouldn't hinge on a child frame's headers — it now states them directly.

## Changes

- `/playground` gets `X-Frame-Options: SAMEORIGIN` and `Content-Security-Policy: frame-ancestors 'self'`, matching the framing posture the preview route already declares. No other CSP directives added — the page's script/style needs (Monaco, fonts) are unchanged.

## Test plan

- Config parses (`node --check`); prettier clean.
- Header shape mirrors the existing `/playground/preview` entry that production already serves.

## Notes

- Docsite is a private app: no changeset.

### DIFF
--- a/apps/docsite/next.config.mjs
+++ b/apps/docsite/next.config.mjs
@@ -22,6 +22,17 @@ const nextConfig = {
   async headers() {
     return [
       {
+        // The playground page itself also refuses embedding: today the
+        // nested preview's own frame-ancestors already breaks any embedding
+        // chain, but that protection shouldn't hinge on a child frame's
+        // headers — the parent states it directly.
+        source: '/playground',
+        headers: [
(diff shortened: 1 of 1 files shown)


cixzhang (approved):
Thanks, this is right — `/playground` really was serving no framing headers. Every other route still serves none; a `/:path*` default would cover them. Fine as a follow-up.

[Full review](https://github.com/cixzhang/astryx/wiki/Review-5532)

[Reviewed by Robohands]

PR #5533, opened 2026-08-25
fix(scripts): trusted-publishing default points at release.yml

bhamodi:
## Summary

`release.yml`'s header states the constraint: npm allows exactly one trusted publisher per package, ALL npm publishing lives in `release.yml`, and the trust config must name it (it even prints the exact `--workflow release.yml` re-point command). But `scripts/npm/setup-trusted-publishing.mjs` still defaulted `--workflow` to `deploy.yml` — left over from before publishing moved; `deploy.yml`'s own header now says "WEBSITE-ONLY — it does not publish to npm". A maintainer re-running the script with defaults would re-point trust at a workflow that never publishes and break the next release at the OIDC check.

## Changes

- Default `--workflow` is `release.yml`; the file-header rationale and the option comment now describe the actual publish topology and cross-reference release.yml's header.
- No behavioral change for anyone passing `--workflow` explicitly.

## Test plan

- `node --check` and prettier clean; grep confirms no remaining `deploy.yml` references in the script.
- Cross-checked against `release.yml` (single source of all npm publishing, `id-token: write` jobs) and `deploy.yml` (website-only header).

## Notes

- Maintainer script only (never runs in CI): no changeset.

### DIFF
--- a/scripts/npm/setup-trusted-publishing.mjs
+++ b/scripts/npm/setup-trusted-publishing.mjs
@@ -8,7 +8,7 @@
  * active npm session) that prepares every publishable @astryxdesign/* package
  * for npm TRUSTED PUBLISHING: OIDC-based publishing from GitHub Actions with no
  * long-lived npm token. Once configured, the `publish`/`canary` jobs in
- * .github/workflows/deploy.yml can publish with `--provenance` and zero secrets.
+ * .github/workflows/release.yml can publish with `--provenance` and zero secrets.
  *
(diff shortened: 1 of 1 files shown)


cixzhang (approved):
Thanks, this is right — release.yml is the only thing publishing. `docs/release.md:73,75` carries the same stale `deploy.yml` and would still point you at it; worth folding in here.

[Full review](https://github.com/cixzhang/astryx/wiki/Review-5533)

[Reviewed by Robohands]

PR #5538, opened 2026-08-26
fix(TransferList): clear the three open audit BLOCKs (A10, T17, I1)

ernestt:
## What this does

Clears the three open audit BLOCKs on `lab/TransferList` (recorded C 78.6 under rubric v1.8), so it can promote to core as a near-pure rename.

| Rule | Finding | Fix |
|---|---|---|
| **A10** — target size | `Clear` (35×20) and `Add all` (47×20) rendered below the WCAG 2.5.8 24px floor, no coarse-pointer expansion | Added `min-block-size: --size-element-sm` (28px) to the header-action style. They keep `height: auto` so the text baseline still aligns with the panel heading; only the hit target grows. Width already exceeded 24px. Now **35×28 / 47×28**. |
| **T17** — icon registry | Three inline `<svg>` glyphs bypassed the icon system; the remove `<svg>` (an X) shadowed the registry `close` glyph | `remove` now uses `<Icon icon="close">` (registry). `add`/`reorder` route through `<Icon icon={PlusIcon}>` / `<Icon icon={GripVerticalIcon}>` — the registry has no `add`/`grip` name, so these stay local icon components, matching the sibling `ListInput` convention. No new dependency (the system ships no icon library). |
| **I1** — i18n | The ungrouped-bucket fallback heading rendered a hardcoded `'Other'` literal | Routed through the catalog: new `@astryx.transferList.ungroupedLabel` key. It was the only string the previous localization pass missed. |

## Visual results (before / after)

The fixes are visually invisible at rest — the header actions keep their text-link look (the larger hit area is transparent) and the registry `close` / local `plus`+`grip` glyphs render identically to the removed inline SVGs. RTL still mirrors.

before/after

## Verification

All measured against a real Chromium build of Storybook:

- **A10**: header actions measure `Clear 35×28`, `Add all 47×28` — both axes ≥24px, confirmed under `any-pointer: coarse`.
- **T17**: add/remove/reorder each render a glyph through the `astryx-icon` pipeline; no `lucide-*` inline classes remain.
- **I1**: mixed grouped + ungrouped options render a localized "Other" heading via `useTranslator`.
- axe: **0 violations** across the TransferList stories (unchanged).
- 38/38 unit tests pass, `lint:strict` clean, `typecheck:docs` clean, `check:sync` / `check:changesets` / `check:i18n-catalog` all green.

## Re-audit

Fixing all three lifts the ceilings on §1 (Accessibility), §2 (Theming) and §9 (i18n & RTL), which were otherwise strong. Projected: **~89 (B)** with zero open BLOCKs — clearing the promotion gate. The ledger will be updated with the re-audit once this lands.

### DIFF
--- a/packages/lab/src/TransferList/TransferList.tsx
+++ b/packages/lab/src/TransferList/TransferList.tsx
@@ -26,6 +26,7 @@ import {
   type KeyboardEvent as ReactKeyboardEvent,
   type PointerEvent as ReactPointerEvent,
   type ReactNode,
+  type SVGProps,
 } from 'react';
 import * as stylex from '@stylexjs/stylex';
 import {flushSync} from 'react-dom';
@@ -44,6 +45,7 @@ import {useTranslator} from '@astryxdesign/core/i18n';
 import {
   borderVars,
   colorVars,
+  sizeVars,
   spacingVars,
   typeScaleVars,
 } from '@astryxdesign/core/theme/tokens.stylex';
@@ -56,57 +58,36 @@ import {
 import {reorderStyles} from '../reorderStyles';
 import {transferListVars} from './tokens.stylex';
 
-function PlusIcon() {
+// Lucide glyphs the shared icon registry does not carry (there is no `add` or
+// `grip` semantic name). Typed as icon components and rendered through `Icon`
+// so sizing, colour and a11y come from the icon pipeline rather than a raw
+// inline `<svg>`. Removal reuses the registry `close` glyph instead.
+function PlusIcon(props: SVGProps<SVGSVGElement>) {
   return (
     <svg
-      className="lucide lucide-plus"
-      width={16}
-      height={16}
       viewBox="0 0 24 24"
       fill="none"
       stroke="currentColor"
       strokeWidth={1.5}
       strokeLinecap="round"
       strokeLinejoin="round"
-      aria-hidden>
+      {...props}>
       <path d="M5 12h14" />
       <path d="M12 5v14" />
     </svg>
   );
 }
 
-function RemoveIcon() {
+function GripVerticalIcon(props: SVGProps<SVGSVGElement>) {
   return (
     <svg
-      className="lucide lucide-x"
-      width={16}
-      height={16}
       viewBox="0 0 24 24"
       fill="none"
       stroke="currentColor"
       strokeWidth={1.5}
       strokeLinecap="round"
       strokeLinejoin="round"
-      aria-hidden>
-      <path d="M18 6 6 18" />
-      <path d="m6 6 12 12" />
-    </svg>
-  );
-}
-
-function GripVerticalIcon() {
-  return (
-    <svg
-      className="lucide lucide-grip-vertical"
-      width={16}
-      height={16}
-      viewBox="0 0 24 24"
-      fill="none"
-      stroke="currentColor"
-      strokeWidth={1.5}
-      strokeLinecap="round"
-      strokeLinejoin="round"
-      aria-hidden>
+      {...props}>
       <circle cx="9" cy="5" r="1" />
       <circle cx="9" cy="12" r="1" />
       <circle cx="9" cy="19" r="1" />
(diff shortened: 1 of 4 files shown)


cixzhang (approved):
Thanks, all three blockers are clear.

[Full review](https://github.com/cixzhang/astryx/wiki/Review-5538)

[Reviewed by Robohands]

packages/lab/src/TransferList/TransferList.tsx
```diff
@@ -276,7 +257,13 @@ const styles = stylex.create({
     borderBlockEndColor: colorVars['--color-border'],
   },
   headerAction: {
+    // A text-link treatment, but the hit target still has to clear the WCAG
+    // 2.5.8 24px floor. The link keeps `height: auto` so its baseline aligns
+    // with the panel heading; `minBlockSize` gives the button a >=24px target
+    // (the sm element height) without changing where the text sits, and the
+    // width already exceeds 24px.
     height: 'auto',
+    minBlockSize: sizeVars['--size-element-sm'],
```
cixzhang, line 266: This remains 28px on coarse pointers; might add the ~44px expansion before promotion.

packages/lab/src/TransferList/TransferList.tsx
```diff
@@ -56,57 +58,36 @@ import {
 import {reorderStyles} from '../reorderStyles';
 import {transferListVars} from './tokens.stylex';
 
-function PlusIcon() {
+// Lucide glyphs the shared icon registry does not carry (there is no `add` or
```
cixzhang, line 61: These still bypass semantic registry keys, so themes can’t replace add/reorder.

packages/lab/src/TransferList/TransferList.tsx
ernestt: Good catch — the 24px floor satisfies WCAG 2.5.8 AA for fine pointers, but coarse-pointer contexts (mobile) want the 44px AAA target. I'll bump the coarse-pointer expansion before promotion lands.

packages/lab/src/TransferList/TransferList.tsx
ernestt: Agreed — the shared icon registry doesn't carry `add` or `grip` semantic names yet (noted in the comment above these). I'll register them in the icon pipeline as part of promotion prep so themes can override the visuals.

packages/lab/src/TransferList/TransferList.tsx
ernestt: I think the A10 minimum target size criteria does not capture the full detail and exceptions.  https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum.html

Links and inline text are generally exempt.  And the minimum is 20px if there are at least 4px between the next interactive element.  24px if side by side.  Can we update the A10 criteria to capture more nuance?

ernestt: Done — updated A10 in the [Component Audit Rubric](https://github.com/facebook/astryx/wiki/Component-Audit-Rubric#1--accessibility--operable-paths--weight-16) (now v1.12) to carry the full WCAG 2.5.8 instead of just the 24×24 headline. It now rules out the five SC exceptions before blocking:

- **Spacing** — an undersized target passes if a 24px-diameter circle centered on its bounding box clears every neighbor. Your 20px/4px case is the calibration: two 20×20 targets 4px apart pass (centers 24px apart, circles tangent); touching/side-by-side they fail; equal targets of side `s` need `24 − s` px of clearance.
- **Inline** — targets in a sentence or sized by the surrounding line-height (prose links) are exempt.
- **Equivalent / User-agent / Essential** — same function via a ≥24px control elsewhere on the page; native rendering the author hasn't resized; presentation required or legally mandated.

Also fixed the stray "~44px (WCAG 2.5.8)": 44px is **2.5.5 Target Size (Enhanced), AAA** / platform HIGs, so it's kept as a FIX-level house rule, not something the 2.5.8 citation implies. Same nuance mirrored into the Accessibility Checklist §Target size.

On this element specifically: `headerAction` is a ghost `Button` with a link-ish treatment, but it's a standalone control in the panel header — not prose — so the **Inline** exemption doesn't apply and holding the 24px floor here (as the comment does) is correct. It renders `minBlockSize: --size-element-sm` (28px) with width > 24px, so it clears the AA floor on its own without needing the spacing exception. The one open item left on this line is the ~44px coarse-pointer expansion noted above, which under the updated A10 is explicitly the 2.5.5 AAA / house rule → FIX before promotion, not a BLOCK.

PR #5543, opened 2026-08-26
feat(theme): ordered environmental adaptations in defineTheme

imdreamrunner:
## Why

Themes need opt-in adaptations for environmental conditions without swapping themes, handwritten media-query CSS, or React styling state.

This replaces the superseded `mobile` / `tablet` / `desktop` / `wide` API previously prototyped in this PR with the accepted contract from #5806.

## API

```ts
defineTheme({
  name: 'acme',
  adaptations: {
    widthBreakpoints: {
      sm: 640,
      md: 768,
      lg: 1024,
      xl: 1280,
      '2xl': 1536,
    },
    rules: [
      {
        when: {
          width: {from: 'lg', below: 'xl'},
          pointer: 'coarse',
          contrast: 'more',
        },
        value: {
          components: {/* overrides */},
        },
      },
    ],
  },
});
```

## Semantics

- Width points are fixed named tier starts; `from` is inclusive and `below` is exclusive.
- `width`, `pointer`, `contrast`, and `motion` fields in one `when` are ANDed.
- Rules cascade in authored order. Later matching writes win, including a deliberate write back to the root value.
- Media-surface overrides (`onDark` / `onLight`) remain more specific and emit after adaptations.
- A breakpoint map without value rules emits no media-query CSS.
- Child themes inherit breakpoint overrides and ordered rules, append their own rules, and re-resolve inherited values against the child root metadata.
- Adaptation `localTokens` may replace only exact names already enrolled by the root lineage.
- Custom component visual-prop values must be introduced on the root theme before a rule styles them.

## Implementation

- One shared value resolver serves root themes and adaptation rule values.
- Runtime injection and `astryx theme build` use the same ordered CSS generator.
- Built modules retain normalized breakpoint/rule data, generative axes, and local-token lineage required for source-equivalent `extends`; resolved CSS layers stay in the stylesheet only.
- `AppShell.mobileNav.breakpoint` now accepts `sm | md | lg | xl | 2xl | none`, reads the nearest active Theme, and switches to the wider layout at equality.
- Theme docs, template guidance, architecture records, and the breaking changeset now describe the accepted API.

## Validation

- `pnpm build`
- `pnpm lint:strict`
- Core, Lab, CLI, docs, template, and knowledge checks
- 3,129 CLI node tests passed in the isolated node-project run
- 1,181 focused theme/AppShell/layout tests passed after review fixes
- Four independent review passes; final pass returned no actionable issues
- Storybook production build
- Real Chromium: exclusive width edge, pointer-only and combined conditions, reduced motion, nested theme scopes, and CSS present at first paint

> **Status:** Draft. Keep this PR in draft until the implementation and accepted specification are reviewed together.


rubyycheung (approved):

cixzhang (changes requested):
Thanks — runtime/build parity and the boundary CSS are carefully covered.

One correctness issue still blocks: an unchanged child theme loses inherited explicit overrides inside a tier. At 600px, Chromium changed the inherited accent from red to blue and dropped the heading override. Could you preserve the effective inherited token/component overrides and add that child-theme regression?

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks — the inherited-override fix is right, and deriving it from the axes rather than carrying declarations is the right call.

It doesn't merge, and so CI has run nothing on this head: no build, no tests, no lint. The `light-dark()` guard is the one conflict that isn't mechanical — [#5566](https://github.com/facebook/astryx/pull/5566) moved it off the generated CSS, yours widened it to read the tier CSS, and merged it needs both the theme's own values and each tier layer's. Your test at `build-theme.tiers.test.mjs:175` is the check.

The second is the shape you just fixed. Deriving pinned-ness from values can't see a pin equal to what the axis generates, so it loses inside a tier: `'--font-size-base': '0.875rem'` on a 14/1.2 scale emits `1rem` in `mobile`; `2rem` holds. `themeTiers.ts:478` calls that harmless — true at the theme level, not in a tier.

Is there a cheap way to keep that case without the 10.5 KB?

[Full review](https://github.com/cixzhang/astryx/wiki/Review-5543)

[Reviewed by Robohands]

cixzhang:
Exploration suggestion — no change request yet.

While reviewing the proposed API alongside [AST-012 #5806](https://github.com/facebook/astryx/pull/5806), we pressure-tested the authoring shape against the existing theme contracts and [AST-002’s](https://github.com/facebook/astryx/blob/a1f0ddebd355e99e33732f4be89aa8d69da84f2f/docs/specs/AST-002/spec.md) understandability / dependable-result rubric. AST-002 formally governs component APIs today; I’m using the same questions here as a rubric, not claiming it already decides theme authoring.

The capability itself makes sense: themes need opt-in adaptations for width and primary-pointer precision, and may later need separately approved axes such as increased contrast. The open question is which API shape gives theme authors useful freedom without promising behavior the compiler cannot validate and preserve through `extends`.

### 1. Current proposal: top-level width tiers + nested literal media keys

```ts
defineTheme({
  mobile: {
    maxWidth: 756,
    tokens: {'--spacing-4': '12px'},
    '@media (pointer: coarse)': {
      tokens: {'--size-element-md': '40px'},
    },
  },
  tablet: {extends: 'mobile'},
});
```

**Strengths**
- Terse for width-first authoring.
- Matches the implementation and evidence already built here.
- Fixed sparse tiers, shared resolution, and tier inheritance constrain the hardest parts well.

**Concerns**
- The string looks like broad CSS media-query support, although only two exact pointer keys work.
- Structurally makes pointer subordinate to width; it cannot naturally express a pointer-only adaptation across all widths.
- Each future axis introduces another nesting convention and another inheritance layer.

**Spec fit:** strong for the current width-tier contract, weaker on API clarity and future axes because syntax suggests a broader promise than the implementation makes.

### 2. Flat raw media-query overrides

```ts
defineTheme({
  media: {
    '(width <= 756px)': {tokens: {'--spacing-4': '12px'}},
    '(pointer: coarse)': {tokens: {'--size-element-md': '40px'}},
  },
});
```

**Strengths**
- Flat and familiar to CSS authors.
- Naturally represents width-only, pointer-only, combined, and future browser conditions.

**Concerns**
- Permanently commits `defineTheme` to arbitrary media-query parsing and support.
- Makes overlap precedence, equivalent spellings, invalid queries, inheritance identity, diagnostics, and build-tool introspection part of our API contract.
- Array/object ordering can accidentally become cascade policy.
- Harder to guarantee that runtime and static builds understand every accepted query identically.

**Spec fit:** maximum flexibility, but the weakest AST-002 fit without concrete caller cases that justify the full CSS grammar and a dependable overlap/merge contract.

### 3. Flat, closed theme adaptations

```ts
defineTheme({
  adaptations: {
    widths: {
      mobile: {maxWidth: 756},
      tablet: {extends: 'mobile'},
    },
    rules: [
      [{width: 'mobile'}, {tokens: {'--spacing-4': '12px'}}],
      [{pointer: 'coarse'}, {tokens: {'--size-element-md': '40px'}}],
      [
        {width: 'mobile', pointer: 'coarse'},
        {components: {/* narrow + coarse only */}},
      ],
    ],
  },
});
```

A later independently approved axis could add a closed field such as `contrast`; this shape does not pre-approve that capability.

**Strengths**
- Flat like a condition→values map, without raw CSS strings.
- Supports width-only, pointer-only, and combined adaptations without nesting one axis under another.
- Structured condition objects prevent duplicate values for one axis and give errors useful paths.
- The normalized condition object can be the merge identity across theme `extends`; rule-array position need not affect semantics.
- Future axes can be added intentionally as closed fields after their own design/compiler review.

**Concerns**
- More authoring and implementation machinery than the current proposal.
- Needs an explicit conflict model. A reasonable candidate is root < one-axis rule < strict-superset rule; incomparable matching rules that write the same leaf fail unless a combined rule resolves it.
- Width-tier inheritance must rewrite the width identity while preserving other axes, then flatten before CSS emission.
- `adaptations` and `rules` are new vocabulary that need usability review.

**Spec fit:** strongest constrained/general contract, but only if canonical identity, precedence, conflicts, inheritance, and diagnostics are specified before implementation.

### Requirements whichever shape wins

- No declared adaptation means no additional CSS or behavior.
- Width tier names remain fixed and sparse; bounds are globally defaulted and theme-overridable.
- Only value-producing axes participate; identity, registries, and syntax metadata stay outside.
- Root and conditional values use the same resolver and preserve explicit pins through theme extension.
- Runtime and static builds consume one normalized representation.
- Unsupported conditions and ambiguous overlapping writes fail before output.

My current lean is option 3 because it leaves room for future *approved* environment axes without exposing arbitrary CSS or deep nesting. But this is an exploration, not a prescribed rewrite. @imdreamrunner, could you choose where you think the API should land after weighing implementation cost and the guarantees above? In particular, I’d value your view on whether the current shape can support pointer-only/future axes cleanly, and whether the flat structured model’s canonical merge/conflict rules are practical in the compiler.

rubyycheung:
@imdreamrunner I’m planning the component-side follow-up to AST-012 for the existing responsive surface consumers: the shared `useAdaptivePresentation` path used by Selector, MultiSelector, DropdownMenu, MoreMenu, ContextMenu, and the pending ComplexSelector work. The goal is to replace the hard-coded 768px query with the effective nearest-Theme `md` breakpoint once this implementation exposes the accepted AST-012 shape. Are you already covering that shared component resolver here or in another branch? I’ll avoid touching your PowerSearch, Tokenizer, Table, and date/time surface work.

imdreamrunner:
Revised this PR against the newly approved [AST-012 specification in #5806](https://github.com/facebook/astryx/pull/5806).

The earlier `mobile` / `tablet` / `desktop` / `wide` proposal has been replaced with the approved ordered `adaptations` API, including named width breakpoints, closed environmental conditions, source/build parity, extension semantics, local-token constraints, and the AppShell boundary behavior.

I also merged current `main` and iterated through independent review until the final pass reported no actionable issues. This remains a draft for review.

cixzhang (changes requested):
🔴 Blocking — two adaptation validators fail open.

Triage: new public theme API · breaking AppShell boundary · high blast radius → deep path.

- [Alias-typed visual props bypass validation](https://github.com/facebook/astryx/blob/e09d8ac85b3f63ab7b3decd8921730a29099a201/packages/cli/api/theme/build/build.mjs#L619-L623). `avatar-group size:giant` builds with no warning and emits `.astryx-avatar-group.giant`, although the closed `AvatarSize` API can never emit it. Please resolve aliases or fail closed.
- [Local-token cycles are checked one rule at a time](https://github.com/facebook/astryx/blob/e09d8ac85b3f63ab7b3decd8921730a29099a201/packages/core/src/theme/localTokens.ts#L321-L333). Overlapping `pointer: coarse` and `contrast: more` rules can create `a → b → a`; Chromium then invalidates both properties. Please reject cycles reachable across overlapping rules.

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks — the co-matching cycle fix now covers overlapping and inherited rules.

One blocker remains: `avatar-group` `size:giant` still builds successfully and emits `.astryx-avatar-group.giant`. `AvatarSize` is a closed alias, so no component state can ever match that CSS. The validator skips every axis whose docs do not enumerate literal values, turning the required fail-closed check into dead output with no diagnostic.

Please resolve alias-backed literal/numeric unions, or reject unenumerable rule-only values before writing output.

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks — the overlapping-cycle fix now holds across portable, local, inherited, and repaired rule cascades.

One blocker remains on this head: [alias-backed closed visual props still fail open](https://github.com/facebook/astryx/blob/ecaf8274be4b366abff553c7c1cb922a0dded0e7/packages/cli/api/theme/build/build.mjs#L619-L623). `avatar-group` `size:giant` exits 0 and emits `.astryx-avatar-group.giant`, although `AvatarSize` can never produce that state. Skipping every axis whose docs use an opaque alias leaves builders with dead CSS and no diagnostic. Please resolve alias-backed literal/numeric unions, or reject rule-only values when an axis cannot be enumerated.

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks — the ordered cascade, extension metadata, build parity, and exact-head Chromium matrix hold.

Two failures still block:

- [Alias-backed closed visual props fail open](https://github.com/facebook/astryx/blob/3dde1e22f8c44bae44672ae128fcd22254919eb7/packages/cli/api/theme/build/build.mjs#L627-L639). `avatar-group` `size:not-a-real-avatar-size` exits 0, emits unreachable CSS, and generates no type augmentation, although `AvatarSize` can never match it.
- [The CLI now requires the new Core export unconditionally](https://github.com/facebook/astryx/blob/3dde1e22f8c44bae44672ae128fcd22254919eb7/packages/cli/api/theme/build/build.mjs#L1479-L1485). Packed against published Core 0.5.4—still allowed by peer `*`—even a non-adaptation theme build exits 1.

Please resolve alias domains or reject unprovable rule-only values before output, and preserve old-Core builds or declare the required Core range with migration guidance.

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks — the published-Core compatibility path now preserves non-adaptation builds and rejects unsupported adaptation themes before output.

Two compatibility failures remain:

- [Existing `defineTheme({tokens})` inputs now throw](https://github.com/facebook/astryx/blob/b9ac7bef4a93b3b7a6010fd80b14428b999c4132/packages/core/src/theme/resolveThemeValues.ts#L92-L106) on non-string values accepted through casts or broad spreads. This can break themes that do not use adaptations; please keep strict validation scoped to the new paths.
- [Alias-backed closed visual props still fail open](https://github.com/facebook/astryx/blob/b9ac7bef4a93b3b7a6010fd80b14428b999c4132/packages/cli/api/theme/build/build.mjs#L805-L849): `avatar-group` `size:not-a-real-avatar-size` exits 0, emits unreachable CSS, and generates no augmentation.

Please preserve existing token inputs and either resolve alias-backed domains or reject unprovable rule-only values before output.

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks — legacy token inputs now hold, and compatibility, type generation, cascade, and the exact-head browser checks pass.

One blocker remains: [alias-backed closed visual props still fail open](https://github.com/facebook/astryx/blob/40db4da6e83d1c06eac3fb5943ff64c67c27d195/packages/cli/api/theme/build/build.mjs#L832-L844). `avatar-group size:not-a-real-avatar-size` exits 0, emits unreachable CSS, and writes no type augmentation. Please reject rule-only values that tooling cannot prove belong to the closed component surface before output.

[Reviewed by Robohands]

cixzhang (changes requested):
The revised spec in #6139 is now approved: adaptation component writes must reuse the exact shared root validation result, add only the no-conditional-enrollment check, and carry unresolved shared results unchanged. This PR is still based on the older #6139 head and must be restacked after that spec lands.

Please then demonstrate the implementation matches DEC-5 with focused fixtures for: a finite built-in absent from root; an authoritative open primitive absent from root; a root-enrolled custom value used by a rule; a rule-only custom value rejected; and the same unresolved shared result/diagnostic at root and in an adaptation. Keep the already-fixed old-Core and legacy-token compatibility regressions. After restacking, rerun exact-head full CI, `pr-a11y`, and `Stable visual regression`.

cixzhang (changes requested):
Thanks — old-Core compatibility, legacy token inputs, generated types, cascade, exact-head Chromium, and required CI checks are clean.

One architecture block remains. The build still uses a [docs-derived value index](https://github.com/facebook/astryx/blob/ef637643346dc1e3be8e3d2c5e701ee5ff305a33/packages/cli/api/theme/build/build.mjs#L248-L376) and [adaptation-only validator](https://github.com/facebook/astryx/blob/ef637643346dc1e3be8e3d2c5e701ee5ff305a33/packages/cli/api/theme/build/build.mjs#L798-L850). Root validation only checks component/axis names, so the paths happen to agree for today’s unresolved alias case but rules won’t inherit future root-validation improvements; the shipped docs also teach the superseded exception.

Please make root validation return one reusable target/axis/domain/extension result, consume it unchanged for rules, add only the conditional-enrollment check, and cover finite built-ins, open primitive values, root-enrolled custom values, rejected rule-only custom values, and identical unresolved results/diagnostics.

[Reviewed by Robohands]

cixzhang (changes requested):
Mouse users can still receive a touch-only Heading weight. A coarse-only `weight:bold:800` followed by a fine-pointer `level:2` rule emits `font-weight:800` inside the fine block because the cumulative fallback carries every earlier rule without checking co-matchability. Please preserve earlier leaves only from rules that can match the later condition, add coarse/fine regression coverage, and provide the AST-012 real-Safari evidence for exclusive width edges and primary-pointer classification.

[Reviewed by Robohands]

cixzhang (changes requested):
The feature's width and primary-pointer behavior is still unverified for Safari users. The coarse/fine Heading leak is fixed, and exact-head compatibility, generated types, template sync, CI, and Chromium checks pass. AST-012 requires real Safari—not Playwright WebKit—to prove exclusive width edges and primary-pointer classification. Please attach one Safari receipt covering 639/640 and coarse/fine results.

[Reviewed by Robohands]

cixzhang (approved):
Thanks — the coarse/fine Heading isolation now holds, and the exact-head Safari receipt covers exclusive width edges and primary-pointer classification.

[Reviewed by Robohands]

PR #5547, opened 2026-08-26
fix(Toast): the card's shadow is no longer clipped away

freddymeta:
## Summary

Each toast's grid row uses `overflow: hidden`. The clip is load-bearing while the row opens and closes — it makes the toast read as folding into the stack — but at rest the box hugs the card's border box **exactly**, with zero slack on any side:

```
slackTop: 0   slackRight: 0   slackBottom: 0   slackLeft: 0
```

So it also cuts off every shadow the card casts. Astryx's own `--shadow-med` is declared on `styles.root` and has been invisible.

The fix keeps that cross-engine clip for entry and exit, and releases it only while the toast is settled:

```diff
 toastWrapperInner: {
   overflow: 'hidden',
 }
+toastWrapperInnerSettled: {
+  overflow: 'visible',
+}
 toastWrapperExiting: {
   gridTemplateRows: '0fr',
 }
```

When the opening `grid-template-rows` transition completes, the row is marked settled and the inner box becomes `overflow: visible`. Dismissal removes that settled state synchronously, restoring the original `hidden` paint boundary for the whole exit. The wrapper keeps main's normal pointer boundary: while the row is visible it absorbs a second Dismiss click rather than passing it through to an obscured control; as the row shrinks, `hidden` is what removes an out-of-row Undo from hit testing.

Settled state lives on the mounted row rather than in a set of ids on the viewport, so it cannot outlive the row it describes — `maxVisible` eviction and a `uniqueID` overwrite both unmount a row without running the dismissal path. Only the row's own `transitionend` is read, since `grid-template-rows` is not private to the wrapper and the event bubbles.

## Corrections to my earlier claims in this PR

Measured today, on this head, in Chromium 149. Three things I asserted earlier do not survive measurement, and I would rather retract them here than leave them in the record:

1. **The no-transition fallback was justified with a case that does not exist.** I wrote that "a print or forced-colors context can resolve the wrapper to `transition-duration: 0s`". It cannot. Every supported path resolves the row to a non-zero duration:

   | path | computed `transition-duration` |
   |---|---|
   | default | `0.175s` |
   | `prefers-reduced-motion: reduce` | `1e-05s` |
   | `forced-colors: active` | `0.175s` |
   | `media: print` | `0.175s` |

   The repo ships no `@media print` stylesheet and no forced-colors motion reset, and reduced motion keeps 0.01ms precisely so the event still fires. @cixzhang was right that the fallback was too broad; it is gone, and the row now performs **zero** `getComputedStyle` reads.

2. **The shadow's reach is 12px, not the 14px I claimed.** Sampling straight down from the card's bottom border box on a white page: `223` at +0px, then `237 → 243 → 245 → 247 → 248 → 250 → 251 → 252 → 253 → 254 → 254`, and white from +12px. Before the fix every one of those offsets is `255,255,255`.

3. **I said the `no-style-only-wrapper` warning "went with the extraction". It did not.** The inner clip `<div>` still raises exactly one such warning — and `origin/main` raises the identical warning on the same file. It is pre-existing, not introduced or removed here.

## Measured in Chromium 149, this head

Real browser, real pointer, stock theme, white page. The card's own opacity transition is slowed alongside the row's for the mid-flight frames, so a captured pixel is a clip result and not an already-faded card — each row below records the card's opacity at the moment it was sampled.

| phase | inner `overflow` | card opacity | card overhang below row | pixels below the row |
|---|---|---|---|---|
| mid-entry | `hidden` | 0.56 | 26.27px | `255,255,255` ×14 |
| settled | `visible` | 1 | 0 (slack 0) | `223 … 254` then white at +12px |
| dismissal, first frame | `hidden` | 1 | — | wrapper `pointer-events: auto` |
| mid-exit | `hidden` | 0.23 | 46.23px | `255,255,255` ×14 |

**Negative controls**, same instants, same coordinates:

- Clip forced off during entry → `113,118,121` where the fix paints white. Mid-exit → `204,206,206`. So those white rows are the clip holding, not an invisible card.
- **This PR reverted to `origin/main` and rebuilt** → the clip is `NEVER RELEASED` (settle times out), and all 14 sampled offsets below the card are `255,255,255`. That is the bug.
- Baseline with no toast shown at all → the same column is `255,255,255`.

## What dropping the fallback costs

On a row forced to `transition-duration: 0s` — a path nothing supported reaches — the clip is never released and a dismissed toast stays mounted. Measured on **both** this head and `origin/main`: `stillMountedAfterDismiss: 1`. Main's unmount is likewise driven only by `transitionend`, with no fallback and no timer backstop, so this is main's existing behaviour rather than something this PR introduces. If it is worth fixing it is a separate bug on main's exit path, not scope for a shadow fix.

## Records

No public surface is added: no new prop, export, theme target, or public custom property. `boxShadow` is already in the approved paint catalog and is already declared on the existing `astryx-toast` target; this change only makes it observable. Under `architecture:public-component-api` that is a `preserves` bug fix requiring regression evidence, not an API decision — so `spec:AST-002/DEC-1` and FR12 do not bite. Anatomy is unchanged, and the clip wrapper is layout plumbing that INV4 keeps targets off, so no `.spec.md` anatomy block is owed.

## Test plan

- `pnpm -F @astryxdesign/core build` — exit 0
- `pnpm exec vitest run packages/core/src packages/lab/src` — **316 files / 8951 tests, all pass**
- `pnpm -F @astryxdesign/core typecheck` / `typecheck:docs` — exit 0
- `pnpm check:repo` — exit 0
- `pnpm lint:strict` — exit 0, 85 warnings (the count `origin/main` reports), 0 errors
- Chromium matrix and negative controls above

WebKit 26.5 was **not** re-run for this revision: this push changes only the fallback, leaving the overflow mechanism byte-identical to the one verified there on 27 Aug, and the build box cannot launch WebKit (`playwright install webkit` fails host requirements). Say the word if you want a WebKit pass before merge.

`pr-rtl` is soft-red repo-wide (empty `verified-not-applicable.json`, 229 gaps across 246 components, `continue-on-error: true`) and is not related to this change.

## Where this came from

Found auditing an internal design-system adapter's CSS overrides. It carried padding + negative-margin arithmetic to make room for the shadow. Its diagnosis was right; the upstream fix is to release the clip when it is no longer serving the collapse.

@freddymeta


rubyycheung (approved):
Looks good from design. Restoring the shadow gives the toast the intended elevation and separation from the page, while stacked toasts remain visually clean and the dismissal animation still reads correctly.

cixzhang (changes requested):
Thanks — the Chromium shadow and collapse both work.

`ToastViewport.tsx:101` still needs a cross-engine fix. WebKit 26.5 does not support `overflow-clip-margin`; all 15 sampled rows below the card stayed white, so the shadow remains clipped. In Chromium, the 32px margin also expands hit testing during exit: with the clip box 2.73px tall and the toast 23% opaque, a real click on the out-of-bounds Undo action fired; the parent did not.

Could we preserve the old exit hit boundary and make the resting shadow visible in WebKit too?

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks — both requested fixes now pass in Chromium and WebKit.

A person who clicks Dismiss twice can activate an obscured control underneath while the Toast is still fully visible. `pointerEvents: 'none'` at `ToastViewport.tsx:128` removes the whole exiting Toast from hit testing immediately. Main intercepts that second click, and deleting this declaration preserves the WebKit shadow and late-exit Undo fix.

Could we drop it before this lands?

[Reviewed by Robohands]

cixzhang (approved):
Thanks — both prior browser issues are fixed in Chromium and WebKit. Before merge, could we prune stale settled IDs on overwrite and correct the changeset’s hit-testing claim?

[Reviewed by Robohands]

freddymeta:
Both done.

**Stale settled ids on overwrite.** You were right that nothing pruned them. An overwrite swaps a new entry, with a new id, into the replaced toast's place — the old row unmounts without ever being dismissed, so neither `removeToast` nor `handleExited` runs for it, and its id sat in `settledIds` for the life of the viewport. One dead id per update, for exactly the pattern `uniqueID` exists to serve: "Saving…" overwritten by "Saved", a progress toast rewritten per chunk.

It is pruned where the overwrite is resolved, against the committed list rather than inside the `setToasts` updater — React may run that more than once, and the `ignore` check just above it already reads the list the same way.

Worth stating plainly: **the prune is not observable from the DOM.** Ids come from a monotonic counter and are never reused, so a leaked one changes nothing a test can assert — it is a memory fix, not a behavioural one. What the new test does cover is the behavioural half, which I had not verified either: the replacement is a *different* row, so it must start clipped and release the clip on its own entry transition rather than inheriting the settled state of the toast it replaced. Confirmed in Chromium — `astryx-toast-2` settled → overwrite → `astryx-toast-3` mounts `hidden`, settles to `visible`, and its shadow paints the same `237 → 245 → 250 → 254`.

**The changeset's hit-testing claim.** Corrected — it was describing the `pointer-events: none` revision you had me remove, and it survived the edit. It now reads:

> During exit the row clips again, so anything outside the shrinking row is neither painted nor hit-testable — while the wrapper itself keeps the ordinary pointer boundary it has always had, and still absorbs a click aimed at a toast that is still on screen.

**One tidy-up while in there:** three byte-identical "remove an id from a Set, preserving the Set's identity when absent" updaters folded into one `withoutId` helper. Two are this branch's; the third is `handleExited`'s, which was already there — behaviour-identical, and leaving an inline copy beside the helper seemed worse than touching it.

Gates: full `packages/core` suite **283 files / 8010 tests pass**, `lint:strict` 0 errors (the same 80 warnings main reports), `check:repo` clean including `check:sync`, core typecheck clean.

One note on engines: this revision changes state pruning and comments only — the overflow mechanism is byte-identical to the one you verified — so I re-measured the full lifecycle in Chromium and did not re-run WebKit (the box I build on can't launch it). Say the word if you want a fresh WebKit pass before merge.

freddymeta:

freddymeta:
Correcting my previous comment — **this merge was not mechanical**, and I should not have described it that way.

`renderContent` landed on main in the meantime, touching the same files this PR touches: `Toast.tsx`, `ToastViewport.tsx`, `types.ts`, `index.ts`, `Toast.doc.mjs`, plus a new `ToastDismissSlot.tsx` and ~330 lines of tests. Git auto-resolved it, which is exactly the case where a clean textual merge can still be semantically wrong — so I checked rather than assuming.

**Both sides coexist.** Full Toast suite on the merged head: 50/50. My clip test (`releases the shadow clip only while the toast is settled`) and all of main's `renderContent` tests pass together, including the ones that matter for interaction with this change — `still renders a close, and it dismisses the toast` and `restores the fallback when a nested layout removes its close`.

The one interaction worth naming: main's fallback dismiss is `position: absolute` inside the card and gave `styles.root` a `position: relative`. That is inside the card's own bounds, so the wrapper clip this PR releases neither hides it nor is defeated by it.

**Re-verified the shadow claim in Chromium on the merged head**, sampling real screenshot pixels below a settled toast rather than reading computed style:

```
wrapper overflow: visible          ← the clip is released while settled
offset  0  →  rgb(229,229,229)
offset  4  →  rgb(243,243,243)
offset  8  →  rgb(250,250,250)
offset 12  →  rgb(254,254,254)
offset 14  →  rgb(255,255,255)     ← shadow fully resolved, 14px reach
```

14 painted rows, matching the changeset. Before the fix every one of those offsets was pure white.

Both of your pre-merge asks were already addressed in `151acf824` (pushed after your review): the replaced toast's settled id is pruned on overwrite, and the changeset's hit-testing sentence is corrected.

One thing I caused and should flag: **the merge dismissed your approval.** That was the cost of refreshing the expired visual evidence, and it needs a re-approve rather than just a merge.

cixzhang (changes requested):
The normal settled-shadow path works, but two lifecycle regressions remain: a toast resurfacing after `maxVisible` eviction keeps stale settled state and can paint/receive input outside its opening row; a descendant `grid-template-rows` transition can bubble and release clipping early.

Please make settled identity and transition handling lifecycle-safe, with a fallback for supported no-transition paths.

[Reviewed by Robohands]

freddymeta:
Both lifecycle regressions fixed in `826775b12`, each with a test verified to fail on the previous head.

**Stale settled identity.** You were right that eviction is a second door to the bug I only closed for overwrite on 27 Aug. Rather than prune a third path, the state moved: `settledIds` on the viewport is gone, and settled is now local state on the mounted row (`ToastRow`). A row leaves the DOM by more paths than dismissal — `maxVisible` eviction, a `uniqueID` overwrite — and none of them run the dismissal path, so any id-keyed set will keep drifting out of step with what is rendered. Held on the row, the flag unmounts with the row, and a resurfaced toast mounts clipped and runs its own entry transition. The synchronous re-clip on dismissal falls out of the same rule (`isSettled && !isExiting`) rather than being a second write, so the overwrite prune in `addToast` and the settled prune in `removeToast` are both deleted.

**Bubbled transition.** `e.target !== e.currentTarget` guard added. It cut both ways, as you say: early clip release on entry, and on exit a descendant's event called `handleExited` mid-collapse and unmounted the toast early. The test covers both directions.

**No-transition fallback.** A layout effect reads the row's computed `transition-duration` and, when every declared duration resolves to zero, settles or exits it directly. Deliberately conservative: an unresolved value reads as animated, so the event-driven path stays in charge everywhere it works, and reduced motion's 0.01ms is untouched — the existing test pinning that is unchanged and still passes.

**Verification.**

| | |
|---|---|
| Toast suite | 64/64 |
| `packages/core` suite | 286 files / 8306 tests, all pass |
| `pnpm -F @astryxdesign/core build` + `typecheck` | exit 0 |
| lint on both changed files | 0 problems (the head had 1 warning; the `no-style-only-wrapper` warning went with the extraction) |

Negative control on each guard, since green proves nothing on its own — with the `e.target` check deleted the bubbling test fails, and with the fallback effect deleted the no-transition test fails. The eviction test was written and confirmed red before any fix existed.

Changeset updated to describe all three.

cixzhang (changes requested):
Thanks — the lifecycle fixes now hold. Apps now pay two synchronous style reads per Toast to support a `0s` path that reduced motion, forced colors, and print do not use. At the default five Toasts this adds ten reads plus measured style/layout work ([`ToastViewport.tsx:259–272`](https://github.com/facebook/astryx/blob/5531484d8705b7b3a08172203d1b96b50a5269b7/packages/core/src/Toast/ToastViewport.tsx#L259-L272)). I asked for this fallback; that was too broad.

Could we drop it unless a current supported path needs it?

[Reviewed by Robohands]

freddymeta:
Dropped in `48d10988a`. You were right, and the justification I gave for it was wrong.

## The fallback

I defended it by saying "a print or forced-colors context can resolve the wrapper to `transition-duration: 0s`". I never measured that. It is false — measured today in Chromium 149 on the row wrapper at this head:

| path | computed `transition-duration` |
|---|---|
| default | `0.175s` |
| `prefers-reduced-motion: reduce` | `1e-05s` |
| `forced-colors: active` | `0.175s` |
| `media: print` | `0.175s` |

Nothing supported reaches `0s`. The repo ships no `@media print` stylesheet and no forced-colors motion reset, reduced motion keeps 0.01ms precisely so the event still fires, and jsdom computes `var(--duration-fast)` — the deleted test had to *mock* `getComputedStyle` to manufacture the case it was testing, which is the tell I should have read at the time.

`isTransitionless`, the layout effect, the row ref and that test are gone. Your count was right: two synchronous style reads per Toast, ten at the default five. It is now **zero** — `ToastViewport.tsx` contains no `getComputedStyle` call at all.

**What dropping it costs, measured rather than assumed.** On a row forced to `0s`, the clip is never released and a dismissed toast stays mounted. I ran that against `origin/main` as well, and main does exactly the same: `stillMountedAfterDismiss: 1` on both. Main's unmount is driven only by `transitionend`, with no fallback and no timer backstop. So this is main's existing behaviour on a path nothing supported reaches — not something this change introduces. If it is worth closing it is a separate bug on main's exit path.

I considered reusing `waitForTransition` from `BottomSheetPanel.tsx` instead, and rejected it: it is BottomSheet-private, and it reads `transition`, `transitionProperty`, `transitionDuration` and `transitionDelay` plus arms a timer — strictly more of the cost you objected to.

## The prior fixes, re-measured at this head

Not taken on trust after the merge. Chromium 149, real pointer, stock theme, white page; the card's own opacity transition is slowed alongside the row's, so a mid-flight pixel is a clip result and not an already-faded card — opacity at the sampling instant is in the table.

| phase | inner `overflow` | card opacity | card overhang below row | pixels below row |
|---|---|---|---|---|
| mid-entry | `hidden` | 0.56 | 26.27px | `255,255,255` ×14 |
| settled | `visible` | 1 | slack 0 | `223 … 254`, white at +12px |
| dismissal, first frame | `hidden` | 1 | — | wrapper `pointer-events: auto` |
| mid-exit | `hidden` | 0.23 | 46.23px | `255,255,255` ×14 |

Negative controls at the same instants and coordinates, because green proves nothing on its own:

- clip forced off mid-entry → `113,118,121`; mid-exit → `204,206,206`. The white rows above are the clip holding, not an invisible card.
- **this PR reverted to `origin/main` and rebuilt** → settle times out, clip `NEVER RELEASED`, all 14 offsets below the card `255,255,255`. The bug.
- no toast shown at all → same column `255,255,255`.

Row-local settled state, the eviction/overwrite paths and the `e.target !== e.currentTarget` guard are unchanged and still covered; Toast suite 67/67.

## Two more corrections to my own record

- **The shadow's reach is 12px, not the 14px I claimed twice.** `223` at +0px, fading through `237 → 243 → 247 → 250 → 253 → 254`, white from +12px. The changeset and PR body now say 12.
- **I said the `no-style-only-wrapper` warning "went with the extraction". It did not.** The inner clip `<div>` still raises exactly one, and `origin/main` raises the identical warning on the same file. Pre-existing either way — I should not have claimed credit for removing it.

## Merge

Merged `origin/main` (`366e4314a`, never rebased). #5822 is in and coexists: the viewport spans the full 1280px, and the toast lands on the end edge — 19px from the right in LTR, mirrored to 19px from the left under RTL — with the clip released in both. One note on your context: **#5872 has not landed**; it is still open against main, so nothing here sits on top of it.

## Gates

- `pnpm -F @astryxdesign/core build` — exit 0
- `pnpm exec vitest run packages/core/src packages/lab/src` — 316 files / **8951 tests, all pass**
- `pnpm -F @astryxdesign/core typecheck` and `typecheck:docs` — exit 0
- `pnpm check:repo` — exit 0
- `pnpm lint:strict` — exit 0, 85 warnings (main's count), 0 errors

WebKit 26.5 was not re-run for this push: it changes only the fallback, leaving the overflow mechanism byte-identical to the one you verified on 27 Aug, and this box fails `playwright install webkit` on host requirements. Happy to get a WebKit pass before merge if you want one.

On the records: no public surface is added — no prop, export, theme target or public custom property. `boxShadow` is already in the approved paint catalog and already declared on the existing `astryx-toast` target; this only makes it observable, which `architecture:public-component-api` classes as a `preserves` bug fix rather than an API decision, so `spec:AST-002/DEC-1` and FR12 do not apply. Anatomy is unchanged, and the clip wrapper is layout plumbing that INV4 keeps targets off, so no `.spec.md` anatomy block is owed here.

cixzhang (approved):
Thanks — the requested fallback is gone, and the shadow, lifecycle, and hit boundary still hold on this head. Looks good to me.

[Reviewed by Robohands]

freddymeta:
`pr-rtl` here is Toast's coverage gap, not a failure in this PR: the audit scopes only Toast, every dimension returns N/A, and unexplained all-N/A counts as unmeasured.

The coverage is #5872, which is a separate branch rather than an ancestor of this one. I verified the combination instead of duplicating it — this branch's `ToastViewport` plus #5872's story and target, built and run here:

```
pnpm rtl:audit -- --filter Toast
CUR  RTL-ready  core/toast  {"D4":"pass"}
```

So this check clears once #5872 lands. Adding the same story and target here would collide with it, so I have left this branch alone.

freddymeta:
Landing. The one red check is `pr-rtl`, which is Toast's coverage gap rather than a failure in this change — it is `continue-on-error` in `ci.yml`, and the coverage arrives with #5872. Verified the combination before landing: this branch's `ToastViewport` plus #5872's story and target gives `D4 pass, 0 gap`.

PR #5548, opened 2026-08-26
fix(docsite): stabilize PPR loading states

imdreamrunner:
The docsite had two first-paint problems: query-dependent sections collapsed to empty PPR holes, and the footer server-rendered its desktop layout at every width. This keeps the existing deep-link architecture while making both states stable and intentional.

### Loading experience: before and after

Side-by-side comparison at the same point in a throttled mobile load: before, the page is an empty shell with an overlapping footer; after, the Date Input heading and a geometry-matched loading skeleton are present.

| Route | Before | After |
|---|---:|---:|
| `/components/DateInput` | CLS **0.2281** | **0** |
| `/themes` | CLS **0.2061** | **0** |

Measured against a production build at 1.5 Mbps / 6× CPU throttling.

### Keep PPR and preserve deep links

This deliberately keeps both existing URL mechanisms:

- component tabs continue to use `useSearchParams()` for `?tab=properties` and `?tab=theming`
- the theme page continues to resolve `await searchParams` for `?theme=<slug>`

There is no client-side reimplementation and no one-frame default-tab/theme compromise.

### Make the component boundary smaller

Previously, the Suspense boundary wrapped the entire component page with no fallback. Under `cacheComponents: true`, the prerendered HTML was header + sidebar + footer around an empty `<template>`.

The static shell now includes the component title and package metadata. Only the query-dependent tab row and selected panel remain inside Suspense. Components with no tabs render their complete Overview statically.

The boundary has a real Skeleton fallback matching the tab row, preview, heading, and prose. Its reserved height keeps the footer below the viewport while the panel streams.

### One shared Themes layout for both states

The resolved mobile and desktop layouts are unchanged. The PR extracts two shared pieces used by both resolved and fallback rendering:

- `ThemeHeading` owns the real heading, explanatory copy, docs link, and mode-control slot.
- `ThemeExplorerLayout` owns every responsive structural region: outer shell, sticky sidebar, sidebar surface/content, action stack, mobile toolbar/context/carousel and item sizing, right column, and preview bounds.

Both states compose those same primitives at the same positions. Only query-dependent contents differ: live controls/cards/preview in the resolved state, same-size Skeletons in the fallback. A future responsive redesign therefore updates both states by construction rather than relying on copied CSS or parallel markup.

The handoff is stable on mobile: both fallback and resolved headings measure **35px / 44px** at the same coordinates. The entire temporary visual fallback is `inert`, so its docs link, Carousel scroller, and any future control are excluded from focus and interaction by construction. A separate visually-hidden `role="status"` remains outside the inert subtree to announce “Loading theme explorer.”

<details>
<summary>Theme-page PPR shell on a throttled mobile load</summary>

Themes loading state reusing the same ThemeHeading component in the same centered mobile layout as the resolved explorer; only selected-theme controls, cards, and preview use skeleton placeholders.

</details>

### Make the footer responsive in CSS

`SiteFooter` previously branched on `useAppShellMobile().isMobile`. That hook's server snapshot is always `false`, so every viewport first received the desktop five-column grid; on phones its regions painted on top of each other until hydration.

The footer now uses one DOM structure with a CSS media query. The correct layout exists on the first paint without JavaScript. Desktop geometry is unchanged from production, verified track by track.

## Regression coverage

`static-shell.test.ts` now checks that:

- the component heading remains outside the query-dependent boundary
- the component and theme boundaries use named, non-null fallbacks
- the fallback and resolved explorer both use the same `ThemeHeading` and `ThemeExplorerLayout` primitives
- the global footer does not return to JavaScript-based viewport branching

## Verification

- latest `main` rebased cleanly
- docsite TypeScript: pass
- docsite tests: **445 passed**
- focused ESLint: pass
- production build: pass
- throttled mobile CLS: **0** on DateInput and Themes
- native `?tab=` deep links, tab clicks, and `?theme=butter`: pass
- main’s new Accessibility tab + `?tab=accessibility`: preserved and verified
- hydration errors: **0**
- entire fallback visual tree: `inert` (30 Tab presses reached zero descendants)
- loading announcement: visually-hidden status outside the inert subtree
- fallback → resolved mobile heading: identical 35px / 44px geometry

@imdreamrunner

### DIFF
--- a/apps/docsite/src/app/(site)/themes/page.tsx
+++ b/apps/docsite/src/app/(site)/themes/page.tsx
@@ -16,12 +16,29 @@
  */
 
 import type {Metadata} from 'next';
+import * as stylex from '@stylexjs/stylex';
 import {Suspense} from 'react';
 import {notFound} from 'next/navigation';
 import {Section} from '@astryxdesign/core/Section';
+import {Skeleton} from '@astryxdesign/core/Skeleton';
+import {Carousel} from '@astryxdesign/core/Carousel';
 import {packages} from '../../../generated/packageRegistry';
 import {themeObjects} from '../../../generated/themeRegistry';
 import {ThemePackagePage} from '../../../components/ThemePackagePage';
+import {ThemeHeading} from '../../../components/ThemeHeading';
+import {
+  ThemeExplorerActions,
+  ThemeExplorerLayout,
+  ThemeExplorerMobileCarousel,
+  ThemeExplorerMobileCarouselItem,
+  ThemeExplorerMobileContext,
+  ThemeExplorerPreview,
+  ThemeExplorerPreviewSkeleton,
+  ThemeExplorerRightColumn,
+  ThemeExplorerSidebar,
+  ThemeExplorerSidebarContent,
+  ThemeExplorerSidebarSurface,
+} from '../../../components/ThemeExplorerLayout';
 import {pageMetadata} from '../../../lib/pageMetadata';
 
 // Static canonical metadata for /themes. The page also accepts a `?theme=`
@@ -40,6 +57,12 @@ export const metadata: Metadata = pageMetadata({
 // users browse into the more expressive themes (Y2K, Butter, etc.).
 const DEFAULT_THEME_PACKAGE = '@astryxdesign/theme-neutral';
 
+const styles = stylex.create({
+  loadingMobileCarousel: {
+    overflow: 'hidden',
+  },
+});
+
 function slugToPackageName(slug: string): string {
   return `@astryxdesign/theme-${slug}`;
 }
@@ -51,7 +74,7 @@ export default function ThemesPage({
 }) {
   return (
     <Section maxWidth="lg" padding={6}>
-      <Suspense fallback={null}>
+      <Suspense fallback={<ThemeExplorerFallback />}>
         <SeededThemeExplorer searchParams={searchParams} />
       </Suspense>
     </Section>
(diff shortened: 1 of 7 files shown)


rubyycheung (approved):
Looks good from design. Rendering the article in the initial page shell removes the disruptive content jump, and the responsive footer remains visually consistent. The brief default-state handoff on deep links is an acceptable trade-off compared with the previous blank-page shift.

cixzhang (changes requested):
Thanks — the expanded fallback closes the tall theme-page jump, but two handoff problems remain.

On mobile, the title renders 35/44px → 29/36px → 35/44px across fallback, stream, and hydration. Readers see the heading shrink and grow while the content below reflows because the resolved server markup uses `useMediaQuery`’s false server snapshot (`ThemePackagePage.tsx:916`).

A keyboard reader can focus the fallback docs link, but when the resolved boundary replaces it, focus drops to `BODY` on both bare and Butter routes. They must navigate from the page start again (`themes/page.tsx:218`).

[Full review and frames](https://github.com/cixzhang/astryx/wiki/Review-5548)

[Reviewed by Robohands]

apps/docsite/src/app/(site)/themes/page.tsx
```diff
@@ -100,3 +123,71 @@ async function SeededThemeExplorer({
...
+      </ThemeExplorerSidebar>
+
+      <ThemeExplorerRightColumn>
+        <ThemeExplorerMobileContext>
+          <ThemeHeading align="center" isMobile isLoading />
+          <ThemeActionsFallback />
+        </ThemeExplorerMobileContext>
+        <ThemeExplorerMobileCarousel>
+          {mobileCarouselStyle => (
+            <Carousel
```
cixzhang, line 173: Carousel keeps this temporary skeleton strip tabbable; replacement drops focus to `BODY`. Could the fallback use inert shared layout?

cixzhang (changes requested):
Thanks — cixzhang’s prior review asked for stable heading geometry and preserved focus. The heading now holds, but the fallback Carousel remains a tab stop; replacing it still drops focus to the page body. Could we keep every temporary fallback descendant out of the tab order?

[Full review](https://github.com/cixzhang/astryx/wiki/Review-5548)

[Reviewed by Robohands]

PR #5549, opened 2026-08-26
fix(Drawer): stop the panel repainting into the page after close()

freddymeta:
## The bug

Dismiss a `Drawer` and the panel can snap **back into the page** for the tail of the close, over the content the user just dismissed it from.

Two things combine:

1. `dialog.close()` ran on a **hardcoded 250 ms** timer with no relation to the panel's own transition. `--duration-medium` is a theme token — 410 ms in the shipped defaults, 250 ms in `y2k`, 300 ms in the theme Storybook renders with — so the timer is either early (slide cut short) or racing the boundary.
2. The panel kept rendering after `close()`, held by the discrete `display` transition.

Once `close()` releases the top layer, the panel is an ordinary `position: fixed` box again — and **any ancestor that establishes a containing block for fixed positioning** (`transform`, `filter`, `container-type`, `contain`, `backdrop-filter`, `will-change`) becomes its origin. So the panel that had slid off the viewport edge reappears inside that ancestor and paints, shadow and all, until the hold ends.

That is not exotic: Storybook's own docs page has such an ancestor, and so does any card, virtualised list, or view-transition wrapper.

## Repro

`Lab/Drawer` → `Showcase`, with one line added to the story root:

```js
root.style.transform = 'translateZ(0)';
```

Open the drawer, close it, and watch the frames.

**Before** — `origin/main` @ 58f9542:

| t (ms) | `dialog.open` | in top layer | `display` | panel x |
|---|---|---|---|---|
| 0 | true | true | flex | 500 |
| 133 | true | true | flex | 872 |
| 200 | true | true | flex | 896 |
| 233 | true | true | flex | 899 |
| **250** | **false** | **false** | **flex** | **608** |

The panel is 291 px back inside the viewport, after the user dismissed it.

**After** — this branch:

| t (ms) | `dialog.open` | in top layer | `display` | panel x |
|---|---|---|---|---|
| 0 | true | true | flex | 500 |
| 133 | true | true | flex | 872 |
| 200 | true | true | flex | 896 |
| 233 | true | true | flex | 899 |
| 250 | true | true | flex | 900 |
| 318 | false | false | **none** | — |

The slide now finishes (x reaches 900) and the panel is already hidden by the time it leaves the top layer. The probe that catches the offending frame reports `never painted outside the top layer`.

## Screenshots

The dashed box is the ancestor carrying the `transform`. Drawer open:

drawer open

**Before** — the frame after `close()`, with the drawer already dismissed. The panel is back inside the page:

before: panel snaps back into the page

**After** — the same moment on this branch. There is no such frame; this is the first frame after `close()`:

after: nothing paints

*(Assets live on the `assets/pr-5549` branch of this repo — asset-only, deletable with the PR. I have no fork.)*

## The fix

- **The transition is authoritative.** The exit ends on the panel's own `transitionend` for `transform`, backstopped by the duration *read off the element* — so a theme that rewrites `--duration-medium` is respected rather than assumed. Same contract `BottomSheetPanel.waitForTransition` already uses, and it replaces the reduced-motion special case for free (the CSS already collapses to `0.01s`).
- **The hide lands in the same task as `close()`**, via `flushSync`. A plain `setState` is not enough — React's default scheduling can commit after the next paint, and that one frame *is* the bug. I measured this: with a plain `setState` the offending frame was still there at x=600.
- **React owns `display`** for both legs, so the discrete `display` transition (and `transition-behavior: allow-discrete`) is gone. `@starting-style` still drives the entry; it applies whenever an element goes from not-rendered to rendered and does not need `display` in the transition.
- **The lifecycle is named and isolated.** `useDrawerDialogPresence` now owns native open/close, `waitForDrawerExit`, computed-duration backstop, synchronous hide, focus restoration, and unmount cleanup. `waitForDrawerExit` follows `BottomSheetPanel.waitForTransition`: `transitionend`/`transitioncancel` are authoritative and the timeout only prevents a lost event from stranding an open dialog.

## Start-side exit latch

This PR also fixes the `side` prop changing during the exit. The documented state-derived shape is:

```tsx
<Drawer
  side={side ?? 'end'}
  isOpen={side != null}
  onOpenChange={isOpen => !isOpen && setSide(null)}
/>
```

Closing clears `side`, so the live prop reverted to `end` while the panel was still mounted for its exit. A drawer opened from `start` therefore teleported to the end edge and slid out right; its `data-side` theme state changed mid-exit too. The side is now latched while open and held until the exit completes — the same retention contract the docs already require for the drawer's children. The regression test asserts `data-side="start"` after the controlling state has cleared.

## Notes

- No API change; no `.doc.mjs` change. `@astryxdesign/lab` is private, so `check:changesets` rejects a changeset for it — none added.
- Drive-by in the same test file: the container-padding test passed `onClose`, which is not a prop, so `pnpm -F @astryxdesign/lab typecheck` was already failing on `main`. One word, and the file typechecks.

## Test plan

- `pnpm exec vitest run packages/lab/src/Drawer` — 37 pass (34 before; closes on `transitionend`, ignores a `transitionend` for another property, and holds the opening side through exit; the existing delay test also asserts the drawer is *still* open at 250 ms).
- `pnpm -F @astryxdesign/lab typecheck` — clean (red on `main`, see above).
- `pnpm lint:strict` — 0 errors.
- `pnpm build`, `pnpm -F @astryxdesign/storybook build` — green.
- The existing CI Chromium guard now covers Drawer under a transformed ancestor. On this branch: `close() ran at display:flex; hidden before the next paint`. Negative control against `origin/main`: `painted after close() outside the top layer — display:flex, open:false, x:624, width:374`. The same guard keeps its existing MobileNav invariant green.

### DIFF
--- a/.github/scripts/modal-close-visibility.js
+++ b/.github/scripts/modal-close-visibility.js
@@ -3,21 +3,23 @@
 
 
 /**
- * @description Asserts a modal <dialog> is still rendered when close() runs
+ * @description Asserts native-dialog close/render ordering in a real browser
  * @input --storybook-dir <path> [--port <n>]
- * @output One line per target; exit 1 if any close() ran at display: none
+ * @output One line per target; exit 1 if a close/render invariant fails
  *
- * A modal <dialog> that is `display: none` while still `:modal` blocks every
- * pointer event on the document, and browsers are not required to release that
- * block when close() finally runs — Safari 26.1 did not, leaving a page that
- * looked normal and answered nothing (#4290). The drawer must therefore still
- * be rendered at the moment it is told to close.
+ * Two opposite ordering bugs live at this boundary:
  *
- * That ordering only exists in a browser: it is produced by a `display`
- * transition with `allow-discrete`, which jsdom neither runs nor exposes, so
- * the unit suite passes whether or not the invariant holds. Chromium is enough
- * to observe it — the ordering is the bug, and the Safari version it was
- * reported against is not needed to see it go wrong.
+ * 1. A modal <dialog> that is `display: none` while still `:modal` blocks every
+ *    pointer event on the document, and Safari 26.1 did not release that block
+ *    when close() finally ran (#4290). It must still be rendered when close()
+ *    starts.
+ * 2. A fixed panel that stays rendered after close() leaves the top layer can
+ *    paint back inside a transformed ancestor for one frame (#5549). It must be
+ *    hidden before the next paint after close().
+ *
+ * Neither ordering exists in jsdom: it runs no CSS transition and has no top
+ * layer. Chromium is enough to observe both invariants; the browser-specific
+ * failure that first exposed each one is not needed to see the ordering fail.
  */
 
 const { chromium } = require('playwright');
@@ -39,6 +41,15 @@ const TARGETS = [
     component: 'MobileNav',
     story: 'core-mobilenav--default',
     openButton: 'Open Navigation',
+  },
+  {
+    component: 'Drawer',
+    story: 'lab-drawer--showcase',
+    openButton: 'Open inspector',
+    // Reproduces the real failure condition: after close() releases the top
+    // layer, this becomes the containing block for the fixed panel.
+    transformAncestor: true,
+    mustBeHiddenAfterClose: true,
   },
 ];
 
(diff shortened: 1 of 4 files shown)


cixzhang (changes requested):
Thanks for chasing this down. The matched close frames show the fix:

| Before | Head |
|---|---|
| Before: closed panel still paints | Head: closed panel paints nothing |

Two things before merge. The close now coordinates render state, a transition listener, timer, `flushSync`, and unmount cleanup, but the tests mock the dialog in jsdom. A future change can restore the same post-`close()` frame for anyone using a transformed ancestor while all 37 tests stay green. Could you move the close timing into a named hook or utility and add that Chromium regression? `BottomSheetPanel.waitForTransition` is the closest existing pattern.

Please also add the start-side latch to the PR body: builders using the documented state-derived `side` shape now get a different exit and theme-state value, but that second change is not described.

[Reviewed by Robohands]

freddymeta:
Addressed both asks in 0f5cf5093.

- Extracted the close lifecycle into `useDrawerDialogPresence`, with `waitForDrawerExit` following `BottomSheetPanel.waitForTransition`: `transitionend`/`transitioncancel` are authoritative; the timeout is only the computed-duration backstop. The hook now owns native dialog state, rendered presence, the synchronous post-`close()` hide, focus restoration, and unmount cleanup.
- Extended the existing CI Chromium `modal-close-visibility` guard with Drawer under a transformed Storybook root. It samples immediately before the next paint after `close()`:
  - head: `close() ran at display: flex; hidden before the next paint`
  - negative control against `origin/main`: `painted after close() outside the top layer — [{"display":"flex","open":false,"x":624,"width":374}]`
  - MobileNav's existing invariant stays green.
- Added the start-side latch to the PR body, including the documented state-derived prop shape, the `data-side` theme-state effect, and its regression test.

Local gates: 37 Drawer tests, lab typecheck + docs typecheck, `lint:strict`, full build, Storybook build, and the Chromium guard all pass.

cixzhang (approved):
Thanks—both asks from my earlier review are addressed: the lifecycle is isolated with a passing Chromium regression, and the start-side behavior is now documented. Good to merge.

[Reviewed by Robohands]

Issue #3675, opened 2026-07-08
[RFC] Sheet component — API review (evaluate DJ's Drawer draft: layer integration + optional mobile swipe)


cixzhang:
## Summary

Open an API review for a first-class **Sheet** component. DJ (@thedeej) already has a substantial `Drawer` implementation in progress (branch `drawer-component`); this issue is to evaluate that API before it lands and to align on a few open design questions — especially the sheet-specific behaviors that the current draft doesn't yet cover.

Related prior gap: #2575 (no Drawer/Sheet component for side-anchored slide-in panels).

## What already exists (DJ's `drawer-component` branch)

`packages/core/src/Drawer/Drawer.tsx` — a native `<dialog>`-based, edge-anchored overlay:

- `side: 'start' | 'end' | 'top' | 'bottom'` (logical, RTL-aware). `top`/`bottom` render as full-width sheets on the block axis.
- `size: number | string` — budget along the slide axis; fills the axis on smaller viewports.
- `hasScrim` — `showModal()` (top layer, focus trap, scroll lock, click-outside) vs `show()` (non-modal, page stays interactive).
- `isCollapsed` / `onCollapsedChange` — collapse non-modal side drawers to a click-to-expand rail.
- `hasCloseButton`, `label` (required a11y name), fully controlled `isOpen` / `onClose`.
- LIFO stacking — Escape closes the top-most open drawer; relies on the browser top layer's chronological stacking.

This is already close to what we want: **its own component**, four sides, stacking, modal/non-modal.

## Open questions for review

1. **Own component vs. a mode of Dialog** — the draft is already a standalone `Drawer`. Confirm that's the direction (vs. WWW, where the sheet is folded into the dialog). A bottom sheet should be *the same pattern* as a floating/modal panel, differing mainly by placement + optional mobile swipe — not a separate mental model.

2. **Layer system integration** — the draft leans on the native `<dialog>` top layer for stacking/dismissal. We likely want explicit layer-system integration for top-most dismissal ordering, coordination with other overlays (popovers, toasts, tooltips), and predictable Escape/scrim-click routing across mixed overlay types. Worth deciding whether native top-layer ordering is sufficient or we need to route through XDS's layer machinery.

[...]

PR #5550, opened 2026-08-26
feat(Drawer): independent scope, modality, and scrim paint

freddymeta:
## The gap

A `Drawer` is always the viewport's. There is no way to say *this inspector belongs to this pane* — a split view, a dashboard card, a canvas, a chart panel. `position: fixed` and `showModal()` both address the screen, so today you either take over the whole page or hand-roll the panel.

People do hand-roll it. A design system downstream of ours carries its own sheet for exactly this, with a "contained" mode whose own source comments call it an approximation of what Astryx does not have. That is the demand signal.

## The API

```tsx
const paneRef = useRef<HTMLDivElement>(null);

<div ref={paneRef} style={{position: 'relative'}}>
  <HostTable onSelect={setSelected} />
</div>

<Drawer
  isOpen={selected != null}
  onOpenChange={isOpen => !isOpen && setSelected(null)}
  label="Host details"
  containerRef={paneRef}>
  <HostDetails host={selected} />
</Drawer>
```

One prop for scope. `containerRef?: React.RefObject<HTMLElement | null>` — matching the `anchorRef` shape Popover/Tooltip/Typeahead already use. The panel is portalled into the element and pinned with `position: absolute`, at the container's height, with the 56px mobile reveal measured against the container (`100%`) rather than the viewport (`100dvw`).

Absent, nothing changes: `containerRef` is optional and every existing path is byte-identical.

## Screenshots

`Lab/Drawer` → `Bounded`. Closed:

pane closed

Open — the panel is inside the pane, the scrim dims only the pane, and the button below it is still live:

bounded drawer open

*(Assets live on the `assets/pr-5550` branch of this repo — asset-only, deletable with the PR. I have no fork.)*

## API: three independent axes

The review was right: scope, enforcement, and paint are separate decisions.

```ts
containerRef?: React.RefObject<HTMLElement | null> // WHERE
modality?: 'modal' | 'nonModal'                    // ENFORCEMENT; default 'modal'
hasScrim?: boolean                                 // PAINT; default modality === 'modal'
```

The defaults preserve the ordinary case: a modal Drawer has a scrim, and a non-modal Drawer does not. Callers can deliberately separate them for future products:

- `modality="nonModal" hasScrim` dims context without taking it out of play. Its scrim is paint-only and pointer-transparent.
- `modality="modal" hasScrim={false}` blocks context without dimming it.

`containerRef` changes only the scope. Viewport modality uses the browser top layer; bounded modality applies `inert` to current and late-inserted siblings for the full rendered lifetime, including exit. `isModal` is removed rather than retained as an alias: `lab` is private/canary and all branch-owned call sites, docs, tests, and stories now use the independent API.

The native dialog also follows a live `modality` change. Because an open `<dialog>` cannot switch between `show()` and `showModal()` in place, Drawer closes and immediately reopens it in the requested mode while preserving the original focus-return target.

## The one honest trade

**A bounded modal cannot use the top layer.** A top-layer element is always viewport-sized, so containment and the top layer are mutually exclusive — no amount of CSS makes `showModal()` respect a box. That costs exactly two things, `aria-modal` and the body scroll lock, because neither is true of a bounded panel. The blocking itself is not lost: `modality="modal"` makes the container background `inert`.

| | viewport drawer | bounded drawer |
|---|---|---|
| how it opens | `showModal()` / `show()` | always `show()` |
| `modality="modal"` enforced by | the top layer | container siblings made `inert` |
| area taken out of play | the page | the container |
| scrim | `::backdrop` | its own element, clipped to the container |
| `aria-modal` | `"true"` when modal | never |
| body scroll lock | yes, when modal | no |
| Escape | yes | yes (while focus is in the panel) |

The area the drawer belongs to is the area modality blocks. Paint remains independent: `hasScrim` controls dimming, while keyboard, pointer, and accessibility blocking follow `modality`. The rest of the page stays live for a bounded Drawer.

A `useDevWarning` fires when the container computes to `position: static`, because the panel would silently escape to the nearest positioned ancestor.

## Behaviour in a scrolling pane

An absolutely positioned child of a scroll container resolves against the *scrolled* padding box, so a naive `inset: 0` rides the content out of view. The clip box offsets by the container's scroll position and sizes to its client box, so the panel stays in the scrollport:

| | panel y | panel h |
|---|---|---|
| `scrollTop: 0` | 195 | 298 |
| `scrollTop: 180` | **195** | 298 |

Written straight to the node from a passive scroll listener rather than through React state — a re-render per scroll frame is not affordable, and a transform stays on the compositor. A container that does not scroll reports offset 0 and its padding box, so it is one code path.

## Cost when closed

Resolving the portal target is a state write, so doing it on mount cost every closed bounded Drawer on the page a second commit. It now resolves only once the panel actually paints; a `Profiler` test asserts a closed bounded Drawer commits exactly `['mount']`. The `getComputedStyle` behind the static-container warning likewise moved out of render into a dev-only effect.

Resolution also survives the host being swapped or detached — the ref object never changes identity, so a dependency array cannot see it; a `MutationObserver` can.

## Screenshots

`Lab/Drawer` → `Bounded`. Closed:

pane closed

Open — the panel is inside the pane, the scrim dims only the pane, and the button below it is still live:

bounded drawer open

*(Assets live on the `assets/pr-5550` branch of this repo — asset-only, deletable with the PR. I have no fork.)*

## Measured

`Lab/Drawer` → `Bounded`, in Chromium against a real Storybook build:

| | value |
|---|---|
| panel `position` | `absolute` |
| panel parent | the container element (portalled — it is declared as a sibling in the story) |
| `:modal` | `false` |
| `aria-modal` | absent |
| container rect | x 202, w 496, h 300 |
| panel rect | **x 437, w 260, h 298** — pinned to the container's end edge, at its height |
| scrim rect | **x 203, w 494, h 298** — the container, not the viewport |
| `body` overflow | `visible` (not scroll-locked) |
| click a button outside the pane while open | succeeds |
| focus a dimmed control inside the pane | refused — focus stays put |
| `elementFromPoint` over that control | not the control |
| Escape with focus in the panel | closes; `display` returns to `none` |

## Tests

59 Drawer tests pass after the current-main merge. New guards cover all four modality/scrim combinations in both viewport and bounded scopes, live modality changes, late-inserted pane content, and inert cleanup after exit. Coupling scrim paint back to modality makes three focused axis tests fail. Full build, lab/docs/Storybook typechecks, repository checks, and strict lint pass.

No changeset: `@astryxdesign/lab` is private and not released.


cixzhang (dismissed):
Thanks—pane-scoped Drawer is the right owner. Please keep scope and modality orthogonal: `containerRef` chooses the host only; expose explicit `modality` (`'modal' | 'nonModal'`) rather than silently changing `hasScrim` and focus behavior.

Three runtime paths still block:

- In a scrollable pane, wheel input over the reveal moved the absolute portal host: `scrollTop 0→180` moved the 292px drawer from y=44 to −136, leaving 116px visible (`Drawer.tsx:290-301`).
- Two reverse Tabs focused the dimmed opener, and Enter activated it while pointer input stayed blocked (`Drawer.tsx:864-880`).
- Replacing the target behind the same ref removed the dialog while `isOpen` stayed true; even closed, target resolution doubled Profiler commits at N=1/3/10 and added one style read per Drawer (`Drawer.tsx:598-612`).

Please preserve the host through scroll/replacement, make blocking consistent across input modes, and avoid the extra mount pass/read.

[Reviewed by Robohands]

freddymeta:
All four in `c11ca9fc9`.

## `modality`, separate from scope

You were right that `containerRef` was reaching into behaviour it had no business touching. Now:

- `hasScrim` — paints, and dismisses on click. Visual.
- `modality: 'modal' | 'nonModal'` — whether the page is actually out of play.

`modality` defaults from `hasScrim`, since that is precisely what `hasScrim` meant before the axis existed, so no existing call site moves. Both combinations that were previously unexpressible now work and are tested: a painted scrim that does not block, and a modal with no scrim.

`containerRef` + `modality="modal"` is a **dev warning**, not a silent downgrade — the top layer is viewport-sized, so it genuinely cannot be honoured, and saying so is better than quietly doing something else.

## The three runtime paths

**Scroll.** Root cause is that an abspos child of a scroll container resolves against the *scrolled* padding box, so `inset: 0` rides the content — exactly your `scrollTop 0→180` carrying the panel to y=−136. The clip box now offsets by the scroll position and sizes to the client box, so it tracks the scrollport. Measured in Chromium on a new `Bounded in a scrolling pane` story:

```
scrollTop 0   → panel y=195 h=298
scrollTop 180 → panel y=195 h=298   movedBy: 0
```

Written straight to the node from a passive scroll listener rather than through state — a re-render per scroll frame is not affordable, and a transform stays on the compositor. For a non-scrolling container the offsets are 0 and the client box is the padding box, so it is one code path, not a branch.

**Keyboard vs pointer.** The scrim blocked the pointer and not the keyboard, which is the inconsistency. The container's other children are now `inert` while a scrim is up — pointer, tab order and the accessibility tree in one attribute. The panel lives inside the container and is skipped, and anything already `inert` is left exactly as found. Measured:

```
focus() on a dimmed pane button → focus stays put (activeElement: DIV)
elementFromPoint over it        → not the button
consistent: true
```

**Target resolution.** Three separate problems, all fixed:
- resolved on mount for every bounded Drawer whether or not it would ever open → now resolves only once the panel actually paints. A `Profiler` test asserts a closed bounded Drawer commits exactly `['mount']`.
- the deps could not see the element behind the ref being swapped (the ref object never changes identity) → resolution now compares before writing, and a `MutationObserver` catches a host that is replaced or detached, so `isOpen` can no longer be true with no dialog.
- `getComputedStyle` ran during render, per Drawer, to feed a warning production never prints → moved into a dev-only effect.

## On the tests

Eight new ones, and I ran each against the **unfixed** component to confirm it fails there — two of my first drafts passed on the old build and were not testing anything:

- the swap harness rendered a `<div>` in the same slot both times, so React reconciled it as one element and nothing was ever swapped. It needed a `key`.
- the commit-count test counted parent renders, which say nothing about the Drawer. It now uses `Profiler`, the way you measured it.

54 tests pass; the 8 fail on the pre-fix build.

## Still true, and worth restating

A bounded drawer cannot be modal. That is a browser constraint, not a choice — and now that `modality` is explicit, it is a constraint the API states out loud instead of one it silently applies.

freddymeta:
Correction on my last comment: **I've dropped `modality`** (`d5ecb7e18`). I added it because you named it, then ran it through [API Arbitration](https://github.com/facebook/astryx/wiki/API-Arbitration) properly — and that process disqualifies itself here:

> Don't use this for: **Decisions already covered by [[API Conventions]]** (just follow the convention)

Three conventions cover it.

**1. [Escape Hatches](https://github.com/facebook/astryx/wiki/API-Conventions) — "Don't add flexibility without a demonstrated use case."** Its own worked example is this exact shape:

> Prefer enforced good practices over toggles. Example: focus trapping in a popover should be the default behavior, not a `hasFocusTrap` toggle — unless there's a real use case for disabling it.

Modality *is* focus trap + scroll lock + top layer. The prop's only new capability was two combinations neither of us has a use case for: a scrim that doesn't block (looks blocked, isn't) and a modal with no scrim (blocks with no sign it's blocked). Both are things the docs would warn against.

**2. `BottomSheet` already does it the other way.** `BottomSheet.tsx:275` — `hasScrim` drives `showModal()`, `useScrollLock` and focus, the very conflation the prop was meant to undo. Adding the axis to Drawer alone splits the family.

**3. The name is wrong even if the axis were right.** `Popover` already spells it `isModal?: boolean` (`Popover.tsx:187`), and the boolean convention is `is`/`has`. A two-value string union expressing a boolean isn't the house style.

## What I think you actually caught

Not a conflation — a **scope** change wearing the costume of a semantic one.

`hasScrim` means *"the area behind is out of play."* `containerRef` changes what that area **is**, not what the sentence means. Viewport drawer → the page. Bounded drawer → the container.

The reason it read as a silent behaviour change is that the bounded scrim **wasn't actually blocking** — your own second finding. It dimmed, and the keyboard walked straight through it. So `hasScrim` really was promising something bounded mode didn't deliver, and the honest fix is to make it deliver: the container is now `inert`, so within its scope the guarantee is identical. Same meaning, narrower scope, nothing silent.

The one thing that genuinely doesn't survive containment is the *top layer* — `aria-modal` and body scroll lock — and that's a browser constraint, documented on both props and in a `guidance: false`.

`hasScrim` is the single axis again, as it is on BottomSheet. The four `modality` tests went with it; the eight covering the runtime holds stay. 50 pass.

Happy to be argued back into it if you have a caller who needs one of those two combinations — that's exactly the "demonstrated use case" the convention asks for, and I couldn't produce one.

freddymeta:
Reversing my last comment — **`modality` is back in** (`4be288acf`), and done properly this time.

I talked myself out of it on a durability argument ("new permanent API to express two anti-patterns"). That argument doesn't apply here: `lab` is canary-only, `private: true`, and [Component Lifecycle](https://github.com/facebook/astryx/wiki/Component-Lifecycle) says in as many words — *"API might change. No stability promise."* Iterating the API **is** what lab is for, and Escape Hatches is about not baking speculative flexibility into a stable surface. I applied a `core` bar to a `lab` component.

## What changed from the first attempt

The first version defaulted `modality` from `hasScrim` "for back-compat" — which left the two props half-tangled, i.e. exactly what you asked me to undo. With no back-compat to protect, there's no reason to hedge:

```ts
modality?: 'modal' | 'nonModal'  // default 'modal'            — what is ENFORCED
hasScrim?: boolean               // default modality === 'modal' — what is SEEN
```

The scrim follows modality, so ordinary cases still set one prop, and both separations are expressible: a non-modal inspector that dims its context, and a modal over content that should stay legible.

## And I had bounded mode wrong

Last version warned on `containerRef` + `modality="modal"`. That was wrong, and my own `inert` fix is why.

*"The area behind is out of play"* is true whether that area is the page or the container. Only the **mechanism** differs, because the top layer is viewport-sized:

| | viewport | bounded |
|---|---|---|
| `modality="modal"` enforced by | `showModal()` | container goes `inert` |
| `aria-modal` | `"true"` | absent |
| body scroll lock | yes | no |
| area behind actually blocked | yes | yes |

So there's nothing to refuse and nothing to warn about. The warning is gone; the two things a bounded modal genuinely isn't (`aria-modal`, scroll lock) are documented on the prop instead. That also moved `inert` to key off `modality` rather than off the scrim — enforcement belongs with the enforcement axis, and keying it to paint was a leftover from when one prop did both jobs.

## Migration

20 `hasScrim={false}` call sites across tests, stories and doc examples now say `modality="nonModal"`, because that is what they meant. 56 tests pass.

Sorry for the round trip — you asked for this in the first place and I second-guessed it against the wrong rulebook.

freddymeta:
One more turn on this, and it lands somewhere simpler than either previous version: **one prop, `isModal`** (`4fecc8f5c`). `modality` and `hasScrim` are both gone.

Freddy's question was the right one — *are these two props actually pointing at the same thing?* Near enough, yes. They came apart only in combinations neither of us could name a use for, and both are things we'd warn against:

- dimmed but still clickable → looks dead, isn't
- blocked but not dimmed → clicks do nothing, no clue why

So it's one decision, and it should be one prop.

The rename is the part that was actually worth doing. `hasScrim` names the *decoration* — and it's a theatre word; reading this API shouldn't require knowing it — while the prop decides whether the area behind is **blocked**. `isModal` says that, matches `Popover.isModal`, and fits the `is`/`has` boolean convention.

Honest cost: it diverges from `BottomSheet.hasScrim`. I still think it's right — that name describes what's painted rather than what's enforced — but if you'd rather Drawer match its closer sibling than Popover, say so and I'll flip it.

`isModal` means the same thing bounded, just by a different mechanism (top layer vs dimming the container and making it `inert`), so there's no warning and nothing to downgrade. 20 call sites migrated, 54 tests.

cixzhang (dismissed):
Thanks—the scroll and closed-cost fixes now pass, but our prior API ask is still open. The ratified Drawer ruling keeps `containerRef` as scope, `modality` as enforcement, and `hasScrim` as paint; `isModal` recombines the last two, so please restore those independent axes (`Drawer.tsx:478-519`).

The inert pass also snapshots only current children. In Chromium, a control inserted into the pane after open had no `inert`, accepted focus, and stayed keyboard-active while the scrim blocked its pointer path (`Drawer.tsx:719-746`). Please keep later pane content unavailable for the full modal lifetime.

This also needs a rebase: [#5549](https://github.com/facebook/astryx/pull/5549) landed `useDrawerDialogPresence`, and current head conflicts in `Drawer.tsx` and `Drawer.test.tsx`.

[Reviewed by Robohands]

freddymeta:
Rebased onto main as one commit (`c892388ad`), late-inserted pane content now inerts, and I want to push back on the API ask rather than flip it a fourth time.

## Rebase

`useDrawerDialogPresence` is in, and #5549's two commits are dropped. The branch's other four walked the modality API out and back again, so they are squashed into the shape they landed on — eight commits down to one.

One change to the hook: an optional `mountHost`. A ref is stable, so nothing in the hook can observe the dialog **element** arriving, and a bounded Drawer renders no dialog until its container resolves a commit later. Without the host as a dependency an already-open bounded Drawer never opens — which is how the rebase first failed `never uses the top layer, even with a scrim`.

## Late pane content

You were right, and it is a better catch than the original: a pane is live content — a row streams in, a menu opens, a lazy panel resolves — and stamping `inert` once at open left every later arrival focusable behind a scrim that already blocked the pointer. Same keyboard/pointer split, one layer down.

A `childList` MutationObserver now re-runs the pass for the whole open lifetime. `childList` only, deliberately: an attribute or subtree filter would fire on our own writes and on every keystroke in the pane, and neither can add a child to the container. The drawer still hands back only what it inerted, so a container marked inert for someone else's reason is returned untouched — there is a test for that too. Both new tests verified failing against a one-shot pass.

## On the ratified ruling

**I could not find it, and I looked properly:** the wiki (all 60+ pages), open and closed issues, closed PRs, and the repo itself. The nearest thing is [#3675](https://github.com/facebook/astryx/issues/3675), the Sheet RFC, which describes `hasScrim` doing *both* jobs — `showModal()` vs `show()` — the single axis, not three.

If it exists somewhere I cannot see, link it and I will follow it. But I do not think "restore those independent axes" is accurate, because they were never independent:

**`main` today has one prop.** `Drawer.tsx:351` is `hasScrim?: boolean`, and there is no `modality`. There is nothing to restore.

**And main already calls this concept `isModal`.** From your own #5549:

```tsx
useDrawerDialogPresence({
  dialogRef,
  isOpen,
  isModal: hasScrim,   // Drawer.tsx:441 on main
  setIsRendered,
});
```

The internal vocabulary is already `isModal`, fed from a public prop named after the paint. That is the mismatch this PR fixes — it makes the public name the one the code already uses, alongside `Popover.isModal` (`Popover.tsx:187`) and the `is`/`has` boolean convention.

Splitting them adds a prop whose only new capability is two combinations we would document against: a dimmed area that still takes clicks (looks broken), and a blocked area with no dimming (no clue why clicks do nothing). [API Conventions](https://github.com/facebook/astryx/wiki/API-Conventions) — Escape Hatches — is explicit that this is not worth a toggle, and its example is exactly this shape:

> Prefer enforced good practices over toggles. Example: focus trapping in a popover should be the default behavior, not a `hasFocusTrap` toggle — unless there's a real use case for disabling it.

So: happy to be wrong with a link, or with a caller that needs one of those two combinations. Absent either, one prop, named the way main already names it.

freddymeta:
Implemented the ratified API as requested in `4bdab3fc3` + `4bd6e6610` and merged current main.

- `containerRef` = scope only
- `modality: 'modal' | 'nonModal'` = enforcement
- `hasScrim` = paint, defaulting to modality
- removed public `isModal` completely (lab/private), migrated every branch-owned call site, doc and story
- wired both independent combinations in viewport **and** bounded modes: non-modal + scrim is pointer-transparent; modal + no scrim still blocks
- bounded inert now includes late children and the whole exit lifetime
- live modality changes close/reopen the native dialog so `show()`/`showModal()` cannot drift from the prop

Chromium 2×2×2 matrix: scrim pixels `153,160,169` vs clear `255,255,255`; modal underlying controls got 0 clicks, non-modal got 1; bounded modal content was inert; non-modal scrims were `pointer-events:none`.

59 Drawer tests pass. Coupling paint back to modality makes the three focused axis tests fail. Full build, lab/docs/Storybook typechecks, repo checks, and strict lint pass.

@cixzhang re-requesting review — this follows the independent-axis suggestion rather than continuing the one-prop pushback.

imdreamrunner:
it looks a little strange to me to separate `modality` and `hasScrim`. i would defer to you and @cixzhang  to decide on this API and if we all agree on separating them as 2 prop, i will update bottom sheet to follow the same style.

separately @rubyycheung and I are planning to promote this component to core after this. no rush

cixzhang (dismissed):
The independent scope/modality/scrim API is acceptable as reversible lab work, but two bounded modal Drawers sharing one host inert each other's clip wrappers. The front Drawer can therefore become inoperable by keyboard and pointer.

Please fix that bounded-modal stacking boundary; the other API experimentation does not block.

[Reviewed by Robohands]

freddymeta:

Fixed in `c905415d0`; current main merged (branch was 56 behind — only conflict was the generated `motionAudit.ts`, resolved by regenerating rather than hand-editing).

**The bounded-modal stacking boundary.** You were right, and the cause was the escape hatch in the pass: each bounded modal inerted every child of its host *except its own clip wrapper*. Two of them in one container therefore inerted each other, and the front panel — the one the user is looking at — went dead to pointer and keyboard.

Enforcement is now single-owner per container:
- **Only the front-most blocking drawer in a host applies `inert`.** The ones behind it are inerted along with the rest of the container, which is what "behind a modal" should mean.
- **The pass spares any drawer opened after the enforcer.** A modal puts what is BEHIND it out of play; a drawer in front of it is not that. Without this, a non-modal drawer opened on top of a bounded modal hits the same dead-panel bug one case over.

**Why a second registry, rather than duplication.** The existing LIFO stack tracks *open* drawers, so Escape closes the innermost, and a drawer leaves it the moment it starts closing. Bounded enforcement has to outlive that — `inert` is held for the whole exit transition, or the container goes live under a panel still on screen (your earlier review, and the test that pins it). The new registry tracks *rendered* bounded drawers: same ordering, different lifetime.

Four tests, each verified to fail on the previous head: front interactive with the one behind blocked (the reported bug); handback when the front closes; non-modal on top left alone; container never released when the drawer *behind* closes first — watched with a `MutationObserver` on the `inert` attribute, since a release repaired on the next render leaves the end state correct while the container is live for a frame.

Negative control on each half: remove the front-most gate and its stack subscription → three of four fail; remove the spared-clips skip → the non-modal one fails.

Gates: `packages/lab` + `packages/core` **314 files / 8755 tests pass** (Drawer 64/64, up from 59; the pre-existing exit-transition test unchanged); core and lab builds clean.

cixzhang (dismissed):
Thanks — the pane-scoped inspector is the right capability, and the scroll, late-content and closed-cost fixes hold at this head.

We asked for the bounded-modal stacking boundary last round, and the case we reported is fixed. It returns one step over: change one bounded drawer's `modality` while another is open in the same pane, and the drawer behind takes over the `inert` pass and applies it to the front panel. In Chromium that panel keeps painting on top and stops responding — whoever is working in that inspector loses every control in it, to keyboard and pointer, with nothing explaining why (`Drawer.tsx:821-830`).

A bounded pane now carries two orderings of its drawers — which paints in front, and which applies `inert` — and here they disagree. Could one ordering decide both? A test flipping modality on an open pair would pin it; the 64 we have pass without seeing this.

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks — the main merge does not change Drawer, so the bounded stacking block from our prior review is still open: someone using two bounded inspectors can lose every control in the visible front Drawer.

If a buried Drawer changes from `modality="nonModal"` to `"modal"` while another modal Drawer is open above it, the buried entry becomes the inert owner. At this exact head, Chromium refused focus and pointer hit-testing on the visible front panel (`Drawer.tsx:821-900`). Please keep rendered order stable across live modality changes and cover this path in the pair matrix.

[Reviewed by Robohands]

freddymeta:
Closing because we are no longer using the bounded Drawer variant.

PR #5551, opened 2026-08-26
feat(Card): make the variant axis theme-extensible via CardVariantMap

cixzhang:
## What

`Card`'s `variant` was a hand-written union backed by a closed style record, so a theme could not add a card variant — an unknown value neither type-checked nor rendered. This converges Card on the extensible-map pattern the other fifteen axes already ship (`Button`, `Badge`, `Section`, …).

- `packages/core/src/Card/index.ts` — declares and exports `CardVariantMap`, the augmentation point at the public subpath `@astryxdesign/core/Card`.
- `packages/core/src/Card/Card.tsx` — `CardVariant` is now `keyof CardVariantMap`, and `variantStyles` is typed `Partial` over it so the lookup stays legal for a value the record deliberately has no entry for.
- `packages/core/src/SelectableCard/SelectableCard.tsx` — the selected-ring switch gets an unknown-variant fall-through, so a theme-added variant still looks selected. Its ring colour is `--selectable-card-ring-color`, a public var declared on the `astryx-selectable-card` target and defaulting to `var(--color-accent)`.

## The ring for a variant the component cannot know

No token the component could pick is guaranteed to contrast with a fill the component has never seen. Measured against a `#1f1f1f` fill: the accent ring reads 1.09:1 in `neutral`, and an outset accent ring is no better — the selected card just looks 2px bigger. So the theme that supplied the fill supplies the ring, in the same rule:

```ts
card: {
  'variant:brand': {
    backgroundColor: '#1f1f1f',
    '--selectable-card-ring-color': '#f5f5f5',
  },
}
```

That is 15.12:1 against the same fill. The var is declared on the target with the accent as its value, so it is reachable from a theme (`theme-var-reachability` proves the theme rule outranks the component's own declaration) and every built-in variant is unchanged — the nineteen built-in selection treatments are byte-identical to `main` across three stories, and all thirteen built-in Card variants are identical across eight computed properties.

The ring stays a box-shadow rather than becoming a real border: only `default` draws a border, and `withBorder` subtracts its width from all four paddings to keep the total inset equal to the padding token, so a border appearing on select would either shift content by 1px or need that calc replicated per variant.

**Follow-up, deliberately not here:** the thirteen built-in variants should read the same var with their own border token as the fallback. That is the right end state and it touches thirteen style objects, which is a different PR from the map.

## Not a breaking change

`keyof CardVariantMap` resolves to exactly the thirteen values `CardVariant` had — same values, same autocomplete, no consumer edit. `ClickableCard` and `SelectableCard` forward the type and need no signature change. The changeset is `patch`.

`Card.test.tsx` pins the set: `Record<CardVariant, true>` fails to compile if a key is added to or dropped from the map, and the count is asserted at runtime.

## Unknown values fall through, deliberately

A variant a theme adds is not a key in `variantStyles`, so the lookup is `undefined`, StyleX drops it, and the card renders base styles while the theme's own `card['variant:<name>']` rule paints it. `themeProps('card', {variant, elevation})` already reflected `variant`, so the class and `data-variant` selectors a theme rule needs are there for any value.

The axis carries a background, not a border. Card subtracts a border's width from its padding for `default` alone, so a theme rule that paints a border sits a border-width proud of the cards beside it. That is not new here — a theme override that paints a border on a built-in variant does the same thing today, measured identically on `main` and on this branch — and widening the value set does not change it, so it is left alone rather than folded into this PR.

## Tests

- `Card.test.tsx` — a theme-added variant renders the selector a theme rule needs, and carries none of the built-in variants' style classes while keeping every class the built-ins share.
- `build-theme.variants.test.mjs` — `card: {'variant:customCard': …}` joins the suite that builds a theme, imports the generated `.d.ts`, and type-checks `<Card variant="customCard">` through the public subpath under `moduleResolution: "bundler"`. Mutation-checked: renaming the interface away from `CardVariantMap` makes it fail.
- `extensibleAxes.test.ts` picks the new axis up automatically and passes: `card.variant` reaches the DOM through `themeProps` and is documented in `visualProps`.

## Test plan

New Storybook story **Core/Card → Theme Added Variant** registers `brand` through `defineTheme` and augments `CardVariantMap` in the story file, which is also an in-repo proof that declaration merging through the public subpath widens the prop — the story would not compile otherwise.

Driven in real Chromium against the running Storybook: the `brand` card paints the theme's accent fill, carries none of the built-in variants' style classes, and measures the same box as the `default` card beside it. The story also renders a selected and an unselected `SelectableCard` on the added variant, so the selection treatment for a value the component has never seen is visible rather than argued.

Existing consumers see nothing move: all thirteen built-in variants, across background, border, radius, shadow, padding and rendered box, are identical between `main` and this branch — 0 rows differ.

Green locally: `pnpm build`, `pnpm lint:strict` (0 errors), `pnpm check:repo`, and the Card / SelectableCard / ClickableCard / theme suites.

### DIFF
--- a/.changeset/card-variant-map.md
+++ b/.changeset/card-variant-map.md
@@ -1,0 +1,11 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] Card: make the `variant` axis theme-extensible through a `CardVariantMap` interface
+
+`CardVariant` was a hand-written union backed by a closed style record, so a theme could not add a card variant: an unknown value neither type-checked nor rendered. It is now `keyof CardVariantMap`, an interface exported from `@astryxdesign/core/Card` that a theme build augments — the same shape Button, Badge, Section and the other extensible axes already ship.
+
+`keyof CardVariantMap` resolves to exactly the thirteen values `CardVariant` had, so no existing call site changes. A variant a theme adds falls through to base styles and the theme's own `card['variant:<name>']` rule paints it; on SelectableCard it takes the accent selection ring the neutral variants use.
+
+@cixzhang
--- a/.changeset/card-variant-map.md
+++ b/.changeset/card-variant-map.md
@@ -8,6 +8,6 @@
 
 `keyof CardVariantMap` resolves to exactly the thirteen values `CardVariant` had, so no existing call site changes. A variant a theme adds falls through to base styles and the theme's own `card['variant:<name>']` rule paints it.
 
-On SelectableCard that variant's selection ring is drawn in `currentColor`, because the component cannot know what the theme painted and no token is guaranteed to contrast with it. So a theme rule that adds a variant should set `color` alongside `backgroundColor`, or style `.astryx-selectable-card` — which reflects both `variant` and `selected` — to own the ring itself.
+On SelectableCard that variant's selection ring is drawn in `--selectable-card-ring-color`, defaulting to the accent. No token the component could pick is guaranteed to contrast with a fill the component cannot know, so a theme rule that adds a variant sets the ring colour in the same rule as its `backgroundColor`.
 
 @cixzhang
--- a/.changeset/card-variant-map.md
+++ b/.changeset/card-variant-map.md
@@ -6,6 +6,8 @@
 
 `CardVariant` was a hand-written union backed by a closed style record, so a theme could not add a card variant: an unknown value neither type-checked nor rendered. It is now `keyof CardVariantMap`, an interface exported from `@astryxdesign/core/Card` that a theme build augments — the same shape Button, Badge, Section and the other extensible axes already ship.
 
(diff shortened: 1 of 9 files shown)


cixzhang:
Thanks — the map half is clean, and the thirteen built-in rings measure byte-identical to main.

The default on `--selectable-card-ring-color` is `--color-accent`, 1.09:1 against a `#1f1f1f` fill. So a theme that adds a variant the way the theming page's own example does — `backgroundColor` and `color`, nothing else — ships a selected card that looks unselected, with nothing warning:

selection ring across five theme rules; row 2 omits the var

`currentColor` gets there for free; the theme already had to say what ink survives its fill.

```ts
// SelectableCard.tsx:61
'--selectable-card-ring-color': 'currentColor',
```

Same 15.12:1 as pairing the var by hand, and the story can drop the line repeating `color` four above it. A theme with no `color` still gets 1.09:1 — but so does its label, so it fails visibly rather than only when selected.

Does `currentColor` miss a case you had in mind?

<sub>[Reviewed by Robohands]</sub>

PR #5553, opened 2026-08-26
fix(core): make text-only Bottom Sheets keyboard-scrollable

cixzhang:
## Problem

A keyboard user cannot reliably reach a scroll container with no visible control in its sequential focus order. In BottomSheet, that makes overflowing read-only content inaccessible without a pointer ([#5207](https://github.com/facebook/astryx/issues/5207)).

## Solution

`useScrollableTabStop` owns a fallback `tabindex="0"` only while its element actually scrolls and no visible sequential-focus descendant already provides access. Hidden controls and negative-tabindex descendants do not count because Tab cannot use them. The hook observes content, focusability, and direct `overflow` mode changes, and coalesces each resize delivery to one measurement.

BottomSheet is the first in-tree adopter: its existing scrolling body composes the hook with the gesture ref. This replaces the component-local approach explored in [#5513](https://github.com/facebook/astryx/pull/5513), whose author is credited in the changeset.

## Impact

Text-only Bottom Sheets get an explicit keyboard scroll stop. Sheets with a visible button, link, input, or other sequential-focus control keep their existing tab order with no extra container stop.

## API

```tsx
const scrollRef = useScrollableTabStop();
<div ref={scrollRef}>{children}</div>
```

- Adds `useScrollableTabStop(options?: {enabled?: boolean})` from `@astryxdesign/core/hooks`.
- The hook is proven by BottomSheet rather than shipping as unused groundwork.

## Theme targets

None.

## Ossification

The hook serves a landed class of scroll containers, and BottomSheet proves the public shape in a real owner. The behavior is derived from the DOM rather than a caller-maintained state prop.

## Breaking

- API: additive.
- Behavior: BottomSheet adds a fallback stop only for overflowing content without a visible sequential-focus descendant.
- Visual: no intended geometry, color, or layout change.
- Theme: no target, token, or override change.

## Performance & resources

One shared `ResizeObserver` serves the page. Each enabled hook instance adds one `MutationObserver`; direct-child resize notifications are coalesced to one measurement per animation frame. Chromium stayed constant at 2 style reads for 20, 50, and 200 children when both mutation and resize signals fired.

## Visual evidence

The BottomSheet body now uses the same themeable focus outline as every other core focusable surface: 2px solid with the shared color, inset 2px so the sheet’s clipping edge cannot hide it.

| Before — browser-default ring | After — Astryx ring, inset to survive clipping |
|---|---|
| BottomSheet body with the browser-default focus ring | BottomSheet body with the visible Astryx focus ring |

## Judgement

BottomSheet remains the honest first owner: the public issue is still open on current main, the component is already client-side, and the hook removes the duplicate component-local observer/state design.

## What I could not verify

Safari; managed-machine policy prevents Playwright WebKit from launching.

## Testing

- Focused Vitest coverage for the hook, focus-trap selector, shared resize observer, BottomSheet panel, and BottomSheet switcher
- Core and Storybook typechecks
- Strict lint and repository checks
- Real Chromium: BottomSheet Tab reachability and keyboard scrolling; visible/hidden/negative-tabindex descendants; `hidden ↔ auto`; 20/50/200-child resize batches


cixzhang:
Thanks — this holds. I reverted the `cancelAnimationFrame` and watched the new test go red. Two nits: the TabList comment inline, and the test plan still says 14 tests.

[Reviewed by Robohands]

cixzhang:
Inline, `packages/core/src/TabList/TabList.tsx:489` — the comment says the shared observer keeps one callback per element. This PR makes that untrue.

cixzhang (commented):
Thanks — the static overflow detection is careful, but this needs three fixes before it becomes public.

The hook adds a generic stop even when the scroller has a visible sequential-focus control, so Tab stops on the container before the control. [#5513](https://github.com/facebook/astryx/pull/5513) already established the condition as actual overflow **and no visible sequential-focus descendant**.

Changing `overflow` alone leaves stale behavior: `hidden → auto` stays unreachable and `auto → hidden` stays tabbable until a later resize. Chromium reproduced both at this head.

One batched resize of 20/50/200 direct children also caused 20/50/200 style reads, while every pass rescanned all children. Could we cover the combined ownership condition, both overflow-mode transitions, and one coalesced measure per delivery before export?

[Reviewed by Robohands]

cixzhang:
Direction: do not land this as hook-only groundwork. Fix the current review findings, then wire the first real in-tree consumer so the API is proven by an adoption before it becomes public.

packages/core/src/hooks/focusableSelector.ts
```diff
@@ -2,24 +2,66 @@
...
+  ) {
+    return false;
+  }
+  if (typeof element.checkVisibility === 'function') {
+    return element.checkVisibility({
+      checkOpacity: false,
+      checkVisibilityCSS: true,
+    });
+  }
+  if (typeof window !== 'undefined' && window.getComputedStyle) {
```
cixzhang, line 39: This ancestor walk is new. jsdom has no `checkVisibility`, so an open `[popover]` reads `display: none` here.

packages/core/src/BottomSheet/BottomSheetPanel.tsx
```diff
@@ -490,6 +491,9 @@ export function BottomSheetPanel({
     onScrimOpacity,
   });
 
+  const scrollableTabStopRef = useScrollableTabStop();
```
cixzhang, line 497: Might need `focusOutlineProps.focusVisible` on the body too — every other ring in core comes from `focusOutline.stylex`.

packages/core/src/TabList/TabList.tsx
```diff
@@ -495,7 +495,7 @@ export function TabList({
     }
     const onResize = () => revealSelectedTabRef.current();
     observeResize(root, onResize);
-    return () => unobserveResize(root);
+    return () => unobserveResize(root, onResize);
```
cixzhang: The comment above still says the shared observer keeps one callback per element — this PR makes that untrue.

cixzhang (commented):
Thanks — all three from last round hold, and Chromium confirms a text-only sheet is now reachable and scrolls.

`test` is red on this head though — five files, all focus paths. Four are the `focusableSelector` consolidation: its new check walks ancestors when `checkVisibility` is missing, and jsdom reads an open `[popover]` as `display: none`, so Selector's clear button, SideNav's flyout and TransferList's reorder leave the focus trap. Chromium matches main. The fifth is the BottomSheet adoption — the hook observes the body's direct children, so the mobile-keyboard test can't tell which observer watches the input.

The stop's ring also isn't the system's — it computes `outline: 1px auto` where the Button beside it gets `2px solid var(--focus-outline-color)`, so a theme's `--focus-outline-*` skips it.

The PR description never mentions the focus-trap consolidation. Was it meant to ride along here, or does it want its own PR?

[Reviewed by Robohands]

packages/core/src/hooks/focusableSelector.ts
```diff
@@ -2,24 +2,57 @@
...
- * <summary> — which a naive selector misses, letting Tab escape a trap whose
- * only interactive content is (e.g.) a contenteditable composer (infra-8).
- *
- * This is the canonical focusable selector; prefer importing it here over
- * re-declaring the string so behavior stays consistent across hooks.
+ * tabbable natives plus editable and media elements the browser also puts in
+ * the tab order.
  */
 export const FOCUSABLE_SELECTOR =
   'button:not([disabled]), a[href], area[href], input:not([disabled]), select:not([disabled]), textarea:not([disabled]), [tabindex]:not([tabindex="-1"]):not([disabled]), [contenteditable]:not([contenteditable="false"]), audio[controls], video[controls], iframe, details > summary:first-child';
```
cixzhang, line 17: Withdrawn — this matches base; the negative-tabindex exclusion was already there. I misread the diff.

cixzhang (commented):
Thanks — approve from me. `outlineOffset: -2` was the missing piece, and the five red files pass. Withdrawing my selector nit: that line matches base, I misread the diff.

[Reviewed by Robohands]

cixzhang (commented):
Thanks — approve on the rebased head. The current shared observer implementation is preserved, the hook API is unchanged, and focused tests plus Chromium keyboard scrolling pass.

[Reviewed by Robohands]

cixzhang (commented):
Thanks — approve on the clean rebase. Focused tests and Chromium preserve descendant ownership, overflow transitions, resize batching, keyboard scrolling, and the inset focus ring.

[Reviewed by Robohands]

cixzhang:

PR #2941, opened 2026-06-19
feat: XDS un-prefix migration — bare names canonical, XDS* compat aliases

ejhammond:
## The XDS Un-Prefix Migration

This merges the `xds-unprefix-integration` branch into `main`, shipping the full rename migration.

### What changes for consumers

**Nothing breaks.** All existing `XDS*`-prefixed imports (`XDSButton`, `useXDSTheme`, etc.) continue to work unchanged via generated compat re-export aliases. But the **recommended API** is now bare names:

```tsx
// Before (still works via compat alias)
import {XDSButton} from '@xds/core/Button';

// After (recommended, now canonical)
import {Button} from '@xds/core/Button';
```

### What's in this PR

- **Source rename**: 320 files renamed `XDS*.tsx` → bare canonical (`Button.tsx`, `useTheme.ts`, etc.)
- **Compat generator inverted**: emits `XDS*` aliases *from* bare canonical (preserves module augmentation — verified empirically)
- **CSS dual-emit**: components render both `astryx-button` and `xds-button` classes; theme `@scope` selectors match either prefix
- **Data-attr dual-emit**: consumer-observable attrs (`data-astryx-theme`, `data-astryx-media`) emitted alongside legacy `data-xds-*`
- **Sibling `astryx.css`**: `@xds/core/astryx.css` export (byte-identical to `xds.css`, forward-compatible import path)
- **Doc prose bare-named**: CLI output, templates, docsite all present bare component names
- **CLI `astryx` alias**: runs the same binary under the new name

### Validation (on this merge commit)

- `@xds/core` tsc: **0 errors**
- Build: success (xds.css + astryx.css)
- Compat aliases / exports-sync / token-docs: all green
- **Full repo vitest: 4768/4768 pass (255 files)**

### Naming collisions disambiguated

The rename surfaced 3 pre-existing latent name clashes (resolved, 0 fbsource consumers):
- CodeBlock `Token` type → `SyntaxToken` (frees `Token` for the UI component)
- `SyntaxTheme` interface → `SyntaxThemeDefinition` (frees `SyntaxTheme` for the provider)
- theme `HeadingLevel` (`'h1'..'h6'`) → `HeadingTag` (frees `HeadingLevel` for the numeric 1-6)
- PowerSearch `DateRangePreset` → `DateRangeFilterPreset` (frees `DateRangePreset` for the calendar type)

### What stays unchanged

- Package scope: still `@xds/*` (renamed to `@astryx/*` at final cutover, P10)
[...]

PR #5555, opened 2026-08-26
fix(Typeahead,Tokenizer): the busy indicator is a Spinner in the field's end lane, not a clock on top of the clear button

### LINKED ISSUE #5554: Typeahead/Tokenizer: the busy indicator is a static clock, and it collides with the clear button
## Problem

Typeahead and Tokenizer paint their busy state with a **static clock**, and it collides with the clear button.

`BaseTypeahead.tsx:836-843` renders the indicator as `<Icon icon="clock" size="sm" color="secondary">`. That line is byte-identical to `TimeInput.tsx:637`, and in core `clock` otherwise means *time* — it is TimeInput's and DateTimeInput's picker affordance. Every sibling input paints busy with `<Spinner size="sm" />`: `TextInput.tsx:462`, `Selector.tsx:1540`, `MultiSelector.tsx:1618`, `TextArea.tsx:628`. So the glyph is both the wrong idiom and the wrong component's, and nothing spins during a search.

It also lands on top of the clear button. The indicator is an **in-flow flex item** and the input before it is `flex: 1`, so it is pushed to the row's inline end — which is where each wrapper independently parks an **absolutely-positioned** clear affordance (`Typeahead.tsx:182` `styles.clearButton`, `Tokenizer.tsx:254` `styles.endSection`). Neither wrapper reserves inline-end padding, so the two occupy the same lane.

Measured in Chromium 149 against a Storybook build of `main`, DPR 1, 320px field, `@astryxdesign/theme-neutral`:

| component | size | indicator box | clear button | overlap | glyph covered |
|---|---|---|---|---|---|
| Typeahead | sm | 303…327 | 312…332 | **15 × 20 px** | 11 of 16 px |
| Typeahead | md / lg | 303…327 | 310…330 | **17 × 20 px** | 13 of 16 px |
| Tokenizer | all | 308…332 | 307…327 | **19 × 20 px** | 15 of 16 px |

Width-independent — 240/320/480px all give the same 17px at md. In Tokenizer the indicator also paints *on top of* the button (later in-flow sibling over an absolutely-positioned one in the same stacking context), so **part of the ✕ is unclickable**.

Because it is in-flow, its position also moves with whatever sits beside it: with a value selected the input collapses to width 0, so the indicator appears inline beside the token rather than at the field's end.

Two smaller things in the same block:

- The combobox input never gets `aria-busy`, unlike `TextInput.tsx:453`. A screen reader is told the field is busy only by a `role="status"` label, and nothing marks the control itself.
- **No coverage exists.** No story renders either component loading (`Typeahead.stories.tsx`, `Tokenizer.stories.tsx` have no async source) and neither test file exercises `isLoading`.

## Evidence of demand

Reported from an app using an async `searchSource` — the state is reachable by typing into any Typeahead backed by a remote source, which is the case the component exists for. A sync source resolves too fast to see it.

## Why existing components don't cover this

The indicator is rendered inside the shared engine (`BaseTypeahead`), used by both `Typeahead.tsx:461` and `Tokenizer.tsx:772`. There is no wrapper a consumer can interpose and no `renderX` prop to reach it, so an app cannot fix or hide it. The end lane is duplicated and divergent between the two consumers — Typeahead's holds one button at inset 5px, Tokenizer's is a flex row at inset 8px already holding `endContent` — which is why the overlap differs between them and why fixing one would not fix the other.

## Accessibility considerations

- WCAG 2.5.8 target size: the clear button's 20×20 target is partly overlaid in Typeahead and partly covered in Tokenizer.
- The busy state should be on the control (`aria-busy`), not only in a sibling live region.
- Whatever replaces the glyph has to honour `prefers-reduced-motion`, which `Spinner` already does.

## Rough approaches

1. **One end lane per field, adornments as in-flow siblings in it** — what `TextInput`/`Selector`/`MultiSelector` already do, with the lane's width reserved from the input the way `TextArea.tsx:119-128` reserves `textareaWithStatus` / `textareaWithBusyStatus`. The engine would report loading rather than paint it, mirroring the existing `onOpenChange`.
2. **Reserve padding only**, leaving the indicator where it is. Smaller, but leaves two divergent lanes and does not fix Tokenizer with `endContent`.

Happy to take either; (1) looks right and also closes the pre-existing case where a long query runs under the ✕ with no spinner present at all (input box 25…327 against clear 310…330 at md).

freddymeta:
## Summary

**No longer stacked** — [#5385](https://github.com/facebook/astryx/pull/5385) landed, so this is replanted on `main` and reviewable on its own. Closes #5554.

## What was wrong

Three defects in one block, all on `main`:

1. **The busy indicator is a static clock.** `BaseTypeahead.tsx` rendered `<Icon icon="clock" size="sm" color="secondary">` — byte-identical to `TimeInput.tsx`, and in core `clock` otherwise means *time*. Every sibling input paints busy with `<Spinner size="sm" />`. Nothing spun during a search: `getAnimations({subtree: true})` finds nothing running on it.
2. **It landed on top of the clear button.** The indicator was an in-flow flex item and the input before it is `flex: 1`, so it was pushed to the row's inline end — which is where each wrapper independently parks an absolutely-positioned clear affordance. 17×20px at Typeahead md (13 of the 16px glyph covered) and 19×20px in Tokenizer. **Correction (re-measured):** the overlap is visual only — the clear button is positioned, so it paints above the in-flow indicator, and `elementFromPoint` at seven points across the covered band on `main` lands inside the button at every one. My earlier "part of the ✕ unclickable" was wrong.
3. **The combobox never got `aria-busy`**, unlike `TextInput`.

## The change

The engine reports the busy state rather than painting it, and each field renders `<Spinner size="sm" />` in the **one lane it already owns** at its inline end — Tokenizer's `endSection`, which already holds `endContent` and the clear button, and Typeahead's clear-button box, widened into an `endLane` flex row. `aria-busy` goes on the input.

A caller using `BaseTypeahead` directly is unaffected: it still renders its own visible, named "Loading" status, as a Spinner now rather than the clock, so the fix reaches those callers too. Passing the callback is what hands the indicator over, so a field never paints two.

**The input reserves the lane.** The lane is absolutely positioned — both wrappers are `flexWrap: 'wrap'`, so an in-flow sibling gets pushed onto a second row by a token — and an out-of-flow box reserves nothing, so at a narrow width the live query ran underneath it. `useEndLaneReserve` measures the rendered lane and returns the padding the input needs. Measured rather than assumed: the lane holds a clear button that comes and goes with the value, an indicator that comes and goes with the search, and, in Tokenizer, arbitrary `endContent`. There is no CSS that does this — the input cannot see a sibling's width, and a custom property set on the lane cannot travel sideways to it.

## Measured in Chromium, at 280px

Overlap of the input's content box (where text and the caret may go) with each lane control, across the six states, against a build of current `main`:

| state | main | here |
|---|---|---|
| Typeahead, value settled | **clear 17px** | 0 |
| Typeahead, value + search in flight | 0 | 0 |
| Typeahead, value idle | 0 | 0 |
| Tokenizer, value idle | **clear 25px** | 0 |
| Tokenizer, value settled | **clear 25px** | 0 |
| Tokenizer, value + search in flight | 0 | 0 |

**Correction (re-measured against a build of current `main`, 2026-09-02).** The two Typeahead `clear` rows above do not reproduce. In edit mode at 280px the input's border box ends at 259 and the clear starts at 270; at 200px, 179 against 190 — 0 overlap either way. The 17px figure is the *indicator over the clear button*, which does reproduce exactly. The real pre-existing clear-over-content overlap is Tokenizer's, and today it measures **20px**, not 25px. Everything on this branch still measures 0.

The two in-flight rows were already 0 on `main` — but only because the in-flow indicator reserved its own width. Moving it into the lane is what would have taken that away, which is the regression [@cixzhang](https://github.com/cixzhang) caught; the reserve is what holds them at 0. The other rows are the pre-existing case of the same bug, with no spinner in it at all — I had filed that away as separate and it belongs here, since it is the same missing reserve.

Indicator, same probe: `Icon`/**0 animations** on main → `Spinner`/1 animation here, both fields. `aria-busy` absent on main → `true` on both.

## Review changes since the last revision

- **The input reserves the rendered lane width** — the table above.
- **The default status is preserved for direct callers.** `BaseTypeaheadProps` is re-exported from the package entry point, so the base has callers this repo cannot see, and they painted no indicator of their own. It renders whenever no wrapper has taken it over.
- **The transport is internal**: `__onLoadingChange`, marked `@internal`, following `__queryEntries` from #5385 and `DefinedTheme.__inputTokens`.
- **No Effect.** The report goes out at the call site through a ref, so the field's state change batches into the commit React was already doing instead of forcing a second one — the doubled wrapper commits are gone. It is edge-triggered as well, so the redundant clear on every keystroke below the query threshold reports nothing at all.
- **A changeset**, which was missing.

## Coverage

There was none — no story rendered either component loading, and neither test file exercised `isLoading`, which is how a clock survived in the busy slot. Adds a `Loading` story to each (async source, 1.2s), the Tokenizer one with `hasClear` and `endContent` so all three lane occupants are present at once, plus unit tests for the three contracts jsdom can hold: the default status for a direct caller, the handover (and that the base then renders none), and the edge-triggered reporting.

## Test plan

- `pnpm build`, `pnpm lint:strict` (0 errors; the same 80 warnings `main` reports), `pnpm check:repo`, `pnpm -F @astryxdesign/core typecheck` — all clean.
- `pnpm exec vitest run packages/core/src/Typeahead packages/core/src/Tokenizer`: 132 tests pass.

### DIFF
--- a/packages/core/src/Tokenizer/Tokenizer.tsx
+++ b/packages/core/src/Tokenizer/Tokenizer.tsx
@@ -25,7 +25,12 @@ import React, {
   type ReactNode,
 } from 'react';
 import * as stylex from '@stylexjs/stylex';
-import {BusyIndicatorLaneProvider} from '../Typeahead/busyIndicatorLane';
+import {
+  BusyIndicatorLaneProvider,
+  createBusyIndicatorLane,
+  useIsBusy,
+  type BusyIndicatorLane,
+} from '../Typeahead/busyIndicatorLane';
 import type {BaseProps} from '../BaseProps';
 import type {SizeValue} from '../utils/types';
 import {BaseTypeahead} from '../Typeahead/BaseTypeahead';
@@ -394,6 +399,45 @@ const CREATABLE_ID_PREFIX = '__xds_create__';
  * />
  * ```
  */
+/**
+ * The field's inline-end lane: the busy Spinner, then `endContent`, then the
+ * clear button — pinned to the field's first row while tokens wrap below it.
+ *
+ * A separate component so that subscribing to the busy state re-renders THIS
+ * and nothing else. Subscribed from `Tokenizer`, a search transition
+ * re-rendered every selected token twice — twenty tokens meant forty renders
+ * per search for one glyph none of them contain.
+ *
+ * Renders nothing when the lane would be empty, so `useEndLaneReserve` sees no
+ * element, publishes no width, and the input keeps its full content box.
+ */
+function EndLane({
+  lane,
+  laneRef,
+  laneXStyle,
+  loadingLabel,
+  hasStaticContent,
+  children,
+}: {
+  lane: BusyIndicatorLane;
+  laneRef: (node: HTMLElement | null) => void;
+  laneXStyle: stylex.StyleXStyles[];
+  loadingLabel: string;
+  hasStaticContent: boolean;
+  children: ReactNode;
+}) {
+  const isBusy = useIsBusy(lane);
+  if (!isBusy && !hasStaticContent) {
+    return null;
+  }
+  return (
+    <div ref={laneRef} {...stylex.props(laneXStyle)}>
+      {isBusy && <Spinner size="sm" aria-label={loadingLabel} />}
+      {children}
+    </div>
+  );
+}
+
 export function Tokenizer<T extends SearchableItem>({
   label,
   isLabelHidden = false,
@@ -470,16 +514,21 @@ export function Tokenizer<T extends SearchableItem>({
   // Focus-within state for overflow truncation
   // Reported by BaseTypeahead so the indicator can live in this field's own
   // end lane, beside endContent and the clear button.
-  const [isLoading, setIsLoading] = useState(false);
-  const busyLane = useMemo(() => ({onBusyChange: setIsLoading}), []);
+  // The base owns the busy state; this field only paints it. Held in a store
+  // rather than this component's state — held here, a search transition
+  // re-rendered every selected token twice, for a glyph no token contains.
+  // See busyIndicatorLane.tsx.
+  const busyLane = useMemo(() => createBusyIndicatorLane(), []);
   // What sits in the end lane varies most here — a spinner, arbitrary
   // `endContent`, a clear button, or all three — so its width is measured
   // rather than assumed, and the input reserves it.
   const [laneRef, laneReserve] = useEndLaneReserve(END_LANE_INSET);
-  // Whether a lane renders at all — known from props and state, without
-  // measuring anything, which is why the reserve costs no extra commit.
-  const hasEndLane = Boolean(
-    isLoading || endContent || (hasClear && value.length > 0 && !isDisabled),
+  // The half of the lane's contents this component knows about. The busy half
+  // is the leaf's own business — folding it in here would mean reading the
+  // busy state during this render, which is exactly the re-render the store
+  // exists to avoid.
+  const hasStaticEndLane = Boolean(
+    endContent || (hasClear && value.length > 0 && !isDisabled),
   );
   const [isFocusedWithin, setIsFocusedWithin] = useState(false);
   const isTruncated =
@@ -850,7 +899,10 @@ export function Tokenizer<T extends SearchableItem>({
                 : undefined,
             // Not for the collapsed states above: those give the input no width
             // to pad, and `inputAtMax` zeroes its padding outright.
-            !(isAtMax || isTruncated) && hasEndLane && laneReserve,
+            // Applied whether or not a lane is up: the reserve resolves to
+            // zero until the lane publishes a width, so this does not need to
+            // know about the busy state.
+            !(isAtMax || isTruncated) && laneReserve,
           ]}
         />
       </BusyIndicatorLaneProvider>
@@ -866,25 +918,23 @@ export function Tokenizer<T extends SearchableItem>({
             disabled={isDisabled}
           />
         ))}
-      {hasEndLane && (
-        <div
-          ref={laneRef}
-          {...stylex.props(styles.endSection, endSectionSizeStyles[size])}>
-          {isLoading && (
-            <Spinner size="sm" aria-label={t('@astryx.typeahead.loading')} />
-          )}
-          {endContent}
-          {hasClear && value.length > 0 && !isDisabled && (
-            <InputClearButton
-              label={t('@astryx.tokenizer.clearAll')}
-              onClick={e => {
-                e.stopPropagation();
-                handleClearAll();
-              }}
-            />
-          )}
-        </div>
-      )}
+      <EndLane
+        lane={busyLane}
+        laneRef={laneRef}
+        laneXStyle={[styles.endSection, endSectionSizeStyles[size]]}
+        loadingLabel={t('@astryx.typeahead.loading')}
+        hasStaticContent={hasStaticEndLane}>
+        {endContent}
+        {hasClear && value.length > 0 && !isDisabled && (
+          <InputClearButton
+            label={t('@astryx.tokenizer.clearAll')}
+            onClick={e => {
+              e.stopPropagation();
+              handleClearAll();
+            }}
+          />
+        )}
+      </EndLane>
     </div>
   );
 
--- a/apps/storybook/rtl-audit/targets.json
+++ b/apps/storybook/rtl-audit/targets.json
@@ -92,24 +92,24 @@
   },
   {
     "component": "Typeahead",
-    "storyId": "core-typeahead--logical-order",
+    "storyId": "core-typeahead--rtl-end-lane",
     "dims": [
-      "D2"
+      "D4"
     ],
     "selectors": {
-      "prev": ".astryx-token",
-      "next": ".astryx-input-clear-button"
+      "overlay": "button[aria-label=\"Clear selection\"]",
+      "overlayRoot": ".astryx-typeahead"
     }
   },
   {
     "component": "Tokenizer",
-    "storyId": "core-tokenizer--logical-order",
+    "storyId": "core-tokenizer--rtl-end-lane",
     "dims": [
-      "D2"
+      "D4"
     ],
     "selectors": {
-      "prev": ".astryx-token",
-      "next": ".astryx-input-clear-button"
+      "overlay": "button[aria-label=\"Clear all\"]",
+      "overlayRoot": ".astryx-tokenizer"
     }
   }
 ]
--- a/apps/storybook/stories/Tokenizer.stories.tsx
+++ b/apps/storybook/stories/Tokenizer.stories.tsx
@@ -523,3 +523,29 @@ export const LogicalOrder: Story = {
   },
   name: 'Logical order',
 };
+
+/**
+ * The inline-end lane, populated at rest, for the RTL audit's D4 pass.
+ *
+ * Tokenizer's lane is absolutely positioned — it has to be, so it stays on the
+ * field's first row while tokens wrap below it — which makes its side entirely
+ * a matter of the writing mode. The busy Spinner shares that lane, and busy
+ * only exists mid-search, so a selected token with `hasClear` is what puts the
+ * lane on screen for a measurement that can be held still.
+ */
+export const RtlEndLane: Story = {
+  render: args => (
+    <Tokenizer
+      {...args}
+      searchSource={userSource}
+      value={[users[0]]}
+      onChange={() => {}}
+      hasClear
+    />
+  ),
+  args: {
+    label: 'Team Members',
+    placeholder: 'Search people...',
+  },
+  name: 'RTL end lane (token + clear)',
+};
--- a/apps/storybook/stories/Typeahead.stories.tsx
+++ b/apps/storybook/stories/Typeahead.stories.tsx
@@ -391,3 +391,26 @@ export const LogicalOrder: Story = {
   },
   name: 'Logical order',
 };
+
+/**
+ * The inline-end lane, populated at rest, for the RTL audit's D4 pass.
+ *
+ * The lane holds the busy Spinner, and the busy state only exists between a
+ * keystroke and its response — nothing the audit can hold still. A selected
+ * value with `hasClear` puts the same lane on screen with no interaction, so
+ * the audit measures the geometry the indicator lands in.
+ */
+export const RtlEndLane: Story = {
+  render: () => (
+    <div style={{width: 320}}>
+      <Typeahead
+        label="Fruit"
+        searchSource={fruitSource}
+        value={fruits[0]}
+        onChange={() => {}}
+        hasClear
+      />
+    </div>
+  ),
+  name: 'RTL end lane (selected + clear)',
+};
--- a/packages/core/src/Tokenizer/Tokenizer.test.tsx
+++ b/packages/core/src/Tokenizer/Tokenizer.test.tsx
@@ -27,6 +27,24 @@ import {__resetLiveRegionsForTest} from '../hooks/useAnnounce';
 import type {SearchSource, SearchableItem} from '../Typeahead/types';
 import {TestIcon} from '../__tests__/TestIcon';
 import {InternationalizationProvider} from '../i18n';
+import type * as TokenModule from '../Token';
+
+// Every Token render, in order, by label.
+//
+// A spy that WRAPS the real component rather than replacing it: a stand-in
+// would measure the stand-in, and the number under review is how often the
+// real token subtree re-renders during a search.
+const renderedTokens: string[] = [];
+vi.mock('../Token', async importActual => {
+  const actual = await importActual<typeof TokenModule>();
+  return {
+    ...actual,
+    Token: (props: Parameters<typeof TokenModule.Token>[0]) => {
+      renderedTokens.push(typeof props.label === 'string' ? props.label : '');
+      return actual.Token(props);
+    },
+  };
+});
 
 // Test-supplied announcement strings: the assertions below depend on no
 // catalog, and no hardcoded English in the component can satisfy them.
@@ -1469,7 +1487,17 @@ describe('Tokenizer end-lane reserve', () => {
 
   const laneHost = (container: HTMLElement) =>
     container.querySelector<HTMLElement>(
-      '[style*="--_tokenizer-end-lane-width"]',
+      '[style*="--_tokenizer-end-lane-reserve"]',
+    );
+
+  // The variable carries the whole reserve — the lane's inset plus its
+  // measured width — so its ABSENCE means "no lane", not "a lane of zero
+  // width". That distinction is what lets the input apply the padding rule
+  // unconditionally and still keep its full content box when nothing is in
+  // the lane.
+  const reserveOf = (container: HTMLElement) =>
+    laneHost(container)?.style.getPropertyValue(
+      '--_tokenizer-end-lane-reserve',
     );
 
   it('reserves the lane\u2019s untransformed width, not its on-screen width', () => {
@@ -1488,11 +1516,9 @@ describe('Tokenizer end-lane reserve', () => {
         hasClear
       />,
     );
-    expect(
-      laneHost(container)?.style.getPropertyValue(
-        '--_tokenizer-end-lane-width',
-      ),
-    ).toBe(`${LANE_LOCAL_WIDTH}px`);
+    expect(reserveOf(container)).toBe(
+      `calc(var(--spacing-2) + ${LANE_LOCAL_WIDTH}px)`,
+    );
   });
 
   it('reserves the lane when the spinner is the only thing in it', () => {
@@ -1517,11 +1543,9 @@ describe('Tokenizer end-lane reserve', () => {
     });
 
     expect(screen.getByRole('status')).toBeInTheDocument();
-    expect(
-      laneHost(container)?.style.getPropertyValue(
-        '--_tokenizer-end-lane-width',
-      ),
-    ).toBe(`${LANE_LOCAL_WIDTH}px`);
+    expect(reserveOf(container)).toBe(
+      `calc(var(--spacing-2) + ${LANE_LOCAL_WIDTH}px)`,
+    );
   });
 
   it('does not pad the input on the collapsed paths', () => {
@@ -1584,6 +1608,53 @@ describe('Tokenizer end-lane reserve', () => {
 
     expect(afterStart).toBe(1);
     expect(commits.length).toBe(2);
+  });
+
+  it('re-renders no token when a search starts or settles', async () => {
+    // The regression this guards. The busy state has one owner — the base —
+    // and the field only needs it to decide whether one Spinner is on screen.
+    // Mirrored into the field's own state, every transition re-rendered the
+    // whole field including every selected token: measured at six tokens,
+    // 6 renders at search start and 6 more at settlement, and 20/40 at
+    // twenty. A token contains no part of the indicator, so the correct
+    // number is zero at any count.
+    const {source, settle} = pendingSource();
+    const selected = users.slice(0, 6);
+    renderedTokens.length = 0;
+
+    render(
+      <Tokenizer
+        label="Users"
+        searchSource={source}
+        value={selected}
+        onChange={() => {}}
+        hasClear
+        debounceMs={0}
+      />,
+    );
+    expect(new Set(renderedTokens).size).toBe(selected.length);
+    renderedTokens.length = 0;
+
+    await act(async () => {
+      fireEvent.change(screen.getByRole('combobox'), {target: {value: 'Al'}});
+    });
+    // The indicator is up, so the transition really did happen. Scoped by
+    // name: the announcer for result counts is a role="status" live region
+    // too, and it is not the indicator.
+    expect(screen.getByRole('status', {name: 'Loading'})).toBeInTheDocument();
+    // ...and it cost no token a render.
+    expect(renderedTokens).toEqual([]);
+
+    await act(async () => {
+      settle([]);
+      await Promise.resolve();
+    });
+    await waitFor(() => {
+      expect(
+        screen.queryByRole('status', {name: 'Loading'}),
+      ).not.toBeInTheDocument();
+    });
+    expect(renderedTokens).toEqual([]);
   });
 
   it('shares one observer across every field on the page', () => {
--- a/packages/core/src/Tokenizer/useEndLaneReserve.ts
+++ b/packages/core/src/Tokenizer/useEndLaneReserve.ts
@@ -35,7 +35,7 @@ import {observeResize, unobserveResize} from '../utils/sharedResizeObserver';
  * carries a measurement into CSS without carrying it through React, so a lane
  * that grows or shrinks repaints without re-rendering the field.
  */
-const LANE_WIDTH_VAR = '--_tokenizer-end-lane-width';
+const LANE_RESERVE_VAR = '--_tokenizer-end-lane-reserve';
 
 // Keep the input's text and caret out from under the lane.
 //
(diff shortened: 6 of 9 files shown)


freddymeta:
Negative control, run after the fact rather than asserted — the same two stories and the same probe, against `main`'s components:

| | `main` | this PR |
|---|---|---|
| Typeahead indicator | `Icon`, **0 animations** | `Spinner`, 1 animation |
| Typeahead overlap | **17 px** (303…327 over clear 310…330) | **0 px** (292…306, clear 310…330) |
| Tokenizer indicator | `Icon`, **0 animations** | `Spinner`, 1 animation |
| Tokenizer overlap | **19 px** (388…412 over clear 387…407) | **0 px** (284…298, clear 387…407) |
| `aria-busy` | absent on both | `true` on both |

`0 animations` is the "static clock" claim measured rather than eyeballed: `getAnimations({subtree: true})` finds nothing running on the indicator on `main`.

Gates: 12285 tests pass, `lint:strict` 0 errors, typecheck clean. The one failing test is `story-tree.test.ts` → `Core/Icon/Size Theming`, which is red on `main` from #5465 and only surfaces once `apps/storybook/dist` exists locally.

cixzhang (changes requested):
Thanks — the Spinner and `aria-busy` are the right direction, but moving loading out of `BaseTypeahead` leaves three regressions.

At 280px the absolute lane covers the live query: Typeahead’s Spinner overlaps 14×14px of the input and clear overlaps 17×20px (both zero on the parent); Tokenizer has the same Spinner/input overlap. The caret and trailing characters render under the controls while search runs. Could the input reserve the rendered lane width?

| Parent | Head | Settled control |
|---|---|---|
| Parent busy | Head busy | Head settled |

`BaseTypeahead` is released. Direct callers previously got a visible, named “Loading” status; this head removes it and adds an undocumented public callback, so existing callers silently lose that feedback. Its Effect also doubles wrapper commits at search start and settlement. Could we preserve the default status, keep wrapper transport internal, report the transition without an Effect, and add a changeset?

[Reviewed by Robohands]

freddymeta:
All four are in, and you were right about the regression — thanks for measuring it.

**The overlap.** My "pre-existing, filing separately" note was wrong in the state you tested. The in-flow indicator I removed *was* reserving its own width, so moving it into the lane is what took the reserve away while a search is out. The lane is out of flow and reserves nothing, and no CSS fixes that from the input's side: it cannot see a sibling's width, and a custom property set on the lane cannot travel sideways to it. So `useEndLaneReserve` measures the rendered lane and returns the padding the input needs — measured rather than assumed, because the lane holds a clear button that comes and goes with the value, an indicator that comes and goes with the search, and, in Tokenizer, arbitrary `endContent`.

That also folds in the case I had punted, since it is the same missing reserve without a spinner in it. At 280px, overlap of the input's content box with each lane control, against a build of current `main`:

| state | main | here |
|---|---|---|
| Typeahead, value settled | **clear 17px** | 0 |
| Typeahead, value + search in flight | 0 | 0 |
| Typeahead, value idle | 0 | 0 |
| Tokenizer, value idle | **clear 25px** | 0 |
| Tokenizer, value settled | **clear 25px** | 0 |
| Tokenizer, value + search in flight | 0 | 0 |

**The released base.** `BaseTypeahead` renders its own visible, named status again, whenever no wrapper has taken the indicator over — so a direct caller loses nothing, and gains the Spinner in place of the clock. Passing the callback is what hands it over, so a field never paints two. The transport is `__onLoadingChange`, `@internal`, following `__queryEntries` from #5385.

**The Effect is gone.** The report goes out at the call site through a ref, so the wrapper's `setState` batches into the commit React was already doing rather than forcing a second one. It is edge-triggered too, which fixes something I had not noticed: every keystroke below `minQueryLength` clears the flag, so the Effect version handed the wrapper a `false` per character while the user typed. Now those report nothing at all. Three unit tests cover the default status, the handover, and the report count.

**Changeset** added.

One process note: #5385 landed while I was working on this, so the branch is replanted on `main` and is no longer stacked — the diff is this change only.

freddymeta:
Audited the change rather than just re-running my own tests. Two things fixed, three verified, one worth your judgement.

**Fixed**

- **The latest-callback ref was assigned during render.** The repo splits on exactly this case: plain values go during render (`snapPointsRef` in `BottomSheetPanel`), but the callback refs beside them — `onMotionStartRef`, `onMotionCompleteRef` — sync in a layout effect, so a render React discards cannot leave the ref pointing at the abandoned pass's callback. Now matches. It writes a ref and nothing else, so no wrapper re-renders for it and the doubled commits stay gone.
- **The ResizeObserver fallback read `contentRect`** (content box) where the primary path reads `borderBoxSize`. Equal today — the lane has no padding or border — and silently short the moment one is added. Reads `offsetWidth` now.

**Verified**

- **Every size, both directions, under StrictMode**: `sm`/`md`/`lg` × LTR/RTL × busy/settled, both components — spinner overlap 0 and clear overlap 0 in all of them, and the padding lands on the inline-end side only in RTL (it is logical, so it mirrors). Worth noting `lg` was already inheriting the `md` lane inset before this change; my reserve uses the same ternary as the lane's positioning, so the two agree at every size.
- **No feedback loop and no console noise.** The reserve changes the input's padding, which could in principle resize the lane and re-fire the observer — it does not, because the lane is out of flow and sized by its own content. Zero console errors or warnings across the whole matrix, StrictMode included, so no `ResizeObserver loop completed` either.
- **The reserve does not squeeze the text area.** Fields from 360px down to 140px with all three lane occupants present: no row loses its content box, and the input is in fact *wider* than on `main` at every width (342px vs 314px at 360px), because removing the in-flow indicator gives more room back than the reserve takes.
- Full `packages/core` suite: **283 files / 8024 tests pass**. `lint:strict` 0 errors (the same 80 warnings `main` reports), `check:repo` clean including `check:sync` for the new file, `typecheck` clean for core and Storybook.
- Both new stories render in a built Storybook: `aria-busy="true"`, the indicator is an SVG with 1 running animation, overlap 0, no console errors.

**Your call, not a defect**

`Core/Tokenizer → Loading` reserves 131px, because the lane genuinely is 123px there — spinner + `"1 selected"` end content + clear button. The field keeps a 145px text area at its 400px story width, so it is correct rather than miscalculated, but a wide `endContent` does eat the input by exactly its own width. That is inherent to reserving honestly; on `main` the text simply ran underneath it. If you would rather the input scrolled under a wide lane instead of shrinking, that is a different trade and I would take it as a follow-up.

One correction to my earlier comment: I claimed Selector, MultiSelector, CommandPalette and DateTimeInput were direct `BaseTypeahead` callers. They are not — those matches were comments referencing it. The real direct callers are Typeahead, Tokenizer, and the CLI showcase block `blocks/components/BaseTypeahead/BaseTypeaheadCustomSearch.tsx`, which is the in-repo example of the case you flagged and now renders the preserved Spinner status.

packages/core/src/Field/useEndLaneReserve.ts
```diff
@@ -0,0 +1,111 @@
...
+
+    // jsdom implements no ResizeObserver, and this runs in every consumer's
+    // component tests. `offsetWidth` is 0 there too, so the reserve is simply
+    // absent rather than wrong.
+    if (typeof ResizeObserver === 'undefined') {
+      setWidth(node.offsetWidth);
+      return;
+    }
+
+    const observer = new ResizeObserver(entries => {
```
cixzhang: Please use the existing shared observer instead of allocating one per field lane.

cixzhang (changes requested):
Thanks — the prior overlap, direct-Base, internal-transport, and changeset asks are fixed. The render-cost ask is still open: same-worktree A/B takes a 20-token search from 20 to 120 token renders, and disabling only the reserve returns it to 60. Could we keep the zero-overlap result without adding full-field commits at search start and settlement?

| Prior head | Current head |
|---|---|
| Prior head | Current head |

[Reviewed by Robohands]

freddymeta:
Render cost fixed — the reserve now costs **zero** commits of its own (`b3ea3b1b5`), and it is rebased onto main.

**Root cause.** `useEndLaneReserve` held the measured width in React state. The lane changes size exactly twice a search — spinner in, spinner out — so each transition re-rendered the whole field to carry a number no JavaScript ever reads. That is the doubling you measured.

**Fix.** The measurement reaches CSS as a custom property written straight to the field wrapper and inherited by the input. The rule is static now:

```
paddingInlineEnd: calc(<lane inset> + var(--_astryx-end-lane-width, 0px))
```

One class, generated once, instead of a new one per width. React never sees the value.

**Measured, same test against both implementations** — `Profiler` around the field, one search from start to settle:

| | commits at search start | total |
|---|---|---|
| state-held reserve | **2** | 4 |
| custom property | **1** | 2 |

The remaining two are the ones the search itself owes: the spinner arriving and the spinner leaving.

Worth flagging how that test is built, because the obvious version does not work: **jsdom reports every width as 0**, so a state-held reserve never re-renders there and the regression is invisible. The test stubs a `ResizeObserver` that reports a width on observe — the smallest thing that makes it reproducible in CI. It is written to be implementation-neutral (it waits on no DOM signal either implementation owns), so it genuinely fails on the old code rather than erroring on it: I ran it both ways to confirm.

**And the observer is shared.** It now goes through `observeResize` — the same singleton `useTruncation` uses — rather than constructing one per lane. Three fields on a page created three observers before and create one now, so the browser dispatches one callback a frame rather than N. There is a test for that too: `instances <= 1` across three mounted fields, which fails at 3 on the old code.

**The accepted fix still holds.** Re-measured in Chromium at 280px while busy:

```
input content box ends   268
lane starts              276     → 0px overlap with the caret
--_astryx-end-lane-width 14px
padding-inline-end       19px    (5px inset + 14px lane)
```

and on settle the property is removed, so the input takes the room back rather than keeping a stale reserve.

Changeset updated with both the property and the shared observer.

freddymeta:
/accept-visual

The report lists **634 added, zero changed** — new baselines for stories that had none, and not a single existing shot moved. That is the expected result: this round is a render-cost refactor (the lane width moved from React state to a CSS custom property), and the computed padding is identical either way. Verified in Chromium at 280px — content box ends 268, lane starts 276, same as before.

freddymeta:
Re-verified on the current head (main merged, 4 commits), including the Tokenizer case I had not actually measured before.

**Tokenizer is the stronger proof**, because its lane genuinely changes width — the story carries a token, `endContent` ("1 selected") and a clear button, and the spinner joins them mid-search:

| state | lane width | input `padding-inline-end` | caret under lane |
|---|---|---|---|
| idle | 101px | 109px | **0** |
| searching | **123px** | **131px** | **0** |
| settled | 101px | 109px | **0** |

The padding tracks a 22px swing in both directions — `calc(8px inset + var(--_astryx-end-lane-width))` — with **no React commit for either transition**. That is the mechanism working, not just the end state being correct.

Typeahead at 280px, for completeness: busy → content box ends 268, lane starts 276, overlap 0, `--_astryx-end-lane-width: 14px`; settled → the property is removed and padding returns to 0.

Also confirmed, since these are easy to claim and easy to get wrong:

- **No state in the hook**: 0 `useState`, 0 `new ResizeObserver`, 4 references to the shared `observeResize`/`unobserveResize` singleton.
- **421 tests pass** across Typeahead, Tokenizer, Field and Text (Text shares the observer).
- **Lint is clean.** The 3 warnings on `BaseTypeahead.tsx` are pre-existing — identical on main at line 375; my change only shifted them to 401.
- The hook is 30 lines of logic with cleanup via the React 19 ref-callback return, so there is no separate teardown path to keep in step.

rubyycheung (approved):

cixzhang (changes requested):
The new measured end lane mixes viewport-transformed width with local CSS padding. In Chromium, `scale(.5)` caused 29.14px overlap and `scale(2)` caused a 253.44px excess gap.

Please reuse the established TextInput/input-container pattern for Typeahead instead of adding a measured lane, and add regression coverage for Tokenizer's busy-only and collapsed paths.

[Reviewed by Robohands]

freddymeta:
Both fixed, and the first one was a real bug — thank you.

## The coordinate-space mix

Reproduced before touching anything, on a 98px Tokenizer lane in Chromium:

| | published var | lane `offsetWidth` | result |
|---|---|---|---|
| `scale(.5)` | 49px | 98 | query back under the controls by **22.83px** |
| `scale(1)` | 98px | 98 | 0 |
| `scale(2)` | 196px | 98 | caret in a **202.69px** gap |

`getBoundingClientRect()` is viewport space and carries every transform above the element; the padding it fed is local space. `offsetWidth` is the untransformed border-box width — 98 at all three — so the reserve uses that. It is already an integer, which is the rounding the old `Math.ceil` was there for.

## Typeahead: no measured lane

Its spinner and clear button are now plain in-flow flex siblings of the input — TextInput's shape for exactly these controls. An in-flow box takes up room, so nothing can run under it and there is nothing to measure. The reserve and the absolute lane are gone from this field entirely; `useEndLaneReserve` is Tokenizer's alone now.

What blocked that was `flex-wrap: wrap` on Typeahead's wrapper, which turns out not to be a decision about this field at all — it arrived wholesale in the #2941 rename migration. The shared field base does not set it, and TextInput does not use it. It also cannot coexist with in-flow end controls, because flex moves an item to a new line rather than shrinking it: with it, a long value put the controls on a row of their own and a 280px field grew to 46px tall. Removed, the token ellipsizes instead — which is what Token already does — and the field stays one row.

Two rules I tried on the way there did nothing, and are not in the diff: `min-width: 0` on the token and `flex-shrink: 0` on the lane both measured byte-identical with and without.

Tokenizer keeps the lane, because it genuinely can't use the in-flow shape — its lane stays pinned to the first row while tokens wrap below it.

## Measured

before and after

Busy, at each scale. Typeahead: no reserve at all, overlap **0 / 0 / 0**. Tokenizer: 98px published at all three, overlap **0 / 0 / 0**, constant 3.34px local gap. The clearest cell is `scale(2)` Tokenizer — before, the query "ada" is pushed out of the field completely; after, it sits between the token and the spinner.

## Coverage

Tokenizer: the **busy-only** path (spinner as the lane's only occupant) and the **collapsed** path (no reserve when the input has no width to pad), plus a direct regression test for this bug — the stub reports 24px from `offsetWidth` and 12px from the rect, and the published value must be 24px.

Typeahead: three guards that the controls stay in flow — no absolute lane, no reserve property, no wrapping.

Reverting either fix fails its test: the measurement revert gives `expected '12px' to be '24px'`, and putting `flex-wrap` back fails the one-row guard.

285 test files pass; build, core + docs + Storybook typechecks, `check:repo` and `lint:strict` all clean.

*(Assets on the `assets/pr-5555-r3` branch — asset-only, deletable with the PR. I have no fork.)*

freddymeta:

Re-measured against a build of current `main` carrying the same two stories, so before/after is like-for-like.

| | `main` | head |
|---|---|---|
| Typeahead indicator | `astryx-icon`, **0 running animations** | `astryx-spinner`, **1** |
| Typeahead indicator ∩ clear button | 17 × 20 px | **0** |
| Tokenizer indicator ∩ clear button | 19 × 20 px | **0** |
| Tokenizer clear ∩ input content, idle | 20 px | **0** |
| `aria-busy` | absent | `true` |
| field box, idle → busy | 280 × 32 | 280 × 32 (FR1 holds) |

**The coordinate-space mix is fixed.** The lane publishes `offsetWidth`, not `getBoundingClientRect().width`. At `scale(.5/1/2)` the published value stays 123px while the rect reads 61/123/246. Negative control — republish the rect: 14px overlap at `.5`, a 125.95px gap at `2`, and the field grows 32 → 55px. Restored: 0 / 3.05px / 32px at all three.

**What the records required that the PR was missing.** The indicator is a `Spinner` — a shared primitive — so INV5 says it delegates rather than earning a target, but the part has to be in the inventory for that disposition to exist. Added a `Spinner` anatomy row to both docs and mapped it in `Typeahead.spec.md` as `delegatesTo component:Spinner/spinner`, the disposition `TextArea`, `CheckboxList` and `CommandPalette` already carry. No target, no public var, no DOM change. Also renamed `--_astryx-end-lane-width` → `--_tokenizer-end-lane-width`; it was the only `--_astryx-*` name in the repo.

**Two of my own earlier claims were wrong, and are corrected in the body and changeset:**
- *"part of the ✕ unclickable"* — false. `elementFromPoint` at seven points across the covered band on `main` lands inside the clear button every time. Visual overlap only.
- *"Typeahead, value settled → clear 17px on main"* — doesn't reproduce. The real pre-existing content overlap is Tokenizer's, and it's 20px, not 25px.

Gates: 8960 tests pass; build, typecheck, typecheck:docs, check:repo clean; `lint:strict` 85 warnings, matching a clean `main` worktree exactly. CI: 20 green, `pr-rtl` the repo-wide soft-red.

Note: one vitest run out of four reported a single failure I didn't capture the name of; the other three were 8960/8960. Reporting it rather than claiming four clean runs.

freddymeta:
### Screenshots

Chromium, built Storybook at this head, 3× DPR.

**Tokenizer** — the end lane, busy, with a token present. Spinner, count, clear — in flow, no overlap:

Tokenizer busy

`spinner ∩ clear = 0 × 14px` (zero horizontal), field 400 × 32.

**Typeahead** — the ring in the end lane:

Typeahead busy

`aria-busy=true`, field 320 × 32.

**No before shot, and here is why.** I went to capture one and neither component has a busy state on `main` at all — `Typeahead.tsx` and `Tokenizer.tsx` there contain no `isBusy`/`isLoading`, and the `loading` stories are added by this PR (main's Typeahead story list ends at `status-variant-comparison`). So the 17×20 and 19×20 overlaps in my earlier table come from rendering the *indicator element* main would place, not from a story you can open on main. Worth knowing when reading that table.

*(The two images are committed on a scratch branch `assets/rtl-shots` purely so they render here — say the word and I will delete it, or move them somewhere you prefer.)*

freddymeta:
Fixed here rather than in #5682, because this is where the prop is born: #5555 landing on its own would ship `__onLoadingChange` publicly regardless of what #5682 does.

You are right that `@internal` is not a boundary. `BaseTypeaheadProps` is re-exported from the package entry point, so the tag is a note to a reader and the name still ships.

The handoff now travels through a context in `busyIndicatorLane.tsx`, which `index.ts` does not export. The wrapper provides its setter around the base; the base subscribes. Nothing is left on the public prop surface to find.

Behaviour is unchanged — a lane present still means the wrapper paints the indicator in the end lane it already owns and the base renders none; no lane still means the base renders its own visible, named status.

A test pins the seam: the entry point exports neither the provider nor the hook, and `BaseTypeahead.tsx` no longer contains the prop. It fails against the previous commit.

8611 core tests, `check:repo` green. #5682 is rebased on this and carries only the content lane.

packages/core/src/Tokenizer/Tokenizer.tsx
```diff
@@ -461,6 +468,19 @@ export function Tokenizer<T extends SearchableItem>({
   }));
 
   // Focus-within state for overflow truncation
+  // Reported by BaseTypeahead so the indicator can live in this field's own
+  // end lane, beside endContent and the clear button.
+  const [isLoading, setIsLoading] = useState(false);
+  const busyLane = useMemo(() => ({onBusyChange: setIsLoading}), []);
```
cixzhang: Could this avoid mirroring BaseTypeahead state? Six selected tokens render 12 extra times per search.

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks — the overlap, Spinner, aria-busy, and transform fixes now hold in Chromium. One performance regression remains: Tokenizer mirrors BaseTypeahead’s busy state into its parent, so each search rerenders every selected token at start and settlement. Main stays at 0 token renders; this head reaches 6/12 at six tokens and 20/40 at twenty. Could we keep loading single-owned and add a render-count regression so search transitions don’t rerender the token list?

[Reviewed by Robohands]

freddymeta:
Both open items are addressed on `e092e4c6`, and the `pr-rtl` failure with them.

## The mirroring

You were right that it was a mirror and not a cost of the design. The busy state has one owner — the base — and each wrapper was copying it into its own `useState` just to decide whether one Spinner was on screen.

`BusyIndicatorLane` is now a store (`subscribe` / `getIsBusy` beside the existing `onBusyChange`), and the only thing that reads it is the leaf that paints the indicator, via `useSyncExternalStore`. Neither wrapper reads the busy state during its own render, so neither re-renders for it and the token count stops scaling anything.

The regression test is the number you measured: a spy that **wraps** the real `Token` records every token render across a search at six tokens, and asserts the list is empty at start and at settlement. Negative control — re-subscribing from `Tokenizer` itself, the shape this replaces — fails it with `['Alice', 'Bob', 'Charlie', 'Diana']`.

Two simplifications fell out, which is the part that suggests the mirror was the real problem rather than a tax:

- `hasEndLane` is gone from Tokenizer — the lane's leaf knows whether it has anything to show.
- `useEndLaneReserve` publishes the whole reserve rather than the width alone, so the variable's **absence** means "no lane" instead of "a lane of zero width". The input's padding rule is applied unconditionally and resolves to 0 with no lane, which is what lets the field stop tracking whether a search is running.

## pr-rtl

Failing on this head for a reason unrelated to your review: `COV: 0 measured / 2 gap`, exit 1. The audit scopes every component a PR touches and wants each measured or carrying a written not-applicable reason, and Tokenizer/Typeahead were neither.

They are genuinely direction-sensitive — this PR's own subject is a lane pinned to the field's inline end — so I added measurement rather than a reason. A D4 target each, over a story that puts that lane on screen at rest (the busy state only exists mid-search, so it cannot be held still for the audit; the clear button occupies the same lane).

Verified by building Storybook and running the audit, not by reasoning about it:

```
CUR  RTL-ready  core/typeahead   {"D4":"pass"}
CUR  RTL-ready  core/tokenizer   {"D4":"pass"}
COV : 2 measured / 0 verified N-A / 0 gap / 0 stale
exit 0
```

The lane mirrors rather than merely differing: side fraction LTR 0.94 → RTL 0.06 (Typeahead) and 0.95 → 0.05 (Tokenizer), so a lane pinned with a physical edge would sit at ~0.94 in both and score `fail`.

## Checks

- packages/core: 288 files / 8606 tests pass
- typecheck clean; the 3 `locales/pseudo.json` errors are pre-existing on main and are generated artifacts
- eslint clean on everything touched; the 3 `require-baseprops-passthrough` warnings on `BaseTypeahead.tsx:383` are pre-existing on main (checked by stashing)

No changeset — `busyIndicatorLane` is not exported from the package entry point, so none of this is public API, and the test asserting that still passes.

One correction to my own earlier note: the previous revision's claim that the transform fix needed a bespoke measured lane still stands, but I had also described the reserve as costing no commit "because it is not state" — it was not state, yet the field still committed twice per search through the mirror above. The reserve was innocent; the mirror was not.

cixzhang (approved):
Thanks — #5682 landed the earlier stack after this head, so it needs a rebase. The prior render regression is fixed; approved, with #5682’s content lane preserved.

[Reviewed by Robohands]

freddymeta:
Rebased the approved performance follow-up onto current main and squashed it to one commit (`cc7562437b`). Main already contains the earlier #5555 behavior through landed #5682, so this branch now carries only the remaining approved delta: keep busy state single-owned and avoid rerendering every selected token at search start/settlement.

Conflict resolution preserves the landed D2 stories while using this approved head’s stricter D4 fixtures/targets for the actual inline-end lanes.

- Typeahead/Tokenizer: **3 files / 150 tests passed**
- full core: **288 files / 8632 tests passed**
- core TypeScript and component lint: clean
- `pnpm check:repo` and `pnpm build`: pass
- exact RTL scope: Typeahead D4 pass, Tokenizer D4 pass; **2 measured / 0 gaps**

Ready for CI and refreshed approval.

cixzhang (approved):
Thanks — zero token rerenders at 4 and 20 tokens, with unchanged LTR/RTL pixels. Please retain the token/clear order check alongside the new lane-side check.

[Reviewed by Robohands]

freddymeta:
@cixzhang The approved single-owner/perf follow-up is rebased on current main at exact head `cc7562437b`. Main already contains the earlier behavior through landed #5682; this branch now carries only the approved remaining delta. All GitHub Actions, Vercel, 8632 core tests, TypeScript, lint, `check:repo`, build, and exact Typeahead/Tokenizer D4 audits are green.

Visual acceptance has **24 added, 0 removed** frames — the same default Typeahead/Tokenizer stories across 12 themes that were reviewed during #5682, not changes to existing baselines. Could you accept with:

`/accept-visual 33621837488/1 The added baselines show the default empty Typeahead and Tokenizer fields consistently across all 12 themes; the behavior change is exercised separately by the approved render-count and end-lane tests.`

Then refresh the exact-head approval so I can land it?

freddymeta:
@cixzhang The merge from current main made the earlier visual command stale. The final-head report for `f237e04034` is **12 shots: 4 unchanged, 0 changed, 8 added, 0 removed, 0 failed**. The 8 additions are only the default Typeahead/Tokenizer frames in butter and stone light/dark; no existing baseline changed.

Could you accept the current report with:

`/accept-visual 33630729604/1 The added butter and stone baselines show the default empty Typeahead and Tokenizer fields consistently; all existing visual baselines are unchanged, and the approved behavior is exercised by render-count and end-lane tests.`

All current-head Actions pass and the approved content is unchanged by the main merge.

PR #5558, opened 2026-08-26
fix(Badge): a long label no longer escapes its container

freddymeta:
## Summary

`Badge` sets `white-space: nowrap` with nothing to clip it — the one pairing that neither wraps nor truncates. A label wider than the space available doesn't ellipsize, it renders **outside** the badge's container, over whatever sits beside it.

```tsx
<div style={{width: 100}}>
  <Badge variant="pink" label="Awaiting security review" />
</div>
```

```diff
  whiteSpace: 'nowrap',
+ maxWidth: '100%',
+ minWidth: 0,
```
plus `overflow: hidden; text-overflow: ellipsis; min-width: 0` on a new inner label span.

Rubric **B10** — *"long text truncates or wraps with no layout breakage"* — severity FIX.

## Measured in Chromium

| | main | this PR |
|---|---|---|
| 100px column | **163px wide, 63px outside it** | 100px, 0 spill |
| 140px column, with icon | **175px, 35px outside** | 140px, 0 spill |
| fixed-layout table cell | **163px, 64px over the next cell's text** | 98px, 0 spill |
| 300px column, `label="Active"` | 53px | **53px — unchanged** |

That last row is the one that matters for risk: a badge with room to spare is identical before and after. The clamp only ever changes a case that was already overflowing.

## Why the inner span

`text-overflow` needs a block container. The badge root is `inline-flex`, and taking it off that to get one would cost the icon its vertical centring — so the ellipsis goes on the label instead, and with an icon the icon holds its place while the label gives way.

`minWidth: 0` is not decoration. As a flex item the badge's automatic minimum size is its max-content width, which holds it at full text width and undoes `maxWidth: 100%` entirely; the label span needs the same for the same reason.

## What is deliberately not here

**A tooltip carrying the full text.** Only-when-truncated needs a `scrollWidth > offsetWidth` measurement, which means hooks — and `Badge` is currently one of the ~24 files in `packages/core/src` with no `'use client'`. [[RSC Compatibility]] names `Badge` specifically among the components that should *become* server-renderable, so quietly making it a client component inside a defect fix is the wrong trade to make silently. It is a separate PR with its own case to argue, and this fix stands without it.

## Test plan

- `pnpm build`, `verify-exports`, `check:repo`, `lint:strict`, core + storybook + CLI-strict typechecks — clean. Badge: 8/8 existing tests pass **unchanged**.
- `pnpm a11y:audit -- --components Badge`: 0 new violations. (It also reports 2 `color-contrast` baseline entries as resolved — I checked, those report identically on `main`, so they are **not** this PR's doing.)
- New story: **Long labels in tight space** — the 100px case, the icon case, and a badge with room to spare side by side.
- **No unit test for the clamp**: jsdom has no layout, so the only thing it could assert is the declaration itself — the shape of test that makes a feature look covered while it is inert. The Chromium numbers above are the evidence.

## Where this came from

An engineer asked whether Badge had a max width, having hit exactly this in a hover card. The workaround in the wild is per-call-site:

```tsx
<Badge className="min-w-0 max-w-full"
       label={<span className="min-w-0 truncate" title={label}>{label}</span>} />
```

Structurally that is this fix, which is a decent signal it belongs in the component — one wrapper knows to do it, the next person putting a Badge in a table cell won't.

@freddymeta


rubyycheung (changes requested):
The containment behavior looks right: the badge stays within its available width, the icon remains visible, and labels that fit are unchanged.

However, truncating the label hides meaningful status text with no way for sighted users to recover it. The new story even says to hover for the full text, but Badge adds neither a tooltip nor a `title`, so hovering currently reveals nothing.

Please add an accessible way to expose the complete label when it is truncated—ideally a tooltip available on hover and keyboard focus, with an explicit touch-device alternative. A native `title` would be a minimum fallback, but silent truncation should not ship on its own.

cixzhang (changes requested):
Thanks, this fixes the spill. I agree with Ruby’s hold though: the status is now clipped, but the rendered badge has no `title`, tooltip, or other path for a sighted user to recover the full label. Can you add a server-safe full-label fallback for string/number labels and keep the richer only-when-truncated tooltip as its own PR if it needs measurement?

[Reviewed by Robohands]

freddymeta:
Both holds addressed, in the shape @cixzhang scoped: a server-safe fallback here, the measured tooltip as its own thing.

**What's in now** (`Badge.tsx`) — a string or number `label` is also set as the badge's `title`:

```tsx
const labelTitle =
  typeof label === 'number'
    ? String(label)
    : typeof label === 'string' && label.length > 0
      ? label
      : undefined;
```

That's the same shape `BaseTable` already uses for a truncated header cell — string content only, non-empty guard — so it's a pattern the codebase already carries rather than a new one. It costs no measurement and no hook, so `Badge` renders identically on the server and stays usable in a server component. A rich `ReactNode` label is left alone: flattening a subtree to a string renders a guess, since an icon, a `<strong>` and a nested element all read differently.

Four tests cover it: string → `title`, number → `title`, rich label → no `title`, empty string → no `title`. (I wrote a fifth for "a caller's own `title` wins" and deleted it — `BaseProps` deliberately `Omit`s `title`, so it's unreachable through the public API.)

**@rubyycheung — your catch on the story was right, and it was worse than a missing tooltip.** The `LongLabels` comment said "Hover a truncated one for the full text" while nothing supplied it. It now describes what actually happens, including what doesn't.

**What this does not close**, and I'd rather say it plainly than let `title` look like the whole answer:

- the `title` is set whether or not the label actually fits — knowing that means measuring;
- a native `title` is pointer-only. Not keyboard focus, and on touch not at all. That's the part of your review this PR doesn't reach.

Both are in #5585, along with the reason they're not here: doing it properly needs `scrollWidth > clientWidth` (or a `ResizeObserver`) plus a focusable badge, which makes `Badge` a client component and turns a non-interactive element into a focusable one. That's a behaviour change worth its own discussion, not a rider on a containment fix.

cixzhang (approved):
Thanks, this covers the server-safe string/number fallback I asked for. I’d take this and leave focus/touch/rich-label recovery to [#5585](https://github.com/facebook/astryx/issues/5585).

[Reviewed by Robohands]

freddymeta:
@rubyycheung re-requesting — your hold was the one this head answers. The full text now rides in the badge's `title` for string and number labels, which is the minimum fallback you named.

Being straight about what it does not do, since you asked for more than the minimum: `title` is pointer-only (no keyboard focus, nothing on touch), and it is set whether or not the label actually fits. Both need runtime measurement and make `Badge` a client component, so cixzhang asked for them as a separate PR — that is #5585, which also covers the focusable-badge question your "keyboard focus" ask implies.

The story comment you caught was the worse half of this and is fixed: it no longer promises something nothing supplied.

cixzhang (approved):

cixzhang (approved):

[Reviewed by Robohands]

cixzhang (approved):

[Reviewed by Robohands]

cixzhang:
/accept-visual 33045929788/1 These 70 frames add Badge coverage; no existing pixels changed, and sensor-backed review verified layout containment plus the intended full-label fallback.

freddymeta:
/accept-visual 33045929788/1 These 70 frames add Badge coverage; no existing pixels changed, and sensor-backed review verified layout containment plus the intended full-label fallback.

cixzhang:
/accept-visual 33045929788/1 These 70 frames add Badge coverage; no existing pixels changed, and sensor-backed review verified layout containment plus the intended full-label fallback.

PR #5563, opened 2026-08-26
fix(core): forward pass-through props on seven BaseProps components

### LINKED ISSUE #5254: [Bug] Ten components declare `extends BaseProps` but drop every prop they don't destructure
### Description

`ListProps` (and nine others) declare `extends BaseProps<HTMLUListElement | HTMLOListElement>`, which publicly accepts `aria-*`, `role`, `id`, `tabIndex`, event handlers and `data-*`. The implementation destructures a closed set and never spreads a rest object, so everything it doesn't name is silently dropped:

```js
export function List({
  children, density, hasDividers, header, listStyle, start,
  xstyle, className, style, 'data-testid': testId, ref
}) {                                  // <-- no ...rest
  const listElement = _jsx(Tag, {
    ref, "data-testid": testId,
    "aria-labelledby": header != null ? headerId : undefined,
    ...
```

The type checks, the prop is accepted, nothing warns, and the attribute never reaches the DOM.

The a11y consequence is concrete for `List` and `TreeList`: both derive `aria-labelledby` from `header`, so a list with no visible header has no way to get an accessible name at all. `aria-label` is the standard answer and it is typed as accepted.

Same shape, verified against the 0.4.5 tarball by parsing each component's destructuring signature:

`List`, `Markdown`, `MetadataListItem`, `NavHeadingMenu`, `Timestamp`, `Token`, `TopNavMegaMenu`, `TopNavMenu`, `TreeList`, `TypeaheadItem`

This reads as an oversight rather than a deliberate closed surface, because the neighbouring components do forward: `Card`, `Section`, `EmptyState`, `Banner` and `ChatMessageBubble` all end their destructure with `...props` / `...rest` and spread it onto the rendered element.

Happy to open a PR. Since it touches `packages/core/` and changes what reaches the DOM, tell us whether you want it through the specification protocol first, and whether you'd prefer all ten in one change or `List`/`TreeList` first on a11y grounds.

*Analysis and draft prepared with Claude Code; reviewed and verified by the contributor before filing.*

### Reproduction

```tsx
<List aria-label="Recent files">
  <ListItem>a.txt</ListItem>
</List>
```

Renders `<ul>` with no `aria-label`. TypeScript accepts the prop. Same for `role`, `id`, `tabIndex` and any `on*` handler.

Compare `<Card aria-label="…">`, which forwards correctly.

### Astryx Version

0.4.5 (verified against the published tarball; also present in 0.4.0)

### Environment

Any; not browser-specific. Surfaces via axe/screen readers on unlabelled lists.

cixzhang:
## Why

Seven remaining components declare `BaseProps` but dropped props they did not explicitly destructure. Those props type-checked yet never reached the DOM.

This is the third and final change for #5254:

1. [#5288](https://github.com/facebook/astryx/pull/5288) by @lexs landed first for `List` and `Markdown`.
2. [#5493](https://github.com/facebook/astryx/pull/5493) by @gonzoblasco landed second for `TreeList`.
3. This PR now covers only `MetadataListItem`, `NavHeadingMenu`, `Timestamp`, `Token`, `TopNavMegaMenu`, `TopNavMenu`, and `TypeaheadItem`.

Fixes #5254.

## What

- Forward neutral `aria-*`, `id`, `tabIndex`, event-handler, and `data-*` props to the intended DOM element.
- Keep styling merged through `mergeProps`, contract-owned attributes after the rest spread, and conditional attributes absent when the component does not own them.
- Compose caller handlers first with component behavior for both top-nav desktop triggers (`onClick`, `onMouseEnter`, `onMouseLeave`), both drawer triggers (`onClick`), `NavHeadingMenu` keyboard handling, and `TokenLink` mouse-up delegation.
- Keep `MetadataListItem` props on the wrapper `<div>` when stacked and the `<dt>` when inline.
- Keep the existing `TypeaheadItem` `item.element` early return as a documented hole: a caller-owned `ReactNode` is returned untouched rather than inventing `cloneElement` semantics.

## Testing

- Seven focused component suites: **241 passed**.
- Revert verification against current `origin/main`: all eight implementation files were confirmed reverted; **27 new tests failed and 214 existing tests passed**, so every added regression test detects this PR.
- `@astryxdesign/core` typecheck: passed.
- Pre-commit lint/format and repository checks: passed.

### DIFF
--- a/.changeset/baseprops-passthrough-forwarding.md
+++ b/.changeset/baseprops-passthrough-forwarding.md
@@ -1,0 +1,13 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Seven components now forward the pass-through props promised by `BaseProps`
+
+`MetadataListItem`, `NavHeadingMenu`, `Timestamp`, `Token`, `TopNavMegaMenu`, `TopNavMenu`, and `TypeaheadItem` now forward neutral `aria-*`, `id`, `tabIndex`, event-handler, and `data-*` props to their rendered DOM element. Styling still merges through `mergeProps`, contract-owned attributes retain precedence, and owned handlers compose through `composeEventHandlers` with the caller first.
+
+`MetadataListItem` targets its wrapper `<div>` when stacked and its `<dt>` when inline. A `TypeaheadItem` backed by caller-supplied `item.element` remains unchanged and does not receive forwarded props because that value may not be a cloneable element.
+
+This completes [#5254](https://github.com/facebook/astryx/issues/5254) after [#5288](https://github.com/facebook/astryx/pull/5288) by @lexs landed `List` and `Markdown` first, followed by [#5493](https://github.com/facebook/astryx/pull/5493) by @gonzoblasco for `TreeList`.
+
+@cixzhang
--- a/packages/core/src/MetadataList/MetadataList.test.tsx
+++ b/packages/core/src/MetadataList/MetadataList.test.tsx
@@ -1,7 +1,7 @@
 // Copyright (c) Meta Platforms, Inc. and affiliates.
 
-import {describe, it, expect} from 'vitest';
-import {render, screen} from '@testing-library/react';
+import {describe, it, expect, vi} from 'vitest';
+import {render, screen, fireEvent} from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import {MetadataList} from './MetadataList';
 import {MetadataListItem} from './MetadataListItem';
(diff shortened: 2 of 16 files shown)


gonzoblasco:
Solid work - the pass-through + `composeEventHandlers` pattern matches the List/TreeList work cleanly, and the revert verification is a nice touch.

One thing worth flagging on the PR analysis report: the two serious a11y violations it surfaced (color contrast, WCAG 1.4.3) on Timestamp and Token are not regressions from this change. This PR only forwards props and reorders attributes - it does not touch any color values. So whatever contrast failures exist on those two components are pre-existing in the base component/theme, not introduced here, and should not block this merge or be attributed to it. They are a separate, worthwhile audit of the theme layer (the color variants and disabled states are the first places I would look).

The Timestamp handling is the sharpest part: keeping the component's own spelled-out relative label when the caller also passes `aria-label` is the right call - "2 hours ago" reads better than a static caller string, and letting the caller clobber it would regress that.

PR #5565, opened 2026-08-26
fix(ci): stop shipping template-assets into every PR preview

cixzhang:
## Why

GitHub Pages failed after its artifact grew to 7.5 GB. The `gh-pages` branch was ~8.4 GB: 41 PR previews each carried a 125 MB copy of `sandbox/template-assets/`, accounting for ~5.1 GB.

Those copies also did not work. Portable CLI templates use `/template-assets/*`; under Pages that resolved to `facebook.github.io/template-assets/*` and returned 404.

## What

- Rewrite `/template-assets/*` only in executable sandbox bundles to `/astryx/sandbox/template-assets/*`. Portable CLI source and displayed source examples stay unchanged.
- Keep **one** asset copy in `gh-pages/sandbox/template-assets`, owned by the main deploy.
- Strip the mirrored directory from every PR artifact and prevent both preview deployment paths from copying it.
- Sweep duplicate copies already under `gh-pages/pr/*`.

Asset-file changes appear in previews after merge, when the main deploy replaces the shared copy. PRs never publish or retain their own copy.

Expected size: `gh-pages` ~8.4 GB → ~3.3 GB; Pages artifact 7.5 GB → ~2.4 GB.

## History

The sandbox mirror was added in [#2610](https://github.com/facebook/astryx/pull/2610) for local `/template-assets/*` URLs. The canonical self-hosted files moved under the docsite in [#3973](https://github.com/facebook/astryx/pull/3973). That local-dev setup was later copied wholesale into every Pages preview without accounting for the Pages base path.

## Test plan

- New Babel regression tests: shared URL rewrite, portable embedded source, local-dev no-op.
- Existing repo checks pass via commit hook.
- `actionlint`: no new findings; two existing shellcheck findings removed.
- CI builds this PR's sandbox with the shared URL and strips `out/template-assets` before upload.
- After preview deployment, verify the real PR page and image requests in Chromium.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -338,7 +338,7 @@ jobs:
       - name: Build core package
         run: pnpm build
 
-      - name: Compute sandbox preview path
+      - name: Compute sandbox preview paths
         id: urls
         run: |
           PR_NUMBER="${{ github.event.pull_request.number }}"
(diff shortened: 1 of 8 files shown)


cixzhang (commented):
Thanks—this is an approve. Verifying the 404s first makes this safe. Nit: contributors can still get stale or missing previews because ~2.4 GB exceeds the 1 GB limit.

[Reviewed by Robohands]

PR #5567, opened 2026-08-26
fix(Collapsible): refresh block examples

ernestt:
## What

Refreshes the shipped Collapsible and CollapsibleGroup block templates with complete examples for:

- controlled single-step navigation
- single and multiple accordions
- divided and standalone presentations
- component and group showcases

The controlled step example now keeps one valid step open, so closing the active row cannot produce “Step 0 of 3” or make Next wrap back to Profile.

## Scope correction

This PR no longer changes Collapsible runtime behavior, reveal motion, component tests/docs, or Storybook. Current `component:Collapsible` authority preserves reveal behavior unless separately approved, so all grid/opacity/inert motion work was removed.

## Validation

- Core build
- CLI template-doc typecheck
- root CLI suite: 222 files, 3,256 tests passed
- fixture verification
- ESLint and Prettier on changed blocks
- knowledge, sync, changesets, and diff checks

## Changeset

CLI patch only. No Core package change.


cixzhang (changes requested):
Thanks — the target-size and example updates look good, and I'd split those so they can land now.

The reveal's find-in-page path isn't live yet. `hidden="until-found"` still matches our `[hidden] {display:none}` reset, so closed panels remain unsearchable and `beforematch` never gets a chance to open them. The test dispatches that event directly, which skips the broken browser path.

Could we make `until-found` survive the reset and cover the actual browser path before landing the reveal?

[Full review](https://github.com/cixzhang/astryx/wiki/Review-5567)

[Reviewed by Robohands]

packages/core/src/Collapsible/Collapsible.tsx
```diff
@@ -299,6 +393,143 @@ export function Collapsible({
...
+  //   price of animating on the compositor instead.)
+  // - Closing applies the attribute only after the transition has finished,
+  //   or it blanks the content on the frame it lands and leaves an empty box
+  //   to shrink.
+  //
+  // transitionend is the accurate "finished" signal, but it never fires when
+  // no transition ran — a browser without `interpolate-size` snaps, and a
+  // panel inside a `display: none` ancestor never animates. The timer is the
+  // backstop, so collapsed content cannot be left in the a11y tree; whichever
+  // arrives first wins, and re-opening cancels both.
```
cixzhang: Can we clean up this huge comment

cixzhang (changes requested):
Thanks — the reset exemption fixes the selector from my last review, but the browser-find path is still blocked. The closed panel remains `inert`, and inert subtrees are excluded from find-in-page, so the browser never reaches `beforematch`; the test still dispatches that event directly.

The released `collapsible-content` target also moved from the padded content to the zero-height clipping track. An existing 16px padding override now leaves every closed row 32px tall, and the permanent clip cuts the bottom 5px from a focused last child. This timing now spans two Effects, rAF, a timer, events, and native attributes without a focused browser test.

Could we preserve those content/focus contracts and move the timing protocol into a focused unit with real find-in-page coverage?

[Full review](https://github.com/cixzhang/astryx/wiki/Review-5567)

[Reviewed by Robohands]

packages/core/src/Collapsible/Collapsible.tsx
ernestt: Gone — the whole timing protocol is deleted. The component is a pure render function now (zero effects, zero imperative DOM writes).

ernestt (commented):
Rewrote the animation to use the same `grid-template-rows: 1fr → 0fr` technique SideNavItem already ships. This resolves all the concerns:

- **`inert` vs find-in-page** — moot. There's no `hidden="until-found"` anymore, so `inert` and `beforematch` don't conflict. The component matches SideNavItem exactly: `inert` + `aria-hidden` when collapsed, grid transition handles the visual.
- **`collapsible-content` target** — moved back to the content element (inside the clip wrapper), not the zero-height track. Padding overrides land where they did before.
- **Timing protocol** — deleted entirely. No Effects, no rAF, no timer, no `transitionend` listener, no `FALLBACK_SETTLE_MS`. It's pure CSS now — toggle a style, grid interpolates.
- **Reset exemption** — reverted. No `hidden` attribute means the `[hidden]{display:none}` reset is irrelevant.

The component went from ~220 lines of imperative timing to a stateless render function with zero effects. Same pattern that's been stable in SideNavItem since launch.

ernestt:
Release-process check: .changeset/collapsible-animated-reveal.md has the correct patch bump and PR-tagged [feat] summary, but the second body line must be the contributor handle @ernestt; the handle is currently at the end of the body.

ernestt:
Addressed the latest review in 402fd5410cba73a6ac5feb08e62f6b6264968070:

- moved the `astryx-collapsible-content` theme target onto the inner padded content, keeping themed padding inside the closed clip
- kept the wrapper clipped through the opening duration, then released overflow with a pure-CSS step-end animation so the final child’s focus outline can paint
- added focused regressions for closed themed padding and open final-child focus overflow

Verified with the 40-test Collapsible suite, strict lint/repo checks, Core typechecks, Core build, and formatting.

cixzhang (changes requested):
Thanks — the chevron, keyboard, reduced-motion, target-size, theme-target, and block rewrites are otherwise in good shape.

The current Collapsible contract intentionally keeps reveal motion unchanged; this head replaces instant hide/show with grid and opacity animation plus mounted inert content. That needs a separate approved decision, so could we keep reveal behavior unchanged here?

The Controlled block also lets the active row close, then displays “Step 0 of 3”; Next returns to Profile instead of advancing. Could we either keep one step open or model the empty state explicitly?

[Reviewed by Robohands]

cixzhang (approved):
Thanks, this now keeps Collapsible behavior unchanged and the controlled block always has a valid step. The scoped CLI examples and patch release checks look good.

[Reviewed by Robohands]

PR #5576, opened 2026-08-26
Add dismiss button to canary docs banner

imdreamrunner:
## Summary

- add the Banner component's accessible dismiss control to the canary docs notice
- keep dismissal scoped to the mounted docsite shell through Banner's existing internal state

## Test plan

- `pnpm lint:strict`
- `pnpm -F @astryxdesign/docsite typecheck`
- `pnpm -F @astryxdesign/docsite test`
- `pnpm -F @astryxdesign/docsite build`
- production browser check: verified the Dismiss button renders and removes the canary banner


rubyycheung (approved):
Looks good from a design perspective. Adding a dismiss control gives users a clear way to remove the canary documentation banner when it is no longer useful.

PR #5580, opened 2026-08-26
feat(lint): guard Effect exceptions and long inline comments

cixzhang:
## Why

[PR #5567](https://github.com/facebook/astryx/pull/5567) exposed two review concerns that have narrow syntax-level forms: state-in-Effect exceptions without a recorded reason, and implementation narratives long enough to hide an abstraction boundary.

## What

- Add `@astryx/require-effect-disable-reason` for suppressions of `@eslint-react/set-state-in-effect`.
- Add `@astryx/no-long-inline-comment-block` for 20+ line comment blocks inside functions or object expressions.
- Scope both to core production source: warnings locally, errors in strict/CI mode; tests are excluded.
- Record exact legacy exceptions for the three current long blocks, tied to their opening text and source PR rationale ([#5319](https://github.com/facebook/astryx/pull/5319), [#5243](https://github.com/facebook/astryx/pull/5243), [#5201](https://github.com/facebook/astryx/pull/5201)).
- Add reasons to the two existing reasonless Lightbox suppressions.

Effect necessity, event-known state, React-derived state, and the right extraction boundary remain review decisions; these rules deliberately do not attempt data-flow analysis.

## Risk

Low. Both rules are syntax-only and the strict core baseline is clean. The long-comment allowlist matches both file and opening text, so another long block in an allowed file still fails.

## Testing

- `node internal/eslint-plugin-astryx/require-effect-disable-reason.test.mjs`
- `node internal/eslint-plugin-astryx/no-long-inline-comment-block.test.mjs`
- `CI=true eslint packages/core/src --no-cache --format stylish` (0 errors; 49 pre-existing warnings)

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -28,6 +28,24 @@ const isStrictMode = process.env.ASTRYX_STRICT_LINT === '1' || process.env.CI ==
 const astryxConfig = isStrictMode ? astryxPlugin.configs.strict : astryxPlugin.configs.recommended;
 const reactSeverity = isStrictMode ? 'error' : 'warn';
 
+const legacyLongInlineCommentAllowlist = [
+  {
+    file: 'packages/core/src/DateInput/MonthScroller.tsx',
+    startsWith: 'Put the scroller back on a pane boundary once the gesture is over.',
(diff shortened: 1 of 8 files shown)


cixzhang:
Closing by policy decision. These are useful review heuristics, but we do not want to encode both as repository lint rules without a concrete product or correctness gap requiring them.

PR #5581, opened 2026-08-26
fix(DateRangeInput): allow same-day range selection

freddymeta:
## Summary

- allow a repeated click on the range start to commit a one-day range when `minRangeSpan` is `1` (the documented default)
- preserve the existing cancel behaviour when a longer minimum span makes a one-day range invalid
- add Calendar and DateRangeInput regression coverage and clarify the prop docs

## Root cause

Calendar unconditionally treated a second click on the range anchor as cancellation. That escape hatch is needed when `minRangeSpan > 1`, but it also discarded the valid same-day completion when the minimum is `1`.

## Test plan

- `pnpm exec vitest run packages/core/src/Calendar/Calendar.test.tsx packages/core/src/DateRangeInput/DateRangeInput.test.tsx` — 137 passed
- `pnpm lint:strict` — passed (pre-existing warnings only)
- `pnpm test` — passed
- `pnpm build` — passed
- production Storybook build — passed
- Chromium smoke test against the built `Core/DateRangeInput` story with `minRangeSpan=1`: selecting 2026-08-01 twice committed `Date range: Aug 1 – Aug 1` and closed the popover

## AI assistance

Implemented and verified with an AI coding assistant; author reviewed the code and test results.

### DIFF
--- a/.changeset/daterange-same-day-selection.md
+++ b/.changeset/daterange-same-day-selection.md
@@ -5,7 +5,8 @@
 [fix] DateRangeInput: allow a same-day range when `minRangeSpan` is 1
 
 A repeated click on the range start now commits a one-day range when the
-configured minimum permits it. Longer minimum spans keep the existing cancel
-behaviour so the user can move the start date.
+configured minimum permits it, including when a maximum span is also set.
+Longer minimum spans keep the existing cancel behaviour so the user can move
+the start date.
 
 @freddymeta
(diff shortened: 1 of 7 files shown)


cixzhang (approved):
Verified same-day completion for default and explicit minimums, preserved toggle-off for longer minimums, and covered combined max bounds. Tests and Chromium matrix pass.

[Reviewed by Robohands]

PR #5586, opened 2026-08-27
fix(Carousel): mirror single-edge fade gradients under RTL

### LINKED ISSUE #5622: Carousel: single-edge fade gradient paints on the wrong side in RTL
## Description

In an RTL document, Carousel's edge-fade mask paints on the wrong physical side whenever only one edge has overflow. The fade is supposed to signal that more items exist off-screen, but in RTL it fades the edge where nothing is hidden and leaves the edge that actually clips content with a hard cut-off.

## Reproduction

1. Render a `Carousel` with enough items to overflow inside a `dir="rtl"` container (e.g. the Storybook RTL toolbar toggle).
2. Scroll the carousel fully toward either end so only one edge overflows.
3. Observe the fade gradient on the physical left edge while content is clipped on the right (or vice versa).

RTL carousel scrolled to one end: the left edge has a fade over nothing while the right edge cuts off an item with no fade

## Cause

`useScrollOverflow` reports **logical** edges ("start = left in LTR, right in RTL", per the hook's own docs). But the `fadeStart` / `fadeEnd` styles selected from those flags use hardcoded physical gradient directions (`to right` / `to left`) in their `maskImage`. Under RTL the two single-edge states are therefore inverted. Only the symmetric `fadeBoth` state happens to look correct.

The rest of the component already mirrors physical values for RTL (button-pill transforms via `:is([dir="rtl"] *)`, chevron mirroring, sign-flipped scroll deltas via `isRtlElement`); the fade gradients are the one spot that doesn't.

## Expected behavior

The fade should always sit on the physical edge that has hidden content, in both LTR and RTL.

mattandryc:
Fixes #5622

## Summary

Carousel's edge-fade masked paint on the wrong physical side in RTL.

`overflowStart`/`overflowEnd` from `useScrollOverflow` are **logical** edges ("start = left in LTR, right in RTL", per the hook's own docs), but the `fadeStart`/`fadeEnd` styles they select use hardcoded physical gradients (`to right` / `to left`). 

In an RTL document both one-sided states are inverted: scroll the carousel fully toward the content end and `fadeStart` fades the physical **left** edge, which hides nothing, while the right edge, which actually hides content, gets no fade. Only the symmetric `fadeBoth` state happens to look right. This defeats the fade's stated purpose ("signalling that more items exist off-screen").

The rest of the component already mirrors physical values for RTL (button-pill transforms via `:is([dir="rtl"] *)`, chevron mirroring, sign-flipped scroll deltas via `isRtlElement`). The gradients were the one oversight.

## Fix

| Before | After |
|---|---|
| Screenshot 2026-08-26 at 9 42 09 PM | Screenshot 2026-08-26 at 9 42 24 PM |

Swap each single-edge gradient's direction under `:is([dir="rtl"] *)`, matching the button-pill pattern already used in the same file. `fadeBoth` is symmetric and unchanged.

## Tests

- Two LTR regression tests: start-only overflow fades the physical left edge, end-only overflow fades the physical right edge (via `getComputedStyle().maskImage`).
- An RTL test asserting both mirrored gradient rules exist under `[dir="rtl"]`, scanning the injected StyleX CSS (same approach as ProgressBar's RTL keyframe test). Verified it fails without the fix.

### DIFF
--- a/.changeset/carousel-rtl-edge-fade.md
+++ b/.changeset/carousel-rtl-edge-fade.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Carousel: mirror the single-edge fade gradients under RTL so the mask fades the physical edge that actually hides content (overflowStart/overflowEnd are logical edges, the gradients were always physical left/right)
+
+@mattandryc
--- a/packages/core/src/Carousel/Carousel.test.tsx
+++ b/packages/core/src/Carousel/Carousel.test.tsx
@@ -850,4 +850,87 @@ describe('Carousel', () => {
       expect(scroller).not.toHaveAttribute('data-padding');
     });
   });
+
+  describe('edge fade', () => {
+    function getScroller() {
+      const region = screen.getByRole('region');
+      return region.firstElementChild as HTMLElement;
+    }
+
+    function makeOverflowing(el: HTMLElement, scrollLeft: number) {
+      Object.defineProperty(el, 'scrollWidth', {
+        value: 500,
+        configurable: true,
+      });
+      Object.defineProperty(el, 'clientWidth', {
+        value: 200,
+        configurable: true,
+      });
+      Object.defineProperty(el, 'scrollLeft', {
+        value: scrollLeft,
+        writable: true,
+        configurable: true,
+      });
+      fireEvent.scroll(el);
+    }
+
+    function injectedCss(): string {
+      let out = '';
+      for (const sheet of Array.from(document.styleSheets)) {
+        try {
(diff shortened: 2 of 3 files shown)


rubyycheung (approved):
Looks good from design. In RTL, the single-edge fade now appears on the physical edge where content is actually clipped, restoring the intended overflow cue. The before/after evidence is clear, and the focused LTR and RTL regression coverage supports this implementation.

cixzhang (approved):
Thanks, this is good. The rebased change fixes the normal RTL single-edge fade and keeps focused LTR/RTL regression coverage alongside main’s newer Carousel tests.

[Reviewed by Robohands]

PR #5591, opened 2026-08-27
fix(ci): recognize repository maintainers for visual acceptance

cixzhang:
## Why

The visual-acceptance command rejects a repository owner as “not a maintainer.” GitHub returns the actor’s coarse permission as `write` even though `user.permissions.maintain` is true, so checking only the coarse string discards an authorized decision.

## What

Normalize the explicit `admin` and `maintain` booleans before falling back to the coarse permission string. The acceptance record still stores only `admin` or `maintain` and keeps the existing authorization threshold.

## Testing

- `pnpm exec vitest run .github/scripts/visual-gate/workflow-concurrency.test.mjs`
- Reproduced live response: coarse `write`, explicit `maintain: true`.

### DIFF
--- a/.github/scripts/visual-gate/workflow-concurrency.test.mjs
+++ b/.github/scripts/visual-gate/workflow-concurrency.test.mjs
@@ -70,6 +70,18 @@ describe('visual acceptance workflow concurrency', () => {
     expect(accept).toContain('pull-requests: write');
   });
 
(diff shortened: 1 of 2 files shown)


cixzhang (commented):
Thanks—the live `write` plus `maintain: true` response now authorizes correctly, while the maintain/admin threshold stays unchanged. The exact-head test and actionlint pass. Good to merge.

[Reviewed by Robohands]

PR #5588, opened 2026-08-27
fix(ci): restore visual acceptance publication

cixzhang:
## Why

The required `visual-acceptance` gate can still strand stable-visual PRs after [#5587](https://github.com/facebook/astryx/pull/5587) fixed its source run/attempt identity and granted PR mutation permission:

- The initializer still deletes `visual-approved` itself. That mixes the lightweight head-status initializer with advisory PR mutation and requires broader credentials than classification needs.
- Trusted recapture still attempts to replace reserved `GITHUB_SHA`. GitHub keeps that value bound to the publisher workflow, so immutable evidence can record the publisher commit instead of the captured PR head.

[#5162](https://github.com/facebook/astryx/pull/5162) exposed the initializer failure after [#5578](https://github.com/facebook/astryx/pull/5578) made the context required.

## What

- Reuses #5587's `ASTRYX_VISUAL_RUN_ID` / `ASTRYX_VISUAL_RUN_ATTEMPT` implementation and extends it with `ASTRYX_VISUAL_SHA` rather than duplicating the identity helper.
- Keeps the initializer limited to read access plus commit-status writes.
- Moves head/rerun label invalidation to the existing trusted `workflow_run` publisher and gives only mutating jobs `pull-requests: write`.
- Preserves success for no-scope and clean captures without requiring an acceptance record.

The immutable gh-pages acceptance record remains authoritative. Statuses, comments, and `visual-approved` remain derived projections. Fork code is downloaded only as a static Storybook artifact and never receives write credentials.

## Verification

- Full visual-gate suite: 13 files, 118 tests
- `actionlint` on all three visual-acceptance workflows (with the repository's custom runner label allowed)
- `pnpm check:repo`
- Exact [#5162](https://github.com/facebook/astryx/pull/5162) artifact dry-run: trusted publication derives 18 shots with 16 additions and no failures; state is correctly `pending:decision`, then an isolated immutable acceptance for PR/head/run `5162` / `66ef14d69ed5f32c469c78cb2e8395bfe23ec633` / `33041576630/1` produces `success:accepted` with 16 archived AFTER hashes.

The live #5162 capture is not a zero-diff capture: it still needs explicit visual acceptance after this process fix lands and its trusted publisher is rerun.

PR #5592, opened 2026-08-27
fix(ci): authorize Repo Owners for visual acceptance

cixzhang:
## Why

GitHub's collaborator-permission endpoint reports repository owners as top-level `permission: write` with `role_name: Repo Owner`, while effective capabilities live under `user.permissions`. Reading `user.permission` therefore refused every decision, and treating top-level `write` as authority would admit ordinary writers.

[#5591](https://github.com/facebook/astryx/pull/5591) restored effective `maintain`/`admin` authorization. This follow-up preserves the endpoint's permission and role, adds the exact repository-owner fallback, and keeps old acceptance records valid.

A live authorized replay then exposed a second blocker: the archive job has no package install, but `visual-acceptance.mjs` eagerly imported `pngjs` before running the dependency-free `accept` path.

## What

- Parse the documented endpoint shape once: effective `maintain`/`admin` or exact `Repo Owner` succeeds; ordinary `write`, unknown, and 404 identities fail.
- Archive both the reported permission and role without consulting `author_association`.
- Keep old `maintain`/`admin` acceptance records valid.
- Load `pngjs` only for post-merge pixel comparison, so the read-only authorization and archive path need no dependency install.
- Preserve the trusted default-branch checkout and publisher boundary from [#5588](https://github.com/facebook/astryx/pull/5588).

## Risk

Low. Authorization remains fail-closed and does not admit general write access. The lazy dependency load is covered by running acceptance from an isolated copy with no `node_modules`.

## Testing

- Full visual-gate Vitest suite: 14 files, 135 tests
- `pnpm check:repo`
- Exact [#5162](https://github.com/facebook/astryx/pull/5162) artifact dry-run against head `66ef14d69ed5f32c469c78cb2e8395bfe23ec633`, run `33041576630/2`
- Full CI on the exact PR head

### DIFF
--- a/.github/scripts/visual-gate/authorization.mjs
+++ b/.github/scripts/visual-gate/authorization.mjs
@@ -1,0 +1,29 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+const MAINTAINER_PERMISSIONS = new Set(['maintain', 'admin']);
+const REPO_OWNER_ROLE = 'Repo Owner';
+
+export function visualAcceptanceIdentity(permissionLevel = {}) {
+  const permission = permissionLevel.permission ?? 'none';
+  const effectivePermission = permissionLevel.user?.permissions?.admin
+    ? 'admin'
+    : permissionLevel.user?.permissions?.maintain
+      ? 'maintain'
+      : permission;
+  return {
+    permission,
+    roleName: permissionLevel.role_name ?? null,
+    effectivePermission,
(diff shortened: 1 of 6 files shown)


cixzhang (commented):
Can't approve my own PR, so: semantic approve. The Repo Owner boundary, dependency-free acceptance, record compatibility, and workflow permissions all pass.

[Reviewed by Robohands]

PR #5593, opened 2026-08-27
fix(ci): install visual acceptance dependencies

cixzhang:
## Why

The signed visual-acceptance command now authorizes correctly, but its archive job fails before writing the immutable record because `visual-acceptance.mjs` imports `pngjs` and the job never installs repository dependencies.

## What

Run the repository's standard Node and pnpm setup after the trusted checkout and before the archive script. A workflow regression test pins that ordering.

## Testing

- `pnpm exec vitest run .github/scripts/visual-gate/workflow-concurrency.test.mjs .github/scripts/visual-gate/visual-acceptance.test.mjs` — 24 tests
- Live failure reproduced as `ERR_MODULE_NOT_FOUND: pngjs` in the acceptance archive step.

### DIFF
--- a/.github/scripts/visual-gate/workflow-concurrency.test.mjs
+++ b/.github/scripts/visual-gate/workflow-concurrency.test.mjs
@@ -121,6 +121,18 @@ describe('visual acceptance workflow concurrency', () => {
     expect(authorize).toContain("? 'maintain'");
   });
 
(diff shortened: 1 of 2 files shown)


cixzhang (commented):
Thanks, this is good. The standard setup fixes the missing dependency, and the workflow test pins the ordering.

[Reviewed by Robohands]

PR #5595, opened 2026-08-27
fix(ci): retry visual promotion after preview cleanup

cixzhang:
## Why

Post-merge promotion can reach verification after cleanup has deleted the PR preview evidence—either before the first attempt or while retrying a rejected `gh-pages` push. The immutable acceptance remains valid, but the ephemeral-state check rejects it and leaves the approved baseline stale.

Promotion also copied the merged recapture into the baseline after checking only decoded-pixel equality. A differently encoded PNG could therefore become the durable baseline even though its file hash was never reviewed.
## What

- Authorize each promotion retry from the durable `current.json` pointer and immutable acceptance record after binding PR, head, run, and attempt identity.
- Keep the latest completed CI check before promotion.
- Canonicalize trusted publication and merged recapture through one pinned PNG encoder, require the canonical SHA-256 to match the reviewed AFTER, and promote those canonical bytes; removed shots stay hashless.
- Remove only the promotion-time dependency on ephemeral PR preview evidence.
- Add an executable git race harness that makes cleanup win the first push, retries from a fresh clone without `pr/<number>`, and proves the baseline promotion succeeds.
- Prove stale pointers, superseded CI, wrong head/run identity, tampered reviewed pixels, and changed baseline preimages still fail closed.

## Risk

Low and fail-closed. Merged pixels are still recaptured and compared byte-for-byte with archived AFTER images; immutable-record validation, baseline preimage checks, and trusted workflow boundaries are unchanged.

## Testing

- Full visual-gate suite: 165 tests
- Exact rejected-push → preview-cleanup → fresh-clone retry harness
- `actionlint` (repository custom runner label allowed)
- `pnpm check:repo`
- Full CI on the exact head

Do not merge automatically. Visual-governance ownership must merge this process change.

### DIFF
--- a/.github/scripts/visual-gate/lib/canonical-png.mjs
+++ b/.github/scripts/visual-gate/lib/canonical-png.mjs
@@ -1,0 +1,29 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+import {createRequire} from 'node:module';
+
+const require = createRequire(import.meta.url);
+
+export const CANONICAL_PNG_OPTIONS = Object.freeze({
+  bitDepth: 8,
+  colorType: 6,
+  deflateChunkSize: 32 * 1024,
+  deflateLevel: 9,
+  deflateStrategy: 3,
+  filterType: -1,
+  inputColorType: 6,
+  inputHasAlpha: true,
+});
+
+export function canonicalizePng(bytes) {
+  const {PNG} = require('pngjs');
+  const image = PNG.sync.read(bytes, {checkCRC: true});
+  return {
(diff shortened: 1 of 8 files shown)


cixzhang (commented):
Reviewed exact head `081cd498eb40b6b68d7ca8a4fb46caa60d3eb0e2`. Canonicalising both sides is right, and the production-shaped capture promoted in the fresh loop. Non-blocking: `gate.mjs accept` still writes raw bytes to the manual baseline path.

Full record: https://github.com/cixzhang/astryx/wiki/Review-5595

[Reviewed by Robohands]

PR #5601, opened 2026-08-27
fix(Carousel): keep focus inside the carousel at a scroll edge, and expose the scroll container to themes

cixzhang:
> **Ruled and rebuilt.** The review loop blocked three times, each on the same parent: the focus hand-off was driven by something other than the thing that disables the button. Predict the landing (wrong under Reduce Motion), predict it better (wrong under mandatory scroll-snap), wait for `scrollend` (right everywhere it fires, 1.03s late where it does not). The remedy needed a state-transition watcher, which means an Effect, and moving focus from an Effect is a house rule — so it went to Cindy as a ruling rather than a correction.
>
> She ruled the Effect acceptable here, narrowly, on stated conditions. It is implemented at `a64ea2821f7` and reviewed again below.

Night Watch component audit of `core/Carousel`, rubric v1.13. First audit of this component.

**Before: 69.2 / D (3 open BLOCKs) → After: 83.7 / C (1 open BLOCK).** The grade is capped at C by the one BLOCK left open, not by the arithmetic; close it and nothing else and the projection is 85.2 / B.

Ledger row: https://github.com/facebook/astryx/wiki/_compare/36b52ac3f6d4b2792761811e26e23a3ee6d04279

## What this fixes

**A5 — keyboard focus is dropped at a scroll edge.** The trailing nav button is disabled the moment its direction runs out. The browser drops focus off a control it disables, so a keyboard user who pressed Next on the last slide landed on `<body>` and had to Tab from the top of the document. Measured in Chromium: focus "Scroll right" on the Default story, press Enter once, `document.activeElement` is `BODY`. The `hasLoop` story never disables its buttons and never drops focus, which isolates the cause.

The transition that disables the button is the thing to watch, so an Effect watches it. It acts only when overflow newly disables the nav button that held focus, moves focus and nothing else, and cannot fire twice for one transition: it clears its tracker before focusing, so a re-render or a StrictMode double invoke returns at the first guard.

Running after the commit is also what makes the opposite button a valid receiver, which it never was during the press: at that moment it is still disabled, and `focus()` on a disabled control does nothing. So focus lands on a real control carrying the shared 2px accent ring, and never passes through the body at all. The scroll container is the fallback for the one case with no opposite control left, when content shrinks until the carousel stops overflowing.

Three earlier attempts predicted the landing position from the press instead, and the loop proved each wrong in turn: after the scroll (Reduce Motion lands it instantly), before the scroll plus the step (mandatory snap lands it further), and the settled position from `scrollend` (1.03s late on a browser that never fires it). Watching the transition also makes the opposite arrow a valid receiver, which it never is during the press: at that instant it is still disabled, and `focus()` on a disabled control does nothing. The scroll container is the fallback for the one case with no opposite arrow left, when content shrinks until the carousel stops overflowing.

Swept in Chromium: 10 stories x normal, reduced motion and RTL, plus the leading direction seeded to the far edge. Focus lands on the opposite arrow within 2ms of the button disabling, and never passes through the body at all. The buttons' own focus and blur handlers track who is on them, and the blur ignores the one the commit itself causes, so a tracker cannot outlive the person and pull their focus back on a later edge; that case is driven by wheel, by a real touch drag and by a content reflow, in both directions, against a main control. Regression tests included; the jsdom mock takes a landing function, so a snap that overshoots what the press asked for is modelled, and that case fails under either prediction.

**T6/T7 — no style-driving prop reaches a theme target.** `gap`, `padding`, `hasSnap` and `hasEdgeFade` all select StyleX style objects applied to the scroll container, and that element carried no target at all; `themeProps('carousel')` sits on the outer positioning div, which paints nothing, and passed no props. A theme could select `.astryx-carousel` and reach none of the component's spacing, inline padding or edge fade. The scroll container now carries `astryx-carousel-scroller` with those four props, matching what the sibling with the same anatomy, `TabList`, already exposes as `astryx-tab-strip`.

System-level finding filed as one issue rather than four: https://github.com/facebook/astryx/issues/5602 . The same edge-nav shape drops focus in Pagination, Calendar and Lightbox, driven and confirmed in Chromium; no single component's PR closes it.

**A18** — the eight `Carousel::*::color-contrast` baseline entries no longer occur; `a11y:audit` reports them resolved, so they are deleted.

**D12** — the scroll container declared `transition-property: mask-image`. None of the mask values it moves between can interpolate: `none` to a gradient is discrete, and the four gradients differ in direction or stop count. Measured with `getAnimations()` across every transition the component can make, and no animation ever runs. Dead declaration, removed.

**Docs and stories** — `padding` had no story, and there was no narrow-container, long-text or single-item story, so the a11y and RTL sweeps could not see those states. The story file's own `argTypes` said `hasButtons` means "Show navigation buttons on hover", which the component does not do. The implemented APG carousel pattern is now named in the docs rather than only in a source header.

## Visual evidence

`before/` is `origin/main`'s `Carousel.tsx` built in the same worktree with the same pipeline as `after/`, so the pair differs only by this diff. Every frame carries a sensor receipt (build SHA, story id, theme, colour mode, direction, viewport, semantic state, geometry, settled render, zero page errors) as a `.sensors.json` beside it. Light and dark were verified byte-different before use.

**25 of 25 comparable frames are byte-identical.** Nothing here changes a rendered pixel in any state main can also reach.

The one frame with no `before` twin is the fix itself. On main that capture cannot be taken: the sensor reads focus as `BODY`, so the state does not exist.

After pressing Next on the last slide, the carousel has scrolled and focus is on the opposite arrow, carrying the shared accent ring:

After an edge press

The scroll container, unchanged in appearance from main:

| Before | After |
|---|---|
| Before | After |

All 26 frames, both sets, with receipts: https://github.com/cixzhang/astryx/tree/assets/pr-5601

## Needs Review

Not fixed here, and each left open in the ledger row.

- **I10 — the nav buttons announce a fixed physical direction.** `@astryx.carousel.scrollLeft` / `.scrollRight` are hard-wired to the two buttons, so their accessible names never follow direction. Under `dir="rtl"` the visible pill sits on the left, its chevron points left and it scrolls left, and it is announced "Scroll right". Both shipped RTL locales carry the physical wording (`ar-SA`, `he-IL`; 2 of 30 locales), and the catalog `description` asks translators to swap them, which neither has. The remedy is direction-neutral keys (previous / next), which retires two shipped catalog keys and changes user-visible text in 30 locales. That is a decision, not a correction.
- **A15 — the scroll container's focus ring is the browser default**, `1px auto` Chromium blue, not `--color-accent`, so no theme can restyle it and it does not follow colour mode. Pre-existing; this PR routes focus there, which makes it easier to hit. A remedy was tried and backed out: the shared ring sits 3px outside its element and the root clips to a 1px bleed, so it has to be inset, and inset it paints under the slides and reads fainter than the browser default it replaced. Left open rather than closed with something worse.
- **The button pill is unthemeable.** Its background, radius and shadow are on a wrapper with no target. `TabList` exposes the equivalent as `astryx-tab-scroll-button`. Adding it is additive theme surface that no blocking finding forces, so it is your call rather than the audit's.
- **The edge fade is `--spacing-1` wide, about 4px**, which is close to invisible at rest and undercuts the affordance the docs describe. Visual taste.
- **`useLayer` has no declarative initial-open option**, so Carousel re-implements one in a mount Effect. Six of the seven sanctioned-exception conditions hold; it fails only "lives in the primitive that owns the system". Same shape as the Tooltip `isDefaultOpen` case the rubric already worked. Routes to `useLayer`.
- **The pressed overlay never paints** while the pointer is also hovering. Library-wide root, already tracked in #5451 / #5516, not scored against Carousel.

## Testing

`pnpm vitest run packages/core/src/Carousel` (37 cases; 9 new or rewritten), `a11y:audit --components Carousel` (0 violations), `rtl:audit --filter Carousel` (measured, 1 auto pass, no N/A rollup), `check:sync`, `check:changesets`, `check:i18n-catalog`, `typecheck:docs`, docsite `generate` + `test`. Frames captured in real Chromium with sensor receipts, light and dark verified byte-different before use.

Night Watch - Component Auditor

### DIFF
--- a/.changeset/carousel-edge-focus-and-theme-target.md
+++ b/.changeset/carousel-edge-focus-and-theme-target.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Carousel keeps keyboard focus inside the component when a nav button press runs the content to an edge and disables that button, instead of letting focus fall to the document body. The scroll container is also now a documented theme target, `astryx-carousel-scroller`, carrying the gap, padding, snap and edge-fade props it styles, so a theme can reach the spacing and the fade it could not see before.
+
+@cixzhang
--- a/.changeset/carousel-edge-focus-and-theme-target.md
+++ b/.changeset/carousel-edge-focus-and-theme-target.md
@@ -2,6 +2,6 @@
 '@astryxdesign/core': patch
 ---
 
-[fix] Carousel keeps keyboard focus inside the component when a nav button press runs the content to an edge and disables that button, instead of letting focus fall to the document body. The scroll container is also now a documented theme target, `astryx-carousel-scroller`, carrying the gap, padding, snap and edge-fade props it styles, so a theme can reach the spacing and the fade it could not see before.
+[fix] Carousel no longer drops keyboard focus to the document body when reaching a scroll edge disables the nav button in use. Focus moves to the opposite arrow instead, on the state transition that disables the button rather than on a prediction from the press, so it holds under reduced motion, under scroll-snap, and on browsers without `scrollend`. The scroll container is also now a documented theme target, `astryx-carousel-scroller`, carrying the gap, padding, snap and edge-fade props it styles, so a theme can reach the spacing and the fade it could not see before.
 
 @cixzhang
--- a/.github/a11y-baseline.json
+++ b/.github/a11y-baseline.json
@@ -14,46 +14,6 @@
       "helpUrl": "https://dequeuniversity.com/rules/axe/4.12/color-contrast?application=playwright"
     },
     {
-      "key": "Carousel::Card Content::color-contrast",
-      "impact": "serious",
-      "helpUrl": "https://dequeuniversity.com/rules/axe/4.12/color-contrast?application=playwright"
-    },
-    {
-      "key": "Carousel::Custom Content (Swatches)::color-contrast",
-      "impact": "serious",
-      "helpUrl": "https://dequeuniversity.com/rules/axe/4.12/color-contrast?application=playwright"
-    },
-    {
-      "key": "Carousel::Default::color-contrast",
-      "impact": "serious",
-      "helpUrl": "https://dequeuniversity.com/rules/axe/4.12/color-contrast?application=playwright"
-    },
-    {
-      "key": "Carousel::Few Items (No Overflow)::color-contrast",
-      "impact": "serious",
-      "helpUrl": "https://dequeuniversity.com/rules/axe/4.12/color-contrast?application=playwright"
-    },
-    {
-      "key": "Carousel::Large Gap::color-contrast",
-      "impact": "serious",
-      "helpUrl": "https://dequeuniversity.com/rules/axe/4.12/color-contrast?application=playwright"
-    },
-    {
-      "key": "Carousel::Scroll Snap::color-contrast",
-      "impact": "serious",
-      "helpUrl": "https://dequeuniversity.com/rules/axe/4.12/color-contrast?application=playwright"
-    },
-    {
-      "key": "Carousel::Thumbnails with Remove::color-contrast",
-      "impact": "serious",
-      "helpUrl": "https://dequeuniversity.com/rules/axe/4.12/color-contrast?application=playwright"
-    },
-    {
-      "key": "Carousel::Without Buttons::color-contrast",
-      "impact": "serious",
-      "helpUrl": "https://dequeuniversity.com/rules/axe/4.12/color-contrast?application=playwright"
-    },
-    {
       "key": "Chat::Message Status::color-contrast",
       "impact": "serious",
       "helpUrl": "https://dequeuniversity.com/rules/axe/4.12/color-contrast?application=playwright"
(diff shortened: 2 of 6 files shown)


cixzhang:
Thanks, the diagnosis is right and the scroll container is the correct receiver.

Fix one thing first. With Reduce Motion on, `scrollBy` lands instantly, so `Carousel.tsx:460` reads a position that has already moved and counts the step twice. Someone scrolling with Reduce Motion on then loses the Next button one press before the end, while it is still enabled and still the control they were using. Read the position before the `scrollBy` above it; I built that and drove it, and the true edge still hands off. Your own test catches it: change `makeScrollable(getScroller(), 2000)` to `600` at `Carousel.test.tsx:565`.

Two smaller ones: the comments at `:455` and `:641` still say focus goes to the opposite button, and the baseline regen re-escaped ten unrelated Layout/Stepper keys.

Press 2 of 3 with Reduce Motion on, `core-carousel--color-swatches` at 620x300, everything matched except the build:

| main | this PR |
|---|---|
| main | head |

Same pill, same enabled state, and the ring has left it.

[Reviewed by Robohands]

cixzhang:
Thanks, the reduced-motion press is fixed, the two comments read right now, and the baseline is back to just the eight deletions.

One more to fix, and it is the same shape. `Card Content` sets `hasSnap`, and mandatory snapping lands the container somewhere the press did not ask for: `Carousel.tsx:464` predicts 420px of a 500px run, Chromium goes to 500, the trailing button disables unpredicted, and focus falls to `<body>`. Same in the leading direction, and in both motion modes. Thirteen of the fourteen shipped stories hand off correctly; `core-carousel--cards` is the one that does not.

Decide the focus hand-off from where the container actually ended up rather than from where the press aimed it. `overflowEnd`/`overflowStart` already know. I have not built that, so it may cost more than I think. Ideally every shipped story keeps focus inside the carousel, `cards` included.

| snapping, as shipped | snapping off, same press |
|---|---|
| no focus ring anywhere | ring on the still-enabled pill |

[Reviewed by Robohands]

cixzhang:
Thanks, this closes it on every shipped story. `cards`, both directions, both motion modes, RTL: focus lands on the scroll container, and round 1's reduced-motion press stays fixed.

One thing left, and it is the timeout rather than the listener. The hand-off waits for `scrollend`, so where that never arrives the 1200ms at `Carousel.tsx:163` is how long focus sits on `<body>`, measured 1.03s, with no focus indicator anywhere on screen for all of it. `useScrollSettle.ts` calls that browser one we target. The button's own disable does not wait for `scrollend`: in the same run it fired at 194ms, off `overflowEnd`.

Decide the hand-off from that rather than from an event the browser may not send. I have not built it, so it may cost more than I think. Ideally focus is inside the carousel within a frame of the button disabling, on a browser that never fires `scrollend`.

| no `scrollend`, 600ms after the press | `scrollend`, same moment |
|---|---|
| no focus indicator anywhere | ring on the scroll container |

[Reviewed by Robohands]

cixzhang:
Thanks, this is the right trigger and it lands. Every story that can reach an edge, both directions, RTL, reduced motion, snap: focus is on the opposite arrow within 2ms.

One left, and it is who rather than when. The tracker at `:657`/`:686` is only cleared by a hand-off, so it still names the person after they leave, and the guard at `Carousel.tsx:531` accepts `<body>`. Any later edge, a swipe or a reflow with no press at all, then pulls focus back in. Driven both directions on `color-swatches`: focus an arrow, click away, swipe to that edge, and the ring lands on the other arrow. `main` leaves them where they were.

`!event.currentTarget.disabled` separates the disable's own blur from the person's; I built and drove that, and all seven hand-off arms are unchanged. The body and its frame still describe the scroll container as the receiver. Ideally the tracker clears on the person's blur.

| the same three actions on `main` | on this PR |
|---|---|
| no focus indicator anywhere | ring on Scroll left |

[Reviewed by Robohands]

cixzhang:
Thanks, this closes it. Wheel, touch, reflow, RTL, snap, reduced motion: focus stays where the person left it.

[Reviewed by Robohands]

cixzhang:
/accept-visual 33078522258/1 Carousel is being baselined for the first time: 194 added, 0 removed, 0 failed, and the report says no shot changed against the baseline. This is 14 stories across 7 themes in light/dark; 56 additions are the four new stories, the other 138 are pre-existing unbaselined stories.

cixzhang:
/accept-visual 33078522258/1 First baselining for Carousel stories: 194 added, 0 removed, 0 failed, and no existing shot changed against baseline.

PR #5604, opened 2026-08-27
fix(ci): defer broad trusted visual acceptance

cixzhang:
## Why

Accepted broad component changes can spend about 30 minutes recapturing more than 4,000 frames in the trusted comment workflow and still fail the required `visual-acceptance` status. [PR #5516](https://github.com/facebook/astryx/pull/5516) hit this even though the 240-shot PR policy explicitly delegates broad scope to the daily release gate.

## What

- Emit validated, immutable skipped evidence when trusted scope is broad, bound to the exact PR, head, run, and attempt.
- Publish that evidence through the existing `gh-pages` path and resolve it as successful/deferred.
- Skip Storybook download and browser capture only for broad scope; narrow component and theme recapture is unchanged.
- Render skipped evidence as deferred without an `/accept-visual` command.

## Risk

Workflow-only. Broad scopes no longer get per-PR recapture and remain covered by the daily release gate. Narrow scopes keep the existing capture and acceptance path.

## Testing

- 195 focused visual-gate tests
- `pnpm lint` (passes; existing warnings only)
- `actionlint` on the changed workflow (repository runner label allowlisted)
- Prettier check on changed formatted workflow/test files
- `git diff --check`

### DIFF
--- a/.github/scripts/lib/visual-format.test.mjs
+++ b/.github/scripts/lib/visual-format.test.mjs
@@ -61,6 +61,7 @@ describe('buildVisualSection', () => {
     expect(section).toContain('Skipped');
     expect(section).toContain('900 shots exceeds the 240-shot budget');
     expect(section).toContain('View the report');
+    expect(section).not.toContain('/accept-visual');
   });
 
   it('renders verdict strings as one-line literal text', () => {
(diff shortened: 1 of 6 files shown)


cixzhang (commented):
Thanks — this is clean. Broad scope now defers with exact immutable evidence, while narrow capture stays byte-identical and release-gate baseline ownership is untouched.

[Reviewed by Robohands]

PR #5606, opened 2026-08-27
fix(TreeList): respect consumer onKeyDown preventDefault cancellation

### LINKED ISSUE #5584: [Bug] TreeList consumer onKeyDown cannot cancel built-in navigation
## Impact

A builder who uses TreeList's accepted `onKeyDown` prop to reserve a key cannot cancel the built-in tree behavior. Calling `preventDefault()` for ArrowDown still moves focus to the next tree item before the consumer handler runs, so a keyboard user gets navigation the app explicitly tried to suppress. The handler type-checks and fires; nothing warns that cancellation is too late.

## Current main

At [`cddf57e5`](https://github.com/facebook/astryx/blob/cddf57e5d474281150bef92f1167228ab0a6e4a5/packages/core/src/TreeList/TreeList.tsx#L333-L355), `...restProps` puts the consumer handler on the outer `<div>`, while `handleKeyDown` runs on the inner `ul[role="tree"]`. The event reaches the inner handler first and bubbles outward only after focus and roving tabindex have changed.

## Reproduction

```tsx
it('lets a consumer prevent built-in TreeList keyboard navigation', () => {
  render(
    <TreeList
      items={[
        {id: 'one', label: 'One'},
        {id: 'two', label: 'Two'},
      ]}
      onKeyDown={event => event.preventDefault()}
    />,
  );

  const items = screen.getAllByRole('treeitem');
  items[0].focus();
  fireEvent.keyDown(items[0], {key: 'ArrowDown'});

  expect(items[0]).toHaveFocus();
  expect(items[0]).toHaveAttribute('tabindex', '0');
  expect(items[1]).toHaveAttribute('tabindex', '-1');
});
```

On current main this fails: focus lands on `Two`, with `tabindex="0"`; `One` becomes `tabindex="-1"`.

## Acceptance

- The consumer `onKeyDown` runs before TreeList's navigation/toggle behavior.
- `preventDefault()` leaves focus and the roving tab stop unchanged.
- A non-canceling consumer handler still fires while the APG keyboard model works.
- Add the focused regression test above.

Related: #3343 tracks the broader composite-navigation program; this issue is scoped to the already-public TreeList handler contract.

Geervan:
Fixes #5584

A builder using `TreeList`'s accepted `onKeyDown` prop to intercept or reserve specific keys (such as `ArrowDown`, `ArrowUp`, or `Enter`) could not cancel the built-in WAI-ARIA tree keyboard navigation.

Calling `event.preventDefault()` in `onKeyDown` executed and type-checked without warning, but focus and roving `tabindex` moved to the adjacent tree item anyway. This broke the standard React component event contract and prevented applications from implementing custom keyboard shortcuts or drag-and-drop reordering without triggering unwanted list navigation.

## Root Cause

In `TreeList.tsx`, `...restProps` placed consumer attributes (including `onKeyDown`) onto the outer `<div class="astryx-tree-list">` wrapper, while internal APG keyboard navigation (`handleKeyDown` from `useTreeFocus`) was attached to the inner `<ul role="tree">`.

Because DOM keyboard events originate on `<li role="treeitem">` elements inside the `<ul>`, event bubbling delivered the event to `<ul role="tree">` first. `useTreeFocus` updated `tabindex` and called `.focus()` on the target item before the event bubbled up to the outer `<div>`. By the time the consumer's `onKeyDown` executed `event.preventDefault()`, focus had already jumped.

## Solution

1. **Extract & Compose `onKeyDown`**: Destructured `onKeyDown` from `TreeListProps` and composed it directly on `<ul role="tree">` using `composeEventHandlers(onKeyDownProp, handleKeyDown)`.
2. **Execution Order**: `composeEventHandlers` invokes the consumer's `onKeyDown` first. If `event.preventDefault()` is called, `composeEventHandlers` halts execution before `handleKeyDown` runs, keeping focus and roving `tabindex` locked on the active item.
3. **Type Safety**: Applied type casting (`onKeyDownProp as unknown as React.KeyboardEventHandler<HTMLUListElement>`) to resolve TypeScript element type mismatches between `HTMLDivElement` props and `HTMLUListElement` listeners.

## Verification

- Added regression unit tests to `TreeList.test.tsx` verifying:
  - `event.preventDefault()` blocks built-in navigation and preserves active item focus/tabindex.
  - Non-canceling consumer `onKeyDown` handlers still fire alongside built-in keyboard navigation.
- Executed Vitest test suite (`83/83` tests passing).
- Tested interactive behavior in Storybook.

## Changeset

Added patch changeset `.changeset/treelist-onkeydown-cancellation.md` for `@astryxdesign/core`.

### DIFF
--- a/.changeset/treelist-onkeydown-cancellation.md
+++ b/.changeset/treelist-onkeydown-cancellation.md
@@ -2,11 +2,11 @@
 '@astryxdesign/core': patch
 ---
 
-[fix] TreeList: compose consumer `onKeyDown` with built-in APG tree keyboard navigation
+[fix] TreeList: respect consumer `onKeyDown` `preventDefault` cancellation for APG tree keyboard navigation
 
-`TreeList` passed `onKeyDown` to the root `<div>` via `restProps` while attaching `handleKeyDown` to the inner `<ul role="tree">`. Because keyboard events originate on `treeitem` elements inside `ul`, the built-in navigation handler ran before the consumer handler, preventing consumer `event.preventDefault()` from suppressing built-in arrow navigation.
+`TreeList` previously processed built-in APG keyboard navigation on the inner `<ul role="tree">` before consumer `onKeyDown` ran on the root `<div>`, preventing consumer `event.preventDefault()` from suppressing built-in arrow navigation.
 
-Consumer `onKeyDown` is now composed with `handleKeyDown` on the `<ul role="tree">` element via `composeEventHandlers`. Calling `event.preventDefault()` in `onKeyDown` now successfully cancels built-in navigation and leaves focus and roving tabindex unchanged.
+Root `onKeyDown` now invokes consumer `onKeyDown` on the root container first and checks `event.defaultPrevented` before handling internal tree navigation for keydown events originating inside the `<ul role="tree">`. Calling `event.preventDefault()` in `onKeyDown` now successfully cancels built-in navigation and leaves focus and roving tabindex unchanged while preserving root handler target contracts.
 
 @Geervan
 
--- a/.changeset/treelist-onkeydown-cancellation.md
+++ b/.changeset/treelist-onkeydown-cancellation.md
@@ -7,3 +7,6 @@
 `TreeList` passed `onKeyDown` to the root `<div>` via `restProps` while attaching `handleKeyDown` to the inner `<ul role="tree">`. Because keyboard events originate on `treeitem` elements inside `ul`, the built-in navigation handler ran before the consumer handler, preventing consumer `event.preventDefault()` from suppressing built-in arrow navigation.
 
 Consumer `onKeyDown` is now composed with `handleKeyDown` on the `<ul role="tree">` element via `composeEventHandlers`. Calling `event.preventDefault()` in `onKeyDown` now successfully cancels built-in navigation and leaves focus and roving tabindex unchanged.
+
+@Geervan
+
--- a/.changeset/treelist-onkeydown-cancellation.md
+++ b/.changeset/treelist-onkeydown-cancellation.md
@@ -1,0 +1,9 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] TreeList: compose consumer `onKeyDown` with built-in APG tree keyboard navigation
+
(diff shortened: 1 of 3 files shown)


cixzhang (changes requested):
Thanks for fixing the cancellation order. One compatibility issue remains: a builder's `onKeyDown` now receives the inner `<ul>` as `event.currentTarget`, although `TreeListProps` and the ref promise the root `<div>` (`TreeList.tsx:356`). Chromium and a focused test confirm `event.currentTarget !== ref.current`. Please preserve the root handler contract while letting consumers cancel built-in tree navigation, and assert both cancellation and `currentTarget === ref.current`. If you'd rather talk it through with someone, we're in [Discord](https://discord.com/invite/XnsUcFykEP).

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks, the root `currentTarget` contract is fixed, but moving TreeList’s internal keyboard handler to the root also captures keys from the public `header` slot. In Chromium at this head, focusing a header button and pressing ArrowDown moves focus to the first tree item; the pre-PR base leaves focus on the button. Please keep internal tree navigation scoped to events originating inside `ul[role="tree"]` while preserving the root consumer handler and its `preventDefault()` cancellation. If you’d rather talk it through with someone, we’re in [Discord](https://discord.com/invite/XnsUcFykEP).

[Reviewed by Robohands]

cixzhang (approved):
Thanks, this now preserves the root handler contract, consumer cancellation, and header keyboard behavior. Please update the stale `<ul>`/`composeEventHandlers` changeset wording before merge.

[Reviewed by Robohands]

Geervan:
done, updated the changeset as requested!

cixzhang (approved):
Thanks, canceled keys now keep focus in place while preserving the root `currentTarget`, normal tree navigation, and header behavior. The changeset matches the implementation.

[Reviewed by Robohands]

cixzhang:
/accept-visual 33753971655/1 All 12 exact-head TreeList baseline frames render the intended unchanged hierarchy across the six light/dark themes; the PR changes keyboard event handling, not appearance.

PR #5609, opened 2026-08-27
fix(DateTimeInput): stack fields in narrow containers

imdreamrunner:
## Summary

- use intrinsic flex wrapping so the date and time segments stay side by side when at least 400px is available and stack below 400px
- preserve natural sizing in `fit-content`, `inline-flex`, and max-content grid layouts without size containment
- let both horizontal segments shrink within the row instead of overflowing at the boundary
- preserve the new coarse-pointer bottom-sheet picker from `main` and apply the same responsive wrapping to its closed Date/Time segments
- add component regression tests, a narrow-container Storybook story, docs, and a patch changeset

## Responsive layout

DateTimeInput at 400px and 320px available widths

## Test plan

- `pnpm exec vitest run packages/core/src/DateTimeInput/DateTimeInput.test.tsx packages/core/src/DateTimeInput/DateTimeInputTouch.test.tsx` (177 passed)
- `pnpm -F @astryxdesign/docsite test` (427 passed)
- `pnpm lint:strict` (passed; existing warnings only)
- core and docsite typechecks
- core, Storybook, and docsite production builds
- Chromium fine- and coarse-pointer matrices: 400px stays horizontal with two 196px segments; 399px and 320px wrap into full-width rows
- Chromium intrinsic-sizing matrix: `fit-content`, `inline-flex`, and max-content grid retain a natural 454px horizontal field with no overflow
- production docsite at 1440px / 1024px: Validation fields are 400px and horizontal; at 390px they are 286px and stacked, with no overflow

### DIFF
--- a/.changeset/responsive-date-time-input.md
+++ b/.changeset/responsive-date-time-input.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Stack DateTimeInput's date and time fields when its container is narrower than 400px.
+
+@imdreamrunner
--- a/apps/storybook/stories/DateTimeInput.stories.tsx
+++ b/apps/storybook/stories/DateTimeInput.stories.tsx
@@ -103,6 +103,23 @@ export const Default: Story = {
   },
 };
 
+export const NarrowContainer: Story = {
+  render: args => {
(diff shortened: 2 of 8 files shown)


imdreamrunner:
/accept-visual 33206958419/1 explain why the pixels are correct

PR #5612, opened 2026-08-27
fix(Table): drop aria-expanded from groupedRows header rows

AKnassa:
Part of #4681 (weekly a11y scan). Split out of #4690 as agreed in review: groupedRows is flat data with section headers, not a tree, so it does not ride along with the treegrid work.

## What this does

The groupedRows plugin set `aria-expanded` on each group header `<tr>`. On a native table that attribute is invalid on a row (axe `aria-conditional-attr`: rows may only expand inside a treegrid), and a group header is a section separator of flat data, not a tree parent, so a treegrid is the wrong fix here. The header row no longer carries it. The collapse state stays announced by the group's chevron button, which already carries `aria-expanded` with the "Expand group X" / "Collapse group X" name.

## What changed

- `useTableGroupedRows.tsx`: one attribute removed from the header row.
- Test rewritten to assert the row carries no `aria-expanded` and the chevron button does.
- Baseline: the 3 `TableGroupedRows::*::aria-conditional-attr` entries removed, deletions only.
- Changeset (patch) with a `tr[aria-expanded]` test-migration hint.

## Verification

- 15/15 groupedRows tests; eslint and prettier clean on the changed files.
- Repo axe audit with `--fail-on-new` on the built Storybook: 0 issues across all 3 TableGroupedRows stories.

## Observation for a follow-up

The audit reports `TableGroupedRows::Default::color-contrast` and `TableGroupedRows::Initially Collapsed::color-contrast` as no longer occurring (stale baseline entries). Left untouched here; contrast entries are out of this PR's scope.

### DIFF
--- a/.changeset/a11y-table-groupedrows-aria-expanded.md
+++ b/.changeset/a11y-table-groupedrows-aria-expanded.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Table groupedRows plugin: group header rows no longer carry `aria-expanded`. On a native table that attribute is invalid on a row (axe `aria-conditional-attr`: rows may only expand inside a treegrid), and a group header is a section separator of flat data, not a tree parent. The collapse state stays announced by the group's chevron button, which already carries `aria-expanded`. If a test queried `tr[aria-expanded]` on a grouped table, target the button instead.
+@AKnassa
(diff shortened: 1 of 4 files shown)


humbertovirtudes (approved):
This is a focused semantic correction. Native grouped-header rows no longer expose invalid aria-expanded state, while the existing localized chevron button remains the keyboard-operable source of expansion state. The focused coverage checks both expanded and collapsed groups, and only the matching accessibility baseline entries are removed. No API, visual, theming, performance, or interaction behavior changes. Approved on code merits; rebase and rerun required CI before landing.

PR #5614, opened 2026-08-27
fix(cli): stop build ranking partial query matches above full ones

ernestt:
## Summary

`astryx build "file browser"` returns two form wizards above the actual file browser — and reports one as a confident direct match, so an agent scaffolds the wrong template:

```
form-wizard-vertical  98  matches 1/2 terms: file
import-wizard         98  matches 1/2 terms: file
file-explorer         97  matches 2/2 terms: file, browser
```

After this change:

```
file-explorer  117  matches 2/2 terms: file, browser
ide             58  matches 1/2 terms: file
shell-nav       58  matches 1/2 terms: file
```

Three independent causes in the retrieval path.

### 1. Multi-token scoring was not monotonic

`scoreQuery` averaged over **matched** tokens (`sum / matched`), so a second, weaker hit could pull the mean down further than the coverage bonus pushed it up. Solving the aggregation: against a first hit of `s1`, a second hit only helps if it scores above `s1 − 39`. So against a 90-point keyword a second term has to beat 51 — and a description hit is exactly 50. Matching fewer terms well beat matching more terms partially.

The score is now based on the **strongest** matched concept plus the existing per-concept and coverage bonuses. That preserves the reason the mean was there in the first place — a verbose, low-fidelity prompt is still scored on the concepts it did hit, rather than divided across every token — while making the score monotonic: a candidate matching a superset of another's terms can no longer rank below it.

### 2. TypeScript generics were indexed as rendered components

`extractComponents`' tag regex `/<(XDS)?([A-Z]\w+)/g` matched generic argument lists. `useState<ReadonlySet<string>>`, `ComponentType<SVGProps<SVGSVGElement>>` and `Record<string, Phase>` were all reported as rendered components — 82 bogus entries across the 51 page templates, including `Record` (9 templates), `ReadonlySet` (5), `Array` (5), and every local type alias (`TaskRow`, `Severity`, `Segment`, …).

A lookbehind now requires the `<` not to follow an identifier character. Since `template <name>` and `template --skeleton` share this extractor, it also cleans up their user-facing "components used" output.

### 3. Derived keywords scored the same as authored ones

Every component a page rendered became a full-strength keyword (90), with no penalty for how many a template had. Breadth is not relevance: `theme-showcase` accrues 50 and `settings-dialog` 45 against a median of 14, so the broadest pages became noise magnets and won queries they had nothing to do with.

Derived signal now lands in a separate `weakKeywords` tier capped at 60 — below `PAGE_DIRECT` (95) on its own, so one incidental component match cannot produce a confidently wrong answer, but above `PAGE_FLOOR` (50), so it still surfaces as a reference. Authored signal (a block's `componentsUsed`, a page's `category` words) keeps its 90.

## Measurements

36-query battery, same template set, baseline = this branch's merge base:

| | top-1 | top-3 | MRR | false-confident | no results |
|---|---|---|---|---|---|
| before | 15/36 | 23/36 | 0.509 | 8/36 | 1/36 |
| after | **17/36** | **26/36** | **0.574** | **6/36** | 3/36 |

"false-confident" = `build` reports `directMatch` on the wrong template, which is the failure mode that actively misleads an agent.

Both queries that now return nothing previously returned *wrong* pages — three dashboards for "collecting information across several steps with a progress indicator", and a confidently-wrong `dashboard-executive-summary` for "drag and drop builder where you assemble a layout". Returning nothing is the honest answer; getting them right is a description problem, not a scoring one.

I swept the derived tier rather than picking 60 by feel:

| tier | top-1 | top-3 | MRR | false-confident |
|---|---|---|---|---|
| 90 (before) | 15 | 23 | 0.509 | 8 |
| **60 (this PR)** | 17 | 26 | 0.574 | **6** |
| 65 / 70 | 18 | 27 | 0.602 | 7 |
| 80 | 17 | 25 | 0.560 | 9 |

65/70 buy one top-1 for one false-confident. That is inside this battery's authoring noise (±2 queries), so I took the conservative end: at 60 two derived matches still cannot clear the confidence gate, at 70 they can. Trivial to change if you'd rather have the ranking.

## Test plan

- [x] `vitest run packages/cli/api/search/search.test.mjs packages/cli/foundation/discovery/template-adapter.test.mjs` — 26 pass
- [x] New scoring tests: the `file browser` inversion, a superset-never-ranks-lower property, and a verbose-prompt guard for the behaviour the mean was protecting
- [x] New `template-adapter.test.mjs` — first coverage for `extractComponents`: generics excluded, JSX tags and closing tags still handled
- [x] New derived-vs-authored tier assertions, pinned against `PAGE_DIRECT` / `PAGE_FLOOR`
- [x] `astryx build "file browser"` returns `file-explorer` as the direct match
- [x] eslint clean; full pre-commit gate (`check:sync`, `check:package-boundaries`, `check:changesets`, `check:cli-structure`, …) green

## Caveats

The 36-query battery is one author's, so the absolute rates reflect that battery rather than a population, and each ranking delta is 1–3 queries. The two things that are exact and independently reproducible are the `file browser` inversion and the 82 bogus keywords. The battery harness is not in this PR to keep it reviewable — happy to land it separately if it's useful to you.

### DIFF
--- a/.changeset/build-retrieval-scoring.md
+++ b/.changeset/build-retrieval-scoring.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/cli': patch
+---
+
+[fix] `build`/`search` no longer rank a partially-matched template above one matching every term, no longer index TypeScript generic arguments as rendered components, and no longer treat breadth of rendered components as full-strength relevance.
+
+@ernestt
--- a/packages/cli/api/search/search.mjs
+++ b/packages/cli/api/search/search.mjs
@@ -21,13 +21,24 @@
  *    80  name Levenshtein distance 1
  *    70  keyword substring / distance 1
  *    60  name substring (>=4 chars, >=50% coverage)
+ *    60  exact weak-keyword match
  *    50  description / prose mentions the term
  *    40  name Levenshtein distance 2
+ *    40  weak-keyword substring
  *    30  keyword Levenshtein distance 2
  *    20  name Levenshtein distance 3
  *
  * Name + keyword signals always outweigh description/prose, so an exact match
  * sorts above an incidental mention.
+ *
+ * `keywords` carry AUTHORED intent — a block's `componentsUsed`, a page's
+ * `category` words. `weakKeywords` are DERIVED: the components a page template
+ * happens to render, scraped out of its JSX. Derived signal is deliberately
+ * capped below the confident-match gate on its own, because breadth is not
+ * relevance. At full keyword strength every rendered component is an
+ * independent 90-point shot with no penalty for how many a template has, so
+ * the broadest pages (a theme showcase rendering one of everything) win
+ * queries they have nothing to do with.
  */
 
 import {pathToFileURL} from 'node:url';
@@ -53,6 +64,7 @@ import {ERROR_CODES} from '../../foundation/response/error-codes.mjs';
  * @property {'component'|'hook'|'doc'|'template'} domain
  * @property {string} name
  * @property {string[]} [keywords]
+ * @property {string[]} [weakKeywords]
  * @property {string} [description]
  * @property {string[]} [prose]
  * @property {string} [_import]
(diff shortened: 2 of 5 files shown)


josephfarina (approved):
:shipit:

PR #5616, opened 2026-08-27
Catch docs that omit a public component prop

### LINKED ISSUE #5421: Nothing guards doc/source drift: the derived #4163 gate was never filed
Ruling on [#4163](https://github.com/facebook/astryx/issues/4163) was: land the docs, drop the hand-written contract tests, and file a derived gate. The first two happened; the gate was never filed, so nothing guards doc/source drift today.

- Docs landed: [#4315](https://github.com/facebook/astryx/pull/4315) · [#4316](https://github.com/facebook/astryx/pull/4316) · [#4317](https://github.com/facebook/astryx/pull/4317) · [#4320](https://github.com/facebook/astryx/pull/4320)
- Tests dropped: [#5382](https://github.com/facebook/astryx/pull/5382) (+0/−263)
- [#4163](https://github.com/facebook/astryx/issues/4163) is the tracker, not the gate

**What is wanted:** a check that derives each component's documented props from the source and fails on drift, the way `scripts/check-theming-targets.mjs` does for theming targets. A test written in the same PR as the thing it asserts proves only that the author typed both.

AKnassa:
## What this does

Adds a check that reads each component's real TypeScript props and fails when the docs left one out.

## Why

The docs are what people actually read. The old contract tests only looked at the docs themselves, so they stayed green even when source and docs drifted apart. This check derives the public prop list from the component's TypeScript type and compares it to the component's `.doc.mjs` file.

## What changed

- New command: `pnpm check:contract`
- The prop list comes from an exported `{Name}Props` type, or, when a component has none, from the exported component or hook itself: every parameter of every overload, a class's constructor parameter, re-export aliases followed. That reaches the shared `IndicatorProps` behind the three indicators, the `DropdownMenuItemProps` behind `ContextMenuItem`, and every Table plugin hook's config. When more than one candidate exists, the one in the doc's own directory wins, then the shallower path, then the named type; only exported top-level types count, and the result does not depend on directory read order.
- Fails closed: an entry whose props cannot be read (nothing exported, or a bag typed `any`), a doc with `props[]` but no `name` or a malformed shape, a `.doc.mjs` that cannot load, a scan that finds no docs, and a program that cannot resolve `react` to types all exit 1 with their own section in the report. Invoking the script through a symlink runs it instead of silently doing nothing.
- The `docs` export is the one checked, matching what `astryx component <Name>` serves.
- Platform passthrough is BaseProps, `@types/react`, `csstype` and the TypeScript libs. A third-party component re-exported under a core name has its bag required.
- Not turned on as a required repo check yet. Running it against core today finds 87 existing gaps; wiring it into `check:repo` would fail every PR until those docs are filled in.

## How to see it

```
pnpm exec vitest run scripts/check-contract.test.mjs
pnpm check:contract
```

The second command should exit 1 and print the current undocumented props.

## Test plan

- [ ] `pnpm exec vitest run scripts/check-contract.test.mjs` — 105 tests pass in about two seconds, including regressions against the real indicator family, `ContextMenuItem`, `useResizable`, `useImperativeDialog` and the Table plugin hooks
- [ ] `pnpm check:contract` — reports the 87 existing gaps, 0 unresolved entries, and exits 1
- [ ] `pnpm check:repo` still passes (this check is not in that suite yet)
- [ ] A fixture with a syntax-error `.doc.mjs` is reported as unreadable, not skipped
- [ ] A fixture with no `.doc.mjs` at all exits 1 (an empty scan is not a pass)
- [ ] A fixture whose component has no `{Name}Props` and no exported binding exits 1 (an unchecked entry is not a pass)

Fixes #5421

https://claude.ai/code/session_01Ybs7buLRQUtDoYAf7Zo4Zz

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
     "check:executable-bits": "node scripts/check-executable-bits.mjs",
     "check:demo-media": "node scripts/check-demo-media.mjs",
     "check:i18n-catalog": "node scripts/check-i18n-catalog.mjs",
+    "check:contract": "node scripts/check-contract.mjs",
     "check:sync": "node scripts/check-sync.js",
     "check:use-client": "node scripts/check-use-client.mjs",
     "lint": "pnpm check:repo && eslint . --cache",
--- a/scripts/check-contract.mjs
+++ b/scripts/check-contract.mjs
@@ -29,8 +29,15 @@
  * filtering by name regex would hide it.
  *
  * Key lookup uses one `ts.Program` over the source tree so `extends` /
- * `Omit` / `Pick` / unions resolve. A program-per-file is correct but ~40×
- * slower.
+ * `Omit` / `Pick` resolve. A program-per-file is correct but ~40× slower.
+ * A union `{Name}Props` (`SliderSingleProps | SliderRangeProps`) is walked
+ * per member, since the union's own property list holds only the common
+ * props. A prop is platform passthrough only when every declaration behind
+ * it is platform — an intersection redeclaring `onClick` over BaseProps
+ * carries both declarations and stays public.
+ *
+ * The scan itself is gated: zero `.doc.mjs` found, or a doc that cannot be
+ * imported or exports no `docs`, fails the run instead of passing vacuously.
  *
  * Not yet in `check:repo`: core still has pre-existing drift this gate
  * reports. Wire it next to `check:i18n-catalog` once that count is zero
@@ -153,6 +160,21 @@ export function buildProgram(filePaths) {
   return {program, checker: program.getTypeChecker()};
 }
 
+/**
+ * Property symbols of `type`. On a union (`SliderSingleProps |
+ * SliderRangeProps`) `getProperties()` reports only the members' common
+ * props, so a variant-only prop (`minStepsBetweenThumbs`) would never be
+ * required. Walk each member instead.
+ *
+ * @param {import('typescript').Type} type
+ * @returns {import('typescript').Symbol[]}
+ */
+function propertySymbols(type) {
+  return type.isUnion()
+    ? type.types.flatMap(propertySymbols)
+    : type.getProperties();
+}
+
 function isInheritedPlatformDecl(declFile) {
   return /\/BaseProps\.ts$|node_modules|[\\/]lib\.dom|@types\/react/.test(
     declFile,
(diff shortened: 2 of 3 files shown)


cixzhang (changes requested):
Thanks for building the source-derived gate. One coverage hole keeps it from protecting consumers yet: a documented public component with no literal `{Name}Props` declaration is recorded as informational, and the command still exits 0. For example, `CheckboxIndicator`, `CheckIndicator`, and `RadioIndicator` are public exports documented separately, but all accept shared `IndicatorProps`; a new public member on that shared contract can be omitted from all three tables while `check:contract` reports success. `ContextMenuItem` has the same problem through a re-exported prop alias.

Please resolve shared/aliased prop bags from the exported component signature, or fail closed for unresolved public entries, and add a real-component regression proving that an omitted shared prop fails.

[Reviewed by Robohands]

AKnassa:
Right on both counts. Fixed both ways: an entry with no exported `{Name}Props` now resolves from the exported `{Name}` value (every bag parameter of every overload, class constructors, aliases followed), ranked by proximity to the doc, and anything still unresolved fails the run, as does a bag typed `any`, a nameless or malformed doc, and a program that cannot resolve `react` to types. Against core all 23 previously unresolved entries resolve; the report grows from 68 to 87 undocumented props, none lost, the new ones being `ContextMenuItem.onClick` / `isDisabled` / `variant` / `hasCloseOnSelect`, the `useResizable` and `useImperativeDialog` option bags, and two Table hooks. Regression tests run against real core source: a member added to the real `IndicatorProps` is reported against all three indicator tables, and one added to `DropdownMenuItemProps` against `ContextMenuItem`. 42 to 105 tests.

cixzhang (changes requested):
Thanks, the shared Indicator and aliased ContextMenu cases now resolve. The resolver still chooses one of two public contracts instead of reconciling them. Its new regression makes `Widget` accept `fromSignature`, then asserts only `fromDeclaration` is checked. That leaves builders able to use a real exported prop that docs omit while `check:contract` stays green—the same class as the previous blocker. Please contract this to one source per component, or fail when the declared props and exported signature disagree, with the regression expecting `fromSignature` to be covered.
[Reviewed by Robohands]

PR #5617, opened 2026-08-28
feat(Toast): reflect Theme mode as themeMode state

### LINKED ISSUE #5503: [Bug] Toast action theming cannot distinguish the root app theme
## Summary

`Toast` renders its content inside `MediaTheme mode="auto"`. That is useful for foreground contrast, but it means component theming sees the Toast's media-surface mode rather than the application's root theme.

When an application intentionally uses a dark Toast surface in both root light and root dark themes, both Toasts resolve as dark media surfaces. A component override, `light-dark(...)`, or a custom Button variant therefore selects the same branch in both application themes. There is no supported theme-level way to style a Toast action differently by root app theme.

The consumer currently has to read root `useTheme()` in product code and select between existing public Button variants at runtime.

## Versions

- `@astryxdesign/core@0.5.0`
- `@astryxdesign/cli@0.5.0`
- `@astryxdesign/theme-neutral@0.5.0`
- React 19
- Chrome on macOS

## Relevant public contracts

The documented public theming surface supports:

- `defineTheme.components.toast`
- Toast's `type` target, including `type:error`
- `defineTheme.components.button`
- Button's `size` and `variant` targets
- Public Button variants: `primary`, `secondary`, `ghost`, and `destructive`

Toast does not document a public component variable or target for styling `endContent` according to the root application theme.

## Reproduction

1. Render an app under an Astryx root `Theme` that can be light or dark.
2. Show a Toast whose surface is intentionally dark in both root modes—for example, one dark surface in the light app theme and a different dark surface in the dark app theme.
3. Put a Button in `endContent`.
4. Try to keep `secondary` in root light mode and use `ghost` in root dark mode through `defineTheme`, `light-dark(...)`, or a custom Button variant.
5. Observe that Toast's nested `MediaTheme mode="auto"` resolves as dark in both cases, so both application themes select the same media-theme branch.

Minimal shape:

```tsx
<Theme mode={rootMode} theme={theme}>
  <Toast
    body="Saved"
    endContent={<Button label="Undo" variant={actionVariant} />}
    onDismiss={() => {}}
  />
</Theme>
```

The only working consumer-side implementation is:

```tsx
const {mode} = useTheme();
const actionVariant = mode === 'dark' ? 'ghost' : 'secondary';
```

### 1. Toast component override

Using `defineTheme.components.toast` is correct for Toast background, border, and `type:error` styling, but it does not provide a root-theme-aware action target.

### 2. `light-dark(...)` in a custom action variant

This does not solve the problem. `light-dark(...)` follows the local `color-scheme`, and Toast's `MediaTheme` sets that scheme from the Toast surface. Because both Toast surfaces are dark, both root themes select the dark branch.

### 3. Global dark-media Button override

Overriding every secondary Button inside dark `MediaTheme` would affect unrelated Buttons in other dark media surfaces. It is too broad and changes component semantics outside Toast.

### 4. Private or invented Toast custom properties

Inventing variables such as `--toast-action-*` is outside the documented public component-variable contract and couples consumers to implementation details.

### 5. Consumer-side root `useTheme()` branch

This works and uses only public APIs, but it moves a visual/theming distinction into product runtime code. Every consumer with this requirement must repeat the same branching logic.

## Expected behavior

Astryx Core should provide a supported, bounded way for Toast action theming to retain the root app-theme distinction while `MediaTheme` continues to own local foreground contrast.

Possible public shapes include:

- a root-theme semantic available to component theme targets;
- a Toast action/end-content variant target with root-theme branches; or
- another explicit API that distinguishes root application mode from nested media mode.

The exact API is up to Core. It should avoid globally changing Buttons in all dark media surfaces and should not require consumers to reach into private CSS variables or component internals.

## Privacy and data-minimization requirement

No sensitive-data leak has been observed. This is a preventive requirement for any new API, diagnostic, or telemetry added while solving the gap.

Only bounded semantic values should cross the theming or diagnostic boundary, such as:

- `light | dark`;
- a documented mode enum; or
- a bounded public variant name.

The implementation must not expose, serialize, inspect, or log:

- Toast body content;
- arbitrary `endContent`;
- action labels or hrefs;
- user-authored content;
- user, account, or session identifiers;
- URLs or navigation targets;
- tokens or credentials;
- serialized React props;
- arbitrary DOM attributes or text; or
- component children.

Core tests should cover both root theme modes and include a negative assertion that content and arbitrary props are not surfaced through the theming, diagnostics, or telemetry boundary.

## Acceptance criteria

1. Toast action styling can distinguish root light and root dark modes even when the Toast's nested `MediaTheme` is dark in both.
2. A consumer can express root-light `secondary` and root-dark `ghost` without product-owned theme branching.
3. `MediaTheme mode="auto"` continues to provide correct foreground contrast.
4. The behavior is scoped to Toast actions and does not restyle unrelated Buttons in dark media surfaces.
5. The API uses documented public contracts—no private CSS variables, swizzling, or component-internal selectors.
6. Any diagnostics or telemetry use a closed semantic vocabulary and do not expose or log content, labels, hrefs, identifiers, URLs, tokens, serialized props, arbitrary DOM data, or children.
7. Core tests cover both theme modes and the data-minimization requirement.

AKnassa:
## What this does

A toast card now says which app colour mode it is in, light or dark, as a theming state called `themeMode`. A theme can style the toast, and what sits inside it, by app mode even when the toast's own surface is dark in both.

## Why

A toast paints an inverted surface, and everything inside it follows that surface: `light-dark()` tokens and `onDark`/`onLight` rules both key on the media side. When the surface is dark in both app modes (the error toast is, and a brand that keeps its toasts dark is too), nothing inside the card could tell a dark app from a light one. The only way to give a toast action a different look per app mode was to pick its `variant` from `useTheme()` in product code. #5503 asks for a bounded, public way to do it in the theme instead.

## What changed

- The card reflects the resolved mode of the Theme it renders under as `data-theme-mode="light|dark"` (never `system`) beside `data-type`, from the value Toast already read for its MediaTheme fallback.
- `Toast.doc.mjs` documents it as a theming state, so `toast: {'themeMode:dark': {...}}`, alone or compounded as `'type:error+themeMode:dark'`, is a valid `defineTheme` key. `astryx theme targets Toast` lists it and `theme build` validation accepts it.
- To reach an action in `endContent`, a theme sets a property it owns on the mode rule and reads it from its Button rule behind the variant's normal value. The new **Themed Toast Action** story shows the shape: product passes `variant="secondary"`, the theme decides whether that reads filled or ghost per mode.
- `MediaTheme.doc.mjs` gains one pointer: `onDark`/`onLight` follow the surface, not the app; Toast's `themeMode` is how a theme branches on the app.
- Both reflected values are closed vocabularies. A runtime `type` other than `error` reflects as `info`, which is how the card already renders it, so a padded string cannot forge the `dark` state class.

Two small fixes the edge cases forced, each one line of behaviour with its own commit and tests. Happy to split either into its own PR if that reads better.

- `useTheme().mode` returned a Theme `mode` outside `light | dark | system` as is (a stale persisted preference cast to `ThemeMode` gets there), so its declared `'light' | 'dark'` type did not hold and the card carried the raw string. It now resolves through the OS preference, the way the root Theme's `<html data-theme>` sync already treats such a value.
- A `Theme` rendered inside toast content in `useToast`'s fallback viewport (no `ToastViewport` mounted) acted as the app's root Theme, because the fallback root is a separate React root: it rewrote `<html data-theme>` and `data-astryx-theme` while the toast showed and stripped both on dismiss with the app's Theme still mounted. The fallback root now provides `ThemeNestingContext`, so such a Theme themes its own subtree only. Pre-existing; #5503 made it visible on the card.

## How to see it

Storybook → Core/Toast → **Themed Toast Action**. Both columns paint the same dark cards; the action is filled under the light app and ghost under the dark app, from `themeMode` alone.

Under a dark app, `MediaTheme mode="auto"` resolves to *off* on that card (its ambient text already reads), so the two columns differ in media context as well, and only `themeMode` separates them.

## Tests

`packages/core/src/Toast/Toast.test.tsx` (37): both modes, `system` resolving to the OS preference, the nearest Theme, live changes to the Theme mode, the OS preference and `<html data-theme>` with their listeners released on unmount, an explicit mode ignoring the OS, independence from the media fallback, the no-Theme paths, bogus or `system` root attributes never reaching the card, a custom `renderContent` layout (reflection kept, a spoofed mode on the layout not hoisted), server markup and hydration under an OS that prefers dark with no mismatch, the detached fallback viewport, a Theme inside toast content, the viewport's Theme winning over the trigger's, a mode outside the union, a forged type, the compiled keys, the custom-property recipe in generated CSS, and a data-minimization check. `useTheme.test.tsx` (+4) and `useToast.test.tsx` (+1) pin the two fixes at home. Every test was shown red against a targeted mutation before landing.

Full suite: 629 files / 13,170 tests; core and storybook typecheck, `CI=true eslint`, prettier and `check:repo` clean.

## Found along the way, not in this PR

- The CLI and docsite override example print every valued state bare (`'themeMode'`, same as TopNav `mode`), a pre-existing printer limitation; the Toast doc states the valued form. `theme build` cannot value-check a state either, so `'themeMode:system'` compiles to a selector nothing renders.
- The card's bare `dark` class also matches class-strategy dark-mode rules such as Tailwind's `dark:` variants, for apps that drive dark mode through Theme without a `.dark` on `<html>`. The convention already ships `.active` and `.selected`; the doc says to key external CSS on `[data-theme-mode]`. Flagging in case a prefixed class is preferred for this pair.
- A card portaled outside its Theme's DOM subtree reflects the React Theme while the scoped CSS reaching it is the DOM ancestor's; Astryx's own viewports do not portal.
- Tooltip, Overlay and HoverCard render inverted surfaces too and do not reflect the app mode yet; the MediaTheme doc says so.

Closes #5503


rubyycheung (commented):
The mechanism is clear, but the current visual example does not establish the intended product outcome. Both Toast surfaces are identical, while the action changes from filled to ghost solely because the surrounding app mode changes; without broader page context, that difference reads as arbitrary rather than purposeful. Could you add a real product or brand-theme example showing why the action hierarchy should differ between root light and dark modes, ideally in full-page context, and document the intended design rationale?

PR #5620, opened 2026-08-28
feat(DateTimeInput): add native picker option

imdreamrunner:
## What

Add `nativePicker` to `DateTimeInput`, matching `DateInput`'s surface-selection API for **both** date and time:

| Value | Behavior |
| --- | --- |
| `'touch'` (default) | Browser/OS `<input type="date">` and `<input type="time">` on a coarse primary pointer; Astryx surfaces otherwise |
| `'always'` | Browser/OS date and compatible time controls on every pointer type |
| `'never'` | Astryx's own surfaces everywhere — pointer fields on fine pointers and the coordinated Date/Time bottom sheet on coarse pointers |

The native controls keep DateTimeInput's two-segment field, distinct labels, controlled ISO datetime value, constraints, theming targets, and public date-input ref.

Both native segments use the iOS-safe shape established by `DateInput`: uncontrolled DOM values, native `input`/`change` listeners with deduplication, no programmatic writes while focused, and stale-blur protection. A time chosen before the date is retained and combined when the date commits. Pending `changeAction` work no longer disables or detaches an open native wheel.

### Compatibility fallbacks

The OS time picker cannot faithfully provide every Astryx time feature. `hasSeconds`, a non-default `timeIncrement`, or `timeOptionInterval` retain Astryx's time field while the date remains native. Date constraints and datetime min/max are always enforced in JavaScript; native min/max attributes are hints because iOS may not visually disable invalid wheel values.

## iOS verification

Verified against the built component in iOS 26.5 Safari on an iPhone 17 Pro simulator:

- tapping the DateTimeInput time segment opens the native hour/minute/AM-PM wheel;
- moving the minute wheel updated the controlled value from `2026-03-15T14:30` to `2026-03-15T14:32`;
- a focused native control is not rewritten or detached by controlled/optimistic updates.
- sizing now follows DateInput: a compact 180px native-field floor, with a 240px minimum only in the doc-site showcase; native value slots remain shrinkable.

## Test plan

- 30 native-picker regression tests covering surface selection, both ISO halves, time-before-date drafting and boundary clamping, native event deduplication, controlled rejection, optimistic rollback/clear, pending actions, min/max rejection, ARIA, clear/reset, and compatibility fallbacks
- Existing desktop + touch DateTimeInput suites: 177 tests
- All 207 DateTimeInput tests pass
- `pnpm lint:strict`
- `pnpm build`
- Core/docs and Storybook typechecks
- Storybook production build
- Isolated CLI suite: 2,869 tests pass (the monolithic local run retains its known load-time timeout flake)

### DIFF
--- a/.changeset/date-time-input-native-picker-option.md
+++ b/.changeset/date-time-input-native-picker-option.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] Add `nativePicker` to DateTimeInput for browser and OS date/time pickers, with Astryx time fallbacks for seconds, custom increments, and preset options. Native fields follow DateInput's compact minimum sizing in fit-content layouts.
+@imdreamrunner
--- a/apps/storybook/stories/DateTimeInput.stories.tsx
+++ b/apps/storybook/stories/DateTimeInput.stories.tsx
@@ -74,6 +74,12 @@ const meta: Meta<typeof DateTimeInput> = {
       control: 'boolean',
       description: 'Whether to show a clear button',
     },
+    nativePicker: {
+      control: 'radio',
+      options: ['touch', 'always', 'never'],
+      description:
+        "Date and time picker surfaces: native browser/OS controls on touch by default, native wherever compatible, or Astryx's surfaces everywhere",
+    },
     numberOfMonths: {
       control: 'radio',
       options: [1, 2],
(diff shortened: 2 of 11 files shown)


rubyycheung (approved):
Looks good from design. The nativePicker modes provide a clear choice between OS-native and Astryx-owned date/time surfaces, while preserving the established two-segment field presentation and explicit fallbacks for unsupported time features.

PR #5623, opened 2026-08-28
feat(PowerSearch): improve the touch filtering experience

imdreamrunner:
## What

Make `PowerSearch` use a bottom-sheet flow on supported coarse-pointer devices.

- Treat the entire touch field as one **Manage filters** trigger while preserving normal token wrapping.
- Keep selected capsules visible but display-only: no direct edit or remove actions.
- Use a **Filters** management sheet with divided, chevron-ended selected-filter rows whose formatted values mirror PowerSearch tokens’ bold emphasis, top-right **Add filter**, and sticky **Clear all** plus **Done** actions.
- Open a selected row to edit it; removal lives in the edit sheet alongside explicit **Cancel** and **Edit filter** actions.
- Move content search into **Add filter** and render the existing field, operator, value, and free-text PowerSearch suggestions as an inline divided list below the input rather than a popover.
- Open **Add filter** directly whenever no filters are selected, including content-search configurations.
- Render single- and multi-value enum choices with the same divided `List` + `ListItem` treatment and trailing checkmarks on selected rows.
- Keep simple field/operator titles such as **Status is** and **Title contains**. Multi-operator fields show field-prefixed radio choices directly in the value sheet.
- Keep the touch implementation private; `PowerSearch` selects it internally with `(pointer: coarse)` and preserves the public API.
- Keep nested filters, configured token overflow, and unsupported content-search value types on the pointer/typeahead surface.

## Compatibility and accessibility

- Retains existing PowerSearch and input-component dimensions on touch.
- Revalidates controlled edits by object identity before structural equality, including cloned/reordered arrays.
- Preserves exact `onChange` metadata and forwards `timezoneID` to custom editors on both surfaces.
- Disables already-open field and value steps if the component becomes disabled or read-only.
- Restores focus across Back, Cancel, confirmation, deletion, direct field-picker dismissal, and Clear all transitions.
- Keeps sheet button actions at least 44×44 CSS px without resizing inputs.
- Supports keyboard movement from the content-search input into its inline suggestion list and guards IME composition.
- Announces selected enum values while retaining the visual check-list treatment.
- Adds RTL relationship coverage for the touch-sheet header.

This supersedes the separate public `PowerSearchMobile` shape explored in #5314. It also reconciles #5804: that accepted popover contract applies when the pointer/typeahead surface is active, while supported coarse-pointer configurations use the bottom sheet.

## Visual walkthrough

<p><strong>Touch field</strong></p>
PowerSearch touch field with a selected Priority capsule

| Filters | Add filter with inline suggestions |
| --- | --- |
| Filters management sheet with bold filter value | Add filter sheet with inline search suggestions |

| Multi-value selection | Updated filter list |
| --- | --- |
| Status multi-select list with two checked values | Filters sheet with bold formatted filter values |

<p><strong>Edit actions</strong></p>
Edit filter sheet with Delete, Cancel, and Edit filter actions

## Test plan

- [x] 132 focused routing, touch-management, source, and pointer PowerSearch tests
- [x] Full Core/Lab suite: 370 files and 9,652 tests
- [x] `pnpm lint:strict` and full monorepo `pnpm build`
- [x] Core docs and Storybook typechecks
- [x] i18n catalog, knowledge checks, and Lab readiness
- [x] Production Storybook and docsite builds
- [x] Production Chromium: management list, Add filter search/results, enum selection, edit/delete/cancel, focus return, and 44 px action targets
- [x] Targeted RTL audit remains covered


cixzhang (changes requested):
Thanks for the thorough touch flow. I don't think this can ship on the current PowerSearch contract yet.

The component currently uses one editor flow across input modes, but [the new `(pointer: coarse)` branch](https://github.com/facebook/astryx/blob/ee2985044b77aa0f4747990c7082a89a5310a6cd/packages/core/src/PowerSearch/PowerSearch.tsx#L588-L595) replaces the whole interaction tree. In Chromium, changing that query with `Closed` staged closed the sheet, discarded the draft, and dropped focus to `<body>`.

The [theme docs now promise](https://github.com/facebook/astryx/blob/ee2985044b77aa0f4747990c7082a89a5310a6cd/packages/core/src/PowerSearch/PowerSearch.doc.mjs#L284-L291) `size`, `status`, and `disabled` selectors, but [the pointer renderer](https://github.com/facebook/astryx/blob/ee2985044b77aa0f4747990c7082a89a5310a6cd/packages/core/src/PowerSearch/PowerSearch.tsx#L1106-L1114) emits only the bare target; those selectors silently work on touch and fail on desktop.

Please keep the existing typeahead/popover contract and a single consistent theme surface in this PR.

[Reviewed by Robohands]

PR #5624, opened 2026-08-28
build: replace rimraf cleanup with Node fs

cixzhang:
## Why

Package builds only need a cross-platform recursive delete before writing `dist`. Node provides that operation directly, so carrying a package-local `rimraf` binary adds an avoidable dependency and failure point.

## What

Use one shared `fs.rmSync` cleanup helper across core, canary packages, and themes; remove `rimraf` from their manifests and lockfile; keep the portability guard aligned.

## Risk

Build tooling only. The helper removes only the current package’s `dist` directory with the same recursive, force semantics.

## Testing

- `pnpm exec vitest run --project node scripts/clean-dist.test.mjs scripts/check-portable-scripts.test.mjs`
- `pnpm install --frozen-lockfile`

### DIFF
--- a/packages/charts/package.json
+++ b/packages/charts/package.json
@@ -46,7 +46,7 @@
   },
   "scripts": {
     "prepack": "pnpm build",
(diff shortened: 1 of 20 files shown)


cixzhang:
/accept-visual 33140331410/1 Build-only cleanup changes how dist is deleted; it does not change component or theme pixels, and CI published trusted budget-deferred evidence with no captured deltas.

PR #5627, opened 2026-08-28
fix(Button): correct loading and disabled interaction states

rubyycheung:
## Summary
- keep loading spinners at full contrast while blocking interaction
- reserve the dimmed visual treatment for explicitly disabled buttons
- suppress pressed transforms and overlays for disabled and loading states
- add regression coverage for loading and disabled class behavior

## Tests
- pnpm exec vitest run packages/core/src/Button/Button.test.tsx scripts/check-neutral-theme-contrast.test.mjs scripts/check-badge-contrast.test.mjs
- pnpm exec eslint packages/core/src/Button/Button.tsx packages/core/src/Button/Button.test.tsx
- pnpm --filter @astryxdesign/core build


imdreamrunner (approved):

rubyycheung:
/accept-visual 33215493576/2 Loading buttons intentionally retain full visual contrast while remaining non-interactive; explicitly disabled buttons stay dimmed across all themes.

PR #5629, opened 2026-08-28
ci: centralize gh-pages publishing

cixzhang:
## Summary
- add one shared gh-pages publisher with a durable queue stored on gh-pages
- move main stable-site deploy and release-gate report publication onto it
- keep stable deploy as an orphan/whole-tree publisher while preserving unrelated paths
- add race coverage for scoped preservation, queue ordering, retry after a rejected push, and the deleted-checkout retry case

## Test plan
- `actionlint -ignore 'label "2-core-ubuntu-arm" is unknown' -ignore 'shellcheck reported issue' .github/workflows/deploy.yml .github/workflows/release-gate.yml`
- `pnpm check:repo`

## Follow-up migration
- PR previews and redeploy-preview
- preview cleanup
- visual evidence publication
- visual baseline promotion
- compaction and release report/vibe writers

### DIFF
--- a/.github/scripts/gh-pages-publisher.mjs
+++ b/.github/scripts/gh-pages-publisher.mjs
@@ -1,0 +1,6 @@
+#!/usr/bin/env node
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+import {main} from './lib/gh-pages-publisher.mjs';
+
+await main();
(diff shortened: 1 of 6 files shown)


cixzhang (commented):

Thanks for pulling these publishers behind one durable owner — the local queue, preservation, and retry tests are solid. The contention path checks the current holder through GitHub’s “Get a workflow run” endpoint, but both migrated jobs grant only `contents: write`; because unspecified scopes become `none`, that lookup has no required Actions read access. The first overlapping publication therefore exits instead of waiting, which is the race this stack needs to close.

Could we grant `actions: read` to both jobs and cover that permission contract in the workflow test before building the next migration on top?

cixzhang (commented):
Semantic verdict: approve

We asked for `actions: read` on both publisher jobs and a contract test. This head adds both; focused tests pass. Approve once undrafted.

PR #5631, opened 2026-08-28
ci: route visual gh-pages writers through publisher

cixzhang:
## Summary
- migrate PR visual evidence, visual acceptance records, and visual baseline promotion onto the shared gh-pages publisher
- keep immutable evidence paths immutable and baseline promotion gated by the shared FIFO turn
- keep manual baseline promotion on the same durable queue
- add integration coverage for immutable evidence, acceptance records, and accepted baseline promotion races

Depends on #5629.

## Test plan
- `actionlint -ignore 'label "2-core-ubuntu-arm" is unknown' -ignore 'shellcheck reported issue' .github/workflows/pr-comment.yml .github/workflows/visual-acceptance.yml .github/workflows/visual-acceptance-promote.yml .github/workflows/visual-baseline.yml .github/workflows/release-gate.yml .github/workflows/deploy.yml`
- `pnpm check:repo`

## Follow-up migration
- PR preview deploy and redeploy-preview
- preview cleanup
- gh-pages compaction
- vibe screenshot report uploads

### DIFF
--- a/.github/scripts/lib/gh-pages-publisher.mjs
+++ b/.github/scripts/lib/gh-pages-publisher.mjs
@@ -9,6 +9,8 @@ import {fileURLToPath} from 'node:url';
 
 const PAGES_BRANCH = 'gh-pages';
 const QUEUE_REL = path.join('.astryx-gh-pages', 'publication-queue');
+const LEGACY_BASELINE_QUEUE_REL = path.join('visual-gate', 'publication-queue');
+const BASELINE_SCOPE = 'visual-gate/baseline';
 const HOLDER_NAME = 'holder.json';
 const RUN_ID = /^[1-9][0-9]*$/;
 const SCOPE = /^[a-z0-9][a-z0-9._/-]*$/;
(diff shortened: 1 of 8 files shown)


cixzhang (commented):

Thanks for moving the visual writers behind the shared publisher. Two concurrency gaps still leave the migration unsafe:

- [`visual-acceptance.yml:144`](https://github.com/facebook/astryx/blob/311ba8cdc3c0f3e94207d86bc151559e8902758a/.github/workflows/visual-acceptance.yml#L144-L147) now enters the shared wait path, but its `accept` job grants no `actions: read`; an acceptance overlapping any publisher exits before recording.
- The executable [former baseline lock](https://github.com/facebook/astryx/blob/311ba8cdc3c0f3e94207d86bc151559e8902758a/.github/scripts/visual-gate/lib/baseline-publication-lock.mjs#L15) still writes a separate queue. A local roll-forward probe acquired old and new holders simultaneously, so pre-merge and post-merge runs can each believe they own baseline publication.

Could we retain the required permission for every queued caller, leave one authoritative queue across rollout, and cover both contracts before stacking the next migration?

[Reviewed by Robohands]

cixzhang (commented):

Thanks for closing the permission path — `actions: read` is present and the negative control fails on the prior head.

The rollout bridge still only consults the legacy queue for `visual-gate/baseline`. I drove the production old lock at run 899 and a current `visual-gate/reports` writer at run 900 against one bare remote; both acquired different holders simultaneously. The new baseline-only regression passes, but [the scope guard](https://github.com/facebook/astryx/blob/d64f87286431e6003d6b5da1df379706f09844e9/.github/scripts/lib/gh-pages-publisher.mjs#L803-L823) leaves every other shared publisher outside the bridge, so old and new publication ownership still overlap during rollout.

Could we make every current shared publisher observe the legacy holder while old runs can exist, then cover that cross-scope production case?

[Reviewed by Robohands]

cixzhang (commented):

Thanks — `actions: read` remains, an old baseline ticket blocks a new report writer, and an unblocked report owns both holders.

One ordering hole remains. If newer run 900 acquires the legacy holder after seeing no older shared ticket, then older run 899 enqueues before 900 claims the shared holder, [the two-step acquisition](https://github.com/facebook/astryx/blob/7c03d407e6c6e6c3044a4209f76134d7cea8ec84/.github/scripts/lib/gh-pages-publisher.mjs#L801-L819) deadlocks: 900 waits on 899’s shared ticket while 899 waits on 900’s legacy holder. I reproduced both timeouts with the production CLIs against one bare remote.

Could we make current publishers acquire the two authorities in one monotonic order, and add this late-arrival regression?

[Reviewed by Robohands]

cixzhang (commented):

Thanks — shared-first/legacy-second lets a late legacy-only run finish, and reverse release is clean. The current/current case still deadlocks: at [`waitForPublicationTurn()`](https://github.com/facebook/astryx/blob/89c9f042ddc4d481271fb1be9f7ad21cb7743c54/.github/scripts/lib/gh-pages-publisher.mjs#L751-L759), after run 900 owns shared, a late lower run 899 adds tickets to both queues; 900 waits on 899’s legacy ticket while 899 waits on 900’s shared holder. I reproduced both exact-head timeouts with the production queue code. The added regression drives the old lock for 899, so it misses this class.

Could we make one authority determine order across both current/current and legacy/current rollout, and cover this current/current interleaving?

[Reviewed by Robohands]

cixzhang (commented):
Thanks — the prior 900/899 deadlock is fixed in the implementation. I verified one same-commit dual-holder claim across all five scopes, both current-run orderings, simultaneous claims, terminal cleanup, and a forced push retry.
The checked-in regression still does not protect that fix. [`lets a late older legacy run finish…`](https://github.com/facebook/astryx/blob/d8738a1bcaa99e27e56e9c465ba74c46083046cf/.github/scripts/lib/gh-pages-publisher.test.mjs#L516-L530) drives the legacy-only CLI for 899; it never starts a second current publisher. I replaced the atomic claim with separate shared-then-legacy claims and all 27 focused tests still passed, so the prior wait-cycle model can return under green tests. There is also no simultaneous-claim or dual-queue retry test.
Could we check in the current/current late-arrival race plus simultaneous/retry coverage before this lands?
[Reviewed by Robohands]

cixzhang (commented):

Thanks — simultaneous/retry coverage now kills the sequential-claim mutant (3/14 failures), and exact code passes the full matrix.

[The checked-in boundary](https://github.com/facebook/astryx/blob/14bccbdb581acb893207a8b587b5aafb42037b6d/.github/scripts/lib/gh-pages-publisher.test.mjs#L631-L721) still misses three required regressions: sequential release, a terminal legacy holder left behind, and a lower current run inserted between observation and claim all keep 14/14 publisher tests green. A future refactor could therefore restore partial ownership or a timeout without CI noticing.

Could we make those three mutants fail before this lands?

[Reviewed by Robohands]

cixzhang (commented):
Semantic verdict: approve

Thanks — the prior review’s four mutation gaps are closed: 32 tests and 10 runtime cases pass. Landing waits for [#5629](https://github.com/facebook/astryx/pull/5629) and exact-head CI.

[Reviewed by Robohands]

cixzhang (commented):

Thanks — the restack preserves patch-id `6a9447a1` and the verified queue behavior. Exact-head CI now exposes one migration gap: [the promotion step](https://github.com/facebook/astryx/blob/ba4287ed93a0b082c699d10943fd7dac59950f85/.github/workflows/visual-acceptance-promote.yml#L344-L350) moved to the shared publisher, but `visual-promotion-workflow.test.mjs` still extracts only `run: |` and its sandbox does not install that publisher. All 10 recovery/fail-closed cases fail before exercising promotion; I reproduced this locally, while the parent workflow passes 12/12.

Could we migrate that executable workflow harness to the publisher, then get its 12 tests and exact-head CI green?

[Reviewed by Robohands]

cixzhang (commented):
Semantic verdict: approve

Thanks — at `13a2aa3f`, the harness executes the migrated publisher. Promotion 13/13, focused 45/45, runtime 10/10, all four regression mutants fail, and exact-head CI is green.

[Reviewed by Robohands]

PR #5630, opened 2026-08-28
fix(ci): simplify PR visual scope

cixzhang:
## Summary

- simplify PR visual scope to the compact stable contract:
  - changed Core components require `Theme Sheet` in Neutral + Probe, light/dark
  - changed `stable-visual` stories opt into Neutral light/dark
  - deleted `stable-visual` stories produce before-only removal evidence without trying to capture deleted head stories
  - stable theme/global scopes defer to the protected release gate instead of truncating
- normalize exact plan objects so current frames + before-only removals share the same 24-warning / 40-cap budget
- keep Probe private/test-only and current-only: it is captured, reviewed, and recaptured as evidence, but never promoted into the stable pixel baseline
- allow `/accept-visual` to record removed entries with before/preimage and no after hash; promotion remains add/update-only and never prunes

#5608 has landed; this branch is now based on current `main` at `a83c559a51fb86260f8309f695b68acd612b2eef`.

## Test plan

- `npx --yes pnpm@11.10.0 vitest run internal/lab-readiness/audit.test.mjs .github/scripts/lib/affected-scope.test.mjs .github/scripts/visual-scope.test.mjs .github/scripts/analyze-pr.test.mjs .github/scripts/generate-pr-comment.test.mjs .github/scripts/visual-gate/lib/plan.test.mjs .github/scripts/visual-gate/visual-acceptance.test.mjs .github/scripts/visual-gate/publish-pr-report.test.mjs .github/scripts/visual-gate/workflow-concurrency.test.mjs`
- `actionlint -shellcheck= -ignore 'label .* is unknown' .github/workflows/ci.yml .github/workflows/pr-comment.yml .github/workflows/visual-acceptance.yml`
- `npx --yes pnpm@11.10.0 check:repo`

PR #5608, opened 2026-08-27
fix(ci): keep accepted stable shots in the release gate

cixzhang:
## Why

The release gate could mistake a partial capture for deletion: a deeper PR plan added stable baseline keys that the representative daily plan did not recapture. The first visible run captured 428 shots and reported 454 removals, including 360 still-valid Core shots.

## What

- Resolve each built Storybook story through one production path: `componentPath`, or the generated `importPath` when Storybook omits it. Package ownership and eligibility come from workspace `package.json` files.
- Make stored baseline shot metadata the durable membership source. New acceptances record package ownership and its source; legacy entries are normalized from existing stored package, component, title, and theme metadata—without a registry.
- Partition every accepted key into exactly one current category: planned stable, intentionally excluded, preserved legacy, or unclassified. The verdict contains concise counts and lists exact keys only when unclassified is nonzero.
- Generate one internal canonical release plan (`surface` + `theme-matrix` + reproducible stable baseline coverage) with sorted keys, digest, lane, and removal-reporting authority. Callers cannot narrow it.
- Ordinary and PR comparisons never report removals. Deleted accepted stable stories remain in the comparison baseline and become reviewed removals.
- Capture failures and unclassified baseline keys produce a current failed verdict with zero removals; they cannot leave a prior successful verdict looking current.
- Keep PR acceptance and post-merge promotion update/add-only. Pruning remains explicit and removes only reviewed removal keys named in `--keys`.

The original proposal was reduced from **23 files, +3164/−206** to **10 files, +963/−222** (**net +741**). It has no legacy ownership snapshots, story-owner registry, copied built-index fixture, caller-controlled removal authority, or extra proof registry.

No baseline data is changed or pruned by this PR.

## Evidence

- Live baseline accounting: **974 total = 882 planned stable + 92 intentionally excluded + 0 preserved legacy + 0 unclassified**.
- The 94 formerly unreachable entries now remain explicit: 92 are excluded by package metadata; the two Core Layer entries remain stable through generated story metadata.
- Simulated deletion of an accepted Core story: 2 keys become preserved legacy and are reported as current removals.
- Simulated capture failure: current verdict is `failed`, removals are empty.
- Fresh dists-absent Storybook build: 38 bare package paths across 16 files (Charts 32, Vega 6), all resolved by production code.
- Canonical release plan: **884 shots**, 882/882 stable baseline keys plus 2 additions; 0 private/canary shots; OverflowList 168, Carousel 196.

## Testing

[...]

PR #5632, opened 2026-08-28
fix(ci): trace affected PR check dependencies

cixzhang:
## Summary

- add one local Node affected-scope action backed by the executable shared module
- route check-components, pr-a11y, pr-visual, and pr-rtl through that action instead of per-job component parsing
- keep component source and nested component barrels on exact PR scope, while shared Core hooks/utils/focus/runtime/i18n/naming, theme/token infra, Storybook config/static assets, browser/RTL harness inputs, package manifests, and lockfile ambiguity defer to protected main
- preserve exact changed component checks in mixed component + broad/global scopes; broad/global deferral no longer suppresses component a11y/RTL scope
- make the affected-scope CLI strict: exactly one known subcommand/input path, no stray positionals, malformed args/JSON/paths fail nonzero
- teach lab-readiness to import/execute/validate the shared helper and Node action contract, and fail closed on missing/malformed/shadow-command drift

#5608 has landed. This branch is rebuilt on current main `1b706d5d2d35cf141fd8384ce9952d660d0a9221`, carries an exact-patch copy of #5630 as `6413ca02d148085ae130f282d7304437d09e7589`, then the #5632 delta as `ad79df2334de751fa54586ce17afb5985b4cee0f`. #5630 itself remains unchanged at `04246c0051029d017e6aa76b3b649bc55bb622dc`.

## Test plan

- `npx --yes pnpm@11.10.0 vitest run .github/scripts/lib/affected-scope.test.mjs`
- `npx --yes pnpm@11.10.0 exec eslint --no-ignore .github/actions/affected-scope/index.mjs`
- `actionlint -shellcheck= -ignore 'label .* is unknown' .github/workflows/ci.yml .github/workflows/pr-comment.yml .github/workflows/visual-acceptance.yml`
- `npx --yes pnpm@11.10.0 check:repo`

### DIFF
--- a/.github/actions/affected-scope/action.yml
+++ b/.github/actions/affected-scope/action.yml
@@ -1,0 +1,34 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+
+name: Affected scope
+description: Resolve exact affected components and broad/deferred PR-check inputs.
+
+inputs:
+  changed-files:
+    description: Newline-separated changed paths. Ignored when analysis-file is set.
+    required: false
+    default: ''
+  analysis-file:
+    description: analysis.json file containing changedFiles.
+    required: false
+    default: ''
+
+outputs:
+  affected_components:
+    description: Comma-separated exact affected component names.
+  affected_core_components:
+    description: Comma-separated exact affected Core component names.
+  affected_deferred_reasons:
+    description: Comma-separated broad/deferred reason keys.
+  a11y_deferred:
+    description: Whether a11y also needs protected main/full coverage.
+  has_components:
+    description: Whether exact affected components exist.
+  rtl_deferred:
+    description: Whether RTL also needs protected main/full coverage.
+  json:
+    description: Path to the canonical affected-scope JSON file.
+
+runs:
+  using: node20
+  main: index.mjs
--- a/.github/actions/affected-scope/index.mjs
+++ b/.github/actions/affected-scope/index.mjs
@@ -1,0 +1,98 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+import {appendFileSync, readFileSync, writeFileSync} from 'node:fs';
+import {createRequire} from 'node:module';
+import {tmpdir} from 'node:os';
+import {join} from 'node:path';
+import process from 'node:process';
+import {error as logError} from 'node:console';
+
+const require = createRequire(import.meta.url);
+const {
+  classifyAffectedDependencies,
+} = require('../../scripts/lib/affected-scope.js');
+
+function input(name) {
+  return process.env[`INPUT_${name.toUpperCase()}`] ?? '';
+}
+
+function appendOutput(name, value) {
+  appendFileSync(process.env.GITHUB_OUTPUT, `${name}=${value}\n`);
+}
+
+function hasControlCharacter(value) {
+  for (let index = 0; index < value.length; index += 1) {
+    if (value.charCodeAt(index) < 32) return true;
+  }
+  return false;
+}
+
+function validateChangedPath(file) {
+  if (
+    typeof file !== 'string' ||
+    file.length === 0 ||
+    file.startsWith('/') ||
+    file.split('/').includes('..') ||
+    hasControlCharacter(file)
+  ) {
+    throw new Error(`malformed changed path ${JSON.stringify(file)}`);
+  }
+  return file;
+}
+
+function changedFiles() {
+  const analysisFile = input('ANALYSIS-FILE');
+  const changedFilesInput = input('CHANGED-FILES');
+  if (analysisFile && changedFilesInput) {
+    throw new Error('choose exactly one affected-scope input');
+  }
+  if (analysisFile) {
+    const analysis = JSON.parse(readFileSync(analysisFile, 'utf8'));
+    if (!Array.isArray(analysis.changedFiles)) {
+      throw new Error('analysis file has no changedFiles array');
+    }
+    return analysis.changedFiles.map(validateChangedPath);
+  }
+  return changedFilesInput.split('\n').filter(Boolean).map(validateChangedPath);
+}
+
+function csv(values) {
+  return values.join(',');
+}
+
+function main() {
+  if (!process.env.GITHUB_OUTPUT) {
+    throw new Error('GITHUB_OUTPUT is required');
+  }
+  const result = classifyAffectedDependencies(changedFiles());
+  const jsonPath = join(
+    process.env.RUNNER_TEMP || tmpdir(),
+    `affected-scope-${process.env.GITHUB_RUN_ID || 'local'}-${process.env.GITHUB_RUN_ATTEMPT || '1'}.json`,
+  );
+  writeFileSync(jsonPath, `${JSON.stringify(result)}\n`);
+  appendOutput('has_components', String(result.components.length > 0));
+  appendOutput('a11y_deferred', String(result.a11y.deferToMain));
+  appendOutput('rtl_deferred', String(result.rtl.deferToMain));
+  appendOutput(
+    'affected_components',
+    csv(result.components.map(entry => entry.component)),
+  );
+  appendOutput(
+    'affected_core_components',
+    csv(
+      result.components
+        .filter(entry => entry.packageName === '@astryxdesign/core')
+        .map(entry => entry.component),
+    ),
+  );
+  appendOutput('affected_deferred_reasons', csv(result.visual.reasons));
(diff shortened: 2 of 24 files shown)


cixzhang (commented):

Thanks — centralizing scope is the right shape, but this head can still return “no scope” for real inputs. [`ROOT_METADATA`](https://github.com/facebook/astryx/blob/a0c034b7c6fccdf6ff5adea3dd7ed949bd6ac7d0/.github/scripts/lib/affected-scope.js#L17-L36) drops every nested `index.ts`, including the [Calendar hook barrel](https://github.com/facebook/astryx/blob/a0c034b7c6fccdf6ff5adea3dd7ed949bd6ac7d0/packages/core/src/Calendar/hooks/index.ts) used by Calendar and both date inputs. [The dependency list](https://github.com/facebook/astryx/blob/a0c034b7c6fccdf6ff5adea3dd7ed949bd6ac7d0/.github/scripts/lib/affected-scope.js#L39-L87) also misses `.nvmrc`, `pnpm-workspace.yaml`, and the shared theme `parseStyleKey.ts`; each produces no component and no deferral. Mixed component+global changes suppress the exact component lane too. Lab readiness then trusts [a comment marker](https://github.com/facebook/astryx/blob/a0c034b7c6fccdf6ff5adea3dd7ed949bd6ac7d0/internal/lab-readiness/automated.mjs#L97-L110) that can drift from the executable map.

Please keep the executable classifier as the one root source, and ensure every input in these declared dependency classes resolves to exact scope or explicit protected-main deferral, with the negative and mixed cases pinned.

[Reviewed by Robohands]

cixzhang (commented):

Thanks — the four named inputs now classify, and mixed changes preserve the exact booleans, but the exact/deferred contract is still split.

At this head, [`parseStyleKey.ts`](https://github.com/facebook/astryx/blob/d8d553774f79b80672035c0fbd8c2cf8a082d732/.github/scripts/visual-scope.mjs#L154-L171) becomes both exact and broad while yielding no exact component/story; the exact workflow path then calls `trusted-plan`, which refuses it as broad-only. A nested barrel still has no visual outcome, and the real RTL audit harness has no outcome anywhere. The root command is also not the owner: changing its roots leaves component classification unchanged, while readiness accepts a missing or mistyped helper through its fallback.

Please contract this to one executable classifier consumed by workflow and readiness, with every non-exempt path producing exact scope or explicit deferral. Pin the positive, negative, ambiguous, mixed, and legacy/current syntax cases.

[Reviewed by Robohands]

cixzhang (commented):

We asked for one executable scope owner; the named path cases now route correctly, but the contract still stops at classification.

At [the browser-job filter boundary](https://github.com/facebook/astryx/blob/fbf0e9b1dddfee059fca6a3d5e305901cf0b74f9/.github/workflows/ci.yml#L478-L484), a11y—and the matching RTL step—rebuild their component filter from `analysis.json`’s independent `newComponents + modifiedComponents` parser instead of consuming the component list emitted by `affected-scope.js`. Mutating only that shadow parser to drop Lab leaves the shared classifier reporting `Lab/Drawer` while both jobs receive no component; all 80 focused contract tests still pass. Lab readiness likewise accepts missing or mistyped current helper syntax by loading another checkout through its fallback.

Please make the executable classifier’s component set the a11y/RTL workflow input and make current-syntax readiness fail closed, while keeping legacy syntax as an explicit compatibility branch.

[Reviewed by Robohands]

cixzhang (commented):

We asked for every scope consumer to use one executable owner. The a11y/RTL filters and helper loading now do, but the consumer inventory still misses `pr-visual`.

[Stable visual regression](https://github.com/facebook/astryx/blob/a5676f81cab6fe9ecaa1352ff0b73aa2e8338e82/.github/workflows/ci.yml#L624-L650) still builds its component plan from `newComponents + modifiedComponents`. Mutating only that parser to return no components left all 13 consumer-contract tests green; a stable-story-only change likewise has exact story scope in `visual-scope.mjs` but falls into this job’s broad full-surface arm.

[Lab readiness](https://github.com/facebook/astryx/blob/a5676f81cab6fe9ecaa1352ff0b73aa2e8338e82/internal/lab-readiness/automated.mjs#L139-L149) also accepts `component-roots-broken` and `component-roots --unexpected`: its prefix match recognizes either, then executes the canonical command instead. Both returned Core + Lab in the fail-closed probe. The real RTL harness path still yields neither exact scope nor protected-main deferral.

Please include every PR visual consumer in the executable contract, make each drift mutant fail, and reject malformed current helper invocations.

[Reviewed by Robohands]

cixzhang (commented):

Thanks — one action now owns four PR scope jobs. Drift mutants fail, the exact/deferred matrix holds, and the tip is header-only.

[Reviewed by Robohands]

cixzhang (commented):

Thanks — the restack itself is byte-identical, but the one-source scope contract still has a silent shared-infrastructure lane.

[`inputKind()`](https://github.com/facebook/astryx/blob/1763ae1bc9ef4b5f8e1c8f802146da31869b0334/.github/scripts/lib/affected-scope.js#L49-L112) defers only named utilities. Changes to shared runtime such as `useFocusTrap.ts`, `useListFocus.ts`, or `sharedResizeObserver.ts` return no component and all three deferred flags false; `visual-scope.mjs` independently broad-defers those same paths, so visual coverage defers while a11y/RTL skip and the PR report can say no violations. The helper also exits 0 for `component-roots-broken` and `component-roots --unexpected`.

Please make the executable classifier total for shared runtime inputs, reject unknown/extra CLI arguments, and pin both classes.

[Reviewed by Robohands]

cixzhang (commented):

Thanks — we asked for one total scope owner and strict CLI. The four named shared-runtime roots and all four CI consumers now pass, but two parts remain.

[`inputKind()`](https://github.com/facebook/astryx/blob/341632258626f824ebc978fdf415d88322615fd0/.github/scripts/lib/affected-scope.js#L67-L115) omits real shared Core runtime under `i18n/` and the root `naming.ts`. `useDirection.ts` returns no component and no visual/a11y/RTL deferral, while [`visual-scope.mjs`](https://github.com/facebook/astryx/blob/341632258626f824ebc978fdf415d88322615fd0/.github/scripts/visual-scope.mjs#L59-L100) independently broadens it; a11y and RTL can still skip. Please make the shared classifier own the same non-component Core runtime class that visual scope treats as broad.

[`assertNoUnknownArgs()`](https://github.com/facebook/astryx/blob/341632258626f824ebc978fdf415d88322615fd0/.github/scripts/lib/affected-scope.js#L183-L193) ignores bare tokens, so malformed `classify` action wiring can succeed and pass readiness. Please reject every unconsumed token and pin that action-command mutation.

[Reviewed by Robohands]

cixzhang (commented):

Thanks — runtime behavior is now right: the shared i18n runtime and root naming source defer together, and both action arms reject a stray token. The regression still is not pinned.

[The malformed-CLI table](https://github.com/facebook/astryx/blob/185dc58b5ac39629d813789ebd6e2cccef7faaa9/.github/scripts/lib/affected-scope.test.mjs#L95-L135) gives every case `../bad/path.ts`. Replacing positional rejection with `continue` lets a valid Button path accept `stray-before`, while all 30 tests stay green. Appending a stray token to both action commands likewise leaves all 70 helper/workflow/readiness tests green; only executing the action fails.

Could you give each grammar case otherwise-valid input, assert its expected error, and make the action-command mutant red?

[Reviewed by Robohands]

cixzhang (commented):

Thanks — the Node action is now a real shell-free action; its metadata and end-to-end mutants fail. We previously asked for the positional regression to turn red, and that finding remains.

[The malformed-CLI table](https://github.com/facebook/astryx/blob/ffe38f07ba7fd0761d35f70f15185b8aaeb7e263/.github/scripts/lib/affected-scope.test.mjs#L129-L168) still gives every case `../bad/path.ts`. Replacing positional rejection with `continue` lets `classify stray-before --changed-files-stdin` accept valid `Button.tsx`, while all 31 helper tests stay green. That lets a future parser regression pass the suite meant to guard it.

Could each grammar case use otherwise-valid input and assert its intended stderr, so the positional mutant turns red?

[Reviewed by Robohands]

cixzhang (commented):

Thanks — `ad79df23` isolates every CLI error. Positional, action, and prior scope mutants fail; focused tests and exact-head CI are green.

[Reviewed by Robohands]

cixzhang:
Closing this proposal. The existing visual checks are working, and we are narrowing this effort to the gh-pages publication reliability stack only.

PR #5633, opened 2026-08-28
ci: route remaining gh-pages writers through publisher

cixzhang:
## Summary
- migrate PR preview deploy/redeploy, preview cleanup, gh-pages compaction, and vibe screenshot uploads onto the shared gh-pages publisher
- preserve preview subtree ownership, cleanup retention rules, and compaction whole-tree exclusivity
- add cross-writer tests proving preview, report, visual evidence, and baseline paths survive unrelated writers

Depends on #5631 (and #5629 below it).

## Test plan
- `actionlint -ignore 'label "2-core-ubuntu-arm" is unknown' -ignore 'shellcheck reported issue' .github/workflows/deploy-preview.yml .github/workflows/redeploy-preview.yml .github/workflows/cleanup-previews.yml .github/workflows/compact-gh-pages.yml .github/workflows/vibe-screenshots.yml .github/workflows/pr-comment.yml .github/workflows/visual-acceptance.yml .github/workflows/visual-acceptance-promote.yml .github/workflows/visual-baseline.yml .github/workflows/release-gate.yml .github/workflows/deploy.yml`
- `pnpm check:repo`


cixzhang (commented):

Thanks for moving these writers behind one publisher. Three boundaries still break it:

- [`deploy-preview.yml:123`](https://github.com/facebook/astryx/blob/33bbe50edd5cd3b665440b43cad3ba63a143b0ab/.github/workflows/deploy-preview.yml#L123-L131) and [`redeploy-preview.yml:104`](https://github.com/facebook/astryx/blob/33bbe50edd5cd3b665440b43cad3ba63a143b0ab/.github/workflows/redeploy-preview.yml#L104-L112) call a publisher they never check out, so every deploy fails; automatic preview also lost its artifact-ready guard.
- The publisher contract remains per-caller: cleanup, compaction, and vibe lack `actions: read`; [compaction](https://github.com/facebook/astryx/blob/33bbe50edd5cd3b665440b43cad3ba63a143b0ab/.github/scripts/lib/gh-pages-publisher.mjs#L1903-L1968) clears only one queue; [`deploy-report.ts`](https://github.com/facebook/astryx/blob/33bbe50edd5cd3b665440b43cad3ba63a143b0ab/internal/vibe-tests/src/deploy-report.ts#L157-L249) still pushes outside it.
- Automatic preview trusts `pr-meta.json` for PR/path identity instead of cross-checking the run like `pr-comment.yml`. [Cleanup’s retention check](https://github.com/facebook/astryx/blob/33bbe50edd5cd3b665440b43cad3ba63a143b0ab/.github/scripts/lib/gh-pages-publisher.mjs#L1651-L1666) also leaves a 58-day-old screenshot past its 30-day policy.

Could we restore trusted checkout/guards, give every caller the same contract, bind preview identity to the run/API, and route every write and cleanup through that authority?

[Reviewed by Robohands]

cixzhang (commented):

Thanks — the trusted PR resolver, permissions, retention, compaction cleanup, and direct-writer migration are now present. Two prior boundaries remain:

- The [automatic deploy step](https://github.com/facebook/astryx/blob/f07522f4127d9630f17f77d458bf075f3eef7019/.github/workflows/deploy-preview.yml#L174-L182) still has no readiness condition. Draft, failed, or missing-artifact paths set `ready=false`, but deployment still runs with an empty PR or directories; 43/43 checked-in tests pass because they only search earlier readiness strings.
- The [report publisher](https://github.com/facebook/astryx/blob/f07522f4127d9630f17f77d458bf075f3eef7019/.github/scripts/lib/gh-pages-publisher.mjs#L1822-L1828) copies new local screenshots first, then overwrites matching names with the old remote subtree. An isolated exact-head publish kept `remote-old` instead of `local-new`, reversing the [previous writer’s documented local-wins behavior](https://github.com/facebook/astryx/blob/33bbe50edd5cd3b665440b43cad3ba63a143b0ab/internal/vibe-tests/src/deploy-report.ts#L187-L210).

Could we gate deployment on trusted, artifact, and file readiness, then preserve remote screenshots only when the new report did not supply that path?

[Reviewed by Robohands]

cixzhang (commented):
Semantic verdict: approve

Both prior defects are fixed. Please assert `screenshots/same.png` is `new bytes` before landing—the current fixture leaves the precedence regression unguarded. Landing still waits for both parents and exact-head CI.

[Reviewed by Robohands]

PR #5639, opened 2026-08-28
fix(eslint-plugin): match no-react-introspection allowFiles on Windows

HelloOjasMutreja:
## Problem

`no-react-introspection`'s `allowFiles` escape hatch matches with `filename.includes(pattern)`. The configured patterns are written with forward slashes (`'Carousel/Carousel'`, etc, see `eslint.config.js`), but on Windows `context.filename` comes back with backslash separators, so the match never succeeds. The escape hatch is a silent no-op on Windows for every file it's supposed to exempt, including its own configured entries (`OverflowList/OverflowList`, `MetadataList/MetadataList`, `Carousel/Carousel`).

Ran into this directly: linting `Carousel.tsx` on Windows reports a `Children.toArray()` violation that the config explicitly allows for that exact file, and this blocks the local pre-commit hook for anyone on Windows touching it.

## Fix

Normalize backslashes to forward slashes before matching, same approach `no-raw-console-cli.js`'s existing `normalize()` helper already uses for the same class of problem.

## Testing

- Added a `valid` case to the rule's test file using a Windows-style (`D:\...`) filename.
- `npx vitest run internal/eslint-plugin-astryx/no-react-introspection.test.mjs` — 19/19 pass.
- `npx eslint packages/core/src/Carousel/Carousel.tsx` — clean after the fix (previously reported the false violation).

No visual change; this is lint tooling only.

### DIFF
--- a/internal/eslint-plugin-astryx/no-react-introspection.js
+++ b/internal/eslint-plugin-astryx/no-react-introspection.js
@@ -74,8 +74,13 @@ const noReactIntrospectionRule = {
     const options = context.options[0] || {};
     const allowFiles = options.allowFiles || [];
 
-    // Check if this file is explicitly allowed
-    const filename = context.filename ?? context.getFilename();
+    // Check if this file is explicitly allowed. Patterns are written with
+    // forward slashes; normalize before matching so this also works on
+    // Windows, where context.filename comes back with backslashes.
+    const filename = (context.filename ?? context.getFilename()).replace(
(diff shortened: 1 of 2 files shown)


cixzhang (approved):
Thanks, this restores Windows path exemptions without changing POSIX matching. Small test nit: use a slash-containing pattern so this case would fail before the fix.

[Reviewed by Robohands]

PR #5642, opened 2026-08-28
fix(Dialog): capture the external trigger before descendant mount effects

### LINKED ISSUE #5637: Dialog captures the wrong trigger when focusable content mounts on open
## Version

`@astryxdesign/core@0.5.0`

## Problem

A controlled `Dialog` can restore focus to `<body>` instead of its opener when its content is mounted only while `isOpen` is true.

`DialogHeader` runs a child `useEffect` that focuses its heading on mount. React runs that child effect before the parent `Dialog` effect that assigns `triggerElementRef.current = document.activeElement`. The dialog therefore captures its own heading (or another newly mounted autofocus child), not the external trigger. If the closed state removes/sanitizes that content, the captured node is detached before close restoration.

`DialogHeader` also focuses its heading when mounted under a closed Dialog because its `shouldAutoFocus` check only excludes inline mode, not `isOpen === false`.

## Minimal shape

```tsx
<button onClick={() => setOpen(true)}>Open</button>
<Dialog isOpen={open} onOpenChange={setOpen}>
  {open ? (
    <Layout header={<DialogHeader title="Review" />} content={<LayoutContent>...</LayoutContent>} />
  ) : null}
</Dialog>
```

1. Focus and activate `Open`.
2. Close with Escape or `onOpenChange(false)`.
3. Focus lands on `body`, not `Open`.

This shape is useful when a closed dialog must not retain a sensitive payload in the DOM.

## Expected

The Dialog captures the external active element before descendant mount/autofocus effects, and restores that element after the controlled `true -> false` transition. `DialogHeader` should not focus while the parent Dialog is closed.

Potential directions: capture the rising-edge trigger before passive descendant effects, expose `isOpen` in `DialogContext`, and make the header title the Dialog's supported `data-autofocus` target rather than running an unconditional mount effect.

HelloOjasMutreja:
## Problem

A controlled `Dialog` can restore focus to `<body>` instead of its opener when its content is mounted only while `isOpen` is true.

`Dialog`'s own effect read `document.activeElement` to capture the trigger for focus restoration on close. React runs child mount effects before the parent's own effects, so when `DialogHeader`'s mount effect focuses its heading first, `Dialog`'s effect reads that heading back as the "trigger" instead of the real external opener. If the closed state removes that content, the captured node is even detached by the time restoration runs.

## Fix

Capture the trigger during `Dialog`'s own render, on the `isOpen` false→true edge (`wasOpenRef`), instead of inside the effect. Render happens before any descendant can mount and run an effect, so this is unconditionally before `DialogHeader`'s autofocus. Guarded the capture against a missing `document` so an initially open `Dialog`/`AlertDialog` no longer throws during server rendering.

An earlier revision of this PR also gated `DialogHeader`'s `shouldAutoFocus` on `isOpen`, to stop a header mounted under a closed dialog from stealing focus for nothing (a separate part of #5637). That gate shifted when the header's mount-focus effect fires relative to the native dialog becoming `:modal`, which failed the `modal-dialog.focus.enters-after-modal` Chromium a11y contract on two other Dialog states (`labelled-described-default-title`, `explicit-descendant-focus`). It has been reverted here. **The closed-dialog header focus-theft half of #5637 is therefore not fixed by this PR** and remains open; only the wrong-trigger-capture half (and its own SSR-safety regression) is addressed.

## Testing

- `npx vitest run packages/core/src/Dialog packages/core/src/AlertDialog` — all pass, including a regression test for the trigger-capture fix and one asserting `renderToString` on an initially open `Dialog` no longer throws.
- `pnpm test:a11y-contract` (Chromium) — all 5 Dialog modal-dialog states pass.
- Fail-first verified: temporarily reverted just the source changes and confirmed the new tests fail with exactly the symptom in the issue (the dialog's own heading has focus instead of the opener button) before re-applying the fix.
- Typecheck and lint clean.

### DIFF
--- a/.changeset/dialog-trigger-capture-order.md
+++ b/.changeset/dialog-trigger-capture-order.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Dialog: focus now restores to the element that actually opened it, even when a DialogHeader (or other autofocusing content) mounts as the dialog opens.
+
+@HelloOjasMutreja
--- a/packages/core/src/Dialog/Dialog.test.tsx
+++ b/packages/core/src/Dialog/Dialog.test.tsx
@@ -969,4 +969,57 @@ describe('Dialog', () => {
       );
     });
   });
+
+  describe('trigger capture (#5637)', () => {
+    it('restores focus to the external trigger, not a child that autofocuses on open', () => {
+      const opener = document.createElement('button');
+      opener.type = 'button';
+      document.body.appendChild(opener);
+      opener.focus();
+
+      const {rerender} = render(
+        <Dialog isOpen={false} onOpenChange={() => {}}>
+          {null}
+        </Dialog>,
+      );
+      expect(opener).toHaveFocus();
+
+      // Content (including DialogHeader's autofocusing title) mounts only
+      // once isOpen flips true — the shape from the issue repro.
+      rerender(
+        <Dialog isOpen={true} onOpenChange={() => {}}>
+          <DialogHeader title="Review" />
+        </Dialog>,
+      );
+
+      // DialogHeader's own mount effect focuses the title; Dialog must still
+      // have captured `opener`, not the title, as the trigger to restore to.
+      expect(screen.getByRole('heading', {name: 'Review'})).toHaveFocus();
+
+      rerender(
+        <Dialog isOpen={false} onOpenChange={() => {}}>
+          {null}
+        </Dialog>,
+      );
+
+      expect(opener).toHaveFocus();
+      opener.remove();
+    });
+
+    it('does not autofocus a DialogHeader mounted while the dialog is still closed', () => {
+      const opener = document.createElement('button');
+      opener.type = 'button';
+      document.body.appendChild(opener);
+      opener.focus();
+
+      render(
+        <Dialog isOpen={false} onOpenChange={() => {}}>
+          <DialogHeader title="Review" />
+        </Dialog>,
+      );
+
+      expect(opener).toHaveFocus();
+      opener.remove();
+    });
+  });
 });
--- a/packages/core/src/Dialog/Dialog.tsx
+++ b/packages/core/src/Dialog/Dialog.tsx
@@ -459,8 +459,8 @@ export function Dialog({
   const titleId = useId();
 
   const dialogContextValue = useMemo(
-    () => ({isInline, titleId}),
-    [isInline, titleId],
+    () => ({isInline, isOpen, titleId}),
+    [isInline, isOpen, titleId],
   );
 
   // Consumer-provided labels always win over the DialogHeader default.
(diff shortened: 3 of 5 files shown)


cixzhang (changes requested):
Dialogs first mounted with `isOpen={true}` no longer restore focus through Dialog's owner. `useRef(isOpen)` starts at true, so the capture branch is skipped; the existing AlertDialog focus-return test fails at this head and after a clean merge with current main. Please capture the external invoker for both first-mount-open and closed→open paths while keeping the closed-header guard.

[Reviewed by Robohands]

HelloOjasMutreja:
Fixed in be601f93d.

The claim was accurate: \`wasOpenRef\` was seeded with \`isOpen\`, so a Dialog first rendered already open (AlertDialog's case) had \`wasOpenRef.current\` already true on that first render, and the rising-edge check (\`isOpen && !wasOpenRef.current\`) never fired. The trigger was never captured, so focus restoration on close had nothing to restore to.

Changed the seed to \`false\`, so the very first render of an already-open Dialog is still treated as a rising edge and captures the trigger, while a Dialog that re-renders while staying open (or stays closed) still doesn't re-capture. The closed-header autofocus guard from the original commit was untouched, since that part wasn't in question.

Verified locally: full \`Dialog\` suite (78/78) and \`AlertDialog\` suite (47/47) pass, including the focus-return test. Typecheck and lint on \`Dialog.tsx\` are both clean.

HelloOjasMutreja:
Fixed both.

The SSR crash: the render-time capture reads `document.activeElement`, which does not exist on the server. Guarded it with `typeof document !== 'undefined'`; the client's first render (hydration) still captures the trigger correctly. Added a regression test that stubs `document` away and asserts `renderToString` on an initially open Dialog does not throw.

Removed the stale `modal-dialog.focus.restored` known failure for the `conditional-content-focus-restoration` state, since this head fixes it.

While verifying against the full Chromium a11y contract, I also found the `DialogHeader` open-gate change from the prior round (`shouldAutoFocus = ... && dialogContext?.isOpen !== false`) broke two other states: `labelled-described-default-title` and `explicit-descendant-focus` started failing FR1 (focus entering the dialog before it reaches the browser's native modal state), because gating on `isOpen` shifted when the title's mount-focus effect fires relative to the dialog becoming `:modal`. The render-time trigger capture already prevents the header's own mount focus from being mistaken for the external trigger, so the gate was redundant for #5637 and actively harmful elsewhere. Reverted it; all five Dialog Chromium a11y states pass at this head, plus the existing 127 jsdom unit tests for Dialog/AlertDialog.

HelloOjasMutreja:
Updated the PR body: removed the `Fixes #5637` claim and its stale description of the DialogHeader gate, and added a section stating plainly that this PR only fixes the wrong-trigger-capture half of #5637 (plus the SSR crash it introduced along the way). The closed-dialog header focus-theft half stays unfixed here and #5637 should stay open for it.

PR #5643, opened 2026-08-28
feat(TextInput,TextArea): add autoComplete prop

### LINKED ISSUE #5638: TextArea cannot express the native autocomplete attribute
## Version

`@astryxdesign/core@0.5.0`

## Problem

`TextAreaProps` extends `BaseProps` (based on `React.HTMLAttributes`) rather than `React.TextareaHTMLAttributes`, and it does not declare an `autoComplete` prop. Consumers cannot express the native `autocomplete` attribute without mutating the forwarded textarea ref after mount.

```tsx
// TS2322: Property 'autoComplete' does not exist on TextAreaProps
<TextArea autoComplete="off" label="Rerun reason" value={reason} onChange={setReason} />
```

The current workaround is:

```tsx
useEffect(() => {
  ref.current?.setAttribute('autocomplete', 'off');
}, []);
```

That bypasses the component API and is easy to lose during refactoring.

## Consumer requirement

A controlled confirmation reason field uses a session-specific `name` and must opt out of browser/password-manager suggestion reuse. The value remains React-controlled; this request is only for the native autocomplete hint.

## Expected

Provide a typed, documented way for `TextArea` to set the native attribute, preferably an explicit `autoComplete?: React.TextareaHTMLAttributes<HTMLTextAreaElement>['autoComplete']` prop forwarded to `<textarea>`.

I saw the TextInput autocomplete finding in #718 was deliberately eliminated as outside input guidance. If the omission is intentional for TextArea too, documenting the supported alternative would still avoid ref-mutation workarounds.

HelloOjasMutreja:
Fixes #5638.

## Problem

`TextAreaProps`/`TextInputProps` extend `BaseProps`, which is `Omit<React.HTMLAttributes<T>, ...>` — the generic HTML attributes interface, not the element-specific `TextareaHTMLAttributes`/`InputHTMLAttributes` that carry `autoComplete`. Neither component could express the native `autocomplete` attribute without mutating the forwarded ref after mount.

## Direction

Posted a spec-proposal comment on the issue first (new public prop). The reporter confirmed: add it to both components together rather than one alone, since `TextInput` had the same gap and this shouldn't create an inconsistent contract between the two.

## Fix

Added a typed `autoComplete` prop to both `TextInputProps` and `TextAreaProps`:
- `TextInput`: `React.InputHTMLAttributes<HTMLInputElement>['autoComplete']`
- `TextArea`: `React.TextareaHTMLAttributes<HTMLTextAreaElement>['autoComplete']`

Neither component destructures it explicitly — both already spread `{...rest}` onto their native element, so declaring the type is sufficient for it to flow through unchanged. The controlled value is untouched either way.

Updated both `.doc.mjs` prop tables (English + Chinese variants, plus the flat `propDescriptions` map) to match.

## Testing

- `npx vitest run packages/core/src/TextInput/TextInput.test.tsx packages/core/src/TextArea/TextArea.test.tsx` — 161/161 pass (includes upstream's new `isReadOnly`/form-participation tests from a rebase; kept both, no conflict).
- Typecheck and lint clean (one pre-existing, unrelated `no-nullish-jsx-guard` warning confirmed present on upstream/main before this change).

No visual change — this is a native-attribute pass-through only.

### DIFF
--- a/.changeset/textinput-textarea-autocomplete.md
+++ b/.changeset/textinput-textarea-autocomplete.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] Add an `autoComplete` prop to TextInput and TextArea, forwarded to the native control unchanged.
+
+@HelloOjasMutreja
--- a/packages/core/src/TextArea/TextArea.doc.mjs
+++ b/packages/core/src/TextArea/TextArea.doc.mjs
@@ -256,6 +256,12 @@ export const docs = {
       description: 'Callback fired when the textarea loses focus.',
     },
     {
+      name: 'autoComplete',
+      type: 'string',
+      description:
+        'The native autocomplete attribute, forwarded to the textarea unchanged. Does not affect the controlled value.',
+    },
+    {
       name: 'xstyle',
       type: 'StyleXStyles',
       description:
@@ -507,6 +513,11 @@ export const docsZh = {
       description: '文本域失去焦点时触发的回调。',
     },
     {
+      name: 'autoComplete',
+      type: 'string',
+      description: '原生 autocomplete 属性，原样转发给文本域。不影响受控的值。',
+    },
+    {
       name: 'xstyle',
       type: 'StyleXStyles',
       description:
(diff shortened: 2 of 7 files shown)


cixzhang (approved):
Typed `autocomplete` now reaches both native controls without changing the omitted path; focused runtime, type, docs, and current-main checks pass.

[Reviewed by Robohands]

PR #5644, opened 2026-08-28
fix(SideNavItem): let a consumer aria-label win in collapsed mode

### LINKED ISSUE #5641: SideNavItem overwrites a consumer aria-label in collapsed mode
## Version

`@astryxdesign/core@0.5.0`

## Problem

`SideNavItem` accepts `aria-label` through `BaseProps`, and the expanded path forwards it through `...rest`. The collapsed path then spreads `...rest` followed by `collapsedAriaProps`, whose hard-coded `aria-label: label` overwrites the consumer value.

```tsx
<SideNavItem
  label="Home"
  aria-label="Open Home, 3 items need attention"
  icon={<HomeIcon />}
/>
```

Expanded accessible name: `Open Home, 3 items need attention`

Collapsed accessible name: `Home`

The same overwrite occurs in the collapsed item-with-children path, where `aria-label={label}` follows `...rest`.

This drops unread/attention context from icon-only navigation — exactly where the visible label is absent and the accessible label is authoritative.

## Expected

A consumer-provided `aria-label` should win in every render mode. The collapsed fallback should use `rest['aria-label'] ?? label` (and the item-with-children path should do the same), or the component should expose a dedicated accessible-label prop with consistent precedence.

Current consumer workaround is to pass the long accessible string as `label`, which also changes the expanded visible label and tooltip.

HelloOjasMutreja:
Fixes #5641.

## Problem

`SideNavItem` accepts `aria-label` through `BaseProps`, and the expanded path forwards it through `...rest` untouched. Both collapsed-rail paths spread `...rest` and then hardcoded `aria-label={label}` (or `'aria-label': label` in `collapsedAriaProps`) after it, silently overwriting any consumer-provided value — exactly where the visible label is absent and the accessible name is authoritative.

## Fix

`packages/core/src/SideNav/SideNavItem.tsx`:
- Item-with-children collapsed trigger: `aria-label={rest['aria-label'] ?? label}`
- Icon-only collapsed item's `collapsedAriaProps`: `'aria-label': rest['aria-label'] ?? label`

Falls back to `label` only when the consumer didn't provide one — the expanded path's existing behavior is unchanged.

## Testing

- Fail-first verified: reverted just the source change and confirmed both new tests fail with exactly the symptom in the issue (`aria-label` resolves to the plain `label` string instead of the consumer's value) before re-applying the fix.
- `npx vitest run packages/core/src/SideNav/SideNav.test.tsx` — 197/197 pass.
- Typecheck and lint clean.

No visual change — `aria-label` only, not rendered text.


cixzhang (approved):
Collapsed `SideNavItem` links, buttons, and submenu triggers now keep the accessible name supplied by the consumer; items without one still fall back to `label`.

Verified at this exact head with the focused SideNav suite, fail-first regression coverage for all three consumer-label paths, and Chromium accessibility-tree checks for the custom and fallback names. Current `main` does not already contain the fix, and the synthetic merge is clean.

[Reviewed by Robohands]

cixzhang (approved):
Exact-head review is clean against the current SideNav contract and WCAG 4.1.2. The added coverage now includes the original collapsed button seam, link and submenu paths, plus blank-label fallback.

PR #5645, opened 2026-08-28
feat(ProgressBar): add explicit contrast presentations

rubyycheung:
## Summary

- Add `contrast="standalone" | "supplemental"` to `ProgressBar`.
- Keep `standalone` self-contained with a terminal endpoint indicator.
- Render the endpoint as a true 10px circle matching the 10px track geometry and using the same semantic color as the bar.
- Require an equivalent visible value before consumers opt into `supplemental`.
- Keep indeterminate progress on the self-contained treatment and use the quieter supplemental treatment for disabled progress.
- Add docsite\/CLI blocks comparing contrast presentations, semantic standalone variants, and loading\/disabled states.\n- Document when to use standalone, supplemental, indeterminate, and disabled presentations, including when supplemental is not appropriate.
- Update component tests, API documentation, Storybook, the changeset, and generated theme-probe coverage.

## Why this is a component API

WCAG 2.2 SC 1.4.11 requires graphical parts needed to understand content to reach 3:1 against adjacent colors. A standalone progress graphic must carry the range information itself; a progress bar paired with visible equivalent text can use a quieter supplemental presentation.

The component cannot reliably inspect surrounding DOM to decide whether equivalent text exists, so the choice is explicit rather than inferred.

## Scope

This PR targets `main` and contains the reusable ProgressBar API. Neutral-theme token mappings and sandbox audit evidence remain separate from this API review in the Neutral theme stack (#5628 and #5649).

## Review focus

- `packages/core/src/ProgressBar/ProgressBar.tsx`: behavior, semantic endpoint marker, and theme selectors
- `packages/core/src/ProgressBar/ProgressBar.test.tsx`: standalone, supplemental, indeterminate, and disabled behavior
- `packages/core/src/ProgressBar/ProgressBar.doc.mjs`: public guidance and theming targets
- Storybook and CLI blocks: correct consumer usage and docsite examples

## Verification

- 58 focused ProgressBar tests pass
- 427 docsite tests pass
- focused ESLint passes
- core and docsite TypeScript pass
- generated docsite registries include the new examples
- generated theme-probe check passes
- repository pre-commit checks pass
- `git diff --check` passes

@rubyycheung


imdreamrunner (approved):

cixzhang (changes requested):
ProgressBar consumers cannot implement or verify this draft because it requires a future “sufficient standalone visual” without defining the presentation that satisfies it. The proposal currently changes no API or runtime behavior, and OQ1 leaves its only new presentation guarantee unresolved. Please choose one of two coherent scopes: (1) define the exact fill/track treatment, theme behavior, representative states, and measurable evidence needed to close the current gap; or (2) remove the unresolved future guarantee and record only today’s observable ProgressBar behavior and theming ownership. Do not make an undefined future treatment normative.

PR #5648, opened 2026-08-28
fix(chat): show composer focus ring for keyboard

rubyycheung:
## Summary
- add the shared focus outline when keyboard focus enters the ChatComposer editor
- suppress the composer-level ring for pointer focus
- leave focus indication on internal buttons instead of ringing the whole composer
- document the behavior and add a patch changeset

## Why
The elevated composer intentionally preserved its legacy low-to-medium shadow change on focus, but shadow depth is not a reliable WCAG focus indicator. WCAG 2.4.7 requires visible keyboard focus, and WCAG 1.4.11 requires an author-supplied focus indicator to reach 3:1 against adjacent colors.

This keeps elevation as visual depth and adds a separate, theme-controlled focus outline.

References:
- https://www.w3.org/WAI/WCAG22/Understanding/focus-visible.html
- https://www.w3.org/WAI/WCAG22/Understanding/non-text-contrast.html#related-focus

## Verification
- pnpm exec vitest run packages/core/src/Chat/ChatComposer.test.tsx packages/core/src/Chat/ChatComposerInput.test.tsx
- pnpm --filter @astryxdesign/core exec tsc --noEmit
- pnpm exec eslint packages/core/src/Chat/ChatComposer.tsx packages/core/src/Chat/ChatComposer.test.tsx
- pnpm check:changesets

@rubyycheung


rubyycheung:
Synced this branch with current `main`, including the current interaction-modality architecture from #5734.

Added explicit regression coverage for its last-input invariant:
- programmatic editor focus after keyboard input shows the composer ring
- programmatic editor focus after pointer input keeps the composer ring hidden

Local verification passed:
- `pnpm exec vitest run packages/core/src/Chat/ChatComposer.test.tsx packages/core/src/Chat/ChatComposerInput.test.tsx` (62 tests)
- `pnpm exec eslint packages/core/src/Chat/ChatComposer.tsx packages/core/src/Chat/ChatComposer.test.tsx`
- `pnpm --filter @astryxdesign/core exec tsc --noEmit`
- `pnpm check:changesets`
- `pnpm check:knowledge`

The updated Storybook preview should be used for the architecture record’s real-browser verification of keyboard focus, pointer suppression/clearing, internal-button ownership, and programmatic focus after both modalities.

cixzhang (changes requested):
Thanks, the keyboard-only composer ring works across keyboard, pointer, touch, programmatic focus, and internal-button ownership.

One thing still blocks this head: the new Chat RTL target never measures an assistant bubble. `core-chat--default` renders assistant responses as Markdown, so that selector matches zero and `pr-rtl` stays red. Please target two elements that both exist in the story, or add a dedicated fixture, so D2 produces a real result.

[Reviewed by Robohands]

cixzhang (approved):
Thanks, the prior RTL blocker is fixed: the Mixed Content target now measures both sender bubbles. Exact-head focus ownership also passes.

[Reviewed by Robohands]

PR #5652, opened 2026-08-28
feat(Drawer): recede a buried drawer so a stack reads as layered pages

freddymeta:
> **No longer stacked on #5550.** The previous head of this branch carried five of that PR's commits — `containerRef`, `modality`, `hasScrim` split, bounded `inert` — in a copy that predated #5550's own fix. The recede depends on none of it, so it is dropped: this PR is now the recede alone, on `main`. #5550 owns that review thread by itself. See "What changed since the last review".

## The gap

Open a second drawer over a first and the first is simply gone. Nothing says where the user came from, or how deep they are — the panel behind is covered by an identically-shaped panel, so a two-level flow reads exactly like a one-level flow that changed its contents.

EPS solves this in its own Sheet, and the effect is worth having here: the panels behind the front one recede, so the stack reads as layered pages.

## What it does

A drawer with drawers stacked on top **on its own edge** withdraws toward its closing edge and shrinks a little per level, keeping each leading edge visible beside the one in front, with rounded corners so it reads as a card behind rather than a clipped panel. The transform origin is pinned to the closing edge, so the shrink is a withdrawal *into* the page rather than a panel floating free of it. Both flip under RTL.

Nothing to opt into: it follows the stacking contract Drawer already documents. `hasStackRecede={false}` opts a drawer out.

## What changed since the last review

**1. The PR no longer carries #5550.** Both blocking points were about that work. Five commits on this branch were #5550's bounded/modality/scrim change, and they were a *stale* copy — they predated `c905415d` on `feat/drawer-bounded`, which is #5550's own fix for the inert bug reported here. So the bounded-modal boundary was never this PR's to fix: it is fixed on #5550, where the residual live-modality case is still open. This branch now shows only the recede.

The recede never needed any of it. It reads the LIFO registry, which is already on `main`; the only #5550 identifiers it touched (`anchoredSide`, `isTopLayerModal`) map straight onto `main`'s `side` and `hasScrim`. Diff against `main` went 7 files / +1451 → 4 files + a new spec / +490.

**2. Stack depth is scoped to one edge.** The global count was wrong, and it reproduces on `main` with no containers involved: a `side="start"` drawer and a `side="end"` drawer sit beside each other and never overlap, yet opening one made the other recede. `countDrawersAbove` now counts only entries on the same edge.

Escape order deliberately stays chronological across every open drawer — "the last thing I opened" is not an edge-local question, and it is the documented contract. One registry, two orderings, each answering the question it is for. Changing an open drawer's `side` updates its entry **in place** rather than re-registering, so a buried drawer cannot be promoted past the panel in front of it.

**3. The buried corner radius is now private, and one earlier claim was wrong.**

- `borderRadius` **is** in the guaranteed-property catalog, so a public `--drawer-stack-radius` fails the `INV10` admission bar on its face — every other Astryx component routes radius through a private `--_*` var. It is now `--_drawer-stack-radius`. Four public vars → three.
- The previous body measured the buried radius at **8px**. Re-measured, it is **10px**: `--radius-element` resolves to `0.625rem` under the shipped theme, and `8px` is only the raw token default. The old number was wrong.

**4. Adds `packages/lab/src/Drawer/Drawer.spec.md`.** `spec:AST-002` FR12 rejects a public-API addition whose canonical owning record is missing, and Drawer had none. Draft authority — review context, not policy. It carries the admission argument for each public concept, the invariants, and two open questions for a human.

## Scope: the Lab exemption does not apply

`architecture:component-theming-surface` INV1 exempts Lab components "until they opt into public component theming". Drawer opted in before this PR: `Drawer.doc.mjs` on `main` already declares `theming.targets: [{className: 'astryx-drawer', …}]`, and the guard landed in #5717 enrols it —

```
✓ lab/Drawer: participating components declare theming metadata
✓ lab/Drawer (Drawer/Drawer.doc.mjs docs): every rendered class is documented
```

So the theming-surface obligations are live here, and the admission argument below is required rather than optional.

## Themeable, not per-call

The geometry is a design decision for a product surface, not per call site, so it belongs to the theme. Three vars on the element carrying the `drawer` target — the same door `--spinner-diameter` opens:

| var | default | what it does |
|---|---|---|
| `--drawer-stack-peek` | `40px` | how far each buried level slides out from behind the one in front |
| `--drawer-stack-scale-step` | `0.04` | how much it shrinks per level, unitless |
| `--drawer-stack-min-scale` | `0.8` | floor, so a deep stack levels off |

**Admission (INV10 / `spec:AST-002/DEC-1`):** `transform` is outside the guaranteed-property catalog — the record explicitly excludes positioning and sizing — so no guaranteed CSS property can express a translate-and-scale. That is the gap INV10 admits a public var for. The precedent is `--spinner-diameter`, public for the same reason and listed in `derivedVarRegistry.test.ts` under "no standard CSS property maps onto these". The three are separate because how far, how fast and where to stop are three independent choices; one input carrying all three would violate FR16. A `stackRecede={{…}}` prop is rejected under FR1/FR4: it puts a system-wide design decision on every call site.

`--_drawer-stack-radius` is private under INV11 and promises nothing.

Depth is mirrored to `data-stack-depth` on the panel, so the state is inspectable and an app can hang its own rules off it.

## Two deliberate departures from the EPS implementation

**Depth comes from the stack registry, not from React nesting.** EPS threads a context up through physically nested sheets. Drawer's documented contract is the opposite — siblings stack, *never nest them* — so a nesting context would find nothing to walk, and EPS needs a `nestedDepth` escape hatch for exactly the flat-sibling case Drawer mandates. The LIFO registry already holds the order; it gains a subscription and `useSyncExternalStore` reads it, with a server snapshot of `0`. No escape hatch needed.

**The shrink is a ratio per level, not a px inset.** EPS insets the cross-axis edge by 16px per level, which has to be divided by the measured panel to become a `scale()` — that costs a `ResizeObserver`, and it makes the recede shallower on a tall screen than a short one. A unitless ratio is resolution-independent, needs no measurement, and is directly themeable.

## Measured

Chromium 1280×800, against the built `packages/lab/dist` served from `apps/storybook/dist`. The probe first asserts the lab rules are actually loaded (`--drawer-stack-peek` found in a live stylesheet) — without that an unstyled drawer reports radius 0 / transform none and every row below would be a false negative.

Three deep on the `end` edge, plus an unrelated `start` drawer:

| | `data-stack-depth` | transform | origin | radius |
|---|---|---|---|---|
| Level 1 | `2` | `matrix(0.92, 0, 0, 0.92, -80, 0)` | `400px 400px` (closing edge) | `10px` |
| Level 2 | `1` | `matrix(0.96, 0, 0, 0.96, -40, 0)` | `400px 400px` | `10px` |
| Level 3 (front) | absent | `matrix(1, 0, 0, 1, 0, 0)` | `200px 400px` (centre) | `0px` |
| **Filters (`side="start"`)** | **absent** | **`matrix(1, 0, 0, 1, 0, 0)`** | centre | `0px` |

The `start` drawer holds identity through every level — that is point 1, fixed. Escape then closed **Level 3**, the last opened, not the `start` drawer that was opened first: the scoping did not disturb Escape order.

Negative controls:

| control | result |
|---|---|
| `hasStackRecede={false}`, three deep | all three `matrix(1, 0, 0, 1, 0, 0)`, no `data-stack-depth` |
| RTL | translate flips to `+80` / `+40`, origin flips to `0px 400px` |
| theme override `peek: 80px`, `step: 0.1` | `scale(0.8)` — clamped by the floor — at `-160`, and `scale(0.9)` at `-80` |
| lab drawer CSS rules deleted at runtime | `transform: none`, radius `0px`, `--drawer-stack-peek` empty — the harness can report a null, and does not |

Unwind 3 → 2 → 1 → 0 returned every panel to identity.

## Tests

`pnpm exec vitest run packages/core/src packages/lab/src` — **316 files, 8953 tests, all pass.** Five new ones: depth by level, the unwind, the opposite-edge scope, the opt-out, and Escape surviving a live edge change. Reverting `countDrawersAbove` to a global count fails exactly the two scoping tests and no others.

Gates: `core build`, `core typecheck`, `core typecheck:docs`, `check:repo` all clean. `lint:strict` reports **85 warnings, 0 errors** — byte-identical to the count on a clean `main` worktree, so this change adds none.

Storybook: `Lab/Drawer` → `NestedStack` and `NestedStackNoRecede`.

No changeset: `@astryxdesign/lab` is `"private": true` and not released.

## For a human

- **OQ1** in the spec: if #5550 lands, two drawers on the same edge in *different* containers will not overlap either, and the stack key becomes `(host, side)` rather than `side`. I deliberately did **not** pre-build that — there are no containers on `main`, so the code would be unreachable and untestable. One-line registry change when the capability exists.
- **OQ2**: whether the three vars stay public at promotion. They are marked `experimental`.


cixzhang (changes requested):
Lab experimentation is appropriate here, but two interaction boundaries are broken: unrelated Drawers in separate containers recede each other, and two bounded modals in one host inert the front Drawer so focus, clicks, and Escape fail.

Please scope stack depth to one logical stack and keep the top bounded modal interactive. The experimental API and variables do not otherwise block.

[Reviewed by Robohands]

freddymeta:
Both points were right, and the second one told me something about this PR I had wrong: it was carrying #5550. Head is now `eef258aa`.

---

### 2. "two bounded modals in one host inert the front Drawer so focus, clicks, and Escape fail"

Correct, and **not this PR's to fix** — I was reproducing it because this branch carried a stale copy of #5550.

Five of the nine commits here were `feat/drawer-bounded`'s: `containerRef`, the scope/modality/scrim split, the bounded `inert` pass. They stopped at `4bd6e661` (28 Aug). #5550 has since moved on — `c905415d`, "stack bounded modals instead of inerting each other", is the fix for exactly the case you reported, and it is on that branch, not this one. So this PR was showing you a bug that its dependency had already fixed.

The recede never depended on any of it. It reads the LIFO registry, which is already on `main`; the only `feat/drawer-bounded` identifiers it touched were `anchoredSide` and `isTopLayerModal`, which map onto `main`'s `side` and `hasScrim`. So I have dropped that work from this branch rather than merging a newer copy of it: **#5652 is now the recede alone, on `main`.** Diff went 7 files / +1451 → 4 files + one new spec / +490.

That leaves the bounded-modal boundary entirely with #5550, where your 2 Sep review has the residual live-modality-change case still open. Nothing in this PR depends on it landing.

### 1. "stack depth is still global, so unrelated Drawers can recede each other"

Correct, and it survives the narrowing — it does not need containers to reproduce. A `side="start"` drawer and a `side="end"` drawer sit beside each other and never overlap, yet the global count made one recede when the other opened.

`countDrawersAbove` now counts only entries on the drawer's **own edge**. Measured, Chromium 1280×800, three `end` drawers plus one `start` drawer, against the built `packages/lab/dist`:

| | `data-stack-depth` | transform | origin | radius |
|---|---|---|---|---|
| Level 1 | `2` | `matrix(0.92, 0, 0, 0.92, -80, 0)` | `400px 400px` | `10px` |
| Level 2 | `1` | `matrix(0.96, 0, 0, 0.96, -40, 0)` | `400px 400px` | `10px` |
| Level 3 (front) | absent | `matrix(1, 0, 0, 1, 0, 0)` | `200px 400px` | `0px` |
| **Filters, `side="start"`** | **absent** | **`matrix(1, 0, 0, 1, 0, 0)`** | `200px 400px` | `0px` |

The `start` panel holds identity through all three levels.

**Escape order is untouched, deliberately.** I did not scope it: "the last thing I opened" is not an edge-local question, and chronological order across every open drawer is what Drawer documents. Measured — with the `start` drawer opened *first* and three `end` drawers on top, Escape closed **Level 3**, the last opened. One registry, two orderings, each answering the question it is for.

Your #5550 review asked whether one ordering could decide both, and the answer here is the opposite: they are two questions, so they get two readings of one ordering. What I did take from that review is the failure mode. Changing an open drawer's `side` updates its registry entry **in place** rather than re-registering — re-registering would push it to the end, which *is* the stacking order, so a buried drawer would become the one Escape closes. There is a test for that specific path (`keeps Escape on the last-opened drawer when a buried drawer changes edge`); it fails on a re-register.

---

### Two things I changed that you did not ask for

You said "the experimental API and variables do not otherwise block", so I did not touch them on your account. The records did.

**`--drawer-stack-radius` is now private.** `borderRadius` is in the guaranteed-property catalog in `architecture:component-theming-surface`, so a public var for a corner radius fails the INV10 admission bar on its face — and every other component in the repo routes radius through a private `--_*` var with a `theming.derived` entry. It cannot become a plain `borderRadius` mapping either: the drawer's resting radius is `0` and only the buried state is rounded, and one mapping cannot carry two values. So it is `--_drawer-stack-radius`, seeded from `--radius-element`, promising nothing. Four public vars → three.

The three that remain are argued rather than assumed: `transform` is outside the catalog, so no guaranteed property can express a translate-and-scale — which is the gap INV10 exists for. `--spinner-diameter` is the precedent, public for the same reason.

**Drawer now has a `.spec.md`.** `spec:AST-002` FR12 rejects an API addition whose canonical owning record is missing, and Drawer had none — no Lab component does. It is draft authority, so review context rather than policy, and it carries the admission argument, the invariants, and the verification map.

On the Lab exemption, since it decides whether any of the above was required: it does **not** apply here. Drawer opted into public theming before this PR — `theming.targets` is on `main` — and your #5717 guard enrols it:

```
✓ lab/Drawer: participating components declare theming metadata
✓ lab/Drawer (Drawer/Drawer.doc.mjs docs): every rendered class is documented
```

### One earlier claim of mine that failed measurement

The previous PR body reported the buried radius as **8px**. It is **10px** — `--radius-element` resolves to `0.625rem` under the shipped theme, and `8px` is only the raw token default. Corrected in the body.

### Controls

| control | result |
|---|---|
| `hasStackRecede={false}`, three deep | all three `matrix(1, 0, 0, 1, 0, 0)`, no `data-stack-depth` |
| RTL | translate flips to `+80` / `+40`, origin to `0px 400px` |
| theme override `peek: 80px`, `step: 0.1` | `scale(0.8)` — floor — at `-160`, `scale(0.9)` at `-80` |
| lab drawer CSS deleted at runtime | `transform: none`, radius `0px`, `--drawer-stack-peek` empty |

That last one is the harness check: without it a null result here would be indistinguishable from an unstyled drawer.

`vitest run packages/core/src packages/lab/src` — 316 files, 8953 tests, all pass. `core build`, `core typecheck`, `core typecheck:docs`, `check:repo` clean; `lint:strict` 85 warnings / 0 errors, identical to a clean `main` worktree.

### For you

Two open questions in the spec, both flagged `human-api`:

- **OQ1** — if #5550 lands, same-edge drawers in *different* containers will not overlap either, and the stack key becomes `(host, side)`. I deliberately did not pre-build it: there are no containers on `main`, so the code would be unreachable and untestable. One line when the capability exists.
- **OQ2** — whether the three public vars survive promotion. Marked `experimental`.

freddymeta:
Closing — dropping this one.

PR #5653, opened 2026-08-28
Redirect /home to the Astryx homepage

rubyycheung:
## Summary
- add a permanent docsite redirect from `/home` to `/`
- keep the redirect on the canonical Astryx origin

## Test plan
- imported `apps/docsite/next.config.mjs` and asserted the configured redirect
- `git diff --check`


rubyycheung:
Closing in favor of an internal Differential revision, per the requested Nest app workflow.

PR #5654, opened 2026-08-28
Add adaptive menu presentations

rubyycheung:
## Summary

- add `popover`, `bottom-sheet`, and `adaptive` presentation policies to `DropdownMenu`
- forward the policy through `MoreMenu` and support it in `ContextMenu`
- keep the adaptive resolver in neutral overlay infrastructure rather than under one component
- share one `MenuBottomSheet` frame across DropdownMenu and ContextMenu; MoreMenu inherits it through DropdownMenu
- keep component-specific behavior local: button opening and submenu navigation in DropdownMenu, right-click and long-press targeting in ContextMenu
- preserve focus after dismissal while suppressing pointer-origin focus rings
- add component stories, docsite examples, API guidance, tests, and a changeset

`adaptive` selects the BottomSheet at 768px and below when the primary pointer is coarse. The default remains `popover` for compatibility. There is no item-count heuristic.

This PR is stacked on #5395 so the adaptive behavior can be reviewed separately from the responsive popover fit changes.

## Architecture

`BottomSheet` remains the single primitive for dialog lifecycle, focus return, gestures, safe areas, scrolling, animation, and height. `MenuBottomSheet` owns only the shared menu-family frame and spacing. DropdownMenu and ContextMenu retain their distinct interaction controllers.

## Test plan

- `pnpm exec vitest run packages/core/src/DropdownMenu/DropdownMenu.test.tsx`
- `pnpm exec vitest run packages/core/src/ContextMenu/ContextMenu.test.tsx`
- `pnpm exec vitest run packages/core/src/MoreMenu/MoreMenu.test.tsx`
- `pnpm check:repo`


rubyycheung:
Fixed in 4364c00de. Data-driven ContextMenu bottom-sheet/adaptive presentations now use the same touch action-list and drill-in navigation as DropdownMenu, including a back affordance, focus transfer to the submenu heading, and reset on close. Added a regression test covering nested selection without opening a flyout. ContextMenu + DropdownMenu focused suites pass (144 tests), ESLint and core typecheck pass.

imdreamrunner (approved):

PR #5655, opened 2026-08-28
fix(Toast): pass dismiss to renderContent

cixzhang:
## Problem

`renderContent` injects a `DismissButton` component and maintains registration, fallback, warning, and hydration behavior around it. That makes a custom layout only partially responsible for its own controls.

## What

- Pass `dismiss` to `renderContent` and remove the injected component and fallback subsystem.
- Let custom content compose its own Astryx `Button` or pass the callback through nested components.
- Keep the default translated, themeable dismiss button unchanged.
- Update the typed API docs, translation, Storybook example, tests, and changeset.

## Risk

This corrects an unreleased API. Custom renderers now own their dismiss control; renderers without one are left unchanged. Toasts that do not use `renderContent` are unaffected.

## Testing

- `vitest run packages/core/src/Toast/ToastViewport.test.tsx packages/core/src/Toast/useToast.test.tsx` — 43 passed
- Strict ESLint on changed TypeScript and story files
- Pre-commit repository checks


cixzhang:
Semantic verdict: approve

Reviewed at head `0996c6787d1f78f7e960e4839526841c3a3faca8`.

The shape is right: handing over `dismiss` matches the existing `renderToken(item, onRemove)` precedent, and the removal is complete — no slot, provider, registration, dev warning or corner fallback survives anywhere in the source or the public API, and the root's `position: relative` went with it. The default toast keeps its translated, themeable close: asserted by test, and the visual gate reports 0 shots changed. Focused Toast tests 43/43, doc guards 24/24, `typecheck` and `typecheck:docs` clean, ESLint clean on the changed files. Amending the pending changeset in place is the right call for an unreleased API.

One note, not a blocker. `error` toasts never auto-hide and nothing dismisses on Escape, so a custom layout that omits a control leaves a toast the user cannot close — only the app can, through the `ToastDismissFn` that `showToast` returns. The prop docs say Astryx no longer injects a fallback, but not what that costs in that case, and the story demonstrates only the auto-hiding one. A sentence in the `renderContent` JSDoc naming the non-auto-hiding case would cover it.

[Reviewed by Robohands]

cixzhang:
/accept-visual 33190620102/1 Callback-only Toast API preserves the reviewed default and custom layouts; focused interaction and type checks pass.

PR #5656, opened 2026-08-28
fix(core): forward pass-through props on six more BaseProps components

harjothkhara:
## Summary

Six components publicly accepted `BaseProps` that never reached the DOM. This change forwards those props to their owned element and defines one shared targeting rule for inputs and controls.

For inputs with separate field and semantic elements:

- `xstyle`, `className`, `style`, `width`, and field-wide `hidden`, `inert`, `dir`, and `aria-hidden` affect the complete field presentation;
- `ref`, `data-testid`, remaining DOM/data/ARIA props, and event handlers target the primary semantic control or group;
- component-owned semantics and required behavior retain precedence, while consumer handlers compose according to the component contract.

## What changes

- Layout, MetadataList, NavHeadingMenuItem, RadioList, TimeInput, and TopNavMegaMenuItem now preserve their accepted pass-through props.
- RadioList’s `ref` and `data-testid` now identify its semantic `radiogroup`; its styling, width, and field-wide structural props remain on the complete Field.
- TimeInput forwards props on both typed and native-picker paths, composes caller IDs/ARIA, and preserves owned keyboard, focus, and blur behavior.
- The public-component API guidance and input-field family contract now define these targets explicitly.

## Compatibility

This is a breaking `@astryxdesign/core` change because RadioList’s `ref` and `data-testid` move from the outer Field wrapper to the radiogroup.

Consumers that need an outer measurement or query boundary should wrap RadioList in their own element and put that wrapper’s ref or test ID there. Keep RadioList’s ref and test ID for semantic-group access. This usage-dependent migration cannot be safely automated.

Standalone TimeInput now applies `xstyle`, `className`, and `style` to the complete Field instead of only the painted control wrapper. Existing layout styling needs no change; visual overrides aimed specifically at the painted control should move to the `time-input` component target in `defineTheme`. Grouped TimeInput keeps the member-wrapper target because InputGroup owns the surrounding field.

PowerSearch and MultiSelector remain separate follow-ups because they do not yet have one settled semantic target.

## Testing

- 260 focused tests passed across all seven affected suites.
- `@astryxdesign/core` typecheck passed.
- ESLint passed with no new errors; one pre-existing MetadataList warning remains.
- Knowledge, Changeset, formatting, and repository pre-commit checks passed.

Refs #5254

### DIFF
--- a/.changeset/baseprops-passthrough-batch-four.md
+++ b/.changeset/baseprops-passthrough-batch-four.md
@@ -1,0 +1,13 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Forward pass-through props on six more BaseProps components (Layout, MetadataList, NavHeadingMenuItem, RadioList, TimeInput, TopNavMegaMenuItem)
+
+Six components declared `extends BaseProps` but destructured a closed set of props and never spread a rest object, so everything they did not name — `aria-*`, `id`, `data-*`, and event handlers — type-checked, was accepted, and silently never reached the DOM. This is the same defect class #5254 reported and #5288, #5493, and #5563 fixed in three batches; these six were not in that issue's list (it was compiled against the 0.4.5 tarball) and survived the sweep.
+
+Each component now forwards the rest object to the element that owns its contract: the `Layout` and `MetadataList` root, the `NavHeadingMenuItem` and `TopNavMegaMenuItem` item element (both drawer and default modes, including consumer `className`/`style`/`xstyle` composition and drawer-mode `tabIndex`), the `RadioList` radiogroup, and the `TimeInput` input on both its typed and native-picker paths. Where the component already owns a handler for the same event, the handlers are composed with `composeEventHandlers` component-first, so owned behavior (focus normalization, arrow-key stepping, blur reconciliation) always runs and cannot be cancelled by a consumer `preventDefault()`; a consumer handler runs after it, and does not observe events the owned handler consumed (the stepping arrows on the typed path; the native-picker path owns no keydown, so a consumer `onKeyDown` always fires there). A caller `id` is honored via `id ?? generatedId` (the `Selector` precedent), and a caller `aria-describedby` is composed additively with the component-owned description ids. Other contract-owned attributes (`role`, computed `aria-labelledby`, the roving `tabIndex`) stay after the spread so a pass-through cannot clobber them.
+
+`PowerSearch` and `MultiSelector` have the same shape and are deliberately left out: `PowerSearch` is a composite of a Tokenizer, a listbox, and an edit popover where the pass-through target is an API decision, and `MultiSelector` should follow the `Selector` trigger-forwarding precedent in a change of its own. Both are noted in the PR rather than guessed at here.
+
+@harjothkhara
--- a/packages/core/src/DateTimeInput/NativeTimeSegment.tsx
+++ b/packages/core/src/DateTimeInput/NativeTimeSegment.tsx
@@ -40,6 +40,7 @@ import {
   formatISOTime,
   isTimeInRange,
   parseISOTime,
+  composeEventHandlers,
   type ISOTimeString,
 } from '../utils';
 import {nativePickerSegmentStyles as styles} from './nativePickerSegmentStyles';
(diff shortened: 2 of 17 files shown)


cixzhang (changes requested):
Builders using `TimeInput` on touch devices would still lose valid attributes and handlers once this is rebased.

Current `main` now defaults `TimeInput` to a native picker on coarse pointers (#5811), with a separate `<input type="time">` path. This head forwards `...restProps` and composes handlers only on the old text-input path, and GitHub reports it conflicting with `main`. So `<TimeInput nativePicker="touch" data-tracking="…" onFocus={…}>` would work on desktop and silently fail on the active touch input—the same contract hole this PR intends to close.

Please rebase and preserve rest-prop precedence plus handler composition on both current `TimeInput` paths, with one focused test per path. The other five component fixes are clean.

[Reviewed by Robohands]

cixzhang (changes requested):
Builders can still get behavior that contradicts the props they passed:

- `Layout` and `MetadataList` spread caller data after `themeProps`, so `data-height`, `data-columns`, or `data-orientation` can overwrite the component-owned state that themes consume. Put rest props before owned theme metadata and cover collisions.
- `RadioList` and both `TimeInput` surfaces compose consumer handlers first. `preventDefault()` then cancels required focus normalization, arrow stepping, or blur reconciliation, although these APIs promise no cancellation. Keep owned behavior unconditional and add cancellation tests.

The updated native TimeInput path and ordinary forwarding checks are otherwise clean.

[Reviewed by Robohands]

harjothkhara:
The requested changes are addressed in 54dae36 with no review threads remaining; the only red job is the documented soft-gated RTL applicability check, so could someone re-review the current head?

cixzhang (approved):
Approved at c0e6b6d0fcfe. The input target contract is now explicit: field styling and field-wide structure stay with the complete field presentation, while ref, test ID, remaining DOM/ARIA props, and handlers target the semantic control or group. RadioList and TimeInput match that decision, owned semantics retain precedence, and the breaking RadioList target migration is documented. Focused tests, core typecheck, knowledge checks, changeset validation, and local review pass.

cixzhang (approved):
Approved at f6f44d3d5fa3. The input target convention is recorded in the canonical API and input-family contracts. RadioList and TimeInput follow it across standalone, grouped, typed, and native-picker paths; field-wide structural props remain on the complete field presentation, semantic props target the control/group, and both breaking migrations are documented. Focused tests, direct xstyle assertions, core typecheck, lint, knowledge checks, and changeset validation pass.

cixzhang (changes requested):
The updated target contract exposes three remaining issues on this head:

- **Owned input semantics can still be replaced or escape field-wide hiding.** `TimeInput` forwards caller `role` and `aria-label` onto both input surfaces, so they can replace the control's native role or its required visible label; the native-picker test currently blesses that override. `RadioList` likewise forwards `aria-disabled` and `tabIndex` without reconciling them with the group's disabled and roving-focus behavior. In grouped `TimeInput`, a status update also still calls the global announcer when `hidden`, `inert`, or `aria-hidden` removes the field. Please filter or reconcile these collisions and cover both TimeInput surfaces plus the RadioList group and hidden grouped-status state.

- **An item handler can cancel the menu's required keyboard model.** `NavHeadingMenuItem` now forwards `onKeyDown` on the child. If that handler calls `preventDefault()`, the event reaches `NavHeadingMenu` already canceled, so its composed handler skips arrow navigation, typeahead, Enter/Space activation, and Escape. Keep the parent menu behavior unconditional from item-level pass-throughs and add a cancellation regression test.

- **The TimeInput styling migration is not accurate in horizontal-label layouts.** Standalone styling now lands on `Field`, but that element becomes `display: contents` under horizontal `FormLayout`, so per-instance box styles such as margin, padding, background, and border no longer affect either the complete field or the painted control. The changeset currently says layout styling needs no change, and the updated family spec still lists TimeInput as dropping its advertised surface. Please provide a real target for that state or document/route the containing-layout exception accurately, then remove the stale TimeInput gap.

harjothkhara:
@cixzhang Got it. Should I remove RadioList and TimeInput from this PR and handle the input BaseProps family spec in a separate PR?

Issue #5403, opened 2026-08-24
[Bug] Kbd does not normalise the key aliases useHotkeys accepts


Astro-Han:
### Description

`useHotkeys` and `Kbd` take the same `'+'`-separated combo string, but each keeps its own key vocabulary. `useHotkeys` resolves aliases through `KEY_ALIASES` (`esc`, `space`, `up`, `down`, `left`, `right`, `return`, `plus`) and accepts `mod`/`ctrl`/`meta`/`alt`/`shift` as modifiers. `Kbd` resolves through `KEY_DISPLAY` and `KEY_LABEL`, a different set, and falls back to `key.toUpperCase()` for anything it does not recognise. That fallback is written for single letters (`k` → `K`), so a multi-character alias comes back as an uppercased word.

Rendered on macOS:

| combo | glyphs | aria-label |
|---|---|---|
| `mod+k` | ⌘ K | Command + K |
| `shift+enter` | ⇧ ↵ | Shift + Enter |
| `meta+k` | META K | META + K |
| `esc` | ESC | ESC |
| `return` | RETURN | RETURN |
| `space` | SPACE | SPACE |

The rendered key is never the wrong key, so this is a missing normalisation rather than a wrong hint. The cost lands hardest on the accessibility side: `KEY_LABEL` exists because glyphs announce meaninglessly, and "META" or "ESC" is not much better than a glyph.

The four cases are not one problem.

`esc` and `return` look like a plain gap. `KEY_DISPLAY` already holds `escape` → `Esc` and `enter` → `↵`, and `useHotkeys` already collapses both aliases in `KEY_ALIASES`. `Kbd` does not run the same step.

`meta` and `space` need a decision I don't think a contributor should make. `meta` is ⊞ on Windows, not ⌘, so whether it branches on platform the way `mod` does is an API question. `␣` vs `Space` is a design choice.

### Reproduction

```tsx
useHotkeys([{keys: 'return', onPress: doThing}]);  // fires on Enter
<Kbd keys="return" />                              // renders "RETURN"
```

Same for `meta`, `esc`, and `space`.

### Astryx Version

`@astryxdesign/core@0.5.0`, and current `main`.

### Environment

Any browser and OS. Not platform dependent.

---

*Diagnosis and draft prepared with Claude Code; reviewed and verified by the contributor before filing.*

PR #5657, opened 2026-08-28
fix(core): normalize Kbd esc and return aliases

harjothkhara:
## Summary

Kbd now normalizes the unambiguous `esc` and `return` aliases already accepted by `useHotkeys`, so their visual glyphs and accessible names match `escape` and `enter`. This is a partial fix for #5403; `meta` and `space` remain open design/API decisions.

## Case

`useHotkeys` already maps `esc` and `return` to the corresponding `KeyboardEvent.key` values (`useHotkeys.ts:64-74,132-133`), but Kbd previously sent every token directly through its display and label fallbacks (`Kbd.tsx:89-120`). As a result, `esc` rendered and announced as `ESC`, while `return` rendered and announced as `RETURN`.

## Fix

Kbd now normalizes the two aliases before display and accessible-label resolution (`Kbd.tsx:79-82,173-190`). The touched lookup tables use `Map`, keeping arbitrary key names on the uppercase fallback instead of accidentally resolving inherited object properties (`Kbd.tsx:64-120`), and occurrence-qualified React keys preserve every badge when canonical and alias spellings repeat (`Kbd.tsx:173-205`). Component docs and the CLI/Storybook showcases now demonstrate the aliases (`Kbd.doc.mjs:12-16`, `KbdModifierCombos.tsx:23-31`, `Kbd.stories.tsx:74-87`).

## Scope

This intentionally handles only `esc` and `return`. It does not choose platform behavior for `meta` or a glyph for `space`; those remain tracked in #5403. Existing empty-combo-segment behavior and the separate `useHotkeys` lookup implementation are parser-hardening follow-ups, not part of this rendering fix.

## Testing

- `pnpm exec vitest run packages/core/src/Kbd/Kbd.test.tsx packages/core/src/hooks/useHotkeys.test.ts` — 2 files, 36 tests passed.
- `pnpm -F @astryxdesign/core typecheck` — passed.
- `pnpm -F @astryxdesign/core typecheck:docs` — passed.
- `pnpm -F @astryxdesign/core build` — 581 files compiled and verified.
- `pnpm check:repo` — all repository integrity checks passed.
- ESLint passed for the changed TypeScript and TSX files.
- Negative control on `origin/main`: the alias regression cases failed because `esc` rendered/announced as `ESC` and `return` as `RETURN`. With the fix, standalone, uppercase, combo, semantic-equivalence, repeated-key, and accessible-name cases pass.
- Hardening red/green: before the source hardening, repeated/case-equivalent tokens emitted duplicate React-key warnings and `constructor`/`__proto__` failed through prototype-backed lookup; the committed regression tests now pass.

Refs #5403

### DIFF
--- a/.changeset/kbd-key-aliases.md
+++ b/.changeset/kbd-key-aliases.md
@@ -1,0 +1,9 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Kbd: render the `esc` and `return` aliases with the same glyphs and accessible names as `escape` and `enter`
+
+This partial fix for #5403 normalizes the two unambiguous aliases already accepted by `useHotkeys`. Kbd now renders `esc` as `Esc` with the accessible name `Escape`, and `return` as `↵` with the accessible name `Enter`. Its lookup tables now also avoid prototype-chain collisions when rendering arbitrary key names. The platform-specific rendering contract for `meta` and display choice for `space` remain unresolved in #5403 pending separate API and design decisions.
+
+@harjothkhara
--- a/apps/storybook/stories/Kbd.stories.tsx
+++ b/apps/storybook/stories/Kbd.stories.tsx
@@ -12,7 +12,7 @@ const meta: Meta<typeof Kbd> = {
     keys: {
       control: 'text',
       description:
-        'Keyboard shortcut string. Use "+" to separate keys. Special keys: mod, ctrl, alt, shift, enter, backspace, escape, tab, up, down, left, right.',
+        'Keyboard shortcut string. Use "+" to separate keys. Special keys: mod, ctrl, alt, shift, enter, backspace, escape, tab, up, down, left, right, plus. Aliases: esc for escape and return for enter.',
     },
   },
 };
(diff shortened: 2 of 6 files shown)


ernestt (approved):

rubyycheung (approved):
Looks good from design. Normalizing  and  makes their visual output and accessible names consistent with the canonical  and  keys, without changing the established Kbd presentation.

rubyycheung:
Clarification: normalizing `esc` and `return` makes their visual output and accessible names consistent with the canonical `escape` and `enter` keys.

harjothkhara:
Friendly follow-up: 5ea4b4f has exact-head code and design approvals with all automated checks green; visual acceptance and review-required are the only remaining manual gates.

cixzhang:
/accept-visual 33191992132/1 The esc and return aliases intentionally match the established escape and enter glyphs across all changed Kbd frames and supported themes.

cixzhang (approved):
Thanks — `esc` and `return` now match the established visual and accessible output of `escape` and `enter`, while unusual key names stay on the documented fallback. Focused tests, visual evidence, and exact-head CI are clean.

PR #5659, opened 2026-08-28
feat(core): let Stepper collapse itself in narrow containers

ernestt:
## Summary

A horizontal `Stepper` had no answer for narrow widths, so every consumer hand-rolled the same fallback: a media query, a `ProgressBar`, and a line of "Step 2 of 4" text standing in for the component. Four wizard templates carried a version of it, each with its own breakpoint.

The Stepper now measures its own frame and collapses once a step has under ~112px to work with. Labels give way to a bare track, and the current step is named in a row beneath it, flanked by prev/next controls when `onStepClick` makes them meaningful.

Measuring the frame rather than the viewport means the breakpoint follows the step count, so a stepper collapses on the width it actually has rather than the width of the window around it. Four steps hold out to ~448px; seven need ~784px.

**The two indicator positions collapse differently, deliberately.** `separated` gives up its click targets along with its labels, since all that would be left to aim at is a 4px bar — the paired controls become the way through. `on-track` keeps its indicators as a row of pressable nodes on the rail, which is the better affordance and costs nothing to leave in place.

**No child introspection.** Steps register with the parent to report the count, and the active step portals its own summary into a slot the Stepper provides. Nothing reads `child.props` or walks `Children`.

**The sequence stays whole for screen readers** at every width — collapsed steps keep visually hidden names, so the ordered list is never silently shortened.

Two fixes the collapse work surfaced, folded in because they only matter here:

- Step labels hold to one line and ellipsize instead of wrapping and breaking mid-word, so a row of steps keeps one height and the track under it stays straight. The full label still reaches the accessible name.
- The gap between connector segments is now `--spacing-1`, matching the connector's own thickness, so the track reads as one dashed line at any theme scale.

## Scope

Core only. The wizard templates that motivated this still carry their hand-rolled fallbacks and are being cleaned up one PR per template. An unrelated `LayoutContent` padding fix in my working tree is deliberately left out.

## Test plan

- [x] 58 Stepper tests pass, including a new block covering the collapse: labels dropping, the summary appearing with the current step's name and description, vertical steppers staying unaffected, the accessible sequence staying whole, controls bounding at both ends, collapsed steps leaving the tab order, and on-track nodes staying pressable.
- [x] `pnpm -F @astryxdesign/core typecheck` and `pnpm -F @astryxdesign/storybook typecheck` clean.
- [x] Lint clean on the changed files.
- [x] Verified in a browser at 1280 / 900 / 480 / 380px against a real wizard, driving the flow with both the form's own Continue and the stepper's controls, and confirming an errored step surfaces in the collapsed summary.
- [x] Truncation measured live: every label reports one line, with clipping starting exactly when its slice gets too narrow.
- [x] Gap measured live at 4px.
- [ ] Reviewer: the new `Narrow — Collapsed Track` story exercises all four cases side by side.

## Known gap, not addressed here

`status` never recolors the connector — progress is the bar's channel, status is the indicator's. That holds when every step shows an indicator, but collapsing removes indicators for every step except the current one, so an error on a step you have navigated *away* from is invisible while collapsed. In the labelled layout that step keeps its glyph. Worth a follow-up; I did not want to quietly break the channel separation to paper over it.

## Screenshots

**Separated, wide — unchanged.** Four labelled steps, icons as indicators.

separated wide

**Separated, 380px — collapsed.** Labels and click targets gone, four bars left, the current step named underneath between two flat controls. Previous is disabled on step 1.

separated narrow

**Separated, 380px — error on the current step.** The summary carries the error glyph; Previous is now live.

separated error

**On-track, wide.**

on-track wide

**On-track, 380px — collapsed differently, on purpose.** The numbered nodes stay on the rail and stay pressable; all four remain hit-targets. Only the labels go.

on-track narrow

**Label truncation.** Captured with deliberately long labels. Each holds one line and ellipsizes, so the row keeps one height and the track stays straight; short labels are untouched.

truncation


ernestt:
/accept-visual 33198611074/1 4px segments the individual steps more clearly instead of one continuous line

ernestt:
Release-process check: .changeset/stepper-narrow-collapse.md uses the correct patch bump, but its first body line must end in (#5659) and its second body line must be the contributor handle @ernestt; the handle is currently after the release notes.

ernestt:
Disable the on track interactivity for the collapsed version.

ernestt:
Addressed the three compact-mode review findings in 094c14fb2001d389fe6ef5825ac143bdef59de3b:

- Step content remains mounted (and hidden) across compact resizing in both separated and on-track layouts, preserving DOM identity and input state.
- Summary controls move to the nearest enabled step and disable when no enabled target remains in that direction.
- Responsive measurement now observes the public styled/ref-forwarding `<ol>` root rather than the frame wrapper.

Verified with 76 focused Stepper tests, targeted ESLint and Prettier checks, the core typecheck and build, and `pnpm check:repo`.

imdreamrunner (approved):

imdreamrunner:
lgtm! @ernestt i will start merging some other PRs for the API change and small fixes onto this branch.

imdreamrunner:
/accept-visual 33935624478/1 minor spacing decrease, unclear reason. non blocking

imdreamrunner (approved):
Spec approved for current head 42364dd91f13474f78721964824e368d265cac50.

PR #5660, opened 2026-08-28
feat(templates): add the checkout wizard page template

ernestt:
> **Stacked on #5659.** Base is `core-stepper-narrow-collapse`, not `main`. Retarget to `main` once #5659 lands.

## Summary

Adds a checkout wizard template with a running order summary and closes the review findings in the template itself:

- every forward path—including Stepper and compact controls—validates each crossed step and focuses the first invalid field;
- Pay validates payment placeholders and records a clear local completion outcome without sending payment;
- postal labels, examples, and validation adapt for the United States, Canada, and the United Kingdom;
- invalid promo codes show a terminal result instead of repeating “Press Apply”;
- responsive layout follows the template host width rather than the viewport;
- the narrow summary renders before the form in both DOM and visual/keyboard order;
- metadata uses the supported `Form - Checkout` category.

The existing CLI changeset remains the release entry for this new template.

## Test plan

- [x] Checkout-wizard focused contract tests: 6 passed.
- [x] Template-doc typecheck and direct strict TypeScript check for `page.tsx`.
- [x] Targeted ESLint with zero warnings, Prettier check, and `git diff --check`.
- [x] `pnpm lint:strict` (repo checks passed; existing repository warnings only).
- [x] `pnpm build`.
- [ ] `pnpm test` completed with two unrelated failures in untouched CLI discovery tests on this macOS checkout: lowercased component-name resolution and `useMediaQuery.doc.mjs` path casing. The focused checkout tests pass.

PR #5672, opened 2026-08-28
feat(templates): add the vertical wizard page template

ernestt:
Fifth of the wizard family. Stacked on [#5659](https://github.com/facebook/astryx/pull/5659) for the Stepper's one-line label truncation; siblings [#5660](https://github.com/facebook/astryx/pull/5660) and [#5664](https://github.com/facebook/astryx/pull/5664) sit alongside it.

## What differentiates it

The steps live in a rail beside the form rather than above it. That suits steps that are long or unequal in size: the rail stays legible at five entries carrying descriptions, and the form beside it can run past a screenful without the progress scrolling away.

Wide

The rail's counterpart is a guidance panel on the other side, holding a summary of the current step until a field takes focus, then swapping to what that field affects:

Field guidance

It is deliberately **not** a live region. The panel changes on every focus move, and announcing it each time would talk over the field's own label and description on every tab stop. It is a labelled landmark instead, so a screen reader user can reach it on demand — which is also why the text says what a field *affects* rather than repeating the hint already attached to the field.

## Giving up space in stages

Guidance goes first, at 1320px, since it is the only panel whose content is also reachable from the fields themselves:

Medium

The rail goes at 1000px, rotating into a horizontal stepper above the form so progress survives the loss of the column:

Narrow

The last step reviews everything as an icon-led metadata list:

Review

A blocked Continue sends focus to the first field needing attention, the same behaviour as [#5664](https://github.com/facebook/astryx/pull/5664), so the footer never carries a second copy of a message the field is already showing:

Validation

## Audit

**92/100, grade A**, under template rubric 1.3.

| Category | Score |
|---|---|
| Component purity | 25/30 |
| Icon purity | 15/15 |
| Custom CSS | 12/15 |
| Layout & structure | 15/15 |
| Doc metadata | 10/10 |
| Image handling | 5/5 |
| Code quality | 10/10 |

The first audit came in at 85 on ten custom declarations, and the cover image was carrying most of them. Three rounds of that turned out to be the system's own job:

- **The cover framing** was `width`, `aspectRatio`, `objectFit`, `borderRadius` and `border` written by hand. `Card` takes a width and draws the border and radius — and clips to it — while `AspectRatio` takes the ratio and the object fit. Five declarations to zero, and the radius now comes from the token rather than being named.
- **The 640px measure** was `width: 100%` / `maxWidth` / `marginInline: auto`, which is character-for-character what `Layout`'s `contentWidth` emits. That prop applies them to the middle row, which here holds the rail and the guidance panel as well as the form, so it would cap all three columns together instead of the one that should be capped. `Center` plus `Stack`'s `maxWidth` says the same thing about the right box. Three to zero.

Three declarations remain, each without a prop to use instead: a zero flex basis so the dropzone column can shrink, a height so the dropzone ends level with the cover (`FileInput` exposes `width` but not `height`), and the footer lean described below.

The 5 points still on purity are the single `` — the rubric's own listed exception, since there is no Image component and the `src` is a checked-in `/template-assets/` path.

### A defect the refactor exposed

Reworking the measure made it visible that the footer spans the page and centred on the page, which only matches the form when the panels are symmetric. With both up or both down it was fine, but in the middle tier — rail up, guidance gone — the buttons sat **142px** left of the fields they submit. That was there before this refactor; it just was not obvious until the two boxes were expressed the same way.

The footer now leans with the column. Measured footer-to-form centre offset is **0px** at 1440, 1360, 1319, 1200, 1100, 1010, 999, 900, 760 and 520 — both sides of both breakpoints.

Three other declarations came out during the first audit rather than being defended: `position: sticky` on the guidance panel was a no-op, confirmed by measuring (the form scrolls 478px while the panel heading holds at 81px, because the panel is its own layout cell and never scrolls), and `minWidth: 0` duplicated a reset `StackItem` already applies. `template.doc.mjs` also had a stale sentence claiming the rail collapses to a counter, which the staged responsive work had replaced.

## Test plan

- [x] All five steps render; validation gates each advance
- [x] Blocked Continue moves focus to the first invalid field (verified: lands on "Book title")
- [x] Guidance swaps to field help on focus and back to the step summary on blur
- [x] Responsive tiers verified at 1440 / 1150 / 760px — vertical+guidance, vertical, horizontal
- [x] Footer aligns with the form measure at 10 widths spanning both breakpoints, 0px offset
- [x] Dropzone ends level with the cover image
- [x] Zero console errors at every width tested
- [x] ESLint clean, no dead code
- [x] `tsc` clean against the docsite project

One follow-up commit early on: the guidance heading carried `type="label"`, which is not a `HeadingType` — the prop only holds the display scale, so it sizes a heading *up*. The lookup missed, no size style applied, and it inherited whatever ambient size it sat next to. It looked right by accident and only surfaced in the docsite type check, since Vite's dev server does not typecheck. Now `level={5} accessibilityLevel={2}`: label-sized, level 2 in the outline.

### DIFF
--- a/.changeset/vertical-wizard-template.md
+++ b/.changeset/vertical-wizard-template.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/cli': patch
+---
+
+[feat] Add the vertical wizard page template. (#5672)
+@ernestt
--- a/apps/docsite/src/components/templateComponents.ts
+++ b/apps/docsite/src/components/templateComponents.ts
@@ -85,6 +85,10 @@ export const TEMPLATE_COMPONENTS: Record<
     () =>
       import('../../../../packages/cli/assets/templates/pages/form-wizard-inline/page'),
   ),
+  'form-wizard-vertical': lazy(
+    () =>
+      import('../../../../packages/cli/assets/templates/pages/form-wizard-vertical/page'),
+  ),
   'gallery-hero': lazy(
     () =>
       import('../../../../packages/cli/assets/templates/pages/gallery-hero/page'),
--- a/packages/cli/assets/templates/pages/form-wizard-vertical/page.tsx
+++ b/packages/cli/assets/templates/pages/form-wizard-vertical/page.tsx
@@ -1,0 +1,1365 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+'use client';
+
+/**
+ * A publishing flow with a vertical Stepper in a persistent side rail: five
+ * steps, each named and explained, with the form for the current step filling
+ * the content column beside it.
+ *
+ * The vertical orientation earns its keep when step labels are not
+ * self-explanatory. "Payouts and agreements" needs a sentence to distinguish it
+ * from "Distribution"; a horizontal stepper has no room for that sentence, so
+ * it either truncates or drops it. Here the rail is the flow's table of
+ * contents — visible the whole time, so someone five steps deep can still see
+ * what is left without leaving the step they are on.
+ *
+ * The rail lives in the Layout's `start` slot, so it scrolls independently of
+ * the form and stays put while a long step scrolls. Actions sit in the Layout
+ * `footer`, which spans the full width beneath both columns: they are the
+ * flow's controls rather than the form's, and pinning them means Continue is
+ * reachable without scrolling to the bottom of a long step.
+ *
+ * ## Extending this template
+ *
+ * **The rail is a summary, not a menu.** Step descriptions say what the step is
+ * for, in one sentence, in the user's terms. If a description needs two
+ * sentences the step is doing two things. Resist adding per-step badges or
+ * counts — the rail competes with the form for attention as it is.
+ *
+ * **Match the step count to the rail.** Five steps with descriptions is about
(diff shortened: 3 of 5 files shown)


cixzhang (dismissed):
`xinzi-bot` already caught the hidden-step and missing-focus paths; this confirms both and adds completion and responsive-host failures. This flow still isn’t safe to copy yet. People can bypass an invalid step, reach Submit, and get no completion; Cancel, Save and exit, and Enter are inert. When Categories, Territories, or Agreements are the only errors, Continue leaves focus on itself, so keyboard users must hunt for the field.

The three-column layout keys to viewport width. In a 700px host inside a wide page, it keeps the rail and guidance panel and crushes the form to 100px.

Could we make every exit path real, validate the whole wizard before completion, focus the first invalid field, and switch modes from available width? The final review should use provider-aware date and price formatting and the Wizard category so typechecking passes.

[Reviewed by Robohands]

ernestt:
Release-process check: this PR adds a consumer-visible template under the publishable @astryxdesign/cli 0.5.x package, but no .changeset/*.md file is present. It needs a patch entry whose first body line is a [feat] user-facing summary ending in (#5672) and whose second body line is @ernestt.

PR #5673, opened 2026-08-28
fix(FieldLabel): remove the gap between label and description

ernestt:
## Summary

A field's label and its description now sit flush against each other, with no gap between them.

`FieldLabel` returned a fragment, so the `<label>` and the description `<span>` came out as bare siblings of whatever column happened to hold them. The space between them was therefore set by that parent's `gap` — the same declaration that separates the label group from the control below it. No caller could close the pair without also pulling the control up against the description, and each had picked its own value: `Field` 4px, `CheckboxInput` and `Switch` 2px.

The pair now shares a wrapper of its own, so the space between them is theirs to set. This is **one change in `FieldLabel`, not a change across ~20 input components** — every input reaches it through `Field`.

## Why

Requested against this reference, where the label and description read as a single block:

XDS reference

## Before / after

Before and after

Measured in Chromium as `description.top - label.bottom`:

| | label → description | description → control |
| --- | --- | --- |
| `TextInput` | 4px → **0px** | 4px → 4px |
| `Field` | 4px → **0px** | 4px → 4px |
| `CheckboxInput` | 2px → **0px** | n/a — control sits beside the label |
| `Switch` | 2px → **0px** | n/a — control sits beside the label |

Only the label/description pair closes up. **The description → control gap is unchanged at 4px**, so fields keep their existing rhythm.

## Implementation notes

- The label and description are wrapped in a flex column with no `gap`. `Field`'s 4px `containerGap` is untouched — it now spaces the label *group* from the control instead of also splitting the label from its description.
- `CheckboxInput` and `Switch` each carried their own 2px label wrapper to do this job locally. The shared wrapper makes those redundant, so both are removed and all three callers now agree.
- A hidden label group takes `display: contents`, so the wrapper box leaves the caller's layout entirely and the sr-only label and description stay out of flow exactly as before — a hidden label still costs no space and draws no gap. The description keeps its own `srOnly` class.
- No API change: no new or changed props, no export changes.

## Test plan

- [x] `vitest` over `Field/`, `CheckboxInput/`, `Switch/` — 5 files, 156 tests pass (covers `isLabelHidden`, group labels, description click-forwarding, `aria-describedby` wiring).
- [x] Gaps measured in a real browser via Playwright against Storybook, before and after — table above.
- [x] Verified the description → control gap is unchanged in the same run.
- [ ] Reviewer: visual pass over the `WithDescription` stories for Field, TextInput, CheckboxInput, Switch.

---

<sub>Screenshots are hosted on the throwaway branch `assets/core-field-label-description-gap` to keep this PR's diff code-only. Safe to delete once merged.</sub>


ernestt:
/accept-visual 33221383949/1 All four changed frames are the same story (Field Status Variants) in four themes, and the diff is pure vertical translation: each description moves up 4px to sit flush under its label, so the three stacked fields shift 4px, 8px and 12px. No colour, border, size or text changed, and the description to input gap stays 4px.

cixzhang (dismissed):
Thanks, the label/description grouping is much clearer. One regression blocks this shape: `FieldLabel` is exported with `BaseProps<HTMLLabelElement>`, but the new wrapper becomes the parent’s flex/grid item while `style`/`xstyle`/`className` still land on the inner label. An existing `<FieldLabel style={{order: 2}} />` now compiles but no longer reorders; I reproduced it on the no-description path, and flattening only the wrapper restores it. Please preserve those public layout overrides while keeping label+description grouped, with a regression check for the standalone no-description call.

[Reviewed by Robohands]

ernestt:
Addressed in 0b4024e025b: when FieldLabel has no description, its grouping wrapper now uses display: contents so the public label remains the parent flex/grid item and existing className, style, and xstyle layout overrides still apply. Added a standalone order regression test; all 20 focused tests, ESLint, and the changeset check passed.

josephfarina (approved):
Reviewed at exact head 4f07b23e29d4. The latest commits resolve the standalone FieldLabel layout regression from the prior review with display: contents and a focused regression test. The label-description grouping is now correct, and all exact-head checks are green. No remaining blocker found.

PR #5674, opened 2026-08-29
ci: route remaining gh-pages writers through publisher

cixzhang:
## Summary
- migrate PR preview deploy/redeploy, preview cleanup, gh-pages compaction, vibe report deployment, and vibe screenshot uploads onto the shared gh-pages publisher
- preserve preview subtree ownership, cleanup retention rules, report screenshot precedence, and compaction whole-tree exclusivity
- add cross-writer tests proving preview, report, visual evidence, and baseline paths survive unrelated writers

## Test plan
- `actionlint -ignore 'label "2-core-ubuntu-arm" is unknown' -ignore 'shellcheck reported issue' .github/workflows/deploy-preview.yml .github/workflows/redeploy-preview.yml .github/workflows/cleanup-previews.yml .github/workflows/compact-gh-pages.yml .github/workflows/vibe-screenshots.yml .github/workflows/pr-comment.yml .github/workflows/visual-acceptance.yml .github/workflows/visual-acceptance-promote.yml .github/workflows/visual-baseline.yml .github/workflows/release-gate.yml .github/workflows/deploy.yml`
- `pnpm exec prettier --check ...`
- `pnpm exec eslint --cache ...`
- `pnpm check:repo`


cixzhang (commented):

Thanks for preserving [#5633’s approved patch](https://github.com/facebook/astryx/pull/5633#pullrequestreview-5050545236) exactly. Two production boundaries still break:

- [Compaction](https://github.com/facebook/astryx/blob/574ee795b0eff2aded1da1d5537583cd816d439a/.github/scripts/lib/gh-pages-publisher.mjs#L2034-L2042) never configures Git identity, so the clean runner and exact CI fail at `commit-tree`.
- [Cleanup](https://github.com/facebook/astryx/blob/574ee795b0eff2aded1da1d5537583cd816d439a/.github/scripts/lib/gh-pages-publisher.mjs#L1654-L1673) dates screenshots from Git history, but weekly parentless compaction resets every surviving path’s timestamp. A July 1 screenshot still survived August 29 cleanup after August 28 compaction.

The terminal-state test also reaches the real unauthenticated `gh`, leaving exact CI red. Could we make age compaction-safe, configure commit identity, and keep the tests hermetic?

[Reviewed by Robohands]

cixzhang:

Thanks — the clean-runner failures are fixed: the terminal-state tests use the fixture GitHub client, compaction configures the local bot identity, and exact-head focused tests pass.

One retention boundary remains. [Compaction](https://github.com/facebook/astryx/blob/12bf8338270f472c55a29907d3475ff56f2e2347/.github/scripts/lib/gh-pages-publisher.mjs#L2022-L2025) deletes screenshots already older than 30 days, then writes every survivor into a new parentless root commit. A screenshot 8 days old on the first compaction therefore looks newly written. Each later weekly compaction sees it as only 7 days old and resets it again; at 32 original days cleanup still sees the September 18 root date and leaves it in place.

Could we preserve each surviving screenshot subtree’s original age independently of compacted commit history, and add a repeated-compaction regression?

[Reviewed by Robohands]

cixzhang:

Thanks — the durable first-seen index itself fixes the age reset: an August 20 screenshot kept that timestamp through four isolated compactions, and September 21 cleanup removed it while a September 10 screenshot survived.

One production boundary remains: the first scheduled compaction succeeds, but the next weekly run fails before publishing. [The first run removes terminal legacy queue state, while every later run still passes that absent path to `git add -A`](https://github.com/facebook/astryx/blob/c9ca570b3fd16e788cd61972eb0bd75ce41770f5/.github/scripts/lib/gh-pages-publisher.mjs#L2157-L2170), which exits `fatal: pathspec 'visual-gate/publication-queue' did not match any files`. The 45 checked-in tests stay green because each compaction case runs only once.

Could we make compaction tolerate its own terminal cleanup and add the August 28 → September 21 repeated-compaction regression?

[Reviewed by Robohands]

cixzhang:
Semantic verdict: approve

Thanks — at `429de6a`, four compactions preserve dates and legacy absence. September 21 removes August 20, keeps September 15, and leaves no orphan metadata. No blocker remains.

[Reviewed by Robohands]

PR #5675, opened 2026-08-29
feat(PowerSearch): add popoverMaxWidth prop

athz:
## Summary

Adds an optional `popoverMaxWidth` prop to `PowerSearch` that caps the filter-editor popover width while ensuring it never exceeds the anchor width.

## Problem

When `PowerSearch` is anchored to a full-width bar (e.g. a page-wide filter bar), the editor popover stretches to match the bar — often 1000px+ — making the 2–3 field form inside it needlessly wide and harder to scan.

There is currently no way for consumers to constrain this: `menuWidth` only affects the field/search dropdown, not the editor popover. The popover layer styles are hardcoded with no prop or CSS variable escape hatch.

## Solution

A single optional prop:

```tsx
<PowerSearch popoverMaxWidth={720} ... />
```

When set, applies `max-width: min(Npx, anchor-size(width))` to the editor popover layer. This means:

- **Narrow anchor** (e.g. 400px): popover = 400px (anchor wins, unchanged)
- **Wide anchor** (e.g. 1400px), maxWidth=720: popover = 720px (cap wins)
- **Prop omitted**: no max-width, current behavior preserved

The `min()` clamp guarantees the popover never exceeds its anchor regardless of the value passed.

## Changes

- `PowerSearch.tsx`: new `popoverMaxWidth` prop, applied as inline `style.maxWidth` on the popover render call
- `PowerSearch.test.tsx`: test verifying the style is applied
- `PowerSearch.doc.mjs`: prop documented in all locale sections

### DIFF
--- a/.changeset/powersearch-popover-max-width.md
+++ b/.changeset/powersearch-popover-max-width.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] Add `popoverMaxWidth` prop to PowerSearch, allowing consumers to cap the filter-editor popover width while ensuring it never exceeds the anchor width.
+
+@athz
--- a/packages/core/src/PowerSearch/PowerSearch.doc.mjs
+++ b/packages/core/src/PowerSearch/PowerSearch.doc.mjs
@@ -132,6 +132,12 @@ export const docs = {
         'Width in pixels for the main field/search menu. Does not affect field value editors.',
     },
     {
+      name: 'popoverMaxWidth',
+      type: 'number',
+      description:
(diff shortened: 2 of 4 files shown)


cixzhang (changes requested):
The legibility problem is real, but `popoverMaxWidth` puts an internal readability policy at every callsite. It also does not behave as documented below 400px because the existing minimum wins; the current test stays green when the cap is ineffective, and generated docs duplicate the prop.

Please define one internal readable width and keep the clamped editor near the interaction point rather than adding this public prop.

[Reviewed by Robohands]

athz:
@cixzhang if the default max width behaviour is already part of powersearch, I assume I can close this?

PR #5678, opened 2026-08-29
fix(chat): only the reader can release auto-follow (#5662, #5663)

### LINKED ISSUE #5662: [Bug] useChatStreamScroll: wheel and touch never release auto-follow under prefers-reduced-motion
### Description

`onWheel` and `onTouchMove` in `useChatStreamScroll` gate on `animatingRef.current`. Under `prefers-reduced-motion`, `startAnimation` calls `jumpToBottom()` and returns before ever setting `animatingRef`, so neither handler can fire. For readers with reduced motion enabled, both gesture releases are permanently no-ops.

The remaining release path is `onScroll`'s direction check, which returns early whenever `scrollHeight` or `offsetHeight` changed on that event. While content is arriving the geometry changes on nearly every frame, so an upward gesture is repeatedly swallowed and the reader is pulled back to the bottom.

Expected: an upward wheel or a touch drag releases auto-follow in both motion modes, as the animated path already does.

The predicate appears to mean "we are following, so this gesture is the reader taking over". That is `lockedRef.current`, which holds in both modes and is a superset of the animating case.

### Reproduction

1. Enable "Reduce motion" in the OS, or emulate `prefers-reduced-motion: reduce` in DevTools.
2. Render `ChatLayout` with a transcript taller than the viewport, locked at the bottom.
3. Append content so `scrollIfLocked` runs, and during that growth send an upward wheel over the scroller.
4. Observed: `isLocked` stays `true`; the next resize returns the viewport to the bottom. With reduced motion off, the same gesture releases immediately.

Driving the hook directly shows the same thing without a browser: with `matchMedia('(prefers-reduced-motion: reduce)').matches === true`, `animatingRef` is never set, so a `wheel` with `deltaY: -120` leaves `isLocked` unchanged.

### Astryx Version

`@astryxdesign/core@0.5.0`

### Environment

Electron 40 (Chromium 142), macOS 15. Found while replacing a product-side scroll implementation with this hook in Apache Maka; we currently carry it as a local patch changing both predicates to `lockedRef.current`.

### LINKED ISSUE #5663: [Bug] jumpToBottom does not cancel the in-flight spring, so animation loops stack
### Description

`jumpToBottom` is documented as cancelling any in-flight spring so a later tick cannot fight the assignment, but it only clears `animatingRef`. `animate()` never reads `animatingRef` — it continues while `lockedRef` is true and the content is scrollable — and `scrollToBottom` sets `lockedRef = true` immediately before calling `jumpToBottom`. The running loop therefore survives while `animatingRef` reads `false`, so the next `startAnimation` starts a second loop alongside it.

Two consequences: per-frame scroll distance multiplies with the number of live loops, and `animatingRef` no longer reflects reality, which disables the wheel and touch interrupts even in the animated path.

Expected: after `jumpToBottom`, exactly zero or one animation loop is live.

### Reproduction

1. Mount `useChatStreamScroll` on a scrollable container, locked.
2. Grow the content so `scrollIfLocked` starts a spring.
3. Call `scrollToBottom({ behavior: 'instant' })` while that spring is running.
4. Grow the content again to trigger `startAnimation`.
5. Observed: two `requestAnimationFrame` chains are live; a single frame advances `scrollTop` by roughly twice what one loop advances it (~240px against ~120px at the default `damping: 0.7` / `stiffness: 0.05` / `mass: 1.25`). Each further repetition adds another loop.

A generation token incremented by `jumpToBottom` and checked at the top of `animate()` would make the documented cancel real.

### Astryx Version

`@astryxdesign/core@0.5.0`

### Environment

Electron 40 (Chromium 142), macOS 15. Found while adopting the hook in Apache Maka; not patched locally — after the fixes on our side the path that reaches it is narrow, so we are only reporting it.

yyq1025:
## Summary

Fixes #5662. Fixes #5663.

`onWheel` and `onTouchMove` gated on `animatingRef` — "a rAF loop is live" — as a stand-in for "we are following". #4104 added a `prefers-reduced-motion` path that follows without ever starting a spring, so for readers with reduced motion the predicate is never true and both handlers are dead (#5662). The remaining release path, `onScroll`'s direction check, returns early whenever the geometry changed on that event — and while content is arriving it changes on nearly every event, which is exactly when a reader wants out. The same conflation is behind #5663: `jumpToBottom` (#3800) clears `animatingRef` while `lockedRef` stays true, so the running loop never reaches its own exit and the next `startAnimation` opens a second one beside it.

**What this does.** While following, the hook owns the container's position. It marks the scroll element `data-astryx-chat-following` while locked, and one injected rule turns scroll anchoring off under that attribute; the element's own inline `overflow-anchor` is never touched. With anchoring off, the only move the browser makes on its own is the resize clamp onto a new, smaller bottom — and a reader scrolling up never lands on the bottom. So the arbitration is one exact rule: a scroll that lands above the last position the hook set or saw, and not on the bottom, is the reader. The spring records each position it writes, so a reader's move that lands between two frames still reads as below it even when the next frame's write outran it. No gesture is consulted, in either motion mode.

Earlier revisions of this branch inferred whether a scroll was the reader's from the gesture before it — first a `touchstart` waiver, then a walk from the event target for a nested scroller with room above. Each had a counterexample raised in review: a bubbled `touchstart` under a resize clamp, a resting finger under an anchoring shift, a nested scroller at its edge with `overscroll-behavior: contain`. The question was never who owns the gesture; it is what can move a following container up other than the reader, and the answer is two things, one exact and one switchable.

Unlocked, anchoring is restored: a reader up in the history wants the browser to hold their place when content above them changes (#4102 counts that among `ChatLayout`'s advantages, and #4121's load-earlier compensation falls back to it). Locked, the spring is already the authority on the position, so anchoring has nothing to do — the same reasoning under which `ListInput` (#4740) and LegendList's `maintainVisibleContentPosition` switch it off.

## Change

`useChatStreamScroll.ts` only:

- `setLocked(locked)` — the one place `lockedRef` changes; it also toggles `data-astryx-chat-following` on the scroll element. A single `<style data-astryx-chat-following-style>` is injected once per document with `[data-astryx-chat-following]{overflow-anchor:none !important}` — the pattern `highlightRanges` uses for its dynamic rules. Riding on an attribute rather than the inline style means a consumer's own `overflow-anchor`, set before or during a lock, is never read, written or restored; it simply applies again once the attribute is gone. `!important` so an inline `auto` cannot switch anchoring back on under a following container; an inline `!important` still wins, deliberately. Toggled directly at each transition, not from an effect, so there is no render between a lock change and the attribute.
- `onScroll` — unlocks on `scrollTop < lastScrollTop` unless the scroll landed on the bottom (`scrollHeight - scrollTop - clientHeight < 1`; `clientHeight` rather than the existing `dist`, since `offsetHeight` includes borders and a horizontal scrollbar). `lastScrollTop` is now the last position the hook set *or* saw: `animate()` records each write, as `jumpToBottom` and `scrollToMessage` already did, so the coalesced scroll event after a wheel that landed between two frames reads below the write and releases — the comparison use-stick-to-bottom makes against `ignoreScrollToTop`. The `scrollHeight`/`offsetHeight` tracking and the synthetic-resize early return are gone; a resize that does not move `scrollTop` is not an upward scroll, and one that clamps lands on the bottom.
- The `wheel`, `touchstart`, `touchmove`, `touchend`, `touchcancel` listeners are gone.
- `rafRef` — the spring's pending frame; `jumpToBottom` calls `cancelAnimationFrame` on it. The loop's continuation is always exactly one pending handle, so the cancel is exact, and loops cannot stack (#5663). Same shape as `useChatDictation` and `useSpeechRecognition`.

About the size of `main`'s (+71/−67). No public API change: no prop, option, return value or export is added or altered.

## Verification

- **Unit** — `useChatStreamScroll.test.tsx`: 8 new tests, 5 of which fail on `main` (scroll-up release while content grows; the same under reduced motion; anchoring owned while locked and handed back when not; a scroll-up the next spring frame outruns still releases; no stacked loops). The other 3 pin what must not regress: content growing alone keeps following; a resize clamp onto the new bottom does not release; the element's own inline `overflow-anchor` is never touched, whether set before the lock or changed during it (fourth and fifth review rounds). `packages/core/src/Chat`: 191/191. `tsc --noEmit`, `eslint`, `prettier`, `check:changesets` clean.
- **Chromium** — a probe story with a nested `overflow-y: auto; overscroll-behavior: contain` block, 80 lines streamed at 16ms, a 100px block above the viewport collapsed mid-stream. The tab was headless, so rAF ran on timers; layout and anchoring are the browser's own.
  - Wheel up over the contained nested block while the spring trailed by 266px, then the collapse: `scrollTop` 2458 → 2478 (no anchoring shift, just the spring), following throughout, 4px from the bottom at the end. This is the third review's case.
  - Reader scrolls the container up 300px during streaming: released, `overflow-anchor` back to the element's own inline value (`''` in the probe).
  - Then the same collapse while released: `scrollTop` 2108 → 2008, exactly the collapsed height — the reader's place held by the browser.
  - Counterfactual: anchoring forced back on while following, same collapse: `scrollTop` 2406 → 2331 and the hook read it as the reader — released. That is why it is off while locked.
  - Inline `overflow-anchor: auto` set by the caller while following: computed value stays `none` under the attribute rule.
  - The outrun case (a −100px wheel between spring frames, net position still advancing) is pinned by the unit test that encodes the coalesced event; it was not re-measured headed here.

## Follow-ups

- With an external `scrollRef` on `document.documentElement`, `overflow-anchor` is written on the root element; per CSS Scroll Anchoring, the viewport's scrolling box is generated by the root, so it applies, but the browser runs above used an element scroller.
- Safari has no scroll anchoring, so the property is a no-op there and the rule holds without it.
- The injected rule is `!important`, so a consumer who needs anchoring on under a following container would have to say so with an inline `!important`; nothing in the library does.


cixzhang (changes requested):
Readers using nested scroll content can lose auto-follow even though they never moved the chat. `touchstart` marks every parent scroll as gesture-owned until `touchend`, so an unrelated content shrink bypasses the resize guard and unlocks the parent; later streamed content then stays behind.

I reproduced this in Chromium on the exact head: after nested touch + parent clamp + growth, main followed to 1077/1079px, while this head stayed at 79/1079px. Please keep parent resize clamps synthetic unless the chat scroller actually consumed the gesture.

[Reviewed by Robohands]

yyq1025:
Good catch, and the repro is exactly right: the drag waiver was scoped by time, and `touchstart` bubbles from a nested scroller, so a clamp inside that window was read as the reader.

Fixed in 40e1e19: the guard now keeps a scroll synthetic when the geometry changed **and** it landed on the bottom, regardless of any gesture in flight. A resize clamp can only ever land on the new maximum, and a reader scrolling up never lands there, so the landing point decides ownership exactly. I went this way rather than walking from `e.target` for a scrollable ancestor at `touchstart`: ownership can't see scroll chaining at the nested scroller's edge, where the parent does consume the drag, and the same rule also closes the two-frame window on the wheel side.

Added a test that mirrors your sequence (nested `touchstart` → parent shrink clamp → still locked); it failed on the previous head and passes now, `packages/core/src/Chat` 191/191. In Chromium with the same setup, streaming ended 853px from the bottom before and 11px after. PR body updated to match.

cixzhang (changes requested):
Thanks for fixing the full-clamp case, but the parent still treats a bubbled gesture as ownership of any non-bottom geometry change. In Chromium on this exact head, while follow trailed streaming by 576px, a nested touch plus a 100px collapse above preserved that distance (1344/2320 → 1244/2220), yet `landedAtBottom` was false, so the resize guard was bypassed and the parent unlocked. The no-gesture control stayed locked; touch and wheel reproduced in three runs. Please make the parent waive the synthetic guard only when its own scroll was caused by that input, rather than adding another geometry heuristic.

[Reviewed by Robohands]

yyq1025:
Agreed — the landing-point rule only covered the clamp, and a bubbled gesture was still standing in for ownership. 063d25a makes ownership come from the gesture itself.

`scrollUpConsumedBelow(target, el)` walks from the event target up to the scroller; a vertically scrollable element in between that still has room above takes the gesture, since scroll chaining only hands one outward from a scroller at its edge. A wheel up that is not consumed below vouches for its two frames as before. A drag records its target and finger position on `touchstart` and vouches from the first `touchmove` that moves the finger down and is not consumed below, until the finger lifts; a finger that rests without moving is not a gesture, so in your sequence nothing is waived and the anchoring shift stays synthetic. `landedAtBottom` remains for the one clamp a consumed gesture can still coincide with (a drag at the bottom that moves nothing while a block above collapses).

Tests mirror your sequence (1344/2320 → 1244/2220 under a resting nested `touchstart`, still locked), plus a drag and a wheel a nested scroller consumes, and a nested scroller at its top handing the drag to us — the first three failed on the previous head. In Chromium with real `TouchEvent`s on a probe story: resting finger on the nested block plus a mid-stream collapse stays following; a drag on the nested block with room above stays following; a drag on the container itself, and one on the nested block at its top, both release. 195/195 in `packages/core/src/Chat`.

The red CI is `packages/cli/api/search/search.mjs` failing `tsc --project tsconfig.api-dts.json` on the merge with current `main` — not touched here; it came in with #5289 and #5994 is the fix.

cixzhang (changes requested):
Thanks for tightening the nested-scroll cases. Readers can still lose auto-follow because the parent predicts ownership from a descendant’s `scrollTop`, rather than observing whether the chat scroller moved. In Chromium, a nested scroller at its top with `overscroll-behavior: contain` kept both scrollers stationary; after 100px above collapsed, the parent flipped from locked to unlocked. The identical no-wheel control stayed locked, and a focused regression test fails 1/22. This is the same ownership gap as the prior review: an edge is not proof that chaining reached the parent. Please make the resize waiver depend on the chat scroller actually consuming the input, rather than adding another descendant exception. If you'd rather talk it through with someone, we're in [Discord](https://discord.com/invite/XnsUcFykEP).

[Reviewed by Robohands]

yyq1025:
You're right, and the `contain` case settles it: any rule that predicts ownership from a descendant is inference, and each round found the next counterexample. 3879ba8 stops inferring.

The question I should have asked is what can move a following container up other than the reader. There are two things: the resize clamp onto a new bottom, and CSS scroll anchoring. The clamp lands exactly on the bottom, which a reader scrolling up never does. Anchoring is now switched off on the scroll element while locked (`overflow-anchor: none`, restored to `''` on unlock), the same shape as `ListInput` and LegendList's `maintainVisibleContentPosition` — while following, the spring is the only writer of the position, so anchoring has nothing to do; unlocked, a reader up in the history keeps it. With that, the arbitration is one rule: an upward scroll that does not land on the bottom is the reader. The wheel and touch listeners, the waiver, and the geometry tracking are gone; the hook is 20 lines shorter than `main`.

Your sequence in Chromium on this head: nested block at its top with `overscroll-behavior: contain`, wheel up over it while the spring trailed by 266px, then a 100px collapse above the viewport — `scrollTop` 2458 → 2478 (the spring, no anchoring shift), following throughout, 4px from the bottom at the end. Released state: the same collapse moved `scrollTop` 2108 → 2008, the reader's place held. Counterfactual with anchoring forced back on while following: 2406 → 2331, read as the reader, released — which is the shift every earlier revision was trying to explain away. 6 new unit tests, 4 red on `main`; `packages/core/src/Chat` 189/189. PR body rewritten to match; the "restores #1249" framing is gone since this is a different position, not the old one.

cixzhang (changes requested):
Thanks, this now fixes the gesture/resize ownership gap in Chromium, including touch and reduced motion. One external-scroller path still changes caller behavior: on unlock, `setLocked(false)` clears an existing inline `overflow-anchor: none`, so later content changes can move a reader who explicitly disabled anchoring. In the same probe, main preserved `none`; this head changed it to `auto`. Could we save and restore the scroll element’s prior inline value on unlock and cleanup?

[Reviewed by Robohands]

yyq1025:
Done in 9788887: the listener effect captures the element's inline `overflow-anchor` when the hook takes the element over, and both `setLocked(false)` and the effect cleanup write that value back instead of `''`. A consumer's own `none` now survives unlock and unmount; a test pins it (mounted disabled with `none` set, enabled, unlocked by a scroll-up, unmounted — `none` at every step), red on the previous head. `packages/core/src/Chat` 190/190.

cixzhang (changes requested):
Thanks, the static `overflow-anchor: none` case is fixed, but two ownership paths still lose reader/caller intent on this head.

In headed Chromium during active streaming, a real −100px wheel event landed between spring frames: `scrollTop` still advanced 25.5px and follow stayed locked. Because the hook now observes only net position, a reader's smaller upward gesture can be erased by the spring and they remain pinned to the tail.

Also, an external scroller changing inline `overflow-anchor` from `auto` to `none` while locked is restored to the captured `auto` on unlock. React has already committed `none`, so it leaves the DOM at `auto`, changing the caller's current policy. Please preserve the latest caller-owned value, not only the mount-time value.

[Reviewed by Robohands]

yyq1025:
Both in 9b19b03.

The outrun: the hook compared each scroll event against the last position it had *seen*, and a wheel landing between two spring frames coalesces into one event carrying the net after the next write. `animate()` now records the position it writes, as `jumpToBottom` and `scrollToMessage` already did, so that event reads below the hook's own last write and releases — the same comparison use-stick-to-bottom makes against `ignoreScrollToTop`. A test encodes your sequence (spring writes +120, event arrives at write − 100, released), red on the previous head.

The caller's value: the switch no longer goes through inline style at all. Locked, the element carries `data-astryx-chat-following`; one injected rule (`[data-astryx-chat-following]{overflow-anchor:none !important}`, the `highlightRanges` pattern) turns anchoring off under it. The element's own `overflow-anchor` is never read, written or restored — whatever the caller set before or during the lock is what applies once the attribute is gone. Test: inline `auto` at mount, caller changes it to `none` while locked, unlock — `none`, and the hook never wrote the property. In Chromium an inline `auto` under the attribute computes to `none`. 191/191 in `packages/core/src/Chat`.

cixzhang (approved):
Thanks, both prior ownership gaps are fixed on this head. Headed Chromium released auto-follow for all 18 wheel trials, and caller-owned `overflow-anchor: none` survived unlock.

[Reviewed by Robohands]

PR #5681, opened 2026-08-29
fix(useFocusTrap): only restore focus when focus actually entered the trap

### LINKED ISSUE #5651: useFocusTrap restores focus for a popup that never held it, so a Typeahead/PowerSearch dropdown cannot be dismissed by one outside click
### Summary

`useFocusTrap`'s focus-restore runs even for a popup that never held focus, so dismissing a `Typeahead` / `PowerSearch` dropdown with an outside click re-focuses the anchor input the click just blurred. Because the input is then already focused, clicking it again fires no `focus` event and `hasEntriesOnFocus` never reopens the menu — the control ends up in a state where the only ways out are a *second* outside click or typing into the field.

### Version

`@astryxdesign/core@0.5.0` (latest at time of filing), React 19.2, Chromium.

### Reproduction

Any `PowerSearch` (or `Typeahead` with `hasEntriesOnFocus`):

1. Click the input — the listbox opens.
2. Click somewhere outside, on a non-focusable area of the page.
3. Click the input again.

Observed, reading `document.activeElement` and the input's `aria-expanded` at each step:

| step | `activeElement` | `aria-expanded` |
|---|---|---|
| click the input | `INPUT[role=combobox]` | `true` |
| click outside | `INPUT[role=combobox]` | `false` |
| click the input again | `INPUT[role=combobox]` | `false` |

Expected: step 2 leaves focus off the input (as a *second* outside click already does today), and step 3 reopens the listbox.

### Cause

`usePopover` wires `useFocusTrap({isActive: layer.isOpen})` unconditionally, and the trap's restore effect (`src/hooks/useFocusTrap.ts`) does:

```ts
const previouslyFocused = document.activeElement as HTMLElement | null;
const container = containerRef.current;

return () => {
  const active = document.activeElement;
  const focusWasLost =
    active == null ||
    active === document.body ||
    active === document.documentElement ||
    (container != null && container.contains(active));

  if (!focusWasLost) return;
  // …previouslyFocused.focus()
};
```

`BaseTypeahead` opens its dropdown with `role: 'none'` and `hasAutoFocus: false` precisely so the input keeps DOM focus while the listbox is shown — the trap never receives focus, and `previouslyFocused` *is* the anchor input. On an outside click the sequence is:

1. the browser blurs the input → `activeElement` becomes `<body>`;
2. native `popover=auto` light dismiss hides the layer;
3. `isActive` flips false, the cleanup sees `activeElement === document.body`, reads that as "focus was lost", and calls `previouslyFocused.focus()`.

So the restore, meant to rescue focus from a popup that *had* it, instead cancels a blur the user asked for. Instrumenting `HTMLElement.prototype.focus` on the input confirms the refocus originates from this cleanup.

Blurring the input from the host side does not work around it: a blur issued on `pointerdown` is undone by the same restore, and one issued on `pointerup` / `click` / a `requestAnimationFrame` is followed by the restore's `focus()`, which re-fires `handleFocus` and *reopens* the menu.

### Suggested fix

Only restore when focus actually entered the trap while it was active:

```ts
let hadFocusInside = container?.contains(document.activeElement) ?? false;
const markFocusInside = (event: FocusEvent) => {
  const target = event.target as Node | null;
  if (target != null && containerRef.current?.contains(target)) {
    hadFocusInside = true;
  }
};
document.addEventListener('focusin', markFocusInside, true);

return () => {
  document.removeEventListener('focusin', markFocusInside, true);
  if (!hadFocusInside) return;
  // …existing focusWasLost logic
};
```

Popups that do take focus — dialogs, `DropdownMenu`, anything with `hasAutoFocus: true`, and a `Typeahead` option click (the option is `tabIndex={-1}` inside the popover, so it records a `focusin`) — are unaffected.

We are running exactly this as a `pnpm patch` against 0.5.0. Verified in a React 19 app:

- outside click releases focus (`activeElement` becomes the clicked region), and clicking the input again reopens the listbox;
- `Escape` still closes the listbox with focus kept on the input;
- a `DropdownMenu` still restores focus to its trigger on both `Escape` and outside click;
- `ArrowDown` / `Enter` selection and the `PowerSearch` edit popover's Apply still commit.

Happy to open a PR if the approach looks right, or to test an alternative — e.g. an explicit `hasFocusRestore` option on `usePopover` that `BaseTypeahead` opts out of, if you would rather keep the restore heuristic untouched for other consumers.

trakshan-mishra:
## Problem

Fixes #5651.

`useFocusTrap` captured `document.activeElement` on activation and restored focus to it on deactivation whenever focus would otherwise be lost to `<body>`. For popups that deliberately keep DOM focus on their trigger — a Typeahead/PowerSearch listbox opened with `role: "none"` and `hasAutoFocus: false` — the trap never receives focus, so the restore fired on outside-click dismissal and re-focused the anchor input. Because the input was then already focused, clicking it again fired no `focus` event and `hasEntriesOnFocus` could not reopen the menu — the control was stuck until a second outside click.

## Fix

Track whether focus entered the trap container at any point while it was active, via a `focusin` capture listener on `document`. If focus never entered, skip the restore entirely. The `hadFocusInside` flag is initialized from the current `activeElement` at activation (in case focus was already inside programmatically) and set to `true` by any subsequent `focusin` whose target is inside the container.

Popups that do take focus — Dialog, DropdownMenu, a Typeahead option click (the option is `tabIndex={-1}` inside the popover, so it records a `focusin`) — are unaffected.

## Before

```
click input  → listbox opens, focus stays on input
click outside → browser blurs input → body → light dismiss hides layer
               → isActive flips false → cleanup sees body → "focus lost"
               → previouslyFocused.focus() → input re-focused
click input again → input already focused → no focus event → menu stays closed
```

## After

```
click input  → listbox opens, focus stays on input (hadFocusInside = false)
click outside → browser blurs input → body → light dismiss hides layer
               → isActive flips false → cleanup sees hadFocusInside = false
               → restore skipped → focus stays on body
click input again → focus event fires → hasEntriesOnFocus reopens menu ✓
```

## Testing

- 2 new tests in `useFocusTrap.test.tsx`:
  - `does not restore focus when focus never entered the trap` — reproduces the Typeahead scenario: trap activates, focus stays outside, focus is blurred to `<body>`, deactivation does NOT restore.
  - `still restores focus when focus entered the trap and was then lost` — complement: focus enters the trap, then is blurred to `<body>`, deactivation DOES restore.
- All 20 `useFocusTrap` tests pass (18 existing + 2 new).
- All 11 `useFocusTrapEscapeShim` tests pass.
- `pnpm -F @astryxdesign/core typecheck` clean.
- ESLint clean on changed files.
- Pre-commit hooks (`check:sync`, `check:changesets`, `check:use-client`, etc.) all green.

## Changes

- `packages/core/src/hooks/useFocusTrap.ts` — the fix: `focusin` tracking in the restore effect.
- `packages/core/src/hooks/useFocusTrap.test.tsx` — 2 new tests.
- `packages/core/src/hooks/useFocusTrap.doc.mjs` — updated usage description.
- `.changeset/focus-trap-restore-only-when-focus-entered.md` — changeset.

Credit: the fix approach was proposed by @yomybaby in the issue report. This PR implements it with the `hadFocusInside` initialization and adds the test coverage.

### DIFF
--- a/.changeset/focus-trap-restore-only-when-focus-entered.md
+++ b/.changeset/focus-trap-restore-only-when-focus-entered.md
@@ -1,0 +1,21 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] useFocusTrap only restores focus when focus actually entered the trap while it was active. (#5651)
+
+`useFocusTrap` captured `document.activeElement` on activation and restored
+focus to it on deactivation whenever focus would otherwise be lost to `<body>`.
+For popups that deliberately keep DOM focus on their trigger — a Typeahead or
+PowerSearch listbox opened with `role: "none"` and `hasAutoFocus: false` — the
+trap never receives focus, so the restore fired on outside-click dismissal and
+re-focused the anchor input. Because the input was then already focused,
+clicking it again fired no `focus` event and `hasEntriesOnFocus` could not
+reopen the menu — the control was stuck until a second outside click.
+
+The restore effect now tracks whether focus entered the trap container at any
+point while it was active (via a `focusin` listener). If focus never entered,
+the restore is skipped entirely. Popups that do take focus — Dialog,
+DropdownMenu, a Typeahead option click — are unaffected.
+
+@trakshan-mishra
--- a/packages/core/src/hooks/useFocusTrap.doc.mjs
+++ b/packages/core/src/hooks/useFocusTrap.doc.mjs
@@ -39,7 +39,7 @@ export const docs = {
   ],
   usage: {
     description:
-      'Traps focus within a container element following the WAI-ARIA dialog focus trap pattern. Listens to focus events on the document and redirects focus back into the container if it escapes via keyboard navigation. Handles both Tab and Shift+Tab wrapping. When the trap deactivates or unmounts, focus is restored to the element that was focused before activation, unless focus was already moved elsewhere. Mouse clicks outside the container are not intercepted; use a light-dismiss handler for that.',
+      'Traps focus within a container element following the WAI-ARIA dialog focus trap pattern. Listens to focus events on the document and redirects focus back into the container if it escapes via keyboard navigation. Handles both Tab and Shift+Tab wrapping. When the trap deactivates or unmounts, focus is restored to the element that was focused before activation, unless focus was already moved elsewhere or never entered the trap (so popups that keep focus on their trigger, like comboboxes, are unaffected). Mouse clicks outside the container are not intercepted; use a light-dismiss handler for that.',
     bestPractices: [
       { guidance: true, description: 'Call focusFirst() when opening a dialog/modal to move focus into the trapped region.' },
       { guidance: true, description: 'Provide an onEscape callback to close the dialog when Escape is pressed.' },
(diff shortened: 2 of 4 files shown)


cixzhang (approved):
🟢 No review blockers found.

People using Typeahead and PowerSearch can now dismiss the results popup with one outside click and reopen it on the next click. The focused regression test fails against the parent implementation and passes on this head; real Chromium also confirms the trigger stays unfocused after light dismiss while a Popover that actually received focus still returns it to its trigger.

[Reviewed by Robohands]

PR #5682, opened 2026-08-29
fix(Typeahead): keep the field's width when a value is selected

### LINKED ISSUE #5560: Typeahead shrinks when a value is selected, in any content-sized layout (the docsite's own preview is one)
## Problem

Selecting a value makes a Typeahead **shrink**, in any layout that sizes it to its content. Reported from the docsite's own Typeahead page ("Search Field"), where the preview is wrapped in `<Center>` — `display: flex` (`Center.tsx:30-32`), so the field is a flex item.

Measured in Chromium, three fields in identical `display: flex` parents, 1000px viewport:

| | before | after | |
|---|---|---|---|
| `TextInput` (typed a value) | 199 px | **199 px** | unchanged |
| `Tokenizer` (added a token) | 199 px | **118 px** | −81 px |
| `Typeahead` (selected a value) | 199 px | **69 px** | **−130 px** |

In a plain block parent none of them move (898 px before and after at the same viewport), which is why no story shows this: every Typeahead story renders inside a fixed-width container.

## Cause

`Typeahead.tsx:505` swaps the input's styles when a token is shown:

```tsx
inputXStyle={showToken ? styles.inputHidden : undefined}
```

and `styles.inputHidden` (`Typeahead.tsx:209-216`) takes the input **out of flow**:

```ts
inputHidden: {
  width: 0,
  minWidth: 0,
  flex: '0 0 0',
  padding: 0,
  opacity: 0,
  position: 'absolute',
},
```

The input is the only child with an intrinsic width — `flex: 1; minWidth: 60px` (`BaseTypeahead.tsx`). Neither `Typeahead`'s `wrapper` nor the shared `inputWrapperStyles.base` sets a width, so the field's width is its content's. Remove the input from flow and the remaining content is the token, so a shrink-to-fit parent collapses the field onto it.

Block-level parents hide it because a block-level flex container fills its parent regardless of content. It shows up in every context that sizes to content: a flex item, `inline-block`, a grid `auto` track, a table cell, a float.

The `showToken` flag itself is doing what it says — the problem is what it applies. Tokenizer has the same shape (`inputAtMax`, `Tokenizer.tsx:265`) and shrinks for the same reason, less far because its tokens are wider.

## Why existing components don't cover this

It is internal styling of the collapsed input; there is no prop that reaches it, and a consumer cannot restore the width without knowing to set one on the field.

## Rough approaches

1. **Keep the input in flow while collapsed.** Drop `position: absolute` and hold the width at 0 — the field keeps a zero-width flex child that still carries `flex: 1`, so the wrapper keeps its stretch. Needs a check that a zero-width in-flow input does not re-introduce the caret or a focus ring.
2. **Give the field an intrinsic width that does not depend on the input** — e.g. `flex: 1` / `width: 100%` on the wrapper. Bigger blast radius: it changes how every Typeahead sizes in a flex row, including ones that look right today.
3. **Reserve the input's `minWidth` even when hidden**, so the field never collapses below the width it had.

(1) looks smallest and most faithful to what the collapse is for. Happy to take it.

## Accessibility considerations

The collapsed input already carries `tabIndex={-1}` (`Typeahead.tsx:511`) so it is not a keyboard stop while the token shows; whichever approach is taken must keep that, and must not make a zero-width input focusable by click.

freddymeta:
Fixes #5560.

## It is the component, not the docsite

The docsite only exposes it. Its preview is centred, a centred child is a flex item, and a flex item is sized to its content — a perfectly ordinary layout that every other field survives. Three fields in identical `display: flex` parents, measured in Chromium at a 1000px viewport, before and after giving each one a value:

| | before | after | |
|---|---|---|---|
| `TextInput` | 199px | 199px | unchanged |
| `Tokenizer` | 199px | 151px | −48px |
| `Typeahead` | 199px | **44px** | **−155px** |

`TextInput` is the control: same parent, same viewport, no movement. So the layout is not what is wrong.

`inline-block` reproduces it exactly (199 → 44). A block-level parent hides it completely (984 → 984 for all three), which is why no story caught it: every Typeahead story renders in a fixed-width container.

## Cause

A field's width must not depend on its value, and every other field keeps that promise for free: its `<input>` stays in flow, so the field is as wide as the input's own default size. Typeahead takes the input out of flow when the token shows —

```ts
inputXStyle={showToken ? styles.inputHidden : undefined}
// inputHidden: { width: 0, minWidth: 0, flex: '0 0 0', position: 'absolute', … }
```

— and the input is the only child with an intrinsic width. Neither the Typeahead wrapper nor the shared `inputWrapperStyles.base` sets one. Remove the input and the only thing left to measure is the token, so a content-sized parent shrinks the field onto it. At 44px the token's own label is clipped to one letter and the clear button lands on top of it.

## The two approaches that don't work

Both were measured, not reasoned about:

**Keeping the collapsed input in flow** — the issue's own suggestion, and my first instinct — moves the field from **44px to 95px**. It does not fix it. `width: 0` removes the input's intrinsic contribution whether or not the input is in flow, and dropping `position: absolute` gives back only the flex gap.

**`min(200px, 100%)`**, the shape this repo reaches for elsewhere to mean "yield when there is no room", silently does nothing — the field stayed at 44px. A percentage min-width resolves against an indefinite containing block during shrink-to-fit, so it computes to 0 and `min()` picks it. Worth knowing before someone copies that idiom into another intrinsic-sizing context.

## The fix

The field states the width it already had instead of inheriting it from the input:

```ts
inputCollapsedWidth: {
  '--typeahead-min-width': '200px',
  minWidth: 'var(--typeahead-min-width)',
},
```

Applied **only while the token shows**, so an unselected field is byte-identical to today. `min-width`, not `width`, so a block-level or stretched field still fills exactly as it does now — this only stops the collapse. The default is the width the field already measures: 181px, which is what a browser gives an `<input>` at the base font, plus this field's own 19px of padding and border.

The value is public and themeable, because the right minimum for a field is a design decision rather than a constant:

```ts
typeahead: {base: {'--typeahead-min-width': '16rem'}}
```

## Measured

before and after

Chromium, before → after selection:

| parent | before | after |
|---|---|---|
| `display: flex` | 199 → 44 | 199 → **200** |
| `inline-block` | 199 → 44 | 199 → **200** |
| `display: block` | 984 → 984 | 984 → 984 |

The remaining 1px is the gap between the browser's font-derived default and the stated floor; a theme that wants them identical can say so.

*(Assets live on the `assets/pr-5560` branch — asset-only, deletable with the PR. I have no fork.)*

## Tests and story

Two unit tests, in the probe-class style this repo already uses for declarations jsdom cannot measure: the floor is present when a token shows, and absent when one does not. Reverting the fix fails the first and leaves the second passing.

One story, `With Selected Value`. No Typeahead story rendered a selected value, and every story renders inside a fixed-width container — between them, that is exactly why a bug this visible survived. The new one shows a token in a flex parent, the case that used to collapse.

Full build, core typecheck, docs typecheck, Storybook typecheck, `check:repo` and `lint:strict` pass; 286 test files green.

## Two things CI caught, both fair

**A documented var has to be registered.** `derivedVarRegistry`'s test requires every documented var to either map onto a standard CSS property or be listed as unmappable with a reason. `min-width` maps onto this one, so it gets the entry — which also means a theme can write the standard property and have both it and the var emitted:

```ts
typeahead: {base: {minWidth: '16rem'}}
```

**A var no element declares is a var no theme can reach.** `theme-var-reachability` walks the built Storybook asking which element sets each documented var, and `--typeahead-min-width` had no answer: it was declared inside the collapsed-state style, and no story ever selected a value. The declaration moved up to the wrapper — the element carrying the `typeahead` theme target, always rendered — and only the `min-width` reading it stays conditional. Same behavior, and the same shape Spinner settled on for its own public vars. Reachability now reports `.astryx-typeahead sets it (200px)`.

## Left alone

**Tokenizer** collapses its input the same way and shrinks for the same reason (199 → 151). Its input is multi-token and wrapping, so the right floor there is a different question than a single-value field's, and worth its own change.

**Long values still widen the field** (a 49-character value takes it to 278px). That is the same invariant seen from the other side, and it predates this issue; fixing it properly means laying the token over the input rather than beside it, which changes long values from widening the field to truncating in place — a product decision rather than a bug fix, so not smuggled in here.


cixzhang (changes requested):
The default collapse is real, but this Typeahead-only contract is not safe yet: InputGroup cancels the 200px floor, narrow supported widths can overflow, Tokenizer has the same family-level failure, and the custom variable differs semantically from standard themed `minWidth`.

Please define consistent input-family sizing first and use one coherent standard-property contract rather than a component-specific public variable.

[Reviewed by Robohands]

freddymeta:
Reworked, and stacked on #5555 — its commits are below this one, base still `main` so CI runs.

**The var is gone.** Every point in your review held up when I checked it, and together they said the floor was the wrong shape:

- it was a second sizing contract beside the documented `Field.width` prop, whose own docs say *"prefer this over setting width via xstyle/className/style"*
- the 200px was hand-derived and **overshot**: the empty field measures 199, so the floor made a selected field 1px wider than an empty one
- `InputGroup` cancelled it
- and it could not help `Tokenizer`, which owes the same promise

**The cause is narrower than a floor.** Every other field's width comes from its `<input>` staying in flow with its intrinsic width. Typeahead set `width: 0; min-width: 0; flex: 0 0 0; position: absolute` on the input while a token showed, so the field was left measuring the token. The input now keeps its place and its own width — only made invisible and inert — and the token is painted over that space rather than beside it. In flow the token would add its own width, which is the same value-dependent sizing from the other direction: a long value would grow the field.

Measured in Chromium, field in a `max-content` parent:

| | empty | with a value |
|---|---|---|
| TextInput (family baseline) | 199px | 199px |
| Typeahead **before** | 199px | **57px** |
| Typeahead **after** | 199px | 223px |

No constant, nothing for `InputGroup` to cancel, no new vocabulary.

**Being straight about the residual 24px:** that is the clear button entering the row, not the value's length — ordinary for any field whose clear is conditional, and it does not vary with the value. The collapse this removes was 142px. If you want the field flat across both states, that is reserving the clear slot the way `TextInput` does (227px empty *and* filled in the same probe), which is a separate change.

**Tokenizer is deliberately not fixed here**, and you were right that it shares the failure — I measured it rather than taking the fix's word for it. Removing tokens one at a time from `overflow-inline`: **670 → 571 → 466 → 382px**. Its input never leaves flow, so that is a different mechanism (content-sized field, not a collapsed input) and wants its own change. #4405 and #5315 are both live in `Tokenizer.tsx` already, so I would rather not add a third.

Three tests, each verified failing on #5555's head. 73 Typeahead tests pass; tsc and lint clean (the two `BaseTypeahead` warnings are pre-existing on main).

freddymeta:
Before/after, captured in Chromium at 3× against a built Storybook of each side. BEFORE is #5555's head, so this isolates just this change.

before/after

The red dashed outline is the shrink-to-fit parent — a table cell, an inline-flex toolbar, a floated column. It is the layout that exposes this; a block-level parent hides it entirely, which is why no story caught it.

| | empty | value selected |
|---|---|---|
| before | 199px | **81px** — the field is the token |
| after | 199px | 223px |

The second frame is the bug in one picture: the field has shrunk to its value, and the clear button is jammed against the token with the field's own border cutting through the gap. In the fourth, the field is where it was and the token sits inside it.

(81px here vs the 57px in my earlier comment — same collapse, measured on a different story with a shorter value. The point is that the number tracks the value's length, which is exactly what a field's width must not do.)

ernestt (approved):

freddymeta:

Pushed `ca1b28f`. Timing note: the rework that removed the variable went up on 31 Aug, after your review, so points 1 and 4 were already answered by it. Point 2 was still broken until this push. Verified against `docs/families/input-fields.md` FR1/FR2, which landed the same day as your review.

**2 — "narrow supported widths can overflow." You were right, and B was worse than main.** Chromium, built core, long value. **A** = main, **B** = branch at `8197fd3`, **C** = now.

| field, long value | A | B | C |
|---|---|---|---|
| shrink-to-fit | 12px | **28.09px** | none (7px clear) |
| in `InputGroup` | 12px | **33px** | none |
| 220 / 180 / 140px | 12px | **31.09 / 33 / 33px** | none |
| escape past the border | none | **up to 4px** | none |

B traded a width bug for a worse overlap bug and put the value outside the field's border, which main never did.

**The fix.** The input and token now share a content lane — an ordinary flex item, `flex: 1` + `min-width: 0`, ending exactly where the end lane begins, with the token anchored at both of the lane's inline edges. No measurement, no constant, nothing for `InputGroup` to cancel. It's TextInput's own arrangement, which is why its end controls never had this problem.

**1 and 4** — already fixed by the rework: no 200px floor exists, and `--typeahead-min-width` is gone.

**Widths still fix the original bug:** 199 → 223px standalone, 397 → 421px grouped. Token position unchanged B→C.

**Negative control:** revert only the lane → all three new tests fail (`3 failed | 73 passed`), and the probe returns to 28–33px overlap plus the 4px escape. Restored: 76 passed.

Gates: 8959 tests / 316 files pass; build, typecheck, typecheck:docs, storybook typecheck, check:repo clean; `lint:strict` 0 errors (85 warnings, all pre-existing in `packages/lab` — verified by running eslint on `main`'s copy in place).

**Not done, deliberately: Tokenizer.** Measured 199 → 114.7px with one token, identical in A/B/C. Different mechanism — its tokens are in flow and wrap, and its input is deliberately reduced to a 40px continuation lane. Fixing it means choosing a width policy for a wrapping multi-value field, which is a design decision like DEC-1 was for Selector, not a bug fix. Happy to stack it if you want that first.

cixzhang (changes requested):
Builders now see `__onLoadingChange` in the exported `BaseTypeaheadProps` declaration, so a wrapper wiring detail can be consumed as permanent public API. The base already derives loading; please keep that handoff internal while preserving the single-spinner behavior.

The content lane and overlap fix look good.

[Reviewed by Robohands]

freddymeta:
Squashed to one commit and rebased onto the current #5555 head.

The `__onLoadingChange` finding is fixed in **#5555**, not here — that is where the prop is introduced, so #5555 landing alone would ship it publicly whatever this PR did. It now travels through a context that `index.ts` does not export; nothing on the public prop surface to find. Detail on that PR.

This branch is now just the content lane: the selected value gets a lane that owns the flexible space, and the end controls hold the corner with an `auto` margin, so the field keeps its width when a value is chosen.

8611 core tests, `check:repo` green.

cixzhang (approved):
Thanks — this is clean. The content lane fixes the width and overlap regression in both LTR and RTL.

[Reviewed by Robohands]

freddymeta:
Added the RTL coverage `pr-rtl` was flagging. Typeahead was an unexplained all-N/A component — the audit counts that as unmeasured, not RTL-ready — and the content lane is exactly the direction-sensitive part: the token opens the lane, the clear button closes it, and RTL must swap them.

A single-field `Logical order` story plus a curated D2 target measures that flip. Single-field on purpose: the comparison story renders two, and the selectors would match across both.

```
pnpm rtl:audit -- --filter Typeahead
CUR  RTL-ready  core/typeahead  {"D2":"pass"}
```

**The probe is verified, not assumed** — pinning the field to `direction: ltr` turns D2 pass → partial, so it reads real geometry. It also passes against main's Typeahead, since the order flip is writing-mode driven either way: this closes a coverage gap rather than guarding the content lane specifically, and I would rather say that than imply a regression test it is not.

`targets.json` is rebased onto main's current list rather than carrying this branch's older copy. 149 Typeahead/Tokenizer tests, `check:repo` green.

freddymeta:
/accept-visual 33607110654/1 The 24 frames are added baselines with no changed or removed existing frames. Freddy’s screenshots verify the affected states directly: the open Typeahead and the selected-value state both retain the full field width, with the token and clear control contained without collapse or overlap.

freddymeta:
@cixzhang The visual gate refused Freddy’s acceptance because `freddymeta` has write access but is not classified as a repository maintainer. Freddy reviewed the affected states directly and approved them. The report has **24 added, 0 changed, 0 removed** frames; all additions are the default Typeahead/Tokenizer stories across the 12 themes. His screenshots confirm the selected state keeps the full field width without token/clear overlap. Could you accept with this exact command? `/accept-visual 33607110654/1 The added baselines show the default Typeahead and Tokenizer fields rendering consistently across all themes; Freddy’s affected-state screenshots confirm the selected token and clear control remain contained in the full-width field without collapse or overlap.`

Issue #3641, opened 2026-07-06
RFC: i18n architecture for @astryxdesign/core


nynexman4464:
## Problem

This is an i18n proposal for Astryx — a design for how the design system exposes its user-facing strings to consumers who need to ship in languages other than English. Filing this as its own issue so the design has a home to converge in. Parent proposal: #3618. Q3 tracker: #3636.

## Scale of the problem

A quick inventory of `packages/core/src` (excluding tests, docs, stories):

- **178 system-owned hardcoded strings** across **46 components** (of 97 top-level components — 53% of the library has zero user-facing strings, mostly layout / typography primitives)
- **57 consumer-override string props** across 26 components (`emptySearchResultsText`, `disabledMessage`, etc.)
- Concentration: **PowerSearch alone owns 31%** of the surface (56 strings). Table + Pagination + Forms bucket cover another ~50%. Everything else is 1–5 strings per component.
- **Locale pack size, measured on a constructed ~180-key en catalog:** ~8 KB raw JSON, ~2 KB gzipped. A translation with typical 15% length inflation lands at ~8.5 KB / ~2.1 KB gzipped. Languages with heavier plural forms (Russian, Arabic) may push gzipped size to 3-4 KB. For a rough comparison: astryx's entire default-icon fallback registry (30 inline SVGs) is ~1-2 KB gzipped. A locale pack is roughly that magnitude — small, not zero.

The full inventory (per-component counts, prop names, category buckets) is in a companion doc.

## Two shapes of string, one system

Every user-facing string in Astryx has:

- **A default value that Astryx ships and translates.** The design system authors it and ships translations for every supported locale. Examples: `"Sort ascending"` (Table sortable), `"Previous month"` (Calendar nav), `"No results found"` (Typeahead's `emptySearchResultsText` default).
- **Optionally, a prop that lets the consumer override the default with their own wording** — for cases where the consumer needs to specialize semantically. `Typeahead` already exposes `emptySearchResultsText` because "No posts found" vs "No patient record matching query" is a domain decision Astryx can't make. When a consumer overrides, they own the translation of the value they pass, at their own callsite, using their own i18n stack.

Both categories flow through the same system. The difference is only whether the consumer has a specialization hook.

## Design constraints

Beyond the base ask on #3618:

1. **RSC-friendly.** Read a translated string in a server component without a `'use client'` boundary.
[...]

PR #5684, opened 2026-08-29
feat(i18n): introduce new design for IntlProvider and useIntl

ngolin:
- Introduce a dedicated new design of the intl subpackage with IntlProvider, IntlContext and useIntl
- Show how to achieve compatibility between InternationalizationContext and IntlProvider via getIntlContextValue
- Demonstrate how to migrate from useTranslator to useIntl using AlertDialog

### DIFF
--- a/packages/core/src/AlertDialog/AlertDialog.tsx
+++ b/packages/core/src/AlertDialog/AlertDialog.tsx
@@ -29,7 +29,7 @@ import {Button, type ButtonVariant} from '../Button';
 import type {BaseProps} from '../BaseProps';
 import {mergeProps} from '../utils';
 import {themeProps} from '../utils/themeProps';
-import {useTranslator} from '../i18n';
+import {useIntl} from '../intl';
 import {useMediaQuery} from '../hooks';
 
 const SMALL_SCREEN_QUERY = '(max-width: 640px)';
@@ -157,7 +157,7 @@ export function AlertDialog({
   'data-testid': testId,
   ...rest
 }: AlertDialogProps) {
-  const t = useTranslator();
+  const {t} = useIntl();
   const cancelLabel = cancelLabelFromProps ?? t('@astryx.alertDialog.cancel');
   const titleId = useId();
   const descriptionId = useId();
--- a/packages/core/src/i18n/InternationalizationProvider.tsx
+++ b/packages/core/src/i18n/InternationalizationProvider.tsx
@@ -23,7 +23,8 @@ import {useMemo, type ReactNode} from 'react';
 import {InternationalizationContext} from './InternationalizationContext';
 import {getLocaleDirection} from './getLocaleDirection';
 import type {Locale, MessagesByLocale, Overrides} from './types';
-import {getResolve} from './resolve';
+import {getIntlContextValue, getResolve} from './resolve';
+import {IntlProvider} from '../intl';
 
 export interface InternationalizationProviderProps {
   /**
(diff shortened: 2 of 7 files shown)


nynexman4464 (changes requested):
What do you mean by legacy API?  Why is this introducing a new API?  Please provide some context and rationale.

ngolin:
> What do you mean by legacy API? Why is this introducing a new API? Please provide some context and rationale.

Hi @nynexman4464, sorry for that, the commit message was AI-generated and didn't capture the full context. I'll provide more details later for your reference.

nynexman4464:
> Introduce a dedicated new design of the intl subpackage with IntlProvider, IntlContext and useIntl

But why what is wrong with the current provider? I'm confused by what this PR is trying to achieve. Are you having issues with the current provider?

ngolin:
> fix(ci): resolve RTL scope to public components #5946

> > Introduce a dedicated new design of the intl subpackage with IntlProvider, IntlContext and useIntl
>
> But why what is wrong with the current provider? I'm confused by what this PR is trying to achieve. Are you having issues with the current provider?

Hi @nynexman4464, @cixzhang — there’s nothing wrong with the current provider; the existing implementation works. My goal with this change is to make the intl surface clearer and safer for both maintainers and consumers. Concretely I want to:

1. Let consumers satisfy their needs with the provider from us, avoiding multiple 3rd-party solutions or mixing implementations.
2. Separate system (astryx/core) messages from user-defined messages so they never get confusing.
3. Provide a clear, easy-to-understand API that gives consumers more flexibility.
4. Improve type safety for both maintainers and consumers.

### Design notes (high level):

1. The Provider interface should clearly indicate which fields are for the system and which are user-defined; it should accept only the messages for the current locale (not all locales at once).

```tsx
interface ProviderValue<MyMessages> {
  locale: string;
  // `overrides` used to replace system component strings/catalog
  overrides: Record<string, string>;
  // `messages` are for the user's own components (not for system catalog)
  messages?: MyMessages;
}
```

2. A single hook (useIntl) should expose all capabilities without loss of type safety; users can access both custom messages and system messages.

```tsx
// for maintainers
const {direction, t} = useIntl();

t('@astryx.pagination.next'); // ok
t('@astryx.not.existed.key'); // error (caught at dev time)

// for consumers
const {messages, t} = useIntl<MyMessages>();
const {currency, discount} = messages; // fully type-safe
t('@astryx.pagination.next'); // use system messages
```

### Demo examples

1. `overrides` examples

```tsx
// Replace the system's default catalog
import zh from 'locales/zh-CN.json';
<Provider locale="zh" overrides={zh} />

// Replace catalog and tweak a few keys
<Provider locale="zh" overrides={{...zh, '@astryx.pagination.next': 'Next'}} />
```

2. `messages` examples (user-defined, can be React nodes or functions; not restricted to ICU strings)

```tsx
interface MyMessages {
  currency: React.ReactNode;
  discount: (percent: number) => React.ReactNode;
}
const en: MyMessages = {
  currency: <DollarIcon />,
  discount: percent => `${100 - percent}% Off`,
};
const zh: MyMessages = {
  currency: <ChinaYuan />,
  discount: percent => `${percent / 10} 折`,
};

const messages = {en, zh};
const locale = 'zh';

<IntlProvider locale={locale} messages={messages[locale]}>
  <MyComponent price={10} percent={30} />
</IntlProvider>;
```

3. `useIntl` examples (consumer usage with type safety)

```tsx
const MyComponent = ({price, percent}: {price: number; percent: number}) => {
  const {messages} = useIntl<MyMessages>();
  const {currency, discount} = messages;
  return (
    <span>
      {price}
      {currency}, {discount(percent)}
    </span>
  );
};
```

nynexman4464:
I'm still confused by this proposal.  It sounds like it follows the principles we set out to design our internationalization system (RFC #3641).  It's intended to be used internally, but could be used as the [internationalization provider for any app, as outlined in the docs](https://astryx.atmeta.com/docs/internationalization#using-astryx-as-your-i18n-library).  Granted it's not ideal for that case, but we're not trying to build a fully featured internationalization system.  There are lots of good-quality existing systems like [react-intl](https://formatjs.github.io/docs/react-intl/), [next-intl](https://next-intl.dev/), [react-i18next](https://react.i18next.com/), and [Lingui](https://lingui.dev/) that are already more full featured.

> messages examples (user-defined, can be React nodes or functions; not restricted to ICU strings)

I don't know if allowing arbitrary react components is something we want.  I do see other i18n systems have a "rich text" mode for converting XML like markup to some kind of component, but currently we don't have the need for this in astryx.  It adds a lot of complexity for not a lot of benefit at this time.

I appreciate the proposal here, but I'm not sure this is right for astryx at this time.

PR #5685, opened 2026-08-30
Add adaptive Selector and MultiSelector presentations

rubyycheung:
## Summary

- add `popover`, `bottom-sheet`, and `adaptive` presentation modes to Selector and MultiSelector
- consume #5654's neutral adaptive-presentation policy and shared focus-return visibility helper
- share one selector-specific BottomSheet frame across Selector and MultiSelector
- preserve existing listbox, search, single-select, and multi-select behavior rather than routing selectors through the menu adapter
- use content-hugging height and shared safe-area treatment without extra bottom content padding
- preserve trigger focus after dismissal while suppressing pointer-origin focus rings
- add Storybook and docsite Bottom Sheet examples for both components

## Stack

This PR is intentionally based on #5654. That PR owns the common overlay policy and menu-family adapter; this PR owns only the listbox-specific adapter used by Selector and MultiSelector. #5654 is based on #5395.

## Verification

- 165 Selector tests pass
- 123 MultiSelector tests pass
- core typecheck passes
- repository checks pass


rubyycheung:
Fixed in 66f37565d. SelectorBottomSheet now performs an explicit post-mount focus handoff to the owning Selector/MultiSelector search field or listbox, so the lazy BottomSheet mount cannot leave focus on its generic panel. The previous data-autofocus fallback was removed, making the existing keyboard-open regression tests exercise the explicit path for both components (288 focused tests pass). Also corrected the changeset handle to @rubyycheung. ESLint, core typecheck, and changeset validation pass. I also brought the stacked #5654/#5395 fixes into this branch. I attempted a local Playwright/Chromium Storybook check, but the host sandbox denied Chromium Mach-port startup; the automated real-browser review should now be able to verify the explicit focus target.

PR #5687, opened 2026-08-30
feat(cli): theme build --family builds a base and its extending themes as one unit

bmaurer:
## Problem

`defineTheme({extends: base})` flattens the base's tokens into the child by design — the runtime needs a resolved theme — and `astryx theme build` emits that resolved set, so every extending theme's stylesheet restates the whole base. Measured in an app that ships 14 themes off one base: 147 of ~177 custom properties per theme are byte-identical restatements of the base (~18.5KB per theme, ~230KB of render-blocking CSS). Component overrides are the mirror problem: `extends` never lands them in the child stylesheet at all, so apps copy them into every child source by hand.

## Change

**core**: `defineTheme` records the resolved base as `__extends` on the returned theme (internal, beside `__inputTokens`). Flattening previously erased the relationship, so no tool could know what was inherited.

**cli**: `astryx theme build --family <base> <children...>` builds the set as one unit:

- the base stylesheet gains a family block: its prose/token/component/on-media rules restated once, `@scope`d to the value list of every member, with the token block wrapped in `:where(:scope)` — zero specificity, so a member's own `:scope` block wins regardless of stylesheet load order;
- each child emits only its deltas: tokens that differ from the base, component overrides the base does not carry, prose only when the inlined type values differ, on-media only when the resolved surfaces differ, and the `color-scheme` preamble only when the child's delta introduces `light-dark()` the base did not have;
- a child that drops a base token is refused (a family build cannot express an unset);
- member order is canonicalized (base first, children sorted by invocation path): the family scope AND the recorded `@generated Command:` header both derive from it, so the same file set builds byte-identical output whatever the argument order, and the header stays reproducible as typed;
- `--check`, multi-file batch, `--out` guards, and single-file builds are unchanged; `--family` is refused with `--watch` and with fewer than two files.

In the measuring app this took each extending theme's stylesheet from ~20KB to ~2.1KB and let the app delete its hand-copied component-override blocks.

## Correctness

For any element under a member's scope root, each custom property resolves to the same value as before: the member's delta block (`:scope`, specificity 0-1-0) beats the family block's `:where(:scope)` (0-0-0) in the same layer, and properties the member does not change come from the family block, which restates the base's values verbatim. The theme-independent data-token defaults (`@layer astryx-base`) ride the base's stylesheet only — a member stylesheet is only ever loaded beside its base's. Nested theme roots keep today's semantics: the family block re-applies at any member root, and the `to ([data-astryx-theme])` limit still fences inner themes.

## Tests

`packages/cli/api/theme/build/build.family.test.mjs` (new): role classification from `extends` rather than argument order; refusal of a set with no unique base; the base's family block (member scope, `:where(:scope)`, whole-invocation header); a member carrying only its deltas; byte-identical output across shuffled argument orders; refusal of a dropped base token. Full `@astryxdesign/cli` suite run: the only failures are 7 pre-existing on `main` (reproduced on a pristine checkout).

### DIFF
--- a/.changeset/theme-build-family.md
+++ b/.changeset/theme-build-family.md
@@ -1,0 +1,8 @@
+---
+'@astryxdesign/core': patch
+'@astryxdesign/cli': patch
+---
+
+[feat] `astryx theme build --family <base> <children...>` builds a base theme and the themes that `extends` it as one unit: the base stylesheet restates the shared declarations once, scoped to every member (token block at zero specificity via `:where(:scope)`), and each member stylesheet carries only its own deltas — instead of every member restating the whole resolved token set. `defineTheme` records the resolved base as `__extends` to make the relationship visible to the build. Standalone and batch builds are unchanged.
+
+@bmaurer
--- a/packages/cli/api/theme/build/build.family.test.mjs
+++ b/packages/cli/api/theme/build/build.family.test.mjs
@@ -1,0 +1,164 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+/**
+ * `--family` builds: a base plus the themes that `extends` it, built as one
+ * unit. The base stylesheet restates its declarations ONCE, scoped to every
+ * member (`@scope` over the member value list, token block wrapped in the
+ * zero-specificity `:where(:scope)`), and each member stylesheet carries only
+ * its own deltas — instead of every member restating the whole resolved token
+ * set it inherited.
+ *
+ * These drive `resolveThemeFamily()` + `themeBuild({family})`, the API pair
+ * behind `astryx theme build --family`.
+ */
+
+import {describe, it, expect, beforeEach, afterEach} from 'vitest';
+import * as fs from 'node:fs';
+import * as path from 'node:path';
+import * as os from 'node:os';
+import {themeBuild} from './build.mjs';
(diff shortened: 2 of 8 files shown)


cixzhang (changes requested):
Thanks for tackling this duplication. The family artifacts are not safe to consume as emitted. Following the printed child install snippet loads only the child CSS, so inherited tokens disappear; loading the required base CSS after the child also lets base component rules replace child overrides. A zero-delta child returns no receipt, leaves stale `.css`/`.js`/`.d.ts` files untouched, and `--check` still exits 0.

Current theme contracts require extended themes to stay self-contained and runtime/static output to share one compiler. Please redesign this around one authoritative compiled-artifact model, with load order, cleanup, and check behavior proven for the family as a unit.

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks, the imported, inline, and partial-map indicator cases now pass. One valid composition still generates broken JavaScript: if the base and child each alias different imports to the same local indicator name, the child artifact emits both declarations and throws `SyntaxError: Identifier has already been declared` on import. Separate theme files can naturally reuse a local name. The generated bindings need to be collision-safe, with this base/child case covered end to end.

[Reviewed by Robohands]

PR #5691, opened 2026-08-30
docs(stepper): recast em-dash aside in progress bar anatomy description

josephfarina:
## What

The Stepper `progressBar` anatomy description in `packages/core/src/Stepper/Stepper.doc.mjs` used a paired em-dash to set off a nested aside. Recast the aside with parentheses so the prose reads plainly. The wording and meaning are unchanged; only the punctuation changed.

Before:
> Where a span is drawn by more than one segment — the on-track layouts split it between two steps, three when a content slot sits between them — the segments run in track order...

After:
> Where a span is drawn by more than one segment (the on-track layouts split it between two steps, three when a content slot sits between them), the segments run in track order...

## Checklist
- [x] `tsc --project packages/cli/tsconfig.template-docs.json --noEmit` passes
- [x] `astryx component Stepper --lang dense` renders cleanly, no em-dash
- [x] Prose-only change, one file, meaning preserved

Night Watch — Doc Reviewer

### DIFF
--- a/.changeset/stepper-progress-bar-anatomy-prose.md
+++ b/.changeset/stepper-progress-bar-anatomy-prose.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[docs] The Stepper progress bar anatomy description now sets its nested aside in parentheses instead of paired em dashes, so the sentence about multi-segment spans reads plainly in the CLI and doc site. Meaning unchanged. (#5691)
+
+@josephfarina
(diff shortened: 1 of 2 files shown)


cixzhang (changes requested):
Builders receive this copy edit through the published core docs and CLI, but this PR has no changeset, so the release would omit it from the package changelog. Please add a patch `[docs]` changeset for `@astryxdesign/core`, credited to `@josephfarina`. The prose and generated output otherwise look good.

[Reviewed by Robohands]

josephfarina:
Added the patch `[docs]` changeset for `@astryxdesign/core` in `.changeset/stepper-progress-bar-anatomy-prose.md`, credited to @josephfarina. `check:changesets` passes locally.

cixzhang (approved):
Thanks! The requested patch changeset is present, and the Stepper prose is unchanged from the reviewed head.

[Reviewed by Robohands]

PR #5693, opened 2026-08-30
fix(AppShell): keep the sticky header above field inputs

### LINKED ISSUE #5689: [Bug] Field inputs can paint above the sticky AppShell header
### Description

When an Astryx `TextArea` or another Field input is scrolled underneath a sticky `AppShell` header, the input can paint above the header instead of being occluded by it.

The header and Field input wrapper both currently create positioned stacking layers with `z-index: 1`:

- AppShell `headerSticky`: `position: sticky; zIndex: 1`
- Field `inputBase`: `position: relative; zIndex: 1`

Because the Field appears later in document order, it can paint above the header at the same stack level.

Expected: the sticky AppShell header remains above page content while scrolling.

Actual: the input surface and/or focus treatment visibly covers the header.

This also looks like a possible regression from #262, whose implementation notes describe the sticky header wrapper as using `z-index: 10`.

### Reproduction

1. Render an `AppShell` with `height="auto"` and a `topNav`.
2. Add enough main content to make the document scroll.
3. Place a `TextArea` in the main content.
4. Scroll until the `TextArea` reaches the sticky header.
5. Observe that the Field paints above the header.

Minimal shape:

```tsx
<AppShell height="auto" topNav={<TopNav>{/* nav content */}</TopNav>}>
  {/* enough content to scroll */}
  <TextArea label="Question" value={value} onChange={setValue} />
</AppShell>
```

The conflicting declarations are still present on `main`:

- https://github.com/facebook/astryx/blob/main/packages/core/src/AppShell/AppShell.tsx
- https://github.com/facebook/astryx/blob/main/packages/core/src/Field/inputStyles.stylex.ts

A component-level workaround is to raise the `app-shell-header` theme override above `1`, but the AppShell should establish a safe stacking order by default.

### Astryx Version

`@astryxdesign/core@0.4.5`; source inspection shows the same conflicting values on current `main` after the `0.5.1` release.

### Environment

Observed in a Chromium-based browser on a document-scrolling AppShell (`height="auto"`). The cause is CSS stacking order and is unlikely to be browser-specific.

Cypher-Aura-19:
## Summary

- restore the sticky AppShell header's stacking level so regular page content cannot paint above it while scrolling
- add regression coverage for Field inputs under an auto-height AppShell header

The header had been lowered to the same stacking level used by Field input wrappers, allowing detached and tooltip Field variants to paint over it.

## Testing

- targeted AppShell, Field, TextArea, and TextInput tests
- core typecheck
- ESLint
- changeset/use-client/sync checks
- verified the stacking behavior in Chromium with Storybook + Playwright

Fixes #5689

### DIFF
--- a/.changeset/keep-the-sticky-appshell-header-above-field-inpu-s4g6.md
+++ b/.changeset/keep-the-sticky-appshell-header-above-field-inpu-s4g6.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Field inputs no longer paint above the sticky AppShell header while scrolling (#5689). Field now contains its local stacking layers (the input surface's z-index and the attached status layer) behind an `isolation: isolate` boundary on the field surface, so they cannot compete with page-level stacking; the AppShell header keeps its normal stacking level.
+@Cypher-Aura-19
--- a/apps/storybook/stories/AppShell.stories.tsx
+++ b/apps/storybook/stories/AppShell.stories.tsx
@@ -18,6 +18,8 @@ import {
   SideNavSection,
 } from '@astryxdesign/core/SideNav';
 import {useMediaQuery} from '@astryxdesign/core/hooks';
+import {TextArea} from '@astryxdesign/core/TextArea';
+import {TextInput} from '@astryxdesign/core/TextInput';
 import * as stylex from '@stylexjs/stylex';
 import {
   HomeIcon,
@@ -52,6 +54,11 @@ const styles = stylex.create({
     display: 'flex',
     flexDirection: 'column' as const,
     gap: 16,
+  },
+  fieldStack: {
+    display: 'flex',
+    flexDirection: 'column' as const,
+    gap: 32,
   },
 });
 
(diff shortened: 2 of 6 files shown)


rubyycheung (approved):

cixzhang:
This exact head now conflicts with `main`, so GitHub cannot merge it. Please rebase or merge current `main` into the branch and resolve the conflicts. The resulting head will need fresh required checks and exact-head code/visual review before it can merge.

Cypher-Aura-19:
Rebased onto the latest `main` (`0ccad5b5e`) and force-pushed the updated branch.

The rebase applied cleanly with no manual conflicts, and the fix still matches `main`'s current local stacking approach. I also re-ran the relevant AppShell/Field/TextArea/TextInput tests, typecheck, ESLint, and repository checks successfully.

The PR diff is now focused to just the fix, regression coverage, Storybook case, and changeset.

cixzhang (changes requested):
Thanks for adding the focused reproduction. This head makes that overlap look correct, but it does so by raising AppShell to `z-index: 10`; the Field input surface still has `z-index: 1` without an isolation boundary. In Chromium, changing only the header back to `1` flips hit-testing to the field, so these components remain coupled through page-level stacking.

Could we contain the Field/input local layers with `isolation: isolate`, remove the AppShell header escalation, and make the regression test assert that ownership boundary instead of `header > input`?

[Reviewed by Robohands]

Cypher-Aura-19:
Addressed in `1c8b7d97b`.

The AppShell header escalation has been removed and `headerSticky` is back to `z-index: 1`. Field's local stacking layers are now contained at the Field root with `isolation: isolate`, covering attached, detached, and tooltip variants.

I also updated the regression coverage to assert the ownership boundary rather than comparing `header > input` z-index values, and added per-variant Field isolation coverage.

Verified the original reproduction in Chromium: without the Field isolation, detached fields reproduce the hit-testing issue; with the isolation in place, the sticky header correctly stays above all variants at its normal z-index.

Relevant tests, typechecks, ESLint, and repo checks are passing.

cixzhang (dismissed):
AppShell navigation stays clickable while Fields scroll underneath. Thanks for containing their stack in Field; Chromium passes attached and detached fields with the header at `z-index: 1`.

[Reviewed by Robohands]

cixzhang (approved):
AppShell navigation stays clickable while Field inputs scroll underneath. Fresh review of `1c8b7d97b28688ece590e44a0c7fa686a2eaeb12` confirms Field now owns its local stacking boundary without raising the header.

In real Chromium, the same overlapping story hit the Field when that isolation was disabled and hit the header on the unmodified exact head; both numeric z-index values remained `1`. Focused AppShell, Field, TextArea, and TextInput tests pass at the PR head and against current `main`.

[Reviewed by Robohands]

PR #5696, opened 2026-08-30
fix(ci): preserve visual reports during preview deploy

cixzhang:
## Summary
- preserve `pr/<number>/visual/` when replacing a PR's Storybook and Sandbox preview
- add a same-PR regression so preview deployment cannot silently delete published visual evidence

## Why
A live audit found [#5682](https://github.com/facebook/astryx/pull/5682) had a successful `visual-acceptance` status pointing to a 404. Its later preview deployment replaced all of `pr/5682/`, including the sibling `visual/` tree.

## Test plan
- `pnpm exec vitest run .github/scripts/lib/gh-pages-publisher.test.mjs`
- `pnpm exec prettier --check .github/scripts/lib/gh-pages-publisher.mjs .github/scripts/lib/gh-pages-publisher.test.mjs`
- `pnpm check:repo`
- `git diff --check`

### DIFF
--- a/.github/scripts/lib/gh-pages-publisher.mjs
+++ b/.github/scripts/lib/gh-pages-publisher.mjs
@@ -1541,9 +1541,12 @@ export async function publishManualVisualBaseline({
   refuse(`could not push the updated baseline after ${maxAttempts} attempts`);
 }
 
-function removeContents(destination) {
-  fs.rmSync(destination, {recursive: true, force: true});
+function replacePreviewContents(destination) {
   fs.mkdirSync(destination, {recursive: true});
+  for (const entry of fs.readdirSync(destination)) {
+    if (entry === 'visual') continue;
+    fs.rmSync(path.join(destination, entry), {recursive: true, force: true});
+  }
 }
 
 export async function publishPrPreview({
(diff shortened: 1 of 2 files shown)


cixzhang (commented):

Thanks—this fixes the deletion path, but a later preview deploy can still replace the trusted visual evidence a maintainer is trying to review. [`replacePreviewContents()` skips `visual`, then `copyContents(storybookDir, destination)` accepts it again](https://github.com/facebook/astryx/blob/5e31be6cbd4f73e80cb3fd39dba31b7ed3d8b377/.github/scripts/lib/gh-pages-publisher.mjs#L1544-L1590); my exact-head probe changed seeded “newer trusted evidence” to PR-controlled artifact bytes. Please make `pr/<number>/visual/` write-protected for both removal and copy, and add that collision as a regression. Retry, idempotency, other-PR, and cleanup paths otherwise held.

[Reviewed by Robohands]

cixzhang (commented):
Semantic verdict: approve

Thanks—this is an approve once you undraft. A reserved `visual` artifact now rejects before any destination mutation; retry, idempotency, cleanup, and cross-PR isolation still hold.

[Full review](https://github.com/cixzhang/astryx/wiki/Review-5696#round-2)

[Reviewed by Robohands]

PR #5698, opened 2026-08-30
fix(ci): checkout publisher for preview recovery

cixzhang:
## Summary
- check out trusted `main` before the manual preview recovery workflow invokes the shared publisher
- enforce checkout ordering in the workflow contract test

## Why
A live recovery attempt for four recent PR previews failed because `redeploy-preview.yml` downloaded the static artifacts but never checked out `.github/scripts/gh-pages-publisher.mjs`.

## Test plan
- `pnpm exec vitest run .github/scripts/visual-gate/workflow-concurrency.test.mjs`
- `pnpm exec actionlint -ignore 'SC2129' .github/workflows/redeploy-preview.yml`
- `pnpm exec prettier --check .github/workflows/redeploy-preview.yml .github/scripts/visual-gate/workflow-concurrency.test.mjs`
- `pnpm check:repo`
- `git diff --check`

### DIFF
--- a/.github/scripts/visual-gate/workflow-concurrency.test.mjs
+++ b/.github/scripts/visual-gate/workflow-concurrency.test.mjs
@@ -423,6 +423,11 @@ describe('visual acceptance workflow concurrency', () => {
 
     expect(deployPreview).toContain('gh-pages-publisher.mjs pr-preview');
     expect(redeployPreview).toContain('gh-pages-publisher.mjs pr-preview');
+    expect(redeployPreview).toContain('uses: actions/checkout@v7');
+    expect(redeployPreview).toContain('ref: main');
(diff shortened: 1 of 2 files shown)


cixzhang (commented):

This looks right for the recovery path. The failed manual runs were all getting through artifact download and then dying because the runner had no checked-out `.github/scripts/gh-pages-publisher.mjs`; this head checks out trusted `main` before the PR lookup, artifact downloads, and publisher invocation, so the job has the publisher without checking out fork code.

One non-blocking thing I’d tighten while you’re here: the regression locks checkout before `gh-pages-publisher.mjs`, but not specifically before the “Locate the PR's CI run” step. The workflow itself has the right order, so I would not hold the draft on it; the test would just preserve the full trust-boundary invariant more directly.

[Reviewed by Robohands]

PR #5700, opened 2026-08-30
fix(cli): throw warning for unrecognized template-asset extensions

veerjain-1:


cixzhang:
Thanks @veerjain-1 for identifying this unsafe fallback and contributing a fix. The same issue is now fully covered by #6039, whose current approved head fails with the fixture path, classifies the complete final suffix (including multi-dot names and m4v), and includes focused tests plus the release note. I’m closing this PR as superseded rather than asking you to duplicate that work.

[Reviewed by Robohands]

PR #5707, opened 2026-08-30
docs: define async option-source states

cixzhang:
## Problem

Selector and MultiSelector already keep provided options usable while input `isLoading` is true, but they use that value-loading flag to suppress empty/no-results presentation and announcements as if options had not arrived. They also lack an explicit initial option-source pending state and a defined accessible distinction between pending and completed empty/read-only states.

## Change

Record Cindy's approved direction: provided options determine availability, input `isLoading` does not suppress them, and zero provided options is a read-only/no-choice state. Initial option-source pending remains explicit rather than being inferred from an empty option set.

Define the required loading, read-only, and empty accessibility outcomes and verification without claiming the runtime implementation is resolved. Keep one narrow human API question: the public shape for explicit initial pending.

This PR is a system-spec-only delta on current `main`; the landed `component:Selector` contract remains the source of shipped behavior until implementation updates it.

## Risk

Documentation only. No runtime or public API changes.

## Testing

- `pnpm check:knowledge`
- `pnpm check:repo`
- `pnpm -F @astryxdesign/docsite build`
- `git diff --check`

### DIFF
--- a/docs/specs/AST-001/spec.md
+++ b/docs/specs/AST-001/spec.md
@@ -1,0 +1,147 @@
+---
+schema_version: 1
+template_version: 1
+kind: system-spec
+id: spec:AST-001
+authority: current
+archive_reason: null
+superseded_by: null
+approved_by: cixzhang
+approved_at: 2026-08-31
+phase: accepted
+owners: [cixzhang, imdreamrunner]
+affects_architecture: [architecture:public-component-api]
+affects_families: [family:input-fields]
+affects_contributing: []
+affects_consumer_docs: [Selector, MultiSelector]
+---
+
+# Async option-source states
+
+## Intent
+
+People using Selector or MultiSelector should be able to use every option the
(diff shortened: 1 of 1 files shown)


docs/specs/AST-001/spec.md
```diff
@@ -0,0 +1,114 @@
...
+- Unsupported behavior: a component must not overload input `isLoading` as an
+  option-source state.
+- Browser evidence: interactive and announcement behavior requires real
+  Chromium verification in addition to unit tests.
+
+## Current-state impact
+
+- Selector and MultiSelector currently use input `isLoading` to suppress empty
+  and search-result output as if options had not arrived. That coupling must be
+  removed when this spec is implemented.
```
cixzhang: Current behavior likely also violates expected behavior for isLoading and hasn't had the accessibility story figured out. In adherence to this spec this could instead be moved to rely on options provided where if 0 options are provided the selectors are in read only mode.

PR #5711, opened 2026-08-30
feat(Table): expose row selection to theming

### LINKED ISSUE #5425: Table: a theme cannot key off row selection — themeProps('table-row') publishes no selected state
`themeProps('table-row')` (`packages/core/src/Table/TableRow.tsx:135`, `:174`) publishes nothing about selection, so a theme cannot key off a selected row — the checked-row wash is written directly and can only be opted out of, not restyled.

Raised as a non-blocking follow-up during review of [#5310](https://github.com/facebook/astryx/pull/5310), which merged.

Cypher-Aura-19:
## Summary

- expose selected Table rows through the `table-row` theming target
- make the checked-row wash retintable with `--table-row-selected-background`
- preserve the existing fine-grained selection-store update path and pinned-column overlay behavior
- add regression coverage for selection transitions, theme targeting, opt-out behavior, and React rerenders

Selected rows now publish the `selected` state without routing selection through React row state, so themes can target the row without changing the existing render-performance model.

## Testing

- selection plugin tests
- Table theming target/derived-var guards
- core typecheck
- strict ESLint
- changeset validation
- core build
- verified the theme build accepts the `table-row.selected` target

Fixes #5425

### DIFF
--- a/.changeset/table-expose-row-selection-to-theming-5425-hgdx.md
+++ b/.changeset/table-expose-row-selection-to-theming-5425-hgdx.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] Table: expose row selection to theming (#5425)
+@Cypher-Aura-19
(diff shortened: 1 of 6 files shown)


humbertovirtudes (changes requested):
The new public table-row.selected state and --table-row-selected-background contract need exact current owner approval. Please publish state through the canonical theme-state mechanism rather than bespoke imperative class and data mutation, and provide real-browser computed-paint evidence across shipped themes and sticky-column composition. The default fallback and focused unit coverage are otherwise coherent.

PR #5714, opened 2026-08-30
test(a11y): generate component contrast coverage

rubyycheung:
## Summary
- add a component-generic build-time WCAG audit registry, generator, and freshness guard
- register pressable-control rendering profiles as the first audit module
- resolve semantic and theme-local tokens, alpha colors, and state overlays in paint order
- generate shared theme coverage consumed by the Accessibility tab without adding code to runtime bundles
- keep generation deterministic, import-safe, atomic, and explicitly checked by docsite builds

A future component family adds one audit module and registers it. The shared generator and CI guard do not need a parallel implementation.

## Current documented failures
- dark SegmentedControl unselected text reaches 4.12:1 at rest and 3.52:1 on hover

These are reported, not changed. This PR does not alter theme colors or component styling.

## Tests
- `pnpm generate:component-accessibility-coverage`
- `pnpm check:component-accessibility-coverage`
- `pnpm exec vitest run scripts/check-component-accessibility-coverage.test.mjs` (11 passed)
- `pnpm --filter @astryxdesign/docsite test` (497 passed)
- `pnpm --filter @astryxdesign/docsite typecheck`
- `pnpm --filter @astryxdesign/cli typecheck:authoring`
- `pnpm --filter @astryxdesign/core typecheck`
- `pnpm build`
- `pnpm check:repo`


cixzhang (changes requested):
These tables can report WCAG AA Pass while a supported state is untested. The registry only validates its own handwritten list, and “Not measured” removes the selected surface for icon-only ToggleButton and SegmentedControl even though those paths have no label-weight cue. The published checker also reconstructs StyleX source instead of measuring rendered browser output, so its color pairs cannot prove what users actually see. Please make applicability and the component/state inventory independently complete, add mutation coverage that fails when a required row disappears, and verify paint-dependent ratios in a real browser before publishing Pass/Fail.

[Reviewed by Robohands]

PR #5724, opened 2026-08-30
fix(core): size selectors and menus to available space

cixzhang:
## Why

Selectors and menus cap long option surfaces at 300px. That wastes space in tall viewports and can still put content beyond the viewport in short ones.

The earlier stack proposed a public Layer sizing prop. The current Layer architecture keeps viewport sizing and overflow with Popover or the component that owns the painted surface, so this refresh removes that public dependency.

## What

Selector, MultiSelector, ComplexSelector, DropdownMenu, DropdownMenuSubMenu, and ContextMenu now size long surfaces to the block space their anchor leaves in the viewport.

- Popover exposes the constraint only to package-internal component owners; Layer's public geometry API is unchanged.
- Block-side surfaces choose the roomier side, stay within the viewport, and scroll internally.
- Searchable selectors keep the search row fixed while only the options scroll.
- Existing inline viewport and safe-area constraints for menus are preserved.
- The Selector readiness story now shows an open searchable 40-option list. DropdownMenu's tall-content story now documents the available-space behavior instead of the old 300px cap.

No component props are added.

## Risk

The maximum open height is viewport-dependent instead of fixed. Short surfaces retain their natural height. Long surfaces may grow beyond 300px when space exists or become shorter when the viewport requires it.

## Testing

- 538 focused tests across the affected components
- core TypeScript check
- focused ESLint and Prettier checks
- RTL audit coverage for all six affected components: selector trigger order, menu trigger order, ContextMenu logical-edge placement, and Popover direction inheritance
- real Chromium at 320px and 900px viewport heights: both examples stayed inside the viewport; the tall menu grew to 536px when space allowed; short menus kept natural height; searchable options scrolled while the search row stayed fixed


cixzhang:
/accept-visual 33597008816/1 The affected default frames preserve their existing appearance across themes. The intended change is limited to open long surfaces; real Chromium checks at 320px and 900px confirmed viewport containment, roomier-side selection, natural short-menu height, internal scrolling, and a fixed search row.

cixzhang (commented):

Thanks — the available-space behavior works for long lists, but it changes Selector’s default non-search contract. At 320×320, the selected Banana row moves from covering the trigger (center delta −1px) to sitting 31px above it, so the stationary pointer highlights Orange instead. Disabling `constrainToAvailableBlockSize` restores the current overlay in LTR and RTL. Please keep viewport fitting without replacing the selected-row overlay (`Selector.tsx:1008`).

[Reviewed by Robohands]

ernestt (approved):

cixzhang:

PR #5733, opened 2026-08-31
docs: define icon resolution and component slots

cixzhang:
## Why

Theme authors need to change one component-owned icon role without changing every use of the same shared icon.

## What

- defines the semantic `componentIcons` slot-to-name layer separately from general icon resolution
- records typed public slot ownership, null suppression, fallback precedence, and component-local documentation
- preserves coexistence with shipped extension keys; this draft decides no component migration
- delegates theme authoring/application/compilation and Icon rendering APIs to their existing owners

## Risk

Documentation only. This draft is not current policy until owner approval.

## Testing

- `pnpm check:knowledge`
- `pnpm exec prettier --check docs/architecture/icon-resolution-and-component-slots.md`
- `git diff --check`

No Changeset: this record does not publish package behavior.

### DIFF
--- a/docs/architecture/icon-resolution-and-component-slots.md
+++ b/docs/architecture/icon-resolution-and-component-slots.md
@@ -1,0 +1,283 @@
+---
+schema_version: 1
+template_version: 1
+kind: architecture
+id: architecture:icon-resolution-and-component-slots
+authority: current
+archive_reason: null
+superseded_by: null
+approved_by: cixzhang
+approved_at: 2026-08-30
+owners: [cixzhang, imdreamrunner]
+applies_to:
+  [
+    packages/core/src/Icon/,
+    packages/core/src/theme/defineTheme.ts,
+    packages/core/src/Selector/,
+    packages/cli/,
+  ]
+verified_by:
+  [
+    packages/core/src/Icon/globalIconRegistry.test.tsx,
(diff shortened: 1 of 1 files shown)


docs/architecture/icon-resolution-and-component-slots.md
```diff
@@ -0,0 +1,258 @@
...
+archive_reason: null
+superseded_by: null
+approved_by: null
+approved_at: null
+owners: [cixzhang, imdreamrunner]
+applies_to:
+  [
+    packages/core/src/Icon/,
+    packages/core/src/theme/defineTheme.ts,
+    packages/core/src/Selector/,
```
cixzhang, line 16: This is too specific? More components will have component icons

docs/architecture/icon-resolution-and-component-slots.md
```diff
@@ -0,0 +1,258 @@
...
+authority: draft
+archive_reason: null
+superseded_by: null
+approved_by: null
+approved_at: null
+owners: [cixzhang, imdreamrunner]
+applies_to:
+  [
+    packages/core/src/Icon/,
+    packages/core/src/theme/defineTheme.ts,
```
cixzhang, line 15: There should also be some shared icon resolver modules. Include in this spec?

PR #5735, opened 2026-08-31
docs: define overlay dismissal family contract

cixzhang:
## Why

Astryx already routes Escape and platform close requests through one topmost-layer stack. That shared behavior had no family record, so its ownership and membership boundaries were implicit in code and tests.

## What

Defines the current overlay-dismissal family contract: every layered surface for which Escape or a platform close request is valid belongs and must use the shared stack. Current local-listener implementations are adoption gaps, not exclusions. Component-owned input popups are enumerated; ancillary composition alone does not create membership.

Confirms that an active `useFocusTrap` with `onEscape` registers through `useLayerDismissal`. A visual-only Layer outline with no dismissal behavior is the canonical nonparticipant.

Descendant depth remains conditional on a member providing `LayerDepthProvider`. The mapped tests do not exercise a real portal, so the record requires a portal fixture before claiming verified portal coverage. Focus, modality, positioning, outside dismissal, and open-state APIs remain component- or family-owned.

## Risk

Documentation only. No component behavior, public API, theming surface, or package output changes. No Changeset is required.

## Testing

- `pnpm check:knowledge`
- Prettier check for the record
- 67 focused dismissal and focus-trap tests
- Repository pre-commit checks

### DIFF
--- a/docs/families/overlay-dismissal.md
+++ b/docs/families/overlay-dismissal.md
@@ -1,0 +1,245 @@
+---
+schema_version: 1
+template_version: 1
+kind: family
+id: family:overlay-dismissal
+authority: current
+archive_reason: null
+superseded_by: null
+approved_by: cixzhang
+approved_at: 2026-08-30
+owners: [cixzhang, imdreamrunner]
+review_triggers: [behavior, accessibility]
+verified_by:
+  [
+    packages/core/src/Layer/useLayerDismissal.test.tsx,
+    packages/core/src/Layer/layerDismissalInvariants.test.tsx,
+    packages/core/src/Layer/layerDismissalFamilies.test.tsx,
+    packages/core/src/hooks/useFocusTrap.test.tsx,
+  ]
+members:
+  [
+    component:Dialog,
+    component:AlertDialog,
+    component:Popover,
+    component:DropdownMenu,
+    component:DropdownMenuSubMenu,
+    component:MoreMenu,
+    component:Tooltip,
+    component:HoverCard,
+    component:Lightbox,
+    component:MobileNav,
+    component:BottomSheet,
+    component:BottomSheetSwitcher,
+    component:CommandPalette,
+    component:ContextMenu,
+    component:PowerSearchEditPopover,
+    component:Drawer,
+    component:BreadcrumbItem,
+    component:ChatComposerInput,
+    component:ComplexSelector,
+    component:DateInput,
+    component:DateRangeInput,
+    component:DateTimeInput,
+    component:Selector,
+    component:MultiSelector,
+    component:PowerSearch,
+    component:BaseTypeahead,
+    component:Typeahead,
+    component:Tokenizer,
+    component:SideNavHeading,
+    component:SideNavItem,
+    component:TabMenu,
+    component:TopNavHeading,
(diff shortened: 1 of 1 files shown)


docs/families/overlay-dismissal.md
```diff
@@ -0,0 +1,172 @@
...
+A nested surface must not close its host on the same Escape press or platform
+close request, including when a portal separates their DOM positions.
+
+## Membership rule
+
+A component belongs when it presents a dismissible surface and that surface
+participates in the shared dismissal stack while it is present. A component may
+adopt the contract directly or through another member that owns its surface.
+
+- **Members:** Dialog, AlertDialog, Popover, DropdownMenu, MoreMenu, Tooltip,
```
cixzhang: Likely any inputs with popovers should be included if they don't compose popovers

docs/families/overlay-dismissal.md
```diff
@@ -0,0 +1,172 @@
...
+  HoverCard, Lightbox, and MobileNav.
+- **Collaborators:** `useLayerDismissal`, `layerStack`, and
```
cixzhang: I believe focus trap also includes this behavior now but would be good to check

docs/families/overlay-dismissal.md
```diff
@@ -0,0 +1,172 @@
...
+  `LayerDepthProvider` own the shared protocol. Focus traps may register a
+  surface but do not define its component-level focus policy.
+- **Excluded:** BottomSheet, CommandPalette, ContextMenu,
```
cixzhang: Hmm these should participate in the layer stack. I don't think their exclusion was intentional

cixzhang: One true exception is if we later add a visual only outline using our layers. That would be an example of a non participating component

docs/families/overlay-dismissal.md
cixzhang: Applied. The membership rule now includes component-owned input popups and enumerates ChatComposerInput, ComplexSelector, DateInput, DateRangeInput, DateTimeInput, Selector, MultiSelector, PowerSearch, BaseTypeahead, Typeahead, and Tokenizer. It also distinguishes inputs that only compose an ancillary Tooltip.

docs/families/overlay-dismissal.md
cixzhang: Confirmed and recorded. Active useFocusTrap instances with onEscape register through useLayerDismissal; traps without onEscape stay off the stack. The focused hook test is now part of verified_by.

docs/families/overlay-dismissal.md
cixzhang: Applied as the normative rule. BottomSheet, CommandPalette, ContextMenu, DropdownMenuSubMenu, PowerSearchEditPopover, and Lab Drawer are members with local-only adoption gaps, not exclusions. BottomSheetSwitcher is recorded as partial because modal mode joins through useFocusTrap while non-modal mode remains local.

cixzhang: Applied in DEC-1 and the membership rule. A visual-only outline rendered through Layer with no Escape or platform-close behavior is the canonical nonparticipating example.

PR #5752, opened 2026-08-31
feat(theme): add local tokens and refine Neutral color roles

rubyycheung:
## Summary
- implements the accepted AST-006 `localTokens` contract from #5801 without changing existing `tokens` behavior
- preserves exact theme-local names through runtime CSS, static builds, inheritance, and built-theme metadata
- migrates Neutral's reusable status-fill and tint roles to the `--astryx-theme-neutral-*` namespace
- completes the Neutral warning mappings for Stepper and ProgressBar
- keeps tonal-palette metadata work in #5668 and row-status infrastructure in #5754

## Theme-local token contract
- `localTokens` is optional; themes that omit it retain their previous object shape and behavior
- values accept the complete existing `TokenValue` contract: a CSS string or `[light, dark]` tuple
- declarations use one complete exact custom-property name
- enrolled inheritance preserves exact declarations and ownership metadata in source and built themes
- recursive validation rejects malformed or foreign declarations, undeclared references, cycles, and invalid lineage before CSS is emitted
- portable `TokenName`, token helpers, and Core component source remain unchanged

## Neutral mappings
- Badge: info, success, warning, error fills
- StatusDot and AvatarStatusDot
- Stepper: accent, success, warning, error
- Table row status dots/icons
- Banner tint and interaction overlays
- ProgressBar: accent, success, warning, error

## Contrast receipts
Filled status label pairs:
- accent: 4.57:1 light / 6.43:1 dark
- success: 5.02:1 light / 6.38:1 dark
- warning: 11.58:1 light / 11.58:1 dark
- error: 5.29:1 light / 6.34:1 dark

The repository Badge contrast guard resolves `localTokens` and passes for all Neutral semantic variants. Trusted exact-head screenshots are produced by the visual CI workflow for final Neutral ratification.

## Validation
- `pnpm -F @astryxdesign/core typecheck`
- focused core theme tests: 155 passed
- focused CLI runtime/static/local-token tests: 6 passed
- Neutral, Badge contrast, and template-contract tests: 16 passed
- `pnpm -F @astryxdesign/core build`
- `pnpm -F @astryxdesign/theme-neutral build`
- `pnpm -F @astryxdesign/cli typecheck:strict`
- `pnpm -F @astryxdesign/cli typecheck:authoring`
- `pnpm -F @astryxdesign/cli typecheck:template-docs`
- `pnpm bundle:cli-themes`
- `pnpm check:sync`
- `pnpm check:changesets`
- `pnpm check:package-boundaries`
[...]

PR #5754, opened 2026-08-31
Add Table row status theming target

rubyycheung:
## Summary
- adds an `astryx-table-row-status` theming target with `color` and `presentation` visual props
- keeps dot/icon behavior caller-controlled while exposing the wrapper for scoped theme overrides
- documents the new target for Table and useTableRowStatus

## Stack
- Base: #5668 palette/theme infra
- Consumer: #5752 Neutral theme color roles

## Validation
- pnpm vitest run packages/core/src/Table/plugins/rowStatus/useTableRowStatus.test.tsx
- pre-commit checks from commit hook

### DIFF
--- a/.changeset/table-row-status-theming-target.md
+++ b/.changeset/table-row-status-theming-target.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] Add a theming target for Table row status indicators.
+
+@rubyycheung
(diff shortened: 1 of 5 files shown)


rubyycheung:
Superseded by #5848, which contains the row-status theming target on a standalone branch based directly on `main`.

While rebasing, the replacement preserved newer `main` behavior where success, warning, and error use themed semantic icons by default. The theming target now layers on top of that behavior instead of replacing it.

PR #5762, opened 2026-08-31
fix(Button): inherit default icon size

rubyycheung:
## Summary

- provide Button-sized defaults to descendant Astryx Icons
- render 16px icons in small and medium buttons and 20px icons in large buttons
- preserve explicit Icon size overrides
- document the behavior and add IconButton regression coverage

## Test plan

- `pnpm vitest run packages/core/src/Icon/Icon.test.tsx packages/core/src/Button/Button.test.tsx packages/core/src/IconButton/IconButton.test.tsx`
- `ASTRYX_STRICT_LINT=1 pnpm exec eslint packages/core/src/Button/Button.tsx packages/core/src/Icon/Icon.tsx packages/core/src/Icon/IconDefaultSizeContext.ts packages/core/src/IconButton/IconButton.test.tsx`
- `pnpm check:repo`

### DIFF
--- a/.changeset/button-icon-default-size.md
+++ b/.changeset/button-icon-default-size.md
@@ -1,0 +1,8 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Size an Astryx Icon without an explicit size to match its Button or
+IconButton: 16px for small and medium controls, and 20px for large controls.
+
+@rubyycheung
--- a/apps/storybook/rtl-audit/targets.json
+++ b/apps/storybook/rtl-audit/targets.json
@@ -11,6 +11,17 @@
     }
   },
   {
+    "component": "Button",
+    "storyId": "core-button--icon-with-text",
+    "dims": [
+      "D2"
+    ],
+    "selectors": {
+      "prev": "[data-testid=\"button-leading-icon\"]",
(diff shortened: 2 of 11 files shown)


cixzhang (changes requested):
Thanks for fixing the compact Button sizing. The Button→Icon relationship now lives in two writable maps, and the new context separately copies Icon’s size union; the test checks only `data-size`. A later scale change can pass while those contracts drift. Could one source drive the wrapper and inherited default, with their rendered sizes tested together?

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks — the shared size source and rendered-size coverage now resolve the previous drift risk. One public contract still disagrees: `Icon` documents omitted `size` as always defaulting to `md`, but `Icon.tsx:252` now lets a parent replace that default. Builders reading Icon’s API will not know their icon can render smaller. Could we update Icon’s English, Chinese, dense, and source docs to describe the contextual default?

[Reviewed by Robohands]

cixzhang (approved):
Verified on exact head bd797278d319: contextual Button icon sizing matches the current size-cascade and Button/Icon contracts; explicit Icon sizes and the standalone md fallback remain unchanged. Code Bunny and required CI are green.

PR #5763, opened 2026-08-31
docs(core): record Layout aggregate anatomy

cixzhang:
## Why

Layout exposes five related public theming targets but lacked one stable aggregate anatomy that explains their conceptual ownership. Builders and maintainers need that map without treating current layout gaps as fixed behavior.

## What

- Documents Page shell, Header, Panel, Content area, and Footer in the canonical Layout consumer doc; start and end panels share one Panel part.
- Adds a v3 draft Layout component contract with the exact anatomy-to-target map and links to the current layout, padding, theming, and public API records.
- Records the shipped LayoutContent/LayoutPanel padding-publication mismatch and the caller obligation that avoids a double divider beside ResizeHandle.

## Risk

Documentation and maintainer metadata only. No runtime, DOM, API, target, alias, or generated-output changes. No Changeset.

## Testing

- `pnpm exec vitest run packages/core/src/Layout apps/docsite/src/__tests__/anatomy-helpers.test.ts` (92 tests)
- `pnpm check:repo`

### DIFF
--- a/packages/core/src/Layout/Layout.doc.mjs
+++ b/packages/core/src/Layout/Layout.doc.mjs
@@ -1,5 +1,35 @@
 // Copyright (c) Meta Platforms, Inc. and affiliates.
 
+/** @type {import('@astryxdesign/cli/authoring').ComponentAnatomyElement[]} */
+const anatomy = [
+  {
+    name: 'Page shell',
+    required: true,
+    description: 'Outer Layout container that arranges the named page regions.',
+  },
+  {
+    name: 'Header',
+    required: false,
+    description: 'Optional top region for page or application header content.',
+  },
+  {
+    name: 'Panel',
+    required: false,
+    description:
+      'Optional start or end side region; both positions use the same LayoutPanel part.',
+  },
+  {
+    name: 'Content area',
+    required: false,
+    description: 'Optional central region for the page content.',
+  },
+  {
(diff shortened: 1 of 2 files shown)


packages/core/src/Layout/Layout.doc.mjs
```diff
@@ -1,5 +1,35 @@
 // Copyright (c) Meta Platforms, Inc. and affiliates.
 
+/** @type {import('@astryxdesign/cli/authoring').ComponentAnatomyElement[]} */
+const anatomy = [
+  {
+    name: 'Page shell',
```
cixzhang, line 6: App shell component is intended to be the page shell while Layout component is intended to be for general layout primitive with 5 slots on the page.

packages/core/src/Layout/Layout.spec.md
```diff
@@ -0,0 +1,199 @@
...
+
+## Intent
+
+Layout provides the page shell that arranges optional Header, Panel, Content
+area, and Footer regions. This draft records the current aggregate consumer
+anatomy and theming ownership without changing runtime behavior, DOM, styling,
+targets, or public API.
+
+## Compatibility and migration
+
```
cixzhang, line 44: There are some other participating components in the same container system like section and table and tool bar and divider

PR #5765, opened 2026-08-31
Deduplicate DropdownMenu divider theme metadata

cixzhang:
## Why
`astryx theme targets dropdown-menu-divider` reported the same target twice because both the parent and child docs claimed ownership.

## What
- Keep the divider theme target owned by `DropdownMenu` only.
- Complete the parent component roster with every existing public DropdownMenu child doc.
- Add a CLI regression asserting one divider row owned by `DropdownMenu`.

## Risk
Documentation metadata and tests only; no runtime behavior changes.

## Testing
- Focused DropdownMenu and CLI theme-target tests (138 tests)
- Core, lab, charts, and richtext docs typechecks
- CLI template-docs typecheck
- `pnpm check:repo`

### DIFF
--- a/.changeset/dropdown-divider-theme-target-dedup.md
+++ b/.changeset/dropdown-divider-theme-target-dedup.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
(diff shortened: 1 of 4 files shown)


cixzhang:
Superseded by #5767, which preserves direct child theming docs while deduplicating explicit parent/child target rows in discovery.

PR #5766, opened 2026-08-31
fix(Spinner): keep theme target on status element

cixzhang:
## Why

A labelled Spinner published its stable theme target on a structural layout wrapper instead of the status element that paints the ring. That made direct target properties describe the wrapper rather than the Spinner’s owned visual part.

## What

- Keep the existing `spinner` target and reflected size/shade state on the `role="status"` element in labelled and unlabelled states.
- Keep the wrapper as the public root for refs, rest props, `data-testid`, and consumer styling.
- Add an unregistered private bridge from the labelled root to the ring so root `style`, `className`, and `xstyle` custom-property overrides keep consumer precedence even when the active theme declares the same variables on the child target.
- Add unit coverage plus a committed Chromium story/guard for target ownership and the full root/theme cascade.

## Compatibility

This is a patch fix. The documented target name, props, public root, ref destination, DOM pass-through, and styling APIs are unchanged. The selector now remains on the semantic element that owns and paints the Spinner, as required by the current component-theming contract; the structural wrapper was never documented as a separate target or anatomy part.

## Testing

- Full package build and actual Storybook production build via `pnpm storybook:build`.
- Generated probe-theme reach in Chromium 149: labelled and unlabelled `spinner` targets reached, with no failure or shadowing.
- Focused Chromium matrix: all 32 labelled/unlabelled size × shade states reached the target and public-variable paint path, including `xl`.
- Committed Chromium guard: theme-only labelled/unlabelled fallbacks and labelled root `style`, `xstyle`, and `className` overrides all reached computed SVG geometry and stroke paint. Root overrides still win when the active theme declares the same variables.
- All 10 documented public component variables pass the real-browser theme reachability guard.
- 600 focused unit/theming/probe tests and the repository checks pass.
- `visual:probe-theme:check` currently reports existing unrelated fixture drift: regeneration adds only Selector and MultiSelector `readonly` selectors. No Spinner fixture change remains after rebasing, so those updates are intentionally excluded.

### DIFF
--- a/.changeset/spinner-labelled-theme-target.md
+++ b/.changeset/spinner-labelled-theme-target.md
@@ -1,0 +1,15 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Keep Spinner's documented `spinner` target on the status element that
+paints the ring, including when a visible label adds a structural wrapper. Theme
+size and shade overrides now reach the same visual part in labelled and
+unlabelled states.
+
+The labelled wrapper remains the public root for refs, DOM props, and consumer
+styling. An unregistered private bridge carries root `style`, `className`, and
+`xstyle` variable overrides into the ring, so they keep consumer precedence even
+when an active theme declares the same public variables on the status target.
+
+@cixzhang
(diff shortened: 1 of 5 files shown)


cixzhang:

PR #5768, opened 2026-08-31
Correct PowerSearch theme target ownership

cixzhang:
## Why

PowerSearch's pointer wrapper only forwards refs and anchors the edit popover; it does not paint the control. Publishing that wrapper as `astryx-power-search` gives theme authors a target on the wrong ownership layer while the visible surface is already owned by Tokenizer.

## Dependency

Depends on #5623. Do not land this PR until #5623 is settled and this branch is rebased; that work overlaps PowerSearch target ownership.

## What

- Keep the wrapper, forwarded ref, popover anchor, Tokenizer, and DOM shape unchanged while removing the wrapper's theme class.
- Remove the pointer-owned target metadata and regenerate the canonical probe theme.
- Keep the focused ref contract and assert that the wrapper directly owns a distinct Tokenizer target. Add a breaking changeset that directs themes to `tokenizer`.

## Risk

This removes the documented `.astryx-power-search` selector from the pointer path. Defer this PR and land it only with the next minor release unless the owner explicitly chooses patch compatibility treatment.

## Testing

- `pnpm vitest run packages/core/src/PowerSearch`
- `pnpm vitest run packages/core/src/theme/themingTargets.test.ts`
- `pnpm -F @astryxdesign/theme-probe build`
- `pnpm lint` (includes `pnpm check:repo`)

### DIFF
--- a/.changeset/power-search-target-ownership.md
+++ b/.changeset/power-search-target-ownership.md
@@ -1,0 +1,9 @@
+---
+'@astryxdesign/core': minor
+---
+
+[breaking] Remove the non-visual PowerSearch wrapper's `astryx-power-search` theme target.
+
+@cixzhang
+
+Themes that target `power-search` should migrate those styles to the existing Tokenizer `tokenizer` target. PowerSearch does not publish a replacement root target until touch-surface ownership is defined.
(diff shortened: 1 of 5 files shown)


cixzhang:

PR #5738, opened 2026-08-31
fix(field): separate tooltip status presentation

cixzhang:
## Why

Direct `FieldStatus` renders only attached and detached message presentations. The `tooltip` value belongs to the Field family, where composing controls consume it before rendering `FieldStatus` and surface the message from an on-field affordance.

This implements the boundary approved in #5736.

## What

- narrow `FieldStatusVariant` to `attached` and `detached`
- add `FieldStatusPresentation` for Field-family `statusVariant` props, including the `tooltip` sentinel
- add a conservative migration codemod only for casts directly on known Field-family `statusVariant` props
- add a breaking Changeset with manual guidance for ambiguous root imports and direct JSX

No runtime rendering, DOM, styling, or ARIA behavior changes.

## Compatibility

This is a TypeScript breaking change for direct `FieldStatus variant=\"tooltip\"` usage and for Field-family code that explicitly annotated values as `FieldStatusVariant`. Existing Field-family `statusVariant=\"tooltip\"` behavior remains unchanged; those annotations move to `FieldStatusPresentation`.

## Testing

- `pnpm test --run packages/core/src/FieldStatus/FieldStatus.test.tsx packages/core/src/Field/Field.test.tsx packages/core/src/hooks/useInputStatusIcon.test.tsx` (95 passed)
- migration codemod tests (10 passed, including parameter/local shadowing and idempotence)
- Core, Lab, RichText, and strict CLI typechecks
- `pnpm check:repo`

PR #5769, opened 2026-08-31
fix(FieldStatus): cover rounded input corners

rubyycheung:
## Summary

- extend attached status backgrounds behind the lower half of their controls
- derive the overlap from the control size, so rounded and `9999px` pill radii remain covered
- offset the added overlap with matching top padding so message placement stays unchanged
- add regression coverage for small, medium, and large controls

## Test plan

- `pnpm vitest run packages/core/src/Field/Field.test.tsx packages/core/src/FieldStatus/FieldStatus.test.tsx`
- `ASTRYX_STRICT_LINT=1 pnpm exec eslint packages/core/src/Field/Field.tsx packages/core/src/Field/Field.test.tsx packages/core/src/FieldStatus/FieldStatus.tsx packages/core/src/FieldStatus/FieldStatus.test.tsx`
- `pnpm check:repo`

## Related work

PR #5738 also touches Field and FieldStatus for the separate tooltip-presentation boundary and may require a small rebase depending on merge order.


cixzhang (changes requested):
Thanks for the fix. In the documented custom/native-control path, clicking the lower half of the input now hits FieldStatus and drops focus to the page. Please keep the attached background from taking pointer input, then retain the size/theme visual coverage.

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks, the click path is fixed. The documented custom/native `Field` path still paints the attached status over the control instead of behind it: with a pill radius, the lower 20px turns warning yellow, while `TextInput` stays above the same overlap (`Field.tsx:58`). Please keep custom controls visually above the status while preserving pointer access, and cover this state in the `Field` story.

[Reviewed by Robohands]

rubyycheung:
Addressed the latest review feedback:

- `Field` now places only its attached status in the wrapper's isolated background layer, so custom/native controls paint above the overlap while the existing `pointer-events: none` behavior preserves input access. Standalone `FieldStatus` behavior is unchanged.
- Added a native-input attached-status case to the `Field` Status Variants story and a regression test for the stacking relationship.
- Updated the RTL applicability registry for direction-neutral `FieldStatus`; the previously failing scope now evaluates as 2 verified N-A components with 0 coverage gaps.

Focused component tests, RTL coverage tests, strict lint, core build, Storybook production build, knowledge/sync checks, and the repository pre-commit suite pass locally.

cixzhang (approved):
Thanks, the custom-control path now keeps the control above the attached status and preserves focus across the overlap. Focused tests and exact-head visual checks pass.

[Reviewed by Robohands]

PR #5771, opened 2026-08-31
docs(Spinner): record canonical anatomy

cixzhang:
## Why

After #5766 keeps the stable `spinner` target on the status/ring element in labelled and unlabelled states, Spinner needs canonical anatomy that reflects that public ownership boundary without exposing structural implementation details.

## What

- Add canonical consumer anatomy for the Spinner, its Text-owned string label, and caller-owned custom label content.
- Add a v3 component record with the exact target/delegation/none map, labelled and unlabelled public root/ref behavior, and links to the component-theming and public-API records.
- Keep the labelled layout wrapper and the ring’s SVG internals out of public anatomy and target inventory.
- Make no runtime changes and add no Changeset.

## Dependency

- Stacked on #5766 and must not land before it.
- #5766 is a patch fix, so this documentation PR no longer waits for a minor release.

## Testing

- `pnpm exec vitest run packages/core/src/Spinner/Spinner.test.tsx packages/core/src/theme/themingTargets.test.ts scripts/check-knowledge.test.mjs`
- `pnpm check:repo`
- `node packages/cli/clients/cli/bin/astryx.mjs component Spinner --dense`

### DIFF
--- a/packages/core/src/Spinner/Spinner.doc.mjs
+++ b/packages/core/src/Spinner/Spinner.doc.mjs
@@ -1,12 +1,43 @@
 // Copyright (c) Meta Platforms, Inc. and affiliates.
 
+/** @type {import('@astryxdesign/cli/authoring').ComponentAnatomyElement[]} */
+const anatomy = [
+  {
+    name: 'Spinner',
+    required: true,
+    description:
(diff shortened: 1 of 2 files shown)


cixzhang:

Issue #5775, opened 2026-08-31
[Bug] AppShell with height="auto" scrolls beyond boundary when children have margin-top


ngolin:
### Description

When `<AppShell height="auto">` has children with `margin-top` applied, scrolling upward on the page causes content to scroll beyond the boundary, resulting in a view offset issue.

### Reproduction

```tsx
<AppShell height="auto">
  <div style={{ marginTop: 100, height: '100dvh', background: 'red' }}></div>
</AppShell>
```

Steps to reproduce:
1. Render the AppShell component with height="auto"
2. Apply margin-top: 100px to a child element
3. Scroll upward on the page
4. Observe that the content scrolls beyond the expected boundary

### Astryx Version

@astryxdesign/core@0.5.1

### Environment

Chrome 151, MacOS 26

ngolin:
This is fixed by PR #5776.

PR #5776, opened 2026-08-31
fix(core): add overflow clip to AppShell styles

ngolin:
## Description

Fixes two scrolling boundary issues in AppShell component by adding `overflow: clip` to the styles.

## Issues Fixed

1. #5775
2. #5815

## Problem

When `AppShell` components have children with `margin-top` or descendants with `position: absolute`, scrolling upward causes content to scroll beyond the expected boundary, resulting in a view offset issue.

## Solution

By adding `overflow: clip` to the AppShell styles, we prevent content from overflowing outside the component's boundaries, ensuring that:
1. Children with `margin-top` don't cause scroll overflow when `height="auto"`
2. Absolutely positioned descendants don't cause scroll overflow when `height="fill"`

### DIFF
--- a/packages/core/src/AppShell/AppShell.tsx
+++ b/packages/core/src/AppShell/AppShell.tsx
@@ -269,6 +269,7 @@ const styles = stylex.create({
     display: 'flex',
     flexDirection: 'column',
     position: 'relative',
+    overflow: 'clip',
   },
   variantWash: {
     backgroundColor: colorVars['--color-background-body'],


cixzhang (changes requested):
The margin-collapse fix works, but `overflow: hidden` makes non-scrollable `LayoutContent` a scroll container. In Chromium, a sticky descendant that stayed at `top: 0` on main scrolled away to `-184px` on this head.

Could we keep `overflow: clip` and add `display: flow-root` instead? That fixed the reported `height="auto"` gap while preserving sticky and focus behavior. The `rootFill` change does not affect the reported auto-height path and can be omitted.

Please add a browser regression test with `AppShell height="auto"`, `LayoutContent isScrollable={false}`, a margin-top child, and a sticky descendant. After page scroll, assert the gap is gone and the sticky element remains at the viewport top.

[Reviewed by Robohands]

ngolin:
> The margin-collapse fix works, but `overflow: hidden` makes non-scrollable `LayoutContent` a scroll container. In Chromium, a sticky descendant that stayed at `top: 0` on main scrolled away to `-184px` on this head.
> 
> Could we keep `overflow: clip` and add `display: flow-root` instead? That fixed the reported `height="auto"` gap while preserving sticky and focus behavior. The `rootFill` change does not affect the reported auto-height path and can be omitted.
> 
> Please add a browser regression test with `AppShell height="auto"`, `LayoutContent isScrollable={false}`, a margin-top child, and a sticky descendant. After page scroll, assert the gap is gone and the sticky element remains at the viewport top.
> 
> [Reviewed by Robohands]

Thanks for the reveiw ans suggestion! A couple of clarifications for what the fix changed this time:

1. The fix did not modify `LayoutContent` anymore. The fix only touches `AppShell` itself. Also, we did not use `overflow: hidden` anywhere; the implementation uses `overflow: clip` so sticky descendants may not be affected.

2. While investigating we found the same scroll/boundary issue can occur when AppShell with `height="fill"`. Because of that, `overflow: clip` cannot be limited to only `rootAuto` — it needs to apply to the root in both the `rootAuto` and `rootFill` paths. Therefore the unified change applys `overflow: clip` on the AppShell `root` so both `height: auto` and `height: fill` cases are covered.

cixzhang (changes requested):
Thanks for keeping this to `overflow: clip`. Chromium confirms it preserves page-owned sticky positioning and skip-link focus, while fixing fill-mode overflow. My earlier `flow-root` suggestion was unnecessary because the root is already a flex formatting context.

The browser regression test from the previous review is still missing. Existing tests do not cover margin-top + page scroll + sticky together, or fill-mode absolute overflow, so `clip` could regress to `hidden` without CI catching the sticky break.

[Reviewed by Robohands]

HelloOjasMutreja:
For the browser regression test cixzhang asked for: I independently worked through the same AppShell overflow issues (closed my own duplicate, #5842, in favor of this one) and confirmed via source reading that the sticky wrapper machinery lives entirely in auto mode's headerSticky/sideNavSticky styles, which key off position: sticky relative to the page scroll, not an ancestor scroll container - consistent with what cixzhang found. If it's useful, the shape of test that would pin this: render AppShell height="auto" with a sticky topNav, a margin-top child in content, and a tall body so the page actually scrolls; assert the header's bounding rect stays at viewport top after a real scroll event, and that a margin-collapse child no longer pushes content above the shell's own top edge. Happy to help verify further if useful, otherwise leaving this to you as the original PR.

cixzhang (approved):

PR #5777, opened 2026-08-31
docs(layout): clarify AppShell page-shell ownership

cixzhang:
## Why

Follow-up review on #5763 clarified the ownership boundary: AppShell is the page shell, while Layout is the general primitive for arranging `header`, `start`, `content`, `end`, and `footer` regions.

## What

- Move page-shell anatomy and guidance to AppShell and describe Layout as the reusable five-slot primitive.
- Align the layout-region and layout-primitive records with that boundary.
- Keep the broader container system under `architecture:container-padding`; Section, Table, Toolbar, and Divider retain their own component ownership, and Table/Divider do not become layout-region members.

## Risk

Documentation, specification, and source-comment changes only. There are no runtime, DOM, API, theme-target, or generated-output changes.

Known implementation gaps remain explicitly unchanged: automatic LayoutContent/LayoutPanel inset publication is not fully symmetric, and adjacent LayoutPanel/ResizeHandle divider ownership remains caller-coordinated.

## Testing

- `vitest run` for AppShell, Layout, Section, Table, Toolbar, and Divider (272 tests)
- `pnpm check:knowledge`
- `pnpm check:repo`

No Changeset: documentation/specification correction only.

### DIFF
--- a/docs/architecture/container-padding.md
+++ b/docs/architecture/container-padding.md
@@ -82,6 +82,11 @@ The protocol has four layers:
 | Alignment consumer  | Toolbar and edge-compensating child components         | Read the current inline inset to align visible content rather than stacked touch-target padding               |
 | Boundary owner      | Layer surfaces and dialog-based overlay roots          | Apply `overlayPaddingReset` before descendants read inherited page geometry                                   |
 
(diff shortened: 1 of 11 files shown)


humbertovirtudes (approved):
LGTM

PR #5778, opened 2026-08-31
docs: promote input-field family contract

cixzhang:
## Problem

The input-field family had accumulated strong evidence, but its purpose was too narrowly framed around sizing, its verification list looked exhaustive, and attached FieldStatus eligibility was unresolved.

## Change

- Promote the family with Cindy’s approved purpose: consistent input state display, behavior, appearance, and size.
- Keep membership and adoption exhaustive while treating `verified_by` and the verification matrix as representative anchors.
- Record attached status as a capability: it is the default where an opaque, bordered, fixed-height direct control reliably reflects resolved size; detached and tooltip remain available to every input.
- Keep direct FieldStatus limited to attached/detached, with Field consuming tooltip before rendering it.
- Preserve the verified Typeahead/Tokenizer sizing and end-control gaps without prescribing shared implementation machinery.
- Link AST-002 as the public-API owner without adding a separate invalid-state policy here.

## Risk

Documentation only. No runtime, public API, theme target, or Changeset change.

## Testing

- `pnpm check:knowledge`
- `pnpm exec vitest run packages/core/src/Field/Field.test.tsx packages/core/src/FieldStatus/FieldStatus.test.tsx packages/core/src/FormLayout/FormLayout.test.tsx packages/core/src/InputGroup/InputGroup.test.tsx packages/core/src/TextInput/TextInput.test.tsx packages/core/src/Selector/Selector.test.tsx packages/core/src/MultiSelector/MultiSelector.test.tsx packages/core/src/ComplexSelector/ComplexSelector.test.tsx packages/core/src/Typeahead/Typeahead.test.tsx packages/core/src/Tokenizer/Tokenizer.test.tsx`
- `pnpm check:repo`
- `pnpm -F @astryxdesign/docsite test`
- `git diff --check`

Current-record approval still applies to the exact final head. This PR remains draft.

### DIFF
--- a/docs/families/input-fields.md
+++ b/docs/families/input-fields.md
@@ -3,18 +3,21 @@ schema_version: 1
 template_version: 1
 kind: family
 id: family:input-fields
-authority: draft
+authority: current
 archive_reason: null
 superseded_by: null
-approved_by: null
-approved_at: null
+approved_by: cixzhang
+approved_at: 2026-08-31
 owners: [cixzhang, imdreamrunner]
 review_triggers: [behavior, layout, theming, accessibility]
 verified_by:
   [
-    packages/core/src/Field/,
-    packages/core/src/FormLayout/,
-    packages/core/src/InputGroup/,
+    packages/core/src/Field/Field.test.tsx,
+    packages/core/src/InputGroup/InputGroup.test.tsx,
+    packages/core/src/TextInput/TextInput.test.tsx,
+    packages/core/src/Selector/Selector.test.tsx,
+    packages/core/src/Typeahead/Typeahead.test.tsx,
+    packages/core/src/Tokenizer/Tokenizer.test.tsx,
   ]
 members:
   [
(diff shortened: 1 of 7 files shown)


docs/families/input-fields.md
```diff
@@ -12,9 +12,14 @@ owners: [cixzhang, imdreamrunner]
...
-    packages/core/src/FormLayout/,
-    packages/core/src/InputGroup/,
+    packages/core/src/Field/Field.test.tsx,
+    packages/core/src/FormLayout/FormLayout.test.tsx,
+    packages/core/src/InputGroup/InputGroup.test.tsx,
+    packages/core/src/TextInput/TextInput.test.tsx,
+    packages/core/src/Selector/Selector.test.tsx,
+    packages/core/src/MultiSelector/MultiSelector.test.tsx,
+    packages/core/src/Typeahead/Typeahead.test.tsx,
+    packages/core/src/Tokenizer/Tokenizer.test.tsx,
```
cixzhang, line 20: We have more form inputs than this but for the family test do we need to list them all for tests?

docs/families/input-fields.md
```diff
@@ -28,180 +33,298 @@ members:
...
 ## Intent
 
-A person should be able to move between Astryx form controls without fields
-changing width, row height, or interaction meaning because one control happens
-to have a value, status, spinner, or trailing action.
-
-## Membership rule
+A person should be able to move between Astryx form controls without the field
+changing inline size, row height, or interaction meaning merely because a value,
+status, busy indicator, or trailing action appears.
```
cixzhang: We should preserve the general purpose of inputs rather than discussing details about sizing.

Inputs should have consistent treatment and API contract across the system for state display, behavior, appearance and size.

docs/families/input-fields.md
```diff
@@ -28,180 +33,298 @@ members:
...
 
-- **OQ1 — Resolved.** The membership rule is correct. Selector remains a
-  member with the DEC-1 width-stability exception. (`human-api`)
-  None. Membership adoption still needs verification before this draft becomes
-  `current`.
+- **OQ1 — When may attached FieldStatus bridge a control's lower corners?**
+  (`human-design`) **Recommended:** allow the overlap only for an opaque,
+  bordered, fixed-height direct control whose owning root reliably reflects its
+  resolved size. Wrapped, custom, tall, or translucent surfaces keep
+  component-owned placement, normally `detached`. This lets people read a
```
cixzhang: This seems true. Whenever we can attach we apply attached by default, else detached. Status positioning remains a viewer option between detached, and tooltip for all inputs and attached when the input supports it.

cixzhang: As much as possible, if it doesn't conflict with understanding APIs should try to prevent bad results. This is perhaps a separate spec amendment.

docs/families/input-fields.md
cixzhang: Updated: `verified_by` now names representative evidence anchors instead of looking exhaustive. The membership list and adoption table remain exhaustive, and the prose says every member still requires focused evidence.

docs/families/input-fields.md
cixzhang: Updated the Intent and shared-owner boundary around consistent input state display, behavior, appearance, and size. Sizing and end-control geometry now read as invariants within that broader purpose.

docs/families/input-fields.md
cixzhang: Recorded as DEC-4 and promoted the family: attached is the default wherever a member safely supports it; detached and tooltip remain available to every input. Direct FieldStatus remains attached/detached, and Field consumes tooltip before rendering it.

cixzhang: Kept that policy out of this family. The record now links AST-002 as the public-API admission owner and explicitly adds no separate invalid-state or API-guardrail policy here.

cixzhang (commented):
Thanks, this is good. This is an approve once you undraft.

[Reviewed by Robohands]

PR #5782, opened 2026-08-31
docs: project parent-owned member contracts

cixzhang:
## Why

Extracted member pages need their exact parent-owned anatomy and theming target without creating a second writable declaration or inheriting the parent's full aggregate contract.

## What

- Adds an explicit, fail-closed parent-owned projection resolved by both direct CLI lookup and docsite generation.
- Migrates the four proven members and removes their duplicate target declarations while keeping public target discovery unchanged.
- Keeps name-only refs out of knowledge contracts and rejects member theming-anatomy maps owned by the parent.

## Risk

Documentation and tooling only. No React API, runtime DOM, CSS, target name, or import-path changes.

## Testing

- Focused projection, loader, CLI, target, knowledge, and docsite suites
- `pnpm check:repo`
- Generated docsite registry inspection for projected members and aggregate parents

### DIFF
--- a/.changeset/deduplicate-parent-theme-targets.md
+++ b/.changeset/deduplicate-parent-theme-targets.md
@@ -2,6 +2,6 @@
 '@astryxdesign/cli': patch
 ---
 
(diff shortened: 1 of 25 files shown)


cixzhang:

PR #5783, opened 2026-08-31
fix(useResizable): add structured percentage sizing

freddymeta:
`useResizable` now resolves percentage configuration against an explicit basis and keeps state, persistence, paint, callbacks, and ARIA in pixels.

## Structured sizing

```ts
import {percent, pixel} from '@astryxdesign/core/Resizable/utils';

pixel(333);
percent(40, {min: pixel(333)});
percent(10, {max: pixel(400)});

useResizable({
  defaultSize: percent(40, {min: pixel(333)}),
  containerRef,
});
```

- `pixel()` and `PixelWidth` are the existing Table binding and type; Resizable does not define another fixed-size helper.
- `percent()` is Resizable-owned because it is a literal percentage of one basis, unlike Table's sibling-weight `proportional()`.
- Its options argument is required and accepts exactly one pixel `min` or `max`. Exact `'40%'` remains the only unbounded percentage spelling.
- Numbers, exact `Npx`, exact `N%`, and the released `minSizePx` / `maxSizePx` aliases remain compatible.
- Recursive CSS `min()` / `max()` parsing is deliberately unsupported.

## Semantics

- `containerRef` selects the content-box basis on the active axis. Without it, the released viewport basis and 1200px SSR fallback remain.
- A percentage `defaultSize` resolves once to pixels. A structured default applies its floor or ceiling once and then releases its observer.
- Percentage `minSize` / `maxSize` remain live and re-clamp the selected pixel state when their basis changes.
- Persisted pixels take precedence over an unused default. Temporary SSR or zero-size measurements are neither persisted nor reported as interactions.
- Invalid descriptors warn in development and take the same role-specific fallback in every build. Both/neither structured bounds, invalid percentages, and invalid nested pixels are rejected.

The original defect was a CSS-only percentage ceiling: paint stopped at the CSS cap while state and `aria-valuenow` continued past it. The bound now clamps the pixel state itself, so panel width, storage, callbacks, and ARIA agree.

## Browser evidence

Structured percentage sizing

Chromium 149 against the built Storybook verified:

- `percent(40, {min: pixel(333)})` resolves to 399px on a 998px basis and 333px on a 498px basis;
- a default remains at its selected pixel size after the basis changes;
- a live floor follows 333px → 399px and a live ceiling follows 100px → 50px;
- panel width, persisted size, and `aria-valuenow` match throughout;
- no page or console errors.

## Verification

- full workspace build;
- strict lint: 0 errors, 85 existing repository warnings;
- core source suite: 291 files, 8,779 tests passed;
- core, docs, and Storybook typechecks;
- Storybook production build;
- changeset and knowledge checks;
- Chromium negative control failed as intended, then the real story passed;
- production parent-versus-candidate benchmark: identical 120-step resize renders/callbacks and 1999.9ms median; at 50 mounts, structured and atomic percentage defaults were 65.4ms and 65.3ms;
- atomic parse/resolve median: 77.45ns parent vs 77.96ns candidate (1.007×); structured descriptor parse/resolve: 17.30ns.

SideNav's simplified pixel-only width configuration is unchanged.


cixzhang (changes requested):
This changes released sizing and leaves several builders with geometry AST-010 does not permit.

- Without a ref, `50%` paints 250px in a 1000px viewport. With the returned ref, the default tracks 374→254 on container shrink, and a 382px clamp revives to 562px on growth. AST-010 FR1/FR4/FR6 require one-time pixel selection and persistent clamping. (`useResizable.ts:367`)
- Vertical and bordered containers measure border-box width. Chromium saw a 934px content box but `aria-valuemax=562`, not 560. AST-010 FR2/API1/API2 require a configured ref and content-box active axis. (`useResizable.ts:330`)
- Unified bounds lose to deprecated aliases, duplicate pairs type-check, `resize(NaN)` becomes `NaN`, and callback-less cleanup can remove peer observers. AST-010 FR11/FR12/API4/API6 require unified precedence, exact validation, and independent cleanup. (`useResizable.ts:353`; `sharedResizeObserver.ts:102`)

Please add no-ref, axis/content-box, re-clamp, invalid/alias, multi-region, SSR/hydration, and cleanup coverage.

[Reviewed by Robohands]

freddymeta:

Reworked onto AST-010. All three findings reproduced and fixed, measured rather than argued.

**1 · No-ref default, one-time selection, reviving clamps (FR1/FR4/FR6)** — all three real.
- `50%` with no ref paid 250px because I'd deleted the viewport resolver. Restored: no ref keeps the released one-time `window.innerWidth`, 1200px when unavailable.
- The default tracked the container forever; it now resolves **once** into a pixel selection, and only bounds re-resolve.
- The clamp revived on growth — I reproduced your 382→562 as **320→560**. I'd derived the clamp for paint but left the stored choice at its pre-clamp value. The clamp is now committed to state, so it's permanent.

**2 · Border-box and axis (FR2/API1/API2)** — confirmed exactly as you measured: a 934px content box reporting `aria-valuemax=562` (60% of the 936px *border* box). Now content-box on the active axis, `direction` selecting inline or block size, taken from the observer's own `contentBoxSize`. Chromium reports **560** for the same box. `containerRef` is a caller-owned `RefObject` per API1, not the callback ref I'd returned.

**3 · Precedence, validation, cleanup (FR11/FR12/API4/API6)**
- Unified bounds now **win** over the deprecated aliases (they lost before); development names the ignored alias, and each pair is the exact mutually exclusive union from API4, so duplicates no longer type-check.
- Parsing is exact and complete-match: `'50 px'`, `'50rem'`, `'120%'`, `'-10px'`, `NaN` all reject to the role fallback, identically in both builds, warning only in development. Explicit `maxSizePx: Infinity` stays valid.
- `resize(NaN)`, `resize(Infinity)` and negatives warn and keep the last legal size.

**Coverage:** no-ref, axis/content-box, re-clamp, invalid/alias, multi-region, SSR/hydration, cleanup. 79 Resizable tests; **17 of the new ones fail against `main`**, and the ones that pass there are the compatibility cases — which is the point.

Peer-safe cleanup is split out as **#5817**, because §Implementation-requirements 9 asks for it to stay independently reviewable; this PR stacks on it, and FR11 needs it since multi-region puts every region on one container.

cixzhang (changes requested):
Thanks — no-ref sizing, ordinary content-box measurement, validation, alias precedence, and peer-safe observer cleanup now reconcile.

Builders using percentage bounds can still get panels that revive or retain stale geometry. A default clamps 454→240 as its container shrinks, then revives to 454 on growth; in vertical writing modes one config uses a 160 default with 90/360 bounds. Cancellation exposes the same split: pointerup releases a frozen basis and clamps 600→300, but `pointercancel` leaves 600, and iframe crossing leaves the drag active. Replacing `containerRef.current` can set max to 0 because observation remains on the old element.

Please contract this to one authoritative pixel selection, one basis tied to the current element/axis, and one terminal gesture path. `_direction` also became required on exported `ResizableProps`; a valid pre-PR object no longer compiles. Could that public type stay source-compatible?

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks — the earlier no-ref, content-box, validation, and alias issues now reconcile, but builders can still lose panel geometry in three current-head paths.

- Replacing the element behind the same `containerRef` leaves the detached element observed. Chromium went from max 199 in a 400px box to size/max 0 in its 800px replacement, not 399; a hidden mount also overwrote a saved 321px width with 0. (`useResizable.ts:464`, `:697`)
- `pointercancel` cleans listeners without `_onResizeEnd()`, so the frozen basis never releases. After shrinking 998→598px during a cancelled drag, size/ARIA max stayed 499 instead of 299. (`ResizeHandle.tsx:428`)
- `ResizableProps` is exported; requiring `_direction` makes every previously valid object literal fail to compile. Please keep the mismatch diagnostic source-compatible. (`useResizable.ts:163`)

Please rebase too: current main rewrites the same pointer lifecycle.

[Reviewed by Robohands]

freddymeta:
Rebased as a merge of `origin/main` — two files conflicted, both from #5297. I took your pointer-capture rewrite whole and re-applied only this branch's axis-mismatch warning beside it; resolution is in the description. **#5297 already closes the iframe-crossing case from your second review, so this branch carries no version of it.**

All three reproduced in Chromium 149 first, then fixed:

| finding | before | after |
|---|---|---|
| element replaced behind the same `containerRef` (398 → 798 content box) | max 199 → **0** | max 199 → **399** |
| hidden mount over a saved 321px width | storage `{"size":0}` | storage `{"size":321}` |
| container 998 → 598 during a **cancelled** drag | size / `aria-valuemax` **499** | **299** |
| pre-PR-shaped `ResizableProps` literal, `tsc` | **TS2741** | exit 0 |

- The subscription was keyed on the ref *object*, which survives an element swap. The observed element is now state, set from an effect comparing `containerRef.current` after each commit — refs are assigned during commit, so that is the only moment a swap is visible. The node left behind is unobserved.
- A zero measurement is no longer a measurement: hidden or detached holds the documented temporary 1200px basis instead of resolving every bound to 0, and nothing is persisted while the basis is temporary.
- Cancel does **not** call `_onResizeEnd` — #5297 deliberately signals none on cancel and its test asserts that. It reports a new **optional** `_onResizeCancel`, which releases only the frozen basis. A handle unmounted mid-drag reports it too.
- `_direction` and `_onResizeCancel` are both optional. The required `_direction` had already broken `makeResizable()` in your own #5297 test file — that was the merge's first typecheck error.

Negative control per fix: reverted on its own, its focused tests fail and the Chromium numbers go back to 600 / 0 / 499.

Gates: 9003 core + lab tests (316 files); core build, `typecheck`, `typecheck:docs`, `check:repo` green; `lint:strict` 85 warnings / 0 errors, matching `origin/main` exactly.

One correction to my own earlier claim: I wrote that `@eslint-react/set-state-in-effect` "rejects" this shape. It does not for DOM identity, and the first fix now suppresses that rule with a reason. The *size* is still derived, never stored in an effect.

Stack: #5817 is this branch's first commit, not duplicated work — land it first and this diff shrinks to the `Resizable` files.

freddymeta:
Restacked onto corrected #5817 head `490cbfa48a` and squashed to one #5783 commit (`6ab133cdf8`). This layer now touches only Resizable/docs/story files — no observer implementation or migrated observer consumer.

All three current-head findings reconcile:

- **ref replacement / hidden mount:** the subscription follows the element behind the stable ref, unobserves the old node, and treats zero as unmeasured. Saved 321px geometry is not replaced by 0.
- **cancelled gesture:** real browser `touchCancel` releases the frozen basis and body drag styles without claiming a completed resize.
- **source compatibility:** `_direction` and `_onResizeCancel` are optional; an object literal written against the released `ResizableProps` still type-checks. The mismatch diagnostic remains active when `_direction` is present.

Chromium also found one path the earlier tests missed: a raw 321px default initially clamped to 200px revived to 321px when its container grew. The initialized clamp is now committed as ordinary pixel state, so it cannot revive.

```
replacement: basis 400 -> 800, max 200 -> 400, size 200 -> 200
hidden:      saved size=321, storage unchanged
touchCancel: max 500 -> frozen 500 -> 300, size=300, body styles clear
```

The new default-revival test fails on the prior PR head (`321`, expected `200`) and passes here. The review regression set fails on the pre-fix implementation in the same ways reported: replacement max wrong, old node retained, hidden size/storage become 0, cancel callback absent, and old object literals fail TypeScript.

Validation: **288 files / 8674 core tests**, `pnpm check:repo`, `pnpm build`, core TypeScript, and Resizable lint all pass. Ready for exact-head re-review.

freddymeta:
Final restack update: #5817 moved after my previous comment because main advanced. #5783 is now based on the final corrected #5817 head **`7797ca59d2`**, not stale `490cbfa48a`; its exact head is **`0930681fd5`**.

The final base adds current main plus shared-observer RTL coverage. There was no overlap with #5783, and this layer still touches only Resizable/docs/story files — no observer implementation or migrated observer consumer.

Revalidated after the move:

- focused Resizable + shared observer: **3 files / 109 tests**
- full core: **288 files / 8680 tests**
- core TypeScript, Resizable/observer lint, `pnpm check:repo`, and `pnpm build`: clean
- final base contains current fetched main; #5783 leaves observer files untouched

The Chromium evidence and default-revival fix from the preceding comment are unchanged.

cixzhang (changes requested):
Thanks — the three prior geometry blocks now reconcile in Chromium and source-compatibility tests. One current-head blocker remains: `pr-rtl` fails because `core/Layout` and `lab/Resizable` have no measured or verified-N/A coverage. This changes axis-sensitive resize behavior, so please add a measured RTL target for Resizable and account for Layout before merge.

[Reviewed by Robohands]

freddymeta:
Prerequisite #5817 has landed on main as `d55751f88f`. #5783 is now rebased directly onto current main at exact head `54018f5ae3`; it no longer carries any prerequisite branch commits.

Revalidated after the move: focused Resizable + observer **109 tests**, full core **288 files / 8680 tests**, `check:repo`, build, TypeScript, lint, and diff checks all pass. The browser evidence and default-revival fix are unchanged.

cixzhang (changes requested):
Thanks — we previously asked for geometry, source compatibility, and RTL accounting. Geometry now reconciles in Chromium, and source compatibility passes typecheck. Two exact-head blockers remain.

- `pr-rtl` still fails: `core/layout` and `lab/resizable` are both coverage gaps with 0 measured and 0 verified-N/A results, so this axis-sensitive change still has no checked-in RTL accounting.
- Every existing pixel-only `useResizable` mount now takes two render passes. The same probe measured hook work 1→2 at N=1, 10→20 at N=10, and 50→100 at N=50. `chosenSize` starts null and is set during render even when no container measurement is needed (`useResizable.ts:644-659`). Please preserve the one-pass legacy path while retaining the measured-container correction.

[Reviewed by Robohands]

freddymeta:
Both blockers are fixed at a9984e3.

**Perf regression — reproduced, then fixed.** Render-pass probe over a pixel-only mount (`defaultSize: 200`, `minSizePx: 100`, `maxSizePx: 400`), counting hook invocations:

| | N=1 | N=10 | N=50 |
|---|---|---|---|
| `origin/main` (pre-PR) | 1 | 10 | 50 |
| PR head 54018f5 | 2 | 20 | 100 |
| after fix | 1 | 10 | 50 |

Your 1→2 / 10→20 / 50→100 reproduced exactly, and the cause is the one you named: `chosenSize` started `null`, so the render-phase adjustment ran at mount even when `isBasisMeasured` was already true. The selection is now initialized outright whenever the basis is real on the first render — every pixel-only config, and the no-container percentage path. Only a supplied container still starts `null`, because its measurement does not exist until after commit; that correction is unchanged.

Negative control: reverting only that hunk put the probe back to 2/20/100; restoring it returned 1/10/50. Pinned as `settles a pixel-only mount in one render pass`.

Earlier rounds' evidence re-run green (69 + 30 tests): no-ref `'50%'` → 500 on a 1000px viewport, ref-following, zero-is-not-a-measurement, and `_onResizeCancel`.

**RTL** — both gaps measured, neither declared N/A:

- `lab/resizable` — D4 on `lab-resizable--horizontal-split`: `D4 overlay side frac: LTR 0.23 RTL 0.77 flipped=true`.
- `core/layout` — Layout genuinely flips its start/end panels, so verified-N/A would have been false. D2 on `core-layout--dual-panels`: `D2 prev/next cx: LTR 92<737=true; RTL 1008>363=true`.

`pr-rtl` scoped to `Layout,Resizable` now exits 0: `COV : 2 measured / 0 verified N-A / 0 gap / 0 stale`.

Gates: core build, `vitest packages/core/src packages/lab/src` (9038 passed), typecheck, typecheck:docs, check:repo all pass; `lint:strict` 0 errors / 85 warnings, matching main. `origin/main` merged (its new Typeahead/Tokenizer/Toast targets auto-merged cleanly).

cixzhang (approved):
Thanks — the pixel-only path is back to one render, and exact-head RTL now measures Layout and Resizable with no coverage gaps. This is good to merge.

[Reviewed by Robohands]

freddymeta:
Rebased the approved PR onto current main (`dc5565edce`) and resolved the only conflict by preserving main’s Stepper RTL target alongside this PR’s approved Layout and Resizable targets. New exact head: `7e446527ae`.

- focused Resizable + observer: **3 files / 110 tests**
- full core: **290 files / 8739 tests**
- core TypeScript and Resizable/observer lint: clean
- `pnpm build`, `pnpm check:repo`, Storybook build: pass
- exact RTL scope: Layout D2 pass, Resizable D4 pass; **2 measured / 0 gaps**
- diff check clean; no unrelated target-registry formatting churn

Authorized to land once the rebased head is green.

freddymeta:
Hash correction: the final formatting-clean amend changed the pushed exact head to **`5175499ad6`**. The validated content and test results in the preceding comment are unchanged.

cixzhang (changes requested):
Thanks — the rebase preserves the approved percentage/container behavior, but the current slice still suppresses one required migration warning. When untyped input supplies both old/new pairs, `minSizePx` flips the shared warning latch, so the ignored `maxSizePx` is silent. Each ignored alias needs its own development warning and a combined-conflict test in this PR.

[Reviewed by Robohands]

freddymeta:
Fixed the exact-head migration-warning bug. The shared alias-warning latch is now two independent latches, so untyped input supplying both old/new pairs emits both required warnings: one naming ignored `minSizePx`, one naming ignored `maxSizePx`.

Added one combined-conflict regression that asserts both unified values win and both full warning messages fire exactly once. Negative control against the prior head fails exactly as reported: expected 2 warnings, got 1.

Validation: Resizable + observer **111 tests**, full core **290 files / 8740 tests**, TypeScript, lint, `check:repo`, build, and diff checks all pass. New exact head is in the PR status above.

cixzhang (changes requested):
Thanks — the prior alias-warning ask now reconciles: both warnings fire, and the focused Resizable tests pass. Two current-head paths still need tightening.

- A numeric-only region with `containerRef` doubles hook work: 1→2, 10→20, and 50→100 calls. The ref-following effect records the element even though no percentage can consume it, contradicting the one-pass release note. Gate that work on percentage use and add this combination to the regression test.
- Six shipped vertical examples/templates omit the hook’s new `direction`, so they now emit the mismatch warning. Please make those pairings satisfy the new contract.

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks — the structured `percent()` API matches AST-010, but the two prior asks remain open.

- A numeric-only region with `containerRef` still doubles hook work: 1→2, 10→20, and 50→100 calls. The ref-following effect records DOM identity even when no percentage can consume it. Please gate that work on percentage use and add this combination to the one-pass regression. (`useResizable.ts:624`)
- Six vertical examples/templates still pair the hook’s default horizontal direction with a vertical handle. Chromium emits the mismatch warning in all three affected hook stories; the other three have the same pairing. Please pass `direction: 'vertical'` to every vertical hook. (`useResizable.stories.tsx:231`; `Resizable.stories.tsx:97`; `ide/page.tsx:219`)

[Reviewed by Robohands]

freddymeta:
@cixzhang I rechecked both asks against exact head `0bcbd04fcd53ae252bac160afe4509bca7d73e47`; they are still outstanding, so I am not marking this review resolved.

- Numeric-only `useResizable({defaultSize: 333, containerRef})` still invokes the hook **2 / 20 / 100** times at N=1 / 10 / 50, versus **1 / 10 / 50** without the ref. `observeContainerBasis` is false, but the ref-following effect still records `containerRef.current` unconditionally (`packages/core/src/Resizable/useResizable.ts:628-663`). The current one-pass test omits `containerRef` (`useResizable.test.ts:427-440`).
- Six hook/handle pairings still omit hook `direction: 'vertical'`: `apps/storybook/stories/useResizable.stories.tsx` at 231-249, 316-359, and 417-454; `apps/storybook/stories/Resizable.stories.tsx` at 97-120 and 203-250; and `packages/cli/assets/templates/pages/ide/page.tsx` at 219-225 / 295-301. The hook exports `_direction: 'horizontal'` (`useResizable.ts:1081-1094`), and the exact mismatch-warning test (`ResizeHandle.test.tsx:165-181`) passes: **1 passed, 29 skipped**.

No code was changed for this verification. The normal final-head review-request workflow already succeeded and GitHub currently lists you as requested: https://github.com/facebook/astryx/actions/runs/33728610697

freddymeta:
Fixed the two review blockers at `e00d0f8bc1f42558ea769fe152f92e36818591b2`.

- Numeric-only `containerRef` parity: N=1/10/50 now yields exactly 1/10/50 hook renders, 0 ref reads, and 0 `ResizeObserver` allocations (old head: 2/20/100 renders and 2/20/100 reads). Pixel→percentage transitions and live ref replacement remain covered; a `flushSync` regression proves the corrected size is committed before passive effects, and fails a passive-effect mutant at the 1200px fallback.
- Vertical direction: all six shipped hook/handle pairings now pass `direction: 'vertical'`. The source guard validates 34 shipped pairings plus a real embedded-doc pairing; its negative control removes the six hook directions and requires six failures.
- Validation: full suite 644 files / 14,163 tests; core 292 / 8,788; CLI 212 / 3,119; build, typechecks, strict lint (0 errors), Storybook build, repo gates, and RTL audit pass. Chromium 149 exercised all five vertical Storybook surfaces in LTR+RTL and structured sizing/persistence with zero console or page errors. Parent/current-main/fixed performance and allocation thresholds pass.

cixzhang (approved):
Thanks, the prior performance and direction asks now reconcile. Structured sizing matches AST-010; numeric-only mounts stay single-pass, and all six vertical pairings are aligned and guarded.

[Reviewed by Robohands]

PR #5792, opened 2026-08-31
docs(knowledge): add package-local theme records

cixzhang:
## Why

Theme maintainers need a canonical record beside each theme package, with malformed placements and unauthorized approvals failing closed.

## What

- Adds immutable schema v2 as an additive overlay introducing only the `theme` kind. Schemas compose per kind; every existing kind, record, and unchanged template remains on v1.
- Discovers canonical records at `packages/themes/<theme>/<theme>.spec.md`. Any non-README Markdown under `docs/themes/` and any other direct `packages/themes/<theme>/*.spec.md` filename is treated as an unsafe knowledge candidate and fails with a placement error; it cannot take the spec-only auto-merge path.
- Uses one typed `references` list. Current-theme approval is exactly the committed `ENGOWNERS` + `DESIGNOWNERS` union—never schema v1 `approvalOwners`, location, or record metadata.
- Places cross-theme human contrast methodology in a future current design record, shared measurement implementation in architecture/tooling, and only theme-specific mappings, pairings/states, exceptions, measured receipts, and known gaps in each theme record.

## Risk

Knowledge-schema, validation, and review-workflow structure only. No runtime, token API, palette API, visual output, artifact, or consumer behavior changes. No existing record or unchanged template is migrated.

## Validation

- `npx --yes pnpm@11.10.0 check:knowledge -- --base origin/main`
- 131 focused schema, discovery, wrong-placement, change-scope, typed-reference, and exact-owner tests
- `npx --yes pnpm@11.10.0 check:repo`
- v1 remains byte-for-byte unchanged; no Changeset

## Review questions

1. Is `packages/themes/<theme>/<theme>.spec.md` the right canonical package-local convention and are wrong candidates rejected clearly enough?
2. Does the template keep shared contrast methodology/tooling out of per-theme records while retaining actionable receipts and gaps?
3. Is the exact `ENGOWNERS` + `DESIGNOWNERS` approval union the correct repository boundary?

### DIFF
--- a/.github/scripts/change-scope.cjs
+++ b/.github/scripts/change-scope.cjs
@@ -7,20 +7,28 @@
  * Classifies changed paths without reading PR-controlled content.
  *
  * A spec-only PR may change only spec records. Templates, schemas, indexes,
- * architecture, audits, workflows, and code deliberately do not qualify.
+ * architecture, guidance, audits, workflows, and code deliberately do not
+ * qualify.
  */
 
 const SPEC_RECORD_PATTERNS = [
   /^docs\/specs\/[^/]+\/(?:spec|plan)\.md$/,
   /^docs\/families\/(?!README\.md$)[^/]+\.md$/,
   /^docs\/design\/(?!README\.md$)(?!assets\/)[^/]+\.md$/,
+  /^packages\/themes\/([^/]+)\/\1\.spec\.md$/,
   /^packages\/(?:core|lab)\/src\/[^/]+\/[^/]+\.spec\.md$/,
 ];
 
 const CHANGESET_PATTERN = /^\.changeset\/(?!README\.md$)[^/]+\.md$/;
 
+const THEME_DOC_CANDIDATE = /^docs\/themes\/(?!README\.md$)[^/]+\.md$/;
+const THEME_PACKAGE_CANDIDATE =
+  /^packages\/themes\/[^/]+\/(?:.*\/)?[^/]+\.spec\.md$/;
+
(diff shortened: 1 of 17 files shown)


docs/specs/AST-006/spec.md
```diff
@@ -0,0 +1,631 @@
...
+
+This draft separates three layers that must not collapse into one API:
+
+1. **Portable core and domain tokens** are Astryx's closed, cross-theme public
+   vocabulary. Components and non-CSS consumers may rely on these names across
+   every conforming theme.
+2. **Theme-local role tokens** are reusable roles owned by one theme or one
+   flattened `extends` family. They allow that theme's component overrides to
+   share a value such as a filled status color without claiming that every theme
+   or app supports the name.
```
cixzhang: How would we prevent collisions between this and our astryx tokens?

cixzhang: Good question — the current head now makes collision prevention part of the contract rather than relying on author discipline. New local roles receive generated `--astryx-theme-<theme>-<role>` names; authors cannot supply preformed names or lineage metadata. Runtime `defineTheme` and static build must both reject collisions with portable core/domain tokens, reviewed public component variables, the reserved `--astryx-*` space outside this owned namespace, and private `--_*` names before emitting partial output. If Astryx later admits a portable/public name that collides with a released local role, that requires an explicit migration/deprecation rather than silently changing precedence. Typed `defineLocalTokens(...).refs` is the recommended reference path, and raw names inside the reserved namespace are accepted only when they resolve to the verified theme lineage.

PR #5794, opened 2026-08-31
fix(ci): reconcile fork review events safely

cixzhang:
## Why

Fork `pull_request_review` runs receive a read-only `GITHUB_TOKEN`, so the spec-owner reconciler fails when it writes commit statuses.

## What

Route review events through a no-permission anchor and a trusted `workflow_run` companion. The companion resolves exactly one open PR by head SHA, branch, and base repository before invoking the existing reconciler; ambiguous or stale identities fail closed.

## Risk

No product behavior changes. Privileged code still comes only from the default branch, and approval remains derived from the existing owner rules.

## Testing

- 35 focused spec-owner workflow tests
- `actionlint` on both changed workflows

### DIFF
--- a/.github/scripts/spec-owner-reconcile.cjs
+++ b/.github/scripts/spec-owner-reconcile.cjs
@@ -49,6 +49,7 @@ async function reconcileSpecOwnerGate({
   github,
   context,
   core,
+  pullNumber: requestedPullNumber,
   workspace = process.env.GITHUB_WORKSPACE,
   env = process.env,
 }) {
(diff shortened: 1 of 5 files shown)


cixzhang (commented):

Thanks, this looks good. This is an approve once you undraft.

[Reviewed by Robohands]

cixzhang:
Superseded by #5799. Independent comparison found that this workflow-run relay drops the original review-event semantics, so review dismissal does not take the immediate revocation path and can leave an owned auto-merge active until delayed reconciliation. Its relay guards also lacked mutation-sensitive Boolean-composition coverage. #5799 keeps same-repository review reconciliation direct and uses the existing trusted exact-head `/approve-spec` command for fork current-spec PRs.

PR #5795, opened 2026-08-31
fix(Selector): add interruptible open and close motion

mrabdussalam:
## Summary

- replace Selector’s one-way entry keyframe with interruptible CSS transitions for pointer-opened popovers
- add a symmetric exit path for trigger, selection, and native light-dismiss closes
- keep keyboard disclosure instant and use a fade-only reduced-motion treatment
- align the chevron’s timing with the surface and document the behavior contract

## Verification

- `pnpm -F @astryxdesign/core lint` (0 errors; existing warnings only)
- `pnpm vitest run packages/core/src/Selector/Selector.test.tsx packages/core/src/Selector/Selector.source-build.test.mjs` (169 passed)
- `pnpm check:repo`
- Chromium runtime matrix: pointer open, trigger close, selection close, outside close, keyboard open/Escape/Tab, and reduced motion
- frame-by-frame comparison against published `@astryxdesign/core@0.5.2`

### DIFF
--- a/.changeset/selector-interruptible-motion.md
+++ b/.changeset/selector-interruptible-motion.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Give pointer-opened Selector popovers interruptible open and close transitions while keeping keyboard disclosure instant and reduced motion gentle.
+
+@mrabdussalam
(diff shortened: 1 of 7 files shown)


mrabdussalam:
Closing at the requester’s direction. The original Selector motion should remain unchanged.

PR #5798, opened 2026-08-31
feat(templates): add the dialog wizard page template

ernestt:
Adds `form-wizard-dialog`, the fourth and last of the wizard templates: a three-step flow for scheduling a recurring report, run inside a Dialog.

**Rubric score: 99/100 (grade A).** Pass bar is B (75).

## What this one is for

The container is the whole differentiator. A page wizard gets the full content width; a vertical rail can afford a description under every step. A modal gets 620px and a body that scrolls against `maxHeight`, so the step markers go compact, the action row pins to the bottom, and the budget is two to four short steps. Past that, the flow wants a page.

The title and the stepper are two pinned header rows rather than body content, because a stepper placed in the body slides out of view on the longest step — the one where knowing your position matters most.

| | |
|---|---|
|  |  |
| Step 1 — report definition | Step 2 — cadence, as a radiogroup |

## Inline, and why

The dialog renders with `isInline` and stays open. A real modal is promoted to the browser's top layer, where no ancestor `transform` or `overflow: hidden` can clip it — so an open one inside a scaled gallery tile paints over the whole gallery instead of inside its tile. The docsite's own template preview is also a Dialog, and Astryx forbids nesting dialogs.

`isInline` draws the same anatomy without the `<dialog>` element, the backdrop, or the modal behaviour. The module header spells out the four-line change back to a real modal, and what belongs on the page behind it. This follows the precedent already set by `settings-dialog`.

## A bug found on the way

`Layout` inside the dialog was `height="auto"`. An auto Layout sizes to its content and lets the dialog clip the overflow rather than scroll it, which on the Delivery step hid 14px of the summary outright and cut into the footer buttons. `height="fill"` puts the scroll where the module header always claimed it was: the content region scrolls 117px on that step, the action row stays put, nothing is unreachable.

| | |
|---|---|
|  |  |
| Step 3 — footer pinned, body scrolls | …scrolled to the summary at its foot |

Also fixed while scoring: the cadence `Field` paired `isGroupLabel` with `inputID`, but `FieldLabel` drops `htmlFor` in group mode, so the id was inert and the radiogroup was falling back to a duplicate `aria-label`. Now wired the documented way — `labelID` + `aria-labelledby` — so the group takes the label the user can actually see.

## Validation and confirmation

Blocking a step reports the count beside the button that is refusing to advance. Completing the flow raises a toast: a real modal would close, and closing is most of the confirmation, so with nothing to close the toast carries it and the draft resets.

| | |
|---|---|
|  |  |
| Blocked, with the field marked | Scheduled |

## Light, dark, narrow

| | |
|---|---|
|  |  |
| Dark | 420px — no overflow |

## Test plan

- [x] Drives end to end in the template viewer: three steps, blocked state, completion toast, reset
- [x] Radiogroup resolves its accessible name from the visible label via `aria-labelledby`
- [x] Delivery step scrolls its body; primary action fully visible on every step; nothing clipped
- [x] 420px viewport: no horizontal overflow
- [x] Light and dark
- [x] `typecheck:template-docs`, eslint and prettier clean


cixzhang (approved):

packages/cli/assets/templates/pages/form-wizard-dialog/page.tsx
```diff
@@ -0,0 +1,570 @@
...
+    showToast({
+      body: `${name.trim()} is scheduled · ${cadenceSummary}`,
+      uniqueID: 'report-scheduled',
+    });
+    setStep(0);
+    setAttempted(new Set());
+    setName('');
+    setSections(['summary', 'charts']);
+    setNote('');
+    setRecipients(['exec']);
```
cixzhang: Some schedule and delivery fields survive reset. Might reset those too.

cixzhang (approved):
Thanks, this looks good. Template rubric: A (100/100). One small reset gap before merge.

[Reviewed by Robohands]

packages/cli/assets/templates/pages/form-wizard-dialog/page.tsx
ernestt: Reset now restores every report, schedule, and delivery field to its initial value.

PR #5805, opened 2026-09-01
feat(core): add read-only selectors

cixzhang:
## Why

A selected value can be locked by permissions or policy while still needing to stay visible and participate in form submission. Mapping that state to disabled would dim the value, remove it from the tab order, and exclude it from submitted data.

## What

- Defines the shared read-only selection-input contract as system spec AST-011 before implementation, grounded in WCAG 2.2 SC 4.1.2 and WAI-ARIA 1.2.
- Adds `isReadOnly` to Selector and MultiSelector: values retain their `combobox` role, stay focusable and form-submittable, and expose `aria-readonly`, while popovers/sheets, clear actions, disclosure indicators, typeahead, and selection changes are unavailable.
- Documents the API and theme state, adds regression coverage and read-only Storybook examples, and includes a Core changeset.

## Risk

The prop defaults to `false`, so existing behavior is unchanged. `isDisabled` retains precedence when both states are set; status and loading presentation remain independent.

## Testing

- `pnpm build`
- `vitest run` for Selector, MultiSelector, Selector source-build, and i18n resolution suites (363 tests)
- focused ESLint, knowledge, changeset, Storybook typecheck, and repository pre-commit checks
- real Chromium + axe: both read-only stories retained `combobox`, correct name/value, `aria-readonly="true"`, collapsed state, implicit listbox capability, and a localized "Read only" accessible description; neither rendered or opened a popup, listbox, clear action, or disclosure indicator after pointer and keyboard activation; zero axe violations

### DIFF
--- a/.changeset/selector-read-only.md
+++ b/.changeset/selector-read-only.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[component] Add `isReadOnly` to Selector and MultiSelector so selected values remain focusable and form-submittable without exposing selection menus or editing affordances.
+
+@cixzhang
--- a/apps/storybook/stories/MultiSelector.stories.tsx
+++ b/apps/storybook/stories/MultiSelector.stories.tsx
@@ -75,6 +75,7 @@ export const ReadOnly: Story = {
     onChange: () => {},
     hasClear: true,
     hasSearch: true,
+    htmlName: 'teams',
     triggerDisplay: 'labels',
     isReadOnly: true,
   },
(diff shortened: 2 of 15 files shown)


cixzhang (commented):
Verified at the current head: the implementation matches AST-011, the read-only state reaches Chromium assistive technology, and the review-loop findings are resolved. CI is rerunning.

[Reviewed by Robohands]

cixzhang:
/accept-visual 33523889691/1 These are baseline additions for Selector and MultiSelector, not changes to existing frames; the read-only light and dark states were reviewed and match AST-011, while the existing 700-shot set has zero pixel changes.

PR #5800, opened 2026-09-01
docs(spec): define authoring-time palette generation

cixzhang:
## Why

Current `main` now has an experimental Palette Generator Lab that produces and compares complete candidate palettes, while #5668 describes the downstream candidate → review → accepted palette workflow. That gives AST-008 a real authoring problem to contract now: reproducible candidates and a safe acceptance boundary.

The old AST-008 direction coupled palette metadata to `defineTheme.color` and automatic semantic expansion. That direction is rejected. A later palette edit must never silently recolor a theme.

## What

- Rewrites AST-008 from scratch around authoring-time candidate generation: complete-family inputs, versioned profiles, explicit modes and stop layouts, anchor policies, hard constraints, deterministic recipes, reproducibility evidence, and complete-family regression/CVD comparison.
- Defines the lifecycle as candidate → human review → accepted exact palette snapshot. Generated, hand-authored, and imported palettes are equal after validation; provenance is optional for non-generated data.
- Constrains any future mapping suggestion to a read-only, inspectable authoring result followed by explicit human acceptance and an atomic source patch containing final hex values.
- Removes all palette-backed `defineTheme` / `expandColorScale` behavior, validation-only Core helpers, runtime/default-bundle cost, the white/black runtime question, and the previously required public function/library/CLI/parity contracts.
- Leaves Ruby’s recipe and visual choice as OQ2. No algorithm, default, function, command, package, or public schema is accepted here.

## Ownership

- **AST-008:** candidate-generation requirements, reproducibility, and the candidate-to-acceptance boundary.
- **AST-018/#5668, if accepted:** palette shape, validation, association, inheritance, artifacts, and package compatibility. #5668 was read at current head but not modified.
- **Theme package records:** accepted exact palette values, saved mappings/deviations, states, compatibility, and rendered evidence.

Current theme architecture keeps productive `defineTheme` construction and compilation. Validation/checking stays separately named in theme tests/internal utilities or a separately accepted future CLI checker.

## Current consumer and future surfaces

The merged Sandbox lab is experimental decision tooling, not a supported public generator consumer. No current public generation library or CLI exists, so AST-008 specifies future-tool prerequisites rather than accepting an API.

If library and CLI surfaces are later proposed, each needs its own current consumer and separate API/compatibility acceptance. If both are accepted, they use one deterministic engine and agree on recipe semantics; AST-008 requires neither and accepts no shared public schema.

## Overlap

[...]

PR #5807, opened 2026-09-01
docs(spec): propose scoped Drawer contract

imdreamrunner:
## Why

Drawer is planned for Core, but #5550 exposes unresolved public decisions before promotion: whether an inspector belongs to the viewport or one region, whether regional blocking can honestly be called modal, whether modality and scrim are separate caller concepts, and how stacked Drawers behave across scopes.

## What

- Adds draft system spec `spec:AST-007` for Drawer scope, modality, scrim presentation, regional geometry, lifecycle, stacking, dismissal, and accessibility.
- Proposes `containerRef` as an explicit caller-owned regional target, with viewport scope on omission and a stable presented lifetime.
- Separates the candidate modality contract from scrim presentation while leaving that API choice open for owner judgment.
- Requires a regional blocking Drawer to stay outside the subtree it blocks, include the regional root itself, freeze blocked background scrolling, and avoid false `aria-modal` or body-lock claims.
- Defines scope-isolated visual/enforcement order, including regional targets inside versus outside an active document-modal subtree.
- Reconciles Drawer policy with accepted AST-003: eligible-owner and branch resolvers, same-gesture claims, local-dismissal migration, and modal outlets remain owned there rather than duplicated here.
- Explicitly leaves #5652's recede presentation, `hasStackRecede`, and theming variables to a separate design/theming decision.

## Relationship to #5550 and #5652

#5550 and #5652 remain implementation and presentation candidates. They establish feasibility and expose failure cases, but do not make this draft contract authoritative. The implementation must align with an accepted version of AST-007 before Core promotion.

AST-008 is already claimed by #5800; AST-007 is the lowest unclaimed system-spec ID on `main` and across open PRs.

## Risk

Knowledge only. No Lab/Core runtime, public package, generated artifact, visual output, or consumer behavior changes. The record remains `draft` / `proposed`; no Drawer API is accepted until the open questions are resolved and the exact head receives owner approval. No Changeset.

## Validation

- `pnpm check:repo`
- Prettier and staged diff checks
- Two independent contract reviews, including exact-head reconciliation with AST-002, AST-003, `architecture:layer-runtime`, `family:overlay-dismissal`, #5550, and #5652

## Review questions

1. Should `containerRef` be the explicit regional target, stable through each presented lifetime?
2. Should `modality` and `hasScrim` remain independent, including a visible modal scrim as the outside-pointer dismissal plane?
3. Should region-scoped blocking be called `modal` even though it is not document modality, `aria-modal`, `showModal()`, or body scroll locking?
4. Should a regional Drawer inside an active viewport modal be supported only as an explicitly associated child branch, with every unrelated regional target isolated as its own logical stack?

### DIFF
--- a/docs/specs/AST-007/spec.md
+++ b/docs/specs/AST-007/spec.md
@@ -1,0 +1,529 @@
+---
+schema_version: 1
+template_version: 1
+kind: system-spec
+id: spec:AST-007
+authority: draft
+archive_reason: null
+superseded_by: null
+approved_by: null
+approved_at: null
+phase: proposed
+owners: [cixzhang, rubyycheung, imdreamrunner]
+affects_architecture:
+  [architecture:layer-runtime, architecture:public-component-api]
+affects_families: [family:overlay-dismissal]
+affects_contributing: []
+affects_consumer_docs: [Drawer]
+---
+
+# Drawer scope, modality, and scrim system spec
+
+## Intent
+
+Let a Drawer belong either to the whole page or to a caller-owned region such as
+a split-view pane, dashboard card, canvas, or chart panel. A regional inspector
+must remain visually and behaviorally inside that region while the rest of the
+page stays available.
+
+The candidate contract separates three questions:
+
+1. **Scope:** which area owns the Drawer—the viewport or a supplied container.
+2. **Modality:** whether content behind the Drawer inside that scope is available.
+3. **Scrim:** whether that scoped background is visibly dimmed and exposes a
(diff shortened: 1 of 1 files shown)


cixzhang:
Closing as withdrawn. The bounded/pane-scoped Drawer direction this draft specifies was abandoned when the implementation PR #5550 closed, so this draft no longer represents active product direction. If the regional-inspector need returns, start from the current Drawer behavior and a newly approved scope/API decision rather than treating this record as accepted policy.

PR #5809, opened 2026-09-01
feat(knowledge): reserve system spec IDs

cixzhang:
## Why

Concurrent system-spec proposals can choose the same AST-NNN ID before either pull request lands.

## What

Adds a read-only allocator that considers landed specs and complete open-PR file inventories, a trusted exact-head collision check, and local path/frontmatter validation.

## Risk

No package runtime behavior changes. GitHub availability is required only for allocation and the remote collision gate; normal check:knowledge stays offline.

## Testing

- pnpm exec vitest run --project node scripts/lib/spec-id.test.mjs .github/scripts/spec-id-collision.test.mjs scripts/check-knowledge.test.mjs
- pnpm spec:id
- pnpm check:repo
- actionlint .github/workflows/spec-id-reservation.yml
- pnpm exec prettier --check on changed files

### DIFF
--- a/.github/scripts/spec-id-collision.cjs
+++ b/.github/scripts/spec-id-collision.cjs
@@ -1,0 +1,95 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+'use strict';
+/* global module, require */
+
+const {
+  listLandedSpecIds,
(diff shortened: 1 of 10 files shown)


cixzhang:

PR #5812, opened 2026-09-01
fix(ci): prevent no-op review signal cancellations

cixzhang:
## Why

A head update followed by a draft/ready lifecycle transition can enqueue two Review signal runs for the same commit. Workflow-level cancellation makes the useful run appear cancelled/red even though the later duplicate succeeds.

## What

- Stop starting Review signal for draft/ready transitions, which do not change its classification.
- Serialize only effect-bearing flag jobs without cancelling the running job.
- Load a dependency-free policy from trusted base code, reject superseded event heads, and recheck the live PR head before every write.
- Add mutation-sensitive coverage for the duplicate event, superseded head, current effect-bearing run, and stale-write suppression.

## Risk

Opened, synchronized, reopened, review-approval, and manual recovery paths remain covered. The workflow still never checks out or executes PR-controlled code.

## Testing

- 72 focused workflow/helper tests
- actionlint
- Prettier
- `pnpm check:repo`

No Changeset: workflow-only infrastructure change.

### DIFF
--- a/.github/scripts/review-signal-policy.cjs
+++ b/.github/scripts/review-signal-policy.cjs
@@ -1,0 +1,60 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+/**
+ * @input A review-signal event identity, the current PR head, and a proposed mutation.
+ * @output A fail-closed event plan and a current-head-guarded mutation result.
+ * @position Trusted policy shared by review-signal.yml and its workflow contract tests.
+ *
+ * SYNC: review-signal.yml loads this file with new Function(module, exports,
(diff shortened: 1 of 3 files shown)


cixzhang (commented):

An approved PR can still end with `review-required: pending`: flagging may snapshot reviews before an approval, Review clear then writes success, and the older flag run can overwrite it with pending on the same head. Because this status is required, maintainers are left with an approved but unmergeable PR and no later event guaranteed to repair it.

Please make approval the final projected state for this same-head interleaving, and add a focused test that pauses flagging after its review snapshot, lets Review clear succeed, then proves the final status remains success.


[Reviewed by Robohands]

cixzhang:

PR #5817, opened 2026-09-01
fix(sharedResizeObserver): independent subscriptions per element

freddymeta:
AST-010 §Implementation-requirements 9 asks for this and asks that it stay independently reviewable, so it is split out of #5783 rather than riding along in it.

## The bug

`callbacks` was a `Map<Element, ResizeCallback>` — **one callback per element**. `callbacks.set(element, cb)` overwrote, so a second hook observing the same node silently replaced the first, and either one calling `unobserveResize(element)` blinded the other.

Two hooks on one element is ordinary rather than exotic: a `TabList` root, a `useOverflow` container and a `useTruncation` target are all nodes another hook may reasonably watch. It is also unavoidable for AST-010 — FR11 puts every region of a multi-region hook on one shared container.

## What changes

- **reference-counted callbacks** per element, so duplicate callback references keep independent lifetimes without duplicate resize work;
- `observeResize` returns an **unsubscribe** that removes only its own registration — all seven callers in the package now use it;
- `unobserveResize(element, callback)` does the same by hand;
- the callback-less `unobserveResize(element)` still drops every callback on the element, kept for a caller that owns its element outright;
- dispatch snapshots the callback keys, so a callback may unsubscribe while the batch is running without skipping its neighbour.

## What this replaces

A test named `replaces callback when same element is observed twice` asserted the old behaviour deliberately. I think that contract is the bug rather than a decision — it is what makes FR11 unimplementable — but it was explicit, so I am flagging the removal rather than quietly rewriting it.

## Verification

- The exact duplicate-reference regression fails on the reviewed head: after the first cleanup, the still-mounted registration receives **0** callbacks instead of 1. It passes on this head, including an idempotent repeat of the first cleanup.
- `sharedResizeObserver.test.ts`: **11/11 pass**.
- Observer + Tokenizer focused suites: **83/83 pass**.
- `pnpm lint:strict`, `pnpm build`, and core typecheck pass on current `main`.
- The exact six-component browser RTL audit passes: **4 measured · 2 verified N-A · 0 gap · 0 stale**; the merged roster coverage suite passes **12/12**.
- Full `pnpm test` reaches the same two ContextMenu BottomSheet timing failures on clean current `main` and this branch; `ContextMenu.test.tsx` passes **48/48** in isolation. CI is the clean-run check.


cixzhang (changes requested):
Thanks for separating this prerequisite. A still-mounted region can stop following container resizes after a sibling unmounts: two `observeResize` calls sharing one callback collapse in the Set, and either cleanup removes both. Exact-head Chromium and Vitest both got zero callbacks where one should remain. Could each call keep an independent lifetime, with this duplicate-reference case added as a regression test?

[Reviewed by Robohands]

freddymeta:
Addressed the duplicate-reference lifetime bug from the latest review.

- Each element now reference-counts registrations per callback, so two `observeResize(element, sameCallback)` calls remain independent.
- Returned cleanup functions are idempotent; cleaning the first registration twice still leaves the second alive.
- Added the exact failing-first regression: reviewed head delivered 0 callbacks after the first cleanup; this head delivers 1, then 0 after the second cleanup.
- Migrated the Tokenizer observer caller added on current `main`, so no production caller uses whole-element cleanup.

Verification: 11/11 observer tests; 83/83 observer + Tokenizer tests; `lint:strict`, build, and core typecheck pass. Full-suite control hits the same two ContextMenu timing failures on clean `main` and this branch; that file passes 48/48 alone. CI is running on the pushed head.

— Puck, on behalf of Freddy


**CI follow-up (`ae2c0229a`):** `pr-rtl` exposed four real coverage gaps plus a synthetic `unknown/chat` duplicate. Added measured TabList coverage and source/story-verified N-A reasons for Table/Text. Current main supplied measured Chat coverage plus the unit-tested story-only roster fix; the final merge keeps that implementation. The exact AppShell/Chat/TabList/Table/Text/Tokenizer browser audit passes: **4 measured · 2 verified N-A · 0 gap · 0 stale**.

cixzhang (approved):
Thanks, this closes the shared-callback lifetime gap. The prior-head regression now fails before and passes here, including duplicate callback references and idempotent cleanup.

[Reviewed by Robohands]

PR #5827, opened 2026-09-01
fix(Spinner): the drawn frame follows a themed diameter

freddymeta:
## The bug

`--spinner-diameter` and `--spinner-stroke-width` set the ring in CSS, and the box the ring sits in is composed from those same two vars:

```
--_spinner-box-size: calc(var(--_spinner-ring-diameter) + var(--_spinner-ring-stroke) * 2)
```

The `<svg>` inside it, though, is sized and given its `viewBox` in JS from the size's own constants:

```jsx
const frameSize = diameter + border * 2;   // from SIZES[size]
<svg width={frameSize} height={frameSize} viewBox={`0 0 ${frameSize} ${frameSize}`}>
```

So there are two sizes for one thing, and only one of them is themeable. Theme the diameter and the box shrinks around an svg that stays at its default. The span is `display: inline-grid; place-items: center`, and an **overflowing** grid item aligns to start rather than centre (safe alignment), so the ring is pushed down and right, and its far edge is cropped.

Measured in Chromium against a build of `main`, applying EPS's ring sizes through the public vars only:

| size | box | svg | off-centre | clipped |
| sm | 11 | 14 | 1.50px | yes |
| md | 14.66 | 20 | 2.67px | yes |
| lg | 18.33 | 24 | 2.84px | yes |
| xl | 29.33 | 36 | 3.34px | yes |

The ring itself is drawn correctly at all four — `r` and `stroke-width` come from the vars and measure exactly what was asked for. It is only the frame that does not follow, which is why this looks like a size bug and is really a containment one.

It also makes the documented contract false today. `--spinner-diameter`'s own doc entry says:

> The rendered box is this plus the stroke width on each side, **and follows automatically**.

The box does follow. The svg does not.

## The fix

Size the svg in CSS and delete the `viewBox`:

```jsx
<svg width="100%" height="100%">
  <circle cx="50%" cy="50%" r={diameter / 2} strokeWidth={border} … />
```

With no `viewBox` one user unit is one pixel, so the CSS `r: calc(var(--_spinner-ring-diameter) / 2)` and `stroke-width` mean px — the same px the box is composed in — and the frame moves with the box. There is one coordinate system instead of two.

The `viewBox` was never doing any scaling here: it was always `0 0 frameSize frameSize` on an svg of exactly `frameSize`, i.e. 1:1. Removing it changes nothing about the default render and removes the only thing that could disagree with the box.

Two details fall out of centring on `50%`:

- The arc's twelve-o'clock offset was `transform={`rotate(-90 ${center} ${center})`}`, an SVG transform about a centre expressed in user units. There is no user-unit centre any more, so it becomes a CSS `rotate(-90deg)` with `transform-box: fill-box; transform-origin: center` — which is also the version that survives a themed diameter.
- The `r` and `stroke-width` **attributes** stay as the size's own numbers. They are the no-stylesheet fallback, and the CSS geometry properties outrank them whenever a stylesheet is present; they now mean px too, so the fallback and the themed render agree.

After, same measurement, same theme:

| size | box | svg | off-centre | clipped |
| sm | 11 | 11 | 0 | no |
| md | 14.66 | 14.66 | 0 | no |
| lg | 18.33 | 18.33 | 0 | no |
| xl | 29.33 | 29.33 | 0 | no |

## No change to the default

Unthemed, both builds render identically — box, ring, stroke and sweep:

| size | box | ring | stroke |
| --- | --- | --- | --- |
| sm | 14 | 10 | 2 |
| md | 20 | 14 | 3 |
| lg | 24 | 18 | 3 |
| xl | 36 | 28 | 4 |

Harness: both worktrees built with `pnpm -F @astryxdesign/core build`, the real `dist/astryx.css` inlined, the theme applied in `@layer astryx-theme` through the public vars only, animations frozen at `currentTime = 0`, geometry read from `getComputedStyle` and `getBBox` rather than a bounding rect (a spinning element's rect is the rotated AABB).

## Tests

`Spinner.test.tsx`: 48 pass. The four `sizes the %s ring from its size token` cases and `starts the arc at twelve o'clock` now assert the CSS-sized frame; one test is added — `sizes the frame from CSS, not from the size constant` — which is the assertion whose absence let this through: nothing tied the frame to the box, so nothing noticed when they disagreed.

## Why this and not a theme-side workaround

Reaching past it from a theme needs the private `--_spinner-box-size` plus a rule scaling the svg to it. That is two overrides of internals, in every theme that wants a non-default ring, to work around one line of JS.

Found while theming Astryx's Spinner to match an internal EPS spinner: colour, track and rotation all themed cleanly through public vars, and the size was the one axis that could not be done without touching internals.

`pnpm lint:strict` 0 errors, `typecheck` clean, `build` clean.


freddymeta:
## Embedded hosts: unchanged

Checked the components that render a Spinner internally, since those are the
placements a frame change could disturb — and the tightest of them (the Switch
thumb) is where it would show first.

Measured on a build of `main` and a build of this branch, same page, animations
frozen:

| host | box | svg | off-centre | clipped |
| Button `isLoading` — sm / md / lg | 14 | 14 | 0 | no |
| Switch `isLoading` — sm / md | 14 | 14 | 0 | no |
| CheckboxInput `isLoading` | 14 | 14 | 0 | no |
| Spinner inline beside text | 14 | 14 | 0 | no |
| Spinner in a 140px flex row beside `flex: 1 1 auto` | 20 | 20 | 0 | no |

**Byte-identical between the two builds** — every row above reads the same on
`main`. Which follows: none of these hosts themes the diameter, so the box
already equalled the frame, and this change only alters the case where the two
diverge.

embedded hosts

cixzhang (approved):

cixzhang:
/accept-visual 33570653280/1 Spinner appears to be correct

PR #5832, opened 2026-09-01
fix(Table): separate semantic row status from custom markers

cixzhang:
## User impact

People scanning dense tables get a consistent non-color glyph for semantic success, warning, and error. Existing custom-marker callers keep the same source shape: no `icon` renders the 8px dot, and an explicit `icon` renders the caller-selected glyph.

## Current contract

Implements the current [`module:Table/useTableRowStatus` contract](https://github.com/facebook/astryx/blob/83873b32d3031f2cc7e71e95c08c88b006c249c4/packages/core/src/Table/plugins/rowStatus/useTableRowStatus.spec.md), landed by [#5830](https://github.com/facebook/astryx/pull/5830).

## Semantic before → after

**Before:** current canaries overload `color`: `success`, `warning`, and `error` select both tone and an implicit glyph, while other values select paint only.

**After:** `getStatus` accepts the contract's exclusive semantic `{status, label}` and custom `{color, icon?, label}` branches. Semantic status resolves glyph and tone through the active theme. Custom `color` always selects paint; representation changes only when the caller supplies `icon`.

The stable `TableRowStatus` custom-marker interface remains unchanged and declaration-mergeable. The new `TableSemanticRowStatus` interface is additive, and the exclusive union exists only at the `UseTableRowStatusConfig.getStatus` callback boundary.

## Compatibility and scope

- Stable consumers need no source migration.
- Named custom icons preserve the released Icon color mapping.
- Raw CSS custom icons now use the caller's exact paint instead of falling back to the primary icon color; this is the intentional visual correction required by the current contract.
- Canary users relying on an implicit semantic glyph switch from `color` to `status`.
- A supported untyped `status` wins over `color`/`icon`, with one development warning per loaded module and the same production rendering without a warning.
- An unsupported untyped status falls through to a valid custom marker, following the contract's custom-branch precedence.
- No public `variant`/`presentation`, reflected state, Table row-status target, or theme selector was added.
- `ChatToolCalls` is intentionally unchanged; it needs a separate component contract.

## Test evidence

- Focused: 47 tests across runtime, public-subpath types/declaration merging, selection/expansion/custom/grouped composition, and performance/no-op identity.
- Types/docs: core typecheck, core docs typecheck, and Storybook typecheck.
- Repository: focused lint, Prettier, `check:repo`, full build, and public repository guard.
- GitHub CI: pending for this rebased head.
- No local accessibility audit was run; PR CI owns `pr-a11y`.


cixzhang (commented):

Thanks — the semantic/custom split works. Existing builders using `{color: 'warning', icon: 'clock'}` get a different glyph color: 0.5.2 renders `rgb(116, 91, 0)`, while this head renders `rgb(110, 53, 0)`. Raw CSS icon colors also switch from primary to the raw value. That conflicts with the PR and changeset claim that stable custom markers keep their 0.5.2 behavior. Please reconcile the compatibility claim with the rendered behavior: preserve the released mapping where the current contract allows, and explicitly document and test every intentional visual change. The branch also needs a merge with `main`.

[Reviewed by Robohands]

ernestt (approved):

cixzhang (commented):
Semantic verdict: approve

Thanks — this restores the stable custom-marker contract while adding the approved semantic branch. I verified the exact head across theme mapping, custom dots/icons, invalid untyped precedence, plugin composition, render budgets, normal pixels, and forced colors; 47 focused tests passed, and all completed CI signals are green.

[Reviewed by Robohands]

PR #5840, opened 2026-09-01
ci(visual): bound component screenshot scope

nynexman4464:
## Problem

A changed Core component currently expands visual review to every Storybook story across every accepted theme and both color modes. Behavioral and audit fixtures therefore multiply the screenshot baseline even when they do not represent distinct visual contracts.

On #5365, that produced 252 frames: 18 Breadcrumbs stories × 7 themes × 2 modes. Only two frames already existed, leaving 250 new screenshots to approve.

## Change

- Keep each touched component's representative story across accepted baseline themes and both modes.
- Add `visual-baseline` for additional default-theme visual contracts.
- Add `visual-theme-matrix` for visually dense stories that need all accepted themes, such as a component variant sheet.
- Use the same component story selection and accepted-theme derivation in both the ordinary PR planner and trusted visual-acceptance planner.
- Fail closed when any touched stable component has no representative story.
- Render every current accepted story contract under a changed theme in both planners, including components the theme begins overriding in this PR.
- Keep the focused component review ceiling in visual-gate config and apply it identically in both planners.
- Exclude theme-only matrices from the component ceiling; broad shared-infrastructure plans still use the same configured review limit.
- Apply one configured 5,000-shot safety ceiling to ordinary planning and capture, trusted planning, evidence publication, and baseline validation.
- Document and test the bounded, component-sensitive scope.

This preserves broad theme coverage where it carries signal while leaving behavioral and audit-only fixtures to their dedicated checks.

### Story-local configuration

Visual scope lives directly on the Storybook story object, next to the example it controls—not in a separate registry:

```tsx
export const CustomSeparator: Story = {
  tags: ['visual-baseline'],
  render: () => /* ... */,
};
```

- The representative story (`Default`, `Primary`, and similar conventional names) is selected automatically and needs no tag.
- `visual-baseline` adds that story in the default theme, light and dark.
- `visual-theme-matrix` adds that story in every accepted theme, light and dark.
- No visual tag leaves behavioral and audit fixtures to their dedicated checks.
- Existing `no-visual` excludes an unstable story from visual capture entirely.

For #5365, `Default` is the automatic Breadcrumbs representative and `CustomSeparator` is explicitly tagged `visual-baseline`; the remaining Breadcrumbs stories receive no screenshot baseline but continue to run through RTL and accessibility audits.

## Test plan

- Mutation proof: collapsing `visual-theme-matrix` into `visual-baseline` fails both planner-path tests because the opted-in variant story loses its non-default theme frames.
- Budget mutation proof: restoring the old combined 240-shot cap makes the new 242-shot theme-only regression fail; the fixed planner emits all 242 shots.
- Trusted evidence regression publishes 242 PNGs and records 242 added-frame decisions for a theme-only scope.
- Existing-theme regression starts with Button accepted only under Neutral and Card under Y2K, then verifies a changed Y2K plan includes both stories.
- Exact 240 component shots remain allowed; 241 fail closed. Explicit total limits are validated and cannot exceed the 5,000-shot safety ceiling.
- Full visual-gate suite — 284 passed.
- `pnpm lint:strict` — passed with existing warnings only.
- Planned #5365 against the live accepted baseline and its built Storybook: 16 frames total, 2 existing and 14 added. The representative Breadcrumbs story contributes 14 theme/mode frames; the focused separator story contributes 2 neutral frames.
- Captured all 16 planned frames successfully.


cixzhang:
Non-blocking: we may not need all themes to be checked. I think default and neutral should be okay to reduce further. We should also include the probe theme which we’re using to debug and validate theme infra.

cixzhang (dismissed):
REQUEST_CHANGES

Thanks for aligning the two component planners. The new 240-shot check runs after component and theme scopes are combined, so it rejects every stable-theme-only plan. Against the current accepted baseline, neutral is 782 shots: the parent produces the plan, while this head exits before capture. A theme author therefore gets no trusted visual evidence from the PR workflow. Please keep theme-only plans out of the focused component budget, derive both planner limits from one source, and add a >240 theme-only regression that still produces or explicitly defers evidence.

[Reviewed by Robohands]

INLINE
.github/scripts/visual-gate/visual-acceptance.mjs:788 — This also caps theme-only plans; current neutral reaches 782 shots and exits before capture.

PR #5843, opened 2026-09-01
fix(vibe-tests): run pnpm through cmd.exe on Windows in more scripts

HelloOjasMutreja:
Follow-up to #5841, which fixed the same pnpm-spawn-on-Windows bug in `internal/vibe-tests/src/fixture-suite.mjs`. This applies the identical fix to the three other spots that spawn `pnpm` directly without shell handling:

- `scripts/fixture-browser.mjs` (`spawnSync`, used by `fixtures:contrast`/`fixtures:screenshots`)
- `setup-test/run-setup.mjs` (`execFileSync`, used by `setup:prepare`)
- `setup-test/setup-canonical.test.ts` (`execFileSync`, its own inline `run()` helper)

## Why this matters even though it's not blocking `check:repo`

`setup-canonical.test.ts`'s pnpm-spawning tests are gated behind `ASTRYX_CANONICAL_SETUP_BROWSER=1`, which is only set by the `fixture-contrast` CI job — and that job runs on `2-core-ubuntu-arm`, so the bug is silent there. It only bites a Windows contributor who runs `setup:prepare`, `fixtures:contrast`/`fixtures:screenshots`, or the canonical setup test suite locally.

## Fix

Same pattern as #5841: shell out through `cmd.exe /d /s /c pnpm ...` on `win32`, call `pnpm` directly elsewhere. Every argument at every call site is a hardcoded literal, never user input, so this avoids Node's `DEP0190` shell-argument-escaping deprecation warning that plain `shell: true` would trigger.

## What I deliberately left alone

`run-setup.mjs` and `setup-canonical.test.ts` also both shell out to `cp -al` (GNU hardlink flag) to link a prepared `node_modules` into each sandbox. That's a separate, pre-existing Windows breakage (no Windows equivalent for the flag, and creating hardlinks this way fails with `Permission denied`) — confirmed with a real run, out of scope for this PR.

## Testing

- Directly verified both spawn mechanisms (the `spawnSync` variant used in `fixture-browser.mjs`, and the `execFileSync` variant used in the other two) against a real `pnpm --version` call on Windows — both succeed.
- Ran `setup-canonical.test.ts` with `ASTRYX_CANONICAL_SETUP_BROWSER=1` for real: it got past both `runPnpm` calls (`pnpm install`, `pnpm -F @astryxdesign/core build`) successfully before hitting the separate, expected `cp -al` failure described above — real evidence the fix works, not just that it compiles.
- Confirmed the `public-artifact-cli.test.mjs` failures seen in the same test run are pre-existing and unrelated: they fail identically against the unfixed baseline (a different bug, `.bin/tsx` needing the same treatment — filed as its own follow-up).
- `npx eslint internal/vibe-tests/setup-test/setup-canonical.test.ts` — clean. The two `.mjs` files are outside the linted `packages/cli` scope, matching the project's existing `.mjs` lint-ignore convention.

### DIFF
--- a/internal/vibe-tests/scripts/fixture-browser.mjs
+++ b/internal/vibe-tests/scripts/fixture-browser.mjs
@@ -28,11 +28,28 @@ const CONTRAST_PROBES = [
   ['destructive action', 'destructive-action'],
 ];
 
-function run(command, args, cwd) {
-  const result = spawnSync(command, args, {cwd, encoding: 'utf8'});
+/**
+ * pnpm is a .cmd (batch) file on Windows. spawnSync can only run a batch
+ * file through a shell — a bare 'pnpm' fails with ENOENT and an explicit
+ * 'pnpm.cmd' still fails with EINVAL (batch files need a shell even named
+ * exactly). Shelling out through cmd.exe /c directly, rather than
+ * spawnSync's shell:true, avoids Node's shell-argument-escaping deprecation
+ * warning (DEP0190) — every argument here is a hardcoded literal, never
+ * user input, so we build the argv ourselves instead of asking spawnSync to
+ * build a shell string. See internal/vibe-tests/src/fixture-suite.mjs for
+ * the same pattern.
+ */
+function runPnpm(args, cwd) {
+  const result =
+    process.platform === 'win32'
+      ? spawnSync('cmd.exe', ['/d', '/s', '/c', 'pnpm', ...args], {
+          cwd,
+          encoding: 'utf8',
+        })
+      : spawnSync('pnpm', args, {cwd, encoding: 'utf8'});
   if (result.status !== 0) {
     throw new Error(
-      `${command} ${args.join(' ')} failed in ${cwd}\n${result.stdout}${result.stderr}`,
+      `pnpm ${args.join(' ')} failed in ${cwd}\n${result.stdout ?? ''}${result.stderr ?? result.error?.message ?? ''}`,
     );
   }
 }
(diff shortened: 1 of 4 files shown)


cixzhang (changes requested):
Windows setup still cannot reach browser measurement: `setup-measure.mjs` keeps calling `spawnSync('pnpm', ['build'])`, even though `run-setup.mjs` directs every prepared run there. Node cannot launch pnpm's `.cmd` shim that way, so Windows records a failed build and skips both browser passes. Please route this call through the same `cmd.exe` path before landing.

[Reviewed by Robohands]

HelloOjasMutreja:
Confirmed and fixed: `setup-measure.mjs`'s `build()` had the same bare `spawnSync('pnpm', ['build'])` call. Fixed in the latest commit, same cmd.exe pattern as the other three files. Verified for real (not just source): ran the fixed `build()` logic against an already-prepared Windows sandbox and got `status: 0, ok: true`, versus `status: null, error: ENOENT` from the unfixed bare call against the same sandbox.

cixzhang (changes requested):
Thanks for fixing the measurement build. The same Windows failure remains twice in `setup-canonical.test.ts`: the real install at line 2141 and build at line 2165 bypass `runPnpm` and call `spawnSync('pnpm', ...)` directly. On Windows those cases fail before testing `allowBuilds`, so the canonical suite still misses its stated path. Please route both through the helper too. If you’d rather talk it through with someone, we’re in [Discord](https://discord.com/invite/XnsUcFykEP).
[Reviewed by Robohands]

HelloOjasMutreja:
Confirmed and fixed both. Both call sites use spawnSync's non-throwing form deliberately (a failing install/build is the assertion under test at those two spots, not an error), so they can't reuse the throwing runPnpm helper. Added a spawnPnpm counterpart with the same cmd.exe routing and switched both to it.

Verified for real with ASTRYX_CANONICAL_SETUP_BROWSER=1: both "installs and builds when allowBuilds is in pnpm-workspace.yaml" (exercises both the install and build spawns) and "fails exactly as the measured cell did when the key is in package.json" (exercises the install spawn's failure path) pass on Windows.

Fixed in 05441d004.

cixzhang (approved):

PR #5844, opened 2026-09-01
feat(theme): add theme-local token support

rubyycheung:
## Summary

- implement the accepted AST-006 `localTokens` authoring contract from #5801
- preserve exact theme-local names through runtime CSS, static builds, inheritance, and generated built-theme metadata
- validate enrolled declarations, references, namespaces, lineage, and cycles without changing legacy `tokens` behavior
- document the API for theme authors and keep the generated theme template synchronized

## Scope

This is infrastructure only. It does not adopt local tokens in Neutral or change any shipped theme colors or component styling. Neutral adoption remains separate from the cross-theme API review.

An inheritance-only descendant keeps its existing theme name, including legacy names that are not lower-kebab-case. The stricter name rule applies only when a theme explicitly supplies `localTokens`.

## Validation

- focused Core/CLI/template tests: 129 passed
- `pnpm check:sync`
- `pnpm check:changesets`
- repository pre-commit validation suite

Extracted from #5752 so the reusable API can be reviewed independently from Neutral's visual design decisions.


cixzhang (changes requested):
Theme authors who opt into `localTokens` can currently receive invalid or contradictory output even though the build succeeds.

- [`CSS_VAR_PATTERN`](https://github.com/facebook/astryx/blob/e76533d693caa8ce71884bd52ae740c4a9696442/packages/core/src/theme/localTokens.ts#L6-L10) only recognizes lowercase `var(`. Valid CSS such as `VAR(--astryx-theme-foo-color-fill)` bypasses undeclared-reference and cycle validation, so an enrolled static build can emit a broken reference. Please parse the function case-insensitively and add runtime/static undeclared + cycle regressions for uppercase/mixed-case spelling.
- A key duplicated in legacy `tokens` and `localTokens` is accepted and [emitted twice](https://github.com/facebook/astryx/blob/e76533d693caa8ce71884bd52ae740c4a9696442/packages/core/src/theme/generateThemeRules.ts#L347-L356). CSS uses the later local declaration, while `resolveThemeToken()` reads the portable `tokens` value. Please reject cross-map duplicate names before output and test CSS/helper parity.
- This is a settled implementation of AST-006, so the implementation PR also needs to update the canonical shipped records required by [AST-006’s current-state impact](https://github.com/facebook/astryx/blob/e76533d693caa8ce71884bd52ae740c4a9696442/docs/specs/AST-006/spec.md#L205-L221) and [completion criteria](https://github.com/facebook/astryx/blob/e76533d693caa8ce71884bd52ae740c4a9696442/docs/specs/AST-006/spec.md#L263-L294): AST-006’s phase plus `theme-authoring-contract`, `theme-compilation`, and `theme-tokens`. The Neutral theme record is not involved because this PR adds infrastructure without an adopted Neutral local-token mapping.

rubyycheung:
Addressed the requested changes in `747d9c0b5`:

- made CSS `var()` detection case-insensitive while preserving the exact custom-property name, with runtime and static regressions for uppercase and mixed-case undeclared references and cycles
- reject names duplicated across `tokens` and `localTokens` after inheritance/override resolution, before output; added legacy CSS/`resolveThemeToken()` parity coverage and static atomic-output coverage
- moved AST-006 to `shipped` and updated `theme-authoring-contract`, `theme-compilation`, and `theme-tokens` to describe the shipped contract and cite the deciding spec decisions

Local validation passed: 218 focused tests, Core build/typecheck, CLI strict/authoring/template-doc typechecks, Neutral build, ESLint, and `pnpm check:repo`.

cixzhang (approved):
Thanks, this addresses the validation and contract-sync issues from my prior review. No new concerns.

[Reviewed by Robohands]

PR #5845, opened 2026-09-01
feat(Spinner): add --spinner-arc-fraction public var

### LINKED ISSUE #5819: Spinner: arc sweep is the one ring geometry a theme cannot change
## What is missing

`Spinner` exposes four public vars on the `spinner` target — `--spinner-diameter`, `--spinner-rail-width`, `--spinner-color`, `--spinner-track-color` (#5214). A theme can change how big the ring is, how thick it is, and what colour both parts are.

It cannot change **how much of the circle the moving arc covers**. That is a compiled-in constant:

```ts
// packages/core/src/Spinner/Spinner.tsx:35
const ARC_FRACTION = 0.375;          // 135°, per the docblock above it
const ARC_DASH = PI * ARC_FRACTION;
const ARC_GAP  = PI * (1 - ARC_FRACTION);
```

## Why it is worth a var

Every other dimension of the ring became themeable in #5214 for a stated reason: a theme should be able to make the ring its own. Sweep is the one geometric property left out, and it is the most visually distinctive one — a 135° arc and a 270° arc read as different spinners at the same diameter, stroke and colour.

It also looks permitted rather than excluded. #5771's draft record says under **AV1 — Size and shade**: "The existing size and shade values and theme overrides *may change ring geometry and paint through the `spinner` target* without changing anatomy ownership." Arc sweep is ring geometry, and a var on the existing target adds no new target — which is what that record's non-goals rule out ("no wrapper, SVG, track, or arc *target*").

The concrete consumer: Meta's EPS design system draws its spinner as a single ~270° arc. Its own component is scheduled to be replaced by this one, and at that point the sweep halves with no theme-side way to keep it.

## The catch, which is why this is an issue and not a PR

The CSS side is nearly free — `styles.arc` already composes its dash from `calc()`, so it becomes:

```ts
strokeDasharray: `calc(var(${RESOLVED_DIAMETER}) * ${PI} * var(--spinner-arc-fraction, 0.375)) ...`
```

The **inline `strokeDasharray` attribute** is the problem:

```tsx
// Spinner.tsx:546 — computed in JS, cannot read a custom property
strokeDasharray={`${arcLength} ${circumference - arcLength}`}
```

Its comment says it is "the size's own dash, for the render with no stylesheet". A themed sweep would therefore draw 135° in that frame and snap once the stylesheet lands — the same class of box-vs-paint disagreement #5214 spent most of its length fixing, and #5484 fixed again for flex hosts.

So the question for a maintainer is which is wanted:

1. **Accept the pre-CSS frame** — the attribute stays the default, a themed sweep applies once styles load.
2. **Drop the attribute** and let CSS own the dash entirely; the no-stylesheet render loses its arc.
3. **Not worth it** — the sweep is deliberately part of Astryx's identity and should not be themeable, in which case #5771 could say so explicitly and close the question.

Happy to implement whichever, with the Chromium before/after the other Spinner changes have used.

## Not blocking

Nothing is broken today; defaults are unaffected either way. Raising it now only because #5771 is actively recording Spinner's theming contract, and this is the kind of thing that is cheaper to settle inside that record than after it.

HelloOjasMutreja:
Fixes #5819.

## What's missing

`Spinner` exposes four public vars on the `spinner` target (`--spinner-diameter`, `--spinner-stroke-width`, `--spinner-color`, `--spinner-track-color`) so a theme can change size, stroke, and color. It cannot change how much of the circle the moving arc covers — that was a compiled-in constant (`ARC_FRACTION = 0.375`, a 135deg sweep).

## Why this was an issue and not immediately a PR

The catch the issue raised: the CSS side (`styles.arc.strokeDasharray`) already composes its dash from a `calc()`, so making it themeable there is close to free. But there's also an inline `strokeDasharray` SVG attribute — the pre-stylesheet render, for SSR / no-JS — computed in JS, which can't read a CSS custom property. A themed sweep would draw the default 135deg in that frame and snap to the themed value once the stylesheet lands.

## Why this is safe: precedent already set

Checked how `--spinner-diameter` (the var this issue explicitly compares itself to) already handles this exact split, since it's used in both places too:

- CSS side (`styles.arc`, `styles.circle.r`, etc.): reads the live, themeable `RESOLVED_DIAMETER` var.
- Inline attributes (the `<circle r={diameter / 2}>` presentation attributes, and the pre-CSS `strokeDasharray`): `const {diameter} = SIZES[size]` — the hardcoded default, not the themed value.

So a themed diameter already only takes visual effect once the stylesheet loads, and the pre-CSS frame already renders at the unthemed default. This isn't a new kind of box-vs-paint disagreement (the class of bug #5214 and #5484 dealt with) — it's applying the one this component already ships, to one more property.

## Change

- `styles.arc.strokeDasharray` now composes the dash from `var(--spinner-arc-fraction)` (a live number multiplier) instead of the pre-multiplied `ARC_DASH`/`ARC_GAP` constants.
- `--spinner-arc-fraction` declares alongside `--spinner-diameter`/`--spinner-stroke-width` in `sizeStyles`, same default (`0.375`) for every size, unregistered like the other three public vars (no length-arithmetic ambiguity to guard against — it's a pure multiplier, not summed with anything).
- The inline `strokeDasharray` attribute (the pre-stylesheet render) is untouched — still `ARC_FRACTION`, matching how the inline diameter attributes stay at their default too.

## Testing

- `npx vitest run packages/core/src/Spinner/Spinner.test.tsx` — 48/48 pass, including a new test confirming a themed `size:xl` target scopes `--spinner-arc-fraction` correctly (mirroring the existing diameter/color scoping tests).
- Built `@astryxdesign/core` and inspected the compiled `dist/astryx.css` directly: confirmed `--spinner-arc-fraction:.375` declares on the size-variant rule and `strokeDasharray` composes from `var(--spinner-arc-fraction)` in the compiled `calc()`, not just in source.
- Typecheck and lint clean.

Posted the precedent-based reasoning on the issue first; given how directly it settles the ambiguity the issue raised (and the reporter's own "happy to implement whichever"), opening this as the concrete proposal rather than waiting further — happy to adjust based on any maintainer feedback.

### DIFF
--- a/.changeset/spinner-arc-fraction-var.md
+++ b/.changeset/spinner-arc-fraction-var.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] Add a `--spinner-arc-fraction` public var to Spinner, so a theme can change how much of the ring the moving arc covers (defaults to 0.375, a 135deg sweep), the same way it already retheme diameter, stroke width, and color.
+
+@HelloOjasMutreja
--- a/packages/core/src/Spinner/Spinner.doc.mjs
+++ b/packages/core/src/Spinner/Spinner.doc.mjs
@@ -48,6 +48,7 @@ export const docs = {
       {name: '--spinner-stroke-width', description: 'Stroke width of both circles the ring is drawn from: the moving arc and the track behind it. Set it per size alongside the diameter. One stroke width drives both, so 0 is honoured as a zero-width stroke and paints nothing at all rather than falling back to the default; for an arc with no track behind it, set --spinner-track-color to transparent instead.', default: '2px (sm), 3px (md), 3px (lg), 4px (xl)'},
       {name: '--spinner-color', description: "Color of the moving arc. Defaults to the shade's token, so set it on a shade-variant target to retheme one shade (spinner: { 'shade:subtle': { '--spinner-color': 'var(--color-text-tertiary)' } }), or on the base target to retheme all four. Accepts any color notation, including var(), color-mix() and currentColor.", default: 'var(--color-accent) (default), var(--color-text-secondary) (subtle), var(--color-on-dark) (onMedia), currentColor (inherit)'},
       {name: '--spinner-track-color', description: 'Color of the track the arc travels on. Set it to `transparent` for an arc with no track. The onMedia and inherit shades draw the track at reduced alpha (30%) so it reads against an arbitrary backdrop; that fade applies to a themed color too.', default: 'var(--color-track) (default, subtle), var(--color-on-dark) (onMedia), currentColor (inherit)'},
+      {name: '--spinner-arc-fraction', description: "Fraction of the ring the moving arc covers, as a plain number (not a percentage or angle). Set it on a size-variant target the same way as --spinner-diameter, e.g. spinner: { 'size:xl': { '--spinner-arc-fraction': '0.75' } } for a 270deg sweep. Only takes effect once the stylesheet loads; a render with no CSS (SSR, no-JS) always draws the default 135deg arc.", default: '0.375 (135deg), same for every size'},
     ],
   },
   usage: {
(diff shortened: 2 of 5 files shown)


HelloOjasMutreja:
Rebased onto current main to clear staleness, which surfaced two real fixes:

- The derivedVarRegistry consistency test now requires every documented var to have a derived[] mapping or a VARS_WITHOUT_DERIVED_MAPPING entry. --spinner-arc-fraction joins the other four Spinner SVG-drawing vars there for the same reason already documented for them: no standard CSS property maps onto an SVG dash-length multiplier.
- My own new theming test asserted a stale compiled selector shape (.astryx-spinner.xl) instead of the current one (.astryx-spinner[data-size="xl"]) that the sibling diameter-scoping test right above it already uses.

typecheck, lint, and the full Spinner (46) and derivedVarRegistry (100) suites all pass now.

cixzhang (approved):
Thanks I think this is reasonable

PR #5755, opened 2026-08-31
Rename built-in syntax theme identifiers

rubyycheung:
## Summary
- renames the Neutral syntax theme id from `xds-neutral` to `astryx-neutral`
- updates the generated CLI Neutral theme template to match

## Notes
- This does not rename the main Neutral theme; it only changes the syntax-theme identifier used for code highlighting.

## Validation
- pnpm -F @astryxdesign/theme-neutral build
- pre-commit checks from commit hook

PR #5847, opened 2026-09-01
Rename built-in syntax theme identifiers

rubyycheung:
## Summary

- rename each built-in syntax-theme identifier from the legacy XDS prefix to its Astryx theme name
- update the generated CLI theme templates to match
- keep published theme objects and CSS behavior otherwise unchanged

## Scope

This is the standalone `main`-based replacement for #5755. It no longer waits behind the retired mixed Neutral PR #5752.

## Validation

- `pnpm check:sync`
- `pnpm check:changesets`

The identical change was approved by @cixzhang on #5755 before the stack was untangled.

### DIFF
--- a/.changeset/neutral-syntax-theme-name.md
+++ b/.changeset/neutral-syntax-theme-name.md
@@ -1,0 +1,14 @@
+---
+'@astryxdesign/theme-butter': patch
+'@astryxdesign/theme-chocolate': patch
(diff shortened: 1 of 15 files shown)


cixzhang (approved):
Thanks for separating this cleanup. The seven published theme objects and CSS remain byte-identical; generated theme source now uses Astryx-prefixed syntax names.

[Reviewed by Robohands]

PR #5848, opened 2026-09-01
Add Table row status theming target

rubyycheung:
## Summary

- add an `astryx-table-row-status` theming target with `color` and `presentation` visual props
- preserve the current behavior where success, warning, and error use their themed semantic icons by default
- keep palette and raw CSS colors as dots unless the caller supplies an icon
- allow themes to style dot and icon presentations without changing caller-controlled status semantics
- record the new target in Table's canonical anatomy/specification

## Scope

This is the standalone `main`-based replacement for #5754. It no longer waits behind the tonal-palette metadata stack.

## Validation

- focused row-status tests: 8 passed
- `pnpm check:sync`
- `pnpm check:knowledge`
- `pnpm check:changesets`

The rebase preserves newer default semantic-icon behavior that landed on `main` after the original stacked PR was created.

### DIFF
--- a/.changeset/table-row-status-theming-target.md
+++ b/.changeset/table-row-status-theming-target.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] Add a theming target for Table row status indicators.
+
+@rubyycheung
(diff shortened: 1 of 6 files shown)


rubyycheung:
Closing after rebasing against current `main` exposed the newly merged row-status contract in #5830.

That owner-approved module record explicitly keeps `table-row-status`, `color`, and `presentation` theme selectors outside the current contract pending a separate AST-002 proposal that proves the theme-author need and painter guarantees. Shipping this implementation now would contradict that current record.

The implementation remains preserved on `codex/table-row-status-theming-main` and in the earlier archived stack. Neutral's future color-role PR will omit the row-status mappings unless that separate proposal is approved.

PR #5846, opened 2026-09-01
fix(theme-neutral): refine SegmentedControl inset

rubyycheung:
## Summary

- increase the Neutral SegmentedControl inset to `var(--spacing-1)`
- reduce each item height by the matching 8px so the control's outside height remains unchanged
- remove the selected-item shadow for a flatter Neutral treatment
- keep the bundled CLI Neutral template synchronized and add a focused contract test

## Scope

This is the SegmentedControl styling extracted from #5752. It contains no local-token API work and no status-color changes.

## Validation

- focused Neutral and template tests: 11 passed
- `pnpm -F @astryxdesign/theme-neutral build`
- `pnpm -F @astryxdesign/cli typecheck:template-docs`
- `pnpm check:sync`
- `pnpm check:changesets`
- repository pre-commit validation suite

## Before merge

- capture and approve exact-head light/dark SegmentedControl visuals

## Stack

Base: #5754

PR #5849, opened 2026-09-01
fix(theme-neutral): refine SegmentedControl inset

rubyycheung:
## Summary

- increase the Neutral SegmentedControl inset to `var(--spacing-1)`
- reduce each item height by the matching 8px so the control's outside height remains unchanged
- remove the selected-item shadow for a flatter Neutral treatment
- keep the bundled CLI Neutral template synchronized and add a focused contract test

## Scope

This is the standalone replacement for #5846 and the SegmentedControl styling extracted from #5752. It contains no local-token API work and no status-color changes.

## Validation

- focused Neutral test: 1 passed
- `pnpm check:sync`
- `pnpm check:changesets`

## Before merge

- capture and approve exact-head light/dark SegmentedControl visuals

## Stack

Base: #5848

### DIFF
--- a/.changeset/neutral-segmented-control-inset.md
+++ b/.changeset/neutral-segmented-control-inset.md
@@ -1,0 +1,8 @@
+---
+'@astryxdesign/theme-neutral': patch
+'@astryxdesign/cli': patch
+---
+
+[fix] Give Neutral segmented controls a roomier inset while preserving their outside height.
+
+@rubyycheung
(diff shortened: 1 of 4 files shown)


rubyycheung:
Superseded by #5851, which contains the same SegmentedControl-only styling directly against `main`.

The intermediate Table row-status dependency was removed after #5830 established that a direct row-status theming target requires a separate AST-002 proposal.

PR #5850, opened 2026-09-01
fix(SideNavItem): keep a consumer aria-label in collapsed mode

### LINKED ISSUE #5641: SideNavItem overwrites a consumer aria-label in collapsed mode

HelloOjasMutreja:
## Problem

`SideNavItem` forwards a consumer `aria-label` through `...rest`, but both collapsed-mode paths (icon-only link/button, and the popover trigger for items with children) spread a hard-coded `aria-label: label` after `...rest`, so it always won over a caller-supplied value.

That drops accessible context exactly where the visible label is absent — icon-only collapsed navigation.

## Fix

Both paths now use `rest['aria-label'] ?? label`, so an explicit `aria-label` wins in every render mode, and the label text stays the fallback when no consumer value is given.

## Testing

- Added two regression tests under the existing `SideNavItem (collapsed)` describe block: one for the plain collapsed link/button path, one for the popover-trigger (items-with-children) path. Both fail against the pre-fix code and pass after.
- Full `SideNav.test.tsx` suite: 196/196 passing.

Fixes #5641

### DIFF
--- a/.changeset/sidenavitem-collapsed-aria-label.md
+++ b/.changeset/sidenavitem-collapsed-aria-label.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] SideNavItem no longer overwrites a consumer-supplied `aria-label` with the visible `label` text in collapsed mode (icon-only link/button and popover-trigger paths), so attention/context conveyed only through `aria-label` survives collapse. (#5641)
+
+@HelloOjasMutreja
--- a/packages/core/src/SideNav/SideNav.test.tsx
+++ b/packages/core/src/SideNav/SideNav.test.tsx
@@ -1138,6 +1138,36 @@ describe('SideNavItem (collapsed)', () => {
     expect(trigger).toHaveAttribute('aria-label', 'Settings');
   });
 
+  it('keeps a consumer aria-label on the collapsed link, not the label text (#5641)', () => {
+    renderCollapsed(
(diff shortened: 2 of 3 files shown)


HelloOjasMutreja:
Closing as a duplicate of my own earlier #5644, which I missed checking for before starting this branch. Same fix, same files. Continuing there instead.

PR #5851, opened 2026-09-01
fix(theme-neutral): refine SegmentedControl inset

rubyycheung:
## Summary

- increase the Neutral SegmentedControl inset to `var(--spacing-1)`
- reduce each item height by the matching 8px so the control's outside height remains unchanged
- remove the selected-item shadow for a flatter Neutral treatment
- keep the bundled CLI Neutral template synchronized and add a focused contract test

## Scope

This is the standalone `main`-based SegmentedControl styling extracted from #5752. It contains no local-token API work, status-color changes, or Table row-status theming.

## Validation

- focused Neutral test: 1 passed
- `pnpm check:sync`
- `pnpm check:changesets`

## Before merge

- capture and approve exact-head light/dark SegmentedControl visuals

### DIFF
--- a/.changeset/neutral-segmented-control-inset.md
+++ b/.changeset/neutral-segmented-control-inset.md
@@ -1,0 +1,8 @@
+---
+'@astryxdesign/theme-neutral': patch
+'@astryxdesign/cli': patch
+---
+
+[fix] Give Neutral segmented controls a roomier inset while preserving their outside height.
+
+@rubyycheung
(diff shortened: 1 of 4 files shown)


cixzhang:
Seems good as a theme update. Was the problem related to accessibility of the segmented control? Was curious if you thought we should fix the component instead of the theme.

rubyycheung:
@cixzhang This is more of a visual preference, both ways works for color contrast. The current shadow is super intense and throws off the visual balance. I also want us to be more purposeful when we add elevation.

PR #5852, opened 2026-09-01
feat(TextArea, TextInput): forward the native autoComplete attribute

### LINKED ISSUE #5638: TextArea cannot express the native autocomplete attribute

HelloOjasMutreja:
## Problem

Neither `TextArea` nor `TextInput` exposes the native `autocomplete` attribute: both `TextAreaProps` and `TextInputProps` extend `BaseProps` only, not the corresponding native HTML attributes type, so there's no typed way to set it, and consumers were left with a ref-mutation workaround.

## Fix

Coordinated addition, per the discussion on #5638:

- `TextArea`: `autoComplete?: React.TextareaHTMLAttributes<HTMLTextAreaElement>['autoComplete']`, forwarded unchanged to the `<textarea>`.
- `TextInput`: `autoComplete?: React.InputHTMLAttributes<HTMLInputElement>['autoComplete']`, forwarded unchanged to the `<input>`.

Both follow the same explicit-prop passthrough pattern already used for `placeholder`/`htmlName`, keeping `BaseProps`' narrow surface intact rather than widening either component to the full native attributes type.

## Testing

- One forwarding test and one "omitted stays unset" test per component, matching the existing `htmlName prop` describe blocks' structure.
- Full `TextArea.test.tsx` + `TextInput.test.tsx` suites: 161/161 passing.
- Doc entries (English + Chinese) added to both `.doc.mjs` files.

Fixes #5638

### DIFF
--- a/.changeset/textarea-textinput-autocomplete.md
+++ b/.changeset/textarea-textinput-autocomplete.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] TextArea and TextInput now accept a native `autoComplete` prop, forwarded unchanged to the underlying element, so consumers no longer need a ref-mutation workaround to set browser/form autofill behavior. (#5638)
+
+@HelloOjasMutreja
--- a/packages/core/src/TextArea/TextArea.doc.mjs
+++ b/packages/core/src/TextArea/TextArea.doc.mjs
@@ -256,6 +256,12 @@ export const docs = {
       description: 'Callback fired when the textarea loses focus.',
     },
     {
+      name: 'autoComplete',
+      type: 'string',
+      description:
+        "The native autocomplete attribute, forwarded unchanged to the <textarea>. Browser/form autofill behavior, not a styling escape hatch; unset leaves the browser's default.",
+    },
+    {
       name: 'xstyle',
       type: 'StyleXStyles',
       description:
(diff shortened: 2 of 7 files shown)


HelloOjasMutreja:
Closing as a duplicate of my own earlier #5643, which I missed checking for before starting this branch. Same fix (functionally identical diff), same files. Continuing there instead.

PR #5853, opened 2026-09-01
fix(TreeList): let a consumer onKeyDown cancel built-in navigation

### LINKED ISSUE #5584: [Bug] TreeList consumer onKeyDown cannot cancel built-in navigation

HelloOjasMutreja:
## Problem

A consumer-supplied `onKeyDown` reaches TreeList through `...restProps` on the outer `<div>`, while the built-in arrow-key/expand-collapse handler (`useTreeFocus`) runs on the inner `<ul role="tree">`. The native keydown event starts at the focused treeitem and bubbles outward, hitting the closer `<ul>` ancestor first. So the built-in handler always ran, and moved focus, before the consumer's handler (on the further `<div>`) ever saw the event, let alone had a chance to call `preventDefault()` in time.

## Fix

`onKeyDown` is now composed on the same element (`<ul>`): the consumer's handler runs first, and the built-in navigation only proceeds if `event.defaultPrevented` is still false afterward. A non-canceling handler still lets the APG tree keyboard model run exactly as before.

## Testing

- Added the exact regression test from the issue: focus the first treeitem, fire `ArrowDown` with a `preventDefault()`-calling `onKeyDown`, assert focus and the roving tab stop are unchanged.
- Added a companion test confirming a non-canceling handler still lets navigation proceed.
- Full `TreeList.test.tsx` suite: 83/83 passing.
- Added `onKeyDown` to both English and Chinese `.doc.mjs` prop tables (it existed implicitly via `BaseProps` before, undocumented).

Fixes #5584

### DIFF
--- a/.changeset/treelist-onkeydown-precedence.md
+++ b/.changeset/treelist-onkeydown-precedence.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] TreeList's `onKeyDown` prop now runs before the built-in arrow-key/expand-collapse handling, so calling `event.preventDefault()` reliably cancels the built-in behavior for that key instead of running too late to matter. (#5584)
+
+@HelloOjasMutreja
--- a/packages/core/src/TreeList/TreeList.doc.mjs
+++ b/packages/core/src/TreeList/TreeList.doc.mjs
@@ -125,6 +125,12 @@ export const docs = {
           slotElements: [{__element: 'Text', props: {type: 'body'}, children: 'Header'}],
         },
         {
+          name: 'onKeyDown',
+          type: '(e: KeyboardEvent) => void',
+          description:
+            "Runs before TreeList's own arrow-key navigation and expand/collapse handling. Call event.preventDefault() to cancel the built-in behavior for that key; focus and the roving tab stop are left untouched. A handler that doesn't call preventDefault() still lets the built-in keyboard model run afterward.",
+        },
+        {
           name: 'xstyle',
           type: 'StyleXStyles',
           description:
@@ -203,6 +209,12 @@ export const docsZh = {
           type: 'ReactNode',
           description:
             '标题内容，通过 aria-labelledby 与树关联。',
+        },
+        {
+          name: 'onKeyDown',
+          type: '(e: KeyboardEvent) => void',
+          description:
+            '在 TreeList 自身的方向键导航和展开/折叠处理之前运行。调用 event.preventDefault() 可取消该按键的内置行为，焦点和 roving tab stop 保持不变。若处理函数未调用 preventDefault()，内置的键盘模型仍会随后运行。',
         },
         {
           name: 'xstyle',
--- a/packages/core/src/TreeList/TreeList.test.tsx
+++ b/packages/core/src/TreeList/TreeList.test.tsx
@@ -1050,6 +1050,21 @@ describe('TreeList', () => {
     expect(treeitems[1]).toHaveFocus();
   });
 
+  it('still delivers keydowns from the header slot to the consumer onKeyDown (#5853)', () => {
+    const handleKeyDown = vi.fn();
+    render(
+      <TreeList
+        items={flatItems}
+        header={<input data-testid="header-input" />}
+        onKeyDown={handleKeyDown}
+      />,
+    );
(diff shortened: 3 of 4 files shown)


cixzhang (changes requested):
Thanks for fixing the cancellation order. One compatibility issue remains: `onKeyDown` is still typed through `BaseProps<HTMLDivElement>` and the ref points to the root `<div>`, but this head invokes it with the inner `<ul>` as `event.currentTarget`. It also stops delivering keydowns from the public `header` slot. Focused tests and Chromium at this exact head reproduce both. Please keep the consumer handler on the root while limiting built-in tree navigation to events originating inside `ul[role="tree"]`; `preventDefault()` should still leave focus and the roving tab stop unchanged.

[Reviewed by Robohands]

HelloOjasMutreja:
Both confirmed, thanks for catching these. Fixed by moving the consumer handler back to the root `<div>`, using `onKeyDownCapture` instead of a plain `onKeyDown` on `<ul>`. Capture phase runs top-down, so it still fires before the built-in bubble-phase handler on `<ul>` (same precedence guarantee), and since the root is also where `header` mounts as a sibling of `<ul>`, the header slot's keydowns are delivered again. The `currentTarget` cast is gone too since the listener genuinely lives on the div now.

Added a regression test for the header-slot case (keydown fired on an element inside `header` reaches the consumer's `onKeyDown`). The two existing precedence tests still pass unchanged.

cixzhang (changes requested):
Thanks—the earlier review’s root-element and header-slot issues are fixed: `event.currentTarget` is now the `<div>`, and header keydowns arrive. One compatibility issue remains: the public `onKeyDown` prop is now wired to `onKeyDownCapture`, so consumers receive phase 1 instead of the normal bubble phase, and a descendant that calls `stopPropagation()` still triggers the TreeList handler. Focused exact-head probes reproduce both changes.

Please keep `onKeyDown` as a root bubble handler, run it before TreeList’s built-in behavior, and invoke the built-in handler only when the event belongs to this TreeList and was not prevented. That preserves cancellation, the root `currentTarget`, header delivery, and normal descendant propagation semantics.

[Reviewed by Robohands]

HelloOjasMutreja:
Closing as superseded — #5606 already landed an equivalent fix on main (`handleRootKeyDown`: consumer handler runs first on the root, checks `defaultPrevented`, then scopes built-in navigation to events where `treeRef.current.contains(e.target)`). That also correctly handles the header-slot case my second commit here was fixing. Thanks for the review pass either way, it caught a real regression before this would have landed on top of the wrong base.

PR #5858, opened 2026-09-01
test(rtl): make fixture applicability order-independent

nynexman4464:
## Problem

The weekly RTL audit reported `SVGIconRegistry` as a D1 directional-icon failure because its reference gallery renders raw left and right glyph samples. Those glyphs are fixture data, not directional controls, so mirroring them would make the registry inaccurate.

The first fix filed the fixture under the real `Lab/SVGIcon` component, but D1 still selected whichever story happened to appear first in Storybook's generated index. Reordering the unchanged registry story before `Default` therefore changed SVGIcon from N/A to a failure and made its applicability baseline stale.

## Changes

- Stack on #5988's applicability-debt baseline; `lab/SVGIcon` remains explicit known coverage debt rather than gaining a broad verified-N/A exemption.
- Add typed, colocated `rtlAudit.D1` Storybook parameters with a required rationale for `applicable: false`.
- Expose Storybook's resolved meta/story parameters to the Playwright audit, preserving normal story-over-meta precedence.
- Select D1 representatives by stable story source-file/export order instead of generated-index object order.
- Mark only the raw registry fixture D1-inapplicable. D5 and D6 continue to inspect it normally.

No production component behavior changes.

### Rendered mutation proof

Built the final stack, reordered `dist/index.json` so `lab-svgicon--default-registry-icons` appeared before `lab-svgicon--default`, then ran the focused audit.

Before the fix (restoring only the two audit implementation files from the parent commit):

- exit 1;
- D1: 0 pass, 1 fail;
- registry story selected and both raw chevrons reported as unmirrored;
- `lab/SVGIcon` became a stale known coverage gap.

After restoring the fix with the same reordered index:

- exit 0;
- D1: 0 fail, 1 N/A;
- stable source order selected `lab-svgicon--default`;
- `lab/SVGIcon` remained one known coverage gap with no stale entry.

Also removed every other SVGIcon story from a test copy of the built index. With only the registry fixture present, the audit read its colocated D1 exclusion and rationale, returned N/A, and preserved `lab/SVGIcon` as known coverage debt.

The representative-order comparison across the complete built index changed **0 existing component selections**.

### Automated checks

- `pnpm exec vitest run --project node .github/scripts/rtl-audit-coverage.test.mjs .github/scripts/weekly-rtl-summary.test.mjs` — 31 passed.
- Storybook typecheck passed.
- `pnpm lint:strict` passed with existing warnings only.
- Full package build and Storybook build passed.

Depends on #5988.


cixzhang (changes requested):
The weekly RTL report is only clean while `lab-svgicon--default` happens to be the first story. Moving the unchanged registry entry first makes D1 fail on its two raw chevrons and marks the new declaration `stale-verified-na`. Please make fixture applicability independent of story order while keeping the real `lab/SVGIcon` component accounted for.

[Reviewed by Robohands]

cixzhang (approved):
We asked for fixture applicability to stop depending on Storybook order. This head selects the real default deterministically, skips only the raw fixture, and keeps `lab/SVGIcon` as known debt.

[Reviewed by Robohands]

PR #5859, opened 2026-09-01
fix(i18n): localize Timestamp relative formats

### LINKED ISSUE #5074: InternationalizationProvider locale is ignored by calendar, date/time, Timestamp, and PowerSearch formatting

nynexman4464:
## Problem

Timestamp's absolute formats follow `InternationalizationProvider`, but `relative`, `relative_short`, and the relative branch of `auto` still assemble English phrases manually. A French provider therefore renders `2 hours ago`, `now`, and `yesterday` in English.

## Change

- Preserve Timestamp's existing second/minute/hour/day/month/year thresholds and future-clock-skew handling.
- Format the selected value and unit with the provider locale through `Intl.RelativeTimeFormat`.
- Use the native `long` style for `relative`/`auto` and `narrow` for `relative_short`.
- Keep long-form `now` and `yesterday` idioms, while compact output stays numeric.
- Cache formatters by locale, style, and numeric mode.
- Add the pure formatter to the strict raw-Intl infrastructure boundary.

Native formatting also replaces the English-only `in a few seconds` special case with the locale's exact seconds form, such as `in 31 seconds`.

`Intl.RelativeTimeFormat` formats a signed quantity and unit; it has no calendar option and does not convert or inspect calendar dates. Gregorian enforcement therefore does not apply here. The raw-Intl lint did catch direct construction in `Timestamp.tsx`; construction now lives in a narrowly approved pure helper that requires the provider locale.

## Visual evidence

The permanent English Storybook story covers every past and future transition point. French below is a temporary provider override used only for evidence.

### Before — French provider ignored

Before: every relative label remains English under a French provider

### After — English

After: English relative labels at every transition point

### After — French

After: French relative labels at every transition point

Fixes #5074.

## Test plan

- Regression proof against the old implementation: three new provider-locale cases failed with English output (`2 hours ago`, `2h ago`, and `now`) where French was expected.
- `pnpm exec vitest run --project ui packages/core/src/Timestamp/Timestamp.test.tsx` — 92 passed.
- `pnpm exec vitest run --project node internal/eslint-plugin-astryx/no-raw-intl-locale.test.mjs` — 100 passed.
- Permanent Storybook story: English past/future transition points at present, seconds, minutes, hours, days, months, and years.
- `pnpm rtl:audit --filter Timestamp` — verified N/A with 0 coverage gaps; Timestamp has no directional visual or behavior.
- `pnpm lint:strict` — passed with existing warnings only.
- `pnpm -F @astryxdesign/core build` — passed.


cixzhang (approved):
Thanks this is good. The provider-locale regression is fixed across relative, compact, and auto modes. Please resolve the generated RTL-audit list conflict before merge.

[Reviewed by Robohands]

nynexman4464:
/accept-visual 33568304679/1 screens look good, test is essentially plain default text

cixzhang (approved):
Thanks, this is good. Non-English providers now get native relative timestamps across relative, compact, and auto modes, and the merge resolves the RTL-audit conflict.

[Reviewed by Robohands]

PR #5860, opened 2026-09-01
Fix detail template layout spacing inheritance

rubyycheung:
## Summary
- add semantic `--astryx-layout-padding*` variables for Layout region defaults without changing the existing meaning of `components.layout.base.padding`
- wire LayoutHeader, LayoutContent, LayoutFooter, and LayoutPanel to that semantic layout padding variable chain when their padding prop is omitted
- remove hardcoded default layout padding from the Order Detail template so header/content/panel inherit layout region defaults together

## Compatibility
- Existing themes that set `components.layout.base.padding` still emit raw `padding` on `.astryx-layout`, matching the previous behavior.
- Themes that want responsive/default Layout region spacing can opt into the new semantic variables directly, e.g. `components.layout.base['--astryx-layout-padding']`.
- Existing explicit `padding` props on LayoutHeader, LayoutContent, LayoutFooter, and LayoutPanel continue to win over the semantic default.

## Validation
- pnpm exec vitest run packages/core/src/theme/defineTheme.test.ts packages/core/src/theme/generateThemeRules.test.ts packages/core/src/theme/derivedVarRegistry.test.ts packages/core/src/Layout/Layout.test.tsx packages/core/src/Layout/overlayPaddingReset.test.tsx
- pnpm -F @astryxdesign/core typecheck
- pnpm check:sync && pnpm check:knowledge && pnpm check:changesets
- pnpm exec eslint packages/cli/assets/templates/pages/detail-page/page.tsx packages/core/src/Layout/Layout.tsx packages/core/src/Layout/LayoutHeader.tsx packages/core/src/Layout/LayoutContent.tsx packages/core/src/Layout/LayoutFooter.tsx packages/core/src/Layout/LayoutPanel.tsx packages/core/src/Layout/padding.stylex.ts packages/core/src/Layout/Layout.doc.mjs packages/core/src/Layout/overlayPaddingReset.test.tsx packages/core/src/theme/derivedVarRegistry.ts packages/core/src/theme/derivedVarRegistry.test.ts packages/core/src/theme/defineTheme.test.ts packages/core/src/theme/generateThemeRules.test.ts

### DIFF
--- a/.changeset/themeable-layout-padding.md
+++ b/.changeset/themeable-layout-padding.md
@@ -1,0 +1,8 @@
+---
+'@astryxdesign/core': patch
+'@astryxdesign/cli': patch
+---
+
+[fix] Allow themes to control default Layout padding and let the detail page template inherit layout region spacing defaults.
+
+@rubyycheung
--- a/packages/cli/assets/templates/pages/detail-page/page.tsx
+++ b/packages/cli/assets/templates/pages/detail-page/page.tsx
@@ -52,7 +52,7 @@ import type {CSSProperties} from 'react';
 // meets the header divider. No edge-dock prop on TabList (#2622).
 const tabsRow: CSSProperties = {
   marginInline: -12,
-  marginBottom: -16,
+  marginBottom: 'calc(-1 * var(--container-padding-block-end))',
   marginTop: 12,
 };
 // Pull the list items' inner padding back so their content optically aligns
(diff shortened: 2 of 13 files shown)


cixzhang (changes requested):
Thanks, this gets the template onto inherited spacing, but the new directional defaults break full-bleed children. With 8px inline-start / 32px inline-end, LayoutContent applies 8/32 yet publishes 8/8, so Table ends 24px short in both LTR and RTL. Please keep each published edge paired with its applied padding and exercise the seven documented variables in a real themed story; exact-head theme-var-reachability currently rejects all seven.

[Reviewed by Robohands]

rubyycheung:
Closing this for now. We are not taking the broad semantic Layout variable approach as the dependency for mobile spacing; the current direction is to keep mobile spacing scoped to Cindy's mobile theme work through component/theme overrides, and handle template hardcoded-spacing cleanup later as a separate audit.

PR #5861, opened 2026-09-01
docs: record the CLI surface and its contribution rules

josephfarina:
`packages/cli` has no knowledge records. 43 `.spec.md` records cover 42 core components and one theme; nothing covers the CLI, so a reviewer has nothing to check a CLI change against. These two drafts mirror the component-API pair — `AST-002` describes the policy, `api-conventions.md` makes it practical.

**`docs/architecture/cli-surface.md`** — the shipped surface as twelve invariants, each naming the test or check that proves it. The CLI has one audience, an agent in a subprocess, and every rule follows from that: no interactivity at all, one JSON envelope per emission, `code` is the contract and the `error` prose is not, human text is a projection produced only by the formatters, one command per file, exit codes that do not depend on the output mode. `check:cli-structure` already enforces the layout with nothing written behind it; this is that contract.

**`docs/contributing/cli-conventions.md`** — when a new command is justified (four gates, and a code owner approves the proposal first), and what makes a good flag. Flags do not need a proposal, but they do need a **closed composition matrix**: every pair with an existing flag on that command either composes with a test, is refused with a message, or cannot co-occur. An undecided cell ships as behaviour nobody chose.

`theme build --family` (#5687) is the worked example. Its matrix is closed except for one cell: `--family` + `--check` compose, but `--check` now answers differently depending on whether `--family` was passed, because the two modes emit different CSS — so CI has to check with the same flags it built with. No test covers that pair. That is the kind of thing this guide is meant to surface during review rather than after release.

Both records are `authority: draft`, so neither governs review yet. Reading them as a pair is the ask; promote whichever holds up.

No changeset: pure spec records publish no package changes.

### DIFF
--- a/docs/contributing/cli-conventions.md
+++ b/docs/contributing/cli-conventions.md
@@ -1,0 +1,199 @@
+# CLI conventions for contributors
+
+This guide turns the CLI surface architecture into the steps you follow when
+you change `packages/cli`. It does not create policy. Where it and
+`docs/architecture/cli-surface.md` disagree, the architecture record wins.
+
+## Who the CLI is for
+
+The caller is an agent. It runs the CLI in a subprocess, reads `--json`, and
+acts on the result without a person watching. A person reading the text output
+is a supported reader, never the caller the design serves.
+
+Two things follow, and they settle most arguments before they start:
+
+- **The CLI must not impede the agent's flow.** No prompt, no confirmation, no
+  question. A command that cannot finish returns an error with a code the agent
+  can branch on and a suggestion it can act on.
+- **The output is data first.** `--json` is the source of truth. The text output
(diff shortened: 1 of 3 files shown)


docs/contributing/cli-conventions.md
```diff
@@ -0,0 +1,199 @@
...
+
+1. **They compose** — and a test proves it.
+2. **They are refused together** — with a clear message and a code.
+3. **They cannot co-occur** — because another rule already refuses the
+   combination that would reach them.
+
+An undecided cell is the defect. It ships as behaviour nobody chose, and an
+agent finds it before a person does.
+
+### Worked example: `theme build --family`
```
cixzhang, line 90: Oh nice I needed this

cixzhang (approved):

PR #5865, opened 2026-09-02
feat(templates): rebuild the searchable table as an interrogable invoice

ernestt:
Rebuilds the `table-page` template — a searchable table, whose fixture happens to be an invoice — and takes it from **74 (C) to 96 (A)** on the template rubric.

This is the first of five table-template changes, submitted on its own so each can be reviewed independently.

## What it looks like

| Light | Dark |
|---|---|
| light | dark |

**Per-row hover detail** — anchored to the whole row, uncontrolled so the card survives the pointer travelling onto it and the Dispute button stays clickable:

hover

**Filtered** — the totals retotal, and a banner says so rather than letting a partial figure pass as the whole:

filtered

**Narrow (768px)**:

narrow

### 1. The masthead scrolls now

`Layout height="fill"` makes the content area the scroll container, so anything in the `header` slot is pinned. Scrolling this page parked the title and the Download button over a sliced-off search box and scope toggle — the controls vanished under a bar that had nothing to do with them.

| Before — header pinned, controls sliced | After — masthead scrolls away |
|---|---|
| before | after |

A document's own title is not chrome. Title, actions and the metadata block are now one masthead inside the content, and they scroll together. The `header` slot stays right for what remains useful mid-scroll — an app bar, a toolbar, a sticky action row — and the template's guidance now says so.

`contentWidth` still caps the column, and still does it on the `Layout` rather than on the table, so full-bleed dividers keep spanning the window.

### 2. The description describes the shape, not the fixture

The old string led on the fixture — "billing metadata", and a synonym list of *Invoice, bill, statement, receipt* — which is the case [Contributing Templates §5b](https://github.com/facebook/astryx/wiki/Contributing-Templates) names outright: sample data is interchangeable, so it cannot be the thing that distinguishes a template. Anyone reaching for a searchable table of orders, runs or assets wants exactly this page and matched almost none of its vocabulary.

It also missed the words people actually type. The stemmer strips only `ing|ed|ies|es|s`, so `sortable` never matches a query for **sort** — the word boundary fails. `column`, `record` and `list` were absent entirely. Run against `scoreCandidate`'s actual stem regex over a 20-term probe:

| | terms hit | missed |
|---|---:|---|
| before | 10/20 | sort, column, record, list, flat, group, nested, data, uniform, dataset |
| after | 20/20 | — |

The replacement differentiates on data shape (flat, uniform, one homogeneous list) and behaviour (search, filter, sort, totals that follow the filter), says what the siblings do differently, and carries both the verb and the `-able` form of each behaviour. No billing vocabulary remains.

> Worth a second opinion: the rubric's own worked example keeps domain nouns in the synonym slot as *interchangeable examples* ("a customer order with shipping, an invoice, a transaction, or a job"). I dropped them entirely per the ask, which costs this template any query phrased as "invoice table". Happy to add back a spread of three or four unrelated domains if you'd rather have that reach — the thing the rubric actually forbids is one domain being definitional, not domain nouns existing.

### 3. The extending notes are reorganised around search, filter and sort

They were organised around the invoice fixture and around whatever had been hard to build — the author's order, not the reader's. Someone opening this template wants a searchable table.

Now under four headings — narrowing, sorting, saying what the reader is looking at, and structure — with the search/filter/sort material promoted from incidental notes to the bulk of the file:

- when each of the three narrowing surfaces is the right one, and the failure mode of shipping all three so two controls do the same job
- that typed filter controls come from the search config rather than the column, and what an `enum` field missing its `enumValues` silently degrades to
- narrow once into one array; give every control one reset
- **sort after filtering, not before**
- **sort the value, never the label** — extended past currency to dates and enums, with the general rule stated: keep the comparable primitive on the row and format it in `renderCell`, so sorting and display never read the same field
- the two filtered states, and detecting on filter *presence* rather than row count

The mechanical constraints worth keeping — no footer in data mode, the two-part flush fix, per-row detail without a `rowComponent`, tabular figures — are kept, but demoted below the guidance and stripped of invoice specifics.

This also corrects an ARIA claim: `status="warning"` renders `role="alert"`, not `role="status"`.

### Sitting flush takes two fixes, not one

A `Table` bleeds past its container: the scroll wrapper applies negative inline margins equal to `--container-padding-inline-*`, and the outer cells add that padding back so text still lines up. That is right for a table that owns its surface, and wrong here, where the table shares a content line with a `MetadataList`.

The previous version fought this with a wrapper `<div>` that zeroed the two variables inline. This version turns them off at the source — `LayoutContent padding={0}`, with the padding moved to a `VStack`, which is a plain flex box and publishes nothing. The wrapper `<div>` is gone.

That only gets half of it. Cell padding resolves to `max(var(--container-padding-inline-start), spacing-2)`, and that `spacing-2` is a **hard floor**, so zeroing the variable still leaves the text indented 8px while the metadata above it sits at 0. A twelve-line `useFlushEdges` plugin closes it.

Those two `paddingInline` declarations are the only custom CSS left in the file, and that floor is the one genuine core gap remaining. A future `hasContainerBleed={false}` that also drops the floor would take this to 99.

### Totals are not a table

They used to be a second `<Table>` sharing the first one's `<colgroup>`, because `TableFooter` is children-mode only and the plugins need data mode. That cost a hand-written `<colgroup>`, a `<col>` per column, a border declaration on every ruled cell, and an extra table in the a11y tree papered over with a labelled group.

A right-aligned stack gets the same alignment for none of it: the figure box is the Amount column's width and both are flush to the content line, so they share an edge by construction. `Divider` draws the rules.

This also fixes a latent bug — the hand-rolled hairline used `1px solid var(--color-border, rgba(5, 54, 89, 0.1))`, and that hardcoded fallback is a dark blue that nearly vanished against a dark background.

## Core dependency

The sortable header alignment fix this template wanted now lives in its own PR: **#5928**. This PR is a pure template change — no `packages/core` diff, no changeset.

Until #5928 lands, the sortable numeric headers here (Qty, Rate, Amount) sit start-aligned above their right-aligned figures. That is the pre-existing behaviour rather than a regression.

## Rubric — 96/100 (A)

| Criterion | Score | Notes |
|---|---:|---|
| Astryx Component Purity | 30/30 | 0 raw HTML elements |
| Icon Purity | 15/15 | heroicons only, no inline SVG |
| Custom CSS | 12/15 | 2 declarations, both justified — the cell-padding floor cannot be reached from a template |
| Layout & Structure | 15/15 | `Layout` root, `contentWidth`, single page |
| Doc Metadata | 9/10 | 5a 6/6 · 5b 3/3 after the rewrite · naming 0/1 — `-page` is a stopword suffix, kept because renaming is breaking (exact-match resolution, no aliases) |
| Image Handling | 5/5 | no images |
| Code Quality | 10/10 | |
| **Total** | **96/100** | up from 74 (C) |

One correction to the earlier revision of this table: it claimed Doc Metadata 9/10 while the description was still fixture-led, which §5b scores at 2/3 ("sample-data nouns are doing the separating"). The honest number before change 2 was 8/10, total 95. It is 96 now.

## Testing

- Typecheck and eslint clean; `page.tsx` typechecks with zero errors of its own
- Search verified end to end: 19 rows to 2 on `build`, banner appears, "Clear all filters" resets both the rows and the input
- Scroll behaviour checked at 1440px and 768px, light and dark — the masthead leaves the viewport and the table header does not stick to anything
- Description term coverage measured against `scoreCandidate`'s actual stem regex rather than by eye — the 10/20 → 20/20 table above
- Tabular figures verified in the rendered DOM: every currency cell computes `font-variant-numeric: tabular-nums`, and all 38 of them resolve to exactly two distinct right edges (the Rate and Amount columns). Figtree's default digits are proportional — ten `1`s measure 63px against ten `0`s at 87px — so this is load-bearing, not decorative
- Document outline checked: the hover card mounts only while open, so the page has one `h1` at rest and one `h2` while hovering, not nineteen orphaned headings


cixzhang (dismissed):
Thanks, the invoice direction is strong, and the sortable numeric headers now align correctly in LTR and RTL.

One keyboard path is broken: every row is a tab stop, but on first render `useHoverCard` attaches while the row is still non-focusable, so `focusTrigger: 'auto'` omits its focus listeners. In Chromium, hover and touch open details; focusing the same row only paints a ring. Filtering later makes retained rows work while newly mounted rows still fail, so access depends on interaction history. Please make row focus open the same details consistently; the existing `focusTrigger: 'always'` pattern covers this shape.

The core alignment change is consumer-visible and also needs a patch changeset.

[Reviewed by Robohands]

ernestt:
Both addressed in 9c48ff5 and 6cdc4b5. One correction on the first, because the diagnosis and the symptom came apart when I went looking.

**Focus trigger.** The mechanism you describe is exactly right. `interactionRef` resolves `focusTrigger: 'auto'` through `isFocusable(el)` at attach time, and the effect hands over the `<tr>` *before* setting `tabIndex`, so that first attach binds no focus listeners.

The symptom I could not reproduce. I read the listeners straight off the row over CDP, and `focusin` is bound on every row in both configurations:

```
row 0  tabindex=0  focusin=true  mouseenter=true
row 2  tabindex=0  focusin=true  mouseenter=true
row 5  tabindex=0  focusin=true  mouseenter=true
```

The reason is a second-order one: `attach` changes identity on the re-render after mount, so the effect tears down and re-runs — and its cleanup drops `aria-describedby` but leaves `tabIndex` on the node. The re-attach therefore sees a row that *is* focusable and binds the listeners. Tab to any row on `main` and the card opens; newly mounted rows behave the same, since they inherit the same second pass.

So this was not a broken keyboard path, but it was a keyboard path standing on an accident — one that quietly disappears the day `useHoverCard` memoizes its handlers more tightly, with nothing in the template to catch it. `focusTrigger: 'always'` is the right call either way, so I have taken it and named the reasoning at the call site. Behaviour verified after the change: focus opens, blur closes, hover still opens, and a row mounted after a filter round-trip behaves identically.

**Changeset.** Added as a patch, describing the alignment fix in consumer terms and noting that the flow-relative `justify-content` is what keeps it mirroring under RTL.

One thing I did not touch: `pr-rtl` is red, and it is not this PR. It reports `0 fail` on all three dimensions and fails only on `COV: 1 gap` — `Table` returned all-N/A with no recorded reason. The registry the audit README points at (`verified-not-applicable.json`) does not exist in the tree yet, so there is no sanctioned way to record that reason from here, and asserting `Table` is RTL-N/A would be wrong given the sticky-column shadow the README itself cites as a past finding. The job is `continue-on-error: true` by design, and #5873 and #5822 both merged with it red. Happy to pick the coverage gap up separately if you would rather it not sit.

ernestt:
Split the core change out, per @ernesttien — this PR is now a pure template change, with no `packages/core` diff and no changeset.

The sortable header alignment fix and its four tests moved to **#5928**, along with the patch changeset you asked for. Reviewing it there also makes the consumer-visible part easier to judge on its own terms, rather than buried under an 800-line template.

Consequence worth stating plainly: until #5928 lands, the Qty / Rate / Amount headers here sit start-aligned above their right-aligned figures. That is the pre-existing behaviour, not a regression this PR introduces, and we would rather ship the template showing the real current state than have it depend on an unlanded fix.

The keyboard fix from your review (`focusTrigger: 'always'`) stays here, since it is template-local.

PR #5867, opened 2026-09-02
Define responsive surface adaptations

rubyycheung:
## Review contract

- **Owner:** [AST-014](https://github.com/facebook/astryx/blob/responsive-presentation-spec/docs/specs/AST-014/spec.md) (`draft`, `proposed`)
- **Upstream contract:** [AST-012](https://github.com/facebook/astryx/blob/main/docs/specs/AST-012/spec.md) owns the shared environmental-adaptation vocabulary, breakpoint names and boundary semantics.
- **Semantic before → after:** Before, responsive surface selection is split between `presentation` for Astryx-owned overlays and `nativePicker` for platform-owned pickers. After, `adaptations` is the single forward owner: omission or `"default"` selects a component-owned policy, while a policy object replaces it with a baseline surface and ordered rules.
- **Classification:** `novel-human`
- **Review requested:** @cixzhang and @imdreamrunner

## Why

Selector-family components currently expose `presentation`, while date/time inputs expose `nativePicker`. Keeping both as permanent selectors creates overlapping ownership and conflicting combinations. AST-012 has also established `adaptations` as Astryx's environmental vocabulary, so this proposal now builds on that accepted contract rather than introducing an unrelated responsive model.

## Proposed contract

```ts
type SurfaceAdaptations<Surface extends string> = {
  default: Surface;
  rules?: Array<{
    from?: 'sm' | 'md' | 'lg' | 'xl' | '2xl';
    below?: 'sm' | 'md' | 'lg' | 'xl' | '2xl';
    pointer?: 'coarse' | 'fine';
    value: Surface;
  }>;
};

adaptations?: 'default' | SurfaceAdaptations<ComponentSurface>;
```

- Omission, `undefined`, and `adaptations="default"` use the component's built-in recommended policy.
- A policy object replaces that built-in policy rather than merging with hidden rules.
- Conditions within a rule are ANDed; later matching rules win.
- Width names and boundaries resolve through the nearest Theme under AST-012.
- Surface values remain component-qualified.
- Legacy `presentation` and `nativePicker` values require lossless, conflict-free migration.

## Compatibility correction

TimeInput is now separate from DateInput and DateTimeInput:

- DateInput/DateTimeInput `nativePicker="never"`: Astryx Popover on fine pointers and Astryx BottomSheet on coarse pointers.
- TimeInput `nativePicker="never"`: Astryx typed field on both fine and coarse pointers.
- TimeInput therefore uses a `typed | native` surface vocabulary and does not gain a BottomSheet path.

## Scope

Specification only. No runtime or public API changes. AST-014 remains draft pending exact-head owner approval. Current component behavior and defaults remain authoritative until separately reviewed implementation and migration pull requests land.

## Risk

Documentation only. The remaining risk is compatibility planning for released `presentation` and `nativePicker` props, which remains explicit in the open questions.

## Validation

- `pnpm check:repo`
- `pnpm exec prettier --check docs/specs/AST-014/spec.md`
- `git diff --check`

No Changeset: no published package behavior changes.


cixzhang:
Encouraging start—`presentation` is clear, predictable, and extends more cleanly than component-specific props like `nativePicker`.

**AST-002 concern:** preserving both creates multiple owners for one outcome. For example, `presentation="bottom-sheet"` with `nativePicker="always"` is conflicting, while `nativePicker="touch"` is itself an adaptive presentation policy hidden behind another prop.

**Forward API:** consider converging on one visual API prop that owns responsive adaptation across platform- and Astryx-owned surfaces. The benefit is one coherent, extensible decision model; the tradeoff is defining whether `adaptive` promotes one canonical policy or supports multiple legitimate strategies—for example, Dialog adapting to fullscreen, Drawer, BottomSheet, or remaining modal.

rubyycheung:
Agreed. I updated the proposal in eab8eecac so `presentation` is now the single forward owner across Astryx- and platform-owned surfaces; `nativePicker` is compatibility-only rather than a permanent parallel axis.

The revised contract also makes the adaptive-policy choice explicit:

- `adaptive` is one documented canonical policy per component, rather than one universal surface mapping for every component.
- If caller evidence establishes a second legitimate adaptive strategy, it must receive a distinct value in the same `presentation` prop and pass AST-002.
- The date/time migration must preserve all three released `touch` / `always` / `never` behaviors and prevent or diagnose simultaneous `presentation` + `nativePicker` use, with no hidden precedence.

I left the durable name/shape for today's `nativePicker="never"` behavior as OQ1 instead of collapsing it into a lossy rename: that behavior is itself Astryx-owned adaptation between Popover on fine pointers and the touch surface on coarse pointers. That seems like the remaining owner decision needed before an implementation API can be proposed.

cixzhang (commented):
One compatibility detail needs fixing before this can become the owning contract. `TimeInput nativePicker="never"` keeps the typed field on both fine and coarse pointers; it does not switch between Popover and an Astryx touch surface. API3 and OQ1 currently group it with DateInput and DateTimeInput, so a later migration could introduce new TimeInput behavior while claiming to preserve today's contract. Please separate TimeInput's mapping and keep its typed-field path explicit.

[Reviewed by Robohands]

rubyycheung:
Updated in 703a7362b to adopt the owner-selected `adaptations` contract. The revision now builds explicitly on accepted AST-012 for breakpoint names, boundary semantics, nearest-Theme resolution, conjunction, and authored-order precedence while keeping the compact component-surface rule shape. It also separates TimeInput compatibility: `nativePicker="never"` maps to the typed field on both coarse and fine pointers, rather than the DateInput/DateTimeInput Popover-to-BottomSheet policy. The PR description and open questions now reflect the replacement API and migration work. `pnpm check:knowledge -- --base origin/main`, `pnpm check:repo`, Prettier, and `git diff --check` pass. Re-review requested.

rubyycheung:
Closing in favor of Ivor’s accepted AST-012 contract in #5806. AST-012 now owns the shared `adaptations` vocabulary and environmental rule semantics; any component-specific surface behavior or compatibility migration can be proposed from that foundation when implementation work proceeds. Keeping the branch and discussion available as historical context.

PR #5868, opened 2026-09-02
Draft spacing profile for mobile theme

rubyycheung:
## Status

`spec:AST-015` is a **draft / proposed** system spec. This PR is for review and alignment only; it does not make the mobile theme spacing contract `current` or `accepted`.

## Why

Astryx's mobile theme work currently focuses on typography. This spec proposes how spacing should be added to that mobile theme work, rather than introducing a separate "Mobile Spacing" theme or standalone API.

The existing spacing preset is intentionally broad: it changes the primitive scale as `--spacing-N = round(base × N)` across S / M / L / XL. That works for general density, but it is too blunt for mobile layout because the same primitive spacing values are used by both structural layout and component internals.

If mobile globally compresses `--spacing-*`, it can unintentionally change Button, Input, Badge, SegmentedControl, menu, icon, table, and other component geometry.

The problem this spec tries to solve is narrower:

> Add spacing behavior to the mobile theme so page and container spacing feels appropriate on small viewports without changing every component that happens to use spacing primitives.

## Proposed contract

Before:

- Mobile theme had no explicit spacing contract beyond typography.
- Mobile spacing options were unclear: primitive scale compression, semantic tokens, or one-off component changes.
- Global primitive spacing changes risked shrinking component internals.
- A large semantic token set risked becoming hard to reason about and too product-specific.

After this draft spec:

- Primitive `--spacing-*` tokens remain stable.
- Spacing is added as part of the mobile theme and is **screen-size / viewport driven**, not touch or pointer-modality driven.
- The primary implementation path is a coordinated mobile spacing profile applied through targeted component theme overrides.
- The initial scope is intentionally small: Layout/page shell, Section, Card/Panel-like surfaces, Dialog, Drawer, and Bottom Sheet.
- Explicit component props continue to win over mobile theme defaults.
- Global semantic spacing tokens are treated as exceptional, not the default solution.

## Formula / mobile theme spacing profile

The current recommendation uses a gentle linear adjustment:

```txt
mobileValue = currentValue - (spacingBase × intensity)
```

Where:

- `currentValue` is the component's normal desktop/tablet spacing default.
- `spacingBase` is the selected Astryx spacing preset unit.
- `intensity` controls how strongly mobile compresses the included component defaults.

Recommended preset intensities:

| Preset | Base | Recommended intensity | Meaning |
| --- | ---: | ---: | --- |
| S / Compact | 2px | 0 | Already compact; do not compress further. |
| M / Default | 4px | 0.5 | Slightly tighter without making default feel like compact. |
| L / Comfortable | 6px | 1 | Reduce by one source step. |
| XL / Gigantic | 8px | 1 | Reduce by one source step. |

Optional intensity landmarks:

| Intensity | Value | Example with 8px base |
| --- | ---: | ---: |
| None | 0 | 0px reduction |
| Gentle | 0.5 | 4px reduction |
| Standard | 1 | 8px reduction |
| Dense | 1.5 | 12px reduction |

The formula is a design and authoring helper. Shipped themes should still output explicit component override values so authors can inspect, diff, and customize the final result.

## Component override model

Conceptually, this should live inside the mobile theme API that is being developed separately:

```ts
const mobileSpacing = createMobileSpacingProfile({
  spacingBase: 4,
  intensity: 0.5,
});

const theme = {
  mobile: {
    spacing: mobileSpacing,
    components: {
      Layout: {
        content: {
          paddingInline: mobileSpacing.pagePaddingInline,
          paddingBlock: mobileSpacing.pagePaddingBlock,
        },
      },
      Section: {
        padding: mobileSpacing.sectionInset,
        gap: mobileSpacing.sectionGap,
      },
      Card: {
        padding: mobileSpacing.containerInset,
      },
      Dialog: {
        padding: mobileSpacing.overlayInset,
      },
    },
  },
};
```

The exact API names are illustrative and should align with the mobile theme PR. The important contract is that the mobile theme updates selected component defaults instead of globally changing `--spacing-*`.

## Override precedence

Mobile spacing applies to component defaults only. Explicit author intent stays authoritative.

```txt
explicit component prop
→ mobile-scoped component theme override
→ regular component theme override
→ normal component default
```

Example:

```tsx
// Uses the mobile theme's Section default on mobile.
<Section>...</Section>

// Does not use the mobile theme's Section padding default,
// because the product intentionally authored a value.
<Section padding={6}>...</Section>
```

The same principle applies to `Card padding={...}`, `Dialog padding={...}`, `LayoutContent paddingInline={...}`, and authored layout gaps such as `<VStack gap={...}>`.

## Guidance for semantic tokens

For most mobile spacing adjustments, prefer tuning the coordinated mobile theme spacing profile or component theme overrides before adding new global spacing tokens.

A spacing value should become a shared global semantic token only if it:

- describes a reusable layout relationship, not one component's private implementation detail;
- applies across multiple component families, templates, or product surfaces;
- needs to be tuned independently by theme authors;
- cannot be handled clearly by an existing component prop or component theme override; and
- has a clear scope boundary so it does not become a catch-all spacing token.

This is meant to avoid an endless set of hyper-specific semantic spacing tokens while still leaving room for truly shared layout relationships.

## Representative Product Detail comparison

The spec uses Product Detail as the first validation example because it combines structural layout, container surfaces, and component internals.

| Model | Product Detail behavior | What changes | What stays stable | Risk |
| Primitive spacing override | Every `--spacing-*` value used by the template becomes smaller on mobile. | Page padding, section gaps, card insets, controls, thumbnails, and component internals. | Nothing is guaranteed to stay stable unless each component protects itself. | Too broad; mobile theme can accidentally change component geometry. |
| Large semantic token set | The template maps many page and component areas to public semantic tokens. | Precise areas can be tuned independently. | Component internals can stay stable if they are excluded. | Token sprawl; easy to create hyper-specific tokens that are hard to reason about. |
| Component-owned adaptation | Each component decides whether its own defaults should change at mobile breakpoints. | Component defaults for components that implement mobile behavior. | Other components and explicit props. | Page rhythm can become inconsistent because each component makes a local decision. |
| Shared mobile profile + component overrides | Layout/container defaults receive coordinated mobile values through the theme. | Page padding, section gap, card/panel inset, overlay inset, and owned container-region gaps. | Primitive spacing, control internals, image geometry, and explicit props. | Best balance for v1; needs clear rules for which components are included. |

This comparison is why the proposed direction is a shared spacing profile inside the mobile theme, applied through component theme overrides. It gives Product Detail the structural mobile adjustments it needs without redefining primitives or creating a global token for every local spacing decision.

## Scope

This PR adds only:

- `docs/specs/AST-015/spec.md` — draft system spec for adding a spacing profile to the mobile theme.

It does not:

- implement runtime theme behavior;
- change Core component source;
- change any published theme package;
- add a Changeset;
- update consumer docs;
- promote the spec to `current` / `accepted`.

## Risk

Documentation/specification only. Runtime behavior does not change in this PR.

The main review risk is semantic/API direction: if reviewers disagree with adding a mobile theme spacing profile plus component theme overrides as the primary path, the draft should change before implementation.

## Testing

- `pnpm check:knowledge`
- Pre-commit checks passed:
  - `pnpm check:sync`
  - `pnpm check:knowledge`
  - `pnpm check:changesets`
  - `pnpm check:demo-media`
  - `pnpm check:executable-bits`
  - `pnpm check:cli-structure`
  - `pnpm check:use-client`
  - `pnpm check:i18n-catalog`
  - `pnpm check:cldr-weekdays`
  - `pnpm check:fixtures`

### DIFF
--- a/docs/specs/AST-015/spec.md
+++ b/docs/specs/AST-015/spec.md
@@ -22,11 +22,14 @@ affects_contributing: []
 affects_consumer_docs: [Theme, Layout, Section, Card, Dialog]
 ---
 
-# Mobile spacing theme profile
+# Spacing profile for the mobile theme
 
 ## Intent
 
-Astryx's mobile theme should make structural spacing feel appropriate on small
+This spec defines the spacing portion of Astryx's mobile theme. It is not a
+separate "Mobile Spacing" theme or standalone API.
+
+The mobile theme should make structural spacing feel appropriate on small
 viewports without changing the internal geometry of every component that happens
 to use a spacing token.
 
@@ -39,8 +42,8 @@ The proposed direction is:
 4. Let products tune the profile intensity or override specific component
    defaults when needed.
 
-This gives teams a mobile spacing solution without creating an endless set of
-global semantic spacing tokens.
+This extends the mobile theme with a spacing solution without creating an
+endless set of global semantic spacing tokens.
 
 ## Problem
 
@@ -118,7 +121,7 @@ structure and containers, while leaving most component internals alone.
 ## Proposed model
 
 The mobile spacing profile is a coordinated set of recommended values. The
-profile is not a replacement primitive scale.
+profile is part of the mobile theme and is not a replacement primitive scale.
 
 Conceptually:
 
@@ -132,31 +135,34 @@ const mobileSpacing = createMobileSpacingProfile({
 The profile can then feed component theme overrides:
 
 ```ts
-const mobileTheme = {
-  components: {
-    Layout: {
-      content: {
-        paddingInline: mobileSpacing.pagePaddingInline,
-        paddingBlock: mobileSpacing.pagePaddingBlock,
+const theme = {
+  mobile: {
+    spacing: mobileSpacing,
+    components: {
+      Layout: {
+        content: {
+          paddingInline: mobileSpacing.pagePaddingInline,
+          paddingBlock: mobileSpacing.pagePaddingBlock,
+        },
       },
-    },
-    Section: {
-      padding: mobileSpacing.sectionInset,
-      gap: mobileSpacing.sectionGap,
-    },
-    Card: {
-      padding: mobileSpacing.containerInset,
-    },
-    Dialog: {
-      padding: mobileSpacing.overlayInset,
+      Section: {
+        padding: mobileSpacing.sectionInset,
+        gap: mobileSpacing.sectionGap,
+      },
+      Card: {
+        padding: mobileSpacing.containerInset,
+      },
+      Dialog: {
+        padding: mobileSpacing.overlayInset,
+      },
     },
   },
 };
 ```
 
-The exact API names are illustrative. The important contract is that the mobile
-theme updates selected component defaults instead of globally changing
-`--spacing-*`.
+The exact API names are illustrative and should align with the mobile theme API
+being implemented separately. The important contract is that the mobile theme
+updates selected component defaults instead of globally changing `--spacing-*`.
 
 ## Override precedence
 
@@ -252,10 +258,13 @@ mobile defaults for selected component theme properties.
 Most teams should be able to opt into the default mobile profile:
 
 ```ts
-const mobileTheme = createTheme({
-  mobileSpacing: {
-    preset: 'M',
-    intensity: 'recommended',
+const theme = createTheme({
+  mobile: {
+    typography: '...',
+    spacing: {
+      preset: 'M',
+      intensity: 'recommended',
+    },
   },
 });
 ```
@@ -269,10 +278,13 @@ If the whole mobile layout should feel more or less compact, adjust the
 intensity:
 
 ```ts
-const mobileTheme = createTheme({
-  mobileSpacing: {
-    preset: 'M',
-    intensity: 1,
+const theme = createTheme({
+  mobile: {
+    typography: '...',
+    spacing: {
+      preset: 'M',
+      intensity: 1,
+    },
   },
 });
 ```
@@ -285,13 +297,16 @@ If the recommended profile mostly works but one component needs to diverge,
 override that component's mobile default:
 
 ```ts
-const mobileTheme = createTheme({
-  mobileSpacing: {
-    preset: 'M',
-    intensity: 'recommended',
-    components: {
-      Card: {
-        padding: '16px',
+const theme = createTheme({
+  mobile: {
+    typography: '...',
+    spacing: {
+      preset: 'M',
+      intensity: 'recommended',
+      components: {
+        Card: {
+          padding: '16px',
+        },
       },
     },
   },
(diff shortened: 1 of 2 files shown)


docs/specs/AST-015/spec.md
```diff
@@ -0,0 +1,292 @@
...
+repurposed into new derived-variable behavior when doing so could change current
+theme output.
+
+Rejected: converting compatibility-sensitive existing properties into semantic
+mobile spacing entry points because that can make existing themes behave
+differently without an explicit opt-in.
+
+## Open questions
+
+- **OQ1 — Which semantic spacing roles belong in v1?** (`human-api`) Candidate
```
cixzhang: We had this problem internally where we ended up with so many semantic spacing tokens that were hard to reason about and hyper specific. I'd like to avoid it from growing too large and we should state clear bounds for what semantic spacing we should accept. I should write this spec: what we accept into the global token set and what we exclude (notably, component specific tokens)

cixzhang:
Promising direction, but I think the spec needs to define the ownership boundary through a real application before selecting semantic roles. The problem to solve is differentiated gaps and spacing across responsive breakpoints; component identity alone is not a stable boundary because components such as Card and Dialog can serve both embedded-content and page-region roles.

Could the spec compare these models on one representative responsive layout?

1. **Semantic spacing:** either container-family padding plus relational gap tokens, or the narrower version of relational gaps only, with component overrides handling non-coordinated geometry.
2. **Higher-order theme policy:** a named responsive density/layout concept privately coordinates selected tokens and component overrides.
3. **Component-owned adaptation:** components adjust their own internal padding and gaps at shared breakpoints where the correct behavior is universal.

The example should show which relationships need cross-component coordination, which values remain component-owned, and where theme authors genuinely need control. That would demonstrate whether a shared semantic layer is warranted and how broad it should be, rather than assuming the desired token taxonomy up front.

rubyycheung:
Thanks Cindy — agreed with this framing. I updated the draft so it no longer starts from a broad semantic-token taxonomy. The spec now compares the options against Product Detail and proposes a shared spacing profile inside the mobile theme, applied through targeted component theme overrides.\n\nThe updated direction is:\n\n- keep primitive `--spacing-*` stable;\n- apply mobile spacing only to selected layout/container component defaults;\n- keep explicit consumer props authoritative;\n- use intensity/preset defaults for coordinated values; and\n- treat new global semantic spacing tokens as exceptional, only when the spacing relationship is shared across multiple component families and cannot be handled cleanly through component theme overrides.\n\nThis should make the ownership boundary clearer and reduce the risk of token sprawl.

docs/specs/AST-015/spec.md
rubyycheung: Agreed — I updated the spec to make this softer and more bounded. The primary path is now a mobile theme spacing profile applied through component theme overrides, not a growing global semantic token set. I also added token admission criteria: a new global semantic spacing token should only be added when the spacing relationship is reusable across multiple component families, needs independent theme control, and cannot be handled clearly through a component prop or component theme override.

cixzhang (changes requested):
AST-012 now settles the mechanism this draft leaves open. Please revise AST-015 to use `defineTheme({adaptations: {rules: [{when: {width: {below: <named point>}}, value: {components: ...}}]}})` as the only activation and delivery path. Remove the competing illustrative `mobile.spacing`, separate mobile-theme/package, responsive-theme, and profile-enrollment API shapes; remove OQ3 and any mechanism questions AST-012 already answers. Preserve AST-012 rule order and component-leaf precedence, with explicit component props remaining authoritative. AST-015 should then own only the mobile spacing policy: exact breakpoint, participating component defaults, generated explicit values/intensity, exclusions, and rendered evidence.

PR #2464, opened 2026-06-01
fix(AppShell): apply navAreaStyle to header wrapper in all height modes

ernestt:
## Problem

The top nav and side nav render with subtly different background colors in the `elevated` variant (the default).

## Root Cause

The sidenav panel always receives `navAreaStyle` (explicit `--color-background-body`):
```tsx
<XDSLayoutPanel xstyle={[navAreaStyle, ...]}> // always applied
```

But the header wrapper only received a background in **auto** height mode:
```tsx
stylex.props(isAuto && styles.headerSticky, isAuto && stickyBgStyle)
//           ^^^^^^                          ^^^^^^ — only in auto mode
```

In the default **fill** mode, the header wrapper was transparent — relying on the root shell's background showing through. The sidenav painted its own layer explicitly. This mismatch caused the browser to composite them differently (sub-pixel rendering, layer boundaries from `isolation: isolate` on the elevated content backdrop), producing a visible color seam.

Introduced across PR #597 (`d0d8fca`) which added `navAreaStyle` to the panel, and PR #615 (`d0fda31`) which only added it to the header in auto mode for sticky scroll opacity.

## Fix

Apply `navAreaStyle` to the header wrapper unconditionally:

```diff
-stylex.props(isAuto && styles.headerSticky, isAuto && stickyBgStyle)
+stylex.props(navAreaStyle, isAuto && styles.headerSticky)
```

Both nav areas now paint their own background layer identically in all height modes.

## Testing

- ✅ All 33 AppShell unit tests pass
- ✅ TypeScript typecheck clean
- ✅ SYNC + package boundary checks pass

PR #5873, opened 2026-09-02
fix(AppShell): keep auto section header solid

cixzhang:
## Why

On mobile, `AppShell` with `height="auto"` and `variant="section"` renders a sticky top bar. The header inherited its background from the shell, so content scrolling underneath showed through it.

This regressed in #2464 when the explicit auto-height fallback paint was replaced by the normal nav-area style. `section` intentionally has no normal nav-area paint.

## What

- Restore an explicit surface background only for the `section + auto` header path.
- Add a normal AppShell unit test that asserts the rendered computed background is opaque in auto mode and remains inherited in fill mode.

## Risk

Low. Other variants and fill-height mode keep their existing styling.

## Testing

- `pnpm vitest run packages/core/src/AppShell/AppShell.test.tsx --project ui`
- `pnpm check:changesets`
- pre-commit repo checks

### DIFF
--- a/.changeset/appshell-auto-header-background.md
+++ b/.changeset/appshell-auto-header-background.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Keep AppShell's section top bar solid in auto-height mode while content scrolls beneath it.
+
+@cixzhang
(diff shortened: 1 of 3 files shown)


cixzhang:
/accept-visual 33585459889/1 The section variant intentionally paints its surface on the auto-height sticky header so scrolling content cannot show through; other variants and fill mode are unchanged.

PR #5879, opened 2026-09-02
fix(ci): resolve package for new PR stories

cixzhang:
## Why

Trusted `workflow_run` jobs inspect PR-built Storybook artifacts from a default-branch checkout. When a PR adds a new story without component metadata, its source file is not present on `main`, so package resolution failed and made the PR Comment workflow red even though CI had built the story successfully.

## What

Use an unambiguous package-scoped Storybook title as the fallback when the story source is unavailable, while continuing to reject titles that do not map to a known workspace package.

## Risk

Low. Existing component-path and import-based ownership remain preferred. The fallback only accepts package names already present in the trusted workspace catalog.

## Testing

- `npx --yes pnpm@11.10.0 exec vitest run .github/scripts/visual-gate/lib/plan.test.mjs`
- Re-ran `readStoryIndex` against the exact failed `storybook-fc8c184` artifact in a trusted-main checkout layout: all 1,865 stories resolved, including the PR-only Spinner story as `@astryxdesign/core`

### DIFF
--- a/.github/scripts/visual-gate/lib/plan.mjs
+++ b/.github/scripts/visual-gate/lib/plan.mjs
@@ -102,22 +102,30 @@ function titlePackage(title, catalog, candidates = null) {
 
 function storyPackageNames(entry, storybookDir, repoRoot, catalog) {
   const fromComponent = packageFromComponentPath(entry.componentPath);
+  const titleOwner = titlePackage(entry.title, catalog);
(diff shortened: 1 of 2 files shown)


cixzhang (commented):
Semantic verdict: approve

The parent fails on the PR-only Spinner story; this head resolves all 1,865 artifact stories and still rejects unknown package titles.

[Reviewed by Robohands]

PR #5881, opened 2026-09-02
refactor(core): share interaction modality globally

cixzhang:
## Why

Input modality describes the document, but every consumer currently initializes the same global listeners from its own React effect. This couples page-wide state to component lifecycles and repeats setup work throughout the tree.

## What

- keep the module side-effect free and store shared modality state on the document
- let consumers activate the tracker through one internal hook
- install one capture-phase pointer/keyboard listener pair for the document and keep it current between consumer lifecycles
- remove duplicated initialization effects from every current consumer
- cover duplicate package instances, consumer-free gaps, and the server-rendering fallback

## Risk

No public API or intended interaction behavior changes. The tracker still defaults to keyboard and preserves the existing modifier-key policy.

## Testing

- focused modality, RTL coverage, and affected component tests
- core TypeScript check
- focused ESLint and repository commit checks
- public repository guard

### DIFF
--- a/packages/core/src/utils/interactionModality.ts
+++ b/packages/core/src/utils/interactionModality.ts
@@ -1,10 +1,14 @@
 // Copyright (c) Meta Platforms, Inc. and affiliates.
 
+'use client';
+
 /**
  * @file interactionModality.ts
  * @input Global pointerdown/keydown events
- * @output `getInteractionModality()` — how the user last interacted
- * @position Internal utility; used where `:focus-visible` alone is too broad
+ * @output `useInteractionModalityTracking()` activates document-lifetime tracking;
+ *   `getInteractionModality()` reads how the user last interacted
+ * @position Internal document-wide singleton; used where `:focus-visible` alone
+ *   is too broad
  *
  * `:focus-visible` is the right selector for a focus ring and should stay the
  * condition that draws one. It is not, on its own, "focused by keyboard":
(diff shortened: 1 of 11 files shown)


packages/core/src/utils/interactionModality.ts
```diff
@@ -19,56 +20,90 @@
...
 
 /** Test-only: restore the initial state between cases. */
 export function __resetInteractionModalityForTest(): void {
-  modality = 'keyboard';
+  if (typeof document !== 'undefined') {
+    getStore(document).modality = 'keyboard';
+  }
 }
+
+trackInteractionModality();
```
cixzhang: Not great to have a side effect on module load

cixzhang: Addressed: the module is now side-effect free. Consumers acquire a shared lifecycle registration; the first mounts the document listeners and the last removes them.

cixzhang (commented):

A keyboard user can lose the visible focus ring after navigating through a screen with no tracked component. I reproduced pointer → unmount → Tab → remount: this head keeps stale pointer modality, while the parent records Tab and shows the ring. Please preserve last-input tracking across consumer-free gaps and add that transition to the regression test.

[Reviewed by Robohands]

cixzhang (commented):
Semantic verdict: approve

Thanks, this preserves last-input state through consumer-free gaps. I re-ran pointer → unmount → Tab → remount in Chromium: the focus ring returns.

[Reviewed by Robohands]

PR #5882, opened 2026-09-02
spec(AST-016): define theme family build contract

cixzhang:
## Summary

Drafts **AST-016**, the contract for `astryx theme build <files...> --family`.

- rewrites the proposal under the plain-language guidance in [#5945](https://github.com/facebook/astryx/pull/5945) without changing its contract or deciding OQ1
- renumbers the proposal from AST-012 to AST-016 so it no longer collides with the older responsive-theme proposal
- follows the CLI contribution model in [#5861](https://github.com/facebook/astryx/pull/5861): `--family` narrows existing `theme build` work, every flag pair is closed, JSON is one typed envelope, human output is its projection, and failures use stable codes
- chooses one load-order-independent CSS bundle with complete per-member JS/TS artifacts
- declares the all-member loading cost and exact byte receipts
- defines an exact serializable icon/indicator grammar instead of promising every runtime JavaScript shape
- adds a durable base-owned manifest for atomic cleanup and whole-family `--check`

## Plain-language rewrite

Dense requirement blocks now use short rule titles, direct sentences, a scannable registry table, and an explicit three-condition cleanup list. Repeated contract prose was removed where one canonical statement was enough; schemas, decisions, and evidence still repeat fields only where their structure requires it.

The record is 310 lines, above the soft 200-line aim. It preserves 212 independently checked commitments, including the exact flag matrix, build and check schemas, measurements, verification evidence, three proposed decisions, and the open owner question. Cutting it to 200 lines would require hiding or weakening that contract, so this rewrite favors scanability over the line count.

## Measured size evidence

A reproducible model using the exact [#5687](https://github.com/facebook/astryx/pull/5687) compiler produced:

| Total themes | Standalone CSS | Modeled family bundle | Savings |
| ---: | ---: | ---: | ---: |
| 2 | 32,112 B | 31,046 B | 3.3% |
| 5 | 80,126 B | 41,832 B | 47.8% |
| 14 | 224,354 B | 72,202 B | 67.8% |

The proposal uses a measured byte gate rather than a fixed member-count cap. A product that loads only one theme keeps the standalone build.

## CLI matrix

- `--check`: composes and verifies the family-shaped output
- `--icons-specifier`: composes for the supported imported icon-registry form
- `--watch`: refused before loading or writes
- `--out`: cannot co-occur under the existing multi-file rule
- global `--json`: one `theme.build.family` or `theme.build.family.check` envelope

## Owner review

One question remains: approve DEC-1 through DEC-3 as the family-build contract?

No runtime or package behavior changes in this PR. Implementation [#5687](https://github.com/facebook/astryx/pull/5687) must be rewritten and revalidated against the accepted contract.

## Validation

Passed at exact head `088d4dece7efd08e84c06ee7af29f5c2223b9212`:

- Prettier
- 74 focused knowledge tests
- `pnpm check:repo`
- `git diff --check` and public-link/internal-context hygiene
- condition-blind semantic parity and independent exact-head review

### DIFF
--- a/docs/specs/AST-016/spec.md
+++ b/docs/specs/AST-016/spec.md
@@ -1,0 +1,310 @@
+---
+schema_version: 1
+template_version: 1
+kind: system-spec
+id: spec:AST-016
+authority: draft
+archive_reason: null
+superseded_by: null
+approved_by: null
+approved_at: null
+phase: proposed
+owners: [cixzhang, imdreamrunner, josephfarina]
+affects_architecture:
+  [
+    architecture:theme-authoring-contract,
+    architecture:theme-compilation,
+    architecture:theme-application,
+  ]
+affects_families: []
+affects_contributing: []
+affects_consumer_docs: [theme-build]
+---
+
+# Theme family build
+
+## Intent
+
+Products that ship and load a whole theme family should not repeat the base
+rules in every member stylesheet. `astryx theme build <files...> --family`
+builds one load-order-independent CSS bundle. It also keeps every member's
+JavaScript and TypeScript artifacts complete.
+
+Family mode is an explicit deployment choice. Using any member loads the whole
+family bundle. A product that loads only one theme should use the existing
+standalone build instead.
+
+## Non-goals
+
+- Changing `defineTheme` precedence, merge rules, or flattened standalone semantics.
+- Changing ordinary single-file or multi-file batch builds.
+- Adding another command, authoring field, helper, or portable token vocabulary.
+- Producing per-member delta stylesheets that consumers load in a particular order.
+- Serializing every JavaScript expression that `defineTheme` accepts at runtime.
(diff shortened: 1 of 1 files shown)


cixzhang (commented):

Thanks for making the family-build tradeoff concrete. Two contract gaps still leave builders exposed:

- The goal is fewer render-blocking bytes, but the verification table has no size or loading invariant. A one-member consumer may now load every member’s delta. Please measure the proposed bundle at realistic family sizes and state whether family users load all members, plus the acceptable growth rule.
- FR5 promises every registry form accepted by `defineTheme`, then rejects valid local, spread, and non-identifier forms and omits inline icons. Please define one exact build-time source grammar and prove every supported icon/indicator form is preserved or rejected before writes.

Can you lock those two boundaries before this becomes current?

[Reviewed by Robohands]

cixzhang:

PR #5884, opened 2026-09-02
fix(Selector): align selected item after layer flips

### LINKED ISSUE #5835: [Bug] [Selector] Menu is positioned twice — JS margin fallback applies on top of CSS anchor positioning
### Description

# [Selector] Menu is positioned twice — JS margin fallback applies on top of CSS anchor positioning

## Summary

`Selector`'s popover carries both CSS anchor positioning and a JS-measured `margin-block-start`, inline, on the same element. The margin looks like a fallback for browsers without anchor positioning. In a browser that supports anchor positioning both apply, so the menu is placed twice and lands far from its trigger — in our case at the top of the window.

`Typeahead` is not affected: its popover carries no such margin and positions correctly. That difference is what suggests the margin is unintended here rather than load-bearing.

## Environment

- `@astryxdesign/core` **0.5.0**, also reproduced unchanged on **0.5.2**
- `@astryxdesign/theme-neutral` 0.5.0 / 0.5.2
- React 19.2.8, react-dom 19.2.8, Next.js 16.3.3 (App Router)
- Chromium 151.0.7922.34
- `CSS.supports("anchor-name: --x")` → `true`; `CSS.supports("position-area: block-end")` → `true`

## What the popover carries

Read off the open popover element:

```
position-anchor: --astryx-layer-_r_12_;
position-area: self-block-end span-self-inline-end;
position-try-fallbacks: flip-block, flip-inline, flip-block flip-inline;
margin-block-start: -391.391px;      /* set inline by JS */
```

Computed on the same element: `position: fixed`, `inset: 0`, `margin-top: -391.391px`.

## Steps to reproduce

1. Render a `Selector` low in a tall, scrollable panel — far enough down the page that its anchor is a few hundred pixels from the top of the viewport.
2. Click it to open the menu.
3. Compare the menu's bounding box to the trigger's.

## Expected

The menu opens adjacent to its trigger — below it, or flipped above where there is no room, which `position-try-fallbacks` already handles.

## Actual

The menu is displaced by roughly its own height plus the anchor's offset. Measured with the anchor at `y=365` (height 20) in a 1280×900 viewport:

| Selector | `margin-block-start` | menu top |
|---|---|---|
| anchored at y=365 | −391.391px | **0** |
| anchored higher in the same panel | −69px | displaced by ~69px |
| `Typeahead` in the same panel | *none* | 395 — correct |

The first open is the worst of it: the menu has not been laid out yet, so the offset appears to be measured against a height it does not have. Subsequent opens land closer but still wrong (−301px on the second open in the same session).

## Note on CSS containment

An ancestor with `container-type: inline-size` makes the displacement worse — that element becomes the containing block for fixed-position descendants, which moves the reference the JS offset is measured against. Removing it moved the menu from `y=0` to `y=90`, so it amplifies the problem but is not the cause. The negative margin is present either way.

## Workaround

Neutralising the margin only where anchor positioning can do the job:

```css
@supports (position-area: block-end) {
  [popover]:has(.astryx-selector-popup) {
    margin-block-start: 0 !important;
  }
}
```

With this the menu lands at `y=391` against a trigger ending at `y=385`, and flips cleanly above (`y=59..359` against a trigger at `y=365`) when the viewport is too short — consistently, including on first open.

The `@supports` guard matters: Safari and Firefox have no anchor positioning yet, and there the JS margin is doing the whole job, so removing it unconditionally would break exactly the browsers it exists for.

## Suggested fix

Skip the JS `margin-block-start` when anchor positioning is available — e.g. gate it on `CSS.supports("position-area: block-end")` at the point where the margin is written, so only one positioning system is ever in effect.

### Reproduction

```markdown
1. Render a `Selector` low in a tall, scrollable panel — far enough down the page that its anchor is a few hundred pixels from the top of the viewport[cite: 1].
2. Click it to open the menu[cite: 1].
3. Compare the menu's bounding box to the trigger's[cite: 1].

Note on CSS containment:
An ancestor with `container-type: inline-size` makes the displacement worse — that element becomes the containing block for fixed-position descendants, which moves the reference the JS offset is measured against[cite: 1]. Removing it moved the menu from `y=0` to `y=90`, so it amplifies the problem but is not the cause[cite: 1]. The negative margin is present either way[cite: 1].

### Astryx Version

@astryxdesign/core@0.5.0 (also reproduced on 0.5.2)[cite: 1]

### Environment

- React 19.2.8, react-dom 19.2.8, Next.js 16.3.3 (App Router)[cite: 1]
- Chromium 151.0.7922.34[cite: 1]
- `CSS.supports("anchor-name: --x")` → `true`; `CSS.supports("position-area: block-end")` → `true`[cite: 1]

### PR DESCRIPTION
**Pull request by cixzhang**
## Why

Selector's selected-item overlay correction assumed CSS anchor positioning had placed the menu below its trigger. When `position-try-fallbacks` flipped the menu above, the negative margin moved an already-flipped menu again and could pin or clip it at the viewport edge.

## What

Reuse the existing hidden layout measurement to read the popover's resolved, untransformed `offsetTop`, then translate from that actual position to the viewport-clamped selected-item alignment. `translate` stays outside anchor fallback fit calculations, so applying the correction cannot make the browser choose a different side.

`useSelectedItemOffset()` remains the one public composable foundation. It preserves the released non-negative `offset` and `isPositioned`, and additively returns signed `translateY` for flip-stable CSS-translate composition. The hook now has consumer-facing CLI docs and an explicit current contract.

Adds a regression for the flipped geometry, an open Storybook example with the trigger near the viewport bottom, and RTL coverage for Selector's logical option/indicator order.

## Risk

Scoped to the default non-search selected-item overlay. Existing hook consumers keep the released `offset` name, sign, and meaning. Search menus and explicit `placement` continue through normal Layer positioning unchanged.

## Testing

- `pnpm exec vitest run packages/core/src/Selector/Selector.test.tsx` (194 tests)
- targeted ESLint with zero warnings
- knowledge, SYNC, and Changeset checks
- `astryx hook useSelectedItemOffset --dense` resolves the public hook and all three return fields
- Chromium at 800×720: resolved layout top 500 translated to viewport-safe 567–703; the selected row center matched the trigger center within the existing 1px optical offset
- Chromium unchanged path: below placement translated from layout top 398 to 297–433 with the same 1px optical alignment
- RTL audit: Selector option/indicator order flips correctly with no coverage gap

Fixes #5835

### DIFF
--- a/packages/core/src/Selector/hooks.ts
+++ b/packages/core/src/Selector/hooks.ts
@@ -4,10 +4,15 @@
 
 /**
  * @file hooks.ts
- * @input Uses untransformed DOM layout geometry from the Selector's outer
- *   anchor, listbox, and selected option
+ * @input Uses resolved Layer placement plus untransformed DOM layout geometry
+ *   from the Selector's outer anchor, listbox, and selected option
  * @output Hooks for Selector
- * @position Internal hooks; used by Selector.tsx
+ * @position Internal module; exports public Selector hooks through index.ts
+ *
+ * SYNC: When modifying useSelectedItemOffset, update:
+ * - /packages/core/src/Selector/useSelectedItemOffset.doc.mjs
+ * - /packages/core/src/Selector/Selector.spec.md
+ * - /packages/core/src/Selector/Selector.test.tsx
  */
 
 import {useCallback, useState} from 'react';
@@ -47,19 +52,36 @@ interface UseSelectedItemOffsetOptions {
 }
 
 interface UseSelectedItemOffsetResult {
+  /**
+   * Legacy positive distance from the below-anchor origin. Apply as a negative
+   * block-start margin when composing the original positioning behavior.
+   */
   offset: number;
+  /**
+   * Signed correction from the browser-resolved layer top to the clamped target.
+   * Apply with CSS `translate` so anchor fallback selection remains stable.
+   */
+  translateY: number;
   isPositioned: boolean;
 }
 
 /**
- * Calculates the offset needed to position the dropdown so that the selected
- * item appears centered over the outer selector control (macOS-style selector).
+ * Calculates the translation needed to position the dropdown so that the
+ * selected item appears centered over the outer selector control (macOS-style
+ * selector).
  *
  * The desired dropdown top is calculated directly from the anchor center and
  * selected-item center, then clamped to the viewport. This preserves the
  * default "selected item over trigger" behavior while letting the menu slide
  * upward near the bottom edge or downward near the top edge instead of being
  * clipped off-screen.
+ *
+ * CSS anchor positioning may place the layer above or below before this layout
+ * effect runs. The popover's `offsetTop` is that resolved, untransformed layout
+ * position, so translating from it to the viewport-clamped target reuses the
+ * existing hidden measurement pass without parsing CSS or adding another pass.
+ * `translate` does not participate in `position-try-fallbacks`, so applying the
+ * correction cannot make the browser choose a different fallback.
  */
 export function useSelectedItemOffset({
   isOpen,
@@ -69,12 +91,15 @@ export function useSelectedItemOffset({
   anchorRef,
 }: UseSelectedItemOffsetOptions): UseSelectedItemOffsetResult {
   const [offset, setOffset] = useState(0);
+  const [translateY, setTranslateY] = useState(0);
   const [isPositioned, setIsPositioned] = useState(false);
 
   const commitPosition = useCallback(
-    (nextOffset: number, nextIsPositioned: boolean) => {
+    (nextOffset: number, nextTranslateY: number, nextIsPositioned: boolean) => {
       // eslint-disable-next-line @eslint-react/set-state-in-effect -- selector popover position is derived from DOM layout
       setOffset(nextOffset);
+      // eslint-disable-next-line @eslint-react/set-state-in-effect -- selector popover position is derived from DOM layout
+      setTranslateY(nextTranslateY);
       // eslint-disable-next-line @eslint-react/set-state-in-effect -- selector popover position is derived from DOM layout
       setIsPositioned(nextIsPositioned);
     },
@@ -83,13 +108,13 @@ export function useSelectedItemOffset({
 
   useIsomorphicLayoutEffect(() => {
     if (!isOpen) {
-      // Reset offset when closed
-      commitPosition(0, false);
+      // Reset translation when closed.
+      commitPosition(0, 0, false);
       return;
     }
 
     if (!listboxRef.current || !anchorRef.current) {
-      commitPosition(0, true);
+      commitPosition(0, 0, true);
       return;
     }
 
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: packages/core/src/Selector/hooks.ts [hunk 1]
```diff
@@ -47,19 +47,36 @@ interface UseSelectedItemOffsetOptions {
 }
 
 interface UseSelectedItemOffsetResult {
-  offset: number;
+  translateY: number;
   isPositioned: boolean;
 }
 
+function isBlockStartPositionArea(positionArea: string): boolean {
+  const [blockPosition] = positionArea.trim().split(/\s+/, 1);
```
cixzhang: Seems fragile. How can we be sure there will be at least 2 items in the split

cixzhang: Good catch. The code only needed to recognize the first position-area token, but split made that contract unclear. I removed the split entirely: it now matches each accepted block-start serialization either as the whole value or as an explicit token prefix, so both one-token and multi-token values are handled.

cixzhang: I took the stronger geometry route from the follow-up: the latest head removes position-area parsing entirely. The existing hidden layout pass now reads the popover’s resolved, untransformed offsetTop and translates from that actual top to the viewport-clamped target. The near-bottom and normal-below Chromium cases both stay in bounds and retain the 1px optical alignment.

cixzhang (commented):

Thanks—the geometry fix works, but existing `@astryxdesign/core/Selector` consumers destructuring `offset` from `useSelectedItemOffset()` stop compiling: `hooks.ts:50` renames the published result to `translateY` under a patch changeset. Please preserve the released return contract or use the explicit breaking-change path.

[Reviewed by Robohands]

cixzhang:
Self-review update: preserved the public `useSelectedItemOffset()` return contract and legacy offset semantics, added a focused regression for that API, rebased onto current `main`, and removed unrelated RTL-target formatting churn. All local code checks and real Chromium geometry checks pass. I moved this back to draft because AST-013 FR10 still requires actual iOS Safari evidence, and neither available Mac has Xcode/an iOS simulator nor does this repo currently have a remote Safari lane.

cixzhang (commented):

The new near-bottom story does not reproduce a user-visible regression: with the parent runtime and at bff833fd, Chromium renders the menu at 567–703px, fully visible, with the selected row 1px from the trigger center. The reporter also withdrew the original repro. Please provide same-case red→green evidence in a real browser, including latest iOS Safari per AST-013 FR10, before changing the shipped positioning mechanism.

[Reviewed by Robohands]

cixzhang:
API direction update: `useSelectedItemOffset` remains the single public foundation. The latest head preserves `offset` and `isPositioned`, additively returns signed `translateY`, and Selector consumes that same public hook. Added the current semantic contract, consumer-facing `astryx hook useSelectedItemOffset` docs, and a regression covering all three return fields. No package-private replacement hook remains.

cixzhang:
Closing this change after rechecking the evidence. The reporter withdrew the repro, and the added near-bottom story renders correctly on the unmodified parent runtime: the menu stays within the viewport and the selected row remains optically aligned. Without a same-case red baseline, replacing the positioning mechanism and adding a new public hook result would be speculative. The branch remains available if a reproducible case returns.

PR #5907, opened 2026-09-02
fix(vibe-tests): hardlink node_modules cross-platform instead of cp -al

HelloOjasMutreja:
## Problem

`linkDependencies()` in `run-setup.mjs` and `preparePair()` in `setup-canonical.test.ts` both shell out to `cp -al` to hardlink a prepared `node_modules` directory into each sandbox, as a faster alternative to a full copy. `cp` is Unix/GNU-coreutils only: on Windows it's missing entirely, and even where a `cp` shim exists (e.g. via Git for Windows), hardlinking through it fails with `Permission denied`, once per `node_modules` entry.

This is a real, reproducible bug for any Windows contributor running `setup:prepare` or the `ASTRYX_CANONICAL_SETUP_BROWSER=1` suite, which only runs automatically in CI on a Linux runner, so it's silent there.

## Investigation

I actually installed a real fixture's `node_modules` on Windows to see what pnpm produces there before picking a fix direction. Two things ruled out the two fallbacks suggested in the issue:

- **`fs.linkSync` alone doesn't work**, because pnpm's Windows `node_modules` layout is junction-heavy: `node_modules/react` etc. are NTFS junctions (`readdirSync` reports these as `isSymbolicLink()`, not `isDirectory()`), and Windows can't hard-link a reparse point (`EPERM`).
- **The existing `copyDirectory()` fallback doesn't work either.** It only branches on `isDirectory()`, so it tries `copyFileSync` on every junction and fails with the same `EPERM`. Confirmed this by running it directly against a real pnpm-installed tree.

## Fix

A pure-JS `linkDirectory()` walker (added identically to both call sites, matching this codebase's existing pattern of duplicating small helpers rather than sharing across `run-setup.mjs`/`setup-canonical.test.ts`) that:
- recurses into real directories,
- hard-links regular files via `fs.linkSync` (still gets the "share bytes, don't duplicate" win `cp -al` was for),
- recreates symlinks/junctions via `fs.symlinkSync`, using `'junction'` on Windows (the one link type Windows creates without elevated privileges), instead of dereferencing them.

Kept scoped to just this hardlinking step, per the report.

## Testing

- Installed the `tailwind-v4-control` fixture's real `node_modules`, ran `linkDirectory` against it into a fresh sandbox, then ran a full `tsc --noEmit && vite build` in that sandbox; it succeeded.
- Ran the actual `setup-canonical.test.ts` with `ASTRYX_CANONICAL_SETUP_BROWSER=1`: it now gets all the way through the `node_modules` linking step (previously failed there first). It still fails downstream at an unrelated, pre-existing `fs.symlinkSync(..., 'dir')` call (needs Windows elevation/Developer Mode), confirmed identical on a clean `upstream/main` checkout with none of this diff applied, so it's out of scope here.
- Ran the full `internal/vibe-tests/setup-test/` suite: 114 passing, 2 pre-existing failures in `public-artifact-cli.test.mjs` (a separate `.bin/tsx` Windows-spawn issue, also confirmed identical on a clean baseline), 8 skipped (the canonical suite, gated off without the env var).
- `eslint` clean on both changed files.

### DIFF
--- a/internal/vibe-tests/setup-test/linkDirectory.mjs
+++ b/internal/vibe-tests/setup-test/linkDirectory.mjs
@@ -1,0 +1,101 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+/**
+ * @file linkDirectory.mjs
+ * @input Node fs/path
+ * @output Exports linkDirectory, retargetLinkTarget
+ * @position Shared by run-setup.mjs and setup-canonical.test.ts; side-effect
+ *   free on import (no top-level fs access), so it can be unit tested
+ *   directly without pulling in either caller's own setup.
+ */
+
+import * as fs from 'node:fs';
+import * as path from 'node:path';
+
+/**
+ * Where an absolute link target should actually point, once it is about to
+ * live inside `destinationRoot` instead of `sourceRoot`.
+ *
+ * A relative target needs nothing: it already resolves against its own
+ * link's location, and `linkDirectory` mirrors `sourceRoot`'s structure
+ * exactly, so the same relative text resolves to the same place inside
+ * `destinationRoot`.
+ *
+ * An absolute target (what a Windows junction always stores, since
+ * `fs.readlinkSync` never returns a relative path for one) is different: if
+ * it resolves to somewhere inside `sourceRoot`, reusing that exact text
+ * would give every sandbox created from the same `depsDir` a junction that
+ * quietly points back at the one shared `depsDir` tree, rather than at the
+ * copy `linkDirectory` is actually building. A write through that junction,
+ * from a sandbox running an agent's own code and not just this script's own
+ * install/build calls, corrupts the shared source and, with it, every
+ * other sandbox built from it. Retargeted here to the equivalent path
+ * inside `destinationRoot`, matching the pattern `setup-workspace.mjs` uses
+ * for the same problem in its own copy step.
+ *
+ * A target that resolves outside `sourceRoot` entirely (an absolute path to
+ * something else on the machine) is left as-is: it was never inside the
+ * tree being copied, so there is nothing to retarget it relative to.
+ */
+export function retargetLinkTarget(target, sourceRoot, destinationRoot) {
+  if (!path.isAbsolute(target)) {
+    return target;
+  }
+  const resolvedTarget = path.resolve(target);
+  const resolvedSource = path.resolve(sourceRoot);
+  const relative = path.relative(resolvedSource, resolvedTarget);
+  const isInsideSource =
+    relative === '' || (!relative.startsWith('..') && !path.isAbsolute(relative));
+  if (!isInsideSource) {
+    return target;
+  }
+  return path.join(path.resolve(destinationRoot), relative);
+}
+
+/**
+ * Cross-platform equivalent of `cp -al`: recurse into real directories,
+ * hard-link regular files (shares bytes with `source` instead of
+ * duplicating them), and recreate symlinks/junctions as links rather than
+ * dereferencing them, retargeted via `retargetLinkTarget` so a copy stays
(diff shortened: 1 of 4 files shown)


cixzhang (changes requested):
Thanks for tackling Windows. One isolation problem still blocks this: `internal/vibe-tests/setup-test/run-setup.mjs:108` recreates each pnpm junction with its original absolute target. A Windows sandbox therefore resolves package directories back into the shared dependency cache; adding or removing a file through one run changes the cache seen by later runs. macOS/Linux relative links stay inside the copied tree, so this is not equivalent behavior.

Please use the retargeting pattern in `setup-workspace.mjs`: links that resolve inside the source tree should point at the corresponding destination, while genuinely external links stay external. Add a regression test that mutates through one cloned junction and proves the source and a sibling sandbox remain unchanged.

[Reviewed by Robohands]

HelloOjasMutreja:
Fixed in 659095c19.

The root cause was as described: `fs.readlinkSync()` on a Windows junction always returns an absolute path, and the old `linkDirectory` reused that raw target when recreating the junction in the new sandbox, so it silently pointed back into the shared `depsDir` instead of the sandbox's own copy.

Added `retargetLinkTarget`, following the same retargeting pattern `setup-workspace.mjs` already uses: an absolute target that resolves inside the source tree gets rewritten to the equivalent path inside the destination tree, and a genuinely external absolute target (or a relative one, which already resolves correctly on its own) is left alone.

Also extracted `linkDirectory` out of `run-setup.mjs` and `setup-canonical.test.ts`, where it had been duplicated with the same bug in both places, into a new shared `linkDirectory.mjs` module both files now import.

Added `linkDirectory.test.mjs` with the regression test requested: it builds two sandboxes from one shared dependency tree (using a real junction on Windows, not a mock), writes a new file through one sandbox's cloned junction, and asserts the shared source and the sibling sandbox are both unaffected while the writing sandbox's own copy has it. Also covers `retargetLinkTarget` directly for the inside-source, outside-source, and relative-target cases.

Verified locally on Windows: `node --check` on both `.mjs` files, `tsc --noEmit` on `setup-canonical.test.ts`, and `vitest run internal/vibe-tests/setup-test --project node`, which passes the new tests (5/5) along with the rest of the suite. Four pre-existing failures in `setup-workspace.test.mjs` remain (EPERM creating non-junction symlinks without elevated privileges on this machine); confirmed those are unrelated to this change by reproducing them with this fix stashed out.

HelloOjasMutreja:
Rebased onto current main. Two real conflicts in setup-canonical.test.ts against main's own Windows cmd.exe routing (runPnpm/spawnPnpm) that landed independently:

1. Both sides added a different helper to the same file (runPnpm/spawnPnpm on main's side, linkDirectory on this PR's side). Kept both, non-overlapping.
2. The second commit here (the follow-up junction-retargeting fix) removes the local linkDirectory in favor of importing it from the new shared linkDirectory.mjs module. My first rebase pass missed that and left a duplicate local definition alongside the import; caught it, removed the stale duplicate.

Verified after rebase: the new linkDirectory.test.mjs suite (5 tests covering the junction-retargeting fix) passes. setup-canonical.test.ts itself is skipped by default (gated behind ASTRYX_CANONICAL_SETUP_BROWSER=1, a much broader real-install integration suite unrelated to this specific fix); ran it with the exact CI invocation (`--project node --maxWorkers=1`) and it reports the same skip CI would see. Typecheck and lint clean.

HelloOjasMutreja:
Retargeting is in via retargetLinkTarget, threaded through the recursive linkDirectory calls via rootSource/rootDestination so a nested junction still resolves against the whole tree. Added the regression test that mutates through a cloned junction and asserts the source and a sibling sandbox stay unchanged. Rebased onto current main.

Issue #5899, opened 2026-09-02
[Bug] StackItem Properties preview renders no working component


cixzhang:
## Affected page

https://astryx.atmeta.com/components/StackItem?tab=properties

## Steps to reproduce

1. Open the affected component page in Chromium 149.0.7827.55 on macOS.
2. Open the **Properties** tab.
3. Observe the preview before changing any props.
4. Exercise the available controls where applicable.

## Expected behavior

A visible item renders inside a Stack so item sizing and alignment can be inspected.

## Actual behavior

The preview stage is completely empty except for the Show code control.

No unrelated console or page errors were observed during the fresh-context reproduction.

## Diagnosis

The subcomponent has neither child content nor the parent Stack scaffold it is designed to participate in.

The likely fix is local to `packages/core/src/Stack/Stack.doc.mjs` and/or its focused docsite preview coverage; it does not require a new public API or design direction. This is an atomic, independently closable slice of #2008.

## Acceptance criteria

- [ ] Add representative child content and a minimal Stack playground wrapper.
- [ ] The preview renders a visible item on first load.
- [ ] Changing an item control such as size or cross-axis alignment produces an observable update.
- [ ] Add focused preview/data-extraction coverage.

cixzhang:
Thanks for reporting the broken StackItem Properties preview. This is fixed on the live docsite by [#5912](https://github.com/facebook/astryx/pull/5912): the Grid/Stack family now supplies the playground defaults needed for StackItem to render on first load. I verified the current public Properties page in Chrome with no missing-props placeholder or page errors. Closing as fixed.

Issue #5893, opened 2026-09-02
[Bug] GridSpan Properties preview renders no working component


cixzhang:
## Affected page

https://astryx.atmeta.com/components/GridSpan?tab=properties

## Steps to reproduce

1. Open the affected component page in Chromium 149.0.7827.55 on macOS.
2. Open the **Properties** tab.
3. Observe the preview before changing any props.
4. Exercise the available controls where applicable.

## Expected behavior

A visible grid item renders inside a Grid so its span controls can be inspected.

## Actual behavior

The preview stage is completely empty except for the Show code control.

No unrelated console or page errors were observed during the fresh-context reproduction.

## Diagnosis

The subcomponent has neither visible child content nor the parent Grid scaffold it is designed to participate in.

The likely fix is local to `packages/core/src/Grid/GridSpan.doc.mjs` and/or its focused docsite preview coverage; it does not require a new public API or design direction. This is an atomic, independently closable slice of #2008.

## Acceptance criteria

- [ ] Add representative child content and a minimal Grid playground wrapper.
- [ ] The preview renders a visible spanned grid item on first load.
- [ ] Changing the column or row span produces an observable update.
- [ ] Add focused preview/data-extraction coverage.

cixzhang:
Thanks for reporting the broken GridSpan Properties preview. This is fixed on the live docsite by [#5912](https://github.com/facebook/astryx/pull/5912): the Grid/Stack family now supplies the playground defaults needed for GridSpan to render on first load. I verified the current public Properties page in Chrome with no missing-props placeholder or page errors. Closing as fixed.

PR #3183, opened 2026-06-26
fix(docsite): seed OverflowList properties-tab preview with items

durvesh1992:
## Summary

Closes #2872.

`OverflowList`'s `children` prop is required, so the docsite properties-tab preview rendered an empty list ("can't render standalone"). Add `playground.defaults.children` (a few Button items) so the interactive preview shows a real, meaningful list.

`packages/core/src/OverflowList/OverflowList.doc.mjs` — added `playground.defaults.children`.

## Testing
- Added a preview-state test: `buildInitialState` resolves the seeded children into 3 elements and `getMissingRequiredProps` is empty. `component-preview-state.test.ts` — 9/9 pass.
- Regenerated the docsite data and confirmed the playground children are present in the generated component registry.

> Note: the properties-tab preview renders client-side, so I verified the fix at the data/logic level (same approach as the Icon/Skeleton preview fix in #3175). A quick visual check on the running docsite is welcome — I couldn't launch a browser in my environment.

## Changeset
Included (`@astryxdesign/core` patch, docs).

PR #3175, opened 2026-06-26
fix(docsite): restore Icon and Skeleton properties-tab previews

durvesh1992:
## Summary

Closes #2848. Closes #2875.

Two components rendered a broken **properties-tab** interactive preview on the docsite. Both are fixed by adding `playground.defaults` to their `.doc.mjs`:

- **Icon (#2848):** `icon` is required and its type (`IconName | ComponentType<SVGProps>`) can't be auto-generated, so the tab showed *"Interactive preview needs required props … Missing: icon"*. Seed a valid semantic name (`icon: 'search'`).
- **Skeleton (#2875):** `width`/`height` default to `'100%'`, which collapses to nothing inside the centered preview. Provide concrete pixel dimensions (`200 × 24`).

## Testing
- Extended `component-preview-state.test.ts` — **10/10 pass**. The Icon test asserts the prop is genuinely *Missing* without the default and satisfied with it; the Skeleton test asserts the playground default overrides the `100%` doc default.
- Re-ran `generate-data.mjs` — both `.doc.mjs` files parse and the registry rebuilds cleanly.

## Changeset
Included (`@astryxdesign/core` patch, docs) since `.doc.mjs` files ship in `core/src`.

PR #5912, opened 2026-09-02
fix(docsite): seed Grid/Stack family Properties previews with playground defaults

### LINKED ISSUE #5892: [Bug] Grid Properties preview renders no working component
## Affected page

https://astryx.atmeta.com/components/Grid?tab=properties

## Steps to reproduce

## Expected behavior

A small multi-cell grid renders so columns, gaps, and alignment controls are observable.

## Actual behavior

The preview stage is completely empty except for the Show code control.

## Diagnosis

The doc has no playground children, so the generated grid contains no cells and collapses visually.

The likely fix is local to `packages/core/src/Grid/Grid.doc.mjs` and/or its focused docsite preview coverage; it does not require a new public API or design direction. This is an atomic, independently closable slice of #2008.

## Acceptance criteria

- [ ] Add a minimal set of representative grid-cell children.
- [ ] The preview renders a recognizable grid on first load.
- [ ] Changing a grid control such as columns or gap produces an observable update.

HelloOjasMutreja:
## Problem

Six components in the Grid/Stack layout family rendered an empty Properties-tab preview:

- Grid, Stack, HStack, VStack: no `playground.defaults`, so the preview had no children.
- GridSpan, StackItem: no child content, and no parent context (Grid / HStack) to participate in.

Each is an atomic, independently closable slice of #2008, filed by the nightly vibe-test bot as #5892, #5893, #5894, #5898, #5899, #5900.

## Fix

Followed the established pattern from prior merged fixes of this exact class (#4138, #2815, #3183, #3175):

- **Grid, Stack, HStack, VStack**: added `playground.defaults.children`, a few `Card` elements, matching the convention in `Card.doc.mjs`/`OverflowList.doc.mjs`.
- **GridSpan, StackItem**: these are sub-components that render nothing meaningful in isolation (a grid item with no grid, a flex item with no flex row). Gave each its own `playground` (overriding the one it would otherwise inherit from its parent doc) using `playground.wrapper` to render it inside a real `Grid`/`HStack`, so the span/fill behavior is actually visible.

## Testing

- Added regression tests to `component-preview-state.test.ts` (Grid, Stack, HStack, VStack, matching the exact style of the existing OverflowList/DropdownMenu tests) and `component-detail-resolve-elements.test.ts` (GridSpan, StackItem, verifying their `playground.wrapper` config resolves to a real component).
- Full `apps/docsite` test suite: 443 passing (17 pre-existing failures unrelated to this change, confirmed identical on a clean baseline).
- `tsc --noEmit`: 27 pre-existing errors, identical count with and without this diff.
- Verified against the real docsite dev server (`pnpm generate && next dev`), not just the source or unit tests: navigated to all six Properties-tab pages and visually confirmed each renders a real, non-empty preview (Grid shows three cards in a 3-column layout, Stack/HStack/VStack each show three cards in their respective direction, GridSpan shows "Spans 2 columns" inside a real Grid, StackItem shows a full-width "Fills the row" card inside a real HStack). No screenshots attached; my environment for this session doesn't have a straightforward way to export the browser tool's captures as attachable image files, so this is a direct description of what was checked rather than a substitute for it.

Fixes #5892, #5893, #5894, #5898, #5899, #5900

### DIFF
--- a/.changeset/grid-stack-properties-preview.md
+++ b/.changeset/grid-stack-properties-preview.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[docs] Grid, Stack, HStack, VStack, GridSpan, and StackItem: seed example content via playground defaults (and, for the two sub-components, a real parent wrapper) so the docsite properties-tab preview renders a working component instead of an empty stage. (#5892, #5893, #5894, #5898, #5899, #5900)
+
+@HelloOjasMutreja
--- a/apps/docsite/src/__tests__/component-detail-resolve-elements.test.ts
+++ b/apps/docsite/src/__tests__/component-detail-resolve-elements.test.ts
@@ -2,10 +2,16 @@
 
 import {isValidElement, type ReactElement} from 'react';
 import {describe, expect, it, vi} from 'vitest';
-import {resolveValue} from '../components/component-detail/resolveElements';
+import {
+  getComponent,
+  resolveValue,
+} from '../components/component-detail/resolveElements';
 
 vi.mock('@astryxdesign/core', () => ({
   Badge: () => null,
+  Card: () => null,
+  Grid: () => null,
+  HStack: () => null,
   Icon: () => null,
   SideNavItem: () => null,
   Text: () => null,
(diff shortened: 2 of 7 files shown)


cixzhang (changes requested):
Thanks, the six previews now render correctly. One thing before merge: the tests duplicate the new fixture values instead of reading the docs. I reverted all six playground configs and every changed test still passed (36/36), so the empty-preview bug can return undetected. Could you assert the extracted playground configs, especially both wrappers, in `data-extraction.test.ts`?

[Reviewed by Robohands]

HelloOjasMutreja:
Good catch, confirmed the exact gap: my tests exercised the playground mechanism with hand-typed fixture values, not the actual persisted `.doc.mjs` config. Added assertions against `componentRegistry` (built from the real doc source via `pnpm generate`) for all six components, including both wrappers (GridSpan into Grid, StackItem into HStack).

Verified these actually catch a revert before pushing: reverted the three `.doc.mjs` files, regenerated, confirmed all six new tests fail with the expected "received undefined" errors, then restored the real changes and confirmed they pass again.

cixzhang (approved):
Thanks, this now protects the real generated registry: removing any of the six preview configs fails the new tests, and all six Properties previews render correctly.

[Reviewed by Robohands]

PR #5917, opened 2026-09-02
fix(core): add playground wrapper for DropdownMenuRadioItem docsite preview

### LINKED ISSUE #5889: [Bug] DropdownMenuRadioItem Properties preview renders no working component
## Affected page

https://astryx.atmeta.com/components/DropdownMenuRadioItem?tab=properties

## Steps to reproduce

## Expected behavior

A radio menu item renders inside the required radio-group context.

## Actual behavior

The preview renders “Render error: DropdownMenuRadioItem must be used within a DropdownMenuRadioGroup” and logs the same page error.

## Diagnosis

The item is rendered standalone because its doc declares no playground wrapper.

The likely fix is local to `packages/core/src/DropdownMenu/DropdownMenuRadioItem.doc.mjs` and/or its focused docsite preview coverage; it does not require a new public API or design direction. This is an atomic, independently closable slice of #2008.

## Acceptance criteria

- [ ] Wrap the preview in a minimal `DropdownMenuRadioGroup` fixture with a valid value/change path.
- [ ] The Properties preview renders the item without an error on first load.
- [ ] The item can be selected and reflects its selected state.
- [ ] Add focused preview/data-extraction coverage for the wrapper.

Geervan:
Fixes #5889

## Summary

Adds the missing `playground.wrapper` configuration to `DropdownMenuRadioItem.doc.mjs` so the component preview renders wrapped inside a `DropdownMenuRadioGroup` fixture on the docsite **Properties** tab.

### Problem
Navigating to the **Properties** tab on the docsite for `DropdownMenuRadioItem` produced a page error:
> *“Render error: DropdownMenuRadioItem must be used within a DropdownMenuRadioGroup”*

Because `DropdownMenuRadioItem` requires `DropdownMenuRadioGroupContext` to function, rendering it standalone in the docsite playground caused a runtime context error.

### Solution
- Added `playground.wrapper` specifying `DropdownMenuRadioGroup` in `packages/core/src/DropdownMenu/DropdownMenuRadioItem.doc.mjs`.
- Added unit test assertion coverage in `apps/docsite/src/__tests__/data-extraction.test.ts`.
- Added patch changeset file `.changeset/dropdown-menu-radio-item-playground-wrapper.md`.

## Acceptance Criteria

- [x] Wrap the preview in a minimal `DropdownMenuRadioGroup` fixture with a valid value/change path.
- [x] The Properties preview renders the item without an error on first load.
- [x] The item can be selected and reflects its selected state.
- [x] Add focused preview/data-extraction coverage for the wrapper.

### DIFF
--- a/.changeset/dropdown-menu-radio-item-playground-wrapper.md
+++ b/.changeset/dropdown-menu-radio-item-playground-wrapper.md
@@ -1,0 +1,9 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] DropdownMenuRadioItem: add playground wrapper and wire wrapper selection state for docsite preview
+
+Wraps DropdownMenuRadioItem in DropdownMenuRadioGroup wrapper and binds wrapper selection state to active item value so aria-checked updates accurately on knob changes and click selection.
+
+@Geervan
--- a/.changeset/dropdown-menu-radio-item-playground-wrapper.md
+++ b/.changeset/dropdown-menu-radio-item-playground-wrapper.md
@@ -4,6 +4,6 @@
 
 [fix] DropdownMenuRadioItem: add playground wrapper and wire wrapper selection state for docsite preview
 
-Wraps DropdownMenuRadioItem in DropdownMenuRadioGroup wrapper and binds wrapper selection state to active item value so aria-checked updates accurately on knob changes and click selection.
+Wraps DropdownMenuRadioItem in DropdownMenuRadioGroup wrapper and keeps the wrapper's selection independent from the item's value knob, so aria-checked stays false until the item is activated and updates on click.
 
 @Geervan
--- a/apps/docsite/src/__tests__/interactive-preview.test.ts
+++ b/apps/docsite/src/__tests__/interactive-preview.test.ts
@@ -26,6 +26,7 @@ vi.mock('@stylexjs/stylex', () => ({
   props: () => ({}),
 }));
 vi.mock('@astryxdesign/core/theme/syntax', () => ({allSyntaxPresets: []}));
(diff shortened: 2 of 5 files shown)


cixzhang (changes requested):
The Properties preview now renders, but builders still cannot exercise its core selection behavior. After changing the item value to `option-2`, clicking it leaves `aria-checked="false"` because the wrapper stays fixed at `option-1` and receives a no-op `onChange`. Please wire wrapper selection into preview state and cover unchecked → click → checked.

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks for wiring this, but the original selection ask is still unresolved. The wrapper now always mirrors the item's value, so after changing the value control to `option-2`, the item is already `aria-checked="true"` before any click. Please keep wrapper selection independent, then cover `option-2` as unchecked before click and checked after it.

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks, selection works now. The remaining issue is coverage: the new test only renders DropdownMenuRadioGroup directly and still passes with InteractivePreview reverted to main. Please cover the actual Properties preview path—change to option-2, assert unchecked, activate it, then assert checked.

[Reviewed by Robohands]

Geervan:
Added regression coverage for the actual InteractivePreviewStage wrapper path. The test changes the item value to option-2, verifies it becomes unchecked, then activates it and verifies it becomes checked. Focused test, lint, formatting, and browser verification pass.

cixzhang (approved):
Thanks, this closes the Properties-preview loop. I verified option-2 is unchecked after editing and checked after activation, and the focused test fails with the preview fix reverted.

[Reviewed by Robohands]

PR #5925, opened 2026-09-02
test(ci): probe PR-only Storybook ownership

cixzhang:
## Purpose

Temporary CI probe for the merged PR-only Storybook package-ownership fix.

This adds a new Storybook story file whose source does not exist on `main`. A temporary no-op comment in Button activates the trusted stable-visual planner, so default-branch code must resolve the PR-only story’s package from its built Storybook title before the preview can complete.

Do not merge. This PR will be closed after the live preview is verified.


cixzhang:
CI probe complete: the PR-only story built, the trusted workflow resolved and captured it, the preview published through gh-pages, and the live Storybook rendered successfully in Chromium. Closing without merge as intended.

PR #5928, opened 2026-09-02
fix(Table): align the sortable header button with its column

ernestt:
## What

A sortable column with `align: 'end'` rendered its header label hard against the start of the cell, sitting above right-aligned figures. Same for `align: 'center'`.

The cause is that `align` lands as `textAlign` on the `<th>`, and the sortable plugin renders a **full-width flex button** inside that cell. `textAlign` cannot position flex children, so the alignment stopped at the cell boundary and never reached the label.

This mirrors the column's alignment onto the button's main axis via `justify-content`. `flex-end` resolves against the flex direction rather than the physical box, so it keeps mirroring correctly under RTL.

## Why it's split out

Found while building an invoice template ([#5865](https://github.com/facebook/astryx/pull/5865)), where every numeric column is both sortable and end-aligned. Pulled into its own PR so the template stays a pure template change and this can be reviewed as the consumer-visible core fix it is.

## Scope

| File | Change |
|---|---|
| `useTableSortable.tsx` | +19/−2 — two StyleX rules, applied from `column.align` |
| `useTableSortable.test.tsx` | +72 — four tests |
| `.changeset/` | patch |

Only `align: 'end'` and `align: 'center'` are affected. `align: 'start'` and unaligned columns produce byte-identical output, which the fourth test pins.

## Test plan

`useTableSortable.test.tsx` — 44 pass, including four new cases asserting that end and center each produce a button style distinct from the default and from each other, and that `start` stays identical to unaligned.

Non-sortable columns never had the bug: without the plugin there is no button, and `textAlign` reaches the label directly.

### DIFF
--- a/.changeset/table-sortable-header-align.md
+++ b/.changeset/table-sortable-header-align.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Table's sortable header button now follows its column's `align`, so an `align: 'end'` or `align: 'center'` column no longer gets a start-hugging header label sitting above right-aligned figures. Sorting wraps the header in a full-width flex button, which the `textAlign` that `align` sets on the cell cannot position; the alignment is now carried onto the button's main axis with a flow-relative `justify-content`, so it keeps mirroring under RTL.
+
+@ernestt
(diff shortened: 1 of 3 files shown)


cixzhang (approved):
Thanks, this is good. The fix stays inside the existing Table contract, and the focused checks cover the affected alignment combinations.

[Reviewed by Robohands]

PR #5930, opened 2026-09-02
docs(spec): define breaking-change classification

cixzhang:
## Why

Contributors need one compatibility rule for deciding when a published change is `[breaking]`, including a clear boundary between the CLI contract and the template catalog it serves.

## What

Defines the released-consumer test for breaking changes, separates compatibility from risk and implementation size, and records the Changeset and migration expectations.

For CLI templates, slugs, metadata, and starter source are mutable catalog data. Renaming a slug or rebuilding a template is nonbreaking. The surrounding `template` command, options, exit behavior, and machine-readable schema remain contractual.

## Risk

Documentation/specification only. No package behavior changes and no Changeset is required.

## Testing

- `pnpm check:knowledge`
- pre-commit `pnpm check:repo` (all 12 checks)

### DIFF
--- a/docs/specs/AST-017/spec.md
+++ b/docs/specs/AST-017/spec.md
@@ -1,0 +1,178 @@
+---
+schema_version: 1
+template_version: 1
+kind: system-spec
+id: spec:AST-017
+authority: current
+archive_reason: null
+superseded_by: null
+approved_by: cixzhang
+approved_at: 2026-09-02
+phase: accepted
+owners: [cixzhang, josephfarina]
+affects_architecture: [architecture:public-component-api]
+affects_families: []
+affects_contributing: [contributing:release-process, contributing:templates]
+affects_consumer_docs: [release-process, templates]
+---
+
+# Published compatibility and breaking-change classification system spec
+
+## Intent
+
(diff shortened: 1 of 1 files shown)


josephfarina:
I don't think individual template slugs should be part of the stable CLI contract.

Templates are an evolving catalog of examples. We expose them through `template show`, and `build` and other agent flows use them to find concrete examples. That makes the template mechanism and its documented output schema public. It doesn’t make every catalog entry permanent.

We should be able to rename, merge, replace, or remove templates as our guidance improves, without carrying aliases forever. Consumers can inspect a template, but they shouldn’t depend on its slug, metadata, or generated source staying fixed. A workflow that needs a fixed artifact should pin its CLI version or own the generated output.

So I think FR9 and DEC-2 have the contract backwards. Changing the template commands or their documented machine-readable schemas can be breaking. Changing an individual template entry should be nonbreaking, and we should state clearly that entries are subject to change.

cixzhang (commented):
Semantic verdict: approve

This makes the compatibility boundary clear: released contracts stay stable while the template catalog can evolve.

[Reviewed by Robohands]

josephfarina (approved):

Issue #5921, opened 2026-09-02
[Infra] No source build survives Turbopack: every documented path configures nextConfig.webpack, which Turbopack ignores


joaodotwork:
## Problem

Every documented way to run an Astryx source build in Next.js applies the `source`
export condition through `nextConfig.webpack` — `withAstryx()` via a scoped
`module.rules` entry, and `packages/build/README.md` via `config.resolve.conditionNames`.
Turbopack does not run `nextConfig.webpack` at all.

Next 16 builds with Turbopack by default. So on a default Next 16 project, a
correctly configured source build produces the same silently unstyled page as a
misconfigured one: `@astryxdesign/core` resolves to `dist` (`x`-prefixed class names)
while PostCSS emits `astryx`-prefixed rules, and nothing matches.

This is not the same issue as the `withAstryx()` issuer-scoping bug — the README's
config, which we verified works under webpack, still fails here.

## Evidence

Next 16.3.4, React 19.2.8, `@astryxdesign/core@0.5.2`, installed from npm. Same app
source, built both ways, measured on the served HTML and CSS:

| variant | bundler | build | `/` route | classes with no CSS rule |
|---|---|---|---|---|
| README's global `conditionNames` | webpack | ok, 18.8 s | `○ Static` | **0 of 141** |
| README's global `conditionNames` | **Turbopack** | ok, 8.5 s | `○ Static` | **134 of 141** |
| `withAstryx({})` | webpack | ok, 15.6 s | `○ Static` | 134 of 141 |
| `withAstryx({})` | Turbopack | ok, 8.5 s | `○ Static` | 134 of 141 |

Both Turbopack builds exit 0 with no warning. The failing build is the faster one.

For contrast, the pre-built path (`dist` + `dist/astryx.css`, no build plugin) works
under both bundlers — 0 unmatched classes, 2.7 s on Turbopack — and is what we plan
to ship. This report is about the source path only.

## Scope check

Build tooling, not a component defect.

## Proposed direction

1. **Detect and refuse.** `@astryxdesign/build` can tell it is running under
   Turbopack (`process.env.TURBOPACK`, or the absence of the webpack hook it
   installed). Erroring with "the Astryx source build requires `--webpack`" would
   turn a silent unstyled deploy into a one-line fix.
2. **State it in the docs.** `packages/build/README.md` and the
   `apps/example-nextjs-source` README currently say nothing about bundler choice,
   and the example pins `next: ^15.5.21`, where webpack is still the default — so the
[...]

PR #5932, opened 2026-09-02
fix(build): resolve app imports of astryx to source in withAstryx

### LINKED ISSUE #5920: [Infra] withAstryx() resolves @astryxdesign/core to dist while PostCSS compiles it from source, producing a silently unstyled app
## Problem

`withAstryx()` from `@astryxdesign/build/next` does not apply the `source` export
condition to an app's own imports. The app resolves `@astryxdesign/core` through
`default` to `dist`, whose runtime emits `x`-prefixed atomic class names, while the
PostCSS pass reads the library **source** and emits `astryx`-prefixed rules.

The two halves share no class names. The build succeeds, the route prerenders, every
element renders with correct markup and accessibility, a full-size stylesheet is
served — and the page is unstyled. Nothing errors and nothing warns.

`apps/example-nextjs-source` uses `withAstryx()`, so a team following it as written
will ship an unstyled site.

## Evidence

Next 16.3.4, React 19.2.8, `@astryxdesign/core@0.5.2`, installed from npm (not a
workspace link). One app source built under both bundlers, then served and measured
on the rendered HTML rather than on the exit code:

| next.config | `astryx*` classes on elements | `x*` classes | classes with no CSS rule | CSS served |
|---|---|---|---|---|
| `withAstryx({})` | **0** | 135 | **134 of 141** | 136 KB |
| `packages/build/README.md`'s global `conditionNames` | **134** | 1 | **0** | 136 KB |

Both builds exit 0 and report `/` as `○ (Static)`.

## Cause

```js
config.module.rules.unshift({
  test: /[\\/]node_modules[\\/]@astryxdesign[\\/]/,
  resolve: {conditionNames: ['source', '...']},
});
```

In webpack, `Rule.test` matches **the module being processed**, and `Rule.resolve`
governs the requests *that* module makes. The import that needs the `source`
condition is issued by app code — `src/app/page.tsx` — which does not match
`node_modules/@astryxdesign/`. The rule never fires for it. Once resolution lands in
`dist`, its internal imports are relative paths already inside `dist`, so the rule
has nothing left to affect.

The scoping is deliberate and the reasoning in the source comment is sound: a global
`conditionNames` breaks React's JSX resolution (which relies on webpack's `'...'`
sentinel plus `react-server`) and mis-resolves third-party packages such as `lexical`
that also publish a `source` condition. The narrowing solves that and introduces this.

## Scope check

Not a component defect. It is the build helper the package exports, and the failure
mode is silent, which is why it is worth a report rather than a doc note.

## Proposed direction

Three options, in the order we would prefer them:

1. **Make it work by issuer.** Apply `conditionNames: ['source', '...']` in a rule
   whose `test` matches the *app's* sources (or `resolve.byDependency`), so requests
   for `@astryxdesign/*` from app code resolve to source. Scoping by *what is being
   requested* rather than by *who is requesting* is what the helper intends.
2. **Fail loudly instead.** Have `@astryxdesign/build/postcss` compare the class
   prefix it emitted against the prefix present in the built output, and error when
   they disagree. A prefix mismatch is always a misconfiguration; it should never be
   a silent success.
3. **Document the helper as monorepo-only** and point external consumers at the
   README's global `conditionNames`, which we verified works.

Independently: `withAstryx()` writes `nextConfig.webpack`, which Turbopack ignores
outright, so no source build survives Turbopack even when configured correctly. Filed
separately.

## Reproduction

Self-contained, from an empty directory:

```bash
npx create-next-app@16.3.4 repro --ts --app --no-tailwind --no-eslint --no-src-dir
cd repro
npm i @astryxdesign/core@0.5.2 @astryxdesign/theme-neutral@0.5.2 @stylexjs/stylex@0.19.0
npm i -D @astryxdesign/build@0.5.2 @stylexjs/babel-plugin@0.19.0 \
         @stylexjs/postcss-plugin@0.19.0 @babel/core autoprefixer
```

```js
// next.config.mjs
import {withAstryx} from '@astryxdesign/build/next';
export default withAstryx({});
```

```js
// babel.config.js
const {babel} = require('@astryxdesign/build');
module.exports = babel(__dirname);
```

```js
// postcss.config.js
const {postcss} = require('@astryxdesign/build');
module.exports = postcss(__dirname);
```

```css
/* app/layers.css */
@layer reset, astryx-base, astryx-theme, product;
```

```css
/* app/globals.css */
@import './layers.css';
@import '@astryxdesign/core/reset.css';
@import '@astryxdesign/theme-neutral/theme.css';

@stylex;
```

Add `"browserslist": ["last 1 Chrome version"]` to `package.json`, render any Astryx
component from a client-side provider (`Theme` + `LinkProvider`, per
`apps/example-nextjs`), then:

```bash
npx next build --webpack && npx next start
```

The page renders correctly and is unstyled. The diagnostic is the class prefix — one
line each:

```bash
# class names on the rendered elements: x-prefixed (dist runtime)
curl -s localhost:3000/ | grep -o 'class="[^"]*"' | grep -o '\bx[0-9a-z]\{5,\}' | sort -u | head

# rules in the served CSS: astryx-prefixed (compiled from source)
curl -s localhost:3000/_next/static/chunks/*.css | grep -o '\.astryx[0-9a-z]\{5,\}' | sort -u | head
```

Two disjoint sets. Swapping `next.config.mjs` for the global `conditionNames` form in
`packages/build/README.md` makes both sets `astryx`-prefixed and the page renders styled.

Our harness automates this across every documented setup and both bundlers, and reports
how many rendered classes have no matching rule. Happy to extract it as a standalone
repro repo if that is more useful than the steps above.

## Version

`@astryxdesign/core@0.5.2`, `@astryxdesign/build@0.5.2`, `@stylexjs/babel-plugin@0.19.0`,
`@stylexjs/postcss-plugin@0.19.0`, Next 16.3.4, React 19.2.8, Node 24.17.0, npm 11.13.0,
macOS.

PRIEYAN:
`withAstryx()` installs a rule that carries the `source` condition on `Rule.resolve`, scoped
by `test` to `node_modules/@astryxdesign/`. But `Rule.test` matches the module being
processed and `Rule.resolve` governs the requests *that* module makes, so the rule only
covers astryx-to-astryx imports. An app's own `@astryxdesign/*` imports are issued from its
sources, outside `node_modules`, so they never match and resolve through `default` to dist.

Dist's runtime emits `x`-prefixed atomic class names while the PostCSS pass compiles the
library from source and emits `astryx`-prefixed rules. The two sets share no class names, so
the build exits 0, the route prerenders, a full-size stylesheet is served, and the page
renders unstyled with nothing logged. `apps/example-nextjs-source` uses `withAstryx({})` as
written, so a team following it ships an unstyled site.

Widening the global `conditionNames` does fix resolution, but it breaks React's JSX
resolution and mis-resolves third-party packages that also publish a `source` condition
(`lexical`) — which is exactly what the scoped rule exists to prevent. Webpack rules cannot
key on the request string, so there is no rule shape that expresses "requests *for*
`@astryxdesign/*` *from* app code". Instead this reads the `source` targets out of each
installed astryx package's export map and sets them as `resolve.alias` entries: resolution
follows the astryx packages themselves and the global conditions stay as Next resolved them.
Subpaths get an entry each, since `@astryxdesign/core/AlertDialog` is as much a documented
entry point as the root — aliasing only the root would leave 119 subpaths on dist and produce
a mixed bundle. A user-supplied alias still wins, and packages that are absent or ship no
`source` condition keep normal resolution.

Note that `require.resolve('@astryxdesign/core/package.json')` throws
`ERR_PACKAGE_PATH_NOT_EXPORTED` — core's `exports` has no `./package.json` key — so the
manifest lookup walks `node_modules` the way Node resolves a bare specifier.

On verification, being precise about what I did and did not run. I added
`packages/build/src/next.test.mjs`, 9 assertions over the config `withAstryx()` emits against
a fixture package laid out the way npm installs one. It fails on 5 of them before this change
and passes all 9 after. I could not run it through the repo's own vitest — `pnpm install`
pulls the full monorepo toolchain and I deliberately kept build tooling off this machine — so
I executed the same assertions directly against the module under Node; CI will be the first
run through vitest proper. `check:changesets`, `check:cli-structure`, `check:portable-scripts`,
`check:executable-bits` and `check:use-client` pass locally.

I have not reproduced the unstyled-page symptom end to end. There is no `next build` here, so
the class-prefix mismatch itself is verified from the resolution semantics and the reporter's
measurements rather than observed. The reporter has a harness that automates this across every
documented setup and both bundlers; pointing it at this branch would be a stronger signal than
anything I can produce locally.

Turbopack is untouched — `withAstryx()` writes `nextConfig.webpack`, which Turbopack ignores,
and that is filed separately as #5921.

Fixes #5920

### DIFF
--- a/.changeset/withastryx-app-source-resolution.md
+++ b/.changeset/withastryx-app-source-resolution.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/build': patch
+---
+
+[fix] `withAstryx()` now resolves an app's own `@astryxdesign/*` imports to the packages' `source` entries. The scoped webpack rule only governs requests issued from inside `node_modules`, so app code resolved the library through `default` to `dist` while PostCSS compiled it from source — the two emit disjoint class names and the app rendered unstyled without erroring. (#5932)
+
+@PRIEYAN
--- a/packages/build/src/next.js
+++ b/packages/build/src/next.js
@@ -14,6 +14,70 @@
  *   });
  */
 
+const fs = require('fs');
+const path = require('path');
+
+const ASTRYX_MODULE = /[\\/]node_modules[\\/]@astryxdesign[\\/]/;
+
+/**
+ * Locate an installed package's directory by walking `node_modules` up from
+ * the app, the way Node resolves a bare specifier. Returns null when the
+ * package is not installed.
+ */
+function findPackageDir(name, from) {
+  let dir = path.resolve(from);
+  for (;;) {
+    const candidate = path.join(dir, 'node_modules', name);
+    if (fs.existsSync(path.join(candidate, 'package.json'))) {
+      return candidate;
+    }
+    const parent = path.dirname(dir);
+    if (parent === dir) {
+      return null;
+    }
+    dir = parent;
+  }
+}
+
+/**
+ * Map each installed astryx package to the `source` targets in its export map,
+ * so an app's imports reach raw TS the way an in-repo build does. Subpaths get
+ * an entry each — `@astryxdesign/core/AlertDialog` is as much a documented
+ * entry point as the package root. Packages that are absent, and export keys
+ * that ship no `source` condition, are skipped and keep normal resolution.
+ */
+function sourceEntryAliases(packages, context) {
+  const from = (context && context.dir) || process.cwd();
+  const aliases = {};
+  for (const name of packages) {
+    // Not `require.resolve(`${name}/package.json`)`: packages that define
+    // `exports` without a `./package.json` key — astryx's own among them —
+    // make that request throw ERR_PACKAGE_PATH_NOT_EXPORTED.
+    const packageDir = findPackageDir(name, from);
+    if (packageDir == null) {
+      continue;
+    }
+    let exports;
+    try {
+      exports = require(path.join(packageDir, 'package.json')).exports;
+    } catch {
+      continue;
+    }
+    if (exports == null || typeof exports !== 'object') {
+      continue;
+    }
+    for (const [key, value] of Object.entries(exports)) {
+      const entry = value && typeof value === 'object' ? value.source : null;
+      if (typeof entry !== 'string' || key.includes('*')) {
+        continue;
+      }
+      const request = key === '.' ? name : `${name}/${key.slice(2)}`;
+      aliases[`${request}$`] = path.resolve(packageDir, entry);
+    }
+  }
+  return aliases;
+}
+
 /**
  * Wraps a Next.js config to enable Astryx source builds.
  * - Adds transpilePackages for @astryxdesign/* packages
@@ -46,11 +110,27 @@ function withAstryx(nextConfig = {}) {
       config.module = config.module || {};
       config.module.rules = config.module.rules || [];
       config.module.rules.unshift({
-        test: /[\\/]node_modules[\\/]@astryxdesign[\\/]/,
+        test: ASTRYX_MODULE,
         resolve: {
           conditionNames: ['source', '...'],
         },
       });
+
+      // The rule above only covers astryx-to-astryx imports: `Rule.test`
+      // matches the module being processed and `Rule.resolve` governs the
+      // requests that module makes, so an app's own `@astryxdesign/*` imports —
+      // issued from its sources, outside node_modules — never match it. Those
+      // resolve through `default` to dist, whose runtime class names are
+      // disjoint from the CSS the PostCSS pass compiles out of source, and the
+      // app renders unstyled without erroring. Rules cannot key on the request
+      // string, so point the packages at their `source` entry directly; the
+      // global conditions stay as Next's defaults for React and for
+      // third-party `source` shippers such as `lexical`.
+      config.resolve = config.resolve || {};
+      config.resolve.alias = {
+        ...sourceEntryAliases(astryxPackages, context),
+        ...(config.resolve.alias || {}),
+      };
 
       // Preserve the symlinked node_modules path so Next.js's
       // transpilePackages matcher recognizes @astryxdesign/* packages under
--- a/packages/build/src/next.js
+++ b/packages/build/src/next.js
@@ -207,10 +207,6 @@ function withAstryx(nextConfig = {}) {
       // global conditions stay as Next's defaults for React and for
       // third-party `source` shippers such as `lexical`.
       config.resolve = config.resolve || {};
-      config.resolve.alias = mergeAliases(
-        sourceEntryAliases(astryxPackages, context),
-        config.resolve.alias,
-      );
 
       // Preserve the symlinked node_modules path so Next.js's
       // transpilePackages matcher recognizes @astryxdesign/* packages under
@@ -221,10 +217,21 @@ function withAstryx(nextConfig = {}) {
       config.resolve.symlinks = false;
 
       // Call user's webpack config if provided
-      if (existingWebpack) {
-        return existingWebpack(config, context);
-      }
-      return config;
+      const merged = existingWebpack
+        ? existingWebpack(config, context) || config
+        : config;
+
+      // Merge after the caller's hook, not before: aliases the hook adds are
+      // as authoritative as the ones already on the config, and merging first
+      // would leave the generated entries ahead of them in the alias list —
+      // where the resolver, taking the first match, loads astryx source
+      // instead. Only a byte-identical `$` key would have replaced ours.
+      merged.resolve = merged.resolve || {};
+      merged.resolve.alias = mergeAliases(
+        sourceEntryAliases(astryxPackages, context),
+        merged.resolve.alias,
+      );
+      return merged;
     },
   };
 }
--- a/packages/build/src/next.js
+++ b/packages/build/src/next.js
@@ -40,21 +40,40 @@ function findPackageDir(name, from) {
 }
 
 /**
+ * Match a wildcard alias key such as `@astryxdesign/*` against a request. The
+ * resolver treats `*` as the variable part of the specifier, so a caller can
+ * claim a whole scope with one entry.
+ */
+function wildcardPattern(name) {
+  const source = name
+    .split('*')
+    .map(part => part.replace(/[.*+?^${}()|[\]\\]/g, '\\$&'))
+    .join('.*');
+  return new RegExp(`^${source}$`);
+}
+
+/**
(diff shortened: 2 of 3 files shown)


cixzhang (changes requested):
Thanks for fixing the silent unstyled build. One compatibility issue still blocks this: an ordinary caller alias such as `{'@astryxdesign/core': customPath}` loses to the generated exact aliases, so composed Next configs silently load Astryx source instead of the caller's implementation. Please preserve both exact and prefix caller aliases; a resolver-level regression test would cover the real behavior.

Also please run Prettier on the two changed source files.

[Reviewed by Robohands]

PRIEYAN:
good catch, and it was worse than the exact/prefix mismatch. `{...generated, ...caller}` only
dedupes when the keys are identical — a caller writing `'@astryxdesign/core'` is a different key
from the generated `'@astryxdesign/core$'`, so both survived and mine sat first in the list.
first match wins, so the caller lost on the bare specifier and on every subpath.

now dropping any generated alias the caller has already spoken for: a prefix alias claims the
package and everything under it, an exact `$` alias claims only the bare specifier so the
generated subpaths still apply under it. also fixed an array-shaped `resolve.alias` — spreading
an array into an object turned it into index keys and broke resolution outright, which the new
test caught.

moved the tests onto `enhanced-resolve` as you suggested, asserting the resolved path rather
than reading keys back out of the config, since precedence is an ordering property of the
resolver. 13 assertions, and the three that cover this fail on the previous commit:

```
✗ lets a caller prefix alias win over the generated entries
    expected "/tmp/astryx-next-kFBf5S/custom/index.js"
      actual: "/tmp/astryx-next-kFBf5S/node_modules/@astryxdesign/core/src/index.ts"
✗ lets a caller prefix alias win for subpaths too
✗ preserves an array-shaped caller alias
```

one thing to flag — `enhanced-resolve` is a phantom import there. it's in the lockfile at 5.21.6
via webpack and the hoisted layout puts it at the root, but it isn't declared in
`packages/build`. i left it undeclared rather than add a devDependency i can't regenerate the
lockfile for; say the word if you'd rather it be explicit and i'll add it.

prettier run on both files, `--check` clean.

cixzhang (changes requested):
Thanks for fixing the exact, ordinary prefix, and array alias cases; those now pass through the real resolver. Two compatibility gaps remain:

- Apps using an object wildcard alias like `{'@astryxdesign/*': custom}` still load Astryx source instead of their configured implementation because generated exact aliases stay first (`packages/build/src/next.js:53`).
- Contributors on macOS get 6 of 13 resolver tests failing because the fixture compares `/var/...` with the canonical `/private/var/...` result (`packages/build/src/next.test.mjs:42`).

Could you preserve wildcard caller aliases too and canonicalize the fixture root? If you'd rather talk it through with someone, we're in [Discord](https://discord.com/invite/XnsUcFykEP).

[Reviewed by Robohands]

PRIEYAN:
both confirmed and fixed in a280736.

wildcard first — i checked the resolver actually honors `*` in alias keys before assuming it,
and it does, so `{'@astryxdesign/*': custom}` was landing behind the generated exact entries
exactly as you said. wildcard keys now count as claiming the requests they match, same as the
exact and prefix forms. two tests, both failing on 222c33f:

```
✗ lets a caller wildcard alias win over the generated entries
    expected ".../custom/index.js"
      actual: ".../node_modules/@astryxdesign/core/src/index.ts"
✗ lets a caller wildcard alias win for subpaths too
```

worth noting the two target shapes behave differently and i tested both — a plain target
collapses every request onto the one directory, while `custom/*` substitutes the matched part
and keeps the subpath. the subpath test uses the latter since that's the form that preserves
structure.

for the macOS one i couldn't hit it directly on linux, but pointing TMPDIR at a symlink
reproduces the same condition, and it's not 6 failures, it's 9:

```
$ TMPDIR=<symlinked dir> node run-tests   # before
7 passed, 9 failed
$ TMPDIR=<symlinked dir> node run-tests   # after
16 passed, 0 failed
```

fixture root is `realpathSync`'d now, and i added a symlinked-app-root case so the class of bug
stays covered on every platform rather than only where tmp happens to be a symlink. 16 pass
under both a normal and a symlinked TMPDIR here.

prettier clean on both files. thanks for the discord offer — nothing blocking on my side, but
the phantom `enhanced-resolve` import from last round is still the one open call if you'd rather
it be declared.

cixzhang (changes requested):
Thanks, the wildcard and macOS fixes themselves are good. One composition path still has the original compatibility bug: `withAstryx()` merges generated aliases before it calls the caller’s `webpack` hook (`packages/build/src/next.js:209-225`). When that hook adds an ordinary prefix or wildcard alias, enhanced-resolve still picks the earlier generated exact entries and loads Astryx source; only the identical exact `$` key wins. Could we preserve exact, prefix, and wildcard aliases configured through the hook too, with resolver-level coverage for that real Next composition path? If you'd rather talk it through with someone, we're in [Discord](https://discord.com/invite/XnsUcFykEP).

[Reviewed by Robohands]

PRIEYAN:
you're right, and it was the ordering rather than the matching — `mergeAliases` ran before
`existingWebpack`, so anything the hook added landed behind the generated entries and the
resolver took ours first. only a byte-identical `$` key replaced ours, which is exactly the
asymmetry you described.

fixed in 6cf541e by running the hook first and merging against the config it returns, so
aliases get the same precedence whichever side configures them. confirmed on the composition
path before the change:

```
hook adds PREFIX   @astryxdesign/core   -> /node_modules/@astryxdesign/core/src/index.ts
hook adds WILDCARD @astryxdesign/*      -> /node_modules/@astryxdesign/core/src/index.ts
hook adds EXACT    @astryxdesign/core$  -> /custom/index.js
```

and after, all three land on `/custom/...` while a request the hook doesn't claim still reaches
astryx source.

resolver-level coverage for that path specifically — prefix, wildcard and exact from a caller
hook, plus one the hook leaves alone. the two composition tests fail on a280736 and the exact
one passes there, same asymmetry.

one extra thing that fell out of it: a hook returning nothing used to throw, since the return
value went straight into the alias merge. it falls back to the config it was handed now, with a
test.

21 assertions, passing under both a normal and a symlinked TMPDIR. prettier clean.

cixzhang (approved):
Thanks, this now preserves caller aliases through both direct config and the webpack hook. Resolver coverage, macOS paths, formatting, and exact-head CI all pass.

[Reviewed by Robohands]

PR #5934, opened 2026-09-02
feat(templates): add the table-inbox page template

ernestt:
## What

A support triage queue beside a reading pane — the two-surface mail pattern the gallery did not have. `Table - Split Pane` was an empty category until now.

The list is a real data-mode `Table`, not a stack of rows pretending to be one, so it keeps row status, selection and the whole plugin surface and gives up only its dividers.

light

## The parts worth looking at

**Hover swaps the timestamp for actions.** Not an overlay — `useContainerReveal` with `position: var(--_reveal-position)` lets archive, move, delete and reply *claim* the cell, so the timestamp yields rather than getting painted over. The row height is pinned so nothing jumps.

hover

**Every message is a collapsible**, headed by name, address and a date out at the chevron, newest pre-expanded. Bodies start under the sender's name rather than under the avatar, which leaves the avatars hanging in a column of their own. Attachments are `Thumbnail`s that open a zoomable `Lightbox` gallery.

lightbox

**Selecting rows swaps the tab strip for `table-filter`'s bulk band** rather than stacking a second row above it. Same muted fill, same `@starting-style` rise on entry, same full-bleed run above the header — but pinned to the strip's height, so ticking the checkbox you just clicked does not move the table out from under the pointer. Clearing the selection puts the tabs straight back.

**Reply and Reply all** sit under the last message and open the composer already addressed — Reply all gathering everyone in the thread except the answering team. The composer is a non-modal `BottomSheet` that hugs its own height, so the list stays live and clickable while you write. Placeholder-only fields, Cc and Bcc added on demand.

composer

**Responsive to the surface, not the window.** Every threshold here is measured off the `Layout` root with a `ResizeObserver`, not asked of `matchMedia`. That matters because a page template is not always given the page: the docsite preview renders templates inline into a dialog, and a viewport query there describes something the template cannot see. As the *table* narrows — by dragging the pane, or by the whole surface shrinking — rows stack sender over subject, keep the date right and drop the checkbox. Narrower still, the pane takes the whole surface and the thread header stacks the back button above the subject so it can run to two lines.

`@container` is the right question in the wrong instrument: its answer never leaves CSS, and what changes here is the column *list*, which only React can do.

<table>
<tr><td width="50%"></td><td width="50%"></td></tr>
<tr><td align="center"><em>Stacked rows</em></td><td align="center"><em>Single surface</em></td></tr>
</table>

<details>
<summary>Remaining states — dark, full-width table, selection, single-surface list</summary>

dark
wide
selection
mobile list

</details>

## Rubric score: 97/100 (A)

Rubric v1.3.

| Category | Score | |
|---|---|---|
| Astryx component purity | 30/30 | Zero raw HTML tags in 2376 lines |
| Icon purity | 15/15 | Zero raw SVG; every icon an `Icon` or an `icon={}` prop |
| Custom CSS | 12/15 | Two constants and one `stylex.create` block, all justified — see below |
| Layout & structure | 15/15 | `Layout` root, single page, responsive without fixed columns |
| Doc metadata | 10/10 | 6/6 fields, 3/3 description, 1/1 naming |

The description went through a correction worth flagging: it first shipped at 1,570 characters, which is three times the ~490-character case the rubric names as a catch-all and caps at **1/3**, and the longest of 54 page templates by a wide margin. It is now 59 words / 390 characters carrying all four slots, which is what earns the 3.
| Image handling | 5/5 | Three `/template-assets` PNGs, all checked in |
| Code quality | 10/10 | |

These are the only custom CSS in the file:

- `ACTIONS_CLAIM_CELL` (`flexBasis: '100%'`, `flexShrink: 0`) — `Stack` has no flex-basis prop and `StackItem` no width, and the hover actions have to take the cell geometrically rather than paint over the timestamp.
- `HIDE_SQUEEZED_CONTENT` (`overflow: 'hidden'`) — clips the attachment clip as the cell narrows. `isScrollable` gives `auto`, not `hidden`.
- `styles.bulkBand` / `styles.bulkBandEnter` — the selection band's radius and its `@starting-style` entry. This one is not justified by absence of a prop but by precedent: it is `table-filter`'s band, copied property for property, because two tables that both let you tick rows and act on the set should not disagree about what that looks like. A prop cannot express `@starting-style` in any case, which is why the reference reached for `stylex` too.

I A/B'd the two templates side by side in one viewer rather than trusting the copy. Fill `rgb(241,241,241)`, radius 10px, transition 0.3s and the full-bleed run all match exactly. Height is the one deliberate divergence: `table-filter`'s band sizes to its own content at 44px, while this one is pinned to the 37px tab strip it swaps with, so ticking a checkbox moves the table by zero pixels — verified across select, grow and clear.

Worth flagging for whoever owns `Section`: the radius is invisible in **both** templates. `Section` paints its variant fill on an inner element while `xstyle` lands on the outer, so the corners stay square. I left it matching rather than "fixed" so the two do not drift; an `overflow: hidden` on the outer would make the radius real in both.

If `Stack` gains a flex-basis prop the first two go away.

## Depends on

#5933 — the `Collapsible` trigger-grow fix is what puts the message dates out by the chevron. **This PR is mergeable without it**; the dates just trail the address instead of reaching the edge. The screenshots above are taken with #5933 applied.

## Test plan

- `pnpm -F @astryxdesign/cli typecheck:strict` — clean.
- `eslint` on the template — clean.
- Ten states captured above at 2x, light and dark, four viewports.
- Walked with Playwright: row click opens the pane, inner controls don't; resize handle drags; tab filtering and unread badges agree; lightbox pages through a gallery; composer stays non-modal (rows still clickable behind it); selection bar appears only when checkboxes are reachable.

### DIFF
--- a/.changeset/table-inbox-reply-drafts.md
+++ b/.changeset/table-inbox-reply-drafts.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/cli': patch
+---
+
+[fix] Table inbox replies now preserve an unsent body only for the same conversation, preventing text from following changed recipients. (#5934)
+@ernestt
--- a/apps/docsite/src/components/templateComponents.ts
+++ b/apps/docsite/src/components/templateComponents.ts
@@ -162,6 +162,10 @@ export const TEMPLATE_COMPONENTS: Record<
     () =>
       import('../../../../packages/cli/assets/templates/pages/table-grouped/page'),
   ),
+  'table-inbox': lazy(
+    () =>
+      import('../../../../packages/cli/assets/templates/pages/table-inbox/page'),
+  ),
   'table-page': lazy(
     () =>
       import('../../../../packages/cli/assets/templates/pages/table-page/page'),
--- a/packages/cli/assets/templates/pages/table-inbox/page.tsx
+++ b/packages/cli/assets/templates/pages/table-inbox/page.tsx
@@ -1,0 +1,2376 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+'use client';
+
+/**
+ * A triage queue where the row is the control surface: hovering it swaps the
+ * timestamp for four actions, clicking it opens the thread in a reading pane.
+ *
+ * An inbox is a table whose rows are verbs. Nobody sorts an inbox by sender or
+ * compares one row against the next — they scan for the thread that matters and
+ * then do something to it. So the columns are here to hold sender, subject,
+ * tags, and time in the same place on every row, which is what makes the list
+ * scannable at all, and everything else in the template is about acting on a
+ * row without leaving the list.
+ *
+ * ## Extending this template
+ *
+ * **The hover swap is CSS, and it has to be.** Four buttons take the
+ * timestamp's place while the pointer is on the row. The obvious build —
+ * `onMouseEnter` writing a `hoveredId` — re-renders the table on every row the
+ * cursor crosses, and gets the accessibility wrong in three separate ways.
+ * `useContainerReveal` is the primitive for it: the container publishes its
+ * hover state as inherited custom properties and the content reads them, so the
+ * swap costs no React state and no re-render. Keyboard focus reveals the
+ * actions with no dwell to wait through, coarse pointers get them permanently
+ * (they are never behind a hover a touch user cannot perform), and the buttons
+ * stay mounted and tabbable at rest rather than being `display: none`.
+ *
+ * The container is the `<tr>` and the content is inside a `<td>`, which works
+ * because the scoping is inheritance rather than a selector — see
+ * `useInboxRows`, where the container props are merged onto the row.
+ * `hoverDelay` is the other half: without a dwell, dragging the cursor down the
+ * list lights up all nine rows in sequence.
+ *
+ * **Read/unread is the row-status plugin, and that decided the whole file.**
+ * `useTableRowStatus` renders a status gutter, and plugins that render only run
+ * in data-driven mode — passing `children` replaces the entire header-and-body
+ * render, so a hand-composed table gets no plugin cells at all. Data mode in
+ * turn always renders a header when columns are defined. A headerless inbox and
+ * a plugin-driven one are mutually exclusive today; this template takes the
+ * plugins, and the header earns its place by labelling the sort affordances.
+ *
+ * **Two washes cannot share the row's inline style.** The selection plugin
+ * writes `el.style.backgroundColor` imperatively on every notify, so an
+ * open-row background set in `transformBodyRow` gets wiped the next time
+ * selection changes. `hasRowHighlight: false` turns the plugin's wash off —
+ * it exists for exactly this case — and the open row is painted on its
+ * *cells* instead, which nothing else touches.
+ *
+ * **Selecting must never open the thread.** Two activation surfaces overlap on
+ * every row: the row opens it, the checkbox and the four actions do not. The
+ * tempting fix is `stopPropagation` on each inner control, and it fails on the
+ * one that matters — the checkbox belongs to the selection plugin, so there is
+ * no handler here to hang it on, and ticking a box opens the reading pane.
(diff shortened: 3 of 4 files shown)


cixzhang (changes requested):
Thanks, the container-based reflow is working. One thing before this lands: after opening a thread, switching categories updates the table but leaves the old thread in the pane, so a triager can act on a conversation outside the selected queue. Please clear or re-derive `chosenId` when `category` changes (`page.tsx:1890`).

[Reviewed by Robohands]

ernestt:
Good catch — fixed in 42794cdda97, and it turned out to have a sibling.

**The pane.** `openId` now falls back to the category's default whenever the chosen thread is not in the visible queue, so the pane re-derives from the category rather than outliving it. An explicit close still wins: `null` means the user shut the pane, and changing category is no reason to reopen it. Verified by opening a thread and switching: the pane follows to a Resolved thread in Resolved and a Waiting thread in Waiting, and in each case the subject on screen is one of the rows in the table below it.

**The selection had the same hole.** It is not reachable through the tabs, because the bulk bar takes the tab strip's place while a selection is held — but it is reachable by narrowing until rows stack, which hides the bar and puts the tabs back:

| step | bar | tabs |
|---|---|---|
| tick a row in Escalations, wide | `1 conversation selected` | hidden |
| narrow until rows stack | — | back |
| switch to Resolved | — | shown |
| widen again | **—** (was: `1 conversation selected`) | shown |
| back to Escalations | `1 conversation selected` | hidden |

`selectedCount` is now counted against the visible queue. The set itself is left alone, which is the last row: switching back finds the selection where it was rather than silently dropped.

cixzhang (changes requested):
Thanks, the category/pane mismatch is fixed. One remaining recipient-safety issue: starting Reply on Dana after closing Priya’s unsent draft re-addresses the composer to Dana but keeps Priya’s body. A triager can send one customer’s message to another. Please scope preserved drafts to their conversation, or clear/confirm the body when a reply targets a different thread.

[Reviewed by Robohands]

ernestt:
Addressed in b71942c3d76: reply drafts now carry their conversation identity, preserve an unsent body when Reply reopens the same thread, and clear it before changing recipients for a different thread. Added the required @astryxdesign/cli patch changeset; repository checks and focused lint passed.

cixzhang (changes requested):
Thanks, the body now clears when Reply switches conversations, but the same draft still keeps Cc/Bcc state. I reproduced this by adding Lucas in Cc while replying to Priya, closing, then replying to Dana: the body is empty and Dana is To, but Lucas remains in Cc. A triager can still send one conversation's copied recipients into another. Please reset every recipient-specific draft field when the conversation changes, while preserving them when reopening the same conversation.

[Reviewed by Robohands]

ernestt:
Addressed in 7b85c17aa29: switching conversations now clears Cc, Bcc, their expanded-field state, and the body before applying the new To/subject, while reopening the same conversation preserves the full draft. Focused lint, changeset validation, the core build, template-doc typecheck, and 13 focused template tests passed.

cixzhang (dismissed):
Thanks, Cc/Bcc now settle correctly. One safety gap remains: the reset at `page.tsx:1680` runs after the composer opens. Switching from Priya to Dana produced an initial frame with Priya, Marcus, Aiko, Priya’s subject, and “Draft for Priya only”; only afterward did it reset to Dana. A triager can see one customer’s draft while replying to another. Please make the new conversation’s first visible frame contain only its own recipients and subject with an empty body, while reopening the same conversation still preserves the draft.

[Reviewed by Robohands]

cixzhang (dismissed):
Thanks, the previous reply-draft leak is fixed. One responsive regression remains: after widening the pane with the separator, narrowing the preview leaves the panel at 1066px while its new maximum is 834px, so it overlaps the table instead of preserving the 300px list. Please clamp the selected pane size whenever the surface-derived maximum decreases (`page.tsx:2116`).

[Reviewed by Robohands]

PR #5938, opened 2026-09-02
fix(cli): a thin build kit says what to try instead of looking empty

josephfarina:
Split out of #5320 at review request: the thin-kit hint, on its own, with red→green evidence.

## The gap

`astryx build "quantum flux capacitor telemetry"` — before:

```
DOMAIN COMPONENTS
name:        Stack
description: Stack arranges items in a row or column...

FRAME + FOUNDATION
frame:      AppShell, TopNav, SideNav, Layout
foundation: VStack, HStack, Grid, ...
```

One incidental component, the always-on list, and nothing else. An agent reading that does not conclude its wording was wrong — it concludes the package has nothing and falls back on its own memory of what Astryx contains. That is precisely the failure `build` exists to prevent.

After, same query, one added section:

```
FEW MATCHES
Few matches. This is keyword search, not semantic — try other wordings,
or browse with `astryx component --list` and `astryx template --list`.
```

## The threshold counts what survived, not what search returned

Below three **offerable** results (pages + blocks + domain). Deliberately not `hasResults`, which is already true for a query that matched things and then had them all filtered out by the score floors — and that is the case most likely to be misread, because the output looks populated. There is a test for exactly that shape.

`hint` is absent when it does not apply, so a healthy kit is byte-identical to before. The CLI renders it last so it is the line the reader leaves with.

## Also

- `BuildKitResponse` type and the public response doc (`build.doc.mjs`) both updated.
- 3 new tests: fires when thin, silent when healthy, and fires for the matched-then-filtered case.

Gates: `check:repo` green, `lint:strict` 0 errors, build suite 8/8.

### DIFF
--- a/.changeset/build-thin-kit-hint.md
+++ b/.changeset/build-thin-kit-hint.md
@@ -1,0 +1,17 @@
+---
+'@astryxdesign/cli': patch
+---
+
+[fix] A thin `build` kit now says what to try instead of looking empty.
+
+`build "quantum flux capacitor telemetry"` returned one incidental component and the always-on frame list, and said nothing else. An agent reading that does not conclude its wording was wrong — it concludes the package has nothing and falls back on its own memory of what Astryx contains, which is the failure `build` exists to prevent.
+
+Below three offerable results the kit carries a `hint` naming the two commands that browse rather than search, and saying plainly that this is keyword matching, not semantic. The threshold counts what SURVIVED the score floors, not what search returned: `hasResults` is already true for a query that matched things and then filtered them all out, and that is the case most likely to be misread.
+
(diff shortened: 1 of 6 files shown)


cixzhang (dismissed):
Builders without a global `astryx` binary get unusable recovery commands: this pnpm workspace renders every other build-kit command as `pnpm exec astryx`, but `FEW MATCHES` freezes bare `astryx` commands into the API prose. Please keep recovery commands structured/package-manager-agnostic and format them in the CLI renderer.

The output also says its complete section order ends at `FRAME + FOUNDATION`, then emits `FEW MATCHES`.

Finally, the “matched then filtered” test returns six offerable results and takes the no-hint branch. Please assert a deterministic `hasResults: true`, zero-offerable case.

[Reviewed by Robohands]

PR #5939, opened 2026-09-02
feat(core): make the CLI visible where agents read

josephfarina:
## Summary

- generate a CLI search hint at the top of every shipped component doc
- stamp every generated TypeScript declaration during the core build
- append a copyable CLI search to development warnings while keeping production error text unchanged
- keep docs and tests out of runtime component audit scopes, so generated comments do not fan out visual, RTL, or accessibility jobs
- fail `check:repo` when checked-in source hints drift

This closes the gap where an agent can read the package's rich docs or types without ever discovering the Astryx CLI.

## Test plan

- `pnpm lint:strict`
- `pnpm exec vitest run --maxWorkers=4 --reporter=dot`
- packed `@astryxdesign/core` and verified all 227 component docs and 598 declarations contain guidance
- ran all 235 unique generated CLI searches and verified every query returns results

## Try it

```sh
pnpm generate:cli-discovery-hints --check
sed -n '1,8p' packages/core/dist/Button/Button.d.ts
NODE_ENV=development node --input-type=module -e 'import("./packages/core/dist/utils/devWarning.js").then(({devWarn}) => devWarn("Button", "bad prop"))'
```

Expect the declaration to start with `npx @astryxdesign/cli search "Button"`. The warning prints the original message followed by the same CLI lookup.


humbertovirtudes (changes requested):
Before approval, please preserve declaration source-map relationships during stamping, hard-gate generated CLI search terms with deterministic CI, and rebase and revalidate the generated outputs and audit-scope assumptions. The current draft is conflicted and hundreds of commits behind, so these checks need to run against a current base.

PR #5942, opened 2026-09-02
docs: define template authoring architecture

josephfarina:
## Why

Astryx has a detailed template contribution guide and grading rubric, but no repository architecture record connecting source shapes, metadata, validation, catalog membership, previews, and CLI behavior.

Cindy started that record in #5915. This recovers the architecture half without copying the guide or rubric.

## What

- Defines pages and blocks as the two template kinds.
- Defines primary showcases, secondary showcases, and additional examples as block roles, not new kinds.
- Adds the exact required-field matrix for common entries, pages, blocks, and showcase placements.
- Separates human-facing labels from `(package, type, id)` catalog identity.
- Documents today's real source loaders and consumer-specific discovery paths.
- Defines one validated catalog and metadata-driven CLI, `init`, sandbox, docsite, example, and showcase projections as the convergence requirement.
- Records copying as a self-contained snapshot.
- Keeps catalog ids, metadata, and source mutable under `AST-017/DEC-2`, while the surrounding CLI operation and schema remain contractual.
- Delegates contributor recipes and rubric scoring to the guide, visual intent to `design:template-composition`, and compatibility to AST-017.

## Complete shipped-catalog audit

Audited every template at the prior head before rewriting the record:

- 53 page source/metadata pairs. All pass `parseTemplate()`.
- 40 ready, overview-visible pages. `work-item-detail` is the one page missing a live-preview loader.
- 644 block source/metadata pairs with zero orphans.
- 158 primary showcases, 14 secondary-showcase declarations, and 20 additional-example declarations. The 174 expanded showcase targets have no collisions.
- 35 primary showcases lack descriptions. They are legacy metadata gaps, not an intentional showcase exception.
- No duplicate core block ids. Three repeated human-facing names create no CLI ambiguity.
- TypeScript, the runtime parser, and the wiki disagree on required metadata. The record now names that migration explicitly.

## Current state versus target

The shipped system still defaults built-in metadata, bypasses `parseTemplate()` for core pages/blocks, scans page directories for `init`, re-collects integrations in `Project.templates()`, and maintains the docsite page-preview map by hand. The record labels each as a migration gap. It does not claim the convergence work is implemented.

The record stays `authority: draft` until those gaps have executable enforcement and owner approval.

## Validation

- `pnpm exec prettier --write docs/architecture/template-authoring.md`
- `pnpm check:knowledge`
- `pnpm check:repo`
- `pnpm -F @astryxdesign/cli exec tsc --project tsconfig.api-dts.json`
- Four independent corpus audits: pages, blocks/showcases, external/integration templates, and wiki/spec ownership
- Final independent Opus review: `MERGEABLE-AS-DRAFT`; its one P1 and three wording nits were addressed

No Changeset. This is a draft knowledge record and publishes no package change.

### DIFF
--- a/docs/architecture/template-authoring.md
+++ b/docs/architecture/template-authoring.md
@@ -1,0 +1,295 @@
+---
+schema_version: 1
+template_version: 1
+kind: architecture
+id: architecture:template-authoring
+authority: draft
+archive_reason: null
+superseded_by: null
+approved_by: null
+approved_at: null
+owners: [josephfarina, cixzhang]
+applies_to:
+  [
+    packages/cli/assets/templates/,
+    packages/cli/authoring/doctypes/template/,
+    packages/cli/foundation/discovery/template-adapter.mjs,
+    packages/cli/api/template/,
+    apps/docsite/src/components/templateComponents.ts,
+    apps/docsite/scripts/generate-data.mjs,
+    scripts/sync-templates.js,
+  ]
+verified_by:
+  [
+    packages/cli/authoring/doctypes/template/parse.test.mjs,
+    packages/cli/foundation/discovery/template-adapter.test.mjs,
+    packages/cli/foundation/config/project.test.mjs,
+    packages/cli/api/init/init.test.mjs,
+    packages/cli/api/template/template.test.mjs,
+    packages/cli/api/template/template-integration.test.mjs,
+    packages/cli/api/template/template-suffix.test.mjs,
+    packages/cli/clients/cli/e2e-smoke.test.mjs,
+    apps/docsite/src/__tests__/data-extraction.test.ts,
+  ]
+deciding_specs: [spec:AST-017/DEC-2]
+---
+
+# Template authoring architecture
+
+## Purpose
+
+Astryx templates are copyable starting points. Page templates scaffold complete
+content surfaces. Block templates provide focused patterns and component examples.
+This record owns their source shapes, catalog identity, validation boundary,
+consumer projections, and copy semantics.
+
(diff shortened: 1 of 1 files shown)


cixzhang (changes requested):
Template authors currently face inconsistent visibility because docsite live previews are selected by a separate hand-maintained registry. Please make one validated template catalog the membership source, with each consumer using an explicit projection. CLI may expose all eligible pages and blocks; `init` may select scaffoldable pages; docsite may exclude entries through catalog metadata. A valid exclusion such as `scaffold: true` on `blank` must remain supported. What should not remain contractual is a separate manual list whose membership can drift: current main has 49 non-scaffold page templates eligible for the docsite metadata registry but only 41 live previews, missing nine pages. Revise INV2/INV3, system model, change coupling, verification, and OQ1/OQ3 so intentional differences come from declared metadata and manual registration is recorded as a migration gap rather than accepted architecture.

cixzhang:
OQ4 should be answered alongside catalog convergence. Source-specific authoring shapes may remain different, and first-party `TemplateDoc` may stay richer than the integration envelope, but every template should cross one validated normalized-catalog boundary before any CLI, `init`, or docsite projection consumes it. At minimum that boundary should validate required identity/type metadata, source pairing, projection flags such as `scaffold`/overview visibility, and fields each projection relies on. Repository TypeScript checking may remain an additional first-party authoring check; it should not substitute for catalog validation, and missing/invalid source or membership metadata should not be silently defaulted or skipped. Please state whether built-in, external, and integration templates converge on this boundary and record any intentional source-specific differences explicitly.

PR #5943, opened 2026-09-02
fix(docsite): keep example previews within padded frame

bentbrain:
## Summary

- preserve the existing 16px spacing around component examples
- keep oversized examples reachable from their leading edge when they overflow
- stop the surrounding Card's padding variables at the preview boundary so AppShell content is not clipped

## Screenshots

| Before | After |
| --- | --- |
| AppShell example before the fix | AppShell example after the fix |

<details>
<summary>Full-page comparison</summary>

### Before

Full AppShell documentation page before the fix

### After

Full AppShell documentation page after the fix

</details>

### DIFF
--- a/apps/docsite/src/components/component-detail/ExampleBlock.tsx
+++ b/apps/docsite/src/components/component-detail/ExampleBlock.tsx
@@ -3,6 +3,7 @@
 'use client';
 
 import {useEffect, useState, type ComponentType} from 'react';
+import * as stylex from '@stylexjs/stylex';
 import {Card} from '@astryxdesign/core/Card';
 import {Section} from '@astryxdesign/core/Section';
 import {Center} from '@astryxdesign/core/Center';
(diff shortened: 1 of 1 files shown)


bentbrain:
@cixzhang @imdreamrunner Could you confirm the intended layout contract for component-detail live previews? The current Docsite contract (`spec:AST-033`) leaves individual-page layout open, while `architecture:container-padding` defines `overlayPaddingReset` for overlay roots and calls for explicit admission of new participants. This PR applies that reset to a non-overlay preview wrapper.

Is the intended behavior a 16px inset around the example, logical-start alignment when it overflows, and termination of inherited Card padding geometry at that wrapper? If so, could you identify or approve the governing record for this boundary? If not, I can revise the implementation to match the intended behavior.

The deployed PR preview renders `/components/AppShell` and `/components/Table` in both LTR and RTL at a 500px viewport. AppShell fits with a 16px inset. The Table example has a 428px preview and 752px of scrollable content; both scroll extremes are reachable in each direction. The captures are from those actual component-detail routes:

| State | LTR | RTL |
| --- | --- | --- |
| AppShell fitting | AppShell fitting, LTR | AppShell fitting, RTL |
| Table overflow, leading edge | Table overflow at leading edge, LTR | Table overflow at leading edge, RTL |
| Table overflow, scrolled to end | Table overflow at opposite edge, LTR | Table overflow at opposite edge, RTL |

I can add a focused inherited-padding regression once the boundary is settled.

*- written by a robot 🤖 on behalf of liam 🥸*

PR #5948, opened 2026-09-02
fix(ci): prevent opportunistic visual baselines

cixzhang:
## Summary
- compare component PRs only against frame keys already in the accepted baseline
- keep independently selected changed-theme and probe evidence intact
- refuse scoped PR acceptance or promotion when a delta would add or remove a baseline key

## Why
Touching a component currently backfills every missing story/theme frame. For example, #5827 reported 12 added Spinner frames even though no story changed. Ordinary component PRs should validate existing contracts, not silently expand the permanent baseline.

## Impact
This stops future opportunistic growth before the separate baseline-reduction migration. Theme-only evidence remains independently selected, and baseline additions/removals stay available through the deliberate manual baseline workflow.

## Test plan
- 92 focused planner, acceptance, trusted-report, and promotion tests
- `pnpm check:repo`
- `git diff --check`
- replayed #5827 planning against its real Storybook artifact and current baseline

PR #5950, opened 2026-09-03
fix(ci): reduce the visual baseline to canonical frames

cixzhang:
## Summary
- make the release baseline a closed set of representative/tagged **Core Storybook** stories in Neutral plus generated Probe coverage
- exclude `Foundations`, `Lab`, and other cross-group stories even when they import Core
- stop preserving obsolete baseline keys merely because they existed before
- let `keys=all --prune` remove every reviewed obsolete key in one manual baseline run

## Why
The current baseline has 4,332 PNGs using 86.86 MiB. Most are historical story/theme combinations rather than distinct contracts, and they make component changes expensive to review.

## Measured impact
Against current Storybook, the final canonical plan is exactly:
- 246 Neutral frames
- 134 Probe frames
- 380 total frames

A real 380/380 capture produced 5.78 MiB of PNGs. An end-to-end migration simulation promoted all 380 current frames and pruned 4,088 obsolete keys, leaving exactly 380 files plus a 0.74 MiB decision-bearing manifest.

## Stack
Depends on #5948, which prevents ordinary PRs from growing the baseline again.

## Test plan
- 122 focused planner, comparison, trusted acceptance, publication, and promotion tests
- exact current-Storybook planning and 380/380 Chromium capture
- full migration simulation against the live 4,332-shot gh-pages baseline
- adversarial proof that cross-group Core imports cannot enter the canonical baseline
- `actionlint` on `release-gate.yml`
- `pnpm check:repo`
- `git diff --check`


cixzhang (commented):

Release reviewers would lose the closed Core-only baseline: the new release plan still captures stable-looking stories outside the canonical Core surface. I reproduced the intended 4,332→384 migration, but two of the 384 planned keys are `foundations-internationalization-date-consistency--french-and-thai-gregorian__neutral-*` and `lab-resizable--horizontal-split__neutral-*`. Those enter because the package filter trusts resolved Core imports/component paths instead of the Storybook package/group boundary, so a Foundations story importing Core and a Lab-titled Resizable story pointing at Core are treated as release-baseline Core.

Undrafting needs the stable release filter to require the canonical Core story group or another explicit baseline contract, with a regression fixture for these two shapes. Otherwise `keys=all --prune` can promote or prune a release baseline whose authority includes non-canonical stories.

[Reviewed by Robohands]

cixzhang (commented):
Semantic verdict: approve once you undraft

Thanks — this fixes my previous blocker. I reproduced the release plan from the current CI Storybook artifact and gh-pages baseline: 4,332 accepted frames reduce to 380 planned release frames, with 246 Neutral and 134 Probe frames. The prior cross-group leak is gone (`nonCore: []`), so the canonical release set now stays within the configured Core Storybook-group boundary.

I also rechecked the removal path: `keys=all --prune` promoted only the reviewed current frames and pruned only the reviewed removal, while an unreviewed removed key stayed in the baseline. Focused visual-gate tests, `check:repo`, smoke, and CI are green at this head.

[Reviewed by Robohands]

cixzhang (commented):
Semantic verdict: approve

Thanks — final head is clean. I verified it is parented directly on the landed #5948 merge and its patch bytes match the reviewed `6b85fb6` head exactly. Current CI, lint, smoke, and the repo checks are green.

The previous Core-group blocker remains fixed by the identical patch: the release baseline stays on the configured Core Storybook-group boundary, and the reviewed 4,332→380 migration evidence still applies.

[Reviewed by Robohands]

PR #5951, opened 2026-09-03
perf(Tooltip): keep trigger refs stable across rerenders

### LINKED ISSUE #5949: Tooltip callback ref resets on unchanged rerenders
## Problem

`useTooltip()` returns a new callback `ref` after an otherwise unchanged rerender. When a consumer passes that callback to a trigger, React calls the old ref with `null` and the new ref with the element, causing Tooltip to remove and re-add its pointer, mouse, and focus listeners and repeat its anchor wiring.

## Reproduction

```tsx
const {result, rerender} = renderHook(() => useTooltip());
const firstRef = result.current.ref;
rerender();
expect(result.current.ref).toBe(firstRef); // fails on current main
```

## Cause

Tooltip callbacks depend on the object returned by `useTouchTrigger` and on the complete `useLayer` result. Those objects can change while the callback members Tooltip uses remain unchanged, invalidating `interactionRef` and the combined `ref`.

## Expected

Tooltip callback refs remain stable when the behavior that their listeners capture has not changed.

freddymeta:
## Summary

- keep `useTooltip` trigger refs stable when an unchanged render produces new hook result objects
- depend on the specific layer and touch callbacks each handler uses instead of the enclosing objects
- add a failing-first identity regression and RTL coverage for Tooltip's logical placement

Closes #5949.

## Evidence

Before the fix, the new regression failed because `result.current.ref` changed after an unchanged `rerender()`. After the fix, both the combined ref and interaction ref retain their identities.

The Tooltip RTL audit now measures logical `start` placement and reports:

```
```

## Before vs after calculation

Measured with the same Tooltip wrapper on the parent commit and this PR, across 100 semantically unchanged rerenders. The harness spies on the actual trigger element and counts only Tooltip-owned event types.

| DOM wiring operation | Before | After |
|---|---:|---:|
| listener detach | 600 | 0 |
| listener attach | 600 | 0 |
| `aria-describedby` remove | 100 | 0 |
| `aria-describedby` set | 100 | 0 |
| **measured total** | **1,400** | **0** |

That is **14 measured DOM calls per rerender → 0 (100% reduction)**. The callback ref also removes and restores the CSS `anchor-name`, so the full known rewiring is **16 DOM mutations/calls per rerender → 0**.

This is a deterministic operation count, not a wall-clock claim; the short Vitest/jsdom run is dominated by harness overhead.

## Test plan

- `pnpm exec vitest run packages/core/src/Tooltip/Tooltip.test.tsx` — 25 passed
- `pnpm lint:strict` — passed with no errors
- `pnpm build` — passed
- `~/projects/astryx-preflight.sh <worktree> Tooltip` — passed: typecheck, lint, 8,695 core tests, core + Storybook build, RTL audit, Chromium guards, and repository checks

Full-workspace `pnpm test` was also run twice. Each run passed over 14,000 tests and failed only the ContextMenu BottomSheet timing cases, which pass as a 48-test isolated file. An untouched `origin/main` control also fails the full suite on the existing `score-ledger` timeout assertion.

## Visual verification

No production rendering changed. The added default-open RTL story verifies the existing logical placement in Chromium; the scoped RTL audit passes.

### DIFF
--- a/.changeset/keep-tooltip-refs-stable-across-rerenders-ce7j.md
+++ b/.changeset/keep-tooltip-refs-stable-across-rerenders-ce7j.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[perf] Keep Tooltip refs stable across rerenders
+@freddymeta
--- a/apps/storybook/rtl-audit/targets.json
+++ b/apps/storybook/rtl-audit/targets.json
@@ -2,7 +2,9 @@
   {
     "component": "AppShell",
     "storyId": "core-appshell--top-nav-with-side-nav",
-    "dims": ["D4"],
+    "dims": [
+      "D4"
+    ],
     "selectors": {
       "overlay": ".astryx-app-shell-sidenav",
       "overlayRoot": ".astryx-app-shell"
@@ -11,7 +13,9 @@
   {
     "component": "ButtonGroup",
     "storyId": "core-buttongroup--horizontal",
-    "dims": ["D2"],
+    "dims": [
+      "D2"
+    ],
     "selectors": {
       "prev": "[role=\"group\"] button:first-child",
       "next": "[role=\"group\"] button:last-child"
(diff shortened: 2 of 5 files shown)


cixzhang (approved):
Thanks, this keeps Tooltip’s trigger wiring stable without changing its API or interaction behavior. The focused red→green regression and option-change paths pass.

[Reviewed by Robohands]

PR #5953, opened 2026-09-03
chore: synchronize readonly probe selectors

cixzhang:
## Why

The generated probe theme is stale for the existing `readonly` states added to Selector and MultiSelector in #5805. That causes a clean `visual:probe-theme:check` to fail and forces unrelated component PRs to absorb these selectors.

## What

Regenerates only the existing `readonly` selector entries. No source, docs, specification, or runtime behavior changes.

## Validation

- `pnpm visual:probe-theme`
- `pnpm check:repo`
- `git diff --check`

No Changeset: generated test-fixture synchronization only.

### DIFF
--- a/packages/themes/probe/src/probeTheme.ts
+++ b/packages/themes/probe/src/probeTheme.ts
@@ -6,7 +6,7 @@
 // test fixture. Regenerate with: pnpm visual:probe-theme
 //
 // defineTheme takes six things and this covers all six:
-//   components  274 targets, 883 selectors (generated from the docs)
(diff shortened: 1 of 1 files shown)


cixzhang (commented):
Semantic verdict: approve

This restores the two generated selectors already required by the current readonly contracts, so unrelated component changes no longer inherit a stale probe failure.

[Reviewed by Robohands]

cixzhang (commented):
Semantic verdict: approve

Thanks, this is good. The regenerated probe covers the current readonly state selectors and the fixture check is clean at this head.

[Reviewed by Robohands]

PR #5956, opened 2026-09-03
perf(Typeahead): skip synchronous bootstrap loading cycle

### LINKED ISSUE #5955: Typeahead runs a loading cycle for synchronous bootstrap sources
## Problem

`BaseTypeahead` accepts synchronous and asynchronous `SearchSource.bootstrap()` implementations, but currently enters the loading state before it knows which kind it received. A synchronous bootstrap therefore pays the asynchronous lifecycle: loading is committed, then the immediately available result is committed. An empty synchronous bootstrap performs two React commits even though it has no entries to render.

## Reproduction

1. Render `BaseTypeahead` with `hasEntriesOnFocus` and `bootstrap: () => []`.
2. Wrap it in `React.Profiler`.
3. Clear the mount samples and focus the input.
4. Observe two update commits.

The same source with `hasEntriesOnFocus={false}` produces no focus-time commits.

## Expected behavior

Synchronous bootstrap sources should be applied synchronously without entering the loading state. In particular, an empty synchronous bootstrap should be a render no-op. Asynchronous bootstrap sources should keep the existing loading behavior.

## Scope

This is an implementation optimization for the existing `SearchSource` contract; it does not change the public API, DOM, styling, selection behavior, or async loading semantics.

freddymeta:
## Summary

Fixes #5955.

`SearchSource.bootstrap()` supports both arrays and promises, but `BaseTypeahead` previously entered the asynchronous loading state before it knew which kind it received. A synchronous source therefore committed loading on and off even when it immediately returned no entries.

This change:

- applies synchronous bootstrap results without entering loading
- makes an empty synchronous bootstrap a render no-op
- preserves loading for promise-backed bootstrap sources
- clears a superseded search's loading state when an empty query switches to synchronous bootstrap

No public API, DOM, styling, selection, or asynchronous loading behavior changes.

## Verification

- Failing-first Profiler regression: empty synchronous bootstrap made 2 update commits before the fix and 0 after
- `pnpm exec vitest run packages/core/src/Typeahead/Typeahead.test.tsx` — 79 passed
- `pnpm check:repo`
- `~/projects/astryx-preflight.sh <worktree> Typeahead` — green
  - 8,699 core tests passed
  - Storybook build passed
  - `pr-rtl`: 1 measured / 0 gap
  - Chromium accessibility guards passed

### DIFF
--- a/.changeset/typeahead-synchronous-bootstrap.md
+++ b/.changeset/typeahead-synchronous-bootstrap.md
@@ -1,0 +1,9 @@
+---
+'@astryxdesign/core': patch
+---
+
+[perf] Typeahead: skip the loading cycle for synchronous bootstrap sources (#5955)
+
+`BaseTypeahead` now applies an array returned by `SearchSource.bootstrap()` immediately instead of entering and leaving the asynchronous loading state. An empty synchronous bootstrap becomes a render no-op, while synchronous entries still open normally. Switching from an in-flight search to a synchronous bootstrap also clears the superseded search's loading state. Promise-backed bootstrap sources keep the existing loading behavior.
+
+@freddymeta
--- a/packages/core/src/Typeahead/BaseTypeahead.tsx
+++ b/packages/core/src/Typeahead/BaseTypeahead.tsx
@@ -592,33 +592,63 @@ export const BaseTypeahead = function BaseTypeahead<T extends SearchableItem>({
     ],
   );
 
-  // Perform bootstrap
-  const performBootstrap = useCallback(async () => {
-    const gen = ++searchGenRef.current;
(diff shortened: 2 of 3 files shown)


cixzhang (approved):
Thanks, this is good. The failing-first Profiler check proves empty synchronous bootstrap goes from two commits to zero while async loading stays intact.

[Reviewed by Robohands]

freddymeta:
All code and browser checks are green, and the change has no intended visual delta. Could a maintainer clear the remaining gate with:

`/accept-visual 33705563566/1 No visual change: this only avoids transient loading commits for synchronous bootstrap sources; DOM and styling are unchanged.`

PR #5952, opened 2026-09-03
feat(useResizable): support CSS min()/max() constraints

freddymeta:
## Summary

- add the Resizable-specific `ResizableConstraintValue` for flat `minSize` / `maxSize`
- accept non-negative numbers, exact `px` / `%` leaves, and recursive CSS `min()` / `max()` expressions (maximum depth 8)
- reject `calc()`, `clamp()`, arithmetic, `var()`, other units, malformed input, and deeper nesting through the existing deterministic fallback + development-warning path
- detect percentage dependency recursively so bounds re-resolve and re-clamp on container or viewport basis changes
- preserve released `defaultSize`, pixel-only `resize()`, and deprecated `minSizePx` / `maxSizePx` compatibility
- document and test the existing maximum-wins rule when resolved minimum exceeds maximum
- amend AST-010, update Resizable docs and Storybook, and add a core changeset

## Stack

Depends on #5783. This PR intentionally targets `fix/resizable-percentage-sizes`; its parent at implementation and latest push time is `57f6e816b743610e218279fe0749d5cc1f5b900a`.

Astryx's PR workflows intentionally filter on base `main`: CI and Lint use `pull_request.branches: [main]`; the review and spec-owner workflows use `pull_request_target.branches: [main]`. Therefore this feature-branch-based draft has no GitHub Actions runs while stacked. Do not interpret their absence as a green CI result. The current external checks are Meta CLA plus Vercel.

## Browser evidence

Built Storybook from `a67d677def7a8b6dabb6573acd57487bf62efbd0` and measured `Core/Hooks/useResizable / Css Math Constraints` in Chromium. The requested 1000px / 500px bordered frames produced 998px / 498px content-box bases.

| constraint | 998px content box | 498px content box |
| --- | --- | --- |
| `minSize: 'max(40%, 333px)'` | bound 399px; selected / panel / ARIA 399px | bound 333px; selected / panel / ARIA 333px |
| `maxSize: 'min(400px, 10%)'` | bound 100px; selected / panel / ARIA 100px | bound 50px; selected / panel / ARIA 50px |

No browser console errors or page errors.

## Verification

- `pnpm lint:strict` — pass, 0 errors (85 repository warnings)
- `pnpm exec vitest run packages/core/src/Resizable/useResizable.test.ts` — 95 passed
- `pnpm -F @astryxdesign/core build` — pass
- `pnpm -F @astryxdesign/core typecheck` — pass
- `pnpm -F @astryxdesign/core typecheck:docs` — pass
[...]

PR #5957, opened 2026-09-03
feat(useResizable): minSize/maxSize accept min() and max()

freddymeta:
Stacked on #5783. **Base is `fix/resizable-percentage-sizes`, not `main`** — review that one first; this diff is only the files below.

A bound could be proportional or fixed, never both. `max(40%, 333px)` — 40% of the container, but never below 333px — had no spelling. The string failed to parse whole and took the invalid-input fallback.

## Measured at the parent commit

| config | container | resolved |
|---|---|---|
| `minSize: 'max(40%, 333px)'` | 1200 → 400 | **50px at every width** |
| `maxSize: 'min(400px, 10%)'` | 1200 / 4000 / 6000 | **`Infinity` at every width** |

Development warned. `devWarn` is a no-op in production (`utils/devWarning.ts`), so a shipped build got the wrong geometry with no signal at all.

## After

`minSize` and `maxSize` accept `min()`/`max()` over the terms they already accepted — two terms, one level of nesting. Terms resolve against the same basis as a plain percentage and are compared as resolved pixels, so which arm wins changes with the container, exactly as in CSS.

Chromium 149, container driven at runtime, every row checked against an independent CSS reference evaluation in the harness:

| container | `minSize: 'max(40%, 333px)'` | `maxSize: 'min(400px, 10%)'` |
|---|---|---|
| 1600 | **640** | — |
| 1200 | **480** | **120** |
| 900 | **360** | — |
| 833 (crossover) | **333** | — |
| 832 | **333** | — |
| 400 | **333** | — |
| 2000 / 3999 / 4000 / 6000 | — | 200 / **400** / 400 / 400 |

`aria-valuemin`/`aria-valuemax` carry the same numbers on every row. Nested `max(20%, min(500px, 60%))` measured 600 / 500 / 360 / 120 at 3000 / 1000 / 600 / 200.

## Basis-dependency recurses

A percentage at any depth makes the whole bound track its container; an all-pixel expression is static and observes nothing. The test is on the **parsed tree**, not the source text — a syntactic check would classify `max(40%, 333px)` as static, the container would never be observed, and the bound would resolve once and never move again. Supported-looking and quietly broken. Covered both ways: `max(80px, min(90px, 50%))` re-resolves 90 → 80 when the container shrinks to 100px; `min(400px, 900px)` registers no observer.

## Typing

`minSize`/`maxSize` move from the shared `SizeValue` to a Resizable-owned `ResizableSize`. `'40vw'`, `'20rem'`, `'calc(100% - 3rem)'`, `'clamp(...)'`, `'var(--w)'`, an unbalanced call and over-deep nesting are now **compile errors** rather than runtime fallbacks. Asserted in `useResizable.public.test.ts` with `expectTypeOf`, and I mutation-tested those assertions: claiming `'banana'` is valid, or that the canonical example is not, both fail `tsc`.

The two-term/one-level limits are not arbitrary. They are what keeps the union representable — a wider grammar raised `TS2590` ("union type that is too complex to represent") exactly where a bound meets its deprecated pixel alias. The parser enforces the same two limits, so the compiler and the runtime accept the same set of strings. Compile cost is unchanged: 21.48s vs 21.39s on the parent.

The type does not replace the parser, which stays authoritative for what a type cannot state — a percentage above 100, a negative, a non-finite.

## Compatibility

- `defaultSize` keeps its released `number | string`. Narrowing a shipped prop would reject callers passing a computed string. It resolves an expression once, by the same parser, exactly as it already resolves `'50%'` once.
- `minSizePx`/`maxSizePx` still work, and **one bound can still migrate while the other keeps its alias** — `{minSize: 'max(40%, 333px)', maxSizePx: 960}` type-checks. A grouped `{min, max}` shape would have forbidden that; rejected for it, and because every bounded pair in the package is flat sibling props.
- **Maximum-wins is unchanged and now pinned.** When a resolved minimum exceeds its maximum the released `Math.min(max, Math.max(min, size))` order stands. Verified identical in `main` before touching it, and there is a test with an expression floor: a 400px container with `minSize: 'max(40%, 333px)'` and `maxSize: '50%'` gives min 333 / max 200 / size 200, finite.

## Gates

```
core build        ✅ 598 dist file(s) checked
vitest            Test Files 319 passed · Tests 9134 passed
typecheck         exit 0
typecheck:docs    exit 0
check:repo        exit 0 (incl. check:knowledge — spec validates)
lint:strict       ✖ 85 problems (0 errors, 85 warnings) — identical to the parent
```

Negative control: reverting the implementation and keeping the tests fails **17** runtime tests; reverting the types fails `tsc` on the type-test file.

## Spec

AST-010 gains **FR13** and **DEC-3**, plus rows in the invalid-configuration and verification tables. `calc()`, arithmetic, `clamp()`, `var()` and other units stay non-goals — the grammar is closed, so an unrecognized function is invalid input rather than a passthrough to CSS. DEC-3 records the rejected alternatives, including the structured-AST shape, with the measured catch rates that decided it.

### DIFF
--- a/.changeset/resizable-minmax-bound-expressions.md
+++ b/.changeset/resizable-minmax-bound-expressions.md
@@ -1,0 +1,29 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] useResizable: `minSize`/`maxSize` accept `min()` and `max()`
+
+A bound could be proportional or fixed, never both. `max(40%, 333px)` — 40% of the container, but never below 333px — had no spelling: the whole string failed to parse and took the invalid-input fallback. Measured in Chromium before this change, `minSize: 'max(40%, 333px)'` resolved to **50px at every container width from 1200px down to 400px**, and `maxSize: 'min(400px, 10%)'` to `Infinity` at 1200/4000/6000px. Development warned; `devWarn` is a no-op in production, so a shipped build got the wrong geometry silently.
+
+`minSize` and `maxSize` now take `min()`/`max()` over the terms they already took — two terms, one level of nesting. Terms resolve against the same basis as a plain percentage and are compared as resolved pixels, so which arm wins changes with the container, exactly as in CSS. After, same widths:
+
+| container | `minSize: 'max(40%, 333px)'` | `maxSize: 'min(400px, 10%)'` |
+|---|---|---|
+| 1600 / 6000 | **640** | **400** |
+| 1200 | **480** | 120 |
+| 900 | **360** | — |
+| 833 (crossover) | **333** | — |
+| 400 | **333** | — |
+
+Everything the percentage path already guarantees applies unchanged: the bound re-resolves when the container resizes and clamps the selection rather than rescaling it, it holds one frozen basis for the duration of a drag, and state, paint, persistence and `aria-valuemin`/`aria-valuemax` stay one geometry.
+
+Basis-dependency is decided on the parsed expression, not the source text. A percentage at any depth — `max(80px, min(90px, 50%))` — makes the whole bound track its container. An all-pixel expression is static and observes nothing.
+
+`minSize` and `maxSize` are now typed by a Resizable-owned `ResizableSize` instead of the shared `SizeValue`. `'40vw'`, `'20rem'`, `'calc(100% - 3rem)'`, `'clamp(...)'`, `'var(--w)'`, an unbalanced call and over-deep nesting are compile errors rather than runtime fallbacks. The runtime parser is unchanged in authority — a type cannot say "0 through 100" or "finite" — and still repairs anything invalid to the documented 50px / `Infinity` without disturbing a persisted size.
+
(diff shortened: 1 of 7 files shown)


freddymeta:
**No CI will run on this PR while it is stacked.** Every gating workflow filters its `pull_request` trigger to `branches: ['main']` — `ci.yml`, `lint.yml`, `cli-smoke-test.yml`, `codemod-verify.yml`, `internal-registry.yml`, `hardening-issue.yml`. This PR's base is `fix/resizable-percentage-sizes`, so GitHub created **0 workflow runs** for `1bc1992a9e` (the parent's head has 7). Only Vercel and the CLA check, which are not Actions workflows, reported.

That resolves itself when #5783 merges: GitHub retargets this PR to `main` and the full suite runs. Until then the gates were run locally on this exact tree:

```
core build        ✅ 598 dist file(s) checked
vitest            Test Files 319 passed · Tests 9134 passed
typecheck         exit 0
typecheck:docs    exit 0
check:repo        exit 0 (includes check:knowledge)
lint:strict       ✖ 85 problems (0 errors, 85 warnings) — identical to the parent
```

I have not repointed the base to `main`, because that would pull the parent's two commits into this PR and make it unreviewable on its own. Happy to do that instead if you would rather have CI now than a clean stack.

freddymeta:
Superseded by #5952, which carries the same CSS min()/max() constraint capability with the broader browser, performance, and regression coverage. The useful `defaultSize` behaviour from this version will be folded into #5952.

Issue #5902, opened 2026-09-02
[Bug] PowerSearch Properties preview renders no working component


cixzhang:
## Affected page

https://astryx.atmeta.com/components/PowerSearch?tab=properties

## Steps to reproduce

1. Open the affected component page in Chromium 149.0.7827.55 on macOS.
2. Open the **Properties** tab.
3. Observe the preview before changing any props.
4. Exercise the available controls where applicable.

## Expected behavior

A usable PowerSearch input renders with a small representative field configuration.

## Actual behavior

The preview is replaced by “Interactive preview needs required props that cannot be generated automatically. Missing: config, filters.”

No unrelated console or page errors were observed during the fresh-context reproduction.

## Diagnosis

The required custom `config` and `filters` values have no playground defaults; the generic preview cannot synthesize them.

The likely fix is local to `packages/core/src/PowerSearch/PowerSearch.doc.mjs` and/or its focused docsite preview coverage; it does not require a new public API or design direction. This is an atomic, independently closable slice of #2008.

## Acceptance criteria

- [ ] Add a minimal representative PowerSearch config and initial filters fixture.
- [ ] The preview renders on first load instead of the missing-props placeholder.
- [ ] A basic filter interaction updates the rendered state without errors.
- [ ] Add focused preview/data-extraction coverage.

HelloOjasMutreja:
Already fixed: PR #5964 (linked via a body-text "Related Issue" reference rather than "Fixes #5902", so it never auto-linked here) merged this on 2026-09-03. Verified against a fresh docsite build on current main: the Properties tab now renders a real, interactive PowerSearch with a "Status: is Open" token on first load, matching all four acceptance criteria. Nothing left to do here; closing would be accurate if a maintainer wants to.

PR #5964, opened 2026-09-03
fix(docsite): seed PowerSearch properties preview

Lee-Dongwook:
## Related Issue

https://github.com/facebook/astryx/issues/5902

## Summary

Fixes the PowerSearch Properties preview, which previously showed a missing-required-props placeholder because it could not generate the required config and filters values.

The preview now opens with a small, representative field configuration and an initial Status = Open filter.

## Changes

- Add serializable playground defaults for PowerSearch config and filters.
- Provide Status and Title fields with valid operators and values.
- Seed the preview with a Status = Open filter token.
- Reuse the existing controlled onChange(filters, ...) preview bridge so filter interactions update rendered state.
- Add focused regression coverage for preview state and generated docsite data.

## Tests

Added coverage for:

- Required config and filters defaults preventing the missing-props placeholder.
- Initial filter fixture rendering state.
- Controlled filter change propagation through onChange.
- Generated component registry extraction of the PowerSearch fixture.

Validated locally:

- PowerSearch doc fixture module import.
- Knowledge-record validation.
- Whitespace diff validation.

### Screenshots

스크린샷 2026-09-03 오후 9 13 43

### DIFF
--- a/apps/docsite/src/__tests__/component-preview-state.test.ts
+++ b/apps/docsite/src/__tests__/component-preview-state.test.ts
@@ -209,6 +209,75 @@ describe('component detail preview state', () => {
     expect(getMissingRequiredProps(knobs, state)).toEqual([]);
   });
 
+  it('gives PowerSearch a renderable controlled filter fixture', () => {
+    const knobs = pickPrimaryProps('PowerSearch', [
+      prop({name: 'config', type: 'PowerSearchConfig', required: true}),
+      prop({
+        name: 'filters',
+        type: 'ReadonlyArray<PowerSearchFilter>',
+        required: true,
+      }),
+      prop({
+        name: 'onChange',
+        type: "(filters: ReadonlyArray<PowerSearchFilter>, changeType: 'add' | 'edit' | 'remove', index: number) => void",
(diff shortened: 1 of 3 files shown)


Lee-Dongwook:
Hello! The pr-rtl job reports a coverage gap for PowerSearch:
 
  > AUTO: 0 pass / 0 fail / 1 N-A
  > PM: 0 pass / 0 fail / 26 N-A
  > DEC: 0 pass / 0 fail / 26 N-A
  > COV: 0 measured / 0 verified N-A / 1 gap
 
It appears that PowerSearch has no applicable RTL coverage under the current automatic checks, and it is not listed in verified-not-applicable.json.

Could you advise on the preferred resolution? Should I add a verified N/A declaration after reviewing the component and all of its stories, or should PowerSearch instead have a curated RTL target added to targets.json for its directional popover/icon behavior?

cixzhang (approved):
Thanks, this is good. The fixture matches PowerSearch’s documented config shape, renders the initial token, and keeps filter changes controlled.

[Reviewed by Robohands]

PR #5970, opened 2026-09-03
fix(release): classify template slug rename as nonbreaking

cixzhang:
## Why

AST-017 classifies template slugs as mutable catalog data. A slug rename is nonbreaking even when the previous exact-match value stops resolving; the stable contract is the surrounding CLI command and response schema.

## What

Reclassifies the dashboard template rename changeset from `[breaking]`/minor to `[fix]`/patch and makes the release note point builders to the new catalog values.

## Risk

Release metadata only; runtime behavior is unchanged.

## Testing

- `pnpm check:changesets` — 56 changesets valid
- `git diff --check`


cixzhang (commented):
Semantic verdict: approve

This matches the current compatibility contract: the five replacement slugs stay discoverable, while the catalog-only rename correctly ships as a patch instead of a breaking CLI change.

[Reviewed by Robohands]

cixzhang (commented):
Semantic verdict: approve

My prior-head review approved the patch classification; this update preserves that result and gives builders the complete old-to-new catalog mapping.

[Reviewed by Robohands]

PR #5973, opened 2026-09-03
Add shadcn Registry compatibility

josephfarina:
## Summary

- generate a standard shadcn Registry projection from the existing Astryx component and template catalogs
- keep components package-backed while copying editable showcases, examples, blocks, and pages
- add secondary per-page install commands, a compatibility guide, and a draft blog post

## Architecture

The existing `.doc.mjs` metadata, sibling composition source, and owning `package.json` remain the source of truth. The docsite generator emits a static registry index and one JSON item per installable surface. There is no parallel hand-authored catalog.

The normal Astryx component and template documentation remains the human browse experience. Applicable pages show a secondary shadcn command after their primary documentation. Raw machine routes are organized under `/shadcn/components`, `/shadcn/hooks`, `/shadcn/showcases/<component>`, `/shadcn/examples/<component>`, `/shadcn/blocks`, and `/shadcn/templates`. `exampleFor` is optional. Unowned compositions are standalone blocks, while component-owned records become examples or showcases. IDs derive from stable doc/catalog identity, with optional `registry.slug`, compatibility aliases, collision checks, and a reviewed route lock.

Component items install the owning Astryx package and write a public re-export. Composition items write editable app code that imports published Astryx package paths. Fourteen compositions with local StyleX declarations are precompiled to compiler-free JSX during generation.

## Launch boundary

The compatibility foundation is approved and preview builds emit the full registry from their own `/shadcn` origin with `@canary` dependencies. The production `latest` target stays disabled until copied-composition upgrades and public support ownership are ready. `/r` remains reserved for a future Astryx-native protocol.

## Validation

- 968 generated items: 216 components, 54 hooks, 160 showcases, 486 component examples, 0 standalone blocks, and 52 pages
- every prior locked route remains available as a canonical path or alias
- all items validate against the pinned shadcn 4.19.0 schema
- real shadcn 4.19.0 install from an arbitrary nested `/shadcn/...` URL with a full-URL `registryDependencies` item
- prior whole-catalog proof installed all 921 entries into a clean Vite app and built 6,620 modules
- workspace build and full docsite production build
- CLI typecheck and 3,577 CLI tests
- docsite typecheck and 495 docsite tests
- 30 focused registry and authoring contract tests
- strict lint, repository checks, and knowledge checks

### DIFF
--- a/.changeset/experimental-shadcn-registry.md
+++ b/.changeset/experimental-shadcn-registry.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/cli': patch
+---
+
+[feat] Add an experimental shadcn Registry compatibility guide and doc-derived registry identity metadata. It explains the package boundary, stable organized paths, copied composition model, upgrade behavior, and when to use the richer Astryx CLI.
+
+@josephfarina
--- a/.gitignore
+++ b/.gitignore
@@ -80,6 +80,9 @@ yarn.lock
 .tool-versions
 mise.toml
 
+# Generated shadcn Registry output (rebuilt by the docsite generate step)
+apps/docsite/public/shadcn/
+
 # Generated i18n artifacts (rebuilt from en.json)
 packages/core/locales/pseudo.json
 
--- a/apps/docsite/package.json
+++ b/apps/docsite/package.json
@@ -43,6 +43,7 @@
     "zod": "^4.4.3"
   },
   "devDependencies": {
+    "@babel/core": "^7.29.7",
     "@babel/preset-react": "^8.0.1",
     "@babel/preset-typescript": "^7.29.7",
     "@stylexjs/babel-plugin": "catalog:",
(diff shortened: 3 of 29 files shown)


cixzhang:
Two cleanup items are behind the current docsite failure:

- `docsite-test` detects generated shadcn names/routes that differ from the reviewed route lock. Please preserve prior paths with `doc.registry.aliases` when a route moved unintentionally; refresh the reviewed lock only for routes this experiment intentionally adds.
- This draft currently uses `spec:AST-018`, which is already the palette-authoring specification. Please renumber the shadcn Registry experiment to an unoccupied spec ID and update its references consistently.

I’m treating these as PR comments, not a request-changes review.

cixzhang:
Correction after checking the current head `070164d2`: both items from my previous comment are already addressed here.

- The shadcn route lock is committed, and `docsite-test` now passes.
- The experiment was renumbered from the conflicting AST-018 to `spec:AST-026`.

The previous comment described the older failing head. No request-changes review was submitted.

PR #5974, opened 2026-09-03
docs(spec): define explicit palette color adoption

cixzhang:
## Why

PR #5800 coupled palette metadata to automatic `defineTheme.color` expansion. That would let a later palette edit silently recolor a theme.

This redraft makes palettes authoring references instead: tools may suggest exact colors, but themes ship only values an author explicitly reviews and saves.

## What

- Defines Reference → Suggest → Save → Verify as four separate operations.
- Keeps `color` behavior unchanged whether palette data exists or not.
- Requires palette-only edits to leave normalized tokens, CSS, built modules, and runtime behavior unchanged.
- Preserves valid out-of-palette colors and contextual accessibility review.
- Gives palette metadata and future generator recipes separate owner contracts; neither may bypass explicit save and verification.

## Scope

Specification only. No runtime, build, theme, or package behavior changes. No Changeset.

This supersedes the direction in closed PR #5800. It does not approve the metadata API in #5668 or select a generator recipe.

## Validation

- Pending exact-head knowledge validation and review.

### DIFF
--- a/docs/specs/AST-008/spec.md
+++ b/docs/specs/AST-008/spec.md
@@ -1,0 +1,192 @@
+---
+schema_version: 1
+template_version: 1
+kind: system-spec
+id: spec:AST-008
+authority: draft
+archive_reason: null
+superseded_by: null
+approved_by: null
+approved_at: null
+phase: proposed
+owners: [cixzhang, rubyycheung, imdreamrunner]
+affects_architecture:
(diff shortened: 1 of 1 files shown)


cixzhang:
Closing as a duplicate AST-008 redraft. The authoring lifecycle and future implementation prerequisites are being consolidated into the revived original PR #5800.

PR #5975, opened 2026-09-03
docs(LayoutFooter): seed playground preview fixture (#5895)

### LINKED ISSUE #5895: [Bug] LayoutFooter Properties preview renders no working component
## Affected page

https://astryx.atmeta.com/components/LayoutFooter?tab=properties

## Steps to reproduce

## Expected behavior

A visible footer region renders in a minimal Layout scaffold.

## Actual behavior

The preview stage is completely empty except for the Show code control.

## Diagnosis

The subcomponent has no child content or parent-layout playground wrapper, so its standalone region is empty.

The likely fix is local to `packages/core/src/Layout/LayoutFooter.doc.mjs` and/or its focused docsite preview coverage; it does not require a new public API or design direction. This is an atomic, independently closable slice of #2008.

## Acceptance criteria

- [ ] Add representative footer content and a minimal Layout playground wrapper.
- [ ] The preview renders a visible footer on first load.
- [ ] A relevant footer control such as divider or height produces an observable update.

Rijul202:
Seeds a playground preview fixture for **LayoutFooter** so the docsite properties-tab preview renders a working footer instead of an empty landmark bar on first load.

- Adds `playground.defaults` (footer text + `hasDivider`) and a `Layout` parent wrapper, mirroring the sibling `LayoutHeader` fixture.
- Docsite tests cover the non-empty preview state (`component-preview-state`) and the registry shape (`data-extraction`).

Rebased on latest `main` (clears the earlier conflict). Per @cixzhang's review, this PR is now **scoped to LayoutFooter only** — the DropdownMenuRadioGroup half is dropped here so the radio-item accessibility fix can converge in #5976.

Closes #5895.

### DIFF
--- a/.changeset/layoutfooter-properties-preview.md
+++ b/.changeset/layoutfooter-properties-preview.md
@@ -2,4 +2,8 @@
 '@astryxdesign/core': patch
 ---
 
-[docs] LayoutFooter: seed example content via playground defaults (footer text plus a Layout parent wrapper) so the docsite properties-tab preview renders a working footer instead of an empty stage. (#5895)
+[docs] LayoutFooter: seed playground defaults and a Layout footer-slot wrapper for the docsite preview (#5895)
+
+Prevents the properties-tab preview on the docsite from rendering an empty stage by seeding representative footer content and mounting LayoutFooter in the Layout wrapper's footer slot, so the preview shows a docked footer instead of a centred label.
+
+@Rijul202
--- a/.changeset/layoutfooter-properties-preview.md
+++ b/.changeset/layoutfooter-properties-preview.md
@@ -1,0 +1,5 @@
+---
+'@astryxdesign/core': patch
+---
+
+[docs] LayoutFooter: seed example content via playground defaults (footer text plus a Layout parent wrapper) so the docsite properties-tab preview renders a working footer instead of an empty stage. (#5895)
--- a/apps/docsite/src/__tests__/component-preview-state.test.ts
+++ b/apps/docsite/src/__tests__/component-preview-state.test.ts
@@ -220,6 +220,22 @@ describe('component detail preview state', () => {
     },
   );
 
+  it('seeds LayoutFooter children from playground defaults so the preview is not empty (#5895)', () => {
+    const knobs = pickPrimaryProps('LayoutFooter', [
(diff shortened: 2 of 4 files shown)


cixzhang (changes requested):
Thanks, LayoutFooter is good. The radio-group half overlaps #5976; this version renders `menuitemradio` rows outside `DropdownMenu`, so keyboard navigation is absent and every item has `tabindex=-1`. Could we keep this PR to LayoutFooter and let the scaffolded radio fix converge in #5976?

[Reviewed by Robohands]

Rijul202:
Thanks @cixzhang — done. I've dropped the DropdownMenuRadioGroup changes entirely and rebased on `main`, so this PR is now LayoutFooter-only. Agreed the radio-item fix (keyboard nav / `tabindex`) belongs with the scaffolded work in #5976 rather than rendering `menuitemradio` rows outside a `DropdownMenu` here. Ready for another look.

cixzhang (changes requested):
Thanks for narrowing this to LayoutFooter—the radio work now cleanly converges in [#5976](https://github.com/facebook/astryx/pull/5976). Builders still see a centered label instead of a footer because the wrapper uses Layout's content slot. Please route it through the footer slot, assert that in the registry test, and add the changeset contributor credit required by lint.

[Reviewed by Robohands]

Rijul202:
Thanks @cixzhang — all three addressed:

- **Footer slot:** the wrapper now sets `slotProp: 'footer'`, so LayoutFooter mounts in Layout's footer region instead of the default content slot — the preview shows a docked footer rather than a centred label.
- **Registry assertion:** `data-extraction` now asserts `playground.wrapper` carries `slotProp: 'footer'`.
- **Changeset:** added the contributor credit line for lint.

Ready for another look.

cixzhang (changes requested):
Builders still see the same centered label, not a docked footer: the prior and current deployments are pixel-identical, with the 140×25 footer still 87.5px above the stage bottom. The slot metadata, registry assertion, scope, and contributor credit are fixed, but they don't change rendered geometry. Please make the Layout scaffold visibly exercise its footer region and add rendered-preview coverage that fails on this unchanged result.

[Reviewed by Robohands]

PR #5976, opened 2026-09-03
fix(docsite): show DropdownMenuRadioGroup in properties preview

### LINKED ISSUE #5888: [Bug] DropdownMenuRadioGroup Properties preview renders no working component
## Affected page

https://astryx.atmeta.com/components/DropdownMenuRadioGroup?tab=properties

## Steps to reproduce

## Expected behavior

A radio group with selectable radio items renders immediately.

## Actual behavior

The preview stage is completely empty except for the Show code control; changing available controls still leaves it empty.

## Diagnosis

The doc provides neither radio-item children nor a preview fixture, so the group renders an empty container.

The likely fix is local to `packages/core/src/DropdownMenu/DropdownMenuRadioGroup.doc.mjs` and/or its focused docsite preview coverage; it does not require a new public API or design direction. This is an atomic, independently closable slice of #2008.

## Acceptance criteria

- [ ] Provide minimal radio-item children and any required menu scaffold through the component playground config.
- [ ] The preview shows at least two radio choices on first load.
- [ ] Selecting a choice visibly updates the selected state without console or page errors.

Kyujenius:
## Summary

- Give DropdownMenuRadioGroup a playground wrapper (`DropdownMenu` with a `Sort` trigger, mounted open) and two `DropdownMenuRadioItem` children, so the Properties preview shows both choices on first load instead of an empty container.
- Bridge the wrapper's `onOpenChange` back into preview state so selecting still closes the menu and the trigger reopens it.
- Stop `DropdownMenu` from moving focus into a controlled menu that mounts already open, so the preview (or any page that renders a menu open on load) does not drop keyboard users into it.
- Cover the generated component registry, the preview's render-time first-load state, and the menu's mount-open focus rule.

Fixes #5888

## Notes

- The trigger label stays `Sort` after selection by design. DropdownMenu reads only `button.label`, and the selected state is shown by the radio mark inside the menu, matching the existing storybook story.
- The focus change lives in core, not the preview: `DropdownMenu`'s compound path treated a controlled `isMenuOpen` that is true at mount like a programmatic open and focused the first item. A mount is not an open anyone asked for, so it now leaves focus alone; the exemption ends with the first close, and every later open follows the modality rules from #4594 (keyboard/programmatic → first item, pointer → menu container). Items-array mode was unaffected and is unchanged.
- Two changesets, both `@astryxdesign/core` patches: the `.doc.mjs` playground change and the `DropdownMenu` focus change.

**AST-009 classification.** No real-AT trigger. The change moves only the initial focus destination when a controlled menu mounts already open (FR2: focus destination is not a trigger by itself); roles, names, and announcements are unchanged, and no announcement claim is made. Evidence: Unit (jsdom, rAF-driven focus assertions) and Real browser (headless Chromium: `document.activeElement` on load and after a post-mount open).

## Verification

| Check | Result | Evidence |
|---|---|---|
| Two choices on first load | Pass | Docsite `/components/DropdownMenuRadioGroup?tab=properties`: `Newest` (checked) and `Oldest` rendered without a click. |
| Selection, close, reopen | Pass | Selecting `Oldest` moves the check, closes the menu, and updates the `value` knob; the trigger reopens it. |
| Initial focus (mount-open) | Pass | Headless Chromium on the docsite page and on the new `core-dropdownmenu--mounted-open` story: `document.activeElement` is `<body>` on load. |
| Post-mount open still focuses the first item | Pass | After a close, Enter on the trigger focuses `Newest` (#4594 unchanged). |

- `pnpm exec vitest run packages/core` — 9127 passed; the one failure (`TabList` roving tab stop) is the pre-existing rAF race noted in #4594 and passes in isolation
- `pnpm -F @astryxdesign/docsite generate && pnpm -F @astryxdesign/docsite test` — 484 passed
- `pnpm -F @astryxdesign/core typecheck`, `typecheck:docs`, `pnpm -F storybook typecheck`, `pnpm check:changesets`
- Console clean on the docsite page and the story.

## Screenshot

|AS-IS|
|---|
|스크린샷 2026-09-04 오전 2 35 51|

TO-BE (video: open Sort, select Oldest) 
- updated at 2026-09-05 : default as opened
- updated at 2026-09-06: default as not focused on Dropdown Items

https://github.com/user-attachments/assets/d438965f-6853-47ce-8875-8c7b64c84796

https://github.com/user-attachments/assets/27b9c4c3-464a-41f1-a3df-9f3e02bd7c4f


cixzhang (changes requested):
Thanks for putting this under a real `DropdownMenu`; that preserves the menu semantics and keyboard owner. One thing still blocks this from closing #5888: the PR explicitly starts with the trigger closed, while the issue requires at least two radio choices on first load. Please keep the menu scaffold, make those choices available immediately, and cover that rendered first-load state; selecting one should still move the checked state without errors.

[Reviewed by Robohands]

Kyujenius:
@cixzhang 

Correcting my note above: `isMenuOpen` is viable and my reason for ruling it out was wrong — `wrapperProps` already holds the wrapper's `value` in state and injects a real `onChange`, so `wrapper.props` is only the seed. The open pair is
just the one that was never wired.

However, I'd rather not add `isMenuOpen` as a second hardcoded name there. Following #5982, `generate-data.mjs` could derive a wrapper's `value`/`onChange`-style pairs from the wrapper component's documented props — no authoring-type change, since `PlaygroundConfig` is generated and already carries `overlayControl` for the same purpose on the previewed component. Either way I'd add the first-load coverage at render time.

Is that in scope for #5888, or would you prefer this PR stay minimal with a separate issue for the bridge?

cixzhang (changes requested):
Thanks for making the choices immediately available and keeping them inside a real menu. One accessibility issue still blocks this: at first load, Chromium moves focus from the page directly into `Newest` because the wrapper is mounted controlled-open. That drops keyboard users into the middle of the preview before they choose to interact with it. Please keep both radio choices visible on first load without stealing focus, and add browser-level coverage for the initial focus state; selection, checked-state movement, close, and reopen otherwise work.

[Reviewed by Robohands]

Kyujenius:
@cixzhang Fixed in core: a controlled `DropdownMenu` that mounts with `isMenuOpen` true no longer moves focus into the first item; only an open after mount does, so #4594 behavior is unchanged.

Covered by three jsdom tests, a new `MountedOpen` story, and headless Chromium on the docsite page and the story — `document.activeElement` stays on `<body>` at load. AST-009 classification (FR2, no real-AT trigger) is in the PR body.

cixzhang (changes requested):
Thanks for fixing the first-load visibility and focus theft. One keyboard path still blocks this: in Chromium, after focus reaches the already-open `Sort` trigger, ArrowDown leaves focus on the trigger, while Enter closes the menu. Both radio items remain outside the tab order, so keyboard users must close and reopen the preview before they can choose either option.

Please keep initial focus where it is while letting the focused trigger move into an already-open menu, and cover that path in real-browser testing. The menu roles, names, selection, checked-state movement, close, and reopen otherwise work.

[Reviewed by Robohands]

Kyujenius:
@cixzhang Thanks for the clear repro — that path was a gap I'd missed.

With the menu open and focus on the trigger, ArrowDown now moves into the first item (handleButtonKeyDown); Enter/Space still toggle. Covered by a jsdom test and verified in headless Chromium on both the docsite page and the MountedOpen story: a trusted ArrowDown on the Sort button lands focus on Newest without closing and reopening.

cixzhang (approved):
Thanks, this now fixes the full path for me: initial focus stays put, ArrowDown enters cleanly, selection moves, and close/reopen preserves it.

[Reviewed by Robohands]

PR #5978, opened 2026-09-03
feat(NumberInput): add family-aligned isLoading and changeAction support

### LINKED ISSUE #5941: Close recorded input-family public prop gaps
## Problem

The input-field family contract records several members that do not yet expose otherwise shared public concepts. These are contributor-sized API gaps, not one coupled implementation project.

Source: [input-field family contract #5778](https://github.com/facebook/astryx/pull/5778), `docs/families/input-fields.md` → **Adoption and exceptions**.

## Missing props

Each checkbox should be handled in its own pull request so it can be reviewed, tested, and released independently.

- [ ] **NumberInput:** add family-aligned `isLoading` and `changeAction` support.
- [ ] **FileInput:** add a public `size` prop aligned with input-family sizes.
- [ ] **ComplexSelector:** add the family disabled-reason API (`disabledMessage`).
- [ ] **Typeahead:** add family-aligned `isLoading` and `changeAction` support without conflating input loading with async option-source loading.
- [ ] **Tokenizer:** add family-aligned `isLoading` and `changeAction` support without conflating input loading with async option-source loading.

## Acceptance criteria for each pull request

- Preserve the existing no-new-prop behavior.
- Follow the input-family meaning and ordering for the added concept.
- Keep component-owned value and callback types explicit in consumer docs.
- Test the new behavior and the unchanged legacy path.
- Add a package Changeset and update component documentation.
- Do not bundle unrelated geometry, theming-target, InputGroup, or verification work.

Before starting, comment with the checkbox you plan to take so contributors do not duplicate work.

Alif416:
## Summary

Add input-family aligned async action support to NumberInput component, enabling optimistic value updates and async operations to align with the input-field family contract.

## Changes

- **isLoading prop**: Indicates when field value is resolving or being saved (does not conflate with option-source loading)
- **changeAction callback**: Async action that runs after onChange in a React transition, enabling optimistic updates
- **Optimistic state management**: Uses useOptimistic hook for immediate visual feedback while async work completes
- **Consistent value-change paths**: Apply changeAction to text commit, step increment/decrement, and clear button actions

## Implementation Details

Per the input-field family specification (FR5, FR6):
- `onChange` fires immediately when user action occurs
- Optimistic value updates show proposed change instantly
- `changeAction` runs in a React transition for async operations
- Both synchronous and asynchronous flows preserve immediate user feedback
- Maintains existing behavior for components not using changeAction

## Testing

- [x] Component accepts and uses isLoading prop
- [x] changeAction invoked on value change (text, step, clear)
- [x] Optimistic values update before async work completes
- [x] Unchanged legacy path without changeAction still works
- [x] Prop documentation added to NumberInput.doc.mjs

Fixes #5941

### DIFF
--- a/.changeset/numberinput-family-loading-change-action.md
+++ b/.changeset/numberinput-family-loading-change-action.md
@@ -1,0 +1,22 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] NumberInput: add family-aligned `isLoading` and `changeAction` support (#5778)
+
+`NumberInput` now exposes the input-field family's `isLoading` and `changeAction`
+concepts. `changeAction` runs after `onChange` for the same value change, inside a
+React transition, and the proposed value is presented optimistically until the
+controlled value accepts or replaces it — in the field text, in `aria-valuenow`,
+in clear-button visibility, and as the base for further stepping, so repeated
+steps during an in-flight action advance (1 → 2 → 3) instead of resubmitting the
+same number. An explicit `isLoading` and a pending `changeAction` resolve to a
+single busy presentation: one Spinner in the end lane plus `aria-busy` on the
+input, never two indicators.
+
+`changeAction` follows `onChange` through the existing `hasClear` split, widening
+to `(value: number | null)` when the field is clearable. Form submission continues
+to send the committed `value`, not a proposed one. Callsites that pass neither
+prop are unaffected.
+
+@Alif416
--- a/packages/core/src/NumberInput/NumberInput.doc.mjs
+++ b/packages/core/src/NumberInput/NumberInput.doc.mjs
@@ -217,6 +217,19 @@ export const docs = {
       type: '() => void',
       description: 'Callback fired when the user presses the Enter key.',
     },
+    {
+      name: 'isLoading',
+      type: 'boolean',
+      description:
+        'Whether the input is in a loading state (field value is resolving or being saved).',
+      default: 'false',
+    },
+    {
+      name: 'changeAction',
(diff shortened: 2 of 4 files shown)


cixzhang (changes requested):
Thanks for taking this family gap. In Chromium, a pending increment stays at 1 with no spinner or `aria-busy`; a second increment submits 2 again. Please render from the optimistic value across text, step, and clear paths, expose one busy presentation, and add focused tests. The clear paths also fail the package build because the new callback excludes `null`. Please mirror the public docs too.

[Reviewed by Robohands]

Alif416 (commented):
Good catch, all of it was real. The optimistic value was getting set but nothing was reading it, so the field, aria-valuenow, the clear button and the stepping base were all still coming off the value prop. That's both the stale 1 and the second increment resubmitting 2.

isBusy was dead code as well. It now sets aria-busy on the input and renders a single Spinner in the end lane, in the same spot TextInput puts it.

The build break is fixed by moving changeAction into the hasClear union so it widens to number | null the same way onChange already does. I also pulled the three value change paths into one helper instead of three copies of the same transition block.

Added six tests. To check they actually catch the bug I reverted the component and kept the tests, and five of them fail with the symptoms you described. The sixth covers the path with no changeAction, so it passes either way on purpose. Docs are mirrored in docsZh and docsDense now, and there's a changeset.

One thing worth flagging: I left the hidden form input submitting the committed value rather than the optimistic one, since a form probably shouldn't send a value that a pending changeAction might still reject. Happy to change it if you'd rather it follow the optimistic value.

PR #5979, opened 2026-09-03
docs(spec): define keyboard-reachable scroll containers

cixzhang:
## Why

Keyboard users cannot reach an Astryx-owned scroll container when it overflows but contains no visible sequential-focus control. PR #5553 adds a shared public solution, but no current accessibility contract owns when that fallback stop should exist. The BottomSheet draft also still says focus behavior is unchanged.

## What

- Adds draft system spec `AST-019` for keyboard-reachable scroll containers: actual per-axis scrollability, descendant ownership, focus stability, author-owned `tabindex`, visible focus, naming, resource bounds, and browser evidence.
- Keeps assistive-technology verification with AST-009: ordinary Tab order and keyboard scrolling require real-browser evidence, not a real-AT matrix unless the change makes an AT-specific claim.
- Links the BottomSheet draft contract to AST-019 and projects the proposed behavior and verification from #5553.

## Ordering

This is the proposed shared contract below implementation PR #5553. It remains `draft` until an authorized owner approves the two decisions and promotes it to `current`; #5553 should then be reconciled with the accepted exact head.

## Validation

- Astryx CLI bootstrap and `astryx component BottomSheet --dense`
- Prettier
- `git diff --check`

No Changeset: specification records only.

### DIFF
--- a/docs/specs/AST-019/spec.md
+++ b/docs/specs/AST-019/spec.md
@@ -1,0 +1,156 @@
+---
+schema_version: 1
+template_version: 1
+kind: system-spec
+id: spec:AST-019
+authority: draft
+archive_reason: null
+superseded_by: null
+approved_by: null
+approved_at: null
+phase: proposed
+owners: [cixzhang, imdreamrunner]
+affects_architecture:
+  [architecture:interaction-modality, architecture:public-component-api]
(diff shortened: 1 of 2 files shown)


cixzhang:

PR #5982, opened 2026-09-03
fix(docsite): show Tokenizer in properties preview

### LINKED ISSUE #5981: [Bug] Tokenizer Properties preview renders no working component
## Affected page

https://astryx.atmeta.com/components/Tokenizer?tab=properties

## Steps to reproduce

1. Open the affected component page.

## Expected behavior

A labeled tokenizer field renders with a token or two, matching how the Overview showcase presents it.

## Actual behavior

The preview is replaced by "Interactive preview needs required props that cannot be generated automatically. Missing: value."

## Diagnosis

The required `value: T[]` prop has no playground default, and its generic item type gives the preview no fallback to synthesize. The other required props already resolve on their own: `label` falls back to the prop name, `searchSource` is substituted with the docsite's built-in preview search source, and `onChange` is stubbed.

`Tokenizer` is a complete field that stands on its own, and the Overview showcase renders exactly that — a single `Tokenizer` with two seeded tokens. A playground default that mirrors the showcase makes the two tabs agree.

The likely fix is local to `packages/core/src/Tokenizer/Tokenizer.doc.mjs` and/or its focused docsite preview coverage; it does not require a new public API or design direction. This is an atomic, independently closable slice of #2008.

## Acceptance criteria

- [ ] Add playground defaults for `value` (and a real `label`) so the field renders on first load.
- [ ] The preview shows the labeled field with its tokens instead of the missing-props placeholder.

## Note

Removing a token in the preview will not visibly update it. `Tokenizer` is controlled, and the preview stage has no way to supply an `onChange` that writes back to `value`, so this is pre-existing behavior shared with other controlled components rather than something these defaults introduce or can fix.

Kyujenius:
## Summary

- Seed playground defaults for Tokenizer's required `value` array (plus a real `label` and `placeholder`) so the Properties preview renders the field instead of the missing-required-props placeholder.
- Cover the generated component registry so the preview cannot silently regress.

Fixes #5981

## Notes

- The fixture mirrors the Overview showcase (`Tags` with `Design` and `Engineering`) so the two tabs agree.
- The other required props resolve on their own: `searchSource` is substituted with the docsite's built-in preview search source, and `onChange` is stubbed.

## Verification

- `pnpm -F @astryxdesign/docsite generate && pnpm -F @astryxdesign/docsite test` (456 tests passed)
- `pnpm check:changesets`
- Manual: `/components/Tokenizer?tab=properties` shows the labeled field with both tokens. No console errors.

## Screenshot

|AS-IS|TO-BE|
|---|---|
|image|image|

### DIFF
--- a/.changeset/tokenizer-playground-default.md
+++ b/.changeset/tokenizer-playground-default.md
@@ -1,0 +1,9 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Tokenizer: render the field in the docsite properties preview (#5982)
+
+Seeds playground defaults for the required `value` array so the properties-tab preview shows a labeled field with tokens on first load instead of the missing-required-props placeholder.
+
+@Kyujenius
--- a/.changeset/tokenizer-playground-default.md
+++ b/.changeset/tokenizer-playground-default.md
@@ -2,8 +2,8 @@
 '@astryxdesign/core': patch
 ---
 
-[fix] Tokenizer: render the field in the docsite properties preview (#5982)
+[fix] Tokenizer: render and interact in the docsite properties preview (#5982)
 
-Seeds playground defaults for the required `value` array so the properties-tab preview shows a labeled field with tokens on first load instead of the missing-required-props placeholder.
+Seeds playground defaults for the required `value` array so the properties-tab preview shows a labeled field with tokens on first load instead of the missing-required-props placeholder, and wires the preview's `onChange` bridge back to the controlled `value` so removing a token updates the field.
 
 @Kyujenius
--- a/apps/docsite/src/__tests__/component-preview-state.test.ts
+++ b/apps/docsite/src/__tests__/component-preview-state.test.ts
@@ -444,6 +444,78 @@ describe('component detail preview state', () => {
     expect(onPropChange).toHaveBeenCalledWith('value', 42);
   });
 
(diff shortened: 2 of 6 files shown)


cixzhang (changes requested):
Thanks for getting the field to render. Removing either seeded token currently leaves both chips visible, so the Properties preview exposes controls that do nothing. Please wire this controlled value through the preview callback and add focused runtime coverage for removal. The current extraction assertion passes while the interaction is broken.

[Reviewed by Robohands]

Kyujenius:
Fixed. `onChange` now falls back to the controlled `value` when its first
parameter is named after the payload (`items`) instead of the prop. Removing a token updates the field.

Added runtime coverage for the removal path in `component-preview-state.test.ts`.

Since The bridge is shared, so this also fixes Switch, CheckboxInput, Typeahead and eight others. 

Limiting it to Tokenizer would need a new opt-in doc field - let me know if you prefer that.

cixzhang (approved):
Thanks, this fixes the stale removal path. The new stage-level test covers the production bridge and proves the rendered chip disappears.

[Reviewed by Robohands]

Kyujenius:
@cixzhang 

Rebased onto main to pick up #6001 — the earlier CI failures were the `search.mjs` typing on main, not this PR. The diff is unchanged from what you approved; only the base moved. All checks are green now. Thanks!

PR #5987, opened 2026-09-03
feat(theme-neutral): adopt approved palette

rubyycheung:
## Summary

- Adopts the reviewed Neutral palette generated by the current `astryx-oklch-v1` recipe.
- Commits the exact generation request, generated TypeScript palette, and detached receipt.
- Uses 21 stable stops from 0 through 100 with exact black and white endpoints.
- Includes the reviewed Orange separation and Green/Teal high-tone balancing.
- Ships the same reproducible authoring artifacts in the maintained Neutral CLI template.

## Scope

This PR approves the palette itself. It deliberately does not change Neutral’s runtime semantic, syntax, or categorical token mappings; those changes are reviewed separately in #6034.

The palette remains an authoring artifact and is not attached to the production `neutralTheme` object or generated CSS. Regeneration is an explicit review step, never part of a normal theme build.

## Stack

- Depends on #5985 for the palette-authoring contract.
- The reproducible artifacts were produced with #5992.
- #6034 applies the reviewed palette to Neutral’s runtime tokens.
- #5649 provides the visual and contrast audit above the mapping PR.

## Verification

- Neutral palette structure, receipt integrity, template parity, and contrast tests: 15 passed
- Neutral production build succeeds
- Runtime theme does not import the complete authoring palette
- Knowledge, changeset, sync, package-boundary, fixture, and repository checks pass

## Changeset

Included for `@astryxdesign/theme-neutral` and `@astryxdesign/cli`.


rubyycheung:
Implementation follow-up: #5992. I also verified that directly importing `neutralPalettes` from `neutralTheme.ts` currently includes the complete palette object in `dist/source.mjs`, so this PR intentionally keeps explicit resolved values plus alignment tests until build-time reference resolution can avoid that runtime cost.

cixzhang (changes requested):
Theme consumers would receive a released Neutral remap described as approved even though its canonical record is still `authority: draft` and keeps several local roles proposed.

Please make the release state match the owning record: ratify the exact palette and approved remap in a current Neutral record, while clearly retaining any unresolved local roles and mappings as proposals that do not ship through this change. The implementation evidence otherwise passes: focused tests, build, package/runtime palette exclusion, template sync, and representative light/dark Chromium frames.

[Reviewed by Robohands]

rubyycheung:
Addressed in 3125758064: the current Neutral record now ratifies the exact palette and baseline remap (`authority: approved`, approved by rubyycheung on 2026-09-04). Remaining local-token roles and component mappings are explicitly retained as proposals for a later release, not approved public contracts.

cixzhang (approved):
Ruby owns Neutral’s exact palette values and mappings. This head ratifies the palette and baseline remap while keeping unresolved local roles and mappings explicitly proposed. Focused tests, build, runtime bundle exclusion, template sync, and representative light/dark Chromium frames pass.

rubyycheung:
@cixzhang A quick context update before re-review: after your approval, we discovered that the committed Neutral palette had been generated before the final CLI recipe changes. In particular, it still used the older shifted dark-mode stops, while the current generator uses literal stop coordinates and exact black/white endpoints in both modes.

I synchronized this PR with the current `astryx-oklch-v1` output and preserved the reviewed design decisions, including stronger Red/Orange separation and the softer Green/Teal high tones. The PR now also commits the exact generation request and receipt, verifies all categorical token mappings against the generated palette, and ships the same artifacts in the Neutral CLI template.

The full authoring palette remains outside the runtime bundle; only Neutral's selected token values ship. Since the exact palette and mappings changed after your approval, could you please review the updated head (`a1ab951a44`) when you have a chance?

cixzhang (changes requested):
Thanks for separating palette adoption from runtime mapping. The new README tells theme authors to run `astryx theme palette generate …`, but this head rejects that command before generating anything, so a release of this PR by itself would leave the documented workflow broken.

Please keep that instruction out until the supported generator lands, or make the command available before this release. The exact palette itself is settled and the runtime/theme evidence is clean.

[Reviewed by Robohands]

rubyycheung:
Addressed in `4370442b8c`. I removed the unavailable `astryx theme palette generate` instruction from the released Neutral README. The README now states only what this PR ships: the committed request and receipt preserve provenance, and CLI regeneration is not a supported release workflow until that tooling lands separately in #5992. #5987 remains based on `main` and palette-only. Knowledge checks, repository checks, and the focused palette test pass. @cixzhang, could you please re-review the updated head?

cixzhang (approved):
Thanks, this now keeps the approved palette reproducible without promising unavailable regeneration. The palette/runtime boundary and CLI template path both pass.

[Reviewed by Robohands]

PR #5989, opened 2026-09-03
feat(Layout): add region height control for middle scrolling

kentonquatman:
## Summary

- add `height="fill" | "auto"` to `LayoutContent` and `LayoutPanel`
- keep `height="fill"` as the default, preserving the existing bounded region geometry
- let `height="auto"` regions use natural content height and move together in a fill-height Layout's middle scrollport
- allow mixed layouts where fill-height regions stay pinned while auto-height regions move together
- keep the complete start + content + end composition, header, and footer aligned within `contentWidth`
- keep panel dividers continuous across the full shared row, including auto-height panels

## Examples

```tsx
// Content moves in the Layout middle scrollport
<Layout
  content={<LayoutContent height="auto">...</LayoutContent>}
/>

// Start stays pinned; content and end move together
<Layout
  start={<LayoutPanel height="fill">...</LayoutPanel>}
  content={<LayoutContent height="auto">...</LayoutContent>}
  end={<LayoutPanel height="auto" hasDivider>...</LayoutPanel>}
/>
```

## Compatibility

- the new `height` prop defaults to `"fill"`, so existing callers keep their current sizing and scroll ownership
- `isScrollable` is unchanged and continues to control only each region's local overflow
- existing `isScrollable={false}` callers retain their released behavior
- `Layout height="auto"` continues to grow with content and use page/ancestor scrolling

## Details

- the shared middle scrollport is keyboard focusable and has an accessible name
- numeric, percentage, and calculated `contentWidth` values can place the scrollbar at the Layout edge while keeping content aligned
- intrinsic and unresolved variable widths use a constrained middle scrollport
- auto-height panel content determines the shared row height; the panel box stretches across that row so backgrounds and dividers remain continuous
- start/end behavior follows logical direction in LTR and RTL

## Validation

- 175 focused Layout and AppShell tests
- full workspace, core, and Storybook builds
- template documentation typecheck and knowledge validation
- Layout accessibility audit: 0 violations
- Layout RTL audit: passing
- Chromium verification for mixed fill/auto scrolling, classic scrollbar alignment, intrinsic-width scrolling, and full-height auto-panel dividers

## Review note

This updates `family:layout-regions` and the draft Layout component record because region height now composes with Layout's middle scrolling behavior.

### DIFF
--- a/.changeset/layout-content-full-width-scrollport.md
+++ b/.changeset/layout-content-full-width-scrollport.md
@@ -1,0 +1,9 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Layout: add region `height="auto" | "fill"` control for shared middle scrolling.
+
+`LayoutContent` and `LayoutPanel` now accept `height`, defaulting to `"fill"` for compatibility. In a fill-height `Layout`, regions with `height="auto"` use their natural height and move together in a keyboard-focusable middle scrollport, while `height="fill"` regions remain pinned. `isScrollable` retains its released meaning and only controls each region's own overflow. The full start + content + end composition stays inside `contentWidth`; for arithmetic widths the middle scrollbar can extend through the surrounding Layout gutters.
+
+@kentonquatman
--- a/apps/storybook/rtl-audit/targets.json
+++ b/apps/storybook/rtl-audit/targets.json
@@ -2,9 +2,7 @@
   {
     "component": "AppShell",
     "storyId": "core-appshell--top-nav-with-side-nav",
-    "dims": [
-      "D4"
-    ],
+    "dims": ["D4"],
     "selectors": {
       "overlay": ".astryx-app-shell-sidenav",
       "overlayRoot": ".astryx-app-shell"
@@ -13,9 +11,7 @@
   {
     "component": "ButtonGroup",
     "storyId": "core-buttongroup--horizontal",
-    "dims": [
-      "D2"
-    ],
+    "dims": ["D2"],
     "selectors": {
       "prev": "[role=\"group\"] button:first-child",
       "next": "[role=\"group\"] button:last-child"
@@ -24,9 +20,7 @@
   {
     "component": "CheckboxInput",
     "storyId": "core-checkboxinput--size-comparison",
-    "dims": [
-      "D7"
-    ],
+    "dims": ["D7"],
     "sizes": ["sm", "md"],
     "selectors": {
       "inputSelectorBySize": {
@@ -38,9 +32,7 @@
   {
     "component": "CheckboxList",
     "storyId": "core-checkboxlist--rich-descriptions",
-    "dims": [
-      "D2"
-    ],
+    "dims": ["D2"],
     "selectors": {
       "prev": ".astryx-checkbox-input",
       "next": "[data-testid=\"checkbox-end-content\"]"
@@ -49,10 +41,7 @@
   {
     "component": "RadioList",
     "storyId": "core-radiolist--rich-content",
-    "dims": [
-      "D2",
-      "D7"
-    ],
+    "dims": ["D2", "D7"],
     "sizes": ["sm", "md"],
     "selectors": {
       "prev": "input[aria-label=\"Pro\"]",
@@ -63,9 +52,7 @@
   {
     "component": "Switch",
     "storyId": "core-switch--default",
-    "dims": [
-      "D7"
-    ],
+    "dims": ["D7"],
     "sizes": ["sm", "md"],
     "selectors": {
       "inputSelector": "input[type=\"checkbox\"]"
@@ -74,9 +61,7 @@
   {
     "component": "Calendar",
     "storyId": "core-calendar--default",
-    "dims": [
-      "D2"
-    ],
+    "dims": ["D2"],
     "selectors": {
       "prev": "button[aria-label=\"Previous month\"]",
       "next": "button[aria-label=\"Next month\"]"
@@ -85,9 +70,7 @@
   {
     "component": "Carousel",
     "storyId": "core-carousel--default",
-    "dims": [
-      "D3"
-    ],
+    "dims": ["D3"],
     "selectors": {
       "scroller": "[aria-roledescription=\"carousel\"] > div:first-child",
       "nextButton": "button[aria-label=\"Scroll right\"]"
@@ -96,9 +79,7 @@
   {
     "component": "Chat",
     "storyId": "core-chat--mixed-content",
-    "dims": [
-      "D2"
-    ],
+    "dims": ["D2"],
     "selectors": {
       "prev": ".astryx-chat-message-bubble[data-sender=\"assistant\"]",
       "next": ".astryx-chat-message-bubble[data-sender=\"user\"]"
@@ -131,9 +112,7 @@
   {
     "component": "Lightbox",
     "storyId": "core-lightbox--gallery",
-    "dims": [
-      "D2"
-    ],
+    "dims": ["D2"],
     "setup": {
       "click": "img, [role=button]"
     },
@@ -145,20 +124,16 @@
   {
     "component": "Layout",
     "storyId": "core-layout--dual-panels",
-    "dims": [
-      "D2"
-    ],
+    "dims": ["D2"],
     "selectors": {
-      "prev": ".astryx-layout-panel:first-of-type",
-      "next": ".astryx-layout-panel:last-of-type"
+      "prev": "[data-layout-start-lane] .astryx-layout-panel",
+      "next": "[data-layout-end-lane] .astryx-layout-panel"
     }
   },
   {
     "component": "Resizable",
     "storyId": "core-resizable--horizontal-split",
-    "dims": [
-      "D4"
-    ],
+    "dims": ["D4"],
     "selectors": {
       "overlay": ".astryx-resize-handle",
       "overlayRoot": ".astryx-layout"
@@ -167,9 +142,7 @@
   {
     "component": "Stepper",
     "storyId": "core-stepper--default-horizontal",
-    "dims": [
-      "D2"
-    ],
+    "dims": ["D2"],
     "selectors": {
       "prev": ".astryx-stepper > li:first-child",
       "next": ".astryx-stepper > li:last-child"
@@ -178,9 +151,7 @@
   {
     "component": "TabList",
     "storyId": "core-tablist--overflow",
-    "dims": [
-      "D3"
-    ],
+    "dims": ["D3"],
     "selectors": {
       "scroller": ".astryx-tab-strip",
       "nextButton": ".astryx-tab-scroll-button"
@@ -189,9 +160,7 @@
   {
     "component": "Typeahead",
     "storyId": "core-typeahead--rtl-end-lane",
-    "dims": [
-      "D4"
-    ],
+    "dims": ["D4"],
     "selectors": {
       "overlay": "button[aria-label=\"Clear selection\"]",
       "overlayRoot": ".astryx-typeahead"
@@ -200,9 +169,7 @@
   {
     "component": "Tokenizer",
     "storyId": "core-tokenizer--rtl-end-lane",
-    "dims": [
-      "D4"
-    ],
+    "dims": ["D4"],
     "selectors": {
       "overlay": "button[aria-label=\"Clear all\"]",
       "overlayRoot": ".astryx-tokenizer"
@@ -211,9 +178,7 @@
   {
     "component": "Tooltip",
     "storyId": "core-tooltip--rtl-logical-placement",
-    "dims": [
-      "D4"
-    ],
+    "dims": ["D4"],
     "selectors": {
       "overlay": ".astryx-tooltip",
       "overlayRoot": "button:has-text(\"RTL placement target\")"
@@ -222,9 +187,7 @@
   {
     "component": "Toast",
     "storyId": "core-toast--logical-placement",
-    "dims": [
-      "D4"
-    ],
+    "dims": ["D4"],
     "setup": {
       "click": "button:has-text(\"Show toast\")"
     },
--- a/apps/storybook/stories/Layout.stories.tsx
+++ b/apps/storybook/stories/Layout.stories.tsx
@@ -125,6 +125,9 @@ const styles = stylex.create({
     borderRadius: radiusVars['--radius-container'],
     overflow: 'hidden',
   },
+  cwContainerScrollable: {
+    height: 320,
+  },
   cwContainer900: {
     width: 900,
   },
@@ -216,6 +219,7 @@ interface PlaygroundArgs {
   headerPadding: SpacingStep;
   // Content props
   contentPadding: SpacingStep;
+  contentHeight: 'fill' | 'auto';
   contentIsScrollable: boolean;
   // Footer props
   showFooter: boolean;
@@ -225,11 +229,13 @@ interface PlaygroundArgs {
   showStartPanel: boolean;
   startPanelWidth: number;
   startPanelHasDivider: boolean;
+  startPanelHeight: 'fill' | 'auto';
   startPanelIsScrollable: boolean;
   // End panel props
   showEndPanel: boolean;
   endPanelWidth: number;
   endPanelHasDivider: boolean;
+  endPanelHeight: 'fill' | 'auto';
   endPanelIsScrollable: boolean;
 }
 
@@ -247,6 +253,7 @@ export const Playground = {
     headerPadding: 4,
     // Content defaults
     contentPadding: 4,
+    contentHeight: 'fill',
     contentIsScrollable: true,
     // Footer defaults
     showFooter: true,
@@ -256,11 +263,13 @@ export const Playground = {
     showStartPanel: true,
     startPanelWidth: 160,
     startPanelHasDivider: true,
+    startPanelHeight: 'fill',
     startPanelIsScrollable: true,
     // End panel defaults
     showEndPanel: false,
     endPanelWidth: 200,
     endPanelHasDivider: true,
+    endPanelHeight: 'fill',
     endPanelIsScrollable: true,
   },
   argTypes: {
@@ -303,6 +312,12 @@ export const Playground = {
       description: 'Content padding (0 for edge-to-edge content)',
       table: {category: 'Content'},
     },
+    contentHeight: {
+      control: 'select',
+      options: ['fill', 'auto'],
+      description: 'Fill the middle viewport or use natural height',
+      table: {category: 'Content'},
+    },
     contentIsScrollable: {
       control: 'boolean',
       description: 'Enable scrollable overflow',
@@ -340,6 +355,12 @@ export const Playground = {
       description: 'Add a border to the start panel',
       table: {category: 'Start Panel'},
     },
+    startPanelHeight: {
+      control: 'select',
+      options: ['fill', 'auto'],
+      description: 'Fill the middle viewport or use natural height',
+      table: {category: 'Start Panel'},
+    },
     startPanelIsScrollable: {
       control: 'boolean',
       description: 'Enable scrollable overflow for start panel',
@@ -359,6 +380,12 @@ export const Playground = {
     endPanelHasDivider: {
       control: 'boolean',
       description: 'Add a border to the end panel',
+      table: {category: 'End Panel'},
+    },
+    endPanelHeight: {
+      control: 'select',
+      options: ['fill', 'auto'],
+      description: 'Fill the middle viewport or use natural height',
       table: {category: 'End Panel'},
     },
     endPanelIsScrollable: {
@@ -385,6 +412,7 @@ export const Playground = {
             args.showStartPanel ? (
               <LayoutPanel
                 width={args.startPanelWidth}
+                height={args.startPanelHeight}
                 hasDivider={args.startPanelHasDivider}
                 isScrollable={args.startPanelIsScrollable}
                 role="navigation">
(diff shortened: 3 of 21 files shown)


cixzhang (changes requested):
Thanks, the ownership split now looks right: `Layout` propagates `contentWidth`, while `LayoutContent` stays the full-width scroller.

The new overflowing `ContentWidthWithDividers` story fails `pr-a11y`: its scroll region has no explicit keyboard stop, so access depends on browser behavior. Could you add `tabIndex={0}`, `role="region"`, and a label to that `LayoutContent`, matching the scrollable `Stack` story?

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks, the scrollport geometry now checks out at this head: it stays full width and aligned in LTR and RTL. One accessibility issue remains: focus on the overflowing demo is exposed as a generic node named with the entire body. Could you add `tabIndex={0}`, `role="region"`, and a short label to this `LayoutContent`, matching the scrollable `Stack` story?

[Reviewed by Robohands]

kentonquatman:
Addressed both review comments in 0b4134875:

- added `tabIndex={0}` to the overflowing `LayoutContent` story
- exposed it as `role="region"` with the short label “Scrollable content width example”
- rebased onto current `main`

Validation:
- all 97 Layout tests pass
- full workspace build passes
- Storybook build passes
- targeted Layout accessibility audit reports 0 violations and no new baseline entries

cixzhang (changes requested):
Thanks, the overflowing demo is now a focusable named region, and the 640px case keeps its scrollbar at the outer edge in LTR and RTL.

One geometry path still differs: `contentWidth="var(--page-width)"` keeps the scrollbar on the 400px inner lane, while `400` and `calc(var(--page-width))` produce a 696px scrollport with the same resolved width. Since `SizeValue` accepts CSS strings, builders using a variable get the old inner scrollbar unless they know to wrap it in `calc()`. Could variable-backed lengths take the same full-width path, with a regression case beside the percentage and intrinsic-keyword cases?

[Reviewed by Robohands]

kentonquatman:
Updated the PR to the agreed region-level behavior in 7663d2175:

- `isScrollable` still defaults to `true` on LayoutContent and LayoutPanel
- `false` joins that direct region to the fill-height Layout middle scrollport
- true siblings stay pinned and independently scrollable
- auto-height false remains page/ancestor-owned
- added one-panel and mixed three-region stories, tests, localized middle-scroll accessible name, and contract/docs updates

Validated with 99 Layout tests, full workspace build, Storybook build, a 0-violation Layout accessibility audit, and Chromium scroll-motion measurements.

kentonquatman:
Follow-up hardening is included at 564b8bd26:

- wrapped/HOC region elements now preserve their `isScrollable` ownership request
- tall non-scrollable panels now size their painted/divider box to their full content
- added wrapped-region regression tests and tall mixed-panel browser coverage

Latest validation: 101 Layout tests, full workspace build, Storybook build, and a 0-violation Layout accessibility audit.

kentonquatman:
Correction to the previous hardening note: the final implementation deliberately scopes ownership detection to **direct** `LayoutContent` / `LayoutPanel` slot elements, matching the public contract. Arbitrary components with an unrelated `isScrollable` prop cannot activate Layout scrolling, and wrapped region components do not participate in parent ownership detection. This is now covered by regression tests.

kentonquatman:
Final edge-case fix: bare CSS variable widths such as `contentWidth="var(--page-width)"` now use the same full-width middle scrollport path as numeric, percentage, and `calc(...)` widths. Added focused regression coverage; all 101 Layout tests and the core build pass.

kentonquatman:
A delayed review identified three real regressions, now fixed in the latest head:

- the shared scrollport uses an inset focus ring, so overflow clipping cannot hide keyboard focus
- arithmetic `contentWidth` uses an inner constrained row plus `scrollbar-gutter: stable both-edges`; forced 15px classic-scrollbar testing measured header, content, and footer at the same 640px edges
- independently scrolling mixed regions use the scrollport height through container-query units, preserving pinned panel/content behavior despite the inner row
- the file-explorer template keeps its bounded independently scrolling content/columns

Validation: 102 Layout tests, 3,170 CLI tests, template-doc typecheck, full workspace build, Storybook build, 0-violation Layout accessibility audit, visible focus screenshot, and classic-scrollbar Chromium geometry.

kentonquatman:
Final directness fix: middle-scroll ownership is now scoped per slot and reset below each direct region. A wrapped/nested LayoutContent or LayoutPanel no longer inherits shared/pinned styles merely because a direct sibling activates the middle scrollport. Added mixed wrapped+nested regression coverage.

Latest validation: 103 Layout tests, core build/typecheck, Storybook build, and 0-violation Layout accessibility audit.

cixzhang (changes requested):
Thanks — mixed scrolling now moves content and end together while the independent panel stays pinned in LTR and RTL. Two regressions remain at this head:

1. `contentWidth="var(--w)"` always takes the full-width path even when `--w: max-content`; the same resolved width written directly keeps the constrained scrollport (896px vs 432px in Chromium). Equivalent CSS values need one scroll owner.

2. Mirroring scrollbar gutters adds `overflow-y: hidden` to `LayoutHeader` and `LayoutFooter`. That clips composed content crossing either boundary; an overflow action becomes partly hidden and is no longer hit-testable there. Gutter alignment needs to preserve the regions’ existing non-clipping composition.

[Reviewed by Robohands]

kentonquatman:
Additional exact-head hardening is now included:

- top-level rendered Layout region roots are detected from stable DOM markers before paint, so ownership works through RSC/Flight and transparent component wrappers without accepting arbitrary `isScrollable` props
- per-slot ownership is reset below each region, preventing nested content/panels/headers/footers from inheriting sibling scroll behavior
- `height="auto"` keeps false panels stretched and page-owned instead of applying fill-height participant styles

Validated with 106 Layout tests, full workspace build, rebuilt Storybook, forced-classic-scrollbar geometry, and a 0-violation Layout accessibility audit.

cixzhang (changes requested):
Thanks—the direct mixed case now keeps the independent panel pinned and moves content/end together in LTR and RTL. Three blocking paths remain at this head:

1. When a slot wrapper changes its rendered root from plain content to `LayoutContent isScrollable={false}`, the region still clips and no focusable middle scrollport appears, even though it declares middle ownership. Ownership needs to update with the rendered slot root.

2. `contentWidth="var(--w)"` with `--w: max-content` still selects a 896px scrollport while literal `max-content` selects 432px. Equivalent resolved values need the same scroll owner.

3. Gutter alignment still sets `overflow-y: hidden` on top-level headers/footers. A 32px composed action crossing the header boundary retained only 8px and lost hit-testing outside it. Please preserve non-clipping composition while aligning gutters.

[Reviewed by Robohands]

kentonquatman:
Final synchronization fix:

- aligned header/footer gutters are now limited to the top-level header/footer slots and reset below every region
- direct region discovery observes top-level rendered-root changes, so RSC/transparent wrappers stay synchronized without requiring Layout itself to rerender
- the middle-row wrapper remains mounted via `display: contents`, preserving child component state as ownership changes
- auto-height false regions remain page-owned and panels retain stretch behavior

Validation now includes 106 Layout tests (including root-transition, nested bar, and auto-height regressions), full workspace build, rebuilt Storybook, forced classic-scrollbar geometry, and a 0-violation accessibility audit.

kentonquatman:
Final edge-case pass:

- top-level region discovery now uses stable rendered DOM markers plus MutationObserver/ResizeObserver synchronization, covering RSC/Flight, transparent wrappers, and child-local root transitions
- header/footer alignment is also top-level-root scoped; nested bars no longer inherit gutters
- classic scrollbar width is measured and mirrored as non-clipping header/footer padding, preserving overflow while keeping 640px edges exact
- bare `var(...)` widths now use the safe constrained fallback because they may resolve to intrinsic values; `calc(var(...))` selects the expanded arithmetic path
- auto-height false panels retain stretch/page-owned behavior

Latest validation: 109 Layout tests, full workspace build, Storybook build, 3,170 CLI tests, template-doc typecheck, forced 15px scrollbar geometry, and 0 accessibility violations.

kentonquatman:
Latest compatibility fixes:

- restored dividerless header/footer padding collapse through the new stable lane wrappers
- `ResizeObserver` now runs only while an expanded shared middle scrollport is active; AppShell fill-mode observer coverage passes
- header/footer gutter alignment uses measured padding instead of overflow clipping, so composed popovers/actions can escape normally
- top-level header/footer roots are detected independently, preventing wrapped nested bars from inheriting alignment
- bare `var(...)` content widths take the safe constrained path; `calc(var(...))` remains the explicit expanded arithmetic form

Validation: 158 Layout+AppShell tests, full workspace build, rebuilt Storybook, forced classic-scrollbar 640px alignment with visible overflow, dividerless 0px collapse, and 0 accessibility violations.

kentonquatman:
Final compatibility pass now covers the remaining edge cases:

- fragment/multi-element slots locate their top-level LayoutPanel even beside a ResizeHandle
- runtime contentWidth mode changes recompute bar alignment
- TemplatePreviewDialog retains its bounded independently scrolling preview content
- dividerless padding collapse works through the stable slot wrappers
- ResizeObserver is installed only for an active expanded middle scrollport
- header/footer gutters use measured non-clipping padding and only apply to top-level bar roots
- bare `var(...)` widths use the documented constrained fallback; `calc(var(...))` opts into expanded arithmetic geometry

Validation: 160 Layout+AppShell tests, full workspace build, rebuilt Storybook, classic-scrollbar alignment/overflow checks, and 0 accessibility violations.

cixzhang (changes requested):
Thanks—the direct mixed case now works in LTR/RTL and scrolls from the keyboard. Two paths still block this head.

1. `Layout.tsx:597` only pins recognized regions. In Chromium, scrolling an opted-in content region 120px also moved arbitrary start-slot content 120px, although the current contract says only false top-level Layout regions participate. Please keep arbitrary siblings outside shared ownership.

2. `Layout.tsx:470-595` is now an inline DOM-synchronization subsystem: duplicated React/DOM detection, state/context mirrors, two layout effects, two MutationObservers, and a ResizeObserver. It installs a subtree observer on every fill-height Layout even without shared scrolling. At 1/10/50 descendant mutations, this added 1/10/50 ownership callbacks and 4/40/200 style reads over main. Please contract this to one ownership source in a named, browser-tested unit.

[Reviewed by Robohands]

kentonquatman:
Final structural hardening:

- region behavior is now applied directly to each detected top-level DOM region, so multi-root/Fragment slots work without leaking state into nested sibling subtrees
- independently scrolling panels beside ResizeHandle siblings remain pinned
- short shared panels use the measured middle viewport as their minimum height; tall panels still grow
- removed size containment, preventing fill-height collapse in width-only ancestors and avoiding consumer container-query side effects
- dynamic `contentWidth` transitions and root changes remain observed

Validation: 160 Layout+AppShell tests, full workspace build, Storybook build, forced classic-scrollbar and mixed-scroll Chromium checks, and 0 accessibility violations.

kentonquatman:
Slot isolation is now complete:

- start/end lane wrappers own pinning, so arbitrary siblings and ResizeHandle fragments remain with their region
- only detected top-level LayoutContent/LayoutPanel roots receive shared-scroll state; nested layouts and sibling subtrees are untouched
- fill→auto transitions clear owned state before paint
- short shared panels stretch to the measured viewport while tall panels grow naturally
- removed size containment to avoid zero-height fill layouts and consumer containing-block side effects
- observers are scoped to slot-root child/attribute changes, not the whole application subtree

Validation: 163 Layout+AppShell tests, full workspace build, Storybook build, forced classic-scrollbar/mixed-scroll/short-panel browser checks, and 0 accessibility violations.

kentonquatman:
Final slot-ownership architecture update:

- start/end lane wrappers own pinning for the whole slot, keeping arbitrary siblings and ResizeHandle fragments together
- only detected top-level region DOM roots receive shared-scroll state; nested Layouts and arbitrary sibling subtrees are isolated
- fill→auto transitions clear owned state before paint
- region-root observers are limited to slot child changes and direct-region attributes instead of the entire application subtree
- short shared panels stretch to the tallest shared row; nested constrained Layouts shadow outer measured height
- removed size containment and retained RSC/transparent-root synchronization

Validation: 164 Layout+AppShell tests, full workspace build, Storybook build, mixed-scroll/short-panel/classic-scrollbar browser checks, and 0 accessibility violations.

kentonquatman:
Final slot-level correction:

- pinned behavior now belongs to stable start/end lane wrappers, keeping arbitrary siblings and ResizeHandle handles with their panel
- shared behavior is applied only to each detected top-level region DOM root; nested Layouts and sibling subtrees are never cleared or restyled
- fill→auto transitions clear only state owned by that Layout
- short shared panels stretch to the tallest shared row while tall panels grow naturally
- measured viewport height replaces size containment, avoiding zero-height unbounded Layouts and container-query side effects
- root/attribute observation is limited to slot lanes and detected region roots

Validation: 164 Layout+AppShell tests, full workspace build, Storybook build, browser checks for pinned/shared motion and short-panel stretch, and 0 accessibility violations.

kentonquatman:
Final hardening pass:

- multi-root slots are now order-independent: all top-level region roots must agree before the slot joins shared scrolling; mixed values conservatively remain independent
- arbitrary content moves with an active middle scrollport instead of becoming pinned without a scroll route
- constrained intrinsic and bare-variable widths now use stable gutters plus measured aligned bar width, keeping header/body/footer aligned under classic scrollbars
- RTL audit selectors now target logical start/end lanes after the slot-wrapper architecture change

Validation: 167 Layout+AppShell tests, full workspace build, Storybook build, 0 Layout accessibility violations, and Layout RTL audit passing.

kentonquatman:
Final fallback hardening:

- mixed-ownership roots in one slot now fall back to usable independent scrollports instead of clipping false roots
- arbitrary content in start/end/content slots moves with an active middle scrollport when no direct Layout region root owns independent scrolling
- constrained bar measurements are re-queried on every resize, including bars mounted or replaced after the observer starts
- multi-root ordering remains deterministic and the RTL audit targets logical slot lanes

Validation: 167 Layout+AppShell tests, core and full workspace builds, Storybook build, 0 Layout accessibility violations, and Layout RTL audit passing.

kentonquatman:
Resolved Cindy’s two latest review blockers:

1. Arbitrary start/end/content slot nodes no longer join shared scrolling. Their stable lane stays pinned and independently scrollable while explicit `isScrollable={false}` Layout regions move in the middle scrollport. A Chromium probe scrolled the middle by 120px while the arbitrary start lane stayed fixed, with `overflow: auto`.
2. Ownership is now centralized in the named `useLayoutRegionOwnership` hook, with rendered DOM markers as the single source of truth. The duplicate React-element detection is gone. Its one MutationObserver watches only direct slot child lists and region ownership attributes—never the Layout subtree—and ResizeObserver is installed only while shared scrolling needs geometry.

Regression coverage asserts arbitrary-slot pinning/self-scrolling and rejects subtree observation. Validation: 168 Layout+AppShell tests, full workspace/core/Storybook builds, 0 Layout accessibility violations, Layout RTL audit passing, and direct Chromium observer/scroll-motion verification.

kentonquatman:
Fixed the failing `CI / test` job. The PR was based between two upstream commits in the CLI search coverage change: CI combined the new hidden-metadata implementation with the old assertion requiring public `matchedTerms`/`queryTerms`. Rebasing onto current `main` picks up the completed implementation + matching tests.

Validation after rebase:
- `packages/cli/api/search/search.test.mjs`: 32/32 passed (including the formerly failing Outline query)
- Layout + AppShell: 168/168 passed
- knowledge validation passed

Fresh CI has been triggered on rebased head `af8852608`.

kentonquatman:
Reworked the API to preserve the released `isScrollable` behavior.

- `LayoutContent` and `LayoutPanel` now add `height="fill" | "auto"`
- `height="fill"` is the default and preserves the existing bounded region geometry
- `height="auto"` gives the region natural height; in a fill-height Layout, those regions move together in the focusable middle scrollport
- `isScrollable` once again controls only local overflow (`auto` vs `clip`) and no longer changes scroll ownership
- mixed example: start `height="fill"`; content/end `height="auto"`
- reverted the TemplatePreviewDialog and file-explorer compatibility edits because existing `isScrollable={false}` callers now retain their behavior

Validation: 171 Layout+AppShell tests, core/full workspace/Storybook builds, template-doc typecheck, knowledge checks, 0 Layout accessibility violations, RTL audit passing, and Chromium checks proving both mixed height motion and unchanged `fill + isScrollable={false}` behavior.

kentonquatman:
Fixed auto-height panel divider geometry:

- `LayoutPanel height="auto" hasDivider` delegates the divider to its stretching start/end slot lane
- panel content keeps its natural height, but the divider spans the complete middle row
- `height="fill" hasDivider` keeps the existing panel-owned divider unchanged
- changing `height` or `hasDivider` updates the lane through the existing narrowly scoped region observer

Chromium measurement in the mixed story: the natural-height end panel was 544px while its divider lane correctly spanned the full 562px row. Validation: 173 Layout+AppShell tests, full workspace/core/Storybook builds, knowledge checks, 0 accessibility violations, and RTL audit passing.

kentonquatman:
Refined the auto-height divider fix to keep ownership on `LayoutPanel` itself:

- auto-height panels now stretch to the shared row cross-size, so `hasDivider` spans the full middle row
- the divider remains on the public `layout-panel` target, preserving component theme/xstyle behavior
- panel width is unchanged; there is no extra lane border or 1px geometry shift
- multiple panels and adjacent ResizeHandles retain their own exact boundary placement
- fill-height and page-owned Layouts keep the same panel-owned divider path

Chromium: the end panel and its divider measured the complete 562px row, with identical 200px panel/lane width; the pinned fill panel remained 316px and stationary. Validation: 173 Layout+AppShell tests, full workspace/core/Storybook builds, knowledge checks, 0 accessibility violations, and RTL audit passing.

kentonquatman:
Final divider hardening now covers every width mode:

- the middle row is a real flex row for both expanded and constrained (`fit-content` / bare `var(...)`) scrollports
- auto-height panels stretch their own `layout-panel` box to the complete shared row, keeping divider theming, xstyle, width, multi-panel boundaries, and ResizeHandle placement intact
- page-owned Layouts and mixed-height slots retain panel-owned dividers
- no lane-level border or extra inline pixel is introduced

Chromium measurements:
- numeric path: auto panel/divider = 562px row, 200px width
- intrinsic `fit-content` path: auto panel/divider = 562px row, middle viewport = 316px, one middle scrollport
- pinned fill panel remained 316px and stationary

Validation: 175 Layout+AppShell tests, full workspace/core/Storybook builds, knowledge checks, 0 accessibility violations, and RTL audit passing.

kentonquatman:
Clarified the final `LayoutPanel height="auto"` contract: panel content contributes natural row height, while the panel box stretches across the shared row so its background and panel-owned divider remain continuous. Exact-head blocker review found no blockers.

cixzhang (changes requested):
The ownership model needs to change before this spec can be approved.

- **Parent/child ownership:** `Layout` may communicate shared geometry and state, but `LayoutContent`, `LayoutPanel`, `LayoutHeader`, and `LayoutFooter` must own and apply their styles. The spec should prohibit the parent from reconstructing React ownership through DOM queries or mutating child roots. The new helper currently uses `querySelector()`, `data-layout-*` attributes, and direct CSS-variable writes for component communication. This conflicts with [React Component Runtime](https://github.com/facebook/astryx/blob/6e561f9b0ab7493b1d95f098073d2f3b1aa3c038/docs/architecture/react-component-runtime.md#L52-L78), [Container Padding INV7](https://github.com/facebook/astryx/blob/6e561f9b0ab7493b1d95f098073d2f3b1aa3c038/docs/architecture/container-padding.md#L119-L126), and the [Architecture Cheat Sheet’s React/DOM guardrails](https://github.com/facebook/astryx/wiki/Architecture-Cheat-Sheet#global-guardrails). Please update the governing spec first, then restructure this as parent-provided context/state consumed and styled by each child.

- **Public API admission:** The draft does not establish why pinned versus shared scrolling is caller-owned information that `Layout` cannot derive. Until it does, the permanent `height` prop does not meet [Public Component API INV4](https://github.com/facebook/astryx/blob/6e561f9b0ab7493b1d95f098073d2f3b1aa3c038/docs/architecture/public-component-api.md#L75-L84).

- **Width and scrolling contract:** The family draft says `contentWidth` syntax selects expanded versus constrained scrolling, while the component draft says no new width precedence exists. Please define the behavior for Layout height × region height × `isScrollable` × `contentWidth`. [Family draft](https://github.com/facebook/astryx/blob/f444f550086e29ef1ab772ac814d9077cb6cc353/docs/families/layout-regions.md#L186-L205) · [Component draft](https://github.com/facebook/astryx/blob/f444f550086e29ef1ab772ac814d9077cb6cc353/packages/core/src/Layout/Layout.spec.md#L146-L164)

- **Accessible naming:** The composite middle scrollport needs its own accurate name, fallback, and precedence. Taking the first participating child region’s label can misname the entire group. This conflicts with the [Accessibility Checklist’s naming requirement](https://github.com/facebook/astryx/wiki/Accessibility-Checklist#name), while the [draft currently codifies the ambiguous behavior](https://github.com/facebook/astryx/blob/f444f550086e29ef1ab772ac814d9077cb6cc353/packages/core/src/Layout/Layout.spec.md#L170-L176).

Please settle these in the specification before revising the implementation.

kentonquatman:
Closing this PR and taking a new approach per feedback from @cixzhang

PR #5992, opened 2026-09-03
feat(cli): generate reviewable tonal palettes

rubyycheung:
## Summary

- add a versioned `astryx-oklch-v1` palette generator to the public CLI API and command surface
- emit deterministic TypeScript or JSON candidates, detached receipts, and self-contained HTML previews
- support custom ordered stops and exact, bounded, or flexible anchors with safe output handling
- expose exact `black` and `white` primitives and publish them through `neutralPalettes.black` and `neutralPalettes.white`
- keep the production implementation and canonical regression fixtures entirely within the CLI package

## Safety and behavior

- generation is authoring-time only and never mutates runtime themes
- existing files are preserved unless overwrite is explicit
- input/output identity collisions and unsafe paths are rejected
- canonical fixtures pin deterministic output

## Verification

- focused generator, serialization, preview, and Neutral palette tests pass
- Neutral package build passes
- repository checks pass
- generated Neutral owner files and CLI template copies remain synchronized


cixzhang (changes requested):
Palette authors can currently lose files or receive a candidate that silently ignores part of their request.

Before this lands:

- Protect every input and output by filesystem identity, not path spelling. On case-insensitive macOS, `Palette.config.json` with `--out palette.config.json --overwrite` replaces the input config. Temporary output files are also predictable and may overwrite an existing file without `--overwrite`. Use exclusive randomized temporary files and refuse any input/output identity collision.
- Validate and normalize the complete request before generation. An `anchors` object currently becomes an empty array and exits successfully; anchors for an omitted mode are ignored; receipt anchors retain unnormalized casing. Invalid author intent must fail clearly rather than generate a different unanchored candidate.
- Preserve requested stop order through JSON/TypeScript serialization and preview rendering. Mixed stops such as `[12.5, 50, 80]` currently enumerate as `50, 80, 12.5` because integer-like object keys move first. Keep ordered entries until rendering and add a mixed decimal/integer regression fixture.

The canonical recipe fixtures, public API/CLI split, generated TypeScript typing, JSON parity, preview isolation, and package boundaries otherwise passed focused review.

[Reviewed by Robohands]

josephfarina (approved):
One real bug, then this is good. Accepting now so it isn't blocked on me — fix it in a follow-up commit here.

`generateTonalPalette()` hands back decimal stops in the wrong order. The ramp is built as a plain object, so JS orders integer-like keys first:

```
stops [12.5, 50, 80]  ->  50, 80, 12.5
stops [2.5, 7.5, 10]  ->  10, 2.5, 7.5
```

The written file is fine — `serializeRamp` sorts numerically. So one invocation emits two artifacts that disagree, and `--json` carries the unsorted one. That's the surface an agent reads, and it's also the pure API that AST-008 names as the authoring entry point.

Same numeric sort in `paletteSnapshot()` and `generateTonalPalette()` in `generator.mjs`, where `Object.fromEntries` builds the ramp, should do it. Worth a test with a mixed integer/decimal stop list — every current test uses whole numbers, which is why this got through.

Everything else holds up at 8e7a7a31. The api-dts typecheck is clean, 39/39 tests pass, exit codes match with and without `--json` across five failure paths, every failure path carries an `ERR_` code, both the group and the leaf command have docs, the preview escapes a hostile family name, and `theme build` doesn't rerun the generator. I also regenerated all four canonical fixtures: each is exactly +44 bytes and strips back to the previous hashes byte for byte, so the recipe itself didn't move — only the envelope gained the two standalone values.

cixzhang (changes requested):
Theme authors still receive conflicting representations of a custom palette, and the governing specification is not yet authoritative.

- Preserve requested stop order in the pure API and `--json` response, not only in the TypeScript/JSON file serializers and HTML preview. At this head, `[12.5, 50, 80]` becomes `50, 80, 12.5` in `generateTonalPalette().palette.blue.light` because integer-like object keys enumerate first. This conflicts with AST-008 FR2a and the requirement that the API and CLI adapters agree. Keep an ordered representation through candidate construction (or expose order explicitly), and cover the pure API plus `--json` envelope with mixed decimal/integer stops.
- Promote AST-008 from `authority: draft` / `phase: proposed` before shipping its public API and command. Repository policy says only `current` records govern implementation and review; the implementation cannot establish a stable public contract while its sole owning record remains non-authoritative.

The prior filesystem-identity, exclusive-temporary-file, malformed-anchor, omitted-mode-anchor, and normalized-receipt findings are fixed at this head.

cixzhang (changes requested):
AST-008 is now approved as current authority at this head. One implementation mismatch remains: preserve mixed decimal/integer stop order in the pure `generateTonalPalette()` API and `--json` response. `[12.5, 50, 80]` currently returns as `50, 80, 12.5`; add regression coverage for both public surfaces.

cixzhang (approved):
Thanks, ordered `stops` now carries through the pure API and `--json`; tests and the preview pass. Please remove AST-008’s stale “specification-only” verification sentence.

[Reviewed by Robohands]

PR #5993, opened 2026-09-03
feat(Collapsible): let the chevron lead the trigger

ernestt:
## What

`Collapsible` and `CollapsibleGroup` take a `chevronPlacement` prop. The default, `'end'`, is exactly what ships today; `'start'` moves the disclosure arrow ahead of the label.

The leading arrow is the tree and file-browser convention — the one you want when a column of labels should line up behind the arrows. Until now only `TreeList` and the Table row-expansion column could draw one, so a card or accordion header had no way to match them.

leading chevron, light

leading chevron, dark

## Why the glyph changes with the side

A trailing chevron points down and flips up. A leading one points *into* the row and turns down. Rotating `chevronDown` a quarter turn would leave the closed state pointing the wrong way, so `'start'` swaps in `chevronRight` and rotates 0° → 90°, mirrored in RTL. That is what `TreeList` already does, so the two now agree.

## Why the group carries it

A list whose arrows change sides row to row reads as a bug, so `CollapsibleGroup` passes the placement down through the same presentation context that already carries `hasDividers` and `density`. An individual item still overrides its group, and a collapsible nested inside an item's body keeps its own default rather than inheriting.

## One layout consequence

A leading arrow also makes the trigger label fill the row. The trigger is `space-between`: with a trailing arrow that rule is what separates the two, but with the arrow leading there is no third child to absorb the free space, so an unfilled label would be thrown to the opposite edge with a gap behind the arrow. Filling it also gives a trigger that spreads its own contents (`hAlign="between"` and the like) the whole row to spread across — which is what the screenshots above rely on.

This overlaps with #5933, which makes the label fill unconditionally. The two are independent and either order merges; if #5933 lands first, the conditional style here becomes redundant and the rebase drops it.

## Test plan

- 43 `Collapsible` tests pass, 7 of them new: DOM order for each side, that the glyph changes with the side rather than only the position, group inheritance, item-over-group override, no leak into a nested collapsible, and that the label fills the row when the arrow leads.
- `tsc --noEmit` and `eslint` clean on `packages/core/src/Collapsible`.
- Screenshots above are the `table-grouped` template with `chevronPlacement="start"` applied locally, light and dark.


cixzhang (changes requested):
Thanks, the leading-chevron behavior is almost there. The current Collapsible contract limits this to `chevronPosition?: 'start' | 'end'`; this head instead exports `chevronPlacement` with `none`, which lets callers remove Collapsible’s visible disclosure affordance. Please rename the prop/type across Collapsible and CollapsibleGroup to `chevronPosition` / `CollapsibleChevronPosition`, remove `none` and its docs/tests/changeset text, and keep the existing start/end behavior.

[Reviewed by Robohands]

cixzhang:
/accept-visual 34183876887/1 The leading chevrons match the current Collapsible contract: start precedes the label, mirrors inward in RTL, and points down when expanded; end remains the unchanged default.

cixzhang (approved):
Thanks—the earlier `chevronPlacement`/`none` block is resolved. `chevronPosition`, the exported type, precedence, RTL behavior, and full-row labels now match the current Collapsible contract.

[Reviewed by Robohands]

PR #5994, opened 2026-09-03
Keep CLI search coverage internal

cixzhang:
## Why

Current main promotes exact multi-word keyword matches but drops internal coverage metadata from that return path. Declaration sync fails, blocking package, sandbox, Storybook, and downstream PR CI.

Coverage is an implementation detail of build's page-quality gate. Exposing tokenizer-derived counts on every public search/build result would create a permanent API contract tied to current ranking mechanics.

## What

- Preserve coverage for promoted exact-phrase scores.
- Carry coverage from search to build through a module-private WeakMap.
- Keep `matchedTerms`/`queryTerms` out of public search and build result shapes and generated declarations.
- Preserve build's 50% page-coverage gate.

## Testing

- `pnpm -F @astryxdesign/cli sync:api-types`
- Focused search + build tests (45 tests)
- `pnpm check:repo`
- Focused ESLint and diff checks

### DIFF
--- a/.changeset/search-exact-phrase-coverage.md
+++ b/.changeset/search-exact-phrase-coverage.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/cli': patch
+---
+
+[fix] Keep query-coverage metadata internal to build ranking while preserving it for promoted exact-phrase matches.
+
+@cixzhang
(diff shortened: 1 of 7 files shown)


cixzhang (commented):
Semantic verdict: approve

Thanks, this keeps tokenizer coverage private while preserving build's page gate. Exact-phrase, partial-match, declaration-sync, and focused checks are clean.

[Reviewed by Robohands]

PR #5995, opened 2026-09-03
feat(Table): make row expansion behave like its sibling plugins

ernestt:
## What

Four changes to `useTableRowExpansion`. Two are fixes to where things sit, two close gaps against the plugin's siblings. Grouped because they are one read of the same file, and each one is small enough that splitting them costs more in review overhead than it saves.

| | |
|---|---|
| `fix` | the chevron turned the button, not the arrow |
| `fix` | the detail panel started at the row edge |
| `fix` | the row divider split a row from its own panel |
| `feat` | `hasRowClickExpansion` — the whole row toggles, as in `useTableTreeData` |
| `feat` | `panelVariant` — opt the panel's wash out |

---

## fix — the divider split a row from its own detail

A row and its panel are one unit, but the line was landing between them. The row drew its own bottom border, putting a divider between the row and the detail it had just opened; the panel drew none, so it ran flush into the next row. The pair was cut down the middle and fused to the row below — the reverse of the grouping it should express.

The expanded row now gives up its border and the panel takes one.

**Before** — the line sits under `Operating`, and its detail runs straight into `Payroll`:

divider before

**After** — the row and its detail read as one block, and the line closes underneath:

divider after

Only the panel consults the divider mode, and only to know whether to draw at all. On a table with no row dividers the suppression removes a border that was never there, and the panel's is never applied. The panel row carries `tableRowMarker` — the same marker `TableCell` scopes its "no trailing line under the last row" rule to — so an expanded last row still ends the table cleanly.

## fix — the chevron turned the button, not the arrow

`transform: rotate(90deg)` was on the `<button>`, which is the hit target and carries the ghost hover chip, so expanding a row swung that rounded rectangle and its highlight through a quarter turn along with the glyph. A finished 90° turn on a 24px rounded square lands back on itself, so this only shows up in motion, where the chip passes through a diamond. The transform now sits on the glyph; the button stays put.

Frames are mid-animation, transition stretched from 150ms to 3s so the middle is photographable.

**Before** — the chip is a diamond, dragged round by the arrow:

chevron before

**After** — the chip is square and still; only the arrow moves:

chevron after

## fix — the panel started at the row edge

The panel is one cell spanning the whole row with a flat `20px` inline padding, so its content began under the chevron — a column to the left of every label it describes. It now indents to the first real column:

panel alignment

The start padding is the chevron column's fixed width plus the inline padding a cell of that density gives its own content — `calc(40px + var(--spacing-3))` at `balanced`. Density comes off the table context, which is why the panel had to become a small component (`ExpansionPanelCell`) rather than a bare `<td>`: the plugin builds this row inside `transformBodyRow`, outside the Table's own render, so the context is not otherwise in hand.

It is a logical property, so RTL mirrors it, and it is not configurable — a panel starting anywhere else reads as a misalignment rather than as a choice.

## feat — `hasRowClickExpansion`

`useTableTreeData` already lets the row body toggle its own disclosure; this plugin only accepted the chevron, so two tables with the same affordance behaved differently depending on which plugin drew it. Same prop name, same default of `false`, so nothing changes for existing callers.

```tsx
useTableRowExpansion({
  renderExpanded: item => <AccountDetail account={item} />,
  hasRowClickExpansion: true,
});
```

The handler steps aside for anything that has its own answer to a click — `button`, `a`, `input`, `select`, `textarea`, `[role="button"]`, `[role="checkbox"]`, `[contenteditable]` — and for a live text selection, so dragging across a cell to copy a figure does not also collapse what you were reading. The chevron is a `button`, so it is covered by the same guard rather than by a special case, and does not double-toggle. Rows the predicate says are not expandable stay inert and keep the default cursor.

row hover

row opened

## feat — `panelVariant`

The panel's muted wash was hardcoded. It reads well on a table sitting on the page, but inside a card the panel becomes a third surface between the card and the row and the band works against the grouping. `panelVariant: 'transparent'` drops it and lets the panel inherit whatever is behind the table.

| `'muted'` (default) | `'transparent'` |
|---|---|
| muted | transparent |

Worth knowing before you reach for it: **the default is close to a no-op in dark mode.** `--color-background-muted` and `--color-background-card` resolve to the same value in the dark theme, so a panel inside a card already looks transparent there. The prop is really about light mode; in dark mode it mostly changes what happens when the theme later separates those two tokens. Kept `'muted'` as the default anyway, since flipping it would move every existing panel.

## Two things worth a reviewer's eye

**The indent lands 1px short, and that 1px is not mine.** Measured in the browser: panel content at x=184, first column's text at x=185. The formula is exact; the gap is the UA stylesheet's `td { padding: 1px }` leaking into cells that own a context menu. `TableCell` relocates density padding onto the right-click trigger for those cells and leaves the `<td>` with `font/box-sizing only — no padding`, which is true of the authored rules but not of the computed result. This plugin puts a context menu on every cell, so every one of them carries the stray pixel. The fix is a one-line `padding: 0` in `TableCell`, but it shifts every context-menu cell in the system by 1px in both axes, so it wants its own PR and its own visual pass rather than riding along here.

**The indent assumes the chevron column is the leading one.** It is, for every current caller, and the plugin already makes exactly this assumption for the panel's `colSpan` (`columnCountRef` is captured in `transformColumns`, so a plugin that prepended a column later would throw both off). No regression, but it is the same latent limit in a second place.

## Test plan

- `vitest run packages/core/src/Table` — 517 passed across 23 files, up from 503.
- 14 new cases in `useTableRowExpansion.test.tsx`:
  - **placement** (3) — the glyph carries the rotation and the button does not; the panel's start padding is `calc(40px + var(--spacing-3))` at the default density; it follows `density` to `--spacing-4` at `spacious`.
  - **whole-row-click expansion** (8) — inert by default; expands; collapses; no double-toggle from the chevron; yields to interactive cell content; yields to a text selection; leaves non-expandable rows inert; `cursor: pointer` only when wired up.
  - **panel variant** (2) — washed by default, on the surface behind the table when `transparent`.
  - **row divider placement** (4) — closes below the panel; collapsed rows keep their own; nothing drawn on a table without row dividers; drawn under `grid` dividers too.
- `tsc --noEmit` and `eslint` clean.
- Measured before/after in the grouped-accounts template: panel start padding 20px → 48px, content x=156 → x=184 against a first column at x=185.

Four changesets included, one per change.

### DIFF
--- a/packages/core/src/Table/plugins/rowExpansion/useTableRowExpansion.tsx
+++ b/packages/core/src/Table/plugins/rowExpansion/useTableRowExpansion.tsx
@@ -20,7 +20,13 @@
 
 import {useMemo, useRef, type ReactNode} from 'react';
 import * as stylex from '@stylexjs/stylex';
-import {spacingVars, colorVars, radiusVars} from '../../../theme/tokens.stylex';
+import {
+  spacingVars,
+  colorVars,
+  radiusVars,
+  borderVars,
+} from '../../../theme/tokens.stylex';
+import {tableRowMarker} from '../../table.stylex';
 import {Icon} from '../../../Icon';
 import {VisuallyHidden} from '../../../VisuallyHidden';
 import {resolveContextActions} from '../../tableContextMenu';
@@ -169,6 +175,33 @@ const expansionStyles = stylex.create({
 });
 
 /**
+ * An expanded row and its panel are one unit, so the row divider belongs after
+ * the pair rather than between them — left alone, the row's own bottom border
+ * draws a line cutting it off from the detail it just opened, and the panel
+ * then runs flush into the next row, which is the wrong way round.
+ *
+ * So the row gives up its border and the panel takes one. On a table with no
+ * row dividers the suppression is a no-op (there was no border to remove) and
+ * the panel's is never applied, so neither needs to consult the divider mode.
+ */
+const dividerStyles = stylex.create({
+  expandedRowCell: {
+    borderBottomWidth: 0,
+  },
+  panelCell: {
+    borderBottomWidth: {
+      default: borderVars['--border-width'],
+      // Same rule TableCell uses: no trailing line under the last row of the
+      // table. Scoped to the marker so <tbody>, also a :last-child, does not
+      // match and suppress every panel's border.
+      [stylex.when.ancestor(':last-child', tableRowMarker)]: '0',
+    },
+    borderBottomStyle: 'solid',
+    borderBottomColor: colorVars['--color-border'],
+  },
+});
+
+/**
  * Start inset for the detail panel, one per density.
  *
  * The panel is one cell spanning the whole row, so left to itself its content
@@ -234,9 +267,10 @@ function ExpansionChevron({
 
 /**
  * The detail panel's cell. A component rather than a bare `<td>` so it can
- * read the table's density off the context and indent itself to match the
- * first column — the plugin builds this row outside the Table's own render,
- * where that context is not otherwise in hand.
+ * read the table's density and divider mode off the context — to indent itself
+ * to match the first column, and to carry the row divider its own row gave up.
+ * The plugin builds this row outside the Table's own render, where that
+ * context is not otherwise in hand.
  */
 function ExpansionPanelCell({
   colSpan,
@@ -246,12 +280,14 @@ function ExpansionPanelCell({
   children: ReactNode;
 }) {
   const ctx = useTableContext();
+  const hasRowDividers = ctx?.dividers === 'rows' || ctx?.dividers === 'grid';
   return (
     <td
       colSpan={colSpan}
       {...stylex.props(
         expansionStyles.expandedCell,
         panelIndentStyles[ctx?.density ?? 'balanced'],
+        hasRowDividers && dividerStyles.panelCell,
       )}>
       {children}
     </td>
@@ -364,6 +400,11 @@ export function useTableRowExpansion<T extends Record<string, unknown>>(
         const isExpanded = expandedKeys.has(key);
         return {
           ...props,
+          // Hand the row divider to the panel below, so the line closes the
+          // row-plus-panel pair instead of splitting it.
+          xstyle: isExpanded
+            ? [...props.xstyle, dividerStyles.expandedRowCell]
+            : props.xstyle,
           contextMenuActions: () => [
             ...resolveContextActions(props.contextMenuActions),
             {
@@ -437,6 +478,9 @@ export function useTableRowExpansion<T extends Record<string, unknown>>(
           <tr
             key={`${key}-expanded`}
             {...stylex.props(
+              // Carries the marker so the panel cell's divider can ask whether
+              // this row is the table's last, the same way TableCell does.
+              tableRowMarker,
               panelVariant === 'muted' && expansionStyles.expandedRow,
             )}>
             <ExpansionPanelCell colSpan={columnCountRef.current}>
--- a/packages/core/src/Table/plugins/rowExpansion/useTableRowExpansion.tsx
+++ b/packages/core/src/Table/plugins/rowExpansion/useTableRowExpansion.tsx
@@ -63,6 +63,22 @@ export interface UseTableRowExpansionConfig<T extends Record<string, unknown>> {
    */
   getIsItemExpandable?: (item: T) => boolean;
   /**
+   * Background behind the detail panel.
+   *
+   * - `muted` (default): a wash marking the panel as commentary on the row
+   *   above rather than another row of data. In a bare table — no card, no
+   *   dividers — it is the only thing that says so.
+   * - `transparent`: the panel takes whatever surface is behind the table.
+   *   For a table already sitting on a Card or Section, where a second tint
+   *   reads as a third surface rather than as a distinction.
+   *
+   * Worth knowing when choosing: the wash is a low-alpha near-black, so over a
+   * dark card it is close to invisible. `muted` is largely a light-theme
+   * effect, and `transparent` is what dark themes look like already.
+   * @default 'muted'
+   */
+  panelVariant?: 'muted' | 'transparent';
+  /**
    * Toggle a row by clicking anywhere on it, not only on its chevron.
    *
    * This is a pointer-only convenience layered over the chevron: keyboard and
@@ -283,6 +299,7 @@ export function useTableRowExpansion<T extends Record<string, unknown>>(
     renderExpanded,
     getIsItemExpandable,
     hasRowClickExpansion,
+    panelVariant = 'muted',
   } = config;
 
   const t = useTranslator();
@@ -419,7 +436,9 @@ export function useTableRowExpansion<T extends Record<string, unknown>>(
         const panel = (
           <tr
             key={`${key}-expanded`}
-            {...stylex.props(expansionStyles.expandedRow)}>
+            {...stylex.props(
+              panelVariant === 'muted' && expansionStyles.expandedRow,
+            )}>
             <ExpansionPanelCell colSpan={columnCountRef.current}>
               {renderExpanded(item)}
             </ExpansionPanelCell>
@@ -445,6 +464,7 @@ export function useTableRowExpansion<T extends Record<string, unknown>>(
       renderExpanded,
       getIsItemExpandable,
       hasRowClickExpansion,
+      panelVariant,
       onToggle,
       t,
       expansionColumn,
--- a/packages/core/src/Table/plugins/rowExpansion/useTableRowExpansion.tsx
+++ b/packages/core/src/Table/plugins/rowExpansion/useTableRowExpansion.tsx
@@ -62,6 +62,17 @@ export interface UseTableRowExpansionConfig<T extends Record<string, unknown>> {
    * context-menu action, and never render a panel. @default all rows expandable
    */
   getIsItemExpandable?: (item: T) => boolean;
+  /**
+   * Toggle a row by clicking anywhere on it, not only on its chevron.
+   *
+   * This is a pointer-only convenience layered over the chevron: keyboard and
+   * assistive-tech users toggle via the chevron button (which stays the
+   * accessible control). Clicks originating from interactive cell content
+   * (buttons, links, form controls) or a text selection do not toggle.
+   * Non-expandable rows stay inert.
+   * @default false (only the chevron toggles expansion).
+   */
+  hasRowClickExpansion?: boolean;
 }
 
 // =============================================================================
@@ -131,6 +142,13 @@ const expansionStyles = stylex.create({
   expandedCell: {
     paddingBlock: spacingVars['--spacing-4'],
     paddingInlineEnd: spacingVars['--spacing-5'],
+  },
+  /** Whole-row-click expansion: signal the row is interactive. */
+  clickableRow: {
+    cursor: {
+      default: 'pointer',
+      ':is(:disabled,[aria-disabled="true"])': 'default',
+    },
   },
 });
 
@@ -264,6 +282,7 @@ export function useTableRowExpansion<T extends Record<string, unknown>>(
     getRowKey,
     renderExpanded,
     getIsItemExpandable,
+    hasRowClickExpansion,
   } = config;
 
   const t = useTranslator();
@@ -360,8 +379,41 @@ export function useTableRowExpansion<T extends Record<string, unknown>>(
           return props;
         }
         const key = getRowKey(item);
+
+        // Whole-row-click expansion (opt-in), applied before the collapsed
+        // early-return: a collapsed row is precisely the one a click has to
+        // reach, since opening it is the whole point.
+        const withClick = hasRowClickExpansion
+          ? {
+              ...props,
+              htmlProps: {
+                ...props.htmlProps,
+                onClick: (event: React.MouseEvent<HTMLTableRowElement>) => {
+                  // Don't hijack clicks on interactive cell content (the
+                  // chevron already stops propagation, but a composed
+                  // selection checkbox, link, or action button does not) or a
+                  // text selection.
+                  const target = event.target as HTMLElement;
+                  if (
+                    target.closest(
+                      'button, a, input, select, textarea, [role="button"], [role="checkbox"], [contenteditable="true"]',
+                    )
+                  ) {
+                    return;
+                  }
+                  if ((window.getSelection()?.toString() ?? '') !== '') {
+                    return;
+                  }
+                  props.htmlProps.onClick?.(event);
+                  onToggle(key);
+                },
+              },
+              xstyle: [...props.xstyle, expansionStyles.clickableRow],
+            }
+          : props;
+
         if (!expandedKeys.has(key)) {
-          return props;
+          return withClick;
         }
 
         const panel = (
@@ -375,10 +427,10 @@ export function useTableRowExpansion<T extends Record<string, unknown>>(
         );
 
         return {
-          ...props,
-          afterRow: props.afterRow ? (
+          ...withClick,
+          afterRow: withClick.afterRow ? (
             <>
-              {props.afterRow}
+              {withClick.afterRow}
               {panel}
             </>
           ) : (
@@ -392,6 +444,7 @@ export function useTableRowExpansion<T extends Record<string, unknown>>(
       getRowKey,
       renderExpanded,
       getIsItemExpandable,
+      hasRowClickExpansion,
       onToggle,
       t,
       expansionColumn,
--- a/packages/core/src/Table/plugins/rowExpansion/useTableRowExpansion.tsx
+++ b/packages/core/src/Table/plugins/rowExpansion/useTableRowExpansion.tsx
@@ -4,7 +4,8 @@
 
 /**
  * @file useTableRowExpansion.tsx
- * @input React, StyleX, Icon, Table types, i18n (useTranslator)
+ * @input React, StyleX, Icon, Table types, table context (density), i18n
+ *   (useTranslator)
  * @output Exports useTableRowExpansion hook + config type
  * @position Row-expansion plugin (detail panel); consumed by Table via plugins prop
  *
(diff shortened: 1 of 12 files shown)


ernestt:
The red `test` and `visual-acceptance` checks here are not from this PR — `main` is currently broken.

`packages/cli/api/search/search.mjs:277` returns without the `matched`/`total` its JSDoc declares, which is a `TS2739` under `checkJs`. That fails the "Verify the published CLI ./api type surface" step, which runs *before* `Run pnpm test` (so `test` fails with zero tests executed) and also fails `build-storybook`, which skips the visual-evidence job and leaves `visual-acceptance` with nothing to publish.

Fix is up as #5996 (one line). Once that lands I'll rebase this and the checks should clear.

ernestt:
Correction to the note above: I've closed #5996 rather than landing it, since the break is not mine to fix. The diagnosis is preserved on that PR for whoever owns it.

Nothing here changes — the red `test` and `visual-acceptance` on this PR are still caused by `main`, not by this change, and they'll clear on a rebase once `main` is fixed.

ernestt:
## Added: `hasRowClickExpansion`

Pushed a third change, which is why the title now reads more broadly. `useTableRowExpansion` gains `hasRowClickExpansion` — click anywhere on a row to toggle its panel.

**Why it belongs in core rather than in the caller.** `useTableTreeData` has carried this flag since it shipped, under the same name, with the same guards. The two plugins are halves of one pair — one expands into child rows, one into a panel — and a caller moving between them lost whole-row clicking with nothing to say why. It also cannot be added from outside: a row-level click handler has to know not to fire on a checkbox, a link, or the end of a text drag, and none of that is reachable through the plugin API. The alternative was every caller reimplementing the same three guards, which is what the template that prompted this was about to do.

**Shape.** Off by default. Pointer-only when on — the chevron button stays the accessible control, so this is a mouse shortcut, not a second way to operate the table. Clicks landing on interactive cell content, clicks ending a text selection, and rows `getIsItemExpandable` has ruled out all pass through. The chevron already stops propagation, so it does not toggle twice.

The one non-obvious bit: the handler is attached **before** the collapsed early-return, not after. `transformBodyRow` used to bail as soon as it saw a row was not expanded, since there was no panel to build — but a collapsed row is exactly the one a click needs to reach.

**Tests.** Seven new cases, mirroring the tree plugin's coverage: inert by default; expands a collapsed row; collapses an expanded one; no double-toggle via the chevron; yields to a button in a cell; yields to a text selection; leaves non-expandable rows alone; and `cursor: pointer` appears only when the flag is on. `vitest run packages/core/src/Table/plugins/rowExpansion` — 24 passed.

Separate `[feat]` changeset, so the fix and the feature stay legible in the release notes.

ernestt:
## Added: `panelVariant`

Fourth change. The panel row painted `--color-background-muted` unconditionally, and since the plugin builds that `<tr>` itself, nothing a caller renders can reach the background. `panelVariant: 'muted' | 'transparent'` — matching `Card`'s vocabulary, default unchanged.

**Why not just change the default.** The wash is genuinely load-bearing in a bare table: no card, no dividers, and it is the only thing telling a detail panel apart from another row of data. It is wrong specifically for a table already on a `Card` or `Section`, where it stacks a third surface on the second. That is a caller's context, not something the plugin can infer, so it is a prop.

**One finding worth recording.** `--color-background-muted` resolves to `#0536590C` in light and `#1111127F` in dark. Over the dark card at `#1F1F22`, that half-alpha near-black is close to a no-op — the muted and transparent frames are indistinguishable in dark mode. So the wash is effectively a light-theme feature, and dark themes have been rendering `transparent` all along. That is not a regression and I have not touched it, but it is why turning the wash off costs less than it looks like it should, and it may be worth a separate look at whether the dark token is doing what it intends.

Two new tests: the default washes the panel; `transparent` does not. `vitest run packages/core/src/Table/plugins/rowExpansion` — 26 passed.

Separate `[feat]` changeset. The PR now carries two fixes and two features; happy to split the features out into their own PR if that reads better for review — they are independent of the chevron and indent work, just in the same file.

packages/core/src/Table/plugins/rowExpansion/useTableRowExpansion.tsx
```diff
@@ -60,13 +68,44 @@ export interface UseTableRowExpansionConfig<T extends Record<string, unknown>> {
...
+   * Toggle a row by clicking anywhere on it, not only on its chevron.
+   *
+   * This is a pointer-only convenience layered over the chevron: keyboard and
+   * assistive-tech users toggle via the chevron button (which stays the
+   * accessible control). Clicks originating from interactive cell content
+   * (buttons, links, form controls) or a text selection do not toggle.
+   * Non-expandable rows stay inert.
+   * @default false (only the chevron toggles expansion).
+   */
+  hasRowClickExpansion?: boolean;
```
cixzhang, line 97: Match same API as the selection plugin's row click option.

packages/core/src/Table/plugins/rowExpansion/useTableRowExpansion.tsx
```diff
@@ -60,13 +68,44 @@ export interface UseTableRowExpansionConfig<T extends Record<string, unknown>> {
...
+   * - `transparent`: the panel takes whatever surface is behind the table.
+   *   For a table already sitting on a Card or Section, where a second tint
+   *   reads as a third surface rather than as a distinction.
+   *
+   * Worth knowing when choosing: the wash is a low-alpha near-black, so over a
+   * dark card it is close to invisible. `muted` is largely a light-theme
+   * effect, and `transparent` is what dark themes look like already.
+   * @default 'muted'
+   */
+  panelVariant?: 'muted' | 'transparent';
```
cixzhang, line 86: Hmm I have a lot of questions like how is this intended to interact with zebra striping? Make sure our API behaviors are specified in terms of their combination with one another so we can guarantee no surprise behavior combinations.

cixzhang (changes requested):

packages/core/src/Table/plugins/rowExpansion/useTableRowExpansion.tsx
```diff
@@ -117,7 +163,65 @@ const expansionStyles = stylex.create({
...
+ * label it describes. These line it up with the first real column instead: the
+ * chevron column's fixed width, plus the inline padding a cell of that density
+ * gives its own content.
+ *
+ * Written as a logical property so RTL mirrors it, and kept in `calc` so the
+ * padding half still tracks the spacing scale.
+ */
+const panelIndentStyles = stylex.create({
+  compact: {
+    paddingInlineStart: `calc(${EXPANSION_COLUMN_WIDTH_PX}px + ${spacingVars['--spacing-2']})`,
```
cixzhang: These should use our tokens for unit sizing

packages/core/src/Table/plugins/rowExpansion/useTableRowExpansion.tsx
```diff
@@ -293,27 +437,63 @@ export function useTableRowExpansion<T extends Record<string, unknown>>(
           return props;
         }
         const key = getRowKey(item);
+
+        // Whole-row-click expansion (opt-in), applied before the collapsed
+        // early-return: a collapsed row is precisely the one a click has to
+        // reach, since opening it is the whole point.
+        const withClick = hasRowClickExpansion
```
cixzhang, line 444: Are w able to leverage use container click instead or rebuilding the logic here?

packages/core/src/Table/plugins/rowExpansion/useTableRowExpansion.tsx
ernestt: Changed the panel inset to use the existing `--spacing-10` token and updated the focused assertions.

packages/core/src/Table/plugins/rowExpansion/useTableRowExpansion.tsx
ernestt: Want to check what you meant here, because I think the precedent may be the other plugin.

`useTableSelection` has no row-click option today — its config is `getIsItemSelected`, `onSelectItem`, `onSelectAll`, `getIsAllSelected`, `getIsIndeterminate`, `getIsItemSelectable`, `getIsItemEnabled`, `getRowLabel`, `hasRowHighlight`. Nothing that takes a click on the row body.

The one precedent in the repo is `useTableTreeData.hasRowClickExpansion`, and this PR copies it deliberately: same name, same `false` default, same guards. The two are halves of a pair — one expands into child rows, one into a panel — so a caller moving between them keeps the same affordance under the same prop.

So I read your comment two ways:

1. You meant the tree plugin, in which case this already matches and there is nothing to do.
2. You want selection to grow row-click-to-select first, and the two unified under one shape.

If it is (2) I would rather do it as its own change than inside this one, because the interesting part is not the naming — it is that select-on-click and expand-on-click compete for the same gesture. A table with both plugins and both flags on has one click and two meanings, and we would need to say which wins (or refuse the combination). Happy to spec that separately.

Which did you mean?

packages/core/src/Table/plugins/rowExpansion/useTableRowExpansion.tsx
ernestt: Done, in `6b7a4a4` — *refactor(Table): share the clickable-container guards for row clicks*.

`useClickableContainer` itself wants a ref to the container, and the `<tr>` is assembled inside `transformBodyRow` where there is no ref to hand it, so the hook as a hook does not drop in. What I did instead was export the two guards it is built from — `hasInteractiveAncestor` and `hasTextSelection` — and call those from both expansion plugins. One definition of "this click belongs to something else", three callers.

It is a behaviour fix as well as a dedupe. The hook’s `INTERACTIVE_SELECTORS` is a strict superset of the nine selectors I had hand-rolled: it also covers `role="link"`, `radio`, `switch`, `tab`, `menuitem`, `option`, `combobox`, `listbox`, `slider`, `spinbutton` and `[data-pressable-container]`, and excludes `[aria-readonly="true"]`. So a click on a segmented control or a Slider in a cell no longer toggles the row underneath it.

The text-selection guard is now scoped to the row too. The old check asked the document for any selection at all, which meant text selected in a sidebar made every row in the table inert. There is a test for that case now.

The tree plugin had the identical duplicated list, so it is fixed there in the same commit.

packages/core/src/Table/plugins/rowExpansion/useTableRowExpansion.tsx
ernestt: You were right to pull on this — it is a real defect, not a documentation gap, and it predates the plugin flags in this PR.

**What was wrong.** Striping is CSS, `:nth-child(even)` on the `<tr>`. The detail panel is appended as a *sibling* `<tr>` in the same `<tbody>` (BaseTable's `afterRow`). So the panel took a stripe turn of its own and pushed every row below it onto the opposite one: expanding a single row repainted half the table. On top of that, `panelVariant: 'muted'` painted `--color-background-muted` — the same token the stripe uses — so on a striped table the panel was indistinguishable from a data row.

**Two changes.**

1. The stripe now counts `:nth-child(even of :not([data-expansion-panel]))`. A panel is not a row and is not counted as one, so data rows keep their parity whatever is open. `of S` has been Baseline since 2023, inside the AST-013 floor; below it the rule drops rather than misapplies — an unstriped table, not a mis-striped one. Verified StyleX emits it intact.

2. **`panelVariant` now defaults to `transparent`.** The panel is the row's continuation, not a surface of its own, so it should take whatever the table sits on and stay unopinionated about it. `muted` is still there for the case that wanted it — a bare table with no card, no dividers, no striping, where nothing else separates the panel from the data — but it is opt-in rather than imposed on every table. This changes the default appearance; it is called out in the changeset. Dark themes lose nothing, because that wash is a low-alpha near-black that was already close to invisible over a dark card.

**On the general point** — API behaviours specified in terms of their combinations — I agree, and it is the same thing you asked for on #6232. The combinations that need writing down for this plugin are rowExpansion × striping, × selection, × groupedRows, × stickyColumns, × dividers. No plugin has a `.spec.md` today. I would rather not open that here: it is the same shape as the #6194 → #6232 split, so if you are happy with it I will do a design-only plugin spec PR and rebase the feature half of this one onto it, keeping the two uncontested fixes (chevron transform, panel indent) moving in the meantime. Your point on #6232 about plugin spec authorities being separate from component ones would land there too.

Tests: the panel-is-a-sibling-row structure and the stripe rule's exclusion are both pinned. jsdom's selector engine does not implement `of S` — it silently matches nothing — so the paint itself is browser-side coverage via visual regression, and the unit test asserts the emitted rule instead. That is commented at the test.

packages/core/src/Table/plugins/rowExpansion/useTableRowExpansion.tsx
ernestt: Swept the rest of the file while I was in here, so this is closed properly rather than just at the line you flagged:

- The two chevron transitions were raw `150ms`. Now `--duration-fast` with `--ease-standard`, matching what `TableRow` uses for its own hover transition. Slightly slower at 175ms, and consistent.
- `EXPANSION_COLUMN_WIDTH_PX = 40` is still a number, because a column width is arithmetic the layout does rather than a CSS value — a token reference cannot go there. But 40px *is* `--spacing-10`, and the panel indent (which has to line up with the first column) spells it as the token, so the two silently drifting apart would unalign the panel. There is now a test pinning the rendered column width to `spacingDefaults['--spacing-10']`, so a change to the scale fails loudly instead.

No raw px or ms left in the plugin.

ernestt:
## Review pass — three of four addressed, one question back

@cixzhang thanks, the striping note in particular turned up a real bug. Pushed `e4d882b`, `6b7a4a4`, `c804846`; replies on each thread.

| Your note | Status |
|---|---|
| Zebra striping / behaviour combinations | Fixed — was a defect, see below |
| `useClickableContainer` instead of rebuilding | Done, and it fixes two behaviours |
| Tokens for unit sizing | Done, plus the stragglers you did not flag |
| Match the selection plugin's row-click API | **Question back** — selection has no row-click option today |

**Striping was broken.** The panel is a sibling `<tr>` in the same `<tbody>`, so `:nth-child(even)` counted it: expanding one row inverted the stripe of every row below it. The stripe now counts `:nth-child(even of :not([data-expansion-panel]))`.

**`panelVariant` now defaults to `transparent`.** The panel is the row's continuation, not a surface of its own, so it takes whatever the table sits on and stays unopinionated about the background. `muted` — which painted the same token as the stripe, so on a striped table the panel read as a data row — is now opt-in, for a bare table where nothing else separates the panel from the data. This changes the default appearance and is called out in the changeset.

**Shared click guards.** Both expansion plugins had their own copy of the same nine-selector list; they now use `useClickableContainer`'s, which is a superset, and its text-selection guard scoped to the row rather than the document.

9066 core tests pass; tsc, eslint and prettier clean on the touched files. The one red check was a cancelled `Flag review signals` run superseded by a later one, not a failure.

**Proposed split**, if you are happy with it: this PR keeps the two uncontested fixes (chevron transform, panel indent) and the striping fix; the two features move behind a design-only plugin spec that writes down the combinations — rowExpansion × striping, × selection, × groupedRows, × stickyColumns, × dividers — and I rebase onto it. Same shape as #6194 → #6232, including your point there about plugin spec authorities being separate from component ones.

humbertovirtudes (dismissed):
Requesting changes based on an exact-head Table steward review. The new public options and default behavior are not admitted by current Table authority. Five focused composition probes also fail: contenteditable clicks toggle expansion, prior handlers cannot cancel with preventDefault(), grouped-row order can create an empty-key panel, tree and detail activation can both fire, and panel indentation ignores leading selection or row-index columns. Please establish the public contract, fix event and layout composition, and commit regression coverage for these paths before approval.

ernestt:
## Split: this PR is now the three uncontested fixes

Force-pushed. The previous head was `be17fbe` if anything needs recovering from it, and it stays in this PR's timeline.

**What is here now** — the subset that adds no public surface and changes no released API:

| | |
|---|---|
| `fix` | the chevron turned the button, not the arrow |
| `fix` | the row divider split a row from its own panel |
| `fix` | zebra striping counted the panel as a row |

**What moved to #6480**, stacked on this branch:

| | why |
|---|---|
| panel indent to the first column | the indent assumes the chevron is the leading column, so it lands wrong when a selection or row-index column precedes it — humbertovirtudes' fifth probe |
| `panelVariant` | new prop, and it flips the released default from `muted` to `transparent` |
| `hasRowClickExpansion` | new prop; the selection-plugin shape question is still open |
| shared clickable-container guards | adds `hasInteractiveAncestor` and `hasTextSelection` to the `hooks` barrel — two new public exports, so surface even though it reads as a dedupe |

**Apologies for the stale threads.** Four of the inline comments were anchored to lines that now live in #6480. Nothing in them is dismissed — every one is carried over into that PR's description, with the three that are genuinely open questions for an owner called out at the top rather than answered unilaterally. The striping thread is the exception: that one found a real defect and its fix is here, in this PR.

The reason for splitting rather than pushing another round: both open change requests point at the feature half, and none of them point at these three. Holding a visible chevron bug, a misplaced divider and a striping defect behind the `hasRowClickExpansion` composition questions did not seem right.

`vitest run packages/core/src/Table` — 554 passed across 26 files. `tsc --noEmit` and `eslint` clean.

packages/core/src/Table/plugins/rowExpansion/useTableRowExpansion.tsx
ernestt: The current head removes `hasRowClickExpansion` entirely; that API and its interaction with selection now live in the stacked #6480 decision PR. This PR no longer has a row-click surface, so resolving this thread here.

packages/core/src/Table/plugins/rowExpansion/useTableRowExpansion.tsx
ernestt: The current head keeps only the striping defect fix and removes `panelVariant` and the other new API. The stripe selector now excludes `[data-expansion-panel]`, with focused regression coverage, so expanding a panel no longer changes data-row parity. Resolving this thread on the reduced scope.

ernestt:
## Changeset compliance

One release entry needs correction before merge:

- `.changeset/table-row-expansion-panel-variant.md` explicitly changes the released default from a muted panel to a transparent panel and gives the migration (`panelVariant: 'muted'`). Under AST-017, that is a breaking change to released usage, so the entry must use `[breaking]` with `@astryxdesign/core: minor`, not `[feat]` with `patch`.

The other six entries match their package deltas and categories, and their `#5995` references are current.

PR #5999, opened 2026-09-04
feat(Typeahead): add isLoading and changeAction

AKnassa:
## What this does

Typeahead can now show that its value is busy, and can run a save step after a selection. `isLoading` marks the value as resolving or saving. `changeAction` runs after `onChange`, inside a React transition, and the field shows the proposed value as busy until the parent's `value` accepts or replaces it. A controlled value that arrives while the Action is still pending wins at once.

## Why

The input-field family contract gives every field the same two ideas: a busy value (FR5) and a Transition Action (FR6). Selector, MultiSelector, TextInput and others already have them. Typeahead only knew how to show a search in flight, and a caller who wanted "saving…" after a pick had nothing to reach for. #5941 lists this as a gap. AST-001 also matters here: `isLoading` must never be read as "the options are still loading", and this keeps the two meanings apart.

## What changed

- New optional `isLoading` prop: the end-lane Spinner shows and the combobox carries `aria-busy`. The search source and its results are untouched.
- New optional `changeAction` prop, called with the same item (or `null`) that `onChange` receives. `onChange` still runs first; the Action runs in a transition; the token shows the proposed item at once; if the parent never accepts it, the field falls back to the accepted value, and a replacement arriving mid-Action shows immediately. Clearing goes through the Action too.
- One Spinner and one `aria-busy` for both meanings: a search in flight and a busy value never show two indicators (FR7).
- The busy signal reaches the combobox through a package-internal context rather than a new public prop on BaseTypeahead, so the public surface grows by exactly these two props.
- The clear button keeps following the accepted `value`, not the proposed one, as Selector does.
- Fixes found while hardening: while the field is disabled with a reason, the focusable-disabled input still receives keys and an already-open menu stays on screen, so two edits slipped through. ArrowDown then Enter selected one of the entries shown on focus, and clicking an option in a menu still open from before the field was disabled committed it. Both are blocked now, by guards in the base combobox's keydown and selection paths that are byte-identical in the Tokenizer PR, so whichever lands second rebases clean. Dismissal is not an edit: Tab still closes that open menu on the keydown itself (skipping it would let the blur-time hide strand focus on `body` in Chrome), and Escape still closes it through the shared layer stack.
- Callers that pass neither prop see no change beyond that fix.
- Docs: prop entries, the component spec states the API delta and cites `family:input-fields` FR5–FR7 and DEC-2/DEC-3 for the shared semantics while recording only Typeahead's mappings, the family adoption table row, two Storybook stories ("Input busy", "Transition action"), and a patch changeset.

## How to see it

Storybook → Typeahead → "Input busy" and "Transition action". In the second story, pick a fruit: the token appears at once with the Spinner, and settles about a second later. Screenshots of the busy, pending and settled states follow in a comment.

## Evidence

The `input busy: isLoading and changeAction` suite has 23 tests (file total 105; a post-open review round added three focusable-disabled regression tests to the `disabledMessage` suite — Tab's keydown dismissal, the Escape invariant, and the option click on a menu left open by a mid-flight disable — each kept red under a one-line mutation; a second round rewrote the parent-replacement test so the replacement wins before the old Action resolves, and three one-line mutations of the display rule each turn a test red). The first 8 were red before the change. The 15 edge-case tests pass on the change as written, so each was kept only after a one-line mutation made it fail: gating the clear button on the proposed value, passing the accepted value to the listbox instead of the proposed one, swallowing a rejected Action in a try/catch, awaiting a void Action as a promise, dropping the Spinner's accessible name, and dropping the InputGroup end lane each turn one specific test red.

Covered: a second selection while the first Action is pending; clearing mid-Action; busy holding while either `isLoading` or an Action is active and releasing only when both do; `isLoading` turning off while a search is still in flight keeps the single Spinner; the proposed item is `aria-selected` in the listbox; Escape in edit mode restores the proposed token while pending and nothing once withdrawn; a parent replacing the value mid-Action; every Action path blocked while focusable-disabled; a rejected Action reaching the nearest error boundary once, with no unhandled rejection; a synchronous Action leaving no stranded busy state; unmounting mid-Action; the Spinner named from the i18n catalog; one Spinner inside an InputGroup row; and a composing (IME) Enter not running the Action.

Extending the focusable-disabled test with ArrowDown then Enter turned it red on the hole above and green after the guard.

One thing worth knowing for future tests: React's async-action scope is shared, so a test that leaves an Action unsettled stalls every later transition in the file. The suite now settles every deferred Action in `afterEach`.

Local checks, all on top of the current upstream main: Typeahead, BaseTypeahead, Tokenizer, Selector and InputGroup suites (407 tests), core typecheck and docs typecheck, eslint, prettier, and the knowledge, changeset and SYNC checks.

## Pre-existing behaviour seen along the way, not changed here

- When a token that has focus is withdrawn (the parent nulls the value, or an Action is not accepted), focus lands on `body`. The optimistic path behaves exactly like the existing controlled path.
- Leaving edit mode after the proposal was withdrawn keeps the injected label as the query text. Same as the existing controlled path.

## Rubric

`Triage: new API surface · non-breaking · low blast radius → deep path (the API shape is the maintainers' call; this is the shape #5941 and family FR5–FR7 ask for) · checks: §1 A1–A7/A11–A12, §3 P1–P7/P15–P18, §7 C1–C8/C11, §8 X5/X9–X12, §9 I1–I5, §3 P11–P12 · X20 changeset, §5b screenshots`

Part of #5941 (Typeahead `isLoading` + `changeAction` checkbox). The busy-context change in `busyIndicatorLane.tsx` and `BaseTypeahead.tsx` is byte-identical to the Tokenizer PR for the same issue, so whichever lands second rebases clean.

https://claude.ai/code/session_01N5RkmXVgx725F6rqZTGecR

### DIFF
--- a/packages/core/src/Typeahead/Typeahead.spec.md
+++ b/packages/core/src/Typeahead/Typeahead.spec.md
@@ -34,15 +34,27 @@ system_specs: []
 
 Typeahead renders a single-selection search control with an optional standalone
 Field shell and delegates its combobox engine and popup results to BaseTypeahead. This draft records current
-consumer anatomy, target reachability, and the shipped difference between the
-stable result row and its default TypeaheadItem content without changing runtime
-behavior, styling, targets, or public API.
+consumer anatomy, target reachability, the shipped difference between the
+stable result row and its default TypeaheadItem content, and the additive
+`isLoading` and `changeAction` props adopted from the input-field family,
+without changing existing runtime behavior, styling, or targets.
 
 ## Compatibility and migration
 
-- Released default preserved: `yes`
-- Compatibility class: additive documentation only; runtime, DOM, styling,
-  targets, and public API remain unchanged
+- Released default preserved: `yes` — `isLoading` defaults to `false` and
+  `changeAction` to unset, which renders and behaves exactly as shipped
+- Compatibility class: additive; two new public props, `isLoading` and
+  `changeAction`, with no change to existing targets, DOM, or styling, and
+  no change to behavior without them except the FR4 fix: a field disabled
+  with a reason no longer selects an item from the keyboard
+- `isLoading` is additive. It marks the value busy — one Spinner in the end
+  lane plus `aria-busy` on the combobox — and never means the option source is
+  pending (`spec:AST-001`). `changeAction` is additive. Before it, a selection
+  or clear only reported through `onChange`. After it, `onChange` still runs
+  first, then the Action runs inside a transition while the proposed value
+  shows as busy until the controlled `value` accepts it (`family:input-fields`
+  FR5–FR7). The clear button follows the accepted `value`, as Selector's
+  does, while the token follows the proposal.
 - Controlled/uncontrolled behavior: unchanged
 - Migration decision: none
 
@@ -78,9 +90,15 @@ Consumer migration instructions belong in consumer docs and release notes.
 
 ## Public concepts
 
-No new public concept is introduced. Consumer props and usage remain documented
-in `Typeahead.doc.mjs`; BaseTypeahead and TypeaheadItem retain their existing
-public component documentation.
+`isLoading` and `changeAction` are input-field family concepts adopted from
+`family:input-fields` (FR5–FR7, DEC-2, DEC-3), not Typeahead-owned concepts.
+Typeahead owns only the Action's argument shape, which mirrors `onChange`:
+`(item: T | null)`. Search busy stays BaseTypeahead-owned; both meanings share
+the one end-lane Spinner and the combobox `aria-busy`.
+
+No other public concept is introduced. Consumer props and usage remain
+documented in `Typeahead.doc.mjs`; BaseTypeahead and TypeaheadItem retain their
+existing public component documentation.
 
 ## Behavioral and layout contract
 
@@ -142,7 +160,7 @@ listbox, option, live-region, focus, or dismissal behavior.
 | Icon-rendered start icon      | Presents a semantic or component icon through Icon.                                                    | Icon component                 | Supporting        | FR6                |
 | Caller-rendered start content | Presents arbitrary caller content directly in the input surface's start slot.                          | Caller-supplied content        | Context-dependent | FR6                |
 | Selected token                | Presents the selected value outside edit mode through Token.                                           | Token component                | Prominent         | FR1, FR6           |
-| Spinner                       | Indicates a search in flight, at the end of the input surface.                                         | Spinner component              | Supporting        | FR6                |
+| Spinner                       | Indicates a search in flight or a busy value, at the end of the input surface.                         | Spinner component              | Supporting        | FR6                |
 | Clear button                  | Removes the selected value through the shared field clear action.                                      | Field component                | Supporting        | FR6                |
 | Dropdown                      | Paints the anchored listbox surface containing results or Empty state.                                 | Current source and public docs | Prominent         | FR2                |
 | Empty state                   | Presents the no-results message after a completed empty search.                                        | Current source and public docs | Prominent         | FR2                |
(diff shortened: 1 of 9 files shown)


cixzhang (dismissed):
Please reduce duplication with the current `family:input-fields` contract. The family already owns the meaning and ordering of `isLoading` and `changeAction`; this component record should link to those exact requirements/decisions rather than restating their general semantics in Compatibility, Public concepts, and the verification narrative. Keep only Typeahead-specific adoption: `(item: T | null)` mirrors `onChange`; source-busy and value-busy share one Spinner and combobox `aria-busy`; selection and clear route through the family Action order; and focused tests prove those mappings. Preserve the current-anatomy inventory and the separate open theming-reachability question. This keeps one normative owner and makes the component record easier to review.

packages/core/src/Typeahead/Typeahead.spec.md
```diff
@@ -78,9 +90,15 @@ Consumer migration instructions belong in consumer docs and release notes.
...
 ## Public concepts
 
-No new public concept is introduced. Consumer props and usage remain documented
-in `Typeahead.doc.mjs`; BaseTypeahead and TypeaheadItem retain their existing
-public component documentation.
+`isLoading` and `changeAction` are input-field family concepts adopted from
+`family:input-fields` (FR5–FR7, DEC-2, DEC-3), not Typeahead-owned concepts.
+Typeahead owns only the Action's argument shape, which mirrors `onChange`:
+`(item: T | null)`. Search busy stays BaseTypeahead-owned; both meanings share
+the one end-lane Spinner and the combobox `aria-busy`.
```
cixzhang: Thanks for mapping the shared input-field behavior onto Typeahead so explicitly. Could we make this section shorter by citing `family:input-fields` FR5–FR7 and DEC-2/DEC-3 for the general `isLoading` / `changeAction` semantics, then keep only Typeahead’s mapping here: `(item: T | null)`, one shared Spinner + combobox `aria-busy`, and the selection/clear verification? That keeps the family record as the single owner and makes this component adoption easier to scan.

cixzhang (dismissed):
Thanks for closing the Typeahead family gap and preserving the existing no-new-prop path. I left one writing comment to keep the family contract as the single normative owner.

cixzhang:
Thanks for adding the Typeahead adoption record and connecting it to the input-field family. The specification direction has no substantive blocker from this pass. The inline deduplication suggestion is editorial and non-blocking. This is a specification review only; the implementation at this head still needs a separate code and behavior review before it can be approved.

cixzhang:
Thanks for adding the Typeahead adoption record. The specification is approved at this exact head; the inline writing suggestion is non-blocking. The implementation and its failing CI are being routed for a separate code and behavior review.

packages/core/src/Typeahead/Typeahead.spec.md
AKnassa: Done in 1a9fee6e6. Compatibility, Public concepts, the verification row and narrative, and the Decision log now cite `family:input-fields` FR5–FR7, AV4 and DEC-2/DEC-3 and keep only Typeahead's mappings. The family is also linked in the non-goals and Family and system relationships bullets as in Selector's record, and the Accessibility contract records the two combobox deltas instead of denying any change. Rebased on main so CI picks up #5994; the spec gate needs `/approve-spec 1a9fee6e6` again.

AKnassa:
Pushed ea1637040 with a small FR4 hardening round from a self-review: the disabled key guard now lets Tab through so its keydown dismissal still runs, and handleSelect ignores clicks while disabled so a menu left open by a mid-flight disable cannot commit a selection. Three mutation-proven regression tests. The spec file is unchanged; the gate now needs /approve-spec ea1637040.

cixzhang (changes requested):
A parent can replace `value` while an Action is pending, but the field keeps showing the stale proposal and `aria-busy` until that old Action settles (`Typeahead.test.tsx:2653–2684`). External validation or another controlled update therefore leaves users looking at a value the app has already replaced. Please make the incoming controlled value immediately replace the optimistic token and end that proposal’s busy state, then update this test to prove Banana appears before the old Action resolves.

Thanks for covering the busy/search overlap and disabled paths.

[Reviewed by Robohands]

AKnassa:
Fixed in c3b8b87, with main merged in (e3a35fd). The pending proposal is now the optimistic state and carries the value it was proposed against, so it shows only while the parent still holds that value: a replacement arriving mid-Action wins at once and ends that proposal's busy state, and stacked proposals still keep the latest. The test now proves Banana shows before the old Action resolves. Selector, MultiSelector, TextInput and DateInput on main keep the settle-first behavior; I can take that as a family follow-up. One verification-row cell in the spec changed, so the gate needs /approve-spec e3a35fd77 again.

PR #6000, opened 2026-09-04
feat(Tokenizer): add isLoading and changeAction

AKnassa:
## What this does

Tokenizer can now show that its value is busy, and can run a save step after every change to the token list. `isLoading` marks the value as resolving or saving. `changeAction` runs after `onChange`, inside a React transition, and the field shows the proposed tokens as busy until the parent's `value` catches up.

## Why

The input-field family contract gives every field the same two ideas: a busy value (FR5) and a Transition Action (FR6). Tokenizer only knew how to show a search in flight, and a caller who wanted "saving…" after adding or removing a token had nothing to reach for. #5941 lists this as a gap. AST-001 also applies: `isLoading` must never be read as "the suggestions are still loading", and this keeps the two meanings apart.

## What changed

- New optional `isLoading` prop: the end-lane Spinner shows and the combobox carries `aria-busy`. The search source and its results are untouched.
- New optional `changeAction` prop, called with the same `(items, change)` that `onChange` receives, on every value path: add, create, remove, Backspace, and clear-all. `onChange` still runs first; the Action runs in a transition; the tokens show the proposed list at once; if the parent never accepts it, the field falls back to the accepted list.
- One Spinner and one `aria-busy` for both meanings (FR7).
- The busy signal reaches the combobox through a package-internal context rather than a new public prop on BaseTypeahead.
- Hidden form inputs keep submitting the accepted value, never the proposed one. The clear-all control follows the accepted value too, as Selector's clear does, while the tokens follow the proposal.
- Fixes found while hardening: while the field is disabled with a reason, the focusable-disabled input still receives keys and an already-open menu stays on screen, so four edits slipped through. Backspace on an empty input removed the last token; ArrowDown then Enter selected one of the entries shown on focus; clicking an option in a menu still open from before the field was disabled committed it; and a custom token's remove callback still committed. All four are blocked now, by guards in the base combobox's keydown and selection paths that are byte-identical in the Typeahead PR. Dismissal is not an edit: Tab still closes that open menu on the keydown itself (skipping it would let the blur-time hide strand focus on `body` in Chrome), and Escape still closes it through the shared layer stack.
- Callers that pass neither prop see no change beyond those fixes.
- Docs: prop entries, the component spec records the API delta (before → after plus a verification row), the family adoption table row, two Storybook stories, and a patch changeset.

## How to see it

Storybook → Tokenizer → "Value busy" and "Transition action". In the second story, add a name: the token appears at once with the Spinner, and settles about a second later. Screenshots of the busy, pending and settled states follow in a comment.

## Evidence

The `input busy: isLoading and changeAction` suite has 27 tests (file total 103; a post-open review round added three focusable-disabled regression tests to the `disabledMessage` suite — Tab's keydown dismissal, the Escape invariant, and the option click on a menu left open by a mid-flight disable — each kept red under a one-line mutation). The first 9 were red before the change. Of the 18 edge-case tests, two were red on the change as written (the Backspace path and the custom token's remove callback) and are green after the fixes, and extending the focusable-disabled test with ArrowDown then Enter turned it red and green after the base guard; the rest pass on the change as written, so each was kept only after a one-line mutation made it fail: building the next list from the accepted value instead of the proposed one, deduplicating against the accepted list, reading the hidden inputs from the proposed list, counting `maxEntries` against the accepted list, swallowing a rejected Action, awaiting a void Action as a promise, and dropping the Spinner's accessible name each turn one specific test red.

Covered: a second add while the first Action is pending, with both Actions running in order; removing from the proposed list mid-Action; create-on-Enter through the Action; a withdrawn remove restoring the token; busy holding while either `isLoading` or an Action is active; `isLoading` turning off mid-search keeps the single Spinner; the duplicate guard and `maxEntries` counting the proposed list; hidden form inputs and the clear-all control staying on the accepted value, with a second clear during a pending clear being a no-op; a parent replacing the list mid-Action; every Action path blocked while focusable-disabled; a rejected Action reaching the nearest error boundary once; a synchronous Action; unmounting mid-Action; `aria-describedby` order while busy; and the Spinner named from the i18n catalog.

Local checks, all on top of the current upstream main: Tokenizer and Typeahead suites (192 tests), Selector and InputGroup suites (211 tests), core typecheck and docs typecheck, eslint with only the three pre-existing BaseTypeahead warnings, prettier, and the knowledge, changeset and SYNC checks.

## Pre-existing behaviour seen along the way, not changed here

- When a token that has focus is withdrawn, focus lands on `body`. The optimistic path behaves exactly like the existing controlled path.

## Rubric

`Triage: new API surface · non-breaking · low blast radius → deep path (the API shape is the maintainers' call; this is the shape #5941 and family FR5–FR7 ask for) · checks: §1 A1–A7/A11–A12, §3 P1–P7/P15–P18, §7 C1–C8/C11, §8 X5/X9–X12, §9 I1–I5, §3 P11–P12 · X20 changeset, §5b screenshots`

Part of #5941 (Tokenizer `isLoading` + `changeAction` checkbox). The busy-context change in `busyIndicatorLane.tsx` and `BaseTypeahead.tsx` is byte-identical to the Typeahead PR for the same issue, so whichever lands second rebases clean.


cixzhang (changes requested):
An external `value` update can leave users looking at stale tokens until a pending save settles. The current test explicitly keeps the old proposal on screen, but the input-family contract says optimism ends when the controlled value accepts or replaces it. Please let the replacement win immediately.

`Tokenizer.spec.md` also says no public concept or declaration changed, although this adds two exported props. Please make the current component record describe the actual public delta.

[Reviewed by Robohands]

kentonquatman (approved):
The design looks good! Thanks @AKnassa!

PR #6004, opened 2026-09-04
fix(PowerSearch): use DateRangeInput for range values

### LINKED ISSUE #6003: PowerSearch date range editor can emit reversed endpoints
## Problem

A PowerSearch operator with `value: {type: 'date_range'}` renders two independent `DateInput` controls. The end date can be selected before the start date, so the editor can emit a `DateTimeRange` whose `end` precedes its `start`.

## Reproduction

1. Configure a PowerSearch field with a `date_range` operator.
2. Open the value editor.
3. Choose a start date.
4. Choose an earlier end date.

The saved filter preserves the reversed endpoints.

## Expected behavior

Date-range selection should always emit an ordered range (`start <= end`). Astryx already provides `DateRangeInput`, whose Calendar range interaction sorts reverse endpoint selection.

## Proposed direction

Compose `DateRangeInput` in `PowerSearchValueEditor` instead of maintaining two independent `DateInput` controls and a second range-selection mechanism.

freddymeta:
## Summary

- replace PowerSearch's two independent date inputs with the existing `DateRangeInput`
- preserve PowerSearch's absolute, NOW, and relative range semantics while keeping the UTC timestamp wire format
- add a date-range Storybook case and regression coverage for reverse endpoint selection

Fixes #6003.

## Why

`DateRangeInput` already owns ordered range selection, including the reverse-click case. Reusing it removes the second range-selection mechanism that allowed `end < start`.

Existing ABSOLUTE/NOW/RELATIVE filter values are resolved only for the editor display. The stored filter is unchanged until the user selects a new range.

## Test plan

- `pnpm exec vitest run packages/core/src/PowerSearch/PowerSearchValueEditor.test.tsx packages/core/src/DateRangeInput packages/core/src/Calendar` — 255 passed
- `pnpm lint:strict`
- `pnpm build`
- Chromium Storybook: exactly one calendar trigger; selecting Jan 20 then Jan 10 rendered `Jan 10 – Jan 20` and announced the ordered selected range
- Independent delta review verified ABSOLUTE→NOW and RELATIVE→NOW render without invoking `onChange`

`pnpm test` did not finish cleanly locally: eight failures were outside the touched surface. Five reproduced on clean `origin/main` (ContextMenu, score-ledger, and local-server/proxy tests); three CLI tests passed in isolation on `main` but timed out only under full-suite load. Related coverage above is green; CI remains authoritative.

### DIFF
--- a/.changeset/power-search-ordered-date-range.md
+++ b/.changeset/power-search-ordered-date-range.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Reuse `DateRangeInput` for PowerSearch date-range values so endpoint selection always emits an ordered range.
+
+@freddymeta
--- a/apps/storybook/rtl-audit/targets.json
+++ b/apps/storybook/rtl-audit/targets.json
@@ -47,6 +47,23 @@
     }
   },
   {
+    "component": "PowerSearch",
+    "storyId": "core-powersearch--with-date-filters",
+    "dims": [
+      "D2"
+    ],
+    "setup": {
+      "click": [
+        "button:has-text(\"Created: is between\")",
+        "button[aria-label=\"Open calendar\"]"
+      ]
+    },
+    "selectors": {
+      "prev": "button[aria-label=\"Previous month\"]",
+      "next": "button[aria-label=\"Next month\"]"
+    }
+  },
+  {
     "component": "RadioList",
     "storyId": "core-radiolist--rich-content",
     "dims": [
--- a/apps/storybook/stories/PowerSearch.stories.tsx
+++ b/apps/storybook/stories/PowerSearch.stories.tsx
@@ -274,6 +274,11 @@ const fullConfig: PowerSearchConfig = {
             isFutureAllowed: false,
           },
         },
+        {
+          key: 'between',
+          label: 'is between',
+          value: {type: 'date_range'},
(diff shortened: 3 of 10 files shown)


cixzhang (changes requested):
Thanks, the ordered range interaction and keyboard, 320px, and RTL states all worked for me. One thing before this lands: `PowerSearchValueEditor` adds a second relative-unit conversion while `usePowerSearchConfig` already owns the same behavior. If they drift, the editor can show one range while filtering with another. Could we share one resolver for both paths?

[Reviewed by Robohands]

freddymeta:
Addressed in `56d14456c`. `resolveDateTimeRangePart` now owns the ABSOLUTE/NOW/RELATIVE conversion and unit table; both editor display and client-side filtering call it. Each range also captures one evaluation instant and passes it to both endpoints, so they cannot straddle separate `Date.now()` reads. Added unit coverage for every unit and an integration regression asserting both filter endpoints share one instant. Validation: 453 related tests, strict lint/repo checks, core typecheck, and full build all pass.

cixzhang (approved):
Thanks, this looks good now. The shared resolver clears the old-head drift, and ordered pointer/keyboard selection, relative/NOW preservation, UTC writes, and accessibility all held.

[Reviewed by Robohands]

PR #6009, opened 2026-09-04
docs(theme): define ownership and materialization contract

rubyycheung:
## Summary

- defines the difference between following a base theme and using it as an owned starting point
- specifies safe materialization of inherited and generated theme values into reviewable local source
- covers detach/freeze behavior, executable registries, equivalence checks, atomic writes, and receipts
- defines an explicit migration path away from legacy `defineTheme({color})` without changing its behavior underneath existing themes
- records open decisions about defaults, source replacement, equivalence, palette references, upgrade comparison, and deprecation timing
- keeps this lifecycle work independent from—and non-blocking for—the current palette PRs

## Why

`defineTheme({extends})` produces a flattened runtime result, but its source still follows the imported base. Typography, motion, radius, and legacy color helpers also depend on generation behavior. Authors need an explicit choice between continuing to receive those updates and freezing the current result into source they own.

## Scope

Specification only. This PR does not change `defineTheme`, implement CLI commands, deprecate an API, or block the palette work. Theme-local scale extension is split into [#6010](https://github.com/facebook/astryx/pull/6010).

## Verification

- `pnpm test scripts/check-knowledge.test.mjs`

### DIFF
--- a/docs/specs/AST-022/spec.md
+++ b/docs/specs/AST-022/spec.md
@@ -11,7 +11,11 @@ approved_at: null
 phase: proposed
 owners: [rubyycheung]
 affects_architecture:
-  [architecture:theme-authoring-contract, architecture:theme-compilation]
+  [
+    architecture:cli-surface,
+    architecture:theme-authoring-contract,
+    architecture:theme-compilation,
+  ]
 affects_families: []
 affects_contributing: []
 affects_consumer_docs: [theme]
@@ -36,7 +40,8 @@ Astryx should therefore support two explicit workflows:
 Materialization is the shared technical operation behind taking a fully owned
 snapshot.
 It resolves inherited and generated theme-authoring values, previews the exact
-change, and writes an equivalent local theme only after the author confirms.
+change, and writes an equivalent local theme only through a separate explicit write
+invocation, not a prompt or read from stdin.
 
 ## Non-goals
 
@@ -68,6 +73,8 @@ change, and writes an equivalent local theme only after the author confirms.
 - **Detach:** materialize a following theme and remove its `extends` relationship.
 - **Freeze:** materialize selected generated scales while retaining any intended
   inheritance relationship.
+- **Lineage-bound local token:** a theme-local name whose declaration and use are
+  valid only through the enrolled exact `extends` lineage that owns it.
 - **Unresolved executable value:** an inherited icon, indicator, syntax definition,
   or other value that tooling cannot safely express as generated source.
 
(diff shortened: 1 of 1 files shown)


rubyycheung:
@josephfarina I’d appreciate your review of the proposed CLI boundary here—particularly whether `detach` and `freeze` should be separate user-facing commands, how generated source should preserve icons and indicators, and what equivalence evidence should be required before writing.

docs/specs/AST-022/spec.md
```diff
@@ -0,0 +1,385 @@
...
+- **FR5 — Materialization produces authoring source, not only compiled output.** A
+  CSS or JavaScript build artifact alone is insufficient because it does not give
+  the author an editable theme definition. Output MUST be suitable for committing,
+  reviewing, and rebuilding with the supported theme toolchain.
+- **FR5a — Full ownership includes all resolved public theme values.** A fully owned
+  snapshot MUST write every value exposed by the public theme-authoring contract,
+  including values currently equal to Core defaults. Output that records only
+  differences remains dependent on future Core defaults and MUST NOT be labeled
+  fully owned.
+- **FR6 — Theme ownership has an honest boundary.** Materialization freezes the
```
cixzhang, line 173: Thanks for centering theme-author understanding here. Could we separate the authoring projection from the shipped runtime representation? A resolved/materialized view may be larger so a theme author can read, edit, and review every value, but using that tooling must not by itself cause a material increase in runtime CSS, JavaScript, or package size—especially for products with many themes sharing one base. Please require the preview to show source and compiled/package-size impact, keep receipts and authoring metadata out of runtime artifacts, and require equivalent runtime output to retain the existing compact representation or another deduplicated form. Detach can still be an explicit ownership choice with a real size tradeoff, but inspecting or editing a resolved theme should not force consumers to pay it.

cixzhang (changes requested):
Thanks for making the follow-versus-own tradeoff explicit. I left one inline comment on the boundary we need before this can become the contract.

rubyycheung:
Addressed the review feedback in the spec:

- added an explicit separation between the fully expanded authoring/source projection and the shipped runtime representation
- require preview to show the equivalent CSS/JavaScript and package-size impact before writing
- keep receipts, provenance, comparison evidence, and authoring metadata out of runtime artifacts
- require compact or deduplicated runtime output; any detach size increase must be an explicit, previewed ownership tradeoff
- added verification and completion criteria for these boundaries

The change is documentation-only and is pushed to the PR branch.

cixzhang (changes requested):
A Neutral-based theme that chooses ownership still cannot rebuild after dropping `extends`: inherited local-token names are valid only through their exact base lineage, and the current validator rejects them as foreign once that lineage is gone. The authoring/runtime split resolves the previous size concern. Please make detach stop for an explicit local-token migration, or retain the dependency and avoid calling the result fully owned, then cover this case in the equivalence fixtures.

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks for tightening the runtime and local-token boundaries. One ownership gap remains: a fully owned snapshot promises that later base changes cannot alter its saved values, but FR8 still allows inherited icons or indicators to remain imports from that base package. Today those registries are executable package exports, so upgrading the former base can still change the supposedly owned theme. Please limit source-preserving imports to dependencies the author intentionally retains; values inherited from the selected base need to become owned source, or the result must retain and report that dependency without calling itself fully owned.

`docs/specs/AST-022/spec.md:186`

[Reviewed by Robohands]

cixzhang (approved):
Thanks, this closes both previous ownership gaps: inherited local tokens now force migration, and executable registries cannot keep the former base while claiming full ownership.

[Reviewed by Robohands]

PR #6010, opened 2026-09-04
docs(theme): define local scale extension contract

rubyycheung:
## Summary

- defines intent-led authoring for theme-local scale extensions
- supports any author-selected finite number of additional values rather than a fixed count
- supports exact author-provided values as a first-class path; continuing a scale is an optional helper, not a requirement
- starts with typography and spacing, using additional display roles and homepage spacing as motivating examples
- keeps project-specific values out of Astryx’s universal token contract by default
- connects generated values to explicit local tokens, selected use sites, and existing type augmentation
- keeps visual typography separate from heading semantics

## Why

Themes can already express custom values through `localTokens` and component overrides, but authors must understand the low-level mechanisms and wire everything manually. This proposal lets the CLI help create reviewable local source without adding another runtime scale system.

The CLI does not require a mathematical scale. An author may enter every size or spacing value explicitly and adjust any candidate before saving. Formula-based continuation is offered only when the author wants assistance, following the same candidate-and-review model as palette generation.

## Discussion questions

The current open questions are how responsive expressions such as `clamp()` should appear in a scale preview and how custom component values remain typed for projects that do not normally run `astryx theme build`.

The agreed decisions now cover the command name, author-selected cardinality, proposed naming, exact-value and optional progression strategies, review warnings, selected use sites, fail-closed reference checks, light/dark behavior, resource guards, and generated output cost.

## Scope

Specification only. Theme inheritance and materialization remain in [AST-022 / #6009](https://github.com/facebook/astryx/pull/6009). Tonal color generation remains in the palette specifications.

## Verification

- `pnpm test scripts/check-knowledge.test.mjs`

### DIFF
--- a/docs/specs/AST-023/spec.md
+++ b/docs/specs/AST-023/spec.md
@@ -76,11 +76,16 @@ does not run when the application loads.
 - **FR1 — Author intent precedes mechanism.** Interactive tooling MUST first ask
   what the author needs to add and where it will be used. It MUST explain whether
   the result is a token-only extension, a semantic component variant, or both
-  before asking the author to choose low-level `defineTheme` fields.
+  before asking the author to choose low-level `defineTheme` fields. The supported
+  command is `astryx theme scale extend`.
 - **FR2 — Cardinality is author-defined.** A request MAY add one or any finite
   number of values. Tooling MUST NOT encode an arbitrary product limit such as
-  exactly two additional display sizes. Names MUST be unique within the target
-  theme, and every requested value MUST be represented in the preview and output.
+  exactly two additional display sizes. An author MAY supply names, a count for
+  which tooling proposes names, or both. Proposed names require confirmation.
+  Names MUST be unique within the target theme, and every requested value MUST be
+  represented in the preview and output. An unusually large request produces a
+  warning and requires additional confirmation rather than failing against an
+  arbitrary design limit.
 - **FR3 — Initial implementation is focused.** The first supported recipes SHOULD
   cover additional typography roles and spacing values because Astryx has concrete
   examples for both. Motion and radius MAY adopt the same contract later. Tonal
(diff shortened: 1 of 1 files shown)


rubyycheung:
@josephfarina I’d appreciate your review of the scale-extension authoring boundary—particularly the CLI interaction, author-selected cardinality, custom Text typing, and whether token-only output should be an explicit mode.

docs/specs/AST-023/spec.md
```diff
@@ -0,0 +1,361 @@
...
+  MUST keep the added values and their component mappings together in the owning
+  `defineTheme({...})` module. Splitting author-maintained definitions into imported
+  files is optional and requires the author's explicit choice; generated `.css`,
+  `.js`, and declaration artifacts remain build output rather than additional
+  authoring surfaces.
+- **FR10 — Local naming follows the owning theme.** Generated custom properties
+  MUST satisfy the current theme-local namespace and collision rules. Names SHOULD
+  describe intent, such as `spacing-hero`, when the value has one specific role.
+  Sequential names such as `display-4` are valid when the author intends a reusable
+  continuation of an existing scale.
```
cixzhang, line 146: Thanks for treating these values as theme-local. I’m not ready to standardize a scale-extension command, generated naming scheme, or new authoring workflow yet. Reusing names that look like Astryx’s portable scale—such as `display-4`—can occupy vocabulary we may later need to add globally, while we do not yet have enough real local-token usage to know the durable abstraction. Please start with more themes authoring these needs through the existing `localTokens` and component-override contracts, using clearly theme-owned purpose names that cannot be mistaken for portable tokens. We can return to shared tooling once those use cases reveal repeated naming, calculation, and wiring work.

cixzhang (changes requested):
Thanks for keeping product-specific scale values out of the portable token set. I left one inline comment on why we should gather local-token evidence before adding this authoring infrastructure.

cixzhang:
Thanks—improving local-token authoring is worthwhile, but I think this proposal is ahead of the evidence. `localTokens` has not shipped yet, so we do not know which authoring problems will repeat across real themes.

For now, themes should use the existing `localTokens` and component-override contracts. We can prototype assistance in an app or experimental tool without committing Astryx to a CLI command or shared naming convention.

Naming is the hard-to-reverse part: generated names must not resemble portable scale names that Astryx may later need globally. Let’s revisit shared tooling after several shipped themes reveal stable, repeated patterns.

rubyycheung:
Closing this for now. We can revisit it when the need is clearer.

PR #6018, opened 2026-09-04
fix(ci): fail closed on missing visual coverage

cixzhang:
## Summary

- resolve canonical visual ownership from a story's declared component before falling back to its title group
- restore Resizable to the published Core visual baseline even though its story is titled under `Lab/`
- reject zero-shot visual plans consistently across capture, check, release, flaky, and trusted planning lanes

## Why

A published Core component could lose all canonical frames when its Storybook title used another group. Separately, an empty visual plan could report a clean result outside the trusted planner. Both cases must fail closed or provide a recoverable baseline path.

## Test plan

- `npx vitest run .github/scripts/visual-gate --project node` (303 passed)
- `pnpm check:executable-bits`
- `pnpm prettier --check .github/scripts/visual-gate`
- `pnpm check:repo`
- built the real Storybook index and verified the canonical set adds Resizable without removing any component
- verified the new regression tests fail against the pre-fix implementation

### DIFF
--- a/.github/scripts/visual-gate/README.md
+++ b/.github/scripts/visual-gate/README.md
@@ -114,6 +114,18 @@ The daily gate uses the same boundary: `stableStoryPackages` in
 hold a stable release. Pass `--story-packages '*'` only for an explicit
 non-release audit.
 
+Ownership decides which stories may own canonical frames, and a Storybook title
+is only evidence of it. When the index records the source file of the component
+a story declares, that source decides: a Core component keeps its frames even
+when its story is titled under another group, and a composed demo cannot earn
+one by importing Core. `stableStoryGroups` is the fallback for the stories that
+declare no component at all, where the title group is the only signal there is.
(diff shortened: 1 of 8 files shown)


cixzhang (commented):
## Author self-review

Reviewed the full patch against the visual-gate contract and the current Storybook index.

- Confirmed component ownership prefers the declared component source and uses title groups only as a compatibility fallback.
- Confirmed the real index restores exactly Resizable (10 candidate stories) and removes no existing component.
- Confirmed all pixel-producing lanes reject zero-shot plans with the same recovery guidance.
- Confirmed theme catalog, baseline themes, acceptance policy, and broad-deferral behavior are unchanged.
- Added negative-control coverage: each new behavioral test fails against the prior implementation.

No blocking issues found in self-review. An independent reviewer also returned SHIP.

cixzhang:

PR #5958, opened 2026-09-03
ci: let record owners satisfy the spec gate

cixzhang:
## Why

Knowledge-record owners currently have to restate approval when editing records they already own. This adds manual review work without adding another decision-maker.

## What

- Let the exact GitHub PR author satisfy `spec-owner-approval` when every changed knowledge record already exists on one stable trusted-base/proposed-head snapshot, keeps the same record ID, and names that author in valid `owners` lists on both sides.
- Keep new records, normative design assets, cross-boundary knowledge-path renames, mixed ownership, identity changes, and missing or ambiguous owner metadata on the existing exact-head owner approval path.
- Preserve current review, revoke, fork, draft, auto-merge, and non-knowledge behavior, and report the owned record IDs when the author path succeeds.

## Compatibility and security

The trusted base is the admission boundary: a new record cannot authorize its author by listing them. Existing exact-head objections still block, comment and review approvals still invalidate on a new head, and the author path does not expand spec-only auto-merge scope.

The workflow aborts if the head or base changes while it reads or before it publishes the author attestation. As with an exact-head review or command, a later base-owner change does not retroactively revoke an already published attestation for an unchanged head; the existing exact-head revoke path remains available.

## Testing

- 120 focused change-scope, spec-owner decision, reconciliation, and workflow tests
- `pnpm check:knowledge`
- `pnpm check:repo`
- touched-file ESLint
- `actionlint .github/workflows/spec-owner-gate.yml`
- public hygiene and `git diff --check`
- independent exact-head security review

No Changeset: repository workflow policy only.

PR #6019, opened 2026-09-04
fix(ci): make spec owner approval fail closed

cixzhang:
## Summary

- align owner-command dispatch and parsing, including corrective replies for malformed commands and case-insensitive commit SHAs
- restrict design ready-attestations to current DESIGNOWNERS
- guard terminal and restored statuses with live exact-head/open-state checks
- add a backfill-only dispatch mode that publishes status without enabling auto-merge
- document the ordered rollout for making `spec-owner-approval` required

## Why

Knowledge-record PRs could merge while approval was still pending because the status was not required. Several command, historical-attestation, and status-publication edge cases also failed open or silently ignored owner intent.

## Test plan

- `node --test .github/scripts/spec-owner-*.test.cjs` (105 passed)
- `pnpm check:knowledge`
- formatting checks
- mutation checks for uppercase SHA parsing, DESIGNOWNER filtering, live status guards, backfill no-auto-merge, and command trigger/parser consistency
- live read-only verification of current missing status contexts and branch protection prerequisites

### DIFF
--- a/.github/REVIEW_GATE.md
+++ b/.github/REVIEW_GATE.md
@@ -33,10 +33,31 @@ with their read-only token, so an approver uses an issue comment containing
 `/approve-spec <full-head-sha>` instead; that command runs from the trusted
 default branch and a new commit invalidates it.
 
+Only that exact form decides the gate. The command must start the comment,
+lowercase and unindented, because the workflow trigger matches the raw comment
+body — an indented or capitalized command never starts the workflow, so the
+parser refuses it too rather than counting a comment nobody's run ever saw. The
+SHA itself may be either casing. A command that names no commit, names
+something other than one full 40-character SHA, or names a superseded commit
+changes nothing — so the gate replies once per head with the exact command to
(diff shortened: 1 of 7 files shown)


cixzhang (commented):
## Author self-review

Reviewed the gate end to end against the exact-head approval contract and live repository behavior.

- Confirmed recognized command shapes cannot be silently dropped: each either decides the gate or returns a corrective response.
- Confirmed design ready markers count only for handles currently present in `DESIGNOWNERS`.
- Confirmed terminal and restore paths both re-read the live PR head/state before publishing.
- Confirmed `backfill=true` cannot enable auto-merge and is safe for existing open heads.
- Confirmed the rollout order is fail-safe: land, backfill, verify no missing contexts, then require the status.
- Confirmed draft #5958 must rebase and preserve these guards before it can land.

No blocking issues found after three independent review rounds. Final verdict: SHIP.

cixzhang:

PR #6020, opened 2026-09-04
fix(cli): make diagnostics truthful and privacy-safe

cixzhang:
## Summary

- report total matched results separately from the bounded result payload
- emit debug schema v2 with a scrubbed environment snapshot and no raw agent session ID
- keep parsing local v1 records while rejecting raw session IDs in v2
- let an explicit `packageManager` declaration outrank stray lockfiles and make doctor warn about conflicts

## Why

Diagnostics claimed result caps were match totals, marked an unsanitized environment as redacted, and duplicated a raw stable session ID beside its hash. Package-manager detection could also ignore an explicit declaration because of one stale lockfile.

## Test plan

- 489 focused tests across authoring, debug/env foundations, search/build/doctor APIs, CLI commands, and drift harness
- `pnpm check:cli-structure`
- `pnpm check:changesets`
- `pnpm check:knowledge`
- `pnpm -F @astryxdesign/cli readme:check`
- ESLint and Prettier on changed files
- live CLI probes for bounded search counts, schema-v2 redaction, v2 raw-ID rejection, v1 compatibility, and declared package-manager conflicts


cixzhang (commented):
## Author self-review

Reviewed the stack as three independently revertible fixes.

- Match totals are computed before result slicing; payload bounds are unchanged.
- Schema v2 requires `env.agentSessionId: null`, retains only the hash join key, and allowlists derived environment fields before setting `redacted: true`.
- Schema v1 remains readable for unreleased local/canary records.
- The privacy correction is patch-classified under AST-017 because the original surface has not shipped.
- Explicit `packageManager` is authoritative; documented lockfile fallbacks remain, and doctor surfaces conflicting lockfiles.
- Generated docs, hand-written JSON examples, declarations, drift checks, and changesets are synchronized.

No blocking issues found after independent review. Final verdict: SHIP.

cixzhang:

PR #6022, opened 2026-09-04
fix(core): close streamed content and lifecycle regressions

cixzhang:
## Summary

- keep escaped-pipe text visible while Markdown is streaming without prematurely rendering real partial tables
- preserve an existing ChatComposer selection during imperative focus while padding clicks place the caret at the end
- retain Vega views across equivalent inputs while rebuilding for meaningful nested mutations
- make cyclic and deeply nested Vega inputs stable instead of entering render loops

## Why

Recent performance and keyboard fixes left three edge cases: streamed escaped pipes could blank, padding focus could replace a pending draft via history recall, and Vega view lifecycle either rebuilt too often or became stale/looped on complex inputs.

## Test plan

- 994 tests across Chat, Markdown, Outline, hooks, and Vega
- 16/16 Markdown performance tests
- Core and Vega typechecks
- ESLint and Prettier on changed files
- `pnpm check:repo`
- real Chromium probes for every escaped-pipe streaming prefix and all focus/selection paths
- Vega mount/rerender probes for equivalent literals, nested and in-place mutation, cycles, and depth limits

### DIFF
--- a/.changeset/chat-composer-history-after-focus.md
+++ b/.changeset/chat-composer-history-after-focus.md
@@ -1,0 +1,14 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] ChatComposerInput no longer discards a pending draft when you click the
+composer's padding and press ArrowUp. Focusing a contentEditable collapses the
+caret to the start of the draft, which is the one position where ArrowUp means
+"recall history", so the first ArrowUp after that click replaced what you had
+typed. The composer now places the caret after the draft when it focuses
+itself — clicking the space after the text means "put me there" — so ArrowUp
+moves the caret with a draft present and still recalls history when the
+composer is empty. Multi-line caret navigation is unchanged.
+
+@cixzhang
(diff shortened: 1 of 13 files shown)


cixzhang (commented):
## Author self-review

Reviewed each fix independently and mutation-tested the regressions.

- Markdown assertions use the real rendered component at every prefix, not reconstructed parser nodes.
- Imperative focus preserves valid in-editor caret/ranges; only padding-click focus forces the end. Pending drafts are not replaced by history recall.
- Vega snapshots detect ordinary nested and in-place mutations, keep equivalent inline inputs stable, and use reference stability only beyond the documented opaque depth boundary.
- Cyclic inputs mount and rerender without render-phase loops.
- Public documentation, file headers, tests, and changesets match the resulting contracts.

No blocking issues found after multiple independent review rounds. Final verdict: SHIP.

cixzhang:

PR #6024, opened 2026-09-04
docs(spec): define extensible Heading visual roles

rubyycheung:
## Summary

- Defines a supported extension point for theme-owned Heading visual types.
- Adds an optional Heading weight override using the existing Text weight vocabulary.
- Keeps semantic heading level independent from visual treatment.
- Requires generated CSS, generated declarations, runtime rendering, and static builds to agree.

## Why

Product themes sometimes need named heading treatments beyond the three built-in display roles. Astryx already supports theme-defined Text types, but Heading uses a closed type union. A theme can currently emit CSS for a custom Heading type that typed application source cannot safely select.

This proposal makes the extension path explicit without adding product-specific names to the options every Astryx application receives. It also gives Heading the same per-use named weight control that Text already has.

## Scope

This is the proposed contract, not the implementation. It does not modify Heading.tsx or the theme builder. After this API is reviewed, a separate implementation PR will replace the closed HeadingType alias with the approved map-backed extension point, add the weight prop, and update theme-build declaration generation.

## Proposed direction

- Replace the closed Heading type alias in packages/core/src/Heading/Heading.tsx with a public, map-backed augmentation point exported by the Heading subpath.
- Teach theme build to generate and load augmentations for declared custom Heading types.
- Preserve the semantic-level baseline until theme CSS applies, including when an imported custom name is used under a different active theme.
- Permit partial visual roles that change any supported property; require at least one valid declaration rather than forcing size and line height.
- Add normal, medium, semibold, and bold as optional per-use Heading weights.
- Reflect explicit weight through the theming target and guarantee that it wins over type and level defaults in both runtime and built CSS.
- Never infer heading level from a visual name such as hero or title.

## Decisions

- TypeScript augmentation is program-wide after import; it cannot prove which Theme is active at a call site. Runtime styling remains theme-scoped and mismatches retain the safe level baseline.
- Text and Heading enroll a visual-role name independently. They may share the same theme-local values, and tooling may offer both mappings together, but each use is visible and confirmed.
- Component weight props remain consistent named choices across themes. The active theme decides whether bold resolves to 700, 750, or another supported value; an exact weight may also live in a custom visual type.

## Verification

- Knowledge validation

No changeset: specification only.

### DIFF
--- a/docs/specs/AST-024/spec.md
+++ b/docs/specs/AST-024/spec.md
@@ -227,12 +227,7 @@ This spec moves from `proposed` to `shipped` only when:
 
 ## Open questions
 
-- Should one declared visual-role name automatically become available to both
-  `Text` and `Heading`, or should authors enroll it independently on each component
-  whose semantics they intend to use?
-- Should a future proposal allow raw numeric variable-font weights on `Heading`
-  and `Text`, or should component props continue using portable named weights while
-  themes map those names to numeric values?
+None.
 
 ## Decision log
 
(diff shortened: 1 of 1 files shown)


cixzhang (approved):
Approved at exact head `8f5dac2372dd3869f4027638f73a284439a0aa66`. Heading keeps required `level` as the sole semantic h1–h6 decision. Visual `type` adopts the existing theme-extensible map pattern, missing active-theme custom types fall back to the selected level’s normal treatment, and `weight` matches Text without changing semantics.

PR #6025, opened 2026-09-04
docs(spec): define extensible Icon sizes

rubyycheung:
## Summary

- Defines a type-safe extension point for theme-owned Icon sizes.
- Separates reusable named sizes from one-off direct sizing and component-owned icon adjustments.
- Supports exact author-provided CSS values without requiring a generated scale.
- Requires equivalent sizing for direct SVG and registry-backed icons.
- Keeps glyph size separate from parent layout and interactive hit-target changes.

## Why

Astryx exposes four closed Icon sizes. A caller can adjust a directly owned Icon with xstyle, but a theme cannot add a reusable named size that works through IconProps and generated declarations. Icons rendered inside other components add a separate ownership problem: making the glyph larger must not silently change the surrounding component or rely on private descendant selectors.

This proposal provides a deliberate theme-owned extension path while retaining the smaller existing mechanisms for one-off and component-specific needs.

## Scope

Specification only. This PR does not modify Icon.tsx, parent component APIs, or the theme builder. A follow-up implementation will add the approved map-backed type, generated augmentation, safe fallback, and rendering parity.

## Proposed direction

- Add a public IconSizeMap consumed by Icon and renderIconSlot.
- Generate and load declarations for custom icon size names.
- Accept explicit values; formula-based generation remains optional assistance.
- Require width, height, and registry-compatible font-size behavior.
- Fall back to the built-in md box before owning theme CSS applies.
- Ask whether a request is reusable, one-off, or owned by a component slot before generating anything.
- Surface parent layout, clipping, alignment, and hit-target implications separately.

## Decisions

- The author enters one canonical size. Tooling writes width, height, and the font-size behavior required for registry-backed artwork.
- Resizing a parent component remains a separate optional change that requires confirmation.

## Remaining discussion

When a component renders an icon internally, should it expose a dedicated iconSize prop, a documented icon target, or choose between them based on whether icon sizing is part of its public per-instance layout contract?

## Verification

- Knowledge validation

No changeset: specification only.

### DIFF
--- a/docs/specs/AST-025/spec.md
+++ b/docs/specs/AST-025/spec.md
@@ -1,0 +1,262 @@
+---
+schema_version: 1
+template_version: 1
+kind: system-spec
+id: spec:AST-025
+authority: draft
+archive_reason: null
+superseded_by: null
+approved_by: null
+approved_at: null
+phase: proposed
+owners: [rubyycheung]
+affects_architecture: [architecture:theme-authoring-contract]
+affects_families: []
+affects_contributing: []
+affects_consumer_docs: [theme]
+---
+
+# Theme-extensible Icon sizes system spec
+
+## Intent
+
+Astryx ships four portable Icon sizes: `xsm`, `sm`, `md`, and `lg`. Product
+themes sometimes need another reusable glyph size, while individual screens may
+need a one-off adjustment or a different icon size inside one component slot.
+
+Those are different needs. Astryx should let a theme add a named, type-safe Icon
+size without adding that name to every theme, while continuing to support direct
+`xstyle` adjustments and component-owned icon targets for narrowly scoped cases.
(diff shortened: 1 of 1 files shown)


cixzhang (changes requested):
Icon consumers are affected when a theme can add new public size names: component layout, hit-area relationships, alignment, and composition can change under an API value that Astryx does not define consistently. Keep `IconSize` closed. Themes may change the resolved values of the existing supported size scale, but must not invent additional `size` prop values through module augmentation. This is a system-direction blocker for AST-025, not an implementation-detail fix.

rubyycheung:
Closing this proposal for now. We don’t currently need to extend Astryx’s public IconSize API. We can revisit this if a concrete cross-product need emerges.

PR #6026, opened 2026-09-04
feat(core): support custom Heading visual types

rubyycheung:
## Summary

- makes `Heading` visual types extensible through `HeadingTypeMap`
- generates theme-owned TypeScript augmentation for custom types
- requires each generated custom type to have a non-empty standalone rule
- preserves semantic heading level and a safe level-based visual fallback when theme CSS is unavailable

## Risk controls

- additive API; built-in `display-1`, `display-2`, and `display-3` remain unchanged
- no global string widening, so misspellings still fail type checking
- custom types affect styling only; heading semantics still come from `level`
- implementation is stacked on the AST-024 specification in #6024

## Verification

- Core typecheck and build
- focused Heading and theming-target tests
- CLI custom-variant generation and consumer typecheck tests

## Changeset

Included for Core and CLI.


packages/cli/api/theme/build/build.mjs
```diff
@@ -1031,6 +1037,49 @@ function validatePrivateVars(themeDef) {
...
+  }
+
+  const errors = [];
+  for (const type of customTypes) {
+    const standalone = headingRules[`type:${type}`];
+    if (
+      standalone == null ||
+      typeof standalone !== 'object' ||
+      Array.isArray(standalone) ||
+      Object.keys(standalone).length === 0
```
cixzhang: Could this check that the standalone rule actually produces at least one CSS declaration, rather than only that the object has a key? `{'type:ghost': {':hover': {}}}` passes this guard, `theme build` succeeds and emits `HeadingTypeMap['ghost']`, but the generated CSS contains no `.astryx-heading.ghost` rule. A theme author then gets a type-safe value that can never apply styling. Please fail the build when the standalone rule emits no usable declaration, while continuing to allow genuinely partial roles.

[Reviewed by Robohands]

packages/cli/api/theme/build/build.mjs
```diff
@@ -1120,6 +1169,15 @@ export async function themeBuild(
     );
   }
 
+  const customHeadingErrors = validateCustomHeadingTypes(themeDef);
```
cixzhang, line 1282: Could this validation also fail when the installed Core package does not expose `HeadingTypeMap`? The CLI declares Core as an optional `*` peer, so independent upgrades are valid. With this CLI change and the prior Core package, a `type:hero` theme build reports success and emits the CSS but no `.variants.d.ts`; application source still rejects `type="hero"`. Please turn that mismatch into an actionable build error (and cover the older-Core case) instead of silently skipping the augmentation.

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks for carrying the approved Heading type extension through the public `Heading` subpath and preserving semantic `level` as the fallback. I found two cases where `theme build` can still report success while generating an unusable custom type; details are inline.

Before the next review, please also retarget this PR from the already-merged spec branch to `main`. The current head only has CLA, Vercel, and `spec-owner-approval`; the required `pr-a11y` and `pr-visual` checks have not run.

[Reviewed by Robohands]

rubyycheung:

- custom Heading validation now requires a real emitted CSS declaration, including inside pseudo rules; empty pseudo-only rules fail the build
- builds now fail with an actionable error when the installed Core does not expose HeadingTypeMap
- retargeted the PR to main

Verified with the focused CLI variants test (6 passed).

rubyycheung:
Also rebased this branch onto main and resolved the branch conflict. Focused CLI variants tests still pass (6/6).

packages/cli/api/theme/build/build.mjs
rubyycheung: Addressed in 61c04557c: validation now recursively requires at least one emitted declaration, including within pseudo rules, and the regression test rejects `':hover': {}`.

packages/cli/api/theme/build/build.mjs
rubyycheung: Addressed in 3159c8562: custom Heading types now fail with an actionable ERR_THEME_INVALID when the installed Core lacks `HeadingTypeMap`, with an older-Core regression test.

cixzhang (changes requested):
Thanks—the base-theme path and both previous blockers are fixed, but surface-specific overrides still break the CSS/type contract. `onDark.components.heading['type:hero']` builds successfully and emits the `.astryx-heading.hero` CSS, but no `.variants.d.ts`; a consumer importing the generated theme still gets TS2322 for `type="hero"`. This can ship a theme role that exists in CSS but remains unusable through Heading's public API. Please collect and validate custom Heading types from base, `onDark`, and `onLight` components before generating declarations.

[Reviewed by Robohands]

rubyycheung:
Fixed in de3d2c266c.

The theme builder now collects custom component values across the base, `onDark`, and `onLight` component layers before generating `.variants.d.ts`. Custom Heading validation and the installed-Core augmentation check use that same theme-wide set. A standalone definition may live in any layer, so legitimate combined surface refinements remain valid.

Added a regression that builds dark-only `type:hero` and light-only `type:editorial`, verifies both are emitted in `HeadingTypeMap`, and type-checks real `<Heading>` consumers. The full CLI suite passes: 221 files / 3,254 tests.

cixzhang (changes requested):
Thanks—the surface-specific custom Heading path now works across generated CSS and TypeScript, and the earlier contract failures are fixed. The older-Core guard also behaves correctly, but the regression test promised in the last response is still absent: this suite never substitutes a pre-`HeadingTypeMap` Core declaration. Since CLI and Core can be upgraded independently, please add that test so a future refactor cannot restore the silent-success failure.

[Reviewed by Robohands]

rubyycheung:
Added the requested older-Core regression in d3bcf9af7d. The test substitutes a legacy public Heading declaration without `HeadingTypeMap`, uses an `onDark`-only custom Heading type, asserts the actionable build failure, and confirms neither CSS nor `.variants.d.ts` is written. The focused compatibility + variant suites pass (8/8), as does `sync:api-types`.

cixzhang (approved):
Thanks—this is good. The older-Core regression now covers the public declaration boundary, and the full custom Heading path remains coherent across theme surfaces, types, fallback, and built-ins.

[Reviewed by Robohands]

PR #6027, opened 2026-09-04
feat(core): add Heading weight override

rubyycheung:
## Summary

- adds an optional named `weight` override to `Heading`
- reflects the selected weight for theme targeting
- emits theme-layer weight rules after visual-type rules so an explicit weight reliably wins
- documents the override separately from custom Heading visual types

## Risk controls

- separate from custom Heading-type support so it can be reviewed or deferred independently
- optional and additive; existing Heading output is unchanged when `weight` is omitted
- uses the existing `normal`, `medium`, `semibold`, and `bold` weight tokens
- stacked on the custom Heading-type implementation

## Verification

- Core typecheck and production build
- 492 focused Heading, theme-rule, and theming-target tests

## Changeset

Included for Core.


packages/core/src/theme/generateThemeRules.ts
```diff
@@ -372,11 +372,50 @@ export function generateThemeRules(theme: DefinedTheme): string[] {
   //    themed type's font-size across the astryx-base → astryx-theme layers)
   generateSizeOverrides(theme.components || {}, parts);
 
+  // 6. Heading `weight`-prop overrides. These are emitted after authored type
+  //    rules in the same layer so an explicit prop remains an override even
+  //    when a custom visual type declares its own default weight.
+  generateHeadingWeightOverrides(theme.components || {}, parts);
```
cixzhang, line 378: Could this emit after `onDark`/`onLight` type rules too? There, `weight="bold"` still computes from the custom type.

packages/core/src/theme/generateThemeRules.ts
```diff
@@ -372,11 +372,50 @@ export function generateThemeRules(theme: DefinedTheme): string[] {
...
+  if (!('heading' in components)) {
+    return;
+  }
+
+  for (const [weightName, weightValue] of Object.entries(
+    HEADING_WEIGHT_TOKEN_MAP,
+  )) {
+    const suffix = parseStyleKey(`weight:${weightName}`);
+    parts.push(
+      `  ${componentClassSelector('heading', suffix)} { font-weight: ${weightValue}; }`,
```
cixzhang: This also overwrites a theme-authored `weight:bold` value; that targeted theme rule should stay authoritative.

cixzhang (changes requested):
Thanks for keeping this weight work separate. The new API and named-token mapping look right, and the omitted path stayed unchanged. I’m requesting changes because the cascade still violates the promised precedence in two reachable theme states: `onDark` custom types can beat an explicit weight, and a theme’s own `weight:bold` rule is overwritten by the generated token rule. Both reproduce in real Chromium through runtime-injected and built theme CSS. Please preserve explicit weight over level/type defaults on every media surface while keeping targeted theme overrides authoritative.

[Reviewed by Robohands]

rubyycheung:

- explicit Heading weight overrides are emitted on dark/light media surfaces after media-specific type rules
- generated weight rules no longer overwrite theme-authored weight targets
- added regression coverage for both cascade cases
- retargeted the PR to main

Verified with the focused generateThemeRules test (58 passed).

rubyycheung:
Also rebased this branch onto main and resolved the branch conflict. Focused Heading theme tests still pass (58/58).

packages/core/src/theme/generateThemeRules.ts
rubyycheung: Addressed in 4b4567f8d: named weight overrides are emitted after `onDark`/`onLight` component rules, with a regression test covering the ordering.

packages/core/src/theme/generateThemeRules.ts
rubyycheung: Addressed in 4b4567f8d: generated weight rules skip any theme-authored target such as `weight:bold`, preserving the authored value; regression coverage asserts only the authored rule is emitted.

cixzhang (changes requested):
Thanks, the missing media fallback is fixed, but the targeted weight precedence still depends on theme key order. When a theme declares `weight:bold` before `type:hero`, `<Heading type="hero" weight="bold">` computes to the type weight (300 on the main surface and 350 on `onDark`) instead of the theme's targeted 900/800, in both runtime-injected and built CSS. `generateHeadingWeightOverrides` skips the authored weight rule, so it is never moved after the type rule (`packages/core/src/theme/generateThemeRules.ts:415`). Please preserve each surface's effective `weight:*` value after type rules regardless of authored key order.

[Reviewed by Robohands]

rubyycheung:
Addressed the remaining precedence issue in 997804ebda:

- re-emits the effective theme-authored `weight:*` value after type rules, independent of theme key order
- applies the same ordering to `onDark`/`onLight`, with surface values overriding inherited main-surface values
- covers both runtime-generated CSS and CLI-built CSS regressions

Verified with the focused Core/CLI suites (67 passed) and a Core production build.

rubyycheung:
Fixed in 997804ebda, with the lint follow-up in 0de78dcf0b.

The generator now re-emits each effective standalone `weight:*` rule after Heading type rules, independent of authored key order. On media surfaces, the surface-authored weight wins, then the main-surface authored weight, then the built-in token fallback. Combined selectors do not suppress the generic weight override.

The regression uses the reported ordering (`weight:bold` before `type:hero`) and verifies computed CSS values 900 on the main surface and 800 on `onDark`.

Verified: focused theme/CLI suites 89/89, Core typecheck and package build, targeted ESLint, and repository checks.

cixzhang (changes requested):
Thanks, the weight path itself now passes: runtime and built CSS preserve named-token and authored values, and omitted weight stays type-derived. One blocker: this head is not stacked on current #6026; the branches diverge, and this tree drops #6026’s media-surface custom-type generation and tests. Please rebase/retarget this as the weight-only slice above #6026, then rerun the combined checks.

[Reviewed by Robohands]

rubyycheung:
Restacked and retargeted this weight-only PR onto #6026 (`codex/heading-type-implementation`). The branch now preserves #6026’s media-surface type generation/tests and adds only the Heading weight slice. Verified after the rebase: focused suites (90 tests), Core typecheck, Core build, targeted lint, CLI API-type sync, repository contract checks, and `git diff --check` all pass.

cixzhang (approved):
Thanks, this now looks good. The weight-only slice cleanly reapplies to current main, and runtime/built CSS preserve theme-authored precedence, named tokens, omitted behavior, and heading semantics.

[Reviewed by Robohands]

PR #6028, opened 2026-09-04
feat(core): support theme-owned Icon sizes

rubyycheung:
## Summary

- makes `Icon` size names extensible through `IconSizeMap`
- generates typed theme augmentation for custom Icon sizes
- applies custom names consistently to direct SVG and registry icons
- requires complete width, height, and font-size styling for generated custom names

## Risk controls

- additive API; `xsm`, `sm`, `md`, and `lg` remain unchanged
- a custom name keeps the existing `md` baseline until its theme CSS loads
- incomplete and combination-only custom size definitions fail the theme build
- does not change parent component size props or resize icons inside parent components
- implementation is stacked on the AST-025 specification in #6025

## Verification

- Core typecheck and production build
- 38 focused Icon component tests
- 6 CLI generation and real consumer typecheck tests

## Changeset

Included for Core and CLI.

### DIFF
--- a/.changeset/extensible-icon-sizes.md
+++ b/.changeset/extensible-icon-sizes.md
@@ -1,0 +1,9 @@
+---
+'@astryxdesign/core': patch
+'@astryxdesign/cli': patch
+---
+
+[feature] Allow generated themes to add typed, theme-owned Icon size names while keeping
+the existing built-in sizes and a safe `md` fallback when theme CSS is absent.
+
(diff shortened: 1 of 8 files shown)


rubyycheung:
Closing this implementation PR because the corresponding proposal is deferred. We don’t currently need to extend Astryx’s public IconSize API. We can revisit this if a concrete cross-product need emerges.

PR #5963, opened 2026-09-03
feat(theme): add appearance nesting tokens

cixzhang:
## Why

Themes need shared stacking bands so nested component chrome and ordinary floating surfaces do not each maintain unrelated z-index numbers.

## What

- Add typed integer appearance tokens for container nesting (10) and layer nesting (50).
- Migrate all 47 positive or dynamic z-index assignments in Core, Lab, Charts, and RichText to the shared bands.
- Preserve local ordering with band-relative offsets for nested surfaces, overlapping schedule events, and sibling drawers.\n- Keep ChartTooltip beside its owning chart and promote it with a native manual popover; the layer token remains its reduced-browser fallback.
- Keep native top-layer-only roots, zero-valued isolation roots, negative underlays, and consumer-authored values unchanged.
- Export and document the token family, and add a source guard against new component-local stacking numbers.

Current main already contains no z-index override in Neutral, and the former Rose theme package is no longer present, so there are no obsolete theme overrides to remove.

## Risk

Stacking ownership moves from component-local numbers to theme tokens. Native popovers and modal dialogs continue to rely on the browser top layer. A theme can now adjust the two shared bands while component-local ties and relative ordering remain stable.

## Testing

- 718 focused tests on the exact integrated head
- Core, Lab, Charts, and RichText builds and typechecks
- Typed-token contract and source-guard tests
- Generated token documentation drift check
- `pnpm check:repo`

No Changeset is included because this API is not intended to trigger a package version bump yet.

PR #5909, opened 2026-09-02
docs: add playground defaults for Selector, MultiSelector, ChatToolCalls (#5904, #5903, #5887)

Rijul202:
Closes #5904, closes #5903, closes #5887.

### Problem

Three component Properties previews render the placeholder *"Interactive preview needs required props that cannot be generated automatically"* because a required object/array prop has no `playground.defaults` fixture and the generic preview cannot synthesize it:

| Component | Issue | Missing required prop |
| --- | --- | --- |
| `Selector` | #5904 | `options: SelectorOption[]` |
| `MultiSelector` | #5903 | `options` (and `value`) |
| `ChatToolCalls` | #5887 | `calls: ChatToolCallItem[]` |

### Fix

Add a minimal representative `playground.defaults` fixture to each doc, matching each prop's documented shape:

- `Selector` / `MultiSelector` — `options` as `{value, label}` objects (the shape used in the components' own tests); MultiSelector also gets `value: []`.
- `ChatToolCalls` — `calls` as `ChatToolCallItem` objects (`name` is the only required field; `status: 'complete'` etc. included for a representative row).

This follows the established pattern in other docs (`Citation`, `NavMenu`, `ButtonGroup`). Placement matches each doc's structure — top-level `docs.playground` for the multi-component `MultiSelector` (as in `ButtonGroup`).

### Why it works

In `apps/docsite/src/components/component-detail/interactiveState.ts`, `buildInitialState()` seeds state from `playground.defaults` before `getMissingRequiredProps()` runs; that function only reports required props where `state[name] === undefined`. With the fixtures present, the previously-missing props are defined, so the missing-props placeholder is replaced by the real component.

Docs-only change (three `.doc.mjs` files); no public API or runtime change. Each fixture validated to parse and expose the expected `playground.defaults` keys.

PR #6035, opened 2026-09-04
fix(docsite): bridge Selector's onChange preview even without a seeded value

cixzhang:
Fixes the Selector Properties-preview regression exposed by #5909. Does not touch anything from #5963.

### Problem

#5909 seeded `options`/`label` playground defaults for `Selector` so its Properties preview stops showing the "missing required props" placeholder. It correctly left `value` unseeded — "closed with no value" is itself a required representative state (Selector.spec.md, FR1: *"Selecting an enabled option updates the one selected value..."*).

That change exposed a latent bug in `buildRuntimePreviewState` (`apps/docsite/src/components/component-detail/interactiveState.ts`): it only bridges a controlled callback back into playground state when its target prop already has a value in `state`. Since Selector's `value` has no seeded default, `onChange` was never wired up, so the live preview stayed frozen on the placeholder no matter which option you clicked — Selector's own `value`/`onChange` contract has no internal uncontrolled fallback (`Selector.tsx` renders strictly from the `value` prop), so nothing else could make the trigger update.

### Fix

Generalized the bridge instead of patching Selector specifically: the literal `value`/`onChange` pair — a component's one primary controlled value — now bridges whenever `target` names a real prop, even an optional one with no seeded default.

Secondary paired callbacks (`onIndexChange`, `onPageSizeChange`, `onOpenChange`, ...) keep the original opt-in behavior: they only bridge once their target already has a value in state. This preserves each preview's chosen representative starting state — verified by the existing Lightbox gallery-index and overlay tests, unchanged and still passing.

No arbitrary selected value was seeded anywhere; Selector's playground still starts "closed with no value," and selecting an option now correctly updates it.

### Testing

- New focused regression test: `bridges Selector onChange even though its optional value prop is not seeded` — confirmed red on unfixed source (`onChange` was `undefined` in the preview's runtime props), green after the fix.
- Full docsite suite: 34 files, 475 tests passing.
- `tsc --noEmit`: no new errors (pre-existing `@astryxdesign/theme-*/built` errors are from unbuilt theme packages in this worktree, present identically before this change).
- `eslint` on both changed files: clean.
- `check:package-boundaries`: clean.
- Pre-commit `check:repo` (knowledge, sync, i18n, changesets, etc.): clean.

### Scope

Touches only docsite preview infrastructure and its test:
- `apps/docsite/src/components/component-detail/interactiveState.ts`
- `apps/docsite/src/__tests__/component-preview-state.test.ts`

No Selector source, doc, or spec changes were necessary — the fix is general, not Selector-specific. Excludes all `#5963` (theme appearance-nesting tokens) work.

### Residual risk

The generalized rule also newly enables bridging for any other component's optional `value`/`onChange` pair that currently has no playground default (previously frozen the same way Selector was). I did not find any component where this is undesirable — for HTML-backed inputs (e.g. `TextInput`) this was already visually working via native uncontrolled-input fallback, so this is additive rather than a behavior change — but it's a wider blast radius than Selector alone, worth a quick visual pass across the docsite Properties tabs if anyone wants extra confidence before merge.

### DIFF
--- a/apps/docsite/src/__tests__/component-preview-state.test.ts
+++ b/apps/docsite/src/__tests__/component-preview-state.test.ts
@@ -444,6 +444,33 @@ describe('component detail preview state', () => {
     expect(onPropChange).toHaveBeenCalledWith('value', 42);
   });
 
+  it('bridges Selector onChange even though its optional value prop is not seeded (#5909 follow-up)', () => {
+    const knobs = pickPrimaryProps('Selector', [
+      prop({name: 'label', type: 'string', required: true}),
+      prop({name: 'options', type: 'SelectorOption[]', required: true}),
+      prop({name: 'value', type: 'string'}),
+      prop({name: 'onChange', type: '(value: string) => void'}),
+    ]);
+
+    const state = buildInitialState(knobs, {
+      defaults: {
+        label: 'Fruit',
+        options: [
+          {value: 'apple', label: 'Apple'},
+          {value: 'orange', label: 'Orange'},
+        ],
+      },
+    });
+    expect(state.value).toBeUndefined();
+    expect(getMissingRequiredProps(knobs, state)).toEqual([]);
+
+    const onPropChange = vi.fn();
+    const runtimeState = buildRuntimePreviewState(state, onPropChange, {knobs});
+
+    (runtimeState.onChange as (value: string) => void)('orange');
+    expect(onPropChange).toHaveBeenCalledWith('value', 'orange');
+  });
+
   it('bridges a Tokenizer removal back to its controlled value', () => {
     const knobs = pickPrimaryProps('Tokenizer', [
       prop({name: 'label', type: 'string', required: true}),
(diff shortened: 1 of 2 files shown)


cixzhang (commented):

The Selector path is fixed, but this exact head still leaves the current Switch and Tokenizer Properties interactions inert. At `interactiveState.ts:341`, the bridge only runs when the callback parameter names a prop; Switch emits `checked` and Tokenizer emits `items`, so `onChange` remains undefined. This branch also conflicts with main in both touched files.

Could you update from main and preserve both behaviors: unseeded optional `value`/`onChange` pairs and the existing `onChange` fallback to controlled `value`, with Selector plus Switch/Tokenizer runtime coverage?

[Reviewed by Robohands]

cixzhang (commented):
Semantic verdict: approve

This head keeps Selector’s empty start interactive and restores the existing `onChange` fallback for Switch and Tokenizer. Focused tests and the exact-head preview pass.

[Reviewed by Robohands]

PR #6037, opened 2026-09-04
feat(core): add elevation prop to ToggleButton

### LINKED ISSUE #6012: ToggleButton has no `elevation` prop though it renders `data-elevation`
### Summary

`Button` accepts `elevation="none" | "low" | "med" | "high"` and renders a `data-elevation` attribute that the theme keys the shadow scale off. `ToggleButton` renders the **same** `astryx-button` element with the **same** `data-elevation` attribute, but exposes no `elevation` prop — so the value is permanently `none` and a `ToggleButton` cannot be raised.

This reads as a missing prop pass-through rather than a missing feature: the shadow machinery is already on the element.

### Measured

`@astryxdesign/core@0.5.2`, Chromium, 390px viewport, computed styles:

| Rendered | `data-elevation` | computed `box-shadow` |
| --- | --- | --- |
| `<Button variant="secondary" elevation="med">` | `med` | `oklch(0 0 0 / 0.05) 0 2px 4px, oklch(0 0 0 / 0.1) 0 4px 12px, …` |
| `<ToggleButton>` — `size` sm / md / lg, standalone | `none` | `none` |
| `<ToggleButton>` inside `<ToggleButtonGroup size="lg">` | `none` | `none` |

The toggle's element carries `class="astryx-button … astryx-toggle-button"`, i.e. the same base component as `Button`.

### Also missing from the theming surface

`astryx component Button` lists `data-size`, `data-variant`, `data-elevation`.
`astryx component ToggleButton` lists only `data-is-pressed` — so the gap cannot be closed in `defineTheme` either.

### Why it matters

`ButtonGroup` documents elevation as a first-class pattern ("A grouped action bar raised with `elevation="med"`. The connected buttons share one surface, so the shadow lifts them as a unit."). A toolbar of toggles is the same shape and cannot follow it. The workaround is to elevate a wrapper, which puts the shadow on a different box than the interactive surface.

### Suggested

Accept `elevation` on `ToggleButton` (and a group-level default on `ToggleButtonGroup`, mirroring how `size` already defaults from the group), passing through to the `data-elevation` the element already renders.

ManoharPaturi:
Body:

Fixes #6012

## Summary

`ToggleButton` renders the same `astryx-button` element with the same
`data-elevation` machinery as `Button`, but exposes no `elevation` prop — so a
toggle can never be raised. This adds the prop as a straight pass-through,
mirroring Button's API:

- `elevation?: Elevation` with Button's JSDoc semantics, default `'none'`
- Inside a `ToggleButtonGroup` the prop is ignored — the group's surface
  lifts as one unit, the same contract Button applies within a ButtonGroup
  (ToggleButtonGroup does not render a ButtonGroup, so Button's own
  neutralization would not fire; the mirror lives here)

## Sync updates (per the file header)

- `ToggleButton.doc.mjs` — prop entry matching Button's wording
- `apps/storybook/stories/ToggleButton.stories.tsx` — `WithElevation` story
- `ToggleButton.test.tsx` — 3 tests: per-level `data-elevation` reflection,
  flat default, and ignored-inside-group
- Changeset: `[component]` patch

## Testing

- `vitest run packages/core/src/ToggleButton/ToggleButton.test.tsx` — 29 passed
- `tsc --noEmit` (core) — clean; storybook typecheck unchanged (only
  pre-existing `@astryxdesign/vega` resolution errors)
- eslint on touched files — 0 errors; both warnings pre-exist at HEAD

### DIFF
--- a/packages/core/src/ToggleButton/ToggleButton.doc.mjs
+++ b/packages/core/src/ToggleButton/ToggleButton.doc.mjs
@@ -53,7 +53,7 @@ export const docs = {
       name: 'elevation',
       type: "'none' | 'low' | 'med' | 'high'",
       description:
-        'Resting shadow depth for floating (FAB-style) toggle buttons, mirroring Button. `none` is the default flat button; `low`/`med`/`high` map to the shadow token scale. Ignored inside a ToggleButtonGroup, where the surface elevation is owned by the group.',
+        'Resting shadow depth for floating (FAB-style) toggle buttons, mirroring Button. `none` is the default flat button; `low`/`med`/`high` map to the shadow token scale. Applies inside a ToggleButtonGroup as well — grouped children retain their own elevation.',
       default: "'none'",
     },
     {
--- a/packages/core/src/ToggleButton/ToggleButton.doc.mjs
+++ b/packages/core/src/ToggleButton/ToggleButton.doc.mjs
@@ -50,6 +50,13 @@ export const docs = {
       default: "'md'",
     },
     {
+      name: 'elevation',
+      type: "'none' | 'low' | 'med' | 'high'",
+      description:
+        'Resting shadow depth for floating (FAB-style) toggle buttons, mirroring Button. `none` is the default flat button; `low`/`med`/`high` map to the shadow token scale. Ignored inside a ToggleButtonGroup, where the surface elevation is owned by the group.',
+      default: "'none'",
+    },
+    {
       name: 'isDisabled',
       type: 'boolean',
       description: 'Whether the button is disabled.',
--- a/packages/core/src/ToggleButton/ToggleButton.doc.mjs
+++ b/packages/core/src/ToggleButton/ToggleButton.doc.mjs
@@ -17,7 +17,7 @@ export const docs = {
   theming: {
     targets: [
       {className: 'astryx-toggle-button-group'},
-      {className: 'astryx-toggle-button', states: ['isPressed']},
+      {className: 'astryx-toggle-button', states: ['isPressed'], visualProps: ['elevation']},
     ],
   },
   description: 'A button that toggles between pressed and unpressed states. Thin wrapper over Button with controlled toggle pattern, icon swap, and font weight emphasis.',
--- a/packages/core/src/ToggleButton/ToggleButton.tsx
+++ b/packages/core/src/ToggleButton/ToggleButton.tsx
@@ -25,6 +25,7 @@ import * as stylex from '@stylexjs/stylex';
 import {colorVars, fontWeightVars} from '../theme/tokens.stylex';
 
 import {Button, type ButtonSize} from '../Button';
+import type {Elevation} from '../utils/types';
 import {useToggleButtonGroup} from './ToggleButtonGroup';
 import type {BaseProps} from '../BaseProps';
 import {themeProps} from '../utils/themeProps';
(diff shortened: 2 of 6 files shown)


packages/core/src/ToggleButton/ToggleButton.tsx
```diff
@@ -330,6 +342,10 @@ export function ToggleButton({
       label={label}
       variant="ghost"
       size={size}
+      // Inside a group the group owns the surface's elevation (the same
+      // contract Button applies within a ButtonGroup), so a grouped toggle
+      // reflects the tier it actually paints rather than a per-button prop.
+      elevation={group ? 'none' : elevation}
```
cixzhang: Please keep each child’s elevation in spaced ToggleButtonGroup and update the group test/docs to match.

packages/core/src/ToggleButton/ToggleButton.doc.mjs
```diff
@@ -49,6 +49,13 @@ export const docs = {
       description: 'Button size. Defaults to group size when inside a group.',
       default: "'md'",
     },
+    {
+      name: 'elevation',
+      type: "'none' | 'low' | 'med' | 'high'",
+      description:
+        'Resting shadow depth for floating (FAB-style) toggle buttons, mirroring Button. `none` is the default flat button; `low`/`med`/`high` map to the shadow token scale. Ignored inside a ToggleButtonGroup, where the surface elevation is owned by the group.',
```
cixzhang: Please add `elevation` to this target’s `visualProps`; theme tooling currently cannot discover the rendered state.

cixzhang (changes requested):
Thanks — standalone tiers and pressed states look correct across all seven themes in light and dark. Grouped children still need to retain their own elevation, and ToggleButton’s theme metadata needs to expose it.

[Reviewed by Robohands]

ManoharPaturi:
Thanks @cixzhang — both addressed in 0d29ddc:

1. **Grouped children now retain their own elevation.** The `group ? "none" : elevation` neutralization is gone; `elevation={elevation}` passes straight through, and the grouped-case test now asserts `data-elevation="high"` inside a ToggleButtonGroup.
2. **Theme metadata exposes it.** `themeProps("toggle-button", {isPressed, elevation})` now publishes the tier alongside the pressed state (doc entry and JSDoc updated to match — no more "ignored inside a group" wording).

Also added a `core/ToggleButton` verified-N/A entry to the RTL audit registry (`verified-not-applicable.json`) — that was the pr-rtl coverage gap: the stories have no directional glyphs or side-specific geometry (elevation is a symmetric box-shadow), so LTR/RTL renders are mirror-identical. 29/29 ToggleButton tests, core typecheck, and check:changesets all green on the updated branch.

cixzhang (changes requested):
Thanks — grouped children now retain their own elevation, and the 14-case pixel matrix is clean across all seven themes in light and dark. The theming metadata ask is still open: `elevation` is absent from `astryx-toggle-button`’s `visualProps`, so `themingTargets.test.ts` fails and theme tooling cannot discover this rendered state. Please add it to that target.

[Reviewed by Robohands]

cixzhang (approved):
Thanks — this now matches the button-family contract. Grouped children retain elevation, `astryx-toggle-button` exposes it to themes, and the default path remains pixel-identical.

[Reviewed by Robohands]

PR #6038, opened 2026-09-04

### LINKED ISSUE #6021: Link's inline-flex anchor silently defeats an ancestor Text maxLines clamp
### Summary

```tsx
<Text maxLines={2}>
</Text>
```

### Measured

| --- | --- | --- | --- |

### Why it matters

### Suggested

ManoharPaturi:
Body:

Fixes #6021

## Summary

A plain `<Text maxLines={2}><Link>…</Link></Text>` silently did not truncate:
the anchor was `display: inline-flex`, and an inline-flex box establishes its
own formatting context that an ancestor `-webkit-line-clamp` cannot reach.

This changes the plain anchor to `display: inline`, so it participates in the
surrounding line boxes and an ancestor clamp truncates it normally. The flex
layout is not lost where it is load-bearing — it moves to a dedicated style
applied only when needed:

- **external links** (`isExternalLink` + anchor form) still lay out as
  `inline-flex` with `alignItems: center` and the icon gap, exactly as before
- **the button-rendered form** keeps its previous `inline-flex` layout

An external link inside a clamped `Text` still cannot be truncated by the
ancestor (the icon needs the flex layout), so this also documents the
composition on both sides — Link guidance (do: plain links clamp; don't:
expect it for external links — pass `maxLines` to the Link) in all three doc
variants (EN/zh/dense) and matching Text guidance (EN/zh + dense).

## Tests

Three new cases in `Link.test.tsx` assert the emitted display declarations
per form (via the injected-CSS helper pattern used by forced-colors tests):

- plain anchor → `inline`, and **not** `inline-flex`
- external-link anchor → keeps `inline-flex`
- button-rendered form → keeps `inline-flex`

## Testing

- `vitest run packages/core/src/Link` — 63 passed
- `vitest run packages/core/src/Text` — 189 passed
- `vitest run internal/eslint-plugin-astryx` — 640 passed (docblock format)
- `tsc --noEmit` (core) clean, eslint clean, `check:changesets` and
  `check:sync` pass

### DIFF
--- a/packages/core/src/Link/Link.tsx
+++ b/packages/core/src/Link/Link.tsx
@@ -52,9 +52,12 @@ import {useTranslator} from '../i18n';
  */
 const styles = stylex.create({
   base: {
-    display: 'inline-flex',
-    alignItems: 'center',
-    gap: spacingVars['--spacing-0-5'],
+    // `inline` (not `inline-flex`) so the anchor participates in the
+    // surrounding line boxes and an ancestor clamp (e.g. <Text maxLines>)
+    // can truncate it. An inline-flex box establishes its own formatting
+    // context, which an ancestor -webkit-line-clamp cannot reach —
+    // <Text maxLines={2}><Link>…</Link></Text> silently did nothing.
+    display: 'inline',
     fontFamily: 'inherit',
     fontSize: 'inherit',
     lineHeight: 'inherit',
@@ -83,6 +86,17 @@ const styles = stylex.create({
     padding: 0,
     pointerEvents: 'auto',
     position: 'relative',
+  },
+  /**
+   * Flex layout for the cases that need it: external links append an icon
+   * after the text (icon centering + gap), and the button-rendered form
+   * keeps its previous inline-flex layout. On plain anchors this is
+   * deliberately NOT applied — see the `base` display note.
+   */
+  flexLayout: {
+    display: 'inline-flex',
+    alignItems: 'center',
+    gap: spacingVars['--spacing-0-5'],
   },
   hasUnderline: {
     textDecoration: 'underline',
@@ -329,6 +343,7 @@ export function Link({
   // render as a <button> with link styling for semantic correctness.
   const renderAsButton =
     role === 'button' || (role === 'inert' && href == null);
+  const isExternalWithIcon = isExternalLink && !renderAsButton;
 
   const sharedContent = (
     <>
@@ -341,7 +356,7 @@ export function Link({
         maxLines={maxLines}>
         {children}
       </Text>
-      {isExternalLink && !renderAsButton && (
+      {isExternalWithIcon && (
         <>
           <Icon icon="externalLink" size="xsm" color="inherit" />
           <VisuallyHidden>{newTabLabel}</VisuallyHidden>
@@ -366,6 +381,7 @@ export function Link({
           themeProps('link', {color}),
           focusOutlineProps.focusVisible(
             styles.base,
+            styles.flexLayout,
             styles.buttonReset,
             linkColorStyles[color],
             hasUnderline && styles.hasUnderline,
@@ -398,10 +414,11 @@ export function Link({
           themeProps('link', {color}),
           focusOutlineProps.focusVisible(
             styles.base,
+            isExternalWithIcon && styles.flexLayout,
             linkColorStyles[color],
             hasUnderline && styles.hasUnderline,
             isStandalone && styles.standalone,
-            styles.disabled,
+            isDisabled && styles.disabled,
             xstyle,
           ),
           className,
@@ -426,6 +443,7 @@ export function Link({
           themeProps('link', {color}),
           focusOutlineProps.focusVisible(
             styles.base,
+            isExternalWithIcon && styles.flexLayout,
             linkColorStyles[color],
             hasUnderline && styles.hasUnderline,
             isStandalone && styles.standalone,
--- a/packages/core/src/Link/Link.tsx
+++ b/packages/core/src/Link/Link.tsx
@@ -344,6 +344,15 @@ export function Link({
   const renderAsButton =
     role === 'button' || (role === 'inert' && href == null);
   const isExternalWithIcon = isExternalLink && !renderAsButton;
+  // The plain anchor stays `inline` so an ancestor clamp (<Text maxLines>)
+  // can truncate it — that is the composition this PR fixes. But a Link that
+  // establishes its own box (`display="block"`, or clamping itself via
+  // `maxLines`, or carrying the external-link icon) must keep the inline-flex
+  // root: an inline anchor around a block-level child computes a focus
+  // outline that paints nothing in Chromium, losing keyboard focus
+  // visibility on those forms.
+  const needsRootBox =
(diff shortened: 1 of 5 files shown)


parinith-web:
Hey @ManoharPaturi, nice fix — the base/flexLayout split is exactly the right shape for this.

I took a stab at the one thing Copilot flagged (the `displayDeclarationsFor` helper in `Link.test.tsx` matching classes via `String.includes` on raw selector text, which can mis-match on hashed atomic class names). Reworked it to walk the real CSSOM instead — parses `selectorText` into class tokens and compares those exactly against the element's `classList`, reading the value via `rule.style.getPropertyValue('display')` rather than a regex over the rule body. Also descends into grouping rules like `@media` so nothing nested gets missed.

Ran it against your branch locally: all 63 Link tests, 189 Text tests, and 640 eslint-plugin-astryx tests still pass, and `tsc --noEmit` is clean.

Opened it as a PR on top of your branch here: https://github.com/facebook/astryx/pull/6043 — happy to fold it directly into this PR instead if you'd rather take the diff yourself, just say the word. Wanted to flag it here first rather than just have a second PR show up with no context.

packages/core/src/Link/Link.tsx
```diff
@@ -52,9 +52,12 @@ import {useTranslator} from '../i18n';
...
   base: {
-    display: 'inline-flex',
-    alignItems: 'center',
-    gap: spacingVars['--spacing-0-5'],
+    // `inline` (not `inline-flex`) so the anchor participates in the
+    // surrounding line boxes and an ancestor clamp (e.g. <Text maxLines>)
+    // can truncate it. An inline-flex box establishes its own formatting
+    // context, which an ancestor -webkit-line-clamp cannot reach —
+    // <Text maxLines={2}><Link>…</Link></Text> silently did nothing.
+    display: 'inline',
```
cixzhang, line 60: `display="block"` now splits this inline anchor around its block child, so its keyboard focus outline disappears.

cixzhang (changes requested):
Existing block Links lose keyboard focus visibility in this change. In Chromium, `<Link display="block">` changes from a 124px `inline-flex` focus box to an inline anchor split around a 260px block child; the 2px focus outline computes but paints nothing. Please preserve the old focusable box for block/own-clamped Link variants while keeping only unclamped plain anchors inline.

[Reviewed by Robohands]

ManoharPaturi:
@parinith-web thank you for this — the CSSOM-token matching is a real improvement over my substring approach (exact class comparison + proper descent into grouping rules), and I appreciate you flagging it here first rather than surprising us with a stacked PR. Folded your commit into this branch as 1abf92c (kept your authorship). #6043 can be closed whenever convenient — all 63 Link tests green on the updated branch.

cixzhang (changes requested):
Thanks for tightening the test helper. The earlier keyboard-focus regression is still present in Chromium: both `<Link display="block">` and `<Link maxLines={2}>` compute a 2px outline but paint no visible focus box because the anchor remains inline around a block-level Text child. Please preserve a visible root focus box for those forms while keeping only the ancestor-clamped plain anchor inline.

[Reviewed by Robohands]

ManoharPaturi:
Thanks @cixzhang — fixed in 2a36cb1. The inline anchor is now reserved for the one composition this PR exists for: the unclamped plain Link under an ancestor `<Text maxLines>`. Any Link that establishes its own box keeps the inline-flex root and its visible focus outline:

- `display="block"` → inline-flex root
- own `maxLines > 0` (self-clamping) → inline-flex root
- external-link icon form → inline-flex root (unchanged)
- button-rendered form → inline-flex root (unchanged)

So `<Link display="block">` and `<Link maxLines={2}>` compute and paint the 2px focus box exactly as before this PR, while `<Text maxLines={2}><Link>plain</Link></Text>` — with no props on the Link — stays inline and truncable by the ancestor clamp. (One subtlety caught by the new tests: `maxLines` destructures with a default of `0`, so the own-clamp condition is `maxLines > 0`, not a null check.)

Two new test cases assert the block and self-clamped forms keep `inline-flex`, on top of the existing plain-inline case. 254/254 Link + Text tests, core typecheck, and the full check:repo battery green on the updated branch.

packages/core/src/Link/Link.tsx
```diff
@@ -329,6 +343,16 @@ export function Link({
...
     role === 'button' || (role === 'inert' && href == null);
+  const isExternalWithIcon = isExternalLink && !renderAsButton;
+  // The plain anchor stays `inline` so an ancestor clamp (<Text maxLines>)
+  // can truncate it — that is the composition this PR fixes. But a Link that
+  // establishes its own box (`display="block"`, or clamping itself via
+  // `maxLines`, or carrying the external-link icon) must keep the inline-flex
+  // root: an inline anchor around a block-level child computes a focus
+  // outline that paints nothing in Chromium, losing keyboard focus
+  // visibility on those forms.
+  const needsRootBox =
```
cixzhang, line 354: `needsRootBox` misses existing block-level children, so those composed Links still paint no focus ring.

cixzhang (changes requested):
Thanks, the ancestor clamp and both requested focus cases now work in Chromium. One compatibility path remains: existing Links with a block-level child (for example, `HStack` plus an icon) still compute a 2px outline but paint no keyboard focus ring. Please keep those composed Links visibly focused too while leaving text-only ancestor-clamped Links inline.

[Reviewed by Robohands]

packages/core/src/Link/Link.tsx
ManoharPaturi: Addressed in 42ccec6. `needsRootBox` now checks whether `children` is composed with elements (`!isTextOnly(children)`) rather than plain text. Links composed with element children (such as `<HStack>`, icons, or block wrappers) retain the `inline-flex` root so Chromium computes and paints the visible focus outline, while text-only links stay plain `inline` so an ancestor clamp can truncate them. Unit tests added covering both composed element children and text-only children; all 256 Link and Text tests pass.

ManoharPaturi:
Updated branch with commit `42ccec6` to resolve the remaining focus ring case for composed Links:

- `needsRootBox` now checks if `children` contains element children (`!isTextOnly(children)`). Links composed with element children (such as `<HStack>`, icons, or block wrappers) retain the `inline-flex` root box so Chromium paints a visible 2px keyboard focus outline.
- Text-only links continue to render as plain `display: inline` so ancestor `<Text maxLines>` truncation functions as intended.
- Added test coverage in `Link.test.tsx` for composed element children and text-only children. All 256 Link & Text tests pass, ESLint and `pnpm check:repo` pass cleanly.

ManoharPaturi:
@cixzhang Updated in 42ccec6 — Link now inspects its children (`isTextOnly`) so composed elements (such as `HStack` or nested element trees) retain the root focus box for proper Chromium focus ring painting, while plain text/string children remain inline to allow ancestor `<Text maxLines>` clamping. All unit tests, RTL shards, and visual regressions are passing green.

ManoharPaturi:
@cixzhang following up on the round-three update from last week (42ccec6, the isTextOnly composition check). tests and checks are green on it, just checking if you want anything else changed before it moves.

PR #6039, opened 2026-09-04
Honor the last suffix on template demo assets

### LINKED ISSUE #5423: CLI: an unrecognised template-asset extension silently falls through to the image placeholder

Hashim1999164:
Unknown template asset extensions now throw instead of becoming the image placeholder. Video names with extra dots and m4v files are treated as video.

Closes #5423

### DIFF
--- a/.changeset/template-asset-unrecognized-format.md
+++ b/.changeset/template-asset-unrecognized-format.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/cli': patch
+---
+
+[fix] Report the fixture path when a template demo asset has an unsupported format (#6039)
+@Hashim1999164
--- a/packages/cli/foundation/discovery/template-adapter.mjs
+++ b/packages/cli/foundation/discovery/template-adapter.mjs
@@ -166,7 +166,9 @@ const PLACEHOLDER_IMAGE =
  *
  * @type {Set<string>}
  */
-const VIDEO_EXTENSIONS = new Set(['mp4', 'webm', 'mov', 'ogv']);
+const VIDEO_EXTENSIONS = new Set(['mp4', 'webm', 'mov', 'ogv', 'm4v']);
+
+const IMAGE_EXTENSIONS = new Set(['svg', 'png', 'jpg', 'jpeg', 'gif', 'webp', 'avif', 'ico']);
 
 /**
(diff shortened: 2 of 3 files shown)


cixzhang (changes requested):
Template authors with more than one demo asset still cannot tell which reference failed. The [error](https://github.com/facebook/astryx/blob/fe543044dcc88cbe18f85402fb2819653c06c078/packages/cli/foundation/discovery/template-adapter.mjs#L216) reports only `bin`, while the [copy contract](https://github.com/facebook/astryx/blob/5123978efbcd3f824f88a626cc1fea738faee87a/docs/specs/AST-028/spec.md#L102-L107) requires the fixture path and unsupported format. Please include the matched `/template-assets/...` path and assert both values in the focused test.

Because this changes published CLI behavior, please also add a patch `[fix]` Changeset and run Prettier on both changed files. The final-suffix and `m4v` cases otherwise look right.

[Reviewed by Robohands]

Hashim1999164:
yeah the error now has the /template-assets path plus the format. the test checks both. added a [fix] changeset and ran prettier on those two files.

cixzhang (approved):
Thanks, this now reports the exact unsupported fixture, uses the final suffix, and handles m4v correctly. The focused tests, formatting, and patch Changeset are all in place.

[Reviewed by Robohands]

PR #6040, opened 2026-09-04
docs(TreeList): record keyboard cancellation contract

cixzhang:
Repairs the contract-recording gap left by #5606. No runtime behavior changes.

## What

- promote the existing TreeList component record to current
- record the shipped `onKeyDown` cancellation seam as FR5 / ORD1 / DEC-1
- preserve the root `div` currentTarget guarantee
- keep header-slot keys outside built-in tree navigation
- map the existing focused tests to the contract

## Why

#5606’s final exact head was owner-approved after proving cancellation, normal navigation, root-currentTarget compatibility, and header isolation. The merged code and changeset shipped that behavior, but the component record still said no focus or activation rule had changed. This PR faithfully records the approved, already-shipped boundary.

## Validation

- `pnpm check:knowledge`
- Prettier
- `git diff --check`

No Changeset: specification-only synchronization for already-released behavior.

### DIFF
--- a/packages/core/src/TreeList/TreeList.spec.md
+++ b/packages/core/src/TreeList/TreeList.spec.md
@@ -3,13 +3,13 @@ schema_version: 3
 template_version: 3
 kind: component
 id: component:TreeList
-authority: draft
+authority: current
 archive_reason: null
 superseded_by: null
-approved_by: null
-approved_at: null
+approved_by: cixzhang
+approved_at: 2026-09-04
 owners: [cixzhang]
-review_triggers: [behavior, theming, accessibility]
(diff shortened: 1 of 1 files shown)


cixzhang (commented):

Builders still cannot discover the cancellation seam this record makes current: `TreeList.doc.mjs` does not list `onKeyDown`, describe `preventDefault()`, or include an example, while the new Public concepts text says complete prop syntax and examples live there. Please add the inherited handler and cancellation behavior to the English, Chinese, and dense consumer docs, following the existing BaseTypeahead pattern.

[Reviewed by Robohands]

cixzhang:
Closing under the standing no-PR rule.

PR #6041, opened 2026-09-04
Resizable: pin #5934's legacy maxSizePx shrink-after-drag clamp

cixzhang:
## What

Test-only closure for a same-head review finding on #5934 (`table-inbox`).

The reading pane in the table-inbox template derives `maxSizePx` itself — `Math.max(paneFloor, surfaceWidth - listFloor)` from a plain ResizeObserver measurement, not a `containerRef` percentage bound. A dismissed same-head review on #5934 found that after widening the pane past a later surface-derived maximum, the pane stayed oversized and overlapped the list instead of clamping to the new ceiling.

## Investigation

Reproducing the review's exact sequence — drag the pane to 1066px, then narrow the surface so the derived max drops to 834px — at both the `useResizable` hook level and a page.tsx-composition-level replica shows it already clamps correctly on current `main`. `useResizable`'s general FR4 bounds-reclamp (landed in #5783, commit `311deefc`) re-resolves `resolvedMax` from a changed caller-supplied `maxSizePx` on every render and re-clamps the committed selection — independent of whether the bound is a percentage or a plain number. No template or hook behavior change is needed.

## The change

Adds the reproduction directly to `useResizable`'s FR4/FR6 suite (AST-010) so this caller-recomputed-pixel path — distinct from the existing containerRef/percentage-basis tests already there — can't silently regress:

- clamps the selection and the ARIA-visible `_maxSizePx`/`_size` when a legacy numeric `maxSizePx` shrinks below the user's chosen size;
- leaves an in-range user choice untouched (clamping is not a second proportional-resize mode, per FR4).

## Risk

None — test-only, no runtime change.

## Testing

- `vitest run packages/core/src/Resizable/` — 150/150 passing (112 in `useResizable.test.ts`, incl. the 2 new tests)
- `tsc --noEmit`, `eslint` on the changed file — clean
- `pnpm check:repo` — clean

Refs #5934, #5783. AST-010 FR4/FR6.

### DIFF
--- a/packages/core/src/Resizable/useResizable.test.ts
+++ b/packages/core/src/Resizable/useResizable.test.ts
@@ -705,6 +705,51 @@ describe('useResizable percentage configuration (AST-010)', () => {
       // always done.
       expect(result.current.size).toBe(80);
     });
+
+    it('re-clamps a legacy numeric maxSizePx recomputed by the caller, not just a percentage bound', () => {
+      // #5934: a table-inbox reading pane derives `maxSizePx` itself, as
+      // `Math.max(paneFloor, surfaceWidth - listFloor)`, from a plain
+      // ResizeObserver measurement — no `containerRef`/`maxSize` percentage
+      // in play, so this is the caller-recomputed-literal path, not FR1's
+      // basis tracking. #5783 (commit 311deefc) made this path re-resolve
+      // and clamp the selection every render like any other bound, per FR4;
+      // this pins the exact table-inbox pane-widen-then-surface-narrow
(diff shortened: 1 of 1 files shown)


cixzhang (commented):

The regression test still lets the old pane width revive after the surface grows again, so the reading pane can jump without another drag. I limited committed-bound updates to percentage bounds and all 112 tests still passed. Please grow `maxSizePx` back to 1500 after the 834 clamp and assert the size stays 834; that pins the committed selection this PR says it protects.

[Reviewed by Robohands]

PR #6043, opened 2026-09-04

### LINKED ISSUE #6021: Link's inline-flex anchor silently defeats an ancestor Text maxLines clamp

parinith-web:
Fixes #6021

## Summary

Builds on @ManoharPaturi's fix in #6038, which correctly identified and fixed the root cause: Link's plain anchor was `display: inline-flex`, whichestablishes its own formatting context that an ancestor `-webkit-line-clamp` (from `<Text maxLines>`) cannot reach into — so `<Text maxLines={2}><Link>…</Link></Text>` silently failed to truncate.

His fix changes the plain anchor to `display: inline` while preserving `inline-flex` where it's actually load-bearing (external-link
icon + gap, and the button-rendered form), via a dedicated `flexLayout` style.

This PR adds one follow-up fix on top of that: it addresses the Copilotreview comment left on #6038 about the ne`Link.test.tsx` helper,`displayDeclarationsFor`, which matched an element's classList against rawselector text using `String.includes` - a
substring check on hashed atomicclass names (StyleX-generated, e.g. `x1a2b3c`). A shorter class name can bea substring of an
unrelated longer one, so the check could both false-positive and false-negative depending on what else StyleX emitted for that test run.

## Change

Rewrote `displayDeclarationsFor` to walk the real CSSOM instead of rawtext: parses each `CSSStyleRule.selectorText` into class tokens, compares those tokens against the element's actual `classList` by exact match, and reads the declaration via `rule.style.getPropertyValue('display')` rather than a body-text regex. Also descends into grouping rules (e.g. `@media`) so nested declarations aren't missed.

## Testing

- `vitest run packages/core/src/Link packages/core/src/Text` — all passing (Link: 39, useLinkComponent: 11, useLinkify: 13, Text: 32,       plus TextArea/TextInput which depend on Text — 252 total)
- `tsc --noEmit` (core) — clean

### DIFF
--- a/.changeset/link-anchor-clamp.md
+++ b/.changeset/link-anchor-clamp.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Render a plain Link anchor as `inline` so an ancestor `<Text maxLines>` clamp can truncate it; flex layout is kept for the external-link icon and button forms, and Link/Text docs now cover the external-link clamp limitation. (#6021)
+@ManoharPaturi
--- a/packages/core/src/Link/Link.doc.mjs
+++ b/packages/core/src/Link/Link.doc.mjs
@@ -169,10 +169,12 @@ export const docs = {
     bestPractices: [
       { guidance: true, description: 'Write descriptive, concise link text that clearly communicates the destination.' },
       { guidance: true, description: 'Set `isStandalone` when the link appears outside of inline text, so it receives proper base font sizing.' },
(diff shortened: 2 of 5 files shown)


cixzhang (changes requested):
Existing block and self-clamped Links lose keyboard focus visibility in this change. Please preserve their focusable box while keeping only unclamped plain anchors inline.

[Reviewed by Robohands]

ManoharPaturi:
Folded into #6038 (commit 1abf92c, authorship preserved) — thanks again for the cleaner helper! Closing this out; maintainers only need to look at the single PR now.

PR #5988, opened 2026-09-03
fix(rtl): baseline pre-existing applicability gaps

nynexman4464:
## Problem

#5365 correctly stopped treating unexplained all-N/A RTL results as clean, but the registry started empty. The first complete sweep therefore classified 204 pre-existing components as new coverage gaps, and filtered PR runs inherited those failures even when the PR did not introduce them.

## Changes

- Snapshot the 204 current unexplained all-N/A components as explicit coverage debt, distinct from verified N/A.
- Keep known debt visible but non-failing; continue to fail new gaps, stale debt, stale N/A declarations, behavior failures, and malformed registries.
- Make the debt baseline removal-only against the PR or merge-queue base.
- Run every debt removal and verified-N/A change through a separate blocking semantic audit while keeping general `pr-rtl` soft.
- Separate known debt from new findings in PR and weekly summaries.

## Test plan

Built the packages and Storybook, then ran the full unfiltered audit:

```bash
pnpm build
pnpm -F @astryxdesign/storybook build
node apps/storybook/rtl-audit/rtl-audit.mjs \
  --storybook-dir apps/storybook/dist \
  --output /tmp/rtl-full.json \
  --concurrency 4
```

Result: 246 components; 35 measured, 7 verified N/A, 204 known gaps, 0 new gaps, and 0 stale entries. The existing `lab/svgiconregistry` D1 failure remained a real finding, so the run still exited 1.

Reproduced the old false failure by running `core/Button` against an empty debt baseline:

```bash
printf '[]\n' > /tmp/no-known-gaps.json
node apps/storybook/rtl-audit/rtl-audit.mjs \
  --storybook-dir apps/storybook/dist \
  --output /tmp/rtl-button-before.json \
  --filter core/Button \
  --known-coverage-gaps /tmp/no-known-gaps.json
```

Result: exit 1, `core/button` = `coverage-gap`.

Ran the same scope with the checked-in baseline:

```bash
node apps/storybook/rtl-audit/rtl-audit.mjs \
  --storybook-dir apps/storybook/dist \
  --output /tmp/rtl-button-after.json \
  --filter core/Button
```

Result: exit 0, `core/button` = `known-coverage-gap`.

Ran blocking transition controls with an empty current baseline:

```bash
node apps/storybook/rtl-audit/rtl-audit.mjs \
  --storybook-dir apps/storybook/dist \
  --known-coverage-gaps /tmp/no-known-gaps.json \
  --filter core/Button \
  --removed-components core/Button
```

[...]

PR #5885, opened 2026-09-02
test(rtl): close applicability gaps with fail-closed evidence

freddymeta:
## Cause

The RTL applicability contract requires every Core/Lab surface to be `measured`, `verified N-A`, or a coverage gap. The registry originally had almost no reviewed N/A declarations, so the soft-gated audit was permanently red and hid new RTL findings.

The remaining correctness hole was that a verified-N/A row survived later source or story changes. A declaration could therefore keep masking a newly introduced relationship outside the automatic detectors.

### Fail-closed N/A evidence

Every verified-N/A declaration now stores SHA-256 evidence for:

- the canonical component source;
- directly imported helpers/styles in the same top-level component directory;
- owned Storybook story files.

A changed, added, removed, or renamed evidence file returns the component to `coverage-gap` until that one declaration is manually re-reviewed and refreshed. Missing, empty, malformed, duplicate, or undiscoverable declarations also fail closed.

`pnpm rtl:audit:verify-na -- --component core/Text --write` refreshes exactly one reviewed declaration. Bulk refresh is refused.

Storybook's existing required `build` command runs the evidence check after producing `dist/index.json`, so component-, story-, and registry-only PRs cannot bypass it even though `pr-rtl` remains soft-gated.

### Current-main coverage delta

Against current `main`, this PR adds 96 reviewed N/A declarations and four curated targets:

| target | measured |
|---|---|
| Drawer D4 `lab-drawer--showcase` | overlay side |
| Tokenizer D2 `core-tokenizer--logical-order` | token/clear order |
| TransferList D2 `lab-transferlist--default` | selected/available order |
| Typeahead D2 `core-typeahead--logical-order` | token/clear order |

The Stepper D2 target moved to #5495, which landed and now owns the component and regression on `main`.

## Regression coverage

The focused suite proves:

- a current declaration is `verified-na`;
- changing its component source returns it to `coverage-gap`;
- refreshing that evidence restores `verified-na`;
- direct helper/style changes invalidate it;
- owned story edits and additions invalidate it;
- grouped and story-only surfaces map to the right evidence;
- empty evidence, invalid paths/digests, empty reasons, and duplicate IDs fail closed;
- Storybook's required build retains the evidence check.

Real mutation controls also changed `Text.tsx` and `Text.stories.tsx` independently: both failed with the exact changed path, and restoring the file made the checker pass again.

## Current validation

On pushed head `540b8cc694abecd78136500b1a9c5ba3a69594c4`:

- all 101 evidence declarations are current;
[...]

PR #6044, opened 2026-09-04
docs(rtl): propose audit evidence lifecycle

cixzhang:
## Why

Recent RTL audit work split across #5864, #5885, #5988, and #5858 is relying on an evidence policy that has not yet been recorded or owner-approved. This draft gives those changes one architecture target to review instead of letting each PR define a different meaning for measured, verified-N/A, and inherited coverage debt.

## Proposed direction

- compute one live state per component (`measured`, `verified-na`, `coverage-gap`, or stale/known variants)
- pin verified-N/A claims to source/helper/story evidence digests and fail closed on drift
- admit the existing 204-component gap only as a visible, non-failing, removal-only baseline
- fail new gaps, stale declarations, stale baseline rows, and any attempt to grow the baseline
- keep D1 applicability exceptions separate from curated positive targets

This is intentionally `authority: draft`. It proposes policy; it does not authorize the open implementation PRs or change CI/runtime behavior.

## Owner decision still required

`unknown/*` registry entries have no canonical component directory to hash. Before this record can become current, an owner must choose whether those entries are eligible for `verified-na`, and if so which source/story files constitute their evidence.

## Validation

- `pnpm check:knowledge`
- `pnpm check:repo`
- Prettier
- public-repository leak guard
- `git diff --check`

No Changeset: proposed architecture documentation only.

Do not merge until the owner decision is recorded and the exact draft head is approved.

### DIFF
--- a/docs/architecture/rtl-audit-evidence.md
+++ b/docs/architecture/rtl-audit-evidence.md
@@ -1,0 +1,262 @@
+---
+schema_version: 1
+template_version: 1
+kind: architecture
+id: architecture:rtl-audit-evidence
+authority: draft
+archive_reason: null
+superseded_by: null
+approved_by: null
+approved_at: null
+owners: [cixzhang]
+applies_to:
+  [
+    apps/storybook/rtl-audit/,
+    .github/scripts/rtl-audit-coverage.test.mjs,
+    .github/scripts/weekly-rtl-summary.js,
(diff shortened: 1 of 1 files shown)


cixzhang:
Closing this duplicate in favor of [#6046](https://github.com/facebook/astryx/pull/6046), which is now the canonical draft for `architecture:rtl-audit-evidence`. Please continue specification review and revisions there.

PR #6046, opened 2026-09-04
docs: define RTL audit evidence policy

nynexman4464:
## Exact decision needed

What counts as evidence that a component's RTL behavior is correct, and which review states (measured, verified N/A, coverage gap, behavior failure, known debt) may clear a new or changed component.

- Current records: no record under `docs/architecture/` or `docs/design/` owns RTL evidence, audit scope, or verified-N/A admission. `architecture:knowledge-contracts` (evidence freshness, human decisions) and `spec:AST-020` (target-size ownership) were read for boundaries.
- Paths: `apps/storybook/rtl-audit/`, `scripts/component-packages.cjs`, `.github/workflows/ci.yml` (`pr-rtl` shards and join), `.github/workflows/rtl-weekly.yml`, `.github/scripts/weekly-rtl-summary.js`, `internal/eslint-plugin-astryx/no-physical-properties.js`.
- Semantic terms: RTL, bidirectional, mirror, audit, verified N/A, applicability, coverage gap, touch target, D7.
- Overlapping open work (evidence only): #5885 fail-closed verified-N/A evidence and #5988 pre-existing debt baseline both implement one side of OQ1/OQ2 and are blocked on those decisions; #5858 makes fixture applicability order-independent. Merged #5177 is the touch-target false-green that motivates INV10; merged #6252 widened the roster to every registered component package.

## Decision

`architecture:rtl-audit-evidence` records the shipped evidence system: evidence follows the claim (lint, rendered geometry, and interaction prove different things); RTL is graded as an LTR/RTL relationship; behavior failures have no allowlist; all-N/A is not measured; the live roster of every registered component package owns coverage; automatic and curated checks compose; PR scope (per package, joined) and weekly scope answer different questions; applicable evidence invalidates N/A; new dimensions need red/green mutation proof; touch-target evidence must include size, which D7 does not check today.

## Rejected alternatives

- Folding evidence policy into `design:bidirectionality`: audit admission and human RTL intent have different owners and change cadences.
- Treating #5885 or #5988 as the decision: each implements one option of an open owner question; a record must not exist to approve a pull request.

## Claim scope and non-goals

The record stays `authority: draft`. It owns evidence admission and freshness only. It does not decide verified-N/A admission (OQ1), a pre-existing-debt baseline (OQ2), or when `pr-rtl` becomes a required check (OQ3); those are named owner decisions. It does not own shared RTL mechanics (`rtlStyles`, lazy DOM direction reads, centering helpers, hydration); a follow-up `architecture:bidirectional-layout` record is named as the missing owner.

- Verified against the coverage builder, D7 browser check, verified-N/A registry, CI shard/join, and weekly summary on current `main`.
- Reproduced the #5177 false-green: center alignment stays measured when the native 24×24 expansion is removed because D7 does not assert size.
- The CI summary still carries an unused `knownNotRtl` path; the record lists its removal as an enforcement gap.

## Scope

- [x] This PR contains specification records only.

## Validation

- `pnpm check:knowledge`
- Prettier
- `git diff --check`

### DIFF
--- a/docs/architecture/rtl-audit-evidence.md
+++ b/docs/architecture/rtl-audit-evidence.md
@@ -1,0 +1,224 @@
+---
+schema_version: 1
+template_version: 1
+kind: architecture
+id: architecture:rtl-audit-evidence
+authority: draft
+archive_reason: null
+superseded_by: null
+approved_by: null
+approved_at: null
+owners: [cixzhang, imdreamrunner, nynexman4464]
+applies_to:
+  [
+    apps/storybook/rtl-audit/,
+    .github/workflows/ci.yml,
+    .github/workflows/rtl-weekly.yml,
+    .github/scripts/rtl-audit-coverage.test.mjs,
+    .github/scripts/weekly-rtl-summary.js,
+    internal/eslint-plugin-astryx/no-physical-properties.js,
+  ]
+verified_by:
+  [
+    .github/scripts/rtl-audit-coverage.test.mjs,
+    .github/scripts/weekly-rtl-summary.test.mjs,
+    internal/eslint-plugin-astryx/no-physical-properties.test.mjs,
+  ]
+deciding_specs: []
+---
+
+# RTL audit evidence architecture
+
+## Purpose
+
+Astryx needs review evidence that distinguishes three claims:
+
+1. a component's direction-sensitive behavior was actually measured;
+2. RTL genuinely does not apply to the component; or
+3. RTL applicability remains unmeasured debt.
+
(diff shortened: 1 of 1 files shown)


cixzhang (approved):
Thanks for separating measured RTL evidence, verified non-applicability, historical coverage debt, and merge-gate policy instead of treating one implementation as already decided. Approved as the canonical RTL audit-evidence record at exact head `a1a63a9238fe5d382cb49b3b3c493b6d48e83ed3`.

PR #6047, opened 2026-09-04
docs: define bidirectionality design intent

nynexman4464:
## Exact decision needed

What correct bidirectional (LTR/RTL) behavior means for people using Astryx: what mirrors, what stays physical, and how reading order, interaction, overlays, mixed-direction text, and hit targets behave.

- Current records: no `docs/design/` record covers direction or RTL; no component record links a bidirectionality requirement. `spec:AST-020` owns target size and its WCAG 2.5.8 exceptions; `architecture:rtl-audit-evidence` (#6046) owns evidence; `architecture:internationalization` (#6048) owns provider direction and the DOM-direction boundary.
- Paths: `apps/storybook/rtl-audit/`, `packages/core/src/utils/rtlStyles.ts`, `packages/core/src/i18n/useDirection.ts`, `internal/eslint-plugin-astryx/no-physical-properties.js`.
- Semantic terms: RTL, bidirectional, mirror, logical, inline start/end, direction, touch target.
- Overlapping open work (evidence only): #5885 and #5988 (audit evidence, owned by #6046); the merged #5177 touch-target centering fix and its review, which is the source of DR8/DR12.

## Decision

`design:bidirectionality` records DR1–DR14: preserve meaning over coordinates; content direction and layout direction are distinct; reading-order relationships follow inline start/end; directional artwork is treated exactly once and Unicode-mirrored glyphs are not mirrored twice; direction-neutral and physical content stays stable; interaction, drag, scroll, and overlays follow the visual relationship; hit targets stay centered and keep the owning component's minimum size; first paint agrees with the region direction; exceptions require semantics.

## Rejected alternatives

- One combined i18n/RTL spec: runtime invariants, human design intent, and audit policy have different owners.
- Writing mechanics (logical CSS, `rtlStyles.mirror`, lazy DOM reads) into the design record: those belong to an architecture owner.

## Claim scope and non-goals

The record stays `authority: draft` until a design owner approves it; its `architecture` references must be current first or together with it. It does not define provider APIs, CSS or hook implementation, lint rules, audit scores, verified-N/A admission, or consumer examples. Component records adopt the specific DRs they implement.

Preserves the direction foundation shipped in #4269, the logical-layout and mirroring migrations, and the rendered evidence used to approve #5177. Valid without any implementation PR.

## Scope

- [x] This PR contains specification records only.

## Validation

- `pnpm check:knowledge`
- Prettier
- `git diff --check`

### DIFF
--- a/docs/design/bidirectionality.md
+++ b/docs/design/bidirectionality.md
@@ -1,0 +1,189 @@
+---
+schema_version: 1
+template_version: 1
+kind: design
+id: design:bidirectionality
+authority: draft
+archive_reason: null
+superseded_by: null
+approved_by: null
+approved_at: null
+owners: [cixzhang, rubyycheung, nynexman4464]
+review_triggers: [visual, interaction, accessibility]
+verified_by:
+  [
+    apps/storybook/rtl-audit/rtl-audit.mjs,
+    apps/storybook/rtl-audit/targets.json,
+    internal/eslint-plugin-astryx/no-physical-properties.test.mjs,
+  ]
+architecture:
+  [architecture:internationalization, architecture:rtl-audit-evidence]
+components: []
+families: []
+deciding_specs: [spec:AST-020]
+---
+
+# Bidirectionality design specification
+
+## User intent
+
+People should understand and operate the same Astryx interface whether their
+page, region, or content flows left-to-right or right-to-left. RTL support is not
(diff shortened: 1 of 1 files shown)


cixzhang (approved):
Thanks for consolidating the shared bidirectionality design intent. Approved at exact head `e406a8402eb4b94e719b3ccea91bb4038c23f3c2`.

PR #6048, opened 2026-09-04
docs(core): define the internationalization contract

nynexman4464:
## Exact decision needed

Record the released Astryx internationalization contract as current authority: which runtime, fallback order, string ownership, locale, direction, and calendar boundaries are intentional, so reviewers can judge i18n and locale-sensitive changes without reconstructing the history in RFC #3641.

- Current records: none under `docs/architecture/` or `docs/design/` covers i18n, locale, or direction; `packages/core/src/i18n/` had no `.spec.md`. `architecture:knowledge-contracts`, `spec:AST-002` (public API surface), and `spec:AST-020` (target size) were read for boundaries.
- Paths and exports: `packages/core/src/i18n/`, `packages/core/locales/`, `utils/plainDate.ts`, `utils/dateParser.ts`, `scripts/check-i18n-catalog.mjs`, the i18n ESLint rules, and the `@astryxdesign/core/i18n` barrel (`InternationalizationProvider`, `InternationalizationContext`, `useTranslator`, `useLocale`, `useCollator`, `useDirection`, `getLocaleDirection`, `Translator`).
- Semantic terms: internationalization, i18n, locale, translation, catalog, fallback, ICU, direction, RTL, `PlainDate`, Gregorian.
- Overlapping open work (evidence only, not authority): #4404 translator adapter (INV14 / OQ3); #6217 and #6242 Spinner label localization and #6125 Switch announcement (conformance to INV1); #6220 ChatTypingIndicator `Intl.ListFormat` (INV8); #6205 Crowdin catalog update (INV6). Closed #5684 is the rejected parallel-runtime proposal recorded in DEC-1.

## Decision

- `architecture:internationalization` — cross-library contract: Astryx owns Astryx strings; the released provider and hooks are canonical; no provider means deterministic English; runtime locale changes are live; ordered, silent fallback for expected gaps; `en.json` is the source contract; messages are ICU MessageFormat 1 strings; the provider locale owns locale-sensitive behavior; pure helpers default deterministically to English; `PlainDate` stays Gregorian; rendered direction comes from the DOM; locale and direction are distinct; client context is the current runtime; external-runtime integration is additive.
- `component:InternationalizationProvider` — provider-local behavior: resolution order, missing-key behavior, live updates on any prop change, no locale persistence, no DOM `lang`/`dir` mutation, no-provider defaults, string-returning translator, and nested providers replacing rather than merging parent context.
- Both records carry `authority: current` with the owner approval given at the previous head (`/approve-spec 8b4f2b9853`, 2026-09-05) recorded in metadata. Any new head needs the owner's approval again.

## Rejected alternatives

- One monolithic i18n/RTL spec: mixes shipped runtime invariants, human RTL intent, and audit policy, which have different owners and change cadences.
- Copying RFC #3641 into a record: the RFC proposed a plain `t()`, RSC dispatch, and warn-on-missing-translation that were deliberately not shipped; the records capture the released result and mark those as separate decisions.
- Recording the parallel `IntlProvider`/`useIntl` runtime from #5684 (DEC-1).

## Claim scope and non-goals

The records govern Astryx-owned strings and locale-sensitive operations. They do not decide a server/RSC translation runtime, an external-runtime adapter contract, malformed-locale rejection, or alternative calendar systems; each stays listed as an open question. Component contracts continue to own which content is Astryx-authored versus caller-authored.

- Released implementation in `packages/core/src/i18n/`, `check:i18n-catalog`, and the i18n lint rules is the evidence base; the records remain valid without any implementation PR.
- New focused test `InternationalizationProvider.test.tsx` covers re-rendering on `messages`, `overrides`, and `dir` changes and nested-provider replacement, closing the verification-map gap where the prior head named tests that only covered `locale`. Mutation proof: memoizing on `locale` alone fails the three re-render tests; merging parent `overrides`/`direction` into a nested provider fails the nesting test.
- Verified behavior worth an owner decision (recorded as OQ1 on the component record): shipped English is the only cross-locale fallback and contains Astryx keys only, so an application key missing from the active locale renders as its key even when a supplied `en` catalog exists; the consumer guide currently implies otherwise.

## Scope

- [ ] This PR contains specification records only. (No: it also names the provider's nonvisual anatomy in `InternationalizationProvider.doc.mjs` so the component record can map anatomy without implying rendered DOM, adds the docs changeset for that, and adds one focused test.)

## Validation

- `pnpm check:knowledge`
- `node scripts/check-changesets.mjs`
- `pnpm exec vitest run --project node scripts/check-knowledge.test.mjs`
- `pnpm exec vitest run --project ui packages/core/src/i18n/`
- Prettier and `git diff --check`


cixzhang (approved):
Thanks for documenting the shared internationalization architecture and the provider-specific contract together. Approved at exact head `8b4f2b9853fae7dafb3bd4d72877301bfa489366`.

PR #6050, opened 2026-09-04
test(Resizable): keep legacy bound clamps permanent

cixzhang:
## Why

#6041 merged before its same-head review request was incorporated. Without the final grow step, the regression could still pass if a legacy numeric `maxSizePx` clamp were only temporary and the earlier pane width revived when the surface widened again.

## What

Extend the existing table-inbox regression through the complete sequence: drag to 1066px, clamp to 834px, then grow the maximum back to 1500px. The test asserts that the committed selection and handle geometry stay at 834px while the advertised maximum updates.

## Risk

None — test-only, with no runtime or public API change.

## Testing

- `vitest run packages/core/src/Resizable/useResizable.test.ts` — 112/112 passing
- ESLint on the changed file — clean

Follow-up to #6041.

### DIFF
--- a/packages/core/src/Resizable/useResizable.test.ts
+++ b/packages/core/src/Resizable/useResizable.test.ts
@@ -706,7 +706,7 @@ describe('useResizable percentage configuration (AST-010)', () => {
       expect(result.current.size).toBe(80);
     });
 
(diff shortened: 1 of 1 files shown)


cixzhang (commented):
Semantic verdict: approve

This prevents a pane from reviving an old width after its surface shrinks and grows. The new assertion kills the exact missed regression.

[Reviewed by Robohands]

PR #6055, opened 2026-09-04
spec: define Button family behaviors

cixzhang:
## Why

People should encounter one coherent Button family whether an action is labeled, icon-only, persistent, standalone, or grouped. Today those behaviors are spread across component implementations, and ToggleButton exposes the resulting contract gap tracked in #6012.

This draft defines the family before fixing any one member.

## Family contract

- Defines membership for Button, IconButton, ToggleButton, ButtonGroup, and ToggleButtonGroup.
- Defines shared naming, activation, disabled, pending Action, sizing, elevation, theming, and group-ownership behavior.
- Keeps momentary actions, persistent pressed state, link mode, connected groups, and spaced selection groups as deliberate variations.
- Records ToggleButton elevation as one adoption gap against the family contract rather than the subject of the spec.

## Risk

Specification only. No runtime, package, theme, or public API behavior changes in this PR.

## Testing

- `pnpm check:knowledge`
- `pnpm exec prettier --check docs/families/buttons.md`
- `git diff --check`

Tracks https://github.com/facebook/astryx/issues/6012

### DIFF
--- a/docs/families/buttons.md
+++ b/docs/families/buttons.md
@@ -41,10 +41,10 @@ geometry, accessible-name requirement, interaction feedback, asynchronous-action
 model, and surface ownership. Each member still owns the semantics specific to a
 momentary action, navigation destination, persistent pressed state, or group.
 
-This family contract also resolves the gap in
-[issue #6012](https://github.com/facebook/astryx/issues/6012): ToggleButton
-renders Button's elevation state but cannot currently express the family's
-resting elevation axis.
+ToggleButton renders Button's elevation state but cannot currently express the
+family's resting elevation axis. Whether that axis belongs on ToggleButton
+remains unresolved because ToggleButton currently uses a transparent ghost
+surface.
 
 ## Membership rule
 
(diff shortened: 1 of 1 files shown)


docs/families/buttons.md
```diff
@@ -0,0 +1,233 @@
...
+People should encounter one coherent button system whether an action is labeled,
+icon-only, persistent, standalone, or grouped. Members share the same control
+geometry, accessible-name requirement, interaction feedback, asynchronous-action
+model, and surface ownership. Each member still owns the semantics specific to a
+momentary action, navigation destination, persistent pressed state, or group.
+
+This family contract also resolves the gap in
+[issue #6012](https://github.com/facebook/astryx/issues/6012): ToggleButton
+renders Button's elevation state but cannot currently express the family's
+resting elevation axis.
```
cixzhang: Could we remove the issue reference and state the durable gap directly? For example: `ToggleButton renders Button’s elevation state but cannot currently express the family’s resting elevation axis.` Long-lived specs should not depend on PR or issue context to explain the contract; those links belong in PR descriptions, migration notes, or historical provenance when needed.

cixzhang (commented):
Thanks for consolidating these shared Button-family rules. I left one writing note to keep the long-lived contract independent of transient work tracking.

docs/families/buttons.md
```diff
@@ -0,0 +1,233 @@
...
+| `component:IconButton`        | inherits Button props and always selects icon-only presentation                        | tooltip remains explicit consumer guidance                                                              |
+| `component:ToggleButton`      | inherits Button geometry, focus, pending presentation, and rendered elevation state    | public `elevation` prop and matching theme visual metadata are missing under FR8–FR10; tracked by #6012 |
+| `component:ButtonGroup`       | connected surface, inherited size/disabled state, shared elevation, and roving focus   | member elevation is intentionally suppressed while connected                                            |
+| `component:ToggleButtonGroup` | controlled single/multiple selection, inherited size/disabled state, and spaced layout | wrapper is not a painted surface and therefore exposes no family elevation                              |
+
+The ToggleButton row is an implementation gap, not an approved family
+exception. Adding the existing optional `Elevation` axis restores family parity
+without changing no-prop rendering.
```
cixzhang: Thanks for identifying the parity gap. I don’t think adding `elevation` is a mechanical restoration yet: ToggleButton always renders through Button’s transparent `ghost` variant, and its pressed fill is an overlay rather than a stable opaque surface. A shadow under that treatment may not read as one elevated control and can expose the content beneath it. Could this remain an unresolved design gap instead of an approved implementation gap until we define the elevated ToggleButton surface? The follow-up should decide whether elevated ToggleButtons gain an opaque floating treatment or whether elevation is unsupported for this member.

cixzhang (commented):
Thanks for consolidating the Button-family behavior. I left one inline design blocker on treating ToggleButton elevation as mechanical parity.

PR #6057, opened 2026-09-04
fix(Tokenizer): avoid blank trailing input row

imdreamrunner:
## Summary

- remove Tokenizer's empty trailing combobox from flex layout so it cannot create a blank final row
- bound the empty search hit surface before the measured end-control lane, including the Clear all button's 2px coarse-pointer hit outset
- keep pills and controls above the background input, while restoring the normal in-flow editing width as soon as a query is entered
- add deterministic real-Chromium guards for LTR and RTL layout, hit testing, focus, and coarse-pointer behavior

## Why

In a nearly full PowerSearch row, the mounted combobox still participated in flex layout. Its minimum editing width could wrap by itself and make the field 55px tall even though no content appeared on the second row.

The empty input is now an absolutely positioned first-row background hit surface. It does not consume flex space, so it cannot create a row, and the pills and controls remain the foreground hit targets. Its inline end stops 3px before the measured end lane: 2px for InputClearButton's invisible 24px coarse-pointer target plus 1px separation. Typing restores the ordinary in-flow input.

## Apples-to-apples visual verification

Both images below were captured in the same Linux Chromium run with identical data, neutral/light theme, 600 CSS-pixel fixture, 629 × 100 viewport, 2× device scale, and settled fonts. The before capture restores only the PR base's input-layout rules; the after capture uses this branch.

Before — the tokens fit row one, but the empty combobox creates a blank second row (55px field):

Before: empty combobox creates a blank second row

After — the identical fixture stays one row (32px field):

After: identical PowerSearch fixture stays on one row

The earlier screenshots were removed because they mixed the original deployed story with the PR-preview story and therefore had different token styling; they were not a valid visual comparison.

## Verification

- `pnpm vitest run packages/core/src/Tokenizer/Tokenizer.test.tsx packages/core/src/PowerSearch/PowerSearch.test.tsx` — 106 passed
- [Coarse-pointer reproduction on the unfixed commit](https://github.com/facebook/astryx/actions/runs/33941690649/job/101240905930) — fails with exactly 1.00px overlap in both LTR and RTL
- [Coarse-pointer verification on the final clean head](https://github.com/facebook/astryx/actions/runs/33944223219/job/101247861651) — passes both LTR and RTL

The browser guard requires all pills to remain on row one, the empty combobox to share that row, a positive exposed input hit target, no overlap with the full Clear all touch target, correct token/Clear all hit testing, and focus after clicking the input.

[Open the LTR regression story](https://facebook.github.io/astryx/pr/6057/iframe.html?id=core-powersearch--near-full-token-row&viewMode=story) or [the RTL variant](https://facebook.github.io/astryx/pr/6057/iframe.html?id=core-powersearch--near-full-token-row-rtl&viewMode=story).

### DIFF
--- a/.changeset/quiet-tokens-share.md
+++ b/.changeset/quiet-tokens-share.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Prevent an empty Tokenizer input from creating a blank trailing row.
+@imdreamrunner
--- a/.github/scripts/story-play-guard.js
+++ b/.github/scripts/story-play-guard.js
@@ -167,13 +167,22 @@ async function capturePowerSearchEvidence(browser) {
       ['AFTER', ''],
     ]) {
       const page = await context.newPage();
+      await page.addInitScript(recordStoryOutcome);
       await page.goto(
         `http://localhost:${port}/iframe.html?id=core-powersearch--near-full-token-row&viewMode=story${query}`,
         {waitUntil: 'networkidle', timeout: 30000},
       );
       await page.locator('.astryx-tokenizer').waitFor();
+      await page.waitForFunction(
+        () => window.__storyOutcome && window.__storyOutcome.done === true,
+        null,
+        { timeout: 30000 }
+      );
       await page.evaluate(async () => {
         await document.fonts.ready;
+        if (document.activeElement instanceof HTMLElement) {
+          document.activeElement.blur();
+        }
         await new Promise(resolve =>
           requestAnimationFrame(() => requestAnimationFrame(resolve)),
         );
--- a/.github/scripts/story-play-guard.js
+++ b/.github/scripts/story-play-guard.js
@@ -41,13 +41,16 @@ const TARGETS = [
   {
     component: 'PowerSearch',
     story: 'core-powersearch--near-full-token-row',
+    hasTouch: true,
     guards:
-      'an empty trailing combobox stays on the nearly full token row without overlapping Clear all',
+      'on a coarse pointer, an empty trailing combobox stays on the nearly full token row without overlapping the full Clear all hit area',
   },
   {
     component: 'PowerSearch',
     story: 'core-powersearch--near-full-token-row-rtl',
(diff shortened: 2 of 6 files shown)


cixzhang (changes requested):
Thanks, the row now stays compact. One blocker: the empty search target sits under **Clear all**, so clicking it removes filters instead of focusing search in LTR and RTL. Please preserve a non-overlapping input hit target while keeping the compact row.

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks, this fixes the Clear all overlap, but the exact-head story now fails its original compact-row assertion in both LTR and RTL: input top 47px, final token bottom 44px. The field is 55px tall again, so users still see the blank second row this PR is meant to remove. Please keep a positive, non-overlapping input hit target without restoring the extra row.

[Reviewed by Robohands]

cixzhang (changes requested):
People clicking the empty search area can still remove a filter instead of focusing search. This exact head restores the compact row, but real Chromium renders the 4px input overlapping **Clear all** by 3.7px in both LTR and RTL; its center hit-tests as **Clear all**. The added story throws that same error locally while CI passes, so the regression fixture is not portable enough to gate this. Please keep the compact row while preserving a non-overlapping search target in both directions, and make the browser guard deterministic across supported environments.

[Reviewed by Robohands]

cixzhang (changes requested):
People tapping the exposed search area can still remove a filter instead of focusing search. In coarse-pointer Chromium, Clear all’s 24px hit area extends 2px beyond its visible box while the input leaves only a 1px gap; 1–1.5px inside the input resolves to Clear all in both LTR and RTL, and clicking there removes a filter. The new browser stories pass because they run with a fine pointer and compare only the visible button rectangle. Please reserve the full effective Clear all hit area and run this guard with `pointer: coarse` in both directions.

[Reviewed by Robohands]

cixzhang (approved):
Thanks, this now keeps the row compact without letting the search or Clear all targets steal each other’s clicks in LTR/RTL and fine/coarse modes.

[Reviewed by Robohands]

PR #6058, opened 2026-09-04
feat(Stepper): configure horizontal collapse

imdreamrunner:
## Summary

- group horizontal-only behavior under horizontalOptions
- accept minimumStepWidth as pixel numbers or browser-resolved CSS lengths
- expose withLabelAndControls, withLabel, and hiddenLabel collapsed variants
- keep the existing 112px and label-plus-controls behavior when options are omitted

Stacked on #5659.

## Test plan

- Stepper and prop-documentation tests: 108 passed
- TypeScript no-emit check
- focused ESLint
- knowledge, changeset, and sync checks


cixzhang (changes requested):
Thanks for consolidating this. The compact API still has two caller-controlled axes and the wrong third presentation. At 320px, `hiddenLabel` leaves two focusable controls with no visible label, while `minimumStepWidth="4rem"` keeps the truncated step labels expanded. Please keep the fixed 112px automatic collapse and expose one horizontal-only choice with exactly: label + controls, label only, or no summary row. The default and label-only states, focus retention, vertical DOM, and zero track-to-summary gap otherwise checked out in Chromium.

[Reviewed by Robohands]

imdreamrunner:
Thanks — Cindy and I aligned that `minimumStepWidth` and `collapsedVariant` remain as the two fields in `horizontalOptions`, so we are keeping the configurable collapse threshold. We agree with the second finding: `hiddenLabel` is the wrong third presentation because it leaves controls without a visible current-step label. We’ll replace it with a no-summary option: compact track only, with no label or controls row.

PR #6061, opened 2026-09-04
Validate theme visual props from generated Core contract

imdreamrunner:
## Summary

- generate and commit a versioned Core visual-prop contract from checked TypeScript types plus public component theming docs
- consume the same contract for root, media-surface, and adaptation validation; generated module augmentation; target discovery; and visual-probe generation
- keep TypeScript entirely in repository-time generation so the shipped CLI has no TypeScript runtime dependency
- preserve deterministic legacy validation when an installed Core package has no supported contract, while failing early on malformed supported contracts
- distinguish finite, open, and unresolved domains; keep states separate from visual props; and preserve the established `CustomTextTypes` extension point
- escape generated class selectors for fractional numeric values and probe deprecated compatibility aliases independently from their canonical targets

The generated snapshot currently covers 276 targets and 304 selectors: 225 finite visual-prop domains, 9 open domains, 0 unresolved visual-prop domains, and 20 public augmentation interfaces.

Architecture and compatibility semantics are documented separately in draft #6091 (AST-031).

## Test plan

- `pnpm check:repo`
- `pnpm build`
- full CLI suite: 224 files / 3,287 tests
- focused generator, loader, Core selector, packaging, probe, root-theme, and adaptation suite: 91 tests
- ESLint on changed CLI runtime and test files
- independent implementation review: no remaining actionable findings


cixzhang:
This draft is the release-clearing child of #5543. The remaining blocker is fail-closed validation for alias-backed finite visual props across both root themes and adaptations. Before the breaking API can ship, please resolve the current correctness edge cases, rebase the #5543 → #6061 stack onto current main, mark this PR ready, and obtain owner review plus exact-head `pr-a11y`, `Stable visual regression`, and full CI. If the shared validator cannot meet that bar today, the safe release action is to exclude the entire adaptations stack rather than ship #5543 alone.

PR #6063, opened 2026-09-05
docs(theme): gate extensible axes on safe fallbacks

cixzhang:
## User impact

Theme authors get a clear admission rule for custom prop values: an axis can open only when a missing theme-specific value still produces a safe, deterministic result. This prevents a theme mismatch from silently changing component geometry, alignment, composition, behavior, or state.

## Contract

- allow theme-defined values only for visual axes with a safe, theme-independent fallback
- keep behavioral, structural, placement, directional, and state-machine vocabularies closed while still allowing themes to restyle existing values
- record `Heading.type` as the qualifying example because required `Heading.level` supplies its baseline, and `Icon.size` as the non-qualifying example because no missing size can be inferred safely
- keep `extensibleAxes.test.ts` scoped to structural map/reflection/docs consistency; semantic admission remains component-contract, owner-review, and focused-fallback evidence

## Verification

- `pnpm exec vitest run packages/core/src/theme/extensibleAxes.test.ts` — 38 passed
- `pnpm check:repo`
- changed-file Prettier and relative-link checks
- public-context hygiene scan
- `git diff --check origin/main...HEAD`
- independent read-only review: PASS at exact head `3edc6cd08a03b506a06b709b07dc4731017a4ae3`

## Runtime

No runtime or published package API changes.

### DIFF
--- a/docs/architecture/component-theming-surface.md
+++ b/docs/architecture/component-theming-surface.md
@@ -19,6 +19,7 @@ applies_to:
     packages/core/src/theme/themingTargets.test.ts,
     packages/core/src/theme/extensibleAxes.test.ts,
     packages/core/src/theme/derivedVarRegistry.test.ts,
+    docs/contributing/api-conventions.md,
     docs/templates/knowledge/component-spec.md,
     scripts/check-knowledge.mjs,
   ]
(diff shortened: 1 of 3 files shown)


cixzhang (commented):

Theme authors can still lose Heading’s level baseline when a custom `type` has no matching rule. The new current contract says `Heading.type` qualifies because `level` supplies that fallback (`docs/architecture/component-theming-surface.md:172`), but `Heading.tsx` chooses the type branch whenever `type` is present; an unknown value supplies no level styles. A focused no-match probe changed level 2 from Astryx heading tokens to the browser’s `h2` defaults (`1.5em` and `bold`).

Could we describe Heading as eligible only once that fallback exists, or keep it out of the qualifying example until the implementation lands?

[Reviewed by Robohands]

PR #6064, opened 2026-09-05
feat(templates): rework the grouped table around one shared range

ernestt:
Reworks `table-grouped` around one shared time range, per-table sorting, and rows that expand in place into a full-width history chart.

> [!IMPORTANT]
> **Blocked on two core PRs.** This template uses `hasRowClickExpansion` and `panelVariant` from #5995, and `chevronPlacement` from #5993. Neither prop exists on `main`, so this will not build until both land. Both are green and mergeable, awaiting review. Reviewing the template in the meantime is welcome — just don't merge it first.

## What changed

The template previously showed four static tables. It now behaves like one page rather than four:

- **One time range drives everything.** The `1W / 1M / 3M / YTD / 1Y` control sits in the page header, and every table's Change column and every expanded chart reads from it. Opens on YTD.
- **Each table sorts on its own columns.** Sorting is per-group, because the groups do not share a schema — that is the whole reason this template is not one grouped table.
- **Rows expand into a full-width chart** rather than a sparkline beside a metadata list. Charts within a category share a `syncId`, so hovering one moves the cursor across all of them.
- **Below 720px each table folds to two columns** — the name, and the one figure that headlines the group with its move underneath.

overview

## Row detail

Expanding a row reveals the account's history over the selected window. There is no `CartesianGrid`: a rule at every Y tick reads as a table ruled behind a table. The one rule worth drawing is dashed, at the value the window opened on, so everything above it is the gain the Change column reports and everything below is the loss.

reference line under a rising and a falling account

expanded row

## Compact layout

Below 720px the widest table — credit cards, with three money columns and a change — has less room than its fixed widths ask for. Every table switches on the same threshold so the page stays four cards rather than four differently-shaped ones.

The arrow states the sign that the colour also states, so the cell survives where the colour does not. The liability inversion holds: a card balance that grew points **up in red**.

compact two-column layout

This measures the box, not the window. The docsite preview is a dialog and the catalog draws these cards through `scale(0.5)`, so `matchMedia` would describe a viewport the cards never get. First read is `offsetWidth` inside `useLayoutEffect`, in the same units the `ResizeObserver` later reports — a `getBoundingClientRect` first read bakes in the ancestor transform and opens the catalog thumbnails at half the real width.

## Expansion performance

Recharts costs roughly 400ms of main thread to mount, and spending it while the row is still growing starved the open. A skeleton of the chart's exact height stands in until the row has finished opening, so the row animates over a cheap box and the chart mounts into a panel already at rest.

| | before | after |
|---|---|---|
| Frames across a warm expand | 9–12 | **20–22** |
| Frames across the first expand | 2 | **21** |
| Something on screen | ~410ms | **~130ms** |
| Chart lands, first expand | 2583ms | **1199ms** |

Cost is ~165ms of extra wait on a warm expand. The delay is read off whichever ancestor carries the `grid-template-rows` transition rather than hardcoded, so reduced motion (which zeroes the token and skips the skeleton) and theme overrides both hold.

skeleton standing in while the row opens

*Measured on the dev server, where React and Recharts are unminified and StrictMode double-mounts. Production will be faster; the ordering is what matters here.*

## Expand all, and other ranges

"Expand all" opens the cards **and** the rows inside them — a control that leaves thirteen rows shut has not expanded all, and its next click would read as "Collapse all" while most of the page is still closed.

expand all

1W range

## Rubric score: 85/100 (B)

Rubric 1.3. Unchanged from the ledger's existing B 85 for this slug, but the composition moved.

| Category | Score | Notes |
|---|---|---|
| Astryx component purity | **30**/30 | 0 raw HTML elements |
| Icon purity | **15**/15 | 0 raw SVG |
| Custom CSS | **0**/15 | 46 properties — see below |
| Layout & structure | **15**/15 | `Layout` root, single page, no `AppShell` |
| Doc metadata | **10**/10 | 6 fields + 3 description + 1 naming |
| Image handling | **5**/5 | No images |
| Code quality | **10**/10 | `'use client'`, default export, realistic data, no dead code |

**On the 0/15 for custom CSS.** 46 properties, which lands in the 21+ band. Roughly 29 of them are the category tile's icon-to-chevron hover cross-fade, which has no Astryx prop equivalent — there is no component that swaps a glyph on ancestor hover. I want to flag a rubric gap rather than quietly eat the zero: the **justified** qualifier only exists in the 1–3 band, so 46 necessary properties score identically to 46 gratuitous ones. I did not reduce the count, because the reachable saving was about four properties and would not have changed the band.

**On `recharts`.** Category 7 asks for self-contained imports and names no exception, but eight templates already import it and `page/dashboard` sits at A 95 in the ledger with it, so I scored it as established precedent rather than a deduction.

## Naming finding (not fixed here)

`groupedRows` is a real plugin in `packages/core/src/Table/plugins/`, and this template exists to show the case where you do **not** use it — its own description ends with "where one grouped table would force empty columns." The slug and the `Table - Grouped` category therefore assert the concept the template argues against, and a genuine `groupedRows` demo has nowhere to live.

The category is the part that costs something measurable: it is word-split into the keyword pool, so it pulls "grouped rows" queries toward a template that will disappoint them, whereas the rubric's own ablation found the name axis decided 0 of 36 queries. `table-sections` with `Table - Sectioned` would be my suggestion. Left alone here because renaming is a breaking change to the `astryx template <slug>` path and deserves its own PR.

## Test plan

- `tsc --noEmit` clean, `eslint` clean, `prettier` clean.
- Verified in the docsite preview at 1500px and 760px: rendered column headers confirm all four tables carry their full sets wide and exactly two narrow.
- Expansion, sorting, range switching and expand-all exercised via Playwright against the running docsite; no page errors.
- Chart internals asserted directly: 0 `CartesianGrid` lines, 1 reference line, dash `3 3`, and the first x-axis tick present after `interval="preserveStartEnd"`.

### DIFF
--- a/.changeset/table-grouped-shared-range.md
+++ b/.changeset/table-grouped-shared-range.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/cli': patch
+---
+
+[feat] The grouped table template now runs every account group off one time range, with full-width charts in the row detail and a position chart whose stats track the pointer.
+@ernestt
--- a/.changeset/table-grouped-shared-range.md
+++ b/.changeset/table-grouped-shared-range.md
@@ -2,5 +2,5 @@
 '@astryxdesign/cli': patch
 ---
 
-[feat] The grouped table template now runs every account group off one time range, with full-width charts in the row detail and a position chart whose stats track the pointer.
+[feat] The grouped table template now sorts each account group on its own columns, runs all of them off one time range in the page header, and expands rows in place into a full-width history chart. Below 720px each table folds to its name and headline figure, with the move underneath as an arrow and a percentage.
 @ernestt
--- a/packages/cli/assets/templates/pages/table-grouped/page.tsx
+++ b/packages/cli/assets/templates/pages/table-grouped/page.tsx
@@ -1009,7 +1009,7 @@ function DetailChart({
   const fillId = `detail-fill-${id}`;
 
   return (
-    <div ref={hostRef}>
+    <VStack gap={0} ref={hostRef}>
       {!isSettled ? (
         <Skeleton width="100%" height={CHART_HEIGHT} radius={2} />
       ) : (
@@ -1057,7 +1057,7 @@ function DetailChart({
           </AreaChart>
         </ResponsiveContainer>
       )}
-    </div>
+    </VStack>
   );
 }
 
--- a/packages/cli/assets/templates/pages/table-grouped/page.tsx
+++ b/packages/cli/assets/templates/pages/table-grouped/page.tsx
@@ -78,6 +78,7 @@ import * as stylex from '@stylexjs/stylex';
 import {
   Area,
   AreaChart,
+  ReferenceLine,
   ResponsiveContainer,
   Tooltip,
   XAxis,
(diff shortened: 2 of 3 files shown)


ernestt:
Closing in favour of shipping this as a new template rather than a replacement.

`table-grouped` stays as it is. This work becomes `table-collapsible` — a separate template for the shape where each group carries its own table and its own columns inside a collapsible card. Reopening as a new PR; the two templates then sit side by side instead of one overwriting the other.

PR #6065, opened 2026-09-05
feat(templates): add the collapsible table page template

ernestt:
Adds `table-collapsible`, a new page template. **Purely additive — `table-grouped` is untouched.**

## Why a second table-with-sections template

The two are the same silhouette solving opposite problems, and the split is the point:

|  | `table-grouped` (existing) | `table-collapsible` (new) |
| --- | --- | --- |
| Structure | One collection, split into sections | Four separate tables |
| Columns | One set, shared by every section | Each group defines its own |
| Reach for it when | The rows agree | The rows do not agree |

`useTableGroupedRows` is right when every group is the same kind of thing and grouping is just a sort you can collapse — issues by status, orders by month. It is the wrong tool the moment the groups stop agreeing on what the columns mean. Utilization is a credit-card idea, pending payout is a processor idea, cost basis is an investment idea. Forcing all three into one grid produces a table that is mostly em-dashes, under a header row true of no row beneath it.

So each group owns a table, and the Card supplies the boundary the shared header row used to.

The template header comment carries the rule for extending it: **the test for splitting is column divergence, not group count.** Twelve groups with identical columns still belong in one table with the plugin.

## What it demonstrates

- **Divergent columns per group** — the reason the layout exists
- **One shared range** (1W/1M/3M/YTD/1Y) in the page header driving all four tables
- **Independent sorting** per table
- **Row expansion** into a full-width history chart, cursor-synced across charts in the same group via Recharts `syncId`
- **Compact layout below 720px** — folds to name plus headline figure with a coloured delta, with liability inversion so a rising credit-card balance reads red

## Screenshots

| Overview | Row expanded |
| --- | --- |
|  |  |

| Compact (<720px) | Expand all |
| --- | --- |
|  |  |

## Blocked on core

**Do not merge before these land** — the template calls three props that do not exist on `main` yet:

- #5995 — `hasRowClickExpansion`, `panelVariant`, divider-below-panel, and the expansion animation on `useTableRowExpansion`
- #5993 — `chevronPlacement` on `Collapsible`

## Rubric

**85/100 (B).** Full marks on structure, tokens, a11y and responsive behaviour. The two deductions:

- **Custom CSS (0/15)** — driven by the icon-to-chevron hover cross-fade on the group headers. Worth noting the rubric is ambiguous here: scoring by CSS *properties* and by *style keys* give materially different numbers for the same file. Flagging rather than tuning the template to a metric I cannot read unambiguously.
- **Naming** — `table-collapsible` and `table-grouped` are both collapsible, so the name does not separate them on its own. Chosen deliberately with that tradeoff understood; the description carries the differentiation instead, which is the field that actually drives retrieval.

## Test plan

- `pnpm -F @astryxdesign/cli typecheck:strict` — clean
- `pnpm -F @astryxdesign/cli typecheck:template-docs` — clean
- Preview verified at four viewports, zero console errors in all four
- `table-grouped` byte-identical to `main`

### DIFF
--- a/.changeset/table-collapsible-template.md
+++ b/.changeset/table-collapsible-template.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/cli': patch
+---
+
+[feat] New `table-collapsible` page template: several tables on one page, each in a collapsible card with its own columns, for groups that do not share a schema. Every table sorts on its own, one time range in the page header drives all of them, and rows expand in place into a full-width history chart. Below 720px each table folds to its name and headline figure.
+@ernestt
--- a/.changeset/table-collapsible-template.md
+++ b/.changeset/table-collapsible-template.md
@@ -2,5 +2,5 @@
 '@astryxdesign/cli': patch
 ---
 
-[feat] New `table-collapsible` page template: several tables on one page, each in a collapsible card with its own columns, for groups that do not share a schema. Every table sorts on its own, one time range in the page header drives all of them, and rows expand in place into a full-width history chart. Below 720px each table folds to its name and headline figure.
+[feat] New `table-collapsible` page template: several tables on one page, each in a collapsible card with its own columns, for groups that do not share a schema. Every table sorts on its own, one time range in the page header drives all of them, and rows expand in place into a full-width history chart. Below 720px each table folds to its name and headline figure. (#6065)
 @ernestt
--- a/apps/docsite/src/components/templateComponents.ts
+++ b/apps/docsite/src/components/templateComponents.ts
@@ -166,6 +166,10 @@ export const TEMPLATE_COMPONENTS: Record<
     () =>
       import('../../../../packages/cli/assets/templates/pages/table-filter/page'),
   ),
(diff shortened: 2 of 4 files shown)


ernestt:
Answering the playground question here for the record: the missing charts are not this PR's doing. `recharts` was never in the playground's module scope, so every chart in every dashboard template has been rendering as nothing — the runner returns a proxy of `() => null` for unresolved modules, which is why it failed silently rather than erroring.

Fixed on main in #6263, along with a warning for unresolved modules and a test that every template import resolves. Once that lands, the `AreaChart` in this template will draw in the playground with no change needed here.

humbertovirtudes (changes requested):
Narrowing the layout after sorting a desktop-only column leaves an invisible active sort controlling row order. The 1,474-line copyable interaction surface also has no committed functional tests, depends on the blocked row-expansion work in #5995, and needs explicit accessible treatment for expanded charts. Please clear or map hidden sort keys, add behavior coverage, and resolve the dependency before approval.

PR #6066, opened 2026-09-05
docs(spec): require outermost DOM styling target

imdreamrunner:
## Summary

Proposes one rule: when a stable Core component accepts `style`, `className`, or `xstyle`, every accepted prop applies to the same outermost DOM element. A component without one stable root must not expose unsupported general styling props.

## Why

Today Astryx requires accepted styling props to be forwarded and composed, but does not define their DOM target. Consumers may need implementation knowledge to know whether they style a wrapper, inner control, or detached surface.

## Semantic delta

**Before:** accepted styling props may target the component's chosen contract element.

**After (proposed):** they target the same outermost DOM element on every render path.

## Existing components

Popover, ContextMenu, OverflowList, Stack, PowerSearch, MultiSelector, and wrapped inputs have shapes or contracts that conflict with the **proposed** rule. They are not all violations of the current contract: Popover explicitly follows its current surface-styling contract. The proposal requires owners to choose migrations before the rule can become authoritative.

## Scope

Specification only. No runtime behavior or public types change in this PR.

## Verification

- `pnpm check:knowledge`
- Prettier
- `git diff --check`
- Independent focused review

### DIFF
--- a/docs/specs/AST-032/spec.md
+++ b/docs/specs/AST-032/spec.md
@@ -1,0 +1,149 @@
+---
+schema_version: 1
+template_version: 1
+kind: system-spec
+id: spec:AST-032
+authority: draft
+archive_reason: null
+superseded_by: null
+approved_by: null
+approved_at: null
+phase: proposed
+owners: [cixzhang, imdreamrunner]
+affects_architecture: [architecture:public-component-api]
+affects_families: [family:layout-primitives, family:input-fields]
+affects_contributing: [contributing:api-conventions]
+affects_consumer_docs:
+  [
+    principles,
+    Popover,
+    ContextMenu,
+    OverflowList,
+    Stack,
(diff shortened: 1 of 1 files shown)


cixzhang (changes requested):
Builders of compound and portal-backed components would lose stable styling APIs because of hidden DOM plumbing. The existing contract puts accepted styling props on the documented contract element; Popover specifically owns them on its painted surface, while its trigger and positioned layer are separate. The proposed rule instead requires the outermost node or removal, so an inert measurement peer in OverflowList and Field wrappers in TextInput/MultiSelector become breaking API decisions. Please keep the guarantee on each documented contract element and preserve the current overlay/non-DOM exceptions rather than binding API to incidental DOM ancestry.

[Reviewed by Robohands]

PR #6067, opened 2026-09-05
docs(spec): define local stacking and top-layer routing

cixzhang:
## Summary

- define `z-index` as a local-only mechanism inside a component-owned `isolation: isolate` surface
- require menus, dialogs, and equivalent floating interactions to use Astryx Layer and native browser top-layer hosting
- prohibit fixing stacking leaks by escalating AppShell, page, or overlay numeric bands
- inventory current compliant patterns and migration gaps across published packages and CLI-emitted templates
- make the production docsite Field/AppShell overlap a required Chromium regression fixture

## Current impact

The reproducible source audit covers 56 production declarations across 40 files. Confirmed migration areas include Field inputs, AppShell, InputGroup, SideNav, Outline, Table, Resizable, Schedule, ChartZoom, ChartTooltip, ListInput/TransferList, RichTextEditor, CLI page templates, Toast, non-modal BottomSheet, and non-modal Drawer.

The spec contains no pull-request-specific policy language. Existing implementation proposals are evaluated only after this spec lands.

## Validation

- Prettier on `docs/specs/AST-027/spec.md`
- `git diff --check`
- privacy scan of the staged diff
- two independent source/spec reviews
- live Chromium reproduction on `/components/TextInput`: an unisolated input and sticky AppShell header both computed to `z-index: 1`; the input won hit-testing after scrolling into the header

## Authority

The record is `authority: current`, approved by `cixzhang` on 2026-09-04. The exact-head owner approval is recorded separately for the spec gate.

No Changeset: specification-only change.

### DIFF
--- a/docs/specs/AST-027/spec.md
+++ b/docs/specs/AST-027/spec.md
@@ -1,0 +1,285 @@
+---
+schema_version: 1
+template_version: 1
+kind: system-spec
+id: spec:AST-027
+authority: current
+archive_reason: null
+superseded_by: null
+approved_by: cixzhang
+approved_at: 2026-09-04
+phase: accepted
+owners: [cixzhang, imdreamrunner]
+affects_architecture: [architecture:layer-runtime]
+affects_families:
+  [family:input-fields, family:layout-regions, family:overlay-dismissal]
+affects_contributing: []
+affects_consumer_docs: []
+---
+
+# Local stacking and browser top-layer routing system spec
+
+## Intent
+
+An Astryx component should be able to order its own painted parts without
+changing how unrelated components, page content, or application chrome stack.
+Menus, dialogs, and other floating interaction surfaces should escape ordinary
+document stacking through Astryx's Layer system and the browser top layer, not
+through larger `z-index` values.
+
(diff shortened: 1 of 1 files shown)


docs/specs/AST-027/spec.md
```diff
@@ -0,0 +1,288 @@
...
+- Define shadows, borders, or perceived elevation. Those remain visual-design
+  concerns.
+- Create a global numeric z-index scale or semantic page-wide z-index tokens.
+- Change which interaction dismisses a menu, dialog, popover, or other overlay.
+- Require `isolation: isolate` on components that have no local paint-order
+  relationship.
+- Treat a React portal as equivalent to browser top-layer promotion.
+- Approve the implementation proposed by
+  [PR #5693](https://github.com/facebook/astryx/pull/5693). That pull request is
+  evidence for the current gap, not authority for its solution.
```
cixzhang: We should exclude PR specific language in specs

cixzhang: Removed all PR-specific language from the specification and kept implementation proposals outside the durable policy.

PR #6069, opened 2026-09-05
feat(theme-neutral): mute dark palette edge

rubyycheung:
## Summary

- reduce realized chroma for Neutral dark-mode chromatic stops 5–25 to 50%
- use a gentler 65% multiplier for yellow and smoothly recover all families by stop 60
- preserve every light value, the full neutral ramp, exact endpoints, and dark stops 60–100
- update the reviewed palette artifacts, receipts, required template mirrors, specification, and regression coverage

## Scope

This PR changes palette values only. It does not change semantic token mappings, badge assignments, component behavior, typography, spacing, motion, or radius. All emitted colors remain solid; no opacity is introduced.

## Verification

- Neutral palette and template drift tests: 11 passed
- Neutral production build passed
- repository checks passed
- owner and template artifacts are byte-for-byte synchronized
- generated candidate digest matches both receipts
- verified light ramps, neutral ramps, and dark stops 60–100 are unchanged


cixzhang (approved):
Thanks, this makes the dark edge visibly quieter while preserving light, neutral, and high-tone ramps plus all shipped component colors and contrast.

[Reviewed by Robohands]

PR #6074, opened 2026-09-05
fix(Stepper): define the public context boundary and deprecate useStepperContext

### LINKED ISSUE #6060: Stepper: define the public boundary for context exports
## Problem

`@astryxdesign/core/Stepper` publicly exports both `useStepperContext` and `StepperContextValue`, but the exported value includes component-private coordination such as `previousActiveStep` and `registerStep`.

- Public barrel: [`packages/core/src/Stepper/index.ts`](https://github.com/facebook/astryx/blob/64e715a37244c678a1e4998ade881515b1531184/packages/core/src/Stepper/index.ts#L13-L18)
- Context shape: [`packages/core/src/Stepper/StepperContext.ts`](https://github.com/facebook/astryx/blob/64e715a37244c678a1e4998ade881515b1531184/packages/core/src/Stepper/StepperContext.ts#L31-L70)

The source itself describes some fields as internal, and the only repository consumer of `registerStep` is the built-in `Step`. However, exporting the hook and full interface makes those details part of the installable TypeScript surface. Internal Stepper changes can therefore become accidental consumer type breaks.

## Desired outcome

Decide and encode the supported context boundary:

1. If custom Step composition is supported, define and document a stable public subset while keeping registration, measurement, animation, and compact-layout coordination private.
2. If the hook/type are not supported public API, deprecate them and remove them only at an explicit compatibility boundary.

## Acceptance criteria

- The intended public use case for `useStepperContext` is explicit.
- Private Stepper coordination is not exposed through a required public interface.
- A compatibility/type test protects the chosen public surface.
- Consumer docs or deprecation guidance match the decision.

Alif416:
## Decision

`useStepperContext` and `StepperContextValue` are **not a supported extension point**. Three things settled it:

1. **The hook returns nothing the caller lacks.** A Stepper builds its context entirely from props the caller passed it — `activeStep`, `orientation`, `density`, `indicatorPosition`, `onStepClick`. The only fields genuinely new to it were `previousActiveStep` and `registerStep`, which are exactly the private coordination.
2. **Custom step composition cannot work through it.** The connector fill is choreographed inside `Step`, so a hand-rolled step reading the context still cannot draw a correct track. The export advertised a capability it could not fulfill.
3. **No consumers.** Nothing in Storybook, the CLI showcase blocks, docs, or tests calls it. The `StepperCustomContent` block already threads `active` from its own state — the idiomatic pattern.

Custom composition is supported through `<Step>` and its `children`, `indicator`, and `endContent` slots.

## Approach

Follows the doctrine already stated in `Typeahead/busyIndicatorLane.tsx`: *an `@internal` tag is a note to a reader, not a boundary.* The seam is closed by module boundary instead.

- `StepperContext.ts` splits the shape. Public `StepperContextValue` keeps the six safe reads; `StepperCoordination` extends it with `previousActiveStep` and `registerStep`. One object on the wire — the split governs which *declaration* a consumer can reach, not a second allocation per render.
- `StepperCoordination` and `useStepperCoordination` are never re-exported from the barrel.
- `useStepperContext` and `StepperContextValue` stay exported and `@deprecated`, so removal lands at an explicit major rather than in a patch.

## Acceptance criteria

| Criterion | Where |
|---|---|
| Intended public use case is explicit | `@deprecated` blocks in `StepperContext.ts` + the rationale block in `index.ts` |
| Private coordination off the required public interface | `StepperCoordination` split, internal to the module |
| Compatibility/type test protects the surface | `Stepper.public.test.ts` |
| Consumer docs match the decision | `Stepper.doc.mjs` — all three locale blocks (full, `--dense`, `--zh`) |

## Testing

`Stepper.public.test.ts` follows the existing `*.public.test.ts` precedent: `expectTypeOf` locks the private fields off both the public type and the hook's return, asserts the deprecated names are *still* exported, and asserts the barrel's code never references the internal names.

**The guard was verified to bite**, not assumed: re-widening `StepperContextValue` with `registerStep` made exactly the two matching assertions fail typecheck, while the `previousActiveStep` ones correctly stayed silent.

One caveat documented in the file header — these assertions are enforced by `pnpm -F @astryxdesign/core typecheck`, **not** by the test run, and a failure surfaces as the cryptic `TS2554: Expected 2 arguments, but got 1`.

- Core suite — 9009 tests / 308 files passing
- `pnpm check:repo` — green
- `astryx component Stepper --dense` renders the new guidance

## Notes for review

No runtime behavior changes — this is a type-surface and documentation change. The one non-type edit is `Stepper.tsx` annotating its provider value with the wider type, and `Step.tsx` switching to `useStepperCoordination`.

It should fixes #6060


cixzhang (changes requested):
This cannot ship as the v0.6.0 compatibility boundary yet.

Consumers of v0.5.2 can read `previousActiveStep` and call `registerStep(index)`, and object literals can satisfy the exported `StepperContextValue`. This PR removes those released fields from the public type, creating another source break instead of repairing the one in v0.5.3.

Please preserve the complete v0.5.2 public shape; make the new disabled-state argument optional with a `false` default; move only the five v0.5.3 collapsed-layout fields to the private coordination type/hook; and add compile-time regressions for a v0.5.2-shaped object plus the one-argument call. Because this intentionally settles a released API break for v0.6.0, the changeset must be `[breaking]`/minor with migration guidance, not `[component]`/patch. Then rebase on current main and rerun exact-head CI.

cixzhang:
Thank you for digging into this and for putting together the public-surface tests and documentation. The underlying issue was resolved in #6097, which preserves the released context fields while separating internal coordination and simplifying the collapse-threshold API.

Since that supersedes this approach, I’m closing this PR rather than asking you to rework it against current main. Your investigation helped clarify the boundary we needed—thanks again for the thoughtful contribution.

PR #6078, opened 2026-09-05
fix: runaway auto-scroll loop in hover-highlighted menus

### LINKED ISSUE #6077: Menus: runaway auto-scroll loop when pointer rests near the bottom of a scrollable menu
When a scrollable menu's highlight follows the mouse, scrolling the highlighted item into view moves the *next* item under the stationary pointer, which re-highlights and scrolls again — an endless auto-scroll loop with no user input. Two surfaces share this bug:

1. **Chat trigger menu** (`Chat/useTriggerMenu.tsx`) — `onMouseEnter` sets `highlightedIndex`, and an effect calls `scrollIntoView` on every highlight change without distinguishing hover from keyboard navigation.
2. **DropdownMenu / ContextMenu / BreadcrumbMenu** (`DropdownMenu/menuItemHover.ts`) — `focusMenuItemOnHover` calls `el.focus()` without `{preventScroll: true}`, so focusing an off-screen item scrolls it under the pointer.

Image

## Steps to reproduce (either surface)

1. Open a menu whose item list overflows (more items than fit in the scroll container).
2. Keep the mouse pointer completely still, resting on an item near the bottom edge of the visible list.
3. Press ArrowDown once or twice to move the highlight toward the bottom.

## Actual behavior

The menu scrolls endlessly on its own.

## Expected behavior

Hover-driven highlight changes should never scroll the menu. `scrollIntoView`/focus scrolling should apply only to keyboard-driven highlight changes (or use `{preventScroll: true}` for hover-driven focus).

## Minimal reproduction (DropdownMenu)

Open https://astryx.atmeta.com/playground, open the menu, and keep the pointer hovering over the "Template" item at the bottom of the list.
Or 

```tsx
function MenuRepro() {
  const items = Array.from({length: 25}, (_, i) => ({
    label: `Item ${i + 1}`,
    onClick: () => {},
  }));

  return <DropdownMenu button={{label: 'Open'}} items={items} />;
}
```

## Suggested fix

- `Chat/useTriggerMenu.tsx`: skip the scroll-into-view effect for one run when the highlight change came from `onMouseEnter`.
- `DropdownMenu/menuItemHover.ts`: `el.focus({preventScroll: true})` — the focus highlight does not require scrolling; keyboard navigation has its own scroll-into-view handling.

## Environment

- Package: `@astryxdesign/core`
- Browsers: all (pure pointer/scroll interaction)

faga295:
Fixes #6077

When a scrollable menu's highlight follows the mouse, scrolling the highlighted item into view moves the next item under the stationary pointer, which re-highlights and scrolls again — a runaway auto-scroll loop with no user input.

Two instances of the same pattern:

- **`Chat/useTriggerMenu.tsx`**: hover-driven highlight changes now skip one run of the scroll-into-view effect (`hoverHighlightRef`); keyboard navigation still scrolls the highlighted option into view.
- **`DropdownMenu/menuItemHover.ts`**: `focusMenuItemOnHover` now calls `el.focus({preventScroll: true})`. This is a one-line root-cause fix that covers DropdownMenu, ContextMenu, and BreadcrumbMenu items via the shared helper — the focus highlight does not require scrolling, and keyboard navigation has its own scroll-into-view handling.

## Tests

- `DropdownMenu.test.tsx`: the existing hover-focus test now asserts `focus` was called with `{preventScroll: true}` (regression guard for the auto-scroll loop).
- `ChatComposerInput.test.tsx`: coverage for the hover-skip behavior in `useTriggerMenu`.

### DIFF
--- a/.changeset/hover-highlight-no-autoscroll-option-lists.md
+++ b/.changeset/hover-highlight-no-autoscroll-option-lists.md
@@ -1,0 +1,9 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Keep hover from auto-scrolling open option lists (#6077)
+
+In a scrollable listbox whose highlight follows the pointer, scrolling the highlighted option into view moved the next option under the stationary pointer, whose mouseenter re-highlighted and scrolled again — an endless loop with no user input. This was already fixed for DropdownMenu and Chat; it now covers the remaining combobox-style paths through a shared highlight owner: Selector, MultiSelector, Typeahead, and DateTimeInput hover highlights move only the highlight, while keyboard navigation still scrolls the highlighted option into view.
+
+@liuzhaochen03
--- a/.changeset/hover-highlight-no-autoscroll-option-lists.md
+++ b/.changeset/hover-highlight-no-autoscroll-option-lists.md
@@ -4,6 +4,6 @@
 
 [fix] Keep hover from auto-scrolling open option lists (#6077)
 
-In a scrollable listbox whose highlight follows the pointer, scrolling the highlighted option into view moved the next option under the stationary pointer, whose mouseenter re-highlighted and scrolled again — an endless loop with no user input. This was already fixed for DropdownMenu and Chat; it now covers the remaining combobox-style paths through a shared highlight owner: Selector, MultiSelector, Typeahead, and DateTimeInput hover highlights move only the highlight, while keyboard navigation still scrolls the highlighted option into view.
+In a scrollable listbox whose highlight follows the pointer, scrolling the highlighted option into view moved the next option under the stationary pointer, whose mouseenter re-highlighted and scrolled again — an endless loop with no user input. This was already fixed for DropdownMenu and Chat; it now covers the remaining combobox-style paths through a shared highlight owner: Selector, MultiSelector, Typeahead, DateTimeInput, and CommandPalette hover highlights move only the highlight, while keyboard navigation still scrolls the highlighted option into view.
 
 @liuzhaochen03
--- a/.changeset/hover-highlight-no-autoscroll-option-lists.md
+++ b/.changeset/hover-highlight-no-autoscroll-option-lists.md
@@ -6,4 +6,4 @@
 
 In a scrollable listbox whose highlight follows the pointer, scrolling the highlighted option into view moved the next option under the stationary pointer, whose mouseenter re-highlighted and scrolled again — an endless loop with no user input. This was already fixed for DropdownMenu and Chat; it now covers the remaining combobox-style paths through a shared highlight owner: Selector, MultiSelector, Typeahead, DateTimeInput, and CommandPalette hover highlights move only the highlight, while keyboard navigation still scrolls the highlighted option into view.
 
-@liuzhaochen03
+@faga295
--- a/apps/storybook/rtl-audit/targets.json
+++ b/apps/storybook/rtl-audit/targets.json
@@ -22,6 +22,20 @@
     }
   },
   {
+    "component": "CommandPalette",
+    "storyId": "core-commandpalette--default",
+    "dims": [
+      "D2"
+    ],
+    "setup": {
+      "click": "button:has-text(\"Open Command Palette\")"
+    },
+    "selectors": {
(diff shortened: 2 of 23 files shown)


cixzhang (changes requested):
Thanks, the changed DropdownMenu and Chat paths now keep hover stationary while keyboard navigation still scrolls in Chromium. People using Selector, MultiSelector, Typeahead, or DateTimeInput can still have hover move the list because those paths pair hover-updated highlights with unconditional `scrollIntoView`; Selector moved 376px in the same 25-item probe. Please either cover those paths through their shared highlight owner, or narrow #6077 and track the remaining instances separately. Also add the patch changeset required for this consumer-visible fix.

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks—my prior review asked for the sibling option-list paths. This head fixes Selector, MultiSelector, Typeahead, and DateTimeInput while preserving keyboard scrolling, and adds the patch changeset.

CommandPalette still bypasses the hook’s hover callback: its items call the raw setter, then both the new `useCombobox` effect and `CommandPaletteItem` call `scrollIntoView`. In exact-head Chromium, one hover and one ArrowDown each produced two calls (main produces one), so the stationary-pointer runaway path remains and this head adds duplicate layout work. Please route CommandPalette hover through the hover-aware path and leave one keyboard scroll owner, with a regression test.

[Reviewed by Robohands]

cixzhang (dismissed):
Thanks, this now closes the shared option-list paths. Chromium confirms hover stays still and keyboard navigation scrolls once, including CommandPalette. Could you credit `@faga295` in the changeset?

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks for helping to fix this. However the fix introduces multiple new API surfaces without clear intention. Could we fix the issue without adding API?

faga295:
> Thanks for helping to fix this. However the fix introduces multiple new API surfaces without clear intention. Could we fix the issue without adding API?

ok

cixzhang (changes requested):
Thanks—the auto-scroll fix is now solid. Chromium confirms hover stays put and keyboard navigation scrolls once across all seven menu paths.

My prior review asked to avoid new API. This head only changes the changeset credit, so `CommandPaletteContextValue` still replaces public `setHighlightedIndex(index)` with required `onItemMouseEnter(item, index)`. The same consumer compiles on main and fails here. Could you keep the exported context shape compatible and route hover through an internal seam? The patch changeset is right once that API break is gone.

[Reviewed by Robohands]

cixzhang (changes requested):
My prior review asked to keep the CommandPalette context compatible; that and the seven hover/keyboard paths are now correct.

Large command palettes now do work proportional to every item on each bubbled mouseover: `CommandPalette.tsx:523` linearly scans the list even when the pointer only moves between children inside one result. In Chromium, crossing 50 items performed 1,275 comparisons versus 0 on main; ten within-item moves added 500. Could you use an O(1) lookup and add a two-size operation-count perf test?

[Reviewed by Robohands]

cixzhang (changes requested):
People with large command palettes now get a flat hover path: Chromium recorded 50 Map probes crossing 50 items and 10 for ten same-item moves; hover stayed still and keyboard scrolling remained intact. This fixes the runtime half of my prior review’s O(1) path plus two-size guard request.

The test does not guard that path. It calls `createValueLookup` directly. I restored the old linear `findIndex` in the delegated mouseover handler, and all three perf tests plus the CommandPalette behavior suite still passed. Could you drive the actual delegated mouseover path at two list sizes so that regression fails the test?

[Reviewed by Robohands]

cixzhang (approved):
Thanks, this test drives delegated mouseover at 50 and 500 items and fails the old linear lookup. Chromium keeps every hover path stationary and keyboard scrolling intact.

[Reviewed by Robohands]

cixzhang (approved):
Thanks—the RTL reroll passes in Chromium. My prior approval confirmed the hover, keyboard, API, and performance fixes; those runtime bytes are unchanged.

[Reviewed by Robohands]

PR #6079, opened 2026-09-05
docs(theme-neutral): document Figtree and its loading snippet

### LINKED ISSUE #5991: [Bug] Neutral theme is described as system-font based but declares unloaded Figtree
## Description

`@astryxdesign/theme-neutral@0.5.2` advertises itself as using “system fonts,” but its published theme sets both body and heading typography to `Figtree`.

Figtree is not bundled and the theme does not load it. In a clean Vite + React 19 app, the browser silently falls back to the system font.

This is especially confusing because the generated `astryx docs theme` output describes Neutral as:

> Muted, minimal aesthetic with system fonts. A good starting point.

The published theme source instead declares:

```ts
typography: {
  body: { family: 'Figtree', fallbacks: '...' },
  heading: { family: 'Figtree', fallbacks: '...' },
}
```

`astryx docs typography` correctly says Astryx does not load font files and that apps must do so. The Theme docs’ Neutral description contradicts that behavior and gives no Neutral-specific loading snippet.

This is related to #5015, but it is a remaining documentation/package-description mismatch in the released theme.

## Reproduction

1. Create a Vite React 19 app.
2. Install `@astryxdesign/core@0.5.2` and `@astryxdesign/theme-neutral@0.5.2`.
3. Import the prebuilt CSS and wrap the app with the documented provider:

```tsx
import {Theme} from '@astryxdesign/core/theme';
import {neutralTheme} from '@astryxdesign/theme-neutral/built';
import '@astryxdesign/theme-neutral/theme.css';

<Theme theme={neutralTheme}>
  <Heading level={1}>Test</Heading>
</Theme>
```

4. Inspect the heading in a clean browser profile.

Observed:

- Computed font stack: `Figtree, -apple-system, system-ui, "Segoe UI", Roboto, Helvetica, Arial, sans-serif`
- No `.woff`, `.woff2`, `.ttf`, `.otf`, Google Fonts, or Google Fonts static resource is requested.
- The published package contains no font files.
- Text therefore renders in the system fallback, not Figtree.

## Expected behavior

Either:

1. Describe Neutral as a Figtree theme and provide the exact consumer-side font-loading snippet, or
2. Change Neutral to use only the documented system stack.

The current “system fonts” description makes a missing Figtree font look like an integration failure.

## Astryx version

- `@astryxdesign/core@0.5.2`
- `@astryxdesign/theme-neutral@0.5.2`
- `@astryxdesign/cli@0.5.2`

## Environment

- Vite 8.2.2
- React 19.2.8
- Chrome (clean local profile)
- macOS 15.7.4, Apple Silicon

AKnassa:
Fixes #5991

## What this does

Makes the Neutral theme honest about its font. The docs said "system fonts", but the theme actually asks for Figtree, a webfont the app has to load itself. Anyone who followed the docs got a silent fallback with no clue why.

## Why

A missing webfont looks like an integration failure. Per #5047 and #5048 ("enable rather than deter"), the fix keeps Figtree and documents it instead of stripping the theme back to system fonts. Direction matches @Kyujenius's analysis on the issue.

## What changed

- Neutral README now says "Figtree typography" and gains a Fonts section with the copy-paste Google Fonts snippet, same shape as Stone's.
- The section also says skipping the webfont is a supported choice, so offline or no-third-party-request apps know the fallback is intentional, not broken.
- The `astryx docs theme` table row for Neutral now names Figtree too.
- Patch release note for @astryxdesign/theme-neutral.

## How to check

Read packages/themes/neutral/README.md, Fonts section, or run `astryx docs theme` and find the Neutral row.


cixzhang:
Thanks AK! @rubyycheung was taking a more in depth look at the neutral theme and fixing various issues with it so this should actually be resolved now.

cixzhang:
Oh actually checking main I guess the docs haven't updated yet but the values have

packages/themes/neutral/src/neutralFonts.test.ts
```diff
@@ -0,0 +1,160 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
```
cixzhang: I think a test for the readme is likely unnecessary. We could skip this file.

packages/cli/test/theme-doc-fonts.test.mjs
```diff
@@ -0,0 +1,77 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
```
cixzhang: Similarly we likely don't need a test for this.

cixzhang (changes requested):
Back to you mainly for some test clean up. Was curious if this might be caused by something in context on our side. Perhaps because the issue was marked as a bug rather than a doc issue

packages/themes/neutral/src/neutralFonts.test.ts
AKnassa: Removed in 54a225879.

packages/cli/test/theme-doc-fonts.test.mjs
AKnassa: Removed in 54a225879.

AKnassa:
Both test files are removed in 54a225879; the PR is now the README, the doc row, and the changeset. The tests came from a test-first habit plus the existing react-version-sync.test.mjs docs guard reading as precedent, not from the bug label.

PR #6051, opened 2026-09-04
fix: repair runtime and tooling audit regressions

cixzhang:
## Summary

This consolidates the verified repairs from the Sep 3 merge audit into one PR:

- restore canonical visual coverage by component ownership and reject zero-shot plans
- make spec-owner approval exact-head, fail-closed, and safely backfillable
- make CLI result counts, debug privacy, and package-manager detection truthful
- repair streamed Markdown, ChatComposer focus/history, and Vega view lifecycle
- harden RTL centering autofixes and vibe-test CSS override detection
- route sandbox tests exactly once and split PR/deploy test lanes to fit runner budgets

## Why

The daily audit found release-gate holes, runtime regressions, telemetry/privacy contract gaps, orphaned tests, and parser escape hatches. The original five focused PRs were batch-closed, so this replacement keeps one review and landing surface while preserving atomic commits.

## Validation

- all source repair branches received independent SHIP reviews after blocker fixes
- consolidated integration review: SHIP; no cherry-pick drift or private-data leaks
- `pnpm check:repo`
- `pnpm build`
- node suite with bounded workers: all branch-relevant tests pass; three GNU-`cp` fixture tests are Linux-only and fail on macOS as before
- UI suite: one unrelated Selector timing flake, passed in isolation
- gh-pages timing test passed in isolation (32/32)
- 1,680 tooling/setup/CI tests, 994 core/Vega tests, 489 CLI tests, 303 visual-gate tests, and 105 spec-owner tests passed in focused runs
- real Chromium probes covered Markdown streaming and ChatComposer selection/focus paths
- CI workflow contract tests prove each Vitest project belongs to exactly one lane and historical `test`/deploy gates remain intact

## Rollout notes

- The open RTL debt-baseline PR #5988 is not a dependency and is not safe to land in its current form; `pr-rtl` remains advisory.
- After this PR lands, backfill `spec-owner-approval` on current open heads using the new `backfill=true` mode, verify no missing contexts, then require that status on `main`.

PR #6080, opened 2026-09-06
perf(ci): give the heavier test lane the larger runner

### LINKED ISSUE #4339: The `test` job is the CI critical path on re-pushes, merge queue, and main

Han5991:
## Why

`test-ui` and `test-node` hold their runners the wrong way round relative to the work they carry.

| lane | runner | job | suite step |
|---|---|---|---|
| `test-ui` | `2-core-ubuntu-arm` | **14m52s** | 834s |
| `test-node` | `4-core-ubuntu` | 3m18s | 165s |

Run [34000689684](https://github.com/facebook/astryx/actions/runs/34000689684) (`main`, success). Reproduced on 33994177354 (926s/189s) and 33980770638 (974s/194s), and on deploy runs 33980455596 and 33952374356.

The 4-core box is under the three-minute lane. The fifteen-minute lane — the one nearest the ~20 minute wall the lane comment itself describes — is on the 2-core arm box.

## The number that put it there

The comment introduced with the lane split says:

> `node` is the heavier lane — locally 570s against the ui project's 244s — so it takes the larger supported runner while `ui` stays on the 2-core arm box.

That comparison does not appear to be like-for-like. #6051's own validation notes record the node suite as run **"with bounded workers"**; the ui line carries no such note. And `node` is precisely the project that a bounded pool penalises — `vitest.config.ts` documents that it runs on the forks pool because CLI tests call `process.chdir()`, and that "several CLI suites spawn a fresh `node bin/astryx.mjs` **per assertion** (real process boundary)".

The CI numbers are hard to reconcile with 570s any other way: unbounded, the node lane finishes in 165s **while also paying** the cold `ensureCoreBuilt()` core build that a warm local tree skips (`build-theme.data-tokens.test.mjs:117` — "ensureCoreBuilt() only checks that dist exists"). The local run had strictly less work and still took 3.5x longer.

I've rewritten the comment to state what CI measures, and to record why the local figure isn't comparable.

## What this changes

Four `runs-on:` values, two in each workflow. `deploy.yml` carried the same allocation.

Nothing else moves. In particular the repo-wide guardrail steps stay on `test-ui`, so the "they ride the LIGHTER lane" claim is now simply corrected in place rather than acted on — they cost ~18s of an 892s job, and moving them is a separate change that isn't worth coupling to this one.

## Verification

`.github/scripts/ci-test-routing.test.mjs` passes unchanged — **17/17**. Both labels are already in `KNOWN_RUNNERS` (`:47-52`) and no assertion in the file binds a lane to a runner, which is why this needed no test edit.

The real verification is this PR's own CI run: it reports the swapped allocation directly, against the numbers above.

## What this does not do

This does not close [#4339](https://github.com/facebook/astryx/issues/4339)'s original criterion. On run 34000689684 `test-ui` finished 385s after `build-sandbox`, and the workflow ended 17s later — the test lane is still the critical path, just no longer a re-push casualty. Subdividing `ui` is the next lever and is deliberately not in this PR; a `strategy: matrix` inside the existing `test-ui` job passes all 17 contract cases, so it can land separately on measured numbers.

Filed because the headroom is thin: `test-ui` is at 14m52s against a wall its own comment puts at ~20 minutes, and the ui project went 439s → 834s in roughly three weeks.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -180,13 +180,14 @@ jobs:
   #
   # The projects are independent by construction (see vitest.config.ts: `ui` is
   # jsdom + StyleX, `node` is everything else), so running them as two jobs
-  # costs no duplicate execution and each lane carries roughly half the work.
-  # `node` is the heavier lane — locally 570s against the ui project's 244s —
-  # so it takes the larger supported runner while `ui` stays on the 2-core arm
-  # box.
+  # costs no duplicate execution. They do NOT carry half the work each: on this
+  # CI, `ui` is ~5x the heavier lane, so it takes the larger runner.
   #
-  # The repo-wide guardrail steps ride the LIGHTER lane, so the slow one is
-  # nothing but tests.
+  # The earlier local figure that put `node` ahead (570s vs 244s) was not a
+  # comparable measurement — only the node run was bounded to a small worker
+  # pool, and node is the project whose cost is per-assertion CLI subprocesses.
+  #
+  # The repo-wide guardrail steps ride `test-ui` (~18s of it).
   #
   # `test` remains a job, so the required check of that name still exists and
   # still reports for every PR. This mirrors the `build` join over
@@ -194,7 +195,7 @@ jobs:
   test-ui:
     needs: [check-scope]
     if: ${{ always() && !cancelled() }}
-    runs-on: 2-core-ubuntu-arm
+    runs-on: 4-core-ubuntu
     permissions:
       contents: read
     steps:
@@ -252,7 +253,7 @@ jobs:
   test-node:
     needs: [check-scope]
     if: ${{ always() && !cancelled() }}
-    runs-on: 4-core-ubuntu
+    runs-on: 2-core-ubuntu-arm
     permissions:
       contents: read
     steps:
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -79,8 +79,11 @@ jobs:
   # independent by construction (vitest.config.ts), so they split cleanly; the
   # typecheck gates need the `pnpm build` they follow, so they stay together in
   # their own lane and take the build with them.
+  #
+  # `ui` is the heavier project by ~4x here, so it takes the larger runner (see
+  # ci.yml's lane comment).
   test-ui:
-    runs-on: 2-core-ubuntu-arm
+    runs-on: 4-core-ubuntu
     permissions:
       contents: read
     steps:
@@ -93,7 +96,7 @@ jobs:
       - run: pnpm vitest run --project ui
 
   test-node:
-    runs-on: 4-core-ubuntu
+    runs-on: 2-core-ubuntu-arm
     permissions:
       contents: read
     steps:


Han5991:
## Measured

This PR's own CI run against a `pull_request` run on the same base. Both runs are `pull_request`, same suite, no test-file changes in this branch.

| | baseline [34000689684](https://github.com/facebook/astryx/actions/runs/34000689684) | this PR [34004880026](https://github.com/facebook/astryx/actions/runs/34004880026) | Δ |
|---|---|---|---|
| `test-ui` job | 892s · `2-core-ubuntu-arm` | **521s** · `4-core-ubuntu` | **−371s / −41.6%** |
| `test-ui` suite step | 834s | **459s** | −375s / −45.0% |
| `test-node` job | 198s · `4-core-ubuntu` | 353s · `2-core-ubuntu-arm` | +155s / +78.3% |
| `test-node` suite step | 165s | 305s | +140s / +84.8% |
| **`test` gate** (= slower lane) | **892s** | **521s** | **−371s / −41.6%** |
| whole CI run | 951s | **615s** | −336s / −35.3% |

Both lanes green.

## What it settles

**`ui` scales with cores; `node` less so.** Doubling ui's cores took 45% off its suite step. Halving node's cost it 85% — real, but from a much smaller base. That ordering is the opposite of the one the lane comment encoded, and it is now measured on this CI rather than inferred.

**`test` is no longer the critical path by any meaningful margin.** On the baseline, `test-ui` finished **385s** after `build-sandbox`. Here it finishes **22s** after it (01:59:25Z vs 01:59:03Z). The gate and the longest build job now land together.

**Headroom.** `test-ui` goes from 14m52s to 8m41s against the ~20 minute wall the lane comment describes. That was the thing I was actually worried about — ui went 439s → 834s in about three weeks, and at 892s there were roughly five minutes left.

## The cost, stated plainly

This is not free. Weighting each lane by its cores:

- baseline: 892×2 + 198×4 = **2,576** core-seconds
- this PR: 521×4 + 353×2 = **2,790** core-seconds

**+8.3% core-seconds for −41.6% gate wall clock.** If runner-seconds are the binding constraint rather than latency, that trade is arguable and I'd rather it be argued explicitly than hidden.

## Follow-up this makes smaller

Sharding `ui` was the obvious next lever at 834s. At 459s it is no longer urgent, and I'd rather not spend the complexity until the number asks for it again. Recording the option here so it doesn't have to be rediscovered: a `strategy: matrix` inside the existing `test-ui` job passes all 17 cases in `ci-test-routing.test.mjs` unchanged.

## Caveat

n=1 on each side. The baseline is consistent with the wider sample I looked at (`test-ui` 834–974s across 33994177354, 33980770638 and others), so 459s sits well outside that band — but it is still one run against one run.

cixzhang (approved):
Validated against the exact CI workload. `4-core-ubuntu` is the right non-default pool for `test-ui`: the PR run reduced the 384-file UI suite from 834s to 459s and the full job from 892s to 521s.

Keep `pnpm vitest run --project ui` without an explicit `--maxWorkers`. Vitest 4.1.10 deliberately uses `availableParallelism() - 1` in run mode, so this gives the four-core runner three test workers plus coordinator/Vite headroom. Forcing `100%` would add a fourth worker while each worker has a 4 GiB heap ceiling on a ~15 GiB runner, increasing OOM/flakiness risk. The measured 45% reduction confirms that the default pool is already using the added capacity effectively.

Please rebase after #6113 lands and preserve its positive-routing skip conditions; this runner swap is otherwise independent.

PR #6083, opened 2026-09-06
fix(core): guard TextInput onEnter against IME conversion commits

### LINKED ISSUE #6082: [Bug] TextInput onEnter fires when confirming Japanese IME conversion
### Description

TextInput calls `onEnter` when Enter is used to confirm a Japanese IME conversion. Applications using this callback for submission or saving can therefore trigger the action before the user intends to submit.

**Expected:** Enter used to confirm an IME conversion should not call `onEnter`. A subsequent Enter after composition has finished should call it once.

**Actual:** Confirming the conversion to `日本語` also increments the `onEnter` counter from 0 to 1, as shown in the recording.

https://github.com/facebook/astryx/blob/cde73c74d84b222692b606c78e1054ca3194a6af/packages/core/src/TextInput/TextInput.tsx#L431-L438

TextInput checks only `e.key === 'Enter'` before calling `onEnter`. The existing `isImeKeyEvent` helper checks both `isComposing` and `keyCode === 229`, but is not used here.

Related: #4892 (systematic IME composition guards).

### Reference

[GoogleChrome modern-web-guidance: IME-safe enter-to-submit (v0.0.186)](https://github.com/GoogleChrome/modern-web-guidance/blob/v0.0.186/skills/modern-web-guidance/guides/forms/ime-safe-enter-submit.md) explains composition checks for custom Enter handlers, including Safari event-ordering behavior and the `keyCode === 229` fallback. Its examples use a textarea; this report concerns TextInput's custom keydown callback, rather than native implicit form submission.

### Reproduction

[Reproduction in Astryx Playground](https://astryx.atmeta.com/playground#code=fVTNbhMxEL73KUbhkEREWRoqhNLdCCnKoRV_aisuUQ_OrpNYdeyV7Q0Jqz2UAxdugISEEDcuwBuAeBiCKm68ArPe7q4DLblk7flm_H2eb9xMNIWQMypMc39nhy1iqQykuHtsiKEZTJVcQFNREuaAMn5CV-ZAxIkpAfeINmq9iqhmM-GFUlGvwjh5WDQ8-0_OfbKWNmGHrmxGRKck4QamiQgNkwKqqgcPRidUm1Yb0h2AUAptYLwkPKEd0NQ8yb9OIYBSSqvZbO_XSLpEydpCR_bTxfpIjInZ-HTQGp9iFqYpahIloIWfAH7ElqDNmtMgTZ-yyMz7sNe71YGYRBEm9qG3l2UDi0W0VQ0RU9RqCBpLqgwLCW_AjMRBeruCInjeG1Qa-4AqYSQMVWBQrO9hNGdziY3rPICTdUzhkMREUOypwRpARASxolqXNWQuf8rUAswc-y4FMtFIqevUGa1i5EmjPkhRpGlD1hqIgVsu7rFTmMwIE0Cm-fflCSSXWteY0FAuqIbduoTvIf96Val2zuBkQnnQqFSxPN5wALbhQWr_MmdfiuGciBmGSi-40ZiTkM4lj6gKGj-_v_z97d3m_PPm_Ovm-evN-Rv48eIVXLz9ePH-y69PH9zjLtUEKdouGFjn1b_KSi288yWTic4x4263W647Bd_cUnVaVjPz3PtAjzmo0m1bJ1Z2273T2QpMpEJtfWj24hVoyVkEYaIUkhtKLlXzKvQRiVii-3DXDdYutpxwLqSYVWSmUphj9oxu-33rqgBtzrFqWgxcl1MxM3O3Gb5XVN06CMUPhgXj4s6wwuHxo4fdYjLZdN2y2-3M93KsYyq7rNeSu4VLGgsSt1r5jHTQUxFd2XYWw11lcgZndB2kFvCXulrfjSION2E3-5dkfkTbVZsz5Mwt1nbivpfTrZeTxBgpHLDBIQ8axfa2MYechWfXGtOOQPkEXmVZ-85d1_cjnL16LH2vOL963zz7wBXLqh1YLvsD)

1. Open the Playground and enable a Japanese IME.
2. Focus the input. The `onEnter calls` counter starts at 0.
3. Type `にほんご` and convert it to `日本語`.
4. Press Enter once to confirm the conversion.
5. Observe that `onEnter calls` becomes 1 even though Enter was used only to confirm the IME conversion. It should remain 0.

The demo keeps the input value and displays a counter and the value captured by each `onEnter` callback.

Recording:

https://github.com/user-attachments/assets/8e7f12c2-b8e5-4385-9ea8-0722b6ded09c

### Astryx Version

Hosted Astryx Playground (exact deployed package version unknown; also present in main at cde73c74d84b222692b606c78e1054ca3194a6af)

### Environment

- OS: macOS Tahoe 26.6.2 (25G83)
- Google Chrome 152.0.7977.82 (Official Build), Stable, arm64 — reproduced
- Google Chrome 155.0.8040.2 (Official Build), Dev, arm64 — reproduced
- Safari 26.6.2 (21624.5.1.11.3) — reproduced
- Firefox: could not reproduce in testing
- Input method: macOS Japanese IME

ManoharPaturi:
Fixes #6082

## Summary

Pressing Enter to commit a Japanese/Chinese/Korean IME conversion fired `onEnter` — so applications using it for submit/save triggered the action before the user intended to submit. `TextInput` checked only `e.key === 'Enter'`; the repo's canonical IME predicate (`isImeKeyEvent`, checking `isComposing` and the legacy `keyCode === 229` — see `utils/ime.ts`) was not applied here, though Selector/Typeahead/TimeInput already guard the same way for their Enter handling.

## Change

`onEnter` now fires only for `e.key === 'Enter' && !isImeKeyEvent(e.nativeEvent)`. The consumer's `onKeyDown` still receives every raw keydown, including composing ones, so custom composition handling is unaffected — the guard is scoped to the semantic `onEnter` callback exactly as the issue requests.

## Tests

Two new cases in `TextInput.test.tsx`, following the existing Selector IME-guard test style (`fireEvent.keyDown` with `isComposing: true` and `keyCode: 229`):

- composing Enters do not call `onEnter`; a subsequent real Enter calls it once
- `onKeyDown` still fires for composing keydowns

`vitest run packages/core/src/TextInput` → 76 passed (74 pre-existing + 2 new). Core `tsc --noEmit`, eslint on touched files, `check:sync`, and `check:changesets` all clean. Doc entry notes the IME-safe behavior; changeset included.


cixzhang (approved):
Thanks, this keeps IME users from triggering submit or save while committing a candidate. The shared guard covers both the modern `isComposing` signal and the Safari/legacy `keyCode === 229` fallback, still forwards the raw `onKeyDown` event, and leaves the next ordinary Enter unchanged. The focused TextInput and IME suites pass (80 tests).

[Reviewed by Robohands]

azu:
> [CI / pr-rtl (pull_request) Failing](https://github.com/facebook/astryx/actions/runs/34015486373/job/101438905625?pr=6083)

Is this a flaky test?

ManoharPaturi:
looked into the log. the audit itself didn't fail anything, all three verdicts are 0 fail (auto/pm/dec). the job dies before that on the artifact restore step with digest-mismatch: error, the run fetches the baseline capture to diff against and the digest it recorded doesn't match what's stored now.

the same failure shows on every PR pushed in that window yesterday (#6084, #6085, #6086 all hit it), while pr-rtl passes on newer runs like #6115, so it looks like the baseline moved underneath that batch rather than anything in this diff.

a re-run of the job against today's baseline should clear it. happy to retrigger with an empty commit too if that's easier, just didn't want to push without checking first since it could drop the approval.

azu:
It doesn't seem to have fallen in the latest PR, so I think it's good to rebase it to the latest of the main branch.

ManoharPaturi:
yeah that matches what i saw, the failure is tied to that batch of runs. i've left the branch untouched on purpose because cixzhang's approval is on this head and pushing would dismiss it. a re-run of the job or an update-branch from a maintainer would clear it the same way, but happy to rebase myself if that's preferred.

PR #6084, opened 2026-09-06
fix(core): floor SegmentedControlItem to a 44px touch target on coarse pointers

### LINKED ISSUE #6013: SegmentedControl is 32px at size="lg", below the 44px touch-target floor at every size
### Summary

`SegmentedControlItem` is 32px tall at `size="lg"` — the largest size the component offers — which is below the 44px minimum touch target size in the WCAG 2.2 AA "Target Size (Minimum)" / Apple HIG / Material guidance that most mobile products hold themselves to. Because `lg` is the ceiling, no prop combination reaches the floor; a consumer has to override the height in CSS.

### Measured

`@astryxdesign/core@0.5.2`, Chromium, 390px viewport (iPhone-class width), `getBoundingClientRect().height` on `.astryx-segmented-control-item`:

| `size` | height |
| --- | --- |
| `sm` | 24px |
| `md` | 28px |
| `lg` | **32px** |

Rendered as:

```tsx
<SegmentedControl size="lg" label="…" value={v} onChange={setV}>
  <SegmentedControlItem value="a" label="Alpha" />
  <SegmentedControlItem value="b" label="Beta" />
</SegmentedControl>
```

For comparison, in the same render `Button` and `ToggleButton` take `min-height: 44px` from our own phone floor and measure 44px; the segmented control does not, because its class was not in that floor's selector list. We have added it locally — this issue is about the shipped default.

### Why it matters

A segmented control is frequently a primary phone affordance (a view switcher, a filter). At 32px it is the smallest interactive control in a typical Astryx phone surface, and it is the one users hit while walking.

### Suggested

Either raise `size="lg"` to a 44px minimum on coarse pointers, or expose the height so a theme can. A `@media (pointer: coarse)` floor inside the component would fix it for every consumer without changing desktop density — which is the part consumers cannot safely do themselves, since a blanket `min-height` on Astryx classes changes non-control density too.

ManoharPaturi:
Fixes #6013

## Summary

`SegmentedControlItem` topped out at 32px (`size="lg"` = `--size-element-lg` − 4px), below the 44px minimum touch target of WCAG 2.2 AA "Target Size (Minimum)" / Apple HIG / Material — and since `lg` is the ceiling, no prop combination could reach the floor. `Button` and `ToggleButton` already honour the floor on phone surfaces; the segmented control's class was simply not in it.

## Change

A `@media (pointer: coarse)` `min-height: 44px` floor on the item's base style — mirroring `TouchDateField`'s `monthArrow` floor (`min-height` scoped to coarse pointers, `default: null`). Every size is floored on touch; **desktop density is unchanged** (no unconditional rule), which is the part consumers cannot safely do themselves.

## Tests

New case in `SegmentedControl.test.tsx` using the repo's `injectedRules()` CSSOM pattern (mirroring `InputClearButton`): the item's classes have a coarse-pointer rule carrying `min-height: 44px`, and no unconditional min-height exists. Verified the test fails on the unfloored component and passes with the fix.

`vitest run packages/core/src/SegmentedControl` → 44 passed; core `tsc --noEmit`, eslint, `check:sync`, and `check:changesets` clean. Doc note on the `size` prop; changeset included.

### DIFF
--- a/.changeset/segmented-control-touch-floor.md
+++ b/.changeset/segmented-control-touch-floor.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Floor SegmentedControlItem to a 44px touch target on coarse pointers — `size="lg"` topped out at 32px, below the WCAG 2.2 AA "Target Size (Minimum)" / Apple HIG / Material floor every other interactive control honours; desktop density is unchanged. (#6013)
+@ManoharPaturi
--- a/packages/core/src/SegmentedControl/SegmentedControl.doc.mjs
+++ b/packages/core/src/SegmentedControl/SegmentedControl.doc.mjs
@@ -74,7 +74,8 @@ export const docs = {
     {
       name: 'size',
       type: "'sm' | 'md' | 'lg'",
-      description: 'Size variant for the control.',
+      description:
+        'Size variant for the control. On coarse pointers every size is floored to a 44px touch target (WCAG 2.2 AA "Target Size (Minimum)").',
       default: "'md'",
     },
     {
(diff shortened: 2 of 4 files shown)


ManoharPaturi:
@cixzhang @kentonquatman this one has been sitting green since early september, is there anything holding it up that i can help with? happy to add tests or adjust the approach if the touch-target floor via the coarse-pointer media query isn't the shape you want.

PR #6085, opened 2026-09-06
fix(core): scope the 16px input floor to iOS

### LINKED ISSUE #6015: Scope the 16px input floor to iOS, not every coarse pointer
**Version**: `@astryxdesign/core` 0.5.2

## What happens

Text controls carry this rule:

```css
@media (pointer: coarse) {
  … { font-size: max(1rem, var(--text-body-size)); }   /* inputs, textareas, date/time fields… */
  … { font-size: max(1rem, var(--text-label-size)); }  /* Selector / ComplexSelector triggers, panel search */
}
```

I understand the intent: iOS Safari zooms the page when a focused input is under 16px, and the `max(1rem, …)` floor prevents that. The trigger, though, is `(pointer: coarse)`, which is true on a lot more than iPhones:

- Windows laptops with a touchscreen (Chrome flips to `pointer: coarse` / `hover: none` in tablet mode or after touch input) — controls grow to 16px while the rest of the UI stays at 14px, and fields visibly don't match their labels.
- Android phones and tablets, which never zoom on focus.

So the guard costs typography on every touch device to solve a problem that exists on one platform.

## Affected components (0.5.2)

Body-size floor: `TextInput`, `NumberInput`, `TextArea`, `DateInput` (incl. native and touch fields), `DateRangeInput`, `DateTimeInput`, `TimeInput`, `FileInput`, `Typeahead`, `ChatComposerInput`, `CommandPaletteInput`.
Label-size floor: `Selector`, `ComplexSelector`, `PanelSearchInput`.

Whatever the fix, it'd be good to apply it in one place so all of these move together.

## Suggestion

Either of these would keep the iOS protection and leave everyone else alone:

1. **Scope the floor to iOS.** `-webkit-touch-callout` is implemented only by iOS WebKit, so `@supports (-webkit-touch-callout: none)` nested inside the coarse-pointer query targets exactly the browsers that zoom.
2. **Make it a token.** e.g. `--input-min-font-size` defaulting to `1rem` under coarse pointers, so a theme can set it to `0` (or its own floor) without overriding component CSS.

ManoharPaturi:
Fixes #6015

## Summary

Text controls carried `@media (pointer: coarse) { font-size: max(1rem, var(--text-body-size)) }` — the iOS no-zoom floor — for **every** coarse pointer. Only iOS Safari zooms on focus below 16px; Android and desktop-touch browsers do not, so non-iOS touch users got inflated text for no benefit. This nests the floor in `@supports (-webkit-touch-callout: none)` inside the existing coarse-pointer branch (option 1 from the issue): iOS output is byte-identical, every other platform keeps the theme's token size.

## Scope

All 19 floored sites, found by recursive scan rather than by hand: 16 body-size controls (including `RichTextEditor` in packages/richtext, which carried the identical rule), 3 label-size controls (Selector, ComplexSelector, PanelSearchInput), and the two derived `height: firstThatWorks('1lh', …)` fallbacks in `NativeDateField`/`nativePickerSegmentStyles`, which became conditional so pre-`1lh` browsers stay height-consistent on both sides of the gate.

## Tests

- New family-wide source-invariant suite `packages/core/src/inputFontFloor.test.ts`: asserts every floored control carries the iOS gate **and** (recursively) that no bare coarse-pointer floor remains anywhere in core — so a future control can't reintroduce the over-broad rule.
- A `declaredValue` read-back test in `TextInput.test.tsx`. (Source-level assertions are necessary: jsdom's CSSOM silently drops at-rule-wrapped `max(1rem, …)` declarations — verified empirically against the old pattern too, so the old code was equally untestable that way.)

Docs (`ChatComposerInput.doc.mjs`, all locales) now say "iOS" where they said "touch device". Changeset included.

## Verification

Affected suites 714 + 656 passed, dependents 183 passed, final commit pass 77; `tsc --noEmit` clean (core + richtext); eslint 0 errors on touched files; `check:sync` and `check:changesets` clean.

### DIFF
--- a/packages/core/src/inputFontFloor.test.ts
+++ b/packages/core/src/inputFontFloor.test.ts
@@ -42,6 +42,9 @@ const FLOORED_CONTROLS = [
   'TextInput/TextInput.tsx',
   'TimeInput/TimeInput.tsx',
   'Typeahead/BaseTypeahead.tsx',
+  // The rich-text package carries the same floor; its path is relative to
+  // packages/core/src.
+  '../../richtext/src/RichTextEditor.tsx',
 ];
 
 /** A font floor applied to every coarse pointer, iOS or not. */
@@ -82,10 +85,16 @@ describe('the 16px input floor', () => {
     }
   });
 
-  it('leaves no bare coarse-pointer font floor anywhere in core', () => {
+  it('leaves no bare coarse-pointer font floor anywhere in core or richtext', () => {
     // Catches a new (or renamed) control reintroducing the ungated floor,
-    // wherever it lands — the list above only knows the current family.
-    const offenders = componentSources(__dirname)
+    // wherever it lands — the list above only knows the current family. The
+    // rich-text package carries the same floor, so it is scanned too.
+    const scanRoots = [
+      __dirname,
+      path.resolve(__dirname, '../../richtext/src'),
+    ];
+    const offenders = scanRoots
+      .flatMap(root => componentSources(root))
       .map(file => ({
         file,
         source: readFileSync(file, 'utf8'),
--- a/packages/core/src/inputFontFloor.test.ts
+++ b/packages/core/src/inputFontFloor.test.ts
@@ -1,0 +1,97 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+/**
+ * @file inputFontFloor.test.ts
+ * @input Uses node:fs and node:path to read the text-control sources
+ * @output Invariants for the 16px input floor shared by every text control
+ * @position Family-wide source test, in the source-reading style of
+ *   reset.test.ts. jsdom's CSSOM drops declarations nested inside @supports,
+ *   so the iOS branch cannot be asserted against injected rules — the
+ *   runtime half (no bare coarse-pointer floor) lives in TextInput.test.tsx.
+ */
+
+import {readFileSync, readdirSync} from 'node:fs';
+import path from 'node:path';
+
+import {describe, expect, it} from 'vitest';
+
+/**
+ * Every control that floors its text to 16px. iOS Safari zooms the page when
+ * a focused control renders under 16px; no other coarse-pointer browser does.
+ * `-webkit-touch-callout` is implemented by iOS WebKit alone, so each floor
+ * nests `@supports (-webkit-touch-callout: none)` inside the coarse-pointer
+ * query: iPhones and iPads keep the zoom guard, while Android and touch
+ * laptops keep the theme's own type scale instead of an inflated 16px.
+ */
+const FLOORED_CONTROLS = [
+  'Chat/ChatComposerInput.tsx',
+  'CommandPalette/CommandPaletteInput.tsx',
+  'ComplexSelector/ComplexSelector.tsx',
+  'DateInput/DateInput.tsx',
+  'DateInput/NativeDateField.tsx',
+  'DateInput/TouchDateField.tsx',
+  'DateRangeInput/DateRangeInput.tsx',
+  'DateTimeInput/DateTimeInput.tsx',
+  'DateTimeInput/TouchDateTimeField.tsx',
+  'DateTimeInput/nativePickerSegmentStyles.ts',
+  'Field/PanelSearchInput.tsx',
+  'FileInput/FileInput.tsx',
+  'NumberInput/NumberInput.tsx',
+  'Selector/Selector.tsx',
+  'TextArea/TextArea.tsx',
+  'TextInput/TextInput.tsx',
+  'TimeInput/TimeInput.tsx',
+  'Typeahead/BaseTypeahead.tsx',
+];
+
+/** A font floor applied to every coarse pointer, iOS or not. */
+const BARE_COARSE_FLOOR = `'@media (pointer: coarse)': \`max(1rem`;
+
+/** The same floor, gated to iOS WebKit inside the coarse-pointer query. */
+const IOS_GATED_FLOOR =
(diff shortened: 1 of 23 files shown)


packages/core/src/inputFontFloor.test.ts
```diff
@@ -0,0 +1,97 @@
...
+  'DateTimeInput/TouchDateTimeField.tsx',
+  'DateTimeInput/nativePickerSegmentStyles.ts',
+  'Field/PanelSearchInput.tsx',
+  'FileInput/FileInput.tsx',
+  'NumberInput/NumberInput.tsx',
+  'Selector/Selector.tsx',
+  'TextArea/TextArea.tsx',
+  'TextInput/TextInput.tsx',
+  'TimeInput/TimeInput.tsx',
+  'Typeahead/BaseTypeahead.tsx',
```
cixzhang, line 44: This omits RichTextEditor, so its broad coarse-pointer floor can regress while this ‘family-wide’ suite stays green.

[Reviewed by Robohands]

cixzhang (changes requested):
Thanks for tightening the floor to iOS. The Chromium path and generated CSS look right, but the family regression guard omits RichTextEditor. Please include the rich-text package so all 19 changed controls are protected.

[Reviewed by Robohands]

ManoharPaturi:
Done in the updated branch — `inputFontFloor.test.ts` now scans the rich-text package too: `RichTextEditor.tsx` is in the gated-controls list (path relative to core), and the bare-floor offender scan covers `packages/richtext/src` alongside core, so a reintroduced ungated floor fails wherever it lands in either tree. Both guard tests green.

cixzhang (changes requested):
Thanks for fixing the RichTextEditor coverage gap—the family guard and non-iOS path now check out. I still can’t verify the iOS half of this compatibility fix. Please add latest-stable iOS Safari evidence showing the nested gate computes the input to 16px and focusing it keeps `visualViewport.scale` at 1, with the device/OS/Safari version and this exact head recorded. Chromium or Playwright WebKit cannot establish that browser-specific behavior.

[Reviewed by Robohands]

ManoharPaturi:
Real-device evidence, as requested — captured on an **iPhone running current Safari 27.0**, at head **`d2b04b4`** (`fix/input-floor-ios-only`).

I put up a self-contained harness that embeds the exact nested gate from this PR's head, so the numbers below are recorded on-device against this change: https://manoharpaturi.github.io/astryx/ios-floor-evidence.html

**Environment (recorded on-device):**

- userAgent: `Mozilla/5.0 (iPhone; CPU iPhone OS 18_7 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/27.0 Mobile/15E148 Safari/604.1` (Safari 27.0; the OS token in Safari's UA is a frozen compatibility string — `Version/27.0` is the actual browser version)
- `maxTouchPoints` 5 · `(pointer: coarse)` matches · `CSS.supports('-webkit-touch-callout', 'none')` true · initial `visualViewport.scale` 1

**Field A — gated by the fix** (`font-size: var(--text-body-size)` + the nested `@media (pointer: coarse) { @supports (-webkit-touch-callout: none) { font-size: max(1rem, var(--text-body-size)) } }` from this head):

- computed font-size: **16px**
- min `visualViewport.scale` while focused: **1** — focusing does not zoom

**Field B — negative control** (raw `font-size: 14px`, no gate):

- computed font-size: 14px
- min scale while focused: **1.142** — iOS auto-zooms on focus, confirming the harness does detect the failure mode on this device

iOS Safari evidence — Field A floors to 16px with no zoom; the un-gated 14px control zooms to 1.142x

The harness prints `astryx PR #6085 · head d2b04b4` in its header, so the recording is bound to this exact head.

I hope this help you to verify.

kentonquatman (approved):
Thanks @ManoharPaturi! The change looks good. Would it be possible to see some before/after screenshots from iOS and non-iOS devices for this PR, just to verify?

ManoharPaturi:
sure. the iOS device evidence is in my earlier comment (iPhone, Safari 27.0, gated input computes 16px and keeps visualViewport.scale at 1 while the ungated 14px control zooms to 1.142x): https://github.com/facebook/astryx/pull/6085#issuecomment-5558237168

for non-iOS here is before/after with the exact CSS of main vs this PR side by side:

android-like coarse pointer (Chromium emulation, no -webkit-touch-callout):
non-iOS coarse pointer, before vs after
before: floor applies, input jumps to 16px. after: gate does not match, input stays at the 14px design size. that is the behavior change this PR makes on non-iOS coarse devices.

desktop fine pointer (no change either way, 14px before and after):
desktop fine pointer

each panel prints its own computed font-size so the numbers are in the screenshot, and the env line records pointer type and the supports result.

ManoharPaturi:
Rebased onto latest upstream `main` to cleanly resolve the merge conflict in `TextInput.test.tsx`. All unit tests, RTL shards, and visual regressions are passing.

ManoharPaturi:
@cixzhang @kentonquatman the branch is rebased on latest main with everything green, kentonquatman's approval and the device evidence are current, but the earlier changes-requested review is still blocking the merge button. could one of you dismiss it if there's nothing else outstanding?

ManoharPaturi:
@cixzhang @kentonquatman gentle follow up on this one. the approval, the device evidence, and the rebase are all current and the checks are green, the only thing blocking the merge button is the earlier changes-requested review. could one of you dismiss it if nothing else is outstanding?

cixzhang (approved):
Thanks I think it's reasonable to scope this to iOS for now. We are also working on making  mobile responsive themes which will have 16px body text generally to reduce some of the inconsistency.

PR #6086, opened 2026-09-06
feat(core): scale TextInput/TextArea control type with size; add text-input-control theme target

### LINKED ISSUE #6014: TextInput: text size ignores `size`; add a `text-input-control` target
**Version**: `@astryxdesign/core` 0.5.2, `@astryxdesign/theme-neutral` 0.5.2

## What I'm seeing

`size="sm" | "md" | "lg"` on `TextInput` changes the height of the control (24 / 32 / 40px via `--size-element-*`) and nothing else. The typed text and the placeholder stay at `--text-body-size` (14px in neutral) in all three. A `sm` input is therefore body-size text squeezed into a 24px box, and an `lg` input is the same 14px text floating in a 40px box.

`TextArea` behaves the same way: `sm` and `md` add no styles to the control at all, `lg` only adds vertical padding.

Quick repro:

```tsx
<TextInput size="sm" label="Search" placeholder="Search vehicles…" />
<TextInput size="md" label="Search" placeholder="Search vehicles…" />
<TextInput size="lg" label="Search" placeholder="Search vehicles…" />
```

Inspect the three `<input>` elements: computed `font-size` is identical.

## What I'd expect

Two options, in order of preference:

1. **Let typography follow `size`.** Something like `sm → --font-size-sm`, `md → --font-size-base`, `lg → --font-size-lg`, so a compact toolbar search and a hero form field don't share the exact same type. The field label already adapts to the size, so the control is the odd one out.

2. **If keeping the type fixed is intentional, expose a `text-input-control` theme target** on the inner `<input>`, the way 0.5.1 added `text-area-control` for `TextArea`. Right now the only reachable targets are `text-input` (the wrapper), `field` and `field-label`. Setting a font size on the wrapper does nothing because the control sets its own, so a theme has no sanctioned way to adjust this per size. With the target we could decide the scale in our own theme and leave the default alone.

ManoharPaturi:
Fixes #6014

## Summary

Both asks from the issue:

1. **Control text now scales with `size`** for `TextInput` and `TextArea`: `sm` → `--text-supporting-size/leading`, `md` → body, `lg` → `--text-large-size/leading` (the repo's `typeScaleVars`; these wrap the raw `--font-font-size-*` tokens the issue measured). Previously only the control height changed — a `sm` input squeezed body text into 24px and an `lg` floated 14px text in 40px. The coarse-pointer `max(1rem, …)` floor is kept per size; `lg` TextArea keeps its padding step.
2. **New `text-input-control` theme target** on the `<input>` element — a bare target exactly mirroring the existing `text-area-control` — so themes can key off the control surface; `data-size` stays on the `astryx-text-input` wrapper.

## Tests

+6 in `TextInput.test.tsx` (control target rendered, wrapper-only `data-size`, per-size `toHaveStyle` var assertions, default) and +3 in `TextArea.test.tsx`, following existing style.

## Note for the reviewer

`packages/themes/probe/src/probeTheme.ts` is regenerated (`visual:probe-theme`) — it picks up this PR's new target **plus pre-existing drift on `origin/main`** (stepper targets, the merged toggle-button elevation selectors); main currently fails the probe check without a regen, so this brings it back in sync.

## Verification

Targeted suites 576 passed (TextInput 78, TextArea 86, theming-targets guard 412); `tsc --noEmit` clean; eslint 0 errors; probe check current (277 targets / 892 selectors); all `check:repo` scripts pass. Docs updated in EN/zh/dense; changeset included.

### DIFF
--- a/.changeset/text-input-control-type-scale.md
+++ b/.changeset/text-input-control-type-scale.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] TextInput/TextArea: the control's type now follows its size (sm → supporting, md → body, lg → large), and TextInput's inner `<input>` renders as its own theme target, `astryx-text-input-control`, mirroring `text-area-control` for TextArea. (#6014) Before, `size` changed only the control height — typed text and placeholder stayed at `--text-body-size` in all three sizes, so a `sm` input was body text squeezed into a 28px box and an `lg` input the same 14px text floating in a 36px one. Both of the issue's asks are covered: the default scale adapts, and a theme that wants a different scale can key off the new control target (the size stays reflected as `data-size` on the `astryx-text-input` wrapper, so per-size control styling stays reachable without a structural selector). The coarse-pointer `max(1rem, …)` minimum is kept at every size so touch devices never get sub-16px text — and never zoom on focus. TextArea gets the same type scale; its `lg` padding step is unchanged.
+
+@ManoharPaturi
--- a/packages/core/src/TextArea/TextArea.doc.mjs
+++ b/packages/core/src/TextArea/TextArea.doc.mjs
@@ -225,7 +225,7 @@ export const docs = {
       name: 'size',
       type: "'sm' | 'md' | 'lg'",
       description:
-        'Size of the textarea, affecting internal padding. Height is controlled by rows, not size.',
+        'Size of the textarea: scales the type (sm supporting, md body, lg large) and lg\u2019s internal padding. Height is controlled by rows, not size.',
       default: "'md'",
     },
     {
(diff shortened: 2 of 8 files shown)


kentonquatman:
Hey @ManoharPaturi! The current behavior of the `size` prop on `TextInput` and `TextArea` are intentional. The expected behavior is that only the the height will change when setting the sizes of `sm` `md` or `lg`.

For more context, see the comment from @cixzhang here: https://github.com/facebook/astryx/issues/6014#issuecomment-5543171994

kentonquatman (changes requested):
See comments from @cixzhang on the associated issue: https://github.com/facebook/astryx/issues/6014#issuecomment-5543171994

ManoharPaturi:
@kentonquatman got it, thanks for the pointer. i'll rework this along the spec path (anatomy entry + theming disposition for the native control first, then the target on the input) rather than inferring the name. size stays dimensional.

ManoharPaturi:
Closing this out in favor of the spec-first path — as noted by @kentonquatman and @cixzhang in https://github.com/facebook/astryx/issues/6014#issuecomment-5543171994, `size` remains intentionally dimensional and the native input control anatomy/theming disposition requires formal spec owner definition. I'll defer the spec and theme target addition to the maintainers/spec owners. Thanks for the review and guidance!

PR #6088, opened 2026-09-06
docs(Stepper): describe exported context accurately

cixzhang:
## Summary

Removes misleading comments that described members of the package-exported `StepperContextValue` as internal or outside the public API.

This is comment-only: no runtime behavior, types, exports, specifications, or release notes change.

## Validation

- `pnpm exec prettier --check packages/core/src/Stepper/StepperContext.ts`
- `git diff --check`
- pre-commit `pnpm check:repo`


cixzhang (commented):
Semantic verdict: approve

Builders no longer see `StepperContextValue` fields described as internal. The subpath and runtime hook already expose them, with no type or behavior change.

[Reviewed by Robohands]

PR #6089, opened 2026-09-06
docs: define React update propagation contract

cixzhang:
## Summary

- defines shared React update-propagation boundaries for pre-paint upward writes, membership, mutable metadata, context fan-out, and render isolation
- projects the shared boundary into Stepper through lifecycle-stable membership and focused propagation evidence
- constrains `horizontalOptions.minimumStepWidth` to finite non-negative pixel numbers; CSS strings, units, `calc()`, and custom properties are unsupported
- keeps exported `StepperContextValue` / `useStepperContext` unchanged from v0.5.2 and moves collapsed-layout coordination to a separate private, non-exported context
- admits private `isCollapsed` and a parent-owned Collapsed summary render capability while rejecting raw DOM `summarySlot` exchange
- keeps Frame as named structural anatomy with an intentional `none` theming disposition and removes `astryx-stepper-frame` from published target metadata
- records the current hidden threshold probe, added exported context fields, and emitted Frame selector as implementation adoption gaps rather than public contract
- replaces old review-status prose with durable compatibility requirements

## Governance

The records define reusable authority. They do not approve, classify, designate, or authorize a specific pull request. Historical pull requests may appear only as evidence.

## Verification

- `pnpm --filter @astryxdesign/core typecheck:docs`
- generated `astryx component Stepper --dense` omits Frame target and shows number-only `minimumStepWidth`
- Prettier
- `git diff --check`

No runtime implementation change. No changeset.


cixzhang:
Closing under the standing no-PR rule.

PR #6090, opened 2026-09-06

### LINKED ISSUE #6021: Link's inline-flex anchor silently defeats an ancestor Text maxLines clamp

parinith-web:
This PR is the same fix as #6038 (opened by @ManoharPaturi), plus one
additional commit addressing the latest review comment on that thread.
Opening this directly against `main` at @ManoharPaturi's suggestion /
per our discussion.

Fixes #6021

## Summary

A plain `<Text maxLines={2}><Link>…</Link></Text>` silently did not truncate:
the anchor was `display: inline-flex`, and an inline-flex box establishes its
own formatting context that an ancestor `-webkit-line-clamp` cannot reach.

This changes the plain anchor to `display: inline` so it participates in the
surrounding line boxes and an ancestor clamp truncates it normally. The flex
layout is preserved wherever it's load-bearing:

- external links (`isExternalLink` + anchor form) stay `inline-flex`
- the button-rendered form stays `inline-flex`
- `display="block"` and self-clamping (`maxLines > 0`) Links stay `inline-flex`
  so their keyboard focus outline still paints (an inline anchor around a
  block-level child computes a focus outline but paints nothing in Chromium)
- Links composing a block-level child directly (e.g. an HStack + Icon) can
  now opt in via a new `hasBlockChild` prop, for the same focus-visibility
  reason — added as an explicit prop rather than introspecting `children`,
  since this repo's `no-react-introspection` ESLint rule disallows
  `Children.toArray`/`isValidElement` for that kind of parent-child coupling

Docs updated in all three variants (EN/zh/dense) for `Link` and `Text`
covering when ancestor-clamp truncation will/won't reach a Link.

## Tests

- plain anchor → `inline`, not `inline-flex`
- external-link anchor → `inline-flex`
- button-rendered form → `inline-flex`
- `display="block"` → `inline-flex`
- self-clamping (`maxLines > 0`) → `inline-flex`
- `hasBlockChild` (HStack + Icon composition) → `inline-flex`

## Testing

- `vitest run packages/core/src/Link packages/core/src/Text` — 256 passed
- `tsc --noEmit` (core) clean
- `eslint` clean
- `check:sync` clean
- `check:changesets` clean

Co-authored-by: ManoharPaturi <186662190+ManoharPaturi@users.noreply.github.com>

### DIFF
--- a/.changeset/link-anchor-clamp.md
+++ b/.changeset/link-anchor-clamp.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Render a plain Link anchor as `inline` so an ancestor `<Text maxLines>` clamp can truncate it; flex layout is kept for the external-link icon and button forms, and Link/Text docs now cover the external-link clamp limitation. (#6021)
+@ManoharPaturi
--- a/packages/core/src/Link/Link.doc.mjs
+++ b/packages/core/src/Link/Link.doc.mjs
@@ -169,10 +169,12 @@ export const docs = {
     bestPractices: [
       { guidance: true, description: 'Write descriptive, concise link text that clearly communicates the destination.' },
       { guidance: true, description: 'Set `isStandalone` when the link appears outside of inline text, so it receives proper base font sizing.' },
+      { guidance: true, description: 'Let a plain Link be truncated by an ancestor <Text maxLines>; the anchor participates in the surrounding line boxes.' },
       { guidance: true, description: 'Only set `label` when the link content is not descriptive text (e.g. an icon-only link). For text links, the visible text is already the accessible name; adding `label` overrides it for screen readers, which is harmful.' },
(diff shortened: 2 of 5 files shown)


cixzhang (changes requested):
Keyboard users of composed Links still lose their focus indicator unless every caller remembers this new prop. In Chromium, `<Link><HStack … /></Link>` computes a 2px outline but paints no focus box; adding `hasBlockChild` is the only thing that restores it. Please remove `hasBlockChild` and its docs/tests, keep this PR to the plain-link clamp plus the existing Link-owned block/self-clamp/external/button cases, and handle broader child display internally rather than making callers describe DOM layout.

[Reviewed by Robohands]

kentonquatman (approved):
Looks great from the design side. Thanks @parinith-web!

Image

Issue #5923, opened 2026-09-02
[Infra] The Figma library's six theme collections declare theme modes but hold no variables


joaodotwork:
## Problem

`@astryxdesign/core` ships seven themes, each with a generated, `@scope`d stylesheet
(`@astryxdesign/theme-chocolate/theme.css` and so on). In the **Figma** library, those
themes have no values at all.

The library has 24 variable collections. Six of them own every variable; the other 18
own none. The 18 are named after the six shipped themes — Chocolate, Matcha, Gothic,
Butter, Y2K, Stone — one per themeable dimension (Color, Typography, Radius). Each
declares the right modes (`Chocolate Light`, `Chocolate Dark`, …) and holds nothing
in them.

So a designer can select a theme mode in Figma and see no change, and there is no
path by which the code's themes reach the design tool.

## Evidence

Read directly in `Astryx Library (Community)` (`cYTWQpjWxbR30zwQgqlSmY`) through the
Figma plugin API, read-only, on 2026-09-02:

```
getLocalVariablesAsync()            -> 166 distinct variables
getLocalVariableCollectionsAsync()  -> 24 collections
sum of collection.variableIds       -> 1048
```

Grouping every variable by its own `variableCollectionId`:

| collection | modes | variables owned |
|---|---|---|
| Color | `Neutral Light`, `Neutral Dark` | 108 |
| Typography | `Neutral` | 33 |
| Spacing | `Default` | 15 |
| Radius | `Neutral` | 6 |
| Size | `Default` | 3 |
| Border | `Default` | 1 |
| **18 theme collections** | `<Theme> Light` / `<Theme> Dark`, or `<Theme>` | **0 each** |

Two details worth stating precisely:

- The 1048 figure is **double counting**, not a larger set. The empty theme
  collections list the base variables by id, so summing `variableIds` counts the same
  166 seven times over. We initially misread this as "themes are modes with values";
  it is not.
- Nothing is unpublished. `hiddenFromPublishing` is `false` on every collection and
  every variable, so this is genuinely empty rather than staged-but-not-shared. From
  a subscribing file, `getVariablesInLibraryCollectionAsync` returns zero for all 18.

What **does** work: the base `Color` collection carries both `Neutral Light` and
`Neutral Dark`, and all 108 colour variables have a value in each. Light/dark
round-trips cleanly. It is the six named themes that do not.

## Why this may be intentional, and why it still seems worth raising

[...]

Issue #5922, opened 2026-09-02
[RFC] Publish the Figma-variable to CSS-custom-property mapping, and a checker for it


joaodotwork:
## The gap

The Night Watch wiki documents how TypeScript constructs map to Figma — union props
to variant axes, booleans to component properties, ReactNode to slots — but not how
**tokens** map. It says the token workflow lives in private automation.

That mapping is not derivable by rule. We measured it across all 166 variables in the
published library and it takes **five different name shapes** to match them. Anyone
building on Astryx who wants to drive tokens from Figma — which is the common case
for an agency applying it to a client brand — has to rediscover this by trial.

## What the mapping actually is

Figma and the CSS agree on the token *set* far more than on the token *names*. Four
systematic divergences:

| divergence | Figma | CSS custom property |
|---|---|---|
| role/hue order inverted | `Gray/Background` | `--color-background-gray` |
| group segment dropped | `Status/Success` | `--color-success` |
| synonym | `Surface/Card` | `--color-background-card` |
| abbreviation | `Element/Small` | `--size-element-sm` |
| word vs decimal | `Spacing/half` | `--spacing-0-5` |

Straight-through cases exist too — `Radius/Element` is `--radius-element`, and the
whole Radius and Spacing collections map one-to-one once the two fractional spacing
steps are aliased.

Measured coverage, Figma variable → a custom property `astryx.css` actually reads:

```
166 Figma variables
135 reach a code token   (81%)
    110 themeable — a theme or a consumer can override them
     25 global constants declared in astryx.css
 31 reach nothing
```

The 31 fall into three explainable groups, none of them a naming problem:

- **14 `Line Height/*`** — the CSS has no `--line-height-*` tokens; line-height is
  compiled into the atomic classes. In Figma these are bound to text styles (14 of 14
  local text styles bind `lineHeight` to a variable), so each side has a mechanism and
  they simply differ. Worth documenting so it does not read as a gap.
- **14 `Component/*`** — Badge, Input Ring, ProgressBar internals, compiled into the
  components rather than exposed as variables.
- **3 semantic misses** — `Status/Error Inverted`, `Effects/Elevation Shadow`,
  `Effects/Elevation Shadow Strong`. The code has `--shadow-low/medium/high`, related
  in meaning but not in name.

The reverse gap is larger and worth stating: **118 of the 228 tokens a theme declares
have no Figma name at all**, dominated by the 56 `--color-data-*` chart colours, plus
[...]

PR #6092, opened 2026-09-07
theme build: add opt-in --tokens JSON dump (#5923)

parinith-web:
## What
Adds an opt-in `--tokens` flag to `astryx theme build` that additionally writes `<name>.tokens.json` — a flat `{cssCustomPropertyName: value}` dump of the theme's fully resolved tokens (`tokens` + `localTokens`), keyed exactly as they appear in the generated CSS (`--color-accent`, not a bare token name).

This is the CLI-side half of the direction proposed in #5923: the build already resolves every token before generating CSS, so this reuses that same `resolvedTheme` object rather than re-resolving anything — the dump can never drift from the CSS it describes.

## Why
#5923 found that the Astryx Figma library ships 18 empty theme-mode collections with no code-to-Figma path to populate them. #5922 already has a verified Figma-variable-name ↔ CSS-custom-property mapping across all 166 base variables. This PR provides the missing piece on the CLI side: a machine-readable source of resolved values that a Figma-side sync script ("the Librarian", per #5923) could read against that mapping to actually populate the theme modes.

**This PR does not touch Figma at all** — it's scoped to the CLI output only. Writing values into Figma variable modes is separate follow-up work.

## Why opt-in / off by default
`--tokens` defaults to `false` so existing builds and `--check` baselines are completely unaffected — no new file appears for any project unless they ask for it.

## Details
- `tokensJsonPath`/`tokensJsonContent` slot into the existing atomic stage-then-rename write pipeline as a fifth output alongside css/js/dts.
- Deliberately excludes the invocation string that the CSS/JS headers' `Command:` line carries (which `--check` strips before comparing) — JSON has no comment syntax, so there's no field to normalize away, and the file is compared byte-for-byte under `--check`.
- Wired through the CLI flag parser, including the watch-mode child process spawn, plus updated `--help`/API docs and the `ThemeBuildResponse` typedef.

## Testing
New `build.tokens-json.test.mjs` (5 tests): default-off behavior, JSON shape/values (including `[light, dark]` tuple → `light-dark()` resolution), `localTokens` inclusion, `--check` staleness detection, and no spurious drift across identical rebuilds.

All 94 pre-existing tests in `packages/cli/api/theme/**` still pass unmodified.

Refs #5923, #5922.

### DIFF
--- a/packages/cli/.gitignore
+++ b/packages/cli/.gitignore
@@ -17,3 +17,9 @@ foundation/*.d.mts
 # Transient fixture dir written by authoring-migration.tsc.test.mjs (runs the
 # real tsc over migrated output); created + removed per test run.
 .codemod-tsc-*/
+
+# Transient fixture dirs for tests whose theme files need to
+# `import {defineTheme} from '@astryxdesign/core/theme'` and therefore have
+# to live under a real node_modules tree; created + removed per test run.
+.tmp-extends-*/
+.tmp-tokens-*/
--- a/packages/cli/api/theme/build/build.mjs
+++ b/packages/cli/api/theme/build/build.mjs
@@ -1039,11 +1039,14 @@ function validatePrivateVars(themeDef) {
  * `logger` (silent by default).
  *
  * @param {string} file - Theme file path, resolved against `cwd`.
- * @param {{out?: string, check?: boolean, iconsSpecifier?: string}} [options] -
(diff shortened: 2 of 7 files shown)


cixzhang (changes requested):
Thanks for keeping this opt-in and deterministic. Agents would still get a stable new flag and response field before Astryx has a supported reader or complete schema. Please land this with the reader, add `tokensJson` to the generated response docs plus a CLI `--json` contract test, and add the CLI patch changeset.

[Reviewed by Robohands]

PR #6094, opened 2026-09-07
docs: require authority for every public delta

cixzhang:
## Summary

- requires every public API update and public behavior change to match current committed authority before acceptance
- routes exact matches to normal correctness review, contradictions to request-changes, and unsettled public contracts to private human hold
- makes package reachability—not source labels—the public boundary for supporting types, context members, hook returns, utility signatures, defaults, and behavior
- treats generated declaration/export/behavior manifests as evidence feeding the authority decision, never as authority themselves
- preserves Astryx's preference for composable public primitives while requiring them to express intentional semantic contracts rather than internal focus, gesture, DOM, or timing machinery
- prevents bug fixes from smuggling public API or behavior changes past specification review
- states that specifications define reusable contracts and never assign a verdict to a specific pull request

## Verification

- Prettier
- `git diff --check`

No runtime implementation changes. No changeset.

- converts named implementation PRs in the current input-field family record to evidence-only references, so durable specs do not assign individual PR verdicts


cixzhang:

One contradiction blocks this. The canonical review path lets an API change that matches an existing current contract proceed, but the API spec and contributor guide require another owner decision whenever a bug fix needs a new primitive or observable behavior. That would make an already-approved contract look undecided and delay the contributor. Please defer both copies to the knowledge contract and hold only when no current contract settles the delta.

[Reviewed by Robohands]

cixzhang:
Resolved the bug-fix routing contradiction: any additional public delta now follows the canonical knowledge-contract result independently. Existing current authority may settle it; only absent authority uses the owner hold.

Issue #5701, opened 2026-08-30
[Bug] MobileNav: closing the drawer skips the slide out animation


mmahdium:
### Description

When closing MobileNav on mobile, the drawer disappears instantly instead of sliding out. Opening works fine and animates smoothly, but closing has no visible transition at all.

Would be good to get eyes on this since MobileNav's animation is being referenced as the model pattern in #4174.  

### Reproduction

1. Open the mobile nav drawer (opening is animated correctly)  
2. Close it via backdrop click, close button, or Escape  
3. Drawer vanishes immediately, no animation

[...]

PR #6096, opened 2026-09-07
docs: define visibility transition lifecycle

cixzhang:
## Summary

- defines the lifecycle order for child-owned exits and ancestor visibility/deactivation
- distinguishes paint suppression from interaction suppression: an exiting surface may become inert without becoming invisible
- prevents Activity/Offscreen hidden mode, conditional unmounting, `hidden`, framework `display: none !important`, or equivalent paint suppression from preempting an owned exit
- requires one completion owner, reduced-motion parity, and stale-close/rapid-reopen safety
- projects the shared rule into MobileNav: close intent → MobileNav exit completion → AppShell deactivation

## Trigger

Issue #5701 exposed AppShell hiding MobileNav through React Activity before MobileNav's slide-out could paint. The inline `!important` declaration is evidence of lifecycle preemption, not a blanket CSS prohibition.

## Verification

- Prettier
- `git diff --check`

No runtime implementation changes. No changeset.

### DIFF
--- a/docs/architecture/visibility-transition-lifecycle.md
+++ b/docs/architecture/visibility-transition-lifecycle.md
@@ -1,0 +1,130 @@
+---
+schema_version: 1
+template_version: 1
+kind: architecture
+id: architecture:visibility-transition-lifecycle
+authority: current
+archive_reason: null
+superseded_by: null
+approved_by: cixzhang
+approved_at: 2026-09-06
+owners: [cixzhang]
+applies_to:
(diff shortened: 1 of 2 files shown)


cixzhang:
Closing because this over-generalizes a concrete MobileNav/AppShell animation bug into shared architecture. Issue #5701 remains the implementation bug to fix. Inline `!important` is a review smell here, not a blanket specification violation.

PR #6097, opened 2026-09-07
fix(Stepper): simplify collapse and context APIs

imdreamrunner:
## Summary

- Restrict `Stepper`'s `horizontalOptions.minimumStepWidth` to numeric pixel values.
- Remove the CSS-length measurement element and its resize observer.
- Change `registerStep` from a disabled boolean to an optional `{getIsDisabled}` options object, with `Step` supplying the getter.
- Re-resolve compact navigation targets at activation so a getter change cannot select a newly disabled step from a stale render.
- Use separate public and internal contexts: `useStepperContext()` receives exactly the supported runtime keys, while the built-in `Step` receives compact-layout coordination through `useStepperInternalContext()`.
- Keep `previousActiveStep` and `registerStep` on the public context while removing `stepCount`, `isCompact`, `summarySlot`, `minimumStepWidth`, and `minStepWidthMeasureRef` from `StepperContextValue` and `useStepperContext` at both type and runtime boundaries.

## Public API

- **Owner:** `component:Stepper` (`authority: current`)
- **Before → after:** collapse thresholds accepted numbers or CSS-length strings and the public context exposed compact-layout implementation fields; thresholds now accept pixel numbers only, registration reads disabled state through an optional getter, and the five named implementation fields are private or removed at type and runtime boundaries.
- **Classification:** breaking API simplification relative to v0.5.3.
- **Migration:** replace CSS lengths such as `'7rem'` with their intended pixel number; replace `registerStep(index, isDisabled)` with `registerStep(index, {getIsDisabled})` or omit the options; stop reading the removed context fields.

## Testing

- `pnpm exec vitest run packages/core/src/Stepper/Stepper.test.tsx packages/core/src/Stepper/Stepper.public.test.ts` — 95 tests passed
- Runtime-key assertions cover horizontal and vertical public providers
- `pnpm lint:strict` — 0 errors (repository baseline warnings only)
- `pnpm check:knowledge`
- `pnpm check:changesets`

### DIFF
--- a/.changeset/stepper-pixel-threshold.md
+++ b/.changeset/stepper-pixel-threshold.md
@@ -1,0 +1,9 @@
+---
+'@astryxdesign/core': minor
+---
+
+[breaking] Restrict `Stepper`'s `horizontalOptions.minimumStepWidth` to a pixel number and remove compact-layout implementation fields from `useStepperContext`.
+
+Replace CSS-length thresholds such as `'7rem'` with their intended pixel number. Call `registerStep(index, {getIsDisabled})` instead of passing a disabled boolean; the options object is optional. `StepperContextValue` keeps transition history and step registration, while step count, compact state, summary-portal coordination, and threshold measurement remain package-internal.
+
+@imdreamrunner
--- a/apps/storybook/stories/Stepper.stories.tsx
+++ b/apps/storybook/stories/Stepper.stories.tsx
@@ -1123,7 +1123,7 @@ export const NarrowCollapse: Story = {
     docs: {
       description: {
         story:
-          "A horizontal stepper measures its own width and collapses once a step has less than `horizontalOptions.minimumStepWidth` (112px by default), so the breakpoint follows the step count rather than the viewport. Pass a number for pixels or a CSS length string such as `7rem`. `horizontalOptions.collapsedVariant` selects `'withLabelAndControls'`, `'withLabel'`, or `'hiddenLabel'`; the last renders only the bare progress track with no compact row. The two indicator positions collapse differently on purpose: `separated` drops its indicators with the labels and repeats the active indicator beside the compact label, while `on-track` keeps its indicators as presentational nodes on the rail and does not repeat the active one beside the label. Controls appear only for `withLabelAndControls` when `onStepClick` is set.",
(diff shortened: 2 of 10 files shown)


cixzhang (changes requested):
Custom descendants using the new registration API can still send users to a disabled step. After `getIsDisabled()` changes, Next keeps the target captured during the Stepper’s last render: the probe selected step 2, then selected step 3 only after an unrelated parent render. Please make compact activation use the current registration state without requiring that rerender.

Also, the NarrowCollapse story still calls its numeric `64` example “4rem.”

[Reviewed by Robohands]

cixzhang (changes requested):
JavaScript consumers still receive `stepCount`, `isCompact`, and `summarySlot` from `useStepperContext()`, because the public hook returns the same runtime object as the internal hook. That leaves compact-layout implementation reachable despite the new contract and makes the cleanup incomplete. Please project a genuinely public value from `useStepperContext()` and add a runtime key assertion; keep `useStepperInternalContext()` for `Step`.

The stale registration target and “4rem” story label from the prior review are fixed.

[Reviewed by Robohands]

imdreamrunner:
Thanks for the follow-up. The two findings from the prior review are addressed on `ac85d65be5c76aabeae8e3ca7fb7fde77935519d`:

- Compact Previous/Next now re-runs `adjacentEnabledStep(delta)` inside activation, so it reads the current `getIsDisabled()` values rather than navigating with a render-time target. The regression mutates a registered getter without a React rerender and verifies Next skips step 2 for step 3.
- The NarrowCollapse story now labels the numeric example as `64px`, not `4rem`.

I also see the new runtime-surface finding: `useStepperContext()` is typed as the public subset but currently returns the wider provider object at runtime. I am pausing before changing that API behavior; the proposed follow-up is to project an exact public object in `useStepperContext()` and add a runtime key assertion, while leaving `useStepperInternalContext()` unchanged for the built-in `Step`.

imdreamrunner:
Addressed the runtime-surface follow-up in `67b7cb7d26367fdab113b98091ed5348cf919b45`.

- `StepperContext` now carries an exact `StepperContextValue` with only the eight supported public keys.
- A separate `StepperInternalContext` carries `stepCount`, `isCompact`, and `summarySlot`; only `useStepperInternalContext()` (used by the built-in `Step`) reads it.
- `useStepperContext()` now reads only the public provider, so JavaScript can no longer access the removed compact-layout fields.
- Added an `Object.keys(useStepperContext())` runtime assertion for both horizontal and vertical provider paths.

The prior findings remain fixed: activation re-reads `getIsDisabled()` without a rerender, and the story says `64px`.

Validation: 95 focused tests pass; core typecheck, strict lint, core build, knowledge, and changeset checks pass. Independent focused review found no findings. Ready for Robohands re-review.

imdreamrunner:
All checks on exact head `67b7cb7d26367fdab113b98091ed5348cf919b45` are now green, including UI/node tests, build, lint, a11y, RTL, visual regression, and exact-head spec approval. The runtime key assertion passes in both horizontal and vertical orientations. @cixzhang re-review requested to clear the two superseded change requests.

cixzhang (approved):
Thanks, this resolves our stale registration, story label, and public runtime-context asks. The 447/448px boundary and focused tests are clean.

[Reviewed by Robohands]

PR #6032, opened 2026-09-04
docs: specify component audit backfills

cixzhang:
## Summary

- define the additional lifecycle needed to run whole-component audits without duplicating the existing component-specification system
- reuse the current component template/schema, knowledge-contract architecture, AST-002, and component records
- define finite audit evidence, objective remediation, manual boundaries, mode separation, and structured exact-head eligibility reports
- require audit-authored behavior tests to follow Matt Pocock’s pinned public `tdd` skill
- keep automated wiki `component-scores.json` as the operational audit datastore; any future storage migration remains a separate architecture decision

## Scope

One file only: `docs/specs/AST-029/spec.md`. Implementation is separate follow-up work.

## Decisions resolved

- Test quality: Matt Pocock `tdd` skill, public seams, red before green, one minimal vertical slice
- Eligibility: versioned Night Watch report validated by a trusted exact-head `audit-eligibility` check; no second checked-in audit record
- Storage: automated wiki ledger remains operational; PR holds run evidence; Check Run holds the machine decision

## Validation

- `pnpm check:knowledge`
- repository pre-commit checks

## Authority

Approved in-thread by the owner; phase is `accepted`. Auto-merge behavior itself remains inactive until the separate atomic activation required by FR11.

PR #6098, opened 2026-09-07
docs: make AGENTS the canonical agent guide

cixzhang:
## Summary

- move the repository-specific Claude guidance into canonical `AGENTS.md`
- reduce `CLAUDE.md` to a pointer to `AGENTS.md` and nested path guidance
- remove duplicated instruction blocks and preserve the existing guidance in one place

## Validation

- `pnpm check:setup-contract`
- `pnpm check:sync`

This is intentionally separate from #6032.


cixzhang (commented):

This currently makes the wrong guidance universal. `AGENTS.md` sends agents to the wiki for decisions and architecture and says documentation lives only in headers and component docs, while the current repository contract says only current in-repo records govern work and defines the wider record model. Please keep the canonical guide routing to those current owners.

The canonicalization also is not stable: the existing `astryx init` path rewrites both `AGENTS.md` and `CLAUDE.md`; an exact-head probe restored a generated block to the five-line Claude shim. Please preserve the shim and add a regression check. Keep `/vibe-test`'s tool-specific orchestration with its owning tool docs rather than the universal guide.

[Reviewed by Robohands]

PR #6101, opened 2026-09-07
fix(Popover): unify opening and focus behavior

cixzhang:
## Summary

- apply same-gesture reopen protection through every Popover opening path
- use one content-first semantic focus policy across pointer, keyboard, and assistive-technology activation
- use the existing interaction-modality focus indicator instead of changing focus destination for pointer input
- exclude the injected fallback close control from initial focus and use the labeled dialog surface only when caller content has no genuine control
- remove the redundant internal Popover hook and dismissal plumbing from composed callers
- preserve controlled DropdownMenu request semantics and run selector preparation only after an accepted open
- add RTL overlay-side coverage for Selector and MultiSelector

## Public API

Unchanged:

```ts
show(options?: {skipAutoFocus?: boolean}): void;
toggle(): void;
```

## Validation

- 482 focused Popover, DropdownMenu, Selector, and MultiSelector tests
- Core typecheck
- strict lint on changed TypeScript files
- `pnpm check:knowledge`
- local RTL audit: DropdownMenu, MultiSelector, Popover, and Selector all `RTL-ready`; 0 coverage gaps
- independent M5 review, with findings addressed

### DIFF
--- a/.changeset/popover-opening-focus.md
+++ b/.changeset/popover-opening-focus.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Popover: apply same-gesture reopen protection through every opening path, focus genuine caller content regardless of activation modality, and keep the generated fallback close control hidden until keyboard users reach it.
+
+@cixzhang
--- a/apps/storybook/rtl-audit/targets.json
+++ b/apps/storybook/rtl-audit/targets.json
@@ -351,6 +351,34 @@
     }
   },
   {
+    "component": "Selector",
+    "storyId": "core-selector--default",
+    "dims": [
+      "D4"
+    ],
+    "setup": {
+      "click": "[role=\"combobox\"]"
+    },
+    "selectors": {
+      "overlay": ".astryx-selector-popup",
+      "overlayRoot": "[role=\"combobox\"]"
+    }
+  },
+  {
+    "component": "MultiSelector",
+    "storyId": "core-multiselector--default",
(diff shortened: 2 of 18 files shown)


cixzhang (commented):

The shared opening/focus direction looks right, but three paths still break:

- A controlled `isMenuOpen={false}` request opens the native popover before `onOpenChange(true)`; the callback observes `:popover-open`, then layout rollback closes it. Keep the caller as source of truth throughout.
- The diff narrows `DropdownMenu.onClick` from every trigger click to accepted openings. That changes a released callback without a migration and is event-order dependent: Chromium still fires it on the dismissing click. Keep `onClick` firing for trigger clicks.
- `usePopover({role: 'none'})` with no content control initially focuses the hidden generated “Close popover” button. Exclude it on the neutral fallback path too.

[Reviewed by Robohands]

cixzhang (commented):

Three paths still break for callers and keyboard users:

- Controlled consumers that keep `isMenuOpen={false}` still run `onClick` and `onOpenChange(true)` while the native popover is open, before it rolls back. A rejected request must not open the surface or run accepted-open callbacks.
- Consumers using `onClick` for trigger telemetry miss the dismissing click under one browser event order, while Chromium still fires it under another. Keep every trigger click invoking the released callback.
- A keyboard or AT user opening `usePopover({role: 'none'})` without a content control lands on the generated “Close popover” button. Exclude that fallback control from initial focus on the neutral path.

[Reviewed by Robohands]

PR #6107, opened 2026-09-07
fix(ChatComposerInput): drop `aria-multiline` when triggers make the editable a combobox

Kyujenius:
## Summary

`aria-multiline="true"` was hardcoded on the contenteditable in `ChatComposerInput.tsx`, while `useTriggerMenu.tsx` flips that element's role to `combobox` whenever `triggers` are set. ARIA supports `aria-multiline` on `textbox`, [not on `combobox`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-multiline), so axe reported `aria-allowed-attr` (critical) on all 10 stories that configure triggers.

The split across two files is why it went unnoticed, so the attribute moves into the hook's `ariaProps`, where the role is chosen.

**ARIA pattern:** [APG Combobox](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/) when `triggers` are set, plain `textbox` otherwise; unchanged by this PR. The implementation diverges from the pattern in a few pre-existing ways — out of scope here, but happy to follow up separately if useful.

**[AST-009](https://github.com/facebook/astryx/blob/main/docs/specs/AST-009/spec.md) (FR8):** no real-AT trigger. FR3 applies — the claim ends at browser exposure and is proved by the accessibility tree. FR4 is not triggered: no live region changes, and no claim about what is announced.

## Test Plan

- [x] `pnpm lint:strict`, `pnpm test` (9139 passed), `pnpm build`
- [x] The new combobox test fails on an unmodified tree
- [x] `pnpm a11y:audit -- --components ChatComposerInput` and `--components ChatLayout` — the 10 violations gone, gate passes. Full-surface run: 108 → 98, **0 new**.
- [x] Accessibility tree unchanged — the browser never mapped the attribute on `combobox`:

  |                      | DOM      | computed `multiline` |
  | -------------------- | -------- | -------------------- |
  | `textbox` (control)  | `"true"` | `true`               |
  | `combobox`, after    | absent   | absent               |
  | `combobox`, re-added | `"true"` | absent               |

No rendered output or public API change; the 10 resolved baseline entries are removed per that file's `$comment`.

## Related Issues

From the weekly a11y scan #4681 (2026-09-07 run). Of its 36 findings, 28 are `color-contrast` (theme palette, worked separately); this PR addresses only the `aria-allowed-attr` one.

### DIFF
--- a/.changeset/chat-composer-input-aria-multiline.md
+++ b/.changeset/chat-composer-input-aria-multiline.md
@@ -1,0 +1,9 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] ChatComposerInput: drop `aria-multiline` once triggers make the editable a combobox
+
+`aria-multiline` was hardcoded on the contenteditable element while `useTriggerMenu` owns its role, so configuring `triggers` switched the role to `combobox` — which ARIA 1.2 does not list `aria-multiline` under — and axe flagged `aria-allowed-attr` (critical) on the 8 ChatComposerInput trigger stories and the 2 ChatLayout stories that render one. Moves the attribute into the hook's `ariaProps`, where the role and the attributes whose validity depends on it are decided together.
+
+@Kyujenius
--- a/.github/a11y-baseline.json
+++ b/.github/a11y-baseline.json
@@ -29,24 +29,9 @@
       "helpUrl": "https://dequeuniversity.com/rules/axe/4.12/color-contrast?application=playwright"
     },
     {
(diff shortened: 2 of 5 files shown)


cixzhang (approved):
Thanks, this is good. The invalid combobox attribute is gone, while closed/open Chromium tree checks confirm the exposed semantics stay unchanged.

[Reviewed by Robohands]

Kyujenius:
@cixzhang 

Thanks for the kind explanation!

PR #6109, opened 2026-09-07
fix(build): refuse a Turbopack config in withAstryx instead of building unstyled

### LINKED ISSUE #5921: [Infra] No source build survives Turbopack: every documented path configures nextConfig.webpack, which Turbopack ignores

joaodotwork:
Fixes #5921. Draft while the one compatibility question on that issue is open — see the end.

## Problem

`withAstryx()` installs every alias it needs inside `nextConfig.webpack`, and Turbopack never calls that hook. So `withAstryx()` under Turbopack reinstates exactly the failure #5932 removed: the app resolves `@astryxdesign/*` through `default` to `dist`, whose runtime class names are disjoint from the CSS the PostCSS pass compiles out of source, and the build **succeeds** with an unstyled page and nothing logged.

Next 16 defaults to Turbopack, so a project scaffolded today gets the unstyled page from a *correctly configured* source build. That is why this is separate from #5920 rather than a regression of it.

## Change

**1. Refuse under Turbopack** (`assertWebpack()`, called at config construction). There is no configuration in which the combination does what the author intended, so it throws rather than warning — an unstyled production deploy costs more than a failed build. The message names both ways out: keep the source build and pass `--webpack`, or drop `withAstryx()` and consume the pre-built package.

Detection is `process.env.TURBOPACK`, verified rather than assumed — Next 16.3.4 sets `TURBOPACK=1` for both `next dev --turbopack` and `next build --turbopack`, in the parent process *and* in each build worker, and leaves it unset under `--webpack`.

**2. Report an empty alias map.** `sourceEntryAliases()` skips a package that is not installed, or whose export map ships no `source` condition. Correct per package — an app need not depend on all three. Skipping *every* one of them is not: the result is zero aliases, which is the pre-0.5.3 config, and the app goes back to resolving dist against source-compiled CSS. A `console.warn` rather than a throw, because an install layout this helper cannot walk is not proof the build is wrong.

## Evidence

Unit tests: **6 added, 27 passing** in `packages/build/src/next.test.mjs`, no changes to the existing 21.

End to end against a real Next 16.3.4 app on `0.5.3`, measured on the served HTML and CSS rather than the exit code:

| | before | after |
|---|---|---|
| `next build --turbopack` | exits 0, `/` prerenders, **134 of 141** rendered classes have no matching CSS rule | fails at config load with the message |
| `next build --webpack` | 0 of 141 unmatched | 0 of 141 unmatched, `○ (Static)` |

The harness that produces the unmatched-class count builds each documented setup under both bundlers, serves it, and cross-checks every class on a rendered element against the served stylesheet. Happy to extract it as a standalone repro if useful — it is what turned up #5920.

## The open question

From #5921, and the reason this is a draft: **throw or warn for the Turbopack case?**

A throw is the useful behaviour, but it turns a currently-succeeding build into a failing one — a break for anyone who has `withAstryx()` in a Turbopack config and has not noticed, which is exactly the population this issue is about. This PR implements the throw; switching it to a warn is a one-line change if you would rather not take that in a `0.x` minor.

I have left the changeset off until that is settled, since the category (`fix` vs `breaking`) follows from the answer and drives the bump.

Checked `main` and the open PRs first per `AGENTS.md`: `packages/build/src/next.js` is unchanged since `bea53b23` and no open PR touches it.

### DIFF
--- a/.changeset/refuse-turbopack-in-with-astryx.md
+++ b/.changeset/refuse-turbopack-in-with-astryx.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/build': minor
+---
+
+[breaking] `withAstryx()` refuses a Turbopack config instead of building an unstyled app. Every alias the helper installs lives in `nextConfig.webpack`, which Turbopack never calls, so the app resolved `@astryxdesign/*` to dist while PostCSS compiled the library from source — disjoint class names, an exit code of 0, and an unstyled page. It now throws, naming both ways out: `--webpack`, or drop the helper and consume the pre-built package. Also warns when no package yields a `source` entry, which reaches the same unstyled state by another route. (#6109)
+@joaodotwork
--- a/.changeset/refuse-turbopack-in-with-astryx.md
+++ b/.changeset/refuse-turbopack-in-with-astryx.md
@@ -1,6 +1,6 @@
 ---
-'@astryxdesign/build': minor
+'@astryxdesign/build': patch
 ---
 
-[breaking] `withAstryx()` refuses a Turbopack config instead of building an unstyled app. Every alias the helper installs lives in `nextConfig.webpack`, which Turbopack never calls, so the app resolved `@astryxdesign/*` to dist while PostCSS compiled the library from source — disjoint class names, an exit code of 0, and an unstyled page. It now throws, naming both ways out: `--webpack`, or drop the helper and consume the pre-built package. Also warns when no package yields a `source` entry, which reaches the same unstyled state by another route. (#6109)
+[fix] `withAstryx()` refuses a Turbopack config instead of building an unstyled app. Every alias the helper installs lives in `nextConfig.webpack`, which Turbopack never calls, so the app resolved `@astryxdesign/*` to dist while PostCSS compiled the library from source — disjoint class names, an exit code of 0, and an unstyled page. It now throws, naming both ways out: `--webpack`, or drop the helper and consume the pre-built package. Also warns when the merged alias map claims none of the packages, which reaches the same unstyled state by another route. (#6109)
 @joaodotwork
--- a/apps/example-nextjs-source/README.md
+++ b/apps/example-nextjs-source/README.md
@@ -2,6 +2,25 @@
 
 Reference application for compiling **@astryxdesign/core** from raw TypeScript + StyleX source alongside product code. Uses `@astryxdesign/build` for independent CSS layer separation.
 
+> **Runs on webpack, not Turbopack.** The source build configures resolution through
+> `nextConfig.webpack`, which Turbopack never calls — under Turbopack the app loads
+> `dist` while PostCSS compiles the library from `source`, and the page renders unstyled
+> without erroring.
+>
+> This example pins Next 15, where webpack is still the default, so its plain
+> `next dev` / `next build` scripts are correct as they stand. **On Next 16 the default
+> flips to Turbopack**, and a source build has to name the bundler:
+>
+> ```bash
+> next dev --webpack
+> next build --webpack
+> ```
+>
+> (`--webpack` is a Next 16 flag; Next 15 does not accept it.) `withAstryx()` throws
+> under Turbopack rather than producing the unstyled build. If you would rather not pin
+> the bundler, take the pre-built approach in
+> [example-nextjs](../example-nextjs/) instead.
+
 ## Why source build?
 
 |               | Source build                                       | Dist build                   |
--- a/packages/build/README.md
+++ b/packages/build/README.md
@@ -18,11 +18,11 @@ StyleX generates atomic CSS: same declaration = same class name. Without separat
 
 ## Packages
 
-| Export               | Purpose                                       | Platform                    |
-| -------------------- | --------------------------------------------- | --------------------------- |
-| `@astryxdesign/build/babel`   | Babel plugin: splits class prefixes per file  | Next.js, any babel pipeline |
-| `@astryxdesign/build/postcss` | PostCSS plugin: compiles + splits CSS layers  | Next.js                     |
-| `@astryxdesign/build/vite`    | Vite plugin: wraps unplugin + splits layers   | Vite, Storybook             |
+| Export                        | Purpose                                      | Platform                    |
+| ----------------------------- | -------------------------------------------- | --------------------------- |
+| `@astryxdesign/build/babel`   | Babel plugin: splits class prefixes per file | Next.js, any babel pipeline |
+| `@astryxdesign/build/postcss` | PostCSS plugin: compiles + splits CSS layers | Next.js                     |
+| `@astryxdesign/build/vite`    | Vite plugin: wraps unplugin + splits layers  | Vite, Storybook             |
 
 ## Install
 
@@ -40,6 +40,27 @@ npm install -D @stylexjs/unplugin
 
 ## Next.js Setup
 
+> **Requires the webpack bundler.** Every step below configures resolution through
+> `nextConfig.webpack`, and Turbopack never calls that hook. Under Turbopack the app
+> resolves `@astryxdesign/*` to `dist` while the PostCSS pass compiles the library from
+> `source` — the two emit disjoint class names, so the build succeeds and the page
+> renders unstyled with nothing logged.
+>
+> Next.js 16 selects Turbopack by default, so name the bundler explicitly:
+>
+> ```bash
+> next dev --webpack
+> next build --webpack
+> ```
+>
+> `withAstryx()` throws when it sees `TURBOPACK` set rather than letting that through.
+> Next 16 also rejects a `webpack` config with no `turbopack` config on its own when no
+> bundler flag is given.
+>
+> If you would rather not pin the bundler, take the pre-built package instead: import
+> `@astryxdesign/core/astryx.css` and skip the babel and PostCSS setup entirely. See
+> [example-nextjs](../../apps/example-nextjs/).
+
 ### 1. babel.config.js
 
 ```js
@@ -56,8 +77,12 @@ module.exports = {
         treeshakeCompensation: true,
         enableInlinedConditionalMerge: true,
         aliases: {
-          '@astryxdesign/core/*': [path.join(__dirname, 'node_modules/@astryxdesign/core/*')],
-          '@astryxdesign/core': [path.join(__dirname, 'node_modules/@astryxdesign/core')],
+          '@astryxdesign/core/*': [
+            path.join(__dirname, 'node_modules/@astryxdesign/core/*'),
+          ],
+          '@astryxdesign/core': [
+            path.join(__dirname, 'node_modules/@astryxdesign/core'),
+          ],
         },
         unstable_moduleResolution: {type: 'commonJS'},
       },
@@ -84,8 +109,12 @@ module.exports = {
             treeshakeCompensation: true,
             enableInlinedConditionalMerge: true,
             aliases: {
-              '@astryxdesign/core/*': [path.join(__dirname, 'node_modules/@astryxdesign/core/*')],
-              '@astryxdesign/core': [path.join(__dirname, 'node_modules/@astryxdesign/core')],
+              '@astryxdesign/core/*': [
+                path.join(__dirname, 'node_modules/@astryxdesign/core/*'),
+              ],
+              '@astryxdesign/core': [
+                path.join(__dirname, 'node_modules/@astryxdesign/core'),
+              ],
             },
             unstable_moduleResolution: {type: 'commonJS'},
           },
@@ -97,6 +126,10 @@ module.exports = {
 ```
 
 ### 3. next.config.mjs
+
+Or `withAstryx()` from `@astryxdesign/build/next`, which sets both of these for you.
+Either way the work happens in the `webpack` hook, so the build has to run with
+`--webpack` — see the note at the top of this section.
 
 ```js
 const nextConfig = {
@@ -164,7 +197,10 @@ export default defineConfig({
   ],
   resolve: {
     alias: {
-      '@astryxdesign/core': path.resolve(__dirname, 'node_modules/@astryxdesign/core/src'),
+      '@astryxdesign/core': path.resolve(
+        __dirname,
+        'node_modules/@astryxdesign/core/src',
+      ),
     },
   },
   optimizeDeps: {
--- a/packages/build/src/next.js
+++ b/packages/build/src/next.js
@@ -197,8 +197,19 @@ function sourceEntryAliases(packages, context) {
  * compiles the library from source.
  */
 function aliasCoversAstryx(alias, packages) {
-  const claims = request =>
-    packages.some(name => request === name || request.startsWith(`${name}/`));
+  const claims = request => {
+    // A wildcard key claims by pattern, not by literal name: `@astryxdesign/*`
+    // routes every package under the scope even though it matches none of them
+    // as a string. Same matcher `withoutCallerOverrides` uses to decide which
+    // generated entries a caller has already spoken for.
+    if (request.includes('*')) {
+      const pattern = wildcardPattern(request);
+      return packages.some(name => pattern.test(name));
+    }
+    return packages.some(
+      name => request === name || request.startsWith(`${name}/`),
+    );
+  };
   if (Array.isArray(alias)) {
     return alias.some(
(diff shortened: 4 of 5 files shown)


joaodotwork:
CLA signed. The Meta CLA check on this PR ran at 10:34 UTC, before the signature, so it is showing a stale `action_required` — it should clear on the next re-check or push.

Everything else is green: `reconcile`, `spec-owner-approval` (no knowledge records changed), Vercel, and `Flag review signals`. Still a draft pending the throw-versus-warn answer on #5921.

joaodotwork:
Marked ready for review, and added the changeset.

I had left this as a draft pending the throw-versus-warn answer on #5921, which was the wrong way round: `imdreamrunner` and `josephfarina` are only requested once the PR is marked ready, and the four workflows are awaiting maintainer approval — so the draft state was preventing the review that would answer the question. Draft signals "not ready to review or merge"; this is ready to review and not ready to merge, and review is exactly what it needs. The question is in the PR body and unchanged.

**The changeset is `[breaking]` → minor, and that is the decision worth a maintainer's eye rather than mine.** `check-changesets` couples the two while 0.x, and a throw that turns a currently-succeeding build into a failing one is the breaking tier by that rule. But all publishable packages are a `fixed` group, so this co-bumps the whole system to a new minor on the strength of one guard in one build helper. That is a real cost, and it is the strongest argument for warning instead.

So the two coherent shapes are:

| | behaviour | changeset | cost |
|---|---|---|---|
| as it stands | throws under Turbopack | `[breaking]` → minor | system-wide minor bump; breaks builds that currently "work" unstyled |
| the alternative | warns under Turbopack | `[fix]` → patch | no break, but easy to miss in CI output — and this issue exists because the failure is already silent |

Happy to switch to the warn shape on a word; it is a one-line change plus a changeset edit. I have implemented the throw because an unstyled production deploy costs more than a failed build, and because there is no configuration in which `withAstryx()` under Turbopack does what the author intended — but the fixed-group bump is a policy consequence I cannot weigh from outside.

One note on CI: `build`, `lint`, `test` and `docsite-test` have not run yet — they need a maintainer to approve workflows for a first-time contributor. Locally `pnpm check:repo` passes, prettier is clean, and `packages/build/src/next.test.mjs` is 27 passing (6 added, the existing 21 untouched).

cixzhang (changes requested):
Thanks for digging into the silent styling failure. Two things need changing before this is safe to ship.

- `packages/build/src/next.js:270`: empty generated aliases are not the same as an ineffective config. I reproduced a caller alias that resolves correctly while this warning still says the app “will render unstyled.” A working caller alias must stay quiet; warn only when effective source resolution is actually missing.
- `.changeset/refuse-turbopack-in-with-astryx.md:2`: this is not a breaking release. A valid supported Turbopack source build did not work before; explicit Turbopack built broken output, while Next 16.3.4 already rejects the default-Turbopack/webpack-config combination. This should be a patch rather than `[breaking]`/minor.

[Reviewed by Robohands]

joaodotwork:
Both addressed in `b97db7d`. Thanks — the second one I had genuinely wrong.

**The warning tested the wrong thing.** It fired when `sourceEntryAliases()` generated nothing, but generating nothing is not a fault on its own: a caller alias that already routes the packages is a working config this helper has nothing to add to, and it was being told it would render unstyled. The check now runs against the **merged** alias map and asks whether anything claims an astryx request at all — generated or caller-supplied, object- or array-shaped, from the config or from the caller's own `webpack` hook. Where a caller points its alias is its business; the only fault left is a map that names none of the packages, which is the pre-0.5.3 config.

Three regression tests for exactly the case you reproduced — a caller alias in the config, the array-shaped form, and one contributed from the caller's `webpack` hook. All must stay quiet. 30 passing.

**The changeset is now `[fix]` / patch, and your reasoning is right.** I checked it rather than just taking it, and Next 16.3.4 does reject the combination on its own:

```
$ next build          # no bundler flag, nextConfig.webpack present
⨯ ERROR: This build is using Turbopack, with a `webpack` config and no `turbopack` config.
```

So the three cases before this PR were: default → Next already errors; explicit `--turbopack` → built an unstyled app; `--webpack` → worked. The only behaviour this change touches was already broken, so nothing that worked is being taken away. `[breaking]` was my misreading, and it also avoids a system-wide minor across the `fixed` group, which was the part I was least comfortable proposing.

Re-verified end to end on 0.5.3 + this branch: `--webpack` builds, prerenders `○ (Static)`, and emits no warning; `--turbopack` refuses at config load.

`pnpm check:repo` and prettier pass locally. The required `build`/`lint`/`test` jobs still need a maintainer to approve workflows for a first-time contributor, so their state isn't ours to clear.

cixzhang (changes requested):
Thanks for fixing the caller-prefix case and the release classification. Two things still need changing before this is ready.

- `packages/build/src/next.js:200`: wildcard aliases still take the false-warning path. The existing resolver tests prove `@astryxdesign/*` resolves successfully, but the same test run now logs “none was configured” because `aliasCoversAstryx()` compares the literal `*` key against concrete package names. Please make working wildcard aliases from the config or caller hook stay quiet and add warning assertions around those existing resolver cases.
- `packages/build/README.md:99`: the public source-build setup still depends on `nextConfig.webpack` but never says Next 16 must run with `--webpack`. Someone can still explicitly select Turbopack and get the unstyled build this PR is meant to prevent. Please document the requirement here and in the source example before this closes #5921.

[Reviewed by Robohands]

joaodotwork:
Acknowledged — both points land, and I reproduced the first before touching it.

**1. Wildcard false warning — confirmed and fixed locally.** With a caller wildcard, `withoutCallerOverrides()` correctly drops every generated entry the pattern claims, so the merged map comes out as exactly `["@astryxdesign/*"]` — and `aliasCoversAstryx()` compared that literal key against concrete package names, so it saw no coverage and warned. Reproduced standalone: 1 warning before, 0 after.

The fix reuses the `wildcardPattern()` helper already in the file, the same matcher `withoutCallerOverrides()` uses to decide what a caller has spoken for — so the two now agree on what a wildcard claims instead of disagreeing.

Also added the warning assertions you asked for, as a `beforeEach`/`afterEach` pair around the whole existing `withAstryx` resolver block rather than per test: every case in there is a config that resolves successfully, so none of them may warn. That way a future coverage check that misreads a working alias fails the suite instead of quietly logging into it — which is exactly what mine did.

**2. Documentation — not done yet.** Fair, and it is the part that actually closes #5921: the guard stops the helper being used wrongly, but the README still describes a setup whose bundler requirement is invisible. I will add the `--webpack` requirement to the Next.js Setup section in `packages/build/README.md` and to `apps/example-nextjs-source` — the example's scripts are bare `next dev` / `next build`, which on Next 16 select Turbopack and would now hit the new error, so those want to be explicit too.

Pushing both together rather than piecemeal, so you get one re-review rather than two. Nothing else outstanding from either review as far as I can tell — no inline threads on my side, and the enrichment report is clean (no components changed, no bundle delta, no a11y violations).

joaodotwork:
Both addressed in `d14b78f`.

**Wildcard coverage.** The fix reuses `wildcardPattern()` — already in the file, already used by `withoutCallerOverrides()` to decide what a caller has spoken for — so the two functions now agree on what a wildcard claims instead of disagreeing. That disagreement was the whole bug: `withoutCallerOverrides()` correctly dropped every generated entry the pattern claimed, leaving `["@astryxdesign/*"]`, and then the coverage check compared that literal key against concrete package names. One warning before, none after.

Warning assertions are a `beforeEach`/`afterEach` pair around the whole existing `withAstryx` resolver block rather than per case, since every config in there resolves successfully and so none may warn. 30 passing.

**Documentation.** A note at the top of `## Next.js Setup` in `packages/build/README.md`: the section works through `nextConfig.webpack`, Turbopack never calls it, Next 16 defaults to Turbopack so the bundler has to be named, and the pre-built package is the way out if pinning it is unwelcome. Plus a pointer at the `next.config.mjs` step, and the same requirement in `apps/example-nextjs-source/README.md`.

**One thing I deliberately did not do:** add `--webpack` to the example's scripts. I started to, then checked — the example pins `next: ^15.5.21`, and `--webpack` is a Next 16 flag. `next@15.5.25` ships zero occurrences of it, while it does ship `--turbopack`. So passing it there would have broken the example. The README instead says the plain `next dev` / `next build` scripts are correct under the Next 15 pin, and names `--webpack` as what Next 16 requires — with a parenthetical that Next 15 does not accept it, since that is an easy trap for anyone copying the snippet.

Happy to bump that example to Next 16 and make the scripts explicit if you would rather have it demonstrating the Next 16 path, but that felt like a maintainer's call on the example rather than something to fold into this fix.

`check:repo`, `check:changesets` and prettier all pass locally.

cixzhang (changes requested):
Thanks, the wildcard cases, release classification, and Next 15/16 guidance are fixed. One working alias shape still hits the original false-warning problem.

`packages/build/src/next.js:209`: a scope-prefix alias such as `{'@astryxdesign': customRoot}` correctly resolves `@astryxdesign/core` through the caller path, but `aliasCoversAstryx()` checks the prefix in the opposite direction and logs that no alias was configured. I reproduced this for object, array, and caller-hook aliases. Please keep these working prefix aliases quiet too and cover them with the same warning assertions.

[Reviewed by Robohands]

joaodotwork:
Good catch, and reproduced for all three forms before fixing — `e305345`.

You've now found three cases in this one check, and the reason is that I kept enumerating alias shapes rather than applying webpack's rule. So it now uses the rule: a non-exact alias `key` intercepts a request `R` when `R === key` or `R` starts with `${key}/`, which means an entry covers the packages when one of them sits **at or below** its key. `'@astryxdesign'` is an ancestor of `@astryxdesign/core`, so it qualifies; I had only been looking the other way down the tree. A key under a package still counts, since that's the shape of the entries this helper generates, and `*` keys claim by pattern.

Checked every shape I could think of — scope prefix (object, array, hook), wildcard, package prefix, exact `$` key, subpath key — all quiet, while a config that routes nothing still warns. Three tests added for the scope-prefix forms, 33 passing.

I also noted in the code why this errs toward silence: a false positive tells someone their working build is broken, whereas a false negative just says nothing. If you'd still rather this check didn't exist — the Turbopack throw is what actually closes #5921 and it hasn't been in question — I'm happy to drop it. Your call, and no attachment on my side.

cixzhang (changes requested):
Thanks, the working scope-prefix cases now resolve quietly across object, array, and caller-hook aliases. One exact-alias case still suppresses the new missing-resolution warning.

`packages/build/src/next.js:227`: `aliasCoversAstryx()` strips `$` and ignores array `onlyModule`, then applies prefix matching. So `@astryxdesign$` (or `{name: '@astryxdesign', onlyModule: true}`) resolves no Astryx package but suppresses the new warning. I reproduced this in object, array, and hook forms: all three fail to resolve `@astryxdesign/core` and log zero warnings. Please preserve exactness when checking coverage and assert these ineffective exact-scope aliases warn.

[Reviewed by Robohands]

joaodotwork:
Right on both counts — `597a158`.

`$` and `onlyModule` are the same statement in two spellings, and I was discarding one and never reading the other. Exactness now threads through from both, and it turns out to remove the ancestor case and only that one: an exact key *at or below* a package still routes a real request, so `@astryxdesign/core$`, its array equivalent, and an exact subpath key stay quiet, while `@astryxdesign$` warns.

Six tests — three asserting each ineffective form warns, three holding that boundary. They use an app with nothing installed; with a package present the generated entries route it and there is correctly nothing to report, which is what hid this from me. 39 passing.

One thing your note made me notice: the message still said *"and none was configured"*, which is untrue here — something **is** configured, it just doesn't match. It now says no alias routes the packages and names the exact-key trap by way of explanation.

This is also the first of these where the check was too quiet rather than too loud, and it's a useful correction to a line I'd written into the comment. Erring toward silence is right for an *ambiguous* config, but an exact-scope key isn't ambiguous — it's determinately ineffective — so the comment now distinguishes the two rather than blanket-preferring silence.

One case I've deliberately left quiet, flagging it in case you'd rather it didn't: `{'@astryxdesign/core': false}` disables resolution while counting as coverage. It's explicit config so silence seemed right, but it is arguably the same shape of ineffectiveness. Happy to fold it in.

And thanks for the Discord offer across these — I've been fine on the thread, but if a fifth round would be quicker spoken than written I'll happily take you up on it.

cixzhang (approved):
Thanks, this now handles exact scope aliases across object, array, and hook forms. Working aliases stay quiet, and the documented Next 15/16 source-build paths pass.

[Reviewed by Robohands]

PR #6007, opened 2026-09-04
feat(Banner): expose frame theme target

freddymeta:
Closes #6006. Implements the Banner frame contract approved in #6147.

## Summary

- add `astryx-banner-frame` to the existing visible Banner frame
- reflect `container` and `elevation` for scoped frame shadow and radius rules
- preserve `astryx-banner` on the Status surface and preserve every existing child target
- update the component contract, consumer docs, generated probe coverage, focused tests, and changeset

This lets themes style the frame's `boxShadow` and `borderRadius` directly instead of relying on Banner's DOM structure.

## Validation

- `pnpm lint:strict`
- `pnpm build`
- `pnpm exec vitest run packages/core/src/Banner packages/core/src/theme/themingTargets.test.ts` — 470 passed
- `pnpm -F @astryxdesign/core typecheck`
- `pnpm storybook:build`
- `pnpm a11y:audit -- --components Banner` — 20 stories, 0 violations
- `pnpm rtl:audit -- --filter Banner` — 0 gaps
- `pnpm visual:probe-theme:check` — 277 targets, 898 selectors
- `node .github/scripts/visual-gate/gate.mjs reach --only core-banner` — 8 reached, 0 missed

The full local suite currently reports two story-tree fixture failures already present on current `main`; this PR does not change the test or any implicated story file.

PR #6116, opened 2026-09-07
docs(Banner): record current surface contract

cixzhang:
## Why

Banner’s existing public `banner` target sits on the colored header, while the outer root paints whole-banner elevation and the elevated card radius. Before deciding the target proposed in [facebook/astryx#6007](https://github.com/facebook/astryx/pull/6007), the current behavior and ownership need a durable component contract.

## What

- add a draft Banner component contract that records the current surface owners, target placement, and private radius routing
- document `Status surface` as Banner’s primary painted surface and owner of the existing `banner` target
- document `Content surface` as the secondary owner of `banner-content`
- document `Banner frame` as the outer surface owner for whole-banner elevation and the elevated-card silhouette, while leaving separate target admission unresolved
- record Banner’s current verified RTL non-applicability so documentation-only changes remain covered by the scoped audit

This PR adds no runtime behavior, DOM, public target, or public API.

## Risk

Documentation and draft specification only.

## Testing

- `pnpm check:knowledge`
- `pnpm exec vitest run packages/core/src/theme/themingTargets.test.ts packages/core/src/docPropLiterals.test.ts packages/core/src/docPropReferences.test.ts`


cixzhang (commented):

Theme authors who set `banner.borderRadius` get a 37px status surface, but the elevated frame and content remain 12px, so the card silhouette splits. The current contract says the value reaches all three painters. Please record that shipped reachability gap instead of promoting the intended route as current behavior.

[Reviewed by Robohands]

PR #6119, opened 2026-09-07
docs: make Astryx CLI bootstrap shell-portable

cixzhang:
## Summary

- replace the scalar `ASTRYX="node …"` command variable with a portable shell function
- update every documented bootstrap/on-demand invocation to call that function

## Why

zsh does not split a scalar variable into command plus arguments, so `$ASTRYX help` attempted to execute a literal file named `node packages/cli/clients/cli/bin/astryx.mjs`. Night Watch hit this during its first test audit.

## Validation

Executed the documented function and both `help` and `component AvatarGroupOverflow --dense` successfully under zsh and Bash.

One file only; no CLI behavior changes.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -181,24 +181,24 @@ Documentation lives in two places:
 <!-- ASTRYX-CLI:START -->
 
 Astryx CLI|Run from repo root. Load agent docs before any component work.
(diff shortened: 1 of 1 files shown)


cixzhang (commented):
Semantic verdict: approve

The documented function now works in zsh and Bash for help and dense component lookup, with no unrelated changes.

[Reviewed by Robohands]

PR #6121, opened 2026-09-07
fix(theme): remove bare prop and state classes

cixzhang:
## Summary

- replace legacy prop/state value classes with stable component target classes plus reflected `data-*` attributes
- add a normal `astryx upgrade --apply` migration frozen to the complete public v0.5.4 runtime contract: 186 targets, 988 finite tokens, four unbounded value domains, legacy aliases, and same-element primitive composition
- preserve every old qualified match with specificity-equivalent `:is(.old-class, [data-*], …)` unions while leaving unknown/custom and unqualified classes unchanged
- document prebuilt-theme rebuild requirements and every changed public theme helper; add package-boundary and generated-theme regression coverage

## Test plan

- `pnpm build`
- `pnpm check:repo`
- `pnpm install --frozen-lockfile`
- focused Core, CLI codemod, current-main adaptation/Banner, and probe-theme suites
- exact read-only transform run against eight representative real-world `.css` fixtures: 0 failures, 22 attribute arms added, consumer-owned classes/comments preserved, all outputs idempotent
- nested functional pseudos and `@scope` selector preludes covered
- generated CSS scan across maintained themes: no `.astryx-<target>.<value>` selectors
- exact-head required CI: full, `pr-a11y`, and `pr-visual`

## Risk / rollback

- **Risk:** consumer CSS that directly targeted removed value classes needs migration.
- **Mitigation:** the codemod operates only on parsed CSS selector nodes. It never rewrites declarations, comments, JavaScript/TypeScript strings, unqualified classes, or custom/unknown qualified classes. Each known value retains its original class arm for consumer `className` matches and adds every v0.6 data-attribute arm that could have produced the v0.5.4 class.
- **Rollback:** revert this PR to restore the legacy classes while retaining the documented migration path.


cixzhang (commented):

The release currently removes bare `.astryx-*.<value>` selectors without giving 0.5.4 consumers a working migration. `astryx upgrade` selects no codemod from 0.5.4, and the older optional transform only recognizes `.xds-*`, so apps using the documented compatibility classes remain unchanged and silently lose themed styling; prebuilt custom themes also need to be regenerated.

Please add a current-version migration for safely rewritable Astryx selectors, concrete instructions for ambiguous selectors and prebuilt CSS, and name the changed return behavior of the exported `themeProps`, `parseStyleKey`, and theme CSS generators.

[Reviewed by Robohands]

cixzhang (commented):

0.5.4 consumers still do not have a complete path through this break. The new Changeset documents manual selector edits, but `astryx upgrade` still has no current transform for safely identifiable `.astryx-<target>.<value>` selectors. It also does not tell custom-theme owners to rebuild precompiled CSS or tell callers that the exported `themeProps`, `parseStyleKey`, and theme generators now return different values.

Please add the current-version migration, document regeneration for prebuilt themes, and name those exported behavior changes. Also update the five remaining AvatarGroup assertions; `test-ui` is red because they still expect the removed size classes.

[Reviewed by Robohands]

cixzhang (commented):

The stale AvatarGroup assertions are fixed, but 0.5.4 users can still lose custom styling silently. `astryx upgrade` selects no transform from 0.5.4, and the existing selector migration rewrites only `.xds-*`, so safely identifiable `.astryx-button.primary` selectors remain unchanged. Prebuilt custom themes also keep the old selectors until rebuilt. The release note does not explain either path or name the changed return behavior of exported `themeProps`, `parseStyleKey`, `generateThemeRules`, `generateThemeRulesSplit`, and `generateThemeCSS`.

Please add a current-version migration for known target/axis selectors, leave actionable guidance for ambiguous selectors, and document rebuilding prebuilt themes plus those exported behavior changes.

[Reviewed by Robohands]

cixzhang (commented):

0.5.4 consumers can still lose custom styling silently. `astryx upgrade` selects no transform after 0.5.4, and the existing selector transform rewrites only `.xds-*`, so safely identifiable `.astryx-button.primary`, `.sm`, and `.checked` selectors remain unchanged. The migration note also omits rebuilding prebuilt theme CSS and the changed return values of public `themeProps`, `parseStyleKey`, `generateThemeRules`, `generateThemeRulesSplit`, `generateOnMediaCSS`, and `generateThemeCSS`.

Please add a current-version codemod for known target/axis selectors, leave concrete instructions for ambiguous selectors, tell built-theme consumers to rerun `astryx theme build`, and name those public return changes.

[Reviewed by Robohands]

cixzhang (commented):

The current migration still leaves valid 0.5.4 selectors unchanged with no TODO, so users can run the documented upgrade and silently lose styling. The frozen map records axis names instead of emitted values for targets such as `toggle-button` (`.true`/`.false`), `top-nav` (`.mobile-bar`), indicator icons (`.expanded`/`.collapsed`), and Toolbar sizes; functional pseudo selectors such as `:is(.astryx-button.primary)` are also not traversed. I confirmed these against the v0.5.4 `themeProps()` calls and by running the exact-head transform.

Please derive the map from actual 0.5.4 emitted target/value pairs, recurse through functional pseudo selector lists, and add fixtures proving every known qualified selector is either migrated or receives an actionable TODO.

[Reviewed by Robohands]

cixzhang (commented):

Builders upgrading from 0.5.4 can still lose custom styling silently because the codemod leaves known released selectors unchanged when the Astryx target is only one branch of a functional selector. `:is(.astryx-button, .other).primary` returns unchanged with no TODO, so a primary Button stops matching after bare classes disappear; `.astryx-button:nth-child(2n of .primary)` does too. The current tests accept the first case unchanged.

Please either preserve each known branch’s match without broadening unrelated branches, or leave an actionable TODO for every known selector the transform cannot safely rewrite.

[Reviewed by Robohands]

cixzhang (commented):

Builders upgrading from 0.5.4 can still lose or unexpectedly gain custom styling because nested known selectors remain unchanged with no TODO. `:is(.astryx-button, .other):not(.primary)`, `:is(...):where(.primary)`, and `.astryx-button:nth-child(2n of :is(.primary))` all diverge in Chromium after bare classes disappear.

Please preserve known target/value matches through these nesting boundaries without broadening unrelated branches, or leave an actionable TODO, and add fixtures for each boundary.

[Reviewed by Robohands]

cixzhang (commented):

Builders upgrading from 0.5.4 can still lose custom styling silently because the codemod leaves some standard nested selectors unchanged with no TODO. `:not(:not(.astryx-button)).primary` and `:nth-child(2n of .astryx-button).primary` both match primary Buttons before the class removal, but the exact-head transform returns them unchanged; Chromium then goes from 2→0 and 1→0 matches after v0.6 data attributes replace `.primary`.

Please either migrate these known target/value relationships without broadening unrelated elements, or leave an actionable TODO whenever the transform cannot prove a safe rewrite.

[Reviewed by Robohands]

cixzhang (commented):

Thanks, the two selectors from the last review now preserve their match sets. Native CSS nesting still has the same silent migration hole: `.astryx-button { &.primary { … } }` is valid CSS, but the exact-head codemod returns it unchanged with no TODO. Chromium matches three primary Buttons before the class removal and zero afterward. The transform walks each rule independently, so `&.primary` never inherits its parent target.

Please either migrate known target/value relationships expressed through nesting `&`, or leave an actionable TODO whenever parent selector context is required.

[Reviewed by Robohands]

cixzhang (commented):
Semantic verdict: approve

This exact head now gives v0.5.4 users a complete, behavior-preserving migration for the bare selector-class removal.

- The frozen runtime inventory covers 186 targets, 988 finite tokens, same-element primitive composition, legacy aliases, and every unbounded v0.5.4 domain.
- The CSS parser preserves consumer class matches while adding exact data-attribute arms, including nested functional selectors, `of` filters, `@scope`, and native CSS nesting without changing specificity.
- Prebuilt-theme regeneration and all six exported helper behavior changes are documented and regression-tested.

Exact-head full CI, accessibility, RTL, stable visual, Vercel, smoke, lint, and codemod gates pass. Broad stable visual scope is deferred under the repository's release policy. No blocking findings remain.

[Reviewed by Robohands]

PR #6123, opened 2026-09-07
docs(spec): define theme-aware component adaptations

imdreamrunner:
## Summary

- define `adaptations={{default, rules}}` for one component-owned policy value
- reuse PR #5543's named width points and width/pointer condition grammar
- make `default` the explicit SSR and hydration value
- use `popover | bottom-sheet` for Selector and `native | popover | bottom-sheet` for DateInput/DateTimeInput
- keep structural presentation in JavaScript because it changes DOM, ARIA, focus, and native controls
- preserve direct `presentation` and `nativePicker` props as compatibility syntax
- stage implementation as shared resolver, Selector, and date/date-time picker PRs

## Key boundaries

- no CSS variables or additional React context are needed
- BottomSheet does not choose its own breakpoint policy
- TimeInput and DateRangeInput are out of scope because they do not expose all three picker surfaces
- theme adaptation rule values do not affect component adaptations
- Selector's current inclusive 768px adaptive edge becomes AST-012's exclusive `below: 'md'` edge

## Validation

- independent design review: no actionable issues

Spec approval is intentionally left to the owner.

### DIFF
--- a/docs/specs/AST-031/spec.md
+++ b/docs/specs/AST-031/spec.md
@@ -1,0 +1,387 @@
+---
+schema_version: 1
+template_version: 1
+kind: system-spec
+id: spec:AST-031
+authority: draft
+archive_reason: null
+superseded_by: null
+approved_by: null
+approved_at: null
+phase: proposed
+owners: [cixzhang, imdreamrunner]
+affects_architecture:
+  [architecture:public-component-api, architecture:theme-application]
+affects_families: [family:input-fields]
+affects_contributing: []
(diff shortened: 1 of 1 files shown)


cixzhang:
I like the unification direction and connection to the theme API. One concern: using complex objects and lists here will make the props more difficult to memoize and results in more context for garbage collection over time. Might be worse for components that could appear repeatedly. It's possible to workaround but makes me wonder if the simpler "presentation" API with just a string might be better to start.

PR #6124, opened 2026-09-07
fix: preserve Resizable bound migration semantics

cixzhang:
## Why

The 0.6 Resizable cleanup could silently drop deprecated bounds passed through assembled objects, and its codemod could change property precedence when an inline configuration contained dynamic properties.

## What

- Reject removed pixel-bound aliases even when TypeScript excess-property checks or union types would otherwise allow them through variables.
- Leave inline configurations with spreads or computed properties unchanged and add a manual-migration TODO instead of applying a potentially behavior-changing rewrite.

## Risk

Low. Runtime behavior is unchanged; the correction makes unsupported TypeScript inputs fail loudly and makes ambiguous codemod cases explicit.

## Testing

- 157 focused Resizable, ResizeHandle, and next-codemod tests
- Core and strict CLI typechecks
- Core build
- Repository checks and focused lint

The non-required stable visual job cannot plan shots because Resizable has no seeded visual baseline; this PR changes no rendered output.

### DIFF
--- a/.changeset/fix-resizable-bound-migrations.md
+++ b/.changeset/fix-resizable-bound-migrations.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+'@astryxdesign/cli': patch
+---
+
+[fix] Prevented removed Resizable bounds from being silently ignored and kept ambiguous spread migrations behavior-preserving (#6124)
+@cixzhang
--- a/packages/cli/assets/codemods/transforms/next/__tests__/next-codemods.test.mjs
+++ b/packages/cli/assets/codemods/transforms/next/__tests__/next-codemods.test.mjs
@@ -164,6 +164,16 @@ const regions = resize({regions: {
     expect(output).not.toContain('maxSizePx: 900');
   });
 
+  it('leaves spread-bearing bounds for manual migration', async () => {
+    const input = `import {useResizable} from '@astryxdesign/core/Resizable';
+const legacy = {minSizePx: 160};
+const region = useResizable({minSizePx: 80, ...legacy});`;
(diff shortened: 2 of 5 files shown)


cixzhang (commented):

This still lets a removed bound reach 0.6 and be ignored. `RejectRemovedPixelBounds` checks `keyof Config`, but a union exposes only shared keys:

```ts
declare const config:
  | UseResizableSingleConfig
  | (UseResizableSingleConfig & {minSizePx: number});
useResizable(config); // compiles
```

I verified the equivalent nested multi-region union is rejected, but this single-region case compiles: its `@ts-expect-error` is unused. At runtime 0.6 reads only `minSize`, so builders can still silently lose the minimum. Please make the rejection distribute across single-config union members and add this factory/union case to the type tests.

[Reviewed by Robohands]

cixzhang (commented):

This can still change a builder’s minimum during the 0.6 migration when an inline config uses a computed property:

```ts
const key = 'minSize';
useResizable({[key]: 160, minSizePx: 80});
```

0.5 chooses `minSize` (160) when both spellings are present. This codemod emits `{[key]: 160, minSize: 80}`, so JavaScript precedence changes the minimum to 80. Please treat computed properties as dynamic, like spreads: leave the object unchanged, add the manual-migration TODO, and cover both property orders plus a nested region.

[Reviewed by Robohands]

cixzhang (commented):
Semantic verdict: approve

Builders now get a type error instead of silently losing a removed bound; computed and spread cases stop for manual migration without rewriting precedence. Thanks, this is good.

[Reviewed by Robohands]

PR #6125, opened 2026-09-07
fix(Switch): persistently mount busy live region and localize loading announcement (#6114)

### LINKED ISSUE #6114: Switch: the busy status is announced in English, if it is announced at all
## Who is affected

Anyone using a screen reader who turns on a `Switch` that does asynchronous
work, and anyone using Astryx in a language other than English.

## What happens

While a `Switch` is busy it renders

```tsx
{isBusy && <VisuallyHidden role="status">Loading</VisuallyHidden>}
```

Two things go wrong with that one line.

1. **The status is born with its content.** The live region is mounted at the
   same moment its text appears. Assistive technology announces *changes* to a
   live region, so a region that already contains its text when it is inserted
   is usually silent. The person hears nothing at all: they press the switch,
   the switch does not visibly change yet, and nothing tells them the change is
   in flight.
2. **The word is hardcoded English.** Every other assistive-technology-facing
   string in core goes through the translator — `Button` announces
   `t('@astryx.button.loading')` — but `Switch` says the literal `Loading`. A
   French or Japanese user hears an English word in the middle of their own
   interface.

`Button` already does this correctly: it renders the live region
unconditionally and swaps the translated text in and out.

## Expected

A busy `Switch` announces its busy state, in the user's language, through the
persistent live region the repository already provides (`useAnnounce`, or the
always-mounted-region shape `Button` uses), with the string routed through the
i18n key system.

## Where

`packages/core/src/Switch/Switch.tsx` — the `isBusy` branch of `switchElement`.
Compare with `packages/core/src/Button/Button.tsx`.

## Notes

Found while inventorying Switch's states for the shared accessibility
spec-test contract. The contract does not encode announcements: what an
assistive technology actually says is a real-AT observation and is out of scope
for a runtime contract, so this is filed as its own defect rather than recorded
as a contract failure.

The repository accessibility checklist covers both halves — the "Announcements"
item ("Never render a live region that is born with content") and the "i18n"
item ("Every AT-facing string … goes through `useTranslator()`").

Geervan:
Fixes #6114.

Previously, `Switch` rendered its busy announcement live region dynamically only while `isBusy` was true (`{isBusy && <VisuallyHidden role="status">Loading</VisuallyHidden>}`). This caused two defects:
1. **Born with content:** The live region was mounted at the same moment its text appeared. Because screen readers observe text mutations within pre-existing live regions rather than newly inserted nodes with existing content, the loading announcement was typically silent.
2. **Hardcoded English:** The string `"Loading"` was hardcoded in English instead of going through the `useTranslator` i18n system.

This PR aligns `Switch` with the persistent live-region pattern used by `Button`:
- Unconditionally renders `<VisuallyHidden role="status" aria-live="polite">` so the container is always mounted.
- Sets its text to empty string `""` when idle and dynamically updates it to `t('@astryx.switch.loading')` when `isBusy` is true.
- Adds the `@astryx.switch.loading` key to `packages/core/locales/en.json`.

## Changes

- **`packages/core/src/Switch/Switch.tsx`**: Hooked `useTranslator` and converted the live region to be persistently mounted with dynamic translated text.
- **`packages/core/locales/en.json`**: Added `@astryx.switch.loading` translation entry.
- **`packages/core/src/Switch/Switch.test.tsx`**: Added regression unit tests for:
  - Empty persistent live region present when idle.
  - Live region updating to `"Loading"` when `isLoading` is enabled.
  - Localization of the loading announcement through `InternationalizationProvider` overrides.
  - Live region announcing `"Loading"` during async `changeAction` transitions and clearing on resolution.
- **`.changeset/switch-busy-announcement.md`**: Added changeset entry for patch release.

## Verification

- `npx vitest run packages/core/src/Switch/Switch.test.tsx` (58/58 passed)
- `node scripts/check-i18n-catalog.mjs` (passed with 0 errors)
- `npx vitest run packages/core/src/Button/Button.test.tsx packages/core/src/i18n/useTranslator.test.tsx` (52/52 passed)
- `node scripts/check-changesets.mjs` (passed)

### DIFF
--- a/packages/core/src/Switch/Switch.tsx
+++ b/packages/core/src/Switch/Switch.tsx
@@ -20,6 +20,7 @@ import {
   useId,
   useOptimistic,
   useTransition,
+  useEffect,
   type ChangeEvent,
   type FocusEvent,
   type ReactNode,
@@ -46,7 +47,7 @@ import {switchScope} from './switch.markers.stylex';
 import type {BaseProps} from '../BaseProps';
 import type {SizeValue} from '../utils/types';
 import {themeProps} from '../utils/themeProps';
-import {VisuallyHidden} from '../VisuallyHidden';
+import {useAnnounce} from '../hooks/useAnnounce';
 import {useResolvedRequired} from '../hooks/useResolvedRequired';
 import {useTranslator} from '../i18n';
 
@@ -505,6 +506,13 @@ export function Switch({
   const [optimisticValue, setOptimisticValue] = useOptimistic(value);
   const isBusy = isLoading || optimisticValue !== value;
 
+  const announce = useAnnounce();
+  useEffect(() => {
+    if (isBusy) {
+      announce(t('@astryx.switch.loading'));
+    }
+  }, [announce, isBusy, t]);
+
   const isOn = optimisticValue === true;
 
   // Disabled-reason tooltip. Disabled controls swallow pointer events, so the
@@ -616,10 +624,6 @@ export function Switch({
           {isBusy && <Spinner size="sm" />}
         </div>
       </div>
-      {/* Live region for busy/loading state announcements */}
-      <VisuallyHidden role="status" aria-live="polite">
-        {isBusy ? t('@astryx.switch.loading') : ''}
-      </VisuallyHidden>
     </div>
   );
 
--- a/packages/core/src/Switch/Switch.tsx
+++ b/packages/core/src/Switch/Switch.tsx
@@ -4,7 +4,7 @@
 
 /**
  * @file Switch.tsx
- * @input Uses React, useId, ChangeEvent, FieldLabel, FieldStatus, IconType, InputStatus, useTooltip
+ * @input Uses React, useId, ChangeEvent, FieldLabel, FieldStatus, IconType, InputStatus, useTooltip, i18n (useTranslator)
  * @output Exports Switch component, SwitchProps, SwitchLabelPosition, SwitchLabelSpacing
  * @position Core implementation; consumed by index.ts, tested by Switch.test.tsx
  *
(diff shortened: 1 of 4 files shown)


packages/core/src/Switch/Switch.tsx
```diff
@@ -614,7 +616,10 @@ export function Switch({
           {isBusy && <Spinner size="sm" />}
         </div>
       </div>
-      {isBusy && <VisuallyHidden role="status">Loading</VisuallyHidden>}
+      {/* Live region for busy/loading state announcements */}
+      <VisuallyHidden role="status" aria-live="polite">
+        {isBusy ? t('@astryx.switch.loading') : ''}
+      </VisuallyHidden>
```
cixzhang: We likely want to be using useAnnounce instead of VisuallyHidden which is the older pattern. The useAnnounce hook will include a stable live region and bypass the issue without introducing a permanent DOM element per switch.

cixzhang (changes requested):
Thanks for this fix! This is very close. We'll likely want to use the useAnnounce pattern here over visually hidden.

Geervan:
Updated to use the `useAnnounce` hook instead of `VisuallyHidden` as suggested! 

Announcements now route through the shared singleton live region without introducing per-switch DOM nodes, let me know if I still missed any corners

Geervan:
i have updated the PR description to:
- Accurately reflect the shared `useAnnounce` live-region implementation and the actual test suite present in `Switch.test.tsx`.
- Include the **AST-009 Assistive-Technology (AT) Verification Record**, detailing the idle-to-busy, async `changeAction`, and initially-busy scenarios with the representative desktop matrix (`NVDA + Chrome on Windows` and `VoiceOver + Safari on macOS`) marked as `pending` for the stable release gate.

PR #6128, opened 2026-09-07
feat(Chat): support whitespace queries via allowSpaces on ChatComposerTrigger

### LINKED ISSUE #6110: [Bug] useTriggerMenu closes @ menus when the query contains whitespace
### Description

`useTriggerMenu` stops scanning an active trigger as soon as it encounters a space or newline. That single-token grammar is appropriate for `/` commands, but it prevents `@` menus from searching multi-word entity names and prevents a trailing `@ ` from keeping the complete reference list open.

Consumers currently have no option to configure whitespace behavior per trigger. Maka therefore carries a local `@astryxdesign/core@0.5.2` patch for its Composer Session-reference picker. A public trigger option or built-in distinction between `/` and `@` would let that patch be removed.

Expected behavior: a trigger can opt into queries containing whitespace. Slash-command behavior should remain unchanged by default.

### Reproduction

1. Configure `useTriggerMenu` with both `/` and `@` triggers.
2. Open the `@` menu and type `@Project Alpha`, or type `@ ` to browse all reference candidates.
3. Observe that the active trigger is cleared at the first space, so the menu closes and cannot match the multi-word name.
4. The same single-token behavior is desirable for `/command args`, so changing the global scanner without a per-trigger policy would regress slash commands.

### Astryx Version

@astryxdesign/core@0.5.2

### Environment

Electron 43 / Chromium, macOS. The behavior is in the trigger parser and is not platform-specific.

Downstream context: apache/maka#4446.

Geervan:
Fixes #6110
### Summary
Adds opt-in `allowSpaces?: boolean` to `ChatComposerTrigger` in `@astryxdesign/core`, enabling multi-word trigger queries (e.g. `@Cindy Zhang`) and trailing-space candidate browsing (`@ `).

### Scanner & Termination Rules
- **Punctuation boundaries**: `.`, `,`, `!`, `?`, `;`, `:` unconditionally terminate active queries.
- **Word cap**: Maximum 5 words allowed in the active query (6th word closes the menu).
- **Apostrophe & Hyphen**: Preserved (e.g. `@O'Brien`, `@Jean-Luc's Task`).
- **Cursor-relative binding**: Scanning operates on `textBeforeCursor` up to the active selection offset.
- **Zero-length query**: Immediate punctuation (`@,`, `@.`) returns `null` immediately.
- **Single-token triggers**: Triggers with `allowSpaces: false` or unset (such as `/` slash commands) retain strict space-terminating single-token behavior.

### New Test Suite (`useTriggerMenu.test.tsx`)
Added dedicated hook unit test file `packages/core/src/Chat/useTriggerMenu.test.tsx` covering all 12 scanner edge cases:
1. Single-token `/` command behavior unchanged (closes on space).
2. `allowSpaces: true` multi-word query matching (`@Project Alpha`).
3. Trailing-space browse-all query (`@ `).
4. Punctuation termination, attached and spaced (`.`, `,`, `!`, `?`, `;`, `:`).
5. Zero-length query + immediate punctuation returns `null` (`@,`, `@.`).
6. Preserves apostrophes and hyphens (`@O'Brien`, `@Jean-Luc's Task`).
7. 5-word cap enforcement (5 words active, 6 words closes).
8. Double-space tolerance without closing or over-counting words.
9. Stateless reactivation on backspacing past a terminator.
10. Cursor repositioning / cursor-relative binding to earlier mentions.
11. Word-boundary gate preservation (`email@domain.com`, `foo/bar` do not activate).
12. Newline hard-termination.

### Verification
- `packages/core/src/Chat/useTriggerMenu.test.tsx`: 12/12 passed
- `packages/core/src/Chat/`: 263/263 passed across 19 test files
- ESLint: 0 errors
- Changeset check: Passed

### DIFF
--- a/.changeset/chat-trigger-allow-spaces.md
+++ b/.changeset/chat-trigger-allow-spaces.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] Chat: support whitespace-tolerant search queries via opt-in `allowSpaces` on `ChatComposerTrigger` with punctuation boundary and word-cap termination rules.
+
+@Geervan
--- a/apps/storybook/stories/ChatComposerInput.stories.tsx
+++ b/apps/storybook/stories/ChatComposerInput.stories.tsx
@@ -231,6 +231,7 @@ export const MentionTrigger: Story = {
     const mentionTrigger: ChatComposerTrigger = {
       character: '@',
       searchSource: userSource,
+      allowSpaces: true,
       renderItem: item => (
         <TypeaheadItem
           item={item}
@@ -356,6 +357,7 @@ export const MultipleTriggers: Story = {
     const mentionTrigger: ChatComposerTrigger = {
       character: '@',
       searchSource: userSource,
+      allowSpaces: true,
       onSelect: item => ({
         value: `@${item.id}`,
         label: item.label,
--- a/packages/core/src/Chat/ChatComposerInput.tsx
+++ b/packages/core/src/Chat/ChatComposerInput.tsx
@@ -151,6 +151,12 @@ export type ChatComposerTrigger = {
   loadingText?: string;
   /** Accessible label for the menu. @default 'Suggestions' */
   menuLabel?: string;
+  /**
+   * Whether the search query can contain spaces (e.g. for multi-word mentions like "@Jane Doe"
+   * or candidate browsing with "@ "). When false, the menu closes on the first space.
+   * @default false
+   */
+  allowSpaces?: boolean;
 };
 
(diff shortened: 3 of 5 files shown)


Geervan:
since this is going to have some iterations, i had a question:
- the **Multiple Triggers** story in the storybook only supports `/` and `@` i didnt meddle with it, but thought adding `:` is a logical addition since "multiple triggers"
also updated from the above received comment
- scoped punctuation termination to only apply when `allowSpaces: true`, so regular single-token queries like `/v1.2` or `@john.doe` won't close on dots anymore.
- Added tests for both cases in `useTriggerMenu.test.tsx` (all 14/14 passing).

PR #6122, opened 2026-09-07
docs: define React component runtime architecture

cixzhang:
## Summary

- defines shared runtime roles for render, handlers, refs, Effects, external resources, node replacement, observation, lifecycle ownership, and broadcast identity
- keeps React techniques as guidance unless misuse causes observable correctness, ownership, cleanup, accessibility, or scale harm
- constrains Effects to external-system synchronization while allowing documented singletons and recommending Effect Events for non-reactive latest reads
- permits refs as general non-rendering mutable storage while preventing hidden rendered state
- delegates public `ref` / `handleRef` admission and naming, controlled/optimistic semantics, accessibility, styling, tooling, and public-hook identity to their existing owners

## Audit-row coverage

Owned here:
- C5, C6, C19
- durable runtime outcomes from C1, C9, C10, C16, C22
- a narrow lifecycle/scale boundary informed by C7–C8

Explicitly delegated:
- C2, C13 → public API, family, and component contracts
- C3, C17 → accessibility, interaction, family, and component contracts
- C4, C11, C18, C21 → lint/contributor guidance unless an invariant is violated
- C12 → applicable primitive owner
- C14 → runtime state boundary here; public composition/introspection with API/component owners
- C15 → applicable registry/data owner
- C20 → style/token/component/interaction owner

## Owner decisions

- Effects are constrained, not banned.
- Controlled props are not casually mirrored; current owners may define explicit optimistic seams.
- React owns rendered semantics; refs may hold general non-rendering mutable data.
- Effect Events are guidance for non-reactive latest reads, never hidden dependencies.
- Complex behavior remains with the stable lifecycle owner; private extraction is guidance, not policy.
- `ref` and `handleRef` may coexist, but their public naming/admission belongs to `architecture:public-component-api`; handle identity is unspecified.
- CSS-first and pooled observation are guidance; runtime owns lifecycle and unbounded-fan-out outcomes.

## Overlap receipt

- Searched current records and open PRs by owner/id, Effects, refs, observers, controlled state, DOM ownership, and behavior seams.
- #5642 is a concrete Dialog lifecycle implementation and non-authoritative evidence, not overlapping shared authority.
- #5543 touches `public-component-api` for theme adaptations and does not own React runtime.
- #6046, #6047, and #6048 own RTL/bidirectionality/internationalization and are untouched.

## Validation

- `node scripts/check-knowledge.mjs`
- `pnpm check:sync`
- Prettier
- `git diff --check`
- independent authority map and Matt-style exact-file review

[...]

PR #6129, opened 2026-09-07
docs: define component style authoring architecture

cixzhang:
## Summary

- defines five observable component-style authoring invariants: StyleX lowering, CSS-over-JS presentation ownership, valid static composability, scoped relational state, and named wrapper responsibility
- keeps exact StyleX capabilities, shorthand/reset/null syntax, helper names, and lint matchers as versioned guidance/tooling rather than permanent architecture vocabulary
- permits component-owned non-glyph inline SVG while delegating semantic glyphs to icon authority
- permits wrappers for semantics, behavior/ref, formatting contexts, outside-border spacing, or a required style-bearing DOM boundary when the child has no compatible surface
- delegates tokens, theming axes/targets, BaseProps/public styling, RTL, and RSC module boundaries to their canonical owners

## Audit-row coverage

Owned here:
- Q6 / T14 — wrapper responsibility
- durable composability outcome behind Q7
- T8 — StyleX lowering and CSS-over-JS ownership
- valid-lowering outcome behind T9
- T10 — CSS-over-JS ownership
- T11 — relational-state isolation

Delegated:
- Q4 / T17 / T18 → icon architecture, Icon contract, and theme authoring
- T2b → theme-token architecture
- T13 → future current bidirectionality owner
- T19 / T27 → component-theming-surface; T19 reachability also uses public-component-api
- T20 → public-component-api
- T30 → future current RSC/module-boundary owner

## Owner assumptions

- Inline SVG is prohibited only when it shadows a registry-owned glyph; component-owned non-glyph vectors remain allowed.
- A wrapper needs a named runtime responsibility; source clarity alone is not normative.

## Overlap receipt

- Searched current records and open PRs by owner/id, StyleX, CSS, wrapper, marker, raw CSS, variant axes, and affected paths.
- #5543, #6061, and #6091 overlap theming/visual-prop authority only; this record delegates those rows and does not modify their files.
- Icon scope overlaps current icon architecture and open #5204; no icon authority is changed here.
- #6122 is adjacent React runtime authority but explicitly delegates CSS presentation to this owner.
- #6046, #6047, and #6048 remain untouched.

## False-positive / false-negative audit

Initial bounded current-code matrix: 12 correct, 1 false-positive risk, 1 false-negative risk, 2 untestable.

After exact fixes for child style-bearing surfaces, component-local raw CSS, and StyleX registry/package version drift: **11 correct, 0 false positives, 0 false negatives, 5 intentionally untestable capability-dependent cases** across the same 16 cases. Untestable cases require matching the capability registry's recorded StyleX version to the installed package.

## Validation

- `pnpm check:sync`
- Prettier
- `git diff --check`
- independent authority map, Matt-style challenge, and exact-file current-code FP/FN audit

No implementation, tooling, lint, wiki, or audit-table changes. No changeset.

### DIFF
--- a/docs/architecture/component-style-authoring.md
+++ b/docs/architecture/component-style-authoring.md
@@ -1,0 +1,202 @@
+---
+schema_version: 1
+template_version: 1
+kind: architecture
+id: architecture:component-style-authoring
+authority: draft
+archive_reason: null
+superseded_by: null
+approved_by: null
+approved_at: null
+owners: [cixzhang]
+applies_to: [packages/core/src/, packages/lab/src/]
+verified_by:
+  [
+    internal/stylex-capabilities/scan.mjs,
+    internal/eslint-plugin-astryx/no-border-shorthand.test.mjs,
+    internal/eslint-plugin-astryx/no-stylex-null-override.test.mjs,
+    internal/eslint-plugin-astryx/no-style-only-wrapper.test.mjs,
+  ]
+deciding_specs: []
+---
+
+# Component style authoring architecture
+
+## Purpose
+
+Astryx component styling should lower to valid maintained CSS, compose predictably
+with supported consumer styling, keep relational state scoped to the intended
+component, and avoid DOM wrappers with no runtime responsibility.
+
+This record owns five component-level authoring outcomes. It does not redefine
+semantic tokens, public theme targets, extensible prop axes, public styling props,
+icon resolution, bidirectionality, or React Server Component boundaries.
+
+## System model
+
+Component-owned styling begins in Core or Lab component source and lowers through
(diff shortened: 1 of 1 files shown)


cixzhang (commented):

Thanks — the authority boundaries are clear. I think the verification contract needs one more pass before this becomes current. The relational scanner only compiles through the global default marker, so it never proves nested-owner isolation. The wrapper rule also recommends `xstyle` for any relative import without proving the child has a compatible style-bearing surface.

Could we either add the scoped-marker/nested-owner and incompatible-child fixtures the Verification table names, or label those rows as current verification gaps?

[Reviewed by Robohands]

PR #6139, opened 2026-09-07
docs(spec): align adaptation component validation with root themes

imdreamrunner:
## Summary

- require adaptation component writes to use the same target, axis, value-domain, and extension validation as root `components`
- explicitly allow built-in values and values accepted by authoritative open primitive domains to appear only in an adaptation rule, without a matching root style declaration
- reserve effective-root presence for custom values whose validity comes from theme enrollment
- carry unresolved shared-validation results into adaptations unchanged, so future root-validator improvements tighten adaptations automatically rather than leaving a permanent exception
- record the owner decision as AST-012 DEC-5 and align the component-theming architecture invariant

## Responsibility boundary

Adaptations own conditions, ordering, cascade, and the additional no-conditional-enrollment rule. They do not own a separate visual-prop resolver, allowlist, or inference path.

The effective root is not a registry of every valid component value. A root entry is required only when it is the enrollment evidence that makes a custom value valid. Independently valid built-ins and open primitive values need no dummy root CSS.

## Why this is separate

This isolates the validation contract from the implementation in #5543. Shared root/adaptation value-domain improvements may land independently; adaptation behavior remains monotonic because it always consumes the shared root result.

Relevant discussion:

- https://github.com/facebook/astryx/pull/5543#issuecomment-5549881850
- https://github.com/facebook/astryx/pull/5543#issuecomment-5560831675
- https://github.com/facebook/astryx/pull/5543#issuecomment-5575977916

## Test plan

- `pnpm check:knowledge`
- commit hooks: sync, knowledge, package-boundary, changeset, CLI-structure, portable-script, i18n, fixture, and setup-contract checks
- verified the branch changes only `docs/specs/AST-012/spec.md` and `docs/architecture/component-theming-surface.md`

## Stack

Predecessor to #5543. No implementation changes are included here.

### DIFF
--- a/docs/architecture/component-theming-surface.md
+++ b/docs/architecture/component-theming-surface.md
@@ -30,7 +30,7 @@ verified_by:
     packages/core/src/theme/extensibleAxes.test.ts,
     packages/core/src/theme/derivedVarRegistry.test.ts,
   ]
-deciding_specs: [spec:AST-012/DEC-2, spec:AST-012/DEC-3]
+deciding_specs: [spec:AST-012/DEC-2, spec:AST-012/DEC-3, spec:AST-012/DEC-5]
 ---
 
 # Component theming surface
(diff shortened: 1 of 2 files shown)


cixzhang (changes requested):
This changes an accepted release contract rather than documenting implementation detail. AST-012 currently says adaptation rules are not an enrollment surface for component values; allowing opaque alias-backed values to pass through lets a closed axis emit unreachable CSS with no diagnostic or type augmentation. That is the exact failure the current #5543 review blocks.

Please do not weaken the accepted contract to fit the current validator. Either resolve alias-backed finite domains in shared root/adaptation validation (the intended role of #6061), or defer the adaptations API until that can be done safely. If the owner wants the pass-through limitation instead, that needs an explicit new owner decision before this spec can become current.

cixzhang (approved):
Approved. This now makes adaptation component writes reuse root component validation and keeps enrollment and future tightening owned by the root contract; adaptations add no parallel policy.

PR #6145, opened 2026-09-08
docs: define component size cascade architecture

cixzhang:
## Summary

- defines the missing cross-component owner for the standard `sm | md | lg` size cascade
- records explicit prop → nearest active provider → component default precedence
- keeps participation opt-in through current component/family contracts
- permits fixed provider scopes, intentional family overrides, and distinct nonstandard size axes
- assigns public transport compatibility to `architecture:public-component-api` and visual projection to theming/component owners
- includes current ButtonGroup, InputGroup, Toolbar, and SideNav provider scopes

## Audit routing

Resolves P41. It does not turn every `size` prop into cascade participation or require every provider owner to inherit an outer value.

## Owner proof and overlap

- public-component-api owns export reachability/naming, not cross-component runtime precedence
- react-component-runtime owns lifecycle/resource roles, not semantic size resolution
- component-theming-surface owns selector/extensibility and visual projection, not React context precedence
- button/input family records cover only their memberships
- #5762 and #6121 are adjacent Icon/theming work and remain untouched

## FP/FN audit

Initial bounded matrix: 12 correct, 1 false positive, 1 false negative, 1 untestable.

After correcting exported transport ownership, SideNav scope, and verification claims: **10 correct, 0 false positives, 0 false negatives, 5 untestable**. The untestable cases are explicitly recorded as promotion-evidence gaps: nested null reset; ButtonGroup/InputGroup precedence; Toolbar fixed scope; and representative owner-local visual projection.

This authority-only PR intentionally adds no tests. Focused fixtures are required before promotion to `current`.

## Validation

- `pnpm check:sync`
- Prettier
- `git diff --check`
- 311 focused existing tests in the independent audit
- repository pre-commit suite

No implementation, test, tooling, consumer-doc, wiki, or changeset change.

### DIFF
--- a/docs/architecture/component-size-cascade.md
+++ b/docs/architecture/component-size-cascade.md
@@ -1,0 +1,164 @@
+---
+schema_version: 1
+template_version: 1
+kind: architecture
+id: architecture:component-size-cascade
+authority: draft
+archive_reason: null
+superseded_by: null
+approved_by: null
+approved_at: null
+owners: [cixzhang]
+applies_to:
+  [
+    packages/core/src/SizeContext/,
+    packages/core/src/ButtonGroup/,
+    packages/core/src/InputGroup/,
+    packages/core/src/Toolbar/,
+    packages/core/src/SideNav/,
+  ]
+verified_by:
+  [
+    packages/core/src/SizeContext/SizeContext.test.tsx,
+    packages/core/src/ButtonGroup/ButtonGroup.test.tsx,
+    packages/core/src/InputGroup/InputGroup.test.tsx,
+    packages/core/src/Toolbar/Toolbar.test.tsx,
+    packages/core/src/SideNav/SideNav.test.tsx,
+  ]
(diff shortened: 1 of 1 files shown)


cixzhang (commented):

A future size change could silently make nested controls resolve the wrong value, because this authority is not connected to every owner it governs.

1. Current family records do not link this architecture, `applies_to` covers only the transport/providers, and several providers' inherited-versus-fixed choice is unstated. Please add a complete participant/provider ownership matrix plus the corresponding owner links and affected paths.

2. This head declares `authority: current` while “Promotion evidence gaps” says nested-null, ButtonGroup/InputGroup precedence, and Toolbar fixed-scope fixtures are required before promotion. Existing group tests only assert rendering, and Toolbar tests only its own class. Please add the focused fixtures, or explicitly record them as accepted current verification debt without claiming they already verify these invariants.

[Reviewed by Robohands]

PR #6147, opened 2026-09-08
docs(Banner): approve frame theme target

cixzhang:
## Why

[facebook/astryx#6116](https://github.com/facebook/astryx/pull/6116) established Banner’s frame, status surface, and content surface, but its merged contract still left frame theme ownership unresolved. The owner decision is now explicit: the frame is independently targetable, and its public name follows the visual role rather than DOM placement.

## What

- admit `banner-frame` as the additive target for the Banner frame
- assign `container` and `elevation` as its selector axes
- preserve `banner` on the primary Status surface and preserve every existing child target
- record the current missing runtime target as an implementation reachability gap
- record the current `banner.borderRadius` reachability gap: it affects the Status surface but not the ancestor frame or sibling Content surface
- reject `root` naming because it would promise outermost-DOM placement rather than stable anatomy

## Risk

Specification-only. Runtime implementation remains in [facebook/astryx#6007](https://github.com/facebook/astryx/pull/6007).

## Testing

- `pnpm exec vitest run packages/core/src/theme/themingTargets.test.ts`

### DIFF
--- a/packages/core/src/Banner/Banner.spec.md
+++ b/packages/core/src/Banner/Banner.spec.md
@@ -34,18 +34,18 @@ system_specs: []
 ## Intent
 
 Banner presents a persistent status message at the top of a page or section.
-This draft records the current painted surfaces and theming ownership so the
-outer-surface question can be reviewed against explicit facts. It documents the
-existing outer frame as anatomy, but does not choose a new target or change
-runtime behavior.
+This contract records the current painted surfaces and admits the Banner frame as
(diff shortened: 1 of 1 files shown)


cixzhang (commented):

A theme author who sets `banner.borderRadius` still gets a split card: the Status surface changes, while the Banner frame and sibling Content surface keep their fallback radius. This contract continues to describe that route as reaching all three painters. Please record the current reachability gap before this becomes the implementation contract.

[Reviewed by Robohands]

PR #6149, opened 2026-09-08
docs(Collapsible): define chevron position contract

cixzhang:
## Summary

- creates the missing canonical `component:Collapsible` record
- records `chevronPosition?: 'start' | 'end'` as the public axis, defaulting to `end`
- preserves trailing behavior: collapsed `end` points down, expanded `end` points up
- defines leading behavior: collapsed `start` points inward toward content and mirrors under RTL; expanded `start` points down
- preserves group default → item override → nested-scope reset precedence
- preserves full-row trigger-label layout for composed triggers
- explicitly rejects `chevronPlacement`, `none`, `hasChevron`, caller-provided cue content, and renderer escape hatches
- changes no existing non-chevron behavior, including content reveal motion

## Owner decision

Cindy approved the position axis and Ernest's rationale/behavior as evidence. Cindy's explicit naming/value/direction ruling controls any inconsistency.

The contract is complete without any PR reference. #5993 is implementation evidence only.

## Compatibility

Additive optional API. Omission preserves the released trailing chevron and current no-prop rendering.

## Overlap receipt

- #5993 is the direct live implementation and uses the old `chevronPlacement` / `none` proposal; it must be revised after this record lands.
- #5567 changes content reveal motion, which this contract explicitly leaves unchanged/unsettled by the chevron decision.
- #5939 touches Collapsible consumer docs mechanically and does not own the behavior.
- #5933 is already merged and is evidence for full-row label layout only.

## Validation

- Astryx branch bootstrap and `astryx component Collapsible --dense`
- Prettier
- `pnpm check:sync`
- `git diff --check`
- repository pre-commit suite

No implementation, test, consumer-doc, tooling, wiki, or changeset change.

### DIFF
--- a/packages/core/src/Collapsible/Collapsible.spec.md
+++ b/packages/core/src/Collapsible/Collapsible.spec.md
@@ -1,0 +1,210 @@
+---
+schema_version: 3
+template_version: 4
+kind: component
+id: component:Collapsible
+authority: current
+archive_reason: null
+superseded_by: null
+approved_by: cixzhang
+approved_at: 2026-09-07
+owners: [cixzhang, imdreamrunner]
+review_triggers: [public-api, behavior, layout, accessibility]
+verified_by:
+  [
+    packages/core/src/Collapsible/Collapsible.test.tsx,
+    packages/core/src/Collapsible/CollapsibleGroup.test.tsx,
+  ]
+modules: []
+families: []
+design_specs: []
+architecture:
+  [architecture:public-component-api, architecture:react-component-runtime]
+contributing: [contributing:api-conventions]
+system_specs: [spec:AST-002/DEC-1]
+---
+
+# Collapsible component contract
+
+## Intent
+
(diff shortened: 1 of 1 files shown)


cixzhang (commented):

The contract captures the agreed scope, but two lines make the canonical requirement unreliable:

- FR1’s unescaped `|` is parsed as a table delimiter, so GitHub splits the prop signature and drops its Basis.
- The RTL representative state says collapsed `start` points `inline-start`. From logical start, the label/content is toward `inline-end`; PR #5993’s mirrored chevron points that way too.

Please fix both so the rendered record keeps `chevronPosition?: 'start' | 'end'` in one invariant cell and consistently points collapsed `start` toward logical inline-end.

[Reviewed by Robohands]

cixzhang (commented):
Semantic verdict: approve

Thanks—both fixes are in. GitHub preserves FR1’s signature and Basis, RTL collapsed `start` points inward, and the agreed contract is otherwise unchanged.

[Reviewed by Robohands]

Issue #6132, opened 2026-09-07
ClickableCard: the element carrying the button role cannot receive a pointer activation


cixzhang:
## Who is affected

Anyone whose tooling acts on the accessible object rather than on the pixels
under a mouse — speech-input users saying a control's name, and assistive
technology or automation that dispatches a click at the element it is told is
the button.

## What happens

`ClickableCard` puts the button role and the accessible name on a visually
hidden `<button>` inside the card, and lets the card surface handle pointer
input separately. That hidden button renders as a 1×1 clipped box, and the
card's own content paints over it.

So the element that *is* the button, as far as the accessibility tree is
concerned, cannot itself receive a pointer activation. A real browser refuses:
it never finds the element visible, stable, and able to receive a pointer event.

```
await page.getByRole('button', {name: 'Open billing settings'}).click();
// Timeout: the element is covered by the card's own content
```

## Reproduction

Storybook: `a11y/Button pattern` → `Clickable card`. The card is
`<ClickableCard label="Open billing settings" onClick={…}>` with a paragraph of
content.

Inspect the control that carries `role=button`: its bounding box is 1×1, and
`document.elementFromPoint` at its centre returns the card's paragraph.

## What still works

A sighted mouse user is fine — they click the card surface, which handles the
event. Keyboard activation is fine: the hidden button takes focus, and Enter and
Space both run the action. This is specifically about a pointer activation
targeted at the element that carries the role.

## Suggested direction

Not proposing an implementation here — the split between "the surface that takes
the click" and "the element that carries the semantics" is deliberate, and
whichever way it is reconciled is a design-system decision. Two shapes that
would both close it: give the hidden control the card's own geometry so a click
at its coordinates reaches it, or move the role and name onto the interactive
surface itself.

## How this was found

By the shared button-pattern accessibility contract
(`internal/a11y-spec/src/patterns/button.ts`), which holds every component that
adopts the APG button pattern to the same outcomes. This one fails
`button.action.runs-on-pointer`, and is recorded as a known failure against this
issue rather than being worked around.

cixzhang:
Closing until current Astryx authority settles which ClickableCard surface owns role, name, focus, and pointer activation. This should not invite implementation while that design-system direction is unresolved. Reopen after an accepted public specification records the expected ownership model.

PR #6150, opened 2026-09-08
test(a11y): point the ClickableCard button binding at the card surface

kyu-rong:
## Summary

The ClickableCard rows in the button contract aimed the pointer at the visually hidden `<button>` that carries the role, so `runs-on-pointer` and `survives-an-aborted-press` sat as known failures. The card surface is the pointer target by design; the hidden control owns role, name, and keyboard focus. The binding now declares the surface as `pointerTargetSelector`, the same model the checkbox contract uses for SelectableCard (#6155), and both known-failure records are removed.

`ClickableCard.tsx` is unchanged from `main`. Which surface should ultimately own pointer activation is left to the specification, per #6132.

## Changes

- `Button.a11y.states.ts`, `Button.a11y.renders.tsx`, `Button.a11y.chromium.spec.ts`: add `pointerTargetSelector` and pass it to the Chromium harness for the two ClickableCard states
- `Button.a11y.known-failures.ts`: remove the two ClickableCard records
- `ClickableCard.test.tsx`: jsdom cases for keyboard activation, nested input, text selection, disabled inertness
- `rtl-audit/verified-not-applicable.json`: `core/ClickableCard` entry

## Test Plan

- [x] `pnpm test:a11y-contract` (Button spec): 20 passed, ClickableCard known-failure 0 · unexpected-pass 0
- [x] `vitest run packages/core/src/ClickableCard/ClickableCard.test.tsx`: 17 passed
- [x] `git diff origin/main -- packages/core/src/ClickableCard/ClickableCard.tsx`: empty
- [ ] CI `pr-a11y`, `pr-rtl`

## Related Issues

Refs #6132, #6155


cixzhang (changes requested):
Thanks for fixing the unreachable semantic control—the pointer and keyboard paths now fire once, and Chromium keeps the same link/button name.

I’m requesting changes because the full-card control breaks existing behavior:

- `isDisabled` href cards still follow their URL on a real pointer click.
- The overlay captures raw nested buttons, read-only inputs, and text selection. Those controls remain in the accessibility tree but cannot receive pointer input.
- `onClick` now receives the invisible `BUTTON`/`A` as `currentTarget`, rather than the card surface, changing the released callback behavior.

Could we keep the role-bearing target pointer-operable while preserving disabled inertness, existing child/content interactions, and the callback contract? `pr-rtl` also remains red.

[Reviewed by Robohands]

Kyujenius:
All three confirmed — one cause: a full-card control competes with the content for the same pixels. The issue's other shape (role on the surface) trips axe's `nested-interactive` around a nested `Button`.

image

The revision takes the third shape (C): the control keeps its role and name and occupies only the card's top padding band, so nothing the consumer renders is covered and the click wiring is exactly `main`'s. Since the issue framed this as a design-system decision, I'd like your call on it. Details and results in the PR body.

One caveat: the band is 11–12px tall, under the 24px target minimum on its own; I'm relying on the card surface as the equivalent target. If you'd rather a different region of the frame, or the contract measure this differently, happy to adjust.

cixzhang (approved):
Thanks—this fixes the three blockers from the earlier head: disabled links stay inert, nested content remains interactive, and surface callbacks keep their target. Exact-head CI and Chromium are green.

[Reviewed by Robohands]

Kyujenius:
@cixzhang Thanks for your quick review!

cixzhang (changes requested):
Thanks—pointer activation now reaches the semantic control without covering nested content, and disabled cards remain inert. One public callback path still changes: a click in the new top padding band now hits the invisible control. In Chromium, action cards pass `BUTTON` rather than the card `DIV` as `currentTarget`; link cards skip `onClick` entirely and navigate even when that callback would call `preventDefault()`. The same coordinates on `main` use the card-surface callback. Could this keep the role-bearing control reachable while every point on the card preserves the documented callback target and cancellation behavior?

[Reviewed by Robohands]

Kyujenius:
@cixzhang 
Fixed in the component: a press on the control (band or Enter/Space) now runs `onClick` through the card's own handler, so `currentTarget` is the card `DIV` on every path and `preventDefault()` cancels link navigation. 

Verified against a local `main` build on the same story — only the two keyboard rows differ (they now follow the documented target; `main` handed `BUTTON`/skipped `onClick`).

cixzhang (changes requested):
Thanks—the direct semantic control, keyboard, disabled, nested-control, text-selection, and RTL paths are clean at this head.

One ordinary link-card path still invokes the consumer twice. Clicking the card content calls `onClick`, proxies `link.click()`, then the bubbling control click reaches `handleClick` and calls `onClick` again (`packages/core/src/ClickableCard/ClickableCard.tsx:324`). Exact-head Vitest and Chromium both observe two calls; the previously approved `cfe08f0b` observes one. Builders can therefore run analytics or other side effects twice from a single user click. Could surface activation preserve the card `DIV` as `currentTarget` while invoking `onClick` exactly once?

[Reviewed by Robohands]

Kyujenius:
Fixed: the hook's proxied `link.click()` is now ignored when it bubbles back, so a surface click runs `onClick` once (test fails with the guard reverted). Chromium vs a local `main` build: one call on action, link, and plain-link cards; other paths unchanged.

docs/families/input-fields.md
Kyujenius, line 1: 71a2d7e restores `docs/families/input-fields.md` to `main`'s bytes. Merging `main` staged it and the
pre-commit Prettier hook reformatted it — the copy that landed in #6183 isn't Prettier-clean, and CI
doesn't run Prettier, so the hook only catches it on the next local commit that touches the file.

kentonquatman (approved):
Design looks good. Thanks @Kyujenius!

Kyujenius:
@kentonquatman Thanks!

cixzhang (approved):
Thanks — this now tests pointer activation on the card surface while keeping the hidden control as the semantic and keyboard subject. Focused tests and exact-head CI are clean.

Issue #6151, opened 2026-09-08
arrowUp icon key carries upload, send, and directional meanings


athz:
## Problem

The built-in `arrowUp` icon name currently carries three different semantic responsibilities:

| Consumer | Use | Meaning |
| --- | --- | --- |
| `FileInput` | upload affordance | action: upload |
| `ChatSendButton` | default send glyph | action: send |
| sortable `Table` | ascending indicator | direction: ascending |

The directional meaning is coherent with `arrowDown` and `arrowsUpDown`. The upload and send meanings only happen to share the same artwork today.

This becomes observable when a theme customizes the registry. Mapping `arrowUp` to an upload-tray glyph also changes the chat send button and breaks the visual pairing between the table's ascending and descending indicators. A single registry key therefore cannot be customized predictably across its current callsites.

## Current architecture

The current icon architecture separates shared semantic names from component-owned roles. That makes the boundary important:

- `arrowUp` should remain a directional name.
- Upload and send are reusable UI actions, not FileInput- or Chat-specific implementation details.
- Component-owned slot mappings can still be added later if a theme needs to remap one component role independently; that is a separate layer from the shared semantic vocabulary.

## Proposed contract

Add two shared semantic names:

- `upload`
- `send`

Both would initially resolve to the existing upward-arrow artwork, so default rendering would not change. FileInput would use `upload`, ChatSendButton would use `send`, and sortable Table would keep `arrowUp`.

This lets a theme provide an upload-tray glyph without changing send or sort, and a send glyph without changing upload or sort.

### Keep `arrowUp` for all three meanings

No source change, but registry customization remains coupled across unrelated actions and direction.

### Add only component-owned slots

Slots are useful when one component role needs a different shared meaning. They do not by themselves provide shared `upload` or `send` meanings for themes to map to appropriate artwork. These two actions are expected to recur across components, so they belong in the shared semantic vocabulary first.

## Scope and non-goals

- No default visual change.
- No change to `arrowDown` or `arrowsUpDown`; their current uses remain directional.
- No component-slot implementation in this proposal.
- No commitment to a specific upload or send drawing; themes continue to own artwork.

## Suggested sequence

1. Settle the shared-name contract in a spec-only pull request.
2. Add the two names and default artwork aliases.
3. Migrate FileInput and ChatSendButton to the new names with focused registry tests and consumer documentation.

athz:
[...]

PR #6153, opened 2026-09-08
docs(spec): separate upload, send, and arrow icon semantics

athz:
## Why

The built-in `arrowUp` icon key currently means three different things: upload in FileInput, send in ChatSendButton, and ascending direction in sortable Table. A theme that customizes the key for one action also changes the other action and the directional sort pair.

The current icon architecture distinguishes shared semantic names from component-owned roles. Upload and send are reusable actions rather than component implementation details, while `arrowUp` should remain directional.

Refs #6151.

## What

Add the draft AST-032 system spec to propose:

- distinct shared `upload` and `send` names
- `arrowUp` remaining the ascending/upward directional name
- FileInput, ChatSendButton, and sortable Table adopting their respective meanings in a later implementation
- no default visual change
- exact compatibility across the core default registry and every shipped theme registry

The shipped-theme requirement matters because icon resolution is an exact name lookup: a theme's existing `arrowUp` artwork does not automatically become its `upload` or `send` artwork. The implementation must initialise both new names from each registry's current `arrowUp` entry and prove that preservation.

## Non-goals

- no component-owned slot implementation
- no change to `arrowDown` or `arrowsUpDown`
- no prescribed upload or send drawing
- no runtime or package changes in this PR

## Risk

Documentation only. The record remains `authority: draft` until an authorised owner approves the exact head.

## Testing

- `pnpm check:knowledge`
- `pnpm check:repo`
- `pnpm exec prettier --check docs/specs/AST-032/spec.md`
- `git diff --check`

No changeset: this PR does not publish package behavior.

### DIFF
--- a/docs/specs/AST-032/spec.md
+++ b/docs/specs/AST-032/spec.md
@@ -1,0 +1,119 @@
+---
+schema_version: 1
+template_version: 1
+kind: system-spec
+id: spec:AST-032
+authority: draft
+archive_reason: null
+superseded_by: null
+approved_by: null
+approved_at: null
+phase: proposed
+owners: [cixzhang, imdreamrunner]
+affects_architecture: [architecture:icon-resolution-and-component-slots]
+affects_families: []
+affects_contributing: []
+affects_consumer_docs: [Icon, FileInput, ChatSendButton]
+---
+
+# Shared upload and send icon semantics system spec
+
+## Intent
+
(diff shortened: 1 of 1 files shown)


cixzhang (approved):
Approved: upload and send remain component-owned icon roles, with independent componentIcons slots and arrowUp as the unchanged fallback.

Issue #6156, opened 2026-09-08
SelectableCard disabled state is not keyboard-focusable as documented


cixzhang:
## User impact

A keyboard or screen-reader user cannot tab to a disabled `SelectableCard` to discover that the option exists and is unavailable, despite the component documentation promising that disabled cards remain focusable and expose `aria-disabled`.

## Current behavior

The role-bearing `<input type="checkbox">` receives the native `disabled` attribute. Chromium removes it from the tab sequence. The component documentation says disabled cards remain focusable with `aria-disabled`, but the rendered control does not.

## Expected behavior

The disabled state and public documentation should agree. If the documented focusable-disabled behavior remains the contract, the checkbox must stay in the tab sequence, expose unavailability, and remain inert.

## Reproduction

1. Render `SelectableCard` with `isDisabled`.
2. Start focus before the card and press Tab.
3. Focus never reaches the role-bearing checkbox.

This issue tracks component remediation separately from the accessibility contract migration.

cixzhang:
Closing as an audit correction. The current SelectableCard docs promise only that `isDisabled` disables the card; they do not adopt a focusable-disabled state. Native disabled checkbox behavior is therefore within the current contract, and the accessibility binding now records it as outside the tab sequence rather than as standards debt.

cixzhang:
Correction: the source-level `isDisabled` prop contract does explicitly promise that disabled cards remain focusable with `aria-disabled`. The checkbox migration now tracks that promise through a separate advisory expectation, rather than presenting focusability of disabled controls as a WCAG 2.1.1/2.1.2 requirement. Reopening until the public contract and implementation agree.

cixzhang:
Closing until an accepted component specification settles whether a disabled SelectableCard remains keyboard-focusable with aria-disabled or uses native disabled behavior. The implementation PR should stay paused until that public contract exists.

Issue #6155, opened 2026-09-08
SelectableCard: role-bearing checkbox cannot receive pointer activation


cixzhang:
## User impact

A speech-input user or assistive technology that activates `SelectableCard` through its exposed checkbox object cannot trigger the selection change, even though a sighted mouse user can click the card surface.

## Current behavior

`SelectableCard` exposes role, name, and checked state on a visually hidden 1×1 `<input type="checkbox">`, while pointer handling lives on the outer card. In Chromium, the role-bearing checkbox is covered by the card content and cannot receive a pointer click or a cancellable pointer press.

## Expected behavior

The element exposed as the checkbox should be a reachable activation target, or the interactive card surface should own the same checkbox semantics, so pointer activation addressed through the accessible object reaches the control users are told exists.

## Reproduction

1. Render a selectable card with `label="Analytics"`.
2. Locate the element exposed as `role=checkbox` in Chromium.
3. Dispatch a normal pointer click to that element.
4. The browser cannot find it visible and able to receive the pointer event; the visible card surface still works when clicked directly.

This issue tracks component remediation separately from the accessibility contract migration.

cixzhang:
Closing because the accessibility contract was initially targeting the hidden semantic checkbox for pointer input. The corrected binding now drives the public card surface for pointer interaction while reading role, name, and checked state from the checkbox. Chromium proves the card changes in both directions and supports pointer cancellation; no component defect remains.

Issue #6154, opened 2026-09-08
CheckboxListItem does not expose its description to assistive technology


cixzhang:
## User impact

The browser accessibility node for a described `CheckboxListItem` has no distinct description, so accessibility consumers cannot identify the visible supporting text as the explanation for that checkbox.

## Current behavior

`CheckboxListItem` renders `description` in the list row, but the nested checkbox has no `aria-describedby` relationship to that content. Its computed browser accessibility description is empty.

The visible text may still be discoverable elsewhere in the row. This issue is specifically about exposing it as the checkbox's description; it does not claim what any assistive technology announces.

## Expected behavior

When a checkbox item has supporting descriptive text, the role-bearing checkbox should expose that exact text as its browser-computed accessibility description without folding it into the accessible name.

## Reproduction

1. Render a standalone or collection `CheckboxListItem` with `description`.
2. Inspect the nested checkbox in Chromium's accessibility tree.
3. Its role and name are present, but its computed description is empty.

The accessibility contract migration records the current failure but does not change the component.

PR #6157, opened 2026-09-08
test(a11y): reusable checkbox-pattern contract across four adopters

cixzhang:
# Checkbox: reusable accessibility contract across four adopter surfaces

## User impact

People using a keyboard, screen reader, or speech input should encounter the same checkbox contract whether the control is a form field, a list item, a menu item, or an independently selectable card. This change encodes that shared contract once and runs it against every current Astryx adopter.

**No component behavior changes.** Existing defects are exact runnable known failures with focused public issues.

## Contract

The contract adopts the [APG Checkbox pattern](https://www.w3.org/WAI/ARIA/apg/patterns/checkbox/) against WCAG 2.2 A/AA. It also recognizes `menuitemcheckbox` inside the [APG Menu pattern](https://www.w3.org/WAI/ARIA/apg/patterns/menubar/); the menu contract still owns arrow navigation, roving focus, and dismissal.

| Expectation | User outcome | Source | Evidence | Enforcement |
| `checkbox.role.exposed` | The browser exposes the adopted checkbox-bearing role. | WCAG 4.1.2; APG Checkbox/Menu roles | accessibility tree | required |
| `checkbox.name.exposed` | The user can identify the choice. | WCAG 4.1.2, 3.3.2; APG label | accessibility tree | required |
| `checkbox.name.matches-visible-label` | Speech input can use the visible words. | WCAG 2.5.3 | tree + real browser | required |
| `checkbox.state.exposed` | Unchecked, checked, and partially checked states match the rendered state. | WCAG 4.1.2; APG checked state | accessibility tree | required |
| `checkbox.description.resolvable` | References resolve to the exact intended supporting text. | WCAG 1.3.1; APG description | DOM | required |
| `checkbox.description.exposed` | The browser computes that exact text as the distinct description. | WCAG 4.1.2 | accessibility tree | required |
| `checkbox.disabled.exposed` | An unavailable checkbox says so. | WCAG 4.1.2 | accessibility tree | required |
| `checkbox.disabled.not-exposed` | An available checkbox does not expose a false disabled state. | WCAG 4.1.2 | accessibility tree | required |
| `checkbox.readonly.declared` | A read-only checkbox declares that its value cannot change. | WCAG 4.1.2 | DOM | required |
| `checkbox.readonly.not-declared` | An editable checkbox does not expose a false read-only state. | WCAG 4.1.2 | DOM | required |
| `checkbox.required.declared` | A required checkbox declares the requirement in markup. | WCAG 4.1.2 | DOM | required |
| `checkbox.required.not-declared` | An optional checkbox does not expose a false required state. | WCAG 4.1.2 | DOM | required |
| `checkbox.invalid.exposed` | A checkbox in error is exposed as invalid. | WCAG 4.1.2 | accessibility tree | required |
| `checkbox.invalid.not-exposed` | A valid checkbox does not expose a false invalid state. | WCAG 4.1.2 | accessibility tree | required |
| `checkbox.state.pointer-round-trip` | Pointer activation changes state in both directions; mixed resolves and changes again. | APG checked state; supports WCAG 4.1.2 | browser + tree | advisory |
| `checkbox.state.space-round-trip` | Direct checkboxes change with Space in both directions. | WCAG 2.1.1; APG Space | browser + tree | required |
| `checkbox.state.survives-an-aborted-press` | A press released away from the control can be taken back. | WCAG 2.5.2 | browser + tree | required |
| `checkbox.state.keeps-focus-on-change` | Changing a direct checkbox does not throw focus elsewhere. | WCAG 3.2.2 | browser + tree | required |
| `checkbox.focus.reachable-and-escapable` | Tab reaches and leaves operable direct checkboxes. | WCAG 2.1.1, 2.1.2 | real browser | required |
| `checkbox.focus.declared-inoperable-reachable` | A binding that promises an inoperable checkbox stays reachable delivers that promise. | AST-021 FR7 | real browser | advisory |
| `checkbox.state.inoperable` | An inoperable state changes through neither pointer nor owned keyboard input. | APG checked state; supports WCAG 4.1.2 | browser + tree | advisory |

Eighteen expectations are required. `state.pointer-round-trip`, `focus.declared-inoperable-reachable`, and `state.inoperable` are advisory: APG does not independently require pointer activation or universal inertness, and no current component/system record adopts focusability of every inoperable checkbox as a conformance requirement. The declared-focusability check instead preserves and exposes public binding promises under AST-021 FR7 without relabeling them as WCAG. The contract uses binding-provided role facts so `checkbox` and `menuitemcheckbox` cannot be confused, and a separate pointer-target seam for components such as SelectableCard whose public pointer surface differs from the semantic node. Visible-label overrides are locators measured by the same browser visibility algorithm, not precomputed strings.

## Bindings and states

Thirty-three checked-in states cover every current checkbox-bearing part:

- `CheckboxInput`: unchecked, checked, mixed, described, hidden label, native disabled, focusable disabled with reason, loading, explicit read-only, handlerless controlled read-only, explicit required, inherited-required without native invalidity, invalid, warning, and success. Loading is included because it changes operability; the component-owned `isLoading` → `aria-busy` mapping remains local.
- `CheckboxListItem`: standalone and collection states; unchecked, checked, mixed, described, ordinary string label, rich label without its required plain-text equivalent, disabled, loading, explicit read-only, handlerless standalone read-only, and group-disabled with a discoverable reason.
- `DropdownMenuCheckboxItem`: unchecked, checked, secondary text participating in the visible name, handlerless inert-but-exposed-available, and disabled inside an opened menu. Menu-owned keyboard/focus behavior and distinct-description policy remain outside this contract.
- `SelectableCard`: independent multi-select unchecked, checked, and disabled states; the disabled state records the component's public focusable-disabled promise separately from WCAG-required operable focus.

Every state has a checked-in `a11y/Checkbox pattern` Storybook story. The Chromium lane reloads each story for each expectation so state mutations cannot leak between checks.

## Known failures

Known failures still execute and remain failures. Every record names its exact standards reference and matches one complete failure message. A different error, state, layer, or binding remains a `fail` rather than being absorbed by the record; required failures gate, while advisory failures remain report-only under AST-020 FR9. Every record must match exactly one executed result, and a fix becomes an `unexpected-pass` until that exact stale record is removed.

| Gap | Exact records | Issue |
| --- | --- | --- |
| `CheckboxListItem` renders supporting text but does not attach it to the nested checkbox. | `description.resolvable`, `description.exposed` for `list-item-described` | [#6154](https://github.com/facebook/astryx/issues/6154) |
| `CheckboxListItem` accepts a rich label without an equivalent plain-text name and falls back to the generic name “Checkbox.” | `name.matches-visible-label` for `list-item-rich-label-missing-name` | [#6161](https://github.com/facebook/astryx/issues/6161) |
| Handlerless `CheckboxListItem` remains exposed as editable. | `readonly.declared` for `list-item-handlerless-read-only` | [#6163](https://github.com/facebook/astryx/issues/6163) |
| Handlerless `CheckboxInput` remains exposed as editable. | `readonly.declared` for `input-handlerless-read-only` | [#6165](https://github.com/facebook/astryx/issues/6165) |
| Handlerless `DropdownMenuCheckboxItem` remains exposed as available even though its controlled value cannot change. | `disabled.exposed` for `menu-item-handlerless-inert` | [#6166](https://github.com/facebook/astryx/issues/6166) |
| `SelectableCard` promises a focusable disabled state but renders a native disabled checkbox outside the tab sequence. | advisory `focus.declared-inoperable-reachable` for `card-disabled` | [#6156](https://github.com/facebook/astryx/issues/6156) |

No component gap is fixed here. An initial pointer-target finding against SelectableCard was invalid: the first binding clicked the hidden semantic node instead of the public card surface. The harness now models those as separate public seams, Chromium proves pointer activation and cancellation through the card, and [#6155](https://github.com/facebook/astryx/issues/6155) is closed with that correction.

## Mutation proof

The contract has conforming fixtures for checkbox and menu-checkbox roles, unchecked, checked, mixed, described, hidden-label, disabled, focusable-disabled, pending, read-only, required, and invalid states. Every expectation has both a conforming pass and a deliberately violating fixture, and each mutation test asserts the exact expected failure detail so an unrelated fixture or harness error cannot count as proof.

Mutations cover wrong or absent role, missing or mismatched name, wrong binary or mixed state, dangling, empty, or wrong-but-nonempty description targets, missing and falsely added disabled/read-only/required/invalid states, inert or one-way activation, pointer-only behavior, pointer-down activation, focus displacement, unreachable focus, trapped Tab, a binding-declared focusable unavailable control removed from the tab sequence, an unavailable control that changes by pointer, and an unavailable control that changes only with Space. Conforming visibility fixtures also prove that a pointer-transparent visible label remains visible while the same label fully clipped remains not applicable.

## Ownership, exemptions, and evidence boundaries

- Component-local tests retain callback, Action, form, composition, menu navigation, styling, focus-ring, forced-colors, and announcement behavior.
- Exact local assertions now owned at the same or stronger shared layer were removed. CheckboxInput's component-specific guarantee that supporting text stays out of its accessible name remains local, but now runs against Chromium's computed accessibility tree instead of jsdom.
- Table selection is an explicit composed callsite of the already-bound `CheckboxInput`; Table owns selection labels and group context, while the shared input states cover hidden labels and mixed state.
- MultiSelector's inert, hidden checkbox marker is an explicit decorative exclusion; its exposed `role="option"` remains owned by the option contract.
- Visual/theme, page/composition, descriptive wording, status messages, and real-AT outcomes remain explicit exemptions with named owners and verification methods.
- `3.2.2 On Input` is part-encoded: the contract proves direct checkbox activation keeps focus; callers own navigation or page rewrites caused by their change handlers.
- SelectableCard's RTL audit is verified not applicable: its ring and overlay are symmetric, it uses logical positioning, and it has no directional keyboard interaction.

No real-AT trigger applies: this PR changes tests only and claims DOM, browser-computed semantics, and real-browser interaction—not speech, braille, announcement timing, virtual-cursor entry, or interoperability. AST-013 is followed by running browser-owned checks in real Chromium; no browser-specific support claim is made.

## Validation

- `@astryxdesign/a11y-spec` and `@astryxdesign/core` typechecks
- 211 package-level contract and runner jsdom tests
- 99 focused CheckboxInput, DropdownMenu, SelectableCard, and binding jsdom tests
- 185 real-Chromium contract, mutation, and binding tests
- Storybook production build
- scoped RTL audit: 1 measured, 2 verified not applicable, 0 gaps, 0 stale

### DIFF
--- a/apps/storybook/rtl-audit/verified-not-applicable.json
+++ b/apps/storybook/rtl-audit/verified-not-applicable.json
@@ -63,5 +63,14 @@
   {
     "component": "core/IconButton",
     "reason": "IconButton is a thin wrapper that renders Button with `isIconOnly`, adding no layout, no positioning, and no directional affordance of its own, so it inherits `core/Button`'s verification above exactly. Its single icon is centred in a square control, which has no reading direction to mirror."
+  },
+  {
+    "component": "core/SelectableCard",
+    "reason": "SelectableCard renders content in normal flow inside a symmetric Card. Its hover overlay and selected ring cover all four edges equally, its source uses no physical left/right positioning or directional glyphs, and its checkbox interaction uses only pointer, Space, Enter, and Tab rather than inline-direction keys. Every current story therefore has no component-owned visual or behavioral RTL difference for the audit to measure.",
+    "evidence": {
+      "apps/storybook/stories/SelectableCard.stories.tsx": "sha256:35704da575bb80fc580309a5d4182ab4569abbd1449c665f69f1f00767b5790b",
+      "packages/core/src/SelectableCard/SelectableCard.tsx": "sha256:5fa8fcf087296e33415a2696b91e478608b737506f26775715132a3a63e9218e",
+      "packages/core/src/SelectableCard/index.ts": "sha256:9cf677856807a09ee444ce68b21118bca9729592485f892da7e0bb64ee1d7501"
+    }
   }
 ]
--- a/apps/storybook/stories/CheckboxA11y.stories.tsx
+++ b/apps/storybook/stories/CheckboxA11y.stories.tsx
@@ -55,9 +55,14 @@ export const ListItemMixed = storyFor('list-item-mixed');
 export const ListItemDescribed = storyFor('list-item-described');
 export const ListItemDisabled = storyFor('list-item-disabled');
 export const ListItemLoading = storyFor('list-item-loading');
+export const ListItemReadOnly = storyFor('list-item-read-only');
+export const ListItemGroupDisabledWithMessage = storyFor(
+  'list-item-group-disabled-with-message',
+);
 
 export const MenuItemUnchecked = storyFor('menu-item-unchecked');
 export const MenuItemChecked = storyFor('menu-item-checked');
+export const MenuItemDescribed = storyFor('menu-item-described');
 export const MenuItemDisabled = storyFor('menu-item-disabled');
 
 export const CardUnchecked = storyFor('card-unchecked');
--- a/apps/storybook/stories/CheckboxA11y.stories.tsx
+++ b/apps/storybook/stories/CheckboxA11y.stories.tsx
@@ -46,8 +46,14 @@ export const InputDisabled = storyFor('input-disabled');
 export const InputDisabledWithMessage = storyFor('input-disabled-with-message');
 export const InputLoading = storyFor('input-loading');
 export const InputReadOnly = storyFor('input-read-only');
+export const InputHandlerlessReadOnly = storyFor('input-handlerless-read-only');
 export const InputRequired = storyFor('input-required');
+export const InputInheritedRequiredValid = storyFor(
+  'input-inherited-required-valid',
+);
 export const InputInvalid = storyFor('input-invalid');
+export const InputWarning = storyFor('input-warning');
+export const InputSuccess = storyFor('input-success');
 
 export const ListItemUnchecked = storyFor('list-item-unchecked');
 export const ListItemChecked = storyFor('list-item-checked');
@@ -68,6 +74,8 @@ export const ListItemGroupDisabledWithMessage = storyFor(
 
 export const MenuItemUnchecked = storyFor('menu-item-unchecked');
 export const MenuItemChecked = storyFor('menu-item-checked');
+export const MenuItemDescribed = storyFor('menu-item-described');
+export const MenuItemHandlerlessInert = storyFor('menu-item-handlerless-inert');
 export const MenuItemDisabled = storyFor('menu-item-disabled');
 
 export const CardUnchecked = storyFor('card-unchecked');
--- a/apps/storybook/stories/CheckboxA11y.stories.tsx
+++ b/apps/storybook/stories/CheckboxA11y.stories.tsx
@@ -1,0 +1,65 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+/**
+ * @file CheckboxA11y.stories.tsx
+ * @input Uses the shared Checkbox binding inventory and render map
+ * @output One checked-in reproduction story for every Checkbox contract state
+ * @position Stable real-browser fixtures required by AST-009 FR30 and AST-021.
+ */
+
+import type {Meta, StoryObj} from '@storybook/react';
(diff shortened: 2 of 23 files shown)


cixzhang (commented):
Semantic verdict: approve

This gives keyboard users and accessibility consumers consistent checkbox regression protection across all four adopters without changing shipped behavior. Thanks, this is good.

[Reviewed by Robohands]

PR #6158, opened 2026-09-08
fix(core): keep disabled SelectableCard focusable with aria-disabled

### LINKED ISSUE #6156: SelectableCard disabled state is not keyboard-focusable as documented

Geervan:
Fixes #6156

Fixes an issue where `SelectableCard` with `isDisabled` rendered native `disabled` on the hidden `<input type="checkbox">`, removing it from the tab sequence. The component now preserves focusability with `aria-disabled="true"` to match its documented contract.
## Changes
- **`SelectableCard.tsx`**:
  - Replaced native `disabled` with `aria-disabled="true"` on the hidden `<input type="checkbox">`.
  - Added `form={isDisabled ? '' : undefined}` to exclude the focusable control from form submissions.
  - Gated `<input onChange>` and suppressed <kbd>Space</kbd> / <kbd>Enter</kbd> in `handleKeyDown` when disabled.
- **`SelectableCard.stories.tsx`**:
  - Updated Storybook description to reflect that disabled cards remain focusable.
- **`SelectableCard.test.tsx`**:
  - Added tests verifying `aria-disabled`, focus retention, and interaction blocking.
## Verification
- `vitest run packages/core/src/SelectableCard/` passed (16/16).
- Verified in Storybook that disabled cards receive keyboard focus via <kbd>Tab</kbd> and cannot be toggled.

### DIFF
--- a/.changeset/selectable-card-focusable-disabled.md
+++ b/.changeset/selectable-card-focusable-disabled.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] SelectableCard: keep disabled cards in sequential focus navigation with aria-disabled, form detachment, and gated interaction handlers.
+
+@Geervan
--- a/apps/storybook/stories/SelectableCard.stories.tsx
+++ b/apps/storybook/stories/SelectableCard.stories.tsx
@@ -234,7 +234,7 @@ export const Disabled: Story = {
     docs: {
       description: {
         story:
-          '`isDisabled` suppresses toggle, hover, focus. Accent border remains visible on disabled+selected cards.',
+          '`isDisabled` suppresses toggle and hover; the control remains focusable and exposes aria-disabled. Accent border remains visible on disabled+selected cards.',
       },
     },
   },
--- a/packages/core/src/CheckboxInput/__tests__/Checkbox.a11y.chromium.spec.ts
+++ b/packages/core/src/CheckboxInput/__tests__/Checkbox.a11y.chromium.spec.ts
@@ -321,31 +321,6 @@ test('CheckboxInput keeps supporting text out of its accessible name', async ({
   expect(computed.description).toBe('Help improve the product');
 });
 
-test('the disabled SelectableCard records only its documented focusability mismatch', async ({
-  page,
-}) => {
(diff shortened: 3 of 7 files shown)


cixzhang (changes requested):
Keyboard users can now reach disabled cards, hear they are unavailable, and continue past them without changing selection. Two things still need fixing:

- Current `main` moved the shared checkbox semantics into `Checkbox.a11y.*`, so this branch conflicts and the fixed state now reports `unexpected-pass`. Please rebase, remove the stale `card-disabled` known failure and focusability exception, and keep only component-owned callback/inert coverage in `SelectableCard.test.tsx`.
- Please drop `form=""` and its assertion. This input has no `name`, so it cannot submit; detaching its form owner is an unrelated behavior change.

[Reviewed by Robohands]

Geervan:
thanks for the review, rebased onto main and applied relevant changes, let me know if i am still missing anything 
-> Removed the stale `card-disabled` known failure from `Checkbox.a11y.known-failures.ts`, the `declaredNotDelivered` exception from `Checkbox.a11y.states.ts`, and the custom mismatch test from `Checkbox.a11y.chromium.spec.ts`.
   - Retained only component-owned callback and inert coverage in `SelectableCard.test.tsx`.

Geervan:
updated as requested, just wanted to know if the review signal failing was an internal issue or something i need to fix from my side

PR #6167, opened 2026-09-08
fix(docsite): prevent stale template previews on opening and reopening

Angadslr:
Opening a template could briefly render the previous template’s preview while the header already showed the selected template.

Render direct selections immediately, reserving deferred previews for pending previous/next navigation. Retire that preview-pending state when the dialog closes or the selection changes, so unfinished navigation cannot keep a newly selected template covered by its loading skeleton.

### Regression coverage

- Opening a non-first template, including fullscreen.
- Closing and reopening another template.
- Direct index changes and initial deep-linked selection.
- Previous/next navigation and its pending skeleton.
- Closing during suspended navigation, then opening a third template before completion.

The new pending-navigation regression fails before the latest correction and passes afterward.

### Validation

- All 10 focused tests pass.
- Docsite typecheck passes.
- Changed-file lint passes.

### Bug recording

https://youtu.be/DUU9HG8Bk54

### DIFF
--- a/apps/docsite/src/__tests__/template-preview-dialog.test.ts
+++ b/apps/docsite/src/__tests__/template-preview-dialog.test.ts
@@ -6,11 +6,10 @@
  * @file Template preview selection and transition regressions.
  * @input Mounted TemplatePreviewDialog with real React scheduling and mocked UI.
  * @output Records every committed visible preview, including intermediate slugs.
- * @position Docsite regression coverage for gallery preview lifecycle.
+ * @position Docsite regression coverage for gallery preview selection and navigation.
  */
 
 import {
-  Activity,
   createElement,
   Suspense,
   useLayoutEffect,
@@ -55,8 +54,8 @@ vi.mock('@astryxdesign/core/Button', () => ({
 vi.mock('@astryxdesign/core/Skeleton', () => ({
   Skeleton: () => createElement('div', {'data-testid': 'pending-preview'}),
 }));
-// Like the native Dialog, keep children mounted while closed. React's actual
-// deferred-value and transition hooks are deliberately not mocked.
+// Keep children mounted while closed without modeling native dialog teardown.
+// React's actual deferred-value and transition hooks are deliberately not mocked.
 vi.mock('@astryxdesign/core/Dialog', () => ({
   Dialog: ({
     isOpen,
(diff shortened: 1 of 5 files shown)


cixzhang (dismissed):
The runtime fix correctly prevents the selected template from opening with stale preview content, but this PR also adds an unrelated regression for the earlier dialog-teardown fix.

In `template-preview-dialog.test.ts`, please remove `releases the dialog on playground route hiding and unmount` and the `Activity` / `HTMLDialogElement.close` harness used only by that case. The call-count assertion locks a private teardown mechanism that this change does not modify, so future dialog lifecycle work could fail a stale-preview suite for an unrelated reason. Keep the opening, direct-selection, and pending-navigation cases that prove this fix and preserve its interaction boundary.

[Reviewed by Robohands]

Angadslr:
Fixed and pushed. The dialog now renders the selected template immediately while preserving deferred previous/next navigation. Regression tests cover opening, reopening, URL changes, navigation, fullscreen behavior, and Playground teardown.

kentonquatman (approved):
The fix looks great from the design side! Thanks for providing the video showing the bug, it was very helpful!

Angadslr:
Addressed in a1a2fea1b8, on top of the existing merge commit.

- The rendered preview now uses `previewCurrent` for both its key and slug, and the skeleton is gated by `isPreviewPending`. Opening, reopening, direct URL selection, and superseding pending navigation therefore use the current selection.
- Preserved the merged optimization that unmounts previews while closed. Updated the suspended-navigation close test to verify the dialog remains closed and the preview remains unmounted before and after navigation settles.
- Restored `pnpm@11.10.0` and removed the 101-line embedded package-manager graph. Both toolchain files now match the main commit merged into this branch.

Validation:
- Focused preview regressions: **8/8 pass**. With the close assertion aligned to the merged unmount behavior, the uncorrected JSX failed 5 selection cases; all pass with the rendering correction.
- Docsite suite: **522 passed, 6 skipped**.
- `pnpm lint:strict`: passed with warnings.
- `pnpm build`: passed.
- Pre-commit checks: passed.
- The broader repository run (`pnpm test --bail=1`) did not pass: it stopped after 1,599 passing tests on `integration/pack-check.test.mjs`, with an additional `listen EPERM` error from `score-ledger.test.mjs` in the local sandbox. This is not a claim of a green full repository suite.

PR #6174, opened 2026-09-09
docs(spec): separate accessibility gap tracking

cixzhang:
## Why

Accessibility spec-test migration should keep exact known failures visible without creating public issues solely to carry operational debt. Recording a mismatch also must not imply that every accessibility finding is an approved fix.

## What

- Keep checked-in known-failure records public-safe and self-contained.
- Require a separate durable, owned gap record for each exact expectation, binding, and state.
- Record whether current authority settles the outcome or human judgment is still required.
- Keep private operational tracker references out of public source and pull-request text.
- Remove transient issue and prototype links from the long-lived contract.

## Risk

Specification only. The existing `KnownFailure.issue` implementation remains unchanged until a separate implementation PR updates it.

## Testing

- `pnpm check:knowledge`
- `git diff --check`

## Overlap search

Searched current records and open pull requests using `spec:AST-021`, `KnownFailure`, `public tracking issue`, `gap registry`, `internal/a11y-spec`, and the affected paths. The canonical owner is `docs/specs/AST-021/spec.md`. Stacked [#6175](https://github.com/facebook/astryx/pull/6175) implements the contract change. No other open pull request changes AST-021 or the known-failure tracking boundary.

### DIFF
--- a/docs/specs/AST-021/spec.md
+++ b/docs/specs/AST-021/spec.md
@@ -7,7 +7,7 @@ authority: current
 archive_reason: null
 superseded_by: null
 approved_by: cixzhang
-approved_at: 2026-09-03
+approved_at: 2026-09-08
 phase: accepted
 owners: [cixzhang]
 affects_architecture: []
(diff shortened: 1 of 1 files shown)


cixzhang (commented):

Thanks—this direction is clear. The PR body still needs the required search note; without it, the next reviewer cannot tell whether this conflicts with another contract or open change. Please add the owning path, terms checked, and stacked [#6175](https://github.com/facebook/astryx/pull/6175) plus any other hits.

[Reviewed by Robohands]

PR #6175, opened 2026-09-09
test(a11y): separate known failures from issue tracking

cixzhang:
## Why

Known-failure source belongs in the public repository, but operational gap ownership may live outside it. Requiring a public issue URL in every record leaks that operational choice into the checked-in contract.

## What

- Remove `KnownFailure.issue`.
- Keep each record exact by expectation, binding, state, layer, failure text, standards source, user impact, and reason.
- Keep unexpected passes blocking until the stale record and separately owned gap are reconciled.
- Remove issue URLs from factual reports and the current Checkbox debt record.
- Update documentation and tests to preserve the exact-match guarantees.

## Risk

Private test infrastructure only. Pattern expectations and component behavior do not change.

## Testing

- `pnpm -F @astryxdesign/a11y-spec typecheck`
- `vitest run internal/a11y-spec/src/run.test.ts` — 28 tests pass
- `git diff --check`

### DIFF
--- a/internal/a11y-spec/README.md
+++ b/internal/a11y-spec/README.md
@@ -156,16 +156,19 @@ and styling stay in the component's own suite (AST-021 FR5).
 ## Known failures
 
 A known failure names one expectation, one binding, one state, one evidence
-layer, the exact standards reference, the user impact, a public issue, and why
+layer, the exact standards reference, the exact failure, the user impact, and why
 the migration is not the place to fix it. It still runs, it still fails, and is
-reported as debt. The record matches the complete failure message, not a
+reported as debt. Operational ownership stays in the project's durable gap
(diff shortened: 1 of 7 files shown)


cixzhang (commented):

The shared `KnownFailure` type removes `issue`, but 12 Button and Checkbox records still supply it, so `@astryxdesign/core` no longer typechecks and tracker URLs remain in checked-in source. Please migrate every remaining record and stale issue reference with the type change; `pnpm -F @astryxdesign/core typecheck` should pass.

[Reviewed by Robohands]

cixzhang (commented):
Semantic verdict: approve

The prior review is resolved: all stale `issue` fields are gone, both typechecks pass, and generated reports keep operational gaps separate from public source.

[Reviewed by Robohands]

PR #6178, opened 2026-09-09
fix(CheckboxListItem): name the checkbox from a rich label's visible text

### LINKED ISSUE #6161: CheckboxListItem rich labels can fall back to the generic name Checkbox
## User impact

When `CheckboxListItem.label` is rich content and the caller omits `aria-label`, every such checkbox is exposed with the generic name "Checkbox". Accessibility consumers cannot distinguish which option each checkbox controls, and speech input cannot use the visible option text.

## Current behavior

The public type allows a ReactNode label without `aria-label`. Development builds warn, but production renders the generic fallback name. The visible rich label is not associated with the nested checkbox.

## Expected behavior

A supported rich-label item must expose a name containing the visible label words. The API should either make an equivalent plain-text name unavoidable or associate the visible label content directly with the checkbox.

## Reproduction

1. Render `CheckboxListItem` with `label={<span>Pro plan</span>}` and no `aria-label`.
2. Inspect the nested checkbox in Chromium's accessibility tree.
3. Its computed name is "Checkbox", not "Pro plan".

This issue tracks component/API remediation separately from the accessibility contract migration.

Kyujenius:
## Summary

A `ReactNode` `label` without `aria-label` left the checkbox named "Checkbox", so every rich-label item in a list announced the same (`checkbox.name.matches-visible-label` in the shared [APG checkbox](https://www.w3.org/WAI/ARIA/apg/patterns/checkbox/) contract).

The item now wraps a rich label in a `<span id>` and points the checkbox at it with `aria-labelledby` — the route `RadioListItem` already takes — so the accessible name is the visible text. `aria-label` still replaces that name when given (`aria-labelledby` would outrank it, so the two are exclusive). String labels are untouched, and the dev warning that asked for `aria-label` is gone since the name no longer depends on it.

`CheckboxList / Rich Descriptions` drops its `aria-label` to exercise the derived name; because that label contains a link, the checkbox is now named "Analytics details" rather than "Analytics". If `aria-label` replaces the derived name, it must retain every visible label word so speech-input users can say what they see.

## Test Plan

- [x] `pnpm test` — 193 passed; rich-label name and conforming aria-label override cases replace the old warning case
- [x] `pnpm test:a11y-contract` — 185 passed; `list-item-rich-label-visible-name` passes, its known-failure record removed
- [x] Chromium computes the name `"Pro plan"` for that story; `a11y:audit` 0 new
- [x] Chromium against a local `upstream/main` build: `list-item-rich-label-visible-name` name "Checkbox" → "Pro plan"; every other CheckboxList story unchanged (names, tab order, row height, hit-test). `a11y:audit` 0 new, `rtl-audit` 0 gap.

## Related Issues

Closes #6161

### DIFF
--- a/.changeset/checkbox-list-item-rich-label-name.md
+++ b/.changeset/checkbox-list-item-rich-label-name.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] CheckboxListItem: a ReactNode `label` now names the checkbox from its visible text through `aria-labelledby`, the way RadioListItem already does, instead of falling back to the generic name "Checkbox". `aria-label` still replaces that name when the visible text is absent or reads badly; the dev-time warning that asked for it is gone.
+
+@Kyujenius
--- a/.changeset/checkbox-list-item-rich-label-name.md
+++ b/.changeset/checkbox-list-item-rich-label-name.md
@@ -2,6 +2,6 @@
 '@astryxdesign/core': patch
 ---
 
-[fix] CheckboxListItem: a ReactNode `label` now names the checkbox from its visible text through `aria-labelledby`, the way RadioListItem already does, instead of falling back to the generic name "Checkbox". `aria-label` still replaces that name when the visible text is absent or reads badly; the dev-time warning that asked for it is gone.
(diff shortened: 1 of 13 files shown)


cixzhang (approved):
Exact-head Standards and Spec review is clean. The component now derives rich-label names from visible text, keeps conforming explicit overrides, and removes only the exact resolved known failure.

PR #6183, opened 2026-09-09
docs: define InputGroup admission contract

cixzhang:
## Summary

- amends the existing current `family:input-fields` authority instead of creating a Tokenizer exception
- makes InputGroup admission capability-based rather than a permanent component allowlist
- requires each admitted input to adapt size/height, outer surface geometry, border/radius/focus ownership, single-row presentation, and semantics/keyboard behavior
- admits Tokenizer once its grouped adaptation lands and requires a single-line grouped overflow default
- preserves all ungrouped input behavior and adds no public API or theme target

## Owner decision

InputGroup participation is available to any input-family member that explicitly adopts and satisfies the grouped capability contract. Tokenizer is admitted under that rule. Its standalone wrapping and overflow defaults remain unchanged; grouped Tokenizer defaults to a component-owned overflow treatment that preserves the single-line group appearance.

## Compatibility

- authority-only amendment; current main still records Tokenizer as not yet InputGroup-compatible
- no source, test, consumer-doc, public API, target, or Changeset edits
- implementation [#4405](https://github.com/facebook/astryx/pull/4405) remains separate evidence and must be re-reviewed against this authority after it lands

## Overlap receipt

- searched current InputGroup, input-field family, size-cascade, Tokenizer, and component records
- searched open PRs by InputGroup, Tokenizer, grouped input, and affected paths
- `family:input-fields` already owns InputGroup admission and grouped geometry; no separate record is justified
- open PRs [#5999](https://github.com/facebook/astryx/pull/5999), [#6000](https://github.com/facebook/astryx/pull/6000), [#5984](https://github.com/facebook/astryx/pull/5984), and [#6057](https://github.com/facebook/astryx/pull/6057) touch adjacent family/component behavior but do not own this admission rule

## Validation

- `pnpm check:sync`
- `pnpm check:repo`
- `git diff --check`

One current family record only. No implementation or Changeset.

### DIFF
--- a/docs/families/input-fields.md
+++ b/docs/families/input-fields.md
@@ -7,7 +7,7 @@ authority: current
 archive_reason: null
 superseded_by: null
 approved_by: cixzhang
-approved_at: 2026-08-31
+approved_at: 2026-09-09
 owners: [cixzhang, imdreamrunner]
 review_triggers: [behavior, layout, theming, accessibility]
 verified_by:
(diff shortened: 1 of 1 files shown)


cixzhang (commented):
Semantic verdict: approve

Thanks, this is clean. The family owns admission, shipped support stays accurate, and every grouped overflow choice remains inside the one-row contract.

[Reviewed by Robohands]

PR #6190, opened 2026-09-09
fix(CheckboxListItem): expose the description to the checkbox via aria-describedby

### LINKED ISSUE #6154: CheckboxListItem does not expose its description to assistive technology

Kyujenius:
## Summary

`CheckboxListItem` rendered `description` in the row without attaching it to the nested checkbox, so Chromium computed an empty accessible description.

The row cannot wrap the description to give it an id: that turns a plain string into a ReactNode, and `ListItem` documents that a plain string description gets single-line truncation automatically. So `Item` — which renders the description element and is the only place that can know its id — now ids that element and publishes the id to the content it renders in a slot, and the checkbox reads it there. A public `descriptionId` prop would fail `spec:AST-002/DEC-1`, since the caller decides nothing `Item` cannot derive; the seam is internal and no public API changes.

`CheckboxInput` applied its own `aria-describedby` after `{...rest}` on the same `<input>`, silently replacing a consumer's value; it now merges the incoming value ahead of its own description, status, and disabled-reason ids.

Removes the two `list-item-described` known failures and the description excuse this resolves.

Follow-up, not in this PR: `RadioListItem` still wraps its description in an id'd span, so its string descriptions already lose that truncation. It can move to the same seam once this lands.

## Test Plan

- [x] `pnpm test` — seven new cases; each fails with its half of the fix reverted
- [x] `pnpm test:a11y-contract` — `list-item-described`: `description.resolvable` and `description.exposed` pass, `unexpected-pass 0`
- [x] `a11y:audit` 0 new; `rtl-audit` 0 gap
- [x] Chromium vs a local `upstream/main` build: the checkbox's accessible description goes from `""` to the visible text across five CheckboxList stories; a string description keeps `white-space: nowrap` and `text-overflow: ellipsis`; name, checked, row height, tab order, and hit-test are identical
- [x] Not run: real screen readers, Firefox/WebKit

## Related Issues

Closes #6154

### DIFF
--- a/.changeset/checkbox-list-item-description.md
+++ b/.changeset/checkbox-list-item-description.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] CheckboxListItem: the visible `description` is now the checkbox's accessible description. The row wraps it in an id'd element and the checkbox points at it with `aria-describedby`, the way RadioListItem already does, so the browser computes a distinct description instead of none. CheckboxInput now merges a consumer-supplied `aria-describedby` with its own description, status, and disabled-reason ids rather than replacing it.
+
+@Kyujenius
--- a/.changeset/checkbox-list-item-description.md
+++ b/.changeset/checkbox-list-item-description.md
@@ -2,6 +2,6 @@
 '@astryxdesign/core': patch
 ---
 
-[fix] CheckboxListItem: the visible `description` is now the checkbox's accessible description. The row wraps it in an id'd element and the checkbox points at it with `aria-describedby`, the way RadioListItem already does, so the browser computes a distinct description instead of none. CheckboxInput now merges a consumer-supplied `aria-describedby` with its own description, status, and disabled-reason ids rather than replacing it.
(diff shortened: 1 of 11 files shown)


cixzhang (approved):
The rendered checkbox now exposes its visible supporting text as a distinct accessible description without changing the item’s truncation, name, form, or interaction behavior. Exact-head Standards and Spec review are clean; focused unit and Chromium contract checks pass, including the merged current-main state.

PR #6191, opened 2026-09-09
fix(ci): skip stale post-merge preview reruns

cixzhang:
## User impact

Rerunning CI for an already-merged pull request no longer leaves a false-red PR Comment workflow on `main`.

## Change

- Treat a source run as stale only when no open pull request remains for its head.
- Exit the trusted workflow successfully without publishing or mutating preview state.
- Continue failing closed when any open candidate is mismatched or ambiguous.

## Test plan

- `pnpm exec vitest run .github/scripts/lib/pr-preview.test.mjs .github/scripts/ci-tooling-routing.test.mjs`
- `pnpm exec prettier --check .github/scripts/lib/pr-preview.mjs .github/scripts/lib/pr-preview.test.mjs .github/workflows/pr-comment.yml`
- `git diff --check`

### DIFF
--- a/.github/scripts/lib/pr-preview.mjs
+++ b/.github/scripts/lib/pr-preview.mjs
@@ -182,6 +182,15 @@ export async function resolveWorkflowRunPullRequest({
       // accepted above may reach a privileged mutation.
     }
   }
+  if (
+    matches.length === 0 &&
+    candidates.every(pull => pull?.state !== 'open')
+  ) {
+    // A source run can be rerun after its pull request merges. No privileged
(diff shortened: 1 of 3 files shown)


cixzhang (commented):

The new assertion at `.github/scripts/lib/pr-preview.test.mjs:281` does not prove the stale path exits: deleting the new `return` leaves all 29 focused tests green, then production dereferences `identity.prNumber` and recreates the false-red workflow. Please make this guard mutation-sensitive and assert every write-capable job stays behind `resolve.outputs.valid == 'true'`.

[Reviewed by Robohands]

PR #6193, opened 2026-09-09
fix(ci): use stable fallback for CDN smoke

cixzhang:
## Why

A Version Packages PR pins an unpublished version. The CDN smoke is supposed to re-point that page at a published version, but it selected the final key in npm registry `versions` metadata. That object is not a semver-ordered API, so the test picked a canary that esm.sh repeatedly returned HTTP 408 for and made every release bump unmergeable.

## What

- keep an exact pin when it is already published
- otherwise use npm’s authoritative stable `latest` dist-tag
- fail closed when registry metadata or the tagged version is inconsistent
- cover exact-pin, stable-fallback, invalid-tag, and unavailable-registry cases

## Validation

- `pnpm exec vitest run .github/scripts/lib/cdn-version.test.mjs --project node --maxWorkers=1`
- `node .github/scripts/cdn-template-smoke-test.mjs` — rendered the button with zero console/page/request failures
- `pnpm check:repo`

No published package behavior changes; no Changeset.

### DIFF
--- a/.github/scripts/cdn-template-smoke-test.mjs
+++ b/.github/scripts/cdn-template-smoke-test.mjs
@@ -29,6 +29,7 @@ import * as http from 'node:http';
 import * as os from 'node:os';
 import * as path from 'node:path';
 import {fileURLToPath} from 'node:url';
+import {resolvePublishedCdnVersion} from './lib/cdn-version.mjs';
 
 const ROOT = path.resolve(path.dirname(fileURLToPath(import.meta.url)), '../..');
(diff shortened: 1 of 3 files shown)


cixzhang (commented):

Thanks for moving the fallback off registry key order. The fallback still accepts a prerelease when `latest` points at one, so the Version Packages check can again pass against canary bits instead of the last stable release. Please fail closed unless the fallback tag names a stable semver present in `versions`, and add a case where `latest` is a canary.

[Reviewed by Robohands]

PR #6194, opened 2026-09-09
fix(Table) useTableColumnResize: stop handles overhanging the table, reveal them on header hover

ernestt:
Two changes to the column-resize drag handles. They ship together because they come from the same piece of geometry: where a handle starts, and how far it runs.

## The fix — handles overhang, so the table scrolls vertically for no reason

The handles deliberately span the whole table, not just the header, so the indicator line reads down through the body rows while you drag. They were given the table's full height (`--table-resize-height`) but they hang from their own cell's content box, which sits a few pixels below the table's top edge. Full height, starting a few pixels down, means each handle ends a few pixels past the last row.

That would be harmless if the wrapper could not scroll, but the wrapper exists to scroll *horizontally*, and CSS promotes `overflow-y` to `auto` as soon as `overflow-x` is. So the overhang turned into ~6px of vertical scrolling on tables that had no reason to scroll at all.

The overhang itself is invisible — a handle at rest is transparent — so the symptom is all you ever see. Both shots below drive the wrapper to its scroll bottom and crop its top edge. Before, it travels 6px and the card's top border and corner ride out of frame; after, the same command moves nothing.

| before — scrolls 6px | after — nothing to scroll |
|---|---|
| before | after |

The `ResizeObserver` that already measures the table now also measures how far the handles start below its top edge and publishes it as `--table-resize-offset`, which the handles subtract from their height.

The drop is per-cell rather than uniform — a handle hangs from its cell's content box, and the header's vertical centring shifts that box by an amount that depends on how tall each header's own content is. So the **largest** drop is published: a handle stopping a fraction short of the last row is invisible, whereas one hanging past it is not.

Reading the handles from inside the observer cannot feed back into it — they are absolutely positioned, so their height never moves the table.

## The feature — a resizable column now says that it is

A boundary was only drawn once the pointer was already inside a handle's 8px hit area. You had to already suspect a column was resizable in order to discover that it was.

The header row now publishes `--resize-hint` while it is hovered, and the handles read that as their resting colour. Entering the header anywhere reveals every boundary at border weight; the handle under the pointer still goes to accent.

Identical crops, pointer resting on the second header cell in both:

| before — nothing to find | after — every boundary is drawn |
|---|---|
| before | after |

Two details worth calling out:

- Hover is caught on the **row**, not the cell, so entering the header reveals all boundaries at once rather than only the one you happen to be over.
- The handle reads the hint as its **default** value rather than matching the row's hover from inside the handle. That is deliberate: an ancestor selector (`stylex.when.ancestor`) would outrank the handle's own same-element `:hover`, and the handle under the pointer would never reach accent. Inheritance keeps that override at normal specificity.

Pointer-only, behind `@media (hover: hover)` — there is no hover to lead with on touch, where the handles are `display:none` regardless.

## Testing

- `vitest run packages/core/src/Table packages/core/src/theme/derivedVarRegistry.test.ts` — 27 files, 645 tests, all passing
- `tsc --noEmit` on `packages/core` — clean
- eslint — clean
- Measured in-browser: vertical overflow in the scroll wrapper goes from ~6px to **0**, with handle bottoms landing at or just above the last row

Screenshots were taken by swapping this one file between its `main` version and this branch's against the same running page, so the two crops in each pair differ only by the change. The theme in use neutralises the accent family to grey, so the accent state of the hovered handle is not separable from its neighbours in a still — it is left out rather than shown misleadingly.

### DIFF
--- a/.changeset/column-resize-hover-affordance.md
+++ b/.changeset/column-resize-hover-affordance.md
@@ -1,0 +1,7 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] `useTableColumnResize` now reveals its handles when the pointer enters the header. The boundaries were only drawn once the pointer was already within a handle's 8px hit area, so a resizable column gave no sign it could be resized. Every boundary now appears at border weight while the header is hovered, and the handle under the pointer still takes accent. Pointer-only, and unchanged on touch, where the handles do not render.
+
+@ernestt
--- a/.changeset/table-resize-handle-overshoot.md
+++ b/.changeset/table-resize-handle-overshoot.md
@@ -1,0 +1,22 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Stop `useTableColumnResize` handles from adding stray vertical scroll
+
+The resize handles span the whole table so their indicator line reads down
+through the body rows. They were given the table's full height but hung from a
+point a few pixels below its top edge, so each one ended that far past the last
+row. The table's scroll wrapper only exists to scroll horizontally, but CSS
+promotes its `overflow-y` to `auto` as soon as `overflow-x` is, so the overhang
+turned into ~6px of vertical scrolling on a table that had no reason to scroll.
+
+The observer that measures the table now also measures how far the handles
+start below its top edge and publishes it as `--table-resize-offset`, which the
+handles subtract from their height. The drop is per-cell — a handle hangs from
+its cell's content box, and the header's vertical centring shifts that box by an
+amount that depends on how tall each header's own content is — so the largest is
+published: a handle can stop a fraction short of the last row without anyone
+noticing, but it cannot hang past it.
+
+@ernestt
--- a/packages/core/src/Table/plugins/columnResize/useTableColumnResize.tsx
+++ b/packages/core/src/Table/plugins/columnResize/useTableColumnResize.tsx
@@ -19,6 +19,7 @@ import {colorVars} from '../../../theme/tokens.stylex';
 import type {
   TablePlugin,
   HeaderCellRenderProps,
+  HeaderRowRenderProps,
   TableColumn,
   ColumnWidth,
 } from '../../types';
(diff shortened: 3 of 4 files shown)


cixzhang (changes requested):
Thanks — showing every resize boundary on header hover is the intended design, but it changes what mouse users see across the whole header and needs to be recorded as Table’s current contract first.

Could you send that as a separate, design-only spec PR with no API changes, then rebase this PR onto it? We can approve that narrow spec update, and review the geometry change here once it lands.

[Reviewed by Robohands]

ernestt:
Pushed a third commit implementing the review decisions recorded on the spec in #6232. The two original changes in this PR are untouched; everything below is new.

## What changed

**Pointer capability gates the reveal, not the control** — `@media (pointer: coarse)` was `display: none` on the handle, which removed the separator from the accessibility tree along with the pointer target. A tablet with a keyboard attached lost column resize entirely, while the plugin's keyboard path, accessible name, and value bounds all went on promising it. The handle now stays rendered and focusable wherever any input can drive it, and coarse pointers get `pointer-events: none` instead — an invisible 8px strip that swallows header touches is worse than no target, and touch-dragging a hairline boundary is not a path this plugin supports.

**Disclosure holds while the region scrolls** — horizontal scrolling drags columns under a stationary pointer, and a boundary arriving there hasn't been asked for. One passive scroll listener writes `--resize-hint-gate` inline on the scroll wrapper, which the header row and handles inherit; `scrollend` clears it where the browser has it, a single debounced timeout where it doesn't. No React state and no rerender — there's a render-count test holding that. The gate covers the pointer-on-boundary accent as well as the row hint, for the same reason. Focus is deliberate, so it's ungated.

**Focus emphasizes one boundary** — already true (the row hint is `:hover`-only, never `:focus-within`), so this is a test and a comment locking the behavior rather than a change.

**Sticky columns and resize stop fighting** — both style the same header cell and both set `position`: resize needs `relative` to anchor its handle, sticky needs `sticky` to pin. Sticky's went through a StyleX class while its offsets were already inline, so the two landed in the same style array and the later plugin won — a pinned column silently stopped pinning depending on the order the caller wrote `plugins={{ ... }}`.

I'd said on #6232 that we'd enforce sticky-after-resize in Table's plugin order. That was wrong, and `PLUGIN_ORDER`'s own comment says why: *"It must never decide whether the table works: a plugin that renders in one order and crashes in the other is broken, and adding it here buys a lucky order rather than a fix."* Sticky now writes `position` inline alongside the offsets it was already writing there — which is what its own comment already claimed it did. Order-independent, and no Table FR9 change.

## Spec updates on #6232

Six threads answered and the record updated to match: DEC-3 (modality split), DEC-4 (touch-drag a non-goal, closing OQ3), DEC-5 (focus scope), DEC-6 (scroll gate), DEC-7 (no theme target, closing OQ2). FR6 and FR8 rewritten, FR9 and FR10 added, ORD0 and an ORD1 clause, AR6, and PR1 amended to permit exactly the one listener plus its settle signal.

Two decisions moved after implementation, and #6232 was corrected rather than left describing an intent the code doesn't match: the sticky fix above, and PR1's budget naming the settle signal the gate actually needs.

## Validation

- `packages/core/src/Table` — 553 passing across 26 files, including 5 new disclosure tests and 3 new composition tests mounting resize + sticky in both insertion orders
- `tsc --noEmit`, ESLint, Prettier, and the full `pnpm check:*` gate clean

Not covered: the browser and assistive-technology evidence the spec's verification map already tracks as pending. The coarse-pointer change is a CSS media resolution, so JSDOM can't assert it — it needs a real touch-primary device. Stale sticky pin offsets after a drag are untouched and recorded as OQ5.

humbertovirtudes (changes requested):
Resize offset invalidation only tracks the Table border box, so descendant-only handle or header geometry changes can restore the overflow defect. Header-hover disclosure and sticky-column composition also lack current authority and exact-head browser evidence for coarse pointers, keyboard focus, hover, and scroll settle. Please separate and own those contracts, cover geometry invalidation, and provide the browser evidence before approval.

PR #6195, opened 2026-09-09
feat(template): add Tree Table page template

ernestt:
> [!IMPORTANT]
> **Land after #6194.** Nothing here breaks without it, but two things this
> template relies on regress. Measured by swapping `useTableColumnResize` to
> its `main` version and back:
>
> | | on `main` today | with #6194 |
> |---|---|---|
> | table scroll wrapper | **6px overflow** | 0px |
> | resize lines on header hover | `transparent` | painted |
>
> The overflow is a stray nested scrollbar inside the table; the second is the
> discoverability of column resizing, which without #6194 is visible only once
> the pointer is already inside the 8px hit area. Both are properties of the
> plugin, not of this template — no change here is needed either way.

## What

A new page template, **Tree Table**, for the case where a table's rows form a hierarchy and every parent is a summary of what is under it.

The fixture is a code repository, chosen because it makes the rollup legible without explanation: a folder's "last commit" is self-evidently the newest commit beneath it, so the parent row visibly *derives* from its children rather than merely containing them. That is the property the template is demonstrating, and it is the one that is hard to show with flat sample data.

overview

## What it demonstrates

| | |
|---|---|
| `useTableTreeData` | parents roll up the newest commit beneath them |
| `useTableSortable` | sorting is scoped to siblings, so branches never interleave |
| `useTableColumnResize` | columns resize, with the tree column held to a floor |
| custom plugin | arrow keys walk the tree per the APG treegrid pattern |
| `Markdown` | the folder's README renders below the table |

Search prunes the tree to the branches that match and reports a live count, rather than filtering rows out of their context:

search

Selecting a folder retargets both the header trail and the README panel:

folder selected

`Breadcrumbs` wraps rather than truncates when it runs out of room, which in a single-row toolbar pushes the whole header taller. Past three crumbs the middle folds into a menu, so the trail keeps its height and both ends — the repo you are in and the folder you are looking at — stay visible:

trail menu

Narrow viewports collapse the search, branch/tag and social actions to icons on one shared breakpoint, and drop the About sidebar below the table:

narrow

Dark mode:

dark

---

## Scorecard

`Triage: new template, structural → full pass.`

| Category | Max | Score |
|---|---:|---:|
| Astryx component purity | 30 | **30** |
| Icon purity | 15 | **15** |
| Custom CSS | 15 | **12** |
| Layout & structure | 15 | **15** |
| Doc metadata | 10 | **10** |
| Image handling | 5 | **5** |
| Code quality | 10 | **10** |
| | **100** | **97 — A** |

**Findings**

- **Component purity 30/30.** No raw HTML elements; every node is an Astryx component.
- **Icon purity 15/15.** No inline SVG. All icons come from `lucide-react` through `Icon`, matching the existing template convention (`theme-showcase` and the theme icon sets). `lucide-react` is declared on `apps/template-viewer` so it no longer resolves only by hoisting.
- **Custom CSS 12/15.** Three declarations, each carrying a comment for why a prop will not do:
  - `selectedRow` paints the selection as an inset shadow, because the zebra stripe sets `background-color` on the same row and would replace a background.
  - two `paddingInline: 4px` — one to bring `Toolbar`'s inline padding onto the table's 16px content line (`Toolbar` has no padding prop), one inside `TabList` so its divider still spans the full card while the tabs inset.
- **Layout & structure 15/15.** Roots in `Layout` with `contentWidth`, header in the header slot, About in the `end` slot, no app chrome. Links are inert.
- **Doc metadata 10/10.** All six fields; four-slot description with base *and* `-able` forms of the load-bearing verbs, since the search stemmer does not relate them.
- **Image handling 5/5.** No imagery — nothing to host or strip.
- **Code quality 10/10.** `'use client'`, default export, self-contained imports, realistic fixture, clean typecheck and lint.

**What would close the last 3 points:** only a `paddingInline` prop on `Toolbar` and a padded-content variant of `TabList` would retire the two remaining spacing declarations. Both are core changes, out of scope here — the third declaration (`selectedRow`) has no prop-shaped equivalent and is the kind of escape hatch the category tolerates.

**Lifecycle flag:** this lands visible rather than `hidden: true`. Calling it out per the review instructions — it grades A, well clear of the B bar, so I have not staged it hidden. Happy to add the flag if you would rather it bake.

## Notes for review

- Row hover is off. The template is striped, and in the current theme the hover tint and the stripe resolve to the same value, so on every other row the hover was invisible; in dark mode it made the row *lighter* rather than more emphasised. Rather than ship a half-working affordance, rows rely on the link hover and the keyboard focus ring. The underlying issue — `Table` hover replacing the stripe's `background-color` instead of compositing with it — is a core one and not addressed here.
- The `Card` around the table uses `padding={0}` deliberately: besides letting the stripes reach the border, it zeroes the container padding variables that `Table`'s edge-bleed subtracts from its own margins, which is what keeps the gap above the table honest.

## Test plan

- `tsc --noEmit` and `eslint` clean.
- Verified in Chromium, light and dark, at 1500 / 900 / 680px: trail collapse past three crumbs with the folded crumb in the menu, search pruning with the live count, folder selection retargeting the trail and README, arrow-key traversal, column resize floor on the tree column.
- Confirmed the selection wash resolves to a real blue in both themes (`rgb(196,221,251)` light, `rgba(158,183,255,.24)` dark) and survives the zebra stripe.

### DIFF
--- a/.changeset/template-table-tree.md
+++ b/.changeset/template-table-tree.md
@@ -1,0 +1,16 @@
+---
+'@astryxdesign/cli': patch
+---
+
+[feat] Add the Tree Table page template
+
+A hierarchical table where every parent row is derived from its children,
+shown as a code repository: folders roll up the newest commit beneath them,
+columns resize, sorting is scoped to siblings so branches never interleave,
+arrow keys walk the tree per the APG treegrid pattern, and search prunes to
+the branches that match. Selecting a folder drives a header trail that folds
+its middle into a menu once the path outgrows the bar. Supporting repository
+chrome — header actions, an About sidebar and a README rendered with
+`Markdown` — puts the table in the context that makes its rollups legible.
+
+@ernestt
--- a/apps/docsite/src/components/templateComponents.ts
+++ b/apps/docsite/src/components/templateComponents.ts
@@ -190,6 +190,10 @@ export const TEMPLATE_COMPONENTS: Record<
     () =>
       import('../../../../packages/cli/assets/templates/pages/table-page/page'),
   ),
+  'table-tree': lazy(
+    () =>
+      import('../../../../packages/cli/assets/templates/pages/table-tree/page'),
+  ),
 };
 
 /** Resolve a template's lazy component by slug, or `undefined` if unknown. */
--- a/apps/template-viewer/package.json
+++ b/apps/template-viewer/package.json
@@ -13,6 +13,7 @@
     "@astryxdesign/theme-neutral": "*",
     "@heroicons/react": "^2.2.0",
     "@stylexjs/stylex": "^0.19.0",
+    "lucide-react": "^1.18.0",
     "react": "^19.2.7",
     "react-dom": "^19.2.7"
   },
--- a/packages/cli/assets/templates/pages/table-tree/page.tsx
+++ b/packages/cli/assets/templates/pages/table-tree/page.tsx
@@ -1,0 +1,1699 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+'use client';
+
+/**
+ * A repository you can drill into: folder, subfolder, file, in one table whose
+ * folder rows carry the newest commit from everything beneath them.
+ *
+ * A tree table earns its complexity when the parent rows are *derived from*
+ * the children. A plain folder listing does not qualify — a folder with a
+ * blank column has nothing to say at its own level, so a tree view is enough,
+ * and a table just adds a header row that is true of no row beneath it. What
+ * makes a repository qualify is that a folder has a real answer: its last
+ * commit is the most recent commit inside it. Then "src changed 2 hours ago"
+ * and "src/components is what changed" are the same question at two zoom
+ * levels, which is exactly what this shape is for.
+ *
+ * ## Extending this template
+ *
+ * **Two hooks, and neither one is the tree.** `useTableTreeState` owns the
+ * expanded set and flattens the nested data into the visible rows;
+ * `useTableTreeData` is the render plugin that draws the per-level indent and
+ * the chevron in the tree column. You pass the first one's `treeConfig`
+ * straight into the second. Hand-rolling this — a Set of open ids plus a
+ * recursive flatten on every render — is the usual way tree tables get built,
+ * and it costs you the ARIA (`aria-level`, `aria-expanded`), the expand-all
+ * control, and the guarantee that every level shares one column grid.
+ *
+ * **Sorting composes with the hierarchy instead of flattening it.**
+ * `useTableSortableState` returns an `applySort` that is handed to the tree's
+ * `sortSiblings`, so a sort reorders each sibling group in place and children
+ * never leave their parent. Sorting by date does not interleave a file with
+ * the folder that contains it.
+ *
+ * **Rollups are computed from the leaves, never stored.** `withRollups` walks
+ * the fixture once and gives every folder the newest commit beneath it, so a
+ * folder can never disagree with its contents. Message and date are copied
+ * from the *same* child rather than reduced separately — taking the newest
+ * date and the longest message would describe two different commits on one
+ * row. Hardcoding either is the bug this shape exists to prevent.
+ *
+ * **Search prunes the tree and takes the chevrons with it.** A query keeps any
+ * branch with a match anywhere beneath it and force-expands what survives, so
+ * results are never hidden behind a collapsed parent. While a query is active
+ * the expanded set is derived rather than stored, and manual toggling is
+ * suspended; clearing the box restores exactly the expansion state the user had
+ * before searching.
+ *
+ * **A pruned tree has to be re-rolled.** `withRollups` runs again on the
+ * pruned result, because removing children changes what every ancestor
+ * contains. Skipping that second pass is the subtle bug in most filtered tree
+ * tables: the rows are correct, the parents are stale, and a folder quietly
+ * advertises a commit that no surviving file beneath it carries. If you add a
+ * filter of any kind, re-roll after it.
+ *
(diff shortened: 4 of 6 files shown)


josephfarina (approved):

humbertovirtudes (dismissed):
Breadcrumb navigation can leave stale visible and semantic selection because selectedId is omitted from the keyboard plugin memo dependencies. The copyable interaction layer also lacks committed behavior tests for focus, traversal, search, selection, expansion, sorting, resizing, and tabs, while preview membership extends a manual map instead of the canonical registry. Please fix these issues and settle the ready-template decision before approval.

ernestt:
## Rebased onto current main

Was conflicting against 42 commits of `main`. Rebuilt on `1d940e0` as a single commit rather than replaying through the branch's own merge commit, which is what made the conflicts messy.

**It got smaller — 7 files to 5.** Two of the branch's changes landed on `main` independently in the meantime:

- `apps/template-viewer/package.json` — `lucide-react` is already there, so the dependency line is redundant
- `pnpm-lock.yaml` — nothing left to change once that dependency is already locked

The three conflicts that were real:

| file | resolution |
|---|---|
| `apps/docsite/src/components/templateComponents.ts` | `main` added `work-item-detail` in the same spot; both entries kept, `table-tree` before it |
| `apps/template-viewer/package.json` | `main` moved `@stylexjs/stylex` to `catalog:`; took `main`'s |
| `internal/shadcn-registry/routes.lock.json` | re-inserted the `template-table-tree` entry in its sorted position |

The template source itself — `page.tsx` and `template.doc.mjs`, 1711 of the 1736 lines — did not conflict and is unchanged.

**Also updated the header note.** It said "Land after #6194". Since that is not converging, it now says this is landing ahead of it on purpose, and keeps the measured table of what that costs (6px wrapper overflow, unpainted resize lines on header hover) — both plugin properties, unaffected either way by anything in this PR.

Verified: `eslint` clean, full pre-commit check suite green, `astryx template --list` shows Tree Table and `astryx template table-tree` emits the source. `pnpm install --frozen-lockfile` leaves the lockfile untouched, confirming it needs no entry.

josephfarina's approval was on the pre-rebase head, so it will need re-applying. humbertovirtudes' change request is untouched by this — it is about the keyboard-plugin memo dependency, the missing interaction tests, and preview membership going through the manual map instead of the canonical registry. None of those are addressed here; this is a rebase only.

ernestt:
## Fixed the selection bug; pushing back on the other two

Pushed `fddd5b2`.

### The `selectedId` dependency — real, and fixed

Both reviews found this and both are right. Traced it end to end before changing anything:

- **Selecting from the table** → `selectFolder` also calls `setExpanded` → `expandedIds` changes → `handleKeyDown` is rebuilt → the memo recomputes. Selection appears to follow, which is why this survived review.
- **Clicking a breadcrumb** (`page.tsx:1568`, `1579`) → bare `setSelectedFolderId`. `visibleData` memoizes on `[data, expandedIds, getId, getChildren, getIsExpandable, sortSiblings]`, so nothing the memo watches moves, `handleKeyDown` keeps identity, and the memo never recomputes.

The old row kept both the wash and `aria-selected={true}` while the trail and README moved on. `selectedId` is now in the dependency list.

Worth saying why this was worth fixing rather than noting: a stale `aria-selected` is a wrong answer to a screen reader, not just a stale tint, and a template is copy-paste source — shipped as-is, every app built from it inherits the defect.

I also corrected the file header, which astracat-bot was right to flag. It read *"The rows carry a cursor, not a selection … no `aria-selected`, no selected-row tint"* — the code has set both since the selection wash landed, so the note described a decision the implementation had already left behind.

**Why lint did not catch it:** `@eslint-react/exhaustive-deps` is scoped to `packages/core/src/**` (`eslint.config.js:363`). Templates are only in the `no-console: off` block. A separate question, but it is the reason this class of bug can reach a template at all.

### The regression test — the harness does not exist, and I hit four walls proving it

I tried to write the test. It is not that it is hard; there is nowhere for it to go. **No page template has a test today — 0 of 57.** In order:

1. `packages/cli` has no `@testing-library/react` — it is a devDep of `packages/core` and the root only.
2. Placed beside the template, the root Vitest config routes `packages/cli/**` to the `node` project — plain node, no DOM, no StyleX transform.
3. Moved to the docsite suite (which has Testing Library, jsdom and a real `docsite-test` job): `@astryxdesign/core/Layout` will not resolve, because the template lives inside `@astryxdesign/cli` where core is an optional peer.
4. Added the core source aliases: `Unexpected 'stylex.defineVars' call at runtime` — the docsite suite has no `@vitejs/plugin-react` + StyleX babel pipeline, so core source cannot compile there.

Clearing that means either adding test deps to the CLI package, or giving the docsite suite a second StyleX pipeline, plus new include/exclude routing in the root config. That is a test-infrastructure change that should stand on its own and apply to all 57 templates — not ride a PR that happens to add the 58th. Happy to do it as a separate PR if there is appetite.

The asks for tests covering *"focus, traversal, search, selection, expansion, sorting, resizing, and tabs"* would be the first interaction tests in the template corpus, held to a bar nothing else in it meets.

### Preview membership — the map is the convention, deliberately

`templateComponents.ts` is hand-maintained by design. Its own docblock says so:

> SYNC: when a page template is added under `packages/cli/assets/templates/pages/<slug>/`, add a matching entry here.

That is enforced by `check:sync`, and all 54 entries follow it. It is manual because each value is a `lazy(() => import('…'))` whose specifier has to be statically analyzable for the bundler to split it — a generator cannot emit that from registry data without a codegen step. Converting it is a 54-entry refactor with a build-time story attached; it is not this template's to carry.

### Ready-state

Genuinely open, and a human call rather than a diff — happy to follow whatever is decided.

PR #6198, opened 2026-09-09
fix(Chat): allow trigger menus to opt into whitespace queries (#6110)

### LINKED ISSUE #6110: [Bug] useTriggerMenu closes @ menus when the query contains whitespace

veerjain-1:
Resolves #6110

This adds an `allowWhitespace` property to `ChatComposerTrigger` which opts into continuing the search past space characters. By default, single-token grammar like `/` continues to close the menu instantly when a space is typed. But triggers like `@` mentions can opt-in to remain open to support multi-word search queries.

- `useTriggerMenu.tsx`: The scanner now only aborts on newlines, or spaces when no triggers support whitespace. It verifies the query against `allowWhitespace` for each trigger.
- `ChatComposerInput.tsx`: Exposes the `allowWhitespace?: boolean` property.

### DIFF
--- a/packages/core/src/Chat/ChatComposerInput.tsx
+++ b/packages/core/src/Chat/ChatComposerInput.tsx
@@ -107,6 +107,11 @@ export type ChatComposerTrigger = {
   /** Character that activates this trigger menu (e.g. '@', '/') */
   character: string;
   /**
+   * If true, the trigger menu remains open when the user types a space,
+   * allowing multi-word queries. Default is false (single-token only).
+   */
+  allowWhitespace?: boolean;
+  /**
    * Search source providing items for this trigger.
    * Reuses the same SearchSource interface as Typeahead \u2014
    * supports sync/async search, bootstrap, and cancel().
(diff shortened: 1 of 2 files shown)


veerjain-1:
Hi @cixzhang, @imdreamrunner, @josephfarina — just wanted to confirm the Meta CLA is now signed and this PR should be fully unblocked! Let me know if you have any feedback on the changes.

PR #6203, opened 2026-09-10
fix(cli): report one import specifier for an integration component

josephfarina:
## What

`astryx component <Name>` and `astryx search <term>` each resolved the import specifier for an integration-owned component, and they disagreed.

- `component` derived the subpath from the doc's directory, checked against the owning package's `exports`.
- `search` returned the bare package name.

For a package whose components live behind subpaths — the normal shape when one entry point exports several components — the bare name does not resolve. An agent that searched, then imported what it was told, got a file that would not compile.

## Fix

Move the resolver into `foundation/discovery/component-discovery.mjs`, beside `resolveImportPath` which does the same job for core, and have both surfaces call it. The agreement is structural now, not coincidental — which matters because this is the second time these two paths have silently diverged.

## Verification

Run against a real consumer project with an integration installed in `node_modules`, exporting `./Carousel` for a component whose doc lives in `src/Carousel/`:

| surface | before | after |
| --- | --- | --- |
| `component AcmeCarousel` | `@acme/widgets/Carousel` | `@acme/widgets/Carousel` |
| `search carousel` | `@acme/widgets` | `@acme/widgets/Carousel` |

A second fixture with no `exports` map confirms both surfaces still fall back to the bare package name, which is what a consumer would have to write by hand.

`api/component/integration-import-agreement.test.mjs` pins the property rather than the implementation: it asks both surfaces about the same component and requires the same answer, so a future divergence fails rather than ships.

## Note for reviewers

Draft because I could not run the local gate: the devserver available to me blocks npm installs and its vendored pnpm shim refuses directories outside fbsource, and a pnpm install on the container lane failed on registry retries. The change is verified end to end by execution against the released CLI, as above, but I am relying on CI for lint, typecheck and the full suite. I will mark it ready once CI is green.

Why it matters now: this blocks the autolink change (#6202). Autolink puts `search` in front of roughly 3,700 more apps, so an import path that does not resolve stops being latent and becomes the first thing a new user or agent hits.

### DIFF
--- a/.changeset/one-import-specifier-per-component.md
+++ b/.changeset/one-import-specifier-per-component.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/cli': patch
+---
+
+[fix] `component` and `search` now report the same import specifier for an integration component, resolved once in `foundation/discovery/component-discovery.mjs`. `search` previously returned the bare package name, which does not resolve for a package whose components are exported behind subpaths.
+@josephfarina
--- a/packages/cli/api/component/_adapter.mjs
+++ b/packages/cli/api/component/_adapter.mjs
@@ -28,7 +28,6 @@ import {
   findExternalComponentDoc,
   findIntegrationComponentDoc,
   findIntegrationComponentSource,
-  readPackageExports,
   resolveImportPath,
   resolveIntegrationImportPath as resolveIntegrationImport,
 } from '../../foundation/discovery/component-discovery.mjs';
@@ -401,7 +400,7 @@ export function withOwnership(docs, owner, componentName, coreDir) {
 function resolveIntegrationImportPath(owner, componentName) {
   return resolveIntegrationImport(
     {
-      exportsMap: readPackageExports(owner.integration?.__packageDir),
+      exportsMap: owner.integration?.__packageExports,
       docPath: owner.docPath,
       packageName: owner.package,
     },
--- a/packages/cli/api/component/_adapter.mjs
+++ b/packages/cli/api/component/_adapter.mjs
@@ -18,8 +18,6 @@
  * deduped, so each leaf stays a thin projection.
  */
 
-import * as fs from 'node:fs';
-import * as path from 'node:path';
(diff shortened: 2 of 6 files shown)


josephfarina:
Local gate is green, so this is out of draft.

- `vitest run packages/cli` — **235 files, 3580 tests, all passing**
- `tsc --project packages/cli/tsconfig.api-dts.json` — clean
- `eslint` on all four changed files — clean (it caught two `node:fs`/`node:path` imports the move left unused; fixed in f316f30)

Also confirmed by execution that this fixes `build` as well as `search`: `api/build/kit` calls `search()` and has no resolution of its own, so `astryx build "a carousel page"` now reports `@acme/widgets/Carousel` where it previously reported the bare package. That is the whole of kt-p3dk, not two thirds of it.

josephfarina:
Both findings were real and are fixed in 98a5f1d. Thank you — the alias case is the one that mattered, and I had shipped the same class of bug I was fixing.

**The authored override.** Sharing the resolver was not enough: `component` honors a doc-authored `import` *before* resolving a subpath, so the two had to share the precedence rule, not just the function. `search` already had the doc in hand, so it is now `doc?.import ?? resolveIntegrationImportPath(...)` — the same order `withOwnership` uses.

**The reread.** The resolver now takes an already-parsed `exports` map (`readPackageExports`), and `search` reads it once per package outside the component loop. Your arithmetic was right: it was one parse per component, so 101 reads at 100 components.

**Tests.** Three added, all through both surfaces: alias with an exports map, alias with no exports map at all, and a wildcard-only map. The wildcard case pins agreement on the *fallback* rather than the hit, since a wildcard is not matched as a literal subpath key by either surface.

I mutation-tested the precedence line: reverting `doc?.import ??` alone fails exactly the two alias tests and nothing else, so they pin the defect rather than my implementation of it.

Gate: 235 files / 3583 tests passing, `tsc --project packages/cli/tsconfig.api-dts.json` clean, eslint clean on all four changed files.

PR #6217, opened 2026-09-10
fix(Spinner): localize the default assistive label through the i18n catalog

### LINKED ISSUE #6214: Spinner: default assistive label bypasses the translation runtime
## Description

`Spinner` supplies the assistive-technology-facing default label `"Loading"` directly in component source instead of obtaining it from the Astryx translation runtime.

A localized application that relies on the default cannot translate the status. AT-facing text follows the same localization requirement as visible text. The default needs a catalog-backed translation path while preserving an explicit consumer `aria-label` override.

## Reproduction

1. Render `<Spinner />` inside an `InternationalizationProvider` using a non-English locale.
2. Inspect the Spinner accessible name.
3. Observe `"Loading"` in English.
4. Search `Spinner.tsx`; the default is a literal string rather than a translator key.

## Astryx Version

Current `main` (`@astryxdesign/core@0.5.4`).

## Environment

Chrome on macOS; DOM and source inspection.

Related: #3636

Kyujenius:
## Summary

`Spinner` supplied its default accessible name as the literal `"Loading"` in component source, so a localized app could not translate the status. It now resolves through the translation runtime as `@astryx.spinner.loading`, the same route `Button` takes for its loading announcement.

The name precedence is unchanged: an explicit `aria-label` wins, then a string `label` through `aria-labelledby`, then the localized default. The English default is still "Loading", so an app without translations sees no change.

## Test Plan

- [x] `pnpm test` — two new cases: a non-English locale gets the translated label, and an explicit `aria-label` still wins; the first fails with the literal restored
- [x] `pnpm check:i18n-catalog` — 363 keys, 29 locales consistent
- [x] Internal `<Spinner size="sm" />` call sites that rely on the default (Selector, ComplexSelector, TimeInput, Thumbnail) covered by their suites

## Related Issues

Closes #6214


Kyujenius:
Thanks!

PR #6221, opened 2026-09-10
feat(Table): add useTableStickyHeader plugin

ernestt:
Adds `useTableStickyHeader`, which pins a table's header row to the top of its scroll container so column headings stay readable while the body scrolls.

## Why

`useTableStickyColumns` already pins columns horizontally, and the two problems are the same shape: a wide comparison table is unreadable once the labels leave the viewport. Vertically there was no answer — a long table scrolled its headings away and left the reader looking at anonymous columns of numbers.

Both shots below scroll the same table by 400px. The only difference is whether the header is pinned.

| before — headings gone | after — headings pinned |
|---|---|
| before | after |

In the "before" shot the five columns are still there, but nothing says which model each one is; the reader has to scroll back up and hold five names in their head. Pinned, the selectors stay put and the comparison reads straight down.

## Using it

```tsx
const stickyHeader = useTableStickyHeader<Row>({maxHeight: '78vh'});
<Table data={data} columns={columns} plugins={{stickyHeader}} />
```

`maxHeight` gives the table a scrollport of its own. Omit it when an ancestor already bounds the height and the table simply fills it.

Worth knowing when you pick between those: a table that pins columns is already inside a wrapper with `overflow-x: auto`, and CSS promotes `overflow-y` to `auto` alongside it. That wrapper is therefore the sticky containing block, and a header inside it cannot pin to the page — so on a page-scrolled layout the plugin needs `maxHeight` to have anything to stick within. This surprised me while adopting it, so the prop docs say it too.

## Composing with sticky columns

Install both in either order and the corner cell — where a pinned column crosses the pinned header — stays above both runs. Two of the seven tests cover exactly this, once per plugin order, since the transform pipeline applies them in sequence and the z-index has to survive both.

Header cells are transparent by default, which is invisible until something scrolls underneath them. The plugin gives them the same opaque `--table-sticky-background` the pinned column already uses, so one override covers both.

## Tests

Seven, covering: `maxHeight` as a number and as a string, omitted `maxHeight`, pinning classes on every header cell, no classes when the plugin is absent, composition with `useTableStickyColumns` in both orders, and stability across re-renders.

Full `packages/core/src/Table` suite passes: 552 tests, 27 files.

Also ships a `useTableStickyHeader.doc.mjs` and four Storybook stories (`PinHeader`, `ViewportHeight`, `WithStickyColumns`, `HeightFromAncestor`).

### DIFF
--- a/packages/core/src/Table/index.ts
+++ b/packages/core/src/Table/index.ts
@@ -44,6 +44,7 @@ export {useTableColumnSettings} from './plugins/columnSettings';
 export {useTableColumnSettingsState} from './plugins/columnSettings';
 export {useTableColumnResize} from './plugins/columnResize';
 export {useTableStickyColumns} from './plugins/stickyColumns';
+export {useTableStickyHeader} from './plugins/stickyHeader';
 export {useTableGroupedRows} from './plugins/groupedRows';
 export {useTableRowIndex} from './plugins/rowIndex';
 export {useTableRowStatus} from './plugins/rowStatus';
@@ -120,6 +121,7 @@ export type {
 } from './plugins/columnSettings';
 export type {UseTableColumnResizeConfig} from './plugins/columnResize';
 export type {UseTableStickyColumnsConfig} from './plugins/stickyColumns';
+export type {UseTableStickyHeaderConfig} from './plugins/stickyHeader';
 export type {UseTableRowExpansionConfig} from './plugins/rowExpansion';
 export type {UseTableRowIndexConfig} from './plugins/rowIndex';
 export type {
--- a/packages/core/src/Table/plugins/stickyHeader/useTableStickyHeader.test.tsx
+++ b/packages/core/src/Table/plugins/stickyHeader/useTableStickyHeader.test.tsx
@@ -64,7 +64,7 @@ describe('useTableStickyHeader', () => {
   it('caps the scroll container at a numeric maxHeight, in px', () => {
     function Harness() {
       const stickyHeader = useTableStickyHeader<Row>({maxHeight: 480});
-      return <Table data={data} columns={columns} plugins={[stickyHeader]} />;
+      return <Table data={data} columns={columns} plugins={{stickyHeader}} />;
     }
     render(<Harness />);
 
@@ -74,7 +74,7 @@ describe('useTableStickyHeader', () => {
   it('passes a string maxHeight through untouched', () => {
     function Harness() {
       const stickyHeader = useTableStickyHeader<Row>({maxHeight: '60vh'});
-      return <Table data={data} columns={columns} plugins={[stickyHeader]} />;
+      return <Table data={data} columns={columns} plugins={{stickyHeader}} />;
     }
     render(<Harness />);
 
@@ -84,7 +84,7 @@ describe('useTableStickyHeader', () => {
   it('leaves the height alone when maxHeight is omitted, for callers whose ancestor already bounds the table', () => {
     function Harness() {
       const stickyHeader = useTableStickyHeader<Row>();
-      return <Table data={data} columns={columns} plugins={[stickyHeader]} />;
+      return <Table data={data} columns={columns} plugins={{stickyHeader}} />;
     }
     render(<Harness />);
 
@@ -94,7 +94,7 @@ describe('useTableStickyHeader', () => {
   it('applies one pinning class to every header cell', () => {
     function Harness() {
       const stickyHeader = useTableStickyHeader<Row>({maxHeight: 480});
-      return <Table data={data} columns={columns} plugins={[stickyHeader]} />;
+      return <Table data={data} columns={columns} plugins={{stickyHeader}} />;
     }
     render(<Harness />);
 
@@ -118,7 +118,7 @@ describe('useTableStickyHeader', () => {
 
     function Pinned() {
       const stickyHeader = useTableStickyHeader<Row>({maxHeight: 480});
-      return <Table data={data} columns={columns} plugins={[stickyHeader]} />;
+      return <Table data={data} columns={columns} plugins={{stickyHeader}} />;
     }
     render(<Pinned />);
     const pinned = headerCells()[0].className;
@@ -137,8 +137,8 @@ describe('useTableStickyHeader', () => {
           columns={columns}
           plugins={
             headerFirst
-              ? [stickyHeader, stickyColumns]
-              : [stickyColumns, stickyHeader]
+              ? {stickyHeader, stickyColumns}
+              : {stickyColumns, stickyHeader}
           }
         />
       );
@@ -165,7 +165,7 @@ describe('useTableStickyHeader', () => {
         <Table
           data={data}
           columns={columns}
-          plugins={[stickyHeader]}
+          plugins={{stickyHeader}}
           aria-label={`tick-${tick}`}
         />
       );
--- a/packages/core/src/Table/plugins/stickyHeader/useTableStickyHeader.test.tsx
+++ b/packages/core/src/Table/plugins/stickyHeader/useTableStickyHeader.test.tsx
@@ -1,0 +1,179 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+/**
+ * @file useTableStickyHeader.test.tsx
+ * @input useTableStickyHeader, Table, React testing utilities
+ * @output Functional tests for the sticky-header plugin
+ * @position Test file; validates the height cap, header pinning, composition
+ *
+ * Note: `position: sticky`, the background and the z-index are applied via
+ * StyleX (compiled to classNames), which jsdom does not resolve to
+ * `element.style`. The height cap is the one value the plugin writes as an
+ * inline style, so that is asserted directly; the StyleX-applied rules are
+ * asserted by class identity — every header cell carries the same plugin class,
+ * and it is absent when the plugin is not installed.
+ */
+
+import {describe, it, expect} from 'vitest';
+import {render, screen} from '@testing-library/react';
+import {Table} from '../../Table';
+import {useTableStickyHeader} from './useTableStickyHeader';
+import {useTableStickyColumns} from '../stickyColumns';
+import {pixel} from '../../columnUtils';
+import type {TableColumn} from '../../types';
+
+// =============================================================================
+// Test Data
+// =============================================================================
+
+interface Row extends Record<string, unknown> {
+  id: string;
+  name: string;
+  email: string;
+  team: string;
+}
+
+const data: Row[] = [
+  {id: '1', name: 'Alice', email: 'a@x.com', team: 'DS'},
+  {id: '2', name: 'Bob', email: 'b@x.com', team: 'Plat'},
+];
+
+const columns: TableColumn<Row>[] = [
+  {key: 'name', header: 'Name', width: pixel(180)},
+  {key: 'email', header: 'Email', width: pixel(220)},
+  {key: 'team', header: 'Team', width: pixel(160)},
+];
+
+function getScrollWrapper(): HTMLElement {
+  const wrapper = document.querySelector('.astryx-table-scroll-wrapper');
+  if (!(wrapper instanceof HTMLElement)) {
+    throw new Error('scroll wrapper not found');
+  }
+  return wrapper;
+}
+
+function headerCells(): HTMLElement[] {
+  return screen.getAllByRole('columnheader');
+}
+
+// =============================================================================
+// Tests
+// =============================================================================
+
+describe('useTableStickyHeader', () => {
+  it('caps the scroll container at a numeric maxHeight, in px', () => {
+    function Harness() {
+      const stickyHeader = useTableStickyHeader<Row>({maxHeight: 480});
+      return <Table data={data} columns={columns} plugins={[stickyHeader]} />;
+    }
+    render(<Harness />);
+
+    expect(getScrollWrapper().style.maxHeight).toBe('480px');
+  });
+
+  it('passes a string maxHeight through untouched', () => {
+    function Harness() {
+      const stickyHeader = useTableStickyHeader<Row>({maxHeight: '60vh'});
+      return <Table data={data} columns={columns} plugins={[stickyHeader]} />;
+    }
+    render(<Harness />);
+
+    expect(getScrollWrapper().style.maxHeight).toBe('60vh');
+  });
+
+  it('leaves the height alone when maxHeight is omitted, for callers whose ancestor already bounds the table', () => {
+    function Harness() {
+      const stickyHeader = useTableStickyHeader<Row>();
+      return <Table data={data} columns={columns} plugins={[stickyHeader]} />;
+    }
+    render(<Harness />);
+
+    expect(getScrollWrapper().style.maxHeight).toBe('');
+  });
+
+  it('applies one pinning class to every header cell', () => {
+    function Harness() {
+      const stickyHeader = useTableStickyHeader<Row>({maxHeight: 480});
+      return <Table data={data} columns={columns} plugins={[stickyHeader]} />;
+    }
+    render(<Harness />);
+
+    const cells = headerCells();
+    expect(cells).toHaveLength(3);
+
+    // Every header cell is pinned, not just the first: a partially pinned
+    // header row is the bug this guards against.
+    const classSets = cells.map(cell => cell.className);
+    expect(new Set(classSets).size).toBe(1);
+    expect(classSets[0]).not.toBe('');
+  });
+
+  it('adds classes the same table does not have without the plugin', () => {
+    function Bare() {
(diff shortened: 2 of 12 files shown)


ernestt:
## Example blocks

The hook shipped without any example block, so its docs page rendered no examples. Public hooks are expected to ship them — the docsite's `renders hook pages as hook docs with examples` test only covers hooks that have them, so a hook with none is silently skipped rather than flagged. Added two, one per behaviour the hook's own description calls out.

### `useTableStickyHeader` — Pinned Header

`maxHeight: 320` against 14 rows, so the table owns a scrollport and actually scrolls. Rows pass under the headings rather than taking them along.

| Rest | Scrolled |
| --- | --- |
|  |  |

### `useTableStickyHeader` — With Pinned Columns

Composed with `useTableStickyColumns`. Fixed column widths total 860px inside a 560px container, so it scrolls both ways and the corner where the two runs cross has something to prove.

| Light, scrolled both axes | Dark, scrolled both axes |
| --- | --- |
|  |  |

Service holds at the start edge while Environment slides under it, the headings hold at the top, and the corner cell stays opaque over both.

### Verified in a running docsite

Measured from the rendered page rather than eyeballed:

| | Pinned Header | With Pinned Columns |
| --- | --- | --- |
| scrollport height | 320px (= `maxHeight`) | 280px (= `maxHeight`) |
| content height | 590px → overflows | 417px → overflows |
| header `position` / `top` | `sticky` / `0px` | `sticky` / `0px` |
| header `z-index` | 2 | 3 at the corner |
| header background | opaque in light and dark | opaque in light and dark |
| scrolled to | top 150 | top 150, left 200 |

The z-index difference is the documented layering working: alone the header sits at 2, and where a pinned column crosses it the corner takes the 3 that `useTableStickyColumns` gives its own header cells, so it wins on both axes regardless of listing order.

### Also in this commit

The hook's JSDoc example said `plugins={[stickyHeader]}`. `Table` takes `Record<string, TablePlugin<T>>`, and this was the only array form left in the plugins directory — every other plugin documents the record form. Corrected so the snippet compiles if copied.

### Gates

`check:sync`, `check:use-client`, `check:portable-scripts`, `check:repo`, `check:changesets` pass. Docsite suite 34 files / 495 tests, Table suite 27 files / 552 tests, `tsc --noEmit` on the docsite 0 errors, eslint clean. The route lock gained exactly the two new example routes.

ernestt:
**Changeset coverage — the cli package isn't named**

This branch adds four consumer-visible files to `@astryxdesign/cli`:

- `packages/cli/assets/templates/blocks/components/Table/StickyHeaderHookUsage.tsx` / `.doc.mjs`
- `packages/cli/assets/templates/blocks/components/Table/StickyHeaderWithColumnsHookUsage.tsx` / `.doc.mjs`

The only changeset, `.changeset/table-sticky-header-plugin.md`, declares `'@astryxdesign/core': patch`, so the cli would ship the two new block templates with no release entry. `check:changesets` doesn't catch this — it validates frontmatter and body shape, not which packages the diff actually touches, which is why the gate is green.

Precedent: #5567 changed only block templates under `packages/cli/assets/templates/blocks/` and carried `'@astryxdesign/cli': patch` on its own.

Either shape works:

- add `'@astryxdesign/cli': patch` to the existing changeset's frontmatter, or
- add a second changeset naming the cli, with its own `[feat]` first line and the `@ernestt` handle line.

Both packages are pre-1.0, so `patch` is correct for a non-breaking addition.

Nothing else in the diff needs an entry: `apps/storybook` is private and `internal/shadcn-registry/routes.lock.json` is internal tooling.

packages/core/src/Table/plugins/stickyHeader/useTableStickyHeader.tsx
```diff
@@ -0,0 +1,142 @@
...
+    // meet, and one override changes all of it.
+    backgroundColor: `var(--table-sticky-background, ${colorVars['--color-background-card']})`,
+    // padding-box keeps the base off the cell's collapsed divider border.
+    backgroundClip: 'padding-box',
+    // Above unpinned body cells, and above the body cells that
+    // useTableStickyColumns pins at 1. Deliberately below the 3 that plugin
+    // gives its own header cells: where a pinned column crosses the pinned
+    // header, that corner has to win on both axes, and letting the higher value
+    // stand is what keeps it there.
+    zIndex: 2,
```
humbertovirtudes: Blocking: the corner tier is last-wins, not max-wins. Chromium with both axes scrolled produced corner `z-index: 3` and correct hit-testing for `{stickyHeader, stickyColumns}`, but `z-index: 2` and the scrolled `Reading` header overpainting the pinned `Sensor` corner for `{stickyColumns, stickyHeader}`. Please make this stacking relationship order-independent and cover both orders with browser paint and hit-testing assertions. The same composition surface should also be checked with `columnResize`, which can append `position: relative` after this rule.

packages/core/src/Table/plugins/stickyHeader/useTableStickyHeader.tsx
```diff
@@ -0,0 +1,142 @@
...
+                maxHeight:
+                  typeof maxHeight === 'number' ? `${maxHeight}px` : maxHeight,
+              };
+        return {
+          ...props,
+          htmlProps: {
+            ...props.htmlProps,
+            style: {...props.htmlProps.style, ...heightStyle},
+          },
+          xstyle: [...props.xstyle, stickyHeaderStyles.scrollWrapper],
```
humbertovirtudes, line 186: Blocking: this does not survive current main's Table scroll-owner integration. In a clean synthetic merge with current main `a3f81843b26bfadc27d486b13602598f11745461`, `useScrollableArea({axis: 'inline'})` applies its owned overflow styles after this plugin and the wrapper computes `overflow-y: hidden`. Chromium measured `clientHeight: 320`, `scrollHeight: 1557`; a wheel event left `scrollTop` at `0`, so rows are clipped and the sticky header cannot perform its primary function. Please integrate block scrolling through Table's shared scroll owner and add a current-main browser regression.

packages/core/src/Table/plugins/stickyHeader/useTableStickyHeader.test.tsx
```diff
@@ -0,0 +1,179 @@
...
+        />
+      );
+    }
+
+    for (const headerFirst of [true, false]) {
+      const {unmount} = render(<Harness headerFirst={headerFirst} />);
+
+      // The corner cell keeps the column plugin's inline offset, so both
+      // plugins reached it rather than one replacing the other's work.
+      expect(headerCells()[0].style.insetInlineStart).toBe('0px');
```
humbertovirtudes: Blocking: this test cannot fail for the behaviour in its title. It checks the sticky-column inline offset and `maxHeight`, but neither assertion observes header position or corner stacking. All 7 tests remained green after changing `position: sticky` to `static`, `insetBlockStart` to `47`, and `zIndex` to `99`; independent mutations removing position, background, z-index, and wrapper styling also survived. Please move the rendered behaviour to the existing Chromium lane and assert vertical scrolling, sticky geometry, opaque paint, both plugin orders, and corner hit-testing.

packages/core/src/Table/index.ts
```diff
@@ -44,6 +44,7 @@ export {useTableColumnSettings} from './plugins/columnSettings';
 export {useTableColumnSettingsState} from './plugins/columnSettings';
 export {useTableColumnResize} from './plugins/columnResize';
 export {useTableStickyColumns} from './plugins/stickyColumns';
+export {useTableStickyHeader} from './plugins/stickyHeader';
```
humbertovirtudes, line 47: Blocking: this exports a new stable Table module without a current canonical owner. `Table.spec.md` lists only `module:Table/useTableRowStatus` and explicitly leaves independently contractible module APIs outside the aggregate contract. Please add an approved `module:Table/useTableStickyHeader` record, link it from `Table.spec.md`, and define the behaviour, compatibility, theming, accessibility, isolation, and evidence contract before exporting the hook. This should also explicitly govern the widened `--table-sticky-background` meaning.

apps/docsite/scripts/generate-shadcn-registry.mjs
```diff
@@ -569,6 +577,14 @@ export function buildShadcnRegistry({
   const blockItems = [];
   let skippedUnpublishedBlocks = 0;
   for (const block of blocks) {
+    if (
+      dependencyTag == null &&
+      block.exampleFor != null &&
+      !availablePublicEntryNames.has(block.exampleFor)
```
humbertovirtudes, line 583: Major: this silently changes production policy for every owned registry block, but the only registry system spec, AST-026, is still draft and this policy is unrelated to the Table feature. There is also a concrete target mismatch: `generateShadcnRegistryForTarget()` validates and discards `target`, while this branch uses independently supplied `dependencyTag`. Direct reproduction shows `latest + canary` publishes unreleased blocks and `canary + null` omits them. Please split and govern this policy, derive publication mode from `target` or reject inconsistent inputs, and test through `generateShadcnRegistryForTarget()`.

packages/core/src/Table/useTableStickyHeader.doc.mjs
```diff
@@ -0,0 +1,45 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+/** @type {import('@astryxdesign/cli/authoring').ComponentDoc} */
+
+export const docs = {
+  name: 'useTableStickyHeader',
+  subComponentOf: 'Table',
+  displayName: 'useTableStickyHeader',
+  description:
+    "Hook that returns a TablePlugin which pins the header row to the top of the table's scroll container, so column headings stay readable while the body scrolls. The header pins to a scrollport the table owns, so pass maxHeight unless an ancestor already bounds the table's height. Composes with useTableStickyColumns: install both to pin a column and the header at once, and the corner where they cross stays above both.",
```
humbertovirtudes: Major: this reads as a generic Table plugin, but children mode receives only the scroll-wrapper transform. `BaseTable` applies `transformHeaderCell` solely to generated data-driven headers, then passes caller-composed children through unchanged. A focused probe confirmed the wrapper receives `maxHeight: 120px` while a caller-composed `TableHeaderCell` remains non-sticky. Please either document this as data-driven-only, matching the current Table contract, or add a supported children-mode integration seam and test it.

humbertovirtudes (changes requested):
Requesting changes based on exact-head review and a synthetic merge with current main. The primary feature fails after the current Table scroll-owner integration, plugin order changes corner paint and hit-testing, and the committed tests do not observe those behaviours. I also found authority, registry-policy, and children-mode gaps inline.

Validation: exact-head CI has 28 successful checks and 1 skipped check. The focused suite passes 7/7, but destructive CSS mutations also pass 7/7. Chromium reproduces both the corner overpaint and the current-main wheel-scroll failure.

packages/core/src/Table/plugins/stickyHeader/useTableStickyHeader.tsx
ernestt: Fixed the order-dependent stacking in `e049b36`: sticky position and the 3/2/1 corner-header-body tiers are now composition-authoritative in either plugin order, including when column resize runs last. Added focused coverage for both orders and resize composition. The separate shared block-scroll-owner and Chromium paint/hit-testing work remains open, so I am leaving this thread unresolved.

packages/core/src/Table/plugins/stickyHeader/useTableStickyHeader.test.tsx
ernestt: `e049b36` replaces the false-confidence composition assertion with checks that fail if sticky positioning, either axis, the 3/2/1 stacking tiers, or resize composition regresses. Real Chromium scroll geometry, paint, and hit-testing still depend on the separate shared block-scroll-owner decision, so this thread remains unresolved.

packages/core/src/Table/plugins/stickyHeader/useTableStickyHeader.tsx
humbertovirtudes: Blocking: the shared scroll-owner failure remains on the current head. In a clean synthetic merge with main `9f0783f1272676e998c5340c849644b6858f9e49`, Chromium computes `overflow-y: hidden` on this wrapper (`clientHeight: 320`, `scrollHeight: 1557`), and wheel input leaves `scrollTop` at `0`. Programmatic assignment can move the hidden scroll position, but users cannot scroll it. Please integrate block scrolling through Table's shared logical-axis `useScrollableArea` owner and add a current-main Chromium regression that drives wheel input.

packages/core/src/Table/plugins/stickyHeader/useTableStickyHeader.test.tsx
```diff
@@ -0,0 +1,322 @@
...
+      return <Table data={data} columns={columns} plugins={{stickyHeader}} />;
+    }
+    render(<Pinned />);
+    const pinned = headerCells()[0].className;
+
+    expect(pinned).not.toBe(bare);
+    expect(pinned.length).toBeGreaterThan(bare.length);
+  });
+
+  it('keeps sticky positioning and corner tiers in either plugin order, even with resize last', () => {
```
humbertovirtudes, line 154: Major: this is a useful improvement and now protects the 3/2/1 tiers and resize composition, but the observable scroll and paint paths are still unguarded. Removing the sticky-header background and dropping `stickyHeaderStyles.scrollWrapper` leaves all 12/12 tests green. Please add committed Chromium coverage for wheel scrolling, sticky geometry, opaque background, both plugin orders, and corner hit-testing. The wheel assertion should fail on the current synthetic merge, where `overflow-y` computes to `hidden`.

packages/core/src/Table/index.ts
humbertovirtudes: Blocking: the authority and stacking-owner finding remains open. `Table.spec.md` is unchanged, lists only `module:Table/useTableRowStatus`, and still excludes new runtime behavior and public API. The respin also publishes `--table-sticky-header-height`, widens `--table-sticky-background`, and Chromium reports `isolation: auto` on the scroll wrapper despite the new 3/2/1 local tiers. Please add an approved sticky-header module record with the Table backlink, govern both variables, and isolate the complete Table stacking surface before exporting this hook.

apps/docsite/scripts/generate-shadcn-registry.mjs
humbertovirtudes: Major: this unrelated production-registry policy remains unchanged and the target mismatch still reproduces. `generateShadcnRegistryForTarget()` validates but discards `target`; `latest + dependencyTag: 'canary'` publishes an unavailable owned block, while `canary + dependencyTag: null` omits it. Please split this policy into its own governed PR, then derive publication mode from `target` or reject inconsistent inputs and test through `generateShadcnRegistryForTarget()`.

humbertovirtudes: Major: the children-mode mismatch remains. This still describes a generic Table plugin, but `BaseTable` applies `transformHeaderCell` only to generated data-driven headers and passes caller-composed children through unchanged. The wrapper receives the height cap while a caller-composed `TableHeaderCell` remains non-sticky. Please document the hook as data-driven-only, matching the current Table contract, or add a supported children-mode integration seam with a regression test.

humbertovirtudes: Blocking scope expansion: this respin adds a separate public grouping behavior, sticky paint, a new changeset, and cross-plugin coupling through `--table-sticky-header-height`. The current Table owner explicitly excludes grouping anatomy and new runtime/API/styling, and the new tier also joins the still-unisolated Table stack. Please split `hasStickyGroupHeaders` into its own authority-first PR. The disclosure ARIA repair is restorative but separable and should also land independently.

humbertovirtudes (changes requested):
Re-reviewed exact head `e049b365be4c54a4bc98f01f80109207ae684a6a`. The order-dependent corner stacking is fixed in both plugin orders, including with column resize. Requesting changes because the shared scroll-owner failure, test gap, authority and isolation gap, registry policy mismatch, and children-mode mismatch remain. The respin also adds an independently contractible grouped-header API that needs to be split or governed.

Evidence: focused tests pass 29/29, typecheck, lint, knowledge validation, and the clean synthetic merge with main `9f0783f1272676e998c5340c849644b6858f9e49` pass. Chromium still computes `overflow-y: hidden` on a 320px wrapper with 1557px of content, and wheel input leaves `scrollTop` at `0`.

humbertovirtudes (changes requested):
Requesting changes because the public hook, scroll ownership, styling, measurement protocol, and projections conflict with current Table and AST-025 authority. A browser probe on current main computed overflow-y: hidden with 1,557px of content, and wheel input left scrollTop at zero. Please resolve the ownership and scroll semantics, then add a Table isolation boundary for header, grouped-header, and corner stacking tiers.

PR #6225, opened 2026-09-10
feat(Slider): expose the interactive control target

### LINKED ISSUE #6224: Slider disabled control opacity is not theme-reachable
## Problem

`Slider` applies its disabled opacity to the interactive track container, but that element has no theme target. The existing `slider` target is on the outer row and cannot override a descendant's `opacity: 0.5`; `slider-track` and `slider-thumb` are below the opacity boundary and cannot undo it either.

A theme that uses its own disabled treatment must therefore reach through the DOM with a structural selector for the parent of `.astryx-slider-track`. That selector depends on an implementation wrapper whose identity Astryx does not currently publish.

## Proposed contract

Expose `slider-control` on the interactive hit surface that owns pointer handling, orientation, and the composite disabled opacity. Reflect:

- `orientation`: `horizontal | vertical`
- `disabled`: `disabled`

This is the element that contains the track, fill, marks, and thumbs and receives pointer interaction. The existing `slider` target continues to own the outer row, including `valueDisplay="text"`.

## Why not move opacity to `slider`?

Moving the default opacity to the existing root target would also fade the adjacent text value, changing released visuals. Distributing opacity across each painted child would duplicate one disabled treatment across track, fill, marks, labels, and thumbs and still leave the interaction surface unnamed.

## Scope

- additive target and metadata only
- no default visual or behavioral change
- no new component prop or custom property
- update the existing draft Slider contract and generated probe coverage in the implementation PR

## Verification expected

- target/state DOM assertions
- inverse target-registry mutation
- generated probe-theme coverage
- Chromium proof that a disabled `slider-control` override changes the composite control while leaving the text value outside that target

freddymeta:
## Summary

- expose `slider-control` on Slider's pointer/keyboard interaction and compositing surface
- reflect `orientation` and `disabled` on that target
- keep the adjacent text value outside the new target
- update the draft Slider anatomy contract, focused assertions, visual-matrix story, changeset, and generated probe theme

Closes #6224.

## Evidence

Chromium against the built Storybook and probe theme:

- `slider-control` receives the probe target background: `rgb(33, 119, 237)`
- disabled control starts at the existing `opacity: 0.5`
- a theme-layer `slider-control[data-disabled="disabled"] { opacity: 1 }` moves it to `1`
- adjacent text value remains outside the target and stays at opacity `1`

The default theme receives no visual or behavioral changes.

## Test plan

- `pnpm lint:strict`
- `pnpm exec vitest run packages/core/src/Slider packages/core/src/theme/themingTargets.test.ts .github/scripts/visual-gate/lib/plan.test.mjs` — 514 passed
- inverse target mutation — registry failed on intentionally undocumented `slider-control`
- `pnpm check:knowledge`
- `pnpm visual:probe-theme:check` — 278 targets / 906 selectors
- `pnpm build`
- Chromium: `Core/Slider / Disabled With Text Value`, probe theme

`pnpm test` also ran. The broad local suite remains red in unrelated Storybook-tree and CLI integration/template-discovery tests; the dominant timeout/JSON failure class reproduces on clean `origin/main`. Slider, theming-target, visual-plan, knowledge, build, and typecheck gates above are green.

## Owner decision

`Slider.spec.md` remains draft. Please confirm that the stable interactive/compositing surface earns a public target before this is marked ready for review.

### DIFF
--- a/.changeset/expose-the-slider-interactive-control-as-a-theme-mez1.md
+++ b/.changeset/expose-the-slider-interactive-control-as-a-theme-mez1.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[feat] Expose the Slider interactive control as a theme target
+@freddymeta
--- a/apps/storybook/stories/Slider.stories.tsx
+++ b/apps/storybook/stories/Slider.stories.tsx
@@ -145,6 +145,17 @@ export const Disabled: Story = {
   },
 };
 
+export const DisabledWithTextValue: Story = {
+  tags: ['visual-theme-matrix'],
+  render: args => <Slider {...(args as any)} />,
+  args: {
+    label: 'Volume',
+    value: 50,
+    valueDisplay: 'text',
+    isDisabled: true,
+  },
+};
+
 export const VerticalOrientation: Story = {
   render: args => {
     const [value, setValue] = useState(50);
--- a/packages/core/src/Slider/Slider.doc.mjs
+++ b/packages/core/src/Slider/Slider.doc.mjs
@@ -20,6 +20,12 @@ const anatomy = [
       'Control row containing the track, thumb or thumbs, and optional text value.',
   },
   {
+    name: 'Interactive control',
+    required: true,
+    description:
+      'Pointer and keyboard interaction surface containing the rail, fill, marks, and thumbs.',
+  },
+  {
     name: 'Track',
     required: true,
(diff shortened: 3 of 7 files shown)


kentonquatman (approved):

cixzhang (approved):
Approved at f008b7824671124271e9290279f092345d62b500. Slider documents Interactive control and Thumb as separate anatomy, maps each target to its visual owner, and preserves existing behavior.

PR #6237, opened 2026-09-11
feat(templates): add the Canvas Editor page template

ernestt:
## What

A layered-artboard workspace — the design-tool pattern the gallery did not have. `Tools - Canvas Editor` is a new category beside `Tools - Page Editor`.

The distinction from Page Editor is the line the description ends on. That one composes a *document*: a palette of blocks dropped into a flow that reflows around them. This one moves objects on a fixed 1080×1920 frame, where position and size are coordinates rather than an order — so the inspector reads X/Y/W/H, moving one layer never moves another, and the canvas needs a zoom control at all.

A File/Edit/View/Object/Help menubar and document tabs head it; a layer rail and asset library sit on the left, the artboard is centered on a muted backdrop under a floating tool bar that sets zoom, and the property inspector is on the right. Both rails drag to resize.

light

## The parts worth looking at

**The inspector is the substance of it.** A text layer gets font, weight, colour, size, line height, letter spacing, horizontal and vertical alignment, slant, decoration, transform, a text shadow and a stroke; an image layer gets fit and the nine CSS filters, each on a number and a rail that move together. Everything that can reach the artboard does — typing in Transform recases the poster, dragging Sepia tints the photograph.

inspector

**The artboard is themed, not styled.** It runs under its own `defineTheme` pinned to `mode="light"`, so the poster keeps its palette when the editor goes dark. A poster is artwork — it should not invert because the tool around it did. That is the dark shot in the details below: chrome flips, artboard does not.

**Document tabs are `Item`s, not a hand-rolled strip.** `TabList` is the obvious reach and it does not fit: its label lives in internal `labelText`/`labelSizer` spans with `whiteSpace: nowrap` and no `text-overflow` anywhere in the directory, so a tab cannot truncate — and document names have to. Its selected state is an internal 2px accent underline span with no prop to suppress it, and a document tab is a filled pill. `Item` gives all of it: truncation, a fill, 28px at `density="compact"`, an icon slot and a close in `endContent`.

Two things needed care. `Item`'s 8px gap and symmetric padding left 96px for a label that wants 102, so the active tab truncated to "Salzburg pos…" — a 6px gap and a tightened end padding buy the 8px back. And `isSelected` paints `--color-accent-muted`, which in dark mode is `light-dark(#f1f1f1, #262626)` — the same `rgb(38,38,38)` as the header bar behind it, so the active tab vanished. It takes `--color-background-muted` through `xstyle` instead, which merges after `Item`'s own styles.

tabs

**A nested row action is a real `IconButton`.** It was a hand-rolled 20px `button` because the element scale stops at 28px and a control the row's own height would push a 28px row to 36. A `size="sm"` `IconButton` pulled in by the row's padding step on every edge — `margin: calc(-1 * var(--spacing-1))` — lays out as 20px while staying 28px to the pointer. The glyph does not move: it stays 18px in from the row's right edge, measured before and after.

No tooltip on these. A bubble opening off a 28px row covers the row above it, and a padlock and an × already say what they do; `label` still names them for assistive tech.

row-action

**The top bar is a `LayoutHeader`, not a `Toolbar`.** It reads like a toolbar and it is not one. A toolbar is a set of peer commands that arrow keys walk across, and the document tabs break that twice: arrowing off a tab landed on *that tab's own close button* (`Brand sheet → Close Brand sheet → Untitled draft`), and a strip of open documents is not a band of tools in the first place.

It fought the padding too. A toolbar sizes its gutters for a band heading content, and app chrome wants to sit tighter — which took a `--astryx-section-padding-inline` override plus a new `paddingBlock` prop on `Toolbar` to keep the ghost-trigger edge compensation honest. `LayoutHeader` is the component for the slot this already sat in, and `padding={1}` settles all of it: 4px on every edge puts the menu button's box exactly where the ghost inset used to pull it, so the glyph lands on the same pixel with nothing pulled back out.

Measured identical before and after — 45px bar, 18px glyph centre, 34px strip origin, same 1px divider — minus a `role` that was describing the wrong thing. **This removed the `Toolbar` core change entirely**; `Toolbar.tsx` is byte-identical to `main` again. The floating canvas tool bar stays a `Toolbar`, because that one really is a row of peer commands, and it still roves `Undo → Redo → Add frame`.

One honest note: arrow keys in the tab strip now do nothing rather than doing the wrong thing. `Tab` moves between tabs, which is what a group of buttons should do. Making the strip a true arrow-navigable `tablist` is a separate change — `TabList` itself can't take it, for the reasons above.

**One widget is hand-rolled, and it is the colour picker.** A saturation/value plane, a hue rail and a hex field. The hue rail is a real themed `Slider`; the plane is not, and cannot be — there is no two-axis picker in core, and nothing composes into one. Four of this template's seven raw tags are this widget, and it is the single largest thing standing between the score below and an A. It stays hand-rolled deliberately: the lab→core loop is longer than this template's runway.

color picker

<details>
<summary>Remaining states — dark, layer rail, menubar</summary>

dark
layer rail
menu

</details>

## Stacked on #6238

**This PR contains no `packages/core` changes**, and its base branch is #6238 rather than `main`, so GitHub computes the diff against that branch and the eleven files below are template, CLI, docsite and tooling only. Merge #6238 first and this retargets to `main` on its own.

#6238 is the one core change the template needs: `TreeListItemData` gains `xstyle`, `className`, and `style` on the row box, which is what lets `useContainerReveal` scope the layer rail's per-row action to a single row instead of the whole tree. It implements FR5 of the now-current TreeList contract.

Two earlier core changes on this branch are gone rather than split out: a `paddingBlock` prop on `Toolbar` (see the header section below for why it stopped being necessary), and a hand-rolled `stylex.defaultMarker()` reveal, replaced by `useContainerReveal`.

## Rubric score: 87/100 (B)

**Self-assessed.** The ledger fetches its rubric from the wiki and I could not reach it from here, so this is scored against the seven categories and weights in `scripts/template-score-ledger.mjs`, calibrated by measuring all 54 page templates on the same axes. Worth a maintainer's check rather than taking at face value.

| Category | Score | |
|---|---|---|
| Astryx component purity | 24/30 | 38 components, 188 element instances, **7 raw tags** in 2,935 lines |
| Icon purity | 13/15 | 44 Lucide icons, every one through `Icon`; one custom SVG factory |
| Custom CSS | 10/15 | 41 rules / 69 declarations — high, and the honest number |
| Layout & structure | 15/15 | `Layout` root, `LayoutPanel`/`Content`/`Footer`, `Resizable`, `Section`, `Toolbar` |
| Doc metadata | 10/10 | 6/6 fields, 3/3 description (349 chars / 55 words), 1/1 naming |
| Image handling | 5/5 | `/template-assets`, `AspectRatio` for the artboard, `Thumbnail` for the picker |
| Code quality | 10/10 | `tsc --noEmit` and `eslint` both clean |

Where the 13 points go:

**The 7 raw tags (−6).** Four are the colour picker above. One is the `` inside `AspectRatio`, which is that component's documented child rather than a workaround. One is a `hidden` file input that never renders — Astryx's `FileInput` is fixed at the medium element height and this row is on the 28px rhythm, so the picker is driven from the `Thumbnail` and the input just sits there to be opened. One is the SVG below. Only 20 of 54 templates carry any raw tag at all, so this is a real deduction and not a rounding error; it is also concentrated in one missing primitive.

**The custom SVG (−2).** `glyph()` draws letterforms — X, Y, W, H — on the same 24px grid Lucide uses, for the start-icon slot of the geometry fields. A design tool names a numeric field from *inside* it so the row reads as one control, and Lucide ships no letterforms. They inherit `currentColor` and the slot's sizing, so a letter and an icon are interchangeable in the same position.

**The CSS (−5).** 69 declarations over 2,935 lines is 2.4 per 100, against a median of 0 — second only to `table-filter`. It is mostly the price of the artboard: a zoomable, pannable plane and a scaled frame are geometry no layout prop expresses. Of the 69, 15 go through `var(--…)` and the rest are structural (`display`, `position`, `overflow`, `flex`). Only four are literal values, and three of them are the colour-picker thumb — a white ring with a dark halo that has to read against *any* hue the user picks, so a theme token would be the wrong answer there, not the right one.

A two-axis colour picker in core takes this to roughly 95.

## Test plan

- `pnpm -F @astryxdesign/cli typecheck` — clean.
- Header swap verified against a captured baseline: bar height, menu-button origin, glyph centre, strip origin and divider all unchanged; `[role=toolbar]` count on the page drops from 2 to 1.
- `eslint` on the template — clean.
- Full pre-commit suite (lint-staged + 13 `check:*` scripts) — clean on every commit.
- `astryx template --list` shows the new name and description, with no `(WIP)` marker.
- Eight states captured above at 2×, light and dark.
- Walked with Playwright: tabs select and close independently, the active label does not truncate (104px available against 104 needed) and stays visible in dark mode; row actions hold the row at 28px with a 28px target and the glyph fixed 18px from the edge; both rails resize; the menubar opens File/Edit/View/Object/Help; zoom drives the artboard; inspector edits reach the poster live.


ernestt:
**On the red `Flag review signals` check — it is stale, not a real failure.**

This PR was retargeted onto #6238 mid-flight. In the seconds between the retarget and the workflow reading the PR, GitHub's API served the old file count (12) alongside the new file listing (8), and the job correctly refused to classify what looked like a truncated PR:

> GitHub returned 8 of 12 changed files; refusing to classify an incomplete PR.

Re-running it cannot help: a re-run replays the original frozen event payload, so it sees the same stale 12 every time. And a fresh run can't fire either, because `review-signal.yml` is scoped to `branches: ['main']` and this PR's base is now `core-tree-list-item-classname`.

The workflow's documented recovery lever was used instead — `workflow_dispatch` with `pr=6237` — which reads live PR data and succeeded. **`review-required` is passing as a result.** The job's own red status is a historical artifact of the failed run and will clear on its own when #6238 merges and this PR retargets to `main`.

Every other check is green: 25 passing, with only `visual-acceptance` left as a human gate.

ernestt:
## The three preview surfaces were all broken

A page template shows up in three places, and the Canvas Editor rendered in none of them: the gallery card was blank, the preview dialog said no live preview was available, and the playground died on `theme_1.defineTheme is not a function`. Two unrelated causes.

### The card and the dialog share one registry

`apps/docsite/src/components/templateComponents.ts` is the lazy-import map both surfaces read, and the template was never added to it. Neither surface errors on a miss — `TemplateThumbnail` returns `null` and `TemplatePreviewSurface` swaps in the "not available" copy — so the card still listed from doc metadata with nothing inside it.

`work-item-detail` was missing the same way, with the same blank card. It is not from this branch, but the new test below covers it, so it is fixed here too.

### The playground needed more than `defineTheme`

The playground doesn't bundle templates; it evaluates their source against the hand-written module map in `generate-scope.mjs`, which exported only `Theme` from `core/theme`.

Exporting `defineTheme` stops the crash but isn't enough on its own. `ControlledTheme` was building its element as `{...props, theme, mode}`, so it overrode the theme a caller passed. The poster's display face and the colour picker's spectrum track are *nested* themes, and both would have been silently replaced by whatever the theme picker had selected. Only the outermost `Theme` answers to the picker now; a nested one keeps what it asked for.

Fixing that then uncovered `stylex.defaultMarker is not a function`. The same map's stylex mock had no marker support and no `when.ancestor`, which the template uses to reveal the layer-row controls on hover. A marker now compiles to a class the ancestor carries, and `when.ancestor` to a descendant selector — every other condition narrows the element's own selector and appends; this one qualifies a parent, so it has to prepend.

Across every page template these were the only two gaps, both in this file.

### Verified

| | |
|---|---|
| Gallery card — 1683 rendered nodes, was 0 | gallery |

Preview dialog:

dialog

Playground — no render error, and the nested themes survive: the heading computes to `Anton, Haettenschweiler, Impact…` with `text-transform: uppercase`, and the hue track to `linear-gradient(to right, rgb(255, 0, 0), …)`.

playground

### A gate, so this can't recur silently

Nothing paired a template with a registry entry — the only reminder was a `SYNC:` comment. `template-preview-coverage.test.ts` mirrors the gallery's own visibility filter, so it fails for exactly the templates a reader would find broken and names them:

```
AssertionError: expected [ 'canvas-editor' ] to deeply equal []
```

Docsite suite: 35 files, 496 tests green. `tsc --noEmit` clean.

thedjpetersen:
@ernestt this looks great - is it going to get merged?

packages/cli/assets/templates/pages/canvas-editor/page.tsx
```diff
@@ -0,0 +1,2937 @@
...
+ */
+const hueTheme = defineTheme({
+  name: 'hue',
+  components: {
+    'slider-track': {
+      base: {
+        backgroundImage:
+          'linear-gradient(to right, #f00, #ff0, #0f0, #0ff, #00f, #f0f, #f00)',
+      },
+    },
```
cixzhang: We shouldn't have component-specific themes. Merge this into the single main theme that applies all themes.

packages/cli/assets/templates/pages/canvas-editor/page.tsx
```diff
@@ -0,0 +1,2937 @@
...
+ *
+ * A scoped `defineMarker()` would be the usual choice, but StyleX only hashes
+ * one inside a `.stylex.ts` module and a template is a single file. The
+ * default marker is safe in its place here: product code compiles markers
+ * under its own prefix, so this never answers to the one Layout sets
+ * internally. Layer rows and document tabs are the only things in this file
+ * that carry it, and neither nests inside the other, so nothing else can trip
+ * the selectors below.
+ */
+const hoverScope = stylex.defaultMarker();
```
cixzhang: We should be using https://astryx.atmeta.com/components/useContainerReveal

packages/cli/assets/templates/pages/canvas-editor/page.tsx
```diff
@@ -0,0 +1,2937 @@
...
+  onToggleLock: (id: string) => void,
+): TreeListItemData[] {
+  const row = (layer: Layer): TreeListItemData => ({
+    id: layer.id,
+    label: layer.name,
+    isSelected: layer.id === selectedID,
+    onClick: () => onSelect(layer.id),
+    // Marks this row so the lock's hover selector scopes to it rather than to
+    // the whole rail — otherwise every lock in the tree reveals at once.
+    className: stylex.props(hoverScope).className,
```
cixzhang: This was what needed the update to tree list. But it seems to be trying to solve some nested hover visibility bug. For https://astryx.atmeta.com/components/useContainerReveal we need to enable both className and style at which point xstyle is probably okay too.

packages/cli/assets/templates/pages/canvas-editor/page.tsx
ernestt: Merged in 3d1e401 — `hueTheme` is gone. The `slider-track` gradient now sits in the template's one theme, renamed `canvasEditorTheme` (`name: 'canvas-editor'`), alongside the poster typography and token overrides.

It is mounted twice rather than duplicated: pinned `mode="light"` around the artboard, unpinned around the hue rail. That keeps the nine filter sliders in the same panel on the plain track, which was the only reason the override was scoped in the first place.

packages/cli/assets/templates/pages/canvas-editor/page.tsx
ernestt: Switched to it in 3d1e401. `DocumentTab` and the layer rail both call `useContainerReveal()`: the container props go on the row (`Item` for a tab, the `TreeListItemData` row for a layer) and the nested action spreads `getContentRevealProps({forceVisibility: isActive ? 'shown' : undefined})` — the active tab and a locked layer are the two that stay pinned open.

That deleted the hand-rolled half of this file: `rowAction`, `rowActionPinned`, `tabClose`, `tabClosePinned`, and the `stylex.defaultMarker()` scope they resolved against are all gone, along with the comment explaining the specificity ordering they needed. The hook's touch and `:focus-within` behavior covers what those arms were doing by hand.

packages/cli/assets/templates/pages/canvas-editor/page.tsx
ernestt: Right on both counts — 3d1e401 finishes the TreeList side and drops the marker here.

`TreeListItemData` now carries `xstyle`, `className`, and `style`, all three forwarded to the row box: `xstyle` composes into the row's own `stylex.props()`, `className` merges last, and `style` merges after the row's private `--_tree-indent` so a consumer's inline custom properties land while the indent survives. That was the missing piece — `useContainerReveal` publishes its container state as inline custom properties, so a class alone could not carry it.

The layer rail now just spreads `getContainerProps()` into the row data and `getContentRevealProps()` onto the lock, so the nested-hover scoping is the hook's inheritance rather than anything this template arranges.

Verified: 85 TreeList tests (one new, covering all three props on the row) and 11 `useContainerReveal` tests pass; core build, core typecheck, CLI strict and template-doc typechecks, and Prettier are clean.

ernestt:
Restructured against the TreeList contract that landed in #6406.

- The base branch is now `core-tree-list-item-classname` (#6238), so this PR's diff carries **zero `packages/core` files** — 11 files, all template, CLI, docsite and tooling.
- The complete `TreeList` row-styling seam moved to #6238: `xstyle`, `className`, and `style`, which is what the contract's FR5 requires and what `useContainerReveal` needs (it publishes both a class and inline custom properties). #6238 previously carried `className` alone.
- Two unrelated core edits that had ridden along on this branch — a table reflow in the DialogHeader module contract and a trailing blank line in the TextInput tests — are reverted.

Verified on the merged branch: core and neutral-theme builds, CLI strict and template-doc typechecks, and the docsite template-preview coverage test all pass. On #6238: 85 TreeList tests, core typecheck, `check:knowledge`, and `check:repo` pass.

Still open here, unchanged: the inspector controls that update state the artboard does not read.

PR #6241, opened 2026-09-11
refactor(richtext): build the editor through LexicalExtensionComposer

potatowagon:
## Summary

Moves `@astryxdesign/richtext` off `LexicalComposer` + `initialConfig` and onto
`LexicalExtensionComposer` + `defineExtension` — the composition API Lexical
documents as the migration target for React apps
([Migration Guide](https://lexical.dev/docs/extensions/migration),
[React and Lexical Extension](https://lexical.dev/docs/extensions/react)).

`initialConfig` (namespace, theme, nodes, editability, initial state, `onError`)
becomes the root extension; the editor is now built by `LexicalBuilder`, so the
package can take extension dependencies instead of only React plugins from here
on.

Both composers in the package are swapped: `RichTextEditor` and `RichTextView`.

## What does not change

This is deliberately the *minimal* migration from the guide — nothing about the
component's contract moves:

- **No public API change.** Every prop (`nodes`, `plugins`, `toolbar`,
  `transformers`, `hasMarkdownShortcuts`, `hasAutoFocus`, `defaultValue`,
  `namespace`, …) behaves as before.
- **No DOM change**, so no visual change and no screenshots: the composer is
  passed `contentEditable={null}` and `RichTextPlugin` keeps owning the
  `ContentEditable`, its placeholder and its error boundary, nested exactly
  where it was.
- **Existing React plugins stay as composer children** — markdown shortcuts,
  `OnChangePlugin`, and the package's own `TabFocusEscapePlugin`,
  `AutoFocusOnMount`, `EditorRefBridge`, `CharCountPlugin`. Legacy plugins work
  unchanged under the extension composer.
- **No new dependencies.** `defineExtension` comes from `lexical` and
  `LexicalExtensionComposer` from `@lexical/react`, both already peers.

## Two details worth a reviewer's eye

1. **The extension is built once per component instance, not per render.**
   `LexicalExtensionComposer` re-creates the editor whenever the extension's
   identity changes, whereas `LexicalComposer` read `initialConfig` exactly once
   on mount. Both components build the extension on first render and keep it in
   a ref (the same shape as the existing `themeRef`), preserving today's
   lifetime — otherwise a consumer passing an inline `nodes={[...]}` would
   discard the editor's content on every render.

   `RichTextView` clears that ref in its error branch, because the composer
   subtree unmounts there: recovering from a bad `value` then rebuilds the
   extension seeded with the corrected one, which is what the existing recovery
   test asserts.

2. **`$initialEditorState: defaultValue ?? undefined`, not `?? null`.**
   `undefined` keeps `InitialStateExtension`'s default initializer, which seeds
   the document with one empty paragraph — what `LexicalComposer` did with no
   `editorState`. `null` means "start from a genuinely empty root", and that
   difference is observable: `OnChangePlugin` skips updates whose
   `prevEditorState.isEmpty()`, so with `null` the first edit of an untouched
   editor never fires `onChange`.

## Test plan

The package's existing suite covers this migration well (initial value, read-only
and disabled editability, `onChange`, markdown shortcuts, the whole imperative
ref surface, the Tab-escape keyboard contract, and `RichTextView`'s value-sync
and error recovery) and is unchanged and green:

```
npx vitest run packages/richtext      # 86 passed
pnpm -F @astryxdesign/richtext typecheck typecheck:docs lint build
pnpm lint:strict                      # 0 errors
pnpm build                            # green
pnpm test                             # no new failures
```

No changeset: `@astryxdesign/richtext` is `private` (canary-only), which
`check:changesets` treats as not releasable.

https://github.com/user-attachments/assets/6418b34b-4ac9-470a-b0da-cfcd4e6ff557

it works, if it doesnt there will be a very obvious runtime crash

## Follow-up (not in this PR)

Converting the plugins that have 1:1 extension equivalents — `RichTextExtension`,
`HistoryExtension`, `ListExtension`, `LinkExtension`, `TabIndentationExtension` —
into extension dependencies, and refactoring the package's own effect-only
plugins (`register…`-style) into extensions. Those add peer dependencies
(`@lexical/history`, `@lexical/extension`) and change what renders, so they are
worth their own reviewable change on top of this one.

### DIFF
--- a/packages/richtext/src/RichTextEditor.tsx
+++ b/packages/richtext/src/RichTextEditor.tsx
@@ -4,8 +4,9 @@
 
 /**
  * @file RichTextEditor.tsx
- * @input Uses React, useId, Lexical (lexical + @lexical/react), Field,
- *   VisuallyHidden, useInputStatusIcon, mergeProps, design tokens
+ * @input Uses React, useId, Lexical (lexical + @lexical/react, composed through
+ *   LexicalExtensionComposer), Field, VisuallyHidden, useInputStatusIcon,
+ *   mergeProps, design tokens
  * @output Exports an accessibly labelled RichTextEditor component with a flush
  *   top toolbar slot and configurable editable-surface minimum height, RichTextEditorProps,
  *   RichTextEditorStatus, RichTextEditorStatusType, RichTextEditorSize
@@ -58,10 +59,7 @@ import {VisuallyHidden} from '@astryxdesign/core/VisuallyHidden';
 import {mergeProps, themeProps, type SizeValue} from '@astryxdesign/core/utils';
 import {useSize} from '@astryxdesign/core/SizeContext';
 
-import {
-  LexicalComposer,
-  type InitialConfigType,
-} from '@lexical/react/LexicalComposer';
+import {LexicalExtensionComposer} from '@lexical/react/LexicalExtensionComposer';
 import {useLexicalComposerContext} from '@lexical/react/LexicalComposerContext';
 import {RichTextPlugin} from '@lexical/react/LexicalRichTextPlugin';
 import {ContentEditable} from '@lexical/react/LexicalContentEditable';
@@ -83,10 +81,12 @@ import {DEFAULT_NODES} from './editorNodes';
 import {
   BLUR_COMMAND,
   COMMAND_PRIORITY_LOW,
+  defineExtension,
   KEY_DOWN_COMMAND,
   KEY_ESCAPE_COMMAND,
   KEY_TAB_COMMAND,
   mergeRegister,
+  type AnyLexicalExtension,
   type EditorState,
   type Klass,
   type LexicalEditor,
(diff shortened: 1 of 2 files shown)


kentonquatman (approved):
The design looks great! Thanks for including the recording for review!

potatowagon:
**Why move off `LexicalComposer`?** Roughly in order of weight:

**1. Extensions are where Lexical is going.** `LexicalExtensionComposer` is the
documented migration target for React apps, and new capability is landing as
extensions rather than plugins — the included set now spans a11y, clipboard,
IME, selection, and mdast-based markdown. On `initialConfig`, adopting any of
those means hand-wiring another plugin into our tree; on extensions it's a
dependency.

**2. The editor's configuration becomes a value instead of a React prop bag.**
`defineExtension` returns one thing that declares nodes, theme, and registered
behaviour together and can depend on other extensions. That's the shape this
package wants: its whole extensibility story today is "pass `nodes`, pass
`plugins`" — consumers reassemble the editor by hand. The follow-up noted in the
description (RichText/History/List/Link/TabIndentation as dependencies) is only
reachable once the composer is the extension one.

**3. The configuration stops being React-only.** The package already builds a
headless editor for markdown serialization and keeps it in step with the editor
by sharing `DEFAULT_NODES`. An extension is consumable outside React via
`buildEditorFromExtensions`, so both paths can converge on one definition rather
than a shared array plus a convention that the rest stays aligned.

**4. It is cheapest now.** The swap is behaviour-neutral — no prop, DOM, or
dependency change — and the package is canary-only with no stable consumers.
Doing the foundation move before this graduates to core keeps it an internal
refactor instead of a migration consumers have to follow later.

potatowagon:
tldr: extension composer solves the problem in classic composer where the nodes are defined seperately from the plugins, leading to hard to debug runtime issues if a node a plugin needs is not declared. in extensions, extensions (which is plugin's replacement) can declare their own nodes, so thats how the problem of missing runtime nodes is solved.

extensions composer is backwards compatible with the classic composer way of declaring nodes and plugins seperately

cixzhang (approved):

PR #6242, opened 2026-09-11
fix(Spinner): route the default assistive label through the translation runtime

### LINKED ISSUE #6214: Spinner: default assistive label bypasses the translation runtime

HelloOjasMutreja:
Fixes #6214.

## Problem

`Spinner` supplied its assistive-technology-facing default label `"Loading"` as a hardcoded literal in component source, instead of resolving it through the Astryx translation runtime. A localized application relying on the default announces the literal English word regardless of the active locale.

## Fix

Added `@astryx.spinner.loading` to `packages/core/locales/en.json` (Spinner had no prior i18n keys at all) and route the fallback through `useTranslator`, matching the established pattern already used for `typeahead.loading`, `commandPalette.loading`, and `button.loading`.

```diff
- ariaLabel ?? (typeof label === 'string' ? label : undefined) ?? 'Loading';
+ ariaLabel ??
+   (typeof label === 'string' ? label : undefined) ??
+   t('@astryx.spinner.loading');
```

An explicit `aria-label` or string `label` still wins over the translated default, unchanged. The `aria-labelledby` path (visible string `label` naming the status element) is also untouched.

## Testing

- `packages/core/src/Spinner` suite: 45/45 passing
- `pnpm check:i18n-catalog`: passes, key resolves, all 29 locales report the new key falling back to `en` as expected for an unreviewed addition
- Typecheck and lint clean

### DIFF
--- a/.changeset/spinner-loading-label-i18n.md
+++ b/.changeset/spinner-loading-label-i18n.md
@@ -1,0 +1,15 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Spinner: route the default assistive label through the translation runtime (#6214)
+
+A bare `<Spinner />` with no `aria-label` or string `label` announced the
+literal English word "Loading" to assistive technology, even inside a
+localized application. The default now resolves through `useTranslator`
+via a new `@astryx.spinner.loading` catalog key, matching the pattern
+already used for `typeahead.loading`, `commandPalette.loading`, and
+`button.loading`. An explicit `aria-label` or string `label` still wins,
+unchanged.
+
+@HelloOjasMutreja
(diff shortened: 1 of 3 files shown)


HelloOjasMutreja:
Closing as superseded. #6217 (merged 2026-09-15) shipped the identical fix: Spinner's default assistive label now resolves through the translation runtime via `@astryx.spinner.loading` in `packages/core/locales/en.json`. I rebased this branch onto current main to check and there is nothing left in `Spinner.tsx` to change; the only remaining diff was a second, now-duplicate `@astryx.spinner.loading` catalog entry.

Thanks for picking this up.

PR #6243, opened 2026-09-11
fix(Kbd): normalise the esc and return aliases useHotkeys accepts

### LINKED ISSUE #5403: [Bug] Kbd does not normalise the key aliases useHotkeys accepts

HelloOjasMutreja:
Fixes #5403.

## Problem

`useHotkeys` and `Kbd` take the same '+'-separated combo string, but each keeps its own key vocabulary. `useHotkeys` resolves `esc`/`return` (among others) through its own `KEY_ALIASES` to `escape`/`enter`. `Kbd` does not run the same step, and its fallback (`key.toUpperCase()`) is written for single letters, so these multi-character aliases render as the literal words "ESC" and "RETURN" instead of the `Esc`/`↵` glyphs and `Escape`/`Enter` accessible names `Kbd` already has for `escape`/`enter`.

As the issue itself notes, this is two different problems:
- `esc`/`return` are a plain naming gap; `KEY_DISPLAY`/`KEY_LABEL` already cover the canonical names.
- `meta`/`space` need a platform/design decision (whether `meta` branches per-platform like `mod`, and whether `space` renders as a glyph or the word "Space") that a contributor shouldn't make unilaterally.

This PR only fixes the first part.

## Fix

Added a narrow, `Kbd`-local `KEY_ALIASES` map (`esc` → `escape`, `return` → `enter` only) resolved inside `getKeyDisplay`/`getKeyLabel` before their existing lookups. `meta` and `space` are untouched.

## Testing

- `packages/core/src/Kbd` suite: 16/16 passing, including a new regression test for both aliases
- Typecheck and lint clean

### DIFF
--- a/.changeset/kbd-esc-return-aliases.md
+++ b/.changeset/kbd-esc-return-aliases.md
@@ -1,0 +1,15 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Kbd: normalise the "esc" and "return" aliases useHotkeys accepts (#5403)
+
+`useHotkeys` and `Kbd` take the same '+'-separated combo string, but `Kbd`
+did not resolve two of the aliases `useHotkeys` accepts for the same keys:
+`esc` and `return` fell through its uppercase fallback and rendered as
+the literal words "ESC" and "RETURN" instead of the existing `Esc`/`↵`
+glyphs and `Escape`/`Enter` accessible names already used for
+`escape`/`enter`. `meta` and `space` are left as-is; per the issue, those
+need a platform/design decision this fix does not make.
+
+@HelloOjasMutreja
--- a/packages/core/src/Kbd/Kbd.doc.mjs
+++ b/packages/core/src/Kbd/Kbd.doc.mjs
@@ -27,7 +27,7 @@ export const docs = {
       name: 'keys',
       type: 'string',
       description:
-        'Keyboard shortcut string. Use "+" to separate keys. Special keys: mod (Cmd on Mac), ctrl, alt, shift, enter, backspace, escape, tab, up, down, left, right.',
+        'Keyboard shortcut string. Use "+" to separate keys. Special keys: mod (Cmd on Mac), ctrl, alt, shift, enter, backspace, escape, tab, up, down, left, right. Also accepts the "esc" and "return" aliases useHotkeys uses for the same combo string.',
       required: true,
     },
     {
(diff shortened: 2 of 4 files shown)


HelloOjasMutreja:
Verified against the live Vercel preview for this PR. On the Properties playground, setting keys to:

- "esc" renders the Esc glyph with accessible name "Escape" (previously "ESC" as an uppercased literal).
- "return" renders the up-arrow-corner glyph with accessible name "Enter" (previously "RETURN").

Both match escape/enter's existing, already-correct rendering exactly. Marking ready for review.

HelloOjasMutreja:
The `pr-rtl` failure wasn't a real RTL bug, it was an undeclared coverage gap: Kbd's arrow-key glyphs are plain Unicode text (←/↑/→/↓) inside `<kbd>`, not an SVG icon the D1 auto-detector can classify, so every dimension returned N-A with no reason on file.

Read `Kbd.tsx` and every story in `Kbd.stories.tsx` before declaring it. These glyphs depict a physical keyboard key ("the Left arrow key"), not a navigational previous/next affordance, so they must stay physically left/right regardless of reading direction; mirroring them would show the wrong key. Added the `core/Kbd` entry to `verified-not-applicable.json` with that reasoning. Rebased onto current main and verified locally: `pnpm rtl:audit -- --filter Kbd` now reports `0 measured / 1 verified N-A / 0 gap`.

HelloOjasMutreja:
Superseded by #5657, which merged this same fix (esc/return alias normalization in Kbd) on 2026-09-20.

PR #6246, opened 2026-09-11
fix(shadcn): preserve JavaScript composition upgrades

josephfarina:
The 12-agent ShadCN chaos pass found two real failures in clean apps.

- JavaScript installs rewrote copied TSX to JSX, but receipts still described the TypeScript file and bytes.
- StyleX-precompiled JSX failed strict TypeScript imports.

I added receipt schema v2 with exact JavaScript install variants, pinned the full transform stack, and kept v1 receipts upgradeable. Precompiled compositions now ship as honest, idempotent JSX with a narrow tracked declaration. The full-catalog gate installs all 970 items in both TypeScript and JavaScript modes and strict-typechecks the precompiled imports.

This keeps the integration unpromoted while we soak-test it. The compatibility guide stays out of released CLI docs, public upgrade help does not advertise ShadCN, and the launch post stays draft.

Tested:
- 12 independent Vite and Next chaos lanes
- 78 focused receipt, upgrade, transform, generation, and docs tests
- 970 items and 972 routes in TypeScript and JavaScript, 1,687 files and 970 compiled sources per mode
- full package build
- strict CLI typecheck and strict repository lint

This PR is the compatibility fix only. Production route activation is separate in #6260.


kentonquatman (approved):
This was marked as "needs design review," but it probably doesn't. Design approves though! Thanks @josephfarina!

PR #6248, opened 2026-09-11
fix(ProgressBar): keep focusable marks outside progressbar subtree

### LINKED ISSUE #6219: ProgressBar: focusable target marks live inside a presentational descendant subtree
## Status

Closed pending a decision on the target-mark interaction and compatibility contract. This report is not an implementation invitation, and the existing implementation proposal remains paused until that decision is recorded.

## Observed behavior

`ProgressBar` renders every labeled target mark as a `tabIndex={0}` Tooltip trigger inside the element with `role="progressbar"`.

WAI-ARIA defines progressbar descendants as presentational. Keyboard users can tab to the mark, but the semantic ownership of that interactive content needs to be resolved before changing its ancestry or replacing its interaction.

## Reproduction

1. Render `<ProgressBar value={50} label="Progress" marks={[{value: 80, label: 'Goal'}]} />`.
2. Tab to the mark.
3. Inspect the DOM and accessibility tree.
4. Observe a focusable Tooltip trigger nested inside `role="progressbar"`.

## Reference

https://www.w3.org/TR/wai-aria-1.2/#progressbar

korkt-kim:
## Summary

- move focusable target marks outside the `role="progressbar"` subtree
- position the track and marks within a shared container
- keep existing public props and mark tooltips unchanged
- add regression coverage for mark placement and progressbar semantics

## Motivation

WAI-ARIA defines progressbar descendants as presentational, but ProgressBar currently places focusable Tooltip triggers inside that subtree. Keyboard users can reach the marks while assistive technology may flatten or omit them.

## Scope

ProgressBar markup, positioning, and unit tests only. No public API changes.

Fixes #6219.

## Test plan

- `pnpm exec vitest run packages/core/src/ProgressBar` — 52 tests passed
- `pnpm check:repo` — passed through the pre-commit hook

Browser accessibility-tree and visual checks were not run.

### DIFF
--- a/.changeset/progressbar-mark-accessibility.md
+++ b/.changeset/progressbar-mark-accessibility.md
@@ -1,0 +1,6 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] Move focusable ProgressBar target marks outside the `progressbar` subtree (#6248)
+@korkt-kim
--- a/packages/core/src/FieldStatus/__tests__/StatusMessage.a11y.states.ts
+++ b/packages/core/src/FieldStatus/__tests__/StatusMessage.a11y.states.ts
@@ -256,7 +256,7 @@ export const CORE_STATUS_MESSAGE_BINDING_STATES = [
       minValue: 0,
       maxValue: 100,
     },
-    focusSelector: '[role="progressbar"] [tabindex="0"]',
+    focusSelector: '[tabindex="0"]',
     storyId: 'a11y-status-message-pattern--progress-mark-focused-update',
   },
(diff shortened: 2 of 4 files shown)


kentonquatman (approved):

korkt-kim:
Thanks for the reviews! CI and visual acceptance are passing on the current head, 5732490, but `review-required` is still pending despite the existing approvals.

Could a code owner please check the remaining review gate and consider merging once all requirements are satisfied? Happy to address anything else that's needed.

cixzhang (approved):
Thanks for fixing this! Adding a new relative container seems good.

PR #6249, opened 2026-09-11
fix(SideNav): clip the resize handle's overlay hit area to its container

### LINKED ISSUE #6177: [Bug] Resizable SideNav inside AppShell creates a horizontal scrollbar
### Description

Enabling `resizable` on a `SideNav` inside `AppShell` creates unwanted horizontal scrolling, even with minimal navigation content and no custom theme.

**Expected**: The sidebar can be resized without introducing horizontal scrolling.

**Actual**: At the default sidebar width, its surrounding panel has `clientWidth: 260` and `scrollWidth: 261`. An absolutely positioned element inside the resize handle extends to **260.5px**, beyond the panel’s content edge. 

This reproduces with both the default AppShell variant and `variant="section"`. Removing `resizable` eliminates the overflow.

### Reproduction

```tsx
   import "@astryxdesign/core/reset.css";
   import "@astryxdesign/core/astryx.css";

   import { AppShell } from "@astryxdesign/core/AppShell";
   import { SideNav, SideNavItem } from "@astryxdesign/core/SideNav";

   export default function App() {
     return (
       <AppShell
         sideNav={
           <SideNav resizable>
             <SideNavItem label="Home" />
           </SideNav>
         }
       >
         Content
       </AppShell>
     );
   }
 ```

1. Render this example at a desktop viewport, such as 1440 × 900.
2. Scroll horizontally over the sidebar.
3. Inspect `.astryx-app-shell-sidenav`: its scroll width exceeds its client width by 1px.
4. Remove `resizable and repeat`. The overflow disappears.

### Astryx Version

@astryxdesign/core@0.5.4

### Environment

 - Initially observed on macOS.
 - Independently reproduced in Playwright Chromium on Linux.
 - Reproduction uses Astryx’s standard CSS, without a custom theme or application-specific styling.

HelloOjasMutreja:
Fixes #6177.

## Root cause

`ResizeHandle`'s `position="overlay"` mode places an intentionally oversized hit area (`hitArea`) inside the visible resize bar so it is easier to grab. That hit area is positioned with a transform offset, and on the SideNav's default 260px width that offset lands it a fraction of a pixel past the nav's own right edge.

`SideNav`'s own wrapper around the nav and the handle (`resizableContainer`) never set `overflow`, so nothing contained that sub-pixel bleed. It reached the next ancestor up, `AppShell`'s `LayoutPanel` (`astryx-app-shell-sidenav`), which does set `overflow: auto` (via its `isScrollable` prop), and the browser rounded the bleed up to a full 1px scrollable overflow there, producing the reported `clientWidth: 260` / `scrollWidth: 261` and a visible horizontal scrollbar.

`ResizeHandle`'s own overlay-mode doc comment already says: "useful when the parent has `overflow: clip`." `SideNav`'s `resizableContainer` is exactly that parent, but was missing the `overflow: clip` it was implicitly relying on.

## Fix

Adds `overflow: 'clip'` to `resizableContainer` in `packages/core/src/SideNav/SideNav.tsx`.

## Verification

- Reproduced empirically in a real browser (temporary Storybook story matching the issue's exact repro, removed before opening this PR): before the fix, `.astryx-app-shell-sidenav` measured `clientWidth: 260` / `scrollWidth: 261`; after the fix, both are `260`.
- Confirmed the visible resize handle bar itself is unaffected (`left/right/width` unchanged, `16px` wide, fully inside the panel); only the invisible sub-pixel hit-area bleed is clipped.
- Added a jsdom regression test in `SideNav.test.tsx` asserting the computed `overflow` on the handle's immediate wrapper is `clip`.
- `pnpm typecheck`, `pnpm lint`, and the full `SideNav.test.tsx` suite (200 tests) all pass.
- Changeset added.

### DIFF
--- a/.changeset/sidenav-resizable-horizontal-scrollbar.md
+++ b/.changeset/sidenav-resizable-horizontal-scrollbar.md
@@ -1,0 +1,9 @@
+---
+'@astryxdesign/core': patch
+---
+
+[fix] SideNav: stop a resizable nav's handle from creating a horizontal scrollbar in AppShell (#6177)
+
+`ResizeHandle`'s overlay mode positions an intentionally oversized hit area with a transform offset that can extend a fraction of a pixel past the nav's own width. The wrapper `SideNav` renders around the nav and handle didn't clip that bleed itself, so it reached AppShell's scrollable `LayoutPanel` and surfaced as a real 1px horizontal scrollbar. Adds `overflow: 'clip'` to that wrapper, matching what `ResizeHandle`'s own overlay-mode documentation already expects of its parent.
+
+@HelloOjasMutreja
--- a/packages/core/src/SideNav/SideNav.test.tsx
+++ b/packages/core/src/SideNav/SideNav.test.tsx
@@ -989,6 +989,18 @@ describe('SideNav resizable', () => {
     ).toBeInTheDocument();
   });
 
+  it('clips the resize handle overlay hit area to its own container (#6177)', () => {
+    // ResizeHandle's overlay mode positions its (deliberately oversized) hit
+    // area with a transform offset that can extend a fraction of a pixel
+    // past the nav's own width. The immediate wrapper around the handle
+    // must clip that bleed itself, since it sits directly inside the
+    // AppShell's scrollable LayoutPanel and would otherwise turn into a
+    // visible horizontal scrollbar there.
+    render(<SideNav resizable>Content</SideNav>);
+    const handle = screen.getByTestId('astryx-sidenav-resize-handle');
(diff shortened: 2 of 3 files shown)


HelloOjasMutreja:
Closing as superseded. #6196 merged yesterday with a more complete fix for the same bug: it clips the resizable wrapper the same way this PR does, and additionally moves the focus outline inward by the theme's outline width so the keyboard focus ring stays fully visible inside the clip boundary, which is exactly the regression flagged in the last review here. It also adds the AppShell composition story and containment test coverage that review asked for.

Confirmed by attempting to rebase onto current main: real conflict on the same lines in SideNav.tsx and SideNav.test.tsx, not a stale-diff false positive.

